PR #3056, opened 2017-08-07
Fix: Footer hides text

robert00s:
Close: #2257 
This patch should resolve issue #2257 (Footer hides text)
`Sometimes, depending on an epub style, some text is hidden by the progress bar (the mini one at the bottom).`

PR #4648, opened 2019-02-23
[fix, UX] ReaderFooter: Don't hide progressbar if it's the only ticked option

### LINKED ISSUE #3914: top progress indicator: 1) doesn't show up, 2) isn't saved to default by long click
* KOReader version: KOReader version: v2015.11-1689-g5db7bab_2018-04-22
* Device: Kobo glo hd

#### Issue #1
the bottom menu - Progress indicator: switching the toggle doesn't make the top line appear / disappear.

#### Steps to reproduce
1) I'm reading a wiki article downloaded as EPUB
2) tap on the bottom
3) last right menu item with a gear icon
4) first toggle Progress Indicator: full / mini
5) whichever position it is in, the progress indicator doesn't show up on top of the screen
6) while switching for the first time, the text shifts up "like" there was space on top "used" by the indicator, the text is getting re-drawn, but no actual change on top of the screen happens (both from full to mini and from mini to full
7) I've never seen "mini" so far; my progress bar was always:
name of the epub file on the left, pageImOn / pagesTotal, battery indicator
and the line thats getting thicker from the left as I scroll towards the end of the EPUB

BTW1: the indicator at the bottom works fine, on/off by tap and configured in the top menu -> gear icon -> last item "status line";
BTW2: all I need in the top status line is current time next to the battery; current time is the only reason I'm using the bottom status line, and it's good that I can turn it off by tapping

#### Issue #2
and long click on the same "progress indicator" being currently "full" or "mini" does ask the question "do you want to set the current value as default" - I say YES - but it never sets it as default; so the next (new) EPUB I open doesn't have any top progress bar at all; the second time I open the same (new) epub - it DOES show up as default (again, I've never seen the "mini" so I read the label as "on" or "off" instead of "full" and "mini")

Frenzie:
Fixes #3914.

Cf. #3056.

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -210,7 +210,7 @@ function ReaderFooter:init()
     self[1] = self.footer_positioner
 
     self.mode = G_reader_settings:readSetting("reader_footer_mode") or self.mode
-    if self.has_no_mode then
+    if self.has_no_mode and self.settings.disable_progress_bar then
         self.mode = MODE.off
         self.view.footer_visible = false
         self:resetLayout()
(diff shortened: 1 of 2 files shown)


Frenzie:
I'll merge this now. It seems to work okay in my testing.

PR #4640, opened 2019-02-22
[fix] GestureDetector: fix multiswipe direction detection code

Frenzie:
Unfortunately a mistake snuck into the final steps of refactoring <https://github.com/koreader/koreader/pull/4607>.

Thanks to @poire-z for pointing it out [here](https://github.com/koreader/koreader/pull/4607#issuecomment-464856789).

PR #4649, opened 2019-02-23
[fix, UX] GestureDetector: fix multiswipe length detection

Frenzie:
Reported by @poire-z, cf. https://github.com/koreader/koreader/pull/4640#issuecomment-466544922

<hr>

Apparently it's natural for me to make the second swipe slightly longer than the first, so I never noticed a logic issue. I did notice that it seemed slightly harder to make 4-swipe multiswipes than I expected it to be, but those are not necessarily easy gestures to make.

The problem was that I needed to prevent obviously silly gestures like `west west west east`. In ignoring such duplication, what I accidentally did can be illustrated in the following `west east` multiswipe:

path819

So what you see there is that at point X, the first direction in the multiswipe is detected. Every further movement west was ignored, meaning that the following swipe east could still end up as a relatively western movement overall.

By simply updating the current multiswipe slot in case of the same direction, both problems are prevented. We'll never get the same direction twice, and X moves over to where it's supposed to be on the left.

### DIFF
--- a/frontend/device/gesturedetector.lua
+++ b/frontend/device/gesturedetector.lua
@@ -517,7 +517,6 @@ function GestureDetector:handlePan(tev)
                     ["slot"] = slot,
                 },
             }
-            pan_direction = self:getPath(slot, false, fake_first_tev)
         end
 
         local msd_direction, msd_distance = self:getPath(slot, true, fake_first_tev)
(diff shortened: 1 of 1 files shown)


Frenzie:
> ok on the emulator before and after it

Well yeah, the error was *unused*. ;-D

PR #4655, opened 2019-02-24
[UX] Gesture manager: show selected action in menu

Frenzie:
As suggested by @poire-z (thanks for testing!)

screenshot_2019-02-24_15-36-16

PS Since #4652 it's feature complete as far as what I'd planned at the outset, other than coming up with the defaults.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -298,7 +298,8 @@ function ReaderGesture:buildMultiswipeMenu()
         local default_action = default_gesture[safe_multiswipe_name] and default_gesture[safe_multiswipe_name] or "nothing"
         table.insert(menu, {
             text_func = function()
-                return T(_("%1 (%2)"), friendly_multiswipe_name, action_strings[gesture_manager[safe_multiswipe_name]])
+                local action_name = gesture_manager[safe_multiswipe_name] ~= "nothing" and action_strings[gesture_manager[safe_multiswipe_name]] or _("Available")
+                return T(_("%1   (%2)"), friendly_multiswipe_name, action_name)
             end,
             sub_item_table = self:buildMenu(safe_multiswipe_name, default_action),
             hold_callback = function(touchmenu_instance)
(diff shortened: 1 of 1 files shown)


poire-z:
Tested, looking and working fine.

I would may be add some spacing between the arrows and the (action) - I used 3 spaces [here](https://github.com/koreader/koreader/blob/ab1132e885ce8eb53e79d0a4eb52837eabcfac4b/frontend/apps/reader/modules/readerfont.lua#L60-L65).
And may be just don't show anything (keep it blank as previously) instead of `(nothing)` - clearer to see what's unnassigned (meaning: available, or no action).

> it's feature complete as far as what I'd planned at the outset, other than coming up with the defaults.

Yep, that, and where to put this Gesture manager menu item :)

KenMaltby:
"Available" might be a nice hint that the user could specify a function for that gesture.

Frenzie:
screenshot_2019-02-24_17-00-23

poire-z:
Not found of `(Available)` as it does not really stand out of the assigned ones (`(Bookmarks)`  has the same length). But ok, I won't visit that so often once I'm settled :)

Frenzie:
@poire-z What about putting `Links` and `Gesture manager` under Navigation?

poire-z:
Dunno. It could make sense (but there are `Screen> Activate menu>` which can sound like Navigation, as much as `Navigation> Back key to exit`).

Other ideas: `Key navigation>` (for the existing Navigation) and `Gestures>`with `Activate menu>`, `Links>`, `Gesture Manager>` (and `Disable double tap`?).

But I'd really miss the `Links>` at a first level submenu (may be only because I often change them for dev and testing, most users may settle on some settings and hardly go back there).

Frenzie:
I quite like the `Gestures` menu idea, except for the part where calling a tap a "gesture" is a bit of a stretch outside of our own technical context. It also pertains to plenty of tapping-related stuff, after all. I think the kernel speaks of `Touches` but that just seems weird.

Classic Opera puts the whole shebang in `Shortcuts`.

Some thoughts:
* Taps & gestures
* Language & input

PR #4623, opened 2019-02-19
[fix] Gesture manager: check multiswipe defaults

Frenzie:
Pointed out by @poire-z

Also add default previous location action for left, down.

PR #4657, opened 2019-02-24
[UX] Gesture manager: add action - go to latest bookmark

Frenzie:


poire-z:
>     multiswipe_west_east = "previous_location",
>     multiswipe_east_west = "latest_bookmark",

Same [old play of ours](https://github.com/koreader/koreader/pull/4623#issuecomment-465304421) about gestures :| are we making a gesture or dragging the page?
For me, your choice feels unnatural - I would swap these two :|

Same thoughts I had in https://github.com/koreader/koreader/pull/4623#issuecomment-466748426 (and next comment) are in play too: in two swipe gestures:
A) is the first one the intention, and the second some kind of release
B) is the first one just taking some room to make the 2nd, and the second one the intention.

Your choice here would work in the mindset B. (But with north/south, for History vs Skim, I initially went with B, but A feels more natural...).

Am I making sense in my way of thinking about that? (left aside your habits with mouse gestures in browsers :)

Frenzie:
Isn't there a setting somewhere to swap left/right behavior?

poire-z:
Hey, I wasn't saying we should use what I feel natural. I'm not even sure after all this today testing :)

Does that reverse make sense for you? Or not, and that's why you asking for a swap left/right toggle?

And what about my stuff about these 2 mindsets? Any of these make more sense to you?

May be you have that habbit with east and west and browser mouse gestures, and it shows even with single swipes.
But what about north and south? What would be your natural gesture (single swipe) to bring a bigger progress bar from the bottom progress bar?
(That swap left/right toggle could be a good thing for those like you who this gesture vs dragging the way you do. But no swap north/south then?)

Frenzie:
For the record, I almost exclusively turn pages using the regular swipes. I don't particularly want to switch anything in that regard.

In any case, without this action there can't be any proper testing. Although I strongly suspect this latest change is quite horribly wrong. :-P

> What would be your natural gesture (single swipe) to bring a bigger progress bar from the bottom progress bar?

Um, the one we have of course? I'm not really following the question tbh.

poire-z:
> Um, the one we have of course? I'm not really following the question 

I'm just trying to see what's people natural mindset related to swipes and double swipes.

The one we have (to bring the Skim widget) is hold on the footer.
As there's this thing with swipe left and right vs dragging page / doing gestures, I switched the discussion to something less left/right, just to see.)

Going to open a new issue/question and ping some people, to gather some naive opinion without all my questions :)

PR #4660, opened 2019-02-25
[UX] Gesture manager: more multiswipe defaults

Frenzie:


poire-z:
(Not discussing your choices.)
But did/do you have a logical thinking flow when making these out, and remembering them? Especially for the zoom ones :)
Can you share it? (I wonder if they should come with the thinking flow detailed in words - or if I have to play with it and guess it so the thinking unfolds naturally and it becomes my own :)

Frenzie:
The regular ones are based on your suggestions regarding a certain attempt at consistency with the extant two-finger swipes. The rationale behind the zoom gestures is this:

zoom-gestures

poire-z:
OK, thanks. Easier to grasp with some drawing and explanation, and material real life idea to associate :)

(I don't use them often, and I would have naively go about thinking about gestures for them with:
- some 2 always the same swipes D1 D2 indicating `Switch zoom to fit`
- a 3rd one D3 to select either `content` or `page`
- a 4th one D4, horizontal for `width`, vertical for `height`)

Issue #4650, opened 2019-02-24
In scroll mode, the minibar hides the very last line of text


Biep:
* KOReader version: v2019-02-43-g1632acc_2019-02-16
* Device: Kobo Aura One

#### Issue
When I read a text file (maybe without final newling), the minibar appears on top of the last line.  I need to toggle the minibar to be able to read that last line

#### Steps to reproduce
Make sure the minibar is visible.  Open a text file, choose scroll mode on from the columns icon on the bottom menu, and scroll all the way down.

poire-z:
Scroll mode is a bit limited. But the other half of that last line half-cut should be visible on the next page (which indeed make it hard to read). Not much to do about that, except using that Overlap setting. Or just using page mode.
Some details and reasoning in https://github.com/koreader/koreader/issues/4533#issuecomment-459630636 , #3870, #4538

Biep:
When I get to the end of the file, and see all but the last line, and then try to scroll more, nothing happens.  When I click at the right side of the screen, I get the "End of book" menu.  When I set overlap (1 line), the whole page greys out when trying to page to the last line.  Without hiding the minibar there seems to be no way to see that line.

My use case is looking at code.  I want to be able to scroll, in order to see a full function at once, without having to flip forward and back.

poire-z:
Oh, I see, the _very very last_ line, indeed.

Frenzie:
I think PDF/DjVu may also have an issue like that. Not 100 % sure.

poire-z:
For cre, there seems to be a simple solution:
```c
/// set vertical position of view inside document
int SetPos( int pos, bool savePos=true, bool allowScrollAfterEnd = false);
```
https://github.com/koreader/crengine/blob/master/crengine/src/lvdocview.cpp#L2030-L2038

We call that from cre.cpp when setting the position in scroll mode:
https://github.com/koreader/koreader-base/blob/15becff11c1b23edbd3f8c74cd949fc0e219d115/cre.cpp#L871-L875
```c
-       doc->text_view->SetPos(pos);
+       doc->text_view->SetPos(pos, true, true);
```
seems to fix that (by somehow adding some kind of blank page after last book line, so you can scroll it up... up to the last to last page - but I guess that's better than not seeing it).

Or we can add a 4th parameter: some extra height to allow - and would we give it the size of the minibar.

Frenzie:
Adding a blank page seems a bit unexpected but it's probably better than hiding a line. :-)

[...]

PR #4661, opened 2019-02-25
[UX] Alternative update message when ota_version < local_version

Frenzie:
See https://github.com/koreader/koreader/issues/4650#issuecomment-466965983


Frenzie:
Updated with @Biep's downgrade suggestion:
filemanager_2019-feb-25_124300

Galunid:
@Frenzie I believe in this case "Would you still like to **downgrade**" would be more fitting.

PR #4664, opened 2019-02-26
[UX] Gesture manager: add action - show frontlight dialog

Frenzie:
Rationale for the default `down, left` gesture: it's pointing at the location for the frontlight toggle.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -37,6 +37,7 @@ local action_strings = {
     full_refresh = _("Full screen refresh"),
     night_mode = _("Night mode"),
     suspend = _("Suspend"),
+    show_frontlight_dialog = _("Show frontlight dialog"),
     toggle_frontlight = _("Toggle frontlight"),
     toggle_gsensor = _("Toggle accelerometer"),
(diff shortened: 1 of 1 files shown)


gerhaher:
Would it be possible to also add a natural light toggle?

Frenzie:
@gerhaher The natural light toggles separately from frontlight?

gerhaher:
Yes.

Frenzie:
How do you go about toggling it precisely? (E.g., what's the text on the button?)

gerhaher:
There is no toggle button. Instead you choose the level from 0 to 100 (if you don't prefer the auto feature).

reader_2019-feb-26_094639

Frenzie:
Do you mean that this PR only brings up the regular frontlight dialog instead of the full extended dialog on supported platforms?

Frenzie:
@gerhaher No, the frontlight dialog does that all by itself. I'm afraid I don't understand what you're asking.

gerhaher:
OK. 
Never mind.

NiLuJe:
Not generally possible, as there's no notion of on or off for natural light. Both edges of the bar are the best you'll ever get, as it's quite literally a mix of color leds that depend on the frontlight's brightness.

PR #4669, opened 2019-02-27
[fix] TimeVal: add dbg:guard against incorrect subtraction order

Frenzie:
In principle, any negative subtraction result should be caused by a logical error. Cf. https://github.com/koreader/koreader/pull/4666#discussion_r260772668

@poire-z @NiLuJe Of course it could also be made into a more forgiving logger.warn, or nothing at all if there are legitimate cases where the result could be negative.

### DIFF
--- a/spec/unit/timeval_spec.lua
+++ b/spec/unit/timeval_spec.lua
@@ -1,0 +1,34 @@
+describe("TimeVal module", function()
+    local TimeVal, dbg, dbg_on
+    setup(function()
+        require("commonrequire")
+        TimeVal = require("ui/timeval")
+        dbg = require("dbg")
+        dbg_on = dbg.is_on
+    end)
+
+    after_each(function()
+        if dbg_on then
+            dbg:turnOn()
+        else
+            dbg:turnOff()
+        end
+    end)
+
+    it("should subtract", function()
+        local timev1 = TimeVal:new{ sec = 5, usec = 5000}
+        local timev2 = TimeVal:new{ sec = 10, usec = 6000}
+
+        assert.is.same({sec = 5,usec = 1000}, timev2 - timev1)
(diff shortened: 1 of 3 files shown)


spec/unit/timeval_spec.lua
```diff
@@ -0,0 +1,33 @@
...
+        assert.is.same({sec = -5,usec = -1000}, timev1 - timev2)
+    end)
+
+    it("should guard against reverse subtraction logic", function()
+        dbg:turnOn()
+        TimeVal = package.reload("ui/timeval")
+        local timev1 = TimeVal:new{ sec = 5, usec = 5000}
+        local timev2 = TimeVal:new{ sec = 10, usec = 5000}
+
+        assert.has.errors(function() return timev1 - timev2 end)
```
Frenzie: Alternatively you can also check `has.error(function() end, "error message")`.

poire-z:
Dunno much about the specifics of what a TimeVal is.
But I guess that i should be able to substract 2 timeval and get a negative TimeVal if the 2nd is more recent than the 1st. That's for the calling code to deal with I would think, and to do a abs() or order correctly the items when calling.

Frenzie:
But is there a legitimate reason for that to ever happen? (Also note it's a debug mode protection only, in normal mode the program will soldier on troublefree.)

poire-z:
I don't know, I don't know all the code of koreader and which is using it and how.
But if there is a substraction function (or comparison function that would use it), might be it could be used in an anonymous sorting function to sort things by timeval where getting them in any order could happen).

But ok if it's debug only, we will notice it I guess.
(And if it's safe with the way the current code uses it, you should specify that requirement in the comment/blurb/doc at top of file then)

Frenzie:
For a sorting function you should just use less than or less or equal? I'll merge it this way and if a legitimate use pops up it's easy to get rid of. :-)

PR #4677, opened 2019-02-28
[UX] Introduce Taps & Gestures menu

Frenzie:
This more logically groups together various settings related to taps and gestures.

Cf. <https://github.com/koreader/koreader/pull/4655#issuecomment-466808273>.


poire-z:
Looks ok. (And as long there are only 9 items in gear and I can bring back Links> to it as the 10th via menusorter, I'm fine :)
Navigation to be renamed `Key navigation`, if it's all dependant of the device having keys?

poire-z:
And I would move `Document>` (are it's mostly useless) below Taps&Gestures and Navigation, so Tap & Gestures is nearer Screen, as ... it's about screen too :)

Frenzie:
Not just physical keys though. Besides, I don't think any of that stuff should be hiding out anymore. Instead of "back key" we can make it "back action" and we can bind the whole shebang to something like `left, down` by default.

poire-z:
"Disable double tap" appears twice in the Taps & gestures menu, when in File Browser (but not when in Reader).

Frenzie:
Not for me. Custom menu order issue perhaps? :-)

Frenzie:
Oh, never mind. But you need to open a book and then the filemanager for it.

PR #4680, opened 2019-02-28
[UX] Navigation settings for all devices

Frenzie:
Before multiswipes and the gesture manager this was impractical on touch-only devices, but no more!

### DIFF
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -168,11 +168,20 @@ end
 common_settings.navigation = {
     text = _("Navigation"),
 }
+local back_to_exit_str = {
+    prompt = {_("Prompt"), _("prompt")},
+    always = {_("Always"), _("always")},
+    disable ={_("Disable"), _("disable")},
+}
 common_settings.back_to_exit = {
-    text = _("Back to exit KOReader"),
+    text_func = function()
+        local back_to_exit = G_reader_settings:readSetting("back_to_exit") or "prompt"
+        return T(_("Back to exit (%1)"),
+                 back_to_exit_str[back_to_exit][2])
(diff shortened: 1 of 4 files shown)


frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -165,92 +165,91 @@ if Device:isTouchDevice() then
...
+                G_reader_settings:saveSetting("back_to_exit", "disable")
+            end,
+        },
+    },
+}
+common_settings.back_in_filemanager = {
+    text = _("Back in file browser"),
+    sub_item_table = {
+        {
+            text = _("Defer to back to exit setting"),
```
Frenzie: @poire-z Btw, I'm not too happy about this one but I thought "default" was rather unclear.

poire-z: Mhhh, indeed, I don't understand it :) Why the "setting" word?
I don't remember how it works, but why not just "Exit" ?

Frenzie: Because it only exits if the back to exit setting is "exit". ;-)

Frenzie: It could also be prompt (the default) or ignore.

poire-z: Oh, ok. I would have understood it better with a capitalized Back as the other item. But you don't like that :) "As in Back to exit", "Same as Back to exit".
No real idea :|

Frenzie: How about something like this?

## Navigation menu


## Back in file browser submenu

poire-z: Clever :)

Issue #4687, opened 2019-03-01
Even more Gesture Manager action suggestions


pickx:
Now that the Gesture Manager has been massively revamped, here are some more features I personally use often:
- [x] Wikipedia lookup
- [x] Dictionary lookup
- [x] Toggle Reflow

There are more, but the action list is already pretty long at this point. Probably shouldn't bloat it with niche features.

Frenzie:
Ah, reflow! I forgot about that one since fit to content width usually does the trick for me.

Fulltext search is another one that goes along with Wikipedia/Dictionary.

gerhaher:
Fulltext search
Go to page
Go to previous/next chapter
"Go to file manager"

Frenzie:
Dictionary, Wikipedia, and fulltext search would require some refactoring so they use events for showing the window. Not difficult, but not quite a quickie.

Frenzie:
- [ ] Follow nearest link

poire-z:
Might need another to be complete (currently doable by long-press on `Go back to previous location`):
- [ ] Clear location stack

Frenzie:
Would you mind doing it? :-) (I won't tonight, in any case.)

poire-z:
ok, on it.

poire-z:
I don't see why anybody would need this one (but I actually missed it today:):
- [ ] Exit (and why not Restart)

(@Frenzie : I feel like Exit might be tricky, I remember you spent some time on making it work correctly - so, is it? if you feel like adding it :)

Also, we have 3 forward-back-forward. The 4th one is missing: `east west east ` - might be worth adding as the available ones by default as it's some easy to do gesture.

Frenzie:
@poire-z It shouldn't be difficult since now you can just call the self.ui.menu:exitorRestart() method. The problem was in getting it right. ;-) Just sending a Close event around like we used to could lead to unpredictable results.

https://github.com/koreader/koreader/blob/5e01a0c264b3c324ec2f3d34190637890ffaf548/frontend/apps/filemanager/filemanagermenu.lua#L375-L401

https://github.com/koreader/koreader/blob/5e01a0c264b3c324ec2f3d34190637890ffaf548/frontend/apps/reader/modules/readermenu.lua#L248-L280

> Also, we have 3 forward-back-forward. The 4th one is missing: east west east - might be worth adding as the available ones by default as it's some easy to do gesture.

Sure. In the same obvious missing easy gestures you've got all the other diagonal ones.

[...]

PR #4688, opened 2019-03-01
[UX] Gesture manager: add action - toggle reflow

Frenzie:
Cf. https://github.com/koreader/koreader/issues/4687

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -554,6 +554,7 @@ function ReaderGesture:gestureAction(action)
         Device:toggleGSensor()
         self:onGSensorToggle()
     elseif action == "toggle_reflow" then
+        if not self.document.info.has_pages then return end
         if self.document.configurable.text_wrap == 1 then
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -550,6 +553,15 @@ function ReaderGesture:gestureAction(action)
         G_reader_settings:flipNilOrFalse("input_ignore_gsensor")
         Device:toggleGSensor()
         self:onGSensorToggle()
+    elseif action == "toggle_reflow" then
+        if self.document.configurable.text_wrap == 1 then
+            self.document.configurable.text_wrap = 0
+        else
+            self.document.configurable.text_wrap = 1
+        end
```
poire-z: might need to check for not credocument?

Frenzie: True, since otherwise you'll set (and subsequently toggle) a nonsense value and send some no-op events. It's fairly harmless otherwise. :-)

PR #4689, opened 2019-03-01
[chore] Deprecate TapForward/TapBackward

Frenzie:
Also fix previous page action in gesture manager.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -512,11 +512,11 @@ function ReaderGesture:gestureAction(action)
     elseif action == "page_jmp_fwd_10" then
         self:pageUpdate(10)
     elseif action == "page_jmp_fwd_1" then
-        self.ui:handleEvent(Event:new("TapForward"))
+        self.ui:handleEvent(Event:new("GotoViewRel", 1))
     elseif action == "page_jmp_back_10" then
         self:pageUpdate(-10)
     elseif action == "page_jmp_back_1" then
-        self:pageUpdate(-1)
+        self.ui:handleEvent(Event:new("GotoViewRel", -1))
     elseif action == "next_chapter" then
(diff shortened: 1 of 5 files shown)


poire-z:
2 things:
1) if you rename the ids `id = "tap_forward"`, you need to update a few other places that reference them:
```
frontend/apps/reader/modules/readerfooter.lua:            overrides = {
frontend/apps/reader/modules/readerfooter.lua:                'tap_forward', 'tap_backward', 'readerconfigmenu_tap',

frontend/apps/reader/modules/readerconfig.lua:            overrides = { 'tap_forward', 'tap_backward', },

frontend/apps/reader/modules/readerhighlight.lua:            overrides = { 'tap_forward', 'tap_backward', 'readermenu_tap', 'readerconfigmenu_tap', },
```
But I'm not sure we should use rel_forward in these - it's clearer to see the event has been mapped to a **tap**_forward elsewhere.

2) If the handler/event is made more generic by renaming `Tap` to `Rev`, well, it should accept an argument (a positive value, default to 1 - to be translated to -1 by the Backward handler).

So, not sure this is the right thing to do. (Might as well add new onSomething - and don't touch the onTap ones).

Frenzie:
> If the handler/event is made more generic by renaming Tap to Rev, well, it should accept an argument (a positive value, default to 1 - to be translated to -1 by the Backward handler).

Those already exist. You have `PagingRel` and `GotoViewRel`. I'm not sure why the names differ.

These TapForward/Backward events seem to be intended as a (perhaps unnecessary) convenience for the most common use case.

So in that case, those particular event handlers should just be stripped out and replaced by the renamed generic handlers.

Frenzie:
@poire-z How about this? It deprecates those two redundant events and renames PagingRel to GotoViewRel for consistency and calling simplicity.

poire-z:
Yep, looks fine to me.

PR #4691, opened 2019-03-01
[UX] creoptions: add more margin values

### LINKED ISSUE #4684: Tiny margins on 10,3" screen
No matter what dpi margin or style override I use, the margins are less than 0,5cm.

In a eink 10,3 screen looks tiny. 

Is there a way to setup, by default, bigger margins? Or what piece of code i could touch to increase the margins? Looks like the margins are absolute instead of relative but I don't know.. 

Maybe it would be good to have custom margins by percentage or pixels.(or more buttons to choose).

Frenzie:
Fixes #4684.

screenshot_2019-03-01_12-55-22


NiLuJe:
:+1:

poire-z:
Looks fine.
Glad you didn't change the 3 first values (so, no shift in settings - no book re-rendering or visual changes for users) and only added new values. 

(It will just be a bit strange for people who did customized the 3 first ones in default.persistent.lua - probably they increased them, possibly beyond the 4th and 5th).

poire-z:
Nothing to do with this PR, just brainstorming (even for a possible personal hack):
On my GloHD, a refresh on a previously solid black area makes the underlying test grayer/thiner instead of black.
So, with these black progress bars like contrast and this new one, I usually follow quiting the bottom menu with a small diagonal swipe to full refresh.
I don't think it's worth doing a full refresh when quiting a bottom config dialog that would include these black bars - might not be needed on other devices.
Just wondering if there's a visual idea for such progress bar without the big solid blacks - while still as readable as to what is selected. I can't think of any :|

NiLuJe:
Hint hint; that's what flash_ui is for ;p.

NiLuJe:
Although I vaguely recall having removed the flashing from the bottom menu
to avoid bad interactions with the top menu ;).

Stuff is generally far better on newer devices though ;).

(reagl devices will flash less w/ flash_ui, while also suffering much less
from ghosting).

Frenzie:
I don't think it'd suffer in usability from using the same gray as the toggle switches, if that's any better for ghosting.

PS I'll fix the unit tests later.

poire-z:
Well, solid gray areas auto-flash-ui themselves on my GloHD :) and I really don't like that either (I've converted some of our gray lines to black in my own patch - separators in the menu were flashing and that did bother me).
But ok, I don't mind that much the gray toggles, so I'll try that progress bar dressed in gray.

Frenzie:
> solid gray areas auto-flash-ui themselves

I might be overlooking something (especially since I haven't looked at the code and am therefore unaware of its comments) but I'd say either so should black or neither should gray, depending on which behavior is best?

poire-z:
No worry with our code. It's the (GloHD at least) hardware that does this auto-flash with grey (it's not a single transition: it goes from white to black, and they from black to grey - or something like that - no matter what we use in our code).

frontend/ui/data/creoptions.lua
```diff
@@ -89,20 +89,30 @@ Note that your selected font size is not affected by this setting.]]),
...
                     DCREREADER_CONFIG_MARGIN_SIZES_SMALL,
                     DCREREADER_CONFIG_MARGIN_SIZES_MEDIUM,
                     DCREREADER_CONFIG_MARGIN_SIZES_LARGE,
+                    DCREREADER_CONFIG_MARGIN_SIZES_X_LARGE,
+                    DCREREADER_CONFIG_MARGIN_SIZES_XX_LARGE,
+                    DCREREADER_CONFIG_MARGIN_SIZES_XXX_LARGE,
+                    DCREREADER_CONFIG_MARGIN_SIZES_HUGE,
+                    DCREREADER_CONFIG_MARGIN_SIZES_X_HUGE,
                 },
-                name_text_hold_callback = optionsutil.showValuesMargins,
```
Frenzie: Whoops, this deletion is a mistake. :-)

Frenzie:
Made it slightly prettier:

screenshot_2019-03-01_15-07-30

poire-z:
> I don't think it'd suffer in usability from using the same gray as the toggle switches, if that's any better for ghosting without a flash.

It's actually far better in grey on my GloHD (because of the HW auto-flash, which leaves no ghosting at all):

image

image

A bit too monotonous though :)

In case one wantd to try it on some other devices:
```diff
--- a/frontend/ui/widget/buttonprogresswidget.lua
+++ b/frontend/ui/widget/buttonprogresswidget.lua
@@ -43,13 +43,15 @@ function ButtonProgressWidget:update()
     local button_margin = Size.margin.tiny
     local button_padding = Size.padding.button
     local button_bordersize = Size.border.button
-    local preselect
+    local preselect, background
     local button_width = math.floor(self.width / self.num_buttons) - 2*button_padding - 2*button_margin - 2*button_bordersize
     for i = 1, self.num_buttons do
         if self.position >= i then
             preselect = true
+            background = Blitbuffer.COLOR_GREY
         else
             preselect = false
+            background = Blitbuffer.COLOR_WHITE
         end
         local button = Button:new{
             text = "",
@@ -60,6 +62,7 @@ function ButtonProgressWidget:update()
             enabled = true,
             width = button_width,
             preselect = preselect,
+            background = background,
             text_font_face = self.font_face,
             text_font_size = self.font_size,
             callback = function()
```

> Although I vaguely recall having removed the flashing from the bottom menu to avoid bad interactions with the top menu ;).

Indeed, no flash on the bottom menu.

poire-z:
Holding on margin title (to have help text) now crashes with:
```
./luajit: frontend/ui/data/optionsutil.lua:88: attempt to get length of field 'toggle' (a nil value)
stack traceback:
        frontend/ui/data/optionsutil.lua:88: in function 'name_text_hold_callback'
        frontend/ui/widget/configdialog.lua:277: in function 'hold_callback'
        frontend/ui/widget/button.lua:241: in function 'handleEvent'
        frontend/ui/widget/container/inputcontainer.lua:255: in function 'handleEvent'
```
(No crash when holding on the Contrast title)

Issue #4333, opened 2018-11-11
FR: more Gesture Manager actions


pickx:
The Gesture Manager is a great feature for making KOReader faster to use.

That said, here are features I use frequently and would love as options when a gesture is triggered:

- [x] Toggle landscape <-> portrait
- [ ] Open last document
- [x] Sleep device
- [x] History (of last opened files)

thanks!

Biep:
In reference to the issue Frenzie just referenced:

Currently I am really wanting a "Return to where I jumped from" gesture (as opposed to a menu item) that is different from the "go to the previous page" one.  Often I jump, then move several times to the next page, and then want to jump back - or sometimes want to leaf back.  A single gesture cannot serve both functions.

More generally, I'd love (as a user) having the ability to assign gestures to specific commands - the way hotkeys can be assigned by the user in many programs.

Frenzie:
The gesture manager is the early stage of exactly that btw. It's still developing, but before @robert00s added it there was nothing. It's slow acquiring more configurable gestures (e.g., short diagonal swipe) and actions.

Frenzie:
A few quick notes while my attention is focused on this issue. (I was a bit busy in November.)

tl;dr I'll implement at least some of these after dinner.

## Toggle landscape <-> portrait
There's already a rotation gesture for that on devices that support two-finger gestures. I haven't looked into where it lives but it should be simple to reuse or otherwise jig up with `Device:setRotation()`.

## Open last document
Probably not too hard but currently implemented locally in the menu.

https://github.com/koreader/koreader/blob/6b3158e497228614591dfd654d6318ffffa27156/frontend/apps/reader/modules/readermenu.lua#L194-L205

https://github.com/koreader/koreader/blob/6b3158e497228614591dfd654d6318ffffa27156/frontend/apps/reader/modules/readermenu.lua#L219

## Sleep device
Huh, no physical buttons? Anyway, that's just `UIManager:suspend()`, couldn't be simpler.

## History (of last opened files)
Sending a `ShowHist` event might do the trick. I'll have to check.

Frenzie:
Related would be rotate clockwise/counterclockwise as well as specific states, but then comes the point where we need gesture categories akin to @poire-z 's style tweaks.

poire-z:
And already came the point we don't need that much :) and some devices don't support multi touch gestures - and multi swipes are already super enough and don't conflict with anything :)

Biep:
[...]

PR #4692, opened 2019-03-01
[UX] Gesture manager: add actions - prev/next chapter

poire-z:
Also adds rudimentary support for these in ReaderPaging (might need more work in some zoom modes or in scroll mode).
paper.pdf includes a workable TOC.
cf https://github.com/koreader/koreader/issues/4333#issuecomment-468420820


Frenzie:
@poire-z I don't really know why CircleCI takes a while to show up occasionally. Anyway, the readerrolling_spec file needs to be updated. I'll take care of it.

poire-z:
Thanks :)

PR #1998, opened 2016-04-30
Multi-line toggle switch & more pdf cropping methods

Hzj-jie:
This issue has been discussed #1923, we have bunch of build-in zooming method, but use none of them for pdf files. So this change adds three more pdf cropping options as 'none' -> 'content', 'width' -> 'content-width', 'height' -> 'content-height'.

PR #4696, opened 2019-03-01
[UX] Add cropping: none mapped to page width zoom

### LINKED ISSUE #1923: Suggestion: "None" as a cropping option
Can we have "None" as a cropping option? 

It was confusing when I opened a pdf for the first time, noticed that cropping was shown in UI as "auto", but the pdf wasn't cropped at all. Sure, tapping on "auto" again activates cropping. But what if I want to disable cropping and don't want to read in scroll mode?

Thanks.

Frenzie:
Fixes #1923.


Frenzie:
@poire-z @NiLuJe I overlooked #1998 which has roughly the same thing with a few more options. Still, it seems that a "none" option should be present in this list.

poire-z:
I'm not a heavy readerpaging user, so I'm not even sure how these crop options relate to the zoom mode options (eg: if i choose Zoom to content width, do these crop do more stuff to that then? - in which case 'none' surely need to be there).

In case they are independant, UI wise, as the bottom config is small with only one item, the zoom mode would be more usable from there (the zoom mode top menu hide quick a bit of the page, so you need to close it to see the effect, and go back to try another.

Frenzie:
Auto is the same as content width, but the other crop options are different.

Semi-auto applies content width, but lets you play around with it:

screenshot_2019-03-01_20-31-20

And manual initially starts from no crop.

But actually this isn't quite a proper implementation of none. If you do a manual crop like this:
screenshot_2019-03-01_20-33-29

Then the fact that the other modes are secretly content width and page width makes 'em both the same:
screenshot_2019-03-01_20-34-14

So the logic of the whole thing is slightly off… anyway, probably not something I want to play around with prior to the stable release coming up (if at all).

> I'm not a heavy readerpaging user, so I'm not even sure how these crop options relate to the zoom mode options

I am, but content width and occasionally page width or page are practically all I use. I don't use reflow or this cropping business.

PR #4698, opened 2019-03-01
[fix] GestureDetector: only initiate multiswipe with clear N/S/E/W direction

Frenzie:
Accidentally lost while refactoring. See <https://github.com/koreader/koreader/pull/4640#issuecomment-468788986>.

### DIFF
--- a/frontend/device/gesturedetector.lua
+++ b/frontend/device/gesturedetector.lua
@@ -526,8 +526,14 @@ function GestureDetector:handlePan(tev)
         local msd_direction_prev = (msd_cnt > 0) and self.multiswipe_directions[msd_cnt][1] or ""
         local prev_ms_ev, fake_first_tev
 
+        if msd_cnt == 0 then
+            -- do not initiate multiswipe unless we have a clear north/south/east/west direction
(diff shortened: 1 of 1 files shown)


Frenzie:
This explains why I didn't really understand what you were saying previously. Before I introduced multiswipes, the current `west, north` (etc.) gestures would've all been (mis)interpreted as diagonal even though they aren't by any stretch of the imagination.

Just try it in v2019.02. Make a nice big half rectangle and you'll end up with a screenshot. ;-) Heck, you can even make it a big triangular shape.

screenshot_2019-03-01_21-49-04

So my initial code was functioning very well (give or take some stuff like that pan delay) but also a hugely complex spaghetti mess. And while refactoring it into something palatable I stupidly managed to lose properly working distance as well as this only start on a proper north/south/east/west restriction. But in my mind they were both still there. :-)

Frenzie:
Btw, it just occurred to me that there might be a way to create actually usable diagonal multiswipes using a similar restriction principle. I.e., don't mix and match, but either all straight or all diagonal. Not that this is something I want to investigate prior to release but it could be an idea for the future. ;-)

Issue #4667, opened 2019-02-26
Panning rate default


Frenzie:
* KOReader version: 2019.2
* Device: Android/AppImage

#### Issue
Smaller panning rate seems to be enabled by default. Instead, it should depend on Device:hasEinkScreen(). Cf. #4470.

https://github.com/koreader/koreader/blob/5495c5d108eae003cf9d36426d59448b79758d62/frontend/device/generic/device.lua#L109-L110

@pazos Also, if hasEinkScreen then hasColorScreen should be no for some memory/performance improvement.

pazos:
> if hasEinkScreen then hasColorScreen should be no for some memory/performance improvement.

For android & sdl targets sure. There is the pocketbook colourLux which is supported and has both EinkScreen and ColorScreen.

Frenzie:
That's what device-specific overrides are for, but it's just a "btw" in any case.

pazos:
Cool. For me this makes sense:

emulator has a switch to toggle panning rate.
on eink devices set low_pan_rate (on the future make the rate GUI editable)
on other devices do not provide a switch to toggle panning rate and do not set low_pan_rate.

Frenzie:
I'm not completely sure about that not providing a switch part. Low pan rate is primarily about screen drawing performance, but there's also blitting performance to take into consideration. (Of course _2_ is a bit extremely low.)

mergen3107:
Guys, can you please explain what does `Use smaller panning rate` mean in general, and specifically for eInk screens?

mergen3107:
I found the answer on [MR](https://www.mobileread.com/forums/showthread.php?t=320937) :D 
Credits to @pazos:


> It sets a lower "refreshes per second" when scrolling (2 vs 30). It is enabled by default on eink devices.
> 
> Most people will want to leave it enabled. Some android users with modern Onyx/Likebook devices and A2 enabled can disable it without pain.

PR #4699, opened 2019-03-01
[UX] Gesture manager: add action - show dictionary/Wikipedia

Frenzie:
References #4687.

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -5,6 +5,7 @@ local DictQuickLookup = require("ui/widget/dictquicklookup")
 local Geom = require("ui/geometry")
 local InfoMessage = require("ui/widget/infomessage")
 local InputContainer = require("ui/widget/container/inputcontainer")
+local InputDialog = require("ui/widget/inputdialog")
 local JSON = require("json")
 local KeyValuePage = require("ui/widget/keyvaluepage")
 local LuaData = require("luadata")
@@ -173,14 +174,9 @@ end
 function ReaderDictionary:addToMainMenu(menu_items)
     menu_items.dictionary_lookup = {
         text = _("Dictionary lookup"),
-        tap_input = {
-            title = _("Enter a word to look up"),
-            ok_text = _("Search dictionary"),
-            type = "text",
-            callback = function(input)
-                self:onLookupWord(input)
-            end,
-        },
+        callback = function()
+            self:onShowDictionaryLookup()
+        end,
     }
     menu_items.dictionary_lookup_history = {
         text = _("Dictionary lookup history"),
@@ -554,6 +550,35 @@ function ReaderDictionary:dismissLookupInfo()
         -- UIManager:forceRePaint()
     end
     self.lookup_progress_msg = nil
+end
+
+function ReaderDictionary:onShowDictionaryLookup()
+    self.dictionary_lookup_dialog = InputDialog:new{
+        title = _("Enter a word to look up"),
+        input = "",
+        input_type = "text",
+        buttons = {
+            {
+                {
+                    text = _("Cancel"),
+                    callback = function()
+                        UIManager:close(self.dictionary_lookup_dialog)
+                    end,
+                },
+                {
+                    text = _("Search dictionary"),
+                    is_enter_default = true,
+                    callback = function()
+                        UIManager:close(self.dictionary_lookup_dialog)
+                        self:onLookupWord(self.dictionary_lookup_dialog:getInputText())
+                    end,
+                },
+            }
+        },
+    }
+    UIManager:show(self.dictionary_lookup_dialog)
+    self.dictionary_lookup_dialog:onShowKeyboard()
+    return true
 end
 
 function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -39,6 +39,9 @@ local action_strings = {
     open_previous_document = _("Open previous document"),
     filemanager = _("File browser"),
 
+    dictionary_lookup = _("Dictionary lookup"),
+    wikipedia_lookup = _("Wikipedia lookup"),
+
     full_refresh = _("Full screen refresh"),
     night_mode = _("Night mode"),
     suspend = _("Suspend"),
@@ -270,6 +273,9 @@ function ReaderGesture:buildMenu(ges, default)
         {"history", true},
         {"open_previous_document", true, true},
         {"filemanager", not self.is_docless, true},
+
+        {"dictionary_lookup", true},
+        {"wikipedia_lookup", true, true},
 
         {"full_refresh", true},
         {"night_mode", true},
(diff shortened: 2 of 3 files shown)


poire-z:
Looks OK, but I tested it (the lookups, not the gestures) and strangely, the keyboard is a lot less responsive with this patch than without...
I can't really see why, so may be it's something with my emulator (but patch -p1 and patch -R -p1 repeatedly show the same differences...)

Frenzie:
You mean the onscreen keyboard?

poire-z:
Yes, our on-screen keyboard.

Frenzie:
I'll give it a try on my H2O this weekend. I don't think I notice anything. Is there a particular way to demonstrate the problem?

poire-z:
Put the mouse cursor on a virtual keyboard key, and hit your mouse left click like mad
And ... **have "Disable double tap" unchecked** :) (I had unchecked earlier today to check some of your stuff I guess).
So, no problem when double tap is disabled (10 mad clicks = 10 letters). But there's a difference when it is not disabled (10 mad clicks = 1 or 2 letters)

Frenzie:
Ah, so then it's probably not a performance thing (which would be bothersome) but simply a case of  (un)intended double tap delay. I still can't reproduce it though, neither in file manager nor reader.

poire-z:
Most probably a property that can be set to a widget to have double tap disabled for this widget (even if it's not disabled globally) - that went missing (or was not set anymore) after your small refactoring.

```
frontend/apps/reader/modules/readerpaging.lua:    Input.disable_double_tap = false
frontend/apps/reader/modules/readerpaging.lua:    Input.disable_double_tap = true
frontend/ui/uimanager.lua:    if widget.disable_double_tap == false then
frontend/ui/uimanager.lua:        Input.disable_double_tap = false
frontend/ui/uimanager.lua:        Input.disable_double_tap = true
frontend/ui/uimanager.lua:    Input.disable_double_tap = true
frontend/ui/uimanager.lua:            if self._window_stack[i].widget.disable_double_tap == false then
frontend/ui/uimanager.lua:                Input.disable_double_tap = false
frontend/ui/widget/virtualkeyboard.lua:    disable_double_tap = true,
frontend/device/input.lua:    disable_double_tap = true,
```

Frenzie:
So then is it in only of dictionary or Wikipedia or in both?

poire-z:
With "Disable double tap" not disabled:

Without this PR / with this PR

**Dictionnary lookup: OK / SLOW**
Wikipedia lookup: **SLOW / SLOW**
Fullteaxt search: OK / OK
Reading statistics > Settings : OK / OK
Text editor: OK / OK
Rename bookmark: **SLOW / SLOW**

So, it's a more general issue when double tap is not disabled, that is taken care of in various places with keyboard input, but not all - and you broke only Dictionary lookup with this PR (so, once unbroken, that trick should be put into these other places).

Frenzie:
I don't really understand why I'm failing to spot the problem but that makes sense. The dictionary lookup is something I actually changed (concretely, by copying it from Wikipedia) so it's a rewrite that (seemingly) functions the same.

Wikipedia was just a tiny reshuffling of the exact same code.

What's not clear to me is why this should be an issue, because the VirtualKeyboard uses `disable_double_tap = true,`.

As a quick workaround you could see if adding `disable_double_tap = true,` to the InputDialog widget resolves the problem.

Frenzie:
@poire-z Instead of using InputDialog directly we should be able to use the InputContainer's onInput event (just like in the menu, which apparently worked well with double tap). It'd make for slightly less code but I don't really see how it should make a difference in actual functionality. :-/

https://github.com/koreader/koreader/blob/779909343c80501fc223376f4b9e0d46f183c9fd/frontend/ui/widget/container/inputcontainer.lua#L263-L291

poire-z:
> As a quick workaround you could see if adding disable_double_tap = true, to the InputDialog widget resolves the problem.

Nope, this does not solve it:
```diff
local InputDialog = InputContainer:new{
+    disable_double_tap = true,
    is_always_active = true,
```

Frenzie:
Hm, curious. Anyway, shall we merge this as is, seeing how the problem is unrelated, or strip out the dictionary part (since Wikipedia is already borked)?

poire-z:
Go ahead. We'll have that into master to fix them all when we find the culprit.

poire-z:
Might be dependant of readerrollin being used.
There's this on widget show:
https://github.com/koreader/koreader/blob/57ce8dc5742a024cd334e9be63d6a7531ab92120/frontend/ui/uimanager.lua#L292-L297
and this on widget close:
https://github.com/koreader/koreader/blob/57ce8dc5742a024cd334e9be63d6a7531ab92120/frontend/ui/uimanager.lua#L321-L340

So, any widget on the stack that didn't disable double tap will enable it, and all the others that did disable it are screwed (#2155).
So, readerui sets disable_double_tap = false (for readerrolling double tap to go to prev/next chapter), so I guess any input/virtualkeyboard that is stacked over it don't get double tap disabled as they requested.
(Why no issue when Input invoked from menu, no idea :)

poire-z:
This seems to solve the issue (but as it's uimanager wild territory, I won't go there alone :) so pinging @NiLuJe - or we might want to wait after release to fix that):

```diff
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -318,8 +318,9 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
     widget:handleEvent(Event:new("FlushSettings"))
     -- first send close event to widget
     widget:handleEvent(Event:new("CloseWidget"))
-    -- make it disabled by default and check any widget that enables it
+    -- make it disabled by default and check if any widget wants it disabled or enabled
     Input.disable_double_tap = true
+    local requested_disable_double_tap = nil
     -- then remove all references to that widget on stack and refresh
     for i = #self._window_stack, 1, -1 do
         if self._window_stack[i].widget == widget then
@@ -333,11 +334,15 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
                 logger.dbg("Lower widget", self._window_stack[i].widget.name or self._window_stack[i].widget.id or tostring(self._window_stack[i].widget), "was dithered, honoring the dithering hint")
             end

-            if self._window_stack[i].widget.disable_double_tap == false then
-                Input.disable_double_tap = false
+            -- Set double tap to how the topest that specified how it wants it want it
+            if requested_disable_double_tap == nil and self._window_stack[i].widget.disable_double_tap ~= nil then
+                requested_disable_double_tap = self._window_stack[i].widget.disable_double_tap
             end
         end
     end
+    if requested_disable_double_tap ~= nil then
+        Input.disable_double_tap = requested_disable_double_tap
+    end
     if dirty and not widget.invisible then
         -- schedule remaining widgets to be painted
         for i = 1, #self._window_stack do
```

Frenzie:
I actually solved a similar issue with the screensaver a couple of years ago (except in that case properly resolving which widget was the topmost for drawing purposes). We'll need at least a few more days prior to release so this might just be on the safe enough side of risky.

Frenzie:
Besides some more finishing touches on the multiswipes (mainly defaults & some actions) #4708 is a prerequisite. I'd also like to have #4667 finished to ensure an optimal F-Droid first impression.

poire-z:
Another thing would be to disable double tap by default, or altogether.
It seems only used by readerolling to double tap in the 1/4 top left & right corners to go to prev/next chapters.
Otherwise, it will just make tap'ing feels slow to new users (that have not yet disabled it).

NiLuJe:
Yeah, that seems a more sensible default if it's barely used anywhere.

Frenzie:
But double tap *is* disabled by default, isn't it?

poire-z:
Oh, looks like it is :) sorry. And good for me to not have to look at that!

PR #4700, opened 2019-03-01
ButtonProgressWidget: grey buttons instead of black

poire-z:
May help with ghosting when closing bottom dialog, and fit better with the other grey toggles.
See https://github.com/koreader/koreader/pull/4691#issuecomment-468646394 and follow ups.

(To be merged or tried before/after release.)

It's actually far better in grey on my GloHD (because of the HW auto-flash, which leaves no ghosting at all):

<kbd>image</kbd>

<kbd>image</kbd>

### DIFF
--- a/frontend/ui/widget/buttonprogresswidget.lua
+++ b/frontend/ui/widget/buttonprogresswidget.lua
@@ -43,13 +43,15 @@ function ButtonProgressWidget:update()
     local button_margin = Size.margin.tiny
     local button_padding = Size.padding.button
     local button_bordersize = Size.border.button
-    local preselect
+    local preselect, background
     local button_width = math.floor(self.width / self.num_buttons) - 2*button_padding - 2*button_margin - 2*button_bordersize
     for i = 1, self.num_buttons do
         if self.position >= i then
             preselect = true
+            background = Blitbuffer.COLOR_GREY
         else
             preselect = false
+            background = Blitbuffer.COLOR_WHITE
         end
         local button = Button:new{
             text = "",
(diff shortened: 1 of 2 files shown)


Frenzie:
Isn't there normally supposed to be some white padding in between?

(I have to say I agree with you that it seems more… blah.)

poire-z:
Indeed.
With last commit it would be:

<kbd>image</kbd>

With one line commented out:

<kbd>image</kbd>

It might be a little busy with the grey when one have "Flash buttons and menu items", so @NiLuJe may put his veto (and I'll keep that to me :)

NiLuJe:
I don't have anything against it specifically, I'll check how it looks
tomorrow, but that seems benign enough ;).

poire-z:
Tweaked a bit to have thin grey borders, like they are on toggle switches.

New thin grey style:

<kbd>image</kbd>

Current default style:
<kbd>image</kbd>

Frenzie:
The thin gray looks nicer. :-)

poire-z:
I'd like to add some light visual indication of which button is the default (the user default, or ours if none set).
Adding a unscaled 1px to that button border size would give:

<kbd>image</kbd>

<kbd>image</kbd>

Looks quite better on the black style than on the grey style. I'll have to check how it looks on my device.
Any other idea (keeping the widget border/invert simple - avoiding imbricating framecontainer...)

Frenzie:
> Adding a unscaled 1px to that button border size would give:

Needs to be scaled of course, just use 0.5. ;-)

I'd say the border should be a different color (i.e., a darker gray or black).

poire-z:
> I'd say the border should be a different color (i.e., a darker gray or black).

Obviously :) Didn't think of that...
I'll go with the above (scaled) extra border for the now-disabled black style.
And i'll just use COLOR_BLACK for the default item with the thin grey style. It's quite discreet, may be too discreet, but I wanted it that way initially :)

<kbd>image</kbd>

PR #4704, opened 2019-03-02
[UX] GestureDetector: add diagonal multiswipes

Frenzie:
Multiswipes consisting of mixed straight and diagonal strokes are not dependable and too easy to mess up, but making them mutually exclusive seems to work out quite well.


poire-z:
Aren't we done yet with multiswipes? :)

Looks fine, and tested on my Kobo that diagonal swipe to refresh still works well.

NiLuJe:
Yay ;)

Issue #2040, opened 2016-05-23
Publish Koreader for Android


Hzj-jie:
- KOReader version:
- Device:
#### Issue

Is there any reason we do not publish Koreader in Google play store? If it's simply because nobody has a Google developer account, I could help.

Frenzie:
Afaik mostly because it's alpha quality.

houqp:
Yep, I don't think we are ready for android play yet. There are two major blockers:

* we have many bug repots tagged with android that need to be resolved
* a properly home screen which @chrox is working on

And I think it's better to start with f-droid first, get some feedback, then move to the play store. Otherwise, we might ended up getting many bad reviews :P

Frenzie:
I prefer F-Droid anyway.

houqp:
Even for publishing in f-droid, we still need to clear the first blocker :)

gerroon:
It would be nice to also have more frequent Android builds, either through F-droid or github releases

nico202:
Any news? Especially for the f-droid build

Frenzie:
Unfortunately nothing's really changed. No one has enough time to work on the Android port and it still doesn't work on many devices.

nico202:
Ok thanks

dimqua:
https://gitlab.com/fdroid/rfp/issues/205

gerroon:
I think that koreader is ready for the initial F-droid release. I have been testing it on multiple devices (lollipop, nougat, kitkat) and so far so good.

Frenzie:
@dimqua Anything specific missing?

taosxx:
Is there actually anything in KOReader that doesn't comply with [F-Droids inclusion criteria](https://f-droid.org/docs/Inclusion_Policy/)?

Frenzie:
Not that I know of. Some of the artwork is http://creativecommons.org/licenses/by/3.0 but that shouldn't be an issue. If there are any concrete actionable points I'll try to address them, of course, but I don't want to directly take on the time investment or responsibility right now.

IzzySoft:
As far as I see, there are [some building issues](https://gitlab.com/fdroid/rfp/issues/205#note_76481497). Did you meanwhile update to a newer SDK level? Recent ones no longer support the `android` command.

Frenzie:
[...]

PR #4708, opened 2019-03-02
Disable ota updates on certain android flavors

pazos:
Requires https://github.com/koreader/android-luajit-launcher/pull/126
Related to #2040

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -10,6 +10,11 @@ local T = require("ffi/util").template
 local function yes() return true end
 local function no() return false end
 
+local function canUpdateApk()
+    -- disable updates on fdroid builds, since they manage their own repo.
+    return (android.getFlavor() ~= "fdroid")
+end
+
 local Device = Generic:new{
     model = android.getProduct(),
     hasKeys = yes,
@@ -21,7 +26,7 @@ local Device = Generic:new{
     display_dpi = android.lib.AConfiguration_getDensity(android.app.config),
     hasClipboard = yes,
     hasColorScreen = yes,
-    hasOTAUpdates = yes,
+    hasOTAUpdates = canUpdateApk,
 }
 
 function Device:init()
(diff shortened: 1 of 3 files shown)


frontend/device/android/device.lua
```diff
@@ -10,6 +10,13 @@ local T = require("ffi/util").template
 local function yes() return true end
 local function no() return false end
 
+local function canUpdateApk()
+    local flavor = android.getFlavor()
+    -- disable updates on fdroid builds, since they manage their own repo.
+    -- disable updates on debug builds too.
```
Frenzie: Is that necessary? It seems like it just complicates testing without a clear gain.

pazos: disable on debug builds? not necessary

Frenzie: Well, I wouldn't then. :-)

pazos: ok, but are you fine with the flavor of debug builds being "debug" or shall I change that too (and make the flavor settable, like on release builds)?

Frenzie: Oh right, that should definitely be settable.

Frenzie: Kinda defeats the point of a debug build otherwise, imho.

pazos:
Done!

BTW, this PR breaks the remote option of reproducible builds for android (ie: resources won't match between two or more different flavors). I think this is fine as both short and long term solution.

frontend/device/android/device.lua
```diff
@@ -21,7 +26,7 @@ local Device = Generic:new{
     display_dpi = android.lib.AConfiguration_getDensity(android.app.config),
     hasClipboard = yes,
     hasColorScreen = yes,
-    hasOTAUpdates = yes,
+    hasOTAUpdates = function() return canUpdateApk() end,
```
houqp: wouldn't it be simpler to do the following?

```lua
hasOTAUpdates = canUpdateApk
```

Frenzie:
> BTW, this PR breaks the remote option of reproducible builds for android (ie: resources won't match between two or more different flavors). I think this is fine as both short and long term solution.

That would mean there's no logical possibility for reproducible builds of any kind of "flavors" at all, on any platform. All that matters is randomness introduced during the build process, such as filesystem order and timestamps. This PR indirectly does that, but not in a way that is unique for the need to adjust the mtimes of generated files. Get rid of the randomness and what you're left with is reproducible, provided you use the same not at all random set of control switches.

PR #4711, opened 2019-03-02
Fix slow keyboard when double tap not disabled

poire-z:
When double tap is not disabled (only ensured in ReaderRolling), repeated key hits were slowed down by it. This ensures Input widgets that temporarily overrides it to be disabled are satisfied.
See https://github.com/koreader/koreader/pull/4699#issuecomment-468910319

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -318,8 +318,9 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
     widget:handleEvent(Event:new("FlushSettings"))
     -- first send close event to widget
     widget:handleEvent(Event:new("CloseWidget"))
-    -- make it disabled by default and check any widget that enables it
+    -- make it disabled by default and check if any widget wants it disabled or enabled
     Input.disable_double_tap = true
+    local requested_disable_double_tap = nil
     -- then remove all references to that widget on stack and refresh
     for i = #self._window_stack, 1, -1 do
         if self._window_stack[i].widget == widget then
@@ -333,10 +334,14 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
                 logger.dbg("Lower widget", self._window_stack[i].widget.name or self._window_stack[i].widget.id or tostring(self._window_stack[i].widget), "was dithered, honoring the dithering hint")
             end
 
-            if self._window_stack[i].widget.disable_double_tap == false then
-                Input.disable_double_tap = false
+            -- Set double tap to how the topest widget that specifies how it wants it want it
+            if requested_disable_double_tap == nil and self._window_stack[i].widget.disable_double_tap ~= nil then
+                requested_disable_double_tap = self._window_stack[i].widget.disable_double_tap
             end
         end
+    end
+    if requested_disable_double_tap ~= nil then
+        Input.disable_double_tap = requested_disable_double_tap
     end
     if dirty and not widget.invisible then
         -- schedule remaining widgets to be painted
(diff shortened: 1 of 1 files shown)


ptrm:
I would like to ask whether this issue had been present for a longer time, or just since the recent multiswipe change? I mean, on Kobos it's hard to type more than 2-3 characters per second (by a rule of thumb), even with today's nightly containing this fix, and I'm wondering if it is something the hardware/software design is still enforcing (nickel's keyboard is faster, though)

poire-z:
> 2-3 characters per second

That's already fast :)
Well, it must have been present for a long time, but only when double tap is not disabled.
May be you find it slow because of the keyboard keys flashing? You can disable that with `Screen> Eink settings> Flash keyboard` - which I use, and it's quite fast: I can type many keys before they are shown in the text box - and I don't seem to lose any.

Frenzie:
This issue has always been present. And I do mean always quite literally. The situation up to yesterday was introduced in 2013 (225ae9f826d91fb59019d34c3276534be7e8130c), before which double tap delay always got in the way of tapping. Since 2013 only in certain circumstances so it went unnoticed. But since 2016 double tap has been disabled by default (3d7fe1002ca2c56d0281d0ea4fd7103931a61768).

I disable keyboard flashing, ftr.

Frenzie:
Btw, I don't disable it for performance reasons, just aesthetics.

[2019-03-03 11.41.01.mp4.zip](https://github.com/koreader/koreader/files/2923011/2019-03-03.11.41.01.mp4.zip)

poire-z:
What is the font you are using in this video?

ptrm:
I do have double tap enabled to navigate between chapters, so I did not notice the connection with slower keyboard since it has always been on :)

@Frenzie, aesthetics aside, should keyboard flashing speed up the typing speed? I tried, and still faster taps are ignored, as if there was a throttle around 300ms long blocking the intermediary events

Frenzie:
@poire-z Just standard Noto Serif?

@ptrm It shouldn't speed up typing speed perceptibly, but logically speaking less processing (in this case of where to paint updates) does of course mean faster. In a theoretical sense, not a practical one. Key *flashes* lag behind, registering taps does not.

I have an H2O.

poire-z:
> Just standard Noto Serif?

Oh, ok, didn't realize it was that nice... It has a bit more contrast on your video that whan can I make it.

> I do have double tap enabled to navigate between chapters

Well, that double tap enabled makes every tap (turning page, opening menu...) slower.
Now that we have multiswipes, and if you can lose that habit of double taping top corners, you would gain quite something switching to gestures for next/prev chapters :)
But ok: if you disable double tap, just for testing, you still have that speed issue with the keyboard?

Frenzie:
@poire-z But shouldn't this fix make it so that the keyboard is responsive even with double tap enabled?

poire-z:
Yes. But what it this fix doesn't fix things enough ? :) So my question about how with it disabled.
(I just tested it on the emulator, and what I had noticed seemed fixed.)

Frenzie:
@ptrm Could you describe the specifics steps you use to get to a keyboard that is responding slowly? It could be as simple as a dialog that accidentally opens the keyboard before the dialog or something.

ptrm:
> Well, that double tap enabled makes every tap (turning page, opening menu...) slower.

Ah, slower *despite* the fix :) I just never tried to use the keyboard with double tap disabled, and since multiswipe gestures are here I will give it a try, because it appears faster indeed.

Also, I'm still not sure if I'm not tapping to lightly when typing faster, it's not a mobile phone screen after all. I will focus more on this and if it is actually an issue with KOreader will post a separate issue. Thanks.

PR #4713, opened 2019-03-02
[fix, UX] Common menu & info menu: switch to dofile

### LINKED ISSUE #4703: "Disable double tap" appears twice in the Taps & gestures menu, when in File Browser (but not when in Reader).
> "Disable double tap" appears twice in the Taps & gestures menu, when in File Browser (but not when in Reader).

_Originally posted by @poire-z in https://github.com/koreader/koreader/pull/4677#issuecomment-468905516_

To reproduce, open a book and then return to the file manager.

Frenzie:
Require is kept in memory, including the modifications made to it by MenuSorter. This can cause trouble when switching between the FileManager and Reader.

Fixes #4703.

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -158,7 +158,7 @@ function FileManagerMenu:setUpdateItemTable()
         }
     end
     -- insert common settings
-    for id, common_setting in pairs(require("ui/elements/common_settings_menu_table")) do
+    for id, common_setting in pairs(dofile("frontend/ui/elements/common_settings_menu_table.lua")) do
         self.menu_items[id] = common_setting
     end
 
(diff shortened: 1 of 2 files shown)


poire-z:
It solves the 2 cases mentionned in https://github.com/koreader/koreader/issues/4703#issuecomment-468927558. So I guess it's fine :)

(There were only these 2x2 ones? - no other require anywhere else? or you just need the initial ones to use dofile?)

Frenzie:
There are plenty of other requires, but I don't think those are potentially problematic (although I could be missing one). Those are all `addToMainMenu()`-based (or something equivalent) and replace the table with a fresh one, instead of just returning a handle to the same table.

The only real equivalent is the menu order itself, but the fact that it's kept in memory just speeds things up ever so slightly without any adverse side-effects I can think of.

Frenzie:
So the other way to go about it would be to require the common settings, but make it more like a proper module with a `generateMenuItems()` method or some such.

PR #4715, opened 2019-03-02
[UX] Gesture manager: add action - fulltext search

### LINKED ISSUE #4687: Even more Gesture Manager action suggestions

Frenzie:
Fixes #4687.

### DIFF
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -21,14 +21,20 @@ end
 function ReaderSearch:addToMainMenu(menu_items)
     menu_items.fulltext_search = {
         text = _("Fulltext search"),
-        tap_input = {
-            title = _("Input text to search for"),
-            ok_text = _("Search all text"),
-            type = "text",
-            callback = function(input)
-                self:onShowSearchDialog(input)
-            end,
-        },
+        callback = function()
+            self:onShowSearchInput()
+        end,
+    }
+end
+
+function ReaderSearch:onShowFulltextSearchInput()
+    self:onInput{
+        title = _("Enter text to search for"),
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readersearch.lua
```diff
@@ -21,14 +21,20 @@ end
...
-            end,
-        },
+        callback = function()
+            self:onShowSearchInput()
+        end,
+    }
+end
+
+function ReaderSearch:onShowFulltextSearchInput()
+    self:onInput{
```
Frenzie: @poire-z This definitely looks tidier. Maybe dictionary/Wikipedia should also do it this way. :-)

poire-z: Feel free :)

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -21,14 +21,20 @@ end
...
-        tap_input = {
-            title = _("Input text to search for"),
-            ok_text = _("Search all text"),
-            type = "text",
-            callback = function(input)
-                self:onShowSearchDialog(input)
-            end,
-        },
+        callback = function()
+            self:onShowSearchInput()
```
poire-z: `readersearch.lua:25: attempt to call method 'onShowSearchInput' (a nil value)`

Frenzie: My bad, I changed the name last second because I thought it sounded too generic this way.

PR #4716, opened 2019-03-02
[UX] Gesture manager: add action - clear location history

poire-z:
Also show notification when `previous_location` can't go back because the location stack is empty.
Ref https://github.com/koreader/koreader/issues/4687#issuecomment-468956892

(I've put it in the menu, not just after `Back to previous location`, but after all the others that put stuff into this location stack, as some kind of self documentation.)

### DIFF
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -444,7 +444,7 @@ function ReaderLink:onClearLocationStack(show_notification)
     self.location_stack = {}
     if show_notification then
         UIManager:show(Notification:new{
-            text = _("Location history cleared"),
+            text = _("Location history cleared."),
             timeout = 2,
         })
     end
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerlink.lua
```diff
@@ -440,6 +440,17 @@ function ReaderLink:addCurrentLocationToStack()
     end
 end
 
+function ReaderLink:onClearLocationStack(show_notification)
+    self.location_stack = {}
+    if show_notification then
+        UIManager:show(Notification:new{
+            text = _("Location history cleared"),
```
Frenzie:
```suggestion
            text = _("Location history cleared."),
```

frontend/apps/reader/modules/readerlink.lua
```diff
@@ -613,12 +624,17 @@ function ReaderLink:onGotoLink(link, neglect_current_location, allow_footnote_po
...
-function ReaderLink:onGoBackLink()
+function ReaderLink:onGoBackLink(show_notification_if_empty)
     local saved_location = table.remove(self.location_stack)
     if saved_location then
         logger.dbg("GoBack: restoring:", saved_location)
         self.ui:handleEvent(Event:new('RestoreBookLocation', saved_location))
         return true
+    elseif show_notification_if_empty then
+        UIManager:show(Notification:new{
+            text = _("Location history is empty"),
```
Frenzie:
```suggestion
            text = _("Location history is empty."),
```

PR #4721, opened 2019-03-03
set panning rate default based on hasEinkScreen

### LINKED ISSUE #4667: Panning rate default

pazos:
in Android, where we support both e-ink/normal screens, hasColorScreen depends on hasEinkScreen.

Fixes #4667

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -21,11 +21,11 @@ local Device = Generic:new{
     hasDPad = no,
     isAndroid = yes,
     hasEinkScreen = function() return android.isEink() end,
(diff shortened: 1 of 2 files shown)


Frenzie:
Thanks!

PR #4725, opened 2019-03-04
[UX] Gesture manager: add Exit and Restart action and a few gestures

poire-z:
Reference https://github.com/koreader/koreader/issues/4687#issuecomment-469054113
Also:
- add and show some separators in the gestures list
- fix gesture removal, and also remove it from settings
- add missing east west east
- add 6 remaining of the 8 knob 3/4 rotations
- add 3 easy knob full rotation (down + east + west)

These 3/4 or full knob rotations are really practical for stuff you don't want to do by error, but still easy to do when needed. Some may appeal more to left-handed persons.

<kbd>image</kbd>

<kbd>image</kbd>
(I let the "Zoom to fit" ones consecutive)

I did not let in these ones in yellow, because the drawing is really ambiguous:
<kbd>image</kbd>

@Frenzie: removing a gesture updated the current menu, but when going up and back, we were still seeing it. Also, it was not removed from the settings, so it was still active!

I think it needs another check somewhere to remove from settings any gesture that is referenced in none of the default or custom list (and to clean up the one recorded and played with and removed before this fix).
Dunno where you'd want to have that in your code.

Also, another thing: would be nice if the gestures settings would not include the gestures mapped to "nothing". We would then have something in settings.reader.lua a little shorted than:
```
    ["gesture_reader"] = {
        ["multiswipe_north_south_north"] = "prev_chapter",
        ["multiswipe_east_north_west"] = "zoom_contentwidth",
        ["multiswipe_southeast_northeast"] = "follow_nearest_link",
        ["multiswipe_east_north"] = "history",
        ["multiswipe_east_west_east"] = "nothing",
        ["multiswipe_south_west_north_east"] = "nothing",
        ["short_diagonal_swipe"] = "full_refresh",
        ["multiswipe_west_east"] = "previous_location",
        ["multiswipe_south_west"] = "show_frontlight_dialog",
        ["tap_right_bottom_corner"] = "nothing",
        ["multiswipe_east_south"] = "go_to",
        ["multiswipe_north_east"] = "toc",
        ["multiswipe_north_west"] = "bookmarks",
        ["multiswipe_southwest_northwest"] = "nothing",
        ["multiswipe_north_south"] = "nothing",
        ["multiswipe_west_north_east"] = "nothing",
        ["multiswipe_east_west"] = "latest_bookmark",
        ["multiswipe_west_south"] = "back",
        ["multiswipe"] = "nothing",
        ["multiswipe_northeast_southeast"] = "nothing",
        ["multiswipe_east_south_west"] = "nothing",
        ["multiswipe_north_west_south"] = "nothing",
        ["multiswipe_south_north"] = "skim",
        ["multiswipe_north_east_south"] = "nothing",
        ["multiswipe_east_north_west_east"] = "zoom_pagewidth",
        ["multiswipe_northwest_southwest"] = "nothing",
        ["multiswipe_south_east_north"] = "zoom_contentheight",
        ["multiswipe_south_north_south"] = "next_chapter",
        ["tap_left_bottom_corner"] = "toggle_frontlight",
        ["multiswipe_west_north"] = "nothing",
        ["multiswipe_south_west_north"] = "nothing",
        ["multiswipe_south_east_north_south"] = "zoom_pageheight",
        ["multiswipe_west_south_east_north"] = "nothing",
        ["multiswipe_west_south_east"] = "nothing",
        ["multiswipe_east_south_west_north"] = "full_refresh",
        ["multiswipe_south_east"] = "toggle_reflow",
        ["multiswipe_south_east_north_west"] = "nothing",
        ["multiswipe_west_east_west"] = "open_previous_document"
    },
```

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -365,14 +365,34 @@ function ReaderGesture:buildMultiswipeMenu()
     local menu = {}
     multiswipes = {}
 
+    -- Build a list of gestures in settings, so we can show those
+    -- that don't appear anymore in default or custom lists, and
+    -- allow removing them (as they will still work)
+    local settings_gestures = {}
+    for k, v in pairs(gesture_manager) do
+        if k:sub(1, 11) == "multiswipe_" then
+            k = k:sub(12):gsub("_", " ")
+            settings_gestures[k] = v
+        end
+    end
+
     for k, v in pairs(default_multiswipes) do
         table.insert(multiswipes, v)
+        settings_gestures[v] = nil -- remove from settings list
     end
 
     if custom_multiswipes_table and #custom_multiswipes_table > 0 then
         table.insert(multiswipes, true) -- add separator
         for k, v in pairs(custom_multiswipes_table) do
             table.insert(multiswipes, v)
+            settings_gestures[v] = nil -- remove from settings list
+        end
+    end
+
+    if next(settings_gestures) then -- there are old gestures in settings
+        table.insert(multiswipes, true) -- add separator
+        for k, v in pairs(settings_gestures) do
+            table.insert(multiswipes, k)
         end
     end
 
@@ -399,6 +419,8 @@ function ReaderGesture:buildMultiswipeMenu()
                             text = T(_("Remove custom multiswipe %1?"), friendly_multiswipe_name),
                             ok_text = _("Remove"),
                             ok_callback = function()
+                                -- Remove associated action from settings
+                                gesture_manager[safe_multiswipe_name] = nil
                                 -- multiswipes are a combined table, first defaults, then custom
                                 -- so the right index is minus #defalt_multiswipes minus 1 added separator
                                 custom_multiswipes:removeTableItem("multiswipes", i-#default_multiswipes-1)
@@ -413,10 +435,6 @@ function ReaderGesture:buildMultiswipeMenu()
                                     table.insert(item_table, v)
                                 end
                                 touchmenu_instance:updateItems()
-                                -- Also remove associated action from settings
-                                gesture_manager[safe_multiswipe_name] = nil
-                                -- TODO: we should also some place else remove any past multiswipe_name that
-                                -- is no more referenced in our default and custom multiswipes lists
                             end,
                         })
                     end
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -714,10 +714,9 @@ function ReaderGesture:multiswipeAction(multiswipe_directions, ges)
     if not self.multiswipes_enabled then return end
     local gesture_manager = G_reader_settings:readSetting(self.ges_mode)
     local multiswipe_gesture_name = "multiswipe_"..self:safeMultiswipeName(multiswipe_directions)
-    for gesture, action in pairs(gesture_manager) do
-        if gesture == multiswipe_gesture_name then
-            return self:gestureAction(action, ges)
-        end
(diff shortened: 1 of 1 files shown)


Frenzie:
> Also, another thing: would be nice if the gestures settings would not include the gestures mapped to "nothing". We would then have something in settings.reader.lua a little shorted than:

You could see if you can rework #4623, although I thought of it as potentially advantageous for manual config editing. :-)

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -345,39 +369,61 @@ function ReaderGesture:buildMultiswipeMenu()
...
-                        end,
-                    })
-                end
-            end,
-        })
+        local separator = false
+        if i < #multiswipes and multiswipes[i+1] == true then
+            separator = true
+        end
+        if type(multiswipes[i]) == "string" then -- skip separators (true)
```
Frenzie: Or just `if type(multiswipes[i]) ~= "string" then return end`?

KenMaltby:
Not quite Witcher Signs, but getting there.

frontend/apps/reader/modules/readergesture.lua
poire-z: I don't want to `return` :) But I would have used `continue` instead, if Lua had provided that useful construct, instead of needing imbricated `if end`...
I just want to skip the `true` entry, as it was carried to the previous item.

poire-z:
> I think it needs another check somewhere to remove from settings any gesture that is referenced in none of the default or custom list (and to clean up the one recorded and played with and removed before this fix).
Dunno where you'd want to have that in your code.

Well, instead of having them auto-removed, I just added them to the list, so allowing to still see, re-assign or remove them.
Which lead me to think we would then not even need the custom multiswipes.lua: just store the recorded gestures into the settings, with "nothing" until they are assigned.

> Also, another thing: would be nice if the gestures settings would not include the gestures mapped to "nothing".

I guess you need them to be there with "nothing"  so you can bring back new ones with their default.
Well, no real idea for now. So let that be.

frontend/apps/reader/modules/readergesture.lua
Frenzie: Oh, you can actually do that (sort of).

```lua
if type(multiswipes[i]) ~= "string" then goto continue end
-- the whole shebang
::continue
```

The *bad* part of this solution is that you need to keep coming up with new names for the darn things if you have even a little bit of nesting.

Frenzie:
> Which lead me to think we would then not even need the custom multiswipes.lua: just store the recorded gestures into the settings, with "nothing" until they are assigned.

There are of course no technical reasons for it. But LuaData does happen to be crash-proof and as a user I like having separate config files for device-independent things like gestures. That is, I can copy them without concern, unlike my DPI/font-size/etc. device-specific settings (i.e., settings.lua).

I'd rather have the actions in there too, tbh. :-)

poire-z:
> having separate config files for device-independent things like gestures

Makes sense indeed.

Well, I'm then done with this PR :)

frontend/apps/reader/modules/readergesture.lua
poire-z: Well, not fond of such constructs :)
(to review the small changes induced by this added if indentation, you can always use Diff settings >  Hide whitespace changes)

Frenzie: I didn't know about that, thanks for the tip. I wasn't talking about the diff though, although it is indeed easier to focus on the code that way. :-) Like you said, I think it'd be cleaner with a continue like in, C, PHP, Python, Java, etc.

(Anyway, it's good.)

PR #4728, opened 2019-03-04
[fix] GestureDetector: deep copies of events for multiswipes when rotated

### LINKED ISSUE #4724: Landscape: swiping & wrong place in book
* KOReader version: v2019.02-155-g8ecdf9d_2019-03-04/  
* Device: KA1

#### Issue
Maybe you are already aware of this.

Since the introduction of multiswipes, swiping (including single swipes) in landscape is completely f*cked up. The biggest problem with this is that you can't even open the menus with a swipe.
Luckily it´s still possible to open the menus by tapping.

Another thing is that you end up 1 or 2 pages back (after a tap). Reproduce like this:
Go from portrait to landscape. Tap once (forward or backward).

Frenzie:
Fixes #4724.

### DIFF
--- a/frontend/device/gesturedetector.lua
+++ b/frontend/device/gesturedetector.lua
@@ -557,22 +557,22 @@ function GestureDetector:handlePan(tev)
         end
 
         if msd_distance > self.MULTISWIPE_THRESHOLD then
-            local pan_ev_ms = pan_ev
+            local pan_ev_multiswipe = pan_ev
             -- store a copy of pan_ev without rotation adjustment
             -- for multiswipe calculations when rotated
             if self.screen.cur_rotation_mode > 0 then
-                pan_ev_ms = util.tableDeepCopy(pan_ev)
+                pan_ev_multiswipe = util.tableDeepCopy(pan_ev)
             end
             if msd_direction ~= msd_direction_prev then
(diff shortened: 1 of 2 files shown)


Frenzie:
There's also some rotation-related issue with #4666.

frontend/device/gesturedetector.lua
```diff
@@ -556,16 +557,22 @@ function GestureDetector:handlePan(tev)
         end
 
         if msd_distance > self.MULTISWIPE_THRESHOLD then
+            local pan_ev_ms = pan_ev
```
poire-z: I just pondered for a few... thousands of milliiseconds :) with that `_ms` suffix - which looks too much like what we'd use to indicate milliseconds. Any other possible suffix? _msd _multiswipe?

Frenzie: I could also just call it _copy, although that might be a bit misleading when it's not a copy.

Frenzie:
@poire-z Any further comments? :-)

poire-z:
None. Not as deep as you in those low-level thingies, and can't test right now, but if it works for you :)

Frenzie:
PS There was no such thing as a rotated panning issue. That was just #4729.

PR #4737, opened 2019-03-05
[UX] Gesture manager: add action - Wallabag download

Frenzie:


frontend/apps/filemanager/filemanager.lua
```diff
@@ -357,8 +357,10 @@ function FileManager:init()
                 plugin_module, { ui = self, })
             -- Keep references to the modules which do not register into menu.
             if ok then
+                local name = plugin_module.name
+                if name then self[name] = plugin_or_err end
```
poire-z: If there is no issue inserting it with a name now, instead of just having it have an index (table.insert), looks alright.

Frenzie: It's the same thing ReaderUI does:

https://github.com/koreader/koreader/blob/39be43ad8441a8f5c2508afba261444672144dc3/frontend/apps/reader/readerui.lua#L76-L80

The table.insert index is used for internal actions like event propagation (which runs on `ipair()`), while the named index is used for our convenience.

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -335,7 +337,9 @@ function ReaderGesture:buildMenu(ges, default)
         {"zoom_pageheight", not self.is_docless},
         {"zoom_column", not self.is_docless},
         {"zoom_content", not self.is_docless},
-        {"zoom_page", not self.is_docless},
+        {"zoom_page", not self.is_docless, true},
+
+        {"wallabag_download", self.ui.wallabag ~= nil, true},
```
poire-z: In some next Gesture manager, we should remove that "true" (separator) on the last item, as it's a bit ugly.
The code I added yesterday will add a separator itself at this point, but only when there are custom gestures.

Frenzie: The whole thing could probably use some refactoring, but I haven't given it much thought. It was good enough to quickly expand with the basic functionality I wanted without too much trouble.

poire-z: Oups, I mean "in some next gesture manager _PR_" :) No problem with _that_ Gesture manager as it is!

Issue #4727, opened 2019-03-04
Gesture manager: action suggestions


mayabtlaloc:
I find it essential to navigate quickly between folders. Thanks

- [x] Folder shortcuts

Frenzie:
And it would go nicely as the file browser equivalent default gesture for bookmarks.

Also:
- [ ] Toggle scroll/page mode

poire-z:
There is one (that I don't want to have anything to do with :) that is strangely missing (that one you use Suspend or your power button to have it, if I remember correctly):
Set Wifi off (and/or Toggle Wifi).

Frenzie:
I occasionally use suspend as a quick wifi turn-off, yes.

Anyway, I think we've got a decent set of actions for release. More can just trickle in. :-)

Frenzie:
Adding a reference to #3978. Because with a multiswipe of at least three strokes, ignoring touch could work on any device.

For example, an L for lock:
<img src=https://user-images.githubusercontent.com/202757/53835532-07c88b00-3f8e-11e9-84e2-92465c68e551.png width=40%>

But definitely not for the stable. Something like that is a little too volatile to throw in last minute. ;-)

gerhaher:
I would like to see the multiswipe gesture manager itself among the actions opened with a multiswipe.

Frenzie:
- [ ] Book information
Also see https://github.com/koreader/koreader/issues/4659#issuecomment-471635233, but not so much forgotten as not wanting another last-minute disaster akin to #4733 (with #4764 and #4772).

> I would like to see the multiswipe gesture manager itself among the actions opened with a multiswipe.

For sanitary code organization purposes I would generalize that as opening a specific (sub)menu. Not likely to happen very soon I'm afraid but it is part of my overarching vision. :-)

Galunid:
It'd be nice to see "Next/Prev Highlight action" gesture(`Document > Highlight action`). Not sure how many people would be interested, but I'd like it for when I'm debugging stuff on my kindle and need to quickly go from `Highlight` to `Ask`.

Frenzie:
You should know more about that than me. :-) What is it this action would do precisely?

Galunid:
Basically it should set `default_highlight_action` in `G_reader_settings` to one of the following
`"highlight"`, `"translate"`, `"wikipedia"`, or `nil`, and display notification which one is enabled now. I'm not familiar with gesture manager, but sth like this would probably work:
```lua
local actions = {
[...]

PR #4739, opened 2019-03-05
[UX] Gesture manager: add action - toggle wifi

Frenzie:
As suggested by @poire-z.

https://github.com/koreader/koreader/issues/4727#issuecomment-469668338

I wanted to make the gesture a `W`, but that's a bit too unwieldy. Instead there's the suggestion of a (backward) `W`.

g1048

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -57,6 +57,11 @@ local action_strings = {
     toggle_frontlight = _("Toggle frontlight"),
     toggle_gsensor = _("Toggle accelerometer"),
     toggle_rotation = _("Toggle rotation"),
+
+    wifi_on = _("Enable wifi"),
+    wifi_off = _("Disable wifi"),
+    toggle_wifi = _("Toggle wifi"),
+
     toggle_reflow = _("Toggle reflow"),
 
     zoom_contentwidth = _("Zoom to fit content width"),
@@ -118,6 +123,9 @@ local default_multiswipes = {
     "northeast southeast",
     -- "southwest northwest", -- visually ambiguous
     -- "northwest southwest", -- visually ambiguous
+    "northwest southwest northwest",
+    "southeast southwest northwest",
+    "southeast northeast northwest",
 }
 local multiswipes = {}
 local multiswipes_info_text = _([[
@@ -155,6 +163,9 @@ function ReaderGesture:init()
         multiswipe_south_east_north_south = self.ges_mode == "gesture_reader" and "zoom_pageheight" or "nothing",
         multiswipe_east_south_west_north = "full_refresh",
         multiswipe_southeast_northeast = self.ges_mode == "gesture_reader" and "follow_nearest_link" or "nothing",
+        multiswipe_northwest_southwest_northwest = Device:hasWifiToggle() and "toggle_wifi" or "nothing",
+        multiswipe_southeast_southwest_northwest = Device:hasWifiToggle() and "wifi_off" or "nothing",
+        multiswipe_southeast_northeast_northwest = Device:hasWifiToggle() and "wifi_on" or "nothing",
     }
     local gm = G_reader_settings:readSetting(self.ges_mode)
     if gm == nil then G_reader_settings:saveSetting(self.ges_mode, {}) end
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -697,6 +701,20 @@ function ReaderGesture:gestureAction(action, ges)
...
+
+        if not NetworkMgr:isOnline() then
+            -- NB Normal widgets should use NetworkMgr:promptWifiOn()
+            -- This is specifically the toggle wifi action, so consent is implied.
+            NetworkMgr:turnOnWifi()
+        else
+            NetworkMgr:turnOffWifi()
+
+            UIManager:show(InfoMessage:new{
+                text = _("Wifi disabled."),
```
poire-z: No timeout to this InfoMessage? If we use gestures, it's that we want things fast and get done :) No need for this additional tap to dismiss the InfoMessage I think.
I would even use a Notification with timeout, for even less disruption.

Frenzie: Right, good point.

I was testing this and had no idea if turning it off worked or not without going to the menu, which seemed to defeat the point.

Frenzie:
As a default gesture or just as available actions?

poire-z:
As alternate available actions. Some users may prefer having two different gestures, some other only one that toggles I think.
No opinion about defaults or which of these should be set as default.

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -330,6 +339,9 @@ function ReaderGesture:buildMenu(ges, default)
         {"toggle_gsensor", Device:canToggleGSensor()},
         {"toggle_rotation", not self.is_docless, true},
         {"toggle_reflow", not self.is_docless, true},
+        {"toggle_wifi", Device:hasWifiToggle() and "toggle_wifi", true},
+        {"wifi_on", Device:hasWifiToggle() and "wifi_on", true},
+        {"wifi_off", Device:hasWifiToggle() and "wifi_off", true},
```
poire-z: Looks strange. Shouldn't they be `{string action_name, boolean enabled, boolean separator}` ?
And I guess the separator could be a bit moved with the added stuff.

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -339,9 +339,9 @@ function ReaderGesture:buildMenu(ges, default)
         {"toggle_gsensor", Device:canToggleGSensor()},
         {"toggle_rotation", not self.is_docless, true},
         {"toggle_reflow", not self.is_docless, true},
-        {"toggle_wifi", Device:hasWifiToggle() and "toggle_wifi", true},
-        {"wifi_on", Device:hasWifiToggle() and "wifi_on", true},
-        {"wifi_off", Device:hasWifiToggle() and "wifi_off", true},
+        {"toggle_wifi", Device:hasWifiToggle(), true},
+        {"wifi_on", Device:hasWifiToggle(), true},
+        {"wifi_off", Device:hasWifiToggle(), true},
```
poire-z: Still too many separators :)
All the 3 wifi ones should be in the same separated area.
And I guess you want to move toggle_reflow below them, near the Zoom to fit as it's kopt stuff

Frenzie: Depends on whether you see it as the "toggle section" or the "zoom-type stuff section". I suppose "zoom stuff" makes more sense as a grouping.

Frenzie:
(No change, just had to manually merge it with the rotation toggle change.)

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -118,6 +123,9 @@ local default_multiswipes = {
     "northeast southeast",
     -- "southwest northwest", -- visually ambiguous
     -- "northwest southwest", -- visually ambiguous
+    "northwest southwest northwest",
```
poire-z: That one is as ambiguous as the previous one I didn't dare added :) So, we might as well dare.

poire-z: Mhhh, actually it is not :) it reads the same backwards following the arrows :)

PR #4740, opened 2019-03-05
[UX] Gesture manager: fix "Toggle rotation" action

### LINKED ISSUE #4738: Gesture manager: something wrong with toggle rotation
* KOReader version: master
* Device: all

#### Issue
Something about page numbers, not sure how to repro yet. See https://github.com/koreader/koreader/issues/4724#issuecomment-469721725

poire-z:
By using the same events used by the bottom menu toggles.
Reference https://github.com/koreader/koreader/issues/4724#issuecomment-469739274. Close #4738.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -692,11 +692,9 @@ function ReaderGesture:gestureAction(action, ges)
         self.ui:handleEvent(Event:new("RestoreZoomMode"))
         self.ui:handleEvent(Event:new("InitScrollPageStates"))
     elseif action == "toggle_rotation" then
-        if Screen:getScreenMode() == "portrait" then
-            self.ui:handleEvent(Event:new("SetScreenMode", "landscape"))
-        else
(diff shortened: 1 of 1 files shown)


Frenzie:
Can SwapScreenMode also do arbitrary rotation?

poire-z:
Some litterature about that :)
https://github.com/koreader/koreader/blob/addd06c1debcb48dfa338b60fdf24e25ce50019d/frontend/apps/reader/modules/readerview.lua#L685-L727

Not sure I understand the whole rotation stuff involved, or if we could/should have used SwapScreenMode for both cre and pdf (@NiLuJe ?) - but I guess doing as the toggles do should be safe.

NiLuJe:
Yep, that's what the second argument can stand for ;).

The best example of that is probably the Forma gsensor stuff in frontend/Device/input.lua ;).

NiLuJe:
@poire-z: The onSwapScreenMode handler(s) will take care of dispatching the right stuff for both CRe & PDF.

PR #4744, opened 2019-03-05
Bump fonts

### LINKED ISSUE #4707: Incorrect north east/west arrows in Noto font
* KOReader version: master

#### Issue
North east and north west arrows are swapped in the Noto font.

Take these symbols:
```lua
    northeast = "⬈", -- U+2B08 NORTH EAST BLACK ARROW
    northwest = "⬉", -- U+2B09 NORTH WEST BLACK ARROW
    southeast = "⬊", -- U+2B0A SOUTH EAST BLACK ARROW
    southwest = "⬋", -- U+2B0B SOUTH WEST BLACK ARROW
```

Only the bottom two are displayed in the correct direction.

Pinging @NiLuJe who knows more about the font.

(For the diagonal multiswipes it doesn't really matter. I can just swap the symbols with a comment.)

### LINKED ISSUE #4743: Update git command in README.md
I couldn't build at my first try cloning using git clone https://github.com/koreader/koreader.git

(missing luafilesystem/)

Then I figure out by the base/.gitmodules that some submodules need to be fetched.

So I suggest update the doc file:

https://github.com/koreader/koreader/blob/master/README.md#getting-the-source

for include the submodules and could be changed to:

git clone --recurse-submodules https://github.com/koreader/koreader.git

Frenzie:
Fixes #4743 (shallow submodule + GitHub bug). Updated FreeSans fixes #4707 todo (swapped arrows).

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -492,8 +492,8 @@ local multiswipe_to_arrow = {
     west = "⬅",
     north = "⬆",
     south = "⬇",
-    northeast = "⬉", -- @TODO swap back, see https://github.com/koreader/koreader/issues/4707
-    northwest = "⬈", -- @TODO swap back, see https://github.com/koreader/koreader/issues/4707
+    northeast = "⬈",
+    northwest = "⬉",
     southeast = "⬊",
     southwest = "⬋",
 }


poire-z:
There may be a problem with FreeSerif. @NiLuJe : did it get a specific treatment?

With 2019-02-25, all is fine, but with current master, with FreeSerif only, and with Kerning "Good" only ("fast" and "best" are fine):

<kbd>image</kbd>

Replacing current FreeSerif.ttf (2 098 176 bytes ) with the one from 2019-02 (3 303 588 bytes) fixes the issue - so probably not some crengine issue - unless there are some new kind of metrics that crengine is not aware of.

NiLuJe:
I'm not seeing anything so egregious here, can you only reproduce this on some type of content?

(Might be dpi related, I only checked on the Forma).

I did have to rebuild Free* from source, so there may be weird quirks.
Plus, I also switched to the OpenType variant...

NiLuJe:
Does either of these work properly? (You'll have to rename the OTF to .ttf).

[FreeFont-OTF.tar.gz](https://github.com/koreader/koreader/files/2969062/FreeFont-OTF.tar.gz)
[FreeFont-TTF.tar.gz](https://github.com/koreader/koreader/files/2969063/FreeFont-TTF.tar.gz)

poire-z:
Happens too with any of the FreeSerif from these 2 files (no matter if I let the extension .otf or rename it to .ttf).

But, it seems to happen only when the previous word contains some accentuated characters. So easier to witness with french texts than with english ones :) And only with "good" kerning.
Might be some codepoint/index/table/encoding mismatch when index > 256 (so, plain ascii chars are fine)?

NiLuJe:
Okay, round two! These ones built w/ the recommended, awfully old, FontForge version...

[FreeFont-OTF-LegacyFF.tar.gz](https://github.com/koreader/koreader/files/2972980/FreeFont-OTF-LegacyFF.tar.gz)
[FreeFont-TTF-LegacyFF.tar.gz](https://github.com/koreader/koreader/files/2972981/FreeFont-TTF-LegacyFF.tar.gz)

poire-z:
Nope, no change with these ones, still ugly with "good".
Can you or not reproduce it? (with frech texts)

NiLuJe:
I don't actually have anything in French on the device that's on my desk :D.

poire-z:
It works with pure harfbuzz ("best"). Just not with hacky "good". But old FreeSerif did.
(hacky "good" get metrics from HB, deals with clusters, and we apply them in crengine - it's not pure HB)
@Frenzie : can you confirm you get these glitches with FreeSerif, with kerning good, with DaVinci.FR.epub :) ?

NiLuJe:
Okay, yeah, found something in French, and, yeah, I can reproduce that.

Not quite sure what's happening, except for obviously broken kerning somewhere ;).

NiLuJe:
@poire-z : Okay, the ones in https://github.com/koreader/koreader-fonts/pull/5 work on my end, despite FontForge shouting at me (a lot).

Plus, they're smaller, so, eh.

poire-z:
These work for me too :)

In the end, do you think it was really an issue in these font files?
Or they were valid and we may have a bug in the "kerning good" code introduced in https://github.com/koreader/crengine/pull/251 ?

NiLuJe:
No idea, but technically that version probably just bypasses the issue
because the offending tables get invalidated by missing glyphs.

Whether the tables are sane on the version with the properly indexed glyphs
is anyone's guess ;).

poire-z:
Mhhh, there must be something rotten in the "good" kerning implementation.
With another font (Fedra Serif A pro, that seems to have a ligature for "Th"):

<kbd>image</kbd>

and when you dismiss the config panel, this happens:

<kbd>image</kbd>

There's something like the rendering is not deterministic - on the 2nd rendering is not the same as the 1st (and the 2nd is not fixed, it's just a bit less bad than the 1st). I noticed that too with FreeSerif: selecting text moved the text !

So, I don't know. But in our "good" code - or bug in HB when used with +kern but -liga ?

May be hold on a bit on the font PR.

poire-z:
The issue (and solution) must be in https://github.com/koreader/crengine/blob/535c0afac77af638c35974989663bcd376668cae/crengine/src/lvfntman.cpp#L1247-L1293

With current FreeSerif:
image
```
hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 4, must be 3, return value (-1)
FAILED measuring triplet 116 114 232 trè
hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 4, must be 3, return value (-1)
FAILED measuring triplet 114 232 115 rès
hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 4, must be 3, return value (-1)
FAILED measuring triplet 232 115 32 ès
```

With Fedra Serif:
image

```
hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 2, must be 3, return value (-1)
FAILED measuring triplet 32 84 104  Th
hbCalcCharWidth
hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 2, must be 3, return value (-1)
FAILED measuring triplet 84 104 232 Thè
```

I get that even if I change:
```diff
        // HarfBuzz features for full text shaping
        hb_feature_from_string("+kern", -1, &_hb_features[0]);      // font kerning
        hb_feature_from_string("+liga", -1, &_hb_features[1]);      // ligatures
        // HarfBuzz features for lighweight characters width calculating with caching
-       hb_feature_from_string("+kern", -1, &_hb_light_features[0]);      // font kerning
+       hb_feature_from_string("-kern", -1, &_hb_light_features[0]);      // font kerning
        hb_feature_from_string("-liga", -1, &_hb_light_features[1]);      // no ligatures
```

Using `if (true || segLen == glyph_count )` avoid these failures, and the rendering is less awful, but still some width issues.

Pinging @virxkane in case you get some idea about how to have your hbCalcCharWidth() deal with such cases (happens only with some specific fonts, including [our current FreeSerif](https://github.com/koreader/koreader-fonts/raw/master/freefont/FreeSerif.ttf)). (Can you check if you have the same issue with this font on CoolReader - if not, we can suspect our Harfbuzz recent versions - or my own tweaks around your code :| Thanks ! :)

poire-z:
I guess some little char index vs cluster index skip or use-previous stuff will be needed:
```
hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 4, must be 3, return value (-1)
  idx0 cluster=0 codepoint=86 offset=0 advance=417
  idx1 cluster=1 codepoint=84 offset=0 advance=496
  idx2 cluster=2 codepoint=71 offset=0 advance=654
  idx3 cluster=2 codepoint=71 offset=0 advance=654
FAILED measuring triplet 116 114 232 trè

hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 4, must be 3, return value (-1)
  idx0 cluster=0 codepoint=84 offset=0 advance=496
  idx1 cluster=1 codepoint=71 offset=0 advance=654
  idx2 cluster=1 codepoint=71 offset=0 advance=654
  idx3 cluster=2 codepoint=705 offset=-70 advance=0
FAILED measuring triplet 114 232 115 rès

hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 4, must be 3, return value (-1)
  idx0 cluster=0 codepoint=71 offset=0 advance=654
  idx1 cluster=0 codepoint=71 offset=0 advance=654
  idx2 cluster=1 codepoint=705 offset=-70 advance=0
  idx3 cluster=2 codepoint=85 offset=0 advance=540
FAILED measuring triplet 232 115 32 ès

hbCalcCharWidthWithKerning(): hb_buffer_get_length() return 2, must be 3, return value (-1)
  idx0 cluster=0 codepoint=412 offset=0 advance=1800
  idx1 cluster=2 codepoint=152 offset=0 advance=930
FAILED measuring triplet 84 104 101 The
```

(When HB "light" was just to avoid having to deal with these clusters)

poire-z:
> I guess some little char index vs cluster index skip or use-previous stuff will be ~~needed~~ painful

But the issues with these 2 fonts are solved by just disabling some additional opentype features:
```diff
         // HarfBuzz features for lighweight characters width calculating with caching
         hb_feature_from_string("+kern", -1, &_hb_light_features[0]);      // font kerning
         hb_feature_from_string("-liga", -1, &_hb_light_features[1]);      // no ligatures
+        hb_feature_from_string("-ccmp", -1, &_hb_light_features[2]);      // no composition/decomposition
+                                                                          // (needed with FreeSerif)
+        hb_feature_from_string("-calt", -1, &_hb_light_features[3]);      // no contextual alternates
+                                                                          // (needed with Fedra Serif)
```

Other fonts may need to have other features disabled...
So, I don't know if we should disable all those that Harfbuzz seems to enable by default in
https://github.com/harfbuzz/harfbuzz/blob/bcb4e505d6ffe33e3268a06698e75d6be0e64957/src/hb-ot-shape.cc#L229-L333

Some docs about opentype features:
https://en.wikipedia.org/wiki/List_of_typographic_features
https://docs.microsoft.com/en-us/typography/opentype/spec/featuretags
https://sparanoid.com/lab/opentype-features/

So, @NiLuJe , I guess you can kill https://github.com/koreader/koreader-fonts/pull/5 - unless the "smaller" factor is appealing enough.

NiLuJe:
I'll just rebuild 'em with the latest FF/FT variant and the least amount of bloat possible then ;).

Frenzie:
> (and up to which point HB is still HB when we disable these features :)

Contextual alternates are at least as important as ligatures in script fonts, but it is in a sense more or less the same thing, so it makes sense that it'd break the same way.

I believe Unicode decomposition is fairly important for actually finding characters because it means the font doesn't have to include separate glyphs for all the combined diacritics and whatnot. (Not entirely sure how that could cause trouble? Could it be a case of incorrectly returning true somewhere when a glyph isn't actually found?)

virxkane:
Can you upload file "DaVinci.FR.epub" for me? I want to test in CoolReader with font "FreeSerif.ttf".
Font I get from repo "koreader-fonts", it's right?

Frenzie:
Oh, it's nothing special. Just a random Wikipedia page. The important thing is only that because it's French, it has a relatively high occurrence of characters like é.

[Da Vinci Code.FR.epub.zip](https://github.com/koreader/koreader/files/2974887/Da.Vinci.Code.FR.epub.zip)

Screenshot_2019-03-17_08-25-14

virxkane:
Hmm, with this font and this file CoolReader just freezes...

poire-z:
@virxkane : freezes, taking 100% cpu ?

If yes, I remember I've had something like that and fixed it with adding a `lastFitChar = i + 1;` somewhere.
Just in case, can you try adding it in your `else` just before the `continue` in
https://github.com/buggins/coolreader/blob/c68a4e174d5b9516df6fda572f361c747ae253ed/crengine/src/lvfntman.cpp#L1523-L1530
like we have it here:
https://github.com/koreader/crengine/blob/535c0afac77af638c35974989663bcd376668cae/crengine/src/lvfntman.cpp#L1587-L1594

virxkane:
@poire-z, yes, this helps and now I see the same glitches. I implemented light HarfBuzz rendering (you call this mode KERNING_MODE_HARFBUZZ_LIGHT) to speed up rendering process by excluding call of function hb_shape() and using own glyph's position cache. But this requires strong conformity: one character - one glyph. Therefore, ligatures are prohibited in this mode, but with font 'FreeSerif', character 'è' divided into two glyphs with indexes 0x47 and 0x02C2. So I think only one right resolution is also prohibit such separation, i.e. using feature '-ccmp' and other...

P.S. Sorry bad english.

Frenzie:
> with font 'FreeSerif', character 'è' divided into two glyphs with indexes 0x47 and 0x02C2

I would expect this to become increasingly common with Unicode composition.

@NiLuJe Is this something you can affect while building the font? It'd grow in size but perhaps not extravagantly.

poire-z:
> But this requires strong conformity: one character - one glyph. 

Yep, I figured that too.
So, we need to exclude all features that may split a char into more glyphs, or merge chars into one glyph.

Will need to read about each one of these and possibly not disable some of these:
```
+        hb_feature_from_string("-ccmp", -1, &_hb_light_features[2]);      // no composition/decomposition
+                                                                          // (needed with FreeSerif)
+        hb_feature_from_string("-calt", -1, &_hb_light_features[3]);      // no contextual alternates
+                                                                          // (needed with Fedra Serif)
+
+        // We may as well disable all features enabled by default
+        // in harfbuzz/src/hb-ot-shape.cc hb_ot_shape_collect_features()
+        hb_feature_from_string("-locl", -1, &_hb_light_features[4]);
+        hb_feature_from_string("-mark", -1, &_hb_light_features[5]);
+        hb_feature_from_string("-mkmk", -1, &_hb_light_features[6]);
+        hb_feature_from_string("-rlig", -1, &_hb_light_features[7]);
+        hb_feature_from_string("-clig", -1, &_hb_light_features[8]);
+        hb_feature_from_string("-curs", -1, &_hb_light_features[9]);
+        hb_feature_from_string("-rclt", -1, &_hb_light_features[10]);
+        hb_feature_from_string("-rvrn", -1, &_hb_light_features[11]);
+        hb_feature_from_string("-ltra", -1, &_hb_light_features[12]);
+        hb_feature_from_string("-ltrm", -1, &_hb_light_features[13]);
+        hb_feature_from_string("-rtla", -1, &_hb_light_features[14]);
+        hb_feature_from_string("-rtlm", -1, &_hb_light_features[15]);
+        hb_feature_from_string("-frac", -1, &_hb_light_features[16]);
+        hb_feature_from_string("-numr", -1, &_hb_light_features[17]);
+        hb_feature_from_string("-dnom", -1, &_hb_light_features[18]);
+        hb_feature_from_string("-rand", -1, &_hb_light_features[19]);
+        hb_feature_from_string("-trak", -1, &_hb_light_features[20]);
+        hb_feature_from_string("-vert", -1, &_hb_light_features[21]);
```

@Frenzie : Note that this is juft for HB light - everything seems full featured with HB full - so no need to worry about extra feature packed fonts I think.

Frenzie:
@poire-z So it's not a problem if you have a font that only comes with the combining acute accent and e, without some derivative like `é`?

poire-z:
I don't know about such fonts. But FreeSerif seems to come with both.
If we don't disable `ccmp`, hb_shape will give 2 glyphs for `é` : `e` + `'` (acute, with negative advance).
If we disable `ccmp`, well, it renders a `é`, so it must be using a single `é` glyph (no other fallback font in use).

Same with Fedra Serif and `calt`. with calt, it insists of giving a single glyph for `Th` - without, it gives properly `T` and `h`

And in any case, if there are such fonts with just separate glyphs, and they display badly in HB light, the user has the option to switch to HB full.

Frenzie:
`é` is a rather contrived example since that grapheme will probably occur as a single character in fonts. It's not terribly difficult to come up with slightly more exotic graphemes.

* X́x́
* X̂x̂
* Ẍẍ 
* X̌x̌
* Ẋẋ 
* X̧x̧ 
* X̱x̱
* X̣x̣
* ᶍ

Based purely on my impression that HB and equivalents have become commonplace, I assume that fonts could theoretically lose a fair bit of weight by not including all manner of such combinations as more than a few alignment instructions.

But regardless of what the font does, if someone creates a document that contains a combining acute accent with an e it still has to be normalized to é. I should have phrased it that way, as opposed to my tangential thought about light-weight fonts.

poire-z:
> if someone creates a document that contains a combining acute accent with an e it still has to be normalized to é

This should work HB full - @virxkane first (full) implementation deals correctly with these "clusters".
It's only the 2nd (light) implementation that requires no clustering (1 codepoint = 1 cluster = 1 glyph).

For HB light, I think we could go with this:
```c
        // HarfBuzz features for lighweight characters width calculating with caching
        // We need to disable all the features, enabled by default in Harfbuzz, that
        // may split a char into more glyphs, or merge chars into one glyph.
        // (see harfbuzz/src/hb-ot-shape.cc hb_ot_shape_collect_features() )

        // We can enable these ones:
        hb_feature_from_string("+kern", -1, &_hb_light_features[0]);  // Kerning: Fine horizontal positioning of one glyph to the next, based on the shapes of the glyphs
        hb_feature_from_string("+mark", -1, &_hb_light_features[1]);  // Mark Positioning: Fine positioning of a mark glyph to a base character
        hb_feature_from_string("+mkmk", -1, &_hb_light_features[2]);  // Mark-to-mark Positioning: Fine positioning of a mark glyph to another mark character
        hb_feature_from_string("+curs", -1, &_hb_light_features[3]);  // Cursive Positioning: Precise positioning of a letter's connection to an adjacent one
        hb_feature_from_string("+locl", -1, &_hb_light_features[4]);  // Substitutes character with the preferred form based on script language

        // We should disable these ones:
        hb_feature_from_string("-liga", -1, &_hb_light_features[5]);  // Standard Ligatures: replaces (by default) sequence of characters with a single ligature glyph
        hb_feature_from_string("-rlig", -1, &_hb_light_features[6]);  // Ligatures required for correct text display (any script, but in cursive) - Arabic, semitic
        hb_feature_from_string("-clig", -1, &_hb_light_features[7]);  // Applies a second ligature feature based on a match of a character pattern within a context of surrounding patterns
        hb_feature_from_string("-ccmp", -1, &_hb_light_features[8]);  // Glyph composition/decomposition: either calls a ligature replacement on a sequence of characters or replaces a character with a sequence of glyphs
                                                                      // Provides logic that can for example effectively alter the order of input characters
                                                                      // (needed with FreeSerif)
        hb_feature_from_string("-calt", -1, &_hb_light_features[9]);  // Contextual Alternates: Applies a second substitution feature based on a match of a character pattern within a context of surrounding patterns
                                                                      // (needed with Fedra Serif)
        hb_feature_from_string("-rclt", -1, &_hb_light_features[10]); // Required Contextual Alternates: Contextual alternates required for correct text display which differs from the default join for other letters, required especially important by Arabic
        hb_feature_from_string("-rvrn", -1, &_hb_light_features[11]); // Required Variation Alternates: Special variants of a single character, which need apply to specific font variation, required by variable fonts
        hb_feature_from_string("-ltra", -1, &_hb_light_features[12]); // Left-to-right glyph alternates: Replaces characters with forms befitting left-to-right presentation
        hb_feature_from_string("-ltrm", -1, &_hb_light_features[13]); // Left-to-right mirrored forms: Replaces characters with possibly mirrored forms befitting left-to-right presentation
        hb_feature_from_string("-rtla", -1, &_hb_light_features[14]); // Right-to-left glyph alternates: Replaces characters with forms befitting right-to-left presentation
        hb_feature_from_string("-rtlm", -1, &_hb_light_features[15]); // Right-to-left mirrored forms: Replaces characters with possibly mirrored forms befitting right-to-left presentation
        hb_feature_from_string("-frac", -1, &_hb_light_features[16]); // Fractions: Converts figures separated by slash with diagonal fraction
        hb_feature_from_string("-numr", -1, &_hb_light_features[17]); // Numerator: Converts to appropriate fraction numerator form, invoked by frac
        hb_feature_from_string("-dnom", -1, &_hb_light_features[18]); // Denominator: Converts to appropriate fraction denominator form, invoked by frac
        hb_feature_from_string("-rand", -1, &_hb_light_features[19]); // Replaces character with random forms (meant to simulate handwriting)
        hb_feature_from_string("-trak", -1, &_hb_light_features[20]); // Tracking (?)
        hb_feature_from_string("-vert", -1, &_hb_light_features[21]); // Vertical (?)

```

poire-z:
OK, PR'ed that at https://github.com/koreader/crengine/pull/271

PR #4748, opened 2019-03-06
import ntx_hwconfig into environment variables

BloodRagg:
sets all ntx_hwconfig variables into the shell environment

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/4748)

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -1,6 +1,9 @@
 #!/bin/sh
 export LC_ALL="en_US.UTF-8"
 
+# import ntx_hwconfig into environment variables
+export $(ntx_hwconfig -s /dev/mmcblk0 |sed -n "/^\[/s,^.[^ ]* ,,;s,',,gp")
+
 # working directory of koreader
 KOREADER_DIR="${0%/*}"
 
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -2,7 +2,7 @@
 export LC_ALL="en_US.UTF-8"
 
 # import ntx_hwconfig into environment variables
-export $(ntx_hwconfig -s /dev/mmcblk0 |sed -n "/^\[/s,^.[^ ]* ,,;s,',,gp")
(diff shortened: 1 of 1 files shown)


NiLuJe:
Move it behind a var check after the siphon attempts, and *only import the relevant variables*.

We need a clean env to be able to restart Nickel on exit, so this can only be done *as needed* (i.e., on Nickel/KSM-less systems).

NiLuJe:
Otherwise, the -standalone shim approach used by @pazos on Cervantes seems like a saner solution, because Nickel is finicky as hell, and I really, really, really don't want to have this messing with 99.9999% of the setups out there ;).

pazos:
Hi @BloodRagg, thanks for your time!

As NiLuJe said, I think it's better to leave koreader.sh as is an implement new stuff outside.

I'm not sure what you want to do, but I'm thinking about something like running KOReader on a buildroot/yocto/debian rootfs. In that case you'll need to write a new file that does everything that kobo needs to run koreader (like having PRODUCT and WIFI, mounting filesystems and disabling leds...). Once you have a new file then please test it using that file on inittab instead of rcS and check that everything works ok.

NiLuJe:
I'm not overly against making the main script slightly more fall-backy, but it has to be done *very* carefully.

In this specific context, I'm not quite seeing the hurry, as we already have fallbacks for everything except WIFI_MODULE ;).

PRODUCT *does* depend on the kobo_config script, but for good reason. Having had to reimplement that in C, I can tell you that, no, we really don't want to maintain yet another copy of that ourselves.

NiLuJe:
On the other hand, stuff like handling rotation quirks on the devices with the 6.8" inverted panel *really* don't belong in there, though, and, in that case, the -standalone shim really makes much more sense.

poire-z:
Why not just a new `koreader-bootstrap.sh`, that would set all the environment/files (from ntx_config or whatever) just like our main `koreader.sh` wants them - and just call our untouched `koreader.sh` (or with possibly new env variables to disable some of the stuff we do, if really needed.)

NiLuJe:
Yep, that's the kind of approach I was thinking of, and kind of what I meant by "shim" ;).

pazos:
Closing this as nobody is working on it and is just a one-liner. Feel free to reopen if you like ;)

Issue #772, opened 2014-08-08
EPUB: In two page mode some highlights do not show up


WS64:
I guess this is related to two pages being shown at the same time, and Koreader just shows the hightlight on the active page, but since there are two pages shown there is a 1:1 chance a highlight will not be displayed.
The highlighting itself works fine, the correct text is selected and everything, it is just that this highlight is not shown in the text. if you switch back to 1 page the hightlight shows up.
Since the drawing of the inverted text before is absolutely fine too I guess this should be very easy to fix, but I cannot find the related part in the source code.

smartscripts-nl:
This issue has been closed, but I have this issue in koreader 2019.2 on my Kobo Forma. So highlights in the second/right column are not shown and only become visible when switching to portrait lay-out.

smartscripts-nl:
Additionally: when I start a highlight in the left column and it continues in the right colum, the first line of text in the right column is shown as not highlighted and the lines below that as highlighted.

poire-z:
Confirmed. It's quite messed up in 2-pages mode.
(I need `DCREREADER_TWO_PAGE_THRESHOLD = 2` in defaults.lua to get them after I resize my emulator to full screen.)

smartscripts-nl:
Hello poire-z,

Thank you for your answer! Changing this setting doesn't solve my problem. Is there any chance this bug will be fixed in an upcoming release?
Sincerely,

Alex Pot
The Netherlands

	versnel en optimaliseer je website
www.smartscripts.nl — www.waalweb.nl  
KVK 34286209 | BTW NL 098677810B01
06 - 420 190 74 | 023 - 5 271 204
Haarlem

Van: poire-z <notifications@github.com>
Beantwoord: koreader/koreader <reply@reply.github.com>
Datum: 7 maart 2019 at 07:46:32
Aan: koreader/koreader <koreader@noreply.github.com>
Cc: smartscripts-nl <info@smartscripts.nl>, Comment <comment@noreply.github.com>
Onderwerp:  Re: [koreader/koreader] In two page mode some highlights do not show up (#772)  

Confirmed. It's quite messed up in 2-pages mode.
(I need DCREREADER_TWO_PAGE_THRESHOLD = 2 in defaults.lua to get them after I resize my emulator to full screen.)

poire-z:
> Is there any chance this bug will be fixed 

There are chances :)
[...]

PR #4753, opened 2019-03-07
cre: fix highlights in 2-pages mode

poire-z:
Also check for boxes in the 2nd page when in 2 page modes. Fix https://github.com/koreader/koreader/issues/772#issuecomment-470106759
(And fix onTapXPointerSavedHighlight(), too quickly cut and pasted from drawXPointerSavedHighlight() recently in #4663).

Bump crengine and cre.cpp to correctly returns rect for the first line of the 2nd page. See https://github.com/koreader/koreader-base/pull/844. Fix https://github.com/koreader/koreader/issues/772#issuecomment-470107420


poire-z:
(Wait for bump for https://github.com/koreader/koreader-base/pull/844 to be included.)

PR #4759, opened 2019-03-08
Gesture manager: allow folder shortcuts as action in Reader

### LINKED ISSUE #4755: Gesture manager: add action - folder shortcuts Directly from the reading section.
It is possible to implement this option to access directly from the reading of a book and not only from the file browser?. Thanks.

Frenzie:
Note that `self.ui:onClose()` properly closes open documents. There might be similar pieces of code around that don't do that.

Fixes #4755.

### DIFF
--- a/frontend/apps/filemanager/filemanagershortcuts.lua
+++ b/frontend/apps/filemanager/filemanagershortcuts.lua
@@ -190,13 +190,24 @@ function FileManagerShortcuts:deleteFolderShortcut(item)
 end
 
 function FileManagerShortcuts:onShowFolderShortcutsDialog()
-    local fm_bookmark =  FileManagerShortcuts:new{
+    local fm_bookmark = self:new{
         title = _("Folder shortcuts"),
         show_parent = self.ui,
-        curr_path = self.ui.file_chooser.path,
+        curr_path = self.ui.file_chooser and self.ui.file_chooser.path or self.ui:getLastDirFile(),
         goFolder = function(folder)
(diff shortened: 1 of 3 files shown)


frontend/apps/filemanager/filemanagershortcuts.lua
```diff
@@ -190,13 +190,24 @@ function FileManagerShortcuts:deleteFolderShortcut(item)
...
-        curr_path = self.ui.file_chooser.path,
+        curr_path = self.ui.file_chooser and self.ui.file_chooser.path or self.ui:getLastDirFile(),
         goFolder = function(folder)
             if folder ~= nil and lfs.attributes(folder, "mode") == "directory" then
-                self.ui.file_chooser:changeToPath(folder)
+                if self.ui.file_chooser then
+                    self.ui.file_chooser:changeToPath(folder)
+                else
+                    local FileManager = require("apps/filemanager/filemanager")
+
```
poire-z: May be just add a comment for that `else` that in that case, we are called from the reader - that's what I suppose it's for  (and remove that blank like)

Frenzie: What's against the blank line? :-)

poire-z: What is it for? :)
(Feels like you forgot it - I use these to indicate some different logical/work/context block of code - but it feels like it's not the case there.)

Frenzie: In this case a combination of legibility and the logical separation between defining the variables being used and doing stuff.

Issue #4726, opened 2019-03-04
Since -155 build, the screen dpi settings are ignored


ptrm:
* KOReader version: v2019.02-155-g8ecdf9d_2019-03-04
* Device: Kobo Forma / Clara HD

#### Issue

I use a 200dpi custom screen dpi setting. After the update from `-130` to `-155`, the screen dpi is forced to 320 dpi, and even checking other checkboxes while in the menu doesn't uncheck old / check new selection.  `["screen_dpi"] = 200` was still present in the settings.reader.lua during all changes, after exiting koreader to make sure it saved.

After unzipping -130 back, the dpi settings are maintained as expected.

#### Steps to reproduce
Attempt to change screen dpi in the `Screen > Screen DPI` menu. The checkboxes remain intact and after restart the dpi is intact as well.

Frenzie:
See https://github.com/koreader/koreader/commit/dd1ab7b409d73d270c4b1477bd30ba89ec6e245f#commitcomment-32573579

poire-z:
That one line added patch seems to work. Who's adding it so we get it in tomorrow's nightlies?

Frenzie:
I'm not sure if it's quite the proper solution. I think there's an idea there about setting fb.dpi to the actual DPI of the screen with a custom_dpi for scaleBySize. (But of course you could stick in there as a quick fix.)

poire-z:
Well, dunno. Looks like #4580 decouples a bit too much :) Pinging @houqp .

There is this added:
https://github.com/koreader/koreader/blob/710ab1ba1b9c2d0d40fac35691989474d56bc952/reader.lua#L37-L41

I guess calling `require("device")` does already many things including calling getDPI(), and the later setDPI() lose:
```
ffi.load (warning): libgtk-3.so.0: cannot open shared object file: No such file or directory
03/04/19-21:26:00 WARN  fb:getDPI
03/04/19-21:26:00 WARN  fb:getDPI 160
03/04/19-21:26:00 WARN  fb:getDPI 160
03/04/19-21:26:00 WARN  fb:getDPI 160
03/04/19-21:26:00 INFO  initializing for device Emulator
03/04/19-21:26:00 INFO  framebuffer resolution: {    ["h"] = 610,    ["w"] = 600}
03/04/19-21:26:00 WARN  fb:setDPI 120    -- too late
03/04/19-21:26:00 WARN  fb:getDPI 160
```

Frenzie:
[...]

PR #4771, opened 2019-03-11
fix: auto dpi menu checkbox display

houqp:
this addresses https://github.com/koreader/koreader/issues/4726#issuecomment-469585801.


Frenzie:
Thanks!

Issue #4115, opened 2018-07-27
Enable ''Add Note'' option when highlighting


embryo10:
This is a feature request.

A standard use of my reader (except of reading ;o) is to fix typos and other book related errors.
The way I do it is to first highlight the part of the text with the error, then press the highlighted block and select edit and then add the description of the fix or the correct word.
I was thinking if I could somehow save some clicks there (some books have tens of errors) when I noticed the disabled "Add Note" option to the highlight dialog.

In the ["Bookmark titles or comments?"](https://github.com/koreader/koreader/issues/3415) I think that we concluded that no new comment/note field is going to get added to KoReader anytime soon.
So, how about using this "Add Note" option to highlight and open the edit dialog in one go?

It might also need some thinking of the naming scheme (bookmark name, comment, note..)

Frenzie:
I think that button is actually a placeholder for PDF comments. See eb4c76bd159363315141822d709ce3661a7d25ef.

embryo10:
Sorry, never got any notifications from GitHub for the replies..

Are they any different than the "text" key of the epub highlight?
If not, it would be nice to enable it for epubs to do that editing.

Frenzie:
PDF comments can either just be applied to a random part of the page or to a highlight. So I guess they can be the same to some degree. Depends a bit on the MuPDF API for them too.

Btw where was this conclusion thing you mentioned? :-)

embryo10:
It wasn't a conclusion per se.
It just left me with the impression that apart from an `"edited" = true` key, nothing more was scheduled for the near future...

embryo10:
Added from v2019.04 onward... Great!

Issue #1095, opened 2014-11-06
Feature request: note taking functionality


bilyyuan:
As the title. It bothered me very much.
I forget whether it is has worked or not.
For both old version [at least 6 month ago] and latest version, notes are not workable.
Added: Other function are works, problem only effects note. it exists on both scanning and non scanning pdf books.

Markismus:
[Chrox mentioned](https://github.com/koreader/koreader/pull/464#issuecomment-32706593) a year ago that the text input module should be improved before we could use notes. This hasn't happened yet. 
The keyboard layout still boggles me, too. For instance when I am trying to find ß. It's there, though buried.

richo67:
I wonder, is it still the case that keyboard layout is blocking factor for having note taking and highlight feature in koreader?

houqp:
Not anymore, we have full keyboard now. Although we still don't have input method for CJK.

richo67:
Please excuse my ignorance what is CJK?

KenMaltby:
Chinese, Japanese, Korean

richo67:
Thanks Ken!
Is the  lack of input for these languages the reason for not having possibility to write notes for highlighted text in ebook (ePub,mobi,..)?

houqp:
Thanks Ken for the explanation. Lack of CJK input method is only a blocker for taking notes in CJK languages. There is no blocker for supporting notes taking for languages like English.

joelostblom:
Does this issue encompass all types of note-taking features in KOreader or specifically annotating highlighted text? Phrased another way: If I want to write something down while reading an epub/PDF, is there currently any possible way of doing this with KOreader (even if it involves switching to a dedicated note taking app or annotating a bookmark/page rather than a selection of text)?

If this is not possible yet, is an annotation feature on the roadmap (or maybe even being worked on already) or is it still in the planning stages and unlikely to see a release in the upcoming months/years?

mergen3107:
So are there any news on adding this notes functionality?
Moreover, why can't you just start implementing this feature at least for current input method?
That's seems unfair to me, because I don't use CJK input, but make notes on my book highlights very often.

sqlicone:
Will the possibility of adding notes to PDF highlights be added any time in the future??
It's a really crucial feature for me, as I work at a university and read lots of pdfs that I need to note.
[...]

PR #4773, opened 2019-03-11
Enable addNote button, by letting it edit highlight

Galunid:
Reuses already existing highlight editing functionality, to enable "AddNote" in highlight menu. This basically does what was asked in #4115 #1095

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -974,7 +974,7 @@ function ReaderHighlight:exportToDocument(page, item)
 end
 
 function ReaderHighlight:addNote()
-    self:handleEvent(Event:new("AddNote"))
+    self.ui:handleEvent(Event:new("AddNote"))
     local page, index = self:saveHighlight()
     self:editHighlight(page, index)
     UIManager:close(self.edit_highlight_dialog)
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -721,13 +721,8 @@ function ReaderHighlight:onHoldRelease()
                     },
                     {
                         text = _("Add Note"),
-                        enabled = true,
                         callback = function()
-                            local page, index = self:saveHighlight()
-                            if page and index then -- saveHighlight can return nil, nil
-                                self:editHighlight(page, index)
-                                UIManager:close(self.edit_highlight_dialog)
-                            end
+                            self:addNote()
                             self:onClose()
                         end,
                     },
@@ -979,7 +974,10 @@ function ReaderHighlight:exportToDocument(page, item)
 end
 
 function ReaderHighlight:addNote()
-    self:handleEvent(Event:new("addNote"))
+    self:handleEvent(Event:new("AddNote"))
+    local page, index = self:saveHighlight()
+    self:editHighlight(page, index)
+    UIManager:close(self.edit_highlight_dialog)
     logger.dbg("add Note")
 end
 
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -721,9 +721,13 @@ function ReaderHighlight:onHoldRelease()
                     },
                     {
                         text = _("Add Note"),
-                        enabled = false,
+                        enabled = true,
                         callback = function()
-                            self:addNote()
+                            local page, index = self:saveHighlight()
+                            if page and index then -- saveHighlight can return nil, nil
+                                self:editHighlight(page, index)
+                                UIManager:close(self.edit_highlight_dialog)
+                            end
                             self:onClose()
                         end,
                     },
@@ -946,6 +950,7 @@ function ReaderHighlight:saveHighlight()
         if self.selected_text.pboxes then
             self:exportToDocument(page, hl_item)
         end
+        return page, #self.view.highlight.saved[page]
     end
 end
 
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -721,9 +721,13 @@ function ReaderHighlight:onHoldRelease()
                     },
                     {
                         text = _("Add Note"),
-                        enabled = false,
+                        enabled = true,
                         callback = function()
-                            self:addNote()
```
Frenzie: This leaves an (even more) orphaned function around. It's probably cleaner to write the below in that function instead.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -721,9 +721,13 @@ function ReaderHighlight:onHoldRelease()
...
+                            local page, index = self:saveHighlight()
+                            if page and index then -- saveHighlight can return nil, nil
+                                self:editHighlight(page, index)
+                                UIManager:close(self.edit_highlight_dialog)
+                            end
```
Frenzie: You might also want to add an `else` like:
```lua
    text = _("Something about why saveHighlight can return nil, nil."),
    show_icon = false,
    timeout = 3,
})
```

Galunid: This comment is mostly for developers, this can only happen when text is not selected, e.g. 
```lua
if self.hold_pos and self.selected_text and self.selected_text.pos0 and self.selected_text.pos1 then
```
is false. I'm not sure what can cause this, so I left it in case this situation occurs.

frontend/apps/reader/modules/readerhighlight.lua
Galunid: This function currently looks like this:
```lua
function ReaderHighlight:addNote()
    self:handleEvent(Event:new("addNote"))
    logger.dbg("add Note")
end
```
`ReaderHighlight:editHighlight` doesn't emit any event. Should I add "addNote" event there, or should I remove it? I'm not sure if there are any handlers for it, I'll check it when I get access to my VM later to make sure.

Frenzie: The statistics plugin responds to it.

Frenzie: When editing you're not adding anything, so I don't quite see why you'd want to add an event there?

frontend/apps/reader/modules/readerhighlight.lua
Frenzie: And this isn't something you can check in the enabled_func?

frontend/apps/reader/modules/readerhighlight.lua
poire-z: So, this one should be fixed to be `self.ui:handleEvent(Event:new("AddNote"))`, as there is a `ReaderStatistics:onAddNote()`
(But it's just a counter, and there is no onDelNote - but I guess that doesn't really matter as notes are just highlights.)

(`self:handleEvent(Event:new("addNote")`) is currently a no-op, as there is no `ReaderHighlight:onaddNote` - which wouldn't make any sense anyway).

I thought I would agree of letting the callback as it is, and moving that bit of code in the `ReaderHighlight:addNote()` - so we have more room there if we want to do more.
But I see most of the other callbacks that do some UIManager:show() have that code in the callback itself - other ReaderHighlight methods used by callbacks are mostly non- UIManager stuff.
So, I guess I wouldn't mind either way.

(just remove the `enabled = true` as it's the default when unspecified).

frontend/apps/reader/modules/readerhighlight.lua
Galunid: This is already being checked in `ReaderHighlight:saveHighlight`, I don't think there is any point checking it again in enabled_func. In fact, I think that dialog won't be shown if this is false, although I haven't checked it yet. This is not checked for other buttons too.

Frenzie: Then the question automatically becomes the reverse. ;-)

Galunid: I'm pretty sure in this context those additional checks were not necessary, so I've removed them.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -974,7 +974,10 @@ function ReaderHighlight:exportToDocument(page, item)
 end
 
 function ReaderHighlight:addNote()
-    self:handleEvent(Event:new("addNote"))
+    self:handleEvent(Event:new("AddNote"))
```
poire-z: self **.ui** :handleEvent - as done at the top of ReaderHighlight:saveHighlight()
We want `self.ui` to handleEvent, as it will propagate it to all the things registered to self.ui, which includes the plugins.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -974,7 +974,10 @@ function ReaderHighlight:exportToDocument(page, item)
 end
 
 function ReaderHighlight:addNote()
-    self:handleEvent(Event:new("addNote"))
+    self.ui:handleEvent(Event:new("AddNote"))
```
Frenzie: This whole business of having these events at the top doesn't seem to make much sense to me. Surely the idea is more along the lines of "hey guys, we've added a note, go wild", and not so much "hey guys, we're going to add a note but you can't do anything about it."

But I actually started writing this comment to say that you should either remove the logger.dbg call or keep 'em together.

poire-z:
None, looks good to me.

Issue #4752, opened 2019-03-07
Kobo Aura H2O rotation issue


Frenzie:
* KOReader version: since #4637
* Device: Kobo Aura H2O, should also be the case on Kobo Aura HD and potentially others with the same display (H2O2?)

#### Issue
Something's off with the fbink update rotation and rotation in KSM on exit. Discussion on Gitter around https://gitter.im/koreader/koreader?at=5c815e368f294b134a04908a and higher.

The update notification looks like this:
![screenshot_2019-03-07_20-23-34](https://user-images.githubusercontent.com/202757/53983448-771db680-4117-11e9-8161-886828bbf200.png)

Frenzie:
Video illustrating the problem:

[2019-03-07 21.03.11.mp4.zip](https://github.com/koreader/koreader/files/2943075/2019-03-07.21.03.11.mp4.zip)

On KSM8 there's also some turning upside down in the menus on exit, but KSM9 is fine so we can probably just ignore that.

NiLuJe:
Yeah, that was an oversight on my part ;). I'd blissfully forgotten about the quirkiness of the rotation pickel sets up on those devices, because it behaves differently on the Forma.

I just need to expand the current checks for that state to 8bpp ;).

poire-z:
Just for info: no issue at all for me with koreader 8bpp with FW 3.19 and good old KSM 07.
Btw, that may concern only users of older devices that were supported by KSM 07 - as newer devices need KSM 09 - but KSM07 has the benefit (for some) that it's all on the /mnt/onboard partition unlike newer ones. So, I'd rather like koreader to stay compatible with older KSM :)

There is just one thing, possibly related to input/gestures changes - but probably not: the issue I described at https://github.com/koreader/koreader/issues/4431#issuecomment-455847181 has happen to me 5 times over last week, when it happened previously only twice a month. But it could be just very well that I multiswipe too much now :)  and the original problem, possibly hardware related, shows itself more often with more touch events...
Still unsure what's the cause - but unlike @ptrm issue with his Clara describer on gitter, when my touch layer is unresponsive, I can still suspend with the power button - wait 20 seconds, resume, and touch is back.

NiLuJe:
@poire-z : You're using a Glo HD, right?

Could I get an excerpt of your KOReader logs, just to double-check which rotation it's in, so that I try not to break it when I fix the H2O issue? ;).

NiLuJe:
[...]

PR #4779, opened 2019-03-12
Fix fb rotation on the Forma w/ KSM

NiLuJe:
KSM is currently emulating a legacy pickel rotation (UR, 0), instead of what both pickel & Nickel do (CCW, 3) on the Forma.

Besides the bad interaction with FBInk described in #4752, I found out that the eInk driver was doing very weird things with A2 updates when UR @ 8bpp...

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -221,7 +221,7 @@ fi
 if [ "${PRODUCT}" = "frost" ]; then
     # Only needed for KSM, pickel -> Nickel will restore its own rota properly
     if [ "${FROM_NICKEL}" != "true" ]; then
-        echo "${ORIG_FB_ROTA}" >/sys/class/graphics/fb0/rotate
+        echo "$(( (4 - ORIG_FB_ROTA) % 4 ))" >/sys/class/graphics/fb0/rotate
         # Sleep a bit, for good measure
         usleep 150000
     fi
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -221,7 +221,7 @@ fi
 if [ "${PRODUCT}" = "frost" ]; then
     # Only needed for KSM, pickel -> Nickel will restore its own rota properly
     if [ "${FROM_NICKEL}" != "true" ]; then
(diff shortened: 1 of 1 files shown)


NiLuJe:
Pinging @ptrm, as I have no way to check if behaves as intended on KSM ;).

NiLuJe:
The most user-visible change as far as A2 quirks are concerned would be that FORCE_MONOCHROME will actually take now, so you won't get crappy fringes around icons when they're inverted.

platform/kobo/koreader.sh
```diff
@@ -217,6 +217,17 @@ if [ -n "${ORIG_FB_BPP}" ]; then
     ./fbdepth -d "${ORIG_FB_BPP}" >>crash.log 2>&1
 fi
 
+# Restore original fb rotation on the Forma, for KSM
+if [ "${PRODUCT}" = "frost" ]; then
+    # Only needed for KSM, pickel -> Nickel will restore its own rota properly
+    if [ "${FROM_NICKEL}" != "true" ]; then
+        # The Forma kernel inverts odd rotation constants, counteract that, àla Plato because one-liner ;p.
```
Frenzie: I'm oddly bothered by that à la without a space. :-P

NiLuJe: That's my own bad habit, AFAICT ;).

ptrm:
In general, the new `koreader.sh` works, and regardless of later orientation the touches correspond with the screen reaction. Did not have time to downgrade back to previous nightly to check if FBInk status overlay is in place, probably it will turn out tomorrow ;)

NiLuJe:
@ptrm: Thanks! I assume KSM's rotated properly on exit?

ptrm:
>I assume KSM's still rotated properly when KOReader exits?

That's right, and touch response is proper as well

NiLuJe:
Off we go, then, thanks!

ptrm:
FTR, FBInk was oriented in portrait during OTA update syncing and unpacking today

NiLuJe:
\o/

PR #4782, opened 2019-03-13
cre landscape view: adds 1 page / 2 pages toggable 

### LINKED ISSUE #4778: "No books were found" when 'Extract and cache book information'
* KOReader version: v2019.02-180-g2536cc2_2019-03-08/ 
* Device: KA1

#### Issue
'Extract and cache book information'-feature doesn't find your books if the following procedure is performed.
#### Steps to reproduce
1. Ensure that you have books in History page
2. Delete cache database and restart KOReader
3. Choose display mode for History page and file manager (something other than classic).
4. Open History page and then close it again.
5. Go to your Home folder (or where your library are), which should have subfolders with books
6. Start 'Extract and cache book information metadata': Continue, Here and under, Don't refresh/Refresh, Don't prune/Prune
7. This produce a "No books were found" message. So it doesn't find books in subfolders, _but_ it seems like it won't even find books that are placed in the folder you perform the 'Extract and cache...', (of course unless they are on page 1 in that folder) as long the first page in that folder contains at least one folder.
8. If you want it to find your books now you can either restart KOReader, or jump into a subfolder and back, and then start 'Extract and cache..." again. If you open History page in between (after restart or jump/jump back) the problem continues.

poire-z:
Fix a few small issues, and introduce a new toggle in the Orientation bottom config menu:

<kbd>2pagesb</kbd>

Note: 2-pages view was the default for landscape for some users depending on their device resolution and selected dpi, and was not toggable. So I guess a few people will discover this existing nice feature of crengine, that I find really cute.

See individual commit messages for details.

Added fixes for CoverBrowser issues detailed in #4778. Closes #4778.

Also:
bump crengine https://github.com/koreader/crengine/pull/270
- docToWindowPoint(): adds fitToPage parameter
- 2-pages mode: add and fix some functions
- 2-pages mode: tweak middle margin sizing
- Text rendering: fix words stuck when hyphenating on italic

bump base https://github.com/koreader/koreader-base/pull/850
- cre.cpp: add a few functions useful with 2-pages mode
- cre.cpp: fix docToWindowRect() to get truncated segments for lines split onto 2 pages, for highlights and links
- Bump CMake minimum version to 3.5.1 https://github.com/koreader/koreader-base/pull/849

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -207,7 +207,7 @@ function ReaderHighlight:onTapXPointerSavedHighlight(ges)
     -- showing menu...). We might want to cache these boxes per page (and
     -- clear that cache when page layout change or highlights are added
     -- or removed).
-    local cur_page, cur_page2, cur_scroll_top, cur_scroll_bottom
+    local cur_view_top, cur_view_bottom
     local pos = self.view:screenToPageTransform(ges.pos)
     for page, _ in pairs(self.view.highlight.saved) do
         local items = self.view.highlight.saved[page]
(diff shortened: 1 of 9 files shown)


poire-z:
2 notes:
- if people are going to play with orientation and margins like I did in https://github.com/koreader/koreader/issues/4752#issuecomment-471215124 - they may get cpuloops and need hard reboot and lose their settings - so we may want to set `jit.opt.start("loopunroll=45")` on Kobo quick :) Pinging @NiLuJe 

- There is something with using buttonprogress for margins (#4702), that often causes a double painting and refreshing of the page (1st paint after margin changed, 2nd paint when repositionning to previous xpointer, so possible on another page).
I still have to fix that, but it's a bit twisted, as this buttonprogress is used for contrast and do things early unlike other bottom widgets - and when I began fixing that on cre side, none were working with pdf :) so...

poire-z:
Ok. But you could have waited for merging https://github.com/koreader/koreader-base/pull/850 as I wrote _no hurry_. I hurried finishing the frontend side this morning so it could be bumped and you could go on with your CMake experimentations :|

But ok, no hurry, this can wait.

Frenzie:
@poire-z I didn't mean for you to hurry, my apologies. You can now rebase on top of master and it should work. :-)

poire-z:
Thanks, that couple of days went fast :)
(I just dont like these interleaving bumps - been painful in the past - and as I don't get any of your CMake stuff, I thought you would possibly need more iterations on both frontend and base and I didn't want to be in the middle of them :)

poire-z:
So, back to language business: `Visible pages: one | two` or `Something else: one page | two pages` ?

Frenzie:
[This one](https://github.com/koreader/koreader-base/pull/851) is just a thought I had last night based on an off-hand remark by @NiLuJe on Gitter yesterday, so you could call it experimental. (But I tried the build on my PC, my H2O and the Android emulator, so I think it's fine.) The very module I added in that PR was actually added to CMake 3.11 itself as FetchContent.

But merely bumping the minimum version requirement was fully prepared around the time I opened that other PR (i.e., October 2018). I just had to make/push a new Docker image with both Python and SDL for CI purposes, but the nightly builds have been running on Ubuntu 16.04 since October. And of course Travis in base was lagging behind as well.

Like I said in <https://github.com/koreader/koreader-base/pull/849>, CMake 3.3 "is slightly less annoyingly stupid about finding libraries," implicitly meaning there's basically no immediate effect other than being able to remove a few workarounds related to that kind of thing. ;-)

> So, back to language business: Visible pages: one | two or Something else: one page | two pages ?

How about simply `Two pages: on/off`?

Frenzie:
PS

Why to switch to Meson ;-)
https://medium.com/@germandiagogomez/getting-started-with-meson-part-4-8bceec6149e1

NiLuJe:
@poire-z : I was giving it a few days to confirm that it worked around the issue in your case ;). I'll take care of it :).

poire-z:
> Why to switch to Meson ;-)

I see, more readable. Well, thank you both for taking care of all that ugly build side :)

> How about simply Two pages: on/off?

Could work, but inverted to avoid some kind of ugly X (crossed checked toggles) in portrait mode:

<kbd>image</kbd>

<kbd>image</kbd>

But I dunno, in the same way we avoid having Yes|No in ConfirmBox, I like to have named buttons, when we can.

poire-z:
OK, let's go with `Two pages: off | on`.
Added fixes for CoverBrowser issues detailed in #4778. Closes #4778.

NiLuJe:
I'd be partial to something like "Split View" or "Two Columns", but that's just be ;).

NiLuJe:
Or, perhaps more adapted to the specific context: "double spread"?

poire-z:
`Two Columns` would be clear for most I guess, but "double spread" doesn't talk to me :) and "Split view" does not announce if it's vertical or horizontal :)

NiLuJe:
Yeah, there's usually a `Left|Right` following `Split View` (or a `Vertical` preceding it) ;).

poire-z:
Somehow "Two Columns" is common and I quite like it. But "Two Pages" (fixed capitalization) says more what it is: page numbers will skip by 2 when taping or swiping, and the middle margin, when large, feels more like a book than real 2 columns in scientific papers or newspapers.

NiLuJe:
Stupid question: Would bumping base without having this bit in break something?

poire-z:
I guess no. But let me not have us wonder about that :)

Frenzie:
I wouldn't call it a double spread. I can quickly pull out an example of one if you like. :-P

You could call it `dual pages`.

> But I dunno, in the same way we avoid having Yes|No in ConfirmBox, I like to have named buttons, when we can (althought off|on makes it easier for translators :)

True, but it's a contextual thing. On/off here is a simple checkbox without any potential for ambiguity.

Frenzie:
Here, I'd say something like this is a proper dual spread. ;-)

poire-z:
> You could call it `Dual Pages`

I kinda like that. It's less abrupt than "Two Pages", sounds more like a feature than can be enabled/disable with some off|on - possibly a bit cryptic but would make sense once seen.

Frenzie:
@poire-z If you'd like to go with that, please push it to master before tomorrow's push to Transifex. ;-)

I prefer `dual pages` too but I don't think there's anything wrong with `two pages`.

I don't have Word at hand atm, but Writer calls it "book view" for the style discussed here and "multiple-page view" for whichever number will fit.

poire-z:
OK, adding it now.

PR #4789, opened 2019-03-14
Various blitting related cleanups

### LINKED ISSUE #4774: Screensaver: Last book's cover renders differently from original image
* KOReader version: 2019.03.1
* Device: Kobo Forma

#### Issue
Since installing the March stable update, whenever I enter screensaver mode and choose to display the last book cover, the image shown is a bit different from the original one. Please see images below. Nothing major, but curious to see if I'm doing anything wrong.

"White background behind message and images" is disabled, and does not seem to have any effect (likely because I chose to "Stretch covers and images to fit screen").

I believe the problem first emerged in a February nightly. I went back to the stable version and everything was fine then. Thank you for any thoughts!

#### Steps to reproduce
Enable screensaver and select last book cover.

Original

Screensaver

NiLuJe:
* Fix the "Enable debug logging" checkbox so that it properly disables "Enable verbose debug logging" when it gets disabled
* Avoid asking ImageWidget for alpha-blending when it's not useful
* Make ImageWidget's alpha-blending code-path double-check that alpha-blending actually is needed, and avoid it if possible
* In the same vein, only do alpha-blending in textboxwidget when absolutely necessary
* Prefer color constants over the gray() method, ensuring that they're part of the eInk palette

Depends on https://github.com/koreader/koreader-base/pull/853

Fix #4774


poire-z:
Been testing this+base on my Kobo, no issue so far.

Except for some Wikipedia images, fine in the textboxwidget, but when you hold on some images to display them in ImageViewer, they now have a black background (they have a white background with 2019-02-25):

https://fr.wikipedia.org/wiki/Allemagne_de_l%27Ouest First map in the top right table ("Carte de localisation") (Or just do a lookup for "Allemagne", it's the image displayed on small results for this article.)
https://en.wikipedia.org/wiki/China The globe in the top right table (not displayed in the textboxwidget, but it is the first image when Save as EPUB: hold and fetched from crengine, displayed in ImageViewer: black background

NiLuJe:
Yeah, that makes sense, I was wondering about that alpha in image viewer...
;).

frontend/ui/elements/screensaver_menu.lua
```diff
@@ -137,6 +138,15 @@ return {
                 checked_func = whiteBackground,
                 callback = function()
                     G_reader_settings:saveSetting("screensaver_white_background", not whiteBackground())
+                    G_reader_settings:flipFalse("screensaver_no_background")
+                end,
+            },
+            {
+                text = _("Leave background as-is behind message and images"),
```
Frenzie: What does that mean precisely? Alpha-channel blending might have weird results or something?

NiLuJe: It just means that alpha-blending will actually be useful, since we won't be clearing to white or black, or drawing any kind of background behind the image.

So you can effectively use transparent images and have the transparency stay useful.

Prepare to be amazed by my test image... :D

zzz

NiLuJe: Hmm, still something a bit off with alpha-blending, it looks darker than it should (and there's a bit of solarization on the edge of the lighter gray...).

NiLuJe: i.e., original overlay:
forma-sleepy

And essentially the same as earlier but rendered w/ FBInk:
zzz_fbink

NiLuJe: Note to self: try with a G8A input image, to rule out a grayscaling pass somewhere.

poire-z:
Testing this PR and some others on my Kobo.
A small issue: in the bottom config panel, font size section: the selected font size is underlined black - but all others are now underlined gray (instead of underline white / not underlined).
Probably something not obvious in https://github.com/koreader/koreader/pull/4789/commits/89c778d937847d791dc7a601fa85631795e96e82#diff-aedbf40026b37bd6245e58e96bbed9a2

poire-z:
(Wikipedia imageviewer black background solved.)

poire-z:
Probably some missing parens in stuff like:
```diff
- color = d == current_item and Blitbuffer.gray(enabled and 1.0 or 0.5) or Blitbuffer.COLOR_WHITE,
+ color = d == current_item and enabled and Blitbuffer.COLOR_BLACK or Blitbuffer.COLOR_DARK_GRAY or Blitbuffer.COLOR_WHITE,
```

poire-z:
Another small issue, exagerated here by using `COLOR_LIGHT_GRAY = Color8(0xCC)` in both widgets:
<kbd>image</kbd>

The button progress in "thin grey style" use some BB invert, so it worked when the gray was 0x80 :)
Not as well with 0x88 where there ButtonProgress is grayer than the toggleswith.

NiLuJe:
@poire-z: Yeah, I though the lack of parenthesis would blow up at luacheck time, to be perfectly honest, so when it went fine I quickly forgot about it. Will fix.

Not quite sure I get the progress thingy (and I was away when the switch to gray was implemented, IIRC, so I don't have much context). Is this "thin grey style" togglable somewhere?

poire-z:
No will fix needed, did fix in #4797 :)
(The thin grey line was done in #4700.)

NiLuJe:
Yeah, just caught up with that, thanks! ;).

frontend/ui/elements/screensaver_menu.lua
NiLuJe: After looking at this a bit,

this looks suspiciously like the PNG matte background bleeding through (also visible when both MuPDF & LodePNG flatten the image), so, meh.

NiLuJe: Couldn't manage to find much info about how to set that matte color (short of old PhotoShop versions), so, let's say it's a weird PNG thing.

NiLuJe: Haha! Figured it out.

MuPDF stores pixmaps in [premultiplied alpha](https://ghostscript.com/~robin/mupdf_explored.pdf#subsection.10.3.2), while our blending code expects straight alpha.

I'll cobble something together...

PR #4790, opened 2019-03-14
[Gesture manager] Add toggle default highlight action

Galunid:
It should do what I discussed here: https://github.com/koreader/koreader/issues/4727#issuecomment-472630607
It works, but I'm not sure changes to `readergesture.lua` are following good practices.

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -843,6 +843,29 @@ function ReaderHighlight:onHoldRelease()
     return true
 end
 
+function ReaderHighlight:onToggleHighlightAction()
+    local actions = {
+        highlight = "translate",
+        translate = "wikipedia",
+        wikipedia = nil
+    }
+    local current_action = G_reader_settings:readSetting("default_highlight_action")
+    if not current_action then
+        G_reader_settings:saveSetting("default_highlight_action", "highlight")
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action is now: highlight"),
+            timeout = 1,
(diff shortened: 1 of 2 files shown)


Frenzie:
I suppose you closed it because you accidentally included the wifi commit? In the long term it might be worthwhile to familiarize yourself with rebasing commits.

What I do goes something like this:

```sh
git checkout bad-branch
git checkout -b bad-branch-bck # in case you mess up you can abort the rebase, but just good practice
git rebase -i HEAD~3 # 3 is the number of commits to include in interactive rebase
```

Then you comment out the commits you don't want, and rearrange and squash the rest so you get a nice PR.

That's just some background info. We can also squash commits through the GitHub web GUI, so it's not as relevant to know as it used to be. But personally I am quite fond of Git's distributed nature. :-)

Like I said the actual code looks fine to me, but being more familiar with the highlighting stuff in general @poire-z might have more to say about it.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -843,6 +843,29 @@ function ReaderHighlight:onHoldRelease()
...
+    local actions = {
+        highlight = "translate",
+        translate = "wikipedia",
+        wikipedia = nil
+    }
+    local current_action = G_reader_settings:readSetting("default_highlight_action")
+    if not current_action then
+        G_reader_settings:saveSetting("default_highlight_action", "highlight")
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action is now: highlight"),
```
Frenzie: I'd use `Default highlight action changed to`.

PR #4791, opened 2019-03-14
[Gesture manager] Add toggle default highlight action

Galunid:
It should do what I discussed here: #4727 (comment)
@Frenzie Thanks for the instruction in the other PR ;-)
@poire-z Could you take a look?

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -843,6 +843,29 @@ function ReaderHighlight:onHoldRelease()
     return true
 end
 
+function ReaderHighlight:onToggleHighlightAction()
+    local actions = {
+        highlight = "translate",
+        translate = "wikipedia",
+        wikipedia = nil
+    }
+    local current_action = G_reader_settings:readSetting("default_highlight_action")
+    if not current_action then
+        G_reader_settings:saveSetting("default_highlight_action", "highlight")
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action is now: highlight"),
+            timeout = 1,
+        })
+    else
+        local next_action = actions[current_action]
+        G_reader_settings:saveSetting("default_highlight_action", next_action)
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action is now: " .. (next_action or "default")),
+            timeout = 1,
+        })
+    end
+end
+
 function ReaderHighlight:highlightFromHoldPos()
     if self.hold_pos then
         if not self.selected_text then
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -843,8 +843,8 @@ function ReaderHighlight:onHoldRelease()
     return true
 end
 
-function ReaderHighlight:onToggleHighlightAction()
-    local actions = {
+function ReaderHighlight:onCycleHighlightAction()
+    local next_actions = {
         highlight = "translate",
         translate = "wikipedia",
         wikipedia = nil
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -843,6 +843,29 @@ function ReaderHighlight:onHoldRelease()
     return true
 end
 
+function ReaderHighlight:onToggleHighlightAction()
+    local actions = {
+        highlight = "translate",
```
poire-z: Looks alright, just some renaming suggestions:
May be just rename that table to `next_actions` (plural), so you have below `next_action = next_actions[current_action]`
And the method `onCycleHighlightAction()` as we're not really toggling but rotating.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -843,6 +843,29 @@ function ReaderHighlight:onHoldRelease()
...
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action changed to highlight"),
+            timeout = 1,
+        })
+    else
+        local next_action = actions[current_action]
+        G_reader_settings:saveSetting("default_highlight_action", next_action)
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action changed to " .. (next_action or "default")),
+            timeout = 1,
```
poire-z: And you might want to see if you'd not prefer `Notification` instead of `InfoMessage`, which makes thing less busy with rereshes. It's possible that this text is too long for a Notification, that may depends on screen dpi (haven't checked if long text is wrapped).

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -843,6 +843,29 @@ function ReaderHighlight:onHoldRelease()
...
+        G_reader_settings:saveSetting("default_highlight_action", "highlight")
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action changed to highlight"),
+            timeout = 1,
+        })
+    else
+        local next_action = next_actions[current_action]
+        G_reader_settings:saveSetting("default_highlight_action", next_action)
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action changed to highlight " .. (next_action or "default")),
```
poire-z: Your last commit added "Default highlight action changed to **highlight**", which feels like cut & paste gone wrong :)

Galunid: Most likely ;)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -843,6 +843,29 @@ function ReaderHighlight:onHoldRelease()
...
+        G_reader_settings:saveSetting("default_highlight_action", "highlight")
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action changed to highlight"),
+            timeout = 1,
+        })
+    else
+        local next_action = next_actions[current_action]
+        G_reader_settings:saveSetting("default_highlight_action", next_action)
+        UIManager:show(InfoMessage:new{
+            text = _("Default highlight action changed to " .. (next_action or "default")),
```
Frenzie: You can't use concatenation. Use T(_("%1"), bla)

frontend/apps/reader/modules/readerhighlight.lua
Galunid: Notifications seem to work nicely and don't depend on DPI, at least on emulator, I'll test it on my kindle later and possibly amend this commit. For now everything seems to work without any problems.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -843,6 +844,29 @@ function ReaderHighlight:onHoldRelease()
...
+        G_reader_settings:saveSetting("default_highlight_action", "highlight")
+        UIManager:show(Notification:new{
+            text = _("Default highlight action changed to \"highlight\""),
+            timeout = 1,
+        })
+    else
+        local next_action = next_actions[current_action]
+        G_reader_settings:saveSetting("default_highlight_action", next_action)
+        UIManager:show(Notification:new{
+            text = T(_("Default highlight action changed to \"%1\""), (next_action or "default")),
```
Frenzie: The house style is single quotes. I should write that down somewhere more formally.

```suggestion
            text = T(_("Default highlight action changed to '%1'."), (next_action or "default")),
```

Frenzie:
Thanks!

PR #4792, opened 2019-03-14
CoverBrowser: speedup "View full size cover"

poire-z:
It was doing a full document load to get the cover. It now does the faster "only metadata" load.
Also move the trick of setting a default font in CreDocument, so that all callers of `document:loadDocument(false)` benefit from it.
Prevent crash when no cover image is available (even if the cache says it has one, the file may have been updated and doesn't have it anymore).

Also:
Gesture manager: add actions: Book cover, Book description - as requested in https://github.com/koreader/koreader/issues/4727#issuecomment-472739389

### DIFF
--- a/plugins/coverbrowser.koplugin/covermenu.lua
+++ b/plugins/coverbrowser.koplugin/covermenu.lua
@@ -1,6 +1,7 @@
 local DocumentRegistry = require("document/documentregistry")
 local FileManagerBookInfo = require("apps/filemanager/filemanagerbookinfo")
 local ImageViewer = require("ui/widget/imageviewer")
+local InfoMessage = require("ui/widget/infomessage")
 local Menu = require("ui/widget/menu")
 local TextViewer = require("ui/widget/textviewer")
 local UIManager = require("ui/uimanager")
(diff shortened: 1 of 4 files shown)


NiLuJe:
I'm assuming the ScreenSaver code already uses this metadata only fats path, right? ;).

poire-z:
Yes, it does that all fine, including falling back without crash if no cover:
https://github.com/koreader/koreader/blob/f637555d93a16d77b78140f3c18a16a2bcdb6d13/frontend/ui/screensaver.lua#L249-L258

plugins/coverbrowser.koplugin/covermenu.lua
```diff
@@ -232,13 +232,22 @@ function CoverMenu:updateItems(select_number)
...
+                                if cover_bb then
+                                    local imgviewer = ImageViewer:new{
+                                        image = cover_bb,
+                                        with_title_bar = false,
+                                        fullscreen = true,
+                                    }
+                                    UIManager:show(imgviewer)
+                                else
+                                    UIManager:show(InfoMessage:new{
+                                        text = _("No cover image available"),
```
Frenzie:
```suggestion
                                        text = _("No cover image available."),
```

plugins/coverbrowser.koplugin/covermenu.lua
```diff
@@ -352,13 +361,22 @@ function CoverMenu:onHistoryMenuHold(item)
...
+                    if cover_bb then
+                        local imgviewer = ImageViewer:new{
+                            image = cover_bb,
+                            with_title_bar = false,
+                            fullscreen = true,
+                        }
+                        UIManager:show(imgviewer)
+                    else
+                        UIManager:show(InfoMessage:new{
+                            text = _("No cover image available"),
```
Frenzie:
```suggestion
                            text = _("No cover image available."),
```

PR #4793, opened 2019-03-14
ConfigDialog: fix ButtonProgressWidget and some refreshes

poire-z:
This internal ButtonProgressWidget widget was behaving differently from all others (OptionTextItem, OptionIconItem and ToggleSwitch) by duplicating some code from ConfigDialog:onConfigChoose() instead of calling it directly. So, similar to:
https://github.com/koreader/koreader/blob/f637555d93a16d77b78140f3c18a16a2bcdb6d13/frontend/ui/widget/toggleswitch.lua#L191-L208

While making it similar to others, I noticed that onConfigChoose() did a full repaint, which was necessary for some settings to be applied (ie: Contrast) - or may be not, see below, but I don't know what did that then...
On CreDocument, this full repaint may cause some double drawing on config changes (ie: Margins, since #4691, drawing once after margin changes,and then re-positionning to previous xpointer).
So, make the need for full repaint a condition on KoptOptions.

These changes seem to make PDF and CRE settings behave all correctly. I'll still have to check on my Kobo.

But I'm still puzzled, mostly about how it works/worked with PDF.
Previously, we had a call in that ButtonProgressWidget: `UIManager:setDirty("all")`. It looks to me this is actually a no-op from UIManager point of view (where there is no mode provided like "full" or "ui") ! Or may be not in some situations?
So I don't know why removing it, or replacing it with a more localized `UIManager:setDirty(self.config, function() return "ui", widget.diment)` made contrast no more applied...

I'm pretty confident these change are fine for CreDocument, a bit less for PDF.
Changing contrast leads to the following log:
```
adjusted ges: tap
setDirty via a func from widget table: 0x4085d3f8
setDirty via a func from widget table: 0x4085d3f8
painting widget: table: 0x40dab098
_refresh: Enqueued fast update for region 163 514 424 32
update_mode: Update refresh mode fast to ui
_refresh: Enqueued ui update for region 163 514 424 32
setDirty via a func from widget table: 0x4085d3f8
setDirty via a func from widget table: 0x4085d3f8
_refresh: Enqueued partial update for region 0 0 600 610
setDirty partial from widget all w/ NO region
painting widget: ReaderUI
loading koptcontext from ./cache/8d22901f6040cb5e7cf1b7ffa9b3a684
painting widget: table: 0x40dab098
_refresh: Enqueued partial update for region 0 0 600 610
_refresh: Enqueued partial update for region 0 0 600 610
```
Previously:
```
adjusted ges: tap
AutoSuspend: onInputEvent
setDirty via a func from widget table: 0x41eb8bc8
painting widget: table: 0x40dd5650
_refresh: Enqueued ui update for region 163 514 424 32
setDirty nil from widget all w/ NO region
painting widget: ReaderUI
loading koptcontext from ./cache/f5078849f8b7e252f3d07136207b5e8b
painting widget: table: 0x40dd5650
no refresh got enqueued. Will do a partial full screen refresh, which might be inefficient
_refresh: Enqueued partial update for region 0 0 600 610
```
So, 3 instead of 1 "Enqueued partial update for region 0 0 600 610" - dunno if that means 3 EInk screen refreshes...

Pinging @NiLuJe for opinion.

### DIFF
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -643,6 +643,15 @@ function ReaderRolling:onUpdatePos()
         -- we have set above) to avoid multiple refreshes.
         return true
     end
+    -- Calling this now ensures the re-rendering is done by crengine
+    -- so the delayed updatePos() has good info and can reposition
+    -- the previous xpointer accurately:
+    self.ui.document:getCurrentPos()
+    -- Otherwise, _readMetadata() would do that, but the positionning
+    -- would not work as expected, for some reason (it worked
+    -- previously because of some bad setDirty() in ConfigDialog widgets
+    -- that were triggering a full repaint of crengine (so, the needed
+    -- rerendering) before updatePos() is called.
     UIManager:scheduleIn(0.1, function () self:updatePos() end)
     return true
 end
--- a/frontend/ui/data/koptoptions.lua
+++ b/frontend/ui/data/koptoptions.lua
@@ -6,6 +6,7 @@ local Screen = Device.screen
 
 local KoptOptions = {
     prefix = 'kopt',
+    needs_redraw_on_change = true,
     {
         icon = "resources/icons/appbar.transform.rotate.right.large.png",
         options = {
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -499,7 +499,8 @@ function ConfigOption:init()
                 local max_buttonprogress_width = option_widget_width
                 local buttonprogress_width = self.options[c].width and Screen:scaleBySize(self.options[c].width)
                                                 or max_buttonprogress_width
-                local switch = ButtonProgressWidget:new{
+                local switch
+                switch = ButtonProgressWidget:new{
                     width = math.min(max_buttonprogress_width, buttonprogress_width),
                     height = option_height,
                     padding = 0,
@@ -509,10 +510,10 @@ function ConfigOption:init()
                     num_buttons = #self.options[c].values,
                     position = self.options[c].default_pos,
                     callback = function(arg)
-                        UIManager:tickAfterNext(function()
-                            self.config:onConfigChoice(self.options[c].name, self.options[c].values[arg])
-                            self.config:onConfigEvent(self.options[c].event, self.options[c].args[arg])
-                            UIManager:setDirty("all")
+                        self.config:onConfigChoose(self.options[c].values, self.options[c].name,
+                                self.options[c].event, self.options[c].args, self.options[c].events, arg)
+                        UIManager:setDirty(self.config, function()
+                            return "fast", switch.dimen
                         end)
                     end,
                     hold_callback = function(arg)
@@ -840,8 +841,23 @@ function ConfigDialog:onConfigChoose(values, name, event, args, events, position
         if events then
             self:onConfigEvents(events, position)
         end
+        -- Even if each toggle refreshes itself when toggled, we still
+        -- need to update and repaint the whole config panel, as other
+        -- toggles may have their state (enabled/disabled) modified
+        -- after this toggle update.
         self:update()
-        UIManager:setDirty("all")
+        if self.config_options.needs_redraw_on_change then
+            -- Some Kopt document event handlers just save their setting,
+            -- and need a full repaint for kopt to load these settings,
+            -- notice the change, and redraw the document
+            UIManager:setDirty("all", "partial")
+        else
+            -- CreDocument event handlers do their own refresh:
+            -- we can just redraw our frame
+            UIManager:setDirty(self, function()
+                return "ui", self.dialog_frame.dimen
+            end)
+        end
     end)
 end
 


NiLuJe:
> Previously, we had a call in that ButtonProgressWidget: UIManager:setDirty("all"). It looks to me this is actually a no-op from UIManager point of view (where there is no mode provided like "full" or "ui") ! Or may be not in some situations?

`UIManager:setDirty("all")` essentially flags *every* top-level window as dirty, essentially forcing a repaint of the full stack (i.e., each toplevel widget's paintTo will be called).

NiLuJe:
And it's indeed generally overkill, if you want to enforce an *eInk* refresh without *repainting* anything, you're better off doing a `UIManager:setDirty(nil, ...)` instead.

poire-z:
> UIManager:setDirty("all") essentially flags every top-level window as dirty

But then:
https://github.com/koreader/koreader/blob/f637555d93a16d77b78140f3c18a16a2bcdb6d13/frontend/ui/uimanager.lua#L539
with `refreshtype / mode = nil`, this would return early:
https://github.com/koreader/koreader/blob/f637555d93a16d77b78140f3c18a16a2bcdb6d13/frontend/ui/uimanager.lua#L778-L786
so, not doing any refresh?

Or may be this call without mode would just flag the widgets as dirty, exiting there early without any refresh. And a later setDirty (possibly smaller and localized) would see all the previously set dirty widgets, and deal with them at this point? (which could explain the async'ness I see and can't explain :)

NiLuJe:
@poire-z : Yep, I'd forgotten that extra-quirkiness of not passing a refreshfunc at all in this case, but, yeah, you're right ;).

I tried to clean most of those up, because they're indeed unintuitive and weird and generally not the right thing to do, but I guess I missed those ;).

NiLuJe:
TL;DR: On paper, your change looks like the perfect way to handle that ;).

(Assuming KOpt really needs to repaint both the ConfigDialog and the View, but I'm trusting you on that front ;)).

(I'm assuming you're using a 600x610 emulator window, right?)

poire-z:
> 600x610 emulator window, right

Right (on a small laptop screen, 1366x768).

> Assuming KOpt really needs to repaint both the ConfigDialog and the View

That surprised me, but I see that with some settings (contrast, line spacing), not with others (margins, font size).

> your change looks like the perfect way to handle that

No worry with the logs and the 3 `_refresh: Enqueued partial update for region 0 0 600 610` ?

```
_refresh: Enqueued ui update for region 163 514 424 32   must be the progress button, small h
setDirty via a func from widget table: 0x4085d3f8
setDirty via a func from widget table: 0x4085d3f8
_refresh: Enqueued partial update for region 0 0 600 610   this one might looks wrong, too early, or too big (if it's the configdialog)
setDirty partial from widget all w/ NO region  that's the one fixed to have kopt do its stuff
painting widget: ReaderUI     forcing kopt to repaint
loading koptcontext from ./cache/8d22901f6040cb5e7cf1b7ffa9b3a684  it repaings
painting widget: table: 0x40dab098   
_refresh: Enqueued partial update for region 0 0 600 610
_refresh: Enqueued partial update for region 0 0 600 610  double refresh ?
```

poire-z:
`_refresh: Enqueued partial update for region 0 0 600 610   this one might looks wrong, too early, or too big (if it's the configdialog)`
Well, no, it shouldn't be the configdialog size, as I ask it "all" "partial" to repaing kopt. So it's right I guess.
(It's for credocument that it will do only the bottom config size).

NiLuJe:
Yeah, not quite sure where that first partial is from... onTapSelect & onConfigChoose don't do a partial via func... Only onClose does :?

The double refresh at the end looks normal for a "all", "partial" with two top-level widgets open, though ;).

poire-z:
Oh, just discovered verbose mode (which does not work as an option to kodev run?).
Anyway, set via Developpers option, it shows the actual screen refreshs !
These shouldn't be hidden into verbose mode only :) May be condensed into a one-liner in simple debug mode?

So, it shows somethings that actually looks good to me ! Right ?
```
03/14/19-17:20:49 DEBUG adjusted ges: tap
03/14/19-17:20:49 WARN  ConfigDialog:onConfigChoose:update
03/14/19-17:20:49 DEBUG setDirty via a func from widget table: 0x40c9df90
03/14/19-17:20:49 DEBUG setDirty via a func from widget table: 0x40c9df90
03/14/19-17:20:49 DEBUG painting widget: table: 0x41718850
03/14/19-17:20:49 DEBUG _refresh: Enqueued fast update for region 163 514 424 32
03/14/19-17:20:49 DEBUG update_mode: Update refresh mode fast to ui
03/14/19-17:20:49 DEBUG _refresh: Enqueued ui update for region 163 514 424 32
# 03/14/19-17:20:49  triggering refresh {
    ["dither"] = false,
    ["mode"] = "ui",
    ["region"] = {
        ["y"] = 514,
        ["x"] = 163,
        ["h"] = 32,
        ["w"] = 424
    }
}
03/14/19-17:20:49 WARN  ConfigDialog:onConfigChoose:update tickAfterNext
03/14/19-17:20:49 WARN  self:update
03/14/19-17:20:49 DEBUG setDirty via a func from widget table: 0x40c9df90
03/14/19-17:20:49 DEBUG setDirty via a func from widget table: 0x40c9df90
03/14/19-17:20:49 WARN  self:update done
03/14/19-17:20:49 WARN  self:update done setDirty(all, partial)
03/14/19-17:20:49 DEBUG _refresh: Enqueued partial update for region 0 0 600 610
03/14/19-17:20:49 DEBUG setDirty partial from widget all w/ NO region
03/14/19-17:20:49 WARN  self:update done setDirty(all, partial) done
03/14/19-17:20:49 WARN  ConfigDialog:onConfigChoose:update tickAfterNext done
03/14/19-17:20:49 DEBUG painting widget: ReaderUI
# 03/14/19-17:20:49  readerview painting {
    ["y"] = 0,
    ["x"] = 0,
    ["h"] = 610,
    ["w"] = 600
} to 0 0
03/14/19-17:20:49 DEBUG loading koptcontext from ./cache/f5078849f8b7e252f3d07136207b5e8b
03/14/19-17:20:49 DEBUG painting widget: table: 0x41718850
03/14/19-17:20:49 DEBUG _refresh: Enqueued partial update for region 0 0 600 610
03/14/19-17:20:49 DEBUG _refresh: Enqueued partial update for region 0 0 600 610
# 03/14/19-17:20:49  triggering refresh {
    ["dither"] = false,
    ["mode"] = "partial",
    ["region"] = {
        ["y"] = 0,
        ["x"] = 0,
        ["h"] = 610,
        ["w"] = 600
    }
}

```

> The double refresh at the end looks normal for a "all", "partial" with two top-level widgets open, though ;).

So, if I get that right:
1) the (all, partial) just sets all widgets as dirty, no real refresh or repaint
2) UIManager makes each widget flagged as dirty paints itself (UIManager:_repaint())
3) after a widget paints itself, there's yet another call to _refresh() where the widget says what area it has painted needs to be refreshed (so, that would be the multiple 600 610 lines) - but it looks it's just when they are refresh_funcs
4) UIManager:_repaint() does at the end `for _, refresh in ipairs(self._refresh_stack) do Screen[refresh_methods...`

But I don't really get how/why the:
```
03/14/19-17:20:49 DEBUG painting widget: table: 0x41718850      done with all painting

03/14/19-17:20:49 DEBUG _refresh: Enqueued partial update for region 0 0 600 610
03/14/19-17:20:49 DEBUG _refresh: Enqueued partial update for region 0 0 600 610
  /\  why these 2 _refresh after all the paintings as I don't see refresh_funcs

  \/ and why/where they get condensed into a single Screen refresh:
# 03/14/19-17:20:49  triggering refresh {
```

NiLuJe:
(They're hidden behind verbose mode because of the table pretty-printing feature, IIRC).

NiLuJe:
And the answer to all your questions is the _refresh_stack in UIManager and how _refresh() manages it ;).

NiLuJe:
Oh, and I remember why I was confused by that initial partial: it's because the logs are inverted (for... Lua reasons) between setDirty and _refresh ;).

poire-z:
OK, that now makes more sense :) Thanks for the time and explanation.
Well, so it looks good. I'll check later on my device, and if no issue, I guess it will be good to go.

poire-z:
I guess we not need to worry about the numerous things like this I see with verbose mode (with the changes from this PR reverted):
```
03/14/19-18:39:00 DEBUG setDirty full from widget ReaderUI w/ NO region
# 03/14/19-18:39:00  INFO: invalid widget for setDirty() stack traceback:
  frontend/ui/uimanager.lua:560: in function 'post_guard'
  frontend/dbg.lua:45: in function 'setDirty'
  frontend/apps/reader/modules/readerview.lua:709: in function 'onSetScreenMode'
  frontend/apps/reader/modules/readerview.lua:754: in function 'handleEvent'
  frontend/apps/reader/readerui.lua:382: in function 'init'
  frontend/apps/reader/readerui.lua:515: in function 'doShowReader'
  frontend/apps/reader/readerui.lua:475: in function <frontend/apps/reader/readerui.lua:474>
03/14/19-18:39:00 DEBUG _refresh: Enqueued full update for region 0 0 600 610
03/14/19-18:39:00 DEBUG setDirty partial from widget ReaderUI w/ NO region
# 03/14/19-18:39:00  INFO: invalid widget for setDirty() stack traceback:
  frontend/ui/uimanager.lua:560: in function 'post_guard'
  frontend/dbg.lua:45: in function 'setDirty'
  frontend/apps/reader/modules/readerview.lua:603: in function 'recalculate'
  frontend/apps/reader/modules/readerview.lua:723: in function 'handleEvent'
  frontend/apps/reader/modules/readerview.lua:711: in function 'onSetScreenMode'

03/14/19-18:43:48 DEBUG setDirty via a func from widget table: 0x403c6070
# 03/14/19-18:43:48  INFO: invalid widget for setDirty() stack traceback:
  frontend/ui/uimanager.lua:560: in function 'post_guard'
  frontend/dbg.lua:45: in function 'setDirty'
  frontend/ui/widget/buttonprogresswidget.lua:90: in function 'update'
  frontend/ui/widget/buttonprogresswidget.lua:97: in function 'setPosition'
  frontend/ui/widget/configdialog.lua:526: in function 'init'
  frontend/ui/widget/configdialog.lua:570: in function 'init'
  frontend/ui/widget/configdialog.lua:768: in function 'update'
  frontend/ui/widget/configdialog.lua:799: in function 'onShowConfigPanel'
  frontend/apps/reader/modules/readerconfig.lua:84: in function 'onShowConfigMenu'
  frontend/apps/reader/modules/readerconfig.lua:92: in function 'handler'
  frontend/ui/widget/container/inputcontainer.lua:248: in function 'handleEvent'
  frontend/ui/uimanager.lua:632: in function 'sendEvent'
  frontend/ui/uimanager.lua:966: in function 'handleInputEvent'
  frontend/ui/uimanager.lua:1027: in function 'handleInput'
  frontend/ui/uimanager.lua:1071: in function 'run'
  ./reader.lua:216: in main chunk
  [C]: at 0x560d1e6dd0c0
03/14/19-18:43:48 DEBUG setDirty via a func from widget table: 0x403c6070
# 03/14/19-18:43:48  INFO: invalid widget for setDirty() stack traceback:
  frontend/ui/uimanager.lua:560: in function 'post_guard'
  frontend/dbg.lua:45: in function 'setDirty'
  frontend/ui/widget/configdialog.lua:802: in function 'onShowConfigPanel'
  frontend/apps/reader/modules/readerconfig.lua:84: in function 'onShowConfigMenu'
  frontend/apps/reader/modules/readerconfig.lua:92: in function 'handler'
  frontend/ui/widget/container/inputcontainer.lua:248: in function 'handleEvent'
  frontend/ui/uimanager.lua:632: in function 'sendEvent'
  frontend/ui/uimanager.lua:966: in function 'handleInputEvent'
  frontend/ui/uimanager.lua:1027: in function 'handleInput'
  frontend/ui/uimanager.lua:1071: in function 'run'
  ./reader.lua:216: in main chunk
  [C]: at 0x560d1e6dd0c0

```

NiLuJe:
Yeah, that's the debug guard firing for some real reasons, but mostly on a
widget's first showing when it's not yet in the top-level list ;).

NiLuJe:
Even in cases where set Dirty really is called on a widget that isn't top
level (there's quite a bit of calls like that left), I added the same basic
check to the runtime code, so it's not fantastically bad, just bad practice
;).

poire-z:
Looks all good on my device. So, merging.

PR #4797, opened 2019-03-14
ConfigDialog: remove gray underline on text items

poire-z:
Fix small issue in #4789 gray refactoring. See https://github.com/koreader/koreader/pull/4789#issuecomment-473034847


poire-z:
Hold on, finding a fix for https://github.com/koreader/koreader/pull/4789#issuecomment-473056137.

PR #4805, opened 2019-03-15
ConfigDialog ButtonProgress: black border on default item

poire-z:
Make the default (user set, or KOReader's default) item among the button progress squares shown with a black border. Details in https://github.com/koreader/koreader/pull/4700#issuecomment-473351859 and followups.

<kbd>image</kbd>

### DIFF
--- a/frontend/ui/widget/buttonprogresswidget.lua
+++ b/frontend/ui/widget/buttonprogresswidget.lua
@@ -19,6 +19,7 @@ local ButtonProgressWidget = InputContainer:new{
     enabled = true,
     num_buttons = 2,
     position = 1,
+    default_position = nil,
     thin_grey_style = false, -- default to black
 }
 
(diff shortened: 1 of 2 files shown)


frontend/ui/widget/buttonprogresswidget.lua
```diff
@@ -47,18 +48,24 @@ function ButtonProgressWidget:update()
...
         local button = Button:new{
             text = "",
             radius = 0,
             margin = margin,
             padding = button_padding,
-            bordersize = button_bordersize,
+            bordersize = button_bordersize + extra_border_size,
             enabled = true,
-            width = button_width,
+            width = button_width - 2*extra_border_size,
```
Frenzie: No concern about height?

poire-z: No. The height can overflow (see pic of the black style). For the width, we don't want that to keep having even dstributed squares all allong the progress bar, so we care about having the final width (inside + borders) similar.
That's for the black style we don't use :) so no worry at all for the thin grey style.

(and uh, uhnit what ? :)

PR #4808, opened 2019-03-16
[build] Bump base to make ninja the default CMake generator when available

Frenzie:
The difference for builds from scratch is negligible, but for rebuilds (i.e., for us hackers & the CI here in front) it should bring a speed improvement.

Automatic fallback to Make when Ninja is not available, or override with `USE_MAKE=1`.

You can (theoretically) choose your own generator with a sensible combination of `CMAKE_FLAGS` and `CMAKE_MAKE_PROGRAM`.

I also added `MAKE_FLAGS` and `NINJA_FLAGS`. That way you can easily pass `MAKE_FLAGS=-n` or `NINJA_FLAGS=-n` for a dry run, for example.

To switch you might have to run something like `make dist-clean USE_MAKE=1`.

Cf. https://github.com/koreader/koreader-base/pull/861, https://github.com/koreader/koreader-base/pull/862 and https://github.com/koreader/virdevenv/pull/34/files

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: frenzie/koappimage:0.1.3
+      - image: frenzie/koappimage:0.1.4
         environment:
           EMULATE_READER: 1
           # this is for shellcheck 0.4.5 and lower; can be removed for 0.4.6


Frenzie:
PS Although what seems to be 1-2 minutes faster on average for fresh builds is arguably not as negligible as I may have implied. It's still 5, maybe even as much as 10 %. It's just that most of the time is, obviously, spent compiling things. Still, those could be improved a fair bit by using gcc-8, which should be perfectly fine in front as long as we test 5 in base. It's not the most pressing concern since base is changed less frequently.

Note to self: https://askubuntu.com/a/1087116
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-get install gcc-8 g++-8
gcc-8 --version
```

I have some further thirdparty CMake build reorganizations in mind that should significantly reduce redundant actions.

poire-z:
Mhh, I don't have ninja installed, but I just noticed that crengine compiles quicker (lvtinydom.cpp was taking ~40s - now only 20s). Which is nice!
Would any of your other CMake related changes these last days have caused that ?
(No update that I know of has been made to my system recently.)

Frenzie:
I did add something about release vs debug, and of course that min CMake 3.5 thing. Maybe in 2.8 compat mode it erroneously recompiled some dependencies or something?

NiLuJe:
The buildsystems shouldn't affect the compilation time of a single file, though, unless compile flags are propagated differently.

Frenzie:
To be clear, all that changed is the switch to 3.5.1 (which should affect fairly little directly):
https://github.com/koreader/koreader-base/commit/b95da7ab8920c3367209b7e657f2a98f37c6d98b#diff-f775286a996a5a8898269c79ffd9d33e

And this little bit about release vs debug:
https://github.com/koreader/koreader-base/commit/31e23b7f6566922074ca00aeb9fcc970a92816b0#diff-f775286a996a5a8898269c79ffd9d33e

But are you talking about the actual compilation or the whole build process? It's possible that the entirety of thirdparty was sped up by up to a minute even without ninja, simply to figure out that there's nothing to do.

poire-z:
I was talking about the recompilation of a single file (+ the libcrengine.so made at the end).
But noticable on each single file when recompiling a few of them.

Also, looking at the size of some previous built of libcrengine.so I kept, it suddenly lose 10% in size:
```
12503856 Jun  7  2018 libcrengine.so.w_upstream
12502056 Jun  8  2018 libcrengine.so.LEADINGSPACEFIX
12583752 Jul 26  2018 libcrengine.so.OLD
12586448 Aug  1  2018 libcrengine.so.TEXT
12577496 Aug  9  2018 libcrengine.so.STYLES
12592328 Aug 11  2018 libcrengine.so.BEFORE
12593552 Aug 12  2018 libcrengine.so.HARFBUZZ
12597080 Aug 22  2018 libcrengine.so.w_UPSTREAM
12607480 Sep 14  2018 libcrengine.so.BEFORE_AVOID_WRAP
12769256 Oct 20 23:18 libcrengine.so.ORIG
12883472 Dec  3 23:59 libcrengine.so.CJK
12784264 Dec 15 14:37 libcrengine.so.before_table_tweak
12784400 Dec 15 14:45 libcrengine.so.after_table_tweak
12858488 Jan 26 20:24 libcrengine.so.master
12890480 Mar 12 07:21 libcrengine.so.GOOD
11549440 Mar 16 14:43 libcrengine.so
```

NiLuJe:
That would seem to indicate a different set of cflags ;).

Frenzie:
That should be the release thing then, I presume.

NiLuJe:
A release build should take *more* time to compile than a debug one,
though.

Frenzie:
I probably introduced a typo or logic error there. I'll double check it tonight.

Frenzie:
No, the logic is good. That CMakeLists.txt could potentially be slightly rewritten from my previous rewrite though, like this:

```diff
diff --git a/thirdparty/kpvcrlib/CMakeLists.txt b/thirdparty/kpvcrlib/CMakeLists.txt
index 24d352c..1832bf5 100644
--- a/thirdparty/kpvcrlib/CMakeLists.txt
+++ b/thirdparty/kpvcrlib/CMakeLists.txt
@@ -54,16 +54,8 @@ add_definitions(
     -DCR3_PATCH=1
 )
 
-if(${CMAKE_BUILD_TYPE} STREQUAL Debug)
-    add_definitions(
-        -D_DEBUG=1
-        -DDEBUG=1
-    )
-else()
-    add_definitions(
-        -DNDEBUG=1
-    )
-endif(${CMAKE_BUILD_TYPE} STREQUAL Debug)
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG=1 -DDEBUG=1")
+set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG=1")
```

Some people seem to call using if CMAKE_BUILD_TYPE an antipattern, but it's not clear to me why using `add_definitions` the way the manual claims you should use it (i.e., for preprocessor definitions) would be an antipattern. Presumably they mean it would make it too easy to do something properly stupid. Calling something an antipattern without saying why isn't helpful.

https://cmake.org/cmake/help/v3.8/command/add_definitions.html?highlight=add_definitions
> This command can be used to add any flags, but it is intended to add preprocessor definitions (see the add_compile_options() command to add other flags).

poire-z:
> crengine compiles quicker (lvtinydom.cpp was taking ~40s - now only 20s). Which is nice!

Again, the current situation is good, don't revert it if there is no logic error and it seems allright.
We don't _have to_ be as slow as before :)

Frenzie:
That's just the same thing written differently, no reversion of anything. ;-)

The old one said this:

```c
ADD_DEFINITIONS(-DUSE_FONTCONFIG=0 -DUSE_FREETYPE=1 -DUSE_HARFBUZZ=1 -DALLOW_KERNING=1 -DCR3_PATCH=1 -DNDEBUG=1)
```

I changed that `-DNDEBUG=1` flag to be release only, and added ` -D_DEBUG=1 -DDEBUG=1` to the debug build to make it potentially slightly more debuggy (for some thirdparty libs like antiword).

poire-z:
Mhhh, crengine on the emulator may have got slower :) Smaller but slower.
My usual benchmark load time + render time that I used to see at 13s + 4s is now 17 + 4s.
Any specific tweak/flag I could do/add to base current master to just get a chance to see previous results - just to be sure this decrease is really due to a different build - and not one of my latest commits I would have failed to benchmark?
Just replace `-D_DEBUG=1 -DDEBUG=1` with `-DNDEBUG=1` ?

NiLuJe:
Verbose build and check the CFLAGS/CXXFLAGS being passed ;).

poire-z:
```
koreader/base/thirdparty/kpvcrlib/crengine/thirdparty/antiword/antiword.h:13:2: error: #error Exactly one of the DEBUG and NDEBUG flags MUST be set
 #error Exactly one of the DEBUG and NDEBUG flags MUST be set
  ^~~~~
```
Even with this:
```
--- a/thirdparty/kpvcrlib/CMakeLists.txt
+++ b/thirdparty/kpvcrlib/CMakeLists.txt
@@ -55,9 +55,9 @@ add_definitions(
 )

 if(${CMAKE_BUILD_TYPE} STREQUAL Debug)
+    remove_definitions(-DDEBUG)
     add_definitions(
-        -D_DEBUG=1
-        -DDEBUG=1
+        -DNDEBUG=1
     )
 else()
     add_definitions(

```

NiLuJe:
Speaking of, was all of that from debug or release builds?

poire-z:
The emulator, do debug I guess (noticed some increase of my usual timings when benchmarking crengine changes with a same document - just wanted to be sure it's because of these changes in build flags.)

Frenzie:
@poire-z That should work. What's your CMake version? Perhaps you have to toss out the /build subdir to force a rebuild.

poire-z:
cmake 3.7.2. cleaning build/ dir didn't help.
Well, no need to spend more time on that.
I'lljust  check with next nightly vs some old 2019-02 that the release performances have not gotten worse.

NiLuJe:
FWIW, just did a Kobo build on my end, after poring through the build logs, I'm not seeing anything out of the ordinary.

(Or course, kpvcrlib is a silent build, so, there may be an extra sneaky behavior, but the ninja rules looked sane).

poire-z:
Forgot to confirm I get the same tinming on my kobo with v2019.03.1-31-g59d8a18_2019-03-16 than with a 2019-02 build.

So it's just the debug build being a bit slower - while compiling faster, which is fine by me :)

PR #4815, opened 2019-03-18
[UX] Gesture manager: add two-finger swipe left and right

Frenzie:
Cf. #4727.

What I don't quite know is how to present it. Should devices without multitouch receive a property in their device setup @pazos @NiLuJe @poire-z ? These events have technically always been present on readers like my H2O where they simply never fired.

Screenshot_2019-03-18_16-26-12


poire-z:
> Should devices without multitouch receive a property in their device setup ?

I guess so - it would be confusing to have them in the menu if they just won't work - or the user will keep trying to set a different action because let's try with another action...
(Not firing in the places where they are currently used is clear: they just don't work :)

There are also two fingers pinch & spread in readerfont.lua (I disable these ones in my own patch, because it is/was just too easy to trigger them by accident - and you have to wait for a re-rendering on a font you didn't want - and a few other re-renderings when you try to get your initial font size back).

(And I wouldn't mind all the two-fingers in a sub-menu if you come to add more, like 2-fingers up down diagonal... :)

Frenzie:
PS This stuff about hasMultitouch() is basically a big shot in the dark.

poire-z:
lgtm.
(Sure it's not cheaper to set the default `hasMultitouch = yes` and disable it just on those kobo and the few others that don't have it?)

Frenzie:
Could be, but then it'd have to be something like this to be accurate:

```diff
diff --git a/frontend/device/generic/device.lua b/frontend/device/generic/device.lua
index 3b7410be..088e7fc7 100644
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -26,7 +26,6 @@ local Device = {
     hasWifiToggle = yes,
     hasWifiManager = no,
     isTouchDevice = no,
-    hasMultitouch = no,
     hasFrontlight = no,
     needsTouchScreenProbe = no,
     hasClipboard = yes, -- generic internal clipboard on all devices
@@ -101,6 +100,10 @@ function Device:init()
         error("screen/framebuffer must be implemented")
     end
 
+    if self.hasMultitouch == nil then
+        self.hasMultitouch = self.isTouchDevice
+    end
+
     self.screen.isColorScreen = self.hasColorScreen
     self.screen.isColorEnabled = function()
         if G_reader_settings:has("color_rendering") then
```

I defaulted to `hasMultitouch = no` because `isTouchDevice = no` by default, which also seems a little odd now that you mention it. Perhaps that technically predates KOReader.

poire-z:
Ah, right, allright then.
(I see kindle and pocketbook start with isTouchDevice = no because probably the first models were not touch - unlike kobo which is touch from the start.)

Frenzie:
I suppose I'll rewrite it as suggested because it does indeed mean adding a lot fewer `yes`es. :-)

Frenzie:
There are a few caveats that I'm not quite sure what to do with. People can hack their older Kobo's to gain multitouch because the hardware supports it, just not the kernel. Presumably my H2O as well, though that shouldn't work with the kernel from this link https://www.mobileread.com/forums/showthread.php?t=232431 Perhaps [okreader](https://github.com/lgeek/okreader) supports it too? 

Not that any existing functionality will break, of course.

gerhaher:
> There are also two fingers pinch & spread in readerfont.lua (I disable these ones in my own patch, because it is/was just too easy to trigger them by accident - and you have to wait for a re-rendering on a font you didn't want - and a few other re-renderings when you try to get your initial font size back).

This annoys me too. How do I disable it?

Thanks!

Frenzie:
I assume @poire-z removes or comments out these lines:

https://github.com/koreader/koreader/blob/0a314e3ab8bbe0613a973986103b387346339ecd/frontend/apps/reader/modules/readerfont.lua#L101-L122

poire-z:
> removes or comments out these lines

Yes, in a small-diff way:
```
-                ges = "spread",
+                ges = "spread_DISABLED",
-                ges = "pinch",
+                ges = "pinch_DISABLED",
```

LucyOne:
Would it also be possible to adjust sensitivity? I often trigger "increase font size" accidentally when I want to swipe up and down for frontlight control.

PR #4820, opened 2019-03-19
[UX] Gesture manager: add diagonal two-finger swipes

Frenzie:
Follow-up to #4815. Pointed out as an easy target [here](https://github.com/koreader/koreader/pull/4815#issuecomment-473990141) by @poire-z. Diagonal two-finger swipes are currently unused.

@ersi Do you have any suggestions for defaults?

Screenshot_2019-03-19_20-34-25

PS Migrating up/down isn't self-evident because it does some fancy stuff with gesture length.

https://github.com/koreader/koreader/blob/728bb187fa0758fc9fbe6804a307638a37405fe1/frontend/apps/reader/modules/readerfrontlight.lua#L40-L65

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -172,8 +172,12 @@ function ReaderGesture:init()
         multiswipe_southeast_southwest_northwest = Device:hasWifiToggle() and "wifi_off" or "nothing",
         multiswipe_southeast_northeast_northwest = Device:hasWifiToggle() and "wifi_on" or "nothing",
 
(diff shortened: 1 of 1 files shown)


poire-z:
Could be nice if the 2-fingers menu would show the assigned action: `⬈ (action)` , like the multiswipe menu does.

Totally unrelated, but I don't know where to mention that:
Twice since yesterday on the emulator after rebuilding, and this morning with today's nightly on Android, it happened that I couldn't show the top menu with tap: nothing happened - while tap on bottom showed the bottom menu.
Restarting koreader made this work again. Dunno yet if it can happen again, of it it's some one-off thing.
I don't see anything in the last 2-3 days commits that could cause this.

Issue #4821, opened 2019-03-20
Handle external links on Android


minijackson:
* KOReader version: v2019.03.01
* Device: a3y17ltexc

#### Issue

As I understand, external links are currently handled only in the Wikipedia case. It would be nice to make KOReader open an external app to handle the links instead of showing an error window.

Thanks for the great work!

#### Steps to reproduce

- Have a book with non-wikipedia external links in it
- Read it
- Click on one of those links
- See error message

Frenzie:
As an aside, on Kobo etc. I want to do something similar to Wikipedia with Wallabag on all links.

PR #4822, opened 2019-03-20
[UX] Add Device:openLink() stub for Linux SDL

Frenzie:
References #4821.

### DIFF
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -614,6 +614,11 @@ function ReaderLink:onGotoLink(link, neglect_current_location, allow_footnote_po
         return true
     end
 
+    -- try opening link in native browser
+    if Device:openLink(link_url) then
+        return true
+    end
+
     -- Not supported
     UIManager:show(InfoMessage:new{
         text = T(_("Invalid or external link:\n%1"), link_url),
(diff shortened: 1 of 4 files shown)


poire-z:
It's minimal for now and non intrusive to other platforms, so probably fine.

But I guess one could still like to see the url in a confirmBox and confirm he really wants to launch a fat web browser to go there.
And more generically, there should probably be a bunch of new submenus in Links>:
`Wikipedia links > Ignore | Lookup in KOReader | Open external browser | Wallabag | WhateverOtherPluginThatCouldDoSomethingWithLinks` - and why not a checkbox `[ ] ask for confirmation`.
And the same for `External links>`.

On Android, I guess it should launch an intent that would be catch by any app able to handle http:// links.

pazos:
> On Android, I guess it should launch an intent that would be catch by any app able to handle http:// links.

It becomes harder on new android versions, but yeah. We need to take a string with url and create a new Uri with http scheme and then start the new activity via intent (we should catch the ActivityNotFound error that triggers if no activity is registered to view that kind of scheme and, optionally, notify the user that no external browser is available via android.notification).

Frenzie:
@poire-z 
> It's minimal for now and non intrusive to other platforms, so probably fine.

I don't really know what intrusive could even mean here? :-)

> But I guess one could still like to see the url in a confirmBox and confirm he really wants to launch a fat web browser to go there.

I'd say that's a matter for the URL handler, not for KOReader. If the URL handler launches a "fat web browser" that's the user's choice outside of KOReader. (Same for Android.) But sure, some kind of additional setting could be added in the future.

@pazos 
I'll try to refrain from commenting on Android, other than stating that it's typical of the system that there should be anything even remotely complicated about opening a simple link post-Windows 95 or so.

Issue #4823, opened 2019-03-20
Suspend action doesn't work on Kindle PW3


Galunid:
* KOReader version: v2019.03.1-24-..._2019-03-15
* Device: Kindle PW3

#### Issue
When I set gesture manager action nothing happens
#### Steps to reproduce
Set "Suspend" as a gesture on kindle

Frenzie:
Is suspend an action that is in fact available on Kindle? By which I mean that the issue title might be misleading and it should simply be something like "suspend is displayed when it shouldn't be".

Galunid:
Oh, I'm not sure if it's supported, it's displayed and can be ticked, so I assumed that it is supported

Frenzie:
But is it available through the exit menu?

pazos:
We can add (device:)canSuspend and canTogglePower instead of
https://github.com/koreader/koreader/blob/master/frontend/ui/elements/common_info_menu_table.lua#L64 and https://github.com/koreader/koreader/blob/master/frontend/ui/elements/common_info_menu_table.lua#L72

Frenzie:
Quite so. :+1: 

Btw, if you press <kbd>y</kbd> then it adds a specific commit so that you have a link that'll always work.

NB The emulator can also "suspend" (but not reboot afaik). Please keep that in mind while refactoring.

https://github.com/koreader/koreader/blob/b1ed2838c8b838e7f1d0b3cbf1aa9ca6ee547932/frontend/apps/reader/modules/readergesture.lua#L815-L816

Anyway @Galunid so it's available in the exit menu judging by what @pazos linked. Is it also functional?

Galunid:
In exit menu, it's called "Sleep", but it doesn't work

Galunid:
```lua
    -- part of UIManager:init()
    elseif Device:isKindle() then
        self.event_handlers["IntoSS"] = function()
            self:_beforeSuspend()
            Device:intoScreenSaver()
        end
        self.event_handlers["OutOfSS"] = function()
            Device:outofScreenSaver()
            self:_afterResume();
        end
        self.event_handlers["Charging"] = function()
            self:_beforeCharging()
            Device:usbPlugIn()
        end
        self.event_handlers["NotCharging"] = function()
            Device:usbPlugOut()
            self:_afterNotCharging()
        end
-- next elseif
```
this is then used in:
```lua
function UIManager:suspend()
[...]

PR #4825, opened 2019-03-20
Kindle suspend using lipc

Galunid:
fixes some stuff discussed in #4823

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -783,7 +783,11 @@ function ReaderGesture:gestureAction(action, ges)
             })
         end
     elseif action == "suspend" then
-        UIManager:suspend()
+        if Device:isKindle() then
+            Device.powerd:toggleSuspend(1)
+        else
+            UIManager:suspend()
+        end
     elseif action == "exit" then
         self.ui.menu:exitOrRestart()
     elseif action == "restart" then
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -783,7 +783,7 @@ function ReaderGesture:gestureAction(action, ges)
             })
         end
     elseif action == "suspend" then
-            UIManager:suspend()
+        UIManager:suspend()
     elseif action == "exit" then
         self.ui.menu:exitOrRestart()
     elseif action == "restart" then
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -783,11 +783,7 @@ function ReaderGesture:gestureAction(action, ges)
             })
         end
     elseif action == "suspend" then
-        if Device:isKindle() then
-            Device.powerd:toggleSuspend(1)
-        else
             UIManager:suspend()
-        end
     elseif action == "exit" then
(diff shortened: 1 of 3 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -783,7 +783,11 @@ function ReaderGesture:gestureAction(action, ges)
             })
         end
     elseif action == "suspend" then
-        UIManager:suspend()
+        if Device:isKindle() then
```
Frenzie: I assume this predates your other fix? The changes to this file are redundant and wouldn't belong here even if they weren't. ;-)

frontend/ui/uimanager.lua
```diff
@@ -1109,7 +1109,7 @@ function UIManager:suspend()
     if Device:isCervantes() or Device:isKobo() or Device:isSDL() or Device:isSonyPRSTUX() then
         self.event_handlers["Suspend"]()
     elseif Device:isKindle() then
-        self.event_handlers["IntoSS"]()
+        Device.powerd:toggleSuspend(1)
```
Frenzie: Is the idea that this power button press emulation will trigger an `IntoSS` event when that's handled in `frontend/device/input`?

frontend/device/kindle/powerd.lua
```diff
@@ -89,4 +89,10 @@ function KindlePowerD:afterResume()
     end
 end
 
+function KindlePowerD:toggleSuspend(status)
+    if self.lipc_handle then
+        self.lipc_handle:set_int_property("com.lab126.powerd", "powerButton", status)
```
Frenzie: lgtm, but @NiLuJe actually knows things about lipc

frontend/apps/reader/modules/readergesture.lua
Galunid: Oh yes, I accidentally added this, I'll remove it in a moment

frontend/ui/uimanager.lua
Galunid: Tbh I'm not certain, the idea is that power button suspends the device properly, so instead of creating such fixes, it directly emulates "power button press", thus triggering events related to this. It works normally and I don't think there could be any problems related to this. It is also what @NiLuJe suggested here https://github.com/koreader/koreader/issues/4823#issuecomment-474962797
It seems to work pretty well too.

Galunid:
@Frenzie It should be labeled as Kindle, not Kobo :wink:

Frenzie:
@Galunid I must've misclicked.

frontend/device/kindle/powerd.lua
NiLuJe: I actually haven't checked what passing 0 to it does. I think it'll do nothing, but I'll have to check.

Setting it to 1 does work when it's "asleep" for a wakeup, though.

TL;DR: This should probably always set it to 1 without accepting an input variable.

frontend/ui/uimanager.lua
NiLuJe: Yep, it'll fire the original event that IntoSS responds to.

frontend/device/kindle/powerd.lua
Galunid: Yes, I just checked, passing 0 does emulate button too

Issue #4845, opened 2019-03-26
Kindle Legacy March 2019 build - Crashes when navigating top menu bar


Tony556:
* KOReader version: March 2019
* Device: Kindle 3 and 4

#### Issue and steps to reproduce:
![image](https://user-images.githubusercontent.com/4276171/55037350-b46cba00-4fda-11e9-9426-05c57d50fec8.png)

Clicking the center in the 5-way here crashes the app.

##### `crash.log` (if applicable)
https://gist.github.com/Tony556/02a8e9c3e577a2e094e9c969ca5f00ea
https://gist.github.com/Tony556/2bb4ece14f9255f5262726fcb24e8dba

Tony556:
It seems to only be the gear, and the others seems to work fine

NiLuJe:
Confirmed on the latest nightly, too.

```
./luajit: frontend/ui/widget/touchmenu.lua:90: attempt to index a nil value
stack traceback:
        frontend/ui/widget/touchmenu.lua:90: in function 'init'
        frontend/ui/widget/widget.lua:48: in function 'new'
        frontend/ui/widget/touchmenu.lua:569: in function 'updateItems'
        frontend/ui/widget/touchmenu.lua:646: in function 'switchMenuTab'
        frontend/ui/widget/touchmenu.lua:311: in function 'callback'
        frontend/ui/widget/iconbutton.lua:106: in function 'action'
        frontend/ui/uimanager.lua:708: in function '_checkTasks'
        frontend/ui/uimanager.lua:985: in function 'handleInput'
        frontend/ui/uimanager.lua:1071: in function 'run'
        ./reader.lua:216: in main chunk
        [C]: at 0x000139b1
```

NiLuJe:
(I have been seeing a few of those `sizeUtf8Text called without text` everywhere, FWIW).

NiLuJe:
@Frenzie: It's apparently trying to show taps_and_gestures despite it behind an ifTouchDevice() block :?

@poire-z: This avoids the crash, but leaves a blank menu entry.

```
diff --git a/frontend/ui/rendertext.lua b/frontend/ui/rendertext.lua
index 6a0ca34c..f5095652 100644
--- a/frontend/ui/rendertext.lua
+++ b/frontend/ui/rendertext.lua
@@ -157,7 +157,7 @@ end
 function RenderText:sizeUtf8Text(x, width, face, text, kerning, bold)
     if not text then
[...]

Issue #4834, opened 2019-03-24
UI doesn't use the full screen on kobo aura H2O 2nd edition


BrainDamage:
* KOReader version: v2019.03.1-41-g54b0e85
* Device: Kobo Aura H2O 2nd Edition

#### Issue
UI does not fill the entire screen, leaving a white stripe on top, this is exacerbated when using night mode.

![IMG_20190324_174854](https://user-images.githubusercontent.com/50474/54882808-1ce25c80-4e5e-11e9-8bcd-882a8a864cad.jpg)

![IMG_20190324_174810](https://user-images.githubusercontent.com/50474/54882798-02a87e80-4e5e-11e9-8221-a2ea00eb07e6.jpg)

I'm not sure if it's some local hardware fault where my screen is misaligned. Is there a way to shift the screen coordinates?

Frenzie:
Presuambly the H2O2v2 doesn't cover up part of the screen with the bezel and no one's mentioned it.

https://github.com/koreader/koreader/blob/54b0e85bddda37be23013f4e4063d50738710d9b/frontend/device/kobo/device.lua#L137-L151

You can simply comment out the viewport, but if part of the screen is still hidden by the bezel you'll need to adjust it instead.

BrainDamage:
Ok, I've done a bit of experimentation:
first of all, my device is  Kobo Aura H2O2, KoboSnow ( I guess there was a miscommunication from my side about naming meant by 2nd edition )

second: the bezel does not cover the top part, but does cover partially the bottom, I have tested this by making the image much smaller, using night more and then gradually expanding towards the borders using successive trials.

the final geometry is:
viewport = Geom:new{x=0, y=0, w=1080, h=1436}
excluding the bottom 4 pixels

... which also explains why I always had issues with grabbing the ui from the bottom, the activation margins were actively much smaller than the screen due to being partially covered

NiLuJe:
Which revision of the H2O2 is that? The initial Mk. 6 version, or the Mk. 7 revision ?

Would it be possible for you to take a screengrab over SSH, from within Nickel, with fbgrab ? That would eliminate all the guesswork ;).

BrainDamage:
![out](https://user-images.githubusercontent.com/50474/54884451-8a4bb880-4e71-11e9-9177-0709ab08424c.png)

resolution seems to be 1080x1440

as for the revision: where am I supposed to check? I know the serial number, but I don't know how that translates to a mk. version

[...]

PR #4846, opened 2019-03-26
A few assorted tweaks

NiLuJe:
* Ignore the accelerometer when sleeping on Kobo (avoids spurious refreshes).
* Allow setting the autosuspend timeout from the UI (Gear > Device)
* Allow toggling the sleepcover from the UI (Gear > Device)
* Enable full granularity in most common font sizes in CRe (depends on https://github.com/koreader/koreader-base/pull/881 for actual effect).
* Don't crash when sizeUtf8Text is called on a nil (#4845)
* Remove viewport on the H2O², as there isn't actually one in effect (#4834)
* Remove viewport on the Aura SE (I'm going to chalk 1-2px on the side to personal preference).

### DIFF
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -138,6 +138,18 @@ if Device:isCervantes() or Device:isKobo() or Device:isSDL() or Device:isSonyPRS
     }
 end
 
+if Device:isKobo() then
+    common_settings.sleepcover = {
+        text = _("Ignore sleepcover events"),
+        checked_func = function()
+            return G_reader_settings:isTrue("ignore_power_sleepcover")
+        end,
+        callback = function()
+            G_reader_settings:flipNilOrFalse("ignore_power_sleepcover")
+        end
+    }
+end
+
 common_settings.night_mode = {
     text = _("Night mode"),
     checked_func = function() return G_reader_settings:readSetting("night_mode") end,
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -106,6 +106,38 @@ if Device:setDateTime() then
     }
 end
 
+if Device:isCervantes() or Device:isKobo() or Device:isSDL() or Device:isSonyPRSTUX() then
+    common_settings.autosuspend = {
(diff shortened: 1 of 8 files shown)


frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -106,6 +106,50 @@ if Device:setDateTime() then
     }
 end
 
+if Device:isCervantes() or Device:isKobo() or Device:isSDL() or Device:isSonyPRSTUX() then
+    common_settings.autosuspend = {
```
Frenzie: If this whole thing comes from the plugin doesn't it make more a lot more sense to simply stick it in the plugin? (MenuSorter doesn't care where it came from.)

Frenzie:
@NiLuJe Do you think that text thing is still necessary? A simple extra check suffices:

```diff
diff --git a/frontend/ui/menusorter.lua b/frontend/ui/menusorter.lua
index 8d7f0ae3..4fc64f37 100644
--- a/frontend/ui/menusorter.lua
+++ b/frontend/ui/menusorter.lua
@@ -56,7 +56,7 @@ function MenuSorter:sort(item_table, order)
             menu_table[order_id].id = order_id
             for order_number,order_number_id in ipairs(order_item) do
                 -- this is a submenu, mark it for later
-                if order[order_number_id] then
+                if item_table[order_number_id] ~= nil and order[order_number_id] then
                     table.insert(sub_menus, order_number_id)
                     tmp_menu_table[order_number] = {
                         id = order_number_id,
```

NiLuJe:
@Frenzie : Depends if we want to crash on error, there's a few call site where a field of the return table is accessed without nil guards.

frontend/ui/elements/common_settings_menu_table.lua
NiLuJe: I wasn't aware we could do that, but, yeah, that makes sense. Scratched my head a bit because AutoSuspend was layed out a bit weirdly, but finally figured it out ;).

NiLuJe:
(This one should probably be rebased, not squashed ;)).

frontend/ui/elements/common_settings_menu_table.lua
Frenzie: One of the main reasons I wrote it. ;-)

PR #4847, opened 2019-03-27
More blitting tweaks

NiLuJe:
Companion PR to https://github.com/koreader/koreader-base/pull/878

* Cleanup BitOpts usage (require & cache)
* Unify oddness checks (MOD -> AND)
* Enforce the native Portrait orientation on Kobo (except @ 16bpp, i.e., KSM w/ 8bpp swap disabled), to allow for faster blitting when unrotted.
* Switch CRe BB to 32BPP on color screens
* Minor cleanups


poire-z:
With today's nightly, all is fine and fast on my Kobo and the emulator.

But on my Android 6 phone, the screen is messed up and unreadable: again, some diagonal noise, like the scanline width is wrong.
That includes the cre book and the UI (even just the initial "Opening blah.epub" - it might have the good y and height, but the noise takes the whole width).

Dunno if it's specific to my Android 6, if some fullscreen/viewport twist is at work...

NiLuJe:
@poire-z : Does this help?

```diff
diff --git a/ffi/framebuffer_android.lua b/ffi/framebuffer_android.lua
index 8ab1af8..9226537 100644
--- a/ffi/framebuffer_android.lua
+++ b/ffi/framebuffer_android.lua
@@ -44,9 +44,9 @@ function framebuffer:_updateWindow()
     if buffer[0].format == C.WINDOW_FORMAT_RGBA_8888
     or buffer[0].format == C.WINDOW_FORMAT_RGBX_8888
     then
-        bb = BB.new(buffer[0].width, buffer[0].height, BB.TYPE_BBRGB32, buffer[0].bits, buffer[0].stride*4)
+        bb = BB.new(buffer[0].width, buffer[0].height, BB.TYPE_BBRGB32, buffer[0].bits, buffer[0].stride*4, buffer[0].stride)
     elseif buffer[0].format == C.WINDOW_FORMAT_RGB_565 then
-        bb = BB.new(buffer[0].width, buffer[0].height, BB.TYPE_BBRGB16, buffer[0].bits, buffer[0].stride*2)
+        bb = BB.new(buffer[0].width, buffer[0].height, BB.TYPE_BBRGB16, buffer[0].bits, buffer[0].stride*2, buffer[0].stride)
     else
         android.LOGE("unsupported window format!")
     end
```

poire-z:
Yep it does! All seems fine with this patch :)

poire-z:
Any combination of night mode and orientation seems fine too.

There's just a strange thing (seen on the emulator and android, no kobo at hand) that I don't think I introduced willingly, so it could be some effect of the BB changes (like background gray filling?) :
In landscape and in 2-pages mode, I now have some gray line in the middle of the middle margin, that I didn't have some days ago.
(And it's actually quite not bad :) I may have introduced that willingly if I had thought about it and seen how it looks :)

Wanted to check that on the emulator with blitbuffer.lua, without `libblitbuffer.so`. But if I move it out, the emulator is stuck on displaying only the widgets: the rendering of the book below stays blank, but the footer page number changes!

NiLuJe:
Cool, I'll push the Android fix in the next round of tweaks ;).

I'll see if I can reproduce the Landscape thing on Kobo, and fix it if needed, because the emulator definitely should work with the Lua blitter ;).

NiLuJe:
Hmm, no issues with either the Lua or C blitter @ 32bpp on Kobo, but then CRe renders @ 8bpp there ;).

NiLuJe:
Okay, can repro on the emulator.

And it actually crashes with the Lua blitter on my end ;D.

NiLuJe:
Okay, caught the crash, stupid typo. -_-".

Still not quite sure where that line is coming from though :D

NiLuJe:
It sill happens with the original BGR stuff, so, that's not it either.

poire-z:
>  so I'm going to assume it's coming from CRe's side, somehow? And only happens when it renders at 32bpp?

Exactly :) and that's a feature (only done when 32bpp), not a bug :
https://github.com/koreader/crengine/blob/8f15e01c8aea15920381f14ea85b0b4abeb36cfd/crengine/src/lvdocview.cpp#L2227-L2232

(Whether to keep it or remove it, or make it available on 8bpp too... is another question I don't want an answer to now :)

PR #4850, opened 2019-03-28
A few minor fixes after #4847

NiLuJe:
* Add a toggle to disable the C blitter in the Dev menu (depends on https://github.com/koreader/koreader-base/pull/882) (never shown if the JIT is disabled, grayed out if the C blitter is not installed)
* Fix a few sizeUtf8Text call sites that were doing a nil check in order to account for the new return type.
* Tweak statusbar handling to avoid spurious sizeUtf8Text warnings when it's hidden, and unify its behavior between being hidden via toggle, and hidden on book open (at least when all-at-once is not enabled).

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -263,6 +263,21 @@ function FileManagerMenu:setUpdateItemTable()
             end,
         })
     end
+    if not Device.should_restrict_JIT then
+        table.insert(self.menu_items.developer_options.sub_item_table, {
+            text = _("Disable use of the C blitter"),
+            enabled_func = function()
+                local lfs = require("libs/libkoreader-lfs")
+                return lfs.attributes("libs/libblitbuffer.so", "mode") == "file"
+            end,
+            checked_func = function()
+                return G_reader_settings:isTrue("dev_no_c_blitter")
+            end,
+            callback = function()
+                G_reader_settings:flipNilOrFalse("dev_no_c_blitter")
+            end,
+        })
+    end
     self.menu_items.cloud_storage = {
         text = _("Cloud storage"),
         callback = function()
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -265,7 +265,7 @@ function FileManagerMenu:setUpdateItemTable()
     end
     if not Device.should_restrict_JIT then
         table.insert(self.menu_items.developer_options.sub_item_table, {
-            text = _("Disable use of the C blitter"),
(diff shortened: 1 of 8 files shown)


frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -263,6 +263,21 @@ function FileManagerMenu:setUpdateItemTable()
             end,
         })
     end
+    if not Device.should_restrict_JIT then
+        table.insert(self.menu_items.developer_options.sub_item_table, {
+            text = _("Disable use of the C blitter"),
```
Frenzie:
```suggestion
            text = _("Disable C blitter"),
```

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -260,6 +260,10 @@ function FileManagerMenu:setUpdateItemTable()
             end,
             callback = function()
                 G_reader_settings:flipNilOrFalse("dev_startup_no_fbdepth")
+                local InfoMessage = require("ui/widget/infomessage")
+                UIManager:show(InfoMessage:new{
+                    text = _("This will take effect on next restart."),
```
Frenzie: Might as well write something more specific, like `The new bit depth will take effect after restarting KOReader.` or `KOReader will need to restart to enable the new bit depth.`

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -275,6 +279,10 @@ function FileManagerMenu:setUpdateItemTable()
             end,
             callback = function()
                 G_reader_settings:flipNilOrFalse("dev_no_c_blitter")
+                local InfoMessage = require("ui/widget/infomessage")
+                UIManager:show(InfoMessage:new{
+                    text = _("This will take effect on next restart."),
```
Frenzie: Same as above. Something like `The new blitter will take effect after restarting KOReader.` or `KOReader will need to restart to enable the new blitter.`

frontend/apps/filemanager/filemanagermenu.lua
Frenzie: Incidentally, I've been thinking that messages of this type could also consider using a ConfirmBox or something instead, with the cancel/OK buttons reading `Restart later` and `Restart now`.

NiLuJe: I mainly went with that generic wording because it's already used elsewhere, so, already translated ;).

Frenzie: That's true. On the flipside, the previous translation will show up as an  `xx %` match so it can be more easily adapted than starting from scratch. Since it's a more advanced config section I suppose we should just leave it then. :-)

PR #4852, opened 2019-03-29
[feat] plugins/wallabag: addArticle() method

Frenzie:
Currently unused, but a prerequisite for dynamically adding links to Wallabag.

### DIFF
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -526,15 +526,15 @@ function Wallabag:processLocalFiles( mode )
     return num_deleted
 end
 
-function Wallabag:addArticle(url)
-    logger.dbg("Wallabag: adding article ", url)
+function Wallabag:addArticle(article_url)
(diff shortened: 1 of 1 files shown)


Frenzie:
Pinging @y-muller in case I missed some more elegant method of doing this.

In the short term I just want to be able to add external links to Wallabag. This is particularly salient in Wallabag articles themselves. In the longer term, the list of articles to add could additionally quite simply be stored offline, and of course since the Wallabag API immediately returns a `content` key it could conceivably also be rendered in a ScrollHtmlWidget. (Basically, think of the way Wikipedia content functions thanks to @poire-z but more generalized thanks to Wallabag.)

y-muller:
This looks good @Frenzie .
I wanted to do this but I have not found the time recently.

Frenzie:
Thanks for the feedback!

It's just something I've been running into quite a lot now that I've started using Wallabag thanks to your plugin. :-)

Issue #4840, opened 2019-03-26
Irregular exit when 'Show more links as footnotes'


gerhaher:
* KOReader version: v2019.03.1-45-gb7b4cdd_2019-03-25/  

* Device: KA1

#### Issue
When I have 'Show more links as footnotes' enabled, some of the links concerned result in a crash when I tap on the screen to close the footnote pop up window.

#### Steps to reproduce
Book (split) in question: 
[The Wretched - Victor Hugo.zip](https://github.com/koreader/koreader/files/3007280/The.Wretched.-.Victor.Hugo.zip)

These links crash KOReader for me:
'palabres'
'de Maistre'
##### 
[crash.log](https://github.com/koreader/koreader/files/3007269/crash.log)

poire-z:
OK, easy fix, I'll PR it later today.
Happened because the link coherency check finds that link not coherent, so no higlighlit of the link and no unhighlight callback made, and try to call that.

Notes to self:
- I thought we were done with the "not a coherent xpointer" - but it happened on this, possibly because it's some italic and we are since some weeks ago adjusting the start of the Rect, so it's possible in this check, we use one without the adjustment which is now outside of the rect, so no more coherency. A simple x+2 makes it valid again:
```diff
 function ReaderLink:isXpointerCoherent(a_xpointer)
     -- Get screen coordinates of xpointer
+    logger.warn("<", a_xpointer)
     local screen_y, screen_x = self.ui.document:getScreenPositionFromXPointer(a_xpointer)
     -- Get again link and a_xpointer from this position
-    local re_link_xpointer, re_a_xpointer = self.ui.document:getLinkFromPosition({x = screen_x, y = screen_y}) -- luacheck: no unused
+    local re_link_xpointer, re_a_xpointer = self.ui.document:getLinkFromPosition({x = screen_x+2, y = screen_y}) -- luacheck: no unused
+    logger.warn(">", re_a_xpointer)
     -- We should get the same a_xpointer. If not, crengine has messed up
     -- and we should not trust this xpointer to get back to this link.
     if re_a_xpointer ~= a_xpointer then
```

<kbd>![image](https://user-images.githubusercontent.com/24273478/54996775-7ebfd500-4fca-11e9-9cbe-a5d895d7be1e.png)</kbd>
- No unhighlight is done when exiting the footnote popup with the Esc key.

PR #4853, opened 2019-03-29
Night mode: keep some images non-inverted

### LINKED ISSUE #2571: Feature Request: can we make night mode ignore photos from being inverted
* KOReader version:
* Device:kobo aura 

**It would be great to ignore them from being inverted.**

**Thanks**

a78c0563a026e149af9a6c012711c7b2
d031b5ba0748c68369f0a0dde1a1206b

poire-z:
Have the code pre-invert some images when in night mode, so they get inverted back to normal by night mode.
This will allow images to be displayed normally in the following contexts:
- images in credocument pages
- images long-pressed in credocument and viewed in ImageViewer
- cover thumbnails in File browser and History
- full size covers when viewed in ImageViewer
- images in Wikipedia lookup results

There is still some issues when the lua blitter is used (Kobo, Kindle), where some images may be wrongly or not inverted, or messed up. No issue seen when the C blitter is used (Android, emulator). Discussed in #2571. Closes #2571.

Includes https://github.com/koreader/crengine/pull/276:
- DrawBuf: allow drawing images inverted (to be used in night mode)
- ldomXPointer::getRect(): adds adjusted=false parameter
- CSS: fix parsing of "! important" with spaces
- CSS: fix non-deterministic combinators
- Delay some style computations until needed https://github.com/koreader/crengine/pull/278

Also:
Footnote popups: fix link unhighlight when closing with keys. (Fix last item in https://github.com/koreader/koreader/issues/4840#issuecomment-476603558)

### DIFF
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -365,14 +365,22 @@ function CreDocument:drawCurrentView(target, x, y, rect, pos)
     end
     -- TODO: self.buffer could be re-used when no page/layout/highlights
     -- change has been made, to avoid having crengine redraw the exact
-    -- same buffer.
-    -- And it could only change when some other methods from here are called
-    --local start_clock = os.clock()
-    self._drawn_images_count, self._drawn_images_surface_ratio = self._document:drawCurrentPage(self.buffer, self.render_color)
-    --print(string.format("CreDocument:drawCurrentView: Rendering took %9.3f ms", (os.clock() - start_clock) * 1000))
-    --start_clock = os.clock()
+    -- same buffer. And it could only change when some other methods
+    -- from here are called
+
+    -- If in night mode, ask crengine to invert all images, so they
+    -- get displayed in their original colors when the whole screen
+    -- is inverted by night mode
+    local invert_images = G_reader_settings:isTrue("night_mode")
+
+    -- local start_clock = os.clock()
+    self._drawn_images_count, self._drawn_images_surface_ratio =
+        self._document:drawCurrentPage(self.buffer, self.render_color, invert_images)
+    -- print(string.format("CreDocument:drawCurrentView: Rendering took %9.3f ms", (os.clock() - start_clock) * 1000))
+
+    -- start_clock = os.clock()
     target:blitFrom(self.buffer, x, y, 0, 0, rect.w, rect.h)
-    --print(string.format("CreDocument:drawCurrentView: Blitting took  %9.3f ms", (os.clock() - start_clock) * 1000))
+    -- print(string.format("CreDocument:drawCurrentView: Blitting took  %9.3f ms", (os.clock() - start_clock) * 1000))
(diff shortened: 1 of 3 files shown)


poire-z:
I'll bump base in a rebase of the last commit once it's merged.

frontend/document/credocument.lua
```diff
@@ -365,14 +365,22 @@ function CreDocument:drawCurrentView(target, x, y, rect, pos)
...
+
+    -- local start_clock = os.clock()
+    self._drawn_images_count, self._drawn_images_surface_ratio =
+        self._document:drawCurrentPage(self.buffer, self.render_color, invert_images)
+    -- print(string.format("CreDocument:drawCurrentView: Rendering took %9.3f ms", (os.clock() - start_clock) * 1000))
+
+    -- start_clock = os.clock()
     target:blitFrom(self.buffer, x, y, 0, 0, rect.w, rect.h)
-    --print(string.format("CreDocument:drawCurrentView: Blitting took  %9.3f ms", (os.clock() - start_clock) * 1000))
+    -- print(string.format("CreDocument:drawCurrentView: Blitting took  %9.3f ms", (os.clock() - start_clock) * 1000))
```
NiLuJe: Whitespace :D.

(Extreme nitpicking aside, that was to differentiate between comments and commented out code ;)).

NiLuJe: Looking at the file and not the diff, I see those aren't right next to comments anymore, so, yeah, that works, too, never mind me ;).

Frenzie: Interesting concept. I've spotted it but I thought it was just an inconsistency. ;-)

(I do in fact keep myself from posting nitpicks now and then. ;-D)

poire-z: It was extra hard to read to me with no whitespace :)
(I would have remove it if you just hadn't added it :) I think profiling code is just as easy to add and remove only when needed. Can/should I ?)

NiLuJe: That was mainly to be able to do it on-device over SSH via nano (instead of a patch/stash + scp from the desktop) ;).

poire-z: Or may be we could just have somewhere a function available for quick profiling, to avoid losing a few minutes to think about os.clock and string.format, that we could use as:
```diff
+ timeit("rendering document", function()
     original code to time
+ end)
```
or something even easier if some Lua constructs allow for easier, like C define...

NiLuJe: Yeah, I'm not terribly fond the the "wrap it in an anonymous func" approach usually taken for that kind of thing in Lua, but that's just me not being fond of many many Lua idiosyncracies in general ;).

NiLuJe: But yeah, having that kind of thing available in any manner somewhere would defintely come in handy ;).

PR #4857, opened 2019-03-30
Fix tap for top menu occasionally not working

### LINKED ISSUE #4856: Tap on top occasionally does not open the top menu
* KOReader version: v2019.03.1-64-g8f019b8_2019-03-30 and current master
* Device: Android, emulator

Twice yesterday on the emulator after rebuilding, and this morning with today's nightly on Android, it happened that I couldn't show the top menu with tap: nothing happened - while tap on bottom showed the bottom menu.
I had never seen that happening before.
But I don't see anything in the last 2-3 days commits that could cause this...

And somehow, when on a startup when it didn't work when tapping in the middle, it seems to work when I tap on that small area here in yellow ...
image
On Android too, when I feel it does not work, taping that same area have it work again.
Does not happen on my Kobo after numerous restarts.

Switching to another cre document does not solve it.
Switching to a PDF document have top menu working when tap on middle, but switching back to a CRE document has it still not working.
Restarting koreader may have it working again, or not...

Differences in logs of 2 startup of the emulator on the same first page of a book, when menu does not open and go to next page (-) vs when the menu open (+):
```diff
- [*] Current time: 03/30/19-16:25:00
+ [*] Current time: 03/30/19-16:25:15
  [*] Version: v2019.03.1-64-g8f019b81_2019-03-30

 ffi.load: SDL2
@@ -41,15 +41,15 @@
 DEBUG data is read from  Wikipedia/Alamans.FR.sdr/metadata.epub.lua
 DEBUG ImageWidget: _render'ing resources/info-i.png
 DEBUG renderImageData: using MuPDF
-DEBUG Mupdf.renderImage true cdata<struct BlitBuffer8>: 0x40ed4458
+DEBUG Mupdf.renderImage true cdata<struct BlitBuffer8>: 0x4009a4c8
 DEBUG cache image|resources/info-i.png||
-DEBUG show widget: table: 0x4072dcf0
-DEBUG setDirty nil from widget table: 0x4072dcf0 w/ NO region
-DEBUG setDirty via a func from widget table: 0x4072dcf0
-DEBUG painting widget: table: 0x41e97680
+DEBUG show widget: table: 0x409afe18
+DEBUG setDirty nil from widget table: 0x409afe18 w/ NO region
+DEBUG setDirty via a func from widget table: 0x409afe18
+DEBUG painting widget: table: 0x40823438
 DEBUG blitFrom 65 274 0 0 60 60
 DEBUG _refresh: Enqueued ui update for region 58 257 483 95
-DEBUG close widget: table: 0x4072dcf0
+DEBUG close widget: table: 0x409afe18
 DEBUG setDirty via a func from widget nil
 DEBUG creating coroutine for showing reader
 INFO  opening file Wikipedia/Alamans.FR.epub
@@ -68,7 +68,7 @@
 DEBUG data is read from  Wikipedia/Alamans.FR.sdr/metadata.epub.lua
 DEBUG ImageWidget: _render'ing resources/icons/appbar.book.open.png
 DEBUG renderImageData: using MuPDF
-DEBUG Mupdf.renderImage true cdata<struct BlitBuffer8>: 0x417c1680
+DEBUG Mupdf.renderImage true cdata<struct BlitBuffer8>: 0x41912360
 DEBUG cache image|resources/icons/appbar.book.open.png||
 DEBUG Getting list of dictionaries
 DEBUG found 12 dictionaries
@@ -85,8 +85,8 @@
 INFO  Plugin  plugins/autofrontlight.koplugin/main.lua  has been disabled.
 INFO  Plugin  plugins/timesync.koplugin/main.lua  has been disabled.
 DEBUG AutoSuspend: deprecateLastTask  1
-DEBUG AutoSuspend: start at  1553959501
-DEBUG AutoSuspend: schedule at  1553963101
+DEBUG AutoSuspend: start at  1553959515
+DEBUG AutoSuspend: schedule at  1553963115
 INFO  RD loaded plugin SSH at plugins/SSH.koplugin
 INFO  RD loaded plugin autosuspend at plugins/autosuspend.koplugin
 INFO  RD loaded plugin batterystat at plugins/batterystat.koplugin
@@ -182,22 +182,22 @@
 DEBUG _repaint: it was dithered, infecting the refresh queue
 DEBUG update_dither: Update dither hint false to true
 DEBUG _refresh: Enqueued full update for region 0 0 600 610 w/ HW dithering
-DEBUG input event => type: 3, code: 57(nil), value: 0, time: 1553959503.360314
-DEBUG input event => type: 3, code: 53(nil), value: 291, time: 1553959503.360334
-DEBUG input event => type: 3, code: 54(nil), value: 56, time: 1553959503.360341
-DEBUG input event => type: 0, code: 0(nil), value: 0, time: 1553959503.360348
+DEBUG input event => type: 3, code: 57(nil), value: 0, time: 1553959516.922920
+DEBUG input event => type: 3, code: 53(nil), value: 290, time: 1553959516.922940
+DEBUG input event => type: 3, code: 54(nil), value: 55, time: 1553959516.922946
+DEBUG input event => type: 0, code: 0(nil), value: 0, time: 1553959516.922953
 DEBUG in tap state...
 DEBUG set up hold timer
 DEBUG adjusted ges: touch
 DEBUG AutoSuspend: onInputEvent
-DEBUG input event => type: 3, code: 57(nil), value: -1, time: 1553959503.497109
-DEBUG input event => type: 0, code: 0(nil), value: 0, time: 1553959503.497118
+DEBUG input event => type: 3, code: 57(nil), value: -1, time: 1553959517.49168
+DEBUG input event => type: 0, code: 0(nil), value: 0, time: 1553959517.49177
 DEBUG in tap state...
 DEBUG set up tap timer
 DEBUG in tap timer true
 DEBUG single tap detected in slot 0 {
-    ["y"] = 56,
-    ["x"] = 291,
+    ["y"] = 55,
+    ["x"] = 290,
     ["h"] = 0,
     ["w"] = 0
 }
 DEBUG adjusted ges: tap
 DEBUG AutoSuspend: onInputEvent
 DEBUG getLinkFromPosition link_xpointer:
 DEBUG getLinkFromPosition a_xpointer:
 DEBUG clear selection
 DEBUG document has no pages at {
     ["page"] = 1,
-    ["x"] = 291,
+    ["x"] = 290,
     ["w"] = 0,
-    ["y"] = 56,
+    ["y"] = 55,
     ["h"] = 0
 }
-DEBUG goto relative screen: 1 , in mode:  page
-DEBUG CreDocument: goto page 2
-DEBUG _refresh: Enqueued partial update for region 0 0 600 610
-DEBUG setDirty partial from widget ReaderUI w/ NO region
-DEBUG CreDocument: check xpointer in current page /body/DocFragment/body/div/p[20]/text()[1].0
-DEBUG CreDocument: check xpointer in current page /body/DocFragment/body/div/p[1]/text()[18].6
+DEBUG ImageWidget: _render'ing resources/icons/appbar.transform.rotate.right.large.png 35 35
+DEBUG renderImageData: using MuPDF
+DEBUG Mupdf.renderImage true cdata<struct BlitBuffer8>: 0x40f1ab00
+DEBUG cache image|resources/icons/appbar.transform.rotate.right.large.png|35|35
+DEBUG ImageWidget: initial offsets 0 0
+DEBUG ImageWidget: _render'ing resources/icons/appbar.column.two.large.png 35 35
+DEBUG renderImageData: using MuPDF
+DEBUG Mupdf.renderImage true cdata<struct BlitBuffer8>: 0x41feb5e8
```

poire-z:
It was missing some ordering hints to have it always precede taps to change page.
(Somehow noticed only when we reached > 32 gestures registered, which may have introduced some shuffling, while previously the addition order was somehow preserved...)
Details and observations in #4856. Closes #4856.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -589,7 +589,7 @@ function ReaderGesture:setupGesture(ges, action)
         if self.is_docless then
             overrides = { 'filemanager_tap' }
         else
-            overrides = { 'readerfooter_tap', }
+            overrides = { 'readerfooter_tap' }
         end
     elseif ges == "tap_left_bottom_corner" then
         ges_type = "tap"
@@ -600,7 +600,7 @@ function ReaderGesture:setupGesture(ges, action)
         if self.is_docless then
             overrides = { 'filemanager_tap' }
         else
-            overrides = { 'readerfooter_tap', 'filemanager_tap' }
+            overrides = { 'readerfooter_tap' }
         end
     elseif ges == "two_finger_swipe_west" then
         ges_type = "two_finger_swipe"
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -587,9 +587,13 @@ function ReaderGesture:setupGesture(ges, action)
             ratio_w = 0.1, ratio_h = 0.1,
         }
         if self.is_docless then
-            overrides = { 'filemanager_tap' }
+            overrides = {
+                "filemanager_tap",
+            }
         else
-            overrides = { 'readerfooter_tap' }
+            overrides = {
+                "readerfooter_tap",
+            }
         end
     elseif ges == "tap_left_bottom_corner" then
         ges_type = "tap"
@@ -598,9 +602,13 @@ function ReaderGesture:setupGesture(ges, action)
             ratio_w = 0.1, ratio_h = 0.1,
         }
         if self.is_docless then
-            overrides = { 'filemanager_tap' }
+            overrides = {
+                "filemanager_tap",
+            }
         else
-            overrides = { 'readerfooter_tap' }
+            overrides = {
+                "readerfooter_tap",
+            }
         end
     elseif ges == "two_finger_swipe_west" then
         ges_type = "two_finger_swipe"
@@ -635,9 +643,14 @@ function ReaderGesture:setupGesture(ges, action)
         direction = {northeast = true, northwest = true, southeast = true, southwest = true}
         distance = "short"
         if self.is_docless then
-            overrides = { 'filemanager_tap' }
+            overrides = {
+                "filemanager_tap",
+            }
         else
-            overrides = { 'rolling_swipe', 'paging_swipe' }
+            overrides = {
+                "rolling_swipe",
+                "paging_swipe",
+            }
         end
 
     else return
--- a/frontend/apps/reader/modules/readermenu.lua
+++ b/frontend/apps/reader/modules/readermenu.lua
@@ -97,7 +97,10 @@ function ReaderMenu:onReaderReady()
                 ratio_x = DTAP_ZONE_MENU.x, ratio_y = DTAP_ZONE_MENU.y,
                 ratio_w = DTAP_ZONE_MENU.w, ratio_h = DTAP_ZONE_MENU.h,
             },
-            overrides = { 'tap_forward', 'tap_backward', },
+            overrides = {
+                "tap_forward",
+                "tap_backward",
+            },
             handler = function(ges) return self:onTapShowMenu(ges) end,
         },
         {
@@ -107,7 +110,10 @@ function ReaderMenu:onReaderReady()
                 ratio_x = DTAP_ZONE_MENU.x, ratio_y = DTAP_ZONE_MENU.y,
                 ratio_w = DTAP_ZONE_MENU.w, ratio_h = DTAP_ZONE_MENU.h,
             },
-            overrides = { "rolling_swipe", "paging_swipe", },
+            overrides = {
+                "rolling_swipe",
+                "paging_swipe",
+            },
             handler = function(ges) return self:onSwipeShowMenu(ges) end,
         },
         {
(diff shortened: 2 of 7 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -589,7 +589,7 @@ function ReaderGesture:setupGesture(ges, action)
         if self.is_docless then
             overrides = { 'filemanager_tap' }
         else
-            overrides = { 'readerfooter_tap', }
+            overrides = { 'readerfooter_tap' }
```
Frenzie: Or alternatively you could write it like like this:
```
overrides = {
    'readerfooter_tap',
}
```
It makes no difference here, but it would for future diffs.

frontend/apps/reader/modules/readermenu.lua
```diff
@@ -97,6 +97,7 @@ function ReaderMenu:onReaderReady()
                 ratio_x = DTAP_ZONE_MENU.x, ratio_y = DTAP_ZONE_MENU.y,
                 ratio_w = DTAP_ZONE_MENU.w, ratio_h = DTAP_ZONE_MENU.h,
             },
+            overrides = { 'tap_forward', 'tap_backward', },
```
Frenzie: Huh, how odd that this never caused any issues before. :-)

KenMaltby:
WWII gremlins coming out of hibernation?

frontend/apps/reader/modules/readergesture.lua
poire-z: Done, for all of them while I was at it.

NiLuJe: Huh, never thought of that, will try to remember it next time I deal with a similar construct ;).

Frenzie: For short things like the above I don't really mind, but if I notice something like this I'll definitely point out the fact that you could have a diff with one line added vs two lines changed. It just makes it ever so slightly harder to focus on what's relevant. :-)

```lua
{
'adfasd',
';lkj' -- no comma
}
```

PR #4862, opened 2019-03-30
[Android] use the same tag for all logs

pazos:
get android logger tag from application name ("KOReader")

~~debug logging can be enabled using developer options like in other platforms.~~

requires https://github.com/koreader/android-luajit-launcher/pull/136 & bumping that module

### DIFF
--- a/platform/android/llapp_main.lua
+++ b/platform/android/llapp_main.lua
@@ -1,6 +1,5 @@
 local A = require("android")
 A.dl.library_path = A.dl.library_path .. ":" .. A.dir .. "/libs"
-A.log_name = 'KOReader'
 
 local ffi = require("ffi")
 local dummy = require("ffi/posix_h")
@@ -17,7 +16,7 @@ end
 -- run koreader patch before koreader startup
 pcall(dofile, "/sdcard/koreader/patch.lua")
 
--- set proper permission for sdcv
+-- set proper permission for binaries
 A.execute("chmod", "755", "./sdcv")
 A.execute("chmod", "755", "./tar")
 A.execute("chmod", "755", "./zsync")
--- a/Makefile
+++ b/Makefile
@@ -352,7 +352,7 @@ androidupdate: all
 	cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=1 \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).7z * \
 		-x!resources/fonts -x!resources/icons/src -x!spec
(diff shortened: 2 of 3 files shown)


platform/android/llapp_main.lua
```diff
@@ -26,5 +25,5 @@ A.execute("chmod", "755", "./zsync")
 C.setenv("TESSDATA_PREFIX", "/sdcard/koreader/data", 1)
 
 -- create fake command-line arguments
-arg = {"-d", file or "/sdcard"}
+arg = {file or "/sdcard"}
```
Frenzie: On Android you can't easily adjust the command line arguments like on other platforms, so I'm not convinced this is a good change.

NiLuJe: I think the intent is to take advantage of the debug toggle done in reader.lua instead of hard-coding this here?

Frenzie: My point is that if you can't reach that toggle, it's a minor inconvenience on other platforms since you can launch the program with -d manually. For Android, by contrast, you need an Android dev environment.

Frenzie: Although now that I think about it you can easily add the option to settings.lua now, so never mind I suppose.

But I very much don't like it regardless. We know when it's a debug flavor, right? So please enable it by default for a debug build. ;-)

NiLuJe: Eh, fair point.

Not necessarily a fan of enforcing debug logging by default, though, but I see what you mean.

NiLuJe: Oh, right, making it depend on the build flavor does take care of the situation in a more elegant manner :).

pazos: Errr, AFAICT it just hides the blitbuffer code a bit. It doesn't affect to luajit-launcher stuff that is calling androig.log directly.

The only difference between debug and release builds should be that in release builds we don't log debug statements (aka Log.d, which is currently unused in both frontend and launcher)

Frenzie: I vehemently disagree. I want my debug builds to fail hard on asserts. You could distinguish between different types of debug builds, but a build with debugging symbols combined with NDEBUG false barely serves a purpose in my eyes, if any. You could always implement that distinction if you want. ;-)

However, that's neither here nor there. This change removes debug statements for any build type regardless of assert behavior. I spent a lot of time and effort on making `./kodev run android` as painless as possible. I'm not responsible for Android being slightly overenthusiastic about debug logging, but I would've nevertheless made sure it behaved that way. An important aspect of painless is the `-d` flag like for the desktop emulator:

https://github.com/koreader/koreader/blob/7396abebf26ecbb3a8cf45f4ee4dcb61cb788287/kodev#L663

I don't much care how it's done, but I strongly object to anything that makes Android even the tiniest bit harder. Because as stupid as it might sound, that extra bit of friction, no matter how small, directly translates into less or no Android testing on my end.

pazos: > You could always implement that distinction if you want. ;-)

Nah, I really don't care.

Frenzie:
Thanks!

PR #4863, opened 2019-03-31
[feat, UX] Add external link to Wallabag

Frenzie:
Offers a choice when Wallabag is available.

## Android/desktop (with Wallabag enabled)
Screenshot_2019-03-31_15-39-29

## Other platforms (with Wallabag enabled)
Screenshot_2019-03-31_15-56-03

### DIFF
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -127,6 +127,9 @@ function ReaderLink:addToMainMenu(menu_items)
             },
             {
                 text = _("External link action"),
+                enabled_func = function()
+                    return self.ui.document.info.has_pages or not isTapIgnoreExternalLinksEnabled
+                end,
                 sub_item_table = {
                     {
                         text = _("Ask with pop-up dialog"),
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -2,6 +2,7 @@
 ReaderLink is an abstraction for document-specific link interfaces.
 ]]
 
+local ButtonDialogTitle = require("ui/widget/buttondialogtitle")
 local ConfirmBox = require("ui/widget/confirmbox")
 local Device = require("device")
 local Event = require("ui/event")
@@ -517,6 +518,57 @@ function ReaderLink:onGotoLink(link, neglect_current_location, allow_footnote_po
     end
     logger.dbg("External link:", link_url)
 
+    local is_http_link = link_url:find("^https?://") ~= nil
+    if is_http_link and self:onGoToExternalLink(link_url) then
+        return true
+    end
+
+    -- Check if it is a link to a local file
+    local linked_filename = link_url:gsub("^file:", "") -- remove local file protocol if any
+    local anchor
+    if linked_filename:find("?") then -- remove any query string (including any following anchor)
+        linked_filename, anchor = linked_filename:match("^(.-)(%?.*)$")
+    elseif linked_filename:find("#") then -- remove any anchor
+        linked_filename, anchor = linked_filename:match("^(.-)(#.*)$")
+    end
+    linked_filename  = ffiutil.joinPath(self.document_dir, linked_filename) -- get full path
+    linked_filename = ffiutil.realpath(linked_filename) -- clean full path from ./ or ../
+    if linked_filename and lfs.attributes(linked_filename, "mode") == "file" then
+        local DocumentRegistry = require("document/documentregistry")
+        local provider = DocumentRegistry:getProvider(linked_filename)
+        if provider then
+            -- Display filename with anchor or query string, so the user gets
+            -- this information and can manually go to the appropriate place
+            local display_filename = linked_filename
+            if anchor then
+                display_filename = display_filename .. anchor
+            end
+            UIManager:show(ConfirmBox:new{
+                text = T(_("Would you like to read this local document?\n\n%1\n"), display_filename),
+                ok_callback = function()
+                    UIManager:scheduleIn(0.1, function()
+                        self.ui:switchDocument(linked_filename)
+                    end)
+                end
+            })
+        else
+            UIManager:show(InfoMessage:new{
+                text = T(_("Link to unsupported local file:\n%1"), link_url),
+            })
+        end
+        return true
+    end
+
+    -- Not supported
+    UIManager:show(InfoMessage:new{
+        text = T(_("Invalid or external link:\n%1"), link_url),
+        -- no timeout to allow user to type that link in his web browser
+    })
+    -- don't propagate, user will notice and tap elsewhere if he wants to change page
+    return true
+end
+
+function ReaderLink:onGoToExternalLink(link_url)
     -- Check if it is a wikipedia link
     local wiki_lang, wiki_page = link_url:match([[https?://([^%.]+).wikipedia.org/wiki/([^/]+)]])
     if wiki_lang and wiki_page then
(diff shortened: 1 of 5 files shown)


frontend/apps/reader/modules/readerlink.lua
```diff
@@ -576,56 +628,60 @@ function ReaderLink:onGotoLink(link, neglect_current_location, allow_footnote_po
             })
         end
         return true
-    end
+    elseif self.ui.wallabag then
+        local settings = G_reader_settings:readSetting("external_link_action")
```
poire-z: Forgot to add the menu item to set this setting? Or just undocumented for now?
(`local settings =` plural seems like a bad name for that variable :)

frontend/apps/reader/modules/readerlink.lua
```diff
@@ -576,56 +628,60 @@ function ReaderLink:onGotoLink(link, neglect_current_location, allow_footnote_po
...
-        local provider = DocumentRegistry:getProvider(linked_filename)
-        if provider then
-            -- Display filename with anchor or query string, so the user gets
-            -- this information and can manually go to the appropriate place
-            local display_filename = linked_filename
-            if anchor then
-                display_filename = display_filename .. anchor
+                },
+            }
+            if Device:openLink() == nil then
```
poire-z: Calling `Device:openLink()` to know if it's supported seems a bit twisted. It returns `false `when set to `no`, and returns `nil` when it is supported, but provided with some invalid input...
If there's no cleaner way, at least `if Device:openLink() ~= false then` ?

frontend/apps/reader/modules/readerlink.lua
```diff
@@ -576,56 +628,60 @@ function ReaderLink:onGotoLink(link, neglect_current_location, allow_footnote_po
...
-                    end)
-                end
-            })
-        else
-            UIManager:show(InfoMessage:new{
-                text = T(_("Link to unsupported local file:\n%1"), link_url),
-            })
+
+            choose_action = ButtonDialogTitle:new{
+                title = T(_("External link:\n\n'%1'"), link_url),
```
poire-z: Not fond of the url inside `' '` on the screenshots. Can't you just let it plain?

frontend/apps/reader/modules/readerlink.lua
Frenzie: I copied this from ReaderStatus. I wouldn't have put the settings in yet if I'd written it from scratch.

frontend/apps/reader/modules/readerlink.lua
Frenzie: Not defining it as `no` and testing `Device.openLink`?

poire-z: Dunno. Is there already functions not defined in generic/device.lua that are defined in others?
I thought the idea of them being `no` was some kind of API listing everything that can be available.
Or add Device:canOpenLink = yes|no ?

Frenzie: I think it's mainly that you can safely call it and be sure it won't crash. API listing is more like a happy side effect.

Having a separate canOpenLink seems a touch inelegant but I suppose that'll have to be the way forward. :-)

Frenzie:
I added a new menu item to the Links submenu. Looking at the screenshots below I realized it might be redundant with that ignore external links thing.

Screenshot_2019-03-31_17-23-18
Screenshot_2019-03-31_17-23-27

poire-z:
> I added a new menu item to the Links submenu. Looking at the screenshots below I realized it might be redundant with that ignore external links thing.

I think it should be just below `Ignore external links` (as the swipe gestures currently ignore external links).
Or merged with it (but then "Ignore" instead of "Do nothing").
But on my Kobo with wallabag not enabled, it just serves as ignoring Wikipedia article links in Wikipedia EPUB for easier page change with Tap.
And ideally, the menu item should show the action selected, ie: `External links: ignore >` :)

Frenzie:
> I think it should be just below Ignore external links (as the swipe gestures currently ignore external links).

Speaking of which, why is that so immensely complicated?

Frenzie:
Screenshot_2019-03-31_17-51-39
Screenshot_2019-03-31_17-53-11

poire-z:
> why is that so immensely complicated

It's not: you just haven't done it yet, and I just haven't requested yet it to be a toggable :) and probably another one from the Tap one - as it involves a bit of expensive work in crengine, but all the work in crengine/cre.cpp is already done, you just have to make that hardcoded `true` a variable:
https://github.com/koreader/koreader/blob/7396abebf26ecbb3a8cf45f4ee4dcb61cb788287/frontend/apps/reader/modules/readerlink.lua#L740-L747
and a bit of tweak at the end of `ReaderLink:onGoToPageLink()` to not do the xpointer checking when there is none but a `uri`.

Frenzie:
I'm not really sure what you're asking tbh. I don't really understand the rationale behind what that setting does on closer examination. This probably comes back to my lack of understanding for why swipe to follow links is ignoring links.

poire-z:
In Wikipedia EPUB, small footnotes are hard to tap on, and may be sitting just between 2 wikipedia URL links. Chances of taping ot swiping on it would be very low. Having avoid external links available makes that a non-issue.

Frenzie:
I see. :-)

Anyhow, I think this PR now includes a response to all comments? (Other than dynamic menu text.)

poire-z:
May be give that `External link action` some enabled_func like `if self.ui.document.info.has_pages or not isTapIgnoreExternalLinksEnabled`?

Otherwise lgtm.
(Except that now, that menu has 11 items and spans 2 pages :) I'll probably kill "Swipe to follow first link on page" as I don't see why one would use it since we have "Swipe to follow nearest link").

Frenzie:
vcfrtdggggggggg ← that's my cat's response

poire-z:
I've been looking at this to see how we could merge `Ignore external links` into the added `External link action>` submenu (and adding the selected action text into the main menu text).
I wondered if we should keep both "ignore" (let the tap propagate) and "do nothing" (eat the tap, don't propagate).

But after some time, I just wonder if this menu is really useful with all the actions.
I guess most will choose only "ignore" or "ask with popup" - just to see the url before getting into the wallabag wifi popup and download, or launching an external browser. These are just one tap away anyway.

We could be just fine with the existing single checkbox `Ignore external links`: if checked, let the tap propagate to change page or find an internal link near - if unchecked, just show the popup with the available actions?

Frenzie:
I'd never choose "nothing" and I don't think add to Wallabag makes all that much sense as a default, but I can definitely see opting for "open in browser" without disabling the Wallabag plugin.

PS It's only got this configuration because you requested it! :-P

poire-z:
>> Forgot to add the menu item to set this setting? Or just undocumented for now?

> I copied this from ReaderStatus

So, we have that menu because of that setting that just got there because it came along in the copying ? :)

poire-z:
Or if we were to keep it, it may be better to have it as a real submenu with various combinable items, instead of an alternative single choice one.
External links>
- Ignore on tap
- Ignore on swipe
- Open in browser without asking
- Open Wikipedia links also in browser without asking (avoid our preview in textboxwidget if you prefer your browser)

Frenzie:
> So, we have that menu because of that setting that just got there because it came along in the copying ? :)

Well no, but I wouldn't have added a menu for it this weekend because studying the menu and the code related to it is an entirely different task. :-P The open in browser setting is obvious but not terribly relevant on most platforms, the "nothing" setting is the kind of thing you like (and easy), and add to Wallabag is the reason I added this in the first place so it would be odd if it weren't a setting. It's the menu you get if you want a menu now rather than next month. ;-)

> it may be better to have it as a real submenu with various combinable items, instead of an alternative single choice one.

Naturally. ;-)

Issue #2555, opened 2017-02-20
The User Interface…


baskerville:
I'm probably not going to surprise you by saying that *koreader*'s UI isn't the greatest UI I've ever used.

Here's the, non-exhaustive, list of problems / suggested improvements:
- Based the average human finger tip area, some icon bars have a correct height, others don't. More generally, the vertical and horizontal padding used around interactive items is often wrong.
- The font used for the dictionary definition needs to be a serif font.
- The top of the file manager should display the menu icons, not *File Manager* (one useless tap…) and the battery percentage and time should appear at the top right, not at the bottom of menus.

Please checkout this [course on user interface design](http://stellar.mit.edu/S/course/6/sp15/6.813/materials.html).

KenMaltby:
A good deal of that is subjective, and referencing a textbook or other academic source, doesn't actually change that.  There can be a good deal to improve the existing interface, but there is a lot to be liked with it as it is.  While you may feel there shouldn't be a UI that appeals to the more technically literate, I don't believe in such oppression/discrimination.  Not everyone thinks the lowest common denominator should rule.

Luck;
Ken

Frenzie:
> Programmers are often affected by a strange disease: they believe that they don't need to know anything about interface design to design an interface.

Who did you ask? :-)

pazos:
UI isn't the best "feature" of koreader, but I like it 💃 

There is room enough to improve Koreader UI, but IMO is better to open specific issues to fix those one by one.

houqp:
>Programmers are often affected by a strange disease: they believe that they don't need to know anything about interface design to design an interface.

I don't think this is the case for us. We do care about UX here. In fact, I interact a lot of with UX designers at work for my projects.

The real problem here is lack of resources. Right now, I am still working on fixing various stability issues of the reader. The UI framework is lagging behind and resulted in horrible callback hell in the code base. It's also not easy to create widgets with the current UI API. Lacking of proper UX design worries me less compared to a shaky foundation of the code base.

I want a good UX for KOReader as much as you do. If you have time to help work on this, that would be awesome.

>Based the average human finger tip area, some icon bars have a correct height, others don't. More generally, the vertical and horizontal padding used around interactive items is often wrong.

[...]

PR #4867, opened 2019-04-02
Links menu and handling tweaks

poire-z:
Follow up to #4863, see discussion there (so, I removed the External link action menu, and as a compensation, I allowed external links on swipe :).

- Removed "Swipe to follow first link on page" menu item and handling code, as it feels not really as practical as "Swipe to follow nearest link".
- Removed recently added "External link action", as we can just always present a popup with the url and the available actions.
- Generic handling of these actions in onGoToExternalLink(), so they are proposed on the Wikipedia lookup popup too.
- Allow external link on PDF documents (previously, only internal links were handled).
- Have "Ignore external links on tap" available on all document types.
- Added "Ignore external links on swipe" (default to true, the current behaviour).
- Added multiswipe gesture "Follow nearest internal link" (the existing "Follow nearest link" now follows the nearest external or internal link)
- ButtonDialogTitle: added an option to look a bit more alike ConfirmBoxes.
- Footnote popups: fix link unhighlight when tap on external link.

I'm not really fond of the switch for Wikipedia popup from the ConfirmBox & MultiConfirmBox to this ButtonDialogTitle, but I guess I'll get used to it (the removal of the Info icon make more room for the article title, so I guess it's good).

<kbd>image</kbd>

For all Wikipedia or external links, the popup will be the same, with just added buttons depending on what's possible/available:

image

image

image

image

image

image

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -31,6 +31,7 @@ local action_strings = {
     previous_location = _("Back to previous location"),
     latest_bookmark = _("Go to latest bookmark"),
     follow_nearest_link = _("Follow nearest link"),
+    follow_nearest_internal_link = _("Follow nearest internal link"),
     clear_location_history = _("Clear location history"),
 
     toc = _("Table of contents"),
@@ -361,6 +362,7 @@ function ReaderGesture:buildMenu(ges, default)
         {"previous_location", not self.is_docless},
         {"latest_bookmark", not self.is_docless},
         {"follow_nearest_link", not self.is_docless},
+        {"follow_nearest_internal_link", not self.is_docless},
         {"clear_location_history", not self.is_docless, true},
 
         {"folder_up", self.is_docless},
@@ -747,6 +749,8 @@ function ReaderGesture:gestureAction(action, ges)
         self.ui:handleEvent(Event:new("GoToLatestBookmark"))
     elseif action == "follow_nearest_link" then
         self.ui:handleEvent(Event:new("GoToPageLink", ges, false, G_reader_settings:isTrue("footnote_link_in_popup")))
+    elseif action == "follow_nearest_internal_link" then
+        self.ui:handleEvent(Event:new("GoToPageLink", ges, true, G_reader_settings:isTrue("footnote_link_in_popup")))
     elseif action == "clear_location_history" then
         self.ui:handleEvent(Event:new("ClearLocationStack", true)) -- show_notification
     elseif action == "filemanager" then
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -99,12 +99,12 @@ local function isSwipeToGoBackEnabled()
     return G_reader_settings:isTrue("swipe_to_go_back")
 end
 
-local function isSwipeToFollowFirstLinkEnabled()
-    return G_reader_settings:isTrue("swipe_to_follow_first_link")
-end
-
 local function isSwipeToFollowNearestLinkEnabled()
     return G_reader_settings:isTrue("swipe_to_follow_nearest_link")
+end
+
+local function isSwipeIgnoreExternalLinksEnabled()
+    return G_reader_settings:nilOrTrue("swipe_ignore_external_links")
 end
 
 local function isSwipeToJumpToLatestBookmarkEnabled()
@@ -126,51 +126,16 @@ function ReaderLink:addToMainMenu(menu_items)
                 help_text = _([[Tap on links to follow them.]]),
             },
             {
-                text = _("External link action"),
-                enabled_func = function()
-                    return self.ui.document.info.has_pages or not isTapIgnoreExternalLinksEnabled()
+                text = _("Ignore external links on tap"),
+                enabled_func = isTapToFollowLinksOn,
+                checked_func = isTapIgnoreExternalLinksEnabled,
+                callback = function()
+                    G_reader_settings:saveSetting("tap_ignore_external_links",
+                        not isTapIgnoreExternalLinksEnabled())
                 end,
-                sub_item_table = {
-                    {
-                        text = _("Ask with pop-up dialog"),
-                        checked_func = function()
-                            local setting = G_reader_settings:readSetting("external_link_action")
-                            return setting == "pop-up" or setting == nil
-                        end,
-                        callback = function()
-                            G_reader_settings:saveSetting("external_link_action", "pop-up")
-                        end,
-                    },
-                    {
-                        text = _("Do nothing"),
-                        checked_func = function()
-                            return G_reader_settings:readSetting("external_link_action") == "nothing"
-                        end,
-                        callback = function()
-                            G_reader_settings:saveSetting("external_link_action", "nothing")
-                        end,
-                    },
-                    {
-                        text = _("Add to Wallabag"),
-                        checked_func = function()
-                            return G_reader_settings:readSetting("external_link_action") == "add_to_wallabag"
-                        end,
-                        enabled_func = function() return self.ui.wallabag ~= nil end,
-                        callback = function()
-                            G_reader_settings:saveSetting("external_link_action", "add_to_wallabag")
-                        end,
-                    },
-                    {
-                        text = _("Open in browser"),
-                        checked_func = function()
-                            return G_reader_settings:readSetting("external_link_action") == "open_in_browser"
-                        end,
-                        enabled_func = function() return Device:canOpenLink() end,
-                        callback = function()
-                            G_reader_settings:saveSetting("external_link_action", "open_in_browser")
-                        end,
-                    },
-                },
+                help_text = _([[
+Ignore taps on external links. Useful with Wikipedia EPUBs to make page turning easier.
+You can still follow them from the dictionary window or the selection menu after holding on them.]]),
                 separator = true,
             },
             {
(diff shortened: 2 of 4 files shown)


frontend/apps/reader/modules/readerlink.lua
```diff
@@ -895,72 +835,63 @@ function ReaderLink:onGoToPageLink(ges, use_page_first_link, allow_footnote_popu
...
-                            if shortest_dist == nil or segment_dist < shortest_dist then
-                                selected_link = link
-                                shortest_dist = segment_dist
-                                link_is_shortest = true
-                            end
-                            if segment.y1 > segments_max_y then
-                                segments_max_y = segment.y1
-                            end
+            -- link.uri may be an empty string with some invalid links: ignore them
+            if link["section"] or (link["uri"] and link["uri"] ~= "") then
```
Frenzie: Seems easier to just write:
```suggestion
            if link.section or (link.uri and link.uri ~= "") then
```

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -746,7 +746,9 @@ function ReaderGesture:gestureAction(action, ges)
     elseif action == "latest_bookmark" then
         self.ui:handleEvent(Event:new("GoToLatestBookmark"))
     elseif action == "follow_nearest_link" then
-        self.ui:handleEvent(Event:new("GoToPageLink", ges, false, G_reader_settings:isTrue("footnote_link_in_popup")))
+        self.ui:handleEvent(Event:new("GoToPageLink", ges,
+            G_reader_settings:nilOrTrue("swipe_ignore_external_links"),
```
Frenzie: I think defaulting to ignoring is an unintuitive default.
```suggestion
           G_reader_settings:isTrue("swipe_ignore_external_links"),
```

frontend/apps/reader/modules/readerlink.lua
```diff
@@ -99,14 +99,14 @@ local function isSwipeToGoBackEnabled()
...
-local function isSwipeToFollowFirstLinkEnabled()
-    return G_reader_settings:isTrue("swipe_to_follow_first_link")
-end
-
 local function isSwipeToFollowNearestLinkEnabled()
     return G_reader_settings:isTrue("swipe_to_follow_nearest_link")
 end
 
+local function isSwipeIgnoreExternalLinksEnabled()
+    return G_reader_settings:nilOrTrue("swipe_ignore_external_links")
```
Frenzie:
```suggestion
    return G_reader_settings:isTrue("swipe_ignore_external_links")
```

frontend/apps/reader/modules/readergesture.lua
poire-z: I disagree for swipe, which is there for a few years and evoke the physical gesture of turning pages to jump to footnotes and back. Having it suddently throw a popup is not a good thing.
(People like you who would still want that can now activate it - but no one else but you complained it was missing :)

frontend/apps/reader/modules/readerlink.lua
poire-z: I disagree for swipe, which is there for a few years and evoke the physical gesture of turning pages to jump to footnotes and back. Having it suddently throw a popup is not a good thing.
(People like you who would still want that can now activate it - but no one else but you complained it was missing :)

frontend/apps/reader/modules/readergesture.lua
poire-z: Well, from the multiswipe gesture, it could be another default or setting. But I don't feel like adding yet another setting for that. So let's have it the same as in readerlink.

Frenzie: It breaks the concept of links as well as the internal logic for ignoring external links.

For the gesture manager you could just make two actions: the current one that follows all links, and a new one called `follow_nearest_internal_link`.

poire-z: Ok, went with 2 actions, hardcoded false and hardcoded true, not using the single swipe ignore setting.

poire-z:
Would it look a bit better, more balanced, with the text centered (with just adding `title_align = "center"`) ?

image

image

image

image

image

Frenzie:
I don't think it does. Better, that is.

NiLuJe:
I like it for actual titles, not so much for links, FWIW ;).

Frenzie:
It's okay for titles, yes. Not really better or worse. But the link is both more balanced and also *much* worse. ;-)

poire-z:
Been toying with adding some kind of PTF (Poor Text Format) to TextBoxWidget, so we could specify alignment hints on individual lines in text fed to TextBoxWidget.
With some bits like:
```diff
@@ -36,6 +36,8 @@ local TextBoxWidget = InputContainer:new{
     editable = false, -- Editable flag for whether drawing the cursor or not.
     justified = false, -- Should text be justified (spaces widened to fill width)
     alignment = "left", -- or "center", "right"
+    simple_formatting = false, -- Set to true to look for simple formatting hints via chars
+                               -- (do not enable it when editable=true, not tested)

+-- Simple formatting chars (picked from Unicode Private Use area)
+
+-- Line alignment (that's all we have for now)
+TextBoxWidget.LINE_ALIGN_LEFT = "\xEE\x90\x80" -- U+E400
+TextBoxWidget.LINE_ALIGN_CENTER = "\xEE\x90\x81" -- U+E401
+TextBoxWidget.LINE_ALIGN_RIGHT = "\xEE\x90\x82" -- U+E402
+TextBoxWidget.LINE_ALIGN_JUSTIFY = "\xEE\x90\x83" -- U+E403
[+ some added code to deal with them]
```
and:
```diff
-        text = T(_("Would you like to read this Wikipedia %1 article?\n\n%2\n"), wiki_lang:upper(), wiki_page:gsub("_", " "))
+        simple_formatting = true
+        local TextBoxWidget = require("ui/widget/textboxwidget")
+        text = T(_("Would you like to read this Wikipedia %1 article?\n\n%3%2\n"), wiki_lang:upper(), wiki_page:gsub("_", " "), TextBoxWidget.LINE_ALIGN_CENTER)
         if epub_fullpath then
-            text = T("%1\n%2", text, _("This article has previously been saved as EPUB. You may wish to read the saved EPUB instead."))
+            text = T("%1\n%3%2", text, _("This article has previously been saved as EPUB. You may wish to read the saved EPUB instead."), TextBoxWidget.LINE_ALIGN_JUSTIFY)
[...]
     dialog = ButtonDialogTitle:new{
         title = text,
+        simple_formatting = simple_formatting,
         use_info_style = true,
         buttons = buttons,
     }
```

We could get when needed results like this (just an example, we'd probably only use LINE_ALIGN_CENTER on the Wikipedia title in that specific example):

image

But I wonder:
- it sucks a bit to have to use templates to insert them, and how to present/explain them in the localized strings
- could we just use more obvious single ASCII chars to hint that alignment? I don't see anything obvious (`[<=_~`). 
(it has to be single chars for easy parsing in TextBoxWidget)

Frenzie:
The way that looks is not great for localization purposes, to put it almost euphemistically. ;-)

If you want to experiment with typography, why not just use the rich text widget we have at our disposal these days? No centering, just something like a simple `<strong></strong>`, possibly combined with a little `margin:0 1em`. After having achieved a general look and functionality, you can switch to Markdown instead for simpler internationalization.

Also see, e.g., https://github.com/koreader/koreader/issues/2555#issuecomment-282991702 although I think that "Are you *sure* you want to **wipe**" might be going a bit overboard. Mainly I'd just make the first line bold:

Screenshot_2019-04-06_13-03-15

Frenzie:
Link dump:

* https://developer.apple.com/design/human-interface-guidelines/macos/visual-design/typography/
* https://docs.microsoft.com/en-us/windows/uwp/design/style/typography
* https://developer.gnome.org/hig/stable/typography.html.en

Frenzie:
From the Gnome HIG, which aligns with my comment above:

> Avoid the use of italic or oblique faces, as these are visually more complex, and can be distracting.

poire-z:
> The way that looks is not great for localization purposes

That's why I asked for alternative suggestions.
It could come down to,  if using `=` char for centering:
```lua
text = T(_("Would you like to read this Wikipedia %1 article?\n\n=%2\n"),
                   wiki_lang:upper(), wiki_page:gsub("_", " "))
```

Or, if hardcoding the layout:
```lua
text = T(_("Would you like to read this Wikipedia %1 article?"), wiki_lang:upper()) ..
                   "\n\n" .. TextBoxWidget.LINE_ALIGN_CENTER .. wiki_page:gsub("_", " "))
```

> If you want to experiment with typography

I don't :) Just wanted a quick solution to this _not that nice_ display of the Wikipedia article title.
It's just that if they were already some kind of simple embedded line formatting syntax from the 1990s, simpler that Markdow, we could justify supporting that, instead of my unicode private use chars.
But I'm not that fond of my TextBoxWidget tweaks. No problem just forgetting it.

> why not just use the rich text widget

Not that keen on having to bring MuPDF to render simple text messages. And using HTML in the localisation strings is prone to errors, unbalanced tags, readability...

Frenzie:
> Not that keen on having to bring MuPDF to render simple text messages. And using HTML in the localisation strings is prone to errors, unbalanced tags, readability...

I am because I'm not talking about simple, although even merely implementing support for `**strong**` would suboptimally improve most current use cases. It would still be a relatively limited number either way.

You shouldn't use HTML, but Markdown.

## Blockquote
```md
Would you like to read this Wikipedia %1 article?

> %2
```

## Strong
```md
Would you like to read this Wikipedia %1 article?

**%2**
```

## List
```md
Would you like to read this Wikipedia %1 article?

* %2
```

## Some header
```md
Would you like to read this Wikipedia %1 article?

## %2
```

## Pre
````md
Would you like to read this Wikipedia %1 article?

```
%2
```
````

## Also note definition lists ([here](https://talk.commonmark.org/t/description-list/289/12?u=chrisalley), might make it into CommonMark soonish?), although like some of the others mentioned probably not so much for this use case
```md
Would you like to read this Wikipedia %1 article?
:    %2
```

The specifics don't necessarily matter too much. Ideally the element base style suffices, but you can pass along some minor adjustments in the stylesheet.

poire-z:
But you and markdown do not even answer my original wish: text centering :(  :)

Frenzie:
I don't agree that it should be centered. I think it looks unbalanced in combination with a ragged-right edge. NB I don't think "balanced" is a particularly important metric, unless all else is roughly equal.

Weight and size are the best way to improve clarity of the dialogs. However, the above is most definitely the answer to your question and not my vision on what it should look like. See below for that. Because no matter which element you go for, if you tell it to `text-align:center` that's what it'll do. ;-)

I think blockquote or emphasized does a better job.

Screenshot_2019-04-06_17-15-26
Screenshot_2019-04-06_17-18-57

<hr>

Screenshot_2019-04-06_17-14-13
Screenshot_2019-04-06_17-17-27

<hr>

Screenshot_2019-04-06_17-19-58

Issue #4870, opened 2019-04-03
the standby with the cover does not work anymore


davicraquette:
* KOReader version: 2019.3.1-72
* Device:kobo aura one
hello, since this last version the folding of the cover of the e-reader does not put in standby any more. Conversely, the opening of the cover does not activate the reader any more.This problem appeared with version 69 then corrected with version 70 but again in default with the latest version

NiLuJe:
Worksforme on -69.

What is missing is a "This needs a restart" popup on the "Ignore sleepcover events" checkbox I added to the Device menu.
Il'l fix that.

In the meantime, @davicraquette , double-chekc that you don't have that enabled, and restart KOReader ;).

davicraquette:
indeed, and at first, the restart has solved the problem. thank you

NiLuJe:
Fixed via #4871

PR #4482, opened 2019-01-15
Rationalize Links menu items. Supports WTF-8. Internal clipboard

poire-z:
Followup to https://github.com/koreader/koreader/pull/4477#issuecomment-453837780:
- Remove duplicated "Show footnote popup". Have the same setting applied for Tap or 'Swipe to follow nearest link'.
- Make some menu items enabled or disabled depending on if they would have some effect with the current state of other menu items, as some kind of visual self-documentation of these dependancies.
- Add menu item to set the footnote popup font size, relative to the book font size.
- SpinWidget: allow for showing some informative text
- TouchMenu: show checkboxes in grey for disabled menu items

<kbd>image</kbd>

<kbd>image</kbd>
(wording corrections welcome :)

Also includes:
- util.splitToChars(): supports text encoded in WTF-8
WTF-8 is a superset of UTF-8, that includes UTF-16 surrogates in UTF-8 bytes (forbidden in well-formed UTF-8). https://en.wikipedia.org/wiki/UTF-8#WTF-8 https://simonsapin.github.io/wtf-8/
We may get UTF-8 with these from bad producers or converters.
We can get such chars in the text we get from Wikipedia API once their (fully valid) JSON has been decoded by our lpeg-based JSON decoder (which is a defect, hard to fix). (Our other pure-Lua json
decoder has no problem and do that correctly). More details in https://github.com/koreader/crengine/pull/252#issue-244297819
We might also find these WTF-8 in some dictionaries, so let's support them.
- Implement generic internal clipboard for all devices https://github.com/koreader/koreader/issues/4471#issuecomment-453921769
Will allow copying selected text from a book in the TextEditor.
Will be overriden by devices that implement interactions with the system clipboard.

PR #3163, opened 2017-09-02
frontlight on kobo: a few fixes

poire-z:
Closes #2974.
I modified Kobo frontlight code to not update `Kobo eReader.conf` on every frontlight change, and never if `KOBO_SYNC_BRIGHTNESS_WITH_NICKEL = false`.
So, I reintroduced the global settings `frontlight_intensity` and `is_frontlight_on` that were removed in june to keep level and state across koreader sessions.

I hope this fix most of the remaining issues on kobo light, and I think it should answer the unresolved problems in #2974 (and hopefully, won't introduce new ones...)
Also ensure settings are saved when needed.
Also fixed: any settings set while in File Browser were not saved when rebooting/powering off from File browser.

The only changes from the advertised behaviour (in `default.lua`) are:
- setting KOBO_LIGHT_ON_START=-1 saves level and state only in `settings.reader.lua`, and restores light across koreader sessions (dunno how was previous behaviour, as we couldn't guess values for "let light as it"). If there is sense to keep this -1 behaviour, i'll introduce a -3 to save only in koreader settings.
- for people who set KOBO_LIGHT_ON_START to a number >= 0: when suspending and resuming, light is set how it was before suspend (and no more to this KOBO_LIGHT_ON_START value - that seems more natural to me, even if setting a fixed value for light on start seems not :).

I've tested that with different values for the KOBO_* settings, and it works fine.
I haven't tested it on firmware versions that set `FrontLightState` in `Kobo eReader.conf`, but from what I understood, it should work fine.

@Hzj-jie : could you check if these changes are faithful to your work on frontlight from last june ?
@djalma2k @a-kohout : if you're able to backup/patch a PR on your device, could you check if this PR solves your issues (@djalma2k: without and with the `FrontLightState=` you added manually to your `Kobo eReader.conf`)

There's a thing that may be a possible bug (that was not noticed on kobo before this change, probbly because of the way state was saved in nickelconf) in BasePowerd:
https://github.com/koreader/koreader/blob/568e8f667a42ba80dd519c7471e4c4b78b7c58aa/frontend/device/generic/powerd.lua#L130-L132
It looks to me `self:_decideFrontlightState()` should be called after `self:_setIntensity(self.fl_intensity)` so that frontlight state (`is_fl_on`) is accurate if we set some intensity > 0 when we were toggled off.
Could someone on Kindle (@robert00s ?) test the following:
frontlight widget > set some positive level > toggle Off > use +1 or set some positive level from the bar
[...]

PR #3143, opened 2017-08-29
Fix kobo fl_min and frontlightwidget

poire-z:
Been inclined to have a look at frontlightwidget after reading some issues, after i suggesting the fl_min=0 fix (https://github.com/koreader/koreader/issues/3103 https://github.com/koreader/koreader/issues/2970#issuecomment-324973222).
I don't use that widget, and I just fixed what was obvious to me, and for it to somehow behave like the 1st-time user I was would expect :) But maybe I changed some of the original expected behaviour. @robert00s , could you please review this and tell me if you're fine with it? And may be if it works as fine on Kindle (looks like most issues are reported on Kobo, but some should happen on Kindle too).

`fl_min=0` for all other devices, and I see no reason for it to be different on Kobo, and it solves the _1% acting as Off_ bug.
The assumption it should be 1 so that toggling is managed another way is not/no more implemented by current code:
https://github.com/koreader/koreader/blob/efac2721bf40ca17a5e04c9983e86cb28927e8bc/frontend/device/generic/powerd.lua#L38-L39

About frontlightwidget.lua, the main ideas around things I changed are:

- `Min` sets light to 1 (the minimal not-Off light level). Toggle can be used to set it off.
- Setting/reaching 0 does toggle: so, when using `-1` and reaching 0, the previous not-Off level was 1, so `Toggle` puts it back to 1 (when toggling/untoggling, light is restored to previous not-Off level).

Seen no issue after that with the footer frontlight indicator.

PR #4871, opened 2019-04-03
Slightly less crappy Nightmode

NiLuJe:
Companion PR to https://github.com/koreader/koreader-base/pull/884
* Basically flags devices known to be stable when using PxP inversion.
* Plus, random fix for #4870 ;).
* A few FrontLight tweaks & cleanups on Kobo:
  * Moved the Kobo-specific startup status insanity to Kobo-specific init
  * Made turnOff/turnOn frontlight do a smooth ramp down/up
  * On Kobo, use turnOff/turnOn for suspend/resume, to get that smooth toggle
  * On Kobo, for NaturalLight w/ a mixer, only set warmth for setWarmth, and only set Brightness for setBrightness, otherwise, it tried to set both with not in-sync values, which made the FL widget jittery.


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -703,7 +703,7 @@ function ReaderGesture:gestureAction(action, ges)
     elseif action == "toc" then
         self.ui:handleEvent(Event:new("ShowToc"))
     elseif action == "night_mode" then
-        local night_mode = G_reader_settings:readSetting("night_mode") or false
+        local night_mode = G_reader_settings:isTrue("night_mode") or false
```
Frenzie: Redundant or false? ;-)

NiLuJe: Indeed, thanks ;).

NiLuJe:
Okay, added some frontlight tweaks on Kobo, because it's kind of a mess there :D. c.f., the commit message from 5b70446 (although it's only worth reviewing the final state, as, err, things moved a lot after testing on various devices :D).

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -764,7 +764,7 @@ function ReaderHighlight:onHoldRelease()
                         text = "Copy",
                         enabled = Device:hasClipboard(),
                         callback = function()
-                            Device.input.setClipboardText(self.selected_text.text)
+                            Device.input:setClipboardText(self.selected_text.text)
```
poire-z: The clipboard functions are actually functions and not methods, so this one (with parameter) should not work anymore as intended when switched from `.` to `:`. The others won't mind, but I guess they should still use `.`. (cf https://github.com/koreader/koreader/pull/4482#issuecomment-454492160 where I thought about making them methods, but didn't :)

poire-z:
(You're courageous for having a take at that frontlight stuff :) I swear I wouldn't anymore after #3143 and #3163, too fragile ... But great if you make that cleaner :)

Frenzie:
Looks alright at a glance; didn't test; nothing to add to @poire-z :-)

frontend/apps/reader/modules/readerhighlight.lua
NiLuJe: By "others" you mean Clipboard stuff, right? Basically, that only leaves hasClipboard`:` ;).

NiLuJe:
I mainly want to murder the unit test that's (failing) to test a non-existent scenario, but the actual changes look sound ;p.

frontend/apps/reader/modules/readerhighlight.lua
poire-z: Yep, (has|get|set)Clipboard(), that you restored as they were it seems, as they are no more in the diff.
Haven't looked at each of the Device:xxx if they are really a method, but mostly everything should be a method :)

NiLuJe: Yeah, I only rg'd for `evice\.`, so I shouldn't have touched anything else ;).

poire-z:
I'll test all that on my Kobo this evening.
(Looks like 2019.04 is nearly ready for release, so may be wait for it to happen before merging this risky frontlight stuff and new'y HW invert :)

Frenzie:
Yes, I had some time yesterday so I prepped a little release note. I can tag it tomorrow unless there are any urgent remaining issues I'm not aware of?

poire-z:
> unless there are any urgent remaining issues I'm not aware of?

Nope. No real changes since a few days, should be pretty stable :)

poire-z:
> I'll test all that on my Kobo this evening.

No problem noticed so far on my GloHD.
No change noticed with nightmode, but I never really used it.
The progressive off<>on frontlight toggling is nice too (it's smooth when ON>OFF, but when OFF>ON, there is somehow a small pause at ~20% and it feels a bit less smooth - GloHD only?).

NiLuJe:
@poire-z : Noticed that on the H2O, too, so, that'd be all devices using the ioctl (i.e., without NaturalLight).

I'm not sure if it's a quirk of the ioctl itself, or load-related, since it tends to hang just to display the "Frontlight Off" toast.
Which, incidentally, makes it readable, so I'm not *too* bothered by it, but it is weird ;).

NiLuJe:
@poire-z: Yep, it's a hardware quirk.

```C
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/ioctl.h>
#include <sys/stat.h>

#define CM_FRONT_LIGHT_SET 241

int main(int argc, char **argv)
{
	int ntx_io, retval, ptr;
	ntx_io = open("/dev/ntx_io", O_RDONLY | O_NONBLOCK | O_CLOEXEC);

	uint8_t fl_min = 0U;
	uint8_t fl_max = 100U;
	uint8_t fl_cur = 20U;
	// Ramp-down
	for (uint8_t i = 1U; i <= 5; i++) {
		ptr = __builtin_ifloorf(fl_cur - ((fl_cur / 5.0f) * i));
		retval = ioctl(ntx_io, CM_FRONT_LIGHT_SET, ptr);
		printf("ioctl(ntx_io, CM_FRONT_LIGHT_SET) = %d; returned %d\n", retval, ptr);
		if (i < 5) {
			usleep(35 * 1000);
		}
	}

	usleep(1000 * 1000);
	// Ramp-up
	for (uint8_t i = 1U; i <= 5; i++) {
		ptr = __builtin_iceilf(fl_min + ((fl_cur / 5.0f) * i));
		retval = ioctl(ntx_io, CM_FRONT_LIGHT_SET, ptr);
		printf("ioctl(ntx_io, CM_FRONT_LIGHT_SET) = %d; returned %d\n", retval, ptr);
		if (i < 5) {
			usleep(35 * 1000);
		}
	}

	close(ntx_io);
}
```

Behaves the same way ;).

NiLuJe:
@poire-z : Re-reading your message, you're seeing the stutter on the opposite side I do. Huh. (I'm seeing it when shutting the light off, i.e., the ramp down).

PR #4884, opened 2019-04-08
[UX] Add keyboard swipes

Frenzie:
Only the `e` has a few swipes in this initial commit.

Screenshot of west, northwest, north, northeast, east.
Screenshot_2019-04-08_18-14-05


frontend/ui/data/keyboardlayouts/std.lua
```diff
@@ -8,7 +17,7 @@ return {
         {  --  1           2       3       4       5       6       7       8       9       10      11      12
             { "Q",        "q",    "„",    "0",    "Й",    "й",    "?",    "!",    "Å",    "å",    "1",    "ª", },
             { "W",        "w",    "!",    "1",    "Ц",    "ц",    "(",    "1",    "Ä",    "ä",    "2",    "º", },
-            { "E",        "e",    "@",    "2",    "У",    "у",    ")",    "2",    "Ö",    "ö",    "3",    "¡", },
+            { "E",          e,    "@",    "2",    "У",    "у",    ")",    "2",    "Ö",    "ö",    "3",    "¡", },
```
poire-z: or name this local variable (and the future ones) `_e_`, so it's balanced in there, and no risk of confusion (and luacheck complain?) when we use `for k, v in pairs() do` and we implement swipes for the letters k or v ?

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -311,21 +336,31 @@ function VirtualKeyboard:addKeys()
...
         for j = 1, #self.KEYS[i] do
+            local key
+            local key_chars = self.KEYS[i][j][self.keyboard_layout]
+            if type(key_chars) == "table" then
+                key = key_chars[1]
+            else
+                key = key_chars
+                key_chars = nil
+            end
+            logger.dbg(key)
```
poire-z: Volontary addition, or forgotten `logger.dbg()` ?

Frenzie: Overlooked. :-)

frontend/ui/data/keyboardlayouts/std.lua
Frenzie: I'm not entirely sure what you mean. It's drop-in and all that's needed is keyboard updates. ;-)

(Did you say k? I think `ķ` is about the only one possible. But of course such keys would also be good for punctuation characters.)

poire-z: I meant just what you just changed, `local _e_ =` on the first line, so nothing else :)
(this `k` thing was just an example of another common single char local variable; `i` in `for i, v in pairs()` would have been another better example, as we could have `ï` `î` :)

poire-z:
(And for a next PR, a little overlay widget that would show onHold, a 3x3 grid with the chars we'd get with a swipe in these 9 directions ? :)

Frenzie:
@poire-z I was actually thinking of ideally doing a little bit more than that, see my latest commit.

```lua
local _e_ = {
    -- default
    "e",
    -- quick swipes
    north = "ë",
    northeast = "é",
    northwest = "è",
    east = "ê",
    west = "ẽ",
    south = "ę",
    southeast = "€",
    southwest = "ė",
    -- extended hold characters, more than 3 probably not obvious
    "ē",
}
```

That could look something like this:

Screenshot_2019-04-09_10-34-43

But GUI is much harder. :-)

Biep:
Nice!

PR #4886, opened 2019-04-09
[UX] Keyboard character selection popup

Frenzie:
Other than positioning, this is a feature-complete proof of concept. For the moment this PR is mainly to gather potential feedback before/while tidying it all up.

Screenshot_2019-04-09_16-29-30

### DIFF
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -21,6 +21,15 @@ local WidgetContainer = require("ui/widget/container/widgetcontainer")
 local logger = require("logger")
 local Screen = Device.screen
 
+
+
+local VirtualKeyPopup = FocusManager:new{
+    modal = true,
+    disable_double_tap = true,
+    inputbox = nil,
+    layout = {},
+}
+
 local VirtualKey = InputContainer:new{
     key = nil,
     icon = nil,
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -66,6 +66,158 @@ function VirtualKey:init()
...
+            table.insert(vertical_group, horizontal_group_middle)
+            table.insert(vertical_group, horizontal_group_bottom)
+
+    local keyboard_frame = FrameContainer:new{
+        margin = 0,
+        bordersize = Size.border.default,
+        background = Blitbuffer.COLOR_WHITE,
+        radius = 0,
+        padding = self.keyboard.padding,
+        CenterContainer:new{
```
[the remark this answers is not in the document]
NiLuJe: True, but it breaks the swipe paradigm a tiny bit (and, for instance, we do kind of like our `à`'s in french ;)), and has to be layout-specific.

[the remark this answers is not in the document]
Frenzie: I think those swipes are already broken by their location regardless. But, um, everything's layout-specific, so I'm not quite sure what you mean there. ;-)

Light-gray highlight:

Screenshot_2019-04-09_17-50-47

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -66,6 +66,158 @@ function VirtualKey:init()
         self.callback = function() self.keyboard:downLine() end
     else
         self.callback = function () self.keyboard:addChar(self.key) end
+        self.hold_callback = function()
+            if not self.key_chars then return end
+
+            local popup_focus_manager = FocusManager:new{
```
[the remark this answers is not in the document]
poire-z: Yes, just to avoid creating and garbage collecting these few methods functions (even if it's little compared to the other stuff created in there :)

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -66,6 +66,158 @@ function VirtualKey:init()
...
+
+            function popup_focus_manager:onPressKey()
+                self:getFocusItem():handleEvent(Event:new("TapSelect"))
+                return true
+            end
+
+            local key_chars = self.key_chars
+--logger.dbg(key_chars)
+--error()
+            local extra_key_chars = {}
```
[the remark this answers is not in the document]
poire-z: Not sure while just reading the code: do all these keys are also tap'able (so, the whole thing works without swipe)? Or it's just showing what's available?
I guess so, if there are these extra keys.
Do the extra_key row shows when there is none defined for that key?

Also, if this is shown on Hold, you might want to handle HoldPan/HoldPanRelase to support a swipe started from a Hold :)

frontend/ui/widget/virtualkeyboard.lua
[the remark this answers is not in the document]
poire-z: And `ê` should probably be associated to swipe north, as the ^ points up.

frontend/ui/widget/virtualkeyboard.lua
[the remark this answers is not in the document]
Frenzie: These are *only* tappable. I suppose making swiping possible could be an idea.

frontend/ui/widget/virtualkeyboard.lua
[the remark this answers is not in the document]
Frenzie: The acute & grave accents happen to correspond to mnemonic gestures in this layout, but in the English/international keyboard I think the diaeresis makes more sense in the prime location than the more or less French-only circumflex. Besides, most diacritics are on top. ;-)

That being said, I haven't done an actual usage frequency analysis.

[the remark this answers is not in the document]
poire-z: OK, so that will be for the french layout only then.
(It's going to be a bit of a mess to keep the layouts sync'ed...)

Frenzie:
Basic positioning. I'll force it in the frame later tonight or tomorrow. In any case it seems to be about ready for some minor refactoring and actual review.

Screenshot_2019-04-09_19-02-12
Screenshot_2019-04-09_19-02-22

Frenzie:
@poire-z @NiLuJe Alright, ready for actual review.

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -17,9 +17,17 @@ local TextWidget = require("ui/widget/textwidget")
...
 local logger = require("logger")
 local Screen = Device.screen
 
+local VirtualKeyPopup = FocusManager:new{
+    modal = true,
+    disable_double_tap = true,
+    inputbox = nil,
+    layout = {},
+}
+
```
poire-z: Why not move all the `VirtualKeyPopup:` methods here ?
I see there's in one of them a reference to `VirtualKey`, so there may be a circular reference that leads to how you laid them. But it looks like it's defined in a function that is created at runtime, so may be it can be looked up fine if you put them here?

Frenzie: I could feed the VirtualKeyPopup down in VirtualKeyboard so they're both defined properly.

Frenzie: The advantage of VirtualKey:init() is that it has all the modifier etc. logic, so that'd result in some wasted memory on all keys though.

Frenzie:
I'd prefer to keep keep actual key layouts to separate PRs. :-)

frontend/ui/widget/virtualkeyboard.lua
Frenzie: Also there's the backspace thing.

Frenzie: Maybe as an extra property `hold_callback_default`?

poire-z: Not really following you :)
I just meant: move the code (just the lines, no indentation shifts needed) AFTER the `local VirtualKeyPopup = FocusManager:new{}`, not INSIDE :)
Just to group object and method definition in the file.

Frenzie: That can't work because VirtualKey isn't defined yet.

Frenzie: I have no idea what you even mean by inside. Where they were before? :-P

poire-z: Fine with me with latest commit :)

Frenzie: Alright, merging then so nightly users can test.

Biep:
Would inverting (as in night mode) the pop-up keyboard be a good idea?  It would make it stand out more from the permanent keyboard.
Or would having a black-on-white popup at night be too disrupting?

NiLuJe:
It's going to ghost like hell, not a fine idea on e-ink.

Biep:
Ah - hardware limitation.  Waiting for the next generation of e-ink: whiter, faster, and non-ghosting.

PR #4887, opened 2019-04-09
[UX] VirtualKeyPopup: respond to hold & pan release

Frenzie:
As suggested by @poire-z in https://github.com/koreader/koreader/pull/4886#discussion_r273569802

This is actually much better than before.

It subtly changes the behavior of every key this way, but I don't think that's a bad thing.

### DIFF
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -131,12 +131,6 @@ function VirtualKey:init()
                     range = self.dimen,
                 },
             },
-            SwipeKey = {
-                GestureRange:new{
-                    ges = "swipe",
-                    range = self.dimen,
-                },
-            },
             HoldReleaseKey = {
                 GestureRange:new{
                     ges = "hold_release",
@@ -146,6 +140,12 @@ function VirtualKey:init()
             PanReleaseKey = {
                 GestureRange:new{
                     ges = "pan_release",
+                    range = self.dimen,
+                },
+            },
+            SwipeKey = {
+                GestureRange:new{
+                    ges = "swipe",
                     range = self.dimen,
                 },
             },
@@ -318,6 +318,8 @@ function VirtualKeyPopup:init()
 
                 if v == key_char_orig then
                     virtual_key[1].background = Blitbuffer.COLOR_LIGHT_GRAY
+                    virtual_key.onHoldReleaseKey = function() end
+                    virtual_key.onPanReleaseKey = function() end
                 end
 
                 table.insert(group, virtual_key)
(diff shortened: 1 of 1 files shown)


Frenzie:
@poire-z Don't merge, the behavior isn't 100 % right yet.

Frenzie:
I'll go ahead and merge this one 'cause it's such a definite improvement. ;-)

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -298,11 +313,21 @@ function VirtualKeyPopup:init()
...
+
+                    -- restore ability to hold/pan release on central key after opening popup
+                    virtual_key._keyOrigHoldPanHandler = function()
+                        virtual_key.onHoldReleaseKey = virtual_key._onHoldReleaseKey
+                        virtual_key.onPanReleaseKey = virtual_key._onPanReleaseKey
+                    end
+                    virtual_key._onHoldReleaseKey = virtual_key.onHoldReleaseKey
+                    virtual_key.onHoldReleaseKey = virtual_key._keyOrigHoldPanHandler
+                    virtual_key._onPanReleaseKey = virtual_key.onPanReleaseKey
+                    virtual_key.onPanReleaseKey = virtual_key._keyOrigHoldPanHandler
```
poire-z: What does all that allow?
Not really understanding - looks to me you hold on the original keyboard key - and this virtual key (even with the same letter) is another VirtualKey object, so when you release that second one, the first one should still work. (?)

Frenzie: This doesn't allow anything. ;-) It restricts the central key in the popup from responding as if there were an intentional hold release after opening. But after opening, it felt odd when it didn't respond like any other key.

poire-z:
Some small remarks:
- On my Kobo, the popup keys are not vertically aligned with the main keyboard keys, while they are on the emulator. So, may be a missing scaleBySize() somewhere (or just me having missed updating my small private hacks in that file for this change, so please check on your Kobo :)
- With flash keyboard enabled, when taping another key in the popup, the taped key is border-flashed. But when panning and releasing on another key, it's the middle key that is border-flashed, and not the selected other key.

Frenzie:
If by not vertically aligned you mean off by a pixel or two then it's probably just something like a wrong number of subtracted borders or something. If you truly mean not vertically aligned then I'm fairly sure it's just you. ;-)

poire-z:
Yeah, I mean off by two pixels or three :)

Other remarks:
- Dunno if that should happen on popup key tap, but may be on popup key panrelease, the popup could close, as it happens on some(all?) android virtual keyboards - to avoid that additional tap to close it
- Possibly, for the extra keys definitions (`local _e_ =`), it would be nice if they were put into an independant file, so we can import them easily into the other layouts (fr_keyboard.lua), without having to redefine them all (even if of fr, we may tweak a few of them to get `ê` at north). Not sure if there's a way in Lua like in python: `from extra_keys import *`.

PR #4889, opened 2019-04-10
[UX] Keyboard: add A diacritics

Frenzie:
Screenshot_2019-04-10_09-36-49
Screenshot_2019-04-10_09-36-57
Screenshot_2019-04-10_09-37-42


Frenzie:
I'll rework this slightly with the suggestion from https://github.com/koreader/koreader/pull/4887#issuecomment-481571624 to externalize the popup definitions.

poire-z:
Follow up to https://github.com/koreader/koreader/pull/4886#discussion_r273577086:

> The acute & grave accents happen to correspond to mnemonic gestures in this layout, but in the English/international keyboard I think the diaeresis makes more sense in the prime location than the more or less French-only circumflex

Well, I don't know how much you care about that, but not seeing the `â` between the `à` and `á` just look super strange. French or not french, these are accents. The other are dots and waves :)
So, will that really _have to be_ tweaked in the french layout?

Frenzie:
The diaeresis wins two times out of three, and the circumflex never does.

1. Keyboard layout based. The correct order is `èéë`, with `ẽ` west and `ê` east.
2. Gesture based. `è` is northwest, `é` is northeast, and up is a diacritic centered on top. Both `ê` and `ë` are, so neither makes more sense than the other. The `ẽ` is off-balance left, so it should be west.
3. Usage based. North is a preferential gesture. The most prevalent character is `é`, while `ë` is a more prevalent character than `ê`.

1+2 = `èëé`
1+3 = `èéë`

At equal weighting, 1+2+3 = `èéë`! The `ë` is the result of giving double weight to 2, plus a judgment call.

frontend/ui/data/keyboardlayouts/std.lua
```diff
@@ -1,27 +1,9 @@
...
-    northwest = "è",
-    east = "ê",
-    west = "ẽ",
-    south = "ę",
-    southeast = "€",
-    southwest = "ė",
-    "ē",
-}
+local en_popup = require("ui/data/keyboardlayouts/keypopup/en_popup")
+local _at = en_popup._at
```
Frenzie: @poire-z Something like this?

poire-z: Yes. `std_popup.lua` maybe?

Frenzie: Good point, except I'd rename std.lua to en_keyboard.lua. :-P

poire-z:
As a courtesy to technical users, will it be possible to have on a key on the first keyboard layout (`.` or `,`) a popup with useful keys for styletweaks CSS or quick Lua hacking?

I'm thinking about `{ ! ; }` for CSS (I have to hit `Sym` + `Shift` to get the `{` and back or half back to get the `;` and back to get lowercase letters for properties and values, and then again `Sym` + `Shift` to get the `}`).
And possibly `:` and `_` too, and the other parens/brackets if there is room.

Frenzie:
I was planning to stick [ and ] on q and w. 

And yes, the . obviously needs all the regular punctuation options.

tl;dr no worries ;-)

Frenzie:
@poire-z std.lua renamed & ready.

For <kbd>q</kbd> and <kbd>w</kbd>, considering something like this:
* <kbd>q</kbd> = `! [ {`
* <kbd>w</kbd> = `@ ] }`

Frenzie:
For the `.`, possibly something like this:

| - | _ | ¿   |
|---|---|-----|
| / | ( | )   |
| ; | **.** | :   |
| ! | ? | ... |

poire-z:
> For q and w, considering something like this

Try to make it easy to associate it in fr_keyboard with <kbd>a</kbd> and <kbd>z</kbd> then :)
Dunno if <kbd>n</kbd> and <kbd>m</kbd> are more sticky across layout - also, q is stick to the left, so less practical to pan on its left.

> For the ., possibly something like this:

For the dot at the bottom (as for the q at the left), it will be impossible to swipe down (resp left), and holdpanrelease left - so the chars on these sides will work only with tap.
So, unless you think about a solution to make these work, try to put the most interesting chars on the other sides.

In both your suggestions, <kbd>!</kbd> is the loser :)

Frenzie:
Anyway, I haven't really thought about those yet. Only the s has received a reasonable amount of thought:

Screenshot_2019-04-10_12-39-13

(I know, sizing isn't quite optimal there. One thing at a time.)

> it will be impossible to swipe down (resp left), and holdpanrelease left

On holdrelease it taps the button you're on; it doesn't care about direction (which would be unintuitive). Only swipes swipe.

Frenzie:
I could fill up extra with sigma (`Σ σ/ς`), but that aside. I'm sure the basic problem will necessarily pop up elsewhere regarless of the <kbd>s</kbd>. ;-)

Biep:
I just updated my Kobo Aura One to v2019.04-9-gfd50dc3_2019-04-10, and now choosing "Book status" in the hamburger menu has KOReader exit irregularly (as reported by KSM).

Frenzie:
I'm not really sure why you'd randomly post that here instead of in a new issue?

Biep:
It seemed connected - as the Book status has a keyboard (in fact I went there in order to try your enhancements), and it seemed not unlikely the bug was related.
But you are right, of course.
(BTW, it also happens when enterering the Book status when reaching the end of a document.)

Frenzie:
Could you open a new issue with the crash.log included?

Biep:
I just did.

Biep:
Could you add the following to the = key, please?

« ≠ »
< = >
<= ≈ >=
(bottom corners should be less-or-equal and greater-or-equal).
Thank you!

And are there gestures associated with the two characters on the top line of a 4-line layout?  Such as a hacek?

Frenzie:
Btw, the code behind the popup itself is somewhat complex, but I made the configuration very friendly.

```lua
    _z_ = {
        "z",
        northeast = "ź",
        northwest = "ζ", -- zeta lowercase
        west = "ž",
        south = "ʐ", -- voiced retroflex sibilant fricative IPA
        southeast = "ʒ", -- ezh, voiced palato-alveolar fricative IPA
        southwest = "ż",
    },
```

By <= you mean ≤ and ≥?

Biep:
Yes - I updated the comment.

Biep:
BTW, would it be possible to add arrow keys this way too?  That swiping left or right on, say, the space bar, would move the cursor in that direction?  If so, then NW and NE could mean delete left and delete right, respectively, freeing a key on the main keyboard.  (The delete key(s) could still, with the arrow keys, be on a secondary keyboard.)

PR #4890, opened 2019-04-10
[UX] Keyboard: add S diacritics

Frenzie:
This one's a bit more difficult to reason about than the a and e because there isn't really much of a standard. Only <kbd>ß</kbd> is easily accessible on your regular international QWERTY layout.

The $ is by analogy with the €, same for a few of the other ones.

Not sure if the long s (`ſ`) should be in both layouts (if it's included) even though S is the uppercase of both s and ſ. Sigma has similar issues.


NiLuJe:
Isn't there a dedicated codepoint for a lowercase eszett?

U+00DF ß LATIN SMALL LETTER SHARP S)

U+1E9E ẞ LATIN CAPITAL LETTER SHARP S

(On my phone, so can't really tell them apart ;p).

Frenzie:
I'll be damned! I didn't even realize there was such a thing as an uppercase eszett.

On your regular QWERTY, it's:

* Alt gr + s = ß
* Alt gr + Shift + s = §

* uppercase: TEẞT (looks *atrocious* here on GH in my browser btw, clearly a glyph borrowed from another typeface)
* lowercase: TEßT

https://en.wikipedia.org/wiki/Capital_%E1%BA%9E

So yes, updating the uppercase ß to ẞ.

NiLuJe:
That's seven years of German and a weird quirk for typography for you ;p.

Frenzie:
You were taught about the intricacies of how the uppercase ẞ may look ever so slightly more B-ish (or fatter) than the lowercase ß when it doesn't look exactly the same? :-P

More seriously though, all (?) German books I have spell it `SS` in capitalized form. Of course ALL CAPS isn't exactly a frequent occurrence, and words don't start with ß…

Frenzie:
Actually I wasn't able to find any with a quick look. These are in a special title face, but they aren't capitalized. I might be thinking of something in the Insel Bucherei, none of which I own. (Some of the older ones of those are on Archive.org btw.)

NiLuJe:
I don't actually remember how it came up, I just remembered there were debates about it and casing ;). According to the Wikipedia article, the uppercase variant has only been "standardized" very recently ;).

NiLuJe:
You sometimes gets caught in weird conversations in university halls, so, who knows ;p.

Frenzie:
Oh right, at a university that's definitely a lot more likely to come up, at least in passing.

PR #4891, opened 2019-04-10
[UX] VirtualKeyPopup: close on hold release

Frenzie:
As suggested by @poire-z.

### DIFF
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -366,7 +366,7 @@ function VirtualKeyPopup:init()
         CenterContainer:new{
             dimen = Geom:new{
                 w = parent_key.width*3 - 2*Size.border.default + 4*parent_key.keyboard.key_padding,
-                h = parent_key.height*4 - 2*Size.border.default + 6*parent_key.keyboard.key_padding,
+                h = parent_key.height*4 - 2*Size.border.default + 5*parent_key.keyboard.key_padding,
             },
             vertical_group,
         }
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -183,14 +183,16 @@ function VirtualKey:onUnfocus()
     self[1].inner_bordersize = 0
 end
 
-function VirtualKey:onTapSelect()
+function VirtualKey:onTapSelect(skip_flash)
     if self.flash_keyboard and not self.skiptap then
         self[1].inner_bordersize = self.focused_bordersize
         self:update_keyboard(false, true)
         if self.callback then
             self.callback()
         end
-        UIManager:tickAfterNext(function() self:invert(false) end)
+        if not skip_flash then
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -183,14 +183,16 @@ function VirtualKey:onUnfocus()
...
-function VirtualKey:onTapSelect()
+function VirtualKey:onTapSelect(skip_flash)
     if self.flash_keyboard and not self.skiptap then
         self[1].inner_bordersize = self.focused_bordersize
         self:update_keyboard(false, true)
         if self.callback then
             self.callback()
         end
-        UIManager:tickAfterNext(function() self:invert(false) end)
+        if not skip_flash then
```
Frenzie: Seeing this in the diff made me realize it should just be part of the top condition.

The flashing behavior of this whole popup stuff isn't quite optimized in any case, with the whole keyboard refreshing and all.

poire-z:
There seems to be a refresh issue with current master, when pan'releas'ing:

image

With flash keyboard enabled and hitting where the <kbd>s</kbd> is hidden:

image

It's ok when tapping, and tapping outside to dismiss the popup.

Frenzie:
That's on your Kobo?

poire-z:
No, on the emulator, after just a `git pull`.

Frenzie:
Lightbulb thought: do you have keyboard flash disabled?

Frenzie:
And if so, does the setDirty from https://github.com/koreader/koreader/pull/4891/commits/4c4efca2879d85dc2c2045dfea25cd581e52a27d do the trick?

```lua
function VirtualKeyPopup:onClose()
    local dimen = self.dimen
    UIManager:close(self)
        return "partial", dimen
    end)
    return true
end
```

poire-z:
Usually I have it disabled, but I tried both (with flash enabled, the <kbd>s</kbd> is shown on one of the screenshots above).

> And if so, does the setDirty from 4c4efca do the trick?

Nope (just cut and pasted that at the bottom of virtualkeyboard.lua, no change).

(Sorry, leaving, can't test much till later this evening.)

poire-z:
It looks like we (logically, as most widgets do) need to have this to avoid this refresh issue:

```lua
+function VirtualKeyPopup:onCloseWidget()
+    UIManager:setDirty(nil, function()
+        return "partial", self[1][1].dimen
+    end)
+end
```
(which is called by UIManager when we do `UIManager:close(self)`).

Without that, when tapping outside, we get a full screen refresh (not sure what triggers it):
```
04/10/19-22:53:16 DEBUG no refresh got enqueued. Will do a partial full screen refresh, which might be inefficient
04/10/19-22:53:16 DEBUG _refresh: Enqueued partial update for region 0 0 600 610
```
and nothing when pan'releas'ing.

With that, when tapping outside, we just get the popup region refreshed:
```
04/10/19-23:02:20 DEBUG _refresh: Enqueued partial update for region 58 287 186 217
```
and the same when pan'releas'ing.

It's strange that you don't reproduce it.

PR #4893, opened 2019-04-10
[UX] VirtualKeyPopup: add B, comma, and period popups

Frenzie:
Adds B because it's on the same row as a comma.

Cf. <https://github.com/koreader/koreader/pull/4889#issuecomment-481624588>.

Will look good with <https://github.com/koreader/koreader/pull/4892>.


poire-z:
Can't we have `[ ] { }` (so we get them all on these 2 keys), instead of these rarely used `‖ ¦ ¿` (and `~` if we need to kill a 4th?)

Frenzie:
They have to be on other keys regardless (like q & w), but they could be doubled I suppose.

Issue #3998, opened 2018-06-05
Enhancement to keyboard layout system


ersi-dnd:
Ability to keep adding layouts indefinitely is a must for users who read in many languages. Looks like the current keyboard layout system cannot do it. Correct me if I am wrong.

When reading in many languages, the user may require several dictionaries and keep adding them as needed to no upper limit. Similarly to search inside books of any arbitrary language or to perform searches in the file browser for filenames/metadata in any arbitrary language (think different alphabets like Cyrillic, Greek, Armenian, Hebrew, Arabic, Ethiopian, Chinese...). 

The keyboard layout cannot be tied to the app interface language. These things are completely unrelated. There should be no limit to the number of possible layouts. Start with a few, yes, but the user should be able to keep adding them as needed without losing the first ones.

Biep:
Another thing I'd like a lot is keyboard gestures.  Starting on a vowel and moving NW would give it an acute accent, and moving NE a grave accent.  Long pressing a key would surround it with the eight symbols associated with the eight main directions - and those symbols should be clickable.  The idea is that in the beginning a long press will bring up these options, but soon one will learn to use the gestures instead.
(Multigestures would be possible, but I can't think of a nice way to show those when long-pressing.)
@Frenzie  @poire-z

Frenzie:
Display would take a fair bit of effort (I mean, it's not a novel idea ;-) ), but just adding some swipes might be somewhat easier.

Biep:
OK, consider it two separate proposals.. :-)
Not all gestures would be obvious, but some, such as the acute and grave accents, or a down swipe for cedilla, could be.  If you want to go wild, multiswipes could indicate other alphabets (as in e.g. swipe left to indicate alphabet change, and then up for Cyrillic, down for Hebrew, right for Greek).  With some reasonable mapping that could become fairly mnemonic too.

Frenzie:
Implemented by #4884, #4886 and #4887.

PR #4898, opened 2019-04-11
[UX] Keyboard: add d diacritics

Frenzie:
Eth (_ð_) is an important letter for people who like Old English (me), and for IPA. It's debatable whether it should go on the _t_ instead, at least from the modern _th_ spelling perspective, but since you already get symbols like _þ_ and _θ_ there (in addition to various native t diacritics) it makes more sense here. Besides which, the eth *is* a modified d.

_Ď_/_ď_ is Czech/Slovak.

_Đ_/_đ_ is sometimes an alternative rendering of _ð_, maybe because you could type it on a traditional typewriter. It's also Slavic and something Vietnamese.

Screenshot_2019-04-11_12-14-00
Screenshot_2019-04-11_12-14-31


poire-z:
(No opinion, for lack of foreign alphabets culture :| so go ahead, we'll meet again when you do `ç` and `ô` :)

Frenzie:
I'm not working on any other keys atm. You could do c (or any other key) already. I was thinking of looking into z or r next, so it'd probably be best to pick one of the other options, including c. ;-)

I would make i and o more or less the same as e. The arguments don't really change I'm afraid. 

Alt gr op = óö, so a traditionalist layout would be òóö. This is perhaps not unsurprisingly quite similar to e and a. O with diaeresis is widely used, o circumflex less so, even though it has more friends outside of French.

Btw, you could also consider a circumflex (and caron) multiswipe, come to think of it.

Alt gr ji = ïí, so a traditionalist layout might favor west or southwest for this outlier, although east (") also has its claims. Here we have an argument of internal consistency with the other vowels because we may be inspired by QWERTY, but we don't follow it too tightly because this is a different input paradigm.

We could always consider swapping the circumflex and the diaeresis if it somehow seems unnatural. Technically only the ä is truly north (alt gr + _q_), while _ë_ and _ö_ are actually east. I was thinking of _é_ and _ë_ as being above (north of) _e_, but actually _é_ is right on top of _e_ and _ë_ lies to the east of it. As long as they are easily accessible through a swipe it's not like it matters too much, besides attempting to attain some fragile sense of familiarity and intuitiveness.

Frenzie:
I'm not working on any other keys atm. You could do c (or any other key) already. I was thinking of looking into z or r next, so it'd probably be best to pick one of the other options, including c. ;-)

I would make i and o more or less the same as e. The arguments don't really change I'm afraid. 

Alt gr op = óö, so a traditionalist layout would be òóö. This is perhaps not unsurprisingly quite similar to e and a. O with diaeresis is widely used, o circumflex less so, even though it has more friends outside of French.

Btw, you could also consider a circumflex (and caron) multiswipe, come to think of it.

Alt gr ji = ïí, so a traditionalist layout might favor west or southwest for this outlier, although east (") also has its claims. Here we have an argument of internal consistency with the other vowels because we may be inspired by QWERTY, but we don't follow it too tightly because this is a different input paradigm.

We could always consider swapping the circumflex and the diaeresis if it somehow seems unnatural. Technically only the ä is truly north (alt gr + _q_), while _ë_ and _ö_ are actually east. I was thinking of _é_ and _ë_ as being above (north of) _e_, but actually _é_ is right on top of _e_ and _ë_ lies to the east of it. As long as they are easily accessible through a swipe it's not like it matters too much, besides attempting to attain some fragile sense of familiarity and intuitiveness.

Frenzie:
I'm not working on any other keys atm. You could do c (or any other key) already. I was thinking of looking into z or r next, so it'd probably be best to pick one of the other options, including c. ;-)

I would make i and o more or less the same as e. The arguments don't really change I'm afraid. 

Alt gr op = óö, so a traditionalist layout would be òóö. This is perhaps not unsurprisingly quite similar to e and a. O with diaeresis is widely used, o circumflex less so, even though it has more friends outside of French.

Btw, you could also consider a circumflex (and caron) multiswipe, come to think of it.

Alt gr ji = ïí, so a traditionalist layout might favor west or southwest for this outlier, although east (") also has its claims. Here we have an argument of internal consistency with the other vowels because we may be inspired by QWERTY, but we don't follow it too tightly because this is a different input paradigm.

We could always consider swapping the circumflex and the diaeresis if it somehow seems unnatural. Technically only the ä is truly north (alt gr + _q_), while _ë_ and _ö_ are actually east. I was thinking of _é_ and _ë_ as being above (north of) _e_, but actually _é_ is right on top of _e_ and _ë_ lies to the east of it. As long as they are easily accessible through a swipe it's not like it matters too much, besides attempting to attain some fragile sense of familiarity and intuitiveness.

Frenzie:
I'm not working on any other keys atm. You could do c (or any other key) already. I was thinking of looking into z or r next, so it'd probably be best to pick one of the other options, including c. ;-)

I would make i and o more or less the same as e. The arguments don't really change I'm afraid. 

Alt gr op = óö, so a traditionalist layout would be òóö. This is perhaps not unsurprisingly quite similar to e and a. O with diaeresis is widely used, o circumflex less so, even though it has more friends outside of French.

Btw, you could also consider a circumflex (and caron) multiswipe, come to think of it.

Alt gr ji = ïí, so a traditionalist layout might favor west or southwest for this outlier, although east (") also has its claims. Here we have an argument of internal consistency with the other vowels because we may be inspired by QWERTY, but we don't follow it too tightly because this is a different input paradigm.

We could always consider swapping the circumflex and the diaeresis if it somehow seems unnatural. Technically only the ä is truly north (alt gr + _q_), while _ë_ and _ö_ are actually east. I was thinking of _é_ and _ë_ as being above (north of) _e_, but actually _é_ is right on top of _e_ and _ë_ lies to the east of it. As long as they are easily accessible through a swipe it's not like it matters too much, besides attempting to attain some fragile sense of familiarity and intuitiveness.

poire-z:
> You could do c 

Well, I won't :) First because I don't have the culture related to all that. And second, I think for consistency, it should be done by one person/mindset :)

> We could always consider swapping the circumflex and the diaeresis if it somehow seems unnatural.

Well, I don't think I'll mind your choice in the end. I'm not even sure I'll really want to put `ê` north in french (@NiLuJe should we?).
Mostly to avoid hacks (and maintenance) across keyboard layouts, and because the grid display and pan'release are good enough, (and because I don't really type much on KOReader :) the brackets and `;!` are the big gain for me).

I'll wait for these to be done to integrate all the `local _x_` into fr_keyboard. Do you feel you can add now empty such local variables that you foresee having keys in the future? So we can make them available already in a few other keyboard layouts than english?

(I use the English UI/translations, but I do prefer the french keyboard. I wouldn't mind having another UI submenu to select the keyboard, or at least a manual setting to set in settings.reader.lua to force it - I currently patch virtualkeyboard.lua in my private patch for that.)

Frenzie:
> Do you feel you can add now empty such local variables that you foresee having keys in the future? So we can make them available already in a few other keyboard layouts than english?

Even on characters without too many obvious diacritics like x, I'm probably still inclined to add characters like χ (voiceless uvular fricative). On the other hand, for optimum IPA input one would want a dedicated layout, which isn't of interest to me on my ereader. ;-)

http://ipa.typeit.org/
http://westonruter.github.io/ipa-chart/keyboard/

* All of the vowels will have diacritics.
* Most consonants as well, but possibly not **p; f, h, j; v**. All the rest most likely will. (Of course, there's always possibilities like π for p and φ for f, not to mention that p could be a good place for additional characters like <kbd>+</kbd>.)
* M will at the very least have µ, and since I'm doing a bit of an IPA thing probably also ɱ.

> (I use the English UI/translations, but I do prefer the french keyboard. I wouldn't mind having another UI submenu to select the keyboard, or at least a manual setting to set in settings.reader.lua to force it - I currently patch virtualkeyboard.lua in my private patch for that.)

In fact we need a selection of keyboards (much like the dictionaries) and an easy way to toggle between them. That's #3998.

A common way of toggling is left/right swipe on the spacebar.

An alternative option is a dedicated button, which can combine itself with the popup.

NiLuJe:
@poire-z: Yeah, that was lost in the fast-moving discussion, but I +1'ed @Frenzie argument for the diacritics placement, it made sense, and I think in practice it'll do fine, even if I kind of had the same reaction as you when first seeing the screenshots ;).

@Frenzie: I am always going to like having IPA stuff in there, even if I'm probably never going to actually use them xD.

In particular, since you've put ð on d, I wouldn't mind θ on t ;).

Frenzie:
Yes, I fully intend to stick that as well as þ on t.

PR #4901, opened 2019-04-11
Update AutoWarmth state on resume

### LINKED ISSUE #4895: Natural light: previous session's light for half a second when starting device
* KOReader version: v2019.04-5-g69e3830_2019-04-09/  
* Device: KA1

#### Issue
It has been like this since some days (or maybe weeks).

I use the auto warmth feature. 

When I start the device nowadays, it seems like it has the previous session's level of natural light for maybe a half second before it finds the correct light.

So for example: if I put the device to sleep in the evening when I have a high level of warmth light, and then start the device in the morning (when the warmth should be 0), it has yesterdays light for like half a second before it jumps to the correct non-warmth light.

NiLuJe:
Regression since #4871
Fix #4895

### DIFF
--- a/frontend/device/cervantes/device.lua
+++ b/frontend/device/cervantes/device.lua
@@ -50,6 +50,7 @@ local Cervantes = Generic:new{
 
     -- currently only Cervantes 4 has coloured frontlight
     hasNaturalLight = no,
+    hasNaturalLightMixer = no,
 }
 -- Cervantes Touch
 local CervantesTouch = Cervantes:new{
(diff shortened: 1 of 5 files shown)


NiLuJe:
Hmm, potentially incomplete, because sysfs_light isn't in sync w/ powerd

NiLuJe:
I don't use the feature, so I can't definitely say if this actually fixes the bug, but at least it doesn't crash ;p.

NiLuJe:
That final commit is a bit hackish, but I really didn't feel like adding yet another has* field to device. Thoughts?

Frenzie:
Yeah, I've faced the same dilemma a few times. This one seems reasonable enough.

NiLuJe:
I might actually need it somewhere else to properly fix this on devices with a mixe,r though, so, stay tuned... :).

NiLuJe:
There, that should do it ;).

PR #4903, opened 2019-04-11
Go back to "UI" refreshes in the FM

NiLuJe:
Going with partial ends up being problematic on REAGL devices in a
number of different ways.

On Kindle (at least pre EPDCv2), it's slightly flickery because of the
whitespace and the boldface (I assume), probably combined with the need to make it FULL.
On Kobo, it's slow when anything else tries to queue an update (i.e.,
flashing the chevrons).
Which is a non-issue on the Forma because then it's fast when using
pageturns to navigate, but, still :D.

The obvious downside is we lose the automatic flash after n refresh,
which is potentially annoying on older, non-reagl devices.
It's not really an issue on REAGL devices, especially on EPDCv2.

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -160,7 +160,7 @@ function FileManager:init()
     function file_chooser:onPathChanged(path)  -- luacheck: ignore
         FileManager.instance.path_text:setText(truncatePath(filemanagerutil.abbreviate(path)))
         UIManager:setDirty(FileManager.instance, function()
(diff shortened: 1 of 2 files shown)


NiLuJe:
That (very noticeable) discrepancy in latency/interactivity between using the chevrons vs. the pageturn buttons was basically driving me nuts :D.

And it's genuinely slightly quirky to look at on a PW2.

NiLuJe:
AUTO *might* actually be a *tad* slower than GLR16 on the Forma, but I'll gladly take it because it's now consistent, no matter how you navigate.

PR #4244, opened 2018-09-24
add spanish keyboard with Ñ character

pazos:
As suggested in https://www.mobileread.com/forums/showthread.php?t=309885  by u/djalma.

This adds a new key in the 2nd row. The same character is used when using key modifiers, to avoid displaying an empty key.

This also translates the word "space" for its spanish equivalent "espacio".

PR #4906, opened 2019-04-12
Extend FR and ES keyboards from the EN keyboard

poire-z:
Avoid duplication with the EN keyboard, as most is the same, and just tweak what has to be changed.
Will allow staying in sync with the EN keyboard recent and futur developments of key popups.

Pinging @pazos (#4244) for info, about the ES keyboard: i updated it because there were so little differences from the EN one - it should be the same as before, but it will be easier to tweak if you feel like removing "espacio" (like I did for the FR one) or re-order the key popups (like I did for the FR ones).

The other keyboard layouts are quite a bit more complicated, but some only have changes in the 3,4,5,6th layouts, so they could probably be tweaked with the same kind of remove/insert/replace tweaks if their users feel like enjoying the new popups.


Frenzie:
It does make it harder to grasp what's going on. I thought you wanted to explicitly reuse the popup variables rather than the whole keyboard. I suppose it shouldn't matter too much. People who want to customize can just base it on the English one.

Incidentally, shouldn't the French layout look more like this? I've never looked at the French keyboard before, but the description of AZERTY as QWERTY with four characters swapped struck me as slightly odd. ;-)

Screenshot_2019-04-12_15-14-29

poire-z:
> It does make it harder to grasp what's going on.

I don't think so. The stuff that changes and why is clearly stated in the comments.
A lot easier, vs the time I spent comparing them to the old `std.lua`.

> I thought you wanted to explicitly reuse the popup variables rather than the whole keyboard.

I thought I would have to do that, and to update it each time you add a popup for a key.
With this, I won't have to (unless some stuff really needs to be adapted).

> Incidentally, shouldn't the French layout look more like this? 

Yes, `M` is just right of the `L`, but it looks like this has never bothered me :) The real pain is AZQW.
And moving the M would require a bit of thinking about what to do with the punctuation and the other layouts, and the popups. So, good enough as it is.

poire-z:
> And moving the M would require a bit of thinking about what to do with the punctuation and the other layouts, and the popups. 

(ok, _bit of thinking_ happening right now)

poire-z:
OK, it will now display as on your screenshot, with some possibly useful chars on the new key in the other modes (except for the russian-like modes where I had no inspiration).

At least, what I thought would be tedious was actually easy with this extend/move/update way of patching the EN keyboard :)

frontend/ui/data/keyboardlayouts/fr_keyboard.lua
```diff
@@ -1,70 +1,50 @@
...
+    keys[1][3][2], -- e
+    keys[1][7][1], -- U
+    keys[1][7][2], -- u
+    keys[1][8][1], -- I
+    keys[1][8][2], -- i
+    keys[1][9][1], -- O
+    keys[1][9][2], -- o
 }
+for _, popup in ipairs(popups) do
+    if popup.east and popup.north then -- not all implemented yet
```
Frenzie: We can just add a couple of stubs instead? It's the other options that require thought but the basic òöóô-type ones should be fine.

Frenzie: PS I'll do that now. Shouldn't take more than a few mins.

poire-z: Dunno (yet) what you mean by stub, but I'll see. But may be no need to add anything.

Anyway, I planned to remove that once you have implemented them for all.
So, no real need for anything more.

(And anyway, it's not even needed (and a bit wrong with my intent) : `local toto = "zzz" ; toto.north` returns nil, when I thought it would crash with "attempt to index a string" - but no - and if it's not a table it's a no-op)

Frenzie: Alright, so let's take out the if condition and merge it. ;-)

poire-z: Alright. (so, UIO are done :)

Frenzie: Just a stub; it'd require some more research to call it done. :-)

PR #4907, opened 2019-04-12
[UX] Keyboard popups: stubs for the rest of the vowels (I, O, U)

Frenzie:
With the basic pattern of diacritics and special characters established by now, the basic diacritics + IPA don't need too much thought for an initial draft.

Cf. https://github.com/koreader/koreader/pull/4906#discussion_r274966796


gerhaher:
You have missed to include the norwegian/danish Ø.
It represents the same sound as the swedish Ö.

Frenzie:
The southwest is reserved for it and it's the IPA character for the same sound. :-) But indeed, I might as well add it directly.

PR #4908, opened 2019-04-12
[UX] Keyboard popup: k and l diacritics

Frenzie:
There doesn't seem to be much going on in k-land, and I like using proper hyphens and dashes.


poire-z:
Why not gather the < 8 chars for `K` in a 3x3 grid? Would look less strange than this 3x4 grid with holes, no? Or the extra upper row has some meaning I haven't guessed yet ? :)

Frenzie:
In this case I think the lowercase kappa makes it clear what the uppercase kappa is, plus it's the same as lambda. I think if it were moved to the top left (for example), it should probably be properly treated as lower/uppercase instead. But true, even if ⓚ is added it'd still all fit on the grid. It's more research pending empty spots than necessarily meaningful extra row spots. ;-)

Issue #4902, opened 2019-04-11
'Go to letter' not working when opening with long tap


gerhaher:
* KOReader version: v2019.04-5-g69e3830_2019-04-09/
* Device: KA1

#### Issue
'Go to letter' doesn't work for me if I open it with a long tap. When I choose a letter and tap 'Go to letter' I'm left on same page.

If I open it with a multiswipe it works fine.

'Go to page' also works fine, irrespective of opening method.

Frenzie:
It's probably working alright, but due to an unintended side effect of #4887 you'd be inserting a space first. (Although arguably whitespace should be stripped.)

gerhaher:
> It's probably working alright, but due to an unintended side effect of #4887 you'd be inserting a space first. (Although arguably whitespace should be stripped.)

Yeah, I see.

NiLuJe:
I've sometimes had it not actually moving pages at all despite a sane-looking output, and that for a good long while, so, there may be something slightly wonky somewhere nonetheless ;).

Frenzie:
It seems to work by filename, while the cover browser displays metadata title or filename.

Potential workaround concept that isn't quite satisfactory:
```diff
diff --git a/frontend/ui/widget/virtualkeyboard.lua b/frontend/ui/widget/virtualkeyboard.lua
index d170027c..ccbe52cc 100644
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -232,7 +232,13 @@ function VirtualKey:onSwipeKey(arg, ges)
     return true
 end
 
-VirtualKey.onHoldReleaseKey = VirtualKey.onTapSelect
+function VirtualKey:onHoldReleaseKey()
+    if self.keyboard.just_opened then
+        self.keyboard.just_opened = false
+        return true
+    end
+    self:onTapSelect()
+end
 VirtualKey.onPanReleaseKey = VirtualKey.onTapSelect
 
 function VirtualKey:invert(invert, hold)
@@ -551,6 +557,7 @@ function VirtualKeyboard:_refresh(want_flash)
 end
 
 function VirtualKeyboard:onShow()
+    self.just_opened = true
     self:_refresh(true)
     return true
 end
```

gerhaher:
And since we are talking about 'Go to letter' in this thread: "Äéß-keyboard"´s letters are case sensitive for this plugin. 

Is there any possibility to make them insensitive?

[...]

PR #4909, opened 2019-04-12
[fix] widget/menu: go to letter - strip whitespace

Frenzie:
Just because it seems like a good idea, not because it happens to work around <https://github.com/koreader/koreader/issues/4902> as a side effect.


poire-z:
Any whitespace, even in the middle?
(If looking for "The revenants" and entering "The r", we would skip it and go to "There were warriors" ?)

Frenzie:
True, it's called go to letter but the code actually does go to search string.

Frenzie:
So never mind, I suppose. :-)

poire-z:
You could have just removed leading spaces? Guess it's rare one has files starting with spaces.

Frenzie:
Sure, that'd just be a `^`. But those do occur.

PR #4912, opened 2019-04-13
[UX] Keyboard popup: add C diacritics

Frenzie:
Screenshot_2019-04-13_10-24-27
Screenshot_2019-04-13_10-24-41


Frenzie:
Btw, apparently there are dedicated symbols for °C and °F. (But no, I just used ° separately from C and F there because I wouldn't know how to type them.)

PR #4921, opened 2019-04-14
Unbreak FL widget on !NTX devices

### LINKED ISSUE #4919: Brightness in Settings close Koreader
Device:Kindle PW1
Koreader version:v2019.04-25-g06a73ba_2019-04-13
I can change brightness by gentures. But Brightness in Settings close Koreader. Whis is a bug or feature?

NiLuJe:
Those caps are NTX only

Fix #4919

Regression since #4901


NiLuJe:
That's what I was wondering, too.

For reference, that device check was how it used to, and what I broke in #4901 

This isn't used in many places, IIRC, but making it safe to use without having to think about that sounds like a Good Plan regardless ;).

Frenzie:
@NiLuJe Kinda off-topic, but just a quick note may suffice. The frontlight keeps turning on after standby since your recent changes on my H2O.

NiLuJe:
@Frenzie : Huh, do you mean it stays on during standby? Or it starts getting turned off and then gets turned back on again?

Frenzie:
1. Turn it off.
2. Standby.
3. Wake up.
4. Light is on again.

NiLuJe:
@Frenzie : Oh, right, that makes sense. I'll take care of that ;).

Issue #1833, opened 2016-02-18
Poor image quality in epub files...


djalma2k:
Using Koreader 2015.11 stable and nightly-20160216
Device: Kobo Glo HD
Comparison between Sony PRS-T3 (left) and Kobo Glo HD using koreader (right):

Showing an image (epub) in normal mode
http://postimg.org/image/dzx7t739f
Showing an image (epub) in sleep mode:
http://postimg.org/image/ikje84myr

Using "nickel" in Kobo Glo HD:
Normal mode:
http://postimg.org/image/roht8kojn
Sleep mode:
http://postimg.org/image/8xfvyetz7

Seems some type of posterization (lack of dithering)... I don't know if it's a bug or maybe a setting to save cpu cycles when the image is rendered (nickel)....
Anyway, in case it's not a bug, how I could increase the image quality to a setting similar to Sony T3?
Regards and Thanks in advance!!!

NiLuJe:
Yup, no dithering, and the hardware 'dithering' is crappy.

djalma2k:
Well, I don't know if you are meaning that "hardware dithering" is crapp in a general sense or that it's the Kobo's hardware implementation? Surely you are right, I don't doubt that; but where or what is the crappy side of the "dithering": Bad side effects like ghosting or something like that? Because to me the"nickel" dithering seems good enough to my eyes or at least an improvement over the "naked"-raw image; and sony's implementation seems almost perfect to me...
Anyway, as I asked before is there any setting in there to enable it (hardware dithering or software if It could be possible)? My guess is not for your answer, but I appreciate you explain it to me.
Thanks in advance and Regards.

houqp:
@chrox and @frankyifei knows more about CRengine. I am not aware of hardware dithering support. Can you upload the book or extract that image for us to test?

djalma2k:
http://m.zbigz.com/myfiles?url=magnet:?xt=urn:btih:6B8597560E1DD5F6C485DD597E3E3ADD93922B60&dn=EPL_El_amante&tr=http://tracker.tfile.me/announce&tr=http://91.121.54.8:6969/announce&tr=udp://tracker.openbittorrent.com:80&tr=udp://tracker.publicbt.com:80&tr=udp://open.demonii.com:1337/announce

NiLuJe:
By HW I meant what the eInk driver/controller does behind the scenes.

[...]

PR #4922, opened 2019-04-14
Allow toggling CRe's new dithering & scaling

NiLuJe:
Requires https://github.com/koreader/koreader-base/pull/888

Fix https://github.com/koreader/koreader/issues/1833

### DIFF
--- a/frontend/ui/widget/imagewidget.lua
+++ b/frontend/ui/widget/imagewidget.lua
@@ -356,25 +356,27 @@ function ImageWidget:paintTo(bb, x, y)
         h = size.h
     }
     logger.dbg("blitFrom", x, y, self._offset_x, self._offset_y, size.w, size.h)
+    -- Figure out if we're trying to render one of our own icons...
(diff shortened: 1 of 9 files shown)


frontend/ui/widget/imagewidget.lua
```diff
@@ -355,17 +356,31 @@ function ImageWidget:paintTo(bb, x, y)
         h = size.h
     }
     logger.dbg("blitFrom", x, y, self._offset_x, self._offset_y, size.w, size.h)
+    -- Figure out if we're trying to render one of our own icons...
+    local not_an_icon = (not self.file or (self.file and not util.stringStartsWith(self.file, "resources/")))
```
poire-z: Isn't `local is_icon = self.file and util.stringStartsWith(self.file, "resources/")))` just more readable, and below `if ... and not is_icon then` ?

NiLuJe:
Yep, I was just too lazy to invert the test ;p.

PR #4924, opened 2019-04-15
Some more FL fixes

### LINKED ISSUE #4923: toggle front light on / off in the status bar and frontlight auto-activation ?
* KOReader version: 2019.04-25 ... 2019-04-13
* Device: KA1

1/ I updated to the last nightly yesterday. My status bar is set to display the frontlight intensity. I noticed that when I toggle the frontlight to on with a bottom left tap, the "off" in the status bar is not updated but It was before. Is this behaviour intended ?  The status bar update the value after a page turn.

2/ After a sleep and wake up I also noticed that the frontlight reactivates itself but even if it was off before I put the device to sleep... and the auto frontlight plugin is off...

Any idea ?

NiLuJe:
Fix #4923 & #4925

### DIFF
--- a/frontend/device/generic/powerd.lua
+++ b/frontend/device/generic/powerd.lua
@@ -120,7 +120,6 @@ function BasePowerD:normalizeIntensity(intensity)
 end
 
 function BasePowerD:setIntensity(intensity)
-    print("BasePowerD:setIntensity", intensity)
     if not self.device:hasFrontlight() then return false end
     if intensity == self:frontlightIntensity() then return false end
     self.fl_intensity = self:normalizeIntensity(intensity)
--- a/frontend/device/generic/powerd.lua
+++ b/frontend/device/generic/powerd.lua
@@ -120,6 +120,7 @@ function BasePowerD:normalizeIntensity(intensity)
 end
 
 function BasePowerD:setIntensity(intensity)
+    print("BasePowerD:setIntensity", intensity)
     if not self.device:hasFrontlight() then return false end
     if intensity == self:frontlightIntensity() then return false end
     self.fl_intensity = self:normalizeIntensity(intensity)
--- a/frontend/device/kobo/powerd.lua
+++ b/frontend/device/kobo/powerd.lua
@@ -142,6 +142,10 @@ function KoboPowerD:init()
         -- See discussion in https://github.com/koreader/koreader/issues/3118#issuecomment-334995879
         -- for the reasoning behind this bit of insanity.
         if self:isFrontlightOnHW() then
+            print("init: wasOn: self.hw_intensity", self.hw_intensity)
+            print("init: wasOn: self.fl_intensity", self.fl_intensity)
+            print("init: wasOn: self.initial_is_fl_on", self.initial_is_fl_on)
+            print("init: wasOn: self.is_fl_on", self.is_fl_on)
             -- On devices with a mixer, setIntensity will *only* set the FL, so, ensure we honor the warmth, too.
             if self.device:hasNaturalLightMixer() then
                self:setWarmth(self.fl_warmth)
@@ -149,9 +153,23 @@ function KoboPowerD:init()
             -- Use setIntensity to ensure it sets fl_intensity, and because we don't want the ramping behavior of turnOn
             self:setIntensity(self:frontlightIntensityHW())
         else
-            -- Use _setIntensity for setIntensityHW so as *NOT* to set fl_intensity, so toggle will still work,
-            -- plus the FrontlightStateChanged event.
-            self:_setIntensity(0)
+            -- Use setBrightnes so as *NOT* to set hw_intensity, so toggle will still (mostly) work.
+            print("init: wasOff: self.hw_intensity", self.hw_intensity)
+            print("init: wasOff: self.fl_intensity", self.fl_intensity)
+            print("init: wasOff: self.initial_is_fl_on", self.initial_is_fl_on)
+            print("init: wasOff: self.is_fl_on", self.is_fl_on)
+            self.fl:setBrightness(0)
+            -- And make sure the fact that we started with the FL off propagates as best as possible.
+            self.initial_is_fl_on = false
+            -- The first toggle won't work, because BasePowerD's init sets fl_intensity to hw_intensity right after this,
+            -- which screws with things for a cycle, but the next cycle will properly toggle between 0 & hw_intensity.
+            -- So, instead of making things behave properly,
+            -- setting hw_intensity to 1 would here bypass any KOBO_LIGHT_ON_START behavior, and instead toggle between 0 & 1.
+            --self.hw_intensity = 1
+            print("init: wasOff: self.hw_intensity", self.hw_intensity)
+            print("init: wasOff: self.fl_intensity", self.fl_intensity)
+            print("init: wasOff: self.initial_is_fl_on", self.initial_is_fl_on)
+            print("init: wasOff: self.is_fl_on", self.is_fl_on)
         end
     end
 end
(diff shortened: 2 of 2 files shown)


NiLuJe:
Note that in this specific case I can't get the first on/off cycle to behave in a sane manner because BasePowerD stomps on our fl_intensity/hw_intensity shenanigan, so it essentially breaks the first toggle, because it tries to set a state that is already the current one, instead of jumpoing directly to the other side of the equation (which is what would have happened had fl_intensity not been stomped on).

NiLuJe:
But at least this way it ends up toggling something, instead of being stuck because it's below fl_min.

NiLuJe:
Aha. Got it.

This shit is insane. :D.

houqp:
Yeah, the frontlight handling logic is not fun to work with...

NiLuJe:
The fact that the FL status couldn't actually be queried from the HW until Mk. 7 is seriously bonkers. That probably didn't help designing this when everyone wanted to support his own workflow ;).

cramoisi:
> so it essentially breaks the first toggle, because it tries to set a state that is already the current one
@NiLuJe : I'm not sure I understand. I applied the PR but nothing really changed, the frontlight doesn't toggle on

cramoisi:
OTA update fixed the thing, thanks :)

Hzj-jie:
It looks like the frontlight status is still breaking on my voyage after this change. Circumstances:
1. When the frontlight is off, level / on is shown in status bar; when the frontlight is on, off is shown.
2. The status bar won't update immediately after auto-frontlight switches the frontlight.
3. Several first clicks on bottom-left (kobofrontlight plugin I remember) do not take effect.

Frenzie:
Also, I don't know if this is an intentional change but now you can reduce the brightness all the way to 0. Adjusting brightness (from 1 to 100) and toggling (between 1-100 and 0) have been separate actions.

Biep:
And see my remark at #4925.

cramoisi:
On my KA1 all is working as intended with the last OTA update... (toggle is working and value is recorded between KOReader session)

NiLuJe:
The Kindle stuff is completely unrelated, as it was modified a while ago,
not with these changes.

@Frenzie : you can drop to 0 only in a specific way, or basically
everywhere? (FL widget, gestures,...).

Frenzie:
I disable the gesture controller. I mean with the widget. But it looks like something is going wrong while updating (probably caused by my being out of disk space the other day). So I'll get back to you on that one.

NiLuJe:
I'll double check on my end, because yeah, that shouldn't haven't changed, and I remember checking ;).

Frenzie:
Pressing `Min` on the widget goes to 1. Pressing `-1` goes to 0. I'm 99 % sure it stopped at 1 before.

Frenzie:
The frontlight still keeps turning on btw. But only through the toggle button on the widget I think, not with the gesture manager action.

NiLuJe:
I definitely remember min doing that, but I don't think there ever was
anything else, as the buttons are based on fl_min & max, which have always
been 0 & 100.

NiLuJe:
That's not right, both use toggleFrontlight :?

Frenzie:
You're right, the stable does seem to behave the same way, going down to 0.

Never mind then, although it is slightly confusing that -1 can go lower than min.

NiLuJe:
@Hzj-jie : I can only reproduce the status bar thing on my end, and it's probably unsolvable without massive kludges like on Kobo (kthxbye).

We do NOT update our own internal tracking of the FL on toggle *on purpose* (so we can restore it to the same level on the next toggle).
So we can't force-feed the status bar the "new" value, because it gets it from that internal variable we're relying on.
And, tiehr way, it might actually be a hard-query to the HW on Kindle, and, since it's done asynchronously, the FL probably hasn't yet been updated when we try to read the "new" status.

Frenzie:
Btw, doesn't the statusbar update once or twice a minute these days for the clock?

NiLuJe:
I vaguely seem to remember a scheduleIn with a * 60 somewher,e so, possibly.

cramoisi:
@Frenzie only if you enable the option "Auto refresh time" (I didn't enable it, the page-turn refresh is good enough for me)

Frenzie:
@cramoisi Right, thanks. So if one wants the FL status to adjust, I'd suggest turning that on. :-) (Or does it *only* refresh the time?)

NiLuJe:
Or just turning the page ;).

Sidebar: I find that status widget fairly useless. Do you really need help
telling that the light is on or off? ;).

cramoisi:
@Frenzie : I don't know ; you could revert the patch for testing purpose :) 
@NiLuJe : I use it at 1% : in darkness I can read without killing my eyes but in daylight, it's invisible, and I couldn't tell if it's on or off without this widget

Biep:
Actually, 1% is too bright for me in the dark - but I understand it is the minimum achievable.
Is that because the Nickel API is used, or because the underlying hardware (or Linux) doesn't allow smaller values?

Frenzie:
There's no such thing as a Linux limit, but it is indeed at the Linux kernel level that the limits we're directly dealing with are implemented.

Either the people at Kobo or the people who provide the hardware to Kobo implement a Linux driver, hopefully based around necessary hardware limits and not around some idea that no one would want it less bright. For example, if you reduce the frequency/brightness of a LED too much you'd effect flickering. Such limits could be hardwired (nigh impossible to get around given how intricately it's all put together) or hardcoded (quite hard to get around, but potentially easier).

You can find the kernel source here: https://github.com/kobolabs/Kobo-Reader

With that you can compile a custom kernel and end up with something like https://github.com/lgeek/okreader

For several Kobo devices we could get multitouch that way even though Kobo doesn't provide it on those devices. I can only assume they thought it didn't work well enough.

On some devices it may also be possible to gain USB OTG support doing something like that (e.g., attach a keyboard via a USB OTG converter). E.g., here's a custom kernel that does that: https://github.com/hjr/XCSoar-Kobo-Build There's a valid reason for not including support out of the box since most USB devices wouldn't be able to get enough power, but I still find it slightly disappointing that they went through the effort of disabling it. Especially since you could always use a powered USB hub to provide the power.

But anyway, that's a level of hacking that's well beyond running KOReader. It's dangerous, hard, and unnecessary. At least on a device like my H2O. Some other devices can be taken apart much easier, so you could just swap the internal SD card for some harmless experimentation.

Biep:
Thanks for the comprehensive answer!

NiLuJe:
Note that you can see all three variables behinds the FL handling via ntx_io ioctls, but, as Frenzie said, there's a chance the 0-100 mappings are made this way because going below what 1 is currently mapped to would yield undesirable behavior.

NiLuJe:
(Theoretical answer, I've never actually tried said ioctls, and the kernel source sometimes don't match what actually currently in current production kernels).

NiLuJe:
c.f., `CM_FRONT_LIGHT_DUTY`, `CM_FRONT_LIGHT_FREQUENCY`, `CM_FRONT_LIGHT_R_EN` @ `arch/arm/mach-mx5/mx50_ntx_io.c` (or `ntx_io.c` in a similar-ish path on newer kernels).

Frenzie:
That seems to indicate that what's mapped to 1 isn't necessarily the lowest it can go. (Based on a quick peek [here](https://gist.github.com/thotypous/6379053) because I don't want to bug my poor laptop with giant kernel sources.)

I did a quick search for the kind of program I would write to test and it already lives [here](https://www.mobileread.com/forums/showthread.php?t=200706). (Of course it'd be a bit harder on devices with three of 'em.)

Frenzie:
That topic also has a hardware hack: https://www.mobileread.com/forums/showpost.php?p=2504565&postcount=71

My H2O isn't nearly that moddable though, plus I think the lowest brightness is about just right. Is it brighter on the Forma?

NiLuJe:
I don't recall, but I'm fairly sure those ioctls are gone on the Forma (IIRC, they're at least empty in the sources).

Biep:
On my Aura One the lowest brightness is way too bright in true darkness.

NiLuJe:
Keep in mind that there's a degree of personal preference in those kinds of assessments.

In my case, in true darkness, lowest brightness settings give me headaches, so ;).

Biep:
So do they to me, but I suppose for the opposite reason.
But yes, I'll amend my statement: "way too bright **for me**".

That is also why I should prefer a logarithmic scale - the step from 1% to 2% is also way too large for me in slightly less dark conditions.  Conversely, I never need the fine grain of one-percent increases in the higher percentages.

PR #4927, opened 2019-04-16
Add more build products to gitignore list

xelxebar:
This is extremely minor, bu the following build/test products end up spamming git status:

- foo.sdr and foofoo.sdr
- koreader-kindle-arm-kindle5-linux-gnueabi*
- koreader-emulator-x86_64-unknown-linux-gnu*,

so I'm adding these to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ tags
 test/*
 *.tar
 *.log
+*.sdr/
 bin/
 install/
 spec/unit/data
@@ -23,13 +24,7 @@ git-rev-base
 emu
 luacov.stats.out
 trace-out.txt
-
-koreader-*.zip
-koreader-*.apk
-koreader-*.deb
-koreader-*.tar.gz
-koreader-*.targz
-koreader-*.click
+koreader-*
 
 l10n/*
 !l10n/Makefile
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -49,6 +50,7 @@ koreader-debian-x86_64-linux-gnu*
 koreader-debian-armel-arm-linux-gnueabi*
 koreader-debian-armhf-arm-linux-gnueabihf*
 koreader-kindle-legacy-arm-kindle-linux-gnueabi*
+koreader-kindle-arm-kindle5-linux-gnueabi*
```
Frenzie: There's also kindlepw2. I'm not sure if there's any particular reason we're not using something like `koreader-android*` though. Is there a deeper reason or has it just grown this way organically because there were only a two or three targets in the beginning?

Pinging @NiLuJe @houqp

.gitignore
```diff
@@ -57,6 +59,8 @@ koreader-emulator-i686-linux-gnu*
 koreader-emulator-x86_64-linux-gnu*
 koreader-emulator-x86_64-pc-linux-gnu*
 koreader-emulator-x86_64-apple-darwin*
+koreader-emulator-x86_64-unknown-linux-gnu*
```
Frenzie: Is that from Clang?

.gitignore
xelxebar: Thanks for bringing that up. I sorta opened this hoping to start that conversation in the first place :P

.gitignore
xelxebar: Pretty sure this is just what `_sysctl(2)` returns for my processor type:
```
$ uname --machine --processor --operating-system
x86_64 unknown GNU/Linux
```

Frenzie: So do most people. ;-)

```
$ uname --machine --processor --operating-system
x86_64 unknown GNU/Linux
$ gcc -dumpmachine
x86_64-linux-gnu
$ gcc --version
gcc (Debian 8.3.0-6) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

xelxebar: Ok. TIL. Here's my machine:
```
$ gcc -dumpmachine
x86_64-unknown-linux-gnu
$ gcc --version
gcc (GCC) 8.3.0
$ lsb_release -a
LSB Version:    1.0
Distributor ID: VoidLinux
Description:    Void Linux
Release:        rolling
Codename:       void
```

Taking a look at my distro's gcc package, it looks to be explicitly compiling with `--target`, setting it to the above.

Frenzie: On older versions of Clang (3-5, maybe also 6) it used to output something like `x86_64-unknown-linux-gnu`, but on Clang 7.0.1 it seems to be `x86_64-pc-linux-gnu`.

Frenzie: But yeah, I'd just make it `koreader-emulator*` (etc. etc.) or probably even just `koreader-*`. There doesn't seem to be a good reason not to.

xelxebar:
Okay, so I just went ahead and ignored all top level `koreader-*`. Also changed the commit log to reflect the change.

NiLuJe:
LGTM, but then I build in a different tree than I work on, so this neever really bothered me ;).

houqp:
LGTM :+1: 

Anyone know what creates foo.sdr and foofoo.sdr? Is it from the test?

xelxebar:
@houqp I believe so. Those pop up on my machine when doing a frontend test.

houqp:
that seems wrong, we should probably get it fixed :)

PR #4936, opened 2019-04-17
[GestureManager] Implement action to change highlight style using gesture manager

### LINKED ISSUE #4896: [Feature Request] Create cycler for highlight types
Just a placeholder so I remember I have to do this

Galunid:
This allows user to switch between "invert", "lighten", "underline" highlight styles using gestures, discussed https://github.com/koreader/koreader/issues/1095#issuecomment-481834569
closes #4896

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -870,6 +870,7 @@ function ReaderHighlight:onCycleHighlightAction()
             timeout = 1,
         })
     end
+    return true
 end
 
 function ReaderHighlight:onCycleHighlightStyle()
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -872,6 +872,20 @@ function ReaderHighlight:onCycleHighlightAction()
...
+        lighten = "underscore",
+        underscore = "invert",
+        invert = "lighten"
+    }
+    self.view.highlight.saved_drawer = next_actions[self.view.highlight.saved_drawer]
+    self.ui.doc_settings:saveSetting("highlight_drawer", self.view.highlight.saved_drawer)
+    UIManager:show(Notification:new{
+        text = T(_("Default highlight style changed to '%1'."), self.view.highlight.saved_drawer),
+        timeout = 1,
+    })
```
Frenzie: `return true` to stop propagation (saves a minor amount of CPU cycles).

Issue #4876, opened 2019-04-06
Line Heights problem


noembryo:
With v2019.04 there is a change in the way the text is presented with #4785.
I can see that a lot of work went on this change and to make it the right way, but it changed the way the books are displayed.

Up until the v2018.11.1 the display was this:
![Reader_2019-Apr-06_144431](https://user-images.githubusercontent.com/24675403/55671343-de09c900-5897-11e9-9b66-3612628d8558.png)
I was using `DCREREADER_CONFIG_LINE_SPACE_PERCENT_SMALL = 90` and `DCREREADER_CONFIG_DEFAULT_FONT_SIZE = 32` and the lines filled up all the screen.

With the v2019.04 the display is this:
![Reader_2019-Apr-06_144317](https://user-images.githubusercontent.com/24675403/55671408-67210000-5898-11e9-8a41-892a716d6b1b.png)
There is a line missing at the bottom.

I tried different settings (changing line heights (80%-100% and/or font sizes (29-33) but always the bottom gap is big.

This is with a bigger line height:
![Reader_2019-Apr-06_150018](https://user-images.githubusercontent.com/24675403/55671456-d7c81c80-5898-11e9-91dd-cd66c722590e.png)

I thought that there is a margin problem there, but I found that when there are different tags and not just `<p>` in the page I could almost fill all of it like this:
![Reader_2019-Apr-06_152314](https://user-images.githubusercontent.com/24675403/55671485-3392a580-5899-11e9-8881-f9bd6a5b3925.png)

I wonder, is there a way I could recreate the page the way it was before?
A setting to use the old numbers? (I know that its nit-picking, but as you can see from the first screenshot, I don't like big margins.. ;o)

Frenzie:
You're saying that, e.g., `85` to reduce the line-height to the previous value still leaves a gap at the bottom?

Also note the widows and orphans, btw.

noembryo:
Using the same values as before does not work as expected.
Using smaller line height brings more lines to the page but the gap is still there..

noembryo:
Aaa, and there are no widows and orphans..

noembryo:
Using 85 height (the same lines as the first screenshot but with a gap):
![Reader_2019-Apr-06_185652](https://user-images.githubusercontent.com/24675403/55671932-c2092600-589d-11e9-947d-a1c00a4edd0f.png)

Frenzie:
The previous values were broken. You might need to use a smaller value to achieve the same result.

[...]

PR #4081, opened 2018-07-16
Bookmarks, CoverBrowser: scale dogear icon

poire-z:
The Dogear icon is 20x20 pixels and was never scaled where used. Now:

The bookmark icon (top right of screen) is scaled to 1/32th of the screen width (previously, it was 1/30th on a 600px wide emulator, 1/53th on a GloHD).
So, it should stay the same size on a small emulator, and become 2 times bigger on a 1000px device.
1/32 looks like this:
<kbd>after_pdf_bookmark</kbd>

On CreDocument, furthermore decrease its size if needed depending on the selected margins so it never overwrite the text (as margins are scaled for DPI, this somehow scale the icon for dpi too).
Before => after
<kbd>before_bookmark</kbd> => <kbd>after_crebookmark</kbd>

CoverBrowser list view: scale it to the available room under the "N % of P page" text, so it does not cover "page".
<kbd>before_listmenu</kbd> => <kbd>after_listmenu</kbd>

CoverBrowser mosaic view: scale it to 1/16th of the cover rectangle, which should prevent if from overwritting the text thanks to a max text width of 7/8 of the cover rectangle.

<kbd>before_mosaicmenu</kbd>
=>
<kbd>after_mosaicmenu</kbd>

The above CoverBrowser screenshots were made on the emulator, and the icon size looks divided by 2. But the after size is actually how it looked before and still looks after on a GloHD.

Also for CoverBrowser: don't index metadata for unsupported document (which could happen when browsing files with PathChooser) and show full filename for such documents.

Also: ImageWidget: small fix in case we use both scale_factor and scale_for_dpi.

The 1/32 (for bookmark icons) and 1/16 (for coverbrowser mosaic view) can be changed if people find it too big or too small (they fit my liking, I like things small, but strangely, I don't bother the bigger bookmark icon).

Close #3265. Close #4046.

PR #4945, opened 2019-04-19
Give more control over CRe margins

NiLuJe:
Without having to resort to weird custom defaults.

* Split the current margins setting in three:
  * Horizontal margins (because you generally want those two to be balanced).
  * Top margin & Bottom margin (because you may want to tweak those separately to deal with quirky status bar/final line shenanigans).

* Also, add a "Reclaim bar height from bottom margin" toggle to the status bar menu, to optionally make sure the status bar won't eat into the bottom margin.

* Includes a free fix to diacritics popup refresh handling in the keyboard ;).


NiLuJe:
For instance, because I'm a crazy person, I'm going to override all of those on my end to basically 0,1,2,3,4,5,6,... ;).

NiLuJe:
Speaking of: what's the use-case for the crazy high values on the top end of the scale?

Granted, I like very minimal margins, but those get run through scaleBySize, so, I'm slightly at a loss as to what the intended use could be ;).

Frenzie:
Something like an Onyx Boox Max2. By which I mean anything with a giant display.

Frenzie:
PS I also wonder what the point is of those contrast settings over 2 or so. That's about where they turn from "meh, not for me" to "GET IT AWAY FROM ME D:". ;-)

NiLuJe:
True ;o).

NiLuJe:
margins

poire-z:
> For instance, because I'm a crazy person, I'm going to override all of those on my end to basically 0,1,2,3,4,5,6,... ;).

I'm not fond of having 2 additional progress bars just for margins (the more stuff there is, the less room there is to see their effect on the current book page).
But I see the point and the attempt to solve the other issue with the bottom margins and the feeling of missing a line at the bottom.
But this PR does not provide the solution to the default user :) only to the crazy person in the know who gonna edit their settings for that.

I'd rather see a single margin progress buttons + a fine tune progressbutton for bottom margins.
Or if really needed: a L/R margins progressbuttons + a T/B progressbuttons + a finetune bottom margin progressbutton.

NiLuJe:
I initially started with a simple H/V split, and in turns out this has a tendency to lead to unbalanced gaps between top/bottom (effective) margins.

So, yeah, one way or another, top/bottom are better controlled separately.

A dynamic fine tune sounds like a great idea, I'm just not quite sure how to present that in terms of UI?

poire-z:
> I also wonder what the point is of those contrast settings over 2 or so. That's about where they turn from "meh, not for me" to "GET IT AWAY FROM ME D:". ;-)

Might depend on the font and the eyesignt of the user... I use 4, and the max when i'm really tired.

> Speaking of: what's the use-case for the crazy high values on the top end of the scale?

Not sure generally, but when I got my first progressive lenses, with a very narrow reading zone at the bottom, and being not used at all to move my head (only my eyes) when reading, i was happy to have koreader show me 25% margin on each side so I got a narrow readable zone with no blur on the sides...

NiLuJe:
Fair enough, although after a scaleBySize, I'm not quite sure we need to go that high ;).

NiLuJe:
But speaking in terms of % of screen estate is interesting, and is indeed much less murky that this essentially somewhat meaningless unit we're currently using ;).

poire-z:
> A dynamic fine tune sounds like a great idea, I'm just not quite sure how to present that in terms of UI? 

Maybe:
L/R as you have it
T/B as you had it
Less B margin : 0 -1 -2 -3 -4 -5 -6 (whether scaleBySize or not, would be removed from the B margin set in the T/B button.
So one could decrease that B till he gets that missing line back.
I initially wrote "extra B margin", but I think it should have it decreased, to give the sense of having back that missing line.

(and this extra B would gain being available as a gesture for quick viewing its effect).

> (And anything I can think of wouldn't actually gain us any vertical space here).

Might be a time to add a 6th bottom button for cre (which has currently 5, while PDF already has 6) :)

poire-z:
> Less B margin : 0 -1 -2 -3 -4 -5 -6 

Or maybe something allowing us a 100% choice along `default font size * interline space` to eat out of the bottom margin, so we can really get a line back.

NiLuJe:
Thinking about it, by virtue of it being a bottom-anchored menu, adding a tab won't help us see through the menu to check out the state of the bottom margin ;).

Having the bars displayed (and stored) as absolute "margin" values help make this not completely crazy in the backend, so I'm wondering how a layout similar to the FL would work:

[-] [======] [+]

What those buttons actually increment/decrement is up for grabs, I was thinking something simple like 1/0.5 like the fonts, but something smarter could probably be done ;).

NiLuJe:
Anyway, brain is mush right now, I'll sleep on it, but I do quite like your final idea... :).

NiLuJe:
Okay, managed to implement the buttonprogress w/ -/+ finetune without having to mangle too many stuff ;).

margins

NiLuJe:
Stepping is currently hard-coded to increments of 1.

I could easily make that configurable if it turns out to be too coarse-grained for high-dpi devices (that's a 265dpi H2O, so, eh, should be okay).

frontend/apps/reader/modules/readerdogear.lua
```diff
@@ -43,10 +43,17 @@ end
 function ReaderDogear:onReadSettings(config)
     if not self.ui.document.info.has_pages then
         -- Adjust to CreDocument margins (as done in ReaderTypeset)
-        self:onSetPageMargins(
-            config:readSetting("copt_page_margins") or
```
houqp: just curious, do we have any plan to migrate the old setting?

poire-z:
Sorry for the delay in following up. I'm still unable to test this to see how practical it is, so just asking: is it?

Now that we have fine tuning, do we really need a progressbar for top and another for bottom? Can't a single T/B one be enough, and it would ensure the margins are balanced, with the option in the footer menu (Reclaim...) to take footer height into account, or not, in this balance?

(It's not so much about losing one progressbar - no pb now that we get that 6th bottom tab - it's just that I feel I'll want to have these T & B margins the same, and after playing with both to find what fits, there will be a last phase of ok, let's see with Hold the numbers we got, do some maths and distribute them to both sides... I'd rather have KOReader do that by default.)

No comment code-wise.

May be that added tab could be made 2nd instead of 3rd, so line-height is nearer font size without a non-font config in between?

> Thinking about it, by virtue of it being a bottom-anchored menu, adding a tab won't help us see through the menu to check out the state of the bottom margin ;).

Some refresh trick to exclude ConfigDialog from being painted on T&B margin change? And a 1 or 2s delay to trigger it's painting after that? :)

NiLuJe:
I do like having different T/B bars, yes, as it helps evening the gaps in case there really isn't enough space for a final line, and even margins would actually lead to uneven vertical alignment ;).

I have a couple flimsy reasons for leaving the tab in third position instead of second (as the crop tab is in KOpt). The first one is a smaller diff, the second is because it depends (to some extent) on the DPI & line-height settings of the previous tab, so it felt sensible to leave them in that order, as you're more likely to fine-tune margins *after* having fine-tuned line-heights.

(i.e., I choose line-heights based on what feels comfortable to actually read, I'm far less likely to fine-tune them for any other purpose, even if it would help with said final line issues).

Ultimately not strongly attached to that, though ;).

----

That's a lot of moving parts for something like that ;p. Probably easier just to float the whole panel up a few dozen pixels ;).

poire-z:
> because it depends (to some extent) on the DPI & line-height settings of the previous tab, so it felt sensible to leave them in that order, as you're more likely to fine-tune margins after having fine-tuned line-heights.
(i.e., I choose line-heights based on what feels comfortable to actually read, I'm far less likely to fine-tune them for any other purpose, even if it would help with said final line issues).

Mhhh :) first, you would choose your font size, so 4th tab :) then line-height (2nd), then margin (3rd)
And, with a prefered font and prefered font size, you'd probably not go touch the (default or prefered) line-height (you would already have them all to what feels confortable to actually read, no matter the book content, no?)

poire-z:
OK, I had a try.
It works and allows getting the missing line at the bottom. But it was a bit tedious :) mostly because we don't see the effect immediately as the bottom margin is hidden by the config panel.

Also, I was a bit confused not seeing my previous margins.
Of course, I haven't yet updated my prefered margins in default.persistent.lua, so I'll have to do that.
But I guess (as @houqp suggested) that we should try to use the previous margins saved in the book metadata.lua under `copt_page_margins` for already opened books - just to not surprise people and have them have to update all the books they have been reading to make them look as before. (And avoid us some issues.)

> I do like having different T/B bars, yes, as it helps evening the gaps in case there really isn't enough space for a final line, and even margins would actually lead to uneven vertical alignment

I guess the point is updating both (evenly) till we have enough space for the additional line and we should have even vertical alignment at some point (you'd lose 1px granularity, and have 2px granularity, if we move each by 1px, but that shouldn't be noticable).
But I see the case when pages are not all laid out the same (because of some paragraph margins, titles...) where you'd get the missing line on most but not on some pages, and you'd want to move down a bit all pages (so, more top margin, less bottom margin) so pages with the missing line have it less noticable.

But I dunno, after only a few minutes of playing, it's not that confortable. May be I'll get a grasp of it.

(Also, minor point: visually, the - and + buttons are not really aligned with the L/R buttons - dunno if it's some computations error, or something due to the size of the button set from the numbers of buttons - in which case, having only 7 values for T & B, and 9 for L/R would solve that.) 

Anyway, fine to have it merged to see how it goes and if we get some idea for more confort. But I think the previous saved margins should really be restored and re-used.

NiLuJe:
I'm seeing 2 pixels on the left and 1 on the right, which feels like a rounding/centering thing given the different amount of buttons, possibly combined with a missing border?

(There's a lot of stuff that does rounding in an unexpected manner. I'd expect coordinates to be floor'ed, and dimensions to be ceil'd, but in practice we mostly floor everything).

----

And yeah, honoring the previous margins should be doable, I'll look into it ;).

poire-z:
Why not do the migration in` ReaderTypeset:onReadSettings()` (for the book settings) where we deal with the new ones, and `ReaderTypeset:init()` (for the global settings, like I did at top of 4d67dd59c), to avoid some noise in the other modules (reader.lua, readerui.lua shouldn't have any mention of margins, not their business).
(I did delete the old ones in ReaderLink(), but may be we wouldn't need that, to allow going back to old versions - although for global settings it's better to keep them clean.)

NiLuJe:
I wanted to do it as early as possibly to make sure *nothing* would actually ever see the old format ever again ;).

NiLuJe:
I vaguely thought about moving all of that to a migrateSettings function tucked away in luasettings/docsettings, though, but that'd still be called as early as possibly (i.e., the exact same place as I'm currently using).

poire-z:
> I wanted to do it as early as possibly to make absolutely sure nothing would actually ever see the old format ever again

Because of ReaderDogear, who uses them too and might be init() before ReaderTypeset?
ReaderTypeSet could just `self.ui:handleEvent(Event:new("SetPageMargins", ...)` once it has migrating them then, onNextTick if needed.

It's ugly seeing margins mentionned in reader.lua and readerui.lua. I'd rather have all that in a dedicated module if it can't be done in the main module that deals with the new settings and dealt with the old ones.

> (And the reasoning behind deleting the old format is to avoid going through the conversion every time.)

We could check if the new ones are there, and if not, check if there are old ones, and only migrate them then. But OK about deleting the old settings, will make things tidy.

Frenzie:
> I vaguely thought about moving all of that to a migrateSettings function tucked away in luasettings/docsettings, though, but that'd still be called as early as possible (i.e., the exact same places as I'm currently using).

The same places technically perhaps, but I agree with @poire-z that this feels out of place at best.

NiLuJe:
Any further comments?

frontend/ui/data/settings_migration.lua
```diff
@@ -0,0 +1,40 @@
...
+local SettingsMigration = {}
+
+-- Shockingly, handles settings migration
+-- NOTE: supports LuaSettings & DocSettings objects as input, as both implement the same API
+function SettingsMigration:migrateSettings(config)
+    -- Figure out what kind of object we were passed, to make the logging more precise
+    local cfg_mt_idx = getmetatable(config).__index
+    local cfg_class
+    if cfg_mt_idx == DocSettings then
+        cfg_class = "local"
```
poire-z: "book" rather than "local"?

poire-z:
I haven't played with it since the other day, but it looks allright.
May be wait for the coming stable release (named "Labor day" ? :) before merging, so we get one month to see how it does? (I'm still on my tedious/confused impression :)

Frenzie:
Labor Day? Such sudden pressure, lol.

NiLuJe:
I'd rather have it *in* the stable, to gather feedback from a potentially wider range of  use-cases ;).

Frenzie:
It seems pretty safe to me. But then better put it in tonight or tomorrow so that it's had some testing by the time we reach the weekend. ;-)

poire-z:
Allright then.

poire-z:
> I have a couple flimsy reasons for leaving the tab in third position instead of second (as the crop tab is in KOpt). The first one is a smaller diff

Now that the smaller diff does not matter anymore, we can rediscuss that point :)

> the second is because it depends (to some extent) on the DPI & line-height settings of the previous tab, so it felt sensible to leave them in that order, as you're more likely to fine-tune margins after having fine-tuned line-heights.

That's one use case, mine would probably be different and justify another order.
But it looks odd the bottom tab icons are different between kopt and cre.
<kbd>image</kbd>

And there was some kind of logical order, from larger to smaller things, in kopt and in cre, that inverting 2nd and 3rd would bring back:
<kbd>image</kbd>
- Screen (orientation)
- Page (crop, page margins)
- Lines (line heights, with zoom for elements/paragraphs margins in px and images)
- Font size
- Font smaller tweaks (kerning, hinting, contrast)
- Last tab with other various thingies

poire-z:
Small issue with the DogEar: when on a page with a bookmark, when reducing margins, it does not get its size adapted anymore, and it may end up eating into the text (previously fixed in #4081).
(It only has `ReaderDogear:onSetPageMargins()`, that we don't send anymore, now that we switched to SetPageHorizMargins SetPageTopMargin and SetPageBottomMargin.)

poire-z:
Still a bit awkward with the distinct top & bottom margins.
My use case (often used, always the same) is that I use font size 19 in the morning, and sometimes in the evening i need 20 or 21. Switching to 20 (with my other prefered unchanged settings) leaves a blank line at the bottom. That's what was discussed in #4876.
So, I want to get it back, and this PR allows it!
So, either I start with Top or Bottom. Starting with bottom is obviously better.
1) Starting with Bottom: I reduce it, one or two steps, two is good with my use-case: one more line gained ! => 2 taps/rendering. But ok, it's no more centered.
2) So, I reduce Top to get that centering back (I guess that if we are bothered with the missing/blank line at bottom, we are bothered by miscentering :) => 1 more tap/rendering.
3) Now, I wouldn't need as much bottom margin, now that I gained some bit at the top (it may not matter much letting it as is, or it may on some pages , not sure): so, I increase again the Bottom margin => 1 more tap/rendering.
So, that made 4 taps/renderings, which with some big books make take 10s to 60s each :|

Now that I get it with my usual use-case, I know I can only tap once on top and once on bottom, so I'll get 2 tap/renderings. Less time wasted. So, I may be somehow fine with this (unnatural) top & margin distinction.

I was going to quick&dirty hack that code to see how it would feel with the Top margin progress bar applied to both top & bottom margins (possibly hiding the Bottom margin progress bar if it felt alright).
But now that we have a specific tab for these margins settings, would it make sense to less quick&dirty hack it and have another toggle `Lock T&B margins   off | on` (or a better name I can't find), so both progress bars would be kept sync'ed and both margins updated to the same value?
That would allow having the centering in 1 tap, and could be unlocked if some more precise tweaking is needed with some books or use-cases?

poire-z:
(OK, labor day or not, having a go at these last 2 points.)

NiLuJe:
Yeah, a keep top & bottom in sync toggle sounds like a neat idea, and
should be pretty painless to implement (just fudge the final set Margins
args in setTop/setBottom if the toggle is true).

NiLuJe:
Oh, and the dogear thing is probably because I'm calling
self:setPageMargins directly instead of sending an Event...

poire-z:
> Some refresh trick to exclude ConfigDialog from being painted on T&B margin change? And a 1 or 2s delay to trigger it's painting after that? :)

Had a quick try at that:
```diff
--- a/frontend/ui/data/creoptions.lua
+++ b/frontend/ui/data/creoptions.lua
@@ -102,2 +102,3 @@
                 event = "SyncPageTopBottomMargins",
+                delay_repaint = 2,
                 name_text_hold_callback = optionsutil.showValues,
@@ -139,2 +140,3 @@
                 },
+                delay_repaint = 2,
                 name_text_hold_callback = optionsutil.showValues,
@@ -171,2 +173,3 @@
                 },
+                delay_repaint = 2,
                 name_text_hold_callback = optionsutil.showValues,
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -516,2 +516,3 @@ function ConfigOption:init()
                     events = self.options[c].events,
+                    delay_repaint = self.options[c].delay_repaint,
                     config = self.config,
@@ -546,3 +547,3 @@ function ConfigOption:init()
                             self.config:onConfigChoose(self.options[c].values, self.options[c].name,
-                                self.options[c].event, self.options[c].args, self.options[c].events, arg)
+                                self.options[c].event, self.options[c].args, self.options[c].events, arg, self.options[c].delay_repaint)
                         end
@@ -870,3 +871,3 @@ end

-function ConfigDialog:onConfigChoose(values, name, event, args, events, position)
+function ConfigDialog:onConfigChoose(values, name, event, args, events, position, delay_repaint)
     UIManager:tickAfterNext(function()
@@ -887,13 +888,22 @@ function ConfigDialog:onConfigChoose(values, name, event, args, events, position
         self:update()
-        if self.config_options.needs_redraw_on_change then
-            -- Some Kopt document event handlers just save their setting,
-            -- and need a full repaint for kopt to load these settings,
-            -- notice the change, and redraw the document
-            UIManager:setDirty("all", "partial")
+        local do_refresh = function()
+            self.skip_paint = nil
+            if self.config_options.needs_redraw_on_change then
+                -- Some Kopt document event handlers just save their setting,
+                -- and need a full repaint for kopt to load these settings,
+                -- notice the change, and redraw the document
+                UIManager:setDirty("all", "partial")
+            else
+                -- CreDocument event handlers do their own refresh:
+                -- we can just redraw our frame
+                UIManager:setDirty(self, function()
+                    return "ui", self.dialog_frame.dimen
+                end)
+            end
+        end
+        if delay_repaint then
+            UIManager:scheduleIn(delay_repaint, do_refresh)
+            self.skip_paint = true
         else
-            -- CreDocument event handlers do their own refresh:
-            -- we can just redraw our frame
-            UIManager:setDirty(self, function()
-                return "ui", self.dialog_frame.dimen
-            end)
+            do_refresh()
         end
--- a/frontend/ui/widget/container/inputcontainer.lua
+++ b/frontend/ui/widget/container/inputcontainer.lua
@@ -70,2 +70,5 @@ function InputContainer:paintTo(bb, x, y)
     end
+    if self.skip_paint then
+        return
+    end

--- a/frontend/ui/widget/toggleswitch.lua
+++ b/frontend/ui/widget/toggleswitch.lua
@@ -204,3 +204,3 @@ function ToggleSwitch:onTapSelect(arg, gev)
     self.config:onConfigChoose(self.values, self.name,
-                    self.event, self.args, self.events, self.position)
+                    self.event, self.args, self.events, self.position, self.delay_repaint)
     UIManager:setDirty(self.config, function()
```
It's quite ok (it just crashes if you switch the toggle too fast before the timeout), but I think it shouldn't be a timeout (to give the time to appreciate the result), and should probably be coupled with the InfoMessage showing the values, with an added line: "(Hit screen to dismiss)" so it shows the ConfigDialog again only then.

poire-z:
> and should probably be coupled with the InfoMessage showing the values, with an added line: "(Hit screen to dismiss)" so it shows the ConfigDialog again only then.

[This patch](https://github.com/koreader/koreader/files/3136968/delay_repaint2a.diff.txt) would give this:

<kbd>margins_tap_dismiss</kbd>

NiLuJe:
@poire-z : I have yet to try it, but I think I like it, and it doesn't look too gnarly ;).

NiLuJe:
@poire-z : Yup, I like it in practice, too ;).

I kinda liked the timeout approach, though, but I don't know how feasible that would be here? (i.e., have a timeout, that you can cancel early by tapping?).

poire-z:
> I don't know how feasible that would be here

Just a matter of having dismiss_callback called here:
https://github.com/koreader/koreader/blob/3009f5ae54f3c77db0b2e9f31671da220287d31f/frontend/ui/widget/infomessage.lua#L185-L187
like it is done here:
https://github.com/koreader/koreader/blob/3009f5ae54f3c77db0b2e9f31671da220287d31f/frontend/ui/widget/infomessage.lua#L200-L202
(Possibly only if specified as an option, dunno if in general, a timeout should call a dismiss_callback, or if it should be called only if explicite dismiss with tap - I don't think we yet use these together.)

> I kinda liked the timeout approach

I didn't :) Your 4 seconds were too near my mean re-action time :) So, either I was fast (3s!) and tapped on another progressbar button, and it didn't react as it was just dismissing the infomessage, and so I needed to tap another time. Or I was slow and it worked. Kinda upsetting :)

With this, we don't know how much time the user will want to appreciate the margins. Sometimes not much when they are bad - sometimes 10 seconds when they are fine. A timeout would show the bottom menu again, and he would have to tap again to dismiss it and see again the full page.
So, hard to guess a timeout value that won't upset some users. A _always tap to dismiss_ gives a more deterministic experience I think.

(I'll make a PR of it after "not labor day but the sunday after" stable release :)

NiLuJe:
Okay, let's forget about timeouts if you feel like 4s is too short then ;). Thanks for the pointer, though :).

PR #4970, opened 2019-04-25
[fix, spec] enable ReaderView tests

Frenzie:
It's not clear to me where the value is set as expected when running all tests linearly because it doesn't seem to be ReaderFooter.

### DIFF
--- a/spec/unit/readerview_spec.lua
+++ b/spec/unit/readerview_spec.lua
@@ -47,8 +47,9 @@ describe("Readerview module", function()
         end
     end)
 
-    -- TODO FIX THESE TESTS THEY'RE BROKEN!!!
-    it("should return and restore view context in page mode #notest", function()
+    it("should return and restore view context in page mode", function()
+        -- we don't want a footer for this test
+        G_reader_settings:saveSetting("reader_footer_mode", 0)
         local sample_pdf = "spec/front/unit/data/2col.pdf"
         local readerui = ReaderUI:new{
             document = DocumentRegistry:openDocument(sample_pdf),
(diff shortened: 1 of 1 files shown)


spec/unit/readerview_spec.lua
```diff
@@ -47,8 +47,9 @@ describe("Readerview module", function()
         end
     end)
 
-    -- TODO FIX THESE TESTS THEY'RE BROKEN!!!
```
Frenzie: To be clear, the test wasn't broken per se, but it was secretly depending on some other test turning off the footer.

NiLuJe:
:+1:

That kind of sneaky inter-test dependency makes a lot of the tests nigh ungrokable when they break :/.

Frenzie:
I suppose I could/should track it down with a clever shellscript.

For each _spec file, remove the whole koreader-emulator-* folder (to be sure), then test the spec file, then run readerview_spec (without these changes).

Assuming that'd be slightly faster than doing it manually. In any case it'd be more interesting than just doing it manually for half an hour (if you're unlucky). :-P

Frenzie:
Oh, I think I may have an idea about the problem:

https://github.com/koreader/koreader/blob/4dc116ed6a43e8ef625a3b96bb1740d309be2eca/spec/unit/commonrequire.lua#L10-L13

You see, `require` caches things.

Frenzie:
Also that `--no-auto-insulate` isn't helping. This will fix that (and it should probably be contributed upstream to Busted):

```lua
-- don't try to overwrite metatables so we can use --auto-insulate-tests
-- shamelessly copied from https://github.com/Olivine-Labs/busted/commit/db6d8b4be8fd099ab387efeb8232cfd905912abb
local ffi = require "ffi"
local old_metatype = ffi.metatype
local exists = {}
ffi.metatype = function(def, mttable)
    if exists[def] then return exists[def] end
    exists[def] = old_metatype(def, mttable)
    return exists[def]
end
```

Result without this patch:
```
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] ./spec/front/unit/readertoc_spec.lua @ 24: Readertoc module should get toc title from page
[  FAILED  ] ./spec/front/unit/readertoc_spec.lua @ 69: Readertoc module should get page left of chapter
[  FAILED  ] ./spec/front/unit/readertoc_spec.lua @ 74: Readertoc module should get page done of chapter
[  FAILED  ] ./spec/front/unit/readerview_spec.lua @ 51: Readerview module should return and restore view context in page mode
[  FAILED  ] ./spec/front/unit/readerview_spec.lua @ 99: Readerview module should return and restore view context in scroll mode

 5 FAILED TESTS
```

So readerview_spec failing is intended; readertoc is apparently also sneakily depending on another test.

Issue #4861, opened 2019-03-30
Table of contents is not working 


hugleo:
If I test the table of contents in Kobo glo the action doesn't go direct to the the chapter title. I need back 5 pages, sometimes 10 pages.

If I test in the Likebook Mars, a android device, then each item in table of contents that I select I'm being redirect to the book cover.

Frenzie:
What kind of document is this? Is it freely available?

hugleo:
Epub, yes. But is not the file. I copy the file and upload the same file to another folder and the table of contents worked normally.
So I deleted the "broken file" and moved the other file to the same folder. The file is the same and the same name as before. And after that process tabled of contents is working normally.

I do know what could be happened but I remember of installed a old koreader git version before and uninstall. I removed some .sdr files, and koreader folder in some place, no sure if I remove all the cache files though.  After that I updated the boyue os since was using a very old version and install koreader again today (most recent git build night). Maybe boyue os missed something with files and caches that occasioned that unexpected behavior to Koreader. Or maybe some koreader cached files.

Since I can't reproduce the problem anymore we can close this issue.

Eduardomb22:
I think it is a problem with the cached files, sometimes I get this too. All itens at ToC will go the book cover and have position number 1 in front of it.

hugleo:
Happened again here after I moved a book to another folder (not used koreader for this, use my own mtp file manager). But I just manually removed the /koreader/cache folder and worked! Maybe a good thing clean the cache folder when koreader restart?

poire-z:
The cache is useful for quicker load. And the crengine code has some checks to see if a cache is valid, and conformed to your current settings. Depending on what it noticed has changed (from what's cached in the cache), it can reload the book, or keep the DOM (the tree of HTML nodes and text) and just do a re-rendering.
Next time you witness that, can you try just changing the font size (and later back) to see if the forced re-rendering is enough to build a working TOC ?
(If you didn't move the .sdr and the previous settings, using the default settings should have made crengine noticed a change in layout, and trigger a re-rendering.)
Also check you crash.log, to see if crengine did output some messages at that moment (usually look like `CRE WARNING: blah blah)`
This happened on the same KOReader version? no KOReader update involved?

hugleo:
[...]

PR #4972, opened 2019-04-25
[spec] Better insulation

Frenzie:
Some combination of Travis and/or older Busted necessitated insufficient insulation.

Follow-up to https://github.com/koreader/koreader/pull/4970

### DIFF
--- a/spec/unit/readertoc_spec.lua
+++ b/spec/unit/readertoc_spec.lua
@@ -8,7 +8,15 @@ describe("Readertoc module", function()
         ReaderUI = require("apps/reader/readerui")
         DEBUG = require("dbg")
 
+        G_reader_settings:reset({})
+
         local sample_epub = "spec/front/unit/data/juliet.epub"
+
+        local DocSettings = require("docsettings")
+        local purgeDir = require("ffi/util").purgeDir
+        purgeDir(DocSettings:getSidecarDir(sample_epub))
+        os.remove(DocSettings:getHistoryPath(sample_epub))
+
         readerui = ReaderUI:new{
             document = DocumentRegistry:openDocument(sample_epub),
         }
@@ -24,9 +32,9 @@ describe("Readertoc module", function()
     it("should get toc title from page", function()
         title = toc:getTocTitleByPage(60)
         DEBUG("toc", toc.toc)
-        assert(title == "SCENE V. A hall in Capulet's house.")
+        assert.is.equal("SCENE V. A hall in Capulet's house.", title)
         title = toc:getTocTitleByPage(187)
-        assert(title == "SCENE I. Friar Laurence's cell.")
+        assert.is.equal("SCENE II. Hall in Capulet's house.", title)
     end)
     describe("getTocTicks API", function()
         local ticks_level_0 = nil
@@ -68,12 +76,12 @@ describe("Readertoc module", function()
     end)
     it("should get page left of chapter", function()
         assert.truthy(toc:getChapterPagesLeft(10, 0) > 10)
-        assert.truthy(toc:getChapterPagesLeft(97, 0) > 10)
+        assert.are.same(10, toc:getChapterPagesLeft(97, 0))
         assert.are.same(nil, toc:getChapterPagesLeft(290, 0))
     end)
     it("should get page done of chapter", function()
         assert.truthy(toc:getChapterPagesDone(12, 0) < 5)
-        assert.truthy(toc:getChapterPagesDone(95, 0) < 5)
+        assert.is.same(6, toc:getChapterPagesDone(95, 0))
         assert.truthy(toc:getChapterPagesDone(290, 0) > 10)
     end)
     describe("collasible TOC", function()
--- a/spec/unit/readertoc_spec.lua
+++ b/spec/unit/readertoc_spec.lua
@@ -9,14 +9,7 @@ describe("Readertoc module", function()
         ReaderUI = require("apps/reader/readerui")
         DEBUG = require("dbg")
 
-        G_reader_settings:reset({})
-
         local sample_epub = "spec/front/unit/data/juliet.epub"
-
-        local DocSettings = require("docsettings")
-        local purgeDir = require("ffi/util").purgeDir
-        purgeDir(DocSettings:getSidecarDir(sample_epub))
-        os.remove(DocSettings:getHistoryPath(sample_epub))
 
         readerui = ReaderUI:new{
             document = DocumentRegistry:openDocument(sample_epub),
--- a/spec/unit/readertoc_spec.lua
+++ b/spec/unit/readertoc_spec.lua
@@ -1,10 +1,11 @@
 describe("Readertoc module", function()
-    local DocumentRegistry, ReaderUI, DEBUG
+    local DocumentRegistry, Event, ReaderUI, DEBUG
     local readerui, toc, toc_max_depth, title
 
     setup(function()
         require("commonrequire")
         DocumentRegistry = require("document/documentregistry")
+        Event = require("ui/event")
         ReaderUI = require("apps/reader/readerui")
         DEBUG = require("dbg")
 
@@ -22,6 +23,8 @@ describe("Readertoc module", function()
         }
         -- reset book to first page
         readerui.rolling:onGotoPage(0)
+        readerui.document:setSpaceCondensing(75)
+        readerui:handleEvent(Event:new("ChangeScreenMode", "portrait"))
         toc = readerui.toc
     end)
 
@@ -34,7 +37,7 @@ describe("Readertoc module", function()
         DEBUG("toc", toc.toc)
         assert.is.equal("SCENE V. A hall in Capulet's house.", title)
         title = toc:getTocTitleByPage(187)
-        assert.is.equal("SCENE II. Hall in Capulet's house.", title)
+        assert.is.equal("SCENE I. Friar Laurence's cell.", title)
     end)
     describe("getTocTicks API", function()
         local ticks_level_0 = nil
@@ -76,12 +79,12 @@ describe("Readertoc module", function()
     end)
     it("should get page left of chapter", function()
         assert.truthy(toc:getChapterPagesLeft(10, 0) > 10)
-        assert.are.same(10, toc:getChapterPagesLeft(97, 0))
+        assert.truthy(toc:getChapterPagesLeft(97, 0) > 10)
         assert.are.same(nil, toc:getChapterPagesLeft(290, 0))
     end)
     it("should get page done of chapter", function()
         assert.truthy(toc:getChapterPagesDone(12, 0) < 5)
-        assert.is.same(6, toc:getChapterPagesDone(95, 0))
+        assert.truthy(toc:getChapterPagesDone(95, 0) < 5)
         assert.truthy(toc:getChapterPagesDone(290, 0) > 10)
     end)
     describe("collasible TOC", function()
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,6 @@ testfront: $(INSTALL_DIR)/koreader/.busted
 	-rm -rf spec/unit/data/*.sdr
 	cd $(INSTALL_DIR)/koreader && ./luajit $(shell which busted) \
 		--sort-files \
-		--no-auto-insulate \
 		--output=gtest \
 		--exclude-tags=notest $(BUSTED_OVERRIDES) $(BUSTED_SPEC_FILE)
 
@@ -136,7 +135,6 @@ coverage: $(INSTALL_DIR)/koreader/.luacov
 	cd $(INSTALL_DIR)/koreader && \
 		./luajit $(shell which busted) --output=gtest \
 			--sort-files \
-			--no-auto-insulate \
 			--coverage --exclude-tags=nocov
 	# coverage report summary
 	cd $(INSTALL_DIR)/koreader && tail -n \
--- a/kodev
+++ b/kodev
@@ -758,8 +758,6 @@ OPTIONS:
 
         echo "Running tests in" "${test_path}"
         busted --lua="./luajit" "${opts}" \
-            --no-auto-insulate \
-            --lazy \
             --output=gtest \
             --exclude-tags=notest "${test_path}"
     } && popd || exit
@@ -809,8 +807,6 @@ OPTIONS:
             echo "Running tests in" ${test_path}
             busted --lua="./luajit" \
                 --sort-files \
-                --no-auto-insulate \
-                --lazy \
                 -o "./spec/${target}/unit/verbose_print" \
                 --coverage \
                 --exclude-tags=nocov "${test_path}" || {
--- a/spec/unit/commonrequire.lua
+++ b/spec/unit/commonrequire.lua
@@ -1,3 +1,14 @@
+-- don't try to overwrite metatables so we can use --auto-insulate-tests
+-- shamelessly copied from https://github.com/Olivine-Labs/busted/commit/db6d8b4be8fd099ab387efeb8232cfd905912abb
+local ffi = require "ffi"
+local old_metatype = ffi.metatype
+local exists = {}
+ffi.metatype = function(def, mttable)
+    if exists[def] then return exists[def] end
+    exists[def] = old_metatype(def, mttable)
+    return exists[def]
+end
+
 require "defaults"
 package.path = "?.lua;common/?.lua;rocks/share/lua/5.1/?.lua;frontend/?.lua;" .. package.path
 package.cpath = "?.so;common/?.so;/usr/lib/lua/?.so;rocks/lib/lua/5.1/?.so;" .. package.cpath


Frenzie:
Apparently readerrolling runs interference.

spec/unit/readertoc_spec.lua
```diff
@@ -1,19 +1,23 @@
...
         DEBUG = require("dbg")
 
         local sample_epub = "spec/front/unit/data/juliet.epub"
+
         readerui = ReaderUI:new{
             document = DocumentRegistry:openDocument(sample_epub),
         }
         -- reset book to first page
         readerui.rolling:onGotoPage(0)
+        readerui.document:setSpaceCondensing(75)
```
Frenzie: @poire-z Could this be related to the cache file btw? It's not quite clear to me why this setting pollutes.

poire-z: I'm not sure, but I don't think so.
It's possible the default is 50 in crengine if not specified.
But we should set it in `ReaderFont:onReadSettings()`:
https://github.com/koreader/koreader/blob/27ddd6f313cf4cc609f6ec064e09ae0f0a049beb/frontend/apps/reader/modules/readerfont.lua#L156-L158

and this setting is accounted in the HASH that is used to see if the cached rendering info is still valid, or if a re-rendering is to be done:
https://github.com/koreader/crengine/blob/cf8a533b20cf8be30b30e79aa5ff0a0ed8b7dc61/crengine/src/lvtinydom.cpp#L10147-L10153

So, any change of minSpaceCondensing should be noticed and cache reused or not accordingly.

Frenzie: I may have very well misidentified space condensing as the culprit. A `printf("_minSpaceCondensingPercent: %i\n", _minSpaceCondensingPercent);` shows it to be at 75. Getting rid of the 50 in readerrolling or setting it to 75 in readertoc seemed to solve the problem.

I can in any case confirm that the cache is at fault, tested by removing it manually. I've attached the cache generated by readerrolling_spec (breaks-test) and by readertoc_spec (works-test) in case it might help.
[cr3cache.zip](https://github.com/koreader/koreader/files/3119911/cr3cache.zip)

```sh
$ md5sum juliet.epub.aaa10123.1.cr3.works-test 
d70ea3f15cc9edd83a0ed4442f5d8c91  juliet.epub.aaa10123.1.cr3.works-test
$ md5sum juliet.epub.aaa10123.1.cr3.breaks-test 
88df0fbee32b3b9aa9225c5610c57da6  juliet.epub.aaa10123.1.cr3.breaks-test
```

So for the purposes of the test it might be best to remove the cache in teardown, even if it'd result in slightly longer test runs. On the flipside, it might hint at some bug. I'm not quite sure. :-/

poire-z: I'm not sure a md5sum on the cache file is a hint it has structurally changed, although I don't see/remember what may change a bit in it in different invocations of KOreader with no setting change:
```
$ md5sum Saint-Empire-ique.FR.epub.e32923e8.1.cr3 ; rm Saint-Empire-ique.FR.epub.e32923e8.1.cr3
    017dba090de3fcee9c52ea032a99b6d8  Saint-Empire-ique.FR.epub.e32923e8.1.cr3
$ md5sum Saint-Empire-ique.FR.epub.e32923e8.1.cr3 ; rm Saint-Empire-ique.FR.epub.e32923e8.1.cr3
    ce0bc5f6bcdd6d7bb402b1ab31190e34  Saint-Empire-ique.FR.epub.e32923e8.1.cr3
$ md5sum Saint-Empire-ique.FR.epub.e32923e8.1.cr3 ; rm Saint-Empire-ique.FR.epub.e32923e8.1.cr3
    be6e9e7648294d6d1415bdc20f3cc22c  Saint-Empire-ique.FR.epub.e32923e8.1.cr3
```

Frenzie: I confirmed the behavior. The md5sum merely illustrates that there seems to be a subtle difference between the generated cache even though it has exactly the same size.

poire-z:
> I'm not sure a md5sum on the cache file is a hint it has structurally changed, although I don't see/remember what may change a bit in it in different invocations of KOreader with no setting change:

> The md5sum merely illustrates that there seems to be a subtle difference between the generated cache even though it has exactly the same size. This may be meaningless.

I figured out what causes these changes in cache file checksums: it's some inplicit padding (by the compiler) in some of the struct that gets serialized to the cache file. The padded area may contain random data, that will get serialized. Everything else was being the same.
So, no issue with the real cache content.

We could avoid these, and have "reproducible :)" cache files with the same hash, with stuff like:
```diff
--- a/crengine/include/lvtinydom.h
+++ b/crengine/include/lvtinydom.h
@@ -674,16 +674,23 @@ private:

     /// 0: packed 32bit data field
     ldomNodeHandle _handle; // _docIndex, _dataIndex, _type
+    // Some implicit padding with 32 additional bits may be done here
+    // to align the next 64bits pointer on x64)

     /// 4: misc data 4 bytes (8 bytes on x64)
     union {                    // [8] 8 bytes (16 bytes on x64)
         ldomTextNode * _text_ptr;   // NT_TEXT: mutable text node pointer
         tinyElement * _elem_ptr;    // NT_ELEMENT: mutable element pointer
+      struct {
+      union {
 #if BUILD_LITE!=1
         lUInt32 _pelem_addr;        // NT_PELEMENT: element storage address: chunk+offset
         lUInt32 _ptext_addr;        // NT_PTEXT: persistent text storage address: chunk+offset
 #endif
         lUInt32 _nextFreeIndex;     // NULL for removed items
+      };
+      lUInt32 _padding;
+      };
     } _data;


--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -399,6 +399,9 @@ struct CacheFileItem
     lUInt64 _dataHash; // additional hash of data
     lUInt64 _packedHash; // additional hash of packed data
     lUInt32 _uncompressedSize;   // size of uncompressed block, if compression is applied, 0 if no compression
+    lUInt32 _padding;  // explicite padding (this struct would be implicitely padded from 44 bytes to 48 bytes)
+                       // so we can set its value to 0 (instead of some random content with implicite padding)
+                       // in order to get reproducible (same file checksum) cache files when this gets serialized
     bool validate( int fsize )
     {
         if ( _magic!=CACHE_FILE_ITEM_MAGIC ) {
@@ -425,6 +428,7 @@ struct CacheFileItem
     , _dataHash(0)          // hash of data
     , _packedHash(0) // additional hash of packed data
     , _uncompressedSize(0)  // size of uncompressed block, if compression is applied, 0 if no compression
+    , _padding(0)           // (padding)
     {
     }
 };
@@ -1997,6 +2001,7 @@ void tinyNodeCollection::recycleTinyNode( lUInt32 index )
         ldomNode * p = &part[index & TNC_PART_MASK];
         p->_handle._dataIndex = 0; // indicates NULL node
         p->_data._nextFreeIndex = _elemNextFree;
+        p->_data._padding = 0;
         _elemNextFree = index;
         _itemCount--;
     } else {
@@ -2006,6 +2011,7 @@ void tinyNodeCollection::recycleTinyNode( lUInt32 index )
         ldomNode * p = &part[index & TNC_PART_MASK];
         p->_handle._dataIndex = 0; // indicates NULL node
         p->_data._nextFreeIndex = _textNextFree;
+        p->_data._padding = 0;
         _textNextFree = index;
         _itemCount--;
     }
@@ -12656,6 +12662,7 @@ ldomNode * ldomNode::insertChildText( lUInt32 index, const lString16 & value )
         //node->_data._ptext_addr._parentIndex = _handle._dataIndex;
         lString8 s8 = UnicodeToUtf8(value);
         node->_data._ptext_addr = getDocument()->_textStorage.allocText( node->_handle._dataIndex, _handle._dataIndex, s8 );
+        node->_data._padding = 0;
 #endif
         me->_children.insert( index, node->getDataIndex() );
         return node;
@@ -12680,6 +12687,7 @@ ldomNode * ldomNode::insertChildText( const lString16 & value )
         ldomNode * node = getDocument()->allocTinyNode( NT_PTEXT );
         lString8 s8 = UnicodeToUtf8(value);
         node->_data._ptext_addr = getDocument()->_textStorage.allocText( node->_handle._dataIndex, _handle._dataIndex, s8 );
+        node->_data._padding = 0;
 #endif
         me->_children.insert( me->_children.length(), node->getDataIndex() );
         return node;
@@ -12702,6 +12710,7 @@ ldomNode * ldomNode::insertChildText(const lString8 & s8)
 #else
         ldomNode * node = getDocument()->allocTinyNode( NT_PTEXT );
         node->_data._ptext_addr = getDocument()->_textStorage.allocText( node->_handle._dataIndex, _handle._dataIndex, s8 );
+        node->_data._padding = 0;
 #endif
         me->_children.insert( me->_children.length(), node->getDataIndex() );
         return node;
@@ -13094,6 +13103,7 @@ ldomNode * ldomNode::persist()
             int childCount = elem->_children.length();
             _handle._dataIndex = (_handle._dataIndex & ~0xF) | NT_PELEMENT;
             _data._pelem_addr = getDocument()->_elemStorage.allocElem(_handle._dataIndex, elem->_parentNode ? elem->_parentNode->_handle._dataIndex : 0, elem->_children.length(), elem->_attrs.length() );
+            _data._padding = 0;
             ElementDataStorageItem * data = getDocument()->_elemStorage.getElem(_data._pelem_addr);
             data->nsid = elem->_nsid;
             data->id = elem->_id;
@@ -13118,6 +13128,7 @@ ldomNode * ldomNode::persist()
             lUInt32 parentIndex = _data._text_ptr->getParentIndex();
             _handle._dataIndex = (_handle._dataIndex & ~0xF) | NT_PTEXT;
             _data._ptext_addr = getDocument()->_textStorage.allocText(_handle._dataIndex, parentIndex, utf8 );
+            _data._padding = 0;
             // change type
         }
     }
```
That was for on the emulator, which is 64bits.
I don't know if our devices are 32 or 64bits :| and if we should care about not adding these uneeded padding on 32bits.
Haven't yet looked at how we can guess that at compile time, if there are some `#define` available for this so we can put that into `#ifdef` & co. I guess so, and I guess @NiLuJe already knows and will tell me how to do that properly :)

Do we want that? (I guess so, just to avoid these false alarms and have a quick tool to compare cached data).

Frenzie:
But then, how come swapping out the cache files has consequences for the tests? :-/

poire-z:
I don't know. But I was just investigating the variable md5sum of cache files that I didn't expect.

For your tests issues, well, I don't know, I'm not even really sure what is it that is bothering, as I'm not as deep in the scenario/order of tests (and you didn't explain much, and assumed it was obvious and I would understand :)

Does it work the same if you had set readerui.document:setSpaceCondensing(50 or 100)? assuming the cache must be set with one among 50, 75, 100, so one of them should not trigger a re-rendering.
Are the readertoc tests always run after readerrolling tests?

Well, just something that came to mind, related to TOC and cache: https://github.com/koreader/koreader-base/pull/601#discussion_r171065144 but it's quite old and I hardly understand it now :| but I don't see why this would not cause problems in daily usage, but just with tests.
Unless the reader do call some other methods that the tests don't.

poire-z:
May be we are somehow reproducing here the kind of issues discussed in https://github.com/koreader/koreader/issues/4861#issuecomment-478312940

NiLuJe:
See [this](http://www.catb.org/esr/structure-packing/) for a good reference about structure padding/packing ;).

The `CacheFileItem` struct in `lvtinydom.cpp` indeed ends up being aligned to 8 bytes boundaries (uit64_t), and the only gap is the padding to alignment of the struct itself (i.e., 4 bytes at the end, which you can make explicit by using `char pad[4]`, even if an uint62_t does the job just fine, both on 32b & 64b ;)).

The trickery with the unions in the header is much more iffy, because it's an union, and these are pointers, so their size *will* change depending on the arch.
It'll be 4 bytes on 32bits, and actually makes sense, as the size of every possible member of the union will match (sizeof(ptr_t) == sizeof(lUInt32) == 4 on 32bit), but it gets iffier on, 64bits. The pointer will be 8 bytes, but uint62_t still 4 bytes. Which works, because an union, but the union itself will inherit the largest sizeof possible, 8 bytes.

If it's serialized as-is somewhere, that may throw a wrench on padding somewhere, but otherwise, it shouldn't be an issue ;).

NiLuJe:
-Wpadded offers some inisght as to what actually happens. Prefer a recent Clang version, as it will actually print the exact amount of bytes, instead of telling you "x was aligned to y for z" ;).

NiLuJe:
Note, that type-punning via unions is also potentially undefined behavior in C++ (while in C, it's explicitly allowed since C99), so that union becomes even more iffier ;).

Frenzie:
@poire-z 

> For your tests issues, well, I don't know, I'm not even really sure what is it that is bothering, as I'm not as deep in the scenario/order of tests (and you didn't explain much, and assumed it was obvious and I would understand :)

The problem is that some of the tests depend on a certain running order in spite of settings being wiped. (Occasionally there's some unintentional `G_reader_settings` spillover — not sure why we don't just make that test-local anyway, but I think the cache is the only thing that purposefully carries over as supposed to be harmless on the one hand & helpful to speed things up on the other.) Also there's no guarantee that the correct tests will end up together in the same CircleCI container.

But the main problem is test development. If I randomly feel like doing some ReaderView tests, for example, I don't want to wait 1-2 minutes for all of them to complete. I just want to test the stuff I'm working on. But if you try that, you might get stuff randomly breaking because a test that seems to depend on default values actually depends on custom values set by a previous test. See #4970 for an example.

> Does it work the same if you had set readerui.document:setSpaceCondensing(50 or 100)?

No, those are broken. Values other than 75 are what causes the problems. But I'm pretty sure it's not only space condensing that can cause problems.

> I don't see why this would not cause problems in daily usage, but just with tests.

Just because the page numbers are different than the test expects doesn't mean they're wrong. My assumption is that the whole document is "wrong."

poire-z:
@Frenzie : well, I don't seem to be able to get any failure on my emulator machine.
I checked out 4dc116ed6a43e8ef625a3b96bb1740d309be2eca (commit just before this PR), and included your first commit of this PR (3fb4342ea3a2b15a2e32456ebdfebe75ea40d6f2).
Running:
```
./kodev test front readerfooter_spec.lua
./kodev test front readerview_spec.lua
```
(in any order, and cleaning emulator cr3cache between pair-runs).

(I haven't upgraded any of my debian busted/lurocks stuff.)

Frenzie:
That was probably a typo. I may have simply meant readerrolling → readerview, which I identified as causing the error on the first commit.

I think readertoc_spec should error if you run it all by itself (without running some other things prior). In fact in that scenario I think it only succeeds if you set it to 100.

Note that there are two separate issues:

1. Some spillover of stuff (mainly due to require) between tests when running them all at once. This is mildly annoying but not too bad.
2. As of yet unexplained spillover of stuff between tests that's affected by cache when supposedly it shouldn't be. Cache could be deleted in teardown but that would just be sweeping it under the rug.

poire-z:
Well, I can't get any failure when running any pair or triplet of readerrolling, readerview, readertoc, readerfooter_spec.lua in various orders...

> I think readertoc_spec should error if you run it all by itself

Indeed, but, still with:
> I checked out 4dc116e (commit just before this PR), and included your first commit of this PR (3fb4342)

if I just don't include the 2 last chunks of your first commit with your changes in tests expected value, I get no failure :)

But OK, I have then one combination that ends up in failure:
readertoc standalone: no failure
readerview > readertoc: no failure
readerrolling > readertoc: no failure
readerfooter > readertoc: FAILURE

If I then add the 2 last chunks from your commit, it does fix this FAILURE (but then, the 3 others do fail now).

So, is that what you observed?
And the point is just to understand what happens in the readerfooter > readertoc order?

Frenzie:
> And the point is just to understand what happens in the readerfooter > readertoc order?

If it helps to phrase it that way, sure. ;-) The point is all failures you mentioned because they're all the same. Whether or not they currently occur is immaterial. That just means the error condition should be distilled into a test.

Btw, are completely different file sizes still plausibly explained by padding?

Screenshot_20190427_145400

poire-z:
> Btw, are completely different file sizes still plausibly explained by padding?

Nope. How did that happen ?
I guess for that to happen, there must be some style change, or some nodes added (so, styles changes, inline/block, autoBoxing...)... so, some settings change.

> The point is all failures you mentioned because they're all the same. Whether or not they're currently likely to occur is immaterial.

Well, I did not follow (and have little interest in) your latest busted/unittests changes.
And the tests never failed previously. Why?

So, maybe because of the removal of `--no-auto-insulate`? But what does that do?

> If it helps to phrase it that way, sure. ;-) 

It helps me localizing one "material" point, if we want to understand what happens. (again, I don't get much of the tests infrasctructure, insulation and stuff - so talk to me in terms of what I know, the reader, please :)
Anyway, this `readerfooter > readertoc` ordered pair test fail with 2c6cc258a2d2d96d1a9701219c04fe7caec3b9bb, before any busted/insulation stuff.
So, I guess if there's some error in our reader code, or some needed stuff that is implicitely done by readerui & children in the context of the reader, but not when called in a unit test, that's what we want to know and understand, so we can explicitely do what's needed so it works the same in the tests.
Am I right in suspecting something like that and phrasing it that way ? (if I'm off topic :) I won't bother investigating :)

poire-z:
@NiLuJe : confirmed that on my 32bits Kobo, CacheFileItem needs the padding. And we only need that to have reproducible cache files - no need to mess with the ldomNode stuff.

So, I won't mess with the ldomNode union/struct/union :)

What I could do if want reproducible cache files on 64bits emulator is to massage the serialized structs to reset to 0 the padding area (if we are on 64bits) before it is written to the cache file. Would cost a small increase in cputime.

Would just `#if __SIZEOF_POINTER__ == 8     massage    #endif` work wtith all our compilers ?
Or should I just use `if (sizeof(void*) == 8) { massage }` ?

poire-z:
@NiLuJe : just asking: `char pad[4]` is good for readability. I would be using `lUint32 _padding`, because it's simpler for the initialization: `_padding(0)`.
How would you go to initialize the pad[4] ? `pad[0](0), pad[1](0) `... ? `pad('\0\0\0\0')` (guess that would not work :))

Frenzie:
> Nope. How did that happen ?
I guess for that to happen, there must be some style change, or some nodes added (so, styles changes, inline/block, autoBoxing...)... so, some settings change.

readerrolling creates a different (smaller) cache. If you follow it by readerfooter or readertoc it rerenders and grows in size. I guess it has to do with that landscape thing.

readerrolling → readerfooter → readertoc = failure

Incidentally, not really related too much but I seem to get a 1 kiB cache if I use:

```lua
        readerui.document:invalidateCacheFile()
        readerui:reloadDocument()
```

> And the tests never failed previously. Why?

They've always failed like this. They currently fail less than ever before when running them in isolation or in a random order. Test interdependence is at an all time low.

> Am I right in suspecting something like that and phrasing it that way ? (if I'm off topic :) I won't bother investigating :)

You mean if the reader does something extra to invalidate the cache? I don't know; the unit tests also trigger crengine's built-in invalidation when relevant most of the time.

poire-z:
>  I guess it has to do with that landscape thing.

OK, if you switch orientation, it may change in size (different number of pages info to cache).

> You mean if the reader does something extra to invalidate the cache? I don't know; the unit tests also trigger crengine's built-in invalidation when relevant most of the time.

For me, whether you re-use saved G_settings_reader or not, and whether there is a cache or not, ReaderUI and its modules (even in tests) should do enough readSetting and call enough CreDocument:setStuff() so crengine gets all initialized correctly, and crengine should be able to notice alone if the cache is to be resused or not.
There shouldn't be any need to invalidateCache from the reader side (this is only needed when toggling the alternative TOC, or switching Styletweaks that involve `display:` and may need a DOM rebuilt) or from the unittests side.

So, I mean may be in _some of_ our tests, our use of ReaderUI could not do enough sendEvent or method calls, so crengine starts with its default settings and so consider a current cache file valid.
In our Reader app, we may have ReaderUI does all that more correctly and fully - and this is why we don't notice such issue in normal use.
(That's pure speculation, haven't had time yet to investigate.)

NiLuJe:
@poire-z: `__SIZEOF_POINTER__` is supported both by GCC & Clang (at least w/ Clang 8) ;).

But does 64bits really need another tweak?

----

I'd just use C++11 initializers, i.e., `char pad[4]{0}` in the struct's declaration.

Frenzie:
> OK, if you switch orientation, it may change in size (different number of pages info to cache).

Sure, but I'd expect it to be the same again after switching back to portrait. Apparently that's wrong. :-)

> For me, whether you re-use saved G_settings_reader or not, and whether there is a cache or not, ReaderUI and its modules (even in tests) should do enough readSetting and call enough CreDocument:setStuff() so crengine gets all initialized correctly, and crengine should be able to notice alone if the cache is to be resused or not.

Exactly! :-)

@poire-z One problem is that there's a discrepancy between the test and the emulator.

Test:

```
04/27/19-17:59:16 DEBUG CreDocument: set view dimen {
    ["h"] = 800,
    ["w"] = 600
}
```

Emulator:

```
04/27/19-17:59:51 DEBUG CreDocument: set view dimen {
    ["h"] = 720,
    ["w"] = 540
}
```

To get the same size you can set it on the command line or use -s=kindle3. Or alternatively, reduce the size for the unit tests. That way you won't get a forced cache invalidation if you try to reproduce things in the GUI (e.g., readerfooter → emulator).

I find this bothersome. Changing the tests is not a very realistic option, but changing the default to 600×800 probably is? I'm not even aware of any ereaders that have a lower resolution than 600×800 tbh. Is there some underlying thought that 540×720 still works on a 1024×600 display, perhaps?

Frenzie:
In fact readerfooter → emulator has some interesting results to be sure:

Screenshot_20190427_182439

Frenzie:
Sorry, I should've added that this is a minimal but very problematic readerfooter, with only this test remaining:

```lua
    it("should disable footer if settings.disabled is true", function()
        local sample_epub = "spec/front/unit/data/juliet.epub"
        purgeDir(DocSettings:getSidecarDir(sample_epub))
        os.remove(DocSettings:getHistoryPath(sample_epub))
        UIManager:quit()

        G_reader_settings:saveSetting("footer", { disabled = true })
        local readerui = ReaderUI:new{
            document = DocumentRegistry:openDocument(sample_epub),
        }
        local footer = readerui.view.footer

        assert.falsy(readerui.view.footer_visible)
        assert.truthy(footer.onCloseDocument == nil)
        assert.truthy(footer.mode == 0)
        
                purgeDir(DocSettings:getSidecarDir(sample_epub))
        os.remove(DocSettings:getHistoryPath(sample_epub))
        UIManager:quit()
    end)
```

poire-z:
> Sure, but I'd expect it to be the same again after switching back to portrait. Apparently that's wrong. :-)

Well, it looks like it might not be truncated to zero when re-saving over an existing one. So, it might only grow and never decrease.
portrait (314k) > landscape (314K) > portrait (314K)
landscape (313k) > portrait (318K) > landscape (318K)
Might also be some stuff re-used, some block aligment in play (explaining that 318K..?)

poire-z:
> this is a minimal but very problematic readerfooter

OK, it seems the failure of `readerfooter > readertoc` is indeed due to the optimization done in https://github.com/koreader/koreader-base/pull/601#discussion_r171065144.
If you remove `G_reader_settings:saveSetting("footer", { disabled = true })` from that test, no failure.
I guess that when just doing the footer test, and disabling some stuff, we have a situation when ReaderUI has never called cre.cpp `getTableOfContent()` and `LVDocView::getToc()`, so crengine has in memory a half-made TOC (without page numbers, which are calculated only when requested, as it may be expensive). So, it saves  this half made TOC in the cache. When readertoc has crengine load from cache, the optimisation makes it trust the TOC from cache, and so the page numbers are not calculated: they are then all set to 1 (as your screenshot shows).

If I just patch crengine with this:
```diff
--- a/crengine/src/lvdocview.cpp
+++ b/crengine/src/lvdocview.cpp
@@ -778,7 +778,7 @@ LVTocItem * LVDocView::getToc() {
         // Avoid calling updatePageNumbers() in that case (as it is expensive
         // and would delay book opening when loaded from cache - it will be
         // called when it is really needed: after next full rendering)
-        if (!m_doc->isTocFromCacheValid())
+//        if (!m_doc->isTocFromCacheValid())
             updatePageNumbers(m_doc->getToc());
        return m_doc->getToc();
 }
```
`readerfooter > readertoc` does not fail.

@Frenzie : are you able to run the tests locally and have them failed with the various ordering? If yes, can you try with that simple crengine patch, to see if this problem is the only cause (and only thing to fix) to all your tests issues?

Then 2 options:
- quicker: don't end readertoc with that disabled test - but end it with some other test that requests the TOC. Or just `invalidateCacheFile()` at the end of readertoc.
- more correct: try to see that the TOC is not really valid before trusting it when loading from cache. I have no idea yet how that will go, but it should be doable - and who knows, it might solve https://github.com/koreader/koreader/issues/4861#issuecomment-478312940 that I have never reproduced.

> One problem is that there's a discrepancy between the test and the emulator

Not a problem for me, I thought we could just debug this by just running tests, which is what I just quickly did.
(But being able to run the tests with `--debug` and have that activate debug logs of our code would be nice :) I just had to add `DEBUG:turnOn()` in setup() to get them thus.)

Frenzie:
> are you able to run the tests locally and have them failed with the various ordering? If yes, can you try with that simple crengine patch, to see if this problem is the only cause (and only thing to fix) to all your tests issues?

It definitely wouldn't fix all, but it might get rid of the mysterious ones. ;-)

Frenzie:
@poire-z Confirmed to resolve the mysterious cache-related stuff.

NB The lack of TOC is only in this specific test. With the whole readerfooter_spec it caused incorrect page numbers exactly as described in #4861.

Screenshot_20190428_195603
Screenshot_20190428_195607
Screenshot_20190428_195614
Screenshot_20190428_195620

> more correct: try to see that the TOC is not really valid before trusting it when loading from cache. I have no idea yet how that will go, but it should be doable - and who knows, it might solve #4861 (comment) that I have never reproduced.

How expensive would it be to simply check if the xpointer of the final TOC entry is in fact present on the relevant page?

poire-z:
Thanks for the checks!

> How expensive would it be to simply check if the xpointer of the final TOC entry is in fact present on the relevant page?

I don't know, But I'd first like to understand the root cause. There might be other obvious ways to know some facts (like the root TOC item having `_percent=0` when pages have neved been calculated, `_percent=-1` when they have been).

> With the whole readerfooter_spec it caused incorrect page numbers exactly as described in #4861

How did you get into having these screenshots?
Running the tests, and launching the emulator then (so, re-using the cr3cache from the tests)?
If yes, with the same geometry and settings?
If yes, did it *has to* have the same geometry - and all would be fine if they did not have the same geometry?

It's possible that I assumed that when loading from cache, the cache can only be in sync with the previous opening settings as we re-use them and they couldn't be changed elsewhere (which is obviously a wrong assumption with tests.) But I don't clearly see why/how the mess happens. It looks to me that if geometry changes (or any other rendering setting), this is noticed and a full rendering is done, and page numbers then calculated next time getToc() is called.

It looks like the readerfooter test with `disabled = true` is testing something that never happens currently with the reader app, and it looks that it's only then that getToc() is never called.
If you can, can you check your #4861 test case with just that `disabled = true` test disabled?
Might help me pinpoint the problem.

Frenzie:
> Running the tests, and launching the emulator then (so, re-using the cr3cache from the tests)?
> If yes, with the same geometry and settings?
> If yes, did it has to have the same geometry - and all would be fine if they did not have the same geometry?

Like I said [above](https://github.com/koreader/koreader/pull/4972#issuecomment-487299449), you need to use the same dimensions (600x800) or it'll trigger an invalid cache.

poire-z:
So, can I assume an implicit _yes_ to my other questions? (Please be explicite, I'm kind impliciteness-blind :)

Was that with or without my `updatePageNumbers(m_doc->getToc());` patch from above?
Explicitely: does this patch solve or not the messed up page numbers?

Frenzie:
Yes, that patch solves the problem. It probably got lost a bit there above the screenshots, but I said so right [here](https://github.com/koreader/koreader/pull/4972#issuecomment-487401919). ;-)

> So, can I assume an implicit yes to my other questions? (Please be explicite, I'm kind impliciteness-blind :)

I'm not sure I fully understand them.

There are no settings, so the same (default) settings.
A different resolution makes crengine say the cache is invalid. You need to launch the emulator in 600x800 in order to reuse the cache from the unit tests.

readerfooter → emulator (in 600x800) is all there is to it.

poire-z:
> readerfooter → emulator (in 600x800) is all there is to it.

I don't seem to be able to reproduce what yous screenshots show with this.
But I can fix the readerfooter_spec > readertoc_spec with this, which feels like it may solve readerfooter → emulator:
```diff
--- a/crengine/include/lvtinydom.h
+++ b/crengine/include/lvtinydom.h
@@ -2009,6 +2009,11 @@ public:
     /// For use on the root toc item only (_page, otherwise unused, can be used to store this flag)
     void setAlternativeTocFlag() { if (_level==0) _page = 1; }
     bool hasAlternativeTocFlag() { return _level==0 && _page==1; }
+
+    /// When page numbers have been calculated, LVDocView::updatePageNumbers()
+    /// sets the root toc item _percent to -1. So let's use it to know that fact.
+    bool hasValidPageNumbers() { return _level==0 && _percent == -1; }
+    void invalidatePageNumbers() { if (_level==0) _percent = 0; }
 };


--- a/crengine/src/lvdocview.cpp
+++ b/crengine/src/lvdocview.cpp
@@ -778,8 +778,10 @@ LVTocItem * LVDocView::getToc() {
         // Avoid calling updatePageNumbers() in that case (as it is expensive
         // and would delay book opening when loaded from cache - it will be
         // called when it is really needed: after next full rendering)
-        if (!m_doc->isTocFromCacheValid())
+        printf("<TOC root: page = %d percent = %d\n", m_doc->getToc()->_page, m_doc->getToc()->_percent);
+        if (!m_doc->isTocFromCacheValid() || !m_doc->getToc()->hasValidPageNumbers())
             updatePageNumbers(m_doc->getToc());
+        printf(">TOC root: page = %d percent = %d\n", m_doc->getToc()->_page, m_doc->getToc()->_percent);
        return m_doc->getToc();
 }

--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -3934,6 +3934,9 @@ int ldomDocument::render( LVRendPageList * pages, LVDocViewCallback * callback,
     if ( !_rendered ) {
         setCacheFileStale(true); // new rendering: cache file will be updated
         _toc_from_cache_valid = false;
+        // force recalculation of page numbers (even if not computed in this
+        // session, they will be when loaded from cache next session)
+        m_toc.invalidatePageNumbers();
         pages->clear();
         if ( showCover )
             pages->add( new LVRendPageInfo( _page_height ) );
```

Could you have a try with it?

Frenzie:
> But I can fix the readerfooter_spec > readertoc_spec with this, which feels like it may solve readerfooter → emulator:

Any custom global settings in the way? The emulator screenshots merely illustrate more visually the exact same thing as the unit test, which does it with a few boring numbers.

I'll try your patch tomorrow. Looks like it should do the trick. :+1:

poire-z:
> `__SIZEOF_POINTER__` is supported both by GCC & Clang (at least w/ Clang 8) ;).
But does 64bits really need another tweak?

Not really _need_, but I think it's nice if we get same-hash cache files when the same settings are used, just to have a quick way to check caches sameness.

I'll go with this, which ensures that on both 32bits and 64bits builds, if that looks good to you:
```diff
--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -399,6 +399,9 @@ struct CacheFileItem
     lUInt64 _dataHash; // additional hash of data
     lUInt64 _packedHash; // additional hash of packed data
     lUInt32 _uncompressedSize;   // size of uncompressed block, if compression is applied, 0 if no compression
+    lUInt32 _padding;  // explicite padding (this struct would be implicitely padded from 44 bytes to 48 bytes)
+                       // so we can set this padding value to 0 (instead of some random data with implicite padding)
+                       // in order to get reproducible (same file checksum) cache files when this gets serialized
     bool validate( int fsize )
     {
         if ( _magic!=CACHE_FILE_ITEM_MAGIC ) {
@@ -425,6 +428,7 @@ struct CacheFileItem
     , _dataHash(0)          // hash of data
     , _packedHash(0) // additional hash of packed data
     , _uncompressedSize(0)  // size of uncompressed block, if compression is applied, 0 if no compression
+    , _padding(0)           // (padding)
     {
     }
 };
@@ -1852,8 +1856,19 @@ bool tinyNodeCollection::saveNodeData( lUInt16 type, ldomNode ** list, int nodec
         }
         ldomNode buf[TNC_PART_LEN];
         memcpy(buf, list[i], sizeof(ldomNode) * sz);
-        for (int j = 0; j < sz; j++)
+        for (int j = 0; j < sz; j++) {
             buf[j].setDocumentIndex(_docIndex);
+            // On 64bits builds, this serialized ldomNode may have some
+            // random data at the end, for being:
+            //   union { [...] tinyElement * _elem_ptr; [...] lUInt32 _ptext_addr; [...] lUInt32 _nextFreeIndex }
+            // To get "reproducible" cache files with a same file checksum, we'd
+            // rather have the remains of the _elem_ptr sets to 0
+            if (__SIZEOF_POINTER__ == 8) { // 64bits
+                lUInt32 tmp = buf[j]._data._nextFreeIndex; // save 32bits part
+                buf[j]._data._elem_ptr = 0;                // clear 64bits area
+                buf[j]._data._nextFreeIndex = tmp;         // restore 32bits part
+            }
+        }
         if (!_cacheFile->write(type, i, (lUInt8*)buf, sizeof(ldomNode) * sz, COMPRESS_NODE_DATA))
             crFatalError(-1, "Cannot write node data");
     }
```

NiLuJe:
@poire-z: Does simply using a pointer-sized int do the trick?

(Includes a free switch to cstdint now that we enforce C++11 :D).

```diff
diff --git a/crengine/include/lvtinydom.h b/crengine/include/lvtinydom.h
index c5d4ae58..712772a2 100755
--- a/crengine/include/lvtinydom.h
+++ b/crengine/include/lvtinydom.h
@@ -287,7 +287,7 @@ public:
     /// call to invalidate chunk if content is modified
     void modified( lUInt32 addr );
 #endif
-    
+
     /// get or allocate space for rect data item
     void getRendRectData( lUInt32 elemDataIndex, lvdomElementFormatRec * dst );
     /// set rect data item
@@ -680,10 +680,10 @@ private:
         ldomTextNode * _text_ptr;   // NT_TEXT: mutable text node pointer
         tinyElement * _elem_ptr;    // NT_ELEMENT: mutable element pointer
 #if BUILD_LITE!=1
-        lUInt32 _pelem_addr;        // NT_PELEMENT: element storage address: chunk+offset
-        lUInt32 _ptext_addr;        // NT_PTEXT: persistent text storage address: chunk+offset
+        lUIntPtr _pelem_addr;        // NT_PELEMENT: element storage address: chunk+offset
+        lUIntPtr _ptext_addr;        // NT_PTEXT: persistent text storage address: chunk+offset
 #endif
-        lUInt32 _nextFreeIndex;     // NULL for removed items
+        lUIntPtr _nextFreeIndex;     // NULL for removed items
     } _data;
 
 
@@ -1232,7 +1232,7 @@ class ldomDocument;
 
 /**
  * @brief XPointer/XPath object with reference counting.
- * 
+ *
  */
 class ldomXPointer
 {
@@ -2198,7 +2198,7 @@ public:
 #if BUILD_LITE!=1
     /// return selections collection
     ldomXRangeList & getSelections() { return _selections; }
-    
+
     /// get full document height
     int getFullHeight();
     /// returns page height setting
diff --git a/crengine/include/lvtypes.h b/crengine/include/lvtypes.h
index ac39209f..fc4d61c5 100644
--- a/crengine/include/lvtypes.h
+++ b/crengine/include/lvtypes.h
@@ -10,6 +10,7 @@
 #ifndef LVTYPES_H_INCLUDED
 #define LVTYPES_H_INCLUDED
 
+#include <cstdint>
 #include <stdlib.h>
 #include "crsetup.h"
 
@@ -19,30 +20,23 @@
 #include <sys/time.h>
 #endif
 
-#ifdef _WIN32
-typedef long lInt32;            ///< signed 32 bit int
-typedef unsigned long lUInt32;  ///< unsigned 32 bit int
-#else
-typedef int lInt32;            ///< signed 32 bit int
-typedef unsigned int lUInt32;  ///< unsigned 32 bit int
-#endif
+typedef int32_t lInt32;      ///< signed 32 bit int
+typedef uint32_t lUInt32;    ///< unsigned 32 bit int
 
-typedef short int lInt16;           ///< signed 16 bit int
-typedef unsigned short int lUInt16; ///< unsigned 16 bit int
+typedef int16_t lInt16;      ///< signed 16 bit int
+typedef uint16_t lUInt16;    ///< unsigned 16 bit int
 
-typedef signed char lInt8;          ///< signed 8 bit int
-typedef unsigned char lUInt8;       ///< unsigned 8 bit int
+typedef int8_t lInt8;        ///< signed 8 bit int
+typedef uint8_t lUInt8;      ///< unsigned 8 bit int
 
-typedef wchar_t lChar16;            ///< 16 bit char
-typedef char lChar8;                ///< 8 bit char
+typedef wchar_t lChar16;     ///< 16 bit char
+typedef char lChar8;         ///< 8 bit char
 
-#if defined(_WIN32) && !defined(CYGWIN)
-typedef __int64 lInt64;             ///< signed 64 bit int
-typedef unsigned __int64 lUInt64;   ///< unsigned 64 bit int
-#else
-typedef long long int lInt64;       ///< signed 64 bit int
-typedef unsigned long long int lUInt64; ///< unsigned 64 bit int
-#endif
+typedef int64_t lInt64;      ///< signed 64 bit int
+typedef uint64_t lUInt64;    ///< unsigned 64 bit int
+
+typedef intptr_t  lIntPtr;   ///< signed pointer-sized int
+typedef uintptr_t lUIntPtr;  ///< unsigned pointer-sized int
 
 /// platform-dependent path separator
 #if defined(_WIN32) && !defined(__WINE__)
@@ -166,7 +160,7 @@ public:
         return left<=pt.x && top<=pt.y && right>pt.x && bottom > pt.y;
     }
        void clear() { left=right=top=bottom=0; }
-
+
     bool intersect(const lvRect &rc)
        {
         if (left < rc.left)
diff --git a/crengine/qimagescale/qglobal.h b/crengine/qimagescale/qglobal.h
index bce56d48..9a38a6c4 100644
--- a/crengine/qimagescale/qglobal.h
+++ b/crengine/qimagescale/qglobal.h
@@ -41,15 +41,12 @@
 #ifndef QGLOBAL_H
 #define QGLOBAL_H
 
+#include <cstdint>
+
 namespace CRe {
 
-#if defined(_WIN32) && !defined(__GNUC__)
-typedef __int64 qint64;            /* 64 bit signed */
-typedef unsigned __int64 quint64;  /* 64 bit unsigned */
-#else
-typedef long long qint64;           /* 64 bit signed */
-typedef unsigned long long quint64; /* 64 bit unsigned */
-#endif
+typedef int64_t qint64;   /* 64 bit signed */
+typedef uint64_t quint64; /* 64 bit unsigned */
 
 }
```

poire-z:
I don't know, but I just don't feel like messing that much with that :)
Because I don't want to diverge too much from upstream.
And because these `_pelem_addr` might be used for some internal packing of ldomNode into chunks of memory/cacheitems, which might be full of shift (pun intended, binary shifts), that I don't feel at all to risk :)

poire-z:
So, no shame going with my tmp swap ? Shouldn't cost too much compared to everything else that happens on cache save.

Unless you can think of the right AND (or XOR?) to zero the 2nd part that would work no matter endianness and stuff like that :)` buf[j]._data._elem_ptr &= 0xFFFFFFFF;` or `buf[j]._data._elem_ptr &= 0xFFFFFFFF00000000;`

NiLuJe:
Hmm, does a simple zero-initialization help?

```diff
diff --git a/crengine/include/lvtinydom.h b/crengine/include/lvtinydom.h
index c5d4ae58..490f4de1 100755
--- a/crengine/include/lvtinydom.h
+++ b/crengine/include/lvtinydom.h
@@ -287,7 +287,7 @@ public:
     /// call to invalidate chunk if content is modified
     void modified( lUInt32 addr );
 #endif
-    
+
     /// get or allocate space for rect data item
     void getRendRectData( lUInt32 elemDataIndex, lvdomElementFormatRec * dst );
     /// set rect data item
@@ -684,7 +684,7 @@ private:
         lUInt32 _ptext_addr;        // NT_PTEXT: persistent text storage address: chunk+offset
 #endif
         lUInt32 _nextFreeIndex;     // NULL for removed items
-    } _data;
+    } _data{0};
 
 
     /// sets document for node
@@ -1232,7 +1232,7 @@ class ldomDocument;
 
 /**
  * @brief XPointer/XPath object with reference counting.
- * 
+ *
  */
 class ldomXPointer
 {
@@ -2198,7 +2198,7 @@ public:
 #if BUILD_LITE!=1
     /// return selections collection
     ldomXRangeList & getSelections() { return _selections; }
-    
+
     /// get full document height
     int getFullHeight();
     /// returns page height setting
diff --git a/crengine/include/lvtypes.h b/crengine/include/lvtypes.h
index ac39209f..d196086f 100644
--- a/crengine/include/lvtypes.h
+++ b/crengine/include/lvtypes.h
@@ -10,6 +10,7 @@
 #ifndef LVTYPES_H_INCLUDED
 #define LVTYPES_H_INCLUDED
 
+#include <cstdint>
 #include <stdlib.h>
 #include "crsetup.h"
 
@@ -19,30 +20,23 @@
 #include <sys/time.h>
 #endif
 
-#ifdef _WIN32
-typedef long lInt32;            ///< signed 32 bit int
-typedef unsigned long lUInt32;  ///< unsigned 32 bit int
-#else
-typedef int lInt32;            ///< signed 32 bit int
-typedef unsigned int lUInt32;  ///< unsigned 32 bit int
-#endif
+typedef int32_t lInt32;      ///< signed 32 bit int
+typedef uint32_t lUInt32;    ///< unsigned 32 bit int
 
-typedef short int lInt16;           ///< signed 16 bit int
-typedef unsigned short int lUInt16; ///< unsigned 16 bit int
+typedef int16_t lInt16;      ///< signed 16 bit int
+typedef uint16_t lUInt16;    ///< unsigned 16 bit int
 
-typedef signed char lInt8;          ///< signed 8 bit int
-typedef unsigned char lUInt8;       ///< unsigned 8 bit int
+typedef int8_t lInt8;        ///< signed 8 bit int
+typedef uint8_t lUInt8;      ///< unsigned 8 bit int
 
-typedef wchar_t lChar16;            ///< 16 bit char
-typedef char lChar8;                ///< 8 bit char
+typedef wchar_t lChar16;     ///< 16 bit char (NOTE: Actually 32 bit on Linux!)
+typedef char lChar8;         ///< 8 bit char
 
-#if defined(_WIN32) && !defined(CYGWIN)
-typedef __int64 lInt64;             ///< signed 64 bit int
-typedef unsigned __int64 lUInt64;   ///< unsigned 64 bit int
-#else
-typedef long long int lInt64;       ///< signed 64 bit int
-typedef unsigned long long int lUInt64; ///< unsigned 64 bit int
-#endif
+typedef int64_t lInt64;      ///< signed 64 bit int
+typedef uint64_t lUInt64;    ///< unsigned 64 bit int
+
+typedef intptr_t  lIntPtr;   ///< signed pointer-sized int
+typedef uintptr_t lUIntPtr;  ///< unsigned pointer-sized int
 
 /// platform-dependent path separator
 #if defined(_WIN32) && !defined(__WINE__)
@@ -166,7 +160,7 @@ public:
         return left<=pt.x && top<=pt.y && right>pt.x && bottom > pt.y;
     }
        void clear() { left=right=top=bottom=0; }
-
+
     bool intersect(const lvRect &rc)
        {
         if (left < rc.left)
diff --git a/crengine/qimagescale/qglobal.h b/crengine/qimagescale/qglobal.h
index bce56d48..9a38a6c4 100644
--- a/crengine/qimagescale/qglobal.h
+++ b/crengine/qimagescale/qglobal.h
@@ -41,15 +41,12 @@
 #ifndef QGLOBAL_H
 #define QGLOBAL_H
 
+#include <cstdint>
+
 namespace CRe {
 
-#if defined(_WIN32) && !defined(__GNUC__)
-typedef __int64 qint64;            /* 64 bit signed */
-typedef unsigned __int64 quint64;  /* 64 bit unsigned */
-#else
-typedef long long qint64;           /* 64 bit signed */
-typedef unsigned long long quint64; /* 64 bit unsigned */
-#endif
+typedef int64_t qint64;   /* 64 bit signed */
+typedef uint64_t quint64; /* 64 bit unsigned */
 
 }
```

poire-z:
`_data{0};                     // Zero-init to ensure padding is zero when pointers are wider than uint32_t (i.e., on x64)`

Nope, that won't do. They have a life of their own after init, and they can switch from being used with one of the thing in the union to another.
I really don't want to do anything fancy withat them :)

NiLuJe:
Gah. Then, yeah, no better idea ;).

poire-z:
> I'll try your patch tomorrow. Looks like it should do the trick.

@Frenzie : did you get the time to test it?

Frenzie:
@poire-z Not yet, but I've got 10 minutes.

Frenzie:
@poire-z That does seem to work around the problem, provided we delete the old cache first (as would be done by bumping the cache version).

poire-z:
You mean it just _did not work_ on your first run (because the toc pages numbers were still marked as valid)? And next full runs worked ok?
(Just the rendering cache version - as we do often - and not the one dom reload version, right?)

Frenzie:
I mean it works fine if you start from scratch, but not if you start from a cache borked by a pre-patch crengine.

poire-z:
OK, thanks. I'll push that in my next bunch of crengine stuff, probably after "Probably Not Labor Day" release :)
(We'll then have to revert the 2nd commit from this PR, right?)

Frenzie:
Yes, it'll no longer be useful.

PR #4973, opened 2019-04-26
kodev: pass correct --lpath for base test

Frenzie:


kodev
```diff
@@ -749,16 +749,18 @@ OPTIONS:
 
     make "${EMU_DIR}/.busted"
     pushd "${EMU_DIR}" && {
-        test_path="./spec/$1/unit"
-        rm -rf "${test_path}"/data/*.sdr
+        test_path_base="./spec/$1/unit"
```
Frenzie: On second thought this may not be the best variable name. Any suggestions @NiLuJe @poire-z ? It's intended as just the path without a filename.

poire-z: Dunno. Something with `basedir`, `basepath `, `dirname`?

Frenzie: Thanks!

PR #4974, opened 2019-04-26
[CI] CircleCI: save test results

Frenzie:
It should be a path to a directory, not a file.

https://circleci.com/docs/2.0/configuration-reference/#store_test_results

> Path (absolute, or relative to your working_directory) to directory containing subdirectories of JUnit XML or Cucumber JSON test metadata files

Perhaps the file needs to be named results.xml. We'll see.


Frenzie:
Success:

Screenshot_20190426_140629

Issue #4894, opened 2019-04-10
Download images in news (rss) feed articles


Thra11:
* KOReader version: v2019.04

#### Issue / Suggestion
When an article in an RSS feed contains images, it would be nice to download the images and embed them in the article. Particularly with scientific articles, it can be very frustrating when you can download the text, but you can't understand it because the accompanying diagrams are missing.

 I assume the thing to do is:
1. Download the html
2. Parse the html to find img tags and extract image urls
3. Download the image urls
4. Substitute image urls to be relative as required
5. Zip everything up into a suitable container, probably epub.

It look like most of the necessary functionality should already exist in the wikipedia article "Save as EPUB" code, so hopefully we can re-use/copy some of it. (When saving as epub with images, does it get the "normal" web html for the wikipedia page? or does it get it in an easier-to-parse form?)

Might be worth adding an optional parameter to the feed specification in feed_config.lua, as downloading images might not be desirable for every feed.

I'm planning to have a go at implementing this myself if no-one else does, but I'm not particularly familiar with lua or the koreader code, so any pointers or suggestions from those who know it better would be welcome.

There are a couple of open issues which are related: #3065, #3516.

Thra11:
Looks like the wikipedia code is parsing html, but it does take advantage of some wikipedia-specific knowledge to select or reject certain elements based on class.

poire-z:
> It look like most of the necessary functionality should already exist in the wikipedia article "Save as EPUB" code, so hopefully we can re-use/copy some of it. (When saving as epub with images, does it get the "normal" web html for the wikipedia page? or does it get it in an easier-to-parse form?)

We don't get the normal web page html, but some HTML from Wikipedia API (the same that, I think, is used to make the web pages, where navigation and other stuff is added to it).
So, it's indeed some quite clean and "normalized" HTML, and after looking at it, we could guess what we can remove or modify with some simple regexps.

Anyway, I think wikipedia.lua is a good starting point to get what you can do with Lua and the few things we already have to help building an EPUB:
https://github.com/koreader/koreader/blob/fd50dc34a6922eed97c69e3ec928694caebe6bd0/frontend/ui/wikipedia.lua#L648-L649
https://github.com/koreader/koreader/blob/fd50dc34a6922eed97c69e3ec928694caebe6bd0/frontend/ui/wikipedia.lua#L817-L818
[...]

PR #4982, opened 2019-04-29
NewsDownloader: epub with images

Thra11:
Implements #4894

### DIFF
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -20,10 +20,10 @@ local EpubDownloadBackend = {
 }
 local max_redirects = 5; --prevent infinite redirects
 
-function EpubDownloadBackend:download(url, path)
+function EpubDownloadBackend:download(url, path, include_images)
     logger.dbg("EpubDownloadBackend:download")
 --    self:createEpub(path, url)
-    self:createEpubWithUI(path, url, function(success)
+    self:createEpubWithUI(path, url, include_images, function(success)
         if (success) then
             logger.dbg("createEpubWithUI success")
         else
@@ -189,9 +189,8 @@ local ext_to_mimetype = {
 }
 
 -- Create an epub file (with possibly images)
-function EpubDownloadBackend:createEpub(epub_path, url)
+function EpubDownloadBackend:createEpub(epub_path, url, include_images)
     logger.dbg("EpubDownloadBackend:createEpub(", epub_path, ",", url, ")")
-    local with_images = true
     -- Use Trapper to display progress and ask questions through the UI.
     -- We need to have been Trapper.wrap()'ed for UI to be used, otherwise
     -- Trapper:info() and Trapper:confirm() will just use logger.
@@ -298,7 +297,6 @@ function EpubDownloadBackend:createEpub(epub_path, url)
     logger.dbg("Images found in html:", images)
 
     -- See what to do with images
-    local include_images = true
     local use_img_2x = false
     if not include_images then
         -- Remove img tags to avoid little blank squares of missing images
@@ -532,7 +530,7 @@ end
 
 -- Wrap EpubDownloadBackend:createEpub() with UI progress info, provided
 -- by Trapper module.
-function EpubDownloadBackend:createEpubWithUI(epub_path, url, result_callback)
+function EpubDownloadBackend:createEpubWithUI(epub_path, url, include_images, result_callback)
     logger.dbg("EpubDownloadBackend:createEpubWithUI(", epub_path, ",", url, ",", title, ", ...)")
     -- To do any UI interaction while building the EPUB, we need
     -- to use a coroutine, so that our code can be suspended while waiting
@@ -545,7 +543,7 @@ function EpubDownloadBackend:createEpubWithUI(epub_path, url, result_callback)
         -- Trapper) would just abort (no reader crash, no error logged).
         -- So we use pcall to catch any errors, log it, and report
         -- the failure via result_callback.
-        local ok, success = pcall(self.createEpub, self, epub_path, url)
+        local ok, success = pcall(self.createEpub, self, epub_path, url, include_images)
         if ok and success then
             result_callback(true)
         else
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -1,0 +1,559 @@
+local Version = require("version")
+local ffiutil = require("ffi/util")
+local http = require("socket.http")
+local https = require("ssl.https")
+local logger = require("logger")
+local ltn12 = require("ltn12")
+local socket = require("socket")
+local socket_url = require("socket.url")
+local _ = require("gettext")
+local T = ffiutil.template
+
+local EpubDownloadBackend = {
+   -- Can be set so HTTP requests will be done under Trapper and
+   -- be interruptible
+   trap_widget = nil,
+   -- For actions done with Trapper:dismissable methods, we may throw
+   -- and error() with this code. We make the value of this error
+   -- accessible here so that caller can know it's a user dismiss.
+   dismissed_error_code = "Interrupted by user",
+}
+local max_redirects = 5; --prevent infinite redirects
+
+function EpubDownloadBackend:download(url, path)
+    logger.dbg("EpubDownloadBackend:download")
+--    self:createEpub(path, url)
+    self:createEpubWithUI(path, url, function(success)
+        if (success) then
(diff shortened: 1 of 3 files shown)


plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -0,0 +1,524 @@
...
+    self.trap_widget = trap_widget
+end
+
+function EpubDownloadBackend:resetTrapWidget()
+    self.trap_widget = nil
+end
+
+function EpubDownloadBackend:loadPage(url)
+    local completed, success, content
+    if self.trap_widget then -- if previously set with Wikipedia:setTrapWidget()
```
Frenzie: You should probably change this comment. ;-)

Thra11: Done :)

Thra11:
> Is `EpubDownloadBackend:createEpub()` the part that differs most from the Wikipedia code?

From memory, the main differences are:
* Wikipedia starts with an article name and constructs various urls from it, whereas EpubDownloadBackend is given a url to fetch
* In createEpub, there's lots of similar epub boilerplate, but the Wikipedia code also performs a series of substitutions to make the result render more like a wikipedia page and fix known issues.
* Most of the code which parses the html to find and process image urls is unchanged.

I think the core of a shared epub-writing module would probably be a function to take a lua structure representing an epub (metadata, html files, images, css, table of contents, etc.) and write it out as an epub, possibly substituting minimal defaults for elements which aren't provided.

Possible future improvements to this module:
* Parsing \<h*\> to construct a proper table of contents.
* Downloading linked css and including it.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -0,0 +1,524 @@
...
+    -- (crengine only uses <meta name="cover" content="cover-image" /> to get the cover image)
+    --   In <spine toc="ncx"> :
+    --     <itemref idref="cover" linear="no"/>
+    --   And a <guide> section :
+    --     <guide>
+    --       <reference href="title.html" type="cover" title="Cover"/>
+    --       <reference href="toc.html" type="toc" title="Table of Contents" href="toc.html" />
+    --     </guide>
+    local koreader_version = "KOReader"
+    if lfs.attributes("git-rev", "mode") == "file" then
```
Frenzie: Could you use Version:getCurrentRevision() instead? Same thing, but only one place to change things.

https://github.com/koreader/koreader/blob/96179670f8d1d29f343db1ab0ce20ef6bd45acb2/frontend/version.lua#L7-L22

Frenzie:
@poire-z Do you have any further comments?

poire-z:
No comment after a quick look (but I'm not familiar with the newsdownloader code, and I've forgotten a bit about that Trapper thingy :| but it looks ok, so, if it works, go ahead.)

Just surprised about that comment (in a commit here, and in the other discussion):

> Lift the Trapper:wrap call out of the individual article processing code,
so that articles are processed one by one, in order to:
Avoid concurrent progress updates fighting over the UI dialog
Avoid trying to download many images at the same time

I didn't know we could download things at the same time ! :) I think Trapper still gets stuff one after the other, just allowing to interrupt the whole flow, and indeed displaying progress. But there shouldn't be any fighting :)
So, unless NewsDownloader really download articles in parallele... (does it really do that?!)

Thra11:
My understanding is that `Trapper:wrap` is a ["Simple wrapper function for a coroutine"](https://github.com/koreader/koreader/blob/96179670f8d1d29f343db1ab0ce20ef6bd45acb2/frontend/ui/trapper.lua#L27), so if you call it from the body of a for loop, you end up creating a bunch of coroutines which run concurrently.

Thra11:
I don't think the old newdownloader backend used Trapper inside the backend, so the concurrency was probably introduced accidentally when I took the Wikipedia code with Trapper and put it inside the feed-processing loop.

poire-z:
OK, I see what may have happened.
But (fyi, if I remember that business right):
- coroutines don't run concurrently: they are not threads. Lua is not multi-threaded. Coroutines are just some functions that can pause themselves (when they call `yield`) and can be resumed. Only one piece of the code run at a time, and that piece of code can allow some other piece to be run with `yield()` or `resume()`.
- Trapper tries to make that a little bit easier for us: we run some download, and give control back to the UI, which waits for 100ms for a touch event (for aborting the next step) - if no touch event, it gives the control back to the coroutine code, which goes on with the next download. We can't interrupt in the middle of a download (we have to wait for it to complete, or for the timeout to be reached). But we can between downloads (so, between the article HTML download and the first image download, and between each image download).
- That's the best we could do with Lue being non multithreaded.

So, may be you launched many coroutines, and depending on which gave control back to the UI, and which was resumed, or which got a touch event, you may have had that mess in "fighting for the UI".
So yes, there should be only one coroutine'd flow, with the loops inside it, so things can be kept simple :)

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -0,0 +1,521 @@
+local Version = require("version")
+local ffiutil = require("ffi/util")
```
Thra11: Having looked at the coding style guide and some of the other lua files, I think some of these should be capitalised (`local FFIUtil = ...`) because they're classes?

Thra11:
> coroutines don't run concurrently: they are not threads.

I guess it depends on your definition of concurrency: some people make a distinction between concurrency (bits code which can be safely paused and resumed or executed in a different order, even on a single core, often aiming to keep the system responsive), versus parallelism (splitting a task across multiple cores with the aim of finishing it faster, without any part necessarily being interruptible).

Anyway, semantics aside, it looks like we understand what's going on. I didn't think too hard about where to put the `Trapper:wrap`, just moved it as far out as required to get it out of the loop and out of the individual `processAtom`/`procesRSS` functions to avoid duplicate copies. However, I'm wondering whether it makes sense to be showing `InfoMessage`s via `Trapper:info` and `UIManager:show` in the same module, or whether it make more sense to move the `Trapper:wrap` further out and use `Trapper:info` everywhere.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Frenzie: Either way is fine by me for ffiutil. I think the existing code (most of which predates that addition to the styleguide) may have ffiutil, FFIUtil as well as ffiUtil.

So I'm good to merge, but if you'd like to change it to FFIUtil I'll wait.

poire-z:
> whether it make more sense to move the Trapper:wrap further out and use Trapper:info everywhere.

Having not really looked at NewsDownloader, I would anyway say this is how it should be used.
Trapper.wrap() should be called quite soon after the menu item is hit and the callback/action called, after having checked there are indeed some configuration, feeds, and things to do, before doing the first thing of them.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Thra11: I'll leave it as it is then :)

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -0,0 +1,521 @@
...
+  <metadata>
+    <dc:title>%s</dc:title>
+    <dc:publisher>%s</dc:publisher>
+    %s
+  </metadata>
+  <manifest>
+    <item id="ncx" href="toc.ncx" media-type="application/x-dtbncx+xml"/>
+    <item id="content" href="content.html" media-type="application/xhtml+xml"/>
+    <item id="css" href="stylesheet.css" media-type="text/css"/>
+]], page_htmltitle, Version:getCurrentRevision(), meta_cover))
```
poire-z: Nobody would really see it, but we'll bee be missing the `KOReader` in:
```xml
    <dc:publisher>KOReader v2015.11-1476-g775b7a36</dc:publisher>
``` 
You can just hardcode it in the string `<dc:publisher>KOReader %s</dc:publisher>`

Thra11: Fixed! Thanks.

Frenzie:
Thanks!

Thra11:
> Trapper.wrap() should be called quite soon after the menu item is hit and the callback/action called, after having checked there are indeed some configuration, feeds, and things to do, before doing the first thing of them.

Looks like I didn't test this well enough: it no longer interleaves articles in a feed, but it will interleave feeds if you have more than one.

I'll change it to use Trapper the way you suggest. (Can I add new commits to this PR or do I create a new one?)

NiLuJe:
I'd go with a new PR, from a new, rebased branch ;).

PR #4987, opened 2019-04-30
NewsDownloader: Fix UI when downloading multiple feeds

Thra11:
Wrap the whole of loadConfigAndProcessFeeds with Trapper:wrap so that feeds are downloaded sequentially, otherwise they become interleaved and the UI progress updates get confusing. Use Trapper:info in place of manually constructed InfoMessages.

Improves #4894 by fixing issues discussed in #4982


poire-z:
Looks fine to me. Ready for merge?

PR #4989, opened 2019-05-01
Blacklist the C BB on devices where it cannot be used

NiLuJe:
Either at all, or because it would impose reduced functionality.

Meaning:
* On 4bpp fb, because those are completely unsupported by the C BB
* On 24bpp fb, because it has limited support for that target format (no alpha blending).
* On 8bpp Linux eInk fb with an inverted palette (that's basically only the K4 ;)).
* On Linux eInk devices w/ no HW inversion support, as that would render NightMode inoperable.

That last point means that it's currently *always* blacklisted on PocketBook, because I don't have a PB device, so I don't know how safe HW inversion would be, meaning none of them are marked as being HW invert capable.

(I was a tiny bit hasty with https://github.com/koreader/koreader-base/pull/912 ;).)


poire-z:
> On Linux eInk devices w/ no HW inversion support, as that would render NightMode inoperable.

That sounds like you would disable it on older Kobo (like my GloHD), althought the code looks like you don't.
(Haven't checked if it is really used, but I understood it would indeed be used since today's nightly :)

NiLuJe:
I did forget about the Kobo case, good point :D.

Although, the only device flagged as unsafe for HW inversion is the Aura, the GloHD should be okay ;).

PR #4990, opened 2019-05-01
CRe margins: add toggle to sync top and bottom margins

poire-z:
Follow up to #4945, see https://github.com/koreader/koreader/pull/4945#issuecomment-488276218 for the reasoning.

<kbd>image</kbd>

<kbd>image</kbd>

Suggestions for some alternate title and wording?

Also fix DogEar size update on margin change, see https://github.com/koreader/koreader/pull/4945#issuecomment-488271564

(Still the tab icon order that is bothering me :) https://github.com/koreader/koreader/pull/4945#issuecomment-488266529 )


NiLuJe:
Feel free to re-order the tabs, BTW, I have no strong opinion on that ;).

frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -347,17 +351,47 @@ end
...
-    self:onSetPageMargins(self.unscaled_margins)
+    if self.sync_t_b_page_margins then
+        self.unscaled_margins[2] = b_margin
+        -- Let ConfigDialog know so it can update it on screen and have it saved on quit
+        self.ui.document.configurable.t_page_margin = b_margin
+    end
+    self.ui:handleEvent(Event:new("SetPageMargins", self.unscaled_margins))
+end
+
+function ReaderTypeset:onSyncPageTopBottomMargins(toggle)
```
NiLuJe: You went the extra mile there ;).

I probably wouldn't have made it somewhat "retroactive" like that, instead just relying on what the user would next touch (top or bottom) and sync to that, but I get the idea, and hey, why the hell not ;).

Plus, it should essentially be a NOP with default values anyway ;).

poire-z:
OK, re-ordered.
Also added the hint about "Reclaim bar height..." in the bottom margin help_text too. (I always have a hard time with saying stuff like `in the top menu's Status bar submenu` so if that sounds tedious, please suggest an alternative wording :)

NiLuJe:
Random remark about the layout after having tested it on a 300dpi device, like on your screenshot: I like the -/+ buttons having a slightly larger padding, to set them apart from the progress bar, but it does make 'em jut out a bit.

I personally don't mind it, and how much (or whether they actually do, on a 265 H2O, that made the mismatch switch from +2, +1 to 0, +1, IIRC) depends entirely on the screen width and the DPI, but since it had came up in the earlier discussions, here we go ;).

NiLuJe:
I think I preferred the wording from your screenshot, FWIW ;).

frontend/apps/reader/modules/readertypeset.lua
poire-z: Yes :) And it's quite useful to rebalance the margin tweaks.

It ended up being an unexpected bonus in my workflow:
- start with them sync'ed, see what you can get
- if you only get the missing line when they are recuded too much (I like some top margin), do one step back (so, missing line back), and increase just the top margin till things get centered)

poire-z:
> about the layout after having tested it on a 300dpi device, like on your screenshot: I like the -/+ buttons having a slightly larger padding, to set them apart from the progress bar, but it does make 'em jut out a bit.

My screenshots were on the emulator with a small dpi, but yes, I noticed on my Kobo the new bars are a bit wider than the others. But as the other progressbars are a bit shorter than the toggles, it should be fine. (I did notice it, but given how tedious solving it would be in the complex ConfigDialog, I decided to no more notice it :)

We could interleave the new toggle between the different kind of progress bar so that it's less noticable:

<kbd>image</kbd>

> I think I preferred the wording from your screenshot

I'll put that first wording back then. Unless @Frenzie comes up with a better one?

Frenzie:
Something like this? Not sure which text I'm supposed to comment on precisely though. :-)

<hr>

Keep top and bottom margins synchronized.
- 'off' allows different top and bottom margins.
- 'on' keeps top and bottom margins locked, ensuring text is vertically centered in the page.
Note that you can select (in the top menu's Status bar submenu) whether the bottom margin should apply from the bottom of the screen, or from above the footer status bar.

poire-z:
> Not sure which text I'm supposed to comment on precisely though. :-)

`in the top menu's Status bar submenu` vs  `in the status bar top menu` vs another wording.

Frenzie:
I think writing more symbolically would be clearer than writing proper sentences. Something like:

In the top menu → Settings → Status bar, you can choose whether the bottom margin applies from the bottom of the screen, or from above the status bar.

poire-z:
OK, will go with your text with the arrows.

Tested on my Kobo having the toggle between the L/R bar and the 2 other bars, and you don't notice the difference in widths (althought it's still there). It also separates the T&B from the other, so your instinct to go hit the top bar to tweak the top margin is a bit prevented :)
But it felt initially a bit ugly.
So, which way to go?

NiLuJe:
@poire-z : Putting the toggle up there was going to be a suggestion, so, eh ;).

Not strongly attached to either positions, but I do like it as a kind of separator, so, let's say in between?

poire-z:
OK, so we end up with:

<kbd>image</kbd>

PR #5000, opened 2019-05-05
CRe margins: hide bottom menu on change

poire-z:
So we can immediately see how's the bottom margin:

<kbd>margins_tap_dismiss</kbd>

(As proposed in https://github.com/koreader/koreader/pull/4945#issuecomment-488571117)

### DIFF
--- a/frontend/apps/reader/modules/readertypeset.lua
+++ b/frontend/apps/reader/modules/readertypeset.lua
@@ -71,7 +71,7 @@ function ReaderTypeset:onReadSettings(config)
         G_reader_settings:readSetting("copt_b_page_margin") or
         DCREREADER_CONFIG_B_MARGIN_SIZES_LARGE
     self.unscaled_margins = { h_margins[1], t_margin, h_margins[2], b_margin }
-    self:onSetPageMargins(self.unscaled_margins, true)
+    self:onSetPageMargins(self.unscaled_margins)
     self.sync_t_b_page_margins = config:readSetting("copt_sync_t_b_page_margins") or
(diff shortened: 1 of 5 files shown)


frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -406,16 +410,19 @@ function ReaderTypeset:onSetPageMargins(margins, silent)
     end
     self.ui.document:setPageMargins(left, top, right, bottom)
     self.ui:handleEvent(Event:new("UpdatePos"))
-    if not silent then
+    if refresh_callback then
         -- Show a toast on set, with the unscaled & scaled values
         UIManager:show(InfoMessage:new{
             text = T(_([[
     Margins set to:
```
Frenzie: @poire-z 
Is this intentionally all indented? It seems like it should be:

```lua
Margins set to

    indented
    indented

Click to dismiss
```

poire-z: That was from @Niluje PR #4945 (to which I added "Tap to dismiss").
(I like it all indented, and I think you once lectured me that if I wanted this visual effect, I'd have to use some left padding - and I went just removing the indentation :)
So, no opinion :)

NiLuJe: I don't *think* I made that decision consciously, FWIW, so, err, no opinion either ;).

Frenzie: In that case I'll get rid of it. It just adds additional complexity to translating the string.

PR #5012, opened 2019-05-08
newsdownloader: rebuild URL on redirect to bare path

avsej:
The server might redirect to only path, in this case the scheme and host
should be taken from original URL


avsej:
For instance, currenly this RSS feed has redirect like this: `https://lobste.rs/t/ruby.rss`

Frenzie:
Thanks!

Issue #5025, opened 2019-05-13
Wallabag can't download some articles


mwoz123:
* KOReader version: v2019.04-80-g4ba7f98_2019-05-05
* Device: Inbook Prime (android)

#### Issue
Can download couple articles from wallabag. Names:
`Inicjatywa Tato.Net | 22 narzędzia, które pomogą Ci wygrać ojcostwo`
`» Czy policja może namierzyć sprawców fałszywych alarmów bombowych? -- Niebezpiecznik.pl…`

Could it be `|` character in first case? coudn't be stored on fat fs ?
but what it could be in second one? I don't think it's `»` as couple similar (same source) were processed e.g. this one was processed: 
`» KeePass — jak zacząć swoją przygodę z managerem haseł? -- Niebezpiecznik.pl --`

#### Steps to reproduce
Add below urls to wallabag : 
https://tato.net/czytelnia/22-narzedzia-ktore-pomoga-ci-wygrac-ojcostwo/
https://niebezpiecznik.pl/post/czy-policja-moze-namierzyc-sprawcow-falszywych-alarmow-bombowych/?more
and then try download them on koreader.

Frenzie:
Wallabag's own export produces the filename `22 narzędzia, które pomogą Ci wygrać ojcostwo.epub`.

Looks like that isn't used at all, but the logic looks sound enough either way:
https://github.com/koreader/koreader/blob/eace8d25c1cbf9bd13e98220098494e8fb63c18f/plugins/wallabag.koplugin/main.lua#L300-L301

The URL `https://tato.net/czytelnia/22-narzedzia-ktore-pomoga-ci-wygrac-ojcostwo/` processes just fine on my H2O (which also uses FAT32).

I can confirm `https://niebezpiecznik.pl/post/czy-policja-moze-namierzyc-sprawcow-falszywych-alarmow-bombowych/?more` to result in failure.

I can't properly investigate atm.

Frenzie:
> the logic looks sound enough

Actually, I think it's not UTF-8 aware, so you get an invalid filename like `[w-id_1955191] » Czy policja może namierzy�.epub`. It's not entirely clear to me why this only sometimes breaks down, but that's probably besides the point.

Frenzie:
@poire-z You've played around with this UTF-8 stuff a bit. Is there a more elegant solution than this?

```diff
diff --git a/plugins/wallabag.koplugin/main.lua b/plugins/wallabag.koplugin/main.lua
[...]

PR #5026, opened 2019-05-14
[fix, chore] Abstract filename logic in util.getSafeFilename()

Frenzie:
Fixes https://github.com/koreader/koreader/issues/5025

The OPDS browser was doing some fancier stuff in a way that should be abstracted away in util (because it applies anywhere files will be saved):

https://github.com/koreader/koreader/blob/eace8d25c1cbf9bd13e98220098494e8fb63c18f/frontend/ui/widget/opdsbrowser.lua#L482-L491

### DIFF
--- a/frontend/ui/widget/opdsbrowser.lua
+++ b/frontend/ui/widget/opdsbrowser.lua
@@ -479,9 +479,8 @@ end
 function OPDSBrowser:downloadFile(item, format, remote_url)
     -- download to user selected directory or last opened dir
     local download_dir = self.getCurrentDownloadDir()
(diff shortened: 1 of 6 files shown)


frontend/ui/widget/opdsbrowser.lua
```diff
@@ -480,13 +480,8 @@ function OPDSBrowser:downloadFile(item, format, remote_url)
     -- download to user selected directory or last opened dir
     local download_dir = self.getCurrentDownloadDir()
     local file_system = util.getFilesystemType(download_dir)
```
Frenzie: Drat, I accidentally dropped a commit.

Issue #5024, opened 2019-05-13
crash on selecting/long press when viewing cbz files


sat3ll:
* KOReader version: v2019.05
* Device: Kindle Fire HD 7" 2nd generation (2012)

#### Issue
When viewing a cbz file, if I long press (accidentally) it selects/highlights some part of the text before crashing.

#### Steps to reproduce
Open a cbz file
Long press/select/highlight somewhere on the content (could be text) (could take a few tries (usually <= 2))

##### `crash.log` (if applicable)
crash.log not found

cramoisi:
Yup, reproduced on my KA1. here is a log

[crash_cbz.txt](https://github.com/koreader/koreader/files/3171841/crash_cbz.txt)

Frenzie:
I'm not entirely sure how one would accidentally highlight something? :-P

cramoisi:
@Frenzie  : falling asleep when reading ;)

PR #5027, opened 2019-05-14
[fix] Don't try to save highlight into non-PDF document

Frenzie:
Fixes <https://github.com/koreader/koreader/issues/5024>.

Highlighting will still work with the .sdr metadata.

### DIFF
--- a/frontend/document/pdfdocument.lua
+++ b/frontend/document/pdfdocument.lua
@@ -142,8 +142,8 @@ function PdfDocument:getPageLinks(pageno)
 end
 
 function PdfDocument:saveHighlight(pageno, item)
-    local _, suffix = util.splitFileNameSuffix(self.file)
-    if suffix ~= "pdf" then return end
+    local suffix = util.getFileNameSuffix(self.file)
+    if string.lower(suffix) ~= "pdf" then return end
 
     self.is_edited = true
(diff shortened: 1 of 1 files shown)


frontend/document/pdfdocument.lua
```diff
@@ -142,6 +142,9 @@ function PdfDocument:getPageLinks(pageno)
 end
 
 function PdfDocument:saveHighlight(pageno, item)
+    local _, suffix = util.splitFileNameSuffix(self.file)
+    if suffix ~= "pdf" then return end
```
poire-z: `if string.lower(suffix) ~= "pdf"`  probably?

Frenzie: Indeed. I wonder where it should go though. (I.e., in this and similar locations or in the function.)

poire-z: Looks like everything related to suffix is done in documentregistry.lua, where I guess that lower(suffix) is done correctly, otherwise we would know :) Elsewhere, we just care about the filename without suffix. So I guess here is fine for this quick workaround.
(Unless you want to support real PDF files with a bofus .tif or .zip extension, anyway opened as a PDF by MuPDF - so we should ask MuPDF for the type of file he has opened :)

Issue #5015, opened 2019-05-09
Feature request: pop-up footnote, endnotes etc..


hiworld3:
* KOReader version: 2019.04
* Device: Android

Hi, could we possible to make the footnote, endnotes etc.. into a small pop-up window? just like the function in Kindle system, Kindle doesn't have this but added it like 4 or 5 years ago, really convenient, no need to jump back and forth in case of reading notes.

Frenzie:
Try Settings → Taps & gestures → Links → Show footnotes in popup.

Alternatively there's also Style tweaks → Miscellaneous → In-page footnotes.

hiworld3:
my lord, thank you

hiworld3:
my lord, thank you

Frenzie:
@poire-z Should we enable one of those by default?

poire-z:
Not sure we should.
_In-page footnotes_ is too dependant on CSS to be made a default.
_Show footnotes in popup_ could, but is also dependant on HTML/CSS quality to be accurate (it may not work on some books, or show other footnotes text, or even whole chapters as footnotes - or need the _Show more links as footnotes_)
The current default of following the link is the only one that works 100% with a coherent behaviour.
(I also like to reward curious people that go browse the menu :)

(Personally, I don't use _Show footnotes in popup_, and prefer following links with the little markers - but I use _In-page footnotes_ when they work, in books where such footnotes aren't too big.)

Frenzie:
> In-page footnotes is too dependant on CSS to be made a default.

I'm not really sure what the argument is there? It just means it wouldn't work in a document unless one of the supported standards is used. :-) E.g., `epub:type="footnote"` that maps to the ARIA role of `doc-footnote`. A footnote that isn't declared a footnote is just a link. That's true regardless whether you use CSS or something hard-coded in C++.

https://github.com/koreader/koreader/blob/b1215812cb99a1e9e8cbb2ce99da94eca12a3324/frontend/ui/data/css_tweaks.lua#L413-L424

I think in-page EPUB footnotes should be enabled by default.
In-page Wikipedia footnotes probably should be too because it's just so specific.
In-page classic classname footnotes are something I enable by default but that may be a bit risky to enable by default.

poire-z:
> I'm not really sure what the argument is there? It just means it wouldn't work in a document unless one of the supported standards is used. :-) 

[...]

PR #5031, opened 2019-05-17
cre: fix 2 footnotes issues

poire-z:
- Popup footnotes: workaround strange behaviour of MuPDF which is adding some unwanted margin above elements with an `id=` attribute. See https://github.com/koreader/koreader/issues/5015#issuecomment-492221431
- In-page footnotes: also remove any left and right margin, to keep footnotes aligned with the main text. See https://github.com/koreader/koreader/issues/5015#issuecomment-492184113
Also add 2 classnames (witnessed in quite a few recent french books) to the list of classic classnames.

Also bump crengine for:
- CSS: fix parsing of multi classnames and attributes selectors https://github.com/koreader/crengine/pull/288

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -419,8 +419,7 @@ This only works with footnotes that have specific attributes set by the publishe
 *[role~="doc-rearnote"]
 {
     -cr-hint: footnote-inpage;
-    margin-top: 0 !important;
-    margin-bottom: 0 !important;
+    margin: 0 !important;
 }
                 ]],
             },
@@ -439,8 +438,7 @@ This only works with footnotes that have specific attributes set by the publishe
 *[role~="doc-rearnote"]
 {
     -cr-hint: footnote-inpage;
-    margin-top: 0 !important;
-    margin-bottom: 0 !important;
+    margin: 0 !important;
     font-size: 80% !important;
 }
                 ]],
@@ -453,8 +451,7 @@ This only works with footnotes that have specific attributes set by the publishe
                 css = [[
 ol.references > li {
     -cr-hint: footnote-inpage;
-    margin-top: 0 !important;
-    margin-bottom: 0 !important;
+    margin: 0 !important;
 }
 /* hide backlinks */
 ol.references > li > .noprint { display: none; }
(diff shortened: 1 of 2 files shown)


poire-z:
Hijacking a random in-page footnote issue to ask a question (could have been a Discussion, but no way.)

I'd like to fix this issue with in-page footnotes:
<kbd>image</kbd>
(I made it "dual page" for easier understanding - but assume for the discussion these are 2 pages - you would not see what's on the other page.)
(Also, this might look quite ok, as it's just text splitted, and actually splitter quite cleanly there - but sometimes it can be the image, or a line cut in the middle - I just couldn't quickly find such a case :)

The float makes the whole yellow block lines be `break:avoid`. If there wasn't any footnote, it would be pushed on page 2 and we'd get a huge blank area at bottom of first page, which is fine.
But the footnote 182 currently cause the current code to abandon this break:avoid.

This would need a footnote to not be pushed immediately, but to wait for its associated "parent" main-text line to be actually pushed to a page before thinking about adding the footnotes to the page the main line will go to.

But asking:
**A** because it feels absurd, but it also feels like I've seen that before: would it be ok to still push such footnotes on the first page, so actually having "anticipated footnotes" - and it would fill a bit of the blank area, instead of possibly being pushed again on a 3rd page if the break-avoid area is really huge?
Or would this be, for a paper book publisher, a really bad practice - and footnotes should always come after the reference, so at the earliest in the page with the reference ?

----
Another thing I want to fix, but I guess there is less discussion about this:
Currently, we could get this:
<kbd>image</kbd>
a blank space at bottom, because the footnote first line + it's top-margin/separator didn't fit - while another line of the main text would have fit.
(It's quite rare - and so, each time I see that, I go chasing widows and orphans - it always took me quite a while before noticing it's the footnote that's causing the issue :/)

I guess this is fine and what we would expect:  **B** right ?
<kbd>image</kbd>

NiLuJe:
No specific opinion as (shocking!) I don't use the feature (I'm the weird guy that actually *prefers* on-demand popups ;)).

That said, and solely in the interest of avoiding weird rivers, I think I'd be fine with the current behavior for your first example, and ambivalent about the second, possibly with a vague pref for B?

poire-z:
> ambivalent about the second, possibly with a vague pref for B

ambivalent?! a vague pref ?! You, the one who implemented _reclaim stuff from bottom margin_ to have minimal margins, would be fine with having such a bottom blank ?! :) (may be you were deceived by the dual pages ?)
Do you also keep publishers widows and orphans? Or do you kill them with the styletweak to avoid such blanks?

NiLuJe:
I definitely prefer B in terms of visuals, but having the footnotes no longer in-sync feels a bit weird ^^.

hius07:
I would allow footnotes to start at the next page if needed.

Frenzie:
I have seen A before, but B is much more common afaik, and also what I prefer.

Issue #3172, opened 2017-09-04
FR: NewsDownloader feed text only


Frenzie:
* KOReader version: d4fd0b9
* Device: all

#### Issue
It doesn't always make sense to download the whole webpage.

#### Steps to reproduce
Use a feed with full text such as http://ftr.fivefilters.org/makefulltextfeed.php?url=http%3A%2F%2Fwww.nu.nl%2Frss%2FAlgemeen&max=3 or https://github.com/koreader/koreader/releases.atom

PS That first feed apparently doesn't work, see https://www.mobileread.com/forums/showthread.php?p=3576634#post3576634

PPS Pinging @mwoz123. :-)

Edit: I can't confirm any issues with http://ftr.fivefilters.org/makefulltextfeed.php?url=http%3A%2F%2Fwww.nu.nl%2Frss%2FAlgemeen&max=3 btw.

mwoz123:
Looks very interesting. 

But basic, "free" service has a [limit of max 3](http://fivefilters.org/content-only/#section-pricing) news (feed) per url (item). IMHO this is too low.

Any other services with bigger or no limit?

Frenzie:
The FR doesn't really have anything to do with that particular service. :-) (But incidentally, you can [host your own](https://bitbucket.org/fivefilters/full-text-rss) and there's also the neat [rss-bridge](https://github.com/RSS-Bridge/rss-bridge).)

mwoz123:
Quick glance and they are both seems PHP applications - I probably won't be able to help.

This would be such big change that I'd suggest to create new plugin e.g. "NewsDownloader2" (from current NewsDownloader code). 
So unit it's fully finished & working we can still have working "NewsDownloader 1" in nightly (or both).

Frenzie:
I think we have a slight misunderstanding. For a full-text feed, such as https://github.com/koreader/koreader/releases.atom (forget about fivefilters, it's just another way to get one), it doesn't make sense to download the whole webpage.

Turning a partial feed into a full feed is an entirely different can of worms.

mwoz123:
So how this enhancement should be done? What you want to use?
Would it have similar architecture:
Koreader N.D. -> sends feeds urls to a server-> server download feeds -> server does fulltext search-> server sends back data (text) -> to korader N.D.

Frenzie:
I would start by renaming `downloadFeed` to something else because that's not what it does and it seems to be fueling a lot of confusion. `downloadPage` might be a better name.

[...]

PR #5040, opened 2019-05-25
NewsDownloader: Allow including images when using text from feed

Thra11:
Split the html page download out of createEpub, so that createFromDescription
can pass in its own html for epub packaging (with optional image download).
Still pass in a url, so that relative urls can be made absolute.

Related to #3172.

### DIFF
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -276,7 +276,7 @@ function NewsDownloader:processAtom(feeds, limit, download_full_article, include
         if download_full_article then
             self:downloadFeed(feed, feed_output_dir, include_images, article_message)
         else
-            self:createFromDescription(feed, feed.context, feed_output_dir)
+            self:createFromDescription(feed, feed.context, feed_output_dir, include_images, article_message)
         end
     end
 end
@@ -296,7 +296,7 @@ function NewsDownloader:processRSS(feeds, limit, download_full_article, include_
         if download_full_article then
             self:downloadFeed(feed, feed_output_dir, include_images, article_message)
         else
-            self:createFromDescription(feed, feed.description, feed_output_dir)
+            self:createFromDescription(feed, feed.description, feed_output_dir, include_images, article_message)
         end
     end
 end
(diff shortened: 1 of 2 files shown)


plugins/newsdownloader.koplugin/main.lua
```diff
@@ -322,39 +322,46 @@ local function getTitleWithDate(feed)
...
         local article_message = T(_("%1\n%2"), message, title_with_date)
-        DownloadBackend:download(link, news_dl_path, include_images, article_message)
+        local link = getFeedLink(feed.link)
+        local html = DownloadBackend:loadPage(link)
+        DownloadBackend:createEpub(news_file_path, html, link, include_images, article_message)
     end
 end
 
-function NewsDownloader:createFromDescription(feed, context, feed_output_dir)
+function NewsDownloader:createFromDescription(feed, context, feed_output_dir, include_images, message)
```
Thra11: The bulk of `NewsDownloader:downloadFeed` and `NewsDownloader:createFromDescription` is now identical.

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -322,39 +322,46 @@ local function getTitleWithDate(feed)
...
-    local footer = _("This is just description of the feed. To download full article go to News Downloader settings and change 'download_full_article' to 'true'")
+                                             title_with_date,
+                                             file_extension)
+    local file_mode = lfs.attributes(news_file_path, "mode")
+    if file_mode == "file" then
+        logger.dbg("NewsDownloader:", news_file_path, "already exists. Skipping")
+    else
+        logger.dbg("NewsDownloader: News file will be stored to :", news_file_path)
+        local article_message = T(_("%1\n%2"), message, title_with_date)
+        local footer = _("This is just description of the feed. To download full article go to News Downloader settings and change 'download_full_article' to 'true'")
```
Frenzie:
```suggestion
        local footer = _("This is just a description of the feed. To download the full article instead, go to the News Downloader settings and change 'download_full_article' to 'true'.")
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -322,39 +322,46 @@ local function getTitleWithDate(feed)
...
-    local footer = _("This is just description of the feed. To download full article go to News Downloader settings and change 'download_full_article' to 'true'")
+                                             title_with_date,
+                                             file_extension)
+    local file_mode = lfs.attributes(news_file_path, "mode")
+    if file_mode == "file" then
+        logger.dbg("NewsDownloader:", news_file_path, "already exists. Skipping")
+    else
+        logger.dbg("NewsDownloader: News file will be stored to :", news_file_path)
+        local article_message = T(_("%1\n%2"), message, title_with_date)
+        local footer = _("This is just a description of the feed. To download the full article instead, go to the News Downloader settings and change 'download_full_article' to 'true'.")
```
Thra11: I wonder if it should mention that _just_ changing `download_full_article` to true and re-running download news will actually cause it to skip the article, as the filename won't change.

Maybe we could encode the (relevant) feed settings in the filename, so that changing settings doesn't appear to do nothing if you don't manually delete the existing articles.

Frenzie: Sure, I suppose. :-)

Issue #3073, opened 2017-08-11
Newsdownloader needs some unit tests


Frenzie:
* KOReader version: 7545d38
* Device: all (dev)

#### Issue
Things might break. ;-)

roygbyte:
Will be attended to by #9225

PR #5041, opened 2019-05-26
NewsDownloader: Fix createFromDescription for atom feeds

Thra11:
Atom feed content is found under the xml tag, 'content', not 'context'.
The content field is a actually a table, with the text/html at [1].

With this, I can finally get a decent changelog from https://github.com/koreader/koreader/releases.atom :-)


Frenzie:
Note that you can either have `content`, `link` or `summary`.

Basically this stuff would need to be abstracted better. I'd personally start by writing some tests to define the behavior and prevent regressions, cf. #3073.

PR #5047, opened 2019-05-28
Paste to shortcuts

### LINKED ISSUE #5046: FR: move current article to favourites (general/wallabag/newsdownloader/send2ebook)
* KOReader version: 2019.05
* Device: 

#### Issue

Often reading a article downloaded from internet (via wallbag/news downloader or send2ebook) I wish there were a function to move current reading article from this folder to a "favourites" folder (so I wont' lose it e.g. with "delete all readed" function). 

There could be a setting to choose one folder and mark it as 'favourite', and then a option "move current article to favourites" folder on current reading book .

It mostly usefull for above metioned plugins but could be an koreader menu option (or even separate plugin to move it if you don't agree to have it in main code).

Frenzie:
Adds a new paste action to folder shortcuts.

Fixes #5046.

Screenshot_2019-05-28_15-17-52

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -678,11 +678,13 @@ function FileManager:pasteHere(file)
                 ok_callback = function()
                     info_file()
                     self:onRefresh()
+                    self.clipboard = nil
                 end,
             })
         else
             info_file()
             self:onRefresh()
+            self.clipboard = nil
         end
     end
 end
--- a/frontend/apps/filemanager/filemanagershortcuts.lua
+++ b/frontend/apps/filemanager/filemanagershortcuts.lua
@@ -125,6 +125,14 @@ function FileManagerShortcuts:onMenuHold(item)
             buttons = {
                 {
                     {
+                        text = _("Paste file"),
+                        enabled = (self._manager.ui.file_chooser and self._manager.ui.clipboard) and true or false,
+                        callback = function()
+                            UIManager:close(folder_shortcuts_dialog)
+                            self._manager.ui:pasteHere(item.folder)
+                        end
+                    },
+                    {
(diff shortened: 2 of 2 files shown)


poire-z:
> Although perhaps that's actually intended and only cut should get rid of it to prevent errors?

No real opinion. Looks like Windows allows single copy & multiple pastes. I don't think it matters much on our devices where we can only copy books (or possibly, an image to the screensaver directory).

So, lgtm.

Frenzie:
The problem I noticed was that if you cut & paste, you can then paste again and you'll get an error message.

Do you think it should be "Paste file" instead to make it clear what we're talking about here?

poire-z:
> The problem I noticed was that if you cut & paste, you can then paste again and you'll get an error message.

Yeah, I got that. So you should indeed set it to nil when Paste after Cut. And if you can't detect Cut vs Copy at Paste time, fine to set it to nil in both cases (if you can detect that, well, no real opinion).

> Do you think it should be "Paste file" instead to make it clear what we're talking about here?

Not especially. We don't add "file" to Cut/Copy/Paste/Delete in the file manager, and we're in a files context in this Shorcuts thingy, so I don't mind just seeing "Paste". And it would make the buttons a bit larger. An other option is to just show the large button with "Paste file" here (in a 2 rows buttondialog)  only if you have something in the clipboard (instead of always showing it either enabled or disabled).

Frenzie:
> and we're in a files context in this Shorcuts thingy

The difference with the file manager is that "edit" and "delete" refer to folder shortcuts, not to performing actions with real files or folders.

poire-z:
So go with "Paste file" to kill any ambiguity. Possibly in a 2-rows buttondialog.

Frenzie:
Alright, I'll probably do that tomorrow. :-)

PR #5053, opened 2019-06-01
Add Korean keyboard (2-beolsik)

limerainne:
Dear KOReader developers,

I want to add Korean keyboard (in 2-beolsik; most popular and somewhat simple to implement) that I implemented, regarding many references on the web, to the KOReader.

Here is the screenshot of the keyboard:
KR

and here is some working example:
Typing with Korean keyboard

Because Korean characters, called `Hangul`, have to be typed with combinations of some `consonnants` and `vowels`, hence Korean keyboard requires some automata code to do those combinations. It seems that keyboards implemented in KOReader did not required automata things, I did not found handler functions required for processing events with automata code.

Therefore, I made those changes that are not matching your code quality standard:

  * I don't know adequete location, so I placed automata code in the same directory of keyboard data.
  * To fetch keyboard event in my custom event handler code, I appended some code to `virtualkeyboard.lua` that:
    * Hooked `addChars()` and `delChar()` functions, and if Korean characters are typed, then they will be handled with Korean automata
    * Other events to reset Korean automata

Please consider my request, and please modify it if its code style, etc. was not appropriate..

Thank you!

### DIFF
--- a/frontend/ui/data/keyboardlayouts/ko_KR_helper.lua
+++ b/frontend/ui/data/keyboardlayouts/ko_KR_helper.lua
@@ -1,0 +1,440 @@
+-- require('utf8')
+local BaseUtil = require("ffi/util")
+local logger = require("logger")
+
+--------
+-- # Hangul-input-method Kit for Lua/KOReader
+--------
+-- ## Input method implemented: 2-beolsik (for simplicity, can retrieve many articles for implementation)
+-- ## Classes and their features
+--  * HgSylbls (= Hangul Syllables)
+--    - Determine if a character is in Hangul consonnant, vowel, initial, medial, or final character
+--    - Combine initial, medial[, and final] character into a complete syllables
+--    - Determine if a medial (or final) character can be a double one (can combine another medial (or final) one)
+--  * HgFSM (= Hangul Finite State Machine)
+--    - Process Hangul syllabus combination if the character that user inputs are valid one to be combined
+--  * UIHandler
+--    - To communicate with the actual UI text input box
+--
+-- ## References
+-- https://ehclub.co.kr/2482
+-- :: Hangul syllables combination formula, Hangul unicode composition, FSM reference
+-- https://en.wikipedia.org/wiki/Hangul_consonant_and_vowel_tables
+--------
+
+----------------------
+-- Hangul Syllables --
+----------------------
+
+local HgSylbls = {
+    -- Hangul character ranges in Unicode
+    UNI_HG_BASE = 0xac00,
+    UNI_HG_UPPER = 0xd7af,
+
+    UNI_HG_CONSONNANT_BASE = 0x1100,
+    UNI_HG_CONSONNANT_UPPER = 0x1112,
+
+    UNI_HG_VOWEL_BASE = 0x1161,
+    UNI_HG_VOWEL_UPPER = 0x1175,
+
+    UNI_HG_COMPAT_CONSONNANT_BASE = 0x3131,
+    UNI_HG_COMPAT_CONSONNANT_UPPER = 0x314e,
+    UNI_HG_COMPAT_VOWEL_BASE = 0x314f,
+    UNI_HG_COMPAT_VOWEL_UPPER = 0x3163,
+
+    -- Initial, medial, and final characters to be combined
+    CHARS_INITIAL = {"ㄱ", "ㄲ",   "ㄴ", "ㄷ", "ㄸ",   "ㄹ", "ㅁ", "ㅂ", "ㅃ",
+                     "ㅅ", "ㅆ",   "ㅇ", "ㅈ", "ㅉ",   "ㅊ", "ㅋ", "ㅌ", "ㅍ", "ㅎ"},
+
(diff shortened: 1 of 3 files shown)


Frenzie:
I think it might be clearer to create a new directory named something like `keyboardhandler` for the actual code? I don't have time to look at it much this weekend; pinging @poire-z

frontend/ui/data/keyboardlayouts/ko_KR_HgKit.lua
```diff
@@ -0,0 +1,445 @@
...
+-- https://ehclub.co.kr/2482
+-- :: Hangul syllables combination formula, Hangul unicode composition, FSM reference
+-- https://en.wikipedia.org/wiki/Hangul_consonant_and_vowel_tables
+--------
+
+----------------------
+-- Hangul Syllables --
+----------------------
+
+HgSylbls = {
```
Frenzie: Should be local. That should also fix the bulk of the luacheck errors.

poire-z:
You type super fast! :)

No issue with your code quality, it's rather clean and readable.

> I think it might be clearer to create a new directory named something like keyboardhandler for the actual code?

Either that (but when it will contain a single file, it's a bit unwelcome), or rename `ko_KR_HgKit.lua` to something that will visually talk to other developers, like:
```
ko_KR_helper.lua
ko_KR_keyboard.lua
```

> To fetch keyboard event in my custom event handler code, I appended some code to virtualkeyboard.lua

Given that there are only 8 `VirtualKeyboard` small methods that just wrap calls to `self.inputbox` (a `InputText` object) methods, may be you can even avoid modifying them by just wrapping `self.inputbox` in an object defined in your `ko_KR_HgKit.lua`, and in `VirtualKeyboard:init()` just have:
```lua
if keyboard.wrapInputBox then
    self.inputbox = keyboard.wrapInputBox(self.inputbox)
end
```
(and may be only reset the previous state of it in that `wrapInputBox()` function when a new one is created, and let `VirtualKeyboard:onCloseWidget()` untouched?).
And have in your ko_KR files something like:
```lua
local InputBoxWrapper = {
    real_inputbox = nil
}
function InputBoxWrapper:addChar(key)
  -- do your state stuff
  self.real_inputbox.addChar(your_composed_key)
end
function InputBoxWrapper:upLine(key)
  self.real_inputbox:upLine()
  _resetStateCustomHandler()
end
```
and in the big `return {` have that wrapping function in it:
```lua
    wrapInputBox = function(inputbox)
        _resetStateCustomHandler() -- reset previous state
        return InputBoxWrapper:new({
            real_inputbox = inputbox,
        })
    end
```
That would leave VirtualKeyboard quite untouched with no notion of that state specific to the korean keyboard.
But if it's not as easy as I envisioned above, I'm fine with this PR as it is.

limerainne:
First of all, I should have run luacheck before making PR 😿

I've fixed PR commit from previous one, [details are](https://github.com/koreader/koreader/compare/adb379d968fa525d5751094f2dab836a5e9c45fd..1878cb4d5ba937317f3c20c8df94e6773ed0a0aa):

  * Renamed the helper classes file to `ko_KR_helper.lua`
    For now, as @poire-z said, creating dedicated directory (for Korean keyboard helper only) would be overwork. But in the future with more keyboards implemented which require helper classes, there might be a chance to create dedicated directory for helper classes :)

  * Fixed lint warnings/errors from `luacheck`
    As @Frenzie told, I put `local` keyword to **local** variables to fix too much error messages 😃 Also, there was a typo: **F**alse 😅

  * Minimized changes to `VirtualKeyboard`, and made a `InputBox` wrapper in Korean keyboard code
    As @Frenzie told, I placed code for wrapping InputBox only in `VirtualKeyboard`. The inputbox handler was moved to keyboard code.

    Unfortunately, I couldn't simplify or reduce target functions to hook for keyboard event. There are 12 (+4 from previous version) hooks exist :(

    If an user moves cursor in the input box (or making other events), previous states about combining single character has to be dropped (or resetting helper state), for avoiding previous character popping out. I thought that resetting state is required when these events are fired:
      * Moving cursors with arrow keys, touching screen, etc.
      * Delete single line or whole box
      * Closing input box (➡️ As @poire-z said, this event can be substituted with `VirtualKeyboard:init()` call.)

    But I couldn't find useful functions used commonly in above events, hence I had to wrap each event handlers. I think that other contributers in the future could fix these inefficiencies 😅

Please consider revised PR. Thank you!

poire-z:
Thanks!

> hence I had to wrap each event handlers. I think that other contributers in the future could fix these inefficiencies sweat_smile

Well, that looks quite ok to me (some hardcore Lua hacker could probably achieve something similar by tweaking _metatable_ stuff).

Just one thing, as it seems you are not _replacing_ the original inputbox (as I suggested) but just wrapping its interesting methods (which is better!), you could just, for clarity, dont `return inputbox` and just use in `virtualkeyboard.lua`:
```diff
     if keyboard.wrapInputBox then
-        self.inputbox = keyboard.wrapInputBox(self.inputbox)
+        keyboard.wrapInputBox(self.inputbox)
     end
```
(that way, we know we're not getting a new object, and don't have to wonder if there's stuff to do in the Korean keyboard code when we think about that inputbox)

limerainne:
Ah, thank you for approving this PR, @poire-z!

Issue #5060, opened 2019-06-05
Book status is not saved on End of Document actions


NiLuJe:
* KOReader version: 2019.05-18
* Device: All

#### Issue

Upon reaching the end of a book, choosing "File browser" as the end of document action will properly drop you back to the FM, but the book's progress won't have been saved: instead of showing "Finished" (or at the very least 100%), it's showing the previous status (i.e., whatever it was when you opened the book).

Going back to the FM from the menu does not exhibit this issue.

#### Steps to reproduce

Open a random book, taking note of the current progress.
Go To > Last page
PgNext until you hit the EoD action.

(I'm using the default setting of "popup", but I suspect *every* action, not just "back to FM" suffers from this).

Not familiar at all with [this piece of code](https://github.com/koreader/koreader/tree/master/frontend/apps/reader/modules/readerstatus.lua), so, open to any ideas you might have ;).

Frenzie:
I'd expect the problem to be in the popup, not in ReaderStatus?

Frenzie:
Incidentally, the Wallabag "go to folder" action doesn't properly close the currently open document either.

PR #5063, opened 2019-06-09
[fix, plugins] Wallabag: close document on go to folder

Frenzie:
See https://github.com/koreader/koreader/issues/5060#issuecomment-499416617


Frenzie:
It works fine, but you're right that a check for the reader context might be in order.

PR #5067, opened 2019-06-09
[fix] util.getSafeFilename() maximum extension length

### LINKED ISSUE #5049: Wallabag plugin:  download failed - name with quotation mark, hash and comma
* KOReader version: 2019.05-10-...2019-05-22
* Device: Inkbook (android)

#### Issue
(This doesn't seem to be correctly process by wallabag server but) noticed that it has problems probably with quatation mark (as it couldn't be downloaded by wallabag plugin). 
It was named on my framabag:
`Jak koduje się pilota (scyzoryk) do #peugeot 307 po lifcie? W necie są "jakieś" instrukcje, każda jest inna, a nic nie idzie. Ktoś zna sposób? #motoryzacja #pytanie`

#### Steps to reproduce
add 
https://www.wykop.pl/wpis/13795011/jak-koduje-sie-pilota-scyzoryk-do-peugeot-307-po-l/
to your wallabag 

Try to download with Koreader's wallabag plugin

Frenzie:
Mainly in case of sentences. Fixes #5049.


poire-z:
(It's super edge-casy, and there's no clean solution - but I guess that if you get a suffix longer than 10 chars, it's not a suffix, and you could just truncate the filename as a whole - instead of truncating the first part and again truncating the 2nd part that's not really a suffix, making the sentence twicely broken?)

Frenzie:
Another indicator would be spaces in the suffix. The conceptual problem with this edge case is that it's all perfectly permissible. The only limit is 255 characters, regardless if it's in the "filename" or the "extension." The fact that extensions are seldom longer than 5 characters and don't contain spaces is just convention.

But for the safe filename function your suggestion makes sense. If the extension comes out longer than 10 or 12 characters or contains spaces, just say `filename = str`.

poire-z:
Also, at least for the documents we save ourselves, which are aimed to be open with a html or epub renderer, it could make sense to force having one of these suffixes (even if wrong, crengine does not care about the suffix, it will guess the content type when opening the file) - so at least, it displays as a supported document with a supported suffix in the file browser.

Frenzie:
I think all the callers already do that? :-)

poire-z:
So, what happened in the issue this is supposed to fix? You shouldn't have had a long suffix then?

Frenzie:
The .epub is appended after processing.

poire-z:
OK, but what really happened?
A long suffix was detected, the name part truncated, the long suffix appended again, making it too long, and the appending of `.epub` kept making it too long, and it failed saving (or was discarded)?

Frenzie:
Yes, exactly. You can't write a filename that's too long, so it fails.

```
[w-id_1969015] Jak koduje się pilota (scyzoryk) do #<a href="#peugeot">peugeot<_a> 307 po lifcie? W necie są &quot;jakieś&quot; instrukcje, każda jest inna, a nic nie idzie. Ktoś zna sposób? #<a href="#motoryzacja">motoryzacja<_a> #<a href="#pytanie">pytanie<_a>.epub: File name too long
```

The part after idzie is the auto-detected suffix.

The idea in the function being you can give it a string or a filename, but that Polish site seems to be a champ at finding length-related edge cases that never come up in my personal use.

Strictly speaking I should go test-driven here. :-)

Alternatively the function could just be ignorant of extensions, although I'd like to be able to toss a Unix-safe filename at it to transform it into one suitable for a Windows filesystem.

Frenzie:
Actually it only just occurred to me that we can get a much nicer filename by simply stripping the HTML:

```
06/10/2019-04:19:39 PM DEBUG Wallabag: DOWNLOAD: filename:  /home/frans/src/kobo/koreader/test/wallabag/[w-id_1969015] Jak koduje się pilota (scyzoryk) do #peugeot 307 po lifcie? W necie są "jakieś" instrukcje, każda jest inna, a nic nie idzie. Ktoś zna sposób? #motoryzacja #pytanie.epub
```

That actually takes care of the original issue, but your remarks still made the code better. :-)

poire-z:
We also have the following, which has the additional effect of removing twicely encoded stuff (if that can happen where we get the filenames from):
https://github.com/koreader/koreader/blob/da12fc9e2a28fdc96bdb7c0b1dc140c920be047c/frontend/util.lua#L688-L716

Frenzie:
@poire-z Thanks! I thing that most sources (e.g., Wallabag, Wikipedia) should be fine, but you never know with any random newsfeeds.

Since we always get rid of bad characters it doesn't really matter for basic functionality, but it might occasionally have slightly more aesthetically pleasing results.

Issue #5070, opened 2019-06-10
Broken gesture detection on Kindle Basic 3


NiLuJe:
* KOReader version: All
* Device: Kindle Basic 3 (KT4)

#### Issue

Touch input is completely broken by the fact the the device appears to emit sliiightly different events than what we're used to (àla Kobo).

Stuff appears to be stuck in a perpetual hold state?

See [the bottom of this page](https://www.mobileread.com/forums/showthread.php?t=270835&page=68) for logs on both a broken & working device.

#### Steps to reproduce

Launch KOReader ;).
Then be glad you have USBNetwork installed so you can properly kill it ^^.

##### `crash.log` (if applicable)

c.f., https://www.mobileread.com/forums/showthread.php?t=270835&page=68 (bottom of this page and the next one).

Frenzie:
Is that a very old model?

NiLuJe:
That's a very new model, in fact ;).

Frenzie:
Hm, old would've made more sense to me. (Also, confusing naming. :-P)

NiLuJe:
[Indeed!](https://github.com/NiLuJe/KindleTool/blob/a9bef14eee8f6dbed9cab4a8d3accbd4b7de195a/KindleTool/kindle_tool.h#L354)

Frenzie:
I suggested trying this patch instead, because the Kobo code features some slight indirection:

```diff
diff --git a/frontend/device/kindle/device.lua b/frontend/device/kindle/device.lua
index 501d0735..2505f8dc 100644
--- a/frontend/device/kindle/device.lua
+++ b/frontend/device/kindle/device.lua
@@ -755,6 +755,7 @@ function KindleBasic3:init()
 
     Kindle.init(self)
 
+    self.input.snow_protocol = true
     self.input.open(self.touch_dev)
     self.input.open("fake_events")
 end
```
https://www.mobileread.com/forums/showpost.php?p=3855048&postcount=1037

PR #5071, opened 2019-06-10
[fix] Kindle Basic 3: use Kobo snow protocol

Frenzie:
Otherwise touch will be stuck in perpetual hold.

Fixes https://github.com/koreader/koreader/issues/5070


NiLuJe:
*facepalm*. Thanks !

Issue #5019, opened 2019-05-09
Feature request: Different page refresh preference in Night Mode


ghost:
* KOReader version: 201905
* Device: Kindle Oasis

#### Issue

I guess this is basically a feature request due to the work that NiLuJe did here: https://github.com/koreader/koreader-base/commit/19e38d0be9191dfdbf4ba0d225e635b98f3ec4fd

The previous behavior of refreshing the page every turn in nightmode, while annoying for some, was genuinely useful as the kindles really don’t draw well after about 3 page turns when in night mode. (LOTS of ghosting).

So it was nice to have page refreshes every chapter when not in night mode, but then have the software default to refreshing every page turn in night mode. If this could be added as a menu option that would be amazing.

Issue #1427, opened 2015-02-13
FR: Light settings profiles


gerroon:
Hi

It would be very neat if there were couple light settings profile listed just below the front light menu or so. 

For example we can have 3  profiles one for night one for day one for indoor type where the user sets what the settings should be in them.

thanks

chrox:
We already have a nice gesture to control light intensity. And I don't think several taps in the menus is any way easier than the gesture to adjust light.

PR #5079, opened 2019-06-15
WIP: Profiles plugin

Thra11:
I've been thinking about how, depending on whether it's day or night, I adjust a number of settings, but I usually end up with one of two states:

Setting | Dark | Light
--------- | ------ | -----
Frontlight Brightness | 1 | 16
Frontlight Warmth | 100 | 30
Night Mode | On | Off
Font Contrast | Higher | Low

My idea to make this more convenient is as follows.
* The user is able to configure two (possibly more) "profiles".
* With a UI similar to the gesture configuration UI, the user can choose multiple settings (with values) to add to the profile.
* Loading a profile simply sets the chosen settings to the selected values (After setting the settings, KOReader doesn't need to remember which profile was loaded, and the user can adjust the settings as usual).
* An action to toggle (or cycle) between profiles could be made available, so a gesture can toggle between profiles.

Would probably also solve some other issues, such as #1427 & #5019.

So far, this is a proof-of-concept which allows you to use gestures to "cycle through profiles" (execute one of two lua files). At the moment, the "profiles" are hand-written scripts (copy them into koreader/profiles/ to use them): It doesn't have a GUI to select desired settings and values yet.

Remarks
I thought about trying to hook into the gesture manager's actions, as a way to have a selection of relevant actions available instead of arbitrary code execution. However, actions are too limited for my use case, as they don't allow you to select custom values (e.g. there's a "toggle frontlight" action, but there's no "set frontlight intensity to _x_", nor support in the GUI for choosing _x_. Similarly, there's a "toggle night mode", but no "turn night mode _on_", etc.).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5079)

### DIFF
--- a/plugins/profiles.koplugin/main.lua
+++ b/plugins/profiles.koplugin/main.lua
@@ -1,0 +1,83 @@
+local DataStorage = require("datastorage")
+local Event = require("ui/event")
+local FFIUtil = require("ffi/util")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local _ = require("gettext")
+local logger = require("logger")
+local T = FFIUtil.template
+
+local Profiles = WidgetContainer:new{
+    name = "profiles",
+}
+
+local last_profile = 1
+local number_of_profiles = 2
+
+function Profiles:init()
+    logger.info("Profiles:init()")
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function Profiles:addToMainMenu(menu_items)
+    logger.info("Profiles:addToMainMenu")
+    menu_items.profiles = {
+        text = _("Profiles"),
+        sub_item_table = {
+            {
+                text = _("Profile 1"),
+                keep_menu_open = true,
+                callback = function()
+                    -- Profile configuration via GUI not yet implemented
+                    logger.dbg("Profile 1 menu callback")
+                end,
+            },
+            {
+                text = _("Profile 2"),
+                keep_menu_open = true,
+                callback = function()
+                    -- Profile configuration via GUI not yet implemented
+                    logger.dbg("Profile 2 menu callback")
+                end,
+            },
+            {
+                text = _("Load next profile"),
+                callback = function()
+                    self.ui:handleEvent(Event:new("LoadNextProfile"))
+                end,
+            },
+            {
+                text = _("Load previous profile"),
+                callback = function()
+                    self.ui:handleEvent(Event:new("LoadPreviousProfile"))
+                end,
+            },
+        },
+    }
+end
+
+function Profiles:loadProfile(profile_number)
+    local profile_path = T(_("%1/profiles/profile_%2.lua"), DataStorage:getFullDataDir(), profile_number)
+    logger.dbg("Executing profile file", profile_path)
+    local ok, err = xpcall(dofile, debug.traceback, profile_path)
+    if not ok then
+        logger.dbg("Problem executing profile:", err)
+    end
+end
+
+function Profiles:onLoadNextProfile()
+    last_profile = last_profile + 1
+    if last_profile > number_of_profiles then
+        last_profile = 1
+    end
+    self:loadProfile(last_profile)
+end
+
+function Profiles:onLoadPreviousProfile()
+    last_profile = last_profile - 1
+    if last_profile <= 0 then
+        last_profile = 1
+    end
+    self:loadProfile(last_profile)
+end
+
+return Profiles
--- a/plugins/profiles.koplugin/profile_1.lua
+++ b/plugins/profiles.koplugin/profile_1.lua
@@ -1,0 +1,14 @@
+local Device = require("device")
+local Screen = require("device").screen
+local UIManager = require("ui/uimanager")
+
+local night_mode = G_reader_settings:isTrue("night_mode")
+if night_mode then
+    Screen:toggleNightMode()
+    UIManager:setDirty("all", "full")
+    G_reader_settings:saveSetting("night_mode", false)
+end
+
+local powerd = Device:getPowerDevice()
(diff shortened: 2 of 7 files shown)


plugins/profiles.koplugin/profile_1.lua
```diff
@@ -0,0 +1,14 @@
+local Device = require("device")
+local Screen = require("device").screen
```
Frenzie: It doesn't make much practical difference but according to the coding style that should be:
```suggestion
local Screen = Device.screen
```

plugins/profiles.koplugin/main.lua
```diff
@@ -0,0 +1,83 @@
...
+local number_of_profiles = 2
+
+function Profiles:init()
+    logger.info("Profiles:init()")
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function Profiles:addToMainMenu(menu_items)
+    logger.info("Profiles:addToMainMenu")
+    menu_items.profiles = {
```
Frenzie: You should also add a `sorting_hint = "tools",`.

PS I haven't tried it yet or thought about where it'd be best to put it.

Frenzie:
> I thought about trying to hook into the gesture manager's actions, as a way to have a selection of relevant actions available instead of arbitrary code execution. However, actions are too limited for my use case, as they don't allow you to select custom values (e.g. there's a "toggle frontlight" action, but there's no "set frontlight intensity to x", nor support in the GUI for choosing x. Similarly, there's a "toggle night mode", but no "turn night mode on", etc.).

Doesn't mean there shouldn't be. ;-)

Thra11:
It should be simple enough to just store a table of parameters (nothing/empty table for actions which don't need a parameter) along with the string which identifies the action. So instead of storing
```"toggle_frontlight"```,
you'd store something like
```{ "set_frontlight_intensity", 50 }```
Lua seems to be quite good at handling things which might not be there: it would be nice if we could leave existing settings untouched.

In addition to the `action_strings`, we'd need a list of parameters that an action takes, with each parameter specifying its type (boolean, integer, enumeration, etc.), valid values (e.g. min & max for integers, possible choices for enumerations). Then the UI could decide what type of control to show, such as a checkbox for a boolean, an input text box for an integer, or a menu for an enumeration.

If we did implement this, then a profile would simply be a list of actions, and loading a profile would consist of iterating over the list triggering each action.

**Comment by poire-z** [comment]
> Then the UI could decide what type of control to show, such as a checkbox for a boolean, an input text box for an integer, or a menu for an enumeration

Just a thought if you feel like going with a UI:
Instead of having a specific UI with again some ways to configure stuff (that we can already configure each in their own place), you could just show a list of checkbox'able settings display name (without their value, and no way to configure them), and the user would just check all the settings he wants included in this profile, and ignore other. You would then just save the ones checked, with their current (not shown to the user) value(s) in your profile file. Like:
```
Settings to include in this new profile
   [x] nighmode
   [x] frontlight level
   [ ] font size
   [x] font contrast
...
```
(No way to edit a current one, except applying it, and just reselect what settings to include, and replace.)

Frenzie:
@Thra11 Btw, other than my minor nits I have no objection to merging this as is for now. :-)

Thra11:
when I have time, I think @poire-z's suggestion of creating profiles by saving the current setup with a filter to select which settings to include is probably the best way to do it, but I haven't had time to implement it so far. In the meantime, I've no objection to it being merged as is (with your changes). It can't hurt anyone by existing, and somebody might find it useful.

I'll leave it to you to decide :-)

Frenzie:
I'm of the opinion that it's easier for others to build upon once it's in. :-)

arooni:
One vote for including this in an upcoming release.

Frenzie:
Effectively superseded by #6106, although some of the functionality implemented here as a proof of concept isn't available in there yet.

yparitcher:
with #6364 plugins should be on par with gestures so this can probably be closed. :)

Thra11:
Haven't had a chance to try it out yet, but sounds good to me!

Thra11:
So having tried out #6106/#6364 in v2020.07.1, it looks there are still a few areas in which it can't quite replicate what this proof of concept did: 
1. Profiles (at least as implemented by this PR) are all about putting certain options in a specific state, which shouldn't depend on the state before the profile was loaded. It is not possible to set the backlight to a specific value using the currently available commands, as there is no "set backlight brightness to x", only "increase frontlight brightness by x" and "decrease frontlight brightness by x". Same problem for backlight warmth.
2. I created a couple of profiles, but I can't seem to trigger profiles via a gesture: Is that supposed to be an option?

NiLuJe:
@Thra11:

1. #6440 ;)
2. @yparitcher mentioned that this was in the pipeline once the groundwork was solid enough ;).

yparitcher:
2. is being implemented in #6441

yparitcher:
Dispatcher is not perfected yet (you can't change every setting in every way yet), but as is seen with #6440 it is very easy just to add a new state to change as long as there is a corresponding handler somewhere in koreader. So if there are any other edge setting missing please open a issue / PR to add it to Dispatcher

Thra11:
> gestures won't call a profile but will natively have all the abilities of a profile.

So is the plan that a gesture can have multiple functions selected?

yparitcher:
> > gestures won't call a profile but will natively have all the abilities of a profile.
> 
> So is the plan that a gesture can have multiple functions selected?

yes

Thra11:
Cool. Awesome work!

solarkite:
I've added a FR issue #8531_
I requested the option to select from available fonts in the profiles menu.
And saving current reading settings under new or existing profiles.

PR #5081, opened 2019-06-16
launch BQ Reader App from KOReader

ZapperDJ:
Script to exit KOReader and launch BQ Reader App

### DIFF
--- a/platform/cervantes/switch.sh
+++ b/platform/cervantes/switch.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+# This script exits KOReader and loads the default BQ Reader App
+# Copy it to /usr/bin so that it can be used by simply writing switch.sh on the terminal emulator plugin of KOReader
+#
+# It can also be invoked with the exit command on KOReader by modifying /etc/rc.local like this:
+#
+# REPLACE THIS CODE:
+# check if KOReader script exists.
+#  if [ -x /mnt/private/koreader/koreader.sh ]; then
+#     # yada! KOReader is installed and ready to run.
+#     while true; do
+#         /mnt/private/koreader/koreader.sh
+#         if [ -x /usr/bin/safemode ]; then
+#             safemode storage || sleep 1
+#         else
+#             sleep 1
+#         fi
+#     done
+#
+#
+# WITH THIS CODE:
+# check if KOReader script exists.
+#  if [ -x /mnt/private/koreader/koreader.sh ]; then
+#     # yada! KOReader is installed and ready to run.
+#     while true; do
+#         /mnt/private/koreader/koreader.sh
+#         /usr/bin/switch.sh
+#     done
+
+killall koreader.sh
+killall reader.lua        # Exit KOReader so that the touchcreen can be used by BQ Reader App
+/etc/init.d/connman start # Start connman in order to be able to use WiFi
+/usr/bin/restart.sh       # Launch BQ Reader App


platform/cervantes/switch.sh
```diff
@@ -0,0 +1,33 @@
...
+# check if KOReader script exists.
+#  if [ -x /mnt/private/koreader/koreader.sh ]; then
+#     # yada! KOReader is installed and ready to run.
+#     while true; do
+#         /mnt/private/koreader/koreader.sh
+#         /usr/bin/switch.sh
+#     done
+
+killall koreader.sh
+killall reader.lua        # Exit KOReader so that the touchcreen can be used by BQ Reader App
```
Frenzie: As an aside, I don't believe we handle SIGTERM (or any of those types of signals).

ZapperDJ: Well I'm running KOReader by default via koreader-standalone.sh as rc.local, so that way when I exit KOReader nothing happens, it just restarts KOReader again. With this script I can change to BQ Reader whenever i need it by issuing a command on the terminal (simple way to do it if the user for whatever reason doesn't want to mess with the code on koreader-standalone.sh) and if I enable it the way I suggest on rc.local, I can make it switch with the exit option (which makes it useful for something, as right now it just restarts KOReader and there is an option already there for that). 

Futhermore, I don't know how am I supposed to execute /mnt/private/koreader/koreader.sh from BQ reader, there is no way to execute scripts from BQ Reader as far as I know, and if there is, exiting BQ Reader and opening KOReader would be as simple as running stopapp.sh and the koreader.sh. Right now, if I want to switch from BQ Reader back to KOReader I just power off and then on again the device, and that loads KOReader back

I made this because right now because of a bug I don't have wifi access on KOReader, and I needed a way to switch to BQ Reader in order to use wifi and access my ebook via telnet. I started tinkering with the device and came to this solution, and as I am no developer and I use to hack my way around things I'm sure there is a better way of doing this (that is regarding to Frenzie's comment on SIGTERM)

On a side note, I noticed that the Store button on BQ's dev firmware doesn't work (I believe it is on purpose because on licenses and all that jazz) so maybe it could be modified to lauch KOReader?

pazos: > Well I'm running KOReader by default via koreader-standalone.sh as rc.local, so that way when I exit KOReader nothing happens, it just restarts KOReader again

> I don't know how am I supposed to execute /mnt/private/koreader/koreader.sh from BQ reader, there is no way to execute scripts from BQ Reader as far as I know

Both things are explained in https://www.mobileread.com/forums/showpost.php?p=3760905&postcount=2. I didn't receive any feedback about the instructions. I'm more than happy to adapt instructions to be easier to read/understand :smile: 

The reason why it is not possible to launch stuff from BQ Reader is explained on https://www.mobileread.com/forums/showpost.php?p=3760904&postcount=1

Now I know that the buggy fw only affects Cervantes3/4 and no older models. So porting fmon/kfmon should be easy since inotify is enabled inside the kernel.

> With this script I can change to BQ Reader whenever i need it by issuing a command on the terminal (simple way to do it if the user for whatever reason doesn't want to mess with the code on koreader-standalone.sh) and if I enable it the way I suggest on rc.local, I can make it switch with the exit option (which makes it useful for something, as right now it just restarts KOReader and there is an option already there for that).

IMO changes that affect the interaction between KOReader and stock software should stay in koreader.sh and koreader-standalone.sh is just a wrapper to run koreader.sh on a device without relying on other system scripts.

> I needed a way to switch to BQ Reader in order to use wifi and access my ebook via telnet. 

Not entirely related, but you can always start telnet at boot (if you're executing koreader-standalone.sh as rc.local) pressing the home button while the little dots are on screen. From shell you can start both KOReader and BQ reader without problems (because both programs lock the home button while they are running). Once you exited from the program pressing the home button again will continue normal boot and start KOReader as usual.

> On a side note, I noticed that the Store button on BQ's dev firmware doesn't work (I believe it is on purpose because on licenses and all that jazz) so maybe it could be modified to lauch KOReader?

Not without adapting and rebuilding the Qt application. Porting fmon seems easier, IDK. Pinging @avsej which did fix a bug before on bq software.

pazos:
@ZapperDJ: please check this commit: https://github.com/koreader/koreader/commit/301aa580cc2e37b4ff1eaea023acb1194304d778

in particular modifications to platform/cervantes/koreader.sh and the function MassStorage:start()

That way, you can start any app (including bq reader) without having to kill KOReader. These commands are useful because they finish KOReader with a magic number and you can check that number and react accordly.

platform/cervantes/switch.sh
ZapperDJ: > Both things are explained in https://www.mobileread.com/forums/showpost.php?p=3760905&postcount=2. I didn't receive any feedback about the instructions. I'm more than happy to adapt instructions to be easier to read/understand

I already know that I can launch scripts over telnet, I meant there is no way to launch scripts inside BQ Reader App in a standalone way, that is without depending on another device to launch them via command line

> Not entirely related, but you can always start telnet at boot (if you're executing koreader-standalone.sh as rc.local) pressing the home button while the little dots are on screen. From shell you can start both KOReader and BQ reader without problems (because both programs lock the home button while they are running). Once you exited from the program pressing the home button again will continue normal boot and start KOReader as usual.

This right now is useless since networking is broken on the Cervantes 2013 via KOReader

> Not without adapting and rebuilding the Qt application. Porting fmon seems easier, IDK. Pinging @avsej which did fix a bug before on bq software

Whatever works best or is easiest to implement, if it gets the job done, then it is enough for me

ZapperDJ:
> @ZapperDJ: please check this commit: [301aa58](https://github.com/koreader/koreader/commit/301aa580cc2e37b4ff1eaea023acb1194304d778)
> 
> in particular modifications to platform/cervantes/koreader.sh and the function MassStorage:start()
> 
> That way, you can start any app (including bq reader) without having to kill KOReader. These commands are useful because they finish KOReader with a magic number and you can check that number and react accordly.
> 
> If you want to launch KOReader at boot and have a graphical way of switching back to stock os (with the caveat of needing a reboot to run KOReader again)
> 
> I think this route is better and we can add a new menu entry somewhere on the ui instead of the terminal plugin

As I said earlier, my solution was kind of a hack and there are better ways to do it, and I think the solution you propose is the way to go, so I'm OK with it

NiLuJe:
I have no idea how the BQ app works, but if there's no database, or you don't want to bother handling the DB right now, a very quick and dirty approach to seeing if KFMon would do the job is to:

* Tweak the Makefile to build a Cervantes FBInk
* Enforce the proper paths via https://github.com/NiLuJe/kfmon/blob/df481d4783bae857777f6ea38c974f067d24ebb5/kfmon.h#L64
* Use the undocumented `skip_db_checks` setting for your watches (https://github.com/NiLuJe/kfmon/blob/df481d4783bae857777f6ea38c974f067d24ebb5/kfmon.c#L423).

(i.e., no actual code changes needed).

pazos:
@ZapperDJ 

> This right now is useless since networking is broken on the Cervantes 2013 via KOReader

Oh, no. The safemode tool relies on usb network, not wifi. It repurposes the same static ip used by developers firmware. It was meant as the last fallback before reinstalling a firmware, in case something went wrong and as a tool for further developments.

> As I said earlier, my solution was kind of a hack and there are better ways to do it, and I think the solution you propose is the way to go, so I'm OK with it

Cool! I have deleted the original application (I deboostraped a new debian on my device and ditch the QBookApp completely) so we still need testers. If you need help with this PR I can submit some changes on your branch for you to test!

@NiLuJe 

Database works like Kobo (but faster). I didn't test kfmon, but when I tested fbink I wrote a script to see if inotify tools worked with a preproccesed image.

- watches are on /mnt/public (noexec, vfat). This partition can be unmounted by the safemode tool via KOReader.
- executables are on /mnt/private (ext?, exec). This partition is not unmounted in any circumstances.

NiLuJe:
Probably not using the same schema, though, which is what I meant by "handling the db" ;).

ZapperDJ:
> Cool! I have deleted the original application (I deboostraped a new debian on my device and ditch the QBookApp completely) so we still need testers. If you need help with this PR I can submit some changes on your branch for you to test!

As I said earlier, I'm not a developer, so I don't know how to handle building, compiling and all that stuff (I have a GitHub account just to be able to report bugs and upload some random scripts I occasionally write for myself, just in case someone finds them useful)
However, I can test whatever precompiled packages you need me to...

pazos:
No need to rebuild anything. It is just a matter of modifying some text scripts (lua and shell). Are you fine with a diff? You need to feel comfortable with usbnet to try such changes.

ZapperDJ:
Okay then, if it is only modifying some scripts I guess I can handle that as well as usbnet

pazos:
Hey @ZapperDJ, here you go.

* on top of latest nightly try this diff:

```diff
diff --git a/frontend/ui/elements/common_settings_menu_table.lua b/frontend/ui/elements/common_settings_menu_table.lua
index 556bc965..ad20a154 100644
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -21,6 +21,17 @@ if Device:hasFrontlight() then
     }
 end
 
+if Device:isCervantes() then
+    local util = require("util")
+    if util.PathExists("/usr/bin/restart.sh") then
+    common_settings.start_bq = {
+        text = T(_("Start %1 reader app"), "BQ"),
+        callback = function()
+            UIManager:quit()
+            UIManager._exit_code = 87
+        end,
+end
+
 if Device:canToggleMassStorage() then
     local MassStorage = require("ui/elements/mass_storage")
 
diff --git a/frontend/ui/elements/filemanager_menu_order.lua b/frontend/ui/elements/filemanager_menu_order.lua
index 9847439d..418b02df 100644
--- a/frontend/ui/elements/filemanager_menu_order.lua
+++ b/frontend/ui/elements/filemanager_menu_order.lua
@@ -149,6 +149,7 @@ local order = {
         "poweroff", -- if Device:isKobo()
         "reboot",   -- if Device:isKobo()
         "----------------------------",
+        "start_bq",
         "exit",
     }
 }
diff --git a/frontend/ui/elements/reader_menu_order.lua b/frontend/ui/elements/reader_menu_order.lua
index 78339f95..af5a91eb 100644
--- a/frontend/ui/elements/reader_menu_order.lua
+++ b/frontend/ui/elements/reader_menu_order.lua
@@ -171,6 +171,7 @@ local order = {
         "poweroff", -- if Device:isKobo()
         "reboot",   -- if Device:isKobo()
         "----------------------------",
+        "start_bq",
         "exit",
     }
 }
diff --git a/platform/android/luajit-launcher b/platform/android/luajit-launcher
index 4c831970..e1de1566 160000
--- a/platform/android/luajit-launcher
+++ b/platform/android/luajit-launcher
@@ -1 +1 @@
-Subproject commit 4c8319700f0b5dae5685df7c561edaa627f19189
+Subproject commit e1de1566d7150c74b9358e152b4c6b79c2dbaa77-dirty
diff --git a/platform/cervantes/koreader.sh b/platform/cervantes/koreader.sh
index ebd3c56b..2ac6b437 100755
--- a/platform/cervantes/koreader.sh
+++ b/platform/cervantes/koreader.sh
@@ -73,7 +73,7 @@ fi
 # check if QBookApp was started before us, then
 # restart the application after leaving KOReader
 export STANDALONE="true"
-if pkill -0 QBookpp; then
+if pkill -0 QBookApp; then
     STANDALONE="false"
 fi
 
@@ -86,6 +86,7 @@ fi
 # any number lower than that will exit this script.
 RESTART_KOREADER=85
 ENTER_USBMS=86
+ENTER_QBOOKAPP=87
 RETURN_VALUE="${RESTART_KOREADER}"
 
 # Loop forever until KOReader requests a normal exit.
@@ -105,6 +106,21 @@ while [ "${RETURN_VALUE}" -ge "${RESTART_KOREADER}" ]; do
 
         safemode storage --force 2>/dev/null
         # waiting forever for home button events.
+
+    # check if user requested to open bq stock reader.
+    elif [ "${RETURN_VALUE}" -eq "${ENTER_QBOOKAPP}" ]; then
+
+        # iface is down and wpa killed. Start connman
+        [ -x /etc/init.d/connman ] && /etc/init.d/connman start
+
+        # (re)start the reading application in the background
+        restart.sh >/dev/null 2>&1
+        sleep 2
+
+        # loop while BQ app is running.
+        while pkill -0 QBookApp; do
+            sleep 10
+        done
     fi
 done
```

pazos:
Of course there are better and safer ways of doing this but all of them require the modification of `koreader-standalone.sh`. I wouldn't touch that file because it is already installed as rc.local and the only file that cannot be upgraded with normal OTAs and require user intervention.

ZapperDJ:
I tried to apply the diff using git apply on koreader-cervantes-v2019.08-88-gb2554ba_2019-09-01 but it complains that the patch is corrupted at line 25

Frenzie:
@ZapperDJ It's multiple separate diffs.

ZapperDJ:
I tried putting each diff in a single file, but got the same errors, here is my console output:

`[zapper@laserdragon koreader]$ git apply patch1                            
error: parche corrupto en la línea 25
[zapper@laserdragon koreader]$ git apply patch2
error: el parche falló: frontend/ui/elements/filemanager_menu_order.lua:149
error: frontend/ui/elements/filemanager_menu_order.lua: el parche no aplica
[zapper@laserdragon koreader]$ git apply patch3
error: el parche falló: frontend/ui/elements/reader_menu_order.lua:171
error: frontend/ui/elements/reader_menu_order.lua: el parche no aplica
[zapper@laserdragon koreader]$ git apply patch4
error: platform/android/luajit-launcher: No existe el fichero o el directorio
[zapper@laserdragon koreader]$ git apply patch5
error: parche corrupto en la línea 43
`

pazos:
please copy and paste. They're less than 25 lines. 

You can modify frontend/ui/elements/common_settings_menu_table.lua and koreader.sh directly on your device and run from there. The other changes are just for place the new item on the exit menu.

ZapperDJ:
OK I tried uninstalling my previous koreader version and installing the latest nightly, applied the patches and now koreader crashes when i try to open the menu. Here is the crash log

./luajit: frontend/ui/elements/common_settings_menu_table.lua:26: attempt to call field 'PathExists' (a nil value)
stack traceback:
        frontend/ui/elements/common_settings_menu_table.lua:26: in main chunk
        [C]: in function 'dofile'
        frontend/apps/filemanager/filemanagermenu.lua:198: in function 'setUpdateItemTable'
        frontend/apps/filemanager/filemanagermenu.lua:554: in function '_getTabIndexFromLocation'
        frontend/apps/filemanager/filemanagermenu.lua:573: in function 'handler'
        frontend/ui/widget/container/inputcontainer.lua:251: in function 'handleEvent'
        frontend/ui/uimanager.lua:670: in function 'sendEvent'
        frontend/ui/uimanager.lua:275: in function '__default__'
        frontend/ui/uimanager.lua:1010: in function 'handleInputEvent'
        frontend/ui/uimanager.lua:1071: in function 'handleInput'
        frontend/ui/uimanager.lua:1115: in function 'run'
        [C]: at 0x00013f05

NiLuJe:
The function name is `pathExists` (small case SNAFU ;)).

pazos:
@ZapperDJ: sorry!

I think you must now but anyways:

To recover from the "broken state":

1. power-off your device
2. power-on your device
3. press the home button while the little dots are on screen and wait for the [big usb-rj45 icon/face](https://github.com/pazos/cervantes-safemode/blob/master/images/usbnet.png)
4. connect via telnet (from memory 192.168.4.1)
5. modify the file frontend/ui/elements/common_settings_menu_table.lua:26. Replace `PathExists` with `pathExists`
6. run manually from telnet using /mnt/private/koreader/koreader.sh
7. hopefully: profit.

Once you're done and KOReader start nicelly again you can stop the current instance with `Ctrl+C` and push the home button again to stop usbnet and start KOReader normally.

ZapperDJ:
Yup, it worked!

pazos:
> Yup, it worked!

Cool. I understand that:

1. new menu entry exits KOReader and starts bq app without problems
2. wifi works fine on bq app
3. there aren't any glitches while the bq app is running
4. there aren't any glitches if you suspend/poweroff from bq app
5. after shutdown the device boots to KOReader just fine.

Please a bit of feedback since I cannot test myself without deleting my wonderful setup.

ZapperDJ:
Yes, everything you pointed out seems to be working ok

Issue #3871, opened 2018-04-13
FR: Change Background color/shade


gerroon:
Hi

It would be so nice if one could change the background color of the app, especially in the content area. It is set to pure white and sometimes not all content fits nicely to the screen so there is some white leaking through edges which gets a bit distracting. This is a bigger issue when used on Lcd devices.

I recommend having at least 3 basic shades like white, gray and black. The menus can stay as is, this is purely about the content area where the book pages are rendered over. I am not talking about background color of the books.

Frenzie:
If/when changing this it'd make no sense to limit it artificially, except in preset menu options.

Since white is hard-coded in so many places for now it's basically impossible except by redefining white. And more changes would be needed to affect the documents themselves. A stretch goal for now, not a quickie. :-)

![screenshot_2018-04-13_19-39-21](https://user-images.githubusercontent.com/202757/38749632-960b58d0-3f52-11e8-94a9-139f28b66f0b.png)

gerroon:
Thanks for considering.

poire-z:
Difficult indeed to have the whole thing consistent.

But in case you want it for epubs/crengine docs, you can set one in `data/cr3.ini`, by replacing:
`background.color.default=0x00FFFFFF` with something like background.color.default=0x00BBFFCC

On Android, I remember i had it working with this in patch.lua (may be it doesn't work anymore):
```lua
-- Set CREngine documents background color
local called_once = false
local fba = require("ffi/framebuffer_android")
fba.refreshFullImp_orig = fba.refreshFullImp
fba.refreshFullImp = function(self)
    fba.refreshFullImp_orig(self)
    if not called_once then
        called_once = true
        return
    end
    fba.refreshFullImp = fba.refreshFullImp_orig
    local CreDocument = require("document/credocument")
    CreDocument.initOrig = CreDocument.init
    CreDocument.init = function(self)
        CreDocument.initOrig(self)
        self._document:setStringProperty("background.color.default", "#eeeeee")
    end
end
```

gerroon:
@poire-z 

I mostly use it to read df files. Will the patch work for pdfs?

poire-z:
No, it won't work. That was easy for epub because crengine internally supports this feature of background color.
I know not much about MuPDF, but I doubt it does.

Frenzie:
StackOverflow result: https://stackoverflow.com/a/15007540/2470572

Alternatively (skimming the headers) maybe something with `fz_shade`?

[...]

PR #5088, opened 2019-06-21
fullscreen toggle for Jelly Bean devices

### LINKED ISSUE #4794: [Android 4.x]: Bottom of Screen being cut-off by navigation bar.
#####KOreader version
KoReader 2019.03

######Device Information
Energy Sistems Ereader Pro Plus
Model Number : Energy_eReaderProP
Android Version : 4.2.2
Kernel Version : Energy_ereaderProP_2016011214
Build No : Energy_ereaderProP_V2.0.0_2016011214
Energy Version : 3.0.0B

####### Other Information :
Running the e-Ink Test returns :
Manufacturer : Energy_Sistem
Brand : Energy_Sistem
Model : eReaderProPlus
Product : Energy_eReaderProP
Hardware : rk30board

fake update = does nothing
rk30xx normal update = screen refresh
rk30xx forced update = screen refresh

#### Issue
The lowest part of the screen is used by Android for the "Home", "Back" etc buttons. This obscures the lowest part of KoReader's display and it is impossible to see all of the bottom menu and dependant on font size, margins etc the bottom lines of text in a book are also obscured.

There is no "Full Screen" option visible in the KoReader settings menu.

Two photo's of the screen are attached - one shows the display with the Android System Menu (Back, Home, Frontlight etc) as a hidden (Black) display attempting to display the KOreader bottom menu The other a menu screen where "Full Screen " option would be expected and with the Android System Menu (visible this time) cutting off the lowest lines of the book text.

#### Illustrations
KoReader-1

KoReader2

`N/A

pazos:
Work in progress: 

Things work but the new viewport is not changed without reloading current app (I mean the reader or the file manager). 
Move the old toggleFullscreen for apis 14-16 to device/android/device.lua?

requires https://github.com/koreader/android-luajit-launcher/pull/148
fixes #4794

### DIFF
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -181,8 +181,8 @@ if Device:isAndroid() then
         }
     end
 
-    -- fullscreen toggle on devices with compatible fullscreen methods (apis 14-16)
-    if Device.firmware_rev <= 16 then
+    -- fullscreen toggle on devices with compatible fullscreen methods (apis 14-18)
+    if Device.firmware_rev < 19 then
         common_settings.fullscreen = {
             text = _("Fullscreen"),
             checked_func = function() return android.isFullscreen() end,
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -1,5 +1,6 @@
 local Generic = require("device/generic/device")
 local A, android = pcall(require, "android")  -- luacheck: ignore
+local Geom = require("ui/geometry")
 local ffi = require("ffi")
 local C = ffi.C
 local lfs = require("libs/libkoreader-lfs")
@@ -132,6 +133,11 @@ function Device:init()
         android.setWakeLock(true)
     end
 
+    -- check if we disable fullscreen support
+    if G_reader_settings:isTrue("disable_android_fullscreen") then
+        self:toggleFullscreen()
+    end
+
     Generic.init(self)
 end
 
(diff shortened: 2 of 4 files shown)


frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -166,8 +166,8 @@ if Device:isAndroid() then
         }
     end
 
-    -- fullscreen toggle on devices with compatible fullscreen methods (apis 14-16)
-    if Device.firmware_rev <= 16 then
+    -- fullscreen toggle on devices with compatible fullscreen methods (apis 14-18)
+    if Device.firmware_rev < 19 then
```
Frenzie: Maybe a little info message that a restart is required for changes to take effect?

pazos: done, but only for jelly bean apis.

pazos:
Also removed the patch.lua workaround now that is useless

pazos:
### before -> toggle -> after
Sin nombre

poire-z:
> Also removed the patch.lua workaround now that is useless

Might be useless in your context of fixing screen geometry, but it could be useful in other contexts :)
I haven't played with it much recently, but on Android, there's nothing much else to be able to play and tweak stuff, eg https://github.com/koreader/koreader/issues/3871#issuecomment-381229246.

So, I don't know. Do you really need to get rid of it so current users who had it to tweak viewport don't go mess with your latest changes?
If not, it would be best to keep that available as a possibility for quick tweaks/tests.

pazos:
> > Also removed the patch.lua workaround now that is useless
> 
> Might be useless in your context of fixing screen geometry, but it could be useful in other contexts :)
> I haven't played with it much recently, but on Android, there's nothing much else to be able to play and tweak stuff, eg [#3871 (comment)](https://github.com/koreader/koreader/issues/3871#issuecomment-381229246).
> 
> So, I don't know. Do you really need to get rid of it so current users who had it to tweak viewport don't go mess with your latest changes?
> If not, it would be best to keep that available as a possibility for quick tweaks/tests.

A half approach would be renaming patch.lua to runtime.lua, so old users who added fancy things there can rename their patch file and users who added things to workaround screen.height can forget about it.

What do you think?

poire-z:
That's fine with me :)
(patch.lua was a nice and explicite filename, runtime.lua a bit less, but better than nothing at all :)

pazos:
The worst thing that can happen if we leave patch.lua as is: bug reports from people saying "the window doesn't fill the screen". I can live with that :+1:

NiLuJe:
I don't know how feasible and/or annoying that would be, but perhaps add a, unobtrusive visual indicator somewhere if a patch.lua is in effect?

pazos:
> I don't know how feasible and/or annoying that would be, but perhaps add an unobtrusive visual indicator somewhere if a patch.lua is in effect?

I don't want to do that unless it can be done with a Text to Speech engine :smile:

zeroheure:
thanks!

szebrowski:
bump?

Frenzie:
@szebrowski This was merged over a year ago. Please ask on the forum or report a bug.

Issue #5091, opened 2019-06-25
OPDS Catalog Removal crash


AluminumTank:
- Version: 2019.06
- Device: Kobo Glo HD

When trying to remove default OPDS catalogs (specifically ManyBooks, Internet Archive, Flibusta (Russian), textos.info (Spanish)), there is a crash as soon as "Delete" is pressed.

Frenzie:
Interestingly, this only happens if you already removed a few of them previously (or perhaps also if you added something). On a clean install this can't be reproduced.

Frenzie:
PS The crash.log file is really quite useful. I could've seen the problem without having the program available then. ;-)

If there are fewer than 4 servers, this line will crash:
https://github.com/koreader/koreader/blob/a791a7df56ff47ac6fa4688ba1610576b0c6272a/frontend/ui/widget/opdsbrowser.lua#L92-L93

PR #4249, opened 2018-09-30
OPDS - change url for Internet Archive

robert00s:
* Change url for Internet Archive from `http://bookserver.archive.org/catalog/` to `https://bookserver.archive.org`
* Fix logger problem in fetchFeed

PR #5092, opened 2019-06-25
[fix] OPDS crash with fewer than 4 servers

Frenzie:
Regression introduced in <https://github.com/koreader/koreader/pull/4249>.

Fixes <https://github.com/koreader/koreader/issues/5091>.


Frenzie:
> But migrating the Internet Archive url only if it is nb 4 feels like another bug

Depends whether the intent is to migrate that one or to migrate that one only if it's fully default. (I'm not sure.) I have little interest in the matter even if it weren't 34°. :-P

poire-z:
>  I have little interest in the matter

Neither do I, so go ahead :)

PR #2940, opened 2017-06-05
CoverBrowser plugin: alt views for File Browser and History

poire-z:
This plugin offers some alternative views of file lists as used by FileManager and History, using "as you browse" extracted metadata and cover images from files.
It exports one new sub-menu (Display mode) to File Browser menu and upon selected mode, replace a few display methods of the FileChooser and FileManagerHistory classes. While being a bit hacky, it has the effect that all the other code of these is still used as is and all features available (sorting, hidden files, select and hold actions...), so very few code duplicated.
As of now, it does not need any modification to core code, and is fully self-contained in its plugin directory.
 
<kbd>1-menu-textcover </kbd>
(If you choose "Classic" mode, this plugin does nothing.)

<kbd>2-browser-coverimages</kbd>

<kbd>3-browser-listimages</kbd>

<kbd>4-history-coverimages</kbd>

What's your feeling about this, a plugin dynamically replacing core code (and putting it back when you choose Classic again) ?
(Could have been named koverbrowser, K-Over-browser :)

Long story ... long :

Got interested by this comment by @Frenzie https://github.com/koreader/koreader/issues/2906#issuecomment-302744109 and wondered how we could do that with our single threaded KOReader.
Thought about having another code starter (indexer.lua, like reader.lua), but it got complicated to re-initialize the whole reader stuff to the point of having the engines working.
Turned out the easiest and cheapest way to have indexing in the background was just to (unix-)fork the main reader process, have it do the opening of documents, get the metadata and cover image, save them, and die. And it works beautifuly (on kobo, should be fine on any linux based device).  Haven't seen any side effect of doing that. A bit of subprocess management and collecting, and it got really safe and sound.
(It's disabled for Android, as I don't know if fork works, and what are the alternatives.)

I didn't want any .sdr directory to be created for each file to hold a thumbnail of the cover.
So I choose SQLite to store all these informations (I learnt the lua-ljsqlite3 way of things by looking at @robert00s 's Stats plugin v2, thanks!).
It's fine as it's really fast, and deals with integrity and keeping concurrent accesses by main process (for reading) and indexer sub-process (for writing) in queue.
Cached book informations are just in one file (`settings/bookinfo_cache.sqlite3`), and it can be deleted as it's just a cache, so it's quite clean. It may get big, 5MB for a directory with 40 books if you choose Mosaic images, smaller for other modes, but it's your choice. (Not done yet: cleanup of removed files.)
[...]

PR #5094, opened 2019-06-25
MovableContainer: avoid refresh glitches on close

poire-z:
The MovableContainer's contained widget may not have integer position/dimen (which should probably be fixed in those that don't ensure that), and our maths could eventually make them even more shifted by fractional parts.
In the end, when closing the widget, the refresh code may miss painting/cleaning 1px and leave a border of the widget on the screen.

Eg, on a sufficiently large resolution, moving this InfoMessage to the screen right border, then back:

<kbd>image</kbd>

On tap elsewhere to close it:

<kbd>image</kbd>

So, @NiLuJe , don't remove the following as long as we haven't make sure there's no fractional x/y/w/h from any widget :)

https://github.com/koreader/koreader/blob/fca51757af6dace42d72163816ca12b52a2ee3aa/frontend/ui/uimanager.lua#L938-L944

Because even with that, when the x/w or y/h may both be shifted by fractional parts, we might be in a off-by-twos situation :)

### DIFF
--- a/frontend/ui/widget/container/movablecontainer.lua
+++ b/frontend/ui/widget/container/movablecontainer.lua
@@ -147,6 +147,9 @@ function MovableContainer:_moveBy(dx, dy, restrict_to_screen)
                 self._moved_offset_y = screen_h - self._orig_y - self.dimen.h
             end
         end
+        -- Ensure the offsets are integers, to avoid refresh area glitches
+        self._moved_offset_x = Math.round(self._moved_offset_x)
+        self._moved_offset_y = Math.round(self._moved_offset_y)
         -- if not restrict_to_screen, we don't need to check anything:
         -- we trust gestures' position and distances: if we started with our
         -- finger on widget, and moved our finger to screen border, a part


NiLuJe:
I've also seen something of the sort when making a FM long-press popup translucent: it's often missing a vertical line of pixel on one side (usually the left).

poire-z:
I think a simple round() is enough, that's just an offset, to move x or y, so no real need to round up or down.

I think the issue is that, if you get the original widget with `x=100 w=33.66` , you get bounds `x0=100 x1=133.66`, so a naive width of 33px.
If I then move x by `+5.5`, i get `x0=105.5 x1=139.16`, so a naive width of 34px.

Depending on how all the other stuff involved (FrameContainer to compute border coordinates, some drawRect/DrawLine, and the UIManager refresh region stuff) do their rounding (round, floor or ceil), they may end up with doing stuff on a width of 33 to 34px intially, but on a width of 33 to 35px after my move.
Just getting fully integer offsets ensure these "other stuff" will always get the same fractional parts to do their rounding on, which should consistently give the same widths, and avoid these little glitches, as far as MovableContainer was the cause.

But I think these non-fully-integers dimen should be chased, or stuff somewhere to ensure they are killed early, cause they are evil :)

From my CoverBrowser plugin PR #2940:
> The less fun part that took me too much time was chasing memory leaks (ImageWidget blitbuffers, and TextWidget/TextBoxWidget too have blitbuffers !), and make sure to call free() where logically needed.
And the half-pixel positionning ! I've seen some paintTo(bb, 18, 124.5) that caused some misalignment of images in their bordered framecontainer, eating the top or bottom border (the same code was fine on the emulator, fine in FileBrowser on my Kobo, but bad in History ; all that because on my Kobo, the "History" header TextWidget had height=18.5 , and so all the below positionning had a y=NNN.5. Hours of cursing containers widgets before finding that :) Comments in the code about how to fix this in TextWidget, if that's the right thing to do.)

PR #5095, opened 2019-06-25
Allow disabling automagic image inversion in Nightmode in CRe

### LINKED ISSUE #4986: Add option to disable non-inverted images in night mode
* KOReader version: v2019.04
* Device: Kobo Aura One

#### Issue
The recent work to keep most images non-inverted in night mode (see #2571) is a massive improvement for most documents. However, some books don't have _pictures_  as such, but have lots of line-art _decorations_: Illustrated chapter numbers, squiggly symbols to separate sections and so on. These now look a bit odd in night mode, as they are rendered as black symbols on a white rectangular background (not sure if the original images have white or transparent backgrounds), whereas if inverted, they render as white symbols and the background rectangle is indistinguishable from the page background (as it is without night mode). Could we maybe have style setting (per document, but with configurable default) which disables the non-inverting?

NiLuJe:
Fix #4986


NiLuJe:
I'm not particularly proud of the public name of the toggle, but I couldn't find anything better without jumping down a rabbit hole of multiple qualifiers ;p.

PR #5100, opened 2019-06-26
Some BookStatus QoL tweaks

NiLuJe:
* Make the cover thumbnail respect the cover's AR in the widget
* Add a "Mark as read/unread" button in the FM's longpress menu.

### DIFF
--- a/frontend/ui/widget/bookstatuswidget.lua
+++ b/frontend/ui/widget/bookstatuswidget.lua
@@ -57,12 +57,27 @@ local BookStatusWidget = InputContainer:new{
 
 function BookStatusWidget:init()
     if self.settings then
-        self.summary = self.settings:readSetting("summary") or {
+        -- What a blank, full summary table should look like
+        local new_summary = {
             rating = nil,
             note = nil,
             status = "",
             modified = "",
         }
+        local summary = self.settings:readSetting("summary")
+        -- Check if the summary table we get is a full one, or a minimal one from CoverMenu...
+        if summary then
+            if summary.modified then
+                -- Complete, use it as-is
+                self.summary = summary
+            else
+                -- Incomplete, fill it up
+                self.summary = new_summary
+                util.tableMerge(self.summary, summary)
+            end
+        else
+            self.summary = new_summary
+        end
     end
     self.total_pages = self.view.document:getPageCount()
     stats_book = self:getStats()
--- a/frontend/ui/widget/bookstatuswidget.lua
+++ b/frontend/ui/widget/bookstatuswidget.lua
@@ -17,6 +17,7 @@ local LeftContainer = require("ui/widget/container/leftcontainer")
 local LineWidget = require("ui/widget/linewidget")
 local OverlapGroup = require("ui/widget/overlapgroup")
 local ProgressWidget = require("ui/widget/progresswidget")
+local RenderImage = require("ui/renderimage")
 local Size = require("ui/size")
 local TextBoxWidget = require("ui/widget/textboxwidget")
 local TextWidget = require("ui/widget/textwidget")
@@ -343,10 +344,20 @@ function BookStatusWidget:genBookInfoGroup()
     }
     -- thumbnail
     if self.thumbnail then
+        -- Much like BookInfoManager, honor AR here
+        local cbb_w, cbb_h = self.thumbnail:getWidth(), self.thumbnail:getHeight()
+        local scale_factor = 1
+        if cbb_w > img_width or cbb_h > img_height then
+            scale_factor = math.min(img_width / cbb_w, img_height / cbb_h)
+            cbb_w = math.min(math.floor(cbb_w * scale_factor)+1, img_width)
+            cbb_h = math.min(math.floor(cbb_h * scale_factor)+1, img_height)
+            self.thumbnail = RenderImage:scaleBlitBuffer(self.thumbnail, cbb_w, cbb_h, true)
+        end
+
         table.insert(book_info_group, ImageWidget:new{
             image = self.thumbnail,
-            width = img_width,
-            height = img_height,
+            width = cbb_w,
+            height = cbb_h,
         })
         -- dereference thumbnail since we let imagewidget manages its lifecycle
         self.thumbnail = nil
--- a/frontend/ui/widget/bookstatuswidget.lua
+++ b/frontend/ui/widget/bookstatuswidget.lua
@@ -346,9 +346,8 @@ function BookStatusWidget:genBookInfoGroup()
     if self.thumbnail then
         -- Much like BookInfoManager, honor AR here
         local cbb_w, cbb_h = self.thumbnail:getWidth(), self.thumbnail:getHeight()
-        local scale_factor = 1
         if cbb_w > img_width or cbb_h > img_height then
-            scale_factor = math.min(img_width / cbb_w, img_height / cbb_h)
+            local scale_factor = math.min(img_width / cbb_w, img_height / cbb_h)
             cbb_w = math.min(math.floor(cbb_w * scale_factor)+1, img_width)
             cbb_h = math.min(math.floor(cbb_h * scale_factor)+1, img_height)
             self.thumbnail = RenderImage:scaleBlitBuffer(self.thumbnail, cbb_w, cbb_h, true)
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -31,6 +31,7 @@ local _ = require("gettext")
 
 local Button = InputContainer:new{
     text = nil, -- mandatory
+    text_func = nil,
     icon = nil,
     preselect = false,
     callback = nil,
@@ -48,6 +49,11 @@ local Button = InputContainer:new{
 }
 
 function Button:init()
+    -- Prefer an optional text_func over text
+    if self.text_func and type(self.text_func) == "function" then
+        self.text = self.text_func()
+    end
+
     if self.text then
         self.label_widget = TextWidget:new{
             text = self.text,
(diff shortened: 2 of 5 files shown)


NiLuJe:
A few quirks of the menu entry:

* I don't think I can add it to the history menu, because we don't have access to the cache there, so it ends up being fairly useless if it's only silently updating the DocSetting...

* It needs the DocSetting, so I hid that behind a "make sure it's been opened" once check. Ideally, I'd like to be able to mark books that have never been opened, but I have no idea how involved that might be...

NiLuJe:
Gah, that actually means the BookStatus widget has to have been opened at least once for that book >_<".

Sooo, yeah, need to handle sidecar creation/population.

poire-z:
Screenshot?
It looks like you'll add a row with a single long button between rows of 2 buttons, which feels like it won't look esthetically promising :)

Also, and I don't really mind that, but, except for the local cache in CoverBrowser that helps, this feature might/should as well be available in Classic mode (so, that code could be in filemanager.lua) - as CoverBrowser up to now only adds buttons that deal with its own business (well, at least what it has in its sqlite db cache).

I personally never set any reading status. Just promise me you won't create any sidecar dir/file if I don't use these buttons :) I like my directories tidy!

> I don't think I can add it to the history menu, because we don't have access to the cache there, so it ends up being fairly useless if it's only silently updating the DocSetting

This cache: ?
https://github.com/koreader/koreader/blob/03bbba249aa1e230f4318dbe0641f7126ba0908e/plugins/coverbrowser.koplugin/covermenu.lua#L469-L470
Only used in ListMenu as Mosaic menu doesn't need yet to display anything from the DocSetting. But it could if you want it to.

> It needs the DocSetting, so I hid that behind a "make sure it's been opened one" check. Ideally, I'd like to be able to mark books that have never been opened, but I have no idea how involved that might be, given that we'd need to create the sidecar folder/file...

Isn't it as easy as: no docsetting = not opened/not read. If the user change the status, just create an empty docsetting and set this value. I guess everything else will be populated when the user really open the book?
And may be, if he marks it as unread, and you don't see some common key, purge the sidecar dir?

poire-z:
And I guess you don't even need that cover_info_cache: just check existence/open the docsetting when the dialog is being built/hacked. Shouldn't cost much.

NiLuJe:
I initially wanted to add a button to open the BookStatus widget, so I was planning on two buttons ;p.
But it pretty much needs the doc open to function, so, meh, threw that idea out the window.

Yep, that's the cache ;). The issue with it not being accessible isn't the Mosaic menu itself, it's that when you go back the the FM, it displays the cached info, which is now stale ;).

In the end, since that specific bit of info isn't actually visible in the Mosoaic menu (or in the classic view), I don't terribly mind having this only there ;).

Yep, I wasn't sure how it all tied together, but DocSetting:open() takes care of everything. The only icky bit is I need to mirror the exact dictionary skeleton BookStatus expects, I can't just set the status key. It's something like 5 keys in total, so, not the end of the world, but slightly icky nonetheless ;p.

poire-z:
> it displays the cached info, which is now potentially stale ;).

I guess you could trash that cache entry in the callback, it might just be refetched when the view is redrawn.

> In the end, since that specific bit of info isn't actually visible in the Mosaic menu (or in the classic view), I don't terribly mind having this only available there ;).

Well, that's a bit ... awkward :)
Given that you don't really need that local cache (the text_func of that button could just check/open the docsetting - and the callback too) , you could really just have all that in filemanager.lua, and it would show in classic mode too like:
```
     Open with...    |    Convert     (or may be inverted, most useful to the right for most common right handed users)
Mark as read/reading | Book information
```
and CoverBrowser would keep adding his own below these.

NiLuJe:
But I do like that cache ;p. The less we touch crappy FAT32 stuff, the happier I am ;p.

poire-z:
OK, I get it, it's only in List mode that you see `Finished` instead of `68% of 120 pages`, so it's only there that you would need that alternate way of setting it.

So, just a bit annoyed having there that big `Mark as read` button that I won't be using :)
But also (a bit less) annoyed if we would need to shuffle buttons and make the layout different only in List mode :| But may be morphing that `Open with... Convert Book info` 1x3 row into a 2x2 as suggested above would make it still pretty, with the Book info still on the right, so not that bit of a shuffle... Dunno.

NiLuJe:
The `Open with... | Convert | Book info` triplet is from higher up the chain than CoverMenu, so I'm not terribly sure how to deal with that...

Right now it looks like this:
mark_as_read

mark_as_reading

I kinda like it close to the "View book desc" / "Book info" stuff, but I'd have no qualms moving down a bit, either before or after the final "Refresh cached book info".

NiLuJe:
In the meantime, took care of the DocSetting issue.

Not touching the sidecar until the actual button press callback ;).

The text_func callback only checks the cache, which is empty if there's no sidecar anyway. (Was a bit puzzled by the weird stacktraces Button not actually supporting text_func initally caused, but, hey, I definitely prefer having this in a callback like this, so, yay!).

NiLuJe:
Fun fact: the "Purge .sdr" button doesn't purge the cache (probably because it also doesn't have access to it).

poire-z:
> I kinda like it close to the "View book desc" / "Book info" stuff, but I'd have no qualms moving it down a bit, either before or after the final "Refresh cached book info".

Well, in between the "View full size cover" and "Ignore cover" (same for "Book info", which is another name for metadata, and "Ignore metadata") is a bit unwelcome, as these are up-down related, to me :)

> The Open with... | Convert | Book info triplet is from higher up the chain than CoverMenu, so I'm not terribly sure how to deal with that...

Just hack orig_buttons to reorganize the buttons:

```diff
                -- Replace Book information callback to use directly our bookinfo
                orig_buttons[4][3].callback = function()
                    FileManagerBookInfo:show(file, bookinfo)
                    UIManager:close(self.file_dialog)
                end

+               -- Add some new buttons to original buttons set
+               table.insert(orig_buttons, {
+                   { -- Your stuff
+                       text = _("Mark as read"),
+                   },
+               })
+               -- Move the "Book info" button near the "Mark as read" we added, for balance
+               table.insert(orig_buttons[5], table.remove(orig_buttons[4])) -- remove item from the end, insert at the end
+
                table.insert(orig_buttons, {
                    { -- Allow user to view real size cover in ImageViewer
                        text = _("View full size cover"),
```
would show it as:

<kbd>image</kbd>

(or even swap the Convert <> Mark as read if it feels better)

(So, no test for list mode, so you're showing that button even in Mosaic mode then? fine by me)

> Fun fact: the "Purge .sdr" button doesn't purge the cache (probably because it also doesn't have access to it).

You could probably hack the callback, like we did for the `Book info` button. Something like:
```lua
local orig_purge_callback = orig_buttons[1][3].callback
orig_buttons[1][3].callback = function()
  orig_purge_callback()
  -- trash that cache entry, whether the callback confirm will be OK'ed or not, as we can't
  -- know easily if it will or not
end
```

NiLuJe:
Okay, took a bit of finaggling because Lua, but here's how it now looks :).

mark_as_reordered

NiLuJe:
And took care of the cache, too.

Thanks for the pointers, @poire-z (I'd never have expected table.remove() to return the pop'ed element)!

NiLuJe:
Oh, and, yeah, this works in Mosaic mode, even if the info isn't visible. I'm okay with that ;p.

NiLuJe:
Random remark: ultimately, most if not all of the info that's in the BookStatus summary table would be a good fit for BookInfo instead.

(So we'd be able to say "show me all unread books" or something, which'd be a bit gnarly currently).

poire-z:
Can't you even not need:
```lua
-- Swap "Convert" ([4][2]) w/ "Book Info" ([4][3], the last item)
table.insert(orig_buttons[4], 2, table.remove(orig_buttons[4]))
```
if you just did:
```diff
 -- Move the "Convert" button (now [4][3], the last item) to the left of the "Mark as..." button [5][1] we've just added
-table.insert(orig_buttons[5], 1, table.remove(orig_buttons[4]))
+table.insert(orig_buttons[5], 1, table.remove(orig_buttons[4], 2))
```

I'm not super fond of that `Mark as read` below `Book information`. Book info, view cover, view description are "get/read/view metadata" related - Mark as read feels like an intruder in that family :) although I could admit it's setting some metadata, but...

Also, I feel I'll hit it often by error when wanting to hit Book info or View description (I might as well hit it by error if it were above Book info thus :|
So, if I hit it by error on a neveropened book, I'll see immediately the little dogear corner (because you created the docsetting), and I can just hit Purge .sdr, right? Or do/can you notice the docsetting is just a dummy one, and do the purging of that dummy docsetting if I hit that button again? Or it's just toggling between read and reading - so no room for unread, and that's the role of Purge .sdr?

> Random remark: ultimately, most if not all of the info that's in the BookStatus summary table would be a good fit for BookInfo instead.

I somehow like/wanted that the bookinfo.sdb is just a cache and can be trashed, and that we don't need to sync things in multiple places :)
Although that local cover_info_cache stuff could be also saved in the sqlitedb - but you would need tp check sidecard dir or file to see if that cached data needs a refresh... might be tricky and complex and ugly :)

Frenzie:
> (I'd never have expected table.remove() to return the pop'ed element)!

I don't think I knew that either, at least not actively. It's always possible I encountered it a few years back and forgot. :O

NiLuJe:
@poire-z : Yep, that works, thanks ;).

Yeah, I kinda see it as a metadata-related action, so the layout does make sense, IMHO.
The only nitpick would be the "view cover" on the left, but it's right above "ignore cover", which is right next to another "ignore", so, it all ties back nicely ;).

Yep, purging would be the way. nil is assumed to mean "reading" when *reading* the info, but writing will always write something relevant to that (i.e., "reading"/"read").

Issue #5098, opened 2019-06-26
Sometimes KOReader doesn't wake up after opening sleepcover


allati:
**KOReader version**: v2019.06
**Device**: Kobo Forma
Sleepcover from Kobo. Not sure if it makes the difference, but I counted 6 magnets in it (https://imgur.com/a/G092yBM). I guess that two magnets on the right keeps cover closed or in it's "origami" position. One magnet in the triangle triggers open/close event. And there are three more magnets doing have no idea what, maybe helping the other two to keep cover closed.

**Steps to reproduce**
Sometimes when I open sleepcover, KOReader is not waking up. Closing and then opening sleepcover again usually wakes it.

In similar-looking issue (#5087) NiLuJe suggested that it may be caused by sleepcover magnets triggering sensors when folded to the back of the device. In my case it is unlikely, because I keep it open at 270° max, it shouldn't be hitting sensors at these angles.

Here are excerpts from crash.log for two recent similar cases:

```
====== Case 1

06/20/19-00:58:26 INFO  Kobo suspend: synced FS
06/20/19-00:58:26 INFO  Kobo suspend: asking for a suspend to RAM . . .
06/20/19-00:58:26 INFO  Kobo suspend: ZzZ ZzZ ZzZ? Write syscall returned:  true

//sleepcover opened, but KOReader keeps sleeping
06/20/19-11:03:14 INFO  Kobo suspend: woke up!
06/20/19-11:03:29 ERROR Kobo suspend: putting device back to sleep, unexpected wakeups: 1
06/20/19-11:03:29 INFO  Kobo suspend: going to sleep . . .
06/20/19-11:03:29 INFO  Kobo suspend: asked the kernel to put subsystems to sleep, ret: true
06/20/19-11:03:31 INFO  Kobo suspend: waited for 2s because of reasons...
06/20/19-11:03:31 INFO  Kobo suspend: synced FS
06/20/19-11:03:31 INFO  Kobo suspend: asking for a suspend to RAM . . .
06/20/19-11:03:31 INFO  Kobo suspend: ZzZ ZzZ ZzZ? Write syscall returned:  true

//sleepcover reopened, KOReader wakes up
06/20/19-11:04:04 INFO  Kobo suspend: woke up!
06/20/19-11:04:11 INFO  Kobo resume: clean up after wakeup

====== Case 2

06/20/19-20:39:55 INFO  Kobo suspend: asking for a suspend to RAM . . .
[...]

Issue #5087, opened 2019-06-19
koreader freezing after waking up device


lbesnard:
* KOReader version: v2019.06
* Device: Kobo Glo

#### Issue
The e-reader is freezing after waking up device directly back on koreader.

I looked at this [issue](https://github.com/koreader/koreader/issues/3936) as it seemed quite similar and was fixed in a recent PR. My wifi was turned off.

#### Steps to reproduce
1. Turn on Kobo Glo
2. Open Koreader
3. Open a book
4. Close sleep mode cover (or just put ebook reader to sleep mode)
5. Wake up
after a few seconds, the entire device freezes. Impossible to do anything. Redo

##### `crash.log` (if applicable)
```
                launching...
  _  _____  ____                _
 | |/ / _ \|  _ \ ___  __ _  __| | ___ _ __
 | ' / | | | |_) / _ \/ _` |/ _` |/ _ \ '__|
 | . \ |_| |  _ <  __/ (_| | (_| |  __/ |
 |_|\_\___/|_| \_\___|\__,_|\__,_|\___|_|

 It's a scroll... It's a codex... It's KOReader!

 [*] Current time: 06/19/19-16:29:13
 [*] Version: v2019.06

ffi.load: SDL2
ffi.load (warning): libSDL2.so: cannot open shared object file: No such file or directory
ffi.load: libSDL2-2.0.so
ffi.load (warning): libSDL2-2.0.so: cannot open shared object file: No such file or directory
SDL2 not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: libSDL2-2.0.so
ffi.load: SDL
ffi.load (warning): libSDL.so: cannot open shared object file: No such file or directory
SDL not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: SDL
ffi.load: blitbuffer
FB: computed size: 1572864
FB: id           : mxc_epdc_fb
FB: smem_len     : 3145728
FB: line_length  : 768
FB: xres         : 758
FB: xres_virtual : 768
FB: yres         : 1024
FB: yres_virtual : 2048
FB: bpp          : 8
06/19/19-16:29:13 INFO  initializing for device Kobo_kraken
06/19/19-16:29:13 INFO  framebuffer resolution: {
    ["h"] = 1024,
    ["w"] = 758
}
[ko-input] Forked off fake event generator(pid:4633).
ffi.load: libs/libmupdf.so
ffi.load: libs/libwrap-mupdf.so
[...]

PR #5102, opened 2019-06-28
Make sure we only pass a single IP to ping

NiLuJe:
For some mysterious reason, I have a duplicate default gateway on my
Forma, one with no metrics, and a proper one.
This ensures we pick the proper one.

I'm fairly sure this is temporary insanity on the part of my device,
but, still, it can't hurt.

`route -n`:
```
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.0.1     0.0.0.0         UG    0      0        0 eth0
0.0.0.0         192.168.0.1     0.0.0.0         UG    208    0        0 eth0
192.168.0.0     0.0.0.0         255.255.255.0   U     208    0        0 eth0
```

`ip r`:
```
default via 192.168.0.1 dev eth0
default via 192.168.0.1 dev eth0  metric 208
192.168.0.0/24 dev eth0  src 192.168.0.49  metric 208
```

### DIFF
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -314,9 +314,9 @@ function Device:retrieveNetworkInfo()
             -- NOTE: No -w flag available in the old busybox build used on Legacy Kindles...
             local pingok
             if self:isKindle() and self:hasKeyboard() then
-                pingok = os.execute("ping -q -c 2 `ip r | grep default | cut -d ' ' -f 3` > /dev/null")
+                pingok = os.execute("ping -q -c 2 `ip r | grep default | tail -n 1 | cut -d ' ' -f 3` > /dev/null")
             else
-                pingok = os.execute("ping -q -w 3 -c 2 `ip r | grep default | cut -d ' ' -f 3` > /dev/null")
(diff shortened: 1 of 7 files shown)


NiLuJe:
Bonus points for tail vs. head:

On a random sampling of one single test (:B), usb0 ends up at the top of the list if I enable USBNet.

```
192.168.2.0/24 dev usb0  src 192.168.2.2 
192.168.0.0/24 dev eth0  src 192.168.0.50 
default via 192.168.0.1 dev eth0
```

```
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
192.168.2.0     0.0.0.0         255.255.255.0   U     0      0        0 usb0
192.168.0.0     0.0.0.0         255.255.255.0   U     0      0        0 eth0
0.0.0.0         192.168.0.1     0.0.0.0         UG    0      0        0 eth0
```

NiLuJe:
Added some SleepCover tweaks (#5098, #5087) and a potential regression fix after #5100, so this ought to be rebased, not squashed ;).

NiLuJe:
Yup, everything went back to normal after a reboot.

PR #5107, opened 2019-07-03
Delete book after reading

### LINKED ISSUE #5050: FR: delete book after reading - new "end of docuemnt action" popup action
#### Issue
Add option to "delete" just read book to "End of document action".
Most usefull for Wallabag, news downloader etc plugins, but can be also used with normal books.

#### Steps to reproduce
Enable "end of docuemnt action" -> "Ask with pop-up dialog".

Go to last page of book (read 100%) of the book, click next page, menu will appear add it there.

robert00s:
Close #5050 
Add option to "delete" just read book to "End of document action".
Before:

After:

### DIFF
--- a/frontend/apps/reader/modules/readerstatus.lua
+++ b/frontend/apps/reader/modules/readerstatus.lua
@@ -70,17 +70,17 @@ function ReaderStatus:onEndOfBook()
             },
             {
                 {
-                    text = _("Open next file"),
-                    enabled = collate,
+                    text = _("Delete file"),
                     callback = function()
-                        self:openNextFile(self.document.file)
+                        self:deleteFile(self.document.file, false)
                         UIManager:close(choose_action)
                     end,
                 },
                 {
-                    text = _("Delete file"),
+                    text = _("Open next file"),
+                    enabled = collate,
                     callback = function()
-                        self:deleteFile(self.document.file, false)
+                        self:openNextFile(self.document.file)
                         UIManager:close(choose_action)
                     end,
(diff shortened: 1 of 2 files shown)


poire-z:
Long time no PR, welcome back :)

May be switch these 2:   `| Delete file | Open next file |` ?
I see there is a ConfirmBox so no risk if we hit `Delete file` by error, but may be `Open next file` should be reachable more easily - instead of the destructive option?

Well, personally, I don't use any End of doc action - but if I were, I'd probbly use `Delete file` the most :) - So, I dunno, may be people will do that most often too... and so it's at the right place.

In the menu, may be move `Delete file` a bit up, so they are in the same order we see them in the button dialog - just for consistency?

AlanSP1:
Certainly very handy option to have. Thanks...

NiLuJe:
One random question: Does it wipe the sidecar? Should it?

robert00s:
@poire-z 
I was busy a bit :)
> May be switch these 2: | Delete file | Open next file | ?

Ok.

> In the menu, may be move Delete file a bit up, so they are in the same order we see them in the button dialog - just for consistency?


Good point.

@NiLuJe 

> One random question: Does it wipe the sidecar? Should it?

Yes. Delete file option works exactly like delete in file browser. Delete file + .sdr + remove from history.

Eduardomb22:
Thanks for this update.  I have a question, is it possible to add a option to exit koreader after the file is deleted instead of go to the filebrowser?

PR #5108, opened 2019-07-04
android: add support for 3rd party dictionary apps

### LINKED ISSUE #3496: Koreader Dictionary lookup is slow on android
hi ,
i own both a kindle voyage and a samsung galaxy tab s
i have noticed that looking up in dictionary much more slow on my android tablet than on my kindle .

(I love Koreader , and thank you for the hard work )

### LINKED ISSUE #5039: [feat. request] autocopy to clipboard
* KOReader version: 2019.05
* Device: Android

#### Issue:
Dictionary look-ups are very slow on my devide and worse, many times fail to find even simple words (eventhough I have 5 dictionaries all of them including it).
Autocopy word on selection could potentially solve many my dictionary look-up problems (depends on whether tasker clipboard scanner is battery efficient, which I have not yet tested). Better would be that selected words are automatically sent to dictionary apps (and a way to disable it easily so that one can highlight text).
Another improvement IMO would be that instead of showing "searching dictionary for:" (which takes ages on my device) koreader could directly display the pop-up window but with wikipedia, highlight and search buttons on top adding a "copy" button, so that the dict search results can appear below (after a few seconds).

pazos:
Added support for Aard2, Alpus, Colordict, Fora, Goldendict, Kiwix, Mdict and Quickdic.

Goldendict Pro is untested. The Free version works fine.

All credits to @poire-z, who did most of the job in https://github.com/koreader/koreader/issues/2333#issuecomment-502010311

**how to add support for new dictionary apps ?**

1. find a target dict app (already installed, configured and ready to work)
2. find its package name (using an app like "Package Names Viewer")
3. try generic actions in this order: picker-share -> picker-send -> picker-text.
4. hopefully the application is available as an option to open the query.
5. try again removing the "picker-", It should open your dictionary without questions.

You might need to look at the app AndroidManifest.xml and find the kind of intents that the application handles.

Requires https://github.com/koreader/android-luajit-launcher/pull/156
Fixes #3496
Fixes #5039

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -298,6 +298,56 @@ If you'd like to change the order in which dictionaries are queried (and their r
             }
         }
     }
+    if Device:canExternalDictLookup() then
+        local function genExternalDictItems()
+            local items_table = {}
+            for i, v in ipairs(Device:getExternalDictLookupList()) do
+                local setting = v[1]
+                local dict_name = v[2]
+                local is_enabled = v[3]
+                table.insert(items_table, {
+                    text = dict_name,
+                    checked_func = function()
+                        return setting == G_reader_settings:readSetting("external_dict_lookup_method")
+                    end,
+                    enabled_func = function()
+                        return is_enabled == true
+                    end,
+                    callback = function()
+                        G_reader_settings:saveSetting("external_dict_lookup_method", v[1])
+                    end,
+                })
+            end
+            return items_table
+        end
+        table.insert(menu_items.dictionary_settings.sub_item_table, 1, {
+            text = _("Use external dictionary"),
+            checked_func = function()
+                return G_reader_settings:isTrue("external_dict_lookup")
+            end,
+            callback = function()
+                G_reader_settings:flipNilOrFalse("external_dict_lookup")
+            end,
+        })
+        table.insert(menu_items.dictionary_settings.sub_item_table, 2, {
+            text_func = function()
+                local display_name = _("none")
+                local ext_id = G_reader_settings:readSetting("external_dict_lookup_method")
+                for i, v in ipairs(Device:getExternalDictLookupList()) do
+                    if v[1] == ext_id then
+                        display_name = v[2]
+                        break
+                    end
+                end
+                return T(_("Dictionary: %1"), display_name)
+            end,
+            enabled_func = function()
+                return G_reader_settings:isTrue("external_dict_lookup")
+            end,
+            sub_item_table = genExternalDictItems(),
+            separator = true,
+        })
+    end
 end
 
 function ReaderDictionary:onLookupWord(word, box, highlight, link)
@@ -708,6 +758,18 @@ function ReaderDictionary:stardictLookup(word, dict_names, fuzzy_search, box, li
             time = os.time(),
             word = word,
         })
+    end
+
+    if Device:canExternalDictLookup() and G_reader_settings:isTrue("external_dict_lookup") then
+        Device:doExternalDictLookup(word, G_reader_settings:readSetting("external_dict_lookup_method"), function()
+            if self.highlight then
+                local clear_id = self.highlight:getClearId()
+                UIManager:scheduleIn(0.5, function()
+                    self.highlight:clear(clear_id)
+                end)
+            end
+        end)
+        return
     end
 
     if fuzzy_search then
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -40,6 +40,23 @@ local function getCodename()
     return codename
 end
 
+local EXTERNAL_DICTS_AVAILABILITY_CHECKED = false
+local EXTERNAL_DICTS = require("device/android/dictionaries")
+local external_dict_when_back_callback = nil
+
+local function getExternalDicts()
+    if not EXTERNAL_DICTS_AVAILABILITY_CHECKED then
+        EXTERNAL_DICTS_AVAILABILITY_CHECKED = true
+        for i, v in ipairs(EXTERNAL_DICTS) do
+            local package = v[4]
+            if android.isPackageEnabled(package) then
+                v[3] = true
+            end
+        end
+    end
+    return EXTERNAL_DICTS
+end
+
 local Device = Generic:new{
     isAndroid = yes,
     model = android.prop.product,
@@ -57,6 +74,22 @@ local Device = Generic:new{
         if not link or type(link) ~= "string" then return end
         return android.openLink(link) == 0
     end,
+    canExternalDictLookup = yes,
+    getExternalDictLookupList = getExternalDicts,
+    doExternalDictLookup = function (self, text, method, callback)
+        external_dict_when_back_callback = callback
+        local package, action = nil
+        for i, v in ipairs(getExternalDicts()) do
+            if v[1] == method then
+                package = v[4]
+                action = v[5]
+                break
+            end
+        end
+        android.dictLookup(text, package, action)
+    end,
+
+
     --[[
     Disable jit on some modules on android to make koreader on Android more stable.
 
@@ -86,6 +119,12 @@ function Device:init()
                 or ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
                 this.device.screen:_updateWindow()
             elseif ev.code == C.APP_CMD_RESUME then
+                EXTERNAL_DICTS_AVAILABILITY_CHECKED = false
+                if external_dict_when_back_callback then
+                    external_dict_when_back_callback()
+                    external_dict_when_back_callback = nil
+                end
+
                 local new_file = android.getIntent()
                 if new_file ~= nil and lfs.attributes(new_file, "mode") == "file" then
                     -- we cannot blit to a window here since we have no focus yet.
--- a/frontend/device/android/dictionaries.lua
+++ b/frontend/device/android/dictionaries.lua
@@ -1,0 +1,15 @@
+return { --[[ supported android dictionary applications.
+
+Most of them should support Intent.ACTION_SEND, Intent.ACTION_SEARCH or
+Intent.ACTION_PROCESS_TEXT. Some applications implement their custom intents. ]]--
+
+    { "Aard2", "Aard2", false, "itkach.aard2", "aard2" },
+    { "Alpus", "Alpus", false, "com.ngcomputing.fora.android", "search" },
+    { "ColorDict", "ColorDict", false, "com.socialnmobile.colordict", "colordict" },
+    { "Fora", "Fora Dict", false, "com.ngc.fora", "search" },
+    { "GoldenFree", "GoldenDict Free", false, "mobi.goldendict.android.free", "send" },
+    { "GoldenPro", "GoldenDict Pro", false, "mobi.goldendict.android.pro", "send" },
+    { "Kiwix", "Kiwix", false, "org.kiwix.kiwixmobile", "text" },
+    { "Mdict", "Mdict", false, "cn.mdict", "send" },
+    { "QuickDic", "QuickDic", false, "de.reimardoeffinger.quickdic", "quickdic" },
+}
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -71,6 +71,7 @@ local Device = {
 
     canOpenLink = no,
     openLink = no,
+    canExternalDictLookup = no,
 }
 
 function Device:new(o)


frontend/device/android/device.lua
```diff
@@ -86,6 +111,7 @@ function Device:init()
                 or ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
                 this.device.screen:_updateWindow()
             elseif ev.code == C.APP_CMD_RESUME then
+                this.device.getExternalDictLookupList = getExternalDicts
```
poire-z: What does that do?
It looks like you just re-assigning that function to the that name, and I see no reason for getExternalDictLookupList to have been assigned something else in between (?).

frontend/device/android/dictionaries.lua
```diff
@@ -0,0 +1,56 @@
...
+
+Most of them should support Intent.ACTION_SEND or Intent.ACTION_SEARCH.
+
+Some applications implement their custom intents. Colordict, for example,
+uses colordict.intent.action.SEARCH. ]]--
+
+    { "Aard2", "Aard2", false,
+        {
+            package="itkach.aard2",
+            action="aard2",
```
poire-z: If there is never anything else than these 2 "package" and "action", you could as well have them in the main list, and access them with v[4] and v[5].

```
{ "Aard2", "Aard2", false, "itkach.aard2", "aard2"},
{ "Alpus", "Alpus", false, "com.ngcomputing.fora.android", "search"}
{ "ColorDict", "ColorDict", false, "com.socialnmobile.colordict", "colordict"}
[...]
```
would look nice enough.

frontend/device/android/device.lua
pazos: it updstes the list of availab

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -298,6 +298,53 @@ If you'd like to change the order in which dictionaries are queried (and their r
             }
         }
     }
+    if Device:canExternalDictLookup() then
+        local function genExternalDictItems()
+            local items_table = {}
+            for i, v in ipairs(Device:getExternalDictLookupList()) do
+                table.insert(items_table, {
```
Frenzie: Some people object to what they call useless variables, but I'm of the opinion that something like this is easier to comprehend at a glance and therefore preferable:

```lua
local setting = v[1]
local text = v[2]
local enabled = v[3]
```

frontend/device/android/device.lua
poire-z: But you're just re-assigning that function to that name, not actually calling it. So that's a no-op.
And you're actually calling it in `doExternalDictLookup `, so doing the `android.isPackageEnabled` at each lookup (as you got rid of my `EXTERNAL_DICTS_AVAILABILITY_CHECKED` flag).
If you really care about rechecking availability, you should keep that kind of flag (to avoid rechecking on each lookup in the same foreground session), and just unset it onResume, so the next lookup (only) do the availability check.
(Or do I miss something?)

pazos:
@poire-z: about your comment in https://github.com/koreader/koreader/issues/2333#issuecomment-508459641

> > We need to refresh the table when the activity is resumed (because an user can install/uninstall software in between).
> 
> Is the packages-installed-checks expensive?

Not really expensive but could be cheaper. The overhead is in the JNI interface for android.isPackageAvailable and we do this for each dict. It takes a few milliseconds anyways.

> Is that really needed to do it each time KOReader is put to foreground? Once some dict are installed, we may never install any new ones. And if the user installs one and doesn't see it in the menu, he will naturally think that it needs a real quit/restart.

:+1: I will remove the onResume code.

What is left to do is clean the hightlight before or after we start the new activity.

untitled

poire-z:
> What is left to do is clean the hightlight before or after we start the new activity.

You could just call `self.highlight:clear()` before calling `Device:doExternalDictLookup()`.

But if launching an external dict lookup always bring another window on the foregroud, and quitting it brings KOReader in the foreground, **and** if that triggers one of your `elseif ev.code == C.APP_CMD_GAINED_FOCUS...` or `elseif ev.code == C.APP_CMD_RESUME then`, you could pass a callback to `Device:doExternalDictLookup()` which would get called when one of these event happen, and it could do a bit of that:
https://github.com/koreader/koreader/blob/a3356483bac01d8e7becd8ca37d83a8df6b14603/frontend/ui/widget/dictquicklookup.lua#L740-L747

Something like
```lua
-- local at the module level
local external_dict_when_back_callback = nil
[...]
    doExternalDictLookup = function (self, text, method, when_back_callback)
        external_dict_when_back_callback = when_back_callback
        local package, action = nil
        [...]

            elseif ev.code == C.APP_CMD_RESUME then
                EXTERNAL_DICTS_AVAILABILITY_CHECKED = false
                if external_dict_when_back_callback then
                    external_dict_when_back_callback()
                    external_dict_when_back_callback = nil
                end

    if Device:canExternalDictLookup() and G_reader_settings:isTrue("external_dict_lookup") then
        Device:doExternalDictLookup(word, G_reader_settings:readSetting("external_dict_lookup_method"), function()
            if self.highlight then
                -- delay unhighlight of selection, so we can see where we stopped when
                -- back from our journey into dictionary or wikipedia
                local clear_id = self.highlight:getClearId()
                UIManager:scheduleIn(0.5, function()
                    self.highlight:clear(clear_id)
                end)
            end
        end)
        return
    end
```

pazos:
thank you very much! It works nice for normal lookups inside documents. I can still reproduce the same issue on the Dictionary lookup history widget.

BTW, logcat reports

```logcat
07-05 01:28:16.772  3300  3314 D DICT    : Package itkach.aard2 is installed. Enabled? -> true
07-05 01:28:16.773  3300  3314 D DICT    : Package com.ngcomputing.fora.android is not installed.
07-05 01:28:16.774  3300  3314 D DICT    : Package com.socialnmobile.colordict is installed. Enabled? -> true
07-05 01:28:16.774  3300  3314 D DICT    : Package com.ngc.fora is not installed.
07-05 01:28:16.775  3300  3314 D DICT    : Package mobi.goldendict.android.free is not installed.
07-05 01:28:16.775  3300  3314 D DICT    : Package mobi.goldendict.android.pro is not installed.
07-05 01:28:16.776  3300  3314 D DICT    : Package org.kiwix.kiwixmobile is installed. Enabled? -> true
07-05 01:28:16.776  3300  3314 D DICT    : Package cn.mdict is not installed.
07-05 01:28:16.777  3300  3314 D DICT    : Package de.reimardoeffinger.quickdic is installed. Enabled? -> true
```

So each jni call <= 1ms.

poire-z:
> I can still reproduce the same issue on the Dictionary lookup history widget.

What issue? The thing I can gather from your animated screenshots (you didn't say it was an issue, so I didn't look for one before :) : the reverted black bar over the item you tap that doesn't get unreverted?

That may be some refresh/redraw issue: the unrevert is `UIManager:scheduleIn`'ed, and KOReader lose focus before the delay happen and may be luajit is suspended, or some push of the buffer to android layer is lost or something. I guess you don't get that if you uncheck `Screen > EInk settings > Flash buttons and menu items` ? (Pinging @NiLuJe who love these flashes :)

About the book selected text highlight and its clearing, also look how it behave if you do the Dict lookup from the multiple-words-selected dialog (my above suggestion may clear the highlight early - it should stay highlighted till that button dialog is closed - but may be just check it does not crash :)

pazos:
> About the book selected text highlight and its clearing, also look how it behave if you do the Dict lookup from the multiple-words-selected dialog (my above suggestion may clear the highlight early - it should stay highlighted till that button dialog is closed - but may be just check it does not crash :)

It behaves exactly as you expected :+1: 

> That may be some refresh/redraw issue: the unrevert is UIManager:scheduleIn'ed, and KOReader lose focus before the delay happen and may be luajit is suspended, or some push of the buffer to android layer is lost or something. I guess you don't get that if you uncheck Screen > EInk settings > Flash buttons and menu items ?

That's correct. Cannot reproduce disabling "Flash buttons and menu items"

Here are the logs

``` adb logcat
07-05 17:01:15.505  3407  3420 V KOReader:  dict lookup word: location
07-05 17:01:15.505  3407  3420 V KOReader:  dict stripped word: location
07-05 17:01:15.517  3407  3420 D KOReader: Package itkach.aard2 is installed. Enabled? -> true
07-05 17:01:15.523  3407  3420 D KOReader: Package com.ngcomputing.fora.android is not installed.
07-05 17:01:15.528  3407  3420 D KOReader: Package com.socialnmobile.colordict is installed. Enabled? -> true
07-05 17:01:15.528  3407  3420 D KOReader: Package com.ngc.fora is not installed.
07-05 17:01:15.529  3407  3420 D KOReader: Package mobi.goldendict.android.free is not installed.
07-05 17:01:15.530  3407  3420 D KOReader: Package mobi.goldendict.android.pro is not installed.
07-05 17:01:15.534  3407  3420 D KOReader: Package org.kiwix.kiwixmobile is installed. Enabled? -> true
07-05 17:01:15.535  3407  3420 D KOReader: Package cn.mdict is not installed.
07-05 17:01:15.538  3407  3420 D KOReader: Package de.reimardoeffinger.quickdic is installed. Enabled? -> true
07-05 17:01:15.545  3407  3407 V threaded_app: WindowFocusChanged: 0xa662dfe0 -- 0
07-05 17:01:15.545  3407  3407 D KOReader: App paused
07-05 17:01:15.546  3407  3407 V threaded_app: Pause: 0xa662dfe0
07-05 17:01:15.550  3407  3420 V KOReader:  Explicit widgetRepaint: table: 0x91a86710 @ ( 18 , 1502.5 )
07-05 17:01:15.555  3407  3420 V KOReader:  setDirty via a func from widget nil
07-05 17:01:15.555  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:15.555  3407  3420 V KOReader:  _refresh: Enqueued ui update for region 18 1502.5 1044 72 
07-05 17:01:15.598  3407  3420 V KOReader:  input event => type: 4, code: 7(0), value: 1, time: 1562338875.598348
07-05 17:01:15.598  3407  3420 V KOReader:  Android application event 7
07-05 17:01:15.598  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:15.598  3407  3420 V threaded_app: activityState=13
07-05 17:01:15.599  3407  3420 V KOReader:  input event => type: 4, code: 13(6), value: 1, time: 1562338875.598819
07-05 17:01:15.599  3407  3420 V KOReader:  Android application event 13
07-05 17:01:15.599  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:15.882  3407  3420 V KOReader:  BackgroundRunner: _execute() @  1562338875
07-05 17:01:15.882  3407  3420 V KOReader:  BackgroundRunnerWidget: start running @  1562338875
07-05 17:01:15.882  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:15.883  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:17.643  3407  3407 V threaded_app: WindowFocusChanged: 0xa662dfe0 -- 1
07-05 17:01:17.643  3407  3420 V KOReader:  input event => type: 4, code: 6(nil), value: 1, time: 1562338877.643446
07-05 17:01:17.643  3407  3420 V KOReader:  Android application event 6
07-05 17:01:17.726  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:17.875  3407  3407 D KOReader: App resumed
07-05 17:01:17.884  3407  3420 V KOReader:  BackgroundRunner: _execute() @  1562338877
07-05 17:01:17.884  3407  3420 V KOReader:  BackgroundRunnerWidget: start running @  1562338877
07-05 17:01:17.884  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:17.884  3407  3420 V KOReader:  not painting 1 covered widget(s)
07-05 17:01:17.909  3407  3407 V threaded_app: Resume: 0xa662dfe0
07-05 17:01:17.910  3407  3420 V threaded_app: activityState=11
07-05 17:01:17.910  3407  3420 V KOReader:  input event => type: 4, code: 11(4), value: 1, time: 1562338877.910033
07-05 17:01:17.910  3407  3420 V KOReader:  Android application event 11
07-05 17:01:17.910  3407  3420 V KOReader:  not painting 1 covered widget(s)
```

I would expect a "cannot blit" message from https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer_android.lua#L70 if the app tries to write to a window when the surface is not available.

poire-z:
> It behaves exactly as you expected :+1: 

Meaning (a) the highlight is cleared (too early?) or (b) stay highlighted till that button dialog is closed? If (a), is it not bothering enough?

> I would expect a "cannot blit" message

Well, I don't know enough about the android framebuffer/native surface business to give you any hint.
May be try to have KOReader do a full window repaint on Resume, or full transfer from its own buffer (if it has an intermediate one) to the Android surface (if it is different). May be the regional highlight-revert was lost and not pushed to android, and after the resume, only other regional paintings/refreshes were pushed to Android?

pazos:
@poire-z

> Meaning (a) the highlight is cleared (too early?) or (b) stay highlighted till that button dialog is closed? If (a), is it not bothering enough?

sorry, I was not clear enough. I meant (b) stay highlighted till that button dialog is closed. All good.

> Well, I don't know enough about the android framebuffer/native surface business to give you any hint.
May be try to have KOReader do a full window repaint on Resume, or full transfer from its own buffer (if it has an intermediate one) to the Android surface (if it is different). May be the regional highlight-revert was lost and not pushed to android, and after the resume, only other regional paintings/refreshes were pushed to Android?

The surface gets destroyed when the application is stopped (not when it is paused). That's why the log above has no "surface created/destroyed/changed" message). Anyways, from KOReader pov we just try to get a native buffer, lock the buffer to write and then unlock the buffer. These three steps obviously won't work if we cannot get the buffer on the first time (hence the cannot blit to a window).

In that specific case the log is saying: 

- KOReader is paused when we do a dict lookup.
- The "pause" callback  started on the Java framework, reachs our code via the app_glue (that is logged as threaded_app)
- KOReader still works on the background (that's why you can see the Background Runner even when the activity is paused. In fact it should still tick on stopped state).
- The messages saying "Explicit widgetRepaint: table: 0x91a86710 @ ( 18 , 1502.5 ) // not painting 1 covered widget(s)" could be related to this specific glitch.
- Even when KOReader gains focus again and the activity is resumed the messages are still saying "not painting 1 covered widget(s)".

In any case I think it can be fixed later. AFAICT this is the only issue that I found with external dicts.

poire-z:
> the messages are still saying "not painting 1 covered widget(s)".

That's not related to your issue. It's just some optimisation in the KeyValuePage (and some other full screen widgets) when they are redrawn, to not redraw the widget that are behind it (eg: the FileManager or the Reader showing the book) because it's fully covered and it's useless (before that, each time you scrolled a page in KeyValuePage, crengine would be redrawing the book page - or FileManager would be redrawing the cover thumbnails).
Remove the following (or set it to false), and you won't see that message anymore:
https://github.com/koreader/koreader/blob/dca24e0ac57f8f5b50077727aa1128864d586b11/frontend/ui/widget/keyvaluepage.lua#L283-L285

The other bit is from:
https://github.com/koreader/koreader/blob/dca24e0ac57f8f5b50077727aa1128864d586b11/frontend/ui/uimanager.lua#L960-L968
which is called by this:
https://github.com/koreader/koreader/blob/dca24e0ac57f8f5b50077727aa1128864d586b11/frontend/ui/widget/keyvaluepage.lua#L231-L252

So, you should have seen one `Explicit widgetRepaint` before all the Package enabled checks, and the log above shows the second one after the pause (from tickAfterNext) (with y=1502, which might be ok if you have a super tall window :) otherwise it's strange).

> In any case I think it can be fixed later. AFAICT this is the only issue that I found with external dicts.

Fine with me, so go ahead merging :)

mafein:
Dear all,

There are livio dictionary apps, which have their database from wiktionary, and are lightweight (or look up dictionary that connects to livio and shows in popup) . Unfortunately, koreader doesn't work with it. Is it possible to integrate a 3rd party android app with koreader Android version? If not how can I suggest it to developers? 
Also sharing the dictionary lookup history wihh an anki kind of app (independent of above app) , would also be great.

https://play.google.com/store/apps/d...ack.lang.en_US

https://play.google.com/store/apps/d...urav.lookuppro

Frenzie:
I meant "see" a bit more literally than repeating the same thing in this old PR. The forum is fine for the moment, or otherwise a new issue

PR #5117, opened 2019-07-14
cre: toggable enhanced rendering (floats, collapsed margins...)

### LINKED ISSUE #2652: Floating image with caption does not respect width/float
* KOReader version: v2015.11-898-gdccbf4c
* Device: Kobo Glo HD

#### Issue
I've got an epub where I need to float images together with a caption, sized in percent.
Koreader renders the size correct when sizing the image directly (no caption) but ignores the float.
If a floating div wrapper is used for sizing with the enclosed image sized 100% then koreader renders the image at 100%, ignoring the size of the surrounding wrapper.

Koreader seems to use the size of the image based on the body, disregarding the size of the parent wrapper.

Is there any way to at least get the image to be rendered at the proper size (without the float)?

Thanks!

This is how it looks in Nickel/Calibre/ADE/iBooks:
screen_001

This is how Koreader renders it:
reader_2017-mar-16_232618

#### Steps to reproduce

See example epub
[test.epub.zip](https://github.com/koreader/koreader/files/849269/test.epub.zip)

### LINKED ISSUE #2843: CSS auto margins are ignored, cannot center element
* KOReader version: v2015.11-1024-g3b813c7
* Device: Emulator

#### Issue
Centering an element with left and right margins set to auto doesn't work.
px/em/% based width doesn't matter.
Koreader renders at 100% width.

Expected:
screen shot 2017-05-02 at 22 51 11

Koreader: 
screen shot 2017-05-02 at 22 53 05

CSS:
```
div.center {
  margin: 0 auto;
  width: 200px;
  background-color: silver;
}
```

This hack-y way doesn't work either: (px or % doesn't make a difference)
CSS:
```
div.wrap {
  text-align: center
}

p.block {
  text-align: left;
  width: 200px;
  background: gray;
  display: inline-block
}
```

HTML:
```
<div class="wrap">
<p class="block">Lorem ipsum...</p>
</div>
```

ADE/Calibre/Sigil/iBooks behave as expected

#### Steps to reproduce
[untitled.epub.zip](https://github.com/koreader/koreader/files/971953/untitled.epub.zip)

### LINKED ISSUE #2858: CSS: Margins do not collapse properly
* KOReader version: v2015.11-1024-g3b813c7
* Device: Emulator

#### Issue
Margins do not collapse properly.
Only the bigger margin between two vertically adjoining boxes should be honored, unless negative.
See https://css-tricks.com/what-you-should-know-about-collapsing-margins/
and https://www.w3.org/TR/CSS2/box.html#collapsing-margins
The margins are not straight up added but still way too big.

Example:
Heading: margin-bottom: 5em
Text: margin-top: 5em
Total margin between these two elements should be 5em
Koreader draws about 8ems worth of margin

Koreader:
screen shot 2017-05-04 at 23 42 54

ADE/Sigil/iBooks/Calibre:
screen shot 2017-05-04 at 23 42 27

#### Steps to reproduce
[untitled.epub.zip](https://github.com/koreader/koreader/files/977657/untitled.epub.zip)

### LINKED ISSUE #2878: Negative margins are ignored
* KOReader version: v2015.11-1024-g3b813c7
* Device: Emulator

#### Issue
Negative margins are ignored

Koreader:
screen shot 2017-05-06 at 16 45 23

Sigil/iBooks/Calibre (ADE can't do the three-column-ul):
screen shot 2017-05-06 at 16 45 09

#### Steps to reproduce
[margins.epub.zip](https://github.com/koreader/koreader/files/985114/margins.epub.zip)

### LINKED ISSUE #3432: Some CSS weaknesses
* KOReader version: all
* Device:

#### Issue
Using Koreader for quite a long time, I can see that some CSS properties are just ignored and this causes some surprising display glitches. I  am unable to provide a comprehensive study but this is what I regularly find and it would be a good beginning:
 
_text-transform_ property (uppercase, lowercase) not applied
_vertical-align_ property (specially text-top value) not applied
_margin_ property (-top, -bottom, etc.) negative values not supported
_float_ property not appplied

thanks for your attention and for a wonderful reading tool

Roger

poire-z:
Adds a "Render mode" toggle with 4 modes, to enable some or all of crengine's new enhanced block rendering features.
Use "legacy" rendering for previously opened books to not mess with bookmarks, and "web" (all of the new features) for new books.

<kbd>image</kbd>

Details and screenshots at https://github.com/koreader/crengine/pull/299

(Unit tests updated mostly because of collapsed margins which make fewer pages, with sample.txt too, and H3 ActI H3 Prologue in same DocFragment now on a same page in juliet.epub.)

Closes #2652 Floating image with caption does not respect width/float
Closes #2843 CSS auto margins are ignored, cannot center element
Closes #2858 CSS: Margins do not collapse properly
Closes #2878 Negative margins are ignored
Closes #3432 Some CSS weaknesses

Also:
- Prettier CSS when viewing HTML 
- Style tweaks: update for latest epub.css, added a few more 
- Update Wikipedia EPUB stylesheet to allow floating thumbnails:
<kbd>wikipedia1</kbd>

bump crengine:
- Fix bad word spacing in lines ending with a single char word https://github.com/koreader/crengine/pull/298
- Fix "list-style-position: inside" with "list-style-type: none"
- Fix missing in-page footnotes from imbricated tables
- Avoid page split with lines taller than page height
- CSS: adds more attribute selectors, handle case insensitivity flag
- Allow reentrant use of LVFormatter https://github.com/koreader/crengine/pull/299
- EBR1/9 Enhanced Block Rendering: selectable features set
- EBR2/9 CSS: parse and store "float:" and "clear:"
- EBR3/9 DOM/initRendMethod: wrap floats in a floatBox element
- EBR4/9 Upgrade RenderRectAccessor to cache new fields
- EBR5/9 Implement enhanced renderBlockElement()
- ERB6/9 Enhanced final block rendering and text formatting
- EBR7/9 Update drawing code to handle floats and overflows
- EBR8/9 Update XPointers/pt translation methods to handle floats
- EBR9/9 Add "-cr-only-if:" private CSS property, update epub.css
- Adds HTML5 suggested rendering stylesheet (html5.css) https://github.com/koreader/crengine/pull/300

bump base
- Bump FreeType to 2.10.1 https://github.com/koreader/koreader-base/pull/932
- [Android] some tolino/nook waveform fixes https://github.com/koreader/koreader-base/pull/933
- Update FBInk to 1.17.0 https://github.com/koreader/koreader-base/pull/934


Frenzie:
@poire-z I'll tag the 2019.07 release today or tomorrow and then we'll merge this after?

poire-z:
> I'll tag the 2019.07 release today or tomorrow and then we'll merge this after?

Yep, fine with me (it felt like you were going to skip the july release :) so I went ahead with this).

Frenzie:
I like to put in a little bit more effort in terms of writing release notes & testing. :-)

(But I do use it regularly, so at least there are no bugs that break my personal workflow. :-D)

frontend/ui/data/css_tweaks.lua
```diff
@@ -83,6 +133,14 @@ h4, h5, h6 { page-break-after: avoid !important; }
                 title = _("Justify all elements"),
                 description = _("Text justification is the default, but it may be overridden by publisher styles. This will re-enable it for all elements, which may lose centering in some of them."),
                 css = [[* { text-align: justify !important; }]],
+                separator = true,
+            },
+            {
+                id = "headings_align_center",
+                title = _("Center headings"),
+                description = _("Middle align headings."),
```
Frenzie: I'm not sure if that's really any clearer than `center align`, if not actually confusing? I think the implicit "align" part is the important one. Middle align means vertical "centering" not just in CSS but also in Excel/Calc etc.

frontend/ui/data/css_tweaks.lua
```diff
@@ -521,6 +579,19 @@ switch > case    { display: inline; }
 switch > default { display: none; }
             ]],
         },
+        {
+            id = "pure_black_and_white";
+            title = _("Pure black and white"),
+            description = _([[Enforce all text and borders to be black, and backgrounds transparent.]]),
```
Frenzie: Enforce black text and borders, and transparent backgrounds.

frontend/ui/wikipedia.lua
```diff
@@ -95,6 +95,11 @@ function Wikipedia:getWikiServer(lang)
     return string.format(self.wiki_server, lang or self.default_lang)
 end
 
+-- Say who we are to Wikipedia (see https://meta.wikimedia.org/wiki/User-Agent_policy)
```
Frenzie: Incidentally, what did it say before? Some standard LuaSec thing? Nothing?

poire-z: Just `LuaSocket 3.0-rc1`.
Which wasn't causing any issue with Wikipedia, because it's a little known library.

But doing the same kind of multiple images download with "python-requests" was causing many images requests to just get some HTML error message with that URL.

So, we'll be fine if/whenever `LuaSocket` gets a little more noticed by Wikipedia.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -586,4 +586,8 @@ function ReaderBookmark:getLatestBookmark()
     return latest_bookmark
 end
 
+function ReaderBookmark:hasBookmarks()
+    return #self.bookmarks > 0
```
Frenzie: I just wrote no further comments, but of course just after writing it occurs to me… could it ever be undefined?

poire-z: Shouldn't have happened with the orders things happen. But might as well check for that, so done :)

Frenzie:
@poire-z There may be some issue with inline-block or something, see https://www.mobileread.com/forums/showthread.php?p=3883743

poire-z:
Thanks for the test that rules out text-align (phew :).

```html
    <div class="chapter-title">
      <div class="chapter-title-decoration-above">
        <div class="chapter-title-decoration-above-inner">
          <img alt="image" class="chapter-title-decoration-above" src="../Images/chapter_title_above.png"/>
        </div>
      </div>
      <h1 class="chapter-title">The Author</h1>
```

```css
    img.chapter-title-decoration-above {
        width: 2.1400in;
        height: 0.5667in;
    }
    div.chapter-title-decoration-above {
        text-align: center;
    }
    div.chapter-title-decoration-above-inner {
        margin: 0;
        padding: 0;
        display: inline-block;
        width: 2.1400in;
        height: 0.5667in;
    }


```
We indeed don't support `inline-block`, even in the CSS parsing part, so it's just like it was never there, and the `div.chapter-title-decoration-below-inner` just gets what's a DIV is: `display: block`.

If you replace `display: inline-block` with `display: inline`, all is fine it seems.

So, if it's a classic block DIV, is it normal that its content is not centered?
And why it is centered in `flat` or `book` mode and not in `web` mode?

I think it's because of the witdh with an absolute unit `width: 2.1400in`, which is ensured in `web` mode but not the others.

So, in `web` mode, we get a 2inch DIV block (containing the image) in another DIV with `text-align: center`.
But `text-align: center` does not center blocks (only `margin-left/right: auto` helps with that), it only centers the text. So, the 2inch div is just put on the left, and it's content (the image) might be centered itself in that 2inch width.

In `flat` and `book` mode, we don't ensure absolute sized widths, so it defaults to taking all the available width. That 2inch DIV is so not 2inch but takes the whole width of its container. It inherits text-align, so its content (even an image) gets properly aligned, inside that whole width.
If you remove `width: 2.1400in;`, all is fine it seems.

So, yes, not supporting `inline-block` was the cause of all that, but quite indirectly :)

(With that limitation, is my understanding/description conforms to what should happen, and we do all  that right?)

Frenzie:
> So, if it's a classic block DIV, is it normal that its content is not centered?
And why it is centered in flat or book mode and not in web mode?
>
> I think it's because of the witdh with an absolute unit width: 2.1400in, which is ensured in web mode but not the others.

Yes, that all sounds correct and is additionally confirmed by a quick double check in Firefox and Vivaldi.

PR #5121, opened 2019-07-18
Add percentage read (and other book information) to sleep message

sladflob:
This is one feature I miss from the stock Kobo firmware, the ability to see the percentage read on the screen while the device is asleep. I know I can see the book status but it's not quite the same.

I thought a good way to do it would be to create escape sequences that the user can put in their screensaver text. I've modified screensaver.lua to use %p for percentage read. While I was at it I added %c and %t for current and total pages respectively, and %dt for document title.

Attached is a picture of the screen of my Kobo Aura One sleeping...

IMG_20190718_093423

### DIFF
--- a/frontend/ui/screensaver.lua
+++ b/frontend/ui/screensaver.lua
@@ -174,6 +174,7 @@ function Screensaver:setMessage()
     end
     self.input_dialog = InputDialog:new{
         title = "Screensaver message",
+        description = _("The message to be displayed by the screensaver. The following escape sequences may be used in the message:\n  %p percentage read\n  %c current page number\n  %t total number of pages\n  %T title"),
         input = screensaver_message,
         buttons = {
             {
@@ -384,10 +385,10 @@ end
 
 function Screensaver:expandSpecial(message, fallback)
     -- Expand special character sequences in given message. Use fallback string if there is no document instance
-    -- %p percentage read, with percent sign
+    -- %p percentage read
     -- %c current page
     -- %t total pages
-    -- %dt document title
+    -- %T document title
 
     local ret = message
 
@@ -402,10 +403,10 @@ function Screensaver:expandSpecial(message, fallback)
         ret = string.gsub(ret, "%%t", totalpages)
 
         local percent = math.floor(((currentpage * 100) / totalpages) + 0.5)
-        ret = string.gsub(ret, "%%p", percent .. "%%")
+        ret = string.gsub(ret, "%%p", percent)
 
         local props = doc:getProps()
-        ret = string.gsub(ret, "%%dt", props.title)
+        ret = string.gsub(ret, "%%T", props.title)
     else
         ret = fallback
     end
--- a/frontend/ui/screensaver.lua
+++ b/frontend/ui/screensaver.lua
@@ -174,7 +174,7 @@ function Screensaver:setMessage()
     end
     self.input_dialog = InputDialog:new{
         title = "Screensaver message",
-        description = _("The message to be displayed by the screensaver. The following escape sequences may be used in the message:\n  %p percentage read\n  %c current page number\n  %t total number of pages\n  %T title"),
+        description = _("Enter the message to be displayed by the screensaver. The following escape sequences can be used:\n  %p percentage read\n  %c current page number\n  %t total number of pages\n  %T title"),
         input = screensaver_message,
         buttons = {
             {
--- a/frontend/ui/screensaver.lua
+++ b/frontend/ui/screensaver.lua
@@ -349,9 +349,14 @@ function Screensaver:show(event, fallback_message)
         if screensaver_message == nil and prefix ~= "" then
             screensaver_message = G_reader_settings:readSetting("screensaver_message")
         end
+
+        local fallback = fallback_message or default_screensaver_message
         if screensaver_message == nil then
-            screensaver_message = fallback_message or default_screensaver_message
+            screensaver_message = fallback
+        else
+            screensaver_message = self:expandSpecial(screensaver_message, fallback)
         end
+
         widget = InfoMessage:new{
             text = screensaver_message,
             readonly = true,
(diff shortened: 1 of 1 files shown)


sladflob:
And a screenshot of the dialog to configure this sleep message:

Reader_2019-Jul-18_100051

frontend/ui/screensaver.lua
```diff
@@ -377,6 +382,38 @@ function Screensaver:show(event, fallback_message)
...
+    local doc = DocumentRegistry:openDocument(lastfile)
+    local instance = require("apps/reader/readerui"):_getRunningInstance()
+    if instance ~= nil then
+        local currentpage = instance.view.state.page
+        ret = string.gsub(ret, "%%c", currentpage)
+
+        local totalpages = doc:getPageCount()
+        ret = string.gsub(ret, "%%t", totalpages)
+
+        local percent = math.floor(((currentpage * 100) / totalpages) + 0.5)
```
Frenzie: We also have our own `Math.round()` function which, admittedly, does exactly this. But the advantage is it's easier to search the source to see where it's used. :-)

https://github.com/koreader/koreader/blob/93b6335421dcc63321e1aac3526b005aa10da52a/frontend/optmath.lua#L19-L21

frontend/ui/screensaver.lua
```diff
@@ -377,6 +382,38 @@ function Screensaver:show(event, fallback_message)
...
+    -- %dt document title
+
+    local ret = message
+
+    local lastfile = G_reader_settings:readSetting("lastfile")
+    local doc = DocumentRegistry:openDocument(lastfile)
+    local instance = require("apps/reader/readerui"):_getRunningInstance()
+    if instance ~= nil then
+        local currentpage = instance.view.state.page
+        ret = string.gsub(ret, "%%c", currentpage)
```
Frenzie: It's indeed also nice and simple this way. @poire-z any thoughts on this "custom" thingy vs the regular template function?

poire-z: Our current [template function](https://github.com/koreader/koreader-base/blob/bf8faaf2d7845c08ca2ca1c32568b4163816ea74/ffi/util.lua#L590-L599) (which uses the same kind of `string.gsub`) only handles %num => param #num, so it wouldn't work for this use case. So I'm fine with it this way.

But we could easily add a function `util.template2(str, arr)` (better name?) and provide it with `{ c=currentpage, t=totalpages ...}` so it does the gsub work there, and we have something shorter here :)

Frenzie: I'm not sure if named replacement strings are really more user friendly than a simple %1, %2, etc., unless perhaps they were truly named.

poire-z: But the simple %1 %2 doesn't allow any random variables order (if we want to allow freedom if the template string comes from elsewhere)

Frenzie: Huh? Random variable order is the entire reason I wrote that function in the first place! :-)

sladflob:
Personally I think using a string is easier for a user to remember than %1, %2 etc. The template function is fine for using in the code because you've got the string and the replacements right next to each other in the function call, but seems a bit obscure to expose to a user? I take the point that %p etc are scarcely better than numbers as they'd both need to be documented somewhere. And %dt is probably too obscure for document title but I couldn't reuse %t. Maybe make them more explicit like %percent, %page, %total, %title?

frontend/ui/screensaver.lua
poire-z: Oh ok, never ever thought they were different :| (Lots of discussion in #1176 to come up with that :)

But anyway, a tad easier to remember: %c is current page, %p is percentage - than: %1 is current page, %4 is percentage :)

Frenzie:
Yes, `%percent` would be self-explanatory, at least while reading. However, they need to be documented regardless.

For that you should add a `description` to the InputDialog.

https://github.com/koreader/koreader/blob/ce7c5cad48f17a7f4b106c8293ecdfbb1d7f32ce/frontend/ui/screensaver.lua#L175-L196

poire-z:
(I'm fine with this PR as-is - just continuing the discussion for discussion sake :)

> And %dt is probably too obscure for document title but I couldn't reuse %t. Maybe make them more explicit like %percent, %page, %total, %title?

Well, if it's one single char, it's easy. And you could have 26 lowercase letters + 26 uppercase letters + 10 numbers to make out something easier than `%dt`? `%T` may be? Somehow, uppercase = document metadata, lowercase = others.

Multiple letters like `%dt` `%percent` may be harder/ambiguous, and may need some kind of boundaries like `%{percent}` which is less pretty and fun and more risky :)

sladflob:
Good idea re "T" for title. Also, I changed %p to not include the percent sign so it's consistent with the others (I only included it because I couldn't see it on the keyboard - d'oh!). I've added a description field to the dialog - how do I see this when testing?

sladflob:
OK, I must have done something wrong then...

sladflob:
Oh yeah, I put it in the wrong spot. Here it is... I have written a short essay.

Reader_2019-Jul-18_181107

sladflob:
(got rid of the spurious trailing ".")

frontend/ui/screensaver.lua
```diff
@@ -174,6 +174,7 @@ function Screensaver:setMessage()
     end
     self.input_dialog = InputDialog:new{
         title = "Screensaver message",
+        description = _("The message to be displayed by the screensaver. The following escape sequences may be used in the message:\n  %p percentage read\n  %c current page number\n  %t total number of pages\n  %T title"),
```
Frenzie:
```suggestion
        description = _("Enter the message to be displayed by the screensaver. The following escape sequences can be used:\n  %p percentage read\n  %c current page number\n  %t total number of pages\n  %T title"),
```

frontend/ui/screensaver.lua
sladflob: OK cool, I'll use that then.

Frenzie:
Thanks!

frontend/ui/screensaver.lua
Frenzie: Oops, I forgot about that (am on my phone). Minor follow-up PR if you want? Doesn't matter too much.

sladflob: Might as well finish it off properly.

AlanSP1:
Thanks, added new info to wiki: https://github.com/koreader/koreader/wiki/Screen-saver-options

protist:
This is pretty nice! Is there any way to have the message overlaid on a different "background", e.g. the cover?

sladflob:
Thanks! I was thinking this would be a good thing too. No idea how hard it is... maybe I'll investigate when I get some time.

protist:
@sladflob Great! Shall I open an issue so it doesn't get lost?

KenMaltby:
You might include an option for the user to position the box on the page.

sladflob:
Sure @protist, and maybe add @KenMaltby's idea about positioning and I'll see how I go.

protist:
Thanks @sladflob!

PR #5123, opened 2019-07-18
Update screensaver.lua

sladflob:
Following on from previous PR #5121,  change to use Math.round()


frontend/ui/screensaver.lua
```diff
@@ -14,6 +14,7 @@ local logger = require("logger")
 local _ = require("gettext")
 local Screen = Device.screen
 local T = require("ffi/util").template
+local Math = require("optmath")
```
Frenzie: Could you please stick it on line 10? See https://github.com/koreader/koreader-base/wiki/Coding-style

Issue #3578, opened 2018-01-06
File manager - add option to view non-ebooks files


mwoz123:
* KOReader version:
* Device:

#### Issue

Koreader file manager seems to be enought mature file browser to be able to handle operations also on non-ebooks files. 

I'd love to see  feature to disable only ebooks filter in koreader file browser.
Which means to add option to view also non - ebooks format e.g. Zips, rars, images etc.

Frenzie:
Agreed, it makes sense. The biggest question is where to put it. In principle it's the same kind of thing as "Show hidden files" but just sticking it there might contribute to overflowing the settings menu.

![screenshot_2018-01-06_13-02-24](https://user-images.githubusercontent.com/202757/34639529-db3b990a-f2e1-11e7-82dd-3816c2f0deb1.png)

Technically there are a few more concerns since it's currently presumed that displayed files can be opened.

KenMaltby:
Some images are listed and can be selected and will display (.png, for instance) as do the compressed .cbz files.  The other files that an advanced user or developer might wish to see or manipulate would be better handled by some sort of tiny Notepad++ plugin. 

If just viewing is the objective, what's on the device could mostly be viewed once they are added to the reader engine's supported list.  (as txt files for the script files, of course)

robert00s:
What do you think about something like this?
`Show files` instead of `Show hidden files` and submenu

![koreader_066](https://user-images.githubusercontent.com/22982594/35167759-edbf713a-fd56-11e7-91f8-28bd92246c2d.png)
![koreader_067](https://user-images.githubusercontent.com/22982594/35167768-f3c94d12-fd56-11e7-9b00-aab714b942dc.png)

Frenzie:
It does make sense, but at least on Kobo the show hidden files toggle is quite useful where it is. As such I might be more inclined to stick the all files vs documents under `Display mode` and to keep `Show hidden files` where it is. I'm not quite sure if that's the best generic option though.

robert00s:
Or maybe we should split the filemanager settings into two parts:

![koreader_068](https://user-images.githubusercontent.com/22982594/35169331-74d4d46c-fd5c-11e7-880a-e334658e75ef.png)

![koreader_069](https://user-images.githubusercontent.com/22982594/35169332-7762d6ca-fd5c-11e7-9714-e7853d918f43.png)

KenMaltby:
[...]

PR #5129, opened 2019-07-20
File manager - adding option to view unsupported files

robert00s:
New option to show all files (also not supported by KOReader) in file manager. Default this option is disabled.
Close: #3578 

Screenshots:

obraz

obraz

obraz


Frenzie:
Thanks!

Issue #4842, opened 2019-03-26
FR: Make gesture intervals configurable


Frenzie:
* KOReader version: v2019.03/master
* Device: all

#### Issue
They're currently hard-coded.

https://github.com/koreader/koreader/blob/cc9f58687a8a8504d9a81928c707fced51e79b6d/frontend/device/gesturedetector.lua#L53-L56

See https://www.mobileread.com/forums/showthread.php?p=3824560#post3824560

robert00s:
Where we should put config dialog?
Option 1: `Setting -> Taps and gestures`
Option 2: `Tools -> Advanced settings`

Frenzie:
Taps & gestures works for me. :-)

robert00s:
I'm not sure that limit values are correct.
![obraz](https://user-images.githubusercontent.com/22982594/61459302-be8d3c80-a96c-11e9-865e-4f532ee4f04f.png)

poire-z:
If I would play with that, I'll be scared moving away from the defaults, and not being able to reset back to them :)
So, I guess you could show the defaults - or if that would take too much space and make the titles multi-lines, you could add a 3rd button: `Use defaults` (as in `Hyphenation>Left/right minimal sizes`).

robert00s:
Thanks for hint :)

Option 1: (use default button)
![obraz](https://user-images.githubusercontent.com/22982594/61461845-dc10d500-a971-11e9-8a3a-75b9b2e515fb.png)

Option 2: (submenu for each interval + individual widget)

![obraz](https://user-images.githubusercontent.com/22982594/61462017-39a52180-a972-11e9-9a6f-05aa89fc3133.png)


![obraz](https://user-images.githubusercontent.com/22982594/61461994-28f4ab80-a972-11e9-95bb-3ff98436b93d.png)

Frenzie:
Do they really need individual limits? I'd just give 'em all a limit from 50-2000 or something.

robert00s:
We can set all limits from 100 to 2000.

poire-z:
The numberpicker is prettier and easier to use (with a default button too, and may be the current value and if it is the default shown in the menu).

But I don't know if these 5 intervals have any relations (like: one shouldn't be smaller than some other), where it would be easier to have a view of them all with option 1.

Frenzie:
> We can set all limits from 100 to 2000.

Alright.

> But I don't know if these 5 intervals have any relations

[...]

PR #5138, opened 2019-07-23
Gesture intervals configurable

robert00s:
Discussion: #4842 
Close: #4842 

Screenshots:

obraz

obraz

obraz

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -369,12 +369,11 @@ function ReaderGesture:addToMainMenu(menu_items)
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
                         text = T(_([[
-Set double tap interval.
-Time interval is in miliseconds.
-Acceptable value from 100 (0,1 sec) to 2000 (2 sec).
+Set double tap interval in miliseconds.
+The interval value can range from 100 (0.1 seconds) to 2000 (2 seconds).
 Default value: %1]]), GestureDetector.DOUBLE_TAP_INTERVAL/1000),
                         width = Screen:getWidth() * 0.6,
-                        value = GestureDetector.ges_double_tap_interval/1000,
+                        value = GestureDetector:getInterval("ges_double_tap_interval")/1000,
                         value_min = 100,
                         value_max = 2000,
                         value_step = 100,
@@ -384,6 +383,7 @@ Default value: %1]]), GestureDetector.DOUBLE_TAP_INTERVAL/1000),
                         default_value = GestureDetector.DOUBLE_TAP_INTERVAL/1000,
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_double_tap_interval", spin.value*1000)
+                            GestureDetector:setNewInterval("ges_double_tap_interval", spin.value*1000)
                         end
                     }
                     UIManager:show(items)
@@ -396,12 +396,11 @@ Default value: %1]]), GestureDetector.DOUBLE_TAP_INTERVAL/1000),
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
                         text = T(_([[
-Set two finger tap duration.
-Time duration is in miliseconds.
-Acceptable value from 100 (0,1 sec) to 2000 (2 sec).
+Set two finger tap duration in miliseconds.
+The duration value can range from 100 (0.1 seconds) to 2000 (2 seconds).
 Default value: %1]]), GestureDetector.TWO_FINGER_TAP_DURATION/1000),
                         width = Screen:getWidth() * 0.6,
-                        value = GestureDetector.ges_two_finger_tap_duration/1000,
+                        value = GestureDetector:getInterval("ges_two_finger_tap_duration")/1000,
                         value_min = 100,
                         value_max = 2000,
                         value_step = 100,
@@ -411,6 +410,7 @@ Default value: %1]]), GestureDetector.TWO_FINGER_TAP_DURATION/1000),
                         default_value = GestureDetector.TWO_FINGER_TAP_DURATION/1000,
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_two_finger_tap_duration", spin.value*1000)
+                            GestureDetector:setNewInterval("ges_two_finger_tap_duration", spin.value*1000)
                         end
                     }
                     UIManager:show(items)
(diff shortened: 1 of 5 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -359,6 +359,146 @@ function ReaderGesture:addToMainMenu(menu_items)
...
+            {
+                text = _("Double tap interval"),
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local GestureDetector = require("device/gesturedetector")
+                    local items = SpinWidget:new{
+                        text = T(_([[
+Set double tap interval.
+Time interval is in miliseconds.
+Acceptable value from 100 (0,1 sec) to 2000 (2 sec).
```
Frenzie: I don't like "acceptable value". I'll think about it more later. :-)

frontend/device/gesturedetector.lua
```diff
@@ -50,12 +50,19 @@ local util = require("util")
...
     input = nil,
-    -- all the time parameters are in us
+    -- default values (all the time parameters are in us)
     DOUBLE_TAP_INTERVAL = 300 * 1000,
     TWO_FINGER_TAP_DURATION = 300 * 1000,
     HOLD_INTERVAL = 500 * 1000,
     PAN_DELAYED_INTERVAL = 500 * 1000,
     SWIPE_INTERVAL = 900 * 1000,
+    -- all the time parameters are in us
+    ges_double_tap_interval = G_reader_settings:readSetting("ges_double_tap_interval") or 300 * 1000,
```
Frenzie: I realize it's duplicated because you can't reference `self.DOUBLE_TAP_INTERVAL` yet but perhaps it'd be better to define them locally above first for clarity? Then it could just be `or DOUBLE_TAP_INTERVAL`.

frontend/apps/reader/modules/readergesture.lua
NiLuJe: `Accepts value from` if you want to keep the same verb. Otherwise, `supports`, `handles`?

Frenzie: I'd make it something like this.

> Set double tap interval in milliseconds.
> 
> The interval value can range from 100 (0.1 seconds) to 2000 (2 seconds).

robert00s: Thanks :)
Changes have been made.

obraz

frontend/device/gesturedetector.lua
robert00s: Done.

poire-z:
Just a quick note regarding https://www.mobileread.com/forums/showthread.php?t=333958
> I don't now why and I'm wondering if I'm the only one having this anoyance but with a variable frequency (sometime 5 to 6 pages, sometimes more than 10) two pages are swiped at once.
It happens in forward or backward direction.
At the begining I was convinced that it was me double tapping on screen but now I'm sure it's software related as it does not happen with the default BOOX reader

Occasional double tap happened to me quite often when I started using KOReader. I'm sure it's double taps, or bounces (might be easier to trigger on IR screen - or just me, never saw anybody complaining about that :). 

I quickly fixed that for me with:
```diff
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -131,12 +132,14 @@
             id = "tap_forward",
             ges = "tap",
             screen_zone = forward_zone,
+            rate = 1.5,
             handler = function() return self:onGotoViewRel(1) end,
         },
         {
             id = "tap_backward",
             ges = "tap",
             screen_zone = backward_zone,
+            rate = 1.5,
             handler = function() return self:onGotoViewRel(-1) end,
         },
     })
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -337,12 +338,14 @@
             id = "tap_forward",
             ges = "tap",
             screen_zone = forward_zone,
+            rate = 1.5,
             handler = function() return self:onGotoViewRel(1) end,
         },
         {
             id = "tap_backward",
             ges = "tap",
             screen_zone = backward_zone,
+            rate = 1.5,
             handler = function() return self:onGotoViewRel(-1) end,
         },
         {
```
So, now, I can't quickly turn pages with tap, but no more bounces !
So, may be we could add a configurable for that in that Gesture interval menu ?

To be complete for my (mis)use case, I have a few more bounce avoidance handled like that:
```diff
+-- discard past queued and upcoming events
+function UIManager:discardEvents(seconds)
+    if seconds == 0 then
+        self._discard_events_till = nil
+        return
+    end
+    if not seconds then
+        -- If none provided, use an adequate delay to account for
+        -- device refresh delay and discard events got during that
+        -- 1s is good on Kobo to avoid bounces
+        seconds = 1
+    end
+    local now = { util.gettime() }
+    local now_us = now[1] * MILLION + now[2]
+    local s = math.floor(seconds)
+    local usecs = (seconds - s) * MILLION
+    self._discard_events_till = now_us + s * MILLION + usecs
 end

 --- Transmits an event to an active widget.
 function UIManager:sendEvent(event)
     if #self._window_stack == 0 then return end

+    -- ensure discardEvents
+    if self._discard_events_till then
+        local now = { util.gettime() }
+        local now_us = now[1] * MILLION + now[2]
+        if now_us < self._discard_events_till then
+            return
+        else
+            self._discard_events_till = nil
+        end
+    end
+
     local top_widget = self._window_stack[#self._window_stack]
     -- top level widget has first access to the event
     if top_widget.widget:handleEvent(event) then

--- a/frontend/ui/widget/confirmbox.lua
+++ b/frontend/ui/widget/confirmbox.lua
@@ -182,6 +182,7 @@
     UIManager:setDirty(self, function()
         return "ui", self[1][1].dimen
     end)
+    UIManager:discardEvents(0.5) -- discard past events to avoid dismissal
 end

 function ConfirmBox:onCloseWidget()

--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
         }
         table.insert(self.dict_window_list, self.dict_window)
         UIManager:show(self.dict_window)
+        -- so we don't accidentally dismiss it
+        UIManager:discardEvents()
     end
 end
```
(and in a few other places in my added code).

poire-z:
Yes, I did that when I wasn't that familiar with all of KOReader.
I guess now it might be better handled by gesturedetector, with a general timer between taps (like I met in #6186) - which looks like a more complex work :/

NiLuJe:
See also [this bit of nastiness](https://github.com/koreader/koreader/blob/e15b7f7d4b38234b07bb96f1ba53fda976de7f56/frontend/device/input.lua#L382-L402) which I'm responsible for, in a slightly similar vein ;p.

PR #5142, opened 2019-07-24
android: add support for different screen timeout setups

### LINKED ISSUE #4604: [FR] For android devices allow a screen timeout for inactivity when "Keep Screen On" is enabled
* KOReader version: 2019.02.16 nightly
* Device: fire tablet 10

#### Issue
So I love the keep screen on feature and use it when I'm eating so I can use my bluetooth mouse to advance pages.

But sometimes I forget to turn off the tablet, or the mouse gets bumped and then turns the tablet back on and if the keep screen on is enabled I'll end up draining the battery life out of the tablet.

FR would be to have a user definable timeout time either as a number entry UI (like the read timer) or a slider bar.  I think the read timer would be more flexible with say a sensible default like 10 minutes.

Would also be useful to see something like this on the kindle in case I want to change my screen turn off timeout for some reason.

pazos:
requires https://github.com/koreader/android-luajit-launcher/pull/160
fixes #4604

### DIFF
--- a/frontend/ui/elements/screen_android.lua
+++ b/frontend/ui/elements/screen_android.lua
@@ -5,8 +5,41 @@ local logger = require("logger")
 local _ = require("gettext")
 local Input = Device.input
 local Screen = Device.screen
+local T = require("ffi/util").template
 
 if not isAndroid then return end
+
+-- custom timeouts (in milliseconds)
+local timeout_custom1 = 30 * 1000
+local timeout_custom2 = 60 * 1000
+local timeout_custom3 = 2 * 60 * 1000
+local timeout_custom4 = 5 * 60 * 1000
+local timeout_custom5 = 10 * 60 * 1000
+local timeout_custom6 = 15 * 60 * 1000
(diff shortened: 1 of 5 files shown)


frontend/ui/elements/screen_android.lua
```diff
@@ -5,9 +5,42 @@ local logger = require("logger")
...
+local can_modify_timeout = not android.needsWakelocks()
+
+local timeout_system = 0	-- same as system (do nothing!)
+local timeout_custom1 = 30000	-- 30s
+local timeout_custom2 = 60000	-- 60s
+local timeout_custom3 = 120000	-- 2m
+local timeout_custom4 = 300000	-- 5m
+local timeout_custom5 = 600000	-- 10m
+local timeout_custom6 = 900000	-- 15m
+local timeout_custom7 = 1800000	-- 30m
```
Frenzie: You could also consider writing 30 * 60 * 1000 if you think that's clearer. (luacheck would like there to be only spaces I'm afraid btw)

Issue #4757, opened 2019-03-08
The space between lines in the table of contents view


mayabtlaloc:
Could be reduced, or an option to customize it The space between lines in the table of contents view?. It has too much “Line spacing”, which results in having to navigate in a large number of pages. Thanks.
![table of contents line spacing very large](https://user-images.githubusercontent.com/2880036/53999672-eab8c700-4109-11e9-900d-add9bb1e349f.jpg)

poire-z:
Not obvious, but what you set in File browser menu > 1st tab > `Items per page` applies to all these kinds of panels (Files list, TOC, Bookmarks...). You probably have it at 8. I have it set at 20.

mayabtlaloc:
It is a good solution but, unfortunately the size of the font is reduced too much, And in long file names it becomes even smaller, almost microscopic. This is a problem for many people who we have tired eyesight and vision problems.
Is there the possibility of increasing the font size in the lists without sacrificing the number of items per page?
It's something like the filename classic view was before. Could you include an option to have the classic old view?. In which you can have a large font with a large number of items per page, due to a more tight Space between lines. Thanks.
![very small font](https://user-images.githubusercontent.com/2880036/54047354-79762400-419c-11e9-9f6e-99d54d839811.jpg)
![classic old view](https://user-images.githubusercontent.com/2880036/54047403-9579c580-419c-11e9-8a3a-0caf5e05d357.jpg)

mayabtlaloc:
Long names look even smaller. 
![20190309_092712](https://user-images.githubusercontent.com/2880036/54073637-68d2b600-424f-11e9-8b6a-47c7f24ffcf4.jpg)

How can you see large font keeping more items per page?

poire-z:
Currently, you can't, there is no option to tweak the algorithm.

But if you feel like editing a file and playing, you can try adding into `frontend/ui/widget/menu.lua` these 2 lines in green around line 145  (without the `+` at the beginning), and experiment with other values than this `30`, depending on what you set for _item per page_:
```diff
 function MenuItem:init()
+    self.font_size = 30
+    self.single_line = true
     self.content_width = self.dimen.w - 2 * Size.padding.fullscreen
     local shortcut_icon_dimen = Geom:new()
     if self.shortcut then
```

mayabtlaloc:
Perfect poire-z!!, I did not think it could be changed so much by adding two simple lines.
Thank you very much, very grateful for the great work you are doing with all your team.

PR #5146, opened 2019-07-26
Set font size for menu items

robert00s:
Ref: #4757
This PR add option to changing the font size of items in menu widget (filemanger in classic mode, TOC) in single line mode.
Close: #4757

obraz
obraz
obraz
obraz

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -149,8 +149,6 @@ function FileManagerMenu:setUpdateItemTable()
                         title_text =  _("Items per page"),
                         callback = function(spin)
                             G_reader_settings:saveSetting("items_per_page", spin.value)
-                            --after changing items per page we reset the font size of item and set it as default
-                            G_reader_settings:saveSetting("items_font_size", math.floor(24 - ((spin.value - 6)/ 18) * 10 ))
                             self.ui:onRefresh()
                         end
                     }
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -172,7 +172,7 @@ function FileManagerMenu:setUpdateItemTable()
                         value_max = 72,
                         default_value = default_font_size,
                         ok_text = _("Set size"),
-                        title_text =  _("Max font size for item"),
+                        title_text =  _("Maximum font size for item"),
                         callback = function(spin)
                             G_reader_settings:saveSetting("items_font_size", spin.value)
                             self.ui:onRefresh()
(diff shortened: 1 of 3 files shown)


KenMaltby:
Perhaps with a warning that "Changing Font Size will effect "items per page""  that and/or line spacing.

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -126,31 +126,62 @@ function FileManagerMenu:setUpdateItemTable()
         checked_func = function() return self.ui.file_chooser.show_unsupported end,
         callback = function() self.ui:toggleUnsupportedFiles() end
     }
-    self.menu_items.items_per_page = {
-        text = _("Items per page"),
-        help_text = _([[This sets the number of items per page in:
+    self.menu_items.items = {
+        text = _("Items"),
```
Frenzie: Maybe something like `View`? But that's just a synonym for `Display`, already used above when CoverBrowser is enabled. Without CoverBrowser you don't get access to settings like home abbreviation? It might be necessary to give that some attention, even if not necessarily as part of this PR. Also pinging @poire-z

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -126,31 +126,62 @@ function FileManagerMenu:setUpdateItemTable()
...
+                    local default_font_size = math.floor(24 - ((curr_items - 6)/ 18) * 10 )
+                    local curr_font_size = G_reader_settings:readSetting("items_font_size") or default_font_size
+                    local items_font = SpinWidget:new{
+                        width = Screen:getWidth() * 0.6,
+                        value = curr_font_size,
+                        value_min = 10,
+                        value_max = 72,
+                        default_value = default_font_size,
+                        ok_text = _("Set size"),
+                        title_text =  _("Max font size for item"),
```
Frenzie:
```suggestion
                        title_text =  _("Maximum font size for item"),
```
(I don't think we're starved for space here?)

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -126,31 +126,62 @@ function FileManagerMenu:setUpdateItemTable()
...
+                        width = Screen:getWidth() * 0.6,
+                        value = curr_items,
+                        value_min = 6,
+                        value_max = 24,
+                        ok_text = _("Set items"),
+                        title_text =  _("Items per page"),
+                        callback = function(spin)
+                            G_reader_settings:saveSetting("items_per_page", spin.value)
+                            --after changing items per page we reset the font size of item and set it as default
+                            G_reader_settings:saveSetting("items_font_size", math.floor(24 - ((spin.value - 6)/ 18) * 10 ))
```
robert00s: After changing items per page should I change font size for default or keep settings from `maximum font size for item`? Now I reset font size to default.

Frenzie: It seems somewhat unexpected to change that when you're just changing one setting.

robert00s: Ok, I removed it.

PR #5147, opened 2019-07-27
feat: customizable articles count per sync

michivi:
Adds a new client setting representing the number of articles that will be fetched from the Wallabag server for each synchronization.

Previously, Wallabag only returns 30 articles at a time by default. With this setting, koreader will be able to download a custom number of articles from the Wallabag server for each synchronization.

If the setting was not present in the user settings, the default value (30) will be used.

### DIFF
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -40,6 +40,7 @@ function Wallabag:init()
     self.is_auto_delete = false
     self.is_sync_remote_delete = false
     self.filter_tag = ""
+    self.articles_per_sync = 30
 
     self.ui.menu:registerToMainMenu(self)
     self.wb_settings = self.readSettings()
@@ -63,6 +64,9 @@ function Wallabag:init()
     end
     if self.wb_settings.data.wallabag.filter_tag then
         self.filter_tag = self.wb_settings.data.wallabag.filter_tag
+    end
+    if self.wb_settings.data.wallabag.articles_per_sync ~= nil then
+        self.articles_per_sync = self.wb_settings.data.wallabag.articles_per_sync
     end
 
     -- workaround for dateparser only available if newsdownloader is active
@@ -130,6 +134,13 @@ function Wallabag:addToMainMenu(menu_items)
                         keep_menu_open = true,
                         callback = function()
                             self:editServerSettings()
+                        end,
+                    },
+                    {
+                        text = _("Configure Wallabag client"),
+                        keep_menu_open = true,
+                        callback = function()
+                            self:editClientSettings()
                         end,
                     },
                     {
@@ -293,7 +304,7 @@ function Wallabag:getArticleList()
     if self.filter_tag ~= "" then
         filtering = "&tags=" .. self.filter_tag
     end
-    local articles_url = "/api/entries.json?archive=0" .. filtering
+    local articles_url = "/api/entries.json?archive=0&perPage=" .. self.articles_per_sync .. filtering
     return self:callAPI( "GET", articles_url, nil, "", "" )
 end
 
@@ -685,7 +696,12 @@ Restart KOReader after editing the config file.]]), DataStorage:getSettingsDir()
                     text = _("Apply"),
                     callback = function()
                         local myfields = MultiInputDialog:getFields()
-                        self:saveSettings(myfields)
+                        self.server_url    = myfields[1]
+                        self.client_id     = myfields[2]
+                        self.client_secret = myfields[3]
+                        self.username      = myfields[4]
+                        self.password      = myfields[5]
+                        self:saveSettings()
                         self.settings_dialog:onClose()
                         UIManager:close(self.settings_dialog)
                     end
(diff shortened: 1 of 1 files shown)


plugins/wallabag.koplugin/main.lua
```diff
@@ -132,6 +136,13 @@ function Wallabag:addToMainMenu(menu_items)
                             self:editServerSettings()
                         end,
                     },
+                    {
+                        text = _("Configure Wallabag client"),
+                        keep_menu_open = true,
+                        callback = function()
```
Frenzie: Just self.editClientSettings doesn't work here?

plugins/wallabag.koplugin/main.lua
```diff
@@ -700,6 +716,47 @@ Restart KOReader after editing the config file.]]), DataStorage:getSettingsDir()
     self.settings_dialog:onShowKeyboard()
 end
 
+
```
Frenzie: Stray newline.

plugins/wallabag.koplugin/main.lua
```diff
@@ -700,6 +716,47 @@ Restart KOReader after editing the config file.]]), DataStorage:getSettingsDir()
...
+function Wallabag:editClientSettings()
+    self.client_settings_dialog = MultiInputDialog:new {
+        title = _("Wallabag client settings"),
+        fields = {
+            {
+                text = self.articles_per_sync,
+                description = T(_("Articles to fetch:")),
+                input_type = "number",
+                hint = _("Articles to fetch")
```
```suggestion
                hint = _("Number of articles")
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -700,6 +716,47 @@ Restart KOReader after editing the config file.]]), DataStorage:getSettingsDir()
...
 end
 
+
+function Wallabag:editClientSettings()
+    self.client_settings_dialog = MultiInputDialog:new {
+        title = _("Wallabag client settings"),
+        fields = {
+            {
+                text = self.articles_per_sync,
+                description = T(_("Articles to fetch:")),
```
```suggestion
                description = T(_("Articles to fetch")),
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -700,6 +716,47 @@ Restart KOReader after editing the config file.]]), DataStorage:getSettingsDir()
...
+            {
+                {
+                    text = _("Cancel"),
+                    callback = function()
+                        self.client_settings_dialog:onClose()
+                        UIManager:close(self.client_settings_dialog)
+                    end
+                },
+                {
+                    text = _("Apply"),
```
```suggestion
                    text = _("Set number"),
```

plugins/wallabag.koplugin/main.lua
michivi: Have to admit I haven't tried, just to be consistent with the `editServerSettings` call a few lines above.

But after a quick edit, it doesn't seem to work (the function is called, but the settings dialog is empty). I'm quite new to Lua, but I suppose the `self` object has been lost in the callback?

plugins/wallabag.koplugin/main.lua
michivi: My bad! :-)

plugins/wallabag.koplugin/main.lua
michivi: Cool, thanks!

plugins/wallabag.koplugin/main.lua
michivi: I was wondering if `Apply` would be better, as I would like later on to add another setting to archive finished articles instead of deleting them (if the user wants to). Should we change it anyway and see to that potential new feature later?

plugins/wallabag.koplugin/main.lua
michivi: Thanks!

plugins/wallabag.koplugin/main.lua
Frenzie: Or rather, `self` would be referring to something else by that time. ;-)

plugins/wallabag.koplugin/main.lua
Frenzie: Alright, in that case we'd just cause unnecessary work for the translators.

plugins/wallabag.koplugin/main.lua
```diff
@@ -700,6 +716,46 @@ Restart KOReader after editing the config file.]]), DataStorage:getSettingsDir()
     self.settings_dialog:onShowKeyboard()
 end
 
+function Wallabag:editClientSettings()
+    self.client_settings_dialog = MultiInputDialog:new {
+        title = _("Wallabag client settings"),
+        fields = {
+            {
+                text = self.articles_per_sync,
+                description = T(_("Number of articles")),
```
Frenzie: Btw, the description and the hint needn't be the same. 

Anyway, you're not using the template function here.
```suggestion
                description = _("Number of articles"),
```

michivi: Right, thanks!

I also added a validation test to prevent crashes on incorrect values.

PR #5148, opened 2019-07-28
feat(wallabag): archiving instead of deleting articles

michivi:
Adds the option to remotely archive finished Wallabag articles instead of permanently deleting them.

Previous versions permanently deleted the remote finished articles. This behavior is preserved by default. If the archiving setting is used, then the finished articles won't be deleted but remotely archived.

The setting is available in the `Settings` menu as `Archive instead of delete articles`.

Perhaps it would be better (if possible) to regroup all the deletion options in a settings dialog? Also, all the options regarding deletion and archiving might get confusing to the user. I thought about rewriting the labels but how would that impact translation?

### DIFF
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -39,6 +39,7 @@ function Wallabag:init()
     self.is_delete_read = false
     self.is_auto_delete = false
     self.is_sync_remote_delete = false
+    self.is_archiving_deleted = false
     self.filter_tag = ""
 
     self.ui.menu:registerToMainMenu(self)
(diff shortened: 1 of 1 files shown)


plugins/wallabag.koplugin/main.lua
```diff
@@ -178,6 +182,14 @@ function Wallabag:addToMainMenu(menu_items)
                             self:saveSettings()
                         end,
                     },
+                    {
+                        text = _("Archive instead of delete articles"),
```
Frenzie: Is that "mark as read"? Because if so, I think something like `mark as read instead of deleting` would be clearer.
```suggestion
                        text = _("Archive instead of deleting"),
```

michivi: Yes, it is. You're right, it is essentially the `Mark as read` functionality. It seems to be called `archived` in the Wallabag API.

Frenzie: It's just not a term I'm familiar with from the Wallabag website, and if I saw it in the API docs I forgot. :-)

Issue #5144, opened 2019-07-25
[question] Restart KOReader after update


robert00s:
I want to add option to restart KOReader immediately after update. Now after update we have message:
![obraz](https://user-images.githubusercontent.com/22982594/61864920-a0768d80-aed2-11e9-8f14-8b79d04ffec5.png)
and we need manualy restart (Exit menu -> Exit -> Restart KOReader).
I want to change infomessage to confirmbox with option to restart:
![obraz](https://user-images.githubusercontent.com/22982594/61864886-90f74480-aed2-11e9-9e78-db0f0ad7a4f2.png)

Is there any method to run `ReaderMenu:exitOrRestart(callback)`
here:
https://github.com/koreader/koreader/blob/1f6af80817bd9ce55201555817d4ce5e7f8f319c/frontend/ui/otamanager.lua#L254-L256 in `OTAManager:fetchAndProcessUpdate()`?

robert00s:
Ok, I think I found solution. I'll try to use that code:

https://github.com/koreader/koreader/blob/1f6af80817bd9ce55201555817d4ce5e7f8f319c/frontend/device/input.lua#L355-L379

NiLuJe:
Does something like what readergesture.lua does would work?

(i.e., `self.ui.menu:exitOrRestart(function() UIManager:restartKOReader() end)`)

robert00s:
Unfortunately not.
OTAUpdate doesn't know self.ui.menu

robert00s:
But "my" solution works great (tested on Kindle) :)

PR #5150, opened 2019-07-28
Restart KOReader after update

robert00s:
Adding the option to restart KOReader after update.
Close: #5144

Before:
obraz
After:
obraz

### DIFF
--- a/frontend/ui/otamanager.lua
+++ b/frontend/ui/otamanager.lua
@@ -46,6 +46,32 @@ local ota_channels = {
     nightly = _("Development"),
 }
 
+local function showRestartMessage()
+    UIManager:show(ConfirmBox:new{
+        text = _("KOReader will be updated on next restart.\nWould you like to restart now?"),
+        ok_callback = function()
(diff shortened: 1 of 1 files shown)


frontend/ui/otamanager.lua
```diff
@@ -46,6 +46,32 @@ local ota_channels = {
     nightly = _("Development"),
 }
 
+local function showRestartMessage()
+    UIManager:show(ConfirmBox:new{
+        text = _("KOReader will be updated on next restart.\nWould you like to restart now?"),
```
Frenzie: ok_text = _("Restart"),

frontend/ui/otamanager.lua
```diff
@@ -46,6 +46,33 @@ local ota_channels = {
...
+        ok_text = _("Restart"),
+        ok_callback = function()
+            local savequit_caller = nil
```
Frenzie: Just declaring it without = nil has the same effect.

PR #5161, opened 2019-08-01
On (really old) Kindles, don't crash if there's no sysfs entry to rea…

### LINKED ISSUE #5159: Kindle 2/DX crash because they lack a sysfs entry to read battery status
* KOReader version: v2019.07
* Device: Kindle 2 International (B003, K2I)

#### Issue

I've installed KOReader to my K2I via KUAL, but when I attempt to launch it the program crashes before booting into anything. According to the crash log, it has something to do with the inability to register some fonts upon loading the QuickStart help page. There appear to be a number of other potential issues, but loading the fonts is where the program crashes and closes out.

#### Steps to reproduce

1. Get a Kindle 2 International
2. Install KUAL
3. Install KOReader
4. Launch KOReader
5. Check the crash.log

[crash.log](https://github.com/koreader/koreader/files/3454438/crash.log)

NiLuJe:
…d for battery capacity

Parse gasgaucge-info's output instead.

Fix #5159

### DIFF
--- a/frontend/device/kindle/powerd.lua
+++ b/frontend/device/kindle/powerd.lua
@@ -46,8 +46,17 @@ end
 function KindlePowerD:getCapacityHW()
     if self.lipc_handle ~= nil then
         return self.lipc_handle:get_int_property("com.lab126.powerd", "battLevel")
-    else
+    elseif self.batt_capacity_file then
         return self:read_int_file(self.batt_capacity_file)
+    else
+        local std_out = io.popen("gasgauge-info -c 2>/dev/null", "r")
+        if std_out then
+            local result = std_out:read("*all"):match("%d+")
+            std_out:close()
+            return result and tonumber(result) or 0
+        else
+            return 0
+        end
     end
 end
 


frontend/device/kindle/powerd.lua
```diff
@@ -46,8 +46,17 @@ end
...
+    elseif self.batt_capacity_file then
         return self:read_int_file(self.batt_capacity_file)
+    else
+        local std_out = io.popen("gasgauge-info -c 2>/dev/null", "r")
+        if std_out then
+            local result = std_out:read("*all"):match("%d+")
+            std_out:close()
+            return result and tonumber(result) or 0
+        else
+            return 0
```
Frenzie: Could potentially be easier to read if you just stick it at the end? (Not asking for a change, result is the same either way.)

NiLuJe: I'm not sure I get what you mean. Stick what at the end of what? ^^.

Frenzie: I mean you can either write:

```lua
if bla then
  return lala
else
  return 0
end
```

or

```lua
if bla then
  lala
end
return 0
```

I find the latter slightly quicker to understand as a default fallback return value, but perhaps I just like that it's quicker to write. /shrugs

Frenzie: So in this case right before the function's `end`.

NiLuJe: Oh ;).

Yeah, I'd have gone with the full verbose, explicit pyramid, but then I checked what `get_int_property` did, and I kind of liked it ;p.

PR #5162, opened 2019-08-01
Don't use WAL on devices where it's not supported

NiLuJe:
(i.e., truly ancient kernels and weird FS).

In which case, we use TRUNCATE, which might be a tad less terrible than
DELETE on said weird crap FS.

Re #5159


NiLuJe:
Yeah, I'm guessing the issue is mostly because /mnt/us is not actually vfat but a fuse proxy, so, the fault lies in the fuse implementation of said proxy, or fuse itself in such old kernels ;).

Issue #5154, opened 2019-07-31
File manager - ability to change drive


AlanSP1:
I'm not really sure is latest version has this ability as I'm at the moment at pretty old version of koreader, but here's the story:

For some reason, when I tried to view one image within ebook, I entered "twilight zone". Everything slowed down, and, somehow, I couldn't exit image and restart koreader in normal way. I had really strange experience, where I had light on and message that device is suspended at the same time, which could be changed into no message, but no light also. And, image viewer was unresponsive.

I know, s--- can happen, so I went for restart, long press of power (I have Kobo Aura One, otherwise I would go for reset button on my H2O). And, here story begins.

After restart, all settings were deleted, probably crash was culprit. But, also, for some reason, I was on wrong side of devices memory, seeing system files, not user files (i.e. wrong partition, or disk, or however is it correctly called).

Anyway, I couldn't find way out of it to my user partition. I don't know if there's option for that, for changing partition.

I had luck, in history, somehow, info about last book was retained, so I found way out, but, it would be great if there is way (if there's not already) to change partition, in case something like this happen to unsuspecting users.

NiLuJe:
The user partition is mounted at `/mnt/onboard` on Kobo.

Anyone can cd out of it, even in perfectly normal situations, the FM isn't sandboxed ;).

Frenzie:
I think this is fairly thoroughly covered by the home button and folder favorites. That's no help if the settings were somehow lost, I suppose. In that case you'd be forced to navigate to `/mnt/something-or-other` manually, but it should always be reachable.

AlanSP1:
Yes, that's something I realized later, after first shock of seeing some strange "unix" directories, with which I'm unfamiliar, as I usually see only user partition (or whatever it is).

Now I know that I should go to mount directory as @NiLuJe mentioned, but at first shock, average user can really be totally lost. Not sure if something can be done in similar case where everything is lost of settings.

Is there maybe some option to check for settings, if they are lost, to check if there's settings backup somewhere. I mean, I know that at the moment there's no that option, but if it can be programmed?

NiLuJe:
Like @Frenzie mentioned, the Home icon will zap you back to your Home, which *should* be set to the user(s preference, as that's the gist of the Quickstart Guide shown on first boot.

AlanSP1:
Hm, I think (not 100% sure, I was bit shocked at that moment) that when I touched home icon, it asked me to set current dir as home dir.

[...]

PR #5163, opened 2019-08-01
Go Big or Go Home!

NiLuJe:
Make the Home button fallback to sane values in case no custom home directory is set (re #5154 ).
And if we're already home, make it jump to the first page.

(The QuickStart Guide properly introduces the long-press as the way to set defaults, so I'm not too torn up about mostly getting rid of that bit of QoL cuteness that may actually mean people wouldn't remember that long-presses are awesome in KOReader).

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -560,6 +560,7 @@ end
 function FileManager:goHome()
     local home_dir = G_reader_settings:readSetting("home_dir")
     if home_dir then
+        -- Jump to the first page if we're already home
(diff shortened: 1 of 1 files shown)


NiLuJe:
Pinging @pazos about the Android default.

On the one hand, I like that it shows people where we live so they can find the setting manually later, on the other hand, no-one should actually set this as their Home :D.

/sdcard/Downloads might be better? Or /sdcard?

Frenzie:
@NiLuJe It goes to /sdcard.

https://github.com/koreader/koreader/blob/e2961097e7b054d97928b6d188353335372b0db6/platform/android/llapp_main.lua#L30

PR #5165, opened 2019-08-02
Fix: set custom screen DPI

robert00s:
Settings -> Screen -> Screen DPI
When custom DPI is not set we see ugly information with %1:

obraz
This PR should fix it and replace input widget with spin widget.

Before:
obraz

After:

obraz

obraz

obraz

### DIFF
--- a/frontend/ui/elements/screen_dpi_menu_table.lua
+++ b/frontend/ui/elements/screen_dpi_menu_table.lua
@@ -20,7 +20,7 @@ local function setDPI(_dpi)
     Device:setScreenDPI(_dpi)
 end
 
-local function spinWidgetSetDPI()
+local function spinWidgetSetDPI(touchmenu_instance)
     local SpinWidget = require("ui/widget/spinwidget")
     local UIManager = require("ui/uimanager")
     local items = SpinWidget:new{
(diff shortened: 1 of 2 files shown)


Frenzie:
I forget, but does the spin widget allow to hold for custom input? The higher the (physical) DPI the less it matters, but the scaling factor may cause multiple (slightly) distinct renderings in steps smaller than 10.

frontend/ui/elements/screen_dpi_menu_table.lua
```diff
@@ -20,6 +20,28 @@ local function setDPI(_dpi)
...
+        value_max = 900,
+        value_step = 10,
+        value_hold_step = 50,
+        ok_text = _("Set DPI"),
+        title_text = _("Set custom screen DPI"),
+        callback = function(spin)
+            G_reader_settings:saveSetting("custom_screen_dpi", spin.value)
+            setDPI(spin.value)
+            local Event = require("ui/event")
+            UIManager:broadcastEvent(Event:new("UpdateTouchMenu"))
```
poire-z: May be you don't need to add this onUpdateTouchMenu and use a broadcast.
A callback called from a menu item gets passed the touchmenu_instance, and you can call updateItems on it.
For example (even if it's a bit different from your situation):
https://github.com/koreader/koreader/blob/e2961097e7b054d97928b6d188353335372b0db6/frontend/apps/reader/modules/readerrolling.lua#L372-L387

robert00s: I didn't know about that trick :)
Thanks, it works :)

robert00s:
@Frenzie You can tap on numbers (between up and down buttons) and you can enter manual value (for example 152).

Frenzie:
@robert00s Alright, cool. I have no real comments at a glance.

Issue #5166, opened 2019-08-02
Strange default in paged media: fit to page + scroll mode


q3cpma:
Hello

* KOReader version: 2019.07
* Device: Kobo Aura HD

#### Issue
When reading PDFs, navigation pans the pages for no reason.
#### Steps to reproduce
1) Turn 2 pages, no problem
2) Go back 2 pages, you get panning toward to bottom each time
3) Go back 1 page more, to return to the first page with correct layout
4) Turn 2 pages, you get panning toward to top each time
The interesting point is that it only happens with PDFs (EPUBs and CBTs work fine) and that if I set the "Zoom mode" to "fit page" between 3) and 4), I go back to 1) (no problem while going forward).
Note that the "Zoom mode" was already set to "fit page", I just clicked on the ticked box.

Example faulty pdf: https://9p.io/sys/doc/9.pdf
I'll upload my koreader directory if nobody can reproduce.

By the way, an option to disable panning gestures would be extra nice.

Frenzie:
What exactly do you mean by "you get panning toward the bottom"?

q3cpma:
I was wrong, it's not setting the zoom mode that resets the loop, touching the top left to get the little book icon does too (but the top right folded corner does not).

q3cpma:
Panning toward the bottom: the top of the page is cut and I see the interpage gray then the end of the previous page

Frenzie:
Are you using zoom to fit page in combination with scroll mode?

q3cpma:
I didn't change anything (deleted my .apps/koreader then copied the one from the zip) from the defaults. No idea what the "scroll mode" is, sorry.

q3cpma:
Disabling scroll mode by editing defaults.lua worked but I can't find the option in the menus and https://github.com/koreader/koreader/wiki/Change-defaults is wrong, the default is 1 (I looked in the release zip). And even if this "solved" it, I wonder why the problem was there anyway.

q3cpma:
Sorry for being daft, finally found it on the bottom menu.

Frenzie:
Scroll mode can be toggled in the bottom menu. It was made the default in <https://github.com/koreader/koreader/pull/2160>.

Perhaps zoom to page could (prompt to) turn off scroll mode. It's just that something like zoom to content or page width in combination with scroll mode by and large makes for a better experience than zoom to page.

q3cpma:
[...]

PR #5169, opened 2019-08-02
[lang] Rename 'Scroll Mode' to 'Continuous'

Frenzie:
The current name causes people to think of scrolling, which is slightly awkward on E Ink. For example, see <https://github.com/koreader/koreader/issues/5166#issuecomment-517771475>.

> But this on LCD, not E-ink where the lag is unbearable, especially if you want to correct any scrolling over/undershoot.

Scroll mode is a misnomer. It should be renamed continuous. I use "scroll mode" as the default, but I seldom or never scroll. Conversely, you can still scroll in paged mode if a page is longer than the screen. As far as I'm concerned if you're scrolling (panning) you may be doing it wrong, but to anyone reading: you do you. ;-)

## Slightly more awkward alternatives
E.g., the wordier `continuous mode` or `continuous display`.

PS I quoted myself from [here](https://www.mobileread.com/forums/showthread.php?p=3803924#post3803924).


poire-z:
No alternate solution with names for each part of the toggle instead of `on | off`?
(Like `scroll | page` in cre, that I'm glad you didn't change :)
`continuous | paged` ?

Frenzie:
Good suggestion! So then it becomes something like:

`Display: continuous | paged`
`View: continuous | paged`

Frenzie:
> (Like scroll | page in cre, that I'm glad you didn't change :)

Also that's an accident. I didn't realize it was using different words for the same thing. ;-)

Frenzie:
I never minded the `scroll mode` name, but it seems to be a not infrequent misunderstanding that scroll mode implies scrolling. I believe some variation of `continuous` is used by Adobe Acrobat and Evince.

poire-z:
Fine with me.
`continuous` instead of `scroll` is clearer even for cre, except may be for newcomers to follow the code :)
(Adds a bit of unbalance to the buttons' texts' widths, the large `continuous` vs the small `page`.)
I don't really like "paged", but you kept "page" so I'm fine - but `adjective | name` feels ok?

Frenzie:
It could also be `per-page` or something.

NiLuJe:
`paginated`?

poire-z:
> `paginated`?

I prefer the unbalance to that one :)

NiLuJe:
Fair enough, both are fine with me ;).

PR #5170, opened 2019-08-02
Change default zoom mode & add warning message for fit page + scroll

### LINKED ISSUE #5166: Strange default in paged media: fit to page + scroll mode

Frenzie:
The alternative is to disable scroll mode by default, but the default combination of fit page + scroll is a bit odd.

Fixes #5166.

Screenshot_2019-08-02_20-50-47

### DIFF
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -9,6 +9,7 @@ local Device = require("device")
 local Geom = require("ui/geometry")
 local Event = require("ui/event")
 local ImageWidget = require("ui/widget/imagewidget")
+local InfoMessage = require("ui/widget/infomessage")
 local OverlapGroup = require("ui/widget/overlapgroup")
 local ReaderDogear = require("apps/reader/modules/readerdogear")
 local ReaderFlipping = require("apps/reader/modules/readerflipping")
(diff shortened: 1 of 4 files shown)


Frenzie:
I figured a little warning message would be better than a full dialog for starters.

Screenshot_2019-08-04_10-32-12
Screenshot_2019-08-04_10-32-19

poire-z:
I rarely read PDF, but I'm reading one these days.
I get these informative popups makes sense and are informative :)
But I get them too in "reflow mode", where, it looks to the pdf newbie me, they don't make much sense. Or do they?

Frenzie:
Reflow should use something like fit page width if you use scroll mode.

"Proper" reflow would arguably create individual pages of exactly the right size. In this case you can (or should?) use fit to page with page mode. This is what you get with k2pdfopt on your PC. You shouldn't use scroll mode due to overlap offset.

Our reflow implementation only works on a single, individual page. This way it can do it very quickly, so that you barely notice it even on the slowest devices. The typical result is that a page will become longer. I'm not sure if you can actually select fit to page, but the result wouldn't be any good due to the length of the page.

poire-z:
Reflow works ok for me in continuous mode and fit to content width.
I'm switching between reflow (for reading) and continuous (for full page images, I get them larger than with fit to page width). It's a bit bothering getting that warning on switch between these modes - and on opening back that PDF when switching documents.

Frenzie:
Yes, it probably shouldn't keep showing.

poire-z:
I think I might be happy with just:
```diff
--- a/frontend/apps/reader/modules/readerzooming.lua
+++ b/frontend/apps/reader/modules/readerzooming.lua
@@ -119,7 +119,7 @@ function ReaderZooming:onReadSettings(config)
     local zoom_mode = config:readSetting("zoom_mode") or
                     G_reader_settings:readSetting("zoom_mode") or
                     self.DEFAULT_ZOOM_MODE
-    self:setZoomMode(zoom_mode)
+    self:setZoomMode(zoom_mode, true) -- avoid informative message on load
 end

 function ReaderZooming:onSaveSettings()
@@ -340,8 +340,8 @@ function ReaderZooming:genSetZoomModeCallBack(mode)
     end
 end

-function ReaderZooming:setZoomMode(mode)
-    if self.ui.view.page_scroll and self.paged_modes[mode] then
+function ReaderZooming:setZoomMode(mode, no_warning)
+    if not no_warning and self.ui.view.page_scroll and self.paged_modes[mode] then
         UIManager:show(InfoMessage:new{
             text = T(_([[
 %1
```
to only have these when actually switching modes via the menu (which is disabled in Reflow mode), and not when opening the document (where we have the previous zoom mode that we already go the warning in the previous session).

Frenzie:
Not 100% sure about the first one, but sure.

poire-z:
The first one what?

Frenzie:
The one that says "avoid informative message on load." But like I said, sure. ;-)

PR #5171, opened 2019-08-02
fix "empty folder" when accessing nextcloud webdav

jp8:
Using the cloud storage feature to connect to a NextCloud server (using WebDAV protocol) results in "/" for every folder name in the list, and "Empty Folder" messages when trying to access any folder.

The cause is a trailing slash in the response from the NextCloud server that needs to be handled if present.


NiLuJe:
Which, arguably, is the *right* (as per specs) way to specify a URL ;).

Nice catch, though!

Frenzie:
Thanks!

Issue #3588, opened 2018-01-10
href in stardict


alex125:
* KOReader version: v2015.11-1457
* Device: Kobo 

#### Issue
Now that html dictionaries are implemented (thanks a lot, this is really great), is there any chance to also implement links within stardict dictionaries?

The syntax in the dict.dz file is  `<a href="bword://dictionary_entry">text</a>`.
A tap on `text` should jump to the entry `dictionary_entry` (or open it in another window).
Of course a long tap on `text` does already work as long as `text` and `dictionary_entry` are identical but often they are not.

Frenzie:
Getting the link content is probably the hardest part, but of course
already implemented elsewhere. Sounds like a nice enhancement. :)

On Jan 10, 2018 09:37, "alex125" <notifications@github.com> wrote:


   - KOReader version: v2015.11-1457
   - Device: Kobo

Issue

Now that html dictionaries are implemented (thanks a lot, this is really
great), is there any chance to also implement links within stardict
dictionaries?

The syntax in the dict.dz file is <a href="bword://dictionary_
entry">text</a>.
A tap on text should jump to the entry dictionary_entry (or open it in
another window).
Of course a long tap on text does already work as long as text and
dictionary_entry are identical but often they are not.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/koreader/koreader/issues/3588>, or mute the thread

TnS-hun:
> The syntax in the dict.dz file is `<a href="bword://dictionary_entry">text</a>`.

I've checked a dozen HTML dictionaries and those that use links all use that URL format. Out of curiosity: is there any standard where this is mentioned?

Frenzie:
It's the bword "protocol" in the first and only I therefore checked: Etymology Online. :-)

What do you mean by URL format though? I don't see anything against the URL/URI specs in that bword thing. Do you mean their most common occurrence with the HTTP protocol?

Frenzie:
@TnS-hun Example from https://github.com/koreader/koreader/pull/3573#issuecomment-356950297 (oddly enough it seems to refer to itself, but whatever)

```html
[...]

PR #5173, opened 2019-08-03
Dict lookups: have them always interruptible

### LINKED ISSUE #5151: [feature request?] [pw4] [19.08 stable release] Ability to cancel dictionary lookups that are taking too long (fuzzy search turned off)
* KOReader version: 19.08 stable
* Device: PW4

#### Steps to reproduce

1) perform a lookup.
2) get tired of how long its taking
3) want to go the next page, but i have to wait until the dialog pops up for 'item not found'  or the dictionary definition which in some cases has taken so long i thought my kindle crashed and had to soft-reset it. 

a speed demon the paperwhite is not.

poire-z:
They should be now interruptible when fuzzy search is disabled, and on Android.
See https://github.com/koreader/koreader/issues/5151 and https://github.com/koreader/koreader/issues/3588#issuecomment-357467866.
Closes #5151. 
(Tested on my Android 6 phone, hope they work as well on other android versions, with hopefully correct shells.)

Also increase the timeout when downloading Wikipedia HTML, needed when making an EPUB out of https://zh.wikipedia.org/wiki/%E4%B8%AD%E5%8D%8E%E4%BA%BA%E6%B0%91%E5%85%B1%E5%92%8C%E5%9B%BD).

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -673,41 +673,29 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
             end
         end
 
-        local results_str = nil
-        if Device:isAndroid() then
-            local A = require("android")
-            results_str = A.stdout(unpack(args))
-        else
-            local cmd = util.shell_escape(args)
-            -- cmd = "sleep 7 ; " .. cmd     -- uncomment to simulate long lookup time
+        local cmd = util.shell_escape(args)
+        -- cmd = "sleep 7 ; " .. cmd     -- uncomment to simulate long lookup time
 
-            if self.lookup_progress_msg then
-                -- Some sdcv lookups, when using fuzzy search with many dictionaries
-                -- and a really bad selected text, can take up to 10 seconds.
-                -- It is nice to be able to cancel it when noticing wrong text was selected.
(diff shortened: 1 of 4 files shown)


NiLuJe:
Is the pipe closed on abort?

I can somewhat reliably break subsequent dictionary lookups by repeatedly cancelling early on a PW2.

poire-z:
> Is the pipe closed on abort?

Yes, it should, we do that correctly I think:
https://github.com/koreader/koreader/blob/d3e02d05c62bcbbe5492c4d37a27aa585a19742e/frontend/ui/trapper.lua#L384-L433

But we'll wait contiuously if it never terminates for some reason... (unlike Trapper:dismissableRunInSubprocess() where we send a kill before collecting it).

> I can somewhat reliably break subsequent dictionary lookups by repeatedly cancelling early on a PW2.

With this PR or without?
I've never seen that on my Kobo, so no idea why - and if sdcv allows concurent executions.

Frenzie:
On that note, do by break reliably do you mean temporarily (meaning it'll work again after a little while) or "forever"?

NiLuJe:
That was before this PR, yeah.

And it appeared to be "forever" (i.e., had to restart KOReader).

To be clear, when it broke, everything instantly returned an empty (no match) result.

Issue #5155, opened 2019-07-31
[UX] Up button in touch menu is too small


robert00s:
I would like to extend the touch zone for up button in touch menu. Now it is to small for me.
Red rectangle is touch zone.
![obraz](https://user-images.githubusercontent.com/22982594/62210731-2431ed80-b39d-11e9-9803-4a3934ee1f7c.png)

Three option:

1. Extend a bit + move button.
![obraz](https://user-images.githubusercontent.com/22982594/62210899-8985de80-b39d-11e9-8bc4-3c84531b5758.png)

2. Extend a lot + don't move button.
![obraz](https://user-images.githubusercontent.com/22982594/62210829-5c393080-b39d-11e9-856d-f89d3fc3957b.png)
3. We don't change that.

Frenzie:
I suppose being slightly larger (like in 1) would be best given this particular design, but I'd like to change how it's presented. @baskerville made a nice mockup of what it should more or less look like: https://github.com/koreader/koreader/issues/2615#issuecomment-290021189

![menu](https://user-images.githubusercontent.com/202757/62215802-e5edfb80-b3a7-11e9-81f1-7724183723b5.png)

The title at the top there is the menu name (like "Style" or "Style tweaks").

robert00s:
Maybe someday I'll try to change touch menu like in that mockup.

PR #5175, opened 2019-08-03
Extend the touch zone for up button in touch menu.

robert00s:
See: #5155 
Close: #5155 
I chose option 1 (red rectangle is touch zone).

Before:
obraz

After:
obraz

### DIFF
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -465,14 +465,16 @@ function TouchMenu:init()
         -- Add some span to balance up_button image included padding
         HorizontalSpan:new{width = Size.span.horizontal_default},
     }
+    local footer_width = self.width - self.padding*2
     local up_button = IconButton:new{
         icon_file = "resources/icons/appbar.chevron.up.png",
         show_parent = self.show_parent,
+        padding_left = footer_width*0.33*0.1,
+        padding_right = footer_width*0.33*0.1,
         callback = function()
             self:backToUpperMenu()
         end,
     }
-    local footer_width = self.width - self.padding*2
     local footer_height = up_button:getSize().h + Size.line.thick
     self.footer = HorizontalGroup:new{
         LeftContainer:new{


frontend/ui/widget/touchmenu.lua
```diff
@@ -465,14 +465,16 @@ function TouchMenu:init()
         -- Add some span to balance up_button image included padding
         HorizontalSpan:new{width = Size.span.horizontal_default},
     }
+    local footer_width = self.width - self.padding*2
     local up_button = IconButton:new{
         icon_file = "resources/icons/appbar.chevron.up.png",
         show_parent = self.show_parent,
+        padding_left = footer_width*0.33*0.1,
```
Frenzie: Wouldn't it make more sense to base it on the IconButton itself or is that not known yet?

robert00s: Size of Icon Button doesn't know yet. Beside of that better to set up left and right padding to 3,3% of footer size.

poire-z:
Mhhh, my obsession with balance makes me not so pleased with that one :)

Previously, the left of the arrow button was aligned with the left edge of the line separator.
Now, on the screenshot, it _looks_ aligned with the left edge of the menu items's text, which could be a valid decision (but on the right side of the footer, the 0% battery level is aligned with the right edge of the line separator...).
But if it _looks_ aligned, it seems it is by chance :) and it may not with various screen ratio, font size, or display DPI, as the left padding of the text items is the actual size of a checkmark (whether it's invisible or not) - and not dependant of the self.padding used to pad the left of the button.

Personally, I would have prefered to just put no padding_left on the button, and double the padding_right, to get the same tapable width.

Frenzie:
I like padding left as the checkmark width. :)

NiLuJe:
I kinda like it to ;).

And I feel like, given the very shape of the icon, it might not suffer too much from slight horizontal misalignment with the text's edge.

(As I imagine pinpointing accurately the exact right padding to align to the text might be slightly tricky?).

Frenzie:
Trickier than worth the effort I'd say but I doubt a workaround would take too long. Simplest/worst case you duplicate the CheckMark + whatever padding logic (should be a Size.padding of some sort), slightly trickier you'd export it somehow.

poire-z:
OK, we'll see tomorrow on our devices :)

NiLuJe:
IIRC, the checkmark is tricky in that the width is mainly calculated by actually rendering a checkmark glyph and then getting it's bbox width...

Frenzie:
Slightly wasteful; not sure I'd call it tricky. :-P

PR #5156, opened 2019-07-31
 Add vertical edge gesture 

robert00s:
* Add left/right edge swipe gesture
* Move kobolight plugin to ReaderFrontlight
* Remove kobolight plugin
* New gestures - increase/decrease frontlight, increase/decrease frontlight warmth
* Supports frontlight (warmth) change in file manger mode 
* Remove `Frontlight gesture controller` menu from Taps and Gestures

Defalut left edge swipe -> frontlight increase/dicrease (if present)
Default right edge swipe -> frontlight warmth increase/decresase (if present) - needs to be tested (I don't have device that supports this function) 

Ref: https://github.com/koreader/koreader/pull/4240#discussion_r220166095

obraz

obraz

obraz

PR #5179, opened 2019-08-04
Add horizontal edge gestures

robert00s:
Add top/bottom edge swipe gesture like in #5156 
obraz

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -166,6 +166,10 @@ function ReaderGesture:init()
         one_finger_swipe_left_edge_up = Device:hasFrontlight() and "increase_frontlight" or "ignore",
         one_finger_swipe_right_edge_down =  Device:hasNaturalLight() and "decrease_frontlight_warmth" or "ignore",
         one_finger_swipe_right_edge_up = Device:hasNaturalLight() and "increase_frontlight_warmth" or "ignore",
+        one_finger_swipe_top_edge_right = "ignore",
+        one_finger_swipe_top_edge_left = "ignore",
+        one_finger_swipe_bottom_edge_right = "ignore",
+        one_finger_swipe_bottom_edge_left = "ignore",
         two_finger_tap_top_left_corner = "ignore",
         two_finger_tap_top_right_corner = "ignore",
         two_finger_tap_bottom_left_corner = "ignore",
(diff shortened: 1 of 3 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -838,6 +857,14 @@ function ReaderGesture:setupGesture(ges, action)
...
         ratio_w = 1/8, ratio_h = 7/8,
     }
+    local zone_top_edge = {
+        ratio_x = 1/8, ratio_y = 0,
+        ratio_w = 7/8, ratio_h = 1/8,
+    }
+    local zone_bottom_edge = {
+        ratio_x = 1/8, ratio_y = 7/8,
+        ratio_w = 7/8, ratio_h = 1/8,
+    }
```
poire-z: I guess these `1/8` and `7/8` are to exclude the `zone_top_left_corner` (and friends) which are themselves all `1/8` rectangles at corners.
So, I guess this `7/8` must be `6/8`, to leave `1/8` on each side.
Or may be it doesn't matter as there is no conflict with tap/hold on corners, as it is swipe.
And there may be no conflict with each other edge swipes because of different directions.
In which case there is no reason to not use the full width.
So, either an offset of 1/8 and a lenght of 6/8 - or an offset of 0 and a length of 1 (8/8).

(I thought that for consistency, these could also be defined in defaults.lua like the corners' ones are, eg. `DTAP_ZONE_TOP_LEFT`, but I guess there's no reason to override them, so no need to bother.)

Frenzie: > Or may be it doesn't matter as there is no conflict with tap/hold on corners, as it is swipe.

You can imagine swipe from corner though. Not sure if it's worth it. :-)

> I guess there's no reason to override them, so no need to bother.)

I don't know, it's kind of cool to hack touch zones to do whatever you want.

robert00s: Zones was simply taken from kobolight plugin :)
https://github.com/koreader/koreader/commit/c6c8c5a0374e89f6a363f2ba9e78834929db3a04#diff-9cf37ee7352547f3ecc054072ae290d6L20

There is no conflict with tap/hold on corners but I don't think that swipe from one corner to another is useful :)

poire-z: OK, but if they were messy there in the plugin, we could fix them now that they are in the core :)

> I don't think that swipe from one corner to another is useful 

If I read them right, they set the following zones:
image
and if I remember correctly, it's the swipe _starting position_ that matters for checking the zone.

So, a right handed person holding the device with his single right hand could bottom swipe left with his thumb (starting in the bottom right corner) with no issue.
While a left handed person holding the device with his single left hand will have a hard time as it can't start in the bottom left corner, and will have to extend his thumb far right to reach the green zone.
So, I guess for consistency and the various aerodynamics of hands and fingers, they should all be 8/8 (=1) long :) (if that cause no other issues I have not foreseen.)

robert00s: I'm absolutely convinced :)
I prepare patch for it.

poire-z:
Old issue, nothing strictly about it, just an observation, this time about the edge zones thickness:
I have associated some stuff to top and bottom edge swipes, which have a height of 1/8 of screen height.
They may conflict (take precendence) over "tap to follow links", if you have a footnote on the first line of text - and another one on the second line of text - if you want to get the first and avoid the second, you can't, as it will trigger top edge swipe left.
I hacked myself the thickness to 1/16, and given my usual top & bottom margins in reader, this is fine for me.
Dunno if this could or should be fixed some better way (ie. by setting the zone height to the actual reader margins (but with PDF, no such margins...), or letting users configure them (more boring UI work that may not always work). Or just one setting "edge thickness" set to 1/8 or 1/16 and applying to all - but then, less risks for the left and right edges (except may be for people using scroll mode, that may like that? or may want different thickness on either sides).
(Just sharing the thought, to see if others have better ideas - or are just fine with that :) Not really planning any work on that.)

NiLuJe:
Not at home, and I can't even remember how these are set and/or overridden (in terms of z-order, I mean; although I think reader always wins for gestures, IIRC?).

poire-z:
No, they win over Reader (which I think is needed for all our gestures, corner tap... to work :) - they are by default set to "passthrough", which is fine.

The zones are configured there (as locals, so not even easy for a user patch):
https://github.com/koreader/koreader/blob/47dae6c5dd11608c02e478f1cd643a908b449a98/plugins/gestures.koplugin/main.lua#L787-L802

NiLuJe:
Yep, of course. I shouldn't attempt intelligent conversations after 36h of uptime ;p.

PR #5181, opened 2019-08-05
Extend edge zones

robert00s:
See: https://github.com/koreader/koreader/pull/5179#discussion_r310676845


frontend/apps/reader/modules/readerfrontlight.lua
```diff
@@ -71,14 +71,14 @@ function ReaderFrontLight:onChangeFlIntensity(ges, direction)
     local powerd = Device:getPowerDevice()
     local gestureScale
     if ges.direction == "south" or ges.direction == "north" then
-        gestureScale = Screen:getHeight() * 7/8 * 0.8
+        gestureScale = Screen:getHeight() * 0.8
```
poire-z: Was that the reason for me feeling these last few days that swiping left did not change frontlight level as much as before ? :)

Issue #2638, opened 2017-03-12
"Zsync" -> "Share this document" crashes koreader if offline


a-kohout:
* KOReader version: v2015.11-896-g4316284
* Device: Kobo Glo HD

#### Issue
"Zsync" -> "Share this document" crashes koreader if not connected to wifi

crash.log:
```
17-03-12 01:28:09 I: server accepts anonymous access
D: 17-03-12 01:28:09 (zsys): zsys_set_interface (wlan0)
E: 17-03-12 01:28:09 (UDP) error 'Address family not supported by protocol' on bind
luajit: /builds/koreader/nightly-builds/koreader/base/thirdparty/czmq/build/arm-kobo-linux-gnueabihf/czmq-prefix/src/czmq/src/zsys.c:823: zsys_socket_error: Assertion `0' failed.
Aborted
```
#### Steps to reproduce
1. Make sure wifi is disconnected
2. Touch the wrench symbol in reader mode.
3. Touch "Zsync" -> "Share this document"

PR #5184, opened 2019-08-07
Fix: Zsync -> Subscribe to book share crashes

lucarota:
Patch similar to PR #3086.
Same problem #2638 when tap Subscribe to book share.
"Zsync" -> "Subscribe to book share" crashes koreader if not connected to wifi

After this fix when koreader is offline after "Zsync" -> "Subscribe to book share" we are asked to enable wifi.


Frenzie:
Thanks for updating this plugin.

Issue #5186, opened 2019-08-07
Double finger up/down & horizontal left edge gestures


gerhaher:
* KOReader version: v2019.07-52-gaec112f_2019-08-06/
* Device: KA1

#### Issue
Double finger up/down does not work at all on my KA1 in portrait mode. However it seems to work in landscape. So I guess that is the opposite side of this problem:  #4843.

Horizontal left edge gesture gets in conflict with the other left swipe. I have 'Swipe to follow nearest link' enabled, so every time there is a link on the page my bottom/top left edge swipe follows that link instead.

gerhaher:
And a possible rotation issue with horizontal edge gestures:

When rotating I end up 1 or 2 pages back. That's not happening when I rotate with the other gestures in Gesture manager.
 
I rotate counterclockwise (in case it's important to know).

robert00s:
I'll try to prepare fixes. Thanks for reporting.

robert00s:
@gerhaher 
Try the latest version v2019.07-58 (from today). All described problems should be solved.

Frenzie:
@gerhaher Does the thumbs up mean it works? Just asking for the August release. :-)

gerhaher:
> @gerhaher Does the thumbs up mean it works? Just asking for the August release. :-)

Yes, it seems to work.

Frenzie:
Alright, closing then.

PR #5189, opened 2019-08-08
Fix: conflict horizontal edge gesture with link swipe

robert00s:
See: #5186 

> Horizontal left edge gesture gets in conflict with the other left swipe. I have 'Swipe to follow nearest link' enabled, so every time there is a link on the page my bottom/top left edge swipe follows that link instead.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -953,6 +953,9 @@ function ReaderGesture:setupGesture(ges, action)
             "paging_swipe",
             "rolling_swipe",
         }
+        overrides_horizontal_edge = {
+            "swipe_link"
+        }
         overrides_pan = {
             "paging_swipe",
             "rolling_swipe",
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -952,9 +952,15 @@ function ReaderGesture:setupGesture(ges, action)
         overrides_vertical_edge = {
             "paging_swipe",
             "rolling_swipe",
+            "readermenu_swipe",
+            "readerconfigmenu_swipe",
         }
         overrides_horizontal_edge = {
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -953,6 +953,9 @@ function ReaderGesture:setupGesture(ges, action)
             "paging_swipe",
             "rolling_swipe",
         }
+        overrides_horizontal_edge = {
+            "swipe_link"
+        }
```
poire-z: Shouldn't it also overrides the ones that `overrides_vertical_edge` does?
And both `overrides_vertical_edge`  and `overrides_horizontal_edge` might also need to override `readerconfigmenu_swipe` and `readermenu_swipe` (they may not overlap as DTAP_ZONE_MENU.x=1.8 in defaults.lua, but some people may increase that).

For info, some months ago (so it may no more be correct), I checked the various gestures overrides and possible conflicts in https://github.com/koreader/koreader/issues/4856#issuecomment-478274191.

frontend/apps/reader/modules/readerlink.lua
```diff
@@ -26,7 +24,30 @@ local ReaderLink = InputContainer:new{
...
+                handler = function(ges) return self:onTap(_, ges) end,
+            },
+            {
+                id = "swipe_link",
+                ges = "swipe",
+                screen_zone = {
+                    ratio_x = 0, ratio_y = 0,
+                    ratio_w = 1, ratio_h = 1,
+                },
+                handler = function(ges) return self:onSwipe(_, ges) end,
```
poire-z: That new one might also need some override (paging_swipe, rolling_swipe...) I guess.

frontend/apps/reader/modules/readergesture.lua
robert00s: Fully agree :)

PR #5195, opened 2019-08-10
Status bar alignment

### LINKED ISSUE #5044: fr: option to make status bar on the right or left side
if we turn off the progess bar the status bar will be sitting at the middle of the bottom, could we add options to choose the status on left or right side, or even at top or at bottom, thanks...

robert00s:
Close #5044 
Option to align status bar without progress bar.

obraz

obraz

obraz

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -1,5 +1,6 @@
 local Blitbuffer = require("ffi/blitbuffer")
 local BottomContainer = require("ui/widget/container/bottomcontainer")
+local CenterContainer = require("ui/widget/container/centercontainer")
 local Device = require("device")
 local Event = require("ui/event")
 local Font = require("ui/font")
@@ -7,6 +8,7 @@ local FrameContainer = require("ui/widget/container/framecontainer")
 local Geom = require("ui/geometry")
 local HorizontalGroup = require("ui/widget/horizontalgroup")
 local HorizontalSpan = require("ui/widget/horizontalspan")
+local LeftContainer = require("ui/widget/container/leftcontainer")
 local ProgressWidget = require("ui/widget/progresswidget")
 local RightContainer = require("ui/widget/container/rightcontainer")
 local TextWidget = require("ui/widget/textwidget")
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -272,11 +287,11 @@ function ReaderFooter:setupTouchZones()
 end
 
 -- call this method whenever the screen size changes
-function ReaderFooter:resetLayout()
+function ReaderFooter:resetLayout(force_reset)
     local new_screen_width = Screen:getWidth()
     local new_screen_height = Screen:getHeight()
     if new_screen_width == self._saved_screen_width
-        and new_screen_height == self._saved_screen_height then return end
+        and new_screen_height == self._saved_screen_height and force_reset == nil then return end
```
Frenzie: Or perhaps `and not force_reset` so that nil and false are treated the same? (Unless the distinction between nil and false is important.)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -201,10 +203,23 @@ function ReaderFooter:init()
         padding = 0,
         padding_bottom = self.bottom_padding,
     }
-    self.footer_container = BottomContainer:new{
-        dimen = Geom:new{ w = 0, h = self.height*2 },
-        self.footer_content,
-    }
```
poire-z: I never understood the why of these double BottomContainer, and why one had self.height*2  :)
It was introduced in 6938b543, I hope it does not matter with the touch zone now after many other changes since 2014 :) and that DTAP_ZONE_MINIBAR works just as expected.
(BottomContainer does centering by default, so it acted as the centerer I think - might as well use a CenterContainer as you did for clarity.)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -492,6 +507,74 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                    self.settings.align = "center"
+                    self.footer_container = CenterContainer:new{
+                        dimen = Geom:new{ w = 0, h = self.height },
+                        self.footer_content,
+                    }
+                    self.footer_positioner = BottomContainer:new{
+                        dimen = Geom:new{},
+                        self.footer_container,
+                    }
+                    self[1] = self.footer_positioner
```
poire-z: That code is duplicated with the one in `ReaderFooter:init()` that you updated above.
Might be moved in a new `ReaderFooter:updateFooterContainer()` so there's a single copy of that code that could be called from both places.
That would give here:
```lua
                callback = function()
                    self.settings.align = "center"
                    self:updateFooterContainer(true)
                    self:resetLayout(true)
                    self:updateFooter()
                    UIManager:setDirty(nil, "ui")
                end,
```

Frenzie: I second that.

frontend/apps/reader/modules/readerfooter.lua
robert00s: I also don't understand why double BottomContainer wa used :)
Touch zones should work because we have separate method:

https://github.com/koreader/koreader/blob/570180196b70b7fc1be502790652400fce2f8927/frontend/apps/reader/modules/readerfooter.lua#L244-L272

frontend/apps/reader/modules/readerfooter.lua
robert00s: Good point.

PR #5197, opened 2019-08-11
avoid hard-coded path

contrun:
Although the current hard-coded paths work on debian, they may fail on many other distros. Other distros may pack koreader based on the released deb file. I personally have problem using current script on nixos, as it does not use the paths `/bin/bash` and `/usr/lib/`.

### DIFF
--- a/platform/debian/koreader.sh
+++ b/platform/debian/koreader.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 export LC_ALL="en_US.UTF-8"
 
 # writable storage: ${HOME}/.config/koreader.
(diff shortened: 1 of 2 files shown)


platform/debian/koreader.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
Frenzie: As an aside, see https://github.com/koalaman/shellcheck/issues/880#issuecomment-298032886

But since KOReader isn't supposed to be run as root that shouldn't be a concern here.

I figure you might also want to change this one:

https://github.com/koreader/koreader/blob/0a2a1fffe8305555ac431c1926c16c1a0a15af3d/kodev#L1

contrun: Thanks for the information. I've also changed kodev.

Frenzie: Thanks!

PR #5198, opened 2019-08-12
Add missing parentheses

jetomit:
Prevents a crash when the WIFI_MODULE environment variable is not
defined.

### DIFF
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -406,7 +406,7 @@ function Kobo:initNetworkManager(NetworkMgr)
     -- NOTE: Cheap-ass way of checking if WiFi seems to be enabled...
     --       Since the crux of the issues lies in race-y module unloading, this is perfectly fine for our usage.
     function NetworkMgr:isWifiOn()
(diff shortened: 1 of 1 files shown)


NiLuJe:
Ooops. :D

NiLuJe:
For reference, what's the context for `WIFI_MODULE` being undefined in your instance?

jetomit:
> For reference, what's the context for `WIFI_MODULE` being undefined in your instance?

I am running koreader in an Alpine rootfs without the original Kobo software, which normally sets this variable.

Frenzie:
The crash being `attempt to concatenate a nil value`?

jetomit:
Yes.

NiLuJe:
How are you dealing with the potential lack of other relevant env vars? (PRODUCT, PLATFORM, WIFI_MODULE_PATH, among others)?

Frenzie:
@NiLuJe To be fair, the line as written clearly implies a fallback in case of nil. ;-P

jetomit:
> How are you dealing with the potential lack of other relevant env vars? (PRODUCT, PLATFORM, WIFI_MODULE_PATH, among others)?

I’m not, everything seems to just work. :) I have not managed to set up wireless yet due to some driver issues (not related to koreader), but I don’t use it anyway.

PR #5200, opened 2019-08-13
[Android] Enforce LZMA2 1MB chunk size

Frenzie:
According to the manual:

> If you don't specify ChunkSize, LZMA2 sets it to max(DictionarySize, min(256M, max(1M, DictionarySize * 4))).

From my understanding the dictionary should be only 64 kB with the `-mx=1` flag, but memory usage on Android suggests we end up with something larger. Enforcing a 1MB chunk size seems to prevent the problem, while mostly keeping the faster LZMA2 decompression advantage.

Cf. <https://github.com/koreader/android-luajit-launcher/issues/16#issuecomment-520588236>.

<hr>

@pazos Could you double check if I didn't accidentally make a mistake with the Android profiling? :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -339,7 +339,7 @@ androidupdate: all
 	# in runtime luajit-launcher's libluajit.so will be loaded
 	-rm $(INSTALL_DIR)/koreader/libs/libluajit.so
 	# make android update apk
-	cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=1 \
+	cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=1 -mc=1M \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).7z * \
 		-x!resources/fonts -x!resources/icons/src -x!spec
 	$(MAKE) -C $(ANDROID_LAUNCHER_DIR) $(if $(KODEBUG), debug, release) ANDROID_APPNAME=KOReader ANDROID_VERSION=$(ANDROID_VERSION) ANDROID_NAME=$(ANDROID_NAME) ANDROID_FLAVOR=$(ANDROID_FLAVOR)


pazos:
> Could you double check if I didn't accidentally make a mistake with the Android profiling? :-)

Sure, give me a few hours (or a day?).

pazos:
Sorry for the delay. I did a few profiles and found a lot of unrelated surprises.

As you said in https://github.com/koreader/android-luajit-launcher/issues/16#issuecomment-520763366 the memory usage seems related to memory available.

But I'm still being able to reproduce crashes even when the memory usage is low:

KOAS

That was a reproducible crash on Debug build on emulator X86-API25 that fails to load libmupdf.

Frenzie:
You mean even though it successfully extracted?

pazos:
> You mean even though it successfully extracted?

That's the problem, I guess. I cannot say when assets were succesfully extracted until it fails to load a library.

I'm not sure what to do. Moving the asset loader to java, using [commons-compress](https://commons.apache.org/proper/commons-compress/) could help as we can throw an exception if something fails during extraction. It is slooooower, though.

Frenzie:
Can't the assets just live uncompressed where the 7z file currently lives or something? They'd still be regular zipped in the APK.

pazos:
> Can't the assets just live uncompressed where the 7z file currently lives or something? They'd still be regular zipped in the APK.

7z package is copied on the files/assets dir from asset loader stream and then uncompressed. AFAICT uncompressing directly from the stream is not supported.

Frenzie:
No wonder Android is always so short of space with all that useless copy action.

https://bitbucket.org/MartinFelis/love-android-sdl2/src/f40577c9cbf01d640c0a8b77813c95ac477a8a4a/love/src/main/java/org/love2d/android/GameActivity.java?at=master#lines-99:148

https://bitbucket.org/MartinFelis/love-android-sdl2/src/f40577c9cbf01d640c0a8b77813c95ac477a8a4a/love/src/main/java/org/love2d/android/GameActivity.java?at=master#lines-258:306

https://stackoverflow.com/a/38464084

But anyway, I just mean that this whole 7z step seems extra complicated even if there may not be a properly better way to do it.

pazos:
> But anyway, I just mean that this whole 7z step seems extra complicated even if there may not be a properly better way to do it.

The idea of having everything in a single package is good. Maybe we can improve it if we:

- switch from lzma to a format that supports uncompression from stream (to avoid copy)
- migrate all the [asset code from install.lua](https://github.com/koreader/android-luajit-launcher/blob/master/assets/install.lua#L20-L72) to java. Add an exception handler too.
- libs outside the zip file.
- avoid noto (25M)  and make CR rely on droid sans/serif.

NiLuJe:
Do we even actually need Noto/Droid on Android? I'd imagine they're available *somewhere* by default on Android?

Frenzie:
@pazos
> switch from lzma to a format that supports uncompression from stream (to avoid copy)

Just say the word (which format). lzma is nice because it's quick in decompression but pretty much anything goes. Heck, even plain tar would probably do the trick.

@NiLuJe
> Do we even actually need Noto/Droid on Android? I'd imagine they're available somewhere by default on Android?

I believe all Android 4+ devices should ship with Droid, Roboto, and Noto. Stupid, I never even thought about that in relation to our package size but that means we can shed about a third of our weight instantly.

That being said, these fonts will also have bugs from 2011/2012 on the oldest supported devices. Is that something that could be of significance?

Frenzie:
The good news, skipping Droid and Noto is a 15 MB package reduction. Closer to half than a third!

The bad/weird news, the Noto fonts aren't quite the same? We have filenames like `NotoSans-Bold.ttf` and `NotoSans-BoldItalic.ttf`, which only exist for NotoSerif on Android.

```
root@generic_x86:/system/fonts # ls
AndroidClock.ttf
AndroidClock_Highlight.ttf
AndroidClock_Solid.ttf
CarroisGothicSC-Regular.ttf
Clockopia.ttf
ComingSoon.ttf
CutiveMono.ttf
DancingScript-Bold.ttf
DancingScript-Regular.ttf
DroidSans-Bold.ttf
DroidSans.ttf
DroidSansMono.ttf
NotoColorEmoji.ttf
NotoNaskhArabic-Bold.ttf
NotoNaskhArabic-Regular.ttf
NotoNaskhArabicUI-Bold.ttf
NotoNaskhArabicUI-Regular.ttf
NotoSansArmenian-Bold.ttf
NotoSansArmenian-Regular.ttf
NotoSansBalinese-Regular.ttf
NotoSansBamum-Regular.ttf
NotoSansBatak-Regular.ttf
NotoSansBengali-Bold.ttf
NotoSansBengali-Regular.ttf
NotoSansBengaliUI-Bold.ttf
NotoSansBengaliUI-Regular.ttf
NotoSansBuginese-Regular.ttf
NotoSansBuhid-Regular.ttf
NotoSansCanadianAboriginal-Regular.ttf
NotoSansCham-Bold.ttf
NotoSansCham-Regular.ttf
NotoSansCherokee-Regular.ttf
NotoSansCoptic-Regular.ttf
NotoSansDevanagari-Bold.ttf
NotoSansDevanagari-Regular.ttf
NotoSansDevanagariUI-Bold.ttf
NotoSansDevanagariUI-Regular.ttf
NotoSansEthiopic-Bold.ttf
NotoSansEthiopic-Regular.ttf
NotoSansGeorgian-Bold.ttf
NotoSansGeorgian-Regular.ttf
NotoSansGlagolitic-Regular.ttf
NotoSansGujarati-Bold.ttf
NotoSansGujarati-Regular.ttf
NotoSansGujaratiUI-Bold.ttf
NotoSansGujaratiUI-Regular.ttf
NotoSansGurmukhi-Bold.ttf
NotoSansGurmukhi-Regular.ttf
NotoSansGurmukhiUI-Bold.ttf
NotoSansGurmukhiUI-Regular.ttf
NotoSansHanunoo-Regular.ttf
NotoSansHebrew-Bold.ttf
NotoSansHebrew-Regular.ttf
NotoSansJP-Regular.otf
NotoSansJavanese-Regular.ttf
NotoSansKR-Regular.otf
NotoSansKannada-Bold.ttf
NotoSansKannada-Regular.ttf
NotoSansKannadaUI-Bold.ttf
NotoSansKannadaUI-Regular.ttf
NotoSansKayahLi-Regular.ttf
NotoSansKhmer-Bold.ttf
NotoSansKhmer-Regular.ttf
NotoSansKhmerUI-Bold.ttf
NotoSansKhmerUI-Regular.ttf
NotoSansLao-Bold.ttf
NotoSansLao-Regular.ttf
NotoSansLaoUI-Bold.ttf
NotoSansLaoUI-Regular.ttf
NotoSansLepcha-Regular.ttf
NotoSansLimbu-Regular.ttf
NotoSansLisu-Regular.ttf
NotoSansMalayalam-Bold.ttf
NotoSansMalayalam-Regular.ttf
NotoSansMalayalamUI-Bold.ttf
NotoSansMalayalamUI-Regular.ttf
NotoSansMandaic-Regular.ttf
NotoSansMeeteiMayek-Regular.ttf
NotoSansMongolian-Regular.ttf
NotoSansMyanmar-Bold.ttf
NotoSansMyanmar-Regular.ttf
NotoSansMyanmarUI-Bold.ttf
NotoSansMyanmarUI-Regular.ttf
NotoSansNKo-Regular.ttf
NotoSansNewTaiLue-Regular.ttf
NotoSansOlChiki-Regular.ttf
NotoSansOriya-Bold.ttf
NotoSansOriya-Regular.ttf
NotoSansOriyaUI-Bold.ttf
NotoSansOriyaUI-Regular.ttf
NotoSansRejang-Regular.ttf
NotoSansSC-Regular.otf
NotoSansSaurashtra-Regular.ttf
NotoSansSinhala-Bold.ttf
NotoSansSinhala-Regular.ttf
NotoSansSundanese-Regular.ttf
NotoSansSylotiNagri-Regular.ttf
NotoSansSymbols-Regular-Subsetted.ttf
NotoSansSyriacEstrangela-Regular.ttf
NotoSansTC-Regular.otf
NotoSansTagbanwa-Regular.ttf
NotoSansTaiLe-Regular.ttf
NotoSansTaiTham-Regular.ttf
NotoSansTaiViet-Regular.ttf
NotoSansTamil-Bold.ttf
NotoSansTamil-Regular.ttf
NotoSansTamilUI-Bold.ttf
NotoSansTamilUI-Regular.ttf
NotoSansTelugu-Bold.ttf
NotoSansTelugu-Regular.ttf
NotoSansTeluguUI-Bold.ttf
NotoSansTeluguUI-Regular.ttf
NotoSansThaana-Bold.ttf
NotoSansThaana-Regular.ttf
NotoSansThai-Bold.ttf
NotoSansThai-Regular.ttf
NotoSansThaiUI-Bold.ttf
NotoSansThaiUI-Regular.ttf
NotoSansTibetan-Regular.ttf
NotoSansTifinagh-Regular.ttf
NotoSansVai-Regular.ttf
NotoSansYi-Regular.ttf
NotoSerif-Bold.ttf
NotoSerif-BoldItalic.ttf
NotoSerif-Italic.ttf
NotoSerif-Regular.ttf
Roboto-Black.ttf
Roboto-BlackItalic.ttf
Roboto-Bold.ttf
Roboto-BoldItalic.ttf
Roboto-Italic.ttf
Roboto-Light.ttf
Roboto-LightItalic.ttf
Roboto-Medium.ttf
Roboto-MediumItalic.ttf
Roboto-Regular.ttf
Roboto-Thin.ttf
Roboto-ThinItalic.ttf
RobotoCondensed-Bold.ttf
RobotoCondensed-BoldItalic.ttf
RobotoCondensed-Italic.ttf
RobotoCondensed-Light.ttf
RobotoCondensed-LightItalic.ttf
RobotoCondensed-Regular.ttf
```

Frenzie:
More bad news, I thought Noto was available on 4+, but clearly I was wrong. On 4.1 it's not present:

```
root@android:/system/fonts # ls
AndroidClock.ttf
AndroidClock_Highlight.ttf
AndroidClock_Solid.ttf
AndroidEmoji.ttf
Clockopia.ttf
DroidNaskh-Regular-SystemUI.ttf
DroidNaskh-Regular.ttf
DroidSans-Bold.ttf
DroidSans.ttf
DroidSansArmenian.ttf
DroidSansDevanagari-Regular.ttf
DroidSansEthiopic-Regular.ttf
DroidSansFallback.ttf
DroidSansGeorgian.ttf
DroidSansHebrew-Bold.ttf
DroidSansHebrew-Regular.ttf
DroidSansMono.ttf
DroidSansTamil-Bold.ttf
DroidSansTamil-Regular.ttf
DroidSansThai.ttf
DroidSerif-Bold.ttf
DroidSerif-BoldItalic.ttf
DroidSerif-Italic.ttf
DroidSerif-Regular.ttf
MTLmr3m.ttf
Roboto-Bold.ttf
Roboto-BoldItalic.ttf
Roboto-Italic.ttf
Roboto-Light.ttf
Roboto-LightItalic.ttf
Roboto-Regular.ttf
RobotoCondensed-Bold.ttf
RobotoCondensed-BoldItalic.ttf
RobotoCondensed-Italic.ttf
RobotoCondensed-Regular.ttf
```

NiLuJe:
We could arguably use Roboto instead of Noto sans, and Droid fallback
instead of Noto cjk.

Frenzie:
An interesting point, but I assume Roboto doesn't have quite the range…

After #5252 it'd be as simple as:

```lua
-- horrible variable name :-P
local main_font = android and "roboto" or "noto"
```

NiLuJe:
Droid Fallback should cover a lot/most of it, but, yeah, of course it won't be the same ;).

pazos:
> Just say the word (which format). lzma is nice because it's quick in decompression but pretty much anything goes. Heck, even plain tar would probably do the trick.

Android does support uncompressing zips from the assets store on the fly. But we need to change permissions accordly (I dunno if is possible to keep unix permissions on a zip file).

Anyways, aapt seems to do a very good job with our current assets, so the uncompressing step might be useless (just 3-10mb more vs lzma on release/debug builds). In that case we copy everything, without compressing, in platform/android/luajit-launcher/module and copy assets on the first run, with:

```diff
+    private boolean copyAssets() {
+        AssetManager assetManager = getAssets();
+        String[] files = null;
+
+        try {
+            files = assetManager.list("module");
+        } catch (IOException e) {
+            Log.e(app_name, "Error reading asset files.", e);
+            return false;
+        }
+
+        // fail if 1+ exceptions happen during asset extraction.
+        boolean copy_failed;
+
+        for(String filename : files) {
+            InputStream input = null;
+            OutputStream output = null;
+            String msg = "asset -> " + assets_path + "/" + filename;
+            try {
+                input = assetManager.open(filename);
+                output = new FileOutputStream(new File(assets_path, filename));
+                copyFile(input, output);
+                input.close();
+                output.flush();
+                output.close();
+                Log.d(app_name, msg + " [ok]");
+            } catch(IOException e) {
+                Log.e(app_name, msg + " [error]:\n" + e.toString());
+                copy_failed = true;
+            }
+        }
+        return !copy_failed;
+    }
+
+    private void copyFile(InputStream input, OutputStream output) throws IOException {
+        byte[] buffer = new byte[1024];
+        int read;
+        while((read = input.read(buffer)) != -1){
+           output.write(buffer, 0, read);
+        }
+    }
```
So, in onCreate we do:

``` java
    // copy assets to the local path.
    boolean assets_copied = copyAssets();
    if (!assets_copied) { /* notify error */ }
```

The example might not work as is because I was testing it embedded into MainApp.java, so it is loaded before the Activity but we want to start in from the activity to be able to repurpose the update dialog.

NiLuJe:
zip *should* preserve basic file permissions (i.e., rwx, not sure about uid/gid). Now whether Android honors that when unzipping... I'd assume so, but I don't know ;).

NiLuJe:
It also makes sense that compression wouldn't necessarily help, as the gist of the stuff taking space is either already compressed or doesn't compress well (fonts, icons).

An archive might simply be easier to handle than recursive copies, so, if zip works, an uncompressed zip might be a great way to handle that?

pazos:
> An archive might simply be easier to handle than recursive copies, so, if zip works, an uncompressed zip (-Z store or -0) might be a great way to handle that?

The problem is: aapt won't try to compress a zip file even when the file itself is uncompressed. It does try to compress things directly and does a pretty good job with text files. So the zip file, if uncompressed, will take more space on the apk and take more time to "uncompress" that the raw assets.

I will try to code something for tomorrow and have timmings for both alternatives.

pazos:
Sooo, I started moving the libraries outside the 7z package. I did everything documentation says:

## luajit-launcher diff

```diff
diff --git a/Makefile b/Makefile
index 19d1fc8..2729904 100644
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ test: update
        @echo "WARNING: You'll need to sign this application to be able to install it"
 clean:
        @echo "Cleaning binaries, assets and LuaJIT build"
-       rm -rf assets/module/ bin/ jni/luajit/build
+       rm -rf assets/module bin/ libs/ jni/luajit/build
        cd jni/luajit && \
                ./mk-luajit.sh clean
 
diff --git a/app/AndroidManifest.xml b/app/AndroidManifest.xml
index eb39428..ec0f804 100644
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -19,6 +19,7 @@
         android:vmSafeMode="true"
         android:allowBackup="false"
         android:fullBackupContent="false"
+        android:extractNativeLibs="true"
         android:usesCleartextTraffic="true">
         <meta-data android:name="android.max_aspect" android:value="3.1" />
         <activity
diff --git a/app/build.gradle b/app/build.gradle
index 77b2902..ce9c033 100644
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -34,6 +34,7 @@ android {
             java.srcDirs = ['src']
             res.srcDirs = ['res']
             assets.srcDirs = [ '../assets' ]
+            jniLibs.srcDirs = [ '../libs' ]
         }
     }
```

## koreader diff

```diff
diff --git a/Makefile b/Makefile
index 7ddffaae..2c154687 100644
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,13 @@ else ifeq ($(TARGET), pocketbook)
 endif
 
 MACHINE=$(shell PATH=$(PATH) $(CC) -dumpmachine 2>/dev/null)
+
+ifeq ($(MACHINE),i686-linux-android)
+       ANDROID_ABI:=x86
+else
+       ANDROID_ABI:=armeabi-v7a
+endif
+
 ifdef KODEBUG
        MACHINE:=$(MACHINE)-debug
 endif
@@ -334,15 +341,31 @@ endif
                mv *.AppImage ../../koreader-$(DIST)-$(MACHINE)-$(VERSION).AppImage
 
 androidupdate: all
+       # folder with application assets
        mkdir -p $(ANDROID_LAUNCHER_DIR)/assets/module
        -rm $(ANDROID_LAUNCHER_DIR)/assets/module/koreader-*
+       -rm -rf $(ANDROID_LAUNCHER_DIR)/libs
+
        # in runtime luajit-launcher's libluajit.so will be loaded
        -rm $(INSTALL_DIR)/koreader/libs/libluajit.so
-       # make android update apk
+
+       # we compress assets manually with 7z
        cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=1 \
                ../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).7z * \
-               -x!resources/fonts -x!resources/icons/src -x!spec
-       $(MAKE) -C $(ANDROID_LAUNCHER_DIR) $(if $(KODEBUG), debug, release) ANDROID_APPNAME=KOReader ANDROID_VERSION=$(ANDROID_VERSION) ANDROID_NAME=$(ANDROID_NAME) ANDROID_FLAVOR=$(ANDROID_FLAVOR)
+               -x!libs \
+               -x!resources/fonts \
+               -x!resources/icons/src \
+               -x!spec
+
+       # compression for libraries is handled by aapt
+       mkdir -p $(ANDROID_LAUNCHER_DIR)/libs
+       cp -pRv $(INSTALL_DIR)/koreader/libs \
+               $(ANDROID_LAUNCHER_DIR)/libs/$(ANDROID_ABI)
+
+       # build the apk.
+       $(MAKE) -C $(ANDROID_LAUNCHER_DIR) $(if $(KODEBUG), debug, release) \
+               ANDROID_APPNAME=KOReader ANDROID_VERSION=$(ANDROID_VERSION) \
+               ANDROID_NAME=$(ANDROID_NAME) ANDROID_FLAVOR=$(ANDROID_FLAVOR)
        cp $(ANDROID_LAUNCHER_DIR)/bin/NativeActivity.apk \
                koreader-android-$(MACHINE)-$(VERSION).apk

```

Reduced launch time from 7 seconds to 2 seconds on the beefy emulator while keeping package size unaltered. Native libraries get copied in the expected path:

```adb
 adb shell ls /data/app-lib/org.koreader.launcher-1
libblitbuffer.so
libcrengine.so
libkoreader-cre.so
libkoreader-djvu.so
libkoreader-lfs.so
liblodepng.so
libluacompat52.so
libluajit.so
liblualongnumber.so
liblzma.so
libmupdf.so
libsqlite3.so
libturbojpeg.so
libwrap-mupdf.so
```

# But
```adb logcat
I/KOReader( 3683): trying to open asset libs/libkoreader-lfs.lua: cdata<struct AAsset *>: NULL
E/KOReader( 3683): [NativeThread]: failed to run script: datastorage.lua:3: module 'libs/libkoreader-lfs' not found:
E/KOReader( 3683): 	no field package.preload['libs/libkoreader-lfs']
E/KOReader( 3683): 	no file 'libs/libkoreader-lfs.lua' (checked with asset loader)
E/KOReader( 3683): 	no file 'libs/libkoreader-lfs.lua'
E/KOReader( 3683): 	no file '/data/data/org.koreader.launcher/files/libs/libkoreader-lfs.lua'
E/KOReader( 3683): 	no file 'libs/libkoreader-lfs.so'
E/KOReader( 3683): 	no file '/data/data/org.koreader.launcher/files/libs/libkoreader-lfs.so'
E/KOReader( 3683): [NativeThread]: Stopping due to previous errors
```

It seems to me that we are looking inside NativeLibrary but the package expected libraries on NativeLibrary/libs. I have no idea what to do next. Any hint @Frenzie ??

pazos:
> It seems to me that we are looking inside NativeLibrary but the package expected libraries on NativeLibrary/libs

in fact:

```diff
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -2021,7 +2021,7 @@ local function run(android_app_state)
     -- set up a sensible package.path
     package.path = "?.lua;"..android.dir.."/?.lua;"
     -- set absolute cpath
-    package.cpath = "?.so;"..android.dir.."/?.so;"
+    package.cpath = "?.so;"..android.dir.."/?.so;"..android.nativeLibraryDir.."/?.so;"
     -- register the asset loader
     table.insert(package.loaders, 2, android.asset_loader)
```

```adb logcat
E/KOReader( 4704): [NativeThread]: failed to run script: datastorage.lua:3: module 'libs/libkoreader-lfs' not found:
E/KOReader( 4704): 	no field package.preload['libs/libkoreader-lfs']
E/KOReader( 4704): 	no file 'libs/libkoreader-lfs.lua' (checked with asset loader)
E/KOReader( 4704): 	no file 'libs/libkoreader-lfs.lua'
E/KOReader( 4704): 	no file '/data/data/org.koreader.launcher/files/libs/libkoreader-lfs.lua'
E/KOReader( 4704): 	no file 'libs/libkoreader-lfs.so'
E/KOReader( 4704): 	no file '/data/data/org.koreader.launcher/files/libs/libkoreader-lfs.so'
E/KOReader( 4704): 	no file '/data/app-lib/org.koreader.launcher-1/libs/libkoreader-lfs.so'
E/KOReader( 4704): [NativeThread]: Stopping due to previous errors
```

I guess it can be fixed easily but I'm clueless so I'm going to work on the zip extractor code while waiting for help :+1:

Frenzie:
`libs/` is hardcoded in every ffi.load call.

So for now I'd just add a couple of lines to this file:
https://github.com/koreader/koreader/blob/f3c3989b7fa652b01a8ab45ce01e49ffacd587d6/setupkoenv.lua#L18-L32

```lua
-- on top
local is_android, android = pcall(require, "android")

-- and on line 19, something like
if is_android and lib:match("^lib/") then lib = lib:sub(lib, 4) end
```

pazos:
@Frenzie: it doesn't work. But if I do the same on assets/android

```diff
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1274,8 +1274,11 @@ function android.deplib_loader(modulename)
     end
 
     local modulepath = string.gsub(modulename, "%.", "/")
+    if modulepath:match("^libs/") then modulepath = string.sub(modulepath, 6) end
+    android.LOGI(modulepath)
     for path in string.gmatch(package.cpath, "([^;]+)") do
         local module = string.gsub(path, "%?", modulepath)
+        android.LOGI(module)
         -- try to load dependencies of this module with our dlopen implementation

```

I get:

```adb logcat
I/KOReader( 7174): libkoreader-lfs
I/KOReader( 7174): /data/app-lib/org.koreader.launcher-1/libkoreader-lfs.so
E/KOReader( 7174): error: [string "dl.lua"]:99: could not find library /data/app-lib/org.koreader.launcher-1/libkoreader-lfs.so
E/KOReader( 7174): [NativeThread]: failed to run script: datastorage.lua:3: module 'libs/libkoreader-lfs' not found:
E/KOReader( 7174): 	no field package.preload['libs/libkoreader-lfs']
E/KOReader( 7174): 	no file 'libs/libkoreader-lfs.lua' (checked with asset loader)
E/KOReader( 7174): 	no file 'libs/libkoreader-lfs.lua'
E/KOReader( 7174): 	no file '/data/data/org.koreader.launcher/files/libs/libkoreader-lfs.lua'
E/KOReader( 7174): 	no file '/data/app-lib/org.koreader.launcher-1/libs/libkoreader-lfs.so'
E/KOReader( 7174): [NativeThread]: Stopping due to previous errors
```

Frenzie:
You'd have to do some debug logging in the library functions; it could be helpful to put the relevant code up in branches others can fetch.

Since we don't see a `try to load module` that probably the readable check returned false. Something like this:

```diff
diff --git a/assets/android.lua b/assets/android.lua
index 29693bf..0620c0e 100644
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1279,8 +1279,11 @@ This loader function just loads dependency libraries for the C module.
 --]]
 function android.deplib_loader(modulename)
     local function readable(filename)
-        local f = io.open(filename, "r")
-        if f == nil then return false end
+        local f, err = io.open(filename, "r")
+        if f == nil then
+            android.DEBUG("can't open module "..module)
+            return false
+        end
         f:close()
         return true
     end
```

pazos:
Ok, will try tomorrow. Now I'm fully into the zip uncompressor and it is showing good results (zip with max compression takes 5% more apk size and 70% less time uncompressing). I'm not able to reproduce the issues that happened with lzma.

Frenzie:
We don't necessarily need max compression though, unless it's fine CPU/memory wise on the decompression end.

Funny though, you're saying that minimum compression lzma is smaller than maximum compression zip? :rofl:

pazos:
> We don't necessarily need max compression though, unless it's fine CPU/memory wise on the decompression end.

I just finished the first implementation but seems to work fine in both CPU/memory/VM heap limited machines.

> Funny though, you're saying that minimum compression lzma is smaller than maximum compression zip?

Yes, 2mb on the release build for x86. But is faster uncompressing.

You can give it a try: https://github.com/koreader/android-luajit-launcher/pull/174

Frenzie:
Closing in favor of <https://github.com/koreader/android-luajit-launcher/pull/174>.

Issue #5021, opened 2019-05-10
Feature request: Option for shutdown in the Gesture Manager


noembryo:
Just a way to avoid a click or two by long pressing a corner :-p

I never put my Glo to sleep, because if I do, it spends more battery than when is open.
So I have to shutdown every time I leave my book.
And I'm lazy...

On a side-note, would it not be cleaner if there was a menu arrangement like this?
```
Gesture Manager
    Top left corner
        Tap
            menu
        Hold
            menu
    Top right corner
        Tap
            menu
        Hold
            menu
    Bottom left corner
        Tap
            menu
        Hold
            menu
    Bottom right corner
        Tap
            menu
        Hold
            menu
```
Anyway, all these are minor nitpicks.
You guys have made KOReader a super reader!

Too bad there is no windows version ;o)

noembryo:
I meant "Hold" not "Right top" ..

Frenzie:
> Too bad there is no windows version ;o)

I'm told the program runs fine in WSL.

noembryo:
What is WSL (apart from that [WSL](https://www.worldsurfleague.com/))

Frenzie:
https://www.howtogeek.com/261575/how-to-run-graphical-linux-desktop-applications-from-windows-10s-bash-shell/

More is on the way: https://venturebeat.com/2019/05/06/microsoft-windows-terminal-wsl-2-coming-june/

robert00s:
Should we reorganize Gesture manager menu or leave as it is? I like arrangement as is now.

Frenzie:
I think it's clearer to have the tap menu with locations, the hold menu, etc. (like we do). To focus on the location first rather than the action strikes me as backwards, but I'm having trouble articulating why.

poire-z:
> I think it's clearer to have the tap menu with locations (like we do)

I agree,
Couldn't we just move the 4 `Tap y x corner` in a `Tap corner >` submenu, like all the others?
It woule make the main menu one single page.
And the `Tap y x corner` ones are the only ones not (yet) showing the associated action in parenthesis in the menu text. Adding that while they are in the main menu would make that main menu a bit ugly, but no pb if they are moved in a submenu.

noembryo:
@Frenzie So, is the "Shut down reader" added to the gesture manager options?

Frenzie:
It was added a few days ago in #5202.

noembryo:
Oops, I missed it... :o)
Thanks!

PR #5202, opened 2019-08-13
Add gesture - poweroff and restart device

robert00s:
Ref: #5021 

Two new gesture:
+ Power off
+ Reboot the device

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -58,6 +58,8 @@ local action_strings = {
     suspend = _("Suspend"),
     exit = _("Exit KOReader"),
     restart = _("Restart KOReader"),
+    poweroff = _("Power off"),
+    reboot = _("Reboot the device"),
     show_menu = _("Show menu"),
     show_config_menu = _("Show bottom menu"),
     show_frontlight_dialog = _("Show frontlight dialog"),
@@ -686,6 +688,8 @@ function ReaderGesture:buildMenu(ges, default)
         {"suspend", true},
         {"exit", true},
         {"restart", not Device:isAndroid()},
+        {"poweroff", Device:isCervantes() or Device:isKobo() or Device:isSonyPRSTUX()},
+        {"reboot", Device:isCervantes() or Device:isKobo() or Device:isSonyPRSTUX()},
 
         {"show_menu", true},
         {"show_config_menu", not self.is_docless, true},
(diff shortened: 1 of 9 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -686,6 +688,8 @@ function ReaderGesture:buildMenu(ges, default)
         {"suspend", true},
         {"exit", true},
         {"restart", not Device:isAndroid()},
+        {"poweroff", Device:isCervantes() or Device:isKobo() or Device:isSonyPRSTUX()},
```
Frenzie: With this being duplicated twice here and also in https://github.com/koreader/koreader/blob/master/frontend/ui/elements/common_info_menu_table.lua#L76 maybe it should be abstracted into a Device:canPowerOff() or some such?

robert00s: Good tip. Thanks.

NiLuJe:
Sure, it's been used in a couple of places by now ;).

poire-z:
Lgtm.
I'm fine with the abstraction, but we may as well add `canRestart`, which Android can't :)
` {"restart", not Device:isAndroid()},` just above the added ones.

(May be in the menu order, PowerOff after Reboot, ordered by _bye_'ness :) ?)

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1415,6 +1419,22 @@ function ReaderGesture:gestureAction(action, ges)
...
         self.ui.menu:exitOrRestart(function() UIManager:restartKOReader() end)
+    elseif action == "poweroff" then
+        UIManager:show(ConfirmBox:new{
+            text = _("Are you sure you want to power off the device?"),
+            ok_text = _("Power off"),
+            ok_callback = function()
+                UIManager:nextTick(UIManager.poweroff_action)
+            end,
+        })
+    elseif action == "reboot" then
```
Frenzie: Bit of a nitpick that I didn't think of before @poire-z mentioned it, but the reboot/poweroff order also happens here

frontend/device/cervantes/device.lua
```diff
@@ -41,6 +41,8 @@ local Cervantes = Generic:new{
     hasOTAUpdates = yes,
     hasKeys = yes,
     hasWifiManager = yes,
+    canPowerOff = yes,
```
Frenzie: and here (and in the other device files)

robert00s: Done.

poire-z:
> but we may as well add canRestart, which Android can't :)
{"restart", not Device:isAndroid()}, => {"restart", Device:canRestart()}, just above the added ones.

Can you add that one in this PR, for coherency/completeness sake?

robert00s:
Yes, soon.

poire-z:
And some last bits there, and we're done, thanks :)
https://github.com/koreader/koreader/blob/aae6acb48c0b32f1f4898e37441cf7133c12b97b/frontend/apps/reader/modules/readermenu.lua#L211-L215
https://github.com/koreader/koreader/blob/aae6acb48c0b32f1f4898e37441cf7133c12b97b/frontend/apps/filemanager/filemanagermenu.lua#L440-L444

Issue #4854, opened 2019-03-30
Suggestion:read percentage up to two digits after the decimal point.


seacatwu:
chage the file
koreader/frontend/apps/reader/modules/readerfooter.lua

`    percentage = function(footer)
        return ("R:%1.f%%"):format(footer.progress_bar.percentage * 100)
    end,
`
to
`    percentage = function(footer)
        return ("R:%.2f%%"):format(footer.progress_bar.percentage * 100)
    end,
`

PR #3078, opened 2017-08-12
WIP: Use background runner to restore wifi and check network state

Hzj-jie:
This change should significantly improve the responsiveness of network menu. I.e. we do not need to ping the www.example.com when menu pops up.

I am still verifying this change on my device.

Issue #2706, opened 2017-04-01
WiFi and network manager states


lgeek:
Hi,

I think what the UI shows doesn't quite match the state of the actual interface, and the internal state of the network manager gets similarly confused. At the moment, there are two states shown for the WiFi connection: on (connected) and off. However, the WiFi adapter has 3 main states: on (up, disconnected), connected, and off (down). 

For example, we can now enable the Wi-Fi adapter via Gear -> Network -> Wi-Fi connection. However, if we don't connect to a network, when we return to the menu, the 'WiFi connection' checkmark will appear as off. In fact, the WiFi adapter is left up in the disconnected state. If we select the 'Wi-Fi connection' button again, then we'll once again be asked whether to turn on Wi-Fi, even though it is already enabled. Similarly, we can only turn off the WiFi adapter after we have connected to a network, there's no way to go from the disconnected state to the off state. With the 'automatically restore Wi-Fi connection after resume' option enabled, this can persist for the uptime of the device.

The way I would propose to fix this is exposing the three states in the UI. The following transitions should be allowed:

* off -> on
* on -> connected
* on -> off
* connected -> off
* connected -> on (via the disconnect button in the network list)

In terms of actual changes to the UI, I would suggest the following:

* renaming the 'Wi-Fi connection' menu entry to 'Wi-Fi' and mapping the check mark to (on || connected)
* changing the actions of the WiFi menu entry to the following: from off, ask whether to enable WiFi, then show the network list if answering yes; from on and connected, show the network list immediately
* from the on and connected states, is must be allowed to turn off Wi-Fi. I'm not sure how to implement that in terms of UI. I'd be tempted to use a long press on the Wi-Fi connection menu, followed by the confirmation dialog currently used. However, I'm not sure if that is consistent with the rest of the UI. In particular, I'm concerned that it might not be intuitive to discover this action. Maybe it would make more sense to add a turn off button to the network list UI.

Any thoughts?

Frenzie:
My preferred way of disabling wifi is by suspending quickly because it's not really any slower than going through the GUI anyway. I don't use restore on resume. Probably a case of https://xkcd.com/1172/

Another possible solution (and probably the easiest short-term) would be to change the network info popup into a dialog with a "Turn off Wi-Fi" button.

houqp:
> renaming the 'Wi-Fi connection' menu entry to 'Wi-Fi' and mapping the check mark to (on || connected)

[...]

PR #5203, opened 2019-08-13
Some footer tweaks

robert00s:
Base on #4594
Close: #4854 

Changes include:
+ adding unicode characters to menu and footer
+ move alignment to settings submenu
+ new text footer separator
+ progress percentage format (#4854)
+ date in 12/24 format
+ time format in 1h30' as option

Edit - new screenshots

obraz

obraz

obraz

obraz

obraz

obraz

obraz

obraz

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -38,7 +38,7 @@ local symbol_prefix = {
         time = nil,
         pages_left = "=>",
         battery = "B:",
-        percentage = nil,
+        percentage = "R:",
         book_time_to_read = "TB:",
         chapter_time_to_read = "TC:",
         frontlight = "L:",
@@ -49,7 +49,7 @@ local symbol_prefix = {
         time = "⌚",
         pages_left = "⇒",
         battery = "⚡",
-        percentage = "◔",
+        percentage = "⤠",
         book_time_to_read = "⏳",
         chapter_time_to_read = "⤻",
         frontlight = "☼",
@@ -120,7 +120,7 @@ local footerTextGeneratorMap = {
     percentage = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].percentage
-        local digits = footer.settings.progress_perc_format or "0"
+        local digits = footer.settings.progress_pct_format or "0"
         local string_percentage
         if not prefix then
             string_percentage = "%." .. digits .. "f%%"
@@ -541,8 +541,32 @@ function ReaderFooter:addToMainMenu(menu_items)
     table.insert(sub_items, {
         text = _("Settings"),
         sub_item_table = {
+            getMinibarOption("all_at_once", self.updateFooterTextGenerator),
+            getMinibarOption("reclaim_height"),
+            {
+                text = _("Auto refresh time"),
+                separator = true,
+                checked_func = function()
+                    return self.settings.auto_refresh_time == true
+                end,
+                -- only enable auto refresh when time is shown
+                enabled_func = function() return self.settings.time end,
+                callback = function()
+                    self.settings.auto_refresh_time = not self.settings.auto_refresh_time
+                    G_reader_settings:saveSetting("footer", self.settings)
+                    if self.settings.auto_refresh_time then
+                        self:setupAutoRefreshTime()
+                    else
+                        UIManager:unschedule(self.autoRefreshTime)
+                        self.onCloseDocument = nil
+                    end
+                end
+            },
             {
                 text = _("Alignment"),
+                enabled_func = function()
+                    return self.settings.disable_progress_bar
+                end,
                 sub_item_table = {
                     {
                         text = _("Center"),
@@ -586,7 +610,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                 }
             },
             {
-                text = _("Text prefix"),
+                text = _("Prefix"),
                 sub_item_table = {
                     {
                         text = _("Icons"),
@@ -613,15 +637,15 @@ function ReaderFooter:addToMainMenu(menu_items)
                 },
             },
             {
-                text = _("Text separator"),
+                text = _("Separator"),
                 sub_item_table = {
                     {
                         text = _("Vertical line") .. " (|)",
                         checked_func = function()
-                            return self.settings.text_separator == "bar" or self.settings.text_separator == nil
+                            return self.settings.items_separator == "bar" or self.settings.items_separator == nil
                         end,
                         callback = function()
-                            self.settings.text_separator = "bar"
+                            self.settings.items_separator = "bar"
                             self:updateFooter()
                             UIManager:setDirty(nil, "ui")
                         end,
@@ -629,21 +653,21 @@ function ReaderFooter:addToMainMenu(menu_items)
                     {
                         text = _("Bullet") .. " (•)",
                         checked_func = function()
-                            return self.settings.text_separator == "bullet"
+                            return self.settings.items_separator == "bullet"
                         end,
                         callback = function()
-                            self.settings.text_separator = "bullet"
+                            self.settings.items_separator = "bullet"
                             self:updateFooter()
                             UIManager:setDirty(nil, "ui")
                         end,
                     },
                     {
-                        text = _("Without separator"),
+                        text = _("No separator"),
                         checked_func = function()
-                            return self.settings.text_separator == "none"
+                            return self.settings.items_separator == "none"
                         end,
                         callback = function()
-                            self.settings.text_separator = "none"
+                            self.settings.items_separator = "none"
                             self:updateFooter()
                             UIManager:setDirty(nil, "ui")
                         end,
@@ -744,12 +768,9 @@ function ReaderFooter:addToMainMenu(menu_items)
             },
         }
     })
-    table.insert(sub_items,
-                 getMinibarOption("all_at_once", self.updateFooterTextGenerator))
-    table.insert(sub_items,
-                 getMinibarOption("reclaim_height"))
     table.insert(sub_items, {
         text = _("Progress bar"),
+        separator = true,
         sub_item_table = {
             {
                 text = _("Show progress bar"),
@@ -764,24 +785,6 @@ function ReaderFooter:addToMainMenu(menu_items)
             },
             getMinibarOption("toc_markers", self.setTocMarkers),
         }
-    })
-    table.insert(sub_items, {
-        text = _("Auto refresh time"),
-        checked_func = function()
-            return self.settings.auto_refresh_time == true
-        end,
-        -- only enable auto refresh when time is shown
-        enabled_func = function() return self.settings.time end,
-        callback = function()
-            self.settings.auto_refresh_time = not self.settings.auto_refresh_time
-            G_reader_settings:saveSetting("footer", self.settings)
-            if self.settings.auto_refresh_time then
-                self:setupAutoRefreshTime()
-            else
-                UIManager:unschedule(self.autoRefreshTime)
-                self.onCloseDocument = nil
-            end
-        end
     })
     table.insert(sub_items, getMinibarOption("page_progress"))
     table.insert(sub_items, getMinibarOption("time"))
@@ -805,9 +808,9 @@ function ReaderFooter:genFooterText() end
 function ReaderFooter:genAllFooterText()
     local info = {}
     local separator = "  "
-    if self.settings.text_separator == "bar" or self.settings.text_separator == nil then
+    if self.settings.items_separator == "bar" or self.settings.items_separator == nil then
         separator = " | "
-    elseif self.settings.text_separator == "bullet" then
+    elseif self.settings.items_separator == "bullet" then
         separator = " • "
     end
     for _, gen in ipairs(self.footerTextGenerators) do
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -698,10 +698,10 @@ function ReaderFooter:addToMainMenu(menu_items)
                     {
                         text = _("Bullet") .. " (•)",
                         checked_func = function()
-                            return self.settings.text_separator == "dot"
+                            return self.settings.text_separator == "bullet"
                         end,
                         callback = function()
-                            self.settings.text_separator = "dot"
+                            self.settings.text_separator = "bullet"
                             self:updateFooter()
                             UIManager:setDirty(nil, "ui")
                         end,
@@ -821,9 +821,9 @@ function ReaderFooter:genFooterText() end
 function ReaderFooter:genAllFooterText()
     local info = {}
     local separator = "  "
-    if self.settings.text_separator == "vertiacal_line" or self.settings.text_separator == nil then
+    if self.settings.text_separator == "bar" or self.settings.text_separator == nil then
         separator = " | "
-    elseif self.settings.text_separator == "dot" then
+    elseif self.settings.text_separator == "bullet" then
         separator = " • "
     end
     for _, gen in ipairs(self.footerTextGenerators) do
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -33,6 +33,31 @@ local MODE = {
     wifi_status = 10,
 }
 
+local symbol_prefix = {
+    letters = {
+        time = nil,
+        pages_left = "=>",
+        battery = "B:",
+        percentage = nil,
+        book_time_to_read = "TB:",
+        chapter_time_to_read = "TC:",
+        frontlight = "L:",
+        mem_usage = "M:",
+        wifi_status = "W:",
+    },
+    icons = {
+        time = "🕒",
+        pages_left = "⇒",
+        battery = "⚡",
+        percentage = "◔",
+        book_time_to_read = "⏳",
+        chapter_time_to_read = "⤻",
+        frontlight = "☼",
+        mem_usage = "⌨",
+        wifi_status = "⚟",
+    }
+}
+
 local MODE_NB = 0
 local MODE_INDEX = {}
 for k,v in pairs(MODE) do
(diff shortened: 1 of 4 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -100,16 +109,16 @@ local footerTextGeneratorMap = {
...
         return ""
     end,
     wifi_status = function()
         local NetworkMgr = require("ui/network/manager")
         if NetworkMgr:isWifiOn() then
-            return "W:On"
+            return "⚟ On"
         else
-            return "W:Off"
+            return "⚟ Off"
```
Frenzie: translation

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -363,18 +372,23 @@ local option_titles = {
...
-    time = _("Current time"),
-    pages_left = _("Pages left in chapter"),
-    battery = _("Battery status"),
-    percentage = _("Progress percentage"),
-    book_time_to_read = _("Book time to read"),
-    chapter_time_to_read = _("Chapter time to read"),
-    frontlight = _("Frontlight level"),
-    mem_usage = _("KOReader memory usage"),
-    wifi_status = _("Wi-Fi status"),
+    page_progress = _("Current page") .. " (/)",
```
Frenzie: Why concatenation? This way a translator can't reverse the order if desired. The string will show up as a 95 % match suggestion so minor changes aren't generally a big problem.

frontend/util.lua
```diff
@@ -99,6 +99,53 @@ function util.secondsToClock(seconds, withoutSeconds)
...
+        if mins == "60" then
+            mins = string.format("%02.f", 0)
+            hours = string.format("%.f", hours + 1)
+        end
+        if withoutSeconds then
+            if hours == "0" then
+                mins = string.format("%.f", round(seconds / 60))
+                return mins .. "'"
+            end
+            return hours .. "h" .. mins .. "'"
```
Frenzie: It's a bit of a weird string to translate sure, but in Dutch for example it'd be `1u30`, not `1h30`.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -43,24 +44,30 @@ end
 local footerTextGeneratorMap = {
     empty = function() return "" end,
     frontlight = function()
-        if not Device:hasFrontlight() then return "L: NA" end
+        if not Device:hasFrontlight() then return "☼ NA" end
```
Frenzie: Huh, it seems odd that this is available as an option at all if `not Device:hasFrontlight()`.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -511,46 +525,181 @@ function ReaderFooter:addToMainMenu(menu_items)
...
-                    self:resetLayout(true)
-                    self:updateFooter()
-                    UIManager:setDirty(nil, "ui")
-                end,
+                text = _("Text separator"),
+                sub_item_table = {
+                    {
+                        text = _("Vertical line") .. " (|)",
+                        checked_func = function()
+                            return self.settings.text_separator == "vertiacal_line" or self.settings.text_separator == nil
```
poire-z: verti**a**cal_line => `vertical_line` - or just `"bar"` ?
and `"none"` below instead of `"no_separator"`, `text_separator == "no_separator"` feels a bit redundant :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -511,46 +525,181 @@ function ReaderFooter:addToMainMenu(menu_items)
...
-                    self:resetLayout(true)
-                    self:updateFooter()
-                    UIManager:setDirty(nil, "ui")
-                end,
+                text = _("Progress percentage format"),
+                sub_item_table = {
+                    {
+                        text = T(_("No decimal point (%1)"), self:progressPercentage(0)),
+                        checked_func = function()
+                            return self.settings.progress_perc_format == "0" or self.settings.progress_perc_format == nil
```
poire-z: `progress_perc_format ` => `progress_pct_format` ? "pct" feels more natural than "perc" to me.
Or progress_pct_digits to be more explicite. (What's the word for digit *after* the dot? fractional digits?)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -511,46 +525,181 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                        callback = function()
+                            self.settings.time_format = "classic"
+                            self:updateFooter()
+                            UIManager:setDirty(nil, "ui")
+                        end,
+                    },
+                    {
+                        text = _("Modern"),
+                        checked_func = function()
+                            return self.settings.time_format == "modern"
```
poire-z: Bit confusing `Date format` (above) vs `Time format` (here).
Your Date format (12/24) indeed sets the formating of the time.
Your Time format (classic/modern) actually set the style of **durations** (12:04 vs 12m4s).

frontend/util.lua
```diff
@@ -99,6 +99,53 @@ function util.secondsToClock(seconds, withoutSeconds)
...
+        end
+        local secs = string.format("%02.f", math.floor(seconds - hours * 3600 - mins * 60))
+        if hours == "0" then
+            mins = string.format("%.f", round(seconds / 60))
+            return mins .. "'" .. secs .. "''"
+        end
+        return hours .. "h" .. mins .. "'" .. secs .. "''"
```
poire-z: Mhh, in french, I'm not used to read minutes abbreviated with `'` :) A simple `m` and `s` would talk to me more.
Or may be just nothing when withoutSeconds=true: `1h09` instead of `1h09'` or `1h09m`

Frenzie: Agreed for English,and it should also be a localizable string.

frontend/apps/reader/modules/readerfooter.lua
Frenzie: Are you talking about decimals? :-P

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: Technically called a `pipe` ;).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -511,46 +525,181 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            return self.settings.text_separator == "vertiacal_line" or self.settings.text_separator == nil
+                        end,
+                        callback = function()
+                            self.settings.text_separator = "vertiacal_line"
+                            self:updateFooter()
+                            UIManager:setDirty(nil, "ui")
+                        end,
+                    },
+                    {
+                        text = _("Dot") .. " (•)",
```
NiLuJe: `Bullet`? (or `Bullet point`)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -100,16 +109,16 @@ local footerTextGeneratorMap = {
             local rss = infos:match("^%S+ (%S+) ")
             -- we got the nb of 4Kb-pages used, that we convert to Mb
             rss = math.floor(tonumber(rss) * 4096 / 1024 / 1024)
-            return ("M:%d"):format(rss)
+            return ("⌨ %d"):format(rss)
```
NiLuJe: That's a keyboard, right? Do we have a chip-like glyph available instead? That vaguely sounds more appropriate her,e but I'm nitpicking ;).

NiLuJe: A quick look at FreeSerif and the Unicode listing would point to "no". Unless "tape drive" (✇ U+2707) counts ;p

poire-z:
(Looking nice on my device, the icons are not distracting.)

frontend/apps/reader/modules/readerfooter.lua
poire-z: Or use these ones, swithing from the non-distracting white one to the heavily distracting black one when memory reaches some level:
image

At 200M on my Kobo, things begin to get slow - but I don't know how to decide that triggering limit scientifically :)

frontend/util.lua
poire-z: But we need something when the minute when zero hour.
`⤻ 9'` is somehow ok,
`⤻ 9m` feels a bit strange (could feel like Mb)
`⤻ 9` without anything is just confusing.
`⤻ 1h09` is ok.

frontend/apps/reader/modules/readerfooter.lua
Frenzie: > At 200M on my Kobo, things begin to get slow - but I don't know how to decide that triggering limit scientifically :)

You won't reach that much memory unless it's a large and/or complex document, so that's probably expected for page turning responsiveness, but not so much for menu responsiveness.

You can do it quote unquote scientifically by simply filling up a couple 100 MB on the side. An interesting suggestion that might work on busybox (depending on how much Bash syntax it supports):

```sh
function malloc() {
  if [[ $# -eq 0 || $1 -eq '-h' || $1 -lt 0 ]] ; then
    echo -e "usage: malloc N\n\nAllocate N mb, wait, then release it."
  else 
    N=$(free -m | grep Mem: | awk '{print int($2/10)}')
    if [[ $N -gt $1 ]] ;then 
      N=$1
    fi
    sh -c "MEMBLOB=\$(dd if=/dev/urandom bs=1MB count=$N) ; sleep 1"
  fi
}
```
From https://unix.stackexchange.com/a/99390

poire-z: > You won't reach that much memory unless it's a large and/or complex document

We reach it very easily when using dict or wikipedia lookups. The TextBoxWidget code uses a lot of memory for the various Lua tables holding chars/widths/justification.
Do that, switch EPUB documents, do that again, and the memory might end up very fragmented and hardly re-used, and the usage rises.
At some point, using a gesture to show the History or TOC or top menu gets from instant to 1 or 2 seconds delay.

Frenzie: Does the problem go away after a little while (i.e., post garbage collection)? Perhaps it could be of interest to trigger garbage collection as the final step in closing a document. On the flipside, it'd probably make it slower to reopen.

poire-z:
Unrelated, just for information, in case other feels it makes more sense than the `=> 8` (which says there are 8 pages left in this chapter, and it becomes 0 when on the last page), I have it modified to show `8/8` ... `1/8` , reverse number of page in this chapter / total number of pages in this chapter.
(I was always mentally computing +1 to `=> 8` for some reason...)
Or this could become another distinct foooter item.

```diff
     pages_left = function(footer)
         local left = footer.ui.toc:getChapterPagesLeft(
             footer.pageno, footer.toc_level)
-        return "=> " .. (left and left or footer.pages - footer.pageno)
+        -- return "=> " .. (left and left or footer.pages - footer.pageno)
+        -- so we see the total nb of page of this chapter
+        local done = footer.ui.toc:getChapterPagesDone(
+            footer.pageno, footer.toc_level)
+        left = left or (footer.pages - footer.pageno)
+        done = done or footer.pageno-1
+        -- return left.." > "..(left+done+1)
+        return (left+1).." > "..(left+done+1)
     end,
```

I had also added Wifi status with an icon to the percentage item - but I just realized we have a Wifi status footer item, which is enabled only on Android. Why is that? It's quite valuable on other devices to not forget to dismiss Wifi to save battery when done looking up Wikipedia. (I'm appending an eye catching enough `¶ WIFI` for that.)

frontend/apps/reader/modules/readerfooter.lua
poire-z: Nope. I may just sometimes get -1 or -2 Mb when switching to another EPUB document, after the next page turn.

> trigger garbage collection as the final step in closing a document

We already do that:
https://github.com/koreader/koreader/blob/0b8d15416a3ebadee2b26a9d41af9c2047836d6e/frontend/document/documentregistry.lua#L146-L151

(It's quite possible the previous document is not yet closed at this point, as crengine may still delay the opening of the new one by saving the cache of the previous one, but switching from a big document, to a small one, and then between 2 small docs does not reclaim the large memory used by the first one.)

collectgarbage() does not help much with the memory usage rises I witness, I've already tried to set it at various places (and let some calls in the code just to be sure to get back what we can... which is not much).

Frenzie:
On Android it's asynchronously updated and returns a value immediately, elsewhere it's very slow. See, for instance, <https://github.com/koreader/koreader/issues/2706#issuecomment-291529583>.

poire-z:
Oh right. I forgot I have other related tweaks, which make it usable.
(In my patch, I also have some `isWifiEnabled()` just checking if the wifi module appears in the content of `/proc/modules` so it can be synchronous and cheap - i'm personnally just interested in that for power saving.)

Frenzie:
Yes, I suppose that makes sense. PS More relevant reference: https://github.com/koreader/koreader/pull/3078

We currently have two states, that'd make three. Wifi enabled, wifi connected (= current wifi enabled), internet connected.

KenMaltby:
I wonder if you need three, once the "Connected to ..." box is dissmissed, you could use the regular WiFi icon and have it change to the internet (globe) icon?

NiLuJe:
@poire-z: I'd gladly merge a /proc/modules approach instead of the lsmod one we currently use in a couple places ;). (:isWifiOn on Kobo, f.g.).

Frenzie:
@KenMaltby In the long run that sounds like a good idea.

poire-z:
@NiLuJe : to spare you a few minutes if you go at it, here's what I have for my Kobo:
```lua
-- We should not standby/suspend if wifi kernel modules are inserted, otherwise crash
function UIManager:isWifiEnabled()
    local wifi = false
    local fd =  io.open("/proc/modules", "r")
    if fd then
        local modules = fd:read("*all")
        fd:close()
        -- lsmod is usually empty, unless wifi or usb is enabled
        --     dhd 240713 0 - Live 0x7f19c000   (= présence directory /proc/sys/net/ipv4/conf/eth0)
        --     sdio_wifi_pwr 486 0 - Live 0x7f198000
        -- we could alternatively check if lfs.attributes("/proc/sys/net/ipv4/conf/eth0", "mode") == "directory"
        if modules:len() > 0 then
            if modules:find("dhd ") or modules:find("sdio_wifi_pwr ") then
                wifi = true
            end
        end
    end
    return wifi
end
```

Frenzie:
Isn't lsmod and /proc/modules basically the same thing, except lsmod may pull some info from other sources like /sys/module/*?

NiLuJe:
Most likely, but we have to fork() to use lsmod, which arguably puts more pressure on the system ;).

NiLuJe:
BusyBox appears to only be checking [https://git.busybox.net/busybox/tree/modutils/lsmod.c?h=1_31_stable](/proc/modules) ;).

Frenzie:
According to the manual for real lsmod it's no different:

> lsmod is a trivial program which nicely formats the contents of the /proc/modules, showing what kernel modules are currently loaded.

https://www.mankier.com/8/lsmod

So yes, clearly using lsmod is a bit stupid for our purposes. :-)

frontend/apps/reader/modules/readerfooter.lua
robert00s: done

frontend/apps/reader/modules/readerfooter.lua
robert00s: done

frontend/util.lua
robert00s: done

frontend/apps/reader/modules/readerfooter.lua
robert00s: Done. Thanks for review :)

frontend/apps/reader/modules/readerfooter.lua
robert00s: Thanks.
I added another settings to switch between icons and letters.

frontend/apps/reader/modules/readerfooter.lua
robert00s: Done.

frontend/apps/reader/modules/readerfooter.lua
robert00s: Good tip. Switched.

frontend/util.lua
robert00s: Done. I hope it will work :)

frontend/apps/reader/modules/readerfooter.lua
robert00s: Done. Thanks :)

frontend/apps/reader/modules/readerfooter.lua
robert00s: > That's a keyboard, right? Do we have a chip-like glyph available instead? That vaguely sounds more appropriate here, but I'm nitpicking ;).

@NiLuJe 
I like that icon. It look almost like chip :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -359,21 +435,40 @@ function ReaderFooter:updateFooterTextGenerator()
...
+        book_time_to_read = T(_("Book time to read (%1)"),symbol_prefix[symbol].book_time_to_read),
+        chapter_time_to_read = T(_("Chapter time to read (%1)"), symbol_prefix[symbol].chapter_time_to_read),
+        frontlight = T(_("Frontlight level (%1)"), symbol_prefix[symbol].frontlight),
+        mem_usage = T(_("KOReader memory usage (%1)"), symbol_prefix[symbol].mem_usage),
+        wifi_status = T(_("Wi-Fi status (%1)"), symbol_prefix[symbol].wifi_status),
+    }
+    return option_titles[option]
+end
 
 function ReaderFooter:addToMainMenu(menu_items)
```
poire-z: That menu is getting crowded :) I wouldn't mind a separator between the global options and the list of available footer items (or have these in a sub menu?).
Also may be move the `Settings>` among the top ones instead of having it burried after the list of footer items?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -298,6 +240,131 @@ function ReaderFooter:setupTouchZones()
     })
 end
 
+-- function that generates footer text for each mode
+function ReaderFooter:footerTextGeneratorMap(option)
+    local ret_func = function() return "" end
+    if option == "empty" then
```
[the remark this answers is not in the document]
robert00s: Long story short. I needed use global variable (eg self.settings.text_prefix) so I created ReaderFooter:footerTextGeneratorMap(). After that I made a trick with local function 

```
frontlight = function(footer)
    local aaa = footer.settings.text_prefix
    ...
end
```

so no need to use global function anymore.
I restore changes with `local footerTextGeneratorMap = {}`.
Thanks for paying attention :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -160,6 +101,7 @@ function ReaderFooter:init()
         frontlight = false,
         mem_usage = false,
         wifi_status = false,
+        text_prefix = "icons"
```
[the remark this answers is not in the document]
robert00s: Done.

frontend/apps/reader/modules/readerfooter.lua
robert00s: Done. Settings moved to top.

robert00s:
I also change time icon to ⌚
I'm not completely happy with the icon percentage (◔). Maybe `◴` or `🌖` or another?

NiLuJe:
> I also change time icon to ⌚
> I'm not completely happy with the icon percentage (◔). Maybe `◴` or `🌖` or another?

U+1F56E (🕮) would work, but I'm not sure we have anything that ships that glyph...

NiLuJe:
Random other idea: U+2920 (⤠)

NiLuJe:
FreeSerif's fancy % sign is at U+E879 (it's in the private use block, so it may be short-circuited by something else entirely...).

poire-z:
> I'm not completely happy with the icon percentage (◔)

Same here :) This black quarter makes it the only icon that catch my eye (there's something black at the bottom: when reading text, my peripheral vision catches it, or something :).
(No alternative suggestion in my small reference epub with a limited set of symbols, sorry :|)

frontend/util.lua
```diff
@@ -99,6 +99,91 @@ function util.secondsToClock(seconds, withoutSeconds)
...
+
+-- Converts seconds to a period of time string.
+
+---- @int seconds number of seconds
+---- @bool withoutSeconds if true 1h30', if false 1h30'10''
+---- @bool hmsFormat, if true format 1h30m10s
+---- @treturn string clock string in the form of 1h30' or 1h30'10''
+function util.secondsToHClock(seconds, withoutSeconds, hmsFormat)
+    seconds = tonumber(seconds)
+    if seconds == 0 or seconds ~= seconds then
```
poire-z: `seconds ~= seconds` ? will always be false and of no use :)

frontend/util.lua
```diff
@@ -99,6 +99,91 @@ function util.secondsToClock(seconds, withoutSeconds)
...
 end
 
+
+-- Converts seconds to a period of time string.
+
+---- @int seconds number of seconds
+---- @bool withoutSeconds if true 1h30', if false 1h30'10''
+---- @bool hmsFormat, if true format 1h30m10s
+---- @treturn string clock string in the form of 1h30' or 1h30'10''
+function util.secondsToHClock(seconds, withoutSeconds, hmsFormat)
```
poire-z: Seems to never be called with `hmsFormat=true` ? Did you mean for this to be a setting?

poire-z:
Quick test of rev2 on my device, just some observations (haven't looked at the code to pinpoint the issues):
- Reading percentage used to get a letter `R:` (to differentiate it from the battery % which got B). I'd like to get it back in letter mode :)
- progress % format does not apply (it worked with rev1), I keep getting 52% even if I chose `2 digits after` (but it shows fine with the additional digits in the menu item text)
- Menu: `Text prefix>` and `Text separator>`, probably better just `Prefix>` and `Separator>`.
- Code: still `self.settings.text_separator`, for coherency with your change to `item_prefix`, it better be `items_separator` or just `separator`
- Menu: `Without separator` => `No separator`
- Alignment does not work when the progress bar is displayed, as the whole thing take the whole width: this item could be made disabled when that happens
- I'd really like a separator after "Auto refresh time" :) (just a matter of adding `separator = true` to it
- May be `Show all at once`, `Reclaim bar height` and `Auto refresh time` could be moved into `Settings>`, leaving in top menu: `Settings>` , `Progress bar >`, separator, and the items - or even have `Progress bar` a toggle, and `Show chapter markers on progress bar` in `Settings>`
- Haven't

frontend/util.lua
robert00s: Not used yet. For future use :)

robert00s:
Should be better now.

poire-z:
Thanks :)
(very minor issue that is not worth a Rev4: when one has never toggled the progress bar, the `Alignment>` menu is not disabled - it is after toggle and back.)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -456,13 +538,239 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                        callback = function()
+                            self.settings.time_format = "12"
+                            self:updateFooter()
+                            UIManager:setDirty(nil, "ui")
+                        end,
+                    },
+                }
+            },
+            {
+                text = _("Durations format"),
```
Frenzie: Duration format, like the variable.

Frenzie: Or just time format?

poire-z:
≡ (seen at #5277), looking like a vu-meter, could be nice to replace the KOreder memory icon (currently a keyboard ⌨). I guess that's:
image

Issue #5205, opened 2019-08-13
Android: bug(s) in runtime permissions 


pazos:
* KOReader version: from 2019.07.
* Device: Android (api23+)

#### Issue

1. ~~If the user don't give write permission on the external storage the activity closes itself but the native thread is still on the background uncompressing assets, until it fails when it can't write to cache folder~~ **solved**

2. ~~If the user uninstalls the program but keeps /sdcard/koreader and had setup a custom timeout the next time the program is installed it will request the permission at launch (because it is stored on our own settings and restored every launch). It should request WRITE_SETTINGS just if the user does an explicit action (ie: changes screen brightness or screen timeout).~~ **solved**

3. Minor: the frontlight widget sets its new value before asking for permission.

pazos:
Regarding:

> **Minor: the frontlight widget sets its new value before asking for permission**

I think we would want to avoid touching system-wide light settings and just modify the light of our activity. In this case we don't need to ask permissions and we don't force the user to manual light levels outside KOReader.

Frenzie:
Fully on board with not messing with system-wide light settings. :-)

pazos:
> Fully on board with not messing with system-wide light settings. :-)

Ok, I will implement this on top of https://github.com/koreader/android-luajit-launcher/pull/171 to avoid ugly casting.

AFAICT the in-activity brightness is [one way only](https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#screenBrightness) (set a float between 0.0 and 1.0 to use custom levels and anything lower than 0 to use system defaults), which means there isn't any way to retrieve current level of the activity (as it is expected to be equal to system brightness or managed/recorded by the activity).

pazos:
Implemented in https://github.com/koreader/android-luajit-launcher/pull/171/commits/7fd77e930f3ee0fcd300c87eb831905346d3a03b.

I kept everything as it was. We get and set int values from MIN(0) to MAX(255). All these values are **custom** Any number below 0 will be treated as system defaults. For convenience  BRIGHTNESS_DEFAULT(-1) will be used if the user doesn't override this setting.

Note 1: we start with default level and we need set custom user level, if exists, during device initialization. Custom level will be used **only** for this app.

Note 2: once a custom level is set changing screen brightness from android settings won't change the value in KOReader.

pazos:
And now my headache begins:

I spent a few minutes reading frontend/ui/widget/frontlightwidget.lua for the first time and I would need some help here.

[...]

PR #5207, opened 2019-08-14
android: some runtime permissions / keys fixes.

### LINKED ISSUE #5206: Android: bug with MUTE hardware key if volume keys are used to turn pages.
Android hw mute key:

 - when it is pressed it goes to volume 0 step by step (it is like is triggering volume_down events).
- when pressed again it goes to the latest volume stored, again step by step (triggering volume_up events).

#### issue

when volume keys are configured to turn pages pressing the mute button will do page forward/back the number of steps between the actual volume and the desired new one.

Probably never reached by anyone because I don't know a single mobile phone with a hardware button for muting the device. It can be reproduced by generating events or with an emulator with a mute key (like https://www.bignox.com/)

#### possible solution:

consume mute key events when volume keys are used for page turning.

pazos:
- do not request write settings permission at launch
- do not uncompress assets if we don't have read and write permissions on external storage.
- do not propagate mute key events to the frontend

Requires https://github.com/koreader/android-luajit-launcher/pull/169
Related to #5205
Fixes #5206

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -172,7 +172,14 @@ function Device:init()
         android.setScreenOffTimeout(-1)
     else
         local timeout = G_reader_settings:readSetting("android_screen_timeout")
-        if timeout ~= nil and timeout ~= 0 then
+        if not timeout then return end
+        if timeout > 0 then
+            -- set a custom timeout if we already have write settings permission.
+            -- do not attempt to request permissions here.
+            if android.canWriteSettings() then
+                android.setScreenOffTimeout(timeout)
+            end
+        elseif timeout == -1 then
             android.setScreenOffTimeout(timeout)
         end
     end


poire-z:
Small (but big :) bug introduced by this PR:

On my Android 6.0.1 phone, `v2019.08-4-g0b8d154_2019-08-13` launches correctly, while `v2019.08-9-gaae6acb_2019-08-14` crashes:

```
1302  1314 V KOReader: dl.lua - dl.dlopen - library lname detected /system/lib/libc.so
1302  1314 V KOReader: dl.lua - sys_dlopen - loading library /system/lib/libc.so
1302  1314 W linker  : /data/data/org.koreader.launcher/files/libs/libwrap-mupdf.so: unused DT entry: type 0xf arg 0xbbb
1302  1314 E KOReader: [NativeThread]: failed to run script: frontend/document/canvascontext.lua:42: attempt to call field 'isColorEnabled' (a nil value)
1302  1314 E KOReader: [NativeThread]: Stopping due to previous errors
```
Probably because of an early/wrong `return`:
https://github.com/koreader/koreader/commit/50ce8d47cdd9433a4442c68d79ed01b1f35ebe98#diff-c14ee0e93f7187eeb6f6fe8af52cf0b7R176
which causes `android Device:init()` to not call `Generic.init(self)` and not get the other settings (like isColorEnabled or hasMultiTouch).

Dunno if you want the `if timeout > 0 elseif timeout == -1` wrapped by a `if timeout then`, or one of these 2 be the default when `G_reader_settings:readSetting("android_screen_timeout")` has not been defined.

poire-z:
pinging @pazos, in case you miss that ^
(the nightly apks would crash for new users or those that have neither played with the timeout setting)

Frenzie:
Which is probably most. ;-)

pazos:
:man_facepalming: Just commented in #5228 without idea of the issue.

PR #5208, opened 2019-08-14
Honor the alpha channel in our icon for the About popup

NiLuJe:
It ends up with a black background otherwise.
(Old regression that dates back to my alpha shenanigans a few months
ago)


NiLuJe:
I've been aware of it for quite a while, but it managed to fell through my TODO lists for, err, a while :D.

PR #5211, opened 2019-08-14
Simplify isWifiOn on Kobo

NiLuJe:
Don't fork for lsmod + grep, we can perfectly well parse /proc/modules
ourselves ;).

c.f., https://github.com/koreader/koreader/pull/5203#issuecomment-521277628

### DIFF
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -406,7 +406,22 @@ function Kobo:initNetworkManager(NetworkMgr)
     -- NOTE: Cheap-ass way of checking if WiFi seems to be enabled...
     --       Since the crux of the issues lies in race-y module unloading, this is perfectly fine for our usage.
     function NetworkMgr:isWifiOn()
-        return 0 == os.execute("lsmod | grep -q " .. (os.getenv("WIFI_MODULE") or "sdio_wifi_pwr"))
+        local fd = io.open("/proc/modules", "r")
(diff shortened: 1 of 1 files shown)


Frenzie:
And I'm not really sure when we gained this function, but this means the icon can also be shown on Kobo.

NiLuJe:
Yeah, I vaguely checked other patforms:

Kindle: lipc via dbus or fork(), meh :/
Cervantes: sysfs, but only true if the *interface* is up. Sill, probably okay for the icon ;).
PB: InkView. Possibly perfectly fine, too, but here be dragons.
Sony: double fork(), meh :/

pazos:
> Cervantes: sysfs, but only true if the interface is up. Sill, probably okay for the icon ;)

Yep, on cervantes we report isWifiOn as true [based on 2nd layer](https://en.wikipedia.org/wiki/Data_link_layer) according to IEEE 802.11 wireless LAN protocol.

This way wifi is on even on networks where we cannot ping remote locations but can still access to LAN services.

The reason we don't care about modules is that having them loaded doesn't mean that the interface is associated with an AP. For instance isWifiOn will return false if the interface is up but wpa_supplicant failed to associate.

NiLuJe:
Yep, whereas I needed isWifiOn as-is on Kindle to deal with crappy launchers that left WiFi in a broken/dangerous/crashy state ;).

PR #4474, opened 2019-01-11
[fix] PocketBook 626 new firmware with new codename

Frenzie:
Might need to be repeated for the others?

Fixes #4473.

PR #5215, opened 2019-08-15
PocketBook 626 new codename

vvasseur:
Similar to #4474

PocketBook Touch Lux 3 (626) seems to have a new codename with the latest firmware (mine is FR626.5.14.1228 released on 20170718_143341)


poire-z:
Thanks!

PR #5218, opened 2019-08-17
Fix after #5203

robert00s:
https://github.com/koreader/koreader/pull/5203#issuecomment-522156268
https://github.com/koreader/koreader/pull/5203#pullrequestreview-276228925

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -565,7 +565,7 @@ function ReaderFooter:addToMainMenu(menu_items)
             {
                 text = _("Alignment"),
                 enabled_func = function()
-                    return self.settings.disable_progress_bar
+                    return not self.settings.disable_progress_bar == false
                 end,
                 sub_item_table = {
                     {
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -565,7 +565,7 @@ function ReaderFooter:addToMainMenu(menu_items)
             {
                 text = _("Alignment"),
                 enabled_func = function()
-                    return self.settings.disable_progress_bar
+                    return not self.settings.disable_progress_bar == false
                 end,
```
Frenzie: Maybe some parentheses for clarity?

poire-z:
Just in case you're not yet bored with the footer :)
As you show some sample in the menu item text for Progress percentage format, you could show it for more/all of them (most probably just useful for the Duration format, classic vs modern is a bit obscure -  although one immediately sees what it means in the footer when updating...)

frontend/apps/reader/modules/readerfooter.lua
poire-z: Or just `return self.settings.disable_progress_bar == true`.
(our touchmenu code considers `nil` to mean: enabled - it just checks for `false` meaning: disabled)

robert00s:
Now all settings have samples:

obraz

obraz

obraz

obraz

PR #5219, opened 2019-08-17
bump crengine and others, adds HTML extended debug view

### LINKED ISSUE #5191: Hyphenation for dash
Is it possible to make the hyphenation treat the dash (or em dash or ??? dash) characters like hyphens?

Hyphen_sml

I don't know what kind of dash is in this screenshot, but it looks ugly.
Maybe it can be done using a "Style tweak"?

Also, does anyone knows where can I find a better hyphenation .pattern for English, since the default does not recognize (for splitting) a lot of words?

poire-z:
Bump crengine https://github.com/koreader/crengine/pull/305 :
- View HTML: option to show text unicode codepoints
- Fix attribute parsing: decode &-encoded chars
- Text: allow wrap after more unicode spaces and hyphen (Closes #5191)
- Fix: no left hanging when hanging punctuation disabled
- lvtextfm.cpp: more comments, some formatting cleanup
- Optimize background image drawing
- CSS: fix background-image file path resolution

Bump harfbuzz to 2.6.0 https://github.com/koreader/koreader-base/pull/917
Bump luasec to 0.8.1 https://github.com/koreader/koreader-base/pull/949

cre.cpp: https://github.com/koreader/koreader-base/pull/950
Adds `setBackgroundImage()` proxy function to crengine facility to set background textures (this is unrelated to the background-image CSS fixes).
No frontend code (yet) to select such an image (dunno if we should provide some free samples, let the user put his ones in styletweaks/ and build a menu for easy selection - or just use a a FileChooser).
For those interested to see, just add some of the tx_*.jpg files from https://github.com/koreader/crengine/tree/master/android/res/drawable in your koreader root directory, and add something like that:
```diff
--- a/frontend/apps/reader/modules/readertypeset.lua
+++ b/frontend/apps/reader/modules/readertypeset.lua
@@ -122,6 +122,7 @@ function ReaderTypeset:onReadSettings(config)
         self.nightmode_images = (global == nil or global == 1) and 1 or 0
     end
     self:toggleNightmodeImages(self.nightmode_images)
+    self.ui.document:setBackgroundImage("tx_old_book.jpg")
 end

 function ReaderTypeset:onSaveSettings()
```

<kbd>image</kbd>
(This is totaly unbearable on eInk - at least on my Kobo GloHD, which doesn't really like grey -, but it may be interesting on Android.)

View HTML: adds a 3rd view (extended debug view), showing the unicode codepoint of each char and crengine rendereing methods.
<kbd>image</kbd>
(might be only useful to me, but I need the first debug view without all that).

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -632,17 +632,22 @@ function ReaderHighlight:viewSelectionHTML(debug_view)
     if self.selected_text and self.selected_text.pos0 and self.selected_text.pos1 then
         -- For available flags, see the "#define WRITENODEEX_*" in crengine/src/lvtinydom.cpp
         local html_flags = 0x3030 -- valid and classic displayed HTML, with only block nodes indented
-        if debug_view then
+        if not debug_view then
+            debug_view = 0
+        end
+        if debug_view == 1 then
             -- Each node on a line, with markers and numbers of skipped chars and siblings shown,
             -- with possibly invalid HTML (text nodes not escaped)
             html_flags = 0x3353
-            -- html_flags = 0x3753 -- use this to additionally see rendering methods
+        elseif debug_view == 2 then
+            -- Additionally see rendering methods and unicode codepoint of each char
+            html_flags = 0x3757
         end
         local html, css_files = self.ui.document:getHTMLFromXPointers(self.selected_text.pos0,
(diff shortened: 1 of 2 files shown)


Frenzie:
> (might be only useful to me, but I need the first debug view without all that).

Nah, I'll (very rarely) consider it useful not to have to open up a document on my computer for that. ;-)

PR #5220, opened 2019-08-17
[Desktop]: Open writable font dir in FM, toggle between system+user and user fonts, fix openLink on mac...

### LINKED ISSUE #5093: Add more fonts for AppImage/desktop
Is it possible to do the same for koreader.AppImage? 
I created a fonts folder under ~/.config/koreader and put some fonts inside, the AppImage does not detect them.

_Originally posted by @ChenguangZhang in https://github.com/koreader/koreader/issues/5013#issuecomment-505567224_

<hr>

Right now you can set the `EXT_FONT_DIR` variable, so it could also be solved through the launch script.

pazos:
~~I enabled it on the emulator just to test (I needed to export EXT_FONT_DIR =~/.local/share/fonts)~~

~~The emulator should be removed since we cannot grant it is linux.~~

~~The user has a menu to switch from its own set of fonts (~/.local/share/fonts) and system fonts (/usr/share/fonts). If the user chooses her/his own fonts she/he can open it on the file manager (via xdg-open).~~

~~It seems a bit hacky but works really well.~~

~~The user can toggle system fonts. The toggle just creates or deletes a symbolic link to /usr/share/fonts in a folder inside our EXT_FONT_DIR.~~

Thanks to @poire-z and @Frenzie for the hints!

**to-do**:
- ~~place the new submenu somewhere.~~
- ~~remove the emulator from isDesktopLinux.~~
- ~~enable "open dir in fm" just if "xdg-open" exists.~~
- ~~infomessage saying "changes will apply after a restart (not true, changes will apply on the next cr3 document loaded, but that is harder to explain :man_facepalming:~~ )
- ~~more sanity checks ¿?~~

Fixes #5093

### DIFF
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -40,6 +40,9 @@ local Device = {
     canToggleGSensor = no,
     canToggleMassStorage = no,
     canUseWAL = yes, -- requires mmap'ed I/O on the target FS
+    canRestart = yes,
+    canReboot = no,
+    canPowerOff = no,
 
     -- use these only as a last resort. We should abstract the functionality
     -- and have device dependent implementations in the corresponting
@@ -53,9 +56,7 @@ local Device = {
     isSonyPRSTUX = no,
     isSDL = no,
     isEmulator = no,
-    canRestart = yes,
-    canReboot = no,
-    canPowerOff = no,
+    isDesktop = no,
 
     -- some devices have part of their screen covered by the bezel
     viewport = nil,
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -6,6 +6,42 @@ local logger = require("logger")
 local function yes() return true end
 local function no() return false end
 
+-- xdg-open is used on most linux systems
+local function hasXdgOpen()
+    local std_out = io.popen("xdg-open --version 2>/dev/null")
+    local version = nil
+    if std_out ~= nil then
+        version = std_out:read()
+        std_out:close()
+    end
+    return version ~= nil
+end
+
+-- open is the macOS counterpart
+local function hasMacOpen()
+    local std_out = io.popen("open")
+    local all = nil
+    if std_out ~= nil then
+        all = std_out:read()
+        std_out:close()
+    end
+    return all ~= nil
+end
+
+-- get the name of the binary used to open links
+local function getLinkOpener()
+    local enabled = false
+    local tool = nil
+    if jit.os == "Linux" and hasXdgOpen() then
+        enabled = true
+        tool = "xdg-open"
+    elseif jit.os == "OSX" and hasMacOpen() then
+        enabled = true
+        tool = "open"
+    end
+    return enabled, tool
+end
+
 local Device = Generic:new{
     model = "SDL",
     isSDL = yes,
@@ -17,10 +53,11 @@ local Device = Generic:new{
     needsScreenRefreshAfterResume = no,
     hasColorScreen = yes,
     hasEinkScreen = no,
-    canOpenLink = yes,
+    canOpenLink = getLinkOpener,
     openLink = function(self, link)
-        if not link or type(link) ~= "string" then return end
-        return os.execute("xdg-open '"..link.."'") == 0
+        local enabled, tool = getLinkOpener()
+        if not enabled or not tool or not link or type(link) ~= "string" then return end
+        return os.execute(tool.." '"..link.."'") == 0
     end,
 }
 
@@ -28,6 +65,7 @@ local AppImage = Device:new{
     model = "AppImage",
     hasMultitouch = no,
     hasOTAUpdates = yes,
+    isDesktop = yes,
 }
 
 local Emulator = Device:new{
@@ -37,10 +75,12 @@ local Emulator = Device:new{
     hasFrontlight = yes,
     hasWifiToggle = yes,
     hasWifiManager = yes,
+    isDesktop = yes,
 }
 
 local Linux = Device:new{
     model = "Linux",
+    isDesktop = yes,
 }
 
 local UbuntuTouch = Device:new{
--- a/frontend/ui/elements/font_settings.lua
+++ b/frontend/ui/elements/font_settings.lua
@@ -1,0 +1,81 @@
+local Device = require("device")
+local logger = require("logger")
+local util = require("util")
+local _ = require("gettext")
+
+--[[ Font settings for desktop linux and mac ]]--
+
+local function getUserDir()
+    return os.getenv("HOME").."/.local/share/fonts"
+end
+
+-- System fonts are common in linux
+local function getSystemDir()
+    local path = "/usr/share/fonts"
+    if util.pathExists(path) then
+        return path
+    else
+        -- mac doesn't use ttf fonts
+        return nil
+    end
+end
+
+local function usesSystemFonts()
+    return G_reader_settings:isTrue("system_fonts")
+end
+
+local function openFontDir()
+    local user_dir = getUserDir()
+    local openable = util.pathExists(user_dir)
+    if not openable then
+        logger.info("Font path not found, making one in ", user_dir)
+        openable = util.makePath(user_dir)
+    end
+    if not openable then
+        logger.warn("Unable to create the folder ", user_dir)
+        return
+    end
+    if Device:canOpenLink() then
+        Device:openLink(user_dir)
+    end
+end
+
+local FontSettings = {}
+
+function FontSettings:getPath()
+    if usesSystemFonts() then
+        local system_path = getSystemDir()
+        if system_path ~= nil then
+            return getUserDir()..";"..system_path
+        end
+    end
+    return getUserDir()
+end
+
+function FontSettings:getMenuTable()
+    return {
+        text = _("Font settings"),
+        separator = true,
+        sub_item_table = {
+            {
+                text = _("Enable system fonts"),
+                checked_func = usesSystemFonts,
+                callback = function()
+                    G_reader_settings:saveSetting("system_fonts", not usesSystemFonts())
+                    local UIManager = require("ui/uimanager")
+                    local InfoMessage = require("ui/widget/infomessage")
+                    UIManager:show(InfoMessage:new{
+                        text = _("This will take effect on next restart.")
+                    })
+                end,
+            },
+            {
+                text = _("Open fonts folder"),
+                keep_menu_open = true,
+                callback = openFontDir,
+            },
+        }
+    }
+end
+
+return FontSettings
--- a/platform/appimage/AppRun
+++ b/platform/appimage/AppRun
@@ -10,9 +10,6 @@ cd "${KOREADER_DIR}" || exit
 # export load library path
 export LD_LIBRARY_PATH=${KOREADER_DIR}/libs:$LD_LIBRARY_PATH
 
-# export external font directory
-export EXT_FONT_DIR="${HOME}/.fonts"
-
 RETURN_VALUE=85
 
 if [ $# -eq 0 ]; then
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -49,6 +49,9 @@ function ReaderFont:init()
     end
     -- build face_table for menu
     self.face_table = {}
+    if Device:isDesktop() then
+        table.insert(self.face_table, require("ui/elements/font_settings"):getMenuTable())
+    end
     local face_list = cre.getFontFaces()
     for k,v in ipairs(face_list) do
         table.insert(self.face_table, {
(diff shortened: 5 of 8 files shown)


pazos:
Pinging @Frenzie @poire-z @robert00s @NiLuJe for suggestions. Thanks!

poire-z:
(Not really using linux for UI stuff needing fonts, I only use the emulator.)
But, does it really need to be a choice?
fontlist.lua's `_readList()` walk subdirectories (you could make it walk symlinks if it doesn't follow them).
You could just use the user's `~/.local/share/fonts` and create in it a link to the system folder.
With just an option to ignore (and remove that link) or not (and recreate that link) system fonts, if needed.

NiLuJe:
> (Not really using linux for UI stuff needing fonts, I only use the emulator.)
> But, does it really need to be a choice?
> fontlist.lua's `_readList()` walk subdirectories (you could make it walk symlinks if it doesn't follow them).
> You could just use the user's `~/.local/share/fonts` and create in **it** a link to the system folder.
> With just an option to ignore (and remove that link) or not (and recreate that link) system fonts, if needed.

Which it? The XDG folder?

poire-z:
**it** = the user's ~/.local/share/fonts
If the user has his own fonts there, I guess he rather want to have them used (if **it**'s empty, create the symlink so **it** does not stay empty). The option would be to additionally bring system fonts into our koreader menu, or not.

(I have no idea what The XDG folder is, and what xdg-open does - assuming its a generic opener that chose the right app, it will open the ~/.local/share/fonts in an external file browser, and then what the user could do there?)

pazos:
> (I have no idea what The XDG folder is, and what xdg-open does - assuming its a generic opener that chose the right app, it will open the ~/.local/share/fonts in an external file browser, and then what the user could do there?)

Copy fonts there or backup already copied fonts. This is the only place where the user has RW granted (because the fonts partition will be in /usr/lib/koreader on debian and inside the app bundle on the appimage).

If an user wants to add a font but not all the fonts already present on the system then leave the *use system fonts* unchecked, click to open font dir and copy fonts on the opened folder. Not sure if it make much sense.

poire-z:
Also, if **it** is a kind of standard user dir for fonts that other apps may use too, couldn't backing it up make these other apps not function correctly anymore (apps launched before, apps launched while koreader is running)?

NiLuJe:
Okay, I would touch the XDG folder ourselves.

A *lot* of stuff depends on it being the way it currently is, we should only ever treat it as RO on our end.

pazos:
> Also, if **it** is a kind of standard user dir for fonts that other apps may use too, couldn't backing it up make these other apps not function correctly anymore (apps launched before, apps launched while koreader is running)?

Yeah, my logic was flawed. Yours is way better and waaaay simpler. I will submit changes in a minute :smile:

frontend/device/sdl/device.lua
```diff
@@ -17,7 +28,7 @@ local Device = Generic:new{
     needsScreenRefreshAfterResume = no,
     hasColorScreen = yes,
     hasEinkScreen = no,
-    canOpenLink = yes,
+    canOpenLink = isXdgOpen,
```
Frenzie:
```suggestion
    canOpenLink = hasXdgOpen,
```

frontend/device/sdl/device.lua
```diff
@@ -6,6 +6,17 @@ local logger = require("logger")
 local function yes() return true end
 local function no() return false end
 
+-- xdg-open should be available on most linux systems
+local function isXdgOpen()
```
Frenzie:
```suggestion
local function hasXdgOpen()
```

poire-z:
Thinking about it, if we make a symlink to the system font dir in **it**, couldn't these other apps see multiple copies of the system fonts if they are made to look in both the user and the system dirs (like I guess they should logically do), with possible side effects?
May be KOReader should do the same thing: look in both directories.
Could be just a matter of having a `getAdditionalExternalFontDir()` reading a `os.getenv("EXT_FONT_DIR2")` and some tweaks in `FontList:getFontList()`.

Frenzie:
I'm not sure why you guys are all being so difficult. ;-)

https://github.com/koreader/koreader/blob/f56e9bcf981c68d91cb37e151b4a02cb70b389be/frontend/fontlist.lua#L97-L100

pazos:
Thanks for the hints!

I updated the PR and added support for openLink on macs too.

Frenzie:
I was going to suggest splitting on `:` but then I noticed the code does `;`.

frontend/ui/elements/common_info_menu_table.lua
```diff
@@ -7,6 +7,11 @@ local T = require("ffi/util").template
 
 local common_info = {}
 
+
+if Device:isDesktop() then
+    common_info.font_settings = require ("ui/elements/font_settings"):getMenuTable()
```
Frenzie: Putting this in the *common* info menu and then disabling it in the file manager with the MenuSorter is pretty weird. ;-)

Frenzie: I created it because there was a lot of duplication that was annoying to deal with.

Reader-specific stuff should simply go around/after here:

https://github.com/koreader/koreader/blob/33e4c36a422d8665fd2e330f930f94446dfe4529/frontend/apps/reader/modules/readermenu.lua#L144-L152

pazos: Wow, I just requested your review, you're fast.

Sorry, I'm not familiar with the menu sorter. Where should I put the menu?, somewhere inside the reading app?.

Frenzie: Just a coincidence, I saw your latest commit before you did that.

frontend/device/generic/device.lua
```diff
@@ -53,9 +56,7 @@ local Device = {
     isSonyPRSTUX = no,
     isSDL = no,
     isEmulator = no,
-    canRestart = yes,
-    canReboot = no,
-    canPowerOff = no,
+    isDesktop = no,
```
Frenzie: Not sure how much it matters, but by the xdg-open = desktop logic I believe this means Sailfish `isDesktop() == true`. Sailfish should probably be able to run on a regular x86 desktop, but, um, well… ;-)

frontend/device/sdl/device.lua
```diff
@@ -6,6 +6,42 @@ local logger = require("logger")
...
+    local std_out = io.popen("xdg-open --version 2>/dev/null")
+    local version = nil
+    if std_out ~= nil then
+        version = std_out:read()
+        std_out:close()
+    end
+    return version ~= nil
+end
+
+-- open is the macOS counterpart
```
Frenzie: And `start` on Windows. Not relevant to the code, but I was curious and looked it up.

frontend/device/sdl/device.lua
```diff
@@ -17,17 +53,20 @@ local Device = Generic:new{
     needsScreenRefreshAfterResume = no,
     hasColorScreen = yes,
     hasEinkScreen = no,
-    canOpenLink = yes,
+    canOpenLink = getLinkOpener,
     openLink = function(self, link)
+        local enabled, tool = getLinkOpener()
         if not link or type(link) ~= "string" then return end
-        return os.execute("xdg-open '"..link.."'") == 0
+        if not enabled or tool == nil then return end
```
Frenzie: Unless the distinction between false and nil is meaningful (which it could hardly said to be here; a boolean concatenates just as badly as nil) I'd personally just write `or not tool`.

frontend/ui/elements/font_settings.lua
```diff
@@ -0,0 +1,68 @@
...
+
+function FontSettings:getMenuTable()
+    return {
+        text = _("Font settings"),
+        sub_item_table = {
+            {
+                text = _("Enable system fonts"),
+                checked_func = usesSystemFonts,
+                callback = function()
+                    G_reader_settings:saveSetting("desktop_uses_system_fonts", not usesSystemFonts())
```
Frenzie: Or just `system_fonts`? The setting name seems a bit verbose. Also it might make sense to have this toggle on Android as well.

frontend/ui/elements/font_settings.lua
```diff
@@ -0,0 +1,68 @@
...
+                    local UIManager = require("ui/uimanager")
+                    local InfoMessage = require("ui/widget/infomessage")
+                    UIManager:show(InfoMessage:new{
+                        text = _("This will take effect on next restart.")
+                    })
+                end,
+            },
+            {
+                text = _("Open fonts dir"),
```
Frenzie:
```suggestion
                text = _("Open fonts folder"),
```

Or directory, but I think we (and most filemanagers) use folder.

platform/appimage/AppRun
```diff
@@ -10,8 +10,9 @@ cd "${KOREADER_DIR}" || exit
 # export load library path
 export LD_LIBRARY_PATH=${KOREADER_DIR}/libs:$LD_LIBRARY_PATH
 
-# export external font directory
-export EXT_FONT_DIR="${HOME}/.fonts"
+# create user font directory if it does not exist.
+USER_FONT_DIR="${HOME}/.local/share/fonts"
+[ ! -d "${USER_FONT_DIR}" ] && mkdir -pv "${USER_FONT_DIR}"
```
Frenzie: Would this make more sense in Lua, in the fairly rare circumstance the user actually taps on the menu item?

frontend/ui/elements/reader_menu_order.lua
```diff
@@ -24,6 +24,7 @@ local order = {
     typeset = {
         "set_render_style",
         "style_tweaks",
+        "font_settings",
         "----------------------------",
         "change_font",
```
poire-z: Better be moved below the separator (that separates style stuff from font stuff)

frontend/device/generic/device.lua
pazos: nah, Sailfish canOpenLink but is not isDesktop. In fact Sailfish isAndroid.

Desktop devices are sdl that run without a sandbox. Which means everything except Ubuntu Touch.

Frenzie: > Desktop devices are sdl that run without a sandbox. Which means everything except Ubuntu Touch.

Pretty sure Sailfish is more desktop than Ubuntu Touch then. ;-)

platform/appimage/AppRun
pazos: Right now the menu item is disabled if the folder don't exist (which could happen only on the emulator). Your idea seems better: item always enabled and mkdir on demand.

frontend/device/sdl/device.lua
pazos: ok, will do 

```lua
if not enabled or not tool or not link or type(link) ~= "string" then return end
```

frontend/apps/reader/modules/readermenu.lua
```diff
@@ -179,6 +179,10 @@ function ReaderMenu:setUpdateItemTable()
         table.insert(self.menu_items.screensaver.sub_item_table, ss_book_settings)
     end
 
+    if Device:isDesktop() then
+        self.menu_items.font_settings = require ("ui/elements/font_settings"):getMenuTable()
```
Frenzie:
```suggestion
        self.menu_items.font_settings = require("ui/elements/font_settings"):getMenuTable()
```

frontend/device/sdl/device.lua
```diff
@@ -6,6 +6,42 @@ local logger = require("logger")
...
+        std_out:close()
+    end
+    return all ~= nil
+end
+
+-- get the name of the binary used to open links
+local function getLinkOpener()
+    local enabled = false
+    local tool = nil
+    if hasXdgOpen() then
```
Frenzie: Might be a bit wasteful on non-Linux when we already have `jit.os`, which will return `Linux` or `OSX`. Maybe something like this?

```suggestion
    if jit.os == "Linux" and hasXdgOpen() then
```

frontend/ui/elements/reader_menu_order.lua
poire-z: And moved again below `change_font` ? :)
(Because I'm so used to find it on the 3rd position, and easthetically, having the following longer items together should look better, instead of having the little "Font >" in between.
(Would it need an added `separator=true` for that `Font settings>`, that would appear only on isDesktop?)

Frenzie: I'd be inclined to move it down one further in line with the general menu order, which would add an extra separator, Though in this case it's defensible that change font, hyphenation & hanging punctuation belong closer together.

pazos: **non Desktop vs Desktop**
Sin nombre

¿?

Frenzie: It's fine I guess, but like I said initially, I think it'd be best within the font menu itself.

poire-z: I'm fine with that screenshot.

pazos: > It's fine I guess, but like I said initially, I think it'd be best within the font menu itself.

Ok, that makes sense. I didn't understand.

pazos: Needs to stay on top of that menu because where system fonts are enabled the list of pages can grow a bit:

Sin nombre

PR #5223, opened 2019-08-18
Remove Polish keyboard

robert00s:
After #4886 Polish keyboard is no more needed. Now we can use standard keyboard with popup selection.

obraz


Frenzie:
Yeah, it's not worthwhile to have a more Polish QWERTZ? See for example the French AZERTY layout which uses only a few minor modifications: https://github.com/koreader/koreader/blob/f56e9bcf981c68d91cb37e151b4a02cb70b389be/frontend/ui/data/keyboardlayouts/fr_keyboard.lua

robert00s:
But Polish standard is QWERTY :) QWERTZ is very rarely used (eg on old typewriters).

Frenzie:
Ah, excellent. *shakes fist at AZERTY and QWERTZ*

Issue #5190, opened 2019-08-08
Never full refresh page


noembryo:
I know that I can change the number of pages for Refresh with `DRCOUNTMAX`, but, since I always do something that triggers a refresh (like dictionary lookup or open the menus or..) is it possible to have a "Never" as an option?

pazos:
I'm not sure if  `DRCOUNTMAX` has some sort of sanity checks. If not you could set it to a million pages :+1:

noembryo:
Hmm, good idea.. ;o)
It would be nice if the 0 value was "Never".
Now that I think about it, it might already be.
I'll have to check...

robert00s:
`Settings -> Screen -> E-ink settings -> Full refresh rate` , long tap on Custom 1 and input 0.
There is a chance that it will also work.

noembryo:
Looks like its working too :o)
I minor nitpick:
After changing the value, the menu label is not updated with the new value..

robert00s:
Thanks for info :)

PR #5227, opened 2019-08-19
Never full refresh

robert00s:
+ add `Never refresh` in `Settings -> Screen -> E-ink settings -> Full refresh rate`.
+ spinnerWidget insteed of inputWidget after long press Cutom x
+ After changing the value, the menu label is now updated see: https://github.com/koreader/koreader/issues/5190#issuecomment-522528644

obraz

obraz

### DIFF
--- a/frontend/ui/elements/refresh_menu_table.lua
+++ b/frontend/ui/elements/refresh_menu_table.lua
@@ -1,30 +1,51 @@
+local Device = require("device")
 local UIManager = require("ui/uimanager")
-local util = require("ffi/util")
 local _ = require("gettext")
+local Screen = Device.screen
+local T = require("ffi/util").template
 
-local function custom_1() return G_reader_settings:readSetting("refresh_rate_1") or 12 end
-
-local function custom_2() return G_reader_settings:readSetting("refresh_rate_2") or 22 end
-
-local function custom_3() return G_reader_settings:readSetting("refresh_rate_3") or 99 end
-
-local function custom_input(name)
-    return {
-        title = _("Input page number for a full refresh"),
-        type = "number",
-        hint = "(1 - 99)",
-        callback = function(input)
-            local rate = tonumber(input)
-            G_reader_settings:saveSetting(name, rate)
-            UIManager:setRefreshRate(rate)
-        end,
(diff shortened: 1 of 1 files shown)


Frenzie:
I use 22 on mine and I'd probably be happy with 99, but *literally* never doesn't strike me as fully desirable. But whatever, I guess that's up to the user. ¯\_(ツ)_/¯

frontend/ui/elements/refresh_menu_table.lua
```diff
@@ -1,30 +1,51 @@
...
 local _ = require("gettext")
+local Screen = Device.screen
+local T = require("ffi/util").template
 
-local function custom_1() return G_reader_settings:readSetting("refresh_rate_1") or 12 end
-
-local function custom_2() return G_reader_settings:readSetting("refresh_rate_2") or 22 end
-
-local function custom_3() return G_reader_settings:readSetting("refresh_rate_3") or 99 end
+local function custom(refresh_rate_type)
```
poire-z: nitpick, but for read/understandability, `refresh_rate`**_type** is a bit misleading, `refresh_rate_num` or `custom_refresh_rate_setting_name` is probably clearer.

frontend/ui/elements/refresh_menu_table.lua
```diff
@@ -1,30 +1,51 @@
...
+    UIManager:show(items)
 end
 
+
 return {
     text = _("Full refresh rate"),
     separator = true,
     sub_item_table = {
+        {
+            text = _("Never refresh"),
```
poire-z: Or just `Never` or `Manually` ? @Frenzie ?

Frenzie: `Manual and UI`?

frontend/ui/elements/refresh_menu_table.lua
robert00s: Done :)

frontend/ui/elements/refresh_menu_table.lua
robert00s: I use `Never`.

Issue #5232, opened 2019-08-21
[i18n] Translation should be context-dependent


Frenzie:
* KOReader version: v2019.09
* Device: Transifex (cf. #3754)

#### Issue
Currently all strings that have the same outward appearance are lumped together. This means, for example, `Pages` in the Style tweaks menu (i.e., page *layout*) and `Pages` in most other contexts (i.e., page *number*).

On the Transifex end that should be a simple switch, even if I can't find it right now, or maybe actually using `msgctxt` would do the trick, but on the frontend and string extraction end it's a lot more problematic.

PS Some semi-related PRs:

https://github.com/koreader/koreader-base/pull/159
https://github.com/koreader/koreader-base/pull/164
https://github.com/koreader/koreader/pull/586

PR #5234, opened 2019-08-21
[i18n, feat] Add basic context (msgctxt) support

Frenzie:
References https://github.com/koreader/koreader/issues/5232

Given an entry in the PO file like the following:

```
msgctxt "systemstat"
msgid "    Total"
msgstr "Totaal"
```

It can be addressed using:

```lua
_("systemstat", "    Total")
```

This allows to distinguish between separate instances of the same string, for example "Pages" meaning "Number of pages" and "Pages" meaning "Display of pages".

Extraction of this code pattern is not yet implemented by nightswatcher. xgettext didn't yet support Lua back in 2013 when all this was first added to the program, but now it does. Therefore it might make the most sense to replace the current Python extraction script with xgettext itself.

By default it only understands gettext.dgettext(), but that can be addressed by passing some extra command line arguments, for example:

```
xgettext -l lua -c --keyword=_:1c,2 *.lua
```

### DIFF
--- a/frontend/gettext.lua
+++ b/frontend/gettext.lua
@@ -21,9 +21,9 @@ function GetText_mt.__call(gettext, ...)
     end
     if argc == 2 then
         -- gettext.dgettext
-        local context = select(1, ...)
+        local domain = select(1, ...)
         local string = select(2, ...)
-        return gettext.context[context] and gettext.context[context][string] or string
+        return gettext.domain[domain] and gettext.domain[domain][string] or string
     end
     --if argc == 3 then -- gettext.ngettext end
     --if argc == 4 then -- gettext.dngettext end
@@ -49,7 +49,7 @@ end
 -- we only implement a sane subset for now
 
 function GetText_mt.__index.changeLang(new_lang)
-    GetText.context = {}
+    GetText.domain = {}
     GetText.translation = {}
     GetText.current_lang = "C"
 
@@ -77,10 +77,10 @@ function GetText_mt.__index.changeLang(new_lang)
             if data.msgid and data.msgstr and data.msgstr ~= "" then
                 local unescaped_string = string.gsub(data.msgstr, "\\(.)", c_escape)
                 if data.msgctxt and data.msgctxt ~= "" then
-                    if not GetText.context[data.msgctxt] then
-                        GetText.context[data.msgctxt] = {}
+                    if not GetText.domain[data.msgctxt] then
+                        GetText.domain[data.msgctxt] = {}
                     end
-                    GetText.context[data.msgctxt][data.msgid] = unescaped_string
+                    GetText.domain[data.msgctxt][data.msgid] = unescaped_string
                 else
                     GetText.translation[data.msgid] = unescaped_string
                 end
--- a/frontend/gettext.lua
+++ b/frontend/gettext.lua
@@ -75,13 +75,14 @@ function GetText_mt.__index.changeLang(new_lang)
         local line = po:read("*l")
         if line == nil or line == "" then
             if data.msgid and data.msgstr and data.msgstr ~= "" then
+                local unescaped_string = string.gsub(data.msgstr, "\\(.)", c_escape)
                 if data.msgctxt and data.msgctxt ~= "" then
                     if not GetText.context[data.msgctxt] then
                         GetText.context[data.msgctxt] = {}
                     end
-                    GetText.context[data.msgctxt][data.msgid] = string.gsub(data.msgstr, "\\(.)", c_escape)
+                    GetText.context[data.msgctxt][data.msgid] = unescaped_string
                 else
-                    GetText.translation[data.msgid] = string.gsub(data.msgstr, "\\(.)", c_escape)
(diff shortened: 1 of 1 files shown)


poire-z:
I know quite nothing about how gettext and our translation works, how we generate the po, etc...
So I don't know if these fields names and order are some gettext requirements or not.
Just wondering, if for readability and writability, it would be better to have to use:
```lua
_("    Total", "systemstat")
_("My english string that I want written here", "that alt context")
_("Pages", "rendermode")
_("Pages", "number")
```

Frenzie:
Yes, the order is demanded by gettext.

https://www.mankier.com/3/gettext

Of course we don't actually use gettext, but doing something custom would be confusing at best. Also we could swap out this pure Lua subset of gettext for the real deal any time we wanted in case we wanted to use the fancier features (i.e., [ngettext](https://www.mankier.com/3/ngettext) and friends).

poire-z:
OK, that order still looks a bit unintuitive to me :)
But may be better in your code to stick with the gettext term of "domain", rather than "context" then (dunno, but it became a bit, just a bit, more understandable when I read "domain" on that manpage...)

Hope it will be just to be used for disambiguation, and that we can stick with our classic `_("stuff")` for most things.
How will we know it's already used somewhere else, and that we need to think if we're in the same generic blank domain - or that we need to add one?

Frenzie:
Of course you could swap the order of the arguments in Lua but that'd still be confusing. ;-)

The PO file is extracted using this custom Python script: https://github.com/koreader/koreader-misc/blob/cc6a7a0d81a4f59dba237e57be7b259c00f7799d/gettext/lua_xgettext.py

It works fine, but in retrospect adding support for `[[]]` and such was a bit of a waste of time (if somewhat fun to play around with the regex). And the "upstream" source/inspiration for that script already moved to xgettext back in 2016 or so.

Frenzie:
> But may be better in your code to stick with the gettext term of "domain", rather than "context" then (dunno, but it became a bit, just a bit, more understandable when I read "domain" on that manpage...)

It uses both. `dgettext` suggests domain, `msgctxt` suggests context. ;-) I'll change it.

> How will we know it's already used somewhere else, and that we need to think if we're in the same generic blank domain - or that we need to add one?

Something along these lines (but that's too inclusive):
```bash
find . -name '*.lua' >POTFILES
xgettext -l lua --keyword=_:1c,2 --files-from=POTFILES
```

Then either directly in the generated file or in a program like Poedit, Virtaal, Qt Linguist, etc. you can search for the string to see where it occurs:

Screenshot_20190821_171247

I prefer any of those over Transifex tbh but there's a lower barrier to entrance with Transifex.

Frenzie:
By the by, here's what a domain looks like in Virtaal (it's not very clear):

Screenshot_20190821_172726

The way everything is displayed there it's hard to accidentally miss whitespace. I wonder how many translators accidentally did that wrong just like me. :-P

poire-z:
So, in regular dev, I'll keep just using `_("my wording")` and wait for someone to open an issue and say there are wording conflicts in such language and we need to add a domain to one of the occurence, right? :)

Frenzie:
Most of the time I assume so, yes. There are some instances of single words that might be suspicious all by themselves though. I've noticed some instances in the past, but in Dutch/French/German they worked just well enough to be slightly awkward and not wrong.

Anyway, the full context for this PR stems from a question on Transifex:

https://www.transifex.com/houqp/koreader/translate/#zh_TW/koreader/110323460

cges30901
> Does this mean the number of pages or the layout of pages? 

Frenzie
> @cges30901 Generally the number, but you're right that this lumping together of all strings that are the same (a leftover from the early days) is no longer a good idea. Because one of those mentioned in context is about page style/layout. I'll open a ticket about it on GitHub.

Frenzie:
Hm, I may have gotten domain and context confused in some way. It's not entirely clear to me how to use context in a normal gettext fashion.

glib defines a `C_()` macro.
https://developer.gnome.org/glib/2.28/glib-I18N.html#C-:CAPS

I seem to be using context effectively correctly, except for the part where I'm relating it to dgettext.

Frenzie:
@poire-z Thanks for making me aware of my mistaking context and domain for the same thing.

I've reworked it so that you could do something like this:

```lua
local _ = require("gettext")
local C_ = _.pgettext
```

poire-z:
So:
```lua
_("Scroll") -- when I don't fear anything
C_("rendering", "Scroll") -- in case I fear Scroll might be used elsewhere
C_("rendering", "Pages")
_C("rendering", "Pages") -- looks nice(r) too
```

Frenzie:
Indeed. Where `C_` is whatever we pass in `xgettext --keyword=C_:1c,2`, because the only thing it understands without special instructions is the awfully unwieldy `gettext.pgettext()`.

Having it as `_()` with two arguments would be ideal imo, but the usual implementation of the `_()` convenience function uses:

1 arg = gettext
2 args = dgettext
3 args = ngettext
4 args = dngettext

Frenzie:
Also, reference for contexts: https://www.gnu.org/software/gettext/manual/html_node/Contexts.html

PR #5237, opened 2019-08-21
[i18n] Switch `make pot` to xgettext

### LINKED ISSUE #5232: [i18n] Translation should be context-dependent

Frenzie:
This introduces context to our strings to differentiate them when necessary.

The syntax chosen is `C_()`, following [glib](https://developer.gnome.org/glib/2.28/glib-I18N.html#C-:CAPS).

```lua
local _ = require("gettext")
local C_ = _.pgettext

C_("Menu|StyleTweaks|", "Pages")
```

Closes #5232.

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -14,7 +14,7 @@ local C_ = _.pgettext
 
 local CssTweaks = {
     {
-        title = C_("Menu|StyleTweaks|", "Pages"),
+        title = C_("StyleTweaksCategory", "Pages"),
         {
             id = "margin_body_0";
             title = _("Ignore publisher page margins"),
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -10,10 +10,11 @@ They may have the following optional attributes:
 ]]
 
 local _ = require("gettext")
+local C_ = _.pgettext
 
 local CssTweaks = {
     {
-        title = _("Pages"),
+        title = C_("Menu|StyleTweaks|", "Pages"),
         {
             id = "margin_body_0";
             title = _("Ignore publisher page margins"),
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -14,7 +14,7 @@ local C_ = _.pgettext
 
 local CssTweaks = {
     {
-        title = C_("StyleTweaksCategory", "Pages"),
+        title = C_("Style tweaks category", "Pages"),
         {
             id = "margin_body_0";
             title = _("Ignore publisher page margins"),
--- a/Makefile
+++ b/Makefile
@@ -462,15 +462,15 @@ pocketbook-toolchain:
 # for gettext
 DOMAIN=koreader
 TEMPLATE_DIR=l10n/templates
-KOREADER_MISC_TOOL=../koreader-misc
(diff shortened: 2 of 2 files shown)


frontend/ui/data/css_tweaks.lua
```diff
@@ -10,10 +10,11 @@ They may have the following optional attributes:
 ]]
 
 local _ = require("gettext")
+local C_ = _.pgettext
 
 local CssTweaks = {
     {
-        title = _("Pages"),
+        title = C_("Menu|StyleTweaks|", "Pages"),
```
poire-z: Why not just `C_("StyleTweaks", "Pages")` ?
The | separators  just add room for errors.

And in that context, we define a list of tweak categories and tweaks, which happen to be used as menu items, but I'd rather not see "Menu" mentionned.
In the case there would be a tweak also named "Pages", I'd rather see `C_("StyleTweaksCategory", "Pages")`

Frenzie: It could also just be `Style tweaks category` or `style tweaks category`. We just want to try to avoid descriptive sentences that would be more likely to (want to be) change.

Frenzie:
And Transifex now nicely separates pages the style tweaks category from pages the number of pages. Oddly enough it doesn't seem to prompt for a retranslation with this context change.

Screenshot_2019-08-22_00-15-59
Screenshot_2019-08-22_00-16-25

Frenzie:
A somewhat related xgettext feature that we could use thanks to the switch is the ability to add notes for translators.

You pass a magic string along on the command line:
```
xgettext --add-comments=TRANSLATORS: file.lua
```

```lua
-- TRANSLATORS: a note for translators
_([[
---------------------------------------------
                launching...
  _  _____  ____                _
 | |/ / _ \|  _ \ ___  __ _  __| | ___ _ __
 | ' / | | | |_) / _ \/ _` |/ _` |/ _ \ '__|
 | . \ |_| |  _ <  __/ (_| | (_| |  __/ |
 |_|\_\___/|_| \_\___|\__,_|\__,_|\___|_|

 It's a scroll... It's a codex... It's KOReader!

 [*] Current time: ]])
```

Then you get a result like this:

```
#. TRANSLATORS: a note for translators
#: reader.lua:17
msgid ""
"---------------------------------------------\n"
"                launching...\n"
"  _  _____  ____                _\n"
" | |/ / _ \\|  _ \\ ___  __ _  __| | ___ _ __\n"
" | ' / | | | |_) / _ \\/ _` |/ _` |/ _ \\ '__|\n"
" | . \\ |_| |  _ <  __/ (_| | (_| |  __/ |\n"
" |_|\\_\\___/|_| \\_\\___|\\__,_|\\__,_|\\___|_|\n"
"\n"
" It's a scroll... It's a codex... It's KOReader!\n"
"\n"
" [*] Current time: "
msgstr ""
```

This note will then be shown in translation programs:
Screenshot_2019-08-22_18-25-02

PR #5244, opened 2019-08-23
[doc] Tag @todo, @fixme and @warning

Frenzie:
This commit tries to standardize the various todos around the code a bit in a manner recognized by LDoc.

Besides drawing more attention by being displayed in the developer docs, they're also extractable with LDoc on the command line:

```sh
ldoc --tags todo,fixme *.lua
```

However, whether that particular usage offers any advantage over other search tools is questionable at best.


frontend/apps/reader/modules/readerzooming.lua
```diff
@@ -116,9 +116,6 @@ function ReaderZooming:init()
 end
 
 function ReaderZooming:onReadSettings(config)
-    -- @TODO config file from old code base uses globalzoom_mode
```
Frenzie: This todo from 2012 doesn't sound terribly relevant. Some old viewing configurations from 2011/2012 have been broken for 7 years now? I imagine people updated their settings manually by now… ;-)

NiLuJe:
Does it have to be lowercase? I kinda prefer those tags in all-caps myself ;).

Frenzie:
It does for LDoc, yes. Looks like this:

Screenshot_2019-08-23_16-15-25

(The alternative looks like nothing. ;-) )

Frenzie:
Although if you really want I think it should be possible to define a custom override.

NiLuJe:
Nah, no need to bother on my account ;).

I double-checked, and I do get speciifc syntax highlighting for it (not the one I'm used to, but good enough) ;).

Thanks for checking, though!

Frenzie:
Btw, you can disable it after all if you use some nonsense alias:

```
alias('todo','____')
alias('TODO','todo')
```

Frenzie:
@NiLuJe Btw, know anything like or better than regexxer? This seems to work fairly well to find and review replacements.

Screenshot_2019-08-23_18-25-28

NiLuJe:
@Frenzie : I tend to do this kind of stuff over CLI via sed if it's across multiple files (followed by a VCS diff to double-check).

Frenzie:
@NiLuJe I almost wrote "because I find the cludge of using sed and then checking the diff tedious." ;-D

Issue #3754, opened 2018-03-14
FR: Switch internationalization to Weblate


Frenzie:
* KOReader version: 78fbc14
* Device: all

#### Issue

According to @anarcat it could simplify and improve certain aspects of the workflow. This issue is to track ideas and requirements.

https://github.com/koreader/koreader/issues/2621#issuecomment-372986629

anarcat:
So I was asked to provide a workflow example, so here we go... I'm using Weblate for the [Monkeysign project](https://0xacab.org/monkeysphere/monkeysign/). There is this [web interface](https://hosted.weblate.org/projects/monkeysphere/monkeysign/) where people can sign in and contribute translations. Say I spoke German, I would go to the [german translation page](https://hosted.weblate.org/projects/monkeysphere/monkeysign/de/) and hit `Tranlsate`. There I am shown with a simple form where I can submit translations, but I can also "suggest" them in which case they wait for moderation before being committed. There's also a batch edit mode (called "Zen") where multiple translations can be entered more quickly.

Once a translation is done, it gets committed into git, in a repository that is publicly available. The main developers (ie. me) will, before a release, pull from that repository and merge it in the release branch after a summary review. One thing I find annoying there is that some "empty" translations sometimes shows up: a translator would come to the project and start a (say) Russian translation but never translate any string. Or we'd have partial translations like Ukrainian (3 words?!). It's harmless, but there's a `.po` file there that's just mostly empty. Completely empty .po files do not get committed into git, so it's not such a big deal.

When new code is pushed to the main Monkeysign repository, there is a (web) hook that tells Weblate to pull changes. Weblate merges those in and will notice new strings needing translation and update the web interface accordingly.

Strings can also be translated using normal po tools by editing the files directly in the normal git repository, which was a requirement for us: we didn't want to depend on a web interface.

There's also a [commandline client](https://github.com/WeblateOrg/wlc) to operate the weblate repository from the commandline. From there you can list projects, translations, download translations and so on.

I'm happy to answer any other question you might have about Weblate. Apart from that one annoyance I've found above, weblate has served us very well. It mostly stays out of the way while allowing people to contribute translations fairly easily, and the web interface is always getting better. All the translations we have of Monkeysign (except a *partial* french translation) come from Weblate users.

You can host your own [weblate instance](https://github.com/WeblateOrg/weblate) or you can ask to be hosted for free, since this is a free software project (although they welcome donations). Here are their [hosting plans](https://weblate.org/en/hosting/).

[...]

PR #5250, opened 2019-08-23
[i18n] Add translator notes

Frenzie:
Thanks to <https://github.com/koreader/koreader/pull/5237> we can now  extract the knowledge currently embedded in Transifex and put it directly in our source. This positively affects <https://github.com/koreader/koreader/issues/3754>.

Translation instructions and knowledge that comes out of localization-related questions should be preserved in the source, because Transifex is too ephemeral. For example, the links from <https://github.com/koreader/koreader/pull/2290> are no longer accessible. Even when they are, it's quite useful to have this information around while dealing with the code as well, and I also hope it'll be informative to contributors who seldom visit Transifex.

This commit also makes a few minor changes to obviate the need for comments where possible.

### DIFF
--- a/frontend/apps/filemanager/filemanagersearch.lua
+++ b/frontend/apps/filemanager/filemanagersearch.lua
@@ -79,7 +79,7 @@ function Search:getCalibre()
         logger.dbg("search Calibre database")
         self.metafile_1 = findcalibre("/mnt")
         if not self.metafile_1 then
-          self.error = _("SEARCH_LIBRARY_PATH should be defined in DEFAULTS.LUA.")
+            self.error = _("The SEARCH_LIBRARY_PATH variable must be defined in 'persistent.defaults.lua' in order to use the calibre file search functionality.")
         end
     else
         if string.sub(SEARCH_LIBRARY_PATH, string.len(SEARCH_LIBRARY_PATH)) ~= "/" then
@@ -189,6 +189,7 @@ function Search:ShowSearch()
                         end,
                     },
                     {
+                        -- @translators Search for books in calibre catalog.
                         text = _("Find books"),
                         enabled = true,
                         callback = function()
@@ -494,7 +495,7 @@ function Search:onMenuHold(item)
         item.info = item.info .. item.path
         local f = io.open(item.path, "r")
         if f == nil then
-            item.info = item.info .. "\n" .. _("File not found!")
+            item.info = item.info .. "\n" .. _("File not found.")
         else
             item.info = item.info .. "\n" .. _("Size:") .. " " .. string.format("%4.1fM", lfs.attributes(item.path, "size")/1024/1024)
             f:close()
--- a/frontend/ui/data/strings.lua
+++ b/frontend/ui/data/strings.lua
@@ -14,14 +14,17 @@ S.B_PAGE_MARGIN = _("Bottom Margin")
 S.SYNC_T_B_PAGE_MARGINS = _("Sync T/B Margins")
 S.LINE_SPACING = _("Line Spacing")
 S.COLUMNS = _("Columns")
-S.TEXT_ALIGN = _("Text Align")
+-- @translators Text alignment. Options given as icons: left, right, center, justify.
+S.TEXT_ALIGN = _("Alignment")
 S.FONTSIZE_FINE_TUNING = _("Fine Tuning")
 S.CONTRAST = _("Contrast")
+-- @translators Reflow text.
 S.REFLOW = _("Reflow")
 S.DEWATERMARK = _("Dewatermark")
 S.DOC_LANG = _("Document Language")
 S.VERTICAL_TEXT = _("Vertical Text")
 S.WORD_GAP = _("Word Gap")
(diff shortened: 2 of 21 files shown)


frontend/ui/data/strings.lua
```diff
@@ -14,14 +14,17 @@ S.B_PAGE_MARGIN = _("Bottom Margin")
...
+S.TEXT_ALIGN = _("Alignment")
 S.FONTSIZE_FINE_TUNING = _("Fine Tuning")
 S.CONTRAST = _("Contrast")
+-- @translators Reflow text.
 S.REFLOW = _("Reflow")
 S.DEWATERMARK = _("Dewatermark")
 S.DOC_LANG = _("Document Language")
 S.VERTICAL_TEXT = _("Vertical Text")
 S.WORD_GAP = _("Word Gap")
+-- @translators The maximum size of a speckle to be ignored instead of being considered a character.
```
Frenzie: This could probably be a little clearer.

k2pdfopt for reference:
```
-de <size>        Defect size in points.  For scanned documents, marks
                  or defects smaller than this size are ignored when bounding
                  rectangular regions.  The period at the end of a sentence is
                  typically over 1 point in size.  The default is 1.0.
```

frontend/ui/translator.lua
```diff
@@ -24,6 +24,7 @@ local _ = require("gettext")
 -- 20181217: 104 supported languages
 local AUTODETECT_LANGUAGE = "auto"
 local SUPPORTED_LANGUAGES = {
+    -- @translators Many of the names for languages can be conveniently found pre-translated in the relevant language of this Wikipedia article: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
```
Frenzie: This is just a semi-random sprinkling in the hope that translators see it. Doing it on every line seemed a bit much.

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -429,7 +430,14 @@ function DictQuickLookup:update()
                 {
                     -- if dictionary result, do the same search on wikipedia
                     -- if already wiki, get the full page for the current result
-                    text = self.is_wiki and _("Wikipedia full") or _("Wikipedia"),
+                    text_func = function()
+                        if self.is_wiki then
+                            -- @translators Full Wikipedia article.
```
Frenzie: This actually attaches itself to the first extracted string if you leave the line as is, but I figured it was better to be abundantly clear about it.

frontend/apps/filemanager/filemanagersearch.lua
```diff
@@ -79,7 +79,7 @@ function Search:getCalibre()
         logger.dbg("search Calibre database")
         self.metafile_1 = findcalibre("/mnt")
         if not self.metafile_1 then
-          self.error = _("SEARCH_LIBRARY_PATH should be defined in DEFAULTS.LUA.")
+            self.error = _("The SEARCH_LIBRARY_PATH variable must be defined in 'persistent.defaults.lua' in order to use the calibre file search functionality.")
```
NiLuJe: :+1:

frontend/util.lua
```diff
@@ -153,6 +153,7 @@ function util.secondsToHClock(seconds, withoutSeconds, hmsFormat)
                 mins = string.format("%.f", round(seconds / 60))
                 return mins .. "'"
             end
+            -- @translators This is the 'h' for hour, like in 1h30.
```
NiLuJe: I'm maybe specify that it's a duration?

plugins/systemstat.koplugin/main.lua
```diff
@@ -112,10 +113,13 @@ function SystemStat:appendSystemInfo()
     local stat = systemInfo()
     if stat.cpu ~= nil then
         self:put({_("System information"), ""})
+        -- @translators Ticks is a highly technical term. See https://superuser.com/a/101202 The correct translation is likely to simply be "ticks".
         self:put({_("  Total ticks (million)"),
                  string.format("%.2f", stat.cpu.total / 1000000)})
+        -- @translators Ticks is a highly technical term. See https://superuser.com/a/101202 The correct translation is likely to simply be "ticks".
         self:put({_("  Idle ticks (million)"),
                  string.format("%.2f", stat.cpu.idle / 1000000)})
+        -- @translators Ticks is a highly technical term. See https://superuser.com/a/101202 The correct translation is likely to simply be "ticks".
```
NiLuJe: That one appears to be erroneous? (it's a a %, not a tick count).

frontend/ui/data/strings.lua
NiLuJe: I kind of like it? Maybe I'd just qualify "speckle" (dust/ink?)

Frenzie:
Lokalize is KDE's tool. This completion visualization is interesting:

Screenshot_20190824_000139

The tool as a whole looks very useful.

Frenzie:
And here's how it looks on Transifex:

Screenshot_2019-08-24_09-41-32

PR #5252, opened 2019-08-24
[Android] Don't package Droid Sans

Frenzie:
This is preinstalled on every Android device. It's only 105.6 KiB, but every space reduction is a good one.

Cf. <https://github.com/koreader/koreader/pull/5200#issuecomment-524501845>.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -347,9 +347,10 @@ androidupdate: all
 	# in runtime luajit-launcher's libluajit.so will be loaded
 	-rm $(INSTALL_DIR)/koreader/libs/libluajit.so
 
-        # assets are compressed manually and stored inside the APK.
+	# assets are compressed manually and stored inside the APK.
 	cd $(INSTALL_DIR)/koreader && zip -r9 \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).zip * \
+		--exclude=*fonts/droid* \
 		--exclude=*resources/fonts* \
 		--exclude=*resources/icons/src* \
 		--exclude=*share/man* \
--- a/frontend/ui/elements/font_settings.lua
+++ b/frontend/ui/elements/font_settings.lua
@@ -6,7 +6,10 @@ local _ = require("gettext")
 --[[ Font settings for desktop linux and mac ]]--
 
 local function getUserDir()
-    return os.getenv("HOME").."/.local/share/fonts"
+    local home = os.getenv("HOME")
+    if home then
+        return home.."/.local/share/fonts"
+    end
 end
 
 -- System fonts are common in linux
@@ -25,9 +28,10 @@ local function usesSystemFonts()
 end
 
 local function openFontDir()
+    if not Device:canOpenLink() then return end
     local user_dir = getUserDir()
     local openable = util.pathExists(user_dir)
-    if not openable then
+    if not openable and user_dir then
         logger.info("Font path not found, making one in ", user_dir)
         openable = util.makePath(user_dir)
     end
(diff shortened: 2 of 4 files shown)


frontend/ui/elements/font_settings.lua
```diff
@@ -6,7 +6,8 @@ local _ = require("gettext")
 --[[ Font settings for desktop linux and mac ]]--
 
 local function getUserDir()
-    return os.getenv("HOME").."/.local/share/fonts"
+    local home = os.getenv("HOME") or ""
+    return home.."/.local/share/fonts"
```
Frenzie: @pazos This workaround isn't quite right obviously, except for preventing a crash.

Also, http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html#tag_08_03

> The system shall initialize this variable at the time of login to be a pathname of the user's home directory. See <pwd.h>.

I'm not entirely sure what's happening here since it should just exist. But it's not related to this PR other than that this code path wasn't hit during the unit tests until now.

frontend/ui/elements/font_settings.lua
```diff
@@ -6,7 +6,10 @@ local _ = require("gettext")
 --[[ Font settings for desktop linux and mac ]]--
 
 local function getUserDir()
-    return os.getenv("HOME").."/.local/share/fonts"
+    local home = os.getenv("HOME")
```
Frenzie: @pazos The unit test doesn't behave exactly right for Kobo because `HOME` is actually set to `/`, but regardless it doesn't seem to make much sense to do all this unless Device:canOpenLink() is true? I adjusted `openFontDir()` accordingly.

frontend/ui/font.lua
```diff
@@ -2,10 +2,11 @@
 Font module.
 ]]
 
+local FontList = require("fontlist")
 local Freetype = require("ffi/freetype")
-local logger = require("logger")
 local Screen = require("device").screen
-local FontList = require("fontlist")
+local logger = require("logger")
+
```
Frenzie:
```suggestion
```

Makefile
```diff
@@ -341,7 +341,8 @@ androidupdate: all
 	# make android update apk
 	cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=1 \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).7z * \
-		-x!resources/fonts -x!resources/icons/src -x!spec
+		-x!resources/fonts -x!resources/icons/src -x!spec \
+		-x!fonts/droid
```
Frenzie: Similarly, the deb package should skip Droid (?) and Noto. `fonts-noto-core` can be a dependency and boom, half size package.

pazos: In that case we need to remove the "enable system fonts" toggle.

NiLuJe: @pazos : Err, I'm not sure I follow? (Or did you mean we need to enforce reading system fonts in that case?)

pazos: > we need to enforce reading system fonts in that case?

Exactly.

Frenzie:
[ ] Wait for https://github.com/koreader/koreader/pull/5264
[ ] Enforce system fonts on Android

pazos:
@Frenzie: system fonts are already used in Android. See https://github.com/koreader/koreader/blob/master/defaults.lua#L237

Frenzie:
@pazos I know, but I thought you'd added a toggle for it. But that was only for Mac/Linux. :-)

Makefile
```diff
@@ -347,9 +347,10 @@ androidupdate: all
 	# in runtime luajit-launcher's libluajit.so will be loaded
 	-rm $(INSTALL_DIR)/koreader/libs/libluajit.so
 
-        # assets are compressed manually and stored inside the APK.
+  # assets are compressed manually and stored inside the APK.
```
Frenzie: Oh ffs GH online editor. :man_facepalming:  Anyway, still to be tested; I don't have Android stuff on this laptop.

Issue #5249, opened 2019-08-23
FR: Localization and plurals


Frenzie:
* KOReader version: 2019.08
* Device: Transifex

#### Issue
Ideally we'd implement ngettext as well, see https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html and https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html

Now that we've switched to xgettext extraction is available right out of the box. I'm not entirely sure how important this is, or in other words how often grammatical constructions of the "one, two, many" type occur as opposed to our split in singular and plural. Also I'm not sure how well GUIs expose this functionality.

Example comment [on Transifex](https://www.transifex.com/houqp/koreader/translate/#nl_NL/koreader/37345969?q=comment%3A*):
> in case of
Cardinal number + noun
the noun remains in its singular form.
by nbuz1, 5 months ago 

> @nbuz1 There's a separate string "1 item" for that scenario. Ideally we would be a little bit more clever about that (see https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html) but for the basic singular/plural it should usually be fine.
by Frenzie, a few seconds ago

Frenzie:
I'm investigating if we could roll out our own ngettext implementation instead of building gettext for all platforms, though it wouldn't be the worst if we did. The first step of getting the header is simple enough:

```diff
diff --git a/frontend/gettext.lua b/frontend/gettext.lua
index ec4ec662..0159a0db 100644
--- a/frontend/gettext.lua
+++ b/frontend/gettext.lua
@@ -62,6 +62,15 @@ function GetText_mt.__index.changeLang(new_lang)
         local line = po:read("*l")
         if line == nil or line == "" then
             if data.msgid and data.msgstr and data.msgstr ~= "" then
+                -- header
+                if data.msgid == "" then
+                    local plural_forms = data.msgstr:match("Plural%-Forms: (.*);")
+                    print(plural_forms)
+                    -- plural handling function
+                    error()
+                end
+
+                -- translated string
                 local unescaped_string = string.gsub(data.msgstr, "\\(.)", c_escape)
                 if data.msgctxt and data.msgctxt ~= "" then
                     if not GetText.context[data.msgctxt] then
```

Looks something like this for many languages:
```
nplurals=2; plural=(n != 1)
```

I think Russian may be the most complex language we have:
```
[...]

PR #5257, opened 2019-08-24
[feat, i18n] Implement ngettext

Frenzie:
Fixes <https://github.com/koreader/koreader/issues/5249>.

Looks a little something like this:
Screenshot_2019-08-24_18-58-11
Screenshot_2019-08-24_18-58-18

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -467,7 +467,7 @@ XGETTEXT_BIN=xgettext
 pot:
 	mkdir -p $(TEMPLATE_DIR)
 	$(XGETTEXT_BIN) --from-code=utf-8 \
-		--keyword=C_:1c,2 --keyword=N_:1,2 \
+		--keyword=C_:1c,2 --keyword=N_:1,2 --keyword=NC_:1c,2,3 \
 		--add-comments=@translators \
 		reader.lua `find frontend -iname "*.lua"` \
 		`find plugins -iname "*.lua"` \
--- a/Makefile
+++ b/Makefile
@@ -466,7 +466,8 @@ XGETTEXT_BIN=xgettext
 
 pot:
 	mkdir -p $(TEMPLATE_DIR)
-	$(XGETTEXT_BIN) --from-code=utf-8 --keyword=C_:1c,2 \
+	$(XGETTEXT_BIN) --from-code=utf-8 \
+		--keyword=C_:1c,2 --keyword=N_:1,2 \
 		--add-comments=@translators \
 		reader.lua `find frontend -iname "*.lua"` \
 		`find plugins -iname "*.lua"` \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -466,7 +466,8 @@ XGETTEXT_BIN=xgettext
 
 pot:
 	mkdir -p $(TEMPLATE_DIR)
-	$(XGETTEXT_BIN) --from-code=utf-8 --keyword=C_:1c,2 \
+	$(XGETTEXT_BIN) --from-code=utf-8 \
+		--keyword=C_:1c,2 --keyword=N_:1,2 \
```
Frenzie: For `npgettext`, given that we're using `C_` for `pgettext`, logically that should be `NC_`. I didn't really test that one yet but other than the lacking npgettext method it should be working.

frontend/gettext.lua
```diff
@@ -57,20 +145,43 @@ function GetText_mt.__index.changeLang(new_lang)
...
-                else
-                    GetText.translation[data.msgid] = unescaped_string
                 end
+            elseif data.msgid and data.msgstr and data.msgstr ~= "" then
+                -- header
+                if not headers and data.msgid == "" then
+                    local util = require("util")
+                    headers = data.msgstr
+                    local plural_forms = data.msgstr:match("Plural%-Forms: (.*);")
+                    local nplurals = plural_forms:match("nplurals=([0-9]+);") or 1
```
Frenzie: 2 actually seems to be right; I'll have to look at the docs a bit better. Clearly you don't have to declare the default.

Frenzie: Oh, I guess perhaps I can't check for sanity because a C boolean expression is 0 or 1.

Makefile
poire-z: But `CN_` would state the order: first context, then the next thing will be N'ed :)
(if we don't need to care about the involved gettext function names - but you decide)

Frenzie: For reference, GTK also has an NC_ (although it means something completely different): https://developer.gnome.org/glib/stable/glib-I18N.html#NC-:CAPS

Anyway, I don't even know if we need `npgettext`. It's just with all the groundwork already there it seems silly not to add the method.

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -381,7 +382,7 @@ function ListMenuItem:update()
                 end
             else
                 if pages then
-                    pages_str = T(_("%1 pages"), pages)
+                    pages_str = T(N_("1 page", "%1 pages", pages), pages)
```
Frenzie: @poire-z Like so:

Screenshot_2019-08-24_23-00-42


Screenshot_2019-08-24_22-58-43

Frenzie:
Merging now so the translations can start rolling in; code should work as designed per-spec but we'll see if anything funny crops up in the wild. ;-)

Frenzie:
And it looks like this on Transifex:

Screenshot_2019-08-24_23-21-21

PR #5262, opened 2019-08-25
[i18n, chore] Some translation stuff

Frenzie:
See https://github.com/koreader/koreader/pull/5000#discussion_r317392259

Btw @NiLuJe is that code path I reference in the second commit even still used?


NiLuJe:
Yes, when setting/viewing defaults on the "Horizontal margins" (it's still a table of two identical values, for, err, reasons? :D).

Frenzie:
Oh right, now I see. :-)

PR #5264, opened 2019-08-25
migrate from lzma to zip file.

pazos:
Requires https://github.com/koreader/android-luajit-launcher/pull/174 ~~and https://github.com/koreader/android-luajit-launcher/pull/175~~

The main benefit of having the libraries outside the zip file is that they will be installed when installing the application instead of copying them on the first run (it could save a few seconds)

**Edit:** I'm leaving the libs thingy for the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,11 +346,24 @@ androidupdate: all
 	-rm $(ANDROID_LAUNCHER_DIR)/assets/module/koreader-*
 	# in runtime luajit-launcher's libluajit.so will be loaded
 	-rm $(INSTALL_DIR)/koreader/libs/libluajit.so
-	# make android update apk
-	cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=1 \
-		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).7z * \
-		-x!resources/fonts -x!resources/icons/src -x!spec
-	$(MAKE) -C $(ANDROID_LAUNCHER_DIR) $(if $(KODEBUG), debug, release) ANDROID_APPNAME=KOReader ANDROID_VERSION=$(ANDROID_VERSION) ANDROID_NAME=$(ANDROID_NAME) ANDROID_FLAVOR=$(ANDROID_FLAVOR)
+
+        # assets are compressed manually and stored inside the APK.
+	cd $(INSTALL_DIR)/koreader && zip -r9 \
+		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).zip * \
+		--exclude=*resources/fonts* \
+		--exclude=*resources/icons/src* \
+		--exclude=*share/man* \
+		--exclude=*spec* \
+		--exclude=*COPYING* \
+		--exclude=*README.md*
+
+	# make the android APK
+	$(MAKE) -C $(ANDROID_LAUNCHER_DIR) $(if $(KODEBUG), debug, release) \
+		ANDROID_APPNAME=KOReader \
+		ANDROID_VERSION=$(ANDROID_VERSION) \
+		ANDROID_NAME=$(ANDROID_NAME) \
+		ANDROID_FLAVOR=$(ANDROID_FLAVOR)
+
 	cp $(ANDROID_LAUNCHER_DIR)/bin/NativeActivity.apk \
 		koreader-android-$(ANDROID_ARCH)$(KODEDUG_SUFFIX)-$(VERSION).apk
 
--- a/kodev
+++ b/kodev
@@ -9,15 +9,7 @@ fi
 
 # Default Android build to arm.
 ANDROID_ARCH="${ANDROID_ARCH:-arm}"
-if [ -z "${ANDROID_FULL_ARCH}" ]; then
-    if [ "${ANDROID_ARCH}" = "arm" ]; then
-        ANDROID_FULL_ARCH_APK="${ANDROID_FULL_ARCH_APK:-arm-linux-androideabi}"
-    elif [ "${ANDROID_ARCH}" = "x86" ]; then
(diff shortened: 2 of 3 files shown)


Frenzie:
Also needs:

```diff
diff --git a/kodev b/kodev
index e44e064b..4f8b7d15 100755
--- a/kodev
+++ b/kodev
@@ -9,15 +9,7 @@ fi
 
 # Default Android build to arm.
 ANDROID_ARCH="${ANDROID_ARCH:-arm}"
-if [ -z "${ANDROID_FULL_ARCH}" ]; then
-    if [ "${ANDROID_ARCH}" = "arm" ]; then
-        ANDROID_FULL_ARCH_APK="${ANDROID_FULL_ARCH_APK:-arm-linux-androideabi}"
-    elif [ "${ANDROID_ARCH}" = "x86" ]; then
-        ANDROID_FULL_ARCH_APK="${ANDROID_FULL_ARCH_APK:-i686-linux-android}"
-    else
-        ANDROID_FULL_ARCH_APK="${ANDROID_ARCH}"
-    fi
-fi
+
 # Default to Android 4.0+; required for NDK 15 but with a custom NDK the strict minimum is 9.
 NDKABI=${NDKABI:-14}
 export NDKABI
@@ -625,7 +617,7 @@ TARGET:
                 # uninstall existing package to make sure *everything* is gone from memory
                 # no assert_ret_zero; uninstall is allowed to fail if there's nothing to uninstall
                 adb uninstall "org.koreader.launcher"
-                adb install "koreader-android-${ANDROID_FULL_ARCH_APK}${KODEBUG_SUFFIX}-${VERSION}.apk"
+                adb install "koreader-android-${ANDROID_ARCH}${KODEBUG_SUFFIX}-${VERSION}.apk"
                 assert_ret_zero $?
                 # there's no adb run so we do this…
                 adb shell monkey -p org.koreader.launcher -c android.intent.category.LAUNCHER 1
```

Frenzie:
@pazos I seem to be missing some libraries in my APK:

```
root@generic_x86:/data/user/0/org.koreader.launcher/lib # ls                   
libblitbuffer.so
libcrengine.so
libkoreader-cre.so
libkoreader-djvu.so
libkoreader-lfs.so
liblodepng.so
libluacompat52.so
libluajit.so
liblualongnumber.so
libmupdf.so
libsqlite3.so
libturbojpeg.so
libwrap-mupdf.so
```

But they also exist here? Bit confused tbh.

```
root@generic_x86:/data/app/org.koreader.launcher-1/lib/x86 # ls
libblitbuffer.so
libcrengine.so
libkoreader-cre.so
libkoreader-djvu.so
libkoreader-lfs.so
liblodepng.so
libluacompat52.so
libluajit.so
liblualongnumber.so
libmupdf.so
libsqlite3.so
libturbojpeg.so
libwrap-mupdf.so
```

Frenzie:
Gradle is of course total trash:

https://android.googlesource.com/platform/tools/build/+/master/builder/src/main/java/com/android/builder/internal/packaging/Packager.java#58

pazos:
>  seem to be missing some libraries in my APK

Ops, I didn't notice. Good catch!

> But they also exist here? Bit confused tbh

`/data/user/x/something` should be a symbolic link to `/data/app/something` in multiuser environments.

Frenzie:
Oh, you're right. I hadn't noticed.

```
root@generic_x86:/data/user/0/org.koreader.launcher # ls -l
drwxrwx--x u0_a159  u0_a159           2019-08-26 21:17 cache
drwxrwx--x u0_a159  u0_a159           2019-08-26 21:17 code_cache
drwxrwx--x u0_a159  u0_a159           2019-08-26 21:17 files
lrwxrwxrwx root     root              2019-08-26 21:17 lib -> /data/app/org.koreader.launcher-1/lib/x86
```

pazos:
So the change from lzma to zip increases the assets 3'1MB on release builds. (Lastest x86 nightly has a 7z package of 32,9MB while the same assets compressed as a zip take 36MB)

There are a few micro optimizations that can save us a few KB:

- enable proguard (reduces classes.dex from 1.4MB to 300KB and the whole APK about 300kb).
- remove a few unused bits (man pages, README and COPYING...)  aprox another 300kb of size reduction.

I think we need to take a decission about Noto:

NotoSansCJKsc-Regular has a size of 17,3MB. All the rest of Noto fonts have a size of 4MB together. If we want to keep CJK support in place we might to remove the rest of Noto fonts and use Roboto instead.

Probably all phones with 4.0+ on CJK countries have NotoSans/NotoSerif CJK available as a system font. This is difficult to check for me :/

NiLuJe:
@pazos: At worst we've established that anything on earth should have Droid Fallback, so, eh, that's good enough for me.

Frenzie:
Yeah, older Androids have Droid Fallback until Noto showed up. The only exception would be if a manufacturer removed the default fonts, which would be pretty crazy but then again it's Android.

One can of course always just grab the fonts from https://github.com/koreader/koreader-fonts/tree/master/noto and stick 'em in `/sdcard/koreader/fonts`.

pazos:
@Frenzie: as the libs thingy is going to take some time I think we can move to zip and do the fonts thing before.

Frenzie:
Sure, just merge the zip migration when ready and then I can look into the font thing tomorrow.

poire-z:
Just a note, not sure it's related, but letting you know.
I updated on android from a version from last weekend (manual download of the nightly apk).
Then, on subsequent launches, I got the black screen with the rolling thingy, that I usually get only once, the first launch after the update. Today, I get it on each KOreader launch.

I then just uninstalled koreader and installed that same nightly: got the rolling thingy only once - no more on each launch.

(Dunno if we need/can do something, users will get that if they don't think about reinstalling.)

pazos:
@poire-z: it happened to me a few times even before the change.

We shouldn't show the dialog unless we are extracting something.

Issue #5269, opened 2019-08-26
Cannot be relaunched after update on KPW3-32G


zivilyyyn:
* KOReader version: v2019.08-51-gba8099f_2019-08-25
* Device: KPW3-32G

##### `crash.log` 
[crash.log](https://github.com/koreader/koreader/files/3541245/crash.log)

Frenzie:
Likely caused by https://github.com/koreader/koreader/pull/5257 (or at least something related).

Which language setting are you using?

zivilyyyn:
chinese

Frenzie:
Neither zh_CN nor zh_TW cause a crash for me, so at least it's (probably) not language specific. I wonder why you're getting a circular dependency though, and I'm not…

Working around it with some code restructuring should be easy enough.

PR #5270, opened 2019-08-26
[fix] Remove gettext util dependency

Frenzie:
A simple `string.gmatch()` should do the trick here, which avoids the need for refactoring.

Fixes <https://github.com/koreader/koreader/issues/5269>.

### DIFF
--- a/frontend/gettext.lua
+++ b/frontend/gettext.lua
@@ -196,7 +196,6 @@ function GetText_mt.__index.changeLang(new_lang)
             elseif data.msgid and data.msgstr and data.msgstr ~= "" then
                 -- header
                 if not headers and data.msgid == "" then
-                    local util = require("util")
                     headers = data.msgstr
                     local plural_forms = data.msgstr:match("Plural%-Forms: (.*);")
                     local nplurals = plural_forms:match("nplurals=([0-9]+);") or 2
@@ -212,7 +211,10 @@ function GetText_mt.__index.changeLang(new_lang)
                         plurals = GetText.plural_default
                     end
 
-                    local pl_tests = util.splitToArray(plurals, " : ")
+                    local pl_tests = {}
+                    for pl_test in plurals:gmatch("[^:]+") do
+                        table.insert(pl_tests, pl_test)
+                    end
 
                     GetText.getPlural = getPluralFunc(pl_tests, nplurals, GetText.plural_default)
                     if not GetText.getPlural then


Frenzie:
Unit tests are happy, so merging now since this is high impact.

NiLuJe:
My go to answer would be the weird fuse layer on the Kindles, but who knows
;).

NiLuJe:
If it still fails after that, IIRC there's also a weird LD_PRELOAD hack on Kindles that speacifically deals with locales, so, that might be screwing with us. But I don't even remember if we're subject to that or not.

Frenzie:
I removed util from gettext so a circular dependency would be impossible. In retrospect the problem should be simple to reproduce with an environment variable (or adding a line calling changeLang around there):

https://github.com/koreader/koreader/blob/cddca1a5b556860256586ae4173f2fec081caf15/frontend/gettext.lua#L361-L369

"Normally" the language isn't changed until just a touch later, avoiding a circular dependency.

https://github.com/koreader/koreader/blob/888d3591b29259884a4d87f0386e0e31154967d3/reader.lua#L31-L35

Frenzie:
Yup, a few commits ago:

```
LANGUAGE=de ./kodev run

[…]

[*] Running KOReader with arguments: ...
~/src/koreader/koreader-emulator-x86_64-linux-gnu-debug/koreader ~/src/koreader
---------------------------------------------
                launching...
  _  _____  ____                _
 | |/ / _ \|  _ \ ___  __ _  __| | ___ _ __
 | ' / | | | |_) / _ \/ _` |/ _` |/ _ \ '__|
 | . \ |_| |  _ <  __/ (_| | (_| |  __/ |
 |_|\_\___/|_| \_\___|\__,_|\__,_|\___|_|

 It's a scroll... It's a codex... It's KOReader!

 [*] Current time: 08/26/19-17:26:56
 [*] Version: v2019.08-48-gece884a3_2019-08-25

./luajit: frontend/util.lua:7: loop or previous error loading module 'gettext'
stack traceback:
        [C]: in function 'require'
        frontend/util.lua:7: in main chunk
        [C]: in function 'require'
        frontend/gettext.lua:199: in function 'changeLang'
        frontend/gettext.lua:360: in main chunk
        [C]: in function 'require'
        ./reader.lua:32: in main chunk
        [C]: at 0x556a7d75b170
~/src/koreader
```

Issue #2431, opened 2016-12-16
Allow transitioning from *sleep* to *power off*


baskerville:
Some users might put the device to sleep by mistake and then try to power it off directly.

This will currently fail.

robert00s:
I think this is done by @poire-z in #3572. You can set deferent settings for sleep and power off.

poire-z:
Not sure this issue is about screensaver.
I guess it's about, after having put the device to sleep with a short press on the power button, to be able to power it off with a 3-seconds press.
Currently, one have to short press again to wake it up, and the long-press it to power it off (which does not sounds too complicated to me :)

robert00s:
@poire-z  But can it be handled by KOReader?
@baskerville What did you mean exactly?

poire-z:
On Kobo, it is handled by KOreader:
https://github.com/koreader/koreader/blob/0d66ea7555accfd0c231f04d7754d85d84fed6e2/frontend/ui/uimanager.lua#L87-L107
https://github.com/koreader/koreader/blob/0d66ea7555accfd0c231f04d7754d85d84fed6e2/frontend/ui/uimanager.lua#L55-L65

So, solving this issue would need some trickery around these (but I won't try it :)

robert00s:
Ok :) I don't have Kobo so I can't help.

avsej:
@Frenzie @poire-z @pazos This patch breaks power button on Cervantes! Is it possible to keep kobo-specific fixes separately?

Frenzie:
@avsej Could you be a bit more specific? This is Kobo-specific code.

NiLuJe:
@Frenzie : From a very quick glance, I *think* the Cervantes/Sony blocks in uimanager rely on the "Resume" event, which went poof with this ;).

Frenzie:
Indeed. We've got a case of confusing design or just a quick hack from years ago spreading out like an oil spill here. :-P

pazos:
> We've got a case of confusing design or just a quick hack from years ago spreading out like an oil spill here. :-P

Indeed. I might be still confused now as I don't understand why this code path should be reachable only for kobos.

Frenzie:
The split in Press/Release is fine and generic enough in theory, but deciding that "Resume" is the correct action to take based on that input event is UIManager logic infiltrating in the Input module.

Back a few years ago this lack of separation was "Kobo only" and I just hadn't quite caught up to that fact. ;-)

avsej:
[...]

PR #5280, opened 2019-08-29
[feat, Kobo] Implement shutdown from suspend

Frenzie:
This allows to shutdown straight from suspend just like Nickel.

Fixes <https://github.com/koreader/koreader/issues/2431>.


Frenzie:
I was about to close it as won't fix when I noticed Nickel does it that way. I'm not completely convinced it makes sense either, but it seems relatively harmless because who's going to accidentally press the button for 2 seconds? So then I wondered how hard it could be anyway, and 15 minutes later there was this PR. (It would've taken 2 minutes but restarting the device and all takes a lot of extra time.)

NiLuJe:
My thoughts exactly ;).

Issue #5054, opened 2019-06-02
[FR] Option to focus on hits on previous pages after fulltext search


cemmilad:
* KOReader version: v2019.05
* Device: Kobo Clara HD 

#### Issue

Now, fulltext search in a document focuses on hits on the next pages. An option to focus on the _previous_ pages instead, would allow for the fulltext search to be used to look up previous sections of the document without possibly spoiling the story or narrative.

#### Steps to reproduce
* Locate a page so there is at least one hit on a page before and one hit on a page after this one.
* Fulltext search for the corresponding query.

Now, the page with at least one hit that is displayed, is located after the original page.

PR #5283, opened 2019-08-30
[feat] Implement fulltext search direction

Frenzie:
Closes <https://github.com/koreader/koreader/issues/5054>.

### DIFF
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -3,6 +3,7 @@ local InputContainer = require("ui/widget/container/inputcontainer")
 local UIManager = require("ui/uimanager")
 local logger = require("logger")
 local _ = require("gettext")
+local C_ = require("gettext").pgettext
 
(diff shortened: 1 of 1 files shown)


poire-z:
Wouldn't `backward | forward` sound nicer than `previous | next` in this context (as there is no hit yet, that could have a previous or a next hit) ?

Frenzie:
Or just the same ⯈ type arrows as in the dictionary? I didn't really pay attention but the search is also grossly abusing `<` `>` there instead of something slightly nicer.

PS It's not that specific ⯈ character but a "white triangle" or something.

poire-z:
Yep, those symbols would look fine there too :)

Frenzie:
Not quite there yet. It needs a different `|`-like character.

Screenshot_20190830_130543
Screenshot_20190830_130556

Frenzie:
I think left/right one eighth block look pretty decent.

Screenshot_20190830_131325

poire-z:
Kinda thick, but so are front and back covers :) so fine with me.

Issue #4983, opened 2019-04-29
RTL works in Cool Reader but there is no menu selection for it


Svahnen:
* KOReader version: 2019.04-60
* Device: Kobo Aura e2

#### Issue
RTL option is remembered and works in Cool Reader engine from having used it in MuPDF engine, but there is no menu setting for turning it on/off when using Cool Reader engine.
#### Steps to reproduce
Open any epub in MuPDF and turn on "Read from right to left", then go back to the file browser and open in Cool Reader engine. The result will be that the setting still applies but can not be turned off unless you open the file using MuPDF again.

I wish this setting would be available in Cool Reader engine as well since it seems to be able to handle RTL.

Issue #4089, opened 2018-07-19
default reading direction


Gil-boy:
* KOReader version: latest nightly v2015.11-1782-g4646721_2018-07-16
* Device: kobo aura one

#### Issue
 can you add a dfeature to change default reading direction  in the menu  when you long press from  on reading direction # right to left#

NiLuJe:
IIRC, a lot of stuff barely supports RTL (Us, CRe, I think only MuPDF does a bit).

NiLuJe:
Fold w/ #4088

Issue #4446, opened 2019-01-05
Enable inverted button mapping on Forma


alethiophile:
* KOReader version: v2018.12-44-g41b885f_2019-01-03
* Device: Kobo Forma

#### Issue
On the Forma, I find the "default" (bottom next, top previous) button layout basically unusable. Nickel offers an option to invert the sense of the buttons, but Koreader doesn't.

(I could always just hack the mapping file manually for this one device, but that's annoying and fragile.)

#### Steps to reproduce
1. Invert button mapping in Nickel
2. Run Koreader
3. Notice buttons are still in default mapping, cringe with anticipated RSI

Frenzie:
> (I could always just hack the mapping file manually for this one device, but that's annoying and fragile.)

Annoying perhaps, but fragile depends on how you hack it. ;-)

I think I might have a branch around somewhere with simple user event map support; I'll have to remember to look for it.

poire-z:
There is also (already implemented) the possibility to override keymap by creating `koreader/settings/event_map.lua` (see mine for the emulator in https://github.com/koreader/koreader/pull/3862#issue-181100453).

If the default keymap is:
https://github.com/koreader/koreader/blob/ea424e621ea88ef0519e86112b0373adb5b6efa2/frontend/device/kobo/device.lua#L240-L250

creating `koreader/settings/event_map.lua` with these 2 keys inverted might work:
```
return {
            [194] = "RPgBack",
            [193] = "RPgFwd",
 }
```

NiLuJe:
Is this documented anywhere? (pinging the great Wiki wizard @AlanSP1 ;)).

poire-z:
I guess that if that works, it could just be made into a menu item in `Navigation> Invert next/prev buttons` (shown on device with keys), and the inversion of `193` and `194` taken care in frontend/device/kobo/device.lua where `G_reader_settings` is available and already used (for `kobo_touch_switch_xy`).

AlanSP1:
> Is this documented anywhere? (pinging the great Wiki wizard @AlanSP1 ;)).

I didn't documented it (before). And I didn't know that there's possibility for keymap editing, as I have H2O and it doesn't have physical keys.

I'm going to create document now:

NiLuJe:
I was thinking about doing it in kobo/device, but it's a tiny bit clunky because the mapping is initalized *after* the per-device setup, in a global init sink.

Which implies setting a flag in the per-devcie bit, and mangling the event map setup in the global init.

[...]

PR #5284, opened 2019-08-30
[UX] Treat inverse_reading_order as a proper setting

Frenzie:
This is supported by ReaderPaging and ReaderRolling, but the menu entry was only shown in ReaderPaging.

Hold now sets the global default.

Fixes <https://github.com/koreader/koreader/issues/4983>.
Fixes <https://github.com/koreader/koreader/issues/4089>.

### DIFF
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -239,8 +239,8 @@ function ReaderPaging:addToMainMenu(menu_items)
         end,
         sub_item_table = page_overlap_menu,
     }
-    menu_items.read_from_right_to_left = {
-        text = _("Read from right to left"),
+    menu_items.invert_page_turn_gestures = {
+        text = _("Invert page turn taps and swipes"),
         checked_func = function() return self.inverse_reading_order end,
         callback = function()
             self.inverse_reading_order = not self.inverse_reading_order
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -253,8 +253,6 @@ function ReaderPaging:addToMainMenu(menu_items)
                 ok_text = self.inverse_reading_order and _("Enable")
                     or _("Disable"),
                 ok_callback = function()
-                    self.inverse_reading_order = not self.inverse_reading_order
-                    self:setupTapTouchZones()
                     G_reader_settings:saveSetting("inverse_reading_order", self.inverse_reading_order)
                     if touchmenu_instance then touchmenu_instance:updateItems() end
                 end,
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -1,13 +1,14 @@
-local InputContainer = require("ui/widget/container/inputcontainer")
-local Geom = require("ui/geometry")
-local Input = require("device").input
+local ConfirmBox = require("ui/widget/confirmbox")
 local Device = require("device")
-local Screen = Device.screen
 local Event = require("ui/event")
-local UIManager = require("ui/uimanager")
+local Geom = require("ui/geometry")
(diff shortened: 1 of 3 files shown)


frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -240,6 +246,21 @@ function ReaderPaging:addToMainMenu(menu_items)
             self.inverse_reading_order = not self.inverse_reading_order
             self:setupTapTouchZones()
         end,
+        hold_callback = function(touchmenu_instance)
+            self.inverse_reading_order = not self.inverse_reading_order
+            
+            self:setupTapTouchZones()
```
poire-z: Is toggling it and calling self:setupTapTouchZones() (here and in readerrolling) before the ok_callback right?
Either not at all (if settings as default should not apply to the current document), or in the ok_callback if it should?

Frenzie: Good point. I think not at all would be consistent with how it works elsewhere, and also with the logic behind text string.

Frenzie:
Looks like this:

Screenshot_20190830_123823
Screenshot_20190830_123832

poire-z:
Might find its way nearer to (or inside) "Taps & gestures" ?

And if we are going to support RTL in documents, we might soon think about supporting it in the UI.
So, may be that setting should apply as a whole, including when browsing menus?

Page/menu turning seems like a behavioral habit, so I don't know if people can naturally switch their page turning reflexes when reading English or Arabic PDFs... and when browsing menus.

Frenzie:
There's no "going to". This has been there for years. It should either be like this or not at all, imho. ;-)

But indeed, it should go inside taps & gestures.

Frenzie:
Although, not necessarily. This is a document-specific setting. Taps & gestures aren't.

Frenzie:
How about next to its little friend in navigation? (I don't really know when that was added, lol.)

Screenshot_20190830_125918

poire-z:
Yep, that looks fine (the other comes from fixing #4446).
Might be nice to add some `help_text=` to it/these :)

Frenzie:
Isn't help_text also a hold_callback?

Frenzie:
I'll merge this one, and leave help text for another commit/PR.

poire-z:
> Isn't help_text also a hold_callback?

Oh, right :)

Frenzie:
Could potentially still be useful on the other one. :-)

hius07:
As we now have a new `Page turns` menu (https://github.com/koreader/koreader/pull/7572), isn't it a better place for `Invert page turns` items?

NiLuJe:
There's also a case to be made for moving those to "Navigation", maybe?

But, in any case, agreed that they should be together ;).

poire-z:
"Invert page turns" is a bit different than all the others found in "Taps and gestures" as it is a per-document setting, and should probably be easily'er accessible for people who switch between arabic and english books. Moving it in `Tap&Gestures> Page turns>` would make it one tap further than what it currently is when in `Navigation>` (but not too top level, as 99% of our users only read LTR and don't need that item). But as I'm not one of these people, I wouldn't mind much :)

hius07:
> it is a per-document setting

I don't have a device with buttons, but am I right that
https://github.com/koreader/koreader/blob/617ed2c0782944efad48e16063a88bc785aa2e61/frontend/ui/elements/common_settings_menu_table.lua#L428
is applicable to the file browser page turns?

Secondly, is it a good idea to extend disabling swipes onto the file browser?
https://github.com/koreader/koreader/pull/7572#issuecomment-828381154

poire-z:
> is applicable to the file browser page turns?

Looks like it is, it's handled very low level in device/generic/device.lua Device:init().

> Secondly, is it a good idea to extend disabling swipes onto the file browser?

Answering that there (not that I have much to say :), to not mix issues.

NiLuJe:
Oh, right, it's for LTR/RTL swaps, right. Yeah, probably ought to stay put, then ;).

Issue #3019, opened 2017-07-21
Autoscroll


rjbergerud:
* KOReader version: v2015.11-1023-g773e24f
* Device: Kobo Aura One

#### Issue
It would be great to have an autoscroll feature, especially for those using their device to read music.  At minimum, this would include being able to control autoscroll speed.  I couldn't find any similar feature requests in the issues.

pibone:
This could be really interesting, there is an alternative, maybe better for e-ink devices.
Instead (or also :smile:) of scrolling, might be useful to implement half page turns.

I guess this feature could be implemented easily.

Frenzie:
Yes, someone could probably write a plugin relatively easily.

For the GUI you could take a cue from https://github.com/koreader/koreader/blob/db6c35c14476c4f28fdc838ae5f97bccec47ccba/frontend/apps/reader/modules/readersearch.lua

(e.g., slowest, slower/faster in small increments, fastest)

![screenshot_2017-10-04_11-11-02](https://user-images.githubusercontent.com/202757/31168175-bbc1a72c-a8f4-11e7-90a2-15eabe6db6a6.png)

For distance look at https://github.com/koreader/koreader/blob/db6c35c14476c4f28fdc838ae5f97bccec47ccba/frontend/apps/reader/modules/readerpaging.lua#L438-L454

Let me know if you need any more help. :-)

KenMaltby:
What formats are used for "emusic" files?

Frenzie:
Either PDF if it's properly digital or some image format (most likely encapsulated in PDF) if it's scanned afaik.

Evidlo:
I think half page turns are very useful.

Often when I'm reading a math text, I'll look up a few lines to reference something.  With half-page scrolling you can ensure that both pieces of information are on the same screen.

Frenzie:
@Evidlo You mean half pageturns in autoscroll?

If you mean generically then I'd recommend turning on scroll mode. It's in the menu at the bottom.

![screenshot_2018-01-25_15-54-01](https://user-images.githubusercontent.com/202757/35394409-02f9f418-01e8-11e8-836e-7a92965a50db.png)

That way you can just scroll up or down a little to put something into view with more precision than half page turns would give you.

Evidlo:
That works, but I find clicking and dragging on eink displays really unpleasant.

Does scroll mode automatically horizontally lock the page so that I don't accidentally pan left or right as I drag my finger up?

Frenzie:
[...]

PR #5295, opened 2019-09-01
[feat, plugin] Add AutoScroll

Frenzie:
Fixes <https://github.com/koreader/koreader/issues/3019>.

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -369,8 +369,6 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
         for i = 1, #self._window_stack do
             self:setDirty(self._window_stack[i].widget)
         end
-        -- Inform everyone who's the boss now.
-        self:broadcastEvent(Event:new("TopWidget", ((self._window_stack[#self._window_stack] or {}).widget or {}).name))
         self:_refresh(refreshtype, refreshregion, refreshdither)
     end
 end
@@ -621,6 +619,11 @@ end
 --- Gets full refresh rate for e-ink screen.
 function UIManager:getRefreshRate(rate)
     return self.FULL_REFRESH_COUNT
+end
+
+--- Get top widget.
+function UIManager:getTopWidget()
+    return ((self._window_stack[#self._window_stack] or {}).widget or {}).name
 end
 
 --- Signals to quit.
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -308,6 +308,8 @@ function UIManager:show(widget, refreshtype, refreshregion, x, y, refreshdither)
     end
     -- and schedule it to be painted
     self:setDirty(widget, refreshtype, refreshregion, refreshdither)
+    -- Inform everyone who's the boss now.
+    self:broadcastEvent(Event:new("TopWidget", widget.name))
     -- tell the widget that it is shown now
     widget:handleEvent(Event:new("Show"))
     -- check if this widget disables double tap gesture
@@ -369,6 +371,8 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
         for i = 1, #self._window_stack do
             self:setDirty(self._window_stack[i].widget)
         end
+        -- Inform everyone who's the boss now.
+        self:broadcastEvent(Event:new("TopWidget", ((self._window_stack[#self._window_stack] or {}).widget or {}).name))
         self:_refresh(refreshtype, refreshregion, refreshdither)
     end
 end
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -308,8 +308,6 @@ function UIManager:show(widget, refreshtype, refreshregion, x, y, refreshdither)
     end
     -- and schedule it to be painted
     self:setDirty(widget, refreshtype, refreshregion, refreshdither)
-    -- Inform everyone who's the boss now.
-    self:broadcastEvent(Event:new("TopWidget", widget.name))
     -- tell the widget that it is shown now
     widget:handleEvent(Event:new("Show"))
     -- check if this widget disables double tap gesture
--- a/plugins/autoscroll.koplugin/main.lua
+++ b/plugins/autoscroll.koplugin/main.lua
@@ -78,6 +78,11 @@ function AutoScroll:init()
     self:_start()
 end
 
+function AutoScroll:onCloseDocument()
+    logger.dbg("AutoScroll: onCloseDocument")
+    self:_deprecateLastTask()
+end
+
 function AutoScroll:onInputEvent()
     logger.dbg("AutoScroll: onInputEvent")
     self.last_action_sec = os.time()
--- a/plugins/autoscroll.koplugin/main.lua
+++ b/plugins/autoscroll.koplugin/main.lua
@@ -117,7 +117,7 @@ function AutoScroll:addToMainMenu(menu_items)
                 cancel_callback = function()
                     self.enabled = false
                     G_reader_settings:flipFalse("auto_scroll_enabled")
-                    self:deprecateLastTask()
+                    self:_deprecateLastTask()
                 end,
                 callback = function(autoscroll_spin)
                     self.auto_scroll_sec = autoscroll_spin.value
--- a/plugins/autoscroll.koplugin/main.lua
+++ b/plugins/autoscroll.koplugin/main.lua
@@ -14,6 +14,7 @@ local AutoScroll = WidgetContainer:new{
     autoscroll_distance = G_reader_settings:readSetting("autoscroll_distance") or 1,
     enabled = G_reader_settings:isTrue("autoscroll_enabled"),
     settings_id = 0,
+    running = false,
     last_action_sec = os.time(),
 }
 
@@ -50,6 +51,7 @@ end
 function AutoScroll:_deprecateLastTask()
     PluginShare.pause_auto_suspend = false
     self.settings_id = self.settings_id + 1
+    self.running = false
     logger.dbg("AutoScroll: deprecateLastTask ", self.settings_id)
 end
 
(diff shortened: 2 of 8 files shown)


plugins/autoscroll.koplugin/main.lua
```diff
@@ -0,0 +1,132 @@
...
+local AutoScroll = WidgetContainer:new{
+    name = 'autoscroll',
+    is_doc_only = true,
+    auto_scroll_sec = G_reader_settings:readSetting("auto_scroll_timeout_seconds") or 0,
+    settings_id = 0,
+    last_action_sec = os.time(),
+}
+
+function AutoScroll:_enabled()
+    return self.auto_scroll_sec > 0
```
Frenzie: This makes more sense to me as a completely separate enabled setting, but for starters I just wanted to get the base functionality working, and deviating from AutoSuspend would've taken more time.

plugins/autoscroll.koplugin/main.lua
```diff
@@ -0,0 +1,136 @@
...
+            local autoscroll_spin = SpinWidget:new {
+                width = Screen:getWidth() * 0.6,
+                value = curr_items,
+                value_min = 0,
+                value_max = 240,
+                value_hold_step = 5,
+                ok_text = _("Set timeout"),
+                cancel_text = _("Disable"),
+                title_text = _("Timeout in seconds"),
+                cancel_callback = function()
```
Frenzie: Looks like SpinWidget doesn't do this atm; will add soon.

poire-z:
No issue with the whole thing.
May be just the name, "Autoscroll", as it doesn't scroll but turn pages :) "Autoturnpage" ? :)
If we ever want to add a real autoscroll (that would  `"GotoViewRel", 0.01` every second, if that ever works), there would be some conflict (unless this could fit into this plugin).

Anyway, if the plugin is named autoscroll, and that setting gets into the main settings store, better name it with the plugin name as prefix, that is: `autoscroll_timeout_seconds` instead of `auto_scroll_timeout_seconds`.

plugins/autoscroll.koplugin/main.lua
```diff
@@ -0,0 +1,136 @@
...
+        logger.dbg("AutoScroll:_schedule registered settings_id ",
+                   settings_id,
+                   " does not equal to current one ",
+                   self.settings_id)
+        return
+    end
+
+    local delay
+
+    if PluginShare.pause_auto_scroll then
```
poire-z: A bit unclear to me: are you using that evil BackgroundRunner? Or doing the UIManager.schedule / unschedule yourself?
(looks like the later - but that PluginShare made me thought initially you went with BackgroundRunner).
Anyway, is that `PluginShare.pause_auto_scroll` needed ? A left-over from copied plugin? Something for the future to stop it from elsewhere?

Frenzie:
> If we ever want to add a real autoscroll (that would "GotoViewRel", 0.01 every second, if that ever works), there would be some conflict (unless this could fit into this plugin).

I see that as an expansion of this plugin, like set scroll distance or something.

plugins/autoscroll.koplugin/main.lua
Frenzie: Oh yeah, a vestigial function. This is a KeepAlive hook for the AutoSuspend plugin.

Frenzie: Come to think of it, it should still be kept. This one and AutoSuspend reset with an InputEvent hook, but this plugin doesn't emit input events, only `GotoViewRel`.

Frenzie:
@poire-z Any further comments?

plugins/autoscroll.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+            local SpinWidget = require("ui/widget/spinwidget")
+            local curr_items = G_reader_settings:readSetting("autoscroll_timeout_seconds") or 30
+            local autoscroll_spin = SpinWidget:new {
+                width = Screen:getWidth() * 0.6,
+                value = curr_items,
+                value_min = -20,
+                value_max = 20,
+                precision = "%.2f",
+                value_step = .1,
+                value_hold_step = .5,
```
poire-z: So, 1 = 1 page, 0.3 = 30% of the page?
Do value_min/max = 20, or anything > 1, has some use case?

plugins/autoscroll.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+function AutoScroll:_start()
+    if self:_enabled() then
+        logger.dbg("AutoScroll: start at ", os.time())
+        PluginShare.pause_auto_suspend = true
+        self.last_action_sec = os.time()
+        self:_schedule(self.settings_id)
+
+        local InfoMessage = require("ui/widget/infomessage")
+        UIManager:show(InfoMessage:new{
+            text = T(_("Autoscroll is now active and will automatically turn the page every %1 seconds."), self.autoscroll_sec),
```
poire-z: "turn the page" - but we now can turn some fraction of the page :)
So, just "scroll", or `if distance == 1 then "turn the page" else T("scroll %1%% of the page", distance)` ?

plugins/autoscroll.koplugin/main.lua
Frenzie: I think so, yes. Or if not it's close enough for it not to matter. One would just decrease or increase the value to achieve the desired effect.

I just noticed something odd though. In paged mode fractional values result in an instant end of book dialog. And related, sending this event when one's already showing pops up more on top of it.

poire-z: What happens when you're doing a dict lookup and enjoying the result? The page still scroll in the background without any issue?

Might need to pause this autoscroll whenever the readerview is not the top one ? :|

Frenzie: > What happens when you're doing a dict lookup and enjoying the result? The page still scroll in the background without any issue?

If you're doing it really slowly without interacting, sure.

> Might need to pause this autoscroll whenever the readerview is not the top one ? :|

Is that currently possible?

poire-z: Dunno :) probably not.
Might need to ask UIManager if there's something in the stack above the reader...
Or have UIManager send some event when covered/uncovered, when a widget has requested it (and give the names of the events, or some fixed event with the widget name/id :)
Probably quite some work, for a little plugin :) (pinging @NiLuJe if he sees something easy.)

NiLuJe: The least intrusive would probably to add something to the effect of "am I on top of the stack", and query that from the plugin as-needed?

NiLuJe: (That might have to be sneakier: am I the highest *visible* widget on the stack).

Or not, Been a while since I played w/ UIManager.

Otherwise, possibly as easy as checking if `self._window_stack[1].widget == widget` ;).

Frenzie: The top of the stack is probably queried easily enough, but ReaderUI isn't the top of the stack. There's statusbars, bookmarks, and all kinds of stuff like that on top of it.

poire-z: Or just a simple method UIManager:getTopWidgetId() => "reader" or "#1234ab32" that the widget could just call before deciding to send the onGotoViewRel.

frontend/ui/uimanager.lua
```diff
@@ -369,6 +369,8 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
         for i = 1, #self._window_stack do
             self:setDirty(self._window_stack[i].widget)
         end
+        -- Inform everyone who's the boss now.
+        self:broadcastEvent(Event:new("TopWidget", ((self._window_stack[#self._window_stack] or {}).widget or {}).name))
```
Frenzie: @poire-z @NiLuJe Feels a bit dirty but it seems to work. Any suggestions on a better way? Instead of unscheduling the plugin on an event there could also be a method that's checked before turning the page or something.

NiLuJe: How much do we end up firing that `TopWidget` event, ultimately?

poire-z: > How much do we end up firing that TopWidget event

Each time we UIManager:show() a widget, which is not that much, but it goes thru all the widgets to broadcast it. Which might be negligeable, I don't know, but still :)

> Instead of unscheduling the plugin on an event there could also be a method that's checked before turning the page or something.

Like I suggested above?:
> Or just a simple method UIManager:getTopWidgetId() => "ReaderUI" or "#1234ab32" that the widget could just call before deciding to send the onGotoViewRel.

poire-z: (Why did you remove the one on :show() and kept only the one on :close() ? )

Frenzie: > How much do we end up firing that TopWidget event, ultimately?

Not much at all. So it's distributing the event to all widgets every once in a while vs always checking the top widget every x seconds. But of course this event implementation would affect everything and the plugin just users of the plugin, so out the event will go. ;-)

> Why did you remove the one on :show() and kept only the one on :close() ? )

It wasn't working right, ran into some kind of loop.

poire-z: (Well, the event could have some other usages, like stopping statistics, to get a more accurate time to read a page.
Unless all the little times we spend dict or wikipedia lookup'ing need to be accounted in this, assuming it's part of the reading time :)

Frenzie: I don't know, I think the statistics are pretty cool but I'm not very concerned about their accuracy personally.

AlanSP1:
To me, automatic page turning, or autopageturner sounds better...

frontend/ui/uimanager.lua
Frenzie: @poire-z Btw, that issue I noticed with it continuously popping up end of document dialogs was magically resolved by your suggestion, because now it automatically pauses when that dialog is up. :-)

Frenzie:
@AlanSP1 Possibly, but note that autoscroll is a "traditional" term. See, e.g., [here](https://youtu.be/znH_x945kpI?t=50). Then again, the same applies to [page turner](https://news.mit.edu/1999/pageturner).

Frenzie:
Alright, final screens:

PR #5297, opened 2019-09-01
handle file:// protocol URI scheme

contrun:
This patch only handles part of  file URI scheme defined in [rfc 8089](https://tools.ietf.org/html/rfc8089), i.e., it currently only handles `file://ABSOLUTE_PATH`.

### DIFF
--- a/reader.lua
+++ b/reader.lua
@@ -201,16 +201,33 @@ SettingsMigration:migrateSettings(G_reader_settings)
 
 local exit_code
 
+local function getPathFromURI(str)
+    local hexToChar = function(x)
+        return string.char(tonumber(x, 16))
+    end
+
+    local unescape = function(url)
+       return url:gsub("%%(%x%x)", hexToChar)
+    end
+
+    local prefix = "file://"
+    if str:sub(1, #prefix) ~= prefix then
+        return str
+    end
+    return unescape(str):sub(#prefix+1)
+end
+
 if ARGV[argidx] and ARGV[argidx] ~= "" then
     local file
-    if lfs.attributes(ARGV[argidx], "mode") == "file" then
-        file = ARGV[argidx]
+    local sanitized_path = getPathFromURI(ARGV[argidx])
+    if lfs.attributes(sanitized_path, "mode") == "file" then
+        file = sanitized_path
     elseif open_last and last_file then
         file = last_file
     end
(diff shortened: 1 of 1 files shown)


reader.lua
```diff
@@ -201,10 +201,28 @@ SettingsMigration:migrateSettings(G_reader_settings)
 
 local exit_code
 
+local function getPathFromURI(str)
+    local function decodeChar(hex)
+        return string.char(tonumber(hex,16))
+    end
+
+    local function decodeString(str)
+        local output, t = string.gsub(str,"%%(%x%x)",decodeChar)
```
Frenzie: We can just leave it out if unused.
```suggestion
        local output = string.gsub(str,"%%(%x%x)",decodeChar)
```

reader.lua
```diff
@@ -201,10 +201,27 @@ SettingsMigration:migrateSettings(G_reader_settings)
 
 local exit_code
 
+local function getPathFromURI(str)
+    local hex_to_char = function(x)
```
Frenzie:
```suggestion
    local hexToChar = function(x)
```
That's in our style guide to distinguish functions/methods from variables.

contrun:
You mean how I came to the thought of handling `file://`?
I was using [mimeo](https://xyne.archlinux.ca/projects/mimeo/) and [fzf](https://github.com/junegunn/fzf) to open pdf files quickly. `mimeo` automatically converts file paths to `file://` format. So I was not able to view pdf files as expected.

reader.lua
```diff
@@ -201,10 +201,27 @@ SettingsMigration:migrateSettings(G_reader_settings)
...
+
+    local unescape = function(url)
+       return url:gsub("%%(%x%x)", hexToChar)
+    end
+
+    local prefix = "file://"
+    if str:sub(1, #prefix) ~= prefix then
+        return str
+    end
+    return unescape(str):sub(#prefix+1)
```
Frenzie: What if the input is just `file://`?

contrun: never thought about that. Corrected it.

Frenzie:
Right, I was wondering if it came up in regular use, semi-advanced use (e.g., copying the URL from a browser that opens PDFs) or in a completely artificial manner.

Frenzie:
Thanks!

PR #5302, opened 2019-09-01
Revert "[feat, Kobo] Implement shutdown from suspend (#5280)"

Frenzie:
This reverts commit cfa73be940e2468c353f5737a0e76f722362dcc6.

Cervantes and Sony secretly depend on hacky Kobo code.
Reported by @avsej, see <https://github.com/koreader/koreader/issues/2431#issuecomment-526925910>.

Closes <https://github.com/koreader/koreader/issues/5292>.

PR #5304, opened 2019-09-01
[fix] Let UIManager handle suspend logic

Frenzie:
See <https://github.com/koreader/koreader/issues/2431#issuecomment-526925910>.

Supersedes <https://github.com/koreader/koreader/pull/5302>.

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -198,12 +198,19 @@ function UIManager:init()
         self.event_handlers["PowerRelease"] = function()
             if not self._entered_poweroff_stage then
                 UIManager:unschedule(self.poweroff_action)
-                self:_beforeSuspend()
-                Device:intoScreenSaver()
-                Device:suspend()
+                -- resume if we were suspended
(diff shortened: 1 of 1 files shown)


pazos:
Just reading UIManager.lua. What a if elseif end mess :man_facepalming: 

Shouldn't Pocketbook react to PowerPress/PowerRelease like the others?. They only react to suspend/resume in https://github.com/koreader/koreader/blob/2600abe5545deb6e05fc9ae0551a55afc1b4f093/frontend/ui/uimanager.lua#L81-L90

Frenzie:
No, PB uses a handleMiscEv plugin to emit Suspend/Resume events.

https://github.com/koreader/koreader/blob/621f36b4a9eb5b29c7aa62cf2d557f7c71e9c9d8/frontend/device/pocketbook/device.lua#L98-L108

Something like that is probably how that Kobo thing should've been done.

And yes, it's clear that ideally a bunch of that UIManager stuff would be refactored into the device abstractions.

PR #5306, opened 2019-09-02
Allow setting a fine-tuned font size as default

NiLuJe:
And a couple bugfix this unearthed:

* Always set the default as the *current* value, no matter what (by chance, this was not an issue for margins).
* Don't crash on the hold gesture in the -/+ buttons for the horizontal margins (regression since #5303).

### DIFF
--- a/frontend/ui/widget/toggleswitch.lua
+++ b/frontend/ui/widget/toggleswitch.lua
@@ -210,9 +210,20 @@ function ToggleSwitch:onTapSelect(arg, gev)
 end
 
 function ToggleSwitch:onHoldSelect(arg, gev)
+    print("Called ToggleSwitch:onHoldSelect:", arg, gev)
+    print("name:", self.name)
+    print("name_text:", self.name_text)
+    print("val:", self.values or self.args)
+    print("toggle:", self.toggle)
     local position = self:calculatePosition(gev)
-    self.config:onMakeDefault(self.name, self.name_text,
-                    self.values or self.args, self.toggle, position)
+    print("pos:", position)
+    if self.name == "font_fine_tune" then
+        self.config:onMakeFineTuneDefault("font_size", _("Font Size"),
+                        self.values or self.args, self.toggle, position == 1 and "-" or "+")
+    else
+        self.config:onMakeDefault(self.name, self.name_text,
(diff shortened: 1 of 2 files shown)


NiLuJe:
I vaguely thought about making the font_size toggle things more generic, but, AFAICT, this is the onlt thing in the code that's built this way, so, meh.

NiLuJe:
@Frenzie : Err, CircleCI went kablooey?

robert00s:
Thanks for fix.

Frenzie:
@NiLuJe Oh, that sucks. :-/ It's not CircleCI exactly but *your* CircleCI. I added the ability to push to Transifex and the koreader-docs repo through hidden environment variables. So if you added a `koreader-vars`, even if completely empty, it should work just fine.

https://github.com/koreader/koreader/blob/b2554ba5da30840503216c5395ab5b6648b546d7/.circleci/config.yml#L79

I wonder if that context thingy can somehow be optional… it's completely irrelevant except for the push to master. I could change the workflow but that'd be quite a bit of effort for essentially the same result. ;_;

Frenzie:
PS You can find that here, or in your own user/org settings for the repo anyway:

Screenshot_2019-09-02_17-36-30

NiLuJe:
@Frenzie: That did the trick, thanks!

frontend/ui/widget/toggleswitch.lua
```diff
@@ -211,8 +211,14 @@ end
 
 function ToggleSwitch:onHoldSelect(arg, gev)
     local position = self:calculatePosition(gev)
-    self.config:onMakeDefault(self.name, self.name_text,
-                    self.values or self.args, self.toggle, position)
+    if self.name == "font_fine_tune" then
+        --- @note Ugly hack for the only widget that uses a dual toggle for fine-tuning (others prefer a buttonprogress)
```
Frenzie: I also still have to make that fine-tuning thing work for EPUB/FB2 in MuPDF. :-P

Frenzie:
Link to (presumably) the reason for this PR, for posterity: https://www.mobileread.com/forums/showthread.php?p=3884539#post3884539

PR #5307, opened 2019-09-02
Frontlight - Add checkbox use system settings

robert00s:
See: https://github.com/koreader/koreader/issues/5205#issuecomment-526935357

Devices with `hasLightLevelFallback = true` (for now Android) has extra checkbutton `Use system settings`. Default checked.

After check button we have Frontlight widget with read only status.

Also frontlight gestures are disabled.

Checkbutton status is written in global variable `light_fallback`.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -1198,7 +1198,7 @@ function ReaderGesture:registerGesture(ges, action, ges_type, zone, overrides, d
 end
 
 local function lightFrontlight()
-    return Device:hasLightLevelFallback() and G_reader_settings:readSetting("light_fallback") or false
+    return Device:hasLightLevelFallback() and G_reader_settings:nilOrTrue("light_fallback")
 end
 
 function ReaderGesture:gestureAction(action, ges)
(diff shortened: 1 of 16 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1302,13 +1307,29 @@ function ReaderGesture:gestureAction(action, ges)
             self.ui:handleEvent(Event:new("ShowFlDialog"))
         end
     elseif action == "increase_frontlight" then
+        -- when using frontlight system settings
+        if lightFrontlight() then
+            UIManager:show(Notification:new{
+                text = _("Frontlight controlled by system settings."),
```
Frenzie: is/will now be?

Also I think 1 s might be a bit short.

robert00s: 2 s or more?

Frenzie: I'd say 2 or 3.

robert00s: 2.5 s :)

pazos:
Awesome job @robert00s !

I will review changes at night and submit a PR on top of your branch.

A few things I noticed:

> Devices with hasLightLevelFallback = true (for now Android) has extra checkbutton Use system settings. Default unchecked.

Default is system settings, at least on Android. No problem with that since it can be retrieved using

```lua
function Device:isUsingSystemSettings()
    return android.getScreenBrightness < 0
end
```

robert00s:
Now the system settings are the default (checkbox checked)

pazos:
Fixes #5205

PR #4582, opened 2019-02-11
Add highlight editing buttons

Galunid:
This allows user to move highlight end/beginning using buttons that appear after pressing on the highlight.
**Screenshot:**
image
This should be merged after https://github.com/koreader/koreader-base/pull/808

Close #4555. Close #975. Provides a solution for #4134 and https://github.com/koreader/koreader/pull/4267#issuecomment-429926765

PR #5309, opened 2019-09-03
Option to show horizontal line separator in footer

robert00s:
Hidden line separator (default)

New option `Show footer separator`

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -9,10 +9,14 @@ local Geom = require("ui/geometry")
 local HorizontalGroup = require("ui/widget/horizontalgroup")
 local HorizontalSpan = require("ui/widget/horizontalspan")
 local LeftContainer = require("ui/widget/container/leftcontainer")
+local LineWidget = require("ui/widget/linewidget")
 local ProgressWidget = require("ui/widget/progresswidget")
 local RightContainer = require("ui/widget/container/rightcontainer")
+local Size = require("ui/size")
 local TextWidget = require("ui/widget/textwidget")
 local UIManager = require("ui/uimanager")
+local VerticalGroup = require("ui/widget/verticalgroup")
+local VerticalSpan = require("ui/widget/verticalspan")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
 local util = require("util")
 local T = require("ffi/util").template
@@ -257,13 +261,6 @@ function ReaderFooter:init()
         margin_span,
     }
 
-    self.footer_content = FrameContainer:new{
-        self.horizontal_group,
-        background = Blitbuffer.COLOR_WHITE,
-        bordersize = 0,
-        padding = 0,
-        padding_bottom = self.bottom_padding,
-    }
     self:updateFooterContainer()
 
     self.mode = G_reader_settings:readSetting("reader_footer_mode") or self.mode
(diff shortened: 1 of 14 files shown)


Frenzie:
Do you intend to offer more options? Because just on/off is what the checkmark is for. ;-)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -649,7 +672,40 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            self:resetLayout(true)
+                            self:updateFooter()
+                            self.ui:handleEvent(Event:new("SetPageBottomMargin", self.view.document.configurable.b_page_margin))
+                            UIManager:setDirty(nil, "ui")
+                        end,
+                    },
+                },
+            },
+            {
+                text = _("Items separator"),
```
```suggestion
                text = _("Item separator"),
```

robert00s:
> Do you intend to offer more options? Because just on/off is what the checkmark is for. ;-)

Of course :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -287,22 +284,48 @@ function ReaderFooter:updateFooterContainer()
...
+
+    if not self.settings.bottom_horizontal_separator then
+        vertical_span.width = 0
+        separator_line.dimen.h = 0
     end
+    self.vertical_frame = VerticalGroup:new{
+        separator_line,
+        vertical_span,
+        self.footer_container,
+    }
```
poire-z: I'd rather not see `separator_line` and `vertical_span` when `not self.settings.bottom_horizontal_separator `, instead of giving them a height of 0 (because that's not pretty :) and may be one day, they won't support a height of 0 :)
So, rather:
```lua
if self.settings.bottom_horizontal_separator then
  local separator_line = ...
  local vertical_span = ...
  self.vertical_frame = VerticalGroup:new{
        separator_line,
        vertical_span,
        self.footer_container,
    }
else
  self.vertical_frame = VerticalGroup:new{
        self.footer_container,
    }
end
```
It looks like we always access things by name, nowhere with indices like `self.vertical_frame[3]`, so that should work.
(Even just `self.vertical_frame = self.footer_container` might work, but may be less risky keeping the same hierarchy.)

Frenzie: And even if we did use `self.vertical_frame[3]` that would generally make it a piece of code that should be changed to a named reference for clarity and maintainability instead.

poire-z:
All fine to me with the last commit.

I get all the previous commits (cervantes, frontlight, noto) are just some other already commited commits that were brought in by some git update stuff, and they'll be no-op when merged.

But now, with this PR, as a reviewer, how can I be sure of that? And that there's not a new thing introduced in the other related files, like font.lua? (I trust @robert00s, it's just an academic question about using git/github and if it can help with that).

(I usually do followup work like in https://github.com/koreader/koreader/pull/4582#issuecomment-462975795 to avoid bringing alien commits in the PR.)

NiLuJe:
I generally rebase against upstream/master instead of merging
upstream/master in my tree when this happens ;).

Frenzie:
It's a bit of a difficult question. On the one hand merging master makes it so that the branch always remains usable for people who are trying it out. They can just fetch/pull from the branch without issue. On the other hand `git pull -r upstream master` is much cleaner, but force pushing breaks things for people trying your branch and potentially kills some commit history relevant to the PR review process too.

For a perspective I mostly disagree with, see [here](https://github.com/FreshRSS/FreshRSS/pull/2481#issuecomment-521726656). I prefer that the Git repo makes sense without GH, which means a mixture between rebasing and squashing.

robert00s:
I should read more about git :)
Should I do something additional in this PR?

Frenzie:
@robert00s It only just occurred to me now, but maybe status bar separator would be better?

robert00s:
Status bar as separator will be next step :)
And I saw yesterday in new Kobo Libra, status bar at the very bottom of edge:

obraz
I like it.

NiLuJe:
The FW update download progress is similar (although at the top of the screen, and possibly thicker) ;).

(That was a fun useless thought I had when I noticed the "new" progress bar ;)).

Frenzie:
What's the stuff at the top & bottom? Top total pages, bottom chapter pages?

NiLuJe:
I'd assume book on top, chapter at the bottom, yeah. I think david detailed it properly in the Libra thread on MR.

PR #5310, opened 2019-09-03
[Android] Don't ship Noto

Frenzie:
Follow-up to <https://github.com/koreader/koreader/pull/5252>. This greatly reduces the Android package size.

See discussion in <https://github.com/koreader/koreader/pull/5264#issuecomment-525517001>.

### DIFF
--- a/frontend/fontlist.lua
+++ b/frontend/fontlist.lua
@@ -81,7 +81,7 @@ local function _readList(target, dir)
         local mode = lfs.attributes(dir.."/"..f, "mode")
         if mode == "directory" and f ~= "." and f ~= ".." then
             _readList(target, dir.."/"..f)
-        elseif mode == "file" or mode == "link"  then
+        elseif mode == "file" or mode == "link" then
             if string.sub(f, 1, 1) ~= "." then
                 local file_type = string.lower(string.match(f, ".+%.([^.]+)") or "")
                 if file_type == "ttf" or file_type == "ttc"
--- a/frontend/fontlist.lua
+++ b/frontend/fontlist.lua
@@ -74,13 +74,13 @@ local function getExternalFontDir()
 end
 
 local function _readList(target, dir)
-    -- lfs.dir non-exsitent directory will give error, weird!
+    -- lfs.dir non-existent directory will give an error, weird!
     local ok, iter, dir_obj = pcall(lfs.dir, dir)
     if not ok then return end
     for f in iter, dir_obj do
         if lfs.attributes(dir.."/"..f, "mode") == "directory" and f ~= "." and f ~= ".." then
             _readList(target, dir.."/"..f)
-        else
+        elseif lfs.attributes(dir.."/"..f, "mode") ~= nil then
             if string.sub(f, 1, 1) ~= "." then
                 local file_type = string.lower(string.match(f, ".+%.([^.]+)") or "")
                 if file_type == "ttf" or file_type == "ttc"
--- a/frontend/fontlist.lua
+++ b/frontend/fontlist.lua
@@ -80,7 +80,7 @@ local function _readList(target, dir)
     for f in iter, dir_obj do
         if lfs.attributes(dir.."/"..f, "mode") == "directory" and f ~= "." and f ~= ".." then
             _readList(target, dir.."/"..f)
-        elseif lfs.attributes(dir.."/"..f, "mode") ~= nil then
+        elseif lfs.attributes(dir.."/"..f, "mode") == "file" then
             if string.sub(f, 1, 1) ~= "." then
                 local file_type = string.lower(string.match(f, ".+%.([^.]+)") or "")
                 if file_type == "ttf" or file_type == "ttc"
--- a/frontend/fontlist.lua
+++ b/frontend/fontlist.lua
@@ -78,9 +78,10 @@ local function _readList(target, dir)
     local ok, iter, dir_obj = pcall(lfs.dir, dir)
     if not ok then return end
     for f in iter, dir_obj do
-        if lfs.attributes(dir.."/"..f, "mode") == "directory" and f ~= "." and f ~= ".." then
+        local mode = lfs.attributes(dir.."/"..f, "mode")
+        if mode == "directory" and f ~= "." and f ~= ".." then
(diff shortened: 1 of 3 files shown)


frontend/ui/font.lua
```diff
@@ -108,13 +111,38 @@ function Font:getFace(font, size)
...
+
         -- Not all fonts are bundled on all platforms because they come with the system.
         -- In that case, search through all font folders for the requested font.
-        if not ok and font ~= realname then
+        if not ok then
             local fonts = FontList:getFontList()
+            local escaped_realname = realname:gsub("[-]", "%%-")
+
             for _k, _v in ipairs(fonts) do
-                if _v:find(realname) then
```
Frenzie: I was so confused why this wasn't working for a little while until I remembered `-` is a special character. One which of course  doesn't occur in `DroidSansMono`, but does in `DroidSans-Bold`/`DroidSans-Regular`…

frontend/fontlist.lua
```diff
@@ -74,13 +74,13 @@ local function getExternalFontDir()
...
 local function _readList(target, dir)
-    -- lfs.dir non-exsitent directory will give error, weird!
+    -- lfs.dir non-existent directory will give an error, weird!
     local ok, iter, dir_obj = pcall(lfs.dir, dir)
     if not ok then return end
     for f in iter, dir_obj do
         if lfs.attributes(dir.."/"..f, "mode") == "directory" and f ~= "." and f ~= ".." then
             _readList(target, dir.."/"..f)
-        else
+        elseif lfs.attributes(dir.."/"..f, "mode") ~= nil then
```
NiLuJe: -> `== file`? Or could this arguably hit a symlink one day?

frontend/ui/font.lua
```diff
@@ -10,30 +12,30 @@ local logger = require("logger")
 local Font = {
     fontmap = {
         -- default font for menu contents
-        cfont = "noto/NotoSans-Regular.ttf",
+        cfont = "NotoSans-Regular.ttf",
```
NiLuJe: Isn't losing the subdirectory an issue on !Android, where they do live in a noto subdirectory?

Frenzie: The fonts are scanned in ./fonts/* first. :-)

frontend/fontlist.lua
Frenzie: Possibly on regular desktop, but good point.

frontend/ui/font.lua
Frenzie: Oh drat, my reversal commit introduced an issue here sometimes skipping that scan. Meh.

Frenzie: @NiLuJe The logic could be slightly different, i.e. try with full path and if fail use [util.spiltFilePathName](http://koreader.rocks/doc/modules/util.html#splitFilePathName) to search all folders. It'd be negligibly faster on platforms that ship all fonts; it's more the future readability of the logic that I'm trying to figure out.

frontend/fontlist.lua
```diff
@@ -74,13 +74,13 @@ local function getExternalFontDir()
...
 local function _readList(target, dir)
-    -- lfs.dir non-exsitent directory will give error, weird!
+    -- lfs.dir non-existent directory will give an error, weird!
     local ok, iter, dir_obj = pcall(lfs.dir, dir)
     if not ok then return end
     for f in iter, dir_obj do
         if lfs.attributes(dir.."/"..f, "mode") == "directory" and f ~= "." and f ~= ".." then
             _readList(target, dir.."/"..f)
-        else
+        elseif lfs.attributes(dir.."/"..f, "mode") == "file" then
```
Frenzie: @NiLuJe Actually on this note, on newer Android Droid is Roboto, so it's not just theory: 

```
lrw-r--r-- root     root              2017-07-12 21:51 DroidSans-Bold.ttf -> Roboto-Bold.ttf
lrw-r--r-- root     root              2017-07-12 21:51 DroidSans.ttf -> Roboto-Regular.ttf
```

(But I don't think

Frenzie: Where by "newer" I presumably mean 4.0 and up. (I can only test 2.3 and 4.1 and up because the 4.0 emulator has always been borked.)

NiLuJe: Ah, nice find ;).

I'm wondering if there are any other possible values for attributes besides dir/file/link? (In which case the original test was perfectly fine ;p).

Frenzie: https://keplerproject.github.io/luafilesystem/manual.html
> string representing the associated protection mode (the values could be file, directory, link, socket, named pipe, char device, block device or other)

NiLuJe: Ah, thanks, it always takes me forever to grok the lfs docs ;).

Frenzie: So I don't think any of the others should be relevant?

NiLuJe: Yup! :+1:

pazos:
With this PR merged and the new apk below 20MB I think it shouldn't hurt to migrate some stuff to Kotlin. It adds 800kb to the final APK and about 1-3% of extra build time but reduces boilerplate code, looks nice and has safer constructs. See an example: https://github.com/pazos/android-luajit-launcher/commit/06658befd2ffcf1390a6e147df6fbc4c11f264ec

Frenzie:
I have no real opinion. Kotlin is certainly less verbose just to get something done, which is nice of course.

NiLuJe:
Can't really be worse than plain java, so, yeah, +1 ;).

pazos:
> Can't really be worse than plain java, so, yeah, +1 ;).

A few killer features IMO:

- [Null safety](https://kotlinlang.org/docs/reference/null-safety.html)
- [Smart casts](https://kotlinlang.org/docs/reference/typecasts.html)
- [Java interop](https://kotlinlang.org/docs/reference/java-interop.html)
- [auto getter/setters for mutable properties](https://kotlinlang.org/docs/reference/properties.html)

Frenzie:
Their code samples are terrible though. I had to turn on their JS just to see them! (A proper JS syntax highlighting design just builds on top of `<pre><code>` so that it's always visible.)

PR #3420, opened 2017-10-24
[readtimer] Separate strings for singular and plural

robert00s:
Close: #3397

PR #5315, opened 2019-09-04
[i18n, plugins] ReadTimer - better plural forms

robert00s:
See: https://github.com/koreader/koreader/pull/3420#pullrequestreview-71776459

### DIFF
--- a/plugins/readtimer.koplugin/main.lua
+++ b/plugins/readtimer.koplugin/main.lua
@@ -2,8 +2,9 @@ local InfoMessage = require("ui/widget/infomessage")
 local TimeWidget = require("ui/widget/timewidget")
 local UIManager = require("ui/uimanager")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
-local T = require("ffi/util").template
 local _ = require("gettext")
+local N_ = _.ngettext
+local T = require("ffi/util").template
 
 local ReadTimer = WidgetContainer:new{
     name = "readtimer",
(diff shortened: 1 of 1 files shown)


plugins/readtimer.koplugin/main.lua
```diff
@@ -154,19 +150,18 @@ function ReadTimer:addToMainMenu(menu_items)
...
-                                    min_str = string.format("%d %s", min, minutes_string)
+                                if min > 0 then
+                                    min_str = T(N_("1 minute", "%1 minutes", min), min)
+                                    if hr_str ~= "" then
+                                        hr_str = hr_str .. " "
+                                    end
                                 end
                                 UIManager:show(InfoMessage:new{
-                                    text = T(_("Timer is set to %1 %2"), hr_str, min_str),
+                                    text = T(_("Timer is set to %1%2."), hr_str, min_str),
```
Frenzie:
```suggestion
                                    text = T(_("Timer set for %1%2."), hr_str, min_str),
```

plugins/readtimer.koplugin/main.lua
```diff
@@ -96,19 +93,18 @@ function ReadTimer:addToMainMenu(menu_items)
...
-                                    min_str = string.format("%d %s", min, minutes_string)
+                                if min > 0 then
+                                    min_str = T(N_("1 minute", "%1 minutes", min), min)
+                                    if hr_str ~= "" then
+                                        hr_str = hr_str .. " "
+                                    end
                                 end
                                 UIManager:show(InfoMessage:new{
-                                    text = T(_("Timer set to: %1:%2\nIt's %3 %4 from now"),
+                                    text = T(_("Timer set to: %1:%2.\nIt's %3%4 from now."),
```
Frenzie: That "it's" is a bit weird. Maybe something like this.
```suggestion
                                    text = T(_("Timer set to: %1:%2.\n\nThat's %3%4 from now."),
```

Issue #3097, opened 2017-08-18
Occasional mysterious ReadHistory failure


Frenzie:
* KOReader version: I've seen it a couple of times, most recently 7376c57
* Device: Travis

#### Issue
```
Failure → spec/front/unit/readhistory_spec.lua @ 226

ReadHistory module should be able to remove three items

spec/front/unit/readhistory_spec.lua:40: Expected objects to be the same.

Passed in:

(string) 'a'

Expected:

(string) 'd'
```

Restarting the build generally works.

#### Steps to reproduce
Unknown.

poire-z:
OK, there are a few more tests (#2985, #3380) that could benefit from a usleep(1000000) between the addItem().

Frenzie:
Oh, now I understand your earlier remark. I'm not entirely sure if the tests are the right place to "fix" this. Pro Starcraft gamers perform 10 actions per second.

poire-z:
Must have been solved by #6991.

PR #5316, opened 2019-09-04
[Cervantes]: launch QBookApp from KOReader

pazos:
Supersedes #5081 

It was tested by @ZapperDJ in https://github.com/koreader/koreader/pull/5081#issuecomment-527653617


pazos:
Err, I think `[  FAILED  ] spec/front/unit/readhistory_spec.lua @ 363: ReadHistory module should not automatically remove deleted items from history if setting has been set to false` has nothing to do with this pull :worried:

Frenzie:
Probably https://github.com/koreader/koreader/issues/3097. You can just restart the build in the once a month case it happens.

PR #5318, opened 2019-09-04
[UX] Allow keyboard layout selection

Frenzie:
Fixes <https://github.com/koreader/koreader/issues/3998>.

This is an unpolished quickie but I've been letting this slip for too
long due to fancier keyboard interaction concepts.

Screenshot_2019-09-04_21-33-41
Screenshot_2019-09-04_21-33-49
Screenshot_2019-09-04_21-33-56

### DIFF
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -210,6 +210,10 @@ if Device:isAndroid() then
 end
 
 if Device:isTouchDevice() then
+    common_settings.keyboard_layout = {
+        text = _("Keyboard layout"),
+        sub_item_table = require("ui/elements/menu_keyboard_layout"),
+    }
     common_settings.taps_and_gestures = {
         text = _("Taps and gestures"),
     }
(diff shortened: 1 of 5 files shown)


poire-z:
The order of the items is random (not the same in the file vs your screenshot) ?
Any chance to get them in some order (with may be en/default at first place)?
Would it make sense to give the menu item better names than these few letters? Or is that the classic naming for keyboards?
(Alternative, the 6 first letters of the keyboard, `qwerty`, `azerty`, `ㅃㅉㄸㄲㅆㅛ`, `あかさたなは`.)

Frenzie:
That's part of what I mean by unpolished. ;-)

But sure, I can probably polish it at least a little tomorrow.

Frenzie:
Screenshot_20190905_215814

Frenzie:
@poire-z Do you think it's acceptable now for a first pass?

noembryo:
Just saw this and I wonder, should I change the `el_keyboard.lua` to make it that the Greek characters are default?

Frenzie:
@noembryo If that's what people expect from a Greek keyboard then that's probably a good idea. Judging by [this](https://en.wikipedia.org/wiki/Keyboard_layout#Greek) it looks like it should be delta for D, etc.?

noembryo:
Yeap..
I'll try a PR (I did one already from the browser, maybe I'll try one from inside my IDE ;o)

PR #5324, opened 2019-09-05
[chore] Drop SDL 1.2

Frenzie:
Depends on https://github.com/koreader/koreader-base/pull/963.

Also includes Initial import of <linux/rtc.h> related stuff https://github.com/koreader/koreader-base/pull/965


pazos:
drop legacy code. I like it!

Frenzie:
PS Rationale in <https://github.com/koreader/koreader-base/issues/828>.

Issue #5323, opened 2019-09-05
A couple of PDF/CBZ crashes


NiLuJe:
* KOReader version: current nightly
* Device: Forma (but possibly irrelevant)

#### Issue

A few `attempt to index local 'bbox' (a nil value)` crashes.

#### Steps to reproduce

First crash happened when trying to enable reflow in a PDF in paged mode (I think, might have been continuous).

Second crash happened in a CBZ in continuous mode, across two different pages, with two different gutter sizes, when trying to enable "auto" cropping (instead of semi-auto).

##### `crash.log` (if applicable)

PDF:
```
./luajit: frontend/document/koptinterface.lua:133: attempt to index local 'bbox' (a nil value)
stack traceback:
        frontend/document/koptinterface.lua:133: in function 'getContextHash'
        frontend/document/koptinterface.lua:190: in function 'getPageBBox'
        frontend/document/document.lua:230: in function 'getUsedBBoxDimensions'
        frontend/apps/reader/modules/readerzooming.lua:246: in function 'getZoom'
        frontend/apps/reader/modules/readerzooming.lua:333: in function 'setZoom'
        frontend/apps/reader/modules/readerzooming.lua:219: in function 'handleEvent'
        frontend/ui/widget/container/widgetcontainer.lua:88: in function 'propagateEvent'
        frontend/ui/widget/container/widgetcontainer.lua:106: in function 'handleEvent'
        frontend/apps/reader/modules/readerview.lua:824: in function 'handleEvent'
        frontend/ui/widget/container/widgetcontainer.lua:88: in function 'propagateEvent'
        frontend/ui/widget/container/widgetcontainer.lua:106: in function 'handleEvent'
        frontend/ui/widget/configdialog.lua:872: in function 'onConfigEvent'
        frontend/ui/widget/configdialog.lua:918: in function 'action'
        frontend/ui/uimanager.lua:746: in function '_checkTasks'
        frontend/ui/uimanager.lua:1029: in function 'handleInput'
        frontend/ui/uimanager.lua:1115: in function 'run'
        ./reader.lua:260: in main chunk
        [C]: at 0x000140c1
```

CBZ:
```
./luajit: frontend/document/document.lua:232: attempt to index local 'bbox' (a nil value)
stack traceback:
        frontend/document/document.lua:232: in function 'getPageArea'
        frontend/apps/reader/modules/readerview.lua:570: in function 'recalculate'
[...]

PR #5328, opened 2019-09-06
[Kobo] Restart KOReader after a crash

NiLuJe:
Because getting booted back to Nickel is mildly annoying ;).

(Keeps track of crashes to avoid getting stuck in boot loops).

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -223,6 +223,15 @@ while [ $RETURN_VALUE -ne 0 ]; do
             CRASH_COUNT=1
         fi
 
+        # Check if the user requested to always abort on crash
+        if grep -q '\["dev_abort_on_crash"\] = true' 'settings.reader.lua' 2>/dev/null; then
+            ALWAYS_ABORT="true"
+            # In which case, make sure we pause on *every* crash
+            CRASH_COUNT=1
+        else
+            ALWAYS_ABORT="false"
+        fi
+
         # Show a fancy bomb on screen
         viewWidth=600
         viewHeight=800
@@ -234,7 +243,7 @@ while [ $RETURN_VALUE -ne 0 ]; do
         ./fbink -q -b -c -B GRAY9 -m -y 1 "Don't Panic! (Crash n°${CRASH_COUNT} -> ${RETURN_VALUE})"
         if [ ${CRASH_COUNT} -eq 1 ]; then
             # Warn that we're waiting on a tap to continue...
-            ./fbink -q -b -O -m -y 2 "Tap the screen to continue!"
+            ./fbink -q -b -O -m -y 2 "Tap the screen to continue."
         fi
         # U+1F4A3, the hard way, because we can't use \u or \U escape sequences...
         # shellcheck disable=SC2039
@@ -248,7 +257,7 @@ while [ $RETURN_VALUE -ne 0 ]; do
 
         echo "!!!!" >>crash.log 2>&1
         echo "Hu oh, something went awry... (Crash n°${CRASH_COUNT}: $(date +'%x @ %X'))" >>crash.log 2>&1
-        if [ $CRASH_COUNT -lt 5 ]; then
+        if [ $CRASH_COUNT -lt 5 ] && [ "${ALWAYS_ABORT}" = "false" ]; then
             echo "Attempting to restart KOReader . . ." >>crash.log 2>&1
             echo "!!!!" >>crash.log 2>&1
         fi
@@ -264,8 +273,16 @@ while [ $RETURN_VALUE -ne 0 ]; do
         CRASH_PREV_TS=${CRASH_TS}
 
         # But if we've crashed more than 5 consecutive times, exit, because we wouldn't want to be stuck in a loop...
+        # NOTE: No need to check for ALWAYS_ABORT, CRASH_COUNT will always be 1 when it's true ;).
         if [ $CRASH_COUNT -ge 5 ]; then
             echo "Too many consecutive crashes, aborting . . ." >>crash.log 2>&1
+            echo "!!!! ! !!!!" >>crash.log 2>&1
+            break
+        fi
+
+        # If the user requested to always abort on crash, do so.
+        if [ "${ALWAYS_ABORT}" = "true" ]; then
+            echo "Aborting . . ." >>crash.log 2>&1
             echo "!!!! ! !!!!" >>crash.log 2>&1
             break
         fi
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -232,7 +232,7 @@ while [ $RETURN_VALUE -ne 0 ]; do
         ./fbink -q -c -B GRAY9 -m -y -5 "Don't Panic! (Crash n°${CRASH_COUNT} -> ${RETURN_VALUE})"
         # U+1F4A3, the hard way, because we can't use \u or \U escape sequences...
         # shellcheck disable=SC2039
-        ./fbink -q -m -t regular=./fonts/freefont/FreeSerif.ttf,px=${bombHeight},top=${bombMargin} $'\xf0\x9f\x92\xa3' -O
+        ./fbink -q -O -m -t regular=./fonts/freefont/FreeSerif.ttf,px=${bombHeight},top=${bombMargin} $'\xf0\x9f\x92\xa3'
         # Cue a lemming's faceplant sound effect!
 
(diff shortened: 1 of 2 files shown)


NiLuJe:
Looks something like that in action:

gsod

With KFMon's replacement boot progress bar on top:

gsod2

poire-z:
Not so fond of the idea, but may be that's because I'm a dev - while users may like it.
I use KSM, and when it crashes, I get back to KSM with some lower light level (that's my grey screen of death :), and I can just hit "start koreader".

Additionally, depending on what I suspect caused the crash (if I'm testing stuff), I do not want to restart koreader. Possibly because then I would just be waiting 15s before being able to get back to usb, and it would fill crash.log with multiple instances of the crash.
And it could possibly overwrite/rotate the settings.reader.lua / book's metadata.lua.old and the previous OK one would be lost (while no-restart would not trash the settings.reader.lua.old or metadata.lua.old and you'll be able to recover them, in a state before some setting change that could be the cause of the crash).
Why 5 restarts ? If the first one crashes too, there's probably no reason for the 2nd, 3rd, 4th and 5th to work better :)

I'll probably patch that out for my own safer usage (may be you could make that an option ? in developpers options?).

(But may be I'd like a 1-time restart, with an excerpt of the crash log stacktrace on your grey screen :) so I know what next step to take: restart or go fix that over usb :)

Frenzie:
@poire-z I think the 5 restarts might be meant to be accumulative over the course of weeks. If so, it might be better to store the date and see if the last two restarts occurred within the last 10 seconds or so.

Frenzie:
PS 2019.10 so that 2019.09 can go out this weekend. ;-)

NiLuJe:
@Frenzie: That's exactly what this does ^^. Plus, any "expected" exit (i.e., a KOReader restart via the menu) will reset the counter.

@poire-z: The idea behind the 5 (well, in practice, 6) crash limit is to avoid getting stuck in a loop in the unlikely event we crash on startup.
In which case, the delay only gets applied to the *first* of those crashes, because, yeah, that'd be annoying otherwise ;). (And the full loop ends up tallying neatly at right below the 5s grace period). 

The vast majority of crashes do *NOT* happen on startup, though, and in 99% of the cases, you will probably restart KOReader straight away ;).

That said, I do get the point about being able to just die on crash as usual, I'll look into it ;).

NiLuJe:
@poire-z: And I was indeed thinking about how best to mangle a relevant bit of the tail of the crash.log somewhere around the bottom of that grey screen ;).

Frenzie:
@NiLuJe My point is more that if you're running the same session for several months you could rack up a total of 5 crashes, each of which might've occurred weeks apart.

NiLuJe:
@NiLuJe: Oh, right, I see. Yeah, that makes sense ;).

NiLuJe:
Okay, log printing is in. Caught an FBInk bug in the process, yay :D.

(Hi, #5323 ;p)

crashy

platform/kobo/koreader.sh
```diff
@@ -218,28 +218,33 @@ while [ $RETURN_VALUE -ne 0 ]; do
...
         bombMargin=$((viewWidth/30))
-        # With a little notice at the bottom of the screen (same line as KFMon), on a big gray screen of death ;).
-        ./fbink -q -b -c -B GRAY9 -m -y -5 "Don't Panic! (Crash n°${CRASH_COUNT} -> ${RETURN_VALUE})"
+        # With a little notice at the top of the screen, on a big gray screen of death ;).
+        ./fbink -q -b -c -B GRAY9 -m -y 1 "Don't Panic! (Crash n°${CRASH_COUNT} -> ${RETURN_VALUE})"
         # U+1F4A3, the hard way, because we can't use \u or \U escape sequences...
         # shellcheck disable=SC2039
         ./fbink -q -b -O -m -t regular=./fonts/freefont/FreeSerif.ttf,px=${bombHeight},top=${bombMargin} $'\xf0\x9f\x92\xa3'
+        # And then print the tail end of the log on the bottom of the screen...
+        crashLog="$(tail -n 25 crash.log)"
```
Frenzie: Fancy, although if it's only for 2 seconds you might have to be a pretty fast reader. ;-)

NiLuJe: Oh, yeah, good point, delay should be higher now that it does this ;p.

poire-z: Yep :) Ideally, I'd like some button dialog on that, that waits for my choice: `Exit | Restart`, so we get the time to read it and decide. Can fbink do that ? :)

(After a crash, I usually read the crash.log with a KSM script in KSM own clumsy text viewer, or launch Koreader and view it in TextEditor - so that fbink display would avoid that need.)

NiLuJe: It's... tricky because of the various crappy input protocol handling on Kobos.

What might be doable is a *single* button, saying "Tap screen to <restart|exit KOReader>", the context 'i.e., restart or exit) being the position in the crash counter.
In which case we only need to listen for the first EV_KEY (or whatever) event that makes sense, and a tap anywhere would do the trick.

poire-z: If I'm able to just hack that counter to make my tap be always _too many already_, and make it always exit, I'm super fine with that :)

NiLuJe: I went with a "tap to continue" on the *first* crash in a series, which should workout well enough in practice ;).

NiLuJe: @poire-z: You raise an interesting conundrum: I *can* add an option in the Dev menu, but what it actually does is up for grabs:

I was originally thinking of simply reverting back to the previous behavior.
But I could also (and, in fact, it's probably going to be tiny bit easier) make it always print the log, wait for tap & *exit*.

NiLuJe: (essentially, check that dev setting, if it's on, always enforce the counter to 1, and at the end, just break).

poire-z: superfinewithme :)

NiLuJe:
Okay, dev option is in!

Final? iteration looks like this:

crashy

NiLuJe:
There are a few other fun icons available, FWIW.

I settled with the bomb because it made me laugh, and (somehow, despite me barely ever having used a classic Mac) reminded me of the old Mac OS system error popup...

In the same vein, there's a "Fire" icon, which could be a nod to HCF/Halt and Catch Fire.

Or less nerdy stuff like skull & bones or something ;p.

poire-z:
Just looked at [some page listing some emoticons](https://apps.timwhitlock.info/emoji/tables/unicode) where I found yours... and there's not many alternatives.
This one is nice in full massive black, except that it expresses that a bomb _will_ trigger - while the crash _has already happened_ :)

Frenzie:
On that note… https://fransdejonge.com/2018/09/im-not-sure-what-to-think-of-the-vivaldi-crash-icon/

NiLuJe:
There's also a "kapow" like explosion outline, but getting the two to align was tricky, so I gave up ;p.

platform/kobo/koreader.sh
```diff
@@ -196,15 +196,100 @@ if [ -e crash.log ]; then
...
+        ./fbink -q -b -O -m -t regular=./fonts/freefont/FreeSerif.ttf,px=${bombHeight},top=${bombMargin} $'\xf0\x9f\x92\xa3'
+        # And then print the tail end of the log on the bottom of the screen...
+        crashLog="$(tail -n 25 crash.log | sed -e 's/\t/    /g')"
+        ./fbink -q -b -O -t regular=./fonts/droid/DroidSansMono.ttf,top=$((viewHeight/2 + viewWidth/20)),left=$((bombMargin / 2)),right=$((bombMargin / 2)),size=6 "${crashLog}"
+        # So far, we hadn't triggered an actual screen refresh, do that now, to make sure everything is bundled in a single flashing refresh.
+        ./fbink -q -f -s top=0,left=0
+        # Cue a lemming's faceplant sound effect!
+
+        echo "!!!!" >>crash.log 2>&1
+        echo "Hu oh, something went awry... (Crash n°${CRASH_COUNT}: $(date +'%x @ %X'))" >>crash.log 2>&1
```
Frenzie: Btw, is this on purpose or a typo? ;-P

NiLuJe: Which part? ^^

pazos:
if I had a Kobo I would try to crash it just to see that screen. :heart:

Frenzie:
https://xkcd.com/2200/

Issue #4551, opened 2019-02-05
Android eink e-readers to add


Frenzie:
## Intro

KOReader has support for some eink update routines. These routines are not device specific, but platform specific. 

In order to add support for newer devices there's a tool included with the app. You can start the tool from KOReader file manager in "Tools" -> "More Tools" -> "Developer options" -> "Start eink test".

## Usage

![einkTest-1 2](https://user-images.githubusercontent.com/975883/59360943-51fd9d00-8d31-11e9-979f-01c086b9fb2e.jpg)

If you get results with the "tolino" update please answer this question too:

> does KOReader works normally (without full screen updates) or it seems stuck and not responding?

### 1. I did run all the tests but no one is working for me. Shall I report my results?

Well, you can but it will stay ignored until a developer adds support for a new routine compatible with your device.

### 2. One of the test works for me but (latest version of) KOReader does not refresh at all. Heeelp!!

You're in the right place. Please report your test results and we can add support on the next KOReader nightly.

### 3. I have an onyx device which seems to work with the eink test utility

Onyx e-ink devices have their own ticket in https://github.com/koreader/koreader/issues/5073. We can't add new supported devices until these issues are fixed.

<details>
<summary>
<b>original message</b>
</summary>
<p> I've been testing the screen refresh on my android e-readers (v2019.02-8-g3f41560_2019-02-04). It works on the Onyx C67 but not on the InkBook Prime HD (Boyue clone).  However, the "einkTest" application does run the screen refresh.  This is the information that appears when you run einkTest:

Manufacturer: ARTATECH
Brand: inkBook
Model: Prime HD
Product:  Prime HD
Hardware: rk30board

_Originally posted by @Cellaris in https://github.com/koreader/koreader/pull/4517#issuecomment-460712068_

</p>
</details>

Frenzie:
I was trying out GitHub's relatively new "create issue from comment" feature btw, hence the slightly odd look for the OP.

@Cellaris Tomorrow's nightly will include support. Nightlies can be found at http://build.koreader.rocks/download/nightly/

ptrm:
I have some curious spieces of InkBook Prime HD, with space missing in the model name:

![20190207_154710](https://user-images.githubusercontent.com/366258/52419104-b2649f80-2aef-11e9-9d35-85f04481adb4.jpg)

Normal and forced updates work fine, 5d21990 commit release fails, however.

ptrm:
[...]

Issue #4291, opened 2018-10-28
Kobo Forma support


patrick-tang:
* KOReader version:
koreader-kobo-arm-linux-gnueabihf-v2015.11-stable
(launch with KFMon-v1.2.6-1-g9ab7d3a)
* Device:
Kobo Forma
(firmware version: N782890001425,4.1.15,4.11.11911,4.1.15,4.1.15,00000000-0000-0000-0000-000000000377)

#### Issue
In new released Kobo Forma, when launching Koreader, it crashed right away, reporting "unrecognized Kobo model frost".

Complete detailed log:
[crash.log](https://github.com/koreader/koreader/files/2523197/crash.log)

#### Steps to reproduce
After installed, try to launch Koread, it will crash with above crash log.

##### `crash.log` (if applicable)
./luajit: frontend/device/kobo/device.lua:177: unrecognized Kobo model frost
stack traceback:
	[C]: in function 'error'
	frontend/device/kobo/device.lua:177: in main chunk
	[C]: in function 'probeDevice'
	frontend/device.lua:37: in main chunk
	[C]: in function 'require'
	frontend/ui/uimanager.lua:1: in main chunk
	[C]: in function 'require'
	./reader.lua:92: in main chunk
	[C]: at 0x0000bea5

Frenzie:
In frontend/device/kobo/device.lua, you could try adding a definition:

Here's an example:

https://github.com/koreader/koreader/blob/c01d105bb31df51850ce67194842c841682ccf7e/frontend/device/kobo/device.lua#L185-L199

Sorry for the brevity, but I have to be off now. :-)

NiLuJe:
When a Kobo asks you to re-login, that wipes the DB, IIRC. Which means you may need to reboot it for it to process stuff again, provided it hasn't done a bonus factory reset, which would explain 3., as that reinstalls the factory FW, but not 2., because I *think* that wipes the user partition, too.

If it actually did a factory reset, tthat's a fairly visible process, though, so you'd have noticed the process happening on screen.

On the other hand, if that was a simple reboot and it just asked you to re-login just after that, that was simply a trashed database, which can happen (or be misdetected as "useful", even when it isn't), after bad crashes.

Sidebar: starting with the extremely old "stable" release was also a fairly bad idea ;).

NiLuJe:
In the meantime, if you got another crash.log from the hard crash, that might be useful, too ;).

patrick-tang:
[...]

PR #4283, opened 2018-10-22
refactor sysfs_light

pazos:
syfs_light was written originally for Kobo Aura One but works fine on other devices w/ natural light (like other new kobos and Bq cervantes 4).

Some devices don't have a green led and /dev/null is used as the sysfs_path, causing log errors trying to write to /dev/null/bl_power and the like.

This PR moves sysfs_light outside kobo realm (because I'm going to use in cervantes) and removes the Kobo Aura One values hardcoded in the file. It doesn't try to write to nil dirs, so errors in log are also removed.

PR #5331, opened 2019-09-06
WIP: [Android] Boox NovaPro sys natural backlight support

ColinKinloch:
I recently bought a Boox Nova Pro (Android eInk device).
It doesn't respect the android system brightness settings.
It has two frontlight types white and orange at `/sys/class/backlight/white/brightness` and `/sys/class/backlight/warm/brightness` respectively.
Much of the code in this merge request is copied from the kobo frontend.
I feel much of the auto warmth code should be move to the generic frontend.
Any pointers on how this code could be improved would be appreciated.

`white/device/uevent` and `warm/device/uevent` contain:
```
DRIVER=lm3630a_bl
OF_NAME=lm3630a
OF_FULLNAME=/i2c@ff140000/lm3630a@38
OF_COMPATIBLE_0=onyx,lm3630a
OF_COMPATIBLE_N=1
MODALIAS=i2c:lm3630a
```

https://github.com/torvalds/linux/blob/master/Documentation/ABI/stable/sysfs-class-backlight
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/video/backlight/lm3630a_bl.c
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/video/backlight/backlight.c

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5331)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -107,6 +107,15 @@ local Device = Generic:new{
     should_restrict_JIT = true,
 }
 
+local BooxNovaPro = Device:new{
+    model = "Onyx_Boox_NovaPro",
+    hasNaturalLight = yes,
+    frontlight_settings = {
+        frontlight_white = "/sys/class/backlight/white",
+        frontlight_red = "/sys/class/backlight/warm",
+    },
+}
+
 function Device:init()
     self.screen = require("ffi/framebuffer_android"):new{device = self, debug = logger.dbg}
     self.powerd = require("device/android/powerd"):new{device = self}
@@ -121,6 +130,8 @@ function Device:init()
                 or ev.code == C.APP_CMD_INIT_WINDOW
                 or ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
                 this.device.screen:_updateWindow()
+            elseif ev.code == C.APP_CMD_PAUSE then
+                self.powerd:beforeSuspend()
             elseif ev.code == C.APP_CMD_RESUME then
                 EXTERNAL_DICTS_AVAILABILITY_CHECKED = false
                 if external_dict_when_back_callback then
@@ -143,6 +154,7 @@ function Device:init()
                         require("apps/reader/readerui"):doShowReader(new_file)
                     end)
                 end
+                self.powerd:afterResume()
             end
         end,
         hasClipboardText = function()
@@ -293,4 +305,12 @@ end
 android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
     android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
 
-return Device
+-------------- device probe ------------
+
+local product = android.prop.product
+
+if product == "NovaPro" then
+    return BooxNovaPro
+else
+    return Device
+end
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -112,8 +112,11 @@ local BooxNovaPro = Device:new{
     hasNaturalLight = yes,
     hasLightLevelFallback = no,
     frontlight_settings = {
-        frontlight_white = "/sys/class/backlight/white",
-        frontlight_red = "/sys/class/backlight/warm",
+        frontlight_white = "/sys/class/backlight/white/brightness",
+        frontlight_mixer = "/sys/class/backlight/warm/brightness",
+        nl_min = 0, nl_max = 1,
+        fl_warmth_min = 0, fl_warmth_max = 250,
+        fl_min = 0, fl_max = 250,
         bl_power_off = nil,
         bl_power_on = nil,
     },
@@ -212,6 +215,10 @@ function Device:init()
     local last_value = G_reader_settings:readSetting("fl_last_level")
     if type(last_value) == "number" and last_value >= 0 then
         Device:setScreenBrightness(last_value)
+    end
+
+    if self:hasNaturalLight() and self.frontlight_settings and self.frontlight_settings.frontlight_mixer then
+        self.hasNaturalLightMixer = yes
     end
 
     Generic.init(self)
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -132,8 +132,6 @@ function Device:init()
                 or ev.code == C.APP_CMD_INIT_WINDOW
                 or ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
                 this.device.screen:_updateWindow()
-            elseif ev.code == C.APP_CMD_PAUSE then
-                self.powerd:beforeSuspend()
             elseif ev.code == C.APP_CMD_RESUME then
                 EXTERNAL_DICTS_AVAILABILITY_CHECKED = false
                 if external_dict_when_back_callback then
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -110,6 +110,7 @@ local Device = Generic:new{
 local BooxNovaPro = Device:new{
     model = "Onyx_Boox_NovaPro",
     hasNaturalLight = yes,
+    hasLightLevelFallback = no,
     frontlight_settings = {
         frontlight_white = "/sys/class/backlight/white",
         frontlight_red = "/sys/class/backlight/warm",
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -113,6 +113,8 @@ local BooxNovaPro = Device:new{
     frontlight_settings = {
         frontlight_white = "/sys/class/backlight/white",
         frontlight_red = "/sys/class/backlight/warm",
+        bl_power_off = nil,
+        bl_power_on = nil,
     },
 }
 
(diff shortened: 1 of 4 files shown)


Frenzie:
The generic SysfsLight module was abstracted out of Kobo because Cervantes also supports (cf. <https://github.com/koreader/koreader/pull/4283>). Pinging @pazos who should probably know the most about it.

frontend/device/android/device.lua
```diff
@@ -120,6 +129,8 @@ function Device:init()
                 or ev.code == C.APP_CMD_INIT_WINDOW
                 or ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
                 this.device.screen:_updateWindow()
+            elseif ev.code == C.APP_CMD_PAUSE then
+                self.powerd:beforeSuspend()
```
pazos: Are you sure you want to turn off light when KOReader activity looses focus? Doesn't this affect to other activities?

pazos:
> I recently bought a Boox Nova Pro (Android eInk device).
It doesn't respect the android system brightness settings.
It has two frontlight types white and orange at /sys/class/backlight/white/brightness and /sys/class/backlight/warm/brightness respectively.

Wow, that's weird. Thanks for your time!

> I feel much of the auto warmth code should be move to the generic frontend.

Yeah, some of the powerd code is now triplicated (it was duplicated before between Kobo and Cervantes). I will leave the abstraction out of the equation here unless you want to take a look on other devices powerd implementations. But yeah, something to figure out in the future.

pazos:
BTW @ColinKinloch: are you aware of https://github.com/koreader/koreader/issues/4551#issuecomment-526942909?

The SDK that onyx provides takes care of lights too among eink updates and other things. Feel free to give it a try if you wish. Anyways until somebody integrates the onyx sdk with [https://github.com/koreader/android-luajit-launcher](android-luajit-launcher) this PR can help!

frontend/device/android/device.lua
ColinKinloch: You're right, I was using that in place of a device sleep event.
On sleep the Nova switches off the backlight and doesn't restore it on wake.
I was using it to get past this line:
https://github.com/koreader/koreader/blob/4005bf69aa5667f520a1e837f9c493874cf192fa/frontend/device/generic/powerd.lua#L88
How about I add a function to `SysfsLight` to read backlight state and use that to implement `AndroidPowerD:frontlightIntensityHW`?

ColinKinloch:
The `setBrightness` of the Boox SDK only seems to take a single `unsigned char`.
I just compiled the demo, the buttons that trigger `FrontLightController.setBrightness` have no effect.
:(

ColinKinloch:
`adb shell settings list system`
Lists values related to brightness
```
brightness_state_key=0
cold_brightness_state_key=1
screen_brightness=2
screen_brightness_mode=0
screen_cold_brightness=120
screen_warm_brightness=88
sync_brightness_adjust_key=0
wake_up_brightness_key=0
warm_brightness_state_key=1
```
However they only seem to map to the boox frontlight config dialog.
Setting them with `settings put system screen_cold_brightness 4` doesn't change anything.

frontend/device/android/device.lua
pazos: Anything that works on onyx and doesn't break other devices is actually an improvement. So feel free to code it.

pazos:
> adb shell settings list system lists values related to brightness

Yeah, they're free to `get` but (on normal systems) put a new value requires that the app has granted WRITE_SETTINGS permissions. We just switched to a less intrusive method on https://github.com/koreader/android-luajit-launcher/commit/3750551c6607aa5fa53c77a7095e1dba0b11237b:

It works this way: 
1. by default we're using the same brightness that the rest of the system. (The one that can be retrieved with `adb shell settings get system screen_brightness` and we retrieve using 
[`readSettingScreenBrightness`](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/helper/ScreenHelper.java#L248-L256). It returns an integer between 0 and 255)

2. if the user overwrites system settings we just write the new one [inside the activity window parameters](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/helper/ScreenHelper.java#L89-L115) unless it is out the valid range, in that case the system will use a value less than 0, which means use system settings again.
3. to populate the frontlight widget we use [`getScreenBrightness()`](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/helper/ScreenHelper.java#L79-L87), which sould cover quirks present in 1 and 2.

> On sleep the Nova switches off the backlight and doesn't restore it on wake.

That's weird. Does this happens on every app?

As I said, feel free to workaround this as you like. But I find your current implementation (writtind directly to sys nodes instead of using WRITE_SETTINGS permission) simpler and less hackier.

ColinKinloch:
> Yeah, they're free to get but (on normal systems) put a new value requires that the app has granted WRITE_SETTINGS permissions. We just switched to a less intrusive method on koreader/android-luajit-launcher@3750551:

I'm writing and verifying the values using `adb shell`.
I think the values are just used to maintain state for the Boox frontlight dialog.

ColinKinloch:
> That's weird. Does this happens on every app?

Yes.
It's not mapped to the standard android brightness settings so it would need to be implemented by the developers at Onyx.

pazos:
I just give a quick look at the sdk. It seems that the methods used to retrieve and set light values are (for rk32xx) in https://github.com/buggins/coolreader/blob/master/android/src/com/onyx/android/sdk/device/RK32XXDevice.java#L797-L817

pazos:
About e-ink devices:

It seems that an already supported epd update routine does work on rk32xx devices.

You can test that with [EinkTest](https://github.com/koreader/koreader/issues/4551#issue-406883935). If the method called "ntx" works on your device you can add it to [our list of supported devices](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/device/DeviceInfo.java) as `EINK_FREESCALE` and `EINK_SUPPORT`. This should enable a submenu to change full refresh rate and the hability of requesting a full refresh with a diagonal swipe by default.

ColinKinloch:
The UI on the Nova Pro has two independant sliders to control white and warm frontlights seen [here](https://user-images.githubusercontent.com/334272/64879736-847ba680-d64e-11e9-9504-2885e66a2d03.png).

Initialising / updating the brightness / warmth values from the hardware values for koreader is not a simple problem.

Would a simpler equation, per device equation or an option for a 0 - 255 slider per backlight colour be possible?

https://github.com/koreader/koreader/blob/840d05b3ae1f0622f6fe98b40d4af1b5f1bf108a/frontend/device/sysfs_light.lua#L86-L97

pazos:
> Initialising / updating the brightness / warmth values from the hardware values for koreader is not a simple problem.

@ColinKinloch: you're using (part) of the interface for devices that have separate green and red leds. You might want to replace frontlight_red with frontlight_mixer to apply the code path: https://github.com/koreader/koreader/blob/840d05b3ae1f0622f6fe98b40d4af1b5f1bf108a/frontend/device/sysfs_light.lua#L68-L80

This makes things a lot of easier than calculting warmth with a mix of R/G leds

frontend/device/android/device.lua
```diff
@@ -107,6 +107,18 @@ local Device = Generic:new{
     should_restrict_JIT = true,
 }
 
+local BooxNovaPro = Device:new{
+    model = "Onyx_Boox_NovaPro",
+    hasNaturalLight = yes,
+    hasLightLevelFallback = no,
+    frontlight_settings = {
+        frontlight_white = "/sys/class/backlight/white",
+        frontlight_red = "/sys/class/backlight/warm",
```
pazos: use `frontlight_mixer =  "/sys/class/backlight/warm"` instead

ColinKinloch: On mixer devices does writing to the `/sys/class/backlight/x/color` change the backlight colour or change the brightness of a separate set of LEDs?

The two brightness files on the Nova Pro control independent sets of LEDs.

pazos: We don't do that but probably the system does something like that (ie: adjust brightness intensity based on warmth level). See https://github.com/koreader/koreader/issues/4291#issuecomment-439700537.

pazos:
@ColinKinloch: please ping me when you want a review. I will also ping @NiLuJe since he implented the frontlight mixer and has awesome reviewer skills 😄

frontend/device/android/device.lua
pazos: I feel like I cannot help because I don't know nothing about your device. Do you found some issues with your current implementation?

pazos:
Superseded by #6426 

@ColinKinloch: feel free to submit an updated PR. You can implement your own controller using one of the existent as a template. See [GenericController](https://github.com/koreader/android-luajit-launcher/pull/239/files#diff-436a5e22272fe275214acafff188096c) and [TolinoWarmthController](https://github.com/koreader/android-luajit-launcher/pull/239/files#diff-ca13463e453d301c98d10a062655c87f).

Issue #3806, opened 2018-03-27
Automatic Power Off


lhannest:
This is a feature request.

Backstory: I put my Kobo into sleep mode one night a little while back at forty-something-percent battery, and came back to it in the morning with the battery drained. As far as I know this is really bad for the battery, and I'm annoyed that this happened with a brand new device. Until this bug (#3706) gets fixed, there should be an automatic power off feature to prevent massive battery drainage.

The user should be able to define a threshold after which the device will attempt to automatically power down. Before actually powering down it should prompt the user to disable the feature, then powering down if the user does not respond after some time.

cramoisi:
The fact is : [there is one](https://github.com/koreader/koreader/blob/e21fe5504299f2c650939da22adefb73aa4f0f2b/plugins/autosuspend.koplugin/main.lua#L36). The kobo goes to sleep after 60 minutes of inactivity. But that's the problem : for inactivity to be registered, it need to be inactivity.

Anyway, you could help gather infos about this. What about putting your kobo in debug mode ( "debug" = true, to your settings.reader.lua ) and adding a log as described [there](https://github.com/koreader/koreader/issues/3706#issuecomment-370152234) ? Perhaps you could describe your settings too ? (wifi enabled ? sleepcover ? )

For letting it sleeps overnight, I noticed I don't have the problem if there's no sleepcover and  nothing touching the screen while it sleeps. (There is also the temporary solution to power it off when unused)

AlanSP1:
@cramoisi I think he thinks powering off, i.e. completely powering device off, not sleep mode.

It can be useful, sometimes we don't use our device for days and then could be useful it powers itself off.

cramoisi:
@AlanSP1 yup you're right. I miss the point because the bug is related to sleep...

Frenzie:
> It can be useful, sometimes we don't use our device for days and then could be useful it powers itself off.

Disagree, battery drain in sleep is pretty minimal even after weeks. Besides, how would the device know I'm not going to use it tomorrow? If it did more than sleep I'd just get annoyed. But that aside. Are we sure this isn't already an option in principle? :-P

But sure, as a last resort someone could write an addition that says if we're still not sleeping after 2*sleep period then shutdown. Although one might argue that if you can write that you can probably debug and solve whatever the problem is. :-)

AlanSP1:
I looked at it from @lhannest perspective, and it can be useful if our reading habits are that we read a lot in certain periods, and don't read at all in different periods.

[...]

Issue #3855, opened 2018-04-10
Battery still drains after a few hours


melochale:
* KOReader version: koreader-beta-20180410
* Device: Kobo aura one (with magnet cover)

#### Issue
I installed the latest koreader(koreader-beta-20180410) once i saw a new update because last one had a issue about battery draining bug, but after a few hours test, it's still there. my battery was from 95% down to 53% in 3~5 hours!

Frenzie:
I think the Kobo Aura One never suffered from that bug in the first place. Pinging @cramoisi  Do you have a crash.log? It should presumably contain some lines like `Kobo suspend: putting device back to sleep, unexpected wakeups:`

Also, what does it say in System statistics → sleeps? Is it a normal number (say less than 50, maybe 100) or something outrageous?

melote:
Seems fine about the sleep times(only 10 sleeps), I will fully charge the device than make more tests, then report to you.

cramoisi:
@Frenzie KA1 didn't suffer of this very bug with something touching the screen. I'm pretty sure of it as I compared the two device side by side due to the capacitive touchscreen KA1 and it's properties.

 But it's true that the KA1 battery doesn't do as well as the gloHD one with koreader. I read a lot every days and reading two hours straight drains something significant (I will make note about that during my lunch reading session). Probably because it's the same battery ( ? ) in the two device and the KA1 has more screen, etc to power. In the case of @melote I would think the device didn't sleep at all during the night :) A crashlog (and perhaps with a device in debug mode) should be interesting to look at.

cramoisi:
It's more or less my estimation. for info, what's the proc % usage average (during effective reading) for koreader ? does it varies significantly between devices?

Frenzie:
I'm not completely sure what you mean, but I would assume so. The program doesn't really do much when you don't. So given the same CPU, a bigger screensize means it'll presumably spend roughly the screensize factor in extra time spent rendering and drawing while turning pages.

There are some compilations there, like a higher DPI might mean exponentially more time spent rendering finer glyph details.

hwhw:
Zero, basically and mostly, except maybe when a user is pressing all kinds of keys in a panic :-)

Processor usage stat is meaningless as an average over time. It would probably be more interesting to see the absolute values of the CPU time.

cramoisi:
ok, I asked that because on KA1, I usually see something around 15-20% if I close a book and go in system stat. (And I have disabled most of the plugins).

[...]

PR #5335, opened 2019-09-07
[feat, Kobo] Autoshutdown

Frenzie:
The methods used here will likely work on most embedded devices, which is why I put them in UIManager.

See <https://www.mobileread.com/forums/showthread.php?p=3885836#post3885836> for more context.

Fixes <https://github.com/koreader/koreader/issues/3806>.


plugins/autosuspend.koplugin/main.lua
```diff
@@ -141,34 +153,55 @@ function AutoSuspendWidget:addToMainMenu(menu_items)
...
+                -- About a minute, good for testing and battery life fanatics.
+                -- Just high enough to avoid an instant shutdown death scenario.
+                value_min = 0.0007,
+                -- More than three weeks seems a bit excessive if you want to enable authoshutdown,
+                -- even if the battery can last up to three months.
+                value_max = 21,
+                value_hold_step = 3,
+                precision = "%.2f",
+                ok_text = _("Set timeout"),
+                title_text = _("Timeout in days"),
```
poire-z: Why not timeout in hours? Easier to compute that 3 days = 72 hours - than compute that 4 hours is 0.17 days.

frontend/ui/uimanager.lua
```diff
@@ -478,6 +480,146 @@ end
 dbg:guard(UIManager, 'unschedule',
     function(self, action) assert(action ~= nil) end)
 
+--- @todo Stick this in base/ffi and reference here?
```
poire-z: Yep, too much C/FFI here :) when there is none already.

Haven't checked much the autosuspend plugin, that I have disabled (so, I won't benefit from autoshutdown, which is fine for me). Haven't checked much of the ffi/rtc stuff either, but reading the other links, I trust you got that right :)

Is that uimanager alarm management abstracted enough from its shutdown use so it can eventually be used for other features (like update rss/news, autoupdate koreader,...)?

Frenzie: > Yep, too much C/FFI here :) when there is none already.

Much easier to prototype. ;-) Actually this is the abstracted version of the proof of concept done purely in kobo/device. For the moment I've still got a branch with that [here](https://github.com/Frenzie/koreader/compare/7019ff1...ba58403). Perhaps I should add a diff here for posterity. ^_^

> Is that uimanager alarm management abstracted enough from its shutdown use so it can eventually be used for other features (like update rss/news, autoupdate koreader,...)?

Sure, it's (purposefully) completely agnostic as to what's done with the wakeup. I'll update it with a wakeup_action or some such later to finalize that.

plugins/autosuspend.koplugin/main.lua
Frenzie: Perhaps, but except for testing any value under 24 hours is basically useless. And I don't find 2 weeks = 336 hours particularly easy tbh. :-)

Although on second thought it might have some use as a workaround for issues like #3855.

Frenzie:
Btw @NiLuJe if you woudn't mind could you add these to base the right way? :-)

```
time_t mktime(struct tm *tm);
double difftime(time_t time1, time_t time0);
```

frontend/ui/uimanager.lua
Frenzie: Although I should add that unless someone writes a wakeup scheduler that manages things, this implementation is one wakeup action only. So that would be either update Wallabag/whatever or autoshutdown. But that's basically unrelated to the methods for scheduling the wakeup.

To clarify, you can only schedule one wakeup. So if you wake up to sync, then afterward you have to schedule either the next sync or the autoshutdown depending on which would come first. I suppose the scheduler could be as simple as a table of upcoming actions in proper order, holding a reference to the scheduled action.

I have no interest in implementing such a thing right now, and I think I'd rather keep it for a follow-up PR even if I do it later.

Frenzie: Diff for the part of the v1 prototype that didn't carry over into this PR:

```diff
diff --git a/frontend/device/kobo/device.lua b/frontend/device/kobo/device.lua
index 35edf5d3..6a124352 100644
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -534,7 +534,13 @@ local function getProductId()
 end
 
 local unexpected_wakeup_count = 0
+local wakeup_scheduled = false
 local function check_unexpected_wakeup()
+    if wakeup_scheduled then
+        logger.dbg("Kobo suspend: scheduled wakeup for shutdown")
+        local UIManager = require("ui/uimanager")
+        UIManager:poweroff_action()
+    end
     logger.dbg("Kobo suspend: checking unexpected wakeup:",
                unexpected_wakeup_count)
     if unexpected_wakeup_count == 0 or unexpected_wakeup_count > 20 then
@@ -553,7 +559,49 @@ end
 
 function Kobo:getUnexpectedWakeup() return unexpected_wakeup_count end
 
+local ffi = require("ffi")
+local C = ffi.C
+-- for ioctl header definition:
+local dummy = require("ffi/posix_h")
+local rtc = require("ffi/rtc_h")
+
 function Kobo:suspend()
+    local t = ffi.new("time_t[1]")
+    t[0] = C.time(NULL)
+    t[0] = t[0] + 12
+    print(t[0])
+
+    local ptm = ffi.new("struct tm")
+    ptm = C.gmtime(t)
+    print(ptm.tm_sec)
+
+    local wake = ffi.new("struct rtc_wkalrm")
+    wake.time.tm_sec = ptm.tm_sec
+    wake.time.tm_min = ptm.tm_min
+    wake.time.tm_hour = ptm.tm_hour
+    wake.time.tm_mday = ptm.tm_mday
+    wake.time.tm_mon = ptm.tm_mon
+    wake.time.tm_year = ptm.tm_year
+    -- wday, yday, and isdst fields are unused by Linux
+    wake.time.tm_wday = -1;
+    wake.time.tm_yday = -1;
+    wake.time.tm_isdst = -1;
+    print(wake.time.tm_sec)
+
+    wake.enabled = 1
+    --wake.time = rtc_time
+
+    local rtc0 = C.open("/dev/rtc0", C.O_RDONLY)
+    print(rtc0, ffi.string(C.strerror(ffi.errno())))
+    local err = C.ioctl(rtc0, C.RTC_WKALM_SET, wake)
+    print(err, ffi.string(C.strerror(ffi.errno())))
+    local err2 = C.close(rtc0)
+    print(err2, ffi.string(C.strerror(ffi.errno())))
+
+    if err == 0 then
+        wakeup_scheduled = true
+    end
+
     logger.info("Kobo suspend: going to sleep . . .")
     local UIManager = require("ui/uimanager")
     UIManager:unschedule(check_unexpected_wakeup)
```

frontend/ui/uimanager.lua
```diff
@@ -478,6 +480,146 @@ end
...
+    local alarm_epoch
+    local alarm_sys = self:getWakeupAlarmSys()
+    local alarm_sys_epoch
+
+    -- this seems a bit roundabout
+    local current_time = ffi.new("time_t[1]")
+    current_time[0] = C.time(nil)
+    local current_time_epoch = C.difftime(C.mktime(C.gmtime(current_time)), 0)
+
+    if not alarm and alarm_sys then return end
```
Frenzie:
```suggestion
    if not (alarm and alarm_sys) then return end
```

NiLuJe:
> Btw @NiLuJe if you woudn't mind could you add these to base the right way? :-)
> 
> ```
> time_t mktime(struct tm *tm);
> double difftime(time_t time1, time_t time0);
> ```

Done! :)

Sidebar: `difftime` is possibly overkill on linux ([time_t is an int64_t](https://github.com/NiLuJe/kfmon/blob/master/kfmon.c#L1526)) ;).

frontend/ui/uimanager.lua
```diff
@@ -478,6 +480,146 @@ end
...
+    wake.time.tm_year = ptm.tm_year
+    -- wday, yday, and isdst fields are unused by Linux
+    wake.time.tm_wday = -1
+    wake.time.tm_yday = -1
+    wake.time.tm_isdst = -1
+
+    wake.enabled = enabled and 1 or 0
+
+    local err
+    local rtc0 = C.open("/dev/rtc0", C.O_RDONLY)
```
NiLuJe: I'd possibly at least add `| O_NONBLOCK` in there, just because ;p.

i.e.,

```lua
local bor = bit.bor
local rtc0 = C.open("/dev/rtc0", bor(C.O_RDONLY, C.O_NONBLOCK))
```

Frenzie: If you wrap it like this then it can be committed right from GH. ;-)
````
```suggestion
    local rtc0 = C.open("/dev/rtc0", C.O_RDONLY)
```
````

Comment (NiLuJe) [reply frontend/ui/uimanager.lua]: Yeah, but ideally the bit.bor local should go way earlier ;p.

Good to know, though, I'd never bothered to look into the proper syntax for those ;).

Comment (Frenzie) [reply frontend/ui/uimanager.lua]: True, but I'll transfer the FFI stuff to something base/ffi/rtc.lua in any case. :-)

**Comment by Frenzie** [comment]
What's this pipefail nonsense on CircleCI all of a sudden?

File: frontend/device/wakeupmgr.lua [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+        end
+    end
+end
+
+function WakeupMgr:wakeupAction()
+    if #self._task_queue > 0 then
+        local task = self._task_queue[1]
+        if self:validateWakeupAlarmByProximity(task.epoch) then
+            task.callback()
+            self:removeTask(1)
```
Frenzie: And schedule the next one on the list. Something as simple as this should do the trick.

```lua
self:setWakeupAlarm((self._task_queue[1] or {}).epoch)
```

frontend/device/wakeupmgr.lua
```diff
@@ -0,0 +1,143 @@
...
+            return true
+        else
+            return false
+        end
+    end
+end
+
+function WakeupMgr:setWakeupAlarm(seconds_from_now, enabled)
```
Frenzie:
```suggestion
function WakeupMgr:setWakeupAlarm(epoch, enabled)
```

frontend/device/wakeupmgr.lua
```diff
@@ -0,0 +1,143 @@
...
+    return RTC:setWakeupAlarm(seconds_from_now, enabled)
```
Frenzie: (Just a variable name here, I changed the functionality around a little since I wrote the function.)
```suggestion
    return RTC:setWakeupAlarm(epoch, enabled)
```

plugins/autosuspend.koplugin/main.lua
```diff
@@ -11,11 +11,17 @@ local UIManager = require("ui/uimanager")
...
+local default_autoshutdown_timeout_seconds = 3*24*60*60
+
+local AutoSuspend = WidgetContainer:new{
+    name = "autosuspend",
+    is_doc_only = false,
+    autoshutdown_sec = G_reader_settings:readSetting("autoshutdown_timeout_seconds") or default_autoshutdown_timeout_seconds,
     settings = LuaSettings:open(DataStorage:getSettingsDir() .. "/koboautosuspend.lua"),
-    settings_id = 0,
     last_action_sec = os.time(),
+    _current_task = nil,
```
Frenzie: Unused?

spec/unit/wakeupmgr_spec.lua
```diff
@@ -0,0 +1,46 @@
...
+        assert.stub(WakeupMgr.setWakeupAlarm).was.called(1)
+
+        WakeupMgr:addTask(123, function() end)
+        assert.is_equal(epoch2, WakeupMgr._task_queue[1].epoch)
+        assert.stub(WakeupMgr.setWakeupAlarm).was.called(2)
+    end)
+    it("should execute top task", function()
+        assert.is_true(WakeupMgr:wakeupAction())
+    end)
+    it("should have removed executed task from stack", function()
```
Frenzie: And should've called WakeupMgr.setWakeupAlarm

plugins/autosuspend.koplugin/main.lua
```diff
@@ -141,34 +147,58 @@ function AutoSuspendWidget:addToMainMenu(menu_items)
...
+                        timeout = 3,
                     })
+                    self:_unschedule()
+                    self:_start()
+                end
+            }
+            UIManager:show(autosuspend_spin)
+        end,
+    }
+    if not (Device:isKobo() or Device:isEmulator()) then return end
```
Frenzie: Should use canPowerOff, not isKobo
 https://github.com/koreader/koreader/blob/07116b98982bc65a158e877e4a001a8563b0ca82/frontend/device/generic/device.lua#L46

plugins/autosuspend.koplugin/main.lua
```diff
@@ -43,90 +49,90 @@ function AutoSuspend:_enabled()
     return self.auto_suspend_sec > 0
 end
 
-function AutoSuspend:_schedule(settings_id)
+function AutoSuspend:_enabledShutdown()
+    return Device:isKobo() and self.autoshutdown_sec > 0
```
Frenzie:
```suggestion
    return Device:canPowerOff() and self.autoshutdown_sec > 0
```

plugins/autosuspend.koplugin/main.lua
```diff
@@ -43,90 +49,90 @@ function AutoSuspend:_enabled()
...
 
--- We do not want auto suspend procedure to waste battery during suspend. So let's unschedule it
--- when suspending and restart it after resume.
 function AutoSuspend:onSuspend()
     logger.dbg("AutoSuspend: onSuspend")
-    self:_deprecateLastTask()
+    -- We do not want auto suspend procedure to waste battery during suspend. So let's unschedule it
+    -- when suspending and restart it after resume.
+    self:_unschedule()
+    if self:_enabledShutdown() then
```
Frenzie:
```suggestion
    if self:_enabledShutdown() and Device.wakeup_mgr then
```
Or something like that.

plugins/autosuspend.koplugin/main.lua
```diff
@@ -141,34 +147,58 @@ function AutoSuspendWidget:addToMainMenu(menu_items)
...
                 title_text = _("Timeout in minutes"),
                 callback = function(autosuspend_spin)
-                    G_reader_settings:saveSetting("auto_suspend_timeout_seconds", autosuspend_spin.value * 60)
-                    -- NOTE: Will only take effect after a restart, as we don't have a method to set this live...
+                    local autosuspend_timeout_seconds = autosuspend_spin.value * 60
+                    self.auto_suspend_sec = autosuspend_timeout_seconds
+                    G_reader_settings:saveSetting("auto_suspend_timeout_seconds", autosuspend_timeout_seconds)
                     UIManager:show(InfoMessage:new{
-                        text = _("This will take effect on next restart."),
+                        text = T(_("The system will automatically suspend after %1 minutes of inactivity."),
```
Frenzie: Maybe some logic to point out suspend will occur first? Only devices with `Device.wakeup_mgr` (or some equivalent check) will be able to shutdown after suspend.

frontend/ui/uimanager.lua
```diff
@@ -743,7 +747,7 @@ function UIManager:_checkTasks()
             -- task is pending to be executed right now. do it.
             -- NOTE: be careful that task.action() might modify
             -- _task_queue here. So need to avoid race condition
-            task.action()
+            task.action(task.context)
```
poire-z: In your context (uh), this 3rd argument is indeed a "context". But may be, generically, in UIManager, this 3rd argument passed as an arg to the callback could be just named `arg` or `args` (we have quite a few of that here and there).

frontend/device/wakeupmgr.lua
```diff
@@ -0,0 +1,143 @@
...
+For example, maintenance, sync, and shutdown.
+I'm not sure if the distinction between maintenance and sync makes sense
+but it's wifi on vs. off.
+--]]
+function WakeupMgr:addTask(seconds_from_now, callback)
+    if not type(seconds_from_now) == "number" and not type(callback) == "function" then return end
+
+    local epoch = RTC:secondsFromNowToEpoch(seconds_from_now)
+
+    local old_upcoming_task = (self._task_queue[1] or {}).epoch
```
poire-z: I initially stumble on that, "task" being assigned an timestamp, while intuitively, a task is a callback.
So, may be just a comment saying "a task here is a combo of (epoch timestamp, callback).

Nothing more to add about WakeupMrg, looks quite fine and explicite.

Frenzie:
The missing job output issue should be fixed: https://status.circleci.com/incidents/qfhr81gjzkb0

@poire-z AutoSuspend should actually be significantly easier to follow now.

frontend/ui/uimanager.lua
Frenzie: Right, I'll make it args or arguments just like in Event.

plugins/autosuspend.koplugin/main.lua
```diff
@@ -11,10 +11,15 @@ local UIManager = require("ui/uimanager")
...
+local T = require("ffi/util").template
 
-local AutoSuspend = {
+local default_autoshutdown_timeout_seconds = 3*24*60*60
+
+local AutoSuspend = WidgetContainer:new{
+    name = "autosuspend",
+    is_doc_only = false,
+    autoshutdown_sec = G_reader_settings:readSetting("autoshutdown_timeout_seconds") or default_autoshutdown_timeout_seconds,
     settings = LuaSettings:open(DataStorage:getSettingsDir() .. "/koboautosuspend.lua"),
```
poire-z: You save into G_reader_settings. Is there still some reason to read this koboautosuspend.lua and loop thru candidates below? (alt: should you not save them into that plugin setting file, and forger about G_reader_settings here?)

plugins/autosuspend.koplugin/main.lua
```diff
@@ -43,90 +48,90 @@ function AutoSuspend:_enabled()
...
+            UIManager:scheduleIn(delay_suspend, self._schedule, self)
+        end
+        if self:_enabledShutdown() then
+            logger.dbg("AutoSuspend: schedule shutdown at ", os.time() + delay_shutdown)
+            UIManager:scheduleIn(delay_shutdown, self._schedule, self)
+        end
     end
 end
 
-function AutoSuspend:_deprecateLastTask()
```
poire-z: good riddance:)

plugins/autosuspend.koplugin/main.lua
```diff
@@ -43,90 +48,90 @@ function AutoSuspend:_enabled()
...
-    self.settings_id = self.settings_id + 1
-    logger.dbg("AutoSuspend: deprecateLastTask ", self.settings_id)
+function AutoSuspend:_unschedule()
```
poire-z: Now you're talking koreaderish.

plugins/autosuspend.koplugin/main.lua
```diff
@@ -43,90 +48,90 @@ function AutoSuspend:_enabled()
...
--- We do not want auto suspend procedure to waste battery during suspend. So let's unschedule it
--- when suspending and restart it after resume.
 function AutoSuspend:onSuspend()
     logger.dbg("AutoSuspend: onSuspend")
-    self:_deprecateLastTask()
+    -- We do not want auto suspend procedure to waste battery during suspend. So let's unschedule it
+    -- when suspending and restart it after resume.
+    self:_unschedule()
+    if self:_enabledShutdown() and Device.wakeup_mgr then
+        Device.wakeup_mgr:addTask(self.autoshutdown_sec, UIManager.poweroff_action)
```
poire-z: No pb with that, but autoshutdown delay applies after autosuspend delay, so it's actually autosuspend_set + autoshutdown_sec ?
Does all that work when suspend is disabled, and only shutdown enabled?

plugins/autosuspend.koplugin/main.lua
Frenzie: Now that you mention it, one of them is probably intended to be legacy. Given that the plugin-specific settings seem to stem from a migration out of the UIManager core, I'm guessing G_reader_settings is the one that should be deprecated.

plugins/autosuspend.koplugin/main.lua
Frenzie: This was convoluted and hard to follow. The changed method Just Works™ (thanks to the scheduleIn/unschedule methods I wrote in 2014 I might add with some pride).

plugins/autosuspend.koplugin/main.lua
```diff
@@ -43,90 +48,90 @@ function AutoSuspend:_enabled()
...
-    self.settings_id = self.settings_id + 1
-    logger.dbg("AutoSuspend: deprecateLastTask ", self.settings_id)
+function AutoSuspend:_unschedule()
+    logger.dbg("AutoSuspend: unschedule")
+    UIManager:unschedule(self._schedule)
 end
 
 function AutoSuspend:_start()
-    if self:_enabled() then
+    if self:_enabled() or self:_enabledShutdown() then
```
Frenzie: @poire-z This is the answer to your question below in https://github.com/koreader/koreader/pull/5335/files#r322896802

plugins/autosuspend.koplugin/main.lua
Frenzie: The scheduling part above is only for suspending/shutting down while the device is active. On suspend that's all unscheduled (see two lines above), while the WakeupMgr takes care of scheduling the wakeup and executing the scheduled task.

Issue #5327, opened 2019-09-05
Reflow isn´t woring at kindle ppw2


Eduardomb22:
* KOReader version: v2019.08-94-gcb2a3d9_2019-09-04
* Device: ppw2

Reflow isn´t working with any pdf. When I use reflow, koreader crashs and goes to the kindle SO.

                launching...

 It's a scroll... It's a codex... It's KOReader!

 [*] Current time: 09/05/19-13:49:35
 [*] Version: v2019.08-94-gcb2a3d9_2019-09-04

ffi.load: SDL2
ffi.load (warning): libSDL2.so: cannot open shared object file: No such file or directory
ffi.load: libSDL2-2.0.so
ffi.load (warning): libSDL2-2.0.so: cannot open shared object file: No such file or directory
ffi.load: libSDL2-2.0.so.0
ffi.load (warning): libSDL2-2.0.so.0: cannot open shared object file: No such file or directory
SDL2 not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: libSDL2-2.0.so.0
ffi.load: SDL
ffi.load (warning): libSDL.so: cannot open shared object file: No such file or directory
SDL not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: SDL
ffi.load: blitbuffer
FB: computed size: 6684672
FB: id           : mxc_epdc_fb
FB: smem_len     : 6782976
FB: line_length  : 1088
FB: xres         : 1072
FB: xres_virtual : 1088
FB: yres         : 1448
FB: yres_virtual : 6144
FB: bpp          : 8
09/05/19-13:49:35 INFO  initializing for device KindlePaperWhite3
09/05/19-13:49:35 INFO  framebuffer resolution: {
    ["h"] = 1448,
    ["w"] = 1072
}
[ko-input] Forked off fake event generator(pid:21542).
ffi.load: libs/libmupdf.so
ffi.load: libs/libwrap-mupdf.so
ffi.load: libs/libfreetype.so.6
09/05/19-13:49:36 INFO  opening file /mnt/us/documents/Direito TributÃ¡rio - Ricardo Alexandre - 2018.pdf
ffi.load: libs/liblept.so.5
ffi.load: libs/libk2pdfopt.so.2
09/05/19-13:49:37 INFO  Loading plugins from directory: plugins
09/05/19-13:49:37 INFO  Plugin  plugins/autofrontlight.koplugin/main.lua  has been disabled.
[...]

PR #5282, opened 2019-08-29
[feat] Add MuPDF EPUB/FB2 dynamic font size

Frenzie:
Closes #4368.

<del>I might leave it like this for now (after some minor cleanup), meaning the full document will have to be reloaded, because `layoutDocument()` has to be called again. Ideally I'd give it that bit of polish too.</del> <ins>Rerendered on the fly.</ins>

PR #5337, opened 2019-09-07
[fix] PdfDocument: Hash collision

Frenzie:
A typo introduced in <https://github.com/koreader/koreader/pull/5282>.

This might resolve <https://github.com/koreader/koreader/issues/5323>, fixes <https://github.com/koreader/koreader/issues/5327>.


Eduardomb22:
Thanks for the fix.

PR #5344, opened 2019-09-08
[plugin/statistics] Synchronize Current book with footer

robert00s:
Better synchronize statistics `Current Book` with footer.
Also show `Current page/Total pages` instead of `Read pages/Total pages`.
Add `Read pages` row.

Before:

After:


plugins/statistics.koplugin/main.lua
```diff
@@ -1016,11 +1016,11 @@ function ReaderStatistics:getCurrentStat(id_book)
         { _("Total time"), util.secondsToClock(total_time_book, false) },
         { _("Total highlights"), tonumber(highlights) },
         { _("Total notes"), tonumber(notes) },
+        { _("Total read pages"), tonumber(total_read_pages) },
```
Frenzie: Total pages read ought to sound better.
```suggestion
        { _("Total pages read"), tonumber(total_read_pages) },
```

NiLuJe: Which made me realize that `Current pages` sounds iffy, too. `Current pages read`?

NiLuJe: I'd also possibly swap "period" for "session", and repeat it where appropriate (i.e., "Pages read this session", "Current session duration", ...).

NiLuJe: "Time spent reading today", "Pages read today"

robert00s:
Thanks. Done :)

Frenzie:
(Merging because I need to test if the OTA file is generated correctly and this is a small one.)

PR #5345, opened 2019-09-08
Disable show chapter markers when progress bar is hidden in footer

robert00s:


Biep:
The progress bar no longer has the option to shrink to the available space.  On my Kobo Aura One, in portrait mode, when showing everything (in default order), memory usage drops off the bar at the far end.  If the bar would shrink ever so little, it would all fit, as it used to.

NiLuJe:
@Biep: Could you open a new, dedicated issue for thazt?

(With a xref top here, if that's indeed where the regression came from).

poire-z:
Possibly #5757 enforcing a 20% min width.

Biep:
@NiLuJe: done (#5778).

Issue #5342, opened 2019-09-08
[UX] NumberPicker value loss


Frenzie:
* KOReader version: v2019.09.1 (but likely since introduction)

#### Issue
When you tap on the number it's not prefilled with the value, making it needlessly hard to amend (e.g., add `.5` to the value).

Similarly, if you enter an out of bounds value it's just discarded completely instead of showing a warning and adjusting it up/down.

#### Steps to reproduce
Tap on the number to bring up an input dialog.

PR #5353, opened 2019-09-09
Messages in numberpicker widget

robert00s:
See: #5342 

Close: #5342

### DIFF
--- a/frontend/ui/widget/numberpickerwidget.lua
+++ b/frontend/ui/widget/numberpickerwidget.lua
@@ -163,7 +163,7 @@ function NumberPickerWidget:paintWidget()
                                     })
                                 elseif input_value and input_value > self.value_max then
                                     UIManager:show(InfoMessage:new{
-                                        text = T(_("This value should be %1 or less"), self.value_max),
+                                        text = T(_("This value should be %1 or less."), self.value_max),
                                         timeout = 2,
                                     })
                                 else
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/numberpickerwidget.lua
```diff
@@ -128,39 +129,61 @@ function NumberPickerWidget:paintWidget()
...
+                                local input_value = tonumber(input_dialog:getInputText())
                                 if input_value and input_value >= self.value_min and input_value <= self.value_max then
                                     self.value = input_value
                                     self:update()
+                                    UIManager:close(input_dialog)
+                                elseif input_value and input_value < self.value_min then
+                                    self.value = self.value_min
+                                    self:update()
+                                    UIManager:show(InfoMessage:new{
+                                        text = _("Invalid value. The minimum value has been set."),
```
Frenzie: Or maybe something like "This value should be self.value_min or less"?

frontend/ui/widget/numberpickerwidget.lua
```diff
@@ -128,39 +130,55 @@ function NumberPickerWidget:paintWidget()
...
                                     self:update()
+                                    UIManager:close(input_dialog)
+                                elseif input_value and input_value < self.value_min then
+                                    UIManager:show(InfoMessage:new{
+                                        text = T(_("This value should be %1 or more"), self.value_min),
+                                        timeout = 2,
+                                    })
+                                elseif input_value and input_value > self.value_max then
+                                    UIManager:show(InfoMessage:new{
+                                        text = T(_("This value should be %1 or less"), self.value_max),
```
Frenzie:
```suggestion
                                        text = T(_("This value should be %1 or less."), self.value_max),
```

Issue #5356, opened 2019-09-10
Unable to exit KOReader onKobo Glo


calverb:
* KOReader version:
2019.09.02
* Device:
Kobo Glo
#### Issue
I have two Kobo Glos one with 2019.09 and one with 2019.09.02.   Both have the same problem. These were updated using OTA from the earlier 2019.08.

When I try to exit KOReader, I get a "Bomb" Screen and KoReader restarts.   This even happens if I try to exit using the Power Off option.   The only way I can shut down KOReader is by using a Hard reset (paperclip reset) on the device,  The Power Slider on the Glo takes me to the power off screen and then restarts KOReader

##### `crash.log` (if applicable)

[crash.log](https://github.com/koreader/koreader/files/3594662/crash.log)

Frenzie:
```
./luajit: frontend/device/kobo/nickel_conf.lua:162: Wrong front light state value type (expect boolean)!
stack traceback:
	[C]: in function 'assert'
	frontend/device/kobo/nickel_conf.lua:162: in function 'set'
	frontend/device/kobo/powerd.lua:187: in function 'saveSettings'
	frontend/device/kobo/device.lua:725: in function 'saveSettings'
	frontend/ui/uimanager.lua:63: in function 'action'
	frontend/ui/uimanager.lua:746: in function '_checkTasks'
	frontend/ui/uimanager.lua:1029: in function 'handleInput'
	frontend/ui/uimanager.lua:1115: in function 'run'
	./reader.lua:260: in main chunk
	[C]: at 0x000140c1
```

Some functions in that file seem to be using a plain assert instead of a dbg:guard. Perhaps those parts predate dbg:guard.

https://github.com/koreader/koreader/blob/6af1bc46b1e43fff06c61015ba46443a643e959d/frontend/device/kobo/nickel_conf.lua#L161-L169

https://github.com/koreader/koreader/blob/6af1bc46b1e43fff06c61015ba46443a643e959d/frontend/device/kobo/nickel_conf.lua#L183-L187

However, it doesn't seem like this would be related to 2019.09.

What's your Kobo firmware version?

calverb:
I thought I had replied using Email; but perhaps not

Kobo is running 4.15.12920 (27/5/2019).
No problem before I updated to KoReader 2019..09

Apologies for any duplication
---------------

NiLuJe:
How are you launching KOReader, and if you use KSM, how did you install the FW update?

What are your frontlight settings in Nickel?

[...]

PR #5361, opened 2019-09-11
[fix, Kobo] Don't crash trying to sync frontlight with Nickel.

Frenzie:
When reasonably possible, the program should only crash in debug mode.

Adds a couple of extra unit tests to prevent regressions and adds docs.

Fixes <https://github.com/koreader/koreader/issues/5356>.

### DIFF
--- a/frontend/device/kobo/nickel_conf.lua
+++ b/frontend/device/kobo/nickel_conf.lua
@@ -1,6 +1,6 @@
---[[
-    Access and modify values in 'Kobo eReader.conf' used by Nickel.
-    Only PowerOptions:FrontLightLevel is currently supported .
+--[[--
+Access and modify values in `Kobo eReader.conf` used by Nickel.
+Only PowerOptions:FrontLightLevel is currently supported.
 ]]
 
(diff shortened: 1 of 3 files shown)


frontend/device/kobo/nickel_conf.lua
```diff
@@ -150,46 +181,76 @@ function NickelConf._write_kobo_conf(re_Match, key, value, dont_create)
...
 
+--[[--
+Set frontlight level.
+
+@int new_intensity
+--]]
 function NickelConf.frontLightLevel.set(new_intensity)
-    assert(new_intensity >= 0 and new_intensity <= 100,
-           "Wrong brightness value given!")
+    if type(new_intensity) ~= "number" or (not (new_intensity >= 0) and (not new_intensity <= 100)) then return end
```
Frenzie: I should've probably split this up into two or three commits. :man_facepalming: Anyway, this line here and the matching line below + moving the assert to dbg:guard are the meat of this thing, plus that one line in powerd. The docs make the diff look a lot more imposing than it is.

PR #5362, opened 2019-09-11
Support the Kobo Libra

NiLuJe:
Basically assume it's a tinier Forma. Hopefully that'll hold up ;).

Depends on https://github.com/koreader/koreader-base/pull/974


reader.lua
```diff
@@ -108,6 +108,7 @@ while argidx <= #ARGV do
     if arg == "-h" then
         return showusage()
     elseif arg == "-d" then
+        -- @note: this is too late for the framebuffer module, since it's initialized via device, which was required above us.
```
Frenzie: Pssht the special LDoc tag is signaled by `---` (or more). ;-)

reader.lua
```diff
@@ -52,6 +52,12 @@ else
     end
 end
 
+-- should check DEBUG option in arg and turn on DEBUG before loading other
+-- modules, otherwise DEBUG in some modules may not be printed.
+local dbg = require("dbg")
```
Frenzie: Can't this entire (new) block just move below argument checking?

reader.lua
```diff
@@ -52,6 +52,12 @@ else
...
+if G_reader_settings:isTrue("debug") then dbg:turnOn() end
+if G_reader_settings:isTrue("debug") and G_reader_settings:isTrue("debug_verbose") then dbg:setVerbose(true) end
+
 local Device = require("device")
```
Frenzie: Or more to the point this part.

Frenzie: Like so:
```diff
diff --git a/reader.lua b/reader.lua
index 86dca3ab..1cae83a3 100755
--- a/reader.lua
+++ b/reader.lua
@@ -52,15 +52,6 @@ else
     end
 end
 
-local Device = require("device")
-local dpi_override = G_reader_settings:readSetting("screen_dpi")
-if dpi_override ~= nil then
-    Device:setScreenDPI(dpi_override)
-end
-
-local CanvasContext = require("document/canvascontext")
-CanvasContext:init(Device)
-
 -- option parsing:
 local longopts = {
     debug = "d",
@@ -121,6 +112,15 @@ while argidx <= #ARGV do
     end
 end
 
+local Device = require("device")
+local dpi_override = G_reader_settings:readSetting("screen_dpi")
+if dpi_override ~= nil then
+    Device:setScreenDPI(dpi_override)
+end
+
+local CanvasContext = require("document/canvascontext")
+CanvasContext:init(Device)
+
 local ConfirmBox = require("ui/widget/confirmbox")
 local QuickStart = require("ui/quickstart")
 local UIManager = require("ui/uimanager")
```

NiLuJe: Nothing blew up ;).

reader.lua
```diff
@@ -52,14 +52,11 @@ else
...
-local dpi_override = G_reader_settings:readSetting("screen_dpi")
-if dpi_override ~= nil then
-    Device:setScreenDPI(dpi_override)
-end
-
-local CanvasContext = require("document/canvascontext")
-CanvasContext:init(Device)
+-- should check DEBUG option in arg and turn on DEBUG before loading other
+-- modules, otherwise DEBUG in some modules may not be printed.
+local dbg = require("dbg")
```
Frenzie: Unless you think it's clearer this way (I don't really think it matters) I'd probably move this part back to where it was to minimize the diff. But whichever way you prefer.

NiLuJe: I do kinda like it there, that leaves the whole "argument parsing" stuff in a single block without that in between ;).

Frenzie: In a way it's part of the argument parsing but I see what you're saying. So yeah, cool. ^_^

Issue #3033, opened 2017-07-27
Poweroff script addition


embryo10:
* KOReader version: 2015.11-1054
* Device: Kobo Glo

Last year I made an [addition to the KSM7 poweroff.sh](https://www.mobileread.com/forums/showthread.php?t=274069) to log the uptime of the reader. 
Since Koreader doesn't use this powerof.sh, is there a way to add that modification to its shutdown script?
The extra commands are:
```bash
uptimelog="/adds/kbmenu/log/uptime.txt"
last=$(awk 'END {print $4}' $uptimelog)
bat="$(cat /sys/devices/platform/pmic_battery.1/power_supply/mc13892_bat/capacity)"
echo "$(date +%Y:%m:%d_%H:%M), $(awk -v last="$last" -v bat="$bat" '{printf("%d:%02d:%02d:%02d total: %d bat: %d",($1/60/60/24),($1/60/60%24),($1/60%60),($1%60),($1+last),(bat))}' /proc/uptime)" >> $uptimelog
```
An alternative would be to somehow use the KSM script from inside Koreader for shutdown.
Is that possible?

Frenzie:
Sure, send a PR? :-)

On Jul 27, 2017 21:08, "embryo10" <notifications@github.com> wrote:


   - KOReader version: 2015.11-1054
   - Device: Kobo Glo

Last year I made an addition to the KSM7 poweroff.sh
<https://www.mobileread.com/forums/showthread.php?t=274069> to log the
uptime of the reader.
Since Koreader doesn't use this powerof.sh, is there a way to add that
modification to its shutdown script?
The extra commands are:

uptimelog="/adds/kbmenu/log/uptime.txt"
last=$(awk 'END {print $4}' $uptimelog)
bat="$(cat /sys/devices/platform/pmic_battery.1/power_supply/mc13892_bat/capacity)"echo
"$(date +%Y:%m:%d_%H:%M), $(awk -v last="$last" -v bat="$bat"
'{printf("%d:%02d:%02d:%02d total: %d bat:
%d",($1/60/60/24),($1/60/60%24),($1/60%60),($1%60),($1+last),(bat))}'
/proc/uptime)" >> $uptimelog

An alternative would be to somehow use the KSM script from inside Koreader
for shutdown.
Is that possible?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/koreader/koreader/issues/3033>, or mute the thread

[...]

PR #5369, opened 2019-09-12
[feat] Allow setting any document cover as screensaver

Frenzie:
Cf. suggestion by @KenMaltby in <https://github.com/koreader/koreader/issues/3033#issuecomment-526657352>.


Frenzie:
I'll merge this now to collect some feedback.

PR #5371, opened 2019-09-13
[UX] Frontlight widget slider

Frenzie:
This responds to pan events.


Frenzie:
This works okay, but I wrote a different concept that I like better.

Issue #5312, opened 2019-09-04
Fine tune and set default in UI for everything


pblasi:
Hi,
We can fine tune T/B margins and font size in UI, but we can't fine tune L/R margins, Line Spacing, Contrast... Please add fine tuning to them (pop-up numerical keyboard to set desired value would be an option).  Then, having option to set this fine tuned value as default would be quite necessary (not only for the squared presets...)
Thanks!

Frenzie:
People are working on it:
https://github.com/koreader/koreader/pull/5303 https://github.com/koreader/koreader/pull/5306

NiLuJe:
Random UX thoughts for interested parties:

Add a vertical ellipsis button on the right of said widgets, which pops a SpinWidget up when tapped, and sets a FineTuneDefault on hold.
Possibly tweak the SpinWidget itself if it doesn't support popping up the keyboard (and make it honor safe clamping values).

Otherwise, the ButtonProgress fine tuning support may not work as-is for everything (the +/- value is fixed, which was fine for the margins, but may not be for stuff with a wider window of value (contrast/line spacing).

pblasi:
To make it more simple, join features and avoid adding more visual stuff, my thought would be: Single tap on widget name, (since long tap is for info) to pop a window containing: 
-  actual value which is editable
- numerical keyboard
- "ok" button
- "set default" button

PR #5375, opened 2019-09-13
Add fine tune in line spacing

robert00s:
See: https://github.com/koreader/koreader/issues/5312#issuecomment-528506421
Progress widget has a new extra button `⋮` that triggers SpinWidget.

### DIFF
--- a/frontend/ui/widget/spinwidget.lua
+++ b/frontend/ui/widget/spinwidget.lua
@@ -33,6 +33,9 @@ local SpinWidget = InputContainer:new{
     value_hold_step = 4,
     ok_text = _("OK"),
     cancel_text = _("Cancel"),
+    -- extra button on bottom
+    custom_text = nil,
+    callback_custom = nil,
     -- set this to see extra default button
     default_value = nil,
     default_text = _("Use default"),
@@ -140,6 +143,20 @@ function SpinWidget:update()
                 end,
             },
         })
+    end
+        if self.custom_text then
+            table.insert(buttons,{
+                {
+                    text = self.custom_text,
+                    callback = function()
+                        if self.callback_custom then
+                            self.value = value_widget:getValue()
+                            self.callback_custom(self)
+                        end
+                        self:onClose()
+                    end,
+                },
+            })
     end
 
     local ok_cancel_buttons = ButtonTable:new{
--- a/frontend/ui/widget/spinwidget.lua
+++ b/frontend/ui/widget/spinwidget.lua
@@ -34,8 +34,8 @@ local SpinWidget = InputContainer:new{
     ok_text = _("OK"),
     cancel_text = _("Cancel"),
     -- extra button on bottom
-    custom_text = nil,
-    callback_custom = nil,
+    extra_text = nil,
+    extra_callback = nil,
     -- set this to see extra default button
     default_value = nil,
     default_text = _("Use default"),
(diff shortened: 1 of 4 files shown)


frontend/ui/widget/spinwidget.lua
```diff
@@ -33,6 +33,9 @@ local SpinWidget = InputContainer:new{
     value_hold_step = 4,
     ok_text = _("OK"),
     cancel_text = _("Cancel"),
+    -- extra button on bottom
+    custom_text = nil,
+    callback_custom = nil,
```
poire-z: nitpick, but either `custom_` as a prefix for both, or as a suffix for both :) it looks strange as they are there.
(and may be `extra` instead of `custom` as you name that button "extra button" - unless this kind of added stuff is already called custom elsewhere.)

poire-z:
(Dunno if that's possible for `Zoom (dpi)` which is quite crowded (it could be useful in some rare cases to be able to set some intermediate value) - while still keeping the buttons with the values shown (I'd rather not like it to become a progressbutton :)

poire-z:
Might be useful with this fine tune to be able to exceed `value_max = values[#values],` (and value_min by symetry). May be with adding to the option something like `fine_tune_value_max = 250` and `fine_tune_value_min = 30` ? If not provided, use `values[#values]` as it is now.
(I'm thinking about arabic text that looks like it may sometimes need more than 130% of the default font interline space.)

frontend/ui/widget/spinwidget.lua
Frenzie: Prefix is probably more in line with the rest of the code around the program, good catch.

robert00s:
> (Dunno if that's possible for Zoom (dpi) which is quite crowded (it could be useful in some rare cases to be able to set some intermediate value) - while still keeping the buttons with the values shown (I'd rather not like it to become a progressbutton :)

Should be doable :) Adds to my todo list.

robert00s:
@poire-z 
What about something like that:
obraz

One problem I see. How to present current intermediate value?

poire-z:
> One problem I see. How to present current intermediate value?

Just as we do for the others: highlight/select the nearest one ? If I set 150 or 175, highlight 167. (Or the one above, or the one below, doesn't really matter.)

> What about something like that

Well, that's the idea. But it's a little bit off aesthetically.
Dunno if the `...` button should have the style of the widget on its left (height, rounded corners or not)- or if they should all have the same style (so they match vertically).
In any case, they should have the same width.
I guess it's probably not easy to fix, just mentionning it :) - if it's too much trouble, better drop the idea (if I have to choose between this little used feature, or staying pretty - I choose pretty :)

Also, just noticing that (on the emulator and on your screenshot), the `...` on the Line Spacing is a bit off, it looks like there are too much padding on its right, so it's not aligned the same on both sides vs the above widget edges. (nitpick, but as I struggled hard to have them aligned, it's worth having a look :)

(Also, what's that greyish background? Not a screenshot but a photograph ? :)

robert00s:
Thanks :)
version 1
obraz

version 2
obraz

version 3
obraz

Sorry for bad screenshots quality. It's VirtualBox problem with driver :)

NiLuJe:
I kinda like 3 personally for this kind of widget ;).

poire-z:
I'm fine with 1 and 3.
I kinda like 1, but it wastes a bit of width room. Unlike 3 which gives more room to the other buttons .
So, may be 3 if you manage the fix the little overflow on the right (vs the View mode and Render mode button toggles).

(But 2 could be ok if you manage to have the `...` correctly aligned with the same width :)
(But 3 is probably the easiest to fix and maintain.)

Frenzie:
I like 3 best. 2 is fine too, but like @poire-z said it would require more coordination.

Issue #5374, opened 2019-09-13
FR: Add rumble support


Frenzie:
More of a brainstorm than an FR. Linkdump without commentary incoming.

Docs:
https://wiki.libsdl.org/SDL_HapticOpenFromJoystick
https://developer.android.com/reference/android/os/Vibrator

https://davidgow.net/handmadepenguin/ch6.html

Background:
https://medium.com/@sdrzn/make-your-ios-app-feel-better-a-comprehensive-guide-over-taptic-engine-and-haptic-feedback-724dec425f10

pazos:
@Frenzie: I do like the idea but after a quick search I think the implementation on android could have some problems:

1. differences between api levels
2. overhead of the JNI calls (about 1ms)
3. Is the call blocking the program until the vibration stops? In that case vibrate on keyboard taps will become impossible.

Also I would like to finish https://github.com/koreader/android-luajit-launcher/pull/183 first and these days I'm very busy trying to find time for both my daily job and my family.

Frenzie:
> Is the call blocking the program until the vibration stops?

Certainly not in SDL.

> Also I would like to finish koreader/android-luajit-launcher#183 first and these days I'm very busy trying to find time for both my daily job and my family.

I'm just asking for some feedback on the device abstraction logic. ;-) Job & family take precedence of course!

pazos:
@Frenzie: I did a first implementation in https://github.com/koreader/android-luajit-launcher/pull/190.

Needs a lot of testing and tweaking because default values are far from perfect (at least in 8.0+)

I'm currently testing with this changes on the frontend:

```diff
diff --git a/frontend/device/android/device.lua b/frontend/device/android/device.lua
index 4deee2c1..b194da09 100644
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -70,6 +70,7 @@ local Device = Generic:new{
     canRestart = no,
     firmware_rev = android.app.activity.sdkVersion,
     display_dpi = android.lib.AConfiguration_getDensity(android.app.config),
+    isHapticFeedbackEnabled = function() return android.canVibrate() end,
     hasClipboard = yes,
     hasOTAUpdates = canUpdateApk,
     canOpenLink = yes,
@@ -225,6 +226,15 @@ function Device:initNetworkManager(NetworkMgr)
     end
 end
 
+function Device:performHapticFeedback(type)
+    local short, long = 80, 160 -- milliseconds
[...]

PR #5380, opened 2019-09-15
[feat] Add haptic feedback

Frenzie:
References <https://github.com/koreader/koreader/issues/5374>.

<hr>

@pazos This reflects my basic thought process on how to implement this. Most actions wouldn't do anything of course, basically just keyboard & hold. In using [performHapticFeedback](https://developer.android.com/reference/android/view/View.html#performHapticFeedback(int)) it would be up to the system whether it wants to do anything with it.

### DIFF
--- a/frontend/apps/reader/modules/readerstatus.lua
+++ b/frontend/apps/reader/modules/readerstatus.lua
@@ -1,5 +1,6 @@
 local BookStatusWidget = require("ui/widget/bookstatuswidget")
 local ButtonDialogTitle = require("ui/widget/buttondialogtitle")
+local Device = require("device")
 local InfoMessage = require("ui/widget/infomessage")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local UIManager = require("ui/uimanager")
@@ -38,6 +39,7 @@ function ReaderStatus:addToMainMenu(menu_items)
 end
 
 function ReaderStatus:onEndOfBook()
+    Device:performHapticFeedback("CONTEXT_CLICK")
     local settings = G_reader_settings:readSetting("end_document_action")
     local choose_action
     local collate = true
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -215,10 +215,20 @@ function Device:init()
         setClipboardText = function(text)
             return input.setClipboardText(text)
         end,
+        gameControllerRumble = function(left_intensity, right_intensity, duration)
+            return input.gameControllerRumble(left_intensity, right_intensity, duration)
+        end,
         file_chooser = input.file_chooser,
     }
 
     self.keyboard_layout = require("device/sdl/keyboard_layout")
+
+    if self.input.gameControllerRumble(0, 0, 0) then
+        self.isHapticFeedbackEnabled = yes
+        self.performHapticFeedback = function(type)
+            self.input.gameControllerRumble()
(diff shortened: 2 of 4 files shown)


frontend/apps/reader/modules/readerstatus.lua
```diff
@@ -1,9 +1,11 @@
 local BookStatusWidget = require("ui/widget/bookstatuswidget")
 local ButtonDialogTitle = require("ui/widget/buttondialogtitle")
+local Device = require("device")
 local InfoMessage = require("ui/widget/infomessage")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local UIManager = require("ui/uimanager")
 local _ = require("gettext")
+local Input = Device.input
```
Frenzie:
```suggestion
```

frontend/device/sdl/device.lua
```diff
@@ -215,11 +215,20 @@ function Device:init()
         setClipboardText = function(text)
             return input.setClipboardText(text)
         end,
+        gameControllerRumble = function(left_intensity, right_intensity, duration)
+            return input.gameControllerRumble(left_intensity, right_intensity, duration)
+        end,
         file_chooser = input.file_chooser,
     }
```
Frenzie:
```suggestion
    }

```

frontend/device/sdl/device.lua
```diff
@@ -215,11 +215,21 @@ function Device:init()
         setClipboardText = function(text)
             return input.setClipboardText(text)
         end,
+        gameControllerRumble = function(left_intensity, right_intensity, duration)
+            return input.gameControllerRumble(left_intensity, right_intensity, duration)
+        end,
         file_chooser = input.file_chooser,
-    }
+     }
```
Frenzie:
```suggestion
    }
```

frontend/apps/reader/modules/readerstatus.lua
```diff
@@ -38,6 +39,9 @@ function ReaderStatus:addToMainMenu(menu_items)
 end
 
 function ReaderStatus:onEndOfBook()
+    if Device:isHapticFeedbackEnabled() then
```
Frenzie: I think it's probably more elegant if this just reads `Device:performHapticFeedback("CONTEXT_CLICK")`.

NiLuJe: You mean fold the cap check inside performHaptic*?

Frenzie: Exactly. It already defaults as a no-op anyway and this way it just adds a lot of extra boilerplate.

frontend/device/sdl/device.lua
```diff
@@ -215,11 +215,21 @@ function Device:init()
...
         end,
+        gameControllerRumble = function(left_intensity, right_intensity, duration)
+            return input.gameControllerRumble(left_intensity, right_intensity, duration)
+        end,
         file_chooser = input.file_chooser,
     }
 
     self.keyboard_layout = require("device/sdl/keyboard_layout")
 
+    if self.input.gameControllerRumble(0, 0, 0) then
```
Frenzie: Note to self: somewhere it should convert -1 into false. Probably best to stick that directly in FFI.

Frenzie:
Merging now. The API can be adjusted if other platforms require it.

Issue #4230, opened 2018-09-17
Feature request: Specify screenshot output folder


onaka-itai:
* KOReader version: koreader-beta-20180729
* Device: Kobo Aura HD

After using koreader for almost a year, I recently discovered that you can take screenshots. The default output folder (.adds/koreader/screenshots) is inaccessible unless you enable hidden folders (and know that it exists there). I was wondering if we could have an option to set a default screenshot location, similar to what the screensaver function has.

PR #5382, opened 2019-09-15
Set screenshot output folder

robert00s:
Close: #4230 

I'm not sure that `Screenshot directory` is located in right place in menu (Settings -> Screen)

### DIFF
--- a/frontend/apps/reader/modules/readermenu.lua
+++ b/frontend/apps/reader/modules/readermenu.lua
@@ -178,6 +178,13 @@ function ReaderMenu:setUpdateItemTable()
         table.remove(self.menu_items.screensaver.sub_item_table, 9)
         table.insert(self.menu_items.screensaver.sub_item_table, ss_book_settings)
     end
+    self.menu_items.screenshot = {
+        text = _("Screenshot directory"),
+        callback = function()
+            local Screenshoter = require("ui/widget/screenshoter")
+            Screenshoter:chooseFolder()
+        end,
+    }
 
     local PluginLoader = require("pluginloader")
     self.menu_items.plugin_management = {
--- a/frontend/apps/reader/modules/readermenu.lua
+++ b/frontend/apps/reader/modules/readermenu.lua
@@ -178,13 +178,6 @@ function ReaderMenu:setUpdateItemTable()
         table.remove(self.menu_items.screensaver.sub_item_table, 9)
         table.insert(self.menu_items.screensaver.sub_item_table, ss_book_settings)
     end
-    self.menu_items.screenshot = {
-        text = _("Screenshot directory"),
-        callback = function()
-            local Screenshoter = require("ui/widget/screenshoter")
-            Screenshoter:chooseFolder()
-        end,
-    }
 
     local PluginLoader = require("pluginloader")
     self.menu_items.plugin_management = {
--- a/frontend/ui/elements/filemanager_menu_order.lua
+++ b/frontend/ui/elements/filemanager_menu_order.lua
@@ -67,6 +67,7 @@ local order = {
     },
     screen = {
         "screensaver",
+        "screenshot",
         "----------------------------",
         "screen_dpi",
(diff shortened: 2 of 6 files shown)


NiLuJe:
IIRC, we have a dedicated screensaver setting submenu ;).

frontend/apps/reader/modules/readermenu.lua
```diff
@@ -178,6 +178,13 @@ function ReaderMenu:setUpdateItemTable()
         table.remove(self.menu_items.screensaver.sub_item_table, 9)
         table.insert(self.menu_items.screensaver.sub_item_table, ss_book_settings)
```
Frenzie: Shouldn't it go over here? :-)

Frenzie: Oh sorry, misunderstood what it was.

frontend/ui/widget/screenshoter.lua
```diff
@@ -54,6 +56,41 @@ function Screenshoter:onScreenshot(filename)
...
+    table.insert(buttons, {
+        {
+            text = _("Choose screenshot directory"),
+            callback = function()
+                UIManager:close(self.choose_dialog)
+                require("ui/downloadmgr"):new{
+                    onConfirm = function(path)
+                        G_reader_settings:saveSetting("screenshot_dir", path .. "/")
+                        UIManager:show(InfoMessage:new{
+                            text = T(_("Screenshoot directory set to:\n%1"), path),
```
Frenzie:
```suggestion
                            text = T(_("Screenshot directory set to:\n%1"), path),
```

poire-z:
Should the little identical stuff you put in readermenu and filemanagermenu be put with all the other common menu stuff in frontend/ui/elements/common_settings_menu_table.lua (or another included file)?

frontend/ui/widget/screenshoter.lua
```diff
@@ -54,6 +56,41 @@ function Screenshoter:onScreenshot(filename)
...
+        {
+            text = _("Close"),
+            callback = function()
+                UIManager:close(self.choose_dialog)
+            end,
+        }
+    })
+    local screenshot_dir = G_reader_settings:readSetting("screenshot_dir") or DataStorage:getDataDir() .. "/screenshots/"
+    self.choose_dialog = ButtonDialogTitle:new{
+        title = T(_("Current screenshot directory:\n %1"), screenshot_dir),
```
Frenzie: Minor final nitpick, but is that space intentional?

frontend/ui/elements/filemanager_menu_order.lua
```diff
@@ -67,6 +67,7 @@ local order = {
     },
     screen = {
         "screensaver",
+        "screenshot",
```
Frenzie: I think the Device menu might make more sense than the Screen menu, but I'm not really sure.

poire-z: No strong opinion on that.
But we might visit the `Screen>` menu more often than the `Device>` one, so, as something of little regular use, I'd put it in `Device>` with the other misc things.

frontend/ui/widget/screenshoter.lua
robert00s: This is parametr of `ButtonDialogTitle` so `title=` should be indent.

frontend/ui/elements/filemanager_menu_order.lua
robert00s: I also thought about put it in `Device>`.

frontend/ui/widget/screenshoter.lua
poire-z: I think he means the one between `\n %1` (which adds some small indent on the 2nd line with the directory path) - but we have it in quite a few places, so either intentional, or it replicated thru cut & paste :)

robert00s: :)

Issue #5329, opened 2019-09-06
Custom sorting of status bar elements


pblasi:
Kind of weird arbitrary sorting by default when "show all at once" is on (battery and clock mixed between reading status elements).
Maybe putting the elements in the order you select them would be easy to implement?
Thanks!

LucyOne:
I would like that, too.

robert00s:
How should look widget?

poire-z:
> How should look widget?

If you're having a go at that, may be it doesn't need any specific UI.
As it seems Holding on the menu items has yet no effect (and no help_text), one simple option would be to just Hold on one item to make it the first. A few holds and you could reorder your items once you get the trick.
For the visual feedback, either just add a number after the text item - or a little more complicated: re-order the menu items following that order (so, Hold would refresh the menu with the item you just hold on at the top).

Or, as suggested, just do that on tap when an item is selected (so uncheck then check => that item is at top).

Frenzie:
@robert00s I would say an ideal widget would allow for some form of drag & drop or selecting an element and moving it up/down with arrow buttons on the bottom/side. But in lieu of such a widget (which I imagine would take at the very least an hour to create) I fully concur with @poire-z's suggestion.

> Or, as suggested, just do that on tap when an item is selected (so uncheck then check => that item is at top).

Preferably not. :-) I do that to enable/disable items (as would anybody else I'm sure) and then I'd end up with some completely random order.

PR #5389, opened 2019-09-16
Sort footer elements

robert00s:
Close: #5329 

- new option for footer - `Sort items`
- new widget `SortWidget`

Screenshots:

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -221,8 +221,11 @@ function ReaderFooter:init()
         local mode_name
         for i = 0, #mode_tbl do
             mode_name = mode_tbl[i]
-            if not ((mode_name == "wifi_status" and not Device:isAndroid())
-                or (mode_name == "frontlight" and not Device:hasFrontlight())) then
+            if mode_name == "wifi_status" and not Device:isAndroid() then
+                do end -- luacheck: ignore 541
+            elseif mode_name == "frontlight" and not Device:hasFrontlight() then
+                do end -- luacheck: ignore 541
+            else
                 self.mode_index[self.mode_nb] = mode_name
                 self.mode_nb = self.mode_nb + 1
             end
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -62,13 +62,6 @@ local symbol_prefix = {
     }
 }
 
-local MODE_NB = 0
-local MODE_INDEX = {}
-for k,v in pairs(MODE) do
-    MODE_INDEX[v] = k
-    MODE_NB = MODE_NB + 1
-end
-
 -- functions that generates footer text for each mode
 local footerTextGeneratorMap = {
     empty = function() return "" end,
@@ -218,6 +211,30 @@ function ReaderFooter:init()
         item_prefix = "icons"
     }
 
+    if not self.settings.order then
+        self.mode_nb = 0
+        self.mode_index = {}
+        local mode_tbl = {}
+        for k,v in pairs(MODE) do
+            mode_tbl[v] = k
+        end
+        local mode_name
+        for i = 0, #mode_tbl do
+            mode_name = mode_tbl[i]
+            if not ((mode_name == "wifi_status" and not Device:isAndroid())
+                or (mode_name == "frontlight" and not Device:hasFrontlight())) then
+                self.mode_index[self.mode_nb] = mode_name
+                self.mode_nb = self.mode_nb + 1
+            end
+        end
+    else
+        self.mode_index = self.settings.order
+        self.mode_nb = #self.mode_index
+    end
+    self.mode_list = {}
+    for i = 0, #self.mode_index do
+        self.mode_list[self.mode_index[i]] = i
+    end
     if self.settings.disabled then
         -- footer featuren disabled completely, stop initialization now
         self:disableFooter()
@@ -227,7 +244,7 @@ function ReaderFooter:init()
     self.pageno = self.view.state.page
     self.has_no_mode = true
     self.reclaim_height = self.settings.reclaim_height or false
-    for _, m in ipairs(MODE_INDEX) do
+    for _, m in ipairs(self.mode_index) do
         if self.settings[m] then
             self.has_no_mode = false
             break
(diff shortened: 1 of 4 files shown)


Frenzie:
That looks pretty cool. :+1:

frontend/ui/widget/button.lua
```diff
@@ -123,7 +123,9 @@ end
 function Button:setText(text, width)
     self.text = text
     self.width = width
+    local dimem = self[1].dimen
     self:init()
+    self[1].dimen = dimem
```
poire-z: Is that with no risk?
Some button with small text, when given some larger text, would not increase its width then?
Like in frontend/ui/widget/menu.lua: `self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))` could become truncated/misaligned when you change from `page 1 of 33` to `page 24 of 33` ?
(not tested, just thinking.)
Might be worth making it an option if it's really needed.
Or, as there is already a width parameter, may be using it like that would work?
```diff
- self.page_cnt[1]:setText(curr .. "/" .. total)
+ self.page_cnt[1]:setText(curr .. "/" .. total, self.page_cnt[1].width) -- or self.page_cnt[1].dimen.w)
```
Or do what you do if you are provide a `width=true` (instead of a size), true meaning: keep the existing dimen.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -218,6 +211,30 @@ function ReaderFooter:init()
...
+        self.mode_index = {}
+        local mode_tbl = {}
+        for k,v in pairs(MODE) do
+            mode_tbl[v] = k
+        end
+        local mode_name
+        for i = 0, #mode_tbl do
+            mode_name = mode_tbl[i]
+            if not ((mode_name == "wifi_status" and not Device:isAndroid())
+                or (mode_name == "frontlight" and not Device:hasFrontlight())) then
```
poire-z: That was a bit hard to read :)
A small indentation might help (I like that kind of stuff, dunno if it's ok with our coding style:):
```lua
            if not (   (mode_name == "wifi_status" and not Device:isAndroid())
                    or (mode_name == "frontlight" and not Device:hasFrontlight()) ) then
```
Or clearer:
```lua
           if mode_name == "wifi_status" and not Device:isAndroid() then
           else if mode_name == "frontlight" and not Device:hasFrontlight() then
           else
             -- do your suff
           end
```

frontend/ui/widget/sortwidget.lua
```diff
@@ -0,0 +1,499 @@
...
+function SortWidget:goToPage(page)
+    self.show_page = page
+    self:_populateItems()
+end
+
+function SortWidget:swipeItem(diff)
+    local swipe_to = self.marked + diff
+    if swipe_to > 0 and swipe_to <= #self.item_table then
+        self.show_page = math.ceil(swipe_to/self.items_per_page)
+        self:swipeItems(self.marked, swipe_to)
```
poire-z: I was thinking you were supporting the swipe gesture when reading this :) and was about to suggest it's better be named a pan :) But no.
Also, got confused by `self:swipeItem()` calling `self:swipeItems()` recursively (I missed the additional **s**) :)

Better this one be named `SortWidget:moveItem(diff)` and the other one below: `SortWidget:moveItemTo(pos1, pos2)`, if they are actually just moving items :)

OR, if I get it better after a re-read :) the one below should be named `SortWidget:swapItems(pos1, pos2)` which I think is what you meant :)

frontend/ui/widget/sortwidget.lua
```diff
@@ -0,0 +1,499 @@
...
+    if idx_offset + self.items_per_page <= #self.item_table then
+        page_last = idx_offset + self.items_per_page
+    else
+        page_last = #self.item_table
+    end
+
+    for idx = idx_offset + 1, page_last do
+        table.insert(self.main_content, VerticalSpan:new{ width = self.item_margin })
+        local invert_status = false
+        if (idx) == self.marked then
```
poire-z: No parens needed around `(idx)`

frontend/apps/reader/modules/readerfooter.lua
Frenzie: I agree that the second suggestion is clearer, but mainly because it's easier to read the conditions separately than having a `not (something or other)`.

You'd gain the same effect from something like this without the (imo) distracting `do end` stuff.
```lua
            if      not (mode_name == "wifi_status" and not Device:isAndroid())
                and not (mode_name == "frontlight" and not Device:hasFrontlight()) then
```

As an aside, you can also avoid LuaCheck's ire by writing:
```lua
if something then do
    -- No-op.
end
```
Because then it's not empty. ;-)

frontend/ui/widget/button.lua
robert00s: I need it because without this line:

`self[1].dimen = dimem
`

I get error (crash) after tap on `▼`

```
./luajit: frontend/ui/widget/button.lua:228: attempt to index field 'dimen' (a nil value)
stack traceback:
	frontend/ui/widget/button.lua:228: in function 'action'
	frontend/ui/uimanager.lua:750: in function '_checkTasks'
	frontend/ui/uimanager.lua:1033: in function 'handleInput'
	[C]: at 0x563cba790964

```
self[1].dimen is null but I really don't know why :(
Any idea how to do it correctly?

poire-z: button.lua:228 is code related to flashing the button when tap: you tap, its inverted to black, and a scheduled action is set to put it back to white. But in the meantime, with setText, you have re-initing the button, so, when in the scheduled action some milliseconds later, `self[1] = self.frame` is another object, which might not yet have a dimen (or it's the old object that has already been partly destroyed).
So, it might then better be fixed in the scheduled callback: it's its job to check that what it has to invert back still exists.
So, may be in there:
https://github.com/koreader/koreader/blob/3d696d9d87c9a068667c956db38f97ddcc26e0c9/frontend/ui/widget/button.lua#L225-L228
add
```lua
 UIManager:tickAfterNext(function() 
+    if not self[1] or not self[1].invert or not self[1].dimen then
+        -- widget no more there (destroyed, re-init'ed by setText(), or not inverted: nothing to invert back
+        return
+    end
     self.callback()
     [...]
```

robert00s: Thanks.
It works perfectly but I have to move self.callback() on top.
`self.callback()` "breaks" `self[1].dimen`

```lua
            UIManager:tickAfterNext(function()              
                self.callback()
                if not self[1] or not self[1].invert or not self[1].dimen then
                    -- widget no more there (destroyed, re-init'ed by setText(), or not inverted: nothing to invert back
                    return
                end
                self[1].invert = false
                UIManager:widgetRepaint(self[1], self[1].dimen.x, self[1].dimen.y)
                [...]

```

frontend/ui/widget/sortwidget.lua
```diff
@@ -445,7 +445,7 @@ function SortWidget:_populateItems()
     end)
 end
 
-function SortWidget:swipeItems(pos1, pos2)
+function SortWidget:moveItemTo(pos1, pos2)
```
poire-z: (Stated that in some previous comment) I think this one, you wanted to name it `swapItems(pos1, pos2)` - and `swapItems` is better than my early suggestion of `moveItemTo`.

AlanSP1:
Looks fantastic, great work.

PR #5391, opened 2019-09-17
[chore] Remove unnecessary indent on new line

robert00s:


Frenzie:
As an aside, I hadn't even noticed the whitespace while translating, so my Dutch translations were already "correct" (if technically wrong for how it was previously).

PR #5393, opened 2019-09-17
fix: Truncate text in spinwidget

robert00s:
Before:

After:

### DIFF
--- a/frontend/ui/widget/spinwidget.lua
+++ b/frontend/ui/widget/spinwidget.lua
@@ -10,8 +10,9 @@ local Font = require("ui/font")
 local HorizontalGroup = require("ui/widget/horizontalgroup")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local LineWidget = require("ui/widget/linewidget")
-local OverlapGroup = require("ui/widget/overlapgroup")
 local NumberPickerWidget = require("ui/widget/numberpickerwidget")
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/spinwidget.lua
```diff
@@ -83,12 +84,23 @@ function SpinWidget:update()
         value_widget,
     }
 
+    local close_button = CloseButton:new{ window = self, padding_top = Size.margin.title, }
+    local btn_width = close_button:getSize().w + Size.padding.default * 2
+    local title_txt_width = RenderText:sizeUtf8Text(
+        0, self.width, self.title_face, self.title_text).x
+    local show_title_txt
+    if self.width < (title_txt_width + btn_width) then
+        show_title_txt = RenderText:truncateTextByWidth(
```
Frenzie: Is there any hold to show full text implementation in place?

robert00s: What do you mean exactly? :)

Frenzie: Like how on most things you can hold or something to show the full text.

robert00s: But we didn't do that on title bar earlier. I'm not sure is very useful.

Frenzie: Just wondering. :-)

PR #5400, opened 2019-09-18
Update el_keyboard.lua

noembryo:
Set the Greek characters as the default


Frenzie:
Discussion in https://github.com/koreader/koreader/pull/5318#issuecomment-532705016

noembryo:
Me thank you for all..

PR #5402, opened 2019-09-18
Update el_keyboard.lua

noembryo:
Moved the accented Greek characters to the "Accented" group.
I hope its finished now..


noembryo:
@Frenzie Trying to update the `el_keyboard.lua` again (using the same way as before, failed.
Any ideas to fix it?

noembryo:
OK, at last found it.
It was a character that was disappeared when pasting the code.
And in 2 places.. :o(

Issue #5399, opened 2019-09-18
Line separator in footer: too short after rotation


gerhaher:
![Reader_2019-Sep-18_163053 (1)](https://user-images.githubusercontent.com/32645430/65160572-43143e00-da36-11e9-8682-769ded6a332c.png)

PR #5403, opened 2019-09-18
Fix: Incorrect width of separator line

robert00s:
Fix: #5399

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -319,15 +319,15 @@ function ReaderFooter:updateFooterContainer()
     end
 
     if self.settings.bottom_horizontal_separator then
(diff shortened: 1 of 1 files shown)


poire-z:
Just noticed that the footer/progress bar get a bit messed up in width and may overflow screen, when you resize the emulator window.
I don't think it was messy in look before your changes - but it probably was with the info, which was bearable. That change in look is a bit discracting :)

(Just mentionning it in case you see a quick fix while it's fresh :) there are other issues with emulator resize, possibly more when the resize switches from portrait to landscape - but I never felt like digging into fixing them, as it may our existing code more complex.)

robert00s:
Ok. Added to my todo list :)

Frenzie:
Yeah, I noticed that too just yet.

Issue #5401, opened 2019-09-18
[FR] Enable/disable corner hold


noembryo:
My problem starts with my dislike for blank margins.
Because of using 0pt margin all around, it is impossible to select a word in a corner if I have assigned an action for when holding that corner.
Most of the time, I have the same problem if I try to select a word from the last line of the page (the skim dialog pops up).

The only solution I could think of, is to have a way to toggle these shortcuts.
It could be a menu entry or even better a gesture.

NiLuJe:
Is that on a current version? IIRC, everything's neuterable in the Gesture manager.

poire-z:
I think he wants a new gesture action: `Neuter/unneuter hold corners`.
So, when you need to highlight some words in the corners, you just draw a gesture (eg. a circle), and hop, you can hold on words. When done, draw it again and corners get their specificied gestury power back.

noembryo:
Exactly.
I can always disable the specific corner hold, select the word and re-select the desired action again (it might be one on the 7th page!), but that's a lot of clicks.. :o)

Another way (not so fast as a gesture) is to have long pressing the `Hold corner` menu item to make the toggle...

Frenzie:
Could be a bit confusing… perhaps more something for #5079? (Gesture/action to toggle profile x.)

noembryo:
A profile with `Hold corner` actions and another profile with out?
Does that mean that when you change a specific action you have to update the profile too?

Frenzie:
That depends on which direction #5079 goes in. The way it's currently written you'd just save a backup of the current action before setting the corners to pass through.

But "Disable something or other" could also just live next to "Disable double tap". Just brainstorming.

poire-z:
Given how easily Hold corners can conflict with other regular hold actions on the book content, I think `Toggle hold corners gestures` deserves to be a gesture action (let's place it at the bottom of the 7th page of our actions if it's that confusing :).
(I personally never bothered with the hold corners gestures just because of that - and I'll probably continue not using them no matter what :)

Frenzie:
Hold corners doesn't jive with me either. Occasionally it might get in the way but mainly it just feels a bit slow and awkward.

PR #5404, opened 2019-09-19
[feat, UX] Ignore hold corners

Frenzie:
Closes <https://github.com/koreader/koreader/issues/5401>.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -1369,8 +1369,7 @@ function ReaderGesture:gestureAction(action, ges)
         Device:getPowerDevice():toggleFrontlight()
         self:onShowFLOnOff()
     elseif action == "toggle_hold_corners" then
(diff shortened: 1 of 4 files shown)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1518,6 +1526,11 @@ function ReaderGesture:pageUpdate(page)
 
 end
 
+function ReaderGesture:onIgnoreHoldCorners(ignore_hold_corners)
+    self.ignore_hold_corners = ignore_hold_corners
```
poire-z: Might be that event/handler  could be named `onIgnoreHoldCornersToggled` and it would `G_reader_settings:isTrue("ignore_hold_corners")` itself, so sender/caller do not have to provide it as an argument to the event?

Frenzie: Makes sense, but I'd still prefer an optional argument to explicitly set it on or off.

Issue #1933, opened 2016-03-29
FR - two lines progress bar


AlanSP1:
- KOReader version: ALL
- Device: ALL

I already asked if we could get two lines for lower status bar, but in a way it is OK if it doesn't happen.

The reason I asked for this is that we have more and more text options to add in status bar, and being statistics freak (is there a word for it?), I like to have them all turned all at the same time. With having all of them turned on, graphic progress bar becomes shorter and shorter.

And I found progress bar very useful to easily see where in the book we are, including information about chapters. So I need to decide, should I use all textual statistics, or have much smaller (shorter) progress bar.

Anyway, either having lower progress bar in second line along whole screen length or using upper one would be just fine with me.

But, upper progress bar now isn't same as lower. It's just one very thin line, or a bit thicker, without information about chapters (or I'm not able to discern them).

So, please, could you make upper graphic progress bar similar to lower one? Coding wise, you already have all the code you need, you just need to implement it. It would be great help. Or, even more preferably would be adding second line in lower position just for progress bar.

KenMaltby:
Perhaps it would be less impact on the normal reading experience, if there were a pop-up info box of maybe four lines size. It might drop back down after a certain amount of time after the finger is lifted.

AlanSP1:
Well, if you ask me, I prefer to see that info all the time, without any special action from my side, like pointing on same screen position.

I'd like to see it all the time, if it is possible.

houqp:
Sounds like a simple task. Now only if someone can tell me how to enable the upper progress bar... When I read, I usually disable everything so I can focus on the text. I don't even know we have a upper progress bar support...

KenMaltby:
It's the bottom reader menu gear icon "Full" progress bar.

houqp:
Oh, i see. That's the one rendered by crengine, which is completely out of
our control. If we want to change its styling, we will need to change
crengine's code instead...

AlanSP1:
If it is a problem, maybe you could put lower progress bar in separate line from textual statistics info, so it is wide as screen. I would even prefer it this way.

AlanSP1:
I already opened similar issue, just to remind: https://github.com/koreader/koreader/issues/1724

[...]

PR #5405, opened 2019-09-19
Two lines footer

robert00s:
Close: #1933

New option that moves progress bar to separate line in footer.

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -651,9 +651,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                 end,
                 callback = function()
                     self.settings.bottom_horizontal_separator = not self.settings.bottom_horizontal_separator
-                    self:updateFooterContainer()
-                    self:resetLayout(true)
-                    self:updateFooter()
+                    self:refreshFooter()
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -930,6 +943,23 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                text = _("Progress bar on separate line"),
+                checked_func = function()
+                    return self.settings.progress_bar_separate_line
+                end,
+                enabled_func = function()
+                    return not self.settings.disable_progress_bar
+                end,
+                callback = function()
+                    self.settings.progress_bar_separate_line = not self.settings.progress_bar_separate_line
+                    self:updateFooterContainer()
```
Frenzie: Should these three lines be refactored into one update function of some sort?

NiLuJe:
Speaking of the FW 4.17 progress bar: it's an overlay, so it doesn't directly affect the page layout (especially given the still somewhat generous hardcoded free footer space).

(It's obvious on a cover page, because it doesn't sit right at the bottom edge of the screen, so you get to see a few lines of the image *below* the bar ;)).

frontend/apps/reader/modules/readerfooter.lua
robert00s: Done :)

AlanSP1:
This looks great, thanks, just like I imagined.

PR #5406, opened 2019-09-21
bump crengine, show progress meter on load/re-rendering

poire-z:
Includes https://github.com/koreader/crengine/pull/310
- Adds tinyNodeCollection::getStatistics()
- epubfmt.cpp: cleanup indentation and small optimisation
- Add more progress callback events
- Optimisation: avoid re-init for EPUBs with embedded fonts
- Fix: don't draw border when border-color: transparent
- CSS: ignore 'inherit' inside font-family

Show a small progress meter at top left of screen when crengine is loading, re-rendering, and saving cache file (after a 2 second delay to not be bothered on small books):

<kbd>image</kbd>

Also add a wrapper for the last remaining bit of code that was accessing _document directly (`:getPageOffsetX()`).

### DIFF
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -1,11 +1,13 @@
-local bit = require("bit")
 local Blitbuffer = require("ffi/blitbuffer")
 local ConfirmBox = require("ui/widget/confirmbox")
 local Device = require("device")
-local InputContainer = require("ui/widget/container/inputcontainer")
 local Event = require("ui/event")
+local InputContainer = require("ui/widget/container/inputcontainer")
+local ProgressWidget = require("ui/widget/progresswidget")
 local ReaderPanning = require("apps/reader/modules/readerpanning")
+local TimeVal = require("ui/timeval")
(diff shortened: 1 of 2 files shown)


poire-z:
Some warnings in unit tests output since this PR:
```
10/09/19-15:06:23 WARN  cre callback() error: frontend/apps/reader/modules/readerrolling.lua:1000: attempt to index global 'G_reader_settings' (a nil value)
stack traceback:
	frontend/apps/reader/modules/readerrolling.lua:1000: in function 'showEngineProgress'
	frontend/apps/reader/modules/readerrolling.lua:983: in function <frontend/apps/reader/modules/readerrolling.lua:963>
```
I guess this should quickly solve it:
```diff
function ReaderRolling:showEngineProgress(percent)
-    if G_reader_settings:isFalse("cre_show_progress") then
+    if G_reader_settings and G_reader_settings:isFalse("cre_show_progress") then
```
But I dunno if there's more than that, or if it's somehow possible/logical that in the unit tests, the Lua stuff may be non standard, and the `G_reader_settings` that was there when we set the callback, is not more there for the callback itself.

poire-z:
No change with these ^ changes

PR #5373, opened 2019-09-13
[UX] Use ProgressWidget in frontlight

Frenzie:
This makes it behave like a pretty decent slider.

Cf. <https://github.com/koreader/koreader/issues/5314>.

<hr>

This is not unlike #5371 but it works much better. I'm attaching some screenshots, but without trying it on a real device they're fairly meaningless.

PR #5407, opened 2019-09-21
[Kobo/Cervantes] Prefer using the ioctl over sysfs when setting the FL.

NiLuJe:
It's much lower latency (re #5373).

We can do so on NL devices with a mixer.

### DIFF
--- a/frontend/device/kobo/powerd.lua
+++ b/frontend/device/kobo/powerd.lua
@@ -126,14 +126,14 @@ function KoboPowerD:init()
                 end
             end
             -- Does this device's NaturalLight use a custom scale?
-            self.fl_warmth_min = self.device.frontlight_settings["nl_min"] or self.fl_warmth_min
-            self.fl_warmth_max = self.device.frontlight_settings["nl_max"] or self.fl_warmth_max
+            self.fl_warmth_min = self.device.frontlight_settings.nl_min or self.fl_warmth_min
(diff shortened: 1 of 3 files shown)


frontend/device/kobo/powerd.lua
```diff
@@ -128,6 +128,14 @@ function KoboPowerD:init()
             -- Does this device's NaturalLight use a custom scale?
             self.fl_warmth_min = self.device.frontlight_settings["nl_min"] or self.fl_warmth_min
             self.fl_warmth_max = self.device.frontlight_settings["nl_max"] or self.fl_warmth_max
+            -- If this device has a mixer, we can use the ioctl for brightness control, as it's much lower latency.
+            if self.device:hasNaturalLightMixer() then
+                local kobolight = require("ffi/kobolight")
+                local ok, light = pcall(kobolight.open)
+                if ok then
+                    self.device.frontlight_settings["frontlight_ioctl"] = light
```
Frenzie: Or just a dot?

pazos:
@NiLuJe: could you please add the same to cervantes?.

frontend/device/kobo/powerd.lua
NiLuJe: @Frenzie: I'm not sure I follow ;).

Frenzie: As in self.device.frontlight_settings.frontlight_ioctl is the same thing.

Issue #5411, opened 2019-09-22
DMINIBAR_ advanced settings are not applied


derungs:
* KOReader version: v2019.09.3
* Device: Android, emulator

#### Issue
Advanced settings DMINIBAR_CONTAINER_HEIGHT, DMINIBAR_FONT_SIZE, and DMINIBAR_HEIGHT have no impact on status bar.

#### Steps to reproduce
Go to Advanced settings menu, set DMINIBAR_FONT_SIZE to 40 and DMINIBAR_CONTAINER_HEIGHT to 140.
Open an ebook (epub format).
**Expected**: large white space at the bottom, large font for pages.
**Actual**: font size and container size are still 14 px.

##### `crash.log` (if applicable)
n/a

Frenzie:
Note that if you exit the emulator *as if* it were a device it'll save the settings just fine. The fact that Alt+F4 or the X doesn't necessarily save all settings is a known issue, although I can't find it right now because it probably has a deceptive issue title.

Your Android problem may be something similar, or an entirely unique Android issue.

Frenzie:
@derungs Could you clarify precisely what issue you have in the emulator (as well as Android) and how you're triggering it? As you can see in https://github.com/koreader/koreader/pull/5412#issuecomment-533910384 it seems to work just fine. Of course you need to restart first, as prompted.

derungs:
@Frenzie Took me a while to identify the broken behaviour, see #5418.

derungs:
If you don't want advanced settings immediately applied, I suggest to close this issue here.

Frenzie:
Closing then. If you're going to change one of the advanced settings to be immediately applied you should simultaneously migrate it away from advanced settings. ;-)

Issue #4225, opened 2018-09-13
calibre catalog prevents exit


bunglegrind:
* KOReader version: v.2015.11-1800-g0924b57_2018-07-29
* Device: Kobo Aura One with KSM 0.9

#### Steps to reproduce

Preliminary step: Import tagged ebooks via calibre (just a couple);

1. In file manager go to search menu->find a book in calibre catalog->browse tags
2. Select one of the tags you have imported in koreader and open one of the related books
3. Open the reader menu->main menu->exit->exit

It does not exit, but instead koreader closes the book and the browse tags panel appears again.
Then, if you close the browse tag panel (ie, press the X on the top right) koreader exits to KSM.

NiLuJe:
(Leaky widget that doesn't close properly?)

Frenzie:
@bunglegrind Could you perhaps add some more info to the issue about how to set up this calibre catalog business (in calibre as well as KOReader)?

bunglegrind:
OK. I'm currently using Calibre 64-bit on Win7 Ultimate. 

Since I'm using KSM and not nickel, I've disabled the KoboTouch driver (see https://www.mobileread.com/forums/showthread.php?t=279299 ) and I'm connecting manually the device by clicking "connect/share->connect to folder" and then selecting the KOReader folder.

In my calibre library I've organized my ebooks (every kind of ebooks) with tags and they are automatically imported in KOReader when I send to device the ebooks. However, I can also force tags import by clicking "update cached metadata on device" on the device menu. 

In KOReader there's no particular setting. Just follow the steps I've described in the first post. 

Let me know if you need more info. ;)

Frenzie:
Ah, connect to folder under device/share.

Following your instructions calibre seems to have written a .metadata.calibre file, but the plugin says /metadata.calibre not found. This seems to be at least partially related to certain presumably Kobo-specific assumptions:

https://github.com/koreader/koreader/blob/85ac59ae0074c321861d57ccc1c3e52518e1d60e/frontend/apps/filemanager/filemanagersearch.lua#L78-L83

Setting that to something relevant in persistent.defaults.lua does the trick.

But if I then follow your steps regarding browse tags, the program does seem to exit properly. Is it possible that the main problem was fixed by some exit-related fixes? I do see it pop up because it was behind the reader, mind you, it's just not preventing exit.

Either way, the widget should be closed when opening a document. I'll take care of that.

Frenzie:
[...]

PR #5412, opened 2019-09-22
always directly reference advanced status bar settings

derungs:
fix for #5411

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -236,20 +236,20 @@ function ReaderFooter:init()
 
     self.footer_text = TextWidget:new{
         text = '',
-        face = Font:getFace(self.text_font_face, self.text_font_size),
+        face = Font:getFace(self.text_font_face, DMINIBAR_FONT_SIZE),
     }
     -- all width related values will be initialized in self:resetLayout()
     self.text_width = 0
     self.progress_bar = ProgressWidget:new{
         width = nil,
-        height = self.bar_height,
+        height = Screen:scaleBySize(DMINIBAR_HEIGHT),
         percentage = self.progress_percentage,
         tick_width = DMINIBAR_TOC_MARKER_WIDTH,
         ticks = nil, -- ticks will be populated in self:updateFooterText
         last = nil, -- last will be initialized in self:updateFooterText
     }
     self.text_container = RightContainer:new{
-        dimen = Geom:new{ w = 0, h = self.height },
+        dimen = Geom:new{ w = 0, h = Screen:scaleBySize(DMINIBAR_CONTAINER_HEIGHT)},
         self.footer_text,
     }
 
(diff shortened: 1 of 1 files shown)


poire-z:
In your issue, you meant that there are not picked up instantly - but they work when you restart KOReader, right?

So, yes, I guess it's bothering that these updated global variables are not picked up instantly by the ReaderFooter instance - expecially when testing and trying to find the right value.
But I guess it's the same for all these advanced/global variables.
And as we tried to get rid of these global variables (keeping them to still allow some tuning for past users), it's a bit bothering too to see them reintroduced that much :)

derungs:
Actually, they weren't picked up after restart. After restart I'd see the old values again.
Other advanced settings, like DMINIBAR_TOC_MARKER_WIDTH, get picked up immediately.

derungs:
Are you suggesting to add these tweaks to the status bar settings? If yes, I'd be willing to add these to the settings. I would appreciate a hint as to which classes to start with, since I'm new to the project.

poire-z:
>  Other advanced settings, like DMINIBAR_TOC_MARKER_WIDTH, get picked up immediately.

Does it survive a restart?
Because testing on the emulator, nothing gets saved :| I always see the previous values.
Anyway, I guess what people do usually is manually add the ones they wish to change, by copying them from [defaults.lua](https://github.com/koreader/koreader/blob/master/defaults.lua) to a new file named `defaults.persistent.lua` - see https://github.com/koreader/koreader/wiki/Change-defaults

> Are you suggesting to add these tweaks to the status bar settings?

Well, I don't know. These might be "niche" settings. Could you post a screenshot of what kind of footer you're trying to achieve?
And there might be combinations and inter-relations to ensure, so they might be merged into 1 or 2 settings that would influence the sizing that used these 3 variables.
The idea is that if we make them into settings, they become stuff stored inside G_reader_settings (settings.reader.lua) and the global variables will be killed.

> If yes, I'd be willing to add these to the settings. I would appreciate a hint as to which classes to start with, since I'm new to the project.

Well, see some of the [latest commits by @robert00s](https://github.com/koreader/koreader/commits?author=robert00s) who added recently many new footer menu items/settings.

Frenzie:
As written this is a definite no-go I'm afraid.

> Actually, they weren't picked up after restart. After restart I'd see the old values again.

Then it sounds like the issue is that they aren't in fact saved. That should be unaffected by the changes in this PR.

derungs:
Mainly, I need the font bigger so I can read it, and the progress bar a bit higher, again for visibility. I agree that 1 setting for everything makes sense, especially since the container makes most sense at the same height as the font, and the progress bar can be tailored to look ok-ish in comparison, e.g. 2/3 or 1/2 of the font height.

poire-z:
> Then it sounds like the issue is that they aren't in fact saved

OK, found out how to have them saved (not intuitive, but I guess that's a hint these should not be used :)
Once you have edited some of them, you need to quit the settings list, and go back to the menu, and HOLD on the menu item "Advanced settings": then, a confirm box appears and only then defaults.persistent.lua is created.

Frenzie:
@poire-z Settings are saved on exit. See https://github.com/koreader/koreader/issues/5411#issuecomment-533909130

Why not immediately I'm not sure…

poire-z:
^ Oh, right :)

Frenzie:
@poire-z I can't reproduce the issue though; perhaps I'm actually thinking of something subtly different or I fixed it. It asks me before exiting.

Screenshot_2019-09-22_21-26-59

poire-z:
Yes, it was me who exited the emulator with the Esc/Backspace key, so not via the regular Exit menu item (which calls setDefault:saveSettings() indeed). Not really an issue then.

Frenzie:
Right, I meant I couldn't reproduce it by pressing Alt+F4/X. I can indeed reproduce by pressing Escape.

NiLuJe:
From what I recall, *every* **sane** exit/reboot codepath we have should be wrapped in that check, so, yeah, that would leave "unexpected" exits ;).

Frenzie:
@NiLuJe @poire-z I think I was actually more directly thinking of this issue rather than the Escape thing: https://github.com/koreader/koreader/issues/4225#issuecomment-472177043

PR #5413, opened 2019-09-22
Add fine tune in contrast

robert00s:
See: #5312


frontend/ui/data/optionsutil.lua
```diff
@@ -42,16 +42,21 @@ function optionsutil.showValues(configurable, option, prefix)
         current = arg_table[current]
     end
     if option.labels and option.values then
-        for i=1,#option.labels do
-            if default == option.values[i] then
-                default = option.labels[i]
-                break
+        if option.more_options_param and option.more_options_param.value_table then
+            current = option.more_options_param.value_table[current + 1]
+            default = option.more_options_param.value_table[default + 1]
```
poire-z: I trust you on the whole thing (= I'm not a friend of configidialog.lua :).
Just enlighten me: why are these `+ 1` needed here, and below (I guess there's a reason, but as tables start at 1 with Lua, we rarely see that).

robert00s: Exactly. Tables in C are counted from 0 in Lua from so I have to add +1.

poire-z: OK, but where is there some C ? :)
I see at some point when calling document:setGamaIndex(), it should start from 0.
But once you have added `value_table = { 0.3, 0.35, 0.4, 0.45...` , it's now a Lua table.
And this code is generic, so if you need that +1 here, it might be bad when you use it for some other pure Lua that does not need it.
(Still, I haven't read precisely this whole PR - so I may be wrong :)

Frenzie: I have to agree that doing something so C-specific in what looks like pure Lua seems a bit suspicious for what's supposed to be generic (Lua-only) frontend code. In that case it would probably be better for an abstraction further down to deal with +1/-1.

poire-z: Some possible solutions:
- another tables, `more_option_params.args_table = { 0, 1, 2, 3...}` where you would pick the real arg value to send to the setGammaIndex()
- a convert function `more_option_parms.value_to_arg_func = function(index, value) return index-1 end`

(I still don't know what is saved/stored/used/send ... just sharing some ideas :)

Frenzie: I think the function idea is more versatile, and similar to how stuff like sorting is implemented in the likes of `table.sort`.

robert00s: Thanks.
I choose `more_option_params.args_table` because it is more readable to me.

PR #5414, opened 2019-09-22
Add Hebrew keyboard

yparitcher:
english - hebrew keyboard first draft, without diacritics, will add them soon.
please review and let me know if i am doing it right, before i spend more time finishing it off.
also how can i change the menu title from en_HE ?
it looks like the hebrew variable name are failing circleci, how can i fix this?
thanks
Reader_2019-Sep-22_133823
Reader_2019-Sep-22_133846
Reader_2019-Sep-22_134321

### DIFF
--- a/frontend/ui/data/keyboardlayouts/he_keyboard.lua
+++ b/frontend/ui/data/keyboardlayouts/he_keyboard.lua
@@ -1,0 +1,152 @@
+local en_popup = require("ui/data/keyboardlayouts/keypopup/en_popup")
+local he_popup = require("ui/data/keyboardlayouts/keypopup/he_popup")
+local com = en_popup.com -- comma (,)
+local prd = en_popup.prd -- period (.)
+local _at = en_popup._at
+local _eq = en_popup._eq -- equals sign (=)
+local _A_ = en_popup._A_
+local _a_ = en_popup._a_
+local _B_ = en_popup._B_
+local _b_ = en_popup._b_
+local _C_ = en_popup._C_
+local _c_ = en_popup._c_
+local _D_ = en_popup._D_
+local _d_ = en_popup._d_
+local _E_ = en_popup._E_
+local _e_ = en_popup._e_
+local _F_ = en_popup._F_
+local _f_ = en_popup._f_
+local _G_ = en_popup._G_
+local _g_ = en_popup._g_
+local _H_ = en_popup._H_
+local _h_ = en_popup._h_
+local _I_ = en_popup._I_
+local _i_ = en_popup._i_
+local _J_ = en_popup._J_
+local _j_ = en_popup._j_
+local _K_ = en_popup._K_
+local _k_ = en_popup._k_
+local _L_ = en_popup._L_
+local _l_ = en_popup._l_
+local _M_ = en_popup._M_
+local _m_ = en_popup._m_
+local _N_ = en_popup._N_
+local _n_ = en_popup._n_
+local _O_ = en_popup._O_
+local _o_ = en_popup._o_
+local _P_ = en_popup._P_
+local _p_ = en_popup._p_
+local _Q_ = en_popup._Q_
+local _q_ = en_popup._q_
+local _R_ = en_popup._R_
+local _r_ = en_popup._r_
+local _S_ = en_popup._S_
+local _s_ = en_popup._s_
+local _T_ = en_popup._T_
+local _t_ = en_popup._t_
+local _U_ = en_popup._U_
+local _u_ = en_popup._u_
+local _V_ = en_popup._V_
+local _v_ = en_popup._v_
+local _W_ = en_popup._W_
+local _w_ = en_popup._w_
+local _X_ = en_popup._X_
+local _x_ = en_popup._x_
+local _Y_ = en_popup._Y_
+local _y_ = en_popup._y_
+local _Z_ = en_popup._Z_
+local _z_ = en_popup._z_
+local aleph = he_popup.aleph
+local beis = he_popup.beis
+local gimmel = he_popup.gimmel
+local daled = he_popup.daled
+local hey = he_popup.hey
+local vov = he_popup.vov
+local zayin = he_popup.zayin
+local tes = he_popup.tes
+local yud = he_popup.yud
+local chof = he_popup.chof
+local lamed = he_popup.lamed
+local mem = he_popup.mem
+local mem_sofis = he_popup.mem_sofis
+local nun = he_popup.nun
+local samech = he_popup.samech
+local ayin = he_popup.ayin
+local pey = he_popup.pey
+local pey_sofis = he_popup.pey_sofis
+local tzadik = he_popup.tzadik
+local kuf = he_popup.kuf
+local reish = he_popup.reish
+local shin = he_popup.shin
+local taf = he_popup.taf
+
+return {
+    shiftmode_keys = {["Shift"] = true},
+    symbolmode_keys = {["Sym"] = true, ["ABC"] = true},
+    utf8mode_keys = {["IM"] = true},
+    umlautmode_keys = {["Äéß"] = true},
+    keys = {
+        -- first row
+        {  --  1           2       3       4       5       6           7       8       9       10      11      12
+            { _Q_,        _q_,    "„",    "0",    "׳",    "״",        "?",    "!",    "Å",    "å",    "1",    "ª", },
+            { _W_,        _w_,    "!",    "1",    "֘֘֙֙ ",    kuf,        "(",    "1",    "Ä",    "ä",    "2",    "º", },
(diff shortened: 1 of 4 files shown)


poire-z:
Looks fine (but obviously, I can't say anything about the hebrew part - that's 99% :)
But how are you using it as the UI doesn't support RTL? Type a letter and use the cursor keys to move back?! :)

> how can i change the menu title from en_HE ?

I guess you also need to add it here:
https://github.com/koreader/koreader/blob/bcceb94603a6ac934a604f979f35f368e39c094a/frontend/ui/language.lua#L7-L12
But then, it becomes a full UI language, falling back to english I guess?

> it looks like the hebrew variable name are failing circleci, how can i fix this?

I guess Lua is fine with hebrew letters in variable name, as it works and you managed to make some screenshots. So probably a Luacheck issue, not supporting them.
But may be it's saner to keep variable names in pure ascii?

pinging @Frenzie

poire-z:
> how are you using it as the UI doesn't support RTL? Type a letter and use the cursor keys to move back?! :)

In case you want more fun, I guess you could use a trick introduced with the koread keyboard (ko_KR_keyboard.lua), which wraps the inputbox with (very complicated) code. I guess you could use that to wrap `addChar()` so it does add the char, and then, if it detects its a hebrew char, just do an additional `leftChar()` to move the cursor before the added char.

yparitcher:
> But how are you using it as the UI doesn't support RTL? Type a letter and use the cursor keys to move back?! :)

the only thing RTL or LTR is the display,
i just type normally it just displays LTR (backwards) for example:
Reader_2019-Sep-22_142501
Reader_2019-Sep-22_142508


> > how can i change the menu title from en_HE ?
> 
> I guess you also need to add it here:
> 
> https://github.com/koreader/koreader/blob/bcceb94603a6ac934a604f979f35f368e39c094a/frontend/ui/language.lua#L7-L12
> 
> But then, it becomes a full UI language, falling back to english I guess?
> 

it works, i just did not add it to the language menu, like other languages that dont have a translation yet

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -526,6 +526,7 @@ local VirtualKeyboard = FocusManager:new{
         ja = "ja_keyboard",
         pt_BR = "pt_keyboard",
         ko_KR = "ko_KR_keyboard",
+        en_HE = "en_HE_keyboard",
```
Frenzie: Please keep it alphabetical.

frontend/ui/data/keyboardlayouts/en_HE_keyboard.lua
```diff
@@ -0,0 +1,152 @@
...
+local _v_ = en_popup._v_
+local _W_ = en_popup._W_
+local _w_ = en_popup._w_
+local _X_ = en_popup._X_
+local _x_ = en_popup._x_
+local _Y_ = en_popup._Y_
+local _y_ = en_popup._y_
+local _Z_ = en_popup._Z_
+local _z_ = en_popup._z_
+local _א_ = he_popup._א_
```
Frenzie: Is this variable name actually working? luacheck doesn't usually object without reason (see the failing CI at the bottom of the main PR page):

```
Luacheck results
Checking frontend/ui/data/keyboardlayouts/en_HE_keyboard.lua 1 error

    frontend/ui/data/keyboardlayouts/en_HE_keyboard.lua:59:8: expected statement near 'א'

Checking frontend/ui/data/keyboardlayouts/keypopup/he_popup.lua 1 error

    frontend/ui/data/keyboardlayouts/keypopup/he_popup.lua:2:6: expected '}' (to close '{' on line 1) near 'א'
```

yparitcher: the variables work in the emulator, see the screenshots, i can rename all of the in ascii if needed it just will be more work.

yparitcher:
> > also how can i change the menu title from en_HE ?
> 
> Note that `en-HE` would mean Hebrew English, a variety of English like en-US, en-UK, en-AU, en-NZ, etc. I should simply be named `he`.

right now the keyboard is just a extension to the english, replacing cyrillic characters with hebrew ones, so i am not sure if it should be `he` or `en-HE` i don't mind changing it either way.
this is the way i primarily use it: english ui, but also able to type hebrew characters

poire-z:
> i just type normally it just displays LTR (backwards) for example

Oh right :) everything works in the logical order, so you must type them that way (until we really support displaying them inverted - but still keeping them in that logical order for store and search.)

frontend/ui/data/keyboardlayouts/en_HE_keyboard.lua
Frenzie: Indeed it seems that it works in LuaJIT, but it doesn't work in plain Lua. We'd better not introduce any needless incompatibility with regular Lua. :-)

Frenzie:
> right now the keyboard is just a extension to the english, replacing cyrillic characters with hebrew ones, so i am not sure if it should be he or en-HE i don't mind changing it either way.

Please keep it as `he` because `en-HE` doesn't make much sense without a special explanation.

poire-z:
> Please keep it as he because en-HE 

But `he` should probably have the hebrew letters on the initial keyboard map (and latin ones on another map) - if it ever gets usable if/when we have RTL support.
This one here is mainly `en` with another map dedicated to hebrew.
We don't have (yet) anything to handle that kind of 2nd level language.
Unless if now that we have the option to choose the keyboard layout via the menu, it becomes less of an issue, and if I ever need to enter japanese, korean or hebrew, I'd just select the keyboard I want. It's just complicated when I want to enter these different letters in a same typing session... as it's not always possible to save what I just typed, go back to menu, select another keyboad, and re-edit what I was typing.

Frenzie:
I want something like holding or swiping on the spacebar to change layouts, probably out of a selection of enabled ones. But like I wrote in the other PR I've been letting it slide because that would take at least a medium time investment, not least of which experimenting with UX/UI, whereas the quick 10-minute menu switch implementation does a lot of good in little time and I should've done it a year ago.

> But he should probably have the hebrew letters on the initial keyboard map (and latin ones on another map) - if it ever gets usable if/when we have RTL support.

Which doesn't change the fact that en-HE is a confusing name. `en-he` might kinda sorta work, especially if you make it something like `en__he`. I'm not entirely sure how clear that would be, but at least it's not plainly saying English of the HE variety.

Frenzie:
@yparitcher Btw, bit of a random question but how did you mark this PR a draft? I don't think I've ever noticed that before.

yparitcher:
should be fixed, `he` for now, has all the letters their variations (via popups), nikkud (punctution), and some (not all) cantillation marks.

yparitcher:
> @yparitcher Btw, bit of a random question but how did you mark this PR a draft? I don't think I've ever noticed that before.

when opening the PR on the submit button there is a down arrow to select a draft PR, for not ready PR's

frontend/ui/data/keyboardlayouts/he_keyboard.lua
```diff
@@ -0,0 +1,152 @@
...
+local daled = he_popup.daled
+local hey = he_popup.hey
+local vov = he_popup.vov
+local zayin = he_popup.zayin
+local tes = he_popup.tes
+local yud = he_popup.yud
+local chof = he_popup.chof
+local lamed = he_popup.lamed
+local mem = he_popup.mem
+local mem_sofis = he_popup.mem_sofis
```
Frenzie: Pity we can't keep 'em three characters but clarity over ASCII art. :-)

yparitcher:
i might add the rest of the cantillation marks via popups in a seperate PR, this one is ready for now

poire-z:
> any comments @poire-z ?

Nope, looks fine to me.
No pb having `he` as a global UI language? It won't magically popup on transiflex and have more people disappointed as we don't display what they added nicely?

Frenzie:
@poire-z Remember when you added a hundred languages for Google Translate or some such? All complete translations have contained the word Hebrew for ages. ;-)

Frenzie:
But also the strings in question here aren't translated. Which is pretty useful when testing some stuff in Chinese. ;-)

poire-z:
I didn't mean the strings in this PR.
Just that now, Hebrew is a UI language. So, people may select it, and find ... english.
Then, they'd go to Transiflex translating all our english strings to hebrew, and the next day go download a nightly, where they'll see `detalsnart yeht sgnirts eht lla` like that ! :)

Frenzie:
I decoupled the language names from the menu items when I added the keyboard selection. My goal was merely to access them through `Language:getLanguageName()` but it also happily obviates any such concerns. :-)

poire-z:
Oh, right, the available menu languages are in another section below in language.lua! :+1:

PR #5416, opened 2019-09-23
Use the proper LED driver on the Libra

NiLuJe:
Doesn't seem to actually help, though?
Which doesn't make much sense to me, as the kernel hasn't changed on
that front...


NiLuJe:
c.f., https://www.mobileread.com/forums/showpost.php?p=3893596&postcount=159

Issue #5418, opened 2019-09-23
Advanced settings are not saved on exit from opened book


derungs:
* KOReader version: 2019.09.3-30-gbcceb946_2019-09-22
* Device: emulator, tolino (android)

#### Issue
Modified advanced settings are not stored when you exit koreader from the reader, that is, from an open book. When you exit from the file manager, the save dialog pops up and the advanced settings are stored.

#### Steps to reproduce
  * Change advanced settings.
  * Open a book.
  * Exit koreader.
  * Start koreader.

**Observed**: The advanced setting has the old value.
**Expected**: The advanced setting has the new value.

Frenzie:
Makes sense (or rather, that's what the code does). The entire workflow for saving advanced settings is pretty weird imho.

derungs:
I could move the "save/don't save" dialog to `filemanagersetdefault`, and pass the "real" close function as a callback rather than the "save/don't save" dialog function.
That would to some degree unweird the save workflow.

Frenzie:
Sounds good to me. :+1:

derungs:
On it, the idea seems to work (once I fixed my rookie mistake with lambdas).

NiLuJe:
Here's me eating humble pie for saying "every sane exit workflow is wrapped in tat check" just yesterda :D.

Hmm, pie! :cake:

derungs:
Two questions:
* ReaderMenu seems to use lfs without explicitly setting it. Do I miss something or should I just require lfs at the top of the file?
* What does OTAManager do? It calls exitOrRestart, but I haven't discovered a way to trigger it.

Frenzie:
> ReaderMenu seems to use lfs without explicitly setting it. Do I miss something or should I just require lfs at the top of the file?

It's not a pure Lua module but a library, so it doesn't have any functional consequences. If you think it should be enforced as a code style then you can remove the exception from `.luacheckrc`. The current situation is probably primarily a case of harmless legacy and less a case of purposeful consideration.

> What does OTAManager do? It calls exitOrRestart, but I haven't discovered a way to trigger it.

OTA updates are only for embedded devices. On Android it merely starts the download of a newer package if available.

PR #5419, opened 2019-09-23
[fix] add save dialog to more exit paths (#5418)

derungs:
The defaults are saved when leaving the "set defaults" dialog/menu, not when exiting koreader as before.
Moved the Don't save / Save dialog from FileManagerMenu to FileManagerSetDefaults.


frontend/apps/filemanager/filemanagersetdefaults.lua
```diff
@@ -356,4 +356,28 @@ function SetDefaults:saveSettings()
...
 
+function SetDefaults:saveBeforeExit(callback)
+    if self.settings_changed then
+        UIManager:show(ConfirmBox:new{
+            text = _("You have unsaved default settings. Save them now?\nTap \"Cancel\" to return to KOReader."),
+            ok_text = _("Save"),
+            ok_callback = function()
+              self.settings_changed = false
+              self:saveSettings()
+              callback()
```
derungs: Does it make sense to check for callback ~= nil, so that the save dialog can be triggered without further action?
The "Don't save" action would become a "Don't save now", the changes would still be included in a next save.

Frenzie:
I think we had a bit of mutual misunderstanding. :-/ I wasn't calling the code weird, but the way the whole thing works.

frontend/apps/filemanager/filemanagersetdefaults.lua
Frenzie: When you said the real function as a callback I thought you simply meant something like `UIManager:close(menu_container)`.

derungs: Not sure I understand what you mean.
If you mean call `UIManager:close(<some_parameter>)` instead of `callback()`, I feel uncomfortable doing that because `ReaderMenu:exitOrRestart()` has quite a bit more logic in it than `FileManagerMenu:exitOrRestart()`, so I'd rather preserve the code.

Frenzie: And moreover, I wrote most of that logic for good reason! ;-)

frontend/apps/reader/modules/readermenu.lua
```diff
@@ -193,19 +195,19 @@ function ReaderMenu:setUpdateItemTable()
     self.menu_items.exit_menu = {
         text = _("Exit"),
         hold_callback = function()
-            self:exitOrRestart()
+            SetDefaults:saveBeforeExit(function() self:exitOrRestart() end)
```
Frenzie: I mean none of this. This is precisely what I meant by "porting the weirdness to the reader" in https://github.com/koreader/koreader/issues/5418#issuecomment-534225444

derungs:
After sleeping on it, I think I understand what you mean by weird workflow.
Since some/most of the advanced settings are only applied on restart, I suggest changing the user experience to:
On closing the Defaults dialog, open a dialog:
koreader-seetings
Discard changes: Read settings from defaults.lua to prevent side effects of immediately-applied settings like DMINIBAR_TOC_MARKER_WIDTH.
Save and restart: Writes defaults.lua and _immediately_ restarts the KOReader.

Frenzie:
Yup, that's pretty much what I was suggesting. ;-)

Btw, that button would need to vary its text, along the lines of `if Device:canRestart() then "Save and restart" or "Save and quit"`.

derungs:
There's still a bug with discarding default settings, plus cleanup needed.
Kinda updated the pull request by mistake. Yeah, I'm tired.

frontend/apps/filemanager/filemanagersetdefaults.lua
```diff
@@ -356,4 +355,37 @@ function SetDefaults:saveSettings()
...
+        saveMessage = "Save and quit"
+    end
+    if self.settings_changed then
+        UIManager:show(ConfirmBox:new{
+            text = _("KOReader needs to restart in order to apply all default settings."),
+            ok_text = _(saveMessage),
+            ok_callback = function()
+                self.settings_changed = false
+                self:saveSettings()
+                --self.ui:onClose()
```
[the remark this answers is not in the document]
derungs: missed that one, thanks

frontend/apps/filemanager/filemanagersetdefaults.lua
```diff
@@ -356,4 +355,37 @@ function SetDefaults:saveSettings()
     self.settings_changed = false
 end
 
+function SetDefaults:saveBeforeExit(callback)
+    local canRestart = Device:canRestart()
```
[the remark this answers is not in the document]
derungs: Assumed that this test might be expensive, so I didn't want to do it twice. Read the code now, agree with you.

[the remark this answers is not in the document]
Frenzie: Even the ones that are (relatively) expensive are by and large preloaded with `yes/no` or other relevant values within device init during the first second or so after first starting the program. :-) I think the only real exception is some of the wifi-related stuff.

derungs:
@Frenzie Thanks for the support.

Frenzie:
@derungs No problem. Thanks for fixing your own bug reports. ;-)

PR #5420, opened 2019-09-24
Better align toggleswitch in configdialog

robert00s:
See: https://github.com/koreader/koreader/pull/5375#issuecomment-533561559

Before (align `View mode` `Render mode` `Zoom`):
ver2

After (align `View mode` `Render mode` `Zoom`):
ver1

### DIFF
--- a/frontend/ui/widget/toggleswitch.lua
+++ b/frontend/ui/widget/toggleswitch.lua
@@ -73,7 +73,15 @@ function ToggleSwitch:init()
         w = item_width,
         h = item_height,
     }
+    local diff_one_step = item_width - (frame_inner_width / self.n_pos - 2*item_border_size)
+    local diff = 0
     for i = 1, #self.toggle do
+        diff = diff + diff_one_step
+        if diff >= 1 then
+            -- One pixel narrower to better align the entire widget
+            center_dimen.w = item_width - math.floor(diff)
+            diff = diff - math.floor(diff)
+        end
         local text = self.toggle[i]
         local face = Font:getFace(self.font_face, self.font_size)
         local txt_width = RenderText:sizeUtf8Text(0, Screen:getWidth(), face, text, true, true).x
--- a/frontend/ui/widget/toggleswitch.lua
+++ b/frontend/ui/widget/toggleswitch.lua
@@ -78,7 +78,7 @@ function ToggleSwitch:init()
         local real_item_width = item_width
         item_width_to_add = item_width_to_add + item_width_adjust
         if item_width_to_add >= 1 then
-            -- One pixel narrower to better align the entire widget
+            -- One pixel wider to better align the entire widget
             real_item_width = item_width + math.floor(item_width_to_add)
             item_width_to_add = item_width_to_add - math.floor(item_width_to_add)
         end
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/toggleswitch.lua
```diff
@@ -73,7 +73,15 @@ function ToggleSwitch:init()
         w = item_width,
         h = item_height,
     }
+    local diff_one_step = item_width - (frame_inner_width / self.n_pos - 2*item_border_size)
```
poire-z: Took me a bit of time to understand were that was coming from - I just had to look above :)

So, may be, put that nearer to its counterpart:
```diff
-local item_width = math.ceil(frame_inner_width / self.n_pos - 2*item_border_size)
+-- We'll need to adjust items width and distribute the accumulated fractional part to some
+-- of them for proper visual alignment
+local item_width_real = frame_inner_width / self.n_pos - 2*item_border_size
+local item_width = math.ceil(item_width_real)
+local item_width_adjust = item_width - item_width_real
```
and name your `diff` to something like `item_width_to_add`, for easier future reading.

robert00s: I like it. It's a lot more clear :)

frontend/ui/widget/toggleswitch.lua
```diff
@@ -65,15 +65,27 @@ function ToggleSwitch:init()
...
         w = item_width,
         h = item_height,
     }
+    --local diff_one_step = item_width - (frame_inner_width / self.n_pos - 2*item_border_size)
+    local item_width_to_add = 0
     for i = 1, #self.toggle do
+        item_width_to_add = item_width_to_add + item_width_adjust
+        if item_width_to_add >= 1 then
+            -- One pixel narrower to better align the entire widget
+            center_dimen.w = item_width - math.floor(item_width_to_add)
```
poire-z: Oups, well, we're now _substracting_ what we named `item_width_to_add` :)
So, I dunno, either name it `item_width_to_substract`, or use `item_width = math.floor(item_width_real)` (instead of `math.ceil()`) above, and a few adaptations here (a bit more thinking than the simple rename, but might be easier to understand with just additions and positive numbers :).

poire-z: Or I dunno. Your choice. Either decrease the width of one when needed. Or incresase the width of one when needed. Dunno if these 2 options are visually different, or if they are equivalent.

robert00s:
I have to move `center_dimen` inside for loop because widths of elements (single buttons) was not as expected.

frontend/ui/widget/toggleswitch.lua
```diff
@@ -65,27 +65,38 @@ function ToggleSwitch:init()
...
-        w = item_width,
-        h = item_height,
-    }
+    local item_height = self.height / self.row_count
+    local item_width_to_add = 0
     for i = 1, #self.toggle do
+        local real_item_width = item_width
+        item_width_to_add = item_width_to_add + item_width_adjust
+        if item_width_to_add >= 1 then
+            -- One pixel narrower to better align the entire widget
```
poire-z: Looks good to me.
Please just update that comment, narrower => wider (now that you changed the logic :)

PR #5421, opened 2019-09-24
[Kobo] Unbreak the FL ramp-up/down.

NiLuJe:
I got my wires crossed.

It's fast on Mk. 5, slow on Mk. 7.
So, slow it down on Mk. 5 so we actually get a ramp and not a jump.

### DIFF
--- a/frontend/device/kobo/powerd.lua
+++ b/frontend/device/kobo/powerd.lua
@@ -321,7 +321,8 @@ function KoboPowerD:turnOffFrontlightHW()
     util.runInSubProcess(function()
         for i = 1,5 do
             self:_setIntensity(math.floor(self.fl_intensity - ((self.fl_intensity / 5) * i)))
-            -- Newer devices appear to block slightly longer on FL ioctls/syfs, so only sleep on older devices...
+            --- @note: Newer devices appear to block slightly longer on FL ioctls/sysfs, so only sleep on older devices,
+            ---        otherwise we get a jump and not a ramp ;).
             if not self.device:hasNaturalLight() then
                 if (i < 5) then
                     util.usleep(35 * 1000)
@@ -359,7 +360,8 @@ function KoboPowerD:turnOnFrontlightHW()
     util.runInSubProcess(function()
         for i = 1,5 do
             self:_setIntensity(math.ceil(self.fl_min + ((self.fl_intensity / 5) * i)))
-            -- Newer devices appear to block slightly longer on FL ioctls/sysfs, so only sleep on older devices...
+            --- @note: Newer devices appear to block slightly longer on FL ioctls/sysfs, so only sleep on older devices,
+            ---        otherwise we get a jump and not a ramp ;).
             if not self.device:hasNaturalLight() then
                 if (i < 5) then
                     util.usleep(35 * 1000)
--- a/frontend/device/kobo/powerd.lua
+++ b/frontend/device/kobo/powerd.lua
@@ -321,8 +321,8 @@ function KoboPowerD:turnOffFrontlightHW()
     util.runInSubProcess(function()
         for i = 1,5 do
             self:_setIntensity(math.floor(self.fl_intensity - ((self.fl_intensity / 5) * i)))
-            -- NOTE: We generally don't need to sleep when using sysfs as a backend...
-            if self.device:hasNaturalLight() and not self.device:hasNaturalLightMixer() then
+            -- Newer devices appear to block slightly longer on FL ioctls/syfs, so only sleep on older devices...
+            if not self.device:hasNaturalLight() then
                 if (i < 5) then
                     util.usleep(35 * 1000)
                 end
@@ -359,7 +359,8 @@ function KoboPowerD:turnOnFrontlightHW()
     util.runInSubProcess(function()
         for i = 1,5 do
             self:_setIntensity(math.ceil(self.fl_min + ((self.fl_intensity / 5) * i)))
-            if self.device:hasNaturalLight() and not self.device:hasNaturalLightMixer() then
+            -- Newer devices appear to block slightly longer on FL ioctls/sysfs, so only sleep on older devices...
+            if not self.device:hasNaturalLight() then
                 if (i < 5) then
                     util.usleep(35 * 1000)
                 end


NiLuJe:
Regression since https://github.com/koreader/koreader/pull/5407

frontend/device/kobo/powerd.lua
```diff
@@ -321,8 +321,9 @@ function KoboPowerD:turnOffFrontlightHW()
     util.runInSubProcess(function()
         for i = 1,5 do
             self:_setIntensity(math.floor(self.fl_intensity - ((self.fl_intensity / 5) * i)))
-            -- NOTE: We generally don't need to sleep when using sysfs as a backend...
-            if self.device:hasNaturalLight() and not self.device:hasNaturalLightMixer() then
+            --- @note: Newer devices appear to block slightly longer on FL ioctls/sysfs, so only sleep on older devices,
+            ---        otherwise we get a jump and not a ramp ;).
+            if not self.device:hasNaturalLight() then
                 if (i < 5) then
                     util.usleep(35 * 1000)
```
Frenzie: Btw, how'd you come up with that value?

frontend/device/kobo/powerd.lua
```diff
@@ -321,8 +321,9 @@ function KoboPowerD:turnOffFrontlightHW()
     util.runInSubProcess(function()
         for i = 1,5 do
```
Frenzie: This explains why it's so slow. The effect kinda sorta but not really works if you toggle from a high value, but from/to a low value it just feels like a delay.

frontend/device/kobo/powerd.lua
NiLuJe: Trial & error, mostly. The goal was having the whole thing stay under a certain total duration (like, 500ms at worst), having enough steps to make it look decent everywhere (barring the ramp-down glitch on < Mk. 7), but not too much that the duration gets blown on slow devices.

Turns out, that's only a tiny bit faster than the delay of Mk. 7 ioctls ;p. Which is what I should have started with, as the ramp looks neat there ;).

frontend/device/kobo/powerd.lua
NiLuJe: There *is* a delay, but one we can do nothing about: turning the frontlight off is not immediate on < Mk. 7. So, essentially, the ramp-down works, but gets "stuck" on the final step.

No issue when turning it on, though. That's always smooth.

The issue doesn't exist on Mk. 7, but there, you do start to see that 5 steps is not enough with larger deltas..

Frenzie: > I live with the quirk because it's one users should be used to

Ah, but it feels as if it takes twice as long (even though it's probably only a third). ;-)

NiLuJe: I'm pretty used to it, so I don't really find it slow ;p. It should be taking around 150ms.

That jitter on the last step kinda ruins it, though. It usually gets hidden away by a flashing update, though.

NiLuJe: We could also increase the amount of steps on those devices, in order to ensure that the final step will mostly always be 1%, that somewhat limits the perceived effect.

NiLuJe: I remember swapping between 5 steps/35ms and 20 steps/10ms sleeps during testing, for instance.

But, of course, that didn't work all that well on Mk. 7, since they block for far more than 10ms.

But we could just make the constants device-specific.

NiLuJe: Or, err, duh, just enforce the penultimate step to 1.

NiLuJe: Meh, doesn't help all that much.

```C
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/ioctl.h>
#include <sys/stat.h>

#define CM_FRONT_LIGHT_SET 241

#define STEPS 20u
#define STEPSF 20.0f
#define SLEEP 7u

int main(int argc, char **argv)
{
        int ntx_io, retval, ptr;
        ntx_io = open("/dev/ntx_io", O_RDONLY | O_NONBLOCK | O_CLOEXEC);

        uint8_t fl_min = 0U;
        uint8_t fl_max = 100U;
        uint8_t fl_cur = 12U;
        // Ramp-down
        for (uint8_t i = 1U; i <= STEPS; i++) {
                // Fudge the penultimate step to limit the perceived effect the the extra delay needed to turn the light off
                if (i == STEPS - 1) {
                        ptr = 1;
                } else {
                        ptr = __builtin_ifloorf(fl_cur - ((fl_cur / STEPSF) * i));
                }
                retval = ioctl(ntx_io, CM_FRONT_LIGHT_SET, ptr);
                printf("ioctl(ntx_io, CM_FRONT_LIGHT_SET) set to %03d%%; returned %d\n", ptr, retval);
                if (i < STEPS) {
                        usleep(SLEEP * 1000);
                }
        }

        usleep(1750 * 1000);
        // Ramp-up
        for (uint8_t i = 1U; i <= STEPS; i++) {
                ptr = __builtin_iceilf(fl_min + ((fl_cur / STEPSF) * i));
                retval = ioctl(ntx_io, CM_FRONT_LIGHT_SET, ptr);
                printf("ioctl(ntx_io, CM_FRONT_LIGHT_SET) set to %03d%%; returned %d\n", ptr, retval);
                if (i < STEPS) {
                        usleep(SLEEP * 1000);
                }
        }

        close(ntx_io);
}
```

NiLuJe: And in the end, more steps don't actually help either there, so I kinda like the 5/35 compromise ;).

Frenzie: > We could also increase the amount of steps on those devices, in order to ensure that the final step will mostly always be 1%, that somewhat limits the perceived effect.

That's what happens when toggling from <10 anyway. The delay idea works best around 20-30. From higher the gaps are too large and from lower it's just an annoying additional delay. (For the record, values higher than 10 are useless. ;-P)

NiLuJe: On those devices, no delay makes the ramp non-operational (i.e., it visually jumps to the final value, barring said turn-off quirk), no matter the steps, and no matter the delta ;).

Frenzie: > no delay makes the ramp non-operational 

The desired situation. ;-)

But perhaps I was being a bit unclear. I meant brightness values higher than 10 being useless, not microsecond*1000 values.

NiLuJe: No, I got that ;p. You'll notice the "12" used in my testcase ;).

Issue #5354, opened 2019-09-09
Regression in right to left reading (inverted taps)


Frenzie:
Most likely introduced in <https://github.com/koreader/koreader/pull/4878>. A fix would probably be similar to <https://github.com/koreader/koreader/pull/5189>.

> Now after the update the problem seems to be solved but you can not bookmark the book after activating the option (invert page turn.....) until you restart the program 
thanks a lot

_Originally posted by @hgrain86 in https://github.com/koreader/koreader/issues/5346#issuecomment-529633769_

Frenzie:
Fixed by #5423.

PR #5423, opened 2019-09-25
[UX] add inverse reading order to gesture manager

yparitcher:
before you merge, i am not very familiar with lua so a few points;

i was not sure what to put here just followed the flow.
```lua
 {"toggle_inverse_reading_order", not self.is_docless, true},
```

is there a way to abstract the code from `readerrolling` and `readerpaging` to avoid duplication?

---

when testing this i came across a very weird related bug that i can not figure out.

on startup of the reader i can change the page turn direction using any gesture.

once the page turn direction was changed in the current session (of crengine) i cannot change it via the corresponding corner gesture.

 i can't go from RTL -> LTR using the right corners,
nor from LTR -> RTL using the top left corner,
nor either way using the bottom left corner.
(all once the page turn direction was changed in the current session)

however other gestures (short diagonal swipe) always work.

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -240,14 +240,6 @@ function ReaderGesture:initGesture()
     G_reader_settings:saveSetting(self.ges_mode, gesture_manager)
 end
 
-function ReaderGesture:onReSetupGesture()
-    local gesture_manager = G_reader_settings:readSetting(self.ges_mode)
-    for gesture, action in pairs(gesture_manager) do
-        self:setupGesture(gesture, action)
-    end
-    return true
-end
-
 function ReaderGesture:genMultiswipeSubmenu()
     return {
         text = _("Multiswipe"),
@@ -1368,7 +1360,7 @@ function ReaderGesture:gestureAction(action, ges)
     elseif action == "toggle_bookmark" then
         self.ui:handleEvent(Event:new("ToggleBookmark"))
     elseif action == "toggle_inverse_reading_order" then
-        self.ui:handleEvent(Event:new("ToggleReadingOrder"))
+        self:onToggleReadingOrder()
     elseif action == "toggle_frontlight" then
         -- when using frontlight system settings
         if lightFrontlight() then
@@ -1572,6 +1564,22 @@ function ReaderGesture:onGSensorToggle()
     UIManager:show(Notification:new{
         text = new_text,
         timeout = 1.0,
+    })
+    return true
+end
+
+function ReaderGesture:onToggleReadingOrder()
+    local document_module = self.ui.document.info.has_pages and self.ui.paging or self.ui.rolling
+    document_module.inverse_reading_order = not document_module.inverse_reading_order
+    document_module:setupTouchZones()
+    -- Needed to reset the touch zone overrides
+    local gesture_manager = G_reader_settings:readSetting(self.ges_mode)
+    for gesture, action in pairs(gesture_manager) do
+        self:setupGesture(gesture, action)
+    end
+    UIManager:show(Notification:new{
+        text = self.inverse_reading_order and _("RTL page turning.") or _("LTR page turning."),
+        timeout = 2.5,
     })
     return true
 end
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -240,6 +240,13 @@ function ReaderGesture:initGesture()
     G_reader_settings:saveSetting(self.ges_mode, gesture_manager)
 end
 
+function ReaderGesture:onReSetupGesture()
+    local gesture_manager = G_reader_settings:readSetting(self.ges_mode)
+    for gesture, action in pairs(gesture_manager) do
+        self:setupGesture(gesture, action)
+    end
+end
+
 function ReaderGesture:genMultiswipeSubmenu()
     return {
         text = _("Multiswipe"),
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -1578,7 +1578,7 @@ function ReaderGesture:onToggleReadingOrder()
         self:setupGesture(gesture, action)
     end
     UIManager:show(Notification:new{
-        text = self.inverse_reading_order and _("RTL page turning.") or _("LTR page turning."),
+        text = document_module.inverse_reading_order and _("RTL page turning.") or _("LTR page turning."),
         timeout = 2.5,
     })
     return true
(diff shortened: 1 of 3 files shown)


NiLuJe:
Wasn't there a thing where changing the reading direction subtly broke bits of the gesture manager that came up recently?

frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -1019,4 +1019,13 @@ function ReaderPaging:onGotoPrevChapter()
     return true
 end
 
+function ReaderPaging:onToggleReadingOrder()
+    self.inverse_reading_order = not self.inverse_reading_order
+    self:setupTapTouchZones()
+    UIManager:show(Notification:new{
+        text = _((self.inverse_reading_order and "RTL" or "LTR") .. " page direction."),
```
Frenzie: Just use the whole string whenever possible. Anything else is bad for localization.

```suggestion
        text = self.inverse_reading_order and _("RTL page direction.") or _("LTR page direction."),
```

As an aside, as written in the PR it wouldn't work regardless, and that'd have to be something like `T(_("%1 page direction"), (self.inverse_reading_order and _("RTL") or _("LTR")))`. But don't do that. ;-)

Frenzie:
> when testing this i came across a very weird related bug that i can not figure out.

That's https://github.com/koreader/koreader/issues/5354.

Frenzie:
> is there a way to abstract the code from readerrolling and readerpaging to avoid duplication?

I'm not sure if it would properly fit into one of the existing modules. ReaderGesture itself might be the best fit for the moment.

Something like:

```lua
local document_module = self.ui.document.info.has_pages and self.ui.paging or self.ui.rolling
-- rest of the code would be the same
```

Slightly inelegant/hackish, but probably still better than the duplication.

yparitcher:
> > is there a way to abstract the code from readerrolling and readerpaging to avoid duplication?
> 
> I'm not sure if it would properly fit into one of the existing modules. ReaderGesture itself might be the best fit for the moment.
> 
> Something like:
> 
> ```lua
> local document_module = self.ui.document.info.has_pages and self.ui.paging or self.ui.rolling
> -- rest of the code would be the same
> ```
> 
> Slightly inelegant/hackish, but probably still better than the duplication.

that would not be enough because the code would still have to be be duplicated in the menu code.

frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -241,8 +242,7 @@ function ReaderPaging:addToMainMenu(menu_items)
         text = _("Invert page turn taps and swipes"),
         checked_func = function() return self.inverse_reading_order end,
         callback = function()
-            self.inverse_reading_order = not self.inverse_reading_order
-            self:setupTapTouchZones()
+            self:onToggleReadingOrder()
```
Frenzie: In reply to https://github.com/koreader/koreader/pull/5423#issuecomment-535194152

> that would not be enough because the code would still have to be be duplicated in the menu code.

This line here is the tiniest bit faster than `self.ui:handleEvent(Event:new("ToggleReadingOrder"))` (or something along those lines) because in the current scenario there's no need to bubble the event through various widgets.

frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -1019,4 +1019,13 @@ function ReaderPaging:onGotoPrevChapter()
     return true
 end
 
+function ReaderPaging:onToggleReadingOrder()
+    self.inverse_reading_order = not self.inverse_reading_order
+    self:setupTapTouchZones()
+    UIManager:show(Notification:new{
+        text = self.inverse_reading_order and _("RTL page turning.") or _("LTR page turning."),
+        timeout = 2.5,
+    })
```
Frenzie: Speaking of bubbling, this should return true to stop propagation.

frontend/apps/reader/modules/readerpaging.lua
yparitcher: i was trying to deduplicate the code, if the speed and not going back and forth between modules is better i will change it back, i am new to the code and not sure of conventions, if there are any.

frontend/apps/reader/modules/readerpaging.lua
yparitcher: will fix

yparitcher:
> > when testing this i came across a very weird related bug that i can not figure out.
> 
> That's #5354.

i figured out the problem: when changing `inverse_reading_order` the page turn zone are placed over the override zones.
the solution re `setupGesture` when changing `inverse_reading_order`

yparitcher:
i am not sure if i fixed it the right way.
please take a look and see if it can be done better as i don't know lua very well so i am ver much c/p from other code.

frontend/apps/reader/modules/readerpaging.lua
Frenzie: And I'm asking for it but not seeing any yet. :-P

```lua
function Reader**Gesture**:onToggleReadingOrder()
-- this is where the line I suggested earlier goes
    local document_module = self.ui.document.info.has_pages and self.ui.paging or self.ui.rolling
-- and most of the selfs should be replaced by the reference to the module
    self.inverse_reading_order = not self.inverse_reading_order
    self:setupTouchZones()
    -- Needed to reset the touch zone overrides
-- this one however can then become self:onReSetupGesture
    self.ui:handleEvent(Event:new("ReSetupGesture"))
        text = self.inverse_reading_order and _("RTL page turning.") or _("LTR page turning."),
        timeout = 2.5,
    })
    return true
end

File: frontend/apps/reader/modules/readergesture.lua [hunk 4]
```diff
@@ -239,6 +240,14 @@ function ReaderGesture:initGesture()
     G_reader_settings:saveSetting(self.ges_mode, gesture_manager)
 end
 
+function ReaderGesture:onReSetupGesture()
```
Frenzie: Not the biggest fan of ReInitGesture either, but at least it's internally consistent. ;-)
```suggestion
function ReaderGesture:onReInitGesture()
```

frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -241,8 +242,7 @@ function ReaderPaging:addToMainMenu(menu_items)
         text = _("Invert page turn taps and swipes"),
         checked_func = function() return self.inverse_reading_order end,
         callback = function()
-            self.inverse_reading_order = not self.inverse_reading_order
-            self:setupTapTouchZones()
+            ReaderGesture:onToggleReadingOrder()
```
yparitcher: should i call this directly like now or through the event system?

frontend/apps/reader/modules/readerpaging.lua
yparitcher: done

frontend/apps/reader/modules/readerpaging.lua
Frenzie: These days I definitely prefer the Event system in cases like this. I think it makes the code easier to use and refactor.

Frenzie:
Thanks!

Issue #3778, opened 2018-03-19
Progress bar with chapter markers inaccurate in landscape mode


gerhaher:
* KOReader version: v2015.11-1573-g0a4fb89_2018-03-15/ 
* Device: KA1

#### Issue
In landscape mode the progress bar isn't coordinated with the chapter markers. The bar seems to lag.

Not all epubs are affected (and no pdfs AFAIK).

I've added an example. First portrait mode and then in landscape. Same place in a book (a new chapter).
![reader_2018-mar-19_183716](https://user-images.githubusercontent.com/32645430/37613591-b19cdf50-2ba8-11e8-9d44-5f67be842f07.png)
![reader_2018-mar-19_183658](https://user-images.githubusercontent.com/32645430/37613598-b4db050c-2ba8-11e8-8bb0-dd52e0fdf672.png)

Frenzie:
Pinging @poire-z to ask if it might be related to #3740.

I haven't managed reproduce this in real-life conditions. I can trigger it by resizing the window but that's more likely to be indirectly related to #3740.

poire-z:
I dunno, it may or may not :| #3740 being pure frontend code, it should be easy to test without.
In the onChangeScreenMode(), there was and is a self:onUpdatePos() which triggers via a scheduleIn a UpdateToc event.
Does it gets ok when you go to next page ?

Frenzie:
Actually I can reproduce it in real-life(ish) conditions. All you need is a sufficiently large window. (You'd think Aura One size would be big enough to notice but I didn't see an obvious problem without pixel counting.)

```
./kodev run -w=2000 -h=2500
```

The progressbar can seemingly run both fast and slow. What's special about resizing the window (as opposed to switching portrait/landscape) is that the program as a whole becomes slightly confused about locations, including the TOC. The reason I thought it might potentially be related is because it looks a bit like sends you where the chapter marks end up but that was probably just an illusion or a fluke. In any case it's a different, otherwise irrelevant issue. ;-)

Here's the problem demonstrated by switching from landscape (good) to portrait (bad). What's good depends on what loads first.

<img src=https://user-images.githubusercontent.com/202757/37682855-fd4dfd98-2c8a-11e8-9dfa-74738501c7a3.png width=30%>
<img src=https://user-images.githubusercontent.com/202757/37682856-fd738400-2c8a-11e8-8826-892eafe42a31.png width=30%>

And portrait (good) to landscape (bad):

[...]

PR #5425, opened 2019-09-25
Fix: incorrect chapter markers

robert00s:
Close: #3778 

We have to set chapter markers in footer every time when screen geometry is changed.
Thanks @poire-z for: https://github.com/koreader/koreader/issues/3778#issuecomment-389967175

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -911,6 +911,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                 end,
                 callback = function()
                     self.settings.disable_progress_bar = not self.settings.disable_progress_bar
+                    self:setTocMarkers()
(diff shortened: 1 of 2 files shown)


poire-z:
Looks simple & good to me :)
Just a note, if we should trust the comment above the function you call:
https://github.com/koreader/koreader/blob/e7cfc657809472b8fe1f89c1dbbe7c605c3263ff/frontend/apps/reader/modules/readerfooter.lua#L986-L987
may be you need some additional check in this PR before calling it ?
Or may be that comment is obsolete, dunno :)

robert00s:
Thanks. 
I added an additional check inside `setTocMarkers()` and found an next bug by the way.
We also need to refresh pages count inside this method

```lua
    if reset then
        self.progress_bar.ticks = nil
       +self.pages = self.view.document:getPageCount()
    end

```

PR #5426, opened 2019-09-25
[wip]: android - timeout without writting system settings

pazos:
Requires https://github.com/koreader/android-luajit-launcher/pull/183

Avoid WRITE_SETTINGS permission, which is wonky and don't work on some devices. 

new restrictions:
- we cannot set a timeout less than system timeout. 
- menu entries below system timeout will be disabled.

to-do:
- reset wakelock timer ~~if no wakelock is held and a custom setting is currently being used.~~ each time a new input event is queued.

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -169,19 +169,16 @@ function Device:init()
         self.isTouchDevice = yes
     end
 
-    -- check if we enabled support for custom timeouts (including wakelocks)
+    -- check if the activity has a custom timeout
     if android.needsWakelocks() then
-        android.setScreenOffTimeout(-1)
+        android.setScreenOffTimeout(C.ATIMEOUT_WAKELOCK)
     else
         local timeout = G_reader_settings:readSetting("android_screen_timeout")
-        if timeout and timeout > 0 then
-            -- set a custom timeout if we already have write settings permission.
-            -- do not attempt to request permissions here.
-            if android.canWriteSettings() then
+        if timeout then
+            if timeout > C.ATIMEOUT_SYSTEM
+            or timeout == C.ATIMEOUT_WAKELOCK then
                 android.setScreenOffTimeout(timeout)
             end
-        elseif timeout and timeout == -1 then
-            android.setScreenOffTimeout(timeout)
         end
     end
 
--- a/frontend/ui/elements/screen_android.lua
+++ b/frontend/ui/elements/screen_android.lua
@@ -1,6 +1,7 @@
 local isAndroid, android = pcall(require, "android")
 local Device = require("device")
 local Geom = require("ui/geometry")
+local ffi = require("ffi")
 local logger = require("logger")
 local _ = require("gettext")
 local Input = Device.input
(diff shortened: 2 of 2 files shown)


pazos:
Need to find a way to reset the timer for the current wakelock duration each time the user touches the screen or uses a hardware key.

I currently don't know how to do that without inserting a new call to wakelock acquire on the event loop. Any ideas @poire-z @NiLuJe @robert00s @Frenzie

Frenzie:
> I currently don't know how to do that without inserting a new call to wakelock acquire on the event loop.

Seems fine to me? Unless you mean something other than http://koreader.rocks/doc/modules/ui.uimanager.html#UIManager:scheduleIn

NiLuJe:
I imagine autosuspend already does something of the sort?

Frenzie:
It schedules a check against the last input time.

https://github.com/koreader/koreader/blob/3de3ea04ffe64f73301e9cbec305fab406530638/plugins/autosuspend.koplugin/main.lua#L71-L86

https://github.com/koreader/koreader/blob/3de3ea04ffe64f73301e9cbec305fab406530638/plugins/autosuspend.koplugin/main.lua#L112-L115

poire-z:
> Need to find a way to reset the timer for the current wakelock duration each time the user touches the screen or uses a hardware key.

I haven't followed how you implemented that, but I thought, as it's some android stuff, Android would handle that itself, no? If you set a wakelock with a specific timeout, I would expect Android to reset it on touch event while the app has the focus.

If not, well, you're then just setting a countdown timer.
Either re-set it on each even (which might be costly on the Android system call - but could be cheap by just having the android part plug a method into UIManager, that UIManager would call on each event - no need to broadcast an event).
Or if you get some event/hook called when the timer expire, just have it check that `last_action_sec` and re-set the countdown time accordingly (so less android system calls in that case).

(^ Just some generic thoughts :)

pazos:
> I haven't followed how you implemented that, but I thought, as it's some android stuff, Android would handle that itself, no?

android has the setting ScreenOffTimeout which applies to all activities. 

Common methods for an activity to request a different timeout are wakelocks and window settings. Acquiring a wakelock for screen_brightness and add the flag `WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON` is internally the same, but google recommends the flag because it handles activity state changes automatically.

We use wakelocks directly and [handle lifecycle manually](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/BaseActivity.kt#L425-L436).

Wakelocks are acquired during a certain time and then released. Each time the activity is resumed from pause the wakelock is re-acquired.

pazos:
I think I will left this in this totally non-borked state for a few weeks. The reason is lack of interest in this specific issue and a few things that I want to achieve in the little time I had.

I would still like to bump luajit-launcher as it includes a few other great things:

- migration to kotlin: that means no more runtime exceptions, except those that can be triggered calling a bad named method in lua/JNI.
- the new e-ink friendly spinner and splash screen that @ColinKinloch wrote.
- the, hopefully, more reactive GC which should now trigger more often due to how things are now allocated. That also means that it should be able to free more memory if requested by the system.

If you think the "semi-borked" timeout state is not good enough for a quick merge I will reimplement the old code to set custom timeouts while we don't have anything better. I call it semi borked as is doesn't affect normal users, which use system settings, or sony DPT-R1 users, which always use wakelocks without taking ScreenOffTimeout into account.

Frenzie:
Wouldn't E Ink devices normally come with a higher timeout than backlit devices by default because the screen doesn't actively use energy or is that not the case?

pazos:
> Wouldn't E Ink devices normally come with a higher timeout than backlit devices by default because the screen doesn't actively use energy or is that not the case?

I guess so but I'm not sure.

NiLuJe:
On a slightly related note, on my One Plus 6, the system's "Reading Mode" (which basically switches to grayscale + night mode) does keep the screen on, disregarding the usual system & app settings.

Frenzie:
> I guess so but I'm not sure.

Basically I just don't really know what the actual impact would be.

pazos:
The timeout setting is probably more apealing to tablet owners. Most usage of the tablet will require little timeouts without user input (search web, mail, background jobs like playing music..) or the screen on (youtube, netflix, vlc..).

In that context having a global timeout of 30-60s and override the timeout on reading apps to 3-10 minutes seem nice.

Frenzie:
It doesn't sound like a horrible flaw to me, just like a possibly nice to have. /shrug

pazos:
Closing this issue, Will open a new one. Sorry for the mess :cry:

Issue #5226, opened 2019-08-19
I cant open epub files (created frome html- by "Save as eBook" Chrome extension")


ocra98:
* KOReader version: v2019.08
* Device: likebook mars t80D, android 6.0.1

#### Issue
Cant open epub file (~6mB); or very slow (5-6s) with small epub file (400kb). See attached file (zip).
This is "Save as eBook" extension i'm using: [https://github.com/alexadam/save-as-ebook](https://github.com/alexadam/save-as-ebook)

[i cant open this epub file- sample.zip](https://github.com/koreader/koreader/files/3516264/i.cant.open.this.epub.file-.sample.zip)

Frenzie:
The smaller EPUB seems to open quickly enough for me, but the larger one definitely takes a while. It's almost certainly the CSS, which looks absurdly complex. 4.4 MB worth of it!

poire-z:
On the small one, which only has 240 CSS classes and only a hundred of node, it still takes 2 seconds on the emulator for applying styles:
```
2019/08/19 16:57:48.2861 TRACE CSS file href: OEBPS/style/style7638.css
2019/08/19 16:57:48.2862 DEBUG Setting StyleSheet attribute to OEBPS/style/style7638.css for document fragment
2019/08/19 16:57:48.2862 TRACE ldomNode::applyNodeStylesheet()
<2s pause here>
2019/08/19 16:57:50.4990 DEBUG Loaded font
```

If I keep the 240 classes, but make their properties smaller, eg just one property, it loads fast:
```
2019/08/19 17:03:12.7378 TRACE CSS file href: OEBPS/style/style7638.css
2019/08/19 17:03:12.7379 DEBUG Setting StyleSheet attribute to OEBPS/style/style7638.css for document fragment
2019/08/19 17:03:12.7379 TRACE ldomNode::applyNodeStylesheet()
2019/08/19 17:03:12.7410 DEBUG Loaded font 
```
I checked with some additional logs that it's really the parsing of that not-that-huge style7638.css that is taking time:
```
2019/08/19 17:11:46.3221 WARN parsing start
2019/08/19 17:11:48.5118 WARN parsing end
```
We may have some issue with the parsing of border* properties, as changing the 240 classes to that:
```css
 .class-24960 {background-color:black;
     zzborder:0px none black;
     zzborder-top:0px none black;
     zzborder-right:0px none black;
     zzborder-bottom:0px none black;
     zzborder-left:0px none black;
color:black;
[...]

PR #4126, opened 2018-07-30
News Downloader - feed config editor

mwoz123:
zrzut ekranu z 2018-07-30 23-12-21

No validation yet - in next phase.

PR #5428, opened 2019-09-26
per directory defaults

yparitcher:
extend `DocSettings` to allow per directory defaults for book settings.
if the file does not have a sidecar it recursively looks for a directory default to put as the settings.
it stops on the first parent directory if finds defaults for (children override parents).

***

it uses three new settings
`directory_defaults_enabled` - a boolean
`directory_defaults_base` the base directory to search until. (there is no need to go back to `/` if you have a documents folder)
`directory_defaults` - a table of settings tables keyed to directories

```lua
    ["directory_defaults_enabled"] = true,
    ["directory_defaults_base"] = "/mnt/us/documents",
    ["directory_defaults"] = {
        ["/mnt/us/documents/hebrew"] = {
            ["inverse_reading_order"] = true
        }
    },
```

***

right now this is the backend but i plan to add a frontend to this (plugin based?) with three options.

1. turn on/off
2. set base directory
3. for a directory set the default settings (from current book?)

***

before this goes anywhere i have a few questions

A) are the setting names ok?
B) this is currently saved in `setting.reader.lua` is that ok? and if not where should i store it?
C) how can i set the base to the home directory by default (on install)?
D) can i assume we are on linux and directories are delineated by `/`?
E) do i have to worry about conflicting setting for different engines?
F) is there already a way to check if a setting is doc specific (ie: `bookmarks`) so as not to allow it?

this is similar but different to #5079.


poire-z:
Well, I don't really like it, I think it's too (your) use-case specific :)

Because I may want do to the same based not on the directory the book is in, but based on the book language - or book author - or book publisher - or book format. And I actually did (a set of style tweaks when the Author is "Wikipedia *", or the publisher is "Save as eBook"), in my own patch (that I apply on Koreader after each install for my specific needs that I think won't interest many people, or would make the code messy).

After your precedent, I or other may want to add `lang_defaults_enabled`, `author_defaults_enabled`, `publisher_defaults_enabled` - and we'll have tons of rarely used and maintained settings and code...

For those like you and I who know how to hack, I'd prefer to have hooks for such things via external files. For others, well, learn a bit of Lua and hack your external files :) (like we have for HTML dicts output tweaking).

But, for your specific case: default reading order by book language, assuming your books have the correct language tags, a real nice feature would be an option to set reading order based on the book language (arabic, hebrews get RTL). May be not as the default, but it could be set with Hold on "Invert reading order" menu item (or "Invert page turn taps and swipe" is that's its new name).

(I thought the above before I had a look at the code - which is so small that I may not really care. So, if others feel like it may have some other use cases, I won't oppose :)

frontend/docsettings.lua
```diff
@@ -150,12 +151,30 @@ function DocSettings:open(docfile)
...
 
     return setmetatable(new, {__index = DocSettings})
 end
 
+function DocSettings:getDirectoryDefaults(docfile)
+    local data = {}
+    if G_reader_settings:readSetting("directory_defaults_enabled") then
+        local defaults = G_reader_settings:readSetting("directory_defaults")
+        local base = G_reader_settings:readSetting("directory_defaults_base")
+        local directory = docfile and docfile ~= '' and docfile:match("(.*)/")
```
Frenzie: Btw, our style is `"`, not `'`.

yparitcher: copied from https://github.com/koreader/koreader/blob/3de3ea04ffe64f73301e9cbec305fab406530638/frontend/docsettings.lua#L32
will fix.

Frenzie:
> A) are the setting names ok?

Looks fine at a glance, but is there a particular rationale behind splitting the existing home directory and the base folder for directory settings?

> directory_defaults - a table of settings tables keyed to directories

I don't think that should go in settings.reader.lua. It could get unwieldy quickly. That should be either something like `.directory` in the directory itself or a separate settings file.

> D) can i assume we are on linux and directories are delineated by /?

No but still yes. You should generally use or create utility functions in any case instead of doing one-offs for path business for various reasons. (From unit tests to prevent regressions to transparent swaps that benefit the whole program.)

> E) do i have to worry about conflicting setting for different engines?

Unlikely.

> F) is there already a way to check if a setting is doc specific (ie: bookmarks) so as not to allow it?

Otoh I don't think so.

@poire-z 
> Because I may want do to the same based not on the directory the book is in, but based on the book language - or book author - or book publisher - or book format.

But to counter that, a directory is the easiest and most generic way of achieving that.

frontend/docsettings.lua
Frenzie: There seem to be a grand total of half a dozen documents with single quotes in 'em. :-P

PS Always feel free to do `'Blabla "this is a quote" blabla'` when it's more elegant that way than escaping.

poire-z:
> But to counter that, a directory is the easiest and most generic way of achieving that.

But why counter that ? :) Why force me to put my hebrew books in a same directory, just to be able to use that feature - and prevent me from putting my philosophy books (french and hebrew) in Philosophie/ and my theology books (english and arabic) in Théologie/ ? The title names are enough for me to see what languages they are in. (I guess it's the same for Wallabag, all articles might come in into the same directory.)
Better to have the feature generic/flexible, if we're adding it.

> (I thought the above before I had a look at the code - which is so small that I may not really care.)

Well, the backend code might be small, but the UI part to set the generic settings per directory won't be :|
A cleaner option would be to have a plugin, easy to extend with various rules, and some event that would be broadcasted by core like `UIManager:broadcastEvent(Event:new("BookFirstLoad", initial_empty_docsetting_table))` , and handlers could add their stuff to that empty directory.

yparitcher:
> > A) are the setting names ok?
> 
> Looks fine at a glance, but is there a particular rationale behind splitting the existing home directory and the base folder for directory settings?

no, i will probably just use `home_dir`
> > directory_defaults - a table of settings tables keyed to directories
> 
> I don't think that should go in settings.reader.lua. It could get unwieldy quickly. That should be either something like `.directory` in the directory itself or a separate settings file.

makes sense, a separate settings file is easier. 
> > D) can i assume we are on linux and directories are delineated by /?
> 
> No but still yes. You should generally use or create utility functions in any case instead of doing one-offs for path business for various reasons. (From unit tests to prevent regressions to transparent swaps that benefit the whole program.)

will probably switch to a util. function
> > E) do i have to worry about conflicting setting for different engines?
> 
> Unlikely.
> 
> > F) is there already a way to check if a setting is doc specific (ie: bookmarks) so as not to allow it?
> 
> Otoh I don't think so.

so i guess i have to whitelist settings then.

 @poire-z

good point, that i didn't think of. 

this is not just for that setting, i have many settings to tweak
btw does style tweaks support settings, or only css properties?

> But why counter that ? :) Why force me to put my hebrew books in a same directory, just to be able to use that feature - and prevent me from putting my philosophy books (french and hebrew) in Philosophie/ and my theology books (english and arabic) in Théologie/ ? The title names are enough for me to see what languages they are in. (I guess it's the same for Wallabag, all articles might come in into the same directory.)

the current abstraction that Koreader offers is file system based not metadata based (no collections etc) this is extending the current abstraction not creating a new one.
as of now the only real way to access books by author or lang is to put them in a corresponding folder hierarchy.

though it would be cool to have a metadata base viewer :)

> A cleaner option would be to have a plugin, easy to extend with various rules, and some event that would be broadcasted by core like `UIManager:broadcastEvent(Event:new("BookFirstLoad", initial_empty_docsetting_table))` , and handlers could add their stuff to that empty directory.

interesting idea, looks more future proof, but i am not sure exactly how i would go about doing that

Frenzie:
@poire-z 
> But why counter that ? :) 

Depends on the implementation complexity of the alternative. ;-)

@yparitcher yparitcher
> interesting idea, looks more future proof, but i am not sure exactly how i would go about doing that

Perhaps in ReaderUI:showReader(). There's already a call to DocSettings there. If it's not present then we're dealing with a first open.

poire-z:
> the current abstraction that Koreader offers is file system based not metadata based (no collections etc) this is extending the current abstraction not creating a new one.
as of now the only real way to access books by author or lang is to put them in a corresponding folder hierarchy.

I like the directory hierarchy as much as you do (I hope you do :) - and care less than you about the metadata :). And that abstraction is for organizing. I was just pointing that book settings tweaking might be orthogonal to that organisation (my use case), and that by using the directory hiearchy for that (albeit a simple first and quick solution to that problem), being non-orthogonal, the possibilites are limited.

But ok, given how small it is, why not. You would not even need any setting. Just do what you do, looking in a few directories up to the home folder - so, no need for your "base dir" setting, we already have it: `G_reader_settings:readSetting("home_dir") or require("apps/filemanager/filemanagerutil").getDefaultDir()`
And I'd say: don't even put any UI, let users set up manually their own base settings table, easy enough by copying from an existing one and cleaning up (because doing a UI part, you would need to decide which settings are valid and which are not, like last_xpointer, current_page, bookmarks...) - and that would be really similar to the WIP PR about profiles.

> this is not just for that setting, i have many settings to tweak
btw does style tweaks support settings, or only css properties?

Nope, only CSS.

Unrelated to what you're doing but for info, for my use case (depending on book metadata), it needed to be delayed till the book is half loaded and we have access to the metadata, so I had to hook it into a onPreRenderDocument handler. My stuff at bottom of https://github.com/koreader/koreader/issues/5226#issuecomment-522664869.

Another option for what you want could be a single user file in `koreader/settings/book.default.settings.lua` where one could define a few functions to fill his needs, like:
```lua
function tweakDocSettings(docsetting, filepath, doc_props, hour_of_day)
  -- docsetting is either empty, or the saved one
  local first_load = false
  if not docsetting.last_xpointer then -- find something similar for pdf
     first_load = true
  end
  if firstload and filepath.startswith("/mnt/onboard/.Library/Hebrew") then
     docsetting["inverse_reading_order"] = true
  elseif firstload and  filepatch.contains("/hebrew/") then
     docsetting["inverse_reading_order"] = true
  elseif firstload and doc_props["language"].lower() == "he" then
     docsetting["inverse_reading_order"] = true
  elseif firstload and filepatch.contains("/comics/") then
     docsetting["night_mode"] = true
  elseif hour_of_day > 20 or hour_of_day < 8 then -- even if not first opening, override the setting one
     docsetting["night_mode"] = true  
end
```
(That needs more Lua knowledge, but would be more flexible.)
Just sharing alternatives :)

yparitcher:
based on your feedback i refactored it int a plugin called by ReaderUI:init()

i have not yet figured out where and how to store the table.

should i include the logic in the plugin or just call a user defined callback for each person to do as they see fit, and if so what is the best way to call it directly or via an event?

also before the settings are initialized (while in ReaderUI:init) the _window_stack is empty therefore `broadcastEvent` does nothing and i had to call the event directly via `handleEvent`. is there any way around this? without event handling i have no way of IPC

frontend/apps/reader/readerui.lua
```diff
@@ -382,6 +382,10 @@ function ReaderUI:init()
         })
     end
 
+    -- Allow others to change settings based on external factors
+    UIManager:broadcastEvent(Event:new("DocSettingsLoad", self.doc_settings, self.document.file))
+    self.docsettingtweak:handleEvent(Event:new("DocSettingsLoad", self.doc_settings, self.document.file))
```
poire-z: Or you could just do as on the next line (if the next line work, this should):
```lua
+    -- Allow others to change settings based on external factors, before feeding it to onReadSettings
+    self:handleEvent(Event:new("TweakSettings", self.doc_settings))
+
     -- we only read settings after all the widgets are initialized
     self:handleEvent(Event:new("ReadSettings", self.doc_settings))
```
Also, because plugins can be disabled, your `self.docsettingtweak:handleEvent` would cause a crash when `self.docsettingtweak` does not exist.

plugins/docsettingtweak.koplugin/main.lua
```diff
@@ -0,0 +1,30 @@
...
+local DocSettingTweak = WidgetContainer:new{
+    name = 'docsettingtweak',
+}
+
+local directory_defaults = {
+}
+
+function DocSettingTweak:onDocSettingsLoad(doc_settings, docfile)
+    if next(doc_settings.data) == nil then
+        local base = G_reader_settings:readSetting("home_dir") or "/"
```
poire-z: instead of `or "/"`:
`or require("apps/filemanager/filemanagerutil").getDefaultDir()`

poire-z:
> based on your feedback i refactored it int a plugin called by ReaderUI:init()
> i have not yet figured out where and how to store the table.
> should i include the logic in the plugin or just call a user defined callback for each person to do as they see fit, and if so what is the best way to call it directly or via an event?

Thanks, I really prefer it now as a plugin. And you're a bit more free to do whatever UI things you want, in that plugin :)
A lazy but flexible option would be to just use something like the TextEditor to edit your default mapping `directory => file with the default table`, or directly the `directory => default table`.
Not the TextEditor plugin directly (you can't), but the powerfull base for it: InputDialog, as done in the NewsDownloader plugin (see #4126 and related):
https://github.com/koreader/koreader/blob/d0f95ec669447e8f3a2ee55972bdfdda4e49baa1/plugins/newsdownloader.koplugin/main.lua#L405-L444

With that, you (or others later) could propose to edit your `directory_defaults.lua` , or even some `docsettings_tweaker.lua` with a function with stuff like I proposed above. You would edit and store these 2 files in the koreader/settings/ directory, like some other plugins do.

(In the same vein, we could even have a `koreadertweak.plugin` to monkey patch koreader :)

Frenzie:
> based on your feedback i refactored it int a plugin called by ReaderUI:init()

/twitch

Probably already mentioned (just passing by real quick) but that would defeat the purpose of a plugin. ;-)

yparitcher:
TODO: stil need to take care of storage

yparitcher:
> Probably already mentioned (just passing by real quick) but that would defeat the purpose of a plugin.

Why? what are the plugins meant for? it calls it via an event.

Frenzie:
Plugins are meant to be architecturally completely separated. The program can run just fine without any. The way you wrote it, it would've crashed without  the plugin. (But I responded  to the text, not the code.)

yparitcher:
@Frenzie @poire-z  please review. thanks

plugins/docsettingtweak.koplugin/main.lua
```diff
@@ -0,0 +1,106 @@
...
+    }
+    UIManager:show(config_editor)
+    config_editor:onShowKeyboard()
+end
+
+function DocSettingTweak:onDocSettingsLoad(doc_settings, docfile)
+    if next(doc_settings.data) == nil and directory_defaults.data ~= nil then
+        local base = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
+        local directory = docfile and docfile ~= "" and FFIUtil.dirname(docfile)
+        while directory:sub(1, #base) == base do
```
poire-z: Can we be called without a `docfile` ?
You take precautions checking if `docfile` is not `nil `and not an empty string to only then get its directory. And if not, directory is `nil`. But then, the `directory:sub()` might crash on a directory that is `nil`.
So, may be just return early if there is no `docfile` provided to make the whole thing simpler clearer.

plugins/docsettingtweak.koplugin/directory_defaults.lua
```diff
@@ -0,0 +1,33 @@
...
+ -- syntax:
+
+ --   ["path/to/folder"] = {
+ --         ["setting_to_override"] = value,
+ --    },
+
+ -- examples:
+
+ --[[
+    ["/mnt/us/documents/hebrew"] = {
```
poire-z: How would it do if I add a directory here with a trailing / (eg. `["/mnt/us/documents/hebrew/"]`) ?
(If it wouldn't work, either make it work :), or add some comment here that a trailing slash should not be included.)

frontend/apps/reader/readerui.lua
```diff
@@ -382,6 +382,9 @@ function ReaderUI:init()
         })
     end
 
+    -- Allow others to change settings based on external factors
+    self:handleEvent(Event:new("DocSettingsLoad", self.doc_settings, self.document.file))
```
Frenzie: Isn't `self.document.file` unnecessarily limiting?

plugins/docsettingtweak.koplugin/_meta.lua
```diff
@@ -0,0 +1,6 @@
+local _ = require("gettext")
+return {
+    name = 'docsettingtweak',
+    fullname = _("Document setting tweak"),
```
Frenzie: That sounds a bit weird. Some suggestions:

* Document settings tweaker
* Document settings tweak tool
* Tweak document settings

plugins/docsettingtweak.koplugin/directory_defaults.lua
```diff
@@ -0,0 +1,33 @@
...
+        ["inverse_reading_order"] = true
+    },
+    ["/mnt/us/documents/smalltext"] = {
+        ["font_size"] = 34,
+        ["line_space_percent"] = 130,
+    },
+--]]
+
+ -- remember to put coma at the end of each line!
```
Frenzie: Comments like to be proper sentences.
```suggestion
-- Remember to put a comma at the end of each line!
```

plugins/docsettingtweak.koplugin/main.lua
```diff
@@ -0,0 +1,106 @@
...
+    name = 'docsettingtweak',
+}
+
+local directory_defaults_name = "directory_defaults.lua"
+local directory_defaults_path = FFIUtil.joinPath(DataStorage:getSettingsDir(), directory_defaults_name)
+local directory_defaults = nil
+local initialized = false
+
+function DocSettingTweak:init()
+    DocSettingTweak:lazyInitialization()
```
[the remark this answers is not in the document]
Frenzie: Agreed. This is needless indirection.

plugins/docsettingtweak.koplugin/main.lua
```diff
@@ -0,0 +1,106 @@
...
+        while directory:sub(1, #base) == base do
+            if directory_defaults:has(directory) then
+                doc_settings.data = util.tableDeepCopy(directory_defaults:readSetting(directory))
+                break
+            else
+                directory = FFIUtil.dirname(directory)
+            end
+        end
+    end
+    return true
```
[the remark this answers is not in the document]
Frenzie: Indeed, stopping propagation here largely defeats the purpose.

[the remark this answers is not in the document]
yparitcher: i had thought that best practice was to always return true from an event, however i see how this can be used by others.

frontend/apps/reader/readerui.lua
yparitcher: not for my purposes, however i will change it for anyone else who wants to extend this based on some other heuristic

plugins/docsettingtweak.koplugin/directory_defaults.lua
```diff
@@ -1,9 +1,11 @@
...
 
  --HELP:
- -- you can override default settings for documents per directory in this file
- -- the directories must be under the home folder.
- -- you can find settings to change in the metadata.lua in the .sdr folder.
+ -- You can override default settings for documents per directory in this file.
+ -- The directories must be under the home folder.
+ -- You can find settings to change in the metadata.lua in the .sdr folder.
+ -- The path must start with "/" (absolute path).
+ -- The path may not end with a "/" it must end with the folder name.
```
poire-z: "_must_ not end with a /" then? ("_may not_" sounds to me like it would be allowed).

plugins/docsettingtweak.koplugin/directory_defaults.lua
```diff
@@ -0,0 +1,38 @@
+return {--do NOT change this line
+
+ --HELP:
+ -- You can override default settings for documents per directory in this file.
```
Frenzie: To prevent people from editing *this* file, maybe the file should be called something like directory_defaults_template.lua instead and not directory_defaults.lua until it's copied over to ./settings?

plugins/docsettingtweak.koplugin/main.lua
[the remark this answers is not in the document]
Frenzie: We're conceptually talking about a generic hook here. Perhaps "hooks" and "events" could be separated, though I'm not currently inclined to think it'd make much of a practical difference in terms of real-world clarity or ease of use.

> i had thought that best practice was to always return true from an event

I would argue that requiring boilerplate code to reach a desired default state of affairs has appearances going against it.

But of course with a user input event, you'll almost certainly want to stop further propagation. Otherwise your on-screen button press would also turn the page, or something along those lines.

plugins/docsettingtweak.koplugin/main.lua
```diff
@@ -0,0 +1,107 @@
...
+local directory_defaults_name = "directory_defaults.lua"
+local directory_defaults_path = FFIUtil.joinPath(DataStorage:getSettingsDir(), directory_defaults_name)
+local directory_defaults = nil
+local initialized = false
+
+function DocSettingTweak:init()
+    if not initialized then
+        -- Make sure our settings file exists
+        if not lfs.attributes(directory_defaults_path, "mode") then
+            FFIUtil.copyFile(FFIUtil.joinPath(self.path, directory_defaults_name) .. ".template",
```
poire-z: This concatenation seems unnecessary. Clearer to just define above `local directory_defaults_template = "directory_defaults_template.lua"` (and put "template" before the extension, and keep it named something.lua so we don't lose Lua syntax highlighting.

yparitcher:
> Just template.lua immediately raises the question of template _for what_?

the plugin which folder it is in.

Frenzie:
That narrows it down a bit. It's not self-descriptive like `directory_defaults_template.lua`.

Frenzie:
The point I'm trying to make is that you have to open a file with a non-descriptive name to know what it is.

PR #5435, opened 2019-09-28
Series in separate line

derungs:
New option to show series in a separate line between book title and author(s) in ListMenu.
New configuration for number of items in non-classic file list.
Slight re-grouping of config menu.

series
display-mode-menu
items-menu


plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -745,10 +752,9 @@ function ListMenu:_recalculateDimen()
     end
     local available_height = self.dimen.h - self.others_height
 
-    -- 64 hardcoded for now, gives 10 items both in filemanager
-    -- and history on kobo glo hd
-    local item_height_min = Screen:scaleBySize(64)
-    self.perpage = math.floor(available_height / item_height_min)
+    -- menu item height based on number of items per page
+    self.perpage = G_reader_settings:readSetting("files_per_page") or 10
+    local item_height_min = math.floor(available_height / self.perpage)
```
poire-z: Dunno if that, when keeping the default of 10, might change the computed item_height_min vs. the value before (if yes, the following issue will affect all users).
There's a small issue anyway when you change files_per_page: the cached thumbnails will be kept as rendered with the first value - so they may look a bit bad when you change that `files_per_page` value.
Not sure there's a way to invalide them and trigger a refresh - or if we let that to the user...

Or may be that's handled by `cover_specs` ? and if not, it could be included in sizetag/cover_sizetag, so you could use `"s10"` or `"s8"` or `"s14"` (when 10 or 8 or 14 files-per_page) instead of the simple `"s"`?

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -266,6 +269,50 @@ function CoverBrowser:addToMainMenu(menu_items)
                 },
                 separator = true,
             },
+            {
+                text = _("Series display mode"),
```
poire-z: I'd rather see it in/as `Other settings > Series display mode >` (a bit _niche_ to be in the 1st level).
(With the 2 `Append series metadata...` in it, and even with your `(Detailed list) Files per page` added, we'll have a `Other settings >` with still and only 10 settings, that will fit on one page.

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -129,6 +129,29 @@ function FileManagerMenu:setUpdateItemTable()
...
         text = _("Items"),
         sub_item_table = {
+            {
+                text = _("Files per page"),
+                help_text = _([[This sets the number of files and directories
+                per page in non-'classic' display modes.]]),
+                keep_menu_open = true,
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local curr_items = G_reader_settings:readSetting("files_per_page") or 10
```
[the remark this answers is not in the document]
derungs: Yes you're right, it only applies to the detailed file list.
It seems to me that the four top items in "Other settings" only apply to the Detailed list - if that's the case, I'll add it there without the parenthesis. Maybe add a sub-menu?
other-settings

plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -228,6 +228,35 @@ function BookInfoManager:loadSettings()
     for i, key in ipairs(keys) do
         self.settings[key] = values[i]
     end
+
+    -- translate old append_series_to_authors and append_series_to_title to new series_mode
+    if not self.settings["series_mode"] then -- initialize from old settings
```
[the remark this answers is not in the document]
derungs: Good point. I considered just dropping backwards compatibility, but that's something that tends to annoy me, so I didn't want to do it to others :)

frontend/apps/filemanager/filemanagermenu.lua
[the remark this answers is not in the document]
poire-z: The `Show hints` must apply to both mosaic & detailed.
May be to give a small visual hint to what applies to what:
```
Show hint for book with description
Show hint for opened books in history
Series>
---------
(Detailed list) Files per page
Show number of page read
Show number of pages left to read
---------
(Classic mode) Show open in bold
---------
The others...
```
Just beware the following if you change the order:
https://github.com/koreader/koreader/blob/a934d2d52e45a52b2232689e04701beb7db5c4b6/plugins/coverbrowser.koplugin/main.lua#L422-L424

plugins/coverbrowser.koplugin/listmenu.lua
derungs: When testing, the thumbnail sizes get adapted immediately, even on my old and slow Tolino Vision 2; so I don't think caching is an issue. Makes me wonder if the distinction between "w" and "s" is even necessary, or if we could just go with one size and have the device adapt it when displaying.

plugins/coverbrowser.koplugin/main.lua
derungs: done

frontend/apps/filemanager/filemanagermenu.lua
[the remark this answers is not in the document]
derungs: Yes, saw it now. And only now that you mention it, I see the little double edge on the top right for descriptions. So everything still seems to work as it should.
From the screenshot above, I assume that the "insert generic items" still works as it should - I'll take your comment as a hint for future changes.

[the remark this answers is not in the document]
derungs: Do you use a visual hint for menu items that open a dialog? In many apps it's three dots. In KOReader I haven't seen a visual hint, e.g. in "Items per page" or "Font size".
Is this something we should consider to introduce on a global level?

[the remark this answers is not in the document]
poire-z: I guess we have that inverted :)
Menu items that don't open anything have a checkbox on their left, and it just toggle the checkbox.
All others either open a dialog, or launch something (possibly a full screen dialog, like TOC).
(Three dots would just add a bit of unneeded noise, imho.)

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: Yes, they are resized to fit the wanted geometry if the saved blitbuffer has not the right size - but only in that case.

But if I took the pain of saving a rightly sized blitbuffer in the DB, it's to avoid that resize. Because if I introduced it, it's probably that I noticed some small slow down when it has to resize 9 or 10 small covers to a different small size, when you're browsing pages fast.
So, if you use only the `Detailed list` mode, we save the rightly sized small size blitbuffer (`"s"`)- and no resize needed when you quicly browse many file browser pages.
If you use Mosaic, we save bigger blitbuffers (`"M"`).

When you have both (Detailed for File browser, Mosaic for History, I think many people do that), we save the `"s"` when books are met in File Browser, but when meeting them in History, we notice if they have the `"s"` sizetag, and if they do, we refetch the original cover and resize it down to the bigger `"M" tag, and replace the blitbuffer in the DB, so they are nice in History (instead of a ugly "s" resized up to a bigger size). For quality.
Then, when back in File Browser, we may find some (those that we met once in History) with a "M" sizetag, and these indeed need to be resized down (on the fly, while browsing pages - we don't replace the M blitbuffer in the DB, we want to keep it for next time in History). There's a small slow down, but there shouldn't be too much books in History vs the amount we'll meet in File Browser.

So, not super necessary, but good to have and sad to drop :)

plugins/coverbrowser.koplugin/bookinfomanager.lua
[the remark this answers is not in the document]
poire-z: (Not resolved).
I insist: you don't need any SQL in there. Just use `saveSetting()`.

There's no real backward compatibility needed, except for users that downgrade from time to time (I do when checking for regressions - but I woudln't mind much losing that series setting :)
I wouldn't even mind if we were not migrating it :) Users would notice and would then go see that menu - and see your new option.

But if you do want migration, and want to keep the old settings, and avoid recursion when dealing with them, just use:
```lua
if not self.settings["zzz_migration_series_setting_migrated"] then
    self:saveSetting("zzz_migration_series_setting_migrated", true)
        -- should stop re-doing the following when saveSetting() re-loadSettings()
    local series_mode
    if self.settings["append_series_to_authors"] then
        series_mode = "append_series_to_authors"
    end -- no elseif, the user may have both
    if self.settings["append_series_to_title"] then
        series_mode = "append_series_to_title"
    end
    self:saveSetting("series_mode", series_mode)
end
```

frontend/apps/filemanager/filemanagermenu.lua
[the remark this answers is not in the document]
Frenzie: > Three dots would just add a bit of unneeded noise, imho.

The tl;dr is that I agree with that.

A summary of the much longer reply is that implementation of this feature (?) is quite inconsistent, and the various HIG/UI guidelines by, e.g., Apple, Microsoft, and GNOME provide neither rationale nor guidance that I've found. At most they imply it by using `Open...` as an example menu item.

Android and Sailfish don't use the principle. Microsoft doesn't tend to use it in ribbon and post-ribbon interfaces.

If we look at older Microsoft apps, the implementation seems oddly technical. I suspect that perhaps it quite simply is (i.e., auto-generated in some manner). Unlike the vast majority of implementations, older Microsoft apps are slightly more consistent. Items like the about dialog are `About...` and options is `Options...`. This is not the case in virtually all other apps. But even so, an item like `Help` doesn't typically have an ellipsis (or three dots). It may not technically be a dialog of the program, but the result is the same. It opens a different window/dialog as opposed to applying an action to the current window. 

You might think that perhaps it's MS that was inconsistent and those other apps have a different idea, for example *dialogs that apply an action to the current document*, but no one seems to have thought that much about it. As far as I can tell it's completely up in the air whether a dialog will be indicated and to what scope it applies.

That's a brief summary of my analysis of existing implementations from a few years ago, when I considered implementing `...`.

Of course I could've decided to do away with the inconsistencies in what seems to be the least bad implementation (by Microsoft), but it just didn't seem to *add* anything worthwhile. This menu item opens a dialog? Does it matter? The only truly relevant distinction is whether it applies to the currently opened document or to the entire program. Whether it opens a dialog to do so strikes me as an irrelevant detail.

PS If you've found anything different, please share your arguments or references. I don't intend this to shut down discussion, merely to indicate that I quite possibly gave it more thought than most.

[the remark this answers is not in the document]
Frenzie: > Whether it opens a dialog to do so strikes me as an irrelevant detail.

Moreover, in our UI many actions apply either locally or as global defaults at the user's request. Tap for current document, long-press for global default.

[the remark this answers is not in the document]
derungs: I was mainly wondering if it was a conscious decision or a case of "we haven't thought about it". Your answer is quite satisfactory, and I have no inclination to argue for a change.

With respect to "tap for current document", selecting a page in the file chooser is triggered with long-press, tap has no effect. It seems the wrong way around to me, unless you have reasons to keep it that way, I would kinda  like to change that.

plugins/coverbrowser.koplugin/bookinfomanager.lua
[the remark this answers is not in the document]
derungs: I don't want to keep the old settings, migrating is sufficient.
The reason I didn't call saveSetting() is because it calls loadSettings(), which leads to a 3-deep recursion if both flags are set. I do get your point about not replicating SQL code, though, so I'll adapt the code.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: I'll give it a go then, and use "s<num>" for indication. I'll try to keep using "s" when files_per_page is not set to prevent unnecessary resizing.

frontend/apps/filemanager/filemanagermenu.lua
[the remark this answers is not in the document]
Frenzie: You'd have to check git blame to see if there's a particular reason. Perhaps it's for consistency with the skim dialog in the reader, which you can trigger by long pressing at the bottom. I don't think I ever gave the FileChooser controls much thought.

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: My maths suck :) so I dunno if the bit of code I commented on here does change the computed `item_height_min`.
If there's a risk it does, you may just use `"s10"` for the default, so they are refreshed for all users next time they browse some area (so that the next times, they are the right size without any resize needed) - better that that a sudden feeling browsing has become a bit slower because of the 9-10 resizes for a few pixels more or less.
You could also use `"s<item_height_min>"` (and `"M<item_height_min>"`), so that next time we do some outer UI tweaks and happen to change `available_height `, that will be automatically taken care of.
Although `item_height_min` may also just change when you call history from portrait book, and then from a landscape book... so, well... dunno :/

Frenzie:
@poire-z @derungs I lost track of this PR. Is the comment about saved icon sizes still relevant?

derungs:
Yes it is. I had a very busy week at work, didn't get to address it yet. I suggest taking the PR off the October release, by November it should be done.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: The height does change. On different size devices the 64 predefined height translates to a different number of files shown on one page.
Went with your suggestion, sizetag is now s<height_in_pixels>. I did not touch the "M" tag or size computation.

frontend/apps/filemanager/filemanagermenu.lua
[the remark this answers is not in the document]
derungs: I'll take that into a separate pull request.

derungs:
Something else I want to fix: The triangle on the bottom right gets size-adjusted to the number of entries shown in the menu. I want the triangle to always be the same size.

poire-z:
> Something else I want to fix: The triangle on the bottom right gets size-adjusted to the number of entries shown in the menu. I want the triangle to always be the same size.

But you would then get more risk of it overwriting the text - see #4081 - so I think it should have a smaller size when the nb of items increases, and the size of each decreases, as so does the font size of the text in each item. (If one wants many items, he's ok with smaller text, so he'll be ok with a smaller dogear.)

poire-z:
Btw, you have some screenshot  `other-settings.png`  included in the PR files.

derungs:
> > Something else I want to fix: The triangle on the bottom right gets size-adjusted to the number of entries shown in the menu. I want the triangle to always be the same size.
> 
> But you would then get more risk of it overwriting the text - see #4081 - so I think it should have a smaller size when the nb of items increases, and the size of each decreases, as so does the font size of the text in each item. (If one wants many items, he's ok with smaller text, so he'll be ok with a smaller dogear.)

I don't mind making it smaller if needed. When it gets bigger it starts to look seriously overweight.
large-dogear

poire-z:
Oh, I see :) I agree on limiting it to some max size (made from some other metric).
Like I did for the bookmark dogear:
https://github.com/koreader/koreader/pull/4081/files#diff-9a17fdf36f33bc9340d6e20ccca83586R19
Whatever that does not make it smaller than what it currently is with the (or my) default of 10 items in Detailed list mode :)

plugins/coverbrowser.koplugin/bookinfomanager.lua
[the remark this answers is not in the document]
derungs: I'm having second thoughts, and cannot decide between keeping obsolete append_* settings lying around and using series_mode="none" instead of series_mode=nil.
Which variant do you prefer?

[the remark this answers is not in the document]
poire-z: Personally, I would just prefer if there was no migration code :) (I don't mind if we don't migrate them - as we would then not carry that migration code for years).
Otherwise, I have no preference (they are stored in `bookinfo_cache.sqlite3` that no one looks at - I would mind just a tad more if they would be stored in G_reader_settings, but not much :)
After that, `series_mode=nil` makes just more sense if you're not ever interested in series.

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -294,6 +298,74 @@ function CoverBrowser:addToMainMenu(menu_items)
...
+                                    self:refreshFileManagerInstance()
+                                end,
+                            },
+                        },
+                        separator = true
+                    },
+                    {
+                        text = _("(Detailed list) Files per page"),
+                        help_text = _([[This sets the number of files and directories
+                        per page in non-'classic' display modes.]]),
```
poire-z: This should be on a single line - possibly wrapped the following way if you feel it's needed:
https://github.com/koreader/koreader/blob/65fcd385c01f6eb6a1bab7d3404899938c04eaa1/frontend/apps/reader/modules/readerlink.lua#L134-L136
(The leading space is removed, but that's the only stuff removed - your big amount of spaces in the middle would stay there, and show up to translators.)

poire-z:
I don't want to dig into all the conversations :) but except for your dogear size limiting, is there anything else you want to add? Should I go test it?

derungs:
dogear is the only bit left.

plugins/coverbrowser.koplugin/main.lua
derungs: So if I were to go with max 100 characters per line, I would use string concatenation?

Frenzie: No, never. The line should be as long as it needs to be. If it feels too long, rewrite it to be shorter.

(But this line is only 86 characters long…?)

Frenzie: Minor correction: xgettext does happen to support the practice, so since #5237/#5242 it's technically possible.

However, the fact remains however that I'm opposed to abiding by line length limits like some kind of machine, especially if it results in ugly string concatenation.

For code it's a fairly strong indicator that you might want to consider refactoring, but text isn't code. At the same time, for text it can also an indicator that you might want to consider rewriting for clarity and brevity depending on context. But of course only if you're talking about line length within the string, not indentation plus string length.

As @poire-z said, for longer strings it works better to do:
```lua
                                some_string = [[
This is the string.]]
```

derungs:
Took a page from your book @poire-z and made the dogear size depend on the ListMenuItem height, like you did on the Mosaic menu. Checked that it fits under/beside the text on all `files_per_page` settings on 4 different reader sizes.
I'm leaving the `series_mode` as is.
Unless there's something else from code review, I won't push any more changes.

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -388,16 +398,25 @@ function ListMenuItem:update()
...
-                face = Font:getFace("cfont", 14),
-                fgcolor = self.file_deleted and Blitbuffer.COLOR_DARK_GRAY or nil,
-            }
+            local fontsize_info = 16
+            local wfileinfo
+            local wpageinfo
+            -- Build file and page info texts with decreasing font size
+            -- till it fits in the space available
+            repeat
+                wfileinfo = TextWidget:new{
```
poire-z: For the right stuff, where file size and type is always one line, and page info another line, I don't think you need to do this repeat stuff and build a widget each time and decrase font size. (It's uneeded cpu/power usage/latency - but it is ok/needed when we have vertical space where we can fit a variiable amount of lines and as much as the text - which is not the case here.)
You ought to be able to compute the right font size that will fit - or just find some value from a factor of the available height.

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -419,11 +438,10 @@ function ListMenuItem:update()
             }
 
             -- Create or replace corner_mark if needed
-            local wright_bottom_pad_available = math.ceil( (dimen.h - wright[1]:getSize().h) *2/3 )
-            -- We should normally use 1/2 because of CenterContainer, but there's
-            -- some space inside the text widget that we can use for a larger marker
-            if wright_bottom_pad_available ~= corner_mark_size then
-                corner_mark_size = wright_bottom_pad_available
+            local mark_size = math.floor(dimen.h / 6)
```
poire-z: Felt a bit sad to lose that curated formula for a vulgar 1/6 :) but it seems to look OK on the emulator.
Might depend on how you go with the previous comment.

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -447,10 +465,12 @@ function ListMenuItem:update()
 
             local fontname_title = "cfont"
             local fontname_authors = "cfont"
-            local fontsize_title = 20
-            local fontsize_authors = 18
+            local fontsize_title = 24
+            local fontsize_authors = 22
```
poire-z: I don't want my title and authors bigger when I'll set the same nb of list items I have.
You must have increased that because you need to start from a bigger font size when you have big items (small number of items) - but it shouldn't affect smaller items (large number of items).
Again, it should be a factor/computation vs the available height (possibly keeping the factor the current static nb of items has, to not surprise users with too much change).

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -745,17 +770,14 @@ function ListMenu:_recalculateDimen()
...
     self.page_num = math.ceil(#self.item_table / self.perpage)
     -- fix current page if out of range
     if self.page_num > 0 and self.page > self.page_num then self.page = self.page_num end
 
-    local height_remaining = available_height - self.perpage * item_height_min
-    height_remaining = height_remaining - (self.perpage+1) -- N+1 LineWidget separators
-    self.item_height = item_height_min + math.floor(height_remaining / self.perpage)
+    -- menu item height based on number of items per page
+    -- add space for the separator
+    self.item_height = math.floor(available_height / self.perpage) - 1
```
poire-z: What's this `-1` about ? Doesn't math.floor() have it right? Is this due to some added padding or... ?

poire-z:
I'm not really happy with the result.
I (and others) might be used to the current density (text vs blank space), and any change might feel like a regression. But changes of that density from one item to the other feels worse.

Screenshots on the emulator:

Original (8 items per page on the emulator):
<kbd>FileManager_2019-Oct-12_211417</kbd>

<kbd>FileManager_2019-Oct-12_215421</kbd>

With this PR (first with 8 items per page too):
<kbd>FileManager_2019-Oct-12_212120</kbd>

<kbd>FileManager_2019-Oct-12_211643</kbd>

<kbd>FileManager_2019-Oct-12_211758</kbd>

<kbd>FileManager_2019-Oct-12_211849</kbd>

(This last one is the only one I find not too bad :)

plugins/coverbrowser.koplugin/listmenu.lua
derungs: Good point - I followed the example from the title and author, where the text length may vary.
Do you have an idea on how to compute the text widget height from the font size? Otherwise, I'll have to dig into documentation and code, I don't want to estimate.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: Shouldn't make a difference wrt changes above, because the 2-line text still has to fit into the ListMenuItem, so it won't overlap.
Your formula is nice :) but on some settings the dogear would actually overlap the text.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: In the default settings, the title and author pretty much fill the height, so not much change. If there is a lot of whitespace above and below, why not make the text a bit biger?
The loop is not my code... But I'll give the computation a try, but it's probably a lot harder than with the file info, since the title can span more than one line. I shoud be able to decrease the font size based on widget size vs text size rather than by 1, to reduce the number of loops.

Frenzie: Whitespace is not a bad thing. It helps direct the eyes.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: It's because the lines between the ListMenuItem widgets are not part of the ListMenuItem widget. Without the `-1`, the last line can get overlapped by the forward/backward icons. That is what I wanted to express with the second comment line.

plugins/coverbrowser.koplugin/listmenu.lua
Frenzie: To put it another way, it's a compromise. I think we can all agree that we want the whitespace as small as reasonably possible in such an information dense environment. Based on some of the screenshots below I'm inclined to agree the "after" situation ends up a touch too busy. I think it's because the size seems to be both bigger (reducing needed whitespace) and kind of random?

derungs:
> But changes of that density from one item to the other feels worse.

That comment I don't understand - isn't this the case in the original as well, in particular on the second screenshot?
Reverting the max font size to the previous values would make the density the same as before the pull request, but honestly, one of my reasons for reducing the number of files per page is to see bigger text. Then again: add more padding around the title and author would also reduce the visual change.
The page info text is relatively long, could probably be shortened.

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: I'd say, if previously we had:
```lua
item_height_min = Screen:scaleBySize(64)
face = Font:getFace("cfont", 14) -- 14 gets scaleBySize'd too
```
A factor of `item_height * 14 / 64` might be worth trying. (eventually adjust it to 16/64 = 1/4 :)

Otherwise, there's that:
https://github.com/koreader/koreader/blob/a71b6250e2ba25476fb4f09b1d1a1b24d6275bd4/frontend/ui/widget/textwidget.lua#L55-L65

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: > why not make the text a bit biger?

Well, that's a question of taste :) I personally don't like it, 

> The loop is not my code... 

Indeed. But the starting font size is yours :) May be just finding the right one will make things fine.
Similar to my comments about the right stuff font size, title font size starts at 20, authors' at 18.
So, starting from `item_height * 20 / 64 ` and `item_height * 18/64` should give a similar result as before.
When it does, and only then, we can discuss about these factors (I'll fight against increasing them, but if people agree with your preference, I'll just have a simple factor to change in my own patch.)

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: Oh, allright, I previously also used some -1:
https://github.com/koreader/koreader/blob/a71b6250e2ba25476fb4f09b1d1a1b24d6275bd4/plugins/coverbrowser.koplugin/listmenu.lua#L757
Dunno if you need to account for the extra one (N+1)?

We use LineWidget:new{... h = Size.line.thin } which happen to always be 1px on most DPI.
That `-1` should probably be replaced with `- Size.line.thin`  for correctness.

poire-z:
> The page info text is relatively long, could probably be shortened

It probably will if you start with that 14/64 factor, and possibly cap it to some max value.

> That comment I don't understand - isn't this the case in the original as well, in particular on the second screenshot?

Yes, possibly, but in some limited ratio. Or the space around make it acceptable (to me).

> Reverting the max font size to the previous values would make the density the same as before the pull request, but honestly, one of my reasons for reducing the number of files per page is to see bigger text

OK, but that can be thought of as 2 orthogonal things.
Font size might just follow linearly the item height (so, it's just a zoom and all ratios would be the same), which seems to be my preference.
And there could be another orthogonal setting to zoom just the font size (so, the starting font size factor).

plugins/coverbrowser.koplugin/listmenu.lua
derungs: I like the idea of removing the fixed upper font size in favor of a variable size. With that, there's no real value for me to fight you on the factors :)
I'll try and keep the font size on the default setting (on the emulator) as close to the original visual as possible.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: Will do. And I'll check the code, but I suggest the top-most line comes from the header rather than the menu.

poire-z:
(And if you ever go at having 2 settings: `items per page` & `font size variation`, you could make these 2 in a submenu of the `Other settings>` menu - because currently it being 10 items, you don't see much of the effect it has as the menu covers most of the file browser - with a submenu of 2 items, you'd have room enough to see the effect of you playing with these 2 settings.)

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: (Note that your item_height will be pixels, so already "scaledBySize'd" - so you'd have to invert-scaleBySize to get the font size (as it will be scaleBySize'd by Font:new)

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: > the top-most line comes from the header rather than the menu.

There's a top line only for Detailed list. Not for Mosaic and not for Classic. So that might not be a good idea.

derungs: Taking the extra line into account, using Size.line.thin instead of 1.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: scaledBySize reacts to changes in height/width ration much more than changes in height alone. So, using a 64/20 ration for title font:
* font size grows bigger in square viewers _or_
* font size grows smaller in 4/3 screens

depending on whether scaleBySize is used on the font or not... Still wondering about that one. Currently I'm going for same density on 720/540.

derungs: Ah, I'll have to revert dbg.lua and change all my logs from warn to debug. Sorry 'bout that.

poire-z: Well, I don't know of any squared device :) But I see what you mean. I guess this would also apply when switching to landscape (FileBrowser don't do landscape, but History does if you wanna test).

I'm really fine with your `font_size = math.ceil(nominal * dimen.h / 64)` rule, as I'm then sure we get the ratio we had before on all size.
And testing a bunch of settings, they ALL look really fine and coherent to me now :) so :+1: !

(May be add a comment about why that 64, like "previously, with a fixed size, blah blah was using 64 to get 10 items on gloHD - so let's start from that rule to keep the old aspect")

Another option would be for you to add a `scaleByHeight()`, either here or alongside `fb:scaleBySize(px)` in `base/ffi/framebuffer.lua`.

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -388,16 +409,27 @@ function ListMenuItem:update()
...
+                    fgcolor = self.file_deleted and Blitbuffer.COLOR_DARK_GRAY or nil,
+                }
+                wpageinfo = TextWidget:new{
+                    text = pages_str,
+                    face = Font:getFace("cfont", fontsize_info),
+                    fgcolor = self.file_deleted and Blitbuffer.COLOR_DARK_GRAY or nil,
+                }
+                local height = wfileinfo:getSize().h + wpageinfo:getSize().h + Screen:scaleBySize(2)
+                fontsize_info = fontsize_info - 1
+            until height < dimen.h -- now it fits
```
poire-z: (Just a reminder: we still don't need that repeat...until.)

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -743,19 +785,16 @@ function ListMenu:_recalculateDimen()
...
     self.page_num = math.ceil(#self.item_table / self.perpage)
     -- fix current page if out of range
     if self.page_num > 0 and self.page > self.page_num then self.page = self.page_num end
 
-    local height_remaining = available_height - self.perpage * item_height_min
-    height_remaining = height_remaining - (self.perpage+1) -- N+1 LineWidget separators
-    self.item_height = item_height_min + math.floor(height_remaining / self.perpage)
+    -- menu item height based on number of items per page
+    -- add space for the separator
+    self.item_height = math.floor(available_height / self.perpage) - Size.line.thin
```
poire-z: You could compute all the font sizes here, to avoid recomputing it for each item (but that would may be make the whole thing less readable, not seeing the font size near the font name), so all that magic is contained and explained in `ListMenu:_recalculateDimen()`.
```lua
fontsize_directory_nb_items = math.min(20, _fontSize(15)))
fontsize_direction_name = fontSize(20))
fontsize_fileinfo = math.min(20, _fontSize(14)) -- don't go too large to give most room to title & authors
fontsize_title = _fontSize(20)
fontsize_authors = _fontSize(18)
fontsize_noinfo = _fontSize(18) -- you forgot that one
```
The last one is another you need to update, in the `else -- bookinfo not found`, it's the one displayed with the filename and an added "..." while waiting for the cover to be fetched. It's currently too large when large number of items.

Anyway, I'm fine with the look now, but you're probably not :)
So, if you want to introduce another setting, `Zoom font : -2 -1 0(default) 1 2 3 4 5 6` (or "Font size shift", dunno) , your `_fontSize()` would use that saved setting to tweak the formula.
Probably one of these 2 solutions (not sure which is best, probably the 1st for consistency).
```lua
local font_size = math.ceil( (nominal+font_zoom) * dimen.h / 64)
local font_size = math.ceil(nominal * dimen.h / 64) + font_zoom
```

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -44,6 +44,9 @@ local BookInfoManager = require("bookinfomanager")
 local corner_mark_size = -1
 local corner_mark
 
+local scale_by_size
+local font_scale_factor
```
poire-z: (No more needed)

plugins/coverbrowser.koplugin/listmenu.lua
derungs: and I'll cap the font size at something like Screen:height / 36.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: I'll check if this function triggers when files_per_page changes. If yes, then I'll go with your suggestion. The extra `zoom font` setting may not be necessary.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: Font sizes are now almost identical with original - one px off in some cases, same size in most. Without explicitly setting `files_per_page`, most users should not see any difference.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: Decided not to move font calculation to `ListMenu:_recalculateDimen()` because of its dependency on `ListMenuItem.underline_h`. I feel that apart from limited readability, it can introduce hard-to-track visual glitches when someone changes `ListMenuItem`.

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -502,6 +493,9 @@ function ListMenuItem:update()
                         authors = { authors[1], T(_("%1 et al."), authors[2]) }
                     end
                     authors = table.concat(authors, "\n")
+                    -- Needs to fit 3 lines instead of 2
+                    fontsize_title = _fontSize(17)
+                    fontsize_authors = _fontSize(15)
```
poire-z: 3 lines of authors or 3 lines in total (1 line of title + 2 line of authors) ?
Because we already could display 2 line of authors, and I don't really see the need to display 3:

<kbd>image</kbd>

Or is it that you prefer to decrease the font size in that case to have more blank space around? (I felt you initially disliked that :)
Dunno about how you prefer it with series, but in my screenshot case, I think I prefer it as on the screenshot.

edit: tested, and I don't see anything changing because of that: I guess we'd already go to some smaller font size than the one you set here when having to display 1+2 lines.
So why did you have to do that? Some particular case I didn't manage to test? Or is it just to start decreasing font size from a lower start value to avoid some loops?

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -189,11 +189,10 @@ function ListMenuItem:update()
...
 
     local function _fontSize(nominal)
---        return math.ceil(nominal * font_scale_factor * scale_by_size
-        -- dimen.h * scaleBySize, font_size / scaleBySize cancel each other out
-        local font_size = math.ceil(nominal * dimen.h / 64)
-        logger.warn("font size", nominal, dimen.h, font_size)
-        return font_size
+        -- nominal font size is based on 64px ListMenuItem height
+        -- keep ratio of font size to item height
+        local font_size = nominal * dimen.h / 64 / scale_by_size
```
poire-z: OK :) (On the emulator, I didn't realized it was missing the scaleBySize inversion :)

derungs: That's because scaleBySize in the emulator (720/540) has no effect. On 1200/900 the difference would be more than 10%.

plugins/coverbrowser.koplugin/listmenu.lua
derungs: When displaying title and 2 authors, or title, series, and author, the loop would run 3 times, ending with title font size 17 and author font size 15. So I decided to not cause the 3 loops and go to the right font size immediately.
Thanks to your input to this PR, we can always compute spacing on a 64px height, and all scaling to the right sizes will be applied afterwards.

plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -237,7 +249,8 @@ function BookInfoManager:getSetting(key)
     return self.settings[key]
 end
 
-function BookInfoManager:saveSetting(key, value)
+function BookInfoManager:saveSetting(key, value, reload)
+    reload = reload or true
```
poire-z: Haven't noticed it before :)
If you provide `reload=false`, you'll end up with `reload = false or true`, so with `true`.
(The right way is: `if reload == nil then reload = true end` instead.)

If you haven't noticed any thing strange since then :) I suggest we drop that reload parameter, and always end up with loadSettings, as before, to keep things simple.
(As you'd use it only when migrating, yes, you'd do 2 or 3 unneeded saves, but that's a one time thing when migrating the old setting.)

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -245,7 +255,17 @@ function ListMenuItem:update()
 
         local bookinfo = BookInfoManager:getBookInfo(self.filepath, self.do_cover_image)
         if bookinfo and self.do_cover_image and not bookinfo.ignore_cover then
-            if not bookinfo.cover_fetched then
+            if bookinfo.cover_fetched then
+                -- trigger recalculation of thumbnail if size changed
+                if bookinfo.has_cover and bookinfo.cover_sizetag ~= "M" and bookinfo.cover_sizetag ~= cover_specs.sizetag then
+                    logger.dbg("ListMenuItem: thumbnail size changed from "
+                        .. (bookinfo.cover_sizetag or "nil") .. " to " .. (cover_specs.sizetag or "nil"))
```
poire-z: `logger.dbg("ListMenuItem: thumbnail size changed from ", bookinfo.cover_sizetag, " to ", cover_specs.sizetag)`
simpler and avoid string concatenation (on devices where we don't log debug).

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: You might want to make that comment `-- Needs to fit 3 lines instead of 2` more explicite then, like `-- as we'll fit 3 lines instead of 2, we can avoid some loops by starting from a lower font size`

plugins/coverbrowser.koplugin/bookinfomanager.lua
derungs: Totally missed the `reload=false` implication :(
If we reload settings on every time, the order in which we reload can impact the correctness of the translation. So I'd rather go with no translation at all, and the old settings will be kept lying around.

poire-z:
I would also drop a few of your `logger.dbg()` (those that give you the sizes that helped during dev might not be needed for others to see now that you got these sizes right :)

Apart from my last nitpicks, looks good to me. Anything else you wish to add? (We'll merge it after next release, so you get a few more days :)

You really don't wish anymore to have a bigger font size? How is that?

(I think it could be nice, just to have a reason to push that `(Detailed list) Files per page` into a sub menu `(Detailed list) Items and font > Files per page` and `> Font size` and have more room below to see the effect - but really not forcing you on that if you no more find it's needed.)

plugins/coverbrowser.koplugin/bookinfomanager.lua
poire-z: Oh, right, indeed.
Well, I'm fine with both options: no migration, or just fixing your `reload = false or true`, so do as you wish

derungs:
The one dbg log I want to keep is when triggering another loop for size calculation.

Nothing more to add from my side.

I don't want to push for bigger font size now, because that functionality is provided by selecting fewer files per page.

poire-z:
Tested, and looks good to me (I'll test it on my Kobo later).

Some small aesthetic issue when using a large number of items par page (so, small font sizes), that is not to be fixed by this PR, is that the spacing between the 2 items on the right is a bit large, larger than what we get on the left:
image

By adding `padding = 1` to the `wfileinfo = TextWidget:new` and `wpageinfo = TextWidget:new`, we get a better display:
image
with `padding=0`:
<kbd>image</kbd>

So, that padding (internal to TextWidget) should be a factor of the font size.
So not to be fixed in CoverBrowser, and I guess I have an answer to my question :):
https://github.com/koreader/koreader/blob/17244cca2f0d9fa8d9cb22068244f79f224f8306/frontend/ui/widget/textwidget.lua#L23-L28

(The left stuff uses TextBoxWidget, which doesn't add any padding - so dunno if we should use it too for the right stuff instead.)
Anyway, I'll look at it after this PR is merged.

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -612,7 +646,7 @@ function ListMenuItem:update()
                     HorizontalSpan:new{ width = Screen:scaleBySize(10) },
                     TextBoxWidget:new{
                         text = self.text .. hint,
-                        face = Font:getFace("cfont", 18),
+                        face = Font:getFace("cfont", _fontSize(18)),
                         width = dimen.w - 2 * Screen:scaleBySize(10),
```
poire-z: With a small number of items per page (4), with long filenames, this one may overflow the dimen.h, and draw on the next item height (which is just a bit ugly), or for the last item, it may draw on the footer, which won't get refreshed (which is uglyer :)
I was going to suggest to just add here `height = dimen.h` - but then the text is top aligned and no more centered.
So, I guess we'll have to go with a `while true ... if height < dimen.h then break else fontsize = fontsize -1` like we do for the authors/titles widgets.
(I can do it later, if you have enough :)

poire-z:
Looking and working really good on my Kobo :+1:
I think I won't even bother fixing the little alignment issue mentionned above, they show only for extreme values (4,5, 18 to 20), but only if you pay attention - and my padding stuff mentionned above affects the lest stuff, so it might not be an easy fix.

I see now why you don't need a zoom font setting :)
With a small number of items (4 to 6), with the large cover and large text info on the right, the width available for authors/titles is quite shorter, so you quite early (90% of my books) swith to 3 lines, (and so the font will be decreased to make 3 lines - and it's already quite fat - so a zoom font won't help at all:)
(With my still prefered value of 10 items per page, only 10% of my books go on 3 lines.)

plugins/coverbrowser.koplugin/listmenu.lua
derungs: No worries, I'll do it - with a `repeat ... until` because I prefer it for aesthetic reasons :)

derungs:
> the width available for authors/titles is quite shorter

Do you think we should limit the max size of the file info further?

plugins/coverbrowser.koplugin/listmenu.lua
```diff
@@ -640,17 +640,25 @@ function ListMenuItem:update()
             if self.file_deleted then -- unless file was deleted (can happen with History)
                 hint = " " .. _("(deleted)")
             end
+            local text_widget
+            local fontsize_deleted = _fontSize(18)
```
poire-z: It's not (only) for items deleted.
It's fot item with no bookinfo (yet) in DB - and it's probably gonna be fetched in a few seconds.
So better name this `fontsize_no_bookinfo`, or just `fontsize` as it's just a local variable.

derungs: It was the only way I could produce this situation, hence the wrong assumption.

poire-z:
> Do you think we should limit the max size of the file info further?

Well, I don't know, no real opinion about what's best when large items - _you_ are the customer there :)
But I don't think reducing the width of the right stuff will really add/help much.
(I have the feeling the starting font is already really too large, but may be that's what "large items" users expect. Otherwise, you could start from/cap the font size for authors and titles to a smaller value, so you'd fit more text into the limited width.)

I noticed that browsing with 4 items per page is quite slow on my GloHD - not sure what's causing that slowness, the bigger cover images (hoping they are sized correctly when saved and there's no additional resize on drawing), or the font size decrease loop, or else...
But I don't mind all that that much, as it exhibits only a 4-6 - users will see, won't like it much, and will use a value of 7 :)
(A quick & dirty & ugly solution is to hide all that by having a min value of 6 or 7 :)

derungs:
Let's leave the file info size cap as is, then. Honestly, I don't expect anyone to use 4 or 5 here, only went so far because Mosaic shows 3 rows.

plugins/coverbrowser.koplugin/listmenu.lua
poire-z: Well, you can produce it by just changing the "File per page" value :) Then as you browse, the cover_sizetag being different, bookinfo is dropped, and you're in a "no bookinfo" situation. That's the temporary display while the covers and metadata are fetched.
(If you don't see them, you must have a super fast device! :)

NiLuJe:
I'd still look into the resize thing @poire-z mentioned, in case there's actually a real performance issue that's just swept under the rug by tamer resizes ;).

poire-z:
Some little logging of browsing 2 pages with 4 items per page on my Glo HD:
```
os.date()           os.clock()
# 10/20/19-23:45:57 4.090  no scaling
# 10/20/19-23:45:57 4.090  178 299 -> 178 299 1    cover imag size -> widget size scale_factor
# 10/20/19-23:45:57 4.100  font loop 1 52 47    logger.warn("font loop", num, fontsize_title, fontsize_authors)
# 10/20/19-23:45:57 4.120  font loop 2 51 46
# 10/20/19-23:45:57 4.150  font loop 3 50 45
# 10/20/19-23:45:57 4.170  font loop 4 49 44
# 10/20/19-23:45:57 4.190  font loop 5 48 43
# 10/20/19-23:45:57 4.200  font loop 6 47 42
# 10/20/19-23:45:57 4.210  font loop 7 46 41
# 10/20/19-23:45:57 4.220  font loop 8 45 40
# 10/20/19-23:45:57 4.240  font loop 9 44 39
# 10/20/19-23:45:57 4.250  font loop 10 43 38
# 10/20/19-23:45:57 4.260  font loop 11 42 37
# 10/20/19-23:45:57 4.270  font loop 12 41 36
# 10/20/19-23:45:57 4.280  font loop 13 40 35
# 10/20/19-23:45:57 4.290  font loop 14 39 34
# 10/20/19-23:45:57 4.300  font loop 15 38 33
# 10/20/19-23:45:57 4.320  font loop 16 37 32
# 10/20/19-23:45:57 4.340  font loop 17 36 31
# 10/20/19-23:45:57 4.360  font loop 18 35 30
# 10/20/19-23:45:57 4.370  font loop 19 34 29
# 10/20/19-23:45:57 4.380  font loop 20 33 28
# 10/20/19-23:45:57 4.390  font loop 21 32 27
# 10/20/19-23:45:57 4.410  font loop 22 31 26
# 10/20/19-23:45:57 4.430  font loop 23 30 25
# 10/20/19-23:45:57 4.440  font loop 24 29 24
# 10/20/19-23:45:57 4.460  font loop 25 28 23
# 10/20/19-23:45:57 4.470  font loop 26 27 22
# 10/20/19-23:45:57 4.490  no scaling
# 10/20/19-23:45:57 4.490  178 299 -> 178 299 1
# 10/20/19-23:45:57 4.490  font loop 1 52 47
# 10/20/19-23:45:57 4.510  font loop 2 51 46
# 10/20/19-23:45:57 4.530  font loop 3 50 45
# 10/20/19-23:45:57 4.550  font loop 4 49 44
# 10/20/19-23:45:57 4.570  font loop 5 48 43
# 10/20/19-23:45:57 4.590  font loop 6 47 42
# 10/20/19-23:45:57 4.600  font loop 7 46 41
# 10/20/19-23:45:57 4.620  font loop 8 45 40
# 10/20/19-23:45:58 4.640  font loop 9 44 39
# 10/20/19-23:45:58 4.650  font loop 10 43 38
# 10/20/19-23:45:58 4.670  font loop 11 42 37
# 10/20/19-23:45:58 4.690  font loop 12 41 36
# 10/20/19-23:45:58 4.700  font loop 13 40 35
# 10/20/19-23:45:58 4.720  font loop 14 39 34
# 10/20/19-23:45:58 4.730  font loop 15 38 33
# 10/20/19-23:45:58 4.750  font loop 16 37 32
# 10/20/19-23:45:58 4.760  font loop 17 36 31
# 10/20/19-23:45:58 4.770  font loop 18 35 30
# 10/20/19-23:45:58 4.790  font loop 19 34 29
# 10/20/19-23:45:58 4.810  no scaling
# 10/20/19-23:45:58 4.810  183 299 -> 183 299 1
# 10/20/19-23:45:58 4.810  font loop 1 52 47
# 10/20/19-23:45:58 4.820  font loop 2 51 46
# 10/20/19-23:45:58 4.830  font loop 3 50 45
# 10/20/19-23:45:58 4.840  font loop 4 49 44
# 10/20/19-23:45:58 4.850  font loop 5 48 43
# 10/20/19-23:45:58 4.860  font loop 6 47 42
# 10/20/19-23:45:58 4.870  font loop 7 46 41
# 10/20/19-23:45:58 4.880  font loop 8 45 40
# 10/20/19-23:45:58 4.880  font loop 9 44 39
# 10/20/19-23:45:58 4.890  font loop 10 43 38
# 10/20/19-23:45:58 4.900  font loop 11 42 37
# 10/20/19-23:45:58 4.910  font loop 12 41 36
# 10/20/19-23:45:58 4.920  font loop 13 40 35
# 10/20/19-23:45:58 4.920  font loop 14 39 34
# 10/20/19-23:45:58 4.930  font loop 15 38 33
# 10/20/19-23:45:58 4.940  font loop 16 37 32
# 10/20/19-23:45:58 4.940  font loop 17 36 31
# 10/20/19-23:45:58 4.950  font loop 18 35 30
# 10/20/19-23:45:58 4.970  no scaling
# 10/20/19-23:45:58 4.970  196 299 -> 196 299 1
# 10/20/19-23:45:58 4.970  font loop 1 52 47
# 10/20/19-23:45:58 4.980  font loop 2 51 46
# 10/20/19-23:45:58 4.990  font loop 3 50 45
# 10/20/19-23:45:58 5.000  font loop 4 49 44
# 10/20/19-23:45:58 5.010  font loop 5 48 43
# 10/20/19-23:45:58 5.020  font loop 6 47 42
# 10/20/19-23:45:58 5.030  font loop 7 46 41
# 10/20/19-23:45:58 5.040  font loop 8 45 40
# 10/20/19-23:45:58 5.050  font loop 9 44 39
# 10/20/19-23:45:58 5.060  font loop 10 43 38
# 10/20/19-23:45:58 5.080  font loop 11 42 37
# 10/20/19-23:45:58 5.090  font loop 12 41 36
# 10/20/19-23:45:58 5.100  font loop 13 40 35

# 10/20/19-23:46:12 5.570  no scaling
# 10/20/19-23:46:12 5.570  185 299 -> 185 299 1
# 10/20/19-23:46:12 5.570  font loop 1 52 47
# 10/20/19-23:46:12 5.580  font loop 2 51 46
# 10/20/19-23:46:12 5.620  font loop 3 50 45
# 10/20/19-23:46:12 5.640  font loop 4 49 44
# 10/20/19-23:46:12 5.640  font loop 5 48 43
# 10/20/19-23:46:12 5.660  font loop 6 47 42
# 10/20/19-23:46:12 5.660  font loop 7 46 41
# 10/20/19-23:46:12 5.680  font loop 8 45 40
# 10/20/19-23:46:12 5.690  font loop 9 44 39
# 10/20/19-23:46:12 5.710  no scaling
# 10/20/19-23:46:12 5.710  206 299 -> 206 299 1
# 10/20/19-23:46:12 5.710  font loop 1 52 47
# 10/20/19-23:46:12 5.730  font loop 2 51 46
# 10/20/19-23:46:12 5.740  font loop 3 50 45
# 10/20/19-23:46:12 5.760  font loop 4 49 44
# 10/20/19-23:46:12 5.780  font loop 5 48 43
# 10/20/19-23:46:12 5.800  font loop 6 47 42
# 10/20/19-23:46:12 5.820  font loop 7 46 41
# 10/20/19-23:46:12 5.840  font loop 8 45 40
# 10/20/19-23:46:12 5.850  font loop 9 44 39
# 10/20/19-23:46:12 5.870  font loop 10 43 38
# 10/20/19-23:46:12 5.880  font loop 11 42 37
# 10/20/19-23:46:12 5.900  font loop 12 41 36
# 10/20/19-23:46:12 5.910  font loop 13 40 35
# 10/20/19-23:46:12 5.930  font loop 14 39 34
# 10/20/19-23:46:12 5.950  font loop 15 38 33
# 10/20/19-23:46:12 5.960  font loop 16 37 32
# 10/20/19-23:46:12 5.980  font loop 17 36 31
# 10/20/19-23:46:12 6.000  font loop 18 35 30
# 10/20/19-23:46:12 6.010  font loop 19 34 29
# 10/20/19-23:46:12 6.030  font loop 20 33 28
# 10/20/19-23:46:12 6.050  font loop 21 32 27
# 10/20/19-23:46:12 6.060  font loop 22 31 26
# 10/20/19-23:46:12 6.070  font loop 23 30 25
# 10/20/19-23:46:13 6.090  font loop 24 29 24
# 10/20/19-23:46:13 6.110  font loop 25 28 23
# 10/20/19-23:46:13 6.130  font loop 26 27 22
# 10/20/19-23:46:13 6.160  no scaling
# 10/20/19-23:46:13 6.160  206 299 -> 206 299 1
# 10/20/19-23:46:13 6.160  font loop 1 52 47
# 10/20/19-23:46:13 6.180  font loop 2 51 46
# 10/20/19-23:46:13 6.210  font loop 3 50 45
# 10/20/19-23:46:13 6.230  font loop 4 49 44
# 10/20/19-23:46:13 6.240  font loop 5 48 43
# 10/20/19-23:46:13 6.250  font loop 6 47 42
# 10/20/19-23:46:13 6.270  font loop 7 46 41
# 10/20/19-23:46:13 6.280  font loop 8 45 40
# 10/20/19-23:46:13 6.290  font loop 9 44 39
# 10/20/19-23:46:13 6.300  font loop 10 43 38
# 10/20/19-23:46:13 6.310  font loop 11 42 37
# 10/20/19-23:46:13 6.320  font loop 12 41 36
# 10/20/19-23:46:13 6.340  font loop 13 40 35
# 10/20/19-23:46:13 6.350  font loop 14 39 34
# 10/20/19-23:46:13 6.360  font loop 15 38 33
# 10/20/19-23:46:13 6.370  font loop 16 37 32
# 10/20/19-23:46:13 6.380  font loop 17 36 31
# 10/20/19-23:46:13 6.390  font loop 18 35 30
# 10/20/19-23:46:13 6.410  no scaling
# 10/20/19-23:46:13 6.410  206 299 -> 206 299 1
# 10/20/19-23:46:13 6.410  font loop 1 52 47
# 10/20/19-23:46:13 6.420  font loop 2 51 46
# 10/20/19-23:46:13 6.420  font loop 3 50 45
# 10/20/19-23:46:13 6.430  font loop 4 49 44
# 10/20/19-23:46:13 6.440  font loop 5 48 43
# 10/20/19-23:46:13 6.440  font loop 6 47 42
# 10/20/19-23:46:13 6.450  font loop 7 46 41

# 10/20/19-23:46:18 6.830  no dialog left to show
# 10/20/19-23:46:18 6.830  quitting uimanager
```

So, good, no issue with the cover image, it's just right and doesn't need scaling.
But the font size decrease loop is the one that takes time, and it's really variable depending on the title and authors length vs the short width.
So, except for adjusting the starting font size (even for small text, `loop 7 46 41` looks really big - `loop 18 35 30` would feel like a good starting point).
Dunno if there's some logic error in the font size computation to start from such a large size, or if it's doomed to happen with the constrained width with large items...
Or let's just not bother about these extrem cases (and let people who really want them fix it :)

poire-z:
OK, going to merge this PR: it looks fine, the cover images are finely made, saved and loaded without resize.
We can work on optimizing the _find the best font size_ algo later.

Idea:
```lua
local fitWithBestFontSize = function(max_height, font_size_max, font_size_min, building_func)
    -- be clever, use dichotomy
end

local widget1, widget2
fitWithBestFontSize(available_height, font_size_start, font_size_min, function(font_size)
   -- do the widget building
   widget1 = TextBoxWidget:new(...)
   widget2 = TextBoxWidget:new(...)
   return widget1:getHeight() + widget2:getHeight()
end
```

poire-z:
Other idea: just step by a larger number instead of 1 with big items (so, function of item height).

poire-z:
Sorry for that other commit (`Update listmenu.lua`), dunno how it ended up individually in master, thought it would be squashed with the others here... (and it's supposed to apply to the next commit...)

poire-z:
Damned, what did I do ?! I did not squashed all that ?! :|

Frenzie:
Looks like it got squashed into https://github.com/koreader/koreader/commit/2034001a5affc975864b165588b256709c3f593c?

poire-z:
Yes, but there are 20 commits mixed with all the others over the last 24 days :|

poire-z:
Should I reset with force push ?

poire-z:
or rebase master by dropping all the intermediate, and then force push?

Frenzie:
Oh right. I guess you must've used a classic merge.

> Should I reset with force push ?

Eh, it's not ideal but it could be worse. I'm find with leaving it as it is.

poire-z:
Not possible to just drop that path in red?
image

Frenzie:
I'm not really sure how well rebasing merged branches goes. If you can fix it locally I'm also fine with a force push, but please be very careful. ;-)

If you temporarily rename the protected branch from master to something else in https://github.com/koreader/koreader/settings/branches then you can force push master.

poire-z:
I cleaned that up and pushed it to my "origin" https://github.com/poire-z/koreader/commits/master
with `git push -f origin master`.
So, `git push -f upstream master` should work once unprotected?

Frenzie:
Yes, indeed.

poire-z:
Won't lose tags or stuff specific to koreader/master?

Frenzie:
Hm, I don't think so but just to be sure perhaps you'd best do something like git fetch --tags upstream master first.

poire-z:
OK, done. Well, let's try it...

poire-z:
Done. Looks ok on the github commit page.
Except for the ref to the PR for the latest commit.

Frenzie:
You could add it real quick while you're at it. :-P

But it doesn't matter.

poire-z:
Well, stopping playing with fire for now :)
Everything looking ok for you too?
(Did you do something for gitlab resync last weekend? It had updated to my mess in the meantime.)

Frenzie:
Yeah, I'll have to force push the changes to GitLab but I'll take care of that tomorrow.

It looks good as far as I can tell. :-)

poire-z:
OK :) Thanks for the help :) phew...

poire-z:
Mhhh, managed to push a commit to upstream/master. Can you check if the protection has not lost some settings with the move?

Frenzie:
The protection is against *force* pushing. Normal everyday pushing isn't presumed dangerous.

poire-z:
Oh, ok. Confirmed that I could not force push a drop of that commit :) So, I guess that's ok.

poire-z:
2 things noticed after having updated on my Android phone:
- my default was and still was 12 items, but when hitting "Files per page", it said 10.
- browsing pages has become slower, even with 10 or 12 - so probably the font size decrease as described above (might be because of phones A/R, high height, small width, we soon hit the case of a really small width to draw title/authors described above, so many font size loops trying to fit all that in that small width).

poire-z:
Going to fix what I can, including #5523.
(I'm off tomorrow for about a week, and won't be able to test/code, and hardly review - and I'd rather go with a fixed koreader :)

poire-z:
Pffff. Previously, with a hardcomputed item height for Detailed list, I would get 10 items in History in portrait mode, and 7 items in History in landscape mode - which felt visually nice and coherent.
With this files_per_page setting, we'll then and have to get 10 items in both - things will look smaller in landscape mode then.


Dunno if we should just let that as that (that setting won't lie) - or compute a dedicated perpage when in landscape mode from the aspect ratio, so we get 7 in landscape when that setting is 10 (as people anyway can only play with that setting, and see its effect, when in file browser which is always in portrait mode.)

Frenzie:
Like you said, it sounds visually incoherent.

PR #5436, opened 2019-09-29
EPUB: more support for RTL documents

poire-z:
bump crengine: enhanced RTL block rendering. Includes:
- Re-enable: Use the same FT load flags in HB as in FT https://github.com/koreader/crengine/pull/311
- Avoid redundant calls to getStyle() https://github.com/koreader/crengine/pull/312
- CSS: parse and store "direction:"
- Enhanced block rendering: handle RTL direction
- Enhanced block rendering: handle RTL tables
- Pagesplitting: account for lines' text direction
- CSS: add support for more pseudo-classes
- epub.css: a few updates for RTL documents
- Fix LVDocView::getBookmark() which could be slow or wrong https://github.com/koreader/crengine/pull/313

Add a few style tweaks useful to RTL readers.

Wikipedia Save as EPUB: build proper RTL documents from persian, arabic, hebrew (and others) Wikipedia articles.

Screenshots in https://github.com/koreader/crengine/pull/312

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -169,12 +169,12 @@ h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !importan
             {
                 title = _("About text direction"),
                 info_text = _([[
-Documents in languages like arabic or hebrew are to be rendered Right-To-Left. This does not only affects text layout, but also list items bullets and numbers (put on the right) and table layout (cells laid from right to left).
+Languages like Arabic or Hebrew use right-to-left writing systems (Right-To-Left or RTL). This doesn't only affect text layout, but also list item bullets and numbers, which have to be put on the right side of the page, as well as tables, where the cells are laid out from right to left.
 
-Usually, the publisher has set the appropriate tags to force such rendering - but if he hasn't, or if you're reading plain text documents, you may want to manually enable it with these tweaks.
-Note that in the absence of such specifications KOReader will try to detect the language of each paragraph and set the approppriate rendering per paragraph.
+Usually, the publisher will have set the appropriate tags to enable RTL rendering. But if these are missing, or if you're reading plain text documents, you may want to manually enable RTL with these tweaks.
+Note that in the absence of such specifications, KOReader will try to detect the language of each paragraph to set the appropriate rendering per paragraph.
 
-You may also want to enable, in top menu → Gear → Navigation →, Invert page turn taps and swipes.]]),
+You may also want to enable, in the top menu → Gear → Navigation →, Invert page turn taps and swipes.]]),
                 separator = true,
             },
             {
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -179,7 +179,7 @@ You may also want to enable, in the top menu → Gear → Navigation →, Invert
             },
             {
                 id = "body_direction_rtl",
-                title = _("Set document direction to RTL (Right-to-Left)"),
+                title = _("Document direction RTL"),
                 css = [[body { direction: rtl !important; }]],
                 priority = 2, -- so it overrides the LTR one below
             },
@@ -201,7 +201,7 @@ You may also want to enable, in the top menu → Gear → Navigation →, Invert
             },
             {
                 id = "body_direction_ltr",
-                title = _("Set document direction to LTR (Left-to-Right)"),
+                title = _("Document direction LTR"),
                 css = [[body { direction: ltr !important; }]],
             },
         },
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -178,7 +178,7 @@ You may also want to enable, in top menu → Gear → Navigation →, Invert pag
                 separator = true,
             },
             {
-                id = "dir_html_rtl",
+                id = "body_direction_rtl",
                 title = _("Set document direction to RTL (Right-to-Left)"),
                 css = [[body { direction: rtl !important; }]],
                 priority = 2, -- so it overrides the LTR one below
@@ -200,7 +200,7 @@ You may also want to enable, in top menu → Gear → Navigation →, Invert pag
                 separator = true,
             },
             {
-                id = "dir_html_ltr",
+                id = "body_direction_ltr",
                 title = _("Set document direction to LTR (Left-to-Right)"),
                 css = [[body { direction: ltr !important; }]],
             },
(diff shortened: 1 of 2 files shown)


frontend/ui/data/css_tweaks.lua
```diff
@@ -161,7 +161,48 @@ h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !importan
...
-                priority = 3, -- so it overrides the ones above
+                priority = 4, -- so it overrides the ones above
+            },
+        },
+        {
+            title = _("Text direction"),
+            {
+                title = _("About text direction"),
+                info_text = _([[
+Documents in languages like arabic or hebrew are to be rendered Right-To-Left. This does not only affects text layout, but also list items bullets and numbers (put on the right) and table layout (cells laid from right to left).
```
Frenzie: Maybe something like this? (In any case, Arabic and Hebrew are capitalized in English. ;-))
```suggestion
Languages like Arabic or Hebrew use right-to-left writing systems (Right-To-Left or RTL). This doesn't only affect text layout, but also list item bullets and numbers, which have to be put on the right side of the page, as well as and tables, where the cells are laid out from right to left.
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -161,7 +161,48 @@ h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !importan
...
+
+Usually, the publisher has set the appropriate tags to force such rendering - but if he hasn't, or if you're reading plain text documents, you may want to manually enable it with these tweaks.
```
Frenzie:
```suggestion
Usually, the publisher will have set the appropriate tags to enable RTL rendering. But if these are missing, or if you're reading plain text documents, you may want to manually enable RTL with these tweaks.
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -161,7 +161,48 @@ h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !importan
...
+Note that in the absence of such specifications KOReader will try to detect the language of each paragraph and set the approppriate rendering per paragraph.
```
Frenzie:
```suggestion
Note that in the absence of such specifications, KOReader will try to detect the language of each paragraph to set the appropriate rendering per paragraph.
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -161,7 +161,48 @@ h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !importan
...
+
+You may also want to enable, in top menu → Gear → Navigation →, Invert page turn taps and swipes.]]),
```
Frenzie:
```suggestion
You may also want to enable, in the top menu → Gear → Navigation →, Invert page turn taps and swipes.]]),
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -161,7 +161,48 @@ h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !importan
...
+                separator = true,
+            },
+            {
+                id = "body_direction_rtl",
+                title = _("Set document direction to RTL (Right-to-Left)"),
```
Frenzie: Seems a bit wordy. Couldn't we suffice with something like this?
```suggestion
                title = _("Force RTL (Right-to-Left)"),
```

Or even (given the explanation item above):

```suggestion
                title = _("Force RTL"),
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -161,7 +161,48 @@ h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !importan
...
+                id = "text_align_all_right",
+                title = _("Right align all elements"),
+                description = _("Enforce right alignment of text in all elements."),
+                css = [[* { text-align: right !important; }]],
+                priority = 3, -- so it overrides the ones from Text alignment
+                separator = true,
+            },
+            {
+                id = "body_direction_ltr",
+                title = _("Set document direction to LTR (Left-to-Right)"),
```
Frenzie: Same comment as above.

frontend/ui/wikipedia.lua
```diff
@@ -655,6 +655,31 @@ local ext_to_mimetype = {
     woff = "application/font-woff",
 }
 
+-- Display from these Wikipedia should be RTL
```
Frenzie: Don't be afraid to use multiline comments. ;-)

frontend/ui/wikipedia.lua
```diff
@@ -916,19 +941,19 @@ h1, h2 {
...
     page-break-inside: avoid;
     page-break-after: avoid;
-    text-align: left;
+    text-align: start;
 }
 /* Styles for our centered titles on first page */
 h1.koreaderwikifrontpage, h5.koreaderwikifrontpage {
     page-break-before: avoid;
     text-align: center;
-    margin-top: 0em;
```
Frenzie: Huh, I somehow overlooked these needless units?

frontend/ui/data/css_tweaks.lua
poire-z: Well, these tweaks do not "force" RTL or LTR. They just set it to the top node. It will be overriden if there are dir= below in the document.
I have added no tweak that "force" a direction (I guess they would need `* {direction: rtl !important;}` - and I don't think they would be useful).

Also, after all this, I don't quite like the wording "text direction", because, as we see, the text itself does not change direction (english stays Left-To-Right). "Document direction" is a bit less confusing, although it could give the idea it means "invert page turn taps and swipes" - that's why I added that little help text.
So, I don't know :) other suggestions ?

frontend/ui/wikipedia.lua
poire-z: Remind me what are these and how they differ from consecutive `-- blah` ?! :)

frontend/ui/data/css_tweaks.lua
Frenzie: How about just "RTL"? Or `Document direction RTL` (it doesn't need to be a sentence).

frontend/ui/wikipedia.lua
Frenzie: I just mean 
```
--[[--lua
Bah blah.

Blah.
--]]
```

poire-z:
Small (or not so small) issue noticed on the crengine side.
Because of [this change](https://github.com/koreader/crengine/commit/5967f723b8937fcac30739732d61db8d7af0eee7#diff-77f6525e33d2fd1aeeb073a64dd8d759R13444-R13460) to get selection working in table cells, the getBookmark() call used to get the last_xpointer for each page can, on some pages (the pages with notes on that Petra.AR.epub), be slow (1-2 second lag in page turn when on some pages on my Kobo), or be wrong (same pages when in scroll mode, we'd always get back to the page with note 1 when we were on some other page from that list than spans 4 or 5 pages).
So, let's wait for another crengine bump before merging this...

frontend/ui/wikipedia.lua
poire-z: Well, I find these uglier :)

frontend/ui/data/css_tweaks.lua
poire-z: That's fine :) Done.

Issue #5408, opened 2019-09-21
annotations not exported


milmike0:
* KOReader version: Starling
* Device: PW2, firmware 5.4.3.2

Hi.
 Thanks for all your work on koreader. It's great.

After I highlight  a few sentences of a PDF, I will occasionally edit it and add a (usually moronic) comment. These comments are visible in the bookmarks but not in the HTML file created by Evernote. This seems odd.

When I look at the metadata.pdf.lua I can see these comments under the section ["test"] but not under the section ["notes']. I'm not a programmer so I don't really know what this means.

I have three questions.

1. Is this behavior peculiar to Kindle devices?
2. If I wanted to copy all notes could I do it directly from the Bookmarks folder when connected to my PC? Where is  this folder and can I open it? 
3. Would it be a PITA to make Koreader copy all notes directly from the bookmarks  or from the ["test'] section to the Evernote HTML file? (Again I'm not a programmer, so have a good laugh at this one if you want).

Thanks thousands.

Frenzie:
> Is this behavior peculiar to Kindle devices?

Unlikely.

> If I wanted to copy all notes could I do it directly from the Bookmarks folder when connected to my PC? Where is this folder and can I open it?

There'll be a `.sdr` folder with the same name as the document. That contains all the data related to the document. Also see https://github.com/noembryo/KoHighlights/

> Would it be a PITA to make Koreader copy all notes directly from the bookmarks or from the ["test'] section to the Evernote HTML file? (Again I'm not a programmer, so have a good laugh at this one if you want).

Probably not but I don't use Evernote.

milmike0:
Thank you.

Just a note that this is really the same problem as noted by maximus83 on his kobo on 1-26-19.

Frenzie:
Do you mean this is a duplicate with an existing issue?

milmike0:
Yes. I had tried to research it  before starting a new thread but could not find anything. When I noticed that you added the plugin label it gave me the hint to look under "Evernote plugin" where I found the thread by maximus83.

Frenzie:
Ah, you mean [this thread](https://www.mobileread.com/forums/showthread.php?t=314688&highlight=Evernote+plugin) on the forum. Thanks.

There's currently some movement in #5086, so perhaps @kereviz would be interested in looking at basic shared functionality after Joplin support is finished.

milmike0:
[...]

PR #5440, opened 2019-09-30
Support for exporting annotations/bookmark renames 

mustafa292:
see https://github.com/koreader/koreader/issues/5408

parseHighlight now must be called with both highlight and bookmarks fields of sidecar file. If bookmarks field have edited text field on it, it removes page and time info from text, saves edited version as clipping.text. 

I tried to making it work only with bookmarks or highlight but both fields lacks necessary data (pboxes, page info for example) and didn't want to fill them with duplicate info.

Also when a highlights borders are changed bookmarks.text field gets rewritten. I think there will also be issues with that later.

### DIFF
--- a/plugins/evernote.koplugin/clip.lua
+++ b/plugins/evernote.koplugin/clip.lua
@@ -226,7 +226,7 @@ function MyClipping:getImage(image)
     end
 end
 
-function MyClipping:parseHighlight(highlights, book)
+function MyClipping:parseHighlight(highlights, bookmarks, book)
     --DEBUG("book", book.file)
     for page, items in pairs(highlights) do
         for _, item in ipairs(items) do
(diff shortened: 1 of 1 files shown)


plugins/evernote.koplugin/clip.lua
```diff
@@ -281,7 +287,7 @@ function MyClipping:parseHistoryFile(clippings, history_file, doc_file)
             title = title,
             author = author,
         }
-        self:parseHighlight(stored.highlight, clippings[title])
+        self:parseHighlight(stored.highlight, stored.bookmarks, clippings[title])
```
Frenzie:
```suggestion
        self:parseHighlight(stored.highlight, stored.bookmarks, clippings.title)
```

plugins/evernote.koplugin/clip.lua
```diff
@@ -312,7 +318,7 @@ function MyClipping:parseCurrentDoc(view)
         title = title,
         author = author,
     }
-    self:parseHighlight(view.highlight.saved, clippings[title])
+    self:parseHighlight(view.highlight.saved, view.ui.bookmark.bookmarks, clippings[title])
```
Frenzie:
```suggestion
    self:parseHighlight(view.highlight.saved, view.ui.bookmark.bookmarks, clippings.title)
```

PR #5442, opened 2019-10-01
Show a progress bar at the bottom of the screen

robert00s:
Similar like in new firmware of Kobo


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -944,6 +960,27 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                    if self.settings.progress_bar_separate_line then
+                        self.settings.progress_bar_bottom = nil
+                    end
+                    self:refreshFooter()
+                    self.ui:handleEvent(Event:new("SetPageBottomMargin", self.view.document.configurable.b_page_margin))
+                    UIManager:setDirty(nil, "ui")
+                end,
+            },
+            {
+                text = _("Progress bar on the bottom"),
```
```suggestion
                text = _("Bottom progress bar"),
```

robert00s:
> Looks cool. What's the reasoning behind avoiding the corners?

No special reason :) I can extend progress bar to the corners.

NiLuJe:
I'd vote *for* avoiding the corners, specifically with Android devices with rounded corners in mind (i.e., most current phones) ;).

Frenzie:
> (i.e., most current phones) ;).

You mean the "waterfall"?

poire-z:
No comment on the code, except 2 things:

You made the LineWidget progress'able - which was indeed quite easy - but it's quite a change in meaning :) I just wonder if for clarity of purpose, it wouldn't be clearer to have a `lineprogresswidget.lua`, that would just inherit from `LineWidget` (or not - it looks like we can't use `style=dashed` with the new progress option) and just implement `LineProgressWidget:paintTo()`. @Frenzie?

Also, do we really want to _just_ do as the new kobo firmware (which is a nice thing). But could some people want the real progress bar with ticks at bottom? Or this new thin line above the footer items (dunno how that would look :) And how would a progressbar without ticks with a smaller height look vs this line widget? So, I dunno, menu wise, it could be a styling option of the progress bar (top|bottom|aside, tick|notick thick|thin).
Just sharing some thoughts, not requesting that nor stopping this from being merged :)

No real opinion on avoiding corners - I see the issue with phones with rounded corners - but are we currently drawing the regular footer in these hidden corners? Or do we add some black on the bottom of such phones? And on regular squared devices, this smaller width could feel arbitrary.

Frenzie:
> clearer to have a lineprogresswidget.lua

Or something like a ProgressWidget type=line? Or for that matter just a regular ProgressWidget with bordersize 0?

> on regular squared devices, this smaller width could feel arbitrary.

Hence my earlier comment. It just looked a bit odd.

NiLuJe:
> > (i.e., most current phones) ;).
> 
> You mean the "waterfall"?

Oh, those, too, probably, yeah. But I was thinking of just plain rounded corners ;).

AlanSP1:
I would respect margin of text, or something along this line... It looks nicer and somewhat more logical, not to go from end to end. Similar to text.

NiLuJe:
Could it simply inherit the CRe margin values? *that* would make sense to me.

Frenzie:
Can't say it does to me I'm afraid. At the edge of the screen that just looks like random empty space.

Speaking of margin, there's a couple of weird margin pixels at at the bottom.

NiLuJe:
I, err, also like those ? :D (What's happening to me, i'm advocating for whitespace?! :D).

robert00s:
proposed menu

poire-z:
My tick/notick and thin/thick were just suggestions for toggle states. So, I guess you can still keep the `Show chapter markers` checkbox instead of 2 items `tick` & `notick`.
Same for thick/thin: just a toggle: either `Thick`, or `Thin` (or a better naming?) - and I dunno if you could have ticks/markers with the thin - so may be some enable/disable relationship with the others.
For top/bottom/aside, either as shown, or in a submenu `Position: aside >` (like our `Sort by: filename>` ).

But all that only if you find some pleasure hacking readerfooter again :) don't feel obliged!

NiLuJe:
To expand on what @poire-z said, I'd propose `Simple` for the "thin/line" style, and I don't think it ought to support chapter ticks, as that kind of defeats the simplistic design.

Which makes the menu simpler, as the chapter markers checkbox can just be greyed out if the chosen style is "simple" ;).

AlanSP1:
Well, I would also ask for one more option, if it is possible, to have it respect margins, or from edge to edge.

On last image it is from edge to edge and I just watched it on my older version where it respects text margins and to me it looks great, much, much more appealing than edge-to-edge version.

Otherwise, it is really great how one of the most visible elements of Koreader has become customized. It makes status bar really users' domain, each of us can make it to suit his/hers own tastes. Great work really. Top notch.

Frenzie:
> I don't think it ought to support chapter ticks, as that kind of defeats the simplistic design.

I dunno, I think the point isn't to be simplistic but to be thin. I think it could potentially work quite well. A quick demo implementation would be nice to see. ;-)

robert00s:
Version 2
- removed LineWidget changes,
- adapted ProgressWidget to show thin bar,
- ability to choose progress bar width (edge to edge, balanced, like book margins),
- ability to choose progress bar height (from 1 to 4 px for thin bar style and from 5 to 14 for thick bar),
- removed DMINIBAR_HEIGHT - now is set up by bar height in Progress bar style,
- ability to set up bar position (top, bottom, aside)

Thanks all for suggestions :)

Some screenshots:

AlanSP1:
Excellent work as always... top notch.

defaults.lua
```diff
@@ -179,9 +179,10 @@ DCREREADER_PROGRESS_BAR = 1
 
 -- configure "mini" progress bar
 DMINIBAR_TOC_MARKER_WIDTH = 2   -- Looses usefulness > 3
-DMINIBAR_HEIGHT = 7             -- Should be smaller than DMINIBAR_CONTAINER_HEIGHT
 DMINIBAR_CONTAINER_HEIGHT = 14  -- Larger means more padding at the bottom, at the risk of eating into the last line
 DMINIBAR_FONT_SIZE = 14
+-- no more needed
```
NiLuJe:
```suggestion
-- no longer needed
```

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -913,41 +942,232 @@ function ReaderFooter:addToMainMenu(menu_items)
...
                     end
                     UIManager:setDirty(nil, "ui")
                 end,
             },
             {
-                text = _("Show chapter markers"),
-                checked_func = function()
-                    return self.settings.toc_markers
+                text_func = function()
+                    local pos = _("aside")
```
NiLuJe: I find the whole "aside" thing highly cryptic.

I'm still not quite sure what it means in this context. Is that supposed to mean how it's always been (i.e., next to the text/icon stuff?).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -913,41 +942,232 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            self:setTocMarkers()
+                            self:updateFooter()
+                            UIManager:setDirty(nil, "ui")
+                        end
+                    },
+                },
+            },
+            {
+                text_func = function()
+                    local pos = _("balanced (10)")
```
NiLuJe: I would *tend* to prefer "centered" over "balanced", but let's wait for @Frenzie for a comment on that ;).

NiLuJe:
Besides making the whole "aside" thing clearer, :+1:, thanks! ;).

frontend/apps/reader/modules/readerfooter.lua
poire-z: That's indeed what I meant (on the same line with the text/icon stuff) when suggesting it with my bad english (well, _aside_ sounded like a snobbish version of "on the side" to me :)
What's the correct word?

NiLuJe: Yeah, that's kind of my point: I took it to mean show me a *vertical* bar on one side of the screen ;).

frontend/apps/reader/modules/readerfooter.lua
Frenzie: What exactly does it mean?

frontend/apps/reader/modules/readerfooter.lua
Frenzie: Text alongside/next to icons?

NiLuJe: I'd somewhat take a different approach, and use something based on the *variable* names, actually.

i.e., "Top", "Bottom", and "Bottom, on its own".

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: With a fixed L+R margin.

Frenzie: Centered sounds like the right term for that, yes.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -913,41 +942,232 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                    if self.settings.progress_margin_width == 0 then
+                        pos = _("no margins (0)")
+                    end
+                    if self.settings.progress_margin then
+                        local margins = self.ui.document:getPageMargins()
+                        pos = T(_("like book margins (%1)"), math.floor((margins.left + margins.right)/2))
```
Frenzie: same as book margins?

frontend/apps/reader/modules/readerfooter.lua
poire-z: Above footer items / Alongside footer items / Below footer items ?

frontend/apps/reader/modules/readerfooter.lua
poire-z: But "like book margins" is centered too.
This "Balanced" item means centered, but with fixed margins, with some default fixed margins values that do please most people (so, statistically balanced :)

Frenzie: Static margins?

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: Oh, wait, the "Top" means above the items, not the top *of the screen* (again :D).

Then, yeah, what you just said ;p.

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: I wasn't as bothered by this one, and the only other idea I had was something like "honor book margins" or "follow" book margins, and I'm not quite convinced that's any better ;).

Frenzie: It's okay, but I think it's less clear than it could be. It can be read as any book. (So about 2 cm perhaps? Who knows. ;-) )

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1266,6 +1494,14 @@ function ReaderFooter:onChangeScreenMode()
     self:resetLayout(true)
 end
 
+function ReaderFooter:onSetPageHorizMargins(h_margins)
+    if self.settings.progress_margin then
+        self.settings.progress_margin_width = math.floor((h_margins[1] + h_margins[2])/2)
```
poire-z: Oh, (following my previous comment) I'm just seeing that. So does that mean it's updated on any book load? :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -913,41 +942,232 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                        callback = function()
+                            local value, value_min, value_max
+                            if self.settings.progress_style_thin then
+                                value = self.settings.progress_style_thin_height or 3
+                                value_min = 1
+                                value_max = 4
+                            else
+                                value = self.settings.progress_style_thick_height or 7
+                                value_min = 5
+                                value_max = 14
```
poire-z: I kinda like our current progress bar height. After I'll play with that setting, may be I'll want to get back the original height - which I guess is 7 (?). Might be useful having a `Use default` button in that spin widget?

yparitcher:
thanks @robert00s for all the work on the footer (and UI in general).

Is it possible while doing this to allow the progressbar to be sorted with the rest of the status bar items, when next to them.

instead of only
```
| progressbar          | pages | time | 
|----------------------|-------|------|
```
also allow
```
| pages | progressbar          | time | 
|-------|----------------------|------|
```
thanks.

robert00s:
@yparitcher It should be possible. I added to my todo list but I don't plan do it before 2019.12.

robert00s:
@poire-z @Frenzie @NiLuJe Thanks for the review. I hope all suggestion were made :)

Frenzie:
Sounds fine to me.

NiLuJe:
Yep, LGTM.

(Not quite sure what's going on with the testsuite in the CI, though?).

NiLuJe:
Ah, my most favorite test ever. :D.

Frenzie:
It's to dissuade us from adding more junk to the file. ;-D

frontend/ui/widget/progresswidget.lua
```diff
@@ -92,4 +92,27 @@ function ProgressWidget:setPercentage(percentage)
     self.percentage = percentage
 end
 
+function ProgressWidget:updateStyle(thick, height)
+    if thick then
+        if height then
+            self.height = height
+        end
+        self.margin_h = Screen:scaleBySize(3)
```
poire-z: You call this `self.progress_bar:updateStyle(true, Screen:scaleBySize(bar_height))`.
As you already do many `Screen:scaleBySize` here, you might as well do that for the provided `height`, so it's a bit simpler at the calling places:
`self.progress_bar:updateStyle(true, bar_height)`.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -912,42 +930,216 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                        checked_func = function()
+                            return self.settings.progress_margin and not self.ui.document.info.has_pages
+                        end,
+                        enabled_func = function()
+                            return not self.ui.document.info.has_pages
+                        end,
+                        callback = function()
+                            self.settings.progress_margin = true
+                            local margins = self.ui.document:getPageMargins()
+                            self.settings.progress_margin_width = Screen:scaleBySize(math.floor((margins.left + margins.right)/2))
```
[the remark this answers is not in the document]
NiLuJe: Yeah, CRe feeds us the scaled margins (i.e., pixels).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -912,42 +930,216 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            self:refreshFooter(nil, true)
+                        end,
+                    },
+                    {
+                        text_func = function()
+                            if self.ui.document.info.has_pages then
+                                return _("Same as book margins")
+                            end
+                            local margins = self.ui.document:getPageMargins()
+                            return T(_("Same as book margins (%1)"), math.floor((margins.left + margins.right)/2))
```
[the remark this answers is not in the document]
robert00s: Maybe better show in the menu scaled value

```lua
                        text_func = function()
                            return T(_("Static margins (%1)"), Screen:scaleBySize(10))
                        end,
```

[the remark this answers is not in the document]
NiLuJe: Whatever you end up with, if it's a scaled value, that fact should be made explicit (with a px suffix), like we did in the feedback popup when setting the CRe margins ;).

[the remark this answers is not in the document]
poire-z: Better, I don't think so (we display font size unscaled, and other things too), but simpler possibly :)
Or may be just don't fetch them from crengine.
The only place that seems to feed them to crengine is:
https://github.com/koreader/koreader/blob/b75fec4971c1b2fa46e9a69cf72c2275fc10d35f/frontend/apps/reader/modules/readertypeset.lua#L502-L512
And this function always get them unscaled.

It's a bit unclear if this event is always sent (it's sometimes called directly, so the duplication mentionned below - and this function `return true`, possibly  stopping propagation).
But we have ReaderDogear (that is loaded by ReaderView, just like ReaderFooter is), that may get the event before ReaderTypeset, that already does react to margins - so, you may do as ReaderDogear does in ReaderFooter:
https://github.com/koreader/koreader/blob/b75fec4971c1b2fa46e9a69cf72c2275fc10d35f/frontend/apps/reader/modules/readerdogear.lua#L43-L58
https://github.com/koreader/koreader/blob/b75fec4971c1b2fa46e9a69cf72c2275fc10d35f/frontend/apps/reader/modules/readerdogear.lua#L60-L73
(That @NiLuJe code will then be triplicated - but I won't mention it as it may get complicated to do differently :)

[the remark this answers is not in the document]
poire-z: So, the idea is that instead of fetching them each time, you would store them in `self`.
Read their initial value in onReadSettings, store them.
Get the new value when it's updated, by reacting to `onSetPageMargins` (you already do it for `onSetPageHorizMargins`, but `ReaderTypeset:onSetPageHorizMargins` will send `self.ui:handleEvent(Event:new("SetPageMargins", self.unscaled_margins)` - so no need for your `onSetPageHorizMargins `then).

robert00s:
@poire-z Done :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -690,7 +698,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                         end,
                         callback = function()
                             self.settings.align = "center"
-                            self:refreshFooter(nil, true)
+                            self:refreshFooter(true, false)
```
poire-z: You can drop the rightmost arguments when they are `false` (the function will get `nil` instead, and yours deals with `nil` the same as `false`) and it makes it a bit shorter here:
`self:refreshFooter(true)`

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1343,6 +1348,16 @@ function ReaderFooter:onReaderReady()
     self:updateFooter()
 end
 
+function ReaderFooter:onReadSettings(config)
+    if not self.ui.document.info.has_pages then
+        local h_margins = config:readSetting("copt_h_page_margins") or
+            G_reader_settings:readSetting("copt_h_page_margins") or
+            DCREREADER_CONFIG_H_MARGIN_SIZES_MEDIUM
+        self.book_margin_left = h_margins[1]
+        self.book_margin_right = h_margins[2]
```
poire-z: So, that's all that was needed !? :) (I expected it would need much much work, so... well done :)

Given that you compute `math.floor((self.book_margin_left + self.book_margin_right)/2))` a few times above, you may as well compute it here, and just store the result as `self.book_margins_footer_width`.
And the same below in `ReaderFooter:onSetPageHorizMargins`.
And you would then use where appropriate:
`self.settings.progress_margin_width = Screen:scaleBySize(self.book_margins_footer_width)`
`return T(_("Same as book margins (%1)"), self.book_margins_footer_width)`

poire-z:
(Just noticing in the test output some warnings related to my crengine loading progress stuff - there were there in previous successful tests, so they are not the cause of your 3 real readerfooter errors.)

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: I kind of like the fact that every call is explicit and consistent for these kind of tricks, it potentially saves a roundtrip of checking the signature to grok why some do it this way and others another ;).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -912,42 +932,213 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                        callback = function()
+                            self.settings.progress_style_thin = true
+                            local bar_height = self.settings.progress_style_thin_height or PROGRESS_BAR_STYLE_THIN_DEFAULT_HEIGHT
+                            self.progress_bar:updateStyle(false, bar_height)
+                            self:refreshFooter(true, true)
+                        end,
+                    },
+                    {
+                        text = _("Set size"),
+                        separator = true,
```
poire-z: This menu deserves to have added `keep_menu_open = true,` - so you can quickly get back after you press OK to try another value :)
(And may be move `separator=true` at the end of the item, after the `callback=` - as it looks more like how it will appear, below the stuff :) and it's how we do everywhere I think.

poire-z:
Just a small thing: I had the feeling (more on the emulator than on my Kobo) that the bottom margin was bigger, and I had a hard time to get my text nearer to the footer/progress bar.

The self.view.footer:getHeight() we get in:
https://github.com/koreader/koreader/blob/99495aaed43bb770804b46f2a186c471c79808ee/frontend/apps/reader/modules/readertypeset.lua#L510
is now a bit greater, for no visual difference in the position of the footer:

Before this PR: self.view.footer:getHeight() => 14
<kbd>image</kbd>

With this PR: self.view.footer:getHeight() => 16
<kbd>image</kbd>

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -341,8 +346,20 @@ function ReaderFooter:updateFooterContainer()
...
+    if self.settings.progress_bar_position == "above" and not self.settings.disable_progress_bar then
+        table.insert(self.vertical_frame, self.progress_bar)
+        table.insert(self.vertical_frame, vertical_span)
+        table.insert(self.vertical_frame, self.footer_container)
+    elseif self.settings.progress_bar_position == "below" and not self.settings.disable_progress_bar then
+        table.insert(self.vertical_frame, self.footer_container)
+        table.insert(self.vertical_frame, vertical_span)
+        table.insert(self.vertical_frame, self.progress_bar)
+    else
+        table.insert(self.vertical_frame, vertical_span)
```
poire-z: Could it be because of this vertical_span? which wasn't added before?

NiLuJe: That might track with the extra blank pixel line @Frenzie originally mentioned (one on top, one on the bottom)?

poire-z:
Also, when using "Alongside items", the Margin menu is disabled.
But it looks exactly as if it was using "Static margins (10)". So, may be we could also apply the margins when using Alongside items (as the "Static margins" entry gives the exact legacy width).

frontend/apps/reader/modules/readerfooter.lua
poire-z: No pb for me with the space below the text.
Which might not be any additional space: note the bar/pipe separator character in my screenshots above, that goes touching the bottom of the screen. So, that space is just the normal space below the baseline for our footer font. Don't reduce it ! :)

NiLuJe: True. Plus, that was with the "thin" bar, not the classic one ;).

robert00s: Yes, I added this line (extra vertical space). Restore as it was before?

poire-z: Yes, I'd prefer it as before, as currently you add a non-cancellable bottom margin (so, when you set the bottom margin to 0, you wouldn't get 0 in the end) - and one can always add it if he wants it by using a bigger bottom margin.

robert00s:
> Also, when using "Alongside items", the Margin menu is disabled.
But it looks exactly as if it was using "Static margins (10)". So, may be we could also apply the margins when using Alongside items (as the "Static margins" entry gives the exact legacy width).

I can enable this option for "Alongside items" but without `same as book margins`. For now this option with wide margins will generate problem to property show all elements with many items. In the future I will need to create truncate items when text is too long.

For example in this situation:

frontend/apps/reader/modules/readerfooter.lua
robert00s: Ok, I will do :)

poire-z:
> For now this option with wide margins will generate problem to property show all elements with many items.

Well, you could propose that option, but you're not obliged to garantee it will always be used.
If you're able to figure out the "minimum needed width", and if the "book margins width" is smaller than it, well, just forget the book margins width and use your "minimum needed width".
Or if you're able to just figure out that items don't fit, fallback to Static margins or 0-margins.

But I'm fine too if you just let things as they are :)

PR #5443, opened 2019-10-01
[wallabag] fetch pdf files directly instead of converting them

teroshan:
Currently, all the articles were converted to .epub by default. When
handling an article that linked to a .pdf file, it resulted in an
unreadable file.

This patch skips the conversion for a pdf file, and download it directly
instead.

I'm using the mimetype attribute of wallabag's articles to determine if they are pdf or not. Unfortunately this doesn't seem to cover every link correctly for some reason (e.g. https://www.usenix.org/system/files/sec19-sathaye.pdf is not recognized as a pdf by wallabag). Maybe looking at the extension of the link is better?

Since some pdfs were a little bit large, I had to extend the timeout, but I still got some errors reported at the end of the sync sometimes. How could I access the errors reporter to the logger?

Finally I reused the callApi function, although not for its initial intended purpose. But it had some error handling, so hey :)

### DIFF
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -329,9 +329,18 @@ end
 -- @treturn int 1 failed, 2 skipped, 3 downloaded
 function Wallabag:download(article)
     local skip_article = false
-    local item_url = "/api/entries/" .. article.id .. "/export.epub"
     local title = util.getSafeFilename(article.title, self.directory, 230, 0)
-    local local_path = self.directory .. article_id_prefix .. article.id .. article_id_postfix .. title .. ".epub"
+    local file_ext = ".epub"
+    local item_url = "/api/entries/" .. article.id .. "/export.epub"
+
+    -- If the article links to a pdf file, we will download it directly
+    ---- @todo use hasProvider to skip all supported mimetypes
+    if article.mimetype == "application/pdf" then
+        file_ext = ".pdf"
+        item_url = article.url
+    end
+
+    local local_path = self.directory .. article_id_prefix .. article.id .. article_id_postfix .. title .. file_ext
     logger.dbg("Wallabag: DOWNLOAD: id: ", article.id)
     logger.dbg("Wallabag: DOWNLOAD: title: ", article.title)
     logger.dbg("Wallabag: DOWNLOAD: filename: ", local_path)
@@ -364,31 +373,47 @@ function Wallabag:download(article)
 end
 
 -- method: (mandatory) GET, POST, DELETE, PATCH, etc...
--- apiurl: (mandatory) excluding the server path
+-- apiurl: (mandatory) API call excluding the server path, or full URL to a file
 -- headers: defaults to auth if given nil value, provide all headers necessary if in use
 -- body: empty string if not needed
 -- filepath: downloads the file if provided, returns JSON otherwise
+---- @todo separate call to internal API from the download on external server
 function Wallabag:callAPI( method, apiurl, headers, body, filepath )
     local request, sink = {}, {}
-    local parsed = url.parse(self.server_url)
-    request["url"] = self.server_url .. apiurl
+    local parsed
+
+    -- Is it an API call, or a regular file direct download?
+    if apiurl:sub(1, 1) == "/" then
+        -- API call to our server, has the form "/random/api/call"
+        parsed = url.parse(self.server_url)
+        request["url"] = self.server_url .. apiurl
+        if headers == nil then
+            headers = { ["Authorization"] = "Bearer " .. self.access_token, }
+        end
+    else
+        -- regular url link to a foreign server
+        local file_url = apiurl
+        parsed = url.parse(file_url)
+        request["url"] = file_url
+        if headers == nil then
+            headers = {} -- no need for a token here
+        end
+    end
+
     request["method"] = method
     if filepath ~= "" then
         request["sink"] = ltn12.sink.file(io.open(filepath, "w"))
     else
         request["sink"] = ltn12.sink.table(sink)
     end
-    if headers == nil then
-        headers = { ["Authorization"] = "Bearer " .. self.access_token, }
-    end
     request["headers"] = headers
     if body ~= "" then
         request["source"] = ltn12.source.string(body)
     end
-    logger.dbg("Wallabag: URL     ", self.server_url .. apiurl)
+    logger.dbg("Wallabag: URL     ", request["url"])
     logger.dbg("Wallabag: method  ", method)
 
-    http.TIMEOUT, https.TIMEOUT = 10, 10
+    http.TIMEOUT, https.TIMEOUT = 30, 30
     local httpRequest = parsed.scheme == "http" and http.request or https.request
     local code, resp_headers = socket.skip(1, httpRequest(request))
     -- raise error message when network is unavailable
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -108,7 +108,7 @@ function Wallabag:addToMainMenu(menu_items)
                         NetworkMgr:promptWifiOn()
                         return
                     end
-                    local num_deleted = self:processLocalFiles( "manual" )
+                    local num_deleted = self:processLocalFiles("manual")
                     UIManager:show(InfoMessage:new{
                         text = T(_("Articles processed.\nDeleted: %1"), num_deleted)
                     })
@@ -125,9 +125,9 @@ function Wallabag:addToMainMenu(menu_items)
                         self.ui:onClose()
                     end
                     if FileManager.instance then
-                        FileManager.instance:reinit( self.directory )
+                        FileManager.instance:reinit(self.directory)
                     else
-                        FileManager:showFiles( self.directory )
+                        FileManager:showFiles(self.directory)
                     end
                 end,
             },
@@ -275,7 +275,7 @@ function Wallabag:getBearerToken()
         })
         return false
     end
-    if string.sub( self.directory, -1 ) ~= "/" then
+    if string.sub(self.directory, -1) ~= "/" then
         self.directory = self.directory .. "/"
     end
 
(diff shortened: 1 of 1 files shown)


plugins/wallabag.koplugin/main.lua
```diff
@@ -364,14 +372,24 @@ function Wallabag:download(article)
 end
 
 -- method: (mandatory) GET, POST, DELETE, PATCH, etc...
--- apiurl: (mandatory) excluding the server path
+-- apiurl: (mandatory) API call excluding the server path, or full URL to a file
 -- headers: defaults to auth if given nil value, provide all headers necessary if in use
 -- body: empty string if not needed
 -- filepath: downloads the file if provided, returns JSON otherwise
-function Wallabag:callAPI( method, apiurl, headers, body, filepath )
+function Wallabag:callAPI( method, apiurl, headers, body, filepath)
```
Frenzie: Either do one or the other. ;-) (Our code prefers no spaces.)
```suggestion
function Wallabag:callAPI(method, apiurl, headers, body, filepath)
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -364,14 +372,24 @@ function Wallabag:download(article)
...
     local request, sink = {}, {}
-    local parsed = url.parse(self.server_url)
-    request["url"] = self.server_url .. apiurl
+    -- Is it an API call, or a regular file direct download?
+    if apiurl:sub(1, 1) == "/" then
+        -- API call, is of the form "/random/api/call", append the server url
+        server = self.server_url
+        parsed = url.parse(self.server_url)
+        request["url"] = self.server_url .. apiurl
```
```suggestion
        request.url = self.server_url .. apiurl
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -364,14 +372,24 @@ function Wallabag:download(article)
...
+    else
+        -- regular url link
+        local file_url = apiurl
+        parsed = url.parse(file_url)
+        request["url"] = file_url
```
```suggestion
        request.url = file_url
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -364,14 +372,24 @@ function Wallabag:download(article)
...
+    end
     request["method"] = method
```
```suggestion
    request.method = method
```

Frenzie:
> Since some pdfs were a little bit large, I had to extend the timeout, but I still got some errors reported at the end of the sync sometimes. How could I access the errors reporter to the logger?

The code already comes with a couple of commented out debug statements iirc. There doesn't seem to be any logging in the else path; just add some? Not entirely sure I understand the question. :-)

https://github.com/koreader/koreader/blob/27de8d9bdf088791558d44513df9d4033cb10e15/plugins/wallabag.koplugin/main.lua#L420-L432

> e.g. https://www.usenix.org/system/files/sec19-sathaye.pdf is not recognized as a pdf by wallabag

It's not? It seems to have an appropriate mimetype.

```shell
$ curl -I https://www.usenix.org/system/files/sec19-sathaye.pdf
HTTP/2 200 
cache-control: public, max-age=75441
content-language: en
content-type: application/pdf
etag: ea3a36f2f0ba240442fd8d420fca5042
expires: Wed, 02 Oct 2019 17:19:20 GMT
last-modified: Tue, 30 Jul 2019 18:33:12 GMT
server: nginx
x-content-type-options: nosniff
x-drupal-cache: MISS
x-frame-options: SAMEORIGIN
x-pantheon-styx-hostname: styx-fe4-9b9c45564-mtjgs
x-styx-req-id: 1f341b4a-e489-11e9-aa3f-6ac4c6779cab
x-timestamp: 1564511591.49125
x-trans-id: tx5bfe77e3f40344e3a1be7-005d8a8e60ord1
date: Tue, 01 Oct 2019 20:22:00 GMT
x-served-by: cache-mdw17322-MDW, cache-ams21043-AMS
x-cache: MISS, MISS
x-cache-hits: 0, 0
x-timer: S1569961320.611933,VS0,VE590
age: 0
accept-ranges: bytes
via: 1.1 varnish
content-length: 3700505
```

I'm not a fan of doing things by extension on the Internet tbh.

Frenzie:
I forgot to say something positive. I like the idea *and* the execution really. Just some nits. ;-)

teroshan:
> It's not? It seems to have an appropriate mimetype.

My server reads the same thing, and the wallabag logs also seem to indicate that "application/pdf" was read. I'll have to investigate on wallabag's side I guess.

> The code already comes with a couple of commented out debug statements iirc. There doesn't seem to be any logging in the else path; just add some? Not entirely sure I understand the question. :-)

Sorry I wasn't clear, I don't see where the logger actually outputs the text. I found a `crash.log` file, but it isn't there. I'm running the script directly on my kobo, maybe I should use the emulator for that?

> That could be resolved by doing more with the if/else paths but I'm inclined to think DRY doesn't really apply here.

Damn, didn't see the authorization header. Right, I'll write another function for directly download through the link of an article instead. Thanks for pointing that out.

Thank you for your remarks, I'll work through them.
Also thank you  CI-bot, it made me understand how `local` works in Lua :)

teroshan:
By the way, how does it work with pull requests, can I just add commits on my branch and rebase at the end to keep a clean log, or will you do that at the end when merging?

plugins/wallabag.koplugin/main.lua
teroshan: Hmm, so I understand that it's easier to read that way, but since the rest of the file uses this same syntax should I replace every other occurrence of this pattern as well? Wouldn't it be weird if I only changed the lines you pointed out?

NiLuJe:
@teroshan: It's highly unlikely anything else will conflict with what you're doing for the lifetime of this PR, so you can basically pile on commits as you see fit ;).

We'll generally squash to merge, or in some rarer cases rebase, so basically, only rebase yourself if you absolutely need to (i.e., if GH complains about conflicts ;)).

plugins/wallabag.koplugin/main.lua
```diff
@@ -376,34 +376,39 @@ end
 -- headers: defaults to auth if given nil value, provide all headers necessary if in use
 -- body: empty string if not needed
 -- filepath: downloads the file if provided, returns JSON otherwise
-function Wallabag:callAPI( method, apiurl, headers, body, filepath)
+function Wallabag:callAPI( method, apiurl, headers, body, filepath )
```
teroshan: Same thing here, the convention in the file is using spaces. If the convention of the project is to not use them, is it ok for me to go over the whole file? Should i make another PR for this specific thing? Or is it ok to make another commit right here

teroshan:
So I removed the header which was leaking the token, I don't see any other reason to rewrite the function. It basically crafts the request, and downloads a file, with some error handling/reporting at the end.

plugins/wallabag.koplugin/main.lua
Frenzie: It improves commit log/git blame clarity to do it separately, so probably better to do it as a follow-up. Or just as two structured commits in this PR if you know how.

plugins/wallabag.koplugin/main.lua
```diff
@@ -329,9 +329,17 @@ end
 -- @treturn int 1 failed, 2 skipped, 3 downloaded
 function Wallabag:download(article)
     local skip_article = false
-    local item_url = "/api/entries/" .. article.id .. "/export.epub"
     local title = util.getSafeFilename(article.title, self.directory, 230, 0)
-    local local_path = self.directory .. article_id_prefix .. article.id .. article_id_postfix .. title .. ".epub"
+    local file_ext = ".epub"
+    local item_url = "/api/entries/" .. article.id .. "/export.epub"
+
+    -- If the article links to a pdf file, we will download it directly
```
Frenzie: Btw, does how does Wallabag deal with other supported filetypes? Most notably DjVu, EPUB, FB2 & CBZ. Not necessarily for this PR, but it might be worth considering doing something with http://koreader.rocks/doc/modules/document.documentregistry.html#DocumentRegistry:hasProvider or the like if Wallabag works as a transport.

Frenzie:
> Sorry I wasn't clear, I don't see where the logger actually outputs the text. I found a crash.log file, but it isn't there. I'm running the script directly on my kobo, maybe I should use the emulator for that?

You need to run the program with the `-d` flag, which you can also enable in `Developer options` (in the tools menu in the file manager, might be on page 2).

Note that you don't need to build the emulator to play around with pure Lua like you're doing. You can get at the content of the AppImage with `--appimage-extract`. Then you can run `reader.lua` directly. See [here](https://github.com/koreader/koreader/blob/27de8d9bdf088791558d44513df9d4033cb10e15/doc/Building.md#setting-up-a-build-environment-for-koreader) for more information on available options.

> So I removed the header which was leaking the token, I don't see any other reason to rewrite the function. It basically crafts the request, and downloads a file, with some error handling/reporting at the end.

I'm inclined to think it could be clearer if refactored slightly but it's acceptable. Note that I'm not talking about clearer in the context of this PR but clearer in a year (or even in a month) just reading through the file.

plugins/wallabag.koplugin/main.lua
teroshan: Wallabag-converted EPUB files also get outputted as unreadable mess, it would totally make sense to skip the conversion for all files that have a provider.

I tried using the "hasProvider" family of functions to do what I want, but I was stuck on the fact that all of them use `filepath` as input. It's easy enough to add a dict to track the supported mimetypes, but I also have to find the extension to add to my downloaded file.

That extension would need to be the one associated to the preferred provider, but `getProvider` need the extension... I'll try to tackle that later.

Maybe an acceptable hack for the meantime would be to add some more hardcoded extensions/mimetypes for the types above?

teroshan:
The developper setting in the menu gets me everything I need, `tail -f crash.log | grep -i wallabag` and I see everything. Thank you for the instructions.

> I'm inclined to think it could be clearer if refactored slightly but it's acceptable. Note that I'm not talking about clearer in the context of this PR but clearer in a year (or even in a month) just reading through the file.

I see what you mean, I'll add a todo in the code if that's okay with you. I have a few other things I wanted to add to the plugin, I may refactor it then.

plugins/wallabag.koplugin/main.lua
teroshan: I made a second commit that removes the extra whitespaces, and replaces the `foo["bar"]` to `foo.bar`

plugins/wallabag.koplugin/main.lua
Frenzie: Sidenote for future refinement, Wallabag can grab files from websites that require authentication (see [here](https://wallabag.org/en/news/new-release-wallabag-2-3-is-there-really)). I'm not sure if this is something the API mentions so it can immediately be taken into account because I've personally never tried it, but it might be something to keep in mind.

Frenzie:
Thanks!

Issue #5444, opened 2019-10-02
[Proposal] Image viewer use the whole screen when opened


max-cw:
It's always bothered me that when opening a full-page picture in imageviewer widget, it is shown in a smaller size because the imageviewer use a significant vertical space for the buttons on the bottom. I made some simple changes to the imageviewer.lua as followed:

- When opened, the buttons are not drawn, and the picture take up the whole screen space.
- If the user taps the bottom quarter of the screen, the buttons are toggled.
- If the user taps the rest of the screen, the widget is closed.

I've been using this modified image viewer for a long time now and I think it works well. I would like to ask if the team thinks this is a reasonable idea. The code I have is based on the old version of koreader, but it should be easy to merge the change into the latest one.

NiLuJe:
That certainly sounds sensible to me, feel free to fire a PR ;).

robert00s:
:+1: 
I was also thinking about something similar.

poire-z:
I kinda like the current display, I don't think the buttons height ever bothered me.
But no opposition to this proposal if people prefer that - as the hidden buttons would be easy to display, and not to hard to discover (if stuck, the user would just tap anywhere).
But you have to keep the current state (buttons shown/not shown) when some other gestures are done (zoomin/panning...)
Just don't forget the top caption that should still be shown by default if there is one (from images in Wikipedia lookup results) - and the rare case of animated gifs, where tap on left/right should still move among frames - with a little progress bar at bottom (that should be displayed even if the buttons are not - it's only with it that one can know it's an animated gif and there's more stuff to be seen).
(So that you don't spend time looking for some sample, here's an EPUB with such an animated gif  - a map of Europe in grey and pink on the first pages: [SaintEmpireRomainGermanique.FR.epub.zip](https://github.com/koreader/koreader/files/3682461/SaintEmpireRomainGermanique.FR.epub.zip))

max-cw:
Thanks for the comments. And especially thanks @poire-z for the detailed one! Indeed, the animated gifs is something I did not notice before. I'm happy to report that I just re-applied my modification on the master and with minimum changes, I manage to make everything works! Everything you said is work as before. When it detected a list of images, I switched back to showing button by default, and disabled the button toggle/close onTap.

I'll fire a PR soon. Just need to figure out how. This will be my first PR on GitHub 😉(please be gentle!)

PR #5446, opened 2019-10-02
Image viewer use the whole screen

max-cw:
As proposed in #5444, when open an image, it is shown using the whole screen. User can toggle the buttons by tapping the bottom quarter. Tapping other area will close the image.

If it is a list of image (e.g. animated gifs), then the above does not applied.

### DIFF
--- a/frontend/ui/widget/imageviewer.lua
+++ b/frontend/ui/widget/imageviewer.lua
@@ -390,7 +390,9 @@ function ImageViewer:update()
     if progress_container then
         table.insert(frame_elements, progress_container)
     end
-    table.insert(frame_elements, button_container)
+    if self._show_buttons then
+        table.insert(frame_elements, button_container)
+    end
 
     self.main_frame = FrameContainer:new{
         radius = not self.fullscreen and 8 or nil,
@@ -455,6 +457,7 @@ function ImageViewer:onTap(_, ges)
     if self.caption_tap_area and ges.pos:intersectWith(self.caption_tap_area.dimen) then
         self.caption_visible = not self.caption_visible
         self:update()
+        return true
     end
     if self._images_list then
         -- If it's a list of image (e.g. animated gifs), tap each size to navigate
--- a/frontend/ui/widget/imageviewer.lua
+++ b/frontend/ui/widget/imageviewer.lua
@@ -81,6 +81,9 @@ local ImageViewer = InputContainer:new{
     _image_wg = nil,
     _images_list = nil,
     _images_list_disposable = nil,
+    
+    -- buttons state
+    _show_buttons = false,
 }
 
 function ImageViewer:init()
@@ -131,6 +134,7 @@ function ImageViewer:init()
         -- also swap disposable status
         self._images_list_disposable = self.image_disposable
         self.image_disposable = self._images_list.image_disposable
+        self._show_buttons = true
     end
     self:update()
 end
@@ -165,52 +169,57 @@ function ImageViewer:update()
         self.width = Screen:getWidth() - Screen:scaleBySize(40)
     end
 
-    local buttons = {
-        {
+    local button_table_size = 0
+    local button_container
+    if self._show_buttons then
+        local buttons = {
             {
-                text = self._scale_to_fit and _("Original size") or _("Scale"),
-                callback = function()
-                    self.scale_factor = self._scale_to_fit and 1 or 0
-                    self._scale_to_fit = not self._scale_to_fit
-                    -- Reset center ratio (may have been modified if some panning was done)
-                    self._center_x_ratio = 0.5
-                    self._center_y_ratio = 0.5
-                    self:update()
-                end,
+                {
+                    text = self._scale_to_fit and _("Original size") or _("Scale"),
+                    callback = function()
+                        self.scale_factor = self._scale_to_fit and 1 or 0
+                        self._scale_to_fit = not self._scale_to_fit
+                        -- Reset center ratio (may have been modified if some panning was done)
+                        self._center_x_ratio = 0.5
+                        self._center_y_ratio = 0.5
+                        self:update()
+                    end,
+                },
+                {
+                    text = self.rotated and _("No rotation") or _("Rotate"),
+                    callback = function()
+                        self.rotated = not self.rotated and true or false
+                        self:update()
+                    end,
+                },
+                {
+                    text = _("Close"),
+                    callback = function()
+                        UIManager:close(self)
+                    end,
+                },
             },
-            {
-                text = self.rotated and _("No rotation") or _("Rotate"),
-                callback = function()
-                    self.rotated = not self.rotated and true or false
-                    self:update()
-                end,
+        }
+        local button_table = ButtonTable:new{
+            width = self.width - 2*self.button_padding,
+            button_font_face = "cfont",
+            button_font_size = 20,
+            buttons = buttons,
+            zero_sep = true,
+            show_parent = self,
+        }
+        button_container = CenterContainer:new{
+            dimen = Geom:new{
+                w = self.width,
+                h = button_table:getSize().h,
             },
-            {
-                text = _("Close"),
-                callback = function()
-                    UIManager:close(self)
-                end,
-            },
-        },
-    }
-    local button_table = ButtonTable:new{
-        width = self.width - 2*self.button_padding,
-        button_font_face = "cfont",
-        button_font_size = 20,
-        buttons = buttons,
-        zero_sep = true,
-        show_parent = self,
-    }
-    local button_container = CenterContainer:new{
-        dimen = Geom:new{
-            w = self.width,
-            h = button_table:getSize().h,
-        },
-        button_table,
-    }
+            button_table,
+        }
+        button_table_size = button_table:getSize().h
+    end						   
 
     -- height available to our image
-    local img_container_h = self.height - button_table:getSize().h
+    local img_container_h = self.height - button_table_size
 
     local title_bar, title_sep
     if self.with_title_bar then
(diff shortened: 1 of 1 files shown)


poire-z:
Looks fine.
I get a crash when toggling the caption when viewing a Wikipedia lookup image:
```
./luajit: frontend/ui/widget/imageviewer.lua:417: attempt to index field 'dimen' (a nil value)
stack traceback:
       frontend/ui/widget/imageviewer.lua:417: in function 'refreshfunc'
       frontend/ui/uimanager.lua:949: in function '_repaint'
       frontend/ui/uimanager.lua:1047: in function 'handleInput'
```
when tapping on that arrow in yellow:
<kbd>image</kbd>

Frenzie:
I'm not entirely sure what I think of the bottom quarter approach. It's probably more intuitive to toggle the buttons on the whole screen and to have a dedicated swipe down action (as in various other widgets) as a close shortcut.

On a similar thought to the one in this PR, perhaps it'd be better to rotate images that are wider than they are high by default.

poire-z:
This ImageViewer already makes heavy use of swipe/pan/hold events, including swipe to pan big images (and swipe on the left/right borders for zoom in/out on devices without multitouch).
It would make it quite complicated to have swipe do other things (depending on if the image overflows screen and pan should work).
So, probably better/simpler/more consistent if that ImageViewer keeps being in its own standalone gesture world.

max-cw:
> Looks fine.
> I get a crash when toggling the caption when viewing a Wikipedia lookup image:
> 
> ```
> ./luajit: frontend/ui/widget/imageviewer.lua:417: attempt to index field 'dimen' (a nil value)
> stack traceback:
>        frontend/ui/widget/imageviewer.lua:417: in function 'refreshfunc'
>        frontend/ui/uimanager.lua:949: in function '_repaint'
>        frontend/ui/uimanager.lua:1047: in function 'handleInput'
>        frontend/ui/uimanager.lua:1119: in function 'run'
>        ./reader.lua:260: in main chunk
> ```
> 
> when tapping on that arrow in yellow:
> image

I need a bit of guidance on how to fix this. My best guess is that because the button_container is now empty, somehow it makes the self.main_frame.dimen a `nil` value?

I add a code to only insert the button_container when the buttons are shown, but as expected that didn't help:
```lua
if self._show_buttons then
	table.insert(frame_elements, button_container)
end
```

poire-z:
> I need a bit of guidance on how to fix this. My best guess is that because the button_container is now empty, somehow it makes the self.main_frame.dimen a nil value?

I don't think so. It's a bit more linked to KOReader internal flow for update/refresh/repaint.

Showing your involved diff, with just one line to add to fix this (hope that's enough):
```diff
@@ -439,13 +455,24 @@ function ImageViewer:onTap(_, ges)
     if self.caption_tap_area and ges.pos:intersectWith(self.caption_tap_area.dimen) then
         self.caption_visible = not self.caption_visible
         self:update()
+        return true                     -- <=== add this
     end
     if self._images_list then
+        -- If it's a list of image (e.g. animated gifs), tap each size to navigate
         if ges.pos.x < Screen:getWidth()/2 and self._images_list_cur > 1 then
             self:switchToImageNum(self._images_list_cur - 1)
         elseif ges.pos.x > Screen:getWidth()/2 and self._images_list_cur < self._images_list_nb then
             self:switchToImageNum(self._images_list_cur + 1)
         end
+    else
+        -- Otherwise, tapping the bottom will toggle the image,
+        -- and tapping the rest will close the widget
+        if ges.pos.y > Screen:getHeight() * 3/4 then
+            self._show_buttons = not self._show_buttons
+            self:update()
+        else
+            self:onClose()
+        end
     end
     return true
 end
```
The issue is that with your added `self:update()` if y>3/4*h, there were then 2 calls to it when I tap on that arrow.
The first update() resets everything, but at the end, there's some delayed stuff (setDirty) with a callback that expects stuff to have been painted in the interval, and have a `dimen` computed in that process.
Your second call to update() must mess with all that, in some ways :)

Adding that `return true` is the right thing to do (generally, to stop propagation to other widgets that could handle tap too, but even here, to stop further tap handling conditions to be checked) as the first test has handled the tap.
(It should have been there, I guess I forgot to add it - may be it would already have crashed before in some conditions - animated gif with caption.)

poire-z:
Or may be not :)
That caption tap y is **not** >3/4*h - so may be the conflict was with your `onClose` - this onClose cleans things up, but the delayed callback set by the first caption update() call then meets everything cleaned up and finalized, so no more `dimen`.
Any ways, that `return true` avoids that too.

max-cw:
> Or may be not :)
> That caption tap y is **not** >3/4*h - so may be the conflict was with your `onClose` - this onClose cleans things up, but the delayed callback set by the first caption update() call then meets everything cleaned up and finalized, so no more `dimen`.
> Any ways, that `return true` avoids that too.

The `return true` works like a charm! Thank you.

> I'm not entirely sure what I think of the bottom quarter approach. It's probably more intuitive to toggle the buttons on the whole screen and to have a dedicated swipe down action (as in various other widgets) as a close shortcut.

While I agree that toggle the buttons with the whole screen is more intuitive, this will make it always a two-tap operation to close a picture while reading, so I really want to have a quick way to close the widget. I actually thought about using swipe up to toggle the buttons, but apart from this fact that it's already used, it would be very hard to discover. I can do half and half, or the opposite (3/4 for toggle), but then I think no one will find out that you can close it by tapping the top quarter. So this bottom quarter thing is a compromised I came up with.

Frenzie:
@poire-z 
> So, probably better/simpler/more consistent if that ImageViewer keeps being in its own standalone gesture world.

Point taken. Still, I wonder if we could implement some kind of standard "close widget" multiswipe.

> I really want to have a quick way to close the widget.

By contrast, I just want a quick way to toggle original size and rotate. ;-) I don't view losing the controls behind a secondary step as an improvement, although with images of a particular aspect ratio I can see some appeal.

max-cw:
> By contrast, I just want a quick way to toggle original size and rotate. ;-) I don't view losing the controls behind a secondary step as an improvement, although with images of a particular aspect ratio I can see some appeal.

Understood. I agree that it would be best if there is a way to do everything the buttons do without having to toggle them back. This is beyond what I know how to do though. Do I need to do anything more on this?

Frenzie:
I mainly have my doubts about the 25 % thing, based on trying it out. It strikes me as either a bit large compared to `DTAP_ZONE_CONFIG` (1/8), or too small compared to the whole screen.

The rest is just something to keep in mind for the future.

poire-z:
Another possibility: Tap always just toggle buttons. And Swipe down closes, but only as long as there are no conflict - so, only on initial launch.
As we (our current code base) always launch with `fullscreen=true` and `scale_factor = 0, -- start with image scaled for best fit`, the image is initially always scaled to fit, and there's no use for panning with swipe - only for swiping on the sides to zoom in/out (for not mutitouch devices, including the emulator).

So, in init():
`if fullscreen=true and scale_factor=0 then self.allow_swipe_down_to_close=true end
`and in update() (so, any call to update reset that, which mean we have seen the buttons and know how to call them back):
`self.allow_swipe_down_to_close=false
`
In OnSwipe()
```lua
    elseif direction == "south" then
        if ges.pos.x < Screen:getWidth() * 1/16 or ges.pos.x > Screen:getWidth() * 15/16 then
            -- allow for zooming with vertical swipe on screen sides
            local dec = ges.distance / Screen:getHeight()
            self:onZoomOut(dec)
        else
+           if self.allow_swipe_down_to_close then
+                self:onClose()
+                return true
+           end
            self:panBy(0, -distance)
        end
```
(or `if scale_factor==0 ` instead of `if self.allow_swipe_down_to_close ` if you want the thing consistent instead of just on initial launch.)

and OnTap would toggle buttons on all height - while allowing a quick swipe down to dismiss - but once you started playing with it, you'll all to use the bottom button to close it.
Dunno if that's better :) just some quick thoughts.

Frenzie:
> Another possibility: Tap always just toggle buttons. And Swipe down closes, but only as long as there are no conflict - so, only on initial launch.

That's basically what I meant above, although I did forget about the "panning" (more like next page). ;-)

poire-z:
> On a similar thought to the one in this PR, perhaps it'd be better to rotate images that are wider than they are high by default.

Not so keen on that. For some landscape image, going from a timestamp to a full screen width size (even if only half of height is used) is often enough to get a better look at the image.
Going automatically to it rotated and indeed taking more of the screen area will need me to rotate the device, or move hands, or twist my neck, with the risk on some devices to trigger the accelerometer that will rotate again things, or will rotate the book below when I quit the ImageViewer in that position ... :)

AlanSP1:
Please, inform what is finally decided for new Image viewer controls, to add info to wiki, as at the moment still can be some changes...

poire-z:
> Do I need to do anything more on this?

@max-cw : looks like we'd prefer to have Tap on any part of the screen toggle buttons, and use Swipe south for quick closing (only on initial view, or under some conditions where there is no conflict - depends on how it will feel). (Swipe south is already used for close in quite a few widgets.)
Would that be fine with you? (or are you already too much used to tap on bottom? :)

Want to follow up on this? Or should we merge this and I can take over?
(I'd actually additionally like that with animated gifs, the buttons start hidden too (but with the progress bar always visible at bottom), for consistency, and with these only, only Tap in the middle half of the screen would toggle buttons - Tap on the 1/3 left and 1/3 right would change frames.)

max-cw:
@poire-z If you could take over, that would be great. The only thing among what you said that I can comfortably implement is using the whole screen to toggle, which would be trivial. So I think it's better to merge and close this first.

I have no opposition to what you proposed. Indeed it would be nice to have a consistent way to close a widget.

poire-z:
I figured I can just push commits to your branch, so all (discussion and changes) will be contained in this PR and make it in a single squashed commit.

poire-z:
@max-cw : github won't let me request a review from you - but give us your OK :)

Frenzie:
You can't approve or request changes on your own PR. I guess because otherwise a setting like 1 approval before merge would be useless.

poire-z:
Thanks !

PR #5451, opened 2019-10-03
cre: properly unmark link position markers

poire-z:
When following or back from link, that marker is shown (here in dual pages mode):
<kbd>image</kbd>

This PR avoids this from happening when it is removed a second later:
<kbd>image</kbd>

or this when some background texture is used:
<kbd>image</kbd>

### DIFF
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -604,18 +604,41 @@ function ReaderRolling:onGotoXPointer(xp, marker_xp)
 
         self.mark_func = function()
             self.mark_func = nil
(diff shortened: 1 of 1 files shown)


NiLuJe:
:D. Sneaky ;).

poire-z:
I guess this is fine as long as we use a white background. But if we end up allowing background textures (the cre.cpp method is there, just no UI to select one yet), this will still leave a little white hole in that background.
So, I guess ultimately, we should keep a small buffer of the original content we're drawing the black marker on, and restore it on unmark.

poire-z:
Don't merge this. I'll have these markers done the proper way, by saving and restoring the original content they are drawn on.

Issue #5415, opened 2019-09-23
Can't tap on confirmbox when keyboard shown


mustafa292:
When LoginDialog:onShowKeyboard called when there is a Confirmbox on the screen, Comfirmbox stays on screen (since it has .modal = true) but events won't propagate to it. Instead all events go to keyboard or login dialog that called keyboard. To see it in effect you can look at evernote login or kosync login. When you click to login, wi-fi dialog pops up but you cant accept or dismiss it. Keyboard and login dialog at bottom still works. I tried isolate the problem as much as I can in this [wbuilder.lua](https://gist.github.com/kereviz/971261625e617f5d8128e2e40520c13f) .

 [to see in gif ](https://postimg.cc/QByXwxzw)

First I thought it was simply two modal widgets clashing but this happens even when I set VirtualKeyboard's .modal = false. So when only login dialog called after Confirmbox everything is fine, Confirmbox still on top of the window_stack and it responds. Only calling :onShowKeyboard stops Comfirmbox responding; VirtualKeyboard and LoginDialog responds as they are .is_always_active widget's. It doesn't move Confirmbox from top of the stack (i think, since it is on top of screen and keyboard responds when is_always_active widgets called in UIManager:sendEvent, not when toplevelwidget called)).  

So i'm confused, and it is possible the paragraph above doesn't make any sense :P . 

There is two workarounds to this .

1- simply call NetworkMgr:checkconnectivity after :onShowKeyboard. But this isn't how modal widgets works, they should be on screen until user responds. So this takes away advantage of ConfirmBox being modal.

2- make ConfirmBox.is_always_active true, but I don't truly understand what is exactly the aim of that flag.

Could this have any relation to VirtualKeyboard inheriting from FocusManager?

Frenzie:
I don't quite have time to fully comprehend what you wrote atm, but I'll just drop this quick link.

http://koreader.rocks/doc/topics/Events.md.html

> When you call UIManager:show on a widget, this widget will be added to the top of the UIManager._window_stack. Events are sent to the first widget in UIManager._window_stack. If it is not consumed, then UIManager will try to send it to all active widgets (widget.is_always_active equals true) in the _window_stack.

mustafa292:
OK, I don't know what i was thinking at that moment but reason of this problem is actually simple. And a comment on code explains it clearly.  
[...]

Issue #4034, opened 2018-06-29
Trying to Enable WiFi just before Logging to Evernote Freezes Device


t5k6:
* KOReader version: v2015.11-1735-g14e7730_2018-06-02
* Device: Kobo Aura - Mark 5

#### Issue

If WiFi is not enabled beforehand and I try to login to Evernote, it brings me the menu to turn the WiFi on. However, once the cursor is in the username textbox, I cannot click any option about WiFi. All I can do is power off the device to get rid of the menu items.

![6_27_18 10_57 pm office lens](https://user-images.githubusercontent.com/12563212/42112839-e8647310-7bf1-11e8-9fae-27122ab5cf53.jpg)

This does not happen when I am already logged in to Evernote.

NiLuJe:
Only slightly related: I have somewhat of a minor issue with the whole WiFi UI:

* When it successfully connects, none of the popups get autodiscarded (neither the list of SSIDS, nor the infobox). I can understand why we'd want to keep the infobox, but the SSID list is fairly useless at that point (and, if you're in an urban area, chances are it'll be nearly full-screen, which makes the "discard it" tap areas somewhat small).

* When a feature requires WiFi and the "ask if offline" option is on, you do get steered towards enabling WiFi, but the previous action is not resumed once connected.

Frenzie:
All true.

poire-z:
Closing as probably obsolete after #6424 and other previous Wi-Fi related PRs.

PR #5452, opened 2019-10-04
Fix frozen confirmbox when keyboard shown

mustafa292:
see https://github.com/koreader/koreader/issues/5415 https://github.com/koreader/koreader/issues/4034

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -694,16 +694,12 @@ function UIManager:sendEvent(event)
                     if active_widget:handleEvent(event) then return end
                 end
             end
-            if widget.widget.is_always_active then
+            if widget.widget.is_always_active or widget.widget.modal then
                 -- active widgets will handle this event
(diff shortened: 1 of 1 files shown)


frontend/ui/uimanager.lua
```diff
@@ -701,6 +701,12 @@ function UIManager:sendEvent(event)
                 checked_widgets[widget] = true
                 if widget.widget:handleEvent(event) then return end
             end
+            if widget.widget.modal then
```
Frenzie: Seems like it could just be an or?

poire-z: It could be a `or` indeed, but I don't mind the split because of the lengthy comments.
But that `if` might be a `elseif`. Dunno if a `is_always_active` might also be a `modal`, but calling _twice_ handleEvent on a same widget, even if it returns false, might have some side effect (can't think of any right now, but logically, it could).
A `or` would solve this issue too, so why not :)

PR #5456, opened 2019-10-06
WIP [UX] Add … button to highlight edit

### LINKED ISSUE #5455: (fr) add a Wikipedia lookup button to the selection expand / close ui
Use case 

You've set default selection to highlight.  This means any multi word selection will be turned to a highlight.  But now say there's a proper noun that you'd like to lookup via Wikipedia.  Currently impossible

Idea would be to add a Wikipedia lookup button to selection expander/reducer so you could look up proper nouns

Frenzie:
Shows regular highlight dialog. Fixes #5455.

<hr>

<del>Perhaps slightly hackish.</del> <ins>Plus a couple of complementary crashes. ;-)</ins>

Anyway, just playing around.

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -167,7 +167,7 @@ function ReaderHighlight:onClearHighlight()
 end
 
 function ReaderHighlight:onTap(_, ges)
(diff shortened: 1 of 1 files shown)


Frenzie:
You're right, the hackish parts are mostly elsewhere. :-P

PR #5458, opened 2019-10-07
[Android] Ship Noto

### LINKED ISSUE #5381: Chinese characters in UI not displaying
* KOReader version: v2019.09.3-13-g2a93e42_2019-9-15
* Device: Nova 
The Chinese characters in the UI are not showing. The device may be using another font as the default font for Chinese characters instead of Noto Sans CJK. I tried to put NotoSans-regular.ttf and NotoSansCJKsc-regular.otf under Storage/fonts but that didn't work.
IMG_20190915_213804

IMG_20190915_214410

Frenzie:
Reverts a small part of #5310.

I don't really have the time to investigate doing font fallback better and the Android landscape is a tad far from uniform.

Fixes #5381.


poire-z:
Still thinking?

Frenzie:
Well, I don't really want to. :-)

PR #5460, opened 2019-10-07
[fix] Try to make screensaver settings slightly less error-prone

### LINKED ISSUE #5459: Missing screensaver settings submenu on Kobos
* KOReader version: 513fd807
* Device: Kobo Aura One, Clara HD and Forma

#### Issue
I cannot find the submenu where setting screensaver directory and picking screensaver image is possible (looks like [this submenu inclusion does not include](https://github.com/koreader/koreader/blob/master/frontend/apps/reader/modules/readermenu.lua#L178)).

https://www.dropbox.com/s/qpbq7vom6ldxfgv/20191007_162824~2_1_1.mp4?dl=0

Frenzie:
Regression inadvertently caused by #5369.

With the way this is constructed, this is almost guaranteed to happen.

Fixes #5459.

### DIFF
--- a/frontend/apps/reader/modules/readermenu.lua
+++ b/frontend/apps/reader/modules/readermenu.lua
@@ -175,7 +175,11 @@ function ReaderMenu:setUpdateItemTable()
             text = _("Screensaver"),
             sub_item_table = require("ui/elements/screensaver_menu"),
         }
-        table.remove(self.menu_items.screensaver.sub_item_table, 9)
+        local table_len = #self.menu_items.screensaver.sub_item_table
+        --- @todo Make this less prone to being messed up by screensaver menu changes, for example with MenuSorter.
+        if self.menu_items.screensaver.sub_item_table[table_len].text == ss_book_settings.text then
+            table.remove(self.menu_items.screensaver.sub_item_table, table_len)
+        end
         table.insert(self.menu_items.screensaver.sub_item_table, ss_book_settings)
     end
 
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readermenu.lua
```diff
@@ -175,7 +175,11 @@ function ReaderMenu:setUpdateItemTable()
             text = _("Screensaver"),
             sub_item_table = require("ui/elements/screensaver_menu"),
         }
-        table.remove(self.menu_items.screensaver.sub_item_table, 9)
+        local table_len = #self.menu_items.screensaver.sub_item_table
+        --- @todo Make this less prone to being messed up by screensaver menu changes, for example with MenuSorter.
+        if self.menu_items.screensaver.sub_item_table[table_len].text == ss_book_settings.text then
```
Frenzie: I'll just write this line a bit more error-proof. Cf. https://github.com/koreader/koreader/issues/5469#issuecomment-541281013

poire-z: Still thinking ? Not going with your suggestion from https://github.com/koreader/koreader/issues/5469#issuecomment-541357260 ? (Which I guess you'll have to use at the 2 places you changed things in this PR.)

poire-z:
https://github.com/koreader/koreader/issues/5469#issuecomment-541329007 Had it crashed 2 or 3 times a day since this... and couldn't reproduce it when willing... but there's a reason :) You have to switch documents with a gesture 20 times! (or 18 or 22, haven't counted).

Enabling `function Device:supportsScreensaver() return true end` on the emulator, I notice that when using a gesture to switch document, the Screensaver shrink by 1 item every 2 document switches (without doing anything else than this gesture - if I go look at the menu content between every 1 document switch, it doesn't shrink :| )
If I use the top hamburger menu to Open previous document, it never shrinks.

So, the quick fix at https://github.com/koreader/koreader/issues/5469#issuecomment-541338936 of adding a check `if table_len > 0` does prevent the crash when it reaches 0, but it actually does not prevent that menu from shrinking...
Not sure why the different behaviour... (Possibly an additional onCloseDocument emited when invoked from a gesture?)

But I guess it's because the other place where you removed last item is not checking if that last item has actually `.text == ss_book_settings.text`:
https://github.com/koreader/koreader/blob/bae69c0ef3639d5a1647815dd27225d0267c587a/frontend/apps/reader/modules/readermenu.lua#L353-L359

Making a PR, by just adding a `is_ss_book_settings=true` to the menu item we add, so we can find it and remove only it in both places.

PR #5463, opened 2019-10-09
[Android, fix] Revert slightly more of the Noto business

Frenzie:
Follow-up to #5458. Cf. https://github.com/koreader/koreader/issues/5381#issuecomment-539873843.


Frenzie:
Thanks, I missed that.

Mayonnnaise:
Hi, I found that CJK characters in the footnotes are not displaying either. It seems that the latin characters are displaying in Noto Sans in the latest version, though.

Frenzie:
Could you please open a new ticket?

PR #5467, opened 2019-10-10
[feat] plugins/Wallabag: add ability to set a list of tags to ignore

teroshan:
A new setting has been adding allowing to set a comma separated list of
tags to ignore.

Entries with either of these tags will be skipped by the client when
fetching the list of articles to download. Extra articles will be
fetched from the server to make up for the skipped articles, ensuring to
meet the target number of articles.

This can be used to ignore some articles that are known to not be readable in koreader: videos, binaries,.. As long as they are correctly tagged on the server. 
Another use case could be, once the ability to tag an entry in the client has been added, to ignore some entries and not download them again.

Most of the code (for the UI) was copied from the existing `filter_tag` setting.

### DIFF
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -188,12 +188,12 @@ function Wallabag:addToMainMenu(menu_items)
                     {
                         text_func = function()
                             local tags
-                            if not self.ignore_tags or self.ignore_tags == "" then
-                                tags = _("None")
+                            if not self.ignore_tags then
+                                tags = ""
                             else
                                 tags = self.ignore_tags
                             end
-                            return T(_("Ignore tags: %1"), tags)
+                            return T(_("Ignore tags (%1)"), tags)
                         end,
                         keep_menu_open = true,
                         callback = function(touchmenu_instance)
@@ -372,10 +372,12 @@ function Wallabag:getArticleList()
         new_article_list = self:filterIgnoredTags(new_article_list)
 
         -- Append the filtered list to the final article list
-        local num_articles_left = self.articles_per_sync - #article_list
-        local num_articles_to_append = math.min(#new_article_list, num_articles_left)
-        for i=0, num_articles_to_append, 1 do
-            table.insert(article_list, new_article_list[i])
+        for i, article in ipairs(new_article_list) do
+            if #article_list == self.articles_per_sync then
+                logger.dbg("Wallabag: hit the article target", self.articles_per_sync)
+                break
+            end
+            table.insert(article_list, article)
         end
 
         page = page + 1
@@ -391,7 +393,7 @@ function Wallabag:filterIgnoredTags(article_list)
     -- decode all tags to ignore
     local ignoring = {}
     if self.ignore_tags ~= "" then
-        for tag in util.gsplit(self.ignore_tags, "[, ]+", false) do
+        for tag in util.gsplit(self.ignore_tags, "[,]+", false) do
             ignoring[tag] = true
         end
     end
@@ -790,7 +792,7 @@ end
 function Wallabag:setIgnoreTags(touchmenu_instance)
    self.ignore_tags_dialog = InputDialog:new {
         title =  _("Tags to ignore"),
-        description = _("Set a comma-separated list of tags to ignore"),
+        description = _("Enter a comma-separated list of tags to ignore."),
         input = self.ignore_tags,
         input_type = "string",
         buttons = {
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -45,6 +45,7 @@ function Wallabag:init()
     self.is_sync_remote_delete = false
     self.is_archiving_deleted = false
     self.filter_tag = ""
+    self.ignore_tags = ""
     self.articles_per_sync = 30
 
     self.ui.menu:registerToMainMenu(self)
@@ -72,6 +73,9 @@ function Wallabag:init()
     end
     if self.wb_settings.data.wallabag.filter_tag then
         self.filter_tag = self.wb_settings.data.wallabag.filter_tag
+    end
+    if self.wb_settings.data.wallabag.ignore_tags then
+        self.ignore_tags = self.wb_settings.data.wallabag.ignore_tags
     end
     if self.wb_settings.data.wallabag.articles_per_sync ~= nil then
         self.articles_per_sync = self.wb_settings.data.wallabag.articles_per_sync
@@ -179,6 +183,21 @@ function Wallabag:addToMainMenu(menu_items)
                         keep_menu_open = true,
                         callback = function(touchmenu_instance)
                             self:setFilterTag(touchmenu_instance)
+                        end,
+                    },
+                    {
+                        text_func = function()
+                            local tags
+                            if not self.ignore_tags or self.ignore_tags == "" then
+                                tags = _("None")
+                            else
+                                tags = self.ignore_tags
+                            end
+                            return T(_("Ignore tags: %1"), tags)
+                        end,
+                        keep_menu_open = true,
+                        callback = function(touchmenu_instance)
+                            self:setIgnoreTags(touchmenu_instance)
                         end,
                     },
                     {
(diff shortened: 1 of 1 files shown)


plugins/wallabag.koplugin/main.lua
```diff
@@ -315,13 +334,62 @@ function Wallabag:getBearerToken()
...
     local filtering = ""
+    local ignoring = {}
     if self.filter_tag ~= "" then
         filtering = "&tags=" .. self.filter_tag
     end
-    local articles_url = "/api/entries.json?archive=0&perPage=" .. self.articles_per_sync .. filtering
-    return self:callAPI("GET", articles_url, nil, "", "")
+    if self.ignore_tags ~= "" then
+        local tag
+        -- take a not of all the tags that we should ignore
```
Frenzie: The typo jumped out at me, but it can be a bit shorter/simpler too.
```suggestion
        -- Store all tags to ignore.
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -695,6 +763,36 @@ function Wallabag:setFilterTag(touchmenu_instance)
     self.tag_dialog:onShowKeyboard()
 end
 
+function Wallabag:setIgnoreTags(touchmenu_instance)
+   self.ignore_tags_dialog = InputDialog:new {
+        title =  _("Set a comma separated list of tags"),
```
```suggestion
        title =  _("Tags to ignore"),
```

I would put the instruction in the description instead.

plugins/wallabag.koplugin/main.lua
```diff
@@ -315,13 +334,62 @@ function Wallabag:getBearerToken()
...
 function Wallabag:getArticleList()
     local filtering = ""
+    local ignoring = {}
     if self.filter_tag ~= "" then
         filtering = "&tags=" .. self.filter_tag
     end
-    local articles_url = "/api/entries.json?archive=0&perPage=" .. self.articles_per_sync .. filtering
-    return self:callAPI("GET", articles_url, nil, "", "")
+    if self.ignore_tags ~= "" then
+        local tag
```
Frenzie: No need to define this; the for loop should be local.
```suggestion
```

```
Luacheck results
Checking plugins/wallabag.koplugin/main.lua       4 warnings

    plugins/wallabag.koplugin/main.lua:348:15: unused variable 'tag'
    plugins/wallabag.koplugin/main.lua:350:13: shadowing definition of variable 'tag' on line 348
    plugins/wallabag.koplugin/main.lua:371:19: unused variable 'tag'
    plugins/wallabag.koplugin/main.lua:373:20: shadowing definition of variable 'tag' on line 371
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -315,13 +334,62 @@ function Wallabag:getBearerToken()
...
+                          .. filtering
+        local articles_json = self:callAPI("GET", articles_url, nil, "", "")
+        if not articles_json then
+            -- we may have hit the last page, there are no more articles
+            logger.dbg("Wallabag: couldn't get page #", page)
+            break
+        end
+        -- Go through the articles and skip the ones that have ignored tags
+        for _, article in ipairs(articles_json._embedded.items) do
+            local tag
```
```suggestion
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -315,13 +334,62 @@ function Wallabag:getBearerToken()
...
+    local ignoring = {}
     if self.filter_tag ~= "" then
         filtering = "&tags=" .. self.filter_tag
     end
-    local articles_url = "/api/entries.json?archive=0&perPage=" .. self.articles_per_sync .. filtering
-    return self:callAPI("GET", articles_url, nil, "", "")
+    if self.ignore_tags ~= "" then
+        local tag
+        -- take a not of all the tags that we should ignore
+        for tag in  util.gsplit(self.ignore_tags, "[, ]+", false) do
```
```suggestion
        for tag in util.gsplit(self.ignore_tags, "[, ]+", false) do
```

plugins/wallabag.koplugin/main.lua
teroshan: Yep I was going to fix this this evening, I didn't know that this variable was local to the loop.
Should've understood this from the other iterators. Good thing there is CI :)

teroshan:
I took your comments into account.

I also refactored the filtering a little bit, and put it in a new function.
I'm not quite sure what the proper way of iterating through an array is in lua, so If someone's got some advice regarding that I'll be glad to take it.

Frenzie:
Depends on the table. Just pairs/ipairs for your basic ones.

plugins/wallabag.koplugin/main.lua
```diff
@@ -181,6 +185,21 @@ function Wallabag:addToMainMenu(menu_items)
                             self:setFilterTag(touchmenu_instance)
                         end,
                     },
+                    {
+                        text_func = function()
+                            local tags
+                            if not self.ignore_tags or self.ignore_tags == "" then
+                                tags = _("None")
```
```suggestion
                                return _("Ignore tags")
```

I don't think anything more is required, but if you think something like `none` is a good addition it should just be part of the full string instead of harder to localize through programmatic cleverness. ;-)

plugins/wallabag.koplugin/main.lua
```diff
@@ -315,13 +334,86 @@ function Wallabag:getBearerToken()
...
+            table.insert(new_article_list, article)
+        end
+
+        -- Apply the filters
+        new_article_list = self:filterIgnoredTags(new_article_list)
+
+        -- Append the filtered list to the final article list
+        local num_articles_left = self.articles_per_sync - #article_list
+        local num_articles_to_append = math.min(#new_article_list, num_articles_left)
+        for i=0, num_articles_to_append, 1 do
```
```suggestion
        for i=1, num_articles_to_append, 1 do
```
This isn't very Lua. I know, I hate it when people say some particular construct isn't "Pythonic" but having some random C-style 0 in there is just confusing both to readers and the built-in table-handling functions.

Frenzie: Also I specifically hate it when people say that without specifying what *is* and I think I dodged that bullet. ;-)

plugins/wallabag.koplugin/main.lua
```diff
@@ -695,6 +787,37 @@ function Wallabag:setFilterTag(touchmenu_instance)
     self.tag_dialog:onShowKeyboard()
 end
 
+function Wallabag:setIgnoreTags(touchmenu_instance)
+   self.ignore_tags_dialog = InputDialog:new {
+        title =  _("Tags to ignore"),
+        description = _("Set a comma-separated list of tags to ignore"),
```
```suggestion
        description = _("Enter one or more tags to ignore, separated by a comma."),
```

That's how it's phrased in Wikipedia languages. That being said, I like the brevity of:

```suggestion
        description = _("Enter a comma-separated list of tags to ignore."),
```

Any thoughts on consistency/style @poire-z?

poire-z: Not really :) Feel free to rewrite it the way you prefer.
About commas vs spaces as separator, some people may already have these Wikipedia lang codes separated by space - so if we go with commas, there would need so code to support both - but well... :|
If tags here should go with spaces, dunno if a tag is allowed to have some space in it - so I can't really say.
But I don't really mind about the inconsistency of the separator - as long as the text says what's to be used.

Frenzie: Tags can (probably?) include a space. I was just talking about the `Enter one or more tags to ignore, separated by a space/comma.` vs `Enter a comma/space-separated list of tags to ignore.` ;-)

poire-z: Well, the 2nd is the nicest. But the first is probably the most explicite for technical and non technical people (One could wonder how to do a list of one tag: `[mytag]` `{"mytag"}`...)

teroshan: Your comments made me realize that tags can have spaces in them, so I fixed that :)
I went with the second string!

plugins/wallabag.koplugin/main.lua
teroshan: Damn that's right, Lua indexes start at 1!
I used `ipairs` instead.

plugins/wallabag.koplugin/main.lua
teroshan: I'm not sure to understand the issue, so I remove the "None" string, it will just be empty now.

teroshan:
I made a separate commit to show the changes.
I can rebase if you want.

plugins/wallabag.koplugin/main.lua
```diff
@@ -188,12 +188,12 @@ function Wallabag:addToMainMenu(menu_items)
                     {
                         text_func = function()
                             local tags
-                            if not self.ignore_tags or self.ignore_tags == "" then
-                                tags = _("None")
+                            if not self.ignore_tags then
+                                tags = ""
```
Frenzie: But then it'll just show `Ignore tags ()`. That's ugly. :-) My `return _("Ignore tags")` suggestion doesn't have that problem.

The problem with the `Ignore tags: %1` string is that you're programmatically combining the word `none` with a string. To translate the word none correctly, chances are it might need to be a contextualized (see [here](http://koreader.rocks/doc/modules/gettext.html#pgettext)). That's a lot of extra mental time and effort for a translator to wrap their head around when the full string `Ignore tags: none` will do the job just fine. Not to mention that a user might have a `none` tag (meaning something like "doesn't fit the other categories).

*However*, that's all a moot point because I think a simple `Ignore tags` is better overall. It's just a PSA: don't be clever with strings that have to be localized. Programmers always want to DRY (for good reason of course), but for localization it's often better to repeat yourself a little.

What works in English will likely to work quite well in closely related languages like Dutch and German, not typically too bad either in mostly similar but more distantly related languages like French. But as soon as you leave Western Europe, programmatic assumptions about language will bite you (or perhaps more to the point, they'll make all of our hard-working translators sad). It's as simple as that. ;-)

Frenzie: So in short, I stick with:
```suggestion
                                return  _("Ignore tags")
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -188,12 +188,12 @@ function Wallabag:addToMainMenu(menu_items)
...
                             else
                                 tags = self.ignore_tags
```
Frenzie: In which case `return T(_("Ignore tags (%1)"), self.ignore_tags)` can probably just go here without extra variables.

Frenzie:
> I can rebase if you want.

I'll just squash & merge it with GH. But feel free to rebase it first if you prefer. ;-)

plugins/wallabag.koplugin/main.lua
teroshan: I was wondering what those `_`, `T("foo")` functions were about, thanks for linking that documentation. 

I think I got it, unless what you wanted was just one `return _("Ignore tags")`? Because if I understand this correctly, if someone localizes `"Ignore tags"`, maybe someone else could localize `"Ignore tags (%1)"` differently right?

Frenzie: Right, that's the translator's prerogative. Even in English it might make sense to make the strings slightly different, like:

* `Ignore tags`
* `Ignored tags (one,two,three)`

Incidentally, I didn't notice this before merging but `Help` is now relegated to the second page.

Screenshot_20191013_091136

It might make sense to group these two together in a `Tags` submenu.

Frenzie: That would also open it up to more easily splitting the ignored tags themselves off from the ignoring. See for example how the multiswipe settings are grayed out when multiswipes aren't enabled, but you don't have to empty your settings to disable them. (Although if you're ignoring more than a handful of tags, you should probably just filter them? Or tag less? lol)

Screenshot_20191013_092208

Anyway, just some thoughts. Thanks for the addition!

PR #5468, opened 2019-10-11
[Kobo] Always set the warmth on the KA1

### LINKED ISSUE #5465: After a koreader restart the frontlight warmth resets to 0
* KOReader version: 2019-10-08
* Device: KoboA1

Noticed since last few updates. After restarting because of an update or just manually, the warmth becomes zero. When I open the frontlight menu, warmth shows 100 but it isn't.

I just have to adjust it twice to get it back to 100 warmth.

NiLuJe:
It might be possible to instead only set the white led, but I don't
have the hardware to test that theory, so, eh.

Fix #5465 (probably.)

### DIFF
--- a/frontend/device/kobo/powerd.lua
+++ b/frontend/device/kobo/powerd.lua
@@ -232,7 +232,13 @@ end
 
 function KoboPowerD:setIntensityHW(intensity)
     if self.fl == nil then return end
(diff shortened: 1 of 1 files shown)


NiLuJe:
(Regression since abc86323fc0e1396b6402ebb38b8a78e21c44e3e)

PR #5477, opened 2019-10-11
[Kobo] Log the FW/kernel version on crash

NiLuJe:
(Because it gets tedious asking for it in every bug report ;p).


platform/kobo/koreader.sh
```diff
@@ -260,6 +260,7 @@ while [ $RETURN_VALUE -ne 0 ]; do
 
         echo "!!!!" >>crash.log 2>&1
```
Frenzie: The CI doesn't seem to have run on your PR? Odd. 

Sidebar, this line triggers https://github.com/koalaman/shellcheck/wiki/SC2129 on newer SC.

Frenzie: Looks like it's not just newer SC. Oops. (Why didn't CirlceCI run? >_<)

PR #5482, opened 2019-10-12
fix spelling mistakes

merin2000:


README.md
```diff
@@ -50,7 +50,7 @@ Please follow the model specific steps for your device:
 
 
 
-## Development
+#a# Development
```
Frenzie:
```suggestion
## Development
```

Frenzie:
@NiLuJe @poire-z @pazos 

Hold on a sec, there seems to be an odd kind of spam here. I didn't notice when I merged #5483 that effectively the same PR existed some four times.

poire-z:
Yep...
There's a button in GitHub project's Settings to forbid PRs from non-previous-contributors for 24 hours, in case his happens again.

PR #5486, opened 2019-10-12
Use system font in Pocketbook

robert00s:
In Pocketbook Koreader should also use system fonts. 
Tested by Cyfranek from forum.eksiazki.org


Frenzie:
Weird, has it changed over the years or was it a typo omitting the `/system` part?

robert00s:
I don't know but people on forum reported problem with that so I found that proper folder is /mnt/ext1/system/fonts.
I think that it has changed overt the years.

PR #5488, opened 2019-10-12
android: fix ko paths when primary storage is a SD card

### LINKED ISSUE #5487: Unable to launch KO if external sdcard is used as main partition.
* KOReader version: all
* Platform: android

#### Issue

Unable to launch KO if external sdcard is used as main partition. 

#### Steps to reproduce
On android settings use the sdcard as the primary storage. System will reboot or not based on vendor. Launch KO. 

It will start on /sdcard which is now empty, because it is hardcoded on llapp_main.lua.

We're in a good position to fix this because the new path is already recognized:

```adb logcat
10-13 00:48:55.251  8588  8588 V KOReader: Application info {
10-13 00:48:55.251  8588  8588 V KOReader:   Flags: user.
10-13 00:48:55.251  8588  8588 V KOReader:   Paths {
10-13 00:48:55.251  8588  8588 V KOReader:     Assets: /data/user/0/org.koreader.launcher/files
10-13 00:48:55.251  8588  8588 V KOReader:     Library: /data/app/org.koreader.launcher-xM0Q166albC8kw0pQAKtYA==/lib/arm
10-13 00:48:55.251  8588  8588 V KOReader:     Storage: /storage/3837-3661
```

So we need to use reported storage as an argument to start KOReader instead of "/sdcard".

pazos:
Requires https://github.com/koreader/android-luajit-launcher/pull/195

Fixes #5487

~~to-do:~~
Removed: 
https://github.com/koreader/koreader/blob/5209493ba881f25e94cb4682d92416d515b2fb7c/defaults.lua#L238

### DIFF
--- a/frontend/ui/elements/font_settings.lua
+++ b/frontend/ui/elements/font_settings.lua
@@ -3,7 +3,7 @@ local logger = require("logger")
 local util = require("util")
 local _ = require("gettext")
 
---[[ Font settings for desktop linux and mac ]]--
+--[[ Font settings for desktop linux, mac and android ]]--
 
 local function getUserDir()
     local home = os.getenv("HOME")
@@ -52,6 +52,20 @@ function FontSettings:getPath()
         end
     end
     return getUserDir()
+end
+
+function FontSettings:getAndroidPath()
+    local A, android = pcall(require, "android")
+    if not A then return end
+    local system_path = "/system/fonts"
+    local user_path = android.getExternalStoragePath()
+    if user_path == "Unknown" then
+        -- unable to identify external storage. Use defaults
+        return system_path..";".."/sdcard/koreader/fonts"
+    else
+        -- use the external storage identified by the app
+        return system_path..";"..user_path.."/koreader/fonts"
+    end
 end
 
 function FontSettings:getMenuTable()
--- a/defaults.lua
+++ b/defaults.lua
@@ -234,9 +234,6 @@ NETWORK_PROXY = nil
 -- Use turbo library to handle async HTTP request
 DUSE_TURBO_LIB = false
 
--- Custom font dir for Android
-ANDROID_FONT_DIR = "/system/fonts;/sdcard/koreader/fonts"
-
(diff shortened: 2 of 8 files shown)


pazos:
damn, not sure how to fix those unit test

Frenzie:
Because you removed a line, that means that everything following will be off by one. So it's as simple as changing 112 to 111.

https://github.com/koreader/koreader/blob/40ba01fd72193d3dbc355f69405b3164d61126c5/spec/unit/defaults_spec.lua#L26

This one should probably be 56:

https://github.com/koreader/koreader/blob/40ba01fd72193d3dbc355f69405b3164d61126c5/spec/unit/defaults_spec.lua#L103-L104

That default file/the way the unit test for it is written is slightly annoying. ;-)

frontend/ui/elements/font_settings.lua
```diff
@@ -54,6 +54,21 @@ function FontSettings:getPath()
     return getUserDir()
 end
 
+function FontSettings:getAndroidPath()
+    local A, android = pcall(require, "android")
+    local system_path = "/system/fonts"
+    local user_path = "/koreader/fonts"
+    local legacy_font_dirs = system_path..";".."/sdcard"..user_path
+
+    if not A then return legacy_font_dirs end
```
Frenzie: Bit confused by this? For easier testing (in the emulator)?

Frenzie: That is, /system/fonts or /sdcard/koreader/fonts won't be very useful on non-Android platforms regardless.

pazos: yeah, more than anything a future proof thingy. I guess nobody will want to call FontSettings:getAndroidPath() outside of android. But if someone tries that it will receive the "legacy" android path.

Not sure if this is a good idea but is safer in the case we cannot load the android module.

Frenzie: Why not just nil or an empty string?

PR #5494, opened 2019-10-14
Minor state handling tweaks when the footer is disabled/invisible

NiLuJe:
* Don't break `doc_pages` updates when it's completely disabled. (c.f., the commit's comment, doing it on every page turn seems a bit overkill, but I get that it's probably done that way because it was easier ;)).
* Detect the empty footerTextGenerator properly in `_updateFooterText` (it's `""`, not `nil`).
* Abort `_updateFooterText` early when the footer is invisible (and has been for a while/ever, i.e., when not requesting a repaint, like a mode switch would).
* Never send `SetBottomPageMargin` events twice.
* Minor tweaks to touchmenu & configdialog to avoid useless repaints (don't redraw the reader/fm unless we absolutely need to).
* Make sure we show the progress bar alone instead of nothing when disabling the last enabled mode in the menu.


NiLuJe:
Pinging @robert00s to double-check that I didn't do anything stupid ;).

Frenzie:
Err, fine besides those anyway:

```
[  FAILED  ] spec/front/unit/readerfooter_spec.lua @ 186: Readerfooter module should switch between different modes
[  FAILED  ] spec/front/unit/readerfooter_spec.lua @ 275: Readerfooter module should pick up screen resize in resetLayout
[  FAILED  ] spec/front/unit/readerfooter_spec.lua @ 303: Readerfooter module should update width on PosUpdate event
```

NiLuJe:
Well, AFAICT, this is yet another case of a test doing handstands to try to replicate stuff, because doing the actual thing manually behaves properly :/.

NiLuJe:
I hate tests with silent, sneaky and stupid side-effects with a fiery passion.

NiLuJe:
(That last one is a regression since #5491, so do double-check that I didn't re-break the stuff that attempted to fix).

NiLuJe:
Now includes a free touchmenu tweak to avoid repainting useless stuff when it's not needed ;p.

(Happy coincidence that the Settings submenu of the Status Bar stuff doesn't change height, so I was intrigued by the useless `setDirty("all")` that I caught in the logs ;).)

NiLuJe:
Okay, I should be done with this now ;).

@robert00s, I've just noticed your thumbs-up, but I'd be glad for another review on the latest tweaks, thanks ;).

robert00s:
Looks good :)
Thanks.

Issue #5388, opened 2019-09-16
Font size for menu items: size oscillation


gerhaher:
Output font sizes for menu items in filemanger in classic mode oscillate between two sizes, when you choose different font sizes that is above what is limited by filename length etc. 
Not the biggest problem in the world, but I suppose, if anyone is interested, the algorithm can be improved. :-)

Anyway, see pictures below (Items per page = 24 on my KA1).

Font size = 12
![1FileManager_2019-Sep-10_184813](https://user-images.githubusercontent.com/32645430/64765723-f2d44200-d544-11e9-873c-a31a16554583.png)
Font size = 13
![1FileManager_2019-Sep-10_184833](https://user-images.githubusercontent.com/32645430/64765728-f7005f80-d544-11e9-8f78-54358e516f5c.png)
Font size = 14
![1FileManager_2019-Sep-10_184849](https://user-images.githubusercontent.com/32645430/64765735-fbc51380-d544-11e9-95ec-c35b4a0994fc.png)
Font size = 15
![1FileManager_2019-Sep-10_184903](https://user-images.githubusercontent.com/32645430/64765747-041d4e80-d545-11e9-9f12-8a75fd4cabf7.png)
Font size = 16
![1FileManager_2019-Sep-10_184919](https://user-images.githubusercontent.com/32645430/64765778-10091080-d545-11e9-8a99-0300b57a7002.png)
Font size = 17
![1FileManager_2019-Sep-10_184931](https://user-images.githubusercontent.com/32645430/64765785-139c9780-d545-11e9-92c6-cdc5b8d5cead.png)
Font size = 18
![1FileManager_2019-Sep-10_184944](https://user-images.githubusercontent.com/32645430/64765793-17c8b500-d545-11e9-97f5-49c693fa6449.png)
Font size = 19
![1FileManager_2019-Sep-10_185000](https://user-images.githubusercontent.com/32645430/64765798-1ac3a580-d545-11e9-8f05-a66eb0ea8236.png)
Font size = 20
![1FileManager_2019-Sep-10_185015](https://user-images.githubusercontent.com/32645430/64765812-1eefc300-d545-11e9-8cd2-ddc0487f575a.png)

Frenzie:
What I see on the picture is that it reduces the text size slightly to show as much text as possible. It's basically form vs function.

poire-z:
The even font size seems to go lower size than the odd ones...
Might be this `-2` decrement and/or the scaleForDpi rounding?
[...]

PR #5496, opened 2019-10-16
New algorithm to show items in menu widget

robert00s:
See: #5388 
New algorithm:
When text is too long or too height:
- first: we try to decrease number of lines (eg from 3 to 2 and from 2 to 1 line)
- second: when 1 line is too height we try to decrease font size (-1)
And at the end we try to add or remove at least one letter to better fit text.

Needs to be tested. Shouldn't be merge before 2019.10.

Close: #5388

### DIFF
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -347,8 +347,8 @@ function MenuItem:init()
             local text_last_line
             -- when lines is more than 1 we see only for last visible line
             if num_lines > 1 then
-                local offset_prev = item_name.vertical_string_list[num_lines - 1].offset - 1
-                text_last_line = table.concat(item_name.charlist, '', offset_prev, offset)
+                local offset_prev = item_name_orig.vertical_string_list[num_lines - 1].offset
+                text_last_line = table.concat(item_name_orig.charlist, '', offset_prev, offset)
             else
                 text_last_line = self.text
             end
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -303,11 +303,11 @@ function MenuItem:init()
             end
             -- remove last line and try again to fit
             offset = item_name.vertical_string_list[num_lines].offset - 1
-            -- remove ending '\n' (new line) to prevent infinity loop
-            if item_name.charlist[offset] == '\n' then
+            -- remove ending "\n" (new line) to prevent infinity loop
+            if item_name.charlist[offset] == "\n" then
                 offset = offset - 1
             end
-            self.text = table.concat(item_name.charlist, '', 1, offset)
+            self.text = table.concat(item_name.charlist, "", 1, offset)
         end
 
         -- second: decrease font size (we have now only one line)
@@ -317,7 +317,7 @@ function MenuItem:init()
                 item_name:free()
             end
             self.font_size = self.font_size - 1
-            item_name = TextBoxWidget:new {
+            item_name = TextBoxWidget:new{
                 text = self.text,
                 face = Font:getFace(self.font, self.font_size),
                 width = self.content_width - mandatory_w - state_button_width - text_mandatory_padding,
@@ -334,7 +334,7 @@ function MenuItem:init()
             -- if text is too height with that really small font we don't show text at all
             -- this shouldn't happen
             if self.font_size <= 8 then
-                item_name = TextBoxWidget:new {
+                item_name = TextBoxWidget:new{
                     text = "…",
                     face = Font:getFace(self.font, self.font_size),
                     width = self.content_width - mandatory_w - state_button_width - text_mandatory_padding,
@@ -352,7 +352,7 @@ function MenuItem:init()
             -- when lines is more than 1 we see only for last visible line
             if num_lines > 1 then
                 local offset_prev = item_name_orig.vertical_string_list[num_lines - 1].offset
-                text_last_line = table.concat(item_name_orig.charlist, '', offset_prev, offset)
+                text_last_line = table.concat(item_name_orig.charlist, "", offset_prev, offset)
             else
                 text_last_line = self.text
             end
@@ -365,7 +365,7 @@ function MenuItem:init()
             local max_offset = #item_name_orig.charlist
             -- try to add chars to better align
             while item_name.width > text_size_increase + ellipsis_size and offset < max_offset
-                and item_name_orig.charlist[offset] ~= '\n' do
+                and item_name_orig.charlist[offset] ~= "\n" do
                 text_size_increase = text_size_increase + item_name_orig:getCharWidth(offset + 1)
                 if text_size_increase + ellipsis_size < item_name.width then
                     -- add one char to text
@@ -380,14 +380,14 @@ function MenuItem:init()
             end
             if offset == max_offset then
                 -- when finally after manipulation we have all original text we don't need to add ellipsis
-                self.text = table.concat(item_name_orig.charlist, '', 1, offset)
+                self.text = table.concat(item_name_orig.charlist, "", 1, offset)
             else
                 -- remove ending '\n' (new line) to prevent increase number of lines
-                if item_name_orig.charlist[offset] == '\n' then
+                if item_name_orig.charlist[offset] == "\n" then
                     offset = offset - 1
                 end
                 -- add ellipsis to show that text was truncated
-                self.text = table.concat(item_name_orig.charlist, '', 1, offset) .. "…"
+                self.text = table.concat(item_name_orig.charlist, "", 1, offset) .. "…"
             end
 
             if item_name then
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -344,10 +344,18 @@ function MenuItem:init()
         end
         -- add ellipsis when text was truncated
         if flag_add_ellipsis then
+            local text_last_line
+            -- when lines is more than 1 we see only for last visible line
+            if num_lines > 1 then
+                local offset_prev = item_name.vertical_string_list[num_lines - 1].offset - 1
+                text_last_line = table.concat(item_name.charlist, '', offset_prev, offset)
+            else
+                text_last_line = self.text
+            end
+            local text_size = RenderText:sizeUtf8Text(0, self.content_width,
+                Font:getFace(self.font, self.font_size), text_last_line, true, self.bold).x
             local ellipsis_size = RenderText:sizeUtf8Text(0, self.content_width,
                 Font:getFace(self.font, self.font_size), "…", true, self.bold).x
-            local text_size = RenderText:sizeUtf8Text(0, self.content_width,
-                Font:getFace(self.font, self.font_size), self.text, true, self.bold).x
 
             local text_size_increase = text_size
             local max_offset = #item_name_orig.charlist
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -303,6 +303,10 @@ function MenuItem:init()
             end
             -- remove last line and try again to fit
             offset = item_name.vertical_string_list[num_lines].offset - 1
+            -- remove ending '\n' (new line) to prevent infinity loop
+            if item_name.charlist[offset] == '\n' then
+                offset = offset - 1
+            end
             self.text = table.concat(item_name.charlist, '', 1, offset)
         end
 
@@ -360,7 +364,8 @@ function MenuItem:init()
             local text_size_increase = text_size
             local max_offset = #item_name_orig.charlist
             -- try to add chars to better align
-            while item_name.width > text_size_increase + ellipsis_size and offset < max_offset do
+            while item_name.width > text_size_increase + ellipsis_size and offset < max_offset
+                and item_name_orig.charlist[offset] ~= '\n' do
                 text_size_increase = text_size_increase + item_name_orig:getCharWidth(offset + 1)
                 if text_size_increase + ellipsis_size < item_name.width then
                     -- add one char to text
@@ -377,6 +382,10 @@ function MenuItem:init()
                 -- when finally after manipulation we have all original text we don't need to add ellipsis
                 self.text = table.concat(item_name_orig.charlist, '', 1, offset)
             else
+                -- remove ending '\n' (new line) to prevent increase number of lines
+                if item_name_orig.charlist[offset] == '\n' then
+                    offset = offset - 1
+                end
                 -- add ellipsis to show that text was truncated
                 self.text = table.concat(item_name_orig.charlist, '', 1, offset) .. "…"
             end
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -268,6 +268,11 @@ function MenuItem:init()
         local mandatory_w = RenderText:sizeUtf8Text(0, self.dimen.w, self.info_face, "" .. mandatory, true, self.bold).x
         local max_item_height = self.dimen.h - 2 * self.linesize
         local flag_fit = false
+        local flag_add_ellipsis = false
+        local num_lines, offset
+        local item_name_orig = nil
+        -- first: try to decrease number of lines in TextBoxWidget
+        -- this loop ends only when text fits or we have only one line of text
         while true do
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,116 @@ function MenuItem:init()
...
+                break
+            end
+        end
+        -- add ellipsis when text was truncated
+        if flag_add_ellipsis then
+            local text_last_line
+            -- when lines is more than 1 we see only for last visible line
+            if num_lines > 1 then
+                local offset_prev = item_name_orig.vertical_string_list[num_lines - 1].offset
+                text_last_line = table.concat(item_name_orig.charlist, '', offset_prev, offset)
```
poire-z: One of the things I intend to do is to move that kind of stuff into TextBoxWidget.
It already has a "private" `TextBoxWidget:_getLineText(vertical_string)` which does what you do, so I guess it just have to have a "public" `TextBoxWidget:getLineText(line_num)` to give you that.
So, not asking you to do it :) This is fine for now.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,116 @@ function MenuItem:init()
...
+            end
+            local text_size = RenderText:sizeUtf8Text(0, self.content_width,
+                Font:getFace(self.font, self.font_size), text_last_line, true, self.bold).x
+            local ellipsis_size = RenderText:sizeUtf8Text(0, self.content_width,
+                Font:getFace(self.font, self.font_size), "…", true, self.bold).x
+
+            local text_size_increase = text_size
+            local max_offset = #item_name_orig.charlist
+            -- try to add chars to better align
+            while item_name.width > text_size_increase + ellipsis_size and offset < max_offset do
```
poire-z: What's the reason for this to be needed?
I wondered for a few minutes - but is it just because TextBoxWidget wrap by words, so the last line you keep may be shorter than the textboxwidget width because a wrap happened and a long word has been pushed onto the next line. So, here, you're trying to grab some chars from that long word to put it at end of last line, before the "...", so the "..." touches the screen right border.
Is it that? Is that the only reason for this to be needed and work?

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,116 @@ function MenuItem:init()
...
+                text_size_increase = text_size_increase + item_name_orig:getCharWidth(offset + 1)
+                if text_size_increase + ellipsis_size < item_name.width then
+                    -- add one char to text
+                    offset = offset + 1
                 end
-                self.text = table.concat(item_name.charlist, '', 1, offset) .. "…"
-                flag_fit = true
+            end
+            -- remove chars when text is too long
+            while item_name.width <= text_size + ellipsis_size do
```
poire-z: And the, why does it happen you need to remove chars? (because you added some and made sure they fitted with the ellipsis)?
Or does this happen and is only needed when you actually did not have to add some chars? So, just to get rid of the orignal chars on that last line, to make room for the ellipsis?

poire-z:
I'm now a bit more familiar with that code, after having contemplated it for some time before you stopped me :)

I think this multi-line display of Menu is only used by File browser when in Classic display mode.
(Are there others?)
So, I probably won't mind this big change, but there is a big change :)

Now, you're truncating text at 1 or 2 lines. If 2 lines fit, well, all done, we'll display the text truncated.
If 1 line does not fit (which I guess is an extrem case), only then you decrease the font size (in that case, I'm not sure by looking at the code, but aren't you left with the text truncated at some point, and when decreasing font size, you won't grab more text (as it is truncated) as they can fit on the single line, now that it has a smaller font size? - or is it what your code at my 2nd comment is also trying to fix? - if yes, it might be a bit convoluted, as you could just recreate a TextBoxWidget with the original text to get the most that fits with that final font size you decided on.)

So, the big change is that we won't get anymore the full text with a smaller font size, as we do currently.

Before:
<kbd>image</kbd>

After:
<kbd>image</kbd>
The 2nd line is shown with the font size like the 1st and the 3rd, so showing a bit less (truncated) that what we previously show untruncated.

Before:
<kbd>image</kbd>

After:
<kbd>image</kbd>

Again, I don't really mind this change if it applies only to File browser classic - as we mostly get the previous behaviour with the Detailed with filename mode:
<kbd>image</kbd>

(I prefer the previous behaviour, but I understand some people may prefer a fixed size - even if less is shown. They currently can not get that. With this PR, they would have an option for that.)

poire-z:
Another thing I noticed earlier.
In the multiline code, we decrease the "mandatory" (the item on the right) font size until it fits.
We don't do that in the single_line code, although everything is the same regarding the mandatory.
Moreover, we try to adjust the infont_size vs perpage, so it's just right and no decrease should be needed:
https://github.com/koreader/koreader/blob/bae69c0ef3639d5a1647815dd27225d0267c587a/frontend/ui/widget/menu.lua#L933-L936

I think the same should apply for the multiline main name font size - so that decreasing of font size, in this PR that displays only one line (but not before this PR), might be just a precaution in case the above formulae is a bit wrong and would find out a too large font size.

So, we could get rid of the mandatory font size loop in any case.
And if going with this PR, the main name font size loop may also be killed, to make the code simpler :)

frontend/ui/widget/menu.lua
robert00s: That's exactly how it works :)

frontend/ui/widget/menu.lua
robert00s: Second sentence. It's to make a room for ellipsis.

robert00s:
> I think this multi-line display of Menu is only used by File browser when in Classic display mode.
(Are there others?)

It's also used in Bookmark.

>So, the big change is that we won't get anymore the full text with a smaller font size, as we do currently.

Yes. I wanted to stay with fixed size, as long as possible. I think it looks prettier than every element looks similar (but it's only my opinion :))

> (I prefer the previous behaviour, but I understand some people may prefer a fixed size - even if less is shown. They currently can not get that. With this PR, they would have an option for that.)

If user want to see more text he/she can always decrease font size in` items -> font size` or decrease `items per page`

At the end. This is only my propose. If you have better implementation or another idea I don't insist on my solution :)

poire-z:
> It's also used in Bookmark.

Oh, right. Seeing more of the text is more useful there :) (But I'm not a heavy bookmarker, so not minding either).

> I think it looks prettier than every element looks similar (but it's only my opinion :))

Valid opinion (and it was my argument in #5435 - so I suddently agree with you :)

> At the end. This is only my propose. If you have better implementation or another idea I don't insist on my solution :)

No better idea (nor any real interest in having one).
Let's see what other people think. (Pinging @gerhaher who seems to use classic mode - but with many per-page item, so used to seeing only one line.)

poire-z:
While testing bookmarks, I got a reproducible infinite loop (I have to Ctrl-C) with just this PR:
Select 10 lines of text in a book, highlight. Go to top menu > Bookmarks:
```
[many of these for many seconds:]
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
10/16/19-21:45:00 DEBUG TextBoxWidget:free called
^C./luajit: interrupted!
stack traceback:
       [C]: in function 'insert'
       frontend/cache.lua:128: in function 'check'
       frontend/ui/rendertext.lua:88: in function 'getGlyph'
       frontend/ui/rendertext.lua:231: in function 'renderUtf8Text'
       frontend/ui/widget/textboxwidget.lua:380: in function '_renderText'
       frontend/ui/widget/textboxwidget.lua:125: in function 'init'
       frontend/ui/widget/menu.lua:281: in function 'init'
       frontend/ui/widget/menu.lua:1040: in function 'updateItems'
       frontend/ui/widget/menu.lua:951: in function 'init'
       frontend/apps/reader/modules/readerbookmark.lua:216: in function 'onShowBookmark'
       frontend/apps/reader/modules/readerbookmark.lua:40: in function 'callback'
       frontend/ui/widget/touchmenu.lua:729: in function 'action'
       frontend/ui/uimanager.lua:752: in function '_checkTasks'
       frontend/ui/uimanager.lua:1035: in function 'handleInput'
       frontend/ui/uimanager.lua:1121: in function 'run'
       [C]: at 0x55bbe69d6782
```
(600x610 emulator, 12 items per page, Font size: 16)

gerhaher:
>Let's see what other people think. (Pinging @gerhaher who seems to use classic mode - but with many per-page item, so used to seeing only one line.)

Yeah, as you said, I (sometimes) use classic, but then always 24 items/page, so I never see the multi-lines.

But I guess I find equal font size looks prettier.

robert00s:
> While testing bookmarks, I got a reproducible infinite loop (I have to Ctrl-C) with just this PR:
Select 10 lines of text in a book, highlight. Go to top menu > Bookmarks:

Should be fixed now. It was problem with ending new line '\n'

>Let's see what other people think. (Pinging @gerhaher who seems to use classic mode - but with many per-page item, so used to seeing only one line.)

Yep, we see only one line (with per-page=24) but we use algorithm for "multiline" so we shoudn't now see items truncated too short like in #5388. See my first and second screenshot :)

poire-z:
@Frenzie @NiLuJe @pazos : any opinion on this change?

I don't mind it (previous algo was @robert00s too, so his choice :)
I admit it's nicer looking than before.
(The only real added value of the previous algo was that in bookmarks, the size of the font for each item was a bit of an indicator of the length of the highlighted text :)

I find some bits in this code a bit convoluted - but I don't want to request any change, as I'll probably soon have to look at TextBoxWidget, so, now that we know what kind of stuff callers expect, we can integrate a bit of that code into new TextBoxWidget methods, and clean a bit of this in Menu (just like in I did for TextWidget in #5503).

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
+                break
+            end
+            flag_add_ellipsis = true
+            num_lines = item_name:getAllLineCount()
+            if num_lines == 1 then   -- widget doesn't fit and we have only one line of text
+                break
+            end
+            -- remove last line and try again to fit
+            offset = item_name.vertical_string_list[num_lines].offset - 1
+            -- remove ending '\n' (new line) to prevent infinity loop
```
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
+            if item_name.charlist[offset] == '\n' then
```
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
+                offset = offset - 1
+            end
+            self.text = table.concat(item_name.charlist, '', 1, offset)
```
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
+            local text_size = RenderText:sizeUtf8Text(0, self.content_width,
+                Font:getFace(self.font, self.font_size), text_last_line, true, self.bold).x
+            local ellipsis_size = RenderText:sizeUtf8Text(0, self.content_width,
+                Font:getFace(self.font, self.font_size), "…", true, self.bold).x
+
+            local text_size_increase = text_size
+            local max_offset = #item_name_orig.charlist
+            -- try to add chars to better align
+            while item_name.width > text_size_increase + ellipsis_size and offset < max_offset
+                and item_name_orig.charlist[offset] ~= '\n' do
```
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
+            while item_name.width <= text_size + ellipsis_size do
+                text_size = text_size - item_name:getCharWidth(offset)
+                -- remove one char from text
+                offset = offset - 1
+            end
+            if offset == max_offset then
+                -- when finally after manipulation we have all original text we don't need to add ellipsis
+                self.text = table.concat(item_name_orig.charlist, '', 1, offset)
+            else
+                -- remove ending '\n' (new line) to prevent increase number of lines
```
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
+                if item_name_orig.charlist[offset] == '\n' then
```
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
                     offset = offset - 1
                 end
-                self.text = table.concat(item_name.charlist, '', 1, offset) .. "…"
-                flag_fit = true
-            else
-                -- If we don't fit, decrease font size
-                self.font_size = self.font_size - 2
+                -- add ellipsis to show that text was truncated
+                self.text = table.concat(item_name_orig.charlist, '', 1, offset) .. "…"
```
Frenzie: Random single quotes instead of double.

frontend/ui/widget/menu.lua
```diff
@@ -281,34 +286,125 @@ function MenuItem:init()
...
+            end
+
+            if item_name then
+                item_name:free()
             end
+            if item_name_orig then
+                item_name_orig:free()
+            end
+            --final item_name that fits
+            item_name = TextBoxWidget:new {
```
Frenzie:
```suggestion
            item_name = TextBoxWidget:new{
```

poire-z:
Testing that again, I'm still frustrated with the result :)
(Not bothered personally - its only use I care about is the TOC display, which uses the single_line=true code and is not affected by this change.)

> If user want to see more text he/she can always decrease font size in items -> font size or decrease items per page

The problem with that is that if you want to see more of the long filenames/highlights, you'll get tiny font size for all items, even short ones, and for all things using our Menu class:
<kbd>image</kbd>
<kbd>image</kbd>
<kbd>image</kbd>
(I could be fine with that for Bookmarks, half-fine for File browser, but not at all with that TOC :)

When you raise font size, you're then switched to single line (but using the multi-lines code):
<kbd>image</kbd>
<kbd>image</kbd>
<kbd>image</kbd>

Raise it again, and you get something better:
<kbd>image</kbd>
<kbd>image</kbd>
<kbd>image</kbd>
(I guess everybody would be fine with that last one - do we all have the font size adjusted that way, or smaller like on previous set of screenshots?)

So, the multi line code switches now (very early when raising font size) to displaying a single line (so, with a static font size).
And it could as well have used the single line code then (that had no problem with oscillation).
And we could have kept the previous multi-line code for auto font sizing - and so, have it as another menu option:
```
Items per page >
Font size >
[x] Multi-lines (with previous auto-decreasing font size)
(or named [x] Auto-adjust font size
```

Then, only the first set of my screenshots above would be impossible - but would some people really want that ?!

Also, if I were not really interested in only having my TOC right, I'd be bothered that the same settings apply to different things (TOC, Bookmarks, Classic file browser), where I might want to see things laid out differently. But strangely, with the previous behaviour, this felt like not being an issue.)

Let's see if/how people feel about that.

poire-z:
> I'll probably soon have to look at TextBoxWidget, so, now that we know what kind of stuff callers expect, we can integrate a bit of that code into new TextBoxWidget methods, and clean a bit of this in Menu

Done in #5509 
I also resurrected the previous algorithm (that I prefered), which is available via an added menu option `Items > Reduce font size to show more text`.

Issue #5449, opened 2019-10-03
how to change size of highlight text box


leela52452:
version: 2019.09.03

i dont know, what the box is called.
it appears when i highlight text of an epub.
is there any way to increase the size of the box ?

i mean the box containing options: highlight, copy, view html, etc..

![Screenshot_2019-09-29-10-44-29](https://user-images.githubusercontent.com/52065244/66127837-26ead200-e5dc-11e9-9f7e-2b3fa45d78f0.png)

Frenzie:
You can set a custom DPI in settings (gear icon) → Screen → Screen DPI. All UI elements will automatically scale along.

poire-z:
If it's for reading purpose, the above suggestion is the one.

If it's because it's tedious taping on the right button, because of the small height of each, and the proximity of others - I share that feeling. When I have to tap on a button in that dialog, I need to be more careful than usually...
But I don't want to increase DPI and the font size. The solution could be to increase these buttons' padding, but that make the UI a bit inconsistent, and would hide a bit more of the text below. So, I dunno :|

leela52452:
> If it's because it's tedious taping on the right button, because of the small height of each, and the proximity of others - I share that feeling. When I have to tap on a button in that dialog, I need to be more careful than usually...

i completly agree with you.

> You can set a custom DPI in settings (gear icon) → Screen → Screen DPI. All UI elements will automatically scale along.

until a solution is found, i will adapt

poire-z:
One might try:
```diff
--- a/frontend/ui/widget/buttontable.lua
+++ b/frontend/ui/widget/buttontable.lua
@@ -59,7 +59,7 @@ function ButtonTable:init()
                 max_width = (self.width - sizer_space)/column_cnt - 2*self.sep_width - 2*self.padding,
                 bordersize = 0,
                 margin = 0,
-                padding = 0,
+                padding = Size.padding.button, -- or Size.padding.default
                 text_font_face = self.button_font_face,
                 text_font_size = self.button_font_size,
                 show_parent = self.show_parent,
```
`Size.padding.button` gives a tad more height without looking too different - `Size.padding.default` adds too much, which could be ok in that ButtonDialog, but makes all other widgets really ugly (quickdictlookup...)

(This change would not be enough as if affects some horizontal paddings and measurements, and leaves things ugly - it would need some other tweaks here and there to counteract that.)

poire-z:
Some suggestion with screenshots:
```diff
[...]

PR #5497, opened 2019-10-17
open unsupported documents as text

yparitcher:
This ~~is a working POC / hack to~~ allow opening unsupported files as txt, for example .lua files.
this can be very helpful when having oddly named files. 
it might be a good idea to allow all providers on this list?

this only triggers via the open with menu, not by directly clicking the item, unless setting the provider as a default.

this probably needs proper integration but i am not 100% sure exactly how the `DocumentRegistry` works and don't want to mess things up.

~~once this works i would like to add a way to clear already set providers from settings.~~
 
FileManager_2019-Oct-17_015418

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -293,8 +293,8 @@ function FileManager:init()
             {
                 {
                     text = _("Open with…"),
-                    enabled = lfs.attributes(file, "mode") == "file" and DocumentRegistry:getProviders(file) ~= nil
-                        and #(DocumentRegistry:getProviders(file)) > 1,
+                    enabled = lfs.attributes(file, "mode") == "file" and (DocumentRegistry:getProviders(file) == nil
+                        or #(DocumentRegistry:getProviders(file)) > 1),
                     callback = function()
                         UIManager:close(self.file_dialog)
                         self:showSetProviderButtons(file, FileManager.instance, ReaderUI)
--- a/frontend/document/documentregistry.lua
+++ b/frontend/document/documentregistry.lua
@@ -56,8 +56,16 @@ end
 function DocumentRegistry:hasProvider(file)
     local filename_suffix = string.lower(util.getFileNameSuffix(file))
 
-    if self.filetype_provider[filename_suffix] then
+    local filetype_provider = G_reader_settings:readSetting("provider") or {}
+    if self.filetype_provider[filename_suffix] or filetype_provider[filename_suffix] then
         return true
+    end
+    local DocSettings = require("docsettings")
+    if DocSettings:hasSidecarFile(file) then
+        local doc_settings_provider = DocSettings:open(file):readSetting("provider")
+        if doc_settings_provider then
+            return true
+        end
     end
     return false
 end
@@ -96,6 +104,12 @@ function DocumentRegistry:getProvider(file)
 
         -- highest weighted provider
         return providers[1].provider
+    else
+        for _, provider in ipairs(self.providers) do
+            if provider.extension == "txt" then
+                return provider.provider
+            end
+        end
     end
 end
 
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -402,8 +402,7 @@ function FileChooser:showSetProviderButtons(file, filemanager_instance, reader_u
             provider = provider.provider
             table.insert(radio_buttons, {
                 {
-                    -- @translators %1 is the provider name, such as Cool Reader Engine or MuPDF.
-                    text = T(_("%1 ~Unsupported"), provider.provider_name),
+                    text = provider.provider_name,
                     checked = DocumentRegistry:getProvider(file) == provider,
                     provider = provider,
                 },
@@ -413,7 +412,8 @@ function FileChooser:showSetProviderButtons(file, filemanager_instance, reader_u
         local provider = DocumentRegistry:getProvider(file)
         table.insert(radio_buttons, {
             {
-                text = provider.provider_name .. _(" ~Unsupported"),
+                -- @translators %1 is the provider name, such as Cool Reader Engine or MuPDF.
+                text = T(_("%1 ~Unsupported"), provider.provider_name),
                 checked = true,
                 provider = provider,
             },
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -395,17 +395,27 @@ function FileChooser:showSetProviderButtons(file, filemanager_instance, reader_u
     local buttons = {}
     local radio_buttons = {}
     local providers = DocumentRegistry:getProviders(file)
-
-    for ___, provider in ipairs(providers) do
-        -- we have no need for extension, mimetype, weights, etc. here
-        provider = provider.provider
+    if providers ~= nil then
+        for ___, provider in ipairs(providers) do
+            -- we have no need for extension, mimetype, weights, etc. here
+            provider = provider.provider
+            table.insert(radio_buttons, {
+                {
+                    text = provider.provider_name,
+                    checked = DocumentRegistry:getProvider(file) == provider,
+                    provider = provider,
+                },
+            })
+        end
+    else
+        local provider = DocumentRegistry:getProvider(file)
         table.insert(radio_buttons, {
             {
-                text = provider.provider_name,
-                checked = DocumentRegistry:getProvider(file) == provider,
+                text = provider.provider_name .. _(" ~Unsupported"),
+                checked = true,
                 provider = provider,
             },
-        })
+         })
     end
 
     table.insert(buttons, {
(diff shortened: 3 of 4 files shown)


frontend/apps/filemanager/filemanager.lua
```diff
@@ -293,8 +293,8 @@ function FileManager:init()
             {
                 {
                     text = _("Open with…"),
-                    enabled = lfs.attributes(file, "mode") == "file" and DocumentRegistry:getProviders(file) ~= nil
-                        and #(DocumentRegistry:getProviders(file)) > 1,
+                    enabled = lfs.attributes(file, "mode") == "file" and (DocumentRegistry:getProviders(file) == nil
+                        or #(DocumentRegistry:getProviders(file)) > 1),
```
Frenzie: Looks like you want to remove everything after file?

frontend/ui/widget/filechooser.lua
```diff
@@ -395,17 +395,27 @@ function FileChooser:showSetProviderButtons(file, filemanager_instance, reader_u
...
+                },
+            })
+        end
+    else
+        local provider = DocumentRegistry:getProvider(file)
         table.insert(radio_buttons, {
             {
-                text = provider.provider_name,
-                checked = DocumentRegistry:getProvider(file) == provider,
+                text = provider.provider_name .. _(" ~Unsupported"),
```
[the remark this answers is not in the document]
yparitcher: can you explain me how the Template / gettext works for strings (or point me to the docs).
i could not figure it out and keep on getting it wrong.
thanks

frontend/apps/filemanager/filemanager.lua
yparitcher: as of now it only showed  `open with` when there was 2 or more providers, i am showing it also when there are 0 providers, i did not change the behavior for 1 provider (don't show), should i do different?

yparitcher:
adds reset default global provider

FileManager_2019-Oct-17_150651

frontend/apps/filemanager/filemanager.lua
Frenzie: If you allow opening any file as if it were plain text it seems somewhat inconsistent to me to restrict it to unknown filetypes; not quite sure how much sense it makes one way or the other.

yparitcher: to be more precise i am opening an unknown file with the provider for txt. (crengine). 
if it actually an epub for example crengine will treat it as an epub, if it is unstable crengine will probably crash.

if a file is only supported by one provider there is no reason to try opening it with a different one that probably wont work, and i only am adding this provider if there are no providers already so you will get a `open with` with the same option as opening it directly

Frenzie: Right, fair enough. Carry on then. ;-)

frontend/ui/widget/filechooser.lua
```diff
@@ -415,6 +426,15 @@ function FileChooser:showSetProviderButtons(file, filemanager_instance, reader_u
                 UIManager:close(self.set_provider_dialog)
             end,
         },
+        {
+            text = _("Reset Default"),
```
Frenzie:
```suggestion
            text = _("Reset default"),
```

frontend/document/documentregistry.lua
```diff
@@ -56,9 +56,17 @@ end
...
-    if self.filetype_provider[filename_suffix] then
+    local filetype_provider = G_reader_settings:readSetting("provider") or {}
+    if self.filetype_provider[filename_suffix] or filetype_provider[filename_suffix] then
         return true
     end
+    local DocSettings = require("docsettings")
+    if DocSettings:hasSidecarFile(file) then
+        local doc_settings_provider = DocSettings:open(file):readSetting("provider")
+        if doc_settings_provider then
+            return true
```
poire-z: Was going to suggest you should properly do a `docsetting:close()` like it is done in many other places (eg. https://github.com/koreader/koreader/pull/5384).
But no, letting the table disappear when out of scope, as done here, is the right thing to do.
These other places, by calling `:close()`, do unnecessarily flush/save the unmodified data to disk...

frontend/ui/widget/filechooser.lua
poire-z: Might want to have the buttons on 2 lines, eg.:
<kbd>image</kbd>

frontend/ui/widget/filechooser.lua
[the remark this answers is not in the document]
Frenzie: gettext isn't involved, but the docs I wrote for gettext should include some examples with it.

tl;dr Concatenation like this removes a translator's ability to freely position a word like unsupported.

It's explained [over here](https://github.com/koreader/koreader/blob/7de022a61742f42224a046b604202e3a8e150ed7/doc/Building.md#translations) in a part of the README that probably should be moved elsewhere. (Pinging @pazos.) It's also explained [here](http://koreader.rocks/doc/modules/ffi.util.html#template) but obviously that file is still a bit of a stub documentation-wise.

```
-- @translators %1 is the provider name, in this case Cool Reader Engine
T(_("%1 ~Unsupported"), provider_name)
```

That being said, unless one can also force MuPDF it actually seems unnecessarily complex to use the template function here.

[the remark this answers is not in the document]
yparitcher: i did not want to hardcode a specific provider, to make it future proof

frontend/ui/widget/filechooser.lua
yparitcher: i find two lines harder to press, personally

yparitcher: also that would require reworking the whole `open as` widget which i am trying to leave as intact as possible.

frontend/ui/widget/filechooser.lua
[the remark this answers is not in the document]
yparitcher: also the word unsupported was a placeholder is there maybe a better word to convey the intent?

yparitcher:
can somone test and make sure i didn't mess up `DocumentRegistry` too much.
thanks

frontend/ui/widget/filechooser.lua
poire-z: > i find two lines harder to press, personally

Yes, me too. See https://github.com/koreader/koreader/issues/5449#issuecomment-537935889. And we should try to get that better.

But for consistency, and because, if these 3 text fit without truncation in english, they may get truncated in german or in some other language, so it should be on 3 lines.

> also that would require reworking the whole 

No, it should be easy.
Instead of 
```lua
table.insert(buttons, { -- line 1
 {button1}, -- columns
 {button2},
 {button3},
})
```
use
```lua
table.insert(buttons,
 { -- line1
  {button1}, -- columns
  {button2},
 },
 { -- line1
  {button3},
 }
)
```

(Not familiar enough with DocumentRegistry to comment more on this PR.)

poire-z: Or better, if you can detect the current value is different than the default, just add that 2nd row with the `Use default` when it is - so in normal conditions, we only see a single row with the 2 buttons.
```lua
table.insert(buttons, { -- line 1
 {button1}, -- columns
 {button2},
})
if there_is_setting_other_than_default then
    table.insert(buttons, { -- line 2
     {button3},
    })
end
```

yparitcher: fixed

yparitcher:
FileManager_2019-Oct-18_140005
FileManager_2019-Oct-18_140020

frontend/ui/widget/filechooser.lua
```diff
@@ -394,18 +394,30 @@ function FileChooser:showSetProviderButtons(file, filemanager_instance, reader_u
...
+                },
+            })
+        end
+    else
+        local provider = DocumentRegistry:getProvider(file)
         table.insert(radio_buttons, {
             {
-                text = provider.provider_name,
-                checked = DocumentRegistry:getProvider(file) == provider,
+                text = provider.provider_name .. _(" ~Unsupported"),
```
Frenzie: It should still read something like:
```suggestion
                T(_("%1 ~Unsupported"), provider.provider_name),
```

yparitcher: sorry i cahnged the wrong one last time. Fixed

frontend/ui/widget/filechooser.lua
```diff
@@ -394,18 +394,30 @@ function FileChooser:showSetProviderButtons(file, filemanager_instance, reader_u
...
             {
-                text = provider.provider_name,
-                checked = DocumentRegistry:getProvider(file) == provider,
+                -- @translators %1 is the provider name, such as Cool Reader Engine or MuPDF.
+                text = T(_("%1 ~Unsupported"), provider.provider_name),
+                checked = true,
                 provider = provider,
             },
-        })
+         })
```
Frenzie:
```suggestion
        })
```

Frenzie: Looks OK to me besides this tiny nitpick, were you done with it?

yparitcher: yes

Frenzie:
Alright, thanks!

PR #5503, opened 2019-10-18
TextWidget: small refactoring, better handle max_width

poire-z:
Lots of code was doing some renderText calls to get the size of some text string, and truncate it to some width if needed, with or without an added ellipsis, before instantiating a TextWidget with that tweaked text string.

This PR fixes/adds some properties and methods to TextWidget so all that can be done by it. It makes the caller codes simpler, as they don't need to use RenderText directly.
(Additionally, if we ever go at using Harfbuzz for text rendering, we'll just have to update or replace textwidget.lua without the need to update any higher level code.)

RenderText: removed the space added by truncateTextByWidth after the ellipsis, as it doesn't feel needed, and break right alignment of the ellipsis with other texts.
KeyValuePage: fix some subtle size and alignment issues.
NumberPickerWidget: fix font size (provided font size was not used)

Menu.lua: I did not touch the part of the code that #5496 affects, so these 2 PRs should be mergeable in any order. If needed, I'll update what can be from #5496 to follow the spirit of this PR.
Pinging @robert00s for info, for your future widgets :)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -50,10 +50,6 @@ local function restoreScreenMode()
     end
 end
 
-local function truncatePath(text)
-    return FileChooser:truncatePath(text)
-end
-
 local FileManager = InputContainer:extend{
     title = _("KOReader"),
     root_path = lfs.currentdir(),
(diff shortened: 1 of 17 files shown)


robert00s:
@poire-z 
Sorry for delay. I'll try to look at it today evening or tomorrow.

poire-z:
(Just rebased to current master.)
@Frenzie : any comment on this one?

PR #5506, opened 2019-10-19
[chore] Add translations to repo

Frenzie:
Apparently Weblate needs the translations in the repo before it can be set up.

References <https://github.com/koreader/koreader/issues/3754>.

Also see <https://github.com/koreader/koreader/issues/2040#issuecomment-485154051>.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,9 +31,6 @@ luacov.stats.out
 trace-out.txt
 koreader-*
 
-l10n/*
-!l10n/Makefile
-!l10n/README.md
 i18n
 
 /.cproject
--- a/l10n/af_ZA/koreader.po
+++ b/l10n/af_ZA/koreader.po
@@ -1,0 +1,8998 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# 
+# Translators:
+msgid ""
+msgstr ""
+"Project-Id-Version: KOReader\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2019-10-18 21:47+0200\n"
+"PO-Revision-Date: 2019-10-18 19:44+0000\n"
+"Last-Translator: Frans de Jonge <fransdejonge@gmail.com>\n"
+"Language-Team: Afrikaans (South Africa) (http://www.transifex.com/houqp/koreader/language/af_ZA/)\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: af_ZA\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#: reader.lua:131
+msgid ""
+"Cannot open last file.\n"
+"This could be because it was deleted or because external storage is still being mounted.\n"
+"Do you want to retry?"
+msgstr ""
+
+#: reader.lua:194
+msgid ""
+"Documents will be rendered in color on this device.\n"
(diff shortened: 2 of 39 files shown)


poire-z:
Just for setup? Or will they stay in the main koreader repo?
And how that will work?
Will we get a bunch of commits each time someone translates something :( ?
Or you'll pull the updated translations only when you're about to release? And how about/if people update translations directly on github/via PR?

Frenzie:
It lives in the repo, yes, but it does [lazy commits](https://docs.weblate.org/en/latest/admin/continuous.html#lazy-commit). That looks like this:

* Small adjustment: https://github.com/phpmyadmin/phpmyadmin/commit/b73733f7a7526023db426f79d5423b99f37a76ca
* Big translation session: https://github.com/phpmyadmin/phpmyadmin/commit/70281956bbb5459c01c9d483409e30035e33fe9a

If you feel strongly about it we could consider a submodule instead, but the Git tracking in and of itself is a definite advantage.

> And how about/if people update translations directly on github/via PR?

I may very well switch to doing exactly that. ;-)

poire-z:
> If you feel strongly about it

I do :|
If I add a string in some dev commit, I'll be seeing dozens of small adjustment commits (one per author/language) following it, and it will be a pain to follow github commits page (1 or 2 real dev commit per page, and 28 translation commits) or use `git log -p` to see what really happened over a period of time (to look for cause of regressions, and for me to look about what's new when adjusting my own patch once a week).

> we could consider a submodule instead

But you'll be bumping it every day into the main repo to build a nightly?
If no, that's fine with me :)
If yes, well, only 1 translation commit per day among 2 to 5 real dev commits. That's better, but still :)

> but the Git tracking in and of itself is a definite advantage.

OK for git tracking in its own repo, but what advantage linking it to the main repo?

Frenzie:
> edit: oh, I already hate that last merged commit :(

We could force push master; apologies I thought we'd basically discussed this about a year and a half ago. Force push to replace this commit with a submodule okay by you @NiLuJe @pazos?

NiLuJe:
No objections to either approach here ;).

I would marginally prefer if the commits came from a bot account instead of a real one, but that's basically cosmetics to me ;).

poire-z:
(OK for me ! and thanks for considering that :)
Given no activity/chat happened since that commit 5 hours ago, go for it before people wake up and git pull :)

Frenzie:
I created https://github.com/koreader/koreader-translations

> But you'll be bumping it every day into the main repo to build a nightly?

The nightly build should pull in the latest translations but it doesn't need to be bumped (daily) for that.

@NiLuJe 
> I would marginally prefer if the commits came from a bot account instead of a real one, but that's basically cosmetics to me ;).

It comes from the weblate bot?

Frenzie:
Okay, it's done.

poire-z:
Thanks.
(No pb with that: not sure why, but all our opened PRs got their last update time updated.)

poire-z:
https://gitlab.com/koreader/koreader has updated in between. Will it be ok with the force push, or does it need manual tweaking? (Is it what is used for building nightlies?)

Frenzie:
We'll see, I have no idea. But yeah, that's why you shouldn't force push. ^_^

master is protected from accidental force pushes by GH btw. I disabled that protection to do it, or rather I temporarily switched the protection to the imaginary `masterr` branch instead.

Frenzie:
The nightly uses GH: https://gitlab.com/koreader/nightly-builds/blob/29ed1f3f781434827cc18849431852d41ea2eb41/build_release.sh#L18

It used to be GitLab but apparently there was a mirroring issue at some point: https://gitlab.com/koreader/nightly-builds/commit/9cc1fec7d9bfbf928c950f422ad3cd365590e140

Frenzie:
Yep, that broke the mirroring. I'm not sure if there's an easy way to fix that besides manually force pushing or deleting the repo and starting the mirror from scratch.

Issue #5037, opened 2019-05-22
Feature request: keep "up a level" button always visible in file browser


Thra11:
* KOReader version: 2019.05
* Device: Kobo Aura One

#### Issue
When using the file browser with the display mode set to (for example) "mosaic with cover images", the entry to move up to the directory above the current one ("..") appears in the mosaic as the first item. As you only get nine items per screen in mosaic mode, this can mean that if you have a lot of files or subdirectories, you can be many pages away from the up button. Would it make sense to have the 'up' button as a smaller, always-visible button at the top of the screen, like the home and '+' buttons?

I'm aware that I can configure a tap or gesture to go up a directory, but I think a more obvious solution by default might be beneficial.

poire-z:
> Would it make sense to have the 'up' button as a smaller, always-visible button at the top of the screen, like the home and '+' buttons?

Dunno about having another 'up' button always shown, but I guess we could add "Go up" into the '+' button menu (so, 2 taps to go up, instead of many swipes - for those who don't know/haven't set a gesture).

cramoisi:
I often find myself searching for this button ; it would be cool to have a way for a fast up

Frenzie:
I think it's fairly normal in file managers.

Thunar:
![Screenshot_2019-05-23_15-55-19](https://user-images.githubusercontent.com/202757/58258514-2e1eeb00-7d73-11e9-8cc8-d75cbec30fd9.png)

Nemo:
![Screenshot_2019-05-23_15-56-04](https://user-images.githubusercontent.com/202757/58258579-4b53b980-7d73-11e9-8783-b2de95f74a41.png)

But of course an alternative or complementary approach (here Dolphin) is the breadcrumbs.
![Screenshot_2019-05-23_15-57-11](https://user-images.githubusercontent.com/202757/58258703-8229cf80-7d73-11e9-9870-bf67c3e82eeb.png)

You don't really need a separate up if you can click parts of the path.

robert00s:
One problem I see. Tap on folder shortcut (eg `frans` or `home` like in Dolphin) could make conflict with tap to open menu :(

Frenzie:
I'm not sure if that's a problem without having tried it. Not in the technical sense (it will overlap and it would need to take precedence) but in the usability sense. If it is, it could always be hold only in combination with tap to show menu.

melyux:
We can put the "up" button at the bottom left corner like we do on KV pages.

hius07:
[...]

PR #5508, opened 2019-10-20
UI Changes

dontcrash:
Have made several changes to the UI to improve the experience and make it less "clunky" and more polished/refined, please let me know what you think

### DIFF
--- a/frontend/apps/filemanager/filemanagerutil.lua
+++ b/frontend/apps/filemanager/filemanagerutil.lua
@@ -25,7 +25,7 @@ end
 
 function filemanagerutil.abbreviate(path)
     if not path then return "" end
-    if G_reader_settings:isTrue("shorten_home_dir") then
+    if G_reader_settings:nilOrTrue("shorten_home_dir") then
         local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
         if path == home_dir then
             return _("Home")
--- a/frontend/apps/filemanager/filemanagerutil.lua
+++ b/frontend/apps/filemanager/filemanagerutil.lua
@@ -6,7 +6,6 @@ local Device = require("device")
 local DocSettings = require("docsettings")
 local util = require("ffi/util")
 local _ = require("gettext")
-local m_util = require("util")
 
 local filemanagerutil = {}
 
@@ -29,7 +28,7 @@ function filemanagerutil.abbreviate(path)
     if G_reader_settings:isTrue("shorten_home_dir") then
         local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
         if path == home_dir then
-            return home_dir:sub(m_util.lastIndexOf(home_dir, "/")+1)
+            return _("Home")
         end
         local len = home_dir:len()
         local start = path:sub(1, len)
--- a/frontend/apps/filemanager/filemanagerutil.lua
+++ b/frontend/apps/filemanager/filemanagerutil.lua
@@ -6,7 +6,7 @@ local Device = require("device")
 local DocSettings = require("docsettings")
 local util = require("ffi/util")
 local _ = require("gettext")
-local util = require("util")
+local m_util = require("util")
 
 local filemanagerutil = {}
 
@@ -29,8 +29,7 @@ function filemanagerutil.abbreviate(path)
     if G_reader_settings:isTrue("shorten_home_dir") then
         local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
         if path == home_dir then
-            return home_dir:sub(util.lastIndexOf(home_dir, "/")+1)
-            --return _("Home")
+            return home_dir:sub(m_util.lastIndexOf(home_dir, "/")+1)
         end
         local len = home_dir:len()
         local start = path:sub(1, len)
--- a/frontend/apps/filemanager/filemanagerutil.lua
+++ b/frontend/apps/filemanager/filemanagerutil.lua
@@ -5,6 +5,8 @@ This module contains miscellaneous helper functions for FileManager
 local Device = require("device")
 local DocSettings = require("docsettings")
 local util = require("ffi/util")
+local _ = require("gettext")
+local util = require("util")
 
 local filemanagerutil = {}
 
@@ -24,13 +26,16 @@ end
 
 function filemanagerutil.abbreviate(path)
     if not path then return "" end
-    local home_dir_name = G_reader_settings:readSetting("home_dir_display_name")
-    if home_dir_name ~= nil then
+    if G_reader_settings:isTrue("shorten_home_dir") then
         local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
+        if path == home_dir then
+            return home_dir:sub(util.lastIndexOf(home_dir, "/")+1)
+            --return _("Home")
+        end
         local len = home_dir:len()
         local start = path:sub(1, len)
         if start == home_dir then
-            return home_dir_name .. path:sub(len+1)
+            return path:sub(len+1)
         end
     end
     return path
--- a/frontend/apps/filemanager/filemanagerutil.lua
+++ b/frontend/apps/filemanager/filemanagerutil.lua
@@ -33,7 +33,7 @@ function filemanagerutil.abbreviate(path)
         local len = home_dir:len()
         local start = path:sub(1, len)
         if start == home_dir then
-            return path:sub(len+1)
+            return path:sub(len+2)
         end
     end
     return path
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -56,7 +56,7 @@ if Device:setDateTime() then
         text = _("Time and date"),
         sub_item_table = {
           {
-            text = _("12 Hour time"),
+            text = _("12 hour clock"),
             keep_menu_open = true,
             checked_func = function()
                 return G_reader_settings:isTrue("twelve_hour_time")
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -59,14 +59,10 @@ if Device:setDateTime() then
             text = _("12-hour clock"),
             keep_menu_open = true,
             checked_func = function()
-                return G_reader_settings:isTrue("twelve_hour_clock")
+                return G_reader_settings:nilOrTrue("twelve_hour_clock")
             end,
             callback = function()
-                if G_reader_settings:isTrue("twelve_hour_clock") then
-                    G_reader_settings:flipFalse("twelve_hour_clock")
-                else
-                    G_reader_settings:flipTrue("twelve_hour_clock")
-                end
+                G_reader_settings:flipNilOrTrue("twelve_hour_clock")
             end,
             },
             {
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -55,6 +55,21 @@ if Device:setDateTime() then
     common_settings.time = {
         text = _("Time and date"),
         sub_item_table = {
+          {
+            text = _("12 Hour time"),
+            keep_menu_open = true,
+            checked_func = function()
+                return G_reader_settings:isTrue("twelve_hour_time")
+            end,
+            callback = function()
+              local twelve_hr = G_reader_settings:isTrue("twelve_hour_time")
+              if twelve_hr then
+                G_reader_settings:flipFalse("twelve_hour_time")
+              else
+                G_reader_settings:flipTrue("twelve_hour_time")
+              end
+            end,
+            },
             {
                 text = _("Set time"),
                 keep_menu_open = true,
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -60,14 +60,14 @@ if Device:setDateTime() then
             -- does the menu need to be kept open?
             keep_menu_open = true,
             checked_func = function()
-                return G_reader_settings:isTrue("twelve_hour_time")
+                return G_reader_settings:isTrue("twelve_hour_clock")
             end,
             callback = function()
-              local twelve_hr = G_reader_settings:isTrue("twelve_hour_time")
+              local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
               if twelve_hr then
-                G_reader_settings:flipFalse("twelve_hour_time")
+                G_reader_settings:flipFalse("twelve_hour_clock")
               else
-                G_reader_settings:flipTrue("twelve_hour_time")
+                G_reader_settings:flipTrue("twelve_hour_clock")
               end
             end,
             },
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -56,7 +56,7 @@ if Device:setDateTime() then
         text = _("Time and date"),
         sub_item_table = {
           {
-            text = _("12 hour clock"),
+            text = _("12-hour clock"),
             -- does the menu need to be kept open?
             keep_menu_open = true,
             checked_func = function()
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -57,7 +57,6 @@ if Device:setDateTime() then
         sub_item_table = {
           {
             text = _("12-hour clock"),
-            -- does the menu need to be kept open?
             keep_menu_open = true,
             checked_func = function()
                 return G_reader_settings:isTrue("twelve_hour_clock")
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -63,12 +63,12 @@ if Device:setDateTime() then
                 return G_reader_settings:isTrue("twelve_hour_clock")
             end,
             callback = function()
-              local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
-              if twelve_hr then
-                G_reader_settings:flipFalse("twelve_hour_clock")
-              else
-                G_reader_settings:flipTrue("twelve_hour_clock")
-              end
+                local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
+                if twelve_hr then
+                    G_reader_settings:flipFalse("twelve_hour_clock")
+                else
+                    G_reader_settings:flipTrue("twelve_hour_clock")
+                end
             end,
             },
             {
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -57,6 +57,7 @@ if Device:setDateTime() then
         sub_item_table = {
           {
             text = _("12 hour clock"),
+            -- does the menu need to be kept open?
             keep_menu_open = true,
             checked_func = function()
                 return G_reader_settings:isTrue("twelve_hour_time")
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -62,8 +62,7 @@ if Device:setDateTime() then
                 return G_reader_settings:isTrue("twelve_hour_clock")
             end,
             callback = function()
-                local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
-                if twelve_hr then
+                if G_reader_settings:isTrue("twelve_hour_clock") then
                     G_reader_settings:flipFalse("twelve_hour_clock")
                 else
                     G_reader_settings:flipTrue("twelve_hour_clock")
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -899,6 +899,9 @@ function Menu:updatePageInfo(select_number)
         end
         -- update page information
         self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
+        if self.page == self.page_num then
+          self.page_info_text:setText("");
+        end
         self.page_info_left_chev:showHide(self.page_num > 1)
         self.page_info_right_chev:showHide(self.page_num > 1)
         self.page_info_first_chev:showHide(self.page_num > 2)
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -898,8 +898,9 @@ function Menu:updatePageInfo(select_number)
             self.selected = { x = 1, y = select_number }
         end
         -- update page information
-        self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
-        if self.page == self.page_num then
+        if self.page_num > 1 then
+          self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
+        else
           self.page_info_text:setText("");
         end
         self.page_info_left_chev:showHide(self.page_num > 1)
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -898,6 +898,7 @@ function Menu:updatePageInfo(select_number)
             self.selected = { x = 1, y = select_number }
         end
         -- update page information
+        -- don't show page x of x if there is only one page
         if self.page_num > 1 then
           self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
         else
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -602,11 +602,16 @@ function TouchMenu:updateItems()
     self.page_info_right_chev:showHide(self.page_num > 1)
     self.page_info_left_chev:enableDisable(self.page > 1)
     self.page_info_right_chev:enableDisable(self.page < self.page_num)
-    local time_info_txt = os.date("%H:%M").." @ "
-    if Device:getPowerDevice():isCharging() then
-        time_info_txt = time_info_txt.."+"
+
+    local time_info_txt = os.date("%H:%M").." - "
+    local twelve_hr = G_reader_settings:isTrue("twelve_hour_time")
+    if twelve_hr then
+      time_info_txt = os.date("%I:%M %p").." - "
     end
     time_info_txt = time_info_txt..Device:getPowerDevice():getCapacity().."%"
+    if Device:getPowerDevice():isCharging() then
+        time_info_txt = time_info_txt .. " ⚡"
+    end
     self.time_info:setText(time_info_txt)
 
     -- recalculate dimen based on new layout
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -611,7 +611,7 @@ function TouchMenu:updateItems()
     else
         time_info_txt = os.date("%H:%M")
     end
-    time_info_txt = " – " .. time_info_txt..Device:getPowerDevice():getCapacity().."%"
+    time_info_txt = time_info_txt .. " – " .. Device:getPowerDevice():getCapacity().."%"
     if Device:getPowerDevice():isCharging() then
         time_info_txt = time_info_txt .. " ⚡"
     end
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -605,10 +605,11 @@ function TouchMenu:updateItems()
     self.page_info_left_chev:enableDisable(self.page > 1)
     self.page_info_right_chev:enableDisable(self.page < self.page_num)
 
-    local time_info_txt = os.date("%H:%M").." – "
-    local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
-    if twelve_hr then
+    local time_info_txt
+    if G_reader_settings:isTrue("twelve_hour_clock") then
         time_info_txt = os.date("%I:%M %p").." – "
+    else
+        time_info_txt = os.date("%H:%M").." – "
     end
     time_info_txt = time_info_txt..Device:getPowerDevice():getCapacity().."%"
     if Device:getPowerDevice():isCharging() then
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -605,7 +605,7 @@ function TouchMenu:updateItems()
     self.page_info_right_chev:enableDisable(self.page < self.page_num)
 
     local time_info_txt
-    if G_reader_settings:isTrue("twelve_hour_clock") then
+    if G_reader_settings:nilOrTrue("twelve_hour_clock") then
         time_info_txt = os.date("%I:%M %p")
     else
         time_info_txt = os.date("%H:%M")
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -594,6 +594,7 @@ function TouchMenu:updateItems()
     table.insert(self.item_group, self.footer_top_margin)
     table.insert(self.item_group, self.footer)
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
+    -- don't show page x of x if there is only one page
     if self.page_num > 1 then
       self.page_info_text:setText(util.template(_("Page %1 of %2"), self.page, self.page_num))
     else
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -595,9 +595,9 @@ function TouchMenu:updateItems()
     table.insert(self.item_group, self.footer)
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
     if self.page_num > 1 then
-      self.page_info_text.text = util.template(_("Page %1 of %2"), self.page, self.page_num)
+      self.page_info_text:setText(util.template(_("Page %1 of %2"), self.page, self.page_num))
     else
-      self.page_info_text:setText("");
+      self.page_info_text:setText("")
     end
     self.page_info_left_chev:showHide(self.page_num > 1)
     self.page_info_right_chev:showHide(self.page_num > 1)
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -596,9 +596,9 @@ function TouchMenu:updateItems()
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
     -- Don't show page x of x if there is only one page.
     if self.page_num > 1 then
-      self.page_info_text:setText(util.template(_("Page %1 of %2"), self.page, self.page_num))
+        self.page_info_text:setText(util.template(_("Page %1 of %2"), self.page, self.page_num))
     else
-      self.page_info_text:setText("")
+        self.page_info_text:setText("")
     end
     self.page_info_left_chev:showHide(self.page_num > 1)
     self.page_info_right_chev:showHide(self.page_num > 1)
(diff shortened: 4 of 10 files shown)


frontend/ui/widget/menu.lua
```diff
@@ -899,6 +899,9 @@ function Menu:updatePageInfo(select_number)
         end
         -- update page information
         self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
+        if self.page == self.page_num then
+          self.page_info_text:setText("");
```
poire-z: In file browser, when on the last page, you'd see blank (between arrow buttons) instead of `Page 3 of 3`.
Probably better that way:
```lua
if self.page_num > 1 then
    self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
else
    self.page_info_text:setText("");
end
```

frontend/ui/widget/touchmenu.lua
```diff
@@ -595,6 +595,9 @@ function TouchMenu:updateItems()
     table.insert(self.item_group, self.footer)
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
     self.page_info_text.text = util.template(_("Page %1 of %2"), self.page, self.page_num)
+    if self.page == self.page_num then
+      self.page_info_text:setText("");
```
poire-z: Not sure why, but in the top menu, when there are multiple pages, I don't see anymore `Page P of N`, I always see blank between arrows (eg, with crowded menu `Gear > Tap & gestures > Gesture manager > Multiswipe`)

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -132,14 +132,14 @@ function CoverBrowser:addToMainMenu(menu_items)
             separator = true,
         },
         {
-            text = _("Shorten home directory to ~"),
+            text = _("Shorten home directory to Home"),
```
poire-z: > Since the file browser can be considered "Home"

_Can be_, but not _must be_ :) So don't _must_ me!
That name can be tweaked, as the setting is just the value to be shown.
I don't mind replacing the default with `Home`, but you should allow me to use `~/`. (OK, one could manually hack it into its settings.reader.lua - and if one has already checked that box, it will stay with `~` - until one play again with that checkbox.)
So, this should be some InputText (so no checkbox? or a checkbox for callback - and a hold_callback to set the name to use)?
Also, `text = _("Shorten home directory"),` , or include the setting value.

And "Home" should be translatable.
And I dunno. It looks nice in english with Home (`Home/Wikipedia/Science`) - may be less with some french translation (`Accueil/Wikipedia/Science`), and probably bothering in languages with long words (`MyLilleHomeInThePrairie/Wikipedia/Science`).

poire-z:
(Tested on my device.) I thought I'd mind most of the changes, but no :)
(Our file browser icon is a bit ugly (hard to recognise it shows files :) - so I don't mind having a house now.)

But regarding your comments on gitter:

> I don’t believe people would have a problem with replacing ~ with Home, the people I have talked to about that, they only have that setting turned on because it’s unpleasant to look at /mnt/onboard/path, you have to remember who the target audience is for this, it’s not just programmer

As you see, I do have a problem with that (but solvable :) see my comment).
And you don't have to remember me/us who the target audience is. My primary target audience is... myself :) and the secondary audience probably mostly technical people (given that it needs a bit of skills to be installed on a device).
And I'm personally not interested in simplifying stuff to make it easier for that third audience you're talking about :)

> a lot of people (myself included) love KOReader as it’s a better experience, but think that the UI is clunky, no offence, it works great but just visually, some things could be improved 

No offence taken. But I think we have done most things right (and some really more cleverly than other software, even for non technical users) - the alternative being to kill or hide features and be limited  just like the other readers software.
(I'm still persuaded people don't like our UI just because we use a Sans font, unlike Kobo Nickel which has a nice Serif font - but I like our Sans font.)

frontend/ui/widget/menu.lua
dontcrash: Absolutely agree, will update my code now, didn't think of that!

frontend/ui/widget/touchmenu.lua
dontcrash: Can be fixed by applying the same solution as menu.lua if page_num > 1, will update code

plugins/coverbrowser.koplugin/main.lua
dontcrash: This change I'm impartial with, I like Home personally but it would be nice to allow the user to set the text, e.g. ~ or Home, I will have to do some more digging to figure out how to do that as I am still learning how all the code for KOReader is structured, haven't worked with LUA before either

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -55,6 +55,21 @@ if Device:setDateTime() then
     common_settings.time = {
         text = _("Time and date"),
         sub_item_table = {
+          {
+            text = _("12 Hour time"),
```
Frenzie: Clock would be better than time.

dontcrash: Agreed

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,25 @@ function TouchMenu:updateItems()
     table.insert(self.item_group, self.footer_top_margin)
     table.insert(self.item_group, self.footer)
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
-    self.page_info_text.text = util.template(_("Page %1 of %2"), self.page, self.page_num)
+    if self.page_num > 1 then
+      self.page_info_text.text = util.template(_("Page %1 of %2"), self.page, self.page_num)
```
poire-z: To be future-proof (I was seeing blank because I had #5503 in), can you use setText() like you did in the other?
`self.page_info_text:setText(util.template(_("Page %1 of %2"), self.page, self.page_num))`

dontcrash: Done, that line of code wasn't mine however :)

poire-z:
2 things noticed while re-testing:

The house icon replacing the filebrowser icon is not right: that icon takes you to File browser, but to the directory that contain the book you were reading. It does _not_ take you to your home directory. And I don't want it to take me to my home directory: the current target directory is right.
So, a Home icon is confusing: the home icon at top left of file browser takes you to your home directory - having that same icon in reader take you elsewhere is confusing.

About the `~/` vs `Home/`: may be we could just show nothing when we are under the home directory.
So, we'll just see blank in home dir, and `Wikipedia/Science/` when in a sub-directory, without a leading slash.
When in a directory outside the home directory, we'll see a leading slash /.
That way, no translation involved, and even shorter.
(If really needed, have a "Home" translatable string for when we're just in that home directory, to replace the blank just in that case.)

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,25 @@ function TouchMenu:updateItems()
...
+    end
     self.page_info_left_chev:showHide(self.page_num > 1)
     self.page_info_right_chev:showHide(self.page_num > 1)
     self.page_info_left_chev:enableDisable(self.page > 1)
     self.page_info_right_chev:enableDisable(self.page < self.page_num)
-    local time_info_txt = os.date("%H:%M").." @ "
-    if Device:getPowerDevice():isCharging() then
-        time_info_txt = time_info_txt.."+"
+
+    local time_info_txt = os.date("%H:%M").." - "
```
Frenzie: Btw, I'd prefer to use something else as a separator, an en-dash (–) would be similar but nicer.

dontcrash: Done, it does look nicer

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -57,6 +57,7 @@ if Device:setDateTime() then
         sub_item_table = {
           {
             text = _("12 hour clock"),
+            -- does the menu need to be kept open?
```
poire-z: I guess yes - it should allow you to see the change in real time, doesn't it?

dontcrash: That's the only reason I left that there, but I wasn't sure if it was necessary to flash the screen, it does look nice though to see the change happen I guess

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -139,7 +139,7 @@ function CoverBrowser:addToMainMenu(menu_items)
                     G_reader_settings:delSetting("home_dir_display_name")
                     if FileManager.instance then FileManager.instance:reinit() end
                 else
-                    G_reader_settings:saveSetting("home_dir_display_name", "Home")
+                    G_reader_settings:saveSetting("home_dir_display_name", _("Home"))
```
poire-z: I don't think we should save to settings a translatable string (if I switch language from English to French, I expect to see "Accueil" and not the value "Home" it has when i was using English).

I think this change will need a little more work than what you intended to do :|
If we are going to use the main translatable string for "Home", we should probably just save the boolean `true` here - and allow for setting it to a string like `"~"` for users who want something else.
Which means some more work on the code that abbreviate the path to use that string, or the translatable `_("Home")`, depending on the setting type.

(Or simplify all that but not using any prefix, but needs comments on my other comment :)

dontcrash: I’ll look into this more, I think not all users would want Home, I’ll see if I can figure out how to let users input a string for it instead, then it could be a simple “Shorten home directory”, click that and it brings up a text field or something of the sorts

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -55,6 +55,22 @@ if Device:setDateTime() then
     common_settings.time = {
         text = _("Time and date"),
         sub_item_table = {
+          {
+            text = _("12 hour clock"),
```
Frenzie:
```suggestion
            text = _("12-hour clock"),
```

dontcrash: Will fix shortly

Frenzie:
Gitter convo [here](https://gitter.im/koreader/koreader?at=5dab199da28cd16529c52410).

plugins/coverbrowser.koplugin/main.lua
poire-z: Don't rush into adding a InputDialog. It might be a little overkill for such a little setting.
People could and still can override it by editing manually the setting into their settings.reader.lua.
I realize that currently, we too are overriding the user specific setting when toggling that menu item, with a `~`. So, you were just changeing the overriding value from `~` to `Home`.

What do you think about my suggestion of going with nothing, from a UI/third audience point of view?
You would display `_("Home")` only when in home_dir, and a relative path (without leading /) when in a sub directory (without a `Home/` prefix).
(and when this checkbox is unchecked, we would display always the full path).

I would be fine with that, and it would make things simpler - just a boolean setting (while still allowing people to use a string to get something else because it's currently possible).

dontcrash: Actually yes, I like that idea, _("Home") for the Home and / for others, I'm keen, don't have time to write that right now as it's getting late here, unless you have time? :D

poire-z: I can have a look. @Frenzie : any opinion on going with that (also explained in https://github.com/koreader/koreader/pull/5508#issuecomment-544232071).

Frenzie: On Kobo that could make a lot of sense, less sure about other platforms.

One thing to keep in mind is that not displaying a home prefix would be antithetical to breadcrumbs, also see <https://github.com/koreader/koreader/issues/5037#issuecomment-495230466>.

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -1,5 +1,6 @@
 local InputContainer = require("ui/widget/container/inputcontainer")
 local UIManager = require("ui/uimanager")
+local InputDialog = require("ui/widget/inputdialog")
```
Frenzie: This would have to go above UIManager, see <https://github.com/koreader/koreader-base/wiki/Coding-style#order-of-requires>.

(More of a generic comment; not entirely sure about this commit in and of itself.)

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -132,15 +133,43 @@ function CoverBrowser:addToMainMenu(menu_items)
...
-                    G_reader_settings:delSetting("home_dir_display_name")
-                    if FileManager.instance then FileManager.instance:reinit() end
+                  G_reader_settings:delSetting("home_dir_display_name")
+                  if FileManager.instance then FileManager.instance:reinit() end
                 else
-                    G_reader_settings:saveSetting("home_dir_display_name", _("Home"))
-                    if FileManager.instance then FileManager.instance:reinit() end
+                  local home_path_dialog
+                  home_path_dialog = InputDialog:new{
+                    title = _("Home directory name"),
```
Frenzie: Also generic, the indentation seems to have ended up as two spaces instead of four.

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -132,15 +133,43 @@ function CoverBrowser:addToMainMenu(menu_items)
...
+                    input = "",
```
Frenzie: Seems like this should be prefilled with `home_dir_display_name`? Generically speaking again.

plugins/coverbrowser.koplugin/main.lua
poire-z: So, that kept in mind :), what do we want?
I guess once a home_dir is set, users rarely goes outside (and when they will, they'll see the full path with a leading /).
But when inside my home_dir, do I need to always see the home_dir breadcrumb/anchor (whether it's `~/` or `Home/` or `🏠/`) ? I don't think so. It's noise. `Home/` is too long for me, `~/` is short and so better, but _nothing_ feels even better.
So, when navigating home_dir/authors/John Martin/Novels/, we'll see
`authors/John Martin/Novels`
or do we really need to see:
`Home/authors/John Martin/Novels` or
`~/authors/John Martin/Novels`  or
`🏠/authors/John Martin/Novels` ?

> less sure about other platforms

Why would that be different? A set home_dir is a home_dir no matter the platform :) If there's a secondary location where books are kept, well, that's outside the home_dir, so we'll display a full path. If a user is confused and have to work with book in multiple locations, the only option, like now, is to not check that option of shortening.

plugins/coverbrowser.koplugin/main.lua
dontcrash: Due to the nature of how that works, when the option is turned off the setting is deleted by default, so the string would be empty

Frenzie: Then the nature of how that works is wrong. ;-)

poire-z: I agree. Logically, there should be 2 settings:
- `shorten_home_dir = true/false`
- `home_dir_shortened_text = "user set text like ~/"` - and if not set, use `_("Home")`

(or that other solution that make all that a non issue :)

dontcrash: I’d be happy to write that tomorrow if Frenzie likes that idea

Frenzie: It sounds like the correct behavior, yes.

plugins/coverbrowser.koplugin/main.lua
Frenzie: > or do we really need to see:

Need not necessarily, but I think it's at best inconvenient not to. If you have an interaction pattern like breadcrumbs that normally works, breaking it to get rid of some noise for a nice clean look isn't a good thing.

Dolphin actually does something very much like what you're proposing and it's really annoying; its biggest flaw perhaps. Any direct subfolder of home doesn't display home in the breadcrumbs, completely breaking the normal `>` quick folder switching functionality. Windows Explorer/Directory Opus does it best.

But I suppose we can table my argument for the moment because there's no such thing as breadcrumbs.

I do like to know where I am though (i.e., Home, not nowhere).

> Why would that be different?

Because Kobo's /mnt/onboard path is very ugly. ;-)

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -55,6 +55,22 @@ if Device:setDateTime() then
     common_settings.time = {
         text = _("Time and date"),
         sub_item_table = {
+          {
+            text = _("12-hour clock"),
+            -- does the menu need to be kept open?
+            keep_menu_open = true,
+            checked_func = function()
+                return G_reader_settings:isTrue("twelve_hour_time")
```
Frenzie: Since the setting isn't legacy, please also change it.

```suggestion
                return G_reader_settings:isTrue("twelve_hour_clock")
```

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -55,6 +55,22 @@ if Device:setDateTime() then
...
+            end,
+            callback = function()
+              local twelve_hr = G_reader_settings:isTrue("twelve_hour_time")
```
Frenzie: Indentation.

frontend/ui/widget/menu.lua
```diff
@@ -898,7 +898,12 @@ function Menu:updatePageInfo(select_number)
             self.selected = { x = 1, y = select_number }
         end
         -- update page information
-        self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
+        -- don't show page x of x if there is only one page
+        if self.page_num > 1 then
+          self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
```
Frenzie: Indentation.

Also weren't you proposing to make it Page or was there an objection to that?

```suggestion
            self.page_info_text:setText(util.template(_("Page %1 of %2"), self.page, self.page_num))
```

frontend/ui/widget/menu.lua
```diff
@@ -898,7 +898,12 @@ function Menu:updatePageInfo(select_number)
...
+        else
+          self.page_info_text:setText("");
```
Frenzie:
```suggestion
            self.page_info_text:setText("");
```

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,26 @@ function TouchMenu:updateItems()
     table.insert(self.item_group, self.footer_top_margin)
     table.insert(self.item_group, self.footer)
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
-    self.page_info_text.text = util.template(_("Page %1 of %2"), self.page, self.page_num)
+    -- don't show page x of x if there is only one page
```
Frenzie: Did you check what xgettext thinks of this? You might have to reverse the order. Also comments like to be proper sentences.

```suggestion
    -- Don't show page x of x if there is only one page.
```

However… I'd say this isn't really the kind of code that requires much comment.

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,26 @@ function TouchMenu:updateItems()
...
+    if self.page_num > 1 then
+      self.page_info_text:setText(util.template(_("Page %1 of %2"), self.page, self.page_num))
```
Frenzie: More indentation.

plugins/coverbrowser.koplugin/main.lua
poire-z: > I do like to know where I am though

So, that's where the new translatable string `_("Home")` would be used, and only when just in home_dir (I was fine without any, but ok).
(That will be confusing if a user create a subdirectory named "Home" thus :)

plugins/coverbrowser.koplugin/main.lua
dontcrash: Consider it done, will push code tomorrow

frontend/ui/widget/menu.lua
dontcrash: In some of the UI it is “page” and in other areas it’s “Page”, will fix that up tomorrow

poire-z:
Regarding the suggestion at https://github.com/koreader/koreader/pull/5508#issuecomment-544232071 and https://github.com/koreader/koreader/pull/5508#discussion_r336775822, that's all it would need:
```diff
--- a/frontend/apps/filemanager/filemanagerutil.lua
+++ b/frontend/apps/filemanager/filemanagerutil.lua
@@ -5,6 +5,7 @@ This module contains miscellaneous helper functions for FileManager
 local Device = require("device")
 local DocSettings = require("docsettings")
 local util = require("ffi/util")
+local _ = require("gettext")

 local filemanagerutil = {}

@@ -24,13 +25,15 @@ end

 function filemanagerutil.abbreviate(path)
     if not path then return "" end
-    local home_dir_name = G_reader_settings:readSetting("home_dir_display_name")
-    if home_dir_name ~= nil then
+    if G_reader_settings:isTrue("shorten_home_dir") then
         local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
+        if path == home_dir then
+            return _("Home")
+        end
         local len = home_dir:len()
         local start = path:sub(1, len)
         if start == home_dir then
-            return home_dir_name .. path:sub(len+1)
+            return path:sub(len+2) -- remove leading / before sub-directory
         end
     end
     return path
```
Or variations depending on whether you want to keep/allow a `home_dir_display_name` setting to for customising "Home" - but I'd rather not allow that - we don't allow customising any other translatable string, so why start now? (Initially, it was a hidden/manual setting - I added that menu item to quickly unhidden it, and used `~` as the value.)
(You get your Home, I get my short display, dunno if people would still really want to see their `~/`  - even if I wanted to see it at the start of this thread :)

dontcrash:
How would you feel about the Home dir text instead of being "Home", set it to the home_dir path substring, e.g. if the home_dir was set to /mnt/onboard/Books it would display "Books"?
This would also mean no translation would be needed @poire-z @Frenzie

Frenzie:
I support @poire-z's suggestion from https://github.com/koreader/koreader/pull/5508#issuecomment-544275728 (with some reservations with regard to breadcrumbs that aren't really applicable while there are no breadcrumbs).

frontend/apps/filemanager/filemanagerutil.lua
```diff
@@ -5,6 +5,8 @@ This module contains miscellaneous helper functions for FileManager
 local Device = require("device")
 local DocSettings = require("docsettings")
 local util = require("ffi/util")
+local _ = require("gettext")
+local util = require("util")
```
Frenzie: Pssht this should break horribly.

dontcrash: :( ?

Frenzie: local *util* = require("ffi/util")
local _ = require("gettext")
local util = require("util")

dontcrash: Oh, I am an idiot, will fix that lol

frontend/apps/filemanager/filemanagerutil.lua
```diff
@@ -24,13 +26,15 @@ end
...
+    if G_reader_settings:isTrue("shorten_home_dir") then
         local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
+        if path == home_dir then
+            return home_dir:sub(m_util.lastIndexOf(home_dir, "/")+1)
+        end
         local len = home_dir:len()
         local start = path:sub(1, len)
         if start == home_dir then
-            return home_dir_name .. path:sub(len+1)
+            return path:sub(len+1)
```
poire-z: You didn't need the `len+2` from my sample code to remove the leading `/`
(We need to remove it, to distinguish shortened paths inside home_dir from full path outside home_dir.)

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -55,6 +55,21 @@ if Device:setDateTime() then
...
+                return G_reader_settings:isTrue("twelve_hour_clock")
+            end,
+            callback = function()
+                local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
+                if twelve_hr then
+                    G_reader_settings:flipFalse("twelve_hour_clock")
+                else
+                    G_reader_settings:flipTrue("twelve_hour_clock")
+                end
+            end,
```
poire-z: You can make that one line:
https://github.com/koreader/koreader/blob/761f332d47bc9ca0fd87f9425b86a87cd4a6d12d/frontend/apps/reader/modules/readerwikipedia.lua#L280-L282
Same for `shorten_home_dir` below.

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,26 @@ function TouchMenu:updateItems()
...
     self.page_info_right_chev:showHide(self.page_num > 1)
     self.page_info_left_chev:enableDisable(self.page > 1)
     self.page_info_right_chev:enableDisable(self.page < self.page_num)
-    local time_info_txt = os.date("%H:%M").." @ "
-    if Device:getPowerDevice():isCharging() then
-        time_info_txt = time_info_txt.."+"
+
+    local time_info_txt = os.date("%H:%M").." – "
+    local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
+    if twelve_hr then
```
poire-z: `if G_reader_settings:isTrue("twelve_hour_clock") then` is ok when it's that short.

poire-z:
> How would you feel about the Home dir text instead of being "Home", set it to the home_dir path substring, e.g. if the home_dir was set to /mnt/onboard/Books it would display "Books"?
This would also mean no translation would be needed 

Why not indeed.

It can be that this home directory can not always have a nice name - on Kobo, some people use a hidden directory with a leading dot so it's not indexed by Nickel - mine is `.LIBRARY`, in uppercase so it jumps out among .adds .kobo... when USB mounted on my PC - that leading dot would be a little ugly below KOReader :)
So, remove any leading dot? On Kobo only? @Frenzie ?

frontend/ui/widget/touchmenu.lua
poire-z: And make that a `if else` to avoid 2 calls to `os.date()`:
```lua
local time_info_txt
if G_reader_settings:isTrue("twelve_hour_clock") then
  time_info_txt = os.date(...)
else
  time_info_txt = os.date(...)
end
time_info_txt = time_info_txt .." – " .. Device:getPowerDevice():getCapacity().."%"
```

Frenzie:
> So, remove any leading dot? On Kobo only? @Frenzie ?

The default home dir on Kobo is `/mnt/onboard`, which would then display as `onboard`. Simply calling it `Home` would deliver a more uniform, friendlier out of the box experience.

As an aside, here's how Dolphin displays `/home/<username>`.

Screenshot_20191021_092752

dontcrash:
> > So, remove any leading dot? On Kobo only? @Frenzie ?
> 
> The default home dir on Kobo is `/mnt/onboard`, which would then display as `onboard`. Simply calling it `Home` would deliver a more uniform, friendlier out of the box experience.
> 
> As an aside, here's how Dolphin displays `/home/<username>`.
> 
> Screenshot_20191021_092752

How about if I did a check, see if a home dir had been set?
`local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()`
Could simply check if home_dir exists, if so use the new method of extracting the name from the set home folder, if it uses the default dir, display _("Home")?

frontend/ui/widget/touchmenu.lua
```diff
@@ -605,10 +605,11 @@ function TouchMenu:updateItems()
...
-    local time_info_txt = os.date("%H:%M").." – "
-    local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
-    if twelve_hr then
+    local time_info_txt
+    if G_reader_settings:isTrue("twelve_hour_clock") then
         time_info_txt = os.date("%I:%M %p").." – "
+    else
+        time_info_txt = os.date("%H:%M").." – "
     end
     time_info_txt = time_info_txt..Device:getPowerDevice():getCapacity().."%"
```
poire-z: Not as per my full suggestion :)
It jumps at me that both have `.." – "` , while having it on the next common line would make it clearer and less risky of desync.

frontend/ui/widget/menu.lua
```diff
@@ -898,7 +898,12 @@ function Menu:updatePageInfo(select_number)
             self.selected = { x = 1, y = select_number }
         end
         -- update page information
-        self.page_info_text:setText(util.template(_("page %1 of %2"), self.page, self.page_num))
+        -- don't show page x of x if there is only one page
```
Frenzie: I like to think of myself as a commentophile, but this one's really quite redundant. :-)
```suggestion
```

dontcrash:
Okay, I'm happy enough with just "Home", no need to get fancy and try and guess what the user wants the home text to be

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,27 @@ function TouchMenu:updateItems()
     table.insert(self.item_group, self.footer_top_margin)
     table.insert(self.item_group, self.footer)
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
-    self.page_info_text.text = util.template(_("Page %1 of %2"), self.page, self.page_num)
+    -- Don't show page x of x if there is only one page.
```
Frenzie:
```suggestion
```

dontcrash: I need to stop commenting, I need help

dontcrash:
Okay, from here I think all the changes I wanted to make have been explored to their full potential, unless any further changes are needed I won't be sending any further commits to this branch :)

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -55,6 +55,21 @@ if Device:setDateTime() then
...
         sub_item_table = {
+          {
+            text = _("12-hour clock"),
+            keep_menu_open = true,
+            checked_func = function()
+                return G_reader_settings:isTrue("twelve_hour_clock")
+            end,
+            callback = function()
+                local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
+                if twelve_hr then
```
Frenzie: The variable is only used once and it doesn't strike me as a case of increasing clarity.
```suggestion
                if G_reader_settings:isTrue("twelve_hour_clock") then
```

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -55,6 +55,21 @@ if Device:setDateTime() then
...
         text = _("Time and date"),
         sub_item_table = {
+          {
+            text = _("12-hour clock"),
+            keep_menu_open = true,
+            checked_func = function()
+                return G_reader_settings:isTrue("twelve_hour_clock")
+            end,
+            callback = function()
+                local twelve_hr = G_reader_settings:isTrue("twelve_hour_clock")
```
Frenzie: Sorry, the GH suggestion diff thing is a bit difficult to work with multiple lines (if it's possible at all) but of course this line should go too.
```suggestion
```

frontend/apps/filemanager/filemanagerutil.lua
```diff
@@ -24,13 +25,15 @@ end
 
 function filemanagerutil.abbreviate(path)
     if not path then return "" end
-    local home_dir_name = G_reader_settings:readSetting("home_dir_display_name")
-    if home_dir_name ~= nil then
+    if G_reader_settings:isTrue("shorten_home_dir") then
```
Frenzie: Do we keep this off by default?

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,26 @@ function TouchMenu:updateItems()
     table.insert(self.item_group, self.footer_top_margin)
     table.insert(self.item_group, self.footer)
     -- @translators %1 is the current page. %2 is the total number of pages. In some languages a good translation might need to reverse this order, for instance: "Total %2, page %1".
```
Frenzie: The comment should remain attached to the string for clarity. I think xgettext will simply link it to the first following string.

frontend/apps/filemanager/filemanagerutil.lua
dontcrash: Oops, missed this earlier, implemented now

frontend/ui/widget/touchmenu.lua
```diff
@@ -594,16 +594,26 @@ function TouchMenu:updateItems()
...
     self.page_info_right_chev:enableDisable(self.page < self.page_num)
-    local time_info_txt = os.date("%H:%M").." @ "
+
+    local time_info_txt
+    if G_reader_settings:isTrue("twelve_hour_clock") then
+        time_info_txt = os.date("%I:%M %p")
+    else
+        time_info_txt = os.date("%H:%M")
+    end
+    time_info_txt = " – " .. time_info_txt..Device:getPowerDevice():getCapacity().."%"
```
Frenzie: I might be mistaken but shouldn't that be the other way around?
```suggestion
    time_info_txt = time_info_txt .. " – " .. Device:getPowerDevice():getCapacity() .. "%"
```

frontend/apps/filemanager/filemanagerutil.lua
poire-z: I think it was off by default.
I guess most people have enabled it (well, I don't really know :), but as the setting name changed, they won't have it enabled. I don't think it's worth handling a migration.
So, dunno, may be it should indeed be on by default.

frontend/ui/elements/common_settings_menu_table.lua
dontcrash: Resolved already, spotted that one as well just then!

poire-z:
Raah, I hate the code review comments autoclosing as resolved when one line change (and there's some many little code reviews that you may have missed it).
So, reposting it here.
You can make your 2 menu items' callbacks (that toggle a setting) one-line:
https://github.com/koreader/koreader/blob/761f332d47bc9ca0fd87f9425b86a87cd4a6d12d/frontend/apps/reader/modules/readerwikipedia.lua#L280-L282
(which would be flipNilOrTrue if you're making it default as true)

frontend/apps/filemanager/filemanagerutil.lua
dontcrash: Unsure of what to do here

dontcrash:
> Raah, I hate the code review comments autoclosing as resolved when one line change (and there's some many little code reviews that you may have missed it).
> So, reposting it here.
> You can make your 2 menu items' callbacks (that toggle a setting) one-line:
> 
> https://github.com/koreader/koreader/blob/761f332d47bc9ca0fd87f9425b86a87cd4a6d12d/frontend/apps/reader/modules/readerwikipedia.lua#L280-L282
> 
> 
> (which would be flipNilOrTrue if you're making it default as true)

I tried using flipNilOrTrue which resulted in it always being stuck on, are you sure this is right?

frontend/apps/filemanager/filemanagerutil.lua
Frenzie: Isn't there an isNilOrTrue or something?

Frenzie: http://koreader.rocks/doc/modules/luasettings.html#LuaSettings:nilOrTrue

poire-z:
You need to use the adequate check in the checked_func, eg:
https://github.com/koreader/koreader/blob/daacb841148f5040faa9e9c990c3bdd52d3ba037/frontend/apps/reader/modules/readerdevicestatus.lua#L58-L66

frontend/apps/filemanager/filemanagerutil.lua
dontcrash: Solved

frontend/apps/filemanager/filemanagerutil.lua
```diff
@@ -24,13 +25,15 @@ end
...
+    if G_reader_settings:nilOrTrue("shorten_home_dir") then
         local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()
+        if path == home_dir then
+            return _("Home")
+        end
         local len = home_dir:len()
         local start = path:sub(1, len)
         if start == home_dir then
-            return home_dir_name .. path:sub(len+1)
+            return path:sub(len+2)
```
poire-z: > I need to stop commenting, I need help

Don't stop :) Add one here `-- remove leading / of sub-directory` (because I will surely wonder _why that +2 ?!_ next time I'll look at that code)

Frenzie: Like I said, I'm a commentophile. It's just that this is almost verbatim saying the same thing.

```lua
-- Only do this if there's more than one thing.
if thing > 1 then
```

Frenzie: Instead of a comment for a mystery number you can also use a self-explanatory variable name.

```lua
local leading_slash_len = 2
```

Depends a bit on the specifics.

ptrm:
Hooray for the charging icon 😁

PR #5511, opened 2019-10-20
Allow any device with networkmanager enabled to have wifi status in the footer

BrainDamage:
Tested on my Kobo H2O2 and gave a read at the code, should work on other platforms as well.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5511)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -222,7 +222,7 @@ function ReaderFooter:init()
         local mode_name
         for i = 0, #mode_tbl do
             mode_name = mode_tbl[i]
-            if mode_name == "wifi_status" and not Device:isAndroid() then
+            if mode_name == "wifi_status" and not Device:hasWifiManager() then
                 do end -- luacheck: ignore 541
             elseif mode_name == "frontlight" and not Device:hasFrontlight() then
                 do end -- luacheck: ignore 541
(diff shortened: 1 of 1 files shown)


Frenzie:
Not on Android, where kernel modules aren't involved.

The reason Android is hard-coded is that it's the only platform that can currently do it async. The #3078 type stuff is a bit in limbo.

NiLuJe:
Hard veto. It's blocking on most devices, which is why we don't do it ;).

Frenzie:
Basically blocking for about 10-20 ms is the best case scenario.

NiLuJe:
And even if it weren't blocking, on most devices, the check implies *at least* a fork(), which would be enough to warrant a veto for the footer IMHO ;).

BrainDamage:
on embedded devices, can we assume we're the only managers of the wifi connection?
if yes, it could just be checked at startup/wakeup, and save the state, then it'd be as fast as reading a variable

Frenzie:
That's how it should function from the footer's perspective regardless (not with a fork *for the footer* ;-) ), and as a simple on/off indicator (as opposed to connected) I figure something like that would do the trick.

NiLuJe:
That might not always hold true.

On Kobo, we inherit whatever the state was on startup (unless it happened
to be very broken, In which case we kill wifi).

On Kindle, we defer to the system.

As do we on PB I think.

So that would require some finaggling on startup to get an accurate idea of
the actual state. But, yeah, that would definitely make more sense ;).

NiLuJe:
There's the `NetworkConnected` event. It might not cover all bases, but it's probably a good start ;).

pazos:
I'm closing this one.

Frenzie:
Basically good visual change in theory, unfortunately unworkable in practice for the moment. :-)

Issue #5501, opened 2019-10-17
Timeout on wifi connected infobox


Frenzie:
* KOReader version: master
* Device: Kobo

#### Issue
It just stays on the screen and I keep forgetting to add a timeout.

#### Steps to reproduce
Enable wifi.

dontcrash:
Apologies about that first pull request, I made the mistake of pushing to my master branch instead of creating a new one for the UI changes, so it messed it all up!

PR #5513, opened 2019-10-21
Added Wi-Fi connected info message timeout

dontcrash:
Resolved issue #5501

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -504,8 +504,7 @@ pot:
 	$(MAKE) -C l10n push
 
 po:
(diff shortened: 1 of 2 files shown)


Frenzie:
Please strip the other commits from the PR. ;-)

Frenzie:
git-revert is one way, but in this case you can just rebase the branch, drop all the commits and force push.

The easiest way would be something like:

```bash
git rebase -i HEAD~30
# comment out all the irrelevant lines & save
git push -f
```

PR #5522, opened 2019-10-23
set global inverse_reading_order independent of current setting

yparitcher:
allow setting the global inverse_reading_order regardless of what the value is for the current document.

### DIFF
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -244,13 +244,14 @@ function ReaderPaging:addToMainMenu(menu_items)
             self.ui:handleEvent(Event:new("ToggleReadingOrder"))
         end,
         hold_callback = function(touchmenu_instance)
+            local inverse_reading_order = G_reader_settings:isTrue("inverse_reading_order")
             UIManager:show(ConfirmBox:new{
-                text = self.inverse_reading_order and _("Enable right to left reading by default?")
-                    or _("Disable right to left reading by default?"),
-                ok_text = self.inverse_reading_order and _("Enable")
-                    or _("Disable"),
+                text = inverse_reading_order and _("Disable right to left reading by default?")
+                    or _("Enable right to left reading by default?"),
+                ok_text = inverse_reading_order and _("Disable")
+                    or _("Enable"),
                 ok_callback = function()
-                    G_reader_settings:saveSetting("inverse_reading_order", self.inverse_reading_order)
+                    G_reader_settings:saveSetting("inverse_reading_order", not inverse_reading_order)
                     if touchmenu_instance then touchmenu_instance:updateItems() end
                 end,
             })
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -370,13 +370,14 @@ function ReaderRolling:addToMainMenu(menu_items)
             self.ui:handleEvent(Event:new("ToggleReadingOrder"))
         end,
         hold_callback = function(touchmenu_instance)
+            local inverse_reading_order = G_reader_settings:isTrue("inverse_reading_order")
             UIManager:show(ConfirmBox:new{
-                text = self.inverse_reading_order and _("Enable right to left reading by default?")
-                    or _("Disable right to left reading by default?"),
-                ok_text = self.inverse_reading_order and _("Enable")
-                    or _("Disable"),
+                text = inverse_reading_order and _("Disable right to left reading by default?")
+                    or _("Enable right to left reading by default?"),
+                ok_text = inverse_reading_order and _("Disable")
+                    or _("Enable"),
                 ok_callback = function()
-                    G_reader_settings:saveSetting("inverse_reading_order", self.inverse_reading_order)
+                    G_reader_settings:saveSetting("inverse_reading_order", not inverse_reading_order)
                     if touchmenu_instance then touchmenu_instance:updateItems() end
                 end,
             })


Frenzie:
It's by design. The idea is:

1. I like this value.
2. Let's make it default.

There are some others that are written the same way (e.g., Hanging punctuation).

I'm inclined to think making it independent of the displayed setting adds a lot of extra holds & taps to achieve the same result.

Pinging @poire-z

poire-z:
Was going to approve, seeing it does what's promised in this PR text :)
But I never really looked at how that work. I might have set one or two as default once and never looked again.

So, our current behaviour is that the "default" button is not a toggle? When you hit it, it just sets as default the current value? I guess that looks fine.
Might help then that the "default" button would be grey/disabled when the default is already the current value (so, you know it applied when you hit it by becoming grey - or that you need to do something else to have it enabled, like hitting the main OK/Cancel.)

Anyway, no real opinion - both our current behaviour and @yparitcher one might be not really obvious. But a single global same unobviousness is better than multiple :)

poire-z:
OK, so it wasn't a button :) discard my previous middle paragraph...

Well, for font & hyphenation, so lists, we set a symbol for the default one. For other menu item with a checkbox, no real way or reason to do the same (it would add noise to the menu).
But may be just a clearer dialog, saying what is the current value, and (instead of `Cancel|OK`) 2 buttons: `Set Disable as default | Set Enable as default` would make all that clearer?
There woule be no real way to escape that dialog (except hitting outside), so you'd have to make a choice.

Frenzie:
It's an option, but as a user I'd feel better with a dialog like this:

Screenshot_20191023_101416

poire-z:
Oh, of course, we _can_ do 3 buttons :)

yparitcher:
> It's by design. The idea is:
> 
>     1. I like this value.
> 
>     2. Let's make it default.
> 
> 
> There are some others that are written the same way (e.g., Hanging punctuation).
> 
> I'm inclined to think making it independent of the displayed setting adds a lot of extra holds & taps to achieve the same result.

i see where you are coming from, however i bumped into this when trying to disable the global while the local was enabled, and kept on getting the enable dialog. 
At first i thought there was a typo or bug and was confused until I read the code.

Your way is better the first time someone sets the global (ie: the global is not known now), however it is confusing when one knows the global and is trying to switch it.

There is no way to toggle the global without toggling the local. and with Inverse_reading_order i want some documents RTL and some LTR.


as poire-z said whatever is decided it should be the same everywhere.

highlighting is set the inverse of your way: global true when local is false
dictionary fuzzy search  and hanging punctuation are like your way.

these are the only ones that have a checkbox selector with a default that can cause this confusion all other toggles (bottom menu) have slider so you an just hold the option you want as default.

yparitcher:
the clearest way would probably be to go wth three buttons `cancel | disable | enable` with the order of disable enable based on the current local like is currently done.

Frenzie:
Pretty sure @poire-z and I already agreed on Cancel | Disable | Enable. ;-)

> as poire-z said whatever is decided it should be the same everywhere.

I kind of felt that was the main point of my first reply actually. >_> :-)

poire-z:
Yes, we all agree on 3 buttons (in a single row, or cancel on the bottom row if we go with 2 rows?).

> the clearest way would probably be to go wth three buttons cancel | disable | enable with the order of disable enable based on the current local like is currently done.

Dunno if the order should be changed (it would not matter much as we would not visit them often and get no chance of getting used to the buttons order :) but in general, thing shouldn't move too much).
But if a global is set, its button might be grey/dimmed (the current global value) - so only the other is tap'able (to change the global value).
And the text should be explicite too.

Frenzie:
Yeah I'd just keep the order the same. Moving them around is needlessly complex in the code and annoyingly unpredictable as a user.

yparitcher:
how does this look?

Reader_2019-Oct-23_175619

i am using `MultiConfirmBox` i am not sure i can gray one out with it.

yparitcher:
once we come to a final decision i will fix the other ones.

poire-z:
> i am using MultiConfirmBox i am not sure i can gray one out with it.

Looks like it does not (yet), but you can quickly make it accept some new keys:
```diff
--- a/frontend/ui/widget/multiconfirmbox.lua
+++ b/frontend/ui/widget/multiconfirmbox.lua
@@ -50,2 +50,4 @@ local MultiConfirmBox = InputContainer:new{
     cancel_callback = function() end,
+    choice1_enabled = true,
+    choice2_enabled = true,
     margin = Size.margin.default,
@@ -104,2 +106,3 @@ function MultiConfirmBox:init()
                     text = self.choice1_text,
+                    enabled = self.choice1_enabled,
                     callback = function()
@@ -111,2 +114,3 @@ function MultiConfirmBox:init()
                     text = self.choice2_text,
+                    enabled = self.choice2_enabled,
                     callback = function()
```

> how does this look?

Good, but not super clear. (I could be tempted to hit "Disable" to get out of it :)
Dunno. Either longer text like (for this one, there is always the default of LTR) like _The default for newly opened books is Left-to-Right reading order. You can set newly opened books to default to Right-To-Left reading order._
And have buttons more explicite. Either "Enable by default" and "Disable by default" (but that is long), or better: `Cancel | Left-To-Right | Right-To-Left` (might be short enough to stick in a single row).
(The text could read then:  _The default for newly opened books is Left-to-Right reading order. What should be the default_ or _What should newly opened book reading direction be?_ (hope you'll find a better english wording than my cumbersome attempts :)

yparitcher:
Reader_2019-Oct-23_190751
Reader_2019-Oct-23_190755

poire-z:
Looks good :) Even better with a newline before `Would you...`?

yparitcher:
Reader_2019-Oct-23_193839

will start on the other ones now

yparitcher:
Reader_2019-Oct-23_195155


Reader_2019-Oct-23_200220


Reader_2019-Oct-23_200225


Reader_2019-Oct-23_201248

yparitcher:
inverse reading order, enable highlighting, dictionary fuzzy search and hanging punctuation are all the same now.
3 buttons, totally independent of the local config.

Frenzie:
I'm not overly enthused by the button disabling. Do you think it improves clarity? I like to be able to emphatically click that button, even if it is already the default.

Regardless of any disabling, I think it'd improve visual clarity and internal unity to follow the example I posted a little more closely.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -1179,17 +1179,20 @@ function ReaderHighlight:onClose()
...
-        new_text = _("Enable highlight by default.")
-    end
-    UIManager:show(ConfirmBox:new{
-        text = new_text,
-        ok_callback = function()
-            G_reader_settings:saveSetting("highlight_disabled", highlight_disabled)
+function ReaderHighlight:toggleDefault()
+    local highlight_disabled = G_reader_settings:isTrue("highlight_disabled")
+    UIManager:show(MultiConfirmBox:new{
+        text = highlight_disabled and _("Enable highlighting by default.")
```
Frenzie: Same basic comment as above, but I assume highlighting is straightforward enough.

```
Would you like to enable or disable highlighting by default? The current default (★) is enabled.
```

frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -244,13 +244,20 @@ function ReaderPaging:addToMainMenu(menu_items)
...
-            UIManager:show(ConfirmBox:new{
-                text = self.inverse_reading_order and _("Enable right to left reading by default?")
-                    or _("Disable right to left reading by default?"),
-                ok_text = self.inverse_reading_order and _("Enable")
-                    or _("Disable"),
-                ok_callback = function()
-                    G_reader_settings:saveSetting("inverse_reading_order", self.inverse_reading_order)
+            local inverse_reading_order = G_reader_settings:isTrue("inverse_reading_order")
+            UIManager:show(MultiConfirmBox:new{
+                text = inverse_reading_order and _("The default for newly opened books is Right To Left page turning.\nWould you like to change it?")
```
Frenzie:
```suggestion
                text = inverse_reading_order and _("The default (★) for newly opened books is right-to-left (RTL) page turning.\n\nWould you like to change it?")
```

frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -244,13 +244,20 @@ function ReaderPaging:addToMainMenu(menu_items)
...
+                or _("The default for newly opened books is Left To Right page turning.\nWould you like to change it?"),
```
Frenzie:
```suggestion
                or _("The default (★) for newly opened books is left-to-right (LTR) page turning.\n\nWould you like to change it?"),
```

frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -428,14 +429,22 @@ function ReaderTypeset:addToMainMenu(menu_items)
...
 
 function ReaderTypeset:makeDefaultFloatingPunctuation()
-    local toggler = self.floating_punctuation == 1 and _("On") or _("Off")
-    UIManager:show(ConfirmBox:new{
+    local floating_punctuation = G_reader_settings:isTrue("floating_punctuation")
+    local toggler = floating_punctuation and _("Disable") or _("Enable")
+    UIManager:show(MultiConfirmBox:new{
         text = T(
-            _("Set default hanging punctuation to %1?"),
+            _("%1 hanging punctuation by default?"),
```
Frenzie: That should be written out in full. There's no guarantee that enable and disable are interchangeable in other languages, let alone take the same form without context.

However, it's a moot point because it should take the same form as the others:

```
Would you like to enable or disable hanging punctuation by default? The current default (★) is enabled.
```

frontend/ui/widget/multiconfirmbox.lua
```diff
@@ -48,6 +48,8 @@ local MultiConfirmBox = InputContainer:new{
     choice1_callback = function() end,
     choice2_callback = function() end,
     cancel_callback = function() end,
+    choice1_enabled = true,
```
Frenzie: Looks like this doesn't support `text_func` atm? That's slightly unfortunate but it'd be easy to add. I assume the underlying Button(Table) already supports it.

yparitcher:
> I'm not overly enthused by the button disabling. Do you think it improves clarity? I like to be able to emphatically click that button, even if it is already the default.

it was @poire-z idea, i don't care either way. when you guys decide which way let me know and i will do so.

frontend/apps/reader/modules/readertypeset.lua
yparitcher: i was trying to leave it as intact as possible, i didn't realize the implications.

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -950,21 +951,23 @@ function ReaderDictionary:onSaveSettings()
...
-    UIManager:show(ConfirmBox:new{
+function ReaderDictionary:toggleFuzzyDefault()
+    local disable_fuzzy_search = G_reader_settings:isTrue("disable_fuzzy_search")
+    UIManager:show(MultiConfirmBox:new{
         text = T(
             disable_fuzzy_search
-            and _("Disable fuzzy search by default?")
-            or _("Enable fuzzy search by default?")
+            and _("Enable fuzzy search by default?")
+            or _("Disable fuzzy search by default?")
```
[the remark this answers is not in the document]
yparitcher: this pr is just for the defaults, you can add docs however you want

yparitcher:
Reader_2019-Oct-24_150424
Reader_2019-Oct-24_150435
Reader_2019-Oct-24_150446

frontend/apps/reader/modules/readerdictionary.lua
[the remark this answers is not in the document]
Frenzie: I assume that comes across more brusque than intended, but I'm still a bit confused by your objection. This is my suggested string for this particular default in line with the rest of the UI, no more no less.

I'll merge it with the button disabling in place for now.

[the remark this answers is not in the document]
yparitcher: sorry, was not trying to be so harsh, i just wanted to get this out without getting tangled up in documentation. i have seen PRs (in other repos) that get sidetracked and stalled.

poire-z:
> it was @poire-z idea, i don't care either way. when you guys decide which way let me know and i will do so.

I don't care either. I was just suggesting ideas to make these dialogs the less confusing possible - and they got less confusing.
So ok if you want to make the buttons not disabled (after all, we don't disable the current font or hyphenation in their respective lists).

PR #5525, opened 2019-10-23
CoverBrowser: some fixes after #5435

### LINKED ISSUE #5523: Metadata extraction: books on History page not extracted
* KOReader version: v2019.10-10-gf38073c_2019-10-23/  
* Device: KA1

#### Issue
Metadata for opened books are not extracted when using 'Extract and cache book information' and    then open History page 

#### Steps to reproduce
1. Open some books
2. Delete cache database and restart KOReader
3. Choose (for example) "Detailed list..." for both file manager and History
4. Extract
5. Open History

poire-z:
- Use a hardcoded default value of 10 files per page. https://github.com/koreader/koreader/pull/5435#issuecomment-545335940
- Save cover sizetag as `s<nb of items per page>` instead of `s<item height>` so they are stable (and not re-fetched) when switching between FileBrowser and History, which have some small geometry differences. Closes #5523.
- When showing History in landscape mode, decrease perpage from the set value, so items looks the same, instead of smaller.
- In list mode, use a step (larger with big items) when iterating and decreasing font sizes to make text fit, to speed up the process a bit. (This was already happening in mosaic mode.) On devices, instead of always -1, it will be -2 at around 8-12 items per page, -3 at around 4-6, possibly -4 on large devices.
There are probably ways to do that better, but that helps quite a bit and will do for now.

### DIFF
--- a/plugins/coverbrowser.koplugin/listmenu.lua
+++ b/plugins/coverbrowser.koplugin/listmenu.lua
@@ -202,7 +202,7 @@ function ListMenuItem:update()
     local max_img_w = dimen.h - 2*border_size -- width = height, squared
     local max_img_h = dimen.h - 2*border_size
     local cover_specs = {
-        sizetag = "s"..max_img_h,
+        sizetag = self.menu.cover_sizetag,
         max_cover_w = max_img_w,
         max_cover_h = max_img_h,
     }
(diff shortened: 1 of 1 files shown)


poire-z:
Pinging @derungs for info.

NiLuJe:
Yeah, I was vaguely thinking about making the step dynamic (f.g., double it each iteration), but that would probably involve a backtracking step at the end to correct for overshoot...

poire-z:
Yep, that was just a quick fix. It might deserve something better and quicker (feel free to have a go at it if you have some idea - I didn't go further than thinking if we were to use dichotomy, it should be tweaked to use top/2 at the bottom step instead of the min value... or something like that :|)

NiLuJe:
Random observation: I've seen a few re-fetch of covers in Details view with the same defaults as before (i.e., 10 items).

Was that expected?

NiLuJe:
Also on the list of things to consider: [FreeType's caching subsystem](https://freetype.org/freetype2/docs/reference/ft2-cache_subsystem.html)?

PR #5527, opened 2019-10-24
Add book to favorites

robert00s:
New function that support organize books to favorites.
We can:
- add files to favorites
- remove files from favorites
- view list of favorites books
- sort list of favorites books
- open books from favorites
- change view mode (like in history)
Now we support only one collection (named favorites) but code is preparted to extend to more collections.

Some screenshots:

Close: #5099

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -9,6 +9,7 @@ local DocumentRegistry = require("document/documentregistry")
 local Event = require("ui/event")
 local FileChooser = require("ui/widget/filechooser")
 local FileManagerBookInfo = require("apps/filemanager/filemanagerbookinfo")
+local FileManagerCollection = require("apps/filemanager/filemanagercollection")
 local FileManagerConverter = require("apps/filemanager/filemanagerconverter")
 local FileManagerFileSearcher = require("apps/filemanager/filemanagerfilesearcher")
 local FileManagerHistory = require("apps/filemanager/filemanagerhistory")
@@ -288,11 +289,13 @@ function FileManager:init()
             },
             -- a little hack to get visual functionality grouping
             {},
-            {
+        }
+        if lfs.attributes(file, "mode") == "file" then
+            table.insert(buttons, {
                 {
                     text = _("Open with…"),
-                    enabled = lfs.attributes(file, "mode") == "file" and (DocumentRegistry:getProviders(file) == nil
-                        or #(DocumentRegistry:getProviders(file)) > 1),
+                    enabled = DocumentRegistry:getProviders(file) ~= nil
+                        and #(DocumentRegistry:getProviders(file)) > 1,
                     callback = function()
                         UIManager:close(self.file_dialog)
                         self:showSetProviderButtons(file, FileManager.instance, ReaderUI)
@@ -300,13 +303,14 @@ function FileManager:init()
                 },
                 {
                     text = _("Convert"),
-                    enabled = lfs.attributes(file, "mode") == "file"
-                        and FileManagerConverter:isSupported(file),
+                    enabled = FileManagerConverter:isSupported(file),
                     callback = function()
                         UIManager:close(self.file_dialog)
                         FileManagerConverter:showConvertButtons(file, self)
                     end,
                 },
+            })
+            table.insert(buttons, {
                 {
                     text = _("Book information"),
                     enabled = FileManagerBookInfo:isSupported(file),
@@ -315,8 +319,26 @@ function FileManager:init()
                         UIManager:close(self.file_dialog)
                     end,
                 },
-            },
-        }
+                {
+                    text_func = function()
+                        if require("readcollection"):checkItemExist(file, "favorites") then
+                            return _("Remove from favorites")
+                        else
+                            return _("Add to favorites")
+                        end
+                    end,
+                    enabled = DocumentRegistry:getProviders(file) ~= nil,
+                    callback = function()
+                        if require("readcollection"):checkItemExist(file, "favorites") then
+                            require("readcollection"):removeItem(file, "favorites")
+                        else
+                            require("readcollection"):addItem(file, "favorites")
+                        end
+                        UIManager:close(self.file_dialog)
+                    end,
+                },
+            })
+        end
         if lfs.attributes(file, "mode") == "directory" then
             local realpath = util.realpath(file)
             table.insert(buttons, {
@@ -359,6 +381,9 @@ function FileManager:init()
     self.active_widgets = { Screenshoter:new{ prefix = 'FileManager' } }
     table.insert(self, self.menu)
     table.insert(self, FileManagerHistory:new{
+        ui = self,
+    })
+    table.insert(self, FileManagerCollection:new{
         ui = self,
     })
     table.insert(self, FileManagerFileSearcher:new{ ui = self })
@@ -695,9 +720,10 @@ function FileManager:pasteHere(file)
                 self:moveFile(DocSettings:getSidecarDir(orig), dest) -- dest is always a directory
             end
             if self:moveFile(orig, dest) then
-                --update history
+                --update history and collections
                 local dest_file = string.format("%s/%s", dest, util.basename(orig))
                 require("readhistory"):updateItemByPath(orig, dest_file)
+                require("readcollection"):updateItemByPath(orig, dest_file)
                 --update last open file
                 if G_reader_settings:readSetting("lastfile") == orig then
                     G_reader_settings:saveSetting("lastfile", dest_file)
@@ -764,7 +790,7 @@ function FileManager:createFolder(curr_folder, new_folder)
 end
 
 function FileManager:deleteFile(file)
-    local ok, err
+    local ok, err, is_dir
     local file_abs_path = util.realpath(file)
     if file_abs_path == nil then
         UIManager:show(InfoMessage:new{
@@ -778,6 +804,7 @@ function FileManager:deleteFile(file)
         ok, err = os.remove(file_abs_path)
     else
         ok, err = util.purgeDir(file_abs_path)
+        is_dir = true
     end
     if ok and not err then
         if is_doc then
@@ -789,6 +816,8 @@ function FileManager:deleteFile(file)
             end
             doc_settings:purge()
         end
+        --remove from collections
+        require("readcollection"):removeItemByPath(file, is_dir)
         UIManager:show(InfoMessage:new{
             text = util.template(_("Deleted %1"), file),
             timeout = 2,
@@ -804,6 +833,8 @@ function FileManager:renameFile(file)
     if util.basename(file) ~= self.rename_dialog:getInputText() then
         local dest = util.joinPath(util.dirname(file), self.rename_dialog:getInputText())
         if self:moveFile(file, dest) then
+            -- update collections
+            require("readcollection"):updateItemByPath(file, dest)
             if lfs.attributes(dest, "mode") == "file" then
                 local doc = require("docsettings")
                 local move_history = true;
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -302,26 +302,18 @@ function FileManager:init()
                     end,
                 },
                 {
-                    text = _("Convert"),
-                    enabled = FileManagerConverter:isSupported(file),
-                    callback = function()
-                        UIManager:close(self.file_dialog)
-                        FileManagerConverter:showConvertButtons(file, self)
-                    end,
-                },
-            })
-            table.insert(buttons, {
-                {
                     text = _("Book information"),
                     enabled = FileManagerBookInfo:isSupported(file),
                     callback = function()
                         FileManagerBookInfo:show(file)
                         UIManager:close(self.file_dialog)
                     end,
-                },
+                }
+            })
+            table.insert(buttons, {
                 {
                     text_func = function()
-                        if ReadCollection:checkItemExist(file, "favorites") then
+                        if ReadCollection:checkItemExist(file) then
                             return _("Remove from favorites")
                         else
                             return _("Add to favorites")
@@ -329,15 +321,27 @@ function FileManager:init()
                     end,
                     enabled = DocumentRegistry:getProviders(file) ~= nil,
                     callback = function()
-                        if ReadCollection:checkItemExist(file, "favorites") then
-                            ReadCollection:removeItem(file, "favorites")
+                        if ReadCollection:checkItemExist(file) then
+                            ReadCollection:removeItem(file)
                         else
-                            ReadCollection:addItem(file, "favorites")
+                            ReadCollection:addItem(file)
                         end
                         UIManager:close(self.file_dialog)
                     end,
                 },
             })
+            if FileManagerConverter:isSupported(file) then
+                table.insert(buttons, {
+                    {
+                        text = _("Convert"),
+                        enabled = true,
+                        callback = function()
+                            UIManager:close(self.file_dialog)
+                            FileManagerConverter:showConvertButtons(file, self)
+                        end,
+                    }
+                })
+            end
         end
         if lfs.attributes(file, "mode") == "directory" then
             local realpath = util.realpath(file)
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -24,6 +24,7 @@ local InputContainer = require("ui/widget/container/inputcontainer")
 local InputDialog = require("ui/widget/inputdialog")
 local MultiConfirmBox = require("ui/widget/multiconfirmbox")
 local PluginLoader = require("pluginloader")
+local ReadCollection = require("readcollection")
 local ReaderDeviceStatus = require("apps/reader/modules/readerdevicestatus")
 local ReaderDictionary = require("apps/reader/modules/readerdictionary")
 local ReaderGesture = require("apps/reader/modules/readergesture")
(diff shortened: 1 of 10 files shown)


frontend/apps/filemanager/filemanager.lua
```diff
@@ -315,8 +319,26 @@ function FileManager:init()
...
+                    text_func = function()
+                        if require("readcollection"):checkItemExist(file, "favorites") then
+                            return _("Remove from favorites")
+                        else
+                            return _("Add to favorites")
+                        end
+                    end,
+                    enabled = DocumentRegistry:getProviders(file) ~= nil,
+                    callback = function()
+                        if require("readcollection"):checkItemExist(file, "favorites") then
```
Frenzie: While the results of `require` are cached, it should still be preferable to do `local ReadCollection = require(…)`. (And it should be fine at the top of the file I figure?)

frontend/apps/filemanager/filemanager.lua
```diff
@@ -695,9 +720,10 @@ function FileManager:pasteHere(file)
                 self:moveFile(DocSettings:getSidecarDir(orig), dest) -- dest is always a directory
             end
             if self:moveFile(orig, dest) then
-                --update history
+                --update history and collections
```
Frenzie: If we're updating comments, they like to be proper sentences. ;-)
```suggestion
                -- Update history and collections.
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -789,6 +816,8 @@ function FileManager:deleteFile(file)
             end
             doc_settings:purge()
         end
+        --remove from collections
```
Frenzie: Seems slightly redundant with the code.
```suggestion
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -804,6 +833,8 @@ function FileManager:renameFile(file)
     if util.basename(file) ~= self.rename_dialog:getInputText() then
         local dest = util.joinPath(util.dirname(file), self.rename_dialog:getInputText())
         if self:moveFile(file, dest) then
+            -- update collections
```
Frenzie:
```suggestion
```

frontend/apps/filemanager/filemanagercollection.lua
```diff
@@ -0,0 +1,124 @@
...
+local FileManagerCollection = InputContainer:extend{
+    coll_menu_title = _("Favorites"),
+}
+
+function FileManagerCollection:init()
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function FileManagerCollection:addToMainMenu(menu_items)
+    -- insert table to main tab of filemanager menu
```
Frenzie: Well, technically just wherever MenuSorter is told to put it by the user. ;-)

But this is the same as every addToMainMenu function; I don't think a comment adds much?

```suggestion
```

frontend/apps/filemanager/filemanagercollection.lua
```diff
@@ -0,0 +1,124 @@
...
+    menu_items.collections = {
+        text = self.coll_menu_title,
+        callback = function()
+            self:onShowColl("favorites")
+        end,
+    }
+end
+
+function FileManagerCollection:updateItemTable()
+    -- try to stay on current page
```
Frenzie:
```suggestion
    -- Try to stay on current page.
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -288,25 +289,28 @@ function FileManager:init()
...
             {},
-            {
+        }
+        if lfs.attributes(file, "mode") == "file" then
+            table.insert(buttons, {
                 {
                     text = _("Open with…"),
-                    enabled = lfs.attributes(file, "mode") == "file" and (DocumentRegistry:getProviders(file) == nil
-                        or #(DocumentRegistry:getProviders(file)) > 1),
+                    enabled = DocumentRegistry:getProviders(file) ~= nil
```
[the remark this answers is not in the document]
robert00s: Yes, you're right. This was taken from commit before #5497.

frontend/apps/filemanager/filemanagercollection.lua
Frenzie: Actually that should be *the* current page.

NiLuJe:
Features sounds nice, but I wonder if it wouldn't be better suited to the SQL db than another LuaSetting?

robert00s:
@NiLuJe
 I think that we can move all (or almost all) lua settings to SQL database. But we need to thinks how to do it (how many databases, which settings should be moved, what about table data (serialize?). )
For now I don't want start that job. We have to discuss this with a larger group.

NiLuJe:
I meant specifically the BookInfo db, because that felt vaguely related (and possibly much more flexible), but fair enough ;).

This isn't designed to handle large collections anyway, so that's fine by me. I can see it being used to track a TBR pile, for instance.

frontend/apps/filemanager/filemanagercollection.lua
```diff
@@ -0,0 +1,124 @@
...
+                callback = function()
+                    UIManager:close(collfile_dialog)
+                    local item_table = {}
+                    for i=1, #self._manager.coll_menu.item_table do
+                        table.insert(item_table, {text = self._manager.coll_menu.item_table[i].text, label = self._manager.coll_menu.item_table[i].file})
+                    end
+                    local SortWidget = require("ui/widget/sortwidget")
+                    local sort_item
+                    sort_item = SortWidget:new{
+                        title = _("Sort favorites book"),
```
poire-z: `Sort favorite books` or just `Sort favorites`

frontend/apps/filemanager/filemanagercollection.lua
```diff
@@ -0,0 +1,124 @@
...
+                    UIManager:close(collfile_dialog)
+                end,
+            },
+        },
+        {
+            {
+                text = _("Book information"),
+                enabled = FileManagerBookInfo:isSupported(item.file),
+                callback = function()
+                    FileManagerBookInfo:show(item.file)
```
poire-z: Later, you might want to tweak the onHold dialog when the user uses a CoverBrowser mode, like we did for History in:
https://github.com/koreader/koreader/blob/a5419439ce8115a8571a8c78c6735618c82c270b/plugins/coverbrowser.koplugin/covermenu.lua#L398-L400
so we have the other buttons like View book description, View full size cover...

frontend/ui/elements/filemanager_menu_order.lua
```diff
@@ -128,6 +128,7 @@ local order = {
     },
     main = {
         "history",
+        "collections",
         "open_last_document",
         "----------------------------",
```
poire-z: For both reader and filebrowser hamburger menu, I'd rather see `history` and `open_last/previous_document` grouped together. As collections/favorites is not related to history/recent file, may be:
```
        "history",
        "open_last_document",
        "collections",
```

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -41,6 +44,7 @@ local init_done = false
 local filemanager_display_mode = false -- not initialized yet
 local history_display_mode = false -- not initialized yet
 local series_mode = nil -- defaults to not display series
+local collection_display_mode = false -- not initialized yet
```
poire-z: Just swap these 2. `series_mode` is a bit unrelated to the now 3 real display modes.

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -663,4 +716,81 @@ function CoverBrowser:setupHistoryDisplayMode(display_mode)
...
+        coll_menu.onCloseWidget = CoverMenu.onCloseWidget
+
+        if display_mode == "mosaic_image" or display_mode == "mosaic_text" then -- mosaic mode
+            -- Replace some other original methods with those from our MosaicMenu
+            local MosaicMenu = require("mosaicmenu")
+            coll_menu._recalculateDimen = MosaicMenu._recalculateDimen
+            coll_menu._updateItemsBuildUI = MosaicMenu._updateItemsBuildUI
+            -- Set MosaicMenu behaviour:
+            coll_menu._do_cover_images = display_mode ~= "mosaic_text"
+            -- no need for do_hint_opened with Collections
```
poire-z: Why not?

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -663,4 +716,81 @@ function CoverBrowser:setupHistoryDisplayMode(display_mode)
...
+            local ListMenu = require("listmenu")
+            coll_menu._recalculateDimen = ListMenu._recalculateDimen
+            coll_menu._updateItemsBuildUI = ListMenu._updateItemsBuildUI
+            -- Set ListMenu behaviour:
+            coll_menu._do_cover_images = display_mode ~= "list_only_meta"
+            coll_menu._do_filename_only = display_mode == "list_image_filename"
+            -- no need for do_hint_opened with Collections
+
+        end
+        coll_menu._do_hint_opened = BookInfoManager:getSetting("history_hint_opened")
```
poire-z: I would naively just set `coll_menu._do_hint_opened = true` , or `= false` (as you wrote above _no need for do_hint_opened with Collections_). Bur probably not the setting that is named after / targeted for History.

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -663,4 +716,81 @@ function CoverBrowser:setupHistoryDisplayMode(display_mode)
...
+    end
+    -- remember current mode in module variable
+    collection_display_mode = display_mode
+    logger.dbg("CoverBrowser: setting Collection display mode to:", display_mode or "classic")
+
+    if not init_done and not display_mode then
+        return -- starting in classic mode, nothing to patch
+    end
+
+    -- We only need to replace one FileManagerHistory method
```
poire-z: one _FileManagerCollection_ method

frontend/readcollection.lua
```diff
@@ -0,0 +1,156 @@
+local DataStorage = require("datastorage")
+local LuaSettings = require("luasettings")
+local getFriendlySize = require("util").getFriendlySize
+local lfs = require("libs/libkoreader-lfs")
+local realpath = require("ffi/util").realpath
+local util = require("frontend/util")
```
poire-z: No need for `frontend/` in our require.

frontend/readcollection.lua
```diff
@@ -0,0 +1,156 @@
...
+
+local collection_file = DataStorage:getSettingsDir() .. "/collection.lua"
+
+local ReadCollection = {}
+
+function ReadCollection:read(collections)
+    local collection_handle = LuaSettings:open(collection_file)
+    local coll = collection_handle:readSetting(collections) or {}
```
poire-z: That file was a bit hard to understand :)

I guess that `collections` (plural) here , `collection` (singular) below is a _collection name_, and you pass "favorites" as a string when you call it (your single default collection name for now, as I understand we'll be able to manage more later).
So, may be use `collection_name` everywhere as the variable holding that collection name string, to make it easier to understand.
And use `collections` (plural) when you're dealing with the whole datastructure. And `collection` (singular) when dealing with a sub-table of a specific `collection_name`.

And may be, to avoid passing "favorites" everywhere else, just pass nothing, and in here:
```lua
local DEFAULT_COLLECTION_NAME = "favorites"
[...]
function ReadCollection:something(collection_name)
  if not collection_name then collection_name = DEFAULT_COLLECTION_NAME end
```

robert00s:
@poire-z Something like that? :)

obraz

poire-z:
Mhhh, I don't like it :)

I'd rather keep all these as-is at the bottom, as they are all related to cover browser, and the last one needs the full width for the long description (couldn't find anything smaller, and I think it needs to be that explicite).
image

So, either, for the middle group, 2 lines (3 + 2 or 2 + 3), or 3 lines (2 + 1 + 2 or 2 + 2 + 1) like:
```
Open with...     |      Book information
Convert | Mark as read | Add to favorite
```
or
```
Open with...     |      Book information
Convert | Add to favorite | Mark as read
```
or
```
Open with... | Book information
     Add to favorite
Convert      |     Mark as read
```

Given that `Convert` is only enabled with `.md` files (Mardown only for now, but there hasn't been any more format added for ages), and these are rare, I would happily just not display it in these menus, and add it only when meeting a converter supported file (even as a single full width button).
That way, we would not need to have 3 buttons in a row.
@Frenzie : any opinion?

Frenzie:
Sure, you can hide convert if you need to win some space.

poire-z:
OK, So may be we can go with this:
Only when converting is supported, add a line with a single button `Convert` above `Open with...`.

That would give in classic mode:
```
Copy  |  Paste  | Purge .sdr
Cut   |  Delete | Rename
========================
     Convert (if convertable)
Open with... | Book information
       Add to favorite
```
That would give in CoverBrowser modes:
```
Copy  |  Paste  | Purge .sdr
Cut   |  Delete | Rename
========================
     Convert (if convertable)
Open with... | Book information
Mark as read |  Add to favorite
View fscover | View book description
Ignore cover | Ignore metadata
Refresh cached book information
```
You'll have to tweak the buttons move and adding code in covermenu.lua to adapt to this new layout, and the present or absent `Convert`.
So, may be change my above drawings and move Convert below Add to favorite, if that can make the above easier, as Book info and the place where to add Mark as read will always have the same indice:
```
Copy  |  Paste  | Purge .sdr
Cut   |  Delete | Rename
========================
Open with... | Book information
       Add to favorite
     Convert (if convertable)
```
```
Copy  |  Paste  | Purge .sdr
Cut   |  Delete | Rename
========================
Open with... | Book information
Mark as read |  Add to favorite
     Convert (if convertable)
View fscover | View book description
Ignore cover | Ignore metadata
Refresh cached book information
```

robert00s:
@poire-z 
Thanks for review and tips. I did all things :)

plugins/coverbrowser.koplugin/main.lua
```diff
@@ -265,6 +270,54 @@ function CoverBrowser:addToMainMenu(menu_items)
                         separator = true,
                     },
                 },
+            },
+            {
+                text = _("Collection display mode"),
```
poire-z: May be plural? `Collections display mode`.
Even if for now, we only have a single collection named `Favorites`. So, to avoid confusion, may be `Favorites display mode` until we get more :).
(Then, the only place we would see the word "Collection" is when holding on a file in the Favorites list, which would be ok - if I didn't miss any other.)

frontend/readcollection.lua
```diff
@@ -3,15 +3,17 @@ local LuaSettings = require("luasettings")
...
+local DEFAULT_COLLECTION_NAME = "favorites"
 local collection_file = DataStorage:getSettingsDir() .. "/collection.lua"
 
 local ReadCollection = {}
 
-function ReadCollection:read(collections)
-    local collection_handle = LuaSettings:open(collection_file)
-    local coll = collection_handle:readSetting(collections) or {}
+function ReadCollection:read(collection_name)
+    if not collection_name then collection_name = DEFAULT_COLLECTION_NAME end
```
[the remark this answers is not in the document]
robert00s: Because all that functions begin with 
`local coll = self:read(collection_name)` 
so we don't need to use
`if not collection_name then collection_name = DEFAULT_COLLECTION_NAME end` 
second time.

robert00s:
rev4:
- prevent crash when we hold on folder,
- hold on folder in cover mode should show original buttondialog (from classic mode),
- should fix #5557 

@poire-z 
Please check changes in last commit :)

poire-z:
Looks simple enough :)
I'm ok with letting the original dialog on directories.
I'd be inclined to hide the test into bookinfomanager.lua,, and do it that way (so we can now that for other stuff if needed):
```diff
--- a/plugins/coverbrowser.koplugin/bookinfomanager.lua
+++ b/plugins/coverbrowser.koplugin/bookinfomanager.lua
@@ -268,5 +268,6 @@ function BookInfoManager:getBookInfo(filepath, get_cover)
     -- returns a bookinfo like-object enough for a correct display and
     -- to not trigger extraction, so we don't clutter DB with such files.
-    if not DocumentRegistry:hasProvider(filepath) then
+    local is_directory = lfs.attributes(filepath, "mode") == "directory"
+    if is_directory or not DocumentRegistry:hasProvider(filepath) then
         return {
             directory = directory,
@@ -278,4 +279,6 @@ function BookInfoManager:getBookInfo(filepath, get_cover)
             ignore_meta = "Y",
             ignore_cover = "Y",
+            -- for CoverMenu to *not* extend the onHold dialog:
+            _is_directory = is_directory,
             -- for ListMenu to show the filename *with* suffix:
             _no_provider = true
diff --git a/plugins/coverbrowser.koplugin/covermenu.lua b/plugins/coverbrowser.koplugin/covermenu.lua
index fd4bbf0b..19abde4c 100644
--- a/plugins/coverbrowser.koplugin/covermenu.lua
+++ b/plugins/coverbrowser.koplugin/covermenu.lua
@@ -205,6 +205,6 @@ function CoverMenu:updateItems(select_number)

                 local bookinfo = BookInfoManager:getBookInfo(file)
-                if not bookinfo then
-                    -- If no bookinfo (yet) about this file, let the original dialog be
+                if not bookinfo or bookinfo._is_directory then
+                    -- If no bookinfo (yet) about this file, or it's a directory, let the original dialog be
                     return true
                 end
```

poire-z:
(Little luacheck warning in case you didn't see it: `plugins/coverbrowser.koplugin/covermenu.lua:9:7: unused variable 'lfs'`)

poire-z:
@Frenzie : can we merge this PR? Or do you prefer we wait after the 2019.11 release (coming soon?), because of translations (this PR adds a few new strings like "Favorites" that are quite top level in the various menus) - or will translators have enough time to translate them?

Frenzie:
Sure, go ahead if you guys think it's ready. About 4 days is enough for quite a few translations.

boriar:
I know that at the moment this is only a favorite list but, ¿will it be a full collection's support?

PR #5528, opened 2019-10-24
[statistic plugin] Don't generate md5 sum on open document

robert00s:
There is no need to generate a md5 sum for documents that were already opened.

### DIFF
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -174,7 +174,9 @@ function ReaderStatistics:initData()
     self.data.series = book_properties.series
 
     self.data.pages = self.view.document:getPageCount()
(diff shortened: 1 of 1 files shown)


plugins/statistics.koplugin/main.lua
```diff
@@ -174,7 +174,9 @@ function ReaderStatistics:initData()
     self.data.series = book_properties.series
 
     self.data.pages = self.view.document:getPageCount()
-    self.data.md5 = self:partialMd5(self.document.file)
+    if not self.data.md5 then
```
Frenzie: Should it also check that it's not an empty string?

robert00s: On initialize `self.data.md5` should be `nil` so we don't need to check that string is empty.

PR #5537, opened 2019-10-26
Unify battery meters look'n feel

### LINKED ISSUE #5535: "Charging" symbol is stuck.
KOReader version: v2019.09.3
Device: Kobo Aura One

Restarting KOReader after updating Nickel, I find the battery indicator is stuck: it stays on charging and 46%.  Restarting the device doesn't help, nor does plugging and unplugging the power cable (and waiting after each such change).
At some point the value changed to 47%, so I guess it is just the charging symbol that is stuck. Taking it off the cable doesn't make the lightning symbol disappear.

NiLuJe:
We unfortunately lack a battery glyph, so we'll make do with the power
glyph instead.

A vertical lignting bolt suffix now always means charging.

Fix #5535

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -52,7 +52,7 @@ local symbol_prefix = {
     icons = {
         time = "⌚",
         pages_left = "⇒",
-        battery = "⚡",
+        battery = "⏻",
         percentage = "⤠",
         book_time_to_read = "⏳",
         chapter_time_to_read = "⤻",
@@ -85,7 +85,7 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].battery
         local powerd = Device:getPowerDevice()
-        return prefix .. " " .. (powerd:isCharging() and "+" or "") .. powerd:getCapacity() .. "%"
+        return prefix .. " " .. powerd:getCapacity() .. "%" .. (powerd:isCharging() and " ⚡" or "")
     end,
     time = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -116,7 +116,7 @@ local footerTextGeneratorMap = {
                 end
             end
         end
-        return prefix .. " " .. batt_lvl .. "%"
+        return prefix .. (symbol_type == "icons" and "" or " ") .. batt_lvl .. "%"
     end,
     time = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -58,7 +58,7 @@ local symbol_prefix = {
         chapter_time_to_read = "⤻",
         frontlight = "☼",
         mem_usage = "",
-        wifi_status = "⚟",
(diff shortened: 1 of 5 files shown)


NiLuJe:
batt_both

Includes a completely gratuitous use of ⌁ instead of an em-dash because it had "electric" in the name and looked mildly cool ;p.

NiLuJe:
In fact, I vaguely considered dropping the power icon from the topmenu and just keeping the electric arrow.

But we'd be back to square one re: footer & topmenu not matching.

(And using the electric arrow in the footer would probably look weird, as it's horizontal, and tiny).

NiLuJe:
Okay, I really wasn't a huge fan of the power icon, so splurged for new icons (https://github.com/koreader/koreader-fonts/pull/7).

batt_nerd

On the upside, we get a different icon every 10%, and a fixed-width icon even when charging.

NiLuJe:
Post whitespace squish:

batt_nerd2

Biep:
Looks good too!

ptrm:
That looks great. Any chance for no-percentage-just-icon option ;) ?

poire-z:
Well, I'm sorry to always be the one to come up with that kind of comment, but I don't like that big black battery icon in the footer :|
I just had a chance to test in on my Android phone (I'll probably patch it out on my Kobo), and the 100% battery icon is just too big and solid black that it catches/attract my eye when reading (there's this black insect at the bottom among the usual grey-like footer text and other icons that I can't help noticing - I'll baptize that effect: Peripheral vision pollution :).
It's less bothering when the battery is at 20% when it's less black :) (it should be the other way around: don't catch my eye when 100% but catch it when i'm reaching 20% and I need to charge it :)

robert00s:
>Well, I'm sorry to always be the one to come up with that kind of comment, but I don't like that big black battery icon in the footer :|

I have the same opinion.

NiLuJe:
I'm in the middle of a minor computer rebuild, so feel free to add an
option to enforce the outline icon in the footer ;). See the link earlier
for the exact codepoint.

poire-z:
> feel free to add an option to enforce the outline icon in the footer ;). 

`[x] Ditch that ridiculously big solid black battery icon` ? That's an accurate wording, but a bit too long I'm afraid :)
(I really think it's a bad icon/design - and as distracting as the white Kobo logo at the bottom of the Kobo Clara, vs the black on black Kobo logo on the GloHD. Yes, people could/would get used to it, but that's just imposing uneeded fancyness vs leanness.)

Anyway, if people really like/want it, it probably shouldn't just be an option for that icon, but may be 2 icons sets (and I'd rather not even have that tall outline battery icon, but something a bit smaller).
So, we might want to think about other "animated/dynamic" icons if we want to propose 2 icons set: a lean one, and a more fancy/dynamic one.

Btw, if your unicode codepoints for the various battery level symbols are consecutive, you might want to swap your 2 long if/else branches for something like:
https://github.com/koreader/koreader/blob/a5419439ce8115a8571a8c78c6735618c82c270b/frontend/ui/translator.lua#L486-L487

Biep:
This is going towards _themes_: people could choose a set of characters (be they initials, emojis, or whatever) to show the various menu and status bar options.  It would just be a variable string, indexed by the code that created the UI elements.

Thra11:
> don't catch my eye when 100% but catch it when i'm reaching 20% and I need to charge it :)

Sounds like you want to hide the battery status when it's above the low battery threshold 😉 😛

poire-z:
> Sounds like you want to hide the battery status when it's above the low battery threshold 

No, I really don't :) I want to see 92% > 77%  > 51% happening - I just don't want a black solid battery icon shouting "look at me, things are happening !" :)

PR #5538, opened 2019-10-27
Minor fixes after #5537

NiLuJe:
* Restore the '+' prefix when charging when the footer is using letter prefixes.
* Use a battery icon with a ? in the *menu* entry for the footer, to make it clear that the one displayed here isn't in sync, unlike the actual footer.


NiLuJe:
> Screenshot_2019-10-27_18-48-17

Tofu! :D

Frenzie:
It's just not something I see very often; I've taken special care to install a variety of fonts ([like so](https://fransdejonge.com/2012/02/debian-international-fonts/) but I'm sure the fonts and/or package names I jotted down there are outdated by now). ;-)

NiLuJe:
Yeah, that's because nerdfonts populates the Private Use/Corporate Use Blocks in a fairly idiosyncratic way.

PR #5546, opened 2019-10-29
Fix a deadlock with empty settings

NiLuJe:
Might be emulator only. Can be easily replicated by deleting `settings.reader.lua*`, `settings/*` & `help/*` (recursively).

Opening the newly generated quickstart guide will deadlock in docsettingtweak:main:97 by busy-looping forever, because `directory == next_directory == base` (`== .`).

### DIFF
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -43,8 +43,6 @@ sdl2 lua@5.1 luarocks gettext pkg-config wget md5sha1sum
 echo 'export PATH="/usr/local/opt/gettext/bin:$PATH"' >> "$HOME"/.bash_profile
 ```
 
-If you run into a gettext error while building glib, try `brew link --force gettext` to override the built-in macOS BSD gettext with GNU gettext.
-
 *Note:* With current XCode versions, you *will* need to set a minimum deployment version higher than `10.04`. Otherwise, you'll hit various linking errors related to missing unwinding libraries/symbols.
 On Mojave, `10.09` has been known to behave with XCode 10, And `10.14` with XCode 11. When in doubt, go with your current macOS version.
 ```
@@ -76,6 +74,8 @@ To run KOReader on your development machine:
 ```
 ./kodev run
 ```
+
+*Note:* On non-Linux hosts, you might want to pass `--no-build` to prevent re-running the buildsystem, as incremental builds are probably not behaving properly.
 
 
 You can specify the size and DPI of the emulator's screen using
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -75,7 +75,7 @@ To run KOReader on your development machine:
 ./kodev run
 ```
 
-*Note:* On non-Linux hosts, you might want to pass `--no-build` to prevent re-running the buildsystem, as incremental builds are probably not behaving properly.
+*Note:* On Mac and possibly other non-Linux hosts, you might want to pass `--no-build` to prevent re-running the buildsystem, as incremental builds may not behave properly.
 
 
 You can specify the size and DPI of the emulator's screen using
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -75,7 +75,7 @@ To run KOReader on your development machine:
 ./kodev run
 ```
 
-*Note:* On non-Linux hosts, you might want to pass `--no-build` to prevent re-running the buildsystem, as incremental builds are probably not behaving properly. Might also need to send a ^C if the very first startup hangs.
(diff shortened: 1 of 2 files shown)


NiLuJe:
Not quite sure this does what it ought to, but at least it doesn't deadlock anymore.

Ping @yparitcher

plugins/docsettingtweak.koplugin/main.lua
```diff
@@ -98,7 +98,13 @@ function DocSettingTweak:onDocSettingsLoad(doc_settings, document)
                 doc_settings.data = util.tableDeepCopy(directory_defaults:readSetting(directory))
                 break
             else
-                directory = FFIUtil.dirname(directory)
+                next_directory = FFIUtil.dirname(directory)
+                if directory == next_directory then
```
yparitcher: good catch i forgot to test the `/` or `.` case.
there is no need for this:
```
next_directory = FFIUtil.dirname(directory)
```

```suggestion
                if directory == "/" or  directory == "." then
                    -- have reached the root, abort
```

yparitcher:
probably only happened on the emulator when the default `home` == `.`
on other devices it has a path so this loop can never reach`.`

yparitcher:
i didn't get these github sugestions to do what i want this is what i meant

```diff
-                directory = FFIUtil.dirname(directory)
+                if directory == "/" or directory == "." then
+                    -- have reached the filesystem root, abort
+                    break
+                else
+                    directory = FFIUtil.dirname(directory)
+                end
```

NiLuJe:
Would adding **..** to the list make sense, too, however unlikely that may be in practice?

doc/Building.md
```diff
@@ -77,6 +75,8 @@ To run KOReader on your development machine:
 ./kodev run
 ```
 
+*Note:* On non-Linux hosts, you might want to pass `--no-build` to prevent re-running the buildsystem, as incremental builds are probably not behaving properly.
```
Frenzie: How do you mean?

Frenzie: I wouldn't appreciate that. ^_^ (Even on my slow laptop it's less than half a second's delay when nothing needs to be done.)

Anyway, surely that just means the Mac build somehow depends on the wrong files? Weird though.

NiLuJe: No idea, so I figured a warning in the docs would be good enough ;).

Frenzie: Since it seems to be Mac/Xcode that broke itself and afaik not *BSD, Windows, etc. ;-)
```suggestion
*Note:* On Mac and possibly other non-Linux hosts, you might want to pass `--no-build` to prevent re-running the buildsystem, as incremental builds may not behave properly.
```

Frenzie: Occasionally something might get slightly confused when a dependency changes; does it still happen if you clear out the thirdparty/lua-spore/build/x86_64-etc subfolder? (Or just the whole build one if you're feeling lazy but then it'll needlessly redownload.)

yparitcher: yes, i narrowed it down.

Cmake expects:
build/x86_64-pc-linux-gnu-debug/rocks/lib/luarocks/**rocks**/lua-spore/0.3.1-1/lua-spore-0.3.1-1.rockspec

my luarocks (luarocks 3.2.1) creates:
/AUR/koreader/base/build/x86_64-pc-linux-gnu-debug/rocks/lib/luarocks/**rocks-5.1**/lua-spore/0.3.1-1/lua-spore-0.3.1-1.rockspec

this is due to luarocks 3.x supporting one luarocks with multiple lua versions.
debian and friends still use 2.x

when i change https://github.com/koreader/koreader-base/blob/2700c5315d0307221a0c49c8bbb2508bdf9432f0/Makefile.defs#L799
`LUAROCKS_DIR=$(OUTPUT_DIR)/rocks/lib/luarocks/rocks`
to 
`LUAROCKS_DIR=$(OUTPUT_DIR)/rocks/lib/luarocks/rocks-5.1`
it no longer rebuilds.

is there a way to fix this, use `rocks-5.1` if luarocks is >= 3.x, i cant find a way to change luarocks behavior.
should i open an issue in base?

yparitcher: see https://github.com/luarocks/luarocks/blob/master/CHANGELOG.md#user-visible-changes for more info

Frenzie: > should i open an issue in base?

Yes please. Some random comment in an unrelated issue is hard to find.

yparitcher: see koreader/koreader-base#1003

Frenzie: That's not an issue. :-P

Issue #5547, opened 2019-10-29
Android crash on startup


Frenzie:
* KOReader version: 21c5d58 but it goes back earlier
* Device: Android

#### Issue
I haven't had the time to run ADB yet. Nothing immediately jumps out to me about the commits, maybe one of those base things like https://github.com/koreader/koreader-base/pull/997?

https://github.com/koreader/koreader/compare/v2019.10...21c5d58

NiLuJe:
I had -19-g8d28164 running fine, FWIW ;).

NiLuJe:
And the latest nightly still works on my end? (Android P).

Frenzie:
Really? Hm, that's very odd. I noticed it a couple days ago on my x86 phone and double checked today on my arm phone with the latest build.

If it was just an x86 thing I would've just quietly fixed it but since I may not be able to look at it for another few days I figured I'd open an issue in the meantime.

Frenzie:
Alright, I'm at my PC with debugging tools ready. Looks like it's caused by #5508.

```
10-29 17:42:36.819  3138  3155 E KOReader: [NativeThread]: failed to run script: frontend/apps/filemanager/filemanagerutil.lua:33: attempt to index local 'home_dir' (a function value)
```

Frenzie:
So that must mean caused by #5488, exposed to me by #5508.

NiLuJe:
`local home_dir = G_reader_settings:readSetting("home_dir") or filemanagerutil.getDefaultDir()`

->

`local home_dir = G_reader_settings:readSetting("home_dir") or self:getDefaultDir()`

Or something?

(I do have a home dir set, which explains why I'm not seeing it).

Frenzie:
I think it should just read this:

```diff
diff --git a/frontend/apps/filemanager/filemanagerutil.lua b/frontend/apps/filemanager/filemanagerutil.lua
index a2bcf019..5691b958 100644
--- a/frontend/apps/filemanager/filemanagerutil.lua
+++ b/frontend/apps/filemanager/filemanagerutil.lua
@@ -11,7 +11,7 @@ local filemanagerutil = {}
 
 function filemanagerutil.getDefaultDir()
     if Device:isAndroid() then
-        return Device.external_storage
+        return Device.external_storage()
     elseif Device:isCervantes() then
         return "/mnt/public"
     elseif Device:isKindle() then
```

I'll test it momentarily.

NiLuJe:
Or that, yeah :D.

[...]

PR #5551, opened 2019-10-31
CoverBrowser: list mode: better default for files_per_page

poire-z:
Don't use a hardcoded default value of 10 files per page, but compute it from available height, as previously (this reverts a bit from #5525). See https://github.com/koreader/koreader/issues/5547#issuecomment-547525726

### DIFF
--- a/plugins/coverbrowser.koplugin/listmenu.lua
+++ b/plugins/coverbrowser.koplugin/listmenu.lua
@@ -788,13 +788,20 @@ function ListMenu:_recalculateDimen()
     end
     local available_height = self.dimen.h - self.others_height - Size.line.thin
 
-    -- (perpage used to be static and computed from a base of 64px per ListMenuItem,
-    -- which gave 10 items both in filemanager and history on kobo glo hd - now that
-    -- we can change the nb of items, let's start with a default of 10 - as it must
-    -- be known as the initial value by the menu selection widget, but there is not
-    -- enough information there to compute it as we could here).
-    -- local default_per_page = math.floor(available_height / scale_by_size / 64)
-    self.perpage = BookInfoManager:getSetting("files_per_page") or 10
+    -- (Note: we can't assign directly to self.perpage and expect it to
+    -- be 'nil' if it was not defined, as we'll find instead the value
+    -- defined in the Menu class (14) because of inheritance.)
+    local files_per_page = BookInfoManager:getSetting("files_per_page")
+    if files_per_page then
+        self.perpage = files_per_page
+    else
+        -- perpage used to be static and computed from a base of 64px per ListMenuItem,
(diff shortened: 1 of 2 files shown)


poire-z:
Some small issue noticed on my Android phone: the items on the first page of file browser get saved with coversizetag `s12`, while those on the next pages get saved with `s12.0` - so, when back on the first page, thumbnails get refreshed to then be properly saved with `s12.0`.
```
sqlite> select cover_sizetag from bookinfo;
s12

s12
s12
s12
s12.0
s12.0
```
I guess we should just make sure to get rid of that `.0` (numbers from math.floor() converted to string get that .0 even when they are integer?)

Frenzie:
Where's the relevant code? I don't think that should happen.

poire-z:
Somewhere around these:
https://github.com/koreader/koreader/blob/eb8795e4e0449e82f9fcf25564496fc6abfbe4b6/plugins/coverbrowser.koplugin/listmenu.lua#L794-L805
https://github.com/koreader/koreader/blob/eb8795e4e0449e82f9fcf25564496fc6abfbe4b6/plugins/coverbrowser.koplugin/bookinfomanager.lua#L240-L248

I guess directly after the math.floor(), we have a plain `s12` - but what is saved in the config table is `12.0` , so, later, when getting it from the sqlite db, we get a string (no notion of number vs string in sqlite, or our bookinfo db wrapper), which is then converted to a number 12.0 for math stuff, but is still `12.0` when used as a string to make cover_sizetag = `s12.0`.
So, may be I should just make sure its a string `12` and not `12.0` and not a number when I call `BookInfoManager:saveSetting("files_per_page", self.perpage) `

Frenzie:
> I guess directly after the math.floor(), we have a plain s12 - but what is saved in the config table is 12.0

Hm, weird. I think math.floor should just be 12. I'll throw in some print/debug statements later I suppose.

poire-z:
Might be how lua-ljsqlite3 do things:
https://github.com/koreader/koreader-base/blob/db0f8e31feb02710a18352b97113fb49286fe2ae/thirdparty/lua-ljsqlite3/init.lua#L236-L244

poire-z:
In our sqlite schema:
```
        cover_sizetag       TEXT,     -- 'M' (Medium, MosaicMenuItem) / 's' (small, ListMenuItem)
[...]
        cover_w             INTEGER,  -- blitbuffer width
        cover_h             INTEGER,  -- blitbuffer height
[...]
CREATE TABLE config (
        key TEXT PRIMARY KEY,
        value TEXT
    );
```
In the cover_w and cover_h columns, we have nice integers (40, 105, 186).
As config.value is TEXT, I guess I should just provide a string for files_per_page, and do some from/to integer conversion myself outside.

Issue #5545, opened 2019-10-28
This is how Arabic characters appear after the last update on kobo aura one :(


hgrain86:
* KOReader version: v 2019.10
* Device: kobo aura one

#### Issue
Arabic letters look mirrored
I cant find arabic language in the languages list
#### Steps to reproduce
![photo_2019-10-29_00-01-38](https://user-images.githubusercontent.com/51149380/67721490-4f010180-f9df-11e9-8eb6-de651d50ef7a.jpg)
![photo_2019-10-29_00-01-33](https://user-images.githubusercontent.com/51149380/67721491-4f999800-f9df-11e9-87fb-bf65ee462f2a.jpg)

##### `crash.log` (if applicable)
`crash.log` is a file that is automatically created when KOReader crashes. It can
normally be found in the KOReader directory:

* `koreader/` directory for Kindle
* `.adds/koreader/` directory for Kobo
* `applications/koreader/` directory for Pocketbook

Please try to include the relevant sections in your issue description.
You can upload the whole `crash.log` file on GitHub by dragging and
dropping it onto this textbox, or you can use one of these paste services.
Please don't forget to set a high expiration date if you do.
https://pastebin.com https://slexy.org https://paste2.org https://fpaste.org
http://ix.io https://paste.kde.org https://paste.debian.net.

Frenzie:
What do you mean by the last update? As in that part that says "KOReader version:" that you left empty? ^_^

Frenzie:
Also pinging @WaseemAlkurdi

NiLuJe:
As far as the UI is concerned, the symbols font may be polluting some needed PUA codepoints, inserting Noto Sans Arabic UI *above* it in the fallback should take care of that.

It shouldn't have any impact on CRe, though, as CRe shouldn't ever touch that font.

Frenzie:
Btw, if this is in regard to the screenshots from https://github.com/koreader/koreader/issues/5359#issuecomment-544586459 then that's something @poire-z is still working on.

Frenzie:
I've been using that as the default since you introduced it. :-)

WaseemAlkurdi:
@hgrain86 Could you upload or give us a link to some of the files you're using as to reproduce the issue?

WaseemAlkurdi:
@poire-z Boom! It works with FreeSerif! 
I didn't find Noto Sans Arabic UI, but that's because I'm using the latest stable release instead of the nightly. Should be rectified by next month.
[...]

Issue #1929, opened 2016-03-29
Investigate font rendering quality in CREngine


houqp:
Quote from the discussion in #1313:

> As for Koreader vs CoolReader question --- no, it is not because of UI, but for some unknown reason CoolReader seems to renger the glyphs slightly better (for some fonts that I use, like Arno Pro and Academy).

NiLuJe:
AutoHint vs. BCI (or even... unhinted?!)?

cf. https://github.com/koreader/koreader-base/pull/199/commits/ed43255b1912d4e388a096ef0c19027a6bea32ae

frankyifei:
I noticed a little difference. Fonts are handled using fontconfig in CoolReader, does it relevant?

houqp:
AFAIK, frontconfig only helps discovering fontface files and has nothing to do with the rendering bit. What @NiLuJe points out is more likely the cause. @frankyifei were you able to spot the rendering difference from CR and KOR? I haven't had time to test it on my side yet.

frankyifei:
a screenshot, very little difference. 
![fonts](https://cloud.githubusercontent.com/assets/8104251/14132509/33618cb8-f68e-11e5-9d6e-525a81be8d57.jpg)

houqp:
I don't see much difference from the screenshot. @tigran123 did mention it only happens to some specific font faces though.

Frenzie:
@houqp fontconfig can provide the configuration for something like `props->setInt(PROP_FONT_HINTING, 2);` though (not saying it does in our case)

NiLuJe:
If you want to see dramatic differences between bytecode hinting and the
autohinter, check with Consolas, if you have a Win7+ somewhere ;).

As for fontconfig, yeah, it's used to handle choosing font families /
styles, and eventually rendering settings, like hinting. No idea to which
extent CRe actually uses it though.

NiLuJe:
Some differences also tend to be much more marked on e-ink than on an lcd
screen...

NiLuJe:
If CR is using an older freetype version than ours, and the original report
was about a CFF font, and not a truetype one, there have been massive
changes (for the better) when hinting those natively.

houqp:
Looking at lvfntman.cpp, looks like it's just using FC for the following font attributes: ` Font family:%s style:%s weight:%d slant:%d spacing:%d file:%s`.

@NiLuJe isn't auto hinting suppose to produce better result? Looks like coolreader is using autothint as default hinting setting.

[...]

PR #5553, opened 2019-10-31
cre: use 'best' (Harfbuzz) as the default kerning method

poire-z:
It might be slower, but is needed to properly display books in arabic, indic...
See https://github.com/koreader/koreader/issues/5545#issuecomment-548011207 and follow ups.

One thing I'm not sure about, but I often felt when trying "best" that it looked nicer with font hinting set to "off". Our default for font hinting is "auto" (Freetype auto hinting).
@NiLuJe : any opinion/science about that?
(Lately, I've been using Font hinting "off" with everything, as the other options were sometimes triggering some font bugs or something.)

Not really needed, but I'll update https://github.com/koreader/koreader-base/blob/master/cre.cpp#L510 next time I tweak base/.


NiLuJe:
That should be better since the flags fix ;). Switching to "light" render
mode should entirely eradicate discrepancies.

In any case, +1 on best ;).

Frenzie (approved):
👍

poire-z:
Just to be sure, if/when you're using kerning "best", can you check on some of your books what you have for font hinting, if you really have the default "auto" (because I'm not sure best is really best with auto :)

Frenzie:
Not sure if you were asking me, but I think hinting off/kerning best gives the best results, mainly in the form of the best baseline alignment,[1] and the best kerning. Combining one of the other hinting settings with kerning best results in slightly worse kerning and sometimes a forced straight baseline.

[1] I.e., most true to the font's intended design. It allows them to peek slightly under or over as designed.

poire-z:
(I was indeed asking you and @NiLuJe :)
I too find it better with Hinting: off - it avoids some kerning and baseline bugs (at least with my favorite font Bitter). But it makes things more condensed (that Harfbuzz already does, hinting off make that even more noticable), which I like - but people used to fast+auto will be surprised.

NiLuJe:
I'm using unhinted rendering on everything >= 300DPI. On lower density
screens, it depends on the font, and I tend to prefer native hinting then
;).

Frenzie:
I might prefer native hinting in theory, at least on smaller font sizes where 300 DPI doesn't make one lick of a difference with 265 and is still fully pixel bound. But in combination with the effect on kerning in crengine/HarfBuzz I don't in practice.

poire-z:
Strangely, for FreeSans and FreeSerif, best + auto is totally ugly on my small emulator (x), while it's ok on my Kobo. It's fine on both for NotoSans and NotoSerif.
(x: although with FreeSans, Auto seems to correct some ugly `s` or `c` that went below the baseline... it makes it better when using "fast"...)

So, what should we do: let Font hinting to Auto for less drastic changes (and wait for any feedback) - or change it to Off?
(Or make that dependant on device DPI - but my above test with the Free fonts makes me think hinting is not even better on small DPI device.)

Frenzie:
I've never cared for the auto default, nor for its name. Some old related comment: https://github.com/koreader/koreader/issues/1929#issuecomment-333471966 I think it should be relabeled something like `FreeType` or `FT` to indicate it's FreeType autohinting, because to me it implies using native when available.

I wouldn't quite call it ugly though? Just not as good. :-)

poire-z:
> I wouldn't quite call it ugly though?

I'm not that an aesthete regarding text, but I find it really ugly, all these glyphs merged into some black mess (that looks like super bad/off kerning), that we don't see on your screenshots with Off or Native:
<kbd>image</kbd>

> it should be relabeled something like FreeType 

I guess "auto" made it sound less technical.

We could go with "native" for a less drastic condensation than "off".
I remember reading "native" was sometimes buggy with fonts that did not provide native instructions (or provided buggy ones).
And once, I got a font that when switching to "native" made crengine go into some infinite loop somewhere (probably inside FreeType)- I didn't manage to track where things went wrong at the time, and I got rid of the font - I don't remember which one it was :|

Frenzie:
Fine, it's ugly. ;-) But it's legible anyway.

NiLuJe:
Light hinting should fix that ;).

poire-z:
> Light hinting should fix that 

Is that about just replacing the 6 `FT_LOAD_TARGET_NORMAL`  in crengine/src/lvfntman.cpp with `FT_LOAD_TARGET_LIGHT` (https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_load_target_xxx and https://www.freetype.org/freetype2/docs/text-rendering-general.html) ?

Just tested on the emulator, and it looks indeed cleaner! There's now hardly any difference between native and auto (to the point I wondered if it worked :) but there is some subtle ones: I notice only some small changes vertically (the tip of the `t` or `f`, or the dot on a `i` going higher with some fonts with Auto, with some other fonts with Native...). But both, like previously, make the text horizontally streched vs off (that makes it more condensed) - so I guess it works.

NiLuJe:
Yep ;). That is essentially the point: no more hinting on the x axis. Bonus points, that simplifies advances computation, as they're no longer modified by hinting ;).

 The UI could probably also use it in freetype or font (I can't remember which does the load).

poire-z:
> But both, like previously, make the text horizontally streched vs off (that makes it more condensed) 

Well, that happens (now?) only with kerning "off" or "fast" - not with "good" or "best".
So, dunno, it might be we've had a bug with kerning vs hinting: should hinting on the x-asis change that much the spacing between letters, for a same kerning mode?
Anyway, it gets more stable with "best", so that's a plus to go with "best" as the default. And now that "auto" and "native" are not so much different, staying with "auto" is ok. (haven't read/understand everything in the 2nd url above, so I dunno if I understood right that auto may try to use the native hinting first with that load_target_light flag).
And I might not want to fix that bug, as that airy vs condensed feeling that this kerning/hinting bug may give is quite a nice bonus feature (now non-default with this PR).

Anyway, looks fine too on my Kobo. So I'll PR and bump crengine with that tomorrow.

> Bonus points, that simplifies advances computation, as they're no longer modified by hinting ;).

For free? No other code change needed on the crengine side?

Using your [FT cache submsystem ](https://freetype.org/freetype2/docs/reference/ft2-cache_subsystem.html), while sounding good, might need a lot of changes in crengine, which does its own caching - dunno how better or not that would be.
But using it on the UI side might be easier - but may be after my harfbuzz/fribidi - so we can see how to rework that freetype/font code to make it play fine in 2 different contexts.

NiLuJe:
Full autohinting can be fairly aggressive on the x axis, yeah ;). Which was
why not letting harfbuzz follow that along led to some fairly egregious
differences.

Don't worry about the auto/native stuff mentioned with the light docs, I
should have made pretty sure no means no, auto means auto, and native means
native ;).

Oh,and yeah, I was very much only thinking about the UI when I mentioned
the FT cache stuff.

poire-z:
But from https://www.freetype.org/freetype2/docs/text-rendering-general.html:
> Setting ‘slight’ hinting usually leads to FT_LOAD_TARGET_LIGHT. This mode implied the auto-hinter before and has now been changed to mean “Use native vertical-grid-only-snapping if driver and font supports it and vertical-grid-only auto-hinter otherwise”. Right now, only the OpenType/CFF driver is supported. In the future, this will hopefully include the TrueType engine once full support for ClearType arrives.

(That doc might be outdated, dunno).
The first sentence is a bit ... unclear :) Should I read it as _Just use FT_LOAD_TARGET_LIGHT to enable 'slight' hinting_?
The next sentences seem to suggest we could have a 4th hinting mode, a real "auto", that would use native if there are some instructions, and freetype hinting only if there are none, by just using some other combinations of `FT_LOAD_NO_AUTOHINT` and `FT_LOAD_NO_HINTING` ?

> I should have made pretty sure no means no, auto means auto, and native means native

I get that you mean your work on these above flags, right?

NiLuJe:
Yep, you can just switch the target flag, we do the right thing with the
others to make the situation less murky ;).

gilgulgamesh:
Hi! the Light hinting change never got made - I just commented this line in `ffi/freetype.lua` on my Libra 2 and it's substantially better

 ```
-- FT_Load_Glyph_flags = bit.bor(FT_Load_Glyph_flags, ft2.FT_LOAD_FORCE_AUTOHINT)
-- (No hinting, as it would mess synthetized bold)
FT_Load_Glyph_flags = bit.bor(FT_Load_Glyph_flags, ft2.FT_LOAD_NO_HINTING, ft2.FT_LOAD_NO_AUTOHINT)
```
 It might be mixing hinting types currently? 
 
 I can't quite tell how this and the cpp and the overriding is all related, but I think this is the right place to leave this. It's really pretty!

poire-z:
> the Light hinting change never got made
It might be mixing hinting types currently?

Not sure what you're talking about. What's the problem ?
(I just tested on a book: with some fonts, any combination of Font Hinting and Font Kerning light|best results in a slight visible change to the letters drawn, thickness or position).

> I can't quite tell how this and the cpp and the overriding is all related, but I think this is the right place to leave this.

The bit you quoted in freetype.lua is about UI font rendering (our menu, our dict results, etc...), not at all about cre.cpp/crengine that deal with the drawing of EPUBs books pages.

PR #5556, opened 2019-11-01
CoverBrowser: fix files_per_page setting type

poire-z:
Avoid saving/getting "12.0" (as the lua-ljsqlite3 driver does), and save/get "12" instead by properly converting it to a string and back, as the config.value table column is of type TEXT.
See https://github.com/koreader/koreader/pull/5551#issuecomment-548698110


kojishoda:
Thank for helping & support! APPRECIATED

PR #5564, opened 2019-11-02
android: touchscreen toggle

### LINKED ISSUE #5558: Touchscreen toggle with camera button request
* KOReader version: 2019.08 and 2019.10
* Device: Onyx Boox Nova Pro

Currently the camera button toggles the touch off/on and goes a page forward (or back (depending on settings) with each click. Would really like the ability to either toggle that function off/on or turn it off completely as it interferes with the way I use my remote. Camera button mapped to page forward only.  Thank you!

pazos:
Using camera button.

Fixes #5558

### DIFF
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -220,6 +220,13 @@ if Device:isAndroid() then
         end,
     }
 
+    -- camera key events
+    common_settings.android_camera_key = {
+        text = _("Camera key toggles touchscreen support"),
+        checked_func = function() return G_reader_settings:isTrue("camera_key_toggles_touchscreen") end,
+        callback = function() G_reader_settings:flipNilOrFalse("camera_key_toggles_touchscreen") end,
+    }
+
     -- fullscreen toggle on devices with compatible fullscreen methods (apis 14-18)
     if Device.firmware_rev < 19 then
         common_settings.fullscreen = {
--- a/frontend/device/input.lua
+++ b/frontend/device/input.lua
@@ -324,7 +324,8 @@ function Input:handleKeyBoardEv(ev)
     end
 
     -- The hardware camera button is used in Android to toggle the touchscreen
-    if keycode == "Camera" and ev.value == EVENT_VALUE_KEY_RELEASE then
+    if keycode == "Camera" and ev.value == EVENT_VALUE_KEY_RELEASE
+        and G_reader_settings:isTrue("camera_key_toggles_touchscreen") then
         local isAndroid, android = pcall(require, "android")
         if isAndroid then
             -- toggle touchscreen behaviour
--- a/frontend/ui/elements/filemanager_menu_order.lua
+++ b/frontend/ui/elements/filemanager_menu_order.lua
@@ -53,6 +53,7 @@ local order = {
         "back_in_filemanager",
         "enable_back_history",
         "android_volume_keys",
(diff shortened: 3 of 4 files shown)


frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -220,6 +220,16 @@ if Device:isAndroid() then
         end,
     }
 
+    -- camera key events
+    common_settings.android_camera_keys = {
```
poire-z: `android_camera_key` singular as there is only one (even if it kills the symetry with the one above :)

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -220,6 +220,16 @@ if Device:isAndroid() then
...
+        text = _("Camera key toggles touchscreen"),
+        checked_func = function() return G_reader_settings:isTrue("camera_key_toggles_touchscreen") end,
+        callback = function()
+            local current = G_reader_settings:isTrue("camera_key_toggles_touchscreen")
+            G_reader_settings:saveSetting("camera_key_toggles_touchscreen", not current)
```
poire-z: `G_reader_settings:flipNilOrFalse("camera_key_toggles_touchscreen")` makes that a single line.

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -220,6 +220,16 @@ if Device:isAndroid() then
         end,
     }
 
+    -- camera key events
+    common_settings.android_camera_keys = {
+        text = _("Camera key toggles touchscreen"),
```
Frenzie: Maybe something like this?

Camera key toggles ignore mode

pazos: Since I'm not native I'm fine with something that means: "(usar la) tecla de cámara (para) conmuta(r) el panel táctil". The easy translation for that is: **camera key switches/toggles touchscreen** but if you find your suggestion a better translation then is good for me :smile:

poire-z: I find "ignore mode" a bit obscure (what is ignored? history, statistics, page turns, taps, swipes, other keys?) - and it's not all user interactions that are ignored: volume keys to turn pages would still work while touchscreen is disabled.
"toggles touchscreen activation" ?

Frenzie: `Touch ignore mode (camera button)`?

poire-z: Would need the word "toggle" or "disable/enable", otherwise one may think it's the checkbox that set the mode.

pazos: +1 to toggle or disable/enable. Feel free to correct and merge when ready :smiley:

poire-z: "Camera key touchscreen toggling", if we want to have the same grammatical construct as the previous "Volume key page turning" ? (Although even for this one, my limited english skills would more easily understand a sentence like "Use/allow volume keys to turn page" - so I'm fine with this PR that uses a sentence.)

pazos: the first one is described as an option, the second as an action. The first sentence is saying "this to use x key to do y action" while the second assumes that "this" is an option and says "use x key to do y action".

I prefeer the second option too (in both cases).

PR #5568, opened 2019-11-04
Avoid double flash when switching or reloading document

poire-z:
Avoid double flash when switching documents (via gesture or Open previous document menu) or reloading document (switch from alt TOC to normal TOC, or when "styles have changes in such a way that fully reloading the document might be needed").
That's just the 2 place I had noticed that, so no big change.
Pinging @NiLuJe in case I missed something where these 2 flashes could make some sense.


NiLuJe:
Nope, that was definitely one too many ;).

Issue #5555, opened 2019-10-31
Keyboard popup propagates


Frenzie:
* KOReader version: 602a821
* Device: all

#### Issue
The keyboard popup propagates the event. I'm not sure which change is responsible atm but #5380 is probably the most suspicious at a glance.

https://github.com/koreader/koreader/commits/master/frontend/ui/widget/virtualkeyboard.lua

#### Steps to reproduce
Hold on a key and let go.

poire-z:
Possibly related:
- Edit some text file in text editor, go at end of some line:
- Long press on Backspace, to delete the line
- When releasing that long press: the last char of the previous line is also deleted

Frenzie:
Indeed, that sounds like the same issue.

#5380 isn't to blame, but the issue didn't exist in 2019.09 while it does in 2019.10. I've traced the undesired behavior to #5452.

Frenzie:
It seems to me that #5452 breaks the way modal widgets are supposed to work. The top modal widget is supposed to be the top modal widget.

Frenzie:
PS I forgot to add this above, but this would be one way to get rid of the biggest problem. The thing is, if you hover over some random key outside of the popup it will still activate. Complex workarounds would be possible, but shouldn't be necessary because we've got the top modal dialog in play here.

```diff
diff --git a/frontend/ui/widget/virtualkeyboard.lua b/frontend/ui/widget/virtualkeyboard.lua
index 3a7d0992..2e48321c 100644
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -377,6 +377,7 @@ function VirtualKeyPopup:init()
                 virtual_key.onHoldReleaseKey = function()
                     virtual_key:onTapSelect(true)
                     UIManager:close(self)
+                    return true
                 end
                 virtual_key.onPanReleaseKey = virtual_key.onHoldReleaseKey
 ```

poire-z:
Actually, with the current code and my above issue:
> When releasing that long press: the last char of the previous line is also deleted

There were 3 stuff done: delete to start of line when long press. And when releasing: one delete that removed the CR, and another delete that removed the last char of previous line.

Looks like we may have forgotten some `return true` in some evenr handlers:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L243-L259
[...]

PR #5569, opened 2019-11-06
Keyboard for Arabic and languages with Arabic script, initial release

WaseemAlkurdi:
This is an Arabic character keyboard layout. It is based in part on the PC keyboard layout, and in part on the Google keyboard.
I have had to add some keys to the first and second rows.
I haven't tested this, because attempting to compile the build from source results in an error:
```
make[1]: *** No rule to make target 'luafilesystem/src/lfs.c', needed by 'build/x86_64-linux-gnu-debug/libs/libkoreader-lfs.so'.  Stop.
```
Diacritic marks haven't been added, but they could be added in a couple of minutes, should this be tested and proven to work.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5569)


Frenzie:
On which system is that? You can use the extracted AppImage (for just testing Lua stuff) or the Docker image to have a tailored build environment, see https://github.com/koreader/koreader/blob/master/doc/Building.md#setting-up-a-build-environment-for-koreader

Also on systems other than Android it's generally simple to just change a few Lua files.

WaseemAlkurdi:
> On which system is that? You can use the extracted AppImage (for just testing Lua stuff) or the Docker image to have a tailored build environment, see https://github.com/koreader/koreader/blob/master/doc/Building.md#setting-up-a-build-environment-for-koreader
> 
> Also on systems other than Android it's generally simple to just change a few Lua files.

I'm building on desktop Linux. 
So is it okay to just copy the source tree over to the `/mnt/us/koreader` directory on my Kindle? 
And do I have to edit options to enable the Arabic keyboard?

Frenzie:
You still need to add it to this table:

https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L526-L535

Frenzie:
> So is it okay to just copy the source tree over to the /mnt/us/koreader directory on my Kindle?

Sort of. As long as you have the libraries in place that should be fine, but it's normally faster and easier to just overwrite the couple of Lua files you edited.

WaseemAlkurdi:
@Frenzie I did that ... KOReader now crashes. Why is that?
The Lua scripts are error-free, except for warnings about unused variables.

frontend/ui/data/keyboardlayouts/ar_keyboard.lua
```diff
@@ -60,7 +60,7 @@ local alef = ar_popup.alef
 local baa = ar_popup.baa
 local jeem = ar_popup.jeem
 local daal = ar_popup.daal
-local h-aa = ar_popup.h-aa
+local haa2 = ar_popup.haa2
```
Frenzie: Incidentally, underscore (`_`) is fine in a variable name.

WaseemAlkurdi: @Frenzie I know, but I removed them in order to isolate an issue. I'll return them in the next round of cleanups. But for now, the keyboard as is would crash KOReader for some reason.

Frenzie:
> I did that ... KOReader now crashes. Why is that?

I have no idea, because all you've got here is two completely independent files that literally can't affect anything at all. ^_^ The crash.log will tell you more if you want to find out, or you can just overwrite the whole thing with the latest nightly from http://build.koreader.rocks/download/nightly/v2019.10-33-gf1f75c5_2019-11-06/ so you've got a working copy.

Frenzie:
Anyway, I can push a few fix-up commits to your branch if you like.

WaseemAlkurdi:
> Anyway, I can push a few fix-up commits to your branch if you like.

It would be great if you do!

Frenzie:
@WaseemAlkurdi See the two individual commits for the (minor) changes.

Screenshot_2019-11-06_14-28-20

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -377,6 +377,7 @@ function VirtualKeyPopup:init()
                 virtual_key.onHoldReleaseKey = function()
                     virtual_key:onTapSelect(true)
                     UIManager:close(self)
+                    return true
```
Frenzie: Whoops, I didn't mean to commit this line but that's okay. This solves the worst part of #5555.

poire-z:
> KOReader now crashes. Why is that? The Lua scripts are error-free,

See your crash.log, you should have a stack trace that will tell you (or us) why it crashes, and what must have happened for it to crash even if your lua files are error-free (truncated transfer, other file touched...)

About AppImage: https://github.com/koreader/koreader/wiki/Installation-on-desktop-linux
(@Frenzie : Looks like this #4304 got lost in the Wiki reorganisation.)

Also, we'll have to think about these HE/AR keyboards: is it fine that you start with querty, and you have to it the little Globa icon key to have the HE/AR keys? Or should we start with these, and get qwerty with that Globe icon key instead. See https://github.com/koreader/koreader/pull/5414#issuecomment-533910298 and around.

Frenzie:
@poire-z Not lost, that's at the top of https://github.com/koreader/koreader/blob/master/doc/Building.md#setting-up-a-build-environment-for-koreader

Frenzie:
@WaseemAlkurdi Now you'll need to `git pull` my commits into your branch before you continue working.

frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua
```diff
@@ -0,0 +1,165 @@
...
+            { "/",        "/",    "’",    "↓",    "ֽ ",    thaa,   "Ю",    "ю",    "Ç",    "ç",    "…",    "¨", },
+            { label = "مَسح",
+              icon = "resources/icons/appbar.clear.reflect.horizontal.png",
+              width = 1.5
+            },
+        },
+        -- fourth row
+        {
+            { "Sym",     "Sym",  "ABC",  "ABC",  "Sym",  "رمز",    "رمز",  "رمز",  "Sym",  "Sym",  "ABC",  "ABC",
+              width = 1.5},
```
poire-z: You could try reducing this width to = 1 to not have these keys overflow the keyboard :)

NiLuJe:
The original build error vaguely looks like you haven't pulled the submodules?

Also, I've posted the one-liner I use to scp updated files to my devices a couple of times around. I'm unfortunately not on my desktop right now, but it's basically based on the list that `git diff --name-only` outputs ;).

Frenzie:
A `./kodev run` should do that automatically though.

frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua
```diff
@@ -0,0 +1,165 @@
...
+            { ".",        ".",    "’",    "↓",    "ֽ ",    zay,   "Ю",    "ю",    "Ç",    "ç",    "…",    "¨", },
+            { "/",        "/",    "’",    "↓",    "ֽ ",    thaa,   "Ю",    "ю",    "Ç",    "ç",    "…",    "¨", },
+            { label = "مَسح",
+              icon = "resources/icons/appbar.clear.reflect.horizontal.png",
+              width = 1.5
+            },
+        },
+        -- fourth row
+        {
+            { "Sym",     "Sym",  "ABC",  "ABC",  "Sym",  "رمز",    "رمز",  "رمز",  "Sym",  "Sym",  "ABC",  "ABC",
```
poire-z: Looks like these ر`مز",    "رمز",  "رمز"` (raaah, so hard to cut and paste and wrap, bidi hell!) need to be added to:
`symbolmode_keys = {["Sym"] = true, ["ABC"] = true, ["رمز"] = true},` on line 91, for them to work.

(That's going to be my new swear word :  _that bidi-mary is bidi disgusting on this bidi sunday_ :)

poire-z:
Otherwise, seems to work ok, and good to let me see I have some bidi (pronounce bloody) ugly cursor positionning issues to solve (when entering latin, then entering arabic, cursor stays at the right of the arabic word that gets letter added on its left...

Frenzie:
@poire-z I just played around with that a little in Firefox (which presumably does it mostly right?) and that cursor/selection stuff is pretty weird & complex. :-)

WaseemAlkurdi:
@NiLuJe
This might have been the cause, since I did `./kodev run` behind a firewall, which might've interfered. Anyway, I'm testing on my Kindle instead.

@poire-z
> Also, we'll have to think about these HE/AR keyboards: is it fine that you start with querty, and you have to it the little Globa icon key to have the HE/AR keys? Or should we start with these, and get qwerty with that Globe icon key instead. See #5414 (comment) and around.

In my opinion, the localized keyboard should appear before the QWERTY layout. This is to avoid users' confusion (_I've set the keyboard to Arabic, yet the keyboard appeared in English!_), and also because the globe icon's meaning is obvious when you have a localized keyboard (get back to QWERTY).

I've done the changes that you and @Frenzie have suggested, and fixed the missing characters. It should now be ready ... except for the diacritics. Where do I place them? We have nine diacritics in Arabic, which makes `north/northeast/...` a non-starter (six directions only). I thought of placing them in the umlaut character selector ... but I don't know if that's possible.

And the spacebar ... localizing that string would cause the QWERTY keyboard's Space to be localized too.

What should I do?

Frenzie:
> In my opinion, the localized keyboard should appear before the QWERTY layout.

Then please do that. I don't understand why people (you included ;-) ) do it the other way around but besides ensuring some basic code sanity the specifics of the keyboard are the implementer's.

> What should I do?

There are actually 8 reachable by swipe and 11 options total.

Screenshot_2019-11-08_14-56-09

WaseemAlkurdi:
@poire-z @Frenzie

Yeah, the RTL caret is _seriously_ complicated ... but I have somehow got used to it, by simply moving in the _counterintuitive_ direction :-D

Frenzie:
@WaseemAlkurdi The table behind that popup:

https://github.com/koreader/koreader/blob/98555908c8fb1eeb9a7985443da32cd6fc87f085/frontend/ui/data/keyboardlayouts/keypopup/en_popup.lua#L129-L142

poire-z:
> In my opinion, the localized keyboard should appear before the QWERTY layout
> Then please do that.

I agree.
Also, having played with your keyboard for a few days, one annoying thing was that the cursor move keys were only accessible from the arabic layout after 2 layout keys tap :) So, I guess they should be on the layout we get with the Sym key, with the digits (1234...) keys (I avoid saying "arabic numbers" to avoid confusion :)
Do arabic letters have the notion of lowercase/uppercase ? If not, the Shift layout is all yours, to put some variations of the original letters, possibly the most used diacritic combination, or the diacritic themselves (as currently).
But these diacritics combined with their letter should also be in the popup for that letter. Those that are below the char in the bottom row, those above in the top rows, etc...
Can all the arabic diacritics apply to all arabic letters? Or are there some forbidden/rarely used combinations (that unicode does not prevent I guess).

Also, you reduce the width of the Sym key (which was fine and does not need to be that small), and not of the overflowing Shift and Backspace keys:
image

There is a bunch of duplicated down cursor keys :) you may want to only set one at the correct position, and add another symbol (non duplicated) to the duplicated ones:
image

image what does this word mean :)

yparitcher:
> @poire-z
> 
> > Also, we'll have to think about these HE/AR keyboards: is it fine that you start with querty, and you have to it the little Globa icon key to have the HE/AR keys? Or should we start with these, and get qwerty with that Globe icon key instead. See #5414 (comment) and around.
> 
> In my opinion, the localized keyboard should appear before the QWERTY layout. This is to avoid users' confusion (_I've set the keyboard to Arabic, yet the keyboard appeared in English!_), and also because the globe icon's meaning is obvious when you have a localized keyboard (get back to QWERTY).
> 

in order to properly support multiple input languages i think:

- each keyboard should only have one language (arabic & hebrew should not have english characters, the english should not have crylic characters)

- the keyboard layout menu should allow for multiple selections and a default.

- when pressing the globe icon the keyboard should rotate between selected keyboards (or show a popup to select a language), or nothing if only one keyboard is allowed. 

this would require some refactoring of the keyboard.

this is how most touch keyboards that i have seen do it.

Frenzie:
> the keyboard layout menu should allow for multiple selections and a default.

That's always been the goal, please feel free to implement that if you have the time. See #5318.

WaseemAlkurdi:
@poire-z
> I agree.

How do I set it to appear first though? Should I remove the English letters in `ar_AA_keyboard.lua`?

> Also, having played with your keyboard for a few days, one annoying thing was that the cursor move keys were only accessible from the arabic layout after 2 layout keys tap :) 

That's why it should be the first layout. If you point me as how to do it, I'll do it ASAP.

> So, I guess they should be on the layout we get with the Sym key, with the digits (1234...) keys (I avoid saying "arabic numbers" to avoid confusion :)

Great! I had the same thought ... it's as if the gap in the fourth row was specifically reserved for these!

> Do arabic letters have the notion of lowercase/uppercase ? 

Thankfully not :-)

> If not, the Shift layout is all yours, to put some variations of the original letters, possibly the most used diacritic combination, or the diacritic themselves (as currently).

That's a really good idea.
Arabic diacritics in computing are considered separate characters, so that wouldn't be an issue.

> But these diacritics combined with their letter should also be in the popup for that letter.
Not strictly necessarily. It would be nice, but lots of extra coding for something only cosmetically significant. We can use a "generic" letter as a placeholder.

> Those that are below the char in the bottom row, those above in the top rows, etc...

Brilliant.

> Can all the arabic diacritics apply to all arabic letters? Or are there some forbidden/rarely used combinations (that unicode does not prevent I guess).

Yep, Unicode doesn't prevent any Arabic character from having any diacritic (you can even put them on English letters :-D )
But in Arabic, the two forms of _alef_ `ا ى ` don't take diacritics.

> Also, you reduce the width of the Sym key (which was fine and does not need to be that small), and not of the overflowing Shift and Backspace keys:

Will do!

> There is a bunch of duplicated down cursor keys :) you may want to only set one at the correct position, and add another symbol (non duplicated) to the duplicated ones:

Will do that as well.

> image what does this word mean :)

That's the word "Symbol" translated. (But it's not connected and HarfBuzz'd. If implementing HarfBuzz on the keyboard just for one key isn't worth it, I'll keep it untranslated. Your call.)

WaseemAlkurdi:
> @WaseemAlkurdi The table behind that popup:
> 
> https://github.com/koreader/koreader/blob/98555908c8fb1eeb9a7985443da32cd6fc87f085/frontend/ui/data/keyboardlayouts/keypopup/en_popup.lua#L129-L142

Won't that screw up the symbols in English too?
I think that @poire-z's solution works best. 
Or, as an alternative, adding the Arabic diacritics to the Shift panel (as on PC keyboards, where Shift + letter produces a diacritic symbol instead)

Frenzie:
Er, you don't change the English anything. @poire-z wrote literally the exact same thing I did. >_>

Frenzie:
> We have nine diacritics in Arabic, which makes north/northeast/... a non-starter (six directions only).

^ That was a direct reply to this. As in "here's how you do 11". Not as in "change this". ;-)

Frenzie:
> How do I set it to appear first though? Should I remove the English letters in ar_AA_keyboard.lua?

Whether you want to remove the Latin alphabet is your call, but each indexed table entry corresponds to a keyboard.

```lua
            { _Q_,        _q_,    "„",    "0",    "Й",    "й",    "?",    "!",    "Å",    "å",    "1",    "ª", },
```

So in keyboard 1, show Q, keyboard 2, show q, keyboard 3, show „, etc.

1/2 (Q/q) is the first keyboard with and without shift,  3/4 is Symbol, 5/6 is "international", 7/8 is special symbols and then there's some way too complex combinations of symbols & international/special.

I believe you should be fairly free to choose to which extent you want to include any of those, if not ping me about it. Since you don't need shift in Arabic I'd probably just remove that button from the layout altogether, keeping only regular, symbol, and special symbol (and possibly "international").

> If implementing HarfBuzz on the keyboard just for one key isn't worth it, I'll keep it untranslated. Your call.

It doesn't need implementing "on the keyboard." It'll just be implemented on textwidgets, and the keyboard uses 'em just the same as any other part of the program. :-)

poire-z:
> How do I set it to appear first though? Should I remove the English letters 

Not remove, but invert them.
```
  \/-------------<-------------\
{ _Q_,        _q_,    "„",    "0",    "Й",    "й",    "?",    "!",    "Å",    "å",    "1",    "ª", },
  \______________>_____________/\
```
So, quite boring to do for each row :)

> Won't that screw up the symbols in English too?

No, that's the popup for `_e_`, and it applies only to the layout that has that `_e_`. So you can have different popup on the same key-square in different layouts.

> Or, as an alternative, adding the Arabic diacritics to the Shift panel 

You can have both :) The Shift panel for the standalone diacritics, as arabic has no uppercase, and the popups for applied diacritics to the letter of that key.

> If implementing HarfBuzz on the keyboard just for one key isn't worth it

(Dunno the context of that comment, can't find it :|)
You can have multiple unicode chars on a key. They should display shaped by Harfbuzz, and add the multiple unicode chars to the text, which will too be shaped by Harfbuzz - as everything else, once we switch the text rendering code to use Harfbuzz/FriBiDi. So you can do that, even if it displays bad currently. I should have that ready by next week.

WaseemAlkurdi:
@Frenzie

>  @poire-z wrote literally the exact same thing I did. >_>

After giving it some thought, it turned out that it is. Apologies :-(

> 1/2 (Q/q) is the first keyboard with and without Shift, 3/4 is Symbol, 5/6 is "international", 7/8 is special symbols and then there's some way too complex combinations of symbols & international/special.

In my most recent commit, I tried two approaches to the issue of removing Latin and Shift, because I wasn't really sure which way is correct. Removing the whole column did _not_ work, for some reason.
* In the first row of the keyboard, I duplicated the second Arabic column (the one without Shift).
* In the second and third row, I replaced every character in the row with `""`.
Which approach do you recommend?

@poire-z

> No, that's the popup for _e_, and it applies only to the layout that has that _e_. So you can have different popup on the same key-square in different layouts.

Noted!

> (Dunno the context of that comment, can't find it :|)

I edited out that part of the comment, as it was kind of embarrassing not to have realized that from the beginning ;-(

WaseemAlkurdi:
Now, as to the remaining issues:
* Running `luacheck` on the files would produce a _lot_ of errors. But these are instances of two actual errors. One is about unused variables (which is expected for removed Latin characters ... more on that coming up later), but the other is:
```
    ../ar_AA_keyboard.lua:168:121: line is too long (207 > 120)
```
But the line _isn't_ actually that long. So what seems to be the issue?
* On line 168 of `ar_AA_keyboard.lua`, we have my diacritics key. I've basically copied the line for `prd` and changed the reference to `prd` to point at `diacritics` in `ar_popup.lua` ... but for some reason, the word `diacritics` has to be repeated eight times. Is that normal? It really feels "unclean".
There is also the `label = "َ ُ ِ"` at the beginning of the line. Replacing this with `diacritics` would cause the `"َ ُ ِ"` to be typed (which we don't want ... this is only to indicate a character picker), while replacing with `label = diacritics` would cause a crash with this message in the log:
```
./luajit: frontend/ui/rendertext.lua:37: bad argument #1 to 'len' (string expected, got table)
stack traceback:
	[builtin:len]: in function 'len'
	frontend/ui/rendertext.lua:37: in function '(for generator)'
	frontend/ui/rendertext.lua:174: in function 'sizeUtf8Text'
	frontend/ui/widget/textwidget.lua:45: in function 'updateSize'
	frontend/ui/widget/textwidget.lua:75: in function 'getSize'
	frontend/ui/widget/container/centercontainer.lua:10: in function 'paintTo'
	frontend/ui/widget/container/framecontainer.lua:85: in function 'paintTo'
	frontend/ui/widget/container/inputcontainer.lua:85: in function 'paintTo'
	frontend/ui/widget/horizontalgroup.lua:36: in function 'paintTo'
	frontend/ui/widget/verticalgroup.lua:37: in function 'paintTo'
	frontend/ui/widget/container/centercontainer.lua:24: in function 'paintTo'
	frontend/ui/widget/container/framecontainer.lua:85: in function 'paintTo'
	frontend/ui/widget/container/bottomcontainer.lua:18: in function 'paintTo'
	frontend/ui/widget/container/inputcontainer.lua:85: in function 'paintTo'
	frontend/ui/uimanager.lua:933: in function '_repaint'
	frontend/ui/uimanager.lua:1049: in function 'handleInput'
	frontend/ui/uimanager.lua:1121: in function 'run'
	[C]: at 0x00404f60
```

Frenzie:
> In the second and third row, I replaced every character in the row with "".
Which approach do you recommend?

I'd use an empty string because then you know which is which at a glance. It's not (currently) possible to just leave it empty? I mean like:

```lua
value, , , another value
```

Anyway, it's looking pretty good to me on the code front. 👍

WaseemAlkurdi:
@Frenzie 

No. Trying that would cause this error:
```
./luajit: error loading module 'ui/data/keyboardlayouts/ar_AA_keyboard' from file 'frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua':
	frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua:126: unexpected symbol near ','
stack traceback:
	[C]: at 0x0044e5d0
	[C]: in function 'require'
	frontend/ui/widget/virtualkeyboard.lua:542: in function 'init'
	frontend/ui/widget/inputtext.lua:380: in function 'initKeyboard'
	frontend/ui/widget/inputtext.lua:224: in function 'init'
	frontend/ui/widget/inputdialog.lua:296: in function 'init'
	frontend/apps/reader/modules/readerdictionary.lua:606: in function 'onShowDictionaryLookup'
	frontend/apps/reader/modules/readerdictionary.lua:178: in function 'callback'
	frontend/ui/widget/touchmenu.lua:731: in function 'action'
	frontend/ui/uimanager.lua:752: in function '_checkTasks'
	frontend/ui/uimanager.lua:1035: in function 'handleInput'
	frontend/ui/uimanager.lua:1121: in function 'run'
	[C]: at 0x00404f60
```

Frenzie:
> but for some reason, the word diacritics has to be repeated eight times. Is that normal?

If each layout needs the diacritics, that's normal. Naively I wouldn't expect them to be required on the symbols and numbers though.

> What is the correct approach to removing the Latin and Shift keyboards? Removing the whole column doesn't work (causes the characters in column 2 to appear in the column 1 keyboard).

I doubt removing it was a consideration while the keyboard was written, but filling it up with "" or nil is fine for now.

> No. Trying that would cause this error:

My bad; `nil` is functionally the same thing except that'd be valid syntax. Anyway, chances are the keyboard wouldn't like it regardless but at least Lua would.

WaseemAlkurdi:
> If each layer needs the diacritics, that's normal. Naively I wouldn't expect them to be required on the symbols and numbers though.

No, you're right. The only layer that requires diacritics is the main character layer. 
But having only one repeat of the word `diacritics` doesn't work (the popup doesn't open). That's why I had to repeat it eight times.

WaseemAlkurdi:
> I doubt removing it was a consideration while the keyboard was written, but filling it up with "" or nil is fine for now.

I think that I'll use `""` as you suggested, for readability.

Frenzie:
> But having only one repeat of the word diacritics doesn't work (the popup doesn't open). That's why I had to repeat it eight times.

That doesn't sound right, for the moment can you add a little line above it like:

```
--- @fixme Diacritics is only required on the main character layer.
```

NB The three `---` is on purpose. It's a special tag.

WaseemAlkurdi:
@Frenzie

I'll do that.
I'm curious, however, as to how @yparitcher would do it in the Hebrew keyboard.

yparitcher:
> @Frenzie
> 
> I'll do that.
> I'm curious, however, as to how @yparitcher would do it in the Hebrew keyboard.

there are many Diacritics so i used the <kbd>Shift</kbd> layer for Diacritics.

WaseemAlkurdi:
@yparitcher More than nine diacritics? Won't a popup work in your case?
I used both the Symbol layer (like on mobile keyboards, but after some thought, I really should have used Shift like you did, as on PC keyboards) and a separate key.

WaseemAlkurdi:
@Frenzie 
That's one issue down. This leaves us with the Lua warnings and the removal of Shift and Latin. After that, it will be ready for merge.

yparitcher:
there are like 20 cantillation marks and 5 - 10 nikkud, the symbol layer i left with numbers etc.

you might want to wait for #5583 to settle before merging this.

Frenzie:
@WaseemAlkurdi #5583 is merged. What you'll want to do is something like:

```
# only if you haven't yet
git remote add upstream git@github.com:koreader/koreader.git
git pull upstream master
git submodule update --recursive
```

Once that's done, you'll need to do something like this: https://github.com/koreader/koreader/pull/5610/commits/50d5350376a6904da1dc30d92efc75695c14d541

Specifically note the new lines at the top:

```lua
    min_layer = 1,
    max_layer = 8,
```

Also see the updated Hebrew keyboard.

If you prefer I could do some of these things for you.

Frenzie:
@WaseemAlkurdi I took the liberty of making it run, so something like this should do the trick:
```sh
git pull
git submodule update --recursive
```

I tried to simplify the layout, but I'm not sure if I completely understood it. In case I messed up completely and you need to start from scratch on that, you can use:

```sh
git revert 470526759134e8955938c73df0080c123e3bc29f
```

I hope it's a good kick-off point, however. ;-)

yparitcher:
@frenzie 

layout #4 is supposed to be numbers / symbols, it is the default layout for nuber related text fields like the `go to page` dialog.

yparitcher:
@frenzie

layout #4 is supposed to be numbers / symbols, it is the default layout =
for nuber related text fields like the `go to page` dialog.

Frenzie:
Makes sense, I'll replace the QWERTY stuff with the original column 4.

WaseemAlkurdi:
@Frenzie I did the steps yet the file seems to have vanished

Frenzie:
Are you sure you're in your `arabic_keyboard` branch?

You can see your current branch with:
```
git branch
```

You can switch branches with:

```
git checkout insert-branch-name-here
```

WaseemAlkurdi:
@Frenzie Ah, can't believe I missed that!
I've moved to a new laptop, so I cloned my own repo again, then added upstream and pulled again. Then, it seems to have switched to `master`, and my branch didn't appear under `git branch`. Out of desperation, I tried `git checkout arabic_keyboard`, and lo and behold:
```
Branch 'arabic_keyboard' set up to track remote branch 'arabic_keyboard' from 'origin'.
Switched to a new branch 'arabic_keyboard'
```
It worked, and my branch now shows. Stupid git, this `git` (pun _very_ intended :-D )

WaseemAlkurdi:
I've updated the keyboard to be more PC-like instead of Android-like. I used the Shift "panel" for diacritics, in addition to the existing diacritics key (which would make it easier for both PC users where the diacritics are accessed using Shift, and mobile users, where the diacritics are on a key, usually the comma key or a dedicated "Diacritics" key on iOS).
Only issues left:
- Do you propose merging the `حركات` (diacritics) key with the comma key, like on Android's AOSP/Google Keyboard, or keep it as is, on a separate key like on iOS?
On the one hand, merging with the Comma key makes the keyboard more generic for non-Arabic languages that use Arabic-like scripts, because the Arabic word isn't used.
But on the other hand, some people wouldn't know what's "below" the comma key until they stumble upon it by accident (Android's keyboard shows a small `...` to indicate that there are keys below a given key).
- There is an extra unused panel (the Shift-Sym panel). This currently holds some Hebrew keys (given that I based my keyboard on the Hebrew keyboard), but I don't know what should go in there. I thought of placing non-Arabic keys in there, but they are already there, accessible by holding the Arabic character that looks the closest to that non-Arabic letter (exactly like on Android).
- The letter "haa" (`هـ`) on the keyboard currently looks like this: `ه`. 
In handwriting, when this letter is alone, not attached to any other word, it's supposed to be `هـ`, but to reproduce this on a computer, you have to add a `ـ` (Shift-J). However, you wouldn't want this character in regular use, as this is considered by the computer to be a character, breaking things like "find in text".
TL;DR: How do I make the keyboard _show_ one form of a letter while actually entering another form?
If it isn't possible, I'll just leave it as is (which is the way it is on mobile keyboards, unfortunately, but is actually linguistically incorrect).

poire-z:
> The letter "haa" (هـ) on the keyboard currently looks like this: ه.
In handwriting, when this letter is alone, not attached to any other word, it's supposed to be هـ, but to reproduce this on a computer, you have to add a ـ (Shift-J). 

Not sure if you mean
(a) there is a Harfbuzz/xtext use/rendering bug, and that a single arabic char that should be rendered as its standalone form is actually not by our code.
(b) Of if everything is correct, and you just need a trick to make it rendered in its not-standalone form on the keyboard key.

If (a), can you check how it works in some HTML file rendered by crengine - and make me a small sample .html that I can use with Firefox or Chrome and compare.

If (b), I think I read there are some tricks by using invisible unicode chars with arabic like [Zero-width joiner](https://en.wikipedia.org/wiki/Zero-width_joiner) and [Zero-width non-joiner](https://en.wikipedia.org/wiki/Zero-width_non-joiner), may be see `https://github.com/w3c/i18n-discuss/issues/2` - dunno if that might help.

> while actually entering another form?

It will enter the unicode letter without specifying the form - you shouldn't decide that: it's harfbuzz, when rendering, depending on neighbour chars, that will decide what form to use :)

Frenzie:
> Do you propose merging the حركات (diacritics) key with the comma key, like on Android's AOSP/Google Keyboard, or keep it as is, on a separate key like on iOS?

Why does Android merge them? Lack of space?

> There is an extra unused panel (the Shift-Sym panel). This currently holds some Hebrew keys (given that I based my keyboard on the Hebrew keyboard), but I don't know what should go in there.

Just disable it?

WaseemAlkurdi:
@poire-z ,

> Not sure if you mean
(a) there is a Harfbuzz/xtext use/rendering bug, and that a single arabic char that should be rendered as its standalone form is actually not by our code.

No, not that. It's definitely not in KOReader, nor it's a bug in HarfBuzz. It's sort of a "design decision" that was taken by whomever had written the first Arabic on a computer. The decision is technically very sound, but not linguistically so (nobody wrote the letter `هـ` as `ه` before).

> (b) Of if everything is correct, and you just need a trick to make it rendered in its not-standalone form on the keyboard key.

Precisely! It's as if you read through my mind!

> If (b), I think I read there are some tricks by using invisible unicode chars with arabic like Zero-width joiner and Zero-width non-joiner, may be see https://github.com/w3c/i18n-discuss/issues/2 - dunno if that might help.

Tried the ZWJ and ZWNJ. The ZWJ is perfect ... except that it still breaks stuff like search in text, because it's still a character after all.
(To reproduce: copy the following to any text editor:
```
ه‍ل
هل
```
The first word has a ZWJ, the second one doesn't. Now do a search in text for the _second_ instance. The first instance won't be highlighted/found.)

The ZWNJ has the opposite effect, it "forces" characters to remain "separate". Not useful.
So, basically, what is needed is either:
- some way of having the "keycap" show one character and type another (I tried with `label = `, like on the Enter key, but that didn't work).
- or something like the ZWJ but one that doesn't affect searches in files.

WaseemAlkurdi:
@Frenzie 
> I don't know, frankly. But it seems to be that way.
Note that the Arabic iOS keyboard doesn't have a comma nor a dot on the last row (unlike Android/AOSP, which has both).
iOS keyboard: http://aratools.com/static/images/keyboard-hamza-alif.png
Most popular Android keyboards (incl. AOSP): http://laviefrugale.blogspot.com/2013/08/typing-arabic-vowelsdiacritics-on.html

poire-z:
Dunno if you should remove Shift+Sym. On the english or french keyboards, Sym and Shift+Sym give:
<kbd>image</kbd>
<kbd>image</kbd>

Some of these like % $ € are not on your keyboard - and might be useful to arabic writers. So I guess you could add a bunch of these (and duplicate some that are on the other layers to fill it).

WaseemAlkurdi:
@poire-z 

I've fixed the Shift+Sym panel by copying over symbols from the English keyboard.
However, I'm somewhat at a loss with the `label =` and `key = ` part. It doesn't seem to work, because the keycap has different labels in different keyboard panels.

poire-z:
OK, I see. It just needs some added tweak to virtualkeyboard.lua:
```diff
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -786,4 +786,6 @@ function VirtualKeyboard:addKeys()
             local key_chars = self.KEYS[i][j][self.keyboard_layer]
+            local label
             if type(key_chars) == "table" then
                 key = key_chars[1]
+                label = key_chars.label
             else
@@ -796,3 +798,3 @@ function VirtualKeyboard:addKeys()
             local key_height = base_key_height
-            local label = self.KEYS[i][j].label or key
+            label = label or self.KEYS[i][j].label or key
             local virtual_key = VirtualKey:new{
```
and you could use something like:
```
-            { _Z_,    _z_,    "&",    "7",    "Á",    "á",    "¿",    "¿", },
+            { _Z_,    _z_,    "&",    {"88", label="77"},    "Á",    "á",    "¿",    "¿", },
```
to have a key in some specific layer labeled 77 but output 88.

Frenzie:
That'll be fine. I basically just wrote what I needed. :-)

WaseemAlkurdi:
@poire-z This works when specifying the character as a constant `"a"`, but not as a variable (what Lua calls a table for some reason).
When specifying a variable, KOReader crashes:
```
/luajit: frontend/util.lua:338: bad argument #1 to 'gmatch' (string expected, got table)
stack traceback:
	[C]: in function 'gmatch'
	frontend/util.lua:338: in function 'splitToChars'
	frontend/ui/widget/inputtext.lua:443: in function 'addChars'
	frontend/ui/widget/virtualkeyboard.lua:838: in function 'addChar'
	frontend/ui/widget/virtualkeyboard.lua:121: in function 'callback'
	frontend/ui/widget/virtualkeyboard.lua:289: in function 'handleEvent'
	frontend/ui/widget/container/inputcontainer.lua:259: in function 'handleEvent'
	...
	frontend/ui/uimanager.lua:674: in function 'sendEvent'
	frontend/ui/uimanager.lua:1014: in function 'handleInputEvent'
	frontend/ui/uimanager.lua:1075: in function 'handleInput'
	[C]: at 0x00013ef9
lipc-wait-event exited normally with status: 0
```
The code in `util.lua` doesn't seem complicated but I don't really understand it.

poire-z:
> This works when specifying the character as a constant "a", but not as a variable (what Lua calls a table for some reason).

Please show the key (even the line) with exactly how you want to write it.
(No need to fix util.lua: it expects a string, it got a table - you have to make it have a string :)

If you mean you want `{key="88", label="77"}` instead of `{"88", label="77"}`, this might work (to add to the the first chunk of my previous patch above):
```diff
-                 key = key_chars[1]
+                 key = key_chars.key or key_chars[1]
+                 label = key_chars.label
```

Frenzie:
The only thing needed to make this PR ready to merge is that little change suggested by @poire-z in https://github.com/koreader/koreader/pull/5569#issuecomment-559274204?

WaseemAlkurdi:
@Frenzie No, not yet ... this did not work. It still crashes with the error described above.

poire-z:
Even with [my above suggestion](https://github.com/koreader/koreader/pull/5569#issuecomment-559595610) ?

WaseemAlkurdi:
@poire-z Yes, even with that:
image

poire-z:
OK, but can you point at which key (the line in ar_AA_keyboard.lua) causes the problem, so we also see how it's defined? Or is it all keys?
And if it's not in what's been commited so far, paste the line here, so we know on what to find a solution :)

WaseemAlkurdi:
@poire-z  I `git push`ed my latest tree (dunno why GitHub says I'm requesting a review). Anyway, this is the line:
`69           { "÷",                       {key=h_aa, label="ه‍"},    "|",    "ـ", },`

poire-z:
Your latest commit brings many permission changes to 578 files :) - and probably some of them were last edited by these guys, so the review, I guess.

You might want to remove that last commit:
Keep a backup of your changed files to not lose them.
`git reset HEAD~1`

Then fix your permissions if you can :) `git status` would show what's about to be commited.
Reapply your changes, and just use `git add path/to/specific/file/to/add` instead of no arg or a wildcard.

(Later, you'll need to rebase to master because there is a conflit where you added `ar_AA = "ar_AA_keyboard",` as another keyboard has been added in the meantime I think. But don't do that yet.)

poire-z:
> `69 { "÷", {key=h_aa, label="ه‍"}, "|", "ـ", },`

Ok, your `h_aa` is already a table, so you need to let it as is:

`69 { "÷", h_aa, "|", "ـ", },`

And set the label in that table, in ar_AA_popup.lua :
```lua
    h_aa = { -- الهاء
        "ه",   -- that's key_chars[1]
        label="ه‍", -- that's key_chars.label
        north = "هـ",
    },
```
And just use my patch from https://github.com/koreader/koreader/pull/5569#issuecomment-559274204 - ignoring the 2nd one https://github.com/koreader/koreader/pull/5569#issuecomment-559595610.

Frenzie:
Yes, it's because you touched every single file in the repo. I think it's probably easier to just do `git revert da36e418bb8f596e8e79ae11e8a48ae01452b767` rather than worrying too much about Git intricacies for the moment.

If you're interested, the review thing is automatically triggered by the codeowners file I added in #5559. I'm not really sure how useful it is, but tbh I never expected it to be triggered quite like this. ;-)

poire-z:
Yes indeed, go with that `git revert`, way simpler!

WaseemAlkurdi:
@Frenzie 
> Yes, it's because you touched every single file in the repo.

I've recently changed distros (but moved my `koreader` directory as is, possibly across filesystem boundaries). It's probably because of that, because the revert produces a huge amount of `mode change`s. 

@poire-z 
> Ok, your h_aa is already a table, so you need to let it as is:

Ah, shoot. Missed that entirely.
Your edit is pretty much it. I should test that, and by tomorrow, it should hopefully be ready to roll!

poire-z:
@WaseemAlkurdi : any update on this (and your arabic and kurdish translations)?

WaseemAlkurdi:
> @WaseemAlkurdi : any update on this (and your arabic and kurdish translations)?

I've been _very_ busy. Everything holding me up (in real life) should be resolved by 15/1.

WaseemAlkurdi:
Something else: my tree is _completely_ messed up at this point.
I cloned my own repo again: `git clone https://github.com/WaseemAlkurdi/koreader`
which brings me back to a tree with commit `da36e418` applied.
Trying to revert doesn't work.
```
$ git revert da36e418
error: Reverting is not possible because you have unmerged files.
hint: Fix them up in the work tree, and then use 'git add/rm <file>'
hint: as appropriate to mark resolution and make a commit.
fatal: revert failed
```
`git rm`-ing the three files that I modified, then retrying the revert would get me this:
```
$ git revert da36e418
error: your local changes would be overwritten by revert.
hint: commit your changes or stash them to proceed.
fatal: revert failed
```

Frenzie:
@WaseemAlkurdi Seems to work for me. Are you in the right branch? ;-)

Frenzie:
Anyhow, I can push that to your branch if you like.

poire-z:
Looks like you have modified files not commited. I would do:
```
git diff > backup_of_my_non_commited_stuff_just_in_case.diff
git reset --hard
git revert da36e418
```

WaseemAlkurdi:
@poire-z @Frenzie Thanks! I was on the wrong branch indeed. Though it didn't clone anything except `master`, so I had to clone again, passing the option `-b arabic_keyboard`.
It's fine now, and the commit has been reverted, but I notice that this repo (`koreader/koreader`) is _not_ included in the project's list of remotes/origins, even though GitHub shows it as having been forked from this repo. Should I worry about this?

Frenzie:
No, only wherever you clone from will normally be included by default. If you wish to add other remotes such as someone else's GH fork or the GitLab mirror you'll need to do so manually.

For every repo I contribute to regularly, I've added a remote called upstream.
```
git remote add upstream https://github.com/koreader/koreader.git
```
```
git remote add upstream git@github.com:koreader/koreader.git
```
That enables me to `git fetch upstream master` to grab the latest commits, followed by a manual cherry pick, merge or rebase. Or one of the various all in one alternative commands like `git pull upstream master` or `git pull -r upstream master`. I've tried to succinctly explain that [here](http://koreader.rocks/doc/topics/Collaborating_with_Git.md.html). But none of that should be very relevant to you right now, unless you want it to be.

WaseemAlkurdi:
> No, only wherever you clone from will normally be included by default. If you wish to add other remotes such as someone else's GH fork or the GitLab mirror you'll need to do so manually.
> 
> For every repo I contribute to regularly, I've added a remote called upstream.
> 
> ```
> git remote add upstream https://github.com/koreader/koreader.git
> ```
> 
> ```
> git remote add upstream git@github.com:koreader/koreader.git
> ```
> 
> That enables me to `git fetch upstream master` to grab the latest commits, followed by a manual cherry pick, merge or rebase. Or one of the various all in one alternative commands like `git pull upstream master` or `git pull -r upstream master`. I've tried to succinctly explain that [here](http://koreader.rocks/doc/topics/Collaborating_with_Git.md.html). But none of that should be very relevant to you right now, unless you want it to be.

I know about this (and I did this back in the first clone as to update my tree with changes from this repo).
But I thought that `git` somehow does this automatically afterwards (a one-time change per repo or something)
Liked your write-up! Pretty much sums it up.

WaseemAlkurdi:
@poire-z 
I've pushed a final commit. The keyboard now works absolutely great - ready to merge.
Any thoughts?

Frenzie:
@WaseemAlkurdi Looks like the conflict with master is really minor, you can resolve it in the GH web GUI or I can quickly do it too if that's easier.

WaseemAlkurdi:
@Frenzie I'd do it, but how?

Frenzie:
There should be a "resolve conflicts" button above the comment box. That will then show you that the edit for the Arabic keyboard and the Romanian keyboard conflict. Of course we want both. ;-)

WaseemAlkurdi:
@Frenzie Done! I've kept both.

frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua
```diff
@@ -0,0 +1,129 @@
...
+    utf8mode_keys = {["🌐"] = true},
+    umlautmode_keys = {["Äéß"] = false},
+    keys = {
+        -- first row
+        {  --  1                          2        3       4
+            { diacritic_fat_ha,          dhad,    "„",   "0", },
+            { diacritic_tanween_fath,    saad,    "!",   "1", },
+            { diacritic_damma,           thaa,    _at,   "2", },
+            { diacritic_tanween_damm,     qaf,     "#",    "3", },
+            { "ﻹ",     fah,     "+",    _eq, },
```
poire-z: Might keep them all aligned the same way with the `1         2   3  4`, for readability :)

frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua
```diff
@@ -0,0 +1,129 @@
...
+            { "إ",   ghayn,    "€",    "(", },
+            { "`",                       ayin,    "‰",    ")", },
+-- h_aa
```
poire-z: You can indent that comment (including the `--`) up to where the `{"ه", label="ه‍"}` (I assume it says what's that thing is :)

Frenzie:
@WaseemAlkurdi Can these just be removed?

```
    frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua:11:7: unused variable 'h_aa'
    frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua:38:7: unused variable 'taamarbouta'
```

frontend/ui/data/keyboardlayouts/keypopup/ar_AA_popup.lua
```diff
@@ -0,0 +1,162 @@
...
+        north = "هـ",
+    },
+    waw = {
+        "و",
+    },
+    zay = {
+        "ز",
+        north = "ژ",
+    },
+    ha = { -- الحاء
```
poire-z: You could translate that arabic word to english      `-- الحاء (english translation)` if that means something other than "ha" :) so we don't have to wonder if we're missing something, and to improve the culture of non arabic readers :)
(I wonder the same above for h_aa, but it's shorter, so it would just be h_aa in arabic :)

frontend/ui/data/keyboardlayouts/keypopup/ar_AA_popup.lua
```diff
@@ -0,0 +1,162 @@
...
+          "ً",
+    },
+   diacritic_tanween_damm = {
+          "ٌ",
+    },
+   diacritic_tanween_kasr = {
+          "ٍ",
+    },
+-- الحركات
+   diacritics = {
```
poire-z: I guess that arabic word means "diacritics" :)

frontend/ui/data/keyboardlayouts/keypopup/ar_AA_popup.lua
WaseemAlkurdi: @poire-z I didn't know how to "transcribe" the two letters, so I used 'ha' for هـ and 'h_aa' for ح. Honestly, this could be improved, and I myself mix up the two "transciptions". 
Please do listen to the first 5 seconds of this YouTube video: https://www.youtube.com/watch?v=YeyVjaWD6-U. The first letter that the announcer pronounces is 'هـ' (my 'ha') and the second letter is 'ح' (my h_aa).

WaseemAlkurdi: The word 'الحاء' is simply the name of the letter, roughly meaning "the letter h_aa". like how "double u" is sort of a name of the letter W in English.

frontend/ui/data/keyboardlayouts/keypopup/ar_AA_popup.lua
WaseemAlkurdi: @poire-z Precisely.

frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua
WaseemAlkurdi: @poire-z Can I do that from the web interface, or do I have to push again?

frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua
WaseemAlkurdi: @poire-z Fixing that!

frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua
WaseemAlkurdi: @poire-z Turns out that I can, but it would be considered another commit. Not a problem!

WaseemAlkurdi:
> @WaseemAlkurdi Can these just be removed?
> 
> ```
>     frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua:11:7: unused variable 'h_aa'
>     frontend/ui/data/keyboardlayouts/ar_AA_keyboard.lua:38:7: unused variable 'taamarbouta'
> ```

There's a slight problem here. I had to hardcode these two letters (instead of defining them as variables) because I had to use `label=`, as you can see in `ar_AA_keyboard.lua`). 
Do you suggest removing these variables, or is there a way in which I could use `label=` with `ar_popup.lua`?

poire-z:
Regarding the trick with the zero-width-joiner/non-joiner stuff mentionned in https://github.com/koreader/koreader/pull/5569#issuecomment-559017799 :
The sad thing is that we don't see it in the code :)
I had the same issue some weeks ago with zero width space, and I used \x utf8 bytes instead. You could do that, with a comment explaining why.
Or, I think arabic has another range in unicode for "arabic presentational forms", which I think are the pre-shaped glyphs, so may be the one you want might be there.

WaseemAlkurdi:
@poire-z Sheesh, I've mistyped it while moving. It was supposed to have a ZWJ, but the ZWJ isn't directly accessible on the keyboard, unlike a non-zero width joiner (an easy Shift-J away). Will fix this instant.

WaseemAlkurdi:
>  The sad thing is that we don't see it in the code :)

GitHub's web UI, at least here on Firefox on Linux, uses a red dot to indicate it instead of the shaped character.

WaseemAlkurdi:
> Or, I think arabic has another range in unicode for "arabic presentational forms", which I think are the pre-shaped glyphs, so may be the one you want might be there.

No, these (from what I see here: https://en.wikipedia.org/wiki/Arabic_Presentation_Forms-A) are the Arabic characters used by non-Arabic languages with Arabic scripts, which are not "legal Arabic". Something like how some French and German characters, while using Latin script, are not "legal Latin".

WaseemAlkurdi:
Replaced as per your suggestion. It's a much neater solution, having the whole thing as one character, instead of bothering with invisible characters.
Looks like this is done then! The build tests check out.

A small note though. The Wikipedia article (https://en.wikipedia.org/wiki/Arabic_Presentation_Forms-B) says:

> The presentation forms are present only for compatibility with older standards, and are not currently needed for coding text.

Does this mean that these might be deprecated? If so, wouldn't this be a "hack" that might stop working with unpredictable consequences?

poire-z:
And in B ? :) Isn't it this one:


Uni­codecode point | char­acter | UTF-8en­co­ding(hex) | Uni­co­de char­ac­ter name | Uni­co­de 1.0 char­act­er name (de­pre­ca­ted)
-- | -- | -- | -- | --
U+FEE9 | ﻩ | ef bb a9 | ARABIC LET­TER HEH ISOLATED FORM | GLYPH FOR ISOLATE ARABIC HA
U+FEEA | ﻪ | ef bb aa | ARABIC LET­TER HEH FINAL FORM | GLYPH FOR FINAL ARABIC HA
U+FEEB | ﻫ | ef bb ab | ARABIC LET­TER HEH INITIAL FORM | GLYPH FOR INITIAL ARABIC HA
U+FEEC | ﻬ | ef bb ac | ARABIC LET­TER HEH MEDIAL FORM | GLYPH FOR MEDIAL ARABIC HA

from https://unicode-search.net/unicode-namesearch.pl?term=HEH

poire-z:
Oh, allright, you found it :)
I don't think Unicode can deprecate or remove codepoints :) by contract, I guess, they can just add stuff.

poire-z:
> Looks like this is done then! The build tests check out.

Stil the alignment with `1 2 3 4` to fix with GH UI :)

WaseemAlkurdi:
> > Looks like this is done then! The build tests check out.
> 
> Stil the alignment with `1 2 3 4` to fix with GH UI :)

Done! Should look really tidy now.

WaseemAlkurdi:
> I think you accidentally lost a space here.

What's a space between friends? :-D

frontend/ui/data/keyboardlayouts/keypopup/ar_AA_popup.lua
```diff
@@ -0,0 +1,162 @@
...
+          "ً",
+    },
+   diacritic_tanween_damm = {
+          "ٌ",
+    },
+   diacritic_tanween_kasr = {
+          "ٍ",
+    },
+--  Diacritics (al-Harakat الحركات)
+   diacritics = {
```
poire-z: Last nitpick: have the comment indented as what follows it.

WaseemAlkurdi: @poire-z Done.

NiLuJe:
I'd possible have gone with IPA for those latest (interesting!) comments, but that's just me ;).

Frenzie:
No, I definitely would've gone with IPA. But I didn't know IPA prior to the university, so I wouldn't really expect anybody else to unless they had a degree in something language-related.

Issue #5476, opened 2019-10-11
Segmentation fault while browsing LAN OPDS catalog


BrainDamage:
* KOReader version: v2019.09.3-64g5209494_2019-10-10
* Device: Kobo Aura H2O2

#### Issue
Segmentation fault while browsing LAN OPDS catalog

#### Steps to reproduce
Just click on the OPDS entry after adding it

##### `crash.log` (if applicable)

>  [*] Current time: 10/11/19-16:18:34
>  [*] Version: v2019.09.3-64-g5209493_2019-10-10
> 
> ffi.load: SDL2
> ffi.load (warning): libSDL2.so: cannot open shared object file: No such file or directory
> ffi.load: libSDL2-2.0.so
> ffi.load (warning): libSDL2-2.0.so: cannot open shared object file: No such file or directory
> ffi.load: libSDL2-2.0.so.0
> ffi.load (warning): libSDL2-2.0.so.0: cannot open shared object file: No such file or directory
> SDL2 not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: libSDL2-2.0.so.0
> ffi.load: blitbuffer
> FB: computed size: 3342336
> FB: id           : mxc_epdc_fb
> FB: smem_len     : 6684672
> FB: line_length  : 1088
> FB: xres         : 1080
> FB: xres_virtual : 1088
> FB: yres         : 1440
> FB: yres_virtual : 3072
> FB: bpp          : 8
> 10/11/19-16:18:34 INFO  initializing for device Kobo_snow
> 10/11/19-16:18:34 INFO  framebuffer resolution: {
>     ["h"] = 1440,
>     ["w"] = 1080
> }
> [ko-input] Forked off fake event generator(pid:2817).
> ffi.load: libs/libmupdf.so
> ffi.load: libs/libwrap-mupdf.so
> ffi.load: libs/libfreetype.so.6
> ffi.load: sqlite3
> ffi.load: z
> ffi.load: sqlite3
> 10/11/19-16:18:35 INFO  Plugin  plugins/autofrontlight.koplugin/main.lua  has been disabled.
> 10/11/19-16:18:35 INFO  Plugin  plugins/hello.koplugin/main.lua  has been disabled.
> 10/11/19-16:18:35 INFO  RD loaded plugin SSH at plugins/SSH.koplugin
> 10/11/19-16:18:35 INFO  RD loaded plugin autosuspend at plugins/autosuspend.koplugin
[...]

PR #5572, opened 2019-11-06
Prevent crash on local calibre OPDS server

robert00s:
Close: #5476 
- Prevent segmentation fault when we enter incorrect address of calibre opds server (eg http://192.168.1.1:8080/ instead of proper http://192.168.1.1:8080/opds
- prevent unnecessary infos in Gutenberg OPDS like

### DIFF
--- a/frontend/ui/opdsparser.lua
+++ b/frontend/ui/opdsparser.lua
@@ -84,7 +84,7 @@ function OPDSParser:parse(text)
         return s:gsub( "%p", {["&"] = "&amp;", ["<"] = "&lt;", [">"] = "&gt;" } )
     end )
     local xlex = luxl.new(text, #text)
-    return self:createFlatXTable(xlex)
+    return assert(self:createFlatXTable(xlex))
 end
 
 return OPDSParser
--- a/frontend/ui/widget/opdsbrowser.lua
+++ b/frontend/ui/widget/opdsbrowser.lua
@@ -298,6 +298,9 @@ function OPDSBrowser:fetchFeed(item_url, username, password, method)
         if xml ~= "" then
             return xml
         end
+    elseif method == "HEAD" then
+        -- Don't show error messages when we check headers only.
+        return
     elseif code == 301 then
         UIManager:show(InfoMessage:new{
             text = T(_("The catalog has been permanently moved. Please update catalog URL to '%1'."), headers['Location']),
@@ -401,6 +404,11 @@ function OPDSBrowser:genItemTableFromCatalog(catalog, item_url, username, passwo
     end
 
     if not feed.entry then
+        if #hrefs == 0 then
+            UIManager:show(InfoMessage:new{
+                text = _("Catalog not found."),
+            })
+        end
         return item_table
     end
 


frontend/ui/opdsparser.lua
```diff
@@ -84,7 +84,7 @@ function OPDSParser:parse(text)
         return s:gsub( "%p", {["&"] = "&amp;", ["<"] = "&lt;", [">"] = "&gt;" } )
     end )
     local xlex = luxl.new(text, #text)
-    return self:createFlatXTable(xlex)
+    return assert(self:createFlatXTable(xlex))
```
Frenzie: Canned response, but is this assert protected by a pcall? ;-)

robert00s: Do you suggest to use 
`assert(pcall(createFleXTable, xlex))` ?

Frenzie: I suggest whatever doesn't crash when the assertion fails, which would be the other way around (`pcall(assert(`). ;-) Note that we also have a debug.assert function (might be named slightly differently) which will crash the program in debug mode.

I'm not sufficiently familiar with the code to know what makes the most sense here specifically. It could also be a level higher up, like: `pcall(OPDSParser.parse, text)`.

robert00s: > I suggest whatever doesn't crash when the assertion fails, which would be the other way around (pcall(assert(). ;-) Note that we also have a debug.assert function (might be named slightly differently) which will crash the program in debug mode.


Example please. I really don't understand :)

>I'm not sufficiently familiar with the code to know what makes the most sense here specifically. It could also be a level higher up, like: pcall(OPDSParser.parse, text).


pcall(OPDSParser.parse, text) won't work. It can't enter to any opds server.

Frenzie: I mean `dbg.dassert()`. But that's not really appropriate here.

https://github.com/koreader/koreader/blob/0885e9947dead0fe4b30d816ffeafca4a3932a19/frontend/dbg.lua#L50-L53

> pcall(OPDSParser.parse, text) won't work. It can't enter to any opds server.

How do you mean? Something like `pcall(OPDSParser.parse, OPDSParser, text)` or `pcall(OPDSParser.parse, self, text)`, whatever works in the relevant context.

My point is simply that `assert()` means the program halts unless it's taken care of in some way.

```
> function bla() assert(false) end
> bla()
stdin:1: assertion failed!
stack traceback:
	[C]: in function 'assert'
	stdin:1: in function 'bla'
	stdin:1: in main chunk
	[C]: in ?
```

poire-z: I was going to ask : but why add that `assert()` ?! Why would youd need that?
But following the code, the only place this is called is actually wrapped in a pcall():
https://github.com/koreader/koreader/blob/0885e9947dead0fe4b30d816ffeafca4a3932a19/frontend/ui/widget/opdsbrowser.lua#L324-L363
So, you actually added it to make that pcall() fail.
So, case closed :)

Frenzie: Excellent, so the answer to my initial question is a straightforward "yes". :+1:

PR #5573, opened 2019-11-06
Keyboard: avoid additional key stroke on hold release

### LINKED ISSUE #5555: Keyboard popup propagates

poire-z:
Added some forgotten `return true` on hold release event handlers: the event would otherwise trigger
a tap (as we generate a tap normally only at release time) and cause duplicate key strokes.
Details in https://github.com/koreader/koreader/issues/5555#issuecomment-550423347. Closes #5555.

(I included your suggested fix for the keyboard popup, as it's more related to my other fixes. You might want to remove it from #5569).

### DIFF
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -58,7 +58,10 @@ function VirtualKey:init()
         self.skiptap = true
     elseif self.label == "Backspace" then
         self.callback = function () self.keyboard:delChar() end
-        self.hold_callback = function () self.keyboard:delToStartOfLine() end
+        self.hold_callback = function ()
+            self.ignore_key_release = true -- don't have delChar called on release
+            self.keyboard:delToStartOfLine()
+        end
         --self.skiphold = true
     elseif self.label =="←" then
         self.callback = function() self.keyboard:leftChar() end
@@ -241,21 +244,31 @@ function VirtualKey:onSwipeKey(arg, ges)
 end
 
 function VirtualKey:onHoldReleaseKey()
+    if self.ignore_key_release then
+        self.ignore_key_release = nil
+        return true
+    end
     Device:performHapticFeedback("LONG_PRESS")
     if self.keyboard.ignore_first_hold_release then
         self.keyboard.ignore_first_hold_release = false
         return true
     end
     self:onTapSelect()
+    return true
 end
 
 function VirtualKey:onPanReleaseKey()
+    if self.ignore_key_release then
+        self.ignore_key_release = nil
+        return true
+    end
     Device:performHapticFeedback("LONG_PRESS")
     if self.keyboard.ignore_first_hold_release then
         self.keyboard.ignore_first_hold_release = false
         return true
     end
     self:onTapSelect()
+    return true
 end
 
 function VirtualKey:invert(invert, hold)
@@ -377,6 +390,7 @@ function VirtualKeyPopup:init()
                 virtual_key.onHoldReleaseKey = function()
                     virtual_key:onTapSelect(true)
                     UIManager:close(self)
+                    return true
                 end
                 virtual_key.onPanReleaseKey = virtual_key.onHoldReleaseKey
 


Frenzie:
The behavior is still incorrect though.

1. Hold.
2. Move over to another key (outside the popup).
3. Release.
4. Now it's pressed!

If the modal behavior were as it should be (i.e., as it was prior to #5452) then there would be nothing to "fix."

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -241,21 +244,31 @@ function VirtualKey:onSwipeKey(arg, ges)
 end
 
 function VirtualKey:onHoldReleaseKey()
+    if self.ignore_key_release then
```
Frenzie: This is @#$@# ugly though. ;-)

poire-z:
I don't understand much about that modal stuff - and hardly a bit much about the inner working of your key popups.
But the absence of these `return true` just looks like a bug - and adding them like the natural fix.

And having a quick look at that uimanager stuff, it look to me that #5452 is naturally correct. Before it, may be the events didn't reach the keyboard because something there in uimanager was (incorrectly) not sending them. But now they are, as I think they should.
For my backspace issue, which is the only one having a hold_callback (with your keypopups), it does some immediate action (delToStartOfLine). On hold release, like for all other keys, onHoldRelease is called, and is supposed to do the tap-like action associated with the key (delChar, or addChar). And it was doing that since #5452 (and I don't know why it didn't before :). Preventing it from doing that (so, "ignore key release" :) looks to me like the correct way to fix the issue (and I don't see an other way that with this ugly flag :)
Your key popup issue is a lot less understandable to me.

Having that fixed so far away in uimanager looks like far fetched side effect :)
But if that really make sense for you, I'm fine not merging this.

Frenzie:
You can think of a modal dialog as a fullscreen widget for most intents and purposes. You wouldn't expect to be able to interact with anything below the fullscreen widget, quite the contrary. However, modal dialogs are arguably overused, including in KOReader. Does the wifi dialog even need to be modal? (The answer is probably affirmative within our context, because it might get lost behind other widgets.)

> But the absence of these return true just looks like a bug - and adding them like the natural fix.

Not a bug exactly, at least not here, but more of a good practice to possibly prevent a few wasted CPU cycles.

> Preventing it from doing that (so, "ignore key release" :) looks to me like the correct way to fix the issue (and I don't see an other way that with this ugly flag :)

But how are you overriding it to remove the line on hold?

poire-z:
> Not a bug exactly, 

On my 2 unwanted delChar was solved by the return true. The other one by the self.ignore_key_release.

> But how are you overriding it to remove the line on hold?

The removal of the line is done by the hold_callback, which is done on the "hold" event by:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L129-L134
handled by:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L206-L214

A tap on a normal key is handled by:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L123-L128
which calls:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L187-L195
But this one is also called when hold on a normal key (so, a slow tap):
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L135-L140
handled by:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L243-L250

So, 2 way to handle tap: quick tap, and slow tap that appears like a hold.
We just want to not handle the hold_release when it's not a slow tap, but a real hold that has its action handled before the hold_release.
So, for the backspace, I'm not overriding anything.

Frenzie:
I simply didn't realize that hold to remove a line was built-in behavior.

Anyway, note that #5452 resulted in *three* removals as opposed to what looks like the intended two.

So this PR could still make sense if you only want one.

poire-z:
> as opposed to what looks like the intended two.

I intended only one: the removal to start of line, not the removal of the `\n` (by that unwanted delChar) :) as without #5452.

Regarding this modal stuff handling:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/uimanager.lua#L683-L707
We have modal:

https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L502-L503

https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/ui/widget/virtualkeyboard.lua#L270-L271

So, if your modal popup key don't consume the event, it looks correct with this code that some other modal we get it: the keyboard, which I guess will give it to some VirtualKey, so your bug.
So, may be, in that uimanager code, when a modal is met, whether it's handleEvent() has consumed it or not (return true or false), we should stop checking other `modal` (but still the `is_always_active`?).
Would that be the expected behaviour when having 2 modal, even when they are not overlapping?

Frenzie:
> Would that be the expected behaviour when having 2 modal, even when they are not overlapping?

No, absolutely not! That's why #5452 is dead wrong. A modal dialog should always be modal. The very idea of "some other modal" makes no sense.

Frenzie:
If the wifi dialog pops up first and then another modal dialog opens on top of it then the wifi dialog *should not be accessible*.

The problem is that the caller there called two modal dialogs in a row. Only modal dialogs themselves are ever supposed to call other modal dialogs.

poire-z:
OK.
In uimanager, a widget having `self.modal = true` gets itself put on top in the window_stack. There's no other check/use of this `modal` attribute.
So, if at some point we have may have 2 widgets with modal=true in the stack.
Now, some event is send and not handled by the 1st modal in the stack.
Should it be send to the other lower leve widgets (so, possibly send to another widget that had/has modal=true) - or meeting a widget with modal=true should just stop the propagation to lower level widgets?
From what you said, and logically, the propagation should be stopped. Our current code does not do that.

poire-z:
OK, with #5574 and this one, I don't notice any issue. So I hope we're fine on the keyboard side.

Frenzie:
Could you clarify that? The behavior described in https://github.com/koreader/koreader/pull/5573#issuecomment-550453338 and https://github.com/koreader/koreader/issues/5415 suggests that it behaves correctly.

poire-z:
Yes, with _your_ added `return true` in `virtual_key.onHoldReleaseKey`, the event is explicitely consumed and not propagated (even if it did not match the key, like tap outside it). So we're fine.

But if it is set to modal, and if we had forgotten this `return true`, it would be propagated to other widgets (modal or not). Shouldn't the stopping of the propagation be handled generically by UIManager because this widget is modal?

Frenzie:
But that return true isn't responsible for that behavior.

poire-z:
OK, forget about that keyboard and that return true. I have enough of it :)

But generically: sendEvent => modal widget handler not returning true => propagated to lower non-modal widgets. Correct behaviour or bug?

Frenzie:
If that happened, that would be a bug. (Assuming that `modal` isn't a misnomer for `always_on_top`.)

PR #5580, opened 2019-11-08
default setting option always enabled

yparitcher:
as per https://github.com/koreader/koreader/pull/5522#issuecomment-548006595 do not grey out the current default


Frenzie:
Works for me. ;-)

PR #5581, opened 2019-11-08
loading bar thicker

yparitcher:
the current thickness for the loading bar when opening a document is to thin to really see on my KT4,
i don't have screenshots as they only trigger once koreader finishes loading.

i would like to make it a little thicker, it still does not fully cover the top menu bar.

let me know if this is the wrong place to change the number, and if it has negative effects on other devices. (my emulator loads books too fast to get the bar)

### DIFF
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -1030,9 +1030,9 @@ function ReaderRolling:showEngineProgress(percent)
         -- Widget size and position: best to anchor it at top left,
         -- so it does not override the footer or a bookmark dogear
         local x = 0
-        local y = 0
+        local y = 3
         local w = Screen:getWidth() / 3
-        local h = Screen:scaleBySize(15)
+        local h = Screen:scaleBySize(7)
         if self.engine_progress_widget then
             self.engine_progress_widget:setPercentage(percent)
         else
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -1032,7 +1032,7 @@ function ReaderRolling:showEngineProgress(percent)
         local x = 0
         local y = 0
         local w = Screen:getWidth() / 3
-        local h = Screen:scaleBySize(5)
+        local h = Screen:scaleBySize(15)
         if self.engine_progress_widget then
             self.engine_progress_widget:setPercentage(percent)
         else
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -6,6 +6,7 @@ local InputContainer = require("ui/widget/container/inputcontainer")
 local MultiConfirmBox = require("ui/widget/multiconfirmbox")
 local ProgressWidget = require("ui/widget/progresswidget")
 local ReaderPanning = require("apps/reader/modules/readerpanning")
+local Size = require("ui/size")
 local TimeVal = require("ui/timeval")
 local UIManager = require("ui/uimanager")
 local bit = require("bit")
@@ -1032,7 +1033,7 @@ function ReaderRolling:showEngineProgress(percent)
         local x = 0
         local y = 3
         local w = Screen:getWidth() / 3
-        local h = Screen:scaleBySize(7)
+        local h = Size.line.progress
         if self.engine_progress_widget then
             self.engine_progress_widget:setPercentage(percent)
         else
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -1031,7 +1031,7 @@ function ReaderRolling:showEngineProgress(percent)
         -- Widget size and position: best to anchor it at top left,
         -- so it does not override the footer or a bookmark dogear
         local x = 0
-        local y = 3
+        local y = Size.margin.small
         local w = Screen:getWidth() / 3
         local h = Size.line.progress
         if self.engine_progress_widget then
--- a/frontend/ui/size.lua
+++ b/frontend/ui/size.lua
@@ -62,6 +62,7 @@ local Size = {
         thin = Screen:scaleBySize(0.5),
         medium = Screen:scaleBySize(1),
         thick = Screen:scaleBySize(1.5),
+        progress = Screen:scaleBySize(7),
     },
     item = {
         height_default = Screen:scaleBySize(30),


poire-z:
> i don't have screenshots as they only trigger once koreader finishes loading.

You could post a photo.

> i would like to make it a little thicker, it still does not fully cover the top menu bar.

I see how 5 could feel too thin. Tested on my device, and I find 15 quite big. 10 would be fine for me, 12 would be the max I would not find ugly. (Somehow, the taller it is, the less black it feels, so it looks like some grey/dirty mass on my GloHD - and at 15, it slightly overrides Kobo Start Menu (kobo launcher) first text line.)

Anyway, I wouldn't really mind it, I see it quite rarely :)

> let me know if this is the wrong place to change the number

I don't see any other place :) But you could define it as some module variable, like those a few lines above (would make it easier for me to patch it back to 5 in my own patch :)
`local ENGINE_PROGRESS_HEIGHT = Screen:scaleBySize(15)`

Frenzie:
> if it has negative effects on other devices

It should have roughly the same physical size on all devices. Personally I rather like that it's an elegant line rather than some distracting blob that just flashes for half a second.

> I don't see any other place :)

I thought it was intended to be more or less the same as any progressbar (like at the bottom), which however happens to use a value of 3 or 7 right now.

https://github.com/koreader/koreader/blob/61a8b63757accbfcb3cf077fc70f9033ba8df0ea/frontend/apps/reader/modules/readerfooter.lua#L64-L65

In regard to which I'll quote myself:

http://koreader.rocks/doc/modules/ui.size.html

> There are values for borders, margins, paddings, radii, and lines. Have a look at the code for full details. If you are considering to deviate from one of the defaults, please take a second to consider:
>
>1.    Why you want to differ in the first place. We consciously strive toward consistency in the UI, which is typically valued higher than one pixel more or less in a specific context.
>2.    If there really isn't anything close to what you want, whether it should be added to the arsenal of default sizes rather than as a local exception.

So I'd say standardize that thing on 7, `Size.line.progress` or something.

https://github.com/koreader/koreader/blob/61a8b63757accbfcb3cf077fc70f9033ba8df0ea/frontend/ui/size.lua#L60-L64

yparitcher:
i can see how the thickness is a little ugly, so i made it the same size as the progress bar (7) but added a margin of 3 so it still can be seen.

how does that look?

poire-z:
> but added a margin of 3 so it still can be seen.

That margin adds a thin line of white above it, which I find less clean that without (without, the bar merges into the black plastic of the device).
If you need that extra height to see it, I'd rather like it to be part of the height, instead of that margin.

Care to show a pic?
Do you use a small Screen DPI?
I would have suspected some pixels are not shown and the KT4 needs to have a viewport to account for that, but I guess if that wass, it would already be known :)
Do you see a varying part of the white margin you added? Depending on how you tilt the device? (I do on my GloHD - when tilted top up 20° from horizontal, I don't see it any more.)

yparitcher:
> Care to show a pic?

i have no way of taking one. sorry.

> Do you use a small Screen DPI?

i use the default, KT4 is 167 PPI
> I would have suspected some pixels are not shown and the KT4 needs to have a viewport to account for that, but I guess if that wass, it would already be known :)
> Do you see a varying part of the white margin you added? Depending on how you tilt the device? (I do on my GloHD - when tilted top up 20° from horizontal, I don't see it any more.)

yes this is due to my screen being slightly recessed so a small section at the edge is hard to tell apart from the case, especially if the device is tilted. (this is why i was not seeing the original bar)

like everything this need to have some distance from the edge in order to be seen on my device, it is the same to me if it is black or has a white margin.

poire-z:
You mean you don't see at all the 3px white margin? Even when tilting bottom up?

yparitcher:
> You mean you don't see at all the 3px white margin? Even when tilting bottom up?

if i look straight i see it, however as i tilt the top towards me i see less, until i don't see it at all

poire-z:
Ok, I get it: when changing some layout option, and it is taking 2mn, you might not keep your device in your hands while doing other stuff, so the need to see more of it at a distance/angle, to know when it is done to get back to it ?
So, some margin or more height ? (Both feels ugly when you're just waiting with your device in your hand with no tilt :)

poire-z:
Having seen it a few times today, I think I'll be ok with the added small margin. It makes it slightly more obvious it is a progress bar. May be use `local y = Size.margin.small` then.

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1030,9 +1030,9 @@ function ReaderRolling:showEngineProgress(percent)
         -- Widget size and position: best to anchor it at top left,
         -- so it does not override the footer or a bookmark dogear
         local x = 0
-        local y = 0
+        local y = 3
         local w = Screen:getWidth() / 3
-        local h = Screen:scaleBySize(5)
+        local h = Screen:scaleBySize(7)
```
Frenzie: This is just using 7, not standardizing to 7.

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1032,7 +1033,7 @@ function ReaderRolling:showEngineProgress(percent)
         local x = 0
         local y = 3
         local w = Screen:getWidth() / 3
-        local h = Screen:scaleBySize(7)
+        local h = Size.line.progress
```
Frenzie: Ideally also applied to the other progressbar but not strictly necessary. :-)

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1032,7 +1033,7 @@ function ReaderRolling:showEngineProgress(percent)
         local x = 0
         local y = 3
```
poire-z: `local y = Size.margin.small`  (`Screen:scaleBySize(2)`, which hopefully makes 3 on your device :)

yparitcher: done

Frenzie: Not likely. :-P The old Kindle 800x600 @ 167 is no longer our baseline since we switched it to an easier 160, but that extra 4 % will only round up for certain fractional values (i.e., the `x.45`–`x.49` range that afaik we don't use).

yparitcher: on my KT4 i dont think `display_dpi` is set so `Screen:scaleBySize()` will scale input by x1.3 rounded up so `Screen:scaleBySize(2)` should be 3

PR #5583, opened 2019-11-10
[feat, UX] Pick keyboard layout from keyboard

yparitcher:
this is a first draft for fixing keyboard locale support
see https://github.com/koreader/koreader/pull/5569#issuecomment-551971811

the idea is that when pressing the globe icon a popup comes up to chose an input language (the same choices that are in the `keyboard layout` menu)
i made some of the necessary changes.

things i need help with:

1. making a popup (similar to the openwith dialog) to chose the language when pressing the globe key.
i put in comments the radio table needed for the languages options but i have no clue how the widget/input ui works to be able to make a proper dialog out of it.

2. fixing all the keyboards:

- make the local characters default: move position 5-8 to 1-4
- shift the `Äéß` from position 9-12 to 5-8
- get rid of the english characters in other languages


frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -571,8 +572,38 @@ function VirtualKeyboard:init()
     end
 end
 
-function VirtualKeyboard:getKeyboardLayout()
-    return G_reader_settings:readSetting("keyboard_layout") or G_reader_settings:readSetting("language")
+function VirtualKeyboard:getKeyboardLocale()
```
Frenzie: I wouldn't change this function name. Not only is "keyboard layout" a standard term, but it's safely decoupled from locale-related shenanigans. Instead this calls up all kinds of locale-related questions and doubts.

poire-z:
> make the local characters default: move position 5-8 to 1-4

Mhhh, that won't do just like that, or you have to fix all the keyboards :)
The 1st layout (except for your hebrew one) is usually the default the keyboard codes wanted. (In the french keyboard, I have some russian on that globe layout.)
We'd need to rethink all current keyboards files then if we go that way.

> get rid of the english characters in other languages

I think english/ascii should be accessible directly from a toggle layout key in all languages, because that's just practical (filenames, ...)

Frenzie:
> Mhhh, that won't do just like that, or you have to fix all the keyboards :)

Indeed, it's not the fault of any other keyboard that the Hebrew keyboard was written one way or the other! There is no bug here whatsoever. If you want the Hebrew keyboard fixed, fix it. Other keyboards should not be broken and nothing should be shifted.

If I recall correctly you said you wrote it this way because it was most convenient for you, switching between English and Hebrew. That's fine, but if that's considered a bug it's exclusively a bug that you purposefully put into that specific keyboard.

> I think english/ascii should be accessible directly from a toggle layout key in all languages, because that's just practical (filenames, ...)

That's not really your or my business to decide but depends on the sensibilities and traditions of the keyboards in question. ;-) The whole point here is to make it unproblematic to quickly switch between layouts.

The most sensible way is to select up to four or five desired layouts that you can easily toggle between. Also see https://github.com/koreader/koreader/pull/5414#issuecomment-533911479

Why five? Because then you can map 'em to swipes on the globe key. ;-) Plus you get a popup along these lines:

poire-z:
> That's not really your or my business to decide but depends on the sensibilities and traditions of the keyboards in question. ;-)

Except keyboard traditions never had 12 layouts like we do :) So we can be creative and build up future traditions :)

Frenzie:
I'd prefer to reduce it to 6 (?) tbh, getting rid of the globe layer entirely. It can simply be refactored into a separate Cyrillic keyboard.

Unfortunately I'm not fully up to snuff on Cyrillic, but at least unlike Arabic I can read most of the characters. :-P

poire-z:
I guess I'd be ok with 6 (all these with some popups where appropriate):

1) lowercase letters in the "locale" (or arabic, hebrew single-case letters, korean thingies)
2) <kbd>Shift</kbd> > uppercase letters in the "locale" (or arabic, hebrew diacritics when no uppercase)
3) <kbd>Sym</kbd> > numbers, arrow keys, punctuations, parenthesis... (the same for all keyboards?)
4) <kbd>Shift</kbd> + <kbd>Sym</kbd> > 2nd layer of punctuation, parenthesis, with arrow keys too
5) <kbd>Äéß</kbd> > alternative lowercase letters, direct accentuated ones, less common letters, or some other common foreign ones, like some greek - or english letters for arabic/hebrew/korean keyboards for quick access (so, the icon would show <kbd>Aes</kbd> :)
6) <kbd>Shift</kbd> + <kbd>Äéß</kbd> > alternative uppercase letters, other stuff, stupid stuff like <kbd>http://</kbd> <kbd>.com</kbd>...
7) <kbd>Globe</kbd> > switch to another layout

And ideally, the activated <kbd>Shift</kbd> <kbd>Sym</kbd><kbd>Äéß</kbd> should be highlighted, so we know where we're at :) so we can quicky unactivate them to get back to the first layout (I sometimes spend too much time on getting back to the first one:)

(Pinging @WaseemAlkurdi who's on the arabic keyboard in #5569.)

Frenzie:
> lowercase letters in the "locale" (or arabic, hebrew single-case letters, korean thingies)

Definitely how I'd do it. Most already work that way regardless. I think Hebrew and the current Arabic PR are the exception rather than the norm. And Cyrillic is an exception too, of course.

> numbers, arrow keys, punctuations, parenthesis... (the same for all keyboards?)

Not necessarily 100 % the same, but it should be pretty close for the most part. I can definitely imagine punctuation being laid out differently (like `«»` for French).

> alternative lowercase letters, direct accentuated ones, less common letters, or some other common foreign ones, like some greek - or english letters for arabic/hebrew/korean keyboards for quick access (so, the icon would show Aes :)

Some layouts (like Chinese and/or Korean) got rid of that button btw.

yparitcher:
the point of this PR is to decouple keyboard language from keyboard layout.


> > Mhhh, that won't do just like that, or you have to fix all the keyboards :)
> 
> @yparitcher Indeed, it's not the fault of any other keyboard that the Hebrew keyboard was written one way or the other! There is no bug here whatsoever. If you want the Hebrew keyboard fixed, fix it. Other keyboards should not be broken and nothing should be shifted.
> 
> If I recall correctly you said you wrote it this way because it was most convenient for you, switching between English and Hebrew. That's fine, but if that's considered a bug it's exclusively a bug that you purposefully put into that specific keyboard.

sorry i was refering to the hebrew and arabic, but as you guys said for french and english the Cyrillic chraracters should be moved into thier own keyboard.

the goal is each keyboard should have one language, however the file needs to be fixed

> I guess I'd be ok with 6 (all these with some popups where appropriate):
> 
>     1. lowercase letters in the "locale" (or arabic, hebrew single-case letters, korean thingies)
> 
>     2. Shift > uppercase letters in the "locale" (or arabic, hebrew diacritics when no uppercase)
> 
>     3. Sym > numbers, arrow keys, punctuations, parenthesis... (the same for all keyboards?)
> 
>     4. Shift + Sym > 2nd layer of punctuation, parenthesis, with arrow keys too
> 
>     5. Äéß > alternative lowercase letters, direct accentuated ones, less common letters, or some other common foreign ones, like some greek - or english letters for arabic/hebrew/korean keyboards for quick access (so, the icon would show Aes :)
> 
>     6. Shift + Äéß > alternative uppercase letters, other stuff, stupid stuff like http:// .com...
> 
>     7. Globe > switch to another layout
> 
> 
> And ideally, the activated Shift SymÄéß should be highlighted, so we know where we're at :) so we can quicky unactivate them to get back to the first layout (I sometimes spend too much time on getting back to the first one:)
> 

this is the point however <kbd>Globe</kbd> > is not switching to a layout rather to a different language. 
each language has up to 8 layouts, instead of the 12 we have now.

this is necessary in order to scale keyboard languages

> The most sensible way is to select up to four or five desired layouts that you can easily toggle between. Also see #5414 (comment)

> Why five? Because then you can map 'em to swipes on the globe key. ;-) Plus you get a popup along these lines:

this is a good idea to implement, however it would require new menu setting to select which are enabled, also to ensure that the current is enabled. and would require reworking the whole lang to keyboard table to track what the current layout is.

frontend/ui/widget/virtualkeyboard.lua
yparitcher: as i commented below i am trying to separate language from layout, as calling it a layout is more confusing to me.

if `Locale` is confusing i can change it to `getKeyboardLanguage` which is more definitive.

Frenzie: >  i am trying to separate language from layout

Which is precisely why it should be layout — not locale and especially not language.

yparitcher: the way i see it is that the 8 or 12 ways in each language are layouts, there are many languages and each language has different layouts. if you want that languages should be called layouts than the current layouts should be called something like keymaps

Frenzie:
> the point of this PR is to decouple keyboard language from keyboard layout.

For terminological purposes I think a term like "keyboard language" should be torpedoed in the bow before it can sail on. ;-) A full-on Dutch keyboard would refer either to the Dutch QWERTY layout (effectively replaced by US QWERTY since the 2000s) or Dutch/Belgian AZERTY (still a thing). There may also have been Dutch QWERTZ, or in any case QWERTZ typewriters were around in the Netherlands.

The tl;dr is that there are at least three distinct keyboard layouts associated with the Dutch language, probably four. English has at least two (US & UK). More for Dutch and English alike if you also count ANSI vs ISO. (I can't stand ISO.) The fact that the current two main keyboards for Dutch are currently referred to as "English" and "Français" in the KOReader interface is mildly unfortunate. KDE presents it like this, but Windows and Mac are very similar:

Screenshot_20191110_213754

A "locale" is a set of properties typically associated with a physical locale. Netherlands = € 1.000,00 with a US English International keyboard layout.

A keyboard layout can be part of a locale, but to refer to a keyboard layout as a keyboard locale is just confusing at best because it implies all that other baggage.

The code currently says something like `setLayout`. If that's confusing, please change it to `setLayer` or `setLevel`.

One layout has multiple layers, also known as levels. The base layer, the shift layer, and various other layers. In KOReader it's called the symbol layer, on a physical keyboard the Alt Gr. layer or the third layer (Shift being the second layer and Shift+Alt Gr. the fourth layer).

With all that out of the way, I don't really know what you mean by decoupling the keyboard language from its layout. I decoupled the UI language from the keyboard layout somewhat recently (Device → Keyboard layout), and the way I see this PR is just to make that slightly easier to put into effect, i.e., without closing the keyboard and going into a menu. The code so far is fully in line with that.

yparitcher:
Now that that is out of the way, 
can someone help me make a widget to chose the layout, and or help implement @Frenzie's idea

>> The most sensible way is to select up to four or five desired layouts that you can easily toggle between. Also see #5414 (comment)
> > Why five? Because then you can map 'em to swipes on the globe key. ;-) Plus you get a popup along these lines:
> 
> this is a good idea to implement, however it would require new menu setting to select which are enabled, also to ensure that the current is enabled. and would require reworking the whole lang to keyboard table to track what the current layout is.

poire-z:
> can someone help me make a widget to chose the layout, and or help implement @Frenzie's idea

You could get both :) by starting with the existing keyboard popup widget (re-using/extending/duplicating it), and you may quickly get something like that:

<kbd>image</kbd>

Then, may be instead of having a grid of squares, just have stacked rectangles with the full language name.
But I'd be fine with these 2 letters squares, as they would allow at least 5 swipes to quickly switch between them - and eventually add something so the most recently used are to be in these 5 squares, or allow organizing and ordering them into these 5 squares with our SortWidget.

Frenzie:
The current menu could activate multiple layouts easily with only a minor adjustment. I'd start with that and the actual layout toggling (by tapping the globe key).

The swipe/popup thing is more of a later refinement/addition, where instead of having to tap two or three times you can just swipe in a particular direction.

Frenzie:
@yparitcher Alright, with 2019.11 out the door I can now focus on this. https://github.com/koreader/koreader/pull/5588 implements the relevant popup functionality.

yparitcher:
> @yparitcher Alright, with 2019.11 out the door I can now focus on this. #5588 implements the relevant popup functionality.

thanks, but i don't think that is going to be enough, as that allows for a callback but does not yet allow dynamically assigning the key label this also has to be customizeable
https://github.com/koreader/koreader/blob/2855fdfdac5a7357dc7be67f03950df5aa872386/frontend/ui/widget/virtualkeyboard.lua#L401-L408

unless i am understanding this wrong

Frenzie:
I can probably add that tomorrow, but I don't think it's related to the few lines you quoted.

I'd add a `VirtualKey:updateText()` method (or some such) which you would then call using `parent_key:updateText(new_lang)` in the callback.

Alternatively it could be made slightly more complex with something like a label_func, in which case a simple `update_keyboard()` would suffice. Probably not really worth it for such a one off thing.

What exactly are you envisioning? Text next to the globe? Instead of the globe?

yparitcher:
the popup keys have to be dynamically assigned to a language so the label on them matches to the language set and to the enabled languages. also the globe should probably change to the current layout once the popup appears.

yparitcher:
i think i figured out how to make the popup, however i can't get it to close the popup once new language is selected. also for now i just used the first five languages just to get it working, but they should be assigned dynamically based on user choice

Frenzie:
Yes, it should be dynamically generated but I see you mostly figured it out. ;-)

Like I said though, I consider the swipe & popup of secondary concern. Just a simple toggle through two or three layouts would do wonders for starters.

yparitcher:
i would like to make a widget like this:

Screenshot_2019-11-12-09-18-27

to be the <kbd>globe</kbd> callback. it should have all the languages as options.

i put in comments the code necessary to generate the `radiobuttontable` however i have no clue which widget to use / extend to get this result. please help.

at the same time the virtualkeypopup will be as the holdcallback, assigned from a list of enabled languages.

Frenzie:
The dialog itself would be nice to have, although I'd probably use it as one of the swipe/popup directions (like east). I doubt many people would use more than two or three layouts, in which case one or two taps is a lot quicker than hold & move & tap.

Anyhow, widgets are normally put in containers. You can nest them in all kinds of fun ways. It'd probably be a lot easier to copy the OpenWithDialog and to remove what's not necessary instead of approaching it the other way around?

Frenzie:
What precisely is the problem with the current code btw? I just fetched your PR here and the commented out code is kind of floating there disconnected from anything else.

yparitcher:
i tried adapting the OpenWithDialog as you can see in my latest commit, however it requires `InputDialog` which requires `VirtualKeyboard` via `InputText` leading to a circular dependency that i could not solve.

Frenzie:
An InputDialog is basically a frame with a text input and a couple of buttons. It's a tad too late now to take a look, but InputContainer is the parent for dealing with input events. I apologize for the suggestion if OpenWithDialog leans too strongly on InputDialog, although atm I'd still be inclined argue that then it's just a case of stripping out the InputDialog and replacing it with a bare InputContainer. :-)

yparitcher:
getting there.

The `virtualkeypopup` still has to be made dynamic by adjusting the keyboard layout menu to allow for a few languages and using those languages.

---

Reader_2019-Nov-12_202601

---

Reader_2019-Nov-12_200917

yparitcher:
and we need to find a way to close the `VirtualKeyPopup` on language change.

Frenzie:
You should probably pass through a reference to self, something like (untested):

```diff
@@ -231,12 +242,12 @@ function VirtualKey:onTapSelect(skip_flash)
         self[1].inner_bordersize = self.focused_bordersize
         self:update_keyboard(false, true)
         if self.callback then
-            self.callback()
+            self.callback(self)
         end
         UIManager:tickAfterNext(function() self:invert(false) end)
     else
         if self.callback then
-            self.callback()
+            self.callback(self)
         end
     end
     return true
@@ -431,6 +442,7 @@ function VirtualKeyPopup:init()
                 local label = v.label or key
                 local icon = v.icon
                 local virtual_key = VirtualKey:new{
+                    popup = self,
                     key = key,
                     label = label,
                     icon = icon,
```

Then you'd call `UIManager:close(self.popup)`, for example.

yparitcher:
got the popup to close, however the key (layout) that was pressed is not closed.

---
Reader_2019-Nov-13_195408

---

Reader_2019-Nov-13_195115

yparitcher:
made virtual popup dynamic however what should be done about `pt_BR` and `ko_KR` that are longer than 2 chars and dont fit into a key?

yparitcher:
ready for review with the 2 above mentioned bugs, please help me fix them.

please make sure the widget code is up to par as this is my first time realy writing a widget.

the remaining keyboard files need to be converted to having up to 8 layers, preferably only chars from one language, and the must have the <kbd>Globe</kbd> key.

poire-z:
> however the key (layout) that was pressed is not closed.

I can't reproduce it. Guess you solved that in later commits?

> please make sure the widget code is up to par as this is my first time realy writing a widget.

Nothing suspicious jumps out at me :)

Just had a quick try, and some remarks, not all related to your work:

<kbd>image</kbd>
Thanks for the highlight (grey background) of the Shift and other keys. Except that it looks ugly when a white background without transparency is used :) Although I dunno if the image has really no transparency (alpha stuff I don't fully grasp, or if we just need to add some `alpha=true` to some ImageWidget somewhere now that it's needed. @NiLuJe ?)

<kbd>image)</kbd>
@NiLuJe : your nerdfont is a quite unpolite font :) and should probably be moved after the more correct FreeSans/FreeSerif fonts - but I guess your icons would be found in FreeSerif - so probably better cleaned up of chars in invalid ranges.
In that nerdfont, there are some icons in some non-private non-symbol unicode ranges, like arabic presentational forms, and as seen here in some hebrew range. (and when trying to find some new icons for the footer, while I was working on that UI RTL code, suddently, the footer was all messed up because some nice icon found in that arabic presentational forms unicode range were making my footer text RTL - as with mostly digits and punctuation in the footer, all direction neutral, that icon was a strong RTL arabic char that imposed its direction to the stuff after it :)

<kbd>image</kbd>
This happened when hitting outside your new widget: lower widgets got my taps and handled them, I could hit the hide keyboard key, then hit it again to show keyboard again, I could then hit the keys and enter text - I was not blocked and could get out of it, by re-hiding the keyboard.
@Frenzie: I guess we're in the situation of multiple modal that I evoked in https://github.com/koreader/koreader/pull/5573#issuecomment-550496568.

frontend/ui/widget/keyboardlayoutdialog.lua
```diff
@@ -0,0 +1,179 @@
...
+end
+
+function KeyboardLayoutDialog:onCloseWidget()
+    UIManager:setDirty(nil, function()
+        return "partial", self[1][1].dimen
+    end)
+    return true
+end
+
+return KeyboardLayoutDialog
```
poire-z: I think this widget should close when we tap outside, for easier getting rid of it :)
(see other widgets that have a onTapClose handler, like frontend/ui/widget/hyphenationlimits.lua).

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -45,16 +47,29 @@ local VirtualKey = InputContainer:new{
...
     elseif self.keyboard.utf8mode_keys[self.label] ~= nil then
-        self.callback = function () self.keyboard:setLayout("IM") end
+        self.callback = function ()
+            self.keyboard_layout_dialog = KeyboardLayoutDialog:new{
+                parent = self,
+            }
+            UIManager:show(self.keyboard_layout_dialog)
+        end
+        self.hold_callback = function()
+            self.key_chars = self:genkeyboardLayoutPopup(self)
```
poire-z: I'd personally prefer if these 2 were inverted:
Tap to get the keypopup and quick switching.
Hold to show your new widget only when needed, when I need another language I have not yet checked to be in the keypopup.

Frenzie: These key_chars need to be outside hold_callback to they can also be used in swipe_callback.

(I've been trying to hint at swipes for a bit but I guess I wasn't explicit enough. :-P)

frontend/ui/widget/keyboardlayoutdialog.lua
Frenzie: Possibly, I'm not sure I agree with that though. I think you might say that mostly because just tapping is a rather easy way to bring up this dialog.

I think tapping should just switch to the next layout, with special dialogs relegated to hold. I'd put activating this dialog on the east side of the swipe/popup.

Frenzie:
> Although I dunno if the image has really no transparency (alpha stuff I don't fully grasp, or if we just need to add some alpha=true to some ImageWidget somewhere now that it's needed. @NiLuJe ?)

It's possible that not all images properly have a transparent background, but they're easy to fix up if they don't.

> @Frenzie: I guess we're in the situation of multiple modal that I evoked in #5573 (comment).

I'll have a look in some follow-up.

Frenzie:
@yparitcher 
> made VirtualKeyPopup dynamic however what should be done about pt_BR and ko_KR that are longer than 2 chars and dont fit into a key?

Step 1: add a screenshot so we can consider it regardless of access to dev machines. ;-)

We might want to just allow it to grow a little bigger if required.

> please make sure the widget code is up to par as this is my first time realy writing a widget.

The code is fine (give or take some nits), it's the behavior that I don't think is sufficiently polished yet. Direction should always be stable. If I have multiple active keyboards, they shouldn't randomly switch positions unless there's a very good reason for them to. The whole point of swipes is to instantaneously switch to the keyboard you want.

The popup is intended as a visual aid for swipes. You can use it as your primary interaction pattern if you desire, but there shouldn't be a keyboard popup without swipes.

Holding is seemingly broken until you set up the active keyboards. At the very least the quick & easy solution would be a little info popup about setting default keyboards, though this wouldn't be an issue in the first place in my sketched interaction pattern.

Capitalizing the text in the popup is kind of ugly imho, and also not the way it's done in Windows/Android/Linux DEs.

Screenshot of the text overflowing & the popup randomly changing:


@poire-z 
> I can't reproduce it. Guess you solved that in later commits?

It only happens when doing hold, let go, tap as opposed to the more self-evident hold, pan, let go. It's a minor polish issue, though at a glance I'm surprised by it because I purposefully wrote UIManager:close() and not something like popup:onClose() even if the code is taking a slightly different approach to doing essentially the same thing I proposed. I'll have to read through it in more detail.

Frenzie:
@poire-z I don't really understand how to reproduce your issue. The popup blocks access to the keyboard for me.

frontend/ui/elements/menu_keyboard_layout.lua
```diff
@@ -6,12 +6,25 @@ local sub_item_table = {}
 
 for k, _ in orderedPairs(VirtualKeyboard.lang_to_keyboard_layout) do
     table.insert(sub_item_table, {
-        text = Language:getLanguageName(k),
+        text_func = function()
+            local text = Language:getLanguageName(k)
+            if VirtualKeyboard:getKeyboardLayout() == k then
+                text = text .. "   ★"
```
Frenzie: Template (for Hebrew no less! ;-))

frontend/ui/widget/keyboardlayoutdialog.lua
```diff
@@ -0,0 +1,179 @@
+--[[--
+This widget displays an keyboard layout dialog.
```
Frenzie:
```suggestion
This widget displays a keyboard layout dialog.
```

frontend/ui/widget/keyboardlayoutdialog.lua
```diff
@@ -0,0 +1,179 @@
...
+    width = Screen:getWidth() * 0.8,
+    face = Font:getFace("cfont", 22),
+
+    title_face = Font:getFace("x_smalltfont"),
+
+    title_padding = Size.padding.default,
+    title_margin = Size.margin.title,
+    button_padding = Size.padding.default,
+    border_size = Size.border.window,
+
```
Frenzie:
```suggestion
```
Bit heavy on stray/double newlines. :-)

frontend/ui/widget/keyboardlayoutdialog.lua
```diff
@@ -0,0 +1,179 @@
...
+    table.insert(buttons, {
+        {
+            text = _("Cancel"),
+            callback = function()
+                UIManager:close(self.parent.keyboard_layout_dialog)
+            end,
+        },
+        {
+            text = _("Chose Language"),
```
Frenzie: I'd say something like:
```suggestion
            text = _("Switch to layout"),
```

(The text as written should read `Choose layout`, but that aside.)

frontend/ui/widget/keyboardlayoutdialog.lua
```diff
@@ -0,0 +1,179 @@
...
+                    h = self.radio_button_table:getSize().h,
+                },
+                self.radio_button_table,
+            },
+            CenterContainer:new{
+                dimen = Geom:new{
+                    w = self.title_bar:getSize().w,
+                    h = Size.span.vertical_large*2,
+                },
+                LineWidget:new{
```
Frenzie: This line should be removed unless you intend to add a "make default" type checkbox below, similar to the open with dialog. (Which might not be a bad idea.)

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -162,9 +177,40 @@ function VirtualKey:init()
...
         }
     end
+    if (self.keyboard.shiftmode_keys[self.label] ~= nil  and self.keyboard.shiftmode) or
+    (self.keyboard.umlautmode_keys[self.label] ~= nil and self.keyboard.umlautmode) then
+        self[1].background = Blitbuffer.COLOR_LIGHT_GRAY
+    end
     self.flash_keyboard = G_reader_settings:readSetting("flash_keyboard") ~= false
 end
 
+function VirtualKey:genkeyboardLayoutPopup(parent)
```
Frenzie: Seems more logical as a method on the popup instead?
```suggestion
function VirtualKeyPopup:genkeyboardLayoutPopup(parent)
```

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -398,9 +444,13 @@ function VirtualKeyPopup:init()
                     h_key_padding[i].width = h_key_padding[2].width
                 end
 
+                local key = v.key or v
```
Frenzie: That could be dangerous (even if the below isn't currently expected), e.g.:

```
> v = true
> print(v.lala)
stdin:1: attempt to index global 'v' (a boolean value)
stack traceback:
        stdin:1: in main chunk
        [C]: in ?
```

It's probably more future proof to write something like:
```lua
type(v) == "table" and v.key or v
```

NiLuJe:
@poire-z: The shift icon (and the few others we might be using in the keyboard) don't have an alpha channel, so that wouldn't help ;).

On the other hand, I'm fairly sure we have a shift/enter/globe symbol available somewhere, so we probably ought to use those instead of a bitmap ;).

----

I'm also aware of the unfortunate PUA issues w/ nerdfont. I was vaguely hoping that putting Noto Sans Arabic UI *above* it would take care of the brunt of it, but do ping me if that doesn't handle everything. There are a few issues open upstream with a few ideas of where to move what.

(We do need to keep it above Free*, as those also use the PUA for some stuff. I recall ligatures in particular, but I had a clash with one of our icons with something else, too).

poire-z:
> I don't really understand how to reproduce your layered widget issue. The popup blocks access to the keyboard for me. Does this involve an always_active widget? What are the precise steps?

You need something that bleed outside the popup :) Like the TextEditor plugin.
Open some file, tap globe to bring the popup. Then, hit that button:
<kbd>image</kbd>.
The keyboard disappear and the TextEditor takes the full screen.
Hit that same key back, and it brings the keyboard back, above the selection popup.
(To be able to close that popup, you need to hide the keyboard again by hitting that same key).

poire-z:
> I was vaguely hoping that putting Noto Sans Arabic UI above it would take care of the brunt of it

Nope. It might override the arabic range that nerddont is usurping. But we might need a Noto Sans Hebrew UI to kick it ouf of the hebrew ranges it is squatting.

But I'm bothered we'd have to add `Noto Sans Arabic`, `Noto Sans Hebrew`, and one for earch of the dozen india languages :) which shouldn't have each more of a few hundreds glyphs - so, they could fit all in a `Noto Sans AllEasternLanguages WithoutCJK.otf`and we would be fine... A kind of have-it-all font like FreeSerif is :)
Could we build a merged one? Because the harfbuzz code I'll bring soon has Harfbuzz do some work with each of the fallback fonts until one has the glyphs - a bit non economical :)

NiLuJe:
@poire-z: Nope, there's no collection variant for non-latin Eastern scripts like there is for CJK.

And if there were, that'd still be a collection (so, essentially, multiple families in a single files), which wouldn't do anything to help performance-wise, it'd just save some space.

I'm not aware of any silver-bullet on the fallback front, short of switching to fontconfig?

poire-z:
> Nope, there's no collection 

OK, there's none :) But could we build one? A frankenfont by merging all the interesting Noto Sans (Arabic|Hebrew|Indic|...) UI.otf in a single `Noto Sans EasternKoreaderMerged UI.otf `?
Legally?
Technically? Assuming all Noto Sans xxx UI share the same metrics and it could work? Can't Fontforge or your favorite command line utilities help ? :)

We can do without fontconfig (I think that's the main bonus of the shaping code I'll bring: iterating fallback fonts until the required glyphs are found - it comes with some caveat as the spaces and punctuations will be drawn with the first font, but well, it works well and I didn't notice any bad glyph misalignment...)
For our UI, with some small revamp of font.lua, we could optimize stuff a bit by setting NotoSansArabic as the default font (for infont, smallinfofont...) when the UI language is set to arabic.
Then, it's just when reading non-default-language Wikipedia lookups, or looking at some specific EPUB's View book description, that some fallback font may be needed.
But still, I'd like a single magic fallback font :)

Anyway, should NotoSans Hebrew UI be added to kill nerdfont usurping these hebrew slots?
Or should our nerdfont be cleaned up (can we, legally? then why not Noto Sans UI, legally? :)

Frenzie:
@poire-z Still looks like an `is_always_active` issue to me, not a modal issue?

frontend/ui/widget/keyboardlayoutdialog.lua
```diff
@@ -27,14 +27,11 @@ local KeyboardLayoutDialog = InputContainer:new{
     modal = true,
     width = Screen:getWidth() * 0.8,
     face = Font:getFace("cfont", 22),
-
```
Frenzie: Ftr, while not exactly applicable here there's a difference between spacing for ordering and random stray lines (like for the closing `}`). :-)

Frenzie:
Basically anything is_always_active is supposed to always catch events. That's not a bug.

frontend/ui/widget/keyboardlayoutdialog.lua
yparitcher: i got rid of them it was mostly there as i copied from input dialog

NiLuJe:
@poire-z: I've never done anything of the sort, but I suppose it's probably scriptable via fontforge.

The issue being that I'm fairly sure each script has very specific and distinct metrics requirements... ;).

Regardless, I'll fix nerfdont and move it down, because that just bugs me, too ;).

frontend/ui/widget/virtualkeyboard.lua
yparitcher: as of now:
swipe works.
tap brings the `VirtualKeyPopup` where the gobe icon will bring up the full popup
hold brings up the whole popup

does that work?

frontend/ui/elements/menu_keyboard_layout.lua
yparitcher: i just copied from the font menu code, for now this will only be the english code for the language that is not translated.

frontend/ui/widget/virtualkeyboard.lua
yparitcher: changed to `VirtualKey:genkeyboardLayoutKeyChars(parent)`

frontend/ui/widget/virtualkeyboard.lua
yparitcher: also if there are no active keyboards, swipe does nothing and tap brings up the full popup

Frenzie:
When you wrote you want the popup to close I thought that was because the regular popup behavior was to close as well, but it's not. I don't think the popup should behave any differently than any other popup. You could argue *all* popups should close on tap, but I'd prefer to leave that that to a different PR.

As to the button that remains behind, it's because of the `self:invert()`. The easiest but not necessarily best way to avoid that is to schedule closure just like the invert is scheduled, i.e.:

```diff
diff --git a/base b/base
--- a/base
+++ b/base
@@ -1 +1 @@
-Subproject commit c140b752e16163ee8836214787caf7343bc4a579
+Subproject commit c140b752e16163ee8836214787caf7343bc4a579-dirty
diff --git a/frontend/ui/widget/virtualkeyboard.lua b/frontend/ui/widget/virtualkeyboard.lua
index 89a0ba51..c8c3f82c 100644
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -214,7 +214,7 @@ function VirtualKey:genkeyboardLayoutKeyChars()
         if v == true and G_reader_settings:readSetting("keyboard_layout") ~= k then
             key_chars[positions[index]] = string.upper(k)
             key_chars[positions[index] .. "_func"] = function()
-                UIManager:close(self.popup)
+                UIManager:tickAfterNext(function() UIManager:close(self.popup) end)
                 self.keyboard:setKeyboardLayout(k)
             end
             if index >= 5 then break end
```

PS The latest change has broken expected hold behavior. The interaction methods for a key are:

1. Swipe in a direction.
2. Hold & drag in a direction, let go.
3. Hold, let go, and tap.

yparitcher:
@Frenzie 

> The code is fine (give or take some nits), it's the behavior that I don't think is sufficiently polished yet. Direction should always be stable. If I have multiple active keyboards, they shouldn't randomly switch positions unless there's a very good reason for them to. The whole point of swipes is to instantaneously switch to the keyboard you want.
> 
> The popup is intended as a visual aid for swipes. You can use it as your primary interaction pattern if you desire, but there shouldn't be a keyboard popup without swipes.

the popup positions are assigned alphabetically, from the list of active layouts up until 5, (all additional ones are ignored)
the current layout is not assigned a popup, this causes all the layouts alphabetically after it to be bumped up one slot.
the locations cannot be statically assigned as there are more languages than popup locations for now.
the only solution would be to also assign the current layout a popup which would then guarantee that for any set of active languages the popup will always have the same order.

Frenzie:
> the current layout is not assigned a popup, this causes all the layouts alphabetically after it to be bumped up one slot.

It should be. Period.

Frenzie:
This:

Screenshot_2019-11-14_20-12-35

Should have three buttons:

Screenshot_2019-11-14_20-12-48

yparitcher:
> > the current layout is not assigned a popup, this causes all the layouts alphabetically after it to be bumped up one slot.
> 
> It should be. Period.

why? if i want to keep my layout i will just tap outside, the point of the swipe and popup is to change to **other** languages.

poire-z:
> Still looks like an is_always_active issue to me, not a modal issue?
> Basically anything is_always_active is supposed to always respond to events. That's not a bug.

You mean it's not a bug at all? Not a bug in our current UIManager code, and not a bug in this PR?
The behaviour feels buggy to me (although rare enough I would not care if we don't fix it).
But I'd expect this KeyboardLayoutDialog to prevent any interaction with the textbox widget or the keyboard below.
This popup has modal=true and is_always_active=true. Should one be removed? If we can't, should the same properties of the widgets below be temporarily switched to false? Or else?

yparitcher:
> As to the button that remains behind, it's because of the `self:invert()`. The easiest but not necessarily best way to avoid that is to schedule closure just like the invert is scheduled, i.e.:
> 
> ```diff
> diff --git a/base b/base
> --- a/base
> +++ b/base
> @@ -1 +1 @@
> -Subproject commit c140b752e16163ee8836214787caf7343bc4a579
> +Subproject commit c140b752e16163ee8836214787caf7343bc4a579-dirty
> diff --git a/frontend/ui/widget/virtualkeyboard.lua b/frontend/ui/widget/virtualkeyboard.lua
> index 89a0ba51..c8c3f82c 100644
> --- a/frontend/ui/widget/virtualkeyboard.lua
> +++ b/frontend/ui/widget/virtualkeyboard.lua
> @@ -214,7 +214,7 @@ function VirtualKey:genkeyboardLayoutKeyChars()
>          if v == true and G_reader_settings:readSetting("keyboard_layout") ~= k then
>              key_chars[positions[index]] = string.upper(k)
>              key_chars[positions[index] .. "_func"] = function()
> -                UIManager:close(self.popup)
> +                UIManager:tickAfterNext(function() UIManager:close(self.popup) end)
>                  self.keyboard:setKeyboardLayout(k)
>              end
>              if index >= 5 then break end
> ```

this works, thanks


> When you wrote you want the popup to close I thought that was because the regular popup behavior was to close as well, but it's not. I don't think the popup should behave any differently than any other popup. You could argue _all_ popups should close on tap, but I'd prefer to leave that that to a different PR.

> PS The latest change has broken expected hold behavior. The interaction methods for a key are:
> 
>     1. Swipe in a direction.
> 
>     2. Hold & drag in a direction, let go.
> 
>     3. Hold, let go, and tap.

if you want to have the VirtualKeyPopup on tap that is also different than other keys, if you want this to be like other keys than the way i had it originally is correct
the only reason i changed it was cuz you guys wanted the VirtualKeyPopup on tap

yparitcher:
@Frenzie @poire-z 
can we have a consensus of what each interaction should do so we are all on the same page 

there is swipe - probably everyone can agree
there is tap - should it be next layout, VirtualKeyPopup, full popup, (all this with and without active keyboards)
there is hold - same question as tap

poire-z:
I previously stated my preference :) And even if it's different from other keys (after all, Shift and Sym too are different: they don't outpout anything :), I'd rather have the keypopup that I can quicky dismiss by re-taping on the globe - or taping outside.
Once you've selected your prefered layouts, there is no need for the KeyboardLayoutDialog - or at least not as easy as a simple tap. So, bringing KeyboardLayoutDialog on hold is fine, less easy to bring up - but I'll be fine too with just having the keys popup too on hold :)
Or possibly more natural:
- tap on globe switches to previous layout (so, it behave somehow like Shift and Sym)
- hold brings the key popup (as other keys do)
- swipe, we agree: it bring the one in the keypopup in that direction
- and then I don't know what to do with your KeyboardLayoutDialog :) a dedicated key in that virtualkeypopup <kbd>...</kbd>?

Frenzie:
@yparitcher 
> why? if i want to keep my layout i will just tap outside, the point of the swipe and popup is to change to other languages.

The result of the same action should always be the same if at all reasonably possible. It should not be dynamic.

Having to read stuff on the screen is *not* a feature. A UI that can't be learned is a bad one.

The advantage of swipes is that you can immediately swipe in a particular direction to switch to a specific layout. The popup uses the exact same principle, except it offers some visual support. The only difference is you hold it for a split second longer.

@poire-z 
> Not a bug in our current UIManager code

I wouldn't call it one.

> not a bug in this PR?

Obviously it's an issue. ;-)

@yparitcher 

> there is swipe - probably everyone can agree

Can we? ;-) Because swipe == popup and you're quite busy disagreeing. :-P

> there is tap - should it be next layout, VirtualKeyPopup, full popup, (all this with and without active keyboards)

I'm in favor of next layout, but while I'm not convinced the dialog popup is quite right for simple tap I think it does better at not breaking expected behavior.

> there is hold - same question as tap

Hold is expected to show a popup.

@poire-z 

> and then I don't know what to do with your KeyboardLayoutDialog :) a dedicated key in that virtualkeypopup ...?

That's what I've been saying all along. Swipe right for the complex dialog. ;-)

Frenzie:
Btw, can we make the KeyboardLayoutDialog immediately preview the keyboard layout? It'd have to move up slightly for that of course. Cancel can simply reset it to the starting value.

Frenzie:
Never mind that for the moment, looks like I didn't write the RadioButtonTable sufficiently easily extensible in that regard. Just some polish for a follow-up PR.

yparitcher:
so i think i have it.

tap is next layout (of enabled, if there are no enabled then of all layouts).
hold brings the key popup (if there are enabled layouts, otherwise straight to the KeyboardLayoutDialog)
swipe, we agree: it bring the one in the keypopup in that direction.

the current layout will also get a key for consistency.

please confirm.

Frenzie:
> tap is next layout (of enabled, if there are no enabled then of all layouts).

I wouldn't cycle through all layouts; that just sounds confusing. I'd either bring up the KeyboardLayoutDialog, show a little notification about enabling keyboard layouts, or do nothing.

> hold brings the key popup (if there are enabled layouts, otherwise straight to the KeyboardLayoutDialog)

Sounds good. :+1:

poire-z:
> please confirm.

Sounds good to me too :)

yparitcher:
> > tap is next layout (of enabled, if there are no enabled then of all layouts).
> 
> I wouldn't cycle through all layouts; that just sounds confusing. I'd either bring up the KeyboardLayoutDialog, show a little notification about enabling keyboard layouts, or do nothing.

right now it does nothing.

the next layout cycle has a lot of loops if someone sees a way to simplify it please do.

yparitcher:
screenshots:

hold popup:
Reader_2019-Nov-14_154839

KeyboardLayoutDialog:
Reader_2019-Nov-14_154845

poire-z:
For the PT_BR, unless there is a way to abbreviate it ("br", "ko"), an option would be to do like we do some other places: get the width of the text after rendering: if it's too wide, loop while reducing the font size until it fits.

Frenzie:
For this specific application, if you want to keep all five characters a manual smaller font size would suffice.

While making it slightly smaller, you could also transform the underscore into a newline:
```
pt
BR
```

Or just:
```
ptBR
```

(Just a matter of reducing the label size less for no good reason.)

> the next layout cycle has a lot of loops if someone sees a way to simplify it please do.

Indeed it does. You might be able to work something out something more efficient with `next()` but as long as the behavior is correct I'm fine leaving that as is for the initial PR.

yparitcher:
> > the next layout cycle has a lot of loops if someone sees a way to simplify it please do.
> 
> Indeed it does. You might be able to work something out something more efficient with `next()` but as long as the behavior is correct I'm fine leaving that as is for the initial PR.

`next()` does not guarantee order, so it would not always give the same behavior.

Frenzie:
Oh right, arbitrary order. ^_^

WaseemAlkurdi:
Forgive me if I misunderstood the way this implementation works, but if we implement multiple keyboards this way, don't we have to remove the Latin character layout in each and every keyboard?

yparitcher:
> Forgive me if I misunderstood the way this implementation works, but if we implement multiple keyboards this way, don't we have to remove the Latin character layout in each and every keyboard?

yes that is the plan, they should only be needed in the latin based languages.

Frenzie:
The 🌐 works fine btw (here instead of Sym, left of the current icon for easy contrast):

Screenshot_2019-11-15_14-56-14

The ⇧ is less obviously good; there might be other options too if you look in Unicode arrows.

Screenshot_2019-11-15_14-58-16

NiLuJe:
@Frenzie: Hmm, U+21E7 looks thicker in FreeSerif (at least in FF), but it's common enough that something higher in the list might be covering it...

yparitcher:
everything should be finished on my end, all that is needed to to consolidate each keyboard layout to 8 layers with one set of characters and get rid of the unnecessary ones.

Reader_2019-Nov-17_220613

Frenzie:
Alright, I'll give it a look later. :+1:

Frenzie:
When no layouts are enabled:
```
./luajit: ./ffi/util.lua:520: bad argument #1 to 'pairs' (table expected, got nil)
stack traceback:
        [C]: in function 'pairs'
        ./ffi/util.lua:520: in function '__genOrderedIndex'
        ./ffi/util.lua:535: in function '(for generator)'
        frontend/ui/widget/virtualkeyboard.lua:62: in function 'callback'
        frontend/ui/widget/virtualkeyboard.lua:294: in function 'handleEvent'
        frontend/ui/widget/container/inputcontainer.lua:259: in function 'handleEvent'
        ...
        frontend/ui/uimanager.lua:674: in function 'sendEvent'
        frontend/ui/uimanager.lua:1014: in function 'handleInputEvent'
        frontend/ui/uimanager.lua:1075: in function 'handleInput'
        [C]: at 0x55d4de3fd77e
```

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -45,16 +47,60 @@ local VirtualKey = InputContainer:new{
...
     elseif self.keyboard.shiftmode_keys[self.label] ~= nil then
-        self.callback = function () self.keyboard:setLayout("Shift") end
+        self.callback = function () self.keyboard:setLayer("Shift") end
         self.skiptap = true
     elseif self.keyboard.utf8mode_keys[self.label] ~= nil then
-        self.callback = function () self.keyboard:setLayout("IM") end
+        self.key_chars = self:genkeyboardLayoutKeyChars()
+        self.callback = function ()
+            local current = G_reader_settings:readSetting("keyboard_layout")
+            local keyboard_layouts = G_reader_settings:readSetting("keyboard_layouts")
```
Frenzie: Like elsewhere in the code.
```suggestion
            local keyboard_layouts = G_reader_settings:readSetting("keyboard_layouts") or {}
```

yparitcher: fixed

Frenzie:
Thanks for putting in the effort! :+1:

PR #5584, opened 2019-11-10
[UX] Reduce ButtonTable vertical separator height

Frenzie:
In preparation for 2019.11 this commit improves the visual style
without completely overhauling it yet.

Cf. https://github.com/koreader/koreader/pull/5554#issuecomment-548931725

### DIFF
--- a/frontend/ui/widget/buttontable.lua
+++ b/frontend/ui/widget/buttontable.lua
@@ -73,7 +73,7 @@ function ButtonTable:init()
                 background = Blitbuffer.COLOR_DARK_GRAY,
                 dimen = Geom:new{
                     w = self.sep_width,
-                    h = button_dim.h,
+                    h = button_dim.h/3*2,
                 }
             }
             buttons_layout_line[j] = button


poire-z:
Mhhh, I don't like the smaller vertical separator, it feels less clean to me (on your screenshot and on my Kobo). I may get used to it, but that's my first impression.
(Feels like we got no better option than using a pipe character `|` as a separactor :))

(You can revert #5554 if you'd like us more time to rethink the whole thing.)
(Also, if you want to keep that, please wrap your `blah/3*2` it in a `math.floor()/ceil()` (fractional x/y/w/h had caused me headaches in the past :)

Frenzie:
Various other UIs use the same trick to make the line feel less tall than it is. My proposal to get rid of the separator lines still stands too. ;-)

poire-z:
> Various other UIs use the same trick to make the line feel less tall than it is.

May be they also reduce and cut the long horizontal separator, so it looks less strange than just doing that to the vertical separator?
Anyway, may be merge this PR for after 2019.11 (and so we can get more feedback about standalone #5554 ugliness :)

> My proposal to get rid of the separator lines still stands too. ;-)

And I prefer this full white one to these shorter vertical separators :)

Frenzie:
Not really, for example here's Firefox doing it (Android looks essentially the same):

Screenshot_20191110_205729

PR #5588, opened 2019-11-11
[feat] Support callback function in VirtualKeyPopup

Frenzie:
Required for <https://github.com/koreader/koreader/pull/5583>.

### DIFF
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -322,6 +322,7 @@ function VirtualKeyPopup:init()
     local parent_key = self.parent_key
     local key_chars = parent_key.key_chars
     local key_char_orig = key_chars[1]
+    local key_char_orig_func = parent_key.callback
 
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -315,27 +322,41 @@ function VirtualKeyPopup:init()
     local parent_key = self.parent_key
     local key_chars = parent_key.key_chars
     local key_char_orig = key_chars[1]
+    local key_char_orig_func = parent_key.callback
```
poire-z: What's that for?

Frenzie: The middle button responds like the original key so I figure it should do the same with a custom callback. Could also just be nil I figure.

poire-z: Oh, allright, I thought you were saving something before allowing/going to override it (that I often would name `something_orig_func` :)

PR #5596, opened 2019-11-15
Add Mark as read to end of document action

robert00s:

### DIFF
--- a/frontend/apps/reader/modules/readerstatus.lua
+++ b/frontend/apps/reader/modules/readerstatus.lua
@@ -89,6 +89,19 @@ function ReaderStatus:onEndOfBook()
             },
             {
                 {
+                    text_func = function()
+                        if self.settings.data.summary and self.settings.data.summary.status == "complete" then
+                            return _("Mark as reading")
+                        else
+                            return _("Mark as read")
+                        end
+                    end,
+                    callback = function()
+                        self:onMarkBook()
+                        UIManager:close(choose_action)
+                    end,
+                },
+                {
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerstatus.lua
```diff
@@ -122,6 +135,12 @@ function ReaderStatus:onEndOfBook()
         end
     elseif settings == "file_browser" then
         self:openFileBrowser()
+    elseif settings == "mark_read" then
+        self:onMarkBook(true)
+        UIManager:show(InfoMessage:new{
+            text = _("You've reached the end of the document.\nCurrent book mark as read."),
```
Frenzie:
```suggestion
            text = _("You've reached the end of the document.\nThe current book is marked as read."),
```

Frenzie:
I think it would make more sense to put `Cancel` in the bottom left. Since Mark as read feels related to book status, that'd basically be a simple swap around like this:

* Mark as read/Book status
* Delete file/Next file
* Cancel/File browser

Any thoughts @poire-z @NiLuJe ?

poire-z:
I agree with the layout you propose, seems logical to me too.

poire-z:
I agree with the layout @Frenzie proposes, seems logical to me too.

NiLuJe:
Yep, makes sense!

robert00s:
@Frenzie Thanks. It really make more sense.

Frenzie:
Thanks!

PR #5598, opened 2019-11-15
TextWidget/TextBoxWidget: enhanced text shaping

poire-z:
Have TextWdiget and TextBoxWidget use the new Lua C module (added in https://github.com/koreader/koreader-base/pull/1010) as an alternative text layout method, using Harfbuzz, FriBiDi and libunibreak.
This is needed to correctly shape Arabic scripts (cursive glyphs), propertly draw RTL text (Hebrew,
Arabic) and bidi text (arabic/hebrew filenames in english UI).
Screenshots and discussion in koreader/koreader#5359.

With this PR, we globally switch to use Harfbuzz for text shaping. It will feel like a change (text more condensed, I didn't like it initially, but switching back to the previous Freetype-only rendering, that old rendering really feels ugly :)

This should be enough to render Arabic text correctly. It still needs more work on the UI widgets and the language bootstrap code, and gettext/L to properly bidi isolate some bit of strings - to properly _align_ arabic and hebrew to the right.
(We'll also need more unrelated work to mirror the UI elements when the UI language is RTL - but this PR and a next one should have addressed all things related to text.)

For now, for testing, to enable RTL magic, add somewhere in reader.lua (this will have to be properly handled in language.lua)
```lua
require("libs/libkoreader-xtext")
xtext.setDefaultParaDirection(true)
xtext.setDefaultLang("ar")
-- xtext.setDefaultLang("zh_CN")
-- xtext.setDefaultLang("ja") -- get alt CJK glyphs
-- xtext.setDefaultLang("fr") -- get libunibreak french guillemets non-breaks
```

Also, I had to rework #5496 as it was using TextBoxWidget internals. So, I ended adding stuff to TextBoxWidget to do the work, and Menu.lua ended up quite shorter :) @robert00s : can you check last commit, and see if the code, with my comments, address correctly what you did with #5496.
I also took the liberty of adding a new setting `Items > Reduce font size to show more text` to bring back the previous behaviour that I prefer :) See https://github.com/koreader/koreader/pull/5496#issuecomment-542798582.

Some misc technical notes:
I initially went with touching as little TextWidget and TextBoxWidget, and add my stuff in xtextwidget.lua and xtextboxwidget.lua, and use something like:
```lua
function XTextBoxWidget:replaceTextBoxWidget()
    if TextBoxWidget._replaced_by_XTextBoxWidget then
        return
    end
    TextBoxWidget._replaced_by_XTextBoxWidget = true
    TextBoxWidget.initText = XTextBoxWidget.initText
    TextBoxWidget._splitToLines = XTextBoxWidget._splitToLines
    TextBoxWidget._renderText = XTextBoxWidget._renderText
    TextBoxWidget._shapeLine = XTextBoxWidget._shapeLine
    TextBoxWidget._getXYForCharPos = XTextBoxWidget._getXYForCharPos
    TextBoxWidget.getCharPosAtXY = XTextBoxWidget.getCharPosAtXY
    TextBoxWidget.onHoldReleaseText = XTextBoxWidget.onHoldReleaseText
    TextBoxWidget._alt_color_for_rtl = true -- for debugging
end
```
to enable it. It resulted in some redundant code (the code that deals with y coordinates), as this updates only the work on placing glyphs on the x asis. So, I ended up merging the new code into our TextWidget and TextBoxWidget, using a simple boolean to toggle one or the other (use_xtext = true/false).
For `TextBoxWidget:_splitToLines()`, I had to use a ugly `goto idx_continue` and `::idx_continue:: -- (Label for goto when use_xtext=true)` - just because I didn't want to indent the previous code and have it part of the diff/git blame. I'll get rid of it later by indenting the whole thing in a later commit if that's really too ugly :)

Haven't benchmarked that code. But on my device, it feels as smooth as before.
I know (from the code and library we use) this needs quite more processor cycles than before, so it feels a bit gratuitous as the previous rendering code was good enough. But I have no idea how much this added work compares to all the other CPU intensive work, like blitting blitbuffers, and if it is negligeable :)
If needed, we could optimize the other widgets to cache more their TextWidgets (our menu create new ones each time we switch menu or menu pages) - or use other cache strategies.

Also, we might need to work on our bold font use and handling. In many cases, we use the normal font, and provide `bold=true` which does synthetized bold (while we may have a real nicer bold font).
With XText/Harfbuzz, we use another Freetype emboldening API that don't change metrics, and we can tune the boldness to our liking. Also, as it doesn't change metrics, it feels different, and I quite like it that the metrics don't change, things get more aligned, eg:
Before (that `Chapter 6` sticks out from the other ones):
<kbd>image</kbd>
After with `bold_strength_factor = 3/8 -- as crengine, lighter`:
<kbd>image</kbd>
With `bold_strength_factor = 1/2 -- bold enough`:
<kbd>image</kbd>
With `bold_strength_factor = 1   -- really too bold`:
<kbd>image</kbd>

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -376,6 +376,20 @@ function FileManagerMenu:setUpdateItemTable()
             end,
         })
     end
+    table.insert(self.menu_items.developer_options.sub_item_table, {
+        text = _("Disable enhanced UI text shaping (xtext)"),
+        checked_func = function()
+            return G_reader_settings:isFalse("use_xtext")
+        end,
+        callback = function()
+            G_reader_settings:flipNilOrTrue("use_xtext")
+            local InfoMessage = require("ui/widget/infomessage")
+            UIManager:show(InfoMessage:new{
+                text = _("This will take effect on next restart."),
+            })
+        end,
+    })
+
     self.menu_items.cloud_storage = {
         text = _("Cloud storage"),
         callback = function()
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -179,6 +179,16 @@ function FileManagerMenu:setUpdateItemTable()
                     }
                     UIManager:show(items_font)
                 end
+            },
+            {
+                text = _("Reduce font size to show more text"),
+                keep_menu_open = true,
+                checked_func = function()
+                    return G_reader_settings:isTrue("items_multilines_show_more_text")
+                end,
+                callback = function()
+                    G_reader_settings:flipNilOrFalse("items_multilines_show_more_text")
+                end
             }
         }
     }
--- a/frontend/ui/font.lua
+++ b/frontend/ui/font.lua
@@ -79,9 +79,10 @@ local Font = {
     fallbacks = {
         [1] = "NotoSans-Regular.ttf",
         [2] = "NotoSansCJKsc-Regular.otf",
-        [3] = "nerdfonts/symbols.ttf",
-        [4] = "freefont/FreeSans.ttf",
-        [5] = "freefont/FreeSerif.ttf",
+        [3] = "NotoSansArabicUI-Regular.ttf",
+        [4] = "nerdfonts/symbols.ttf",
+        [5] = "freefont/FreeSans.ttf",
+        [6] = "freefont/FreeSerif.ttf",
     },
 
     -- face table
@@ -90,6 +91,49 @@ local Font = {
 
 if is_android then
     table.insert(Font.fallbacks, 3, "DroidSansFallback.ttf") -- for some ancient pre-4.4 Androids
+end
+
+-- Synthetized bold strength can be tuned:
+-- local bold_strength_factor = 1   -- really too bold
+-- local bold_strength_factor = 1/2 -- bold enough
+local bold_strength_factor = 3/8 -- as crengine, lighter
+
+-- Callback to be used by libkoreader-xtext.so to get Freetype
+-- instantiated fallback fonts when needed for shaping text
+local _getFallbackFont = function(face_obj, num)
+    if not num or num == 0 then -- return the main font
+        if not face_obj.embolden_half_strength then
+            -- cache this value in case we use bold, to avoid recomputation
+            face_obj.embolden_half_strength = face_obj.ftface:getEmboldenHalfStrength(bold_strength_factor)
+        end
+        return face_obj
+    end
+    if not face_obj.fallbacks then
+        face_obj.fallbacks = {}
+    end
+    if face_obj.fallbacks[num] ~= nil then
+        return face_obj.fallbacks[num]
+    end
+    local next_num = #face_obj.fallbacks + 1
+    local cur_num = 0
+    for index, fontname in pairs(Font.fallbacks) do
+        if fontname ~= face_obj.realname then -- Skip base one if among fallbacks
+            local fb_face = Font:getFace(fontname, face_obj.orig_size)
+            if fb_face ~= nil then -- valid font
+                cur_num = cur_num + 1
+                if cur_num == next_num then
+                    face_obj.fallbacks[next_num] = fb_face
+                    if not fb_face.embolden_half_strength then
+                        fb_face.embolden_half_strength = fb_face.ftface:getEmboldenHalfStrength(bold_strength_factor)
+                    end
+                    return fb_face
+                end
+            end
+        end
+    end
+    -- no more fallback font
+    face_obj.fallbacks[next_num] = false
+    return false
 end
 
 --- Gets font face object.
@@ -144,12 +188,34 @@ function Font:getFace(font, size)
         -- @field hash hash key for this font face
         face_obj = {
             orig_font = font,
+            realname = realname,
             size = size,
             orig_size = orig_size,
             ftface = face,
-            hash = hash
+            hash = hash,
         }
         self.faces[hash] = face_obj
+
+        -- Callback to be used by libkoreader-xtext.so to get Freetype
+        -- instantiated fallback fonts when needed for shaping text
+        face_obj.getFallbackFont = function(num)
+            return _getFallbackFont(face_obj, num)
+        end
+        -- Font features, to be given by libkoreader-xtext.so to HarfBuzz.
+        -- (Could be tweaked by font if needed. Note that NotoSans does not
+        -- have common ligatures, like for "fi" or "fl", so we won't see
+        -- them in the UI.)
+        -- Use HB defaults, and be sure to enable kerning and ligatures
+        -- (which might be part of HB defaults, or not, not sure).
+        face_obj.hb_features = { "+kern", "+liga" }
+        -- If we'd wanted to disable all features that might be enabled
+        -- by HarfBuzz (see harfbuzz/src/hb-ot-shape.cc, quite unclear
+        -- what's enabled or not by default):
+        -- face_obj.hb_features = {
+        --      "-kern", "-mark", "-mkmk", "-curs", "-locl", "-liga",
+        --      "-rlig", "-clig", "-ccmp", "-calt", "-rclt", "-rvrn",
+        --      "-ltra", "-ltrm", "-rtla", "-rtlm", "-frac", "-numr",
+        --      "-dnom", "-rand", "-trak", "-vert", }
     end
     return face_obj
 end
--- a/frontend/ui/widget/textboxwidget.lua
+++ b/frontend/ui/widget/textboxwidget.lua
@@ -823,7 +823,7 @@ function TextBoxWidget:getCharPosAtXY(x, y)
     if x > self.vertical_string_list[ln].width then -- no need to loop thru chars
         local pos = self.vertical_string_list[ln].end_offset
         if not pos then -- empty line
-            pos = self.vertical_string_list[ln].offset
+            return self.vertical_string_list[ln].offset
         end
         return pos + 1 -- after last char
     end
@@ -883,6 +883,12 @@ function TextBoxWidget:moveCursorToCharPos(charpos)
         self.virtual_line_num = self.virtual_line_num + scroll_lines
         -- needs to deal with possible overflow ?
         y = y - scroll_lines * self.line_height_px
+    end
+    -- We can also get x ouside current view, when a line takes the full width
+    -- (which happens when text is justified): move the cursor a bit to the left
+    -- (it will be drawn over the right of the last glyph, which should be ok.)
+    if x > self.width - self.cursor_line.dimen.w then
+        x = self.width - self.cursor_line.dimen.w
     end
     if not self._bb then
         return -- no bb yet to render the cursor too
--- a/frontend/ui/widget/textboxwidget.lua
+++ b/frontend/ui/widget/textboxwidget.lua
@@ -43,6 +43,8 @@ local TextBoxWidget = InputContainer:new{
     fgcolor = Blitbuffer.COLOR_BLACK,
     width = Screen:scaleBySize(400), -- in pixels
     height = nil, -- nil value indicates unscrollable text widget
+    height_adjust = false, -- if true, reduce height to a multiple of line_height (for nicer centering)
+    height_overflow_show_ellipsis = false, -- if height overflow, append ellipsis to last shown line
     top_line_num = nil, -- original virtual_line_num to scroll to
     charpos = nil, -- idx of char to draw the cursor on its left (can exceed #charlist by 1)
 
@@ -131,6 +133,15 @@ function TextBoxWidget:init()
         self.text_height = self.lines_per_page * self.line_height_px
         self.virtual_line_num = 1
     else
+        if self.height_overflow_show_ellipsis and #self.vertical_string_list > self.lines_per_page then
+            self.line_with_ellipsis = self.lines_per_page
+        end
+        if self.height_adjust then
+            self.height = self.text_height
+            if #self.vertical_string_list < self.lines_per_page then
+                self.height = #self.vertical_string_list * self.line_height_px
+            end
+        end
         -- Show the previous displayed area in case of re-init (focus/unfocus)
         -- InputText may have re-created us, while providing the previous charlist,
         -- charpos and top_line_num.
@@ -448,7 +459,8 @@ function TextBoxWidget:_shapeLine(line)
     -- Get glyphs, shaped and possibly substituted by Harfbuzz and re-ordered by FriBiDi.
     -- We'll add to 'line' this table of glyphs, with some additional
     -- computed x and advance keys
-    local xshaping = self._xtext:shapeLine(line.offset, line.end_offset)
+    local xshaping = self._xtext:shapeLine(line.offset, line.end_offset,
+                                            line.idx_to_substitute_with_ellipsis)
     -- logger.dbg(xshaping)
     -- We get an array of tables looking like this:
     --     [1] = {
@@ -594,6 +606,23 @@ function TextBoxWidget:_renderText(start_row_idx, end_row_idx)
     if self.use_xtext then
         for i = start_row_idx, end_row_idx do
             local line = self.vertical_string_list[i]
+            if self.line_with_ellipsis and i == self.line_with_ellipsis and not line.ellipsis_added then
+                -- Requested to add an ellipsis on this line
+                local ellipsis_width = RenderText:getEllipsisWidth(self.face)
+                    -- no bold: xtext does synthetized bold with normal metrics
+                if line.width + ellipsis_width > line.targeted_width then
+                    -- The ellipsis would overflow: we need to re-makeLine()
+                    -- this line with a smaller targeted_width
+                    line = self._xtext:makeLine(line.offset, line.targeted_width - ellipsis_width)
+                    self.vertical_string_list[i] = line -- replace the former one
+                end
+                if line.end_offset and line.end_offset < #self._xtext then
+                    -- We'll have shapeLine add the ellipsis to the returned glyphs
+                    line.end_offset = line.end_offset + 1
+                    line.idx_to_substitute_with_ellipsis = line.end_offset
+                end
+                line.ellipsis_added = true -- No need to redo it next time
+            end
             self:_shapeLine(line)
             if line.xglyphs then -- non-empty line
                 for __, xglyph in ipairs(line.xglyphs) do
@@ -628,7 +657,19 @@ function TextBoxWidget:_renderText(start_row_idx, end_row_idx)
         end
         -- Note: we use kerning=true in all RenderText calls
         -- (But kerning should probably not be used with monospaced fonts.)
-        RenderText:renderUtf8Text(self._bb, pen_x, y, self.face, self:_getLineText(line), true, self.bold, self.fgcolor, nil, self:_getLinePads(line))
+        local line_text = self:_getLineText(line)
+        if self.line_with_ellipsis and i == self.line_with_ellipsis then
+            -- Requested to add an ellipsis on this line
+            local ellipsis_width = RenderText:getEllipsisWidth(self.face, self.bold)
+            if line.width + ellipsis_width > self.width then
+                -- We could try to find the last break point (space, CJK) to
+                -- truncate there and add the ellipsis, but well...
+                line_text = RenderText:truncateTextByWidth(line_text, self.face, self.width, true, self.bold)
+            else
+                line_text = line_text .. "…"
+            end
+        end
+        RenderText:renderUtf8Text(self._bb, pen_x, y, self.face, line_text, true, self.bold, self.fgcolor, nil, self:_getLinePads(line))
         y = y + self.line_height_px
     end
 
@@ -811,6 +852,25 @@ function TextBoxWidget:getVisibleHeightRatios()
     local low = (self.virtual_line_num - 1) / #self.vertical_string_list
     local high = (self.virtual_line_num - 1 + self.lines_per_page) / #self.vertical_string_list
     return low, high
+end
+
+-- Helper function to be used before intanstiating a TextBoxWidget instance
+function TextBoxWidget:getFontSizeToFitHeight(height_px, nb_lines, line_height_em)
+    -- Get a font size that would fit nb_lines in height_px.
+    -- A font with the returned size should then be provided
+    -- to TextBoxWidget:new() (as well as the line_height_em given
+    -- here, as the line_height= property, if not the default).
+    if not nb_lines then
+        nb_lines = 1 -- default to 1 line
+    end
+    if not line_height_em then
+        line_height_em = self.line_height -- (TextBoxWidget default above: 0.3)
+    end
+    -- We do the revert of what's done in :init():
+    --   self.line_height_px = Math.round( (1 + self.line_height) * self.face.size )
+    local font_size = height_px / nb_lines / (1 + line_height_em)
+    font_size = font_size * 1000000 / Screen:scaleBySize(1000000) -- invert scaleBySize
+    return math.floor(font_size)
 end
 
 function TextBoxWidget:getCharPos()
--- a/frontend/ui/widget/textboxwidget.lua
+++ b/frontend/ui/widget/textboxwidget.lua
@@ -77,11 +77,23 @@ local TextBoxWidget = InputContainer:new{
     --                  when bb or hi_bb is nil: its job is to load/build bb or hi_bb.
     --                  The page will refresh itself when load_bb_func returns.
     images = nil, -- list of such images
-    line_num_to_image = nil, -- will be filled by self:_splitCharWidthList()
+    line_num_to_image = nil, -- will be filled by self:_splitToLines()
     image_padding_left = Screen:scaleBySize(10),
     image_padding_bottom = Screen:scaleBySize(3),
     image_alt_face = Font:getFace("xx_smallinfofont"),
     image_alt_fgcolor = Blitbuffer.COLOR_BLACK,
+
+    -- Additional properties only used when using xtext
+    use_xtext = G_reader_settings:nilOrTrue("use_xtext"),
+    lang = nil, -- use this language (string) instead of the UI language
+    para_direction_rtl = nil, -- use true/false to override the default direction for the UI language
+    auto_para_direction = false, -- detect direction of each paragraph in text
+                                 -- (para_direction_rtl or UI language is then only
+                                 -- used as a weak hint about direction)
+    alignment_strict = false, -- true to force the alignemnt set by the alignment= attribute.
+                              -- When false, specified alignment is inverted when para direction is RTL
+    _xtext = nil, -- for internal use
+    _alt_color_for_rtl = nil, -- (for debugging) draw LTR glyphs in black, RTL glyphs in gray
 }
 
 function TextBoxWidget:init()
@@ -98,12 +110,18 @@ function TextBoxWidget:init()
         if self.height < self.line_height_px then
             self.height = self.line_height_px
         end
-        -- if no self.height, these will be set just after self:_splitCharWidthList()
+        -- if no self.height, these will be set just after self:_splitToLines()
         self.lines_per_page = math.floor(self.height / self.line_height_px)
         self.text_height = self.lines_per_page * self.line_height_px
     end
-    self:_evalCharWidthList()
-    self:_splitCharWidthList()
+
+    if self.use_xtext then
+        self:_measureWithXText()
+    else
+        self:_evalCharWidthList()
+    end
+    self:_splitToLines()
+
     if self.charpos and self.charpos > #self.charlist+1 then
         self.charpos = #self.charlist+1
     end
@@ -168,8 +186,32 @@ function TextBoxWidget:_evalCharWidthList()
     self.idx_pad = {}
 end
 
+function TextBoxWidget:_measureWithXText()
+    if not self._xtext_loaded then
+        require("libs/libkoreader-xtext")
+        TextBoxWidget._xtext_loaded = true
+    end
+    if type(self.charlist) == "table" then
+        self._xtext = xtext.new(self.charlist, self.face, self.auto_para_direction,
+                                                self.para_direction_rtl, self.lang)
+    else
+        if not self.text then
+            self.text = ""
+        elseif type(self.text) ~= "string" then
+            self.text = tostring(self.text)
+        end
+        self._xtext = xtext.new(self.text, self.face, self.auto_para_direction,
+                                                self.para_direction_rtl, self.lang)
+        self.charlist = self._xtext
+            -- Just to have many common bits of code using #self.charlist work
+            -- as expected (will crash if used as a real table with "table
+            -- expected, got userdata", so we know)
+    end
+    self._xtext:measure()
+end
+
 -- Split the text into logical lines to fit into the text box.
-function TextBoxWidget:_splitCharWidthList()
+function TextBoxWidget:_splitToLines()
     self.vertical_string_list = {}
 
     local idx = 1
@@ -180,7 +222,7 @@ function TextBoxWidget:_splitCharWidthList()
     local image_num = 0
     local targeted_width = self.width
     local image_lines_remaining = 0
-    while idx <= size do
+    while idx and idx <= size do
         -- Every scrolled page, we want to add the next (if any) image at its top right
         -- (if not scrollable, we will display only the first image)
         -- We need to make shorter lines and leave room for the image
@@ -218,6 +260,45 @@ function TextBoxWidget:_splitCharWidthList()
 
         -- end_offset will be the idx of char at end of line
         offset = idx -- idx of char at start of line
+
+        if self.use_xtext then
+            -- All of what's done below when use_xtext=false is done by the C++ module.
+            local line = self._xtext:makeLine(offset, targeted_width)
+            -- logger.dbg("makeLine", ln, line)
+            -- We get a line such as this:
+            --    {
+            --        ["next_start_offset"] = 9272,
+            --        ["width"] = 511,
+            --        ["end_offset"] = 9270,
+            --        ["targeted_width"] = 548,
+            --        ["offset"] = 9208,
+            --        ["can_be_justified"] = true
+            --    },
+            -- Notes:
+            -- - next_start_offset is nil when reaching end of text
+            -- - On empty lines made from a standalone \n\n, we get end_offset = offset-1,
+            --   which is a bit strange but that's what the use_xtext=false does.
+            -- - Between a line end_offset= and the next line offset=, there may be only
+            --   a single indice not included: the \n or the space that allowed the break.
+            self.vertical_string_list[ln] = line
+            if line.hard_newline_at_eot and not line.next_start_offset then
+                -- Add an empty line to reprensent the \n at end of text
+                -- and allow positionning cursor after it
+                self.vertical_string_list[ln+1] = {
+                    offset = size+1,
+                    end_offset = nil,
+                    width = 0,
+                }
+            end
+            ln = ln + 1
+            idx = line.next_start_offset -- nil when end of text reached
+            -- Skip the whole following non-use_xtext code, to continue
+            -- this 'while' loop (to avoid indentation diff on the
+            -- following code if we were using a 'else'...)
+            goto idx_continue
+        end
+
+        -- Only when not self.use_xtext:
 
         -- We append chars until the accumulated width exceeds `targeted_width`,
         -- or a newline occurs, or no more chars to consume.
@@ -333,6 +414,8 @@ function TextBoxWidget:_splitCharWidthList()
         end
         ln = ln + 1
         -- Make sure `idx` point to the next char to be processed in the next loop.
+
+        ::idx_continue:: -- (Label for goto when use_xtext=true)
     end
 end
 
@@ -348,6 +431,146 @@ function TextBoxWidget:_getLinePads(vertical_string)
         table.insert(pads, self.idx_pad[idx] or 0)
     end
     return pads
+end
+
+-- XText: shape a line into positionned glyphs
+function TextBoxWidget:_shapeLine(line)
+    -- line is an item from self.vertical_string_list
+    if not line.end_offset then
+        return -- empty line (hard newline at end of file)
+    end
+    if line.end_offset < line.offset then
+        return -- empty line (hard newline while not at end of file)
+    end
+    if line.xglyphs then
+        return -- already done
+    end
+    -- Get glyphs, shaped and possibly substituted by Harfbuzz and re-ordered by FriBiDi.
+    -- We'll add to 'line' this table of glyphs, with some additional
+    -- computed x and advance keys
+    local xshaping = self._xtext:shapeLine(line.offset, line.end_offset)
+    -- logger.dbg(xshaping)
+    -- We get an array of tables looking like this:
+    --     [1] = {
+    --         ["y_offset"] = 0,
+    --         ["x_advance"] = 10,
+    --         ["can_extend"] = false,
+    --         ["can_extend_fallback"] = false,
+    --         ["is_rtl"] = false,
+    --         ["text_index"] = 1,
+    --         ["glyph"] = 68,
+    --         ["font_num"] = 0,
+    --         ["x_offset"] = 0,
+    --         ["is_cluster_start"] = true,
+    --         ["cluster_len"] = 1
+    --     },
+    --     [...]
+    --     [12] = {
+    --         ["y_offset"] = 0,
+    --         ["x_advance"] = 0,
+    --         ["can_extend"] = false,
+    --         ["can_extend_fallback"] = false,
+    --         ["is_rtl"] = true,
+    --         ["text_index"] = 8,
+    --         ["glyph"] = 1292,
+    --         ["font_num"] = 3,
+    --         ["x_offset"] = -2,
+    --         ["is_cluster_start"] = true,
+    --         ["cluster_len"] = 2
+    --     },
+    --     [13] = {
+    --         ["y_offset"] = 0,
+    --         ["x_advance"] = 10,
+    --         ["can_extend"] = false,
+    --         ["can_extend_fallback"] = false,
+    --         ["is_rtl"] = true,
+    --         ["text_index"] = 8,
+    --         ["glyph"] = 1321,
+    --         ["font_num"] = 3,
+    --         ["x_offset"] = 0,
+    --         ["is_cluster_start"] = false,
+    --         ["cluster_len"] = 2
+    --     },
+    -- With some additional keys about the line itself, that will help
+    -- with alignment and justification:
+    --     ["para_is_rtl"] = true,
+    --     ["nb_can_extend"] = 6,
+    --     ["nb_can_extend_fallback"] = 0,
+    --     ["width"] = 457
+
+    local alignment = self.alignment
+    if not self.alignment_strict and xshaping.para_is_rtl then
+        if alignment == "left" then
+            alignment = "right"
+        elseif alignment == "right" then
+            alignment = "left"
+        end
+    end
+
+    local pen_x = 0 -- when alignment == "left"
+    if alignment == "center" then
+        pen_x = (line.targeted_width - line.width)/2 or 0
+    elseif alignment == "right" then
+        pen_x = (line.targeted_width - line.width)
+    end
+
+    local space_add_w = 0
+    local space_add1_nb = 0
+    local use_can_extend_fallback = false
+    if self.justified and line.can_be_justified then
+        local space_to_fill = line.targeted_width - xshaping.width
+        if xshaping.nb_can_extend > 0 then
+            space_add_w = math.floor(space_to_fill / xshaping.nb_can_extend)
+            -- nb of spaces to which we'll add 1 more pixel
+            space_add1_nb = space_to_fill - space_add_w * xshaping.nb_can_extend
+            line.justified = true
+            line.width = line.targeted_width
+            pen_x = 0 -- reset alignment
+        elseif xshaping.nb_can_extend_fallback > 0 then
+            use_can_extend_fallback = true
+            space_add_w = math.floor(space_to_fill / xshaping.nb_can_extend_fallback)
+            -- nb of spaces to which we'll add 1 more pixel
+            space_add1_nb = space_to_fill - space_add_w * xshaping.nb_can_extend_fallback
+            line.justified = true
+            line.width = line.targeted_width
+            pen_x = 0 -- reset alignment
+        end
+    end
+
+    local prev_cluster_start_xglyph
+    for i, xglyph in ipairs(xshaping) do
+        xglyph.x0 = pen_x
+        pen_x = pen_x + xglyph.x_advance -- advance from Harfbuzz
+        if xglyph.can_extend or (use_can_extend_fallback and xglyph.can_extend_fallback) then
+            -- add some pixels for justification
+            pen_x = pen_x + space_add_w
+            if space_add1_nb > 0 then
+                pen_x = pen_x + 1
+                space_add1_nb = space_add1_nb - 1
+            end
+        end
+        -- These will be used by _getXYForCharPos() and getCharPosAtXY():
+        xglyph.x1 = pen_x
+        xglyph.w = xglyph.x1 - xglyph.x0
+        -- Because of glyph substitution and merging (one to many, many to one, many to many,
+        -- with advance or zero-advance...), glyphs may not always be fine to position
+        -- the cursor caret. For X/Y/Charpos positionning/guessing, we'll ignore
+        -- glyphs that are not cluster_start, and we build here the full cluster x0/x1/w
+        -- by mergin them from all glyphs part of this cluster
+        if xglyph.is_cluster_start then
+            prev_cluster_start_xglyph = xglyph
+        else
+            if xglyph.x1 > prev_cluster_start_xglyph.x1 then
+                prev_cluster_start_xglyph.x1 = xglyph.x1
+                prev_cluster_start_xglyph.w = prev_cluster_start_xglyph.x1 - prev_cluster_start_xglyph.x0
+            end
+            -- We don't update/decrease prev_cluster_start_xglyph.x0, even if one of its glyph
+            -- has a backward advance that go back the 1st glyph x0, to not mess positionning.
+        end
+    end
+    line.xglyphs = xshaping
+    --- @todo Should we drop these when no more displayed in the page to reclaim memory,
+    -- at the expense of recomputing it when back to this page?
 end
 
 ---- Lays out text.
@@ -367,6 +590,34 @@ function TextBoxWidget:_renderText(start_row_idx, end_row_idx)
     self._bb = Blitbuffer.new(self.width, h, bbtype)
     self._bb:fill(Blitbuffer.COLOR_WHITE)
     local y = font_height
+
+    if self.use_xtext then
+        for i = start_row_idx, end_row_idx do
+            local line = self.vertical_string_list[i]
+            self:_shapeLine(line)
+            if line.xglyphs then -- non-empty line
+                for __, xglyph in ipairs(line.xglyphs) do
+                    local face = self.face.getFallbackFont(xglyph.font_num) -- callback (not a method)
+                    local glyph = RenderText:getGlyphByIndex(face, xglyph.glyph, self.bold)
+                    local color = self.fgcolor
+                    if self._alt_color_for_rtl then
+                        color = xglyph.is_rtl and Blitbuffer.COLOR_DARK_GRAY or Blitbuffer.COLOR_BLACK
+                    end
+                    self._bb:colorblitFrom(glyph.bb,
+                                xglyph.x0 + glyph.l + xglyph.x_offset,
+                                y - glyph.t + xglyph.y_offset,
+                                0, 0, glyph.bb:getWidth(), glyph.bb:getHeight(), color)
+                end
+            end
+            y = y + self.line_height_px
+        end
+        -- Render image if any
+        self:_renderImage(start_row_idx)
+        return
+    end
+
+    -- Only when not self.use_xtext:
+
     for i = start_row_idx, end_row_idx do
         local line = self.vertical_string_list[i]
         local pen_x = 0 -- when alignment == "left"
@@ -375,8 +626,8 @@ function TextBoxWidget:_renderText(start_row_idx, end_row_idx)
         elseif self.alignment == "right" then
             pen_x = (self.width - line.width)
         end
-            --- @todo don't use kerning for monospaced fonts.    (houqp)
-            --- refer to [cb25029dddc42693cc7aaefbe47e9bd3b7e1a750](https://github.com/koreader/koreader/commit/cb25029dddc42693cc7aaefbe47e9bd3b7e1a750) in master tree
+        -- Note: we use kerning=true in all RenderText calls
+        -- (But kerning should probably not be used with monospaced fonts.)
         RenderText:renderUtf8Text(self._bb, pen_x, y, self.face, self:_getLineText(line), true, self.bold, self.fgcolor, nil, self:_getLinePads(line))
         y = y + self.line_height_px
     end
@@ -578,6 +829,19 @@ end
 function TextBoxWidget:paintTo(bb, x, y)
     self.dimen.x, self.dimen.y = x, y
     bb:blitFrom(self._bb, x, y, 0, 0, self.width, self._bb:getHeight())
+end
+
+function TextBoxWidget:onCloseWidget()
+    -- Free all resources (BlitBuffers, XText) when UIManager closes this
+    -- widget (as it won't be painted anymore), without waiting for Lua gc()
+    -- to kick in.
+    -- Free the between-renderings (between page scrolls) freeable resources
+    self:free()
+    if self.use_xtext and self._xtext then
+        -- Free our self._xtext (that we can't free in :free()
+        -- as we need to keep it across renderings)
+        self._xtext:free()
+    end
 end
 
 function TextBoxWidget:free()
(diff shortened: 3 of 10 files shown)


poire-z:
Will need a base bump for https://github.com/koreader/koreader-base/pull/1010 in 3rd commit.

frontend/ui/font.lua
```diff
@@ -144,12 +188,31 @@ function Font:getFace(font, size)
...
+        -- Callback to be used by libkoreader-xtext.so to get Freetype
+        -- instantiated fallback fonts when needed for shaping text
+        face_obj.getFallbackFont = function(num)
+            return _getFallbackFont(face_obj, num)
+        end
+        -- Font features, to be given by libkoreader-xtext.so to HarfBuzz.
+        -- (Could be tweaked by font if needed. Note that NotoSans does not
+        -- have common ligatures, like for "fi" or "fl", so we won't see
+        -- them in the UI.)
+        face_obj.hb_features = { "+kern", "+liga" }
```
Frenzie: So is this technically redundant with HB defaults? I'm slightly confused by the comment following this line.

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -376,6 +376,20 @@ function FileManagerMenu:setUpdateItemTable()
             end,
         })
     end
+    table.insert(self.menu_items.developer_options.sub_item_table, {
+        text = _("Disable UI enhanced text shaping (xtext)"),
```
Frenzie: The order sounds a bit weird to me.

Disable enhanced UI text shaping

frontend/ui/widget/textwidget.lua
```diff
@@ -46,14 +57,44 @@ function TextWidget:updateSize()
...
+    --   self._height = math.ceil(self.face.size * 1.5)
+    --   self._baseline_h = self._height*0.7
+    -- But better compute baseline alignment from freetype font metrics
+    -- to get better vertical centering of text in box
+    -- (Freetype doc on this at https://www.freetype.org/freetype2/docs/tutorial/step2.html)
+    local face_height, face_ascender = self.face.ftface:getHeightAndAscender()
+    self._height = math.ceil(face_height) + 2*self.padding
+    self._baseline_h = Math.round(face_ascender) + self.padding
+    -- With our UI fonts, this usually gives 0.72 to 0.74, so text is aligned
+    -- a bit lower than before with the hardcoded 0.7
```
Frenzie: Excellent!

frontend/ui/widget/textwidget.lua
```diff
@@ -46,14 +57,44 @@ function TextWidget:updateSize()
...
+
     -- In case we draw truncated text, keep original self.text
     -- so caller can fetch it again
     self._text_to_draw = self.text
 
+    if self.text and type(self.text) ~= "string" then
```
Frenzie: When does this happen?

frontend/ui/widget/textwidget.lua
```diff
@@ -46,14 +57,44 @@ function TextWidget:updateSize()
...
+    -- Compute width:
+    if self.use_xtext then
+        self:_measureWithXText()
+        return
+    end
+
+    -- Only when not self.use_xtext:
+
     -- Note: we use kerning=true in all RenderText calls
     --- @todo Don't use kerning for monospaced fonts. (houqp)
```
Frenzie: This still relevant?

frontend/ui/widget/textboxwidget.lua
```diff
@@ -367,6 +590,34 @@ function TextBoxWidget:_renderText(start_row_idx, end_row_idx)
...
+    if self.use_xtext then
+        for i = start_row_idx, end_row_idx do
+            local line = self.vertical_string_list[i]
+            self:_shapeLine(line)
+            if line.xglyphs then -- non-empty line
+                for i, xglyph in ipairs(line.xglyphs) do
+                    local face = self.face.getFallbackFont(xglyph.font_num) -- callback (not a method)
+                    local glyph = RenderText:getGlyphByIndex(face, xglyph.glyph, self.bold)
+                    local color = self.fgcolor
+                    if self._alt_color_for_rtl then
```
Frenzie: So this is always disabled unless you change it in the code?

Frenzie: (Might be of some interest in that developer settings menu.)

frontend/ui/widget/textboxwidget.lua
```diff
@@ -580,6 +831,17 @@ function TextBoxWidget:paintTo(bb, x, y)
     bb:blitFrom(self._bb, x, y, 0, 0, self.width, self._bb:getHeight())
 end
 
+function TextBoxWidget:onCloseWidget()
+    -- free when UIManager:close() was called
```
Frenzie: Free when UIManager:close() was called.

(I mean, it just looks so out of place with the next comment. :-P)

frontend/ui/widget/textboxwidget.lua
```diff
@@ -580,6 +831,17 @@ function TextBoxWidget:paintTo(bb, x, y)
...
+    -- Free the between-renderings (between page scrolls) freeable ressources
```
Frenzie: ressources is French ^_^

frontend/ui/widget/textwidget.lua
poire-z: (Nothing excellent :) that's some previous comment that I just moved - it appears added here, but it's the same that is deleted below :)

Frenzie: Oh right, I remember now. :man_facepalming: 

Still excellent though. ;-D

frontend/ui/font.lua
poire-z: Half :) I've indeed been unclear, but that's mostly just cut and pasted from crengine.
And what the HB defaults are are also hard to track. It's in its `src/hb-ot-shape.cc`.
So, it looks to me `+liga` is the default, but not `+kern` (or may be it is...).
I'm just setting here the defaults we use in crengine HB full/best, even if one of them is already the default.
The commented next line is the list of all stuff mentionned in HB "good", that I just cut and pasted with all disabled to see if enabling +liga / +kern made a difference :)
I don't plan on being the reference as I'm not sure what it is :|. I'll update the comment to not state anything about "default" but "If we'd wanted to disable all the features HB may enable" :)

frontend/ui/widget/textwidget.lua
poire-z: Should not, but can if we provide a number or else.
We'll see it displayed instead of crashing, and it simplify my next check:
`if not self.text or #self.text == 0 then self._is_empty = true self._length = 0`
It avoid having `nil` fixed/handled in lower level code, like in:
https://github.com/koreader/koreader/blob/e166a77eba30e84cd72a9854b289bd9466ecc200/frontend/ui/rendertext.lua#L210-L214
https://github.com/koreader/koreader/blob/e166a77eba30e84cd72a9854b289bd9466ecc200/frontend/util.lua#L311-L313

frontend/ui/widget/textwidget.lua
poire-z: Not with the use_xtext code (Haven't checked, but I hope HB does the right thing with monospace fonts.
But for the legacy code, it might (it would still use kerning with monospaced fonts).

frontend/ui/widget/textboxwidget.lua
poire-z: > So this is always disabled unless you change it in the code?

Right, always disabled.
Not worth busying developper settings with it I think - was useful when doing this PR, but I guess RTL is alright now, and I shouldn't need it - unless we feel it's needed when we play with mirroring the UI, but I doubt it.

frontend/ui/widget/textboxwidget.lua
poire-z: Yes :) That's verbatim from the other widgets that have it, so, I just kept it for coherency even if it sounds gramatically strange.
And... it's I that introduced that strange wording :)

frontend/ui/widget/textboxwidget.lua
poire-z: French ressources need freeing too ! :)

frontend/ui/widget/textwidget.lua
Frenzie: Would something this work in this context? (That'd go underneath the method.) It's a convenient way to ensure conformity without bothering users.
```lua
dbg:guard(TextWidget, "updateSize",
    function()
        assert(type(self.text) == "string",
            "Wrong text type (expected string)")
    end)
```

frontend/ui/widget/textwidget.lua
poire-z: OK, I'll remove the `@todo` and keep that comment in parenthesis. As we'll probably don't have to fix it if we keep using xtext.

frontend/ui/widget/textwidget.lua
poire-z: I can, it works as you expect (crash on the emulator). Same in setText(text) then.
I think I've seen some output from `logger.warn("renderUtf8Text called without text")` in the past, so it may happen we get a nil. It's ok if we crash then when that happens?
(Can't cause any crash while tapping around - but I'm sure we'll get some.)

Frenzie: Assuming it only crashes when it comes across a problem so that the problem can be addressed, that's indeed the idea. ;-)

But if it's too big an issue atm then it'd probably be best to leave that for a follow-up PR.

NiLuJe:
Yeah, I'd probably +1 using a real Bold font (which we already ship anyway ^^) ;).

frontend/ui/widget/textwidget.lua
poire-z: I've already added and tested it, so it will go :)
If you confirm dbg.guard() does not kick in nor add overhead on our devices, it's fine with me - as long as I get there the safety of my tostring() (othewise, XText Lua check will throw an error).

frontend/ui/widget/textboxwidget.lua
```diff
@@ -562,6 +854,25 @@ function TextBoxWidget:getVisibleHeightRatios()
...
+    if not nb_lines then
+        nb_lines = 1 -- default to 1 line
+    end
+    if not line_height_em then
+        line_height_em = self.line_height -- (TextBoxWidget default above: 0.3)
+    end
+    -- We do the revert of what's done in :init():
+    --   self.line_height_px = Math.round( (1 + self.line_height) * self.face.size )
+    local font_size = height_px / nb_lines / (1 + line_height_em)
+    font_size = font_size * 1000000 / Screen:scaleBySize(1000000) -- invert scaleBySize
```
NiLuJe: Doing the proper math in a real matching function in ffi/framebuffer.lua might be worth it?

(i.e., having a real px to pt helper function available).

frontend/ui/widget/textwidget.lua
NiLuJe: Yeah, we possibly have a few `nil` `renderUtf8Text` lying around, that does ring a bell ;).

frontend/ui/widget/textboxwidget.lua
poire-z: Probably, we used that same trick in coverbrowser some weeks ago.
(But I suck at proper math :)

NiLuJe: Apparently as simple as basically copying `scaleBySize` but changing the return to `math.floor(px / (size_scale + dpi_scale) * 2)` ;).

poire-z: So, should it be that (scaleBySize() input being pt and not px)?
```diff
-function fb:scaleBySize(px)
+function fb:scaleBySize(pt)
     -- larger screen needs larger scale
@@ -299,3 +299,15 @@ function fb:scaleBySize(px)
     -- scaled positive px should also be positive
-    return math.ceil(px * (size_scale + dpi_scale) / 2)
+    return math.ceil(pt * (size_scale + dpi_scale) / 2)
+end
+
+function fb:unscaleBySize(px)
+    -- Revert of previous function:
+    -- px = scaleBySize(pt)
+    -- pt = unscaleBySize(px)
+    local size_scale = math.min(self:getWidth(), self:getHeight())/600
+    local dpi_scale = size_scale
+    if self.device and self.device.display_dpi ~= self.dpi then
+        dpi_scale = self.dpi / 160
+    end
+    return math.floor(px / (size_scale + dpi_scale) * 2)
 end
```

NiLuJe: Well, it's not technically points per-se, as that's pretty much set in stone as `px = round(dpi / 72.0 * pt)`, which isn't what we're doing.

It's more something like absolute pixels or whatever (I only mentioned point earlier because the theory is similar ;)).

poire-z: OK, so keeping using px in scaleBySize (pt means portuguese to me anyway:).

frontend/ui/widget/textwidget.lua
Frenzie: No overhead, when debug is disabled it's just nothing:

https://github.com/koreader/koreader/blob/b21b4c8776393f57f2c8fc15f06d7b90b952556f/frontend/dbg.lua#L65

And when enabled:

https://github.com/koreader/koreader/blob/b21b4c8776393f57f2c8fc15f06d7b90b952556f/frontend/dbg.lua#L37-L49

frontend/ui/widget/textboxwidget.lua
poire-z: @Frenzie: please confirm you're ok with that, before I add it to my base PR and trigger the whole CI:
```lua
function fb:unscaleBySize(px)
    -- Invert of fb:scaleBySize(px)
    local size_scale = math.min(self:getWidth(), self:getHeight())/600
    local dpi_scale = size_scale
    if self.device and self.device.display_dpi ~= self.dpi then
        dpi_scale = self.dpi / 160
    end
    return math.floor(px / (size_scale + dpi_scale) * 2)
end
```

Frenzie: Link to the originals for convenience:

https://github.com/koreader/koreader-base/blob/c140b752e16163ee8836214787caf7343bc4a579/ffi/framebuffer.lua#L285-L301

It looks sane enough at a glance, so I'm okay with it provided you checked it actually outputs the right values. ;-)

Frenzie: Actually you can do that by simply adding an extra line here:

https://github.com/koreader/koreader-base/blob/c140b752e16163ee8836214787caf7343bc4a579/spec/unit/framebuffer_spec.lua#L23-L29

poire-z: Well... given the rounding we do, unit tests fails...
```
        fb:setDPI(167)
        assert.are.equals(20, fb:scaleBySize(20 * 1000000 / fb:scaleBySize(1000000)))
        assert.are.equals(20, fb:scaleBySize(fb:unscaleBySize(20)))
        fb:setDPI(167 * 3)
        assert.are.equals(18, fb:scaleBySize(18 * 1000000 / fb:scaleBySize(1000000)))
        assert.are.equals(18, fb:scaleBySize(fb:unscaleBySize(18)))
```
whether I use floor or ceil or round, one of these fail. The non rounding 1000000 stuff doesn't fail.
So:
```
    -- Because of possible rounding error, as we only pass and
    -- return integers, scaleBySize(unscaleBySize(N)) != N
    -- In the context this is used, we prefer getting a smaller
    -- one. So use math.floor()
    return math.floor(px / (size_scale + dpi_scale) * 2)
```
?
Or return a float, and let callers do the rounding they want?
And how to unit-test that if we can't check equality? `> 0.9*N and < N` ?

(And please, stop bringing me into your bidi math games! :)

poire-z: Well, no frankly, not my war.
That api would look suspicious, and the unit tests even more fuzzy - so better to not have it, and let callers do they fuzzy math themselves.

Frenzie: You didn't mention how they fail precisely. You could just do an imprecise test, like:

```lua
local unscaled_size = fb:scaleBySize(fb:unscaleBySize(18))
assert.is_true(unscaled_size > 17 && unscaled_size < 19)
```

Or given that we control the test, something like `assert.are.equals(19, fb:scaleBySize(fb:unscaleBySize(18)))` would also work. The point of the unit test is primarily to ensure we still get the result we expect in the future.

poire-z: Yeah, 19 vs 20, stuff like that. And may be more than +/-1 with a high dpi value.
But this function is ridiculous if it's unprecise. And if the caller wants it rounded up, it would need to be some twins unscaleBySizeRoundedUp/unscaleBySizeRoundedDown.
I get more precise results with the 1000000 stuff. May be it's not needed by the 2 callers (textwidget/menu, & coverbrowser), but I don't want to investigate if they would be ok with floor or a little lower, or if they would prefer ceil and a little higher- as it's all just fine and tested with precise 1000000 stuff.

poire-z:
I have not spend much time on the cursor positionning in RTL/bidi contexts. So, this will need some more work. See `@todo` in `TextBoxWidget:_getXYForCharPos()`.

frontend/ui/widget/textboxwidget.lua
NiLuJe: Oh, yeah, round trips are going to be hard to keep stable without preserving the precise value (i.e., the float).

I'm not quite sure by which arcane trickery this currently appears to be the case with the integer trickery.

poire-z: > I'm not quite sure by which arcane tricke

Because I lied a bit and did not do any rounding in `18 * 1000000 / fb:scaleBySize(1000000)` may be :) But still.

NiLuJe: That said, I'm slightly miffed by the fact that it appears to behave in Python...

```py
#!/usr/bin/env python3

import math

dpi = 167
width = 600
height = 800

def scaleBySize(px):
   # larger screen needs larger scale
   size_scale = min(width, height)/600.0
   # if users custom screen dpi, also scale by dpi
   dpi_scale = size_scale

   dpi_scale = dpi / 160.0

   # scaled positive px should also be positive
   scaled_px = px * (size_scale + dpi_scale) / 2.0
   return (scaled_px, math.ceil(scaled_px))

def unScaleBySize(px):
   # larger screen needs larger scale
   size_scale = min(width, height)/600.0
   # if users custom screen dpi, also scale by dpi
   dpi_scale = size_scale

   dpi_scale = dpi / 160.0

   # scaled positive px should also be positive
   unscaled_px = px / (size_scale + dpi_scale) * 2.0
   return (unscaled_px, math.floor(unscaled_px))

print("scaleBySize(500): {}".format(scaleBySize(500.0)))
print("unScaleBySize(511): {}".format(unScaleBySize(511.0)))

print("scaleBySize(20): {}".format(scaleBySize(20.0)))
print("unScaleBySize(21): {}".format(unScaleBySize(21.0)))

dpi = 167 * 3

print("scaleBySize(18): {}".format(scaleBySize(18.0)))
print("unScaleBySize(38): {}".format(unScaleBySize(38.0)))

print("18 via * 10**6: {}".format(scaleBySize(18 * 1000000.0 / scaleBySize(1000000.0)[1])))
```

```
scaleBySize(500): (510.93750000000006, 511)
unScaleBySize(511): (500.06116207951067, 500)
scaleBySize(20): (20.4375, 21)
unScaleBySize(21): (20.550458715596328, 20)
scaleBySize(18): (37.18125, 38)
unScaleBySize(38): (18.3963691376702, 18)
18 via * 10**6: (17.999999999999996, 18)
```

The emulator defaults to 600x800 @ 167 dpi, right?

NiLuJe: > 
> The emulator (and as such the testsuite) defaults to 600x800 @ 167 dpi, right?

Possibly not, otherwise scaleBySize would do nothing (as min(600, 800)/600 == 1), and if dpi == device_dpi that's the scale factor).

poire-z: Looks like this test does not  have any screen h/w, so possibly only using dpi_scale.

Or we can just have that unscaleBySize return a float, and document that the caller is responsible to do a floor or ceil depending on what he prefers.
That way, the unit tests would be always fine: `assert.are.equals(18, fb:scaleBySize(fb:unscaleBySize(18)))`

NiLuJe: Not a Python quirk ;).

```lua
#!/usr/bin/env luajit

dpi = 167
width = 600
height = 800

function scaleBySize(px)
   -- larger screen needs larger scale
   size_scale = math.min(width, height)/600
   -- if users custom screen dpi, also scale by dpi
   dpi_scale = size_scale

   dpi_scale = dpi / 160

   -- scaled positive px should also be positive
   scaled_px = px * (size_scale + dpi_scale) / 2
   return {scaled_px, math.ceil(scaled_px)}
end

function unScaleBySize(px)
   -- larger screen needs larger scale
   size_scale = math.min(width, height)/600
   -- if users custom screen dpi, also scale by dpi
   dpi_scale = size_scale

   dpi_scale = dpi / 160

   -- scaled positive px should also be positive
   unscaled_px = px / (size_scale + dpi_scale) * 2
   return {unscaled_px, math.floor(unscaled_px)}
end

---

function tprint(tbl, indent)
  if not indent then indent = 0 end
  for k, v in pairs(tbl) do
    formatting = string.rep("  ", indent) .. k .. ": "
    if type(v) == "table" then
      print(formatting)
      tprint(v, indent+1)
    elseif type(v) == 'boolean' then
      print(formatting .. tostring(v))
    else
      print(formatting .. v)
    end
  end
end

print("scaleBySize(500)")
tprint(scaleBySize(500))
print("unScaleBySize(511)")
tprint(unScaleBySize(511))

print("scaleBySize(20)")
tprint(scaleBySize(20))
print("unScaleBySize(21)")
tprint(unScaleBySize(21))

dpi = 167 * 3

print("scaleBySize(18)")
tprint(scaleBySize(18))
print("unScaleBySize(38)")
tprint(unScaleBySize(38))

print("18 via * 10**6")
tprint(scaleBySize(18 * 1000000 / scaleBySize(1000000)[2]))
```

```
scaleBySize(500)
1: 510.9375
2: 511
unScaleBySize(511)
1: 500.06116207951
2: 500
scaleBySize(20)
1: 20.4375
2: 21
unScaleBySize(21)
1: 20.550458715596
2: 20
scaleBySize(18)
1: 37.18125
2: 38
unScaleBySize(38)
1: 18.39636913767
2: 18
18 via * 10**6
1: 18
2: 18
```

NiLuJe: TL;DR: Smells like yet another case of "I don't know what the hell's happening inside the testsuite" ;).

NiLuJe: (GH is still being clunky, I've only just now got your previous answers interleaved in the thread -_-").

Frenzie: GH has recently been "helpfully" hiding comments behind a "More comments" type thing for me. D:

The presumed default *used* to be 800x600 @ 167 but that was changed to 160 for easier math. See https://github.com/koreader/koreader-base/commit/c978857658fd5343f2fe90534c0d998cdead9aa8

Otherwise the unit test would make no sense! ;-)

```lua
        fb:setDPI(167)
        assert.are.equals(31, fb:scaleBySize(30))

        fb:setDPI(167 * 3)
        assert.are.equals(62, fb:scaleBySize(30))
```

Frenzie: Or at least, I supported it for that reason: https://github.com/koreader/koreader-base/pull/843#issuecomment-469597361

Frenzie:
@poire-z I cleared everything and rebuilt from scratch, but I'm still getting these errors:

```
11/17/19-15:24:39 DEBUG refresh on physical rectangle 47 262 446 196
munmap_chunk(): invalid pointer
Aborted (core dumped)
```

What's weird though is that I don't get them with the AppImage, which does seem to suggest something about my local setup. :-/

Frenzie:
Also:

```
corrupted size vs. prev_size
Aborted (core dumped)
```

poire-z:
:|
I got these while developping, when poking our of array bounds... (and messing with the bytes reserved by malloc for housekeping on each side of the array - so that happens on free()).
To get the line where that happens, and where the bad poke started, i needed to run koreader with:
`valgrind --log-file=/tmp/valgrind.log --leak-check=full --track-origins=yes` (hardcoded in kodev, the options parsing kills that if provided on the command line).
Then, you should fine "xtext.cpp" in that log when one of these has happened. (Unless it's something else).

Frenzie:
Yeah, that part of `kodev` is slightly broken. I just end up running valgrind with reader.lua manually since it's typically rare and fairly custom anyway.

Btw, possibly related:

```
/home/frans/src/kobo/koreader/base/xtext.cpp:366:9: warning: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119 [clang-analyzer-security.insecureAPI.strcpy]
        strcpy(m_lang, lang);
        ^
/home/frans/src/kobo/koreader/base/xtext.cpp:366:9: note: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119
/home/frans/src/kobo/koreader/base/xtext.cpp:376:30: warning: Call to 'malloc' has an allocation size of 0 bytes [clang-analyzer-optin.portability.UnixAPI]
        m_text = (uint32_t *)malloc(m_length * sizeof(*m_text));


/home/frans/src/kobo/koreader/base/xtext.cpp:1849:5: warning: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119 [clang-analyzer-security.insecureAPI.strcpy]
    strcpy(default_lang, lang);
```

Screenshot_2019-11-17_16-00-45

Frenzie:
> Then, you should fine "xtext.cpp" in that log when one of these has happened. (Unless it's something else).

How would I do that exactly? Valgrind makes it not crash.

Frenzie:
I can also check gdb of course.

Frenzie:
You mean like this?

```

==10289== Invalid read of size 4
==10289==    at 0x16FF0DBC: FT_Reference_Face (in /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1)
==10289==    by 0x16DDBA48: hb_ft_font_create_referenced (in /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0)
==10289==    by 0x1865DB4F: getHbFontData (xtext.cpp:566)
==10289==    by 0x1865DB4F: measureSegment (xtext.cpp:771)
==10289==    by 0x1865DB4F: XText::measure() (xtext.cpp:734)
==10289==    by 0x1865CA44: XText_measure(lua_State*) (xtext.cpp:2026)
==10289==    by 0x129DA6: lj_BC_FUNCC (buildvm_x86.dasc:809)
==10289==    by 0x11A081: lua_pcall (lj_api.c:1129)
==10289==    by 0x11093E: docall (luajit.c:121)
==10289==    by 0x1111EB: handle_script (luajit.c:292)
==10289==    by 0x1118D9: pmain (luajit.c:553)
==10289==    by 0x129DA6: lj_BC_FUNCC (buildvm_x86.dasc:809)
==10289==    by 0x11A0C3: lua_cpcall (lj_api.c:1153)
==10289==    by 0x111985: main (luajit.c:582)
```

Whole file attached.

[valgrind.log](https://github.com/koreader/koreader/files/3855703/valgrind.log)

poire-z:
gdb wasn't helpful, it's too late when you don't know where it could happen. Valgrind checks bounds, and tell you at which point you started to bad read/write. I think it might also prevent the crash, not sure.

poire-z:
> You mean like this?

Yep, exactly, that's the niceness of valgrind :)

poire-z:
^^ that's all clang-tidy has to say? (Didn't install it as it was too much stuff). I'd be interested by the full output, if there's other interesting reported stuff.

Frenzie:
It doesn't say anything other than the three things I quoted. Unless you want the full story of "taking if/else/etc path".

poire-z:
Nope, not needed, thanks :)

poire-z:
Invalid read of size 4
at 0x16FF0DBC: FT_Reference_Face (in **/usr/lib/x86_64-linux-gnu/** libfreetype.so.6.16.1)
by 0x16DDBA48: hb_ft_font_create_referenced (in **/usr/lib/x86_64-linux-gnu/** libharfbuzz.so.0.20301.0)

Any reason it references your system freetype and harfbuzz?` ldd build/x86_64.../libs/libkoreader-xtest.so` show the ones alongside it on my side.
Just valgrind printing wrong stuff? Or your emulator is using these, possibly old versions with structs smaller, so xtext.so or even harfbuzz poking outside expected structs larger?

Frenzie:
@poire-z I wanted to try asciinema, so in case it's useful: https://asciinema.org/connect/bc65adc2-2322-4932-9eb7-0e9f0d387b7f

Frenzie:
> Any reason it references your system freetype and harfbuzz?

That's a question for yourself, isn't it? ;-)

poire-z:
I'm not the build expert :/
At least try ldd ...libkoreader-xtest.so , see what it says.

Frenzie:
```
$ ldd libkoreader-xtext.so 
	linux-vdso.so.1 (0x00007fffccbdd000)
	libfreetype.so.6 => /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/./libfreetype.so.6 (0x00007fa2b99e3000)
	libharfbuzz.so.0 => /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/./libharfbuzz.so.0 (0x00007fa2b98bd000)
	libfribidi.so.0 => /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/./libfribidi.so.0 (0x00007fa2b989e000)
	libunibreak.so.3 => /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/./libunibreak.so.3 (0x00007fa2b9888000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fa2b96c9000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fa2b96af000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa2b94ec000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa2b9ad6000)
```

poire-z:
Can you run it with gdb, to have it stopped but not killed when it crashes, and look with (dunno which) command line tools which library are used, just to be sure. Like:
```
ls -l /proc/pid of luajit/map_files/ | grep harfb & freetype
.../koreader/base/build/x86_64-linux-gnu-debug/libs/libharfbuzz.so.0
```

Frenzie:
I don't know what luajit/map_files is I'm afraid.

Btw, this isn't some obscure thing. It happens on all my PCs & OSes. :-P

Frenzie:
Is ls -l the one you're looking for or do you mean something with ps?

```
ls -l 15894
```

Frenzie:
Oh, you mean /proc/<pid> :-P

Frenzie:
I still have no idea what you want me to do though.

Frenzie:
This does nothing:
```
frans@frenzie-desktop:/proc/15894/map_files$ ls -l /proc/15894/map_files/ | grep harfb
frans@frenzie-desktop:/proc/15894/map_files$ ls -l /proc/15894/map_files/ | grep freetype
```

poire-z:
Yes, /proc/PID_OF_LUAJIT/map_files/ should have links to the libraries mmaped.
Mhhh. If that's the pid of luajit koreader, and it is still running, just do ls, you should see some libraries... And among them some freetype and harfbuzz.so.

Frenzie:
Should I? It looks like this. :-P
Screenshot_2019-11-17_16-56-40

Frenzie:
To be clear, it actually looks like this, not like that:

Screenshot_2019-11-17_16-58-29

poire-z:
Yes ! :) And with `ls -l`, some magic will happen !

Frenzie:
Ah, right.

```
frans@frenzie-desktop:/proc/16465/map_files$ ls -l | grep harfb
lr-------- 1 frans frans 64 Nov 17 16:58 7fffcc079000-7fffcc082000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 16:58 7fffcc082000-7fffcc12b000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 16:58 7fffcc12b000-7fffcc16f000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 16:58 7fffcc16f000-7fffcc170000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 16:58 7fffcc170000-7fffcc171000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
```

```
frans@frenzie-desktop:/proc/16465/map_files$ ls -l | grep freetype
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3173000-7fffb317f000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb317f000-7fffb31e9000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb31e9000-7fffb3214000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3214000-7fffb321b000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb321b000-7fffb321c000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3bcc000-7fffb3bd9000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3bd9000-7fffb3c56000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3c56000-7fffb3c7f000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3c7f000-7fffb3c80000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3c80000-7fffb3c87000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 16:58 7fffb3c87000-7fffb3c88000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
```

poire-z:
OK, so your emulator is using your system libraries (and both for freetype?!)
While mine is only using those from koreader/libs (and not crashing).
Why is that ? :|

Frenzie:
I assume the "correct" FreeType is the one loaded through FFI (see https://github.com/koreader/koreader-base/blob/6c480198b00ec7ae4e8af25216e8a12fcabda9e6/ffi/freetype.lua#L14-L21) and the other one is pulled in through some library like xtext or kpvcrlib.

poire-z:
OK. That was when running koreader without valgrind? (To be sure it's not valgrind loading these)?

Frenzie:
Yes, that's just with gdb. I can do the same thing with KOReader that actually runs with a commit from one or two days ago?

NiLuJe:
I don't recall if that's a Gentoo only thing, by we have a magical script by the name of `lddtree` in there which prints the full dep-tree of a binary ;).

NiLuJe:
Eh.

```
┌─(ROOT@illyria:pts/1)────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~SVN/Bash/trunk)─┐
└─(0.10:23%:17:09:#)── equery b $(which lddtree)                                                                                                                                                                                                                                                                                                        ──(Sun, Nov 17)─┘
 * Searching for /usr/bin/lddtree ... 
app-misc/pax-utils-1.2.4 (/usr/bin/lddtree)

┌─(ROOT@illyria:pts/1)────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~SVN/Bash/trunk)─┐
└─(0.16:23%:17:10:#)── eix pax-utils                                                                                                                                                                                                                                                                                                                    ──(Sun, Nov 17)─┘
[I] app-misc/pax-utils
     Available versions:  1.2.3 (~)1.2.3-r3 (~)1.2.4 {caps debug python seccomp PYTHON_SINGLE_TARGET="python2_7 python3_5 python3_6 python3_7" PYTHON_TARGETS="python2_7 python3_5 python3_6 python3_7"}
     Installed versions:  1.2.4(10:52:16 PM 11/11/2019)(seccomp -caps -debug -python PYTHON_SINGLE_TARGET="python3_7 -python2_7 -python3_5 -python3_6" PYTHON_TARGETS="python2_7 python3_6 python3_7 -python3_5")
     Homepage:            https://wiki.gentoo.org/index.php?title=Project:Hardened/PaX_Utilities
     Description:         ELF utils that can check files for security relevant properties
```

Tough luck ;p.

Frenzie:
Tough luck how? pax-utils is a Debian package, so if you tell me what to do with it I'll follow the instructions like a dumb sheep. :-P

poire-z:
Also try without gdb (in case it's gdb that loads them :|) And yes, with an earlier commit.
@NiLuJe : can you tell us with that (ls -l /proc/pid/map_files) trick what your emulator is using.

Btw, `lsof -p 21092` tells that too :)

Frenzie:
> Also try without gdb (in case it's gdb that loads them :|)

Well that's not really possible, is it? :-P

NiLuJe:
@Frenzie: Ah, thought it was Gentoo Hardened specific ;). Basically use it instead of `ldd` on everything you can think of ;p.

poire-z:
Ctrl-Z before it crashes :)

Frenzie:
Considering it insta-crashes… that means removing the `kodev run`/`./reader.lua` line? :-P

Frenzie:
Oh, hold on, looks like I can get it to run a little bit if I disable CoverBrowser. Anyway, same results.

```
$ ls -l |grep harf
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8fa1000-7fdcf8faa000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8faa000-7fdcf9053000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf9053000-7fdcf9097000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf9097000-7fdcf9098000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf9098000-7fdcf9099000 -> /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20301.0
```

```
$ ls -l |grep free
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8274000-7fdcf8280000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8280000-7fdcf82ea000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf82ea000-7fdcf8315000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8315000-7fdcf831c000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf831c000-7fdcf831d000 -> /home/frans/src/kobo/koreader/base/build/x86_64-linux-gnu-debug/libs/libfreetype.so.6
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8cc7000-7fdcf8cd4000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8cd4000-7fdcf8d51000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8d51000-7fdcf8d7a000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8d7a000-7fdcf8d7b000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8d7b000-7fdcf8d82000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
lr-------- 1 frans frans 64 Nov 17 17:19 7fdcf8d82000-7fdcf8d83000 -> /usr/lib/x86_64-linux-gnu/libfreetype.so.6.16.1
```

Frenzie:
Anyway, here's @NiLuJe 's suggestion.

[have-fun.txt](https://github.com/koreader/koreader/files/3855799/have-fun.txt)

poire-z:
Add use_xtest=false in your reader.settings.lua,, and try the same - just to be sure it's not libkoreader-xtext.so that pulls them. Check for where freetype is (there shouldn't be any harfbuzz). open an EPUB, harfbuzz should be pulled by crengine, see where it comes from.

NiLuJe:
The iffiest looking one is harfbuzz. It only has a build rpath. (And I can confirm that on our Kobo builds, too).

NiLuJe:
Okay, just finished building the emu on my end, and I can reproduce the same crashes ;).

Frenzie:
@poire-z No crash with `use_xtext=false`, or at least not one that's excessively easy to trigger. The libraries look the same though, including HarfBuzz?

Frenzie:
@poire-z Speaking of, MuPDF uses HarfBuzz, doesn't it?

Frenzie:
@NiLuJe Like I said, happens on all my machines with various versions/OSes (even if none of them are Gentoo) so I shouldn't think it's particularly niche. ;-)

NiLuJe:
Also, lightbulb moment: the system libraries we see in maps_file are because of SDL2.

Frenzie:
Plenty of them are, but I don't think SDL2 pulls in HarfBuzz (or even FreeType?).

Screenshot_2019-11-17_17-40-40

Frenzie:
Oh wait, I can use your tools for that.

NiLuJe:
Double-lightbulb moment: having two copies of the same libraries initialized in the process is a recipe for kablooey.

`env LD_LIBRARY_PATH="${PWD}/libs:${LD_LIBRARY_PATH}"` prevents that from happening, and, magic, no more crashes ;).

Frenzie:
```
frans@frenzie-desktop:/usr/lib/x86_64-linux-gnu$ lddtree libSDL2-2.0.so
libSDL2-2.0.so => ./libSDL2-2.0.so (interpreter => none)
    libasound.so.2 => /usr/lib/x86_64-linux-gnu/libasound.so.2
        ld-linux-x86-64.so.2 => /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
    libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2
    libpulse-simple.so.0 => /usr/lib/x86_64-linux-gnu/libpulse-simple.so.0
        libpulsecommon-12.2.so => /usr/lib/x86_64-linux-gnu/pulseaudio/libpulsecommon-12.2.so
            libX11-xcb.so.1 => /usr/lib/x86_64-linux-gnu/libX11-xcb.so.1
            libxcb.so.1 => /usr/lib/x86_64-linux-gnu/libxcb.so.1
                libXau.so.6 => /usr/lib/x86_64-linux-gnu/libXau.so.6
                libXdmcp.so.6 => /usr/lib/x86_64-linux-gnu/libXdmcp.so.6
                    libbsd.so.0 => /usr/lib/x86_64-linux-gnu/libbsd.so.0
            libICE.so.6 => /usr/lib/x86_64-linux-gnu/libICE.so.6
            libSM.so.6 => /usr/lib/x86_64-linux-gnu/libSM.so.6
                libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1
            libXtst.so.6 => /usr/lib/x86_64-linux-gnu/libXtst.so.6
            libsystemd.so.0 => /lib/x86_64-linux-gnu/libsystemd.so.0
                liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5
                liblz4.so.1 => /usr/lib/x86_64-linux-gnu/liblz4.so.1
                libgcrypt.so.20 => /lib/x86_64-linux-gnu/libgcrypt.so.20
                    libgpg-error.so.0 => /lib/x86_64-linux-gnu/libgpg-error.so.0
            libwrap.so.0 => /lib/x86_64-linux-gnu/libwrap.so.0
                libnsl.so.1 => /lib/x86_64-linux-gnu/libnsl.so.1
            libsndfile.so.1 => /usr/lib/x86_64-linux-gnu/libsndfile.so.1
                libFLAC.so.8 => /usr/lib/x86_64-linux-gnu/libFLAC.so.8
                libogg.so.0 => /usr/lib/x86_64-linux-gnu/libogg.so.0
                libvorbis.so.0 => /usr/lib/x86_64-linux-gnu/libvorbis.so.0
                libvorbisenc.so.2 => /usr/lib/x86_64-linux-gnu/libvorbisenc.so.2
            libasyncns.so.0 => /usr/lib/x86_64-linux-gnu/libasyncns.so.0
                libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2
            libdbus-1.so.3 => /lib/x86_64-linux-gnu/libdbus-1.so.3
        libcap.so.2 => /lib/x86_64-linux-gnu/libcap.so.2
    libpulse.so.0 => /usr/lib/x86_64-linux-gnu/libpulse.so.0
    libsndio.so.7.0 => /usr/lib/x86_64-linux-gnu/libsndio.so.7.0
    libX11.so.6 => /usr/lib/x86_64-linux-gnu/libX11.so.6
    libXext.so.6 => /usr/lib/x86_64-linux-gnu/libXext.so.6
    libXcursor.so.1 => /usr/lib/x86_64-linux-gnu/libXcursor.so.1
        libXrender.so.1 => /usr/lib/x86_64-linux-gnu/libXrender.so.1
        libXfixes.so.3 => /usr/lib/x86_64-linux-gnu/libXfixes.so.3
    libXinerama.so.1 => /usr/lib/x86_64-linux-gnu/libXinerama.so.1
    libXi.so.6 => /usr/lib/x86_64-linux-gnu/libXi.so.6
    libXrandr.so.2 => /usr/lib/x86_64-linux-gnu/libXrandr.so.2
    libXss.so.1 => /usr/lib/x86_64-linux-gnu/libXss.so.1
    libXxf86vm.so.1 => /usr/lib/x86_64-linux-gnu/libXxf86vm.so.1
    libwayland-egl.so.1 => /usr/lib/x86_64-linux-gnu/libwayland-egl.so.1
    libwayland-client.so.0 => /usr/lib/x86_64-linux-gnu/libwayland-client.so.0
        libffi.so.6 => /usr/lib/x86_64-linux-gnu/libffi.so.6
    libwayland-cursor.so.0 => /usr/lib/x86_64-linux-gnu/libwayland-cursor.so.0
    libxkbcommon.so.0 => /usr/lib/x86_64-linux-gnu/libxkbcommon.so.0
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0
    librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
```

NiLuJe:
@Frenzie: I said SDL2, but it might be GTK pulling pango, more likely ;).

Frenzie:
Oh right, I forgot about that GTK open dialog. I guess I should remove it since I can't figure out how to fix it (within 10-20 minutes which is basically how long I spent implementing it in the first place…). /lesigh Yeah, that'd be pulling in FT & HB.

NiLuJe:
@poire-z: FWIW, we'll need `static-libstdc++`, I'm pretty sure it won't load on some Kindles otherwise ;).

NiLuJe:
@Frenzie: In any case, enforcing `LD_LIBRARY_PATH` seems like a Good Thing (both in kodev & AppImage).

Frenzie:
Tbh I thought we already did.

poire-z:
(I don't have any gtk stuff installed, which might explain why I was unaffected.)

Frenzie:
I'm not wrong, I did put that in the script. :-P

https://github.com/koreader/koreader/blob/89c0bd08863748dc8338987cb98dcb59436e26d3/platform/appimage/AppRun#L10-L11

Frenzie:
But this would mean I can no longer just run `./reader.lua`? :-/

NiLuJe:
@Frenzie: Which explains why you couldn't repro w/ the AppImage. Yay! :).

NiLuJe:
> But this would mean I can no longer just run `./reader.lua`? :-/

Hmm, yeah, I can't think of a neat solution to that.

I can think of a very terrible one: an LD_PRELOAD hack that sets the end var. Kindles do that >_<".

poire-z:
> FWIW, we'll need static-libstdc++, I'm pretty sure it won't load on some Kindles otherwise

You mean that in relation to the Android5 issue?
Btw, adding `-static-libstdc++` does not make my non-debug arm libkoreader-xtext.so grows by a single bit. If I add somewhere `std::vector<int> v = {7, 5, 16, 8};`, it does grow by 36 bytes.
It does grow more in debug mode for x86.
So, dunno if it really changes something if I don't pull any stdc++. Would it on Kindle?

Frenzie:
@NiLuJe This doesn't seem to actually do anything?

```
diff --git a/kodev b/kodev
index 0c1ac28f..a41c842e 100755
--- a/kodev
+++ b/kodev
@@ -101,6 +101,7 @@ function setup_env() {
     files=$(find . -maxdepth 1 -name 'koreader-emulator-*' | ${SETUP_ENV_GREP_COMMAND})/koreader
     export EMU_DIR=${files[0]}
+    export LD_LIBRARY_PATH=${EMU_DIR}/libs:$LD_LIBRARY_PATH
 }
 
 function kodev-fetch-thirdparty() {
```

NiLuJe:
@poire-z:

The simple fact of linking against the STL implies ABI compat checks.

Here:

```
 Version needs section '.gnu.version_r' contains 2 entries:
  Addr: 0x0000000000001360  Offset: 0x001360  Link: 3 (.dynstr)
   000000: Version: 1  File: libstdc++.so.6  Cnt: 2
   0x0010:   Name: CXXABI_1.3.9  Flags: none  Version: 5
   0x0020:   Name: GLIBCXX_3.4  Flags: none  Version: 3
```

That won't do on Kindle ;).

NiLuJe:
@Frenzie: PWD changes between when that's set and when we run ;).

Something like

```diff
diff --git a/kodev b/kodev
index 0c1ac28f..84bca093 100755
--- a/kodev
+++ b/kodev
@@ -100,7 +100,8 @@ function setup_env() {
     fi
     files=$(find . -maxdepth 1 -name 'koreader-emulator-*' | ${SETUP_ENV_GREP_COMMAND})/koreader
     assert_ret_zero $? "Emulator not found. Please build it first."
-    export EMU_DIR=${files[0]}
+    export EMU_DIR="${files[0]}"
+    export LD_LIBRARY_PATH="$(realpath ${EMU_DIR})/libs:${LD_LIBRARY_PATH}"
 }
 
 function kodev-fetch-thirdparty() {
```

Works, by ensuring that's an canonical absolute path ;).

NiLuJe:
That, or `readlink -f`.

One, or both of those, will gloriously not work on macOS, because Apple. Unless one install coreutils.

yparitcher:
btw with archlinux i also get the system versions of FT and harfbuzz, but the emulator works even with xtext

```
/usr/lib/libharfbuzz.so.0.20600.4
/AUR/koreader/base/build/x86_64-pc-linux-gnu-debug/libs/libfreetype.so.6
/usr/lib/libfreetype.so.6.17.1
```

i do have gtk installed, and it works both if run via kodev or via ./reader.lua

NiLuJe:
I'm running FT snapshots on my system, and @Frenzie is on a Debian thingy that may be running some older versions, while you're exactly matching the vendroed version on Arch, which might explain why nothing blows up.

Having two copies of the same library initialized in the same process is still a highly volatile proposition ;).

yparitcher:
@Frenzie  I found a difference, my `libkoreader-xtext.so` pulls in `libstdc++` yours does not.
i do not know if it is connected.
[lddtree.txt](https://github.com/koreader/koreader/files/3856046/lddtree.txt)

my results
```
libkoreader-xtext.so (interpreter => None)
    libfreetype.so.6 => /AUR/koreader/base/build/x86_64-pc-linux-gnu-debug/libs/libfreetype.so.6
    libharfbuzz.so.0 => /AUR/koreader/base/build/x86_64-pc-linux-gnu-debug/libs/libharfbuzz.so.0
    libfribidi.so.0 => /AUR/koreader/base/build/x86_64-pc-linux-gnu-debug/libs/libfribidi.so.0
    libunibreak.so.3 => /AUR/koreader/base/build/x86_64-pc-linux-gnu-debug/libs/libunibreak.so.3
    libstdc++.so.6 => /usr/lib/libstdc++.so.6
        ld-linux-x86-64.so.2 => /usr/lib/ld-linux-x86-64.so.2
    libm.so.6 => /usr/lib/libm.so.6
    libgcc_s.so.1 => /usr/lib/libgcc_s.so.1
    libc.so.6 => /usr/lib/libc.so.6
```
your results
```
libkoreader-xtext.so => ./libkoreader-xtext.so (interpreter => none)
    libfreetype.so.6 => ./libfreetype.so.6
    libharfbuzz.so.0 => ./libharfbuzz.so.0
    libfribidi.so.0 => ./libfribidi.so.0
    libunibreak.so.3 => ./libunibreak.so.3
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
    ld-linux-x86-64.so.2 => /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
```

Frenzie:
> a Debian thingy

I generally run Debian stable or testing on my desktop. That's currently stable, it'll turn into testing whenever I start to feel it's a bit outdated, i.e., when I start to feel the need to backport more than a select handful of programs. Right now my only backport is po4a, because Debian comes with 0.55 (even in unstable ;_;) and 0.56 contains this fix: `https://github.com/mquinson/po4a/pull/173`.

On my laptop I normally run the latest *buntu. Then I've got another old laptop which varies (currently Mint), doesn't do too much since I got my current sidegrade in '16, and an external drive with a bunch of live images. I've also got an old netbook that runs the latest Xubuntu LTS. And of course Ubuntu 16.04 in Docker. I think that's about all of it.

So yeah, mostly Debubuntu in various degrees of "outdated" (compared to Arch/Gentoo).

NiLuJe:
@yparitcher: I'm guessing that was a simply a post `-static-libstdc++` build ;).

yparitcher:
nope koreader @ 89c0bd08863748dc8338987cb98dcb59436e26d3 base was @6c48019 before  `-static-libstdc++`,

however my kindle KT4 build, built last night has 
```
lddtree libkoreader-xtext.so 
libkoreader-xtext.so (interpreter => None)
    libfreetype.so.6 => /home/yechiel/Downloads/dtmpfs/kindle/mnt/us/koreader/libs/libfreetype.so.6
    libharfbuzz.so.0 => /home/yechiel/Downloads/dtmpfs/kindle/mnt/us/koreader/libs/libharfbuzz.so.0
    libfribidi.so.0 => /home/yechiel/Downloads/dtmpfs/kindle/mnt/us/koreader/libs/libfribidi.so.0
    libunibreak.so.3 => /home/yechiel/Downloads/dtmpfs/kindle/mnt/us/koreader/libs/libunibreak.so.3
    libm.so.6 => None
    libgcc_s.so.1 => None
    libc.so.6 => None
    ld-linux.so.3 => None
```
and it is working just fine without `-static-libstdc++`

NiLuJe:
@yparitcher: You've just reminded me that we enforce `-static-libstdc++` on a bunch of !emulator platforms anyway ;).

poire-z:
Note: I forgot to implement this in the use_xtext variant of TextBoxWidget:
https://github.com/koreader/koreader/blob/ea67b9be77131ada5ddc1afa781f11ef934f76b5/frontend/ui/widget/textboxwidget.lua#L356-L361
It's only use is in CoverBrowser Mosaic mode, with "fake" text covers, where we can now have words splitted in the middle :| (but it also avoids going to excessive small font size)
https://github.com/koreader/koreader/blob/ea67b9be77131ada5ddc1afa781f11ef934f76b5/plugins/coverbrowser.koplugin/mosaicmenu.lua#L215-L219

poire-z:
@Frenzie regarding your gitter comment _maybe there's a sneaky mistake in the xtext/fribidi Lua API use_: you can disable all use of xtext via `Developper options> Disable enhanced UI text shaping (xtext)` and see if you still crash or not (let me hope you still do ! :)

PR #5600, opened 2019-11-16
ConfigDialog: allow larger toggle title text

poire-z:
-  ConfigDialog: allow larger toggle title text : avoid a few truncated text, as some fit when allowed
to bite a bit on the (large) left padding.
Allows `Embedded fonts`, `Sync T/B Margins`, and `Render Mode` (truncated since we added the finetuning button) to no more be truncated.

- cre fonts list: ignore nerdfonts/symbol.ttf : it's rejected anyway by crengine, this just avoids
an error from being logged.

### DIFF
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -80,7 +80,7 @@ function CreDocument:engineInit()
         -- we need to initialize the CRE font list
         local fonts = FontList:getFontList()
         for _k, _v in ipairs(fonts) do
-            if not _v:find("/urw/") then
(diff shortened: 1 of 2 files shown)


NiLuJe:
Oh, good point the the CRe blacklist.

Which reminds me, there's already a blacklist in place for some Kindle fonts. I don't recall where/how it's implemented (and I'm not on my dev box RN), but that doesn't look like it's around here.

(Mostly a note to self to check it out tomorrow ;)).

Frenzie:
The blacklist is here:

https://github.com/koreader/koreader/blob/89c0bd08863748dc8338987cb98dcb59436e26d3/frontend/fontlist.lua#L11-L59

PR #2032, opened 2016-05-17
russian keyboard layout improvement from mergen3107

houqp:
This is the new keyboard layout proposed by mergen3107 in https://github.com/koreader/koreader/pull/2029. Let's discuss it here.

PR #5610, opened 2019-11-18
[chore] Extract Cyrillic/Russian keyboard

Frenzie:
Follow-up to #5583.


Frenzie:
@yparitcher Quick sanity check before I continue?

Frenzie:
Cross-ref to #2032 for specifics on the Cyrillic/Russian keyboard.

yparitcher:
> @yparitcher Quick sanity check before I continue?

looks good so far. 

btw if you only need 4 layers for a specific language you can, see hebrew for example.

Frenzie:
I haven't looked at them all yet. Portuguese seems to be using all layers fully so it's a bit problematic, but I think it should be covered by my updated English keyboard with swipes/popups.

yparitcher:
either the english one, or if you want to have a Portuguese specific keyboard, you can move the letters with diacritics from other layers to a popup.

Frenzie:
I don't know if I want anything at this point. It's not like I've researched stuff like https://en.wikipedia.org/wiki/Portuguese_keyboard_layout yet, nor our own history. ;-) Iirc @robert00s wrote it.

Frenzie:
@NiLuJe @poire-z @pazos With the exception of Portuguese, ready for review.

frontend/ui/data/keyboardlayouts/ja_keyboard.lua
```diff
@@ -1,103 +1,106 @@
...
-            { "ァ",        "ぁ",    ")",    "0",    "ァ",     "ぁ",    ")",    "0",    "ァ",    "ぁ",    ")",    "0", },
-            { "ャ",        "ゃ",    "_",    "-",    "ャ",     "ゃ",    "_",    "-",    "ャ",    "ゃ",    "_",    "-", },
-            { "゛",         "゛",     "+",    "=",    "゛",      "゛",     "+",    "=",    "゛",     "゛",     "+",    "=", },
+        {  --  1       2       3       4
+            { "ア",    "あ",    "~",    "`", },
+            { "カ",    "か",    "!",    "1", },
+            { "サ",    "さ",    "@",    "2", },
+            { "タ",    "た",    "#",    "3", },
+            { "ナ",    "な",    "$",    "4", },
+            { "ハ",    "は",    "%",    "5", },
```
poire-z: Any idea why that one (2nd char among the 4) is highlighted by github diff?
(And actually always one at the same place in the 3 following sections.)

Frenzie: I double checked the characters and they seem to be exactly the same, so I can only assume it's an obscure GH bug.

Screenshot_20191118_232622

Frenzie:
I'll merge it so that it makes it into the next nightly together with the keyboard layout selection. I tested all the keyboards and I didn't do anything new/weird, so it should be good. ^_^

poire-z:
The korean keyboard got a large globe key, which make the popup large too:
<kbd>image</kbd>

Also, for usability, it would be nice if all the globe keys were at the same position when switching layouts (the hebrew keyboard makes it move position) - possibly by making the space bar larger so the bottom row always take the full width:
`[Sym] [Globe] [Optional Aéss] [Variable width space] [0, 1 or 2 punct keys] [Enter]`
Sym and Enter always stuck to their sides of screen.

Frenzie:
You know, I was going to suggest just making the buttons larger to that `pt_BR` question. ;-) I thought it looked a bit out of place at the regular size but feel free to touch it up slightly.

PS Japanese doesn't have Sym. (But it does have roughly the same location.)

poire-z:
Not clear in my post, but I wanted to say I don't like it larger :) I like it at the smaller size it has in the english keyboard.
(I'm not planning on touching the keyboards these busy days :))

frontend/ui/data/keyboardlayouts/ja_keyboard.lua
Frenzie: @poire-z I now understand why it is. It relates to why I accidentally deleted a few layers in the Japanese layout. The tenth layout is exactly the same as the second, except on this specific line where it's `ぱ` instead of `は`. But GH (intelligently?) regards this as a change rather than as deleting the third to tenth layout. It probably has something to do with the number of spaces.

Screenshot_2019-12-30_16-26-15

PR #5625, opened 2019-11-22
Rewrite unicodeCodepointToUtf8 w/ bitopt

NiLuJe:
Avoids costly divs & modulos

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5625)

### DIFF
--- a/frontend/util.lua
+++ b/frontend/util.lua
@@ -7,6 +7,10 @@ local dbg = require("dbg")
 local _ = require("gettext")
 local T = BaseUtil.template
 
+local rshift = bit.rshift
+local band = bit.band
+local bor = bit.bor
+
 local util = {}
 
 --- Strips all punctuation and spaces from a string.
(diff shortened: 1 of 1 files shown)


NiLuJe:
Random optimization because I noticed I had a copy of that in C lying around in FBInk ;p.

NiLuJe:
(We can't use the unsigned overflow trickery from the first ref. for the surrogates test because Lua :/).

poire-z:
For info: https://luapower.com/luajit-notes
> - multiplications and additions are cheaper than memory access, so storing the results of these operations in temporary variables might actually harm performance (more register spills).
> - divisions are 4x slower than multiplications on x86, so when dividing by a constant, it helps turning x / c into x * (1 / c) since the constant expression is folded -- LuaJIT does this already for power-of-2 constants where the semantics are equivalent.
> - the % operator is slow (it's implemented in terms of math.floor() and division) and really kills hot loops; math.fmod() is even slower; I don't have a solution for this except for x % powers-of-two which can be computed with bit ops.

Issue #3031, opened 2017-07-27
Battery statistics are wrong


embryo10:
* KOReader version: 2015.11-1054
* Device: Kobo Glo

In Battery statistics, in the "Awake since last charge" the total minutes is still counting even when my Glo is powered off.
Same thing happens in the "Since last charge" section too (must be based in the same measurement).
What can I do to debug this problem?

I would also like to ask if its possible to have the time in minutes but also in DD:HH:MM:SS format, that is more easy to humans than say "2645.60 minutes".

And since I don't remember if I already ask for this (I definitely ask in the forum) I would like the option to also move through pages with taps.

Issue #2937, opened 2017-06-04
Battery statistics: "sleeping" time is not identified correctly


protist:
* KOReader version: koreader-nightly-20170430
  * koreader-kobo-arm-kobo-linux-gnueabihf-v2015.11-1023-g773e24f
* Device: Kobo H2O

#### Issue
"Battery statistics" presents incorrect information.

#### Steps to reproduce
1. Open up "Battery statistics".
2. Observe incorrect data.

Currently the statistics say
```
Awake since last charge
  Total minutes     39703.47
Sleeping since last charge
  Total minutes     0.15
```
My kobo has been sleeping for most of this time, and only on for less than ~1 hour per day.

Frenzie:
Same here, also see https://www.mobileread.com/forums/showthread.php?t=286933

Pinging @Hzj-jie

protist:
Hmmm… I presumed that the Kobo really *was* sleeping, and that the battery statistics just got it wrong. After the last koreader update, my battery life included phenomenally, which I presumed was due to fixes relating to issue #2239. That's made me think that it really *was* suspending properly now. However, the first post of that linked thread hypothesises otherwise, although it might just be pure speculation.

Frenzie:
Mine definitely sleeps or it would've died many weeks ago.

protist:
That makes sense. As I implied above, my battery life is ridiculously good now. However, I have no idea how much drain occurs when awake, and how much when asleep. If only there were a way to inspect those statistics… 😁

cramoisi:
yup, it was sleeping. But I was experiencing  the kobo drain bug at the same time and tried to use battery stat to find the issue...
So, since I remove the magnet, it's ok regarding suspending and it's just the battery statistics which is wrong.

Hzj-jie:
Several changes are related,

https://github.com/koreader/koreader/commit/30378eb2a82ac1b2f686b0db638b194b642af4ab
  - This ensures the battery statistics can receive callbacks in filemanager

https://github.com/koreader/koreader/commit/d4756d777fd40e73bbea70307d9c16af76f9c82f
  - Fix a typo which causes the state to be not reset.

So you may try the latest version.

cramoisi:
Great ! I will test it as soon as it synch from the ota server.

protist:
[...]

PR #5626, opened 2019-11-22
Fix and improvements of battery stats plugin

robert00s:
Close: #3031, #2937

What was done:
- fix incorrectly show awake and sleeping,
- fix incorrectly show charging and discharging,
- remove unneeded debug mode,
- remove logging to external file,
- another method to show indentation text for better translate,
- prevent showing values like inf, -inf, nan in estimate time (we show now n/a),
- prevent showing values below zero,
- show extra confirm box when we want to reset data,
- show time in format xx**h**xx**m** (hours, minutes) instead only pure minutes,
- check at initialization that the device was charging when it was turned off (the battery is larger than at the time of the KO exit),

ToDo: 
- more tests
- show time in format xx**d**xx**h**xx**m** (days, hours, minutes)

Before:

After:

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5626)

### DIFF
--- a/plugins/batterystat.koplugin/main.lua
+++ b/plugins/batterystat.koplugin/main.lua
@@ -1,14 +1,12 @@
-
+local ConfirmBox = require("ui/widget/confirmbox")
 local DataStorage = require("datastorage")
 local KeyValuePage = require("ui/widget/keyvaluepage")
 local LuaSettings = require("luasettings")
 local PowerD = require("device"):getPowerDevice()
 local UIManager = require("ui/uimanager")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
-local T = require("ffi/util").template
 local dbg = require("dbg")
-local logger = require("logger")
-local util = require("ffi/util")
+local util = require("util")
 local _ = require("gettext")
 
 local State = {}
@@ -29,6 +27,7 @@ function State:toString()
 end
 
 local Usage = {}
+local INDENTATION = "   " --three spaces
 
 function Usage:new(o)
     o = o or {}
@@ -43,7 +42,7 @@ end
 
 function Usage:append(state)
     local curr = State:new()
-    self.percentage = self.percentage + (state.percentage - curr.percentage)
+    self.percentage = self.percentage + math.abs(state.percentage - curr.percentage)
     self.time = self.time + os.difftime(curr.timestamp - state.timestamp)
 end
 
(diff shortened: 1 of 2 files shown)


plugins/batterystat.koplugin/main.lua
```diff
@@ -29,6 +27,7 @@ function State:toString()
 end
 
 local Usage = {}
+local INDENTATION = "   " --three spaces
```
Frenzie: Been trying to unify comment style. :-)
```suggestion
local INDENTATION = "   " -- Three spaces.
```

plugins/batterystat.koplugin/main.lua
```diff
@@ -64,59 +63,64 @@ function Usage:percentagePerHour()
...
 end
 
 function Usage:dumpRemaining(kv_pairs)
-    table.insert(kv_pairs, {_("    Estimated remaining hours"), shorten(self:remainingHours())})
+    table.insert(kv_pairs, {string.format("%s%s",INDENTATION, _("Estimated remaining hours")), shorten(self:remainingHours())})
 end
 
 function Usage:dumpCharging(kv_pairs)
-    table.insert(kv_pairs, {_("    Estimated hours for charging"), shorten(self:chargingHours())})
+    table.insert(kv_pairs, {string.format("%s%s",INDENTATION, _("Estimated hours for charging")), shorten(self:chargingHours())})
```
Frenzie: @poire-z Does this work for RTL?

poire-z: Possibly not - but go ahead with that as it is, it is done the same way in the System Statistics plugin:
<kbd>sysstats_rtl</kbd>

There are much more stuff than I expected that will need to be adapted for proper and code-clean RTL mirroring, and I have yet no idea how we'll go with these indented keys in Keyvaluepage...
We'll fix all these, either manually in each of them, or globally if we find a trick to have that done in KeyValuePage or even upper.

As an example, our lazy aligment tricks to use a different padding_left vs padding_right in FrameContainer - would need to be replaced by a HorizontalSpan and the FrameContainer in a HorizontalGroup, so HorizontalGroup is the only one that'd need to deal with mirroring.
My wish is that we should continue to just think LTR when developping, with some simple rules to follow for special cases (like chevron, or string bits to be BD.wrap() before concatenating them - and all should be done automagically by low level widgets when switching to some RTL language.

robert00s:
___
*[plugins/batterystat.koplugin/main.lua, line 30 at r1](https://reviewable.io/reviews/koreader/koreader/5626#-LuHFugYaECeXc_l1ul8-r1-30:-LuJOL5h36AxfPTJDJiW:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/594c425fd64d8f2f34254941fbeb9fb7aced59af/plugins/batterystat.koplugin/main.lua#L30)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

Been trying to unify comment style. :-)
```suggestion
local INDENTATION = "   " -- Three spaces.
```
</blockquote></details>

Done.

plugins/batterystat.koplugin/main.lua
Frenzie: What I mean is that lazy `   some spaces for alignment` can be translated however it needs to be in RTL if they're part of the string, as they are now. If this code would require extra code for inverting then this is a bad change.

Frenzie: Note that a string like `الذاكرة الإجمالية (MB)` overlooked and left out the spaces, unless it was a conscious decision. This is a Transifex flaw.

poire-z: > left out the spaces.

Oh, right. If I add the leading spaces back in ar_AA/koreader.po, then it looks ok! So, no issue with RTL at having spaces at start for indentation (I thought the bidi algo might push all spaces at end, but it doesn't). So good.

Anyway, the way this PR does it (indentation outside of translated string) is fine. (The way System statistics plugin does it is not.)

Frenzie: Right, but does "    ".."RTL string" actually put them at the start or at the end?

poire-z: At the start of the logical order of bytes. So it will be on the right (so at the RTL start) when text is rendered RTL (the bidi algo consider the spaces neutral, so all RTL when next real char is RTL).

There are many things the bidi algo (and/or xtexp.cpp :) do just right - we can still keep thinking LTR, and assume it will work ok with RTL.

Frenzie: Alright, in that case this bit of programmatic cleverness should okay. But in matters of internationalization anything clever is suspect by default. :-)

plugins/batterystat.koplugin/main.lua
```diff
@@ -64,59 +63,64 @@ function Usage:percentagePerHour()
...
+    table.insert(kv_pairs, {INDENTATION .. _("Estimated remaining hours"), shorten(self:remainingHours())})
 end
 
 function Usage:dumpCharging(kv_pairs)
-    table.insert(kv_pairs, {_("    Estimated hours for charging"), shorten(self:chargingHours())})
+    table.insert(kv_pairs, {INDENTATION .. _("Estimated hours for charging"), shorten(self:chargingHours())})
 end
 
 local BatteryStat = {
     settings = LuaSettings:open(DataStorage:getSettingsDir() .. "/batterstat.lua"),
```
poire-z: Can we fix that filename to "battery_stats.lua" ?
(I guess we don't need to care about migration for such transient stuff.)

robert00s: Of course :)

poire-z:
Merge'able? Was about to, but just saw your:
> ToDo:
    more tests
    show time in format xxdxxhxxm (days, hours, minutes)

robert00s:
Yes, it's ready to merge. 
Todo things I'll be do in near future in another PR :)

poire-z:
(A little request @robert00s : can you try deleting your branchs soon after they are merged? Instead of deleting a whole batch of branches a few times a year :) which makes the "sort PR/issues by recently updated" push all the alive issues below all your PRs of the trimester :) - unless you have a reason to do it that way)

Frenzie:
Deleting the branch after merge "updates" the PR? Sounds more like a GH bug/misfeature tbh.

Btw, since we never (?) have more than about a dozen PRs what's the use case precisely? They all fit on one page. :-)

robert00s:
Ok, I'll try to.

poire-z:
Thanks! :)

> what's the use case precisely

My landing page is https://github.com/koreader/koreader/issues?q=sort:updated-desc - where I see all the active PRs **and** issues - and I just have to look for "updated X minutes/hours ago" to see what has happened :) or quickly find out older but recent stuff I planned on commenting on.
(A few days ago, when @robert00s did some clean up, they all went to page 2, and it was hard to find the boundary - but ok, it all settles back after a few days)

> Sounds more like a GH bug/misfeature tbh

Surely is - but it has been like that for years :)

PR #5629, opened 2019-11-22
Other minor frontend.util cleanups

NiLuJe:
Shouldn't have kept that tab open ;p.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5629)


NiLuJe:
> > Reviewed 3 of 4 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/koreader/koreader/5629)_ status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @NiLuJe)
> 
> _[frontend/apps/reader/modules/readerdictionary.lua, line 570 at r1](https://reviewable.io/reviews/koreader/koreader/5629#-LuK-9No-IZdcmwmg7Bk:-LuK-9No-IZdcmwmg7Bl:blvxt9a) ([raw file](https://github.com/koreader/koreader/blob/e0f14a336efcfc3dee18c801353cd1f9bfa48b24/frontend/apps/reader/modules/readerdictionary.lua#L570)):_
> 
> > ```lua
> >     text = text:gsub("\xE2\x80\x99", "'") -- U+2019 (right single quotation mark)
> >     -- Strip punctuation characters around selection
> >     text = util.stripePunctuations(text)
> > ```
> 
> Hehe, that one bothers me every time I see it.

Candy cane punctuation marks! :D

NiLuJe:
(And by "big numbers", I mean 64bit, i.e., > UINT32_MAX)

frontend/util.lua
```diff
@@ -652,12 +653,12 @@ function util.getFriendlySize(size)
     size = tonumber(size)
     if not size or type(size) ~= "number" then return end
     local s
-    if size > 1024*1024*1024 then
-        s = string.format("%4.1f GB", size/1024/1024/1024)
-    elseif size > 1024*1024 then
-        s = string.format("%4.1f MB", size/1024/1024)
+    if size > 1073741824 then
```
poire-z: From the luajit notes posted at https://github.com/koreader/koreader/pull/5625#issuecomment-557425843
> the constant expression is folded

So, it shouldn't hurt letting it be `1024*1024*1024` - but if you don't trust it :) at least, put that as a comment so we don't have to wonder what that magic number `1073741824` is :)
`if size > 1073741824 then -- 1024*1024*1024`

frontend/util.lua
```diff
@@ -652,12 +653,12 @@ function util.getFriendlySize(size)
...
+        s = string.format("%4.1f GB", rshift(size / 1024, 10) / 1024)
```
poire-z: I find these ones quite unreadable :) vs the clearer `size/1024/1024/1024`.
Given how non-performance impacting this function is (called in Book information and nothing much else), I think that optimisation vs clarity would not be really needed.
(If you insist, please put the simple/naive formula as a comment.)

Frenzie: Yeah, I have to agree with that. Maintainability over performance, unless it's clear there's an issue. ;-)

NiLuJe: Fair enough ;).

PR #5638, opened 2019-11-24
Screensaver: avoid crash / CoverBrowser Mosaic: fix alignment

### LINKED ISSUE #5363: PocketBook Inkpad 3 (PB740) - time is a few minutes off when waking up from sleeping mode.
* KOReader version: v2019.09.2
* Device: Inkpad 3 / PB740 - Firmware: T740.5.17.1680

#### Issue

When I open a document (PDF) with koreader and let the device go to sleep (either pushing the button on the bottom of the device or waiting) when the device wakes up the timer is off a few minutes (can be longer if it stayed in sleep mode for a long time, it's around 10 minutes / 12h (?)). The only way is then to configure the right time on the device again, but if it goes to sleep again (when running koreader) the problem will come again at next wake up and the internal clock will be late again.
When using the 'original' reading software, or entering sleeping mode when the device is on a menu page, this problem doesn't happen, but only when entering sleep mode from within koreader.
Note: wifi is always off, didn't try with wifi on.

#### Steps to reproduce

1 Update the time using the wireless connection or manually and check that it's the same than an another clock.
2 Install Koreader
3 Open a document (PDF file in my example)
4 Push the button at the bottom of the device or wait for it to enter sleep mode
5 Wait a few hours
6 Start the device again.
7 Look at the time on the device: it's XX minutes late compared to the device used in #2 or the time servers.

Thanks!

### LINKED ISSUE #5633: Crash & wake up when: Sleep/power off + empty History + Use last book's cover as screensaver
* KOReader version: v2019.11-27-g4740ab1_2019-11-22/  
* Device: KA1

#### Issue
When I have "Use last book's cover as screensaver" and History is empty my device crashes when power off/put to sleep.
#### Steps to reproduce
1. "Use last book's cover as screensaver", purge .sdr for all books on History page
2. Put device to sleep: device will go into sleep with a crashlog as screensaver and then wake up again after 10-15 seconds
3. Put device to sleep again. Now it won't crash.

[crash.log](https://github.com/koreader/koreader/files/3882875/crash.log)

### LINKED ISSUE #5634: File browser displays items on new rows centred
**Current behaviour:**

Before

**Desired behaviour:**

After

This could possibly be a setting that users can change to get the desired result?

poire-z:
Small fixes:
- Screensaver: avoid crash when no lastfile. Closes #5633. 
- CoverBrowser: Mosaic: fix alignment on non-fully filled rows. Closes #5634.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5638)

### DIFF
--- a/frontend/ui/screensaver.lua
+++ b/frontend/ui/screensaver.lua
@@ -261,7 +261,7 @@ function Screensaver:show(event, fallback_message)
             exclude = doc_settings:readSetting("exclude_screensaver")
         end
         if exclude ~= true then
-            if lfs.attributes(lastfile, "mode") == "file" then
+            if lastfile and lfs.attributes(lastfile, "mode") == "file" then
                 local doc = DocumentRegistry:openDocument(lastfile)
                 if doc.loadDocument then -- CreDocument
                     doc:loadDocument(false) -- load only metadata
@@ -287,7 +287,7 @@ function Screensaver:show(event, fallback_message)
         end
     end
     if screensaver_type == "bookstatus" then
-        if lfs.attributes(lastfile, "mode") == "file" then
+        if lastfile and lfs.attributes(lastfile, "mode") == "file" then
             local doc = DocumentRegistry:openDocument(lastfile)
             local doc_settings = DocSettings:open(lastfile)
             local instance = require("apps/reader/readerui"):_getRunningInstance()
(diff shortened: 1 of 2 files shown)


NiLuJe:
I, err, kind of liked the center alignment ;'(.

dontcrash:
@NiLuJe I did originally say it could be a setting :( But it does look super strange, especially when there’s no files and the back button is centred

poire-z:
> But it does look super strange, especially when there’s no files and the back button is centred

Well, that did not happen, because of the initial inconsistency mentionned in https://github.com/koreader/koreader/issues/5634#issuecomment-557866878

What about https://github.com/koreader/koreader/issues/5634#issuecomment-557915523 ?:
> I guess centering could be preferable in History, for those that have only a few items and use it as a kind of home page - and centering makes it nicer - while left alignment kinda makes more sense in File browser.
That's doable (just use a CenterContainer instead of the added LeftContainer).

NiLuJe:
Yep, that makes sense ;).

PR #5639, opened 2019-11-24
[RFC] Switch remaining keyboard icons to glyphs

NiLuJe:
The only two left were Del/Backspace & Enter.

This also allows one to render a specific key in synthetic bold, in order to do so for those two glyphs.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5639)


NiLuJe:
Before:
kb_icon

After:
kb_glyph

Frenzie:
Thanks, I was just about to ask. ;-)

NiLuJe:
The enter glyph is slightly smaller, hence the "bold" trick to compensate.

FWIW, unbolded:

kb_cr_bis

NiLuJe:
Other candidates for the enter glyph (c.f., https://github.com/koreader/koreader/pull/5583#issuecomment-554717198):

* The actual Enter glyph, which is unfortunately super-weird in FreeSerif:
kb_enter

* CR:
kb_cr

* Nerdfont's Enter key:
kb_nerdf

NiLuJe:
TL;DR: The only other decent one is probably nerdfont's ;).

I wasn't necessarily going to bold the Backspace, but it behaves slightly less worse that way re: A2/ghosting on eInk when you tap it ;).

NiLuJe:
(gnu.org / savannah.org is down, hence the CI failures).

poire-z:
Not fond of the bold backspace (our orignal icon looked a bit airy'er and nicer to me, but ok).
Really nothing else for backspace in nerdfont?

(No problem with the bolded or unbolded enter key.)

NiLuJe:
I'll double-check, but, surprisingly, no. There's all the crappy Apple meta keys, but not that one -_-".

In case I don't find anything else, are you okay with that one unbolded?

NiLuJe:
... he said before remembering there was the Material design one.

Which is decidedly NOT airy ;p.

kb_mdi

NiLuJe:
The current MDI set *does* include an outline variant, though (https://materialdesignicons.com/), so I may be able to finagle something if that one's more amenable ;).

poire-z:
> In case I don't find anything else, are you okay with that one unbolded?

I'll be ok :) Unbolded prefered, but hava a go with bold it it avoids that ghosting issue.

> Which is decidedly NOT airy ;p.

Yep, quite too solid black (it looks like that battery icon knocked down! :)

NiLuJe:
Okay, that looks better with the outline variant ;).

kb_mdi_outline

poire-z:
Way better ! :+1:

Issue #5632, opened 2019-11-23
Various keyboard crashes and issues


poire-z:
* KOReader version: v2019.11-27-g4740ab1

Possibly related to recent refactoring (#5583, #5610), but might be present before.

There's some issue with the Japanese keyboard.

Hitting that yellow key on the emulator causes a crash:
<kbd>![image](https://user-images.githubusercontent.com/24273478/69483833-9935a100-0e2c-11ea-966f-2b7aea9af7ec.png)</kbd>
```
./luajit: frontend/ui/widget/textwidget.lua:135: Wrong text type (expected string)
stack traceback:
        [C]: in function 'assert'
        frontend/ui/widget/textwidget.lua:135: in function 'pre_guard'
        frontend/dbg.lua:41: in function 'updateSize'
        frontend/ui/widget/textwidget.lua:192: in function 'getSize'
        frontend/ui/widget/container/centercontainer.lua:10: in function 'paintTo'
        frontend/ui/widget/container/framecontainer.lua:96: in function 'paintTo'
        frontend/ui/widget/container/inputcontainer.lua:85: in function 'paintTo'
        frontend/ui/widget/horizontalgroup.lua:49: in function 'paintTo'
        frontend/ui/widget/verticalgroup.lua:48: in function 'paintTo'
        frontend/ui/widget/container/centercontainer.lua:24: in function 'paintTo'
        frontend/ui/widget/container/framecontainer.lua:96: in function 'paintTo'
        frontend/ui/widget/container/bottomcontainer.lua:18: in function 'paintTo'
        frontend/ui/widget/container/inputcontainer.lua:85: in function 'paintTo'
        frontend/ui/uimanager.lua:931: in function '_repaint'
```
But on my Kobo, when hitting that same key, I first get a japanese keyboard with all keys blank. Then, when I hit the screen, I get this crash (note the strange reference to the ko_KR keyboard!):
```
./luajit: ./ffi/util.lua:412: attempt to get length of local 'charstring' (a nil value)
stack traceback:
    ./ffi/util.lua:412: in function 'utf8charcode'
    frontend/ui/data/keyboardlayouts/ko_KR_helper.lua:148: in function 'in_consonnant_char'
    frontend/ui/data/keyboardlayouts/ko_KR_helper.lua:270: in function '_should_handle_as_target_char'
    frontend/ui/data/keyboardlayouts/ko_KR_helper.lua:253: in function 'process_char'
[...]

PR #5642, opened 2019-11-25
Keyboard layout change and Coverbrowser Mosaic tweaks

poire-z:
Avoid some of the issues noticed in #5632:
- Korean keyboard: unwrap InputText on layout change:
The korean keyboard wraps InputText and overrides some of its methods to process input in some specific way (details in #5053). When switching to another keyboard layout, the original methods need to be restored.

- Keyboard: properly handle keyboard layout height change
The japanese keyboard being taller than the others, when switching to/from it from/to another layout:
Re-init InputDialog for proper sizing and positionning
Refresh the whole screen, to remove any trace of a previous taller keyboard
Also add calls to :free() here and there to free keyboard keys' TextWidgets' XText C objects without waiting for GC.
(The `width = icon_height * 100` change is just because I was seeing some huge values in the log like a refresh on x=-1500..., and I figured it would just be cleaner to not see such huge values - possibly some small bug in ImageWidget. No issue noticed - and with #5639 it shouldn't be visited anymore).

Also included:
- CoverBrowser: Mosaic: only left-align for File browser
Left align partially filled rows for File browser/chooser, but have them centered for History and Favorites for a prettier look when only a few items are displayed. Followup to #5638.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5642)

### DIFF
--- a/frontend/ui/data/keyboardlayouts/ko_KR_keyboard.lua
+++ b/frontend/ui/data/keyboardlayouts/ko_KR_keyboard.lua
@@ -26,9 +26,21 @@ local wrapInputBox = function(inputbox)
     if inputbox._wrapped == nil then
         inputbox._wrapped = true
 
-        -- helper function
+        -- helper functions
+        local copied_names = {}
+        local function restore_func_references(obj)
+            for __, name in ipairs(copied_names) do
+                local orig_name = "_" .. name
+                if obj[orig_name] then
+                    obj[name] = obj[orig_name]
+                    obj[orig_name] = nil
+                end
+            end
+        end
+
         local function copy_func_reference(obj, name)
             obj["_" .. name] = obj[name]
+            table.insert(copied_names, name)
         end
 
         -- override original implementations with helper object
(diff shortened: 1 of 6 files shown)


poire-z:
Not strictly necessary. But given that:
- VirtualKeyboard is far from InputDialog (InputDialog > InputText > VirtualKeyboard), and we may have (now or one day) VirtualKeyboard work with some other widget than InputDialog
- VirtualKeyboard wants to refresh its area (the taller area of the previous or the next layout, at the bottom of the screen)
- InputDialog want to reposition its own title/scrolltextwidget/buttons, at the top of the screen - so they request the full screen refresh for different reason
- both `UIManager:setDirty("all", "flashui")` end up being merged into a single refresh

I figure it's clearer to have it done in both places.

poire-z:
Hold on: tested on my Kobo, and there's a small alignment issue when using a CenterContainer in the Mosaic view.

poire-z:
> small alignment issue when using a CenterContainer in the Mosaic view.

Fixed. (Having the width wrong doesn't matter when aligning to the left...)

PR #5643, opened 2019-11-25
Android fixes

### LINKED ISSUE #5621: Add /sdcard/fonts to Android
* KOReader version: v2019.12
* Device: Computer running Ubuntu Linux 18.04.3 or any Android device

#### Issue
Even with the option "System Fonts" enabled Koreader does not see all of the system fonts (i.e. user fonts installed on the system via fontconfig mechanism), namely it seems to ignore all the fonts installed in the user's `~/.fonts` directory. But this is where the best fonts (various optical families that user actually bought) are normally installed.

Oh btw, there is another little bug, clicking on "Open font folder" results in these errors:

```
nautilus: symbol lookup error: /usr/lib/x86_64-linux-gnu/libpangoft2-1.0.so.0: undefined symbol: hb_glib_get_unicode_funcs

```

As for Android, a similar problem happens here: the `fonts` subdirectory of internal storage which is used for user/system fonts on other platforms (and on Android by CoolReader and, maybe, ebookdroid) is ignored. Please either use this standard location or allow the user to set it manually.

### LINKED ISSUE #5631: System statistics freeze on Android
* KOReader version: 15fe0fe 2019-11-18
* Device: Android 6, Zenfone 2, x86

#### Issue
Opening system statistics freezes the program.

pazos:
Fixes #5631 
Fixes #5621 

Fonts can be placed in the following directories (under the primary storage partition): `fonts/` and `koreader/fonts`

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5643)

### DIFF
--- a/frontend/ui/elements/font_settings.lua
+++ b/frontend/ui/elements/font_settings.lua
@@ -61,10 +61,10 @@ function FontSettings:getAndroidPath()
     local user_path = android.getExternalStoragePath()
     if user_path == "Unknown" then
         -- unable to identify external storage. Use defaults
-        return system_path..";".."/sdcard/koreader/fonts"
+        return system_path..";".."/sdcard/fonts;/sdcard/koreader/fonts"
     else
         -- use the external storage identified by the app
-        return system_path..";"..user_path.."/koreader/fonts"
+        return system_path..";"..user_path.."/fonts;"..user_path.."/koreader/fonts"
     end
 end
 
--- a/plugins/systemstat.koplugin/main.lua
+++ b/plugins/systemstat.koplugin/main.lua
@@ -24,8 +24,6 @@ function SystemStat:init()
         self.storage_filter = "' /mnt/us$'"
     elseif Device:isSDL() then
         self.storage_filter = "/dev/sd"
-    elseif Device:isAndroid() then
-        self.storage_filter = Device.external_storage()
     end
 end
 


frontend/ui/elements/font_settings.lua
```diff
@@ -61,10 +61,10 @@ function FontSettings:getAndroidPath()
     local user_path = android.getExternalStoragePath()
     if user_path == "Unknown" then
         -- unable to identify external storage. Use defaults
-        return system_path..";".."/sdcard/koreader/fonts"
+        return system_path..";".."/sdcard/fonts;/sdcard/koreader/fonts"
```
poire-z: Logically, shouldn't /sdcard/koreader/fonts/ comes first, even before system_path?
(e.g. if a user needs to tweak some fonts, even system fonts, to have them work better with koreader (interline space, weight), while keeping that original same font available for other android apps?)

(I hope the fonts we ship are checked for before all those in these directories. Is it the case?)

pazos: The fonts we ship are stored in `/data/data/org.koreader.launcher/files`. These internal files cannot be modified by the user so we add a font dir under shared storage.

> Logically, shouldn't /sdcard/koreader/fonts/ comes first, even before system_path?
(e.g. if a user needs to tweak some fonts, even system fonts, to have them work better with koreader (interline space, weight), while keeping that original same font available for other android apps?)

Good point. I'm not a font expert and just add a few fonts myself that are not available as system fonts nor bundled. I'm not sure about the font loading mechanism and what happens if the same font is stored in a few different places.

poire-z: Well, we read the fonts in the directories in the order given here.
But at the end, we sort() them:
https://github.com/koreader/koreader/blob/0cbceca98a09e04783cb7fb176be8d72add501bf/frontend/fontlist.lua#L98-L105
And we feed that sorted set, in the sorted order, to crengine - which will ignore the duplicates (but crengine ignores the duplicate by font name, the internal name of the font - the filename doesn't matter).

So, it depends on the filename (if `table.sort()` keeps the original order for 2 identical elements - and if yes, good, the order of the directories would be preserved), and the internal name (to be or not considered a duplicate).

I guess a trick would be that if you want your font to be used first, wherever it is located: prefix its filename with "000"...

So, that order here is not 100% follproof, but it does still matter.

PR #5644, opened 2019-11-26
Remove DRCOUNTMAX

robert00s:
DRCOUNTMAX in Advanced settings is no longer needed.

```
-- number of page turns between full screen refresh
-- default to do a full refresh on every 6 page turns
--DRCOUNTMAX = 6
```
Full refresh can be modified in `Settings -> Screen -> E-ink settings -> Full refresh rate`

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5644)


poire-z:
More to come, to be added to this PR ? or you prefer one by one?

robert00s:
I plan to remove defaults completely. Some settings will be removed, some will be moved to the menu in different places and others will create new `advanced settings` with submenu.
It's a lot of work so I want to do it one by one.

poire-z:
OK.
But I don't know if it's really desirable for some settings.
For example:
https://github.com/koreader/koreader/blob/5a28cbd630f656c2aab1e47c51bce329be87948f/defaults.lua#L117-L127

I don't want just to be able to fine-adjust the current margins of my current book, and save that as **my one single** default - to use it on future books.
Currently, and I had done that for the example above and its buddies, I have changed them all to values that make more sense to me, so I have **9** defaults - but they are my defaults :) - that I can tap on the bottom menu to get any of them.

If I can't edit (copy to defaults.persistent.lua) a file to tweak them, and I have to go thru some UI (which you'll have to make very complicated to allow me that), it will be tedious.

(Personally, it won't hurt me much, as I patch nightlies anyway, I'll just have to patch them in their new places in the code - but still, I guess some people like being able to edit defaults.persistent.lua for such settings and some others.)

robert00s:
Thanks for your opinion. I'm not a big fan of defaults but if people want to use defaults file I will leave it. We back to it later.
One for sure. I want to rebuild advanced settings.

PR #5647, opened 2019-11-27
Dictionary font size

robert00s:
Add new option: Dictionary font size.
Before it was configurable through advanced settings.
Also DICT_FONT_SIZE is removed.

```
-- dictionary font size
--DDICT_FONT_SIZE = 20
```

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5647)

PR #5650, opened 2019-11-28
[fix] Crash dictionary font size in reader mode

robert00s:
See: https://github.com/koreader/koreader/pull/5647#issuecomment-559376164

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5650)


frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -302,7 +302,7 @@ If you'd like to change the order in which dictionaries are queried (and their r
                     local font_size = G_reader_settings:readSetting("dict_font_size") or 20
                     return T(_("Font size (%1)"), font_size)
                 end,
-                callback = function()
+                callback = function(instance)
```
poire-z: Can you call it `touchmenu_instance`, so we don't have to wonder _an instance of what?_ :)

PR #5655, opened 2019-11-28
bump crengine: word spacing and hyphenation tweaks

### LINKED ISSUE #5645: Word before a quotation mark will not be hyphenated
* KOReader version: v2019.11
* Device: PocketBook Inkpad 3 (PB740)

#### Issue
The first word in a line before a quotation mark is not hyphenated even if there is more than enough space (German hyphenation).
IMG_20191126_15112
IMG_20191126_150001915

poire-z:
Bump crengine, includes:
- New option to tune word spacing: space width scale percent https://github.com/koreader/crengine/pull/319
- Text: look for hyphenation in more words if needed https://github.com/koreader/crengine/pull/320 . Closes #5645 
- CSS: fix "hyphens:none" should override inherited "hyphens:auto"
- getHtml(): grab dir= and lang= attributes from upper nodes

It replaces our Word Gap/Space condensing toggle/setting with a new Word Spacing toggle/setting, made of 2 values:
- 1st number scales the normal width of spaces in all font (100% uses the font space width untouched)
- 2nd number applies after the 1st has been applied, and tells how much these spaces can additionally be condensed to make more text fit on a line.

(Had to use a new setting to avoid issues and allow downgrading - while migrating previous setting in the hope people won't notice the change - I wonder if I should use 95% instead of 100% to limit the cold reception of #5628 :)

Discussion in https://github.com/koreader/koreader/pull/5628#issuecomment-557811720 and https://github.com/koreader/crengine/pull/319

<kbd>image</kbd>

@robert00s : I had to tweak `optionutils.showValues()` to make it show that, and I cleaned / reorganized it a bit (I replaced your `suffix` with a more generic `show_true_value_func`) - if you can have a quick look if you see if I might have broken things, thanks.
Should fix:
<kbd>image</kbd>

 (I was tempted to add a `more_option` <kbd>⋮</kbd>, but I thought you'll have more fun than me implementing a DoubleSpinWidget in ConfigDialog :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5655)


poire-z:
(I changed `Word Gap` to `Word Spacing` for no real reason :) should i put back `Word Gap` to avoid the need to have a visible name translated? `Word Gap` is shared with PDF, `Word Spacing` rhymes with `Line Spacing` in a panel not far away.)

Frenzie:
Word spacing, line spacing, and letter spacing are the standard terms afaik (as also implemented in CSS properties). And as you said, it's internally inconsistent. I'd just make it `Word spacing`. :+1:

NiLuJe:
Oh, yeah, just realized that's one of the only instance where we have Title Cased labels ;).

robert00s:
@poire-z 
> I had to tweak optionutils.showValues() to make it show that, and I cleaned / reorganized it a bit (I replaced your suffix with a more generic show_true_value_func) - if you can have a quick look if you see if I might have broken things, thanks.

I know it too late but looks ok 👍 

> I was tempted to add a more_option ⋮, but I thought you'll have more fun than me implementing a DoubleSpinWidget in ConfigDialog :)

Yep :) I'll add this option in a few days.

PR #5657, opened 2019-11-29
fix continuous scroll & footer overlap

### LINKED ISSUE #5654: slim status bar crops content
* KOReader version: v2019.11.43-gb073896_2019-11-27
* Device: Kobo Aura H2O2

#### Issue
Slim status bar overlaps over the content of comic books ( cbz ) cropping off a part of the content

as you can see the speech bubble gets the text "I'll die" truncated by the status bar overlap.

This happens regardless of which zoom mode is selected, I would guess that status bar size is not accounted / added after the picture frame is processed and rendered

yparitcher:
closes  #5654

like #5620 but for continuous mode.

~~may fail CI because the page sizes change.~~

needs testing for side effects.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5657)

### DIFF
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -564,7 +564,7 @@ function ReaderPaging:onInitScrollPageStates(orig_mode)
         self.orig_page = self.current_page
         self.view.page_states = {}
         local blank_area = Geom:new{}
-        blank_area:setSizeTo(self.view.dimen)
+        blank_area:setSizeTo(self.view.visible_area)
         while blank_area.h > 0 do
             local offset = Geom:new{}
             -- caculate position in current page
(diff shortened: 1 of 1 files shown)


Frenzie:
@robert00s Could you take a look and test as well? It seems fine to me but I don't know all the ins and outs of the mini statusbar.

poire-z:
(I was willing to have a go at testing this, but I can't reproduce the issue before applying this PR :) I always get some overlap... Note that I'm not really familiar with our PDF/CBZ/ReaderPaging code and options...)

Anyway, should we merge this, and let users test it? :)

Frenzie:
Overlap is supposed to be there; tbh I never understand any of these issues because there's plenty. Scroll mode will be cut off for obvious reasons; paged won't. Usually you'll get a nice full line on the next page, sometimes you won't. Margin stuff is completely unrelated, but theoretically you could detect lines and make sure(r).

Frenzie:
Anyway, the "issue" is that some overlap is eaten by the status bar. But imo the issue is the reverse if anything.

Frenzie:
That is, the overlap is rather large, basically good with statusbar and oversized without. So I'd combine this PR with reduced overlap.

robert00s:
I'm also not familiar with CBZ ReaderPaging.
Like @poire-z said, let users test it :)

yparitcher:
the issue is not so much the overlap rather the the end of a "page" is cut off due to being under the footer.
with `page` view it is every page
with `continous view` it is only at the end of the document
the efect is more when using a width based zoom.
this pr should just put the status bar area out of the margin, so it is not used in calculating how much of the page to show.

PR #5658, opened 2019-11-29
Touch zones: fix loss of overrides when re-registering a zone

poire-z:
Happens with "Inverse reading order", which re-registers the Reader tap forward/backward zones and would prevent tap menu and bookmark from working.
Before this PR, when removing them, we would lose all the `override=` set on them by other touch zones (tap menu, bookmarks...), and so they were no more ensured after re-adding the zone.
So, make sure we don't lose that info. (Pinging @houqp for info, as DepGraph comes from you.)

This must be the proper fix for #5354, and revert a bit from #5423 (which made only the gestures set by ReaderGesture re-work - but not all others not managed by ReaderGesture like ReaderMenu Tap on top of screen). Pinging @yparitcher for info.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5658)

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -1577,11 +1577,6 @@ function ReaderGesture:onToggleReadingOrder()
     local document_module = self.ui.document.info.has_pages and self.ui.paging or self.ui.rolling
     document_module.inverse_reading_order = not document_module.inverse_reading_order
     document_module:setupTouchZones()
-    -- Needed to reset the touch zone overrides
-    local gesture_manager = G_reader_settings:readSetting(self.ges_mode)
-    for gesture, action in pairs(gesture_manager) do
-        self:setupGesture(gesture, action)
-    end
     UIManager:show(Notification:new{
         text = document_module.inverse_reading_order and _("RTL page turning.") or _("LTR page turning."),
         timeout = 2.5,
--- a/frontend/depgraph.lua
+++ b/frontend/depgraph.lua
@@ -40,7 +40,18 @@ function DepGraph:addNode(node_key, deps)
 end
 
 function DepGraph:removeNode(node_key)
-    self.nodes[node_key] = nil
+    -- We should not remove it from self.nodes if it has
+    -- a .deps array (it is the other nodes, that had this
+    -- one in their override=, that have added themselves in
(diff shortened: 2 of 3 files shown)


frontend/depgraph.lua
```diff
@@ -40,7 +40,19 @@ function DepGraph:addNode(node_key, deps)
 end
 
 function DepGraph:removeNode(node_key)
-    self.nodes[node_key] = nil
+    -- self.nodes[node_key] = nil
```
Frenzie: I don't see much point in leaving this line here?

spec/unit/depgraph_spec.lua
```diff
@@ -98,4 +98,43 @@ describe("DepGraph module", function()
             'readerhighlight_hold_release',
         }, dg:serialize())
     end)
+
+    it("should serialize complex graph and keep dependancies after remove and re-add", function()
```
Frenzie: I wouldn't have commented if that dependancies didn't stand out so much. :-P
```suggestion
    it("should serialize complex graph and keep dependencies after removing and re-adding", function()
```

spec/unit/depgraph_spec.lua
```diff
@@ -98,4 +98,43 @@ describe("DepGraph module", function()
...
+        local dg = DepGraph:new{}
+        dg:addNode('tap_backward')
```
Frenzie: There doesn't seem to be any reason to use single quotes?

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1577,11 +1577,6 @@ function ReaderGesture:onToggleReadingOrder()
     local document_module = self.ui.document.info.has_pages and self.ui.paging or self.ui.rolling
     document_module.inverse_reading_order = not document_module.inverse_reading_order
     document_module:setupTouchZones()
-    -- Needed to reset the touch zone overrides
```
Frenzie: So this part was the ugly hack?

poire-z: Well, just _a_ hack :) I don't think we ever did remove & re-add before, so that little bug in that trusted but forgotten bit of code was staying unnoticed.

yparitcher:
lgtm, i was fixing the symptom, i had not realized there was a deeper cause .

PR #5659, opened 2019-11-30
[UX] Simple symbol keyboard layer paging

Frenzie:
Here's a quick concept for consideration. Use 1/2 and 2/2 instead of Shift.

Screenshot_2019-11-30_16-01-28
Screenshot_2019-11-30_16-01-35

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5659)

### DIFF
--- a/frontend/ui/data/keyboardlayouts/en_keyboard.lua
+++ b/frontend/ui/data/keyboardlayouts/en_keyboard.lua
@@ -59,7 +59,7 @@ local _z_ = en_popup._z_
 return {
     min_layer = 1,
     max_layer = 8,
-    shiftmode_keys = {[""] = true},
+    shiftmode_keys = {[""] = true, ["1/2"] = true, ["2/2"] = true,},
     symbolmode_keys = {["Sym"] = true, ["ABC"] = true},
     utf8mode_keys = {["🌐"] = true},
     umlautmode_keys = {["Äéß"] = true},
@@ -91,7 +91,7 @@ return {
         },
         -- third row
         {  --  1       2       3       4       5       6       7       8
-            { label = "",
+            { "",   "",   "2/2",  "1/2",   "",   "",   "",    "",
               width = 1.5
             },
             { _Z_,    _z_,    "&",    "7",    "Á",    "á",    "Ű",    "ű", },


SilverGreen93:
I also renamed the Sym key to 123 in my romanian keyboard, and changed that to an alt key when you are in the second character page (see commit)

poire-z:
Well, before the highlighting of the Shift key, I would have liked it :)
But with the highlighting, it feels a bit less needed - and having <kbd>2/2</kbd> highlighted (<kbd>1/2</kbd> is not) feels strange/duplicated...
Also, I feel the <kbd>Shift</kbd> symbol less ambiguous. I could expect hitting <kbd>1/2</kbd> would output `1/2` :)  Because the <kbd>...</kbd> just above the <kbd>2/2</kbd> will output `...`, while I could as well expect it will move me to the ...next... layer :)

Anyway, just a thought. Go ahead if you prefer it that way.

NiLuJe:
I'd also tend to prefer a symbol over this, too.

I'm okay with Shift, as it's fairly universal, but we also have access to Apple's Command & Alt/Option glyph which might make sense in this context (the alt one, especially).

SilverGreen93:
The shift symbol here I can say it seems a bit ambiguous. On the main letter keyboard it makes sense, but on the symbols keyboard it should indicate something that "there are more symbols here". I would go with anything else BUT the shift symbol.

NiLuJe:
FWIW, I have no strong opinion on that subject. I'd *vaguely* prefer a symbol (because I also find that it's enough since the HL tweak), but that's it ;).

Frenzie:
I think any symbol would be _significantly_ less clear than a simple 1/2 and 2/2. Shift would at least be consistent and therefore better than anything trying to be too clever.

Frenzie:
I'm still in favor of this proposal. Has it grown on you guys by any chance @poire-z @NiLuJe? ;-)

poire-z:
Well, I didn't give it a chance to grow on me :) but as I get now that it applies only on the symbol/number layers, it's less drastic that I initially thought. So go ahead.

PR #5662, opened 2019-12-01
[UX] ru keyboard ё popup and translation

SilverGreen93:
Changes in this commit:
- Fix arrays to large (10 elements instead of 8) copied from the old keyboard system.  Issue introduced in #5610 
- Add ё popup.
- Change shift key as per #5659 (for both ro and ru keyboards)
- Translate Russian space and symbol keys.
- Rename Sym key to 123

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5662)


Frenzie:
Thanks!

PR #5664, opened 2019-12-03
android: allow overrides on 3rd party dictionary list

pazos:
After a conversation with @Norbi24 in https://www.mobileread.com/forums/showthread.php?t=325088 I finally implemented [his original suggestion](https://www.mobileread.com/forums/showpost.php?p=3922401&postcount=4)

Requires https://github.com/koreader/android-luajit-launcher/pull/200 for being able to launch an action with no specific package

I tested it placing the file "dictionaries.lua" inside the koreader folder with

```lua
return {
    { "Generic", "App picker", true, nil, "send" },
    { "ColorDict", "ColorDict", false, "com.socialnmobile.colordict", "colordict" },
}
```
If no package is specified the item must be true to avoid checking if the package is available.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5664)

### DIFF
--- a/frontend/device/android/dictionaries.lua
+++ b/frontend/device/android/dictionaries.lua
@@ -1,15 +1,19 @@
-return { --[[ supported android dictionary applications.
+local user_path = require("datastorage"):getDataDir() .. "/dictionaries.lua"
+local ok, dicts = pcall(dofile, user_path)
 
-Most of them should support Intent.ACTION_SEND, Intent.ACTION_SEARCH or
-Intent.ACTION_PROCESS_TEXT. Some applications implement their custom intents. ]]--
-
-    { "Aard2", "Aard2", false, "itkach.aard2", "aard2" },
-    { "Alpus", "Alpus", false, "com.ngcomputing.fora.android", "search" },
-    { "ColorDict", "ColorDict", false, "com.socialnmobile.colordict", "colordict" },
-    { "Fora", "Fora Dict", false, "com.ngc.fora", "search" },
-    { "GoldenFree", "GoldenDict Free", false, "mobi.goldendict.android.free", "send" },
-    { "GoldenPro", "GoldenDict Pro", false, "mobi.goldendict.android.pro", "send" },
-    { "Kiwix", "Kiwix", false, "org.kiwix.kiwixmobile", "text" },
-    { "Mdict", "Mdict", false, "cn.mdict", "send" },
-    { "QuickDic", "QuickDic", false, "de.reimardoeffinger.quickdic", "quickdic" },
-}
+if ok then
+    return dicts
+else
+    return {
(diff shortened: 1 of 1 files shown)


Markismus:
A bit of topic, but I was wondering whether these changes make it possible to do it other way around, too: Register Koreader's dictionary as a dictionary choice in other apps on Android?

In Onyx Boox e-reader you can write in epubs/PDFs in Neoreader, which makes it quite nice for studying. However, the dictionary support is not that good. Even when using the same dictionaries both for Koreader and Neoreader. With the coming update to 3.0 it will be possible to select an third party dictionary inside Neoreader. 

And I know that Pocketbook is severly lacking on that front, too. If Koreader's dictionary could be used independently from Koreader, that platform would get a boost, too.

Markismus:
Just read the dates, I am rather late to the party it seems. Sorry. (No idea how I got here. I thought I was checking new posts. :) )

pazos:
> A bit of topic, but I was wondering whether these changes make it possible to do it other way around, too: Register Koreader's dictionary as a dictionary choice in other apps on Android?
> 
> In Onyx Boox e-reader you can write in epubs/PDFs in Neoreader, which makes it quite nice for studying. However, the dictionary support is not that good. Even when using the same dictionaries both for Koreader and Neoreader. With the coming update to 3.0 it will be possible to select an third party dictionary inside Neoreader.
> 
> And I know that Pocketbook is severly lacking on that front, too. If Koreader's dictionary could be used independently from Koreader, that platform would get a boost, too.

@Markismus: in Android is feasible. I don't think so on Pockebook.

But, aside from android specific changes, the frontend needs to be adapted. This goes from effort to **huge** effort.

In the end, if something like this is implemented, I would expect that it would work on the desktop/emulator too. Basically `koreader --dict query` should start a new instance directly with the dict result in fullscreen mode. Closing that will close the program.

Then it would be a matter of adding the android specific bits, which are waay easier to implement.

Markismus:
Koreader didn't used to be so much code. :) 
Nowadays, I get lost whenever I want to hack something simple. (Doesn't help that I've haven't really programmed Lua in 4 years, too.)
If we reuse the widgets and just draw a window with the output from sdcv, we of course lose the possibility of looking up another dictionary or another word. 
And your mentioning of full screen stresses the fact that it is also not the only mode that would be wanted.
Still, it would be an nice way to get back into both Koreader and Lua.

pazos:
> And your mentioning of full screen stresses the fact that it is also not the only mode that would be wanted.

That's the only hard limit of NativeActivity. A fullscreen window. Things like overlays are unfeasible, if not impossible. The rest of stuff is a matter of playing with KO widgets until you get a result that works for you.

But yeah, it would be better and probably less effort to write an android app that uses sdcv directly.

PR #5667, opened 2019-12-03
UI mirroring for RTL languages (Arabic, Farsi, Hebrew...)

### LINKED ISSUE #2936: No support for font variants in UI based on language
* KOReader version: current
* Device: all

#### Issue

There is currently no way to specify different UI fonts, both default and fallback, based on the user's selected language.  This isn't a major issue, but for languages such as Chinese, Japanese, and Korean, the appearance of the same unicode characters can vary noticeably based solely on which language a given font is designed for (see example below).

Although I am less familiar with the details, apparently this is an issue between different variations of Western-language fonts as well.  For example, the English (top) and Romanian (bottom) versions of Veranda render the same glyphs differently:

Ideally, the correct fonts would be automatically selected based on the user's chosen language.

#### Steps to reproduce

Set KOReader language to Japanese and restart.  The UI language is now Japanese, but the rendered glyphs are the Chinese variants.

#### Proposed solution

Make all the fonts in `frontend/ui/font.lua` translatable so that custom fonts can be selected on a per-language basis.  Then, for example, the Japanese "translation" of the fallback font "`noto/NotoSansCJK-Regular.ttf`" could be an alternate font such as "`noto/NotoSansCJKjp-Regular.otf`" which has the appropriate Japanese versions of the necessary glyphs.  Of course, the appropriate font would also need to be added to the repository.

### LINKED ISSUE #5359: Right to left support (Arabic, Hebrew, ...)
* Device: kobo aura one 

#### Issue

Duplicate topic but I wrote for the importance
Will we see support for Arabic soon?
In the keyboard
And in writing the names of files and folders
Finally in the menus and settings
Thank you very much for this wonderful support and this extraordinary effort

poire-z:
Follow up to #5598 (_TextWidget/TextBoxWidget: enhanced text shaping_) that plugs the UI language and its text direction as defaults into the XText C module, which should be all that's needed to have RTL text right aligned and drawn correctly. (Should also close #2936 _No support for font variants in UI based on language_).

Also adds UI mirroring support in many widgets and apps (many files modified, but it's usually just a few lines in each of them).
Should finally close #5359 when the full arabic translation is in to have a nice mirrored UI in arabic (pinging @WaseemAlkurdi :)

@Frenzie @NiLuJe @robert00s 
The 2nd commit defines some helpers in the bidi.lua module, that are used by stuff in the 3rd commit.
Can you have a quick look at the function names defined in the 2nd commit, if they are obvious/concise/short enough, and if the way they are used in the 3rd commit looks natural and non-obtrusive enough? And if it's worthwhile introducing all this little noise to get a RTL UI none of us will ever use :)

I may split the 3rd commit in multiple smallers ones (low-level widgets, symbols mirroring, gesture mirroring, geometry arithmetic mirroring, textboxwidget parameters forwarding...), but it will be easier to do that while all is still flat :)
So some quick OK or comments on the BD API are appreciated :)

This mirrors all the UI elements I met that felt they needed mirroring.
The simple stuff in the low level widgets (*Container, HorizontalGroup, OverlapGroup, FrameContainer) was enough to make 90% of the stuff mirrored naturally - and we can just continue developping in LTR - and things should work in RTL. There are just some special cases with symbols, swipe directions, and some direct geometry arithmetic or direct painting, that need additional care.

What's left to do is some additional `BD.wrap()`, `BD.filename()`, `BD.path()` where needed, where they are give as parameters to some `T(L())`, but that's probably for another PR to not make this one even larger.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5667)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -399,6 +399,37 @@ function FileManagerMenu:setUpdateItemTable()
             })
         end,
     })
+    table.insert(self.menu_items.developer_options.sub_item_table, {
+        text = "UI layout mirroring and text direction",
+        sub_item_table = {
+            {
+                text = _("Reverse UI layout mirroring"),
+                checked_func = function()
+                    return G_reader_settings:isTrue("dev_reverse_ui_layout_mirroring")
+                end,
+                callback = function()
+                    G_reader_settings:flipNilOrFalse("dev_reverse_ui_layout_mirroring")
+                    local InfoMessage = require("ui/widget/infomessage")
+                    UIManager:show(InfoMessage:new{
+                        text = _("This will take effect on next restart."),
+                    })
+                end
+            },
+            {
+                text = _("Reverse UI text direction"),
+                checked_func = function()
+                    return G_reader_settings:isTrue("dev_reverse_ui_text_direction")
+                end,
+                callback = function()
+                    G_reader_settings:flipNilOrFalse("dev_reverse_ui_text_direction")
+                    local InfoMessage = require("ui/widget/infomessage")
+                    UIManager:show(InfoMessage:new{
+                        text = _("This will take effect on next restart."),
+                    })
+                end
+            }
+        }
+    })
 
     self.menu_items.cloud_storage = {
         text = _("Cloud storage"),
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -1,3 +1,4 @@
+local BD = require("ui/bidi")
 local CenterContainer = require("ui/widget/container/centercontainer")
 local CloudStorage = require("apps/cloudstorage/cloudstorage")
 local ConfirmBox = require("ui/widget/confirmbox")
@@ -611,10 +612,10 @@ function FileManagerMenu:_getTabIndexFromLocation(ges)
         return last_tab_index
     -- if the start position is far right
     elseif ges.pos.x > 2 * Screen:getWidth() / 3 then
-        return #self.tab_item_table
+        return BD.mirroredUILayout() and 1 or #self.tab_item_table
     -- if the start position is far left
     elseif ges.pos.x < Screen:getWidth() / 3 then
-        return 1
+        return BD.mirroredUILayout() and #self.tab_item_table or 1
     -- if center return the last index
     else
         return last_tab_index
--- a/frontend/ui/bidi.lua
+++ b/frontend/ui/bidi.lua
@@ -1,0 +1,189 @@
+--[[--
+Bidirectional text and UI mirroring setup and helpers.
+
+There are 2 concepts we attempt to handle:
+- Text direction: Left-To-Right (LTR) or Right-To-Left (RTL)
+- UI elements mirroring: not-mirrored, or mirrored
+
+These 2 concepts are somehow orthogonal to each other in
+their implementation, even if in the real world there are
+only 2 valid combinations:
+- LTR and not-mirrored: for western languages, CJK, Indic...
+- RTL and mirrored: for Arabic, Hebrew, Farsi and a few others.
+
+Text direction is handled by the libkoreader-xtext.so C module,
+and the TextWidget and TextBoxWidget widgets that handle text
+aligment. We just need here to set the default global paragraph
+direction (that widgets can override if needed).
+
+UI mirroring is to be handled by our widget themselves, with the
+help of a few functions defined here.
+
+Fortunately, low level widgets like LeftContainer, RightContainer,
+FrameContainer, HorizontalGroup, OverlapGroup... will do most of
+the work.
+But some care must be taken in other widgets and apps when:
+- some arrow symbols are used (for next, previous, first, last...):
+  they might need to be swapped, or some alternative symbols or
+  images can be used.
+- some geometry arithmetic is done (e.g. detecting if a tap is on the
+  right part of screen, to go forward), which need to be adapted/reversed.
+- handling left or right swipe, whose action might need to be reversed
+- some TextBoxWidget/InputText might need to be forced to be LTR (when
+  showing HTML or CSS code, or entering URLs, path...)
+
+Some overview at:
+https://material.io/design/usability/bidirectionality.html
+]]
+
+local Language = require("ui/language")
+local _ = require("gettext")
+
+local Bidi = {
+    _mirrored_ui_layout = false,
+    _rtl_ui_text = false,
+}
+
+-- Setup UI mirroring and RTL text for UI language
+function Bidi.setup(lang)
+    local is_rtl = Language:isLanguageRTL(lang)
+    -- Mirror UI if language is RTL
+    Bidi._mirrored_ui_layout = is_rtl
+    -- Unless requested not to (or requested mirroring with LTR language)
+    if G_reader_settings:isTrue("dev_reverse_ui_layout_mirroring") then
+        Bidi._mirrored_ui_layout = not Bidi._mirrored_ui_layout
+    end
+    -- Xtext default language and direction
+    if G_reader_settings:nilOrTrue("use_xtext") then
+        local xtext = require("libs/libkoreader-xtext")
+
+        -- Text direction should normally not follow ui mirroring
+        -- lang override (so that Arabic is still right aligned
+        -- when one wants the UI layout LTR). But allow it to
+        -- be independantly reversed (for testing UI mirroring
+        -- with english text right aligned).
+        if G_reader_settings:isTrue("dev_reverse_ui_text_direction") then
+            is_rtl = not is_rtl
+        end
+        Bidi._rtl_ui_text = is_rtl
+        xtext.setDefaultParaDirection(is_rtl)
+
+        -- Text language: this helps picking localized glyphs from the
+        -- font (eg. ideographs shaped differently for Japanese vs
+        -- Simplified Chinese vs Traditional Chinese).
+        -- Allow overriding xtext language rules from main UI language
+        -- (eg. English UI, with French line breaking rules)
+        local alt_lang = G_reader_settings:readSetting("xtext_alt_lang") or lang
+        if alt_lang then
+            xtext.setDefaultLang(alt_lang)
+        end
+    end
+    -- Optimise Bidi.default and Bidi.wrap by aliasing them to the right wrappers
+    if Bidi._rtl_ui_text then
+        Bidi.default = Bidi.rtl
+        Bidi.wrap = Bidi.rtl
+    else
+        Bidi.default = Bidi.ltr
+        Bidi.wrap = Bidi.noop
+    end
+end
+
+
+-- Use this function in widgets to check if UI elements mirroring
+-- is to be done
+function Bidi.mirroredUILayout()
+    return Bidi._mirrored_ui_layout
+end
+
+-- This function might only be useful in some rare cases (RTL text
+-- is handled directly by TextWidget and TextBoxWidget)
+function Bidi.rtlUIText()
+    return Bidi._rtl_ui_text
+end
+
+-- Small helper to mirror gesture directions
+local mirrored_directions = {
+    east = "west",
+    west = "east",
+    northeast = "northwest",
+    northwest = "northeast",
+    southeast = "southwest",
+    southwest = "southeast",
+}
+
+function Bidi.flipDirectionIfMirroredUILayout(direction)
+    if Bidi._mirrored_ui_layout then
+        return mirrored_directions[direction] or direction
+    end
+    return direction
+end
+
+function Bidi.flipIfMirroredUILayout(bool)
+    if Bidi._mirrored_ui_layout then
+        return not bool
+    end
+    return bool
+end
+
+-- Wrap provided text with bidirectionality control characters, see:
+--   http://unicode.org/reports/tr9/#Markup_And_Formatting
+--   https://www.w3.org/International/questions/qa-bidi-unicode-controls.en
+--   https://www.w3.org/International/articles/inline-bidi-markup/
+-- This works only when use_xtext=true: these characters are used
+-- by FriBidi for correct char visual ordering, and later stripped
+-- by Harfbuzz.
+-- When use_xtext=false, these characters are considered as normal
+-- characters, and would be printed. Fortunately, most fonts know them
+-- and provide an invisible glyph of zero-width - except FreeSans and
+-- FreeSerif which provide a real glyph (a square with "LRI" inside)
+-- which would be an issue and would need stripping. But as these
+-- Free fonts are only used as fallback fonts, and the invisible glyphs
+-- will have been found in the previous fonts, we don't need to.
+local LRI = "\xE2\x81\xA6"     -- U+2066 LRI / LEFT-TO-RIGHT ISOLATE
+local RLI = "\xE2\x81\xA7"     -- U+2067 RLI / RIGHT-TO-LEFT ISOLATE
+local FSI = "\xE2\x81\xA8"     -- U+2068 FSI / FIRST STRONG ISOLATE
+local PDI = "\xE2\x81\xA9"     -- U+2069 PDI / POP DIRECTIONAL ISOLATE
+
+function Bidi.ltr(text)
+    return string.format("%s%s%s", LRI, text, PDI)
+end
+
+function Bidi.rtl(text) -- should hardly be needed
+    return string.format("%s%s%s", RLI, text, PDI)
+end
+
+function Bidi.auto(text) -- from first strong character
+    return string.format("%s%s%s", FSI, text, PDI)
+end
+
+function Bidi.default(text) -- default direction
+    return Bidi._rtl_ui_text and Bidi.rtl(text) or Bidi.ltr(text)
+end
+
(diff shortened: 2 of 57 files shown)


frontend/ui/bidi.lua
```diff
@@ -0,0 +1,157 @@
...
+local _ = require("gettext")
+
+local Bidi = {
+    _mirrored_ui_layout = false,
+    _rtl_ui_text = false,
+}
+
+-- Setup UI mirroring and RTL text for UI language
+function Bidi.setup(lang)
+    local is_rtl = Language:isLanguageRTL(lang)
```
Frenzie: I assume you didn't use `isRtlLanguage` because it looks a bit odd (as does `isRTLLanguage) but I'd still prefer it for easier comprehension.

frontend/ui/bidi.lua
```diff
@@ -0,0 +1,157 @@
...
+    -- Mirror UI if language is RTL
+    Bidi._mirrored_ui_layout = is_rtl
+    -- Unless requested not to (or requested mirroring with LTR language)
+    if G_reader_settings:isTrue("reverse_lang_ui_layout_mirroring") then
```
Frenzie: What about something like `ui_direction_rtl`?

frontend/ui/bidi.lua
```diff
@@ -0,0 +1,157 @@
...
+local mirrored_directions = {
+    east = "west",
+    west = "east",
+    northeast = "northwest",
+    northwest = "northeast",
+    southeast = "southwest",
+    southwest = "southeast",
+}
+
+function Bidi.mirrorDirectionIfMirroredUILayout(direction)
```
Frenzie: Um, what? :-)

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -399,6 +400,37 @@ function FileManagerMenu:setUpdateItemTable()
             })
         end,
     })
+    table.insert(self.menu_items.developer_options.sub_item_table, {
+        text = "UI layout mirroring and text direction",
+        sub_item_table = {
+            {
+                text = _("Reverse language natural UI layout mirroring"),
```
Frenzie: Just `Reverse UI layout mirroring`?

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -399,6 +400,37 @@ function FileManagerMenu:setUpdateItemTable()
...
+                callback = function()
+                    G_reader_settings:flipNilOrFalse("reverse_lang_ui_layout_mirroring")
+                    local InfoMessage = require("ui/widget/infomessage")
+                    UIManager:show(InfoMessage:new{
+                        text = _("This will take effect on next restart."),
+                    })
+                end
+            },
+            {
+                text = _("Reverse language natural UI text direction"),
```
Frenzie: Same.

frontend/ui/bidi.lua
poire-z: Yes, flat camelCAse is a a bit strange.
But also, it puts the accent on what's checked for at the end: IsThisLanguageRTL., while your proposal sounds like IsItSomeRtlLanguage (among others) - less obvious.

frontend/ui/bidi.lua
poire-z: To make things less ambiguous, I tried to always use "rtl" for text, and "mirrored" for UI layout.
These 2 things are quite orthogonal to each other (in the implementation, even if in the real world, they should be either both enabled or both disabled).
I'd like to keep that as separate as possible, so we can all distinquish RTL issues (text, text aligmnent) and Mirroring issues (widgets, gestures...), so also for the settings names.

frontend/ui/bidi.lua
poire-z: You should see where they are used.
But ok, `flipDirectionIfMirroredUILayout()` to be less heavy ? :)

frontend/ui/bidi.lua
Frenzie: That depends on if you think of it as more of a query or more of a boolean, I suppose. As a natural language question I'd agree with you, but I see it in line with isKobo and isTouchDevice.

```lua
if isKobo() then
if isRtlLanguage() then
if isLanguageRTL() then
```

frontend/ui/bidi.lua
Frenzie: Since flip is used elsewhere already it's probably better that way.

poire-z:
A 3mn watch with 2 examples, that might show/explain why/when we'll need to `BD.wrap()` individual bits that will be concatenated (like in our footers): https://www.youtube.com/watch?v=xpumLsaAWGw&t=1692

frontend/ui/bidi.lua
poire-z: But these Device: methods refer to the Device object they are methods of, so Device:isKobo() makes sense - a boolean/property.
Language here is a module, not an object, so Language:isLanguageRTL(lang) does not refer to a property/boolean of the Language - but it's a function that takes an argument and should tell if that argument is RTL or not - a query.
Anyway, it's only used at 2 places, so I don't mind a name that feel strange to me - if you still feel isRtlLanguage is preferable (`isRTL(lang)` could work, but it sound even more like a method than the 2 other longer solutions).

poire-z:
So, no veto on this?
I can go on with coloring all our code with stuff like that?
```lua
local BD = require("ui/bidi")
if BD.mirroredUILayout() then ... end
if BD.rtlUIText() then ... end -- very rare
rotation_angle = BD.mirroredUILayout() and 90 or 0,
local direction = BD.flipDirectionIfMirroredUILayout(ges.direction)
if BD.flipIfMirroredUILayout(ges_ev.pos.x < Screen:getWidth()/2) then ... end
BD.wrap(prefix)
```

Frenzie:
I haven't really had time to think more about the names but the principle always looked good to me. :-)

poire-z:
^ force-pushed just a rebase to master - can be ignored

poire-z:
Prepended 2 commits that fix some stuff not necessarily related to UI mirroring - but that actually showed up more often when mirroring.
Splitted the initial 3rd commit in 3 commits. People working on fixing mirroring issues might only need to look at the last for inspiration.

@Frenzie: may be for you to later think about, a comment in bidi.lua:
```lua
-- See at having GetText_mt.__call() wrap untranslated strings in Bidi.ltr()
-- so they are fully displayed LTR.
```

poire-z:
Some screenshots (with the current incomplete Arabic translation):

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

NiLuJe:
Looks fun :).

WaseemAlkurdi:
From what I see, it's gorgeous. Absolutely gorgeous. Pure gold.
Lots of people, at least people around me, are going to love this!

poire-z:
> @Frenzie: may be for you to later think about, a comment in bidi.lua:
> `See at having GetText_mt.__call() wrap untranslated strings in Bidi.ltr() so they are fully displayed LTR.`

Would you be OK with something like that:
```diff
--- a/frontend/gettext.lua
+++ b/frontend/gettext.lua
@@ -30,2 +30,3 @@ local GetText = {
     plural_default = "n != 1",
+    -- wrap_untranslated_noop = function(text) return text end,
 }
@@ -36,2 +37,5 @@ local GetText_mt = {

+GetText.wrap_untranslated_noop = function(text) return text end
+GetText.wrap_untranslated = GetText.wrap_untranslated_noop
+
 --[[--
@@ -50,3 +54,3 @@ Returns a translation.
 function GetText_mt.__call(gettext, msgid)
-    return gettext.translation[msgid] or msgid
+    return gettext.translation[msgid] or GetText.wrap_untranslated(msgid)
 end
@@ -293,5 +297,5 @@ function GetText_mt.__index.ngettext(msgid, msgid_plural, n)
     if plural == 0 then
-        return GetText.translation[msgid] and GetText.translation[msgid][plural] or msgid
+        return GetText.translation[msgid] and GetText.translation[msgid][plural] or GetText.wrap_untranslated(msgid)
     else
-        return GetText.translation[msgid] and GetText.translation[msgid][plural] or msgid_plural
+        return GetText.translation[msgid] and GetText.translation[msgid][plural] or GetText.wrap_untranslated(msgid_plural)
     end
@@ -326,5 +330,5 @@ function GetText_mt.__index.npgettext(msgctxt, msgid, msgid_plural, n)
     if plural == 0 then
-        return GetText.context[msgctxt] and GetText.context[msgctxt][msgid] and GetText.context[msgctxt][msgid][plural] or msgid
+        return GetText.context[msgctxt] and GetText.context[msgctxt][msgid] and GetText.context[msgctxt][msgid][plural] or GetText.wrap_untranslated(msgid)
     else
-        return GetText.context[msgctxt] and GetText.context[msgctxt][msgid] and GetText.context[msgctxt][msgid][plural] or msgid_plural
+        return GetText.context[msgctxt] and GetText.context[msgctxt][msgid] and GetText.context[msgctxt][msgid][plural] or GetText.wrap_untranslated(msgid_plural)
     end
@@ -359,3 +363,3 @@ See [gettext contexts](https://www.gnu.org/software/gettext/manual/html_node/Con
 function GetText_mt.__index.pgettext(msgctxt, msgid)
-    return GetText.context[msgctxt] and GetText.context[msgctxt][msgid] or msgid
+    return GetText.context[msgctxt] and GetText.context[msgctxt][msgid] or GetText.wrap_untranslated(msgid)
 end
--- a/frontend/ui/bidi.lua
+++ b/frontend/ui/bidi.lua
@@ -88,2 +88,15 @@ function Bidi.setup(lang)
     end
+    -- If RTL UI text, have untranslater string (so english) still rendered LTR
+    if Bidi._rtl_ui_text then
+        _.wrap_untranslated = function(text)
+            -- if true then return text end
+            lines = {}
+            for s in text:gmatch("[^\r\n]+") do
+                table.insert(lines, Bidi.ltr(s))
+            end
+            return table.concat(lines, "\n")
+        end
+    else
+        _.wrap_untranslated = _.wrap_untranslated_noop
+    end
 end
```
It would turn this (that we have currently):
<kbd>image</kbd>
into something a bit more readable (I guess, even for Arabic readers) until the thing is translated:
<kbd>image</kbd>

Frenzie:
Seems okay. It does add some minor complication in case of switching from our Lua GetText to GetText through FFI so it should also mention that somewhere in a comment.

WaseemAlkurdi:
@poire-z There's a slight problem here. In the first screenshot ("before"), things like bullet points started from right to left, as they should be. However, in the second screenshot ("after"), they seem to start from the left. Is that normal or a regression?

poire-z:
No regression for something that does not yet exist :)

That was just some idea to display yet-not-translated english text a bit more correctly, I thought.
For this multi-lines english text fully non-translated, with bullets and punctuations, it looks to me quite bogus when displayed in the RTL UI right aligned with the RTL text rules, with a comma just after the bullet, and punctuations out of place.
I thought it would be more readable displayed LTR if it is plain english, even in the RTL UI - but you tell me.
Also, it may jump more at your eyes (if it needed help to be noticed...) that it needs to be translated that way :)
But once it's translated (it it exists in the translation, even if you let it be plain english), it would not suffer this trick, and would be fully RTL.
But you tell me if that's a plus - or really not. If the "before" is totally fine with you, no need to follow up on my idea :)

WaseemAlkurdi:
> No regression for something that does not yet exist :)
That was just some idea to display yet-not-translated english text a bit more correctly, I thought.

Ah, I misunderstood the purpose then. I thought that this is going to be the case for both localized and English.

> For this multi-lines english text fully non-translated, with bullets and punctuations, it looks to me quite bogus when displayed in the RTL UI right aligned with the RTL text rules, with a comma just after the bullet, and punctuations out of place.
I thought it would be more readable displayed LTR if it is plain english, even in the RTL UI - but you tell me.

Agreed. 
But why not go all the way and LTR flow the whole untranslated dialog then? If there's a single RTL character in the dialog, then render it RTL. What do you think?

> Also, it may jump more at your eyes (if it needed help to be noticed...) that it needs to be translated that way :)
But once it's translated (it it exists in the translation, even if you let it be plain english), it would not suffer this trick, and would be fully RTL.
But you tell me if that's a plus - or really not. If the "before" is totally fine with you, no need to follow up on my idea :)

You're right ... the second variation is better, now that I understand the reasoning.
However, it does need to be LTR in that case, doesn't it? If it can't be, then it's alright.

poire-z:
> If there's a single RTL character in the dialog, then render it RTL

There are 4 or 5 RTL chars in the above screenshots, at the bottom :)

> But why not go all the way and LTR flow the whole untranslated dialog then? If there's a single RTL character in the dialog, then render it RTL. What do you think?

Because that would be quite too much overhead - and very complicated to do.

And we can have bits translated and bits untranslated concatenated in a same paragraph.
My idea was just to have the GetText module which, from the english string, returns the arabic string, and if none, the english string as is - just return in the 2nd case the english string wrapped in LTR-isolate.

The paragraph direction is now something quite global (RTL for arabic, hebrew.. LTR for most others), and changing a single TextBoxWidget (that would happen to display untranslated text) paragraph direction would need kludge in all the places that use it.

In the example above, there are 6 translatable strings (the title "Font hinting", the long description, "Current value", the translated value, "Default value", "not set"), , and some of them can be either translated or not translated. All of them displayed in a single TextBoxWidget, that doesn't know about the little bits that made out the full text.

WaseemAlkurdi:
> There are 4 or 5 RTL chars in the above screenshots, at the bottom :)

Six! ;-P
`ت` `ل` `ق` `ا` `ئ` `ي`

> and changing a single TextBoxWidget (that would happen to display untranslated text) paragraph direction would need kludge in all the places that use it.

Thought that this could be done by GetText (passing the text orientation along with the translated/untranslated text), but then again, I know nothing of Lua.

> All of them displayed in a single TextBoxWidget

This is the dealbreaker here. 
So if I understand correctly, `TextBoxWidget`s take only one value for paragraph direction and apply it throughout the text? 
Not something like a word processor where you can specify paragraph direction for each paragraph separately?
If it's the first scenario, then what we have at present is good.

poire-z:
> So if I understand correctly, TextBoxWidgets take only one value for paragraph direction and apply it throughout the text?

Correct.
There are [unicode bidi control characters](https://www.w3.org/International/questions/qa-bidi-unicode-controls) to specify some direction/isolation inside the text (that I use in this trick) - but they work mostly like HTML inline SPANs.
The HTML block direction is something that is specified/set outside of the text content, and it's what we set as parameters of each TextBoxWidget (see below).

So, for the HTML analogy: TextBoxWidget is a single outer `<DIV>`, that can have a `dir=` attribute (the TextBoxWidget parameters).
The text that it contains is inline, with the `\n` in it being like `<br/>`, that are just inline linefeeds.
And the unicode bidi control chars we can inject in the text are just `<SPAN>`s 

> Not something like a word processor where you can specify paragraph direction for each paragraph separately?

We can't specify paragraph direction for each line (as explained above), but we have the option to use "auto", like HTML `<div dir="auto">` (I think):
https://github.com/koreader/koreader/blob/ea67b9be77131ada5ddc1afa781f11ef934f76b5/frontend/ui/widget/textboxwidget.lua#L91-L99

But still, we use these defaults ([unless we specify other values](https://github.com/koreader/koreader/commit/d6d49a64a79bf07820651ee2418ed6d1c5a3ed59) - e.g. when displaying CSS or HTML, we force `para_direction=false` to get LTR - and the only place we use `auto_para_direction=true` is for dictionary results, as there's no way for us to know the language we'll get).
So most textboxes use the UI language direction, and I think that's what you want: if you have a some real english text in a line among other lines, like a filename, a url, or some number/code, or some remote error message in english we are reporting to the user, you want them still right-aligned in the arabic UI - you don't want `auto_para_direction=true` to align them to the left and have a twisted text box :)

Anyway, we can do pretty nice things if needed.
But in this edge/temporary case, while we wait for you and others to translate strings to RTL languages, I guess we'll be fine with that simple & quick trick of wrapping the string in LTR-isolate, mostly just for the punctuations to not fly to the other side.
(So, for a full line or paragraph, the untranslated string becomes a "isolated LTR fragment" (or if multilines, each line becomes an isolated LTR fragment), but still in a RTL line/paragraph - so it's right aligned.)

Btw, for later: if you have room in the arabic keyboard, I think some of these [unicode bidi control characters](https://www.w3.org/International/questions/qa-bidi-unicode-controls) could be set to some key, they can be useful, like in the case you described elsewhere that if you have to start an arabic line with an english word, you rethink your line so to not start with that english word :) The alternative would be to just start the line with a, invisible strong RTL character:

Character | Name | Code point | Equivalent markup | Comment
-- | -- | -- | -- | --
RLM | RIGHT-TO-LEFT MARK | U+200F | none | strongly typed RTL character

(May be that does even work in your word processor :)

Frenzie:
@poire-z 

> @Frenzie: may be for you to later think about, a comment in bidi.lua:

Thinking about it now, but already implemented (by you :-P) in #5694.

poire-z:
Yep, couldn't wait :)
(How come you came to think about that old comment? Cleaning up your mailbox ? :)

Frenzie:
Yup, it's actually an old e-mail I purposefully didn't delete at the time. ^_^

PR #5670, opened 2019-12-06
android: make system fonts toggable

pazos:
like on desktop linux, but without the menu entry to open fonts dir.

Wip: works fine on fresh installation (without the koreader folder or with the folder without further modifications) but crashes on my main device, which is actually is/was heavily modified.

I'll update the crash logs when I have a bit of time.

**Edit:** the main reason for this PR is to avoid *15* pages of noto fonts on my huawei tablet. Others will find this useful too, specially on generic, modern, phones/tablets.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5670)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -350,7 +350,6 @@ androidupdate: all
 	# assets are compressed manually and stored inside the APK.
 	cd $(INSTALL_DIR)/koreader && zip -r9 \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).zip * \
-		--exclude=*fonts/droid* \
 		--exclude=*resources/fonts* \
 		--exclude=*resources/icons/src* \
 		--exclude=*share/man* \
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -50,7 +50,7 @@ function ReaderFont:init()
     end
     -- build face_table for menu
     self.face_table = {}
-    if Device:isDesktop() then
+    if Device:isAndroid() or Device:isDesktop() then
         table.insert(self.face_table, require("ui/elements/font_settings"):getMenuTable())
     end
     local face_list = cre.getFontFaces()
--- a/frontend/fontlist.lua
+++ b/frontend/fontlist.lua
@@ -64,9 +64,7 @@ local function isInFontsBlacklist(f)
 end
 
 local function getExternalFontDir()
-    if CanvasContext.isAndroid() then
-        return require("frontend/ui/elements/font_settings"):getAndroidPath()
-    elseif CanvasContext.isDesktop() then
+    if CanvasContext.isAndroid() or CanvasContext.isDesktop() then
         return require("frontend/ui/elements/font_settings"):getPath()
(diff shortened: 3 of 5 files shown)


NiLuJe:
Oh, yeah, +1, I also have a bazillion Noto variants on my OnePlus phone ;).

pazos:
okey, the error happens if/when a system font was used for a document and now is disabled:

```adb logcat
12-07 11:32:44.131 12016 12036 E KOReader: [NativeThread]: failed to run script: frontend/ui/widget/textwidget.lua:68: attempt to index field 'face' (a nil value)
12-07 11:32:44.131 12016 12036 E KOReader: [NativeThread]: Stopping due to previous errors
```

It should affect desktop linux too, but I didn't test.

poire-z:
> a system font was used for a document
> textwidget.lua:68: attempt to index field 'face' (a nil value)

That's not for a document, but for text in the UI elements - which should use only the NotoSans & Free* fonts we define in font.lua - which we still ship on Android.
So, the question is whether disabling system fonts make KOReader not use its own shipped fonts - or if, in spite of this disabling, it would still look for them, see them before the shipped ones, and ignore the shipped ones, and end up not using them because they are disabled (haven't look at the code, just some naive thoughts).

(Just checked how crengine behave when the main and the fallback fonts previously set disappear: it does not crash, and seem to use some Noto font.)

pazos:
@poire-z: thanks for the info. I was stupid enough and didn't notice the error. This time with a debug build. As you can see the error is not related to noto, but with DroidSansFallback:

```adb logcat
12-07 15:34:28.887 14143 14163 V KOReader:  Found font: NotoSans-Regular.ttf in ./fonts/noto/NotoSans-Regular.ttf
12-07 15:34:28.899 14143 14163 V KOReader:  Found font: NotoSans-Regular.ttf in ./fonts/noto/NotoSans-Regular.ttf
12-07 15:34:28.899 14143 14163 V KOReader:  Found font: DroidSansMono.ttf in ./fonts/droid/DroidSansMono.ttf
12-07 15:34:28.901 14143 14163 V KOReader:  Found font: NotoSansCJKsc-Regular.otf in ./fonts/noto/NotoSansCJKsc-Regular.otf
12-07 15:34:28.905 14143 14163 E KOReader:  #! Font  DroidSansFallback.ttf  ( DroidSansFallback.ttf ) not supported:  ffi/freetype.lua:142: Failed to load font './fonts/DroidSansFallback.ttf', freetype error code: 1
```

Reverting https://github.com/koreader/koreader/commit/503a57b32badb09a43e2210643bb3ecadb707258 fixes the error

poire-z:
The alternative to ressurecting all the fonts we got rid of :) would be to still pick "our no more shipped" fonts from system fonts, even when system fonts are disabled - which might be a bit tedious to code :|
But as we ressurected all of them except that DroidSans (did we?), I guess it's fine just getting it back and forget about that tricky font picking stuff.

Frenzie:
Alright, let's just revert it. :-/ The Droid Sans one is pretty small anyway; it's big fat Noto I wanted to get rid of.

PR #3625, opened 2018-01-21
luasettings: backup previous file and auto-recover

poire-z:
Similar to what was done for docsettings in #3093 (will apply to settings.reader.lua, but also to plugins' own settings).
Had it since after my early MuPDF 1.12 kobo crashes, and it has recovered once when I had wrongly manually edited settings.reader.lua.

PR #5678, opened 2019-12-09
Settings: use fsync() for more robust setting files saving

poire-z:
Bump base for util.fsyncOpenedFile() and util.fsyncDirectory() https://github.com/koreader/koreader-base/pull/1021

Use these to force flush to storage device when saving luasetting, docsetting, and history.lua.
Also dont rotate them to .old until they are at least 60 seconds old.
Discussion in #3625.

(Unrelated, but I also updated Luadata so it does not output stuff when there is no file - I used to see a few lines as I got not custom multiswiple.lua defined.)

Also bump crengine: open (as text) small or empty files https://github.com/koreader/crengine/pull/322

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5678)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -43,12 +43,19 @@ local symbol_prefix = {
     letters = {
         time = nil,
         pages_left = "=>",
+        -- @translators This is the footer letter prefix for battery % remaining.
         battery = C_("FooterLetterPrefix", "B:"),
+        -- @translators This is the footer letter prefix for percentage read.
         percentage = C_("FooterLetterPrefix", "R:"),
+        -- @translators This is the footer letter prefix for book time to read.
         book_time_to_read = C_("FooterLetterPrefix", "TB:"),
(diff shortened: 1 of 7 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -43,12 +43,19 @@ local symbol_prefix = {
     letters = {
         time = nil,
         pages_left = "=>",
+        -- @translators This is the footer letter prefix for battery % remaining.
```
Frenzie: Looks good :)

poire-z:
@robert00s : can you confirm a nightly with this PR does not crash on Kindle?
I tested on Kobo and Android, but I can't test Kindles. @NiLuJe : do all Kindle on any firmware use that fuse layer you mentionned a few times?

NiLuJe:
Yup, all of 'em ;/.

robert00s:
@poire-z 
I tested yesterday on Kindle. No problem so far :)

PR #5679, opened 2019-12-10
New widget DoubleSpinWidget based on HyphenationLimits

robert00s:
More generic HyphenationLimits widget is now known as DoubleSpinWidget. TextBoxWidget was changed to TextWidget to prevent multiline in title above NumberPickerWidgets.

This is first step to create `more_option ⋮` for `Word Spacing` in ConfigDialog.
See: https://github.com/koreader/koreader/pull/5655#issue-346842175

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5679)

### DIFF
--- a/frontend/apps/reader/modules/readerhyphenation.lua
+++ b/frontend/apps/reader/modules/readerhyphenation.lua
@@ -39,13 +39,20 @@ function ReaderHyphenation:init()
             local alg_left_hyphen_min = hyph_settings.left_hyphen_min
             local alg_right_hyphen_min = hyph_settings.right_hyphen_min
             local hyph_limits_widget = DoubleSpinWidget:new{
+                -- Min (2) and max (10) values are enforced by crengine
                 left_value = G_reader_settings:readSetting("hyph_left_hyphen_min") or alg_left_hyphen_min or 2,
+                left_min = 1,
+                left_max = 10,
                 right_value = G_reader_settings:readSetting("hyph_right_hyphen_min") or alg_right_hyphen_min or 2,
+                right_min = 1,
+                right_max = 10,
                 left_default = alg_left_hyphen_min or 2,
                 right_default = alg_right_hyphen_min or 2,
                 -- let room on the widget sides so we can see
                 -- the hyphenation changes happening
                 width = Screen:getWidth() * 0.6,
+                default_values = true,
+                default_text = _("Use language defaults"),
                 title_text = _("Hyphenation limits"),
                 info_text = _([[
 Set minimum length before hyphenation occurs.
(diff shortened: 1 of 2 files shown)


frontend/ui/widget/doublespinwidget.lua
```diff
@@ -23,29 +23,30 @@ local WidgetContainer = require("ui/widget/container/widgetcontainer")
...
+local DoubleSpinWidget = InputContainer:new{
+    title_text = "",
     title_face = Font:getFace("x_smalltfont"),
+    info_text = "",
     width = nil,
     height = nil,
-    -- Min (2) and max (10) values are enforced by crengine
     left_min = 1,
     left_max = 10,
     left_value = 2,
```
poire-z: May be put back these values (and the comment) in readerhyphenation.lua.
And here, set defaults similar to the ones of (single)spinwidget, for coherency (even if no other widget will use these defaults).

frontend/ui/widget/doublespinwidget.lua
```diff
@@ -23,29 +23,30 @@ local WidgetContainer = require("ui/widget/container/widgetcontainer")
...
     left_default = nil,
+    left_text = _("Left"),
     right_min = 1,
     right_max = 10,
     right_value = 2,
     right_default = nil,
+    right_text = _("Right"),
+    button_default_text = _("Use language defaults"),
```
poire-z: This text belongs to readerhyphenation, it's really not a good default for DoubleSpinWidget :)
Either a more generic `_("Use defaults")`, or make that additional button optional if that value is nil (and use nil as the default).

robert00s:
@poire-z 
Thanks for review :)

frontend/apps/reader/modules/readerhyphenation.lua
```diff
@@ -39,13 +39,20 @@ function ReaderHyphenation:init()
             local alg_left_hyphen_min = hyph_settings.left_hyphen_min
             local alg_right_hyphen_min = hyph_settings.right_hyphen_min
             local hyph_limits_widget = DoubleSpinWidget:new{
+                -- Min (2) and max (10) values are enforced by crengine
```
poire-z: OK, I must have written that false comment before some change I made to crengine :)
```c
// min value supported by algorithms is 1 (max is arbitrary 10)
// value enforced by algorithm previously was 2, so it's the default
#define HYPH_DEFAULT_HYPHEN_MIN 2
#define HYPH_MIN_HYPHEN_MIN 1
#define HYPH_MAX_HYPHEN_MIN 10
```
Can you please correct it, and just change `Min (2)` to `Min (1)`, so it's coherent with the `left_min = 1, right_min = 1,`?

poire-z:
(@robert00s : please see https://github.com/koreader/koreader/pull/5678#issuecomment-564477482 if you missed it)

PR #5680, opened 2019-12-10
android: unbreak KO on scoped storage devices

pazos:
Requires https://github.com/koreader/android-luajit-launcher/pull/201

Tested against the Android 10 - emulator

Build with targetApi 28: normal storage, RW path retrieved: `/storage/emulated/0`
Build with targetApi 29: scoped storage, RW path retrieved: `/storage/emulated/0/Android/data/org.koreader.launcher/files`
 
This isn't intended as a long term solution but as a workaround. Lets see what other reading software do to avoid scoped storage limitations.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5680)

### DIFF
--- a/datastorage.lua
+++ b/datastorage.lua
@@ -11,7 +11,7 @@ function DataStorage:getDataDir()
     if data_dir then return data_dir end
 
     if isAndroid then
-        data_dir = android.externalStorage() .. "/koreader"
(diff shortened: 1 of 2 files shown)


pazos:
The bumped launcher module adds a couple of new nooks as known eink devices too.

PR #5685, opened 2019-12-12
Ensure we flush everything to disk after an update.

NiLuJe:
Will stall for longer, but with a visible message (vs. less stalling on
a white screen during the startup and a laggy FM for a while).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5685)

### DIFF
--- a/platform/cervantes/koreader.sh
+++ b/platform/cervantes/koreader.sh
@@ -34,6 +34,8 @@ ko_update_check() {
         fi
         rm -f "${NEWUPDATE}" # always purge newupdate in all cases to prevent update loop
         unset BLOCKS CPOINTS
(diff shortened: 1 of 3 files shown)


NiLuJe:
Only done one devices where we have visual feedback during an OTA, otherwise it's a random weird stall with no visible warning ;).

NiLuJe:
Let's see how annoying this is in practice ;).

ptrm:
wow, so this lack of sync might explain why I had most of koreader files 0b in size when hard rebooting just after an upgrade :D

Just wonder if it would work in a way of creating empty files and writing the contents later on sync, or would there be no files at all.

NiLuJe:
That's probably entering the territory of "undefined behavior", so, who knows ^^.

PR #5686, opened 2019-12-13
android: import from Storage Access Framework

pazos:
Add an option to the filemanager "plus" menu to import files in current path using Android storage access framework.

The new menu entry will be displayed on Android 4.4+. Once the user select one or more files in the dialog these files will be imported on the specified folder.

Only supported mimetypes can be imported.

On folders *outside* the file sandbox the "import files here" entry will be disabled.

saf (1)

**Edit**: updated description and sample

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5686)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -83,6 +83,10 @@ local Device = Generic:new{
         if not link or type(link) ~= "string" then return end
         return android.openLink(link) == 0
     end,
+    canImportFiles = function() return android.app.activity.sdkVersion >= 19 end,
+    importFile = function(path) android.importFile(path) end,
+    isValidPath = function(path) return android.isPathInsideSandbox(path) end,
+
     canExternalDictLookup = yes,
     getExternalDictLookupList = getExternalDicts,
     doExternalDictLookup = function (self, text, method, callback)
@@ -133,7 +137,6 @@ function Device:init()
                     external_dict_when_back_callback()
                     external_dict_when_back_callback = nil
                 end
-
                 local new_file = android.getIntent()
                 if new_file ~= nil and lfs.attributes(new_file, "mode") == "file" then
                     -- we cannot blit to a window here since we have no focus yet.
@@ -148,6 +151,20 @@ function Device:init()
                     UIManager:scheduleIn(0.2, function()
                         require("apps/reader/readerui"):doShowReader(new_file)
                     end)
+                else
+                    -- check if we're resuming from importing content.
+                    local content_path = android.getLastImportedPath()
+                    if content_path ~= nil then
+                        local FileManager = require("apps/filemanager/filemanager")
+                        local UIManager = require("ui/uimanager")
+                        UIManager:scheduleIn(0.5, function()
+                            if FileManager.instance then
+                                FileManager.instance:onRefresh()
+                            else
+                                FileManager:showFiles(content_path)
+                            end
+                        end)
+                    end
                 end
             end
         end,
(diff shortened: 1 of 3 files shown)


Frenzie:
Are there any other apps that do something similar that could be used for inspiration? Looks fine to me as a proof of concept anyway.

pazos:
To figure out what can be done with SAF: https://play.google.com/store/apps/details?id=com.ianhanniballake.localstorage&hl=es_419

To test content providers I'm using both Termux and Google Drive.

pazos:
on the SAF side of the things we can do a lot of things but that doesn't allow us to access the filesystem, so I'm out of ideas.

Probably if scoped storage is enforced we can use SAF to do/restore a backup of both ko settings and books, but for a couple of years we're safe with legacy storage.

pazos:
Okay, I simplified a bit the backend and the menu. It adds a "import files here" entry to the plus menu. Files will be imported directly on that folder. Everything works but I'm not able to refresh the folder view to include the new imported files. Not sure if I'm doing this wrong because I'm not familiarized enough with filemanager and filechooser code.

Maybe @poire-z has some hints.

poire-z:
We have a `self:onRefresh()` called a few lines above your filemanager.lua button, for "Paste".
But I don't know how your `Device.importFile(current_dir)` works, if it blocks and switches to another activity, and continue once it is done - or if once it's called, a `self:onRefresh()` added just after would be executed before files are imported. In the later case, you'll need to have that done in a callback passed to Device.importFile, so it's it when done that would call that callback. Something like:
```lua
                text = _("Import files here"),
                callback = function()
                    local current_dir = self.file_chooser.path
                    Device.importFile(current_dir, function()
                           self:onRefresh()
                    end)
                end,
```
Or in the first case:
```lua
                text = _("Import files here"),
                callback = function()
                    local current_dir = self.file_chooser.path
                    Device.importFile(current_dir)
                    self:onRefresh()
                end,
```

poire-z:
You could try just storing that callback in some variable `onresume_callback`, and just call if it it is not nil in onActivityResult.

pazos:
The code in https://github.com/koreader/koreader/blob/master/plugins/wallabag.koplugin/main.lua#L131-L135 and works fine with UIManager:ScheduleIn.

frontend/device/android/device.lua
```diff
@@ -148,6 +151,20 @@ function Device:init()
                     UIManager:scheduleIn(0.2, function()
                         require("apps/reader/readerui"):doShowReader(new_file)
                     end)
+                else
+                    -- check if we're resuming from importing content.
+                    local content_path = android.getLastImportedPath()
```
poire-z: I don't get much of the Android code part :) so just asking: does this lastImportedPath/getLastImportedPath() gets reset back to null/nil after this, so you don't get stuck getting in this if when you resume koreader later?

pazos: Sure, it is in https://github.com/koreader/android-luajit-launcher/pull/203/files#diff-272d98f2e8dcd4377fdca0ca2a53178aR193-R197

pazos:
@Frenzie: ready to be merged, except for:

> * Archs: F-droid provides a package for arm. They need a different version code for each apk. The version code will be changed in [koreader/android-luajit-launcher#203](https://github.com/koreader/android-luajit-launcher/pull/203) to add a first number to differenciate between archs. (1) for arm and (2) for x86. Maybe the fdroid script and/or the feed 'koreader-android-fdroid-latest ' need to be updated.

But I think I won't break anything on our end and the file https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/org.koreader.launcher.yml needs to be updated anyways to point to weblate instead of transifex translations.

Issue #5271, opened 2019-08-26
Koreader doesn't start after v2019.08-43


Eduardomb22:
After the last two updates, my koreader at kindle can't start. Anyone knows how to fix it? It works fine at v2019.08-39

              launching...
  _  _____  ____                _
 | |/ / _ \|  _ \ ___  __ _  __| | ___ _ __
 | ' / | | | |_) / _ \/ _` |/ _` |/ _ \ '__|
 | . \ |_| |  _ <  __/ (_| | (_| |  __/ |
 |_|\_\___/|_| \_\___|\__,_|\__,_|\___|_|

 It's a scroll... It's a codex... It's KOReader!

 [*] Current time: 08/26/19-12:38:32
 [*] Version: v2019.08-51-gba8099f_2019-08-25

./luajit: frontend/util.lua:7: loop or previous error loading module 'gettext'
stack traceback:
	[C]: in function 'require'
	frontend/util.lua:7: in main chunk
	[C]: in function 'require'
	frontend/gettext.lua:199: in function 'changeLang'
	frontend/gettext.lua:362: in main chunk
	[C]: in function 'require'
	./reader.lua:32: in main chunk
	[C]: at 0x00013ef9

Frenzie:
Duplicate of https://github.com/koreader/koreader/issues/5269. I just fixed that like an hour ago. ;-)

PR #5687, opened 2019-12-13
Adds Document> Auto-save book metadata sub-menu

poire-z:
Replaces `DAUTO_SAVE_PAGING_COUNT` in defaults.lua with a new option available in menu, that allows setting an interval in minutes instead of a page count.
Long discussion in #3625.
(Probably to be merged after 2019-12 release, because not yet tested on my device, and there's some long bit that won't be translated in time, and might need discussion :)

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5687)


poire-z:
Didn't get your last comment :/ Can you clarify? (_Obvious comment about min != pages is obvious_ )
(Or it's pointing to a line you didn't mean to?)

NiLuJe:
> Didn't get your last comment :/ Can you clarify? (_Obvious comment about min != pages is obvious_ )
> (Or it's pointing to a line you didn't mean to?)

That's because I'm an idiot and read that as if you were assigning to `SYNC_PAGING_COUNT` the value of the new settings, which is in minutes, not in pages.

Except you're not, you're assigning the now deprecated defaults, which will be nil on new installs, so everything's fine :D.

poire-z:
OK :)

> Sidebar: I'm slightly on the fence about keeping the whole "wear" thingy. First, it shouldn't be a huge issue in practice. Second: it's explicitly mentioned that it's a full rewrite (in fact, that was specifically why I went with "rewrite" instead of flush/write), so: duh. (I think that bit of techy business is mostly a known fact, and/or irrelevant in practice, much more so that FS shenanigans).

Well, I'm all for being less dramatic :)
But if that text is also a bit to cover our asses, and give as much info to the user so he can decide between the options, better to not let any hint out and assume the user knows stuff. And this will avoid the question "_why not every 1 minute?! I don't want to have to lose at min 5 minutes of changes!_".
(Tbh, since you brought that dicussion up, I was regretting that I had it save every 10 page turns for 3 years! All these useless writes... - once/15 mn feels somehow less heavy :)

And... why so much noise about that myth of FAT32 being shit ? :)
> That said, I've never managed to actually trash a FAT32 FS ;). https://github.com/koreader/koreader/pull/3625#issuecomment-565247901

I too never had any corruption or any loss of files or directory content - only 0-bytes files, or changes not reflected because of USB mount badly unmounted, or forgotten to...

So, _may exacerbate the potential for filesystem corruption and complete data loss after a hardware crash_, is that a myth or a know fact ? I don't remember having seen any issue mentionning such disaster.

May be more risk on Kindle with that fuse layer?

You haven't tell me your thought about my comment at https://github.com/koreader/koreader/pull/3625#issuecomment-565237355:
> Given that we do it at document close / FM opening and at suspend, times where more things happen, and so more chances of bad luck - than once every N minutes in a peaceful reading session :)

NiLuJe:
> OK :)
> 
> > Sidebar: I'm slightly on the fence about keeping the whole "wear" thingy. First, it shouldn't be a huge issue in practice. Second: it's explicitly mentioned that it's a full rewrite (in fact, that was specifically why I went with "rewrite" instead of flush/write), so: duh. (I think that bit of techy business is mostly a known fact, and/or irrelevant in practice, much more so that FS shenanigans).
> 
> Well, I'm all for being less dramatic :)
> But if that text is also a bit to cover our asses, and give as much info to the user so he can decide between the options, better to not let any hint out and assume the user knows stuff. And this will avoid the question "_why not every 1 minute?! I don't want to have to lose at min 5 minutes of changes!_".
> (Tbh, since you brought that dicussion up, I was regretting that I had it save every 10 page turns for 3 years! All these useless writes... - once/15 mn feels somehow less heavy :)

Fair enough ;).

> 
> And... why so much noise about that myth of FAT32 being shit ? :)
> 
> > That said, I've never managed to actually trash a FAT32 FS ;). [#3625 (comment)](https://github.com/koreader/koreader/pull/3625#issuecomment-565247901)
> 
> I too never had any corruption or any loss of files or directory content - only 0-bytes files, or changes not reflected because of USB mount badly unmounted, or forgotten to...
> 
> So, _may exacerbate the potential for filesystem corruption and complete data loss after a hardware crash_, is that a myth or a know fact ? I don't remember having seen any issue mentionning such disaster.
> 
> May be more risk on Kindle with that fuse layer?
> 

Oh, I definitely have murdered FAT32 thingies in the past because of it being terrible. Just not an eReader ;).

On Kindle things will usually be sneakier when things get awry, but the general idea is still "some data loss implied" ^^.

Note that by "complete data loss" may be overdoing it a bit, I meant complete in the sense *all of the modified content*, not *all the things*. Although, side-effects of a bad failure may kill random apparently unrelated things.


> You haven't tell me your thought about my comment at [#3625 (comment)](https://github.com/koreader/koreader/pull/3625#issuecomment-565237355):
> 
> > Given that we do it at document close / FM opening and at suspend, times where more things happen, and so more chances of bad luck - than once every N minutes in a peaceful reading session :)

That'd be fairly bad luck in any case ;).

poire-z:
> Should this also be hidden behind a readSettings check like checkAutoSaveSettings?

I don't think so: one can change the option from never > every N minutes without needing a restart, so better to update that timestamp in all case and have it accurate - a `os.time()` in onResume doesn't cost much - the alternative being some ugly added code in the menu callbacks.

Frenzie:
I've had FAT32 completely crash on a hard drive due to a power interruption. It was definitely complete data loss, except that since the actual files were still physically present I was able to retrieve them all with a data rescue program.

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -356,7 +356,7 @@ This sets how often to rewrite to disk global settings and book metadata, includ
 
 The normal behavior is to save those only when the document is closed, or your device suspended, or when exiting KOReader.
 
-Setting it to some interval may help prevent losing new settings/sidecar data after a software crash, but will cause more I/O writes the lower the interval is,  and may wear your storage media in the long run.]])
+Setting it to some interval may help prevent losing new settings/sidecar data after a software crash, but will cause more I/O writes the lower the interval is, and may wear out your storage media in the long run.]])
```
Frenzie: Maybe "slowly wear out" sounds less frightening?

poire-z:
Are we skipping 2019.12 ? Should we merge this PR, or should we still wait?

Frenzie:
It's still half a week. I'll probably put it out on the 30th. ;-)

AlanSP1:
Only thing that "bothers" me is abbreviation for minutes.

As you can see here: https://english.stackexchange.com/questions/125110/abbreviation-of-units-of-measure-hours-vs-minutes/125111

Usually it is just m, or something longer. Not mn, as per image. I would go with "5m", "15m", etc, it looks much better to me.

Frenzie:
Agreed, mn looks off. In my French dictionary (Petit Robert) it says it can be abbreviated as _min_ or _mn_, so I think we know where it came from. ;-)

poire-z:
D'accord :) Will fix that.

poire-z:
Currently is: `text = T(_("Every %1 mn"), value)`

`Every 15m` or `Every 15 m` (with a space) in english? I guess without, right?
Or the longer (we have room): `Every 15 minutes`

poire-z:
I'm not familiar with the GetText stuff for singularvs  plural, but if we go with the longer version, which should it be:
```diff
     if not value then
         text = _("Only on close, suspend and exit")
     else
-        text = T(_("Every %1 mn"), value)
+        -- (Min value below is 5 minutes, so we can keep that plural.)
+        text = T(_("Every %1 minutes"), value)
+
+        -- (Or we go fully grammaraware: - is that correct?)
+        text = T(N_("Each minute", "Every %1 minutes", value), value)
+            -- Or "Every minute" or "Every minutes" ?
     end
     return {
         text = text,
```
And how this is done for languages with multiple plurals? Even if I just put 2 alternatives in that english code, languages with more plural will get presented with the possibility to add more than 2 alternatives on weblate?

Frenzie:
With a space is preferred by most style guides and also favored by me.

> And how this is done for languages with multiple plurals? Even if I just put 2 alternatives in that english code, languages with more plural will get presented with the possibility to add more than 2 alternatives on weblate?

Exactly, how many plurals they get depends on the language. I can't think of a specific string otoh but Russian and Polish make good showcases for multiple plurals. Just search for `N_` in the code and then look it up on Weblate to see what that looks like if you're curious. ;-)

Frenzie:
Here's one, 1 item, %1 items:

Screenshot_2020-01-02_16-33-09

poire-z:
Thanks.
So, how do you want that one? a) `"Every %1 m"` or b) `N(..., "Every %1 minutes...")`
(And if b), should we now think that: as we can do b), there's no reason for a) as it may be not adapted to some languages?)
(You can do it if you want, would be quicker than co-piloting me :)

Frenzie:
What you're asking is why Mozilla's Fluent/Pontoon take the whole singular/plural business completely into the translation tool.

For us, `N_("Every 1 m", "Every %1 m")` (or minute/minutes if there's space) is probably the safer bet.

AlanSP1:
For me "15m" looks better then "15 m", but, don't mind that. About translation, not sure how things work in different languages (mine is very complex, one of hardest).

poire-z:
Ok, done in #5271 .
Using "minute" instead of "m" makes us not to have to decide between "15m" and "15 m" :)

PR #5693, opened 2019-12-16
bump crengine, fribidi, xtext - some CoverBrowser Mosaic tweaks

poire-z:
- bump crengine: (Upstream) DocX: fix paragraphs with single hyperlink https://github.com/koreader/crengine/pull/323
- bump fribidi to 1.0.8 (for nested isolates fix, https://github.com/koreader/koreader/issues/5359#issuecomment-565001148)
- xtext.makeLine(): return `no_allowed_break_met=true` when that happens, so we know the line breaks inside a word. https://github.com/koreader/koreader-base/pull/1022 https://github.com/koreader/koreader/pull/5598#issuecomment-565836669

Also CoverBrowser: Mosaic: better words wrap in text covers
- Update words wrapping tweaks for use_xtext: add zero-width spaces around dots and underscores to allow more wraps.
- Fix overlap of directory name and nbitems for long directory names.

Before:
<kbd>image</kbd>
<kbd>image</kbd>

After:
<kbd>image</kbd>
<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5693)

### DIFF
--- a/frontend/ui/widget/textboxwidget.lua
+++ b/frontend/ui/widget/textboxwidget.lua
@@ -304,6 +304,10 @@ function TextBoxWidget:_splitToLines()
             -- - Between a line end_offset= and the next line offset=, there may be only
             --   a single indice not included: the \n or the space that allowed the break.
             self.vertical_string_list[ln] = line
+            if line.no_allowed_break_met then
+                -- let the fact a long word was splitted be known
+                self.has_split_inside_word = true
+            end
             if line.hard_newline_at_eot and not line.next_start_offset then
                 -- Add an empty line to reprensent the \n at end of text
                 -- and allow positionning cursor after it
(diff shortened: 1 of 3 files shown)


poire-z:
Guess I'd better use utf8 hex codes for the zero width spaces... They're not visible in the diff, and probably to most people who use a modern text editor that supports utf8, unlike me :)

poire-z:
OK. The last dot before suffix trick:
`title = title:gsub("%.\xE2\x80\x8B(%w%w?%w?%w?%w?)$", "\xE2\x80\x8B.%1")`
works in History (where I tested on the emulator), but not in FileBrowser (where I tested on my Kobo), because... of the bidi isolate wrap for filename, which adds a 3-bytes "Pop Directional Isolate" at the end of the filename there, so the 5x %w I try to match failed because there's 3+3 or 4+3 bytes there...
I was beginning to think these Bidi wrapping were going to get bothering - but I just realized we wrap them too early (in getMenuText, where I added that cause it was convenient), and it should be delayed to the last moment, where it's given to a TextWidget or TextBoxWidget.
Anyway, should not prevent merging this. But I'll need another pass on Menu.lua and these CoverBrowser files (maybe tomorrow).
No real issue not having that clean for the release, but you tell me how much time I have :)

PR #5694, opened 2019-12-17
[RTL UI] Bidi wrapping tweaks

poire-z:
Some followup tweaks to #5667:
- Alias everything to `Bidi.nowrap()` when in LTR UI, as using LTR isolates seems uneeded when already LTR. This might avoid unforeseen side effects for LTR languages (safer for the coming up release, but we might then notice general wrapping issues only in RTL UI, so less quickly :)
- Better wrapping of RTL filename by using auto direction and LTR-isolating only the suffix so it's always on a side. That's option C in https://github.com/koreader/koreader/issues/5359#issuecomment-564767926, used in the screenshot below.
- menu.lua: handle `bidi_wrap_func` outside `getMenuText()`, which may be used in other contexts. See https://github.com/koreader/koreader/pull/5693#issuecomment-566240993.
- GetText/Bidi: wrap untranslated strings in LTR. See https://github.com/koreader/koreader/pull/5667#issuecomment-564491264
- `util.getFriendlySize()`: remove leading space, which will fix:
<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5694)

### DIFF
--- a/frontend/gettext.lua
+++ b/frontend/gettext.lua
@@ -34,6 +34,15 @@ local GetText_mt = {
     __index = {}
 }
 
+-- wrapUntranslated() will be overriden by bidi.lua when UI language is RTL,
+-- to wrap untranslated english strings as LTR-isolated segments.
+-- It should do nothing when the UI language is LTR.
+GetText.wrapUntranslated_nowrap = function(text) return text end
+GetText.wrapUntranslated = GetText.wrapUntranslated_nowrap
+-- Note: this won't be possible if we switch from our Lua GetText to
+-- GetText through FFI (but hopefully, RTL languages will be fully
+-- translated by then).
+
 --[[--
 Returns a translation.
 
@@ -48,7 +57,7 @@ Returns a translation.
     local translation = _("A meaningful message.")
 --]]
 function GetText_mt.__call(gettext, msgid)
-    return gettext.translation[msgid] or msgid
+    return gettext.translation[msgid] or gettext.wrapUntranslated(msgid)
 end
 
 local function c_escape(what)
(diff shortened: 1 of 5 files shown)


poire-z:
@Frenzie: I'll be removing the leading space and your _with minimum field width alignment_ from #3650 - as I checked we don't need this left padding anywhere (and the `%4.1f` was my code, and probably not intended and cut and pasted from somewhere else - and wrong anyway as it should have been `%5.1f` to have `123.4` and `1.2` right aligned similarly).
Or I can make it an option, even if we don't use it for now.

poire-z:
OK, adding it as an option. For now RTL UI will be as on the left, but we can switch to like on the right if prefered:
<kbd>image</kbd> vs <kbd>image</kbd>

frontend/gettext.lua
```diff
@@ -34,6 +34,15 @@ local GetText_mt = {
     __index = {}
 }
 
+-- wrapUntranslated() will be overriden by bidi.lua when UI language is RTL,
+-- to wrap untranslated english strings as LTR-isolated segments.
+-- It should do nothing when the UI language is LTR.
+GetText.wrapUntranslated_nowrap = function(text) return text end
+GetText.wrapUntranslated = GetText.wrapUntranslated_nowrap
+-- Note: this won't be possible if we switch from our Lua GetText to
```
Frenzie: I'm sure it should be possible, it's just sneakily slightly incompatible. ;-)

poire-z:
> edit: but with RTL, the KB/MB should be on the left of the number, so we'll be on the left aligned case, with no left padding needed

OK, they will. But they are not currently because of the added (by this PR) _wrap untranslated strings in LTR_. Once they are translated, they should be displayed like the following (but just for showing, I removed the GetText.untranslated > LTR, so now the N items in directory looks stranger):

<kbd>image</kbd>

<kbd>image</kbd>

Still to understand why the ellipsis when truncating is on the wrong side.

PR #5696, opened 2019-12-18
[RTL UI] Bidi-wrap filenames, paths, urls, metadata

### LINKED ISSUE #1426: File names in file browser do not respect RTL
I have some files where the filename is in Hebrew.
Using the latest Android release, in the file browser, those names are displayed backwards.

poire-z:
Follow up to #5667 and #5694. Discussion in https://github.com/koreader/koreader/issues/5359#issuecomment-566959691.

This also helps displaying correctly RTL and bidi filenames and directories in a LTR UI.
Not tested, may have typo and parens errors :)
The Bidi._path and co at the bottom of bidi.lua and the aliasing will need some tuning.

I hardcoded Bidi.ltr() for IP adresses and IP:port combo. Dunno if they need some other kind of wrapping, or if `192.0.0.1:433` could really be displayed `443:1.0.0.192` in RTL :)

Also fix UI mirroring of SortWidget that I forgot in #5667.
Closes #1426.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5696)

### DIFF
--- a/frontend/apps/cloudstorage/cloudstorage.lua
+++ b/frontend/apps/cloudstorage/cloudstorage.lua
@@ -1,3 +1,4 @@
+local BD = require("ui/bidi")
 local ButtonDialog = require("ui/widget/buttondialog")
 local ButtonDialogTitle = require("ui/widget/buttondialogtitle")
 local ConfirmBox = require("ui/widget/confirmbox")
@@ -343,7 +344,7 @@ end
 
 function CloudStorage:onMenuHold(item)
     if item.type == "folder_long_press" then
-        local title = T(_("Select this directory?\n\n%1"), item.url)
+        local title = T(_("Select this directory?\n\n%1"), BD.dirpath(item.url))
         local onConfirm = self.onConfirm
         local button_dialog
         button_dialog = ButtonDialogTitle:new{
@@ -524,7 +525,7 @@ function CloudStorage:synchronizeSettings(item)
     local dropbox_sync_folder = item.sync_source_folder or "not set"
     local local_sync_folder = item.sync_dest_folder or "not set"
     syn_dialog = ButtonDialogTitle:new {
-        title = T(_("Dropbox folder:\n%1\nLocal folder:\n%2"), dropbox_sync_folder, local_sync_folder),
+        title = T(_("Dropbox folder:\n%1\nLocal folder:\n%2"), BD.dirpath(dropbox_sync_folder), BD.dirpath(local_sync_folder)),
         title_align = "center",
         buttons = {
             {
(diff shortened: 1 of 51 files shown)


poire-z:
> Seems like that %1 - %2 should just be part of the string so that it can be freely positioned by the translator. Not just for RTL purposes but also so you can say something like:
Minimal size left: %1 Minimal size right: %2

Yes, but this one is a menu item text, and I think most of them showing some parameters have a fixed text on the left, and the the selected param/option on the right, either after a `:` or inside parenthesis.
Free positionning is fine in other contexts, but may be for menu, it's better to have a rather contant look.
That menu item may also have an alternate text: 
```
Left/right minimal sizes: 3 - 2
Left/right minimal sizes: language defaults
```
So, dunno if we should have 2 translatable strings, one with free positionnning, that could make them really different.

Frenzie:
Your translation app of choice will normally show you similar strings, so it should be a conscious choice to deviate. And in any case, there would be no deviation whatsoever if the translation were:

```
Right/left minimal sizes: %2 - %1
Right/left minimal sizes: language defaults
```

yparitcher:
looks good to me so far :)
i have not tested all the corner cases though

poire-z:
(Last force-push is just a rebase to master - no change related to this PR.)

poire-z:
Updated to fix the issue mentionned at https://github.com/koreader/koreader/issues/5359#issuecomment-567671590 - will need a base bump for https://github.com/koreader/koreader-base/pull/1025 too.
TextBoxWidget: fix some issue (line.width not updated when adding the ellipsis) which caused the ellipsis to not be displayed when right align, because we were shifting x too much (no issue in LTR where we left align).
BD.auto() for title, author(s).... some rework of CoverBrowser bits, that I haven't all tested - but well, they looks logical.
(Getting a bit fed up with Bidi...)

Frenzie:
Just take a break from bidi if it's getting too much. It's already working great if you ask me! ;-)

poire-z:
I just don't want a break! I want it behind me, done and gone forever ! :)
Should be over soon when this one is merged (except for the possible edge case bugs...).
Might still be some stuff to fix with time and durations, depending on how they should be in AR/HE.

poire-z:
OK, merging this one, so we get feedback and fixes sooner, and be done with it.

(I didn't do the replacements of the existing `file:match("([^/]+)$")` & co with `util.splitFilePathName()` - there were already too much changes.)

Possibly something to try to do: when displaying book metadata (title, authors...), provide the
language from the book metadata to TextWidgets/XText.

Issue #5656, opened 2019-11-28
Continous scroll mode adds an unnecessary black line to comic books


BrainDamage:
* KOReader version: v2019.11.43-gb073896_2019-11-27
* Device: Kobo Aura H2O2

#### Issue
Load a comic book archive ( cbr )
Set the scroll mode and zoom type to fit to width, there'll be horizontal black lines on page breaks, even if the original content didn't.

Since several comics have the format "long vertical strip", and get paginated over a fixed size, this produces ugly artefacts such as below

![break](https://user-images.githubusercontent.com/50474/69833484-717f7800-1234-11ea-9108-f38d6be4960b.png)

PR #5705, opened 2019-12-24
Option to change page gap in continuous mode

mustafa-001:
for request see https://github.com/koreader/koreader/issues/5656

Adds an option to select page gap in continuous mode. Option only shows up on documents that have numbered pages (.pdf, .cbz) because i think there is not much use to this option when reading arbitrarily paged documents. If seen necessary same logic can be duplicated for .epub files.

koptoptions.lua handles user selection and saving that selection
ReaderView:readSettings() handles reading previously selected option when opening book
ReaderVİew:onPageGapUpdate() renders changes when selection made

Page gap is same in both landscape and portrait modes, I think adding different settings for both is little unnecessary. 
The options for gap is 0, 8, 16 and 32, default one before was 8.

commit log
> 
> page_gap setting on opening a book
> if it does not exist defaults to 8
> -adds page_gap selector to koptoptions.lua,
> user can select 0, 8, 16, 32 via footer settings
> -adds :onPageGapUpdate event handler to ReaderView

page-1
page-gap2

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5705)

### DIFF
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -774,7 +774,7 @@ function ReaderView:onReadSettings(config)
     self.highlight.saved = config:readSetting("highlight") or {}
     self.page_overlap_style = config:readSetting("page_overlap_style") or G_reader_settings:readSetting("page_overlap_style") or "dim"
     self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or
-    G_reader_settings:readSetting("kopt_page_gap_height") or 8)
+        G_reader_settings:readSetting("kopt_page_gap_height") or 8)
 end
 
 function ReaderView:onPageUpdate(new_page_no)
@@ -838,6 +838,9 @@ function ReaderView:onSetViewMode(new_mode)
     return true
 end
 
+--Refresh after changing a variable done by koptoptions.lua since all of them
+--requires full screen refresh. If this handler used for changing page gap from
+--another source (eg. coptions.lua) triggering a redraw is needed.
 function ReaderView:onPageGapUpdate(page_gap)
     self.page_gap.height = page_gap
     return true
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -773,7 +773,7 @@ function ReaderView:onReadSettings(config)
     self.page_scroll = page_scroll == 1 and true or false
     self.highlight.saved = config:readSetting("highlight") or {}
     self.page_overlap_style = config:readSetting("page_overlap_style") or G_reader_settings:readSetting("page_overlap_style") or "dim"
-    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or 
+    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or
     G_reader_settings("kopt_page_gap_height") or 8)
 end
 
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -51,10 +51,6 @@ local ReaderView = OverlapGroup:extend{
     scroll_mode = "vertical",
     -- properties of the gap drawn between each page in scroll mode:
     page_gap = {
-        -- width in pixels (when scrolling horizontally)
-        width = Screen:scaleBySize(G_reader_settings:readSetting("page_gap_width") or 8),
-        -- height in pixels (when scrolling vertically)
-        height = Screen:scaleBySize(G_reader_settings:readSetting("page_gap_height") or 8),
         -- color (0 = white, 8 = gray, 15 = black)
         color = Blitbuffer.gray((G_reader_settings:readSetting("page_gap_color") or 8)/15),
     },
@@ -782,8 +778,7 @@ function ReaderView:onReadSettings(config)
     self.page_scroll = page_scroll == 1 and true or false
     self.highlight.saved = config:readSetting("highlight") or {}
     self.page_overlap_style = config:readSetting("page_overlap_style") or G_reader_settings:readSetting("page_overlap_style") or "dim"
-    self.page_gap.height = config:readSetting("kopt_page_gap") or 8
-    self.page_gap_width = config:readSetting("kopt_page_gap") or 8
+    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or 8)
 end
 
 function ReaderView:onPageUpdate(new_page_no)
@@ -849,7 +844,6 @@ end
 
 function ReaderView:onPageGapUpdate(page_gap)
     self.page_gap.height = page_gap
-    self.page_gap.width = page_gap
     return true
 end
 
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -48,7 +48,6 @@ local ReaderView = OverlapGroup:extend{
     page_scroll = nil,
     page_bgcolor = Blitbuffer.gray(DBACKGROUND_COLOR/15),
     page_states = {},
-    scroll_mode = "vertical",
     -- properties of the gap drawn between each page in scroll mode:
     page_gap = {
         -- color (0 = white, 8 = gray, 15 = black)
@@ -385,11 +384,7 @@ function ReaderView:getScrollPageRect(page, rect_p)
 end
 
 function ReaderView:drawPageGap(bb, x, y)
-    if self.scroll_mode == "vertical" then
-        bb:paintRect(x, y, self.dimen.w, self.page_gap.height, self.page_gap.color)
-    elseif self.scroll_mode == "horizontal" then
-        bb:paintRect(x, y, self.page_gap.width, self.dimen.h, self.page_gap.color)
-    end
+    bb:paintRect(x, y, self.dimen.w, self.page_gap.height, self.page_gap.color)
 end
 
 function ReaderView:drawSinglePage(bb, x, y)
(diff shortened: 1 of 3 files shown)


mustafa-001:
https://github.com/koreader/koreader/blob/bfe9cd8527db340ad73bcc0f8466df205081a133/frontend/ui/data/koptoptions.lua#L8-L11

Changes on this settings forces screen wide refresh, and ReaderView:painttto() draws everything again. Also I didn't quite figure out what to do other than full screen refresh thats already queued :). Any suggestions on proper way of doing this are welcomed.

NiLuJe:
Yeah, IIRC, koptsettings can force repaints without any specific events because... reasons? :D.

Frenzie:
It probably mostly makes sense to just redraw everything after changing a setting… a bit wasteful depending on what it is perhaps but I doubt it's bothersome.

So if that's not an issue, it seems fine to me.

frontend/apps/reader/modules/readerview.lua
```diff
@@ -782,6 +782,8 @@ function ReaderView:onReadSettings(config)
     self.page_scroll = page_scroll == 1 and true or false
     self.highlight.saved = config:readSetting("highlight") or {}
     self.page_overlap_style = config:readSetting("page_overlap_style") or G_reader_settings:readSetting("page_overlap_style") or "dim"
+    self.page_gap.height = config:readSetting("kopt_page_gap") or 8
+    self.page_gap_width = config:readSetting("kopt_page_gap") or 8
```
yparitcher: page gap is already defined here:
https://github.com/koreader/koreader/blob/7fb2db8f02897b8ff27f529f752f9a6150ca9939/frontend/apps/reader/modules/readerview.lua#L53-L59

so you might want to change this to:
```diff
-   self.page_gap.height = config:readSetting("kopt_page_gap") or 8
+   self.page_gap.height = config:readSetting("kopt_page_gap") or G_reader_settings:readSetting("page_gap_width") or 8
```
and not define it on line 55.

also there it uses `scaleBySize` instead of a direct number.

mustafa-001:
Aww, when trying to separate landscape and portrait page gap I just realized they are both calculated from `ReaderView.page_gap.height`, `page_gap.width` works in horizontal scrolling (left-right, which I think not much useful and currently not implemented. Maybe with few and very long lines?). 

So I'm removing width, moving height to :onreadSettings(), leaving page_gap.color only.
https://github.com/koreader/koreader/blob/7fb2db8f02897b8ff27f529f752f9a6150ca9939/frontend/apps/reader/modules/readerview.lua#L52-L60

frontend/apps/reader/modules/readerview.lua
```diff
@@ -782,8 +778,7 @@ function ReaderView:onReadSettings(config)
     self.page_scroll = page_scroll == 1 and true or false
     self.highlight.saved = config:readSetting("highlight") or {}
     self.page_overlap_style = config:readSetting("page_overlap_style") or G_reader_settings:readSetting("page_overlap_style") or "dim"
-    self.page_gap.height = config:readSetting("kopt_page_gap") or 8
-    self.page_gap_width = config:readSetting("kopt_page_gap") or 8
+    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or 8)
```
yparitcher: you might want
```diff
-   self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or 8)
+   self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or G_reader_settings:readSetting("page_gap_width") or 8)
```

frontend/apps/reader/modules/readerview.lua
```diff
@@ -51,10 +51,6 @@ local ReaderView = OverlapGroup:extend{
     scroll_mode = "vertical",
     -- properties of the gap drawn between each page in scroll mode:
     page_gap = {
-        -- width in pixels (when scrolling horizontally)
-        width = Screen:scaleBySize(G_reader_settings:readSetting("page_gap_width") or 8),
```
yparitcher: if you are getting rid of this you should also clean up
https://github.com/koreader/koreader/blob/7fb2db8f02897b8ff27f529f752f9a6150ca9939/frontend/apps/reader/modules/readerview.lua#L392-L396
and
https://github.com/koreader/koreader/blob/7fb2db8f02897b8ff27f529f752f9a6150ca9939/frontend/apps/reader/modules/readerview.lua#L51

does anyone know what `scroll_mode` is used for and if it is ever set to `horizontal`?

Frenzie: I renamed scroll mode to continuous in the GUI.

frontend/apps/reader/modules/readerview.lua
yparitcher: that should be 
`self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or G_reader_settings:readSetting("kopt_page_gap_hight") or 8)`

also this does not work with defaults (holding the option) with or without this change

frontend/apps/reader/modules/readerview.lua
yparitcher: @Frenzie 
`scroll_mode` or `page_scroll`?

frontend/apps/reader/modules/readerview.lua
yparitcher: sorry typo caused that.
to get defaults to work you need this
```diff
-    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or 8)
+    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or G_reader_settings:readSetting("kopt_page_gap_height") or 8)
```

frontend/apps/reader/modules/readerview.lua
mustafa-001: `page_scroll` is a boolean that holds `True` in continuous mode, `False` in page mode. `scroll_mode` set `vertical` in `ReaderView` creation and doesn't change.

Frenzie: That's part of the same thing, perhaps separated specifically for that vertical vs horizontal thing. I don't know the specific plan behind 63e5e7e577a0fb18b27f0b50400e38ef650b3a71

frontend/apps/reader/modules/readerview.lua
```diff
@@ -778,7 +773,8 @@ function ReaderView:onReadSettings(config)
     self.page_scroll = page_scroll == 1 and true or false
     self.highlight.saved = config:readSetting("highlight") or {}
     self.page_overlap_style = config:readSetting("page_overlap_style") or G_reader_settings:readSetting("page_overlap_style") or "dim"
-    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or 8)
+    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or
+    G_reader_settings("kopt_page_gap_height") or 8)
```
yparitcher: ``diff
-    G_reader_settings("kopt_page_gap_height") or 8)
+    G_reader_settings:readSetting("kopt_page_gap_height") or 8)
```

**Comment by yparitcher** [comment]
lgtm besides for the typo

**Comment by poire-z** [comment]
> Yeah, IIRC, kopt settings can force repaints without any specific event because... reasons? :D.

https://github.com/koreader/koreader/blob/c3a0bd4def2243b66cacf27222b62d9441ebbaff/frontend/ui/widget/configdialog.lua#L909-L913
Part of #4793, with some long discussion trying to understand how that works, that I'd rather not re-read :)
I guess that some options are only used by the kopt code, and there's no (or no need for) Lua code to handle some event, as there's no real action - except having kopt do its stuff with the updated settings.
For this one (page_gap.height), that's our Lua code that draws the gray line, so it looks allright to have an event and an handler (or we could just re-read the setting each time we draw a page, but no need to bother changing what's here I guess).

File: frontend/apps/reader/modules/readerview.lua [hunk 5]
```diff
@@ -782,6 +773,8 @@ function ReaderView:onReadSettings(config)
     self.page_scroll = page_scroll == 1 and true or false
     self.highlight.saved = config:readSetting("highlight") or {}
     self.page_overlap_style = config:readSetting("page_overlap_style") or G_reader_settings:readSetting("page_overlap_style") or "dim"
+    self.page_gap.height = Screen:scaleBySize(config:readSetting("kopt_page_gap_height") or
+    G_reader_settings:readSetting("kopt_page_gap_height") or 8)
```
poire-z: (nitpick: I'd like a bit more indentation on that 2nd line, to avoid confusion and make it clear it's continued from previous)

Frenzie:
___
*[frontend/ui/data/koptoptions.lua, line 88 at r6](https://reviewable.io/reviews/koreader/koreader/5705#-LxW_IaV8mfK-Iglr0KW:-LxW_IaW8ad5luR-7gbH:b-zc79rk) ([raw file](https://github.com/koreader/koreader/blob/455f1cdd67dd830d51b41c2dc29716d244e658b1/frontend/ui/data/koptoptions.lua#L88)):*
> ```Lua
>             {
>                 name = "page_gap_height",
>                 name_text = S.PAGE_GAP,
> ```

I'd just use the string here instead of adding it to that strings.lua file. For some reason those strings were globals back in 2013 but separating them from their surroundings just makes translation harder.

poire-z:
Is this ready to merge?
(I think @Frenzie last comment about `S.PAGE_GAP` should be adressed in a dedicated "chore" PR, where we'd do all of them in one batch.)

Frenzie:
If you have no further comments, sure. I haven't tested it.

poire-z:
> I haven't tested it.

Neither did I, but that's what users are for :)

I guess the Circle CI build (repeatedly) fails because of the transiflex>weblate move, and this PR base commit is not adapted to the CircleCI updated tests?

Frenzie:
Sort of, a minor accident on my part caused the submodule reference to become invalid.

My plan was simple, to update Weblate with the latest from Transifex when the time came, the end.

I did that, but it turned out Weblate had already attracted a few dozen new translations. So instead of overwriting I had to redo it by merging the two translations, and I semi-accidentally rebased that one commit. I wasn't fully cognizant of the fact that it'd change the commit hash if I just kept it in there.

tl;dr Now older commits referencing the submodule as it was can't find their reference.

Issue #5500, opened 2019-10-17
When chapters have subchapters


gerhaher:
#### Issue
When a chapter has subchapters the Mini progress bar has chapter markers only for the subchapters, but not for the chapter itself. 
For example in the book below there is a marker for 'Making everything gay' but not for '1 Gay'. 
And when using 'Next chapter'-gesture you jump directly from 'Introduction' to 'Making everything gay'.

The Full progress bar works the other way around: it leaves out the subchapters and instead has chapter markers for "1 Gay", "Interlude", "2 Women" etc.

Is this behaviour intentional? Would it be possible to change so that it takes account of both (chapter & subchapters).
Usually this is a nonissue because the chapter and its first subchapter are on the same page. But not so with this book as there is a lot of text before the first subchapters appear. 

![Reader_2019-Oct-17_140041](https://user-images.githubusercontent.com/32645430/67008442-c1bdd300-f0e9-11e9-84fa-bcc3bc454b44.png)
![Reader_2019-Oct-17_140111](https://user-images.githubusercontent.com/32645430/67008445-c5515a00-f0e9-11e9-94d2-71ecc5a5e828.png)

NiLuJe:
How does the Skim widget deal with this?

(I *think* it does both, but, as you said, it's usually a non-issue because of the same-page-iness ;p).

gerhaher:
> How does the Skim widget deal with this?
> 
> (I think it does both, but, as you said, it's usually a non-issue because of the same-page-iness ;p).

No, just the subchapter. For example, it jumps from 'Introduction' to 'Making Everything Gay'.

poire-z:
> Is this behaviour intentional?

I have no idea. It's like this for years, and it's kinda explicite (even if no comment...) to behave in this strange way: walk the TOC from the end, and ignore TOC entries with a lower level than one we have met:
https://github.com/koreader/koreader/blob/7de022a61742f42224a046b604202e3a8e150ed7/frontend/apps/reader/modules/readertoc.lua#L142-L158

Looking around, it's like this since https://github.com/koreader/koreader/pull/984. Explanation at https://github.com/koreader/koreader/issues/976#issuecomment-58483192.

Personnally, since a few years, I fix that in my own patch by just commenting line 153:
```diff
@@ -153,7 +153,7 @@
                 if v.depth >= depth then
                     table.insert(ticks, v.page)
                 end
-                depth = v.depth
[...]

PR #5708, opened 2019-12-28
Footer - add option chapter markers width

robert00s:
Move DMINIBAR_TOC_MARKER_WIDTH from Advanced settings to Status bar -> Progress bar -> Style  

- Thin (DMINIBAR_TOC_MARKER_WIDTH =1)
- Normal (DMINIBAR_TOC_MARKER_WIDTH =2)
- Thick (DMINIBAR_TOC_MARKER_WIDTH =3)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5708)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1160,10 +1165,58 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            self:refreshFooter(true)
                         end
                     },
+                    {
+                        text_func = function()
+                            local markers_width_text = _("thick")
+                            if self.settings.toc_markers_width == 1 then
+                                markers_width_text = _("thin")
+                            elseif self.settings.toc_markers_width == 2 then
+                                markers_width_text = _("normal")
```
Frenzie: I'd probably say either medium or default.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1160,10 +1165,58 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            end
+                            return T(_("Chapter markers width (%1)"), markers_width_text)
```
```suggestion
                            return T(_("Chapter marker width (%1)"), markers_width_text)
```

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1160,10 +1165,58 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                                    return self.settings.toc_markers_width == 1
+                                end,
+                                callback = function()
+                                    self.settings.toc_markers_width = Screen:scaleBySize(1)
+                                    self:setTocMarkers()
+                                    self:refreshFooter(true)
+                                end,
+                            },
+                            {
+                                text = _("Normal"),
```
Frenzie: medium/default?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1160,10 +1165,58 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            return not self.settings.progress_style_thin and self.settings.toc_markers
+                        end,
+                        sub_item_table = {
+                            {
+                                text = _("Thin"),
+                                checked_func = function()
+                                    return self.settings.toc_markers_width == 1
+                                end,
+                                callback = function()
+                                    self.settings.toc_markers_width = Screen:scaleBySize(1)
```
poire-z: Looks like this might work on an emulator where scaleBySize does x1 - but unless I miss something, it won't be checked (or worse) when scaleBySize does x1.7 or x2.
So, may be just save the unscale value - and do the scaleBySize on lines 1330 and/or 338.

poire-z:
(An other option, for later :), might be to have the marker width larger for top level TOC items, and thinner for lower levels - but I remember there's some code that only take items from a single level when showing marks, https://github.com/koreader/koreader/issues/5500#issuecomment-543201699.)

robert00s:
How to ignore this? 

```
Checking for unscaled sizes

Warning: it looks like you might be using unscaled sizes.
It is almost always preferable to defer to one of the predefined sizes in ui.size in the following files:
frontend/apps/reader/modules/readerfooter.lua:1191:                                    self.settings.toc_markers_width = 1
frontend/apps/reader/modules/readerfooter.lua:1202:                                    self.settings.toc_markers_width = 2
frontend/apps/reader/modules/readerfooter.lua:1213:                                    self.settings.toc_markers_width = 3

```

Frenzie:
You could either add it to the list of exceptions (that `grep -Ev '(default_option_height|default_option_padding)'` part) or write the variable name in such a way that it doesn't trigger. Perhaps we could add a generic `unscaled_width`/`height` override.

https://github.com/koreader/koreader/blob/c3a0bd4def2243b66cacf27222b62d9441ebbaff/.ci/check.sh#L12

The point of the script is merely to force you to think about it, and 9 times out of 10 it'll remind you not to do something stupid. The other time you have to amend the script or work around it.

poire-z:
Or just add to that line a generic `grep -v 'unscaled_size_check: ignore"` , so we can just use (after we have thought about it :):
```lua
self.settings.toc_markers_width = 1 -- unscaled_size_check: ignore
```

robert00s:
Thanks for tip :)

Frenzie:
True, it's probably clearer that way. :+1:

PR #5715, opened 2019-12-31
Unbreak the KUAL extension

NiLuJe:
Nasty symlink shenanigans meant we weren't actually shipping a required script...

Also, unclutter it while I'm here ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5715)

### DIFF
--- a/platform/kindle/extensions/koreader/bin/koreader-ext.sh
+++ b/platform/kindle/extensions/koreader/bin/koreader-ext.sh
@@ -22,7 +22,7 @@ logmsg() {
         f_log I koreader kual "" "${1}"
     fi
 
-    # And handle user visual feedback via eips...
+    # And handle user visual feedback via FBInk/eips...
     eips_print_bottom_centered "${1}" 1
 }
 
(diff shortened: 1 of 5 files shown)


NiLuJe:
(Waiting on CI to double-check that I haven't borked the symlink).

platform/kindle/extensions/koreader/bin/koreader-ext.sh
```diff
@@ -93,55 +93,6 @@ install_koreader() {
     update_koreader "clean"
 }
 
-# Handle cre's settings...
```
Frenzie: What even is all this? Some vestigial thing from 2011? >_>

NiLuJe: Yeah, reaaaaaallly old stuff ^^.

NiLuJe:
@Frenzie: Any idea why the CI is only now complaining about not having ragel to build HB from git?

Frenzie:
Some failed download? That's… pretty weird.

NiLuJe:
No, it tried to regenerate some headers (which I think actually *are* shipped in the git repo, so, some kind of timestamp mishap?).

Frenzie:
Just rerun it?

NiLuJe:
Just remembered that front's CI only does an emulator build, so, running a local kindle build to double-check ;p.

NiLuJe:
```
  adding: extensions/koreader/bin/libkohelper.sh (deflated 53%)
```

Wheee!

PR #5716, opened 2019-12-31
[i18n] Use sort to keep POT order more stable

Frenzie:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5716)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -494,9 +494,9 @@ pot: po
 	$(XGETTEXT_BIN) --from-code=utf-8 \
 		--keyword=C_:1c,2 --keyword=N_:1,2 --keyword=NC_:1c,2,3 \
 		--add-comments=@translators \
-		reader.lua `find frontend -iname "*.lua"` \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -494,9 +494,9 @@ pot: po
 	$(XGETTEXT_BIN) --from-code=utf-8 \
 		--keyword=C_:1c,2 --keyword=N_:1,2 --keyword=NC_:1c,2,3 \
 		--add-comments=@translators \
-		reader.lua `find frontend -iname "*.lua"` \
-		`find plugins -iname "*.lua"` \
-		`find tools -iname "*.lua"` \
+		reader.lua `find frontend -iname "*.lua" | sort` \
```
Frenzie: @NiLuJe I figure this is probably the best approach, right? I noticed the whole freaking file seems to be rewritten all the time.

NiLuJe: Yup, sounds good :).

PR #5727, opened 2020-01-02
Various fixes (reader.lua, ConfirmBox, In-page footnotes style tweaks)

### LINKED ISSUE #5707: Clarify how to set customised footnote markers
* KOReader version: koreader-kobo-v2019.11
* Device: Kobo H2O

#### Steps to reproduce
1) Attempt to display in-line footnotes
2) None are shown

#### Issue
With my epub document, I have enabled the setting where I can tap a footnote, and it will display in a pop-up at the bottom of the page. This works perfectly.

However, when I try to display footnotes inline, as per [this page](https://github.com/koreader/koreader/wiki/In-page-footnotes), nothing shows. I assume it's because the footnotes in my document are in an unexpected format. An image on that linked page suggests that custom classes can be set. However, KOReader's menu currently looks different to the menu on that page. How can I either:
a) determine what KOReader currently uses to recognise footnotes (hence I can edit my epub document), or
b) customise KOReader to recognise my document's custom footnotes.

(For reference, my document has `<a href="#rch1star1">*</a>`, which links to `<p class="note3"><a href="#ch1star1" id="rch1star1">*</a> ... </p>`, where the class is `note[0-9]`.)

### LINKED ISSUE #5717: Does not start on Kobo Touch (attempt to call field 'isKindle')
* KOReader version: koreader-kobo-v2019.12
* Device: Kobo Touch model N905C

#### Issue
Koreader crashes immediately at startup.

#### Steps to reproduce
1. Install KSM 09
2. Install Koreader
3. Boot
4. Select "start koreader"

[crash.log](https://github.com/koreader/koreader/files/4013662/crash.log)

poire-z:
- reader.lua: re-order touchScreenProbe() after CanvasContext. Should close #5717.
- ConfirmBox: prevent lower widgets from responding to tap: could happen when toggling Render Mode in the bottom menu, when we get the "Styles have changed in such a way... do you want to reload the document": taping on that text instead of the buttons would activate some other buttons in the bottom config menu hidden by that ConfirmBox :)
- Style tweaks: In-page footnotes: add some classic classnames: just adds `note2` and `note3`. Closes #5707.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5727)

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -604,7 +604,9 @@ ol.references > li > .mw-cite-backlink { display: none; }
 Show footnotes with classic class names at the bottom of pages.
 This tweak can be duplicated as a user style tweak when books contain footnotes wrapped with other class names.]]),
                 css = [[
-.footnote, .note, .note1, .ntb, .ntb-txt, .ntb-txt-j
+.footnote, .footnotes, .fn,
+.note, .note1, .note2, .note3,
+.ntb, .ntb-txt, .ntb-txt-j
 {
     -cr-hint: footnote-inpage;
     margin: 0 !important;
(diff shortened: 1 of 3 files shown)


robert00s:
> Style tweaks: In-page footnotes: add some classic classnames: just adds note2 and note3. Closes #5707.

Could you add also `.fn` and `.footnotes` classnames? I saw books with that classnames.

poire-z:
Done.
(Hoping the small `fn` classname is not commonly used for other things.)

PR #5729, opened 2020-01-02
[fix, i18n] Auto-save book metadata: fix interval string

poire-z:
See https://github.com/koreader/koreader/pull/5687#issuecomment-570226614 and followups.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5729)

### DIFF
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -373,7 +373,7 @@ local function genAutoSaveMenuItem(value)
     if not value then
         text = _("Only on close, suspend and exit")
     else
-        text = T(N_("Every 1 minute", "Every %1 minutes", value), value)
+        text = T(N_("Every minute", "Every %1 minutes", value), value)
     end
     return {
         text = text,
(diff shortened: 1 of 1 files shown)


frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -372,7 +373,7 @@ local function genAutoSaveMenuItem(value)
     if not value then
         text = _("Only on close, suspend and exit")
     else
-        text = T(_("Every %1 mn"), value)
+        text = T(N_("Every 1 minute", "Every %1 minutes", value), value)
```
Frenzie: Did I write that? If so, I meant:

```suggestion
        text = T(N_("Every minute", "Every %1 minutes", value), value)
```

Frenzie: Oh right, I wrote `Every 1 m`, which does work. ;-)

poire-z:
Forgot to add another one, where I use the shorter version, if that's allright (or I could use the full for consistency, but it may get more easiliy truncated as there's a prefix).

PR #5736, opened 2020-01-03
Footer font size

robert00s:
Move footer font size (`DMINIBAR_FONT_SIZE`) from Advanced Settings to `Status bar -> Settings
`

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5736)


poire-z:
Looks good to me.
May be just move the added menu item `Font size` above (before `Sort items` or after `[ ] Show footer separator`, before the separator ?), so the little less interesting `Time format>` and `Duration format>` both end up on the 2nd page?

robert00s:
Done :)
Moved after `[ ] Show footer separator`, before the separator

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -766,6 +769,41 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                end,
+                separator = true,
+                callback = function(touchmenu_instance)
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local font_size = self.settings.text_font_size
+                    local items_font = SpinWidget:new{
+                        width = Screen:getWidth() * 0.6,
+                        value = font_size,
+                        value_min = 10,
+                        value_max = 18,
```
Frenzie: Is a max value technically required for this widget? (Not in a position to test atm.) See https://www.mobileread.com/forums/showthread.php?p=3946432#post3946432

robert00s: Yes. max_value is required but I can prepare PR with `value_min = 8` and `value_max = 36`

Frenzie: Alright, please do! :+1:

PR #5737, opened 2020-01-03
android: call EPD test from within KOReader

### LINKED ISSUE #5730: Crash if no dictionary is installed
* KOReader version: 2019.12
* Device: Tolino Epos 2

#### Issue
Crash if a word in the book is selected an no dictionaries are installed.

#### Steps to reproduce
Select a word -> Crash

pazos:
Most users can call it from the developers menu:

test (1)

On some devices, where KOReader doesn't work at all, it can be called using adb with:

`adb shell am start -n org.koreader.launcher/.EPDTestActivity`

Fixes #5730 too

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5737)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -389,7 +389,7 @@ function FileManagerMenu:setUpdateItemTable()
     end
     if Device:isAndroid() then
         table.insert(self.menu_items.developer_options.sub_item_table, {
-            text = _("Start Eink Test"),
+            text = _("Start E-ink test"),
             callback = function()
                 Device:epdTest()
             end,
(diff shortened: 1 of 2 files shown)


frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -387,6 +387,15 @@ function FileManagerMenu:setUpdateItemTable()
             end,
         })
     end
+    if Device:isAndroid() then
+        table.insert(self.menu_items.developer_options.sub_item_table, {
+            text = _("Start Eink Test"),
```
Frenzie: I believe it's spelled E-ink in the rest of the program. (Please ignore if I recall incorrectly. ;-) )
```suggestion
            text = _("Start E-ink Test"),
```

PR #5738, opened 2020-01-03
android: OTA fallback using the browser

### LINKED ISSUE #5731: Crash when trying to run an update
* KOReader version: 2019.11
* Device: Tolino Epos 2

#### Issue
Crash when trying to run an update

#### Steps to reproduce
Select Update -> Check for Update. If there is an update available, click Update -> Crash

pazos:
Requires https://github.com/koreader/android-luajit-launcher/pull/215
Fixes #5731 

If the device is borked and fails to download using DownloadManager show a confirmBox to retry using the browser.

See how it works/looks (simulated because I don't have any borked device)

test (2)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5738)

### DIFF
--- a/frontend/ui/otamanager.lua
+++ b/frontend/ui/otamanager.lua
@@ -261,11 +261,18 @@ function OTAManager:fetchAndProcessUpdate()
                 ok_callback = function()
                     local isAndroid, android = pcall(require, "android")
                     if isAndroid then
-                        -- download the package if not present.
-                        if android.download(link, ota_package) then
-                            android.notification(T(_("The file %1 already exists."), BD.filename(ota_package)))
+                        -- try to download the package
+                        local ok = android.download(link, ota_package)
+                        if ok == 1 then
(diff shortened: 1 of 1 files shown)


frontend/ui/otamanager.lua
```diff
@@ -260,11 +260,18 @@ function OTAManager:fetchAndProcessUpdate()
...
+                        -- try to download the package
+                        local ok = android.download(link, ota_package)
+                        if ok == 1 then
                             android.notification(T(_("The file %1 already exists."), ota_package))
-                        else
+                        elseif ok == 0 then
                             android.notification(T(_("Downloading %1"), ota_package))
+                        else
+                            UIManager:show(ConfirmBox:new{
+                                text = _("Your device seems unable to download packages.\nRetry using the browser?"),
```
Frenzie:
```suggestion
                                text = _("Your device seems to be unable to download packages.\nRetry using the browser?"),
```

PR #5431, opened 2019-09-27
Joplin support

mustafa292:
adds joplin support https://github.com/koreader/koreader/issues/5086

Changes
-adds a submenu to evernote menu 
   - -Joplin
      |-Set IP and port
      |-Set authorization token
      |-Export to Joplin
      |-Help

-adds EvernoteExporter:exportBooknotesToJoplin()
-adds JoplinClient.lua
-modifies html_export, txt_export and joplin_export flags to work with each other. (eg if user selects one others deactivated)

Because joplin only listens on localhost, [see diccussion on joplin forums, ](https://discourse.joplinapp.org/t/clipper-api-listens-only-localhost-requesting-to-include-local-addresses/3560)  user needs to port forward on computer. To do that :
on linux `$socat tcp-listen:41185,reuseaddr,fork tcp:localhost:41184`
on windows `netsh interface portproxy add listeningaddress:0.0.0.0 listeningport:41185 connectaddress:localhost connectport:41184`
If you think this is too hacky, feel free to close this PR.  I can add all this into wiki in case someone may want to use Joplin.

PR #5744, opened 2020-01-04
[plugin] fix netsh command

rffontenelle:
The edited `netsh` command line had errors:
1- missing v4tov4 (or v6 variant)
2- should have `=` instead of `:`
3- typo in listen(ing)address and listen(ing)port

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5744)


rffontenelle:
FWIW: https://docs.microsoft.com/windows-server/networking/technologies/netsh/netsh-interface-portproxy

Frenzie:
For reference, added in #5431.

PR #5745, opened 2020-01-04
[android]: share text with other apps

pazos:
I'm an user of [Markor](https://github.com/gsantner/markor), an awesome markdown/text tool for android.

It has a Quicknote feature which is greeeat. Works with ACTION_SEND, so it is compatible with a lot of applications, including all web browsers. When a user shares text with markor and press "quicknote" he/she will return to the previous activity and the note will be stored in Quicknote.md.

See how it works with a browser:

test1

It is extremely useful to do initial researching about a topic, at least for me. So I did the same with KO
test2 (1)

It works exactly the same, but in my case I setup markor as my preferred app to send text from KO, so no application picker is shown.

This is kind of a one-user feature, but the share text can be repurposed for other uses: twitter?, whatsapp? I don't know.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5745)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -498,6 +498,17 @@ function ReaderHighlight:onShowHighlightMenu()
             },
         })
     end
+    if Device:canShareText() then
+            table.insert(highlight_buttons, {
+            {
+                text = _("Share text"),
+                callback = function()
+                    Device.doShareText(self.selected_text.text)
+                end,
+            },
+        })
+    end
+
     self.highlight_dialog = ButtonDialog:new{
         buttons = highlight_buttons,
         tap_close_callback = function() self:handleEvent(Event:new("Tap")) end,
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -86,6 +86,8 @@ local Device = Generic:new{
     canImportFiles = function() return android.app.activity.sdkVersion >= 19 end,
     importFile = function(path) android.importFile(path) end,
     isValidPath = function(path) return android.isPathInsideSandbox(path) end,
+    canShareText = yes,
+    doShareText = function(text) android.sendText(text) end,
 
     canExternalDictLookup = yes,
     getExternalDictLookupList = getExternalDicts,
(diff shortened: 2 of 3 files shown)


pazos:
@Frenzie: my bad, I was sure I labeled this PR as wip but it seems I didn't.

It needs https://github.com/koreader/android-luajit-launcher/pull/216 to work :man_facepalming: :woman_facepalming: Sorry!

Also I'm not sure if @poire-z is fine with the layout of the new entry in the menu.

poire-z:
Well, as it's for now only on Android, I'll be fine :)
Ideally, it would have been near "Copy", but no other item deserves to be full width.
And the other optional item is a Follow link that is show full width like yours only when selected text starts with a link.
So, we could have 10 items by always showing Follow link and Share text, but as the latter is not shown on other devices than Android, this is not a solution.
So, I guess it's fine :)
(In my own patch, I have something like that that sends the selected text to a web server of mine - which I replace "Add note" (that I never use) with.)

pazos:
Cool. From the pov of my workflow I would be better closing the menu before sharing the text:

```diff
diff --git a/frontend/apps/reader/modules/readerhighlight.lua b/frontend/apps/reader/modules/readerhighlight.lua
index 15046c16..28952863 100644
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -503,7 +503,10 @@ function ReaderHighlight:onShowHighlightMenu()
             {
                 text = _("Share text"),
                 callback = function()
-                    Device.doShareText(self.selected_text.text)
+                    local text = self.selected_text.text
+                    -- call self:onClose() before calling the android framework
+                    self:onClose()
+                    Device.doShareText(text)
                 end,
             },
         })
```

This will match the behaviour of the first gif. But I'm fine leaving the menu open (and the text selected) if this makes sense.

poire-z:
Highlight, Add note, Search, Follow link - close the dialog.
Copy, View HTML, Wikipedia, Dictionary, Translate - don't close the dialog.

From that, no obvious hint about how Share Text should be :) and no real opinion about what is best.
So, I guess, do as what you and your workflow prefer - someone disagreeing will open an issue :)

PR #5746, opened 2020-01-05
Spelling: Wi-Fi, turn on/off, on/off

comradekingu:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5746)

### DIFF
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -76,7 +76,7 @@ local action_strings = {
     toggle_rotation = _("Toggle rotation"),
 
     wifi_on = _("Turn on Wi-Fi"),
-    wifi_off = _("Tun off Wi-Fi"),
+    wifi_off = _("Turn off Wi-Fi"),
     toggle_wifi = _("Toggle Wi-Fi"),
 
     toggle_bookmark = _("Toggle bookmark"),
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -1603,7 +1603,7 @@ function ReaderGesture:onGSensorToggle()
     if G_reader_settings:isTrue("input_ignore_gsensor") then
         new_text = _("Accelerometer rotation events on.")
     else
-        new_text = _("Accelerometer rotation events off.")
+        new_text = _("Accelerometer rotation events on.")
     end
     UIManager:show(Notification:new{
         text = new_text,
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -1601,7 +1601,7 @@ end
 function ReaderGesture:onGSensorToggle()
     local new_text
     if G_reader_settings:isTrue("input_ignore_gsensor") then
-        new_text = _("Accelerometer rotation events on.")
+        new_text = _("Accelerometer rotation events off.")
     else
         new_text = _("Accelerometer rotation events on.")
     end
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -75,9 +75,9 @@ local action_strings = {
     toggle_gsensor = _("Toggle accelerometer"),
     toggle_rotation = _("Toggle rotation"),
 
-    wifi_on = _("Enable wifi"),
-    wifi_off = _("Disable wifi"),
-    toggle_wifi = _("Toggle wifi"),
+    wifi_on = _("Turn on Wi-Fi"),
+    wifi_off = _("Tun off Wi-Fi"),
+    toggle_wifi = _("Toggle Wi-Fi"),
 
     toggle_bookmark = _("Toggle bookmark"),
     toggle_page_flipping = _("Toggle page flipping"),
@@ -248,7 +248,7 @@ function ReaderGesture:init()
         local reset = false
         for k, v in pairs(mirrored_if_rtl) do
             -- We only replace them if they are still the other direction's default.
-            -- If not, the user has changed them: let him deal with setting new ones if needed.
+            -- If not, the user has changed them: Let it deal with setting new ones if needed.
             if gm[k] == self.default_gesture[v] then
                 gm[k] = self.default_gesture[k]
                 reset = true
@@ -259,7 +259,7 @@ function ReaderGesture:init()
             end
         end
         if reset then
-            logger.info("UI language direction changed: resetting some gestures to direction default")
+            logger.info("UI language direction changed: Resetting some gestures to default direction.")
         end
         G_reader_settings:flipNilOrFalse(ges_dir_setting)
     end
@@ -346,7 +346,7 @@ function ReaderGesture:addToMainMenu(menu_items)
         text = _("Gesture manager"),
         sub_item_table = {
             {
-                text = _("Enable multiswipes"),
+                text = _("Turn on multiswipes"),
                 checked_func = function() return self.multiswipes_enabled end,
                 callback = function()
                     G_reader_settings:saveSetting("multiswipes_enabled", not self.multiswipes_enabled)
@@ -1220,13 +1220,13 @@ function ReaderGesture:registerGesture(ges, action, ges_type, zone, overrides, d
                 if ges == "multiswipe" then
                     if self.multiswipes_enabled == nil then
                         UIManager:show(ConfirmBox:new{
-                            text = _("You have just performed a multiswipe gesture for the first time.") .."\n\n".. multiswipes_info_text,
-                            ok_text = _("Enable"),
+                            text = _("You have just performed your first multiswipe gesture.") .."\n\n".. multiswipes_info_text,
+                            ok_text = _("Turn on"),
                             ok_callback = function()
                                 G_reader_settings:saveSetting("multiswipes_enabled", true)
                                 self.multiswipes_enabled = true
                             end,
-                            cancel_text = _("Disable"),
+                            cancel_text = _("Turn off"),
                             cancel_callback = function()
                                 G_reader_settings:saveSetting("multiswipes_enabled", false)
                                 self.multiswipes_enabled = false
@@ -1449,7 +1449,7 @@ function ReaderGesture:gestureAction(action, ges)
 
         if not NetworkMgr:isOnline() then
             UIManager:show(InfoMessage:new{
-                text = _("Enabling wifi…"),
+                text = _("Turning on Wi-Fi…"),
                 timeout = 1,
             })
 
@@ -1460,7 +1460,7 @@ function ReaderGesture:gestureAction(action, ges)
             NetworkMgr:turnOffWifi()
 
             UIManager:show(InfoMessage:new{
-                text = _("Wifi disabled."),
+                text = _("Wi-Fi off."),
                 timeout = 1,
             })
         end
@@ -1470,7 +1470,7 @@ function ReaderGesture:gestureAction(action, ges)
         NetworkMgr:turnOffWifi()
 
         UIManager:show(InfoMessage:new{
-            text = _("Wifi disabled."),
+            text = _("Wi-Fi off."),
             timeout = 1,
         })
     elseif action == "wifi_on" then
@@ -1483,7 +1483,7 @@ function ReaderGesture:gestureAction(action, ges)
             })
 
             -- NB Normal widgets should use NetworkMgr:promptWifiOn()
-            -- This is specifically the toggle wifi action, so consent is implied.
+            -- This is specifically the toggle Wi-Fi action, so consent is implied.
             NetworkMgr:turnOnWifi()
         else
             local info_text
@@ -1587,9 +1587,9 @@ function ReaderGesture:onShowFLOnOff()
     local powerd = Device:getPowerDevice()
     local new_text
     if powerd.is_fl_on then
-        new_text = _("Frontlight is on.")
+        new_text = _("Frontlight on.")
     else
-        new_text = _("Frontlight is off.")
+        new_text = _("Frontlight off.")
     end
     UIManager:show(Notification:new{
         text = new_text,
@@ -1601,9 +1601,9 @@ end
 function ReaderGesture:onGSensorToggle()
     local new_text
     if G_reader_settings:isTrue("input_ignore_gsensor") then
-        new_text = _("Accelerometer rotation events will now be ignored.")
+        new_text = _("Accelerometer rotation events on.")
     else
-        new_text = _("Accelerometer rotation events will now be honored.")
+        new_text = _("Accelerometer rotation events off.")
     end
     UIManager:show(Notification:new{
         text = new_text,


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -75,9 +75,9 @@ local action_strings = {
     toggle_gsensor = _("Toggle accelerometer"),
     toggle_rotation = _("Toggle rotation"),
 
-    wifi_on = _("Enable wifi"),
-    wifi_off = _("Disable wifi"),
-    toggle_wifi = _("Toggle wifi"),
+    wifi_on = _("Turn on Wi-Fi"),
+    wifi_off = _("Tun off Wi-Fi"),
```
Frenzie: It would be preferable to do the reverse. That is, change every accidental occurrence of "Wi-Fi" to wifi.

(Also, major typo. ;-) )

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -248,7 +248,7 @@ function ReaderGesture:init()
         local reset = false
         for k, v in pairs(mirrored_if_rtl) do
             -- We only replace them if they are still the other direction's default.
-            -- If not, the user has changed them: let him deal with setting new ones if needed.
+            -- If not, the user has changed them: Let it deal with setting new ones if needed.
```
Frenzie:
```suggestion
            -- If the user has changed directions, they can deal with setting new ones if needed.
```

They is a gender-neutral third person pronoun, but it is not normally used for people.

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -259,7 +259,7 @@ function ReaderGesture:init()
             end
         end
         if reset then
-            logger.info("UI language direction changed: resetting some gestures to direction default")
+            logger.info("UI language direction changed: Resetting some gestures to default direction.")
```
Frenzie: While I don't typically police comments too much, our house style doesn't capitalize after colons.
```suggestion
            logger.info("UI language direction changed: resetting some gestures to default direction.")
```
```suggestion
            logger.info("UI language direction changed. Resetting some gestures to default direction.")
```

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1601,9 +1601,9 @@ end
 function ReaderGesture:onGSensorToggle()
     local new_text
     if G_reader_settings:isTrue("input_ignore_gsensor") then
-        new_text = _("Accelerometer rotation events will now be ignored.")
+        new_text = _("Accelerometer rotation events on.")
```
Frenzie: I'm pretty sure that's the reverse. :-P

frontend/apps/reader/modules/readergesture.lua
poire-z: If we need to be that excessively politically correct in comments, I'm gonna stop putting comments :|

frontend/apps/reader/modules/readergesture.lua
poire-z: No, it should stay `resetting some gestures to direction default`.
We're not resetting gestures to any "direction" (so, not to "default direction).
We're resetting gestures to the gestures that are the default for that new direction.

poire-z:
Mhhh, these changes feel all gratuitous, less precise, or wrong. (I seem to prefer our current enable/disable vs turn on/off - but english being not my first language...)

pazos:
> Mhhh, these changes feel all gratuitous, less precise, or wrong. (I seem to prefer our current enable/disable vs turn on/off - but english being not my first language...)

On linux NetworkManager is using "enable/disable" toggles for wireless and wired networks. I would stay with these too. 

For multigesture "turn on/off" doesn't make any sense as we're not turning out anything.

frontend/apps/reader/modules/readergesture.lua
Frenzie: Ftr, I'm objecting exclusively to "it".

Frenzie:
Multigestures (or basically anything that can be toggled) are either turned on/enabled or turned off/disabled. Turn on is arguably more natural. It's an unnecessary change, but not a bad one.

poire-z:
For things with a state, like light or wifi, I'd be ok with _turn on/off_.
For something that is a feature, that is there and can apply to stuff that needs it, or not if there's no such stuff, _turn on/off_ sounds strange to me.
For example, with EPUB rendering, I _enable floats_, I don't _turn on floats_. But I may _turn on floats support_.
Same way, I think we _enable multigestures_. Or _turn on multigestures support_. But the former is simpler, shorter, obvious :)

frontend/apps/reader/modules/readergesture.lua
NiLuJe: Yep, purely grammatical: *it* for a person just doesn't work ;).

frontend/apps/reader/modules/readergesture.lua
NiLuJe: But I kind of *like* Wi-Fi! ;p.

(In fact, most of those occurrences are probably from me ;p).

Frenzie: I'm likely responsible for a decent amount of wifi, but it wasn't a mistake. I'm well aware of what the Wi-Fi Alliance calls itself, but in my usage hifi and wifi are fundamentally the same. That applies regardless whether we're talking about it as an abbreviation of a phrase like wireless fidelity or as a generic concept of wireless network, free from the confines of any particular trademark, patent, or technical standard.

(I recognize that in well-edited English prose, wifi is still in a minority. This will change.)

comradekingu: If you read the actual claims of the Wi-Fi Alliance, it only concerns selling a physical product that makes use of Wi-Fi technology. By any stretch of trademark law, other uses are unenforceable, and especially now that "Wi-Fi" is a word. As it will remain being outside of whatever possibly consistent action is taken in KOReader. "wifi" just looks wrong, and pollutes the otherwise "Wi-Fi" translation memory of Weblate. Alternatives are "WLAN" and "wireless network".

frontend/apps/reader/modules/readergesture.lua
comradekingu: It isn't political, just poor English. There is no need to personify the user, as it is a concept.
If for example, (without that) "If anyone wants to use the fire escape in case of a fire, they will find it around the corner." "they" can be employed.

frontend/apps/reader/modules/readergesture.lua
comradekingu:
```suggestion
            logger.info("UI language direction changed: Resetting some gestures to directional defaults.")
```

comradekingu: Pretty sure I saw both. Using the one that is only allowed in English baits errors in translation.

frontend/apps/reader/modules/readergesture.lua
comradekingu:
```suggestion
        new_text = _("Accelerometer rotation events off.")
```

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1601,9 +1601,9 @@ end
...
     else
-        new_text = _("Accelerometer rotation events will now be honored.")
+        new_text = _("Accelerometer rotation events off.")
```
comradekingu:
```suggestion
        new_text = _("Accelerometer rotation events on.")
```

frontend/apps/reader/modules/readergesture.lua
comradekingu:
```suggestion
    wifi_off = _("Turn off Wi-Fi"),
```

frontend/apps/reader/modules/readergesture.lua
Frenzie: Could you please point out the problematic strings? This is just random code.

frontend/apps/reader/modules/readergesture.lua
poire-z: If we can't use poor english in comments, I'm gonna stop putting comments :|

(But I doubt it's worse than "it" or "they"... I often use "the user" when I refer to users-only in comments - when I write about how developpers would/should do things, I write "you". I respect both of them, and neither is more "a concept" than the others ! They're both human beings :)

And at least, "him", personifying the user, drop some ambiguity:
What would be "it" in your initial suggestion? the user? the code that follows? the widget class we're implementing? the widget object that will be instantiated?
With "him", no ambiguity.

So please, leave this comment as it is - and ignore the above suggestion as it is dropping some contextual logic by removing the leading `If not, `. Unclear logic is worse than poor english.

frontend/apps/reader/modules/readergesture.lua
poire-z: Indeed, that's just some string that is logged in crash.log - it shouldn't reach weblate in any way - or does it ?!

poire-z: I don't like this one either :) even if directional is the english adjective for direction, I'd rather keep the word direction to again be explicite (as we're not all native english speakers, simple english is sometimes better in comments).
But it's just a log, and if you'd really want it to be clear and unambiguous, you'd use:
```diff
-UI language direction changed: resetting some gestures to direction default
+UI language direction changed: Resetting some gestures to their default for the new direction.
```

frontend/apps/reader/modules/readergesture.lua
Frenzie: Using "he" is perhaps slightly outdated stylistically, but it's definitely not poor English.

frontend/apps/reader/modules/readergesture.lua
Frenzie: The original sentence would be better as "direction defaults."

As an aside, for comments & logs I purposefully try not suggest corrections or improvements as long as they're comprehensible and not ungrammatical. Because there the goal is the future maintainability of the code.

pazos:
In spanish, galician etc. "Apaga/enciende el wifi" sounds horrible and is not used outside the context of physical turning on/off wifi routers. What users will do is "comprobar si el wifi esta activado" in their phones.

Both MacOS, Linux and Android have actions to "Desactivar el WiFi", not for "Apagar el Wifi". On some devices, like the Cervantes, is plain false as wireless chip cannot be turned off.

In short: enable/disable is what we do. In some devices it does imply turning on/off vcore but that's too technical to be presented to the user. Disable means: drop a lease, break a link, unload a module. Enable the oposite: load a module, link to ap, get a lease.

Frenzie:
> In spanish, galician etc.

That's not relevant to English I'm afraid. That means the proper translation of "turn off" would be "desactivar." No more, no less. :-)

> In short: enable/disable is what we do. In some devices it does imply turning on/off vcore but that's too technical to be presented to the user. 

By this argument, the conclusion may well be the reverse. Turn on/off is how regular people talk, while enable/disable is more technical.

comradekingu:
@Frenzie Could just the name of the thing being either on/off/activated/deactivated be used, with a checkbox next to it?

frontend/apps/reader/modules/readergesture.lua
Frenzie: No one said anything about the claims of the Wi-Fi Alliance. What I wrote is that no matter if you look at it as an abbreviation of wireless fidelity or as a generic descriptor, the correct spelling is _wifi_.

WLAN is something like Central European, not exactly regular English. It can be an excellent translation for wifi/Wi-Fi. Wireless network is a bit verbose and possibly goofy, but better.

As for any "pollution," such loaded language makes me want to take up a principled stance against Wi-Fi pollution. :imp: But seriously, a translation memory can't be "polluted" by a slightly different spelling. Next thing you know someone will spell _color_ like _colour_. Oh, the humanity…

Frenzie:
@comradekingu That's the general pattern. Turn on wifi/W-Fi is one of a small number of exceptions because it occurs in a scenario like:

1. The user selects an action that requires network access.
2. The program notices a lack of networking, so it pops up a dialog.

comradekingu:
> Mhhh, these changes feel all gratuitous, less precise, or wrong. (I seem to prefer our current enable/disable vs turn on/off - but english being not my first language...)

You are turning off functionality. The distinction is when capability is underlying said functionality, but I don't think that is the case here (?)

frontend/apps/reader/modules/readergesture.lua
comradekingu: I grant you that point, and it is one of semantics. Apply it to that of "Wi-Fi*", and you will find nobody spells it _wifi".

https://dictionary.cambridge.org/dictionary/english/wi-fi?q=wifi
https://www.dictionary.com/browse/wifi?s=t
https://dictionary.cambridge.org/dictionary/english/wi-fi?q=wifi
https://www.vocabulary.com/dictionary/WiFi https://en.wikipedia.org/wiki/Wi-Fi

Meanwhile, spelling it "Wi-Fi", _isn't wrong_, _and_ enjoys the most popularity.

Some readers have 3G/WiMAX/4G and so on too. My personal preference is to use something more generic when it is possible. Even https://emojipedia.org/antenna-with-bars/ would do.

pazos:
> That's not relevant to English I'm afraid. That means the proper translation of "turn off" would be "desactivar." No more, no less. :-)

Exactly. Translations have their own life, I guess :+1:

frontend/apps/reader/modules/readergesture.lua
NiLuJe: @poire-z: I *really* wouldn't worry about that ;). We're being nitpicky *because* the PR itself is nitpicky. Kind of a vicious circle ;).

frontend/apps/reader/modules/readergesture.lua
Frenzie: Nobody except for Nerdist, Business Insider, Gizmodo, Techcrunch, Public Contract Law Journal, Analog, Literary Review, Daily Beast, MotherJones, NewStatesMan, Denver, CSMonitor, and FantasySciFi, just to name a few. Many other publications use WiFi, Wifi, as well as wi-fi. TIME Magazine switched from Wi-Fi to wi-fi in 2003 when it turned generic.

Note that I [said](https://github.com/koreader/koreader/pull/5746#discussion_r363202383) I want to be part of the change:

> I recognize that in well-edited English prose, wifi is still in a minority. This will change.

I assume you included the Cambridge Dictionary twice to emphasize that it does, in fact, fully agree with me. ;-) The fact that it happens to spell the words in question _hi-fi_ and _wi-fi_ is not the salient point, only that it treats them the same.

PS I thought the corpus of US Supreme Court Opinions was quite amusing.

https://caselaw.findlaw.com/us-supreme-court/13-896.html
> Suppose an extensive business headquarters or a resort or a college campus wants a single, central wireless system (sometimes called a Wi-Fi network). In order to cover the large space, the system needs multiple base stations so a user can move around the area and still stay connected.

Frenzie:
> Edit: "Turn on" should be "OK" there. (According to convention.)

Convention in the '90s perhaps. ;-) This design pattern has been well established in the 2000s by Apple and GNOME.

https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/dialogs
dialog_rs2_delete_file

https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/dialogs/
Dialog-buttons_2x

https://material.io/components/dialogs/#usage
mio-components_assets_1Vs3vgE4jR4V5x-jjaQUhC1LqdOmdi2o2_dialogs-alertdialog

https://developer.gnome.org/hig/stable/dialogs.html.en
Screenshot_2020-01-06_19-01-02

Frenzie:
Anyway, let's just drop the changes to the comments & keep the rest of the PR as is (with turn on/off and Wi-Fi).

frontend/apps/reader/modules/readergesture.lua
poire-z: Thanks :) but I don't _really_ worry - if we can't put a bit of drama-dy in our GitHub comments, I'm gonna stop writting GitHub comments! :)
Just trying to discourage rich-english translators from too easily thinking they can change comments or non-translated strings for pseudo grammar reasons, without having to read and understand the code and logic surrounding these comments (because the code _is_ some comment to the comments - in case the comments are not clear enough :) - better they not contradict each others).

frontend/apps/reader/modules/readergesture.lua
Frenzie: Using the comment suggest functionality to fix this up for merge.
```suggestion
            logger.info("UI language direction changed: resetting some gestures to their default for the new direction.")
```

Frenzie: @comradekingu You didn't enable push access. Could you please fix up/revert the random comments so the PR can be merged?

comradekingu: @Frenzie I did, but your are confusing the plural "gestures" with the singular "default". I didn't randomly change that.

comradekingu:
@Frenzie Not sure what discussion this is. There is _plenty_ of stuff to actually fix in KOReader.
I think the greater in this specific case is to explain what is happening. Adding the imperative up front gets to the point quicker, with less arbitration.
"Turn on Wi-Fi to *?"
Cancel | OK

frontend/apps/reader/modules/readergesture.lua
comradekingu: @Frenzie You will never see anything but "Wi-Fi" from the Wi-Fi alliance. This ensures anything but "Wi-Fi", which is in clear minority for good reason, will always be and look like a knock-off.
My changes improve consistency within KOReader, and also help avoid errors in translation, as anything but "Wi-Fi" isn't allowed elsewhere.
Using "wifi" for wireless network is an erroneous practice.

frontend/apps/reader/modules/readergesture.lua
comradekingu: So you use "he", for archaic reasons, sure. The odds of anyone thinking that isn't in poor taste and for no good reason is slim. Slimmer still considering the state of the rest of the strings in KOReader. Speaking of users, lets have them, and lets consider what they are likely to understand rather than what is merely a technically dependable position.

frontend/apps/reader/modules/readergesture.lua
comradekingu: @Frenzie Didn't find any, but most are unexplained placeholders, so could be I remembered it as being one of those. https://github.com/koreader/koreader/blob/a7def7e8338659e51f4bc956487787391b76a41a/frontend/apps/reader/modules/readerwikipedia.lua#L484 is off, and I saw about 10 other clear errors looking for it.
Do you want me to find evidence of errors in translation resulting from : small-versal?

Frenzie:
@comradekingu The discussion was about passive labels.

pazos:
@comradekingu: please differenciate between issues in the Ui and issues in the translation. You are making translation changes in this PR. If you want to make Ui changes then please open other PR. If you don't want to make the changes yourself but suggest them for others to do the work, then open a new ticket per each issue describing current behaviour and expected behaviour.

That makes things easier for contributions.

frontend/apps/reader/modules/readergesture.lua
NiLuJe: Doesn't change the fact that `it` is ungrammatical in this context?

(TL;DR: While "he" isn't great, at least it makes a proper sentence).

frontend/apps/reader/modules/readergesture.lua
Frenzie: @comradekingu 
Screenshot_2020-02-12_18-36-25

> but you are confusing the plural "gestures" with the singular "default".

You can certainly say English generally prefers a plural noun for a repeated idea, but that's besides the point.

NiLuJe: Except it's "value" taking the plural in his suggestion. default values OR defaults. defaults values would be wrong, though.

Frenzie:
On that note, this PR has been blocked for over a month because it changes more than what it says on the tin (Wi-Fi and turn on/off). Ideally those would be three separate PRs.

One for Wi-Fi (probably merged), one for turn on/off (merged immediately), and one for nitpicking internal comments (probably not merged but in any case not holding up anything else).

frontend/apps/reader/modules/readergesture.lua
Frenzie: @NiLuJe I just edited it to "default values"; it just said "default" previously because I copied it straight from @poire-z as a statement. ;-)

I figured most people would miss the subtle statement; I wasn't quite expecting it to be caught yet still missed. ^_^

NiLuJe: @Frenzie: Ah, okay, now it makes sense ;). Thanks.

Frenzie: Yes please! :-)

frontend/apps/reader/modules/readergesture.lua
Frenzie: @comradekingu https://github.com/koreader/koreader/pull/5746#issuecomment-571270197

> Anyway, let's just drop the changes to the comments & keep the rest of the PR as is (with turn on/off and Wi-Fi).

comradekingu:
@Frenzie Ideally I would have the time, but I don't. KOReader has 10k words that I haven't gotten through yet. If at some time it stops being all over the place, one could start going easier on the changes. For how long it has been this way, it seems me doing something about it is how that would end up getting done.

Frenzie:
Changes merged in #5856.

PR #5757, opened 2020-01-10
Prevent overwriting text in footer

robert00s:
This patch fix problem with overwriting text in footer in `Show all at once` mode and when we select many items to show. When text is to wide we truncate it and end with ellipsis(...).
Now we can also set minimal width of progress bar (from 20% to 50% of screen size) in all at once mode.

Before:

After: 

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5757)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -307,7 +307,7 @@ function ReaderFooter:init()
         self.settings.toc_markers_width = DMINIBAR_TOC_MARKER_WIDTH
     end
     if not self.settings.progress_bar_min_width then
-        self.settings.progress_bar_min_width = 20
+        self.settings.progress_bar_min_width = 20  -- unscaled_size_check: ignore
     end
     self.mode_list = {}
     for i = 0, #self.mode_index do
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -1490,7 +1490,7 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
             self.footer_container.dimen.h = 0
             self.footer_text.height = 0
         end
-        self.footer_text.max_width = math.floor(self._saved_screen_width - 2 * self.settings.progress_margin_width)
+        self.footer_text:setMaxWidth(math.floor(self._saved_screen_width - 2 * self.settings.progress_margin_width))
         self.footer_text:setText(text)
         self.progress_bar.width = math.floor(self._saved_screen_width - 2 * self.settings.progress_margin_width)
         self.text_width = self.footer_text:getSize().w
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -1497,6 +1497,7 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
     else
         if self.has_no_mode or not text or text == "" then
             self.text_width = 0
+            self.footer_text:setText(text)
         else
             local min_progress_bar_width = (100 - self.settings.progress_bar_min_width) / 100
             self.footer_text:setMaxWidth(math.floor(min_progress_bar_width * self._saved_screen_width - 2 * self.settings.progress_margin_width))
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -306,8 +306,8 @@ function ReaderFooter:init()
     if not self.settings.toc_markers_width then
         self.settings.toc_markers_width = DMINIBAR_TOC_MARKER_WIDTH
     end
-    if not self.settings.progress_bar_min_width then
-        self.settings.progress_bar_min_width = 20  -- unscaled_size_check: ignore
+    if not self.settings.progress_bar_min_width_pct then
+        self.settings.progress_bar_min_width_pct = 20  -- unscaled_size_check: ignore
     end
     self.mode_list = {}
     for i = 0, #self.mode_index do
@@ -1324,7 +1324,7 @@ function ReaderFooter:addToMainMenu(menu_items)
             },
             {
                 text_func = function()
-                    return T(_("Minimal width (%1%)"), self.settings.progress_bar_min_width)
+                    return T(_("Minimal width (%1%)"), self.settings.progress_bar_min_width_pct)
                 end,
                 enabled_func = function()
                     return not self.settings.progress_bar_position and not self.settings.disable_progress_bar
@@ -1334,7 +1334,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                     local SpinWidget = require("ui/widget/spinwidget")
                     local items = SpinWidget:new{
                         width = Screen:getWidth() * 0.6,
-                        value = self.settings.progress_bar_min_width,
+                        value = self.settings.progress_bar_min_width_pct,
                         value_min = 20,
                         value_step = 5,
                         value_hold_step = 20,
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -306,6 +306,9 @@ function ReaderFooter:init()
     if not self.settings.toc_markers_width then
         self.settings.toc_markers_width = DMINIBAR_TOC_MARKER_WIDTH
     end
+    if not self.settings.progress_bar_min_width then
+        self.settings.progress_bar_min_width = 20  -- unscaled_size_check: ignore
```
poire-z: So we don't have to read until `text = _("Minimal progress bar width in percentage of screen width"),` to understand it's a percentage :) either add a comment above saying `-- This is a percentage value, % of the screen width`, or rename that variable/setting to `progress_bar_min_width_pct`.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1320,6 +1322,35 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                end,
+                callback = function(touchmenu_instance)
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local items = SpinWidget:new{
+                        width = Screen:getWidth() * 0.6,
+                        value = self.settings.progress_bar_min_width,
+                        value_min = 20,
+                        value_step = 5,
+                        value_hold_step = 20,
+                        value_max = 50,
```
poire-z: Will this limit the progress bar witdh to 50% - and not allow it to take the remaining 80% when I want it to show only 2 items (we need all at once enabled for that, with all items unchecked except two) ?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1442,29 +1473,35 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
         return
     end
     local text = self:genFooterText()
-    if text then
-        self.footer_text:setText(text)
-    end
```
poire-z: You seem to have duplicated that single one 5 times below in all if/else branches. Any reason for removing it from here?
No other solution to avoid duplicating it? May be have it simpler by just keeping it here, with before a `if not text then text = "" end` if we ever can get `text=nil`.
Same for the generic `setMaxWidth(math.floor(self._saved_screen_width - 2 * self.settings.progress_margin_width))`: you could do it here for all case, even if in one case, you do a duplicate call to that (it does not cost much).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1442,29 +1473,35 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
...
+        self.footer_text:setMaxWidth(math.floor(self._saved_screen_width - 2 * self.settings.progress_margin_width))
+        self.footer_text:setText(text)
         self.progress_bar.width = math.floor(self._saved_screen_width - 2 * self.settings.progress_margin_width)
         self.text_width = self.footer_text:getSize().w
     else
         if self.has_no_mode or not text or text == "" then
             self.text_width = 0
+            self.footer_text:setText(text)
         else
+            local min_progress_bar_width = (100 - self.settings.progress_bar_min_width) / 100
```
poire-z: That local variable `min_progress_bar_width` is not what its name says :) it should be named something like `text_max_available_ratio`.

frontend/apps/reader/modules/readerfooter.lua
robert00s: No. This is only minimal bar width.
With 2 items progress bar will be wider than 50%.
obraz


Of course I can increase value_max to 80 if you want :)

frontend/apps/reader/modules/readerfooter.lua
robert00s: I didn't know that I can use `setText()` before `setMaxWidth()`. Before first commit I used `self.footer_text.max_width` and I had to duplicate that to proper work. Later I notice that better to use method `setMaxWidth()`.

Thanks for tip. Now should be much cleaner.

frontend/apps/reader/modules/readerfooter.lua
poire-z: Oh, ok, thanks :) Then, no need to increase it I guess.

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: IIRC, I had a vague reason for skipping the setText entirely if text is nil/empty, I'd like to keep that, if possible ;).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -306,6 +306,9 @@ function ReaderFooter:init()
     if not self.settings.toc_markers_width then
         self.settings.toc_markers_width = DMINIBAR_TOC_MARKER_WIDTH
     end
+    if not self.settings.progress_bar_min_width_pct then
+        self.settings.progress_bar_min_width_pct = 20  -- unscaled_size_check: ignore
```
poire-z: I think you don't need that `-- unscaled_size_check: ignore` anymore (.ci/check.sh checks for `width =`, `width_pct =` shouldn't trigger it).

frontend/apps/reader/modules/readerfooter.lua
poire-z: May be in https://github.com/koreader/koreader/pull/4850/commits/4e8841654044f4f05e6b28cfd556e0e739b7cb7c from #4850.
Looks like it was just to prevent some issue with nil being returned when text is empty - which I guess might no more be needed since my recent changes to TextWidget. Also, since it, setText and setMaxWidth costs nothing until draw time or getSize() is called - and there are some shortcuts when text is empty (not rendering anything, but just getting the height from the font).
Also, (quick) seeing no logical change as most tests where `or not text or text == ""` and became `or text == ""`

NiLuJe: I'm wary of piling on "almost nothing"s on page turns, even if the situation is better than it used to be ;).

Plus, I've used this to debug what the status bar was actually doing on every page turn re: disabled stuff.

NiLuJe: But: yeah, at least for the other branches (i.e., nil -> ""), you're right, this is slightly simpler ;).

poire-z: Given all the (more expensive) things readerfooter already does on page turn, this (really) nothing won't hurt. Furthermore if we also now do setMaxWidth (any one of them will call _xtext:free(), which is the only "expensive" stuff that could be done), so setText() is really a no-op.
The setText() could be just wrapped in a `if text then` as before and it wouldn't hurt (as it did not hurt before), if you really want - but I think that's less readable - my first reflex when reading the PR was actually worrying : why don't we reset setText() to empty, just to have no risk later in the branches to have the previous text displayed or previous width re-used. One would have to really read the followup code to be sure there's no risk of that happening (I did not re-read it :)  but I expect you to do that to find a good reason of keeping your trickery :)

NiLuJe: You're completely right, I'm more concerned with the setMaWidth than the setText by now ;).

poire-z: (sorry, did not see your 3rd post just before mine.)

NiLuJe: It's been too long since I played with readerfooter for me to tell whether hiding the setMaxWidth behind a text = "" would be enough/right, though.

My understanding is that it's only needed for visible modes that don't already have a custom setMaxWidth?

NiLuJe: (To explain my reasoning and the previous comments: the goal being to skip useless string formatting/layout when the status bar is invisible).

poire-z: > My understanding is that it's only needed for visible modes that don't already have a custom setMaxWidth?

setText and setMaxWidth are also no-op when the text or max_width are the same as the existing ones.
So, when text = "" always, when no mode, we'll just have these no-ops executed on a TextWidget with some already empty text that won't ever be sized/rendered (because of is_empty shortcut and paintTo skip) even if it were to be rendered by the complex readerfooter code or errors within it :)

NiLuJe: Cool! Did not know that ;).

Then everything's purrfect ;).

PR #5758, opened 2020-01-10
Unbreak OTA on Kindle

### LINKED ISSUE #5721: Cannot be relaunched after update
* KOReader version:v2019.11-51-ge8f91ab_2019-12-03
* Device:KPW3-32G

Cannot be relaunched after update. Just like issue #5269.
[crash.log](https://github.com/koreader/koreader/files/4014171/crash.log)

NiLuJe:
Regression since #5715 (because I forgot to test with tarballs, and links be weird).

Also, pile on yet another stupid Kindle workaround to make sure an OTA update won't break the launcher script itself because of truncation, which would break the exit sequence, leaving the user with an apparently hung device, simply because the WM was never unpaused...

Yay Kindle. -_-"

Fix #5721

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5758)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,7 @@ kindleupdate: all
 	# note that the targz file extension is intended to keep ISP from caching
 	# the file, see koreader#1644.
 	cd $(INSTALL_DIR) && \
-		tar -I"gzip --rsyncable" -cah --no-recursion -f ../$(KINDLE_PACKAGE_OTA) \
+		tar --hard-dereference -I"gzip --rsyncable" -cah --no-recursion -f ../$(KINDLE_PACKAGE_OTA) \
 		-T koreader/ota/package.index
 
 KOBO_PACKAGE:=koreader-kobo$(KODEDUG_SUFFIX)-$(VERSION).zip
@@ -233,7 +233,7 @@ koboupdate: all
 		koreader/ota/package.index koreader.png README_kobo.txt
 	# make gzip koboupdate for zsync OTA update
 	cd $(INSTALL_DIR) && \
-		tar -I"gzip --rsyncable" -cah --no-recursion -f ../$(KOBO_PACKAGE_OTA) \
+		tar --hard-dereference -I"gzip --rsyncable" -cah --no-recursion -f ../$(KOBO_PACKAGE_OTA) \
 		-T koreader/ota/package.index
 
 PB_PACKAGE:=koreader-pocketbook$(KODEDUG_SUFFIX)-$(VERSION).zip
@@ -272,7 +272,7 @@ pbupdate: all
 		applications/koreader/ota/package.index system
 	# make gzip pbupdate for zsync OTA update
 	cd $(INSTALL_DIR)/applications && \
-		tar -I"gzip --rsyncable" -cah --no-recursion -f ../../$(PB_PACKAGE_OTA) \
+		tar --hard-dereference -I"gzip --rsyncable" -cah --no-recursion -f ../../$(PB_PACKAGE_OTA) \
 		-T koreader/ota/package.index
 
 utupdate: all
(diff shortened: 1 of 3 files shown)


NiLuJe:
(Wasn't technically broken (well, more broken than usual :D) in the previous stable, so not putting this in a milestone).

Frenzie:
> (Wasn't technically broken (well, more broken than usual :D) in the previous stable, so not putting this in a milestone).

I'm not sure if that quite makes sense. I see it more as things that are part of it, in any case.

Makefile
```diff
@@ -202,7 +202,7 @@ kindleupdate: all
 	# note that the targz file extension is intended to keep ISP from caching
 	# the file, see koreader#1644.
 	cd $(INSTALL_DIR) && \
-		tar -I"gzip --rsyncable" -cah --no-recursion -f ../$(KINDLE_PACKAGE_OTA) \
+		tar --hard-dereference -I"gzip --rsyncable" -cah --no-recursion -f ../$(KINDLE_PACKAGE_OTA) \
```
Frenzie: This is technically only for Kindle?

(I'm fine keeping it the same everywhere for simplicity; just asking.)

platform/kindle/koreader.sh
```diff
@@ -1,4 +1,15 @@
 #!/bin/sh
+
+# NOTE: Stupid workaround to make sure the script we end up running is a *copy*,
+# living in a magical land that doesn't suffer from gross filesystem deficiencies.
+# Otherwise, the vfat+fuse mess means an OTA update will break the script on exit,
+# and potentially leave the user in a broken state, with the WM still paused...
+if [ "$(dirname "${0}")" != "/var/tmp" ]; then
```
Frenzie: Does it need to be preserved on reboot?

NiLuJe: It's a tmpfs, so it'll go poof on reboot.

I'm fine with leaving however kb this takes when we exit, it's 32MB by default, we've got more than enough space in there ;).

Makefile
NiLuJe: Yeah, although I figured this couldn't hurt, since most of those will be deployed on vfat anyway, and all of them were already using `--dereference` (`h`).

platform/kindle/koreader.sh
NiLuJe: Although, since it's no longer on a crappy FS, it also wouldn't hurt to unlink on exit, now that I think about it ;).

Frenzie: It will? Hmm… odd.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html
>5.15. /var/tmp : Temporary files preserved between system reboots
>5.15.1. Purpose
>
>The /var/tmp directory is made available for programs that require temporary files or directories that are preserved between system reboots. Therefore, data stored in /var/tmp is more persistent than data in /tmp.
>
>Files and directories located in /var/tmp must not be deleted when the system is booted. Although data stored in /var/tmp is typically deleted in a site-specific manner, it is recommended that deletions occur at a less frequent interval than /tmp.

Frenzie: Anyway, I guess I mean that /var/tmp seems needlessly longer than /tmp if there's no special reason to prefer /var/tmp

NiLuJe: I don't have the Kindle plugged in for details, but /var is a tmpfs on Kindle, and /tmp is either not one at all, or another, much smaller one, and that one might be noexec in some setups.

So, yeah, /var/tmp should be a safer bet than /tmp, on Kindle at least ;).

(FWIW, both /tmp and /var/tmp are tmpfs on my desktop ^^).

NiLuJe: The TL;DR being: That was a concerted decision after checking the state of /tmp vs. /var/tmp on various Kindle generations ;).

Hzj-jie:
Interesting, on my voyage, /var is tmpfs, no /tmp entry. E.g. it's rootfs?

NiLuJe:
Possibly (so, triple threat: not a tmpfs, noexec, and, worst of all, ro :D).

Issue #545, opened 2014-04-03
Text to Speech (TTS)


Sawasdee:
Is there anyway to add TTS to the latest koreader, please?   I've seen that in the previous version (kdfviewer) has this ability but not in the latest version.   I tried to learn the TTS code in the previous version (dialog.lua), but it was written in the entirely different way the latest version written.   So I'm really don't know how to modified or use the code.   Any help or suggestion, please?   Thanks a lot.

Korb:
Neither in the [README.md](https://github.com/koreader/koreader#readme) file, nor in the [User Guide](https://koreader.rocks/koreader-user-guide.pdf), nor on [the application site](https://www.google.com/search?q=android+tts+site%3Akoreader.rocks), I found a mention of TTS. Do I understand correctly that this application, at least its version for Google Android, does not have the ability to read open text documents by voice?

poire-z:
> this application, at least its version for Google Android, does not have the ability to read open text documents by voice?

You understand correctly.
There is no code in the app to generate speech and neither do anything related to audio.
See also #9548 and #7975.

foss-:
> Do I understand correctly that this application, at least its version for Google Android, does not have the ability to read open text documents by voice?

Been asking myself the same for a long time.

https://github.com/koreader/koreader/issues/9548 seems to be the correct feature request. While I understand there currently is no interest in implementing a solution, I am not sure I understand why this valid feature request would be closed.

NiLuJe:
> I am not sure I understand why this valid feature request would be closed.

Because it's out of scope. For technical reasons, which then snowballs into a human resources reason, as it would take an enormous amount of work.

ryanwwest:
@NiLuJe would it be appropriate to build a plugin for this, that for example, only works on android to start? I see some of the plugins are only for specific devices, but I don't really know how much work it might take to integrate it with KOReader since it has a totally customized UI.

NiLuJe:
I'm not sure if a plugin would have the required access necessary to even attempt to implement this. But, yes, from a design standpoint, that's probably better. It'would most likely forever be stuck in contrib land because of the aforemenetion reasons, though.

ryanwwest:
[...]

PR #5762, opened 2020-01-11
wip: android - share files

pazos:
bluetooth, mail and social apps are the main targets, but should work with any app that's able to handle SEND intents for a given mimetype.

Requires https://github.com/koreader/android-luajit-launcher/pull/218

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5762)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -311,6 +311,20 @@ function FileManager:init()
                     end,
                 }
             })
+
+            if Device:canShareFile() then
+                local title = _("Share")
+                table.insert(buttons, {
+                    {
+                        text = title,
+                        callback = function()
+                            UIManager:close(self.file_dialog)
+                            Device.doShareFile(file, title)
+                        end,
+                    }
+                })
+            end
+
             table.insert(buttons, {
                 {
                     text_func = function()
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -87,8 +87,9 @@ local Device = Generic:new{
     importFile = function(path) android.importFile(path) end,
     isValidPath = function(path) return android.isPathInsideSandbox(path) end,
     canShareText = yes,
+    canShareFile = yes,
     doShareText = function(text) android.sendText(text) end,
-
+    doShareFile = function(path, title) android.sendFile(path, title) end,
     canExternalDictLookup = yes,
     getExternalDictLookupList = getExternalDicts,
     doExternalDictLookup = function (self, text, method, callback)
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -47,6 +47,7 @@ local Device = {
     hasBGRFrameBuffer = no,
     canImportFiles = no,
     canShareText = no,
+    canShareFile = no,
     canToggleGSensor = no,
     canToggleMassStorage = no,
     canUseWAL = yes, -- requires mmap'ed I/O on the target FS


poire-z:
What's the use case ? :)
(Sharing a file to another phone/tablet from KOReader file browser? is it really worthy for it be do'able from KOReader? May be sending progress & highlights would be more useful?)

Anyway, a few notes:
- You're adding a full-width `Share` button, on all Android - in between 2 2-buttons rows. A bit aesthetically unpleasant I guess, and making wide a small/rarely used button. No real suggestion though :| Would be nice if we could  have one more button in a new 2-buttons row for Android only too :)
- Beware that coverbrowser covermenu.lua replaces/inserts these buttons by `[row][column]` position, so beware of side effects. Some details in https://github.com/koreader/koreader/pull/5527#issuecomment-548265851. Please check it works fine in both classic mode, and one of list/mosaic modes.

pazos:
> What's the use case ? :)
> (Sharing a file to another phone/tablet from KOReader file browser? is it really worthy for it be do'able from KOReader?...)

Good question :smile: My user case is very specific. I send *some* files to termux.  Including the quicknotes I take from KOReader itself (using markor) :p 

I guess other can send files via Whatsapp, Telegram, Dropbox, Gmail or any of the other apps that support the mechanism. Since it is possible to display unsupported documents it is actually possible to do other kind of fancy -useless- stuff, like
test


> Anyway, a few notes

Thank you! I tested with cover browser but not without it. Here is the difference:

**with** coverbrowser

device-2020-01-11-221827

**without** coverbrowser

device-2020-01-11-221744


> May be sending progress & highlights would be more useful?

Yeah, sure. I won't be too bothered if this "enhancement" never reaches the program. I can use it locally. But the (wip) backend is, IMO, a good way to "export" content from KOReader to other systems/apps. Nowadays is probably unneded, but with enforced sandboxing in the future is the only way to send documents outside the program.

Anyways, I guess we can add a setting in Settings->Device and just display these options if the user wants *or* sandbox is enforced.

poire-z:
Might be better to add your "Share" button after the (optional) "Convert", so CoverBrowser can prepend the "Mark as read" button to the originally alone "Add to favorite".
So your "Share" would come after - and before the buttons CoverBrowser adds at the bottom.

>  the only way to send documents outside the program.

Except for your uselessusecase :) why would we want/need to send documents outside the program? Would we need that even for the progress sync/kosync (whatever it's called) feature? I hope we can still do HTTP requests :)

> Anyways, I guess we can add a setting in Settings->Device and just display these options if the user wants or sandbox is enforced.

I'd prefer that :) Your use case is indeed quite remote to what a book reading app should be :)

No problem with getting the backend part in the program, and your added button (shown only when a manual setting, or with a menu item quite deep :) (and full-width if you don't mind that - I would but I won't see it :) after the generic buttons, so we don't have to complicate the buttons hacking on the CoverBrowser side.

pazos:
Thanks for all the feedback @poire-z.  I will close this PR and reimplement in the future with your suggestions! 

> Except for your uselessusecase :) why would we want/need to send documents outside the program? Would we need that even for the progress sync/kosync (whatever it's called) feature? I hope we can still do HTTP requests :)

without sandbox we have read access in a bunch of places and RW in /sdcard. Settings in /sdcard/koreader will be stored even when the program is deleted.

with sandbox we have RW in /sdcard/Android/data/org.koreader.launcher/files and the settings (stored in /sdcard/Android/data/org.koreader.launcher/files/koreader) and **all** the books will be deleted if/when the app is deleted.

And yes, http is still valid via native code, but was flagged as deprecated in the SDK, which now requires https, so we need [this flag](https://github.com/koreader/android-luajit-launcher/blob/master/app/manifest/base.xml#L21) to download app updates.

ryanwwest:
Btw, would love to see this sharing of files brought in - the 'usefulusecase' that a few others have requested would be allowing easy TTS by using a third-party reader that supports TTS, and supplying it with the absolute filepath of the currently-opened book (for details: https://github.com/koreader/koreader/issues/545#issuecomment-1529180324). I think there are other use cases as well, though, even like opening the file in a file explorer app like MiX.

I think https://github.com/koreader/android-luajit-launcher/pull/218 is related.

pazos:
> Btw, would love to see this sharing of files brought in - the 'usefulusecase' that a few others have requested would be allowing easy TTS by using a third-party reader that supports TTS, and supplying it with the absolute filepath of the currently-opened book (for details: [#545 (comment)](https://github.com/koreader/koreader/issues/545#issuecomment-1529180324)). I think there are other use cases as well, though, even like opening the file in a file explorer app like MiX.
> 
> I think [koreader/android-luajit-launcher#218](https://github.com/koreader/android-luajit-launcher/pull/218) is related.

A lot of things changed since 2020. 
We now have both userpatches and support for 3rd party plugins.

If you really want to do that on your machine it is a matter of doing something like https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/ApkUpdater.kt#L62-L73 into a function called `android.sendFile(path)`

And then use `android.sendFile("/full/path/to/file")` in your patch or user plugin. I think there're alrady some user plugins that hook on that specific menu, so it looks like a 5 minutes hack.

ryanwwest:
Good to know. Are you saying that since sharing involves OS-specific operations, it would need to be integrated as e.g. a 3rd party plugin rather than be integrated by default? I could see this as somewhat cross platform even so (for example, the file explorer and loading files must have OS-specific code but are not plugins). But maybe that doesn't make sense.

I'd love to contribute (and in fact have a side project doing highlight export/import to Zotero and KOReader), but don't currently have the time to get into this codebase and make a good tested change. Thanks for pointing out the place to start, though—if I want to scratch the itch hard enough maybe I'll be able to do it over some future holiday.

pazos:
> Good to know. Are you saying that since sharing involves OS-specific operations, it would need to be integrated as e.g. a 3rd party plugin rather than be integrated by default?

Since TTS will always be something out of scope for this app the only thing that you can do to get what you want is build your own plugin/patch.

It also has some advantages to roll your own even when a feature is actually wanted: no code reviews, no testing, no nothing.. so you can choose to ship a full fledged implementation or something that barely fits your needs.
You can also choose to not release it and keep it for your private usage if you feel you don't want to give support for it.

ryanwwest:
> Since TTS will always be something out of scope for this app the only thing that you can do to get what you want is build your own plugin/patch.

For the case of TTS specifically, I agree. But sharing a filepath outside of KOReader is much more general than that, and while it _could_ be used for TTS in another app, it could just as well be used for many other reasons in another app. I used TTS as an example motivating use case, but the implementation of sharing a book filepath from KOReader to another app seems general enough to maybe warrant a more core implementation.

pazos:
The only core implementation we need is to make `android.sendFile(path)` available. All the reasons to use that can be implemented as a 3rd party plugins.

The only real limitation of a 3rd party plugin nowadays is that you need to do your own translations. But you don't need translations if you use _("Share"), _("Close") or any other string already covered by our own translations :)

PR #5765, opened 2020-01-14
fix, confusing info/error message when exporting bookmarks

mustafa-001:
When multiple bookmarks exported Koreader shows following message/error:
nothingwasexportedbug

This PR removes `all_count == 1` check (holds `false` with multiple books). 

I guess it was added to decide which message to display but string template already does that.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5765)

### DIFF
--- a/plugins/evernote.koplugin/main.lua
+++ b/plugins/evernote.koplugin/main.lua
@@ -575,25 +575,21 @@ function EvernoteExporter:exportClippings(clippings)
     local msg = "Nothing was exported."
     local all_count = export_count + error_count
     if export_count > 0 and error_count == 0 then
-        if all_count == 1 then
-            msg = T(
-                N_("Exported notes from the book:\n%1",
-                   "Exported notes from the book:\n%1\nand %2 others.",
-                   all_count-1),
-                export_title,
-                all_count-1
-            )
-        end
+        msg = T(
+            N_("Exported notes from the book:\n%1",
+                "Exported notes from the book:\n%1\nand %2 others.",
+                all_count-1),
+            export_title,
+            all_count-1
+        )
     elseif error_count > 0 then
-        if all_count == 1 then
-            msg = T(
-                N_("An error occurred while trying to export notes from the book:\n%1",
-                   "Multiple errors occurred while trying to export notes from the book:\n%1\nand %2 others.",
-                   error_count-1),
-                error_title,
-                error_count-1
-            )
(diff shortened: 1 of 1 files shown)


plugins/evernote.koplugin/main.lua
```diff
@@ -575,25 +575,21 @@ function EvernoteExporter:exportClippings(clippings)
...
-                N_("Exported notes from the book:\n%1",
-                   "Exported notes from the book:\n%1\nand %2 others.",
-                   all_count-1),
-                export_title,
-                all_count-1
-            )
-        end
+        msg = T(
+            N_("Exported notes from the book:\n%1",
+                "Exported notes from the book:\n%1\nand %2 others.",
```
Frenzie: Would you mind keeping the indentation as it was for aesthetic reasons? (Not that this isn't logical.)

mustafa-001: Don't know why but both VS Code and Vim gave me a fight over one character.

plugins/evernote.koplugin/main.lua
```diff
@@ -575,25 +575,21 @@ function EvernoteExporter:exportClippings(clippings)
...
-                   "Exported notes from the book:\n%1\nand %2 others.",
-                   all_count-1),
-                export_title,
-                all_count-1
-            )
-        end
+        msg = T(
+            N_("Exported notes from the book:\n%1",
+               "Exported notes from the book:\n%1\nand %2 others.",
+                all_count-1),
```
Frenzie:
```suggestion
               all_count-1),
```

plugins/evernote.koplugin/main.lua
```diff
@@ -575,25 +575,21 @@ function EvernoteExporter:exportClippings(clippings)
...
-                   "Multiple errors occurred while trying to export notes from the book:\n%1\nand %2 others.",
-                   error_count-1),
-                error_title,
-                error_count-1
-            )
-        end
+        msg = T(
+            N_("An error occurred while trying to export notes from the book:\n%1",
+               "Multiple errors occurred while trying to export notes from the book:\n%1\nand %2 others.",
+                error_count-1),
```
Frenzie:
```suggestion
               error_count-1),
```

Frenzie:
Or I would but you I guess you didn't tick the write permission box.

Issue #5766, opened 2020-01-14
Koreader getting slower and slower as I add highlights


abdoulayeYATERA:
* KOReader version: v2019.12-13-g59deaf6_2020-01-02
* Device: Kobo format

#### Issue
 While reading big books, KoReader get slower and slower as i add highlights until it's barely usable anymore.
This is the folder containing the book and all my highlights and metada so far.
[Mises_human_action.zip](https://github.com/koreader/koreader/files/4058676/Mises_human_action.zip)
The book : Human action by Mises
No copyright issue, the book is free and provided by https:///mises.org

As you'll see the metadata file is quiet big .... 5000 lines, 500kb.
I have to note that if I open an other book everything is fine.
A help will be appreciated !
Thank's !

#### Steps to reproduce
- read a big book
- add many highlights

poire-z:
Related/duplicate: #3261 #3503 (but let's keep this issue open).
#3503 is actually the exact same/duplicate of this issue, thas has been opened 2 years ago by ... *you* :) and that I had hope it was fixed by #3508 - that you confirmed :).

Do you think it has gotten worse at some point in time since?
Or it's just that this is a bigger book with very much more highlights (250) than what you usuall read/do?

Anyway, thank you for the zip with the epub AND the sdr. Never seen that many highlights :) it's good stuff to see/test what we can do.

abdoulayeYATERA:
"the exact same/duplicate of this issue, thas has been opened 2 years ago by ... you :) " X)
Hi Poire-z ! Yeah, i thought the symptoms were the same but not the cause.
That's why I openned a new issue.

abdoulayeYATERA:
This book contains clearly much more highlights that usually....

abdoulayeYATERA:
And I can add this, the navigation between pages is slow but i can cope with it, however, the text selection is really tough !

poire-z:
OK, so it's the 500 calls to `getPosFromXPointer(xp)` (to get the pages the start and the end of each of your 250 highlights xpointers/xpath are, to see if they are part of the current page and if we need to get more details about them to draw the highlights) that are happening on each page turn, that are causing your slow down.
#3508 was about doing that, using `getPosFromXPointer(xp)` to check for the page, before doing the more expensive "getting more details" work - so, it's good it is that way.
Nothing much we can do to speed up these calls, 500 x 1 or 2ms, and you get your slowdown.
But...:

[...]

PR #5773, opened 2020-01-16
cre: add toggable crengine call cache

poire-z:
Optimise usage of some of CreDocument methods by caching their results, either globally, or per page/pos for those whose result may depend on current page number or y-pos.
This helps a lot on cre books with a lot of highlights. More info in https://github.com/koreader/koreader/issues/5766#issuecomment-574410890.

Currently off by default, can be enabled via File browser > Developer options. Please do so to help testing it. We'll see how it does before enabling it by default.

Also adds OPDS for https://gallica.bnf.fr (not super interesting, but I was testing it, found some possible bug in their output, and added some "fix" to workaround it, so, might as well add it to the list to justify the fix :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5773)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -440,6 +440,27 @@ function FileManagerMenu:setUpdateItemTable()
             }
         }
     })
+    table.insert(self.menu_items.developer_options.sub_item_table, {
+        text_func = function()
+            if G_reader_settings:isTrue("use_cre_call_cache")
+                    and G_reader_settings:isTrue("use_cre_call_cache_log_stats") then
+                return _("Enable CRE call cache (with stats)")
+            end
+            return _("Enable CRE call cache")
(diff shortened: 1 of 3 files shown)


poire-z:
Yep, a good amount of code, but a lot less if you remove the one accounting stats :) It's quite generic thus - there's some rarely used stuff like select/unpack, and the possibly unoptimized asString... But well, indeed, can't be helped.

> wondering if it should be split up into a crecache module in its own file. :-)

Thought about it, but decided not to: it's the same object we hack and override methods, and it refers to methods defined above - so better be in the same file, at its bottomest bottom, so we more easily get the full picture and less chances to forget to update it when we add new methods :)

(Ideally, if Lua supported stuff like python decorators, we could tag each methods as we define them - but it doesn't I think. But it's as good with the full list of the method names we deal with, easier to grasp what can be cached, what resets cache...)

Issue #3690, opened 2018-02-22
How to export highlights as plain text


agaAdams:
* KOReader version: nightly v2015.11-1525-gca6a486
* Device: KA1

#### Issue
I'm not sure how to export highlights from my documents to a plain text format 
#### Steps to reproduce
Menu > Tools Icon > Evernote
- export to local HTML files
- export to local clipping text file

If I check the option to export to text file, only ONE menu option to export becomes available: "Export all notes in your library"
Activating this option creates a "KOReaderClipping.txt" file with ONLY the titles of books with highlights. No highlights are included.

If I choose export to HTML, BOTH menu options to export become available and produce the intended html-files.

Is exporting to plain text somehow broken or am I doing this the wrong way?

Frenzie:
I just gave it a quick try and both options seem to work for me.
```
quickstart
　
　　-- Page: 1, added on Thu Feb 22 16:55:39 2018
the
-=-=-=-=-=-

quickstart
　
　　-- Page: 2, added on Thu Feb 22 16:56:48 2018
language and other settings through the gear icon.
File
-=-=-=-=-=-
　　-- Page: 3, added on Thu Feb 22 16:56:53 2018
folders
-=-=-=-=-=-

[clipboard.zip](https://github.com/koreader/koreader/files/1748574/clipboard.zip)

agaAdams:
This is how my menu looks in koreader:
![01 koreader_menu](https://user-images.githubusercontent.com/22838606/36618001-bd78fb32-18e9-11e8-85ed-1f22e40a3c34.jpg)

The message I get after exporting all notes:
![02 koreader_message](https://user-images.githubusercontent.com/22838606/36618005-c1142f5a-18e9-11e8-9443-c1d676a9fff1.jpg)

The plain text file I end up with after exporting:
![03 koreader_txt_file](https://user-images.githubusercontent.com/22838606/36618007-c360d06a-18e9-11e8-9a19-1d43909ecf09.png)

One of the html files generate for one of the books:
![04 koreader_html_file](https://user-images.githubusercontent.com/22838606/36618009-c4da2edc-18e9-11e8-9e84-a650bf9306c1.png)

What I don't understand is why I get all the highlights exported per html but only the one from the first book per plain text.

Frenzie:
Did you try the Suble Art book independently? In the list it looks like that's where things go wrong, because the Paleo Approach did apparently work out.

As an aside, those are some nice encoding issues in your screenshot of the HTML export.

[...]

Issue #4525, opened 2019-01-28
Feature request - highlighting images and text in .epub files and possibility of exporting them to  


konjski-tramvaj:
* KOReader version:
* Device:

#### Feature request:

Was wondering is it even possible to implement functionality of exporting highlights from a certain book with highlighted images or other media found in .epub file. Could something like this be even done to .epub files?
Was fantasizing of func. which would store all highlights in .html file but how could we "cut" (highlight) parts of .epub file?

poire-z:
I don't really see the purpose  :|

But we have mostly everything to do that:
- the code used when selecting text and then `View HTML` can give you a completely valid and balanced HTML snippet containing the selected text, including references to stylesheets, keeping any `<img src=filepath_in_the_epub>` in the HTML (you could then extract the IMG from the HTML with some regex matching).
- that same code allows extracting stylesheet files content and viewing it: you could do the same to extract CSS content as well as image files content, that you could then save as local files
- if you'd rather want to make a self-contained EPUB from all these, there is our Save Wikipedia article as EPUB code that does that kind of thing (building an EPUB from some HTML, CSS and image content downloaded from the net).

But once you've done that, you've done the most fun part. But you still have to provide some way to organise what you just save: allow ordering or removing previous selections, giving them names if local files, or editing/updating the currently appended EPUB...

But if you don't really care about images, there is already some stuff (I just discovered some weeks ago) to export highlights' text into a HTML file, see https://github.com/koreader/koreader/issues/1095#issuecomment-453726356

konjski-tramvaj:
> 
> 
> I don't really see the purpose :|
> 
> But we have mostly everything to do that:
> 
>     * the code used when selecting text and then `View HTML` can give you a completely valid and balanced HTML snippet containing the selected text, including references to stylesheets, keeping any `<img src=filepath_in_the_epub>` in the HTML (you could then extract the IMG from the HTML with some regex matching).
> 
>     * that same code allows extracting stylesheet files content and viewing it: you could do the same to extract CSS content as well as image files content, that you could then save as local files
> 
>     * if you'd rather want to make a self-contained EPUB from all these, there is our Save Wikipedia article as EPUB code that does that kind of thing (building an EPUB from some HTML, CSS and image content downloaded from the net).
> 
> 
> But once you've done that, you've done the most fun part. But you still have to provide some way to organise what you just save: allow ordering or removing previous selections, giving them names if local files, or editing/updating the currently appended EPUB...
> 
[...]

Issue #3415, opened 2017-10-23
Bookmark titles or comments?


embryo10:
This is not an issue.
On the contrary, its about a new (as of 2015.11-1370) feature, the Tap-edit bookmarks #3369.
Now that we can just click on a highlight and change its title, I think that its almost like adding a comment *about* the highlight.
I did read them with the [KoHighlights](https://github.com/noembryo/KoHighlights) app like this
![screen1](https://user-images.githubusercontent.com/8684739/31944052-769e77b2-b8d3-11e7-9ac7-e14386e35ccf.png)
and I wonder if I should call them comments.
If there is a plan (or a reason) to add another kind of comments to the highlights, then I should call these bookmarks titles something different to avoid the confusion.
Any suggestions?

Frenzie:
@poire-z @robert00s ?

embryo10:
While adding the ability to edit the title/comment from the PC, I observed that although the "Edit" dialog does not let you create a multi-line text, it does shows it to you (if it is) and lets you edit it.

Since the bookmarks seems to be working fine, do you think it is dangerous to not restrict the text to a single line?

FYI the new line is written the same way as the new line in the highlighted text.

Frenzie:
Sounds more like a missing feature on the KOReader side to me. :-)

KenMaltby:
I tried to add text to a highlighted paragraph, as a note, using the "rename" function, when it was first proposed.  I could add a note but it wouldn't save, with the highlight.  I saw it as a good way to add a much asked for note taking function/feature.  I didn't try it in a .pdf.  It might have worked there as the highlight is reportedly added directly into the .pdf itself.

embryo10:
The "Edit"/"Rename" function of the reader works fine in the 2015.11-1370 version.
This is how I enter the initial titles and started the experiments...
Tested in epub and pdf.

Frenzie:
It does save it but wrt highlight there's some kind of display issue.

```lua
        [3] = {
            ["page"] = "/body/DocFragment[3]/body/div/div[1]/p/text().720",
            ["highlighted"] = true,
-- this is displayed in the details view
            ["notes"] = "began\
[...]

PR #5774, opened 2020-01-16
[fix, plugin] Evernote exporter only writes one documents clippings when txt export used, rest is ignored

mustafa-001:
When using txt export with multiple books koreader produces an output like this: [bugged_exports.txt](https://github.com/koreader/koreader/files/4073633/bugged_exports.txt) Only first entry has clippings. fixes https://github.com/koreader/koreader/issues/3690

On every `booknotes` `:exportToTXT()` compares `booknotes.time` with file modification time. After first run this modification time is updated to current time, then comparison returns false. 
I think idea behind it was a .txt file for every book without using flag like `exported_stamp`. Or an artifact from here: https://github.com/koreader/koreader/blob/b111ccc2b803fc78425d9791fcc8bb7a9a704e6e/frontend/apps/reader/modules/readerhighlight.lua#L1134-L1136

Solution
I think best solution here is rewriting .txt file everytime. To append on existing file we must locate previous entry from book, find end of it and delete it; overly complicated process for little reason. In other export schemes flag method works because every book has its own output file.

In my plan, we delete KOReaderClippings.txt file at every export, and append to it book by book overriding `exported_stamp`.

Also I wrote a small explanation at `:exportAllNotes()` about how this highlight thing works and a unit test if I got it right.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5774)

### DIFF
--- a/plugins/evernote.koplugin/main.lua
+++ b/plugins/evernote.koplugin/main.lua
@@ -485,14 +485,16 @@ Parses highlights and calls exporter functions.
 Entry point for exporting highlights. User interface calls this function.
 Parses current document and documents from history, passes them to exportClippings().
 Highlight: Highlighted text or image in document, stored in "highlights" table in
-documents sidecar file. Parser uses this table.
-Bookmarks: Text in bookmark explorer, user can edit this fields. If user didn't
-edit highlight or "renamed" bookmark this is not created. Stored in "bookmarks" table
-in documents sidecar file. Parser looks to this file for edited notes.
+documents sidecar file. Parser uses this table. If highlight._._.text field is empty parser uses
+highlight._._.pboxes field to get an image instead.
+Bookmarks: Data in bookmark explorer. Stored in "bookmarks" table of documents sidecar file. Every
+field in bookmarks._ has "text" and "notes" fields When user edits a highlight or "renames" bookmark,
+text field is created or updated. Parser looks to bookmarks._.text field for edited notes. bookmarks._.notes isn't used for exporting operations.
+https://github.com/koreader/koreader/blob/605f6026bbf37856ee54741b8a0697337ca50039/plugins/evernote.koplugin/clip.lua#L229
 Clippings: Parsed form of highlights, stored in clipboard/evernote.sdr/metadata.sdr.lua
-for all documents. Used only for exporting bookmarks. Internal functions does not use
-this table.
-Booknotes: Every table in clippings table. Clippings = {"title" = booknotes}
+for all documents. Used only for exporting bookmarks. Internal highlight or bookmark functions
+does not usew this table.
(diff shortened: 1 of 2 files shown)


poire-z:
Can't comment much as I don't know (and don't want to know) this plugin :)
I read your comment, and just pointing you to past posts where I wrote about the `bookmarks` and `highlight` tables in sidecar files, so you can check if we understand them the same way:
https://github.com/koreader/koreader/issues/3415#issuecomment-338996571
https://github.com/koreader/koreader/issues/1095#issuecomment-458918504
(And just in case you want to follow up on fixing it: https://github.com/koreader/koreader/issues/1095#issuecomment-453726356)

mustafa-001:
https://github.com/koreader/koreader/issues/1095#issuecomment-453726356 is somewhat fixed at https://github.com/koreader/koreader/pull/5440 but I don't think it's reliable. For example scraping page-date info logic doesn't respect locale, bookmarks without highlights are annoying and having two different source of data bugs me. 

About note taking there are so different requests in related issues: write a note randomly anywhere independent of page, edit a bookmark even scrape html https://github.com/koreader/koreader/issues/4525. Knowing what to do is harder than doing it :) Maybe make a plugin for each use case (a plugin for each person  that is :smile:) and deactivate all by default.

poire-z:
> having two different source of data bugs me

Yes, me too, it's a big mess, but it would be quite some work to clean all that, migrate old stuff to new structures (which would have to be followed up in external app KOHighlight...)

> bookmarks without highlights are annoying 

and in your comment in the code:
> Bookmarks: Text in bookmark explorer, user can edit this fields. If user didn't
edit highlight or "renamed" bookmark this is not created. 

It looks to me that everything should be there from the start in `bookmarks`. Bookmarks indeed includes bookmarked pages, but they have no text/pos0/pos1.  `highlight` includes some of the stuff in bookmarks but not some other, and not bookmarked pages, and it's mostly a duplicate , except it has a field "drawer" (lighten, underline) telling how it should be highlighted.
So, it feels that plugin could just use `bookmarks` (I see you didn't change any logic in the parsing, so no need to update it).
Just saying your comment might not be accurate (unless you meant with `Bookmarks: Text` : `bookmarks[].text` in which case the capitalization and plural "fields" fooled me :).

mustafa-001:
> Just saying your comment might not be accurate (unless you meant with Bookmarks: Text : bookmarks[].text in which case the capitalization and plural "fields" fooled me :). 
Yes, I guess it's confusing. Thanks for pointing out.

`bookmarks` lacks `pboxes` to export images (for Evernote, not sure if that works), and duplicating them would mean twofold increase in sidecars file size. I think changing all highlight logic to use `bookmarks` as default is way to go. 

But for now I'm thinking a PR to use `bookmarks` in parser and falling back to `highlights` for images.

Issue #4010, opened 2018-06-13
chapter title option in status bar


kokoreader:
* KOReader version: all
* Device: KA1

Good day.

I was wondering if the option to display the chapter title in the status bar at the bottom of the screen could be implemented. Thank you.

cramoisi:
my two cent : title and chapter name would be great !

AlanSP1:
Well, title of book and of chapter, it may better be placed in upper part, but if it goes down, it maybe "asks" for third line in status bar. I already asked for two lines status bar here: https://github.com/koreader/koreader/issues/1933

As we now have many options to see, having book and chapter title can take one line more, as it can get pretty big.

cramoisi:
you don't need to enable everything.... I could be just an option to enable or not for people who want it. For myself, I enable page i am / total and number of page left int the chapter. Plenty of place left.

cramoisi:
@AlanSP1 : I just do the opposite : I activate the less option. Everything which is not immediately relative to reading I have disabled :-P

KenMaltby:
With the "Full" progress bar (with the cr3 "Status" bar that has the book title, at the top of the page) you can still have the "Mini" progress bar also (although you may have to tap the bottom right corner, to bring it back up), so you do have two lines.  One at the top and one at the bottom.  Maybe, with the page numbering removed from the top, there could be room for a short chapter title.  Would be some trick though messing with the cr3 code and grabbing the chapter title some how.

cramoisi:
to note : all the options to remove things in the full-header-status-c3-bar are somehow functional from cr3.ini : displaying or not : page you are / total page, battery icon, battery status (disabled by default), time.

But really, this full-header-status-c3-bar is ugly :) and it's far better to use the little mini-bar which can be hidden and displayed so easily.

AlanSP1:
@cramoisi Yes, I agree that we all have our own different customs and no one is same.

For me, I don't look at all information that is offered all the time, but from time to time I want to know some of info that is there. As getting some info that isn't included at the moment would require going to setting and enabling, for me, it is easier to already include all info, e.g. info on memory usage (to know when to restart koreader, just in case).

Anyway, that's why I think that all that info should be optional, that we all can set what we prefer.

LucyOne:
I would also like to see the chapter title in the status bar. A rotation with author name - book title - chapter title would be great.

PR #5775, opened 2020-01-17
Option to show title and current chapter in footer

robert00s:
Close: #4010 
Option to show title and current chapter of book in footer. 
Default width is 30% of screen size and can be configurable by `Status bar -> Settings -> Maximum width of items`

I need help @poire-z with RTL when title or chapter is truncate and ended with ellipsis. For LTR I use `RenderText:sizeUtf8Text()` and `RenderText:truncateTextByWidth()` in `book_title` and `book_chapter` local function.
Should I use trick with reversed text like in TextWidget?
https://github.com/koreader/koreader/blob/b111ccc2b803fc78425d9791fcc8bb7a9a704e6e/frontend/ui/widget/textwidget.lua#L122-L144

Screenshot:

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5775)


poire-z:
> For LTR I use RenderText:sizeUtf8Text() and RenderText:truncateTextByWidth() in book_title and book_chapter local function.
Should I use trick with reversed text like in TextWidget?

My idea was that we shouldn't use RenderText stuff anymore. Since #5598, Text*Widgets do not use it, and use the xtext C module instead.
Looks like we always could just let TextWidget do the truncation, and we never really needed to know where it would truncate, until this PR :)
I'll look later today how we could get what you need. Might just be an additional method to TextWidget: you would instantiate a TextWidget with a max_width, and could get back the logical string up to the truncation, and delete the TextWidget without ever drawing it.

robert00s:
> I'll look later today how we could get what you need. Might just be an additional method to TextWidget: you would instantiate a TextWidget with a max_width, and could get back the logical string up to the truncation, and delete the TextWidget without ever drawing it.

Thanks. I tried to create method getTextToDraw() that return logical string but I gave up. I'm not familiar with xtext :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -807,6 +868,61 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                            local SpinWidget = require("ui/widget/spinwidget")
+                            local items = SpinWidget:new{
+                                width = Screen:getWidth() * 0.6,
+                                value = self.settings.book_chapter_max_width_pct,
+                                value_min = 10,
+                                value_step = 5,
+                                value_hold_step = 20,
+                                value_max = 100,
+                                title_text =  _("Maximum width"),
+                                text = _("Maximum book current chapter width in percentage of screen width"),
```
Frenzie: Another book current chapter, but I'm not sure the current is really necessary here.
```suggestion
                                text = _("Maximum current chapter width in percentage of screen width"),
```
```suggestion
                                text = _("Maximum chapter width in percentage of screen width"),
```

poire-z:
OK, so it will need #5776 and https://github.com/koreader/koreader-base/pull/1031.
I have just tested that with logging the result from various truncated text in the UI, but not with your code from this PR.
Please have a go at trying to use it (with the base xtext.cpp patch) with this PR, to confirm I have it working correctly :) Might be used like this:
```lua
local TextWidget = require("ui/widget/textwidget")
local textw = TextWidget:new{
    text = "This is a very long text",
    max_width = 100,
    face = Font:getFace("cfont", 20),
}
local fitted_text, add_ellipsis = textw:getFittedText()
textw:free()
if add_ellipsis then
    fitted_text = fitted_text .. "..." -- with the real ellipsis code :)
    -- that's logical text, so add it to the end, no need to bother if LTR or RTL
end
print("#"..fitted_text.."#")
```

Also, I don't think you need this to be translatable: `T(_("%1 %2"), prefix, doc_info.title)`
T() should be enough. Or just use .. like everywhere else. But you might need to BD.wrap() the title.
so, may be just `title_text = prefix .. " " .. BD.wrap(doc_info.title)`

poire-z:
Here's an arabic epub with a bidi title and some sufficiently long chapter names, if you want to test how RTL text displays in the footer: [Jordanie.AR.epub.zip](https://github.com/koreader/koreader/files/4077294/Jordanie.AR.epub.zip).
Also, when you're done, have a try at seeing how it goes with the untested code I added for when `not use_xtext` by disabling its use in Developpers option.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -296,9 +341,17 @@ function ReaderFooter:init()
             end
         end
     else
+        if not self.settings.max_possible_modes then
+            self.settings.max_possible_modes = 13
```
poire-z: I guess you want to force add these new items to the user already chosen set of items.
It's good for advertising the new stuff, but I'm not sure it's a really friendly decision :) I would just not mess with it, people will discover them if they need them.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -68,6 +75,8 @@ local symbol_prefix = {
         frontlight = "☼",
         mem_usage = "",
         wifi_status = "",
+        book_title = "✎",
+        book_chapter = "☝",
```
poire-z: Might find nicer icons in the nerdfont @NiLuJe recently added (he uses some of them for the battery icon). https://github.com/ryanoasis/nerd-fonts
Or let these (or just one?) without icons/prefix. They should be sufficiently different from all the other items (which are mostly numbers) so there's no doubt about what they are.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -234,6 +243,40 @@ local footerTextGeneratorMap = {
...
+            local chapter_text = T(_("%1 %2"), prefix, chapter_title)
+            local face = Font:getFace(footer.text_font_face, footer.settings.text_font_size)
+            local tsize = RenderText:sizeUtf8Text(0, Screen:getWidth(), face, chapter_text, true, false)
+            if tsize.x > footer._saved_screen_width * footer.settings.book_chapter_max_width_pct / 100 then
+                chapter_text = RenderText:truncateTextByWidth(chapter_text, face,
+                    footer._saved_screen_width * footer.settings.book_chapter_max_width_pct / 100)
+            end
+            return chapter_text
+        else
+            return T(_("%1 N/A"), prefix)
```
poire-z: Or just an empty string for books with no chapter?
Or even `nil`, and have them skipped in the relevant code, to not include at all the item and its separator?

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: I'm pretty sure there's at least an (open) book icon, yeah.

But I think I agree with @poire-z that these don't necessarily need a prefix/icon ;).

Frenzie: +1 There's no need for an icon just like there's no need for an icon with the date or something.

NiLuJe: c.f., a search fro `book` on https://www.nerdfonts.com/cheat-sheet

(The codepoint won't match with our version, ping me if you find something you like, I'll pull the CP for you).

AlanSP1:
Well, this opens one more question, at least if you use footer I as I do: You want to have them all visible. :)

I guess we can approach this in two ways (at least I think so):

1. We can add third line for book and chapter titles (of course, as an option for users that already use all other footer items), otherwise it goes where other options go.

2. We can place it with reading progress if it is in separate line leaving other line for other footer options, or it goes as before with all other items.

What do you think about this? And I guess it would be in different PR, if realized. And yes, I really use all options there are. Maybe I can be treated by good doctors for this? :)

frontend/apps/reader/modules/readerfooter.lua
robert00s: I also like idea to not add icons/prefix to title and chapter.

frontend/apps/reader/modules/readerfooter.lua
robert00s: No :)
This is needed to see new new items (title and chapter).
Look at sample settings before this PR:"

```
    ["footer"] = {
        ["order"] = {
            [0] = "off",
            [1] = "page_progress",
            [2] = "time",
            [3] = "pages_left",
            [4] = "battery",
            [5] = "percentage",
            [6] = "book_time_to_read",
            [7] = "chapter_time_to_read",
            [8] = "frontlight",
            [9] = "mem_usage",
        },
[...]
```
and code:

```
    if not self.settings.order then
        self.mode_nb = 0
        self.mode_index = {}
        local mode_tbl = {}
        for k,v in pairs(MODE) do
            mode_tbl[v] = k
        end
        local mode_name
        for i = 0, #mode_tbl do
            mode_name = mode_tbl[i]
            if mode_name == "wifi_status" and not Device:isAndroid() then
            elseif mode_name == "frontlight" and not Device:hasFrontlight() then
            else
                self.mode_index[self.mode_nb] = mode_name
                self.mode_nb = self.mode_nb + 1
            end
        end
    else
        self.mode_index = self.settings.order
        self.mode_nb = #self.mode_index
    end

```
and almost always we run `else` statement and `self.mode_index` take `self.settings.order` so we see only old modes (without title and chapter). When we need to add new ones we need to add flag `self.settings.max_possilble_modes`. In the future after adding new modes we need increase max_possible_modes and 
`self.settings.order[self.mode_nb+1] = "whatever"`

I hope I explained clearly :)

robert00s:
@poire-z
Something is wrong with RTL (no problem with LTR). Look:





That's totally unclear to me:

poire-z:
> Something is wrong with RTL (no problem with LTR). Look:
> That's totally unclear to me:

Welcome to bidi (hell) land ! :)
(I met something like that in https://github.com/koreader/koreader/issues/5359#issuecomment-570490169 and had it fixed with https://github.com/koreader/koreader-base/pull/1025.)

But may be I just gave you the wrong advice of using `BD.wrap()` (which is a no-op and don't wrap at all in LTR UI :) - and you should use `BD.auto()`.
I still fear it won't have https://github.com/koreader/koreader-base/pull/1025 do its magic as it is not xtext that adds the ellipsis but you explicitely... But have a try before I start scratching my head :)

frontend/apps/reader/modules/readerfooter.lua
poire-z: You did, thanks :)
OK, so that's just the ordering, and they won't be enabled be default?
I get that when ordering, the user orders all the items, so yes, the order table contains them all.
But it looks like you don't need `self.settings.max_possible_modes` and neither that hardcoded `13`, as you have all the existing items in `local MODE = {` on line 28. So, may be something like:
```
while #self.settings.order < #MODE do
  self.settings.order[#self.settings.order+1] = MODE[#self.settings.order+1]
end
```
but ok, with that, you're not skipping wifi_status and frontlight...
So, may be just merge these 2 branches (the rare if and the almost always else), just using the if code, and you would skip the ones already in:
```diff
        for i = 0, #mode_tbl do
            mode_name = mode_tbl[i]
            if mode_name == "wifi_status" and not Device:isAndroid() then
                do end -- luacheck: ignore 541
            elseif mode_name == "frontlight" and not Device:hasFrontlight() then
                do end -- luacheck: ignore 541
+            elseif self.settings.order and i+1 < #self.settings.order then
+                -- already in
+                do end -- luacheck: ignore 541
            else
                self.mode_index[self.mode_nb] = mode_name
                self.mode_nb = self.mode_nb + 1
            end
        end
```

Or may be keep the first solution above, and just have MODE cleaned up at top of file, just after `local MODE`:
```
if not Device:isAndroid() then
    MODE.wifi_status = nil
end
if not Device:hasFrontlight() then
    MODE.frontlight = nil
end
```
Or something like that :)
(I just want to not have `settings.max_possible_modes` and neither that hardcoded `13`)

robert00s:
Wow. Now works :)

poire-z:
Phew :) :+1:

(Enable FreeSerif as the fallback font, and try it in RTL UI: `KO_RTL=yes ./kodev run ...` for the complete arabic experience :)

robert00s:
> (Enable FreeSerif as the fallback font, and try it in RTL UI: KO_RTL=yes ./kodev run ... for the complete arabic experience :)

Done:)

frontend/apps/reader/modules/readerfooter.lua
robert00s: Done in https://github.com/koreader/koreader/pull/5775/commits/2aef4ef5d60eb0a240ed60604dbc80a3f6dfdfeb

robert00s:
>I guess we can approach this in two ways (at least I think so):
  
>1. We can add third line for book and chapter titles (of course, as an option for users that already use all other footer items), otherwise it goes where other options go.
>2. We can place it with reading progress if it is in separate line leaving other line for other footer options, or it goes as before with all other items.

I have another idea. Maybe header (footer at the top of screen) as option. See: https://github.com/koreader/koreader/issues/1933#issuecomment-396679746

AlanSP1:
@robert00s 

> I have another idea. Maybe header (footer at the top of screen) as option. 

From what I know (and maybe I'm completely wrong), this upper part is controlled differently and it already has ability to show book title (but not chapter). Of course, you may hack into it somehow (I don't know anything about programming) as well.

That footer is part of CR engine, from what I gathered, and upper part isn't.

But, whatever could be made, I'd like to have options to set it up and, personally I would prefer everything in only one part, not in upper and lower, but I'd prefer to have everything in lower, i.e. footer part, it's easier to see and comprehend info in my opinion if everything is in same "zone", and not in two "zones".

Of course, upper part could be used by other people who maybe would prefer it.

NiLuJe:
@AlanSP1: To avoid confusing, the *top* status bar (I can't recall how it's actually called in the setting,s since I've never used it) is generated inside CRe, and we have minimal control over it.

Our *own* (bottom) status bar, we control, and is what's being tweaked here ;).

AlanSP1:
Thanks for correcting me. I know that there's something with that top status bar and that it is somehow CR related, but wasn't sure how exactly.

theboomr:
> Well, this opens one more question, at least if you use footer I as I do: You want to have them all visible. :)
> 
> I guess we can approach this in two ways (at least I think so):
> 
> 1. We can add third line for book and chapter titles (of course, as an option for users that already use all other footer items), otherwise it goes where other options go.
> 2. We can place it with reading progress if it is in separate line leaving other line for other footer options, or it goes as before with all other items.
> 
> What do you think about this? And I guess it would be in different PR, if realized. And yes, I really use all options there are. Maybe I can be treated by good doctors for this? :)

Hello, I am extremely happy to see the chapter title has finally been added as an option in the status bar! Thank you for all your hard work! However, I have to echo @AlanSP1 in saying that it would be really really nice to have an option to add a third line where the chapter/book title goes, and/or the ability to organize items onto whichever lines of the footer we want. I am also one of the weird ones who uses almost everything down there, and adding the chapter title just gets far too cramped, at least in portrait orientation.

protist:
Thank you for putting the chapter title in! This is a fantastic idea! However, I also agree with the earlier comments that's it's pretty cramped in the status bar. I use both top and bottom bars, in order to see the book title in the top bar. There's a fair bit of spare room left in the top bar, and putting the chapter name there would be fantastic!

PR #5776, opened 2020-01-17
TextWidget: adds getFittedText()

poire-z:
Needs a base bump for https://github.com/koreader/koreader-base/pull/1031.
Needed for #5775.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5776)

### DIFF
--- a/frontend/ui/widget/textwidget.lua
+++ b/frontend/ui/widget/textwidget.lua
@@ -201,6 +201,57 @@ function TextWidget:_measureWithXText()
     end
 end
 
+-- Returns the substring of text that fits in self.max_width
+-- The substring does not include the ellipsis that could
+-- be added when drawn.
+-- 2nd returned value is nil if no truncation, false when truncated
+-- and no ellipsis would be added, true if truncated and an ellipsis
+-- will be added (on the right or left of string in logical order,
+-- caller knows the side from the provided 'truncate_left').
+function TextWidget:getFittedText()
+    if not self.max_width then
+        return self.text, nil
+    end
+    self:updateSize()
+    if self._is_empty then
+        return "", nil
+    end
+    if not self.use_xtext then
+        if self._text_to_draw == self.text then
+            return self.text, nil
+        end
+        if not self.truncate_with_ellipsis then
+            return self._text_to_draw, false
+        end
+        -- ellipsis is 3 bytes
+        if self.truncate_left then
+            return self._text_to_draw:sub(3), true
+        else
+            return self._text_to_draw:sub(1, -4), true
+        end
+    end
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/textwidget.lua
```diff
@@ -201,6 +201,49 @@ function TextWidget:_measureWithXText()
...
+        return self.text, nil
+    end
+    local with_ellipsis = false
+    local start_idx, end_idx = self._shape_start, self._shape_end
+    if self._shape_idx_to_substitute_with_ellipsis then
+        with_ellipsis = true
+        if self.truncate_left then
+            start_idx = start_idx + 1
+        else
+            end_idx = end_idx -1
```
Frenzie: -1 or - 1?

frontend/ui/widget/textwidget.lua
```diff
@@ -201,6 +201,49 @@ function TextWidget:_measureWithXText()
     end
 end
 
+function TextWidget:getFittedText()
+    -- Returns the substring of text that fits in self.max_width, without
```
Frenzie: Looks like LDoc-type documentation to me, for in http://koreader.rocks/doc/modules/ui.widget.textwidget.html

Like one of these molds:
https://github.com/koreader/koreader/blob/b111ccc2b803fc78425d9791fcc8bb7a9a704e6e/frontend/util.lua#L27-L47
https://github.com/koreader/koreader/blob/b111ccc2b803fc78425d9791fcc8bb7a9a704e6e/frontend/util.lua#L948-L953

poire-z: Yeah, but none of the other methods have a ldoc doc. Feels awkward to document such a minor one and not have the other essential ones not documented :) But I don't feel like spending the time needed to document the other ones...

Frenzie: Alright, just keep the comment as is then, but I think it'd be clearer above the function. (Also you could change the first line to three dashes to convert it into an LDoc comment in that case, resulting in a smaller diff.)

poire-z:
Note about:
```
-- The substring does not include the ellipsis that could
-- be added when drawn.
```
So, this returns a smaller text string than the one that could fit, as it removes the width needed for the ellipssis. That's fine for #5775 use case.
If one wants the text that fully fit in max_width, without removing the width needed for the ellipsis, one just have to provide `truncate_with_ellipsis = false`, and he will get the expected text.

poire-z:
Just fixed the `not use_xtext` case, after some last minute test.
I'll merge it this evening, even if not validated where it will be used - I'll fix it if needed.

PR #5783, opened 2020-01-22
Make Inverted Portrait a first-class citizen

### LINKED ISSUE #5772: Don't remeber screen orientation
* KOReader version: 2019.12
* Device: Kobo Forma

#### Issue
When reading from the device in a specific orientation, let's say portrait with hardware buttons on the left, then switch to main file browser, rotate the screen in portrait with harware buttons to the right and open the previous book again, the book automatically rotates the screen back 180 degrees as it remembered the orientation from when it was closed.
This is a pain to rotate back. You have to rotate the device again 180 degrees, wait, and then another 180 degrees to register.
My suggestion here would be NOT to remember the previous rotation of the book and always open the book in the current screen orientation.

#### Steps to reproduce
Open book, rotate screen. Close book, rotate screen. Open book again, screen auto-rotates as it was previously rotated instead of opening in the current screen orientation.

NiLuJe:
Mainly by ensuring we actually *always* remember it properly in a book sidecar, and that we support it in ScreenSaver mode, as well as in the FM.

This basically makes the whole "lefty" workflow consistent, without any spurious rotations back to Portrait.

Landscape is another kettle of fish, because the FM doesn't support it (and neither do image ScreenSaver modes).

Fix #5772

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5783)

### DIFF
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -205,7 +205,7 @@ function Device:onPowerEvent(ev)
                     network_manager:scheduleConnectivityCheck()
                 end
                 self:resume()
-                -- restore to previous rotation mode, if need be.
+                -- Restore to previous rotation mode, if need be.
                 if self.orig_rotation_mode then
                     self.screen:setRotationMode(self.orig_rotation_mode)
                 end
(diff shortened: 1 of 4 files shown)


NiLuJe:
This got rid of a few duplicate/bogus setScreenMode/setRotationMode calls, too.

NiLuJe:
Not quite sure what's going on with the CI, @Frenzie ?

(I'm on my laptop right now, so I won't have a chance to take a look at the re-run's result until later tonight/tomorrow).

Frenzie:
Well, I just see tests passed. ;-)

frontend/device/generic/device.lua
```diff
@@ -228,11 +228,17 @@ function Device:onPowerEvent(ev)
...
         local UIManager = require("ui/uimanager")
         logger.dbg("Suspending...")
-        -- Mostly always suspend in portrait mode...
-        -- ... except when we just show an InfoMessage, it plays badly with landscape mode (c.f., #4098)
+        -- Mostly always suspend in Portrait/Inverted Portrait mode...
+        -- ... except when we just show an InfoMessage, it plays badly with Landscape mode (c.f., #4098)
         if G_reader_settings:readSetting("screensaver_type") ~= "message" then
             self.orig_rotation_mode = self.screen:getRotationMode()
-            self.screen:setRotationMode(0)
+            -- Leave Portrait & Inverted Portrait alone, that works just fine.
```
NiLuJe: Note to self: possibly could have gone with a getScreenMode == "portrait" here instead.

But I'm not a fan of getScreenMode anyway ;p (which basically only does a w > h check).

Issue #5777, opened 2020-01-17
Bookmark icon size is to small


oldman63:
* KOReader version: 2019.12
* Device: Kindle Paperwhite 1

If the page is bookmarked, the icon in up-right corner is very small (dogear.png 20x20 px).

How do increase the size of this icon?
Change for larger icon, not working. :-(

Please, put this icon size setting to menu.

poire-z:
It should have a max of 1/32 of screen width - and can additionally be reduced depending on the margins you use (in cre/EPUB documents) - so as to not be drawn OVER the book text.
That's a design choice, that fixed a few issues. Details in #4081.

If you have small margins, try increasing them, and see if you get a bigger size.
The code if you feel like experimenting is:
https://github.com/koreader/koreader/blob/b111ccc2b803fc78425d9791fcc8bb7a9a704e6e/frontend/apps/reader/modules/readerdogear.lua#L11-L18
https://github.com/koreader/koreader/blob/b111ccc2b803fc78425d9791fcc8bb7a9a704e6e/frontend/apps/reader/modules/readerdogear.lua#L62-L75

I don't think it's worth having its size in the menu - but we may add a hidden/manual setting to either change the 1/32 or skip the cre margins restriction - depending on which of them bothers you.

But it has been like this for 18 months, any reason you're reacting just now (assuming you're a long time user as you're involved in translations) ? May be you have unusually set small margins? :)

pazos:
I'm not sure why @oldman63 wants a bigger icon but if is just for figuring it which pages have a bookmark on a quick look I would recommend the "two finger right to left swipe", which will show you all the markers in the entire book.

oldman63:
Thank you, for "Two finger right to left swype" tip!

But small bug found...

Tap upper-right corne -> bookmarked this page
Two finger right to left swype -> bookmarks
Long press on bookmark, Delete bookmark
Close Bookmarks page with [X] on upper-right
The bookmark icon always visible on the up-right corner!!! Don't removed.
After page left, Page right, the icon is disappeared.

poire-z:
It's also available via multiswipe gestures, north-west by default - see picture in https://github.com/koreader/koreader/releases/tag/v2019.03.

Any answer to my above question about why your need for bigger icon? :)

Frenzie:
There's also a wiki page: https://github.com/koreader/koreader/wiki/multiswipes

(Not entirely sure how you can reach it without searching for the page though.)

[...]

PR #5786, opened 2020-01-23
Render metadata text with book language

poire-z:
Gives the book language to Text*Widget (and XText) when drawing title, authors and other metadata text (mentionned at https://github.com/koreader/koreader/pull/5696#issuecomment-570739039).
Might be needed to properly display a Japanese book title when the UI language is Chinese when the glyphs are different.
The following 2 books have the same UTF8 text in their metadata title and authors, but a different language - our CJK font would default to the chinese glyphs (but it does contain the japanese glyphs, and Harfbuzz will use them when we tell it to):
<kbd>image</kbd>
<kbd>image</kbd>
<kbd>image</kbd>

Also:
- [Fix] Update dogear when back from bookmarks list - Issue described at https://github.com/koreader/koreader/issues/5777#issuecomment-575703642.
- Footnote popups: drop justification if RTL content
MuPDF (even latest version, looking at the code) does not handle "text-align: justify" correctly on
RTL text: the last line is left-aligned, but it should be right-aligned. Not using justify makes it correctly right align the whole text.
We still need to tweak our MuPDF patch to have it use FreeSerif instead of FreeSans to be able to display arabic or hebrew in our footnote popups.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5786)

### DIFF
--- a/frontend/apps/filemanager/filemanagerbookinfo.lua
+++ b/frontend/apps/filemanager/filemanagerbookinfo.lua
@@ -210,10 +210,18 @@ function BookInfo:show(file, book_props)
     end
     table.insert(kv_pairs, { _("Cover image:"), _("Tap to display"), callback=viewCoverImage })
 
+    -- Get a chance to have title, authors... rendered with alternate
+    -- glyphs for the book language (e.g. japanese book in chinese UI)
+    local values_lang = nil
+    if book_props.language and book_props.language ~= "" then
+        values_lang = book_props.language
+    end
+
     local widget = KeyValuePage:new{
         title = _("Book information"),
         value_overflow_align = "right",
         kv_pairs = kv_pairs,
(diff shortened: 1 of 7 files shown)


frontend/apps/filemanager/filemanagerbookinfo.lua
```diff
@@ -210,10 +210,18 @@ function BookInfo:show(file, book_props)
     end
     table.insert(kv_pairs, { _("Cover image:"), _("Tap to display"), callback=viewCoverImage })
 
+    -- Get a chance to have title, authors... rendered with alternate
+    -- glyphs for the book language (e.g. japanese book in chinese UI)
+    local values_lang = nil
```
Frenzie: Obligatory = nil isn't necessary comment. :-P

Issue #5778, opened 2020-01-18
Status bar overflow due to large progress bar


Biep:
* KOReader version: v2019.12-38-g7eee758_2020-01-14
* Device: Kobo Aura One

#### Issue
The progress bar no longer has the option to shrink to the available space. On my Kobo Aura One, in portrait mode, when showing everything (in default order), memory usage drops off the bar at the far end. If the bar would shrink ever so little, it would all fit, as it used to.
#### Steps to reproduce
Read a document in portrait mode, have the status bar visible and showing everything.

Probably related to #5757

Biep:
And a related request: it would be great if the bar could:
- Have the setting "fill remaining space"
- Be sortable like the other elements.  I'd really like to have one kind of information to the left, and the other to the right of the bar.

PR #5788, opened 2020-01-23
Extend minimal progress bar width to 5%

robert00s:
Close: #5778 
Minimal progress bar in footer can be now from 5% to 50%. Before 20%-50%.
Settings: `Status bar -> progress bar -> Minimal width`
Regression from #5757

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5788)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -1335,7 +1335,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                     local items = SpinWidget:new{
                         width = Screen:getWidth() * 0.6,
                         value = self.settings.progress_bar_min_width_pct,
-                        value_min = 20,
+                        value_min = 5,
                         value_step = 5,
                         value_hold_step = 20,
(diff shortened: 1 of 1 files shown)


Biep:
Thanks!
Tbough I'd still prefer a "Take remaining space"option (which would make it dynamic: if "off" takes more space than "on" besides the light symbol, the bar would shrink a bit to accommodate the "off".
And being able to sort it among the other elements would be wonderful too.

poire-z:
This PR allows setting the minimal width. The progressbar will continue taking the remaining space to make the footer full width. "Take remaining space" is the default.
Before, as you noticed, if you had many stuff, it could only shrink to 20%: now, you can tell it shrink it to 5%. So, it's not "auto-accomodate": it's you who decide how it should accomodate (and tomorrow, you'll have options to show chapter and book titles, so a few more items that can make things crowded, and so a few more min/max widths to decide which can take more than the others).

Biep:
I discovered this.  Thanks, for "take available space" is a great feature!
And thanks for the new stuff too!

PR #5789, opened 2020-01-23
Reader: strict ordering of tap & hold gestures

poire-z:
The lack of proper "overrides" could make their order/precedence/priority non-deterministic.
Issued described at https://github.com/koreader/koreader/pull/4878#issuecomment-577371343 - but this PR does not exactly set them as I proposed there :)

I had quite a hard time following the flow, having it not working at all with circular override, having to decide where to break it :) and hopefully, making it consistent.
So, not describing right now what's the ordering/priorities: I put in some comments, please have a look and see if the way you read it and understand it make sense :)

(I let in some `print()` because it's not the first time I had to use that - and so probably not the last time :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5789)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -484,15 +484,17 @@ function ReaderFooter:setupTouchZones()
                 "tap_backward",
                 "readerconfigmenu_tap",
             },
+            -- (Low priority: tap on existing highlights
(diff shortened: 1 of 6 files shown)


NiLuJe:
Not familiar with the code, but it seems sound and simple enbough, and I definitely agree with the rationale ;).

PR #5795, opened 2020-01-27
Add turkish q keyboard layout

mustafa-001:
Pretty much self explanatory. I tried to imitate layouts used by mobile phones. 5th and 6th levels of added keys feature Kurdish and Azeri characters. Also added a commented part to make top row 10 key rather than standart 11 key row.

Since keyboard layouts gets their name from language locale names, it shows up as "Türkçe" which commonly refers to F layout. Just writing here to be safe in case anybody complains in the future.
Ekran Görüntüsü - 2020-01-28 01-32-50

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5795)


Frenzie:
Thanks!

> Since keyboard layouts gets their name from language locale names

Yeah, a bit of a hack.

PR #5800, opened 2020-01-30
Bump crengine, "Enforce steady line heights" style tweak, Wikipedia EPUB tweaks

### LINKED ISSUE #5766: Koreader getting slower and slower as I add highlights

poire-z:
See https://github.com/koreader/crengine/pull/326 for details and screenshots (including some about the added style tweak purpose, and galleries in wikipedia EPUBs).

The wikipedia gallery CSS solution is a bit hacky :) but I had a hard time getting them as I wanted, and this trick seems quite fine, and survives render mode change without the need for reloading the book.

(A bit saddened after I had cleaned up the HTML tweaking code and removed these `:gsub()`, as they were kinda my syntax reference when I needed to write a complex one :) I'll have now to dig them out from git history.)

Also includes:
MuPDF: switch fallback font from FreeSans to FreeSerif.  https://github.com/koreader/koreader-base/pull/1033
Bump LodePNG to 2020/01/20. https://github.com/koreader/koreader-base/pull/1032
Bump jpeg-turbo to 2.0.4.

Also enable cre call cache (#5773) by default. Closes #5766.

Style tweak wording:
<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5800)

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -269,6 +269,16 @@ DocFragment {
             },
         },
         {
+            id = "hyphenate_all_auto";
+            title = _("Allow hyphenation on all text"),
+            description = _("Allow hyphenation on all text (except headings), in case the publisher has disabled it."),
+            css = [[
+* { hyphens: auto !important; }
+h1, h2, h3, h4, h5, h6 { hyphens: none !important; }
+            ]],
+            separator = true,
+        },
+        {
             id = "sub_sup_smaller";
             title = _("Smaller sub- and superscript"),
             description = _("Prevent sub- and superscript from affecting line-height."),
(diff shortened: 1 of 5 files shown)


frontend/ui/data/css_tweaks.lua
```diff
@@ -268,6 +268,16 @@ DocFragment {
                 ]],
             },
         },
+        {
+            id = "hyphenate_all_auto";
+            title = _("Allow hyphenation on all text"),
+            description = _("Allow hyphenation to happen on all text (except headings), in case the publisher has disabled it."),
```
Frenzie: I'd probably say something like "occur" or "be applied" instead of "happen," but I'd be more inclined to just leave it out. (I just realized this isn't new, but either way. ;-) )
```suggestion
            description = _("Allow hyphenation on all text (except headings), in case the publisher has disabled it."),
```

PR #5801, opened 2020-01-30
[android]: nuke wifiManager

### LINKED ISSUE #5787: When you're not so close to WiFi ..
* KOReader version: 2019-12-27
* Device: Android (inkbook)

#### Issue

When you're not sitting next to wifi,  using online koreader functions can be very painfull:

https://youtu.be/mV6LyCLQtm8

often you have to click many times 
menu -> wallabag/newsdownloadr/send2ebook -> "retrievie articles from server" -> "turn on wifi" 
(and again, on example video this path was repeated around 7-8 , each 4 clicks, as above, gives 32 clicks) unit it really captures internet, and starts processing. That's horrible!

Year or 2 back I remember I did at least for newsDownloader or send2ebook callback function for enabling wifi that waited until it's connected (or terminates after 30sek) and then automaticly starts processing eg wallabag articles. 

But it seems it's not there anymore or is broken ? 
how about restoring that function and implementing it every where were you need to use "enable wifi" function ?

#### Steps to reproduce
described in issue

pazos:
Fixes #5787

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5801)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -230,22 +230,21 @@ end
 
 function Device:initNetworkManager(NetworkMgr)
     function NetworkMgr:turnOnWifi(complete_callback)
-        android.setWifiEnabled(true)
-        if complete_callback then
-            local UIManager = require("ui/uimanager")
-            UIManager:scheduleIn(1, complete_callback)
-        end
+        android.openWifiSettings()
+    end
+    function NetworkMgr:turnOffWifi(complete_callback)
+        android.openWifiSettings()
     end
 
-    function NetworkMgr:turnOffWifi(complete_callback)
-        android.setWifiEnabled(false)
-        if complete_callback then
-            local UIManager = require("ui/uimanager")
-            UIManager:scheduleIn(1, complete_callback)
-        end
+    function NetworkMgr:openSettings()
+        android.openWifiSettings()
     end
-    NetworkMgr.isWifiOn = function()
-        return android.isWifiEnabled()
+
+    function NetworkMgr:isWifiOn()
+        local ok = android.getNetworkInfo()
+        ok = tonumber(ok)
+        if not ok then return false end
+        return ok == 1
     end
 end
 
@@ -254,12 +253,23 @@ function Device:performHapticFeedback(type)
 end
 
 function Device:retrieveNetworkInfo()
-    local ssid, ip, gw = android.getNetworkInfo()
-    if ip == "0" or gw == "0" then
+    local ok, type = android.getNetworkInfo()
+    ok, type = tonumber(ok), tonumber(type)
+    if not ok or not type or type == C.ANETWORK_NONE then
         return _("Not connected")
     else
-        local BD = require("ui/bidi")
-        return T(_("Connected to %1\n IP address: %2\n gateway: %3"), BD.wrap(ssid), BD.ltr(ip), BD.ltr(gw))
+        if type == C.ANETWORK_WIFI then
+            return _("Connected to Wi-Fi")
+        elseif type == C.ANETWORK_MOBILE then
+            return _("Connected to mobile data network")
+        elseif type == C.ANETWORK_ETHERNET then
+            return _("Connected to Ethernet")
+        elseif type == C.ANETWORK_BLUETOOTH then
+            return _("Connected to Bluetooth")
+        elseif type == C.ANETWORK_VPN then
+            return _("Connected to VPN")
+        end
+        return _("Unknown connection")
     end
 end
 
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -153,6 +153,16 @@ function NetworkMgr:isOnline()
     return socket.dns.toip("dns.msftncsi.com") ~= nil
 end
 
+function NetworkMgr:isNetworkInfoAvailable()
+    if Device:isAndroid() then
+        -- always available
+        return true
+    else
+        --- @todo also show network info when device is authenticated to router but offline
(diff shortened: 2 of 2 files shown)


frontend/device/android/device.lua
```diff
@@ -254,12 +250,19 @@ function Device:performHapticFeedback(type)
...
-    if ip == "0" or gw == "0" then
+    local ok, type = android.getNetworkInfo()
+    if ok == "0" then
         return _("Not connected")
     else
-        local BD = require("ui/bidi")
-        return T(_("Connected to %1\n IP address: %2\n gateway: %3"), BD.wrap(ssid), BD.ltr(ip), BD.ltr(gw))
+        local status = nil
+        if type == "1" then
+            status = _("connected to Wi-Fi")
```
Frenzie: Unless you want to change all the strings, it seems better to keep them consistent? :-)
```suggestion
            status = _("Connected to Wi-Fi")
```

frontend/device/android/device.lua
```diff
@@ -254,12 +250,19 @@ function Device:performHapticFeedback(type)
...
+        elseif type == "2" then
+            status = _("connected to mobile data")
```
Frenzie: Mobile data seems a bit odd? But we/I should double check what's used in Android and such too.
```suggestion
            status = _("Connected to mobile data network")
```

frontend/device/android/device.lua
```diff
@@ -254,12 +250,19 @@ function Device:performHapticFeedback(type)
...
+        elseif type == "3" then
+            status = _("connected to ethernet")
+        end
+        return status or _("connecting")
```
Frenzie: No point adding another string to translate.
```suggestion
        return status or _("Connecting…")
```

frontend/ui/network/manager.lua
```diff
@@ -165,6 +165,7 @@ function NetworkMgr:setHTTPProxy(proxy)
 end
 
 function NetworkMgr:getWifiMenuTable()
+    if not Device:isAndroid() then
```
Frenzie: Is this weird indentation here intentional? (If so I'd like to hear the rationale before saying it doesn't align with the coding style. :-P)

frontend/device/android/device.lua
pazos: I copied it from https://developer.android.com/training/monitoring-device-state/connectivity-status-type#DetermineType.

I also forgot WiMAX as a valid network type :p.

frontend/ui/network/manager.lua
pazos: intentional: lazyness porting from other branch. Will fix!

pazos: Fixed!

frontend/device/android/device.lua
```diff
@@ -230,22 +230,21 @@ end
 
 function Device:initNetworkManager(NetworkMgr)
     function NetworkMgr:turnOnWifi(complete_callback)
```
Frenzie: It seems that if the `complete_callback` thing is useless here you could just as well write `NetworkMgr.turnOnWifi = android.openWifiSettings`? But whichever, writing it this way is probably friendlier to potential future changes.

frontend/device/android/device.lua
```diff
@@ -254,12 +253,23 @@ function Device:performHapticFeedback(type)
...
-        local BD = require("ui/bidi")
-        return T(_("Connected to %1\n IP address: %2\n gateway: %3"), BD.wrap(ssid), BD.ltr(ip), BD.ltr(gw))
+        if type == C.ANETWORK_WIFI then
+            return _("Connected to Wi-Fi")
+        elseif type == C.ANETWORK_MOBILE then
+            return _("Connected to mobile data network")
+        elseif type == C.ANETWORK_ETHERNET then
+            return _("Connected to ethernet")
+        elseif type == C.ANETWORK_BLUETOOTH then
+            return _("Connected to bluetooth")
```
Frenzie: As you know I'm in favor of wifi (and therefore possibly bluetooth by similar reasoning) but for now that should be Bluetooth in English.

```suggestion
            return _("Connected to Bluetooth")
```

frontend/device/android/device.lua
pazos: yup, ignoring is a valid way of treating a parameter :smile:

frontend/device/android/device.lua
pazos: The same for Ethernet?

Frenzie:
https://github.com/koreader/koreader/pull/5801#discussion_r376160439
> The same for Ethernet?

Good point, if wifi is spelled Wi-Fi then ethernet should be spelled Ethernet.

On a related note, I just noticed this Microsoft styleguide: https://docs.microsoft.com/en-us/style-guide (naturally it's very trademarkophilic)

frontend/device/android/device.lua
```diff
@@ -254,12 +253,23 @@ function Device:performHapticFeedback(type)
...
         return _("Not connected")
     else
-        local BD = require("ui/bidi")
-        return T(_("Connected to %1\n IP address: %2\n gateway: %3"), BD.wrap(ssid), BD.ltr(ip), BD.ltr(gw))
+        if type == C.ANETWORK_WIFI then
+            return _("Connected to Wi-Fi")
+        elseif type == C.ANETWORK_MOBILE then
+            return _("Connected to mobile data network")
+        elseif type == C.ANETWORK_ETHERNET then
+            return _("Connected to ethernet")
```
Frenzie:
```suggestion
            return _("Connected to Ethernet")
```

PR #5804, opened 2020-01-31
Allow running shell scripts from the FileManager

NiLuJe:
* Available via the onHold menu, where I stuck it inside the extra separator below cut/delete/rename. Only on shell scripts (.sh).

* That implies making CRe recognize those (which it will, ideally, read as plain text).

* Also available in the Favorites onHold menu.

* Currently runs in the main process, so, blocking, which feels like the right call to me ;). It's bracketed between InfoMessages to make that clear.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5804)


NiLuJe:
Currently looks like this:

ko_shell

Any suggestions, @poire-z?

NiLuJe:
@pazos: Should I disable this on Android?

NiLuJe:
Also, @poire-z: did I use the BiDi stuff right?

NiLuJe:
FWIW, the intended use case is small, simple stuff (say, toggling USBNet, or running a custom sync script (f.g., rclone)).

frontend/apps/filemanager/filemanager.lua
```diff
@@ -297,8 +298,42 @@ function FileManager:init()
...
+        if lfs.attributes(file, "mode") == "file" and string.lower(FrontUtil.getFileNameSuffix(file)) == "sh" then
+            -- NOTE: We populate the empty separator, in order not to mess with the button reordering code in CoverMenu
+            table.insert(buttons[3],
+                {
+                    text = _("Execute shell script"),
+                    enabled = true,
+                    callback = function()
+                        UIManager:close(self.file_dialog)
+                        local script_is_running_msg = InfoMessage:new{
+                                text = T(_("Running shell script %1 ..."), BD.filepath(util.basename(file))),
```
poire-z: If `basename()` is stripping the directory path, and it becomes just a filename, it would then be `BD.filename(...)`

frontend/apps/filemanager/filemanager.lua
```diff
@@ -297,8 +298,42 @@ function FileManager:init()
                 }
             },
             -- a little hack to get visual functionality grouping
-            {},
+            {
+            },
         }
+
+        if lfs.attributes(file, "mode") == "file" and string.lower(FrontUtil.getFileNameSuffix(file)) == "sh" then
+            -- NOTE: We populate the empty separator, in order not to mess with the button reordering code in CoverMenu
```
poire-z: Nice trick :)

frontend/apps/filemanager/filemanager.lua
```diff
@@ -297,8 +298,42 @@ function FileManager:init()
...
+                                text = T(_("Running shell script %1 ..."), BD.filepath(util.basename(file))),
+                        }
+                        UIManager:show(script_is_running_msg)
+                        UIManager:scheduleIn(0.5, function()
+                            local rv = os.execute(util.realpath(file))
+                            UIManager:close(script_is_running_msg)
+                            if rv == 0 then
+                                UIManager:show(InfoMessage:new{
+                                    text = _("It exited successfully."),
+                                })
```
poire-z: You could also use `Trapper:dismissablePopen()` (like readerdictionary.lua does for sdcv) to get the script output, and allow interrupting it.

frontend/apps/filemanager/filemanager.lua
```diff
@@ -41,6 +41,7 @@ local filemanagerutil = require("apps/filemanager/filemanagerutil")
 local lfs = require("libs/libkoreader-lfs")
 local logger = require("logger")
 local util = require("ffi/util")
+local FrontUtil = require("util")
```
poire-z: First time we'd see that name "FrontUtil" :) I'd prefer renaming the other for consistency, BaseUtil is used and ok (I prefer ffiutil) - but that would make many unrelated changes (but well, you did it for util.template > T :)

Frenzie:
Boy, that's really getting crowded.

frontend/apps/filemanager/filemanager.lua
```diff
@@ -297,8 +298,42 @@ function FileManager:init()
                 }
             },
             -- a little hack to get visual functionality grouping
-            {},
+            {
+            },
         }
+
+        if lfs.attributes(file, "mode") == "file" and string.lower(FrontUtil.getFileNameSuffix(file)) == "sh" then
```
Frenzie: More of a general remark, but many shellscripts are just "executables" where whether it's a shell script, a python script or even plain old assembly is just an implementation detail.

frontend/apps/filemanager/filemanager.lua
NiLuJe: I don't actually care about the *output* (it's sent to the log, which is just fine by me), only the return code ;).

frontend/apps/filemanager/filemanager.lua
NiLuJe: Yeah, it felt awkward to me, too, will do ;).

frontend/apps/filemanager/filemanager.lua
```diff
@@ -297,8 +298,42 @@ function FileManager:init()
...
+                        local script_is_running_msg = InfoMessage:new{
+                                text = T(_("Running shell script %1 ..."), BD.filepath(util.basename(file))),
+                        }
+                        UIManager:show(script_is_running_msg)
+                        UIManager:scheduleIn(0.5, function()
+                            local rv = os.execute(util.realpath(file))
+                            UIManager:close(script_is_running_msg)
+                            if rv == 0 then
+                                UIManager:show(InfoMessage:new{
+                                    text = _("It exited successfully."),
```
Frenzie: Admittedly this whole thing is rather technical, but maybe just "(The) Script (was) successful" or something?

frontend/apps/filemanager/filemanager.lua
```diff
@@ -297,8 +298,42 @@ function FileManager:init()
...
+                                })
+                            else
+                                --- @note: Lua 5.1 returns the raw return value from the os's system call. Counteract this madness.
+                                UIManager:show(InfoMessage:new{
+                                    text = T(_("It returned a non-zero status code: %1!"), bit.rshift(rv, 8)),
```
Frenzie:
```suggestion
                                    text = T(_("The script returned a non-zero status code: %1!"), bit.rshift(rv, 8)),
```
Better to be clear. ;-)

frontend/apps/filemanager/filemanager.lua
NiLuJe: Yeah, I'd ideally have just chucked a whole load of extensions in DocumentRegistry, and done the check based on the exec bit, but I can't do that because FAT32 :/.

I'll probably add .py to the list, though ;).

Frenzie: But the only one we know for sure is executable is a Lua script :-P

frontend/apps/filemanager/filemanager.lua
NiLuJe: I, err, may have lost a few hairs before realizing that I was mangling the nesting in my first attempts >_<".

pazos:
Yeeeah! Running shell scripts on root account! What can go wrong? :smiley: :hear_no_evil:

pazos:
> Should I disable this on Android?

Yes for this PR. The only problem I see is the ANR, that can be avoided with a dialog on top of the activity (like the animation on 1st launch).  

If the script is guarantied to return or is cancellable we're fine. I guess the same happens on other platforms. What happens if the script has a while loop and blocks forever?

frontend/apps/filemanager/filemanager.lua
NiLuJe: I'm slightly wary of adding lua to the list because of us being full of lua...

NiLuJe:
> 
> If the script is guarantied to return or is cancellable we're fine. I guess the same happens on other platforms. What happens if the script has a while loop and blocks forever?

Then it blocks forever ^^.

The other possibility would be chucking it inside a `BaseUtil.runInSubProcess` which seems a bit iffy for that use-case?

Otherwise, there's what @poire-z  proposed above (`Trapper:dismissablePopen`) which I'm not terribly familiar with.

NiLuJe:
Okay, reading `Trapper:dismissablePopen`, that only works for stuff that outputs stuff to stdout, i.e., can no longer be cancelled as soon as there's output to be read, which wouldn't work here.

NiLuJe:
That leaves `Trapper:dismissableRunInSubprocess`, which would probably work but feels a tad overkill here.

I'm obviously biased by being okay as having this blocking the UI, though ;).

frontend/apps/filemanager/filemanager.lua
NiLuJe: But the interruptible aspect might be worth looking into, see discussion below ;).

pazos:
> Then it blocks forever ^^.

If this is intended then we can make the same in android, replacing InfoMessages by calls to dialog.show and dialog.dismiss. As in other platforms it will block forever if the script does not return. At least it won't be killed by the activity manager.

I'm not sure if is worth the hassle, though

poire-z:
I'm fine whichever way.
(I thought one could build fancy scripts with wget|grep|... and display the output in a TextViewer... but well, whenever I need such fancy stuff, *i* can patch koreader and have it done in superfancy Lua via some menu item :)

> reading Trapper:dismissablePopen, that only works for stuff that outputs stuff to stdout: it can no longer be cancelled as soon as there's output to be read, which wouldn't work here.

You could solve both cases by wrapping the script in something like:
```
output=`./original_script.sh`; echo "$output" ; echo
```

NiLuJe:
Yeah, I'm kind of on the "let people shoot their own feet off if they want to" camp, but good to know all that :).

AlanSP1:
Added to [wiki](https://github.com/koreader/koreader/wiki/Shell-scripts)

NiLuJe:
Are we all good with this approach for a first iteration?

Frenzie:
Besides the UI space it takes up, sure. :-P

NiLuJe:
It only takes space on an actual script, though ;p.

Frenzie:
Fair enough, forgot about that, all good then :+1:

poire-z:
(minor remark: it will show `Execute shell script` even on  non-shell .py scripts :)

NiLuJe:
@poire-z: Yeah, but switching to "Execute script" felt wonky to me for some reason (and making the string dynamic depending on the extension seems a wee bit overkill ^^).

NiLuJe:
Actually, making that label templated is trivial, so, wheee! \o/

frontend/apps/filemanager/filemanager.lua
```diff
@@ -306,11 +306,13 @@ function FileManager:init()
             -- NOTE: We populate the empty separator, in order not to mess with the button reordering code in CoverMenu
             table.insert(buttons[3],
                 {
+                    -- @translators This is the script's programming language (f.g., shell or python)
```
Frenzie:
```suggestion
                    -- @translators This is the script's programming language (e.g., shell or python)
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -306,11 +306,13 @@ function FileManager:init()
...
                     text = T(_("Execute %1 script"), util.getScriptType(file)),
                     enabled = true,
                     callback = function()
                         UIManager:close(self.file_dialog)
                         local script_is_running_msg = InfoMessage:new{
+                                -- @translators %1 is the script's programming language (f.g., shell or python), %2 is the filename
```
Frenzie:
```suggestion
                                -- @translators %1 is the script's programming language (e.g., shell or python), %2 is the filename
```

NiLuJe:
You mean to say I've been using it wrong for all these years? :D.

Latin, man.

ptrm:
Wow, I've been just looking for an alternative to not check my dropbox book storage via udev event! (as rclone sometimes downloads a dozen books even after ntp time sync on kobo). 

Thanks 😁

pazos:
Just spent half and hour implementing a longTaskDialog to be used while running scripts. After that I found that internal/external storage are mounted noexec :man_facepalming:  So nothing to do in android.

rigogsilva:
I saw python in your comments. Does Koreader support run python scripts? (Android 9.0)

pazos:
> I saw python in your comments. Does Koreader support run python scripts? (Android 9.0)

Python isn't bundled in any platform, but is usually available on some of them. Android is not the case unless you have a **very customized** device.

Bundling python is out of the scope as the disk space required by the interpreter and the "batteries included" is greater than the entire program, including all fonts for all languages and all translations :)

What could be possible for android is to run python or any other interpreted code using intents. Termux is a pretty powerful environment but, AFAIK, it doesn' allow 3rd party apps to execute things directly, because security reasons.

rigogsilva:
Gotcha. I was surprised when I saw that and thought I’d check. Thanks 🙏

Issue #5806, opened 2020-02-01
Wrong translations 


elvvis:
* KOReader version: v2020.01
* Device: Tolino Epos 2

#### Issue

The status bar has two new options: „Book title“ and „Current chapter“.
In the german translation, these are translated with „Buchstatus“ (book status) and „Nächstes Kapitel“ (next chapter), that‘s wrong.
When I look at the translation at Weblate, the strings are correct.
The same problem exists in another language: in French, these two options are also incorrectly translated: „État du livre“ and „Chapitre suivant“.

The assignment of the translated strings is probably incorrect.

Frenzie:
This is in the translated string,
```
#: frontend/apps/reader/modules/readerfooter.lua:667
#, fuzzy
msgid "Book title"
msgstr "Buchstatus"
```

That was fixed a few days after the release, hence why it's correct on Weblate: https://github.com/koreader/koreader-translations/commit/a6f728f01f5c55a7f79049db79929077c6814058

Frenzie:
But perhaps I should prevent fuzzy strings from being displayed… :thinking:

PR #5807, opened 2020-02-01
[i18n] GetText: ignore fuzzy strings

Frenzie:
Strings are prefilled by msgmerge with closely matching ones to reduce the amount of labor required by the translator. Often the string requires only minor adjustment, making the process faster, and when the string isn't a good match it's no big deal. However, these so-called fuzzy strings shouldn't be treated as if they were actually translated.

Fixes the effect seen in <https://github.com/koreader/koreader/issues/5806>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5807)


Frenzie:
### Before/after

PR #5808, opened 2020-02-02
Add some basic info on using ffi-cdecl

tcrs:
I found this was a big stumbling block for me when porting to remarkable. I couldn't see any documentation on how these FFI Lua files were generated, so I added some where I expected to find in. Maybe there is some somewhere else which I missed.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5808)

### DIFF
--- a/doc/Porting.md
+++ b/doc/Porting.md
@@ -25,6 +25,19 @@ Otherwise, when there'd be signficiant gain in doing so (i.e., when displaying m
 
 The actual framebuffer content is then refreshed (i.e., displayed) via device-specific ioctls, making the best effort in using device-specific capabilities, whether that be [optimized waveform modes](https://github.com/koreader/koreader-base/blob/a1fc4e43b7cce7a76b13224e145f9bada343d8ea/ffi/framebuffer_mxcfb.lua#L643-L655), hardware dithering or [hardware inversion](https://github.com/koreader/koreader-base/blob/a1fc4e43b7cce7a76b13224e145f9bada343d8ea/ffi/framebuffer_mxcfb.lua#L253-L256).
 
+The platform-specific Lua FFI modules for using the mxcfb driver are generated from the [kernel header files](https://github.com/koreader/koreader-base/ffi-cdecl/include) using [ffi-cdecl](https://github.com/koreader/ffi-cdecl). `ffi-cdecl` operates on [C files](https://github.com/koreader/koreader-base/ffi-cdecl/) which specify what to export.
+
+To use `ffi-cdecl` first you need to build the gcc plugin for the platform toolchain, e.g.:
+```
+cd ffi-cdecl
+make clean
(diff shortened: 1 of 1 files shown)


Frenzie:
Most of the FFI isn't platform-specific or related to porting, and generating them is primarily for consistency and reproducibility. I.e., someone else doesn't have to figure out what's up; they can just run ffi-cdecl safe in the knowledge that it'll be good. That way you gain long-term maintainability.

For prototyping I typically consider ffi-cdecl an obstacle rather than an aid. I'm looking at the headers to figure out what's what and I just stick them in an ffi.cdef statement right then and there. Maybe I just copy the whole header file so I can go ahead and do things.

That's just some rambling while I'm trying to figure out what I think of adding this section to Porting. ;-)

NiLuJe:
I find a diff of the "generated" headers a great way to catch API/ABI updates when updating a dependency ;).

NiLuJe:
As far as the content of this PR is concerned, it perfectly sums it up ;).

(I'd probably do away with the explicit PATH shenanigans and just mention that the target TC should obviously be in PATH, though).

----

On a slightly sneakier level, the generated files *are* platform specific to a certain extent: the data types are resolved by the compiler being used. So a 64bit compiler may generate slightly different results than a 32bit compiler.
That's mainly a concern for system headers, though, where the data structures themselves may be architecture-dependent (c.f., the pthreads stuff, for instance).
That's also much less annoying in practice with the latest fii-cdecl version which no longer mangles C99 types ;).

The TL;DR; being: I generally use my main, native compiler (in which case, you only need to pass CHOST to make to build the plugin), and I do a second run with `-m32` added to CPPFLAGS to check for bitness issues ;).

NiLuJe:
Oh, speaking of, I *would* mention that ffi-cdecl will honor `CPPFLAGS`, as that's actually extremely useful when dealing with actual libraries (to set `-I` properly) ;).

NiLuJe:
> Oh, speaking of, I _would_ mention that ffi-cdecl will honor `CPPFLAGS`, as that's actually extremely useful when dealing with actual libraries (to set `-I` properly) ;).

Well, to be fair, that bit's already part of ffi-cdecl's own README ;).

tcrs:
So I was thinking that it sounds like these ffi lua files should be generated at build time using the target toolchain? I can see looking at e.g. the pthreads lua ffi files there are several versions for different architectures which are selected at runtime. I assume those where generated using "representative" toolchains for those architectures?
When I was generating the mxcfb ffi lua for remarkable I didn't really think about this as it was only going to be used on that one platform anyway.

tcrs:
(That would also mean having all the `CPPFLAGS` etc required to generate each lua ffi being written down in a makefile or similar, which seems like pretty useful documentation to have!)

NiLuJe:
In an ideal world, sure. Unfortunately, some of them still need manual tweaking to prevent stupid things from slipping through ;).

NiLuJe:
IIRC, I ended up jotting down my own ffi-cdecl call at the top of the gnarliest ones (as far as CPPFLAGS are concerned, mainly).

NiLuJe:
In practice, we check/update them as needed when we bump the matching dependency, so it's not too bad, once you get the hang of it ;).

Issue #5152, opened 2019-07-29
Use curly braces check in ShellCheck 0.7.0


Frenzie:
* KOReader version: 2019.07

When upgrading, look into the new optional checks. See https://www.vidarholen.net/contents/blog/?p=793

@NiLuJe has a strong preference for this one. ;-)

```
$ cat foo
#!/bin/sh
# shellcheck enable=require-variable-braces
name=World
echo "Hello $name"

$ shellcheck foo
In foo line 4:
echo "Hello $name"
            ^---^ SC2250: Prefer putting braces around variable references even when not strictly required.

Did you mean:
echo "Hello ${name}"
```

NiLuJe:
I do like me some curly braces indeed ;p.

Frenzie:
This is still open? Fixed in #5809.

PR #5809, opened 2020-02-02
[CI] Add curly braces check

Frenzie:
Update shellcheck and shfmt to the latest version.

Fixes <#5152>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5809)

### DIFF
--- a/kodev
+++ b/kodev
@@ -145,116 +145,116 @@ ${SUPPORTED_TARGETS}"
         PARAM=$(echo "${1}" | awk -F= '{print $1}')
         VALUE=$(echo "${1}" | awk -F= '{print $2}')
         case "${PARAM}" in
-            -v | --verbose)
-                export VERBOSE=1
-                ;;
-            -h | --help)
-                echo "${BUILD_HELP_MSG}"
-                exit 0
-                ;;
-            --no-debug)
-                export KODEBUG=
-                KODEBUG_NO_DEFAULT=1
-                ;;
-            --debug)
-                export KODEBUG=1
-                KODEBUG_NO_DEFAULT=1
-                ;;
-            *)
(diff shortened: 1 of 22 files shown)


kodev
```diff
@@ -145,116 +145,116 @@ ${SUPPORTED_TARGETS}"
         PARAM=$(echo "${1}" | awk -F= '{print $1}')
         VALUE=$(echo "${1}" | awk -F= '{print $2}')
         case "${PARAM}" in
-            -v | --verbose)
```
Frenzie: @NiLuJe This seems to be a new default; it can be reverted to how it was with `-ci`. Any opinions? The rationale is that it's the same as if elif else for all intents and purposes and should therefore be indented the same.

See https://github.com/mvdan/sh/commit/2e429b0d9b620c0cc360d0adba08f6818fe7c0cf and the associated issue for more details.

NiLuJe: I personally prefer the indented style, but that's basically what it is: a stylistic choice ;).

NiLuJe: (Hell, I actually personally align the `;;` with the conditional, and not the branch ;).)

NiLuJe: TL;DR: Neither are really my own preference, so, no strong opinion on this ;).

Frenzie: Guess I'll just keep it at the default then. Google does it the other way, after all. :-P

NiLuJe: (Although, the indented style is closer to mine, so I'll have less of a chance to shout at shfmt for getting it wrong, which I'll most likely will ^^).

Frenzie: lol, I was just about to click merge

NiLuJe: Sorry ^^.

Issue #4158, opened 2018-08-12
Serve code over HTTPS


alethiophile:
* KOReader version: n/a
* Device: n/a

#### Issue

As it stands, both the nightly builds from the build.koreader.rocks server and the OTA updates from ota.koreader.rocks are served over plain HTTP.

Nowadays, serving runnable code over unencrypted connections is highly contraindicated; it's now easy to get an intercepting proxy server from as little as a host on the same LAN, and software to transparently intercept and substitute files served over HTTP is readily available. For code that will be run on a client's device, using a protocol with integrity verification, such as HTTPS, is a basic security measure. It's also now easy to set up HTTPS on normal servers, given the wide availability of free certificates from e.g. LetsEncrypt.

This is obviously not something I can submit code for, but it's quickly approaching the baseline of security for any code-distribution channel; distributing builds over unencrypted HTTP puts the current servers notably in the lower end of the distribution with respect to security precautions, and this will only get worse as time goes on.

NiLuJe:
`zsync` doesn't support HTTPS.

PR #5810, opened 2020-02-03
Switch to zsync2

NiLuJe:
It's as easy as that ;p.

Requires https://github.com/koreader/koreader-base/pull/1036

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5810)

### DIFF
--- a/frontend/ui/otamanager.lua
+++ b/frontend/ui/otamanager.lua
@@ -19,6 +19,7 @@ local T = require("ffi/util").template
 local ota_dir = DataStorage:getDataDir() .. "/ota/"
 
 local OTAManager = {
+    -- NOTE: Each URL *MUST* end with a /
     ota_servers = {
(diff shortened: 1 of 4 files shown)


NiLuJe:
Fun fact: if I got my ducks in a row, we should handle HTTPS for free!

My own OpenStack mirrors can actually be served over HTTPS, but not through my domain (because HTTPS). Since we don't care all that much, and the real URL are humongous, I left them in plain HTTP.

(Fix #4158 for realz!).

NiLuJe:
Yup, `./zsync2 -i installed.tar -o updated.tar http://storage.gra1.cloud.ovh.net/v1/AUTH_2ac4bfee353948ec8ea7fd1710574097/koreader-ota-fr/koreader-kobo-latest-nightly.zsync` works just fine :).

NiLuJe:
Let's break all the things!

PR #5813, opened 2020-02-04
Stop using the "calibre catalog" wording

NiLuJe:
It's extremely confusing, as it has nothing to do with calibre catalogs,
which are actual things the user can generate manually.

What it's using instead is actually a metadata cache in a flatfile db
that's always present, as long as "Send To Device" was used to upload
the content to the device.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5813)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -477,7 +477,7 @@ function FileManagerMenu:setUpdateItemTable()
 
     -- search tab
     self.menu_items.find_book_in_calibre_catalog = {
-        text = _("Find a book in calibre catalog"),
+        text = _("Find a book via calibre metadata"),
         callback = function()
             Search:getCalibre()
             Search:ShowSearch()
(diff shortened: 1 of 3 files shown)


NiLuJe:
In the OPDS menu, where it could arguably make sense to call it a catalog, instead call that Library, as that's the term calibre uses to refer to its own, well, Library :D.

KenMaltby:
Within the OPDS function, I wonder if the use of "calibre library" might create an expectation that they would get a display such as they see on their PC; rather than the display options that they get with the other sites.  (And with the current "Local calibre catalog" option, under the "OPDS catalog" option of the search tab.)   You could, of course just use "Local calibre", as all the other sites that were provided don't mention catalog. (like "Project Gutenberg")

NiLuJe:
@KenMaltby: You're not wrong. I was working under the assumption that, being under an OPDS header, the fact that you'd be served an OPDS feed would be a given ;).

NiLuJe:
Or, just "calibre server", actually, which is also mildly confusing. Gah.

PR #5814, opened 2020-02-04
end of document: add go to beginning

yparitcher:
Allow `go to beginning` in end actions.
I Use this quite often
Reader_2020-Feb-04_151700
If anyone has any opinions where in the menu it belongs i am all ears, i put it at the end by default.

I would like to add `go to beginning` and `go to end` to the goto UI,
but i am not sure if to put it in the `goto` or `skim` menu.
please advise, as i am sure everyone has their own preferred UI look :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5814)

### DIFF
--- a/frontend/apps/reader/modules/readergoto.lua
+++ b/frontend/apps/reader/modules/readergoto.lua
@@ -119,4 +119,17 @@ function ReaderGoto:gotoPage()
     end
 end
 
+function ReaderGoto:gotoBeginning()
+    self.ui.link:addCurrentLocationToStack()
+    self.ui:handleEvent(Event:new("GotoPage", 1))
+end
+
+function ReaderGoto:gotoEnd()
+    local endpage = self.document:getPageCount()
+    if endpage then
+        self.ui.link:addCurrentLocationToStack()
+        self.ui:handleEvent(Event:new("GotoPage", number))
+    end
+end
+
 return ReaderGoto
--- a/frontend/apps/reader/modules/readergoto.lua
+++ b/frontend/apps/reader/modules/readergoto.lua
@@ -119,17 +119,19 @@ function ReaderGoto:gotoPage()
     end
 end
 
-function ReaderGoto:gotoBeginning()
+function ReaderGoto:onGoToBeginning()
     self.ui.link:addCurrentLocationToStack()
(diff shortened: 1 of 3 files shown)


poire-z:
About the dialog: I don't think `Go to beginning` deserves to be the widest :) So, to not make any of the real action buttons jealous, may be have Cancel as the wide bottom button?
Then, (my) logic would like this:
```
   Mark as read | Book status
Go to beginning | Open next file
    Delete file | File browser
              Cancel
```

> I would like to add go to beginning and go to end to the goto UI,
but i am not sure if to put it in the goto or skim menu.

Menu or dialog?
I often (mostly when testing and browsing books fast up to the end) with the SkimTo widget trying to tap at the beginning/end of the progress bar to go there, and end up with +10/-10 to really reach page 1 or last page.
Buttons for that could be convenient, but the SkimTo widget is already crowded - and there are other ways (like just described above) to do that. So, dunno :)

yparitcher:
reordered the dialog, updated the pic in top post

frontend/apps/reader/modules/readergoto.lua
```diff
@@ -119,4 +119,17 @@ function ReaderGoto:gotoPage()
     end
 end
 
+function ReaderGoto:gotoBeginning()
```
Frenzie: Btw, for calling purposes it's easier to use an event that can be called from the gesture manager.

poire-z:
> I would like to add go to beginning and go to end to the goto UI,

I approved and it's merged, but not stopping you on that :)
I just have no obvious idea to suggest on how to present it.
Or may be non intrusively: when Hold on the prev/new chapter buttons ? currently, hold is managed by (or just reach has not other handler above) MovableContainer, which make the SkimTo widget translucide.

yparitcher:
> > I would like to add go to beginning and go to end to the goto UI,
> 
> I approved and it's merged, but not stopping you on that :)
> I just have no obvious idea to suggest on how to present it.
> Or may be non intrusively: when Hold on the prev/new chapter buttons ? currently, hold is managed by (or just reach has not other handler above) MovableContainer, which make the SkimTo widget translucide.

aready beat you to it ;) see #5820

PR #5816, opened 2020-02-04
File search: Don't traverse hidden folders if we're not showing them

NiLuJe:
Re https://www.mobileread.com/forums/showpost.php?p=3949194&postcount=21

Makes it much faster if you happen to have a lot of crap in dot folders, like I do ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5816)


NiLuJe:
Also ignore macOS resource forks, while we're there.

NiLuJe:
Sneaked in the same fix in BookInfo.

And also hidden macOS resource forks from the FM, because we're mostly always on vfat, so those can proliferate fairly quickly for macOS users...

PR #5819, opened 2020-02-05
[fix] Don't waste time on dir/file attributes in subdirs

Frenzie:
An awful lot of effort was wasted getting attributes just for getting a count of the number of files/directories in subdirectories. Getting the count is pretty fast. Getting all the attributes is not, and in this context we don't even use the attributes.

See <https://www.mobileread.com/forums/showthread.php?t=327085> and <https://gitter.im/koreader/koreader?at=5e3ac3b9e8a838355918ba13> for discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5819)

### DIFF
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -56,12 +56,14 @@ function FileChooser:init()
         end
         return true
     end
-    self.list = function(path, dirs, files)
+    self.list = function(path, dirs, files, count_only)
         -- lfs.dir directory without permission will give error
         local ok, iter, dir_obj = pcall(lfs.dir, path)
         if ok then
             for f in iter, dir_obj do
-                if self.show_hidden or not string.match(f, "^%.[^.]") then
+                if count_only then
+                    table.insert(dirs, true)
+                elseif self.show_hidden or not string.match(f, "^%.[^.]") then
                     local filename = path.."/"..f
                     local attributes = lfs.attributes(filename)
                     if attributes ~= nil then
@@ -211,7 +213,7 @@ function FileChooser:genItemTableFromPath(path)
         local sub_dirs = {}
         local dir_files = {}
         local subdir_path = self.path.."/"..dir.name
-        self.list(subdir_path, sub_dirs, dir_files)
+        self.list(subdir_path, sub_dirs, dir_files, true)
         local num_items = #sub_dirs + #dir_files
         local istr = ffiUtil.template(N_("1 item", "%1 items", num_items), num_items)
         local text
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -61,7 +61,8 @@ function FileChooser:init()
         local ok, iter, dir_obj = pcall(lfs.dir, path)
         if ok then
             for f in iter, dir_obj do
-                if count_only then
+                if count_only
+                   and (self.show_hidden or (not string.match(f, "^%.[^.]") and not util.stringStartsWith(f, "._"))) then
                     table.insert(dirs, true)
                 elseif self.show_hidden or not string.match(f, "^%.[^.]") then
                     local filename = path.."/"..f
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -61,9 +61,10 @@ function FileChooser:init()
         local ok, iter, dir_obj = pcall(lfs.dir, path)
         if ok then
             for f in iter, dir_obj do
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/filechooser.lua
```diff
@@ -56,12 +56,14 @@ function FileChooser:init()
...
-    self.list = function(path, dirs, files)
+    self.list = function(path, dirs, files, count_only)
         -- lfs.dir directory without permission will give error
         local ok, iter, dir_obj = pcall(lfs.dir, path)
         if ok then
             for f in iter, dir_obj do
-                if self.show_hidden or not string.match(f, "^%.[^.]") then
+                if count_only then
+                    table.insert(dirs, true)
+                elseif self.show_hidden or not string.match(f, "^%.[^.]") then
```
poire-z: If `count_only`, shouldn't we count only the not hidden ones if `not show_hidden`?
So the user don't see `Directory name ..... 13 items` and when entering the directory, possibly see only 4 files.

poire-z: At a quick read, I'll fear we might take the elseif branch even if count_only.
Shouldn't it be:
```lua
if count_only then
  if ... check whether hidden ... then
    table.insert(dirs, true)
  end
elseif .. the original checks ... then
  .. the original stuff ...
end
```

Frenzie: If I'm not mistaken the Windows File Manager (or perhaps the one that came with DOS, or an early version of Explorer) did something like that. I always liked it as a kind of insider secret. ;-)

poire-z: (unresolved that conversation, as our last comments might have been simultaneous?)

Frenzie: You're right, that wasn't an issue the way I wrote it not caring about that, but of course it is now.

Frenzie:
There are some other somewhat related seemingly obvious optimizations. For example, here we're using lfs.attributes again instead of our own table that was just filled with all the attributes. I'll have to check on my H2O, but I imagine that could have an impact on slower filesystems.
```diff
diff --git a/frontend/ui/widget/filechooser.lua b/frontend/ui/widget/filechooser.lua
index b91e8725..785917d4 100644
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -239,8 +239,9 @@ function FileChooser:genItemTableFromPath(path)
     local show_file_in_bold = G_reader_settings:readSetting("show_file_in_bold")
 
     for _, file in ipairs(files) do
+    --~ require'logger'.dbg(file)
         local full_path = self.path.."/"..file.name
-        local file_size = lfs.attributes(full_path, "size") or 0
+        local file_size = file.attr.size or 0
         local sstr = getFriendlySize(file_size)
         local file_item = {
             text = file.name,
```

NiLuJe:
Yeah, I don't recall how lfs handles that internally, but it's probably at the very least a variant of stat, and that's notoriously expensive ;).

Frenzie:
@NiLuJe Well… on my laptop it looks like lfs.attributes is just as fast or possibly faster. But I'll also check my H2O.

Frenzie:
On my H2O results seem to be quite comparable (except on my laptop I need 10k files for ~.3 s while the H2O gets ~.6 s with merely 1k files).

```
elapsed time lfs.attributes: 0.56

elapsed time file.attr.size: 0.56

elapsed time lfs.attributes: 0.56

elapsed time file.attr.size: 0.61

elapsed time lfs.attributes: 0.57

elapsed time file.attr.size: 0.58

```

That's just a stupid:

```lua
local x = os.clock()
-- the relevant for loop, see above
print(string.format("elapsed time lfs.attributes: %.2f\n", os.clock() - x))
```

It's even more interesting if we add the timing around the initial `self.list(path, dirs, files)` (but if there *is* caching, obviously I failed to reboot the device):
```
elapsed time initial list: 0.16

elapsed time lfs.attributes: 0.55

elapsed time file.attr.size: 0.56
```

So really it's all the later table processing stuff that's the heavy hitter, which makes intuitive sense. As I wrote on MR, the only real way to improve performance on giant folders would be to leave most of the processing to actually displaying the page. Then you'd effectively only have that initial 0.2, instead of 0.2 + 0.6.

I.e., on the H2O you currently add almost an extra second for every 1000 files, while you could be adding merely an extra 0.2 s.

But given that "normal" folders (< 200 items or so) average out at 0.0 s in the os.clock() timing method even on the H2O, it's hard to get motivated to look into such optimization.

PR #5820, opened 2020-02-05
skimto: add beginning and end as hold callback

yparitcher:
hold the prev chapter button goes to beginning.
hold next chapter goes to the end.
add functionality without messing with the UI

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5820)

### DIFF
--- a/frontend/apps/reader/skimtowidget.lua
+++ b/frontend/apps/reader/skimtowidget.lua
@@ -134,7 +134,7 @@ function SkimToWidget:init()
         width = self.button_width,
         show_parent = self,
         callback = function()
-            self:gotoPageCallback(self.curr_page - 1)
+            self:goToPage(self.curr_page - 1)
         end,
     }
     local button_minus_ten = Button:new{
@@ -146,7 +146,7 @@ function SkimToWidget:init()
         width = self.button_width,
         show_parent = self,
         callback = function()
-            self:gotoPageCallback(self.curr_page - 10)
+            self:goToPage(self.curr_page - 10)
         end,
     }
     local button_plus = Button:new{
@@ -158,7 +158,7 @@ function SkimToWidget:init()
         width = self.button_width,
         show_parent = self,
         callback = function()
-            self:gotoPageCallback(self.curr_page + 1)
+            self:goToPage(self.curr_page + 1)
         end,
     }
     local button_plus_ten = Button:new{
@@ -170,7 +170,7 @@ function SkimToWidget:init()
         width = self.button_width,
         show_parent = self,
         callback = function()
-            self:gotoPageCallback(self.curr_page + 10)
+            self:goToPage(self.curr_page + 10)
         end,
     }
     self.current_page_text = Button:new{
@@ -206,11 +206,11 @@ function SkimToWidget:init()
         callback = function()
             local page = self:getNextChapter(self.curr_page)
             if page and page >=1 and page <= self.page_count then
-                self:gotoPageCallback(page)
+                self:goToPage(page)
             end
         end,
         hold_callback = function()
-            self:gotoPageCallback(self.page_count)
+            self:goToPage(self.page_count)
         end,
     }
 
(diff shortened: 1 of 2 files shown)


poire-z:
Great idea! :)

Frenzie:
That was my first thought, although I had/have some concerns about putting it under the "bookmark" buttons. ("Go to first bookmark" or something might be a more intuitive thing there.)

yparitcher:
> That was my first thought, although I had/have some concerns about putting it under the "bookmark" buttons. ("Go to first bookmark" or something might be a more intuitive thing there.)

?
it is only under the chapter buttons, the bookmark buttons are as is.

Frenzie:
Typo, doesn't make a difference as to my point.

poire-z:
We could implement go to first/last bookmark for consistency.
But I'm fine with hold on prev/next chapter actually going to first page / last page, as, from there, real first and last chapter are just a tap on prev/next away. (If we make them actually go to first/last chapter, well, we lose that unintrusive way of going to book start/end... which is quite more useful that going to first or last chapter :)

yparitcher:
> We could implement go to first/last bookmark for consistency.
> But I'm fine with hold on prev/next chapter actually going to first page / last page, as, from there, real first and last chapter are just a tap on prev/next away. (If we make them actually go to first/last chapter, well, we lose that unintrusive way of going to book start/end... which is quite more useful that going to first or last chapter :)

done,

one caveat, `first bookmark` only goes to a bookmark before the current page (like previous), if you only have later bookmarks it does nothing.
so it really is `go to the first bookmark that is also before the current page`
and the same vice versa.
if anyone feels this should be changed let me know.

poire-z:
> first bookmark only goes to a bookmark before the current page 

Feels right to me. "Go to last chapter" works the same: if you're in a page inside last chapter, it does not take you back to the start of last chapter.

Frenzie:
> (If we make them actually go to first/last chapter, well, we lose that unintrusive way of going to book start/end... which is quite more useful that going to first or last chapter :)

First chapter is the same thing as book start. It's final chapter that's different.

frontend/apps/reader/skimtowidget.lua
```diff
@@ -134,10 +134,7 @@ function SkimToWidget:init()
         width = self.button_width,
         show_parent = self,
         callback = function()
-            self.curr_page = self.curr_page - 1
-            self:update()
-            self:addOriginToLocationStack()
-            self.ui:handleEvent(Event:new("GotoPage", self.curr_page))
+            self:gotoPageCallback(self.curr_page - 1)
```
Frenzie: Is there a maintainability rationale behind including the current use case for the function in the name?
```suggestion
            self:goToPage(self.curr_page - 1)
```

yparitcher:
> > (If we make them actually go to first/last chapter, well, we lose that unintrusive way of going to book start/end... which is quite more useful that going to first or last chapter :)
> 
> First chapter is the same thing as book start. It's final chapter that's different.

many times the first chapter is only after the cover/TOC so it is also different

frontend/apps/reader/skimtowidget.lua
yparitcher: it was originally less confusing to me vs the goToPage event, should i change it back to `goToPage`?

yparitcher: done

frontend/apps/reader/skimtowidget.lua
```diff
@@ -248,7 +248,7 @@ function SkimToWidget:init()
             else
                 page = self.ui.bookmark:getNextBookmarkedPageFromPage(self.curr_page)
             end
-            self:gotoBookmarkCallback(page)
+            self:gotoBookmark(page)
```
Frenzie: Plus the case :angel: 
```suggestion
            self:goToBookmark(page)
```

Frenzie:
> many times the first chapter is only after the cover/TOC so it is also different

Well alright, if a cover is included then the first chapter is sometimes the title page (i.e., page 2). Unless you're making a point about the distinction between the human definition of the first chapter and various more primitive computer TOC definitions of it, but I don't think I've ever seen something like the TOC, colophon or preface fundamentally excluded from the computer TOC.

Anyway, I first/last page may be more predictable even if it does break with internal consistency.

PR #5823, opened 2020-02-06
Add “searchable” capabilities to OPDS catalogs

bateast:
OPDS catalogs may have search query capabilities. It is very usefull to find a
book in huge catalog, even when they are organized as directories. This commit
provides the possibility to add a %s mark in an OPS catalog URL, that would
trigger a search box when browsing the OPDS in koreader.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5823)

### DIFF
--- a/frontend/ui/widget/opdsbrowser.lua
+++ b/frontend/ui/widget/opdsbrowser.lua
@@ -115,14 +115,14 @@ end
 function OPDSBrowser:addServerFromInput(fields)
     logger.info("input catalog", fields)
     local servers = G_reader_settings:readSetting("opds_servers") or {}
-    local new_item = {
+    local new_server = {
         title = fields[1],
         url = (fields[2]:match("^%a+://") and fields[2] or "http://" .. fields[2]),
         searchable =  (fields[2]:match("%%s") and true or false),
         username = fields[3],
         password = fields[4],
     }
-    table.insert(servers, new_item)
+    table.insert(servers, new_server)
     G_reader_settings:saveSetting("opds_servers", servers)
     self:init()
 end
--- a/frontend/ui/widget/opdsbrowser.lua
+++ b/frontend/ui/widget/opdsbrowser.lua
@@ -123,7 +123,6 @@ function OPDSBrowser:addServerFromInput(fields)
       password = fields[4],
     }
     table.insert(servers, new_item)
-    logge.info("new catalog fields", new_item)
     G_reader_settings:saveSetting("opds_servers", servers)
     self:init()
 end
@@ -661,19 +660,19 @@ function OPDSBrowser:showDownloads(item)
     UIManager:show(self.download_dialog)
 end
 
-function OPDSBrowser:browse(url, username, password)
-  logger.info("launch url", url)
+function OPDSBrowser:browse(browse_url, username, password)
+  logger.info("launch url", browse_url)
   table.insert(self.paths, {
-                 url = url,
+                 url = browse_url,
                  username = username,
                  password = password,
   })
-  if not self:updateCatalog(url, username, password) then
+  if not self:updateCatalog(browse_url, username, password) then
     table.remove(self.paths)
   end
 end
 
-function OPDSBrowser:browseSearchable(url, username, password)
+function OPDSBrowser:browseSearchable(browse_url, username, password)
     self.search_server_dialog = InputDialog:new{
         title = _("Search OPDS catalog"),
         input = "",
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/opdsbrowser.lua
```diff
@@ -645,7 +660,55 @@ function OPDSBrowser:showDownloads(item)
     UIManager:show(self.download_dialog)
 end
 
+function OPDSBrowser:browse(browse_url, username, password)
+  logger.info("launch url", browse_url)
```
Frenzie: Seems more of a logger.dbg?

frontend/ui/widget/opdsbrowser.lua
```diff
@@ -645,7 +660,55 @@ function OPDSBrowser:showDownloads(item)
...
+  table.insert(self.paths, {
+                 url = browse_url,
+                 username = username,
+                 password = password,
+  })
+  if not self:updateCatalog(browse_url, username, password) then
+    table.remove(self.paths)
```
Frenzie: Could you please observe the style guide? Four spaces indentation. :-) (That applies to everything in this function; I'm just attaching it to this line more or less at random.)

https://github.com/koreader/koreader-base/wiki/Coding-style

frontend/ui/widget/opdsbrowser.lua
```diff
@@ -645,7 +660,55 @@ function OPDSBrowser:showDownloads(item)
...
+                    callback = function()
+                        UIManager:close(self.search_server_dialog)
+                    end,
+                },
+                {
+                    text = _("Search"),
+                    is_enter_default = true,
+                    callback = function()
+                                UIManager:close(self.search_server_dialog)
+                                local search, _ = self.search_server_dialog:getInputText():gsub(" ", "+")
```
Frenzie: If you want to discard the second one you can just use `local search = function that returns multiple things`.

frontend/ui/widget/opdsbrowser.lua
```diff
@@ -645,7 +660,55 @@ function OPDSBrowser:showDownloads(item)
...
+                                local searched_url, _ = browse_url:gsub("%%s", search)
```
Frenzie: Same.

frontend/ui/widget/opdsbrowser.lua
```diff
@@ -645,7 +660,55 @@ function OPDSBrowser:showDownloads(item)
...
+                                logger.info ("Searched url becomes", searched_url)
+                                self:browse(searched_url, username, password)
+                    end,
+                },
+            }
+        },
+    }
+    logger.info ("display broswse search for url", browse_url)
```
Frenzie: Doesn't really seem necessary anymore.
```suggestion
```

frontend/ui/widget/opdsbrowser.lua
```diff
@@ -645,7 +660,55 @@ function OPDSBrowser:showDownloads(item)
...
+    UIManager:show(self.search_server_dialog)
+    self.search_server_dialog:onShowKeyboard()
+end
+
 function OPDSBrowser:onMenuSelect(item)
+    logger.info("selected catalog item", item)
```
Frenzie: Same as above, seems more of a more of a logger.dbg to me.

Frenzie:
Thanks!

frontend/ui/widget/opdsbrowser.lua
```diff
@@ -645,6 +660,51 @@ function OPDSBrowser:showDownloads(item)
...
+        table.remove(self.paths)
+    end
+end
+
+function OPDSBrowser:browseSearchable(browse_url, username, password)
+    self.search_server_dialog = InputDialog:new{
+        title = _("Search OPDS catalog"),
+        input = "",
+        hint = _("Search string"),
+        input_hint = _("author:dumas alexandre"),
```
Frenzie: I overlooked the fact that this was put in as a translated string, but this `author:` prefix sounds like a technical thing that can't be translated?

ptrm:
Maybe worth mentioning that the list from init()!gets overriden, not merged, by the one from `settings.reader.lua::["opds_servers”]`. That probably desirable, but might confuse users. Confused me at least, until I grepped the other one and removed it ;)

ptrm:
But most of all, I longed for this feature for a long time :D

ptrm:
Also, the redirects might not be supported. Or are supported, but the second download is not allowed to overwrite?

As for the overwrite popup, the result when the epub file had not existed also rendered the same results.

Frenzie:
@ptrm It'd be better to open a new issue.

PR #5824, opened 2020-02-06
Print zsync2's output on screen

NiLuJe:
Instead of the blind spinner we used to have.

This is a bit of a hack (on both sides), and it currently extends towards the middle of the screen, which means the tail end of it will probably get overwritten by the "Success/Failure" popup.

Can easily be tweaked later, I'd like to get it in as-is to try it for real first ;).

Requires https://github.com/koreader/koreader-base/pull/1038

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5824)

### DIFF
--- a/frontend/ui/otamanager.lua
+++ b/frontend/ui/otamanager.lua
@@ -308,8 +308,8 @@ function OTAManager:fetchAndProcessUpdate()
                                         timeout = 3,
                                     })
                                     -- Clear the installed package, as well as the complete/incomplete update download
-                                    os.execute("rm -f" .. self.installed_package)
-                                    os.execute("rm -f" .. self.updated_package .. "*")
+                                    os.execute("rm -f " .. self.installed_package)
+                                    os.execute("rm -f " .. self.updated_package .. "*")
                                     -- As well as temporary files, in case zsync went kablooey too early...
                                     os.execute("rm -f ./rcksum-*")
                                     -- And then relaunch zsync in full download mode...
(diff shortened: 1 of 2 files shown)


NiLuJe:
Obligatory screenshot ;). (From a manual run, hence the lack of InfoMessage).

zsync2

Frenzie:
Not an issue in busybox?
> In platform/common/spinning_zsync line 4:
> set -o pipefail
>        ^------^ SC2039: In POSIX sh, set option pipefail is undefined.
> 
> For more information:
>   https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, set option pipefail ...

NiLuJe:
Not on Kobo, and not on Kindle 5.x at least. I'll double-check on legacy
Kindle just to be sure.

NiLuJe:
Hmm, appeared in busybox 1.16.0 (2010-01-26), conditional on BASH_COMPAT...

I smell a rat. I'll double-check if it behaves properly for real.

If not, `fbink < zsync2` should work instead, right?

Frenzie:
I'd expect it to anyway. We'll just have to try it. ^_^

NiLuJe:
Kobo: BusyBox v1.22.0.kobo (pipefail works)

NiLuJe:
Kindle 2.5: BusyBox v1.7.2 (2011-01-13 18:01:58 PST) (ouch.)

NiLuJe:
Insane shell wizardry incoming!

PR #5827, opened 2020-02-07
[chore] FileChooser: use "traditional" for loop for ~20% better performance

Frenzie:
Apparently there's less overhead this way.

See https://github.com/koreader/koreader/pull/5819 for context.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5827)

### DIFF
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -242,7 +242,8 @@ function FileChooser:genItemTableFromPath(path)
     -- otherwise, show new files in bold
     local show_file_in_bold = G_reader_settings:readSetting("show_file_in_bold")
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
Yay, lua :D.

Were the results consistent, or did it get faster once the JIT was 'primed'?

Frenzie:
I didn't test that in detail, but I assume that's where the 15 % came from when it generally seemed to lean toward 20 %. There seems to be some kind of function call (?) overhead in ipairs. 

In regular Lua the improvement exists as well, but it looks like it can be even more pronounced. But there I just checked some synthetic stuff so it's possible the difference is simply testing loop vs. a real life loop that does Real Things™ other than looping.

PR #5834, opened 2020-02-08
A couple of minor remarkable port fixes

tcrs:
First change just removes some unnecessary code as pointed out by @NiLuJe in the original remarkable support pull request.

Second change is quite important but I don't quite understand the reasoning for the code in `reader.lua` which requires it. If you look at the if statement in `reader.lua` starting at line 239, you can see that if no argument is passed, *and* no `lastfile` setting is set then koreader will just exit with the usage help output.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5834)

### DIFF
--- a/platform/remarkable/koreader.sh
+++ b/platform/remarkable/koreader.sh
@@ -132,7 +132,7 @@ while [ ${RETURN_VALUE} -ne 0 ]; do
         ko_do_fbdepth
     fi
 
-    ./reader.lua "$HOME" >>crash.log 2>&1
+    ./reader.lua "/home/root/" >>crash.log 2>&1
     RETURN_VALUE=$?
 
     # Did we crash?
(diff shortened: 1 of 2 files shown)


platform/remarkable/koreader.sh
```diff
@@ -132,7 +132,7 @@ while [ ${RETURN_VALUE} -ne 0 ]; do
         ko_do_fbdepth
     fi
 
-    ./reader.lua >>crash.log 2>&1
+    ./reader.lua "$HOME" >>crash.log 2>&1
```
Frenzie: The CI will say to make it ${HOME}
```suggestion
    ./reader.lua "${HOME}" >>crash.log 2>&1
```

Frenzie: Actually it would be better to do something like ${args}, like in Kobo. Then you can call `koreader.sh` with arguments that'll be passed to `reader.lua`.

https://github.com/koreader/koreader/blob/6dfe65279c75f37348760fad9f75cb49c8a1eb72/platform/kobo/koreader.sh#L105-L110

tcrs: It makes sense to pass any arguments to koreader.sh through, might be useful

PR #5836, opened 2020-02-09
Honor dithering on diagnoal swipe refreshes

NiLuJe:
c.f., https://github.com/koreader/koreader-base/pull/1039#issuecomment-583494496

Generally affects `setDirty(nil, "full")`, the vast majority of which are indeed diagonal swipes. There are a few stray other uses where this still does make sense ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5836)

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -601,7 +601,7 @@ function UIManager:setDirty(widget, refreshtype, refreshregion, refreshdither)
             for i = 1, #self._window_stack do
                 -- If any of 'em were dithered, honor their dithering hint
                 if self._window_stack[i].widget.dithered then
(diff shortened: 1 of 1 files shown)


NiLuJe:
I'm still unclear on why this sometimes worked before:

That kind of setDirty call shouldn't really ever trigger a repaint, but it sometimes did, and `_repaint` was already doing that check.

NiLuJe:
Probably a different gesture detection.

Short diagonal swipes will repaint, long ones won't. :?

Frenzie:
Long ones take a screenshot? :-P

NiLuJe:
@Frenzie: Oh, right :D.

Turns out they don't in picdocument, but that's probably the source of the discrepancy anyway ;).

PR #5840, opened 2020-02-10
bump crengine: various table and other fixes

poire-z:
Includes https://github.com/koreader/crengine/pull/327 :
- (Upstream) Fix potential segfault
- lvtextfm: avoid spurious spaces when hyphenating (bis)
- View HTML: cosmetic fixes for floats and inline blocks
- Text: avoid clipping of overflowing inline-block content
- Fix images with "display: inline-block" not shown
- elementFromPoint(): skip moved-down block floats
- Rend methods init: fix inconsistency when floats disabled
- getRectEx(): don't crash in case of rend method bug
- getRenderedWidths(): slightly better estimate table width
- Tables cells: handle vertical-align: baseline
- Page splitting: add missing 'discard at start' handling
- Page splitting: split single column tables by lines instead of row
- Hardcoded elements list: add `<legend>`
- Drawing: fix some y/y+h vs page top/bottom comparisons
- Drawing: disable HidePartialGlyphs in scroll mode

Also, some small change to Wikipedia EPUBs to allow this in "web" mode ((just to look more like the original article on Wikipedia "web"...)
<kbd>image</kbd>

while we'll get in 'book' mode as previously (which I think I prefer reading with):
<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5840)

### DIFF
--- a/frontend/ui/wikipedia.lua
+++ b/frontend/ui/wikipedia.lua
@@ -1030,6 +1030,20 @@ body > div:dir(rtl) > div.thumb { /* invert if RTL */
         clear: left;
         margin:  0 0.5em 0.2em 0 !important;
 }
+/* Allow original mix of left/right floats in web mode */
+body > div > div.thumb.tleft {
+    -cr-only-if: float-floatboxes allow-style-w-h-absolute-units;
+        float: left !important;
+        clear: left;
+        margin:  0 0.5em 0.2em 0 !important;
+}
+body > div > div.thumb.tright {
+    -cr-only-if: float-floatboxes allow-style-w-h-absolute-units;
+        float: right !important;
+        clear: right;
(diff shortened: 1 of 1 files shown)


poire-z:
Looks like Wikipedia has changed the structure of their HTML - when Save as EPUB, I do not get the nice little floating bordered boxes with images, like on the screenshots above.

Before:
image

Now:
image


Can you (any reader of this) confirm you also no longer get them in the Wikipedia in your preferred language ? And that you used to get them previously ?

Looks like they now use `<figure>` `<figcaption>` ...
image

I hate it when remote API/services change their stuff....

Frenzie:
This seems to be where it was changed: https://phabricator.wikimedia.org/T51097

PR #5842, opened 2020-02-10
Log zsync2's output, too.

NiLuJe:
Thanks to our good friend tee ;).

As suggested by @Frenzie on gitter ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5842)

### DIFF
--- a/platform/common/spinning_zsync
+++ b/platform/common/spinning_zsync
@@ -30,7 +30,7 @@ rc=0
 
 if [ "${WITH_PIPEFAIL}" = "true" ]; then
     # We can use pipefail, let the shell do the heavy lifting for us...
-    ./zsync2 "$@" 2>&1 | ./fbink -q -z -pm -F scientifica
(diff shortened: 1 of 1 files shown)


NiLuJe:
While the `tee` applet happens to be present on kindle-legacy (at least it was on a K3), I, err, don't quite feel like mangling the pipefail workaround to add another filter in the middle... ^^.

So this only applies to platforms where we can use pipefail ;).

Pinging @pazos to double-check that Cervantes ships with the tee applet, because we've had mishaps with their busybox builds before (sleep/usleep) ;).

NiLuJe:
Fun fact: tee is a good little soldier, and managed to honor the CR, so we don't get a log full of spammy progressbar updates, but only the final state :).

avsej:
```

NiLuJe:
Ooh, real coreutils, cool! Thanks :).

PR #5844, opened 2020-02-11
always use bash for spinning_zsync

avsej:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5844)

### DIFF
--- a/platform/common/spinning_zsync
+++ b/platform/common/spinning_zsync
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # We're going to need to remember failures from the *left* side of a pipeline...
 # shellcheck disable=SC2039


NiLuJe:
Hmm why would the local one be missing? That's the first thing ota manager does to show the update available pop-up.

avsej:
there is no file in the cervantes package, and neither OTA manager, not any other process creates it. Basically right now my cervantes fails to upgrade without this patch.

NiLuJe:
It's not supposed to be in the package.

`OTAManager:checkUpdate` downloads it from the OTA server. As I said, that's the first thing being done, because we parse it to show the "update available popup".

I'd be very interested in some logs ;).

(My reasoning being I don't really want to re-download a file we've already got in full, especially since zsync2 handles that in two passes).

avsej:
Ok, I will apply #5842 and will try to grab logs

avsej:
actually it does not output anything in the log, but when I run it manually, this is what it says:

```
root@cervantes:/mnt/private/koreader# ls -l ota/
total 44
-rw-r--r-- 1 root root 42778 Feb 11 09:30 package.index
root@cervantes:/mnt/private/koreader# ./zsync2 -o ./ota/koreader.updated.tar -u http://ota.koreader.rocks:80/ ./ota/koreader-cervantes-latest-stable.zsync
```

zsync2 version 2.0.0-alpha.1 (commit 7565c08), build <KOReader> built on 2020-02-11 06:29:20 UTC
Checking for changes...
No such file or directory and not a URL: ./ota/koreader-cervantes-latest-stable.zsync
Reading and/or parsing .zsync file failed!
Failed to check for changes!
```

Could you show where exactly it does create the zsync file in the `checkUpdate`?

**Comment by Frenzie** [comment]
https://github.com/koreader/koreader/blob/24f1f435deb556694846752651c890fcf279d69a/frontend/ui/otamanager.lua#L185-L193

**Comment by NiLuJe** [comment]
(i.e., enable debug logging first ;)).

**Comment by avsej** [comment]
@NiLuJe, I found the problem #5824 assumes that `spinning_zsync` is using bash (or that `/bin/sh` is Bash) which is not the case on cervantes. therefore `set -o pipefail` exits with error code and terminates the wrapper. I've updated the PR with the fix

**Comment by NiLuJe** [comment]
Unfortunately, we have no /bin/bash on most every other platform, so that doesn't work.

I'm confused as to why the `set` actually exists. `set -e` isn't, well, set, and the command is handled in a test o_O.

**Comment by NiLuJe** [comment]
Oh, apparently I failed to test this in an actual script.

ash indeed aborts on a set failure o_O.

**Comment by NiLuJe** [comment]
Oh, there's also a very stupid typo in there :D

(I left `WITH_PIPEFAIL` set to true in both branches xD).

**Comment by NiLuJe** [comment]
Yeah, that was "fixed" to mimic Bash (i.e., not exit on set -o failures) later :/.

Raaaaaah.

File: platform/common/spinning_zsync [hunk 1]
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
Frenzie: Definitely not, there's no better way to guarantee a problem like this. ;-)

NiLuJe:
Okay, #5850 should handle that in a "portable" (lol) manner. Thanks, POSIX.

avsej:
I will let you guys to decide. But please propagate `tee -a` patch to non-pipefail version too

NiLuJe:
I personally have no interest in mangling the no-pipefail codepath further, so, PRs welcome ;).

avsej:
@NiLuJe then maybe just fix it to use bash for cervantes and that's all? something like @Frenzie  proposed https://github.com/koreader/koreader/pull/5850#discussion_r377862347 or `sed` in Makefile

Issue #5843, opened 2020-02-11
wrong file count in folders


BrainDamage:
* KOReader version:V2020.01-28-ga2eea88d_2020-02-10
* Device: Kobo Aura H2O2

#### Issue
File count in folders is inconsistent and/or wrong, there's multiple issues here
#### Steps to reproduce
first bug:
create an empty directory
enable "show hidden files" in the file manager
the empty directory will show 2 elements, and if new files are dragged in, the offset will be always shifted by 2
I suspect it's counting the '.' current directory and '..' parent path

second issue:
create an empty directory
set "show hidden files" to disabled ( although it's irrelevant since koreader hides .sdr folders regardless )
drag inside a new ebook, never read, never cached
verify item count, it should be 1 if "show hidden files" is disabled, 3 if enabled
open the ebook so it generates a .sdr directory even if it's hidden by the ui
verify item count of the directory, it has been increased by 1
delete the .sdr directory and the count goes back by 1

while it's not technically wrong because the content exists, the count is inconsistent with the displayed elements, making it confusing

BrainDamage:
second issue still present as of commit de15ab3, but now only when "show hidden files" is disabled

to reproduce:

set "show hidden files" to disabled
create an empty directory
drag inside a new ebook, never read, never cached
verify item count, it should be 1
open the ebook so it generates a .sdr directory even if it's hidden by the ui
verify item count of the directory, it is now 2
delete the .sdr directory and the count goes back to 1

PR #5846, opened 2020-02-11
[fix] More accurate count in FileChooser

Frenzie:
Fixes <https://github.com/koreader/koreader/issues/5843>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5846)

### DIFF
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -62,7 +62,11 @@ function FileChooser:init()
         if ok then
             for f in iter, dir_obj do
                 if count_only then
-                    if self.show_hidden or not util.stringStartsWith(f, ".") then
+                    if
+                        not self.show_hidden and not util.stringStartsWith(f, ".")
+                    or
+                        (self.show_hidden and f ~= "." and f ~= ".." and self.dir_filter(f) and not util.stringStartsWith(f, "._"))
+                    then
                         table.insert(dirs, true)
                     end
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/filechooser.lua
```diff
@@ -62,7 +62,11 @@ function FileChooser:init()
         if ok then
             for f in iter, dir_obj do
                 if count_only then
-                    if self.show_hidden or not util.stringStartsWith(f, ".") then
+                    if
+                        not self.show_hidden and not util.stringStartsWith(f, ".")
+                    or
+                        (self.show_hidden and f ~= "." and f ~= ".." and self.dir_filter(f) and not util.stringStartsWith(f, "._"))
+                    then
```
poire-z: Not fond of that syntax :) (`if or then` on the same indent level - `if` with nothing following it.
I'd rather have multiple if / elseif with the action (small table.insert) duplicated.
If you prefer keeping a single if, I'd then prefer to have parens around the first alternative `if (not and not) or (and and and not)`, make it easier to understand and not worry about logical operator precedence (and > or).

And in the first alternative, won't you be including the `._` Mac ressources?

Frenzie: Also starts with `.`, doesn't it? ;-)

poire-z: Oh, right :| :)

PR #5849, opened 2020-02-11
Footer: bold font option

robert00s:
cf. https://www.mobileread.com/forums/showpost.php?p=3946581&postcount=5 and https://www.mobileread.com/forums/showpost.php?p=3938492&postcount=5

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5849)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -871,7 +871,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                         keep_menu_open = true,
                     },
                     {
-                        text = _("Bold"),
+                        text = _("Use bold font"),
                         checked_func = function()
                             return self.settings.text_font_bold == true
                         end,
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -855,16 +855,13 @@ function ReaderFooter:addToMainMenu(menu_items)
                                 title_text =  _("Footer font size"),
                                 callback = function(spin)
                                     self.settings.text_font_size = spin.value
-                                    --local text = self.footer_text.text
+                                    self.footer_text:free()
                                     self.footer_text = TextWidget:new{
                                         text = self.footer_text.text,
                                         face = Font:getFace(self.text_font_face, self.settings.text_font_size),
                                         bold = self.settings.text_font_bold,
                                     }
-                                    self.text_container = RightContainer:new{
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -828,27 +835,55 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                        end,
+                        callback = function(touchmenu_instance)
+                            self.settings.text_font_bold = not self.settings.text_font_bold
                             self.footer_text = TextWidget:new{
-                                text = text,
-                                face = Font:getFace(self.text_font_face, self.settings.text_font_size)
+                                text = self.footer_text.text,
+                                face = Font:getFace(self.text_font_face, self.settings.text_font_size),
+                                bold = self.settings.text_font_bold,
                             }
```
poire-z: Here and above:
- please call `self.footer_text:free()` before replacing it, to release some resources (xtext) used by the old one - it would be free'd by the Garbage Collector, but it's best to think about doing it explicitely. (When the widget is closed by UIManager as all usually are, no need to do it explicitely.)
- a bit unfortunate to have that RightContainer re-created here (if we change some stuff at the original place, we'll probably forget to change it here) - I think just doing: `self.text_container[1] = self.footer_text` should work
- I would say "a bit unfortunate to have that TextWidget re-created here" but well, it's quite simple, so that's ok (the alternative would have been to add a `TextWidget:setFace(face, bold)` like the existing setText/setMaxWidth, but you'd probably be the only one bit of code that would use it, so, well, let's keep it that way :)

robert00s: Thanks. Done :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -828,39 +835,61 @@ function ReaderFooter:addToMainMenu(menu_items)
...
-                            self.text_container = RightContainer:new{
-                                dimen = Geom:new{ w = 0, h = self.height },
-                                self.footer_text,
+                            UIManager:show(items_font)
+                        end,
+                        keep_menu_open = true,
+                    },
+                    {
+                        text = _("Bold"),
```
poire-z:
```suggestion
                        text = _("Use bold font"),
```

("Bold" feels a bit short for a menu item. I used "Use monospace font" in the TextEditor plugin - but there, I dunno, it's not like "use any bold font", it's rather "make that font bold")

Frenzie: `Bold statusbar`?

PR #5850, opened 2020-02-11
Unbreak OTA on crappy shell

NiLuJe:
Because everything is terrible.

Fix #5844

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5850)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -451,8 +451,8 @@ cervantesupdate: all
 	# remove old package if any
 	rm -f $(CERVANTES_PACKAGE)
 	# Cervantes launching scripts
+	cp $(COMMON_DIR)/spinning_zsync $(INSTALL_DIR)/koreader/spinning_zsync.sh
 	cp $(CERVANTES_DIR)/*.sh $(INSTALL_DIR)/koreader
-	cp $(COMMON_DIR)/spinning_zsync $(INSTALL_DIR)/koreader
 	# create new package
 	cd $(INSTALL_DIR) && \
 		zip -9 -r \
--- a/frontend/ui/otamanager.lua
+++ b/frontend/ui/otamanager.lua
@@ -399,14 +399,22 @@ end
 function OTAManager:zsync(full_dl)
     if full_dl or self:_buildLocalPackage() == 0 then
         local zsync_wrapper = "zsync2"
+        local use_pipefail = true
         -- With visual feedback if supported...
         if self.can_pretty_print then
             zsync_wrapper = "spinning_zsync"
+            -- And because everything is terrible, we can't check for pipefail's usability in spinning_zsync,
+            -- because older ash versions abort on set -o failures...
+            -- c.f., ko/#5844
+            -- So, instead, check from this side of the fence...
+            -- (remember, os.execute is essentially system(), it goes through sh)
+            use_pipefail = (os.execute("set -o pipefail 2>/dev/null") == 0)
         end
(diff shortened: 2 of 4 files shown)


NiLuJe:
Compare https://git.busybox.net/busybox/tree/shell/ash.c?h=1_7_stable#n9004 to https://git.busybox.net/busybox/tree/shell/ash.c?h=1_31_stable#n11090

(With the relevant explanation above in the current code).

frontend/ui/otamanager.lua
```diff
@@ -399,14 +399,22 @@ end
...
+        local use_pipefail = true
         -- With visual feedback if supported...
         if self.can_pretty_print then
             zsync_wrapper = "spinning_zsync"
+            -- And because everything is terrible, we can't check for pipefail's usability in spinning_zsync,
+            -- because older ash versions abort on set -o failures...
+            -- c.f., ko/#5844
+            -- So, instead, check from this side of the fence...
+            -- (remember, os.execute is essentially system(), it goes through sh)
+            use_pipefail = (os.execute("set -o pipefail 2>/dev/null") == 0)
```
avsej: which shell do you test here? are you sure that `os.execute` uses `/bin/sh`?

It is sad that on cervantes we can have pipefail if you will check for `/bin/bash` availability and fix the shebang.

NiLuJe: [Yup](https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/lib_os.c#L52)

>       The system() library function uses fork(2) to create a child process that executes the shell command specified in command using execl(3) as follows:
>          execl("/bin/sh", "sh", "-c", command, (char *) NULL);

NiLuJe: Unfortunately, checking the for availability of a bash symlink wouldn't be enough, because it could be a symlink to ash or dash or any other shell that may not support pipefail.

Which leaves hard-coding a Cervantes-only codepath, which is probably best served by shipping a dedicated spinning_zsync in the platform-specific code, and overriding the common one with it when building the package.

NiLuJe: i.e., https://github.com/koreader/koreader/blob/master/Makefile#L455

Frenzie: Btw, that platform-specific file can have /bin/bash shebang just fine.

Would it do to simply use something like this?

```sh
#!/bin/bash
source common_zsync_spinner
```

NiLuJe: Without touching anything else, that would probably work with an `exec` instead but that's indeed a neat idea to avoid code duplication ;).

NiLuJe:
@avsej: Okay, that should *probably* work ;).

Frenzie:
Excellent ;-)

Makefile
```diff
@@ -451,8 +451,8 @@ cervantesupdate: all
 	# remove old package if any
 	rm -f $(CERVANTES_PACKAGE)
 	# Cervantes launching scripts
+	cp $(COMMON_DIR)/spinning_zsync $(INSTALL_DIR)/koreader/spinning_zsync.sh
 	cp $(CERVANTES_DIR)/*.sh $(INSTALL_DIR)/koreader
```
avsej: I don't think this line covers `platform/cervantes/spinning_zsync` file

NiLuJe: Nope, I missed the `*.sh` (which I read as `*`) in the CERVANTES one, good catch ^^.

PR #5853, opened 2020-02-12
Actually ship the Cervantes version of spinning_zsync

NiLuJe:
Followup to #5850 ;).

Ooops. Thanks @avsej ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5853)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -453,6 +453,7 @@ cervantesupdate: all
 	# Cervantes launching scripts
 	cp $(COMMON_DIR)/spinning_zsync $(INSTALL_DIR)/koreader/spinning_zsync.sh
 	cp $(CERVANTES_DIR)/*.sh $(INSTALL_DIR)/koreader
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -453,6 +453,7 @@ cervantesupdate: all
 	# Cervantes launching scripts
 	cp $(COMMON_DIR)/spinning_zsync $(INSTALL_DIR)/koreader/spinning_zsync.sh
 	cp $(CERVANTES_DIR)/*.sh $(INSTALL_DIR)/koreader
+	cp $(CERVANTES_DIR)/spinning_zsync $(INSTALL_DIR)/koreader
```
Frenzie: What happens with the readme and those service files? Is it supposed to be specific things or basically just `/*`?

NiLuJe: That's probably why the first one is `*.sh` ;).

The second one is a single file, so it shouldn't pull anything else.

And if you were asking where the other stuff goes: I don't know ^^.

PR #5854, opened 2020-02-12
Statistics: new Calendar view

poire-z:
Because KOReader knows what you read last summer 😬
and when you get old, you don't anymore. So, this should help us remembering, and give us an interesting view of our reading activity.

<kbd>image</kbd>

Books are vertically ordered by reading duration in that day (but continuations from previous days override that), and when there are more than 3, there's a little `+2` above, and tap/hold on that day open the classic Statistics book per day list view.
The bottom of each day shows some histogram of reading duration per hour, so one can know when a book kept it awake at night, or when he had a day off work.

No setting, but if needed, stuff is already there to specify starting day of week (monday, sunday...), the nb of book spans to show per day (default is 3, less makes font larger, but then, there's less width to show the title, so they are more easily truncated, and it's less interesting), and whether to show or not the histogram.
Also not done, as it's to be done more generically for all Statistics views: correct localization of dates. So, they all have their day/month or month/year order a bit hardcoded.

@WaseemAlkurdi : is that how calendars are supposed to look in your flipped side of civilisation? :)

<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5854)

### DIFF
--- a/frontend/ui/widget/container/inputcontainer.lua
+++ b/frontend/ui/widget/container/inputcontainer.lua
@@ -272,7 +272,7 @@ function InputContainer:onInput(input, ignore_first_hold_release)
     local InputDialog = require("ui/widget/inputdialog")
     self.input_dialog = InputDialog:new{
         title = input.title or "",
-        input = input.input,
+        input = input.input_func and input.input_func() or input.input,
         input_hint = input.hint_func and input.hint_func() or input.hint or "",
         input_type = input.type or "number",
         buttons =  input.buttons or {
--- a/frontend/ui/widget/overlapgroup.lua
+++ b/frontend/ui/widget/overlapgroup.lua
@@ -56,16 +56,21 @@ function OverlapGroup:paintTo(bb, x, y)
     local size = self:getSize()
 
     for i, wget in ipairs(self) do
+        local wget_size = wget:getSize()
         local overlap_align = wget.overlap_align
         if self._mirroredUI and self.allow_mirroring then
+            -- Checks in the same order as how they are checked below
             if overlap_align == "right" then
                 overlap_align = "left"
-            elseif overlap_align ~= "center" then
+            elseif overlap_align == "center" then
+                overlap_align = "center"
+            elseif wget.overlap_offset then
+                wget.overlap_offset[1] = size.w - wget_size.w - wget.overlap_offset[1]
+            else
                 overlap_align = "right"
             end
             -- see if something to do with wget.overlap_offset
         end
-        local wget_size = wget:getSize()
         if overlap_align == "right" then
             wget:paintTo(bb, x+size.w-wget_size.w, y)
         elseif overlap_align == "center" then
--- a/plugins/statistics.koplugin/calendarview.lua
+++ b/plugins/statistics.koplugin/calendarview.lua
@@ -301,7 +301,7 @@ function CalendarWeek:update()
     local bspan_margin_v = Size.margin.tiny
     local bspan_padding = Size.padding.tiny -- only used to compute the adequate font size
     local bspan_border = Size.border.thin
-    local text_height = self.span_height - 2*(bspan_margin_v + bspan_border + bspan_padding)
+    local text_height = self.span_height - 2 * (bspan_margin_v + bspan_border + bspan_padding)
     local inner_font_size = TextBoxWidget:getFontSizeToFitHeight(text_height, 1, 0)
     for col, day_books in ipairs(self.days_books) do
         for row, book in ipairs(day_books) do
@@ -406,7 +406,7 @@ function CalendarView:init()
     -- 7 days in a week
     self.day_width = math.floor((self.dimen.w - 2*outer_padding - 6*self.inner_padding) / 7)
     -- Put back the possible 7px lost in rounding into outer_padding
-    outer_padding = math.floor((self.dimen.w - 7*self.day_width - 6*self.inner_padding)/2)
+    outer_padding = math.floor((self.dimen.w - 7*self.day_width - 6*self.inner_padding) / 2)
     self.content_width = self.dimen.w - 2*outer_padding
 
     if not MIN_MONTH then
@@ -470,7 +470,7 @@ function CalendarView:init()
         text = "",
         hold_input = {
             title = _("Enter month"),
-            input = self.cur_month,
+            input_func = function() return self.cur_month end,
             callback = function(input)
                 local year, month = input:match("^(%d%d%d%d)-(%d%d)$")
                 if year and month then
--- a/plugins/statistics.koplugin/calendarview.lua
+++ b/plugins/statistics.koplugin/calendarview.lua
@@ -1,0 +1,762 @@
+local BD = require("ui/bidi")
+local Blitbuffer = require("ffi/blitbuffer")
+local BottomContainer = require("ui/widget/container/bottomcontainer")
+local Button = require("ui/widget/button")
+local CenterContainer = require("ui/widget/container/centercontainer")
+local CloseButton = require("ui/widget/closebutton")
+local Device = require("device")
+local Font = require("ui/font")
+local FrameContainer = require("ui/widget/container/framecontainer")
+local Geom = require("ui/geometry")
+local GestureRange = require("ui/gesturerange")
+local HorizontalGroup = require("ui/widget/horizontalgroup")
+local HorizontalSpan = require("ui/widget/horizontalspan")
+local InputContainer = require("ui/widget/container/inputcontainer")
+local KeyValuePage = require("ui/widget/keyvaluepage")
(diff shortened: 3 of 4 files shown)


NiLuJe:
Ooh, that's mighty cool :).

NiLuJe:
Taking a look at the screenshot: would this work better in Landscape? (or, well, more specifically, could one swap that to landscape on demand?)

poire-z:
It's displayed in landscape if you were in landscape, but it's not as nice (less height for each week, less bookspan height, smaller font size) - although may be that's readable on a device:

<kbd>image</kbd>

NiLuJe:
Couldn't the massive river between the last week and the pageturn buttons be reduced to gain a bit of extra content and/or breathing room between day cards?

poire-z:
That space is reserved for months that span 6 weeks. If you adapt the height of each week depending on the nb of weeks the month spans, it gets a bit ugly when browsing prev/next months as things would be growing/shrinking. And breathing room between weeks gives less space for book titles and a smaller font size.

poire-z:
Btw, I initially thought I would have some fun with using real (red, blue...) and varied background colors depending on some book title hash, and having to find the most readable white|black foreground color to use for text with that background color... and gave up after an hour trying to make it display colors :) Which is fine!

So, I guess we have everything working only on 8bits blitbuffer, that makes everything becoming gray, and only images use some codepath that can keep colors? Or did I miss something?
(Just curious, not planning on redoing all that with colors for Android :)

NiLuJe:
Yep, the BB.COLOR_* constants will render true to their name (i.e, Grayscale) everywhere ;).

poire-z:
OK for the GRAYishes named ones, but naively using `TextWidget:new{fgcolor = BlitBuffer.Color32(0xFF, 0x00, 0x00, 0x80)}` or `FrameContainer:new{color=...}` will render untrue to their 32 bits nickname: as gray and not red :)
There's some `v=getColor8(value)` in most codepaths (paintRect for this one) that clip it to gray, I gathered.

robert00s:
Wow, awesome :)

WaseemAlkurdi:
@poire-z Simply brilliant!
And yeah, from the flipped side of civilization, this is how we do it. 😇

WaseemAlkurdi:
@poire-z I remembered something:
Where we live, the week starts from Sunday. (The weekend is Friday and Saturday, instead of Saturday and Sunday).
Can we have a setting to reflect that?
If it's too much work, then don't mind, leave it as is.

poire-z:
The work is already done, there is a hardcoded:
https://github.com/koreader/koreader/blob/ecbfb4d44c8daca00ad1368d1add2a26e1cac092/plugins/statistics.koplugin/calendarview.lua#L366
so it's just a matter of setting it to the appropriate value.

Is that orthogonal to the UI language? Or could I just hardcode a few UI languages to use `startDayOfWeek = 1`.

Otherwise, indeed, a menu setting, but that's some less fun and interesting coding :)

WaseemAlkurdi:
@poire-z Well, I'd rather prefer it to be a menu setting. This accounts for people who use an RTL language, yet live in a country where the week starts on Monday.

Frenzie:
> Is that orthogonal to the UI language?

Only somewhat. English uses both, although there's a bit of a split between North American English (mostly Sunday) and English elsewhere (mostly Monday).

And while officially the Dutch week starts on Monday (it's actually defined in an ISO standard and a NEN standard), it wouldn't surprise me if devout Christians for example preferred calendars that showed it starting on Sunday.

I think/hope no one would make a big issue out of it either way though. ^_^

poire-z:
Can I assume that there's need for only 2 choices for the first day of a week: Monday or Sunday - and make that a checkbox?
Or can Saturday or other days be too - and it will have to be more than a checkbox?
(https://en.wikipedia.org/wiki/Week#Week_numbering says Saturday is in "Much of the Middle East" - but there's not much discussion about it in https://en.wikipedia.org/wiki/Talk:Week#Dubious, unlike about Monday vs Sunday :)

Frenzie:
In Outlook you can pick any day you want as the first day of the week. In Google Calendar you're limited to Saturday, Sunday, and Monday.

NiLuJe:
Long-press on the day's name in the header to set as first day of the week? That'd save you hiding that away in the far away date/time menu ;).

poire-z:
> Long-press on the day's name in the header to set as first day of the week

Damned, that was a good idea ! But I already did some work on extending `Statistics>Settings` into a submenu, where start of day selection will find its place (hoping we won't need it for other things :)

poire-z:
The existing settings needed some explanation (the hourly histograms on my screenshots have me look like a lazy reader, while I'm only a slow reader :) - and I never changed these default settings).

Wording suggestions welcomed (using "calendar" in each of the 3 settings that apply to Calendar view to avoid a submenu):

<kbd>image</kbd>

<kbd>image</kbd>

I'll add may be `[ ] Allow browsing months in the future` for those who want to use it as a real calendar (the default of being stopped at current month when browsing is quite fine otherwise).

NiLuJe:
`included` or `accounted for` (`accounted` on its own doesn't quite work).

Frenzie:
Grammatically speaking it should be `accounted for`, but semantically it should just be something like included or counted.

KenMaltby:
Not interested in keeping reading statistics, but looks like it could be a good little Tasks/projects management calendar.  Of course, the functions to allow a user to populate the calendar would be not exactly that closely related to reading.  I like it as a stand alone feature.

Frenzie:
My second thought when I saw this was actually something like:

* Download ICS
* Parse ICS
* Display as screensaver

My first thought was the weather forecast (see [here](https://openweathermap.org/api) or something).

WaseemAlkurdi:
@Frenzie 
Screensavers don't work on new Kindles (at least the newer firmware versions). Instead, only "Sleeping" is shown.
@KenMaltby Seconded. It'd be useful as a separate widget that has a "[  ] track my reading" checkbox that turns it into what we currently have.
@poire-z I gave that Wikipedia talk page a quick look. The discussion has delved into minutiae about when the day starts, et cetera, which definitely needs a good "reading".
But it's mainly about the _definition_ of the beginning of a week, not about which days are taken as a weekend.
So, to avoid the semantics of "beginning of the week":
In most countries of the Middle East, Friday and Saturday are the days on which businesses close, et cetera. IIRC, the rationale for Friday being taken off is because Muslims have a Friday congregational prayer on Friday afternoon (but it isn't strictly religiously mandated to take the day off, which is why i.e. Morocco doesn't take Friday off), and for Saturday is because Christians and Jews observe the Sabbath, in which working is actually prohibited by scripture (though do correct me if I'm mistaken).

Frenzie:
> Saturday is because Christians and Jews observe the Sabbath, in which working is actually prohibited by scripture (though do correct me if I'm mistaken).

Sunday is "Christian Sabbath" or whatever you want to call it. (There may be some minority sects who do it on Saturday.)

WaseemAlkurdi:
@Frenzie It seems that it has become a Sunday at some later point in time: https://www.ucg.org/bible-study-tools/bible-questions-and-answers/which-day-is-the-sabbath-according-to-the-bible
I'm tending to think that when the days for a weekend were "decided", the Sabbath was, at least in the Middle East, on a Saturday.
However, Christians here have church service on a Sunday, so I'm not really sure what this might mean.

poire-z:
(Not interested at all in doing anything non-reading related :)
We probably have the best UI toolkit to quickly do any app on Kobo/Kindle..., and I would use KOReader as the base/OS for any specific applet need of mine. But what we could do will always be half-baked compared to what's already available on our phones :) So, I'd rather keep KOReader a fully baked reader app, than having a KoOS with clumsy and limited sketch, note taking, diary, tasks, todo / shopping lists, calendar, weather... apps.)

Frenzie:
There's also Python + FBInk. Not so much for interactivity I imagine, but if you want to fairly quickly repurpose an old Kobo as some kind of weather display or something.

arooni:
this is way neat.  i think any future ereader i purchase is going to have to support koreader or i won't be buying it.  

you folks should link all the major ereaders from your project page so you can get any commissions for future sales.

arooni:
Would also love to see this linked from the weekly reading status screen that I also use as my Kindle device screensaver.  That way I could jump from how many minutes a day I read to what I read.  Would also love a week by week view.

rigogsilva:
I am not seeing this calendar view option in my device: boox nova 2. 🤔

poire-z:
You should have the "Reader statistics" plugin enabled (there, in `More tools > Plugin management`), and it should show up  in this `Reading statistics` submenu:

rigogsilva:
So, I had to to to settings and check for an update. After updating it showed up. 

The download on the nova 2 is a bit old 2019...

PR #5855, opened 2020-02-12
Spelling: e.g

comradekingu:
As per https://en.wiktionary.org/wiki/e.g.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5855)

### DIFF
--- a/frontend/apps/cloudstorage/webdav.lua
+++ b/frontend/apps/cloudstorage/webdav.lua
@@ -50,7 +50,7 @@ The start folder is appended to the server path.]])
 
     local hint_name = _("Server display name")
     local text_name = ""
-    local hint_address = _("WebDAV address e.g. https://example.com/dav")
+    local hint_address = _("WebDAV address, for example https://example.com/dav")
     local text_address = ""
     local hint_username = _("Username")
     local text_username = ""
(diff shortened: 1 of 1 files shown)


frontend/apps/cloudstorage/webdav.lua
```diff
@@ -50,7 +50,7 @@ The start folder is appended to the server path.]])
 
     local hint_name = _("Server display name")
     local text_name = ""
-    local hint_address = _("WebDAV address eg https://example.com/dav")
+    local hint_address = _("WebDAV address e.g. https://example.com/dav")
```
Frenzie: Formally speaking:
```suggestion
    local hint_address = _("WebDAV address (e.g., https://example.com/dav)")
```

But maybe just:
```suggestion
    local hint_address = _("WebDAV address, for example https://example.com/dav")
```

pazos: I do like the second string.

NiLuJe: It'll probably go easier on non-Western translators, for which Latin may not be as pervasive in their native language/culture ;).

PR #5858, opened 2020-02-13
file count: Fixes #5843

### LINKED ISSUE #5843: wrong file count in folders

yparitcher:
as per https://github.com/koreader/koreader/issues/5843#issuecomment-585531475
better implementation of #5846 
never count `.sdr` directory

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5858)

### DIFF
--- a/frontend/ui/widget/filechooser.lua
+++ b/frontend/ui/widget/filechooser.lua
@@ -62,8 +62,8 @@ function FileChooser:init()
         if ok then
             for f in iter, dir_obj do
                 if count_only then
-                    if (not self.show_hidden and not util.stringStartsWith(f, "."))
-                        or (self.show_hidden and f ~= "." and f ~= ".." and self.dir_filter(f) and not util.stringStartsWith(f, "._"))
+                    if self.dir_filter(f) and ((not self.show_hidden and not util.stringStartsWith(f, "."))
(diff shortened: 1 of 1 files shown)


Frenzie:
My bad, I overlooked the extra parentheses on my phone.

Issue #5859, opened 2020-02-13
Blockquote handling in html-dictionary. (Sametypesequence=h)


Markismus:
* KOReader version: 2020.01
* Device: linux/kobo Aura h2o

#### Issue
The indent on blockquote's is rather fierce. It's making a dictionairy such as Oxford Dictionary 2nd Edition non-readable on e-readers. Enlarging the koreader window to full screen shows that it is rendered nicely on a larger scale. 
What css controls the dictionairy display? Can we set the indent for blockquote as a percentage of the display width?

I've included some screen shots. Due to resizing it might not be directly apparent that the indent for the blockquote is of equal size in all screenshots.
Full screen:
![Screenshot from 2020-02-13 10-06-55](https://user-images.githubusercontent.com/5269101/74419637-868aa100-4e4a-11ea-8824-9e76c1eb0a1f.png)
![Screenshot from 2020-02-13 10-07-06](https://user-images.githubusercontent.com/5269101/74419642-87233780-4e4a-11ea-9755-f22fcfdddcd1.png)

Normal window:
![Screenshot from 2020-02-13 10-07-33](https://user-images.githubusercontent.com/5269101/74419644-87bbce00-4e4a-11ea-9816-44d6656cb066.png)
![Screenshot from 2020-02-13 10-07-49](https://user-images.githubusercontent.com/5269101/74419647-87bbce00-4e4a-11ea-9c77-d1608ad07052.png)
![Screenshot from 2020-02-13 10-08-17](https://user-images.githubusercontent.com/5269101/74419651-88ecfb00-4e4a-11ea-8457-8b9d9c9c3cfb.png)

#### Steps to reproduce
I've created a small excerpt from the Oxford dictionary and uploaded it [here.](https://markismus.stackstorage.com/s/Y6J6Xic3FVgWQHa) Lookup 'a'.

Frenzie:
There's a CSS snippet in the dictionary or rich text widget file iirc. I'll check in a bit.

Frenzie:
@Markismus It's not so much the default styles (something like margin-left & right `1em`) but the content of your dictionary:

```
[...]

PR #4358, opened 2018-11-24
[UX] Add paragraph whitespace/indent style tweaks

Frenzie:
Requested on the forum, and it makes a lot of sense.

https://www.mobileread.com/forums/showthread.php?p=3777972#post3777972

PR #5863, opened 2020-02-14
Fonts, Style tweaks: smaller menus to see more of the book

poire-z:
Show only 5 fonts by menu page to get more room below to see how the text looks with the selected font. Not sure everybody will like that (needs to extend your thumb more to browse menu pages), but the net effect is that you see the font and don't need to close the menu as previously to appreciate how it looks, and go back re-opening that menu to test another font.

<kbd>fontmenu</kbd>

Re-organize style tweaks into more submenus to also get more room below.
Add a few styletweaks: "No indentation on first/following paragraphs" https://github.com/koreader/koreader/pull/4358#issuecomment-585591943, "Paragraphs margins and paddings", and Links underlining.
Popup footnotes: some CSS fixes https://github.com/koreader/koreader/issues/5859#issuecomment-585664640

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5863)

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -34,6 +34,7 @@ local CssTweaks = {
                 title = _("Ignore vertical margins"),
                 priority = 2,
                 css = [[* { margin-top: 0 !important; margin-bottom: 0 !important; }]],
+                separator = true,
             },
             {
                 id = "padding_horizontal_all_0";
@@ -50,78 +51,81 @@ local CssTweaks = {
             separator = true,
         },
         {
-            id = "titles_page-break-before_avoid ";
-            title = _("Avoid blank page on chapter start"),
-            priority = 2, -- so it can override the one put back by publisher_page-break-before_avoid
-            css = [[h1, h2, h3 { page-break-before: auto !important; }]],
+            title = _("Page breaks and blank pages"),
+            {
+                id = "titles_page-break-before_avoid ";
+                title = _("Avoid blank page on chapter start"),
+                priority = 2, -- so it can override the one put back by publisher_page-break-before_avoid
+                css = [[h1, h2, h3 { page-break-before: auto !important; }]],
 
-        },
-        {
-            id = "docfragment_page-break-before_avoid ";
-            title = _("Avoid blank page on chapter end"),
-            priority = 2, -- so it can override the one put back by publisher_page-break-before_avoid
-            css = [[DocFragment { page-break-before: auto !important; }]],
-        },
-        {
-            id = "publisher_page-breaks_avoid ";
-            title = _("Avoid publisher page breaks"),
-            description = _("Disable all publisher page breaks, keeping only KOReader's epub.css ones.\nWhen combined with the two previous tweaks, all page-breaks are disabled."),
-            css = [[
+            },
+            {
+                id = "docfragment_page-break-before_avoid ";
+                title = _("Avoid blank page on chapter end"),
+                priority = 2, -- so it can override the one put back by publisher_page-break-before_avoid
+                css = [[DocFragment { page-break-before: auto !important; }]],
+            },
+            {
+                id = "publisher_page-breaks_avoid ";
+                title = _("Avoid publisher page breaks"),
+                description = _("Disable all publisher page breaks, keeping only KOReader's epub.css ones.\nWhen combined with the two previous tweaks, all page-breaks are disabled."),
+                css = [[
 * { page-break-before: auto !important; page-break-after: auto !important; }
 /* put back epub.css page-breaks */
 DocFragment { page-break-before: always !important; }
 h1 { -cr-only-if: -epub-document; page-break-before: always !important; }
 h2, h3 { -cr-only-if: legacy -epub-document; page-break-before: always !important; }
 h1, h2, h3, h4, h5, h6 { page-break-after: avoid !important; }
-            ]],
-            separator = true,
-        },
-        {
-            title = _("New page on headings"),
-            {
-                id = "h1_page-break-before_always";
-                title = _("New page on <H1>"),
-                css = [[h1 { page-break-before: always !important; }]],
+                ]],
+                separator = true,
             },
             {
-                id = "h2_page-break-before_always";
-                title = _("New page on <H2>"),
-                css = [[
+                title = _("New page on headings"),
+                {
+                    id = "h1_page-break-before_always";
+                    title = _("New page on <H1>"),
+                    css = [[h1 { page-break-before: always !important; }]],
+                },
+                {
+                    id = "h2_page-break-before_always";
+                    title = _("New page on <H2>"),
+                    css = [[
 h2 { page-break-before: always !important; }
 h1 + h2 { page-break-before: avoid !important; }
-                ]],
-            },
-            {
-                id = "h3_page-break-before_always";
-                title = _("New page on <H3>"),
-                css = [[
+                    ]],
+                },
+                {
+                    id = "h3_page-break-before_always";
+                    title = _("New page on <H3>"),
+                    css = [[
 h3 { page-break-before: always !important; }
 h1 + h3, h2 + h3 { page-break-before: avoid !important; }
-                ]],
-            },
-            {
-                id = "h4_page-break-before_always";
-                title = _("New page on <H4>"),
-                css = [[
+                    ]],
+                },
+                {
+                    id = "h4_page-break-before_always";
+                    title = _("New page on <H4>"),
+                    css = [[
 h4 { page-break-before: always !important; }
 h1 + h4, h2 + h4, h3 + h4 { page-break-before: avoid !important; }
-                ]],
-            },
-            {
-                id = "h5_page-break-before_always";
-                title = _("New page on <H5>"),
-                css = [[
+                    ]],
+                },
+                {
+                    id = "h5_page-break-before_always";
+                    title = _("New page on <H5>"),
+                    css = [[
 h5 { page-break-before: always !important; }
 h1 + h5, h2 + h5, h3 + h5, h4 + h5 { page-break-before: avoid !important; }
-                ]],
-            },
-            {
-                id = "h6_page-break-before_always";
-                title = _("New page on <H6>"),
-                css = [[
+                    ]],
+                },
+                {
+                    id = "h6_page-break-before_always";
+                    title = _("New page on <H6>"),
+                    css = [[
 h6 { page-break-before: always !important; }
 h1 + h6, h2 + h6, h3 + h6, h4 + h6, h5 + h6 { page-break-before: avoid !important; }
-                ]],
+                    ]],
+                },
             },
         },
     },
@@ -204,6 +208,7 @@ You may also want to enable, in the top menu → Gear → Navigation →, Invert
                 title = _("Document direction LTR"),
                 css = [[body { direction: ltr !important; }]],
             },
+            separator = true,
         },
         {
             title = _("Widows and orphans"),
@@ -279,43 +284,47 @@ h1, h2, h3, h4, h5, h6 { hyphens: none !important; }
             separator = true,
         },
         {
-            id = "sub_sup_smaller";
-            title = _("Smaller sub- and superscript"),
-            description = _("Prevent sub- and superscript from affecting line-height."),
-            priority = 5, -- so we can override "font_size_all_inherit"
-            -- https://friendsofepub.github.io/eBookTricks/
-            -- https://github.com/koreader/koreader/issues/3923#issuecomment-386510294
-            css = [[
+            title = _("Fonts and line heights"),
+            {
+                id = "font_family_all_inherit";
+                title = _("Ignore publisher font families"),
+                description = _("Disable font-family specified in embedded styles."),
+                -- we have to use this trick, font-family handling by crengine is a bit complex
+                css = [[* { font-family: "NoSuchFont" !important; }]],
+            },
+            {
+                id = "font_size_all_inherit";
+                title = _("Ignore publisher font sizes"),
+                description = _("Disable font-size specified in embedded styles."),
+                css = [[* { font-size: inherit !important; }]],
+                separator = true,
+            },
+            {
+                id = "lineheight_all_inherit";
+                title = _("Ignore publisher line heights"),
+                description = _("Disable line-height specified in embedded styles, and may allow KOReader's line spacing settings to work on books where they would not."),
+                css = [[* { line-height: inherit !important; }]],
+            },
+            {
+                id = "lineheight_all_normal_strut_confined";
+                title = _("Enforce steady line heights"),
+                description = _("Prevent inline content like sub- and superscript from changing their paragraph line height."),
+                priority = -5, -- so other -cr-hint can override (this one has effect only on inline content)
+                css = [[* { -cr-hint: strut-confined; }]],
+                separator = true,
+            },
+            {
+                id = "sub_sup_smaller";
+                title = _("Smaller sub- and superscript"),
+                description = _("Prevent sub- and superscript from affecting line-height."),
+                priority = 5, -- so we can override "font_size_all_inherit"
+                -- https://friendsofepub.github.io/eBookTricks/
+                -- https://github.com/koreader/koreader/issues/3923#issuecomment-386510294
+                css = [[
 sup { font-size: 50% !important; vertical-align: super !important; }
 sub { font-size: 50% !important; vertical-align: sub !important; }
-            ]],
-        },
-        {
-            id = "lineheight_all_inherit";
-            title = _("Ignore publisher line heights"),
-            description = _("Disable line-height specified in embedded styles, and may allow KOReader's line spacing settings to work on books where they would not."),
-            css = [[* { line-height: inherit !important; }]],
-        },
-        {
-            id = "lineheight_all_normal_strut_confined";
-            title = _("Enforce steady line heights"),
-            description = _("Prevent inline content like sub- and superscript from changing their paragraph line height."),
-            priority = -5, -- so other -cr-hint can override (this one has effect only on inline content)
-            css = [[* { -cr-hint: strut-confined; }]],
-            separator = true,
-        },
-        {
-            id = "font_size_all_inherit";
-            title = _("Ignore publisher font sizes"),
-            description = _("Disable font-size specified in embedded styles."),
-            css = [[* { font-size: inherit !important; }]],
-        },
-        {
-            id = "font_family_all_inherit";
-            title = _("Ignore publisher font families"),
-            description = _("Disable font-family specified in embedded styles."),
-            -- we have to use this trick, font-family handling by crengine is a bit complex
-            css = [[* { font-family: "NoSuchFont" !important; }]],
+                ]],
+            },
         },
     },
     {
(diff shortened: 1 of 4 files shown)


frontend/ui/data/css_tweaks.lua
```diff
@@ -338,38 +347,89 @@ p {
...
-            title = _("No indentation on first paragraph line"),
-            description = _("Do not indent the first line of paragraphs."),
-            css = [[p { text-indent: 0 !important; }]],
-        },
-        {
-            id = "paragraph_indent";
-            title = _("Indentation on first paragraph line"),
-            description = _("Indentation on the first line of a paragraph is the default, but it may be overridden by publisher styles. This will force KOReader's defaults on common elements."),
-            css = [[
+            title = _("Paragraphs first line indentation"),
```
Frenzie:
```suggestion
            title = _("Paragraph first-line indentation"),
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -338,38 +347,89 @@ p {
...
+                priority = 5,
+                css = [[p, li { margin-top: 0 !important; margin-bottom: 0 !important; }]],
+            },
         },
         {
-            id = "paragraph_no_whitespace";
-            title = _("No spacing between paragraphs"),
-            description = _("No whitespace between paragraphs is the default, but it may be overridden by publisher styles. This will re-enable it for paragraphs and list items."),
-            css = [[p, li { margin-top: 0 !important; margin-bottom: 0 !important; }]],
+            title = _("Paragraphs margins and paddings"),
```
Frenzie:
```suggestion
            title = _("Paragraph margin and padding"),
```
```suggestion
            title = _("Paragraph margins and paddings"),
```
```suggestion
            title = _("Paragraph spacing"),
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -338,38 +347,89 @@ p {
...
+            {
+                id = "margin_horizontal_most_0";
+                title = _("Ignore paragraphs horizontal margins"),
```
Frenzie: This collection sounds a bit awkward, mainly because of the adjective order.

```suggestion
                title = _("Ignore horizontal paragraph margins"),
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -338,38 +347,89 @@ p {
...
+                priority = 3, -- Override "Pages > Ignore margin and padding"
+                css = [[p, li { margin-left: 0 !important; margin-right: 0 !important; }]],
+            },
+            {
+                id = "margin_vertical_most_0";
+                title = _("Ignore paragraphs vertical margins"),
```
Frenzie:
```suggestion
                title = _("Ignore vertical paragraph margins"),
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -338,38 +347,89 @@ p {
...
+                priority = 3,
+                css = [[p, li { margin-top: 0 !important; margin-bottom: 0 !important; }]],
+                separator = true,
+            },
+            {
+                id = "padding_horizontal_most_0";
+                title = _("Ignore paragraphs horizontal padding"),
```
Frenzie:
```suggestion
                title = _("Ignore horizontal paragraph padding"),
```

frontend/ui/data/css_tweaks.lua
```diff
@@ -338,38 +347,89 @@ p {
...
+                priority = 3,
+                css = [[p, li { padding-left: 0 !important; padding-right: 0 !important; }]],
+            },
+            {
+                id = "padding_vertical_most_0";
+                title = _("Ignore paragraphs vertical padding"),
```
Frenzie:
```suggestion
                title = _("Ignore vertical paragraph padding"),
```

Frenzie:
@poire-z Btw did you want to implement https://github.com/koreader/koreader/issues/5859#issuecomment-585942116 or shall I? I basically just want to add those same minor adjustments to the readerdictionary styles, like this:

```css
blockquote { margin: 0 1em; } /* or maybe 0 0 0 1em */
dd { margin: 0 1em; }  /* or maybe 0 0 0 1em */
```

poire-z:
> did you want to implement [CSS tweaks for HTML dict]

Not really, please go at it (and I have no complex dict to check and decide which is best :)

poire-z:
About `Paragraph margin and padding`, I can't use your suggested "Paragraph spacing" because there is already one, that contains the 3 tweaks you added in the past:

<kbd>image</kbd>

The 3 sub-menus in that order:
<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

But actually, "No spacing between paragraphs" has the same CSS content as "Ignore paragraphs vertical margins".
But "Spacing between paragraphs" is more for esthetics or personal preferences, while the ones I added in "Ignore paragraphs vertical margins" is more as a help to kill publisher errors or really bad choices.
So, I dunno, may be I could move "Ignore vertical paragraphs padding" into the "Spacing between paragraphs", as it's also about vertical spacing - and that avoids the duplication.
And move the 2 Horizontal padding/margins in another submenu, as they are less often needed (but good to have available). Either in the main paragraph menu, or as a submenu inside "First line indentation" (as it's about horizontal stuff, but it's not indentation as advertized by the menu name).

Any thoughts?

Frenzie:
It sounds the most logical to put vertical margin/padding in the existing spacing menu, and horizontal margin/padding as a submenu in the main paragraph menu.

poire-z:
Simplified the TouchMenu per_page handling (we didn't really need a stack... updating a single place was enough :)

As for the paragraphs tweaks, reorganized and avoiding the duplicate, that would give:
<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

Frenzie:
@poire-z Did you still want to change anything or ready to go?

poire-z:
Nope, ready to go (was just waiting a bit after release in case we needed some hot fix without any new stuff - or for another first commit after release, that you just did :)

poire-z:
I'd like to add a style tweak like this one:
image
I think it might help with the kind of book that sets a font size and needs us to need to change our default font size value to try to approximate what we are used to (@zwim has met this kind of book, https://github.com/koreader/koreader/pull/9159#issuecomment-1146414802), with usually a feeling of "something's still not as I'm used to" :)
(But I'm not yet super sure it's efficient - it looks alright, I may have use somethig like it when fighting books, but somehow I don't think about it everytime, so not much experience yet.)
It's less radical than the "Ignore publisher font size", which would also reset the font size of heading/titles which makes everything becomes flat :)

Any rewording suggestion, for this one and/or the existing one ?

Also, I'm a bit bothered getting now 6 items in this menu (although 6 is still fine, but 5 is best for such things we want to see more of its effect without closing the menu).
Thoughts about reorganizing all that ?
Or moving `Smaller sub- and superscript` elsewhere ? (I rarely use it, it came from before `Enforce steady line-height` which is better at most things. But I'm not sure I want to remove existing ones, people may be using them.)

It wouldn't help with making it smaller, but I realize there are a few things that may be badly organized:
- `Widows and orphans` is in `Text>` , while I think it's rather about `Paragraphs`, or paragraphs in the context of `Pages`
- `Paragraphs>` is a bit crowded (with `List items>` which is somehow still a bit related)
- `Horizontal paragraph margins>` could go into `Pages>` if we rename it `Pages and margins>`
- could `In-page footnotes>` have their place into... `Pages>` / `Pages and margins>` ?

Any thoughts ?

NiLuJe:
* Yes please on the new tweak ;).
* Moving sub/super-script elsewhere sounds good (where to, though :D).
* Windows & orphans probably makes more sense in the context  of Pages, yeah.
* Pages and margins sounds good.
* Moving footnotes makes some sense, but might not necessarily be needed?

Frenzie:
Nothing much to add to that.

Moving the footnotes would probably be fine but I don't think it's necessary.

Maybe a few more elements? Such as `div` and `blockquote`? And what about line-height: initial or 1.5 or something?

zwim:
I am not quite sure, there are ignore publisher entries scattered around in the whole style tweaks menu.  They are sorted into sane places. But the question is: Does the user recognise if the padding happens in a table environment or anywhere else? Is it a font or a line hight problem...

Most books get rendered very well (I like it!). But  if I open a misbehaving one, where do I look first? -> In the "Style tweak - Ignore publisher menu". (Because I trust KOReader ). Hhmm there is no such menu entry, so let's go through the numerous submenus until I find the right one (maybe by trial an error).

So shouldn't we discuss to bundle all the "Ignore vendors" in one submenu?
(I am aware that the style tweak top menu us already crowded, but eventually we can merge some if the ignore entries go away.)

poire-z:
> And what about line-height: initial or 1.5 or something?

That's best left to the "Ignore publisher line-height" - and we can then use our bottom menu widget to tweak it (rather than forcing 1.5 or needing other values).

> But the question is: Does the user recognise if the padding happens in a table environment or anywhere else? Is it a font or a line hight problem...
Most books get rendered very well (I like it!). But if I open a misbehaving one, where do I look first? -> In the "Style tweak - Ignore publisher menu". (Because I trust KOReader ). Hhmm there is no such menu entry, so let's go through the numerous submenus until I find the right one (maybe by trial an error).
So shouldn't we discuss to bundle all the "Ignore vendors" in one submenu?

That could be a way to lay out the menu, but I don't think it would really be better:
- There are `Style tweaks`, not just `Style killers` :) We propose a few other stuff than just killing and restoring, and depending on some, there is no real killing, just overwritting individual bits (ie paragraphs first line indendation and spacing, there are multiple combinations possible before killing everything and getting the normal epub.css rendering. Also, some tweaks are named "Ignore publisher xyz" possibly arbitrarily, but could as well be "Force xyz to be abc" - and I'm sure we have some of these, depending on how the name and description would sound.
- if you would move all the ones that "Ignore publisher something" into its own section, you would have > 15 items that you would need to put into subsections to get 5 items per page and see part of the book content - or you would have a 3 pages menu with many things not really organized. And if subsections for Ignore tweaks, you would duplicate the sections in the other menu for non-Ignore publisher stuff. And related things would be scattered around.
- I trust a user to at least know if what he wants to fight is text/fonts, paragraphs margins, or tables :) and go to the right place to see what's offered. It might be technical HTML/CSS stuff, but the casual name and description we give should be simple enough - and may educate the curious user.
- Otherwise, we may have a "Enable some random Ignore tweaks" button, that the user will hit and we'll pick a random combination of Ignore tweaks :) and at some point (after 137 re-renderings), he may get lucky and have its issues solved :)
- And yes, for some issues, it's trial and error, even for me. Especially everything margins: that's why we have many such tweaks, that may make more sense once in `Pages and margins`: killing Page margins is obvious, but may not be enough. Killing paragraph margins may not work if the book use DIV or BLOCKQUOTE. Killing horizontal margins may kill too much (and you lose the margin/indentation that may be used for quotes/blockquotes). And sometimes none of these will work because the publisher uses padding, not margin... So, we have Kill horizontal or paragraph margin and kill paddings :)
The distinction between what is padding and margin is indeed technical CSS and may not talk to some users. But I think it's still better that using `Ignore paragraph margins (try this first)` and `Ignore paragraph margins (try that one if the other fails)`

tldr: Style tweaks is a zoo showing the diversity of CSS/HTML layout that one should enjoy and learn. It's not a mafia restaurant where you go to hire killers, that will miss their target half of the time :) You often end up needing to do the dirty work yourself :)

poire-z:
About the idea of moving Ignore horizontal paragraph margins from `Paragraphs>` to `Page and margins>`, to have all the margins arsenal at one place:
image
image (a bit so-so though)

it would add another kind of confusion, as some of the paragraphs vertical margins is already (rightly I think) located in `Paragraphs> Spacing between paragraphs>`
image

Unless we consider horizontal margins an annoyance, and vertical margins a feature/style choice of paragraphs.

Or instead of regrouping like on the second screenshots (section for h/v margins, section for h/v padding - the distinction between padding and margin being a bit obscure), we go with grouping by horizontal:
`Ignore all vertical margins` (which will include paddings) as a radical killer for vertical spacing
`----------------`
`Ignore all horizontal margins` (which will include paddings)
`Ignore horizontal containers margins` (which will include paddings) (for DIV, just in case? and/or "sub containers" like BLOCKQUOTE ?)
`Ignore horizontal paragraphs margins` (which will include paddings) (for P, LI)

Although sometimes, it's nice to have that padding/margin granularity as it happens using one or the other kill the unwanted margin while leaving some other wanted ones, but it's mostly luck :)

But such changes would need us ro rename the tweaks' ids and may change some current users layouts...
So, may be we should keep that.
And just keep them and put them in a horizontal stuff section and in a vertical stuff section.

I could also duplicate (same twead id and css) some of the paragraph spacing stuff into Ignore margin stuff...

Well, no real obviously less confusing ideas for now...

Also, a precision to my previous post: Most of the `Ignore publisher xyz` are actually doing `set xyz to 0` (which is ok with our epub.css where we like using 0 - but html5.css does specify some margins/padding, that will be killed by these tweaks).
And `Ignore publisher font sizes` is actually a lie: it sets all font size (publisher or ours) to normal, and it overrides our sizes we set in epub.css for H1 H2 H3...

Frenzie:
> Most of the Ignore publisher xyz are actually doing set xyz to 0

What happens right now if you use `initial`?

poire-z:
I have reorganized a bit our style tweaks, making the main menu and others smaller:
image
Too many stuff moved around or into submenus to make screenshots, so I'm giving the file here:
[css_tweaks.lua.txt](https://github.com/koreader/koreader/files/8928024/css_tweaks.lua.txt)
Please replace your `frontend/ui/data/css_tweaks.lua` with this file and browse around and tell me if all make sense.

Changes:
- moved margins stuff into Page and margins, splitting horizontal from vertical (as we know if what we're fighting is space on the left or right, or between blocks)
- moved In-page footnotes inside Page and margins, as it really feels about Pages :) as much as Widows and orphans - but tell me is this is too much a change and people who are used to find it in Misc> may panic :) (I guess it's one of the most often accessed):
image
- Splitted Font size and Line heights into 2 submenus - so more room to see effects and eventually add more tweaks. Move `Smaller sub/super scripts` into `Line heights>` (feels at home there).
- Thought about duplicating (as mentionned in previous posts) paragraph vertical spacing into Page and margins, but there's a small issue that we would see `Style tweaks (1)` (as it's the same tweak id), but when opening the menu, we would see `Pages and margins (1)` and `Paragraphs (1)` , and `1+1 = 1` is something that bothers me :). So, I made the paragraph one targets only `p {}` and the `Page and margins> Ignore paragraphs margins` target `p, li {}`.
- Move `List items> Force bullets/decimal` into `Miscellaenous>` as it doesn't really need to be in `Paragraphs>`
- Moved Tables, links, images into their own submenu, as rarely used

poire-z:
> moved In-page footnotes inside Page and margins

Minor personal issue with this: previously, I saw `Miscellaneous (1)>` and `Pages (1)>` which was a quick hint for me that epub in-page foonotes was enabled, and that I had probably enabled `Ignore publisher page margins`.
Now, I see `Page and margins (2)>` and the hint is less clear :)

Another question, regarding in-page footnotes.
I often get this kind of footnote rendering, with text-indent:
image
that I always end up fighting with a book style tweak, to get:
image

Is this only my personal taste ?  Or is it obvious :) the second one is better (the number being enough to see a new footnote start, and easier to find when not indented) ?

Another issue I have with the `Reset main text font side` I want to add (and added in the css_tweaks.lua in previous post):
`body, p, li, div, blockquote { font-size: 1rem !important; }`
is that it would override the smallness of `In-page EPUB footnotes (smaller)`
`*[type~="footnote"] { -cr-hint: footnote-inpage; ... font-size: 0.8rem !important;}`
when we have the type=footnote set on a `<aside>` with a `<p>` inside that contains the text:
```html
<aside class="ntb" epub:type="footnote" id="ftn102">
  <p class="note"><a class="notesymbol" href="chapitre13.xhtml#bodyftn102">102</a> Some stuff.</p>
</aside>
```
This construct is very common in French books.

I can't/don't want to add to the existing tweak:
```css
*[type~="note"] p,
*[type~="endnote"] p,
*[type~="footnote"] p,
*[type~="rearnote"] p,
*[role~="doc-note"] p,
*[role~="doc-endnote"] p,
*[role~="doc-footnote"] p,
*[role~="doc-rearnote"] p
{
        font-size: 0.8rem !important;
}
```
as it feels these would make the CSS checks&apply a lot slower on books (checking all ancestors of all P)...
I may end up just adding:
`aside[type~="footnote"] > p { font-size: 0.8rem !important; }`
(I have only 2 mobileread-user-made books that use `<aside type="rearnote">` that I'm ok ignoring.)

Thoughts?

mergen3107:
> Is this only my personal taste ?

I too think unindented looks better. Also it saves space, so more useful notes can fit.

Frenzie:
> And, to counteract any nested blocks (or DL/DT/DD like I've met recently) that would put the footnote number on a paragraph, and the text in another, and make everything smoothly inline

Do you mean they were excessively large? Having a bit of a list or a couple of paragraphs or something sounds a-okay to me and forcing them to be inline like a net negative.

> I would use "cr_hinted" as the name for this internal property - but any thought about a better name for it, that would be sufficiently different from "cr_hint" ? I found "clue" or "cue" as small synonyms, but I'm not really used to these words (or they don't sound as clear as "hint" to me). Or "colored: inside-inpage-footnote" or "scent: block" :) Thoughts?

I'm not sure if I quite understand the "hinted" phrasing. Isn't it more of a context?

poire-z:
> Do you mean they were excessively large?

No, it was a bunch of:
```html
<body>
<dl class="notes">
<dt id="footnote-102" class="calibre5">[<a href="index_split_018.xhtml#citation-102" class="calibre6">102</a>
]</dt>
<dd class="calibre7" id="calibre_pb_131"><p class="p-p53">blah blah</p></dd>
</dl>
</body>
```
which where rendered (as dt and dd are block elements) as:
image


We can also get that kind of stuff with other contexts, like a bunch of DIV (when the publisher decides to make a new page for each footnote).

> Having a bit of a list or a couple of paragraphs or something sounds a-okay to me and forcing them to be inline like a net negative.

It would only be an option. Fighting the above dt dl ended me having to type/test/use:
```css
.notes { -cr-hint: footnote-inpage; margin: 0 !important;}
.notes > dt { display: inline !important; font-size: 1em !important}
.notes > dd, .notes > dd > p { display: inline !important; font-size: 1em !important}
```
With the above solution, I would just enable these 2 tweaks, and they would work.
Of course, they would break footnotes that have multiple paragraphs by making all of them inline - but that's my decision (no forcing :) it's just an available tweak, to use when needed) - I could initially have a quick browsing in scroll mode of all the footnotes to see if they have multiple paragraphs and blockquotes and such - and what loss I would get.

> I'm not sure if I quite understand the "hinted" phrasing. Isn't it more of a context?

Yep, `cr_context` would work ! and would not sound like anything existing in the CSS domain. I will probably go with that, thanks :)

Frenzie:
> It would only be an option. Fighting the above dt dl ended me having to type/test/use:

Makes sense.  :-)

poire-z:
About my above idea of using:
```css
* {
  -cr-only-if: inside-inpage-footnote block;
    display: inline !important;
}
```
I had some obvious issue with CSS specificity: `*` being super low specificity, it would be checked/applied very early, before any other `p[type=footnote]` or `p.footnote` get a chance to actually make the node "inpage-footnote".

So, I had to add some way to increase the specificity of some selectors.
I could have gone with so at-rule like `@cr-late { * { .. } }`, but that's a bit complicated to implement, and for users to add/remove/temporarily disable.
(CSS specs has got recently something for that purpose, `@layer` `https://github.com/w3c/csswg-drafts/issues/4470` but it's even more complicated to grasp.)

So, I'll go with just `-cr-hint: late` : add this in a declaration, and all the selectors gets a specificy huge bump. Make a typo in `late` to just disable it.

poire-z:
So, I'm reworking the "In-page footnotes" submenu.
With the idea that we enable some footnote type, and we tweak how they should look with other menu items.
It's a bit more logical and practical, and it allows tuning / taming down publisher footnotes look, including the one we would need to enable via a book style tweak when none of the 3 existing enabling tweak work.
But then, if  we remove "EPUB footnote (smaller)" (that we currently set as global default), because it's redundant, either the user will get footnote with the publisher text size (which can be normal or smaller), which might look not as nice as our current default.
Or we have to set as global "Footnote font size 80%", but then, we force all footnotes to be 80%, and users won't see the publisher initial size choices...
(And if we force as global default a font size of 80%, we also killing the FB2 footnotes 0.75rem / 0.85rem distinction for footnotes vs endnotes...)

So, I have a super toolkit menu to tame down footnotes, but no real nice first time experience for a normal user.
The experience is ok for me and my workflow: I open a new book, I see what it has: footnotes with the normal publisher text size - and then, I tweak/tame down everything.
So, I'm a bit stuck on how to go with that :/

----
My updated "Inpage-footnotes" menu would be:
<h3>In-page EPUB footnotes</h3>
<p>Show EPUB footnote text at the bottom of pages that contain links to them.
This only works with footnotes that have specific attributes set by the publisher.</p>
<pre style='background-color: #cccccc'>*[type~=&quot;note&quot;],
*[type~=&quot;endnote&quot;],
*[type~=&quot;footnote&quot;],
*[type~=&quot;rearnote&quot;],
*[role~=&quot;doc-note&quot;],
*[role~=&quot;doc-endnote&quot;],
*[role~=&quot;doc-footnote&quot;],
*[role~=&quot;doc-rearnote&quot;]
{
    -cr-only-if: -fb2-document;
        -cr-hint: footnote-inpage;
        margin: 0 !important;
}</pre>
<h3>In-page Wikipedia footnotes</h3>
<p>Show footnotes at the bottom of pages in Wikipedia EPUBs.</p>
<pre style='background-color: #cccccc'>ol.references &gt; li {
    -cr-hint: footnote-inpage;
    list-style-position: -cr-outside;
    margin: 0 !important;
}
&#47;* hide backlinks *&#47;
ol.references &gt; li &gt; .noprint { display: none; }
ol.references &gt; li &gt; .mw-cite-backlink { display: none; }</pre>
<h3>In-page classic classname footnotes</h3>
<p>Show footnotes with classic classnames at the bottom of pages.
This tweak can be duplicated as a user style tweak when books contain footnotes wrapped with other class names.</p>
<pre style='background-color: #cccccc'>.footnote, .footnotes, .fn,
.note, .note1, .note2, .note3,
.ntb, .ntb-txt, .ntb-txt-j,
.przypis, .przypis1, &#47;* Polish footnotes *&#47;
.voetnoten &#47;* Dutch footnotes *&#47;
{
    -cr-hint: footnote-inpage;
    margin: 0 !important;
}</pre>
<h2>In-page footnote font size</h2>
<h3>Footnote font size: 100 %</h3>
<pre style='background-color: #cccccc'>*, autoBoxing {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        font-size: 1rem !important;
}</pre>
<h3>Footnote font size: 90 %</h3>
<pre style='background-color: #cccccc'>*, autoBoxing {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        font-size: 0.9rem !important;
}</pre>
<h3>Footnote font size: 85 %</h3>
<pre style='background-color: #cccccc'>*, autoBoxing {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        font-size: 0.85rem !important;
}</pre>
<h3>Footnote font size: 80 %</h3>
<pre style='background-color: #cccccc'>*, autoBoxing {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        font-size: 0.8rem !important;
}</pre>
<h3>Footnote font size: 75 %</h3>
<pre style='background-color: #cccccc'>*, autoBoxing {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        font-size: 0.75rem !important;
}</pre>
<h3>Footnote font size: 70 %</h3>
<pre style='background-color: #cccccc'>*, autoBoxing {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        font-size: 0.7rem !important;
}</pre>
<h3>Footnote font size: 65 %</h3>
<pre style='background-color: #cccccc'>*, autoBoxing {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        font-size: 0.65rem !important;
}</pre>
<h2>In-page footnote cleanup</h2>
<h3>No footnote indentation</h3>
<pre style='background-color: #cccccc'>* {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote;
        text-indent: 0 !important;
}</pre>
<h3>Justify footnote text</h3>
<pre style='background-color: #cccccc'>* {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote;
        text-align: justify !important;
}</pre>
<h3>Remove block margins</h3>
<p>Remove block margin and padding inside inpage footnotes.</p>
<pre style='background-color: #cccccc'>* {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inline;
        margin: 0 !important;
        padding: 0 !important;
}</pre>
<h3>Remove cosmetic spacing</h3>
<p>Remove any margin and padding inside inpage footnotes, including between inline elements.</p>
<pre style='background-color: #cccccc'>* {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote;
        margin: 0 !important;
        padding: 0 !important;
}</pre>
<h3>Inline all footnote content</h3>
<p>If the footnote number is on a single line and the footnote text on the next line, or if there are variable levels of indentation, this tweak can make all the footnote content become a single paragraph.
This will break any complex footnote containing quotes or lists.</p>
<pre style='background-color: #cccccc'>* {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inpage-footnote -inline;
        display: inline !important;
}
* {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote;
        list-style-position: inside;
}</pre>
<h3>Regularize text size on inline elements</h3>
<p>If the footnote text uses variable or absolute font sizes, line height or vertical alignments, which would make it too irregular, you can reset all of them to get a leaner text (to the expense of loosing superscripts).</p>
<pre style='background-color: #cccccc'>* {
    -cr-hint: late;
    -cr-only-if: inside-inpage-footnote -inpage-footnote;
        font-size: inherit !important;
        line-height: inherit !important;
        vertical-align: inherit !important;
}</pre>

(Generated with https://github.com/koreader/koreader/issues/9261#issuecomment-1167361313 - that I allowed myself to fix :)

Frenzie:
Pardon, what fix? I thought there were no bugs (in the final edit; the initial version was missing escaping). :-P

poire-z:
See the github "Edited by" diff: you were iterating twice, they were duplicated stuff, and some missing ones.

Frenzie:
Ah, I wouldn't have been able to spot any difference without the diff. In my defense, I wrote it in 5-10 minutes. ;-D

poire-z:
Or we should keep the 3 `... (smaller)` ones , which would avoid any migration or loss of current user tweaks and habits
image

It just makes/keeps this menu quite crowded, with a bit of redundancy with the font size - although it could work: the `(smaller)` sets the footnote outer container font size (which may have no effect depending on publisher styles), and the other new options can help forcing the font size on inner containers when needed.
The last 2 menus I would be adding could even go into a submenu if I want to save 1 line.

poire-z:
Mostly unrelated:
Anything against adding swipe south in our top menu to "go back/up", same as the top pointing chevron in the bottom left corner of the menu ? (because, when navigating the menu with one hand and the device in the right hand, reaching this button on the left is tedious, and one needs his left hand).
```diff
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -815,16 +815,18 @@ end
 function TouchMenu:onSwipe(arg, ges_ev)
     local direction = BD.flipDirectionIfMirroredUILayout(ges_ev.direction)
     if direction == "west" then
         self:onNextPage()
     elseif direction == "east" then
         self:onPrevPage()
     elseif direction == "north" then
         self:closeMenu()
+    elseif direction == "south" then
+        self:onBack()
     end
 end
```

zwim:
Maybe a stupid question. The chevron points up (north ) but the swipe should go down (south )?
Isn't that confusing?

poire-z:
> The chevron points up (north ) but the swipe should go down (south )?
> Isn't that confusing?

Yep, a bit - but people may already get used to close it all with swipe north - and only swipe south is yet unused :)
Unless we just make swipe north "go back", and people used to that will just need to use it multiple times (when go back reach top, next "go back" closes) - and let them discover that swipe south is a shortcut :)

Frenzie:
> Unless we just make swipe north "go back"

Closing is significantly more important than going back. Even if you don't want to close the menu per se, it's usually significantly faster to start from scratch unless you specifically want to go just one back (which is what the arrow is for). I very explicitly decided against north to go back.

As to using south to go back, seems a little weird but probably fine?

mergen3107:
> Or we should keep the 3 `... (smaller)` ones , which would avoid any migration or loss of current user tweaks and habits image
> 
> It just makes/keeps this menu quite crowded, with a bit of redundancy with the font size - although it could work: the `(smaller)` sets the footnote outer container font size (which may have no effect depending on publisher styles), and the other new options can help forcing the font size on inner containers when needed. The last 2 menus I would be adding could even go into a submenu if I want to save 1 line.

PR #5865, opened 2020-02-14
Make dictionary margin depend on font size

Frenzie:
The default left margin on blockquote and dd is `40px`. There are two potential problems with this:

1. 40px is a lot on a small font and perhaps rather little on a large font. Conversely, *if* blockquote font size were variable you'd get strange differences in indentation. `rem` should be a better choice.
2. <del>We don't interact with MuPDF quite as I believe we should (cf., the way I implemented this stuff with layoutDocument for opening EPUB/FB2 with MuPDF). While effectively the same problem, the distinction is that point (1) is real but nevertheless largely academic, while this point constitutes a very real problem both on DPIs over ~300 and under ~200.</del><ins>Actually there may be some MuPDF issue involved.</ins>

For problem (2) this may be a bit of a quick & lazy fix, but problem (1) exists regardless.

PS If we assume 1em = 16px then 40px is actually 2.5em, not 1em. This reduction in indentation is mainly intended for the benefit of smaller E Ink devices.

Closes <https://github.com/koreader/koreader/issues/5859>.

### Before/after medium DPI (300)
 

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5865)

### DIFF
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -191,14 +191,18 @@ function DictQuickLookup:getHtmlDictionaryCss()
             ]]..css_justify..[[
         }
 
-        blockquote {
-            margin-left: 1em;
-            margin-right: .5em;
-        }
-
-        dd {
-            margin-left: 1em;
-            margin-right: 0;
+        /* MuPDF doesn't currently scale CSS pixels, so we have to use a font-size based measurement.
+         * Unfortunately MuPDF doesn't properly support `rem` either, which it bases on a hard-coded
+         * value of `16px`, so we have to go with `em` (or `%`).
+         *
+         * These `em`-based margins can vary slightly, but it's the best available compromise.
+         *
+         * We also keep left and right margin the same so it also displays as expected in RTL.
+         * Because MuPDF doesn't currently support `margin-start`, this results in a slightly
+         * unconventional but hopefully barely noticeable right margin for <dd>.
+         */
+        blockquote, dd {
+            margin: 0 1em;
         }
     ]]
 
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -189,18 +189,19 @@ function DictQuickLookup:getHtmlDictionaryCss()
             margin: 0;
             line-height: 1.3;
             ]]..css_justify..[[
-        }
-
-        /* MuPDF doesn't currently scale CSS pixels, so we have to use a font-size based measurement.
-         * Unfortunately MuPDF doesn't properly support `rem` either, which it bases on a hard-coded
-         * value of `16px`, so we have to go with `em` (or `%`).
-         *
-         * These `em`-based margins can vary slightly, but it's the best available compromise.
-         *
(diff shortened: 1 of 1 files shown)


poire-z:
I'm fine with the values.
For the footnotes popup, I went with margin-left = margin-right in https://github.com/koreader/koreader/pull/5863/commits/bd574b3e32cd2c3888f2d3233e5f46ef0534907a (but ok, not for dd :) I guess I should?)
Because I think (may be I checked, don't remember) MuPDF has no support for margin-start, and does (rightly) not mirror them for RTL input - so having them similar should gives probably same good output on RTL dicts/footnotes - while having them unbalanced will make arabic HTML dicts output ugly (but I dunno if there's a single really fancy arabic stardict HTML dict in existence :)

poire-z:
lgtm.
(Not insisting much on that, as I put inline but less giant comments for footnote popup CSS too :) - but may be that giant comment could be put outside the CSS (a few lines below) as a Lua comment to skip that from being parsed by MuPDF - and less output if we ever debug what's given to HTMLBoxWidget.)

Frenzie:
@poire-z You mean something like this?

```diff
diff --git a/frontend/ui/widget/dictquicklookup.lua b/frontend/ui/widget/dictquicklookup.lua
index 1e36c143..7e9b1396 100644
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -189,18 +189,19 @@ function DictQuickLookup:getHtmlDictionaryCss()
             margin: 0;
             line-height: 1.3;
             ]]..css_justify..[[
-        }
-
-        /* MuPDF doesn't currently scale CSS pixels, so we have to use a font-size based measurement.
-         * Unfortunately MuPDF doesn't properly support `rem` either, which it bases on a hard-coded
-         * value of `16px`, so we have to go with `em` (or `%`).
-         *
-         * These `em`-based margins can vary slightly, but it's the best available compromise.
-         *
-         * We also keep left and right margin the same so it also displays as expected in RTL.
-         * Because MuPDF doesn't currently support `margin-start`, this results in a slightly
-         * unconventional but hopefully barely noticeable right margin for <dd>.
-         */
+        }]]
+        ..
+        -- MuPDF doesn't currently scale CSS pixels, so we have to use a font-size based measurement.
+        -- Unfortunately MuPDF doesn't properly support `rem` either, which it bases on a hard-coded
+        -- value of `16px`, so we have to go with `em` (or `%`).
+        --
+        -- These `em`-based margins can vary slightly, but it's the best available compromise.
+        --
+        -- We also keep left and right margin the same so it also displays as expected in RTL.
+        -- Because MuPDF doesn't currently support `margin-start`, this results in a slightly
+        -- unconventional but hopefully barely noticeable right margin for <dd>.
+        --
+        [[
         blockquote, dd {
             margin: 0 1em;
         }
```

poire-z:
I would just have moved the comment after the CSS (as the CSS snippet ends with the `blockquote {} `that this comment most likely applies to, shouldn't be too far away:
```lua
    local css = [[
        @page {
        [...]
        blockquote, dd {
            margin: 0 1em;
        }
    ]]
        -- MuPDF doesn't currently scale CSS pixels...
```

PR #5867, opened 2020-02-15
Statistics: revamp settings, add calendar settings

poire-z:
Details and discussion in https://github.com/koreader/koreader/pull/5854#issuecomment-586383563.

Also adds options to remove statistics for books with small reading durations (to clean up stats for
books just quickly browsed and not yet read, or small articles - been using one with 30mn for months).

(Also adds TextWidget:getFontSizeToFitHeight() even if I ended up not using it, might be useful in other contexts.)

Current wording and defaults:
<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

Allows tuning Calendar view to each's own liking or eyes strength:

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>
(that last one is ugly there but ok on a device)

(bit tedious to get the font size right, and so is the code for that... but checked on the emulator and my Kobo, hope it's also fine on other devices/dpi and that I'm done with it :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5867)


poire-z:
> Set min and max duration (in seconds) _for a page_ to be counted as read in statistics.

Tweaked to:
Set min and max duration (in seconds) _on a page for it_ to be counted as read in statistics.

Frenzie:
"Set duration on a page" sounds a bit odd to me. You can set a duration on a timer or something (although you'd probably just "set the timer," not the "timer duration").

How about:

* Set min and max seconds to count a page as read in statistics.
* Set min and max seconds for counting a page as read in statistics.

poire-z:
OK, my bad (english) :)
My original and your 2 last proposals don't establish a clear link between the duration/delay (of what?) and the page being counted (what's counted?).
I'd like to make it clearer that it is "how long you stay on a page before leaving it/going to an other" that must be between min and max, and that is counted (as one page read, and that adds to the total reading duration).

Frenzie:
I'd stick with my original suggestion for that. Grammatical analysis something like this:

[ [ Set min and max duration (in seconds)]  [ [ for a page to be counted as read ] in statistics. ] ]

set min & max duration
why?
for a page to be counted as read

poire-z:
I've put back your original suggestion, for grammar sake.
> set min & max duration
why?
for a page to be counted as read

But this, to me and my limited english, does not tell me the duration of what - but may be it's implicitely understood by native english readers that the duration = the time you stay on the page.

Frenzie:
Set min & max time spent on a page for it to be counted as read?

poire-z:
Perfect! That's clearer and reword "duration" that's already in the title.

<kbd>image</kbd>

gerhaher:
Settings doesn't stick: When I change 'Books per calendar day' to 5 and enable 'Allow browsing...' and then open a book and then check Reading statistics again the changes are gone.

'Reset stats for books read for <....' doesn't work. I got "No statistics removed" all the time.

poire-z:
Can't reproduce: they stick for me.
Do you see them updated in the menu, after you change them?
_When I change 'Books per calendar day' to 5_ - Be sure to tap the `Set` Button.

How do you "and then open a book" ? Gesture, menu? Please check how it behaves when you open the next book via History/Open last/File browser. May be there's only one case when this happen?

> 'Reset stats for books read for <....' doesn't work. I got "No statistics removed" all the time.

Well, it's supposed to work only once :) Once you removed stats < 5, and they are removed, if you hit that again, there's nothing left to remove - until you read books again for a small time. Unless you select another longer duration.
(Make a backup of your stats db if you're testing this :)

gerhaher:
> Do you see them updated in the menu, after you change them?
Yes.

> When I change 'Books per calendar day' to 5 - Be sure to tap the Set Button.

I did.

> Well, it's supposed to work only once :) Once you removed stats < 5, and they are removed, if you hit that again, there's nothing left to remove - until you read books again for a small time. Unless you select another longer duration.

I have stats that should be removed but aren't with this feature. 
I can reset statistics manually using the old 'Reset statistics per book'.



Hmm, now I have found out that it's because I'm in File manager it doesn't work. When I make the changes from File manager they don't stick. But if I'm in Reader and change them, they stick. Same thing for reset statistics.

poire-z:
> it's because I'm in File manager it doesn't work

Ok, that makes more sense :) Didn't test that as I'm not often in FM.
FM indeed does not broadcast/tell plugins to save their settings.
As for reset statistics, the SQL statements fail because there is no current book id to exclude.
Going to fix that.

PR #5871, opened 2020-02-16
Warn if color rendering is enabled on a grayscale device.

NiLuJe:
Can happen if a user imports settings from a color device (see today's discussion with @ptrm on gitter ;)).

Also, make sure the "Color rendering" checkbox is actually available in these instances ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5871)

### DIFF
--- a/reader.lua
+++ b/reader.lua
@@ -182,6 +182,16 @@ if Device:hasColorScreen() and not G_reader_settings:has("color_rendering") then
     })
 end
 
+-- Conversely, if color is enabled on a Grayscale screen (e.g., after importing settings from a color device), warn that it'll break stuff and adversely affect performance.
+if G_reader_settings:isTrue("color_rendering") and not Device:hasColorScreen() then
+    -- Don't disable it straight-away, might be useful for developers?
+    --G_reader_settings:delSetting("color_rendering")
+    local InfoMessage = require("ui/widget/infomessage")
+    UIManager:show(InfoMessage:new{
+        text = _("Color rendering is mistakenly enabled on your grayscale device.\nThis will subtly break some features, and adversely affect performance, please disable color rendering in the screen sub-menu."),
+    })
+end
+
 -- Helpers
 local lfs = require("libs/libkoreader-lfs")
 local function retryLastFile()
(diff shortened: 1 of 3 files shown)


NiLuJe:
I refrained from hard-deleting the setting on startup, because there might be legitimate uses for such a broken setup? Maybe? I don't know :D.

(If there really aren't, I'm all for uncommenting the `delSetting` ;)).

reader.lua
```diff
@@ -182,6 +182,16 @@ if Device:hasColorScreen() and not G_reader_settings:has("color_rendering") then
     })
 end
 
+-- Conversely, if color is enabled on a Grayscale screen (e.g., after importing settings from a color device), warn that it'll break stuff and adversely affect performance.
+if G_reader_settings:isTrue("color_rendering") and not Device:hasColorScreen() then
+    -- Don't disable it straight-away, might be useful for developers?
+    --G_reader_settings:delSetting("color_rendering")
+    local InfoMessage = require("ui/widget/infomessage")
+    UIManager:show(InfoMessage:new{
+        text = _("Color rendering is mistakenly enabled on your grayscale device.\nThis will subtly break some features, and adversely affect performance.\nPlease disable color rendering in the screen sub-menu."),
```
Frenzie: ConfirmBox?

NiLuJe: Oh, duh. Will do, thanks !

reader.lua
```diff
@@ -184,11 +184,17 @@ end
...
     UIManager:show(InfoMessage:new{
-        text = _("Color rendering is mistakenly enabled on your grayscale device.\nThis will subtly break some features, and adversely affect performance.\nPlease disable color rendering in the screen sub-menu."),
+        text = _("Color rendering is mistakenly enabled on your grayscale device.\nThis will subtly break some features, and adversely affect performance."),
+        cancel_text = _("Ignore"),
+        cancel_callback = function()
+                return
+        end,
+        ok_text = _("Disable"),
+        ok_callback = function()
+                G_reader_settings:delSetting("color_rendering")
```
poire-z: At the time the user hit OK, KOReader is already fully launched. Is that ok? (no side effect changing that value in the middle of a run? - or that it will have effect only on next launch?)

NiLuJe: It certainly doesn't implode, at least, but you're right that I had intended to check what the menu does in case there' s an extra switch to flip for immediate effect ;).

NiLuJe: And, of course, yes, there is ;p.

NiLuJe: Thanks, that had slipped my mind ;).

Fixed, and double-checked that it does take effect immediately :).

PR #5874, opened 2020-02-17
CoverBrowser: fix "Extract and cache" crash

### LINKED ISSUE #5873: 'Extract and cache book information': "No books were found"
* KOReader version: v2020.02-7-g95d18fe_2020-02-16/ 
* Device: KA1

#### Issue
'Extract and cache book information'-feature seems to be out of order.

I get "No books were found" all the time. 

I've tried both 'Here only' and 'Here and under', neither works.

poire-z:
Fix little util/ffiutil misuse in the last addition in #5816.
(Too many `util.` expecting it to be `ffi/util.lua` to change them all to `ffiutil`. or `BaseUtil.`.)
Closes #5873.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5874)

### DIFF
--- a/plugins/coverbrowser.koplugin/bookinfomanager.lua
+++ b/plugins/coverbrowser.koplugin/bookinfomanager.lua
@@ -651,6 +651,7 @@ function BookInfoManager:cleanUp()
 end
 
 local function findFilesInDir(path, recursive)
+    local stringStartsWith = require("util").stringStartsWith
     local dirs = {path}
     local files = {}
     while #dirs ~= 0 do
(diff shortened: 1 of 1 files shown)


plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -651,6 +651,7 @@ function BookInfoManager:cleanUp()
 end
 
 local function findFilesInDir(path, recursive)
+    local stringStartsWith = require("util").stringStartsWith
```
NiLuJe: Oh, yeah, good idea, makes the whole thing slightly less clunky to read ;).

poire-z: Not really intended for that :) I just thought that if I used `local util = require("util")` here, luacheck would complain about it shadowing the top one (and I didn't want to name it `frontutil` :)

PR #5885, opened 2020-02-19
continuous: do not limit visible area to page area

### LINKED ISSUE #5876: pdf continuous view mode doesn't cover all the screen space (tall aspect ratio)
* KOReader version: v2020.02
* Device: hisense a5

#### Issue
When reading pdf in my Hisense a5 with continuous view mode. There is always a big blank space at the bottom. No matter how I adjust the configuration. It can't show any content at the bottom. I have photographed some photos. 
4030e0c6a7efce1b6b83e6b4b851f3deb58f6500
715411d6277f9e2f879133760830e924b999f30c
3502711190ef76c6537ad5558a16fdfaae516700

#### Steps to reproduce
1. Open a pdf file in koreader with my Hisense a5
2. Change the view mode to continuous
3. There will be a big blank space at the bottom

yparitcher:
in scroll_mode allow visible area to be larger than one page

allows proper scrolling on tall aspect ratios. Fixes #5876

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5885)

### DIFF
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -575,8 +575,10 @@ function ReaderView:recalculate()
             self.visible_area.x = self.page_area.x + self.page_area.w - self.visible_area.w
             self.visible_area.y = self.page_area.y
         end
-        -- and recalculate it according to page size
-        self.visible_area:offsetWithin(self.page_area, 0, 0)
+        if not self.page_scroll then
+            -- and recalculate it according to page size
+            self.visible_area:offsetWithin(self.page_area, 0, 0)
+        end
         -- clear dim area
         self.dim_area.w = 0
         self.dim_area.h = 0
--- a/spec/unit/readerlink_spec.lua
+++ b/spec/unit/readerlink_spec.lua
@@ -44,7 +44,7 @@ describe("ReaderLink module", function()
         readerui:handleEvent(Event:new("SetZoomMode", "page"))
         readerui.paging:onGotoPage(1)
         assert.is.same(1, readerui.paging.current_page)
-        readerui.link:onTap(nil, {pos = {x = 250, y = 534}})
+        readerui.link:onTap(nil, {pos = {x = 228, y = 534}})
         UIManager:run()
         -- its really hard to get the exact page number in scroll mode
         -- page positions may have unexpected impact on page number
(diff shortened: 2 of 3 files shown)


NiLuJe:
Pretty much in the same boat as you, so, tomorrow it is ^^.

Frenzie:
There's a problem at the end of the document related to scrolling position offsets. I tried to illustrate it in these GIFs but it might be clearer if you try it for yourself.

### Current
This has good dragging/scrolling behavior, but with the cut-off issue. The cut-off issue just seldom happens to occur on your typical ereader-type aspect ratio.


### This PR
You see weird jumps. It's basically impossible to control correctly by panning because it jumps up.


Besides that I haven't noticed any suspicious behavior.

Frenzie:
Actually the scrolling offset problem is pretty global, just in paper.pdf it seemed to work out.

Frenzie:
My bad, this is a preexisting condition, unrelated to this PR. It's related to doing a quick pan, because then it's detected as a swipe and PanRelease is never called. That's probably why I didn't notice the problem at first.

If you make sure PanRelease is called, like this (which breaks many things ;-) ), it's not an issue:
```diff
diff --git a/frontend/apps/reader/modules/readerpaging.lua b/frontend/apps/reader/modules/readerpaging.lua
index 2f20312d..476a2ccd 100644
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -172,6 +172,14 @@ function ReaderPaging:setupTouchZones()
             },
             handler = function(ges) return self:onPanRelease(nil, ges) end
         },
+        {
+            id = "paging_swipe",
+            ges = "swipe",
+            screen_zone = {
+                ratio_x = 0, ratio_y = 0, ratio_w = 1, ratio_h = 1,
+            },
+            handler = function(ges) return self:onPanRelease(nil, ges) end
+        },
     })
 end
```

NiLuJe:
@Frenzie: No idea if related, but I've never actually managed to pan properly on an actual device in continuous mode, so if that's the same underlying issue, it's indeed fairly old ;).

Frenzie:
On a device panning naturally includes a delay so you'd be fairly likely to keep panning for longer than whatever the swipe time limit is (300? 500?)

Frenzie:
Anyway @NiLuJe shall we merge?

yparitcher:
I have also had this panning issue for a while. I just never got around to investigating it. Thanks

NiLuJe:
Yeah, haven't really looked into it, but this approach is the least intrusive possible (i.e., only if continuous), so it definitely makes sense anyway ;)..

NiLuJe:
> Anyway @NiLuJe shall we merge?

Yeah, haven't really looked into it, but this approach is the least intrusive possible (i.e., only if continuous), so it definitely makes sense anyway ;)..

PR #5891, opened 2020-02-21
macOS build info tweaks

NiLuJe:
Following https://github.com/koreader/koreader-base/pull/1052

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5891)


NiLuJe:
Made it neutral: Catalina (10.15) switched the default shell to ZSH, brew info will output the right info automagically.

PR #5894, opened 2020-02-24
kosync: validate username and password against empty/blank strings

### LINKED ISSUE #5832: Register Progress Sync
* KOReader version: latest stable 2020.01
* Device: Kindle voyage

#### Issue
I am not able to register a new account for syncing the progress.

#### Steps to reproduce
Open book, open progress sync add-on and press register

Unknown server error

pazos:
Fixes #5832

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5894)

### DIFF
--- a/plugins/kosync.koplugin/main.lua
+++ b/plugins/kosync.koplugin/main.lua
@@ -85,11 +85,11 @@ local function validateUser(user, pass)
     local user_ok = validate(user)
     local pass_ok = validate(pass)
     if not user_ok and not pass_ok then
-        error_message = _("Invalid user and password")
+        error_message = _("invalid username and password")
     elseif not user_ok then
-        error_message = _("Invalid user")
+        error_message = _("invalid username")
     elseif not pass_ok then
-        error_message = _("Invalid password")
+        error_message = _("invalid password")
     end
 
     if not error_message then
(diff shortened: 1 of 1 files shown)


plugins/kosync.koplugin/main.lua
```diff
@@ -277,31 +304,47 @@ function KOSync:login()
...
-                        })
+                        local ok, err = validateUser(username, password)
+                        if not ok then
+                            UIManager:show(InfoMessage:new{
+                                text = err,
+                                timeout = 3,
+                            })
+                        else
+                            self:closeDialog()
+                            UIManager:scheduleIn(0.5, function()
```
Frenzie: Seems like a very long delay. Is this still in line with similar code?

plugins/kosync.koplugin/main.lua
```diff
@@ -72,6 +72,33 @@ local function showSyncError()
...
+    end
+    return true
+end
+
+local function validateUser(user, pass)
+    local error_message = nil
+    local user_ok = validate(user)
+    local pass_ok = validate(pass)
+    if not user_ok and not pass_ok then
+        error_message = _("Invalid user and password")
```
Frenzie: Username is probably clearer.

plugins/kosync.koplugin/main.lua
pazos: 2s is enough. 1s could be but not if the user doesn't expect a message here.

PR #5898, opened 2020-02-25
[i18n] allow translation of time (12/24-hours formats)

poire-z:
Discussion around https://github.com/koreader/koreader/issues/5359#issuecomment-590593945

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5898)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -157,9 +157,16 @@ local footerTextGeneratorMap = {
         local prefix = symbol_prefix[symbol_type].time
         local clock
         if footer.settings.time_format == "12" then
-            clock = os.date("%I:%M%p")
+            if os.date("%p") == "AM" then
+                -- @translators this is the time in the morning in 12-hours format (%I is the hour, %M the minute)
+                clock = os.date(_("%I:%M AM"))
+            else
+                -- @translators this is the time in the afternoon in 12-hours format (%I is the hour, %M the minute)
+                clock = os.date(_("%I:%M PM"))
+            end
         else
-            clock = os.date("%H:%M")
+            -- @translators this is the time in 24-hours format (%H is the hour, %M the minute)
+            clock = os.date(_("%H:%M"))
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -157,9 +157,16 @@ local footerTextGeneratorMap = {
         local prefix = symbol_prefix[symbol_type].time
         local clock
         if footer.settings.time_format == "12" then
-            clock = os.date("%I:%M%p")
+            if os.date("%p") == "AM" then
+                -- @translators this is the time in the morning in 12-hours format (%I is the hour, %M the minute)
```
Frenzie: Could you make it a proper sentence? ("This … blah blah.")

Also it's 12-hour format & 24-hour format (or maybe just `clock` instead of format?) ^_^

poire-z: You mean like this?:
This is the time in the morning in 12-hour clock (%I is the hour, %M the minute).

Frenzie:
This is somewhat of a difficult one to translate. Historically, in the mid-19th century, the Dutch national railway company used `V.M.` and `N.M.` (before midday and after midday). No one knows those terms.

Presently, `a.m.` and `p.m.` are in the dictionary. I don't think anyone actually knows them as Dutch terms though, just people like me who are well-versed in English language & culture.

That leaves `8:00 's ochtends` and `3:00 's middags`, which is rather long. So I guess I'll go with obtuse `a.m.` and `p.m.`, hoping most people won't want to see a 12-hour clock.

poire-z:
One could also just let them both be `%I:%M`, without any am/pm.
I think that's how it should be in french - "9h du matin" or "9h du soir" would be too long, and analog clocks don't tell which half it is either.
And one usually know if it's the morning of the afternoon - unless on books really immersive :)

Frenzie:
True, that'd probably work. I'll leave it as `a.m.`/`p.m.` for now though.

PR #5904, opened 2020-02-26
bump crengine: normalized xpointers fix, FB2 footnotes tweaks

poire-z:
Will need a base bump for https://github.com/koreader/koreader-base/pull/1055.

Includes https://github.com/koreader/crengine/pull/329 :
- toStringV2(): fix when target node is a boxing node
- CSS font-family parsing: set usual default of sans-serif
- FB2 footnotes: fix spurious newline ("display: run-in")
- FB2 footnotes: disable hardcoded handling as in-page

cre.cpp: accepts classic FB2 footnotes in footnote popup detection

Make FB2 footnotes behaviour consistent with what we have for footnotes in EPUB:
- FB2 footnotes are no more rendered in-page by default
- In-page rendering can be enabled by the added Style tweak
- FB2 footnotes can also show in footnote popup (with some added not-so-nice CSS so MuPDF can render them as crengine)

`cre: setFontFace(): increase bias given to the main font`
An attempt at solving this:
<kbd>image</kbd>
when selecting FreeSerif as the default font, and using "Ignore publisher font-family": any of our other fonts could be used for the italic text (actually, the first of our font that has an italic variant).
So we can get the nicer:
<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5904)

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -614,6 +614,22 @@ This is just an example, that will need to be adapted into a user style tweak.]]
         {
             title = _("In-page footnotes"),
             {
+                id = "footnote-inpage_fb2";
+                title = _("In-page FB2 footnotes"),
+                description = _([[
+Show FB2 footnote text at the bottom of pages that contain links to them.]]),
+                -- (fb2.css already set font-size to 70% - so no need for a "smaller" variant)
+                css = [[
+body[name="notes"] section,
+body[name="comments"] section
+{
+    -cr-hint: footnote-inpage;
+    margin: 0 !important;
+}
+                ]],
+                separator = true,
+            },
+            {
                 id = "footnote-inpage_epub";
                 title = _("In-page EPUB footnotes"),
                 description = _([[
--- a/frontend/ui/widget/footnotewidget.lua
+++ b/frontend/ui/widget/footnotewidget.lua
@@ -89,6 +89,21 @@ body > li { list-style-type: none; }
 
 /* Remove any (possibly multiple) backlinks in Wikipedia EPUBs footnotes */
 .noprint { display: none; }
+
+/* Attempt to display FB2 footnotes as expected (as crengine does, putting
+ * the footnote number on the same line as the first paragraph via its
+ * support of "display: run-in" and a possibly added autoBoxing element) */
+body > section > autoBoxing > *,
+body > section > autoBoxing > title > *,
+body > section > title,
+body > section > title > p,
+body > section > p {
+    display: inline;
+}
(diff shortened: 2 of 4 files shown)


frontend/ui/data/css_tweaks.lua
```diff
@@ -613,6 +613,22 @@ This is just an example, that will need to be adapted into a user style tweak.]]
         },
         {
             title = _("In-page footnotes"),
+            {
+                id = "footnote-inpage_fb2";
+                title = _("In-page FB2 footnotes"),
```
Frenzie: Btw, did you see my comment on MR about being in favor of enabling these by default?

frontend/ui/widget/footnotewidget.lua
```diff
@@ -89,6 +89,21 @@ body > li { list-style-type: none; }
 
 /* Remove any (possibly multiple) backlinks in Wikipedia EPUBs footnotes */
 .noprint { display: none; }
+
+/* Attempt to display FB2 footnotes as expected (as crengine does, putting
+ * the footnote number on the same line as the first paragraph via its
+ * support of "display: run-in" and a possibly added autoBoxing element) */
+body > section > autoBoxing > *,
```
Frenzie: Bit late since it's already merged, but that `body >` part looks a bit unnecessary?

frontend/ui/data/css_tweaks.lua
poire-z: Yes I did. And you already mentionned that nearly a year ago.
See (again) my answer at https://github.com/koreader/koreader/issues/5015#issuecomment-491292942.
(Nothing much to add or remove from it one year later :| )

frontend/ui/widget/footnotewidget.lua
poire-z: Well, not necessary to catch these FB2 footnotes.
But having them, as we know these FB2 footnotes are pure SECTION > (TITLE/P), and they'll be child of the HTMLBoxWidget added BODY, having them like this, we'll be sure to NOT have that applied on any (probably super rare) DIV > SECTION > TITLE/P that we could find in the inner content of some EPUB footnotes :)

Frenzie: If you can end up with div > section couldn't you also end up with body > section?

poire-z: Yes, we could...
But in this case, I can't, here, distinguish if it comes from a FB2 footnote (and need these CSS tweaks) or if it comes from some EPUB footnote (and does need it).
But in the other case DIV > SECTION, I can.
So, just doing what I can :)

Issue #5794, opened 2020-01-27
[discussion] Structure of bookmarks/highlights/clippings data


mustafa-001:
While previously working on Evernote exporter I've come to see there is couple of unimplemented to dos and broken functionalities. Here is couple of them and my opinions:

1- Exporting functionality depends on two different source; highlights and bookmarks. In the past I've suggested moving everything to bookmarks and using highlights as fallback for no-text-only-image clippings. But `bookmarks` lacks page number info and changing `bookmarks` creation code is little tricky. See below point 4 for more info on this. [little discussion about this](https://github.com/koreader/koreader/pull/5774#issuecomment-575890933)

2- `clippings` structured with chapter name but for now all notes indexed with `chapter` equals `1`. While adding this info to `bookmarks` is easy, additional nesting makes sorting tricky. For now it's like this:
`table.sort(book, function(v1, v2) return v1[1].page < v2[1].page end)`
but if chapter  where not to be `1` it would require sorting function to flatten data before to be:
` function(v1, v2) 
	local p1, p2
	for _, c1 in pairs(p1) do
		p1 = c1.page
	...
	return p1<p2` in top of having unnecessary additional for loops in exporter functions. Advantage of not "fixing" this is not having to worry about existing evernote.metadata file. Having nested structures in Lua tables if we don't need to select with that info (for example the need to select bookmarks of page vs to select clippings of chapter) is unnecessary complication to me. 

3- Title and author of exported book notes are parsed from file name, and if file isn't formatted with calibre or similar tool output format gets messy. Fixing this is easy, just use `doc_props` from book metadata. Caveat here is existing clippings indexed with old info, new exports won't overwrite old ones.

4- `bookmarks`'s  `page`  is pre-calculated value for `page = if self.has_pages and xpos0.page or xpos0` . 
So it's actually duplicate info and gives two different types to a variable. Puts responsibility of deciding how to interpret it to called function, while similar behaviour can be achieved with only passing `pos0` with same amount of if checks. [Bookmark code has check for it 12 times](https://github.com/koreader/koreader/blob/2736661bfcf6b472da8fa06202cb0f5f92013180/frontend/apps/reader/modules/readerbookmark.lua#L319)
I'm bringing this not to change existing code(not worth it while it's working), but to ask if I were to add "actual page info" as seen by user how should I rename new value and should I add it in the first place? 

[...]

PR #5906, opened 2020-02-27
Fix dogear not shown - Fix FileChooser in RTL UI

poire-z:
When, on a same page, you bookmark the page (dogear) and highlight some text, changing page and going back, the dogear might be not shown: remove (some of) the higlights, and it's back !
Caused by some random ordering of bookmarks and highlights in the list, and the binary seach code stopping on the first highlight, possibly not visiting a bookmark just before or after.
So, just do some ordering and tweak the binary search.
(That code is quite messy, and hard to get into, because it feels like it does the reverse of what it says, ordering in the reverse order - bookmarks are stored last-page first, to be ready for the Bookmarks list descending order I guess - but then the whole list is shifted each time you add a bookmark as you progress in the book.... not fixing the whole thing, hoping we'll get that right with the migration to "annotations" work in #5794 :)

Also fix blank PathChooser when used in RTL UI when a CoverBrowser view is used (no issue when classic mode is used): just some stuff present in Menu.lua that didn't look necessary at the time I made CoverBrowser when things were all left-aligned - that I had to copy there.
Reported at https://github.com/koreader/koreader/issues/5359#issuecomment-591969974


poire-z:
For once, thank you, unit tests! And for this PR, I really do trust them to confirm I didn't mess that up :)

Issue #5194, opened 2019-08-10
ToC should show and expand current chapter


LucyOne:
* KOReader version: all
* Device: all

When a book has chapters and sub-chapters, the ToC does not expand to show them all. Even when I am in a specific chapter, this is not shown and I have to expand the entry myself. It would be useful to at least know your place in the book.

I use KOReader on my Pocketbook devices.

Frenzie:
I can't confirm. That's how it behaves?

I open a book at the beginning, the TOC is closed:
<img src=https://user-images.githubusercontent.com/202757/62821972-6fa68180-bb7d-11e9-9648-b6b7251b0766.png width=40%><img src=https://user-images.githubusercontent.com/202757/62821973-6fa68180-bb7d-11e9-8c97-12e20cc83658.png width=40%>

I go to somewhere much further in the book; the TOC is now automatically expanded:
<img src=https://user-images.githubusercontent.com/202757/62821975-6fa68180-bb7d-11e9-897b-b75b93c568c8.png width=40%><img src=https://user-images.githubusercontent.com/202757/62821976-703f1800-bb7d-11e9-9262-a0c430db5588.png width=40%>

LucyOne:
I'm sorry, I didn't realise that it only happens in the book I'm currently reading (Blink by Malcolm Gladwell).

Frenzie:
Likely a duplicate of #3062.

poire-z:
@LucyOne : so, it happened really with one single book of yours?
Anything specific about that book? (Found a copy of that book somewhere that did not exhibit any problem, but it had only one TOC level - so you must have another edition.)

Anyway, a known reason for TOC to not update is when a same ID= is used multiple times, which would result in a TOC target resolving to different pages, and this messing the ordering. Or just the TOC being initially not ordered. See https://github.com/koreader/koreader/issues/3062#issuecomment-521273214.

Could you post a screenshot of what your TOC looks like (like mine there), so we see if the pages are ordered or not?

LucyOne:
I looked at the ToC in Calibre, it seems to have unique IDs for each entry.
When opening the book in the last chapters "Notes" or "Conclusion", those were shown in bold. But when I switched to one of the other chapters, they weren't. In the screenshot I was in "2. Blink in Black and White".
![Reader_2019-Aug-14_165225](https://user-images.githubusercontent.com/15887012/63031585-7baf7d80-beb4-11e9-9836-e56cb225d55e.png)

[...]

PR #5907, opened 2020-02-28
TOC: validate and fix misordered page numbers

### LINKED ISSUE #3062: large nested epub file table of contents entries disappear and re-appear
* KOReader version: koreader-kobo-arm-kobo-linux-gnueabihf-v2015.11-1087-gba96506
* Device: kobo aura(original)

It is awesome to find that multi-level table of contents are supported in koreader. I plan to make good use of that.

I have an epub document i have created with a lot of small files inside. Something like 1400 files. 
I have created a .ncx file with 3 levels of nesting. Documents are listed more than once, organizing them in different way.

What i find is when i expand some of the later sections in the table of contents, the items appear correctly. However when i collapse the same section, then some following sections disappear on the table of contents screen. I can re-open the .epub file and missing entries re-appear. If there is a higher level section containing the problematic portion of the table of contents, if i contract and then expand this, then also the lower level stuff gets fixed.

I did try the sample file i found here, juliet.epub. This has a similar structure in the table of contents, and it works fine on the kobo. 

I wonder if there is some limit i am hitting here about number of sections? 

I also have viewed this same epub with other tools and it works fine there too.

poire-z:
Our code expects (rightfully) the TOC items to be ordered and to have increasing page numbers, but we may occasionally not get that from the engines (usually, because of bugs or duplicated IDs in the document).
Check for that and fix page numbers (possibly making them wrong) to have a working TOC.
Some details in https://github.com/koreader/koreader/issues/3062#issuecomment-402746843
Closes #3062. Issue also met in #5194.

That China.EN.epub fixed TOC would then appear as:
<kbd>image</kbd>
(which is not super right, the real page number of that History entry is 14 - but I'm just attempting to have the TOC working, not to make tha page right, which is impossible, just less bogus enough to not be misordered).

On a TOC with randomly shifted page numbers, it would show as:
<kbd>image</kbd>

(Thought about saving toc_is_valid = true/false in the TOC, so that scan is only done on the first opening of a book - but that quick scan single loop is fast, so probably sane and cheap to do that on each opening or re-rendering.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5907)

### DIFF
--- a/frontend/apps/reader/modules/readertoc.lua
+++ b/frontend/apps/reader/modules/readertoc.lua
@@ -10,8 +10,10 @@ local Geom = require("ui/geometry")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local Menu = require("ui/widget/menu")
 local UIManager = require("ui/uimanager")
+local logger = require("logger")
 local _ = require("gettext")
 local Screen = Device.screen
+local T = require("ffi/util").template
 
 local ReaderToc = InputContainer:new{
     toc = nil,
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readertoc.lua
```diff
@@ -89,6 +91,113 @@ function ReaderToc:fillToc()
...
+
+    -- For testing: shuffle a bit a valid TOC and make it randomely invalid
+    -- for i = first, last do
+    --     toc[i].page = toc[i].page + math.random(10) - 5
+    -- end
+
+    -- Do a cheap quick scan first
+    logger.dbg("validateAndFixToc(): quick scan")
+    local has_bogus
+    local cur_page = -1
```
Frenzie: Not 0?

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -280,6 +389,9 @@ function ReaderToc:onShowToc()
         for _,v in ipairs(self.toc) do
             v.text = self.toc_indent:rep(v.depth-1)..self:cleanUpTocTitle(v.title)
             v.mandatory = v.page
+            if v.orig_page then -- bogus page fixed: show original page number
+                v.mandatory = T("(%1) %2", v.orig_page, v.page)
```
Frenzie: Bidi?

poire-z: Looks like BiDi will handle it automagically:
<kbd>image</kbd>
(wider screen width, so not same numbers as first post screenshots)

Issue #5920, opened 2020-03-06
In landscape mode frontlight level on status bar causes screen artifact on poweroff.


clarkspark:
* KOReader version: 2020.02
* Device: kobo_nova (Kobo ClaraHD)

#### Issue
In landscape mode the frontlight level on status bar causes screen artifact on poweroff.  The frontlight level status bar item being selected is what causes the artifact, the presence (or absence) of any of the other possible status bar items makes no difference.

#### Steps to reproduce
1 ensure statusbar is enabled
2 ensure the frontlight level box is checked
3 open an epub
4 set to landscape view
5 power off device

A vertical bar is then seen across the landscape text on my device.  The bar remains on the screen when the device is powered back on, and screen refresh removes it until the next power-off.  The frontlight indicator on the status bar (along the bottom) remains at the value when the device was on, and the 'anomalous' frontlight level shows the correct value.

![device off, horziontal, w_ light status bar showing issue](https://user-images.githubusercontent.com/14674610/76088389-f7266880-5f85-11ea-9dfb-67a0de8dca27.jpg)

clarkspark:
I just verified that if the frontlight is set to 0 (which is displayed as 'Off' in the status bar) then the artifact is not triggered on suspend/poweroff.

Which would seem to indicate that it is the frontlight statusbar update from the '15%' string (in the scenario shown in the photo) to the 'Off' string which is mis-written to a perpendicular orientation.

poire-z:
Probably due to what happens on suspend/poweroff:
- we set back the screen rotation to portrait (for screensaver images that are usually portrait).
- we turn off frontlight
- turning off frontlight requests the footer to update itself so, displaying itself in portrait mode (but then, no idea why just that FL indicator would be displayed...)
- some screen refresh here and there in that process :)

When the light is off, there's no need for these steps 2 & 3.

https://github.com/koreader/koreader/blob/8502a96a6a0e1d76b8c8d2747f540102b48b3d8b/frontend/ui/uimanager.lua#L56-L68
https://github.com/koreader/koreader/blob/8502a96a6a0e1d76b8c8d2747f540102b48b3d8b/frontend/device/generic/device.lua#L193
https://github.com/koreader/koreader/blob/8502a96a6a0e1d76b8c8d2747f540102b48b3d8b/frontend/device/generic/device.lua#L227-L230
[...]

PR #5928, opened 2020-03-07
Correct Frontlight status on suspend when screensaver mode is 'Leave …

clarkspark:
Address #5920 to correct Frontlight status on suspend when screensaver mode is 'Leave screen as it is' in landscape mode.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5928)

### DIFF
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -232,7 +232,8 @@ function Device:onPowerEvent(ev)
         -- Mostly always suspend in Portrait/Inverted Portrait mode...
         -- ... except when we just show an InfoMessage or when the screensaver
         -- is disabled, as it plays badly with Landscape mode (c.f., #4098 and #5290)
-        if (G_reader_settings:readSetting("screensaver_type") ~= "message" and G_reader_settings:readSetting("screensaver_type") ~= "disable") then
+        if G_reader_settings:readSetting("screensaver_type") ~= "message" and
+           G_reader_settings:readSetting("screensaver_type") ~= "disable" then
             self.orig_rotation_mode = self.screen:getRotationMode()
             -- Leave Portrait & Inverted Portrait alone, that works just fine.
             if bit.band(self.orig_rotation_mode, 1) == 1 then
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -231,7 +231,7 @@ function Device:onPowerEvent(ev)
         logger.dbg("Suspending...")
         -- Mostly always suspend in Portrait/Inverted Portrait mode...
         -- ... except when we just show an InfoMessage, it plays badly with Landscape mode (c.f., #4098)
-        if G_reader_settings:readSetting("screensaver_type") ~= "message" then
+        if (G_reader_settings:readSetting("screensaver_type") ~= "message" and G_reader_settings:readSetting("screensaver_type") ~= "disable") then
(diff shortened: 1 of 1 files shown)


NiLuJe:
I'd add a ref to your own issue in the comment above, but that's just me ;).

clarkspark:
Sorry I'm not too familiar with the workings of github, I thought adding the #5290 would do that.  Or do you mean adding a short synopsis of the issue here?

NiLuJe:
No, I meant in the actual code ;).

frontend/device/generic/device.lua
```diff
@@ -231,7 +231,7 @@ function Device:onPowerEvent(ev)
         logger.dbg("Suspending...")
         -- Mostly always suspend in Portrait/Inverted Portrait mode...
         -- ... except when we just show an InfoMessage, it plays badly with Landscape mode (c.f., #4098)
```
NiLuJe: i.e., here ;).

NiLuJe:
:1:, lgtm, thanks ;).

clarkspark:
Thanks for the feedback, and the great software.

frontend/device/generic/device.lua
```diff
@@ -230,8 +230,9 @@ function Device:onPowerEvent(ev)
         local UIManager = require("ui/uimanager")
         logger.dbg("Suspending...")
         -- Mostly always suspend in Portrait/Inverted Portrait mode...
-        -- ... except when we just show an InfoMessage, it plays badly with Landscape mode (c.f., #4098)
-        if G_reader_settings:readSetting("screensaver_type") ~= "message" then
+        -- ... except when we just show an InfoMessage or when the screensaver
+        -- is disabled, as it plays badly with Landscape mode (c.f., #4098 and #5290)
+        if (G_reader_settings:readSetting("screensaver_type") ~= "message" and G_reader_settings:readSetting("screensaver_type") ~= "disable") then
```
poire-z: No real need for parens, as it's a single `and`.
(When I see parens, I expect more twisted conditions - my expectations would be deceived here :)

clarkspark: I can remove the parens.  Are overly-long lines frowned upon?  Like maybe instead break the if across two lines, in which case I'd need the parens, right?

poire-z: Yep, 145 chars long is a bit too long (120 is ok) - so you can split it on 2 lines.
And still no need for parens :) look a few lines below for a multilines `and` and around to get a feel for the Lua style and simple needs.

Frenzie: You could also do

```lua
local screensaver_type = G_reader_etc_long_long()
if screensaver_type ~= blah and screensaver_type ~= blahblah then -- much shorter and more legible
```

clarkspark: @Frenzie:  not a big deal I can make that change but if its alright I'd prefer just to keep it consistent with with the code around it (~12 lines below with the multi-line if using the same function)

NiLuJe: Quite possibly because it's from me (I like long lines & parentheses, and I cannot lie 🎵).

clarkspark: I went with shorter lines for the whole if branch (even for un-effected code), let me know if that's OK.

clarkspark: @Frenzie: I went with a variable to replace the long function name.  Thanks for the feedback.

PR #5931, opened 2020-03-08
Make xpointers migration ConfirmBox non-dismissable

poire-z:
Because dismissing it by tapping outside would have "not for this book" selected, which is the least welcome action...
Details at https://github.com/koreader/koreader/pull/5897#issuecomment-596209498.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5931)

### DIFF
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -1276,6 +1276,10 @@ Note that %1 (out of %2) xpaths from your bookmarks and highlights have been nor
 
     UIManager:show(ConfirmBox:new{
         text = text,
+        -- Given the layout of the buttons (Cancel|OK, and a big other button below
+        -- with "Not now"), we don't want cancel_callback to be called when dismissing
+        -- this ConfirmBox by taping outside. So, make it non dismissable.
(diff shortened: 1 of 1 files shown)


ptrm:
And having dismissed it for one book, how to get xpointers migrated anyway :D ?

poire-z:
sorry :|
Just edit metadata.epub.lua for this book, and remove the line:
```lua
["cre_keep_old_dom_version"] = true,
```
and re-open the book: you'll get tha ConfirmBox again.

poire-z:
@Frenzie : I see you've made a 2020.03.1 yesterday (thanks!), but there's no binary asset uploaded.
I also see there's been 4 or 5 pipeline runs yesterday outside regular nightlies :) any problem ?

Frenzie:
It's not picking up on the tag, not sure why, the result is exactly the same for all intents and purposes but it gives an uglier filename/version number.

Frenzie:
Actually it looks like doing it through the GH web GUI doesn't do an annotated tag. /sigh

Frenzie:
So yeah, what's wrong is that GH is pretty useless if you aren't at your regular PC with proper command-line Git.

PR #5940, opened 2020-03-12
Add additional MIME types to OPDS Browser

jstirling:
Updating the OPDS browser formats list to include CBR and CBZ to bide time until the DocumentRegistry rework

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5940)


jstirling:
> 
> 
> Err, might be going a bit overboard? Pinging @NiLuJe @poire-z for second opinion. ;-)

Yeah, I'll dial it back. Just pulled the list from [COPS](https://github.com/seblucas/cops/blob/dad459f12cafa4af1a31dfdea03590b52ebfe934/lib/Data.php). 
Didn't even think to check for duplicates which is causing it to fail.

poire-z:
Obviously, we don't support all added formats, but it seems this list is just used to filter/show the downloads format list - so, I guess on android, one could be happy to be able to download a MP3 version of a book from some OPDS server :).
Another option would be to filter that list with a checkbox: formats supported by KOReader | all formats.
(note that I don't use the OPDS browser, so I don't know how this can be inconvenient to those who do)

jstirling:
it only shows formats that exist on the OPDS server. So having more formats doesn't clutter the view unless someone has a ton of different formats for a single book.

I removed all but CBZ and CBR now, but can add them back (minus the duplicates that caused issues) if you'd prefer. I just figured I'd add them originally to prevent someone having this same issue with a different type. Seems unlikely though so might not be an issue.

frontend/ui/widget/opdsbrowser.lua
```diff
@@ -53,6 +53,8 @@ local OPDSBrowser = Menu:extend{
         ["text/plain"] = "TXT",
         ["application/x-mobipocket-ebook"] = "MOBI",
         ["application/x-mobi8-ebook"] = "AZW3",
+        ["application/x-cbz"] = "CBZ",
+        ["application/x-cbr"] = "CBR",
```
avsej: we could also have djvu here

avsej: https://github.com/koreader/koreader/pull/5982

Issue #5022, opened 2019-05-11
Framework drawing clock & battery over KOReader


Pandentia:
* KOReader version: 2019.05, 2019.04
* Device: Kindle Paperwhite 4 (No SOs/8GB/WiFi-only) with (consumer) firmware 5.10.1.3

#### Issue

Kindle's framework (Home UI) seems to be drawing certain UI elements over KOReader spontaneously.

#### Steps to reproduce

None specifically (sorry), though I have noticed an increase in occurrence after a pop-up message appears at the top of the screen ("Frontlight is off.", etc). I've also seen it happen with the top menu open or shortly after closing, and just today I've seen it shortly after the screensaver triggered. (Just as I was writing this report, I've seen it occur while opening a book and the "Opening {filename}" message was displaying, twice.) **TL;DR:** Pretty much everywhere, mostly after touch events, randomly.

Presumably some condition combined with a framework update requiring a redraw (clock roll-over, battery percentage update, etc). Thus far I've seen it happen with battery status and clock, but not WiFi indicator. ~~I'll attach some screenshots later if I can catch it again (and find a way to scrape it off the framebuffer).~~

Edit: Upon closer inspection of related (Kindle) issues, this seems to be a known but unreported bug. I'll leave it open.

#### File browser, shortly after KOReader restart through top menu

![screenshot_2019_05_11T21_59_26+0200_blurred](https://user-images.githubusercontent.com/10685096/57574444-90a1ee00-7439-11e9-8b15-acf8e2683c71.png)

NiLuJe:
Is the clock up to date when it seeps through?

But, yeah, there's nothing much we can do about it, there's already a giant pile of hacks in place to make it mostly behave...

Pandentia:
It is, yes. Same applies for battery status.

Frenzie:
I'll close this since @NiLuJe says there's nothing that can be done. I hope it's not too annoying. :-)

Pandentia:
Okay. I'll report back if I do happen to find a fix nonetheless.

yparitcher:
this is fixed by #5943

PR #5943, opened 2020-03-13
[Kindle] Hide the status bar on current FW, too.

NiLuJe:
That was probably not a great idea, since we live inside a stock-UI controlled window? I dunno, I don't use KPVBooklet ;p.

Also comment on an unsuccessful experiment re: https://www.mobileread.com/forums/showthread.php?t=328015

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5943)


NiLuJe:
That said, if we're all amenable to shipping a pre-compiled 34KB binary (it depends on X11, would be slightly hellish to build), I can fix it "properly" by telling the WM to basically make the status bar go sit in a corner.

NiLuJe:
Hmm, it's (wmctrl) not perfect either, it makes the WM timeout when it attempts to refresh... Which happens on exit, or on suspend/resume on SO devices :/.

yparitcher:
> That said, if we're all amenable to shipping a pre-compiled 34KB binary (it depends on X11, would be slightly hellish to build), I can fix it "properly" by telling the WM to basically make the status bar go sit in a corner.

Would this fix the clock bleeding issue on later firmware >= 5.11?
if so could i just build it against the x11 libs on my device?

Frenzie:
> would be slightly hellish to build

PS Obviously maintainability is a much larger concern than 30 kB more or less. ^_^

NiLuJe:
@yparitcher: When does it happen, exactly? (And I assume your *not* using a Special Offers device?).

(I've got a binary, that's not the issue, but thanks ;)).

@Frenzie: It's unlikely to ever be updated again, anyway ^^.

yparitcher:
> @yparitcher: When does it happen, exactly? (And I assume your _not_ using a Special Offers device?).

mostly once in a while when changing/refreshing pages, the clock or wifi icon from kindle bleed into koreader as reported in #5022 and others. (the device is technically SO, but never connected to amazon so it does not know it, so it runs like a regular device.)

> (I've got a binary, that's not the issue, but thanks ;)).

thanks, I just managed to build it, after finding the right libs and headers from my kindle.

the "wmctrl -r L:C_N:titleBar_ID:system -b toggle,hidden" command works on 5.11
in a few days i can let you know if it stops the clocks bleeding

NiLuJe:
@yparitcher: Does it get detected as such by KOReader (i.e., do you get the stock screensavers, or ours?).

Just for kicks, because twobob's binary is a wee bit old, I'm currently running a BR build of it, because I'm lazy, and it supports crosstool-ng toolchains ;p.

yparitcher:
> @yparitcher: Does it get detected as such by KOReader (i.e., do you get the stock screensavers, or ours?).

the SS hack works fine, i can also get koreader SS. I happen to have them set to "do nothing" and running my own screensavers with https://github.com/yparitcher/kindle-zmanim

> Just for kicks, because twobob's binary is a wee bit old, I'm currently running a BR build of it, because I'm lazy, and it supports crosstool-ng toolchains ;p.

i downloaded wmctrl from http://tripie.sweb.cz/utils/wmctrl/ and built it against the X libraries from my kindle, with headers from the kindle sources and old ubuntu packages.

yparitcher:
If you dont want to maintain a build, Maybe add a check in the startup script, if the user has wmctrl in the PATH then run it, and leave it up to the user to get it.

NiLuJe:
It's 30KB, I'm just going to ship a K5 binary everywhere ;).

yparitcher:
mine is 26KB after stripping ;)

NiLuJe:
Indeed it is, twobob's binary probably wasn't stripped ;).

NiLuJe:
I'd simplify all the mess and just use wmctrl everywhere, but I'm too lazy to check how it behaves on older FWs :D.

NiLuJe:
Note that as I said last night, I went with a hard resize instead of the minimize, because otherwise it was tripping a timeout when ligl couldn't find the titlebar, which was triggering an unsightly stalls and extra flash on suspend/resume (at least).

yparitcher:
Your branch works on my KT4 5.11.

resizing instead of hiding the status bar has the side effect of the stock UI shifting up while koreader loads :)

NiLuJe:
Yep. I don't actually remember how it behaved back when the pillow command worked ;).

I suppose it did the same thing, because the winmgr visibility event (which I cannot manage to fire manually outside of winmgr) appears to be doing that? Eh.

At least it makes the refresh very visible, which should take care of the "oh, no, I left wifi on" concern ;).

hius07:
PW2, nightly v2020.03.2-9
Upon exit from KOReader status bar always displays correct state of Aeroplane mode.
Perfectly, thanks a lot.

yparitcher:
just want to say that this doe fix the stock clock from showing.

a month later and i have not seen the clock once.
Thanks :+1:

PR #5949, opened 2020-03-13
Don't break OPDS parsing on HR tags.

### LINKED ISSUE #5948: OPDS browser not listing content when encountering <hr/> tags
* KOReader version: master
* Device: probably irrelevant, tested on a PW2.

#### Issue

Incomplete listings in a local Calibre OPDS listing.

#### Steps to reproduce

Find > OPDS Catalog > Local calibre library

Try to look at stuff, realize it looks a wee bit emptier than it ought to.
Then, thanks to a recent addition with an "é" in the Author's name, get a lightbulb moment.

I've never used the thing, I basically gave it a try to double-check something on gitter, so I don't know if it's just me, it's a known issue, or a weird quirk of my setup.

(Calibre 4.12.0, Python 2.7. Stuff is visible in Calibre's own Content server web UI. I don't think i've got an OPDS browser on the desktop to double-check, though ;D).

NiLuJe:
Apply the same crappy workaround than for BR.

Fix #5948

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5949)

### DIFF
--- a/frontend/ui/opdsparser.lua
+++ b/frontend/ui/opdsparser.lua
@@ -79,6 +79,9 @@ function OPDSParser:parse(text)
     -- but will kick the ass of luxl
     text = text:gsub("<br>", "<br />")
     text = text:gsub("<br/>", "<br />")
+    -- Same deal with hr
+    text = text:gsub("<hr>", "<hr />")
+    text = text:gsub("<hr/>", "<hr />")
     -- some OPDS catalogs wrap text in a CDATA section, remove it as it causes parsing problems
     text = text:gsub("<!%[CDATA%[(.-)%]%]>", function (s)
         return s:gsub( "%p", {["&"] = "&amp;", ["<"] = "&lt;", [">"] = "&gt;" } )
--- a/frontend/luxl.lua
+++ b/frontend/luxl.lua
@@ -1,19 +1,19 @@
 --[[
-	This code was derived from the pico_xml project
-	which can be found here:
+    This code was derived from the pico_xml project
+    which can be found here:
 
-	http://kd7yhr.org/bushbo/pico_xml.md
+    http://kd7yhr.org/bushbo/pico_xml.md
 
-	The original C code, written by: Brian O. Bush
+    The original C code, written by: Brian O. Bush
 
-	Pure Lua Version written by: William A Adams
-	Dramatic Speed Improvements by: Robert G Jakabosky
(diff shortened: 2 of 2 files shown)


NiLuJe:
Also resync luxl.lua w/ upstream, and mention said upstream, because it's surprisingly sneaky to find.

frontend/ui/opdsparser.lua
```diff
@@ -79,6 +79,9 @@ function OPDSParser:parse(text)
     -- but will kick the ass of luxl
     text = text:gsub("<br>", "<br />")
     text = text:gsub("<br/>", "<br />")
+    -- Same deal with hr
+    text = text:gsub("<hr>", "<hr />")
+    text = text:gsub("<hr/>", "<hr />")
```
Frenzie: That's necessary for this luxl thing? `<hr/>` is perfectly valid. The main reason `<hr />` is still practiced is compatibility with some late '90s/early 2000s browsers and either people don't know that and they're stupidly copying the way it's always been done or perhaps they just prefer the look of `<hr />`.

Btw, while I imagine [the full list](http://xahlee.info/js/html5_non-closing_tag.html) is unlikely to show up, what about ``?

NiLuJe: I don't have anything with img tags to check, but at the very list it's rarely (if ever) going to be empty, which might not trip the state machine the same way?

Frenzie: Depends if the person writing the code wants to send technically valid HTML when JS is disabled. Of course `` or `` is invalid, but if the src is going to be filled out by JS (doesn't matter one bit if we're talking vanilla, jQuery, React, Angular, etc., etc.) then `placeholder` might be technically valid but obviously in a completely meaningless way.

Fancier still is something like: `placeholder`. Valid per spec, no HTTP request, and the validator will give it a nice green checkmark, but any person who thinks it's therefore oh-so-much better than just `` lacks a brain. Of course if the validator is part of your process you want to avoid false positives, no argument there.

But anyway, even if it's not done on purpose I'm sure it can happen by accident. ;-) Like if you're using React to do something like `` where `src` is accidentally `null` instead of `src="image.jpg"`. Sounds like I made up some pretty stupid code but I'm sure there's way worse out there in the wild.

Frenzie: PS `<wbr>` is always used just like that, in any case.

NiLuJe: I dunno, that all seems fairly unlikely ^^.

PR #5962, opened 2020-03-15
bump crengine: handle block elements inside inline element

poire-z:
Includes https://github.com/koreader/crengine/pull/330
- FB2 footnotes: disable hardcoded handling as in-page 2
- CSS: also support gray colors spelled "grey"
- Text: move past floats when wide image or inlineBox does not fit
- Fix list item marker clipping check
- Ensure nodeDisplayStyleHash change when boxing prevented by cache
- Text: adds LFormattedText.is_reusable
- Handle block elements inside inline element
- Fix getBookmark() on multi-pages embedded block
- (Upstream) Support raw picture in RTF
- (Upstream) Fixed integer overflow in cache size calculations
- Add libCURL's CA bundle to the ignore list https://github.com/koreader/crengine/pull/332
- FB2 footnotes: fix setBoxingWishedButPreventedByCache() https://github.com/koreader/crengine/pull/333
- fb2.css: handle body[name="comments"] as footnotes

Also: Update base/thirdparty/turbo to v2.1.3 https://github.com/koreader/koreader-base/pull/1064

Also includes some FB2 footnotes tweaks, discussion around https://github.com/koreader/crengine/pull/329#issuecomment-599046771.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5962)

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -621,19 +621,52 @@ This is just an example, that will need to be adapted into a user style tweak.]]
         {
             title = _("In-page footnotes"),
             {
-                id = "footnote-inpage_fb2";
                 title = _("In-page FB2 footnotes"),
-                description = _([[
+                {
+                    id = "footnote-inpage_fb2";
+                    title = _("In-page FB2 footnotes"),
+                    description = _([[
 Show FB2 footnote text at the bottom of pages that contain links to them.]]),
-                -- (fb2.css already set font-size to 70% - so no need for a "smaller" variant)
-                css = [[
-body[name="notes"] section,
-body[name="comments"] section
-{
+                    -- Avoid 75% of 75% in case of nested <section>
+                    css = [[
+body[name="notes"] section {
     -cr-hint: footnote-inpage;
     margin: 0 !important;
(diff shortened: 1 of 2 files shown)


poire-z:
imbricated ? The romans did :)

https://www.merriam-webster.com/dictionary/imbricate
> The ancient Romans knew how to keep the interior of their villas dry when it rained. They covered their roofs with overlapping curved tiles so the "imber" (Latin for pelting rain or "rain shower") couldn't seep in. The tiles were, in effect, "rain tiles," so the Romans called them "imbrices" (singular "imbrex"). The verb for installing the tiles was "imbricare," and English speakers used its past participle - "imbricatus" - to create "imbricate," which was first used as adjective meaning "overlapping (like roof tiles)" and later became a verb meaning "to overlap."

But OK, let's go with "nested" then :)

poire-z:
Note that with `Ensure nodeDisplayStyleHash change when boxing prevented by cache`, we may get more often:
image
If you get this when just changing font size (or Font hinting, that's enough for testing it), it's that we have a bug somewhere (or that I misjudge how this can be triggered): please report, and share the book on which it happens if you can.

PR #5968, opened 2020-03-15
Next/Previous Bookmark Gestures

### LINKED ISSUE #5965: Gestures for jumping to next or previous bookmark?
Would it be possible to have menu items for jumping to the next or previous bookmark?

I know this is possible with the skimmer, but the skimmer when active covers almost a third of the visible page, so it is not really handy for jumping from bookmark to bookmark. To read the page I skimmed to I have to close the skimmer, then re-invoke it to jump to the next bookmark, close the skimmer, read the page, re-invoke the skimmer, etc. Tedious!

When gestures could be added for jumping back and forth between bookmarks (just like double tapping for jumping between chapters), this would be a real power feature!

yparitcher:
Fixes #5965

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5968)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -209,8 +209,10 @@ function ReaderBookmark:onPosUpdate(pos)
 end
 
 function ReaderBookmark:gotoBookmark(pn_or_xp)
-    local event = self.ui.document.info.has_pages and "GotoPage" or "GotoXPointer"
-    self.ui:handleEvent(Event:new(event, pn_or_xp))
+    if pn_or_xp then
+        local event = self.ui.document.info.has_pages and "GotoPage" or "GotoXPointer"
+        self.ui:handleEvent(Event:new(event, pn_or_xp))
+    end
 end
 
 function ReaderBookmark:onShowBookmark()
@@ -599,6 +601,16 @@ end
 
 function ReaderBookmark:onGotoPreviousBookmark(pn_or_xp)
     self:gotoBookmark(self:getPreviousBookmarkedPage(pn_or_xp))
+    return true
+end
+
+function ReaderBookmark:onGotoNextBookmarkFromPage(pn_or_xp)
+    self:gotoBookmark(self:getNextBookmarkedPageFromPage(pn_or_xp))
+    return true
+end
+
+function ReaderBookmark:onGotoPreviousBookmarkFromPage(pn_or_xp)
+    self:gotoBookmark(self:getPreviousBookmarkedPageFromPage(pn_or_xp))
     return true
 end
 
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -604,18 +604,18 @@ function ReaderBookmark:onGotoPreviousBookmark(pn_or_xp)
     return true
 end
 
-function ReaderBookmark:onGotoNextBookmarkFromPage(pn_or_xp)
-    self:gotoBookmark(self:getNextBookmarkedPageFromPage(pn_or_xp))
-    return true
-end
-
-function ReaderBookmark:onGotoPreviousBookmarkFromPage(pn_or_xp)
-    self:gotoBookmark(self:getPreviousBookmarkedPageFromPage(pn_or_xp))
-    return true
-end
-
 function ReaderBookmark:onGotoNextBookmark(pn_or_xp)
     self:gotoBookmark(self:getNextBookmarkedPage(pn_or_xp))
+    return true
+end
+
+function ReaderBookmark:onGotoNextBookmarkFromPage()
+    self:gotoBookmark(self:getNextBookmarkedPageFromPage(self.ui:getCurrentPage()))
+    return true
+end
+
+function ReaderBookmark:onGotoPreviousBookmarkFromPage()
+    self:gotoBookmark(self:getPreviousBookmarkedPageFromPage(self.ui:getCurrentPage()))
     return true
 end
 
--- a/frontend/apps/reader/modules/readergesture.lua
+++ b/frontend/apps/reader/modules/readergesture.lua
@@ -28,6 +28,8 @@ local action_strings = {
     page_jmp_fwd_1 = _("Next page"),
     prev_chapter = _("Previous chapter"),
     next_chapter = _("Next chapter"),
+    prev_bookmark = _("Previous bookmark"),
+    next_bookmark = _("Next bookmark"),
     go_to = _("Go to"),
     skim = _("Skim"),
     back = _("Back"),
@@ -702,6 +704,8 @@ function ReaderGesture:buildMenu(ges, default)
         {"page_jmp_fwd_1", not self.is_docless},
         {"prev_chapter", not self.is_docless},
         {"next_chapter", not self.is_docless},
+        {"prev_bookmark", not self.is_docless},
+        {"next_bookmark", not self.is_docless},
         {"go_to", true},
         {"skim", not self.is_docless},
         {"back", true},
@@ -1300,6 +1304,10 @@ function ReaderGesture:gestureAction(action, ges)
         self.ui:handleEvent(Event:new("GotoNextChapter"))
     elseif action == "prev_chapter" then
         self.ui:handleEvent(Event:new("GotoPrevChapter"))
+    elseif action == "next_bookmark" then
+        self.ui:handleEvent(Event:new("GotoNextBookmarkFromPage", self:getCurrentPage()))
+    elseif action == "prev_bookmark" then
+        self.ui:handleEvent(Event:new("GotoPreviousBookmarkFromPage", self:getCurrentPage()))
     elseif action == "go_to" then
         self.ui:handleEvent(Event:new("ShowGotoDialog"))
     elseif action == "skim" then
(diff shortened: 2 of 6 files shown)


frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -209,8 +209,10 @@ function ReaderBookmark:onPosUpdate(pos)
 end
 
 function ReaderBookmark:gotoBookmark(pn_or_xp)
-    local event = self.ui.document.info.has_pages and "GotoPage" or "GotoXPointer"
-    self.ui:handleEvent(Event:new(event, pn_or_xp))
+    if pn_or_xp then
```
Frenzie: Is this related to a specific problem you came across?

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1631,4 +1639,14 @@ function ReaderGesture:onToggleReadingOrder()
     return true
 end
 
+function ReaderGesture:getCurrentPage()
```
Frenzie: We don't already have something like this? And are we sure this is where it should go?

frontend/apps/reader/modules/readerbookmark.lua
yparitcher: at the first bookmark and previous, or last bookmark and next.
`pn_or_xp` will be nill crashing koreader.
instead of putting a check in every function, i put one check at execution

frontend/apps/reader/modules/readergesture.lua
yparitcher: not sure, if we do please let me know it would be helpfull all over the place :)
right now we have a bunch of 
```lua
    if self.document.info.has_pages then
        page = self.ui.paging.current_page
    else
        page = self.document:getCurrentPage()
    end
```
all over the codebase. so i assumed there wasnt one yet.

In general i think this should be fixed everywhere

Frenzie: I think something like `Document:getCurrentPage` would probably make the most sense.

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1300,6 +1304,10 @@ function ReaderGesture:gestureAction(action, ges)
         self.ui:handleEvent(Event:new("GotoNextChapter"))
     elseif action == "prev_chapter" then
         self.ui:handleEvent(Event:new("GotoPrevChapter"))
+    elseif action == "next_bookmark" then
+        self.ui:handleEvent(Event:new("GotoNextBookmarkFromPage", self:getCurrentPage()))
```
poire-z: Does this getCurrentPage() business have to really be in this readergesture file?
Can't we have just `self.ui:handleEvent(Event:new("GotoNextBookmarkFromCurrentPage"))` and have the getting of the current page in the 2 functions you added?

frontend/apps/reader/modules/readergesture.lua
yparitcher: a qiuck `ag` shows
https://github.com/koreader/koreader/blob/8d8b8cd0f79455ac4e4bc1dc25524a98a24cdeea/frontend/apps/reader/modules/readergesture.lua#L1568
https://github.com/koreader/koreader/blob/8d8b8cd0f79455ac4e4bc1dc25524a98a24cdeea/frontend/apps/reader/skimtowidget.lua#L66
https://github.com/koreader/koreader/blob/8d8b8cd0f79455ac4e4bc1dc25524a98a24cdeea/frontend/apps/reader/skimtowidget.lua#L424
https://github.com/koreader/koreader/blob/8d8b8cd0f79455ac4e4bc1dc25524a98a24cdeea/plugins/statistics.koplugin/main.lua#L680

yparitcher:
> I think something like `Document:getCurrentPage()` would probably make the most sense.

CreDocument already implements this.
the other Mupdf based ones don't, as the pages are managed in `ReaderPaging`. what would be the best way to get the `paging.current_page` value in `Document:`?

poire-z:
```lua
    if self.document.info.has_pages then
        curr_page = self.ui.paging.current_page
    else
        curr_page = self.document:getCurrentPage()
    end
```

> what would be the best way to get the paging.current_page value in Document:?

If you're asking, may be there's none obvious :) and that's why it's the way it is.
May be forcing it to be a method of Document is not the most logical way (haven't checked), even if it looks like it would in the snippets you mentionned.

(I'd say: don't bother "fixing" that for such a little PR :)

yparitcher:
I think i managed it. Let me know if this looks good or i should abandon the whole mess

frontend/apps/reader/readerui.lua
```diff
@@ -270,6 +270,10 @@ function ReaderUI:init()
             ui = self,
             document = self.document,
         })
+        -- Current page
+        self.getCurrentPage = function()
+          return self.paging.current_page
+        end
```
poire-z: For clarity, I think these could just be regular methods: `ReaderUI:getCurrentPage()` with a `if self.document.info.has_pages then` in it (like we have everywhere).
Just so we find it where we expect to find it, instead of having to think about a dynamically plugged function.

(And you call it like a methods, while it really is an unbound function.)

yparitcher: done

yparitcher: what is the difference between lua methods and functions?

poire-z:
```
Object.function(arg1, arg2)
Object:method(arg1, arg2)
```
The `:` implicitely defines self as being expected as the 1st argument, and when you call `instance:method(arg1, arg2)`, it implicitely passes `instance` as `self` in the call.

In the definitions, these are equivalents, but the 1st is more readable:
```
Object:method(arg1, arg2)
Object.function(self, arg1, arg2)
```

yparitcher: thanks for that, lua is not my primary language :)

frontend/apps/reader/readerui.lua
```diff
@@ -747,4 +747,12 @@ function ReaderUI:switchDocument(new_file)
     self:showReader(new_file)
 end
 
+function ReaderUI:getCurrentPage()
```
poire-z: Was about to suggest to put it as a more appropriate place (that at the end), with some other alike functions... but there are none :)
So, indeed, it's not the module where we expect such functions... so let it be there.
May be just add a comment `-- little convenient helper`
(Or move if into ReaderView, where it would be more at home?)

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1300,6 +1304,10 @@ function ReaderGesture:gestureAction(action, ges)
         self.ui:handleEvent(Event:new("GotoNextChapter"))
     elseif action == "prev_chapter" then
         self.ui:handleEvent(Event:new("GotoPrevChapter"))
+    elseif action == "next_bookmark" then
+        self.ui:handleEvent(Event:new("GotoNextBookmarkFromPage", self.ui:getCurrentPage()))
+    elseif action == "prev_bookmark" then
+        self.ui:handleEvent(Event:new("GotoPreviousBookmarkFromPage", self.ui:getCurrentPage()))
```
poire-z: Still want to pass current page from there?

yparitcher: I dont care either way, i am just keeping to the style in `readerbookmark` where all the functions are passed a page.

poire-z: I prefer as I proposed in https://github.com/koreader/koreader/pull/5968#discussion_r392711103 (no need to pass a page if the function/event is named ...FromCurrentPage)

frontend/apps/reader/readerui.lua
```diff
@@ -320,6 +324,11 @@ function ReaderUI:init()
             ui = self
         })
         self.disable_double_tap = G_reader_settings:readSetting("disable_double_tap") ~= false
+        -- Current page
+        self.getCurrentPage = function()
+          return self.document:getCurrentPage()
```
Frenzie: nit

frontend/apps/reader/readerui.lua
```diff
@@ -270,6 +270,10 @@ function ReaderUI:init()
             ui = self,
             document = self.document,
         })
+        -- Current page
+        self.getCurrentPage = function()
+          return self.paging.current_page
```
Frenzie: nit

frontend/apps/reader/modules/readergesture.lua
yparitcher: done

frontend/apps/reader/skimtowidget.lua
```diff
@@ -421,11 +395,7 @@ function SkimToWidget:goToBookmark(page)
     if page then
         self:addOriginToLocationStack()
         self.ui.bookmark:gotoBookmark(page)
-        if self.document.info.has_pages then
-            self.curr_page = self.ui.paging.current_page
-        else
-            self.curr_page = self.document:getCurrentPage()
-        end
+        self.curr_page  = self.ui:getCurrentPage()
```
Frenzie: more nit :-P

```suggestion
        self.curr_page = self.ui:getCurrentPage()
```

plugins/statistics.koplugin/main.lua
```diff
@@ -677,11 +677,7 @@ function ReaderStatistics:getStatisticEnabledMenuItem()
                 self:initData()
                 self.pages_stats = {}
                 self.start_current_period = TimeVal:now().sec
-                if self.document.info.has_pages then
-                    self.curr_page = self.ui.paging.current_page
-                else
-                    self.curr_page = self.document:getCurrentPage()
-                end
+                self.curr_page  = self.ui:getCurrentPage()
```
Frenzie:
```suggestion
                self.curr_page = self.ui:getCurrentPage()
```

PR #5980, opened 2020-03-19
bump crengine: support font-variant, more FB2 metadata

### LINKED ISSUE #5821: Enhancement proposed: to display Open Type layout features.
* KOReader version: all
It's about making use of the _font-variant_ property (or font-variant-_something_ property) to display some or all alternate properties of Open-Type fonts. 

This is a feature Calibre (editor and viewer), and Sigil already display. It's accepted by ePubcheck. It lacks support among commercial ebook-viewers.

Rationale

Implement the use of some OT features like _oldnum_ or _hlig_ which have  a gracious fallback . Users with an advanced viewer like Koreader could use it, and nothing would change for other users (like it happens today with in-page note display).

Here is an example with a joint ePub3 book which makes use of this "oldnum" OT property. You can see it displayed properly with Calibre viewer and/or Sigil. The OT font of this ePub has been subsetted with the relevant Calibre editor tool and the SubsetFont plugin of Sigil.

The font used is a derivative of Linux Libertine.

[villars.zip](https://github.com/koreader/koreader/files/4165093/villars.zip)

### LINKED ISSUE #5956: Suggestions: show annotation in FM on long tap
* KOReader version: from new release
* Device: Kindle PW3 (5.9.7)

It would be nice if there was an option to read a book's annotation (for epub, fb2 at least) via File Manager's long tap on a book. It would make it easier to choose a book without opening and rendering it.

What do you think guys?

poire-z:
Includes https://github.com/koreader/crengine/pull/334
- FB2: merge other metadata in returned keywords and description. Closes #5956.
- CSS: adds support for font-variant. Closes #5821.

Also:
- Rebase MD5 module against upstream https://github.com/koreader/koreader-base/pull/1065
- Update cURL to 7.69.1 https://github.com/koreader/koreader-base/pull/1066

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5980)


poire-z:
Would it be worth adding this style tweak? :
`Disable ligatures (in kerning mode 'best')` `* { font-variant: no-common-ligatures; }`

(I had a moment when I found ligatures ugly - it has passed :) - and tried this, and it works.)

If yes, where:
<kbd>image</kbd>
Would a sub-menu `Hyphenation and ligatures>` be ok, to not make this one taller?

(And the more we add features that work only well in "best" kerning mode, the hacky "good" mode feels a bit less needed - vs "fast" that might be good enough for most people that need speed. And when adding features to crengine, I always have a moment I have to think about that mode, and wonder how to make that thing work - often ending not having it work in that mode (font-variants, HB font languages...), and often not testing much it :) So, I don't know, may be we could remove it ?)

mergen3107:
I use “good” most of the time due to speed/quality balance. “Fast” is really fast, but some parts of text look a little out of place. “Best” is some times very slow for large books (1000+ CR pages)

poire-z:
OK, so let's keep "good" :)
(Could that be specific to cyrillic ? I have never really noticed any issue with "fast" with latin letters text.)

NiLuJe:
IIRC, "fast" only does legacy kerning (i.e, kern tables vs. GPOS tables), which might indeed be an issue with some fonts ;).

mergen3107:
Unfortunately, I can't remember or reproduce exact reason I didn't like `fast` so much. 
One thing for sure if that if has different spaces between small words, but this effect is insignificant (e.g. produces 370 CR pages instead of 368 in `good` rendering):
`fast`:
Reader_2020-Apr-10_175203
`good`:
Reader_2020-Apr-10_175223

I am sure about Cyrillic-specific deterioration neither. `good`'s description explicitly warns about lack of support for non-western characters anyway :D

mergen3107:
This is on KOReader Stable v2020.03.2 on Kindle PW3 (5.9.7)

PR #5982, opened 2020-03-20
Add DJVU MIME type to OPDS browser

avsej:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5982)


jstirling:
Is this the proper MIME type for djvu? I've never used it before but when I searched around I found "image/vnd.djvu".

NiLuJe:
A quick look at `/etc/mime.types` and the F.D.O MIME thingy would indeed seem to concur.

@avsej: What happened to be service those as application/djvu?

Frenzie:
What @NiLuJe wrote is also why it says this elsewhere:

https://github.com/koreader/koreader/blob/e9d0c7a96e0040dd031e4f80a23a30fcf52482a3/frontend/document/djvudocument.lua#L148-L149

PR #5984, opened 2020-03-20
Highlights: turn page when start or end crosses pages

### LINKED ISSUE #5973: Blind highlights editing at the edge
* Device: Kindle PW3

#### (minor UX) Issue
I was pleased to find out that making a highlight across two pages is so convenient! After dragging to the next page, showed text shifts further a little to show highlighting pieces. 

However, editing such two-page highlight is not so obvious.

For example (in a fb2.zip book):
#### Steps to reproduce
1) Start highlighting from a bottom right corner on a page, drag to the next page. 
2) It will allow to select text on the second page just fine, and return to showing the first page. 
3) Now, tap this highlight, move the starting point to the right. 
4) It will de-select highlight area fine, but once it is beyond this page, it doesn’t turn to the next page. So further de-highlighting happens blindly, although correctly. 

Maybe it should detect where this highlight is interrupted by a page break? And then turn the page accordingly?

poire-z:
More generally: switch to the page that contains the highlight edge that we are currently moving.
Closes #5973.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5984)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -329,6 +329,26 @@ function ReaderHighlight:updateHighlight(page, index, side, direction, move_by_c
         datetime = highlight_time,
         updated_highlight = new_highlight
     }, true)
+    if side == 0 then
+        -- Ensure we show the page with the new beginning of highlight
+        if not self.ui.document:isXPointerInCurrentPage(new_beginning) then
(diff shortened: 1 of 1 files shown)


arooni:
multi page highlighting never worked that well for me before, but you've gone and fixed it with https://github.com/koreader/koreader/pull/5984 .. i really have no idea how you can improve this app anymore but you folks keep finding ways!

PR #5989, opened 2020-03-21
[Kindle] More accurate on/off FL detection/toggle

### LINKED ISSUE #5986: Frontlight: doesn't remember previously used value
* Device: Kindle PW3

#### Issue
After exit, KOReader doesn't remember Frontlight value used before turning off.

#### Steps to reproduce
1) Kindle menu: set Frontlight (e.g. 10)
2) Open KOReader via KUAL
3) It sees the correct value of Frontlight (10), OK
4) Turn off Frontlight and try to toggle - resumes back to 10, OK
5) Turn off Frontlight via "Toggle" (in menu, or with gesture)
6) Exit KOReader
7) Return to KOReader via KUAL. Try to toggle Frontlight - it stays at 0, while gesture popup says "Frontlight off" each time.

So probably KOReader should remember the previously used value of Frontlight before exit?

NiLuJe:
Fix #5986 

Probably breaks a bunch of random and weird corner-cases (Auto FrontLight plugin?). Meh.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5989)

### DIFF
--- a/frontend/device/generic/powerd.lua
+++ b/frontend/device/generic/powerd.lua
@@ -19,7 +19,6 @@ function BasePowerD:new(o)
     if o.init then o:init() end
     if o.device and o.device:hasFrontlight() then
         o.fl_intensity = o:frontlightIntensityHW()
-        print("BasePowerD:new", o.fl_intensity)
         o:_decideFrontlightState()
     end
     return o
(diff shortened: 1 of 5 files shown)


NiLuJe:
Sidebar: our peculiar handling of "0 == off" on <= PW3 effectively renders the lowest "on" step (0) unusable in KOReader, since we coopt it to mean off instead of whatever it does on its own.

That's visible when you suspend/resume, as the system will restore the light to its own interpretation of "0" before we stomp it back to "off" ourselves.

On those devices, our own lowest "on" step is "1" instead (and it matches the system's 1 insead of the system's 0).

Just mentioning it for posterity's sake. I don't intend to fix it, and I wouldn't really know how best to go about it without a bunch of madness-inducing indirection layers like on Kobo.

NiLuJe:
@hius07: You mentioned that the PW3 was the last device where 0 wasn't off for the FL. Does that apply to the Oasis, too (chronologically, it should *not*, as the Oasis was released later), but I honestly don't remember ;).

NiLuJe:
Random drive-by fix for the fbink part of https://github.com/koreader/koreader/pull/5758

I'd actually forgot to update the tar progress bar calls to point to the fbink copy >_<".

PR #5999, opened 2020-03-22
stardict path override

### LINKED ISSUE #5995: Make dictionaries folder location configurable
* KOReader version: 2020.02 (Android)
* Device: Onyx Boox Note

#### Issue

Enhancement proposal:

koreader uses its own private location for dictionaries at koreader/data/dict as common storage for external stardict dictionaries. If users would be possible to reconfigure this storage location it would be able to share an already existing location between multiple reader applications. On the Onyx Boox devices there is typically an upper level folder named dicts used for storing stardict dictionaries for its internal applications. If users would be able to simply point koreader's dictionary location to that same location there would be no need anymore to maintain duplicate dictionary files on the device and save considerable storage capacity.

pazos:
Fixes #5995

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/5999)

### DIFF
--- a/defaults.lua
+++ b/defaults.lua
@@ -251,6 +251,10 @@ NETWORK_PROXY = nil
 -- Use turbo library to handle async HTTP request
 DUSE_TURBO_LIB = false
 
+-- Absolute path to stardict files (override)
+-- By default they're stored in data/dict under dataDir.
+STARDICT_DATA_DIR = nil
+
 -- ####################################################################
 -- following features are not supported right now
 -- ####################################################################
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -121,6 +121,7 @@ read_globals = {
     "KOBO_LIGHT_ON_START",
     "NETWORK_PROXY",
     "DUSE_TURBO_LIB",
+    "STARDICT_DATA_DIR",
     "cre",
     "lfs",
     "lipc",
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -101,7 +101,8 @@ end
 
 function ReaderDictionary:init()
     self.ui.menu:registerToMainMenu(self)
-    self.data_dir = os.getenv("STARDICT_DATA_DIR") or
+    self.data_dir = STARDICT_DATA_DIR or
+        os.getenv("STARDICT_DATA_DIR") or
         DataStorage:getDataDir() .. "/data/dict"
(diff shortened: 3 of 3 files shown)


defaults.lua
```diff
@@ -251,6 +251,10 @@ NETWORK_PROXY = nil
 -- Use turbo library to handle async HTTP request
 DUSE_TURBO_LIB = false
 
+-- Absolute path to stardict files.
+-- By default they're stored in data/dict under dataDir.
+STARDICT_DATA_DIR = nil
```
Frenzie: We should consider something like this to keep compatibility with system-wide sdcv configuration out of the box. Unless sdcv picks up on that regardless?
```suggestion
STARDICT_DATA_DIR = os.getenv("STARDICT_DATA_DIR")
```

pazos: I don't know:

```bash
pazos@bowie:/k$ ./sdcv --only-data-dir -l
Dictionary's name   Word count
pazos@bowie:/k$ export STARDICT_DATA_DIR=/test/dict/
pazos@bowie:/k$ ./sdcv --only-data-dir -l
Dictionary's name   Word count
pazos@bowie:/k$ ./sdcv --only-data-dir --data-dir=/test/dict -l
Dictionary's name   Word count
Oxford English Dictionary    75753
Diccionario Español RAE    86023
Diccionario Inglés-Español    20737
```

Frenzie: --only-data-dir overrides it

https://github.com/Dushistov/sdcv/blob/08461acab8e25dc3b1a596fc7146e079dcef3d64/src/sdcv.cpp#L131-L142

pazos: Sadly it's not possible to commit your suggestion as it clashes with advanced settings. One way to circumvent this is

```lua
@@ -101,7 +101,8 @@ end
 
 function ReaderDictionary:init()
-    self.data_dir = os.getenv("STARDICT_DATA_DIR") or
+    self.data_dir = STARDICT_DATA_DIR or
+        os.getenv("STARDICT_DATA_DIR") or
         DataStorage:getDataDir() .. "/data/dict"
```

and leave `STARDICT_DATA_DIR = nil`

Frenzie: That was my first thought actually, but I figured it might be slightly more elegant in defaults.lua.

But I suppose I was mixing logic and configuration, actually being inelegant.

tl;dr Yes please! :+1:

pazos: done!

Issue #5924, opened 2020-03-06
FR: News downloader - toggle "images download" in N.D. menu


mwoz123:
#### Issue
After last changes (downloading images, generating epub etc) processing all sources can take long time  and blocks the Koreader until finished.

How about quick option in menu - could be an item/"button" in Newsdownloder 
menu "(toogle?) images download"  
which will override defualt "download images" settings from feed_config.lua.

This menu item could display current downloading status:
"Images download : off" -> click on reverses to "on" 
"Images download: on" -> click on reverses to "off" 

This can be "global" option that enables download images only for feeds which have " include_images=true" in feed_config.lua. Shouldn't donwload images for other feed sources.

Can be useful when you're just about to leave home and have no time for full downloading all news with images and processing to epub.

mwoz123:
Better name might be, "force skip images download" but it's too long.
Maybe then "force skip images" or something similar

mwoz123:
The hole idea was to make it a quick way to sometimes disable images (e.g. slow wifi, no time for downloading all data, use smaller amount of wifi in case of being currently connected to limited transfer wifi etc).

I didn't want to make it a replacement for" include_images " which can be set for each rss/atom source independly.

PR #6003, opened 2020-03-25
5924 nd force skip imasges dl

mwoz123:
Solves: #5924

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6003)

### DIFF
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -29,7 +29,7 @@ local config_key_custom_dl_dir = "custom_dl_dir";
 local file_extension = ".epub"
 local news_download_dir_name = "news"
 local news_download_dir_path, feed_config_path
-local global_exlude_images_config_name = 'global_exlude_images_config_name';
+local global_skip_images_download_config_name = 'global_skip_images_download_config_name';
 
 -- if a title looks like <title>blabla</title> it'll just be feed.title
 -- if a title looks like <title attr="alb">blabla</title> then we get a table
@@ -110,8 +110,8 @@ function NewsDownloader:addToMainMenu(menu_items)
                 text = _("Global skip images download"),
                 keep_menu_open = true,
                 callback = function()
-                    local skip = news_downloader_settings:readSetting(global_exlude_images_config_name) or false;
-                    news_downloader_settings:saveSetting(global_exlude_images_config_name, not skip)
+                    local skip = news_downloader_settings:readSetting(global_skip_images_download_config_name) or false;
+                    news_downloader_settings:saveSetting(global_skip_images_download_config_name, not skip)
                     UI:info("Global skip images download set to %1", not skip)
                 end,
             },
@@ -185,7 +185,7 @@ function NewsDownloader:loadConfigAndProcessFeeds()
         return
     end
 
-    local global_exlude_images = news_downloader_settings:readSetting(global_exlude_images_config_name) or false
+    local global_exlude_images = news_downloader_settings:readSetting(global_skip_images_download_config_name) or false
 
     local unsupported_feeds_urls = {}
 
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -107,7 +107,7 @@ function NewsDownloader:addToMainMenu(menu_items)
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
             {
-                text = _("Toogle: force skip images download"),
+                text = _("Force skip images download"),
                 keep_menu_open = true,
                 callback = function()
                     local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
@@ -190,7 +190,7 @@ function NewsDownloader:loadConfigAndProcessFeeds()
         logger.err('NewsDownloader: empty feed list.', feed_config_path)
         return
     end
-    
+
     local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
     local force_exlude_images = news_downloader_settings:readSetting(force_skip_images_download_config_name) or false
 
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -107,12 +107,18 @@ function NewsDownloader:addToMainMenu(menu_items)
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
             {
-                text = _("Global skip images download"),
+                text = _("Force skip images download toggle"),
                 keep_menu_open = true,
                 callback = function()
+                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
                     local skip = news_downloader_settings:readSetting(global_skip_images_download_config_name) or false;
+                    logger.warn('NewsDownloader: global_skip_images_download', skip)
                     news_downloader_settings:saveSetting(global_skip_images_download_config_name, not skip)
-                    UI:info("Global skip images download set to %1", not skip)
+                    news_downloader_settings:flush()
+                    UIManager:show(InfoMessage:new{
+                        text = T(_("Global skip images download set to %1"),
+                        string.format("%s", not skip))
+                    })
                 end,
             },
             {
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -111,14 +111,13 @@ function NewsDownloader:addToMainMenu(menu_items)
                 keep_menu_open = true,
                 checked_func = function()
                     local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
-                    local skip = news_downloader_settings:readSetting(never_download_images_config_name)
-                    return skip
+                    return news_downloader_settings:readSetting(never_download_images_config_name)
                 end,
                 callback = function()
                     local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
-                    local skip = news_downloader_settings:readSetting(never_download_images_config_name) or false
-                    logger.info('NewsDownloader: previous never_download_images: ', skip)
-                    news_downloader_settings:saveSetting(never_download_images_config_name, not skip)
+                    local never_download_images = news_downloader_settings:readSetting(never_download_images_config_name) or false
+                    logger.info("NewsDownloader: previous never_download_images: ", never_download_images)
+                    news_downloader_settings:saveSetting(never_download_images_config_name, not never_download_images)
                     news_downloader_settings:flush()
                 end,
             },
@@ -193,7 +192,7 @@ function NewsDownloader:loadConfigAndProcessFeeds()
     end
 
     local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
-    local force_exlude_images = news_downloader_settings:readSetting(never_download_images_config_name) or false
+    local never_download_images = news_downloader_settings:readSetting(never_download_images_config_name) or false
 
     local unsupported_feeds_urls = {}
 
@@ -202,7 +201,7 @@ function NewsDownloader:loadConfigAndProcessFeeds()
         local url = feed[1]
         local limit = feed.limit
         local download_full_article = feed.download_full_article == nil or feed.download_full_article
-        local include_images = not force_exlude_images and feed.include_images
+        local include_images = not never_download_images and feed.include_images
         if url and limit then
             local feed_message = T(_("Processing %1/%2:\n%3"), idx, total_feed_entries, BD.url(url))
             UI:info(feed_message)
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -25,11 +25,11 @@ local initialized = false
 local wifi_enabled_before_action = true
 local feed_config_file_name = "feed_config.lua"
 local news_downloader_config_file = "news_downloader_settings.lua"
-local config_key_custom_dl_dir = "custom_dl_dir";
+local config_key_custom_dl_dir = "custom_dl_dir"
 local file_extension = ".epub"
 local news_download_dir_name = "news"
 local news_download_dir_path, feed_config_path
-local force_skip_images_download_config_name = 'force_skip_images_download_config_name';
+local never_download_images_config_name = "never_download_images_config_name"
 
 -- if a title looks like <title>blabla</title> it'll just be feed.title
 -- if a title looks like <title attr="alb">blabla</title> then we get a table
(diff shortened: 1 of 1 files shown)


plugins/newsdownloader.koplugin/main.lua
```diff
@@ -29,6 +29,7 @@ local config_key_custom_dl_dir = "custom_dl_dir";
 local file_extension = ".epub"
 local news_download_dir_name = "news"
 local news_download_dir_path, feed_config_path
+local global_exlude_images_config_name = 'global_exlude_images_config_name';
```
Frenzie:
```suggestion
local global_exlude_images_config_name = "global_exlude_images_config_name"
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -105,6 +106,15 @@ function NewsDownloader:addToMainMenu(menu_items)
                 keep_menu_open = true,
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
+            {
+                text = _("Global skip images download"),
```
Frenzie: If you reverse the logic you could just call it:

```suggestion
                text = _("Download images"),
```

Or maybe:
```suggestion
                text = _("Download images (default)"),
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -29,7 +29,7 @@ local config_key_custom_dl_dir = "custom_dl_dir";
 local file_extension = ".epub"
 local news_download_dir_name = "news"
 local news_download_dir_path, feed_config_path
-local global_exlude_images_config_name = 'global_exlude_images_config_name';
+local global_skip_images_download_config_name = 'global_skip_images_download_config_name';
```
Frenzie:
```suggestion
local global_skip_images_download_config_name = "global_skip_images_download_config_name"
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -107,12 +107,18 @@ function NewsDownloader:addToMainMenu(menu_items)
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
             {
-                text = _("Global skip images download"),
+                text = _("Force skip images download toggle"),
                 keep_menu_open = true,
                 callback = function()
+                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
                     local skip = news_downloader_settings:readSetting(global_skip_images_download_config_name) or false;
+                    logger.warn('NewsDownloader: global_skip_images_download', skip)
```
Frenzie:
```suggestion
                    logger.debug("NewsDownloader: global_skip_images_download", skip)
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -29,7 +29,7 @@ local config_key_custom_dl_dir = "custom_dl_dir";
 local file_extension = ".epub"
 local news_download_dir_name = "news"
 local news_download_dir_path, feed_config_path
-local global_skip_images_download_config_name = 'global_skip_images_download_config_name';
+local force_skip_images_download_config_name = 'force_skip_images_download_config_name';
```
Frenzie:
```suggestion
local force_skip_images_download_config_name = "force_skip_images_download_config_name"
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -107,16 +107,16 @@ function NewsDownloader:addToMainMenu(menu_items)
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
             {
-                text = _("Force skip images download toggle"),
+                text = _("Toogle: force skip images download"),
```
Frenzie: Typo, but I'm not entirely sure what the text means. I suspect that "toggle" shouldn't be there since it'd already be visually in the checkbox.
```suggestion
                text = _("Toggle: force skip images download"),
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -105,6 +106,21 @@ function NewsDownloader:addToMainMenu(menu_items)
                 keep_menu_open = true,
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
+            {
+                text = _("Force skip images download"),
+                keep_menu_open = true,
+                callback = function()
+                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
+                    local skip = news_downloader_settings:readSetting(force_skip_images_download_config_name) or false;
+                    logger.warn('NewsDownloader: previous force_skip_images_download: ', skip)
```
Frenzie:
```suggestion
                    logger.info("NewsDownloader: previous force_skip_images_download: ", skip)
```

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -105,6 +106,21 @@ function NewsDownloader:addToMainMenu(menu_items)
                 keep_menu_open = true,
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
+            {
+                text = _("Force skip images download"),
```
Frenzie: @NiLuJe Do you have any thoughts? I'm not overly fond of this string. (Setting aside that `Force skip image download` would be the more idiomatic form of it.)

NiLuJe: `Never download images`?

NiLuJe: Probably applies to a bunch of similar strings up the chain, then ;).

(and/or variable names if someone wants to go the extra mile and it makes sense re: the rest of the code).

Frenzie: Yes, my thoughts exactly (also see [here](https://github.com/koreader/koreader/pull/6003/commits/2bd831205941b6b7c5cc2310772890b4565d2e7b#r398161735) because I mistakenly thought the commits were distinct units). :-)
```suggestion
                text = _("Never download images"),
```

Frenzie: And yes, the variable name would also be much clearer as never_download_images

mwoz123: Guys IMHO  "never" means something that will burn down function (or even hardware? (if I would have been non-tech) :p) responsible for downloading images.
I would NEVER click that (if I didn't write that function):p
I'm serious. 

This is "temporarily disable", you can quickly reverse it and its gets back to normal downloading images not forever! 

Please rethink once again you request

mwoz123: How about " temporary skip images"?

Frenzie: That's because I overlooked that you made it scary, and it also explains the weird popup. ;-)



Should be like this instead:
```diff
diff --git a/plugins/newsdownloader.koplugin/main.lua b/plugins/newsdownloader.koplugin/main.lua
index 5f3921b1..debbaa64 100644
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -107,18 +107,20 @@ function NewsDownloader:addToMainMenu(menu_items)
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
             {
-                text = _("Force skip images download"),
+                text = _("Never download images"),
                 keep_menu_open = true,
+                checked_func = function()
+                    -- Not sure this should all be quite this local all over the file ;-)
+                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
+                    local skip = news_downloader_settings:readSetting(force_skip_images_download_config_name)
+                    return skip
+                end,
                 callback = function()
                     local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
                     local skip = news_downloader_settings:readSetting(force_skip_images_download_config_name) or false;
                     logger.warn('NewsDownloader: previous force_skip_images_download: ', skip)
                     news_downloader_settings:saveSetting(force_skip_images_download_config_name, not skip)
                     news_downloader_settings:flush()
-                    UIManager:show(InfoMessage:new{
-                        text = T(_("Force skip images download set to %1"),
-                        string.format("%s", not skip))
-                    })
                 end,
             },
             {
```

plugins/newsdownloader.koplugin/main.lua
Frenzie: But that whole never/don't stuff isn't necessary and harder to parse, better yet is simply reversed logic:

Screenshot_20200326_085725

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -105,6 +106,21 @@ function NewsDownloader:addToMainMenu(menu_items)
...
+                keep_menu_open = true,
+                callback = function()
+                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
+                    local skip = news_downloader_settings:readSetting(force_skip_images_download_config_name) or false;
```
Frenzie:
```suggestion
                    local skip = news_downloader_settings:readSetting(force_skip_images_download_config_name) or false
```

plugins/newsdownloader.koplugin/main.lua
mwoz123: The hole idea was to make it a quick way to sometimes disable images (e.g.  slow wifi, no time for downloading all data, use smaller amount of wifi in case of being currently connected to limited transfer wifi etc). 

I didn't want to make it a replacement for" include_images " which can be set for each rss/atom source independly.

And  the above screenshot looks this way for me that 's why I wanted to mark it as "skip" / "temporally" etc..

mwoz123: Btw I like the idea with checkbox (and would like to have it) but remembered that moving "lua Settings open" to lazy load was because there was a need/request to speed up menu. 
That's why I didn't use it..

Frenzie: Maybe you're thinking of something wifi-related? It already says `self:lazyInitialization()` in addToMainMenu so regardless how slow that may or may not be, the cost is incurred.

https://github.com/koreader/koreader/blob/0b898624acc8b332d44a16c95b45f68f4ccb1297/plugins/newsdownloader.koplugin/main.lua#L72-L73

mwoz123: Ok, how about checkbox + name that indicates enabling won't cause automatically downloading images from all sources (as it still has to be enabled in feed_config.lua)?

Frenzie: I have no problem with `Never download images`.

mwoz123: Ok checkbox + "never download images " is much safer will do so. 

BTW I'm not familiar with recent git changes. 
Although I haven't touch the majority of the PR request changes they are marked as 'outdated'. 
Do I need to changed them or you somehow patched it?

Frenzie: No, that just means I thought you submitted a PR with 4 distinct commits which I mistakenly reviewed independently.

mwoz123:
changes done.

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -105,6 +106,21 @@ function NewsDownloader:addToMainMenu(menu_items)
                 keep_menu_open = true,
                 callback = function() self:removeNewsButKeepFeedConfig() end,
             },
+            {
+                text = _("Never download images"),
+                keep_menu_open = true,
+                checked_func = function()
+                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
```
Frenzie: To be clear, I meant that the cost wouldn't be incurred if you didn't keep the settings excessively local. In this case there is an extra cost — not one that would be noticeable of course, but surely it should be more measurable than whatever profit you could possibly get out of that `never_download_images_config_name` thing.

I mean something like this:

```diff
diff --git a/plugins/newsdownloader.koplugin/main.lua b/plugins/newsdownloader.koplugin/main.lua
index 0a141c15..7695fcd3 100644
--- a/plugins/newsdownloader.koplugin/main.lua
+++ b/plugins/newsdownloader.koplugin/main.lua
@@ -25,6 +25,7 @@ local initialized = false
 local wifi_enabled_before_action = true
 local feed_config_file_name = "feed_config.lua"
 local news_downloader_config_file = "news_downloader_settings.lua"
+local news_downloader_settings
 local config_key_custom_dl_dir = "custom_dl_dir"
 local file_extension = ".epub"
 local news_download_dir_name = "news"
@@ -110,11 +111,9 @@ function NewsDownloader:addToMainMenu(menu_items)
                 text = _("Never download images"),
                 keep_menu_open = true,
                 checked_func = function()
-                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
                     return news_downloader_settings:readSetting(never_download_images_config_name)
                 end,
                 callback = function()
-                    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
                     local never_download_images = news_downloader_settings:readSetting(never_download_images_config_name) or false
                     logger.info("NewsDownloader: previous never_download_images: ", never_download_images)
                     news_downloader_settings:saveSetting(never_download_images_config_name, not never_download_images)
@@ -153,7 +152,7 @@ end
 function NewsDownloader:lazyInitialization()
    if not initialized then
         logger.dbg("NewsDownloader: obtaining news folder")
-        local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
+        news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
         if news_downloader_settings:has(config_key_custom_dl_dir) then
             news_download_dir_path = news_downloader_settings:readSetting(config_key_custom_dl_dir)
         else
@@ -191,7 +190,6 @@ function NewsDownloader:loadConfigAndProcessFeeds()
         return
     end
 
-    local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
     local never_download_images = news_downloader_settings:readSetting(never_download_images_config_name) or false
 
     local unsupported_feeds_urls = {}
@@ -409,7 +407,6 @@ function NewsDownloader:setCustomDownloadDirectory()
     require("ui/downloadmgr"):new{
        onConfirm = function(path)
            logger.dbg("NewsDownloader: set download directory to: ", path)
-           local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))
            news_downloader_settings:saveSetting(config_key_custom_dl_dir, ("%s/"):format(path))
            news_downloader_settings:flush()

```

More for ease of reading than performance really, that `local news_downloader_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), news_downloader_config_file))` is quite distracting.

plugins/newsdownloader.koplugin/main.lua
```diff
@@ -25,10 +25,11 @@ local initialized = false
 local wifi_enabled_before_action = true
 local feed_config_file_name = "feed_config.lua"
 local news_downloader_config_file = "news_downloader_settings.lua"
-local config_key_custom_dl_dir = "custom_dl_dir";
+local config_key_custom_dl_dir = "custom_dl_dir"
 local file_extension = ".epub"
 local news_download_dir_name = "news"
 local news_download_dir_path, feed_config_path
+local never_download_images_config_name = "never_download_images_config_name"
```
Frenzie: I find this unnecessarily complex. I don't follow the reasoning behind not to just using `readSetting("never_download_images")`, but setting that aside:
```suggestion
local never_download_images_config_name = "never_download_images"
```

mwoz123:
done

PR #6004, opened 2020-03-26
Adds ReaderPageMap, to optionally show source pages numbers

### LINKED ISSUE #4521: Epub page numbers
Hi. Was wondering if there's a way to display page numbers embedded in the epub file, instead of number of page turns. Maybe I cannot find the option

### LINKED ISSUE #4915: FR: Support EPUB3 navigation document
* KOReader version: 2019.04
* Device: Linux

#### Issue
Because Sigil does not automatically update toc.ncx when editing a EPUB3 file, many files I download online does not contain an up-to-date toc.ncx. It will be convenient if KOReader can read TOC from [navigation document](http://www.idpf.org/epub/301/spec/epub-contentdocs.html#sec-xhtml-nav).

Here is a sample file: [nav.zip](https://github.com/koreader/koreader/files/3076330/nav.zip)

### LINKED ISSUE #5991: Bookmarks don't work correctly in some rare cases
* KOReader version: Last one (Mars 21, 2020)
* Device: Kobo Mini

#### Issue
Bookmarks don't work correctly in the book I am reading when the bookmark is in the first page of a new chapter. The book is in epub format.
If I touch the upper right corner at the beginning of a chapter, the bookmark mark will appear in that corner of the page and the bookmark will appear in the bookmarks list.
If I touch again that corner to delete the mark and the bookmark, the mark will disappear, but the bookmark will continue appearing in the bookmarks list. The only way to delete the bookmark is from the bookmarks list.
If I bookmark the book at the beginning of a chapter and turn one page or several, and then I go back to the beginning of the bookmarked chapter, the bookmark mark won't be there although the bookmark will appear in the list of bookmarks.
If I bookmark any other page of the book that is not in the beginning of a chapter, bookmarks will work correctly with this book.

#### Steps to reproduce

Try the above on the extract of the book in epub format that I include here.
(Steps I followed to create t he extract:
The book is commercial and I cannot upload it here, so I unzipped the epub, extracted 2 whole chapters in html, deleted most of the text of these two chapters and created an epub using calibre editor. I tried that extract (a minibook of 3 pages) on my e-reader and the problem continues. So it must be something in the format of the beginning or end of the chapter.)

[Bookmark problem.zip](https://github.com/koreader/koreader/files/4362997/Bookmark.problem.zip)

poire-z:
bump crengine: support for EPUB3 nav toc and page maps
Includes https://github.com/koreader/crengine/pull/336
- Fix lvRect:isRectInside(rc) with 0-width or 0-height rect - closes #5991
- TOC: parse EPUB3 nav toc, fallback to spine when no toc - closes #4915
- Parse and cache various hardcopy page list maps
- epub.css: hide EPUB3 <span epub:type="pagebreak"> content
cre.cpp: add a few PageMap helper functions.

Adds ReaderPageMap which will add a new menu (under TOC and Bookmarks) that will allow:
- to list source page numbers (like a TOC)
- to show visible page labels in the right margin
- to use these source page numbers in the footer, the TOC, the GoTo and SkimTo widgets, and to use the source page number in the standard bookmark and highlight initial text.
- (and to unwontfix and close #4521)

It will not be often used (less than 5% of my publishers' EPUB have a page map/list), and most people will probably disable it when it appears, as it's rarely useful. It's just nice to have (for when you see in a book "See note page 48" and that references the hardcopy book, with no internal link to that note in the EPUB).

So, we'll get this (only when there is a pagemap in the book):
<kbd>image</kbd>
An option to use the source/hardcopy page number in the footer, TOC, Bookmark names, Goto and SkimTo widgets:
<kbd>image</kbd>
and one to show the page labels in the margin:
<kbd>image</kbd>

Opinions and suggestions welcome on the wording in the screenshots above for this feature and these page maps/lists I used "Source pages" because I'm not sure it's always reference to hardcopy paper book pages, but I dunno...

(still need to include the bump for https://github.com/koreader/koreader-base/pull/1070)

Also include:
`BookInfo: add File date, handle zip sub-suffix (.fb2.zip)` , as mentionned in https://github.com/koreader/koreader/issues/5956#issuecomment-603166021

Also include https://github.com/koreader/koreader-base/pull/1068: Update OpenSSL to 1.1.1e Update OpenSSH to 8.2p1

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6004)

### DIFF
--- a/frontend/apps/filemanager/filemanagerbookinfo.lua
+++ b/frontend/apps/filemanager/filemanagerbookinfo.lua
@@ -43,13 +43,27 @@ function BookInfo:show(file, book_props)
 
     local directory, filename = util.splitFilePathName(file)
     local filename_without_suffix, filetype = util.splitFileNameSuffix(filename) -- luacheck: no unused
+    if filetype:lower() == "zip" then
+        local filename_without_sub_suffix, sub_filetype = util.splitFileNameSuffix(filename_without_suffix) -- luacheck: no unused
+        sub_filetype = sub_filetype:lower()
+        local supported_sub_filetypes = { "fb2", "htm", "html", "log", "md", "txt" }
+
+        for __, t in ipairs(supported_sub_filetypes) do
+            if sub_filetype == t then
+                filetype = sub_filetype .. "." .. filetype
+                break
+            end
+        end
+    end
     local file_size = lfs.attributes(file, "size") or 0
+    local file_modification = lfs.attributes(file, "modification") or 0
     local size_f = util.getFriendlySize(file_size)
     local size_b = util.getFormattedSize(file_size)
     local size = string.format("%s (%s bytes)", size_f, size_b)
     table.insert(kv_pairs, { _("Filename:"), BD.filename(filename) })
     table.insert(kv_pairs, { _("Format:"), filetype:upper() })
     table.insert(kv_pairs, { _("Size:"), size })
+    table.insert(kv_pairs, { _("File date:"), os.date("%Y-%m-%d %H:%M:%S", file_modification) })
     table.insert(kv_pairs, { _("Directory:"), BD.dirpath(filemanagerutil.abbreviate(directory)) })
     table.insert(kv_pairs, "----")
 
--- a/frontend/apps/reader/modules/readerpagemap.lua
+++ b/frontend/apps/reader/modules/readerpagemap.lua
@@ -1,0 +1,418 @@
+local BD = require("ui/bidi")
+local Blitbuffer = require("ffi/blitbuffer")
+local CenterContainer = require("ui/widget/container/centercontainer")
+local Device = require("device")
+local Font = require("ui/font")
+local FrameContainer = require("ui/widget/container/framecontainer")
+local Geom = require("ui/geometry")
+local GestureRange = require("ui/gesturerange")
+local InputContainer = require("ui/widget/container/inputcontainer")
+local Menu = require("ui/widget/menu")
+local MultiConfirmBox = require("ui/widget/multiconfirmbox")
+local OverlapGroup = require("ui/widget/overlapgroup")
+local TextBoxWidget = require("ui/widget/textboxwidget")
+local TextWidget = require("ui/widget/textwidget")
+local UIManager = require("ui/uimanager")
+local Screen = Device.screen
+local T = require("ffi/util").template
+local _ = require("gettext")
+
+local ReaderPageMap = InputContainer:new{
+    label_font_face = "ffont",
+    label_default_font_size = 14,
+    -- Black so it's readable (and non-gray-flashing on GloHD)
+    label_color = Blitbuffer.COLOR_BLACK,
+    show_page_labels = nil,
+    use_page_labels = nil,
+    _mirroredUI = BD.mirroredUILayout(),
+}
+
+function ReaderPageMap:init()
+    self.has_pagemap = false
+    self.container = nil
+    self.max_left_label_width = 0
+    self.max_right_label_width = 0
+    self.label_font_size = G_reader_settings:readSetting("pagemap_label_font_size")
+                                or self.label_default_font_size
+    self.use_textbox_widget = nil
+    self.initialized = false
+    self.ui:registerPostInitCallback(function()
+        self:_postInit()
+    end)
+end
+
+function ReaderPageMap:_postInit()
+    self.initialized = true
+    if self.ui.document.info.has_pages then
+        return
+    end
+    if not self.ui.document:hasPageMap() then
+        return
+    end
+    self.has_pagemap = true
+    self:resetLayout()
+    self.ui.menu:registerToMainMenu(self)
+    self.view:registerViewModule("pagemap", self)
+end
+
+function ReaderPageMap:resetLayout()
+    if not self.initialized then
+        return
+    end
+    if self[1] then
+        self[1]:free()
+        self[1] = nil
+    end
+    if not self.show_page_labels then
+        return
+    end
+    self.container = OverlapGroup:new{
+        dimen = Screen:getSize(),
+        -- Pages in 2-page mode are not mirrored, so we'll
+        -- have to handle any mirroring tweak ourselves
+        allow_mirroring = false,
+    }
+    self[1] = self.container
+
+    -- Get some metric for label min width
+    self.label_face = Font:getFace(self.label_font_face, self.label_font_size)
+    local textw = TextWidget:new{
+        text = " ",
+        face = self.label_face,
+    }
+    self.space_width = textw:getWidth()
+    textw:setText("8")
+    self.number_width = textw:getWidth()
+    textw:free()
+    self.min_label_width = self.space_width * 2 + self.number_width
+end
+
+function ReaderPageMap:onReadSettings(config)
+    local h_margins = config:readSetting("copt_h_page_margins") or
+        G_reader_settings:readSetting("copt_h_page_margins") or
+        DCREREADER_CONFIG_H_MARGIN_SIZES_MEDIUM
+    self.max_left_label_width = Screen:scaleBySize(h_margins[1])
+    self.max_right_label_width = Screen:scaleBySize(h_margins[2])
+
+    self.show_page_labels = config:readSetting("pagemap_show_page_labels")
+    if self.show_page_labels == nil then
+        self.show_page_labels = G_reader_settings:nilOrTrue("pagemap_show_page_labels")
+    end
+    self.use_page_labels = config:readSetting("pagemap_use_page_labels")
+    if self.use_page_labels == nil then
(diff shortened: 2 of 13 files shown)


frontend/apps/reader/modules/readerpagemap.lua
```diff
@@ -0,0 +1,419 @@
...
+                    self.ui.doc_settings:saveSetting("pagemap_use_page_labels", self.use_page_labels)
+                    -- Reset a few stuff that may use page labels
+                    self.ui.toc:resetToc()
+                    self.ui.view.footer:updateFooter()
+                    UIManager:setDirty(self.view.dialog, "partial")
+                end,
+                hold_callback = function(touchmenu_instance)
+                    local use_page_labels = G_reader_settings:isTrue("pagemap_use_page_labels")
+                    UIManager:show(MultiConfirmBox:new{
+                        text = use_page_labels and _("The default (★) for newly opened books that have a page map is to use these page labels instead of the renderer page numbers.\n\nWould you like to change it?")
```
Frenzie: Not ecstatic about "renderer page numbers" but atm I can't think of a good alternative.

poire-z: Yeah, this one was hard.
I wanted to have the buttons "Renderer pages" | "Source pages (★)" but these got truncated...
So, we could end up with a sentence and Disable | Enable.

poire-z:
So, no particular thought against calling that/them "Source pages" ?

Frenzie:
Oh, I don't like that one bit. :-P

What about something along the lines of basic page numbers and advanced page numbers?

NiLuJe:
Reference page number?

Or, let's finally call it what it actually is: screen number vs. page number?

poire-z:
> Or, let's finally call it what it actually is instead of a stupid skeuomorphism: screen number vs. page number?

No, our page numbers are page numbers - renderer page numbers.
And most books won't have that menu - and we won't switch all the naming to mention "screen page numbers" :)
So, it's these source/reference/hardcopy page numbers that need the qualification/adjective.

> basic page numbers and advanced page numbers

None of each is either basic or advanced. These "source page numbers", althought sometimes useful, are more bothering (jumps, gaps, multiples consecutive on the same screen page...)

> Reference page number?

That could work I guess.

NiLuJe:
> No, our page numbers are page numbers - renderer page numbers.

But isn't a renderer page number technically one screen's height?

poire-z:
Yes, technically, in page mode, we show one page per screen :) (even more technically, and out of this context, pages can be smaller than a screen height, with page breaks between chapters - so, when switching to scroll mode, you can have 2 or 3 renderer pages in the screen view).

Just insisting I'd like to keep these ones called simply "page numbers" :)

poire-z:
Better ?:

(And where "page" should be plural, "pages" or not? Reference pages label?)

NiLuJe:
Plural mark on the rightmost noun of a compound is correct ;).

poire-z:
What about "Reference page list" vs "Reference pages list" ?

NiLuJe:
It's a single list of reference page numbers? (I'd tend to prefer adding an explicit 'number' mentioned somewhere there, so `Reference page numbers list`?)

poire-z:
<kbd>image</kbd>

The `Use reference page numbers` is a bit obtuse... It's supposed to mean "use reference page numbers instead of renderer page numbers where it might make sense" :) and it's not super perfectly implemented (in the SkimTo widget, it will show the ref page number as the number, it will allow you to input it in the GoTo input (including page numbers in latin like `xvii` - but taping on +10 or -1 will skip you 10 or -1 rendered pages... but well, I'll fix that when someone complain :)

Frenzie:
> None of each is either basic or advanced. These "source page numbers", althought sometimes useful, are more bothering (jumps, gaps, multiples consecutive on the same screen page...)

By basic I mean 1-99 without Latin, letters or whatever. As such we have some slight miscommunication here in that you can have "advanced" page numbers perfectly fine without jumps or gaps.

Anyway, I said *along the lines of* for a reason. ^_^ I don't literally mean the word advanced. I mean the general concept or association of advanced, complete, proper, thorough, paper-like, etc.

> Or, let's finally call it what it actually is instead of a stupid skeuomorphism: screen number vs. page number?

In paged media the distinction isn't between screens and pages but in the fact that page 1 doesn't start till page 20, before that it's page i-ixx, a-p, or whatever. That's basically my angle here. Dumb/basic Arabic page numbers vs labeled page numbers. Ideally the phrasing is compatible with both labeled paper-like page numbers in EPUB and labeled DjVu/PDF page names/numbers.

We can also do it (partially) without words, something like

* Page number format: 1-99
* Page number format: labeled

Frenzie:
Screenshot from LibreOffice:
Screenshot_2020-03-26_21-26-22

poire-z:
But you can have source/reference page numbers that are also 1-99, starting at 1 without any roman numbers, contiguous, increasing with no gap.
Distinguising them by how they look feels artificial. Better by where they come from (renderer vs reference/source/publisher).

(You cross ref'ed the issues with PDF & DJVU page numbers - but I have no idea how/if the code I added could be of use with them - probably simpler as for them, there is a one-to-one mapping between renderer pages vs reference/source page numbers.)

Frenzie:
> But you can have source/reference page numbers that are also 1-99, starting at 1 without any roman numbers, contiguous, increasing with no gap.

What are they called on Kindle or Kobo?

mergen3107:
In Kindle there are three types of positions. 
1) “Locations” are proprietary Amazon-calculated values, basically like a cumulative sum of all characters on a “page” (current screen). Location step depends on font size
2) “Pages” are present only if supporting apnx file is provided for this book. This is just “conversion” of how many locations should be in a pribt version of this book. 
3) “html-positions” which a raw data extracted from the book.

Frenzie:
@poire-z 
> But you can have source/reference page numbers that are also 1-99, starting at 1 without any roman numbers, contiguous, increasing with no gap.

PS In DjVu/PDF it's often as simple as the cover not being included in numbering.

@mergen3107 
Thanks! Sounds like there's nothing of interest to be gained there. ;-)

poire-z:
> What are they called on Kindle or Kobo?

No idea :)

But in the various mobiread links given in #4521:
https://wiki.mobileread.com/wiki/NCX  _1 for 1 mapping to the actual page numbers in the physical book_
https://wiki.mobileread.com/wiki/Adobe_Digital_Editions#Page-map _mapping between the hardcopy book page numbers and the eBook_
https://www.mobileread.com/forums/showthread.php?t=243222 _the start of the real page in the matching printed book._
https://wiki.mobileread.com/wiki/Page_numbers  _Absolute page numbers_ (I don't like it :) _The page numbers to be used in eBooks could be aligned with paper book versions of the same book for easy referencing_

I'm fine with "Reference", which is a little less absolute than "Absolute" :)

Frenzie:
I think "page labels" might be clearer than "reference page numbers," but maybe I'm just saying that because of PDF.

frontend/apps/filemanager/filemanagerbookinfo.lua
```diff
@@ -43,13 +43,27 @@ function BookInfo:show(file, book_props)
 
     local directory, filename = util.splitFilePathName(file)
     local filename_without_suffix, filetype = util.splitFileNameSuffix(filename) -- luacheck: no unused
+    if filetype:lower() == "zip" then
+        local filename_without_sub_suffix, sub_filetype = util.splitFileNameSuffix(filename_without_suffix) -- luacheck: no unused
+        sub_filetype = sub_filetype:lower()
+        local supported_sub_filetypes = { "fb2", "htm", "html", "log", "md", "txt" }
+
+        for __, typ in ipairs(supported_sub_filetypes) do
```
Frenzie: typ or type?

poire-z: "type" on first intention, but it got highlighted by vim, as `type("abc")` is a lua builtin function.
(It works with "type", but it feels strange to have it highlighted.)
So, I went for `typ`. I can use `t` if we'd prefer to not have orthograph errors :)

Frenzie: I think I prefer t, lol (but whichever)

frontend/apps/reader/modules/readerpagemap.lua
```diff
@@ -0,0 +1,416 @@
...
+        end
+    end
+end
+
+function ReaderPageMap:addToMainMenu(menu_items)
+    menu_items.page_map = {
+        text = _("Reference pages"),
+        sub_item_table ={
+            {
+                text = _("Reference source info"),
```
Frenzie: I think we might want to throw in some `@translators` notes. A reference source just sounds like a dictionary or something.

poire-z: OK, added 2 - feels a bit excessive to add more... Hopefully, translators will meet all of them near each others and will know what it's about.

PR #6014, opened 2020-03-29
blanked space

ayush855-cloud:
finally done

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6014)

PR #6015, opened 2020-03-29
blanked space which may lengthy your code

ayush855-cloud:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6015)


Frenzie:
See https://github.com/koreader/koreader/pull/6014#issuecomment-605682222

PR #6031, opened 2020-04-07
Fix remarkable pen input

### LINKED ISSUE #6030: Remarkable: Pen Input Broken
* KOReader version: 2020.03.2
* Device: ReMarkable

#### Issue

The pen does not work as an input device.  Either touching the screen with the pen has no result or, depending on where on the screen the pen makes contact, seemingly random behavior occurs.

#### Steps to reproduce

Start koreader and touch the screen with the pen.

dfhall:
This fixes issue #6030 and allows the ReMarkable pen to work as a touch input device.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6031)

### DIFF
--- a/frontend/device/remarkable/device.lua
+++ b/frontend/device/remarkable/device.lua
@@ -22,19 +22,36 @@ local ABS_Y = 01
 local ABS_MT_POSITION_X = 53
 local ABS_MT_POSITION_Y = 54
 -- Resolutions from libremarkable src/framebuffer/common.rs
+local screen_width = 1404 -- unscaled_size_check: ignore
+local screen_height = 1872 -- unscaled_size_check: ignore
+local wacom_width = 15725 -- unscaled_size_check: ignore
+local wacom_height = 20967 -- unscaled_size_check: ignore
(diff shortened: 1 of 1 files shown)


NiLuJe:
I was going to say: shouldn't this be an actual if ladder (i.e., if ... elseif .. elseif ... instead of a bunch of separate ifs), but apparently frontend/device/input.lua does the same thing, so, eh. :D.

frontend/device/remarkable/device.lua
```diff
@@ -22,20 +22,37 @@ local ABS_Y = 01
 local ABS_MT_POSITION_X = 53
 local ABS_MT_POSITION_Y = 54
 -- Resolutions from libremarkable src/framebuffer/common.rs
+local screen_width = 1404
```
Frenzie, line 25:
```suggestion
local screen_width = 1404 -- unscaled_size_check: ignore
```
For all four of them of course. ;-)

dfhall: Done.

Issue #6026, opened 2020-04-06
Default file browser folder is wrong


mirh:
* KOReader version: 2020.03.2
* Device: Tolino Vision 4 HD (on android 4.4.2)

#### Issue
Title. Not sure if it's specific to my device, or all androids. 
Probably regressed after 87334801ea250c1a77304685e0d82685b08bab09

#### Steps to reproduce
Start the program for the first time
Instead of the sdcard you are greeted with /data/data/org.koreader.blalbla/

pazos:
Cannot reproduce. Could you please attach the log file?

mirh:
:/ 
I cannot reproduce this myself either. 
I deleted both the program data folder, and the koreader folder on the /sdcard. 

Perhaps having upgraded from ics to kitkat caused this? Is it possible that if the previously saved "home folder" is wrong/missing, the fallback is to /data?

mergen3107:
Did you have your old config file before the update?
Maybe you accidentally set default home folder before?

mirh:
I did some more tests, and if I could confirm my later theory. 
If I corrupt the value of the home directory (and/or the last viewed folder) inside the lua settings file in /sdcard/koreader, it falls back to /data/data/whatever.

pazos:
> "home folder" is wrong/missing, the fallback is to /data?

Indeed, setting a home directory other than default will override [path](https://github.com/koreader/koreader/blob/master/platform/android/llapp_main.lua#L16). If the directory doesn't exist it won't be opened so `chdir` won't do anything. [android.dir](https://github.com/koreader/koreader/blob/master/platform/android/llapp_main.lua#L38) is always stored in private storage partition, under `/data`. That's not a bug, but a limitation of the platform.

mirh:
?? Shouldn't you fallback again on /sdcard if the value is wrong, and only then as a last desperate chance go into private storage?

mergen3107:
Maybe something related to mount order? For example, when /sdcard has not been mounted yet, KOReader tries to reach it?

mirh:
I'm not wondering *why* this happened (as I said, I updated from 4.0 to 4.4, and this may have slightly changed mount points). 
I'm saying that it shouldn't have happened. 
Surely it shouldn't be impossible to be more cautious on opening, or catch a notfound exception?

pazos:
> Shouldn't you fallback again on /sdcard if the value is wrong

The issue is not specific for android. It will happen on any other platform, for example:

[...]

PR #6035, opened 2020-04-10
Gestures: add action "Statistics calendar view"

poire-z:
Add a gesture action to show Statistics calendar view - because I realised I was hitting that menu too often :)
Also flush statistics to DB when showing it, so we see the last minutes/hours reading in the hourly histogram.

Also: File browser: sane fallback when home_dir invalid or missing - see https://github.com/koreader/koreader/issues/6026#issuecomment-610856059

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6035)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -680,14 +680,7 @@ end
 
 function FileManager:goHome()
     local home_dir = G_reader_settings:readSetting("home_dir")
-    if home_dir then
-        -- Jump to the first page if we're already home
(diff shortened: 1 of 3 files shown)


poire-z:
In this ? :
```lua
    ReaderGesture.getCalendarView = function()
        return self:getCalendarView()
    end
```
It's similar to the previous function, where `self` is the upvalue/closure passed to/from `function ReaderStatistics:init()`.
Anyway, it works for me in Reader and File browser, so i guess it's fine :)

PR #6040, opened 2020-04-11
desktop: disable battery on touch menu and reader footer

pazos:
The emulator will still show the battery.

## Emulator

emulator

## Desktop

desktop

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6040)


pazos:
> Btw, while probably not useful, a laptop can display battery status

Probably not useful for desktop OS, but could be cool for some future gnome/kde tablet. I will merge this PR until then. :+1:

PR #6052, opened 2020-04-14
Statistics: fix md5 leaking between newly opened books

poire-z:
Which could cause a same book appearing multiple times as different entries in statistics.
More details in https://github.com/koreader/koreader-base/pull/1077#issuecomment-613353913

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6052)

### DIFF
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1962,7 +1962,7 @@ function ReaderStatistics:saveSettings()
 end
 
 function ReaderStatistics:onReadSettings(config)
-    self.data = config.data.stats
+    self.data = config.data.stats or {}
 end
 
 function ReaderStatistics:onReaderReady()


poire-z:
I think this might have introduced another minor issue:

When I highlight some text, this is logged:
`05/16/20-10:28:05 ERROR failed to call event handler onAddHighlight plugins/statistics.koplugin/main.lua:1913: attempt to perform arithmetic on field 'highlights' (a nil value)`

Do you get that too? possibly only on documents opened this last?

Probably because since this PR, these fields are not set:
https://github.com/koreader/koreader/blob/58a6b01d039d3c383ea2f2a0eda76cec85eb30ac/plugins/statistics.koplugin/main.lua#L1912-L1926
(and I think previously, these were leaking from the previous document too).

So, we could just check here if they don't exist and set them to 0.
Or init them to 0 in (like all these other fields):
https://github.com/koreader/koreader/blob/58a6b01d039d3c383ea2f2a0eda76cec85eb30ac/plugins/statistics.koplugin/main.lua#L185-L205

Or just ignore them when not set.

Or just remove these handlers?
Do we care about them? On a quick look, it seems they are updated - but this self.data is mostly pre-SQL stuff, and might not need to be set/updated. Althought some other code might want to access them (but I guess that other code does wrong then :)
Also, the fact that the handlers in the first snippet above `return true` may prevent these events to reach other plugins and widgets. Any reason for that?

Any opinion on how to quick fix (without much thinking :) this minor logged error?
Pining @robert00s

poire-z:
Solution to the above issue made into #6173.

PR #6067, opened 2020-04-17
Add export to JSON option in Evernote plugin. Sort requires in Evernote plugin sources.

RCoeurjoly:
Tested in a Boox Max Carta.

The purpose (for me) of having JSON formatted clippings is that they are easy to parse, and then convert to org mode format, which is the format I am interested in.

I tried exporting html and converting to org with pandoc, but the result didn't please me.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6067)

### DIFF
--- a/plugins/evernote.koplugin/main.lua
+++ b/plugins/evernote.koplugin/main.lua
@@ -16,6 +16,7 @@ local _ = require("gettext")
 local N_ = _.ngettext
 local slt2 = require('slt2')
 local MyClipping = require("clip")
+local json = require("json")
 local realpath = require("ffi/util").realpath
 
 local EvernoteExporter = InputContainer:new{
@@ -32,6 +33,7 @@ local EvernoteExporter = InputContainer:new{
 
 function EvernoteExporter:init()
     self.text_clipping_file = self.clipping_dir .. "/KOReaderClipping.txt"
+    self.json_clipping_file = self.clipping_dir .. "/KOReaderClipping.json"
     local settings = G_reader_settings:readSetting("evernote") or {}
     self.evernote_domain = settings.domain
     self.evernote_username = settings.username or ""
@@ -44,13 +46,18 @@ function EvernoteExporter:init()
     self.html_export = settings.html_export or false
     self.joplin_export = settings.joplin_export or false
     self.txt_export = settings.txt_export or false
-    --- @todo Is this if block necessarry? Nowhere in the code they are assigned both true.
+    self.json_export = settings.json_export or false
+    --- @todo Is this if block necessary? Nowhere in the code they are assigned both true.
     -- Do they check against external modifications to settings file?
 
     if self.html_export then
         self.txt_export = false
         self.joplin_export = false
+        self.json_export = false
     elseif self.txt_export then
+        self.joplin_export = false
+        self.json_export = false
+    elseif self.json_export then
         self.joplin_export = false
     end
 
@@ -73,6 +80,7 @@ function EvernoteExporter:readyToExport()
     return self.evernote_token ~= nil or
             self.html_export ~= false or
             self.txt_export ~= false or
+            self.json_export ~= false or
             self.joplin_export ~= false
 end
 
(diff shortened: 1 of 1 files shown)


plugins/evernote.koplugin/main.lua
```diff
@@ -1,22 +1,23 @@
 local BD = require("ui/bidi")
-local InputContainer = require("ui/widget/container/inputcontainer")
-local LoginDialog = require("ui/widget/logindialog")
-local InfoMessage = require("ui/widget/infomessage")
-local NetworkMgr = require("ui/network/manager")
+local DEBUG = require("dbg")
```
Frenzie, line 2: I said you didn't have to change these, but that DEBUG isn't really the current style. ;-)
```suggestion
local dbg = require("dbg")
```

RCoeurjoly: I just figured it might help.

RCoeurjoly:
I converted to draft because I encountered some weird behavior while reading a pdf.
I am going to check if it was due to the wrong order of imports.
On a side note, I think it would be great to do the same checks that the CI does when building a apk.
Otherwise it builds something with errors.

NiLuJe:
Running the tests requires being able to.... run the tests ;).

Which is obviously not possible when cross-compiling. Ergo, you can only run the testsuite when doing a native emulator build (c.f., kodev's test command).

NiLuJe:
Given the scope of this PR, it is exceedingly unlikely to be causing a PDF issue.

Frenzie:
@RCoeurjoly 
Your previous PR was perfectly fine except for the minor whitespace issue with using two spaces instead of four. The require thing was just a parenthetical remark. Also because touching those lines will necessitate me to do a lot of nitpicking (like some of the naming schemes, the one require randomly using single quotes, etc.)

See https://github.com/koreader/koreader-base/wiki/Coding-style#naming and https://github.com/koreader/koreader-base/wiki/Coding-style#order-of-requires

A concrete example in practice:

https://github.com/koreader/koreader/blob/377bbe77ede168f120e16149a28a799450e38afb/frontend/ui/otamanager.lua#L5-L17

@NiLuJe 
> Which is obviously not possible when cross-compiling. Ergo, you can only run the testsuite when doing a native emulator build (c.f., kodev's test command).

I'd say that's not entirely true, the Android x86 binaries are usable in a run-of-the-mill Linux environment. Of course there's always a chance of minor oddities in that case, although generally I'd put my money on things miraculously unbreaking.

I rather doubt there's any point in doing that of course.

plugins/evernote.koplugin/main.lua
```diff
@@ -1,22 +1,23 @@
...
 local N_ = _.ngettext
-local slt2 = require('slt2')
-local MyClipping = require("clip")
+local NetworkMgr = require("ui/network/manager")
+local Screen = require("device").screen
+local T = require("ffi/util").template
+local UIManager = require("ui/uimanager")
+local json = require("json")
 local realpath = require("ffi/util").realpath
+local slt2 = require('slt2')
```
Frenzie, line 19:
```suggestion
local slt2 = require("slt2")
```

plugins/evernote.koplugin/main.lua
```diff
@@ -1,22 +1,23 @@
...
+local LoginDialog = require("ui/widget/logindialog")
+local MyClipping = require("clip")
 local _ = require("gettext")
 local N_ = _.ngettext
-local slt2 = require('slt2')
-local MyClipping = require("clip")
+local NetworkMgr = require("ui/network/manager")
+local Screen = require("device").screen
+local T = require("ffi/util").template
+local UIManager = require("ui/uimanager")
```
Frenzie, line 16: This one a little higher. ;-)

RCoeurjoly: Are there any issues blocking this PR?

Frenzie: Well yes, the requires. Like I wrote above:

> Your previous PR was perfectly fine except for the minor whitespace issue with using two spaces instead of four. The require thing was just a parenthetical remark. Also because touching those lines will necessitate me to do a lot of nitpicking (like some of the naming schemes, the one require randomly using single quotes, etc.)

RCoeurjoly: Let me know what you think now. I have removed all changes in requires apart from adding json library

PR #6072, opened 2020-04-19
Adds ReaderTypography (replaces ReaderHyphenation)

poire-z:
Replace Hyphenation menu with Typography menu.
This works mostly like before:
- typography/hyphenation is chosen according to the book metadata language,
- one can set a default or a fallback language,
- hyphenation is now just a subset of typography, and can be disabled while still setting a language and enabling the other features,
- the typography language enables newly added features to crengine (https://github.com/koreader/crengine/pull/337): per-language line breaking rules and per-language Harfbuzz glyph selection.
- by default (this can be disabled), we follow the books HTML `lang=` attributes to adapt typography/hyphenation/line breaking rules to which language the book says this section is in:
<kbd>image</kbd>

People who'd like some behaviour fixes/tweaks for their own usual languages can add or suggest stuff (alas to be hardcoded in the sources) in https://github.com/koreader/crengine/blob/master/crengine/src/textlang.cpp.

This should solve a few issues (even if we got no report for them :):
- german with its inverted quotations marks will get proper line breaking https://github.com/koreader/crengine/issues/307#issuecomment-523920580 
- bulgarian, japanese, traditional chinese... can get their specific glyphs from fonts (that have them) (when in Font kerning "best") - previously, I guess people had to install specific fonts, like the other variants of our shipped NotoSerifCJKsc.otf: NotoSerifCJKtc.otf, NotoSerifCJKjp.otf
- specific tweaks for polish, czech, slovak, portuguese (single letter prepositions should not be at end of line, duplicated real hyphen on next line...) when the publisher has not set the proper `&nbsp;` https://github.com/koreader/koreader/issues/5645#issuecomment-559165061

Screenshots, for the usual rewording suggestions :):

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>
(I added some icons from nerdfont - because I needed a break and that's always some fun chosing - for the typography features per languages, dunno if it's a good idea or if its confusing, or if nobody will care :)

Typography language submenu:

<kbd>image</kbd>

<kbd>image</kbd>

First item shows the book metadata lang tag, for easy going back to it:
<kbd>image</kbd>

Hyphenation submenu:
<kbd>image</kbd>
For a typography language, it allows switching between these last 3 hyphenation methods (didn't know if I could get rid of Algorithmic, so, well, it's available as an option with all typography languages).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6072)


NiLuJe:
I'm kind of on the fence about "Typography" on its own and/or when paired with "language".

I don't hate it, but something still bothers me a bit with it, not sure what, exactly...

What about "Typography rules"? (in both cases, I think when a language name is shown, it should be pretty obvious that it's a language ;p).

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+-- Languages to be shown in the menu.
+-- Aliases are language tags that could possibly be found in the book
+-- language metadata, that we wish to map to the correct lang tag.
+-- Features:
+--   H = language specific hyphenation dictionary
+--   b = language specific line breaking rules
+--   B = language specific additional line breaking tweaks
+-- Update them when language tweaks and features are added to crengine/src/textlang.cpp
+local LANGUAGES = {
+    -- lang-tag          aliases    features    menu title
```
NiLuJe, line 61: Random fun fact: 2 letters language codes are legacy ISO 639-1 (-ish, I think the language_DIALECT bastardization is a GNU thing) & ISO 639-2 (where I learned that fre is for French Belgian ;)).

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+        end
+        if not g_text_lang_set then
+            -- If nothing migrated, set the fallback to DEFAULT_LANG_TAG,
+            -- as we'll always have one of text_lang_default/_fallback set.
+            G_reader_settings:saveSetting("text_lang_fallback", DEFAULT_LANG_TAG)
+        end
+    end
+
+    local info_text = _([[
+Some languages have specific typographic rules: these include hyphenation, line breaking rules, and fonts language specific variations glyphs.
```
NiLuJe, line 163: ..., and language specific glyph variants.

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+KOReader will chose one according to the book metadata language tag, but you can select another one.
```
NiLuJe, line 164: ... according to the language tag from the book's metadata

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+You can also set a default language or a fallback one with a long-press.
+
+Features available per language are marked with:
+‐ : language specific hyphenation dictionary
+ : specific line breaking rules (mostly related to quotation marks)
+ : more line breaking rules (e.g. single letter prepositions not allowed at end of line)
```
NiLuJe, line 170: e.g.,

(extreme nitpicking, that's an en_US convention).

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+
+Note that when a language does not come with its own hyphenation dictionary, the English US one will be used.
```
NiLuJe, line 172: US English

(Unless you want to specifically refer to what's shown in the menu/filename, in which case I'd keep the underscore found in the original).

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+When the book language tag is not among our presets, no specific feature will be active, but it might be fine enough to trigger the fonts to use language variations known to them.]])
```
NiLuJe, line 173: book's
no specific features will be enabled
but it might be enough to get language specific glyph variants (when supported)

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+                self.ui:handleEvent(Event:new("UpdatePos"))
+            end,
+            hold_callback = function(touchmenu_instance)
+                UIManager:show(MultiConfirmBox:new{
+                    -- No real need for a way to remove default one, we can just
+                    -- toggle between setting a default OR a fallback (if a default
+                    -- one is set, no fallback will ever be used - if a fallback one
+                    -- is set, no default is wanted; so when we set one below, we
+                    -- remove the other).
+                    text = T( _("Would you like %1 to be used as the default (★) or fallback (�) typography language?\n\nDefault will always take precedence while fallback will only be used if the language of the book can't be automatically determined."), BD.wrap(lang_name)),
```
NiLuJe, line 282: Hmm, that one's mildly trickier with my initial comment about typography/typography rules...

...) language for typography rules?

Maybe?

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+    table.insert(self.menu_table, {
+        text_func = function()
+            local lang_name = LANG_TAG_TO_LANG_NAME[self.text_lang_tag] or self.text_lang_tag
+            return T(_("Typography language: %1"), lang_name)
+        end,
+        sub_item_table = self.language_submenu,
+    })
+
+    table.insert(self.menu_table, {
+        text = _("Follow document embedded lang tags"),
```
NiLuJe, line 313: document's

Or simply `Honor embedded lang tags`?

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -0,0 +1,802 @@
...
+        callback = function()
+            self.text_lang_embedded_langs = not self.text_lang_embedded_langs
+            self.ui.document:setTextEmbeddedLangs(self.text_lang_embedded_langs)
+            self.ui:handleEvent(Event:new("UpdatePos"))
+        end,
+        hold_callback = function()
+            local text_lang_embedded_langs = G_reader_settings:nilOrTrue("text_lang_embedded_langs")
+            UIManager:show(MultiConfirmBox:new{
+                text = text_lang_embedded_langs and _("Would you like to enable or disable support for document embedded lang tags by default?\n\nThe current default (★) is enabled.")
```
NiLuJe, line 322: Ditto.

frontend/apps/reader/modules/readertypography.lua
Frenzie: Something like `use` instead of `honor` is probably a bit easier to understand.

frontend/apps/reader/modules/readertypography.lua
NiLuJe: Or it's actually https://tools.ietf.org/html/rfc5646, as `ReaderTypography:parseLanguageTag()` mentions below ;).

frontend/apps/reader/modules/readertypography.lua
NiLuJe: Fair enough ;).

frontend/apps/reader/modules/readertypography.lua
Frenzie: Not that obscure, just grab a style guide and it's in there. :-P

frontend/apps/reader/modules/readertypography.lua
Frenzie: I think language for typography rules is clearer than typography language.

frontend/apps/reader/modules/readertypography.lua
poire-z: Well, I did not dig into that much (or at all). Not much interested in learning about that - so, if anyone wants to specialize and add stuff so we do the right thing with tags we'll meet in the wild, please jump on it :) (here and in textlang.cpp).
One thing I get is that Harfbuzz has lots of code to handle that/translate, and I hope it will do the right thing with what we provide as is. (Except may be with the 3 letters language code like "fre" that we translate here...)

frontend/apps/reader/modules/readertypography.lua
poire-z: Thanks for the previous feedback.
One thing just to be sure: it's really typograph**y** rules/languages, and not typograph**ic** rules like I'd be inclined to write in french?

frontend/apps/reader/modules/readertypography.lua
poire-z: `Use embedded language tags` ?
`Use document's embedded language tags` ?
`language tags` (for gramatically english) or `lang tags` (for HTMLically english :) ?

I'm not fond of the word "honor", but it's clearer that just "use". "Use" is a bit vague.
It's not like we "use" them, we "use" something else that will do according to them.
We don't _use_ road signs :)

frontend/apps/reader/modules/readertypography.lua
mergen3107: One small comment: it should be “Ukrainian” on line 49

Frenzie: I do have at least one Belgian French book that uses “ ”. I don't know if that's specifically Belgian and if it actually makes any difference to the algorithm.

poire-z: > I do have at least one Belgian French book that uses “ ”

Hard to recognize in github:
image

but pasting them into google, it sees them as left and right double quotation mark, so perfectly balanced and handled by:
https://github.com/koreader/crengine/blob/65d4adf843b4c85627187d4e2f685b8f42dfa1f4/crengine/src/textlang.cpp#L395-L398

So, whether there are space inside or not, they should stick to their first and last content.

poire-z: > One small comment: it should be “Ukrainian” on line 49

That's just a filename (well, that's how that file is named for ages), that might be saved in current users settings.
I was going to say it won't be shown, but it will, as the selected hyph dict.
So, dunno if I should rename it (may be later, when things are migrated), or if we should have another table for translating typos in filenames...

frontend/apps/reader/modules/readertypography.lua
Frenzie: You could just leave out the verb; the checkmark already says `use`/`honor`/etc.

poire-z: `[ ] Document's embedded lang tags` looks very very strange and incomplete to me :)

frontend/apps/reader/modules/readertypography.lua
Frenzie: I'd say they're both correct. Google Books seems to agree.

frontend/apps/reader/modules/readertypography.lua
mergen3107: @poire-z
Sorry I didn’t know that :) Looks like it comes all the way from the first ones who PR’ed Ukrainian language files

Frenzie: @mergen3107 On the contrary, thanks for pointing it out!

@poire-z If I go to hyphenation it already says Ukrainian, which probably implies we already have that?

poire-z: image
We have the other mapping lang_tag => language shows alongside typography, but we can't really use it, as zh-CN would show "Chinese (Simplified)" while it uses English_US.pattern.

I don't mind a small typo table or better: a function to remove the .pattern and fix some known typos, to be used instead of:
`hyph_dict_name = hyph_dict_name:gsub(".pattern$", "")`

poire-z: Well, I guess I'll update the HYPH_DICT_NAME_TO_LANG_TAG table to include a translatable string of the language - instead of using the filename stripped from .pattern, which won't be translatable.
If we remove the .pattern for pretty menus, I guess there's no more reason to keep the relation to the hyph dict filename.

frontend/apps/reader/modules/readertypography.lua
poire-z: `_("Honor embedded lang tags")`
`Would you like to honor or ignore embedded lang tags by default?` is fine by me, because `Ignore | Honor` buttons rhymes likes `Disable | Enable` and I like when things sing :)
Fine with you ?

Frenzie: Is this stuff about honoring Frenglish? Or are we just very corteous? :-P

poire-z:
Update screenshots:

<kbd>image</kbd>
No problem with it being longer than all others?
(Would `[ ] Hanging punctuation` be better placed inside `Typography rules >` ? May be not now as it's in ReaderTypeset and may need more work - but in principle?)

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

I updated this for consistency, but tell me if I should let the first one to keep already made translations:
```diff
-                    choice2_text = C_("Hyphenation", "Fallback"),
+                    choice2_text = C_("Typography", "Fallback"),
```

(Note: not to be merged before tomorrow, I want a nightly with just the crengine bump.)

frontend/apps/reader/modules/readertypography.lua
poire-z: Dunno, It's understandable in that sense in french, and @NiLuJe uses it quite often in his comments. What would be another word for that, less generic than "use" ?
(But in french, it's also used in a meaning a bit more further than corteous, rather meaning intercour-teou-se :) which is the only reason I'm a bit not fond of using it for our text :)

poire-z: Well, what we want is a positive antonym to "ignore".
And there's not many of them in the various thesaurus and dict on the web.
"Honorer" is the right one in french - didn't see among the english antonyms to "ignore"...
The only one I found that could work in english in our context is: Obey (or Follow, that I had initially).

Frenzie: apply

poire-z: `Apply document's embedded lang tags` ? So so, dunno...
What about my initial "Follow" ?

(There's "honor" in there :) http://www.antonymsfor.com/disregard )

NiLuJe:
> (Would `[ ] Hanging punctuation` be better placed inside `Typography rules >` ? May be not now as it's in ReaderTypeset and may need more work - but in principle?)
> 

Agreed (on both counts ;)).

frontend/apps/reader/modules/readertypography.lua
NiLuJe: Much like *règles de typographie* and *règles typographiques* in French :).

frontend/apps/reader/modules/readertypography.lua
Frenzie: I didn't say honor doesn't mean that, but I think in French the meaning of granting/complying with/agreeing to/abiding by/conforming to/obeying may be less formal than in English.

Apply isn't intended to be synonymous, but as a rewrite to avoid the problem for non-native non-French speakers.

Frenzie: But does everything have to be positive? I know there's a UI/UX school of thought that says it should be, but if the default is to abide by the lang tags I don't think it's weird to have an unchecked "Ignore lang tags" setting.

Or even just to leave out any kind of verb completely. The checkmark is a verb too. ^_^

poire-z: Overall, we are on the "check = positive = enable feature" side of UI thoughts :) (we have a few "Avoid...") - and in this new menu, all checkmarks are this positive/enable way. Would feel strange to have a "Ignore lang tags", even more because it's a new feature.
Still not fond of not having a verb, for this feature that is new and not really obvious and where "use" or "apply" do not convey much - while "honor" does.

> to avoid the problem for less advanced non-native non-French speakers

If "honor" works for english and french, can't we just get away with a translator note? :)

Frenzie: I don't really see a difference between honor and apply tbh. ;-)

poire-z: I "honor lang tags" by giving them the typograhy rules they request/deserve :)
I can't "apply lang tags", this has no sense or feels incomplete - lang tags are there, they say something - I can't _apply them_, I can _apply what they say_.
I could "apply typography rules according to lang tags", but that's a bit long.

poire-z: <kbd>image</kbd>

to/with ? Or just `[ ] Adjust to embedded lang tags`

Frenzie: If you don't honor lang tags it kind of feels like you're breaking an agreement with the document to me. :-P

> I can't "apply lang tags", this has no sense or feels incomplete - lang tags are there, they say something - I can't apply them, I can apply what they say.

But how is that any different for honor? I don't see the implied "content of" part as a problem either way.

poire-z: > If you don't honor lang tags it kind of feels like you're breaking an agreement with the document to me. :-P

That's why it's the default - should I remove "embedded styles off" and style tweaks, so we don't allow KOReader to break any agreement ? :)
(noticing you used naturally "If you don't honor lang tags" :)

> I don't see the implied "content of" part as a problem either way.

Well, I do :/
I can honor the Queen - by implicitely following/honoring her orders.
Can i apply the Queen ? :)

Requesting @NiLuJe 's input to settle this :)

Frenzie: > (noticing you used naturally "If you don't honor lang tags" :)

But I was making a point about the word honor. The sentence "if you don't use lang tags it feels like you're breaking an agreement" wouldn't make any sense.

If you don't use or apply lang tags for typography it's a neutral choice.
If you don't honor lang tags it might imply more.

> I can honor the Queen - by implicitely following/honoring her orders.

That's a very different meaning of the word. One even more evocative of what I'm calling slightly odd in this context. Now you're going as far as to imply disrespect to the author if you don't honor the lang tags. ;-)

poire-z: So, any alternative suggestions to "honor lang tags" (that you don't like) and "use/apply lang tags" (that I don't like) ? (so we can stop trying to convince each other :)

`Adapt/adjust/use/blah rules to/with/for embedded lang tags` (with the right grammar) ?
(no need for the longer word "typography", it's there in the 2 menu items above.

Frenzie: Enable?

poire-z: Full wording please.

Frenzie: Enable embedded lang tags
Enable embedded lang tag typography
Enable embedded typography language

Although really I'm still in favor of having the checkmark as the verb.

poire-z: Really, I'm still in favor of Honor :)
Again, `Enable/apply/use lang tags` doesn't make any sense to me.
And "Enable embedded typography language" is super ambiguous (_what, there is a typography language inside HTML ? no one told me, let's enable it!_)

Please go along trying to find something else, with the vocabulary we use in the other menu item, and trying to be explicite rather than cryptic (at least, Honor is poetic :)
`[verb] rules [preposition] embedded lang tags` feels the clearest to me, in the context of this menu.
And we'll see if the verb is replaceable by the checkmark.

Frenzie: You apply/use/honor lang tags [while rendering the document]. If `apply lang tags` doesn't make sense, surely neither does `honor lang tags`. But there's nothing *to* honor. There's no promise, agreement, debt or duty. But other than that minor shift in semantics I see no difference between the three.

poire-z: I understand you don't. But I do :) and you're not convincing me.
So, let's clear our mind from these and try to work out an alternative...

Frenzie: Set typography based on (embedded) lang tags

NiLuJe: I'll readily admit to quite liking honor, possibly a bit too much ;).

A quick wordnet check shows

`
     * S: (v) respect, honor, honour, abide by, observe (show respect
       towards) "honor your parents!"
     * S: (v) honor, honour (accept as pay) "we honor checks and drafts"
`

`Abide by` is definitely too formal, but `respect` could work?

Frenzie: Respect is probably the precise word you guys are actually looking for in honor. (Which isn't quite the same thing as being fully on board with it, but the objections I mentioned don't apply.)

poire-z: <kbd>image</kbd>

Fine with Respect, but somehow, after all this discussion, I now wish we had "rules" in here. "Honoring or respecting lang tags" feels more more enough :)

<kbd>image</kbd>

I could go with that - but "Set typography" feels a bit too global, could mean that the other settings will no more apply.
Anything less radical than "set", like adjust/adapt/tweak ?

`[ ] Typography rules by embedded lang tags` ? (no verb :) 
`[ ] Typography rules driven by embedded lang tags` ?

Frenzie: I want to say `Typography rules as per embedded lang tags` but that's a bit science slangy. "by" is short for something like "as specified by" and I'm not quite sure if it works by itself.

NiLuJe: @poire-z: I think we can get away without `rules` since in a two-option section between separators with the one right above explicitly saying  `Typography rules` ;).

I find the other suggestions way too wordy, an possibly not quite as clear (of which I prefer @Frenzie's, FWIW ;)).

poire-z: > my vote would be for @Frenzie

Which one ?

NiLuJe: > Which one ?

I meant as far as the wordy ones are concerned: `Typography rules as per embedded lang tags`.

But, to be clear, I much prefer

`Respect embedded lang tags` ;).

poire-z: I'm fine with all of these:
`Typography rules as per embedded lang tags`
`Rules as per embedded lang tags`
`As per embedded lang tags`
`Respect embedded lang tags`

So, please you both agree on one :)

(Really nothing could work with an initial verb like Adapt/adjust - to not have the feeling that if you check that box, and the document has no embedded lang tags, you will get no rule applied?)

poire-z:
<kbd>image</kbd>

Alternate wording for `their language..their content`?

frontend/apps/reader/modules/readertypography.lua
Frenzie: The respect one is fine by me. (I thought I already posted this. :-) )

Frenzie:
use relevant typographic rules while rendering their content

use language-specific/tailored typography rules while rendering their content

poire-z:
<kbd>image</kbd>

No problem with the double "while" here ? :)

Frenzie:
to render their content
for rendering their content

Not sure about the typography vs typographic but it doesn't seem like it'd hurt while adding some variation?

NiLuJe:
You can avoid the double "while" by keeping the first "while", and switching the second to "whereas" (which is used to emphasize contrast in these kind of constructs).

roshavagarga:
@poire-z Thanks for the information. Would it be able to add a check somewhere? If there's no font that support Bulgarian typography, but it's been enabled manually or automatically, then Koreader should send out a message noting the lack of a compatible font? Option B is to just include 1 font that supports Bulgarian typography by default.

poire-z:
> Would it be able to add a check somewhere? If there's no font that support Bulgarian typography, but it's been enabled manually or automatically, then Koreader should send out a message noting the lack of a compatible font? 

No, we can't know that. That's handled by low level HarfBuzz, that will give us the glyphs it decides without telling us how/why it did.
But a bulgarian reader will probably notice it visually and know something is wrong and act :)

> Option B is to just include 1 font that supports Bulgarian typography by default.

That's for our our font guy/girl/entity :) (@NiLuJe) although if the Noto fonts do not, it would be sad to have to put another one before these that would be used instead for anything latin...
I guess bulgarian readers, given the rarity of support, must have a prefered font that does provide bulgarian, and may as well add it as a user font and use it as their default font.
(And publishers of bulgarian EPUBs should add such font as embedded fonts, otherwise their books might render bad on most devices.)

roshavagarga:
Ah, I see, thanks for the explanation once again!

Noto and Free Serif should have these, but I don't see it working in Free Serif, not sure which version we have. More information is available here - https://github.com/googlefonts/noto-fonts/issues/89

NiLuJe:
Might be a job for #6272 then ;).

strn:
@roshavagarga: I can confirm that [Libertinus](https://github.com/alerque/libertinus) font correctly supports Serbian Cyrillic specific glyphs.

Issue #5969, opened 2020-03-16
Could the mini progress bar be prevented from disappearing after accidental click?


smartscripts-nl:
Sometimes I accidentally make the mini progress bar disappear by clicking somewhere in the neighbourhood of the bottom screen border. Then I have to click again to make the bar re-appear.

It would be nice if there was a setting to lock this bar, so that an accidental click can't make it disappear. Is this possible?

PR #6083, opened 2020-04-22
Option to prevent tap on status bar

robert00s:
See: #5969 

New option in Status bar -> Settings
`Lock tap status bar` (mayby better name?)

When we enable it - tap on status bar is disabled (locked)

Close: #5969

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6083)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -845,7 +845,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                 end,
             },
             {
-                text = _("Lock tap status bar"),
+                text = _("Lock status bar"),
                 checked_func = function()
                     return self.settings.lock_tap
                 end,
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -1785,10 +1785,7 @@ function ReaderFooter:onExitFlippingMode()
 end
 
 function ReaderFooter:onTapFooter(ges)
-    if self.settings.lock_tap then
-        return true
-    end
-    if self.has_no_mode then
+    if self.has_no_mode or self.settings.lock_tap then
         return
     end
     if self.view.flipping_visible then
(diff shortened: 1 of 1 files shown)


poire-z:
That's the obvious implementation for people (like most of us) that use "all at once".
But for those who don't, tap is used to rotate among items.
So, I dunno if it shouldn't be done to just skip the "nothing shown" step among the rotating items
(all at once being 2 steps : all or nothing, skipping nothing will stay with all, achieving the same effect) ?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1776,6 +1785,9 @@ function ReaderFooter:onExitFlippingMode()
 end
 
 function ReaderFooter:onTapFooter(ges)
+    if self.settings.lock_tap then
+        return true
```
poire-z, line 1789: Returning `true` will prevent the tap from being handled by other code. After it, there are page turns and bottom menu (see #5789).
So, may be just `return` as with next `if self.has_no_mode `.

robert00s: You're right :+1:

KenMaltby:
Just " Lock status bar" or  "Lock progress bar" or "Lock mini progress bar" or "Lock Footer" would do it, reference to tap not needed.

mergen3107:
Or maybe “Allow tap to hide status bar”, with default turned on?

robert00s:
I like ` Lock status bar`.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -844,6 +844,15 @@ function ReaderFooter:addToMainMenu(menu_items)
                     self:refreshFooter(true, true)
                 end,
             },
+            {
+                text = _("Lock tap status bar"),
```
Frenzie, line 848: Could you update the PR with that text?
```suggestion
                text = _("Lock status bar"),
```

Issue #5461, opened 2019-10-08
minor bug: "more plugins" stays when all plugins are disabled.


pazos:
* KOReader version: all
* Device: all

#### Issue

More plugins stays as a menu even when there's no content to show.

#### Steps to reproduce

Disable the following plugins:         

"auto_frontlight",
"battery_statistics",
"synchronize_time",
"keep_alive",
"doc_setting_tweak",
"terminal",

![device-2019-10-08-183232](https://user-images.githubusercontent.com/975883/66414593-06ff4800-e9fa-11e9-9469-808bc44b987b.png)

daemongloom:
Is there any way to disable that submenu but keep plugins on? Make them show on main menu?
I have two plugins in main part and 4 plugins in "more plugins" and it would be better to show all of them without additional submenu.

Frenzie:
Of course, the menus are customizable. You can put your overrides in settings/reader_menu_order.lua and settings/filemanager_menu_order.lua. The defaults can be found in ui/elements.

NB You should only put your overrides in there unless you want a maintenance burden.

Concretely, something like this:

```lua
local order = {
    tools = {
        "read_timer",
        "calibre_wireless_connection",
        "evernote",
        "statistics",
        "progress_sync",
        "wallabag",
        "zsync",
        "news_downloader",
        "send2ebook",
        "text_editor",
        --~ "----------------------------",
        --~ "more_plugins",
        "plugin_management",
    --~ },
    --~ more_plugins = {
        "auto_frontlight",
        "battery_statistics",
        "synchronize_time",
        "keep_alive",
        "doc_setting_tweak",
        "terminal",
    },
    ["KOMenu:disabled"] = {
        "more_plugins"
    },
}

return order
```

If you don't explicitly disable submenus or items, they'll pop up as new in the left-most menu. That's intended so that people who customize their menus are alerted of new features.

daemongloom:
> Of course, the menus are customizable. You can put your overrides in settings/reader_menu_order.lua and settings/filemanager_menu_order.lua. The defaults can be found in ui/elements.

It works, thanks. Would be great if there is a toggle (in plugin management, for example) that allows to do that without creating lua files though.

Frenzie:
The menu structure is the way it is by design. The fact that these are all plugins is an irrelevant technical detail from a UX viewpoint.

PRs welcome for a full LibreOffice-style menu customization UI. ;-)

PR #6092, opened 2020-04-25
Don't show "more plugins" when all plugins are disabled

robert00s:
Close: #5461

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6092)


Frenzie:
Also see #5512. I think that when I looked at it I decided the test was pointing out a legitimate issue but I haven't really had the time or motivation to look at it again.

robert00s:
Thanks, I'll look at it later.

Frenzie:
Closing for now since #6282 is much simpler.

PR #6090, opened 2020-04-24
bump crengine: multiple fallback fonts

poire-z:
Includes https://github.com/koreader/crengine/pull/339 :
- Simplify libunibreak includes
- Text: fix read/write outside array bounds
- lvtextfm: dont adjust space after initial quotation mark/dash (rework)
- Fonts: allow providing and using multiple fallback fonts

Users can set their prefered fallback font, which will be completed with a few of our shipped fonts for maximum coverage.
Ref https://github.com/koreader/koreader/issues/5277#issuecomment-525847129. Closes #5277.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6090)

PR #6095, opened 2020-04-26
Allow toggling use of additional fallback fonts

poire-z:
Allow toggling the use of multiple fallback fonts (#6090).
Might be handly when testing new fonts or choosing the first fallback font - when you really want to see what a font supports.

I've put it at the bottom of the font menu (if will be alone, the last one is optionnal):
<kbd>image</kbd>

Dunno if it's better if it was at start (with the Font settings that is not shown on most devices - or inside it, which would made this menu shown on all devices, so with a single item on most devices):
<kbd>image</kbd>

InfoMessage shown on long-press (for rewording suggestions):

<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6095)

### DIFF
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -362,14 +362,14 @@ These fonts will be used in this order:
 
 %1
 
-You can set a prefered fallback font with a long-press on a font name, and it will be used before these.
+You can set a preferred fallback font with a long-press on a font name, and it will be used before these.
 If that font happens to be part of this list already, it will be used first.]]),
             table.concat(self.ui.document.fallback_fonts, "\n")),
         separator = true,
     })
 
     table.insert(settings_table, {
-        text = _("Generate font test HTML document"),
+        text = _("Generate font test document"),
         callback = function()
             UIManager:show(ConfirmBox:new{
                 text = _("Would you like to generate an HTML document showing some sample text rendered with each available font?");
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -49,11 +49,15 @@ function ReaderFont:init()
                 event = "ChangeLineSpace", args = "decrease" },
         }
     end
-    -- build face_table for menu
+    -- Build face_table for menu
     self.face_table = {}
-    if Device:isAndroid() or Device:isDesktop() or Device:isEmulator() then
-        table.insert(self.face_table, require("ui/elements/font_settings"):getMenuTable())
-    end
+    -- Font settings
+    table.insert(self.face_table, {
+        text = _("Font settings"),
+        sub_item_table = self:getFontSettingsTable(),
+        separator = true,
+    })
+    -- Font list
     local face_list = cre.getFontFaces()
     for k,v in ipairs(face_list) do
         table.insert(self.face_table, {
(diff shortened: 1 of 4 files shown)


hius07:
A rare bird will reach the last page of the fonts list...

frontend/apps/reader/modules/readerfont.lua
```diff
@@ -82,8 +82,28 @@ function ReaderFont:init()
...
+        checked_func = function()
+            return G_reader_settings:nilOrTrue("complementary_fallback_fonts")
+        end,
+        callback = function()
+        G_reader_settings:flipNilOrTrue("complementary_fallback_fonts")
+            self.ui.document:setupFallbackFontFaces()
+            self.ui:handleEvent(Event:new("UpdatePos"))
+        end,
+        help_text = T(_([[
+Enable additional fallback fonts, for the most complete scripts and languages coverage.
```
NiLuJe, line 97: script and language

(Yeah, I know ;p).

frontend/apps/reader/modules/readerfont.lua
```diff
@@ -82,8 +82,28 @@ function ReaderFont:init()
...
+Complementary fonts, used in this order:
```
NiLuJe, line 98: Additional?

(Complementary is... clunky in English, possibly because it's so close to complimentary, which is another thing entirely ^^)

frontend/apps/reader/modules/readerfont.lua
```diff
@@ -82,8 +82,28 @@ function ReaderFont:init()
...
+
+%1
+
+The fallback font set with a long-press on a font name will be used before these.
+Note that the first font of the whole fallback fonts set is still used when this option is disabled.]]),
```
NiLuJe, line 103: Does this refer to this actual list, or the the "old" fallback font?

(In which case, it's potentially a duplicate if your global fallback is set to Noto Sans CJK SC?)

poire-z: (Not sure I understand your question :/ (the "old" fallback font?))

I tried to summarize the following with these 2 sentences :)

In https://github.com/koreader/crengine/pull/339, I explained a bit how that works:
> Users will still be able to only set one fallback font, which will be put at start of this list. If the font is already in that list, it will be moved first. This allows some bit of customisation, while always providing large coverage.

A user can set one fallback font. If not in the list: added at top. If in the list, moved to top.
So, if you set the user fallback to NotoSansCJK, this list does not change, it's the default behaviour :)
Dunno if we need to explain all this in that help_text - users shouldn't much care about duplicates.

(The only case where a duplicate/double rendering/drawing might happen is when you set as the main font one of the fonts in that list.)

frontend/apps/reader/modules/readerfont.lua
poire-z: As we have "additional" in the first sentence, may be for the 2nd sentence just:
`These fonts will be used in this order:`

poire-z: I guess your warning applies to the menu too:
`Complementary standard fallback fonts`
Suggestion for less heavy wording for this one? :)

NiLuJe: `Recommended` maybe? Not such a fan of that either, but it mostly works ;p.

frontend/apps/reader/modules/readerfont.lua
NiLuJe: I vastly prefer that quote from the second sentence used here, FWIW.

poire-z:
> Font settings menu at the beginning seems better.

OK, going to make it proper, which means bringing  `frontend/ui/elements/font_settings.lua` (from #5220) into readerfont.lua for oneplaceness.
I guess I can add to it the undocumented and hidden `Generate fonts test HTML document` - so we have at least 2 items in it.

frontend/apps/reader/modules/readerfont.lua
NiLuJe: i.e., keep the first sentence, but replace the second with something inspired from that like `If that font happens to be part of this list, it'll be moved to the top of the list`.

poire-z:
So, new Font settings submenu:
<kbd>image</kbd>

<kbd>image</kbd>
(only the 2 last items on Kobo, Kindle...)

<kbd>image</kbd>

The last item was hidden and undocumented, but well, now it is no more.
I made some sample text that we'll ship, so users will see this when hitting it:

<kbd>image</kbd>

Feel free to suggest some rewording in it - and if anyone wants to correct my chinese, hebrew or arabic gibberish with stuff that makes sense to people that can read it, please do :)

poire-z:
(Found a bidi bug in crengine that I'll have fixed in my next PR - so that text might show ugly under certain conditions as:
<kbd>image</kbd>
)

poire-z:
No more comment about this one?

frontend/apps/reader/modules/readerfont.lua
```diff
@@ -346,24 +336,79 @@ function ReaderFont:onAdjustFontSize(ges, direction)
...
+            self.ui.document:setupFallbackFontFaces()
+            self.ui:handleEvent(Event:new("UpdatePos"))
+        end,
+        help_text = T(_([[
+Enable additional fallback fonts, for the most complete script and language coverage.
+These fonts will be used in this order:
+
+%1
+
+You can set a prefered fallback font with a long-press on a font name, and it will be used before these.
```
Frenzie, line 365: preferred

frontend/apps/reader/modules/readerfont.lua
```diff
@@ -346,24 +336,79 @@ function ReaderFont:onAdjustFontSize(ges, direction)
...
+If that font happens to be part of this list already, it will be used first.]]),
+            table.concat(self.ui.document.fallback_fonts, "\n")),
+        separator = true,
+    })
+
+    table.insert(settings_table, {
+        text = _("Generate font test HTML document"),
```
Frenzie, line 372: I'd just leave it out, but otherwise I think HTML font test document sounds better.
```suggestion
        text = _("Generate font test document"),
```

poire-z:
Probably some bits in the shipped sample text that could be made less clumsy - but I guess we can fix that later (when you get a chance to read it on your device :)

Issue #6036, opened 2020-04-10
Gestures for jumping to first page and to random page


smartscripts-nl:
The upcoming version of KOReader will have gestures for jumping to the next or previous bookmark.

In the same vein I would like to suggest gestures for jumping to the first page or to a random page. That last functionality is for example handy for randomly selecting an inspirational text of the day.

In my local KOReader installation I've already incorporated working code for these functionalities:

in readergesture.lua:

    wallabag_download = _("Wallabag retrieval"),
    first_page = _("First page"),
    random_page = _("Random page"),

    [...]

        {"next_chapter", not self.is_docless},
        {"first_page", not self.is_docless},
        {"random_page", not self.is_docless},

    [...]

    elseif action == "first_page" then
        self.ui:handleEvent(Event:new("GotoPage", 1))

    elseif action == "random_page" then

        local last_page = self.view.document:getPageCount()
        local random_page = math.random(1,last_page)
        self.ui:handleEvent(Event:new("GotoPage", random_page))

hius07:
I would be happy to have gestures for jumping to the first page and to the last page of the book.

smartscripts-nl:
Jumping to last page is also easy to implement:

    elseif action == "last_page" then
        local last_page = self.view.document:getPageCount()
        self.ui:handleEvent(Event:new("GotoPage", last_page))

smartscripts-nl:
I thought of an improvement when using these gestures: add the location from which the jump was made to the locations stack. So we can navigate back to that location.

The statement to accomplish that is: self.ui.link:addCurrentLocationToStack():

    elseif action == "prev_bookmark" then
        self.ui.link:addCurrentLocationToStack()
        self.ui:handleEvent(Event:new("GotoPreviousBookmarkFromPage", self:getCurrentPage()))
    elseif action == "next_bookmark" then
        self.ui.link:addCurrentLocationToStack()
        self.ui:handleEvent(Event:new("GotoNextBookmarkFromPage", self:getCurrentPage()))
    elseif action == "first_page" then
        self.ui.link:addCurrentLocationToStack()
        self.ui:handleEvent(Event:new("GotoPage", 1))
    elseif action == "last_page" then
        self.ui.link:addCurrentLocationToStack()
        local last_page = self.view.document:getPageCount()
        self.ui:handleEvent(Event:new("GotoPage", last_page))
    elseif action == "random_page" then
        self.ui.link:addCurrentLocationToStack()
[...]

PR #6096, opened 2020-04-27
[Gesture] Add go to first and last page

robert00s:
See: #6036

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6096)


hius07:
Oh tanks a lot.

PR #4434, opened 2018-12-30
add support for debian packages

pazos:
Fixes #3108 and **maybe** #4398 

Requires https://github.com/koreader/koreader-base/pull/785

PR #6099, opened 2020-04-27
Debian - fix some lintian errors

pazos:
Follow-up to https://github.com/koreader/koreader/pull/4434#issuecomment-451219568

Fixed some errors and warnings. ~~I didn't care of warnings (there are soooo many)~~.
The mail address is a dummy address, feel free to suggest a change :p 

~~I didn't fix the `no-copyright-file` and `debian-changelog-file-missing` as I don't know how to fix them on non source debian packages.~~

The rest of errors left are:

`binary-or-shlib-defines-rpath` -> fixable during build
~~`unstripped-binary-or-object`-> fixable during build. I don't know why these are yet unstripped.~~
`embedded-library` -> not easily fixable without dismiss our own versions of libpng, libturbojpeg, libfreetype...

The whole list of remaining issues can be obtained running `lintian --no-tag-display-limit` against a debian package.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6099)

### DIFF
--- a/platform/debian/do_debian_package.sh
+++ b/platform/debian/do_debian_package.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 # Script to generate debian packages for KOReader
+# shellcheck disable=SC2164
 
 if [ -z "${1}" ]; then
     echo "${0}: can't find KOReader build, please specify a path"
@@ -21,6 +22,13 @@ uname_to_debian() {
     fi
 }
 
+link_font() {
+    font_name="$(basename $1)"
+    system_name="${2}/${font_name}"
+    rm -rf "$1"
+    ln -s "$system_name" "$i"
+}
+
 if [ -z "${2}" ]; then
     ARCH="$(uname_to_debian)"
 else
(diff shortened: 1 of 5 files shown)


platform/debian/do_debian_package.sh
```diff
@@ -38,20 +38,42 @@ if command_exists "${COMMAND}"; then
     {
         echo "Section: graphics"
         echo "Priority: optional"
-        echo "Depends: libsdl2-2.0-0"
+        echo "Depends: libsdl2-2.0-0, libc6 (>= 2.2.1)"
```
Frenzie, line 41: Shouldn't that (probably) be 2.23? What's 2.2.1? :-)

platform/debian/do_debian_package.sh
```diff
@@ -38,20 +38,42 @@ if command_exists "${COMMAND}"; then
...
         echo "Architecture: ${ARCH}"
         echo "Version: ${VERSION}"
         echo "Installed-Size: $(du -ks "${INSTALL_DIR}/debian/usr/" | cut -f 1)"
 
-        echo "Package: KOReader"
-        echo "Maintainer: KOReader team"
+        echo "Package: koreader"
+        echo "Maintainer: KOReader team <dummy@koreader.rocks>"
```
Frenzie, line 47: Are you sure it's not better just to let that error be?

platform/debian/do_debian_package.sh
NiLuJe: That's the first glibc 2 version with a stable API/ABI, IIRC ^^.

(Although, in practice, that's more like glibc 2.4).

platform/debian/do_debian_package.sh
pazos: Nope, you're right :+1:

platform/debian/do_debian_package.sh
pazos: I had that lying around on some old debian source package from 2016 era  <\_/>

In theory source packages need to include `Depends: ${shlibs:Depends}` and a helper tool will replace them based on real requirements. Buth yeah 2.2.3 is what Xenial ships. I will update.

pazos:
Updated commit to fix almost every warning too.

Now I would like to know why all the libraries and binaries are reported as unstripped. Maybe @NiLuJe knows :smile:

platform/debian/do_debian_package.sh
```diff
@@ -38,20 +39,65 @@ if command_exists "${COMMAND}"; then
...
         echo "Package: KOReader"
         echo "Maintainer: KOReader team"
         echo "Homepage: https://koreader.rocks"
-        echo "Description: An ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats"
-        echo " KOReader is a document viewer application, originally created for Kindle e-ink readers."
-        echo " It currently runs on Kindle, Kobo, PocketBook, Ubuntu Touch, Android and Linux devices"
+        echo "Description: Ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats"
+        echo " KOReader is a document viewer application, originally created for"
+        echo " Kindle e-ink readers. It currently runs on Kindle, Kobo, PocketBook,"
+        echo " Ubuntu Touch, Android and Linux devices"
```
Frenzie, line 52: I think the current main one is:

> KOReader is a document viewer for E Ink devices. Supported fileformats include EPUB, PDF, DjVu, XPS, CBT, CBZ, FB2, PDB, TXT, HTML, RTF, CHM, DOC, MOBI and ZIP files. It’s available for Kindle, Kobo, PocketBook, Android and desktop Linux.

(And if it's not, I/someone needs to update the website. It's not mentioning Cervantes for one.)

In any case you'd better not mention Ubuntu Touch since it doesn't actually start there. ;-)

Frenzie:
The strip command is definitely executed?

pazos:
> The strip command is definitely executed?

It should as in other platforms for non-debug builds, but I'm not sure. 

If being hackish is not an issue we can run it manually in do_debian_package.sh.

Frenzie:
We can just check in https://gitlab.com/koreader/nightly-builds to see what the output says then; I'll probably take a quick look.

Frenzie:
There doesn't seem to be any `--strip-unneeded` in the Debian build.

NiLuJe:
Emulator/Debug builds never strip anything: https://github.com/koreader/koreader-base/blob/57c848f3630e150321cfb8a975cc905070cb316e/Makefile#L37-L56

Frenzie:
And that's definitely how it should be, but if these Debian builds are debug builds it's a bug. ;-)

NiLuJe:
True, but they're still flagged as emulator builds, right?

Frenzie:
Maybe, too late to look at it now. ^_^

pazos:
Ok, thanks for the info. I will check base Makefiles.

pazos:
Yup, only happens on native builds. It happens because https://github.com/koreader/koreader-base/blob/60c8dc7f43910ccaf0457dd560ba37dceab45246/Makefile.defs#L116, but doesn't happen on cross armel and armhf packages.

Cross packages still report some unstripped binaries. but that's a huge improvement over native one:

```lintian
E: KOReader: unstripped-binary-or-object usr/lib/koreader/plugins/evernote.koplugin/lib/libluabitwise.so
E: KOReader: unstripped-binary-or-object usr/lib/koreader/plugins/evernote.koplugin/lib/libluabpack.so
E: KOReader: unstripped-binary-or-object usr/lib/koreader/plugins/evernote.koplugin/lib/libluacompat52.so
E: KOReader: unstripped-binary-or-object usr/lib/koreader/plugins/evernote.koplugin/lib/liblualongnumber.so
E: KOReader: unstripped-binary-or-object usr/lib/koreader/plugins/evernote.koplugin/lib/libluasocket.so
E: KOReader: unstripped-binary-or-object usr/lib/koreader/rocks/lib/lua/5.1/lpeg.so
```

Now, if I remove the `EMULATE_READER=1` from the debian definition I will need to set CC=gcc instead of CHOST-gcc. I don't really know where to do that. Hints?

NiLuJe:
@pazos: You're probably in for a painful experience if you go the "get rid of EMULATE_READER" way.

I'd simply tweak the conditions around the stripping tests I linked earlier (and fix the stripping to include the luarocks properly ;)).

pazos:
Good to know. I'm in for the hackish route:

```diff
--- a/platform/debian/do_debian_package.sh
+++ b/platform/debian/do_debian_package.sh
@@ -95,6 +95,27 @@ if command_exists "${COMMAND}"; then
     find . -type f -name "defaults.lua" -print0 | xargs -0 chmod 644
     find . -type f -name ".gitignore" -print0 | xargs -0 rm -rfv
 
+    ### strip binaries and shared libraries
+    case "${ARCH}" in
+        "armel")
+            STRIP="arm-linux-gnueabi-strip"
+            ;;
+        "armhf")
+            STRIP="arm-linux-gnueabihf-strip"
+            ;;
+        *)
+           STRIP="strip"
+            ;;
+    esac
+    find . -type f -name "*.so*" -print0 | xargs -0 "${STRIP}" --strip-unneeded
+
+    ### try to remove RPATH
+    if command_exists chrpath; then
+        find . -type f -name "*.so*" -print0 | xargs -0 chrpath -d
+    else
+        echo "chrpath tool not found. Skipping RPATH deletion"
+    fi
+
```

NiLuJe:
Something along the lines of 

```
DO_STRIP := $(if $(or $(EMU),$(DBG)),,1)
DO_STRIP := $(if $(or $(DO_STRIP),$(DEB)),1,) 
ifeq ($(DO_STRIP),1)
    <stuff>
endif
```

Except with the actual variable names ;p.

pazos:
Ok. Added some files (changelog, copyright and manpage) . Fixed all issues that made the package wonky. The rest of the issues reported by lintian are expected even on official debian repos.

The name of the package is now lowercase. That matches the wrapper script that launches the program, the manpage, the changelog et al. 

The changelog and the control file both require an email address. I added mine for this PR but that can be changed. Foliate uses a [github-noreply one](https://github.com/johnfactotum/foliate/blob/master/debian/control#L4). I'm not sure how that works but it will be a nice thing to have.

Frenzie:
lol, that changelog. :-P

Anyway, still seems fine at a glance. ;-)

pazos:
Ok, did a few changes to be even more compliant with debian/lintian:

removed noto fonts (except CJK) and freefonts: use standard debian deps.
removed droid, linked with droidfallback (because license issues)

After applying this patch (assuming chrpath is on path) the rest of lintian warnings and errors are strictly non-fixable:

```
E: koreader: embedded-library usr/lib/koreader/libs/libfreetype.so.6: freetype
E: koreader: embedded-library usr/lib/koreader/libs/libjpeg.so.8: libjpeg
E: koreader: embedded-library usr/lib/koreader/libs/libmupdf.so: openjpeg
E: koreader: embedded-library usr/lib/koreader/libs/libpng16.so.16: libpng
E: koreader: embedded-library usr/lib/koreader/libs/libturbojpeg.so: libjpeg
E: koreader: embedded-library usr/lib/koreader/libs/libz.so.1: zlib
E: koreader: embedded-library usr/lib/koreader/sdcv: zlib
W: koreader: shared-lib-without-dependency-information usr/lib/koreader/plugins/evernote.koplugin/lib/libluabitwise.so
```

Frenzie:
Dunno about strictly non-fixable (zlib potentially might be?) but certainly not *worth* fixing imho. ;-)

pazos:
Yeah, I meant without devendoring libraries, a la https://github.com/koreader/koreader-base/issues/988 :p

Frenzie:
Sure, you could do that for a lot of things that aren't MuPDF, but I purposefully singled that one out as one of the least problematic ones. ;-)

Then again, they all look fine really:
https://abi-laboratory.pro/?view=timeline&l=freetype
https://abi-laboratory.pro/?view=timeline&l=libjpeg-turbo
https://abi-laboratory.pro/?view=timeline&l=libpng
https://abi-laboratory.pro/?view=timeline&l=zlib

pazos:
I can sleep with that lintian error. It seems to be triggered by a bunch of other packages: https://lintian.debian.org/tags/embedded-library.html

Feel free to review &| merge

hyiltiz:
Is it going to get into debian unstable or experimental sometime soon?

NiLuJe:
That was never the goal.

PR #6101, opened 2020-04-28
Move to archive plugin

mwoz123:
Plugin to move/copy current reading book to archive folder

Zrzut ekranu z 2020-04-28 20-58-19

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6101)


frontend/ui/elements/reader_menu_order.lua
```diff
@@ -120,6 +120,7 @@ local order = {
         "zsync",
         "news_downloader",
         "send2ebook",
+        "moveToArchive",
```
Frenzie, line 123: This pushes more plugins to the next page, not necessarily the best move.

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+            {
+                text = _("Copy current book to archive"),
+                callback = self.copyToArchive,
+            },
+            {
+                text = _("Go to archive folder"),
+                callback = function()
+                    if not archive_dir_path then
+                        UIManager:show(InfoMessage:new{
+                            text = _("No archive directory. Please set it first")
```
Frenzie, line 43: You could make it a simple ConfirmBox.

Please set an archive folder first.

Cancel/Set archive folder

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+                        FileManager:showFiles(archive_dir_path)
+                    end
+                end,
+            },
+            {
+                text = _("Go to last copied/moved from folder"),
+                callback = function()
+                    if not last_copied_from_dir then
+                        UIManager:show(InfoMessage:new{
+                            text = _("No previous dir found")
```
Frenzie, line 59:
```suggestion
                            text = _("No previous folder found.")
```

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+                callback =  self.setArchiveDirectory,
+            }
+        },
+    }
+end
+
+function MoveToArchive:moveToArchive()
+    if not archive_dir_path then
+        UIManager:show(InfoMessage:new{
+            text = _("No archive directory. Please set it first")
```
Frenzie, line 82: Same as above

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+    logger.dbg("MoveToArchive: last_copied_from_dir :", last_copied_from_dir)
+
+    FileManager:moveFile(document_full_path, archive_dir_path)
+
+    move_to_archive_settings:saveSetting(last_copied_from_config_key, ("%s/"):format(last_copied_from_dir))
+
+    ReadHistory:removeItemByPath(document_full_path)
+
+    UIManager:show(InfoMessage:new{
+        text = _("Book moved. Please reopen it from archive location.")
```
Frenzie, line 97: I'd be inclined to leave off the second sentence or to make it a ConfirmBox.

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+     })
+end
+
+function MoveToArchive:copyToArchive()
+    if not archive_dir_path then
+        UIManager:show(InfoMessage:new{
+            text = _("No archive directory. Please set it first")
```
Frenzie, line 104: Same as above

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+    local document_full_path = G_reader_settings:readSetting("lastfile")
+    last_copied_from_dir = util.splitFilePathName(document_full_path)
+
+    logger.dbg("MoveToArchive: last_copied_from_dir :", last_copied_from_dir)
+    move_to_archive_settings:saveSetting(last_copied_from_config_key, ("%s/"):format(last_copied_from_dir))
+
+    FileManager:cpFile(document_full_path, archive_dir_path)
+
+    UIManager:show(InfoMessage:new{
+        text = _("Book copied. Please reopen it from archive location if needed.")
```
Frenzie, line 117: Same as above

frontend/ui/elements/reader_menu_order.lua
mwoz123: ok will be last here.

Robertmccalister747:
Thanks. Maybe you can send me some code

mwoz123:
@Robertmccalister747  I don't understand what you mean. Code is in this repo/PR.

mwoz123:
> Interesting concept!

Thanks:)

plugins/moveToArchive.koplugin/_meta.lua
```diff
@@ -0,0 +1,6 @@
+local _ = require("gettext")
+return {
+    name = "moveToArchive",
```
Frenzie, line 3:
```suggestion
    name = "move_to_archive",
```

frontend/ui/elements/reader_menu_order.lua
```diff
@@ -121,6 +121,7 @@ local order = {
         "news_downloader",
         "send2ebook",
         "text_editor",
+        "moveToArchive",
```
Frenzie, line 124: Is more plugins always on the next page? I meant the menu item just under this one.

```suggestion
        "move_to_archive",
```

mwoz123:
Changes done.

frontend/ui/elements/reader_menu_order.lua
mwoz123: done. 

> Is more plugins always on the next page? I meant the menu item just under this one.

I think it depends on your screen resolution + no. of enabled plugins.

Any I made it last one as you wanted...

mwoz123:
Changes done.

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,139 @@
...
+                text = _("Set archive directory"),
+                keep_menu_open = true,
+                callback =  self.setArchiveDirectory,
+            }
+        },
+    }
+end
+
+function MoveToArchive:moveToArchive()
+    local move_done_text = _("Book moved. \nDo you want to open it from archive folder?")
```
Frenzie, line 80:
```suggestion
    local move_done_text = _("Book moved.\nDo you want to open it from the archive folder?")
```

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,139 @@
...
+    local doc_old_path = MoveToArchive:commonProcess(true, move_done_text)
+    ReadHistory:removeItemByPath(doc_old_path)
+end
+
+function MoveToArchive:copyToArchive()
+    local copy_done_text =_("Book copied. \nDo you want to open it from archive folder?")
```
Frenzie, line 86:
```suggestion
    local copy_done_text =_("Book copied.\nDo you want to open it from the archive folder?")
```

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,139 @@
...
+            move_to_archive_settings:saveSetting(archive_dir_config_key, ("%s/"):format(path))
+            move_to_archive_settings:flush()
+
+            archive_dir_path = path
+        end,
+    }:chooseDir()
+end
+
+function MoveToArchive:showNoArchiveConfirmBox()
+    MoveToArchive:showConfirmBox(_("No archive directory. \nDo you want to set it now?"), self.setArchiveDirectory)
```
Frenzie, line 126:
```suggestion
    MoveToArchive:showConfirmBox(_("No archive directory.\nDo you want to set it now?"), self.setArchiveDirectory)
```

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,139 @@
...
+end
+
+function MoveToArchive:showConfirmBox(text, ok_callback)
+    UIManager:show(ConfirmBox:new{
+        text = text,
+        cancel_text = _("Cancel"),
```
Frenzie, line 132: That's the default, no need.
```suggestion
```

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,139 @@
...
+        ok_text = _("Ok"),
```
Frenzie, line 133: That's the default, but it should describe the action to be confirmed, see http://koreader.rocks/doc/modules/ui.widget.confirmbox.html
```suggestion
        ok_text = _("Set archive folder"),
```

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,139 @@
...
+local archive_dir_path = move_to_archive_settings:readSetting(archive_dir_config_key)
+local last_copied_from_dir = move_to_archive_settings:readSetting(last_copied_from_config_key)
+
+function MoveToArchive:init()
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function MoveToArchive:addToMainMenu(menu_items)
+    menu_items.move_to_archive = {
+        text = _("Move to Archive"),
```
Frenzie, line 30:
```suggestion
        text = _("Move to archive"),
```

frontend/ui/elements/reader_menu_order.lua
```diff
@@ -121,6 +121,7 @@ local order = {
         "news_downloader",
         "send2ebook",
         "text_editor",
+        "move_to_archive",
```
Frenzie, line 124: I'm inclined to think this would fit better near progress_sync if we insert it in this part of the menu. The more_plugins menu will be pushed down no matter where above it you insert an extra entry.

Pinging @NiLuJe @poire-z @pazos for some thoughts.

poire-z: I guess the more_plugins menu isn't really needed anymore, now that the user can disable the plugins he doesn't need.
If you put some in more_plugins, no matter the user cleaning, he would have to tap once more to reach that unfortunate plugin we put into more_plugins.
If we'd want to keep more_plugins, only non-user-action plugins (those with only settings or some permanent behaviour) should go in there.

mwoz123:
> A few minor string nitpicks

Resolved

mwoz123:
Btw off topic : do we still need "more plugins" as there's something like "next page" in menu if there're too many entries?

Frenzie:
Like @poire-z said we can organize it differently.

frontend/ui/elements/reader_menu_order.lua
hius07: Yes, I have only a few plugins activated, and definitely do not need More plugins menu.

poire-z:
Just unsure what's happening when moving:
- you're currently reading a book
- you move the book file (only the file, not the docsetting (book.sdr/metadata.lua ?)
- you quit the book, the docsetting is saved and stays at the old place?

If that is what's really happening, you have:
1) and old docsetting unused and dangling
2) possibilities for undefined behaviours in other contexts as that's a case we don't expect until now: having the book opened but no underlying file anymore

In FileManager, when we move/rename a file (and we do that the book closed), there are a bit more stuff taken care of:
https://github.com/koreader/koreader/blob/be72203b8302730b69f92b34acecb97baa96a83a/frontend/apps/filemanager/filemanager.lua#L782-L806

frontend/ui/elements/reader_menu_order.lua
yparitcher: I agree that the more plugins menu in not necessary, and just annoying for the plugins that i use that are there. (I only use ~5 plugins)

mwoz123:
Ok, cool off topic extracted to : https://github.com/koreader/koreader/issues/6105

mwoz123:
@poire-z right sdr is currently not moved.

This was designed mainly for wallabag /nd/S2E so it shouldn't be a big issue.

But anyway I wanted to add this next PR.
Are you ok with that?

poire-z:
> This was designed mainly for wallabag /nd/S2E so it shouldn't be a big issue.
> But anyway I wanted to add this next PR.
Are you ok with that?

Files lying around is an issue for me :) People might test it tomorrow if merged, and that's rather not cool leaving mess on their device.
It's a plugin, so I wouldn't really mind it having bugs. But now that you know of them, it would be a bit unfortunate to not fix them before merging :)

But I reckon it's not easy fixing... Even if you move the docsettings, it will be saved at the old place when the book is closed. So, ideally, when moving, you would ask/tell the user that the file will be moved and re-opened, and if he says OK: close the book, move file + docsettings, re-open the book from the archived file (and if you propose and option to not re-open the book, what to do from there?).
And when you close the book, your plugin is gone... so... not easy (may be adding a `after_close_callback` in `ReaderUI:switchDocument()` like it's done for `ReaderUI:reloadDocument(after_close_callback)` )

Anyway, I let @Frenzie decide how critical that is :)
(I won't use this plugin, I don't really mind people testing it having mess on their devices, and it would probably take us quite a bit of time to help you getting it fully right :)

Frenzie:
I didn't realize it didn't deal with SDRs properly; I'd definitely prefer to see that fixed first too.

mwoz123:
Ok will try.
Can you help me how to force close current book or force save unsaved book changes j

poire-z:
> Can you help me how to force close current book or force save unsaved book progress/changes in sdr?

```
$ rgrep ui:onClose frontend/
frontend/apps/reader/modules/readermenu.lua:                self.ui:onClose()
frontend/apps/reader/modules/readermenu.lua:        self.ui:onClose()
frontend/apps/reader/modules/readergesture.lua:        self.ui:onClose()
frontend/apps/reader/modules/readerstatus.lua:    self.ui:onClose()
```

But it mostly depends on what you want to do after (open another document, go to file browser, exit koreader).

It's a bit unclear to me what's your workflow with this plugin (and what it will be for other people):
- do you really need/want to stay reading the book after you archived it? Isn't it something you do when you're done with it?
- if that's something you invoke at end of book, it could just be a new item in "Gear > Document > End of document action" and in it's "ask with popup dialog" - in which case, one just have to "go to end of book" to invoke this action/popup.
- does "_copy_ to archive folder" has some use case?

mwoz123:
Don't know what I'm doing wrong but in my `commonProcess()`
>    logger.dbg("MoveToArchive: self.ui == nil : ", self.ui == nil )

always returns
> 04/30/20-11:15:55 DEBUG MoveToArchive: self.ui == nil :  true

Can you help here? I assume I need to only execute `self.ui:onClose()` when I want to close book?

> do you really need/want to stay reading the book after you archived it? Isn't it something you do when you're done with it?

Opening from new place allows to add it to the history from archive. 

>if that's something you invoke at end of book, it could just be a new item in "Gear > Document > End of document action" and in it's "ask with popup dialog" - in which case, one just have to "go to end of book" to invoke this action/popup.

Good idea. Once this is merged I could make it in new PR.

>does "copy to archive folder" has some use case?

This is for Wallabag. If you sync it it's able to remove from server but only articles that are 100% read but not deleted.

poire-z:
> Can you help here? I assume I need to only execute self.ui:onClose() when I want to close book?

I think that EVERYWHERE you use, inside a `function MoveToArchive:someMethod()`:
```lua
     MoveToArchive:commonProcess(false, copy_done_text)
         MoveToArchive:showNoArchiveConfirmBox()
```
You need to use `self`:
```lua
     self:commonProcess(false, copy_done_text)
         self:showNoArchiveConfirmBox()
```

> Opening from new place allows to add it to the history from archive (and ofcourse continue reading;)

Yes, but not doing anything while you're not yet done reading works just as well :)

> Good idea. Once this is merged I could make it in new PR.

Well, if you could just be OK with a End of document action (so, at the end, decide to delete it or archive it), you'll avoid yourself many problems to solve (as the End of document code deals with what to do and what to do next).

> This is for Wallabag. If you sync it it's able to remove from server but only articles that are 100% read but not deleted.

Well, I know nothing about Wallabag.
But this plugin with the use cases you describe seems a bit like a quick hack bad solution.
No reason to move it to archive until you're done reading, unless Wallabag too is involved.
And it there are risks of Wallabag removing it, or the need to mark it as "keep it on the device, delete it from server" while continuing reading, it might be something to implement in the Wallbag plugin, rather that make that "Move to archive" plugin with a not-so-obvious workflow.

I see the problem is generic: NewsDownloader, send2ebook, Wallabag, even with a directory you would sync both ways with a script, where you want to keep the book on the device but delete it from the remote end. But having a single "archive" dir is a bit limited, I guess I would fill it (via File browser) in a better suited place.
But I dunno really, I don't use any of that.
Might need a more global thinking about how to solve that issue if it is one. May be this plugin is not the solution, or may be it is :)

Frenzie:
> This is for Wallabag. If you sync it it's able to remove from server but only articles that are 100% read but not deleted.

The way I interpreted it was generic. There's your "currently reading"/home folder and your "archive" folder, or something to that extent. (Probably as a subdirectory of the home folder, to complicate matters.) Of course folder structure should otherwise remain intact but that's not necessarily a blocker for a proof of concept.

> This is for Wallabag. If you sync it it's able to remove from server but only articles that are 100% read but not deleted.

I have to say I'm a bit confused by this statement. Doesn't that mean you'd want something in the Wallabag plugin?

poire-z:
You also need to update your callback:

```diff
- callback = self.moveToArchive
+ callback = function() self:moveToArchive() end,
   -- ^ equivalent to callback = function() MoveToArchive.moveToArchive(self) end,
```
Otherwise, "self" (that has self.ui) is never passed so not accessible.

mwoz123:
I rethought the idea.

And was wondering if we could make option to wallabag, (not in this Pr) to delete  from server each successfully downloaded article. @Frenzie? If we can remove it later (once it 100%read on next sync) we should be also to do it right a way in current "Download" 
This will remove the need for "copy to archive".

Making it as a plugin also allows to move to archive anytime not only at the end of book.
Can it stay plugin?

mwoz123:
Recent changes:
- Moving .sdr file - done (logic from filemanger - not sure if all needed - can you check?)
- menu now appears also in filebrowser mode 
- "coping to archive" marked as "experimental"
- fix issue with confirm ok label

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,163 @@
...
+                callback =  self.setArchiveDirectory,
+            }
+        },
+    }
+end
+
+function MoveToArchive:moveToArchive()
+    local move_done_text = _("Book moved.\nDo you want to open it from the archive folder?")
+    local doc_old_path = self:commonProcess(true, move_done_text)
+    ReadHistory:removeItemByPath(doc_old_path)
```
poire-z, line 91: Better have that done in `self:commonProcess(true...` so all that housekeeping is in a single place).

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,163 @@
...
+    self:commonProcess(false, copy_done_text)
+end
+
+function MoveToArchive:commonProcess(is_move_process, moved_done_text)
+    if not self.ui.document then
+        UIManager:show(InfoMessage:new{
+            text = _("Only works in reading book mode.")
+         })
+        return
+    end
```
poire-z, line 105: Not needed. Just have in your menu items: `enabled_func = function() return self.ui.document ~= nil end,` so the user sees it's not possible.
And to make that even better, include in your enabled_func a check if that book is actually from the archive folder: if it is, don't allow moving it to archive folder (a hint for the user to remember him he has already moved it.)
(How does it work currently if you hit that 2 times?)

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,163 @@
...
+    if not archive_dir_path then
+        self:showNoArchiveConfirmBox()
+        return
+    end
+    local document_full_path = G_reader_settings:readSetting("lastfile")
+    self.ui:onClose()
```
poire-z, line 111: So, that works? This function continues running and all is well after the document is closed?

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,163 @@
...
+
+    logger.dbg("MoveToArchive: last_moved/copied_from_dir :", last_copied_from_dir)
+    move_to_archive_settings:saveSetting(last_copied_from_config_key, ("%s/"):format(last_copied_from_dir))
+
+    if is_move_process then
+        FileManager:moveFile(document_full_path, archive_dir_path)
+        FileManager:moveFile(DocSettings:getSidecarDir(document_full_path), archive_dir_path)
+    else
+        FileManager:copyFileFromTo(document_full_path, archive_dir_path)
+        FileManager:copyFileFromTo(DocSettings:getSidecarDir(document_full_path), archive_dir_path)
```
poire-z, line 123: Does that work? copyFileFromTo uses /bin/cp, without any -r option to copy the directory and its content.
I still think it's just not "experimental", it's a bad idea copying the book and the docsettings, and have them in 2 places, not knowing which docsetting/progress is the active one (and the other one will be stalled - if the user makes highlights while continuing reading, they won't be in the other book docsettings).

plugins/moveToArchive.koplugin/main.lua
```diff
@@ -0,0 +1,163 @@
...
+    end
+    local dest_file = string.format("%s/%s", archive_dir_path, BaseUtil.basename(document_full_path))
+    require("readhistory"):updateItemByPath(document_full_path, dest_file)
+    ReadCollection:updateItemByPath(document_full_path, dest_file)
+    -- Update last open file.
+    if G_reader_settings:readSetting("lastfile") == document_full_path then
+        G_reader_settings:saveSetting("lastfile", dest_file)
+    end
+
+    self:showConfirmBox(moved_done_text, _("Ok"), function () ReaderUI:showReader(archive_dir_path .. filename) end)
```
poire-z, line 133: What happens if the user hits Cancel ? Now that you have closed the book?

Frenzie:
@mwoz123 
> I rethought the idea.
> […]
> Can it stay plugin?

If you thought of something better I'd say it's not a bad thing to just throw away the first draft, using what you learned to your advantage. In code as in writing you shouldn't be afraid to scratch things to make things better. ;-)

plugins/moveToArchive.koplugin/main.lua
mwoz123: Yes it does. I assumed it just clears unneeded ui widget.
But thanks to that question I spot I haven't check "cancel" in "do you want to open it from archive".
Anyway can be here "end of book" actions

plugins/moveToArchive.koplugin/main.lua
mwoz123: Copy is essentially "copy" ;) of "move" command. But with replaced "mv" to "cp". So as it works for move should work same way for copy

poire-z: "should work" ? Have you tested it?
```
/tmp/test $ touch this_is_a_file
/tmp/test $ mkdir this_is_a_dir
/tmp/test $ mkdir other_dir
/tmp/test $ mv this_is_a_file other_dir/
/tmp/test $ mv this_is_a_dir other_dir/
/tmp/test $

/tmp/test $ touch this_is_a_file2
/tmp/test $ mkdir this_is_a_dir2
/tmp/test $ mv this_is_a_file2 other_dir
/tmp/test $ cp this_is_a_dir2 other_dir/
cp: -r not specified; omitting directory 'this_is_a_dir2'
/tmp/test $ ls
other_dir  this_is_a_dir2
/tmp/test $
```
`this_is_a_dir2` would be your sidecar directory.

mwoz123:
```
05/01/20-23:10:50 INFO  AAAAAAAAAAAAAAAA new_folder_path= /mnt/Dane/koreader/metadata///2col.sdr/
05/01/20-23:10:50 INFO  BBBBBBBBBB full_file_path= /mnt/Dane/koreader/2col.sdr/metadata.pdf.lua
05/01/20-23:10:50 INFO  AAAAAAAAAAAAAAAA2 new_folder_path= /mnt/Dane/koreader/metadata///2col.sdr/
05/01/20-23:10:50 INFO  CCCCCCCCCC copyFileFromTo from /mnt/Dane/koreader/metadata///2col.sdr/
05/01/20-23:10:50 INFO  CCCCCCCCCC copyFileFromTo to
05/01/20-23:10:50 INFO  BBBBBBBBBB full_file_path= /mnt/Dane/koreader/2col.sdr/metadata.pdf.lua.old
05/01/20-23:10:50 INFO  AAAAAAAAAAAAAAAA2 new_folder_path= /mnt/Dane/koreader/metadata///2col.sdr/
05/01/20-23:10:50 INFO  CCCCCCCCCC copyFileFromTo from /mnt/Dane/koreader/metadata///2col.sdr/
05/01/20-23:10:50 INFO  CCCCCCCCCC copyFileFromTo to
```

Can you help me with this fight ? I've spent much time on that copy Recursive in FileManager.lua

For unknow for me reasons once it's in `copyRecursive()` the values are ok (BBBBBBBBBB in logs) but when I called from it `copyFileFromTo()` and pass those values (CCCCCCCCCC)  `to` and `from` logs something are completely different. 

What's going on?

NiLuJe:
Hmm, why reinvent the wheel? What's wrong with `cp -r`?

mwoz123:
Btw there's still no tools/ide that would be able to debug that code other than just logger.debug/info/error?

mwoz123:
>Hmm, why reinvent the wheel? What's wrong with cp -r?

 `   cp_recursive_bin = Device:isAndroid() and "/system/bin/cp - r" or "/bin/cp - r",`
+ baseutils.execute() 
But nothing happened:(

NiLuJe:
`cp` in the command, `-r` is already an arg. You basically want to pass a table to execute, i.e., `{ "cp", "-r", from, to}`.

poire-z:
> What's going on?

Again, your dot vs colon eternal curse :/
http://lua-users.org/wiki/ColonForMethodCall http://www.lua.org/pil/16.html

You define: `function FileManager:copyFileFromTo(from, to)`, a method, expecting an implicit `self` when called with a colon.
But you call: `self.copyFileFromTo(full_file_path, new_folder_path)`, as a function, with a dot - not passing the implicit `self`.
So, in your defined `FileManager:copyFileFromTo()`, you get:
```
self = full_file_path
from = new_folder_path 
to = nil (cause you passed only 2 values)
```

mwoz123:
Thanks Guys now it works, and I know why:)

mwoz123:
@poire-z changes done. Please re review

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,159 @@
...
+    require("readhistory"):updateItemByPath(document_full_path, dest_file)
+    ReadCollection:updateItemByPath(document_full_path, dest_file)
+    -- Update last open file.
+    if G_reader_settings:readSetting("lastfile") == document_full_path then
+        G_reader_settings:saveSetting("lastfile", dest_file)
+    end
+
+    self:showConfirmBox(moved_done_text, _("Ok"), function () ReaderUI:showReader(archive_dir_path .. filename) end,
+        function () self:openFileBrowser(last_copied_from_dir) end)
+    ReadHistory:removeItemByPath(document_full_path)
```
poire-z, line 116: Might be redundant with `require("readhistory"):updateItemByPath(document_full_path, dest_file)` above?
(If not, move it just after that one, all that housekeeping before the ending ConfirmBox)).

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,159 @@
...
+        onConfirm = function(path)
+            archive_dir_path = path
+            move_to_archive_settings:saveSetting(archive_dir_config_key, ("%s/"):format(path))
+            move_to_archive_settings:flush()
+        end,
+    }:chooseDir()
+end
+
+function MoveToArchive:showNoArchiveConfirmBox()
+    self:showConfirmBox(_("No archive directory.\nDo you want to set it now?"), _("Set archive folder"), self.setArchiveDirectory)
```
poire-z, line 131: Multilines for easier reading.

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,159 @@
...
+end
+
+function MoveToArchive:showConfirmBox(text, ok_text, ok_callback, cancel_callback)
+    UIManager:show(ConfirmBox:new{
+        text = text,
```
poire-z, line 136: You might as well directly use `UIManager:show(ConfirmBox:new{})` anywhere you use MoveToArchive:showConfirmBox().
This function does not bring any added value to that.

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,159 @@
...
+        self.ui:onClose()
+    end
+    if FileManager.instance then
+        FileManager.instance:reinit(path)
+    else
+        FileManager:showFiles(path)
+    end
+end
+
+return MoveToArchive
```
poire-z, line 159: "No newline at end of file" says github.

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,159 @@
...
+function MoveToArchive:copyToArchive()
+    local copy_done_text =_("Book copied.\nDo you want to open it from the archive folder?")
+    self:commonProcess(false, copy_done_text)
+end
+
+function MoveToArchive:commonProcess(is_move_process, moved_done_text)
+    if not archive_dir_path then
+        self:showNoArchiveConfirmBox()
+        return
+    end
```
poire-z, line 90: Dunno if you should not check for `self.archive_dir_path` existence in isActionEnabled().
On the first use, a user may belive the action was done after it has set the archive dir, which it has not. So, may be while the archive dir is not set, have the menu items disabled?

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,159 @@
...
+    local dest_file = string.format("%s/%s", archive_dir_path, filename)
+    require("readhistory"):updateItemByPath(document_full_path, dest_file)
+    ReadCollection:updateItemByPath(document_full_path, dest_file)
+    -- Update last open file.
+    if G_reader_settings:readSetting("lastfile") == document_full_path then
+        G_reader_settings:saveSetting("lastfile", dest_file)
+    end
+
+    self:showConfirmBox(moved_done_text, _("Ok"), function () ReaderUI:showReader(archive_dir_path .. filename) end,
```
Frenzie, line 114: Please don't pass OK as text, as I said before it's already the default, on top of which it should be replaced by something more appropriate if at all reasonably possible.

See http://koreader.rocks/doc/modules/ui.widget.confirmbox.html

mwoz123:
changes done

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local logger = require("logger")
+local util = require("frontend/util")
+local BaseUtil = require("ffi/util")
+local _ = require("gettext")
+
+local MoveToArchive = WidgetContainer:new{
+    name = "move2archive",
+    move_to_archive_settings = LuaSettings:open(("%s/%s"):format(DataStorage:getSettingsDir(), "move_to_archive_settings.lua")),
```
poire-z, line 19: You can move this one in` MoveToArchive:init()`, that's where I think this should be done.

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+    if not self.archive_dir_path then
+        self:showNoArchiveConfirmBox()
+        return
+    end
+    local document_full_path = self.ui.document.file
+    local filename
+    self.last_copied_from_dir, filename = util.splitFilePathName(document_full_path)
+
+    logger.dbg("MoveToArchive: last_moved/copied_from_dir :", self.last_copied_from_dir)
+    self.move_to_archive_settings:saveSetting("last_copied_from_dir", ("%s/"):format(self.last_copied_from_dir))
```
poire-z, line 95: Might need a `self.move_to_archive_settings:flush()` (as you did below in `MoveToArchive:setArchiveDirectory()`) to have it saved if you want to have that saving done each time.
_Or_ do that in a `MoveToArchive:onFlushSettings()` so it's done only once, when exiting a book.

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+    local dest_file = string.format("%s/%s", self.archive_dir_path, filename)
+    ReadHistory:updateItemByPath(document_full_path, dest_file)
+    ReadCollection:updateItemByPath(document_full_path, dest_file)
+    -- Update last open file.
+    if G_reader_settings:readSetting("lastfile") == document_full_path then
+        G_reader_settings:saveSetting("lastfile", dest_file)
+    end
+
+    UIManager:show(ConfirmBox:new{
+    text = moved_done_text,
```
poire-z, line 114: indentation

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+        ok_callback = function ()
+            ReaderUI:showReader(self.archive_dir_path .. filename)
+        end,
+        cancel_callback = function ()
+            if DocSettings:open(dest_file).data.percent_finished == 1 then
```
poire-z, line 119: (We should use `:readSetting("percent_finished")` instead of accessing the internal `data` table, but I see other plugins do that :/)

Anyway, that might be out of that plugin contract, to remove an (archived) finished book from History.
I would personally don't care about having it removed and having a clean History, but @Frenzie might :) He likes to keep even deleted files in History - and you're not even deleting it here: you have moved it into the archive directory.
Unless it can feel like implied by "Cancel". Or you should ask the user with another ConfirmBox?

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+            self.move_to_archive_settings:flush()
+        end,
+    }:chooseDir()
+end
+
+function MoveToArchive:showNoArchiveConfirmBox()
+    UIManager:show(ConfirmBox:new{
+        text = _("No archive directory.\nDo you want to set it now?"),
+        ok_text = _("Set archive folder"),
+        ok_callback = self.setArchiveDirectory,
```
poire-z, line 142: Again not using `:`, so not providing `self` to `MoveToArchive:setArchiveDirectory()`, where you expect it.
`ok_callback = function() self:setArchiveDirectory() end,`

plugins/movetoarchive.koplugin/main.lua
mwoz123: Done

poire-z:
No further comment, once you have fixed your last typo/quick edit (and push without testing).
But please have a full re-test of the whole thing in all conditions (including when no setting yet).

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,165 @@
...
+    end
+
+    UIManager:show(ConfirmBox:new{
+        text = moved_done_text,
+        ok_callback = function ()
+            ReaderUI:showReader(self.archive_dir_path .. filename)
+        end,
+        cancel_callback = function ()
+            if DocSettings:open(dest_file):readSetting("percent_finished") == 1 then
+                ReadHistory:removeItemByPath(dest_file)
```
Frenzie, line 122: This is weird.

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,165 @@
...
+            end
+            self:openFileBrowser(self.last_copied_from_dir)
+        end,
+    })
+end
+
+
```
Frenzie, line 129: stray

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1052,6 +1052,23 @@ function FileManager:moveFile(from, to)
     return BaseUtil.execute(self.mv_bin, from, to) == 0
 end
 
+--[[
+A shortcut to execute cp command (self.cp_bin) with from and to as parameters.
+Returns a boolean value to indicate the result of cp command.
+--]]
+
```
Frenzie, line 1059: stray

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1052,6 +1052,23 @@ function FileManager:moveFile(from, to)
     return BaseUtil.execute(self.mv_bin, from, to) == 0
 end
 
+--[[
```
Frenzie, line 1055:
```suggestion
--[[--
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1052,6 +1052,23 @@ function FileManager:moveFile(from, to)
...
+A shortcut to execute cp command (self.cp_bin) with from and to as parameters.
+Returns a boolean value to indicate the result of cp command.
+--]]
+
+function FileManager:copyFileFromTo(from, to)
+    return BaseUtil.execute(self.cp_bin, from, to) == 0
+end
+
+--[[
```
Frenzie, line 1064:
```suggestion
--[[--
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1052,6 +1052,23 @@ function FileManager:moveFile(from, to)
     return BaseUtil.execute(self.mv_bin, from, to) == 0
 end
 
+--[[
+A shortcut to execute cp command (self.cp_bin) with from and to as parameters.
+Returns a boolean value to indicate the result of cp command.
+--]]
```
Frenzie, line 1058: I think that should be just like this, unless ldoc is happy to ignore it.
```suggestion
]]
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1052,6 +1052,23 @@ function FileManager:moveFile(from, to)
...
+
+function FileManager:copyFileFromTo(from, to)
+    return BaseUtil.execute(self.cp_bin, from, to) == 0
+end
+
+--[[
+A shortcut to execute cp recursive command (self.cp_bin) with from and to as parameters.
+Returns a boolean value to indicate the result of cp command.
+--]]
```
Frenzie, line 1067:
```suggestion
]]
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1052,6 +1052,23 @@ function FileManager:moveFile(from, to)
     return BaseUtil.execute(self.mv_bin, from, to) == 0
 end
 
+--[[
+A shortcut to execute cp command (self.cp_bin) with from and to as parameters.
+Returns a boolean value to indicate the result of cp command.
```
Frenzie, line 1057:
```suggestion
@treturn boolean result of cp command
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1052,6 +1052,23 @@ function FileManager:moveFile(from, to)
...
+--]]
+
+function FileManager:copyFileFromTo(from, to)
+    return BaseUtil.execute(self.cp_bin, from, to) == 0
+end
+
+--[[
+A shortcut to execute cp recursive command (self.cp_bin) with from and to as parameters.
+Returns a boolean value to indicate the result of cp command.
```
Frenzie, line 1066:
```suggestion
@treturn boolean result of cp command
```

frontend/apps/filemanager/filemanager.lua
Frenzie, line 1055: Here's a proper example:

https://github.com/koreader/koreader/blob/be72203b8302730b69f92b34acecb97baa96a83a/frontend/util.lua#L107-L111

Frenzie: The suggestions I included are for the proper multiline style, but there is no reason here to use multiline comments.

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,165 @@
...
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local logger = require("logger")
+local util = require("frontend/util")
+local BaseUtil = require("ffi/util")
+local _ = require("gettext")
+
+local MoveToArchive = WidgetContainer:new{
+    name = "move2archive",
+    move_to_archive_settings = nil,
```
Frenzie, line 19: Why so verbose?

mwoz123: for readablity. 
other plugins have the same thing e.g. : plugins/evernote.koplugin/main.luam plugins/perceptionexpander.koplugin/main.lua etc

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1044,14 +1044,25 @@ function FileManager:showFiles(path, focused_file)
     self.instance = file_manager
 end
 
---[[
-A shortcut to execute mv command (self.mv_bin) with from and to as parameters.
-Returns a boolean value to indicate the result of mv command.
---]]
+
```
Frenzie, line 1047:
```suggestion
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1044,14 +1044,25 @@ function FileManager:showFiles(path, focused_file)
...
+-- A shortcut to execute mv command (self.mv_bin) with from and to as parameters.
```
Frenzie, line 1048: It's not a normal comment, so there are a few special conventions. Otherwise you don't get those nice auto-generated docs like at http://koreader.rocks/doc/modules/util.html#lastIndexOf Three dashes is the activation method.
```suggestion
--- A shortcut to execute mv command.
```

If you want to add a comment about from and to, use
```
-- @param from description goes here
```

But just saying "with from and to as parameters" is redundant and distracting.

plugins/movetoarchive.koplugin/main.lua
Frenzie: I can't tell what you're referring to in EverNote, but PerceptionExpander surely does not:

https://github.com/koreader/koreader/blob/be72203b8302730b69f92b34acecb97baa96a83a/plugins/perceptionexpander.koplugin/main.lua#L16-L28

Note how it says `settings`, not `perception_expander_settings`.

NB Other plugins do it too is not a good argument.

mwoz123:
done

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1044,14 +1044,24 @@ function FileManager:showFiles(path, focused_file)
...
     return BaseUtil.execute(self.mv_bin, from, to) == 0
 end
 
+--- A shortcut to execute cp command (self.cp_bin) with from and to as parameters.
+-- @treturn boolean result of cp command
+function FileManager:copyFileFromTo(from, to)
+    return BaseUtil.execute(self.cp_bin, from, to) == 0
+end
+
+--- A shortcut to execute cp recursive command (self.cp_bin) with from and to as parameters.
```
Frenzie, line 1059:
```suggestion
--- A shortcut to execute cp recursively.
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1044,14 +1044,24 @@ function FileManager:showFiles(path, focused_file)
...
-A shortcut to execute mv command (self.mv_bin) with from and to as parameters.
-Returns a boolean value to indicate the result of mv command.
---]]
+--- A shortcut to execute mv command.
+-- @treturn boolean result of mv command
 function FileManager:moveFile(from, to)
     return BaseUtil.execute(self.mv_bin, from, to) == 0
 end
 
+--- A shortcut to execute cp command (self.cp_bin) with from and to as parameters.
```
Frenzie, line 1053:
```suggestion
--- A shortcut to execute cp.
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -1044,14 +1044,24 @@ function FileManager:showFiles(path, focused_file)
     self.instance = file_manager
 end
 
---[[
-A shortcut to execute mv command (self.mv_bin) with from and to as parameters.
-Returns a boolean value to indicate the result of mv command.
---]]
+--- A shortcut to execute mv command.
```
Frenzie, line 1047:
```suggestion
--- A shortcut to execute mv.
```

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+                         })
+                        return
+                    end
+                    self:openFileBrowser(self.last_copied_from_dir)
+                end,
+            },
+            {
+                text = _("Set archive directory"),
+                keep_menu_open = true,
+                callback =  function ()
```
Frenzie, line 70:
```suggestion
                callback = function()
```

plugins/movetoarchive.koplugin/main.lua
```diff
@@ -0,0 +1,161 @@
...
+    }
+end
+
+function MoveToArchive:moveToArchive()
+    local move_done_text = _("Book moved.\nDo you want to open it from the archive folder?")
+    self:commonProcess(true, move_done_text)
+end
+
+function MoveToArchive:copyToArchive()
+    local copy_done_text =_("Book copied.\nDo you want to open it from the archive folder?")
```
Frenzie, line 84:
```suggestion
    local copy_done_text = _("Book copied.\nDo you want to open it from the archive folder?")
```

poire-z:
Looks fine, no more comment.

Frenzie:
Thanks @mwoz123 !

poire-z:
Can't be disabled. I guess it needs the names to be coherent with the plugin directory name:
```diff
--- a/plugins/movetoarchive.koplugin/_meta.lua
+++ b/plugins/movetoarchive.koplugin/_meta.lua
@@ -1,6 +1,6 @@
 local _ = require("gettext")
 return {
-    name = "move_to_archive",
+    name = "movetoarchive",
     fullname = _("Move to archive"),
     description = _([[Moves/copies current book to archive folder]]),
 }
diff --git a/plugins/movetoarchive.koplugin/main.lua b/plugins/movetoarchive.koplugin/main.lua
index f575a849..08cba643 100644
--- a/plugins/movetoarchive.koplugin/main.lua
+++ b/plugins/movetoarchive.koplugin/main.lua
@@ -14,7 +14,7 @@ local BaseUtil = require("ffi/util")
 local _ = require("gettext")

 local MoveToArchive = WidgetContainer:new{
-    name = "move2archive",
+    name = "movetoarchive",
 }

 function MoveToArchive:init()
```

Robertmccalister747:
Thank you for giving me the time of day lol

Issue #5938, opened 2020-03-10
Request: progress sync gesture


dowiew:
* KOReader version: latest
* Device: any

Hello, I frequently use the progress sync feature between Kindle at home and mobile phone in a bus to work or back. Just before I step out of the bus I have to sync, but it takes too much time: swipe to menu +tool icon +Progress sync +Push. It's hard to do with one hand (the second is busy to keep the bus handle).
I can't see any possibility to assign a gesture to push/pull the progress.
Such a feature would be very nice.

Thank you for your great work!
Wieslaw

fazlerabbi37:
@Frenzie Hi, new user here. I just installed KOReader AppImage from the latest release and followed the OP's instruction but couldn't find Progress sync plugin. I tried the lasted release for Android from GitHub and F-Droid but the result is same. What am I doing wrong?

Frenzie:
@fazlerabbi37 The menu differs depending on if you have a document open or not. Btw, the forum's better suited for asking user to user questions.

fazlerabbi37:
> Btw, the forum's better suited for asking user to user questions.

I know this is not appropriate to ask those questing here instead of forum but I asked for two reason.
- I don't have an account in the forum
- Maybe we should have a clear documentation saying how to access and use it for new users like me?

Would you mind if I update the Wiki?

Frenzie:
Absolutely, that's what the wiki's for!

PR #6103, opened 2020-04-29
[Gesture] Add progress sync gesture

robert00s:
New gesture:
- Push progress from this device
- Pull progress from other devices

Close: #5938

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6103)

### DIFF
--- a/plugins/kosync.koplugin/main.lua
+++ b/plugins/kosync.koplugin/main.lua
@@ -105,18 +105,6 @@ local function validateUser(user, pass)
     end
 end
 
-function KOSync:init()
-    local ReaderGesture = require("apps/reader/modules/readergesture")
-    ReaderGesture.KOSyncPushProgress = function()
-        if not self.kosync_userkey then return end
-        self:updateProgress(true)
-    end
-    ReaderGesture.KOSyncPullProgress = function()
-        if not self.kosync_userkey then return end
-        self:getProgress(true)
-    end
-end
-
 function KOSync:onReaderReady()
     local settings = G_reader_settings:readSetting("kosync") or {}
     self.kosync_custom_server = settings.custom_server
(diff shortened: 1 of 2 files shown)


plugins/kosync.koplugin/main.lua
```diff
@@ -105,6 +105,18 @@ local function validateUser(user, pass)
...
+    ReaderGesture.KOSyncPushProgress = function()
+        if not self.kosync_userkey then return end
+        self:updateProgress(true)
+    end
+    ReaderGesture.KOSyncPullProgress = function()
+        if not self.kosync_userkey then return end
+        self:getProgress(true)
+    end
+end
+
```
yparitcher, line 119: It might be better to use an event here?
something like (untested)
```lua
function KOSync:onKOSyncPushProgress()
    if not self.kosync_userkey then return end
    self:updateProgress(true)
end
```
called via
```lua
self.ui:handleEvent(Event:new("KOSyncPushProgress"))
```
for each of push and pull

robert00s: Thanks for suggestion but my code base on that PR: #5167 and #6035.
Should I change all to use events instead solutions with ReaderGesture.someFunction()?

Frenzie: If you want, sure. It's probably not a very big deal. What do you think @poire-z ?

yparitcher: Please use events, at least for new code.
It makes the code easier to read.
Also I am in middle of putting together a way to dispatch many events at once 
(to allow for profiles and gestures that do 2 things) which will only work with events.
Thanks

Frenzie: Right, I was just talking about the existing code.

robert00s: Ok. Done.

Issue #6105, opened 2020-04-29
Reorganize "more plugins" submenu


mwoz123:
* KOReader version: git mater 
* Device: all

#### Issue
Extracted oftopics from : 
https://github.com/koreader/koreader/pull/6101

"do we still need "more plugins" as there's something like "next page" in menu if there're too many entries?

Imho it's misleading."

"I guess the more_plugins menu isn't really needed anymore, now that the user can disable the plugins he doesn't need.
If you put some in more_plugins, no matter the user cleaning, he would have to tap once more to reach that unfortunate plugin we put into more_plugins.
If we'd want to keep more_plugins, only non-user-action plugins (those with only settings or some permanent behaviour) should go in there.
So, I wouldn't mind an initial Plugins menu with 20 items :) as I can clean it up to the 5 that matters to me., "
Source : https://github.com/koreader/koreader/pull/6101

#### Steps to reproduce
Open menu plugins although there's kind of" next page " functionality there's also" more plugins "

.

Frenzie:
Simply doing what I wrote in https://github.com/koreader/koreader/issues/5461#issuecomment-604938066 is a thing one could do, but I changed the title because I don't want that to be presupposed to be the solution.

hius07:
I would suggest the menu as follows:
Plugin management
-separator-
Text editor
-separator-
all other plugins

Frenzie:
I should clarify the menu vision. The `More plugins` menu is vestigial in its current form, the name perhaps most of all. As @mwoz123 said, it's misleading. For the sake of thinking about order it should be regarded as `Miscellaneous tools` or `More tools`. It looks like neglecting to change that name may well be the biggest mistake in this minor bit of laziness. ;-)

`More tools` is a submenu in the tools menu, not in the plugins menu. That everything in there happens to be plugins is merely a technical detail and not considered part of the unifying menu vision. Plugin management should be last as it is because it's only used once in a blue moon, if it should be in the tools menu at all. The same applies to settings more broadly. Putting plugin management in the tools menu is traditional, however.

Plugins should not and are not supposed to be most at home in the tools menu. Those plugins that happen to be in the tools menu are by and large there because that's where they fit best. Those that don't are new and I didn't have the heart or energy to make much of a fuzz about it provided they had a reasonable claim to the tools menu. That includes plugins like tweak document settings which should be more at home in settings → document.

[...]

PR #6106, opened 2020-05-01
Dispatcher: for profiles & gestures

yparitcher:
This PR is still very much a RFC draft so give you comments and criticisms on the concept but please don't nitpick :smile:

add a Dispatcher module that allows for dispatching multiple events at once.
This will allow for profiles & for gestures that do multiple things.

it has 2 methods: 
`Execute` which is given a kv table of settings to change and fires an event for each of them.
`addSubMenu` adds a menu item to a menu to allow for modifying which events are called
it also has `settingsList` which is a master table of all allowed setting and their corresponding info
(it is mostly from `ReaderGesture` and needs a lot of work)

to allow for a new setting all one has to do is add a entry to `settingsList` with a corresponding event and it will work out of the box.

the profile plugin is right now still a stub, just to test `Dispatcher`. the plan is to finish it and eventually refactor `ReaderGesture` to rely on this. 
This also needs effort to move many functions out of reader gesture into events where they belong.

 **A few things i still am not sure about:**

- Right now `ReaderGesture` has conditions (if the document is open, etc) however if we are using events they are unnecessary and i am considering dropping them.
- I currently allow for a user to set conflicting setting (this shouldn't be dangerous) It just will try to fire all of them and the unexpected could happen (there is no guarantee on order) ie: if a user add `gotochapter` & `changepage` +10 they might end up on the tenth page of the next chapter.
- I need help with the name of an appropriate widget to display a radio/check list of options for the `string` category.
- I am sure there will be more as I remember them.
- I am thinking of replacing `values` with a function pointer that if available can reference the function that built the list for the UI (think fonts or kerning methods etc)

I would like to add unit test eventually and will need help with how they work (but that is later)
The very long term goal is to allow for any UI setting to be automated

Enjoy
(This all happened because I got frustrated that I could not have one gesture to toggle night mode & the backlight together )

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6106)


frontend/dispatcher.lua
```diff
@@ -0,0 +1,258 @@
...
+    title: for use in ui.
+and optionally
+    min/max: for number
+    values: allowed values for string.
+    conditions: (is document open, etc.)
+--]]--
+local settingsList = {
+    page_jmp = { ["category"]="incrementalnumber", ["event"]="GotoViewRel" , ["title"]=_("Go %1 pages"),["min"]=-50, ["max"]=50},
+    prev_chapter = { ["category"]="none", ["event"]="GotoPrevChapter", ["title"]=_("Previous chapter"), },
+    next_chapter = { ["category"]="none", ["event"]="GotoNextChapter", ["title"]=_("Next chapter"), },
```
poire-z, line 24: Sorry, this is a nitpick, but rather too early than too late :)
This hurts my eyes :)

For the sake of readability, if your keys are always that simple, just use them as is:
`next_chapter = { category="none", event="GotoNextChapter", title=_("Next chapter"), },`

And even, if there's required and optional items, just use an array:
```lua
page_jmp =     { "incrementalnumber", "GotoViewRel" ,     _("Go %1 pages"),   -50, 50},
next_chapter = { "none",              "GotoNextChapter",  _("Next chapter"), },
```
You can then, if you really need it to be accessible by keys, build another (module local) table with these values as associated arrays.

poire-z:
I can't really comment much, as I'm not a customer for this feature. So, I'd be more concerned about it not impacting too much the code base - and it looks currently fine.
Anyway, some thoughts:
- Can you post a screenshot of how the menu and some widget would look?
- So, a user will see a menu with all possible items, check the one he wants in his profile, set some value - and activate that profile by some yet-not-included-in-this-PR way (and create new profiles by some yet-not-included-in-this-PR way), right?
- Any thought of the other way to handle that I suggested in https://github.com/koreader/koreader/pull/5079#issuecomment-502403519 - just to avoid these many widgets, list widgets, need to duplicate settings values strings. (But I reckon this won't work for actions, like page+10, go to end of book...)
- May be to avoid duplicating/spreading settings' values strings and bits in that new module, other modules that are dealing with these settings could just register into this module their own stuff.
- Possibly some concerns with events that do scheduleIn/nextTick (dunno if there are) that would trigger something like redraw/rerender/updateToc - triggering multiple ones would stack these scheduled actions that could do the same thing multiple times (although I guess, for what I know: crengine, you would set a bunch of settings, and it will only do a re-render once when drawing the page next time it can, so they could all coalesce in a single re-rendering - and most actions might probably have checks to not do stuff when not needed).

> I need help with the name of an appropriate widget to display a radio/check list of options for the string category.

There's already radiobuttontable.lua. If you need a higher level one, dunno, OptionListWidget ? :)

> I am thinking of replacing values with a function pointer that if available can reference the function that built the list for the UI (think fonts or kerning methods etc)

If the modules can register their stuff (my 4th point above), you could may be have `readerconfig.lua` / `configdialog.lua` fill a bunch of them when they handle `frontend/ui/data/creoptions.lua` and `koptoptions.lua`.

Frenzie:
Looks reasonable enough at a glance; I don't really have anything to add to what @poire-z already wrote right now.

frontend/dispatcher.lua
yparitcher: thanks,

an array makes this easier to read but the code harder to understand, so i am not sure if i will go there.

yparitcher:
> I can't really comment much, as I'm not a customer for this feature. So, I'd be more concerned about it not impacting too much the code base - and it looks currently fine.

This is a big concern of mine, and in this PR the only thing would be to clean up the functions at the end of `ReaderGesture` and refactor them into events where they belong.(i might do this in a separate PR first once i know that this one is looking good)
In the future however the effect should be limited to whoever wants a setting to be available will have to make sure it has an event handler (similar to how a gesture event is currently added).
however your 4th piont might affect this.

> Anyway, some thoughts:
> 
>     * Can you post a screenshot of how the menu and some widget would look?

Yes.

>     * So, a user will see a menu with all possible items, check the one he wants in his profile, set some value - and activate that profile by some yet-not-included-in-this-PR way (and create new profiles by some yet-not-included-in-this-PR way), right?

i imagine the UI (at least for profiles) to be: a menu with a list of profiles and a option to add a new one, click a profile activates it, hold the profile and a menu comes up with all possible items to check, when checking one it is saved to the profile.
More work will be done to make those profiles file based and persistent etc.

>     * Any thought of the other way to handle that I suggested in [#5079 (comment)](https://github.com/koreader/koreader/pull/5079#issuecomment-502403519) - just to avoid these many widgets, list widgets, need to duplicate settings values strings. (But I reckon this won't work for actions, like page+10, go to end of book...)

For event that are just called (`gotohistory`) and for toggles it is just a checkbox.
number based actions instead of having many fixed actions `go1page` and `go10pages` it makes more sense to have one `gopages` that opens a numberpicker.
the hard one is string based settings, i will address next.

>     * May be to avoid duplicating/spreading settings' values strings and bits in that new module, other modules that are dealing with these settings could just register into this module their own stuff.

This is a great idea that i had not thought of.
Pros: cleaner, less duplication.
Cons: this might require many changes in other parts of the code. a module can only register once it is brought up so there might be ordering issues to be ironed out. (the setting might no be available yet to change)

I am thinking that for simple actions/ toggles/ numbers to leave as is, but for string actions to replace the values array with a function callback that the other module can have (it might be the one that draws the existing UI) to return a list of available values. i think it would be a good balance between modularity and non duplication.
parsing `readerconfig.lua / configdialog.lua` might be a good idea, however that is only setting from the bottom menu :)

>     * Possibly some concerns with events that do scheduleIn/nextTick (dunno if there are) that would trigger something like redraw/rerender/updateToc - triggering multiple ones would stack these scheduled actions that could do the same thing multiple times (although I guess, for what I know: crengine, you would set a bunch of settings, and it will only do a re-render once when drawing the page next time it can, so they could all coalesce in a single re-rendering - and most actions might probably have checks to not do stuff when not needed).

I was going to ask you about this but forgot, for example if one set the font & font size it might rerender multiple times unnecessarily. i will test this.

NiLuJe:
I vaguely seem to remember that there's already stuff in place to handle deduplication of events (for the kind raised in your final point), because DocSettings on open happens to also be firing all of them at once, IIRC.

Frenzie:
Events or just overlapping screen updates?

NiLuJe:
Overlapping screen updates will definitely get coalesced by UIManager ;).

Frenzie:
Right, I just don't recall events doing that. ^_^

yparitcher:
I figured out how to parse all the bottom menu settings from `creoptions.lua/koptoptions.lua` i did not add the list yet as i am waiting for #6122 to settle. but the parser is done.

the bottom menu is all events so easy to parse, the top menu setting will be a little harder.

before i do the top menu settings i will probably open seperate PR's to cleanup `ReaderGesture` into proper events so i can use those events.

The profile plugin is mostly finished.

yparitcher:
screenshots:
Reader_2020-05-06_215504
Reader_2020-05-06_215850
Reader_2020-05-06_215858
Reader_2020-05-06_215605

yparitcher:
Ready for review:

This add a profile plugin. this can be used with any event registered with dispatcher.lua.

for now I added support for all `creoptions` (the bottom crengine menu) items. I  plan on adding (in further PR's) all the other events spread out in the top menu. for `koptoptions` (mupdf) it looks like many of the items just change a setting and do not call an event, so it will need further work to integrate. In the long term I would like to have cleanup `ReaderGesture` to rely on this.

As i mentioned before I have no way of knowing why CircleCi builds failed as they have stopped allowing me to access the logs, unless i login and give the R/W access to my repos :(

frontend/dispatcher.lua
```diff
@@ -0,0 +1,264 @@
+local CreOptions = require("ui/data/creoptions")
+local Event = require("ui/event")
+local KoptOptions = require("ui/data/koptoptions")
```
Frenzie, line 3: This variable isn't used.

frontend/dispatcher.lua
```diff
@@ -0,0 +1,264 @@
...
+    })
+    for k, v in pairs(settingsList) do
+            if settingsList[k].category == "none" then
+                table.insert(menu, {
+                   text = settingsList[k].title,
+                   checked_func = function()
+                   return self[location][settings] ~= nil and self[location][settings][k] ~= nil
+                   end,
+                   callback = function(touchmenu_instance)
+                      if self[location][settings] ~= nil and self[location][settings][k] then self[location][settings][k] = nil else self[location][settings][k] = true end
```
Frenzie, line 126: Readability-wise I'd rather you space this out a bit more. ;-)

frontend/dispatcher.lua
yparitcher: thanks

frontend/ui/elements/reader_menu_order.lua
```diff
@@ -124,6 +124,7 @@ local order = {
         "news_downloader",
         "send2ebook",
         "text_editor",
+        "profiles",
```
Frenzie, line 127: How does this affect the paging?

plugins/profiles.koplugin/main.lua
```diff
@@ -0,0 +1,128 @@
...
+                UIManager:show(name_input)
+                name_input:onShowKeyboard()
+            end,
+            separator = true,
+        }
+    }
+    for k,v in pairs(self.data) do
+        local sub_items = {
+            {
+                text = _("Delete Profile"),
```
Frenzie, line 89:
```suggestion
                text = _("Delete profile"),
```

frontend/ui/elements/reader_menu_order.lua
yparitcher: Reader_2020-06-17_170154

Frenzie: That probably means it pushes plugin management to the next page on 3:2 devices. But wouldn't it be better in something like taps & gestures regardless?

frontend/dispatcher.lua
```diff
@@ -123,7 +123,12 @@ function Dispatcher:addSubMenu(menu, location, settings)
                    return self[location][settings] ~= nil and self[location][settings][k] ~= nil
                    end,
                    callback = function(touchmenu_instance)
-                      if self[location][settings] ~= nil and self[location][settings][k] then self[location][settings][k] = nil else self[location][settings][k] = true end
+                      if self[location][settings] ~= nil
+                      and self[location][settings][k] then
```
Frenzie, line 127: Might be going a bit overboard. ;-P

frontend/ui/elements/reader_menu_order.lua
yparitcher: I put it there by default (this type of thing was discussed in another issue I believe). I am happy to put it somewhere else, once everyone voices their opinion. I only want to move it once :)

Frenzie: For onlookers, that would be https://github.com/koreader/koreader/issues/6105#issuecomment-621653800

plugins/profiles.koplugin/main.lua
```diff
@@ -0,0 +1,128 @@
...
+                logger.dbg("Profile callback")
+                Dispatcher.execute(self, v)
+            end,
+        })
+    end
+    return sub_item_table
+end
+
+function Profiles:newProfile(name)
+    if self.data[name] == nil then
```
Frenzie, line 119: More of a minor detail but it seems like where this is used it should perhaps warn that it already exists or something?

yparitcher: If the profile exists, it will do nothing, leaving the old profile as is.

Frenzie: Right, it's a minor detail because it's not dangerous, but the effect is the same as cancel.

yparitcher: It is in the callback, so it would be hard to  bring them back to the input dialog (the best idea), and I think a warning will just scare a user unnecessarily.

Frenzie: I'm not sure I follow, just don't close the dialog? Along these lines:

```lua
                                local name = name_input:getInputText()
                                if not self:newProfile(name) then
                                    -- InfoMessage or something
                                    return
                                end
                                UIManager:close(name_input)
```

yparitcher: wasnt thinking :)

yparitcher:
I did not add the profiles to the FM menu as right now there are only book bases setting available, but it still comes up in the menu.
How to i specify that i dont want it?

FileManager_2020-06-17_183607

frontend/dispatcher.lua
```diff
@@ -0,0 +1,279 @@
+local CreOptions = require("ui/data/creoptions")
+local Event = require("ui/event")
+local Screen = require("device").screen
+local UIManager = require("ui/uimanager")
+local T = require("ffi/util").template
+local _ = require("gettext")
+--local logger = require("logger")
```
Frenzie, line 7:
```suggestion
```

NiLuJe:
Yup, looks pretty good ;). :+1:

poire-z:
Some issue with this: the bottom toggles aren't updated to reflect what's been fired and correctly applied.
(I have the feeling that this might not be easy to solve, as the toggles manage their own state, and were supposed to be the only source for changing the settings, so there's no retrofeedback)

yparitcher:
> Some issue with this: the bottom toggles aren't updated to reflect what's been fired and correctly applied.
> (I have the feeling that this might not be easy to solve, as the toggles manage their own state, and were supposed to be the only source for changing the settings, so there's no retrofeedback)

would calling `ConfigDialog:update()` (the function that is used when a setting is canged from a toggle) when the menu is opened help?

poire-z:
Haven't tried it, but I guess that if I execute a profil, then show the bottom pannel (so, wrong value displayed), toggle an other toggle (which should have called :update()) : the wrong value is still displayed - so I'd say it doesn't.

yparitcher:
> Haven't tried it, but I guess that if I execute a profil, then show the bottom pannel (so, wrong value displayed), toggle an other toggle (which should have called :update()) : the wrong value is still displayed - so I'd say it doesn't.

right, update is called per item not on the whole menu :(

hius07:
Thank you for a nice feature!

Assume I created a new profile (with many changes in settings) and have applied it.
Is there a quick way to return to my previous state of the settings?

yparitcher:
no,

maybe in the future but first i need to iron out the current wrinkles

yparitcher:
> Well, and finally I got KOReader crashed as follows:
> 
>     * create a new profile with None actions
> 
>     * apply it with a long tap
> 
>     * call the bottom menu
> 
>     * try to open Margins tab
> 
>     * crash with no records in crash.log
> 
> 
> PW2 5.8.7, v2020.06-25-g3220752_2020_06_26
> 
> Got the same crash confirmation from a PW4 user.

thanks for letting me know, fixed in #6324

Galunid:
@yparitcher Cool stuff, one thing I've found out is when you create new profile and set orientation to "portrait", it's upside down 🙃

yparitcher:
> @yparitcher Cool stuff, one thing I've found out is when you create new profile and set orientation to "portrait", it's upside down upside_down_face

I know, it was because of the whole rotation mess, fixed in #6309

PR #6112, opened 2020-05-03
Word Expansion: use of letter spacing for justification

poire-z:
bump crengine:
- Text: fix occasional BiDi bad word splitting https://github.com/koreader/crengine/pull/341
- Font: fix HB fallback measurement/drawing mismatches
- Font: do not add letter spacing on diacritics
- Text: tunable use of letter spacing for justification
- Text: dont adjust space after initial quotation mark/dash (rework)
- Text: fix possible bad widths after collapsed spaces https://github.com/koreader/crengine/pull/342

xtext.cpp: pick crengine fix with HB fallback measurement/drawing mismatches
Also bump base: update FBInk, cURL; enable scrensaver on SDL

Making this available here for suggestions about the name and value labels for this new bottom toggle, as I don't really like `Word Stretching`.
Previous discussions (and alternative layouts for this bottom menu) at https://github.com/koreader/crengine/pull/341#issuecomment-623018904 and https://github.com/koreader/koreader/pull/5628#issuecomment-622354587.

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

Word spacing has been moved out of the `B` menu (and the new one not added there) so that these menu don't take too much space over the text, allowing to see their effect on more of the text:
<kbd>image</kbd>

Rewording suggestions welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6112)


NiLuJe:
Nothing better than "Letter Spacing" comes to mind right now ;).

mergen3107:
Letter spacing  sounds familiar and obvious, yeah

poire-z:
Some stuff about text justification at https://drafts.csswg.org/css-text-3/#text-justify-property.
Some [previous version](https://www.w3.org/TR/2003/CR-css3-text-20030514/#justification-prop) of this draft (from 2003, wow, that CSS3 is an old draft) had a keyword that allowed this behaviour: "newspaper".
> newspaper
    Selects the justification behavior in which both inter-word and inter-letter spacing can be expanded or reduced to spread the text across the whole line. Also, text distribution on any given line may depend on the layout or the contents of the previous or the following several lines. This is the significantly slower and more sophisticated type of the full justify behavior preferred in newspaper and magazines, as it is especially useful for narrow columns. For example, typically, compression is tried first. If unsuccessful, expansion occurs: inter-word spaces are expanded up to a threshold, and finally inter-letter expansion is performed.

Some thoughts, just to give some inspirations to others :)
`Space Reduction`
`Space Killing: hard | soft | off`

poire-z:
> Letter spacing sounds familiar and obvious

Yes, familiar in its generic sense.
But here, it would only be used under some conditions, on 1 line out of 5-10.
So, if we find nothing better than `Letter Spacing`, we should find some better adjective/labels for the toggle buttons, to express permission/_allowness_, `allow little`, `allow more`....

mergen3107:
Maybe we should put `Letter spasing*` and thrn explain this thing in its desscription?

poire-z:
Or:
`Justification: classic | stretch words | explode words`
(or some better words for the level of letter spacing allowed)

hius07:
Increasing from left to right: off, small, large.

mergen3107:
How about `Spread letters on a line`?

poire-z:
> Increasing from left to right: off, small, large.

I somehow agree that it should be in the classic order of "increasing" - even if the other way around makes it work against blank space in the same direction as Word spacing (small word spacing + large word strecthing = minimal blank space - large word space + no word strecthing = maximal blank space).

> How about `Spread letters on a line`?

Bit too long to fit in the widget :)

Frenzie:
Letter spacing is misleading. It has to do with letter spacing of course, but not in the way I interpret the term.

Justification potentially has a somewhat similar problem, but I think labels can help more there to alleviate the problem.

I'm inclined to think the stretching/compression/reduction line of thought sounds the most fruitful. What about just `compression`?

@mergen3107 Something that long would just read `Spread letters…` ;-)

mergen3107:
Or maybe options like “compact”, “tight”, “off”?

poire-z:
> What about just _compression_?

Which would enable word _expansion_ :) might be confusing.
But it depends on what the toggle labels are, how they can make the title clearer.

poire-z:
GitHub might be losing comments (or @mergen3107 removed it), but got his suggestion by github mail, which might bring some new ideas:
> Or maybe options like `compact`, `tight`, `off`

Frenzie:
> Which would enable word expansion :) might be confusing.

Okay, so what about just expansion? :-P

poire-z:
> Okay, so what about just expansion

Could work for me, sounds better than "stretching". Just `Expansion `or `Word Expansion`?
If we could have labels other than small|large, that would be even better.
`none/off | compact/tight | loose/wider` ?

NiLuJe:
Yeah, a variation on expansion works for me, too. Not quite sure there's anything better than "Word" that would more or less fit without requiring more space :s.

I still heavily tend towards the "inverted" order of options because it makes more sense in contrast to what's right above it, though ;p.

As for said options, KISS? more / less / none.

poire-z:
> I still heavily tend towards the "inverted" order of options because it makes more sense in contrast to what's right above it, though ;p.

Dunno, I usually expect the left toggle to be the off/normal/classic/regular one, as the reference.
Which is obviously not the case for Word spacing small|medium|large :) (and not for the right most bottom panel in both epub & pdf...)

NiLuJe:
Eh, more reason to have the two work in unison, no matter the individual order, then ;p.

mergen3107:
@poire-z
Yeah, I put this comment and then after a while deleted it. I thought that “compact” is not very obvious as a default option.

mergen3107:
As for another source of reference, let’s have a look at MS Word: https://support.office.com/en-us/article/change-the-spaces-between-text-e9b96011-1c42-45c0-ad8f-e8a6e4a33462

They call this feature “Character spacing” with “Normal”, “Expanded”, “Condensed”. The latter won’t apply to our case though

poire-z:
So, `Word expansion` fits on my device.
<kbd>image</kbd>

Still not fond of the inversion large > small (any other opinion on that)?

Bit better by using different words:
<kbd>image</kbd>
(dunno if having letter spacing in `w i d e` to express it is clever - probably visually disruptive when you're not interested in it :).

light, mild (léger in french)
wide, ... ?

NiLuJe:
> (dunno if having letter spacing in `w i d e` to express it is clever - probably visually disruptive when you're not interested in it :).
> 

I mean, it certainly looks off, but I like the idea in theory ;p.

The main issue I have with the option names is that they don't *quite* work with "expansion", the least terrible I could come up with on that front was

> EDIT: moderate / limited / none? (EDIT: moderate might fit better as the "small" option... maybe large / moderate / none, then?)

poire-z:
> Bonus points for keeping the (S) / M / L letter 

`Word Expansion:  Smutty | Modest | Locked`

mergen3107:
I support the idea or ordering from left to right. It looks natural to me that the default choice is the leftmost one, then other choices are in increasing order to the right. 

I also like the idea of `w i d e` with spaces, and so this thing comes in even more natural: default choice starts with default spacing on the left, and then other choices are getting wider.

Btw, there is a symbol of a thin space character (https://www.compart.com/en/unicode/U+2009), so maybe we can use it for intermediate choice? E.g. all three choices would be:
`off` , `m o d e r a t e` , `w i d e` (or whatever words we will agree to use in the end)

mergen3107:
(Not sure if Github interpreted the pasted thin spaces correctly above, or maybe it is because of monospace font for highlighted code)

NiLuJe:
Hmm, yeah, if that works, thin spaces would be neat.

poire-z:
(Yep, GitHub monospace probably.)
Thin-space would somehow truncate the word just as space does.
But hair-space (U+200A) gives a bit longer of the word:
<kbd>image</kbd>

Well, not super fond of of that spacing, and that will make it probably hard/confusing for translators.

NiLuJe:
I may have a vague preference for wider vs. wide, but none/moderate/wide(r) works for me ;).

mergen3107:
How about off-wider-widest? :D

poire-z:
<kbd>image</kbd>
with thin-spaces in "wider".

Showing looks with each toggle enabled:

<kbd>image</kbd>

<kbd>image</kbd>

Without thin-spaces:
<kbd>image</kbd>

No problem with "wider", but "moderate" feels a bit long - but ok with it if nothing else comes up.

NiLuJe:
I kinda like @mergen3107's proposition ;).

That's the kind of thing I initially wanted to come up with, but I wasn't happy with whatever I could come up with for small. But, duh, this works just as well ;).

Frenzie:
Sure, wfm.

poire-z:
<kbd>image</kbd>

Not super fan (a bit weird, weirder, weirdest :) but I can go with it, better than the long "moderate".
(Might not be easy for translators, "plus large", "encore plus large" :)
(And my non-native english feels strange using -er & -est when there is no preceeding -e - wider without a preceeding wide... But ok, wider is wider than none.)

We don't have any -er or -est. But we did in the past, for contrast:
`S.LIGHTEST , S.LIGHTER, S.DEFAULT, S.DARKER, S.DARKEST`
not used anymore.

So, we go with that?

Frenzie:
You could also say something like none, some, more.

poire-z:
For opinions:
<kbd>image</kbd>
(I like it, in spite of the similarities of the `_o_e` words :)

poire-z:
Btw, was reading that [thesis about microtypography in TeX](http://www.pragma-ade.com/pdftex/thesis.pdf) (from https://github.com/koreader/crengine/issues/307#issuecomment-524255803), and it's good we use "Word Expansion", because there is also "Font Expansion", where, to solve the same issue, it scales the font horizontally.
Now, I'm wondering how that would look, and if better, and if I should try to do that in crengine :) FreeType matrix and all the math stuff., distorted glyphs caching.. :/

poire-z:
Oh, I see a +1 on `none some more` then ? Let's go with that?

NiLuJe:
> Oh, I see a +1 on `none some more` then ? Let's go with that?

Yep!

> Now, I'm wondering how that would look, and if better, and if I should try to do that in crengine :) FreeType matrix and all the math stuff., distorted glyphs caching.. :/

I have no idea how it's implemented in Latex, but it's probably somewhat expensive to do at runtime...

poire-z:
Some issue with the last crengine commit `Text: dont adjust space after initial quotation mark/dash (rework)`
Before | After:
<kbd>image</kbd>
I'll look at that tomorrow.

mergen3107:
One last thought on words: “none”, “mild”, “harsh” (by harsh I mean aggressive word expansion)

poire-z:
I like `light` or `mild` (but @NiLuJe wrote `light` doesn't quite work with "expansion").
`harsh` is a bit harsh :)

NiLuJe:
If you want to go with light, I'd prefer "slight" instead ;).

poire-z:
As words never go alone, what would be a good buddy for slight ? slarge ?

poire-z:
> As words never go alone, what would be a good buddy for slight ? slarge ?

But now that we're back with toggle labels in logical order, `off | small | large` would work for me :)

Frenzie:
I prefer the some/more I suggested in the sense that I think it feels less permanent.

"Do *some* expansion if necessary" vs "Do *small* expansion… always maybe?"

But that's not really an objection or anything.

poire-z:
> in the sense that I think it feels less permanent

OK, that convinces me :) (that's what I wanted initially, expressing "allowness", so if some/more do that, well, that's good.)

poire-z:
> because there is also "Font Expansion", where, to solve the same issue, it scales the font horizontally.
Now, I'm wondering how that would look, and if better, and if I should try to do that in crengine :) FreeType matrix and all the math stuff., distorted glyphs caching.. :/

I had a quick & dirty attempt at scaling fonts instead of adding letter spacing:
```diff
-    virtual LVFontGlyphCacheItem * getGlyph(lUInt32 ch, lChar16 def_char=0, bool is_fallback=false) {
+    virtual LVFontGlyphCacheItem * getGlyph(lUInt32 ch, lChar16 def_char=0, bool is_fallback=false, int added_letter_spacing=0) {
[...]
         LVFontGlyphCacheItem * item = _glyph_cache.getByChar( ch );
+        if (added_letter_spacing) {
+            item = NULL;
+        }
[...]
-            if (_embolden) { // Don't render yet
+            if (_embolden || added_letter_spacing) { // Don't render yet
                 rend_flags &= ~FT_LOAD_RENDER;
                 // Also disable any hinting, as it would be wrong after embolden
                 rend_flags |= FT_LOAD_NO_AUTOHINT | FT_LOAD_NO_HINTING;
             }

             /* load glyph image into the slot (erase previous one) */
             updateTransform(); // no-op
             int error = FT_Load_Glyph( _face, /* handle to face object */
                     ch_glyph_index,           /* glyph index           */
                     rend_flags );             /* load flags, see below */
             if ( error == FT_Err_Execution_Too_Long && _hintingMode == HINTING_MODE_BYTECODE_INTERPRETOR ) {
                 // Native hinting bytecode may fail with some bad fonts: try again with no hinting
                 rend_flags |= FT_LOAD_NO_HINTING;
                 error = FT_Load_Glyph( _face, ch_glyph_index, rend_flags );
             }
             if ( error ) {
                 return NULL;  /* ignore errors */
             }
+            int orig_matrix_xx = _matrix.xx;
+            if (added_letter_spacing) {
+                int advance = (lUInt16)(FONT_METRIC_TO_PX( myabs(_slot->metrics.horiAdvance) ));
+                _matrix.xx = _matrix.xx * (advance + added_letter_spacing) / advance;
+                FT_Set_Transform(_face, &_matrix, NULL);
+                int error = FT_Load_Glyph( _face, /* handle to face object */
+                        ch_glyph_index,           /* glyph index           */
+                        rend_flags );             /* load flags, see below */
+            }
             if (_embolden) { // Embolden and render
                 // See setEmbolden() for details
                 FT_GlyphSlot_Embolden(_slot);
                 FT_Render_Glyph(_slot, _drawMonochrome?FT_RENDER_MODE_MONO:FT_RENDER_MODE_LIGHT);
             }
+            else if (added_letter_spacing)
+                FT_Render_Glyph(_slot, _drawMonochrome?FT_RENDER_MODE_MONO:FT_RENDER_MODE_LIGHT);

             item = newItem( &_glyph_cache, (lChar16)ch, _slot ); //, _drawMonochrome
-            if (item)
+            if (item && !added_letter_spacing)
                 _glyph_cache.put( item );
+            if (added_letter_spacing) {
+                _matrix.xx = orig_matrix_xx;
+                FT_Set_Transform(_face, &_matrix, NULL);
+            }
         }
         return item;
     }
```
So, each glyph is scaled by `(its width + letter space) / (its width)` (so, a `i` is scaled more than a `m`, which is bad, but good enough to see what it would give).
For those interested to see on their Kobo how it looks: [libcrengine-Kobo-FontScalingInsteadOfLetterSpacing.zip](https://github.com/koreader/koreader/files/4587836/libcrengine-Kobo-FontScalingInsteadOfLetterSpacing.zip)
It's enabled only for Kerning mode `off` or `fast` - so, use `good` (which still uses letter spacing) for comparing.

"Good" = current, with letter spacing:
None:
<kbd>Reader_2020-05-06_162209</kbd>

Some:
<kbd>Reader_2020-05-06_162216</kbd>

More:
<kbd>Reader_2020-05-06_162229</kbd>


"Off" = using font scaling:
None:
<kbd>Reader_2020-05-06_162250</kbd>

Some:
<kbd>Reader_2020-05-06_162306</kbd>

More:
<kbd>Reader_2020-05-06_162314</kbd>

Not really noticable on these github images, but the expanded fonts might appear a bit bolder.
The expanded lines that previously made the text whiter (because of added blank spacing) now makes the text a bit darker (bolded glyphs).
One quite expanded line (the 7th) in this page:
With letter spacing:
<kbd>Reader_2020-05-06_163459</kbd>
With font expanded:
<kbd>Reader_2020-05-06_163529</kbd>


Anyway, I don't think I'll go on on that.
Quotes from that thesis:
> The experiments have shown that font expansion does help to make interword spacing more even, but it also impairs text appearance. Limits of font expansion depend on many factors and must be chosen very carefully A very rough estimation is that the variation in character width caused by font expansion must be limited to ±2%.

We would go further than 2%.

> In Knuth’s opinions, the results of set 3 and set 6 of the samples are excellent (i.e. expansion
limit 20 for Computer Modern/Type 1 fonts and 40 for Multiple Master fonts;
no selected expansion). Unlike my opinions, he preferred variable interletter spacing
(set 6) to font expansion (set 3).
> > About tiny amounts of letter spacing as an alternative to font expansion:
Contrary to what you say in the paper, I actually prefer set 6 to set 3
(in my first impression at least). Obviously that alternative is also a lot simpler,
because it does not increase the number of fonts and it does not lead
to the complications of discreteness in point (2)26 above.
. . .
> > But based on what I see here, I would not go to the bother of making
slightly expanded/condensed fonts. Also therefore I wouldn’t bother to select
characters. The letter spacing would look worst in lines with fi – ff – fl
ligatures, but I couldn’t see any bad lines in the samples I checked of set 6.

> Zapf said that composing with font expansion should be used for book composition,
where there are at least 8 words within a line. At very narrow columns as in the samples
I prepared, composing with font expansion cannot help much to improve interword
spacing, due to small number of words in a line. On the other hand, Knuth thinks that
the benefit of font expansion is considerable mainly for typesetting narrow columns.

> Unlike margin kerning, where the benefit is clear, composing with font expansion is not
encouraged by many typographers. Many people working with typography implicitly
treat font expansion as a “typographic evil” and do not want to hear anything about
such manipulations with typefaces. The reason of this behaviour probably originated
in the fact that font expansion and interletter spacing is often misused in computer
typesetting.

NiLuJe:
Yeah, I'm also less of a fan ;).

That was interesting though, thanks for the experiment!

ptrm:
This looks great, thanks! I have some softhyphen-related glitches on screenshots, but inaccessible atm, and will post them later.

I like the `small` setting most and I guess spacing between 5 and 0% would please me most, but that would go beyond the mentioned (somewhere? I can't find it now :o) one pixel limit. But is this one pixel akin to the CSS font sizes in pixels, i.e. when on 300dpi screen the font size is actually more pixels, and when on 96dpi screen, less pixels in reality?

poire-z:
> I have some softhyphen-related glitches on screenshots, but inaccessible atm, and will post them later.

Post some sample source too, and try to see if some specific setting you may use triggers them. I don't get any glitch (I think :) with my polish softhyphen sample:
<kbd>image</kbd>
With word spacing small:
<kbd>image</kbd>

> I like the small setting most and I guess spacing between 5 and 0% would please me most, but that would go beyond the mentioned (somewhere? I can't find it now :o) one pixel limit. But is this one pixel akin to the CSS font sizes in pixels, i.e. when on 300dpi screen the font size is actually more pixels, and when on 96dpi screen, less pixels in reality?

Nope, the Zoom/DPI should have no effect: it doesn't affect font size (unless specified with CSS fixed units, like 12pt), and the increment of 1px in letter spacing we try are really screen pixels. We move glyphs by 1px, so that the original FreeType rendered glyph is still aligned to its pixels (othewise, that's sub-pixel positionning, and I haven't yet meet that part of the technology, hoping I won't have to :)

ptrm:
> Nope, the Zoom/DPI should have no effect: it doesn't affect font size (unless specified with CSS fixed units, like 12pt), and the increment of 1px in letter spacing we try are really screen pixels. We move glyphs by 1px, so that the original FreeType rendered glyph is still aligned to its pixels (othewise, that's sub-pixel positionning, and I haven't yet meet that part of the technology, hoping I won't have to :)

I see. I will have a closer look and post screenshots as well. Could it be that anyway that some ceiling/rounding is forcing either 0 or >=2px glyph movement on changing percentage in the lowest range :o ?

poire-z:
I think it should do only floor()ing. At lower expansion % or word spacing %, you should get stuck to getting 0px and see no change.
The idea is that ALL glyphs on a line get either all 0px, all 1px, all 2px... If they don't, it's a bug :)
I'll see on your screenshots what might be happening :)

ptrm:
@poire-z here you go.
 
These ones are about the soft hyphens, font is Noto Serif. A bit of wider spacing seems to appear in place of the hyphens:

Reader_2020-05-06_091821 | Reader_2020-05-06_091830 | Reader_2020-05-06_091844
--|--|--
 Reader_2020-05-06_091931 | Reader_2020-05-06_092011


And these are about >1px spacing:

Reader_2020-05-06_092802-diff | Reader_2020-05-06_092815 | Reader_2020-05-06_092802
--|--|--
Screenshot from 2020-05-06 20-21-35 | Screenshot from 2020-05-06 20-21-07

poire-z:
OK, for the soft-hyphen, I can reproduce it. Can you confirm it happens to you, like me, only in Kerning mode: best ? I guess I erroneously add 1px to the soft hyphens.

For your 2nd issue (> 1px), I'm not certain I get it.
But I think it's ok that you screenshots don't superpose: when I add 1px to letters, you can think a 7-letters word gets 7px larger, a 3-letters word gets 3px larger. Then, the words are re-positionned on the line to ensure similar spacing between them.
In your top left "ny opis", "ny" didn't grow much, "opis" did. But the spacing between them has been reduced, so "op" moves a bit left - but with the 1px added between o p i s, it realigns (by luck) on the original.
Did that make sense to you, and explain that what you noticed might not be an issue?

ptrm:
> OK, for the soft-hyphen, I can reproduce it. Can you confirm it happens to you, like me, only in Kerning mode: best ? I guess I erroneously add 1px to the soft hyphens.

Confirmed.

> For your 2nd issue (> 1px), I'm not certain I get it.
> But I think it's ok that you screenshots don't superpose: when I add 1px to letters, you can think a 7-letters word gets 7px larger, a 3-letters word gets 3px larger. [...] Did that make sense to you, and explain that what you noticed might not be an issue?

Yeaaah, totally, forgot how it should with a pixel per letter :D

poire-z:
OK, here's a [libcre-kobo-fix-softhyphens.zip](https://github.com/koreader/koreader/files/4589089/libcre-kobo-fix-softhyphens.zip) with this soft-hyphens bug fixed, so you can confirm it's better and find other bugs :/

It could get really worse than on your screenshots :)
<kbd>image</kbd>
Fixed:
<kbd>image</kbd>

Just a matter of:
```diff
                     // Whole cluster drawn: add letter spacing
-                    x += letter_spacing;
+                    if ( x != prev_x ) {
+                        // But only if this cluster has some advance
+                        // (e.g. a soft-hyphen makes its own cluster, that
+                        // draws a space glyph, but with no advance)
+                        x += letter_spacing;
+                    }
```
Thanks for having noticed that bug!

PR #6113, opened 2020-05-03
[Plugin] Fix move to archive plugin names

poire-z:
Mismatches in these names prevent this plugin from being disabled'able. https://github.com/koreader/koreader/pull/6101#issuecomment-623131390

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6113)

### DIFF
--- a/plugins/movetoarchive.koplugin/_meta.lua
+++ b/plugins/movetoarchive.koplugin/_meta.lua
@@ -1,6 +1,6 @@
 local _ = require("gettext")
 return {
-    name = "move_to_archive",
+    name = "movetoarchive",
     fullname = _("Move to archive"),
     description = _([[Moves/copies current book to archive folder]]),
 }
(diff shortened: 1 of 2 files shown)


plugins/movetoarchive.koplugin/_meta.lua
```diff
@@ -1,6 +1,6 @@
 local _ = require("gettext")
 return {
-    name = "move_to_archive",
+    name = "movetoarchive",
```
Frenzie, line 3: The before seems preferable to me? But yeah, I'd overlooked the move2archive one.

poire-z: We have no plugin with `_` in the directory name.
Could have been "move2archive", as we have already "send2ebook", but well, not minding.

Frenzie: Oh, that was the problem? My bad, I interpreted it as just the mismatch between the metadata name and the widget name.

PR #6114, opened 2020-05-03
Enable Highlight action on single word selection

### LINKED ISSUE #6111: Expose search icons immediately on the long-press menu
* KOReader version: 2020.04.1
* Device: Kobo H2O

#### Issue
I'm currently reading a foreign book, and I find the names extremely hard to remember. Hence, I'm having to search once every few minutes, to remind myself of who is who. To search, I have to navigate the long-press menu, and then select search. It would be much more convenient to have the search buttons directly exposed on this long-press menu.

#### Current usage
1. Long press a word
2. Touch `Search`
3. Use the arrows

#### Preferred usage
1. Long press a word
2. Immediately use the `Search` arrows

#### Rationale
Currently, there are six icons exposed when long pressing a word. I don't use most of them.
* Double arrows: what do these even do? They always seem greyed out for me.
* Highlight: I _do_ use this occasionally, maybe ~3 times per book.
* Wikipedia: I never use this.
* Follow link: I never use this. Isn't this just identical to (short-) pressing a word?
* Search: In my current book, I'd use this >50 times. In a normal book, I'd probably use it ~6–10 times. Not a massive amount more than any other button, but it's definitely my most-used function.

Adding the search icons directly to the menu would add four icons, but lose one. On my device, there would be plenty of room for that.

After a search was initiated, the pop-up window could then be reduced to the minimal four-icon window, obscuring less of the page. (As an aside, I sometimes find even this small window covers important text. I'm not so sure how easy it is, but IMO a better option would be to crop the main text view, and move the popup to the top or bottom edge.)

poire-z:
Also add "Fulltext search" as an option for highlight action.
See https://github.com/koreader/koreader/issues/6111#issuecomment-623088751. Closes #6111.

@protist, in this [highlight_search_singleword.zip](https://github.com/koreader/koreader/files/4571118/highlight_search_singleword.zip) are the 2 modified files, if you want to confirm this works for you.

<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6114)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -550,6 +550,14 @@ function ReaderHighlight:onShowHighlightMenu()
     UIManager:show(self.highlight_dialog)
 end
 
+function ReaderHighlight:_resetHoldTimer(clear)
+    if clear then
+        self.hold_last_tv = nil
+    else
+        self.hold_last_tv = TimeVal.now()
+    end
+end
+
 function ReaderHighlight:onHold(arg, ges)
     -- disable hold gesture if highlighting is disabled
     if self.view.highlight.disabled then return false end
@@ -600,7 +608,7 @@ function ReaderHighlight:onHold(arg, ges)
         --- @todo only mark word?
         -- Unfortunately, CREngine does not return good coordinates
         -- UIManager:setDirty(self.dialog, "partial", self.selected_word.sbox)
-        self.hold_start_tv = TimeVal.now()
+        self:_resetHoldTimer()
         if word.pos0 then
             -- Remember original highlight start position, so we can show
             -- a marker when back from across-pages text selection, which
@@ -615,11 +623,13 @@ end
 function ReaderHighlight:onHoldPan(_, ges)
     if self.hold_pos == nil then
         logger.dbg("no previous hold position")
+        self:_resetHoldTimer(true)
         return true
     end
     local page_area = self.view:getScreenPageArea(self.hold_pos.page)
     if ges.pos:notIntersectWith(page_area) then
         logger.dbg("not inside page area", ges, page_area)
+        self:_resetHoldTimer()
         return true
     end
 
@@ -653,6 +663,7 @@ function ReaderHighlight:onHoldPan(_, ges)
                                       and self.holdpan_pos.x > 7/8*Screen:getWidth()
         end
         if is_in_prev_page_corner or is_in_next_page_corner then
+            self:_resetHoldTimer()
             if self.was_in_some_corner then
                 -- Do nothing, wait for the user to move his finger out of that corner
                 return true
@@ -742,6 +753,7 @@ function ReaderHighlight:onHoldPan(_, ges)
         -- no modification
         return
     end
+    self:_resetHoldTimer() -- selection updated
     logger.dbg("selected text:", self.selected_text)
     if self.selected_text then
         self.view.highlight.temp[self.hold_pos.page] = self.selected_text.sboxes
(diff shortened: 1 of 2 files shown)


poire-z:
Dunno if we should also add `Dictionary lookup`.
It might work well only on single word selection - although I dunno if dicts can/do have multiwords word definitions and if sdcv works with that well.
But for a user that checks "Enable on single word selection", it can be handy I guess.

Frenzie:
@poire-z Works fine afaict ;-)

Screenshot_20200503_212113

mergen3107:
@poire-z
Isn’t single word dictionary lookup currently set as default? On 2020.04.02 release at least. That’s what I extensively use right now

poire-z:
> Isn’t single word dictionary lookup currently set as default?

Yes it is. But people enabling the new option "Enable (highlight action) on single word selection" to be able to use one of the highlight action wouldn't have it in the set. So my question about including it.

These highlight actions can also be cycled with a gesture, so I guess some people might stick with it, and would cycle depending on the book content, and the need for quick "translate" or "wikipedia", or the good usual of "dictionary".

poire-z:
OK, this is buggy: when enabling "on single word", selecting multiple words only uses the first :)

poire-z:
OK, added "Dictionary", fixed bug.
Should we change the order, in the menu and in the cycling?
I did not, to not mess with current users that may be used to the current order.
<kbd>image</kbd>

NiLuJe:
> Should we change the order, in the menu and in the cycling?

Current order looks fine to me ;).

protist:
Thank you so much for the quick work! This seems to work well for me. I had a few comments regarding the implementation though, back in [#6111](https://github.com/koreader/koreader/issues/6111#issuecomment-623239467).

poire-z:
https://github.com/koreader/koreader/issues/6111#issuecomment-623392959

|What|short hold|long hold at end of selection (3s)|
|----|----|----|
|Single word lookup |dict | button menu|
|Single word lookup with [x] Enable on single word lookup|highlight action|button menu|
|Multiple words selection|highlight action|button menu|

@protist: here's the 3rd version: [highlight_search_singleword3.zip](https://github.com/koreader/koreader/files/4574680/highlight_search_singleword3.zip)

protist:
@poire-z Looks great! This seems to work really well for my device! Thank you!

The last thing is I find the UX quite confusing for the settings. In particular, the `Enable on single word lookup` phrasing, but also the implied actions for the other inputs. I wonder if it would be possible to configure it similar to the gesture manager. That is, have a first tier where you specify the input trigger, i.e. select one of

* `Single word lookup — short hold`
* `Single word lookup — long hold`
* `Multiple words selection — short hold`
* `Multiple words selection — long hold`

Then a second tier where you specify one of the six associated actions, i.e. the popup dialog, highlight, translate, Wikipedia, dictionary, fulltext search.

poire-z:
Thanks for the feedback.

> I wonder if it would be possible to configure it similar to the gesture manager

Could make sense. But as it may have less simple implications to other code, I let that for some (someone else's :) other PR.

protist:
Thanks again @poire-z. That definitely sounds reasonable.

(Just for the record, ideally I think the multiple word selection would be different. e.g. if single-word lookup, short hold, is search, it doesn't make as much sense to force multiple-word selection, short hold, to be search too, but perhaps something like dictionary instead. Hence the finer tuning would be useful.)

poire-z:
(Just for the record too :)
I think the contrary: there's no real reason to have single word vs multiple words work differently - moreover with search: what you want to search for (or wikipedia lookup) might be one word, or two words: no reason for the user to wonder and think about what will happening.
It only makes sense with dictionary lookup, where we assume(d) dictionary have only definitions for single words.

protist:
Ooops, sorry, I mistyped. In my usage, probably 90% of the time I select multiple words it's for a highlight. For me, single words are never highlights, and are either search or dictionary.

PR #6118, opened 2020-05-04
Send2Ebook - memory optimalization

mwoz123:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6118)


Frenzie:
Seems okay, can't check atm. This is the same way we do it elsewhere? :-)

Issue #6117, opened 2020-05-04
Kindle PW4 freeze after exit from KOReader


mergen3107:
* KOReader version: 2020.04.01
* Device: Kindle PW4

#### Issue
A user from 4pda forum reports that PW4 freezes after exiting from KOReader.
Logs files from Kindle and KOReader are attached. Hang is visible AFAIK from line 18895 in `all_system_logs` (I just compared the timestamps, there are 17 minutes different)

#### Steps to reproduce
Read a book (allegedly from latest logs, a djvu book);
Exit KOReader;
Kindle freezes.

##### `crash.log` (if applicable)

[4pda KOReader logs.zip](https://github.com/koreader/koreader/files/4575880/4pda.KOReader.logs.zip)

NiLuJe:
An actual idea of which one's supposed to be a hang (i.e., a rough timestamp) would be helpful ;).

Because the final entry is perfectly sane, since there's a ;dm dump done right after.

NiLuJe:
Hmm, the nest to last startup is suspicious, though. It appears to have been using an older, buggy startup script.

My best guess would be the vfat proxy fucked up during an OTA.

A reboot should clear things up.

mergen3107:
One guys says that this problem is definitely thr case on Kindle FW 5.12.5. I understand that it’s not your concern to support all FWs, but probably we should add a warning on KOReader installation instructions for Kindle to stay away from 5.12.x if possible?

NiLuJe:
There's a giant warning about that in the JB thread already, which is linked to in the KOReader instructions :/.

I'm getting a wee bit miffed about people getting burned when clearly playing with fire they've been warned about...

NiLuJe:
I mean, the last guy I had to repeat that to answered something along the lines of "cool. I've updated anyway". `>_<"`.

mergen3107:
@NiLuJe 
Ah, ok! Thank you :) 
Yeah, it is really annoying sometimes

NiLuJe:
That said, in this specific instance, the log clearly shows the startup script not running properly in one run and then properly in the other.

That's typical of the fuse proxy delaying an update because of a held lock. It's been an issue since basically forever on Kindle. The startup script takes some precautions to try to avoid that, but, apparently, shit still happens.

NiLuJe:
(Incidentally, that's why the post-OTA on-screen message essentially reads "reboot if shit happens" on Kindle ^^).

NiLuJe:
[...]

PR #6120, opened 2020-05-04
Tweak the post-OTA warning on Kindle

NiLuJe:
Because everything is terrible, and the fuse proxy can apparently still
manage to screw us over.

(Re: #6117)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6120)


NiLuJe:
As usual on Kindle, if shit gets wonky on an OTA, reboot, and eventually forget about OTAs and keep to manual updates and/or assisted ones via KUAL.

Issue #6119, opened 2020-05-04
Wallabag removes not completed/not read, previously downloaded files.


mwoz123:
* KOReader version: git-master
* Device: android

#### Issue

By default wallabag downloads last 30 articles.

When: 
- `Synchronize remotely deleted files` is checked
- and on koreader you already have 30 article (**not read**) in you wallabag folder, 
- and on server you have 50 new ( 30 previously downloaded + 20 new), 

on after next sync in Wallabag Koreader dir you will have 30 articles (20 new and 10 already downloaded) and Wallabag plugin will remove 20 previously downloaded not read articles from koreader device.

This seems to be caused by this lines:
https://github.com/koreader/koreader/blob/master/plugins/wallabag.koplugin/main.lua#L637-L639
`remote_article_ids` is the list of article ids from current download

In this case not being on the current download list doesn't mean it was deleted from server..

#### Steps to reproduce
Add 30 articles to wallabag server, "download" (sync) them from koreader. Don't read them. 
Add 20 more articles to wallabag server. "download" sync them with Koreader. 20 previously downloaded, not read are gone.

arooni:
Not really the same issue; but is there any way to add articles to wallabag if you're not actively connected to wifi?  Queue them up for later?

Frenzie:
No, that's still waiting for someone to implement it.

mwoz123:
> I've noticed this issue myself.

Workaround that worked for me was to increate download limit (in "settings" -> "configure Wallagag client" ) from default 30 to 100 (I've ~95 articles). The first sync was very long ~5min but next time was much quicker. 

> PS: Not really the same issue; but is there any way to add articles to wallabag if you're not actively connected to wifi? Queue them up for later?

Great idea. If it's not already requestested I think we should extract it to new FR;)

Frenzie:
It's not in any issue, but I explicitly stated it as the obvious next step when I added "add to wallabag." ;-)

arooni:
> > I've noticed this issue myself.
> 
> Workaround that worked for me was to increate download limit (in "settings" -> "configure Wallagag client" ) from default 30 to 100 (I've ~95 articles). The first sync was very long ~5min but next time was much quicker.
> 

What if I already have hundreds of articles in wallabag ? Wouldn’t I need to set the number on the client to 1000 to make this work for me ?  Not sure I want to overload my poor kindle

Frenzie:
[...]

PR #6121, opened 2020-05-05
[wip] Remove article from Wallabag server right after download

mwoz123:
Before I found bug #6119 I did code such change:
To remove just downloaded files from wallabag server, right after download. 

Please check if it might be useful and someone might want to see such feature in Koreader plugin.

Due to the bug #6119 it was not tested because I need to first reoganize my files.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6121)

### DIFF
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -83,7 +83,7 @@ function Wallabag:init()
     if self.wb_settings.data.wallabag.articles_per_sync ~= nil then
         self.articles_per_sync = self.wb_settings.data.wallabag.articles_per_sync
     end
-    self.is_remove_downloaded_from_remote = self.wb_settings.data.wallabag.is_remove_downloaded_from_remote or false
+    self.remove_downloaded_from_remote = self.wb_settings.data.wallabag.remove_downloaded_from_remote or false
     self.remove_finished_from_history = self.wb_settings.data.wallabag.remove_finished_from_history or false
 
     -- workaround for dateparser only available if newsdownloader is active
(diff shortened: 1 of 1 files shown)


plugins/wallabag.koplugin/main.lua
```diff
@@ -83,6 +83,7 @@ function Wallabag:init()
     if self.wb_settings.data.wallabag.articles_per_sync ~= nil then
         self.articles_per_sync = self.wb_settings.data.wallabag.articles_per_sync
     end
+    self.is_remove_downloaded_from_remote = self.wb_settings.data.wallabag.is_remove_downloaded_from_remote or false
```
Frenzie, line 86:
```suggestion
    self.remove_downloaded_from_remote = self.wb_settings.data.wallabag.remove_downloaded_from_remote or false
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -241,6 +242,14 @@ function Wallabag:addToMainMenu(menu_items)
                             self:saveSettings()
                         end,
                     },
+                    {
+                        text = _("Remove downloaded from remote"),
+                        checked_func = function() return self.is_remove_downloaded_from_remote end,
```
Frenzie, line 247: Please write it on multiple lines, that's easier to deal with in the future (+ better diffs).

mwoz123:
>Shouldn't there be some logic to prevent disaster in combination with settings like is_auto_delete? 

Can you exprain? you mean that `is_auto_delete` checked + `remove_downloaded_from_remote` checked can be dagerous? as far as I see `is_auto_delete` removes finished, read or cancelled..

>may be mistaken, but I think there are currently no secretly disastrous combinations.

I'm confused here. Did you put accidentally "no" in that sentence?

Frenzie:
> Can you exprain? you mean that is_auto_delete checked + remove_downloaded_from_remote checked can be dagerous? as far as I see is_auto_delete removes finished, read or cancelled..

Emphasis on _like_. Maybe it's that setting, maybe it's not. The exact specifics aren't really the point of questions like that. It's intended as a trigger, whether for an aha-moment or just to communicate the until now unspoken results of a thought process.

To put it in more concrete terms, normally server deletion equals local deletion upon synchronization. That would seem to result in the following problem:

1. "remove just downloaded files from wallabag server, right after download"
2. Synchronize
3. Just downloaded files are now removed locally because they were removed from the server

As far as I know no such potential issue currently exists.

> I'm confused here. Did you put accidentally "no" in that sentence?

This PR is new, not current. ^_^

mwoz123:
I probably won't have time soon to continue work on this.
Guys feel free to overtake it.

PR #6128, opened 2020-05-06
Delegate "lastfile" management to ReadHistory

### LINKED ISSUE #6126: Entry cleaned from history but stays in "Previous"/" Open last"
* KOReader version: git nightly
* Device: android

#### Issue
Cleaning last entry from history is not with sync with "open last". Still can be reached from that feature. 
"Open last" should be with sync with history. 

#### Steps to reproduce
Open book , remove it from history, use "Open last"/"previous" to oepn it again.

poire-z:
Simplify (and avoid edge cases) in other code by having ReadHistory manage the "lastfile" setting on add, remove, rename...
Fixed a few other cases of things not updated.
See https://github.com/koreader/koreader/issues/6126#issuecomment-624263968. Closes #6126.

Wallabag & MoveToArchive plugins not tested.
I have tested the rest, but I may have missed some (edge or not) cases, so...

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6128)


plugins/movetoarchive.koplugin/main.lua
```diff
@@ -106,12 +106,8 @@ function MoveToArchive:commonProcess(is_move_process, moved_done_text)
         FileManager:copyRecursive(DocSettings:getSidecarDir(document_full_path), self.archive_dir_path)
     end
     local dest_file = string.format("%s%s", self.archive_dir_path, filename)
-    ReadHistory:updateItemByPath(document_full_path, dest_file)
+    ReadHistory:updateItemByPath(document_full_path, dest_file) -- (will update "lastfile" if needed)
     ReadCollection:updateItemByPath(document_full_path, dest_file)
```
mwoz123, line 110: `ReadHistory:updateItemByPath()` updates history and what does 
`ReadCollection:updateItemByPath()` ? 
I glanced at file but didn't understand what's it's doing. 

Isn't it something that should be also done in `ReadHistory`? If we also have to do this toughter?

frontend/apps/reader/modules/readerstatus.lua
```diff
@@ -198,11 +198,9 @@ function ReaderStatus:deleteFile(file, text_end_book)
         ok_text = _("Delete"),
         ok_callback = function()
             local FileManager = require("apps/filemanager/filemanager")
-            local filemanagerutil = require("apps/filemanager/filemanagerutil")
             self.ui:onClose()
             FileManager:deleteFile(file)
-            filemanagerutil.removeFileFromHistoryIfWanted(file)
-            filemanagerutil.ensureLastFileExists()
+            require("readhistory"):fileDeleted(file) -- (will update "lastfile")
```
mwoz123, line 203: is it `isFileDeleted()` check ? or `deleteFile()` ? It's not quite sure what it does before reading full implementation.

How about replacing the comments (as "Uncle Bob" recommends in his book `Clean code`) with self explanatory method name? so here for you name whould be something like:
`fileDeletedAndUpdateLastFile()` ?

So we get rid of comments that after some time are usually orphaned and only mislead devs..

Frenzie: That it also happens to update `lastfile` is an irrelevant internal detail.

plugins/movetoarchive.koplugin/main.lua
poire-z: ReadCollection is another bit of code that manages what's called "Favorites" in the UI.
ReadHistory doesn't know about it - and should not.
So, unless we add a wrapper module `HouseKeeper` and a `HouseKeeper:updateItemByPath(old, new)` that would call this on all the relevant modules, we'll do with it :)
(Or events, but not all modules might be registered to UIManager and get that event.)

frontend/apps/reader/modules/readerstatus.lua
poire-z: > is it isFileDeleted() check ? or deleteFile() ? It's not quite sure what it does before reading full implementation.

None of that. It's just a notification that `thisFileHasJustBeenDeleted(file)`. Not a question, not a request for the action to delete it.
I thought about calling it `onFileDeleted()` - but it might invite people to think this might work as an UI event - and I'm not sure it does/will in any/all conditions.

> How about replacing the comments (as "Uncle Bob" recommends in his book Clean code) with self explanatory method name? so here for you name whould be something like:
fileDeletedAndUpdateLastFile() ?

Well, the idea is that you shouldn't care much about what ReadHistory might do (now and in the future). So you just tell it that this file has been fileDeleted() - and let it do what it should do.
(and yes, the curious dev will have to look at the implementation to know what's done or not - but I expect any serious dev to be that curious :)

Frenzie: > I thought about calling it onFileDeleted() - but it might invite people to think this might work as an UI event - and I'm not sure it does/will in any/all conditions.

You could consider something similar in meaning like process/notify that doesn't have the event handling baggage.

poire-z: Any suggestion? (I thought the `'ed` suffix was enough to convey that meaning).

Frenzie: Well, I'd have said something if I didn't think it was fine. But `notifyFileDeleted` would make it slightly more explicit.

Frenzie: Actually scratch that, I like that much less. Please just keep it. ;-)

poire-z:
> Wallabag & MoveToArchive plugins not tested.

@mwoz123 : can you check that with tomorrow's nightly, all is working well with these plugins? And have some quick checks it works well too in normal reader usage and it actually fixes your issue. thanks :)

mwoz123:
sure

mwoz123:
move to archive is ok. wallabag not yet tested

mwoz123:
@poire-z Actually there seems to be an issue.
When you're in eg. News downloader dir, wallabag or anywhere else where book is removed from history on quit, if you quit book to file manager you are in different directory. Can you fix it here or should I extract it to new issue?

poire-z:
You can keep it here.
But can you describe what happens and what you expect in more details? ("_you are in different directory_" -  in which are you and what you expected).

Might be (or not) related to the `lastdir` setting, that the ReadHistory does not manage (after and before this PR).
https://github.com/koreader/koreader/blob/bee5cd10982721bac3877eeac8964ea255e6be40/frontend/apps/filemanager/filemanager.lua#L1026-L1028
https://github.com/koreader/koreader/blob/bee5cd10982721bac3877eeac8964ea255e6be40/frontend/apps/reader/readerui.lua#L431-L440

```
frontend/apps/reader/modules/readerstatus.lua:                FileManager:showFiles()
frontend/apps/reader/readerui.lua:        FileManager:showFiles(last_dir, last_file)
frontend/apps/filemanager/filemanagerfilesearcher.lua:                        FileManager:showFiles(f.path)
frontend/apps/filemanager/filemanager.lua:function FileManager:showFiles(path, focused_file)
frontend/apps/filemanager/filemanagershortcuts.lua:                            FileManager:showFiles(folder)
frontend/apps/cloudstorage/dropboxapi.lua:function DropBoxApi:showFiles(path, token)
frontend/apps/cloudstorage/cloudstorage.lua:    local remote_files = DropBox:showFiles(item.sync_source_folder, item.password)
frontend/apps/cloudstorage/dropbox.lua:function DropBox:showFiles(url, password)
frontend/apps/cloudstorage/dropbox.lua:    return DropBoxApi:showFiles(url, password)
frontend/device/android/device.lua:                                FileManager:showFiles(content_path)
plugins/newsdownloader.koplugin/main.lua:                        FileManager:showFiles(news_download_dir_path)
plugins/movetoarchive.koplugin/main.lua:        FileManager:showFiles(path)
plugins/wallabag.koplugin/main.lua:                        FileManager:showFiles(self.directory)
plugins/send2ebook.koplugin/main.lua:                        FileManager:showFiles(download_dir_path)
```

May be you can try to see what lastdir and lastfile are and how they change (or should change) in your test case.

mwoz123:
- I open file from dir 'xyz' (which is not removed from history) it stays in history
- next I open file manager and navigate to e.g. News downloader dir (which files are removed from history) 
- I open a file from news downloader
- I close the file by "quitting" to file manager,
- current file is removed from history. 
- file manager is opened and shows dir "xyz" (instead last used news downloader dir)

mwoz123:
For me it looks as if the "lastdir" settings was removed (or rather replaced with previous entry).

poire-z:
> I close the file by "quitting" to file manager,

OK, so that gets `ReaderUI:getLastDirFile()` (code above) triggered, which make up the directory to open File Browser in from `G_reader_settings:readSetting("lastfile")` - so from the first file in xyz/, as it was reset to that when newsdownloader did: _current file is removed from history_.

I don't know how and if that should be fixed :/

mwoz123:
Well, from what your wrote, if there isn't a better (single, global) place, this could be solved by setting it (lastdir) manually in every "onClose()" that removes from history book. We should have ~3-4 such places

poire-z:
No, `lastdir` should be correctly set to the newsloader dir in your case.
The thing is that FileBrowser, invoked from Reader (which you do when you hit the top file browser icon in the Reader menu), gets its dir from `ReaderUI:getLastDirFile()` - so using `lastfile` and not `lastdir`.
The question is why it uses `lastfile` and not just let FileBrowser uses its `lastdir`: probably because you want to browser in the dir the file your current book is (that's ReaderMenu calling `self.ui:showFileManager()` which uses `ReaderUI:getLastDirFile()`.
While End of document action sets to Delete, ReaderStatus uses `FileManager:showFiles()`.
So, the thing is that changing from `ReaderUI:getLastDirFile()` might need a lot of testing, and there are probably other cases where we'll find the current behaviour works better (using current book dir that previous filebrowser lastdir, e.g when you switch books from History without ever going thru File Browser to have it updated its `lastfile`).

May be just swapping these 2 lines:
https://github.com/koreader/koreader/blob/bee5cd10982721bac3877eeac8964ea255e6be40/frontend/apps/reader/modules/readermenu.lua#L48-L49
so filebrowser is called before newsdownloader has updated lastfile can work (but it's logically ugly).

poire-z:
Or, to limit changes and risks, in `NewsDownloader:onCloseDocument()` (and other rare similar places)
```diff
        ReadHistory:removeItemByPath(document_full_path)
+       self.ui.prefered_last_dir = news_download_dir_path
```
and use that `self.prefered_last_dir` in `ReaderUI:getLastDirFile()` if it is set.

mwoz123:
How about, as this remove from history on close is non standard (I don't want to use word "hack";) simply set there "last dir file" here? Without creating new variable and condition to maintain and increasing complexity?

poire-z:
Except that "there", you don't know if you're going to FileBrowser (in which case, your hack will do what you want), or if you're just exiting KOReader (in which case, if you have KOReader "start with last book", your hack will cause next launch to try to open a non-existant file, and you'll get the warning "file not found, do you want to retry?".
Given than when your `NewsDownloader:onCloseDocument() ` is called, `self.ui` is doomed to have no future and possibly just have `ReaderUI:getLastDirFile()` called, may be we could do something like this (untested):

```diff
+function ReaderUI:setLastDirForFileBrowser(dir)
+    self.last_dir_for_file_browser = dir
+end

 function ReaderUI:getLastDirFile(to_file_browser) 
+    if to_file_browser and self.last_dir_for_file_browser then
+        return self.last_dir_for_file_browser
+    end
     local QuickStart = require("ui/quickstart") 
     local last_dir 
     local last_file = G_reader_settings:readSetting("lastfile") 
     -- ignore quickstart guide as last_file so we can go back to home dir 
     if last_file and last_file ~= QuickStart.quickstart_filename then 
         last_dir = last_file:match("(.*)/") 
     end 
     return last_dir, last_file 
 end

function ReaderUI:showFileManager(file)
    local FileManager = require("apps/filemanager/filemanager")

    local last_dir, last_file
    if file then
        last_dir, last_file = util.splitFilePathName(file)
        last_dir = last_dir:match("(.*)/")
    else
-       last_dir, last_file = self:getLastDirFile()
+       last_dir, last_file = self:getLastDirFile(true)
    end
    if FileManager.instance then
        FileManager.instance:reinit(last_dir, last_file)
    else
        FileManager:showFiles(last_dir, last_file)
    end
end

```
And use the non-standard but pretty hack we made :
```diff
        ReadHistory:removeItemByPath(document_full_path)
+       self.ui.setLastDirForFileBrowser(news_download_dir_path)
```

mwoz123:
Ok. You're right. Let's do it your way then:)

Fyi I don't think we remove the file phisicly from disk anywhere in onCloseDocument() only the entry in history so you can skip this case;)

poire-z:
@Frenzie: are you ok with the suggested above little trick?
If yes, @mwoz123 : feel free to test it and push a PR.

Frenzie:
@poire-z Seems ok.

PR #6131, opened 2020-05-07
Reorganize bottom menu config panels

### LINKED ISSUE #6122: Reorganize PDF bottom menu config panels
Not often reading PDF, but giving some thoughts about / asking if the bottom menu could may be be reorganized.

For CRE documents, we moved some stuff in the Font size panel (#6112)
<kbd>image</kbd>

--

It looks to me we could do the same (and more) for PDF, and make the rightmost one smaller:
<kbd>image</kbd>

I guess `Reflow` (which actually is the one that enables the font size widgets - correct me if I'm wrong) could be moved in there, with `Word Gap` (for consistency with the CRE changes):
<kbd>image</kbd>

And may be `Dewatermark` (which I use to remove grayness) (and `Render Quality`?) could be moved to this one:
<kbd>image</kbd>

Dunno if one or two from this one could find a place in one of the others:
<kbd>image</kbd>

Thoughts?

poire-z:
Mostly for the PDF bottom menu. See #6122. Closes #6122.
- Reorganize by topic, trying to limit the nb of widgets   per panel to 4. Screenshots and reasoning in https://github.com/koreader/koreader/issues/6122#issuecomment-624668715
- Re-order some toggles from low to high ('off' then 'on').
- Show font size as number instead of a list of "Aa". Screenshots in https://github.com/koreader/koreader/issues/6122#issuecomment-624564895.
- PDF: add more font size values, and increase usable contrast values.
- Adds help_text to most PDF toggle titles (as I understand the feature, please correct me and my english when I'm wrong :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6131)


frontend/ui/data/koptoptions.lua
```diff
@@ -129,6 +149,8 @@ local KoptOptions = {
                 end,
                 labels = {S.AUTO, S.LEFT, S.CENTER, S.RIGHT, S.JUSTIFY},
                 name_text_hold_callback = optionsutil.showValues,
+                help_text = _([[In reflow mode, sets the lines alignment.
```
Frenzie, line 152: I think you meant "lines' alignment" but I'm not sure if that's any clearer than just alignment?

frontend/ui/data/koptoptions.lua
```diff
@@ -268,25 +285,45 @@ This can also be used to remove some gray background or to convert a grayscale o
...
                 name = "forced_ocr",
                 name_text = S.FORCED_OCR,
-                toggle = {S.ON, S.OFF},
-                values = {1, 0},
+                toggle = {S.OFF, S.ON},
+                values = {0, 1},
                 default_value = 0,
                 advanced = true,
                 name_text_hold_callback = optionsutil.showValues,
+                help_text = _([[Force the use of OCR for text selection, even if the document has some text layer.]]),
```
Frenzie, line 314: You might be talking about something technical with multiple text layers or whatnot? But I think that just unnecessarily complicates things.
```suggestion
                help_text = _([[Force the use of OCR for text selection, even if the document has a text layer.]]),
```

frontend/ui/data/strings.lua
```diff
@@ -7,16 +7,16 @@ S.DUAL_PAGES = _("Dual Pages")
 S.PAGE_CROP = _("Page Crop")
 S.FULL_SCREEN = _("Full Screen")
 S.ZOOM_DPI = _("Zoom (dpi)")
-S.PAGE_MARGIN = _("Margin")
+S.PAGE_MARGIN = _("Margins")
```
Frenzie, line 10: This change looks a bit suspicious, what's it about?

frontend/ui/data/koptoptions.lua
```diff
@@ -30,15 +46,35 @@ local KoptOptions = {
...
-                values = {0, 1, 2, 3},
+                values = {3, 1, 2, 0},
                 default_value = DKOPTREADER_CONFIG_TRIM_PAGE,
                 enabled_func = Device.isTouchDevice,
                 event = "PageCrop",
-                args = {"manual", "auto", "semi-auto", "none"},
+                args = {"none", "auto", "semi-auto", "manual"},
                 name_text_hold_callback = optionsutil.showValues,
-            }
+                help_text = _([[Allows cutting out blank page margins in the original document.
```
NiLuJe, line 59: I'd keep using crop here, it should be common enough (i.e., cropping or croppping out instead of cutting) ;).

poire-z: Initially went with "crop out", but Google tells me it's only used in "(of rock) appear or be exposed at the surface of the earth." :)
Thought cutting out might be an alternative explanation for people not used to "crop" (is that a technical word - or casual english word?)

frontend/ui/data/strings.lua
poire-z: Asked you the question in https://github.com/koreader/koreader/issues/6122#issuecomment-624668715:
> Now includes Margin (should it be plural?) 

<kbd>image</kbd>
It is plural in cre:
<kbd>image</kbd>

frontend/ui/data/koptoptions.lua
poire-z: `text alignment` then ? (avoid wondering if we align words (on a line) or the line itself)

frontend/ui/data/koptoptions.lua
NiLuJe: That's an outcrop, yeah ;).

Cropping should be instantly recognizable by anyone who's ever done even the tiniest bit of image manipulation or photography.

Cropping allows you to crop something out.

And crop can be synonymous to cut in a number of contexts ;).

frontend/ui/data/strings.lua
Frenzie: I'd say that it shouldn't be.

poire-z: Why?

NiLuJe: IIRC, margin is the four all at once (e.g., page margin), or a single one. Two/three of them (e.g., left & right margins) are plural.

Frenzie: Because it's just "the margin." As @NiLuJe wrote.

Frenzie: "The margin of a written page is the empty space on the side."

(There's empty space on all four sides.)

poire-z: Allright, english has the right to be strange.
So, even when talking about _its_ size, and not only as a general concept? _The page margin is small_ ?

NiLuJe: Yep!

frontend/ui/data/creoptions.lua
```diff
@@ -254,7 +254,7 @@ Note that your selected font size is not affected by this setting.]]),
                 more_options = true,
                 more_options_param = {
                   value_min = 70,
```
NiLuJe, line 256: I wouldn't mind lowering this to 50, too ;p.

NiLuJe: (I have a vague memory that Noto has insanely large line-gaps by default, and 70 was still pretty roomy. I may be misremembering :D).

NiLuJe: Okay, scratch that, I had tried Noto, and 70 is when ascenders and descenders start to meet, so, eh, fine by me at 70 ;).

Sorry for the noise.

poire-z: Done.

NiLuJe: (Can't hurt anyway, thanks!)

poire-z: Oh, github comments lags.
Well, still keeping the 50, might be useful with some crappy large line height fonts.

poire-z:
Can I merge this this evening (and give some work to translators) - or should I wait a bit more?

Frenzie:
@poire-z Let's get translation started ASAP, release shouldn't be too far off besides me not having written any release notes. :-)

poire-z:
OK, and easier to spot typo or wording issues on the device than on GitHub.

hius07:
Can I reduce the font size of the info tips?
I would prefer the same size as the bottom menu items are.

poire-z:
Not without tweaking values in the code.
(It uses the same default font & size that most InfoMessages get - and InfoMessage will aditionally reduce it if the text would overflow the screen.)
But why? It's not something you would spent hours contemplating :)

poire-z:
Some unrelated question: should we normalize a bit more our top menus, and generalize the use of `text_func` for menu items with a list of selectable stuff as their submenu - or setable via some widget, to show the thing selected or set?

We have a few of them, with the thing shown in parenthesis or after a colon (should these be normalized too?)
```
Typography rules: French >
Font size (20)
Installed dictionaries (3/5) >
Back to exit (always) >
Previous: That so-so book.epub
```

Should we show:
```
Font: Noto Sans Devanagari UI >
Language: French >
Style: epub.css >
Highlight action: Translate >
Wikipedia languages: en fr ru
Screen DPI: Medium (160) >
```
Or that would make the menu more crowded and less readable - and it's desirable to not abuse that?

Frenzie:
As a rule of thumb I'd say shorter is indeed better.

PR #6134, opened 2020-05-09
Normalize SpinWidgets, add fine tuning to Word Spacing

poire-z:
`Font test sample: tweak footnote links`
Make footnotes links (to non existant footnotes) external links so they don't trigger a "link is invalid" popup and we can use swipe left/right to navigate pages (and jump to latest bookmark/go back for comparing).

`[UX] Normalize SpinWidget and DoubleSpinWidget`
- Use same names for options;
- Have them both movable
- Add option to keep widget shown on Apply (and use that when appropriate: when the value may have an immediate visual effect, so one can tweak it without having to go thru menus to re-open it again).

`[UX] Add fine tune for Word Spacing`
Using DoubleSpinWidget.
Also keep SpinWidgets open on Apply.
Added little trick: if widget has been moved, close bottom config menu on next Apply to show more of the book.

Addded mostly for consistency with the other toggles in that panel (and because some people might want to scale their spaces by 300%, which we allow :) - but it gives a little more room for `small|medium|large` translations :/.

<kbd>image</kbd>

Regarding wording:
Renamed `Set default` (that I find ambiguous, does it restores the picker to the default, or pick the current value and set it as the future default) to `Set as default`

I felt like for grammar, we should add a **by** in here, but thought it wasn't worth screwing already made translations by changing it:
_Tells the rendering engine how much to scale the width of each 'space' character in the text from its regular width, and **by** how much it can additionally reduce them to make more words fit on a line (100% means no reduction)._

I wondered for these toggles with fine tuning if we should not "highlight" the fine tuning toggle when the value is not one of the predefined ones (on its left), instead of highlighting the one with the nearest value:
<kbd>image</kbd>
Which would show as this when I set DPI to be 97:
<kbd>image</kbd>
But decided not to. Thoughts?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6134)


poire-z:
OK, added 2 `by`. Is that gramatically ok?

Any opinion on the last point about which toggle to highlight?

hius07:
> Any opinion on the last point about which toggle to highlight?

To highlight two nearest values, lower and upper.

Frenzie:
@poire-z Yes, that's fine. :)

Wrt the toggle I'd say just round it? (i.e., 96)

mergen3107:
@poire-z
How about `Save as default`? IMO this makes it really clear that it is going to store this setting. One UI comment I have though: for now, the bottom button (Set as default) looks like as if it is the main one, i.e. like “OK” to close this window. So it by design looks a little confusing, because Apply (the actual OK) is above. Shouldn’t we change their order?

Regarding highlighted ellipses, looks fine and intuitive. Just in case if the user set lets says 97 as default, it will tell right away that it is not one of predefined values. Another suggestion might be to include the exact number in parentheses after the ellipses? Don’t know how this will work for 3-digit values though.

Frenzie:
> One UI comment I have though: for now, the bottom button (Set as default) looks like as if it is the main one, i.e. like “OK” to close this window. So it by design looks a little confusing, because Apply (the actual OK) is above. Shouldn’t we change their order?

Why do you say that's not the main one? :-) I tend to think of global as the main thing.

mergen3107:
Well I think about this window as a testing bed. Like, you try 3-4 different things via Apply and only then you save it as default. 

Plus, in different languages there might be different suitable defaults ... :D

poire-z:
> Wrt the toggle I'd say just round it? (i.e., 96)

OK, that's the current behaviour. So, letting it as it is.

> To highlight two nearest values, lower and upper.

Dunno how pleasing that would look, and what about when the setting is below the lower, or above the upper?

> Regarding highlighted ellipses, looks fine and intuitive. Just in case if the user sets let’s say 97 as default, it will tell right away that it is not one of predefined values.

That was my thought. Anyway, might be a little more complicated to implement correctly that what I initially thought - so, may be keeping that for a later PR when I'm bored :)

> Another suggestion might be to include the exact number in parentheses after the ellipses?

Could be nice indeed. Quick dirty little hack to see how that could look (last one not highlighted to see how that would do if we'd just set the value as text when selected):

<kbd>image</kbd>


> How about `Save as default`? 

`Set as default` (to me non-native englisher) is as clear as `Save as default`, so, dunno. We mostly use `Set` in all text, rarely `Save` I think (dunno really why, but I wouldn't really like `Save as default` there :)

> for now, the bottom button (Set as default) looks like as if it is the main one, i.e. like “OK” to close this window. So it by design looks a little confusing, because Apply (the actual OK) is above. Shouldn’t we change their order?

Dunno. For me, I guess I read the buttons from top to bottom, so `Apply` looks like it is the main one. And I'm used (may be it's because I have added quite a few of such dialogs) to see the buttons below as added/extra buttons.
Might also be because on my screenshots, I moved the window to top of screen - usually, it's centered, so the buttons are near the bottom, and the temptation to reach the bottomest as the default would be less attractive :)
Also, it's dictated by the length of the text, `Set as default` might be truncated if set to a half-sized button, so it has to be a large one - otherwise, with the logic of _the bigger the most intuitive_, it could also be:
```
Cancel | Save as default
     Apply
```

hius07:
> Dunno how pleasing that would look

OK, anyway the fine tuning toggle should be highlighted.
Now, for example, I have font size 23, highlighted 24, no indication that it is not 24 actually.

"Set as default" seems absolutely clear.

NiLuJe:
Knowing that the actial value is visible on a long-press on the option's name, I'd be okay w/ an HL on the ellipsis.

Having the actual value printed there sounds fun, but it could get hairy real quick depending on the type of values...

hius07:
The last screenshot doesn't look great, sorry.

NiLuJe:
> The last screenshot doesn't look great, sorry.

Yeah, my point exactly ;).

poire-z:
Even if I remove the ugly `100%, 90%` ?

<kbd>image</kbd>

<kbd>image</kbd>

Anyway, I'll merge this later this evening - and keep the highlighting change (if we end up agreeing, and it's not a mess to code) for another PR probably for after next release - so keep discussing it :)

NiLuJe:
Yeah, not a fan of the added values, no matter the format ;). But the HL itself is cool.

poire-z:
You might be saying that because, like me and most, you don't set custom values.
But may be the people who do would be happy seeing their value here (and avoid having to tap or hold to see it) - and we would not, using predefined values.

NiLuJe:
I mean, the idea is fine and all, I think it just looks weird ;).

ptrm:
\o/ no more restarts to fine tune the spacing. I had done that every other day, thanks 😁

poire-z:
Oh, you're not yet over that word spacing disease :)
I mentionned another setting I externalized, that is too complicated to explain in the UI, and I don't really see a common explainable effect.
But you can try playing with it, for science!
It has the same limits (0% <= x <= 20%) as the Word Expansion one, so, if you can temporarily give up Word Expansion and use its toggle and fine tuning to experiement with that other setting:
```diff
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -815,5 +815,6 @@ end
 function CreDocument:setWordExpansion(value)
     logger.dbg("CreDocument: set word expansion", value)
-    self._document:setIntProperty("crengine.style.max.added.letter.spacing.percent", value or 0)
+    -- self._document:setIntProperty("crengine.style.max.added.letter.spacing.percent", value or 0)
+    self._document:setIntProperty("crengine.style.unused.space.threshold.percent", value or 5)
 end

```
Explanation: % (of line width) of unused space on a line to trigger hyphenation, or addition of letter spacing for justification  - default is 5%

poire-z:
Asking some questions in there to get the attention of you who contributed.
(Previous Word spacing work also in #5655.)

I'm working on CJK typography, and I may have in crengine a new setting to scale CJK glyphs:
```
    CJK: allow scaling CJK glyphs' widths

    This may allow "Word Spacing" to have some effect
    on CJK text (where each glyphs is a word).
    Scaling glyphs (or letter spacing) is considered
    bad practice by many typography sites, but it might
    be helpful with very "solid" CJK fonts and to users
    learning Chinese.
```
We have these settings, and I would add the last one in this:
```c
    /// set space glyph width scaling percent option (10..500%)
    // (scale the normal width of all spaces in all fonts by this percent)
    void setSpaceWidthScalePercent(int spaceWidthScalePercent);

    /// set space condensing line fitting option (25..100%)
    // (applies after spaceWidthScalePercent has been applied)
    void setMinSpaceCondensingPercent(int minSpaceCondensingPercent);

    /// set unused space threshold percent option (0..20%)
    void setUnusedSpaceThresholdPercent(int unusedSpaceThresholdPercent);

    /// set max allowed added letter spacing (0..20% of font size)
    void setMaxAddedLetterSpacingPercent(int maxAddedLetterSpacingPercent);

    /// set CJK glyph width scaling percent option (100..150%)
    // (scale the nominal width of all CJK chars in all fonts by this percent)
    void setCJKWidthScalePercent(int cjkWidthScalePercent);
```
Our Word Spacing toggle and fine tune DoubleSpinWidget allow tuning the 2 first ones.

I think that, for CJK readers, the new one could also be considered tweaking "word spacing".

Some samples, all text-align'ed left (using a Sans CJK font, which is dense, and doesn't have all their CJK glyphs the same width, so the east-asian grid being a bit broken - which does not happen with Serif CJK fonts, which have thinner stroke, but would still look too condensed):

With cjkWidthScalePercent=100% (so, natural stacking of glyphs):
image

With cjkWidthScalePercent=105% 
image

With cjkWidthScalePercent=110% 
image

With cjkWidthScalePercent=115% 
image

(More than 120% may make the glyphs too spread out, and I guess confusing for CJK readers as they could be tempted to read it vertically :)

For now, I didn't change anything and based this value from our space_scaled_width_percent + 5%:

```diff
 function CreDocument:setWordSpacing(values)
     -- values should be a table of 2 numbers (e.g.: { 90, 75 })
     -- - space width scale percent (hard scale the width of each space char in
     --   all fonts - 100 to use the normal font space glyph width unchanged).
     -- - min space condensing percent (how much we can additionally decrease
     --   a space width to make text fit on a line).
     logger.dbg("CreDocument: set space width scale", values[1])
     self._document:setIntProperty("crengine.style.space.width.scale.percent", values[1])
     logger.dbg("CreDocument: set space condensing", values[2])
     self._document:setIntProperty("crengine.style.space.condensing.percent", values[2])
+    -- No dedicated UI tuning for CJK width scaling: base it on space.width.scale.percent + 5%:
+    -- 'medium' space scaling being 95% will make cjk scaling 100% (and so will 'small')
+    -- 'large' space scaling being 100% will make CJK scaling 105% which might look ok;
+    -- users who want more for CJK can increase Space scaling over 100%.
+    local cjk_scaling_percent = math.min(math.max(values[1] + 5, 100), 150) -- min/max 100%/150% are ensured by crengine itself, but get the log accurate
+    logger.dbg("CreDocument: set cjk width scale", cjk_scaling_percent)
+    self._document:setIntProperty("crengine.style.cjk.width.scale.percent", cjk_scaling_percent)
 end
```

This solution is a bit bothering, because:
- to get the most natural spacing according to the font, we westerners choose medium (or large). So, to get "medium" to also looks sane for CJK Reader, "medium" has to get cjk_scaling=100%, so east-asianers get also no space added
- but then "small" has no effect - and "large" has a very small effect (105%) - and I think for it to be noticable, we'd need to have 110% and 120% as the values.
- there's no documentation in the widget and the fine tuning DoubleSpinWidget that this is happening and possible (and it would be a little hard to explain "that this also tweak CJK glyphs with this formula - do the maths yourself" :/)

So, hardcoding the linking of cjk_scaling to our space_scaling is a bit tedious: space_scaling goes down from the natural font look - while cjk_scaling should go up from the natural font glyph widths.

So, ideally, we should have 3 values instead of 2 in:
https://github.com/koreader/koreader/blob/6804b77251431c6c5a3a57ea0e17ef25d8260dca/defaults.lua#L179-L191
Something like:
```lua
DCREREADER_CONFIG_WORD_SPACING_SMALL = {75, 50, 100}
DCREREADER_CONFIG_WORD_SPACING_MEDIUM = {95, 75, 110}
DCREREADER_CONFIG_WORD_SPACING_LARGE = {100, 90, 120}
```

But... : 
- we don't have a TripleSpinWidget to allow tuning these independantly (and given that our SpinWidget and DoubleSpinWidget are so often tweaked and tried to be made similar, having to handle a TripleSpinWidget that would be used only for that is a bit bothering...)
- users wanting to have their natural font widths would need to use "medium" or "large" when reading English or other non-CJK text - and they would need to switch to "small" when reading CJK... and would have a decision to make when reading mixed text (although I guess that if they like their CJK condensed, they would be fine with "small" and their ascii spaces 75% instead of 95%).

So, here I am - I don't know how to go with this.
Any thought about this ?

hius07:
We have an extra_button in DSW, to call a SpinWidget for CJK setting.

poire-z:
@weijiuqiao (from #9132: _I am actually working as a client-side developer in the daytime dealing with UI a lot_), if I can assume from your username that you may read Chinese (and not only [Qiang/Rma](https://weijiuqiao.github.io/qiang_script/) :), can I request your thoughts about the above?
Is increasing glyph spacing for CJK a nice thing to have (cf the screenshots) ?
Would it be ok if you read both english and chinese text to have to toggle from small to medium (or stay with small)?
Any bright idea about making this clearer in the UI (that wouldn't need too much rework) ?

weijiuqiao:
>if I can assume from your username that you may read Chinese

Yes you can :) I read Chinese at a native level. 

>Is increasing glyph spacing for CJK a nice thing to have (cf the screenshots) ?

Just not too much :) Regarding the screenshots, only say for myself, `cjkWidthScalePercent=100%` is normal and only slightly narrow but totally OK. I'd said about 3/4 way from small to medium of western texts. `cjkWidthScalePercent=105%` seems to be the maximum before the spaces become noticeable. It still looks natural. At 110% and above the spaces are more noticeable, requires a bit effort to ignore them while reading. Spaced Chinese are like t h i s  i n  E n g l i s h.

It looks like Chinese characters are analogous to western letters, in terms of spacing.

>Would it be ok if you read both english and chinese text to have to toggle from small to medium (or stay with small)?

I always toggle to medium reading English as small is quite crowded. As to Chinese text, I seldom change it.

>Any bright idea about making this clearer in the UI (that wouldn't need too much rework) ?

Haha, how about making 100% small, 102/103 medium and 105 large? ie make the correspondence non-linear? 75 - 100, 95 - 102.5, 100 - 105?

ptrm:
@poire-z just out of curiosity: would adding letter-width modification to latin-ish letters be a drop-in change or is the thing you describe a whole different thing?

poire-z:
@ptrm : no, this would be some really specific code affecting only CJK characters (to actually "center" the glyph in its scaled square - I had some initial try with letter-spacing at https://github.com/koreader/koreader/issues/6162#issuecomment-1120421679 - but it was not the right solution for this CJK scaling).
And no, I'd rather not mess with western fonts and their kerning or Harfbuzz fine positionning and glyph subsitution and ligatures :) (~~but you could, with `* { letter-spacing: -0.1em }` and take the blame if it's messsed up :)~~ no, we don't support negative values for letter-spacing).

@weijiuqiao : thank you.
Well, if 105% is really the maximum that would not feel odd, I can stay with my "for now" implementation above (that is, small = medium = 100%, and only large = 105%, barely noticable, and we would not need to even write about it - and users who want more will discover that they can add more via the fine tuning widget).
I also feel more than 105% might be too much with Japanese, which has other small characters that are already quite spaced comparing to the few Hanji characters, so adding more spacing could feel even more odd:
image 100%, natural
image 110%
(but well, even at 100%, the spacing between hanji chars and the spacing between non-hanji chars is already very different...)

(I thought that for people learning Chinese, a bit more spacing could help, ie. with these 2 glyphs image I might have a hard time guessing where there is a boundary and that they are not 4 glyphs - but what do I know :) may be it's the same with english and noticing `w` is not `vv` :)
@Tang906 , with your https://github.com/koreader/koreader/issues/6162#issuecomment-650469410 , how do you feel about the 100% 105% 110% 115% screenshots above ?

@weijiuqiao : what's your device ? By any chance a Kobo (that I could give a build for) ? (or do you have the emulator and you can quickly apply a patch to crengine to check?) I'm finishing some work on CJK typography, and haven't had much feedback from CJK readers lately in #6162 - a quick validation by a native reader would be nice.
(Any experience with reading horizontal Traditional Chinese and its expectations with consecutive centered punctuations ? which is some topic I haven't found much discussion about.)

weijiuqiao:
@poire-z Regarding your 都叫 example, as Chinese are strictly grid based, especially in printing, I don't think it'll be a problem - but I already have a hardwired intuition of character grids, so what do I know for learners :)

Unfortunately my device is a kindle. I do have the emulator installed and running fine. And I also managed to get the `kodev release kindle` to run. So let me know what to do.

My experiences of reading traditional Chinese are limited to materials published in mainland China, and I just checked one of them and the punctuations are SC style.

poire-z:
Ok, thanks. I'll give you the patch and instructions of what to expect and check in #6162 so the discussions is kept there.

poire-z:
> We have an extra_button in DSW, may be used to call a SpinWidget for CJK setting.

Alas, it's used for the `Set as default` in Word Spacing (don't even want to dig in to understand why, and if it could use the default_values button).
But SSW has an extra extra button with other_text/callback.

So, I might use the SSW of Word Expansion - as for CJK user, Word Spacing or Word Expansion could both work for adjusting spacing between CJK glyphs - so he might naturally go look there :)

But as it's a bit orthogonal to any of these settings, I guess it should be an independant setting.
And as it's not super useful to most CJK readers, and as even us westerners must not play often with Word Expansion, we might as well add it there. Which would give:
image

which would just close this one and open the other one:
image
(or even with a button allowing to go back to the Word Expansion SW):
image

(I'd rather not make this a DSW with both values, as `none | some | more` don't really map to each other - a CJK reader may want to keep normal CJK, and choose "more" here to get the western words expanded which will allow the CJK glyphs to stay near 100%.)

I found a nice trick to get all this working without having to add some new code to manage the setting and to build a SpinWidget - this might be used for other "small"/rarely used setting:

```diff
--- a/frontend/ui/data/creoptions.lua
+++ b/frontend/ui/data/creoptions.lua
@@ -487,6 +487,31 @@ Note that your selected font size is not affected by this setting.]]),
                     return string.format("%d%%, %d%%", val[1], val[2])
                 end,
             },
+            {
+                -- This setting is now shown in the bottom menu, but via the other_button in Word Expansion fine tuning.
+                -- We still need to define it as a setting here for it to be known and handled as the others,
+                -- but we just don't show it.
+                show = false,
+                name = "cjk_width_scaling",
+                default_value = 100,
+                values = {}, --  100, 105, 110 },
+                event = "SetCJKWidthScaling",
+                more_options = true,
+                more_options_param = {
+                    value_min = 100,
+                    value_max = 150,
+                    value_step = 1,
+                    value_hold_step = 5,
+                    name = "cjk_width_scaling",
+                    name_text = _("CJK width scaling"),
+                    info_text = _([[Scale all CJK (Chinese, Japanese, Korean) glyphs' widths by this %. This has the effect of adding spacing between these glyphs, and might make them easier to
 distinguish to some readers.]]),
+                    event = "SetCJKWidthScaling",
+                    other_button = {
+                        text = _("Word expansion"),
+                        other_setting = "word_expansion",
+                    }
+                },
+            },
             {
                 name = "word_expansion",
                 name_text = _("Word Expansion"),
@@ -500,6 +525,10 @@ Note that your selected font size is not affected by this setting.]]),
                     name_text = _("Max word expansion"),
                     info_text = _([[Set max word expansion as a % of the font size.]]),
                     event = "SetWordExpansion",
+                    other_button = {
+                        text = _("CJK scaling"),
+                        other_setting = "cjk_width_scaling",
+                    }
                 },
                 toggle = {C_("Word expansion", "none"), C_("Word expansion", "some"), C_("Word expansion", "more")},
                 values = {
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -1334,6 +1334,27 @@ function ConfigDialog:onConfigMoreChoose(values, name, event, args, name_text, m
                             end,
                         })
                     end,
+                    option_text =  more_options_param.other_button and more_options_param.other_button.text,
+                    option_callback =  more_options_param.other_button and function()
+                        widget:onClose()
+                        local name = more_options_param.other_button.other_setting
+                        -- Find it among all the options
+                        local option
+                        for i=1, #self.config_options do
+                            local options = self.config_options[i].options
+                            for j=1, #options do
+                                if options[j].name == name then
+                                    option = options[j]
+                                    break
+                                end
+                            end
+                            if option then
+                                break
+                            end
+                        end
+                        self:onConfigMoreChoose(option.values, option.name, option.event,
+                                            "â<8B>®", option.name_text, option.more_options_param)
+                    end,
                 }
             end
             UIManager:show(widget)
```

PR #6146, opened 2020-05-13
Add chapter title when exporting notes in Evernote(JSON/HTML, remote)

### LINKED ISSUE #4566: Chapter title not present in the notes exported to evernote
* KOReader version: v2019.02
* Device: Kobo Aura

#### Issue
When exporting notes to Evernote, the notes should include the Chapter title as well. 

It will be a great help to see the chapter names when revising the notes. 

PS: Brilliant feature!! Thanks.

### LINKED ISSUE #6138: Save chapter name when highlighting text
Is it possible to add additional entry to <ebook_name>.sdr/metadata.epub.lua that also saves the chapter where highlight was made? I'm working on a wikia page for one book, and it would make my life a lot easier. Not sure if there are already functions that take xpath for example and translate it into ToC entry. I'd be grateful if someone could point me in the right direction.

Galunid:
closes #4566
closes #6138 

This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6146)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -218,7 +218,7 @@ end
 -- This function adds "chapter" property to highlights already saved in the document
 function ReaderBookmark:updateHighlightsIfNeeded()
     local version = self.ui.doc_settings:readSetting("bookmarks_version") or 0
-    if version >= 20200511 then
+    if version >= 20200615 then
         return
     end
 
@@ -242,7 +242,7 @@ function ReaderBookmark:updateHighlightsIfNeeded()
             bookmark.chapter = chapter_name
         end
     end
-    self.ui.doc_settings:saveSetting("bookmarks_version", 20200511)
+    self.ui.doc_settings:saveSetting("bookmarks_version", 20200615)
 end
 
 function ReaderBookmark:onShowBookmark()
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -234,7 +234,7 @@ function ReaderBookmark:updateHighlightsIfNeeded()
     for _, bookmark in ipairs(self.bookmarks) do
         if bookmark.pos0 then
             local pg_or_xp = self.ui.document.info.has_pages and
-                    bookmark.pos0.page or bookmark.pos0
+                    bookmark.page or bookmark.pos0
                 local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
             bookmark.chapter = chapter_name
         elseif bookmark.page then -- dogear bookmark
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -224,7 +224,9 @@ function ReaderBookmark:updateHighlightsIfNeeded()
 
     for _, highlights in pairs(self.view.highlight.saved) do
         for _, highlight in pairs(highlights) do
-            local chapter_name = self.ui.toc:getTocTitleByPage(highlight.pos0)
+            local pg_or_xp = self.ui.document.info.has_pages and
+                    highlight.pos0.page or highlight.pos0
+            local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
             highlight.chapter = chapter_name
         end
     end
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -222,10 +222,10 @@ function ReaderBookmark:updateHighlightsIfNeeded()
         return
     end
 
-    for _, highlights in pairs(self.view.highlight.saved) do
+    for page, highlights in pairs(self.view.highlight.saved) do
         for _, highlight in pairs(highlights) do
             local pg_or_xp = self.ui.document.info.has_pages and
-                    highlight.pos0.page or highlight.pos0
+                    page or highlight.pos0
             local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
             highlight.chapter = chapter_name
         end
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -226,13 +226,11 @@ end
 
 function ReaderBookmark:onShowBookmark()
     -- check if highlights already have "chapter" property
+    -- this executes only once, just to set has_chapter_property variable
     local has_chapter_property = false
     for _, highlights in pairs(self.view.highlight.saved) do
-        for _, highlight in ipairs(highlights) do
-            if highlight.chapter then
-              has_chapter_property = true
-            end
-            break
+        if highlights[1].chapter then
+            has_chapter_property = true
         end
         break
     end
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -232,10 +232,13 @@ function ReaderBookmark:updateHighlightsIfNeeded()
     end
 
     for _, bookmark in ipairs(self.bookmarks) do
-        if bookmark.pos0 then -- this prevents tests from failing, shouldn't be necessary otherwise
+        if bookmark.pos0 then
             local pg_or_xp = self.ui.document.info.has_pages and
                     bookmark.pos0.page or bookmark.pos0
                 local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
+            bookmark.chapter = chapter_name
+        elseif bookmark.page then -- dogear bookmark
+            local chapter_name = self.ui.toc:getTocTitleByPage(bookmark.page)
             bookmark.chapter = chapter_name
         end
     end
@@ -567,6 +570,7 @@ function ReaderBookmark:toggleBookmark(pn_or_xp)
     else
         -- build notes from TOC
         local notes = self.ui.toc:getTocTitleByPage(pn_or_xp)
+        local chapter_name = notes
         if notes ~= "" then
             notes = "in "..notes
         end
@@ -574,6 +578,7 @@ function ReaderBookmark:toggleBookmark(pn_or_xp)
             page = pn_or_xp,
             datetime = os.date("%Y-%m-%d %H:%M:%S"),
             notes = notes,
+            chapter = chapter_name
         })
     end
 end
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -215,29 +215,22 @@ function ReaderBookmark:gotoBookmark(pn_or_xp)
     end
 end
 
-function ReaderBookmark:regenerateHighlights()
+function ReaderBookmark:updateHighlightsToAddChapterPropertyIfNeeded()
+    if G_reader_settings:has("bookmarks_chapters_updated") then
+        return
+    end
+
     for _, highlights in pairs(self.view.highlight.saved) do
         for _, highlight in pairs(highlights) do
             local chapter_name = self.ui.toc:getTocTitleByPage(highlight.pos0)
             highlight.chapter = chapter_name
         end
     end
+    G_reader_settings:saveSetting("bookmarks_update_version", true)
 end
 
 function ReaderBookmark:onShowBookmark()
-    -- check if highlights already have "chapter" property
-    -- this executes only once, just to set has_chapter_property variable
-    local has_chapter_property = false
-    for _, highlights in pairs(self.view.highlight.saved) do
-        if highlights[1].chapter then
-            has_chapter_property = true
-        end
-        break
-    end
-
-    if not has_chapter_property then
-        self:regenerateHighlights()
-    end
+    self:updateHighlightsToAddChapterPropertyIfNeeded()
     -- build up item_table
     for k, v in ipairs(self.bookmarks) do
         local page = v.page
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -232,10 +232,12 @@ function ReaderBookmark:updateHighlightsIfNeeded()
     end
 
     for _, bookmark in ipairs(self.bookmarks) do
-        local pg_or_xp = self.ui.document.info.has_pages and
-                bookmark.pos0.page or bookmark.pos0
-        local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
-        bookmark.chapter = chapter_name
+        if bookmark.pos0 then -- this prevents tests from failing, shouldn't be necessary otherwise
+            local pg_or_xp = self.ui.document.info.has_pages and
+                    bookmark.pos0.page or bookmark.pos0
+                local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
+            bookmark.chapter = chapter_name
+        end
     end
     self.ui.doc_settings:saveSetting("bookmarks_version", 20200511)
 end
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -215,7 +215,31 @@ function ReaderBookmark:gotoBookmark(pn_or_xp)
     end
 end
 
+function ReaderBookmark:regenerateHighlights()
+    for _, highlights in pairs(self.view.highlight.saved) do
+        for _, highlight in pairs(highlights) do
+            local chapter_name = self.ui.toc:getTocTitleByPage(highlight.pos0)
+            highlight.chapter = chapter_name
+        end
+    end
+end
+
 function ReaderBookmark:onShowBookmark()
+    -- check if highlights already have "chapter" property
+    local has_chapter_property = false
+    for _, highlights in pairs(self.view.highlight.saved) do
+        for _, highlight in ipairs(highlights) do
+            if highlight.chapter then
+              has_chapter_property = true
+            end
+            break
+        end
+        break
+    end
+
+    if not has_chapter_property then
+        self:regenerateHighlights()
+    end
     -- build up item_table
     for k, v in ipairs(self.bookmarks) do
         local page = v.page
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -215,8 +215,10 @@ function ReaderBookmark:gotoBookmark(pn_or_xp)
     end
 end
 
-function ReaderBookmark:updateHighlightsToAddChapterPropertyIfNeeded()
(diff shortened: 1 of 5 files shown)


plugins/evernote.koplugin/clip.lua
```diff
@@ -235,6 +235,7 @@ function MyClipping:parseHighlight(highlights, bookmarks, book)
             clipping.sort = "highlight"
             clipping.time = self:getTime(item.datetime or "")
             clipping.text = self:getText(item.text)
+            clipping.chapter = item.chapter
             for _, bookmark in pairs(bookmarks) do
                 if bookmark.datetime == item.datetime and bookmark.text then
```
poire-z, line 240: (Making this comment here, because it makes it more obvious.)
You set `chapter` on the items in the `highlight` setting - but I get the feeling it makes sense to have it also (or instead, dunno) in `bookmarks`.
Although this could be dealt with by #5794.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -1146,13 +1146,15 @@ function ReaderHighlight:getHighlightBookmarkItem()
         local datetime = os.date("%Y-%m-%d %H:%M:%S")
         local page = self.ui.document.info.has_pages and
                 self.hold_pos.page or self.selected_text.pos0
+        local chapter_name = self.ui.toc:getTocTitleByPage(page)
```
poire-z, line 1149: OK, as I mentionned in https://github.com/koreader/koreader/issues/6138#issuecomment-626563596, this might be innacurate for CRE documents.
You should try to get the TOC entry whose xpointer is the nearest before your highlight/bookmark xpointer, see https://github.com/koreader/koreader/issues/6138#issuecomment-626543969.
May be by adding a new method to readertoc.lua.

`ReaderToc:getTocTitleByPage(pn_or_xp)` looks like it might be it, but it does like you do, which might be good enough for bookmarked page (although I guess it's already innacurate).
https://github.com/koreader/koreader/blob/12675be0659ad3c1d099610490a4322e3c994fa2/frontend/apps/reader/modules/readerbookmark.lua#L539-L548

plugins/evernote.koplugin/clip.lua
Galunid: I set it in highlights, because Evernote plugin operates mostly based on highlights(it uses bookmarks only to get user notes). That storage system is pretty confusing, it's basically duplicating entries for highlights and bookmarks, and duplicating "text" and "notes" in bookmarks. I didn't really see the point of adding this property to bookmarks. It's not a big deal though, I can add it.

Galunid:
Any idea why `readhistory_spec.lua` fails? I run it on my branch and it fails 2/10 times, when I run it on master it fails 1-3/10 times.

plugins/evernote.koplugin/clip.lua
Galunid: Actually now that I double checked, it saves it for bookmarks too in `ReaderHighlight:getHighlightBookmarkItem` which is called by `ReaderHighlight:saveHighlight()`.

poire-z: > That storage system is pretty confusing, it's basically duplicating entries for highlights and bookmarks

Yes, it's a mess. #5794 (currently stalled) is an aim at making that saner.

poire-z:
> Any idea why readhistory_spec.lua fails? I run it on my branch and it fails 2/10 times, when I run it on master it fails 1-3/10 times.

This happens, don't worry, just re-run the workflow (dunno if you can, or closing/re-opening the PR used to trigger a re-run).
I think it happens in the test that try to check some ordering of items inserted, which then can fail if inserted in the same second. I guess no one really looked at fixing it so these failures don't happen.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -1146,13 +1146,15 @@ function ReaderHighlight:getHighlightBookmarkItem()
         local datetime = os.date("%Y-%m-%d %H:%M:%S")
         local page = self.ui.document.info.has_pages and
                 self.hold_pos.page or self.selected_text.pos0
+        local chapter_name = self.ui.toc:getAccurateTocTitleByXPointer(self.selected_text.pos0)
```
poire-z, line 1149: I think you want chapters working too with PDF, that don't use xpointers. For PDF, we can't do better than using the getTocTitleByPage you previously used.
May be hide that business in betterly named `ReaderToc:getTocChapter(pn_or_xp)`

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -215,6 +215,27 @@ function ReaderToc:getTocIndexByPage(pn_or_xp)
...
 
+function ReaderToc:getAccurateTocIndexByXPointer(xptr)
+    local index = self:getTocIndexByPage(xptr)
+    if self.toc[index + 1] then
+        local next_index_xptr = self.toc[index + 1].xpointer
+        local comparison = self.ui.document:compareXPointers(next_index_xptr, xptr)
+        if comparison and comparison > 0 then
+            return index + 1
+        end
+    end
```
poire-z, line 226: I don't think looking at +1 is enough: you can have multiple TOC entries for a single page:

<kbd>image</kbd>

These 3 highlights should each have a different chapter (the first one should have the chapter that is shown on a previous page).
Actually, the dogear bookmark says it's in chapter "Etats impériaux" (getTocTitleByPage), which I guess is allright as nobody complains :) but I think it should be too the chapter that started on previous page.

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -217,11 +217,25 @@ end
...
+                return index - i
+            end
+            i = i + 1
+        end
+    else
+        while self.toc[index + i] do
+            local toc_xptr = self.toc[index + i].xpointer
+            local cmp = self.ui.document:compareXPointers(toc_xptr, xptr)
+            if cmp and cmp <= 0 then
+                return index + i
```
poire-z, line 236: Initial comparison: xptr comes after the toc xpointer - so there (in the `else`), you go look at the ones after.
Here, when `cmp <= 0`, you just found a `toc_xptr` that is after `xptr`, so, not a good one (but it is when `cmp == 0`) - but the one just before was allright. So, shouldn't you return `index + i - 1` ?
Overall, logic seems allright, but these `index + i` make that a bit harder to read than it should :) Can you try just using:
```lua
local i = index
self.toc[i]
return i
i = i + 1
```

Galunid: >So, shouldn't you return index + i - 1 ?
Of course it should! I must have missed it, because `cmp <=  0` should be `cmp < 0`. Actually it probably doesn't matter if `getTocIndexByPage` doesn't change. For now it always returns the last "chapter" on the page, thus making `else` unnecessary.

frontend/apps/reader/modules/readerhighlight.lua
Galunid: I moved `getAccurateTocTitleByXPointer` to `getTocTitleByPage`, it uses the xpointer for accuracy(xpointer is passed), otherwise it uses page, as it used to.

poire-z:
> Actually now that I double checked, it saves it for bookmarks too in ReaderHighlight:getHighlightBookmarkItem which is called by ReaderHighlight:saveHighlight().

Ok, so new ones get added to bookmark with their chapter.
Dunno how complicate it is as it's then no more the same object, but for consitency, would it be possible to also add their chapter to each bookmark in `ReaderBookmark:updateHighlightsIfNeeded()` (where you add it to the "highlight" clone)?
(Don't if it's too complicated - just add a comment that past bookmarks don't get chapters, so that later work on removing the current mess know it can get chapter only in highlight.)

Also, even if it's probably rare we cross chapter when moving some highlight boundaries, chapter could be updated in `ReaderHighlight:updateHighlight()` / `self.ui.bookmark:updateBookmark()`.

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -215,6 +215,34 @@ function ReaderToc:getTocIndexByPage(pn_or_xp)
...
 end
 
+function ReaderToc:getAccurateTocIndexByXPointer(xptr)
+    local pageno = self.ui.document:getPageFromXPointer(xptr)
+    local index = self:getTocIndexByPage(pageno) -- get toc entry on the page
+    local i = index
+    if not index or not self.toc[index] then return end
+    local initial_comparison = self.ui.document:compareXPointers(self.toc[index].xpointer, xptr)
+    if initial_comparison and initial_comparison < 0 then
+        while self.toc[i] do
```
poire-z, line 225: To avoid a loop (with i=index, that you already checked it's not a candidate), move the `local i = index+/-1` here)

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -215,6 +215,34 @@ function ReaderToc:getTocIndexByPage(pn_or_xp)
...
+            local toc_xptr = self.toc[i].xpointer
+            local cmp = self.ui.document:compareXPointers(toc_xptr, xptr)
+            if cmp and cmp >= 0 then
+                return i
+            end
+            i = i + 1
```
poire-z, line 231: I can't access the previous commit where you were using `index`, but I think you want to scan backward here, so it should be `i = i - 1` until you find `i` where `cmp>=0`.
Also, please add a comment along `if cmp and cmp >= 0 then -- toc_xptr is before xptr` (and the opposite alongside the `else` so we don't have to go look at base/cre.cpp to know what a return valud of +1/0/-1 means :)

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -215,6 +215,34 @@ function ReaderToc:getTocIndexByPage(pn_or_xp)
     return pre_index
 end
 
+function ReaderToc:getAccurateTocIndexByXPointer(xptr)
+    local pageno = self.ui.document:getPageFromXPointer(xptr)
+    local index = self:getTocIndexByPage(pageno) -- get toc entry on the page
```
poire-z, line 220: It's a bit confusing here. Looking at that, it seems you could get in an infinite loop with:
getTocIndexByPage calling getAccurateTocIndexByXPointer callling getTocIndexByPage calling getAccurateTocIndexByXPointer ...
I guess you get out of that infinite loop because on first call you provide the xp string, and on the 2nd call, you provide a pageno integer.
But it's quite confusing :)
So, may be there's a clearer way to do that, and if not, please add some comments about why doing it that way without risk.

Except for that, I'm happy with the refactoring :)

poire-z:
Unit tests are not my friends, but often, they tell us the truth and that we might have missed something.
It might be the case for this one - and if not, your comment is wrong:

> if bookmark.pos0 then -- this prevents tests from failing, shouldn't be necessary otherwise

The tests in questions call this:
https://github.com/koreader/koreader/blob/4d8cf983e26008a15451ff4255a816f0b392d626/frontend/apps/reader/modules/readerbookmark.lua#L534-L548

So, dogear bookmarks dont have a `pos0` / `pos1`, but only `page` (which is a xpointer too, some kind of `pos0`).
So, either you go at adding a `chapter=` to dogear bookmarks (which would be nice, for completeness and future features that could use it - but ok if you don't),
Or you update your comment to say dogear bookmarks don't need chapter :)

Galunid:
Actually, I didn't even know there is something called dogear. Hopefully this fixes that.

poire-z:
That's what you get when you tap the top right corner of screen.

Looks allright and quite clean in the end.
I'd still wait to merge this after next release (@Frenzie: this weekend?) - just to get another month to validate the migration and stuff (and not mess bookmarks and highlighs of a larger amount of users and need to do quick another migration/fix and rerelease :)

Frenzie:
@poire-z If I find the time, hopefully. :-)

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -215,7 +215,38 @@ function ReaderBookmark:gotoBookmark(pn_or_xp)
...
+function ReaderBookmark:updateHighlightsIfNeeded()
+    local version = self.ui.doc_settings:readSetting("bookmarks_version") or 0
+    if version >= 20200511 then
+        return
+    end
+
+    for _, highlights in pairs(self.view.highlight.saved) do
+        for _, highlight in pairs(highlights) do
+            local pg_or_xp = self.ui.document.info.has_pages and
+                    highlight.pos0.page or highlight.pos0
```
poire-z, line 228: I don't know much about PDF, but can you check how this works with PDFs ?
It looks to me `highlight.pos0.page` does not exist. When highlighting in a PDF, I get saved:
```lua
    ["highlight"] = {
        [1] = {
            [1] = {
                ["drawer"] = "lighten",
                ["datetime"] = "2020-05-24 16:57:39",
                ["pboxes"] = {
                    [1] = {
                        ["y"] = 887,
                        ["x"] = 112,
                        ["h"] = 33,
                        ["w"] = 525
                    }
                },
                ["pos0"] = {
                    ["y"] = 891.06819958378,
                    ["x"] = 388.23636363636
                },
                ["pos1"] = {
                    ["y"] = 891.06819958378,
                    ["x"] = 388.23636363636
                },
                ["text"] = "blah"
            }
        }
    },
```
The page looks like it's actually the key `k` from `for k, v in pairs(self.view.highlight.saved) do`.

(The only other place I found that would use pos0.page is in `ReaderBookmark:isBookmarkSame()`, where it serves nothing but won't cause harm as `nil == nil` :)

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -215,7 +215,38 @@ function ReaderBookmark:gotoBookmark(pn_or_xp)
...
+        end
+    end
+
+    for _, bookmark in ipairs(self.bookmarks) do
+        if bookmark.pos0 then
+            local pg_or_xp = self.ui.document.info.has_pages and
+                    bookmark.pos0.page or bookmark.pos0
+                local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
+            bookmark.chapter = chapter_name
+        elseif bookmark.page then -- dogear bookmark
```
poire-z, line 240: Also `bookmark.pos0.page` with PDF:
```lua
    ["bookmarks"] = {
        [4] = {
            ["page"] = 5,
            ["highlighted"] = true,
            ["notes"] = "T steps. We use the T simulation",
            ["datetime"] = "2020-02-27 12:19:58",
            ["pos1"] = {
                ["page"] = 5,
                ["x"] = 354.14966868083,
                ["y"] = 278.53742190043,
                ["rotation"] = 0,
                ["zoom"] = 1.5473683825295
            },
            ["pos0"] = {
                ["page"] = 5,
                ["x"] = 222.31293070475,
                ["y"] = 281.12245597839,
                ["rotation"] = 0,
                ["zoom"] = 1.5473683825295
            }
        },
        [1] = {
            ["page"] = 8,
            ["highlighted"] = true,
            ["notes"] = "converges uniformly on compact sets forthe case of random walk Metropolis algorithm with proposal",
            ["datetime"] = "2020-05-24 17:06:49",
            ["pos1"] = {
                ["y"] = 529.75777777778,
                ["x"] = 349.50444444444
            },
            ["pos0"] = {
                ["y"] = 520.07150326797,
                ["x"] = 296.26222222222
            }
        },
```
The highlight I just made today has no `pos0.page` - while one I made some months ago has one... (no idea why, don't wanna investigate :) Our highlights are really a mess...)
But they both have a `page`

Galunid: Ok, that doesn't make sense :)
I checked my highlights in pdfs and all had `pos0.page` property, that's why I did it like that. I'll fix them soon. Actually, I'm not sure if my highlights had `page` property, I'll need to investigate some more.

Galunid: Hopefully I fixed it, it's probably some old version, that didn't add `page` to `pos0`. Could you confirm, that doesn't happen with highlights? I tried to replicate, but couldn't. I took a look at v2019.01 `readerhighlight.lua`/`readerbookmark.lua` and it seems it was already adding `pos0.page`. This is probably a result of some ancient version.

poire-z: > it's probably some old version, that didn't add page to pos0

Well, the bookmark that didn't have pos0.page was made the day I posted that comment :)
`["datetime"] = "2020-05-24 17:06:49",`
(don't have time to look at the code what may make it add it and what not)

> Could you confirm, that doesn't happen with highlights?

In my other comment above, https://github.com/koreader/koreader/pull/6146#discussion_r429646441 , it looks like I saw it happening with highlights too.

poire-z:
Regarding last 2 code domments above.
It seems ReaderHighlight uses a few `self.hold_pos = self.view:screenToPageTransform(ges.pos)`, which does:
https://github.com/koreader/koreader/blob/0fb9113078bb727134a010178a1caedd79e47f8e/frontend/apps/reader/modules/readerview.lua#L230-L240
which may call:
https://github.com/koreader/koreader/blob/0fb9113078bb727134a010178a1caedd79e47f8e/frontend/apps/reader/modules/readerview.lua#L348-L366
which adds to `pos` some keys named `x/y/page/zoom/rotation` - and might not add them in some cases - or may be they are added, but they can be _not defined_ in `state`.

Anyway, don't want to investigate how PDF/kopt highlights works :)
Best to pick the `page` from a saner place :) (that I mentionned above, different for highlight and for bookmarks - if I'm not wrong).

Galunid:
https://github.com/koreader/koreader/blob/1f994f8ede420de9140e8a64fc4de7443b8b15a2/frontend/apps/reader/modules/readerhighlight.lua#L1137-L1145

So as a final fix, I decided to take `page` from `self.hold_pos.page`. Based on the above snipped, it should be guaranteed to be present. As for `ReaderBookmark:updateHighlightsIfNeeded()`, for highlights I use it's key in the `self.view.highlight.saved[page]` table and `bookmark.page` for bookmarks. `bookmark.page` is guaranteed by `ReaderHighlight:getHighlightBookmarkItem`, so there shouldn't be any problems. Any other suggestions @Frenzie @poire-z ? Sorry for the delay, but it's exam period for me right now.

Frenzie:
Not right now, no.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -215,7 +215,38 @@ function ReaderBookmark:gotoBookmark(pn_or_xp)
...
+            local pg_or_xp = self.ui.document.info.has_pages and
+                    bookmark.page or bookmark.pos0
+                local chapter_name = self.ui.toc:getTocTitleByPage(pg_or_xp)
+            bookmark.chapter = chapter_name
+        elseif bookmark.page then -- dogear bookmark
+            local chapter_name = self.ui.toc:getTocTitleByPage(bookmark.page)
+            bookmark.chapter = chapter_name
+        end
+    end
+    self.ui.doc_settings:saveSetting("bookmarks_version", 20200511)
```
poire-z, line 245: Can you just update this `20200511` (here and above) to something closer to the day this will be merged, let's say `20200615` if a release happen this weekend :)
So we get a more accurate date for when this migration has happened.

Galunid: done :+1:

PR #6147, opened 2020-05-14
Some more Kobo startup script tweaks

NiLuJe:
This works purely because busybox (GNU grep wouldn't stand for this ;p).

(c.f., http://trac.ak-team.com/trac/browser/niluje/Configs/trunk/Kindle/Misc/zshenv#L39 for an explanation).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6147)

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -103,7 +103,7 @@ if [ "${VIA_NICKEL}" = "true" ]; then
         # shellcheck disable=SC2046
         export $(grep -s -E -e '^(DBUS_SESSION_BUS_ADDRESS|NICKEL_HOME|WIFI_MODULE|LANG|WIFI_MODULE_PATH|INTERFACE)=' "/proc/$(pidof -s nickel)/environ")
         # NOTE: Quoted variant, w/ the busybox RS quirk (c.f., https://unix.stackexchange.com/a/125146):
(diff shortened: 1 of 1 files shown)


NiLuJe:
Bonus point about the "it's unquoted!" thing: it's actually safer now, because we only feed garbage to export, instead of eval.

Not that this matters in practice, since none of the stuff we siphon requires quoting.

platform/kobo/koreader.sh
```diff
@@ -100,8 +100,8 @@ if [ "${VIA_NICKEL}" = "true" ]; then
     # If we were spawned outside of Nickel, we'll need a few extra bits from its own env...
     if [ "${FROM_NICKEL}" = "false" ]; then
         # Siphon a few things from nickel's env (namely, stuff exported by rcS *after* on-animator.sh has been launched)...
-        eval "$(xargs -n 1 -0 <"/proc/$(pidof -s nickel)/environ" | grep -s -F -e DBUS_SESSION_BUS_ADDRESS -e NICKEL_HOME -e WIFI_MODULE -e LANG -e WIFI_MODULE_PATH -e INTERFACE)"
-        export DBUS_SESSION_BUS_ADDRESS NICKEL_HOME WIFI_MODULE LANG WIFI_MODULE_PATH INTERFACE
+        # shellcheck disable=SC2046
+        export $(grep -s -E -e '^(DBUS_SESSION_BUS_ADDRESS|NICKEL_HOME|WIFI_MODULE|LANG|WIFI_MODULE_PATH|INTERFACE)=' "/proc/$(pidof -s nickel)/environ")
```
Frenzie, line 104: scary :-P

NiLuJe: Not as scary as [this](https://unix.stackexchange.com/a/125549) :D.

(Apparently, doing this stuff portably and safely is harder than it ought to be ^^).

PR #6163, opened 2020-05-18
Fix/exit dir

mwoz123:
fix exiting dir (when current doc remove from history)
releated: #6128

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6163)

### DIFF
--- a/frontend/apps/reader/readerui.lua
+++ b/frontend/apps/reader/readerui.lua
@@ -428,7 +428,14 @@ function ReaderUI:init()
     -- end
 end
 
+function ReaderUI:setLastDirForFileBrowser(dir)
+    self.last_dir_for_file_browser = dir
+end
+
 function ReaderUI:getLastDirFile()
+    if to_file_browser and self.last_dir_for_file_browser then
+        return self.last_dir_for_file_browser
+    end
     local QuickStart = require("ui/quickstart")
     local last_dir
     local last_file = G_reader_settings:readSetting("lastfile")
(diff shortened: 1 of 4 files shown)


frontend/apps/reader/readerui.lua
```diff
@@ -447,7 +456,7 @@ function ReaderUI:showFileManager(file)
         last_dir, last_file = util.splitFilePathName(file)
         last_dir = last_dir:match("(.*)/")
     else
-        last_dir, last_file = self:getLastDirFile()
+        last_dir, last_file = self:getLastDirFile(true)
```
poire-z, line 459: Either you remove the `true` here (as it's not expected and managed by `ReaderUI:getLastDirFile()` above.
Or you have it manage it as in my sample code:
```lua
+ function ReaderUI:getLastDirFile(to_file_browser) 
+    if to_file_browser and self.last_dir_for_file_browser then
+        return self.last_dir_for_file_browser
+    end
```

(It's possible `ReaderUI:getLastDirFile()` might not be used in other context, but I guess it makes sense to trigger that explicitely with a `to_file_browser=true` parameter)

mwoz123: I overlooked that function parameter. 
That's why it didn't want to work at first time!;)

plugins/send2ebook.koplugin/main.lua
```diff
@@ -227,6 +227,7 @@ function Send2Ebook:onCloseDocument()
         logger.dbg("Send2Ebook: download_dir_path:", download_dir_path)
         logger.dbg("Send2Ebook: removing Send2Ebook file from history.")
         ReadHistory:removeItemByPath(document_full_path)
+        self.ui:setLastDirForFileBrowser(document_full_path)
```
poire-z, line 230: Shouldn't it be `download_dir_path` (a dir, instead of document_full_path which is a file) ?
The same below for walalbag?

PR #6164, opened 2020-05-18
[wip] "move to archive" as option in after read book

mwoz123:
As I promised to someone that I'll create PR for it:

init draft. 
Is there a way to check if plugin is enabled and if so then import it?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6164)

### DIFF
--- a/frontend/apps/reader/modules/readerstatus.lua
+++ b/frontend/apps/reader/modules/readerstatus.lua
@@ -114,6 +114,13 @@ function ReaderStatus:onEndOfBook()
             },
             {
                 {
+                    text = _("Move to archive"),
+                    callback = function()
+                        local MoveToArchive = require("plugins/movetoarchive.koplugin/main")
+                        MoveToArchive:moveFileToArchive(self.document.file)
+                    end,
+                },
+                {
                     text = _("Cancel"),
                     callback = function()
                         UIManager:close(choose_action)
--- a/plugins/movetoarchive.koplugin/main.lua
+++ b/plugins/movetoarchive.koplugin/main.lua
@@ -76,21 +76,25 @@ function MoveToArchive:addToMainMenu(menu_items)
 end
 
 function MoveToArchive:moveToArchive()
+    self:moveFileToArchive(self.ui.document.file)
+end
+
+function MoveToArchive:moveFileToArchive(file)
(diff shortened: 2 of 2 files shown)


poire-z:
> Is there a way to check if plugin is enabled and if so then import it?

Not sure.
Usually, plugins plugged themselves into code modules, or replace core methods by their own (like statistics plugins replace some stuff in footer).

There is also a thing called PluginShare that I guess you could use to add your own buttons into, and ReaderStatus could just find some stuff in here and add it to its buttons set.
^ But I guess you could do that directly.

In ReaderStatus, have a `self.additional_actions = {}`
and in the plugin, just use `table.insert(self.ui.status.additional_actions, { my button title and action code in some way or another })`

mwoz123:
are you sure it's `additional_actions` ? 
I searched for it but couldn't find. 
Also didn't find any releated objects searching for `_actions` and `additional_` .

Was it moved/ removed?

poire-z:
> I searched for it but couldn't find.

Because we don't have it yet. But you could add it (meaning: have ReaderStatus dynamically create and add buttons from this new table to its own set of buttons).
(Just suggesting some ways to go solving your issue - dunno if that will work and won't be too ugly.)

frontend/apps/reader/modules/readerstatus.lua
```diff
@@ -113,6 +113,13 @@ function ReaderStatus:onEndOfBook()
                 },
             },
             {
+                {
+                    text = _("Move to archive"),
+                    callback = function()
+                        local MoveToArchive = require("plugins/movetoarchive.koplugin/main")
```
Frenzie, line 119: I would probably prefer to do that more like:

https://github.com/koreader/koreader/blob/1e244d035832d3abd3cf27625d94aeae08c5e8e0/frontend/apps/reader/modules/readerlink.lua#L594-L602

mwoz123:
I probably won't have time soon to continue work on this.
Guys feel free to overtake it.

grasegger:
I'd really like this, is this a good first issue (I have never developed for koreader)?

Frenzie:
I'm not quite sure how easy it'd be to find your way around but what @poire-z said shouldn't be too hard in principle.

PR #6167, opened 2020-05-19
desktop: add support for 3rd party dictionary apps

pazos:
Like #5108 but for desktop.

* Add support for goldendict in Linux and Mac
* Add support for apple dictionary on Mac.

goldendict

Users can also add other dictionaries, which will be appended in the list. Both commands and urls work.
```lua
return {
    { "MerriamWebster", "Merriam-Webster", false, "https://www.merriam-webster.com/dictionary/" },
    { "Wolfram", "WolframAlpha", false, "https://www.wolframalpha.com/input/?i=" },
    { "Qstardict", "QStardict", false, "qstardict" },
}

```

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6167)

### DIFF
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -41,6 +41,34 @@ local function getLinkOpener()
     return enabled, tool
 end
 
+-- differenciate between urls and commands
+local function isUrl(s)
+    if type(s) == "string" and s:match("*?://") then
+        return true
+    end
+    return false
+end
+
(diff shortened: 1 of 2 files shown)


frontend/device/sdl/device.lua
```diff
@@ -41,6 +41,34 @@ local function getLinkOpener()
     return enabled, tool
 end
 
+-- differenciate between urls and commands
```
Frenzie, line 44:
```suggestion
-- differentiate between urls and commands
```

PR #6170, opened 2020-05-19
init wallabag queue

mwoz123:
idea:
https://github.com/koreader/koreader/issues/6119#issuecomment-624329183

not tested yet

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6170)

### DIFF
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -3,7 +3,6 @@
 ]]
 
 local BD = require("ui/bidi")
-local ConfirmBox = require("ui/widget/confirmbox")
 local DataStorage = require("datastorage")
 local DocSettings = require("docsettings")
 local Event = require("ui/event")
@@ -578,8 +577,8 @@ function Wallabag:synchronize()
         info = InfoMessage:new{ text = _("Adding articles from queue…") }
         UIManager:show(info)
         UIManager:forceRePaint()
-        for url in self.download_queue do
-            self:addArticle(url)
+        for articleUrl in self.download_queue do
+            self:addArticle(articleUrl)
         end
         self.download_queue = {}
         self:saveSettings()
@@ -1060,8 +1059,8 @@ function Wallabag:getLastPercent()
     end
 end
 
-function Wallabag:addToDownloadQueue(url)
-    table.insert(self.download_queue, url)
+function Wallabag:addToDownloadQueue(articleUrl)
+    table.insert(self.download_queue, articleUrl)
     self:saveSettings()
 end
 
--- a/plugins/wallabag.koplugin/main.lua
+++ b/plugins/wallabag.koplugin/main.lua
@@ -1018,7 +1018,7 @@ function Wallabag:onAddWallabagArticle(article_url)
     if not NetworkMgr:isOnline() then
         self:addToDownloadQueue(article_url)
         UIManager:show(InfoMessage:new{
-            text = _("Article added to download queue"),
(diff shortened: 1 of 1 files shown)


plugins/wallabag.koplugin/main.lua
```diff
@@ -1003,7 +1017,18 @@ end
 
 function Wallabag:onAddWallabagArticle(article_url)
     if not NetworkMgr:isOnline() then
-        NetworkMgr:promptWifiOn()
+        UIManager:show(ConfirmBox:new{
+            text = _("No Wi-Fi connection. Connect now or add to queue for later download?"),
```
Frenzie, line 1021: If the proposed online/offline distinction is useful (of which I'm not convinced) then I'd suggest solving the problem elsewhere for better UX instead of constantly having to deal with a popup.

mwoz123: What you propose?

Frenzie: I'd just silently add it to the queue. Is there a particular reason not to?

mwoz123: let me at least display for 1 second info that it has been added to queue. Ok?

Frenzie: Yes, I just meant the logical alternative to what happens when online.

plugins/wallabag.koplugin/main.lua
```diff
@@ -981,6 +993,7 @@ function Wallabag:saveSettings()
         is_sync_remote_delete = self.is_sync_remote_delete,
         articles_per_sync     = self.articles_per_sync,
         remove_finished_from_history = self.remove_finished_from_history,
+        download_queue      = self.download_queue,
```
Frenzie, line 996: If you do ASCII art you have to do it properly. ;-)
```suggestion
        download_queue        = self.download_queue,
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -1003,7 +1016,11 @@ end
 
 function Wallabag:onAddWallabagArticle(article_url)
     if not NetworkMgr:isOnline() then
-        NetworkMgr:promptWifiOn()
+        self:addToDownloadQueue(article_url)
+        UIManager:show(InfoMessage:new{
+            text = _("Article added to download queue"),
```
Frenzie, line 1021: We should keep it uniform.
```suggestion
            text = T(_("Article added to download queue:\n%1"), BD.url(article_url)),
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -1042,6 +1059,11 @@ function Wallabag:getLastPercent()
     end
 end
 
+function Wallabag:addToDownloadQueue(articleUrl)
```
Frenzie, line 1062: Coding style, articleUrl would be a function.
```suggestion
function Wallabag:addToDownloadQueue(article_url)
```

plugins/wallabag.koplugin/main.lua
```diff
@@ -1042,6 +1059,11 @@ function Wallabag:getLastPercent()
...
+    table.insert(self.download_queue, articleUrl)
```
Frenzie, line 1063:
```suggestion
    table.insert(self.download_queue, article_url)
```

mwoz123:
>Lgtm besides the final comments; is that not yet tested thing still true?
untrue

PR #6173, opened 2020-05-20
Statistics: fix number of highlights

poire-z:
Fixes issue noticed at https://github.com/koreader/koreader/pull/6052#issuecomment-629610275: no error log message anymore, and the number of highlights will be updated (and corrected if previously wrong because leaked from another book) to the number of highlights in the settings on each book opening. 
Note that the number of "Notes" shown does not mean much.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6173)


poire-z:
Well, not that complicated, it's just counting things :)
But yes, i didn't see any other way - although this is quite allright (except for the additional counting at book opening, which shouldn't be that expensive).

PR #6177, opened 2020-05-23
unified calibre plugin

### LINKED ISSUE #1916: Wrong SEARCH_LIBRARY_PATH
- KOReader version: koreader-stable-v2015.11
- Device: Kobo Glo HD
#### Issue

With this device I need to force the default search library path to:

defaults.persistent.lua:

```
SEARCH_LIBRARY_PATH = "/mnt/onboard/"
SEARCH_LIBRARY_PATH2 = "/mnt/onboard/"
```

(I don't know if I only need one of them or both.  Although, anyway, all it is working well right now).
#### Steps to reproduce

When I try to search a specific book using calibre's db and koreader shows the results, and then I try to load this book, koreader doesn't find it (wrong path).

PS: I don't know if this is still present in "nightly" versions (I keep the same defaults.persistent.lua file with them).

### LINKED ISSUE #3028: Search calibre catalog from inside a book
* KOReader version: 2015.11-1054
* Device: Kobo Glo

I wonder what is the reason for not having the "Find book in calibre catalog" entry in the, otherwise very similar, Find menu while reading a book.

### LINKED ISSUE #5911: Calibre not retaining book status
* KOReader version: v2020.02
* Device: KPW4

#### Issue

If I push a book to my Kindle via Calibre wireless device connection, everything works. *.calibre files are created at USB storage root, too (``/mnt/us``), and the books are shown as on the device. However, once I disconnect and reconnect to Calibre, the books are shown as not on the device, even though they are.

#### Steps to reproduce

1. Connect to Calibre via wireless device connection
2. Push some books
3. Disconnect from Calibre
4. Reconnect to Calibre

### LINKED ISSUE #6219: [question]: calibre search metadata refactor
In #6177 I implemented a few calibre wireless fixes.

I ended up writting a way to [read and write calibre metadata directly](https://github.com/koreader/koreader/pull/6177/files#diff-040c04971ea664bab5607114ae52de35). That's nice as it makes the code easy to read and allow us to keep the metadata updated across calibre connections (wired via device drivers, wired via connect to folder or wireless via the "companion" plugin).

The implementation works well with calibre but fails to work with our current "search calibre metadata feature". Here's why:

1. with #6177 each "inbox_dir" used with the calibre wireless plugin becomes a full calibre library (it has .driveinfo.calibre and .metadata.calibre files)
2. the code in Filemanager local function [`findcalibre`](https://github.com/koreader/koreader/blob/master/frontend/apps/filemanager/filemanagersearch.lua#L44-L74) will find all that metadata.
3. The full https://github.com/koreader/koreader/blob/master/frontend/apps/filemanager/filemanagersearch.lua
file is intended to **manually** parse a calibre metadata file, as pretty-printed by calibre, but fails to parse the same file if we store it ourselves - without prettyprint. It is still valid JSON data, and calibre can read it without problems.

Since I'm not an user of the "search in calibre metadata" I barely know how it behaves from the user POV. Things are even harder at trying to read `filemanagersearch.lua` code. I'm full aware that most of the file is used to find calibre metadata files, save their path as global settings and write parts of their metadata in "simplified" files to be consumed by the search; but I cannot understand how the search works.

I would love a helper hand to help me figure out how to move current search functions to the new calibre db helper. Mainly, with the new helper we need to do:

1. find a calibre library of your choice
2. initializate it with `CalibreDB:init(path)`
3. all books are now stored in an indexed table `CalibreDB.books` that can be iterated direclty by a search function or used by any of the embedded functions, like `CalibreDB:searchTags(tag)`. 

Now the real question: **how to migrate the UI part of `filemanagersearch.lua` to use `CalibreDB`?**

pazos:
### unified calibre plugin

joins calibre metadata search and calibre wireless connections into a single plugin: `calibre.koplugin` which handles all integration with calibre except OPDS.

#### search metadata changes:

- search directly into calibre metadata files.
- search can be performed on more than one library (configurable from a menu)
- device scans now find all calibre libraries under a given root
- search options can be configured from a menu. (case sensitive, find by title, author and path)
- removed legacy global variables.
- *option* to search from the reader
- *option* to generate a cache of books for faster searches.

#### calibre wireless connection changes:

- keep track of books in a library (includes prunning books from calibre metadata if the file was deleted locally)
- remove files on device from calibre
- support password protected connections
- FM integration: if we're in the inbox dir it will be updated each time a book is added or deleted.
- disconnect when requested by calibre, available on newer calibre versions (+4.17)
- remove unused opcodes.
- better report of client name, version and device id
- free disk space checks for all calibre versions
- bump supported extensions to match what KOReader can handle. Users can override this with their own list of extensions (or from calibre, by configuring the wireless device).

People that don't use calibre at all can disable the plugin and remove all these options from the menu :smile: 

Fixes #5911
Fixes #6219
Fixes #3028
Fixes #1916

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6177)


plugins/calibrecompanion.koplugin/main.lua
```diff
@@ -11,6 +11,44 @@ local T = require("ffi/util").template
...
+        f:write("return "..dump(t))
+        f:close()
+    end
+end
+
+-- delete file
+local function delete(file)
+    if not file then return end
+    if lfs.attributes(file, "mode") == "file" then
+        os.remove(file)
```
Frenzie, line 48: With or without .sdr?

pazos: I would say without. Mainly because:

1. There's no way to sync our metadata with calibre yet.
2. User can restore the book from its calibre library. Book will be named the same, so the same metadata will work again.

Frenzie: Alright. :+1:

plugins/calibrecompanion.koplugin/main.lua
```diff
@@ -49,6 +49,44 @@ local function delete(file)
...
+    end
+    -- we only work in valid directories
+    if not dir or lfs.attributes(dir, "mode") ~= "directory" then
+        return diskUsageFallback()
+    end
+    -- fill a table with total, available and used kb of the mountpoint that holds the directory,
+    -- this relies on common shell utilities: cut, df, grep & sed.
+    local std_out =  io.popen(
+        "df "..dir.." 2>/dev/null | sed -r 's/ +/ /g' "..
+        " | grep /dev | cut -d ' ' -f 2,3,4"
```
pazos, line 70: Pinging @NiLuJe about this one. Does it work on kobos, kindles?. It works on cervantes and should work as well in the Remarkable. It certainly works on desktop.

Frenzie: See the system stats plugin.

Frenzie: Or was it a function in util (am on phone)

pazos: system stats isues a general df command and greps the directory. Here we rely on df to figure out the mountpoint of a given dir. It should work fine even on busybox df but I'm not sure.

Frenzie:
```
# df /mnt/onboard 2>/dev/null | sed -r 's/ +/ /g' | grep /dev | cut -d ' ' -f 2,3,4
3316908 2945656 371252
```

NiLuJe: If you want a battle-tested (MRPI) variant with fewer forks:

```
df -k /mnt/onboard | awk '$3 ~ /[0-9]+/ { print $2,$3,$4 }'
```

frontend/util.lua
```diff
@@ -563,6 +563,61 @@ function util.makePath(path)
...
+    elseif file then
+        return nil, file .. " is not a file"
+    else
+        return nil, "file is nil"
+    end
+end
+
+-- Gets total, used and available bytes for a given directory.
+-- If the directory does not exists (or any other error happens) it
+-- will return fallback values.
```
Frenzie, line 582: I'm not sure I quite understand the fallback here. If it needs to be a table with numbers and can't be something like false or nil, how about `-1` for example?

plugins/calibrecompanion.koplugin/main.lua
```diff
@@ -430,9 +541,14 @@ function CalibreCompanion:getBookCount(arg)
...
         willStream = true,
         willScan = true,
-        count = 0,
+        count = #self.book_list
     }
     self:sendJsonData('OK', books)
+    for i, _ in ipairs(self.book_list) do
+        local book = self.book_list[i]
+        book.priKey = i
+        self:sendJsonData('OK', book)
```
Frenzie, line 550: Probably best to stick with our modern coding style. ;-)
```suggestion
        self:sendJsonData("OK", book)
```

plugins/calibrecompanion.koplugin/sha1.lua
```diff
@@ -0,0 +1,330 @@
+-- $Revision: 1.5 $
```
Frenzie, line 1: Should we put this in base?

frontend/util.lua
pazos: Ok. We need the fallback values in calibre but it can return nil here and in calibre use the fallback.

plugins/calibrecompanion.koplugin/main.lua
```diff
@@ -11,6 +12,25 @@ local T = require("ffi/util").template
 
 require("ffi/zeromq_h")
 
+-- supported formats
+local extensions = require("extensions")
+local function getExtensionPathLengths()
+    local t = {}
+    for _, v in pairs(extensions) do
+        t[v] = 37
```
poire-z, line 20: What's this 37 ? Some opcode like the another anonymous numbers in this file ? Some real path length, but why 37 (while it was 42 before - but why 42 ? because it's the answer ? so arbitrary ? so many questions :)
Might be worth some comments, as you must have investigated all that, so don't let the knowledge be lost :)

plugins/calibrecompanion.koplugin/sha1.lua
poire-z: Yep, could be used one day by other code than this plugin.

plugins/calibrecompanion.koplugin/main.lua
pazos: It is some magical thing. I really don't know. It could be the maximum file name until the extension or any other stuff. I did change it to 37 because it is what Calibre Companion and Uncaged reported to calibre server. I do trust what CC/uncaged are doing. Plus it was the only difference on the `calibre-debug -g` logs.

It makes sense to notify the server which is the maximum length of a filename, so it can feed books with adjusted lpath.

plugins/calibrecompanion.koplugin/main.lua
pazos: Single quotes are used across the whole file. Shall I change all of them?. See opcodes

plugins/calibrecompanion.koplugin/sha1.lua
pazos: Sounds cool. Could you please do it for me?. I'm not sure where to place the file and have no idea how to exclude it from shellcheck. It is spawning a lot of warnings

Frenzie: I'd just put it in base/ffi. The bit manipulations are close enough, so to speak. Pinging @NiLuJe for another opinion.

You can exclude files in .luacheckrc, see https://github.com/koreader/koreader-base/blob/4a91eefe3a42f926ebe262658c258c9929544836/.luacheckrc#L12-L16

Frenzie: Or with specific files rather than wildcards:

https://github.com/koreader/koreader/blob/0fb9113078bb727134a010178a1caedd79e47f8e/.luacheckrc#L134-L141

We could also say this is more like [luxl](https://github.com/koreader/koreader/blob/0fb9113078bb727134a010178a1caedd79e47f8e/frontend/luxl.lua)… although there I'm inclined to wonder why it's in front.

plugins/calibrecompanion.koplugin/main.lua
Frenzie: Sure, but maybe as a follow-up for git blame purposes in the future?

plugins/calibrecompanion.koplugin/main.lua
Frenzie: Agreed, probably best to add a `-- magic number from calibre` comment or something. ^_^

frontend/util.lua
Frenzie: Sounds good!

plugins/calibrecompanion.koplugin/sha1.lua
```diff
@@ -0,0 +1,330 @@
...
+-- 40-byte hex string.
+--
+--
+-- The two "bin" versions do the same, but return the 20-byte string of raw
+-- data that the 40-byte hex strings represent.
+--
+-------------------------------------------------------------------------------
+--
+-- Description
+-- Due to the lack of bitwise operations in 5.1, this version uses numbers to
```
NiLuJe, line 40: Sidebar: That applies to Lua 5.1, not LuaJIT, which ships a JIT-enabled bitop module by default ;).

plugins/calibrecompanion.koplugin/sha1.lua
NiLuJe: That'd be a +1 on moving it to base, and looking into https://github.com/mpeterv/sha1 instead, which'll likely be much faster ;).

plugins/calibrecompanion.koplugin/sha1.lua
Frenzie: I skimmed through it really quickly and I failed to notice it wasn't using bit.bor.

On a related note: https://github.com/Egor-Skriptunoff/pure_lua_SHA/blob/304d4121f080e68ef209d3f5fe093e5a955a4978/sha2.lua

> this is the fastest pure Lua implementation of SHA you can find.

It also does MD5. This might be worth checking out. (MIT license)

NiLuJe: Ooh, shiny :). And it does partial MD5, which we need.

plugins/calibrecompanion.koplugin/sha1.lua
pazos: Wow. Indeed your suggestion is several orders of magnitude faster. It isn't very noticeable on the usage here but in a `for i = 1, 1000` loop here are the results:

the one in this post:

```
real	0m13,375s
user	0m13,369s
sys	0m0,004s
```

mpeterv one using LuaJIT bitops:

```
real	0m0,030s
user	0m0,029s
sys	0m0,001s
```

plugins/calibrecompanion.koplugin/main.lua
pazos: Sure. Maybe @shermp can shed some light on this issue.

shermp: I'm afraid this is a magic number to me as well. I just copied whatever CC was using.

Looking at the Calibre source, it looks like this is part of determining max path length. Also does some sums with `PATH_FUDGE_FACTOR`, which in the source, is:
```
    # guess of length of MTP name. The length of the full path to the folder
    # on the device is added to this. That path includes the device's mount point
    # making this number effectively around 10 to 15 larger.
    PATH_FUDGE_FACTOR           = 40
```
Seems to be primarily used here:
```
maxlen = (self.MAX_PATH_LEN - (self.PATH_FUDGE_FACTOR +
                   self.exts_path_lengths.get(dotless_ext, self.PATH_FUDGE_FACTOR)))
```

pazos:
Now a question to @poire-z:

I want to update the filemanager each time a book is added or removed. The following code works fine:

```lua
local function updateDir(dir)
    if FileManager.instance then
        FileManager.instance:reinit(dir)
    end
end
``` 

If possible I would like to trigger the reinit if the filemanager is **currently** on the same dir. I tried comparing `dir` against `FileManager.instance.root_path` but they are always equal.

poire-z:
Not sure but it looks like `FileManager.root_path` is given to the `FileChooser` sub-instance - and it's there that may be it has a life of its own (and `root_path` is may be not updated when you browse?).
So, may be you could check `dir` against `FileManager.instance.file_chooser.path`.

(If that's right, it might be a bit ugly checking such internal stuff from outside, so you might want to create `FileManager:getCurrentlyBrowsedDirectory() return self.file_chooser.path end` or something like that.)

pazos:
Thanks! Indeed adding the following function works:

```lua
function FileManager:getCurrentDir()
    if self.instance then
        return self.instance.file_chooser.path
    end
end
```

pazos:
> disconnect when requested by calibre (without reconnecting again and again)

Calibre wireless server doesn't notify clients. It just closes the socket on [eject](https://github.com/kovidgoyal/calibre/blob/master/src/calibre/devices/smart_device_app/driver.py#L1457-%20L1459). Unless somebody figures out a way to check if the server closed the socket (something like [`checkCalibreServer`](https://github.com/koreader/koreader/blob/master/plugins/calibrecompanion.koplugin/main.lua#L85-L96) periodically) I'm going to leave that feature outside this PR.

NiLuJe:
If this were C, I'd go with a poll() loop, and you'd get a POLLHUP when the remote closes the socket ;).

NiLuJe:
But, at a quick glance, that's handled via ZMQ (https://github.com/koreader/koreader/blob/0fb9113078bb727134a010178a1caedd79e47f8e/frontend/ui/message/streammessagequeue.lua#L66), and in the hilariously old versions (seriously, there's been *at least* 4 major API revisions since :D), the czmq wrapper don't appear to care all that much about POLLHUP...

https://github.com/zeromq/czmq/blob/2a0ddbc4b2dde623220d7f4980ddd60e910cfa78/src/zpoller.c#L118

ZMQ itself does:

https://github.com/zeromq/libzmq/blob/883e95b22e0bffffa72312ea1fec76199afbe458/src/poll.cpp#L146

pazos:
> If this were C, I'd go with a poll() loop, and you'd get a POLLHUP when the remote closes the socket ;).
> 
> (I, err, may have had to deal with [sockets recently](https://github.com/NiLuJe/kfmon/blob/1958fba1bec3ecf203742e242f66d4850ffca0d1/utils/kfmon-ipc.c#L247) ;p).

The following function will work:
```lua
local function remoteClosedSocket(host, port)
    local socket = require("socket")
    local tcp = socket.tcp()
    tcp:settimeout(0)
    local ok, err = tcp:connect(host, port)
    if not ok and err == "closed" then
        -- calibre closed the socket
        return true
    elseif ok then
        tcp:close()
    end
    return false
end
```

but needs to be scheduled each few seconds because that's the timeframe between calibre closes a port and it opens it to listen to clients again.

NiLuJe:
Yeah, but having a sane poll loop would prevent the need for such an insane workaround in the first place ;).

But alas, old czmq is old ;).

Frenzie:
You're welcome to update it, but a word of warning: I tried to update it a few years ago but since I couldn't figure it out within some sane amount of time (don't recall the specifics; let's say more than 10 minutes and fewer than 60) and ran into some problems. But maybe I was just being stupid or insufficiently motivated. :-)

NiLuJe:
Oh, no, rest assured, I'm not touching that with a 10 foot pole ;p.

pazos:
Ok, everything is in place now.

> disconnect when requested by calibre (without reconnecting again and again)

Will be done in recent calibre versions. Old versions will still reconnect if the user does the request from calibre, not from KOReader.

plugins/calibrecompanion.koplugin/main.lua
pazos: I will use single quotes on this PR and change them later on :+1:

frontend/util.lua
```diff
@@ -563,6 +563,52 @@ function util.makePath(path)
...
+
+-- Gets total, used and available bytes for a given directory.
+-- @string path
+-- @treturn table with total, used and available bytes
+function util.diskUsage(dir)
+    -- safe way of testing df & awk
+    local function doCommand(d)
+        local handle = io.popen("df -k " .. d .. " 2>&1 | awk '$3 ~ /[0-9]+/ { print $2,$3,$4 }' 2>&1 || echo ::ERROR::")
+        if not handle then return end
+        local output = handle:read("*all")
```
poire-z, line 588: I don't know if it's stictly needed or if Lua will properly close the handle when GC'ing.
But we usually follow that with `handle:close()`

pazos: From https://www.lua.org/work/doc/manual.html

> Note that files are automatically closed when their handles are garbage collected, but that takes an unpredictable amount of time to happen.

I would expect the handle to be able to be collected at the end of the function, but the quote doesn't say that. So better stick with `handle:close()`

Frenzie: Not strictly, but then it won't close until whenever it decides to garbage collect.

pazos:
Now with a clean reconnect.

final

pazos:
Now, with lastest commits it is possible to import data from calibre JSON documents (if available and newer than our db data) without disturbing each library UUID. Series and tags are stored in our table/file too.

I'm wondering if it makes sense to move all calibre code from https://github.com/koreader/koreader/blob/master/frontend/apps/filemanager/filemanagersearch.lua to this plugin and use the newer, simplified, scheme of storing metadata in each dir and use that data as the source for searches.

Honestly I don't understand most of the code in `filemanagersearch.lua`, but looks quite possible to do the same searches using the ".metadata.calibre.lua" table.

NiLuJe:
IIRC, it either uses the metadata.calibre file, or a filtered, simpler version that it generates from that if up to date.

The interest of using a single flatfile at the root of the Library is that you don't have to traverse the fs to find multiple shards of metadata ;).

pazos:
> The interest of using a single flatfile at the root of the Library is that you don't have to traverse the fs to find multiple shards of metadata ;).

This PR implements that. But each initialized directory is considered its own library/device. Metadata is still at the root of the library and books are allowed in nested folders (or even outside the root of the library)

I will spend a bit of time to figure out an *universal* calibre metadata handling.

pazos:
**Update**: I'm in the process of moving all calibre related features to a single plugin (renaming it from calibrecompanion.koplugin to calibre.koplugin). It is almost done: here some screenshots:
Sin nombre

Still need to figure at way to make checkboxes and toggle them from the contents of a table like

```lua
return {
    [1] = {
        path = "/mnt/test",
        enabled = true,
    },
    [2] = {
        path = "/mnt/test1",
        enabled = true,
    }
}
```

NiLuJe:
@pazos: Building the menu out of the table, or building the table out of the menu? (or, both, I guess? ^^).

poire-z:
> Still need to figure at way to make checkboxes and toggle them from the contents of a table like

You might look at:
readerfont.lua for simple menu building from a list of external stuff, but one checkable only
readerstyletweak.lua for more complex menu building (with submenus, that I guess you wouldn't need), but multiple checkboxable items

pazos:
> Building the menu out of the table, or building the table out of the menu?

Building the menu out of the table.

> You might look at:
readerfont.lua for simple menu building from a list of external stuff, but one checkable only
readerstyletweak.lua for more complex menu building

Thank you!. I managed to get the checkboxes from the table. But now I find the, hopefully, last issue:

## menu without libraries



## clicking on rescan should give you a prompt (pointing to the rootdir for each platform, ignore "/mnt/calibre" here, it should say "/mnt" on kobos, kindles, etc)



## results of the scan



## the same menu with libraries.



To get the last image I need to reload menus, by entering reader if I was in the FM or by restarting the program. Any way to upload menus automagically without switching apps or restarting?

poire-z:
You might look at:
```
$ rgrep sub_item_table_func plugins/
plugins/evernote.koplugin/main.lua:                sub_item_table_func = function()
plugins/texteditor.koplugin/main.lua:        sub_item_table_func = function()
$ rgrep sub_item_table_func frontend/
frontend/ui/widget/menu.lua:    if item.sub_item_table ~= nil or item.sub_item_table_func then
frontend/ui/widget/touchmenu.lua:        if item.sub_item_table_func then
frontend/ui/widget/touchmenu.lua:            sub_item_table = item.sub_item_table_func()
```
I remember that for the TextEditor plugin, where I update the list of recent files when opening one, I had to do more stuff:
https://github.com/koreader/koreader/blob/879f8a7624ae0e44a6d658c08793c1f2cf4b4441/plugins/texteditor.koplugin/main.lua#L216-L234

Dunno if you'll need that.

pazos:
All right!

Time for a new review. Nothing changed too much on the wireless front. I just added an option to avoid keeping calibre waiting when it requests a book (not supported yet).

The other change is how metadata search work. @NiLuJe tested that a few days ago. Since then I changed a few things: it basically abuses indexed tables and tries to join the minimum possible strings. It also has UI settings to manage search options and folder rescans.

So, beyond code review, another "user" review would be great. I didn't implement the "book cache" yet, but shouldn't be needed to outperform the current metadata search.

NiLuJe:
@pazos: Looking pretty good ;).

My only comment thus far would be to *NOT* default to case sensitive search, because people are lazy, so defaulting to case insensitive matches is less confusing ;).

NiLuJe:
Also, since I stupidly played with the settings first: What happens if you try to search without having done a scan in the settings first?

pazos:
> What happens if you try to search without having done a scan in the settings first?

Same prompt but with a header: one for no libraries `self:prompt(_("No calibre libraries"))` and other for empty/disabled libraries `self:prompt(_("No metadata found"))`

plugins/calibre.koplugin/_meta.lua
```diff
@@ -0,0 +1,6 @@
+local _ = require("gettext")
+return {
+    name = "calibre",
+    fullname = _("Calibre"),
+    description = _([[Integration with calibre. Send documents from calibre library via Wi-Fi and search calibre metadata]]),
```
Frenzie, line 5:
```suggestion
    description = _([[Integration with calibre. Send documents from calibre library via Wi-Fi and search calibre metadata.]]),
```

plugins/calibre.koplugin/main.lua
```diff
@@ -0,0 +1,269 @@
...
+local CalibreSearch = require("search")
+local CalibreWireless = require("wireless")
+local LuaSettings = require("luasettings")
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local _ = require("gettext")
+local T = require("ffi/util").template
+
+local Calibre = WidgetContainer:new{
+    name = 'calibre',
```
Frenzie, line 13:
```suggestion
    name = "calibre",
```

plugins/calibre.koplugin/main.lua
```diff
@@ -0,0 +1,269 @@
...
+            },
+            {
+                text = _("Wireless settings"),
+                keep_menu_open = true,
+                sub_item_table = self:getWirelessMenuTable(),
+            },
+        }
+    }
+    -- insert the wireless connection at the top of plugin menu
+    table.insert(menu_items.calibre.sub_item_table, 1, {
```
Frenzie, line 40: I'm a bit confused by the use of table.insert here? For indentation?

plugins/calibre.koplugin/search.lua
```diff
@@ -0,0 +1,494 @@
...
+end
+
+-- get tags that partially match the search criteria and their frequency
+local function searchByTag(t, query, case_sensitive)
+    local freq = {}
+    for _, book in ipairs(t) do
+        for __, tag in ipairs(book.tags) do
+            -- case insensitive
+            if not query or (query and string.find(
+                string.upper(tag), string.upper(query))) then
```
poire-z, line 70: Not fond of such a like break after `string.find(` :) (here and below)

plugins/calibre.koplugin/search.lua
```diff
@@ -0,0 +1,494 @@
...
+end
+
+-- get series that partially match the search criteria and their frequency
+local function searchBySeries(t, query, case_sensitive)
+    local freq = {}
+    for _, book in ipairs(t) do
+        if type(book.series) ~= "function" then
+            if not query or (query and string.find(
+                string.upper(book.series), string.upper(query))) then
+                freq[book.series] = (freq[book.series] or 0) + 1
```
poire-z, line 85: Also not fond when the `if` continuation is at the same level as the then action.
```lua
if something and
    other thing then
    dosomestuff
end
```
I'd rather see more indentation on the if continuation:
```lua
if something and
        other thing then
    dosomestuff
end
```
or
```lua
if something
        and other thing then
    dosomestuff
end
```
Makes it easier to read.

plugins/calibre.koplugin/search.lua
Frenzie: I'm more than fine with keeping it on one line. And I agree that breaking after `string.find(` seems like a slightly odd/misleading location.

pazos:
thanks for the review. I addressed the comments. Also added a proto cache (without timestamps yet) and mesure the elapsed time in search and metadata loading. In a quick test for a little library of ~120 books on the emulator:

```
06/05/20-15:31:43 INFO searching books | case-insensitive authors title path
06/05/20-15:31:43 WARN  invalid cache cannot open ./settings/calibre-books.lua: No such file or directory
06/05/20-15:31:43 INFO  calibre info loaded from disk in 20.568132 milliseconds: 128 books. 0 pruned
06/05/20-15:31:43 INFO  metadata: 128 books imported from calibre in 20.771027 milliseconds
06/05/20-15:31:43 INFO  search done in 28.142929 milliseconds

``` 

```
06/05/20-15:32:03 INFO  searching books | case-insensitive authors title path
06/05/20-15:32:03 INFO  metadata: 128 books imported from cache in 1.882076 milliseconds
06/05/20-15:32:03 INFO  search done in 9.289026 milliseconds
``` 

Probably worth testing it with a **big** library on a **real** device

plugins/calibre.koplugin/main.lua
pazos: Nope, I don't remember why but certainly I had a good reason at a given time. Not now :p

plugins/calibre.koplugin/search.lua
pazos: Will take your suggestion for next commits. Thanks!

NiLuJe:
@pazos: Thanks, I'll give it a go on my slow-ass H2O w/ a ~500 books db ;).

NiLuJe:
No cache:

```
06/05/20-18:27:49 INFO  searching series | case-insensitive authors title 
06/05/20-18:27:52 INFO  calibre info loaded from disk in 2337.571859 milliseconds: 556 books. 0 pruned
06/05/20-18:27:52 INFO  metadata: 556 books imported from calibre in 2351.696968 milliseconds
06/05/20-18:27:52 INFO  search done in 2363.420010 milliseconds
06/05/20-18:28:07 INFO  searching series | case-insensitive authors title 
06/05/20-18:28:07 INFO  search done in 15.335083 milliseconds
06/05/20-18:28:31 INFO  searching books | case-insensitive authors title 
06/05/20-18:28:31 INFO  search done in 30.802965 milliseconds
06/05/20-18:28:38 INFO  searching books | case-insensitive authors title 
06/05/20-18:28:38 INFO  search done in 42.178869 milliseconds
```

```
06/05/20-18:29:02 INFO  searching series | case-insensitive authors title 
06/05/20-18:29:04 INFO  calibre info loaded from disk in 2161.569834 milliseconds: 556 books. 0 pruned
06/05/20-18:29:04 INFO  metadata: 556 books imported from calibre in 2175.912142 milliseconds
06/05/20-18:29:04 INFO  search done in 2191.108942 milliseconds
06/05/20-18:29:33 INFO  searching books | case-insensitive authors title 
06/05/20-18:29:33 INFO  search done in 50.467968 milliseconds
```

NiLuJe:
And I can't get you results with the cache enabled, because...

```
06/05/20-18:31:38 INFO  searching series | case-insensitive authors title 
06/05/20-18:31:38 WARN  invalid cache cannot open ./settings/calibre-books.lua: No such file or directory
06/05/20-18:31:40 INFO  calibre info loaded from disk in 2143.218994 milliseconds: 556 books. 0 pruned
06/05/20-18:31:40 INFO  metadata: 556 books imported from calibre in 2157.848835 milliseconds
./luajit: plugins/calibre.koplugin/search.lua:41: bad argument #1 to 'upper' (string expected, got nil)
stack traceback:
        [C]: in function 'upper'
        plugins/calibre.koplugin/search.lua:41: in function 'match'
        plugins/calibre.koplugin/search.lua:93: in function 'searchBySeries'
        plugins/calibre.koplugin/search.lua:503: in function 'browse'
        plugins/calibre.koplugin/search.lua:349: in function 'find'
        plugins/calibre.koplugin/search.lua:430: in function 'close'
        plugins/calibre.koplugin/search.lua:384: in function 'callback'
        frontend/ui/widget/button.lua:239: in function 'action'
        frontend/ui/uimanager.lua:812: in function '_checkTasks'
        frontend/ui/uimanager.lua:1095: in function 'handleInput'
        frontend/ui/uimanager.lua:1181: in function 'run'
        ./reader.lua:298: in main chunk
        [C]: at 0x000140c1
```

```
06/05/20-18:32:18 INFO  searching series | case-sensitive authors title 
06/05/20-18:32:18 INFO  metadata: 556 books imported from cache in 46.593904 milliseconds
./luajit: plugins/calibre.koplugin/search.lua:93: bad argument #1 to 'match' (string expected, got nil)
stack traceback:
        [C]: in function 'match'
        plugins/calibre.koplugin/search.lua:93: in function 'searchBySeries'
        plugins/calibre.koplugin/search.lua:503: in function 'browse'
        plugins/calibre.koplugin/search.lua:349: in function 'find'
        plugins/calibre.koplugin/search.lua:430: in function 'close'
        plugins/calibre.koplugin/search.lua:384: in function 'callback'
        frontend/ui/widget/button.lua:239: in function 'action'
        frontend/ui/uimanager.lua:812: in function '_checkTasks'
        frontend/ui/uimanager.lua:1095: in function 'handleInput'
        frontend/ui/uimanager.lua:1181: in function 'run'
        ./reader.lua:298: in main chunk
        [C]: at 0x000140c1
```

^^

NiLuJe:
Or, actually, I can, hadn't noticed that only the Serie search is crashing ;).

```
06/05/20-18:35:27 INFO  searching books | case-insensitive authors title 
06/05/20-18:35:27 INFO  metadata: 556 books imported from cache in 48.688173 milliseconds
06/05/20-18:35:28 INFO  search done in 81.616163 milliseconds
06/05/20-18:35:53 INFO  searching tags | case-insensitive authors title 
06/05/20-18:35:53 INFO  search done in 47.326088 milliseconds
```

So, weeeery nice :).

NiLuJe:
Okay, a simple nil guard later (because the `series` key doesn't exist for books without a series in the cache, which makes perfect sense).

```diff
diff --git a/plugins/calibre.koplugin/search.lua b/plugins/calibre.koplugin/search.lua
index d366c30d..37d9037e 100644
--- a/plugins/calibre.koplugin/search.lua
+++ b/plugins/calibre.koplugin/search.lua
@@ -89,7 +89,7 @@ end
 local function searchBySeries(t, query, case_insensitive)
     local freq = {}
     for _, book in ipairs(t) do
-        if type(book.series) ~= "function" then
+        if book.series and type(book.series) ~= "function" then
             if match(book.series, query, case_insensitive) then
                 freq[book.series] = (freq[book.series] or 0) + 1
             end
```

```
06/05/20-18:42:51 INFO  searching series | case-insensitive authors title 
06/05/20-18:42:51 INFO  metadata: 556 books imported from cache in 31.759977 milliseconds
06/05/20-18:42:51 INFO  search done in 51.898956 milliseconds
06/05/20-18:42:58 INFO  searching series | case-insensitive authors title 
06/05/20-18:42:58 INFO  search done in 8.093834 milliseconds
```

\o/

pazos:
@NiLuJe: actually the metadata search only worked on Kobo or devices where the user used "connect to folder". That's because the way the fullpath of the file was (and still is) derived is taking account the "lpath" of the file based on the path of the library. That holds true in kobos/wireless devices and connect to folder but no on others. For instance on the cervantes the "bq cervantes driver" will place books in `books/` but there's no trace of that part of relative path in book's lpath.

TL;DR: I would need the contents of `driveinfo.calibre` for a Kindle and a Kobo. Specially the fields `device_name` and `prefix` to figure out a way to search metadata and correctly open books in spite of the calibre driver used.

NiLuJe:
Kobo:

`"prefix": "/run/media/niluje/KOBOeReader/", "device_name": "Kobo Aura H2O"`

(That's obviously the usual mountpoint on my Desktop ;). That's based on the FAT label, which is also how macOS deals with it (i.e., something like `/Volumes/KOBOeReader` IIRC)).

Kindle:

`"prefix": "/run/media/niluje/Kindle/", "device_name": "Amazon Kindle"`

Same deal, label is Kindle ;).

pazos:
> I only ever use "Connect to Device" via USB

Cool. So `device_name` can be used to derive the full path of books.

On kobos the full path of books is the same of the mountpoint: `/mnt/onboard` (and `/mnt/sd`) plus the regex used in calibre to setup dirs and file names based on book attributes. Does happens as well in Kindles or they're always stored in a subpath of the USB share?

NiLuJe:
On Kindle, I do get the `documents` folder in the lpath:

`"lpath": "documents/Gaiman, Neil/American Gods_ The Tenth Anniversary Edition - Neil Gaiman.azw3"`

(That's the standard "root" for the books on Kindle. The metadata file itself is an the root of the USB partition, the parent folder of `documents`, `/mnt/us`).

NiLuJe:
On Kobo:

```
grep lpath /mnt/onboard/metadata.calibre | ag Gaiman
    "lpath": "Gaiman, Neil/American Gods_ The Tenth Anniversary Edition - Neil Gaiman.kepub.epub",
```

NiLuJe:
TL;DR (If I understood correctly):

If device_name contains (Kobo|Kindle), fullpath = libpath + lpath?

pazos:
#### loading metadata from calibre files

```
06/09/20-14:11:12 WARN  invalid cache: cannot open ./cache/calibre-books.lua: No such file or directory
06/09/20-14:11:12 INFO  calibre info loaded from disk in 8.457899 milliseconds: 133 books. 0 pruned
06/09/20-14:11:12 INFO  metadata: 133 books imported from calibre in 11.012077 milliseconds
06/09/20-14:11:12 INFO  search done in 8.471012 milliseconds (books, case sensitive: false, title: true, authors: true, path: true)
```

#### loading metadata from cache

```
06/09/20-14:11:36 INFO  metadata: 133 books imported from cache in 2.768993 milliseconds
06/09/20-14:11:36 INFO  search done in 6.221056 milliseconds (books, case sensitive: false, title: true, authors: true, path: true)
```

plugins/calibre.koplugin/search.lua
```diff
@@ -189,8 +187,8 @@ local CalibreSearch = InputContainer:new{
         "find_by_authors",
         "find_by_path",
     },
-    user_libraries = DataStorage:getSettingsDir() .. "/" .. libraries_file,
-    user_book_cache = DataStorage:getSettingsDir() .. "/" .. metadata_file,
+    user_libraries = DataStorage:getDataDir() .. "/cache/" .. libraries_file,
+    user_book_cache = DataStorage:getDataDir() .. "/cache/" .. metadata_file,
```
poire-z, line 191: Just mentionning this, to make sure you're ok with it:
https://github.com/koreader/koreader/blob/c1be488a11bc9efb3082b77affe990e189fbdbe6/frontend/apps/filemanager/filemanagermenu.lua#L241-L256
So, this menu item (that people seldom use I guess) will trash your calibre caches too.

pazos: I'm totally fine with the books cache. It gets redone each time:
- we scan calibre libraries
- a single calibre library has a newer timestamp than the cache

About the libraries cache I don't know. I think most users will use just one calibre library per mountpoint. If they're using calibre device drivers the metadata will be at the root of each mountpoint and the scan will finish in a few (milli)seconds.

If the user has a lot of calibre libraries under a remote path `/mnt/onboard/.books/managed/calibre/libraries` **and** a lot of files in between the scan could take a couple of minutes. The worst case scenario is on desktop, where we scan the whole home dir. Maybe it makes sense to keep libraries under settings. Pinging @NiLuJe

NiLuJe: I mean, it does say "Clear cache" and the Calibre setting says "Cache stuff", so, err, I don't see the problem ;).

NiLuJe: (i.e., I totally agree that cache is where that belongs).

defaults.lua
```diff
@@ -224,22 +224,23 @@ FRONTLIGHT_SENSITIVITY_DECREASE = 2
 -- insensitive sort
 DALPHA_SORT_CASE_INSENSITIVE = true
 
+-- no longer needed
```
NiLuJe, line 227: luacheck is probably complaining because it keeps a list of those globals to ignore, right? ;).

Frenzie: Didn't check what you mean by complaining, but if something like `SEARCH_LIBRARY_PATH` is removed here it should also be removed from .luacheckrc to make sure that it would complain.

pazos: Also `spec/unit/defaults_spec.lua` needs to be updated. I will take a look tonight.

pazos:
This PR is ready to be reviewed/tested/merged. I won't plan to work more on this unless you find something that needs to be fixed/improved. :+1:

NiLuJe:
Lemme give it a try on the Forma, but my only comment the last time I tested it was to enable the cache by default, which I think I saw you did ;).

pazos:
> Lemme give it a try on the Forma

If you try to set the wireless input box in /mnt/onboard (and you're using stock calibre kobo drivers) you should see a warning. Please cancel to dismiss. Do not press Ok unless you know what you're doing (it will strip most of the metadata from metadata.calibre. The metadata can be recovered if you  connect calibre to your Kobo again, but you were warned :smile: )

NiLuJe:
@pazos: Yup, the warning works, and should be explicit enough ;).

Other than that:

* I'd possibly reword the "Find metadata from reader" checkbox to something slightly less confusing, like "Allow searches in the reader" or "when reading", because it does just that, right, add the search entry in the reader menu? ;).

* After a search by Series, the result list used to display the series index in addition to author/title, that was useful ;).

pazos:
> After a search by Series, the result list used to display the series index in addition to author/title, that was useful ;).

Oops, I didn't notice. Something like:

```
Patrick Rothfuss: The name of the wind - the kingkiller chronicle 1
Patrick Rothfuss: The wise man's fear - the kingkiller chronicle 2
```
Maybe does more sense to put the series first (and sort them):
```
The kingkiller chronicle 1: The name of the wind - Patrick Rothfuss
The kingkiller chronicle 2: The wise man's fear - Patrick Rothfuss
```


Otherwise I would like a screenshot to figure out how that looked.

pazos:
Ok, a better solution:

```
     The kingkiller chronicle

1: The name of the wind - Patrick Rothfuss
2: The wise man's fear - Patrick Rothfuss
```

NiLuJe:
Sorry, dinner break, but, yeah, I was going to suggest using the Title field ;).

So that looks perfect (just keep in mind that series index might be a decimal ;)).

pazos:
> So, yep, that looks perfect (just keep in mind that series index might be a decimal ;)).

precision?

NiLuJe:
> > So, yep, that looks perfect (just keep in mind that series index might be a decimal ;)).
> 
> precision?

Two decimal places ought to be more than enough (that said, the FM itself appears to be rounding/truncating at the first decimal place, so, eh, that'll do, too ;)).

pazos:
Ok, changes done. We will preserve the two decimal points if *at least* one result has them. Otherwise both decimal points will be striped.





I added shortcuts to launch the search and to explore all tags/series directly.

plugins/calibre.koplugin/metadata.lua
```diff
@@ -0,0 +1,246 @@
...
+    "publisher",
+    "rating",
+    "rights",
+    "thumbnail",
+    "timestamp",
+    "title_sort",
+    "user_categories",
+    "user_metadata",
+    "_series_sort_",
+}
```
Galunid, line 35: > Please cancel to dismiss. Do not press Ok unless you know what you're doing (it will strip most of the metadata from metadata.calibre. The metadata can be recovered if you connect calibre to your Kobo again, but you were warned smile )

I assume these are the metadata that will be stripped, if that's the case, does that mean I lose covers/thumbnails?

pazos: Nope. Cover is just a path to a temporary cover made by calibre. Thumbnail is a base? encoded  image that is not used by KOReader.

All the metadata is sent by calibre during wireless transfers. We just discard most of them to have a lean json file. We do the same if we "initialize" a directory containing calibre metadata to use as the wireless inbox.

The warning is there not because we strip the metadata but because each calibre driver has its own set of rules and naming conventions, so if you set a library sent using Kindle device driver as the library used as wireless client some things won't match (like logical path to files).

On search we don't strip any metadata. We just read the whole json file and extract a cache of relevant info.

NiLuJe: Said metadata *should* be rebuilt by the Calibre driver the next time you connect over the usual method (i.e., USB), AFAIR.

NiLuJe:
@pazos: Hmm, just tried tested the latest changes.

When searching by Series, I get the initial search result, but tapping on a specific series throw me back to the FM :?.

(By tags works).

Galunid:
So I've played around with this for a bit, and it's great. One little nitpick, device doesn't disconnect when "wireless device connection"(server) is stopped on calibre's end. It works ok, when I do `Device -> Eject this device`.

plugins/calibre.koplugin/search.lua
```diff
@@ -0,0 +1,607 @@
...
+            self.libraries = libs
+        end
+    end
+
+    if #self.books == 0 then
+        self.books = self:getMetadata()
+    end
+    -- this shouldn't happen unless the user disabled all libraries or they are empty.
+    if #self.books == 0 then
+        logger.warn("no metedata to search, aborting")
```
NiLuJe, line 337: Typo ;).

NiLuJe:
@pazos: Narrowed it down to somewhere inside `bookCatalog`, it never returns on my end.

(dinner break, more on that later if you can't repro).

pazos:
> When searching by Series, I get the initial search result, but tapping on a specific series throws me back to the FM :?.

Probably because the cache doesn't have the key series_index (ie: was built before last commits). Remove the cache file and check that series_index is actually on metadata.calibre. It works here :p 

> One little nitpick, device doesn't disconnect when "wireless device connection"(server) is stopped on calibre's end. It works ok, when I do Device -> Eject this device.

Yup, you need calibre 4.18 for that. On previous versions is a known bug. See https://github.com/koreader/koreader/pull/6177#issuecomment-635457675 and next comments.

Galunid:
Actually I tested it on calibre 4.18.

Galunid:
Another thing, series are not sorted in numerical order, but in alphabetical. Perhaps adding a trailing zeroes would be a valid fix?
ss

NiLuJe:
> > When searching by Series, I get the initial search result, but tapping on a specific series throws me back to the FM :?.
> 
> Probably because the cache doesn't have the key series_index (ie: was built before last commits). Remove the cache file and check that series_index is actually on metadata.calibre. It works here :p
> 
> **edit**: or just do a scan, cache will be built again
> 
> > One little nitpick, device doesn't disconnect when "wireless device connection"(server) is stopped on calibre's end. It works ok, when I do Device -> Eject this device.
> 
> Yup, you need calibre 4.18 for that. On previous versions is a known bug. See [#6177 (comment)](https://github.com/koreader/koreader/pull/6177#issuecomment-635457675) and next comments.

Huh. Good hunch ;D. Were earlier iterations perhaps stripping stuff from metadata.calibre without warning?

In any case, I'll do another Calibre over USB session and check again, thanks ;).

(Building the cache seemed strangely speedy on the Forma, that may be why ^^).

NiLuJe:
> Another thing, series are not sorted in numerical order, but in alphabetical. Perhaps adding a trailing zeroes would be a valid fix?

:+1:

Although you probably meant *leading* zero ;).

NiLuJe:
@pazos: Okay, metadata.calibre is now sane, regenerated the cache and... every series_index is set to a function in the cache o_O.

NiLuJe:
@pazos: Okay, nope, metadata.calibre is still incomplete.

Looking at `jq '.[].series_index' < /mnt/onboard/metadata.calibre`, I can see that mostly everything has series_index set to `null`, which is where those functions are coming from, I guess?

(Sidebar: shouldn't those be nil insetad? ^^).

Frenzie:
@NiLuJe @pazos Do you want it in this release or rather the one after?

NiLuJe:
@Frenzie: It'd be pretty nice in this one, I guess? ;).

NiLuJe:
Okay, so, that was a *NO* on calibre automatically regenerating discarded fields from the metadata.calibre file ;D.

Killed it with fire, it jumped from 612K to 5.1MB, and is now actually complete and up to date ^^.

pazos:
Uh, oh. All those issues are related with the series_index added recently. I was testing it with books *with* series_index. I forgot they're nullable values. 

> Were earlier iterations perhaps stripping stuff from metadata.calibre without warning?

Nope. Metadata is only removed from metadata.calibre on wireless connection, not on searches.

> Okay, so, that was a NO on calibre automatically regenerating discarded fields from the metadata.calibre file ;D

No automatically, but on `device -> update metadata in device cache` should do the job. It does here with both folder and Bq driver.

NiLuJe:
@pazos: Got it.

This took care of it (once I had a sane metadata.calibre):

```diff
diff --git a/plugins/calibre.koplugin/search.lua b/plugins/calibre.koplugin/search.lua
index 00a48066..9d0b9105 100644
--- a/plugins/calibre.koplugin/search.lua
+++ b/plugins/calibre.koplugin/search.lua
@@ -110,7 +110,7 @@ end
 local function getBooksBySeries(t, series)
     local result = {}
     for _, book in ipairs(t) do
-        if book.series and type(book.series) ~= "function" then
+        if book.series then
             if book.series == series then
                 table.insert(result, book)
             end
@@ -136,7 +136,7 @@ end
 local function searchBySeries(t, query, case_insensitive)
     local freq = {}
     for _, book in ipairs(t) do
-        if book.series and type(book.series) ~= "function" then
+        if book.series then
             if match(book.series, query, case_insensitive) then
                 freq[book.series] = (freq[book.series] or 0) + 1
             end
@@ -171,7 +171,7 @@ local function getBookInfo(book)
         tags = _("Tags:") .. " " .. tags
     end
     local series
-    if book.series and type(book.series) ~= "function" then
+    if book.series then
         series = _("Series:") .. " " .. book.series
     end
     return string.format("%s\n%s\n%s%s%s", title, authors,
@@ -288,7 +288,7 @@ function CalibreSearch:bookCatalog(t, option)
         local entry = {}
         entry.info = getBookInfo(book)
         entry.path = book.rootpath .. "/" .. book.lpath
-        if series then
+        if series and book.series_index then
             local major, minor = string.format("%.2f", book.series_index):match("([^.]+).([^.]+)")
             if minor ~= "00" then
                 subseries = true
@@ -592,9 +592,13 @@ function CalibreSearch:getMetadata()
         local dump = {}
         for index, book in ipairs(books) do
             local dumped_book = book
+            -- Convert rapidjson.null to nil
             if type(dumped_book.series) == "function" then
                 dumped_book.series = nil
             end
+            if type(dumped_book.series_index) == "function" then
+                dumped_book.series_index = nil
+            end
             table.insert(dump, index, dumped_book)
         end
         util.dumpTable(dump, self.user_book_cache)
```

pazos:
> This took care of it (once I had a sane metadata.calibre ^^):

It works from cache but crashes if cache is disabled and we're loading things directly from a json file. 
Also if series is not null then series_index is also not null

NiLuJe:
> > This took care of it (once I had a sane metadata.calibre ^^):
> 
> It works from cache but crashes if cache is disabled and we're loading things directly from a json file.
> Also if series is not null then series_index is also not null

Ah, true, that explains why those pesky typechecks were kept ^^.

Although, with my broke-ass metadata.calibre file, I *definitely* had null series_index with a series set.

NiLuJe:
As for the leading zero thing, this shoudl take care of it:

`local major, minor = string.format("%05.2f", book.series_index):match("([^.]+).([^.]+)")`

Spoiler alert: this *may* be a tad undefined behavior (and/or printf's manpage sucks). Turns out the field length applies to the *full* decimal, not just one side of the dot. The precision field still only applies to the decimal part (and will round).

NiLuJe:
@pazos: Oh. Found out how I got a stripped metadata.calibre ;).

Turns out attempting to set the folder as an inbox strips the metadata *before* the "Warning" popup ^^.

Galunid:
> As for the leading zero thing, this should take care of it:
> 
> `local major, minor = string.format("%05.2f", book.series_index):match("([^.]+).([^.]+)")`
> 
> Spoiler alert: this _may_ be a tad undefined behavior (and/or printf's manpage sucks). Turns out the field length applies to the _full_ decimal, not just one side of the dot (so, two decimal points (i.e., our precision), the decimal separator, and a two character integer makes 5). The precision field still only applies to the decimal part (and will round).

I just tested it, it works as it should I think. When there is no decimal, we get DD format, when there is decimal we get DD.DD format.

Another thing is what happens if series has >15 books(for example), is there paging/scrolling implemented in the widget?

NiLuJe:
@Galunid: Yup!

disc

pazos:
> Turns out attempting to set the folder as an inbox strips the metadata before the "Warning" popup ^^.

Oh dear, oh dear, oh dear.

shermp:
Hey, just thought I'd provide a quick heads-up in case you do, or ever plan on using any of the 'date' fields from Calibre.

You CANNOT assume that Calibre will send a valid timestamp, or null. I've seen it send the string "none" as well.

This has bitten me more than a few times (I keep forgetting about it every time I go to refactor UNCaGED...)

It's probably not such a big issue for a dynamically typed language like lua, but is a huge PITA in Go.

Frenzie:
Not a huge issue, but the *string* `none` would certainly be something to be aware of.

shermp:
> Not a huge issue, but the _string_ `none` would certainly be something to be aware of.

Yep. Trying to unmarshal that into a time field in a struct doesn't work very well...

Galunid:
So, I've managed to crash calibre! This happens if you disconnect from wifi, while still connected to server and then try to eject device from calibre. Not a big deal, but can be problematic if Kobo goes to sleep while being connected to calibre. I'm not sure if it's fixable, but it's probably a good idea to be aware of it, as it can possibly lead to Calibre library corruption.

NiLuJe:
@pazos: Okay, FWIW, I indeed can't reproduce having series set but series_index null with the current stripping code ;).

(And it definitely shouldn't be able to happen with full Calibre metadata, either :)).

So, on my end, that leaves:

```diff
diff --git a/plugins/calibre.koplugin/main.lua b/plugins/calibre.koplugin/main.lua
index f462ba58..a4badcbf 100644
--- a/plugins/calibre.koplugin/main.lua
+++ b/plugins/calibre.koplugin/main.lua
@@ -131,7 +131,7 @@ function Calibre:getSearchMenuTable()
             end,
         },
         {
-            text = _("Find metadata from reader"),
+            text = _("Enable searches in the reader"),
             checked_func = function()
                 return G_reader_settings:isTrue("calibre_search_from_reader")
             end,
diff --git a/plugins/calibre.koplugin/search.lua b/plugins/calibre.koplugin/search.lua
index 1500cce7..86db4860 100644
--- a/plugins/calibre.koplugin/search.lua
+++ b/plugins/calibre.koplugin/search.lua
@@ -334,7 +334,7 @@ function CalibreSearch:find(option)
     end
     -- this shouldn't happen unless the user disabled all libraries or they are empty.
     if #self.books == 0 then
-        logger.warn("no metedata to search, aborting")
+        logger.warn("no metadata to search, aborting")
         self:prompt(_("No metadata found"))
         return
     end
```

:+1:

NiLuJe:
FWIW, Calibre's "Update cached metadata on device" only rebuilds the structure, AFAICT (e.g., comments fields are back, but they're all null). ;).

NiLuJe:
And, another random Calibre issue: the timestamp on my metadata.calibre is wonky as hell:

`Modify: 2020-06-15 02:16:26.000000000 +0200`

`Mon Jun 15 00:27:02 CEST 2020`

And, you've guessed it... CEST is GMT+2, so, TZ shenanigans.

Possibly not entirely Calibre's fault, or vfat being weird or something.

pazos:
Probably is better to get the timestamp from driveinfo.calibre instead. I will give it a look.

@Frenzie: I think we can merge this one after 2020.06, so it will be fully translated for 2020.07 and hopefully battle tested.

NiLuJe:
Anyway, the idea was to give you some performance numbers, so ;).

Cold VFS cache:
```
06/15/20-00:23:00 INFO  calibre info loaded from disk in 508.933067 milliseconds: 618 books. 0 pruned
06/15/20-00:23:00 INFO  metadata: 618 books imported from calibre in 687.426090 milliseconds
06/15/20-00:23:00 INFO  search done in 83.262920 milliseconds (series, case sensitive: false, title: true, authors: true, path: false)
06/15/20-00:23:38 INFO  calibre info loaded from disk in 408.795118 milliseconds: 618 books. 0 pruned
06/15/20-00:23:00 INFO  search done in 83.262920 milliseconds (series, case sensitive: false, title: true, authors: true, path: false)

```

Hot VFS cache:

```
06/15/20-00:24:31 WARN  cache is older than metadata, ignoring it
06/15/20-00:24:32 INFO  calibre info loaded from disk in 477.095127 milliseconds: 618 books. 0 pruned
06/15/20-00:24:32 INFO  metadata: 618 books imported from calibre in 596.668959 milliseconds
06/15/20-00:24:32 INFO  search done in 16.270876 milliseconds (series, case sensitive: false, title: true, authors: true, path: false)
```

Cached (so, assuming hot, too):

```
06/15/20-00:34:20 INFO  metadata: 618 books imported from cache in 39.472818 milliseconds
06/15/20-00:34:20 INFO  search done in 20.455837 milliseconds (series, case sensitive: false, title: true, authors: true, path: false)
```

That's *dramatically* faster than before ;).
:+1:

pazos:
> Hey, just thought I'd provide a quick heads-up in case you do, or ever plan on using any of the 'date' fields from Calibre.

Thanks for the info @shermp. And huge thanks for UNCaGED. It helped me a lot to figure out NOOP's behaviour and the password/challenge thing.

> You CANNOT assume that Calibre will send a valid timestamp, or null. I've seen it send the string "none" as well.

I think you're talking about book timestamps. We skip these. We just care about driveinfo.calibre "date_last_connected", which is, AFAICT, always populated by calibre.

BTW: one of the things left unimplemented is sending books back to calibre wirelessly. I'm aware UNCaGED has that but never looked how it is implemented. 
Do you scan a certain directory for books that are not in the DB and add them to it (generate an uuid, extracting metadata from files, ...) ?. If so I would like to know how fast the whole thing is. It could be interesting having local files generated by KOReader, like epubs from wiki, in the calibre DB and ready to send to calibre.

pazos:
> So, I've managed to crash calibre! This happens if you disconnect from wifi, while still connected to server and then try to eject device from calibre. Not a big deal, but can be problematic if Kobo goes to sleep while being connected to calibre. I'm not sure if it's fixable, but it's probably a good idea to be aware of it, as it can possibly lead to Calibre library corruption.

We can send an event before suspending from UIManager and consume it on the plugin, I think!.

shermp:
> I think you're talking about book timestamps. We skip these. We just care about driveinfo.calibre "date_last_connected", which is, AFAICT, always populated by calibre.

Yeah, I was talking about timestamps in the metadata. I THINK the smart_device driver sends metadata as-is, without checking it for any constancy.

BTW: one of the things left unimplemented is sending books back to calibre wirelessly. I'm aware UNCaGED has that but never looked how it is implemented.
Do you scan a certain directory for books that are not in the DB and add them to it (generate an uuid, extracting metadata from files, ...) ?. If so I would like to know how fast the whole thing is. It could be interesting having local files generated by KOReader, like epubs from wiki, in the calibre DB and ready to send to calibre.

In Kobo-UNCaGED, part of the init process is getting a list of all books on device. If it is in the nickel DB, I think I create a new UUID for the book as part of that process. (I used to read metadata from epub/kepub files, but ran into issues involving XML parsing in go, so I currently just use whatever metadata is in the sqlite database)

pazos:
@Galunid: I tried to reproduce the calibre crash in https://github.com/koreader/koreader/pull/6177#issuecomment-643828248 but I'm not able to do that.

I think you talk about an error handled by calibre (ie: a raised exception like Control Exception) but these exceptions never kill calibre. If you happen to break calibre to the point the GUI is completely gone I would like you to reproduce while running calibre from the shell using `calibre-debug -g` and attach here the log.

If is just an error that calibre shows to the user we can ignore it. These are harmless and can be reproduced in many ways, being the more obvious to send a bunch of books from calibre and disconnect the client while it is receiving them.

Galunid:
@pazos 
**steps to replicate:**
1. Connect device to calibre's server
2. Disable network on device(Kobo clara HD in my case)
3. Select `Device -> Eject device` in calibre gui.

**What happens:**
My bad,  the gui was completely gone for a while, but it started responding after some time, although it probably took 2-4 minutes. I'm using i3-gaps configuration, but I don't think it matters.

`uname -a`:
```
Linux kris-pl 5.7.2-arch1-1 #1 SMP PREEMPT Wed, 10 Jun 2020 20:36:24 +0000 x86_64 GNU/Linux
```

`calibre --version`:
```
calibre (calibre 4.18)
```

`calibre-debug -g`
```
SMART_DEV (  19.33:  0.000) _call_client FREE_SPACE wfr True
DeviceJob: 3 Send metadata to device done, calling callback
DeviceJob: 3 Send metadata to device callback returned
SMART_DEV (  20.37:  1.034) _call_client NOOP wfr True
Job: 3 Send metadata to device finished
No details available.
SMART_DEV (  22.41:  2.044) _call_client NOOP wfr True
SMART_DEV (  24.51:  2.098) _call_client NOOP wfr True
SMART_DEV ( 324.61:300.099) _write_metadata_cache
SMART_DEV ( 324.71:  0.095) _write_metadata_cache wrote 148 entries, purged 0 entries
SMART_DEV ( 324.71:  0.000) _receive_from_client timeout communicating with device
SMART_DEV ( 324.71:  0.000) _call_client other exception
Traceback (most recent call last):
  File "/usr/lib/calibre/calibre/devices/smart_device_app/driver.py", line 643, in _call_client
    return self._receive_from_client(print_debug_info=print_debug_info)
  File "/usr/lib/calibre/calibre/devices/smart_device_app/driver.py", line 675, in _receive_from_client
    raise TimeoutError('Device did not respond in reasonable time')
TimeoutError: There was a timeout while communicating with the device in function: Device did not respond in reasonable time
SMART_DEV ( 324.71:  0.000) eject
SMART_DEV ( 324.72:  0.001) _call_client NOOP wfr True
SMART_DEV ( 324.72:  0.001) post_yank_cleanup
```

pazos:
Thanks for the info @Galunid. That error could be annoying but is harmless.

Probably if you wait more than 10seconds between step 2 and step 3 you'll get an error that doesn't block calibre all that time:

Calibre sends keep-alive messages each few seconds. If the device doesn't reply to one of them it wil throw an error, still harmless. Since you disconnect the wifi after one of these keep-alive messages but eject the drive before the next keep-alive message calibre will ask the client to eject and will wait forever the reply. The reply doesn't happen and calibre will wait until timeout, thus blocking the UI.

pazos:
@NiLuJe: I saw your comment in https://www.mobileread.com/forums/showpost.php?p=4001400&postcount=2. Sadly mark a book as read is not part of stock calibre metadata. I guess is part of the `user_metadata` field but I'm not sure since I don't use it (or any other custom column). If your library has a book read Y/N checkmark - or any other custom column - then I think I could use some samples to figure out how to deal with read/unread books.

NiLuJe:
@pazos: I was kinda thinking of maintaining a dynamic tag manually, but, you're right that the user_data stuff makes it to the metadata dump...

IIRC, mine is pretty much named '#read', but I'll get you an actual example later (because, err, mine's a bit of a mess, the actual read column is generated based on both a "progress on Kindle" and "progress on Kobo" column ^^).

plugins/calibre.koplugin/metadata.lua
```diff
@@ -0,0 +1,250 @@
...
+    Calibre uses JSON to store metadata on device after each wired transfer.
+    In wireless transfers calibre sends the same metadata to the client, which is in charge
+    of storing it.
+--]]
+
+local rapidjson = require("rapidjson")
+local logger = require("logger")
+local util = require("util")
+
+local unused_metadata = {
```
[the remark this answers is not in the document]
NiLuJe: On the other hand, `user_metadata` is hands down going to be the *largest* (and most complex) structure in there, so that's potentially a strike against it... ;).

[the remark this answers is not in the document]
pazos: Fair warning. I think I will leave it for somebody that uses the feature :smiley:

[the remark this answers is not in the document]
NiLuJe: Yeah, to give you an example....

```json
"user_metadata": {
      "#hyphenated": {
        "is_category": false,
        "#extra#": null,
        "kind": "field",
        "is_custom": true,
        "is_csp": false,
        "colnum": 3,
        "column": "value",
        "rec_index": 22,
        "search_terms": [
          "#hyphenated"
        ],
        "link_column": "value",
        "is_multiple2": {},
        "is_multiple": null,
        "datatype": "bool",
        "#value#": null,
        "category_sort": "value",
        "table": "custom_column_3",
        "is_editable": true,
        "label": "hyphenated",
        "display": {},
        "name": "Hyphenated"
      },
      "#koboreadpct": {
        "is_category": false,
        "#extra#": null,
        "kind": "field",
        "is_custom": true,
        "is_csp": false,
        "colnum": 5,
        "column": "value",
        "rec_index": 26,
        "search_terms": [
          "#koboreadpct"
        ],
        "link_column": "value",
        "is_multiple2": {},
        "is_multiple": null,
        "datatype": "int",
        "#value#": 100,
        "category_sort": "value",
        "table": "custom_column_5",
        "is_editable": true,
        "label": "koboreadpct",
        "display": {
          "description": "Current Reading status on Kobo",
          "number_format": null
        },
        "name": "Progress"
      },
      "#read": {
        "is_category": false,
        "#extra#": null,
        "kind": "field",
        "is_custom": true,
        "is_csp": false,
        "colnum": 7,
        "column": "value",
        "rec_index": 28,
        "search_terms": [
          "#read"
        ],
        "link_column": "value",
        "is_multiple2": {},
        "is_multiple": null,
        "datatype": "composite",
        "#value#": "Yes",
        "category_sort": "value",
        "table": "custom_column_7",
        "is_editable": true,
        "label": "read",
        "display": {
          "description": "Read status",
          "composite_template": "program:\n#\tIf it's Read (i.e. at 100%) on Kobo, flag it as read, otherwise, use the Kindle status as-is.\n\ttest(field('#koboreadpct'), cmp(field('#koboreadpct'), 100, '', 'Yes', ''), field('#kindleread'));",
          "composite_sort": "bool",
          "contains_html": false,
          "make_category": false,
          "use_decorations": 2                                                                                                                                                                                                                                                                                                                                           
        },
        "name": "Read"
      },
      "#kindleread": {
        "is_category": false,
        "#extra#": null,
        "kind": "field",
        "is_custom": true,
        "is_csp": false,
        "colnum": 1,
        "column": "value",
        "rec_index": 23,
        "search_terms": [
          "#kindleread"
        ],
        "link_column": "value",
        "is_multiple2": {},
        "is_multiple": null,
        "datatype": "bool",
        "#value#": null,
        "category_sort": "value",
        "table": "custom_column_1",
        "is_editable": true,
        "label": "kindleread",
        "display": {
          "description": "Read status on Kindle"
        },
        "name": "Read on Kindle"
      },
      "#kobolastread": {
        "is_category": false,
        "#extra#": null,
        "kind": "field",
        "is_custom": true,
        "is_csp": false,
        "colnum": 8,
        "column": "value",
        "rec_index": 24,
        "search_terms": [
          "#kobolastread"
        ],
        "link_column": "value",
        "is_multiple2": {},
        "is_multiple": null,
        "datatype": "datetime",
        "#value#": "2020-06-03T06:08:34+00:00",
        "category_sort": "value",
        "table": "custom_column_8",
        "is_editable": true,
        "label": "kobolastread",
        "display": {
          "date_format": null,
          "description": "Last time the book was Read on Kobo"
        },
        "name": "Last Read"
      },
      "#koboreadloc": {
        "is_category": true,
        "#extra#": null,
        "kind": "field",
        "is_custom": true,
        "is_csp": false,
        "colnum": 6,
        "column": "value",
        "rec_index": 25,
        "search_terms": [
          "#koboreadloc"
        ],
        "link_column": "value",
        "is_multiple2": {},
        "is_multiple": null,
        "datatype": "text",
        "#value#": null,
        "category_sort": "value",
        "table": "custom_column_6",
        "is_editable": true,
        "label": "koboreadloc",
        "display": {
          "description": "Current Reading Location on Kobo",
          "use_decorations": 0                                                                                                                                                                                                                                                                                                                                           
        },
        "name": "Reading Location"
      },
      "#pages": {
        "is_category": false,
        "#extra#": null,
        "kind": "field",
        "is_custom": true,
        "is_csp": false,
        "colnum": 2,
        "column": "value",
        "rec_index": 27,
        "search_terms": [
          "#pages"
        ],
        "link_column": "value",
        "is_multiple2": {},
        "is_multiple": null,
        "datatype": "int",
        "#value#": 288,
        "category_sort": "value",
        "table": "custom_column_2",
        "is_editable": true,
        "label": "pages",
        "display": {
          "number_format": "{0:,}"
        },
        "name": "Pages"
      }
    },
```

:oO:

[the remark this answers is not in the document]
NiLuJe: Granted, ideally, one would not deal with my insane composite column, and instead only something like #kindleread (which I maintained manually), or #koboreadpct (which is automatically kept in sync w/ Nickel via the Kobo driver).

But, still, potentially neat, but probably a huge can of worms ^^.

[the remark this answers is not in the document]
NiLuJe: And that's still considered a fairly tame use of custom columns ^^.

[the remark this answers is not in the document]
pazos: Wow. I'm starting to understand what a calibre user looks like.

I just use it to keep purchased ebooks (deDRMed). I had to download a bunch of Gutenberg books to test this PR since my purchases are not so many. (~ 200 books).

PR #6179, opened 2020-05-23
Kill the RMSDK, too.

NiLuJe:
It's a big pile 'o RAM we could use.
Plus, it might be causing bad interactions with DRM on Nickel restart if we keep it around.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6179)

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -110,7 +110,7 @@ if [ "${VIA_NICKEL}" = "true" ]; then
     sync
     # And we can now stop the full Kobo software stack
     # NOTE: We don't need to kill KFMon, it's smart enough not to allow running anything else while we're up
-    killall -TERM nickel hindenburg sickel fickel fmon 2>/dev/null
+    killall -TERM nickel hindenburg sickel fickel adobehost fmon 2>/dev/null
 fi
 
 # fallback for old fmon, KFMon and advboot users (-> if no args were passed to the script, start the FM)


NiLuJe:
Skipped my mind until then because I don't actually read straight ePubs because the performance is just not up to par (and used to be downright abysmal).

ptrm:
Big pile meaning ~10MB or ~100MB ;) ?

Frenzie:
@ptrm I dunno, all my EPUBs are DRM-free (e.g., Gutenberg, DBNL, Humble Bundle).

ptrm:
From the changes I assumed it just load with nickel on startup. And from experience I know you can't even get metadata from drmed book without decryption ;)

Frenzie:
There's no `adobe*` process if I just start Nickel.

ptrm:
Yeah, even after opening a book I bought on Kobo store and launching koreader there's no sign of `adobe*` processes :0 @NiLuJe when dies this change apply, then?

Frenzie:
I thought I'd seen more info somewhere… and I had: https://www.mobileread.com/forums/showthread.php?p=3991983#post3991983
> Huh. A quick highly scientific (^^) test based on a single DRM-free ePub (as I mentioned, I don't read ePubs, so it's my only goto test file, and my Kobo is not linked to ADE ;p) revealed that I couldn't actually reproduce that behavior (on an H2O).
> 
> In any case, killing the RMSDK is a sound decision, and it may actually help with this, so: https://github.com/koreader/koreader/pull/6179
> 
> (Plus, 75MB of RAM, wheee!).

ptrm:
Thanks 😁


 A peeked into Accounts section of nickel and now it makes sense to me :)

Frenzie:
Is Adobe Digital Editions that absolutely atrocious app? My wife used it for some book last month I believe.

ptrm:
Close to atrocious. I use the Adobe AIR based 3.x version in a process enabling me to read those books on koreader :3 There's a 4.x version, too however, but judging from the screenshots the UX is about the same.

ptrm:
Lucky fact being all Polish made ebooks are drm-free, just watermarked.

NiLuJe:
Yeah, the RMSDK is launched on-demand by Nickel on the first ePub opened (DRM or not. My Settings page looks just like yours, @ptrm ;)).

ptrm:
I see, so it might just mean rmsdk gracefully exits along with nickel?

Frenzie:
@ptrm She used it on Windows.

NiLuJe:
@ptrm: It doesn't, hence the kill ;).

ptrm:
My, the edit in my last post was to be a reply :/

As for nickel, maybe rmsdk it just monitors a pid to quit with like some systemd services? I just wonder why I don't see it after opening books on nickel, despite having a version from 18th May ;)

ptrm:
All right then, so kobo epubs are the ones that don't trigger adobe DE start, after opening an epub from outside kobo store the adobehost is there as supposed.

NiLuJe:
Yep, reading KePubs will never spawn it, which makes sense, Nickel is like us in that respect, it's not terribly fond of dealing with a 75MB foreign entity if it doesn't have to ;).

ptrm:
@NiLuJe lol, AFAIR I only launched search and opened one book, so no idea either :D

poire-z:
Off-topic, but not worth an issue.
You who kill (and so know about :) the Kobo processes:
- what is this "hindenburg" one ? (I always wondered what it does, and why it is named like that :)
- would it be possible to launch/use the kobo web browser from KOReader ?

NiLuJe:
I was going to say hindenburg was handling the watchdog, but I may be crossing wires here, because there's sickel, whose job is to murder Nickel when it hangs. Maybe e a hardware watchdog -> hindenburg vs. software watchodg -> sickel?

Possibly named so because it watches over the system in case stuff comes crashing down?

----

Nope, that's a Nickel plugin.

Frenzie:
It's always reminded me of https://www.youtube.com/watch?v=pUVDmXvXcbk

Issue #4431, opened 2018-12-30
Short taps on wakeup are interpreted as long taps (kobo glo hd N437)


worldexplorer:
* KOReader version: v2018.12-31-gf6743a4_2018_12_28
* Device: Kobo Glo HD, model N437

#### Issue
after I wake up my kobo (that has been put to sleep with short press on physical button on top), any tap on the screen is considered as a long tap. I tap to turn the page and a dictionary window pops up with highlighted word where I tapped; closing this popup window sometimes succeeds, sometimes it doesn't; I can sometimes move this popup window with the dictionary, but how koreader interprets my taps (single / double / swipe / drag and drop) is completely messed; the real long tap is also considered as a long tap, and doing long tap to kobo doesn't make short taps work;

what doesn't help:
1) having a screen cleaned up via Screensaver > Use Message as screensaver; I was hoping that re-drawing the page on wake up will unfreeze the tap sensor;
2) I'm not sure about the book format displayed; I think PDFs are not suffering from this issue;
3) putting kobo to sleep, and waking it up while holding finger on the screen without motion; having two fingers tap without motion while kobo wakes up and cleans the screen by redrawing white to black and back;
4) upgrading KSM (I used kobo start menu 08 for 1.5 years, and 2 weeks ago I upgraded to kobo start menu 09);
5) none of the koreader updates during 2018 addressed this issue - I didn't try every dev version, but "just waiting for the fix" didn't help :) 

what helps:
1) putting kobo to sleep, and waking it up while moving finger on the screen;
2) of course, full restart (twice, because first time kobo boots into nickel, after a forced long press; second time it boots to Kobo Start Menu 9 after second forced long press);

suggestion: koreader needs to initialize/reset the digitizer somehow (via kernel driver call?); it definitely has something to do with hardware, sensors, debouncing (as a hardware phenomenon) and kobo specifics; may be nickel has some weird code executed on wakeup that is visible under disassembler;

extra observations:
1) nickel (a factory rakuten reader software) never has this issue; while in nickel, putting to sleep by short pressing device's button on top and waking it up by the second press - always makes taps correct (never my taps to turn page are "seen" by nickel as long taps);
2) plato doesn't seem to have this issue, but I didn't use it too much yet (inconvenient folder access, I can't find a book I put into 3/4-level folder structure; with hundreds of books I need a listing and an explorer-like interface, instead of tags-like);
3) KSM never had this misinterpretation of taps (probably because it always starts from "cold hardware" reset state);

[...]

PR #6186, opened 2020-05-25
[fix] GestureDetector: prevent hold timers overlap

poire-z:
Prevent the Tap>Hold setTimeout'ed function to trigger on another later Tap.
Fix a Tap quickly following a Swipe (which triggers this Tap>Hold timer) from becoming a Hold.

I would often get that behaviour on my Kobo...
It's a bit harder to trigger on the emulator with the default gesture interval settings, but it become easier if you set Hold interval to 1000ms:
- do a few long-press on a word to get a feeling of how long this takes
- do a diagonal swipe (to full screen refresh) and then quickly a tap/short long-press : the hold will trigger quicker than previous feeling

I had hope that somehow this could magically fix the issue I have once a week of touch becoming unresponsive (https://github.com/koreader/koreader/issues/4431#issuecomment-455847181) - where my next tap after power button press often becomes a Hold - but no, still got it this morning in spite of this fix :(

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6186)

### DIFF
--- a/frontend/device/gesturedetector.lua
+++ b/frontend/device/gesturedetector.lua
@@ -76,6 +76,7 @@ local GestureDetector = {
     },
     -- states are stored in separated slots
     states = {},
+    hold_timer_id = {},
     track_ids = {},
     tev_stacks = {},
     -- latest feeded touch event in each slots
@@ -234,6 +235,7 @@ end
 
 function GestureDetector:clearState(slot)
     self.states[slot] = self.initialState
+    self.hold_timer_id[slot] = nil
     self.detectings[slot] = false
     self.first_tevs[slot] = nil
     self.last_tevs[slot] = nil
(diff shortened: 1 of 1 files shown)


frontend/device/gesturedetector.lua
```diff
@@ -367,7 +369,7 @@ function GestureDetector:handleDoubleTap(tev)
     }
 
     if not self.input.disable_double_tap and self.last_taps[slot] ~= nil and
-    self:isDoubleTap(self.last_taps[slot], cur_tap) then
+                self:isDoubleTap(self.last_taps[slot], cur_tap) then
```
NiLuJe, line 372: Stray indent?

NiLuJe: Gah, nevermind, it's a multiline test.

PR #6212, opened 2020-05-30
Footer: fix items possibly missing

### LINKED ISSUE #6210: Debian/AppImage packages don't show all items on status bar
* KOReader version: 2020-05
* Device: linux emulator running on Ubuntu 20.04

#### Issue

I have the status bar set to "show all items at once". Some items selected (e.g., percentage read) do not show.

Changing things like alignment, position of progress bar, toggling the status bar items, sorting them, etc do not change the result

#### Steps to reproduce

Turn on Show all items
Turn on Progress Percentage

See attached pictures. I expect to see Current page, Current time, Pages left in Chapter, and Percentage read; but I don't see Percentage read. See attached screen shots.
Screenshot from 2020-05-29 09-24-39
Screenshot from 2020-05-29 09-29-25

poire-z:
Iterating a k/v table as an array would have us stopped on the first disabled MODE (frontlight, battery... on some devices). Details at https://github.com/koreader/koreader/issues/6210#issuecomment-636132193. Closes #6210.
Rework that initial setup, and make it correctly handle added or removed MODEs when using the ordered items saved in the user settings.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6212)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -326,10 +326,7 @@ function ReaderFooter:init()
         text_font_bold = false,
     }
 
-    local mode_tbl = {}
-    local mode_name
-    self.mode_nb = 0
-    self.mode_index = {}
+    -- Remove items not supported by the current device
     if not Device:isAndroid() then
         MODE.wifi_status = nil
     end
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -339,23 +336,40 @@ function ReaderFooter:init()
...
+    self.mode_nb = 0
+
     if self.settings.order then
-        while #self.settings.order < #self.mode_index do
-            self.settings.order[#self.settings.order + 1] = self.mode_index[#self.settings.order + 1]
+        -- Start filling self.mode_index from what's been ordered by the user and saved
+        for i=0, #self.settings.order do
+            local name = self.settings.order[i]
+            -- (if name has been removed from our supported MODEs: ignore it)
+            if MODE[name] then -- this mode still exists
```
Frenzie, line 351: Is that MODE thing a global?

poire-z: Kind of: a module local variable:
https://github.com/koreader/koreader/blob/03140f505ca475f185bdc197b175ee8a59cf774f/frontend/apps/reader/modules/readerfooter.lua#L24-L42

poire-z:
Oups, didn't noticed until now that with this, after switching documents, we get:

<kbd>image</kbd>

Probably because I hack the global MODE, which is then empty for the next document...

poire-z:
That one is fixed - but I'm not certain we won't see other bugs in use cases I didn't think of...
We'll see :)

JasonInOttawa:
I dropped the edited file into my install and it worked. Thanks!

Is fixing the icons (e.g., Book time to read that show up as question marks outside the scope of this PR?

Screenshot from 2020-05-30 14-26-04
)

poire-z:
Thanks for the feedback :)

> Is fixing the icons outside the scope of this PR?

Yes, I don't think that's a code issue (that I could solve), more probably a debian packaging issue (that I don't know much about) https://github.com/koreader/koreader/issues/6210#issuecomment-636112725 - or something on your side?
Best to keep the discussion about it in #6210.

PR #6218, opened 2020-05-31
fix missing glyphs by packaging freefonts again.

pazos:
Also disable battery completely from defaults/option in desktop linux as it always return 0%

Background in #6210

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6218)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -308,7 +308,7 @@ function ReaderFooter:init()
         all_at_once = false,
         reclaim_height = false,
         toc_markers = true,
-        battery = true,
+        battery = not Device:isDesktop(),
         time = true,
         page_progress = true,
         pages_left = true,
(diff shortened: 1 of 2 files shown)


NiLuJe:
Ah, yeah, we're shipping freefont svn, the latest release is hilariously old ;).

pazos:
lintian outputs
```
W: koreader: opentype-font-wrong-filename usr/lib/koreader/fonts/freefont/FreeSans.ttf
W: koreader: opentype-font-wrong-filename usr/lib/koreader/fonts/freefont/FreeSerif.ttf
```

are they otf?

NiLuJe:
Yeah, they're OTF renamed as TTF to avoid having to change seven billion references, most notably in the mupdf patches.

poire-z:
(Oh, allright, so that was the icons from FreeSerif that were missing, and not the ones from nerdfonts like I too quickly assumed :/)

JasonInOttawa:
I built and installed a deb and the status bar looks great:

Screenshot from 2020-06-01 19-56-30

PR #6225, opened 2020-06-03
Footer: skip tap lock when needed

poire-z:
7d139263 / #6083 was preventing some features to work when lock status bar is enabled, so skip it in some cases:
- When toggling between top & bottom status bar (a tap is simulated to show the bottom bar).
- When in flipping mode (jumps to page at tap position).
- When toggling via menu (shown only when zero tap zone).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6225)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -723,7 +723,7 @@ function ReaderFooter:addToMainMenu(menu_items)
             enabled_func = function()
                 return not self.view.flipping_visible
             end,
-            callback = function() self:onTapFooter() end,
+            callback = function() self:onTapFooter(true) end,
         })
     end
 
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1809,40 +1809,46 @@ function ReaderFooter:onExitFlippingMode()
...
-                if self.mode == i then
-                    if self.settings[m] then
-                        break
-                    else
-                        self.mode = (self.mode + 1) % self.mode_nb
-                    end
+            self.mode = self.mode_list.page_progress
+        end
+    else
+        self.mode = (self.mode + 1) % self.mode_nb
```
Frenzie, line 1838: That's the remainder of available modes?

poire-z: (Original code not changed, it just appears in the diff because of my change in if/else indentation.)
I guess it's just a way to cycle thru all modes, and when reaching end of list, go back to start - which is not `1` because we have a mode 0 = off :)
(It' s a bit of magic mixed with mess :) See https://github.com/koreader/koreader/issues/6210#issuecomment-636150143. )

Issue #6185, opened 2020-05-25
[FR] News Downloader - html filtering


georgew21:
* KOReader version: 2020.5
* Device: PW3

#### Feature Request

Hello dear developers,

I find that News Downloader is inconvient. In many feeds, I need to turn a lot pages to find the text of article [on the other side, if i set to downlaod full aricle to false, i have only the description and not the full article). Also, it's annoying going back and after waiting to open each new artictle seperatly.

I would like to suggest 2 changes.

1. First, the option to download only specific tags of the wepbage (for example, div "body", mainarticle etc) - with the possibility of user to add more tags - so, if the plugin  doesn't automatically get full article, the user can add some tags).

2.  Saving all the articles of a feed in a single html file with contents table in the begining and hyper-links to each article. 

Have a great day!

Frenzie:
@NiLuJe Better to discuss that proxy thing here. ;-)

#6205 mentioned https://flak.tedunangst.com/post/miniwebproxy which looks like an interesting alternative to Proxomitron and its modern semi-clone Privoxy. Definitely an interesting program.

lich-tex:
I think trying to put Go into KOReader is not really the optimal solution. There is [LuaXML](https://github.com/michal-h21/LuaXML) and it seems it would fit our use-case perfectly. It is just a matter of implementing it, and I am not that good with Lua.

Frenzie:
I didn't mean "interesting" in the "ship it" sense, quite the opposite. Apologies for any confusion. It was more thinking out loud in the "once upon a time I used Proxomitron" sense.

> EDIT: It seems LuaXML is already implemented, so isn't it a matter of just creating a CSS element filter list?

Probably, `query_selector` would be the most obvious thing to look into if you want to implement something like that. (You'll need a much newer version of LuaXML for parseQuery & friends of course.)

lich-tex:
@Frenzie 

After a day of research, I have finally managed to find solution that is comparable in complexity to miniwebproxy. [Luarock "htmlparser"](https://github.com/msva/lua-htmlparser) does the task very well and I've managed to write simple filter just to check whether it would fit the check

```
local file = io.input("./file.html")
local text = io.read("*a") file:close()
local htmlparser = require("htmlparser")
local root = htmlparser.parse(text)

local selectors = {
        "main",
        "article",
        "div#main",
        "#main-article",
        ".main-content",
        "#body",
        "#content",
        ".content",
        "div#article",
        "div.article",
        "div.post",
[...]

PR #6228, opened 2020-06-03
[NewsDownloader] Added an HTML filter through a CSS selector

lich-tex:
Works exactly as expected, requires [htmlparser luarock](https://msva.github.io/lua-htmlparser/) which I do not know now how to include.

I am open to any corrections/edits. This is my first Lua code so please be wary.
Solves #6185 . I have tested it on Kindle PW3 now, and it works surprisingly quickly with convoluted HTML of Reuters (by just throwing my local .luarocks to rocks directory 🤣 .

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6228)

### DIFF
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -27,7 +27,7 @@ local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 -- filter HTML using CSS selector
 local function filter(text, element)
     local htmlparser = require("htmlparser")
-    local root = htmlparser.parse(text, 10000)
+    local root = htmlparser.parse(text, 5000)
     local filtered = nil
     local selectors = {
         "main",
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -28,7 +28,7 @@ local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 local function filter(text, element)
     local htmlparser = require("htmlparser")
     local root = htmlparser.parse(text)
-    local filtered = ""
+    local filtered = nil
     local selectors = {
         "main",
         "article",
@@ -49,22 +49,22 @@ local function filter(text, element)
         "div#newsstorytext",
         "div.general",
         }
-    if element then table.insert(selectors, 1, element) end
-    for _, sel in pairs(selectors) do
+    if element then 
+        table.insert(selectors, 1, element)
+    end
+    for _, sel in ipairs(selectors) do
        local elements = root:select(sel)
-       local stop = false
        if elements then
            for _, e in ipairs(elements) do
                filtered = e:getcontent()
                if filtered then 
-                   stop = true
                    break
                end
            end
-           if stop then break end
+           if filtered then break end
        end
     end
-    if filtered == "" then return text end
+    if not filtered then return text end
     return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
 end
 
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -24,11 +24,11 @@ local max_redirects = 5; --prevent infinite redirects
 local TIMEOUT_CODE = "timeout" -- from socket.lua
 local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 
+-- filter HTML using CSS selector
 local function filter(text, element)
     local htmlparser = require("htmlparser")
     local root = htmlparser.parse(text)
     local filtered = ""
-    
     local selectors = {
         "main",
         "article",
@@ -49,20 +49,21 @@ local function filter(text, element)
         "div#newsstorytext",
         "div.general",
         }
-    
     if element then selectors[0] = element end
-    
-    for _,sel in pairs(selectors) do
+    for _, sel in pairs(selectors) do
        local elements = root:select(sel)
-       if not elements then return text end
-       for _,e in ipairs(elements) do
+       local stop = false
+       for _, e in ipairs(elements) do
            filtered = e:getcontent()
-           if filtered then goto endloop end
+           if filtered then 
+               stop = true
+               break
+           end
        end
-end
-
-::endloop::
-return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
+       if stop then break end
+    end
+    if not filtered then return text end
+    return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
 end
 
 -- Sink that stores into a table, aborting if maxtime has elapsed
@@ -222,7 +223,6 @@ local ext_to_mimetype = {
     ttf = "application/truetype",
     woff = "application/font-woff",
 }
-
     
 -- Create an epub file (with possibly images)
 function EpubDownloadBackend:createEpub(epub_path, html, url, include_images, message, filter_enable, filter_element)
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -61,10 +61,14 @@ local function filter(text, element)
                    break
                end
            end
-           if filtered then break end
+           if filtered then
+              break 
+           end
        end
     end
-    if not filtered then return text end
+    if not filtered then 
+        return text 
+    end
     return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
 end
 
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -24,6 +24,47 @@ local max_redirects = 5; --prevent infinite redirects
 local TIMEOUT_CODE = "timeout" -- from socket.lua
 local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 
+local function filter(text, element)
+    local htmlparser = require("htmlparser")
+    local root = htmlparser.parse(text)
+    local filtered = ""
+    
+    local selectors = {
+        "main",
+        "article",
+        "div#main",
+        "#main-article",
+        ".main-content",
+        "#body",
+        "#content",
+        ".content",
+        "div#article",
+        "div.article",
+        "div.post",
+        "div.post-outer",
+        ".l-root",
+        ".content-container",
+        ".StandardArticleBody_body",
+        "div#article-inner",
+        "div#newsstorytext",
+        "div.general",
+        }
+    
+    if element then selectors[0] = element end
+    
+    for _,sel in pairs(selectors) do
+       local elements = root:select(sel)
+       if not elements then return text end
+       for _,e in ipairs(elements) do
+           filtered = e:getcontent()
+           if filtered then goto endloop end
+       end
+end
+
+::endloop::
+return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
+end
+
 -- Sink that stores into a table, aborting if maxtime has elapsed
 local function sink_table_with_maxtime(t, maxtime)
     -- Start counting as soon as this sink is created
@@ -182,14 +223,14 @@ local ext_to_mimetype = {
     woff = "application/font-woff",
 }
 
+    
 -- Create an epub file (with possibly images)
-function EpubDownloadBackend:createEpub(epub_path, html, url, include_images, message)
+function EpubDownloadBackend:createEpub(epub_path, html, url, include_images, message, filter_enable, filter_element)
     logger.dbg("EpubDownloadBackend:createEpub(", epub_path, ")")
     -- Use Trapper to display progress and ask questions through the UI.
     -- We need to have been Trapper.wrap()'ed for UI to be used, otherwise
     -- Trapper:info() and Trapper:confirm() will just use logger.
     local UI = require("ui/trapper")
-
     -- We may need to build absolute urls for non-absolute links and images urls
     local base_url = socket_url.parse(url)
 
@@ -201,7 +242,7 @@ function EpubDownloadBackend:createEpub(epub_path, html, url, include_images, me
     -- Not sure if this bookid may ever be used by indexing software/calibre, but if it is,
     -- should it changes if content is updated (as now, including the wikipedia revisionId),
     -- or should it stays the same even if revid changes (content of the same book updated).
-
+    if filter_enable then html = filter(html, filter_element) end
     local images = {}
     local seen_images = {}
     local imagenum = 1
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -49,20 +49,22 @@ local function filter(text, element)
         "div#newsstorytext",
         "div.general",
         }
-    if element then selectors[0] = element end
+    if element then table.insert(selectors, 1, element) end
     for _, sel in pairs(selectors) do
        local elements = root:select(sel)
        local stop = false
-       for _, e in ipairs(elements) do
-           filtered = e:getcontent()
-           if filtered then 
-               stop = true
-               break
+       if elements then
+           for _, e in ipairs(elements) do
+               filtered = e:getcontent()
+               if filtered then 
+                   stop = true
+                   break
+               end
            end
+           if stop then break end
        end
-       if stop then break end
     end
-    if not filtered then return text end
+    if filtered == "" then return text end
     return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
 end
 
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -49,7 +49,7 @@ local function filter(text, element)
         "div#newsstorytext",
         "div.general",
         }
-    if element then 
+    if element then
         table.insert(selectors, 1, element)
     end
     for _, sel in ipairs(selectors) do
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -27,7 +27,7 @@ local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 -- filter HTML using CSS selector
 local function filter(text, element)
     local htmlparser = require("htmlparser")
-    local root = htmlparser.parse(text)
+    local root = htmlparser.parse(text, 10000)
     local filtered = nil
     local selectors = {
         "main",
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -229,7 +229,6 @@ local ext_to_mimetype = {
     ttf = "application/truetype",
     woff = "application/font-woff",
 }
-    
 -- Create an epub file (with possibly images)
 function EpubDownloadBackend:createEpub(epub_path, html, url, include_images, message, filter_enable, filter_element)
     logger.dbg("EpubDownloadBackend:createEpub(", epub_path, ")")
--- a/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
+++ b/plugins/newsdownloader.koplugin/epubdownloadbackend.lua
@@ -57,17 +57,17 @@ local function filter(text, element)
        if elements then
            for _, e in ipairs(elements) do
                filtered = e:getcontent()
-               if filtered then 
+               if filtered then
                    break
                end
            end
            if filtered then
-              break 
+               break
            end
        end
     end
-    if not filtered then 
-        return text 
+    if not filtered then
+        return text
     end
     return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
 end
--- a/plugins/newsdownloader.koplugin/feed_config.lua
+++ b/plugins/newsdownloader.koplugin/feed_config.lua
@@ -28,7 +28,6 @@ return {--do NOT change this line
  -- 'filter_element="name_of_css.element.class" - means to filter the chosen CSS selector, it can be easily picked using a modern web browser
  -- default value is empty and it uses the default list of common selectors and they are used as fallback if this value is et
 
- -- 
  -- comment out line ("--" at line start) to stop downloading source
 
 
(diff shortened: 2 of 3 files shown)


plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,47 @@ local max_redirects = 5; --prevent infinite redirects
...
+       if not elements then return text end
+       for _,e in ipairs(elements) do
+           filtered = e:getcontent()
+           if filtered then goto endloop end
+       end
+end
+
+::endloop::
+return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
+end
```
poire-z, line 66: You can avoid that (ugly) goto by doing:
```lua
    for _, sel in pairs(selectors) do
       local elements = root:select(sel)
       if not elements then return text end
       local stop = false
       for _, e in ipairs(elements) do
           filtered = e:getcontent()
           if filtered then
               stop = true
               break
           end
       end
       if stop then
           break
       end
    end
    return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
```

Try to keep if/end aligned (your indentation of the `end` at line 62 was not aligned to its `for`, which made it strange to read :) - and a space after comma (`for _, sel`  rather than `for _,sel `) to make that easier to read.
Looks fine otherwise.

lich-tex: ok, i'll clean it up, thanks for a cleaner solution

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,47 @@ local max_redirects = 5; --prevent infinite redirects
...
+        "div#article-inner",
+        "div#newsstorytext",
+        "div.general",
+        }
+    
+    if element then selectors[0] = element end
+    
+    for _,sel in pairs(selectors) do
+       local elements = root:select(sel)
+       if not elements then return text end
```
poire-z, line 57: Is it right that you return when the first selector gives no element ? Shouldn't you try next selectors?
(There's no `continue` in Lua, only `break` so you have to imbricate all positive checks and let it continue naturally when it reaches the end of the `for` .)

lich-tex: Thanks, I did not notice.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,48 @@ local max_redirects = 5; --prevent infinite redirects
...
+        "div.post-outer",
+        ".l-root",
+        ".content-container",
+        ".StandardArticleBody_body",
+        "div#article-inner",
+        "div#newsstorytext",
+        "div.general",
+        }
+    if element then selectors[0] = element end
+    for _, sel in pairs(selectors) do
```
poire-z, line 53: There's a thing in Lua with table-tables and array-tables - and array tables start at 1.
So, this is not super proper (but we have some occurences of that kind of stuff in the code, and it often caused us some headaches :), so it's best to do that right.
Your `selectors` table is defined as an array of 18 elements: 1 .. 18.

You shouldn't create `selectors[0]`  as 0 is not an index (Lua starts at 1) - you're creating a `key=0 / value=element`  item in the table-table (one Lua table can be both a table-table and an array-table !).
Best to use `table.insert(selector, 1, element)` instead.

Then, using `for k, v = pairs(selectors)`, you iterate the table as a k/v table, and you get the keys in any order (may be that does not bother you, but if there's some sense of priority in the selectors list, it is lost).
Best to use `for i, v = ipairs(selectors)` (_ipairs_ with a _i_ like _iterate_) to iterate thru 1..18 (your selectors[0] would be ignored there).

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,48 @@ local max_redirects = 5; --prevent infinite redirects
...
+       local elements = root:select(sel)
+       local stop = false
+       for _, e in ipairs(elements) do
```
poire-z, line 56: I guess you want to insert an if and indent the rest:
```lua
local elements = root:select(sel)
if elements then
    for _, e in ipairs(elements) do
```

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
lich-tex: Huh, that is why it worked, and I wasn't made aware why does it not work. Thanks.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,48 @@ local max_redirects = 5; --prevent infinite redirects
...
+           filtered = e:getcontent()
+           if filtered then 
+               stop = true
+               break
+           end
+       end
+       if stop then break end
+    end
+    if not filtered then return text end
```
poire-z, line 65: Your initial assignment of the empty string `local filtered = ""` will make that test fail (so, you won't return text if there's no elements).
`if not filtered then ` will only trigger if filtered is `false` or `nil`. And not when it is an empty string `""` or an empty table `{}`. (That's different from most other languages.)

lich-tex: Huh, that's again quite a large difference.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,50 @@ local max_redirects = 5; --prevent infinite redirects
 local TIMEOUT_CODE = "timeout" -- from socket.lua
 local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 
+-- filter HTML using CSS selector
+local function filter(text, element)
+    local htmlparser = require("htmlparser")
+    local root = htmlparser.parse(text)
+    local filtered = ""
```
poire-z, line 31: You could just use `filtered = nil` - and use it instead of `stop` below. It would become more natural.
Just 2 `if filtered then break end` and at the end `if not filtered then return text end`

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,50 @@ local max_redirects = 5; --prevent infinite redirects
...
+        "div.post-outer",
+        ".l-root",
+        ".content-container",
+        ".StandardArticleBody_body",
+        "div#article-inner",
+        "div#newsstorytext",
+        "div.general",
+        }
+    if element then table.insert(selectors, 1, element) end
+    for _, sel in pairs(selectors) do
```
poire-z, line 53: `ipairs`

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,50 @@ local max_redirects = 5; --prevent infinite redirects
...
+        "div.post",
+        "div.post-outer",
+        ".l-root",
+        ".content-container",
+        ".StandardArticleBody_body",
+        "div#article-inner",
+        "div#newsstorytext",
+        "div.general",
+        }
+    if element then table.insert(selectors, 1, element) end
```
poire-z, line 52: Single line is ok for minor stuff like if ... then return/break end.
But this one is important for the logic, so best to make it 3 lines as it's more readable.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,50 @@ local max_redirects = 5; --prevent infinite redirects
...
+    for _, sel in ipairs(selectors) do
+       local elements = root:select(sel)
+       if elements then
+           for _, e in ipairs(elements) do
+               filtered = e:getcontent()
+               if filtered then 
+                   break
+               end
+           end
+           if filtered then break end
```
poire-z, line 64: Make these `if then break end` consistent: either 1-line, or either 3-lines (same for the `if not filtered then return text end` below).
Given the medium level of thinking we both needed to get that right :), I'd prefer if you went with 3-lines, for clarity.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,50 @@ local max_redirects = 5; --prevent infinite redirects
...
+        "div.post-outer",
+        ".l-root",
+        ".content-container",
+        ".StandardArticleBody_body",
+        "div#article-inner",
+        "div#newsstorytext",
+        "div.general",
+        }
+    if element then 
+        table.insert(selectors, 1, element)
```
poire-z, line 53: Just a comment to make sure that's your intended bahaviour: when the user has specified an element, you insert it at 1 (but you keep all the others at 2...19) - and you don't drop the table to use just the user-specified element.
So the other selectors are used as a fallback when the user element is not found. Right?

lich-tex: Yeah, that is my point, I think it is a good safety net.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
 local TIMEOUT_CODE = "timeout" -- from socket.lua
 local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 
+-- filter HTML using CSS selector
+local function filter(text, element)
+    local htmlparser = require("htmlparser")
+    local root = htmlparser.parse(text)
```
lich-tex, line 30: Here, I am not sure whether we should use `htmlparser` default looplimit/page depth. With pages with more than 1000 tag depth (which are more popular than expected) the entire thing just does not work and outputs the inserted text. I don't know what value should I set this to then, or to allow user to change it on per feed basis (but then am I not overloading the config?).

poire-z:
(Good for me as far as Lua coding is concerned. I can't help much with htmlparser integration.)

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Frenzie: Is it actually a tag depth of over 1000 or a deadloop because of some invalid HTML that confuses the parser? Please provide a sample of what we're talking about to fully understand the issue.

Assuming it's not a deadloop, the relevant aspect is how much time it takes on an actual device. Is a limit of 10000 still executed in under 100 ms? I'd say that's perfectly fine. Would it take 10 seconds? Maybe consider making it an option. I hope you get the idea, it's meaningless to just talk about the number 1000. ;-)

lich-tex: The only case I had this issue was with [BussinessInsider](https://www.businessinsider.com/3-boogaloo-men-terror-charges-george-floyd-protest-riot-conspiracy-2020-6?IR=T). Setting it to 10000 solves the issue, and *on desktop* it takes 0.24 secs, I can check later on my kindle what is the result there.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
 local TIMEOUT_CODE = "timeout" -- from socket.lua
 local MAXTIME_CODE = "maxtime reached" -- from sink_table_with_maxtime
 
+-- filter HTML using CSS selector
+local function filter(text, element)
+    local htmlparser = require("htmlparser")
```
Frenzie, line 29: It's clearer to put this at the top on line 3.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Frenzie: I'm not quite sure to what extent I think a "safety net" makes sense. But you can always say enable_filter = false, correct?

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
lich-tex: Isn't it better to not import the library if it is only used in case enable_filter=true? Or is the time needed for such import negligent?

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
...
+           end
+           if filtered then
+              break 
+           end
+       end
+    end
+    if not filtered then 
+        return text 
+    end
+    return "<!DOCTYPE html><html><head></head><body>" .. filtered .. "</body></html>"
```
Frenzie, line 72: Should this take the title from the original HTML document?

lich-tex: I do not think such is needed as the title is already imported before the filter function is initiated and the title exists as epub title, so it is visible in the list.

Matter of preference nonetheless.

Frenzie:
Oh wait, PC World is just the feed text. :facepalm: Regardless, it seems to be working on Reuters.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
lich-tex: I have set it to 10000 now, and checking on Kindle it seems the slowdown is null so it is a fine resolve.

The complex sites still take a bit longer, but nothing which would disrupt usage.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Frenzie: This is supposed to go directly into an EPUB? Is this valid EPUB3? An EPUB content file would normally be expected to start with something like this since it's X(HT)ML:

```xml
<?xml version="1.0" encoding="utf-8"?>
<html xmlns="http://www.w3.org/1999/xhtml">
```

lich-tex:
> Oh wait, PC World is just the feed text. 🤦 Regardless, it seems to be working on Reuters.

If you want a complex example that takes a bit longer I recommend (my reddit world news atom with direct links to articles)[https://dataswamp.org/~lich/news.atom]. It does seem to struggle with couple of more complex sites (interestingly the depth has no relation to the duration of parsing, oh the wonders of modern HTML). But, in the end, except for sites that just resign after being downloaded by something that is not a modern web browser the result is quite fine.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
lich-tex: Yes, if someone does not want to filter then they are free to disable it. And it is the default anyway (as to not force changes upon people).

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
lich-tex: Entire HTML in the default setting was going in anyway, and I've have had bad experience offering the later EPUB parser the HTML without this syntax header and footer. The rest of the program works the same after filtering so any changes done towards EPUB are there anyway. Instead of getting the full html, it just gets the things in the chosen selector and this header and footer.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Frenzie: It's negligible, yes. But it can stay here if you want it here.

Frenzie:
@lich-tex I just meant that this + https://github.com/koreader/koreader-base/pull/1110 seems to be functional.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Frenzie: You mean that crengine doesn't use the XML parser then?

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
Frenzie: Btw, 10000 was just a somewhat extreme suggestion; what's the actual value on that Business Insider thing? 2500 or so?

Frenzie: (Just wondering if it might be safer to go with 5000 to still have a large safety margin, for example.)

lich-tex: > Btw, 10000 was just a somewhat extreme suggestion; what's the actual value on that Business Insider thing? 2500 or so?

I checked with some simple js and it seems to be around 1050, in the most extreme case, that being biblegateway it is 2050. I think we can lower it to 5000 and that would leave only edge cases

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
lich-tex: Well, the rest of the program uses the old regex (kindly avoids summoning Zalgo) and then maybe xml parser, but the htmlparser is only used here.

poire-z: (crengine = our EPUB/HTML renderer)

With EPUB, crengine will always use the XML parser (while for standalone HTML files, it uses the loosy HTML parser that do tag autoclose and stuff).
But crengine does not care much about the DOCTYPE/XML/HTML header - no matter what it says, it won't change crengine's parsing behaviour.

I guess htmlparser, when spitting out the HTML from the selected element, will spit out a perfectly balanced serialized DOM, so crengine will probably be happier with that than with the original possibly messy downloaded HTML.

And indeed, as these HTML files ar just added to an EPUB, they don't need a TITLE - it would not be used by crengine anyway.

Frenzie: Two minor things, crengine (or alternatively MuPDF) is what renders the thing.

And Lua patterns aren't regex.

Anyway, like I said it can stay here.

lich-tex: I see now, I was mixed up, but basically everything is fine in your opinion regarding this?

Frenzie: If there's no reason not to use `<?xml version="1.0" encoding="utf-8"?><html xmlns="http://www.w3.org/1999/xhtml">`, I'd rather see it in there as a good example for people learning by looking at the output. One thing I didn't fully realize a few years ago is that what seems like allowing a fairly innocent shortcut that we all know isn't strictly right, just not wanting to nitpick too much, often somehow comes back to bite you in the future.

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -181,15 +229,14 @@ local ext_to_mimetype = {
     ttf = "application/truetype",
     woff = "application/font-woff",
 }
-
+
```
Frenzie, line 232:
```suggestion

```

plugins/newsdownloader.koplugin/feed_config.lua
```diff
@@ -21,12 +21,19 @@ return {--do NOT change this line
  -- 'include_images=false' - means ignore any images, only download the text (faster download, smaller file sizes)
  -- default value is 'false' (if no 'include_images' entry)
 
+ -- 'enable_filter=true' - means filter a css selector and thus delimit the page to just that (does not apply if download_full_article=false)
+ -- 'enable_filter=false' - means no such filtering and having unlimited document
```
Frenzie, line 25:
```suggestion
 -- 'enable_filter=false' - means no such filtering and including the full page
```

plugins/newsdownloader.koplugin/feed_config.lua
```diff
@@ -21,12 +21,19 @@ return {--do NOT change this line
  -- 'include_images=false' - means ignore any images, only download the text (faster download, smaller file sizes)
  -- default value is 'false' (if no 'include_images' entry)
 
+ -- 'enable_filter=true' - means filter a css selector and thus delimit the page to just that (does not apply if download_full_article=false)
```
Frenzie, line 24:
```suggestion
 -- 'enable_filter=true' - means filter using a CSS selector to delimit part of the page to just that (does not apply if download_full_article=false)
```

plugins/newsdownloader.koplugin/feed_config.lua
```diff
@@ -21,12 +21,19 @@ return {--do NOT change this line
...
+ -- 'enable_filter=false' - means no such filtering and having unlimited document
+ -- default value is 'false'
+
+ -- 'filter_element="name_of_css.element.class" - means to filter the chosen CSS selector, it can be easily picked using a modern web browser
+ -- default value is empty and it uses the default list of common selectors and they are used as fallback if this value is et
```
Frenzie, line 29:
```suggestion
 -- The default value is empty. The default list of common selectors is used as fallback if this value is set.
```

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
...
+        "div.post",
+        "div.post-outer",
+        ".l-root",
+        ".content-container",
+        ".StandardArticleBody_body",
+        "div#article-inner",
+        "div#newsstorytext",
+        "div.general",
+        }
+    if element then
```
Frenzie, line 52:
```suggestion
    if element then
```

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
...
+        }
+    if element then 
+        table.insert(selectors, 1, element)
+    end
+    for _, sel in ipairs(selectors) do
+       local elements = root:select(sel)
+       if elements then
+           for _, e in ipairs(elements) do
+               filtered = e:getcontent()
+               if filtered then
```
Frenzie, line 60:
```suggestion
               if filtered then
```

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
...
+       local elements = root:select(sel)
+       if elements then
+           for _, e in ipairs(elements) do
+               filtered = e:getcontent()
+               if filtered then 
+                   break
+               end
+           end
+           if filtered then
+              break
```
Frenzie, line 65:
```suggestion
               break
```

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
...
+               if filtered then 
+                   break
+               end
+           end
+           if filtered then
+              break 
+           end
+       end
+    end
+    if not filtered then
```
Frenzie, line 69:
```suggestion
    if not filtered then
```

plugins/newsdownloader.koplugin/epubdownloadbackend.lua
```diff
@@ -24,6 +24,54 @@ local max_redirects = 5; --prevent infinite redirects
...
+                   break
+               end
+           end
+           if filtered then
+              break 
+           end
+       end
+    end
+    if not filtered then 
+        return text
```
Frenzie, line 70:
```suggestion
        return text
```

PR #6195, opened 2020-05-26
[UX] Implement hasFewKeys prototype

Frenzie:
This is a quick exploration into how the new Device:hasFewKeys() method could work to make things more usable on very limited devices.

In the reader, the right is repurposed to open the menu. Left in turn closes it.

The same principle is applied to ButtonDialog. This means you can select anything in principle, but once you go right you can't go back.

References <https://github.com/koreader/koreader/issues/4029>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6195)

PR #6229, opened 2020-06-04
[fix] Ignore sha2.lua in doc generation

Frenzie:
Otherwise it'll choke on the comments in that file.

See <https://github.com/koreader/koreader/pull/6195#issuecomment-638801075>.

Accidentally introduced in <https://github.com/koreader/koreader-base/pull/1105>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6229)

### DIFF
--- a/doc/config.ld
+++ b/doc/config.ld
@@ -22,4 +22,5 @@ file = {
     '../frontend',
     '../base/ffi',
     '../platform/android/luajit-launcher/assets',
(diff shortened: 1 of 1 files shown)


Frenzie:
@poire-z You can check it locally if you run `ldoc .` in `/docs`, provided you have ldoc in PATH.

PR #6233, opened 2020-06-04
Don't exit KUAL on the Voyage

### LINKED ISSUE #6038: Kindle voyage  button problem and fix
When use koreader in my kindle voyage everything is well except paper up button.

What i found: 
this problem root in kindle voyage system in which we can find "lab126_whisper_touch.lua". When your kindle in the first screen the whisper_touch is disable, but when the second screen of your document which is available.  
If you start koreader using kual which will back to the first screen of your document (home) meanwhile the whisper touch is disable.

**Solution: When you start koreader using kual,  continue tap next page button  when you back to the first screen of your home page. After koreader start back page button also work well.**
👇↓

update:

> here is my solution:
first: add INSERT OR IGNORE INTO "properties" VALUES('com.mobileread.ixtab.kindlelauncher','whisper-touch','supported'); into /var/local/appreg.db use sqlite3 /var/local/appreg.db < whispertouch.install.sql
second: check if kindle voyage do not exit kual menu before koreader starts.
like this :
"name": "Start the filemanager kv",
"if": "\"KindleVoyage\" -m || || || ||",
"priority": 7,
"action": "/mnt/us/koreader/koreader.sh",
"params": "--kual /mnt/us/documents",
"exitmenu": false,
"status": false,
"internal": "status Start KOReader on the File Manager"
here is what I use:
[test.zip](https://github.com/koreader/koreader/files/4466454/test.zip)

can run it using kterm under kindle, then replace koreader/menu.json.

I test many times in my Kindle voyage. Running: 5.9.7

NiLuJe:
This should allow WhisperTouch support to stay enabled, allowing
consistent button behavior.

c.f., https://github.com/koreader/koreader/issues/6038#issuecomment-612564693

Fix #6038

(Requires today's KUAL build, Booklet only).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6233)

### DIFF
--- a/platform/kindle/extensions/koreader/menu.json
+++ b/platform/kindle/extensions/koreader/menu.json
@@ -6,6 +6,7 @@
 		"items": [
 		{
 			"name": "Start the filemanager",
+			"if": "\"KindleVoyage\" -m!",
 			"priority": 1,
 			"action": "/mnt/us/koreader/koreader.sh",
 			"params": "--kual /mnt/us/documents",
@@ -13,10 +14,31 @@
 			"internal": "status Start KOReader on the File Manager"
 		},
 		{
+			"name": "Start the filemanager",
+			"if": "\"KindleVoyage\" -m",
+			"priority": 1,
+			"action": "/mnt/us/koreader/koreader.sh",
+			"params": "--kual /mnt/us/documents",
+			"exitmenu": false,
+			"status": false,
+			"internal": "status Start KOReader on the File Manager"
+		},
+		{
 			"name": "Open the last document",
+			"if": "\"KindleVoyage\" -m!",
 			"priority": 2,
 			"action": "/mnt/us/koreader/koreader.sh",
 			"params": "--kual",
+			"status": false,
+			"internal": "status Start KOReader on the last document"
+		},
+		{
+			"name": "Open the last document",
+			"if": "\"KindleVoyage\" -m",
+			"priority": 2,
+			"action": "/mnt/us/koreader/koreader.sh",
+			"params": "--kual",
+			"exitmenu": false,
 			"status": false,
 			"internal": "status Start KOReader on the last document"
 		},
@@ -40,6 +62,7 @@
 		},
 		{
 			"name": "Start the filemanager (ASAP)",
+			"if": "\"KindleVoyage\" -m!",
 			"priority": 5,
 			"action": "/mnt/us/koreader/koreader.sh",
 			"params": "--kual --asap /mnt/us/documents",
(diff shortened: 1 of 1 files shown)


platform/kindle/extensions/koreader/menu.json
```diff
@@ -6,17 +6,39 @@
...
+			"priority": 1,
+			"action": "/mnt/us/koreader/koreader.sh",
+			"params": "--kual /mnt/us/documents",
+			"exitmenu": false,
+			"status": false,
+			"internal": "status Start KOReader on the File Manager"
+		},
+		{
+			"name": "Open the last document",
+			"if": "\"KindleVoyage\" -m!",
```
yparitcher, line 28: you added the condition by both "Open the last document" :)

yparitcher:
sorry missed the "!"

niveKKumar:
Doesn't work me for me on both latest KUAL (2.7) and latest Koreader (202102)
If you could explain how to send the log I could also attach it :)

NiLuJe:
Keep in mind the mention about requiring the *Booklet* build of KUAL, because that might not be a given on a Voyage, especially if you've had it for a while.

niveKKumar:
Okay I will try it with the booklet but it isn't stable is it?
It does work through a workaround (already mentioned on other thread).
When I open KOReaded and press on home (top left) and press the next button on the side of the voyage then if timed right it does work. It's maybe because KUAL kills the original homescreen (if it does 😂 I have no clue what KUAL does haha)

NiLuJe:
Err, I have no idea what you're thinking about, but, no, the Booklet is perfectly fine, and you're probably not currently using it, hence the weirdness.

hius07:
@yparitcher , is it possible to add this solution to the KOL?
I've got the Voyage and can test if needed.

yparitcher:
> @yparitcher , is it possible to add this solution to the KOL?
> I've got the Voyage and can test if needed.

I can try, but it will be slightly complex.
In order for whispertouch to work the Booklet that the kindle opens must stay open . KUAL does this by not closing when running Koreader in the voyage. KUAL should stay open after exiting Koreader (Please confirm.) With KOL this is a little harder as it never really opens, rather just launches Koreader and dies.

@NiLuJe 
Is there a reliable trigger for when Koreader exits that i can use for cleanup in KOL on the voyage?

hius07:
> KUAL should stay open after exiting Koreader (Please confirm.)

No, it doesn't. After exiting Koreader I see the device home screen.

NiLuJe:
@yparitcher: The presence (or not) of the startup script copy in /tmp might be a good check (this vaguely rings a bell ^^).

NiLuJe:
@hius07: I honestly don't recall what `exitmenu` does or doesn't do, but this is mildly surprising, I'd have expected what @yparitcher just described.

yparitcher:
@hius07 
Please try https://github.com/yparitcher/KUAL_Booklet/releases/tag/V1.1

@NiLuJe 
As hius07 pointed out it just doesn't kill KUAL right away, rather is gets lifecycled somewhere ?? probably by the kindle?
so i can replicate this behavior by just not self-killing right away

yparitcher:
Btw, If you want, this can be simplified to not exit on any kindle with no side effects so far.

hius07:
> Please try

Great, works perfectly, thank you very much!

yparitcher:
> > Please try
> 
> Great, works perfectly, thank you very much!

I will wait a bit, and no one complains i will make it a full release.

hius07:
To be noted, I'm using koreader.sh from the 2021.05 release, because I wasn't happy with my PW2 behaviour after https://github.com/koreader/koreader/pull/7858/commits/9755ceafe7c464aa9882805133883c170c85888b.

NiLuJe:
@hius07: What kind of issues were you having?

hius07:
Unstable behaviour after exiting KOReader: long screen restoring with many redrawings and even rebooting a couple of times.

hius07:
@NiLuJe more info from KOA3 user: frequent hanging on exit from KOReader since June, he removed `kb` service in
https://github.com/koreader/koreader/blob/c16d2712efb7fc1c9db7c6a03ec98d1daa280a8c/platform/kindle/koreader.sh#L67-L68
reports no hanging for a week.

NiLuJe:
@hius07: That's good news, thanks ;).

Feel free to either test this some more, or just send a PR to unleash it on the unsuspecting masses ;p.

PR #6238, opened 2020-06-06
Allow message to be overlaid on other screensavers

sladflob:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6238)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.un~
 cscope.*
 *.orig
+*.lua~
 lua
 lua-*
 .reader.kpdfview.lua
@@ -36,5 +37,4 @@ i18n
 
 /.cproject
 /.project
-frontend/ui/screensaver.lua~
-*.lua~
+.gitignore~
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ i18n
 /.cproject
 /.project
 frontend/ui/screensaver.lua~
+*.lua~
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ i18n
 
 /.cproject
 /.project
+frontend/ui/screensaver.lua~
--- a/frontend/ui/elements/screensaver_menu.lua
+++ b/frontend/ui/elements/screensaver_menu.lua
@@ -109,14 +109,16 @@ return {
         end,
         callback = function()
             G_reader_settings:saveSetting("screensaver_type", "disable")
-        end
+        end,
+        separator = true
     },
     {
         text = _("Add message to screensaver"),
         checked_func = showMessage,
         callback = function()
             G_reader_settings:saveSetting("screensaver_show_message", not showMessage())
-        end
+        end,
+        separator = true
     },
     {
         text = _("Settings"),
--- a/frontend/ui/elements/screensaver_menu.lua
+++ b/frontend/ui/elements/screensaver_menu.lua
@@ -110,7 +110,7 @@ return {
         callback = function()
             G_reader_settings:saveSetting("screensaver_type", "disable")
         end,
-        separator = true
+        separator = true,
     },
     {
         text = _("Add message to screensaver"),
@@ -118,7 +118,7 @@ return {
         callback = function()
             G_reader_settings:saveSetting("screensaver_show_message", not showMessage())
         end,
-        separator = true
+        separator = true,
     },
     {
         text = _("Settings"),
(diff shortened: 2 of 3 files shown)


frontend/ui/elements/screensaver_menu.lua
```diff
@@ -98,29 +99,23 @@ return {
...
     {
-        text = _("Leave screen as it is"),
-        checked_func = function()
-            if screensaverType() == "disable" then
-                return true
-            else
-                return false
-            end
-        end,
+        text = _("Add message to screensaver"),
```
poire-z, line 115: I guess you can add a `separator = true,` to the previous menu item - and may be also to this one (if it needs to be separated from the following "Settings" - or not if we consider it a bit like a setting).

frontend/ui/screensaver.lua
```diff
@@ -383,7 +396,7 @@ function Screensaver:show(event, fallback_message)
         end
 
         if message_pos == "middle" or message_pos == nil then
-            widget = InfoMessage:new{
+            message_widget = InfoMessage:new{
```
poire-z, line 399: I think you'll need `local message_widget` somewhere, so it's really a local.
There was a complain about `local message_widget = nil` just because of the `= nil` assignment I guess.

frontend/ui/screensaver.lua
```diff
@@ -410,8 +423,30 @@ function Screensaver:show(event, fallback_message)
...
             }
         end
+
         -- No overlay needed as we just displayed the message
         overlay_message = nil
+
+        -- check if message_widget should be overlaid on another widget
+        if message_widget ~= nil then
+            if widget ~= nil
+            then
```
poire-z, line 433: On a single line please :) And even just `if widget then -- we have a screensaver widget`

frontend/ui/screensaver.lua
sladflob: Oops, I went to delete the "= nil" and deleted the whole line.

I think it was complaining because message_widget gets assigned in both the if and else clauses so the nil value was never used? Anyway, hopefully will work now.

frontend/ui/elements/screensaver_menu.lua
sladflob: Just tried it and it looks better with separators above and below. Will commit that.

.gitignore
```diff
@@ -37,3 +37,4 @@ i18n
 /.cproject
 /.project
 frontend/ui/screensaver.lua~
```
poire-z, line 39: Remove that too specific one, and may be move the next one `*.lua~` above around line 9 near its buddies.

frontend/ui/elements/screensaver_menu.lua
```diff
@@ -109,14 +109,16 @@ return {
         end,
         callback = function()
             G_reader_settings:saveSetting("screensaver_type", "disable")
-        end
+        end,
+        separator = true
```
poire-z, line 113: (For info: you can let a comma at the end of the last item - it will avoid showing in diff when you add an item - like this `end,` as you had to add the comma.)

frontend/ui/screensaver.lua
```diff
@@ -429,8 +430,7 @@ function Screensaver:show(event, fallback_message)
 
         -- check if message_widget should be overlaid on another widget
         if message_widget ~= nil then
-            if widget ~= nil
-            then
+            if widget then      -- we have a screensaver widget
```
poire-z, line 433: No real need for that much spacing before the comment (I guess it's ok when you want to align it with other comments, but it looks strange here).

poire-z:
May be you could add in ScreenSaver:show():
```lua
-- pseudo code, use the right variables and setting names :)
if screensaver_type == "message" then
    -- obsolete screensaver_type: migrate it to new setting
    screensaver_type == "disable" 
    saveSetting("screensaver_type", "disable")
    saveSetting("show_message", true)
end
```
(it's just an additional test, shouldn't cost much)

frontend/ui/screensaver.lua
sladflob: Sure... I do like a bit of extra whitespace but happy to change it.

frontend/ui/elements/screensaver_menu.lua
```diff
@@ -121,7 +109,16 @@ return {
...
```
Frenzie, line 113:
```suggestion
        separator = true,
```

frontend/ui/elements/screensaver_menu.lua
```diff
@@ -121,7 +109,16 @@ return {
...
+    },
+    {
+        text = _("Add message to screensaver"),
+        checked_func = showMessage,
+        callback = function()
+            G_reader_settings:saveSetting("screensaver_show_message", not showMessage())
+        end,
+        separator = true
```
Frenzie, line 121:
```suggestion
        separator = true,
```

frontend/ui/screensaver.lua
```diff
@@ -410,8 +431,28 @@ function Screensaver:show(event, fallback_message)
                 }
             }
         end
+
         -- No overlay needed as we just displayed the message
         overlay_message = nil
+
+        -- check if message_widget should be overlaid on another widget
+        if message_widget ~= nil then
```
Frenzie, line 439: The reverse works well enough.
```suggestion
        if message_widget then
```

PR #6240, opened 2020-06-07
Ensure feature-parity in ScreenSaver handling on Kindle

### LINKED ISSUE #5724: Landscape screensaver to portrait
* KOReader version: 2019.11
* Device: PW4
Recently, I have some pdf books. My best choice was to read in landscape mode with support of semi-auto crop. Everything was fine until I hit sleep button, the book cover was in landscape mode which was not the direction that it was designed.

May I ask or a feature to auto rotate book cover to portrait mode even we're reading in landscape mode?
Thank you.

NiLuJe:
We were missing the fancy extra white flash and landscape handling that
most every other eInk device gets ;).

(On non-SO devices, of course).

Fix #5724

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6240)

### DIFF
--- a/frontend/device/kindle/device.lua
+++ b/frontend/device/kindle/device.lua
@@ -161,10 +161,10 @@ function Kindle:usbPlugIn()
 end
 
 function Kindle:intoScreenSaver()
-    local Screensaver = require("ui/screensaver")
-    if self:supportsScreensaver() then
-        -- NOTE: Meaning this is not a SO device ;)
-        if self.screen_saver_mode == false then
+    if self.screen_saver_mode == false then
+        if self:supportsScreensaver() then
+            -- NOTE: Meaning this is not a SO device ;)
+            local Screensaver = require("ui/screensaver")
             -- NOTE: Pilefered from Device:onPowerEvent @ frontend/device/generic/device.lua
             -- Mostly always suspend in Portrait/Inverted Portrait mode...
             -- ... except when we just show an InfoMessage or when the screensaver
@@ -195,10 +195,8 @@ function Kindle:intoScreenSaver()
                 self.orig_rotation_mode = nil
             end
             Screensaver:show()
-        end
-    else
-        -- Let the native system handle screensavers on SO devices...
-        if self.screen_saver_mode == false then
+        else
+            -- Let the native system handle screensavers on SO devices...
             if os.getenv("AWESOME_STOPPED") == "yes" then
                 os.execute("killall -cont awesome")
             end
@@ -210,8 +208,8 @@ end
 
 function Kindle:outofScreenSaver()
     if self.screen_saver_mode == true then
-        local Screensaver = require("ui/screensaver")
         if self:supportsScreensaver() then
+            local Screensaver = require("ui/screensaver")
             -- Restore to previous rotation mode, if need be.
             if self.orig_rotation_mode then
                 self.screen:setRotationMode(self.orig_rotation_mode)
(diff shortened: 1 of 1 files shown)


NiLuJe:
Thanks for the heads up in https://github.com/koreader/koreader/issues/5724#issuecomment-570089493 @poire-z ;).

Don't know why I managed to forget about that (new year's eve might not have been a great time to report a bug :D).

NiLuJe:
(This only takes care of the Kindle thing. If @poire-z's hunch is right abou the PDF and AR thingy, I don't think there's anything much we could do without a giant pile of hacks anyway...).

NiLuJe:
Ping @yparitcher, because the extra delay caused by the pre-ss white flash might not play nice with your stuff?

yparitcher:
> Ping @yparitcher, because the extra delay caused by the pre-ss white flash might not play nice with your stuff?

Thanks for asking, i use `screensaver_type  == disable` and manage it myself so this works fine

frontend/device/kindle/device.lua
```diff
@@ -165,6 +165,35 @@ function Kindle:intoScreenSaver()
...
+                if bit.band(self.orig_rotation_mode, 1) == 1 then
+                    -- i.e., only switch to Portrait if we're currently in *any* Landscape orientation (odd number)
+                    self.screen:setRotationMode(0)
+                else
+                    self.orig_rotation_mode = nil
+                end
+
+                -- On eInk, if we're using a screensaver mode that shows an image,
+                -- flash the screen to white first, to eliminate ghosting.
+                if self:hasEinkScreen() and
```
yparitcher, line 185: arent all kindles (supported by koreader) Eink?

NiLuJe: Yep, left it in to make it extremely obvious this was a c/p ^^.

NiLuJe:
> rewritten like this requires less if's.
> 
> ```diff
> @@ -161,10 +161,10 @@ function Kindle:usbPlugIn()
>  end
>  
>  function Kindle:intoScreenSaver()
> -    local Screensaver = require("ui/screensaver")
> -    if self:supportsScreensaver() then
> +    if self.screen_saver_mode == false then
> +        if self:supportsScreensaver() then
>          -- NOTE: Meaning this is not a SO device ;)
> -        if self.screen_saver_mode == false then
> +            local Screensaver = require("ui/screensaver")
>              -- NOTE: Pilefered from Device:onPowerEvent @ frontend/device/generic/device.lua
>              -- Mostly always suspend in Portrait/Inverted Portrait mode...
>              -- ... except when we just show an InfoMessage or when the screensaver
> @@ -195,10 +195,8 @@ function Kindle:intoScreenSaver()
>                  self.orig_rotation_mode = nil
>              end
>              Screensaver:show()
> -        end
> -    else
> +        else
>          -- Let the native system handle screensavers on SO devices...
> -        if self.screen_saver_mode == false then
>              if os.getenv("AWESOME_STOPPED") == "yes" then
>                  os.execute("killall -cont awesome")
>              end
> @@ -210,8 +208,8 @@ end
>  
>  function Kindle:outofScreenSaver()
>      if self.screen_saver_mode == true then
> -        local Screensaver = require("ui/screensaver")
>          if self:supportsScreensaver() then
> +            local Screensaver = require("ui/screensaver")
>              -- Restore to previous rotation mode, if need be.
>              if self.orig_rotation_mode then
>                  self.screen:setRotationMode(self.orig_rotation_mode)
> ```

I prefer to keep the extra outer if as `self:supportsScreensaver() ` to make the separation between !SO & SO very clear ;).

The tighter scope on the require does make sense, though ;).

NiLuJe:
Huh. I say that, and I realize that's not even consistent between in and out :D.

Oh, well, will clean this up after dinner ;).

yparitcher:
> Huh. I say that, and I realize that's not even consistent between in and out :D.

That why i noticed :)

PR #6241, opened 2020-06-07
Remove a dodgy scheduleIn

NiLuJe:
It was causing a double-refresh on orientation change via ConfigDialog

I could only reproduce it on Kindle, for some strange reason (slower?).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6241)

### DIFF
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -786,7 +786,7 @@ function ReaderRolling:onUpdatePos()
         return true
     end
     -- Calling this now ensures the re-rendering is done by crengine
-    -- so the delayed updatePos() has good info and can reposition
+    -- so updatePos() has good info and can reposition
     -- the previous xpointer accurately:
     self.ui.document:getCurrentPos()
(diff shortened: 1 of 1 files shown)


NiLuJe:
Dates back to #415 (!)

I believe @poire-z actually fixed it in #4793

We're not threaded, so, I assume getCurrentPos is blocking, and as such the delay is unnecessary.

At a (very) quick glance, this didn't break anything, and fixed the double-refresh ;).

NiLuJe:
Sidenote: sticking a `debug.traceback()` at the top of `setDirty` was incredibly helpful ;).

poire-z:
Seems to work allright on my Kobo.
I was worried this section of the code (that I spent time polishing - but I have now forgotten most of what's at play) was modified - but git blame tells me this scheduleIn() is actually not mine :) So I probably just went with it and didn't date touching it. So, hopefully it wasn't really needed.

NiLuJe:
Yep, it's *really* old, and you tackled the issue properly, so it *feels* right, but thanks for double-checking ;).

PR #6242, opened 2020-06-07
Update bundled fonts

NiLuJe:
* Minor FreeFont update
* Noto Sans CJK downgrade to fix a number of issues w/ the git version
* Initial import of Noto Naskh Arabic

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6242)

### DIFF
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -46,7 +46,7 @@ local CreDocument = Document:new{
     --   (Noto Sans & Serif do not have these symbol glyphs).
     fallback_fonts = {
         "Noto Sans CJK SC",
-        "Noto Sans Arabic UI",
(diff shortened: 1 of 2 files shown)


NiLuJe:
@poire-z: Should I add Naskh to a fallback list somewhere?

poire-z:
For CRE, possibly just a matter of _replacing_ Noto Sans Arabic by the new one in:
https://github.com/koreader/koreader/blob/9c4e246038a30a405957b3f537c8c1c27d4a872b/frontend/document/credocument.lua#L47-L55

NiLuJe:
@Frenzie: This *is* front? ;o).

Frenzie:
@NiLuJe Ho, hum, so it is. >_>

PR #6244, opened 2020-06-08
Style tweaks: adds "Book specific tweak" menu item

poire-z:
Allows editing a CSS snippet to be applied to this book only, without the need to create and edit a User style tweak.
Allows copying any other tweak CSS by just taping on it (and pasting into this with Hold).
Limit User style tweaks nb of items per menu page to 6 (like we try to do for other tweaks menus).

(Dunno if there's a better wording than "Book specific tweak" to make that clearer... "Book own tweak" ?)

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6244)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -10,6 +10,7 @@ local TimeVal = require("ui/timeval")
 local Translator = require("ui/translator")
 local UIManager = require("ui/uimanager")
 local logger = require("logger")
+local util = require("util")
 local _ = require("gettext")
 local C_ = _.pgettext
 local T = require("ffi/util").template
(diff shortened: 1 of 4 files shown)


poire-z:
Btw, our keyboard popups are not super style-tweak friendly : you can't use a direct swipe from the key to add `{}`, you need to open the popups:

Dunno if there's a better layout possible - or if I should just ask my keyboard.

And the fact that I replace here <kbd>,</kbd> by <kbd>;</kbd> on the fr keyboard (and switch <kbd>,</kbd> and <kbd>;</kbd> in the popup) makes these keys more confusing than they are in english...

(Another issue caused by me is that fr_keyboard.lua hacks the en popups - in a way the hack stays when switching to the en_keyboard :/ )

hius07:
> Allows editing a CSS snippet to be applied to this book only, without the need to create and edit a User style tweak.

Is it the same as if we have a checked User style tweak without the "star" (which is applied to the current book only)?
To help creating a new user style tweak (for this or all books), can we have a new item in the User style tweaks menu: "Create new style tweak"? (instead of Book specific tweak at all)

Frenzie:
> Btw, our keyboard popups are not super style-tweak friendly : you can't use a direct swipe from the key to add {}, you need to open the popups:

I believe I did that because I reasoned they'd be rarely if ever used in "normal" text. I may have clarified in the relevant PR/commit message. However, they're also available on page 2 of the symbols.

Screenshot_2020-06-08_10-42-07

poire-z:
> Is it the same as if we have a checked User style tweak without the "star" (which is applied to the current book only)?

No.

> To help creating a new user style tweak (for this or all books), can we have a new (first) item in the User style tweaks menu: "Create new style tweak" which will call the editor as above?

Well, I hope we don't need that now.

You can create User style tweaks on the device by using the TextEditor plugin.
Before this PR, it's what I did: TextEditor, create (or edit my existing tmp.css tweak) - then restart KOReader (so it sees the new style tweak), and apply it.

With this PR, you don't need to have a new file created in styletweaks/, and have it polluted with book specific files (or have a tmp.css 30 lines long from past tweaks that I prefered keeping in case I needed some bits for a new book: I just had to update the classnames - and sometimes having unwanted old stuff I kept there applied to that book).
You just hit "Book specific tweak", the editor is launched, you type CSS bits, and it is saved in the book metadata.epub.lua - and you can toggle if off and on with the checkbox like other tweaks.

I hope I don't have to add `User style tweak > Create user style tweak` because:
- it's quite some work to get the editor and workflow right (ask to give a name to the new CSS tweak file, check if existing, repopulate the menu, allow deleting it...)
- it might not be that useful now that we have this Book specific style tweak
- if you have the idea that the CSS you thought of should be re-usable and deserve to be a User style tweak, well, it's a bit more taps, but just use the exising TextEditor plugin, as I/we did until now :)

> Book-specific tweak 

I'll go with that, the `-` helps limiting ambiguity :)

hius07:
> and it is saved in the book metadata.epub.lua

OK, I see it's the main difference, thanks.

PR #6247, opened 2020-06-09
handle newlines in exported pdf highlights

### LINKED ISSUE #6045: Exporting highlights from PDF - problem with the newlines
* KOReader version: the last one (2020.03.2)
* Device: (reMarkable; I remember a similar issue with a Kindle in the past )

#### Issue

When I highlight text in a PDF file, it seems that the break lines are ignored and so the word ending each line and the one starting the next line are recorded as a single word with no blank space between them. 

This error can be noticed already in the metadata.pdf.lua file within the .sdr directory.
The error is also shown when exporting the notes with the Evernote plugin (no matter if you use the option to export as HTML or as a text file, in both cases the error can be seen).

At first I thought that the problem could be the way the newline was encoded in the output file, but it seems that this is not the case (I cannot see any special character between the two words).
The problem is not either related to the PDF formatting, as copying and pasting from the PDF to a text file keeps the break lines as expected.

To reproduce the problem: this seems to happen with any PDF file. As an example, you can try with the one available at http://ceur-ws.org/Vol-1756/paper06.pdf. For example try to highlight the first paragraph of the Introduction section and then export the notes.

The problem seems to arise in any condition (no zoom, no reflow, etc.).

pazos:
Fixes #6045 

Implemented @sr1920 suggestion at https://github.com/koreader/koreader/issues/6045#issuecomment-633215253

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6247)

### DIFF
--- a/frontend/document/koptinterface.lua
+++ b/frontend/document/koptinterface.lua
@@ -828,7 +828,7 @@ function KoptInterface:getTextFromBoxes(boxes, pos0, pos1)
         end
         -- append a space at the end of the line unless its a hyphenated word
         line_text = line_text .. " "
-        line_text = line_text:gsub("- ", "")
+        line_text = line_text:gsub("- $", "")
         -- insert line box
         local lb = boxes[i]
         if i > i_start and i < i_stop then
--- a/frontend/document/koptinterface.lua
+++ b/frontend/document/koptinterface.lua
@@ -826,6 +826,9 @@ function KoptInterface:getTextFromBoxes(boxes, pos0, pos1)
                 line_text = line_text..word..space
             end
         end
+        -- append a space at the end of the line unless its a hyphenated word
+        line_text = line_text .. " "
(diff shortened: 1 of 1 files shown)


frontend/document/koptinterface.lua
```diff
@@ -826,6 +826,9 @@ function KoptInterface:getTextFromBoxes(boxes, pos0, pos1)
                 line_text = line_text..word..space
             end
         end
+        -- append a space at the end of the line unless its a hyphenated word
+        line_text = line_text .. " "
+        line_text = line_text:gsub("- ", "")
```
poire-z, line 831: This would remove `"- "` in the middle of line_text. Should you match that only at end of line_text?

Frenzie: You're quite right of course. :facepalm:  Not the biggest issue in most text I suppose but it should be `- $`.

PR #6253, opened 2020-06-10
Add the possibility to execute a shell script on startup

zwim:
Add the possibility to execute the shell script
    apk_install_directory/files/scripts/startup-script.sh at startup of koreader.
    
    The startup-scripts.sh checks, if it is the first startup of koreader after an installation.
    If so, two the two directorys sdcard/koreader/scripts.[once,update]
    are created and populated with working example scripts,  but only if
    the directories did not exist.
    
    On the first startup all scripts in the sdcard/koreader/scripts.once
    (koreader user-directory) are executed.
    
    On every startup all scripts in sdcard/koreader/scripts.update are executed.
    
    The scripts do have full to access the installation directory and the
    user directory of koreader.
    ........
    
    The example scripts allow to update the hyphenation patterns from user supplied
    patterns in /sdcard/hyph/*.pattern on a new installation of koreader, or if
    the user patterns are newer than the installed ones.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6253)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,8 @@ endif
 ifdef ANDROID
 	cd $(INSTALL_DIR)/koreader && \
 		ln -sf ../../$(ANDROID_DIR)/*.lua .
+	@echo "[*] Install scripts"
+	$(RCP) -pL scripts $(INSTALL_DIR)/koreader/
 endif
 ifdef WIN32
 	@echo "[*] Install runtime libraries for win32..."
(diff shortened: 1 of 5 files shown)


zwim:
Yes, I will let sit the whole thing for now. In the meantime I try to get familiar with lua.

zwim:
I am closing this now. 
Thanks for everyone helping.
I try to provide a better solution soon.

PR #6255, opened 2020-06-10
Feature/characters

smartscripts-nl:
With this plugin it is possible to use selected text in an ebook for describing characters in the ebook:

1) When you want to describe a character, select his or her name. Activate the plugin (through a button defined in readerhighlight.lua, see below). An edit dialog with the selected text will be opened. Add a description to the name and save the character.

2) Later on text selections which contain (part of) the name of the character can be used to lookup the character, via the same route. When part of the name is found in the table with characters, the description will be shown, in stead of opening an edit dialog.

Handy for ebooks with lots of characters in them.

The plugin uses a helper widget, placed in ui/extensions. Maybe not the best of places, can of course be put somewhere else.

To use this plugin in my personal KOReader I've adapted readerhighlight.lua, to implement a button for defining/looking up characters:

{
                text = 'Character',
                callback = function()

                    self:highlightFromHoldPos()

                    if self.selected_text then

                        self.ui:handleEvent(Event:new("SaveCharacter", self.selected_text.text))

                        self:onClearHighlight()

                        UIManager:close(self.highlight_dialog)
                    end
                end,
            },

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6255)

### DIFF
--- a/frontend/extensions/helpers.lua
+++ b/frontend/extensions/helpers.lua
@@ -1,0 +1,275 @@
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local Screen = require("device").screen
+
+local SQ3 = require("lua-ljsqlite3/init")
+local DataStorage = require("datastorage")
+local LuaSettings = require("luasettings")
+local UIManager = require("ui/uimanager")
+local InfoMessage = require("ui/widget/infomessage")
+local Font = require("ui/font")
+local ConfirmBox = require("ui/widget/confirmbox")
+local _ = require("gettext")
+local lfs = require("libs/libkoreader-lfs")
+
+-- Plugin for global methods
+-- Some modules don't use this, but communicate directly with each other by exchanging global variables by triggering events. For an example, search for variable AX_remaining_time.
+
+-- for widgets use :extend instead of :new ?:
+local Helpers = WidgetContainer:extend{
+
+	debug_once_used = false,
+
+	file_paths = {},
+
+	garbage = nil,
+}
+
+function Helpers:init(ui)
+	self.ui = ui
+end
+
+function Helpers:getHomeDir()
+
+	local homedir = G_reader_settings:readSetting("home_dir")
+
+	local current_ebook
+
+	if homedir == nil then
+		current_ebook = G_reader_settings:readSetting("lastfile")
+
+		if current_ebook ~= nil then
+			homedir = string.gsub(current_ebook, '/[- .,:;?!_A-Za-z0-9]+', '')
+		end
+    end
+
+	return homedir
+end
+
+function Helpers:getKoreaderDir()
+
+    local settings_dir = DataStorage:getSettingsDir()
+
+    return string.gsub(settings_dir, '/settings', '')
+end
+
+-- use timeout = nil for second argument when calling for no timeout:
+function Helpers:alertDebug(message, timeout, dismiss_callback)
+
+	if dismiss_callback == nil then
+		UIManager:show(InfoMessage:new { text = message, icon_file = 'resources/info-bug.png', timeout = timeout })
+
+	else
+		UIManager:show(InfoMessage:new { text = message, icon_file = 'resources/info-bug.png', timeout = timeout, dismiss_callback = dismiss_callback })
+	end
+end
+
+function Helpers:alertDebugOnce(message, timeout, dismiss_callback)
+
+	if not self.debug_once_used then
+
+		self:alertDebug(message, timeout, dismiss_callback)
+
+		self.debug_once_used = true
+	end
+end
+
+-- use timeout = nil for second argument when calling for no timeout:
+function Helpers:alertError(message, timeout, dismiss_callback)
+
+	if dismiss_callback == nil then
+		UIManager:show(InfoMessage:new { text = message, icon_file = 'resources/info-error.png', timeout = timeout })
+
+	else
+		UIManager:show(InfoMessage:new { text = message, icon_file = 'resources/info-error.png', timeout = timeout, dismiss_callback = dismiss_callback })
+	end
+end
+
+-- use timeout = nil for second argument when calling for no timeout:
+function Helpers:alertInfo(message, timeout, dismiss_callback)
+
+	if dismiss_callback == nil then
+		UIManager:show(InfoMessage:new { text = message, timeout = timeout })
+
+	else
+		UIManager:show(InfoMessage:new { text = message, timeout = timeout, dismiss_callback = dismiss_callback })
+	end
+end
+
+-- use timeout = nil for second argument when calling for no timeout:
+function Helpers:alertNoIcon(message, timeout)
+
+	UIManager:show(InfoMessage:new { text = message, show_icon = false, timeout = timeout })
+end
+
+function Helpers:shuffle(tbl)
+	for i = #tbl, 2, -1 do
+		local j = math.random(i)
+		tbl[i], tbl[j] = tbl[j], tbl[i]
+	end
+end
+
+function Helpers:split(str, pat)
+	local t = {}  -- NOTE: use {n = 0} in Lua-5.0
+	local fpat = "(.-)" .. pat
+	local last_end = 1
+	local s, e, cap = str:find(fpat, 1)
+	while s do
+		if s ~= 1 or cap ~= "" then
+			table.insert(t, cap)
+		end
+		last_end = e + 1
+		s, e, cap = str:find(fpat, last_end)
+	end
+	if last_end <= #str then
+		cap = str:sub(last_end)
+		table.insert(t, cap)
+	end
+	return t
+end
+
+function Helpers:split_path(str)
+	return self:split(str, '[\\/]+')
+end
+
+function Helpers:table_sort_by_text_prop(subject)
+
+	table.sort(subject, function(v1, v2)
+		return v1.text < v2.text
+	end)
+end
+
+function Helpers:table_sort_alphabetically(subject)
+
+	table.sort(subject, function(v1, v2)
+		return v1 < v2
+	end)
+end
+
+function Helpers:ucfirst(str)
+	return (str:gsub("^%l", string.upper))
+end
+
+function Helpers:getMiddleDialogWidth()
+
+	local orientation = Screen:getScreenMode()
+	local iwidth = (Screen:getWidth() / 2) + 20
+
+	if orientation == 'portrait' then
+		iwidth = Screen:getWidth() - 80
+	end
+
+	return iwidth
+end
+
+function Helpers:getWideDialogWidth()
+
+	local orientation = Screen:getScreenMode()
+	local iwidth = Screen:getWidth() - 120
+
+	if orientation == 'portrait' then
+		iwidth = Screen:getWidth() - 80
+	end
+
+	return iwidth
+end
+
+function Helpers:round(x)
+
+	if x % 2 ~= 0.5 then
+		return math.floor(x + 0.5)
+	end
+
+	return x - 0.5
+end
+
+function Helpers:substrCount(subject, needle)
+	return select(2, subject:gsub(needle, ""))
+end
+
+function Helpers:readSetting(options_file, setting, default)
+
+	if not string.match(options_file, '[.]lua$') then
+		options_file = options_file .. '.lua'
+	end
+
+	local path = self.file_paths[options_file] or DataStorage:getSettingsDir() .. "/" .. options_file
+
+	self.file_paths[options_file] = path
+
+	local file_handle = LuaSettings:open(path)
+
+	return file_handle:readSetting(setting) or default
+end
+
+function Helpers:saveSetting(options_file, setting, value)
+
+	if not string.match(options_file, '[.]lua$') then
+		options_file = options_file .. '.lua'
+	end
+
+	local path = self.file_paths[options_file] or DataStorage:getSettingsDir() .. "/" .. options_file
+
+	self.file_paths[options_file] = path
+
+	local options_file_handle = LuaSettings:open(path)
+
+	options_file_handle:saveSetting(setting, value)
+	options_file_handle:flush()
+end
+
+function Helpers:listItemNumber(nr, title)
+
+	return tostring(nr) .. '. ' .. self:removeListItemNumber(title)
+end
+
+function Helpers:removeListItemNumber(text)
+
+	return string.gsub(text, '^[0-9]+[.][ ]*', '')
+end
+
+function Helpers:confirm(question, callback)
+
+	UIManager:show(ConfirmBox:new {
+		text = question,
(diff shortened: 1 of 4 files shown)


poire-z:
> The plugin uses a helper widget, placed in ui/extensions. Maybe not the best of places, can of course be put somewhere else.

Really not fond of that :/
You're introducing a 2nd API (that mostly just wrap with zero or only a few added lines our functions and UI, and Lua function/string and table methods), like if you were not happy with how Lua and us present stuff, and thought: I'll rewrite the function calls and names in my own words and way of thinking :)

_You_ should speak our language :) - and we should not encourage people to learn another - and we should not have to maintain a duplicated API. If you think some function and utils are missing, you should add them to our relevant places and modules, so people can find them where they would expect it - and they don't go pick a bit in our core, and a bit (the same, with another name :) in your extensions/helpers.lua...

As for the plugin, I'm not a customer for that - but I feel it might be too specific.
A more generic plugin would allow recording/remembering any "topics", "words", "references", "bibliography", which would allow it to be used in other contexts/workflows.
Or tomorrow, somebody will add "notions.plugin", "references.plugin", "years.plugin", "clues.plugin" (useful in crime novels! :)...

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+
+		if current_ebook ~= nil then
+			homedir = string.gsub(current_ebook, '/[- .,:;?!_A-Za-z0-9]+', '')
+		end
+    end
+
+	return homedir
+end
+
+function Helpers:getKoreaderDir()
```
Frenzie, line 48: I agree with @poire-z that there's a lot of stuff in this file that doesn't really seem to fit. This one for example is doing the reverse of `DataStorage:getSettingsDir()`, which is just `DataStorage:getDataDir()` + `/settings`/ I believe?

smartscripts-nl: Maybe you're right. I'll look it up. I've learned lua by looking at the KOReader source code, so my code will not be always optimal.

smartscripts-nl:
OK, just testing the water. I've also written a plugin which saves texts for inspirational quotes, which can be viewed and, if one wishes, displayed automatically every quarter hour. But from your reactions I gather you won't be enthusiastic about that one too.

I've written many other things, for example an improvement (imho) of the file search script. Maybe I'll test the water some more in the time to come.

smartscripts-nl:
NB: in the code you can see that I sometimes repaint a dialog, when showing it. I don't know if my method is the best method to go about that, but I needed this behaviour to fix a problem on my Kobo Forma: the borders of dialogs are not always painted completely, parts of them are often missing. When I repaint the same dialog, the problem is gone.

poire-z:
> the borders of dialogs are not always painted completely, parts of them are often missing. When I repaint the same dialog, the problem is gone.

Yeah, there's quite a bit of nastiness here and there :)
https://github.com/smartscripts-nl/koreader/blob/a8d270a15da3cdadb278cc0690f06e6f8c2a0af5/plugins/characters.koplugin/main.lua#L328-L342
(One minor thing that makes it hard to get into your code is all the blank lines.)

> I've learned lua by looking at the KOReader source code

Like most of us did :)
But there's Lua to learn, and there's KOReader's way of doing things (which is harder than Lua). You usually learn that by copying stuff from things that work, getting to understand the magic, and adapting it.
So, you picked the `UIManager:show()` which looked simple enough - but there's the whole business on refresh (that _is_ complicated, more complicated than other UI framework, because we target eInk panels) that you didn't pick: the `setDirty()` that is often set after showing stuff, to indicate the eInk driver which region to repaint. I guess that's something like that that is missing.
Your show/close/show double repaint solves that, I guess, by the fact that when closing, you're having the lower widgets asked to repaint themselves, some of them take the fullscreen, so it does `setDirty(the whole screen)`, then you show your little widget - and our UIManager enqueue/merge the setDirty regions to do a single refresh at the end (because eInk is slow) - so your little widget that you show() a 2nd time before the refresh is asked benefit from the fullscreen refresh - and is painted fully.

smartscripts-nl:
> > The plugin uses a helper widget, placed in ui/extensions. Maybe not the best of places, can of course be put somewhere else.
> 
> Really not fond of that :/
> You're introducing a 2nd API (that mostly just wrap with zero or only a few added lines our functions and UI, and Lua function/string and table methods), like if you were not happy with how Lua and us present stuff, and thought: I'll rewrite the function calls and names in my own words and way of thinking :)
> 
> _You_ should speak our language :) - and we should not encourage people to learn another - and we should not have to maintain a duplicated API. If you think some function and utils are missing, you should add them to our relevant places and modules, so people can find them where they would expect it - and they don't go pick a bit in our core, and a bit (the same, with another name :) in your extensions/helpers.lua...
> 
> As for the plugin, I'm not a customer for that - but I feel it might be too specific.
> A more generic plugin would allow recording/remembering any "topics", "words", "references", "bibliography", which would allow it to be used in other contexts/workflows.
> Or tomorrow, somebody will add "notions.plugin", "references.plugin", "years.plugin", "clues.plugin" (useful in crime novels! :)...

OK, I was just testing the waters. It seems I don't fit in with your team. But keep up the good works!

Frenzie:
Don't give up, I'll try to give some feedback in a more concrete manner. ;-)

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+end
+
+function Helpers:shuffle(tbl)
+	for i = #tbl, 2, -1 do
+		local j = math.random(i)
+		tbl[i], tbl[j] = tbl[j], tbl[i]
+	end
+end
+
+function Helpers:split(str, pat)
```
Frenzie, line 111: This seems to be used solely as a helper function for split_path? For split in general, see http://koreader.rocks/doc/modules/util.html#gsplit

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+		s, e, cap = str:find(fpat, last_end)
+	end
+	if last_end <= #str then
+		cap = str:sub(last_end)
+		table.insert(t, cap)
+	end
+	return t
+end
+
+function Helpers:split_path(str)
```
Frenzie, line 130: Not quite the same thing as http://koreader.rocks/doc/modules/util.html#splitFilePathName but without seeing what it's used for it's hard to say one way or the other.

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+end
+
+function Helpers:ucfirst(str)
+	return (str:gsub("^%l", string.upper))
+end
+
+function Helpers:getMiddleDialogWidth()
+
+	local orientation = Screen:getScreenMode()
+	local iwidth = (Screen:getWidth() / 2) + 20
```
Frenzie, line 155: More of a side note, but you'll generally want to wrap any numbers like that in `Screen:scaleBySize()`.

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+	local iwidth = Screen:getWidth() - 120
+
+	if orientation == 'portrait' then
+		iwidth = Screen:getWidth() - 80
+	end
+
+	return iwidth
+end
+
+function Helpers:round(x)
```
Frenzie, line 176: We actually have a bunch of math helpers like that in optmath.

I'll add some comment strings in so they'll show up in the docs.

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+
+	if x % 2 ~= 0.5 then
```
Frenzie, line 178: As an aside, this is not how I round. 1.5 rounds to 2, not to 1.

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+	end
+
+	return x - 0.5
+end
+
+function Helpers:substrCount(subject, needle)
+	return select(2, subject:gsub(needle, ""))
+end
+
+function Helpers:readSetting(options_file, setting, default)
```
Frenzie, line 189: At first glance this looks like needless duplication, but if it's not it should probably be fixed in LuaSettings.

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+function Helpers:removeListItemNumber(text)
+
+	return string.gsub(text, '^[0-9]+[.][ ]*', '')
+end
+
+function Helpers:confirm(question, callback)
+
+	UIManager:show(ConfirmBox:new {
+		text = question,
+		ok_text = _("Yes"),
```
Frenzie, line 234: Note that you can just leave those empty for the default values.

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+		cancel_text = _("No"),
+		ok_callback = function()
+			callback()
+		end,
+	})
+end
+
+function Helpers:exists(path)
```
Frenzie, line 242: http://koreader.rocks/doc/modules/util.html#pathExists

plugins/characters.koplugin/main.lua
```diff
@@ -0,0 +1,869 @@
...
+
+        self.filter = character
+
+        return {
+            character = item_with_most_hits['character'],
+            characterno = item_with_most_hits['characterno'],
+        }
+    end
+
+    return nil
```
Frenzie, line 215: Not very important, but you can just leave that out.

smartscripts-nl:
The problem is maybe that I am a PHP programmer and I think Lua is sometimes rather clunky. So I like to write wrapper methods which to my taste (biased by PHP) are easier to "read" and to use.

The methods in helpers.lua can and should of course be placed somewhere else (if handy at all), but for my personal development work it is easier to keep them all together in one place.

frontend/extensions/helpers.lua
```diff
@@ -0,0 +1,275 @@
...
+	else
+		UIManager:show(InfoMessage:new { text = message, icon_file = 'resources/info-bug.png', timeout = timeout, dismiss_callback = dismiss_callback })
+	end
+end
+
+function Helpers:alertDebugOnce(message, timeout, dismiss_callback)
+
+	if not self.debug_once_used then
+
+		self:alertDebug(message, timeout, dismiss_callback)
```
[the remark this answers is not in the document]
smartscripts-nl: When I want to view values or until how far a script proceeds without errors, on my computer I insert temporary calls to alertDebug of alertDebugOnce, so I can see onscreen in KOReader itself how things work out. Completely developing on the Kobo itself would be very impractical.

smartscripts-nl:
> One minor thing that makes it hard to get into your code is all the blank lines

Funny, to my taste the blank lines make the code more readable, yield better oversight... One can't argue about taste, but should I contribute code in the future, then I'll remove the blank lines.

frontend/extensions/helpers.lua
smartscripts-nl: When you run the code, you'll see that it rounds 1.5 nicely to 2:
 1.5 % 2 = 0 => so not 0.5, so return math.floor(1.5 + 0.5) => 2

Frenzie: 1.5 mod 2 = 1.5, so you return 1.5 - 0.5 = 1

0.7 mod 2 = 0.7, so you return 0.7 - 0.5 = 0.2

Frenzie: Wait, you and/or I confused me, but this is how your function behaves pasted into a Lua console. >_>

```
> print (Helpers:round(1.5))
1
> print (Helpers:round(2))
1.5
> print (Helpers:round(2.5))
2
> print (Helpers:round(0.7))
0.2
> print (Helpers:round(2.5))
2
> print (Helpers:round(1.5))
1
```

In any case the exception makes no sense.

```
function round(x)

	if x % 2 ~= 0.5 then
		return math.floor(x + 0.5)
	end

	return x - 0.5
end

print(round(.5))
print(round(1.5))
```
```
$ lua bla.lua 
0
2
```

NiLuJe: Don't we already have a round in optmath?

NiLuJe: (Which should essentially be doing floor + 0.5).

Frenzie: We do. GH may have helpfully hidden my comment saying exactly that right above. ;-)

NiLuJe: Why, yes we do, and yes it does ;).

https://github.com/koreader/koreader/blob/23862fd0ebd51f8a21ed4dd82861806f356f6128/frontend/optmath.lua#L43

Frenzie: This one: https://github.com/koreader/koreader/pull/6255#discussion_r438648691

NiLuJe: Oh, actually, it might be wrong. Does floor(n + 0.5) == trunc(floor(n) + 0.5)?

Frenzie: I don't think trunc is a thing in Lua, but that's probably the same thing. Why is that wrong?

NiLuJe: I dunno, but in C I'd have done floor(n) + 0.5f cast to an integer (so, truncated).

Frenzie: My bad, I was assuming trunc(n + 0.5) as trunc(floor(n) + 0.5) looks completely meaningless to me.

Frenzie: As in:

floor(1.9) = 1
trunc(1.9) = 1
trunc(1 + 0.5) = 1

Frenzie: It'll be different for negative integers of course.

NiLuJe: Yeah, no, I still suck at maths ;D.

floor(n) + 0.5f essentially does a ceil, not a round.

TL;DR: Everything's fine :D.

NiLuJe: Never mind me, I still suck at maths ;D.

(i.e., floor(n + 0.5) works just fine ;)).

NiLuJe: And I was actually thinking of `(int) (n + 0.5f)`, and not the nonsense I mentioned earlier ^^.

NiLuJe: (So, basically, yeah, floor() :D).

Frenzie: Difference illustrated in JS (roundToZero would be a much clearer name, cf. our [roundAwayFromZero](http://koreader.rocks/doc/modules/optmath.html#Math.roundAwayFromZero)):
```
Math.floor(-1.9)
-2
Math.trunc(-1.9)
-1
```

PR #6257, opened 2020-06-11
A few footer tweaks

NiLuJe:
* Refresh the footer instantly when changing the prefix, separator, or formatting.
  (Either that never worked, or it's an old-ish regression).
* Allow setting the container height (formerly DMINIBAR_CONTAINER_HEIGHT)
* Allow setting the container bottom padding (formerly 1 flex pixel)
* Refactor ReaderFooter:applyFooterMode
  Handle toggling the genFooterText function even in all at once mode
  Make sure the layout is properly reset when changing visibility state, to avoid inconsistencies with complex layouts.
* Allow taller progress bars in the footer (max = default * 4)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6257)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -385,14 +385,12 @@ function ReaderFooter:init()
     -- Container settings
     if not self.settings.container_height then
         self.settings.container_height = DMINIBAR_CONTAINER_HEIGHT
-    else
-        self.height = Screen:scaleBySize(self.settings.container_height)
     end
+    self.height = Screen:scaleBySize(self.settings.container_height)
     if not self.settings.container_bottom_padding then
         self.settings.container_bottom_padding = 1 -- unscaled_size_check: ignore
-    else
-        self.bottom_padding = Screen:scaleBySize(self.settings.container_bottom_padding)
     end
+    self.bottom_padding = Screen:scaleBySize(self.settings.container_bottom_padding)
 
     -- default margin (like self.horizontal_margin)
     if not self.settings.progress_margin_width  then
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -300,7 +300,7 @@ local ReaderFooter = WidgetContainer:extend{
     -- NOTE: Used for vertical spans by the VerticalGroup for multi-line footers.
     --       Needs to be separate from the actual bottom_padding, which is user-configurable.
     stock_bottom_padding = Screen:scaleBySize(1),
-    bottom_padding = stock_bottom_padding,
+    bottom_padding = Screen:scaleBySize(1),
     settings = {},
     -- added to expose them to unit tests
     textGeneratorMap = footerTextGeneratorMap,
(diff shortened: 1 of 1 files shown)


NiLuJe:
The intended use-case being having a multi-line "all at once" setup with a bit more breathing room, that can be use as a quick "status panel" peek, while otherwise reading with it hidden.

i.e.,

bottom_status

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -374,6 +379,18 @@ function ReaderFooter:init()
...
+    if not self.settings.container_height then
+        self.settings.container_height = DMINIBAR_CONTAINER_HEIGHT
+    else
+        self.height = Screen:scaleBySize(self.settings.container_height)
+    end
+    if not self.settings.container_bottom_padding then
+        self.settings.container_bottom_padding = 1 -- unscaled_size_check: ignore
+    else
+        self.bottom_padding = Screen:scaleBySize(self.settings.container_bottom_padding)
+    end
```
poire-z, line 392: The `if else` logic seems a bit odd (on first exec, if feels like you forgot to set self.height and self.bottom_padding, but ok, you're using the initial values - but you set them in settings - on next exec, you'll use them from settings and are computing them). It works, but it's a bit unexplicite (you have to look around to be sure it's not a bug).
I would just drop the else and do the computation in all cases, for clarity :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -928,6 +944,60 @@ function ReaderFooter:addToMainMenu(menu_items)
                     },
                 }
             },
+            {
+                text_func = function()
+                    return T(_("Container height (%1)"), self.settings.container_height)
```
poire-z, line 949: (Dunno if these 2 `Container something` would be better in their own submenu, `Container size`?)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1056,8 +1126,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                         end,
                         callback = function()
                             self.settings.item_prefix = "letters"
-                            self:updateFooter()
-                            UIManager:setDirty(nil, "ui")
+                            self:refreshFooter(true)
```
poire-z, line 1129: > Refresh the footer instantly when changing the prefix, separator, or formatting.
(Either that never worked, or it's an old-ish regression).

I noticed that too - I guess it was just forgotten when making the other stuff changes refresh the footer (migrating them to use refreshFooter())

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: Fair enough ;).

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: None of the other stuff is in a submenu (see all the "change <stuff> formatting" at the end), so I opted for spacers between "logical" group of settings (whatever is at the beginning of the menu, spacer, dimension stuff, spacer, formatting stuff).

PR #6259, opened 2020-06-11
Feature/save location before goto next or previous bookmark

smartscripts-nl:
Save location to stack before jumping to a bookmark. So you can easily navigate back after jumping to bookmarks.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6259)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -210,6 +210,7 @@ end
 
 function ReaderBookmark:gotoBookmark(pn_or_xp)
     if pn_or_xp then
+        self.ui.link:addCurrentLocationToStack()
         local event = self.ui.document.info.has_pages and "GotoPage" or "GotoXPointer"
         self.ui:handleEvent(Event:new(event, pn_or_xp))
     end
--- a/frontend/extensions/helpers.lua
+++ b/frontend/extensions/helpers.lua
@@ -1,0 +1,275 @@
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local Screen = require("device").screen
+
+local SQ3 = require("lua-ljsqlite3/init")
+local DataStorage = require("datastorage")
+local LuaSettings = require("luasettings")
+local UIManager = require("ui/uimanager")
+local InfoMessage = require("ui/widget/infomessage")
+local Font = require("ui/font")
+local ConfirmBox = require("ui/widget/confirmbox")
+local _ = require("gettext")
+local lfs = require("libs/libkoreader-lfs")
+
+-- Plugin for global methods
+-- Some modules don't use this, but communicate directly with each other by exchanging global variables by triggering events. For an example, search for variable AX_remaining_time.
+
+-- for widgets use :extend instead of :new ?:
+local Helpers = WidgetContainer:extend{
+
+	debug_once_used = false,
+
+	file_paths = {},
+
+	garbage = nil,
+}
+
+function Helpers:init(ui)
+	self.ui = ui
+end
+
+function Helpers:getHomeDir()
+
+	local homedir = G_reader_settings:readSetting("home_dir")
+
+	local current_ebook
+
+	if homedir == nil then
+		current_ebook = G_reader_settings:readSetting("lastfile")
+
+		if current_ebook ~= nil then
+			homedir = string.gsub(current_ebook, '/[- .,:;?!_A-Za-z0-9]+', '')
+		end
+    end
+
+	return homedir
+end
+
+function Helpers:getKoreaderDir()
+
+    local settings_dir = DataStorage:getSettingsDir()
+
+    return string.gsub(settings_dir, '/settings', '')
+end
+
+-- use timeout = nil for second argument when calling for no timeout:
+function Helpers:alertDebug(message, timeout, dismiss_callback)
+
+	if dismiss_callback == nil then
+		UIManager:show(InfoMessage:new { text = message, icon_file = 'resources/info-bug.png', timeout = timeout })
+
+	else
+		UIManager:show(InfoMessage:new { text = message, icon_file = 'resources/info-bug.png', timeout = timeout, dismiss_callback = dismiss_callback })
+	end
+end
+
+function Helpers:alertDebugOnce(message, timeout, dismiss_callback)
+
+	if not self.debug_once_used then
+
+		self:alertDebug(message, timeout, dismiss_callback)
+
+		self.debug_once_used = true
+	end
+end
+
+-- use timeout = nil for second argument when calling for no timeout:
+function Helpers:alertError(message, timeout, dismiss_callback)
+
+	if dismiss_callback == nil then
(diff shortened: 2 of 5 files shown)


smartscripts-nl:
> But please remove the two other commits. ;-)

That may be caused by my lack of experience with pull requests. I pushed a separate branch and from there created a PR. Why then are the other PR's I made still shown here? I thought they were separate, because on other branches.

poire-z:
You probably created your new branch while being on that other branch. You should have rewind back to master (with `git checkout master`) to start from our master base.

smartscripts-nl:
> You probably created your new branch while being on that other branch. You should have rewind back to master (with `git checkout master`) to start from our master base.

OK, I'll try that out and then post a new PR.

Frenzie:
The PR is to merge the entire branch into another branch.

To fix it before creating a PR, generally speaking (maybe not worthwhile for something this small):

```sh
git checkout master
#new branch from master, not from some other branch
git checkout -b feat/save-location-bookmark 
# cherry pick the one commit we want into our new, proper branch
git cherry-pick f622f93fda5d78677377d1b84c12d9f0db05f6d4
```

But for something like this I'd probably do something like
```sh
git rebase -i HEAD~3
# comment out the two commits we don't want and rebase
# now force push to update the live branch
# NB force pushing overwrites everything, be careful with it
git push -f
```

smartscripts-nl:
> The PR is to merge the entire branch into another branch.
> 
> To fix it before creating a PR, generally speaking (maybe not worthwhile for something this small):

I understand. But when I create a new branch on master afterwards to use for a PR, won't I then encounter the same problem: multiple commits visible in the PR?
> 
> ```shell
> git checkout master
> # new branch from master, not from some other branch
> # you can do this all at once but atm I think it's easier/safer to separate out the steps
> git checkout -b feat/save-location-bookmark 
> # or: git checkout -b feat/save-location-bookmark master
> # cherry pick the commits we want into our new, proper branch
> git cherry-pick f622f93fda5d78677377d1b84c12d9f0db05f6d4
> ```
> 
> But for something like this I'd probably do something like
> 
> ```shell
> git rebase -i HEAD~3
> # comment out the two commits we don't want and rebase
> # now force push to update the live branch
> # NB force pushing overwrites everything, be careful with it
> git push -f
> ```

poire-z:
> That's not currently done? Sounds like a good fix.

It looks to me that it is working without this PR.
So, what's the use case where you need that?

Because, the 2 cases I can think of:
- Bookmark list, and taping on a bookmark, we already add the current location to stack (and this PR will add it a 2nd time):
https://github.com/koreader/koreader/blob/e3f978be2c3771d421268f7bae0ba720b35aacfa/frontend/apps/reader/modules/readerbookmark.lua#L266-L269
- Via the SkimTo widget, where we explicitely add only the page we were on, and not the other ones we reach when tapping (otherwise, we'll have dozens of swipe back to reach back our original place) - and I really want to keep that behaviour:
https://github.com/koreader/koreader/blob/e3f978be2c3771d421268f7bae0ba720b35aacfa/frontend/apps/reader/skimtowidget.lua#L351-L358

Other use case I can't think of ?

smartscripts-nl:
> > That's not currently done? Sounds like a good fix.
> 
> It looks to me that it is working without this PR.
> So, what's the use case where you need that?
> 
> Because, the 2 cases I can think of:
> 
> * Bookmark list, and taping on a bookmark, we already add the current location to stack (and this PR will add it a 2nd time):
>   https://github.com/koreader/koreader/blob/e3f978be2c3771d421268f7bae0ba720b35aacfa/frontend/apps/reader/modules/readerbookmark.lua#L266-L269
> * Via the SkimTo widget, where we explicitely add only the page we were on, and not the other ones we reach when tapping (otherwise, we'll have dozens of swipe back to reach back our original place) - and I really want to keep that behaviour:
>   https://github.com/koreader/koreader/blob/e3f978be2c3771d421268f7bae0ba720b35aacfa/frontend/apps/reader/skimtowidget.lua#L351-L358
> 
> Other use case I can't think of ?

When navigating to next or previous bookmark using a gesture.

Frenzie:
> won't I then encounter the same problem: multiple commits visible in the PR?

To be clear, multiple commits in and of themselves are not a problem, far from it, but they should be related to each other. So yes, if you have those multiple commits currently on master then you'd encounter an issue and you should create a new branch based on a specific commit prior to your changes. For example, like:

```sh
# base branch blah on commit from a few days ago
git checkout -b blah c1be488a11bc9efb3082b77affe990e189fbdbe6
```

poire-z:
> When navigating to next or previous bookmark using a gesture.

OK, so indeed, that's missing for that one.
So, with this PR, you put that `addCurrentLocationToStack()` too deep, while other upper bits of code (that will end up calling `gotoBookmark()`) already do it themselves.

I was going to suggest to put that in readergesture, just adding `self.ui.link:addCurrentLocationToStack()` before you send the GotoNextBookmarkFromPage/GotoPreviousBookmarkFromPage events:
https://github.com/koreader/koreader/blob/e3f978be2c3771d421268f7bae0ba720b35aacfa/frontend/apps/reader/modules/readergesture.lua#L1316-L1329

But it is a bit ugly, as the other stuff there like goToNextChapter do not have to do it, as their event handler does that itself - so it would be nice if we also did not have to do it.

So, possible more elegant solution:

Don't touch ReaderGesture, but do it like this:
```diff
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -613,7 +613,10 @@ function ReaderBookmark:onGotoNextBookmark(pn_or_xp)
     return true
 end

-function ReaderBookmark:onGotoNextBookmarkFromPage()
+function ReaderBookmark:onGotoNextBookmarkFromPage(add_current_location_to_stack)
+    if add_current_location_to_stack ~= false then -- nil or true
+        self.ui.link:addCurrentLocationToStack()
+    end
     self:gotoBookmark(self:getNextBookmarkedPageFromPage(self.ui:getCurrentPage()))
     return true
 end
diff --git a/frontend/apps/reader/skimtowidget.lua b/frontend/apps/reader/skimtowidget.lua
index f6927098..2a28a642 100644
--- a/frontend/apps/reader/skimtowidget.lua
+++ b/frontend/apps/reader/skimtowidget.lua
@@ -412,7 +412,8 @@ end
 function SkimToWidget:goToByEvent(event_name)
     if event_name then
         self:addOriginToLocationStack()
-        self.ui:handleEvent(Event:new(event_name))
+        self.ui:handleEvent(Event:new(event_name, false))
+            -- add_current_location_to_stack=false, as we managed that ourselves
         self.curr_page = self.ui:getCurrentPage()
         self:update()
     end
```
I haven't tested this quick attempt with a gesture (so please do :), but it works with the SkimToWidget. If it works for your use case, and this code makes sense, feel free to complete it for the GotoPrevious, and update this PR with it.

poire-z:
OK, PR'ed my suggestion in #6306.
(I considered pushing it to this PR, but given the noise brought in with the branch mismatch, it was easier to do it via an other PR.)

PR #6264, opened 2020-06-12
Floor dimension computations (mul/div).

NiLuJe:
This should essentially be a giant NOP, but might help with quirky off-by-ones due to accumulation of weird truncation on high-resolution/high-dpi devices.

Also bump base while we're here ;).

(https://github.com/koreader/koreader-base/pull/1113)
(https://github.com/koreader/koreader-base/pull/1114)
(https://github.com/koreader/koreader-base/pull/1115)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6264)


pazos:
Could you bump android-luajit-launcher to latest master too? Basically to stay in sync with LuaJIT

frontend/apps/reader/skimtowidget.lua
```diff
@@ -37,12 +37,12 @@ function SkimToWidget:init()
...
     self.screen_height = Screen:getHeight()
     self.span = math.ceil(self.screen_height * 0.01)
-    self.width = self.screen_width * 0.95
+    self.width = math.floor(self.screen_width * 0.95)
     self.button_bordersize = Size.border.button
     -- the buttons need some kind of separation but maybe I should just implement
     -- margin_left and margin_right…
     self.button_margin = self.button_bordersize
-    self.button_width = self.screen_width * 0.16 - (2*self.button_margin)
+    self.button_width = math.floor(self.screen_width * 0.16 - (2*self.button_margin))
```
poire-z, line 45: math.floor **(** self.screen_width * 0.16 **)** - (2*self.button_margin)
may be? (you did it that way below, which looks clearer)

frontend/apps/filemanager/filemanagersearch.lua
```diff
@@ -206,8 +206,8 @@ function Search:ShowSearch()
                     },
                 },
             },
-            width = Screen:getWidth() * 0.8,
-            height = Screen:getHeight() * 0.2,
+            width = math.floor(Screen:getWidth() * 0.8),
```
Frenzie, line 209: As an aside, I'm not sure if there's much rationale behind this one being 0.8 and many of those others being 0.95 or whatever.

frontend/apps/reader/skimtowidget.lua
NiLuJe: Right, I tried to keep as much precision as possible (so there may be a few cases where I only floor the complete equation *on purpose*), but I indeed missed that one!

frontend/apps/filemanager/filemanagersearch.lua
NiLuJe: True ;o).

That was vaguely what I was hoping to make clear bu using "constants" from somewhere, but as the intermediate commit explain, doing that while handling dimension changes & rotation would require greater change (e.g., moving that to a table in Screen (and not ui/size), which would be responsible for keeping it updated, and then accessing it via a getter to never use a local cached value).

NiLuJe:
FTR, this is *probably* why the actual refresh code has always been playing it safe:

https://github.com/koreader/koreader/blob/19b15462b209bdd6a12d44877dafbda2b6b418a2/frontend/ui/uimanager.lua#L1034-L1040

PR #6266, opened 2020-06-13
fix ANRs in Kosync plugin login/register

### LINKED ISSUE #6237: Can't login "Progress sync" on Android
* KOReader version: 2020.05
* Device: Android - LG G6

Whenever I hit login after filled my username & password in progress sync plugin, the android - Koreader app got freeze and then force close .
I did try with version 2020.04.01, but I encountered the same.

Please kindly take a look.

Thank you

pazos:
Fixes #6237

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6266)

### DIFF
--- a/plugins/kosync.koplugin/main.lua
+++ b/plugins/kosync.koplugin/main.lua
@@ -4,8 +4,7 @@ local InfoMessage = require("ui/widget/infomessage")
 local ConfirmBox = require("ui/widget/confirmbox")
 local NetworkMgr = require("ui/network/manager")
 local UIManager = require("ui/uimanager")
-local Screen = require("device").screen
-local DeviceModel = require("device").model
+local Device = require("device")
 local Event = require("ui/event")
 local Math = require("optmath")
 local DEBUG = require("dbg")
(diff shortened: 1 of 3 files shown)


plugins/kosync.koplugin/main.lua
```diff
@@ -384,6 +383,7 @@ function KOSync:doRegister(username, password)
         custom_url = self.kosync_custom_server,
         service_spec = self.path .. "/api.json"
     }
+    Device:setIgnoreInput(true)
```
poire-z, line 386: Might be worth adding a small comment to all of them - to avoid future readers to have to go look at that/why:
`-- on Android to avoid ANR (no-op on other platforms)`

plugins/kosync.koplugin/main.lua
```diff
@@ -361,8 +360,8 @@ function KOSync:login()
                 },
             },
         },
-        width = math.floor(Screen:getWidth() * 0.8),
-        height = math.floor(Screen:getHeight() * 0.4),
+        width = math.floor(Device.screen:getWidth() * 0.8),
```
Frenzie, line 363: Could you stick with the general coding style? ^_^

local Screen = Device.screen

(less sure about that DeviceModel thing, that seems like a one-off)

PR #6267, opened 2020-06-13
Fix partial HW dithered refreshes sometimes appearing to shift refreshed content

NiLuJe:
My best guess is it's an artifact caused by the PxP alignment constraints...

The EPDC should *already* be taking care of this (c.f., epdc_process_update @ drivers/video/fbdev/mxc/mxc_epdc_v2_fb.c), but, apparently, something's still not *quite* right  ;).

TL;DR: If we request a HW dithered refresh, grow the rectangle ourselves until the alignment constraints are met.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6267)

### DIFF
--- a/frontend/apps/reader/modules/readercropping.lua
+++ b/frontend/apps/reader/modules/readercropping.lua
@@ -112,7 +112,7 @@ function ReaderCropping:onPageCrop(mode)
         self.ui:handleEvent(Event:new("SetZoomMode", "page", "cropping"))
     end
     self.ui:handleEvent(Event:new("SetDimensions",
-        Geom:new{w = Screen:getWidth(), h = Screen:getHeight()*11/12})
+        Geom:new{w = Screen:getWidth(), h = math.floor(Screen:getHeight()*11/12)})
     )
     self.bbox_widget = BBoxWidget:new{
         crop = self,
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -154,7 +154,7 @@ function ReaderFont:onShowFontMenu()
         title = self.font_menu_title,
         item_table = self.face_table,
         width = Screen:getWidth() - 100,
-        height = Screen:getHeight() / 2,
+        height = math.floor(Screen:getHeight() * 0.5),
         single_line = true,
         perpage_custom = 8,
     }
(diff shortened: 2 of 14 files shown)


NiLuJe:
Also bump base for https://github.com/koreader/koreader-base/pull/1116

NiLuJe:
And get rid of the legacy rectangle fixup mentioned in #6264. It shouldn't be necessary, and, if we ever hit an issue, I'd rather fix the root cause.

(e.g., ConfigMenu has always been painted/ asking for refresh off-screen on the vertical axis, for some reason).

NiLuJe:
Oh, an easy reproduction for the issue, on a Forma, was a Spin Widget (I was using the Text Editor's font size setting, but any SPin Widget should do) on a dithered FM page (i.e., one with covers shown).

The initial refresh of the SpinWidget is not dithered, but subsequent ones are, which made the shift obvious.

NiLuJe:
(I'd already noticed the shift myself when testing HW dithering in FBInk, so it is not a weird KOReader quirk, or specific to 8bpp).

poire-z:
> Looks math to me :)

Looks math to you too :)
The readerconfig change screws quite a bit the bottom panel, with toggle names like Embedded styles or the Margins ones getting now truncated.
```diff
-    local default_name_align_right = (max_option_name_width + default_option_hpadding + 2*padding_small) / Screen:getWidth()
+    local default_name_align_right = math.floor((max_option_name_width + default_option_hpadding + 2*padding_small) / Screen:getWidth())
```
I tried changing it to `math.ceil()`, but it got worse.
Cause it's not a size we're computing, but a ratio :)
https://github.com/koreader/koreader/blob/3e9b780111c45544c7bd124b2fc2ad21225ab63c/frontend/ui/widget/configdialog.lua#L219-L226
to later compute a width:
https://github.com/koreader/koreader/blob/3e9b780111c45544c7bd124b2fc2ad21225ab63c/frontend/ui/widget/configdialog.lua#L267

So, no flooring required.
(Haven't checked all your other math.floor'ing, but I don't remember having played with ratio elsewhere else than there.)

NiLuJe:
@poire-z: Gah. I know I noticed some other stuff that were explicitly named as being a ratio (so I skipped those ;p), but the variable name on this didn't make me think this one would ;).

Feel free to revrert that one, I won't be on my main box tonight ;).

PR #6273, opened 2020-06-16
Fix bottom menu item titles truncation

poire-z:
Revert bit from #6267 , see https://github.com/koreader/koreader/pull/6267#issuecomment-644903608.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6273)

### DIFF
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -220,7 +220,7 @@ function ConfigOption:init()
     -- They will carry the left default_option_hpadding, but the in-between
     -- one (and the right one) will be carried by the option items.
     -- (Both these variables are between 0 and 1 and represent a % of screen width)
-    local default_name_align_right = math.floor((max_option_name_width + default_option_hpadding + 2*padding_small) / Screen:getWidth())
+    local default_name_align_right = (max_option_name_width + default_option_hpadding + 2*padding_small) / Screen:getWidth()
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/configdialog.lua
```diff
@@ -220,7 +220,7 @@ function ConfigOption:init()
     -- They will carry the left default_option_hpadding, but the in-between
     -- one (and the right one) will be carried by the option items.
     -- (Both these variables are between 0 and 1 and represent a % of screen width)
-    local default_name_align_right = math.floor((max_option_name_width + default_option_hpadding + 2*padding_small) / Screen:getWidth())
+    local default_name_align_right = (max_option_name_width + default_option_hpadding + 2*padding_small) / Screen:getWidth()
```
Frenzie, line 223: Actually since it's a ratio or something maybe the variable name should be slightly changed?

poire-z: It could - but it spread up to into creoptions.lua and koptoptions.lua (`item_align_center = 1.0`) with:
https://github.com/koreader/koreader/blob/3e9b780111c45544c7bd124b2fc2ad21225ab63c/frontend/ui/widget/configdialog.lua#L242-L243
So it's a ratio that conveys a notion of alignment...
Would need a bit more attention than I can give this evening, to make that less confusing :)

PR #6283, opened 2020-06-19
More robust way to detect our own path on Kobo

NiLuJe:
Should take care of all the weird and interesting ways people manage to
find to break it...

NOTE: "$(dirname $(realpath "${0}"))" works, too, but I'm not sure if
really old devices ship with the realpath applet...

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6283)

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -2,7 +2,7 @@
 export LC_ALL="en_US.UTF-8"
 
 # working directory of koreader
-KOREADER_DIR="$(CDPATH= cd -- "$(dirname -- "$0")" && pwd -P)"
+KOREADER_DIR="$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd -P)"
 
 # we're always starting from our working directory
 cd "${KOREADER_DIR}" || exit
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -2,7 +2,7 @@
 export LC_ALL="en_US.UTF-8"
 
 # working directory of koreader
-KOREADER_DIR="${0%/*}"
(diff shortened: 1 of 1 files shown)


NiLuJe:
xref: https://www.mobileread.com/forums/showpost.php?p=4002319&postcount=19

platform/kobo/koreader.sh
```diff
@@ -2,7 +2,7 @@
 export LC_ALL="en_US.UTF-8"
 
 # working directory of koreader
-KOREADER_DIR="${0%/*}"
+KOREADER_DIR="$(CDPATH= cd -- "$(dirname -- "$0")" && pwd -P)"
```
Frenzie, line 5: I don't quite follow the logic here? Maybe it's too late. If I've got it right:

* Our koreader.sh here is probably `$0` (for example /mnt/onboard/.adds/koreader/koreader.sh)
* `dirname` makes that `/mnt/onboard/.adds/koreader`
* So only if those commands were successful, do `pwd -P`? Is there some kind of check missing here? Wouldn't that first part always be successful?

NiLuJe: The idea is to only print the (absolute) path (via `pwd`) if we can successfully reach it. Given the fact that we know the shell is going to be busybox ash and that no-one will probably ever be attempting to source us, there's not much reason for the cd to fail, but the idea was to be extremely defensive and catch degenerate command lines early ;).

We also doublez-check do that by aborting if we can't cd to `KOREADER_DIR` below.

NiLuJe: In the same vein, the `-P` is overkill, because vfat makes it highly unlikely for us to encounter a symlink ^^.

NiLuJe: c.f., https://stackoverflow.com/a/29835459

PR #6288, opened 2020-06-21
Add the possibility run shell scripts from filemanager on android

zwim:
Works on my Tolino Epos 2.

~~The emulator shows "Permission denied". This has to be fixed, but I don`t know where.~~ (This was just a misbehaving shebang)

Executes only shell scripts, python scripts are not allowed on Android.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6288)


frontend/apps/filemanager/filemanager.lua
```diff
@@ -47,6 +47,7 @@ local C_ = _.pgettext
 local Screen = Device.screen
 local T = require("ffi/util").template
 
+
```
Frenzie, line 50:
```suggestion
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -316,7 +320,12 @@ function FileManager:init()
                         }
                         UIManager:show(script_is_running_msg)
                         UIManager:scheduleIn(0.5, function()
-                            local rv = os.execute(BaseUtil.realpath(file))
+                            local rv = 0;
+                            if Device:isAndroid() then
+                                  rv = os.execute("sh " .. BaseUtil.realpath(file)) -- run by sh, because sdcard has no execute permissions
```
Frenzie, line 325:
```suggestion
                                rv = os.execute("sh " .. BaseUtil.realpath(file)) -- run by sh, because sdcard has no execute permissions
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -316,7 +320,12 @@ function FileManager:init()
...
+                            else
+                                  rv = os.execute(BaseUtil.realpath(file))
```
Frenzie, line 327:
```suggestion
                                rv = os.execute(BaseUtil.realpath(file))
```

frontend/apps/filemanager/filemanager.lua
```diff
@@ -301,7 +302,10 @@ function FileManager:init()
             },
         }
 
-        if not Device:isAndroid() and lfs.attributes(file, "mode") == "file" and util.isAllowedScript(file) then
+        -- checks if file is a python/shell script on non android devices
+        -- or if file is a shell script on android
+        if ( not Device:isAndroid() and lfs.attributes(file, "mode") == "file" and util.isAllowedScript(file) ) or
```
Frenzie, line 307: Wouldn't it be more logical to add the Android exception to util.isAllowedScript() instead of wherever it may be used?

Frenzie:
Pinging @NiLuJe who wrote #5804.

frontend/apps/filemanager/filemanager.lua
zwim: We could add an argument to util.isAllowedScript(). If this argument is nil then the function acts as yet. IF it is there, only the listed script-types deliver true? 
The if on lines 307-308 would get smoother.

NiLuJe: Yep, +1 on adding an `isAndroid` flag to `util.isAllowedScript` to drop python.

(I mean, you can pass it a table of supported formats if you like, but that'd likely be a tad overengineered ^^).

frontend/apps/filemanager/filemanager.lua
```diff
@@ -316,7 +319,12 @@ function FileManager:init()
                         }
                         UIManager:show(script_is_running_msg)
                         UIManager:scheduleIn(0.5, function()
-                            local rv = os.execute(BaseUtil.realpath(file))
+                            local rv;
+                            if Device:isAndroid() then
+                                rv = os.execute("sh " .. BaseUtil.realpath(file)) -- run by sh, because sdcard has no execute permissions
```
NiLuJe, line 324: `sh -c`

frontend/apps/filemanager/filemanager.lua
Frenzie: Passing it a table doesn't address my point, which is that you're defeating the purpose of util.isAllowedScript(). It's not the way lines 307 and 308 look that bothers me but what they do.

pazos:
> The emulator shows "Permission denied". This has to be fixed, but I don`t know where.

Shows or crashes completely?

zwim:
No crash. Koreader shows "The script returned a non-zero status code : 126!"
and the log shows
```
06/22/20-12:22:25 DEBUG _refresh: Enqueued partial update for region 0 0 540 720 w/ HW dithering
06/22/20-12:22:25 DEBUG refresh on physical rectangle 0 0 540 720
06/22/20-12:22:25 DEBUG BackgroundRunner: _execute() @  1592821345
06/22/20-12:22:25 DEBUG BackgroundRunnerWidget: start running @  1592821345
sh: 1: /home/moscht/work/koreader/test/test.sh: Permission denied
06/22/20-12:22:25 DEBUG close widget: table: 0x7f3bc2a952d0
06/22/20-12:22:25 DEBUG setDirty via a func from widget nil
06/22/20-12:22:25 DEBUG Lower widget table: 0x7f3bc292fcb8 was dithered, honoring the dithering hint
```

Frenzie:
Looks like it works for me. Are you sure it's not something you're doing inside the script?

Screenshot_1592821881

frontend/apps/filemanager/filemanager.lua
Frenzie: Nah, it's just `sh /path/to/script`, not `sh -c 'content of script'`. ^_^

frontend/apps/filemanager/filemanager.lua
```diff
@@ -316,7 +319,12 @@ function FileManager:init()
                         }
                         UIManager:show(script_is_running_msg)
                         UIManager:scheduleIn(0.5, function()
-                            local rv = os.execute(BaseUtil.realpath(file))
+                            local rv;
```
Frenzie, line 322:
```suggestion
                            local rv
```

zwim:
> Looks like it works for me. Are you sure it's not something you're doing inside the script? What do you mean by "the emulator," any particular version?

I mean ```./kodev run``` with emulator.

The scripts contents are
```
#!/system/bin/sh
echo "Script run"
```

Frenzie:
And your script is executable?

zwim:
> And your script is executable?
no, was not. 
Now I have set it to executable and the error code changed to 127
the log file now shows
```
sh: 1: /home/moscht/work/koreader/test/test.sh: not found
```
BTW, what does the "1:" mean, where does it come from?

Frenzie:
Is `./test.sh` actually successful on the command line? You're skipping some troubleshooting 101 there. ;-)

zwim:
On the command line:

 `./test.sh` shows: `bash: ./test.sh: /system/bin/sh: Defekter Interpreter: Datei oder Verzeichnis nicht gefunden`. That should be clear as I do not have /system/bin/sh on my linux machine.

`dash -c test.sh` and `bash -c test.sh` shows me the content of the actual directory (very strange).

`dash test.sh` and `bash test.sh` execute the script as expected.

pazos:
> No crash. Koreader shows "The script returned a non-zero status code : 126!"

That's good. Calling a shell script shouldn't crash by itself. Anything executed inside the shell script could crash the entire VM. There's no way to prevent this. For instance:

`mkdir -p /data/data/org.koreader.launcher/files/something-new` will crash the whole program.

Now, as some scripts can take more than 5 seconds to finish, you can consider using `Device:setIgnoreInput(true)` before calling the script and `Device:setIgnoreInput(false)` when the script returns. If not you can reproduce an ANR with the script:

`sleep 20`. Run it and press the touchscreen. 5 secons later android will trigger an ANR.

zwim:
That was new to me, thank you. I will add it.

frontend/util.lua
```diff
@@ -723,8 +723,15 @@ end
 --- Basically a helper method to check a specific list of file extensions.
 ---- @string filename
 ---- @treturn boolean
-function util.isAllowedScript(file)
+function util.isAllowedScript(file, has_restricted_scripts)
     local file_ext = string.lower(util.getFileNameSuffix(file))
+    if has_restricted_scripts ~=nil then
+        if file_ext == "sh" then
+          return true
```
Frenzie, line 730:
```suggestion
            return true
```

frontend/util.lua
```diff
@@ -723,8 +723,15 @@ end
...
+        else
+          return false
```
Frenzie, line 732:
```suggestion
            return false
```

Frenzie:
> My findings: If the shebang of the scripts is #!/system/bin/sh it works on the android device, but not in the emulator.
If the shebang is #!/bin/sh it works on the emulator but not on android.

@zwim So there's no issue, just a script that won't run. In this case I believe Android isn't technically broken (which it usually is), just annoying. /usr/bin/env isn't a thing there either iirc but last I checked that's not a POSIX requirement.

On a related note: https://www.in-ulm.de/~mascheck/various/shebang/

zwim:
It was just a misbehaving shebang combined with `sh -c`. No `-c` solves the problem here (on the emulator and on the android device).

pazos:
@zwim: while developing android stuff you might want to test things on a emulator that is `Device:isAndroid()` using Android Studio and AVD. Testing things on the `KOReader emulator` :tm: doesn't help.

It is also a good idea to test on various android emulators, because things change between android apis and the emulator has these changes enforced. Usually it is ok to test on a couple different APIs: I'm normally using API17 and API25. Most of the emulators are very broken. The ones that are not broken are: 17, 23, 25 and 28

Frenzie:
There are some tips here btw: https://github.com/koreader/koreader/wiki/Development-and-Debugging-on-Android

> 23

This morning it randomly crashed on me, so maybe it's also starting to suffer from decay. I thought 23+ was basically fine though? Admittedly I use exactly the ones you mentioned because I find them the most useful wrt API changes, so I suppose I wouldn't have noticed.

I generally go with 23 or 25 as "default," 17 as a quick check for "old" and 28 as a quick check for "new".

zwim:
I have tested the pull request on API 17. The script thing works there (independently of the shebang).

On 23 and 25 I have some problems with the emulator (has nothing to do with koreader). Please could someone test it there.

Frenzie:
Like I said, works fine. ;-)

frontend/util.lua
```diff
@@ -723,8 +723,15 @@ end
 --- Basically a helper method to check a specific list of file extensions.
 ---- @string filename
 ---- @treturn boolean
-function util.isAllowedScript(file)
+function util.isAllowedScript(file, has_restricted_scripts)
     local file_ext = string.lower(util.getFileNameSuffix(file))
+    if has_restricted_scripts ~=nil then
```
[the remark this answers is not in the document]
pazos: I would rename it to has_python and invert the logic

```lua
if lfs.attributes(file, "mode") == "file" and util.isAllowedScript(file, not Device:isAndroid())
```

```lua
function util.isAllowedScript(file, has_python)
    local file_ext = string.lower(util.getFileNameSuffix(file))
    if file_ext == "sh"
    or file_ext == "py" and has_python then
        return true
    else
        return false
    end
end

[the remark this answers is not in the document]
Comment (pazos) [reply frontend/util.lua]: Or even better

```lua
if lfs.attributes(file, "mode") == "file" and  Device:canExecuteScript(file)
```

### generic
```lua
function Device:canExecuteScript(file)
    local file_ext = string.lower(util.getFileNameSuffix(file))
    if file_ext == "sh" or file_ext == "py"  then
        return true
    end
end
```

### android
```lua
function Device:canExecuteScript(file)
    local file_ext = string.lower(util.getFileNameSuffix(file))
    if android.prop.flavor ~= "fdroid" and file_ext == "sh"  then
        return true
    end
end
```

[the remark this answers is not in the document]
zwim: I have implemented the second variant and tested it on Android.
I cannot test it on other devices, since I don't have them.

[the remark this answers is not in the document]
pazos: @zwim: you can test it on the emulator (not **android** emulator in this case :p). You need to copy the generic function somewhere below https://github.com/koreader/koreader/blob/master/frontend/device/generic/device.lua#L287 and just run the emulator with `./kodev run`. It should be able to run scripts (.py too).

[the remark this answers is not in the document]
zwim: Scripts work on `./kodev run` too.

frontend/apps/filemanager/filemanager.lua
```diff
@@ -301,7 +301,9 @@ function FileManager:init()
             },
         }
 
-        if not Device:isAndroid() and lfs.attributes(file, "mode") == "file" and util.isAllowedScript(file) then
+        -- checks if file is a python/shell script on non android devices
+        -- or if file is a shell script on android
```
pazos, line 305: comment no longer needed

frontend/device/android/device.lua
```diff
@@ -7,6 +7,7 @@ local lfs = require("libs/libkoreader-lfs")
 local logger = require("logger")
 local _ = require("gettext")
 local T = require("ffi/util").template
+local util = require("util")
```
pazos, line 10: util before _

frontend/device/generic/device.lua
```diff
@@ -6,6 +6,7 @@ This module defines stubs for common methods.
 
 local logger = require("logger")
 local _ = require("gettext")
+local util = require("util")
```
pazos, line 9: util before _

frontend/util.lua
```diff
@@ -723,8 +723,15 @@ end
 --- Basically a helper method to check a specific list of file extensions.
 ---- @string filename
 ---- @treturn boolean
-function util.isAllowedScript(file)
+function util.isAllowedScript(file, has_restricted_scripts)
```
pazos, line 726: remove the entire function. Move comment to generic `Device:canExecuteScript(file)`

zwim:
Please ignore the last commit, I have a typo in there. I will correct it.

zwim:
Now it should do.

frontend/device/android/device.lua
```diff
@@ -352,4 +353,15 @@ end
 android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
     android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
 
+--- Returns true if the file is a script we allow running
```
pazos, line 356: Not needed. The function `Device:canExecuteScript()` does the same on all devices. We override its behaviour on android but there's no need to explain what the function does again.

See for example `Device:exit()`. It is explained in https://github.com/koreader/koreader/blob/master/frontend/device/generic/device.lua#L337-L343 but not again in https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L347-L350 (because that is just a device specific implementation of the exit function).

Also, I didn't notice on last review. Could you move the function before the log in 353?

zwim: OK

poire-z:
Some little cleanup still needed after this, see https://github.com/koreader/koreader/issues/6365#issuecomment-655930523.

PR #6292, opened 2020-06-22
ReaderGesture: cleanup

yparitcher:
Remove duplicate function

More commits coming

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6292)


yparitcher:
Ready for review.

part one of a cleanup of `ReaderGesture` making all actions events and putting functions in their appropriate files.

all the network listeners were put into a new `networklistener.lua` that is registered with both the reader and FM and can always handle network events. thanks to @pazos for #6314 allowing me to test this in the emulator

all device listeners (night mode, FL, GSensor) were put into a new `devicelistener.lua` that is registered with both the reader and FM and can always handle device events. this also supersedes `ReaderFrontlight` which was only available in the reader leading to a lot of hacks, this cleans that all up

please test.

yparitcher:
Whoops forgot to add devicelistener.lua :)

plugins/statistics.koplugin/main.lua
```diff
@@ -2101,4 +2069,36 @@ function ReaderStatistics:getReadBookByDay(month)
     return per_day
 end
 
+function ReaderStatistics:onGetReaderProgress()
```
poire-z, line 2072: I guess you could now rename all these from `onGet...` to `onShow...`

frontend/device/devicelistener.lua
```diff
@@ -150,43 +207,29 @@ function ReaderFrontLight:onShowOnOff()
...
 
-function ReaderFrontLight:onShowIntensity()
-    if not Device:hasFrontlight() then return true end
-    local powerd = Device:getPowerDevice()
+function DeviceListener:onShowFlDialog()
+    local FrontLightWidget = require("ui/widget/frontlightwidget")
+    UIManager:show(FrontLightWidget:new{
+        use_system_fl = Device:hasLightLevelFallback()
+    })
+end
```
Frenzie, line 215: It does keep the diff nice and clean but before merging it's probably better to indent for future readers.

yparitcher: done

yparitcher:
Finaly finished with this.

all gestures now call events

frontend/apps/reader/readerui.lua
```diff
@@ -369,6 +370,13 @@ function ReaderUI:init()
         view = self.view,
         ui = self,
     })
+    --  event listener for reader
+    self:registerModule("readerlistener", ReaderListener:new{
```
poire-z, line 374: Is it worth adding yet another module ? Or do you think you'll need that to add other stuff ?

yparitcher: I would rather not have added one but togglereadingorder does not fit into any of the existing ones. any better ideas?

poire-z: Given that it's ReaderRolling and ReaderPaging that holds `self.inverse_reading_order` and handle page turning, I would just have that handled in each of them (so, duplicated, but it's consistent - and they are mutually exclusive).

Frenzie: I agree that a little bit of duplication should be fine for this one, plus that's how it's been, hasn't it? (Albeit with some comment saying it shouldn't be.)

yparitcher: fixed.
discussed earlier in #5423

yparitcher:
any other objections?
I am ready to merge this

poire-z:
There's a little remnant that still references the method you renamed to onUpdateFooter:
https://github.com/koreader/koreader/blob/62daa7b7839f6ecf4910939f46f1723586bb7943/frontend/apps/reader/modules/readerfooter.lua#L1794-L1796
(Dunno if it has any purpose, and if it should be removed or updated - but the current state should cause no crash as it replaces a no-op function with nil :)

yparitcher:
> There's a little remnant that still references the method you renamed to onUpdateFooter:
> 
> https://github.com/koreader/koreader/blob/62daa7b7839f6ecf4910939f46f1723586bb7943/frontend/apps/reader/modules/readerfooter.lua#L1794-L1796
> 
> (Dunno if it has any purpose, and if it should be removed or updated - but the current state should cause no crash as it replaces a no-op function with nil :)

It ensures that the footer is updated when the zoom, fontsize etc. is changed. fixed in #6392

Issue #2778, opened 2017-04-17
Update menu items of evernote plugin


Hzj-jie:
* KOReader version: Up to #2775
* Device: Independent

#### Issue

The menu items of evernote plugin are confusing; we would be better to update it.

A potential solution is,
```
Login
Export nodes from all books
Export nodes from this book (Unclickable in FileManager)
Export to --- (Unclickable)
  >> Evernote
  >> HTML
  >> Text
```

houqp:
I think this would be better served with a pop up dialog. Then you can have the following UI setup:

```
From:
[x] export notes from all books
[ ] export ntoes from this book (hide or uncheckable in file browser)

To:
[button: Evernote] [button: HTML] [button: Text]
```

Frenzie:
I'm not sure if this is still relevant, pinging @Galunid.

Galunid:
Sure, I can do that, although is there a point? And if so, which one should be implemented, the OP's one, or houqp's one? Personally I like it how it is, but it shouldn't be a big deal to change it.

Frenzie:
There's one thing that hadn't been implemented in 2017, the separator. I'll stick some in.
<img src=https://user-images.githubusercontent.com/202757/85256496-9d7a7900-b464-11ea-8237-19169c29f1d6.png width=50%>

I understand what's being called confusing here. You can't click export until you select one of those checkmarks.

I think houqp's suggestion is better but personally I have no interest in implementing/testing it.

Issue #6293, opened 2020-06-22
FR: Translator, allow for displaying word definition


Galunid:
I think it would be pretty cool, to display word definition on top of translation in `Translate` menu.
Google Translate provides this already.
![google_tl](https://user-images.githubusercontent.com/10298730/85248130-3d79d780-b450-11ea-9bbd-4d9167031d89.png)
Interestingly supports making this query, but doesn't display the result. I'll try to work on this when I have a bit of free time.

PR #6294, opened 2020-06-22
[plugin] Add menu separators to Evernote menu

Frenzie:
This makes the menu structure slightly clearer.
See <https://github.com/koreader/koreader/issues/2778#issuecomment-647324742>.

### Before/After
 

Pinging @Galunid

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6294)


Galunid:
Makes sense, looks way better. I'll do #2778 after I finish #6293

PR #6295, opened 2020-06-22
Translator: Add definition of the word to the translation

### LINKED ISSUE #6293: FR: Translator, allow for displaying word definition

Galunid:
output
This gives us result when doing same language -> same language translation

closes #6293

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6295)

### DIFF
--- a/frontend/ui/translator.lua
+++ b/frontend/ui/translator.lua
@@ -502,7 +502,7 @@ function Translator:_showTranslation(text, target_lang, source_lang)
                 for j, res in ipairs(r[2]) do
                     table.insert(output, "\t● ".. res[1])
                 end
-		        end
+            end
         end
     end
 
--- a/frontend/ui/translator.lua
+++ b/frontend/ui/translator.lua
@@ -165,7 +165,7 @@ local Translator = {
             -- "bd",  -- dictionary (articles, reverse translations, etc)
             -- "ex",  -- examples
             -- "ld",  -- ?
-            -- "md",  -- definitions of source text
+            "md",  -- definitions of source text
             -- "qca", -- ?
             -- "rw",  -- "see also" list
             -- "rm",  -- transcription / transliteration of source and translated texts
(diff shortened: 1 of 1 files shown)


frontend/ui/translator.lua
```diff
@@ -489,7 +489,20 @@ function Translator:_showTranslation(text, target_lang, source_lang)
...
-            table.insert(output, "")
+        end
+    end
+
+    if result[13] and type(result[13]) == "table" and #result[13] > 0 then
+        -- Definition(word)
+        table.insert(output, "________")
+        for i, r in ipairs(result[13]) do
+            if r[2] and type(r[2]) == "table" then
+                local symbol = util.unicodeCodepointToUtf8(10101 + (i < 10 and i or 10))
```
Frenzie, line 500: More than 10 is unusual?

Galunid: There shouldn't be more than 10, as it puts number before parts of speech(noun/verb/etc.)
according to wikipedia there should be 9. 8 without articles. I copied this part from above and left this check just in case.
`local symbol = util.unicodeCodepointToUtf8(10101 + i)` *should* be alright too.

Frenzie: How do you mean according to Wikipedia there should be 9? Nine of what?

Galunid: https://en.wikipedia.org/wiki/Part_of_speech
9 parts of speech when you look at the sceenshot in PR it's formatted like this
[symbol1]Noun
     meaning 1
     meaning 2
[symbol2]Verb
     meaning 3
     meaning

Galunid: Once you list them all: 
    noun
    verb
    adjective
    adverb
    pronoun
    preposition
    conjunction
    interjection
    article
you get 9 top, that is assuming one word can be all of the above.

Frenzie: Oh, this code is copied from a few lines up and it's only about that. Your screenshot also shows synonyms or some such for example.

PR #6300, opened 2020-06-22
Fdroid: prevent runtime modification

pazos:
Also disables background runner plugin on android. It isn't unused by any plugin and somewhat breaks what an activity is on android.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6300)

### DIFF
--- a/platform/android/llapp_main.lua
+++ b/platform/android/llapp_main.lua
@@ -38,31 +38,31 @@ local function runUserScripts(dir, migration, parent)
     end
 end
 
--- run scripts once after an update of koreader,
--- it can also trigger a recursive migration of user data
-local run_once_scripts = path .. "/koreader/scripts.afterupdate"
-if lfs.attributes(run_once_scripts, "mode") == "directory" then
-    local afterupdate_marker = android.dir .. "/afterupdate.marker"
-    if lfs.attributes(afterupdate_marker, "mode") ~= nil then
-         if lfs.attributes(run_once_scripts .. "/migrate", "mode") ~= nil then
-             android.LOGI("after-update: running migration")
-             runUserScripts(run_once_scripts, true)
-         else
-             android.LOGI("after-update: running shell scripts")
-             runUserScripts(run_once_scripts)
-         end
-         android.execute("rm", afterupdate_marker)
+if android.prop.flavor ~= "fdroid" then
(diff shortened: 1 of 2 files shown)


poire-z:
> Also disables background runner plugin on android. It isn't unused by any plugin and somewhat breaks what an activity is on android.

I think it is not used directly, but by using `table.insert(PluginShare.backgroundJobs, {...})`, which seems currently used by autofrontlight and autowarmth stuff (on Kobo, Kindle and Cervantes only it seems):
```
--
frontend/device/cervantes/powerd.lua-
frontend/device/cervantes/powerd.lua-    -- Enable background job for setting Warmth, if not already done.
frontend/device/cervantes/powerd.lua-    if not self.autowarmth_job_running then
frontend/device/cervantes/powerd.lua:        table.insert(PluginShare.backgroundJobs, {
frontend/device/cervantes/powerd.lua-                         when = 180,
frontend/device/cervantes/powerd.lua-                         repeated = true,
frontend/device/cervantes/powerd.lua-                         executable = function()
--
frontend/device/kobo/powerd.lua-    end
frontend/device/kobo/powerd.lua-    -- Enable background job for setting Warmth, if not already done.
frontend/device/kobo/powerd.lua-    if not self.autowarmth_job_running then
frontend/device/kobo/powerd.lua:        table.insert(PluginShare.backgroundJobs, {
frontend/device/kobo/powerd.lua-                         when = 180,
frontend/device/kobo/powerd.lua-                         repeated = true,
frontend/device/kobo/powerd.lua-                         executable = function()
--
plugins/autofrontlight.koplugin/main.lua-        return true
plugins/autofrontlight.koplugin/main.lua-    end
plugins/autofrontlight.koplugin/main.lua-
plugins/autofrontlight.koplugin/main.lua:    table.insert(PluginShare.backgroundJobs, {
plugins/autofrontlight.koplugin/main.lua-        when = 2,
plugins/autofrontlight.koplugin/main.lua-        repeated = enabled,
plugins/autofrontlight.koplugin/main.lua-        executable = function()
```
and some unused frontend/ui/plugin/background_task_plugin.lua

(I wish we never had that ugly BackgroundRunner , and just used `scheduleIn()` and cancelation like we do in other places :)

PR #6305, opened 2020-06-25
bump crengine: add support for <ruby> elements

poire-z:
Includes https://github.com/koreader/crengine/pull/347 :
- Fix some non-recursive subtree walkers
- Text: skip formatting optimisations when inlineBoxes
- CSS parsing: parse selectors starting with [attrib]
- Use T="" for internal elements attributes
- CSS: re-order css_d_* (display) enum for easier checking
- Rendering methods: remove erm_table_cell
- Rendering methods: remove erm_table_caption
- Rendering methods: remove erm_list_item
- getRenderedWidths(): better estimate table width
- styleToTextFmtFlags(): adds is_block parameter
- Adds support for "display: ruby" and `<ruby>` elements
- Fix prev/next-VisibleWord-Start/End() with CJK text

Also:
- bump HarfBuzz to 2.6.8 https://github.com/koreader/koreader-base/pull/1122
- bump evernote-lua-sdk https://github.com/koreader/koreader-base/pull/1121

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6305)

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -650,14 +650,13 @@ This is just an example, that will need to be adapted into a user style tweak.]]
                     title = _("In-page FB2 footnotes"),
                     description = _([[
 Show FB2 footnote text at the bottom of pages that contain links to them.]]),
-                    -- Avoid 75% of 75% in case of nested <section>
                     css = [[
 body[name="notes"] section {
     -cr-hint: footnote-inpage;
     margin: 0 !important;
 }
 body[name="notes"] > section {
-    font-size: 75%;
+    font-size: 0.75rem;
 }
                     ]],
                 },
@@ -672,7 +671,7 @@ body[name="comments"] section {
     margin: 0 !important;
 }
 body[name="comments"] > section {
-    font-size: 85%;
+    font-size: 0.85rem;
 }
                     ]],
                     separator = true,
@@ -686,7 +685,7 @@ FB2 footnotes and endnotes get a smaller font size when displayed in-page. This 
 body[name="notes"] > section,
 body[name="comments"] > section
 {
-    font-size: 100% !important;
+    font-size: 1rem !important;
 }
                     ]],
                 },
@@ -727,7 +726,7 @@ This only works with footnotes that have specific attributes set by the publishe
 {
     -cr-hint: footnote-inpage;
     margin: 0 !important;
-    font-size: 80% !important;
+    font-size: 0.8rem !important;
 }
                 ]],
                 separator = true,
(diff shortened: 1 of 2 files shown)


poire-z:
I updated the book style tweak sample to use `0.8rem` instead of `80%` because that's what I've been happily using for some time, and it's easily changeable there if it's not the best option.

I'm not sure if I should do the same for the 3 other "In-page footnotes (small)" variants, as it happened to me that I had books with `body { font-size: 0.8em or 1.2em}` which had me increase or decrease KOReader's font size to get my usual prefered font size - using 80% would work in that case, while using 0.8rem would make the footnotes font too large.

poire-z:
> my quick ruby in the margin demo could be included

It's a bit tedious to get right and not ugly :)
With your absolute sizes in px, I would get on the emulator:
<kbd>image</kbd>
and one would need to finely adjust Zoom/DPI to get something somehow OK (and bad if you go larger than OK).

And if there are 2 ruby in the same line, the 2 annotations would override each other:
<kbd>image</kbd>

So, I'd rather let that to the user via User style tweaks - so he can try to solve/size it according to his setup.

Also, I'd rather not advocate the use of ruby for wordwise stuff, because soon, someone will see the extra space added when the definition is longer than the base word:
<kbd>image</kbd>
and will ask for this to be fixed - and that won't really be as easily fixable as implementing this ruby support with inline-table (because that would mean sizing the inline-table only with the main word, and find a way to have the other cells overflow them, and on which side ? that would depend on the position of the inline-table in the line, and whether there are other such ruby around...  And I don't want to have to add such kludge if wordwise (not in any specs) in KOReader would become popular :)

Frenzie:
> It's a bit tedious to get right and not ugly :)

Yup, hence why I called it quick and dirty. I didn't intend it as anything more than a quick demo of how one could approach taht.

> And I don't want to have to add such kludge if wordwise (not in any specs) in KOReader would become popular :)

Agreed.

frontend/ui/data/css_tweaks.lua
```diff
@@ -301,6 +301,14 @@ h1, h2, h3, h4, h5, h6 { hyphens: none !important; }
 * { font-variant: no-common-ligatures; }
                 ]],
             },
+            {
+                id = "ruby_inline";
+                title = _("Render <ruby> content inline"),
+                description = _("Disable handling of <ruby> tags and have them rendered inline."),
```
Frenzie, line 307: I just realized "have them rendered inline" seems a bit wordy. Maybe like this?
```suggestion
                description = _("Disable handling of <ruby> tags and render them inline."),
```

PR #6306, opened 2020-06-25
Bookmarks prev/next via gestures: add location to stack

poire-z:
Was correctly done when jumping bookmarks via the Bookmarks menu and the SkimTo widget, but not from gestures. Details in https://github.com/koreader/koreader/pull/6259#issuecomment-642677750.  Supersede #6259.
Also prevent fillToc() on each page turn when book has no TOC.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6306)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -647,12 +647,18 @@ function ReaderBookmark:onGotoNextBookmark(pn_or_xp)
     return true
 end
 
-function ReaderBookmark:onGotoNextBookmarkFromPage()
+function ReaderBookmark:onGotoNextBookmarkFromPage(add_current_location_to_stack)
+    if add_current_location_to_stack ~= false then -- nil or true
+        self.ui.link:addCurrentLocationToStack()
+    end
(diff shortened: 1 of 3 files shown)


frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -647,12 +647,18 @@ function ReaderBookmark:onGotoNextBookmark(pn_or_xp)
     return true
 end
 
-function ReaderBookmark:onGotoNextBookmarkFromPage()
+function ReaderBookmark:onGotoNextBookmarkFromPage(add_current_location_to_stack)
+    if add_current_location_to_stack ~= false then -- nil or true
```
Galunid, line 651: Isn't this more appropriate?
```lua
if not add_current_location_to_stack then
```

poire-z: I want `nil` (what other code I won't touch will provide) to behave as `true`. Only in the explicite places I want to not have that default behaviour to happen, I'll provide `false`.
(We use that kind of construct to have a positive-named parameter in other places.)

PR #6308, opened 2020-06-25
EvernoteOAuth set logger

Galunid:
This sets default EvernoteOAuth logger introduced in https://github.com/koreader/evernote-sdk-lua/pull/4 to `DEBUG`, instead of default `print`

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6308)

### DIFF
--- a/plugins/evernote.koplugin/main.lua
+++ b/plugins/evernote.koplugin/main.lua
@@ -7,9 +7,9 @@ local DataStorage = require("datastorage")
 local DocSettings = require("docsettings")
 local UIManager = require("ui/uimanager")
 local Screen = require("device").screen
+local logger = require("logger")
 local util = require("ffi/util")
 local Device = require("device")
-local logger = require("logger")
 local JoplinClient = require("JoplinClient")
 local T = require("ffi/util").template
 local _ = require("gettext")
@@ -607,7 +607,7 @@ function EvernoteExporter:exportClippings(clippings)
                 NetworkMgr:promptWifiOn()
                 return
             elseif not ok and err then
-                logger.dbg("Error occurs when exporting book:", title, err)
+                logger.dbg("Error while exporting book", title, err)
                 error_count = error_count + 1
                 error_title = title
             elseif ok then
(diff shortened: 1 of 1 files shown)


plugins/evernote.koplugin/main.lua
```diff
@@ -405,6 +405,7 @@ function EvernoteExporter:doLogin(username, password)
         domain = self.evernote_domain,
         username = username,
         password = password,
+        logger = DEBUG
```
Frenzie, line 408: Would you mind changing them all to logger.dbg? That `DEBUG` is kind of a blast from the past.

More important, we prefer a trailing comma at the end. ;-)

Galunid: Sure

plugins/evernote.koplugin/main.lua
```diff
@@ -9,7 +9,7 @@ local UIManager = require("ui/uimanager")
 local Screen = require("device").screen
 local util = require("ffi/util")
 local Device = require("device")
-local DEBUG = require("dbg")
+local logger = require("logger")
```
Frenzie, line 12: Although the order here's a bit random could you put it right above util? ^_^

plugins/evernote.koplugin/main.lua
```diff
@@ -405,6 +405,7 @@ function EvernoteExporter:doLogin(username, password)
         domain = self.evernote_domain,
         username = username,
         password = password,
+        logger = logger.dbg
```
Frenzie, line 408:
```suggestion
        logger = logger.dbg,
```

plugins/evernote.koplugin/main.lua
```diff
@@ -606,11 +607,11 @@ function EvernoteExporter:exportClippings(clippings)
                 NetworkMgr:promptWifiOn()
                 return
             elseif not ok and err then
-                DEBUG("Error occurs when exporting book:", title, err)
+                logger.dbg("Error occurs when exporting book:", title, err)
```
Frenzie, line 610:
```suggestion
                logger.dbg("Error when exporting book:", title, err)
```

Issue #5725, opened 2020-01-01
FR: Android TV


pazos:
I know, I know. It is as stupid as it sounds. Who wants to run an e-reader app on a tv?.

Technically is possible and pretty easy: just detect if we run on a tv platform, set hasDpad=true and change orientation from "sensorPortrait" to "landscape".

I'm not sure if there's some hardcoded value somewhere (on the filemanager?) for height > width. Today tried to run an almost "vanilla" version on my new tv box and while the splash screen was in landscape mode once it reached the filemanager it turned to reverse portrait mode.

Frenzie:
Depends a bit on the kind of document. ;-)

The FileManager goes through some trouble to ensure portrait iirc.

pazos:
> The FileManager goes through some trouble to ensure portrait iirc.

Awh. I was expecting some kind of magic port without any effort  :smile:

NiLuJe:
Yeah, IIRC, the FM is very explicitly Portrait only. It can do Inverted Portrait (but you can only achieve that on devices with a Gyro, f.g., Forma/Libra).

KenMaltby:
I wonder about the latest game machines, like the 4K Xbox One, they seem to be headed into a more general app support approach, than just game machines.  Not as obvious as Android, and there could be Microsoft licensing issues, but might be a future thing, especially combined with the voice activated gadgets people are letting into their lives.
Luck;
Ken

NiLuJe:
Unfortunately, those platforms are extremely hostile to homebrew. You basically have to be a registered developer, access to the SDKs usually costs big money, and they're an NDA riddled mess that makes everything harder (impossible?N) for OSS projects...

Frenzie:
Yeah, the Xbox One could be seen as a locked down nightmare version of Windows. They'd make for a pretty interesting PC otherwise, as a viable alternative for one of those € 300 or possibly even as much as € 400 NUCs. Except they only cost a little over € 200.

Unfortunately it's not just the software, but also the hardware that's purposefully distinct in minute details.

pazos:
Reading PDFs on a big tv:

![test](https://user-images.githubusercontent.com/975883/71689985-af2b6f80-2da3-11ea-955d-0d284e4ce7f9.gif)

The shadows you see here are not visible in the tv. I guess is something related to scrcpy on androidTv.
The dpad controls are rotated.

Frenzie:
[...]

Issue #5910, opened 2020-02-29
Incorrect rotation when width larger than height


pazos:
* KOReader version: 2020.02
* Device: SDL

#### Issue

Screen is rotated 90º CW when started in fullscreen, like on Android TV (#5725). Both touch and keys are not, so is pretty unusable even if the screen is in vertical position.

#### Steps to reproduce

```
export SDL_FULLSCREEN
koreader
```

![Captura de pantalla -2020-02-29 18-21-21](https://user-images.githubusercontent.com/975883/75612093-d2169d80-5b20-11ea-81ad-3e432a333e13.png)

Frenzie:
The issue title is slightly wrong. Just try something like `kodev run -w=300 -h=200` to see why. ;-)

pazos:
@Frenzie: same thing. What I should see?

NiLuJe:
There's a w > h check that does stupid things somewhere (let's say, because of Kobos? :D).

NiLuJe:
(I'm thinking of the isAlwaysPortrait crap, which possibly also applies here for some unfathomable reason).

Frenzie:
@pazos I mean that `SDL_FULLSCREEN` doesn't affect a thing. On typical Ubuntu & Sailfish phones height is bigger than width so nothing will be "rotated." It's a generic width larger than height issue which happens to be triggered by fullscreen in your specific condition.

Frenzie:
Put another way, we're pretty much running in fullscreen on all platforms. ;-)

Frenzie:
See https://github.com/koreader/koreader-base/blob/5f9bc28fe7b599a30cce0272d7e44ebc435373e9/ffi/framebuffer.lua#L307-L313

pazos:
Ok, thanks for the info. This quick and dirty diff fixes initial rotation. Things go boom on further rotations:

```diff
diff --git a/ffi/framebuffer.lua b/ffi/framebuffer.lua
index a8f1f89..8313df3 100644
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -313,10 +313,12 @@ function fb:getScreenMode()
 end
 
 function fb:setRotationMode(mode)
+    --[[
     self.bb:rotateAbsolute(-90 * (mode - self.native_rotation_mode - self.blitbuffer_rotation_mode))
     if self.viewport then
         self.full_bb:setRotation(self.bb:getRotation())
     end
+    --]]
[...]

Issue #5106, opened 2019-07-03
Gyro enhancement


sopincariu:
* KOReader version:
* Device:Kobo Forma

#### Issue
We can only turn on or off the accelerometer  events. It should be useful if we can lock in a certain certain orientation (portrait or landscape).
Thanks

pazos:
If you disable sensor events you can get your desired orientation per document. Just use the bottom menu, rotation -> landscape/portrait (press twice for 180º rotation). 

AFAIK the file manager will always be in portrait

NiLuJe:
c.f., comments in the relevant bits of frontend/device/input.lua

I don't have any specific interest in tweaking that, but it's there for the taking for interested parties.

wildertux:
KOReader version 2019.10
DEVICE Acer Iconia Tab 10 (A3a50)

KoReader and inbuild File Manger starts with wrong Orientation (is upside down).
a config option for this would be nice.
and/or a Help info for 180° toggle.

Frenzie:
Is that an Android device? Either way it's wholly unrelated to this Kobo Forma issue. ;-)

You can rotate 180° in landscape and portrait by tapping a second time on the rotation in the bottom menu.

NiLuJe:
(Orientation lock not yet implemented, re-opening ;)).

PR #6309, opened 2020-06-26
Landscape FM / Refactor rotation

yparitcher:
refactor and simplify the orientation handling code. 
the user generally cares about the rotation (what direction the device is facing) and not about if koreader is displaying in portrait or landscape mode

This allows the user to directly set any rotation directly from the UI without resorting to hacks.

Before this PR we were always setting a screenmode and the checking what rotation to use, now we set rotations, and let koreader figure out what is the screenmode. (besides for the FM, which does not support landscape) now the only thing setting a screenmode is the FM.

This separates the rotation (physical) from the screenmode (portrait/landscape or H>W/W>H). I did not get into why the FM has to be portrait but if someone fixes that then the rotation and screenmode will be totally independent.

After this is merged i will fixup the `setScreenMode` in framebuffer.lua in base.

~I did not fix the tests yet, after i get feedback on this I will convert them.~

pinging @NiLuJe 

Screenshot:
Reader_2020-06-26_030919

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6309)


poire-z:
Wow, you're courageaous to go at that :) (that's some part of the code, with the frontlight stuff, I wish I'll never have to look at because of the many possible edge cases:/)

Would that magically solve the issue I mentionned at https://github.com/koreader/koreader/issues/4752#issuecomment-470708895 and https://github.com/koreader/koreader/issues/4752#issuecomment-573119439 ?

yparitcher:
for some stupid reason the test suite uses `setrotation` as a refresh even when the rotation is not changing (when nothing should happen) this leads to some ugly stuff in https://github.com/koreader/koreader/pull/6309/commits/70171b0ef0842562feb9e282b47f2a6089d4a169 

Does anyone know why this is and how i can change it? (maybe reset the document on each test?)

Frenzie:
Related: https://github.com/koreader/koreader/issues/5910

yparitcher:
> Wow, you're courageaous to go at that :) (that's some part of the code, with the frontlight stuff, I wish I'll never have to look at because of the many possible edge cases:/)

it took a while to build up the courage :) but this has been driving me a little nuts
> Would that magically solve the issue I mentionned at [#4752 (comment)](https://github.com/koreader/koreader/issues/4752#issuecomment-470708895) and [#4752 (comment)](https://github.com/koreader/koreader/issues/4752#issuecomment-573119439) ?

yes, the setting now does not know about portrait/landscape it only knows up/down/right/left. the dual/single code checks for H>W

Frenzie:
@yparitcher See http://olivinelabs.com/busted/#before-each for the mechanisms provided by busted. But it looks to me like the commit you linked is about testing a chain of rotations or is the diff misleading?

yparitcher:
> Related: #5910

Does not fix this (it happens because of the FM mess)

yparitcher:
> @yparitcher See http://olivinelabs.com/busted/#before-each for the mechanisms provided by busted. But it looks to me like the commit you linked is about testing a chain of rotations or is the diff misleading?

without this part:
```diff

-function ReaderView:onSetRotationMode(rotation)
+function ReaderView:onSetRotationMode(rotation, force)
    if rotation ~= nil then
-        if rotation == Screen:getRotationMode() then
+        if rotation == Screen:getRotationMode() and not force then
            return true
        end
        Screen:setRotationMode(rotation)
@@ -731,7 +731,7 @@ function ReaderView:onReadSettings(config)
        rotation_mode = config:readSetting("rotation_mode") or G_reader_settings:readSetting("copt_rotation_mode") or 0
    end
    if rotation_mode then
-        self:onSetRotationMode(rotation_mode)
+        self:onSetRotationMode(rotation_mode, true)
    end
    self.state.gamma = config:readSetting("gamma") or 1.0
```
it blows up the rest of the test suite also.
there are 2 fixes in that commit.

yparitcher:
> for some stupid reason the test suite uses `setrotation` as a refresh even when the rotation is not changing (when nothing should happen) this leads to some ugly stuff in [70171b0](https://github.com/koreader/koreader/commit/70171b0ef0842562feb9e282b47f2a6089d4a169)
> 
> Does anyone know why this is and how i can change it? (maybe reset the document on each test?)

The tests were relying on the call to `setScreenMode` to set the document dimensions.

Fixed by setting the dimensions in the call to `ReaderUI:new`.

NiLuJe:
Will test on a device with a gyro tonight, but looks good at first glance, thanks ;).

----

I *think* we have an old DEFAULT that says "landscape should always be clockwise" somewhere that can probably go poof now ;).

yparitcher:
> I _think_ we have an old DEFAULT that says "landscape should always be clockwise" somewhere that can probably go poof now ;).

yes, `DLANDSCAPE_CLOCKWISE_ROTATION`, but the framebuffer uses it so I will delete it after I clean up that part

yparitcher:
> > I _think_ we have an old DEFAULT that says "landscape should always be clockwise" somewhere that can probably go poof now ;).
> 
> yes, `DLANDSCAPE_CLOCKWISE_ROTATION`, but the framebuffer uses it so I will delete it after I clean up that part

actually we might want to keep it for things only care about the portrait/landscape and are using `setScreenMode`. otoh right now that is only the FM & only for portrait so maybe not.

yparitcher:
Please don't merge yet.

Sorry, I took it all the way. Landscape FM coming. It works fine so far, I am not sure why it was never enabled

A few questions.

1. What is the best way to reload the FM? (for when switching rotation in the FM).

2. (The hard one) I think we are going to need a setting to prevent the Reader and FM orientation from affecting each other for those that want. Please give your opinion for or against. :)

yparitcher:
>     2. (The hard one) I think we are going to need a setting to prevent the Reader and FM orientation from affecting each other for those that want. Please give your opinion for or against. :)

I made a new setting called `lock_rotation` that when enabled prevents the rotation from bleeding between the FM and Reader.

TODO:

Reload the FM on rotation

Add setting to the UI to set the rotation & default in FM.

Please test

Frenzie:
> (The hard one) I think we are going to need a setting to prevent the Reader and FM orientation from affecting each other for those that want. Please give your opinion for or against. :)

I'm in the camp that likes them separate. I want them both in portrait, but some documents are a bit wider. I almost always prefer to use landscape over reflow but that doesn't mean I want the file manager to go along.

NiLuJe:
What @Frenzie said ;).

(Also, the fact that there's a hard 100ms penalty to a rotated blit doesn't help. Given that a fast unrotated blit averages between 1 and 3ms ;p).

frontend/ui/data/strings.lua
```diff
@@ -77,8 +77,10 @@ S.HIGH = _("high")
 S.ZERO_DEG = _("0 deg")
 S.FIVE_DEG = _("5 deg")
 S.TEN_DEG = _("10 deg")
-S.PORTRAIT = _("portrait")
-S.LANDSCAPE = _("landscape")
+S.PORTRAIT = _("up")
```
Frenzie, line 80: I'm not quite sure what I think of up/down/left/right, but either way such generic strings should probably be context-disambiguated:
```suggestion
S.PORTRAIT = C_("Rotation", "up")
```
http://koreader.rocks/doc/modules/gettext.html#pgettext

yparitcher: didnt know about this, thanks

NiLuJe:
(I will *almost definitely* have time to test this tomorrow ^^).

frontend/ui/data/strings.lua
NiLuJe: The LinuxFB convention is Upright, Clockwise, Upside Down, Counter-Clockwise. I'm kinda use to it by now, but it's potentially not really any better, so, just throwing it out there ;).

yparitcher: > The LinuxFB convention is Upright, Clockwise, Upside Down, Counter-Clockwise. I'm kinda used to it by now, but it's potentially not really any better, so, just throwing it out there ;).

those strings are long and won't fit well in the UI toggles. thats why I made it short and simple

frontend/ui/data/strings.lua
```diff
@@ -77,8 +77,10 @@ S.HIGH = _("high")
...
+S.LANDSCAPE = _("right")
+S.PORTRAIT_ROTATED = _("down")
```
NiLuJe, line 82: In the same vein, I tend to refer to those as "Inverted", elsewhere in the code.

But it's only a variable name here, so, meh ;).

yparitcher: I copied `ROTATED` from the framebuffer code and in the cleanup happened to get rid of all the inverted references, to me rotated or inverted are the same, but we should make sure this matches the FB code

yparitcher:
Ready for review.

in summary:

makes portrait/landscape RO determined by W>H used by koreader to determine how to display the UI.

the user can set the rotation (up/down/right/left) directly in both the FM and reader.

added a new setting `lock_rotation` that controls if the rotation bleeds between the FM and reader.

when opening the reader: 
1. the books rotation 
2. if this is the first time for the book, 
- if not locked use the current rotation
- if locked use the reader default.

when opening the FM: 
- if not locked use the current rotation
- if locked use the FM default.

add settings for this under Screen> and moved the gyro setting into this.

Should is use "Rotation" or "Orientation" for the heading text?

Please let me know of better strings for the UI settings.

Screenshot:
FileManager_2020-06-28_011528

frontend/ui/elements/screen_rotation_menu_table.lua
```diff
@@ -0,0 +1,96 @@
...
+            callback = function()
+                G_reader_settings:flipNilOrFalse("input_ignore_gsensor")
+                Device:toggleGSensor(not G_reader_settings:isTrue("input_ignore_gsensor"))
+            end,
+        })
+    end
+
+    table.insert(rotation_table, {
+        text = _("Lock rotation"),
+        help_text = _("When checked the rotation of the filemanager and reader will not affect each other"),
```
Frenzie, line 28:
```suggestion
        help_text = _("When checked the rotation of the file browser and the reader will not affect each other."),
```

yparitcher:
Fixes: #5725 #5106 #2565

poire-z:
> Should is use "Rotation" or "Orientation" for the heading text?

Yeah, the naming are a bit confusing. Up/down are orientations, while Left/right are rotations (at least, in my way of thinking: I rotate the top of page to the right - I wouldn't know what right orientation means: is it the the top, or the bottom, of the page that is on the right? or should I turn my device right?...).
I think I would still prefer "Orientation" as the name for all this stuff.

> added a new setting lock_rotation that controls if the rotation bleeds between the FM and reader.

This one is understandable thanks to your help_text, but the word "lock" is a bit confusing standalone:
- does "lock" means the orientation set in FM is locked, and the one set in Reader is locked too (so they won't bleed and are independant) - your decision
- or does it mean they are both locked to each other, and so will bleed/carry the other too - my initial naive expectation

Just some thoughts, I can't think of any other less confusing wording :/

Frenzie:
@pazos While rotating the entire interface is a bit awkward, I'd rather have it like that than not at all. Being able to rotate the document is definitely desirable.

poire-z:
> Being able to rotate the document is definitely desirable.

I agree. One could then read on an Android TV or a normally sitting laptop with the document rotated 90° while lying horizontal on his side, head on a pillow :) and rotate it -90° when switching side.

yparitcher:
also if one has a device with a messed up FB they can just switch the default rotation instead of having to add a device quirk

Frenzie:
I just checked what my laptop did in WIndows (because I recall thinking that was clear) and it uses degrees.

> The rotation feature of Intel® graphics drivers allows the user to rotate a monitor or display's entire image to one of four orientations when enabled. The image can be rotated 90°, 180°, 270°, and 0° (Normal).
<https://www.intel.com/content/www/us/en/support/articles/000005491/graphics.html>

Combined with a little symbol, perhaps it would be clearer at a glance.

* `⤹ 90°`
* `↺ 180°`
* `⤸ 270°`

Frenzie:
Looks something like this:

Screenshot_2020-06-28_23-48-55

NiLuJe:
@Frenzie: ⤹ and ⤸ are inverted, but, yeah ;).

(i.e., according to the initial screenshot, the first rotation is CW, i.e., right, i.e., ⤸ 90° ;).

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -1479,9 +1479,8 @@ function ReaderGesture:gestureAction(action, ges)
         self.ui:handleEvent(Event:new("RestoreZoomMode"))
         self.ui:handleEvent(Event:new("InitScrollPageStates"))
     elseif action == "toggle_rotation" then
-        local event_name = self.document.info.has_pages and "SwapScreenMode" or "ChangeScreenMode"
-        local arg = Screen:getScreenMode() == "portrait" and "landscape" or "portrait"
-        self.ui:handleEvent(Event:new(event_name, arg))
+        local arg = (Screen:getRotationMode() +1)%4
```
NiLuJe, line 1482: I generally do (+1 & 3) (as do the odd/even checks).

NiLuJe:
The `self.ui.doc_settings:saveSetting("screen_mode", Screen:getScreenMode())` @ frontend/apps/reader/modules/readerview.lua L824 can also probably be removed now ;).

NiLuJe:
Okay, finally had time to try this ;).

* Gyro rotations in the FM do not refresh the top 'menu' row.
* Not quite sure why the rotation entries in the top menu use checkboxes, since none of them are actually ever checked ;). (either make sure the current rota is checked, or get rid of the checkboxes?)

NiLuJe:
And I'd consider this one a feature and not a bug:

* A gyro rotation will never trigger if a menu widget is open (top/bottom).

NiLuJe:
>     * Gyro rotations in the FM do not refresh the top 'header/menu' row.

Actually a tiny bit worse than that: the actual FM state seems to get lost along the way.
As soon as I trigger a gyro rota in the FM, the folder I did that in will "stick", and the header will never reflect the actual folder I'm in, and subsequent rotations will throw me back to that original folder.

NiLuJe:
This works slightly better:

```lua
function FileManager:onSetRotationMode(rotation)
    if rotation ~= nil and rotation ~= Screen:getRotationMode() then
        Screen:setRotationMode(rotation)
        if self.instance then
            self:reinit(self.instance.path, self.instance.focused_file)
            UIManager:setDirty(self.instance.banner, function()
                return "ui", self.instance.banner.dimen
            end)
        end
    end
    return true
end
```

Leverages the existing `reinit` call, and explicitly ask for a banner refresh.

The only thing we lose is the current page, you're thrown back to the first page of the current folder.

Frenzie:
@NiLuJe That's on purpose. Degrees are measured CCW (albeit more typically from the x-axis). To revert them it would be -90°, -180° and -270°.

The order could also be shuffled in various other ways, like 270°, 0°, 90°, 180° or 0°, 180°, 90°, 270°. (Or of course 0°, 270°, 180°, 90° but it seems less logical to me.)

NiLuJe:
@Frenzie: Ah. The LinuxFB convention (and what we're following here) is in CW 90° turns ;).

NiLuJe:
Other than that, my only other issue with wording is for the "Lock rotation" thingy. The help is clear, but the label itself isn't ;).

I'd possibly reverse it, and say "Honor the FM rotation when opening a book" or something like that?

poire-z:
> a way to somewhat preserve pagination, although, given the fact that Landscape will show a different amount of items per page than Portrait, a pagination reset isn't necessarily a bad thing?

I remember I once had to handle or fix that, might be in #3351 and #3391, the `focused_file` or `focused_path` stuff. And may be too in CoverBrowser.
But I've seen that it's also broken in CoverBrowser, when switching nb of items per page, we're often flying to some other page. Didn't really care about investigating that.
So, may be keep that for being fixed later, as it's probably not new (unless of course if you want to :)

Frenzie:
@NiLuJe Another option would be something like `90°` and `-90°` (in which case `+90°` as CW makes a certain kind of intuitive sense). Then something like `90°, 0°, 270°, 180°` could be presented as `-90°, 0°, 90°, 180°` which might look a little nicer than that `270°`. But the Linux framebuffer strikes me as going against normal convention.

Here's Intel:
cui2_rotation

Other word-based options: normal/left/right/invert. But I think the concept with arrows is clearer and you can almost regard the number of degrees as filler.

NiLuJe:
`-90°, 0°, 90°, 180°` is fine by me ;). (i.e., ⤹/3/CCW, 0/UR, ⤸/1/CW, 2/UD;

frontend/apps/reader/modules/readergesture.lua
NiLuJe: Another strike against the modulo approach that the latest comment about -90° made me remember:

```
┌─(niluje@illyria:pts/9)─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~/MPLAYER/FBInk)─┐
└─(0.65:27%:18:31:%)── echo $(( (0 - 1) & 3 ))                                                                                                                                                                                                                                                                                                         ──(Mon, Jun 29)─┘
3

┌─(niluje@illyria:pts/9)─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~/MPLAYER/FBInk)─┐
└─(0.31:26%:18:37:%)── echo $(( (0 - 1) % 4 ))                                                                                                                                                                                                                                                                                                         ──(Mon, Jun 29)─┘
-1
```

;).

NiLuJe: Granted, it happens to behaves in Lua ;)

NiLuJe:
Another thing, which may be a side-effect of the `reinit` thing, since I hadn't noticed it before:

* When switching rotation from the top-menu, (i.e., UR -> CW), the menu itself won't be resized yet on refresh. If you close it and open it again, everything's fine, but if you try to keep using the badly sized open one, things go wonky.
This may only be happening on the very first time the top menu is open after startup/rota...

NiLuJe:
Other than that, haven't found anything better than `reinit`. Trying to sneakily resize stuff manually didn't go all that well... ^^

```lua
            --[[
            self.instance.dimen = Screen:getSize()
            -- TouchZone reset
            print("Pre self.path_text max_width", self.instance.path_text.max_width)
            print("Pre Banner's CenterContainer 1:", dump(self.instance.banner[1][1].dimen))
            print("Pre Banner's Button width:", self.instance.banner[1][1][1][2][1].width)
            print("Pre Banner's CenterContainer 2:", dump(self.instance.banner[1][2].dimen))
            print("Pre FC width", self.instance.file_chooser.width)
            print("Pre FC height", self.instance.file_chooser.height)

            self.instance.path_text.max_width = Screen:getWidth() - 2*Size.padding.small
            self.instance.banner[1][1].dimen.w = Screen:getWidth()
            local icon_size = Screen:scaleBySize(35)
            self.instance.banner[1][1][1][2][1].width = Screen:getWidth() - 2 * icon_size - 4 * Size.padding.large
            self.instance.banner[1][2].dimen.w = Screen:getWidth()

            self.instance.file_chooser.width = Screen:getWidth()
            self.instance.file_chooser.height = Screen:getHeight() - self.banner:getSize().h

            print("Post self.path_text max_width", self.instance.path_text.max_width)
            print("Post Banner's CenterContainer 1:", dump(self.instance.banner[1][1].dimen))
            print("Post Banner's Button width:", self.instance.banner[1][1][1][2][1].width)
            print("Post Banner's CenterContainer 2:", dump(self.instance.banner[1][2].dimen))
            print("Post FC width", self.instance.file_chooser.width)
            print("Post FC height", self.instance.file_chooser.height)

            -- NOTE: This doesn't actually do anything?
            --       Move registerTouchZones to a dedicated function called at init & onSetDims, like FileManagerMenu...
            self:handleEvent(Event:new("SetDimensions", self.instance.dimen))

            UIManager:setDirty(self.instance.banner, function()
                return "ui", self.instance.banner.dimen
            end)
            UIManager:setDirty(self.instance, function()
                return "ui", self.instance.path_text.dimen, self.instance.dithered
            end)
            --]]
```

NiLuJe:
Oh, just figured out what you were trying to do with the checkboxes after reading the code ;).

That's a wee bit obscure ;).

I'd suggest having the checkmark on the *current* screen rota, and denote the default via a star postfix, like we do elsewhere (mainly, fonts).

pazos:
> (Also, the fact that there's a hard 100ms penalty to a rotated blit doesn't help. (as opposed to somewhere between 1 and 3ms for an unrotated blit)).

The penalty is even higher on android, where we ask to write on a surface provided by the Window Manager but the Window manager just uses whatever the SurfaceFlinger provides (ie: we're doing a transformation over a transformation).

Now that @yparitcher fixed the wonky W>H behaviour it is actually possible to skip "software" rotation on android and rely on api calls, so the penalty is gone. See:

test-rota (1)

Actually the gif looks terrible on landscape because scrcpy transforms the screen, but looks nice on device (except the top menu, but that's another thing).

NiLuJe:
> (except the top menu, but that's another thing).

AH, good to know that's not a side-effect of my tweaks, as I feared in https://github.com/koreader/koreader/pull/6309#issuecomment-651272923 ;).

Frenzie:
The menu's been that way ever since I wrote the initial resizing code for SDL. I never fixed it up like I did the basic document display. (Actually basic document display is more hacked than fixed. Definitely a candidate for some follow-up refactoring if anyone feels the calling.)

frontend/apps/reader/modules/readergesture.lua
yparitcher: needs bit module no builtin &

yparitcher:
> Another thing, which may be a side-effect of the `reinit` thing, since I hadn't noticed it before:
> 
>     * When switching rotation from the top-menu, (i.e., UR -> CW), the menu itself won't be resized yet on refresh. If you close it and open it again, everything's fine, but if you try to keep using the badly sized open one, things go wonky.
>       This may only be happening on the very first time the top menu is open after startup/rota...
> 
> 
> EDIT: If there's no quick fix, just close the menu on tap?

close the menu done

> Oh, just figured out what you were trying to do with the checkboxes after reading the code ;).
> 
> That's a wee bit obscure ;).
> 
> I'd suggest having the checkmark on the _current_ screen rota, and denote the saved default via a star postfix, like we do elsewhere (mainly, fonts).

done


I didn't see the reinit but now changed to using it, Thanks

yparitcher:
> Other than that, my only other issue with wording is for the "Lock rotation" thingy. As was mentioned earlier, the help is clear, but the label itself is slightly misleading ;).
> 
> I'd possibly reverse it, and say "Honor the FM rotation when opening a document" or something like that?
> 
> (Or "sticky rotation" or something along those lines, conveying the fact that the FM rotation will stick ^^).

Can anyone come up with a label for the menu? "Honor the FM rotation when opening a document" is a little long

yparitcher:
Screenshot:
FileManager_2020-06-29_215139

NiLuJe:
I still have the same issue raised in https://github.com/koreader/koreader/pull/6309#issuecomment-650884658 with the arrow going CCW while the rotation will be actually be CW ;p.

Unless someone has a better idea, https://github.com/koreader/koreader/pull/6309#issuecomment-651231549 feels more readable to me.

NiLuJe:
> > Other than that, my only other issue with wording is for the "Lock rotation" thingy. As was mentioned earlier, the help is clear, but the label itself is slightly misleading ;).
> > I'd possibly reverse it, and say "Honor the FM rotation when opening a document" or something like that?
> > (Or "sticky rotation" or something along those lines, conveying the fact that the FM rotation will stick ^^).
> 
> Can anyone come up with a label for the menu? "Honor the FM rotation when opening a document" is a little long

"Use file-manager rotation everywhere"?

yparitcher:
> I still have the same issue raised in [#6309 (comment)](https://github.com/koreader/koreader/pull/6309#issuecomment-650884658) with the arrow going CCW while the rotation will be actually be CW ;p.

The arrow is the right direction
the screen turns CCW
the user is on the right of the device instead of the bottom
Reader_2020-06-29_221717

yparitcher:
> @Frenzie: Ah. The LinuxFB convention (and what we're following here) is in CW 90° turns ;).
> 
> (i.e., 0 is Upright, formerly Portrait, 1 is ClockWise (+90° clockwise, formerly Landscape), etc.)

0 is Upright, formerly Portrait, 1 is **CounterClockWise** (+90° **counter clockwise**, formerly Landscape), etc

NiLuJe:
:?

`<linux/fb.h>`

```C
/*                          
  * Display rotation support 
  */                         
 #define FB_ROTATE_UR      0 
 #define FB_ROTATE_CW      1 
 #define FB_ROTATE_UD      2
 #define FB_ROTATE_CCW     3
```

I'm *definitely* rotating CW to get to 1 ^^.

yparitcher:
> :?
> 
> `<linux/fb.h>`
> 
> ```c
> /*                          
>  * Display rotation support 
>  */                         
> #define FB_ROTATE_UR      0 
> #define FB_ROTATE_CW      1 
> #define FB_ROTATE_UD      2
> #define FB_ROTATE_CCW     3
> ```
> 
> I'm _definitely_ rotating CW to get to 1 ^^, and `setRotationMode` follows that.

```lua
function fb:setRotationMode(mode)
    self.bb:rotateAbsolute(-90 * (mode - self.native_rotation_mode - self.blitbuffer_rotation_mode))
```
**-90**

and in the blitbuffer
```lua
function BB_mt.__index:rotateAbsolute(degree)
    local mode = (degree % 360) / 90
```

so a rotation of 1 = 270°

1 = CCW

NiLuJe:
Oohkay, then that's needlessly complex and/or broken for some reason, because:

forma_rota

That's definitely 90° CW unless I'm going mad?

NiLuJe:
Okay, wait, we're talking about two different things ^^.

I meant rotating the *device* physically. The fb content indeed goes -90 CCW/270° CW ;).

yparitcher:
> Okay, wait, we're talking about two different things ^^.
> 
> I meant rotating the _device_ physically. The fb content indeed goes -90° CCW/270° CW ;).


I always assumed the device stays the same and the screen rotates ;) (What actually happens)
this is why I like the current way (CCW)
This is more natural to me

> And I do think that's the most user-friendly way to display it ;).
 
by that do you mean current (CCW) or CW?

NiLuJe:
Right now, the arrow/degree combo don't make much sense anyway, since 90° CCW == 270° CW ;).

NiLuJe:
I'd consider the case where people rotate without the device following suit (i.e., the displayed bottom/top matching the way they're holding the device) to be edge-cases (e.g., lying down with the device askew or something).

Which is why I thought the -90° CCW / 0° / 90° CW / 180° compromise made the most sense.

yparitcher:
There are 2 things here:

-  the order: which i am fine with ` -90° CCW / 0° / 90° CW / 180°` aside from that the default and almost always used 0° should be first
-  which of the 90° get which arrow

NiLuJe:
> There are 2 things here:
> 
>     * the order: which i am fine with ` -90° CCW / 0° / 90° CW / 180°` aside from that the default and almost always used 0° should be first
 
While I would generally agree that the default should come first, I vaguely feel like having it between the two landscape make more sense in this context?

>     * which of the 90° get which arrow

CCW/CW was me avoiding a c/p ^^.

⤹ = CCW & ⤸ = CW

yparitcher:
> While I would generally agree that the default should come first, I vaguely feel like having it between the two landscape makes more sense in this context?

What would make most sense is:
```
      0°
⤹90°      90°⤸
     180°
```

yparitcher:
for now:
Reader_2020-06-29_230139
Reader_2020-06-29_230148

NiLuJe:
Indeed, I was going to suggest that, with the proper arrow, the `-` didn't make any sense ;).

In the same vein, I'd possibly go with an ↓ for 180°?

----

Incidentally, the discrepancy between screen rotation and buffer rotation is where the `-90° *` shenanigans come in in `setRotationMode`: while the argument denotes a *screen* rotation (i.e., linuxfb), what the function is responsible for is rotating the *blitbuffer* ;).

NiLuJe:
And, following that last comment (i.e., speaking in  terms of *screen* rotation), `⤹90°` ought to be `3`, not `1` ;).

yparitcher:
FileManager_2020-06-29_231145
Reader_2020-06-29_231117
Reader_2020-06-29_231124

yparitcher:
> And, following that last comment (i.e., speaking in terms of _screen_ rotation), `⤹90°` ought to be `3`, not `1` ;).
> 
> Making the table, as currently labeled, `{3, 0, 1, 2}`.

do you mean device?

NiLuJe:
> do you mean device?

(https://github.com/koreader/koreader/pull/6309#issuecomment-651490226)

> (For all intents and purposes here, I'll happily conflate device with screen, since the device is 95% screen anyway ;)).

NiLuJe:
And for more extra nit-pickiness, I'd assume S.PORTRAIT to be 0 and S.LANDSCAPE to be 1 ;).

yparitcher:
when i read screen I hear buffer.

for every rotation the device goes 90° one way or the content goes 90° the other way.

NiLuJe:
> when i read screen I hear buffer/content.
> 
> for every rotation the device goes 90° one way or the content goes 90° the other way.

Which is why I initially said *device* ^^. (I'd have the same reflex with the way the LinuxFB comment refers to a *display*).

yparitcher:
> And for more extra nit-pickiness, I'd assume S.PORTRAIT to be 0 and S.LANDSCAPE to be 1 (⤸ 90°) ;).

that is changing the current (before this PR) default

NiLuJe:
> > And for more extra nit-pickiness, I'd assume S.PORTRAIT to be 0 and S.LANDSCAPE to be 1 (⤸ 90°) ;).
> 
> that is changing the current (before this PR) default

That's a bit more complex than that because of the whole LANDSCAPE_SHOULD_GO_CCW DEFAULT ;) (which, IIRC, is disabled by default, hence Landscape == CW == 1, matching LinuxFB, which is mainly what I'm getting at ^^.

NiLuJe:
Actual variable name being `DLANDSCAPE_CLOCKWISE_ROTATION`, and being indeed `true` by default.

(Speaking of, this makes it obsolete, too, for your matching base ffi/framebuffer.lua PR ;)).

NiLuJe:
(Probably harmless, but the old `DEFAULT`s are also listed in `.luacheckrc` ;)).

yparitcher:
imageviewer still uses DLANDSCAPE_CLOCKWISE_ROTATION

NiLuJe:
> imageviewer still uses DLANDSCAPE_CLOCKWISE_ROTATION

Ah, true, good catch. Should probably be refactored later to actually read the current rotation mode instead, or something.

Frenzie:
> when i read screen I hear buffer/content.
>
> for every rotation the device goes 90° one way or the content goes 90° the other way.

The user shouldn't be bothered with the framebuffer going against what's usual for what are hopefully very good technical reasons. ^_^

frontend/ui/elements/screen_rotation_menu_table.lua
```diff
@@ -0,0 +1,124 @@
...
+            end,
+            callback = function()
+                G_reader_settings:flipNilOrFalse("input_ignore_gsensor")
+                Device:toggleGSensor(not G_reader_settings:isTrue("input_ignore_gsensor"))
+            end,
+        })
+    end
+
+    table.insert(rotation_table, {
+        text = _("Lock rotation"),
```
Frenzie, line 28: What about:
```suggestion
        text = _("Keep file browser rotation"),
```

frontend/ui/elements/screen_rotation_menu_table.lua
```diff
@@ -0,0 +1,124 @@
...
+local Event = require("ui/event")
+local FileManager = require("apps/filemanager/filemanager")
+local UIManager = require("ui/uimanager")
+local Screen = Device.screen
+local S = require("ui/data/strings")
+
+return {
+    text = _("Rotation"),
+    sub_item_table_func = function()
+    local rotation_table = {}
```
Frenzie, line 12: Is all this purposefully not indented?

NiLuJe:
Okay, not discussing this stuff at 5AM helps ;p.

I took another look at the LinuxFB stuff, and, it turns out...

```
 /*
  * Display rotation support
  */
 #define FB_ROTATE_UR      0
 #define FB_ROTATE_CW      1
 #define FB_ROTATE_UD      2
 #define FB_ROTATE_CCW     3
````

but...

```
>-------__u32 rotate;>-->------->-------/* angle we rotate counter clockwise */
```

i.e., what's user-facing is *device* orientation (in CW steps), and it's the responsibility of the *kernel* to make that a *buffer* rotation (in CCW).

I'd *very much* like to keep that philosophy here, since it's already what `setRotationMode` does.

Which means:

LANDSCAPE_INVERTED = 3 = CCW = ⤹90°
PORTRAIT = 0 = UR = ↑0°
LANDSCAPE = 1 = CW = ⤸90°
PORTRAIT_INVERTED = 2 = UD = ↓180°

**Comment by Frenzie** [comment]
(or rebase; I just noticed you force pushed some nicely rebased commits unlike most people :+1: )

File: frontend/ui/elements/screen_rotation_menu_table.lua [hunk 7]
```diff
@@ -0,0 +1,124 @@
...
+                G_reader_settings:flipNilOrFalse("lock_rotation")
+            end,
+            separator = true,
+        })
+
+        if FileManager.instance then
+            table.insert(rotation_table, {
+                text_func = function()
+                    local text = S.LANDSCAPE_ROTATED
+                    if G_reader_settings:readSetting("fm_rotation_mode") == 3 then
```
NiLuJe, line 43: Could probably use the fancy constants from https://github.com/koreader/koreader-base/blob/68b3a7c5d74cb77d82a9eff5b9d9dd9c5bb368f7/ffi/framebuffer.lua#L51-L55 throughout, instead of magic numbers ;).

spec/unit/readerrolling_spec.lua
```diff
@@ -108,7 +111,7 @@ describe("Readerrolling module", function()
 
     describe("test in landscape screen mode", function()
         it("should go to landscape screen mode", function()
-            readerui:handleEvent(Event:new("ChangeScreenMode", "landscape"))
+            readerui:handleEvent(Event:new("SetRotationMode", 1))
```
NiLuJe, line 114: Ditto

spec/unit/screenshoter_spec.lua
```diff
@@ -9,19 +9,21 @@ describe("ReaderScreenshot module", function()
...
 
         readerui = ReaderUI:new{
+            dimen = Screen:getSize(),
             document = DocumentRegistry:openDocument(sample_epub),
         }
     end)
 
     teardown(function()
-        readerui:handleEvent(Event:new("ChangeScreenMode", "portrait"))
+        readerui:handleEvent(Event:new("SetRotationMode", 0))
```
NiLuJe, line 21: Ditto

frontend/ui/data/creoptions.lua
```diff
@@ -19,14 +19,15 @@ local CreOptions = {
...
+                name = "rotation_mode",
                 name_text = S.SCREEN_MODE,
-                toggle = {S.PORTRAIT, S.LANDSCAPE},
+                toggle = {S.LANDSCAPE_ROTATED, S.PORTRAIT, S.LANDSCAPE, S.PORTRAIT_ROTATED},
                 alternate = false,
-                args = {"portrait", "landscape"},
-                default_arg = "portrait",
-                current_func = function() return Device.screen:getScreenMode() end,
-                event = "ChangeScreenMode",
+                values = {3, 0, 1, 2},
```
NiLuJe, line 26: Ditto

frontend/ui/data/koptoptions.lua
```diff
@@ -26,14 +26,15 @@ local KoptOptions = {
...
+                name = "rotation_mode",
                 name_text = S.SCREEN_MODE,
-                toggle = {S.PORTRAIT, S.LANDSCAPE},
+                toggle = {S.LANDSCAPE_ROTATED, S.PORTRAIT, S.LANDSCAPE, S.PORTRAIT_ROTATED},
                 alternate = false,
-                args = {"portrait", "landscape"},
-                default_arg = "portrait",
-                current_func = function() return Screen:getScreenMode() end,
-                event = "SwapScreenMode",
+                values = {3, 0, 1, 2},
```
NiLuJe, line 33: Ditto

pazos:
@yparitcher: Indeed this PR fixes the underlying issue with rotation on Android Tv. But the platform is not supported yet. Is just a lot easier to add proper support with your patch :smile:

NiLuJe:
Many thanks for cleaning up this mess :).

poire-z:
I tested this for the first time this morning, and I was quite confused by the bottom menu rotated arrows:

I hit the 90° buttons, and the screen rotate in the other direction than the arrows indicate :|

It has been discussed above - but I still find that confusing.

Currently, to me, the arrows actually mean: _"hit that button, and **after** hitting, this is how you should rotate your device to get text readable"_.

Which is counterclockwiseintuitive :)
If the current behaviour has really some sense, it might need a `help_text=` to explain a bit how to think about it.

Frenzie:
@poire-z What exactly has been discussed above? You mean that what arrows only really make sense in the default position? I don't understand what you mean by the rotation being the other direction.

Screenshot_2020-07-06_10-39-35

poire-z:
> What exactly has been discussed above? 

Whether the screen or the device is rotating - somewhere, didn't re-read.

> I don't understand what you mean by the rotation being the other direction.

To me, your screenshot is what I would expect when hitting image (the title "200 words" going to the right of screen) - but it actually turns the other way around : your title "200 words" will be on the left of the screen.

Frenzie:
Oh, you're right. It technically rotates the wrong way. I didn't realize because it's a lot harder to grok than a simple 0/90/180/270 and it feels right because you just follow the arrow.

yparitcher:
> What exactly has been discussed above? 

for the 2 viewpoints see:

https://github.com/koreader/koreader/pull/6309#issuecomment-651487406

> Okay, wait, we're talking about two different things ^^.
> 
> I meant rotating the _device_ physically. The fb content indeed goes -90° CCW/270° CW ;).
> 
> And I do think that talking in terms of _device_ rotation is the most user-friendly way to display it ;).

https://github.com/koreader/koreader/pull/6309#issuecomment-651488660


> I always assumed the device stays the same and the screen rotates ;) (What actually happens)
> this is why I like the current way (CCW)
> This is more natural to me
> 
> I think _screen_ rotation is the way to go :)

NiLuJe:
See https://github.com/koreader/koreader-base/pull/1137 ;).

It possible makes more sense if you actually have a device in your hands, and not the emulator and/or a giant screen ;).

NiLuJe:
Well, actually, Nickel's is mildly unhelpful in that it's an *orientation* icon, not a *rotation* one:

nickel_rota

The modern Kindle one is similar, IIRC (with a few straight lines in the middle to indicate text).

The one I was actually thinking about is the old K4 one, which *did* actually denote a *rotation*, and which makes the fact that this is a device rotation exceedingly clear:

kindlel_rota

Frenzie:
@NiLuJe 
> It possibly makes more sense if you actually have a device in your hands, and not the emulator and/or a giant screen ;).

Not just possibly. On my PC I mildly agree with @poire-z and @yparitcher but on my H2O (or phone) my opinion's the reverse. It feels completely natural the way it is right now.

NiLuJe:
Ideally, I'd have gone with icons similar to the K4 stuff to make that clearer, but we don't have any ;/.

Fun fact: it wasn't using the LinuxFB standard, that was custom [einkfb](https://github.com/NiLuJe/FBInk/blob/435f92abca05bfb4a0602ea68236b9c75c2a02f9/eink/einkfb.h#L266-L272) implementation.

```
[FBInk] Variable fb info: 600x800, 8bpp @ rotation: 0 (Upright, 0°)
[FBInk] Actual einkfb orientation: 0 (Portrait, 0°)


----

[FBInk] Variable fb info: 800x600, 8bpp @ rotation: 0 (Upright, 0°)
[FBInk] Actual einkfb orientation: 2 (Landscape, 90°)

----

[FBInk] Variable fb info: 600x800, 8bpp @ rotation: 0 (Upright, 0°)
[FBInk] Actual einkfb orientation: 1 (Inverted Portrait (Upside Down), 180°)

----

[FBInk] Variable fb info: 800x600, 8bpp @ rotation: 0 (Upright, 0°)
[FBInk] Actual einkfb orientation: 3 (Inverted Landscape (Upside Down), 270°)
```

poire-z:
> It possibly makes more sense if you actually have a device in your hands, and not the emulator and/or a giant screen ;).

I had tested it on a small Glo HD.

> on my H2O (or phone) my opinion's the reverse. It feels completely natural the way it is right now.

How do you explain that?

It feels the inverse of what I expect on both a small device and a big emulator.
BUT:
- it feels the inverse when I'm in portrait and I want to switch to landscape - as an action
- when I'm in landscape, and I show the menu, it represents no more an anction - but a state: and then, the curved arrow shows that something has been rotated that way. And there, it might feel OK/natural to consider showing how the device itself is rotated - I think I could also be ok to consider that it's how the screen has been rotated (but may be my mind then needs an additional 200ms to think _the screen was that way, it has been turned that way_, that might be not needed when considering it's the device rotated in my hand... :)

So, dunno.
But the initial meeting with this menu, when you are in portrait, is surprising, because one intuitively considers the arrow to indicate an action.
On a device with a gyro, you mostly always expect this as showing the state as you won't be using it as an action - so I understand it might be not surprising like it was to me (who never switch to landscape :)
So, the arrows are what make that ambiguous.

(If we had an icon with a device, with some obvious orientation, that would not make it strange on a PC/TV/phone - we have an ImageWidget that accepts a `rotation_angle` parameter :) but that would need some addition to the ugly ConfigDialog :/ )

Frenzie:
> How do you explain that?

Because I don't turn my monitor. Initially I didn't understand what your problem was because I was using it on my H2O. It's only when I thought about it that I realized that the super intuitive device rotation is technically "wrong."

Frenzie:
> when I'm in landscape, and I show the menu, it represents no more an anction - but a state: and then, the curved arrow shows that something has been rotated that way. And there, it might feel OK/natural to consider showing how the device itself is rotated - I think I could also be ok to consider that it's how the screen has been rotated (but may be my mind then needs an additional 200ms to think the screen was that way, it has been turned that way, that might be not needed when considering it's the device rotated in my hand... :)

Like I said, *that*'s what I thought you were talking about. That's kind of weird.

NiLuJe:
Much like the K4 example above, ti's technically *always* a state.

It just feels liek both in Portrait, because Portrait is the reference point ;).

Incidentally, this is why I wanted to keep Portrait between 90°CCW and 90°CW instead of keeping it at the front, I felt like it kinda helped visualize that better.

Issue #6311, opened 2020-06-26
Bug: Pocketbook Basic 3 not recognized + other


roshavagarga:
* KOReader version: 2020.06
* Device: Pocketbook Basic 3 (614W or 614-2?)
* Device firmware: [5.17.964](http://download.pocketbook-int.com/fw/Yitoa/614-2/user/20171123/sw_20171123_614-2_U_5.17.964.user.zip) (23rd November 2017, discontinued firmware support)
#### Issue
Merged both folders, opening a book does nothing, neither does opening the app.

#### Steps to reproduce
Merge folders (applications + system), disconnect, try to open the app from 'applications' menu or a book from the main menu. Screen flickers for a second and goes back to where it was.

##### `crash.log` (if applicable)
[crash.log](https://github.com/koreader/koreader/files/4836025/crash.log)

Frenzie:
```
06/26/20-11:53:59 INFO  SoftwareVersion:  U614-2.5.17.964
./luajit: frontend/device/pocketbook/device.lua:379: unrecognized PocketBook model PocketBook 614W
```

You could replace `"PocketBook 613"` by `"PocketBook 614W"` on this line for example to see if it works. The exact device definition may be slightly different but it should suffice for basic testing.

https://github.com/koreader/koreader/blob/43693c9fe9f03e7862f268bfe4b40e1d4bed3fbf/frontend/device/pocketbook/device.lua#L344

The specs at https://productz.com/en/pocketbook-basic-3-614/p/zlYX suggest it's extremely similar to the 613.

roshavagarga:
@Frenzie Already saw that change and I've been tinkering with my zero knowledge.

Added the following to device.lua:

1. New device at lines 233 to 242 (+ two empty lines, one before and one after)
`-- PocketBook 614W Basic
local PocketBook614W = PocketBook:new{
    model = "PB614W",
    display_dpi = 166,
    isTouchDevice = no,
    hasWifiToggle = no,
    hasFrontlight = no,
    hasDPad = yes,
    hasFewKeys = yes,
}`
2. New device model at lines 355 and 366 (codename decided based on existing model file in original system folder)
`elseif codename == "PocketBook 614W" then
    return PocketBook614W`

Now, I'm not entirely sure how the hasDPad and hasFewKeys is supposed to work, but I have noticed this so far:
[...]

PR #6312, opened 2020-06-26
Add PocketBook 614W (Basic 3) definition

roshavagarga:
Cf. <#6311>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6312)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -240,7 +240,8 @@ local PocketBook614W = PocketBook:new{
     hasWifiToggle = no,
     hasFrontlight = no,
     hasDPad = yes,
-    hasFewKeys = yes, }
+    hasFewKeys = yes,
+}
 
 -- PocketBook Basic Lux 2
 local PocketBook616 = PocketBook:new{
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -232,6 +232,16 @@ local PocketBook613 = PocketBook:new{
     hasFewKeys = yes,
 }
 
+-- PocketBook 614W Basic
+local PocketBook614W = PocketBook:new{ 
+    model = "PB614W",
+    display_dpi = 166,
+    isTouchDevice = no,
+    hasWifiToggle = no,
+    hasFrontlight = no,
+    hasDPad = yes,
+    hasFewKeys = yes, }
+
 -- PocketBook Basic Lux 2
 local PocketBook616 = PocketBook:new{
     model = "PBBLux2",
@@ -343,6 +353,8 @@ if codename == "PocketBook 515" then
     return PocketBook515
 elseif codename == "PocketBook 613" then
     return PocketBook613
+elseif codename == "PocketBook 614W" then
+    return PocketBook614W
 elseif codename == "PB616W" or
     codename == "PocketBook 616" then
     return PocketBook616
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -235,7 +235,7 @@ local PocketBook613 = PocketBook:new{
 -- PocketBook 614W Basic
 local PocketBook614W = PocketBook:new{ 
     model = "PB614W",
-    display_dpi = 166,
+    display_dpi = 167,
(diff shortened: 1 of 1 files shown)


frontend/device/pocketbook/device.lua
```diff
@@ -232,6 +232,16 @@ local PocketBook613 = PocketBook:new{
...
 
+-- PocketBook 614W Basic
+local PocketBook614W = PocketBook:new{ 
+    model = "PB614W",
+    display_dpi = 166,
+    isTouchDevice = no,
+    hasWifiToggle = no,
+    hasFrontlight = no,
+    hasDPad = yes,
+    hasFewKeys = yes, }
```
Frenzie, line 243:
```suggestion
    hasFewKeys = yes,
}
```

roshavagarga:
Just a small overview of those changes:
1. I did not try dpi of 167, but it might work, just saw 166 in device specs, either one may be correct.
2. I'm not sure what hasWifiToggle means, I guessed that it meant having a button to enable/disable WiFi, so I set it as no based on that.
3. There may be other tags/options that I might have missed which may be appropriate for this device.

Frenzie:
> I did not try dpi of 167, but it might work, just saw 166 in device specs, either one may be correct.

Please stick to 167 for consistency. :-) (Also 166.6666… rounds up to 167.)

> I'm not sure what hasWifiToggle means, I guessed that it meant having a button to enable/disable WiFi, so I set it as no based on that.

It's about if you can just tell it to toggle the wifi on/off, not a physical toggle iirc.

roshavagarga:
Changed it for consistency, but other devices seem to have it set to 166 instead of 167, so those might need to be changed too?

roshavagarga:
Within the native settings UI, there is a wifi menu, which has an on/off switch, so if that constitutes a Wifi Toggle, then I should probably change that too, right?

frontend/device/pocketbook/device.lua
```diff
@@ -232,6 +232,17 @@ local PocketBook613 = PocketBook:new{
     hasFewKeys = yes,
 }
 
+-- PocketBook 614W Basic
+local PocketBook614W = PocketBook:new{
```
Frenzie, line 236:
```suggestion
local PocketBook614W = PocketBook:new{
```

roshavagarga:
Thanks for the merge, taking it that the wifi thing I found isn't a toggle :+1:

Frenzie:
> Changed it for consistency, but other devices seem to have it set to 166 instead of 167, so those might need to be changed too?

If they round to 167, probably. It shouldn't affect anything other than perhaps people with OCD though. ^_^

> Within the native settings UI, there is a wifi menu, which has an on/off switch, so if that constitutes a Wifi Toggle, then I should probably change that too, right?

I didn't really notice this before I merged, but that's more of a thing to determine experimentally. Does the wifi toggle in the KOReader menu work? That's Settings → Network → Wi-Fi connection. You'll probably have to uncomment or remove that `hasWifiToggle` line to test.

roshavagarga:
I'll test it out later, not sure I can enter that part of settings because of the whole hasFewKeys issues.

Frenzie:
You can.

1. Open a book.
2. Press right.
3. Navigate to settings (but I just realized it doesn't rotate, so that might be problematic if you're too far right)

roshavagarga:
If I remember correctly, I was stuck at the furthest to the right, if that one has it, I can test it, if not, no dice :)
Actually, now that I think about it, if both menus rotated and the 'left for leave' thing worked on the bottom menu like on the top one, it would be pretty usable even without remapping keys.

Frenzie:
> If I remember correctly, I was stuck at the furthest to the right, if that one has it, I can test it, if not, no dice :)

You can set the last opened menu in settings.reader.lua.

```
["filemanagermenu_tab_index"] = 4,
-- and probably readermenu_tab_index
```

Removing the line will reset it.

> Actually, now that I think about it, if both menus rotated and the 'left for leave' thing worked on the bottom menu like on the top one, it would be pretty usable even without remapping keys.

Well yes, that's what I proposed, for starters anyway. That's very much with remapping keys though? ;-)

PR #6314, opened 2020-06-26
emulator: add fake network manager

### LINKED ISSUE #6080: koreader don't do user request if need connect to the wifi
* KOReader version: v2020.04.1-1-***_2020-04-19
* Device: Kindle pw3

#### Issue
if your kindle not connect to wifi and you try to do something that need internet (for example update or translate) it show a dialog box that ask you to turn on modem but after that koreader close that dialog box but don't do user requested action.
so user should repeat that step again
#### Steps to reproduce
1. wifi should be off
2. go to menu > update > check for update
3. dialog box ask for turn on wifi so press turn on
update don't start  until you repeat step 2 again
#### Expected behavior
after step 3 it should do this steps
4. show small icon in corner that show device try to connect wifi
5. if device connect to wifi ,icon update to an wifi icon with a check mark .if it not show a wifi icon with X mark on it.(program should go to step 6 but hide icon after 2 or 3 second)
6. koreader do the command that user request (translate or update or any other action that need internet)

pazos:
Useful to test things that rely on network state changes, like #6299 

Probably useful to fix #6080

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6314)

### DIFF
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -329,6 +329,29 @@ function Device:simulateResume()
     })
 end
 
+-- fake network manager for the emulator
+function Emulator:initNetworkManager(NetworkMgr)
+    local connectionChangedEvent = function()
+        local UIManager = require("ui/uimanager")
+        if G_reader_settings:nilOrTrue("emulator_fake_wifi_connected") then
+            UIManager:broadcastEvent(Event:new("NetworkConnected")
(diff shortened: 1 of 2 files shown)


pazos:
> Btw, if you want to get really fancy you could go even further

This is certainly one of these times where I don't want to get fancy ;-p.

Issue #4029, opened 2018-06-29
Won't start on PocketBook 613


TenMirim:
Firstly, i know the app isn't designed for this low-end reader, just wanted to warn others that it **won't** work there since it crashes on start (from what i read from crash report it seems it can't detect it). 
* KOReader version: beta 20180602
* Device: PocketBoot 613 Basic
It doesn't even load up, stops on the clock icon, crashes and returns to the previous menu.
Crash log: [crash.log](https://github.com/koreader/koreader/files/2150056/crash.log)

Frenzie:
To test if it works you'll just need to add the codename. For example, change `elseif codename == "PB627" then` to `elseif codename == "PocketBook 613" then`.

https://github.com/koreader/koreader/blob/ca6c7316e6fa550f511db72f9c1cdefd1775a2fe/frontend/device/pocketbook/device.lua#L295-L296

TenMirim:
Oh, looks like it booted after creating PB 613 device in `frontend/PocketBook/device.lua` , but the app can't be controlled, because it probably uses different/non-standard key layout. Turning the "has-touch" option off didn't help. Just for context, it has only D-pad, button in middle of the D-pad, power button and two page navigating buttons, but can't find any official names for them.

Frenzie:
If you add `-d` to the script and touch all the keys it'll log their values.

You did put in hasKeys = yes? And it actually doesn't have a touch screen?

NB There's also something about screen refreshes, so please grab the latest nightly instead of the stable.

TenMirim:
No, it really doesn't have a touch screen. And it looks like it's receiving something, i can turn pages, but can't access menu or anything, really. Where am I supposed to put the argument? Somewhere in the .app file?

Frenzie:
Indeed, if you add -d on this line like so:

https://github.com/koreader/koreader/blob/8ecdf9d88814e19ba6b3aeae2db9167a70f90740/platform/pocketbook/koreader.app#L47

```
./reader.lua -d "${args}" >>crash.log 2>&1 
```

The key codes seem to be defined here:
https://github.com/koreader/koreader/blob/8ecdf9d88814e19ba6b3aeae2db9167a70f90740/frontend/device/pocketbook/device.lua#L21-L38

You seem to be having an issue with the menu key?

TenMirim:
[...]

PR #6318, opened 2020-06-27
[UX] Add left to close to ConfigDialog for hasFewKeys

Frenzie:
Follow-up to <https://github.com/koreader/koreader/pull/6195>.

Part of <https://github.com/koreader/koreader/issues/4029>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6318)

### DIFF
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -824,7 +824,8 @@ function ConfigDialog:init()
     end
     if Device:hasKeys() then
         -- set up keyboard events
-        self.key_events.Close = { {"Back"}, doc = "close config menu" }
+        local close_keys = Device:hasFewKeys() and { "Back", "Left" } or "Back"
+        self.key_events.Close = { { close_keys }, doc = "close config menu" }
     end
     if Device:hasDPad() then
         self.key_events.Select = { {"Press"}, doc = "select current menu item" }


Frenzie:
@roshavagarga This combined with https://github.com/koreader/koreader/pull/6315 should take care of the worst.

roshavagarga:
Thanks for the heads up and for the hard work! :)

Frenzie:
Btw, besides manually replacing the files you can always grab the latest nightly from http://build.koreader.rocks/download/nightly/

That's the same as running update from within the program (after selecting the "development" channel) but that may be problematic without the horizontal wrap around in place.

roshavagarga:
@Frenzie Installed 20.07.1 (hadn't checked the device since these merges) - left as close works for most windows, but does not work for the statistics window. Getting out of that one requires a full-blow shutdown and restarting the device, since no buttons help.

Frenzie:
@roshavagarga It's supposed to close on any keypress whatsoever, but the problem is that it's only on hasKeys() and not on hasDPad. It's probably an issue just with statistics and not with the overall logic behind these things but I'm not 100% sure. :-)

https://github.com/koreader/koreader/blob/b4d5cfccb79cf749bee83311b9b54899c1052c19/plugins/statistics.koplugin/readerprogress.lua#L60-L66

Frenzie:
Could you check if this works satisfactorily?

```diff
diff --git a/plugins/statistics.koplugin/readerprogress.lua b/plugins/statistics.koplugin/readerprogress.lua
index 84c3650a..527a712c 100644
--- a/plugins/statistics.koplugin/readerprogress.lua
+++ b/plugins/statistics.koplugin/readerprogress.lua
@@ -57,7 +57,7 @@ function ReaderProgress:init()
     UIManager:setDirty(self, function()
         return "ui", self.dimen
     end)
-    if Device:hasKeys() then
+    if Device:hasDPad or Device:hasKeys() then
         self.key_events = {
             --don't get locked in on non touch devices
             AnyKeyPressed = { { Device.input.group.Any },
```

roshavagarga:
@Frenzie Nope, it does not.
Tried your line, also tried:
1. `if Device:hasDPad or Device:hasFewKeys or Device:HasKeys() then`
2. `if Device:hasDPad() or Device:hasFewKeys() or Device:HasKeys() then` 

None of the above worked. The left and right buttons weirdly enough let me scroll through the 3 stats pages, but the DPad does nothing.

Frenzie:
@roshavagarga That's combined with the relevant device definitions? It works for me.

roshavagarga:
@Frenzie 2020.07.1 stable on a PocketBook 614W, which already had a proper definition - changing that plugin file does nothing for me when I open device stats, issue persists sadly.

Issue #2721, opened 2017-04-04
FR: double tap to go back or forward 10 pages


Frenzie:
Split out from https://github.com/koreader/koreader/issues/2410#issuecomment-267044394

@retrue wrote:
> Another gesture that I would like to see implemented in KOReader is the option to advance or go back 10 pages. It makes browsing pages inside a book nimbler. It is an option that I miss and was included in Coolreader in my first e-reader, a Jinke Hanlin v3+. And I think a double tap would be the best gesture in this case.

poire-z:
Let's not forget the following comments:

From me:
> The problem with double tap is that it will slow down reaction of other single taps (you can enable double tap in the Screen menu to see).

To which houqp replied:
> This is a bug in the double tap detection code (gesturedetector.lua). I suspect it's problem of finding the right value for DOUBLE_TAP_INTERVAL. If it's too high, wait time for single tap detection will increase, if it's too low, it would be too hard for user to perform a double tap.

I'm just not sure it's a bug, just a matter of "finding the right value", which will slow down single tap reaction anyway, no ?
But may be this existing setting ["disable_double_tap"] = true is enough to disallow double tap waiting in the gesture detector, and any double_tap feature would be disabled for those who prefer quick reactions, and those who want double tap, but slower reaction to single tap, can let it to false ?

Frenzie:
Right, good point. I'd like to add that I consider the implementation details secondary to the request. I'd probably assign a new action to swiping east and west at the bottom of the screen. This largely comes down to #1008. It'd also be cool to do something like `left, down` for previous chapter and `right, down` for next chapter. But just like how my new menu implementation enables changing these things sensibly, something similar is really required for registering gestures. And of course GestureDetector doesn't currently have support for chaining gestures like that in the first place, which would hardly be trivial either.

The 300ms used to be pretty standard (and horribly annoying) in mobile browsers for zoom: http://www.telerik.com/blogs/what-exactly-is.....-the-300ms-click-delay

I think that 100ms feels *a lot* better, although it depends on the document in question. A borderline slow document suddenly becomes slow that way. But while I think 300ms is an insensibly long default, the fact that I'd personally go for something closer to 50ms or less (I'd have to test) would similarly not work. I'd say that double-tap timeout is definitely the kind of thing that is served by a setting.

retrue:
[...]

PR #6322, opened 2020-06-27
[UX] Move double tap to gesture manager

Frenzie:
Also replaces the default with +10/-10 pages instead of prev/next chapter.

Closes <https://github.com/koreader/koreader/issues/2721>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6322)


frontend/apps/reader/modules/readergesture.lua
```diff
@@ -352,6 +360,41 @@ function ReaderGesture:addToMainMenu(menu_items)
...
+                enabled_func = function() return self.ges_mode == "gesture_reader" end,
+                sub_item_table = self:buildMenu("double_tap_left_side", self.default_gesture["double_tap_left_side"]),
+            },
+            {
+                text_func = function() return actionTextFunc("double_tap_right_side", _("Right side")) end,
+                sub_item_table = self:buildMenu("double_tap_right_side", self.default_gesture["double_tap_right_side"]),
+            },
+            {
+                text_func = function() return actionTextFunc("double_tap_top_left_corner", _("Top left")) end,
+                enabled_func = function() return self.ges_mode == "gesture_reader" end,
```
poire-z, line 380: Why does that top left one has this enabled_func, and not the others?

Frenzie: Copy paste oversight. Top left (tap) isn't implemented in the gesture manager in the file manager.

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -191,6 +191,12 @@ function ReaderGesture:init()
         one_finger_swipe_top_edge_left = "ignore",
         one_finger_swipe_bottom_edge_right = "ignore",
         one_finger_swipe_bottom_edge_left = "ignore",
+        double_tap_left_side = self.ges_mode == "gesture_reader" and "prev_chapter",
+        double_tap_right_side = self.ges_mode == "gesture_reader" and "next_chapter",
```
poire-z, line 195: > Also replaces the default with +10/-10 pages instead of prev/next chapter.

Not yet done ?

Frenzie: Whoops, I thought I'd picked one. I couldn't quite decide whether it was better to stick to the current or change it. Thanks for being vigilant. :-)

(Any thoughts?)

poire-z: (not really, I have double tap disabled - May be +10/-10 is better for skimming, some books might not have any chapter)

Frenzie: Yeah, I don't use it either. As I said in the referenced issue I don't like the delay, even when I make it about a 100. Much lower than that isn't practical.

Frenzie:
@poire-z Any other comments?

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -243,6 +249,8 @@ function ReaderGesture:init()
     local mirrored_if_rtl = {
         tap_top_left_corner = "tap_top_right_corner",
         tap_right_bottom_corner = "tap_left_bottom_corner",
+        double_tap_left_side = "double_tap_right_side",
+        double_tap_right_side = "double_tap_left_side",
```
poire-z, line 253: (I don't remember why I just set tap_top_left_corner  and tap_right_bottom_corner, and not their counterparts.
Oh, but I switch them, so one of the swapped pair is enough. But it's strange I made that kinda crossed, and not both: left_corner = "right" ...)

Issue #5768, opened 2020-01-14
[FR/Android]: Simulate light warmth and extra dimming on lcd screens


pazos:
I'm aware that KOReader for android is full of tricks. That's the last one I want to implement :smile: 

That's it: implement an overlay on top of KOReader window, so features like screen dimming beyond hardware levels and light warmth can be simulated.

At first tried to do that on top of KOReader itself, but given the record of bugs that using a surfaceView on top of the native view had on some devices I would go for implementing this outside the program.

There are plenty of 3rd party screen filters that can do that, but the hability to use slide gestures within KOReader would be awesome. I commented on a issue opened at [Red Moon for 3rd party integrations](https://github.com/LibreShift/red-moon/issues/202)

While I'm still thinking that integration with free apps is the way to go I played a bit with a standalone implementation for us based on a service, that needs to be installed as an extension to KOReader.

In any case the behaviour would be the same: check if a 3rd party service is enabled, bound the activity to it, promote the device as one with natural light support and forward messages to the service, which is the one that draws the overlay.

The task is feasible and has the same latency as in my ereader, but the whole idea sound hackish as hell, so I'm looking for feedback about the implementation. 

Shall I write it as part of core app or as a plugin?.

How to distribute the apk that needs to be installed? I guess it can be downloaded using the same method used for updates.

Frenzie:
What do you mean by plugin or part of the core precisely? It sounds like the amount of Lua code would be fairly minimal so I figure that should be fine in some Android-specific device implementation. Depending on some other app sounds quite fine to me too, to reduce complexity.

pazos:
I think I will work a bit on a example app to show you how it behaves and then ask the questions, because explain how it behaves is 100x harder for me :smile:

pazos:
Ok, I've implemented a service, intended to run as a "server", which holds permissions to start on boot and to draw overlays on top of other activities.

It implements a simple interface for other programs:

```aidl
// IService.aidl
package org.koreader.aidl;

interface IService {
 	String status();
 	void color(int color);
 	void alpha(float alpha);
}
```

And a test application to, err, test.

![testing](https://user-images.githubusercontent.com/975883/72619384-68aa4900-393d-11ea-8ac4-a9261f522deb.gif)

As you can see this simulates warmth. Probably high values (above 5000k) and low values (below 1400k) are not interesting.
New to add a view to simulate extra dim steps, but that's easy.

The code used on the test is

```java
class ColorHelper {
[...]

PR #6327, opened 2020-06-28
[wip]: add support for android TV

### LINKED ISSUE #5725: FR: Android TV

pazos:
Fixes #5725

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6327)


frontend/device/android/device.lua
```diff
@@ -67,7 +67,9 @@ local Device = Generic:new{
     isAndroid = yes,
     model = android.prop.product,
     hasKeys = yes,
-    hasDPad = no,
+    hasDPad = function() return android.isTv() end,
+    hasFewKeys = function() return android.isTv() end,
```
Frenzie, line 71: Aha! I knew it would pay off someday. :-)

(I'd think TV remotes had more buttons though?)

pazos: The documentation says: https://developer.android.com/training/tv/start/controllers

So we have 6 buttons. (4 dpad, back and select). I use `hasFewKeys` as it opens the top menu in the reader with the right dpad button. Some things are still rough (ie: we don't need to close the menu with left dpad button, back is ok; there's no way to open top menu in the FM...)

pazos:
Note to self:
- wait until the rotation is fixed for other devices.
- wait until a proper/native rotation is available on android
- implement device abstraction for Eink vs Lcd devices (https://github.com/koreader/koreader/issues/5768#issuecomment-652443499)
- implement android tv as a kind of Lcd device.

Finally figure out which is the best scheme to navigate the whole UI with dpad + back button.

frontend/device/android/device.lua
Frenzie: I see. On Kindle etc. we have a menu key for the top menu and an ok/select/whatever key for the bottom menu. Just a menu key lacking is a lot less bad, but kind of annoying that it's not sufficiently full-featured.

> there's no way to open top menu in the FM

Perhaps something could be done with left and right, e.g., by reversing it from the normal FM button layout. Left for context menu, right for top menu.

pazos: My suggestion would be:

## Reader
dpad-reader

## File manager
dpad-fm

## Menus
dpad-menu

Frenzie: The reason I suggested right for top menu in the file manager is so it'd be consistent with the reader, or was that one for the bottom menu? We should try to keep it internally consistent with the reader too, within reason. Overall the above proposal looks good! :+1:

pazos: Right now on the reader the center d-pad button opens the bottom menu. On the FM the right dpad button opens file menu.

Frenzie: In that case I added → for the top menu (on hasFewKeys).

pazos: That works on the reader, not on the filemanager, where it opens selected file/directory properties.

pazos:
Everything works fine as is with a controller or a bluetooth keyboard. We're still missing keys with the (5dpad + back) remote that comes with most Tv boxes. I'll leave this as a perpetual **WIP** until we figure the right key combos. My proposal is in https://github.com/koreader/koreader/pull/6327#discussion_r450787837, but I don't want to implement it right now :P

Commodore64user:
for the reader: if you simply add access to the bottom menu from the top one (like there is for the plus menu on the FM), you could free up the up/down keys. You could have:
left/right for page navigation
press or enter to call up the top menu (and bottom now too)
up/down are now free for other things like content selection on one and something else you value on the other.

for the menus: why not sacrifice the right and use that as long press?
up/down/left for navigation as you say
press or enter as select
right as long press

fm is about right that way.

doesn't that android documentation state that you have a menu button too? if that is the case you have even more leeway. otherwise consider my proposal.

PR #6332, opened 2020-06-30
add natural light for Tolino Epos 2 

zwim:
This is a first attempt to add natural light settings to the Tolino Epos 2. 
Currently the device is runnung on Android 4.4.2.
The device uses `/sys/class/backlight/tlc5947_bl/color` for the warmth, similar as on Kobo Forma.

There are some things to do, help is needed:

1. `/sys/class/backlight` is not writeable. But if the permissions are set with `chmod 666 /sys/class/backlight/tlc5947_bl/color` , the  warmth and the brightness of the backlight can be set.
~I do not know how to grant that permission within the app. Can anyone help with this?~
I decided to detect, if the color file is writeable. If not, the user gets an information about this. It he wants to get instructions to make the file writeable, a message window is shown.
Since every Tolino Epos2 has to be rooted to run koreader, this additional steps should be doable for advanced users.
 
~2. Safety check if `/sys/class/backlight/tlc5947_bl/color` is writeable, is not implemented yet. (easy)~ done

~3. Do a better check if the device is really a Tolino Epos 2, not just check `android.prop.product` (easy)~ not necessary

4. The brightness and warmth values have to be in the range 0-100, otherwise the slider of the warmth is broken. I don't know why.

5. The automatic warmth setting is duplicated from Kobo -> should be combined later.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6332)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -358,7 +358,7 @@ function Device:canExecuteScript(file)
 end
 
 local AndroidNaturalLight = Device:new{
-    model = "android_with_natural_lights",
+    model = "Android_with_natural_lights",
     hasNaturalLight = yes,
     frontlight_settings = {
         frontlight_white = "/sys/class/backlight/mxc_msp430_fl.0/brightness",
@@ -382,10 +382,10 @@ android.LOGI(string.format("Android %s - %s (API %d, product %s) - flavor: %s",
 
 local product = android.prop.product
 
--- todo: check if color frontlight_settings.frontlight_mixer is readable
-is_color_writeable = true
-
-if product == "ntx_6sl" and is_color_writeable then
+if product == "ntx_6sl" then
+    if lfs.touch(AndroidNaturalLight.frontlight_settings.frontlight_mixer) == nil then
+        AndroidNaturalLight.hasNaturalLight = no
+    end
     return AndroidNaturalLight
 else
     return Device
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -357,7 +357,36 @@ function Device:canExecuteScript(file)
     end
 end
 
-android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
-    android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
+local AndroidNaturalLight = Device:new{
+    model = "android_with_natural_lights",
+    hasNaturalLight = yes,
+    frontlight_settings = {
+        frontlight_white = "/sys/class/backlight/mxc_msp430_fl.0/brightness",
+        frontlight_mixer = "/sys/class/backlight/tlc5947_bl/color",
+        nl_min = 0,
+        nl_max = 10,
+        nl_inverted = true,
+    },
+}
 
-return Device
+function AndroidNaturalLight:init()
+    -- Automagically set this so we never have to remember to do it manually ;p
+    if self:hasNaturalLight() and self.frontlight_settings and self.frontlight_settings.frontlight_mixer then
+        self.hasNaturalLightMixer = yes
+    end
+    Device.init(self)
+end
+
+android.LOGI(string.format("Android %s - %s (API %d, product %s) - flavor: %s",
+    android.prop.version, getCodename(), Device.firmware_rev, android.prop.product, android.prop.flavor))
+
+local product = android.prop.product
+
+-- todo: check if color frontlight_settings.frontlight_mixer is readable
+is_color_writeable = true
+
+if product == "ntx_6sl" and is_color_writeable then
+    return AndroidNaturalLight
+else
+    return Device
+end
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -6,7 +6,7 @@ local _, android = pcall(require, "android")
 local AndroidPowerD = BasePowerD:new{
     fl_min = 0, fl_max = 100,
     fl_intensity = 10,
-    
+
     fl_warmth_min = 0, fl_warmth_max = 10,
     fl_warmth = nil,
     auto_warmth = false,
@@ -65,7 +65,6 @@ function AndroidPowerD:init()
         end
     end
 end
-
 
 function AndroidPowerD:setWarmth(warmth)
     if self.fl == nil then return end
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -1,9 +1,17 @@
 local BasePowerD = require("device/generic/powerd")
+local SysfsLight = require ("device/sysfs_light") -- for android eInk devices using natural light
+local PluginShare = require("pluginshare")
 local _, android = pcall(require, "android")
 
 local AndroidPowerD = BasePowerD:new{
-    fl_min = 0, fl_max = 25,
+    fl_min = 0, fl_max = 100,
     fl_intensity = 10,
+    
+    fl_warmth_min = 0, fl_warmth_max = 10,
+    fl_warmth = nil,
+    auto_warmth = false,
+    max_warmth_hour = 23,
+    fl_was_on = nil,
 }
 
 function AndroidPowerD:frontlightIntensityHW()
@@ -21,5 +29,104 @@ end
 function AndroidPowerD:isChargingHW()
     return android.isCharging()
 end
+
+function AndroidPowerD:init()
+    if self.device:hasFrontlight() then
+        -- If this device has natural light (currently only Android on Tolino Epos 2)
+        -- Use the SysFS interface, and ioctl otherwise.
+        -- NOTE: On the Forma, nickel still appears to prefer using ntx_io to handle the FL,
+        --       but it does use sysfs for the NL...
+        if self.device:hasNaturalLight() then
+            local nl_config = G_reader_settings:readSetting("natural_light_config")
+            if nl_config then
+                for key,val in pairs(nl_config) do
+                    self.device.frontlight_settings[key] = val
+                end
+            end
+            -- Does this device's NaturalLight use a custom scale?
+            self.fl_warmth_min = self.device.frontlight_settings.nl_min or self.fl_warmth_min
+            self.fl_warmth_max = self.device.frontlight_settings.nl_max or self.fl_warmth_max
+            -- If this device has a mixer, we can use the ioctl for brightness control, as it's much lower latency.
+            if self.device:hasNaturalLightMixer() then
+                local kobolight = require("ffi/kobolight")
+                local ok, light = pcall(kobolight.open)
+                if ok then
+                    self.device.frontlight_settings.frontlight_ioctl = light
+                end
+            end
+            self.fl = SysfsLight:new(self.device.frontlight_settings)
+            self.fl_warmth = 0
+        else
+            local kobolight = require("ffi/kobolight")
+            local ok, light = pcall(kobolight.open)
+            if ok then
+                self.fl = light
+            end
+        end
+    end
+end
+
+
+function AndroidPowerD:setWarmth(warmth)
+    if self.fl == nil then return end
+    if not warmth and self.auto_warmth then
+        self:calculateAutoWarmth()
+    end
+    self.fl_warmth = warmth or self.fl_warmth
+    -- Don't turn the light back on on legacy NaturalLight devices just for the sake of setting the warmth!
+    -- That's because we can only set warmth independently of brightness on devices with a mixer.
+    -- On older ones, calling setWarmth *will* actually set the brightness, too!
+    if self.device:hasNaturalLightMixer() or self:isFrontlightOnHW() then
+        self.fl:setWarmth(self.fl_warmth)
+    end
+end
+
+-- Sets fl_warmth according to current hour and max_warmth_hour
+-- and starts background job if necessary.
+function AndroidPowerD:calculateAutoWarmth()
+    local current_time = os.date("%H") + os.date("%M")/60
+    local max_hour = self.max_warmth_hour
+    local diff_time = max_hour - current_time
+    if diff_time < 0 then
+        diff_time = diff_time + 24
+    end
+    if diff_time < 12 then
+        -- We are before bedtime. Use a slower progression over 5h.
+        self.fl_warmth = math.max(20 * (5 - diff_time), 0)
+    elseif diff_time > 22 then
+        -- Keep warmth at maximum for two hours after bedtime.
+        self.fl_warmth = 100
+    else
+        -- Between 2-4h after bedtime, return to zero.
+        self.fl_warmth = math.max(100 - 50 * (22 - diff_time), 0)
+    end
+    self.fl_warmth = math.floor(self.fl_warmth + 0.5)
+    -- Make sure sysfs_light actually picks that new value up without an explicit setWarmth call...
+    -- This avoids having to bypass the ramp-up on resume w/ an explicit setWarmth call on devices where brightness & warmth
+    -- are linked (i.e., when there's no mixer) ;).
+    -- NOTE: A potentially saner solution would be to ditch the internal sysfs_light current_* values,
+    --       and just pass it a pointer to this powerd instance, so it has access to fl_warmth & hw_intensity.
+    --       It seems harmless enough for warmth, but brightness might be a little trickier because of the insanity
+    --       that is hw_intensity handling because we can't actually *read* the frontlight status...
+    --       (Technically, we could, on Mk. 7 devices, but we don't,
+    --       because this is already messy enough without piling on special cases.)
+    if self.fl then
+        self.fl.current_warmth = self.fl_warmth
+    end
+    -- Enable background job for setting Warmth, if not already done.
+    if not self.autowarmth_job_running then
+        table.insert(PluginShare.backgroundJobs, {
+                         when = 180,
+                         repeated = true,
+                         executable = function()
+                             if self.auto_warmth then
+                                 self:setWarmth()
+                             end
+                         end,
+        })
+        self.autowarmth_job_running = true
+    end
+end
+
 
 return AndroidPowerD
--- a/frontend/ui/widget/frontlightwidget.lua
+++ b/frontend/ui/widget/frontlightwidget.lua
@@ -65,7 +65,7 @@ function FrontLightWidget:init()
         self.nl_min = self.powerd.fl_warmth_min
         self.nl_max = self.powerd.fl_warmth_max
         -- NOTE: fl_warmth is always [0...100] even when internal scale is [0...10]
-        self.nl_scale = (100 / self.nl_max)
+        self.nl_scale = (self.powerd.fl_max / self.nl_max)
     end
 
     -- button width to fit screen size
--- a/reader.lua
+++ b/reader.lua
@@ -201,6 +201,53 @@ if G_reader_settings:isTrue("color_rendering") and not Device:hasColorScreen() t
     })
 end
 
+-- Inform that write permissions for NaturalLight are not set on Tolino Epos2
+-- also show a short manual for activating it.
+if Device.model == "Android_with_natural_lights" and not Device.hasNaturalLight() then
+    if not G_reader_settings:has("color_write_permission") then
+       G_reader_settings:saveSetting("color_write_permission", 1)
+    end
+    if G_reader_settings:readSetting("color_write_permission") == 1 then
+      local MultiConfirmBox = require("/ui/widget/multiconfirmbox")
+      UIManager:show(MultiConfirmBox:new{
+          text = _("You are using a Tolino Epos2!\nKOReader needs permissions to set the backlight warmth.\n" ..
+            "\nDo you want instructions to activate NaturalLight?"),
+          cancel_text = _("Now"),
+          cancel_callback = function()
+              G_reader_settings:saveSetting("color_write_permission", 2) -- show again in 2-1=1 starts
+              local InfoMessage = require("ui/widget/infomessage")
+              UIManager:show(InfoMessage:new{
+              text = _("Connect the rooted device per USB\n" ..
+                       "Execute the following commands:\n\n" ..
+                       "adb shell\n" ..
+                       "su\n" ..
+                       "mount -o remount,rw /system\n" ..
+                       "PERM=\"/sys/class/backlight/tlc5947_bl/color\"\n" ..
+                       "FILE=\"/system/bin/upgrade_check.sh\"\n" ..
+                       "echo \"chmod 666 $PERM\" >> $FILE\n" ..
+                       "mount -o remount,ro /sytem\n" ..
+                       "\nRestart the System"
+                       ),
+              })
+          end,
+          choice1_text = _("Later"),
+          choice1_callback = function()
+              G_reader_settings:saveSetting("color_write_permission", 6) -- show again in 6-1=5 starts
+          end,
+          choice2_text = _("Never"),
+          choice2_callback = function()
+              G_reader_settings:saveSetting("color_write_permission", 0) -- never show again
+          end,
+        })
+    else
+        local value = G_reader_settings:readSetting("color_write_permission")
+        if value > 0 then
+            value = value - 1
+        end
+        G_reader_settings:saveSetting("color_write_permission", value)
+    end
+end
+
 -- Helpers
 local lfs = require("libs/libkoreader-lfs")
 local function retryLastFile()


Frenzie:
Btw, did you see https://github.com/koreader/koreader/pull/5331?

> I do not know how to grant that permission within the app. Can anyone help with this?

I don't think that would be possible with non-root apps.

pazos:
@zwim: I would suggest calling the system frontlight dialog direclty (like we do with the wifi toggle). Or alternatively implement a generic one for warmth devices in https://github.com/koreader/android-luajit-launcher/

> Do a better check if the device is really a Tolino Epos 2, not just check android.prop.product

Are you aware of https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/device/DeviceInfo.kt. ?

It is used to determine if a device has a eink screen and other workarounds (well, the only workaround is enabling wakelocks on a sony model). It could be used to detect warmth devices.

zwim:
> @zwim: I would suggest calling the system frontlight dialog directly (like we do with the wifi toggle). Or alternatively implement a generic one for warmth devices in https://github.com/koreader/android-luajit-launcher/

Unfortunately the system frontlight dialog only shows the brightness. The warmth can only be set in the original Tolino reader apk.

Does using a generic dialog for warmth solve the problem with the permissions?

> 
> > Do a better check if the device is really a Tolino Epos 2, not just check android.prop.product
> 
> Are you aware of https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/device/DeviceInfo.kt. ?

I have compared some tolino getprops. It seems that "ntx_6sl" is unique under the Tolinos. Maybe I do an additional `ro.product.hardwareType` check. That should do.

pazos:
> Unfortunately the system frontlight dialog only shows the brightness. The warmth can only be set in the original Tolino reader apk.

> Does using a generic dialog for warmth solve the problem with the permissions?

I would expect the app is using `setprops` internally. You can try to RE a bit of it. It should be pretty straightforward to figure out what is doing.

Reading  #5331 would give you some hints. All info is still valid today (except the hard vetto against WRITE_SETTINGS that no longer holds true, so you're free to use that if it does what you need).

> I have compared some tolino getprops. It seems that "ntx_6sl" is unique under the Tolinos. Maybe I do an additional ro.product.hardwareType check. That should do. 

Some properties are just in some vendor firmware. It is safe to "get" them if you catch exceptions. See how its done for `ro.board.platform` in https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/EPDTestActivity.kt#L43-L53

zwim:
Every Tolino Epos2 running KOReader has to be rooted. Otherwise KOReader would not have been installed. So the user can change the write permissions to the color file with `adb shell`. Instructions therefore will be given if necessary.

I have updated the code:

If running on a Tolino Epos2, it is checked if the color file is writeable. If yes we are fine.
If no, the user gets an information window where he can decide, if he wants instructions how to set  write permissions permanently. The decision can be delayed or switched off.
If the user follows the instruction, the information window does not show up again.

In any case the brightness dialog shows only sane entries.

That's the best I can do.

zwim:
> This is my first veto, I feel a little guilty
 
No problem, thats your job :-)

As my hard-disc died over the night I have to buy a new one and get my backups right now. So I am busy the next days. 
I will get back to you next week.

zwim:
@pazos I am back again. It was not the hard-disk it was my motherboard.

I really want the possibility to change the warmth of the backlight on Tolino Epos 2.

You wrote, the patch will break on normal devices. I don`t see that point. Please could you explain this to me.

I agree, that placing the instructions (spaghetti?) in reader.lua is not beautiful. It would be much better not to touch reader.lua and gain write access to `/sys/class/backlight/tlc5947_bl/color` directly. The only way I see is with `chmod 666`. Should this be done indevice.lua` by spawning a su shell? Or would it be better to reengineer the original Tolino app?

--------

I also do not understand your comment about epd updates. My Tolino works very well with KOReader.

pazos:
Hey @zwim, good to see here you again 😄

> You wrote, the patch will break on normal devices.

You changed fl scale and now 3/4 values are out of range.

> I also do not understand your comment about epd updates. My Tolino works very well with KOReader.

That's because it uses the proper calls. It involved some reverse engineering of the Tolino APK, made by @char11.  Another example is the bug 

By definition e-ink "android" devices are not android devices. There are no android APIs to perform an eink update or to change the warmth light. These things need to be RE'd. Other APIs are broken, like https://github.com/koreader/koreader/issues/5731 and need work too, but these are usually easier to solve with just a crash.log.

> I agree, that placing the instructions (spaghetti?) in reader.lua is not beautiful. It would be much better not to touch reader.lua and gain write access to /sys/class/backlight/tlc5947_bl/color directly. The only way I see is with chmod 666. Should this be done indevice.lua` by spawning a su shell? Or would it be better to reengineer the original Tolino app?

Try to figure out what the Tolino app is doing. Probably you can write the proper value using WRITE_SETTINGS. See https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/utils/SystemSettings.kt#L44. If that's not possible we might want to and the code to spawn a su shell in the same file and use that instead.

> I really want the possibility to change the warmth of the backlight on Tolino Epos 2.

I really want that too. Hopefully, if we implement it well, it can be used on other devices as well. That would be nice!

zwim:
I have tried to find out what the Tolino app is doing. The app is not using the android system dialog.

However logcat shows:
```I/ActivityManager( 2504): Displayed de.telekom.epub/.ui.activities.menu.settings.SettingsMenuActivity: +426ms
I/ActivityManager( 2504): START u0 {cmp=de.telekom.epub/.ui.activities.menu.settings.SettingsIlluminationActivity} from pid 2573
I/View    ( 2573): Waveform mandatory mode is now disabled.
D/wingAndyBsystem( 2504): setFrontLightState with: on:true and fromButton:false
D/PhoneStatusBarPolicy( 2553): received intent :android.intent.action.FL_CHANGED
D/PhoneStatusBarPolicy( 2553): update:E80K00
D/ControllerBrightness( 2573): Set new screen brightness to: 99
D/PhoneStatusBarPolicy( 2553): received intent :android.intent.action.FL_CHANGED
D/PhoneStatusBarPolicy( 2553): update:E80K00
D/wingAndyBsystem( 2504): setFrontLightState with: on:true and fromButton:false
I/ActivityManager( 2504): Displayed de.telekom.epub/.ui.activities.menu.settings.SettingsIlluminationActivity: +467ms
D/lights  ( 2504): ================ brightness_color = 54
D/lights  ( 2504): ================ brightness_color = 81
D/lights  ( 2504): ================ brightness_color = 24
D/lights  ( 2504): ================ brightness_color = 50
D/lights  ( 2504): ================ brightness_color = 68
D/lights  ( 2504): ================ brightness_color = 67
D/ElectronBeam( 2504): dismiss
D/ControllerBrightness( 2573): Set new screen brightness to: 127
D/FrontLight( 2504): Light====>SYS. brightness orig 127 conv 127
D/lights  ( 2504): ================ brightness_color = 43
D/PhoneWindowManager( 2504): mPstate.get() = 1
W/PhoneWindowManager( 2504): ====> sendPowerkeyLongpressBroadcast
```
Changing the brightness gives something like `D/FrontLight( 2504): Light====>SYS. brightness orig 127 conv 127` and changing the color shows something like ` D/lights  ( 2504): ================ brightness_color = 24`. Notice the difference between `SYS.brightness` and `brigthness_color`.

I would go for spawning a `su shell` and give the `color` file the right permission. What do you think?
You mentioned that the su shell should be called in SystemSettings.kt. Why not do that in android/device.lua? Similar to the Kobos, where the device specific stuff is also done in kobo/device.lua.

pazos:
> You mentioned that the su shell should be called in SystemSettings.kt. Why not do that in android/device.lua? Similar to the Kobos, where the device specific stuff is also done in kobo/device.lua.

Because every SDK method is called using JNI. And using JNI is unsafe and can't be protected using `pcall` from lua. Unfortunately we can't avoid the oddities of android. So you need to implement unsafe things in Kotlin and catch exceptions.

I think that you can:

- a: keep the changes to yourself and change things wildy without caring about the compatibility of other devices. You can do that thanks to your previous contributions (running shell scripts). If you want to distribute things for other you can do that on the wiki.

- b: reverse engennering `de.telekom.epub/.ui.activities.menu.settings.SettingsIlluminationActivity`.  Look how @carlinux did in https://github.com/koreader/koreader/issues/4595.

zwim:
OK, thank you. 
I will close this PR and keep the changes.

pazos:
> OK, thank you.
> I will close this PR and keep the changes.

Fair enough!

Feel free to ping back if you change your idea. I think it can be solved in 30' hanging on Gitter

elvvis:
Too bad, I would also like to have this feature on my epos 2...

zwim:
@pazos as elvvis has is interested in this feature, too. I want to start another try.
Is it OK, when I put my code together and start a new PR tomorrow morning? As a base for discussions.
On Thursday I would have time do discuss and solve problems.

elvvis:
That's good news, I'm happy.
@zwim: Falls ich Dich bei Tests unterstützen kann, lass es mich wissen. Bin selbst leider noch nicht dazu gekommen, mich in die Sourcen einzuarbeiten :-(

zwim:
@elvvis: That's good. Let us stay with English so that pazos can follow us.
Today I will set up all my code together and test it on my Phone and on the Epos2, so I can exclude negative impacts of my modifications on "non Epos" Android.
Tomorrow (morning) I do a PR. 

Let's see what pazos thinks about it.

pazos:
> As elvvis has is interested in this feature, too. I want to start another try.

That's cool. Start by decompiling `de.telekom.epub/.ui.activities.menu.settings.SettingsIlluminationActivity` and see what happens each time you move the slider for both normal and warmth lights.

zwim:
Hello pazos. I am not familiar with decompiling android stuff. Could you please give me some hints how to start. apktool ...?

elvvis:
@zwim: Great, I'm glad you're staying on

pazos:
Indeed, use apktool to decompile and http://java-decompiler.github.io/ to see the results. Then navigate to the class that's in charge of updating the values.

You said that the Tolino app is the one that has the controls. Copy that apk to your computer, use `apktool d nameOfApp.apk` and open the results in jd-gui. I'm aware you can do that on your device (like in https://github.com/koreader/koreader/issues/4595#issuecomment-465972270). That doesn't matter. The important stuff is what you'll see there. As I said I would expect something like `android.provider.Settings.System.putInt` is used to set the new value system-wide. Maybe is a bit more complicated than that. Lets see.

zwim:
OK, I have apktool already, but I have to install the java-decompiler and play with it. I will be back tomorrow.

zwim:
apktool was not the right. I used `d2j-dex2jar.sh`. There are not much `putInt` in the apk. 

Anyway I have found `CTMUtil.class` where there is a suspect:
```
public static void setPrefCTMColor(ContentResolver paramContentResolver, int paramInt) {
    Settings.System.putInt(paramContentResolver, "screen_brightness_color", paramInt);
  }
```
Does this seem right? What is the next step?

pazos:
> apktool was not the right. I used d2j-dex2jar.sh.

Of course. Sorry about the missguide :facepalm: 

> Does this seem right?

Absolutely right. That seems the interface for warmth light. Now you need to find another one for your white light. As I understand from your original PR both white/color lights go from 0 to 100?

> What is the next step?

Run KOReader, go to filemanager -> tools -> more tools -> developer options -> start e-ink test. Anotate device id strings and paste them here. We need to use them to flag the Epos2 as a warmthDevice.

There're a few steps to integrate that on KOReader but testing if changes work should be pretty easy with just these two things. Again:

- a interface for changing white led intensity (hopefully similar of the one you pasted here)
- device id.

zwim:
I found two occurences of putInt for screen brighness
This seems to be for the brightness
```
public void setBrightnessInPreferences(int paramInt) {
    paramInt += 4;
    if (this.activity == null) {
      Log.w(LOG_TAG, "onStopTrackingTouch called without activity, after cleanup, this should never happen, possibly a threading issue, TODO: investigate. Handling by ignoring the call");
      return;
    } 
    Log.d(LOG_TAG, "Set new screen brightness to: " + paramInt);
    Settings.System.putInt(this.activity.getContentResolver(), "screen_brightness", paramInt);
  }
```
and this tor manual mode 
```
   private void setBrightnessModeToManual() {
    Settings.System.putInt(getContentResolver(), "screen_brightness_mode", 0);
  }
```
The brightness goes from 0..100, the color from 0..10 (zero means warm, 10 is cold).

--------------
eInk Test shows:
Manufacturer: Rakuten Kobo Inc.
Brand: RakutenKobo
Model: tolino
Product: ntx_6sl
Hardware: E80K00
Platform: imx6

I hope this helps.

pazos:
Surely. That's all the info we need. Now need some free time to think how to implement the interface. I'll ping you when I have something to test.

pazos:
@zwim: a test build is in https://www.dropbox.com/s/gl3kwf841uvnx0y/koreader-arm-fl-demo.apk?dl=0. If you want to test it you'll need to remove KOReader apk first as it is signed with a different key.

On your phone it should work as always. On the FM you'll have a new option called "NEW: show native confirmbox". Clicking on it should open an android dialog with controls for screen brightness (called intensity).

On your EPOS2 if you do the same it should open an android dialog with controls for both screen brightness and warmth. Warmth controls are reversed.

Anyways, I can't be sure that changing bright/warmth works on your device but it shouldn't crash. Also not sure if that's the case on your device but on normal ones you should grant KOReader write settings permissions before trying to change bright/warmth. (You can do that on settings gear -> screen -> timeout -> allow overrides. If you don't see that entry that means that WRITE_SETTINGS is already enabled for the app.

zwim:
> On your phone it should work as always. On the FM you'll have a new option called "NEW: show native confirmbox". Clicking on it should open an android dialog with controls for screen brightness (called intensity).

> 
> On your EPOS2 if you do the same it should open an android dialog with controls for both screen brightness and warmth. Warmth controls are reversed.


The dialog shows up, I can slide both brightness and warmth. Brightness works, but color NOT.

> 
> Anyways, I can't be sure that changing bright/warmth works on your device but it shouldn't crash. Also not sure if that's the case on your device but on normal ones you should grant KOReader write settings permissions before trying to change bright/warmth. (You can do that on settings gear -> screen -> timeout -> allow overrides. If you don't see that entry that means that WRITE_SETTINGS is already enabled for the app.

I don`t see that entry.

pazos:
Thanks for the feedback.

> The dialog shows up, I can slide both brightness and warmth. Brightness works, but color NOT.

 Sounds weird. Both are using the same interface. You can get specific logs using `adb logcat lights:* *:S`

> I don`t see that entry.

Cool. One thing less to worry about.

If you want to experiment a bit you can build your own APK on top of the android-luajit-launcher branch mentioned above. The only change needed on the frontend is:

```diff
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -201,6 +201,15 @@ if Device:isAndroid() then
     local isAndroid, android = pcall(require, "android")
     if not isAndroid then return end
 
+    -- TEST FRONTLIGHT NATIVE DIALOG BEGINS
+    common_settings.native_confirmbox = {
+        text = _("show native confirmbox"),
+        callback = function()
+            android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
+        end,
+    }
+    -- TEST FRONTLIGHT NATIVE DIALOG ENDS
+
     -- screen timeout options, disabled if device needs wakelocks.
     common_settings.screen_timeout = require("ui/elements/screen_android"):getTimeoutMenuTable()
```

You can also grep on jd-gui for a string "screen_brightness_color" or for methods in `CTMUtil.class`.  I'm afraid I cannot help more without the device and/or the dump of the Tolino apk.

zwim:
> Sounds weird. Both are using the same interface. You can get specific logs using `adb logcat lights:* *:S`

logs are only shown in the tolino app, not in koreader:
```
 adb logcat lights:* *:S
--------- beginning of /dev/log/system
--------- beginning of /dev/log/main
D/lights  ( 2487): ================ brightness_color = 4
D/lights  ( 2487): ================ brightness_color = 97
D/lights  ( 2487): ================ brightness_color = 1
D/lights  ( 2487): ================ brightness_color = 2
D/lights  ( 2487): ================ brightness_color = 5
I/lights  ( 2487): [set_light_leds_battery 224]RGB = (0, ff ,0)
I/lights  ( 2487): battery led on
E/lights  ( 2487): _PowerLEDon
D/lights  ( 2487): ================ brightness_color = 8
D/lights  ( 2487): ================ brightness_color = 70
D/lights  ( 2487): ================ brightness_color = 27
I/lights  ( 2487): [set_light_leds_battery 224]RGB = (0, 0 ,0)
I/lights  ( 2487): battery led off
E/lights  ( 2487): _PowerLEDoff
D/lights  ( 2487): ================ brightness_color = 8
I/lights  ( 2487): [set_light_leds_battery 224]RGB = (ff, ff ,0)
D/lights  ( 2487): ================ brightness_color = 9
D/lights  ( 2487): ================ brightness_color = 24
D/lights  ( 2487): ================ brightness_color = 45
D/lights  ( 2487): ================ brightness_color = 62
D/lights  ( 2487): ================ brightness_color = 71
D/lights  ( 2487): ================ brightness_color = 78
```
> If you want to experiment a bit you can build your own APK on top of the android-luajit-launcher branch mentioned above. 

Tomorrow.

You can find the Tolino-App in the update: https://download.pageplace.de/ereader/13.2.1/alldevices/update.zip (this is the official update link) in `/system/app/EPubProd.apk`

pazos:
Try using reflection:

```diff
--- a/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
+++ b/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
@@ -1,6 +1,7 @@
 package org.koreader.launcher.device.lights
 
 import android.app.Activity
+import android.content.Context
 import android.provider.Settings
 
 import org.koreader.launcher.interfaces.LightInterface
@@ -50,11 +51,21 @@ class TolinoWarmthController : LightInterface {
 
     override fun setWarmth(activity: Activity, warmth: Int) {
         if (warmth < MIN || warmth > WARMTH_MAX) return
+
+        /*
         try {
             Settings.System.putInt(activity.applicationContext.contentResolver,
                 "screen_brightness_color", warmth)
         } catch (e: Exception) {
             Logger.w(TAG, "$e")
+        }*/
+
+        try {
+            Class.forName("android.os.PowerManager").getMethod("setFrontlightBrightnessColor",
+                Integer.TYPE).invoke(activity.applicationContext.getSystemService(Context.POWER_SERVICE),
+                warmth)
+        } catch (e: Exception) {
+            Logger.e(TAG, e.toString())
         }
     }
```

And yeah, my bad. Is not lights but `adb logcat KOReader:* *:S | grep lights`

zwim:
I have included luajit-launcher#239, changed common_settings_menu_table.lua and added the reflection.

When touching the Warmth slider `adb logcat KOReader:* *:S | grep lights` shows

`E/KOReader( 3396): [lights] java.lang.reflect.InvocationTargetException`

pazos:
in invoke use `null` as first parameter instead of a power_service instance. Please do a full log and attach it here.

zwim:
With 
```
Class.forName("android.os.PowerManager").getMethod("setFrontlightBrightnessColor",
Integer.TYPE).invoke(null,  warmth)
```
I get the following log. Starting from uninstalling the old koreader, installing koreader, starting koreader, going to "show native confirmbox", tapping on warmth and intensity.
[invoke-null.log](https://github.com/koreader/koreader/files/4931095/invoke-null.log)

pazos:
Good to see that I was not totally wrong :p.

`java.lang.NullPointerException: expected receiver of type android.os.PowerManager, but got null`

I think it has something to do with `activity.applicationContext.getSystemService(Context.POWER_SERVICE)` being a PowerManager?. I think you can cast it to a PowerManager as pass a instance instead of null.

```kotlin
val powerManager = activity.applicationContext.getSystemService(Context.POWER_SERVICE) as PowerManager
Class.forName("android.os.PowerManager").getMethod("setFrontlightBrightnessColor",
    Integer.TYPE).invoke(powerManager, warmth)
```

zwim:
Did the change and added `import android.os.PowerManager`, warmth is not working.
First a few taps to warmth, then intensity (works), then taps on warmth again.

logcat shows:
```I/KOReader( 4069):  RD loaded plugin profiles at plugins/profiles.koplugin
I/KOReader( 4069):  RD loaded plugin readtimer at plugins/readtimer.koplugin
I/KOReader( 4069):  RD loaded plugin send2ebook at plugins/send2ebook.koplugin
I/KOReader( 4069):  RD loaded plugin statistics at plugins/statistics.koplugin
I/KOReader( 4069):  RD loaded plugin systemstat at plugins/systemstat.koplugin
I/KOReader( 4069):  RD loaded plugin terminal at plugins/terminal.koplugin
I/KOReader( 4069):  RD loaded plugin texteditor at plugins/texteditor.koplugin
I/KOReader( 4069):  RD loaded plugin wallabag at plugins/wallabag.koplugin
I/KOReader( 4069):  RD loaded plugin zsync at plugins/zsync.koplugin
V/KOReader( 4069): [MainActivity] requesting epd update, mode:7, delay:0, [x:0, y:0, w:1440, h:1920]
I/epd     ( 4069): requested eink refresh, type: 7 x:0 y:0 w:1440 h:1920
V/KOReader( 4069): [MainActivity] requesting epd update, mode:39, delay:0, [x:0, y:0, w:1440, h:1096]
I/epd     ( 4069): requested eink refresh, type: 39 x:0 y:0 w:1440 h:1096
V/KOReader( 4069): [MainActivity] requesting epd update, mode:39, delay:0, [x:0, y:0, w:1440, h:1920]
I/epd     ( 4069): requested eink refresh, type: 39 x:0 y:0 w:1440 h:1920
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
W/BackupManagerService( 2487): dataChanged but no participant pkg='com.android.providers.settings' uid=10049
D/ElectronBeam( 2487): dismiss
D/FrontLight( 2487): Light====>SYS. brightness orig 63 conv 63
W/BackupManagerService( 2487): dataChanged but no participant pkg='com.android.providers.settings' uid=10049
D/ElectronBeam( 2487): dismiss
D/FrontLight( 2487): Light====>SYS. brightness orig 18 conv 18
W/BackupManagerService( 2487): dataChanged but no participant pkg='com.android.providers.settings' uid=10049
D/ElectronBeam( 2487): dismiss
D/FrontLight( 2487): Light====>SYS. brightness orig 100 conv 100
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
E/KOReader( 4069): [lights] java.lang.reflect.InvocationTargetException
W/InputMethodManagerService( 2487): Window already focused, ignoring focus gain of: com.android.internal.view.IInputMethodClient$Stub$Proxy@4224fc50 attribute=null, token = android.os.BinderProxy@421bc1e0
```

pazos:
According to https://stackoverflow.com/a/27055928 trying to invoke powerManager @hide methods requires android.permission.DEVICE_POWER (only available for system apps).

I don't know if `setFrontlightBrightnessColor` is @hide. We need the Tolino's AOSP sources for that, since it is not part of android APIs.

But I really don't know. It bugs me a lot that the initial implementation didn't work. It really should, specially if white light control works.Anyways you can print a stacktrace on the catch block adding `e.printStackTrace()`.

zwim:
e.printStackTrace() does not show anything. Shouldn`t it go to the log?

pazos:
Yup `Caused by: java.lang.SecurityException: Neither user 10051 nor current process has android.permission.DEVICE_POWER`

What a dumb device!. So the TolinoWarmthController won't work unless the app has the relevant permission and is installed as a system app. Even worst: probably needs the same signature as other system apps. Kind of a dead end.

Which leaves us the other option: write to /sys/ directly. I would still want to do that inside TolinoWarmthController to repurpose the native controls. So it is a matter of changing `getWarmth` and `setWarmth` to read/write the specific bits on the relevant file.

zwim:
I changed getWarmth and setWarmth methods:
```
   override fun getWarmth(activity: Activity): Int {
        val colorFile = File(COLOR_FILE)    
        return try {
            WARMTH_MAX-colorFile.readText().toInt()
        } catch ( e: Exception ) {
            Logger.w(TAG, "$e")
            0
        }
    }
```
and
```
    override fun setWarmth(activity: Activity, warmth: Int) {
        if (warmth < MIN || warmth > WARMTH_MAX) return

        val colorFile = File(COLOR_FILE)    
        try {    
            colorFile.setLastModified(System.currentTimeMillis())
        } catch( e: Exception ) {
            Runtime.getRuntime().exec("su -c chmod 666 "+COLOR_FILE)
        }
        
        try {
            colorFile.writeText((WARMTH_MAX-warmth).toString())
        } catch ( e: Exception ) {
            Logger.w(TAG, "$e")
            0
        }
    }
```
With this changes the native controlls work. I have changed `warmth` to `WARMTH_MAX-warmth` so that the slider increases the warmth.
On the first start (after a reboot) you get the dialog of `SuperSu` to grant write access. On consecutive starts this dialog is not shown.

pazos:
@zwim: thanks for the patch. I've updated the PR against luajit-launcher. 

Now is just a matter of using the native dialog instead of the frontlight widget and adjust the swipe gestures to trigger relevant events. Probably we need to add a few wrapper functions to get min/max values from lua and to decide if the device hasNaturalLight.

zwim:
When I add

```
------------- frontend/ui/elements/common_settings_menu_table.lua -------------
index 4abaeec9..037c92ed 100644
@@ -201,6 +201,14 @@ if Device:isAndroid() then
     if not isAndroid then return end
 
+    -- on Android: overwrite generic frontlight with a native Dialog 
+    common_settings.frontlight = {
+        text = _("Frontlight"),
+            android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
+            end,
+    }
+
     -- screen timeout options, disabled if device needs wakelocks.
     common_settings.screen_timeout = require("ui/elements/screen_android"):getTimeoutMenuTable()
```

I get the native android dialog. It is working on my mobile and on the Epos2. The warmth is only shown on the Epos2.

Touch gestures for in/decreasing intensity works on the Epos2.

BUT on the phone: With the touch gestures I get the popups "Frontlight disabled"/"Frontligh enabled". (Even with the original dialog). No change in intensity.

pazos:
@zwim: there are a few things that need to be done on the frontend. Namely:

in https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua: 

- set `hasNaturalLight` to  `android.isWarmthDevice`

in  https://github.com/koreader/koreader/blob/master/frontend/device/android/powerd.lua: 

- add setWarmth and other things based on exposed backend functions.

Ie: what needs to be done is to interface with the backend. Probably some stuff is missing in the backend. In that case it is better to document what's missing and implement it there.

pazos:
By exposed backend functions I'm talking about https://github.com/koreader/android-luajit-launcher/pull/239/files#diff-6c2e53facd9a74b879f24ca79562f214

pazos:
Probably missing are a way to retrieve the max/min brightness/warmth for each implementation of the lights interface.

zwim:
Changing intensity and warmth work now on the Epos2 and the phone. 
But I would definitely need a method to get maxWarmth and minWarmth from the backend.

zwim:
`getBrightness(activity: Activity)` from GenericController.kt delivers `-1.0` on start and until the intensity is set the first time.
It would be desirable to set the initial `fl_intensity` in android/device.lua with the intensity on startup, so that consecutive changes start from the initial intensity.

On the TolinoWarmthController.kt this is no problem.

pazos:
@zwim: that's correct. The difference between in activity implementation vs using system settings is that the first has a fallback value (managed by the system) and the second hasn't.

Your suggestion is already implemented in https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L247-L251. But I won't want to apply it unconditionally. For most users the app is just another app and should use the system wide fl value.

Anyways I feel I miss something (beyond adding a couple of lua functions to assets/android.lua). Could you please update this PR or send a new one, so I can figure out how to help?.

zwim:
@pazos
> Your suggestion is already implemented in https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L247-L251. But I won't want to apply it unconditionally. For most users the app is just another app and should use the system wide fl value.

That was one of my problems, as I updated from an older KOReader to the actual one. `fl_last_value` was set in `settings.reader.lua`. We should make sure, that this is not the case in an official app. I have commented out the lines247-251 for now.
> 
> Anyways I feel I miss something (beyond adding a couple of lua functions to assets/android.lua). Could you please update this PR or send a new one, so I can figure out how to help?.

is comming in a few minutes

zwim:
@pazos: The new PR is here https://github.com/koreader/koreader/pull/6426

PR #6333, opened 2020-06-30
[i18n] Add support for fledgling Hebrew translation

Frenzie:
See <https://hosted.weblate.org/languages/he/koreader/> if you'd like to contribute.

 

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6333)


Frenzie:
It's easier to read if you break it down the same way the code I wrote a while back does. (Un?)fortunately none of the languages I had at my disposal at the time used nested parentheses. But as far as the outcome of the formula goes, they don't seem necessary in the first place.

* `n == 1 ? 0` (string 0 if the number is one)
* `: (n == 2) ?  1` (string 1 if the number is two)
* `: (n > 10 && n % 10 == 0) ? 2` (string 2 if it's larger than 10 and divides nicely by ten)
* ` : 3` (otherwise string 3)

I don't know atm if any relevant strings have been translated into Hebrew already, but for example in Romanian:
Screenshot_2020-06-30_17-28-08

PR #6347, opened 2020-07-04
Allow locking the gyro to the current screen mode

### LINKED ISSUE #5106: Gyro enhancement

NiLuJe:
Fix #5106 

Also a few minor cleanups following #6309

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6347)


NiLuJe:
Apparently useful for people who like switching hands on those asymmetric devices ;).

(On the plus side, it allows us to use a much faster rotation mechanism, because there's no layout change necessary).

NiLuJe:
Not super-happy about burying a runtime flag in a Device method, but for some reason I was convinced that ReaderSettings was a no-go in input?

(Although the Android Camera Key stuff does use it, so, eh, :D).

poire-z:
> Not super-happy about burying a runtime flag in a Device method, but for some reason I was convinced that ReaderSettings was a no-go in input?

I have no issue with that. (you might be confusing input with the CanvasContext, that houqp wanted to be frontend agnostic, so no setting?)

frontend/device/generic/device.lua
```diff
@@ -332,6 +341,28 @@ function Device:setIgnoreInput(enable) return true end
...
+-- Whether or not the GSensor should be locked to the current orientation (i.e. Portrait <-> Inverted Portrait or Landscape <-> Inverted Landscape only)
+function Device:lockGSensor(toggle)
+    if not self:hasGSensor() then
+        return
+    end
+
+    if toggle and toggle == true then
+        -- Lock GSensor to current roientation
+        self.isGSensorLocked = yes
+    elseif toggle and toggle == false then
```
poire-z, line 353: Looks like this will never trigger :)
Just `if toggle == true` / `elseif toggle == false` and a `else` (for toggle == nil or anything else) should do.

frontend/ui/elements/screen_rotation_menu_table.lua
```diff
@@ -24,9 +25,29 @@ return {
             })
         end
 
+        if Device:hasGSensor() then
+            table.insert(rotation_table, {
+                text = _("Lock auto rotation to current orientation"),
+                help_text = _([[When checked, the gyro will only be honored when switching between the two inverse variants of your current rotation,
+i.e., Portrait <-> Inverted Portrait OR Landscape <-> Inverted Landscape.
+Switching between (Inverted) Portrait and (Inverted) Landscape will be inhibited.
```
poire-z, line 33: You can put the first line at the start of a new line (so the whole text is left aligned and the code is easier to read).
Lua strips out the firt `\n`.

frontend/device/generic/device.lua
NiLuJe: Oh, right. I was somehow convinced that `nil` or unset evaluated as `false`, too.

NiLuJe:
@pazos: Gyro rotations are as transient as a rotation via the UI.

The only thing that should stick is if one sets a default FM rotation ;).

So, if you lock in Landscape without having (Inverted) Landscape set as your default FM rota, you'll effectively be locked in Portrait after a KOReader restart.

NiLuJe:
Which made me realize than setting a default FM rota doesn't actually stick on my end ^^.

NiLuJe:
Ah, that only happens if "Keep FM rotation" is enabled.

frontend/ui/elements/screen_rotation_menu_table.lua
```diff
@@ -81,7 +81,7 @@ When checked, in addition to the obvious, setting a default rotation below will
                     if G_reader_settings:nilOrFalse("lock_rotation") then
                         UIManager:show(InfoMessage:new{
                             text = _("Can't set a default FM rotation when 'Keep file browser rotation' is disabled!"),
-                            timeout = 2,
+                            timeout = 4,
```
poire-z, line 84: I would even remove the timeout: it's an InfoMessage, it can be dismissed - and I might need more than 4s to process it and understand what I should do :)
Also, the `text = _("Can't set ...")` string looks like it's the same duplicated 4 times: might be better put in a local variable above, so we never risk to update one and forget the 3 others.

yparitcher:
> Which made me realize than setting a default FM rota doesn't actually stick on my end ^^.
> Ah, that only happens if "Keep FM rotation" is enabled.

Fixed by #6348

frontend/ui/elements/screen_rotation_menu_table.lua
```diff
@@ -24,9 +26,30 @@ return {
...
+                    G_reader_settings:flipNilOrFalse("input_lock_gsensor")
+                    Device:lockGSensor(G_reader_settings:isTrue("input_lock_gsensor"))
+                end,
+            })
+        end
+
         table.insert(rotation_table, {
             text = _("Keep file browser rotation"),
-            help_text = _("When checked the rotation of the file browser and the reader will not affect each other"),
+            help_text = _("When unchecked, the rotation of the file browser and the reader will not affect each other"),
```
yparitcher, line 52: It is correct the way it is, it should say checked

NiLuJe: Ah, that's indeed a problem, because that menu label is now nonsensical ^^.

NiLuJe:
Note to self: double-check what happens when exiting a rotated document to the FM post #6347 & #6348

frontend/apps/reader/modules/readerview.lua
```diff
@@ -724,13 +724,21 @@ end
...
-    if not rotation_mode and locked then
-        if self.ui.document.info.has_pages then
-            rotation_mode = G_reader_settings:readSetting("kopt_rotation_mode") or 0
-        else
-            rotation_mode = G_reader_settings:readSetting("copt_rotation_mode") or 0
+    local rotation_mode = nil
+    local locked = G_reader_settings:isTrue("lock_rotation")
+    if locked then
+        -- Keep FM rotation
+        rotation_mode = G_reader_settings:readSetting("fm_rotation_mode") or Screen.ORIENTATION_PORTRAIT
```
yparitcher, line 731: This will reset the rotation to the default FM rotation not the current FM rotation as is desired.

it should be:
``` lua
local locked = G_reader_settings:isTrue("lock_rotation")
if  not locked then
        -- Honor docsettings's rotation
        rotation_mode = config:readSetting("rotation_mode") -- Doc's
        if not rotation_mode then
```

The current rotation will be kept by doing nothing

NiLuJe: Oh, yeah, I was approaching that slightly differently, but that actually makes much more sense ;).

And points towards making `lock_rotation` actually do what you probably originally meant (as is, kepe the current rotation *both* ways when going FM <-> reader).

NiLuJe:
Okay, rebased, and tweaked as per @yparitcher's recommendations, this should finally makes sense now ;p.

I suppose *that's* what `locked_rotation` was supposed to be about (hence the settings name), but for some reason it took some doing for me to grok that, sorry ^^.

NiLuJe:
@yparitcher: Yep. And, most importantly, the defaults (i.e., !locked) are left unchanged from pre #6309 / #6347 / #6348 ;).

BuzzKnightbeer:
This seems to be long closed, but I can't figure out where to activate the setting that only both portrait orientations work.

PR #6348, opened 2020-07-05
allow default FM rotation on init when `Keep FM rotation` is disabled

yparitcher:
fixes issue discovered in #6347

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6348)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -72,7 +72,7 @@ end
 
 function FileManager:setRotationMode(init)
     local locked = G_reader_settings:readSetting("lock_rotation")
-    local rotation_mode = G_reader_settings:readSetting("fm_rotation_mode") or 0
+    local rotation_mode = G_reader_settings:readSetting("fm_rotation_mode") or Screen.ORIENTATION_PORTRAIT
     if locked or init then
         self:onSetRotationMode(rotation_mode)
     end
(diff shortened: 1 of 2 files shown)


reader.lua
```diff
@@ -277,7 +277,7 @@ if ARGV[argidx] and ARGV[argidx] ~= "" then
         local home_dir =
             G_reader_settings:readSetting("home_dir") or ARGV[argidx]
         UIManager:nextTick(function()
-            FileManager:showFiles(home_dir)
+            FileManager:showFiles(home_dir, "", true)
```
NiLuJe, line 280: `nil` instead of `""`, or does something wonky happen?

frontend/apps/filemanager/filemanager.lua
```diff
@@ -70,10 +70,10 @@ function FileManager:onSetRotationMode(rotation)
     return true
 end
 
-function FileManager:setRotationMode()
+function FileManager:setRotationMode(init)
     local locked = G_reader_settings:readSetting("lock_rotation")
     local rotation_mode = G_reader_settings:readSetting("fm_rotation_mode") or 0
```
NiLuJe, line 75: Use the constant instead of 0, so I can get rid of that chunk in #6347 too ;p.

yparitcher: done

reader.lua
yparitcher: > `nil` instead of `""`, or does something wonky happen?

Didn't check but assumed so.
Not relevant anyways now

Issue #6357, opened 2020-07-07
Emulator: link cache folder error


yparitcher:
* Device: Emulator

If i delete the cache by pressing `Developer options > Clear cache/ and cr3cache/ ?`
then try to run the emulator via `./kodev run`
I get an error
```
[*] create symlink instead of copying files in development mode
cd koreader-emulator-x86_64-pc-linux-gnu-debug/koreader && \
	ln -sf ../../base/build/x86_64-pc-linux-gnu-debug/* .
ln: ./cache: cannot overwrite directory
make: *** [Makefile:77: all] Error 1
```

caused by:
https://github.com/koreader/koreader/blob/580b38e7754ce95ef98a404d030e280b18707beb/frontend/apps/filemanager/filemanagermenu.lua#L252-L255

the code remakes the cache directory as a folder not a link so when the makefile tries to link it fails.

https://github.com/koreader/koreader/blob/580b38e7754ce95ef98a404d030e280b18707beb/Makefile#L84

Frenzie:
Probably just remove the symlink from the Makefile. The code creates the cache directory if it's not there anyway.

yparitcher:
> Probably just remove the symlink from the Makefile. The code creates the cache directory if it's not there anyway.

the makefile is linking all the directories in the base build for use by koreader

Frenzie:
Tbh I've considered it a non-issue for half a decade because that which could be called a workaround (using my FM) is actually easier than bothering going all the way for that menu. ^_^

Anyhow, I think it'd be easier to exclude the cache (is something like `[!c]*` the appropriate syntax here?) but if you prefer you could also write a purgeDirContent or something, assuming it doesn't already exist.

pazos:
purgeDIr won't be called because cachedir is a link, not a directory. But lfs.mkdir is called without checking the attributes. Maybe something like:

```lua
local type = lfs.attributes(cachedir, "mode") 
 if type == "directory" then 
     purgeDir(cachedir)
     lfs.mkdir(cachedir) 
elseif type == "link" then
    -- remove contents of cachedir recursively
end
```

PR #6358, opened 2020-07-07
do not link cache dir as it may already exist

yparitcher:
closes koreader/koreader#6357

thanks @Frenzie

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6358)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ else
 	cp -f $(KOR_BASE)/ev_replay.py $(INSTALL_DIR)/koreader/
 	@echo "[*] create symlink instead of copying files in development mode"
 	cd $(INSTALL_DIR)/koreader && \
-		ln -sf ../../$(KOR_BASE)/$(OUTPUT_DIR)/* .
+		bash -O extglob -c "ln -sf ../../$(KOR_BASE)/$(OUTPUT_DIR)/!(cache) ."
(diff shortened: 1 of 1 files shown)


yparitcher:
it passes the tests locally, 
as usual i cant see the circleci logs so i cant know exactly what broke, but i wonder if it is caused by the ci cacheing the builds?

Frenzie:
@yparitcher `base` is cached so it doesn't have to do the full 15-minute build for every PR, but that's prior to running tests for basically exactly that reason.

yparitcher:
I figured it out.
[!cache] matches `c` or `a` or `h` or `e` as the first letter not `cache` as a word. there is no way is sh to not match a word and `[!c]*` as you suggested would also match clipboard & common :(

Frenzie:
@yparitcher I believe that in Bash you can do `!(cache)` although depending on Bash is potentially problematic. Pinging @NiLuJe who knows more about various shells.

NiLuJe:
We could *probably* enforce using Bash, but it's mildly annoying in Make (it involves pointing SHELL to the right place, IIRC?).

NiLuJe:
Barring that, moving away from shell globs to a filtered find might do the trick?

Frenzie:
Or just `bash -c` in that case? It's only one line.

yparitcher:
fixed

PR #6359, opened 2020-07-08
android: support for native surface rotation

pazos:
Native surface rotation is used on Kitkat+ devices. On devices with older versions and android Tv/ChromeOS blitbuffer rotation is still used.

Also bumps curl, fribidi and LuaJIT

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6359)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -129,6 +129,7 @@ function Device:init()
         device = self,
         event_map = require("device/android/event_map"),
         handleMiscEv = function(this, ev)
+            local UIManager = require("ui/uimanager")
             logger.dbg("Android application event", ev.code)
             if ev.code == C.APP_CMD_SAVE_STATE then
                 return "SaveState"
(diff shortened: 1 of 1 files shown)


frontend/device/android/device.lua
```diff
@@ -324,10 +336,16 @@ end
...
 function Device:info()
     local is_eink, eink_platform = android.isEink()
+    local product_type = android.getPlatformName()
 
     local common_text = T(_("%1\n\nOS: Android %2, api %3\nBuild flavor: %4\n"),
         android.prop.product, getCodename(), Device.firmware_rev, android.prop.flavor)
 
+    local platform_text = ""
+    if product_type ~= "android" then
+        platform_text = T(_("\nDevice type: %1\n"), product_type)
```
Frenzie, line 346: For translation it's friendlier to keep the newlines outside. (And I hardly think the theoretical overhead of string concat is worth worrying about. ;-) )

Issue #6331, opened 2020-06-29
FR: MacOS Desktop Application


JasonInOttawa:
Has any work been done, or is there any interest in, a MacOS desktop application (rather than building/running the emulator)?

NiLuJe:
Cocoa is pain. Swift is pain. macOS is pain.

But PRs welcome ;p.

JasonInOttawa:
So no work's been done, and there might be some interest :)

I haven't done any Mac development, but I do have a Macbook. I'll see if I can't figure something out.

NiLuJe:
If by "interest" you mean a negative amount of, sure?

Frenzie:
Unless someone's done that in secret, what you see is what you get. I don't know what goes into MacOS apps? If you're lucky it'd be a lot like the AppImage:

https://github.com/koreader/koreader/blob/86db43a0527e94f1ebe2a124dae92f8fd3f70e6e/Makefile#L313-L348

https://github.com/koreader/koreader-base/blob/849231bf518a6b1457459b408b9ef04e36f44b83/Makefile.defs#L441-L445

Incidentally, does the build process also randomly blow up on real Mac or is that just GitHub Actions? I added https://github.com/koreader/koreader-base/blob/849231bf518a6b1457459b408b9ef04e36f44b83/.github/workflows/build.yml to prevent accidentally breaking the Mac build and it worked great… for a few weeks. Ever since Mac clang has been segfaulting on MuPDF without us having changed anything, see [here](https://github.com/koreader/koreader-base/runs/820936534).

pazos:
> Has any work been done, or is there any interest in, a MacOS desktop application (rather than building/running the emulator)?

See https://github.com/koreader/koreader/issues/4375#issuecomment-444717730.

JasonInOttawa:
Thank you everyone. If the procedure for building a Mac application bundle the ugly way works, would it be acceptable to add it as a build target

NiLuJe:
Note that that'd still be building the emulator (i.e., SDL2). I got the impression you were after a truly native backend, which is a whole other can of worms.

JasonInOttawa:
Thank you again. I use the emulator to read on a laptop when I haven't got the Kobo. I was hoping for an easier upgrade method and tidier setup than rebuilding and running the emulator. 

Is SDL sufficient, or would only going native be acceptable?

Frenzie:
SDL is fine.

pazos:
> SDL is fine.

[...]

PR #6360, opened 2020-07-08
add mac ttf paths & simplify user fonts using Device.home_dir

pazos:
The basic abstraction to be used in #6331

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6360)

### DIFF
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -72,6 +72,7 @@ end
 local Device = Generic:new{
     model = "SDL",
     isSDL = yes,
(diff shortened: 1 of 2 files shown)


NiLuJe:
Isn't there a user-specific one in ${HOME}/Library/Fonts (or... something, can't quite recall what $HOME resolves to) too?

pazos:
Thanks for the review. Indeed there are a few things to improve. Now each device has `Device.home_dir` so there's no need to get home dir again in fonts. Will submit later tonight!

Issue #6351, opened 2020-07-07
MacOS Build fails on crengine


JasonInOttawa:
* KOReader version: current
* Device: emulator on MacOS

#### Issue

Build fails with "implicit instantiation of undefined template" errors while building lvmemman.cpp.o

#### Steps to reproduce

Set up build environment per instructions:
jasonbenwell@Jasons-MacBook koreader % echo $PATH
/usr/local/bin:/usr/local/sbin:/usr/local/opt/gettext/bin:/usr/bin:/bin:/usr/sbin:/sbin
jasonbenwell@Jasons-MacBook koreader % echo $MACOSX_DEPLOYMENT_TARGET
10.14

then run

git clone https://github.com/koreader/koreader.git
cd koreader && ./kodev fetch-thirdparty
kodev build

Results in this error:

[ 57%] Building CXX object CMakeFiles/crengine.dir/crengine/crengine/src/lvmemman.cpp.o
/Users/jasonbenwell/Documents/koreader/base/thirdparty/kpvcrlib/crengine/crengine/src/lvmemman.cpp:77:26: error: 
      implicit instantiation of undefined template 'std::__1::basic_string<char,
      std::__1::char_traits<char>, std::__1::allocator<char> >'
    std::string errstr = std::string("FATAL ERROR #") + std::string(strb...
                         ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/iosfwd:209:32: note: 
      template is declared here
    class _LIBCPP_TEMPLATE_VIS basic_string;
                               ^
/Users/jasonbenwell/Documents/koreader/base/thirdparty/kpvcrlib/crengine/crengine/src/lvmemman.cpp:77:57: error: 
      implicit instantiation of undefined template 'std::__1::basic_string<char,
      std::__1::char_traits<char>, std::__1::allocator<char> >'
    std::string errstr = std::string("FATAL ERROR #") + std::string(strb...
                                                        ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/iosfwd:209:32: note: 
      template is declared here
    class _LIBCPP_TEMPLATE_VIS basic_string;
                               ^
/Users/jasonbenwell/Documents/koreader/base/thirdparty/kpvcrlib/crengine/crengine/src/lvmemman.cpp:77:80: error: 
[...]

PR #6361, opened 2020-07-08
add macOS target

pazos:
Requires https://github.com/koreader/koreader-base/pull/1138

To be used in #6331

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6361)


platform/mac/do_mac_bundle.sh
```diff
@@ -0,0 +1,40 @@
...
+if [ -z "${1}" ]; then
+    echo "${0}: can't find KOReader build, please specify a path"
+    exit 1
+else
+    INSTALL_DIR="${1}"
+    VERSION="$(cut -f2 -dv "${1}/koreader/git-rev" | cut -f1,2 -d-)"
+fi
+
+{
+    echo '<?xml version="1.0" encoding="UTF-8"?>'
```
Frenzie, line 14: Wouldn't that be more legible with a heredoc?

kodev
```diff
@@ -125,6 +125,7 @@ SUPPORTED_TARGETS="
     debian          Debian package for current arch
     debian-armel    Debian package for generic armel devices
     debian-armhf    Debian package for generic armhf devices
+    macos           MacOS app bundle (x86_64)
```
Frenzie, line 128: Should it warn that it's only on Mac?

JasonInOttawa:
The attached do_mac_bundle produces an application with relocatable shared libraries. (I hope I'm using those words properly.)
[do_mac_bundle.sh.txt](https://github.com/koreader/koreader/files/4898118/do_mac_bundle.sh.txt)

pazos:
> The attached do_mac_bundle produces an application with relocatable shared libraries. (I hope I'm using those words properly.)

@JasonInOttawa: Indeed. That's a great job!. I adapted your patch to match code style and fix some shellcheck warnings.

That makes the app *almost* redistributable. It should work on a stock osx with just sdl2 and gettext installed via brew. Feel free to check and suggest more changes if you wish. 👍

platform/mac/do_mac_bundle.sh
```diff
@@ -0,0 +1,86 @@
...
+# remove things from the bundle
+rm -rf \
+    "${APP_PATH}/Contents/Resources/resources/fonts" \
+    "${APP_PATH}/Contents/Resources/resources/koreader.icns"
+
+# rename the bundle as an app
+mv "${APP_PATH}" "${APP_BUNDLE}.app"
+
+# optional: create dmg package
+if command_exists "create-dmgg"; then
```
pazos, line 81: typo here

kodev
pazos: yeah, or

```bash
if [ $(uname -s) != "Darwin" ]; then
    echo "You need a mac to build this package"
    exit 1
fi
```

JasonInOttawa:
Great stuff! Once this gets merged, will it be possible to just clone koreader per the build instructions and do a kodev release macos to get an application?

I'll be sure to report any issues I see. Thanks for everything.

pazos:
> Great stuff! Once this gets merged, will it be possible to just clone koreader per the build instructions and do a kodev release macos to get an application?

I do hope so!. With `./kodev release macos`. It will build KOReader-$VERSION.dmg if you have `create-dmg` installed, KOReader-$VERSION.app otherwise.

As I said is not strictly self-contained but work in an almost naked environment.
It is possible to target El Capitan and newer versions, even in Catalina. Just use the mininum supported abi  `export MACOSX_DEPLOYMENT_TARGET=10.11`

Users just need to install sdl2 and gettext with brew: `brew install sdl2 gettext`

platform/mac/do_mac_bundle.sh
```diff
@@ -0,0 +1,40 @@
...
+    echo '    <string>APPL</string>'
+    echo '    <key>CFBundleIconFile</key>'
+    echo '    <string>koreader</string>'
+    echo '</dict>'
+    echo '</plist>'
+} >"${INSTALL_DIR}/bundle/Contents/Info.plist"
+
+mv -v ${INSTALL_DIR}/bundle ${INSTALL_DIR}/../KOReader-mac-${VERSION}.app
+
+exit 0
```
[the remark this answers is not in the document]
pazos: done

platform/mac/do_mac_bundle.sh
pazos: done

kodev
pazos: done

pazos:
@Frenzie: if it is possible to build the mac package once a month (like debian ones) using some sort of free cloud service then makes some sense to try to remove dependency from brew (because brew itself requires xcode commandline tools). If not then we're fine with the package as is.

BTW, latest SDL version (2.0.12) added a way to populate the main app menu using a nib/xib file. So it is possible to tweak these menus without recompiling SDL (See https://github.com/SDL-mirror/SDL/commit/ea29fe5f84348e6aca50c3460b82c564719e0c0b#diff-728e2aec325ffff319c5c78479aac668). IMHO makes more sense to tweak SDL2 a litle further instead of trying to do our own cocoa app. In the end SDL2 uses cocoa under the hood and we aren't going to do a better job than SDL2 in a million years.

I won't plan to do any work on that. I'm fine with the app as is. But if @JasonInOttawa wants to tweak system menu or integrate a bit more with mac ecosystem then I would go for reverting https://github.com/SDL-mirror/SDL/commit/1179ddf3b9f8e5bb98fd3a3ad17e5e705b47efd5#diff-728e2aec325ffff319c5c78479aac668, which isn't needed in our case and just makes the behaviour a bit wonky.

Frenzie:
@pazos The Debian ones are also built daily, I just didn't see much point in making them as easily available since the AppImage is pretty much the same thing. Also it's a potential disk space issue. But you can grab them from GitLab every day.

> if it is possible to build the mac package once a month (like debian ones) using some sort of free cloud service then makes some sense to try to remove dependency from brew (because brew itself requires xcode commandline tools). If not then we're fine with the package as is.

I think GitHub Actions should be able to do that on tag, except as you know it's very dependably failing.

JasonInOttawa:
I'm also very happy with the application as it is. I'm running a fresh build and it's working great!

I'd love to see the brew dependencies go away and will look at the SDL situation @pazos mentions above. I'd also love to see the Mac application on the downloads page :)

pazos:
Ok. Merging as is. We can improve it later!

Fixes #6331 
Fixes #6351

JasonInOttawa:
> We can improve it later!

That's how all great things are accomplished! 

A fresh clone and build is working great. Thanks  - I'm really enjoying having the same reader on my Kobo as on my computer.

pazos:
I think we should force a single instance app on Catalina. Trying to open the same app twice will make the second instance to loop forever. It doesn't react to force close. 

To reproduce:
1 - double click app bundle in Catalina. It works
2 - open the app again by double clicking the same bundle. Now you need to restart your computer to be able to "quit" the 2nd instance of the program.

That doesn't happen on previous osx versions, where clicking the app bundle the second time does nothing.

pazos:
Ok. So SDL is awesome and generates a DROPFILE event when trying to open a file associated with the bundle. Huge win.

That's better than my hacky code. Leaving it here because it can be used on other platforms for single instance mode. It generates standard linux events and sends them via a named pipe:

```bash
#!/bin/sh

# IPC for macOS. Used to redirect commands to a running instance.
# Events generated with this script need to be handled by handleMiscEv.

# $1: full path to named pipe
# $2: event code
# $3: event value

KOREADER_DIR="${0%/*}/../Resources/koreader"

cd "${KOREADER_DIR}" || exit

NewMiscEv() {
    ./luajit <<EOF
    require("setupkoenv")
    local dump = require("dump")
    local ffi = require("ffi")
    local ev_time = ffi.new("timeval")
    ffi.C.gettimeofday(ev_time, nil)
    local ev = {
        type = 4,
        code = tonumber(${2}),
        value = "${3}",
        time = { sec = tonumber(ev_time.tv_sec), usec = tonumber(ev_time.tv_usec) }
    }
    local pipe = io.open("${1}", "wb")
    if pipe then
        pipe:write(dump(ev))
        pipe:close()
    end
EOF
}

NewMiscEv "${@}"
```

Issue #6362, opened 2020-07-09
corrupt ftp file download from CloudStorage


clach04:
* KOReader version: 2020.06 (also seen in 2020.4.1)
* Device: K3 (very old Kindle with keyboard)

#### Steps to reproduce

1. Setup ftp server under CloudStorage (my server happens to be ProFTPD under Linux but I suspect this is not relevant)
2. Download a file/ebook that is binary (e.g. zip file, cbz, mobi, etc.)
3. Download appears successful but file size will be incorrect and reader will fail to open/read metadata) - issue is newlines were converted to DOS format during download

PR #6363, opened 2020-07-09
Fix issues #6362 - ftp corruption of binary files

clach04:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6363)

### DIFF
--- a/frontend/apps/cloudstorage/ftpapi.lua
+++ b/frontend/apps/cloudstorage/ftpapi.lua
@@ -27,6 +27,7 @@ function FtpApi:ftpGet(u, command)
     p.password = util.urlDecode(p.password)
     p.command = command
     p.sink = ltn12.sink.table(t)
+    p.type = "i"  -- binary
     local r, e = ftp.get(p)
     return r and table.concat(t), e
 end


NiLuJe:
Obvious question is obvious: what happens for plain-text content?

They make the trip bit-exact, too?

(What does the default test mode does? OS-specific line-ending mangling?)

clach04:
> Obvious question is obvious: what happens for plain-text content?
> 
> They make the trip bit-exact, too?

* Before fix they get "changed" (corrupted is misleading) and KOreader displays them fine.
* With fix they do not get changed/corrupted and are exact-copies of the source -  and KOreader displays them fine.

> (What does the default test mode does? OS-specific line-ending mangling?)

Can you explain this more? I'm not sure what test mode is? I only ran this on a K3 device manually, I did not run any test suite (not sure how to do that).

NiLuJe:
Your first bullet-point answered my final question, thanks ;).

Frenzie:
@NiLuJe Was this one good to go?

NiLuJe:
It certainly makes sense, so I'd say yes, but I don't use the feature myself ;).

PR #6364, opened 2020-07-09
 Dispatcher: use sections

yparitcher:
use `Device` and `Crengine` and soon `Mupdf` sections to organize the dispatcher menu and allow the user to know when the action will apply

add a few device settings/events to dispatcher and by extension profiles.

depends on koreader/koreader#6292

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6364)

### DIFF
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -1,4 +1,5 @@
 local CreOptions = require("ui/data/creoptions")
+local Device = require("device")
 local Event = require("ui/event")
 local Screen = require("device").screen
 local UIManager = require("ui/uimanager")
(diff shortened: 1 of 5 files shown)


frontend/dispatcher.lua
```diff
@@ -254,6 +244,40 @@ function Dispatcher:addSubMenu(menu, location, settings)
...
+            return next(self[location][settings]) == nil
+        end,
+        callback = function(touchmenu_instance)
+            self[location][settings] = {}
+            if touchmenu_instance then touchmenu_instance:updateItems() end
+        end,
+    })
+    for _,section in ipairs({"Device", "Crengine",}) do
+        local submenu = {}
+         Dispatcher.addItem(self, submenu, location, settings, section)
```
Frenzie, line 275: nit

poire-z: (I guess the nit was about the indentation.)
But there's also `Dispatcher.addItem(self, ...)` which looks like it could be `Dispatcher:addItem(...)` (with a colon and without self, as usually when calling a method) - unless there's something else involved :)

frontend/dispatcher.lua
```diff
@@ -254,6 +244,40 @@ function Dispatcher:addSubMenu(menu, location, settings)
...
+        table.insert(menu, {
+            text = section,
```
Frenzie, line 277: Btw, I'm not sure this is necessarily the best idea (e.g., is crengine transliterated in different scripts?), although for advanced settings it probably doesn't really matter. I do however dislike looking at that uppercase `Crengine=true`.

poire-z: Or name these sections "Rolling" and "Paging" (we have a single "rolling" engine: crengine - but multiple "paging" engines: mupdf, djvu).
or RollingEngine, or isRolling, or else

frontend/dispatcher.lua
yparitcher: I added a comment:

Must be executed in the context of the caller to add items to the callers menu.
therefore declared as Dispatcher:addSubMenu(menu, location, settings)
but should be called as Dispatcher.addSubMenu(self, menu, location, settings)

the same applies to addItem

frontend/dispatcher.lua
yparitcher: I changed to using a k/v table for separate key and text. This should fix this issue

For the display text most users probably are not aware of rolling vs paging (crengine also has pages).
but maybe i should use epub and pdf? as those are the most obvious to the user.
please let me know of any better naming strategy :)

Frenzie: Something like dynamic and fixed layout?

yparitcher: let me explain better, the sections are used as submenus to divide the actions according to when they apply.

Device is always, Crengine (or whatever we name it) when using crengine etc. some items may be both in crengine and mupdf.
for example: font size is in both but calls a different event for each, to the user they would appear to be the same `Font Size` so we need different sections to let the user know which one to chose.

Therefore the string has to convey to the user which engine is being used something that the user does not necessarily pay attention to.

screenshot:
Reader_2020-07-09_140707

poire-z: `Reflowable documents` (epub, fb2, txt)
`Fixed layout documents` (pdf, djvu, pics)
(except that we have the option to "Reflow" PDF, making them reflowable :/)

So, that means you should have a profile for things crengine, and another for things pdf? Can't have one with both kind of settings (that would apply as no-op if the document if from the other kind) ?

yparitcher: > So, that means you should have a profile for things crengine, and another for things pdf? Can't have one with both kind of settings (that would apply as no-op if the document if from the other kind) ?

You can have both in one profile (I just did not yet implement mupdf events yet in the screenshot), and it will be a no-op, the sections are so the user is aware which ones will apply to a given document.

also some may be checked in both `GoToBeginning` is the same event for both so one action can be in both submenus, as it is reader specific

frontend/dispatcher.lua
poire-z: I'm not sure I get it, and what you mean with "context".
What object exactly is `self` when these are called in real-life?
A Dispatcher instance ? The Dispatcher static class object ? Something not a Dispatcher ? :)
If `self` is something else than a Dispatcher thingy, better to declare and use them with a dot so there's no confusion, and name the first argument something other than `self`.

yparitcher: self is whoever is calling Dispatcher, right now that is only Profiles but in the future also ReaderGesture.

see:
https://github.com/koreader/koreader/blob/509ee7bb8612ddb370f4f61d3e4a3209d7ec46ad/plugins/profiles.koplugin/main.lua#L107

i changed it to be declared as `Dispatcher.addSubMenu(caller, menu, location, settings)`

yparitcher:
> lgtm, was there anything left to do here?

I am waiting for #6292 and then i will add all the events to Dispatcher. Hopefully in time for 2020.07, that way it can be tested in the wild while only affecting profiles.

frontend/apps/reader/modules/readerlink.lua
```diff
@@ -685,13 +685,13 @@ function ReaderLink:onGoToExternalLink(link_url)
 end
 
 --- Goes back to previous location.
-function ReaderLink:onGoBackLink(show_notification_if_empty)
+function ReaderLink:onGoBackLink(dont_show_notification_if_empty)
```
poire-z, line 688: Why was that needed? Because you can't pass arguments when the events are sent by profile? (So we should build our future onSomething() that take argument to have the more sensible defaults when used as profile?)
May be just a shorter `no_notification_if_empty` for the name.

yparitcher: I ended up having a separate category that passes a argument

yparitcher:
Dispatcher should now have the same as ReaderGesture + what we parse from creoptions.

I still have to finish the events that expect a gesture object (frontlight increase/decrease, fontsize, and gotolink)

I think i got the sections right, but they are not set in stone, if anyone has a better way of sorting them please let me know.

since we are using pairs() the menu order is not the same every time. i am fine with that.

screenshot:
Reader_2020-07-12_171802
Reader_2020-07-12_172032

poire-z:
> since we are using pairs() the menu order is not the same every time. i am fine with that.

(May be for later, but it would be nice to have them ordered, by topic, a bit like the actions in the Gestures menu. And not have `Rotation` in the middle of 2 other `Something Wi-Fi` :)

We also have a (ffi/)util.orderedPairs() in case that would help to at least get them alphabetically ordered.

yparitcher:
> > since we are using pairs() the menu order is not the same every time. i am fine with that.
> 
> (May be for later, but it would be nice to have them ordered, by topic, a bit like the actions in the Gestures menu. And not have `Rotation` in the middle of 2 other `Something Wi-Fi` :)

I would do this but how can i process a table in order?

> We also have a (ffi/)util.orderedPairs() in case that would help to at least get them alphabetically ordered.

done for now

poire-z:
> I would do this but how can i process a table in order?

You made it a key/value table:
```lua
local settingsList = {
  show_frontlight_dialog = { category="none", event="ShowFlDialog", ...},
  toggle_frontlight = { category="none", event="ToggleFrontlight", ...},
}
```
I don't know if you're using the keys (somewhere else?), but if you do, yes, that table has no-order. You'd need another array to list the keys in the wanted order, and you could iterate that other table.

But if you don't need these keys (`show_frontlight_dialog`, `toggle_frontlight`...), just make that table an array:
```lua
local settingsList = {
  { category="none", event="ShowFlDialog", ...},
  { category="none", event="ToggleFrontlight", ...},
}
```
There, the order is preserved and you can use `ipairs`.

yparitcher:
> I don't know if you're using the keys (somewhere else?)

yes I use the keys, as for each menu that i call Dispatcher for I don't replicate the whole table rather i only save a k/v of the enabled items and the value to use. this makes execution much simpler and faster.

yparitcher:
I went with using another array for the order.
I have no preference for the order and where to put separators.
can someone post here a order for the `dispatcher_menu_order` reordered how you like and when everyone agrees I will use that as the order.

frontend/dispatcher.lua
```diff
@@ -12,37 +13,132 @@ local Dispatcher = {
...
...anager", "dictionary_lookup", "wikipedia_lookup", "fulltext_search", "file_search", "full_refresh", "night_mode", "suspend", "exit", "restart", "reboot", "poweroff", "show_menu", "toggle_hold_corners", "toggle_rotation", "wallabag_download", "calibre_search", "calibre_browse_tags", "calibre_browse_series", "favorites", "folder_up", "show_plus_menu", "folder_shortcuts", "prev_chapter", "next_chapter", "first_page", "last_page", "prev_bookmark", "next_bookmark", "go_to", "skim", "back", "previous_location", "latest_bookmark", "clear_location_history", "toc", "bookmarks", "book_statistics", "book_status", "book_info", "book_description", "book_cover", "show_config_menu", "toggle_bookmark", "toggle_inverse_reading_order", "cycle_highlight_action", "cycle_highlight_style", "kosync_push_progress", "kosync_pull_progress", "toggle_page_flipping", "toggle_reflow", "zoom", "rotation_mode", "visible_pages", "h_page_margins", "sync_t_b_page_margins", "t_page_margin", "b_page_margin", "view_mode", "block_rendering_mode", "render_dpi", "line_spacing", "font_size", "font_weight", "font_hinting", "font_kerning", "status_line", "embedded_css", "embedded_fonts", "smooth_scaling", "nightmode_images",
```
poire-z, line 140: This on a single line will be unmanageable :)

> I went with using another array for the order.

Might not be the best, we'd have to not forget to update 2 places when adding a new one.

I think defining them once in an array and building your k/v table from it for quick lookup is cleaner (suggested at the bottom of https://github.com/koreader/koreader/pull/6364#issuecomment-657282715).

poire-z:
Although it's du/tri-plicated in ReaderGesture.
https://github.com/koreader/koreader/blob/70f89c4df113bbfe4b8ea89730e71a36dfaddedc/frontend/apps/reader/modules/readergesture.lua#L20-L29
https://github.com/koreader/koreader/blob/70f89c4df113bbfe4b8ea89730e71a36dfaddedc/frontend/apps/reader/modules/readergesture.lua#L745-L753

> can someone post here a order for the dispatcher_menu_order reordered how you like and when everyone agrees I will use that as the order.

You could follow the order defined in the above array (2nd snippet).

frontend/dispatcher.lua
NiLuJe: The concern raised in https://github.com/koreader/koreader/pull/6364#issuecomment-657283193 was to avoid duplicating the full table, and instead only duplicating the key ;).

I'm of the mind that the performance point is moot either way, so I kinda like the current approach, but I definitely agree it should be broken down line by line (for readability, and diff-friendliness in case of updates).

poire-z: > the "new" key never makes it to the table, while forgetting to update the actual data table will crash by trying to index a nil or something, right?

Possibly for the crash (haven't checked how it is used), and yes, we won't see it in the menu - but will we always check/have to check if it's present in the menu? :)

But just adding an `id=` to the elements in a single table and using:
```lua
local settingsListById = {}
for _, t in ipairs(settingsList) do
   if t ~= true then
       settingsListById[t.id] = t
   end
end
```
saves 100 lines of code :) (yep, we have nearly that many settings).
(Well, that's just my preference - I like things tidy :)

NiLuJe: I'm also not one to be bothered by extra LOCs, so, eh ;D.

yparitcher: > This on a single line will be unmanageable :)

this was jsut for the first draft untill i got a proper order, was not planning on leaving it this way.

> I think defining them once in an array and building your k/v table from it for quick lookup is cleaner (suggested at the bottom of [#6364 (comment)](https://github.com/koreader/koreader/pull/6364#issuecomment-657282715)).
 
> As for your point about needing to update two things: forgetting to update the order array will essentially mean the "new" key never makes it to the table, while forgetting to update the actual data table will crash by trying to index a nil or something, right?

It will crash.

I like having the k/v table in the order it is in, as it makes adding new things easier, beacuse the items are sorted by where they come from & where they apply as is commented in the table. therefore i would rater have the display order in a seperate array even if it adds loc. (It still will be shorter than ReaderGesture :)

yparitcher:
> You could follow the order defined in the above array (2nd snippet).

that is also a out of date order. some reader stuff at the beginning some towards the end.
There is also the stuff i parse from `Creopt` which is not in that list. 
Also the `Dispatcher` list is sub-sorted by section so there are 3 independent parts to be sorted. (the sorting of device keys wont affect reader keys.)

frontend/dispatcher.lua
poire-z: > beacuse the items are sorted by where they come from & where they apply as is commented in the table. therefore i would rater have the display order in a seperate array 

OK, that's a convincing enough reason :)
(But may be _where they come from & where they apply_  would be just fine for the menu order ? :)

yparitcher: > (But may be _where they come from & where they apply_ would be just fine for the menu order ? :)

not enough, for example: rotation is (now) a device setting and should be listed to the user with the other device settings. however the values are generated by parsing `creopt` so in the k/v it is listed in it's place in `creopt`, this allows a developer making a change in `creopt` to easily see what is affected etc.

NiLuJe: Then simply the same order as the original menus themselves, in a dumb top-to-bottom, left-to-right order?

yparitcher: > Then simply the same order as the original menus themselves, in a dumb top-to-bottom, left-to-right order?

many of them are not in the menus but spread out over the ui (GoTo dialog or frontlight toggle)

poire-z:
Here, your single line made multilines, and re-ordered similarly to how they are in ReaderGesture menu (done quite dumb-ly, I let you do the sorting by sections :)

```
-- Missing page_jmp_back/fwd_10/1
"prev_chapter",
"next_chapter",
"first_page",
"last_page",
"prev_bookmark",
"next_bookmark",
"go_to",
"skim",
"back",
"previous_location",
"latest_bookmark",
-- missing follow_nearest_internal/_link
"clear_location_history",

"folder_up",
"show_plus_menu",
"folder_shortcuts",

"toc",
"bookmarks",
"reading_progress",
"book_statistics",

"book_status",
"book_info",
"book_description",
"book_cover",

"history",
"open_previous_document",
"favorites",
"filemanager",
"stats_calendar_view",

"dictionary_lookup",
"wikipedia_lookup",
"fulltext_search",
"file_search",

"full_refresh",
"night_mode",
"suspend",
"exit",
"restart",
"reboot",
"poweroff",

"show_menu",
"show_config_menu",
"show_frontlight_dialog",
"toggle_frontlight",
-- missing increase/decrease frontlight/_warmth

"toggle_hold_corners",
"toggle_gsensor",
"toggle_rotation",

"wifi_on",
"wifi_off",
"toggle_wifi",

-- missing increase/decrease_font
-- These could be moved below among the CRE stuff
"font_size",
"font_weight",
"font_hinting",
"font_kerning",

"toggle_bookmark",
"toggle_page_flipping",
"toggle_reflow",
"toggle_inverse_reading_order",
"zoom",
-- missing zoom individual values stuff
"cycle_highlight_action",
"cycle_highlight_style",
"wallabag_download",

"kosync_push_progress",
"kosync_pull_progress",

"calibre_search",
"calibre_browse_tags",
"calibre_browse_series",

-- Done with ReaderGesture stuff, now creoptions ordered as in the bottom pannel
"rotation_mode",
"visible_pages",

"h_page_margins",
"sync_t_b_page_margins",
"t_page_margin",
"b_page_margin",

"view_mode",
"block_rendering_mode",
"render_dpi",
"line_spacing",

-- Font stuff could be here

"status_line",
"embedded_css",
"embedded_fonts",
"smooth_scaling",
"nightmode_images",
```

yparitcher:
This should be ready.

yparitcher:
Please Review the Text strings.
They are mostly copied from ReaderGesture but there was not always a 1:1 mapping, and strings are not my strong point :)

frontend/dispatcher.lua
```diff
@@ -12,37 +13,241 @@ local Dispatcher = {
...
+
+    --paging reader settings
+    toggle_page_flipping = { category="none", event="TogglePageFlipping", title=_("Toggle page flipping"), paging=true,},
+    toggle_reflow = { category="none", event="ToggleReflow", title=_("Toggle reflow"), paging=true,},
+    zoom = { category="string", event="SetZoomMode", title=_("Zoom to"), args={"contentwidth", "contentheight", "pagewidth", "pageheight", "column", "content", "page"}, toggle={"content width", "content height", "page width", "page height", "column", "content", "page"}, paging=true,},
+
+    --parsed from CreOptions
+    rotation_mode = {category="string", device=true},
+    visible_pages = {category="string", rolling=true},
+    h_page_margins = {category="string", rolling=true},
```
poire-z, line 117: No `title=` for all of these ?
If it is pulled from elsewhere, might be worth a comment, so I don't have to look at the rest of the code (which I just quickly did and did not found anything).

frontend/dispatcher.lua
```diff
@@ -12,37 +13,241 @@ local Dispatcher = {
...
+    "calibre_browse_series",
+
+    "rotation_mode",
+
+    -- filemanager
+    "folder_up",
+    "show_plus_menu",
+    "folder_shortcuts",
+
+    --reader
```
poire-z, line 189: space

frontend/dispatcher.lua
```diff
@@ -94,84 +299,48 @@ function Dispatcher:init()
...
-                          self[location][settings][k] = nil
-                      else
-                          self[location][settings][k] = true
-                      end
-                      if touchmenu_instance then touchmenu_instance:updateItems() end
-                  end,
-               })
-            elseif settingsList[k].category == "toggle" then
+function Dispatcher.addItem(caller, menu, location, settings, section)
+    for _,k in ipairs(dispatcher_menu_order) do
```
poire-z, line 303: space: `for _, k`

frontend/dispatcher.lua
```diff
@@ -94,84 +299,48 @@ function Dispatcher:init()
...
+        if settingsList[k][section] == true and
+        (settingsList[k].condition == nil or settingsList[k].condition) then
```
poire-z, line 305: Indentation (even 2 x 4 to make it stand out of the next body). On on a single line, it's not too long.

frontend/dispatcher.lua
```diff
@@ -94,84 +299,48 @@ function Dispatcher:init()
...
+                    return caller[location][settings] ~= nil and caller[location][settings][k] ~= nil
                     end,
                     callback = function(touchmenu_instance)
-                        self[location][settings][k] = not self[location][settings][k]
-                        if touchmenu_instance then touchmenu_instance:updateItems() end
-                    end,
-                    hold_callback = function(touchmenu_instance)
-                        self[location][settings][k] = nil
+                        if caller[location][settings] ~= nil
+                        and caller[location][settings][k] then
```
poire-z, line 314: Same.

frontend/dispatcher.lua
```diff
@@ -214,46 +385,93 @@ function Dispatcher:addSubMenu(menu, location, settings)
...
                 for i=1,#settingsList[k].args do
                     table.insert(sub_item_table, {
                         text = tostring(settingsList[k].toggle[i]),
                         checked_func = function()
-                            return self[location][settings] ~= nil
-                            and self[location][settings][k] ~= nil
-                            and self[location][settings][k] == settingsList[k].args[i]
+                            return caller[location][settings] ~= nil
+                            and caller[location][settings][k] ~= nil
+                            and caller[location][settings][k] == settingsList[k].args[i]
```
poire-z, line 403: I also don't like that, would prefer more indentation for the continuation lines.

frontend/dispatcher.lua
```diff
@@ -214,46 +385,93 @@ function Dispatcher:addSubMenu(menu, location, settings)
...
                 table.insert(menu, {
                     text_func = function()
-                        return T(settingsList[k].title, self[location][settings][k])
+                        return T(settingsList[k].title, caller[location][settings][k])
                     end,
                     checked_func = function()
-                        return self[location][settings] ~= nil
-                        and self[location][settings][k] ~= nil
+                        return caller[location][settings] ~= nil
+                        and caller[location][settings][k] ~= nil
```
poire-z, line 416: Same.

frontend/dispatcher.lua
```diff
@@ -214,46 +385,93 @@ function Dispatcher:addSubMenu(menu, location, settings)
...
+        callback = function(touchmenu_instance)
+            caller[location][settings] = {}
+            if touchmenu_instance then touchmenu_instance:updateItems() end
+        end,
+    })
+    local section_list = {
+        {"device", _("Device")},
+        {"filemanager", _("Filemanager")},
+        {"rolling", _("Reflowable documents (epub, fb2, txt)")},
+        {"paging", _("Fixed layout documents (pdf, djvu, pics)")},
```
poire-z, line 464: Would it need an ellipsis after the 3 example formats, as there are more ?

frontend/dispatcher.lua
```diff
@@ -262,17 +480,19 @@ function Dispatcher:execute(settings, gesture)
             if settingsList[k].category == "none" then
                 self.ui:handleEvent(Event:new(settingsList[k].event))
             end
-            if settingsList[k].category == "toggle"
-            or settingsList[k].category == "absolutenumber"
+            if settingsList[k].category == "absolutenumber"
             or settingsList[k].category == "string" then
```
poire-z, line 484: Even if it aligns the content, I prefer some indentation :)

frontend/dispatcher.lua
```diff
@@ -262,17 +480,19 @@ function Dispatcher:execute(settings, gesture)
...
                 self.ui:handleEvent(Event:new(settingsList[k].event, v))
             end
+            if settingsList[k].category == "arg" then
+            -- the event can accept a gesture object or an argument
```
poire-z, line 488: Also with comments, that are comments about the body.

frontend/dispatcher.lua
```diff
@@ -262,17 +480,19 @@ function Dispatcher:execute(settings, gesture)
...
+                local arg = gesture or settingsList[k].arg
+                self.ui:handleEvent(Event:new(settingsList[k].event, arg))
+            end
             if settingsList[k].category == "incrementalnumber" then
-                if v then
-                    self.ui:handleEvent(Event:new(settingsList[k].event, v))
-                else
-                    self.ui:handleEvent(Event:new(settingsList[k].event, gesture))
-                end
+            -- the event can accept a gesture object or a number
```
poire-z, line 493: Same.

frontend/dispatcher.lua
```diff
@@ -12,37 +13,241 @@ local Dispatcher = {
...
+    show_config_menu = { category="none", event="ShowConfigMenu", title=_("Show bottom menu"), rolling=true, paging=true,},
+    toggle_bookmark = { category="none", event="ToggleBookmark", title=_("Toggle bookmark"), rolling=true, paging=true,},
+    toggle_inverse_reading_order = { category="none", event="ToggleReadingOrder", title=_("Toggle page turn direction"), rolling=true, paging=true,},
+    cycle_highlight_action = { category="none", event="CycleHighlightAction", title=_("Cycle highlight action"), rolling=true, paging=true,},
+    cycle_highlight_style = { category="none", event="CycleHighlightStyle", title=_("Cycle highlight style"), rolling=true, paging=true,},
+    kosync_push_progress = { category="none", event="KOSyncPushProgress", title=_("Push progress from this device"), rolling=true, paging=true,},
+    kosync_pull_progress = { category="none", event="KOSyncPullProgress", title=_("Pull progress from other devices"), rolling=true, paging=true,},
+    page_jmp = { category="absolutenumber", event="GotoViewRel" , min=-100, max=100, title=_("Go X pages"), rolling=true, paging=true,},
```
Frenzie, line 103:
```suggestion
    page_jmp = { category="absolutenumber", event="GotoViewRel", min=-100, max=100, title=_("Go X pages"), rolling=true, paging=true,},
```
Meaningless nit :-)

NiLuJe:
Nothing much to say as far as strings are concerned on my end, :+1: !

frontend/dispatcher.lua
```diff
@@ -12,37 +13,241 @@ local Dispatcher = {
...
+    prev_bookmark = { category="none", event="GotoPreviousBookmarkFromPage", title=_("Previous bookmark"), rolling=true, paging=true,},
+    next_bookmark = { category="none", event="GotoNextBookmarkFromPage", title=_("Next bookmark"), rolling=true, paging=true,},
+    go_to = { category="none", event="ShowGotoDialog", title=_("Go to"), rolling=true, paging=true,},
+    skim = { category="none", event="ShowSkimtoDialog", title=_("Skim"), rolling=true, paging=true,},
+    back = { category="none", event="Back", title=_("Back"), rolling=true, paging=true,},
+    previous_location = { category="arg", event="GoBackLink", arg=true, title=_("Back to previous location"), rolling=true, paging=true,},
+    latest_bookmark = { category="none", event="GoToLatestBookmark", title=_("Go to latest bookmark"), rolling=true, paging=true,},
+    follow_nearest_link = { category="arg", event="GoToPageLink", arg={pos={x=0,y=0}}, title=_("Follow nearest link"), rolling=true, paging=true,},
+    follow_nearest_internal_link = { category="arg", event="GoToInternalPageLink", arg={pos={x=0,y=0}}, title=_("Follow nearest internal link"), rolling=true, paging=true,},
```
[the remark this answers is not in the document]
yparitcher: I changed the frontlight handler to also accept a direct number value to change if there is no gesture (via profiles), the link handler requires a gesture object to determine where the user touched the screen (ges.pos) to get the closest link, or if there is no close link the first one on the page.
It was simpler to pass a dummy gesture object `{pos={x=0,y=0}}` then to try to rewire the link handler to handle not getting a gesture especially as there is no value the user can give to change it.

frontend/dispatcher.lua
yparitcher: as per the comment `-- parsed from CreOptions`

yparitcher:
@poire-z
about the indentation, when a condition is part of the if & the line starts with an operator (and, or) I think it is better to indent with the condition that it is part of, and not with the body.

poire-z:
> I think it is better to indent with the condition that it is part of, and not with the body

I don't, and I would probably fix that if I were to edit your file - but I probably won't ever touch it :)
(I also wouldn't indent it with the body, but further, so 2 indent levels just to distinguish it from the body.)

I don't think I had seen other `if` formatted like you prefer, so it's also about having our coding style consistent.
But if others are fine with your prefered formatting, go ahead.

frontend/dispatcher.lua
```diff
@@ -112,6 +112,7 @@ local settingsList = {
     zoom = { category="string", event="SetZoomMode", title=_("Zoom to"), args={"contentwidth", "contentheight", "pagewidth", "pageheight", "column", "content", "page"}, toggle={"content width", "content height", "page width", "page height", "column", "content", "page"}, paging=true,},
 
     -- parsed from CreOptions
+    -- the rest of the table elements are built from their counterparts in CreOptions
```
poire-z, line 115: But where ? I didn't find it when I was looking for it :)

Frenzie:
I prefer either ASCII art or dumb indentation for this use case.

```lua
-- ASCII art style looks nicer
-- Slightly more effort though
if blah
   or lala

-- Dumb indentation makes sense too
-- Arguably more so with tabs than with spaces
if blah
    or lala

-- Same level is a bit confusing at a glance
if blah
or lala
```

@yparitcher 
> about the indentation, when a condition is part of the if & the line starts with an operator (and, or) I think it is better to indent with the condition that it is part of, and not with the body.

Just stick `then` on a newline?

```lua
-- "indenting with the body"
if blah
    or lala then
    local body
```
vs
```lua
-- similar but not "indenting with the body"
-- it's indenting the conditions
if blah
    or lala
then
    local body
```

frontend/dispatcher.lua
yparitcher: https://github.com/koreader/koreader/blob/4e5def4282ebafe4eb0d089ecdd6a4af5448f96e/frontend/dispatcher.lua#L51-L93

yparitcher:
fixed all the concerns in review, I think

PR #6370, opened 2020-07-11
some mac app fixes

pazos:
Follow up:  #6361

Fixes a nasty bug in catalina by preventing multiple app instances. More info in https://github.com/koreader/koreader/pull/6361#issuecomment-657123046

Pinging @JasonInOttawa for a review.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6370)

### DIFF
--- a/platform/mac/do_mac_bundle.sh
+++ b/platform/mac/do_mac_bundle.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 # Script to generate mac application bundles for KOReader
+COPYRIGHT="Copyright © $(date +"%Y") KOReader"
 
 command_exists() {
     type "$1" >/dev/null 2>/dev/null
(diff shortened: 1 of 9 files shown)


JasonInOttawa:
I understand the changes to the plist file, and starting in $HOME, but I don't follow the multiple instance fix. Something has been done using SDL?

pazos:
> I understand the changes to the plist file, and starting in $HOME, but I don't follow the multiple instance fix. Something has been done using SDL?

Nope, is just this property: 

```
        <key>LSMultipleInstancesProhibited</key>
        <true/>
```

Until 10.14 it prevents multiple users running the application at once but in Catalina it does prevent running multiple instances within the same user. I couldn't find relevant info online but I think it has something to do with the new apps ported from ipad using wrapper code (with a name I can't remember). New IOS versions allow apps to run more than one instance so I think they removed the restriction on macOS too.

Until 10.14 the only way to open multiple instances of the same app is using `open -a KOReader` on the terminal. That should work in spite of this setting but I didn't test

JasonInOttawa:
Thanks very much. I read about the LSMultipleInstancesProhibited variable and my understanding matches yours. It's unfortunate that multiple users can't run the application, but keeping the nasty crash (having to reboot isn't nice :) ) is more important.

I did a clone and build (and did not set the MACOSX_DEPLOYMENT_TARGET variable) and things look good to me. Double-clicking on the bundle a second time brings the already-running application to the foreground. Right-clicking on an epub and selecting Open-With Koreader does open the epub in Koreader. 

I'm happy with the changes.

platform/mac/do_mac_bundle.sh
```diff
@@ -34,6 +48,48 @@ cat <<EOF >"${INSTALL_DIR}/bundle/Contents/Info.plist"
...
+                    <string>xps</string>
+                </array>
+                <key>CFBundleTypeIconFile</key>
+                <string>icon</string>
+                <key>CFBundleTypeName</key>
+                <string>docs</string>
+                <key>CFBundleTypeRole</key>
+                <string>Viewer</string>
+            </dict>
+	</array>
```
Galunid, line 84: Not sure if it matters, but `CFBundleTypeExtensions` is "Deprecated in OS X v10.5."(see [documentation](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-101685)). Also the outer `</array>` closing tag should be indented once more.

pazos: They still work in Catalina. Apple guidelines now recommend using UTIs. OS supported ones are in https://developer.apple.com/library/archive/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html#//apple_ref/doc/uid/TP40009259-SW1

I, personally, found too much of work trying to figure out which long name matches each extension (and adding our own UTI's, since most of the types are not recognized by osx), so I opted for document extensions and did a quick copy from https://github.com/koreader/koreader/blob/master/plugins/calibre.koplugin/extensions.lua :octocat: 

But yeah. UTIs are welcome!

Galunid: Well, if it works, then it's alright. Please check indentation, I believe the last `</array>` here uses [Tab], while the rest use spaces.
image

pazos: > Well, if it works, then it's alright.

Yeah. Thanks for pointing it out BTW. I did check the last version of calibre (4.20) and they are still using `CFBundleTypeExtensions`. As calibre has hundred of thousand of macOS users I think it is safe to use deprecated methods until they fail. Then copy the new UTIs from calibre as it supports all the filetypes we're supporting and much more!.

pazos:
Err, I don't understand this issue:

```
Warning: platform/mac/do_mac_bundle.sh does not abide by coding style, diff for expected style:
 
16d15
 
< 
```

platform/mac/do_mac_bundle.sh
```diff
@@ -13,6 +13,20 @@ else
     VERSION="$(cut -f2 -dv "${1}/koreader/git-rev" | cut -f1,2 -d-)"
 fi
 
+
```
Frenzie, line 16: @pazos It's complaining about this extra newline.
```suggestion
```

pazos:
I played a bit with brew dependencies. It turns out that except SDL2 the rest can be packaged in catalina and they will work in previous version (tested in High Sierra). Sadly that's not the case of SDL2, which fails with:

```
dlyd: lazy symbol binding failed: Symbol not found: _objc_alloc_init
  Referenced from: libSDL2.dylib (which was built for Mac OS X 10.15)
  Expected in: /usr/lib/libobjc.A.dylib

dlyd: Symbol not found: _objc_alloc_init
  Referenced from: libSDL2.dylib (which was built for Mac OS X 10.15)
  Expected in: /usr/lib/libobjc.A.dylib
```

So we can build a self-contained bundle with https://github.com/koreader/koreader-base/issues/1141

Frenzie:
@pazos Was this waiting for something?

pazos:
@Frenzie: Yeah, give me a minute!

pazos:
Is this ok to merge?.  It will fail some tests if the emulator *has* a battery

Frenzie:
@pazos Fine by me as long as it doesn't reintroduce fake noise failures here on GH since I just fixed all of them. ;-)

Also I assume it'd only be a few very specific tests that'd fail on a laptop, right?

pazos:
> Also I assume it'd only be a few very specific tests that'd fail on a laptop, right?

readerfooter_spec.lua has battery percent hardcoded to 0%, that's it. But since it will test against a few documents it will fail on all of them.

Frenzie:
Sounds good enough for now; I'll probably fix it up with a stub this afternoon unless you want to do that before merging.

Frenzie:
See http://olivinelabs.com/busted/#spies-mocks-stubs for what I mean, for example used here:

https://github.com/koreader/koreader/blob/e257c4e45ee6be7d21cd0358018ca577a8fbafce/spec/unit/autosuspend_spec.lua#L10-L13

pazos:
>  I'll probably fix it up with a stub this afternoon unless you want to do that before merging.

👍

JasonInOttawa:
Great stuff. Are the build instructions the same, or can I remove brew SDL(or any other things)?

pazos:
@JasonInOttawa: same instructions. You need brew to build the emulator (and to run it since it relies on system sdl2) and the mac application.

If you're not interested in building the emulator at all you can remove brew's sdl. The command `./kodev release macos` will build a fresh sdl2 library from source.

Sadly, libintl and libwebp are still needed on the build host.

JasonInOttawa:
Thank you! I've done a kodev release macos and brew uninstall sdl2 and will report any issues.

PR #6371, opened 2020-07-11
Tame a few tests that relied on `pairs` being somewhat deterministic

NiLuJe:
It *really* isn't ;p.

Re https://github.com/koreader/koreader-base/pull/1133

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6371)


NiLuJe:
@Frenzie: MenuSorter is possibly similarly affected (and/or that one test is), but it's much more of a blackbox than DepGraph to me at first glance :s

NiLuJe:
By which I mean, the only thing left failing on my end is

```
Running tests in ./spec/front/unit/menusorter_spec.lua
[==========] Running tests from scanned files.
[----------] Global test environment setup.
[----------] Running tests from ./spec/front/unit/menusorter_spec.lua
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 10: MenuSorter module should put menu items in the defined order
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 10: MenuSorter module should put menu items in the defined order (0.08 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 38: MenuSorter module should attach submenus correctly
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 38: MenuSorter module should attach submenus correctly (0.05 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 59: MenuSorter module should put orphans in the first menu
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 59: MenuSorter module should put orphans in the first menu (0.22 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 85: MenuSorter module should put orphans with sorting_hint in the right menu
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 85: MenuSorter module should put orphans with sorting_hint in the right menu (0.14 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 158: MenuSorter module should display submenu of orphaned submenu
{
    [1] = {
        [1] = {
            ["text"] = "NEW: Main",
            ["new"] = true,
            ["id"] = "main"
        },
        [2] = {
            [1] = {
                ["id"] = "submenu_item2",
                ["text"] = "Submenu item 2"
            },
            [2] = {
                ["id"] = "submenu_item1",
                ["text"] = "Submenu item 1"
            },
            ["sub_item_table"] = {
                [1] = {
                    ["id"] = "submenu_item2",
                    ["text"] = "Submenu item 2"
                },
                [2] = {
                    ["id"] = "submenu_item1",
                    ["text"] = "Submenu item 1"
                }
            },
            ["text"] = "NEW: Submenu",
            ["new"] = true,
            ["id"] = "submenu"
        },
        [3] = {
            ["text"] = "NEW: Search",
            ["new"] = true,
            ["id"] = "search"
        },
        [4] = {
            ["text"] = "NEW: Tools",
            ["new"] = true,
            ["id"] = "tools"
        },
        ["id"] = "setting",
        ["text"] = "Settings"
    },
    ["id"] = "KOMenu:menu_buttons"
}
./spec/front/unit/menusorter_spec.lua:185: Expected to be truthy, but value was:
(nil)

stack traceback:
        ./spec/front/unit/menusorter_spec.lua:185: in function <./spec/front/unit/menusorter_spec.lua:158>

[  FAILED  ] ./spec/front/unit/menusorter_spec.lua @ 158: MenuSorter module should display submenu of orphaned submenu (0.30 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 189: MenuSorter module should not treat disabled as orphans
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 189: MenuSorter module should not treat disabled as orphans (0.08 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 211: MenuSorter module should attach separator=true to previous item
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 211: MenuSorter module should attach separator=true to previous item (0.16 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 233: MenuSorter module should ignore separator as first item
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 233: MenuSorter module should ignore separator as first item (0.12 ms)
[ RUN      ] ./spec/front/unit/menusorter_spec.lua @ 255: MenuSorter module should compress menus when items from order are missing
[       OK ] ./spec/front/unit/menusorter_spec.lua @ 255: MenuSorter module should compress menus when items from order are missing (0.12 ms)
[----------] 9 tests from ./spec/front/unit/menusorter_spec.lua (17.55 ms total)

[----------] Global test environment teardown.
[==========] 9 tests from 1 test file ran. (17.60 ms total)
[  PASSED  ] 8 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ./spec/front/unit/menusorter_spec.lua @ 158: MenuSorter module should display submenu of orphaned submenu

 1 FAILED TEST
```

I may be missing a few, because ljsqlite3 is randomly failing to load on weird cdef and/or GC issues.

NiLuJe:
Same dealio w/ `Menu.itemTableFromTouchMenu`, it's using `pairs` and the test expects that to keep insertion order.

frontend/depgraph.lua
```diff
@@ -45,72 +93,92 @@ function DepGraph:removeNode(node_key)
     -- one in their override=, that have added themselves in
     -- this node's .deps). We don't want to lose these
     -- dependencies if we later re-addNode this node.
-    local node = self.nodes[node_key]
+    local node, index = self:getNode(node_key)
     if node then
         if not node.deps or #node.deps == 0 then
-            self.nodes[node_key] = nil
+            -- We need to clear *all* references to that node for it to be GC'ed...
+            node = nil --luacheck: ignore
```
poire-z, line 100: Why do you insist on `node = nil` (even luacheck tells you it's not needed)?  It's a local, it will get nil when out of scope.
(We usually have to do it for self.something = nil to be gc'ed, but there, I don't see why.)

frontend/depgraph.lua
```diff
@@ -45,72 +93,92 @@ function DepGraph:removeNode(node_key)
...
+        -- Is not the to be removed node, and has deps
+        if curr_node.key ~= node_key and curr_node.deps then
+            -- Walk that node's deps to check if it depends on us
+            for idx, dep_node_key in ipairs(curr_node.deps) do
+                -- If it did, wipe ourselves from there
                 if dep_node_key == node_key then
-                    remove_idx = idx
+                    -- Wipe all refs
+                    curr_node.deps[idx] = nil
+                    table.remove(self.nodes[i].deps, idx)
```
poire-z, line 115: Making a hole in the array ? Why not `table.remove(curr_node.deps , idx)` ?
Oh, but that's what that 2nd line does, as curr_node = self.nodes[i].
So, that first line is not needed and confusing.

I think you got yourself too much drown in this thinking about "refs" :/ I usually don't need to think like that.

frontend/depgraph.lua
NiLuJe: Because it irks me that this one annoys luacheck, but not the two others below ;p.

It is indeed completely superfluous, as are the other two, especially since those local's are never re-used later in the code.

(i.e., it's more a pedantic point than anything)

NiLuJe:
Random being random:

Current CI only shows one of the menu failures. Both `widget_menu_spec.lua` and `menusorter_spec.lua` have potentially-failing `pairs` usage.

NiLuJe:
Speaking of random:

This is probably related to the PRNG changes (hash beeing seeded properly?), as I left `LUAJIT_SECURITY_PRNG` set to 1 on all platforms.

Fun fact: this requires the `getrandom` *syscall*, so you may need a fairly recent device (Linux 3.17+, possibly later on ARM) to actually duplicate that.

Frenzie:
@NiLuJe As long as the orphans are there I'd say the order of the orphaned ones is less relevant, but sure, ipairs is probably more expected.

frontend/depgraph.lua
poire-z: "technically for show" ? What does that mean ? :)

What annoys me there is not so much the fact that you uneededly set it to `nil` first - but that on the first line, you access the thing via `curr_node.deps` - and on the second line via `self.nodes[i].deps` - which makes the reader think you're dealing with 2 different things - and spend some time verifying that they are indeed _not_ different things.
So, either the reader (me) is confused - or the author was :)

One of these make it less confusing:
```lua
curr_node.deps[idx] = nil -- uneeded but artistically for show :)
table.remove(curr_node.deps, idx)
```

```lua
self.nodes[i].deps[idx] = nil -- uneeded but artistically for show :)
table.remove(self.nodes[i].deps, idx)
```
Or I don't really get it ?!

NiLuJe: The nil assign is technically useless here because that local is never re-user in the function.

But that leads to an interesting query:

What happens to `curr_node.deps[idx]` after the `table.remove`? Is the ref still pointing at old content? Or would accessing it crash because it's attempting to index a nil? Or would it be pointing to something else because the array has been re-ordered by remove?

NiLuJe: On the other hand, the ref being used for the  `table.remove` doesn't matter, because it'll get at the actual table (i.e., it deref's).

Hence, using the "main" reference (self's) for it was also to make that fact more obvious (i.e., also documentation).

NiLuJe:
> @NiLuJe As long as the orphans are there I'd say the order of the orphaned ones is less relevant, but sure, ipairs is probably more expected.

Meaning? Should the test(s) be removed? Or are they exposing a real issue in these modules assumptions about pairs?

Because I *really* don't want to refactor MenuSorter & Menu's item_table handling to be array-based ^^.

frontend/depgraph.lua
NiLuJe: Okay turns out it's actively harmful because it does make the array sparse ;p.

NiLuJe: And as `ipairs` stops at the first hole (i.e., `nil` or hash), thats Bad(TM).

NiLuJe: And `table.remove` is also useless with sparse arrays (i.e., can't drop an index set to nil).

Frenzie:
> Meaning? Should the test(s) be removed? Or are they exposing a real issue in these modules assumptions about pairs? (i.e., are the tests broken, or are the modules broken?)

Meaning we're talking about this very specific piece of code, not the module.

https://github.com/koreader/koreader/blob/f4dad2fae8548b44a59a4ae6de9361673717ae77/frontend/ui/menusorter.lua#L160-L187

Please don't remove the tests. Just let it fail for the moment. ;-)

NiLuJe:
@Frenzie: Noted ;).

Any insight about the other one?

```
Running tests in ./spec/front/unit/widget_menu_spec.lua
[==========] Running tests from scanned files.
[----------] Global test environment setup.
[----------] Running tests from ./spec/front/unit/widget_menu_spec.lua
[ RUN      ] ./spec/front/unit/widget_menu_spec.lua @ 8: Menu widget should convert item table from touch menu properly
k       exit    v       table: 0x7f81949a77e8
k       navi    v       table: 0x7f81949a7650
./spec/front/unit/widget_menu_spec.lua:22: Expected objects to be the same.
Passed in:
(table: 0x7f81949a7e98) {
 *[1] = {
    [sub_item_table] = {
      [1] = { ... more }
      [2] = { ... more }
      [icon] = 'foo/bar.png' }
    [text] = 'navi' }
  [2] = {
    [callback] = function: 0x7f81949a75a0
    [text] = 'exit' } }
Expected:
(table: 0x7f81949a7868) {
 *[1] = {
   *[callback] = function: 0x7f81949a75a0
    [text] = 'exit' }
  [2] = {
    [sub_item_table] = {
      [1] = { ... more }
      [2] = { ... more }
      [icon] = 'foo/bar.png' }
    [text] = 'navi' } }

stack traceback:
        ./spec/front/unit/widget_menu_spec.lua:22: in function <./spec/front/unit/widget_menu_spec.lua:8>

[  FAILED  ] ./spec/front/unit/widget_menu_spec.lua @ 8: Menu widget should convert item table from touch menu properly (0.39 ms)
[----------] 1 test from ./spec/front/unit/widget_menu_spec.lua (23.06 ms total)

[----------] Global test environment teardown.
[==========] 1 test from 1 test file ran. (23.13 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ./spec/front/unit/widget_menu_spec.lua @ 8: Menu widget should convert item table from touch menu properly
```

Frenzie:
Hm, that looks more worrisome. But not otoh.

NiLuJe:
Okay, then I *think* I'm done for now ;).

I'll let it sit for a day in case another DepGraph nit comes to mind, then I'll drop the draft flag ;).

frontend/depgraph.lua
NiLuJe: Okay, it's actually a terrible idea anyway because it makes the array sparse, which is Very Bad :).

(c.f., other comments above).

frontend/depgraph.lua
poire-z: Not sure I knew about all that :) but yes, generally, the simpler the better :)

frontend/depgraph.lua
poire-z: > What happens to curr_node.deps[idx] after the table.remove? 

table.remove returns it, and shift everything after it in the array. And if you're not assigning it (e.g. you're not doing `local thatThing = table.remove(curr_node.deps, idx)`), and if it's not referenced elsewhere, its ref count should be down to 0 and it'll be candidate for gc.
If you previously accessed it `local oldThing = curr_node.deps[idx]`, you've added 1 to the ref count, that thing still exists (it's _old content_ if you want, but it's that thing :) it's just no more in the table.

> As such, the nil is for documentation purposes, to make it clear that that local ref ought to be dead, 

Possibly excessive and confusing documentation to me :) once you get the Lua way of things describe above.

But thanks for the simplication. It all looks fine to me, clear, and nothing confusing :)

frontend/util.lua
```diff
@@ -301,6 +301,53 @@ function util.tableMerge(t1, t2)
...
+    if key then
+        return key, t[key]
+    end
+
+    -- no more value to return, cleanup
+    t.__orderedIndex = nil
+    return
+end
+
+function util.orderedPairs(t)
```
poire-z, line 345: We have a function with the same name in base/ffi/util.lua :)
https://github.com/koreader/koreader-base/blob/b4124d64d4ea4403438f5a79e048f50e2f481f40/ffi/util.lua#L585-L632

NiLuJe: Oops :D.

I'll fix it later tonight ;).

NiLuJe:
(Ready for final review!)

PR #6377, opened 2020-07-15
Perception expander: disable margin increase when set to 0

### LINKED ISSUE #6376: Perception expander settings periodic resets
* KOReader version: 2020.06 Mosaic
* Device: Kindle KT3
I use perception expander since the last version of KOreader is presented. I found optimal settings for me and set it up to train my speed reading skill. So, periodically, KOreader resets custom settings to default (or some else?) ones. It happens approximately every 40-60 pages, sometimes more frequent.

These are my settings:
Line thickness: 1.5
Margin from edges: 0.275
Line intensity: 10

The values after reset I'll write a bit later

yparitcher:
Closes #6376

as per @NiLuJe suggestion.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6377)

### DIFF
--- a/plugins/perceptionexpander.koplugin/main.lua
+++ b/plugins/perceptionexpander.koplugin/main.lua
@@ -128,7 +128,7 @@ function PerceptionExpander:showSettingsDialog()
             {
                 text = "",
                 input_type = "number",
-                hint = T(_("Increase margin after pages. Current value: %1"),
+                hint = T(_("Increase margin after pages. Current value: %1\nSet to 0 to disable."),
                     self.shift_each_pages),
             },
         },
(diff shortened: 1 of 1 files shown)


plugins/perceptionexpander.koplugin/main.lua
```diff
@@ -128,7 +128,7 @@ function PerceptionExpander:showSettingsDialog()
             {
                 text = "",
                 input_type = "number",
-                hint = T(_("Increase margin after pages. Current value: %1"),
+                hint = T(_("Increase margin after pages. Current value: %1\n Set to 0 to disable."),
```
Galunid, line 131: `\n Set` the space probably shouldn't be there, as it may result in text looking like this
```
Increase margin after pages. Current value: %1
 Set to 0 to disable
```
Otherwise seems fine to me

PR #6378, opened 2020-07-15
Implement numeric collator in file manager

### LINKED ISSUE #6374: Question: File manager - numeric collator implementation ideas
Current result when sorting using "plain" string comparison:
```
AAA 1
AAA 10
AAA 2
```
Expected result when using numeric collator:
```
AAA 1
AAA 2
AAA 10
```

I'm not sure if it's a good idea, but my idea to implement would be something like this:
Strip both strings of numbers and compare them. If they are different, then fallback to "plain" string comparison(`strcoll(a, b)`), otherwise get numbers from the string and compare them.
```lua
if a.name:gsub("%d", "") == a.name:gsub("%d", "") then
    -- just draft, it would need additional checks in case there are no numbers in the text
    return tonumber(a.name:match("%d+")) < tonumber(b.name:match("%d+")
else
    return self.strcoll(a.name, b.name)
end
```
This works for my purposes, but it would fail on something like this
```
AAA 1 -- after stripping numbers we get AAA
AAA 2
AA1A 10 -- after stripping numbers we get AAA too
```
Another idea would be to iterate over both strings at the same time, until we find number/difference. Are both of this viable options? My concern is mostly about performance and supporting decimal numbers. Alternatively I was thinking of adding leading zeros to filenames when comparing them, but that's more of a hack. I'd be grateful for any input.

Galunid:
Allows users to detect numbers in text when sorting filenames. We get:
```
AAA 1
AAA 2
AAA 10
```
instead of 
```
AAA 1
AAA 10
AAA 2
```
Most of the code was adapted from http://notebook.kulchenko.com/algorithms/alphanumeric-natural-sorting-for-humans-in-lua.
closes #6374

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6378)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -944,7 +944,7 @@ function FileManager:getSortingMenuTable()
     local fm = self
     local collates = {
         strcoll = {_("filename"), _("Sort by filename")},
-        numeric = {_("numeric"), _("Sort by filename(detect numbers)")},
+        numeric = {_("numeric"), _("Sort by filename (natural sorting)")},
         strcoll_mixed = {_("name mixed"), _("Sort by name – mixed files and folders")},
         access = {_("date read"), _("Sort by last read date")},
         change = {_("date added"), _("Sort by date added")},
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -944,6 +944,7 @@ function FileManager:getSortingMenuTable()
     local fm = self
     local collates = {
         strcoll = {_("filename"), _("Sort by filename")},
+        numeric = {_("numeric"), _("Sort by filename(detect numbers)")},
         strcoll_mixed = {_("name mixed"), _("Sort by name – mixed files and folders")},
         access = {_("date read"), _("Sort by last read date")},
         change = {_("date added"), _("Sort by date added")},
@@ -985,6 +986,7 @@ function FileManager:getSortingMenuTable()
             set_collate_table("modification"),
             set_collate_table("size"),
             set_collate_table("type"),
+            set_collate_table("numeric"),
             {
                 text_func =  get_collate_percent,
                 checked_func = function()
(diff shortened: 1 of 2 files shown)


frontend/ui/widget/filechooser.lua
```diff
@@ -193,6 +193,16 @@ function FileChooser:genItemTableFromPath(path)
...
         end
+    elseif self.collate == "numeric" then
+        -- adapted from: http://notebook.kulchenko.com/algorithms/alphanumeric-natural-sorting-for-humans-in-lua
+        local function addLeadingZeroes(d)
+            local dec, n = string.match(d, "(%.?)0*(.+)")
+            return #dec > 0 and ("%.12f"):format(d) or ("%s%03d%s"):format(dec, #n, n)
+        end
+        sorting = function(a, b)
+            return tostring(a.name):gsub("%.?%d+", addLeadingZeroes)..("%3d"):format(#b.name)
+         < tostring(b.name):gsub("%.?%d+",addLeadingZeroes)..("%3d"):format(#a.name)
```
Frenzie, line 204: slightly odd indent here?

Galunid: Indeed, I tried to fix it, but I'm not sure if it's alright. I always had trouble indenting code spanning multiple lines 😅

Frenzie: I'd say don't worry too much about line length. :-)

It's mainly an indicator that things could be getting too complex if you're getting super deep loops or something.

frontend/apps/filemanager/filemanager.lua
```diff
@@ -944,6 +944,7 @@ function FileManager:getSortingMenuTable()
     local fm = self
     local collates = {
         strcoll = {_("filename"), _("Sort by filename")},
+        numeric = {_("numeric"), _("Sort by filename(detect numbers)")},
```
NiLuJe, line 947:
```suggestion
        numeric = {_("numeric"), _("Sort by filename (detect numbers)")},
```

NiLuJe: (Also usually called "natural sorting", FWIW).

Galunid: "Sort by filename (natural sorting)" - want me to rename it like that, or keep as is?

NiLuJe: Yup! (c.f. https://en.wikipedia.org/wiki/Natural_sort_order)

PR #6383, opened 2020-07-16
desktop/emulator changes

pazos:
SimulateSuspend/SimulateResume are implemented for emulator and no-op for other sdl platforms.
viewport/portrait/touchless toggles via environment can be useful on other sdl platforms as well.
Fixed calling commands on OSX
~~Don't set a window icon on OSX, the bundle icon will be shown on the dock.~~

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6383)

### DIFF
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -5,40 +5,24 @@ local logger = require("logger")
 local function yes() return true end
 local function no() return false end
 
--- xdg-open is used on most linux systems
-local function hasXdgOpen()
-    local std_out = io.popen("xdg-open --version 2>/dev/null")
-    local version = nil
-    if std_out ~= nil then
-        version = std_out:read()
-        std_out:close()
-    end
-    return version ~= nil
-end
-
--- open is the macOS counterpart
-local function hasMacOpen()
-    return os.execute("open >/dev/null 2>&1") == 256
-end
-
--- get the name of the binary used to open links
-local function getLinkOpener()
-    local enabled = false
-    local tool = nil
-    if jit.os == "Linux" and hasXdgOpen() then
-        enabled = true
-        tool = "xdg-open"
-    elseif jit.os == "OSX" and hasMacOpen() then
-        enabled = true
-        tool = "open"
-    end
-    return enabled, tool
-end
-
--- differentiate between urls and commands
 local function isUrl(s)
-    if type(s) == "string" and s:match("*?://") then
-        return true
+    return type(s) == "string" and s:match("*?://")
+end
+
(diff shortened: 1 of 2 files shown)


frontend/device/sdl/device.lua
```diff
@@ -130,45 +108,43 @@ local Emulator = Device:new{
...
         self.isAlwaysPortrait = yes
     end
 
     self.hasClipboard = yes
     self.screen = require("ffi/framebuffer_SDL2_0"):new{device = self, debug = logger.dbg}
-
-    local ok, re = pcall(self.screen.setWindowIcon, self.screen, "resources/koreader.png")
-    if not ok then logger.warn(re) end
-
+    if self.isEmulator or jit.os ~= "OSX" then
```
Frenzie, line 144: Um, what?

pazos: ??. There's something I don't undertand. The intent is don't set the window icon if we're running within an app bundle on osx. That doesn't happen on the emulator even when `jit.os == "OSX"`

Frenzie: Since the checks go from left to right, self.isEmulator is an oddly restrictive one to start with.

> The intent is don't set the window icon if we're running within an app bundle on osx.

Right, but why not?

pazos: err, I give up! To be honest is not that important :p. It is not need when the app runs inside an application bundle (resources/koreader.icns) will be used instead.

pazos:
This one is ready for a quick review. I'm going to bump base with a few SDL changes and don't want to mix things

frontend/device/sdl/device.lua
```diff
@@ -80,16 +62,12 @@ local Device = Generic:new{
...
-        else
-            return os.execute('env -u LD_LIBRARY_PATH '..tool.." '"..link.."'") == 0
-        end
+        return runCommand(tool .. " '" .. link .. "'")
     end,
     canExternalDictLookup = yes,
     getExternalDictLookupList = getExternalDicts,
     doExternalDictLookup = function(self, text, method, callback)
-        external_dict_when_back_callback = callback
+        external_dicts_callback = callback
```
poire-z, line 70: I don't remember much how that works - but if you're just renaming it for aesthetics, and if it is still really called only when back from external dict, add this fact as a comment so a future reader knows that. (Unless it's mentionned elsewhere).

pazos: Thank you! I will leave it out the equation for now. I have a branch with refactored 3rd party app integration which I plan to PR soon.

PR #6385, opened 2020-07-16
german keyboard layout qwertz

zwim:
Add support for German keyboard. It's mainly `en_keyoard.lua` with changed z/Z and y/Y

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6385)


poire-z:
If you're only changing a few keys, it might be better to tweak the en_keyboard like we do in the es_ and fr_ ones (so you benefit from future updates to the other layouts or popups of the en_ one).

zwim:
You are right.

frontend/ui/data/keyboardlayouts/de_keyboard.lua
```diff
@@ -0,0 +1,9 @@
+-- Start with the english keyboard layout (deep copy, to not alter it)
+local de_keyboard = require("util").tableDeepCopy(require("ui/data/keyboardlayouts/en_keyboard"))
+
+local keys = de_keyboard.keys
+
+keys[1][6][1], keys[3][2][1] = keys[3][2][1], keys[1][6][1] -- Z <-> Y
+keys[1][6][2], keys[3][2][2] = keys[3][2][2], keys[1][6][2] -- z <-> y
+
+return de_keyboard
```
poire-z, line 9: `\n` at end of file please :)
(not super important, but github complains, and I sometimes had to fight with git when updating such files).

PR #6386, opened 2020-07-16
SetNightMode event & Night Mode refresh rate

yparitcher:
idea from #5079

allows setting night mode directly not only toggling it.

If the user has not set a separate refresh rate for night mode
the default one will be used, as was the previous behavior

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6386)

### DIFF
--- a/frontend/device/devicelistener.lua
+++ b/frontend/device/devicelistener.lua
@@ -17,7 +17,15 @@ function DeviceListener:onToggleNightMode()
     local night_mode = G_reader_settings:isTrue("night_mode")
     Screen:toggleNightMode()
     UIManager:setDirty("all", "full")
+    UIManager:ToggleNightMode(not night_mode)
     G_reader_settings:saveSetting("night_mode", not night_mode)
+end
+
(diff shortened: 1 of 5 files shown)


frontend/device/devicelistener.lua
```diff
@@ -20,6 +20,15 @@ function DeviceListener:onToggleNightMode()
     G_reader_settings:saveSetting("night_mode", not night_mode)
 end
 
+function DeviceListener:onSetNightMode(night_mode_on)
+    local night_mode = G_reader_settings:isTrue("night_mode")
+    if (night_mode_on and not night_mode) or ( not night_mode_on and night_mode) then
```
Frenzie, line 25: That space on purpose? :-)
```suggestion
    if (night_mode_on and not night_mode) or (not night_mode_on and night_mode) then
```

yparitcher:
@Frenzie Should i wait for the release to merge this?

Frenzie:
Seems safe enough. I just triggered a manual pipeline on GitLab to ease testing for more people.

PR #6397, opened 2020-07-19
remarkable: add more detail to README about launcher

tcrs:
This adds a little more detail about how to use koreader with a different launcher on reMarkable

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6397)

### DIFF
--- a/platform/remarkable/README_remarkable.txt
+++ b/platform/remarkable/README_remarkable.txt
@@ -16,15 +16,15 @@ with a USB cable.
 
 # Launching using the included launcher
 
-koreader includes a very simple launcher called `button-listen` which starts
-koreader when the middle button is held down for 3 or more seconds in the
-xochitl UI.
+KOReader includes a very simple launcher called `button-listen` which starts
+KOReader when the middle button is held down for 3 or more seconds in the
+default UI (xochitl).
 
    cp -v koreader/*.service /etc/systemd/system/
    systemctl enable --now button-listen
 
-To return to xochitl just exit koreader (swipe down from the top of the screen,
-select icon in the top right, Exit, Exit).
+To return to the default UI just exit KOReader (swipe down from the top of the
+screen, select icon in the top right, Exit, Exit).
 
 Some reMarkable software updates will wipe the new systemd units so you will have
 to run the two install steps again when that happens.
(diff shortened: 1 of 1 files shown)


platform/remarkable/README_remarkable.txt
```diff
@@ -13,12 +13,35 @@ with a USB cable.
    ssh root@<your-remarkable>
 
    unzip <koreader-package>.zip
+
+# Launching using the included launcher
+
+koreader includes a very simple launcher called `button-listen` which starts
```
Frenzie, line 19:
```suggestion
KOReader includes a very simple launcher called `button-listen`, which starts
```

platform/remarkable/README_remarkable.txt
```diff
@@ -13,12 +13,35 @@ with a USB cable.
...
+koreader when the middle button is held down for 3 or more seconds in the
+xochitl UI.
```
Frenzie, line 21: Do people know what that is? A subtle hint like "the default xochitl UI" might be preferable.

platform/remarkable/README_remarkable.txt
```diff
@@ -13,12 +13,35 @@ with a USB cable.
...
+
    cp -v koreader/*.service /etc/systemd/system/
    systemctl enable --now button-listen
 
-Hold down the middle button for 3 seconds to start koreader. To return to
-xochitl just exit koreader (swipe down from the top of the screen, select icon
-in the top right, Exit, Exit).
+To return to xochitl just exit koreader (swipe down from the top of the screen,
```
Frenzie, line 26:
```suggestion
To return to xochitl just exit KOReader (swipe down from the top of the screen,
```

platform/remarkable/README_remarkable.txt
```diff
@@ -13,12 +13,35 @@ with a USB cable.
...
+Alternatively, if you want the default behaviour of koreader exit starting up
+xochitl you can run (once on installation):
+
+   cp -v koreader/koreader.service /etc/systemd/system/
+
+And then the launcher needs to run:
+
+   systemctl start koreader
+
+To start koreader (this is the command which `button-listen` runs).
```
Frenzie, line 47:
```suggestion
To start KOReader (this is the command which `button-listen` runs).
```

But it's probably clearer to just merge it into the sentence above.

platform/remarkable/README_remarkable.txt
tcrs: Ah yes, I've corrected the captialisation of koreader->KOReader throughout

platform/remarkable/README_remarkable.txt
tcrs: Good point, I've mentioned xochitl the first time and just used Default UI from then on.

Issue #6344, opened 2020-07-03
Question: fb2 in-page footnotes justified


hius07:
https://github.com/poire-z, can you please help me to have fb2 in-page footnotes justified, same as popup footnote.
I've played with User style tweaks, cr3.ini, footnotewidget.lua with no success.
![Reader_2020-07-03_062817](https://user-images.githubusercontent.com/62179190/86462203-e6b7ac00-bd33-11ea-9297-51b4d07f34f6.png)
![Reader_2020-07-03_062826](https://user-images.githubusercontent.com/62179190/86462226-f20ad780-bd33-11ea-9e84-92d6093a8101.png)

poire-z:
A user style tweak with this should work:
```css
body[name="notes"] section,
body[name="comments"] section {
   text-align: justify !important;
}
```
Does that work ?
If not, can you try to see by inspecting the HTML where a `text-align: left` comes from.

I added recently `text-align: justify;` to the "book style tweak" sample:
https://github.com/koreader/koreader/blob/8a61e70fadb74189b1196ac9a7d07e49f6c0111a/frontend/apps/reader/modules/readerstyletweak.lua#L591-L602
Because I sometimes (but rarely) got them non-justified, and I figured it would not hurt there (a user can just erase that sample line).
Dunno if we should switch all in-page footnotes tweaks to `text-align: justify`.

poire-z:
Even if it works, can you try to see where a text-align: left comes from. Obviously, it's not you who set it via another style tweak.

hius07:
The user style tweak doesn't work.
I've checked cr3.ini, fb2.css, cr3.css, the book itself - no left-aligned instructions.
The text of the footnote is left-aligned both in the in-page footnote and in the end of the book.

hius07:
I've applied the sample of Book-specific tweak - doesn't work.

hius07:
I cannot trace the source of the text-align:left instruction, the html code shows just <body name="notes">, <section> and <p>.

poire-z:
Try this, forcing it on all descendants:
```css
body[name="notes"] section *,
body[name="comments"] section * {
   text-align: justify !important;
}
```

It's only on this specific FB2 book (can you share it?), or does it happen with all FB2 books?

hius07:
Doesn't work. All books.

poire-z:
So, please share one (that you can share), so we can see if it's you only - and if I can reproduce I'll investigate.

[...]

PR #6405, opened 2020-07-20
bump crengine: various fixes and FB2 tweaks

### LINKED ISSUE #6391: Koreader crashes when turning to a particular page
* KOReader version: 2020.06
* Device: Kobo Forma

#### Issue

Koreader crashes when turning to a particular page. I'd include the epub, but it's copyrighted :)

I get a popup to sync reading position, and then a crash when I go to a particular page.

Crash log attached.
[crash.log](https://github.com/koreader/koreader/files/4942203/crash.log)

### LINKED ISSUE #6394: KOReader crashes when opening epub with embedded fonts
* KOReader version: v2020.06-60
* Device: Kobo Aura One

#### Issue

When opening some ePub with embedded fonts, KOReader crashes with the following message:

`CRE: document loaded, but styles re-init needed (cause: embedded fonts)
CRE: styles re-init needed after load, re-rendering
Segmentation fault
!!!!
Uh oh, something went awry... (Crash nÂ°1: 07/19/20 @ 14:03:07)`

If I open the epub with the stable release v2020.06 the book opens fine.
If I remove the embedded font information with Calibre the book also opens fine in v2020.06-60.

#### Steps to reproduce

Open attached embedded_font_crash_scrambled.epub file

[embedded_font_crash_scrambled.zip](https://github.com/koreader/koreader/files/4943634/embedded_font_crash_scrambled.zip)

[crash.log](https://github.com/koreader/koreader/files/4943631/crash.log)

poire-z:
Includes:
- New German hyphenation patterns https://github.com/koreader/crengine/pull/356
- CSS: adds `text-align-last: -cr-right-if-not-first` private keywords https://github.com/koreader/crengine/pull/357
- CSS: adds `-cr-only-if: fb2-document`
- fb2.css: fix verses, minor cleanup
- FB2 footnotes: fix spacing between number and text
- ldomTextCollector: guess block boundaries better
- Text: fix read/write outside array bounds
- Text: fix read outside array bounds https://github.com/koreader/crengine/pull/358
- Fix crash when more than 64 nested elements - Closes #6391 
- Fix possible crash on book with embedded fonts load - Closes #6394 
- (Upstream) lvstsheet: avoid compilation warnings https://github.com/koreader/crengine/pull/359
- Strut confining: don't when there is no strut
- CSS: switch `-cr-hint` from enum to bitmap

`Style tweaks: limit EPUB/FB2 footnotes to these formats`
Avoid conflicts having "In-page EPUB footnotes (smaller)" affects FB2 documents that use `<a type=note>12</a>` which would then get smaller if both tweaks are enabled (which they are, by default). See https://github.com/koreader/koreader/issues/6344#issuecomment-658273890
Footnote popups: add styles for some non-HTML FB2 tags, so MuPDF can render them as expected. See https://github.com/koreader/koreader/issues/6344#issuecomment-657007634 (I just styled the obvious, may be more styling is needed, dunno.)

This PR fixes most issues with FB2 noticed from https://github.com/koreader/koreader/issues/6344#issuecomment-655936850 and later comments.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6405)

### DIFF
--- a/frontend/ui/data/css_tweaks.lua
+++ b/frontend/ui/data/css_tweaks.lua
@@ -338,8 +338,8 @@ ruby { display: inline !important; }
                 id = "lineheight_all_normal_strut_confined";
                 title = _("Enforce steady line heights"),
                 description = _("Prevent inline content like sub- and superscript from changing their paragraph line height."),
-                priority = -5, -- so other -cr-hint can override (this one has effect only on inline content)
-                css = [[* { -cr-hint: strut-confined; }]],
+                -- strut-confined is among the few cr-hints that are inherited
+                css = [[body { -cr-hint: strut-confined; }]],
                 separator = true,
             },
             {
@@ -650,13 +650,17 @@ This is just an example, that will need to be adapted into a user style tweak.]]
                     title = _("In-page FB2 footnotes"),
                     description = _([[
 Show FB2 footnote text at the bottom of pages that contain links to them.]]),
+                    -- Restrict this to FB2 documents, even if we won't probably
+                    -- match in any other kind of document
                     css = [[
 body[name="notes"] section {
-    -cr-hint: footnote-inpage;
-    margin: 0 !important;
+    -cr-only-if: fb2-document;
+        -cr-hint: footnote-inpage;
+        margin: 0 !important;
 }
 body[name="notes"] > section {
-    font-size: 0.75rem;
+    -cr-only-if: fb2-document;
+        font-size: 0.75rem;
 }
                     ]],
                 },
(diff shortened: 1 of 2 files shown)


frontend/ui/data/css_tweaks.lua
```diff
@@ -650,12 +650,16 @@ This is just an example, that will need to be adapted into a user style tweak.]]
                     title = _("In-page FB2 footnotes"),
                     description = _([[
 Show FB2 footnote text at the bottom of pages that contain links to them.]]),
+                    -- Restrict this to FB2 documents, even if we won't probably
+                    -- match in any other kind of document
                     css = [[
 body[name="notes"] section {
+-cr-only-if: fb2-document;
```
Frenzie, line 657: This rather… distracting indentation is on purpose to make it stand out?

poire-z: Yes. I used that kind of layout to express more clearly what the "only if" affects, as in:
https://github.com/koreader/koreader/blob/201d3d8ad1c9e6432283439e3c14abfd9569bc2f/frontend/ui/wikipedia.lua#L1015-L1039
I can add another level of indentation if you prefer, we have enough width to show them non wrapped:
image
=>
image

Frenzie: I think that would be nicer to look at (like in the Wikipedia one). :+1:

poire-z:
Don't merge yet - I forgot some simplifications in the style tweaks, now that we have inheritance, we can replace i.e. `* { -cr-hint: strut-confined }` with just setting it on BODY.

Issue #6406, opened 2020-07-20
FR: [zoom clicked area in comics to read text]


msh2050:
It will be a great enhancement if we can show a box of a zoomed image from the clicked area.
I'm not a Lua developer (although it seems easy to learn) I need help to find these codes in the app to make my widget. 
1- handle touch (event and location)
2- crop the image
3- display the image from the buffer

I complete the development environment and find som good starting stuff.
any hit will be appreciated

regards

poire-z:
Can you explain your use case, and why it would be a great enhancement ?
With PDF only, or also with EPUBs ? (because in EPUBs, you can zoom on image, and increase the font size).
On PDF I understand it could be needed, but we have Reflow mode that usually helps when it's needed.

Anyway, no real code location to give to you. It depends on how you would like it to work, and how you would go at avoiding conflicts with the existing stuff that happens on touch :)
[For now](https://github.com/koreader/koreader/blob/201d3d8ad1c9e6432283439e3c14abfd9569bc2f/frontend/apps/reader/modules/readerhighlight.lua#L57-L84), touching a screen location, or drawing a rectangle would select the text (for dict lookup or for text highlighting). So, you'd have to make that an option to get that alternative behaviour - and find a way to toggle easily between the behaviours - or find an alternative yet not used gesture.
Nothing obvious comes to my mind that would avoid conflict - and code complexity :/

msh2050:
@poire-z thanks for replay
I attached two images to explain my use case... 
It will be great because reading CBZ files(manga/comics)  is difficult even with rotation/reflow... 

<img src="https://user-images.githubusercontent.com/9201213/87970780-e73e9980-cacc-11ea-9d8a-dcb8634afc17.jpg" width="300" />


<img src="https://user-images.githubusercontent.com/9201213/87970812-f6bde280-cacc-11ea-8486-1f9753ea1991.jpg" width="300" />

Galunid:
Perhaps `bottom menu -> page crop -> manual` may work for this purpose?

poire-z:
Thanks for the illustrative pictures :)

So I guess one option would be to have it done a bit like for EPUBs. In a PDF/PIC/CBZ...:
- detect if the Hold (long press) is on some text or some non text area (dunno if MuPDF/Kopt provides some API for that)
- if on text: dict lookup, or show the highlight dialog
[...]

Issue #5524, opened 2019-10-23
FR: Free zoom


Frenzie:
* KOReader version: 2019.10
* Device: all

#### Issue
While the various zoom to fit X options are fantastic and cover just about any document I normally come across, messing about with cropping just to zoom to a section of the page is arduous at best. Page flipping mode features a slightly more free zoom & pan but it doesn't always work out.

Here's how it used to work in KPV (from [here](https://www.mobileread.com/forums/showthread.php?p=3906971#post3906971)), which is fairly straightforward:

> In KPV (Kindle DX) I can zoom out/in using ALT + <> (prev/next page) buttons, by predefined zoom step e.g. 16 %. (using SHIFT + <> buttons for zoom step decrease/increase)
>
>I can pan the active view using joypad stick (moving it up/down/left/right), also by predifined panning step e.g. 60 pixels.
>
>SHIFT + left/right would decrease/increase predifined x panning step.
>
>SHIFT + up/down would decrease/increase y panning step.

Frenzie:
Copying from https://github.com/koreader/koreader-base/issues/1148#issuecomment-661894276

* The zoom to box feature is great *as a starting point* and it's neat most of the time but it doesn't always work out. You want much freer zoom.
  Some details on things we might want in free zoom here: https://github.com/koreader/koreader/issues/5524
* Ideally all this would somehow be available in the main reader mode easily without having to trigger a mostly hidden special mode.
  I'm not sure page flipping mode still has much purpose with the greatly improved skim widget.

Frenzie:
Copying from https://github.com/koreader/koreader/pull/6408#issuecomment-661877691

We actually already have a zoom implemented in page flipping mode, although I just noticed something about the double tap there might be broken. https://github.com/koreader/koreader/pull/6322 is a potential suspect but it'd need to be investigated. Also see https://github.com/koreader/koreader/issues/5524

If you run KOReader on a multitouch device then you can also enter page flipping mode and zoom in/out with pinch gestures. As you'll see it's more of a guided zoom, not truly free.

https://github.com/koreader/koreader/blob/86db43a0527e94f1ebe2a124dae92f8fd3f70e6e/frontend/apps/reader/modules/readerzooming.lua#L315-L331

The main obstacle is that I haven't figured out if there's a good way to combine free scroll & zoom with the ease of use of zoom to fit content/page width, and also some undefined behavior outside of the page proper.

[...]

PR #4486, opened 2019-01-16
Added choice of default action when highlighting

Galunid:
This resolves #4480. Adds `Action on highlight` subsection in `Document>` menu, that allows for using default action when user highlights multiple words.

PR #6408, opened 2020-07-21
add Zoom when hold(not finished)

msh2050:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6408)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -588,12 +588,12 @@ function ReaderHighlight:_resetHoldTimer(clear)
 end
 
 function ReaderHighlight:getRegionalZoomCenter(pos)
-    if not self.dimen then
-        self.dimen = self.ui.dimen
-    end
+    
+    local dimen = self.ui.dimen
+   
     local page_size = self.ui.document:getNativePageDimensions(pos.page)
     local margin = self.ui.document.configurable.page_margin * Screen:getDPI()
-    local zoom = 2*self.dimen.w / page_size.w
+    local zoom = 2* dimen.w / page_size.w
     zoom = zoom/(1 + 3*margin/zoom/page_size.w)
     local x_ratio = (pos.x/ page_size.w) -- x center ratio
     local y_ratio = (pos.y / page_size.h) -- y center ratio
@@ -703,8 +703,8 @@ function ReaderHighlight:onTouchZoom(_, ges)
         local magnifier = Magnifier:new{
         image = image,
         timeout = 5,
-        -- set zoom to dubble now - setting variable to be add
-        zoom = zoom * 2 ,
+        -- set zoom to double now - setting variable to be add
+        zoom = zoom * 1.2 ,
         x_ratio = x_ratio,
         y_ratio = y_ratio,
         location = location
@@ -715,6 +715,9 @@ function ReaderHighlight:onTouchZoom(_, ges)
 end
 
 function ReaderHighlight:onHoldPan(_, ges)
+    if self.view.highlight.disabled then
+        return true
+    end
     if self.hold_pos == nil then
         logger.dbg("no previous hold position")
         self:_resetHoldTimer(true)
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -15,7 +15,6 @@ local _ = require("gettext")
 local C_ = _.pgettext
 local T = require("ffi/util").template
 local Screen = Device.screen
-
 local ReaderHighlight = InputContainer:new{}
 
 function ReaderHighlight:init()
@@ -127,24 +126,24 @@ function ReaderHighlight:genHighlightDrawerMenu()
             end,
             callback = function()
                 self.view.highlight.disabled = not self.view.highlight.disabled
-                self.view.highlight.zoom =  self.view.highlight.disabled
+                self.view.highlight.magnifier =  self.view.highlight.disabled
             end,
             hold_callback = function(touchmenu_instance)
                 self:toggleDefault()
             end,
         },
         {
-            text = _("Zoom Touched"),
+            text = _("use magnifier (beta)"),
             checked_func = function()
-                return self.view.highlight.zoom
+                return self.view.highlight.magnifier
             end,
             callback = function()
 
-                self.view.highlight.zoom = not self.view.highlight.zoom
-                self.view.highlight.disabled =  self.view.highlight.zoom
+                self.view.highlight.magnifier = not self.view.highlight.magnifier
+                self.view.highlight.disabled =  self.view.highlight.magnifier
             end,
             hold_callback = function(touchmenu_instance)
-                self:toggleDefault()
+                self:magnifiertoggleDefault()
             end,
             separator = true,
         },
@@ -586,58 +585,17 @@ function ReaderHighlight:getRegionalZoomCenter(pos)
     zoom = zoom/(1 + 3*margin/zoom/page_size.w)
     local x_ratio = (pos.x/ page_size.w) -- x center ratio
     local y_ratio = (pos.y / page_size.h) -- y center ratio
-       
     return zoom, x_ratio, y_ratio
-    
 end
 
 
-function ReaderHighlight:onHold(arg, ges)
+function ReaderHighlight:onHold(_, ges)
     -- disable hold gesture if highlighting is disabled
     if self.view.highlight.disabled then
-        if self.view.highlight.zoom then
-            --zoom the hold area
-            self:clear() -- clear previous highlight (delayed clear may not have done it yet)
-            self.hold_ges_pos = ges.pos -- remember hold original gesture position
-            self.hold_pos = self.view:screenToPageTransform(ges.pos)
-            logger.dbg("hold position in page ", self.hold_pos)
-            if not self.hold_pos then
-                logger.dbg("not inside page area")
-                return false
-            end
-            
-            -- Screen:shot("/home/m/1.png")
-            logger.dbg("hold on image")
-
-            local getImage = require("document/koptinterface")
-            local Image = getImage:getNativePageImage(self.ui.document, self.hold_pos.page)
-            if Image then
-                logger.dbg("we got image")
-                local zoom, x_ratio, y_ratio
-                zoom, x_ratio, y_ratio = self:getRegionalZoomCenter(self.hold_pos)
-                logger.dbg("zoom:", zoom)
-                logger.dbg("xpos:", x_ratio)
-                logger.dbg("ypos:", y_ratio)
-                
-                --call magnifier widget
-                local Magnifier = require("ui/widget/magnifier")
-                local magnifier = Magnifier:new{
-                image = Image,
-                timeout = 5,
-                -- set zoom to dubble now - setting variable to be add
-                zoom = zoom * 2 ,
-                x_ratio = x_ratio,
-                y_ratio = y_ratio
-
-                }
-            UIManager:show(magnifier)
-            end
-            
-            return false
-
+        if self.view.highlight.magnifier then
+           self:onTouchZoom(nil, ges) 
         end
         return false
-
     end
     self:clear() -- clear previous highlight (delayed clear may not have done it yet)
     self.hold_ges_pos = ges.pos -- remember hold original gesture position
@@ -696,6 +654,52 @@ function ReaderHighlight:onHold(arg, ges)
         return true
     end
     return false
+end
+
+--display zoomed image 
+function ReaderHighlight:onTouchZoom(_, ges)
+    self:clear() -- clear previous highlight (delayed clear may not have done it yet)
+    self.hold_ges_pos = ges.pos -- remember hold original gesture position
+    self.hold_pos = self.view:screenToPageTransform(ges.pos)
+    logger.dbg("hold position in page ", self.hold_pos)
+    if not self.hold_pos then
+        logger.dbg("not inside page area")
+        return false
+    end
+    local image
+    if self.ui.document.koptinterface ~= nil then
+        image = self.ui.document.koptinterface:getNativePageImage(self.ui.document, self.hold_pos.page)
+    else
+        logger.dbg("document has no koptinterface")
+    end
+    if image then
+        logger.dbg("we got image")
+        local zoom, x_ratio, y_ratio
+        zoom, x_ratio, y_ratio = self:getRegionalZoomCenter(self.hold_pos)
+        logger.dbg("zoom:", zoom)
+        logger.dbg("xpos:", x_ratio)
+        logger.dbg("ypos:", y_ratio)
+
+        --call magnifier widget
+        local Magnifier = require("ui/widget/magnifier")
+        --set the magnifier location
+        local location = ""
+        if ges.pos.y > (Screen:getHeight() / 2) then
+            location = "top"
+            -- if not top will be bottom by default
+        end
+        local magnifier = Magnifier:new{
+        image = image,
+        timeout = 5,
+        -- set zoom to dubble now - setting variable to be add
+        zoom = zoom * 2 ,
+        x_ratio = x_ratio,
+        y_ratio = y_ratio,
+        location = location
+        }
+    UIManager:show(magnifier)
+    end
+    return true
 end
 
 function ReaderHighlight:onHoldPan(_, ges)
@@ -1356,21 +1360,22 @@ end
 function ReaderHighlight:onReadSettings(config)
     self.view.highlight.saved_drawer = config:readSetting("highlight_drawer") or self.view.highlight.saved_drawer
     local disable_highlight = config:readSetting("highlight_disabled")
+    local disable_magnifier = config:readSetting("disable_magnifier")
     if disable_highlight == nil then
         disable_highlight = G_reader_settings:readSetting("highlight_disabled") or false
     end
-    self.view.highlight.disabled = disable_highlight
-    local zoom = config:readSetting("hold_to_zoom")
-    if zoom == nil then
-        zoom = G_reader_settings:readSetting("hold_to_zoom") or false
+   
+    if disable_magnifier == nil then
+        disable_magnifier = G_reader_settings:readSetting("disable_magnifier") or false
     end
-    self.view.highlight.zoom = zoom
+    self.view.highlight.disabled = disable_highlight
+    self.view.highlight.magnifier = disable_magnifier
 end
 
 function ReaderHighlight:onSaveSettings()
     self.ui.doc_settings:saveSetting("highlight_drawer", self.view.highlight.saved_drawer)
     self.ui.doc_settings:saveSetting("highlight_disabled", self.view.highlight.disabled)
-    self.ui.doc_settings:saveSetting("hold_to_zoom", self.view.highlight.zoom)
+    self.ui.doc_settings:saveSetting("disable_magnifier", self.view.highlight.magnifier)
 end
 
 function ReaderHighlight:onClose()
@@ -1399,4 +1404,23 @@ function ReaderHighlight:toggleDefault()
     })
 end
 
+function ReaderHighlight:magnifiertoggleDefault()
+    local disable_magnifier = G_reader_settings:isTrue("disable_magnifier")
+    UIManager:show(MultiConfirmBox:new{
+        text = disable_magnifier and _("Would you like to enable or disable magnifier on hold by default?\n\nThe current default (★) is enabled.")
+        or _("Would you like to enable or disable magnifier on hold by default?\n\nThe current default (★) is disabled."),
+        choice1_text_func =  function()
+            return disable_magnifier and _("Disable (★)") or _("Disable")
+        end,
+        choice1_callback = function()
+            G_reader_settings:saveSetting("disable_magnifier", true)
+        end,
+        choice2_text_func = function()
+            return highlight_disabled and _("Enable") or _("Enable (★)")
+        end,
+        choice2_callback = function()
+            G_reader_settings:saveSetting("disable_magnifier", false)
+        end,
+    })
+end
 return ReaderHighlight
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -139,7 +139,7 @@ function ReaderHighlight:genHighlightDrawerMenu()
                 return self.view.highlight.zoom
             end,
             callback = function()
-                
+
                 self.view.highlight.zoom = not self.view.highlight.zoom
                 self.view.highlight.disabled =  self.view.highlight.zoom
             end,
@@ -581,13 +581,13 @@ end
 
 function ReaderHighlight:onHold(arg, ges)
     -- disable hold gesture if highlighting is disabled
-    if self.view.highlight.disabled then 
+    if self.view.highlight.disabled then
         if self.view.highlight.zoom then
             --zoom the hold area
             self:clear() -- clear previous highlight (delayed clear may not have done it yet)
             self.hold_ges_pos = ges.pos -- remember hold original gesture position
             self.hold_pos = self.view:screenToPageTransform(ges.pos)
-            logger.dbg("hold position in page", self.hold_pos)
+            logger.dbg("hold position in page msh", self.hold_pos)
             if not self.hold_pos then
                 logger.dbg("not inside page area")
                 return false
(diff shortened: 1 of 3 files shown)


msh2050:
this pull linked to this  [issue](https://github.com/koreader/koreader/issues/6406)

poire-z:
From https://github.com/koreader/koreader/issues/6406#issuecomment-661576800 :
> OK it is good starting I faced many issues but I get what i want
first:- the self.ui.document:getImageFromPosition(self.hold_pos, true) will only work with epub (not cbz)
so I used screenshoter as a work around ..
if I get help on how to get the image from the diplayed page it will be better
also if I have to go with screenshoter it will be better to shot to buffer rather than save and load from file..
finally I add option on the menu to that will disable highlight if the zoom activated
it need some tweaks like specifying the image widget dimensions and add the save option for the menu and so on... butthis will be done after fix the above mentioned issues..
regards

We have this function to get the native first page image:
https://github.com/koreader/koreader/blob/fca7f9e7d56820c5d9ea573d87ed42db7fbefd71/frontend/document/koptinterface.lua#L279-L290

I guess you could duplicate it as `getNativePageImage(page_number)` to get the image for any page, not just the first (I guess one of the `1` in there tells it we want page 1 :)
And fix/implement wrappers/simplify these to use that generic new one:
```
frontend/document/pdfdocument.lua:function PdfDocument:getCoverPageImage()
frontend/document/pdfdocument.lua:    return self.koptinterface:getCoverPageImage(self)
frontend/document/credocument.lua:function CreDocument:getCoverPageImage()
frontend/document/credocument.lua:    local data, size = self._document:getCoverPageImageData()
frontend/document/credocument.lua:            elseif name == "getCoverPageImage" then no_wrap = true
frontend/document/document.lua:function Document:getCoverPageImage()
frontend/document/koptinterface.lua:function KoptInterface:getCoverPageImage(doc)
frontend/document/picdocument.lua:function PicDocument:getCoverPageImage()
frontend/document/djvudocument.lua:function DjvuDocument:getCoverPageImage()
frontend/document/djvudocument.lua:    return self.koptinterface:getCoverPageImage(self)
```

> it need some tweaks like specifying the image widget dimensions 

I guess you already looked at it, but you have these options available:
https://github.com/koreader/koreader/blob/fca7f9e7d56820c5d9ea573d87ed42db7fbefd71/frontend/ui/widget/imageviewer.lua#L30-L75

Frenzie:
What's in this PR is a good attempt, but I think we can do much better. ;-) I don't necessarily mean that in a perfect is the enemy of good kind of way, but we can already get a nice high-res picture of a region. There's no need to get a blurry one.

We actually already have a zoom implemented in page flipping mode, although I just noticed something about the double tap there might be broken. https://github.com/koreader/koreader/pull/6322 is a potential suspect but it'd need to be investigated. Also see https://github.com/koreader/koreader/issues/5524

If you run KOReader on a multitouch device then you can also enter page flipping mode and zoom in/out with pinch gestures. As you'll see it's more of a guided zoom, not truly free.

https://github.com/koreader/koreader/blob/86db43a0527e94f1ebe2a124dae92f8fd3f70e6e/frontend/apps/reader/modules/readerzooming.lua#L315-L331

The main obstacle is that I haven't figured out if there's a good way to combine free scroll & zoom with the ease of use of zoom to fit content/page width, and also some undefined behavior outside of the page proper.

But most or perhaps all of the puzzle pieces are already in place. The guided zoom to text columns is a great concept — but only when it works nicely.

Anyway, it's one of those help most welcome kind of things. ^_^

msh2050:
yes... agreed ... performance and quality of zoom is critical here...
for manga/comic reading it is good to have total view of the scenario with a click to get what the character is saying ....
I think it is better to use direct functions from mupdf/etc to process the images ...

regards

msh2050:
@poire-z @Frenzie thanks for your help and guidance...
I add a **KoptInterface:getNativePageImage(doc , pageno)**  function it is working now - but it still dirty hack :) 
please note that `self.ui.document:getCoverPageImage()` will give null reference for (self) so I implement  **KoptInterface:getNativePageImage** directly..

@Frenzie I tried to use flipping mode but it has some bugs (mainly the slow navigation) and the main issue is  the losing of  the full  view ... and the code you mention is directed to zoom all the screen 
still the getRegionalZoomCenter function and zooming function is a must use to get the right results in most other zooming and display configurations.

The issue is that imageviewer widget will cover most of the screen.. and the x,y parameter will work for the image only (the frame will continue to cover the screen) 
also I think the imageviewer is too much for this task we do not need all the loaded features ..

I will try to use the ImageWidget directly but I got error `./luajit: frontend/ui/widget/imageviewer.lua:418: attempt to index field 'dimen' (a nil value)`  because it need a container and so on .. 

I'm thinking in modifying the imageviewer or use other container to be flashed in the bottom if the click in the upper half of the screen or otherwise in the upper ..  it will be like magnifier...

finally I'm happy to see other    [request](https://github.com/koreader/koreader-base/issues/1148) to solve the view for comic and I think double click is faster than hold I will try to check that.

regards

Frenzie:
> I tried to use flipping mode but it has some bugs (mainly the slow navigation)

I'm not sure what you're referring to, unless…

> I think double click is faster than hold I will try to check that.

Double tap is slow. It introduces delay for single tap.

msh2050:
I need help to know why I'm getting this error


any widget other than image viewer will show same error
I even copy this code from same file that I'm editing and it is not working laso `UIManager:show(InfoMessage:new{
            text = _("gust to test if i can run container"),
        })`

poire-z:
May be because you create `local imgviewer ` and `local imagewedget ` and you never `UIManager:show()` them (their code may register into UIManager in their :init() and expect them to be shown).
Comment them all out, and it should work.

msh2050:
Waw ... That solve my issue...
Hope I can sleep now without thinking in continue coding😂

msh2050:
I think It is mature now.... the zooming and  centralization are perfect...
I copied the infomessage widget to make new magnifier widget..
now  the widget :-
1- will show on the top or bottom according to where you hold the finger.
2- will flash for 5 sec or till tab on screen
3- will work with different page zoom option 
4- will work with continuous view mode  option 

I will keep the below image size 

please advise if I need to correct any think..or go a head to build for my kindle..

what I'm thinking in:-
1-  change zoom choice in menu to (magnifier) 
2- add a setting to configure magnification zoom (it is hard-coded to double now) 

week point :-
- the image will be loaded in each hold.(it is better to be loaded once when page viewed and used with each hold)
- currently highlighting and zoom can not be both disabled (enabling on will disable the other and disabling one will enable the other) ... need to be corrected

regard



Screenshot from 2020-07-23 22-25-14
 

Screenshot from 2020-07-23 22-25-00

Screenshot from 2020-07-23 22-23-40

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -127,6 +127,21 @@ function ReaderHighlight:genHighlightDrawerMenu()
             end,
             callback = function()
                 self.view.highlight.disabled = not self.view.highlight.disabled
+                self.view.highlight.zoom =  self.view.highlight.disabled
```
poire-z, line 130: Dunno if we people should be allowed to disable both ? (for lazy people that let their fingers rest on the screen :)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -127,6 +127,21 @@ function ReaderHighlight:genHighlightDrawerMenu()
...
+            end,
+            hold_callback = function(touchmenu_instance)
+                self:toggleDefault()
```
poire-z, line 133: The new menu item triggers the same `self:toggleDefault()` - should probably be another one, or adapted to decide which of highlight or zoom (or none) should be the default.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -561,9 +576,69 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+    end
+    local page_size = self.ui.document:getNativePageDimensions(pos.page)
+    local margin = self.ui.document.configurable.page_margin * Screen:getDPI()
+    local zoom = 2*self.dimen.w / page_size.w
+    zoom = zoom/(1 + 3*margin/zoom/page_size.w)
+    local x_ratio = (pos.x/ page_size.w) -- x center ratio
+    local y_ratio = (pos.y / page_size.h) -- y center ratio
+       
+    return zoom, x_ratio, y_ratio
+
```
poire-z, line 591: These kind of blank lines around the return are not really - they just make the reader scroll more :)
(I may use one to separate different logical works in a same function.)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -561,9 +576,69 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+    
+end
+
+
 function ReaderHighlight:onHold(arg, ges)
     -- disable hold gesture if highlighting is disabled
-    if self.view.highlight.disabled then return false end
+    if self.view.highlight.disabled then
+        if self.view.highlight.zoom then
+            --zoom the hold area
```
poire-z, line 599: You might want to externalize that in another `ReaderHighlight:onTouchZoom()` method - to keep that onHold() shorter and logically handling highlights - and we know the zoom stuff is another thing.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -561,9 +576,69 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+            logger.dbg("hold position in page ", self.hold_pos)
+            if not self.hold_pos then
+                logger.dbg("not inside page area")
+                return false
+            end
+            
+            -- Screen:shot("/home/m/1.png")
+            logger.dbg("hold on image")
+
+            local getImage = require("document/koptinterface")
```
poire-z, line 612: Keep giving it the local name we give it everywhere please :)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -561,9 +576,69 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+            local Image = getImage:getNativePageImage(self.ui.document, self.hold_pos.page)
```
poire-z, line 613: No uppercase for local standard variables, just name it `image` or `img` - or we could believe it's a module.
(What you do below for Magnifier/magnifier is the right way to do it.)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -561,9 +576,69 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+                -- set zoom to dubble now - setting variable to be add
+                zoom = zoom * 2 ,
+                x_ratio = x_ratio,
+                y_ratio = y_ratio
+
+                }
+            UIManager:show(magnifier)
+            end
+            
+            return false
```
poire-z, line 636: You handled something: you return true, so the event does not propagate to other lower widgets.
When you don't and want to let it propagate as some lower widget might handle it: you return false.

frontend/document/koptinterface.lua
```diff
@@ -289,6 +289,20 @@ function KoptInterface:getCoverPageImage(doc)
     end
 end
 
+function KoptInterface:getNativePageImage(doc , pageno)
+    --[[cancel zoom to get better image quality for zooming
+
+    local native_size = Document.getNativePageDimensions(doc, pageno)
+    local canvas_size = CanvasContext:getSize()
+    local zoom = math.min(canvas_size.w / native_size.w, canvas_size.h / native_size.h)
+    ]]
```
poire-z, line 298: Might replace that with a comment `-- Unlike in getCoverPageImage(), we cancel zoom to get better image quality for zooming` - although written like that, I don't understand it very well :/ "cancell zoom to get better zoom" ? :)

frontend/ui/widget/magnifier.lua
```diff
@@ -0,0 +1,156 @@
...
+    local Screen = require("device").screen
+    local magnifier
+    magnifier = Magnifier:new{
+        image = Image,
+        height = Screen:scaleBySize(400),
+        width = Screen:scaleBySize(400),
+        timeout = 5,  -- This widget will vanish in 5 seconds.
+    }
+    UIManager:show(magnifier)
+    magnifier:onShowKeyboard()
```
poire-z, line 18: Either remove the example, or have it simple and correct (no gettext, no keyboard needed :)

frontend/ui/widget/magnifier.lua
```diff
@@ -0,0 +1,156 @@
...
+]]
+
+local Blitbuffer = require("ffi/blitbuffer")
+local BottomContainer = require("ui/widget/container/bottomcontainer")
+local Topcontainer = require("ui/widget/container/topcontainer")
+local Device = require("device")
+local FrameContainer = require("ui/widget/container/framecontainer")
```
poire-z, line 25: Re-order them Alphabetically, following the exceptions you might see in other modules.

frontend/ui/widget/magnifier.lua
```diff
@@ -0,0 +1,156 @@
...
+                    w = Screen:getWidth(),
+                    h = Screen:getHeight(),
+                }
+            }
+        }
+    end
+   
+    
+    local image_widget
+    image_widget = ImageWidget:new{
```
poire-z, line 86: On a single line is fine (unless you need to reference the object itself in one of its callback or other function set as a property of it).

frontend/ui/widget/magnifier.lua
```diff
@@ -0,0 +1,156 @@
...
+        image = self.image,
+        width = Screen:getWidth(),
+        height = Screen:scaleBySize(200),
```
poire-z, line 89: What about japanese comics with vertical text ? :) (dunno if there are any, or they all get horizontal text)

poire-z:
I don't have much opinion about the feature and interactions - as I rarely read PDF and never comics. 
Dunno either if this PR duplicate/overlap with #5524 - or if this PR solves it.

Just some thoughts:
- do you really need a timeout ? people might want to have it stay as a reminder of where they were, or to lookup a word in a paper dictionary (because you can't invoke dict from there :)
- another nice thing could be to require the user to keep long-pressing and dismiss on finger release (but dunno if that way, the finger might be in the way of reading) - and then, one could just pan his finger to have the picture centered on the new finger position - so you can keep reading taller captions with just moving your finger down.

frontend/apps/reader/modules/readerhighlight.lua
yparitcher, line 612: this include might break crengine document (epubs) and is unnecessary for mupdf based document as it is already loaded.
you can check for this and use the preloaded one, see:
https://github.com/koreader/koreader/blob/7c1a716d57b092bb786eeec3ca9ecb5405a8195d/frontend/apps/reader/modules/readerconfig.lua#L16-L18

with something like: (untested)
```lua
if self.ui.document.koptinterface ~= nil then
    local Image = self.ui.document.koptinterface:getNativePageImage(self.ui.document, self.hold_pos.page)
```

yparitcher:
I think it might be better (less confusing) to add this to the highlight popup menu rather than having it a separate behavior when highlighting is disabled

poire-z:
> I think it might be better (less confusing) to add this to the highlight popup menu rather than having it a separate behavior when highlighting is disabled

I think it's the separate `Highlighting > [ ] Allow highlighting` that is confusing - because if you don't allow highlighting (and dont care about the style of highlights), currently, you can't select text and do dictionary/wikipedia/translate/share... any on the actions in that other menu far away from this one.

yparitcher:
> I think it's the separate `Highlighting > [ ] Allow highlighting` that is confusing - because if you don't allow highlighting (and dont care about the style of highlights), currently, you can't select text and do dictionary/wikipedia/translate/share... any on the actions in that other menu far away from this one.

correct, really the whole menu should be clarified. I was just saying that this should not be another exception making it even more confusing :)

frontend/apps/reader/modules/readerhighlight.lua
msh2050: I Dunno ..so I will leave it as it is.

frontend/apps/reader/modules/readerhighlight.lua
msh2050: Done

frontend/apps/reader/modules/readerhighlight.lua
msh2050: done.
I need to change my habit :)

frontend/apps/reader/modules/readerhighlight.lua
msh2050: I change it and it worked but what should I do it koptinterface is nill...
should I add:- 
```
else
local koptinterface = require("document/koptinterface")
local image = getImage:getNativePageImage(self.ui.document, self.hold_pos.page)
```


sorry I'm new to the app and learning curve need time to let me know how thinks work..

frontend/apps/reader/modules/readerhighlight.lua
msh2050: done

frontend/apps/reader/modules/readerhighlight.lua
msh2050: Done ...

frontend/apps/reader/modules/readerhighlight.lua
msh2050: I used the code suggested by @yparitcher

frontend/document/koptinterface.lua
msh2050: sorry for englisg :)
corrected

frontend/ui/widget/magnifier.lua
msh2050: Done

frontend/ui/widget/magnifier.lua
msh2050: Done

frontend/ui/widget/magnifier.lua
msh2050: -done

frontend/ui/widget/magnifier.lua
msh2050: yes it may need more tweaks.
I may be the next version will be a square box near the finger and updated with finger move as you suggest...

msh2050:
> > I think it's the separate `Highlighting > [ ] Allow highlighting` that is confusing - because if you don't allow highlighting (and dont care about the style of highlights), currently, you can't select text and do dictionary/wikipedia/translate/share... any on the actions in that other menu far away from this one.
> 
> correct, really the whole menu should be clarified. I was just saying that this should not be another exception making it even more confusing :)

so do you suggest hold text --> show highlight popup menu ---> select zooming ?
I think this will make it very slow in reading 
may be it is better to add toggle for faster moving between zooming and highlight.

frontend/apps/reader/modules/readerhighlight.lua
msh2050: done

poire-z:
(You forgot to push your changes.)

> so do you suggest hold text --> show highlight popup menu ---> select zooming ?

I think @yparitcher means moving your new menu into or around `Gear > Document > Highlight action >` . Where it could benefit from a gesture that cycle thru these actions.
But it's a different kind of action: the others apply on text, yours not - and yours has no effect on EPUB documents, only PDF - while the others work everywhere.
So, no real idea how that should look & work :/ pinging @Galunid would did the action stuff, if you have some ideas.

frontend/apps/reader/modules/readerhighlight.lua
poire-z: > should I add...

No. The documents that work with kopt have it (pdfdocument, djvudocument) - those who don't do not have it (picdocument). Using it yourself would just crash on picdocument that can't.

Also, pdfdocument & co have some wrapping functions that use it internally:
https://github.com/koreader/koreader/blob/db843929f0893383a3dcddbb6d77708691b8e393/frontend/document/pdfdocument.lua#L124-L126
that overrides generic functions that return nil when not available:
https://github.com/koreader/koreader/blob/db843929f0893383a3dcddbb6d77708691b8e393/frontend/document/document.lua#L286-L288

So, ideally, higher level code like readerhighlight should not need to know the document works with koptinterface or not: ask, if you get nil, it doesn't work.

msh2050:
I will build this for my kindle

msh2050:
I try to build release for pw2 but it is not starting in my kindle before that the build was stopping due to DROPBEAR error I add the below code to the file and the build completed but not working
```
I solve the issue by writing :
    #if !HAVE_CRYPT
    #define DROPBEAR_SVR_PASSWORD_AUTH 0
    #endif

```
so I just download the final release and changed the 3 files in this PR...
It is working OK now

I just noticed that the zooming is not correct (the centralization) if reflow enabled
and may be I need to decrease zooming power ...

regards

poire-z:
> so I just download the final release and changed the 3 files in this PR

Oh, yes, you didn't need to compile anything. Just copying the update .lua files on the device is enough :)
(I sometimes just edit files directly on the device for testing little changes.)

Galunid:
> I think @yparitcher means moving your new menu into or around `Gear > Document > Highlight action >` . Where it could benefit from a gesture that cycle thru these actions.
> But it's a different kind of action: the others apply on text, yours not - and yours has no effect on EPUB documents, only PDF - while the others work everywhere.
> So, no real idea how that should look & work :/ pinging @Galunid who did the action stuff, if you have some ideas.

It was done in #4486. Basically it reads 
```lua
local default_highlight_action = G_reader_settings:readSetting("default_highlight_action")
```
And then checks which one is set(`highlight`, `translate`, `wikipedia`). if `default_highlight_action` is not set, it defaults to showing popup menu.
While it should be possible to add this to `Gear > Document > Highlight action >`, I was thinking about setting it as default for `.cbz` considering it's all images and "text detection" doesn't really work here. As for pdf's I'm not sure. So far, all the `Highlight actions` are applicable to all documents. Perhaps you may add another setting like `default_pdf_highlight_action` that overwrites `default_highlight_action` in pdf/cbz?

Galunid:
I didn't read the code/test this yet, I'll try to do so today, and perhaps a better solution will come up.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -573,9 +587,28 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+    return zoom, x_ratio, y_ratio
+end
+
+
+function ReaderHighlight:onHold(_, ges)
     -- disable hold gesture if highlighting is disabled
-    if self.view.highlight.disabled then return false end
+    if self.view.highlight.disabled then
+        if self.view.highlight.magnifier then
+           self:onTouchZoom(nil, ges)
```
Galunid, line 608: This could probably be rewritten as 
```lua
UIManager:broadcastEvent(Event:new("TouchZoom", ges))
```

poire-z: No need to, when the single and only onTouchZoom handler is in here.

Galunid: I was thinking that @yparitcher's dispatcher can use this as an event, either is fine imo.

poire-z: Yes, he can - and should be able to just by having it named here `onTouchZoom` - use that via broadcastEvent elsewhere - but here, we should call it directly `self:onTouchZoom()` for clarity.
(Although given that this functions needs a gesture with x/y, it's probably not a candidate for profiles or multiswipe gestures.)

Frenzie: Multiswipes work with coordinates. No problem there.

Galunid:
`ReaderHighlight:onHoldPan` should have additional check to see if highlights are enabled
```diff
diff --git a/frontend/apps/reader/modules/readerhighlight.lua b/frontend/apps/reader/modules/readerhighlight.lua
index 7da9e281..5e764318 100644
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -715,6 +715,9 @@ function ReaderHighlight:onTouchZoom(_, ges)
 end

 function ReaderHighlight:onHoldPan(_, ges)
+    if self.view.highlight.disabled then
+        return true
+    end
     if self.hold_pos == nil then
         logger.dbg("no previous hold position")
         self:_resetHoldTimer(true)

```
seems to do the trick. Alternatively one can add another condition to the `if self.hold_pos == nil then`. Without it, once the magnifier disappears(by timeout, not tap), one can highlight text, which is probably not what the user wants ;)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -573,9 +587,28 @@ function ReaderHighlight:_resetHoldTimer(clear)
     end
 end
 
-function ReaderHighlight:onHold(arg, ges)
+function ReaderHighlight:getRegionalZoomCenter(pos)
+    if not self.dimen then
+        self.dimen = self.ui.dimen
+    end
```
Galunid, line 593: Since `self.dimen` is only used in this function, it doesn't need to be object's property. You could replace this with
`local dimen = self.ui.dimen`

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -635,6 +668,52 @@ function ReaderHighlight:onHold(arg, ges)
...
+        --set the magnifier location
+        local location = ""
+        if ges.pos.y > (Screen:getHeight() / 2) then
+            location = "top"
+            -- if not top will be bottom by default
+        end
+        local magnifier = Magnifier:new{
+        image = image,
+        timeout = 5,
+        -- set zoom to dubble now - setting variable to be add
```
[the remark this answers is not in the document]
msh2050: yes , thanks

frontend/apps/reader/modules/readerhighlight.lua
msh2050: done

msh2050:
after testing this for some days ...
it is not that much practical to hold the finger many times in one page... 
page flip will be a better way to go ...
I'm closing this pull request till a new idea coming up ...

regards and thanks to all for your help and time ...

Frenzie:
I hope you had some fun playing around with the code in any case. :-)

msh2050:
> I hope you had some fun playing around with the code in any case. :-)

yes, of course, thanks all for your help.
I will try to see if I can add google search(some times ) it gives better information than Wikipedia ...
and I will see if I cand change the article language(you search in English term --> get the English article ---> check other languages available for this article -->   and so on)

the OCR needs to be enhanced also. Dunno if it is easy updating the k2pdfopt to the latest version that has much enhance buildin OCR features.

NiLuJe:
Updating k2pdfopt is a... daunting task. c.f., https://github.com/koreader/koreader-base/pull/762

PR #6417, opened 2020-07-22
Fake battery status in readerfooter_spec to succeed on devices with batteries

Frenzie:
Cf. <https://github.com/koreader/koreader/pull/6370#issuecomment-662382487>.

@pazos No spies/stubs necessary here. ^_^

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6417)

### DIFF
--- a/frontend/device/sdl/powerd.lua
+++ b/frontend/device/sdl/powerd.lua
@@ -5,7 +5,7 @@ local SDLPowerD = BasePowerD:new{}
 
 function SDLPowerD:getCapacityHW()
     local _, _, _, percent = SDL.getPowerInfo()
-    -- never return negative values, since tests rely on battery being 0%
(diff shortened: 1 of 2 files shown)


pazos:
Oh! It seems that we don't need the workaround in https://github.com/koreader/koreader/blob/master/frontend/device/sdl/powerd.lua#L8-L9

Frenzie:
@pazos Oh, -1 is what SDL returns on desktops? It was something like 60% on my laptop when I looked at this.

Frenzie:
@pazos I don't know, doesn't that look a bit weird? 
Screenshot_2020-07-22_18-02-05

pazos:
>  doesn't that look a bit weird?

Yeah, just a bit :p. I really don't care about -1% vs 0%, but the comment in the code is no longer true.

PR #6419, opened 2020-07-22
Fix a few logic holes in dithering mode selection

NiLuJe:
* Allow switching to SW dithering on a HW-capable device without that being lost on boot (and, worse, left in an undithered state).
* Make sure the automagic toggles between HW/SW in the Dev menu are properly saved.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6419)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -357,7 +357,7 @@ function FileManagerMenu:setUpdateItemTable()
             end,
             callback = function()
                 G_reader_settings:flipNilOrFalse("dev_no_hw_dither")
-                Device.screen:toggleHWDithering()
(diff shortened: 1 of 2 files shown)


NiLuJe:
Depends on https://github.com/koreader/koreader-base/pull/1150

Fakerr:
Curious here. Since this PR depends on another one, would you be interested by this [Github app](https://www.dpulls.com/) that I built that prevents PRs with "Depends on" from being merged before their dependencies ? Do you think it's something that can help you in your projects ? cc @NiLuJe

NiLuJe:
@Fakerr: I can't recall anyone *ever* fat-fingering such a PR, FWIW.

Then again, we're somewhat extensively using submodules that we control, so we're kind of used to that kind of workflow.

PR #3395, opened 2017-10-22
News downloader: wifi off prompt after download

mwoz123:
Prompt to turn off wifi after download.
Extracted from #3388

PR #6424, opened 2020-07-24
Various Wi-Fi QoL improvements

### LINKED ISSUE #6421: UI freezes frequently, seemingly correlated with network usage
* KOReader version:2020.07.1
* Device: Kobo Libra H2O

#### Issue
Lately (since updating to 2020.07, I think) I've noticed the KOReader UI become frequently unresponsive during active use of the application.

#### Steps to reproduce
With continued observation over time I think it coincides with actions that use the network, such as manually pushing reading progress to cloud, going from an open book back to the file list, locking the screen via the power button, checking for an application update, etc. More basic functionality like turning pages and navigating the book work as expected. Opening a book happens quickly, however the application then becomes unresponsive to actions such as turning pages. After a period of around 15 seconds, screen taps and button presses all happen quickly, as if they had been queued.

The ssh server does not appear to be affected; it enables/disables/can be accessed instantly.

[lag-log.txt](https://github.com/koreader/koreader/files/4967981/lag-log.txt)

NiLuJe:
* Revamped most actions that require an internet connection to a new/fixed backend that allows forwarding the initial action and running it automatically once connected. (i.e., it'll allow you to set "Action when Wi-Fi is off" to "turn_on", and whatch stuff connect and do what you wanted automatically without having to re-click anywhere instead of showing you a Wi-Fi prompt and then not doing anything without any other feedback).
* Speaking of, fixed the "turn_on" beforeWifi action to, well, actually work. It's no longer marked as experimental.
* Consistently use "Wi-Fi" everywhere.
* On Kobo/Cervantes/Sony, implemented a "Kill Wi-Fi connection when inactive" system that will automatically disconnect from Wi-Fi after sustained *network* inactivity (i.e., you can keep reading, it'll eventually turn off on its own). This should be smart and flexible enough not to murder Wi-Fi while you need it, while still not keeping it uselessly on and murdering your battery.
(i.e., enable that + turn Wi-Fi on when off and enjoy never having to bother about Wi-Fi ever again).
* Made sending `NetworkConnected` / `NetworkDisconnected` events consistent (they were only being sent... sometimes, which made relying on 'em somewhat problematic).
* restoreWifiAsync is now only run when really needed (i.e., we no longer stomp on an existing working connection just for the hell of it).
* We no longer attempt to kill a bogus non-existent Wi-Fi connection when going to suspend, we only do it when it's actually needed.
* Every method of enabling Wi-Fi will now properly tear down Wi-Fi on failure, instead of leaving it in an undefined state.
* Fixed an issue in the fancy crash screen on Kobo/reMarkable that could sometime lead to the log excerpt being missing.
* Worked-around a number of sneaky issues related to low-level Wi-Fi/DHCP/DNS handling on Kobo (see the lengthy comments [below](https://github.com/koreader/koreader/pull/6424#issuecomment-663881059) for details). Fix #6421 
Incidentally, this should also fix the inconsistencies experienced re: Wi-Fi behavior in Nickel when toggling between KOReader and Nickel (use NM/KFMon, and run a current FW for best results).
* For developers, this involves various cleanups around NetworkMgr and NetworkListener. Documentation is in-line, above the concerned functions.

----

Original message:

Possibly a bit heuristic-y, but seems to do the job well enough on my H2O.

Limited to `hasWifiManager` devices (i.e., Kobo/Cervantes/Sony).

Could possibly work on a few other pure-Linux devices with a sane `turnOffWifi`? (Kindle? [EDIT: Nope, see [below](https://github.com/koreader/koreader/pull/6424#issuecomment-663795608)])
(This would require implementing `NetworkMgr:getNetworkInterfaceName()` there, too).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6424)

### DIFF
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -384,15 +384,10 @@ function ReaderWikipedia:onLookupWikipedia(word, box, get_fullpage, forced_lang)
 end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
-    if not NetworkMgr:isOnline() then
-        --- @note: Avoid infinite recursion, beforeWifiAction only guarantees isConnected, not isOnline.
-        if not NetworkMgr:isConnected() then
-            NetworkMgr:beforeWifiAction(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
-        else
-            NetworkMgr:beforeWifiAction()
-        end
+    if not NetworkMgr:rerunWhenOnline(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end) then
         return
     end
+
     -- word is the text to query. If get_fullpage is true, it is the
     -- exact wikipedia page title we want the full page of.
     self:initLanguages(word)
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -529,16 +529,7 @@ function ReaderWikipedia:onShowWikipediaLookup()
     local connect_callback = function()
         self:lookupInput()
     end
-    if NetworkMgr:isOnline() then
-        connect_callback()
-    else
-        --- @note: Avoid infinite recursion, beforeWifiAction only guarantees isConnected, not isOnline.
-        if not NetworkMgr:isConnected() then
-            NetworkMgr:beforeWifiAction(connect_callback)
-        else
-            NetworkMgr:beforeWifiAction()
-        end
-    end
+    NetworkMgr:runWhenOnline(connect_callback)
     return true
 end
 
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -385,7 +385,7 @@ end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
     if not NetworkMgr:isOnline() then
-        --- @note: Wanted: a way to actually forward a working callback for this specific function call to beforeAction!
+        --- @note: Wanted: a way to actually forward a working callback for this specific function call to beforeWifiAction!
         NetworkMgr:beforeWifiAction()
         return
     end
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -385,7 +385,7 @@ end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
     if not NetworkMgr:isOnline() then
-        NetworkMgr:promptWifiOn()
+        NetworkMgr:beforeWifiAction(function() ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
         return
     end
     -- word is the text to query. If get_fullpage is true, it is the
@@ -525,10 +525,13 @@ function ReaderWikipedia:onSaveSettings()
 end
 
 function ReaderWikipedia:onShowWikipediaLookup()
-    if NetworkMgr:isOnline() then
+    local connect_callback = function()
         self:lookupInput()
+    end
+    if NetworkMgr:isOnline() then
+        connect_callback()
     else
-        NetworkMgr:promptWifiOn()
+        NetworkMgr:beforeWifiAction(connect_callback)
     end
     return true
 end
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -385,6 +385,7 @@ end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
     if not NetworkMgr:isOnline() then
+        --- @note: Wanted: a way to actually forward a working callback for this specific function call to beforeAction!
         NetworkMgr:beforeWifiAction()
         return
     end
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -385,7 +385,12 @@ end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
     if not NetworkMgr:isOnline() then
-        NetworkMgr:beforeWifiAction(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
+        --- @note: Avoid infinite recursion, beforeWifiAction only guarantees isConnected, not isOnline.
+        if not NetworkMgr:isConnected() then
+            NetworkMgr:beforeWifiAction(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
+        else
+            NetworkMgr:beforeWifiAction()
+        end
         return
     end
     -- word is the text to query. If get_fullpage is true, it is the
@@ -531,7 +536,12 @@ function ReaderWikipedia:onShowWikipediaLookup()
     if NetworkMgr:isOnline() then
         connect_callback()
     else
-        NetworkMgr:beforeWifiAction(connect_callback)
+        --- @note: Avoid infinite recursion, beforeWifiAction only guarantees isConnected, not isOnline.
+        if not NetworkMgr:isConnected() then
+            NetworkMgr:beforeWifiAction(connect_callback)
+        else
+            NetworkMgr:beforeWifiAction()
+        end
     end
     return true
 end
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -385,7 +385,7 @@ end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
     if not NetworkMgr:isOnline() then
-        NetworkMgr:beforeWifiAction(function() ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
+        NetworkMgr:beforeWifiAction()
         return
     end
     -- word is the text to query. If get_fullpage is true, it is the
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -384,7 +384,8 @@ function ReaderWikipedia:onLookupWikipedia(word, box, get_fullpage, forced_lang)
 end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
-    if not NetworkMgr:rerunWhenOnline(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end) then
+    if NetworkMgr:willRerunWhenOnline(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end) then
+        -- Not online yet, nothing more to do here, NetworkMgr will forward the callback and run it once connected!
         return
     end
 
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -385,8 +385,7 @@ end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
     if not NetworkMgr:isOnline() then
-        --- @note: Wanted: a way to actually forward a working callback for this specific function call to beforeWifiAction!
-        NetworkMgr:beforeWifiAction()
+        NetworkMgr:beforeWifiAction(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
         return
     end
     -- word is the text to query. If get_fullpage is true, it is the
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -253,6 +253,30 @@ function NetworkMgr:setHTTPProxy(proxy)
     end
 end
 
+-- Helper functions to hide the quirks of using beforeWifiAction properly ;).
+
+-- Mild variants that are used for recursive calls at the start of a complex function call.
+-- Returns false when not online, in which case you should *abort* (i.e., return) the initial call.
+function NetworkMgr:rerunWhenOnline(callback)
+    if not NetworkMgr:isOnline() then
+        --- @note: Avoid infinite recursion, beforeWifiAction only guarantees isConnected, not isOnline.
+        if not NetworkMgr:isConnected() then
+            NetworkMgr:beforeWifiAction(callback)
+        else
+            NetworkMgr:beforeWifiAction()
+        end
+        return false
+    end
+
+    return true
+end
+
+function NetworkMgr:rerunWhenConnected()
+
+end
+
+
+
 function NetworkMgr:getWifiMenuTable()
     if Device:isAndroid() then
         return {
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -127,7 +127,7 @@ function NetworkMgr:promptWifiOff(complete_callback)
 end
 
 function NetworkMgr:turnOnWifiAndWaitForConnection(callback)
-    local info = InfoMessage:new{ text = _("Enabling Wi-Fi…") }
+    local info = InfoMessage:new{ text = _("Connecting to Wi-Fi…") }
     UIManager:show(info)
     UIManager:forceRePaint()
 
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -255,10 +255,38 @@ end
 
 -- Helper functions to hide the quirks of using beforeWifiAction properly ;).
 
+-- Run callback *now* if you're currently online (ie., isOnline),
+-- or attempt to go online and run it *ASAP* without any more user interaction.
+-- NOTE: If you're currently connected but without Internet access (i.e., isConnected and not isOnline),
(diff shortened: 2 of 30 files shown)


NiLuJe:
The noise margin is still high enough to make an idle SSH session (w/ KeepAlive) enough to keep Wi-Fi up ;).

poire-z:
I'm not super familiar with all the Wi-Fi code - as I hacked my own dirty auto-enable/disable wifi stuff, based just on explicit KOReader network actions resetting a timer (and disabling this timer when doing some long action like Wikipedia Save as EPUB) - 2mn works for me. I shared some bit of it in https://github.com/koreader/koreader/pull/3395#issuecomment-357796949 (in case it can give you some idea, but I hope it won't :)
(I always fear these kind of PRs that may overlap with my own private hacks, and would need me to rework them - but I think this PR does not, so I shall be fine :)

NiLuJe:
Yeah, I purposefully went with a low level system statistics to keep it simple ;).

NiLuJe:
@poire-z: Yeah, there's a metric shit-ton of `promptWifiOn` that could at the very least be switched to `beforeWifiAction` ;p.

NiLuJe:
Okay, I keep forgetting where the Draft button is hidden, and/or it may no longer be available once a PR is approved...

So, don't merge: i'm fixing the `turn_on` action and replacing all those hard `promptWifiOn`!

frontend/ui/network/manager.lua
```diff
@@ -105,23 +119,14 @@ function NetworkMgr:promptWifiOff(complete_callback)
...
 
 function NetworkMgr:turnOnWifiAndWaitForConnection(callback)
-    NetworkMgr:turnOnWifi()
-    local timeout = 30
-    local retry_count = 0
-    local info = InfoMessage:new{ text = T(_("Enabling Wi-Fi. Waiting for Internet connection…\nTimeout %1 seconds."), timeout)}
+    local info = InfoMessage:new{ text = _("Enabling Wi-Fi. Waiting for Internet connection…") }
     UIManager:show(info)
     UIManager:forceRePaint()
-    while not NetworkMgr:isOnline() and retry_count < timeout do
```
NiLuJe, line 101: Yes, this is swapped to an `isConnected` on purpose.

It's generally good enough, and if WAN connectivity happens to be required by the callback, it'll just timeout on its own.

poire-z:
> Okay, I keep forgetting where the Draft button is hidden, and/or it may no longer be available once a PR is approved...

It's below these great people:
<kbd>image</kbd>

frontend/ui/otamanager.lua
```diff
@@ -456,7 +456,7 @@ function OTAManager:getOTAMenuTable()
         text = _("Update"),
         hold_callback = function()
             if not NetworkMgr:isOnline() then
-                NetworkMgr:promptWifiOn()
+                NetworkMgr:beforeWifiAction(OTAManager.fetchAndProcessUpdate)
```
poire-z, line 459: You might want to provide as the callback an anonymous `function() OTAManager:fetchAndProcessUpdate() end`.
You're lossing the reference to sefl with just `OTAManager.fetchAndProcessUpdate` - which may be is fine with this particular one, but won't be with most real methods of a real object :)

frontend/ui/network/manager.lua
```diff
@@ -105,23 +119,14 @@ function NetworkMgr:promptWifiOff(complete_callback)
 end
 
 function NetworkMgr:turnOnWifiAndWaitForConnection(callback)
-    NetworkMgr:turnOnWifi()
-    local timeout = 30
-    local retry_count = 0
-    local info = InfoMessage:new{ text = T(_("Enabling Wi-Fi. Waiting for Internet connection…\nTimeout %1 seconds."), timeout)}
+    local info = InfoMessage:new{ text = _("Enabling Wi-Fi. Waiting for Internet connection…") }
```
poire-z, line 122: For this one, I personnaly use:
```lua
        -- mieux si on utilise trapwidget, on peut ainsi continuer à lire pendant que ça connecte.
        -- local info = InfoMessage:new{
        local TrapWidget = require("ui/widget/trapwidget")
        local info = TrapWidget:new{
            text = "Connecting to friendly WIFI...",
            image_file = "../mykoreader/resources/info-wifi.png",
        }
```

NiLuJe: Where is the icon from? ;).

Frenzie: Icon's besides the point, lol. :-P

(I mean, we can always consider it but we already have a whole series of 'em if desired.)

poire-z: (yep, just happen to be this: info-wifi )

NiLuJe:
I just realized that my mention of Kindle in the first post was a bit of a pipe dream: given how the network activity check is implemented, it can only work on platforms where we're the *only* thing running, so there won't be extra activity coming from outside...

Given how chatty the Kindle framework is, that would... never really work there ;p.

NiLuJe:
Okay, after the testing w/ the Wikipedia thingy, I'll probably unbreak the few stupid complex recursive callbacks I blindly added ;p.

NiLuJe:
... aaaand DNS resolution is broken again. Yay -_-".

frontend/apps/reader/modules/readerwikipedia.lua
```diff
@@ -385,7 +385,7 @@ end
 
 function ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang)
     if not NetworkMgr:isOnline() then
-        NetworkMgr:beforeWifiAction(function() ReaderWikipedia:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
```
poire-z: May be it would work better with using `self` :  `function() self:lookupWikipedia(...) end)` (the instance, not the class).
Also beware that the Wikipedia might be using Trapper and coroutines - dunno how that would interact with your recursions.

plugins/calibre.koplugin/wireless.lua
```diff
@@ -203,6 +203,11 @@ Do you want to continue? ]]), driver),
 end
 
 function CalibreWireless:connect()
+    if not NetworkMgr:isConnected() then
+        NetworkMgr:beforeWifiAction(function() CalibreWireless:connect() end)
+        return
+    end
```
poire-z, line 209: No risk of perpetual recursion if beforeWifiAction fails? (I guess the callback is called just on success so you're 100% sure `NetworkMgr:isConnected() ` is true when the recursive call is done ? haven't checked in your code, cheaper to just ask :)

poire-z:
(Another option instead of these recusrive calls would have been to use coroutines - so you would suspend the current code to do the connecting, and it would have coroutine.resume() on success - but that, even if may be not really complex line-of-code-wise, would make the whole thing less understandable. And may be make things even more fishy if we were already in a coroutine.)

plugins/calibre.koplugin/wireless.lua
NiLuJe: Yes. Yes there is :D.

That was on the list of things to look into, because if the callback actually works, that's indeed a (low, but not null) risk ;).

NiLuJe:
Okay, finally got to the bottom of #6421 & friends (#6429, #6430).

Apparently, `resolv.conf` isn't reloaded during Lua's lifetime (I don't know if this is a quirk on the old glibc being used on Kobo, or an actual braindead POSIX design. NetBSD has a mechanism in place to fix that, according to their doc. Linux's start with a giant "this doc is outrageously deprecated", so, who knows. `man 5 resolv.conf` does mention "loaded the first time a process uses the resolver", so, eh.).

Anyway, this means that starting with an empty `resolv.conf` breaks name resolution, even if it's later populated by the DHCP client.

Speaking of... killing udhcpc apparently now *clears* resolv.conf (despite not being passed the "release on exit" `-R` flag). I suppose this wasn't the case before the busybox update.
Killing dhcpcd (currently) doesn't, which is why moving to dhcpcd "worked".
Asking dhcpcd to release the IF *does* clear it, which is why I temporarily broke it.

Unfortunately, while dhcpcd has specific options to avoid touching resolv.conf when *acquiring* a lease, it's not honored when *releasing* one.

So, giant pile of workarounds incoming.

frontend/apps/reader/modules/readerwikipedia.lua
NiLuJe: Huh. That works, thanks ;).

plugins/calibre.koplugin/wireless.lua
NiLuJe: Actually, with the `turn_on` action, the callback will *definitely* only fire if the connection was successful (i.e., `isConnected`).

Things potentially get dicier if the check is actually an `isOnline`.

And `prompt` is possibly also iffier, i'm not sure if the callback is only called on success...

NiLuJe: There, fixed `prompt` to do the same ;).

NiLuJe:
More info on the `resolv.conf` madness:

A "modern" workaround for systemd users (ugh): https://unix.stackexchange.com/questions/487612/update-of-etc-resolve-conf-needs-restart-of-application

And the actual rabbit-hole, as discovered by a few prominent projects (do read the linked issues for more details): https://github.com/chef/chef/issues/2894

TL;DR: It's fixed in... glibc 2.26 (2017!).

platform/kobo/disable-wifi.sh
```diff
@@ -1,8 +1,15 @@
...
-# NOTE: Trying to do this nicely with 'wpa_cli terminate' and 'dhcpcd -d -k "${INTERFACE}"' trips mysterious buggy corner-cases... (#6424)
-killall udhcpc default.script dhcpcd wpa_supplicant 2>/dev/null
+# NOTE: Save our resolv.conf to avoid ending up with an empty one, in case the DHCP client wipes it on release (#6424).
+cp -a "/etc/resolv.conf" "/tmp/resolv.ko"
+if [ -x "/sbin/dhcpcd" ]; then
+    env -u LD_LIBRARY_PATH dhcpcd -d -k "${INTERFACE}"
+else
+    killall udhcpc default.script 2>/dev/null
+fi
+mv -f "/tmp/resolv.ko" "/etc/resolv.conf"
```
poire-z, line 12: By principle (and to avoid ultra minor SDcard wear), not super fond of having stuff rewritten on the root partition everytime we'll disable Wifi. Can't we check if it's really needed ?

NiLuJe: I could over-engineer it and add a checksum, but, err, really? ^^. It's ~100 bytes.

poire-z: "principles" :) 
You could just `test -f` & `grep nameserver` and if boths are true, no need to, right?

I have this in my my_enable-wifi.sh:
```
    # On évite de recréer /etc/resolv.conf à chaque fois
    if ! grep -q '#MyStaticDNSconfig' /etc/resolv.conf 2>/dev/null  ; then
        echo '#MyStaticDNSconfig' > /etc/resolv.conf
        for i in $dns ; do echo nameserver $i >> /etc/resolv.conf ; done
    fi
```

NiLuJe: Ideally, I'd like to simply keep it as-is, so just checking if it's sane isn't quite enough.

NiLuJe: And, more to the point: it's already being rewritten by the DHCP client on acquire/renew/release, so I don't really see why we should be worried about *this* rewrite but not the others ;).

poire-z: (The fact that someone shitted in my living room doesn't mean that I should now too :)
But well, OK. (I'm not using DHCP, so I guess I'll have to patch all these scripts - but, I'm already doing that, so, I guess I'm OK.)

NiLuJe: > (I'm not using DHCP, so I guess I'll have to patch all these scripts - but, I'm already doing that, so, I guess I'm OK.)

That was kind of my point ;).

For 99.999% of the userbase, this is necessary ;).

poire-z: btw, are you sure the killall or dhcpd -k are done removing the resolv.conf when you are mv'ing it back?

NiLuJe: Yep, both are blocking!

frontend/ui/network/manager.lua
```diff
@@ -139,6 +139,9 @@ function NetworkMgr:turnOnWifiAndWaitForConnection(callback)
     self:scheduleConnectivityCheck(callback, info)
 end
 
+--- @note: The callback will only run *after* a *succesful* network connection.
+---        The only guarantee it provides is isConnected (i.e., an IP & a local gateway),
+---        *NOT* isOnline (i.e., WAN), se be careful with recursive callbacks!
 function NetworkMgr:beforeWifiAction(callback)
```
poire-z, line 167: A bit hard to grasp these kind of constructs (and a bit ugly, all with a @note):
```lua
    if not NetworkMgr:isOnline() then
        --- @note: Avoid infinite recursion, beforeWifiAction only guarantees isConnected, not isOnline.
        if not NetworkMgr:isConnected() then
            NetworkMgr:beforeWifiAction(function() self:lookupWikipedia(word, box, get_fullpage, forced_lang) end)
        else
            NetworkMgr:beforeWifiAction()
        end
```
Can't we passe a 2nd parameter to `beforeWifiAction(callback, wants_online=true/false)` (so it guarantees what we want) and have that shorter and more obvious ?
(just asking, I'm not even familiar with beforeWifiAction & co)

NiLuJe: That crossed my mind, but I kind of like those constructs, they make it extremely obvious whether the action requires isOnline or not (since lua doesn't have named parameters).

That said, it's not particularly pretty, so I might come up with a helper function that wraps the whole thing and has  a potentially more meaningful name than beforeWifiAction, though ;p

NiLuJe:
Okay, this should be in pretty good shape by now, I'll write a proper changelog, and flag it ready ;).

A few quick notes:

* I have no idea how the beforeWifiAction stuff behaves on devices other than a Kobo (i.e., does the callback forwarding behave? Do things horribly implode?).

As far as callback forwarding is concerned, I tested:

* CloudStorage
* ReaderWikipedia
* OTAManager
* Translator
* OPDSBrowser

I did NOT test stuff that I don't usually use and/or is annoying to setup. I don't expect things to break, but the callback forwarding *may* be misbehaving when attempting those while disconnected, although *most* of those are fairly trivial:

* ReaderDictionary
* CalibreWireless
* Evernote
* GoodReads
* KOSync
* NewsDownloader
* Send2Ebook
* TimeSync
* Wallabag
* ZSync

platform/kobo/disable-wifi.sh
poire-z: (Thanks for old_hash vs new_hash :)

frontend/ui/network/manager.lua
```diff
@@ -174,6 +253,68 @@ function NetworkMgr:setHTTPProxy(proxy)
...
+end
+
+-- This one is for callbacks that only require isConnected, and since that's guaranteed by beforeWifiAction,
+-- you also have a guarantee that the callback *will* run.
+function NetworkMgr:willRerunWhenConnected(callback)
+    if not self:isConnected() then
+        self:beforeWifiAction(callback)
+        return true
+    end
+
```
poire-z, line 313: (nitpick, but unnecessary blank line - here and above)

NiLuJe: That one was on purpose, to separate the actual core logic from the menu/UI shenanigans that come later.

The one in networkListener was done with the same kind of thing in mind, but doesn't really make sense because there are event handlers above AND below it ;p.

NiLuJe:
@pazos: I don't expect much trouble on Cervantes, FWIW. Might want to resync the Wi-Fi scripts at one point, but that's it ;).

----

I will run a test on Kindle later today, though :).

platform/kobo/disable-wifi.sh
NiLuJe: Was kind of free candy after the whole "let's implement killall -w manually" mess ^^.

NiLuJe:
> BTW, Does make sense to invest time to simulate the same auto down when idle behaviour on the (linux) emulator?

Do we even handle Wi-Fi *at all* on the emulator? (i.e., a [NetworkManager](https://wiki.gnome.org/Projects/NetworkManager) backend or something?).

NiLuJe:
Okay, made it not regress and mostly usable on Kindle, I'll leave UI fine-tuning to others if necessary ;).

At a quick glance, on PB, I'm hoping the InkView stuff is blocking, so we should be okay.

On reMarkable, err, do we even support controlling the Wi-Fi at all?

And on Android, the callbacks are never forwarded, so it shouldn't regress.

Frenzie:
> Do we even handle Wi-Fi at all on the emulator? (i.e., a NetworkManager backend or something?).

Haha no, nor should we.

pazos:
Nope, no Wi-Fi handled on the emulator, but there's a fake networkManager instance in https://github.com/koreader/koreader/blob/master/frontend/device/sdl/device.lua#L313-L334, which is useful to test things that rely on network state changes.

> On reMarkable, err, do we even support controlling the Wi-Fi at all?

Nope, reMarkable has Wi-Fi but we don't manage it. It should be pretty similar to Nickel: if we kill the blob running then WiFi will stay alive until the battery is empty.

frontend/ui/network/manager.lua
```diff
@@ -276,9 +416,26 @@ function NetworkMgr:getProxyMenuTable()
     }
 end
 
+function NetworkMgr:getPowersaveMenuTable()
+    return {
+        text = _("Kill Wi-Fi connection when inactive"),
```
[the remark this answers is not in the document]
NiLuJe: I'll admit that was somewhat colored by rfkill buttons, where the term is already used to make it exceedingly clear that the modem is actually turned off ;).

Issue #6297, opened 2020-06-22
[Fdroid]: potencial insecure behaviour


pazos:
@licaon-kter @Rudloff:

### introduction
KOReader runs on a bunch of devices and users are encouraged to modify the source of the program directly in order to fit their needs. In that respect Android is an *special* platform because users are not expected to modify the runtime behaviour of installed applications. 

To workaround platform limitations we enabled some options. We're totally fine with them but they can be considered **local code execution vulnerabilities.**

### impact

A local attacker can take control of granted permissions (internet!, full storage!, writte_settings?) and available api calls.

LuaJIT runs in a separate thread and stays alive until the activity is destroyed. It is possible to execute code even when the app is paused or stopped.

### ways to alter runtime behaviour

These don't include any potential exploit, just intended behaviour.

##### lua

- `/sdcard/koreader/patch.lua` will be executed always on early init.

##### shell scripts

- scripts in `/sdcard/koreader/scripts.always` will be executed always on early init.
- scripts in `/sdcard/koreader/scripts.afterupdate` will be executed once per version on early init.

##### requested by user

- the file manager can run shell scripts if requested by the user.
- The `BackgroundRunner` plugin breaks expected activity behaviour and makes the app a mix between an activity and a service. Please consider that too.

### conclusion

Is not too serious for most KOReader users but could be serious enough for most Fdroid users. We can disable all/part of them in the fdroid flavor, if you wish.

licaon-kter:
This is available in the Play version?

pazos:
> This is available in the Play version?

There's no play version. Just yours and the one that we distribute in both github and http://build.koreader.rocks/download/

licaon-kter:
Silly me, didn't expect that :)

licaon-kter:
I wouldn't mind a toggle in settings to make this opt-in functionality, not ON by default, so there are no surprises.

Also, I small note in https://github.com/koreader/koreader-metadata/blob/98fd6c8fe0f3442a62346ee44454301007db3d8b/en-US/full_description.txt would not hurt either.

@IzzySoft any thoughts?

IzzySoft:
I agree with @licaon-kter: opt-in and full transparency should be sufficient. You might also consider offering those features via an add-on, so users "more paranoid" than me (are there any? :rofl:) feel safe as well.

An alternative build flavor could combine the two of course: one that has the functionality ("opt-in", as described) – and a second one ("lite"?) coming completely without, so users have the choice.

[...]

PR #6426, opened 2020-07-24
[Android]: Native light dialog

zwim:
Native Lights dialog on android (Tolino Epos 2, Phone)
has to be applied togegher with https://github.com/koreader/android-luajit-launcher/pull/239.patch and changed `BRIGHTNESS_MAX = 255` in TolinoWarmthController.kt.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6426)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -380,7 +380,7 @@ end
 function Device:showLightDialog()
     local usleep = require("ffi/util").usleep
     local title = android.isEink() and _("Frontlight settings") or _("Light settings")
-    android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("Ok"), _("Cancel"))
+    android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("OK"), _("Cancel"))
     repeat
         usleep(25000) -- sleep 25ms before next check if dialog was quit
     until (android.lights.dialogState() ~= C.ALIGHTS_DIALOG_OPENED)
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -247,10 +247,10 @@ function Device:init()
 
     -- TODO fl_last_level is a leftover from frontlightwidget
     -- check if we enable a custom light level for this activity
-    local last_value = G_reader_settings:readSetting("fl_last_level")
-    if type(last_value) == "number" and last_value >= 0 then
-        Device:setScreenBrightness(last_value)
-    end
+    --local last_value = G_reader_settings:readSetting("fl_last_level")
+    --if type(last_value) == "number" and last_value >= 0 then
+    --    Device:setScreenBrightness(last_value)
+    --end
 
     Generic.init(self)
 end
@@ -379,7 +379,8 @@ end
 
 function Device:lightDialog()
     local usleep = require("ffi/util").usleep
-    android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
+    local title = android.isEink() and _("Frontlight settings") or _("Light settings")
+    android.settings.dialog(title, _("Brightness"), _("Warmth"), _("Ok"))
     repeat
         usleep(75000) -- sleep 75ms before next check if dialog was quit
     until (not android.isFrontlightDialogRunning())
@@ -395,4 +396,3 @@ android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
     android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
 
 return Device
-
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -245,7 +245,8 @@ function Device:init()
         android.setBackButtonIgnored(true)
     end
 
-    -- TODO fl_last_level is a leftover from frontlightwidget
+    --- @todo remove fl_last_level and revert hasLightFallback if frontlightwidget
+
     -- check if we enable a custom light level for this activity
     --local last_value = G_reader_settings:readSetting("fl_last_level")
     --if type(last_value) == "number" and last_value >= 0 then
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -377,18 +377,28 @@ function Device:canExecuteScript(file)
     end
 end
 
-function Device:lightDialog()
+function Device:showLightDialog()
     local usleep = require("ffi/util").usleep
     local title = android.isEink() and _("Frontlight settings") or _("Light settings")
-    android.settings.dialog(title, _("Brightness"), _("Warmth"), _("Ok"))
+    android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("Ok"), _("Cancel"))
     repeat
-        usleep(75000) -- sleep 75ms before next check if dialog was quit
-    until (not android.isFrontlightDialogRunning())
-    logger.info("new brightness: " .. self.powerd.fl_intensity)
-    self.powerd:setIntensityHW(self.powerd:frontlightIntensityHW())
-    if android.isWarmthDevice() then
-        self.powerd:setWarmth(self.powerd:getWarmth())
-        logger.info("new warmth: " .. self.powerd.fl_warmth)
+        usleep(25000) -- sleep 25ms before next check if dialog was quit
+    until (android.lights.dialogState() ~= C.ALIGHTS_DIALOG_OPENED)
+    local action = android.lights.dialogState()
+    if action == C.ALIGHTS_DIALOG_OK then
+        self.powerd.fl_intensity = self.powerd:frontlightIntensityHW()
+        logger.dbg("Dialog OK, brightness: " .. self.powerd.fl_intensity)
+        if android.isWarmthDevice() then
+            self.powerd.fl_warmth = self.powerd:getWarmth()
+            logger.dbg("Dialog OK, warmth: " .. self.powerd.fl_warmth)
+        end
+    elseif action == C.ALIGHTS_DIALOG_CANCEL then
+        logger.dbg("Dialog Cancel, brightness: " .. self.powerd.fl_intensity)
+        self.powerd:setIntensityHW(self.powerd.fl_intensity)
+        if android.isWarmthDevice() then
+            logger.dbg("Dialog Cancel, warmth: " .. self.powerd.fl_warmth)
+            self.powerd:setWarmth(self.powerd.fl_warmth)
+        end
     end
 end
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -377,6 +377,20 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:lightDialog()
+    local usleep = require("ffi/util").usleep
+    android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
+    repeat
+        usleep(75000) -- sleep 75ms before next check if dialog was quit
+    until (not android.isFrontlightDialogRunning())
+    logger.info("new brightness: " .. self.powerd.fl_intensity)
+    self.powerd:setIntensityHW(self.powerd:frontlightIntensityHW())
+    if android.isWarmthDevice() then
+        self.powerd:setWarmth(self.powerd:getWarmth())
+        logger.info("new warmth: " .. self.powerd.fl_warmth)
+    end
+end
+
 android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
     android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -72,6 +72,7 @@ local Device = Generic:new{
     hasEinkScreen = function() return android.isEink() end,
     hasColorScreen = function() return not android.isEink() end,
     hasFrontlight = yes,
+    hasNaturalLight = android.isWarmthDevice,
     hasLightLevelFallback = yes,
     canRestart = no,
     canSuspend = no,
@@ -244,6 +245,7 @@ function Device:init()
         android.setBackButtonIgnored(true)
     end
 
+    -- TODO fl_last_level is a leftover from frontlightwidget
     -- check if we enable a custom light level for this activity
     local last_value = G_reader_settings:readSetting("fl_last_level")
     if type(last_value) == "number" and last_value >= 0 then
@@ -379,3 +381,4 @@ android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
     android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
 
 return Device
+
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -29,7 +29,7 @@ function AndroidPowerD:setWarmth(warmth)
 end
 
 function AndroidPowerD:getWarmth()
-   return android.getScreenWarmth() * self.warm_diff
+    return android.getScreenWarmth() * self.warm_diff
 end
 
 function AndroidPowerD:getCapacityHW()
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -2,17 +2,47 @@ local BasePowerD = require("device/generic/powerd")
 local _, android = pcall(require, "android")
 
 local AndroidPowerD = BasePowerD:new{
-    fl_min = 0, fl_max = 25,
-    fl_intensity = 10,
+    fl_min = 0, fl_max = 100, -- consistant scale across android, kobo ...
+--    fl_intensity = 10,
 }
 
 function AndroidPowerD:frontlightIntensityHW()
-    return math.floor(android.getScreenBrightness() / 255 * self.fl_max)
+    local intensity = math.floor(android.getScreenBrightness() / android.getScreenMaxBrightness() * self.fl_max)
+    if intensity < 0 then
+        intensity = 0
+    end
+    return intensity
 end
 
 function AndroidPowerD:setIntensityHW(intensity)
-    android.setScreenBrightness(math.floor(255 * intensity / self.fl_max))
+    self.fl_intensity = intensity
+    android.setScreenBrightness(math.floor(intensity * android.getScreenMaxBrightness() / self.fl_max))
 end
+
+function AndroidPowerD:init()
+    if self.device:hasNaturalLight() then
+        self.fl_warmth = self.getWarmth()
+    end
+end
+
+function AndroidPowerD:setWarmth(warmth)
+    self.fl_warmth = warmth
+    android.setScreenWarmth( warmth / AndroidPowerD:getMaxWarmth() )
+end
+
+function AndroidPowerD:getWarmth()
+    local warmth = android.getScreenWarmth() * AndroidPowerD.getMaxWarmth()
+    return warmth
+end
+
+function AndroidPowerD:getMaxWarmth()
+    return android.getScreenMaxWarmth()
+end
+
+function AndroidPowerD:getMinWarmth()
+    return android.getScreenMinWarmth()
+end
+
 
 function AndroidPowerD:getCapacityHW()
     return android.getBatteryLevel()
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -4,8 +4,6 @@ local _, android = pcall(require, "android")
 local AndroidPowerD = BasePowerD:new{
     fl_min = 0,
     fl_max = 100,
-    bright_diff = android.getScreenMaxBrightness() - android.getScreenMinBrightness(),
-    warm_diff = android.getScreenMaxWarmth() - android.getScreenMinWarmth(),
 }
 
 function AndroidPowerD:frontlightIntensityHW()
@@ -18,19 +16,20 @@ function AndroidPowerD:setIntensityHW(intensity)
 end
 
 function AndroidPowerD:init()
+    self.bright_diff = android:getScreenMaxBrightness() - android:getScreenMinBrightness()
     if self.device:hasNaturalLight() then
-        self.fl_warmth = self.getWarmth()
+        self.warm_diff = android:getScreenMaxWarmth() - android:getScreenMinWarmth()
+        self.fl_warmth = self:getWarmth()
     end
 end
 
 function AndroidPowerD:setWarmth(warmth)
     self.fl_warmth = warmth
-    android.setScreenWarmth(warmth / (android.getScreenWarmth() * self.warm_diff))
+    android.setScreenWarmth(warmth / self.warm_diff)
 end
 
 function AndroidPowerD:getWarmth()
-    local warmth = android.getScreenWarmth() * self.bright_diff
-    return warmth
+   return android.getScreenWarmth() * self.warm_diff
 end
 
 function AndroidPowerD:getCapacityHW()
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -2,21 +2,19 @@ local BasePowerD = require("device/generic/powerd")
 local _, android = pcall(require, "android")
 
 local AndroidPowerD = BasePowerD:new{
-    fl_min = 0, fl_max = 100, -- consistant scale across android, kobo ...
---    fl_intensity = 10,
+    fl_min = 0,
+    fl_max = 100,
+    bright_diff = android.getScreenMaxBrightness() - android.getScreenMinBrightness(),
+    warm_diff = android.getScreenMaxWarmth() - android.getScreenMinWarmth(),
 }
 
 function AndroidPowerD:frontlightIntensityHW()
-    local intensity = math.floor(android.getScreenBrightness() / android.getScreenMaxBrightness() * self.fl_max)
-    if intensity < 0 then
-        intensity = 0
-    end
-    return intensity
+    return math.floor(android.getScreenBrightness() / self.bright_diff * self.fl_max)
 end
 
 function AndroidPowerD:setIntensityHW(intensity)
     self.fl_intensity = intensity
-    android.setScreenBrightness(math.floor(intensity * android.getScreenMaxBrightness() / self.fl_max))
+    android.setScreenBrightness(math.floor(intensity * self.bright_diff / self.fl_max))
 end
 
 function AndroidPowerD:init()
@@ -27,22 +25,13 @@ end
 
 function AndroidPowerD:setWarmth(warmth)
     self.fl_warmth = warmth
-    android.setScreenWarmth( warmth / AndroidPowerD:getMaxWarmth() )
+    android.setScreenWarmth(warmth / (android.getScreenWarmth() * self.warm_diff))
 end
 
 function AndroidPowerD:getWarmth()
-    local warmth = android.getScreenWarmth() * AndroidPowerD.getMaxWarmth()
+    local warmth = android.getScreenWarmth() * self.bright_diff
     return warmth
 end
-
-function AndroidPowerD:getMaxWarmth()
-    return android.getScreenMaxWarmth()
-end
-
-function AndroidPowerD:getMinWarmth()
-    return android.getScreenMinWarmth()
-end
-
 
 function AndroidPowerD:getCapacityHW()
     return android.getBatteryLevel()
--- a/frontend/device/devicelistener.lua
+++ b/frontend/device/devicelistener.lua
@@ -114,12 +114,18 @@ if Device:hasFrontlight() then
 
         if new_intensity == nil then return true end
         -- when new_intensity <=0, toggle light off
+        self:onSetFlIntensity(new_intensity)
+        self:onShowIntensity()
+        return true
+    end
+
+    function DeviceListener:onSetFlIntensity(new_intensity)
+        local powerd = Device:getPowerDevice()
         if new_intensity <= 0 then
             powerd:turnOffFrontlight()
         else
             powerd:setIntensity(new_intensity)
         end
-        self:onShowIntensity()
         return true
     end
 
@@ -202,13 +208,19 @@ if Device:hasFrontlight() then
             direction = 1
         end
         local warmth = powerd.fl_warmth + direction * delta_int
+        self:onSetFlWarmth(warmth)
+        self:onShowWarmth()
+        return true
+    end
+
+    function DeviceListener:onSetFlWarmth(warmth)
+        local powerd = Device:getPowerDevice()
         if warmth > 100 then
             warmth = 100
         elseif warmth < 0 then
             warmth = 0
         end
         powerd:setWarmth(warmth)
-        self:onShowWarmth()
         return true
     end
 
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -205,7 +205,7 @@ if Device:isAndroid() then
     common_settings.frontlight = {
         text = _("Frontlight"),
         callback = function()
-            Device:lightDialog()
+            Device:showLightDialog()
         end,
     }
 
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -201,31 +201,13 @@ if Device:isAndroid() then
     local isAndroid, android = pcall(require, "android")
     if not isAndroid then return end
 
-    local logger = require("logger")
-    local powerd = Device:getPowerDevice()
 
-    local function FrontlightDialogWrapper( )
-        local usleep = require("ffi/util").usleep
-        android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
-        logger.info("intensity: " .. powerd:frontlightIntensityHW())
-        repeat
-            usleep(75000) -- sleep 75ms before next check if dialog was quit
-        until (not android.isFrontlightDialogRunning())
-        logger.info("powerd.fl_intensity: " .. powerd.fl_intensity)
-        powerd:setIntensityHW(powerd:frontlightIntensityHW())
-        if android.isWarmthDevice~=nil and android.isWarmthDevice() then
-            powerd:setWarmth(powerd:getWarmth())
-            logger.info("powerd.fl_warmth: " .. powerd.fl_warmth)
-        end
-
-    end
-
-    -- on Android: overwrite generic frontlight with a native Dialog
+    -- overwrite generic frontlight with a native Dialog
     common_settings.frontlight = {
         text = _("Frontlight"),
         callback = function()
-            FrontlightDialogWrapper()
-            end,
+            Device:lightDialog()
+        end,
     }
 
     -- screen timeout options, disabled if device needs wakelocks.
@@ -436,7 +418,7 @@ common_settings.document = {
                 local interval = G_reader_settings:readSetting("auto_save_settings_interval_minutes")
                 local s_interval
                 if interval == false then
-                    s_interval = "only on close"
+                    s_interval = _("only on close")
                 else
                     s_interval = T(N_("every 1 m", "every %1 m", interval), interval)
                 end
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -201,6 +201,14 @@ if Device:isAndroid() then
     local isAndroid, android = pcall(require, "android")
     if not isAndroid then return end
 
+    -- on Android: overwrite generic frontlight with a native Dialog
+    common_settings.frontlight = {
+        text = _("Frontlight"),
+        callback = function()
+            android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
+            end,
+    }
+
     -- screen timeout options, disabled if device needs wakelocks.
     common_settings.screen_timeout = require("ui/elements/screen_android"):getTimeoutMenuTable()
 
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -201,11 +201,30 @@ if Device:isAndroid() then
     local isAndroid, android = pcall(require, "android")
     if not isAndroid then return end
 
+    local logger = require("logger")
+    local powerd = Device:getPowerDevice()
+
+    local function FrontlightDialogWrapper( )
+        local usleep = require("ffi/util").usleep
+        android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
+        logger.info("intensity: " .. powerd:frontlightIntensityHW())
+        repeat
+            usleep(75000) -- sleep 75ms before next check if dialog was quit
+        until (not android.isFrontlightDialogRunning())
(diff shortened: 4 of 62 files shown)


frontend/device/android/device.lua
```diff
@@ -375,7 +377,8 @@ function Device:canExecuteScript(file)
     end
 end
 
-android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
-    android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
+android.LOGI(string.format("Android %s - %s (API %d, product %s) - flavor: %s",
```
pazos, line 380: That's not needed. android.prop.product is == Device.model and that's already logged on a bunch of sites.

frontend/device/android/powerd.lua
```diff
@@ -3,17 +3,48 @@ local _, android = pcall(require, "android")
 
 local AndroidPowerD = BasePowerD:new{
     fl_min = 0, fl_max = 25,
```
pazos, line 5: probably these values can be retrieved from the backend?

frontend/device/android/powerd.lua
```diff
@@ -3,17 +3,48 @@ local _, android = pcall(require, "android")
...
-    fl_intensity = 10,
+--    fl_intensity = 10,
 }
 
 function AndroidPowerD:frontlightIntensityHW()
-    return math.floor(android.getScreenBrightness() / 255 * self.fl_max)
+    intensity = math.floor(android.getScreenBrightness() / 255 * self.fl_max)
+    if intensity < 0 then
```
pazos, line 11: is this because the -1? . I would like this on the backend too. So we can add a checkbox there to use system settings, like we do now.

frontend/device/android/powerd.lua
zwim: I think no. These are just for scaling the backend value to 0-25 in the frontend (gestures report).

frontend/device/devicelistener.lua
```diff
@@ -202,10 +202,18 @@ if Device:hasFrontlight() then
             direction = 1
         end
         local warmth = powerd.fl_warmth + direction * delta_int
-        if warmth > 100 then
```
pazos, line 205: I'm not familiar with devicelistener. Pinging @yparitcher to make sure we don't clash with other warmth devices

frontend/device/android/powerd.lua
zwim: Yes, a checkbox in the backend to use system settings would be good. As it/was is in the frontlightwidget, then we could use the logic behind `fl_last_level`.

frontend/device/devicelistener.lua
yparitcher: TBH, i am not very familiar with the warmth code, I don't have a warmth device, and the code was moved from `readerfrontlight` which is originally from `kobolight.koplugin`.

However from looking around, the frontend expects warmth to be 1 .. 100 (in frontlightwidget etc) and that powerd should translate that to whatever scale the device actually uses. so either have luajit-launcher do this for android, or fix that assumption everywhere.

pinging @NiLuJe & @Frenzie who should actually know what they are talking about. ;)

NiLuJe: Yeah, the front-end expects a 0-100 range, and it's the backend's responsibility to scale that appropriately.

NiLuJe: e.g., https://github.com/koreader/koreader/blob/db843929f0893383a3dcddbb6d77708691b8e393/frontend/device/kobo/powerd.lua#L47-L50

Some devices used a Kelvin scale on their end, others use an arbitrary [0, ??] range, or whatever else they want, really ^^.

(i.e., it's a mess :D).

zwim: @NiLuJe: The Tolino Epos2, on which I am developing should have almost the same Hardware as the Kobo Forma (but has an android firmware). Therefore, it should also use (direct or through the firmware) the same `/sys/class/backlight/tlc5947_bl/color` file.

I know nothing about the Kobo. Could it be, that also the kobo has only 10 steps of warmth and that the "color temperature" is just calculated somehow within the kobo firmware? Do you notice steps, when you change the warmth of the light? How many steps?

NiLuJe: That excerpt was for the Aura One, which had fancier (and buggier) LEDs.

Newer devices do rely on a hardware mixer which only exposes a [0, 10] range:

https://github.com/koreader/koreader/blob/db843929f0893383a3dcddbb6d77708691b8e393/frontend/device/kobo/device.lua#L231-L239

zwim: Now I see. Thank you.

zwim: @NiLuJe: When I look at the `devicelistener.lua` I realize, 
1.) that there is a scale for the intensity, so that you can swipe from `fl_min` to `fl_max`. 
2.) For warmth there is a fixed scale from 0 to 100.

Especially for my device, where I have only 11 steps of warmth a scale from 0..100 is annoying. 
Are there reasons, why the 100-scale should be there. Or shall we change it?

NiLuJe: No idea, I'm not the one who initially implemented warmth ;).

Since the first one was probably the KA1, which happened to be extremely flexible, a [0, 100] range probably felt like a good user-facing scale.

IIRC, I simply hacked the scaling for the NL widget.

I don't particularly care about the gesture, but I don't really see a great way to handle it either, it's already super vague to begin with, so it works well enough as-is.

zwim:
@elvvis: If you like, you can try my development version of the KOReader app. Native Lights on Epos 2 should work. You can also test it on your phone (without warmth).
https://www.dropbox.com/s/13gruqglatp83i9/koreader-android-arm-lights-test.apk?dl=0
If you want to test it, you have to uninstall KOReader first, as it is signed with a different key.

Differences to the original: 
*) Native Light Dialog on Android (phone and Epos2) under Gear->Frontlight/Beleuchtung,
*) Intensity scale goes from 0 to 100 (swipe on the left side up/down)
*) Warmth scale goes from 0 to 100 (swipe on the right side up/down)
*) Don't care about hanging punctuation (Hängende Satzzeichen), this is a different story.

I would appreciate feedback from you.
Ich würde mich sehr über deine Rückmeldung freuen.

elvvis:
@zwim: I have tested it on my Epos 2, I can't test it on the mobile phone because I don't have an android (only an iPhone).
Adjusting the brightness works fine, but adjusting the warmth doesn't (is this maybe because my Epos 2 is not rooted?)

zwim:
@elvvis: Unfortunately the Epos has to be rooted to get the warmth work. I don't see a way around it.
Do you get an error message with the warmth, or does it just not work?

elvvis:
@zwim: No, there was no error message, it just didn't work.
Well, I'll just have to see how I can root my Epos 2...

zwim:
@elvvis: Now in German:
 Ich bin der Anleitung von https://forum.alles-ebook.de/board/26-tolino/ gefolgt. Eventuell brauchst du noch die SuperSu app. Ich verwende `SuperSU Free v2.79` von Chainfire.

elvvis:
@zwim: Okay, vielen Dank für den Link, werde ich mir gleich mal anschauen. Von dieser Seite hatte ich auch die Anleitung, um überhaupt Apps auf dem Epos 2 zu installieren. Es ist denke ich keine schlechte Idee den Epos 2 zu rooten, dann bekomme ich wohl nun auch einen eigenen Screensaver zum Laufen, dank deiner letzter Implementierung, die jetzt Skripte erlaubt.

zwim:
@elvvis: The possibility to execute scripts has nothing to do with rooting your device. There are different "flavors" of KOReader: `F-Droid` and `rocks`.
The F-Droid flavor does not allow to execute user scripts. (see https://github.com/koreader/koreader/issues/6297)
The rocks flavor (http://koreader.rocks/) allows scripts.

elvvis:
@zwim: I just meant that with a rooted device you have direct access to the directory where the screensaver image is located.

frontend/device/android/powerd.lua
```diff
@@ -3,17 +3,47 @@ local _, android = pcall(require, "android")
 
 local AndroidPowerD = BasePowerD:new{
     fl_min = 0, fl_max = 25,
-    fl_intensity = 10,
+--    fl_intensity = 10,
 }
 
 function AndroidPowerD:frontlightIntensityHW()
-    return math.floor(android.getScreenBrightness() / 255 * self.fl_max)
+    local intensity = math.floor(android.getScreenBrightness() / 255 * self.fl_max)
```
pazos, line 10: the way to calcule intensity can be made generic for future implementations with:

`math.floor(android.getScreenBrightness() / android.getScreenMaxBrightness() * self.fl_max)`

frontend/device/android/powerd.lua
```diff
@@ -3,17 +3,47 @@ local _, android = pcall(require, "android")
...
+    if intensity < 0 then
+        intensity = 0
+    end
+    return intensity
 end
 
 function AndroidPowerD:setIntensityHW(intensity)
+    self.fl_intensity = intensity
     android.setScreenBrightness(math.floor(255 * intensity / self.fl_max))
```
pazos, line 19: same as above:

`android.setScreenBrightness(math.floor(android.getScreenMaxBrightness() * intensity / self.fl_max))`

frontend/device/android/powerd.lua
zwim: Ok. Shall I leave the check `if intensity < 0` for now?

zwim:
I have done the changes and then I did:
> git reset --soft HEAD~6
> git reset HEAD l10n
> git commit -m "name of new commit"

gitk shows sane results. But when I try to
`git push origin NativeLights` I get after the password input the following error:
```
To https://github.com/zwim/koreader.git
! [rejected]          NativeLights -> NativeLights (non-fast-forward)
error: Fehler beim Versenden einiger Referenzen nach 'https://github.com/zwim/koreader.git'
Hinweis: Aktualisierungen wurden zurückgewiesen, weil die Spitze Ihres aktuellen
Hinweis: Branches hinter seinem externen Gegenstück zurückgefallen ist. Führen Sie
Hinweis: die externen Änderungen zusammen (z. B. 'git pull ...') bevor Sie "push"
Hinweis: erneut ausführen.
Hinweis: Siehe auch die Sektion 'Note about fast-forwards' in 'git push --help'
Hinweis: für weitere Details.
```

Can you help me?

elvvis:
@zwim: Looks like your local branch is out of date. Did you do git pull?

Galunid:
`git push -f origin NativeLights` should fix your problem. There is a free ebook [ProGit](https://www.amazon.com/Pro-Git-Scott-Chacon-ebook-dp-B01ISNIKES/dp/B01ISNIKES/) on amazon if you're interested.

zwim:
> `git push -f origin NativeLights` should fix your problem. 

Thank you, that did  the trick.
P.S.: I will get the book.

frontend/device/android/powerd.lua
```diff
@@ -2,18 +2,48 @@ local BasePowerD = require("device/generic/powerd")
...
+    self.fl_warmth = warmth
+    android.setScreenWarmth( warmth / AndroidPowerD:getMaxWarmth() )
+end
+
+function AndroidPowerD:getWarmth()
+    local warmth = android.getScreenWarmth() * AndroidPowerD.getMaxWarmth()
+    return warmth
+end
+
+function AndroidPowerD:getMaxWarmth()
```
pazos, line 38: Probably AndroidPowerD:getMaxWarmth(),  AndroidPowerD:getMinWarmth() etc are not needed. We might want them initialized once. For now using maxWarmth is ok as minWarmth/minBrightness is always 0, but that doesn't need to apply on some circunstances.

So I would go for adding two ranges on the new instance, for example

```lua

local AndroidPowerD = BasePowerD:new{
    brightness_range = android.getScreenMaxBrightness() - android.getScreenMinBrightness(),
    warmth_range = android.getScreenMaxWarmth() - android.getScreenMinWarmth(),
}
```

Then you can use `self.brightness_range` instead of `AndroidPowerD:getMaxBrightness` and `self.warmth_range` instead of `AndroidPowerD:getMaxWarmth`

frontend/device/android/powerd.lua
pazos: I don't really understand what's the issue with fl_last_level.

On GenericController each time the activity is initialized it will pick light level from system settings. Then fl_last_level is used to override default/system settings.

Could you please explain me again why is this a problem? and where?. Just on Tolinos?

Sorry but I try to understand without testing, and that seems to break my head :/

frontend/device/android/powerd.lua
```diff
@@ -2,18 +2,48 @@ local BasePowerD = require("device/generic/powerd")
 local _, android = pcall(require, "android")
 
 local AndroidPowerD = BasePowerD:new{
-    fl_min = 0, fl_max = 25,
-    fl_intensity = 10,
+    fl_min = 0, fl_max = 100, -- consistant scale across android, kobo ...
+--    fl_intensity = 10,
```
pazos, line 6: is fl_intensity unused?, what's the reason to comment it out?. If is unused and not needed please remove it first.

frontend/device/android/powerd.lua
zwim: We can do so, but then we have to scale the brightness to our brightness range AND add the minBrightness as an offset.

frontend/device/android/powerd.lua
```diff
@@ -2,18 +2,48 @@ local BasePowerD = require("device/generic/powerd")
 local _, android = pcall(require, "android")
 
 local AndroidPowerD = BasePowerD:new{
-    fl_min = 0, fl_max = 25,
-    fl_intensity = 10,
+    fl_min = 0, fl_max = 100, -- consistant scale across android, kobo ...
```
pazos, line 5: I don't understand why this is needed here, but if it works I see no problem.

zwim: I changed that, because NuLuJe recommended a consistent scale across devices.
>  (I'd tend to agree, FWIW, consistent UI/settings across devices is a good thing to aim for).

frontend/device/android/powerd.lua
pazos: ???

frontend/device/android/powerd.lua
zwim: As I see, the fl_intensity is only used on Kobos and Kindles. 
* In Kobos powerd it is initialized in the init().
* On Kindle its a bit more complicated.
* It is also in generic, but there it is also nil.
* In `devicelistener.lua` it is checked, if fl_intensity~=nil, so no problem there.

I will remove it.

frontend/device/android/powerd.lua
zwim: In my reading: fl_last_level is used by the frontlightwidget to store the last level of frontlight. So if someone just copies his koreader folder from a kobo (using frontlightwidget) to an android device, there is a chance that fl_last_level is set to a special value. -> The App would set that level on every start and you cannot change it, as there is no frontlightwidget on android.
We could extend the native dialog with a checkbox "System Light Settings". We could clear the fl_last_level, if the checkbox is checked, if it is not set, we could store the values (intensity and warmth) in fl_last_level and a new warmth_last_level. On the next start KOReader would use the Systems values or *_last_level values.

frontend/device/android/powerd.lua
zwim: Lets say it different. The hardware has a range from `min` to `max`.  Min does not have to be 0.
The software has a range from [0..100]. 
If I want to map `value`  from the hardware scale to the software range I need `min` and `max`.
If I want to map `value` form the software range to the hardware scale I also need `min` and `max`.

So if I introduce a brightness_range, I would also need a brightness_offset. 
That`s why I am saying two variables.

frontend/device/android/powerd.lua
pazos: Nope, it was added in https://github.com/koreader/koreader/commit/ae50569055a86f2672058b55dc1f1abaed8cb291 and is only used on android for now.

It is used to setup a custom fl_level at launch because, as I already said, when the activity starts it uses generic values from system settings.

frontend/device/android/powerd.lua
pazos: Ehm, nope.

You need to scale whatever difference between max and min values given by the android SDK to the scale used by the frontend (now 0 - 100). Probably `range` is not a good name for this variable, but that's it.

frontend/device/android/powerd.lua
pazos: I think I will start an emulator and see what happens on your branch, because I don't get it :p

pazos: Ok. With https://github.com/koreader/android-luajit-launcher/pull/239/commits/d2de9aab3e39119b4da4bfac1f71ea3705620229 getBrightness won't report default intensity (-1). Instead it will report system settings values, so you won't need to check intensity < 1 here.

pazos:
Just tested on an emulator.

Things work but each time you use the frontlight dialog we lost sync with gestures. For instance:

1. Exit KOReader
2. Set brightness to max on android settings
3. Start KOReader, open a book.
4. Perform a gesture. The initial position is retrieved from android system settings, so it matches.
5. Open frontlight dialog, it will report current brightness, change it to other value
6. Perform a gesture again. Since fl_level wasn't updated in the native dialog the gesture will work but uses the previous fl_level to calculate distance.

Not sure how to fix that (ie: setting fl_level after a native confirmbox). As we use one-way communication between frontend and SDK methods there's no possible to set things the other way around.

Pinging @poire-z, who saved my ass a lot of times with this kind of things: calling the frontlight dialog doesn't block and we cannot guess when the user leave the frontlight dialog. In a ideal world it should be a matter of updating fl_level each time we leave the dialog, but I'm not sure how to proceed.

poire-z:
I don't know much about the frontlight details (and I swear I would never look at it anymore now that it works ! :) - and I don't want to dig into this PR, but, from far away from all that, one idea:

When it's hard to properly forward infos or values with functions and callbacks: use Events :)
You could have the frontlight dialog do on each change:
`UIManager:broadcastEvent(Event:new("FrontlightLevelChanged", new_level))`
and may be somewhere far away you can catch the event and get the new value and update the value on that side.

frontend/device/android/powerd.lua
zwim: Now I see. We proceed like you suggest. My view was biased by embedded programming.

In the future we should not forget: Our backend controller functions (setWarmth, setBrightness) expect values from 0 to Max_x. Our getXXXX functions return values from 0 to Max_x.
0 means dark or cold,  Max_x means their respective maximum brightness or warmth. That is the way we do now.
So the backend controller functions  (setWarmth, setBrightness) have to take care if the range is inverted (as we do now for warmth) and if a new super duper fancy device has its hardware minimum not at 0 then the setWarmth/setBrightness function shift the value.

If we do so, we can drop getMinWarmth and getMinBrightness as they are not needed.
Would that be a feasible solution?

pazos: the backend code defines a interface but doesn't implement it. The interface allow us to encapsulate device specific implementations in a common way. Nor the generic impl or the Tolino impl do what you said: they check that a given value is between min and max before setting that value (for warmth and brightness) and they return implementation min/max values. It happens that in all the current implementations min = 0 but that's not part of the interface, just a specific implementation of it.

frontend/device/android/powerd.lua
zwim: OK.

frontend/device/android/powerd.lua
zwim: > the backend code defines a interface but doesn't implement it. The interface allow us to encapsulate device specific implementations in a common way.

OK with that.
> Nor the generic impl or the Tolino impl do what you said: they check that a given value is between min and max before setting that value (for warmth and brightness)

The warmth interface inverts the range. On the /sys/class../color file 0 means warm and 10 means cold. In line 51 of TolinoWarmthController there is  `WARMTH_MAX - colorFile.readText().toInt()`

> and they return implementation min/max values. It happens that in all the current implementations min = 0 but that's not part of the interface, just a specific implementation of it.

Why shouldn't we make it part of the interface? Up to now it would make no difference.
Again: If we don`t make it part of the interface then we would need 2 values an offset and a range.

pazos: > The warmth interface inverts the range.

Nope. The **Tolino implementation** of the interface inverts the range. The interface is generic enough and doesn't care about specific implementations.

> Why shouldn't we make it part of the interface? 

Please do read https://www.tutorialspoint.com/java/java_interfaces.htm.  I can't solve your issues if you don't try to understand what an interface is :p.

NiLuJe: > The warmth interface inverts the range. On the /sys/class../color file 0 means warm and 10 means cold. In line 51 of TolinoWarmthController there is WARMTH_MAX - colorFile.readText().toInt()

Historical quirk. The first backend didn't. Newer ones does (hence nl_invert on Kobos w/ a mixer).

zwim: Thank you for your patience. The whole discussion is about mapping a value from one range to another as for example is described in https://www.arduino.cc/reference/en/language/functions/math/map/

Somewhere it has to be done! For the mapping on every side two values are necessary. In our case it can be assumed one of the four parameters is zero. So we need two ranges and one offset. 

I have the feeling we are not talking the same language. So I give you the code. Do whaterver you want with it, I won't work on it anymore.

My first PR (https://github.com/koreader/koreader/pull/6332) was not worse than the actual one. IMHO the user experience in KOReader was more consistent with the former one.

I am out now!

pazos: @zwim: please do whatever you want but don't take it personally. As I said your first PR was ok as a patch for your own usage. It will crash on @elvvis device as he didn't have root and JNI calls are not protected.

This whole PR is almost done. We don't speak the same mother tongue (I speak galician and you speak german)  but we're both almost good with english (and have google translate to workaround our limitations :))

> I won't work on it anymore.

That's fine too but I'm afraid I cannot test the implementation myself, so I will wait for any tester before merging

> The whole discussion is about mapping a value from one range to another

That can be implemented on the backend or on the frontend, but has nothing to do with the interface itself, just with the implementation. Feel free to code it as you want.

> I am out now!

Closing this issue for the moment. Please ping me (or any other person here) if you want to open it again. AFAICT there's nothing wrong with current implementation, except for https://github.com/koreader/koreader/pull/6426#issuecomment-663864729, but that has nothing to do with your code.

zwim: @pazos: 
First, the most important thing: It is nothing personal. You did a good job. It was fine to work with you. My saying about the language has nothing to do with our mother tongue,  I meant it abstract, we do not understand each other. Maybe this is due to the medium we are talking. In real life it would be different. You know what you are doing, so do I.

Second thing: If you need testers, feel free to give me a link to an apk.

Third thing: I could be implemented in  devicelistener.lua line 204 and 113 by calling checking if getXXXX is the same than powerd.fl_intensity, but poire-z's solution seems more elegant.

Fourth thing: I will look on the changes in the showcase tomorrow.

zwim:
@pazos: I see no problem.

frontend/device/android/powerd.lua
pazos: Ok, thank you. This conversation was resolved successfully 👍. I will comment the third thing below.

pazos:
> When it's hard to properly forward infos or values with functions and callbacks: use Events :)
You could have the frontlight dialog do on each change:
UIManager:broadcastEvent(Event:new("FrontlightLevelChanged", new_level))

> Third thing: I could be implemented in devicelistener.lua line 204 and 113 by calling checking if getXXXX is the same than powerd.fl_intensity, but poire-z's solution seems more elegant.

@poire-z @zwim: thank you both for your suggestions. I think the main issue here is that we maintain two different states (one is in android PowerD and the other provided by android sdk).

If I do understand correctly the very first reason to keep track of the state in PowerD is the lack of ways to retrieve current fl_level on original Kindle/Kobo implementations (pinging @NiLuJe because he might remember).  That's not a problem in android since the SDK has the info ready for us.

So I was thinking in a stateless (from PowerD POV) AndroidPowerD. Still have no idea how to do that without touching a lot of code and because usually lua blows my mind. Could be possible to define fl_level and the warmth equivalent as functions that return the relevant SDK methods? This way when a gesture happens the base brightness/warmth would be retrieved from android and we solve the out-of-sync problem.

For instance:
```lua
self.fl_level = function()
    return android.getScreenBrightness()
end
```

It is an untested solution. I don't know if current code can handle fl_level (and nl_level or whatever is used to keep track of warmth) being functions.

Frenzie:
@pazos 
> If I do understand correctly the very first reason to keep track of the state in PowerD is the lack of ways to retrieve current fl_level on original Kindle/Kobo implementations (pinging @NiLuJe because he might remember).

Dunno about Kindle but that's true for Kobo.

NiLuJe:
Yeah, each and every part of the FL/warmth handling is essentially a giant clusterfuck because of this (and/or various device-specific quirks).

It's not pretty, but "fixing" it would require essentially breaking everything, which means re-wrirting a ton of device-specific code... without having the actual devices. And I think every one of us here has at least attempted some form of cleanup in there once and burned their fingers off in the process...

zwim:
@pazos: Why not doing it like that. We add a private flag which is set, before the Dialog is started. When pressing OK, the flag gets cleared.
In the common_settings_menu_table, we start the native dialog and wait there until the flag gets cleared. Then we update powerd.fl_intensity and powerd.warmth so the rest of the app can use the new values.

This works for me on Phone and Epos:
add to JNILuaInterface.kt
`fun isFrontlightDialogRunning(): Int`

add to MainActivity in MainActivity.kt
```
    private var runningFrontlightDialog: Boolean = false

    fun clearFrontlightDialogRunning() {
        runningFrontlightDialog=false
    }
```

and change ShowFrontlightDialog to:
```
    override fun showFrontlightDialog(title: String, dim: String, warmth: String, okButton: String): Int {
        val hasWarmth = lights.hasWarmth()
        runningFrontlightDialog = true
        runOnUiThread {
            val dimText = TextView(this@MainActivity)
            val dimSeekBar = SeekBar(this@MainActivity)
            dimText.text = dim
            dimText.gravity = Gravity.CENTER_HORIZONTAL
            dimText.textSize = 18f
            dimSeekBar.max = lights.getMaxBrightness()
            dimSeekBar.progress = lights.getBrightness(this@MainActivity)
            dimSeekBar.setOnSeekBarChangeListener(object: SeekBar.OnSeekBarChangeListener {
                override fun onStartTrackingTouch(p0: SeekBar?) {}
                override fun onStopTrackingTouch(p0: SeekBar?) {}
                override fun onProgressChanged(p0: SeekBar?, p1: Int, p2: Boolean) {
                    lights.setBrightness(this@MainActivity, p1)
                }
            })
            val linearLayout = LinearLayout(this@MainActivity)
            linearLayout.orientation = LinearLayout.VERTICAL
            linearLayout.addView(dimText)
            linearLayout.addView(dimSeekBar)
            if (hasWarmth) {
                val warmthText = TextView(this@MainActivity)
                val warmthSeekBar = SeekBar(this@MainActivity)
                warmthText.text = warmth
                warmthText.gravity = Gravity.CENTER_HORIZONTAL
                warmthText.textSize = 18f
                warmthSeekBar.max = lights.getMaxWarmth()
                warmthSeekBar.progress = lights.getWarmth(this@MainActivity)
                warmthSeekBar.setOnSeekBarChangeListener(object : SeekBar.OnSeekBarChangeListener {
                    override fun onStartTrackingTouch(p0: SeekBar?) {}
                    override fun onStopTrackingTouch(p0: SeekBar?) {}
                    override fun onProgressChanged(p0: SeekBar?, p1: Int, p2: Boolean) {
                        lights.setWarmth(this@MainActivity, p1)
                    }
                })
                linearLayout.addView(warmthText)
                linearLayout.addView(warmthSeekBar)
            }

            val builder = AlertDialog.Builder(this@MainActivity)
            builder.setTitle(title)
                   .setCancelable(false)
                   .setPositiveButton(okButton) { dialog, which ->  clearFrontlightDialogRunning() }

            val dialog: AlertDialog = builder.create()
            dialog.setView(linearLayout)
            dialog.show()
        }
        return 0
    }
```

then add in android.lua:
```    
     android.isFrontlightDialogRunning = function()
        return JNI:context(android.app.activity.vm, function(jni)
            return jni:callIntMethod(
                android.app.activity.clazz,
                "isFrontlightDialogRunning",
                "()I"
            ) == 1
        end)
    end
```

at last change in common_settings_menu_table.lua
```

    local function FrontlightDialogWrapper( )
        android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
        logger.warn("intensity: " .. powerd:frontlightIntensityHW())
        repeat
            usleep(50000)
        until (not android.isFrontlightDialogRunning())
        logger.warn("intensity new: " .. powerd:frontlightIntensityHW())

        logger.warn("powerd.fl_intensity: " .. powerd.fl_intensity )
        powerd:setIntensityHW(powerd.frontlightIntensityHW())
            powerd:setWarmth(powerd:getWarmth())
            logger.warn("powerd.fl_warmth: " .. powerd.fl_warmth )
        end

    end

    -- on Android: overwrite generic frontlight with a native Dialog
    common_settings.frontlight = {
        text = _("Frontlight"),
        callback = function()
            FrontlightDialogWrapper()
            end,
    }
```

What is your opinion about that solution?

pazos:
> What is your opinion about that solution?

:+1: 

Looks wonderful, could you submit a patch against https://github.com/koreader/android-luajit-launcher ?. It can supersed https://github.com/koreader/android-luajit-launcher/pull/239 entirely.

Very happy to see you working on this :smile:. Believe me this contributions would make a lot easier to add new implementations with custom lights!!! Thanks again!

zwim:
@pazos: Please apply my patch for luajit-launcher before trying this PR.
On my phone and Epos the native lights dialog and the light gestures are in sync now.

frontend/device/android/device.lua
```diff
@@ -375,7 +377,22 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:lightDialog()
+    local usleep = require("ffi/util").usleep
+    android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
```
pazos, line 382: Nit: not sure about the title: frontlight dialog. On the old widget is called "frontlight", which is true for ereaders. On android "frontlight" only applies to e-ink devices.

Maybe

```lua
local title = android.isEink() and _("Frontlight settings") or _("Light settings")
```
:thinking: Pinging @Frenzie 

Probably brightness is a better name than intensity.

frontend/device/android/device.lua
```diff
@@ -244,6 +245,7 @@ function Device:init()
         android.setBackButtonIgnored(true)
     end
 
+    -- TODO fl_last_level is a leftover from frontlightwidget
```
pazos, line 248: This comment is right now. The old code can be removed from frontlightwidget. I think this can be done later and now just comment the lines below.

frontend/device/android/device.lua
```diff
@@ -375,7 +377,22 @@ function Device:canExecuteScript(file)
...
+        self.powerd:setWarmth(self.powerd:getWarmth())
+        logger.info("new warmth: " .. self.powerd.fl_warmth)
+    end
+end
+
 android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
     android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
 
 return Device
+
```
pazos, line 398: Intended?

frontend/ui/elements/common_settings_menu_table.lua
```diff
@@ -201,6 +201,15 @@ if Device:isAndroid() then
     local isAndroid, android = pcall(require, "android")
     if not isAndroid then return end
 
+
```
pazos, line 204: two consecutive empty lines, intended?

frontend/device/android/device.lua
Frenzie: If that's about brightness, agreed. :+1:

zwim: Changed.

frontend/device/android/device.lua
zwim: OK.

frontend/device/android/device.lua
zwim: poire-z asked me in https://github.com/koreader/koreader/pull/6385 to add a `\n` at end of file.

frontend/ui/elements/common_settings_menu_table.lua
zwim: OK.

frontend/device/android/device.lua
NiLuJe: That's technically two of 'em ;).

(i.e. we definitely want a LF after `Device`, but then you have another one on L398).

zwim: OK I changed that, it a peculiarity of my editor?

frontend/device/android/device.lua
```diff
@@ -375,6 +377,21 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:lightDialog()
+    local usleep = require("ffi/util").usleep
+    local title = android.isEink() and _("Frontlight settings") or _("Light settings")
+    android.settings.dialog(title, _("Brightness"), _("Warmth"), _("Ok"))
```
Frenzie, line 383: Should be OK or something more descriptive.

See http://koreader.rocks/doc/modules/ui.widget.confirmbox.html

frontend/device/android/device.lua
```diff
@@ -375,6 +377,21 @@ function Device:canExecuteScript(file)
...
+    repeat
+        usleep(75000) -- sleep 75ms before next check if dialog was quit
```
Frenzie, line 385: What's the rationale behind this value?

frontend/device/android/device.lua
```diff
@@ -375,6 +377,21 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:lightDialog()
```
Frenzie, line 380: showLightDialog might be better?

frontend/device/android/device.lua
pazos: It blocks the Lua VM while the dialog is shown. It should be a number at least x10-20 the time spent checking the condition (~1ms) and lower enough, so the user cannot perceive it as a UI block once it press the ok button. 

I think any value from 10 to 250ms is fine

Frenzie: If 10 is perfectly fine then it sounds like doubling or tripling to 20 or 30 for safety should be plenty. :-)

frontend/device/android/device.lua
pazos: It is the button that allows you to exit the dialog. See https://github.com/koreader/koreader/issues/5768#issuecomment-652443499

pazos:
Now tested :D, works quite well. I did notice that we're actually reapplying brightness / warmth instead of updating the values directly. The following patch fixes that, logs the current brightness **after** updating fl_intensity and address the rest of review comments.

Of course it wasn't tested on a warmth device, but I can imagine the logic behind fl_intensity and fl_warmth is the same.

```diff
diff --git a/frontend/device/android/device.lua b/frontend/device/android/device.lua
index a67af16c..8a1a4ac1 100644
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -377,17 +377,17 @@ function Device:canExecuteScript(file)
     end
 end
 
-function Device:lightDialog()
+function Device:showLightDialog()
     local usleep = require("ffi/util").usleep
     local title = android.isEink() and _("Frontlight settings") or _("Light settings")
     android.settings.dialog(title, _("Brightness"), _("Warmth"), _("Ok"))
     repeat
-        usleep(75000) -- sleep 75ms before next check if dialog was quit
+        usleep(25000) -- sleep 25ms before next check if dialog was quit
     until (not android.isFrontlightDialogRunning())
+    self.powerd.fl_intensity = self.powerd:frontlightIntensityHW()
     logger.info("new brightness: " .. self.powerd.fl_intensity)
-    self.powerd:setIntensityHW(self.powerd:frontlightIntensityHW())
     if android.isWarmthDevice() then
-        self.powerd:setWarmth(self.powerd:getWarmth())
+        self.powerd.fl_warmth = self.powerd:getWarmth()
         logger.info("new warmth: " .. self.powerd.fl_warmth)
     end
 end
diff --git a/frontend/ui/elements/common_settings_menu_table.lua b/frontend/ui/elements/common_settings_menu_table.lua
index df0163d0..fca66bb4 100644
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -205,7 +205,7 @@ if Device:isAndroid() then
     common_settings.frontlight = {
         text = _("Frontlight"),
         callback = function()
-            Device:lightDialog()
+            Device:showLightDialog()
         end,
     }
```

frontend/device/android/device.lua
Frenzie: For a single button it might be better to go with "Close" than "OK"?

Better UX might be "Cancel" and "OK," which is also what's on the dialog my phone pops up from settings. (Which I've probably never used because the dropdown from the top is much more relevant.)

pazos: > For a single button it might be better to go with "Close" than "OK"?

Both are fine for me :)

> Better UX might be "Cancel" and "OK," which is also what's on the dialog my phone pops up from settings.

But cancel does the same as ok/close. Changes are already applied.

A cancel button would mean discard any change done inside the dialog and revert to previous values. It is doable and can be done quite fast if the change is desiderable.

Frenzie: > A cancel button would mean discard any change done inside the dialog and revert to previous values.

Exactly. ;-)

>  if the change is desiderable.

It seems to be in line with how Android works, unless they've changed it in recent versions.

pazos: ok, done in the backend. Also updated the diff at the end of this PR.

Frenzie: I think you still need to push it? :+1:

zwim:
In 33a9e50 all discussed changes and the expanded dialog (cancel, OK) are combined. Tested on a warmth device.

frontend/device/android/device.lua
```diff
@@ -375,6 +377,31 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:showLightDialog()
+    local usleep = require("ffi/util").usleep
+    local title = android.isEink() and _("Frontlight settings") or _("Light settings")
+    android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("Ok"), _("Cancel"))
```
Frenzie, line 383:
```suggestion
    android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("OK"), _("Cancel"))
```

To be clear, this shows up as the intended
Cancel/OK
?

frontend/device/android/powerd.lua
```diff
@@ -2,16 +2,34 @@ local BasePowerD = require("device/generic/powerd")
...
+    end
+end
+
+function AndroidPowerD:setWarmth(warmth)
+    self.fl_warmth = warmth
+    android.setScreenWarmth(warmth / self.warm_diff)
+end
+
+function AndroidPowerD:getWarmth()
+   return android.getScreenWarmth() * self.warm_diff
```
Frenzie, line 32: Indent

frontend/device/android/device.lua
zwim: OK

frontend/device/android/powerd.lua
zwim: OK

PR #6438, opened 2020-07-27
Minor cleanups after #6424

NiLuJe:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6438)

### DIFF
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -420,7 +420,7 @@ end
 
 function NetworkMgr:getPowersaveMenuTable()
     return {
-        text = _("Kill Wi-Fi connection when inactive"),
+        text = _("Disable Wi-Fi connection when inactive"),
         help_text = _([[This will automatically turn Wi-Fi off after a generous period of network inactivity, without disrupting workflows that require a network connection, so you can just keep reading without worrying about battery drain.]]),
         checked_func = function() return G_reader_settings:isTrue("auto_disable_wifi") end,
         enabled_func = function() return Device:hasWifiManager() and not Device:isEmulator() end,
--- a/frontend/ui/network/networklistener.lua
+++ b/frontend/ui/network/networklistener.lua
@@ -155,26 +155,28 @@ function NetworkListener:_scheduleActivityCheck()
         end
     end
 
-    -- If we've just killed Wi-Fi, onNetworkDisconnected will take care of unscheduling us
-    if keep_checking then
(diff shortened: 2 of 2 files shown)


NiLuJe:
Quick question @pazos: I mentioned in #6424 that on Android `turnOnWifi`/`turnOffWifi` don't forward callbacks. I don't suppose there's an easy way to fix that?

(The callbacks are also used to send the NetworkConnected/NetworkDisconnected events, which makes those unreliable on Android, which break the Sync's plugin auto-sync mode).

pazos:
> turnOnWifi/turnOffWifi don't forward callbacks. 

Is worse than that. `turnOnWifi` and `turnOffWifi` don't do what their name imply. The just open android wifi settings. It could be the case that one call to `turnOnWifi` leaves the wifi off if the user returns to KOReader without toggling wifi in android settings.

>  I don't suppose there's an easy way to fix that?

Is not hard but is a PITA (as anything android for us):

1. each time you call android.openWifiSettings set a variable to `true`. It could be a variable part of `Device` or anthing in `android.something`.
2. as the activity is paused while android settings are on front and `RESUME` cmd event is called when we're back to KOReader you can check if the variable is true there, check current network state and send the relevant event if the network state changed.

pazos:
Something like that is actually (ab)used to: clean text selection after returning back from a 3rd party app, open a file from a intent and importing files from the SAF file picker.

See https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L153-L187 for everything we check on resume.

NiLuJe:
@pazos: Ah, thanks for the details.

So a cheap workaround like on Kindle might not be enough, then :/

https://github.com/koreader/koreader/blob/93eadbaf888a68a74ec92688148a5f07e0de3d55/frontend/device/kindle/device.lua#L103-L120

(Especially since the connectivity check might ultimately call turnOffWifi again if isConnect is still false after ~45s).

Oh, well :/.

pazos:
Not sure how important is that on android, because even if we don't trigger a turnOn/turnOff action other programs or the system can trigger it, so a check of current state is always needed on resume.

What we can do is:

1. keep track of last network state before calling ` android.openWifiSettings()` (and the callback in case of `NetworkMgr:turnOnWifi` and `NetworkMgr:turnOffWifi`)
2. always check on resume (async, very fast) if network is on.
3. if the state changed then update last network state. If the callback is not nil then execute it and set it to nil again.

PR #6440, opened 2020-07-27
Add Set Frontlight/Set Frontlight Warmth to dispatcher

Galunid:
Currently dispatcher has ability to change frontlight/warmth by a given number, this PR aims to allow user to change FL intensity/warmth to a new value. This is pretty useful, as it allows users to make profiles for day reading/night reading. I\m not sure if I should check all the conditions in events, since dispatcher already handles that.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6440)

### DIFF
--- a/frontend/device/devicelistener.lua
+++ b/frontend/device/devicelistener.lua
@@ -114,19 +114,18 @@ if Device:hasFrontlight() then
 
         if new_intensity == nil then return true end
         -- when new_intensity <=0, toggle light off
+	self:onSetFlIntensity(new_intensity)
+        self:onShowIntensity()
+        return true
+    end
+
+    function DeviceListener:onSetFlIntensity(new_intensity)
+        local powerd = Device:getPowerDevice()
         if new_intensity <= 0 then
             powerd:turnOffFrontlight()
         else
             powerd:setIntensity(new_intensity)
         end
-        self:onShowIntensity()
-        return true
-    end
-
-    -- Used by dispatcher only (all the conditions such as min/max frontlight value should be handled by dispatcher already)
-    function DeviceListener:onSetFlIntensity(new_intensity)
-        local powerd = Device:getPowerDevice()
-        powerd:setIntensity(new_intensity)
         return true
     end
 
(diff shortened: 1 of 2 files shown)


Galunid:
Currently setting fl intensity works, setting warmth partially works(it works in range of 0..10 on my Kobo Clara HD, instead of 0..100).

frontend/dispatcher.lua
```diff
@@ -33,8 +33,10 @@ local settingsList = {
     -- Device settings
     show_frontlight_dialog = { category="none", event="ShowFlDialog", title=_("Show frontlight dialog"), device=true, condition=Device:hasFrontlight(),},
     toggle_frontlight = { category="none", event="ToggleFrontlight", title=_("Toggle frontlight"), device=true, condition=Device:hasFrontlight(),},
+    set_frontlight = { category="incrementalnumber", event="SetFlIntensity", min=1, max=Device:getPowerDevice().fl_max, title=_("Set frontlight brightness"), device=true, condition=Device:hasFrontlight(),},
```
yparitcher, line 36: this should be `category="absolutenumber"` as this cannot be a relative number supplied by a gesture.
min should be 0 to allow setting the FL off.

frontend/dispatcher.lua
```diff
@@ -33,8 +33,10 @@ local settingsList = {
...
     increase_frontlight = { category="incrementalnumber", event="IncreaseFlIntensity", min=1, max=Device:getPowerDevice().fl_max, title=_("Increase frontlight brightness"), device=true, condition=Device:hasFrontlight(),},
     decrease_frontlight = { category="incrementalnumber", event="DecreaseFlIntensity", min=1, max=Device:getPowerDevice().fl_max, title=_("Decrease frontlight brightness"), device=true, condition=Device:hasFrontlight(),},
+    set_frontlight_warmth = { category="incrementalnumber", event="SetFlWarmth", min=1, max=Device:getPowerDevice().fl_warmth_max, title=_("Set frontlight warmth"), device=true, condition=Device:hasNaturalLight(),},
```
yparitcher, line 39: Same as above.
Also max for the warmth should be 100 as powerd takes care of scaling the number in the backend, the frontend always sees 0 .. 100
This will fix your issue with the Clara.

frontend/device/devicelistener.lua
```diff
@@ -123,6 +123,13 @@ if Device:hasFrontlight() then
         return true
     end
 
+    -- Used by dispatcher only(all the conditions such as min/max frontlight value should be handled by dispatcher already)
```
NiLuJe, line 126: nit: space before the parentheses ;).

frontend/device/devicelistener.lua
```diff
@@ -212,6 +219,13 @@ if Device:hasFrontlight() then
         return true
     end
 
+    -- Used by dispatcher only(all the conditions such as min/max frontlight warmth should be handled by dispatcher already)
```
NiLuJe, line 222: Ditto

Galunid:
Makes a lot of sense. Should be done now, I tested it on my Kobo Clara HD and it works as expected. Hopefully ready for final review and merge :)

Galunid:
If you mean why I don't call it in `onSetFlIntensity`, it's because I don't want dispatcher to display it when recalling(?) a profile. It'd make a mess if I set a few actions that display notification, i.e. setting warmth, setting intensity, turning Wi-Fi on, in one action would just display mess on my screen. This simply avoids it.

Issue #6375, opened 2020-07-13
A new "Minimal-Edition" with Removal of Gesture and Plug-in System:


JagdishKumar1970:
Hello,
I'm a non-Programmer currently using:
org.koreader.launcher(v2020.05 from F-Droid.org Repository)@Android-7.1.2;

I request the respected devs to build and maintain a separate 'Minimal-Edition' for KOReader which:
1: Remove ALL the gestures except Left, Righ, Up, Down Screen Tap which trigger page-up/-down and Options;
2: Remove the Plugin System/Framework Completely;
3: Follow Standard Android System APIs for Rotation and all other functions/features except Screen-timeout(because it is embedded deep in System Settings and not as straightforward as brightness from qs panel).

Where, Points 1 and 2 would reduce Package Size and 3 will reduce unneeded use of system sensors which will save Battery, as I found there was no way to switch off the Auto Screen_Rotation which was useless for me. 

I noticed that KOR used 2% battery/hour, which impressed me deeply.

Thank you, for atleast reading till here and Please Comment the reason about: "Why not?"

Frenzie:
> Where, Points 1 and 2 would reduce Package Size

The difference would only be a few kilobytes. Size comes from libraries and fonts. I tried to get rid of most of the fonts but unfortunately Android built-in fonts are all over the place, even though we literally depend on some of the same Noto fonts. A "you'll probably come across broken characters" edition is the only thing that would cut the size and you don't want that. ;-)

> 3 will reduce unneeded use of system sensors which will save Battery, as I found there was no way to switch off the Auto Screen_Rotation which was useless for me.

I'm not sure what you're referring to. The program doesn't waste battery listening to rotation events independently from the system. In the version you're using it can go upside down as controlled entirely by the system; in the next stable it'll be able to do that in any direction. But again, only if you enable "autorotate."

JagdishKumar1970:
> > Where, Points 1 and 2 would reduce Package Size
> 
> The difference would only be a few kilobytes. Size comes from libraries and fonts. I tried to get rid of most of the fonts but unfortunately Android built-in fonts are all over the place, even though we literally depend on some of the same Noto fonts. A "you'll probably come across broken characters" edition is the only thing that would cut the size and you don't want that. ;-)
> 
> > 3 will reduce unneeded use of system sensors which will save Battery, as I found there was no way to switch off the Auto Screen_Rotation which was useless for me.
> 
[...]

PR #6441, opened 2020-07-28
Gestures: refactor as plugin based on Dispatcher

yparitcher:
Gesture should be functionally identical except you can now add more than one action per gesture.
Also Custom multiswipes were moved to their own submenu and Multiswipe recorder was moved into it. this make handling custom multiswipes much cleaner (we don't have to cobble the mutiswipe menu)

this will make gestures much more capable, and cleaner.

Todo:
~~menus in progress~~. Done
~~multiswipe menu~~
~~migration of current settings~~
~~reverse default menu for RTL?~~

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6441)


yparitcher:
Please review.

Gesture should be functionally identical except you can now add more than one action per gesture.
Also `Custom multiswipes` were mover to their own submenu and `Multiswipe recorder` was moved into it. this make handling custom multiswipes much cleaner (we don't have to cobble the mutiswipe menu)

I did not implement migration yet, I have a plan, I just wanted a review on the implementation first.

Screenshots:
Reader_2020-07-29_165633
Reader_2020-07-29_165724
Reader_2020-07-29_165834

plugins/gestures.koplugin/main.lua
```diff
@@ -0,0 +1,1039 @@
...
+        callback = function()
+            self.gestures[ges] = nil
+        end,
+    })
+    Dispatcher:addSubMenu(sub_items, self.gestures, ges)
+    return sub_items
+end
+
+function Gestures:genSubItem(ges, separator, hold_callback)
+    local reader_only = {tap_top_left_corner=true, hold_top_left_corner=true}
```
poire-z, line 232: Is that different from previous code?

yparitcher:
> Just one thing about moving it as a plugin: some gestures are quite essential, like the early ones that were migrated from core to be a gesture like tap top right for bookmarking or Filemanager "plus menu".
> That means that if someone disables the Gesture plugin, KOReader is quite unusable. (Or may be that can be considered a feature: limited functionality ? :)

Someone actually asked for it recently :)

Galunid:
@poire-z Actually there as such FR: #6375

plugins/gestures.koplugin/main.lua
yparitcher: no, they were/are disabled in FM. (I guess as not to interfere with the home & plus button)

yparitcher:
To help
Gestures:init() changed
Gestures:gestureTitleFunc(ges) new
Gestures:genMenu(ges) changed
Gestures:genSubItem(ges, separator, hold_callback) new
Gestures:genSubItemTable(gestures) new
Gestures:genMultiswipeMenu() new
Gestures:friendlyMultiswipeName(multiswipe) moved
Gestures:safeMultiswipeName(multiswipe) moved
Gestures:multiswipeRecorder(touchmenu_instance) moved
Gestures:genCustomMultiswipeSubmenu() new
Gestures:addIntervals(menu_items) moved
Gestures:addToMainMenu(menu_items) changed
Gestures:initGesture() changed
Gestures:setupGesture(ges) moved
Gestures:registerGesture(ges, ges_type, zone, overrides, direction, distance) moved
Gestures:gestureAction(action, ges) changed
Gestures:multiswipeAction(multiswipe_directions, ges) moved
Gestures:onIgnoreHoldCorners(ignore_hold_corners) moved
Gestures:onFlushSettings() new

yparitcher:
Finished ;)

NiLuJe:
Not really familiar with the code, but at a glance, nothing jumps out, :+1: ;).

yparitcher:
>> *plugins/gestures.koplugin/_meta.lua, line 5 at r2 <https://reviewable.io/reviews/koreader/koreader/6441#-MDUbgG77tWaSzAWLCqJ:-MDUbgG8D1d7DWzrS1Wr:bui621p> (raw file <https://github.com/koreader/koreader/blob/f9ee79c54c1979379f926790a3ea4ada7e2fad83/plugins/gestures.koplugin/_meta.lua#L5>):*
>>     name = 'gestures',
    fullname = _("Gestures"),
    description = _([[This plugin allows using gestures.]]),
> provides/implements gestures?

will change

> *plugins/gestures.koplugin/defaults.lua, line 5 at r2 <https://reviewable.io/reviews/koreader/koreader/6441#-MDUc-cUBCPPHJVd9DYD:-MDUc-cX3IcMfrT0_m7K:b-g4jmox> (raw file <https://github.com/koreader/koreader/blob/f9ee79c54c1979379f926790a3ea4ada7e2fad83/plugins/gestures.koplugin/defaults.lua#L5>):*

>> return {
    ["gesture_fm"] = {
> You don't need all that boilerplate. Just `gesture_fm =` will do. Same for all of this file. Is it a user manually editing the file concern?


no, I was following the format luasettings uses to write to disk, but i see it is not needed, I will clean it up

now is a good time to add/change defaults, as existing users wont change as their old defaults will be migrated over
> *plugins/gestures.koplugin/main.lua, line 232 at r1 <https://reviewable.io/reviews/koreader/koreader/6441#-MDIKIMseXxYmwz8Uun9-r1-232:-MDUcStX7REFmNjcaI66:bvbm628> (raw file <https://github.com/koreader/koreader/blob/b7f974e170f2b823fec3d8cb216fc6472f1671e8/plugins/gestures.koplugin/main.lua#L232>):*

> *Previously, yparitcher wrote…*
>> no, they were/are disabled in FM. (I guess as not to interfere with the home & plus button)

> I did the corners one at a time, since it took a lot of time to migrate that stuff over. Plus the icon's not dynamic. There's some stuff there that can be tidied up further but it's not related to this PR. :)


a lot of functionality was added over time so i was not sure what was needed or not, so i tried keeping the same behavior but i did clean up the implementation, so it will me much easier to fix all these corner cases :)

poire-z:
Is that totally independant of the Profiles plugin? Can one disable the Profiles plugin and enable the Gestures plugin - and vice versa - and the one enabled will just work fine?

poire-z:
Also, if one wants a gesture to change 5 settings - and it already has a profile for these 5 settings:
can he just points the gesture to that profile ?
or does he have to re-create the gesture with these 5 settings ? (and then have to edit both gesture and profile when he wants to change the value of a single one of these 5 settings?)

poire-z:
OK, as I really don't want to read the code, I tested it on my device :) Some observations:
- On first launch (I guess a migration is done): no gesture work. A restart (via menu, as my usual gesture does not work), solves it.
- After this first launch & restart, it seems all my usual gestures work as expected - and bookmark and "plus button" work too.

In the gesture menu, when tapping on a gesture and seeing the actions list, it's a little less obvious than before :) where we only had a full list of what's available, and you quickly saw what's the current affected action (as it was single). Now, you have to navigate the submenus to see what it is.
- I'd like it to see it as one of the first items (default|passthought|nothing|what-it-is).
- Or may be a marker on the Device|File browser|Reflowable... menu item to see we have something enabled in which of them
- The order of the actions is not obvious (but I won't complain, as I guess it's the same as for profiles, that I helped set their order :)
- There's no separator, which could make the previous comment a little less relevant
- You can set `Go X pages` and `Previous chapter` as actions to the same gesture, haven't tested what would be the result :)
- `Go X pages`  in the action list, and alongside the gesture name that have it as its action: it could be shown as `Go 10 pages`  or `Go -20 pages`, so you know which one goes back and which goes forward :)
- A little refresh region glitch on my Kobo : when hitting `Go X pages` or `Font size`, and seeing the number picker widget, when closing it (by tapping outside or close button), the right border line is still visible (I had this in the past in possibly some other widget, don't remember how I solved it) over the kept-opened menu, when in Reader - but strangely, I don't see that line when I do that while in File browser. Seems to not happen on Reader with other pickers that keep menu open (Text editor > Set text font size).

Minor issue not related to this PR I guess that I've witnessed : when toggling (either via gesture or the menu item) nightmode, and having `Invert images` `on` in the bottom menu: on toggling (whether to/from day/night/day, the images on the current page are not as expected - and a diagonal swipe refresh doesn't change it. But going to next page and back have it correctly drawn. (pinging @NiLuJe just in case you remember something about this thing you added but don't use :)

NiLuJe:
>     * A little refresh region glitch on my Kobo : when hitting `Go X pages` or `Font size`, and seeing the number picker widget, when closing it (by tapping outside or close button), the right border line is still visible (I had this in the past in possibly some other widget, don't remember how I solved it) over the kept-opened menu, when in Reader - but strangely, I don't see that line when I do that while in File browser. Seems to not happen on Reader with other pickers that keep menu open (Text editor > Set text font size).
> 

I fixed something related to that somewhat recently, but it should only affect devices w/ hardware dithering o_O.

https://github.com/koreader/koreader/pull/6267

So, err, I'd double-check if there are float coordinates left somewhereN

> 
> Minor issue not related to this PR I guess that I've witnessed : when toggling (either via gesture or the menu item) nightmode, and having `Invert images` `on` in the bottom menu: on toggling (whether to/from day/night/day, the images on the current page are not as expected - and a diagonal swipe refresh doesn't change it. But going to next page and back have it correctly drawn. (pinging @NiLuJe just in case you remember something about this thing you added but don't use :)

Yeah, that would require re<i>painting</i> the page to take, instead of just re<i>freshing</i> it ;).

yparitcher:
> Is that totally independant of the Profiles plugin? Can one disable the Profiles plugin and enable the Gestures plugin - and vice versa - and the one enabled will just work fine?

yes

yparitcher:
> OK, as I really don't want to read the code, I tested it on my device :) Some observations:
> 
>     * On first launch (I guess a migration is done): no gesture work. A restart (via menu, as my usual gesture does not work), solves it.
> 

I can't reproduce in the emulator did you test before or after https://github.com/koreader/koreader/pull/6441/commits/f9ee79c54c1979379f926790a3ea4ada7e2fad83 ?

> In the gesture menu, when tapping on a gesture and seeing the actions list, it's a little less obvious than before :) where we only had a full list of what's available, and you quickly saw what's the current affected action (as it was single). Now, you have to navigate the submenus to see what it is.
> 
>     * I'd like it to see it as one of the first items (default|passthought|nothing|what-it-is).
> 
>     * Or may be a marker on the Device|File browser|Reflowable... menu item to see we have something enabled in which of them
>     * There's no separator, which could make the previous comment a little less relevant

These would be good features for Dispatcher (both gestures and Profiles) but doing so would require maintaining in the menu a list of what is enabled which is a big headache. :( PR welcome

>     * You can set `Go X pages` and `Previous chapter` as actions to the same gesture, haven't tested what would be the result :)

as with any conflicting actions (toggle NM & NM on etc.) whichever gets executed last (whichever order `pairs` decides to use that time). It is better to give a user a little more control and trust they won't be too stupid ;)

>     * `Go X pages`  in the action list, and alongside the gesture name that have it as its action: it could be shown as `Go 10 pages`  or `Go -20 pages`, so you know which one goes back and which goes forward :)

I agree, but what happens if the amount of pages is supposed to be based on the gesture? or is not currently set?

poire-z:
> I can't reproduce in the emulator did you test before or after f9ee79c ?

With it (with this PR as 4 hours ago) over v2020.07.1-18-g744e933 (but nothing in the 6 commits after seems relevant).

> I agree, but what happens if the amount of pages is supposed to be based on the gesture? or is not currently set?

Show a little arrow/~/symbol instead of the number ?

About the separator, didn't we talk about it in the dispatcher/profile plugin PR ? I see it mentionned in a comment but no setting has it:
https://github.com/koreader/koreader/blob/a9d0990f848ce33ebbdc67291d645c4a5296fb1c/frontend/dispatcher.lua#L28-L34

yparitcher:
> About the separator, didn't we talk about it in the dispatcher/profile plugin PR ? I see it mentionned in a comment but no setting has it:

You are right dispatcher does support separators, someone just needs to decide which ones should get one. and add them to that table

poire-z:
(As a base, wherever there is a blank line in the following :) but probably a few others could be added:)
https://github.com/koreader/koreader/blob/a9d0990f848ce33ebbdc67291d645c4a5296fb1c/frontend/dispatcher.lua#L142-L151

yparitcher:
> Show a little arrow/~/symbol instead of the number ?

Done, It is still X when no pages are selected, please suggest another glyph/symbol if you want

yparitcher:
>     * A little refresh region glitch on my Kobo : when hitting `Go X pages` or `Font size`, and seeing the number picker widget, when closing it (by tapping outside or close button), the right border line is still visible (I had this in the past in possibly some other widget, don't remember how I solved it) over the kept-opened menu, when in Reader - but strangely, I don't see that line when I do that while in File browser. Seems to not happen on Reader with other pickers that keep menu open (Text editor > Set text font size).

@NiLuJe fixed this in  #6264 for the rest of the UI but i had copied the spinwidget code before that PR and merged it after so it was not done for Dispatcher.
Should be fixed now.

yparitcher:
Updated screenshots:
Reader_2020-07-30_145900
Reader_2020-07-30_145921

yparitcher:
> In the gesture menu, when tapping on a gesture and seeing the actions list, it's a little less obvious than before :) where we only had a full list of what's available, and you quickly saw what's the current affected action (as it was single). Now, you have to navigate the submenus to see what it is.
> 
>  I'd like it to see it as one of the first items (default|passthought|nothing|what-it-is).
> 
>  Or may be a marker on the Device|File browser|Reflowable... menu item to see we have something enabled in which of them

I think I fixed everything aside from this. To implement this would require rebuilding the menu in place each time a item is selected, this would be slow and wasteful I think. We do something like this to allow adding/removing a custom multiswipe to the menu without restarting and it is not pretty lines 408 - 417:
https://github.com/koreader/koreader/blob/462dc5bfb50cdd975ad661b788eb18655e688dfd/plugins/gestures.koplugin/main.lua#L384-L424

It would be much worse for a list of enabled actions.


>  * `Go X pages`  in the action list, and alongside the gesture name that have it as its action: it could be shown as `Go 10 pages`  or `Go -20 pages`, so you know which one goes back and which goes forward :)

This is done but I am still looking for an alternate symbol instead of `X` when it is not set.

**Please do not squash when merging or github will lose the file rename and thereby its history**

plugins/gestures.koplugin/main.lua
```diff
@@ -0,0 +1,1036 @@
...
+            end,
+            overrides = overrides,
+        },
+    })
+end
+
+function Gestures:gestureAction(action, ges)
+    if G_reader_settings:isTrue("gestures_migrated") then
+        UIManager:show(InfoMessage:new{
+            text = _("Gestures have been upgraded. You may now set more than one action per gesture."),
```
yparitcher, line 981: This is a popup on the first gesture after migration to let users know about the change.

Any preference for the text?

NiLuJe: Short and to the point, no complaint here ;).

NiLuJe:
> 
> This is done but I am still looking for an alternate symbol instead of `X` when it is not set.
> 

Apparently maths says `U+2124`: ℤ

We have it in FreeSerif, so it should probably work?

NiLuJe:
> > This is done but I am still looking for an alternate symbol instead of `X` when it is not set.
> 
> Apparently maths says `U+2124` for an integer: ℤ
> 
> We have it in FreeSerif & FreeSans, so it should _probably_ work?
> 
> (Spoiler alert: I'm terrible at maths :D).

That said, a simple lowercase `n` is fine by me ;).

poire-z:
Yes, no italic. But just a regular `n` is fine.

If `Go n pages` and _n not set_ means it's gesture dependant, you could look how these do (when adding symbols, usually what's bad is that there are too bold or too thin, or vertically/horizontally badly aligned):
☇ https://www.compart.com/en/unicode/U+2607
☈ https://www.compart.com/en/unicode/U+2608

Or a `n + ◌͐` = n&#x0350;  https://www.compart.com/en/unicode/U+0350 Combining Right Arrowhead Above :)

> To implement this would require rebuilding the menu in place each time a item is selected, this would be slow and wasteful I think

Or just some text_func if we don't change the number of items ?
But ok with letting this as it is, was just mentionning some ideas :) nothing blocking.

yparitcher:
> > To implement this would require rebuilding the menu in place each time a item is selected, this would be slow and wasteful I think
> 
> Or just some text_func if we don't change the number of items ?
> But ok with letting this as it is, was just mentionning some ideas :) nothing blocking.

You are right again :) I was talking about your idea of a separate menu of the enabled items.

I added a checked_func for each of the sections that is checked if it contains a selected item. Also a hold_callback to clear all the sub selections.

If wanted I can make it a number of selected items in the text instead but i think that is added complexity for no real gain. Up to you guys.

Screenshot:
Reader_2020-07-31_030306

yparitcher:
For the number text (Go n pages): this idea can be extended to other items that are numbers, if someone wants to open a PR to change the text (I personally try to stay away from formatting user facing text as much as possible. Too many opinions ;)

yparitcher:
I would like to merge this soon, so nightly users can test and iron out any bugs before the release. are there any issues left?

I plan on opening a separate PR to extend the number text (Go n pages) idea to other gestures (backlight/fontsize) and i will decide on a final text for `n` there.

Frenzie:
I haven't found the time to properly look at it yet but if @pazos and @poire-z think it's good to go. :+1:

poire-z:
Been using it since a few days ago (without even remembering I had it), so I guess it's fine.
(Except for the first startup after it when no gestures worked - I guess we'll se if others experience that too, and then we can fix it - I can't exclude that it was caused by my own patched stuff :)
I haven't really made any new gestures nor multi-action ones thus.

poire-z:
It looks like the settings/gesture.lua is written on every book close. Might be nice to notice that gestures have not changed and avoid rewritting it when not needed.
(It's possible that other plugins do that - but as I have most of them disabled, I only care about the gestures one :) may be the profiles one needs that too.)

yparitcher:
> It looks like the settings/gesture.lua is written on every book close. Might be nice to notice that gestures have not changed and avoid rewritting it when not needed.
> (It's possible that other plugins do that - but as I have most of them disabled, I only care about the gestures one :) may be the profiles one needs that too.)

The only other option is to write them whenever a item changes which would require passing a callback to `Dispatcher` to be executed after selecting a item. This is way more complicated and much easier to break. Is it worth it?

Frenzie:
> The only other option

You could probably also do a util.tableEquals and only write if it doesn't match?

>  Is it worth it?

But that, I don't know. :-)

yparitcher:
> You could probably also do a `util.tableEquals()` and only write if it doesn't match?

Then i would have to keep around 2 tabes for the whole time and deep copy on each book startup.

poire-z:
> would require passing a callback to Dispatcher to be executed after selecting a item

Dunno how many place modify the settings tables (I guess these places don't call methods to modify stuff, they just update it), but to avoid callback, you could send an event and have OnGestureChanged() (if Dispatcher is able to receive events, dunno).

> Is it worth it?

It would be to me :) given that once set up, users rarely change their gestures.

Frenzie:
> Then i would have to keep around 2 tables for the whole time and deep copy on each book startup.

Not necessarily, but it's pretty much just a few dozen strings and booleans?

yparitcher:
> Dunno how many place modify the settings tables (I guess these places don't call methods to modify stuff, they just update it), but to avoid callback, you could send an event and have OnGestureChanged() (if Dispatcher is able to receive events, dunno).

they update it directly. Dispatcher can't receive events, and it would have to be the other way around anyway, when dispatcher finishes changing a setting it would have to either via a callback or event, cause it's caller to flush settings. A event probably wont work as dispatcher is not aware of who calls it (gestures or profiles).

I get the benefit and and also don't want it to be written on every book close, and will probably do something about this, but it will also make the saving a little more fragile (when someone tries extending this they can easily cause something not to get saved)


> Not necessarily, but it's pretty much just a few dozen strings and booleans?

not any more, it is tables of different actions/event to do

poire-z:
> dispatcher finishes changing a setting it would have to either via a callback or event, cause it's caller to flush settings

So, I guess Dispatcher is just a generic object. And Profiles and Gestures use it, but are both responsible for saving their stuff.
So, you could have Dispatcher add a key to each subtables (the top it manages) `_updated_ = true` when things have changed - and the clients that save it could just walk the settings table, notice the _updated_ keys/flags, and remove them when saving it ?

PR #6443, opened 2020-07-28
Add additional media types to OPDS Browser

gotson:
The media types for CBZ and CBR (added in #5940) are deprecated, and replaced by `vnd.comicbook+zip` and `vnd.comicbook+rar`. In addition, many OPDS servers will use the actual media type of the file, ie `application/zip` and `application/x-rar-compressed`.

See here about the new `vnd` types:
- https://www.iana.org/assignments/media-types/application/vnd.comicbook+zip
- https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6443)


gotson:
I also added `application/vnd.rar` which is replacing `application/x-rar-compressed`

Frenzie:
Thanks!

PR #6448, opened 2020-07-28
standalone osx application bundle

pazos:
Requires https://github.com/koreader/koreader-base/pull/1151

replaces the shell script used as the main executable with a proper native binary. This makes possible to use Interface Builder XIBs for the main menu.

As XIBs are just plain xml I changed the extension here.

XIBs can be translated (instructions attached in `do_mac_bundle.sh` script). There're just a few strings and it is a PITA there's no easy way to integrate them in `koreader.pot` template.

The script will try to find a `mac.strings` for each l10n directory and copy translated resources to app bundle. Files are in `xliff` format which is not plain text but weblate supports translation (it is called IOS string files).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6448)


platform/mac/do_mac_bundle.sh
```diff
@@ -135,23 +154,46 @@ install_name_tool -change ${BREW}/webp/lib/libwebp.7.dylib libs/libwebp.7.dylib
...
-if command_exists "create-dmg"; then
-    # create KOReader-$VERSION.dmg with KOReader.app inside
-    create-dmg "${APP_BUNDLE}.app" --overwrite
-    rm -rf "${APP_BUNDLE}.app"
-else
-    # rename as KOReader-$VERSION.app
-    mv -v "${APP_BUNDLE}.app" "${APP_BUNDLE}-${VERSION}.app"
-fi
+
+#if command_exists "create-dmg"; then
```
Frenzie, line 192: What's the problem with this?

platform/mac/menu.xml
```diff
@@ -0,0 +1,122 @@
...
+        <menu title="Main Menu" systemMenu="main" id="AYu-sK-qS6">
+            <items>
+                <menuItem title="KOReader" id="1Xt-HY-uBw">
+                    <modifierMask key="keyEquivalentModifierMask"/>
+                    <menu key="submenu" title="KOReader" systemMenu="apple" id="uQy-DD-JDr">
+                        <items>
+                            <menuItem title="About KOReader" id="5kV-Vb-QxS">
+                                <modifierMask key="keyEquivalentModifierMask"/>
+                                <connections>
+                                    <action selector="orderFrontStandardAboutPanel:" target="-1" id="Exp-CZ-Vem"/>
```
Frenzie, line 19: Are those `id`s purposefully random?

platform/mac/do_mac_bundle.sh
pazos: That was me commenting the create-dmg stuff to check changes faster! Will fix soon.

platform/mac/menu.xml
pazos: I created the main menu in XCode -> New -> File -> MainMenu and removed most of options that are useless to us. Then renamed `MainMenu.xib` to `menu.xml` and here is the result.

As the Interface Builder stuff isn't linked to any code by itself it should work even if we modify their id to some other value. It should be still forwarded to the first responder, which is the SDL application.

JasonInOttawa:
This is great! I did notice that hitting Cmd-F to put the application into full screen also opens a font dialog box.

pazos:
> This is great! I did notice that hitting Cmd-F to put the application into full screen also opens a font dialog box.

Which font dialog? I don't see it. Probably there's a good change of me missing something as I removed from my laptop some apple software.

The original menu had `Cmd + Alt + F` as the keyboard shortcut. I changed it to `Cmd + F` because it is easier for me to type as is what VLC uses to toggle fullscreen, but I'm totally fine to return back to stock combo if that fixes your issue 👍

JasonInOttawa:
I only see if it a book is open when I hit Cmd-F. See attached.

Chrome uses Ctrl-Cmd-F to enter fullscreen.
Screen Shot 2020-07-28 at 6 01 16 PM

platform/mac/menu.xml
pazos: changed to simpler ids: 1 .. n

JasonInOttawa:
Ctrl-Cmd-F works. 

I also noticed that the application is asking for permission to access Desktop/Documents/etc and I can set and access the Home folder in the file browser there now. Before those folders were showing as empty in the file browser and there was no prompt to access them.

pazos:
> I also noticed that the application is asking for permission to access Desktop/Documents/etc and I can set and access the Home folder in the file browser there now. Before those folders were showing as empty in the file browser and there was no prompt to access them.

Oh, that's cool. That could be something that SDL applications do when they are inside an app bundle (or maybe is apple magically?) because I didn't touch something special for this to happen :smile: 

Anyways I couldn't reproduce it before. Not sure if has something to do with having gatekeeper disabled for all apps or is because I disabled System Integrity Protection all togegher. Could you tell me your SIP/gatekeeper settings?

There's another thing that could be useful to explore. We can define app [entitlements](https://developer.apple.com/documentation/bundleresources/entitlements). But I'm not sure if these entitlements can be applied without codesign.

Do you see any error on Console.app if you filter by `koreader`?. I can only see an error saying the app is not codesigned. That's probably something than we can't address as it requires a paid Apple id.

pazos:
BTW @Frenzie: need help again to interpret circleci checks.

```
Warning: platform/mac/do_mac_bundle.sh does not abide by coding style, diff for expected style:
17c17
<  
---
>
```

:facepalm:

poire-z:
It's probably just complaining about that line with only a space - and expects empty lines to be really empty, like line 19:
<kbd>image</kbd>

pazos:
Thanks @poire-z, I'm still facepalming :)

JasonInOttawa:
In the older application bundle with the script executable, I see this error in Console:

error   08:29:22.938596-0400    sandboxd        Sandbox: luajit(52869) System Policy: deny(1) file-read-data /Users/jason/Documents/Books
Violation:       System Policy: deny(1) file-read-data /Users/jason/Documents/Books
Process:         luajit [52869]
Path:            /Applications/KOReader-2020.07.1-6.app/Contents/koreader/./luajit

My Gatekeeper setting allows applications from the App Store and Identified Developers. I haven't changed my SIP setting.

With the newer application bundle, I see the following error in Console:

error	08:40:14.383639-0400	tccd	Failed to copy signing info for 52955, responsible for file:///Applications/KOReader-2020.07.1-24.app/Contents/MacOS/koreader: #-67062: Error Domain=NSOSStatusErrorDomain Code=-67062 "(null)"

pazos:
@JasonInOttawa: thanks for the info!

> error 08:40:14.383639-0400 tccd Failed to copy signing info for 52955, responsible for file:///Applications/KOReader-2020.07.1-24.app/Contents/MacOS/koreader: #-67062: Error Domain=NSOSStatusErrorDomain Code=-67062 "(null)"

That's exactly the error I saw. Nothing to do here without an Apple Id.

I read a bit about entitlements. They are a required to skip certain limitations of hardened runtime. Hardened runtime is a requisite for notarize. We can skip all of this since the app won't be notarized nor codesigned.

> My Gatekeeper setting allows applications from the App Store and Identified Developers. I haven't changed my SIP setting.

I had the idea you need to disable gatekeeper to run untrusted apps. Did you add an exception for KOReader?. Could you test it the app in the link works for you?:  https://www.dropbox.com/s/q7180zhqq03btri/koreader%202020.06-94.dmg?dl=0

NiLuJe:
@pazos: Right Click > Open > Say yes to GateKeeper ;).

JasonInOttawa:
Opening the DMG, 
copying the application to my Applications folder, 
right-clicking and Open, 
saying yes to Gatekeeper worked

JasonInOttawa:
One nit-pick: if you resize the window, quit, and restart, the window goes back to the default size.

Frenzie:
It's a general bug/inconvenience. I imagine anything macOS-specific would be handled by SDL internally.

pazos:
> It's a general bug/inconvenience. I imagine anything macOS-specific would be handled by SDL internally.

Totally agree. This affects all sdl platforms with floating windows and shouldn't be coded as something mac specific. Sadly the codepath is quite convulted to fix it.

The function that sets the window size and position is [`SDL.open`](https://github.com/koreader/koreader-base/blob/master/ffi/SDL2_0.lua#L81-L121). It uses environment variables or fallback to 800 x 600.

One idea would be used width, height, x and y as arguments of `SDL.open` and change the code to:

```lua
width or tonumber(os.getenv("EMULATE_READER_W")) or 600
height or tonumber(os.getenv("EMULATE_READER_H")) or 800
x or tonumber(os.getenv("KOREADER_WINDOW_POS_X")) or SDL.SDL_WINDOWPOS_UNDEFINED
y or tonumber(os.getenv("KOREADER_WINDOW_POS_Y")) or SDL.SDL_WINDOWPOS_UNDEFINED
```

But the codepath to set arguments there makes impossible this without refactoring:

`SDL.open` is part of the input subsystem, which starts in https://github.com/koreader/koreader-base/blob/master/ffi/input.lua and goes to https://github.com/koreader/koreader-base/blob/master/ffi/input_SDL2_0.lua

So the other option would be: in isDesktop devices other than the emulator (and maybe appImage if we want to use it as a self contained emulator)  set these environment variables before starting the program, keep track of last values, updated each time a SDL move/resize event happens, and save them to disk in `Device:exit()`. That's not really hard but we need to figure out a way to do it :thinking:

JasonInOttawa:
I know this is out of scope for this PR, but is this at all helpful:

https://developer.apple.com/documentation/appkit/nswindow/1526255-restorable

pazos:
> I know this is out of scope for this PR, but is this at all helpful:

Hi @JasonInOttawa, thanks for your suggestions. As @Frenzie said all these specific platform quirks are handled by SDL internally. After all we're not using cocoa direclty. We're using SDL on all desktop systems (which is using the cocoa backend on mac, but that's another topic).

The [proposed SDL patch](https://github.com/koreader/koreader-base/pull/1151/commits/ebe27cc3df2e44eb3488796d453d64532fee1936#diff-74f21b5b1b5c03385cef018a22135004) just removes a couple of workarounds SDL developers added to make possible to run apps without an app bundle. We use this to run the emulator on a mac but we don't need it for the application bundle. It also hardcodes a couple of strings to open a few urls and the license from the mac menu, because SDL doesn't has an API for this. For all the rest SDL has an API (an awesome API, indeed) and we should use that instead of hacking our own way.

Please read my previous comment if you want to know a bit more about how all this can be handled using SDL for all supported platforms, including mac. Since you already improved a few bits of our sdl usage I hope you find it pretty easy to understand.

JasonInOttawa:
Thanks. I'll try to figure things out.

PR #6449, opened 2020-07-28
Allow user to set higher hold pan rate

### LINKED ISSUE #6381: Question: Is it possible to make highlighting more responsive?
Today, out of curiosity I tried Plato for the first time and highlighting there is way more responsive. I'm not sure how it works in KOReader, but I wonder if it's possible to achieve a similar effect. While it could have problems with ghosting, I believe it would make a huge difference on android.

Galunid:
This PR adds new menu entry called `Hold pan rate` to `Taps and Gestures -> Gesture intervals`. This allows users to increase speed at which highlighting is refreshed, instead of the default 2 refreshes per second. Discussed partially in #6381.
closes #6381

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6449)

### DIFF
--- a/plugins/gestures.koplugin/main.lua
+++ b/plugins/gestures.koplugin/main.lua
@@ -429,6 +429,36 @@ function Gestures:addIntervals(menu_items)
         text = _("Gesture intervals"),
         sub_item_table = {
             {
+                text = _("Text selection rate"),
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local current_value = G_reader_settings:readSetting("hold_pan_rate")
+                    if not current_value then
+                        current_value = Screen.low_pan_rate and 5.0 or 30.0
+                    end
+                    local items = SpinWidget:new{
+                        text = T(_([[
+Used when highlighting text. Higher values mean faster screen updates, but also use more CPU.
+Default value: %1]]), current_value),
+                        width = math.floor(Screen:getWidth() * 0.6),
+                        value = current_value,
+                        value_min = 1.0,
+                        value_max = 60.0,
+                        value_step = 1,
+                        value_hold_step = 15,
+                        ok_text = _("Set rate"),
+                        title_text = _("Hold pan rate"),
+                        default_value = Screen.low_pan_rate and 5.0 or 30.0,
+                        callback = function(spin)
+                            G_reader_settings:saveSetting("hold_pan_rate", spin.value)
+                            UIManager:broadcastEvent(Event:new("UpdateHoldPanRate"))
+                        end
+                    }
+                    UIManager:show(items)
+                end,
+                separator = true,
+            },
+            {
                 text = _("Double tap interval"),
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
@@ -531,35 +561,6 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_pan_delayed_interval", spin.value*1000)
                             GestureDetector:setNewInterval("ges_pan_delayed_interval", spin.value*1000)
-                        end
-                    }
-                    UIManager:show(items)
-                end,
-            },
-            {
-                text = _("Hold pan rate"),
-                callback = function()
-                    local SpinWidget = require("ui/widget/spinwidget")
-                    local current_value = G_reader_settings:readSetting("pan_rate")
-                    if not current_value then
-                        current_value = Screen.low_pan_rate and 5.0 or 30.0
-                    end
-                    local items = SpinWidget:new{
-                        text = T(_([[
-Used when highlighting text. Higher values mean faster screen updates, but also use more CPU.
-Default value: %1]]), current_value),
-                        width = math.floor(Screen:getWidth() * 0.6),
-                        value = current_value,
-                        value_min = 1.0,
-                        value_max = 60.0,
-                        value_step = 1,
-                        value_hold_step = 5,
-                        ok_text = _("Set rate"),
-                        title_text = _("Hold pan rate"),
-                        default_value = Screen.low_pan_rate and 5.0 or 30.0,
-                        callback = function(spin)
-                            G_reader_settings:saveSetting("hold_pan_rate", spin.value)
-                            UIManager:broadcastEvent(Event:new("UpdateHoldPanRate"))
                         end
                     }
                     UIManager:show(items)
--- a/plugins/gestures.koplugin/main.lua
+++ b/plugins/gestures.koplugin/main.lua
@@ -438,7 +438,10 @@ function Gestures:addIntervals(menu_items)
                     end
                     local items = SpinWidget:new{
                         text = T(_([[
-Used when highlighting text. Higher values mean faster screen updates, but also use more CPU.
+Used when selecting the text.
+The rate is the number of times per second screen will be refreshed,
+while selecting the text (for translating/highlighting/wikipedia search).
+Higher values mean faster screen updates, but also use more CPU.
 Default value: %1]]), current_value),
                         width = math.floor(Screen:getWidth() * 0.6),
                         value = current_value,
--- a/plugins/gestures.koplugin/main.lua
+++ b/plugins/gestures.koplugin/main.lua
@@ -441,7 +441,7 @@ function Gestures:addIntervals(menu_items)
 Used when selecting text.
 The rate is how often screen will be refreshed per second while selecting text.
 Higher values mean faster screen updates, but also use more CPU.
-Default value: %1]]), current_value),
+Default value: %1]]), Screen.low_pan_rate and 5.0 or 30.0),
                         width = math.floor(Screen:getWidth() * 0.6),
                         value = current_value,
                         value_min = 1.0,
--- a/plugins/gestures.koplugin/main.lua
+++ b/plugins/gestures.koplugin/main.lua
@@ -546,7 +546,7 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
                     end
                     local items = SpinWidget:new{
                         text = T(_([[
-Used when highlighting text, higher value means faster refresh, but uses more CPU.
+Used when highlighting text. Higher values mean faster screen updates, but also use more CPU.
 Default value: %1]]), current_value),
                         width = math.floor(Screen:getWidth() * 0.6),
                         value = current_value,
--- a/plugins/gestures.koplugin/main.lua
+++ b/plugins/gestures.koplugin/main.lua
@@ -3,6 +3,7 @@ local ConfirmBox = require("ui/widget/confirmbox")
 local DataStorage = require("datastorage")
 local Device = require("device")
 local Dispatcher = require("dispatcher")
+local Event = require("ui/event")
 local FFIUtil = require("ffi/util")
 local Geom = require("ui/geometry")
 local GestureRange = require("ui/gesturerange")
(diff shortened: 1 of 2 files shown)


Galunid:
Not sure if there is a way to call `function ReaderHighlight:setupTouchZones()` again safely, to set the new value without restarting the device.  `function ReaderHighlight:setupTouchZones()` calls `self.ui:registerTouchZones`, with the argument I'm interested in
```lua        
        {
            id = "readerhighlight_hold_pan",
            ges = "hold_pan",
            rate = hold_pan_rate,
            screen_zone = {
                ratio_x = 0, ratio_y = 0, ratio_w = 1, ratio_h = 1,
            },
            handler = function(ges) return self:onHoldPan(nil, ges) end
        },
```

poire-z:
I think we made it safe, for this:
https://github.com/koreader/koreader/blob/fc35c6acc856ed41c433496a8748a8b0460f5ec5/frontend/apps/reader/modules/readerrolling.lua#L1304-L1306
So, you could just call `self.ui.highlight:setupTouchZones()` when you have `rate` updated - on add a `ReaderHighlight:onHoldPanRateUpdate()` that would do it in there.

NiLuJe:
Yeah, I recall having done it for... something I don't remember right now (rotation?) a while back ;).

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -631,6 +631,36 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
...
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local current_value = G_reader_settings:readSetting("pan_rate")
+                    if not current_value then
+                        current_value = Screen.low_pan_rate and 2.0 or 30.0
+                    end
+                    local items = SpinWidget:new{
+                        text = T(_([[
+Used when highlighting text, bigger value faster refresh and bigger CPU usage.
+5.0 should be good for most e-ink ereaders.
```
Frenzie, line 645: If so, why is 2 the default?

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -631,6 +631,36 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
...
+                text = _("Hold pan rate"),
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local current_value = G_reader_settings:readSetting("pan_rate")
+                    if not current_value then
+                        current_value = Screen.low_pan_rate and 2.0 or 30.0
+                    end
+                    local items = SpinWidget:new{
+                        text = T(_([[
+Used when highlighting text, bigger value faster refresh and bigger CPU usage.
```
Frenzie, line 644:
```suggestion
Used when highlighting text, bigger values refresh faster use more CPU.
```

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -631,6 +631,36 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
...
+5.0 should be good for most e-ink ereaders.
+Default value: %1]]), current_value),
+                        width = math.floor(Screen:getWidth() * 0.6),
+                        value = current_value,
+                        value_min = 1.0,
+                        value_max = 30.0,
```
Frenzie, line 650: Or 60? Not for E Ink of course but on desktop or a regular Android device.

Galunid: Sure, I can change that, but tbh I don't see a point. 30 gives 1 refresh every ~33ms. I'd say 30 and 60 don't have noticeable difference.

frontend/apps/reader/modules/readergesture.lua
Galunid: To be fair, that's a good question. It was default in `readerhighlight.lua` so I kept it that way. I think it's because older devices may have trouble handling higher value, it was kept low just in case. My Kobo Clara HD had no troubles handling 10, but there was no difference between 10 and 5, because of screen refresh(black thingy).

frontend/apps/reader/modules/readergesture.lua
NiLuJe: Matches the non-gaming standard of 60Hz monitors ;).

frontend/apps/reader/modules/readergesture.lua
Frenzie: In any case, it's weird to say that x is a good value and then not to have x as the default.

frontend/apps/reader/modules/readergesture.lua
Frenzie: Perfectly fine as a default, annoying as an artificial limitation. ;-)

NiLuJe: Oh, right, didn't notice this was the max, not the default ;).

poire-z:
Anything blocking this?
Just the decision to move the default for eInk from 2 to 5 ? (Tested 5 on my GloHD, and it works and is speedy - some possible white->black->grey refreshes/animations when you jump multiple lines fast that are surprising - but indeed, no feeling of slowness.)

yparitcher:
this will need to be rebased after #6441

frontend/apps/reader/modules/readergesture.lua
```diff
@@ -631,6 +631,36 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
                     UIManager:show(items)
                 end,
             },
+            {
```
yparitcher, line 634: to add this to the redone gestures plugin it should be put here:
https://github.com/koreader/koreader/blob/9e0ac734baf704f708c19150f41442ead7745c09/plugins/gestures.koplugin/main.lua#L538

Frenzie:
@poire-z My (fairly minor) comments still stand.

Galunid:
I was waiting for #6441 and reinstalling the system, it should be done in the next 2-3 days.

Galunid:
Should be ready for final review.

poire-z:
Re-tested, 2 issues:
- In the gesture interval menu, Hold pan rate shows 30 as the starting value on my Kobo - because `default_value = Screen.has_low_pan_rate and 5.0 or 30.0,` and `has_low_pan_rate` does not exist :)
- A bit confusing that it advertizes as a rate in the menu item (5, 30...) while the other gesture type in that "Gesture interval" menu advertizes as an interval (duration in ms) - so their meaning/direction of growth are opposite

plugins/gestures.koplugin/main.lua
```diff
@@ -535,6 +536,36 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
...
+                text = _("Hold pan rate"),
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local current_value = G_reader_settings:readSetting("pan_rate")
+                    if not current_value then
+                        current_value = Screen.has_low_pan_rate and 2.0 or 30.0
+                    end
+                    local items = SpinWidget:new{
+                        text = T(_([[
+Used when highlighting text, higher value means faster refresh, but uses more CPU.
```
Frenzie, line 549:
```suggestion
Used when highlighting text. Higher values mean faster screen updates, but also use more CPU.
```

plugins/gestures.koplugin/main.lua
```diff
@@ -535,6 +536,36 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
...
+5.0 should be good for most e-ink ereaders.
```
Frenzie, line 550:
```suggestion
```

plugins/gestures.koplugin/main.lua
```diff
@@ -535,6 +536,36 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
...
+Default value: %1]]), current_value),
+                        width = math.floor(Screen:getWidth() * 0.6),
+                        value = current_value,
+                        value_min = 1.0,
+                        value_max = 60.0,
+                        value_step = 1,
+                        value_hold_step = 5,
+                        ok_text = _("Set rate"),
+                        title_text = _("Hold pan rate"),
+                        default_value = Screen.has_low_pan_rate and 5.0 or 30.0,
```
Frenzie, line 560:
```suggestion
                        default_value = Screen.low_pan_rate and 5.0 or 30.0,
```

Galunid:
@poire-z 
First issue - already fixed
Second issue - I kept it this way, because there are already `Hold Interval`/`Pan Interval` menu entries. 
Having `Hold Pan Interval` would make it more confusing in my opinion. 
Perhaps I can name it `Hold Pan Interval(while highlighting)` and change it to [ms] instead of rate? I'm fine with either.

poire-z:
Given that it seems it is only used by readerhighlight, you could indeed name it:
`Text selection rate / interval`
hiding the "hold pan" technical term?

I dunno what's best:
- name it "interval", and display/save the value 1000/x
- keep naming it "rate", so it's nearer to how the implementation - but in this case, may be just move it in the menu: put it below the others and have a separator=true to separate it from the other "interval" above ? With its name "Text selection rate", it would be different enough to justify the separator.
(or put it first above all the others instead of below?

Galunid:
Thanks, renaming menu entry to `Text selection rate` is great idea. I'm more inclined towards second option. Currently max value is at 60[rate] which gives us 16.67[ms] interval. Spin widget had trouble dealing with floating point numbers(not sure if it's a bug), so we cannot achieve all the values that are possible otherwise. Alternatively we could set new max to 40[rate] -> 25[ms], but NiLuJe mentioned that 60[rate] matches monitor refresh rate. If there is a good way to deal with this that I'm not seeing, I'm open to suggestions.

poire-z:
I'm fine with this.
May be a more explicite description text:

_Set text selection rate (for highlighting, translating...). This rate is the number of time per second the selection will be updated as you move your finger.
Higher values mean faster screen updates, but also use more CPU._

(should we not use _finger_, as ona can use a mouse?)

NiLuJe:
IIRC, the SpinWidget *UI* just rounds the displayed value, but the setting being input manually is kept full-precision.

(I *think*, haven't checked).

Galunid:
```
Used when selecting the text.
The rate is the number of times per second screen will be refreshed,
while selecting the text (for translating/highlighting/wikipedia search).
Higher values mean faster screen updates, but also use more CPU.
```

@poire-z is this acceptable?

poire-z:
The meaning is fine with me.
But I'm not that good with English, so I let @Frenzie and @NiLuJe review the phrasing.

Frenzie:
Used when selecting text.

The rate is how often screen will be refreshed per second while selecting text.

Higher values mean faster screen updates, but also use more CPU.

NiLuJe:
Could even be shortened to `Screen refresh rate while selecting text` ;).

plugins/gestures.koplugin/main.lua
```diff
@@ -438,9 +438,8 @@ function Gestures:addIntervals(menu_items)
                     end
                     local items = SpinWidget:new{
                         text = T(_([[
-Used when selecting the text.
-The rate is the number of times per second screen will be refreshed,
-while selecting the text (for translating/highlighting/wikipedia search).
+Used when selecting text.
+The rate is how often screen will be refreshed per second while selecting text.
 Higher values mean faster screen updates, but also use more CPU.
 Default value: %1]]), current_value),
```
poire-z, line 444: Default value: %1 = current_value ?
The other items seems to show there the default value, so I guess you should too. (`Screen.low_pan_rate and 5.0 or 30.0`,)

plugins/gestures.koplugin/main.lua
```diff
@@ -427,6 +428,38 @@ function Gestures:addIntervals(menu_items)
...
+Higher values mean faster screen updates, but also use more CPU.
+Default value: %1]]), Screen.low_pan_rate and 5.0 or 30.0),
+                        width = math.floor(Screen:getWidth() * 0.6),
+                        value = current_value,
+                        value_min = 1.0,
+                        value_max = 60.0,
+                        value_step = 1,
+                        value_hold_step = 15,
+                        ok_text = _("Set rate"),
+                        title_text = _("Hold pan rate"),
```
Frenzie, line 452: Was this still supposed to say hold pan rate or should it also be text selection rate?

poire-z: Probably should.
And then remove the first line of the text :) 

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`Text selection rate`
```diff
< Used when selecting text.
< The rate 
> This rate
```

Galunid: Yes, I'll fix it tomorrow, I forgot to change it

PR #6453, opened 2020-07-30
CoverBrowser: delay GC for quicker browsing/exiting

poire-z:
Slow browsing and exiting (ex: closing History) could be noticed when the memory usage is already high (as the GC has a lot of memory to visit).
Delaying it to after the new screen is shown limits the feeling of slowness.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6453)


NiLuJe:
Sounds good ;).

PR #6456, opened 2020-07-30
bump crengine: reworked hanging punctuation

### LINKED ISSUE #2844: Rendering issues when floating punctuation is enabled

### LINKED ISSUE #6235: Hanging Punctuation does not work, only the hyphen hangs.
* KOReader version: 2020.05
* Device: Tolino Epos 2, Linux-Emulator

#### Issue
Hanging punctuation does not work. Only the "hanging" of the hyphen can be  toggled.
If you want hanging punctuation, you want all small characters (e.g. . , -  : ; ) to hang a bit  - and not only the hyphen.
I do prefer hanging punctuation (at least in German) and I am using it in my publications.

Please see the pictures:

Here the Hanging Punctuation is **turned off**. Note the hyphens and commata
Hanging Punctuation turned off

Here the Hanging Punctuation is **turned on**. Note the hyphens and commata
Hanging Punctuation turned on

## Solution
At least for German it is possible to activate the hanging of punctuation marks by the following patch

```
--- work/lvtextfm.cpp	2020-06-05 08:29:31.557364010 +0200
+++ koreader/base/thirdparty/kpvcrlib/crengine/crengine/src/lvtextfm.cpp	2020-06-05 08:34:38.299067829 +0200
@@ -2987,12 +2986,13 @@
                         if ( word->flags & LTEXT_WORD_CAN_HYPH_BREAK_LINE_AFTER ) {
                             word->width -= font->getHyphenWidth(); // TODO: strange fix - need some other solution
                         }
-                        else if ( lastc=='.' || lastc==',' || lastc=='!' || lastc==':' || lastc==';' || lastc=='?') {
+                        else if ( lastc=='.' || lastc==',' || lastc=='!' || lastc==':' || lastc==';' || lastc=='-' || lastc=='<' /*|| lastc==0xAB *//*'«'*/ ) {
                             FONT_GUARD
-                            int w = font->getCharWidth(lastc);
+                            int w = (font->getCharWidth(lastc)+1)/2; // half of the character-width, rounded up integer divison no truncation
                             TR("floating: %c w=%d", lastc, w);
-                            if (frmline->width + w + wAlign + x >= maxWidth)
+//                            if (frmline->width + w + wAlign + x >= maxWidth)  // this line has to be removed!!!!
                                 word->width -= w; //fix russian "?" at line end
+                                
                         }
                         else if ( lastc==0x2019 || lastc==0x201d ||   // ’ ” right quotation marks
                                   lastc==0x3001 || lastc==0x3002 ||   // 、 。 ideographic comma and full stop
```

Some remarks to the patch:
1. I do not want the punctuation marks with their full character width. Instead it looks better if they hang only half of their width. This is done with
```
                          int w = (font->getCharWidth(lastc)+1)/2; 
```
"+1)/2" is for rounding the integer division up!!

2. The equation
```
 (frmline->width + w + wAlign + x >= maxWidth)  
```
is always false (at least for my German books). So remove it (at least in German) and hanging punctuation works.

At last two pictures with

**Optimized Hanging Punctuation:** I love this!
HangingHyphenAndOptimizedHangingPunctuation

**Full Hanging Punctuation:** does not look so good
HangingHyphenAndHangingPunctuationFull

poire-z:
Discussion in https://github.com/koreader/crengine/pull/355.
It has become more some kind of "optical margin alignment" than "fully hanged punctuation".
Some comments and links about the difference in https://github.com/koreader/crengine/issues/307.
Closes #2844. Closes #6235.

I have some frontend patch that moves `Hanging punctuation` from the style first menu panel into the `Typography rule>` sub-menu, where I guess it has a better place alongside `Hyphenation`.
Will need an additional tap to reach it, but it leaves a bit more room for the book text to see its effect.
<kbd>image</kbd>
Opinions?

I still have a bug (that was there before moving the menu item) to solve: setting enabled as the default has it enabled on a new book, but the menu item is unchecked - will look at it later this evening or tomorrow.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6456)


poire-z:
OK, bug fixed. Settings are a bit of non-standard - but I don't want to think about it or any kind of migration :) so it's pure copy & pasta.

NiLuJe:
Definitely agree on the move, in fact, I was going to suggest moving somewhere more appropriate than where it used to be ;).

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -718,6 +763,15 @@ function ReaderTypography:onReadSettings(config)
     self.ui.document:setHyphLeftHyphenMin(G_reader_settings:readSetting("hyph_left_hyphen_min") or 0)
     self.ui.document:setHyphRightHyphenMin(G_reader_settings:readSetting("hyph_right_hyphen_min") or 0)
 
+    -- Default to disable hanging/floating punctuation
+    -- (Stored as 0/1 in docsetting for historical reasons, but as true/false
+    -- in global settings.)
```
NiLuJe, line 768: That's the "reason" behind the funky setting.

It's a really old quirk, so much so that I don't remember exactly why/when that quirk was implemented ^^.

poire-z:
(Pfff, github glitch dropped my clean squashed message...)

poire-z:
Some issue when hanging punctuation is enabled: if you bookmark (dogear) a page whose first word is hanging in the left margin (any word starting with any of `AJTVWXYvwxy`), the dogear will not show when you revisit this page.

To be fixed in cre.cpp https://github.com/koreader/koreader-base/blob/2e53592573718a59399df1739a92b39cf8d92e2d/cre.cpp#L3478-L3489
`page_rect.isRectInside(xp_rect)` will fait when the xp (first char box), as it hangs in the left margin, overflows the page rect:
```
page range: 0,127910 - 1072,129243
xp range: -1,127910 - 0,127953   
```
I guess a y-only check should be enough. (But may be a drop cap that overflows the page top might fail too...)

PR #6466, opened 2020-08-02
Oops, too much Python/Lua ^^.

NiLuJe:
This passes shellcheck, for some reason, but it really doesn't work in busybox...

(I *probably* tested this in ZSH and didn't think twice about it at the time).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6466)

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -230,7 +230,7 @@ ko_do_fbdepth() {
 # Fun fact: this wouldn't be necessary if Kobo were using a non-prehistoric glibc... (it was fixed in glibc 2.26).
 ko_do_dns() {
     # If there aren't any servers listed, append CloudFlare's
-    if not grep -q '^nameserver' "/etc/resolv.conf"; then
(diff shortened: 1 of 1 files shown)


platform/kobo/koreader.sh
```diff
@@ -230,7 +230,7 @@ ko_do_fbdepth() {
 # Fun fact: this wouldn't be necessary if Kobo were using a non-prehistoric glibc... (it was fixed in glibc 2.26).
 ko_do_dns() {
     # If there aren't any servers listed, append CloudFlare's
-    if not grep -q '^nameserver' "/etc/resolv.conf"; then
```
Frenzie, line 233: It's perfectly valid syntax after all. You're running the program `not` with some arguments. ;-P

NiLuJe: Indeed ^^.

And either ZSH has a builtin for it, or I checked that locally, where LLVM does ship one which probably do the obvious ^^.

PR #6468, opened 2020-08-03
gesture: messages and sensitivity changes for small (warmth) scales; devicelistener dedup

zwim:
deduplicate code on warmth/brightness swipes
    
If device has a small scale (e.g. warmth is going from 0..10)
    * adjust warmth message,
    * use a smaller gesture sensitivity.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6468)

### DIFF
--- a/frontend/device/devicelistener.lua
+++ b/frontend/device/devicelistener.lua
@@ -10,7 +10,7 @@ local _ = require("gettext")
 local T = require("ffi/util").template
 
 local DeviceListener = InputContainer:new{
-    steps_fl = { 0.1, 0.1, 0.2, 0.4, 0.7, 1.1, 1.6, 2.2, 2.9, 3.7, 4.6, 5.6, 6.7, 7.9, 9.2, 10.6, },
+
 }
 
 function DeviceListener:onToggleNightMode()
@@ -47,10 +47,17 @@ end
 function DeviceListener:onShowWarmth(value)
     local powerd = Device:getPowerDevice()
     if powerd.fl_warmth ~= nil then
-        UIManager:show(Notification:new{
-            text = T(_("Warmth set to %1."), powerd.fl_warmth),
+        if powerd.fl_warmth_max == nil then
+            UIManager:show(Notification:new{
+                text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth)),
+                timeout = 1.0,
+            })
+        else
+            UIManager:show(Notification:new{
+            text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth/100*powerd.fl_warmth_max)),
             timeout = 1.0,
-        })
+            })
+        end
     end
     return true
 end
@@ -58,13 +65,20 @@ end
 -- frontlight controller
 if Device:hasFrontlight() then
 
-    -- direction +1 - increase frontlight
-    -- direction -1 - decrease frontlight
-    function DeviceListener:onChangeFlIntensity(ges, direction)
-        local powerd = Device:getPowerDevice()
+    local function calculateGestureDeltaIntervall(ges, direction, min, max)
         local delta_int
-        --received gesture
         if type(ges) == "table" then
+            -- here we are using just two scales
+            -- big scale is for high dynamic ranges (e.g. brightness from 1..100)
+            --           original scale maybe tuned by hand
+            -- small scale is for lower dynamic ranges (e.g. warmth from 1..10)
+            --           scale entries are calculated by math.round(1*sqrt(2)^n)  
+            local steps_fl_big_scale = { 0.1, 0.1, 0.2, 0.4, 0.7, 1.1, 1.6, 2.2, 2.9, 3.7, 4.6, 5.6, 6.7, 7.9, 9.2, 10.6, }
+            local steps_fl_small_scale = { 1.0, 1.0, 2.0, 3.0, 4.0, 6.0, 8.1, 11.3 }
+            local steps_fl = steps_fl_big_scale
+            if ( min - max < 50 ) then
+                steps_fl = steps_fl_small_scale
+            end
             local gestureScale
             local scale_multiplier
             if ges.ges == "two_finger_swipe" then
@@ -75,6 +89,7 @@ if Device:hasFrontlight() then
             else
                 scale_multiplier = 1
             end
+
             if ges.direction == "south" or ges.direction == "north" then
                 gestureScale = Screen:getHeight() * scale_multiplier
             elseif ges.direction == "west" or ges.direction == "east" then
@@ -85,17 +100,17 @@ if Device:hasFrontlight() then
                 -- diagonal
                 gestureScale = math.sqrt(width * width + height * height) * scale_multiplier
             end
-            if powerd.fl_intensity == nil then return false end
 
             local steps_tbl = {}
-            local scale = (powerd.fl_max - powerd.fl_min) / 2 / 10.6
-            for i = 1, #self.steps_fl, 1 do
-                steps_tbl[i] = math.ceil(self.steps_fl[i] * scale)
+            local scale = (max - min) / steps_fl[#steps_fl] / 2 -- full swipe gives half scale
+            for i = 1, #steps_fl, 1 do
+                steps_tbl[i] = math.ceil(steps_fl[i] * scale)
             end
 
             if ges.distance == nil then
                 ges.distance = 1
             end
+
             local step = math.ceil(#steps_tbl * ges.distance / gestureScale)
             delta_int = steps_tbl[step] or steps_tbl[#steps_tbl]
         else
@@ -106,8 +121,19 @@ if Device:hasFrontlight() then
             -- set default value (increase frontlight)
             direction = 1
         end
-        local new_intensity = powerd.fl_intensity + direction * delta_int
+        return direction, delta_int
+    end
 
+    -- direction +1 - increase frontlight
+    -- direction -1 - decrease frontlight
+    function DeviceListener:onChangeFlIntensity(ges, direction)
+        local powerd = Device:getPowerDevice()
+        local delta_int
+        --received gesture
+
+        direction, delta_int = calculateGestureDeltaIntervall( ges, direction, powerd.fl_min, powerd.fl_max)
+
+        local new_intensity = powerd.fl_intensity + direction * delta_int
         if new_intensity == nil then return true end
         -- when new_intensity <=0, toggle light off
         self:onSetFlIntensity(new_intensity)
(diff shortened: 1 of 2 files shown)


zwim:
@yparitcher: I am sorry, I don't know, how your commit did get in my commit. If I should remove #6467 from this commit, can you please give me some advice on how to do it.

frontend/device/devicelistener.lua
```diff
@@ -10,7 +10,7 @@ local _ = require("gettext")
 local T = require("ffi/util").template
 
 local DeviceListener = InputContainer:new{
-    steps_fl = { 0.1, 0.1, 0.2, 0.4, 0.7, 1.1, 1.6, 2.2, 2.9, 3.7, 4.6, 5.6, 6.7, 7.9, 9.2, 10.6, },
+
```
Frenzie, line 13: No real need for a random newline. ^_^
```suggestion
```

frontend/device/devicelistener.lua
```diff
@@ -51,24 +51,38 @@ end
...
-        UIManager:show(Notification:new{
-            text = T(_("Warmth set to %1."), powerd.fl_warmth),
+        if powerd.fl_warmth_max == nil then
+            UIManager:show(Notification:new{
+                text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth)),
+                timeout = 1.0,
+            })
+        else
+            UIManager:show(Notification:new{
+            text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth/100*powerd.fl_warmth_max)),
```
Frenzie, line 61:
```suggestion
                text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth/100*powerd.fl_warmth_max)),
```

frontend/device/devicelistener.lua
```diff
@@ -51,24 +51,38 @@ end
...
             timeout = 1.0,
```
Frenzie, line 62:
```suggestion
                timeout = 1.0,
```

frontend/device/devicelistener.lua
```diff
@@ -51,24 +51,38 @@ end
...
         if type(ges) == "table" then
+            -- here we are using just two scales
+            -- big scale is for high dynamic ranges (e.g. brightness from 1..100)
+            --           original scale maybe tuned by hand
+            -- small scale is for lower dynamic ranges (e.g. warmth from 1..10)
+            --           scale entries are calculated by math.round(1*sqrt(2)^n)  
+            local steps_fl_big_scale = { 0.1, 0.1, 0.2, 0.4, 0.7, 1.1, 1.6, 2.2, 2.9, 3.7, 4.6, 5.6, 6.7, 7.9, 9.2, 10.6, }
+            local steps_fl_small_scale = { 1.0, 1.0, 2.0, 3.0, 4.0, 6.0, 8.1, 11.3 }
+            local steps_fl = steps_fl_big_scale
+            if ( min - max < 50 ) then
```
Frenzie, line 83: Unless you write this, those parentheses don't do anything to improve clarity.
```suggestion
            if (min - max) < 50 then
```

frontend/device/devicelistener.lua
```diff
@@ -51,24 +51,38 @@ end
...
 end
 
 -- frontlight controller
 if Device:hasFrontlight() then
 
-    -- direction +1 - increase frontlight
-    -- direction -1 - decrease frontlight
-    function DeviceListener:onChangeFlIntensity(ges, direction)
-        local powerd = Device:getPowerDevice()
+    local function calculateGestureDeltaIntervall(ges, direction, min, max)
```
Frenzie, line 72: interval has only one l, but I find this an odd function name. It seems to be about the delta of distance moved; where does an interval come in?
```suggestion
    local function calculateGestureDeltaInterval(ges, direction, min, max)
```

frontend/device/devicelistener.lua
zwim: OK

frontend/device/devicelistener.lua
zwim: OK

frontend/device/devicelistener.lua
zwim: OK

frontend/device/devicelistener.lua
zwim: OK, double 'l' is German. I think Interval is not so bad, as the function calculates the delta from the gesture move width and maps this to min-max.
Which name would you suggest?

frontend/device/devicelistener.lua
zwim: OK

NiLuJe:
You'll want to rebase manually and force push. Assuming you've setup an `upstream` remote:

```
git fetch upstream
git rebase -i upstream/master
<interactive rebase here>
<checking that stuff looks sane here, i.e., git whatchanged, git diff upstream/master)
git push -f
```

frontend/device/devicelistener.lua
Frenzie: Since you seem to be using them as a tautology, just delta (e.g., calculateGestureDelta).

zwim: OK, I will use that.

frontend/device/devicelistener.lua
```diff
@@ -110,8 +123,19 @@ if Device:hasFrontlight() then
...
+    end
 
+    -- direction +1 - increase frontlight
+    -- direction -1 - decrease frontlight
+    function DeviceListener:onChangeFlIntensity(ges, direction)
+        local powerd = Device:getPowerDevice()
+        local delta_int
+        --received gesture
+
+        direction, delta_int = calculateGestureDelta( ges, direction, powerd.fl_min, powerd.fl_max)
```
Frenzie, line 136:
```suggestion
        direction, delta_int = calculateGestureDelta(ges, direction, powerd.fl_min, powerd.fl_max)
```

frontend/device/devicelistener.lua
```diff
@@ -9,9 +9,7 @@ local bit = require("bit")
 local _ = require("gettext")
 local T = require("ffi/util").template
 
-local DeviceListener = InputContainer:new{
-    steps_fl = { 0.1, 0.1, 0.2, 0.4, 0.7, 1.1, 1.6, 2.2, 2.9, 3.7, 4.6, 5.6, 6.7, 7.9, 9.2, 10.6, },
-}
+local DeviceListener = InputContainer:new{}
```
Frenzie, line 12: I'm fine with this, but it's not necessarily what I was saying.

```lua
-- all that's strictly necessary
local DeviceListener = InputContainer:new{}
```

```lua
-- reduced diff if you add something in the future
local DeviceListener = InputContainer:new{
}
```

```lua
-- random new line, that'll be two lines on the diff rather than one if you add something
local DeviceListener = InputContainer:new{

}
```

Frenzie:
Anyway, seems fine at a glance but @pazos knows more about the subject.

frontend/device/devicelistener.lua
```diff
@@ -51,24 +49,38 @@ end
...
-            timeout = 1.0,
-        })
+        if powerd.fl_warmth_max == nil then
+            UIManager:show(Notification:new{
+                text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth)),
+                timeout = 1.0,
+            })
+        else
+            UIManager:show(Notification:new{
+                text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth/100*powerd.fl_warmth_max)),
```
pazos, line 59: I don't understand this one.

What is the intention? if fl_warmth == fl_warmth_max the result will be 0.01. Am I missing something? Sorry but I didn't sleep very well :p

pazos: ok I added parenthesis :p. I guess. I still don't understand but maybe is my bad

zwim: It is annoying if you have a slider (gesture) showing values from 0..100, but the warmth only changes on 0,10,20,30...100.
In this PR the message of the gesture should be changed, if the devices `fl_warmth_max - fl_warmth_min` is much smaller than the **default warmth range (0..100)**.

Internally nothing is changed. The whole frontend works with the default warmth range (0..100). It always applies that **0<=warmth<=100**

The only changes are (only on smaller scaled devices):
1.) The Notification will show values from fl_warmth_min..fl_warmth_max: e.g. with warmth = 58 and fl_warmth_max=10 then floor(58/100*10) -> 5
2.) The scale for swiping must be greater, if `fl_warmth_max - fl_warmth_min` is smaller than the original 100. I find the new `steps_fl_small_scale` very comfortable, but it is not engraved in stone.

Kobos and other warmth devices are not touched by this PR, up to now it will only affect android e-Ink devices.

frontend/device/devicelistener.lua
zwim: OK, I don't want to add anything.

frontend/device/devicelistener.lua
NiLuJe: I'm not *quite* fond of showing the internal implementation detail only in *some* places like this (i.e., specifically talking about 1.) here).

I think it makes much more sense is we *always* display a [0, 100] scale to users, no matter the internal implementation.

Now, that's coming from a Kobo, where the stock UI doesn't actually show numbers, just a progress bars.
If the stock UI itself displays those numbers, that's another kettle of fish, and I'd tend to agree with you, with a caveat: make it consistent ;).

(I have no qualms about tweaking the swipe tweak, although I've never found it necessary myself on Kobo, FWIW).

pazos: > I think it makes much more sense is we always display a [0, 100] scale to users, no matter the internal implementation.

Me too, but I understand the idea behind the change. I'm also a bit of lover of little aesthetics ultraoptimizations :p. In that case my Bq does the same as a Tolino or a Forma: it shows 0 - 100 but just changes warmth each 10 steps. I'm fine with current behaviour but a 1 - 10 message won't bother me either.

In any case I think the logic has some flaws. I will review in a moment.

frontend/device/devicelistener.lua
```diff
@@ -51,24 +49,38 @@ end
 function DeviceListener:onShowWarmth(value)
     local powerd = Device:getPowerDevice()
     if powerd.fl_warmth ~= nil then
-        UIManager:show(Notification:new{
-            text = T(_("Warmth set to %1."), powerd.fl_warmth),
-            timeout = 1.0,
-        })
+        if powerd.fl_warmth_max == nil then
```
pazos, line 52: This codepath is never reached. All devices with powerd.fl_warmth have powerd.fl_warmth_max too (the android has it since this very PR).

frontend/device/devicelistener.lua
```diff
@@ -51,24 +49,38 @@ end
...
+            UIManager:show(Notification:new{
+                text = T(_("Warmth set to %1."), math.floor(powerd.fl_warmth)),
+                timeout = 1.0,
+            })
+        else
```
pazos, line 57: This code path would affect some devices with a different scale. All of them are kobos. Not sure which of them but KOA1 for sure.

frontend/device/devicelistener.lua
pazos, line 59: `math.floor(powerd.fl_warmth/100*powerd.fl_warmth_max))` is difficult to understand without making obvious that fl_warmth and fl_warmth_max are using different scales.

frontend/device/devicelistener.lua
zwim: Thank you for the clarification. This check was just for safety, because I am not familiar with the Kobo-stuff at all. 
We can drop it.

frontend/device/devicelistener.lua
zwim: @pazos Just for understanding: I suppose that your Bq is a phone? 

@pazos and @NiLuJe: What shall we do? A scale from `0-100` or `0-wl_warmth_max`. If we have very few steps, I would prefer the small scale. But anyway, if the decission will be `0-100`, shouldn't we add a %-symbol to the number (and also to the brightness number)?

Frenzie: The regular frontlight adjusts its steps, doesn't it? It has a 100 (?) on my Kobo.
Screenshot_2020-08-05_08-44-16

frontend/device/devicelistener.lua
zwim: You are right. Then I it would be good, if someone with an affected Kobo could test it.

frontend/device/devicelistener.lua
zwim: Would a comment be enough for clarification?
```
-- powerd.fl_warmth holds the warmth-value in the internal koreader scale [0,100]
-- powerd.fl_warmth_max is the maximum value the hardware accepts
```

zwim: @frenzie: Is this the original "Frontlight dialog" on your Kobo?

Frenzie: No, on my Kobo it's 0-100. This is a scale from 0-10 or so as on the emulator and presumably some older Kindle that it's mimicking. Which is my point, that it adapts.

pazos: Yep, the frontlight widget adapts it (the same on android). Here we're talking about the little message that pops up when you perform a warmth change with a gesture. Right now is 0 - 100. With this change is 0 - 10. It is fine in devices with 10  warmth steps but devices with more warmth steps need more resolution.

Solution? I don't know. I'm fine with the current behaviour and I'm fine too adding a % symbol to it.

zwim: > ... It is fine in devices with 10 warmth steps but devices with more warmth steps need more resolution.

The popup shows values between 0 and `fl_warmth_max`. So it adapts the resolution itself.

NiLuJe: Huh, I was convinced the FL widget was always showing a [0,100], my bad ;).

zwim:
Nothing new just a rebase, and renamed PR

zwim:
@pasos @frenzie:
I just wanted to inform you, that the native warmth dialog #6426 (as well as this commit) works on the "**Tolino Epos 2**" and on the "**Tolino Vision 4HD**". Maybe this should be added to the changelog of the upcoming 2008.8 milestone?

PR #6471, opened 2020-08-03
Dispatcher Gestures Profiles: only flush settings when changed

yparitcher:
Should work, did not fully test every corner case.

one side effect, this invalidates a profile named `_updated`. it will fail to be created. (if a user had such a profile they will see very weird behavior)

based on @poire-z idea in https://github.com/koreader/koreader/pull/6441#issuecomment-668220743

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6471)

### DIFF
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -338,7 +338,7 @@ function Dispatcher:addItem(caller, menu, location, settings, section)
                         else
                             location[settings][k] = true
                         end
-                        caller._updated = true
+                        caller.updated = true
                         if touchmenu_instance then touchmenu_instance:updateItems() end
                    end,
                    separator = settingsList[k].separator,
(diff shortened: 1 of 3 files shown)


poire-z:
Looks allright.
May be you can make the key a bit more complicated, `__updated__flag__` .
Also dunno if all this could be delegated to/hidden in Dispatcher.
```lua
dispatcher:setUpdated(settings_table) -- set the flag
if dispatcher:isUpdated(settings_table) then -- check if flag exists
  -- strip flags
  ready_to_be_saved_settings = dispatcher:getSettingsToBeSaved(settings_table)
end
```

yparitcher:
> Also dunno if all this could be delegated to/hidden in Dispatcher.

there are settings that can be set on the gesture/profile side (new custom multiswipe) so that might not be the best idea.

I tried a slightly different approach do you think it is better or worse?

it does not affect the settings table at all, but it requires passing the callers context.

poire-z:
Yes, the new approach looks cleaner. So, no more risk of conflict with a profile/gesture name ? If no risk, you might give it a better name or lose the leading `_`

yparitcher:
> So, no more risk of conflict with a profile/gesture name ? 

correct

PR #6472, opened 2020-08-03
Dispatcher: show selected number

yparitcher:
followup to #6441

screenshots with alternative ideas for what to use when no number is selected.
Reader_2020-08-03_192159
Reader_2020-08-03_192233

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6472)


NiLuJe:
`#` as a placeholder for a number *may* be less common outside of the US, but it's fine by me ;).

I could possibly see `<n>` being used there, too, but I don't have a strong preference for either.

frontend/dispatcher.lua
```diff
@@ -312,8 +312,10 @@ function Dispatcher:getNameFromItem(item, location, settings)
     if location[settings] ~= nil and location[settings][item] ~= nil then
         amount = location[settings][item]
     end
-    if amount == nil or amount == 0 then
-        amount = "X"
+    if amount == nil
+        or (amount == 0 and settingsList[item].category == "incrementalnumber")
+    then
+        amount = "#"
```
Frenzie, line 318: > # as a placeholder for a number may be less common outside of the US, but it's fine by me ;).

Just leave it to the translators. ;-)
```suggestion
        amount = C_("Number placeholder", "#")
```

NiLuJe: Huh, that's cool :).

frontend/dispatcher.lua
```diff
@@ -312,8 +313,10 @@ function Dispatcher:getNameFromItem(item, location, settings)
     if location[settings] ~= nil and location[settings][item] ~= nil then
         amount = location[settings][item]
     end
-    if amount == nil or amount == 0 then
-        amount = "X"
+    if amount == nil
+        or (amount == 0 and settingsList[item].category == "incrementalnumber")
+    then
+        amount = C_("Number placeholder", "#")
```
Frenzie, line 319: A comment on `Go %1 pages`, not sure otoh if that's related to this specifically but it does bring up a point about not being too clever with constructing strings programmatically.
Screenshot_2020-08-05_12-07-02

yparitcher: this does not change that.

I see no simple way to fix that as we don't know the value originally, and would need to return a function for the title instead of just a string.

Frenzie: > and would need to return a function for the title instead of just a string.

I'll take your word for it that it's somewhat complicated since I don't want to look into it right now (or this weekend at 36°…), but that part sounds rather easy. ;-)

PR #6483, opened 2020-08-07
[chore] Proper plurals in CoverBrowser processing

Frenzie:
As reported by @lescheck on Weblate

> I think this should be in plural form (because of second line)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6483)

### DIFF
--- a/plugins/coverbrowser.koplugin/bookinfomanager.lua
+++ b/plugins/coverbrowser.koplugin/bookinfomanager.lua
@@ -890,7 +890,7 @@ Do you want to prune the cache of removed books?]]
         end
     end
     UIManager:close(info)
-    info = InfoMessage:new{text = T(_("Processed %1 / %2 books.\n%3 extracted successfully."), nb_done, nb_files, nb_success)}
+    info = InfoMessage:new{text = T(_("Processed %1 / %2 books."), nb_done, nb_files) .. "\n" .. T(N_("One extracted successfully.", "%1 extracted successfully."), nb_success)}
     UIManager:show(info)
 end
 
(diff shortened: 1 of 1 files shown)


Frenzie:
@poire-z I didn't really know how to trigger the message for testing. Still don't. ;-)

poire-z:
I mean: is that normal ?:
`T(N_("One extracted successfully.", "%1 extracted successfully."), nb_success)`
to impose the 1/many distinction from english to all languages - and limit that to this 2 variant?
Will they be able to get "Uno extracted successfully", "Deuces extractodino successfully" and "%1 extradomultidino successfullo" ?

Frenzie:
Yes, English is just taken as the default. The function selects the right plural from the options. In practice that ranges from… 0 to 5, possibly?

poire-z:
> I didn't really know how to trigger the message for testing. Still don't. ;-)

File browser's "plus" menu I think, and "Extract and cache book information".

poire-z:
So, "One extracted successfully." won't be proposed as-is to be translated ?
Only "%1 extracted successfully." will, where the translator can decide about the variations he needs?

Frenzie:
It's presented in a language-appropriate manner, with the relevant number of fields for possible plurals.

Screenshot_2020-08-07_20-57-45

Screenshot_2020-08-07_20-59-19

Frenzie:
Chinese doesn't have any, so for better or worse translators aren't bothered with the singular option at all:

Screenshot_2020-08-07_21-02-45

Frenzie:
(German/English/Dutch/Spanish/French are the languages I've set up to be displayed simultaneously since they can inform my translation if they got around to it before me.)

PR #6489, opened 2020-08-09
Kosync network callback

Galunid:
I found that KOSync doesn't work with #6424 and since it was pretty simple I fixed it.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6489)

### DIFF
--- a/plugins/kosync.koplugin/KOSyncClient.lua
+++ b/plugins/kosync.koplugin/KOSyncClient.lua
@@ -1,4 +1,5 @@
 local UIManager = require("ui/uimanager")
+local NetworkMgr = require("ui/network/manager")
 local DEBUG = require("dbg")
 
 local KOSyncClient = {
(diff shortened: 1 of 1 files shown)


plugins/kosync.koplugin/KOSyncClient.lua
```diff
@@ -1,4 +1,5 @@
 local UIManager = require("ui/uimanager")
+local NetworkMgr = require("ui/network/manager")
```
Frenzie, line 2: nit

NiLuJe:
That plug-in should already have an onNetworkConnected handler, so I'm not quite sure what this is attempting to fix?

Android? (Because the events are unreliable there).

I'd much rather fix the events and/or the handler than adding a hack in this specific plug-in ;).

(On mobile, so context is a bit gnarly to grok).

Galunid:
@NiLuJe It was supposed to fix it on my Kobos, tested it before(latest nightly) on both and none worked ;/
I took a look at https://github.com/koreader/koreader/blob/66e70fcbbecef77b5da8d456f1b25d99a52894fc/plugins/kosync.koplugin/main.lua#L297-L300
I assumed you missed those two functions. As for `onNetworkConnected`, the only thing it does is https://github.com/koreader/koreader/blob/66e70fcbbecef77b5da8d456f1b25d99a52894fc/plugins/kosync.koplugin/main.lua#L695-L697
`onNetworkManager` calls `_onNetworkManager`, which in turn calls `_onResume`. Pushing pulling progress with Wi-Fi off basically results in a massage like `Something went wrong, check network`.

NiLuJe:
I'm not in a position to actually review this for another week, but then again, I don't actually use the plugin, so if it works for you, that's fine enough by me ;).

PR #6495, opened 2020-08-10
sdl window position and size

pazos:
Requires https://github.com/koreader/koreader-base/pull/1157

Applies to all sdl targets that are not fullscreen. Environment overrides still take precedence over saved settings.

It can be tested on the emulator by removing `EMULATE_READER_W` and `EMULATE_READER_H` from [kodev](https://github.com/koreader/koreader/blob/master/kodev#L707)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6495)

### DIFF
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -145,7 +145,9 @@ function Device:init()
     end
 
     self.hasClipboard = yes
-    self.screen = require("ffi/framebuffer_SDL2_0"):new{device = self, debug = logger.dbg,
+    self.screen = require("ffi/framebuffer_SDL2_0"):new{
+        device = self,
+        debug = logger.dbg,
         w = self.window.width,
         h = self.window.height,
         x = self.window.left,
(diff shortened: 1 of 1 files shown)


frontend/device/sdl/device.lua
```diff
@@ -29,6 +29,23 @@ local function getLinkOpener()
     return false
 end
 
+local function getWindowSizeAndPos()
+    local overrides = {
+        top = "KOREADER_POS_Y",
```
pazos, line 34: KOREADER_WINDOW_POS_Y :facepalm:

frontend/device/sdl/device.lua
```diff
@@ -144,7 +145,12 @@ function Device:init()
     end
 
     self.hasClipboard = yes
-    self.screen = require("ffi/framebuffer_SDL2_0"):new{device = self, debug = logger.dbg}
+    self.screen = require("ffi/framebuffer_SDL2_0"):new{device = self, debug = logger.dbg,
```
Frenzie, line 148: Probably clearer to stick them all on a separate line. In my five-minute hack I didn't bother with formatting. ;-)

Issue #6501, opened 2020-08-13
ReaderRolling: gotoPercent is broken


yparitcher:
* KOReader version: master
* Device: emulator

#### Issue

the number keys are bound to `gotoPercent`:
 1 goes to beginning, 2 11% etc. 
this is broken as currently 1 goes to page 1, 2 to 2 etc.

also `gotoPercent` currently only works in scroll mode not in page mode.

Cre.cpp has a `gotoPercent` function which i tried using instead of `gotoPos` but then nothing happens in page or scroll mode.

reported by @mendelmunkis

poire-z:
By changing this 10000 to 100 (why was it 10000 ?! may be some other callers expect that, dunno, haven't checked), it works (but only in scroll mode):
https://github.com/koreader/koreader/blob/d935ca69373487ae985b4133c96f08e18a5c49d0/frontend/apps/reader/modules/readerrolling.lua#L898-L900

I guess go to pos/percent were only aimed at being used in scroll mode (when there is no page, the % of doc height is the only position availlable).

Might need some tweaking to make it work in page mode.
But except for these number keys, I don't think it has any use in page mode. Or make the number keys go to page number a % of total pages - or let it guess in some way on which page is that % of doc height (might need some cre.cpp help for that).

(Not much interested in digging into that...)

PR #6503, opened 2020-08-14
ReaderRolling: fix gotoPercent

yparitcher:
Before it only worked in scroll mode not page mode & it used 10000 instead of 100%

thanks @poire-z 

Fixes: #6501

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6503)

### DIFF
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -896,7 +896,11 @@ function ReaderRolling:_gotoPos(new_pos, do_dim_area)
 end
 
 function ReaderRolling:_gotoPercent(new_percent)
-    self:_gotoPos(new_percent * self.ui.document.info.doc_height / 10000)
+    if self.view.view_mode == "page" then
+        self:_gotoPage(new_percent * self.ui.document:getPageCount() / 100)
+    else
+        self:_gotoPos(new_percent * self.ui.document.info.doc_height / 100)
+    end
 end
 
 function ReaderRolling:_gotoPage(new_page, free_first_page)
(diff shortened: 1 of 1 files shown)


poire-z:
There's this:
https://github.com/koreader/koreader/blob/d935ca69373487ae985b4133c96f08e18a5c49d0/frontend/apps/reader/modules/readerrolling.lua#L187-L191

I don't know what value this old setting might have had, but if it's like that other one:
`["percent_finished"] = 0.037974683544304,` meaning 3%, this could have been the reason for the 10000.
So, may be just do a `self:_gotoPercent(last_per * 100)` here ?

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -188,7 +188,7 @@ function ReaderRolling:onReadSettings(config)
     --- @fixme remove this branch with migration script
     elseif last_per then
         self.setupXpointer = function()
-            self:_gotoPercent(last_per)
+            self:_gotoPercent(last_per * 100)
```
Galunid, line 191: Is there a point in multiplying by 100 here, just to divide by 100 in `ReaderRolling:_gotoPercent`?

poire-z: It's rarely used stuff, and intuitively, a gotoPercent(percent) function would accept a number between 0 and 100 - even if our internal handling expect one between 0 and 1.
We could have the public `ReaderRolling:onGotoPercent(percent)` accept 0-100 and do a `/ 100` and give a 0-1 to the private `ReaderRolling:_gotoPercent(new_percent)`, but it is still named "Percent", so that would be privately confusing :)
But well, rarely used stuff, *100 and then /100 won't cost much.
(I'm sure we have bits in the code with "percent" that required 0-1, but well :)

PR #6510, opened 2020-08-16
Support for Apache WebDAV module

themanifold:
This change to the parser in `cloudstorage.lua` adds support for the [Apache WebDAV module][1]

It was manually tested using the [bytemark/webdav][2] docker container.

I developed this in Windows, with a docker container that hosted an extracted AppImage and a VNC server that I viewed using a VNC client.

I will write up my work flow at a later point.

Changes have not been tested against other webdav servers (what was this originally tested against?). Please could someone test against other webdav servers?

I also noticed a logic inversion error where we were looking for a slash at the end of the URL and if it exists, then we explicitly set `has_trailing_slash=false` - so I fixed it to set to `true`. I had to do this so that we weren't visiting the URL without a trailing slash - apache sends back a 301 redirect with a `location` header with a trailing slash, if you don't put a trailing slash.

As a side note, I think we should consider replacing this regexp pattern matching parser with the [XML parser in the newsreader plugin][3]

Also, I've never used Lua before.

[1]: https://httpd.apache.org/docs/2.4/mod/mod_dav.html
[2]: https://github.com/BytemarkHosting/docker-webdav
[3]: https://github.com/koreader/koreader/blob/master/plugins/newsdownloader.koplugin/lib/xml.lua

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6510)

### DIFF
--- a/frontend/apps/cloudstorage/webdavapi.lua
+++ b/frontend/apps/cloudstorage/webdavapi.lua
@@ -98,17 +98,11 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
         return nil
     end
 
-    local res = table.concat(sink)
-    --[[
-    Apache Webserver WebDAV mod_dav module adds this fragment to the XML tags
-    It also uses upper casing for the XML tags, amongst other things, which we
-    deal with in the pattern matching that follows
-    --]]
-    local res_data = res:gsub(' xmlns:lp1="DAV:"','')
+    local res_data = table.concat(sink)
 
     if res_data ~= "" then
         -- iterate through the <d:response> tags, each containing an entry
-        for item in res_data:gmatch("<[dD]:response>(.-)</[dD]:response>") do
+        for item in res_data:gmatch("<[^:]*:response[^>]*>(.-)</[^:]*:response>") do
             --logger.dbg("WebDav catalog item=", item)
             -- <d:href> is the path and filename of the entry.
             local item_fullpath = item:match("<[dD]:href>(.*)</[dD]:href>")
@@ -127,7 +121,7 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
                         type = "folder",
                     })
                 end
-            elseif item:find("<[dDlp1]+:resourcetype/>") and (DocumentRegistry:hasProvider(item_name)
+            elseif item:find("<[^:]*:resourcetype/>") and (DocumentRegistry:hasProvider(item_name)
                 or G_reader_settings:isTrue("show_unsupported")) then
                 table.insert(webdav_file, {
                     text = item_name,
--- a/frontend/apps/cloudstorage/webdavapi.lua
+++ b/frontend/apps/cloudstorage/webdavapi.lua
@@ -105,14 +105,14 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
         for item in res_data:gmatch("<[^:]*:response[^>]*>(.-)</[^:]*:response>") do
             --logger.dbg("WebDav catalog item=", item)
             -- <d:href> is the path and filename of the entry.
-            local item_fullpath = item:match("<[dD]:href>(.*)</[dD]:href>")
+            local item_fullpath = item:match("<[^:]*:href[^>]*>(.*)</[^:]*:href>")
             if string.sub( item_fullpath, -1 ) == "/" then
                 item_fullpath = string.sub( item_fullpath, 1, -2 )
             end
             local is_current_dir = self:isCurrentDirectory( item_fullpath, address, path )
             local item_name = util.urlDecode( FFIUtil.basename( item_fullpath ) )
             local item_path = path .. "/" .. item_name
-            if item:find("<[dD]:collection/>") then
+            if item:find("<[^:]*:collection/>") then
                 item_name = item_name .. "/"
                 if not is_current_dir then
                     table.insert(webdav_list, {
--- a/frontend/apps/cloudstorage/webdavapi.lua
+++ b/frontend/apps/cloudstorage/webdavapi.lua
@@ -99,7 +99,7 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
     end
 
     local res = table.concat(sink)
-    --[[ 
+    --[[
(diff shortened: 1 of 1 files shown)


frontend/apps/cloudstorage/webdavapi.lua
```diff
@@ -95,20 +98,27 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
         return nil
     end
 
-    local res_data = table.concat(sink)
+    local res = table.concat(sink)
+    --[[
+    Apache Webserver WebDAV mod_dav module adds this fragment to the XML tags
+    It also uses upper casing for the XML tags, amongst other things, which we
```
Frenzie, line 104: Hm? XML is case sensitive. Judging by a quick peek at https://tools.ietf.org/html/rfc4918 only `<D>` is correct. Pinging @yparitcher

themanifold: OK. Makes me further wonder what WebDAV server this code was originally written against.

Frenzie: See https://github.com/koreader/koreader/commit/a3a17dbbeb7bcd94702c2f12acf1e09809340b50#diff-fff66d39dd02c73d9bc768e0e65b97eb

> Tested with:
NextCloud
HubZilla

poire-z: I think that in a XML tag with a namespace nickname/prefix like `<BlaH:response>`, the `response` part must be in the case specificed in the schema. The namespace nickname part `BlaH` can be in any case as long as it's the same that was specified in the XML document with `xmlns:BlaH="DAV:"` or `xmlns:BlaH="http://something"`
So, logically, you should either fetch the nickname/prefix specified to strip it from tags.
But may be in our context (we know it's a webdav response), just accept any nickname/prefix, not just `D` or `d` or `lp1`, and even the absence of any prefix (we should accept `<response>`) (but skip that absence case if it complicates thing and if all webdav server use a prefix).

Frenzie: @poire-z 
Thanks, the part I was looking at actually didn't include a full XML document, though even if D were the default of course you could do xmlns:d="D:". But there's a complete sample included further down.

It's probably easier/more fruitful to toss in that real XML parser than to mess about with Lua patterns though.

themanifold: Agreed that we should use a real XML parser over Lua patterns. When I get a bit more time, I will test this against NextCloud and HubZilla (unless someone else gets to it before me).

Whoever implements the real XML parser should probably include some functional tests too, so we don't have to manually test every code change.

(I'm already using this on my kobo touch so I know it works.)

yparitcher: @Frenzie I have no clue about XML :) if thats why you pinged me. did you intend for @y-muller who added the webDAV code?

Frenzie: @yparitcher D'oh, must've misclicked. But thanks to @poire-z's observation we've already figured it out. :-)

themanifold:
Are we thinking that we can approve this PR? If not, what will it take?

Frenzie:
It's fine as a quick mini-workaround but I was under the impression that you wanted to update it with a better solution.

poire-z:
Can you try just using this kind of pattern for all of them (and remove your `res:gsub(' xmlns:lp1="DAV:"','')`):

`<[^:]*:response[^>]*>(.*)</[^:]*:response>`

which should work with any namespace prefix, and any tag attribute:
```lua
> s="<blah:response z=y>toto</blah:response>"
> print(s:match("<[^:]*:response[^>]*>(.*)</[^:]*:response>"))
toto
```

Frenzie:
True, while the disadvantage is that there might be multiple namespaces… that's already a problem.

poire-z:
> True, while the disadvantage is that there might be multiple namespaces… that's already a problem.

Not sure I understand what you mean.
But we're parsing a response from a webdav server, so no matter the namespaces, a `<blah:response>` or a `<blue:href>` or a `<blip:collection>` will most probably be webdav stuff :) So, we'll match them all with little risk of catching unrelated stuff.

themanifold:
> 
> 
> Can you try just using this kind of pattern for all of them (and remove your `res:gsub(' xmlns:lp1="DAV:"','')`):
> 
> `<[^:]*:response[^>]*>(.*)</[^:]*:response>`
> 
> which should work with any namespace prefix, and any tag attribute:
> 
> ```lua
> > s="<blah:response z=y>toto</blah:response>"
> > print(s:match("<[^:]*:response[^>]*>(.*)</[^:]*:response>"))
> toto
> ```

@poire-z This is done, though there seems to be a special meaning behind `(.-)` in Lua. `(.*)` does not work.

poire-z:
Indeed. https://www.lua.org/pil/20.2.html
> Like `*´, the modifier `-´ also matches zero or more occurrences of characters of the original class. However, instead of matching the longest sequence, it matches the shortest one

Please update similarly the other ones (unless there is a reason not to?):
```lua
local item_fullpath = item:match("<[dD]:href>(.*)</[dD]:href>")
[...]
if item:find("<[dD]:collection/>") then
[...]
elseif item:find("<[^:]*:resourcetype/>") and (DocumentRegistry:hasProvider(item_name)
```

themanifold:
> Please update similarly the other ones (unless there is a reason not to?):

@poire-z Done, and thanks for the pointer on the regex.

themanifold:
> I tried and gave up understanding the has_trailing/leading_slash as it feels like a mess.
> Ex:
> 
> ```lua
> -- too many slashes, remove one
> path = string.sub( path, 1 ) -- doesn't do what it says
> ```
> 
> ```
> > print(string.sub("/toto", 1))
> /toto
> ```

Agreed that it's a mess. If I get round to rewriting this using an XML parser, I will likely revist the trailing/leading slash too.

y-muller:
The original WebDAV code was tested against Nextcloud and Hubzilla. It's been a while since I looked at it so I can't comment from memory. And unfortunately it is going to be at least a week until I have any time.

Issue #2478, opened 2017-01-05
feature request: PDF "read"-flow


Tobster85:
* KOReader version: latest
* Device: any

#### Issue
I am thinking of enhancing the PDF reading experience. Reading very individually designed PDFs  is quite hard at the moment because of the need to constantly zoom in and out and move the already enlarged view.

![1 page](https://cloud.githubusercontent.com/assets/22470848/21678198/11f47c88-d33e-11e6-8d29-14ffd09bdfa3.png)

A possible solution could be to select the flow of reading as path **before** actually reading the PDF page. Please see the attached image:

![2 first frame](https://cloud.githubusercontent.com/assets/22470848/21678209/180a7c58-d33e-11e6-80ba-8365af532319.png)
![3 second frame](https://cloud.githubusercontent.com/assets/22470848/21678213/1a95ac04-d33e-11e6-830a-a49692055ba7.png)
![5 final](https://cloud.githubusercontent.com/assets/22470848/21678460/16c82b5a-d33f-11e6-9f6c-6ad96fd92598.png)

Frame 1 and 4 are supposed to be shown in portrait mode.

And then enjoying the read flow:
![6 read flow](https://cloud.githubusercontent.com/assets/22470848/21678243/44fc83c8-d33e-11e6-968b-1a9032501fc8.png)

Frenzie:
Self-evident to a person, but I have no idea how this could be automated (at least without gobs of data and processing power in a machine learning context).

Tobster85:
Hi Frenzie,
this feature does not need to be automated, as the read"flow" is generated by user input before reading the "flow".

Frenzie:
Still not trivial but theoretically that's definitely doable. :+1:

KenMaltby:
So you want to have the user able to "map" a sequence of areas (user defined, for each page of the PDF) that a "next page" gesture would have display full screen, in the order specified.

Tobster85:
Exactly! This could be a solution for reading complex layouted and large formated PDF files (like magazines) on our small screen e-readers.
If I can support you with more details of the idea, please let me know.

Frenzie:
I think the bigger problem is finding something with both the time and the will to specifically implement something like this. :-)

houqp:
@Tobster85 do you expect the user to manually tag the flow for each page?

One thing would be nice though is to upload all these manually labeled flows to our server so we can train an machine learned model ;)

Tobster85:
[...]

PR #6511, opened 2020-08-16
Comics: zoom to panel

Galunid:
This pull requests aims to provide convenient way to zoom in comics. The idea is when user holds/double taps (not decided yet) on a manga/comic panel, it gets cut out from the rest of the image and zoomed. More details in https://github.com/koreader/koreader-base/issues/1148. Depends on https://github.com/koreader/koreader-base/pull/1159

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6511)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -583,6 +583,21 @@ function ReaderHighlight:onPanelZoom(arg, ges)
     local hold_pos = self.view:screenToPageTransform(ges.pos)
     local res = self.ui.document:getPanelFromPage(hold_pos.page, ges)
     logger.dbg("result: ", res)
+    logger.dbg(hold_pos)
+    local image = self.ui.document:getImageFromPosition(self.hold_pos, true)
+
+    if rendered_page then
+        local img = rendered_page.bb:copy()
+        local ImageViewer = require("ui/widget/imageviewer")
+        local imgviewer = ImageViewer:new{
+            image = img,
+            -- title_text = _("Document embedded image"),
+            -- No title, more room for image
+            with_title_bar = false,
+            fullscreen = true,
+        }
+        UIManager:show(imgviewer)
+    end
     return true
 end
 
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -620,7 +620,6 @@ end
 function ReaderHighlight:onTogglePanelZoomSetting(arg, ges)
     if not self.document.info.has_pages then return end
     self.panel_zoom_enabled = not self.panel_zoom_enabled
-    self:onSaveSettings()
 end
 
 function ReaderHighlight:onPanelZoom(arg, ges)
@@ -644,7 +643,7 @@ function ReaderHighlight:onPanelZoom(arg, ges)
 end
 
 function ReaderHighlight:onHold(arg, ges)
-    if self.panel_zoom_enabled and self.document.info.has_pages then
+    if self.document.info.has_pages and self.panel_zoom_enabled then
         return self:onPanelZoom(arg, ges)
     end
 
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -99,10 +99,12 @@ function ReaderHighlight:addToMainMenu(menu_items)
         text = _("Highlighting"),
         sub_item_table = self:genHighlightDrawerMenu(),
     }
-    menu_items.panel_zoom_options = {
-        text = _("Panel zoom (manga/comic)"),
-        sub_item_table = self:genPanelZoomMenu(),
-    }
+    if self.document.info.has_pages then
+        menu_items.panel_zoom_options = {
+            text = _("Panel zoom (manga/comic)"),
+            sub_item_table = self:genPanelZoomMenu(),
+        }
+    end
     menu_items.translation_settings = Translator:genSettingsMenu()
 end
 
@@ -158,15 +160,21 @@ function ReaderHighlight:genPanelZoomMenu()
             checked_func = isPanelZoomEnabled,
             callback = function()
                 local toggled = not isPanelZoomEnabled()
-                G_reader_settings:saveSetting("panel_zoom_enabled", toggled)
+                DocSettings:saveSetting("panel_zoom_enabled", toggled)
+            end,
+            hold_callback = function()
+                local ext = util.getFileNameSuffix(self.ui.document.file)
+                local supported_ext = getPanelZoomSupportedExt()
+                supported_ext[ext] = not supported_ext[ext]
+                G_reader_settings:saveSetting("panel_zoom_ext", supported_ext)
             end,
             separator = true,
         },
-        genExtension("cbz"),
-        genExtension("cbt"),
-        genExtension("zip"),
-        genExtension("pdf"),
-        genExtension("djvu"),
+        -- genExtension("cbz"),
+        -- genExtension("cbt"),
+        -- genExtension("zip"),
+        -- genExtension("pdf"),
+        -- genExtension("djvu"),
     }
 end
 
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -121,7 +121,7 @@ local function getPanelZoomSupportedExt()
     return G_reader_settings:readSetting("panel_zoom_ext") or default_supported_ext
 end
 
-local function isDocumentComicOrManga(file)
+local function isPanelZoomSupported(file)
     local filetype = util.getFileNameSuffix(file)
     local supported_filetypes = getPanelZoomSupportedExt()
     return supported_filetypes[filetype]
@@ -1374,7 +1374,7 @@ function ReaderHighlight:onReadSettings(config)
     if self.document.info.has_pages then
         self.panel_zoom_enabled = config:readSetting("panel_zoom_enabled")
         if self.panel_zoom_enabled == nil then
-            self.panel_zoom_enabled = isDocumentComicOrManga(self.ui.document.file)
+            self.panel_zoom_enabled = isPanelZoomSupported(self.ui.document.file)
         end
     end
 end
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -606,7 +606,6 @@ function ReaderHighlight:onHold(arg, ges)
 
     -- disable hold gesture if highlighting is disabled
     if self.view.highlight.disabled then return false end
-
     self:clear() -- clear previous highlight (delayed clear may not have done it yet)
     self.hold_ges_pos = ges.pos -- remember hold original gesture position
     self.hold_pos = self.view:screenToPageTransform(ges.pos)
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -613,7 +613,7 @@ end
 
 local function isDocumentComicOrManga(file)
     local filetype = util.getFileNameSuffix(file)
-    return filetype == "cbz" or filetype == "djvu"
+    return filetype == "cbz" or filetype == "cbt" or filetype == "zip"
 end
 
 function ReaderHighlight:onHold(arg, ges)
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -107,6 +107,10 @@ local highlight_style = {
     invert = _("Invert"),
 }
 
+local function isPanelZoomAllowed()
+    return G_reader_settings:nilOrTrue("panel_zoom_allowed")
+end
+
 function ReaderHighlight:genHighlightDrawerMenu()
     local get_highlight_style = function(style)
         return {
@@ -133,6 +137,17 @@ function ReaderHighlight:genHighlightDrawerMenu()
             end,
             hold_callback = function(touchmenu_instance)
                 self:toggleDefault()
+            end,
+            separator = true,
+        },
+        {
+            text = _("Allow panel zoom in manga/comic"),
+            checked_func = function()
+                return isPanelZoomAllowed()
+            end,
+            callback = function()
+                local toggled = not isPanelZoomAllowed()
+                G_reader_settings:saveSetting("panel_zoom_allowed", toggled)
             end,
             separator = true,
         },
@@ -596,10 +611,13 @@ function ReaderHighlight:onPanelZoom(arg, ges)
     return true
 end
 
+local function isDocumentComicOrManga(file)
+    local filetype = util.getFileNameSuffix(file)
+    return filetype == "cbz" or filetype == "djvu"
+end
+
 function ReaderHighlight:onHold(arg, ges)
-    -- if it's manga/comic, panel zoom
-    local filetype = util.getFileNameSuffix(self.ui.document.file)
-    if filetype == "cbz" or filetype == "djvu" then
+    if isDocumentComicOrManga(self.ui.document.file) and isPanelZoomAllowed() then
         self:onPanelZoom(arg, ges)
         return false
     end
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -98,6 +98,10 @@ function ReaderHighlight:addToMainMenu(menu_items)
         text = _("Highlighting"),
         sub_item_table = self:genHighlightDrawerMenu(),
     }
+    menu_items.panel_zoom_options = {
+        text = _("Panel zoom (manga/comic)"),
+        sub_item_table = self:genPanelZoomMenu(),
+    }
     menu_items.translation_settings = Translator:genSettingsMenu()
 end
 
@@ -107,8 +111,58 @@ local highlight_style = {
     invert = _("Invert"),
 }
 
-local function isPanelZoomAllowed()
+local function getPanelZoomSupportedExt()
+    local default_supported_ext = {
+        cbz = true,
+        cbt = true,
+    }
+    return G_reader_settings:readSetting("panel_zoom_ext") or default_supported_ext
+end
+
+local function isDocumentComicOrManga(file)
+    local filetype = util.getFileNameSuffix(file)
+    local supported_filetypes = getPanelZoomSupportedExt()
+    return supported_filetypes[filetype]
+end
+
+local function isPanelZoomEnabled()
     return G_reader_settings:nilOrTrue("panel_zoom_allowed")
+end
+
+function ReaderHighlight:genPanelZoomMenu()
+    local function genExtension(ext)
+        return {
+            text = ext,
+            checked_func = function()
+                local supported_ext = getPanelZoomSupportedExt()
+                return supported_ext[ext]
+            end,
+            enabled_func = function()
+                return isPanelZoomEnabled()
+            end,
+            callback = function()
+                local supported_ext = getPanelZoomSupportedExt()
+                supported_ext[ext] = not supported_ext[ext]
+                G_reader_settings:saveSetting("panel_zoom_ext", supported_ext)
+            end
+        }
+    end
+    return {
+        {
+            text = _("Allow panel zoom"),
+            checked_func = isPanelZoomEnabled,
+            callback = function()
+                local toggled = not isPanelZoomEnabled()
+                G_reader_settings:saveSetting("panel_zoom_enabled", toggled)
+            end,
+            separator = true,
+        },
+        genExtension("cbz"),
+        genExtension("cbt"),
+        genExtension("zip"),
+        genExtension("pdf"),
+        genExtension("djvu"),
+    }
 end
 
 function ReaderHighlight:genHighlightDrawerMenu()
@@ -137,17 +191,6 @@ function ReaderHighlight:genHighlightDrawerMenu()
             end,
             hold_callback = function(touchmenu_instance)
                 self:toggleDefault()
-            end,
-            separator = true,
-        },
-        {
-            text = _("Allow panel zoom in manga/comic"),
-            checked_func = function()
-                return isPanelZoomAllowed()
-            end,
-            callback = function()
-                local toggled = not isPanelZoomAllowed()
-                G_reader_settings:saveSetting("panel_zoom_allowed", toggled)
             end,
             separator = true,
         },
@@ -611,13 +654,8 @@ function ReaderHighlight:onPanelZoom(arg, ges)
     return true
 end
 
-local function isDocumentComicOrManga(file)
-    local filetype = util.getFileNameSuffix(file)
-    return filetype == "cbz" or filetype == "cbt" or filetype == "zip"
-end
-
 function ReaderHighlight:onHold(arg, ges)
-    if isDocumentComicOrManga(self.ui.document.file) and isPanelZoomAllowed() then
+    if isDocumentComicOrManga(self.ui.document.file) and isPanelZoomEnabled() then
         self:onPanelZoom(arg, ges)
         return false
     end
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -603,7 +603,7 @@ end
 function ReaderHighlight:onHold(arg, ges)
     -- if it's manga/comic, panel zoom
     local filetype = util.getFileNameSuffix(self.ui.document.file)
-    if filetype == 'cbz' or filetype == 'djvu' then
+    if filetype == "cbz" or filetype == "djvu" then
         self:onPanelZoom(arg, ges)
         return false
     end
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -577,10 +577,12 @@ function ReaderHighlight:_resetHoldTimer(clear)
 end
 
 function ReaderHighlight:onPanelZoom(arg, ges)
-    -- @todo: add necessary checks
     self:clear()
+    -- if not pdf/cbz return
+    if not self.ui.document.info.has_pages then return false end
     local hold_ges_pos = ges.pos
     local hold_pos = self.view:screenToPageTransform(ges.pos)
+    if not hold_pos then return false end
     local rect = self.ui.document:getPanelFromPage(hold_pos.page, hold_pos)
     if not rect then return false end
     local image = self.ui.document:getPagePart(hold_pos.page, rect, 0)
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -1,7 +1,6 @@
 local BD = require("ui/bidi")
 local ButtonDialog = require("ui/widget/buttondialog")
 local Device = require("device")
-local DocSettings = require("docsettings")
 local Event = require("ui/event")
 local InfoMessage = require("ui/widget/infomessage")
 local InputContainer = require("ui/widget/container/inputcontainer")
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -580,7 +580,6 @@ function ReaderHighlight:onPanelZoom(arg, ges)
     self:clear()
     -- if not pdf/cbz return
     if not self.ui.document.info.has_pages then return false end
-    local hold_ges_pos = ges.pos
     local hold_pos = self.view:screenToPageTransform(ges.pos)
     if not hold_pos then return false end
     local rect = self.ui.document:getPanelFromPage(hold_pos.page, hold_pos)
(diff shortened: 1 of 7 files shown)


poire-z:
I won't be able to comment much on the kopt/leptonica stuff as I'm not familiar with it.
Just curious: for a page with 3 full-width panels, stacked vertically, what will this do ? Will there be a zoom ? Or just one of the 3 panels extracted and shown standalone not zoomed as it takes the screen width ?

Galunid:
Right now, it depends on the specific case. It will probably detect one page as one panel, and "panel of that page" as another, so it depends which one is "earlier" in the array. Additionally there is a check whether the panel is bigger than 1/10 of width/height, so it's possible the smaller one will get rejected. So far, this only triggers image detection in `KOPTContext` and it writes to files with index numbers. I'm figuring out how to get image/boundary from `KOPTContext` back to ReaderHighlight, so it can be displayed. It shouldn't be difficult, but I'm taking a break from programming/studying for today ;-)

poire-z:
(While browsing old issues, I just met #2478 - so just mentionning it here as it's also about detecting panels.)

Galunid:
Assuming they are within "borders" (black rectangles enclosing the text), it will work on that too! ;-)

Frenzie:
I think those borders are merely illustrative. :-)

Galunid:
Welp, there are some examples of page segmentation in leptonica, so I may take a look after I'm done with this

Galunid:
Is there a way to set default value format-wise? What I mean is for `.cbz` to have different value than `.epub/.pdf`.
If anyone is interesting in testing it needs https://github.com/koreader/koreader-base/pull/1159 and disabling highlighting. Any ideas on how to add this to menu greatly appreciated.

Frenzie:
> Is there a way to set default value format-wise? What I mean is for .cbz to have different value than .epub/.pdf.

You mean currently? I don't think so; we pretty much do things engine-wise, and CBZ is handled by MuPDF. I've thought about it a little bit (in the context of MuPDF's EPUB support) but never implemented anything.

ezdiy:
Tangentially related: Reading comic books on 7" is not always all that pleasant not necessarily because I'd care enough to zoom into panels, but because the bubble font is too small. So I have this idea floating around to:

1) Identify text bubbles
2) Auto-detect if font is too small (and hallucinate zoom factor from it) in there best scenario, but good enough would be even just user setting per book
2) Zoom text in there just by stretching it (ocr too slow/inaccurate). This should look ok because the pages are usually far higher resolution than native device one anyway. Still, needs to "organically" enlarge the bubble area if need be to fit the zoomed text.

Would leptonica be up to the task, or does this mean bringing more image processing in?

Galunid:
Leptonica should be able to do that. The algorithm I have in mind would be something like detect text regions, remove them, use edge detection algorithm -> crop and zoom. We could do it even faster, assuming user "holds the speech bubble". In that case we only need to find text only in that region. I'd like to finish implementing EDP controller for Onyx devices, since whether I keep it or not depends on that and I have 2 days left to decide. After that I'll finish this PR and try to add some documentation to stuff related to Leptonica. I can try to implement your suggestion when I'm done with that. It would be cool to have double tap to zoom to panel, hold to zoom to speech bubble.

Frenzie:
> We could do it even faster, assuming user "holds the speech bubble".

Potentially sounds like overcomplicating things compared to just zoom to region. ^_^

ezdiy:
Yep, the problem is that whenever you'd want to do this, you'd end up tapping all bubbles anyway. I'd prefer to limit interaction with the page to no interaction (as my device input speed/accuracy is ... not great). Hence why the wholesale "for each bubble, detect bubble clipping mask, zoom that, plaster it over center of the original bubble). The point is to deal scanlations where typesetting isn't particularly great.

Galunid:
Makes sense, I'll see what I can do.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -595,13 +595,21 @@ function ReaderHighlight:onPanelZoom(arg, ges)
...
     end
+    logger.dbg("File:", self.ui.document.file)
+    logger.dbg("Extension:", util.getFileNameSuffix(self.ui.document.file))
     return true
 end
 
 function ReaderHighlight:onHold(arg, ges)
+    -- if it's manga/comic, panel zoom
+    local filetype = util.getFileNameSuffix(self.ui.document.file)
+    if filetype == 'cbz' or filetype == 'djvu' then
```
Frenzie, line 606: Nit:
```suggestion
    if filetype == "cbz" or filetype == "djvu" then
```

This is just a proof of concept atm I assume? Otherwise I might have bigger comments. :-P

Galunid: It's PoC for now, I'll try to get it ready for review tomorrow, when I add menu entry to enable/disable this feature (and figure out where to put it ;-))

Galunid:
I put the "toggle" option in `Highlighting` menu, I couldn't really think of a better place, but suggestions are really welcome here, should be ready for review.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -596,10 +611,13 @@ function ReaderHighlight:onPanelZoom(arg, ges)
     return true
 end
 
+local function isDocumentComicOrManga(file)
+    local filetype = util.getFileNameSuffix(file)
+    return filetype == "cbz" or filetype == "djvu"
```
Frenzie, line 616: I can see treating CBZ as a comic by default, but DjVu is basically the same as PDF here? I read most of my manga/comics in PDF. On a related note, we also support CBT.

ezdiy: `filetype:match("^cb[zrt7]") or filetype=="zip"`

NiLuJe: We don't currently support CBR and CB7(or do we? :D), and zip is potentially a bit of an issue for zipped FB2 or something?

But, yeah, +1 on the cb pattern regardless ;).

ezdiy: @NiLuJe mupdf can do it, provided it's built that way. As for zip, it's extremely common format for ... "certain" kinds of manga :>

That said whether it's zip or fb2 inside is discerned by mupdf again, so perhaps more reliable method would be to inquire mupdf about what's it doing.

Galunid: I think for now, it's better not to support `djvu`, since they sometimes fail, when `cbz` works (due to lower resolution believe). I just tested `cbt` and it works without problems. I'm not sure about `cb7`, but `cbr` isn't supported, right?

Galunid: @NiLuJe zipped fb2 don't have extension of `.fb2.zip`?

Galunid: Perhaps a better solution would be to add a separate menu entry, where we let user decide which formats they want supported?

NiLuJe:
I like that idea, yeah ;).

Galunid:
Red rectangle is what's enabled by default, is that ok?
image

frontend/apps/reader/modules/readerhighlight.lua
Frenzie: I think this whole notion of support is weird when it seems like a feature similar to reflow. We talked about defaults earlier.

Frenzie: > Perhaps a better solution would be to add a separate menu entry, where we let user decide which formats they want supported?

Strange, GH didn't show this to me. Anyway, I'd say it should be a per document thing (cf. reflow), possibly with a per format default.

Frenzie:
Turn on panel zoom

poire-z:
> I think this whole notion of support is weird when it seems like a feature similar to reflow. We talked about defaults earlier.

I kind of agree, it doesn't feel right to tie features to extensions, even if we let the user choose the extensions.
I don't read comics, so I haven't followed much this PR. But tell me if I get it right:
when long press on a panel (the 2-6 small images in a comic page), it should zoom that vignette?
Nothing to do with the text bubbles, right?
So, can't we have both working at the same time?:
- if long press at some position, and text is found at that position: do normal text selection action (dict lookup if single word, otherwise show dialog)
- if no text dedected at position, and if panel zoom is enabled, do panel detection and zoom

(Dunno if we can detect there's text at position, or if the stuff will look for nearest text...)

Frenzie:
@poire-z My PDF Humble Bundle comics often have selectable text. (But I've only read a manga on my device once, Cells at Work. Most stuff isn't b&w so unsuitable.) I prefer the PDFs because they render sharper, with vector text among other things.

But also atm I think it'll just select the nearest text, even if it's pretty far away?

poire-z:
> But also atm I think it'll just select the nearest text, even if it's pretty far away?

So, may be when "panel detection" is enabled, it should additionally to not go too far away :) 
I don't know if that's possible, just throwing some ideas, @Galunid might now more what's possible :)

Btw, be sure that new menu item "Panel zoom" only shows with `document.info.has_pages`, or does it work on EPUBs?

ezdiy:
> (Dunno if we can detect there's text at position, or if the stuff will look for nearest text...)

koreader doesn't have a robust scene text detection. That's a thing you throw whole page at and it tells you *all* the rect that most likely are text. Algorithms for this exist and are resonably simlpe (no need for fancy NNs), but not sure if tesseract alone expose any.

Galunid:
@poire-z I had little luck selecting text in comics, more often than not, it selects part of the image


> when long press on a panel (the 2-6 small images in a comic page), it should zoom that vignette?
> Nothing to do with the text bubbles, right?

Yes, it has nothing to do with text bubbles


> So, may be when "panel detection" is enabled, it should additionally not go too far away :)
> I don't know if that's possible, just throwing some ideas, @Galunid might know more what's possible :)

That's how it's supposed to work. The only issue is when 2 panels are "connected" (by text bubble for example), then it'll select 2 panels. Like this
image
For now I've read some papers on how to deal with that. [This one](http://visal.cs.cityu.edu.hk/static/pubs/conf/mm14-panels.pdf) seems promising, but I haven't gotten around implementing anything. 

@Frenzie 
>  Anyway, I'd say it should be a per document thing (cf. reflow), possibly with a per format default.

I kind of disagree, usually when someone reads manga series they binge it I think (at least I do), so it'd be pretty comfortable to say enable this for all pdf's and disable it once I'm done. I assume `per document` would mean I have to enable it for every single file. Since series are often 100+ files, I don't find the idea appealing.

Galunid:
> I kind of agree, it doesn't feel right to tie features to extensions, even if we let the user choose the extensions.

Would you mind elaborating a bit? I don't really see the downside.



> Btw, be sure that new menu item "Panel zoom" only shows with `document.info.has_pages`, or does it work on EPUBs?

Doesn't work in epub, I forgot to add the condition for displaying it. I will do that soon.

poire-z:
> Would you mind elaborating a bit? I don't really see the downside.

Just that we didn't do anywhere that kind of features/extension tie before this PR (at most, we do it for picdocument/is_pic, that we assume are non-text images).
I see it's practical, so one can enable it for cbz while still reading PDF.
But in the absolute, a cbz may contain text images, and a pdf may be a comic, and the user would have to tweak these settings when meeting these non-classic cbz and pdf.
Ideally, you would detect this per page, or when needed by finger position.

Frenzie:
> I kind of disagree, usually when someone reads manga series they binge it I think (at least I do), so it'd be pretty comfortable to say enable this for all pdf's and disable it once I'm done. I assume per document would mean I have to enable it for every single file. Since series are often 100+ files, I don't find the idea appealing.

No, per document means the ability to override whatever you set as the default. ;-) Any per-document setting should have a long press set default setting associated with it.

Galunid:
@Frenzie So you're not against associating those settings with extensions, but against a way in which they are presented? Do you just want me to move it from top menu to the bottom menu near reflow with on/off switch?

Frenzie:
No, there have to be per-document settings. That's unrelated to what you're talking about.

* Normal means per-document and global default.
* Custom would mean per-document and extension-specific default.

ezdiy:
+1 per document, just like stroke boldness font size etc. All of these are tracked by default per document, but if you long-press you configure the default value for documents that don't set their own. This is far more ergonomic than any settings menu.

Galunid:
Do I get it right? There is global setting with a list of extensions where `panel zoom` is enabled by default. Then there is a document wide setting that allows for toggling `panel zoom` on/off for that specific document. When I hold on that document-wide setting button, it should change global setting?

poire-z:
I think you should get rid of the extension selection, and just have an independant menu item with toggle checkbox (i.e. like Floating punctuation):
- tap on that menu item : save state in docsetting (per-document)
- hold on that menu item : save state in settings.reader.lue (global)
- open a new document: use the default value from settings.reader.lua, or your default value (and may be, if there's no default value, start with true for CB* and false for anything else) (and the thing to discuss: whether to save it in docsettings or not)

So, when you're in a Comics reading campaign, you just hold on that menu to set it globally to true. If in this campaign, you open a text PDF, you sut tap on that menu to change it for this book. Open a new CBZ and it opens with the global value.
When you're done with comics, and plan reading more text, you just hold to set it globally to false.

It would feel more accessible in the bottom menu - but the bottom menu is essentially about rendering options... So, having it here fell strange, and I can't find any bottom panel where it would logically fit.

Frenzie:
It's in roughly the same domain as something like crop or reflow, although it's not exactly about rendering. But also I think it would just feel nicely accessible there. :-)

tl;dr What @poire-z said.

poire-z:
And don't forget we have gestures, so if you make it a dispatcher action, toggling between text selection and panel zoom is just a gesture away.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -125,8 +126,12 @@ local function isDocumentComicOrManga(file)
     return supported_filetypes[filetype]
 end
 
-local function isPanelZoomEnabled()
-    return G_reader_settings:nilOrTrue("panel_zoom_allowed")
+local function isPanelZoomEnabled(file)
+    local doc_setting = DocSettings:readSetting("panel_zoom_enabled")
```
poire-z, line 130: You can access directly the setting for the currently opened book with:
`self.ui.doc_settings:readSetting("panel_zoom_enabled")`
without re-opening it (and it it has been updated but not yet flushed, you get the updated value).
But as you probably will set the menu to toggle it in this file, you'll probably keep a copy as self.panel_zoom_enabled that you can access directly - and load/save it in the existing:
```lua
function ReaderHighlight:onReadSettings(config)
function ReaderHighlight:onSaveSettings()
```

Galunid: Thanks, saved me quite a bit of time!

Galunid:
Anyone knows how to add stuff to the bottom menu? I was checking and it seems `frontend/ui/data/koptoptions.lua` is responsible for that, but it works differently(?) from the top menu and I don't really see a simple way to include things from `ReaderHighlight` there.

poire-z:
> I don't really see a simple way to include things from ReaderHighlight there.

Yes, we don't populate the bottom menu, like we do for the top menu.
The bottom panels are standalone and defined in koptoptions.lua / creoptions.lua. In there, you tell everything about what you want in the bottom panels, and you tell which events are fired. Then, in ReaderHighlight, you just add the `onEventName()` method that will trigger when that event is fired.
(There's a bit of confusion about which from the bottom pannel and the target module is responsible for saving/loading the state of the toggle - the bottom panel is - and there's also an intermediate readerkoptlistener - so you'll see sometimes that a setting is saved twice with a different name :) it's a bit of a mess.)

For the PDF bottom panel, I don't know where you plan to add your toggle if you really want it there (again, mostly rendering stuff). Try to keep the height of each panel small: the max is 4 widgets par panel currently, you could go to 5 if needed, but if you can fill the 2 left panels that have 1 or 2 items, that would be nicer - althought panel zooming is a bit unrelated, except for the bottom panels icons :) the 2 left ones could visually evoke a comic panel extraction/cropping:
image :)

NiLuJe:
I don't recall what's currently in it, but, yeah, that'd certainly make sense in the crop tab ;).

frontend/dispatcher.lua
```diff
@@ -103,7 +103,8 @@ local settingsList = {
     cycle_highlight_style = { category="none", event="CycleHighlightStyle", title=_("Cycle highlight style"), rolling=true, paging=true, separator=true,},
     kosync_push_progress = { category="none", event="KOSyncPushProgress", title=_("Push progress from this device"), rolling=true, paging=true,},
     kosync_pull_progress = { category="none", event="KOSyncPullProgress", title=_("Pull progress from other devices"), rolling=true, paging=true, separator=true,},
-    page_jmp = { category="absolutenumber", event="GotoViewRel", min=-100, max=100, title=_("Go %1 pages"), rolling=true, paging=true,},
+    page_jmp = { category="absolutenumber", event="GotoViewRel", min=-100, max=100, title=_("Go %1 pages"), rolling=true, paging=true, separator=true,},
+    panel_zoom_toggle = { category="none", event="TogglePanelZoomSetting", title=_("Toggle panel zoom"), rolling=false, paging=true,},
```
yparitcher, line 107: the separator will be shown after the item in the order of`dispatcher_menu_order` so there is no need to add it here, also there is no need to set `rolling=false`

```suggestion
    page_jmp = { category="absolutenumber", event="GotoViewRel", min=-100, max=100, title=_("Go %1 pages"), rolling=true, paging=true,},
    panel_zoom_toggle = { category="none", event="TogglePanelZoomSetting", title=_("Toggle panel zoom"), paging=true,},
```

frontend/dispatcher.lua
```diff
@@ -206,6 +207,7 @@ local dispatcher_menu_order = {
     "back",
     "previous_location",
     "latest_bookmark",
+    "panel_zoom_toggle",
```
yparitcher, line 210: this should probably go before or after
```lua
    "cycle_highlight_action",
    "cycle_highlight_style",
```

if you put it after then please remove the separator from `cycle_highlight_style` and add it to `panel_zoom_toggle`

yparitcher (changes requested):

Galunid:
To be fair, I don't really mind it in top menu, but I believe @Frenzie preferred it at the bottom.
I think I'll give up on that, since setting/getting default values is very problematic (no `hold_callback` and default value must be value (can't get it from function since `koptoptions` doesn't have `file` variable)). On first run we depend on the file extension. We also depend on extension when setting defaults.

Frenzie:
I think it'd be easier in the bottom from a UX perspective but programmatically speaking it's most the same thing needing to support the same actions no matter where you put it.

frontend/dispatcher.lua
```diff
@@ -100,10 +100,11 @@ local settingsList = {
...
     toggle_inverse_reading_order = { category="none", event="ToggleReadingOrder", title=_("Toggle page turn direction"), rolling=true, paging=true,},
     cycle_highlight_action = { category="none", event="CycleHighlightAction", title=_("Cycle highlight action"), rolling=true, paging=true,},
-    cycle_highlight_style = { category="none", event="CycleHighlightStyle", title=_("Cycle highlight style"), rolling=true, paging=true, separator=true,},
+    cycle_highlight_style = { category="none", event="CycleHighlightStyle", title=_("Cycle highlight style"), rolling=true, paging=true,},
     kosync_push_progress = { category="none", event="KOSyncPushProgress", title=_("Push progress from this device"), rolling=true, paging=true,},
     kosync_pull_progress = { category="none", event="KOSyncPullProgress", title=_("Pull progress from other devices"), rolling=true, paging=true, separator=true,},
     page_jmp = { category="absolutenumber", event="GotoViewRel", min=-100, max=100, title=_("Go %1 pages"), rolling=true, paging=true,},
+    panel_zoom_toggle = { category="none", event="TogglePanelZoomSetting", title=_("Toggle panel zoom"), rolling=false, paging=true, separator=true,},
```
yparitcher, line 107:
```suggestion
    panel_zoom_toggle = { category="none", event="TogglePanelZoomSetting", title=_("Toggle panel zoom"), paging=true, separator=true,},
```

yparitcher (approved):

Galunid:
>I think it'd be easier in the bottom from a UX perspective

@Frenzie 
I agree, but I don't think it's actually doable without a lot of hacky code, I think it's better left in the top menu, is that acceptable?

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -107,6 +113,41 @@ local highlight_style = {
...
 
+local function getPanelZoomSupportedExt()
+    local default_supported_ext = {
+        cbz = true,
+        cbt = true,
+    }
+    return G_reader_settings:readSetting("panel_zoom_ext") or default_supported_ext
+end
+
+local function isDocumentComicOrManga(file)
```
poire-z, line 124: Unless you plan later to detect if document is really a comic or manga, I would rename this to what it actually does :) 
`isPanelZoomSupported(file)`

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -576,7 +617,37 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+        }
+        UIManager:show(imgviewer)
+    end
+    return true
+end
+
 function ReaderHighlight:onHold(arg, ges)
+    if self.panel_zoom_enabled then
+        self:onPanelZoom(arg, ges)
+        return false
```
poire-z, line 648: You're voluntarily letting the hold propagate?

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -576,7 +617,37 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+            with_title_bar = false,
+            fullscreen = true,
+        }
+        UIManager:show(imgviewer)
+    end
+    return true
+end
+
 function ReaderHighlight:onHold(arg, ges)
+    if self.panel_zoom_enabled then
```
poire-z, line 646: I would still wrap that into a `if self.document.info.has_pages then` to make it clear it works only on PDF and not on CRE.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -1298,6 +1369,12 @@ function ReaderHighlight:onReadSettings(config)
         disable_highlight = G_reader_settings:readSetting("highlight_disabled") or false
     end
     self.view.highlight.disabled = disable_highlight
+
+    -- panel zoom settings
+    self.panel_zoom_enabled = config:readSetting("panel_zoom_enabled")
```
poire-z, line 1374: I would wrap that too (loading & saving) into a `if self.document.info.has_pages then`.
(I can't update this setting as the menu is not displayed on CRE document, but if I were to toggle manually the saved setting in a metadata.epub.lua, things could break - so, I feel it's better to have explicite checks.)

poire-z (commented):
I'm fine with letter that in the top menu.
If it happens it works nice and becore a top feature, we can then decide to move it into the bottom menu.

frontend/apps/reader/modules/readerhighlight.lua
Galunid: Good catch, since it's toggable via gesture

Frenzie:
> I think it's better left in the top menu, is that acceptable?

Yes, it can always be changed later.

Galunid:
Should be ready for the final review

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -576,7 +617,37 @@ function ReaderHighlight:_resetHoldTimer(clear)
     end
 end
 
+function ReaderHighlight:onTogglePanelZoomSetting(arg, ges)
+    if not self.document.info.has_pages then return end
+    self.panel_zoom_enabled = not self.panel_zoom_enabled
+    self:onSaveSettings()
```
poire-z, line 623: I believe this onSaveSettings is unnecessary (it will be called when reader is closed and will save the updated self.panel_zoom_enabled ).

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -576,7 +617,37 @@ function ReaderHighlight:_resetHoldTimer(clear)
...
+            with_title_bar = false,
+            fullscreen = true,
+        }
+        UIManager:show(imgviewer)
+    end
+    return true
+end
+
 function ReaderHighlight:onHold(arg, ges)
+    if self.panel_zoom_enabled and self.document.info.has_pages then
```
poire-z, line 647: (nitpick: logical order, just for readability: first self.document.info.has_pages , and then self.panel_zoom_enabled  that makes sense only then.)

poire-z (approved):
lgtm (with these 2 nitpicks)

poire-z (approved):

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -1298,6 +1368,14 @@ function ReaderHighlight:onReadSettings(config)
         disable_highlight = G_reader_settings:readSetting("highlight_disabled") or false
     end
     self.view.highlight.disabled = disable_highlight
+
+    -- panel zoom settings shouldn't work in EPUB
```
Frenzie, line 1372: I'd phrase it slightly differently because if I look at my comics EPUB is a distinct download option! ;-D
```suggestion
    -- panel zoom settings isn't supported in EPUB
```

Screenshot_2020-09-24_15-15-05

Frenzie (approved):

poire-z:
Just tested on the emulator with the only CBZ I have, it extracts things very fast !
But it seems there's some issue with the quality/scaling:
Original page in KOReader:
<kbd>image</kbd>
Zoomed panel in ImageViewer:
<kbd>image</kbd>
And in ImageViewer when I tap "Original size":
<kbd>image</kbd>

Dunno what's the reason (bad scaling by kopt/leptonica? bad scaling/dithering by our ImageWidget? better up-scaling by MuPDF? or my low DPI emulator?)

Frenzie:
I thought the idea was to grab that part of the image at a higher resolution for optimal display, but that looks like it went wrong and grabbed lower res? I see the same effect actually.

Screenshot_2020-09-24_15-38-51
Screenshot_2020-09-24_15-39-14

Frenzie:
But anyway, the panel detection works beautifully! ^_^

Galunid:
@poire-z @Frenzie 
https://github.com/koreader/koreader/blob/15455b594d89810842aa0c1f6aa5d5431dfbcdb1/frontend/document/document.lua#L416-L421
Is responsible for cutting out the panel. I'm not really familiar with the whole blitbuffer/rendering stuff, so I don't know if it's possible to improve it. I guess there should be better way getting image, instad of using `Document:renderPage`, since that's most likely what's causing low resolution. Any suggestions are welcome.

Frenzie:
It's not so much the renderPage as that you're passing a zoom of 1. What we want instead is something like this. :-)

https://github.com/koreader/koreader/blob/15455b594d89810842aa0c1f6aa5d5431dfbcdb1/frontend/document/koptinterface.lua#L285-L286

Frenzie:
To make sure I get my implicit message across clearly, I think a `local zoom = math.min(canvas_size.w / rect.w, canvas_size.h / rect.h)` should do the trick but I'm on my laptop now so annoyingly I don't have an appropriate CBZ at hand to quickly test.

Frenzie:
@Galunid Gave it a quick try, it works but you also need to adjust rect with the zoom factor after calculating it.

Galunid:
```lua
function Document:getPagePart(pageno, rect, zoom, rotation)
    rect.x = rect.x * zoom
    rect.y = rect.y * zoom
    rect.w = rect.w * zoom
    rect.h = rect.h * zoom
    local tile = self:renderPage(pageno, rect, zoom, rotation, 1, 0)
    local target = Blitbuffer.new(rect.w, rect.h, self.render_color and self.color_bb_type or nil)
    target:blitFrom(tile.bb, 0, 0, rect.x, rect.y, rect.w, rect.h)
    return target
end
```

should work, but I'm not sure how to get canvas size from readerhighlight

poire-z:
(I thought I could have mentionned the following - but only thing vaguely related is https://github.com/koreader/koreader/pull/6511#issuecomment-695798892.)

In there:
https://github.com/koreader/koreader/blob/c20ad8f5e9385dd18138a306b874c5b20302f299/frontend/apps/reader/modules/readerhighlight.lua#L685-L708
can't we tweak things a bit so that, when panel zoom is enabled, but no image is detected, just continue handling Hold the normal way and do dict lookup or text selection ?

Use case: I'm reading a PDF with a few illustations: with PanelZoom, long-press on an image shows it just like in EPUB, which is nice - but then dict doesn't work. I have to toggle panel zoom to do dict - and then toggle it again for the next image.
Would this be a problem with other contexts/use-cases ?

Galunid:
@poire-z I was considering this when I first wrote this, the reason I opted not to do this is because in comics/manga (main application) it'd perform OCR, which is usually unwanted, and is pretty broken there. So yes, it'd be problem with comics/manga.

I don't know if we can detect text at given locations, since it's more of a cre/kopt stuff. I'm somewhat interested in being able to zoom PDF images, so it's possible I'll look a bit more into this in the near future.

poire-z:
> I don't know if we can detect text at given locations

Before invoking OCR ? Dunno. It feels there should be a first step of detecting text boxes before invoking OCR, but all that might be handled as one step way far away :) 
May be the easier would be just another option in the `Panel zoom >` (where `[ ] Allow panel zoom` must feel a bit lonely :) : `[ ] Fallback to text selection when no image detected` - that I would enable, but comics/manga readers would not.

Galunid:
I was thinking of something along those lines

PR #6513, opened 2020-08-16
refactor thirdparty app integration

pazos:
Removes some duplicities between desktop and android.
Splits platform specific code from common code used in both platforms.
Makes possible to assign roles to 3rd party apps (only dict is used, but translator is left as an example)

App discovery will be done once at launch and will be logged.

example with no overrides
```
08/17/20-00:55:53 INFO  platform thirdparty apps
-> Goldendict (goldendict), role: dict, available: true
```

example with random overrides

```
08/17/20-00:58:29 INFO  user thirdparty apps
-> SomeThing (asdf), role: dict, available: false
-> Qdict (qdict), role: dict, available: false
-> Webster (https://webster.com/?=), role: dict, available: false
```

Tested on the emulator and android. Works fine ~~Need to try in android~~

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6513)


poire-z (approved):
Looks allright (I'm not super familiar with :new and metatable stuff, but if it works, it must be fine :)
(The .."s" to have the list name plural but the thing name singular is a bit confusing at first, but I guess it makes it more natural language in the device code.)

Frenzie (approved):

Issue #6278, opened 2020-06-17
Night mode in kobo 


amir2mi:
* KOReader version: last version
* Device: Kobo Aura 2

##### `crash.log` (if applicable)
`crash.log` is a file that is automatically created when KOReader crashes. It can
normally be found in the KOReader directory:

* `/mnt/private/koreader` for Cervantes
* `koreader/` directory for Kindle
* `.adds/koreader/` directory for Kobo
* `applications/koreader/` directory for Pocketbook

Android won't have a crash.log file because Google restricts what apps can log, so you'll need to obtain logs using `adb logcat KOReader:I ActivityManager:* AndroidRuntime:* *:F`.

Please try to include the relevant sections in your issue description.
You can upload the whole `crash.log` file on GitHub by dragging and
dropping it onto this textbox, or you can use one of these paste services.
Please don't forget to set a high expiration date if you do.
https://pastebin.com https://slexy.org https://paste2.org https://fpaste.org
http://ix.io https://paste.kde.org https://paste.debian.net.

NiLuJe:
Try with an up to date firmware.

amirmp3:
> Try with an up to date firmware.

I updated kobo today and installed Koreader with kobo start menu 9.

NiLuJe:
Then we're going to need more details, an actual FW version, and an actually accurate device name, because there's no such thing as an "Aura 2" (there's an Aura SE, and Aura H2O, and two different revisions of an Aura H2O²).

Also, KSM is unsupported, not that it should matter much now, but, still.

NiLuJe:
And also an accurate and detailed description of what's happening, because right now it milldy looks like you're confusing actual nightmode (i.e., persistent white on black display) with the black flash that intermittently happens on screen refreshes.

Frenzie:
> Also, KSM is unsupported, not that it should matter much now, but, still.

I don't intend to stop using it quite yet. ;-)

NiLuJe:
> 
> 
> > Also, KSM is unsupported, not that it should matter much now, but, still.
> 
> I don't intend to stop using it quite yet. ;-)

It might be mildly problematic if it turns out this is actually one of the Mk. 7 Aura ;).

But not on that front (usually).

Frenzie:
> Aura SE

PS Sounds like a funny way of saying "2". xD

amirmp3:
[...]

PR #6514, opened 2020-08-17
fix nightmode after standby on Kobo Aura HD

hurzl:
https://github.com/koreader/koreader/issues/6278#issuecomment-669167527

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6514)

### DIFF
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -106,6 +106,7 @@ local KoboDragon = Kobo:new{
     hasFrontlight = yes,
     hasMultitouch = no,
     display_dpi = 265,
+    canHWInvert = no,
 }
 
 -- Kobo Glo:


frontend/device/kobo/device.lua
```diff
@@ -106,6 +106,7 @@ local KoboDragon = Kobo:new{
     hasFrontlight = yes,
     hasMultitouch = no,
     display_dpi = 265,
+    canHWInvert = no,
```
Frenzie, line 109: I'll wait for @NiLuJe (who may not reply for another week or so) but I'd add a comment referring to this PR or the issue saying that it's just for after standby, otherwise that'd be confusing. (In principle you can read these as a feature list.)

NiLuJe:
I kinda feel like NACK'ing that. It's a kernel issue that can be worked around pretty easily, and the downsides of this don't warrant it, IMHO.

Hardware inversion is broken in some corner cases on a *lot* of devices, and the only hard disable we enforce are on one's where it actually hangs the device.

NiLuJe:
I'd be mildly happier with a soft, user-togglable option, like we do for dithering. But not by much.

hurzl:
Then please work around it, if it is so easy.
Togglable option for HWInvert?

Frenzie:
> Then please work around it, if it is so easy.

I suspect @NiLuJe means something like disabling and reenabling it rather than assuming state is restored in a sane manner. (And possibly on all platforms because why not.)

hurzl:
I wouldn't call that a workaround. 
I think disabling HWinvert is a workaround.

Frenzie:
Doing it in software will definitely be slower, could potentially be uglier too if they do something clever with the screen updates when inverted but I don't know if that applies in practice.

hurzl:
You really expect users to
1. open the menu bar
2. select the settings menu
3. uncheck invert
4. check invert
5. close the menu

each time when switching on after standby? I usually just want to start reading.

Galunid:
You can set gesture that does that for you though?
`Gesture manager -> Something you want -> Select gesture you want -> Toggle Nightmode & Nightmode -> ON`
Should do all of that in 1 swipe

Frenzie:
@hurzl …no? You misinterpreted what I wrote. :-) I meant "disabling and reenabling it" as part of the resume procedure.

Like this:
https://github.com/koreader/koreader/blob/d90e221a1797ac79c21bad58869013b75480045a/frontend/device/kobo/powerd.lua#L400-L415

Or that:
https://github.com/koreader/koreader/blob/9037a1f4f28f5bc77d6ac74249b57538f643c9bb/frontend/device/generic/device.lua#L206-L296

hurzl:
@Frenzie ok why not if that re-enables night mode automatically? But it's not there (yet)

Frenzie:
I just mean something like this:
```diff
diff --git a/frontend/device/generic/device.lua b/frontend/device/generic/device.lua
index e2e56e23..09c8b55c 100644
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -228,6 +228,11 @@ function Device:onPowerEvent(ev)
                 if self:needsScreenRefreshAfterResume() then
                     UIManager:scheduleIn(1, function() self.screen:refreshFull() end)
                 end
+                -- Workaround for HW invert being disabled after resume on some devices
+                if G_reader_settings:isTrue("night_mode") then
+                    UIManager:broadcastEvent(Event:new("SetNightMode", false))
+                    UIManager:broadcastEvent(Event:new("SetNightMode", true))
+                end
                 self.screen_saver_mode = false
                 self.powerd:afterResume()
             end
```

hurzl:
thats a crash. 
On this buggy device its a real pain to apply changes so I will keep my HWInvert workaround.

pazos:
@hurzl: the workaround is fine for you but we cannot accept it. It would mean to downgrade the experience of all Aura HD users even if most of them don't use night mode at all!

hurzl:
you prefer to keep your bugs, fine.

pazos:
A proper implementation, like it was suggested to you, is more than welcome.

Of course we don't expect you to do the work if you don't want.

> neither is there a downgrade or do you know whether "most of them don't use night mode at all!"

Please do take a bit of time to understand conditionals. Thank you

hurzl:
Even if you knew it wouldn't even matter.

Frenzie:
@hurzl 
> thats a crash.

What did the crash say?

hurzl:
I couldn't read it because it restarted after a few seconds.

Frenzie:
This will work (or at least not crash).

```diff
diff --git a/frontend/device/generic/device.lua b/frontend/device/generic/device.lua
index e2e56e23..00c8e864 100644
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -224,6 +224,12 @@ function Device:onPowerEvent(ev)
                 if self.orig_rotation_mode then
                     self.screen:setRotationMode(self.orig_rotation_mode)
                 end
+                -- Workaround for HW invert being disabled after resume on some devices
+                if G_reader_settings:isTrue("night_mode") then
+                    local Event = require("ui/event")
+                    UIManager:broadcastEvent(Event:new("SetNightMode", false))
+                    UIManager:broadcastEvent(Event:new("SetNightMode", true))
+                end
                 require("ui/screensaver"):close()
                 if self:needsScreenRefreshAfterResume() then
                     UIManager:scheduleIn(1, function() self.screen:refreshFull() end)
```

NiLuJe:
Except probably not, unless you maybe sneak a dummy refresh between the
two?

Frenzie:
Note there's a user patch now: https://gist.github.com/NiLuJe/4be737eb123cf1199590b626be131345

More information: https://github.com/koreader/koreader/wiki/User-patches

Issue #6516, opened 2020-08-17
Favorites are not displayed when accessed via multiswipe


elvvis:
* KOReader version: 2020.08
* Device: Tolino Epos 2

#### Issue

When calling up the favorites via a multiswipe, an empty page is displayed. When calling up the favorites via the menu, they are displayed correctly.

yparitcher:
caused by my gesture -> dispatcher refactor, will fix later today

PR #6519, opened 2020-08-17
Dispatcher: fix events with static argument

yparitcher:
Fixes: #6516

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6519)


frontend/dispatcher.lua
```diff
@@ -547,7 +547,11 @@ function Dispatcher:execute(ui, settings, gesture)
     for k, v in pairs(settings) do
         if settingsList[k].conditions == nil or settingsList[k].conditions == true then
             if settingsList[k].category == "none" then
-                ui:handleEvent(Event:new(settingsList[k].event))
+                if settingsList[k].arg ~= nil then
+                    ui:handleEvent(Event:new(settingsList[k].event, settingsList[k].arg))
```
Frenzie, line 551: It's probably not relevant here or we would've noticed already, but of course there are also events with multiple arguments. For that you could use probably use `table.unpack()` when necessary.

yparitcher: not relavant here, and perfer to keep the code simpler for now.

PR #6525, opened 2020-08-19
Update PocketBook device definitions

### LINKED ISSUE #4141: PocketBook 615 Plus doesn't start
* KOReader version: any
* Device: PocketBook 615 Plus

#### Issue

Copied and merged `applications` folder. Tried to start it from "Applications" but it's just refreshing System UI without starting.
Tried to find `crash.log` file, but it doesn't created at all.

In "Applications" menu I see icon "@Koreader".

I also tried to add this model to `device.lua`

```lua
-- PocketBook 615 Plus
local PocketBook615 = PocketBook:new{
    isTouchDevice = no,
    hasKeys = yes,
    hasFrontlight = yes,
    display_dpi = 212,
    emu_events_dev = "/var/dev/shm/emu_events",
}
```

And on end tried all possible variants of model definition:
```lua
if codename == "PB615" then
    return PocketBook615
elseif codename == "PocketBook 615" then
    return PocketBook615
elseif codename == "PocketBook 615 Plus" then
    return PocketBook615
-- ...
```

But it doesn't work also

roshavagarga:
**Changes:**
- [x] Added PocketBook 615 Plus definition. (Fixes #4141?)
- [x] Added PocketBook 616 definition.

**Reasons:**
- It seems like the 615 Plus is a CIS-only device, there's a [manual](http://support.pocketbook-int.com/fw/615-2/u/5.17.999/manual/cis/User_manual_PocketBook_615_Plus_RU.pdf).
- 616 is a CIS-only device ([manual](http://support.pocketbook-int.com/fw/616/u/0/manual/cis/User_Manual_PocketBook_616_RU.pdf)), while the Basic Lux 2 is a worldwide device that goes under 616W ([manual](http://support.pocketbook-int.com/fw/616/u/0/manual/ww/User_Manual_Basic_Lux_2_BG.pdf)).

**Other:**
- @NiLuJe @Frenzie Should I get rid of the hasColorScreen and has3BytesWideFrameBuffer lines for the Color (633)?
- #6505 fixes might be included in this, if there's an answer.

Besides that, this is ready for review, can't test as always.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6525)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -246,7 +246,7 @@ local PocketBook614W = PocketBook:new{
     hasFewKeys = yes,
 }
 
--- PocketBook Basic Lux (615)
+-- PocketBook Basic Lux / 615 Plus (615/615W)
 local PocketBook615 = PocketBook:new{
     model = "PBBLux",
     display_dpi = 212,
@@ -255,7 +255,7 @@ local PocketBook615 = PocketBook:new{
     hasFewKeys = yes,
 }
 
--- PocketBook Basic Lux 2 (616)
+-- PocketBook Basic Lux 2 (616/616W)
 local PocketBook616 = PocketBook:new{
     model = "PBBLux2",
     display_dpi = 212,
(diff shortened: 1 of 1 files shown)


Frenzie:
> has3BytesWideFrameBuffer

I'd get rid of that one since it's not doing any good, but if I'm not mistaken hasColorScreen should be perfectly fine on any device. The only problem is that things would be rendered in 24-bit and subsequently downconverted to 8-bit. Which wastes memory and processing power but should otherwise be fine.

frontend/device/pocketbook/device.lua
```diff
@@ -402,12 +402,13 @@ elseif codename == "PocketBook 611" then
     return PocketBook611
 elseif codename == "PocketBook 613" then
     return PocketBook613
-elseif codename == "PocketBook 614W" or codename == "PocketBook 614" then
+elseif codename == "PocketBook 614" or codename == "PocketBook 614W" then
     return PocketBook614W
-elseif codename == "PocketBook 615" or codename == "PB615" then
+elseif codename == "PB615" or codename == "PB615W" then
+    codename == "PocketBook 615" or codename == "PocketBook 615W" or
```
Frenzie, line 408: Typo for this I imagine:
```suggestion
elseif codename == "PB615" or codename == "PB615W" or
    codename == "PocketBook 615" or codename == "PocketBook 615W" then
```

PR #6528, opened 2020-08-19
Add an option to *always* flash on chapter boundaries

NiLuJe:
Minor QoL change, because I like the concept ;).

Also, when this (or refresh rate is set to chapter) is enabled, also flash on the *second* page of a chapter, as chapter headings usually mean extra whitespace, meaning more chnace for visible ghosting/contrast deviations.

(Note that the current algo only flashes when paging *forward*, I didn't touch that).

### DIFF
--- a/frontend/apps/reader/modules/readertoc.lua
+++ b/frontend/apps/reader/modules/readertoc.lua
@@ -63,7 +63,10 @@ function ReaderToc:onPageUpdate(pageno)
     if UIManager.FULL_REFRESH_COUNT == -1 then
         if self:isChapterEnd(pageno, 0) then
             self.chapter_refresh = true
-        elseif self:isChapterBegin(pageno, 0) and self.chapter_refresh then
+        elseif self:isChapterStart(pageno, 0) and self.chapter_refresh then
+            UIManager:setDirty(nil, "full")
+            self.chapter_refresh = true
+        elseif self:isChapterSecondPage(pageno, 0) and self.chapter_refresh then
             UIManager:setDirty(nil, "full")
             self.chapter_refresh = false
         else
@@ -356,7 +359,7 @@ function ReaderToc:getPreviousChapter(cur_pageno, level)
     return previous_chapter
 end
 
-function ReaderToc:isChapterBegin(cur_pageno, level)
+function ReaderToc:isChapterStart(cur_pageno, level)
     local ticks = self:getTocTicks(level)
     local _begin = false
     for i = 1, #ticks do
@@ -368,9 +371,21 @@ function ReaderToc:isChapterBegin(cur_pageno, level)
     return _begin
 end
 
+function ReaderToc:isChapterSecondPage(cur_pageno, level)
+    local ticks = self:getTocTicks(level)
+    local _end = false
+    for i = 1, #ticks do
+        if ticks[i] + 1 == cur_pageno then
+            _end = true
+            break
+        end
+    end
+    return _end
+end
+
 function ReaderToc:isChapterEnd(cur_pageno, level)
     local ticks = self:getTocTicks(level)
-    local _end= false
+    local _end = false
     for i = 1, #ticks do
         if ticks[i] - 1 == cur_pageno then
             _end = true
@@ -390,7 +405,7 @@ function ReaderToc:getChapterPagesLeft(pageno, level)
 end
 
 function ReaderToc:getChapterPagesDone(pageno, level)
-    if self:isChapterBegin(pageno, level) then return 0 end
+    if self:isChapterStart(pageno, level) then return 0 end
     local previous_chapter = self:getPreviousChapter(pageno, level)
     if previous_chapter then
         previous_chapter = pageno - previous_chapter
--- a/frontend/apps/reader/modules/readertoc.lua
+++ b/frontend/apps/reader/modules/readertoc.lua
@@ -60,7 +60,7 @@ end
 
 function ReaderToc:onPageUpdate(pageno)
     self.pageno = pageno
-    if UIManager.FULL_REFRESH_COUNT == -1 then
+    if UIManager.FULL_REFRESH_COUNT == -1 or G_reader_settings:isTrue("refresh_on_chapter_boundaries") then
         if self:isChapterEnd(pageno, 0) then
             self.chapter_refresh = true
         elseif self:isChapterStart(pageno, 0) and self.chapter_refresh then
@@ -361,26 +361,26 @@ end
 
 function ReaderToc:isChapterStart(cur_pageno, level)
     local ticks = self:getTocTicks(level)
-    local _begin = false
+    local _start = false
     for i = 1, #ticks do
         if ticks[i] == cur_pageno then
-            _begin = true
+            _start = true
             break
         end
     end
-    return _begin
+    return _start
 end
 
 function ReaderToc:isChapterSecondPage(cur_pageno, level)
     local ticks = self:getTocTicks(level)
-    local _end = false
+    local _second = false
     for i = 1, #ticks do
         if ticks[i] + 1 == cur_pageno then
-            _end = true
+            _second = true
             break
         end
     end
-    return _end
+    return _second
 end
 
 function ReaderToc:isChapterEnd(cur_pageno, level)
--- a/frontend/ui/elements/refresh_menu_table.lua
+++ b/frontend/ui/elements/refresh_menu_table.lua
@@ -105,6 +105,12 @@ return {
             text = _("Every chapter"),
             checked_func = function() return refreshChecked(-1, -1) end,
             callback = function() UIManager:setRefreshRate(-1, -1) end,
+            separator = true,
+        },
+        {
+            text = _("Always refresh on chapter boundaries"),
+            checked_func = function() return G_reader_settings:isTrue("refresh_on_chapter_boundaries") end,
+            callback = function() G_reader_settings:flipNilOrFalse("refresh_on_chapter_boundaries") end,
         },
     }
 }


NiLuJe:
A more complex CRe-only algorithm could try to simply detect vertical rivers between pages by keeping track of the amount of *lines of text* rendered, and flash if there's a significant discrepancy between pages.

But, in practice, that only matter when scene breaks are involved, and it'd make handling the crazy people reading with vertical padding between paragraphs harder, maybe?

NiLuJe:
Speaking of scene breaks, I don't *think* we do that currently, but flashing when there's image content should probably be NTH, too.

I think all the bits necessary are here, because we already detect those cases for dithering.

NiLuJe:
(Those extra forced flashes *will* reset the refresh_rate counter).

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -60,10 +60,13 @@ end
...
 function ReaderToc:onPageUpdate(pageno)
     self.pageno = pageno
-    if UIManager.FULL_REFRESH_COUNT == -1 then
+    if UIManager.FULL_REFRESH_COUNT == -1 or G_reader_settings:isTrue("refresh_on_chapter_boundaries") then
         if self:isChapterEnd(pageno, 0) then
             self.chapter_refresh = true
-        elseif self:isChapterBegin(pageno, 0) and self.chapter_refresh then
+        elseif self:isChapterStart(pageno, 0) and self.chapter_refresh then
+            UIManager:setDirty(nil, "full")
+            self.chapter_refresh = true
```
Galunid, line 68: shouldn't it be `false`? Otherwise it's not needed, since the requirement to enter this if is `self.chapter_refresh == true`

NiLuJe: It's indeed redundant, but I kept it to keep the ladder "symmetric" in order to make the logic more obvious ;).

NiLuJe: I'm half of a mind to zap that heuristic and just flash on start/second no matter the pagination direction...

NiLuJe: Mildly intrigued that luacheck doesn't catch that as a dead store, by the way...

poire-z:
I didn't know we had that :/ I may like it, dunno yet :)

> A more complex CRe-only algorithm could try to simply detect vertical rivers between pages by keeping track of the amount of lines of text rendered, and flash if there's a significant discrepancy between pages.

That might be too dependant on the content. You may get the same number of lines, but have a little shift because of some bigger font size header or just sup/sub that shifted the line height for one single line, or some small padding/margin added to DIV/P/BLOCKQUOTE containers by the publisher.
Althought it might be fine for a regular styletweaked novel, I'd prefer not to have to add a `self.ui.document:getDrawnTextStatistics()` that might not have a real scientific base to it :)

If I end up liking it, I guess the only little annoying thing is that for the kind of book I'm reading currently, with a heading every 2 pages (so, a chapter, but I don't mind having 1000 chapters and a crowder progress bar), it will flash every 2 page - and the only way to disable that is the toggle you added - which is global. So, I'll have to toggle the global setting back on when switching to another book. Some local override ("Disable for current book") might come handy later (if I really like it :)

NiLuJe:
> That might be too dependant on the content.

Agreed, hence my next § ;). That was mainly just me thinking out loud ;).

> If I end up liking it, I guess the only little annoying thing is that for the kind of book I'm reading currently, with a heading every 2 pages (so, a chapter, but I don't mind having 1000 chapters and a crowder progress bar), it will flash every 2 page - and the only way to disable that is the toggle you added - which is global. So, I'll have to toggle the global setting back on when switching to another book. Some local override ("Disable for current book") might come handy later (if I really like it :).

Well, it is global by design ;p.

But that ties in to something I wanted to ask @yparitcher: what do I need to do to make this accessible to dispatcher/profile, because that'd take care of @poire-z's concerns nicely.

taosxx:
> Also, when this (or refresh rate is set to chapter) is enabled, also flash on the _second_ page of a chapter, as chapter headings usually mean extra whitespace, meaning more chance for visible ghosting/contrast deviations.

Please, make it possible to override this manually without having to compile KOreader. For me, there's nothing more distracting in Nickel while reading than the flash between the first and second page of the chapter (okay, a complete flash after every 6 page turns is worse ;-) ). For some reason, I never read books that would benefit from a second flash -- even if high-contrast ornaments or images are used as chapter headings, for me, some ghosting is much less intrusive than the flash afterwards.

yparitcher:
> But that ties in to something I wanted to ask @yparitcher: what do I need to do to make this accessible to dispatcher/gesture/profile, because that'd take care of @poire-z's concerns nicely.

1. make sure what you want to change is callable via an event
2. add a entry to this table following the comment at the beginning. https://github.com/koreader/koreader/blob/01ed79fb9237ace0c6f6f2f29f7db1ced753b5b5/frontend/dispatcher.lua#L33
3. add a entry to this table for where in the menu order you want it displayed. https://github.com/koreader/koreader/blob/01ed79fb9237ace0c6f6f2f29f7db1ced753b5b5/frontend/dispatcher.lua#L143

What is the best way to add these instructions to the docs for the future?

yparitcher:
once you make a event/dispatcher for this you might want to add one for the regular refresh frequency :)

NiLuJe:
Okay, good to know I got it right in #6530 then ;p.

So, yeah, have to make new events for those.

NiLuJe:
> 
> 
> > Also, when this (or refresh rate is set to chapter) is enabled, also flash on the _second_ page of a chapter, as chapter headings usually mean extra whitespace, meaning more chance for visible ghosting/contrast deviations.
> 
> Please, make it possible to override this manually without having to compile KOreader. For me, there's nothing more distracting in Nickel while reading than the flash between the first and second page of the chapter (okay, a complete flash after every 6 page turns is worse ;-) ). For some reason, I never read books that would benefit from a second flash -- even if high-contrast ornaments or images are used as chapter headings, for me, some ghosting is much less intrusive than the flash afterwards.

What's your actual setup? `refresh_rate`set to `-1`?

There's no silver bullet here.

I can either create an inconsistency by having rf -1 behave as usual, but flash_on_boundaries flash on both, or add *two* new settings to handle botch cases: rf -1 for the usual, rf -2 for both, flash_on_chapter for rf + rf -1 behavior, flash_on_boundaries for rf + rf -2.

I'm not particularly happy about either, but since I don't personally use rf -1, the selfish and easy solution would be the former.

poire-z:
Bit strange to have one of these positive and the other negative:
```
[ } Always flash on chapter boundaries
[ } Don't flash on the second page of a new chapter
```

`[ ] Also flash on a chapter 2nd page` would read better (with may be a enabled_func to have it greyed out when the first one is not enabled ?

NiLuJe:
> Bit strange to have one of these positive and the other negative:
> 
> ```
> [ } Always flash on chapter boundaries
> [ } Don't flash on the second page of a new chapter
> ```
> 
> `[ ] Also flash on a chapter 2nd page` would read better (with may be a enabled_func to have it greyed out when the first one is not enabled ?

That's the new default behavior. I'm not a fan of inverted settings (i.e., whose defaults is nil, which would be necessary to make this behavior default, which I do want ;)).

NiLuJe:
Okay, managed not to break anything when switching to Events, so this should be good to go ;).

poire-z:
Just mentionning that (as I don't mind much this strange phrasing as we won't visit that menu often):
The checkbox does not have to reflect exactly the true/false value of the internal setting.
By negating the checked_func, you could use the positive phrasing (and if you don't want it to be checked even when not enabled for aesthetic reason, include the enabled tests into the checked_func tests.

NiLuJe:
> Just mentionning that (as I don't mind much this strange phrasing as we won't visit that menu often):
> The checkbox does not have to reflect exactly the true/false value of the internal setting.
> By negating the checked_func, you could use the positive phrasing (and if you don't want it to be checked even when not enabled for aesthetic reason, include the enabled tests into the checked_func tests.

True, but I'm also not a fan of those either, and I managed to subtly bork stuff when doing just that in the past ;p.

Galunid:
Sidenote, would you be interested in adding this to `Reader timer` (something like `Till the next chapter`)?

NiLuJe:
> Sidenote, would you be interested in adding this to `Reader timer` (something like `Till the next chapter`)?

Not sure I follow? (I'm wholly unfamiliar with the readtimer plugin).

frontend/dispatcher.lua
```diff
@@ -66,6 +66,15 @@ local settingsList = {
     toggle_rotation = { category="none", event="SwapRotation", title=_("Toggle orientation"), device=true,},
     invert_rotation = { category="none", event="InvertRotation", title=_("Invert rotation"), device=true,},
     iterate_rotation = { category="none", event="IterateRotation", title=_("Rotate by 90° CW"), device=true, separator=true,},
+    set_refresh_rate = { category="absolutenumber", event="SetBothRefreshRates", min=-1, max=200, title=_("Flash every %1 pages (always)"), device=true,},
+    set_day_refresh_rate = { category="absolutenumber", event="SetDayRefreshRate", min=-1, max=200, title=_("Flash every %1 pages (not in nightmode)"), device=true,},
+    set_night_refresh_rate = { category="absolutenumber", event="SetNightRefreshRate", min=-1, max=200, title=_("Flash every %1 pages (in nightmode)"), device=true,},
+    set_flash_on_chapter_boundaries = { category="none", event="SetToggleFlashOnChapterBoundaries", title=_("Always flash on chapter boundaries"), device=true,},
+    unset_flash_on_chapter_boundaries = { category="none", event="UnsetToggleFlashOnChapterBoundaries", title=_("Never flash on chapter boundaries"), device=true,},
```
yparitcher, line 73: you might want to combine these into one event that takes on/off true/false, and have one menu item of category="string" for both true and false.

taosxx:
> What's your actual setup? `refresh_rate`set to `-1`?

Exactly.

frontend/dispatcher.lua
NiLuJe: Oh, yeah, that's much better, thanks :).

NiLuJe: Also made me realize I'd typo'd the event name anyway :D.

frontend/dispatcher.lua
```diff
@@ -66,6 +66,13 @@ local settingsList = {
     toggle_rotation = { category="none", event="SwapRotation", title=_("Toggle orientation"), device=true,},
     invert_rotation = { category="none", event="InvertRotation", title=_("Invert rotation"), device=true,},
     iterate_rotation = { category="none", event="IterateRotation", title=_("Rotate by 90° CW"), device=true, separator=true,},
+    set_refresh_rate = { category="absolutenumber", event="SetBothRefreshRates", min=-1, max=200, title=_("Flash every %1 pages (always)"), device=true,},
+    set_day_refresh_rate = { category="absolutenumber", event="SetDayRefreshRate", min=-1, max=200, title=_("Flash every %1 pages (not in nightmode)"), device=true,},
+    set_night_refresh_rate = { category="absolutenumber", event="SetNightRefreshRate", min=-1, max=200, title=_("Flash every %1 pages (in nightmode)"), device=true,},
```
Frenzie, line 71:
```suggestion
    set_day_refresh_rate = { category="absolutenumber", event="SetDayRefreshRate", min=-1, max=200, title=_("Flash every %1 pages (not in night mode)"), device=true,},
    set_night_refresh_rate = { category="absolutenumber", event="SetNightRefreshRate", min=-1, max=200, title=_("Flash every %1 pages (in night mode)"), device=true,},
```

I notice you snuck another one in and I overlooked it:
> Disable the automagic inversion of images when nightmode is enabled. Useful if your book contains mainly inlined mathematical content or scene break art.

I don't particularly care if we use "night mode" or "nightmode," but the former is probably slightly more idiomatic. Plus you'd have to chance the 4 night modes. ;-P

Frenzie (approved):

poire-z:
> (Can't we find some heuristic for such book so that it doesn't flash on every page ? :)

Just for info, I helped myself with:
```diff
--- frontend/apps/reader/modules/readertoc.lua
+++ frontend/apps/reader/modules/readertoc.lua
@@ -53,2 +53,3 @@
     self.expanded_nodes = {}
+    self.avoid_chapter_flash = nil
 end
@@ -62,2 +63,10 @@
     if UIManager.FULL_REFRESH_COUNT == -1 or G_reader_settings:isTrue("refresh_on_chapter_boundaries") then
+        if self.avoid_chapter_flash == nil then
+            -- No flash on chapter if too many ticks
+            self.avoid_chapter_flash = (#self:getTocTicks(0) / self.ui.document:getPageCount()) > 0.3
+        end
+        if self.avoid_chapter_flash then
+            self.pageno = pageno
+            return
+        end
         local flash_on_second = G_reader_settings:nilOrFalse("no_refresh_on_second_chapter_page")
```
Not a super reliable check (if most of all the ticks are in some small part of the book), but good enough to not have it with Wikipedia EPUBs.

NiLuJe:
Another approach would be counting consecutive chapter  flashes (or near consecutive flashes), and throttle them. The event handler where the current heuristic is should have all the info.

PR #6530, opened 2020-08-19
Make the manual rotation events useful

### LINKED ISSUE #6524: 'Toggle rotation': toggle back doesn't work
* KOReader version: v2020.07.1-51-g5a7ecab_2020-08-15/ 
* Device: KA1

#### Issue
I have 'Toggle rotation' in 'Tap corner' > 'Top left' > Device, and then one of the 4 rotation options.

When I toggle back I got stuck. It looks like the backrotation & the first rotation cancel out each other.

NiLuJe:
Restore a Swap event like before #6309
Implement an Invert one
And keep the fairly useless +90° CW one, but under an accurate name.

Fix #6524

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6530)


NiLuJe:
(As of yet untested, but the logic feels sound ;p).

NiLuJe:
Yep, it works.

Sidebar: Modifying a gesture only appears to take after a restart. Is it just me? Ping @yparitcher / @Frenzie

yparitcher:
> Sidebar: Modifying a gesture only appears to take after a restart. Is it just me? Ping @yparitcher / @Frenzie

not as far as i remember

yparitcher:
> Yep, it works.
> 
> Sidebar: Modifying a gesture only appears to take after a restart. Is it just me? Ping @yparitcher / @Frenzie

If it is a gestures issue, than switching between reader/FM reloads gestures like a restart, so you can use that to debug.

Frenzie (approved):
Looks good, didn't test.

NiLuJe:
> 
> 
> I think toggle sounds better than swap

I'll probably settle on `Toggle orientation` ;).

NiLuJe:
> If it is a gestures issue, than switching between reader/FM reloads gestures like a restart, so you can use that to debug.

That does work. Still puzzled why it doesn't take immediately, though.

Frenzie:
I think toggle's sensible enough. That's between portrait & landscape or reverse portrait & reverse landscape respectively, right?

NiLuJe:
> 
> 
> I think toggle's sensible enough. That's between portrait & landscape or reverse portrait & reverse landscape respectively, right?

Yup!

yparitcher:
> Sidebar: Modifying a gesture only appears to take after a restart. Is it just me? Ping @yparitcher / @Frenzie

Can't reproduce, Is it with a specific gesture or a specific action?

frontend/dispatcher.lua
```diff
@@ -62,8 +62,10 @@ local settingsList = {
     reboot = { category="none", event="Reboot", title=_("Reboot the device"), device=true, condition=Device:canReboot(),},
     poweroff = { category="none", event="PowerOff", title=_("Power off"), device=true, condition=Device:canPowerOff(), separator=true,},
     show_menu = { category="none", event="ShowMenu", title=_("Show menu"), device=true,},
-    toggle_hold_corners = { category="none", event="IgnoreHoldCorners", title=_("Toggle hold corners"), device=true,},
-    toggle_rotation = { category="none", event="ToggleRotation", title=_("Toggle rotation"), device=true, separator=true,},
+    toggle_hold_corners = { category="none", event="IgnoreHoldCorners", title=_("Toggle hold corners"), device=true, separator=true,},
+    swap_rotation = { category="none", event="SwapRotation", title=_("Toggle orientation"), device=true,},
```
yparitcher, line 66: can you change this back to `toggle_rotation` to mach the title and that way the event does not go away for users.

NiLuJe:
> > Sidebar: Modifying a gesture only appears to take after a restart. Is it just me? Ping @yparitcher / @Frenzie
> 
> Can't reproduce, Is it with a specific gesture or a specific action?

Randomly tested on a (FM) corner tap gesture with nothing/any of those rota ones. The initial switch from nothing to it is... weird. First tap rotates, then nada. Going from something to nothing doesn't exhibit the same weirdness (i.e., still rotates).

yparitcher:
> > > Sidebar: Modifying a gesture only appears to take after a restart. Is it just me? Ping @yparitcher / @Frenzie
> > 
> > 
> > Can't reproduce, Is it with a specific gesture or a specific action?
> 
> Randomly tested on a (FM) corner tap gesture with nothing/any of those rota ones. The initial switch from nothing to it is... weird. First tap rotates, then nada. Going from something to nothing doesn't exhibit the exact same weirdness (i.e., fully keeps the previous gesture until restart/reload).

can confirm, when testing on the rotation one

yparitcher:
Does not happen in the reader :(

yparitcher:
> Sidebar: Modifying a gesture only appears to take after a restart. Is it just me?

should be fixed by #6532

PR #6535, opened 2020-08-20
Revert "Kosync network callback"

Galunid:
Reverts koreader/koreader#6489 since it introduced problems for people having `[*]Auto sync now and in the furture` & `Action when wifi off: prompt`.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6535)

### DIFF
--- a/plugins/kosync.koplugin/KOSyncClient.lua
+++ b/plugins/kosync.koplugin/KOSyncClient.lua
@@ -1,4 +1,3 @@
-local NetworkMgr = require("ui/network/manager")
 local UIManager = require("ui/uimanager")
 local DEBUG = require("dbg")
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
Now that I'm back, remind me what was the issue that made you try this approach?

Galunid:
@NiLuJe Kosync plugin doesn't adapt to your QoL wifi improvements, i.e. doesn't enable wifi when pushing/pulling progress and disabling it after. It just displays message along the lines of "Something went wrong, check your network"

NiLuJe:
@Galunid: On which device?

NiLuJe:
Ah, yeah, `KOSync:getProgress` and `KOSync:updateProgress` may be called without Wi-Fi (i.e., pagination).

There's not really much we can do, as they may be called at a *lot* of weird moments, and, the rerunWhenOnline stuff will (as it should) always show the relevant UI bits.

That said, on Kobo, with restore Wi-Fi on resume, it *should* shoudl behave automatically, as it's frequent enough (I think?), that it should keep Wi-Fi alive.

NiLuJe:
Frequent enough given an appropriate `DAUTO_SAVE_PAGING_COUNT`. Since that's gone now, a better fix would probably to make that threshold an actual plugin setting ;).

Galunid:
Yes, actually that's the reason I disabled `Auto progress sync`, since those weird times hung the device for ~30s. A better way to handle this would be to check if `manual == true` was passed when calling https://github.com/koreader/koreader/blob/1b904f0f0df731c3949063f54173687008e3b1a8/plugins/kosync.koplugin/main.lua#L533-L545
So if user called it, enable wifi, otherwise skip.

Issue #5550, opened 2019-10-31
PocketBook Inkpad 3 Pro: Everything is weird and slow


kindfulkirby:
* KOReader version: 2019.10
* Device: PocketBook Inkpad 3 Pro

#### Issue
When I open KOReader, I get the popup with the circled i and it says it's loading `./help/quickstart-en-v2019.10.html`.
Then nothing happens.
After a few minutes, the Inkpad's LED goes out, and a while later it turns off (as usual).

_Update_: KOReader shows up in the Inkpad's task manager, and when I use it to switch to KOReader, it just shows a blank screen.

#### Steps to reproduce

Open KOReader on Inkpad 3 Pro.

##### `crash.log`
```
---------------------------------------------
                launching...
  _  _____  ____                _
 | |/ / _ \|  _ \ ___  __ _  __| | ___ _ __
 | ' / | | | |_) / _ \/ _` |/ _` |/ _ \ '__|
 | . \ |_| |  _ <  __/ (_| | (_| |  __/ |
 |_|\_\___/|_| \_\___|\__,_|\__,_|\___|_|

 It's a scroll... It's a codex... It's KOReader!

 [*] Current time: 10/31/19-06:08:29
 [*] Version: v2019.10

ffi.load: SDL2
ffi.load (warning): libSDL2.so: cannot open shared object file: No such file or directory
ffi.load: libSDL2-2.0.so
ffi.load (warning): libSDL2-2.0.so: cannot open shared object file: No such file or directory
ffi.load: libSDL2-2.0.so.0
ffi.load (warning): libSDL2-2.0.so.0: cannot open shared object file: No such file or directory
SDL2 not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: libSDL2-2.0.so.0
ffi.load: inkview
10/31/19-06:08:29 INFO  SoftwareVersion:  U740-2.5.20.1155
10/31/19-06:08:29 INFO  Blacklisting the C BB on this device
Tweaking LuaJIT's max loop unroll factor (15 -> 45)
PB FB: smem_len    : 2637824
PB FB: line_length : 1872
PB FB: xres        : 1872
PB FB: xres_virtual: 1872
PB FB: yres        : 1404
PB FB: yres_virtual: 1404
PB FB: bpp         : 8
PB FB: xres_virtual -> 1888
PB FB: yres_virtual -> 1408
PB FB: line_length -> 1888
PB FB: yres_virtual => 1408
PB FB: line_length <- 1872
[...]

PR #6541, opened 2020-08-22
wip: pocketbook fixes

### LINKED ISSUE #3477: PB: Book-State not saved on powerdown
* KOReader version: debug-v2015.11-1404-g15157536
* Device: PB Inkpad 2

#### Issue
It seems like putting the PB in sleep-mode does not update the Book-State: The PocketBook will sooner or later power down fully when in sleep-mode.
After booting again the page-count is lost, and the book starts at page one (or another, previously saved state). I made it a habit to exit the application before putting the PB to sleep, in this way the state is preserved.

#### Steps to reproduce
Open a book, skip some pages, put the PB to sleep, wait till it powers down (is it even supposed to do this? My PB-setting is 'Power down after 10 Minutes', but this is surely not whats happening - the power down comes much later - maybe it is in fact a crash?)

[crash.log](https://github.com/koreader/koreader/files/1463823/crash.log)

### LINKED ISSUE #3561: PocketBook: suspension and timer-behaviour
While dealing with https://github.com/koreader/koreader/issues/3419 I noticed that at least the Inkpad 2 does a real suspend approx. two seconds after the last key- or touch-event if certain conditions are met*. That means no [BackgroundRunner-Tasks](https://github.com/koreader/koreader/tree/a16c500ac745ed956f4afa6122ddd3b656e22829/plugins/backgroundrunner.koplugin), no [UIManager:scheduleIn()-Timers](https://github.com/koreader/koreader/blob/a16c500ac745ed956f4afa6122ddd3b656e22829/frontend/ui/uimanager.lua#L292) etc, leading to various issues like 

- https://github.com/koreader/koreader/issues/3419 (suspend while loading a book)
- Display glitches:  
  - reader suspends while in a page-redraw
  - display elements not updated (e.g. statusbar-time)
  - screensaver sometimes does not show the title-page because the redraw comes too late
- unresponsiveness in menus as wake-up, albeit being fast, is noticeable
- "unexpected" keypresses, as multiple touches while in suspend are not executed momentarily but seconds later in succession when the reader woke up.
- possible malbehaviour with plugins which depend on timed tasks.

*If the reader is connected to Wifi or USB the issues do not occur because auto-suspend does for some reason not kick in.

Most of the issues are just annoying and may be overlooked as bad performance (and are 'fixed' on the next key-/touch-event). Nevertheless i am a bit surprised that this has never come up before, leaving me wondering if it is a model- or firmware-specific issue.

So I am interested if other PB-users experience the same (**please post you Firmware-Version**). Regarding possible solutions i am tempted to disable the suspension completely, but this will affect battery-time quite a bit. An other way would be to wrap time-critical sections in ``setSuspendState()``-blocks disabling and re-enabling the suspension if needed.

### LINKED ISSUE #5114: Strange graphics glitches on Pocketbook 622
* KOReader version: v2019.06
* Device: Pockebook 622
* FW: E622.4.4.566

#### Issue
Doesn't start.

#### Steps to reproduce
Setup according to instructions, try to start.

##### `crash.log`
```
 [*] Current time: 07/10/19-21:55:51
 [*] Version: v2019.06

ffi.load: SDL2
ffi.load (warning): libSDL2.so: cannot open shared object file: No such file or directory
ffi.load: libSDL2-2.0.so
ffi.load (warning): libSDL2-2.0.so: cannot open shared object file: No such file or directory
SDL2 not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: libSDL2-2.0.so
ffi.load: SDL
ffi.load (warning): libSDL.so: cannot open shared object file: No such file or directory
SDL not loaded:	./setupkoenv.lua:27: Not able to load dynamic library: SDL
ffi.load: inkview
07/10/19-21:55:51 INFO  SoftwareVersion:  E622.4.4.566
./luajit: frontend/device/pocketbook/device.lua:352: unrecognized PocketBook model PocketBook 622
stack traceback:
	[C]: in function 'error'
	frontend/device/pocketbook/device.lua:352: in main chunk
	[C]: in function 'probeDevice'
	frontend/device.lua:48: in main chunk
	[C]: in function 'require'
	./reader.lua:55: in main chunk
	[C]: at 0x0000c4a9
```

### LINKED ISSUE #5140: Blank screen when reader wakes up (Pocketbook 740)
* KOReader version:
koreader-pocketbook-arm-obreey-linux-gnueabi-v2019.07.zip

* Device:
    Pocketbook 740 (InkPad 3)
    Firmware: sw_20190709_U740_5.20.1010_user.zip

#### Issue
Screen stays blank when the device gets awaken from the sleep mode.

#### Steps to reproduce
- Open a book;
- Set screen saver option to "Leave screen as it is";
- Press the power button to put the device to sleep;
- Press the power button to wake it up.;
Result: The screen is blank. 

- Touch the screen or refresh it with a swipe. 
Result: Text of the book is visible now.

### LINKED ISSUE #6536: [Pocketbook]: screensaver behaviour
* KOReader version: master
* Device: PocketBook

#### Issue

The behaviour of the suspend/screensaver on Pocketbooks is broken. Since we don't handle power functions on the platform I went and disable screensaver support, but the "sleeping" screen is still displayed on two different ocasions:

- when pressing the home button to return to the launcher
- when pressing the power button to sleep the device.

In both the event `EVT_BACKGROUND` is sent by inkview and we handle that in `onPowerEvent` in `device/generic/device.lua`.

There are a few issues that are related to this one but were masked by the "wonky" suspend behaviour.

#### Steps to reproduce

Disable screensaver support (and a few other unrelated things on the diff):

```diff
diff --git a/frontend/device/pocketbook/device.lua b/frontend/device/pocketbook/device.lua
index 10302761..2a621f3a 100644
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -59,18 +59,20 @@ local PocketBook = Generic:new{
     model = "PocketBook",
     isPocketBook = yes,
     isInBackGround = false,
+    hasExitOptions = no,
     hasOTAUpdates = yes,
     hasWifiToggle = yes,
     isTouchDevice = yes,
     hasKeys = yes,
     hasFrontlight = yes,
+    canRestart = no,
     canSuspend = no,
     emu_events_dev = "/dev/shm/emu_events",
     home_dir = "/mnt/ext1",
 
     -- all devices that have warmth lights use inkview api
     hasNaturalLightApi = yes,
-
+    needsScreenRefreshAfterResume = no,
 }
 
 -- Make sure the C BB cannot be used on devices with a 24bpp fb
@@ -163,8 +165,6 @@ function PocketBook:init()
     Generic.init(self)
 end
 
-function PocketBook:supportsScreensaver() return true end
-
 function PocketBook:setDateTime(year, month, day, hour, min, sec)
     if hour == nil or min == nil then return true end
     local command
```

#### How it looks

VID_20200820_141633736

#### Expected behaviour

Exactly the same, without the sleeping message, as is not true when returning home and will be overriden by PB screensaver when going to sleep.

Any non orthodox workaround is fine if that makes the sleeping message disapear :)

pazos:
Requires https://github.com/koreader/koreader-base/pull/1164

Fixes #6536
Fixes #5140
Fixes #3561
Fixes #3477

Should fix #5114 (but a confirmation is needed :))

Changes:

- Pocketbook suspend behaviour is disabled when app is on the foreground. It can be reenabled based on user preferences (still need to implement that on menus). Probably that's not needed for most users since the Pocketbook cpu suspend stub won't save much power compared to a sensible user management.

- Read state is saved each time the system is going to poweroff

- Screensaver is disabled since we don't manage power states.

~~I will upload a test build soon~~

Test build: https://www.dropbox.com/s/wouijkf0g8989xu/koreader-pocketbook-v2020.03.2-260-ga4cfdf3f_2020-08-22.zip?dl=0

It should fix all issues reported in https://github.com/koreader/koreader/issues/5550#issuecomment-670827065 (except the slowness of cpu/io compared to modern Kobos/Cervantes and even Android). Pinging @Cellaris, feel free to test and report. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6541)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -59,11 +59,13 @@ local PocketBook = Generic:new{
     model = "PocketBook",
     isPocketBook = yes,
     isInBackGround = false,
+    hasExitOptions = no,
     hasOTAUpdates = yes,
     hasWifiToggle = yes,
     isTouchDevice = yes,
     hasKeys = yes,
     hasFrontlight = yes,
+    canRestart = no,
     canSuspend = no,
     emu_events_dev = "/dev/shm/emu_events",
     home_dir = "/mnt/ext1",
@@ -71,6 +73,7 @@ local PocketBook = Generic:new{
     -- all devices that have warmth lights use inkview api
     hasNaturalLightApi = yes,
 
+    needsScreenRefreshAfterResume = no,
 }
 
 -- Make sure the C BB cannot be used on devices with a 24bpp fb
@@ -120,6 +123,11 @@ function PocketBook:init()
                     self.isInBackGround = false
                     return "Resume"
                 end
+                self:setSuspendState("resume")
+            elseif ev.code == EVT_SHOW then
+                self:setSuspendState("show")
+            elseif ev.code == EVT_HIDE then
+                return "SaveState"
             end
         end,
     }
@@ -135,10 +143,12 @@ function PocketBook:init()
             ev.type = 1 -- linux/input.h Key-Event
         end
 
-        -- handle EVT_BACKGROUND and EVT_FOREGROUND as MiscEvent as this makes
+        -- handle app state events as misc events as this makes
         -- it easy to return a string directly which can be used in
         -- uimanager.lua as event_handler index.
-        if ev.type == EVT_BACKGROUND or ev.type == EVT_FOREGROUND then
+        if ev.type == EVT_BACKGROUND or ev.type == EVT_FOREGROUND or
+            ev.type == EVT_SHOW or ev.type == EVT_HIDE
+        then
             ev.code = ev.type
             ev.type = 4 -- handle as MiscEvent, see above
         end
@@ -163,7 +173,22 @@ function PocketBook:init()
     Generic.init(self)
 end
 
-function PocketBook:supportsScreensaver() return true end
+-- system sleep is disabled when the app is in the foreground.
+-- This is handled automatically based on application state.
+-- see https://github.com/koreader/koreader-base/blob/master/input/input-pocketbook.h
+--
+-- reenable it based on user preferences, to be used on handleMiscEv
+
+function PocketBook:setSuspendState(state)
+    local msg = string.format("system sleep on %s", state)
+    if G_reader_settings:isTrue("pocketbook_timeout") then
+        logger.info("enabling " .. msg)
+        local ms = G_reader_settings:readSetting("pocketbook_timeout_ms") or 5 * 60 * 1000
+        require("ffi/input").setSuspendState(1, ms)
+    else
+        logger.info("ignoring " .. msg .. ". Timeout is disabled")
+    end
+end
 
 function PocketBook:setDateTime(year, month, day, hour, min, sec)
     if hour == nil or min == nil then return true end
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -82,10 +82,10 @@ function UIManager:init()
     if Device:isPocketBook() then
         self.event_handlers["Suspend"] = function()
             self:_beforeSuspend()
-            Device:onPowerEvent("Power")
         end
         self.event_handlers["Resume"] = function()
-            Device:onPowerEvent("Power")
+            self:setDirty("all", "full")
+            self:forceRePaint()
             self:_afterResume()
         end
     end


LucyOne:
Excellent job! Will test on InkPad 3 and Touch HD 3 with a book that kept freezing. Looking good so far.

pazos:
@LucyOne @Cellaris: I'm interested in battery reports too. If you feel the program drains the battery faster to a significant degree please comment. There's no way to change defaults right now other than edit the lua settings file directly. In case you want to try re-enabling suspend after a certain time you can set "pocketbook_timeout" to true and "pocketbook_timeout_ms" to the number of milliseconds before suspending.

If pocketbook_timeout is enabled and pocketbook_timeout_ms not set the effective timeout will happen 5 minutes after each session starts.

LucyOne:
Your version of KoReader is really snappy, so far no problems. The menus react fine, changing the font size is also faster.
Both my devices are in sleep covers, so they go to sleep when I close the covers. That seems to work well so far, also waking up. I don't know why I would need to set a timeout value.

pazos:
Superseded by #6559

PR #6557, opened 2020-08-24
Make sure "Clear history of deleted files" actually persists.

NiLuJe:
The current code was only clearing the live table, it was forgotten
after a restart.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6557)

### DIFF
--- a/frontend/readhistory.lua
+++ b/frontend/readhistory.lua
@@ -200,7 +200,7 @@ function ReadHistory:clearMissing()
     assert(self ~= nil)
     for i = #self.hist, 1, -1 do
         if self.hist[i].file == nil or lfs.attributes(self.hist[i].file, "mode") ~= "file" then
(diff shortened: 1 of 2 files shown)


poire-z:
Any idea how the "Huh?" happened ? it looks like init() > reload() > sort() should ensure all items have an item.index.

NiLuJe:
With autoremove, that's no longer true, apparently.

poire-z:
So, `ReadHistory:removeItemByPath(path)` not updating the index, may be missing a `self:_indexing(removed_item.index)` ?

NiLuJe:
Nope, just that the remove is done before the indexing, so there's no index ;).

Frenzie (approved):
Huh.

poire-z (approved):

Issue #6545, opened 2020-08-22
[Pocketbook] RFC: Input driver rewrite for direct polling


ezdiy:
The code at https://gist.github.com/ezdiy/a1e300bc05160032ea8390781f52abdd is a proof of concept rewrite of pocketbook frontend driver.

- Uses undocumented inkview polling mode running on same thread. The input side is pure ffi, similar to sdl and android.
- ~Can precisely suspend, that is, no freezes when doing something else. Currently the suspend is set to very aggressive (2s delay) for testing. Down the line, adaptive delay can be done.~ Adaptive suspend implemented.
- Some other minor features (proper multitouch handling on newer devices etc)
- ~Still rather buggy, sometimes crashing in callbacks or entering infinite loops, usually when dealing screen lock/unlock~ It's in works for me with no issues (TM) state. Still, YMMV.
- Inkview can be called directly without data races or awkward message passing

Testing would be welcome for older devices, to figure out whether this approach is workable. The point is to get rid of InkViewMain - a massive source of headaches and C boilerplate code.

ezdiy:
In rev4:
* The hangs/crashes should be now all fixed. Turns out this was related to the weird suspend/resume spoofing, in same vein as #6536.
* Implemented adaptive suspend. This is a heuristic to have cake (battery life), and it eat too (no lag). Every tap within a time window prolongs a window of time for which we *ban* suspends, as well as the window within which further taps are accepted. If a tap occurs beyond current window, we reset the window back to 5s (ie suspend becomes very aggressive again). The effect of all this is that when you're paging through a book, doing something in UI, you won't be annoyed by suspend as it never has a chance to kick in except between first few tap or two. However once you start reading a book, the autosuspend goes in effect again - as you're quite likely to just tap once every 30 seconds, thus saving battery life. Finally, when battery is below some threshold (current 60%), we disable this logic and make autosuspend aggressive as is the default.

Things that are some more conceptual issues are now labeled FIXME, there are direct explanations about those in the code.

roshavagarga:
@ezdiy Will be able to test out today, you reverted the 515 frontlight change between revisions though?

roshavagarga:
@ezdiy Tested this on a 614W, seems to work without any issues so far, minus ones that existed beforehand like being unable to exit the Statistics window, due to the device having hasDPad/hasFewKeys. On that subject, I'd welcome an explanation on how those are added, as I have a few older devices I might add definitions for.

Frenzie:
[...]

PR #6558, opened 2020-08-24
Make UIManager track prevent/allowStandby state.

ezdiy:
Conversely, Trapper and plugins report standby interruptability.

Ground work for #6545, effectively no-op now. To test it, run in debug mode and see if the usual places trigger standby bans as they should, fe for covers:

```
$ strings crash.log |egrep 'standby|BG'
08/25/20-01:31:51 DEBUG prevent standby
08/25/20-01:31:51 DEBUG   BG extraction started
08/25/20-01:31:51 DEBUG   BG extracting: ...
08/25/20-01:31:51 DEBUG   BG extracting: ...
08/25/20-01:31:56 DEBUG   BG extraction done
08/25/20-01:32:01 DEBUG allow standby
```

~Of special note is Trapper:info() - while not background process stuff like the rest, it is used for "tap me to abort this" of unattended loops, so we need to prevent standby in there as well.~ Trapper now nukes standby wholesale. This is coarse, but simple and probably safer to catch possible corner cases.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6558)

### DIFF
--- a/frontend/ui/trapper.lua
+++ b/frontend/ui/trapper.lua
@@ -43,8 +43,10 @@ function Trapper:wrap(func)
     -- Catch and log any error happening in func (an error happening
     -- in a coroutine just aborts silently the coroutine)
     local pcalled_func = function()
+        UIManager:preventStandby()
         -- we use xpcall as it can give a whole stacktrace, unlike pcall
         local ok, err = xpcall(func, debug.traceback)
+        UIManager:allowStandby()
         if not ok then
             logger.warn("error in wrapped function:", err)
             return false
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -37,6 +37,8 @@ local UIManager = {
     _refresh_func_stack = {},
     _entered_poweroff_stage = false,
     _exit_code = nil,
+    _prevent_standby_count = 0,
+    _prev_prevent_standby_count = 0,
 
     event_hook = require("ui/hook_container"):new()
 }
@@ -80,12 +82,11 @@ function UIManager:init()
         end)
     end
     if Device:isPocketBook() then
+        -- Only fg/bg state plugin notifiers, not real power event.
         self.event_handlers["Suspend"] = function()
             self:_beforeSuspend()
-            Device:onPowerEvent("Power")
         end
         self.event_handlers["Resume"] = function()
-            Device:onPowerEvent("Power")
             self:_afterResume()
         end
     end
@@ -1189,6 +1190,11 @@ function UIManager:handleInput()
         wait_us = math.min(wait_us or math.huge, self.ZMQ_TIMEOUT)
     end
 
+    -- If allowed, entering standby (from which we can wake by input) must trigger in response to event
+    -- this function emits (plugin), or within waitEvent() right after (hardware).
+    -- Anywhere else breaks preventStandby/allowStandby invariants used by background jobs while UI is left running.
+    self:_standbyTransition()
+
     -- wait for next event
     local input_event = Input:waitEvent(wait_us)
 
@@ -1290,6 +1296,35 @@ function UIManager:resume()
     elseif Device:isKindle() then
         self.event_handlers["OutOfSS"]()
     end
+end
+
+-- Release standby lock once. We're done with whatever we were doing in the background.
+-- Standby is re-enabled only after all issued prevents are paired with allowStandby for each one.
+function UIManager:allowStandby()
+    assert(self._prevent_standby_count > 0, "allowing standby that isn't prevented; you have an allow/prevent mismatch somewhere")
+    self._prevent_standby_count = self._prevent_standby_count - 1
+end
+
+-- Prevent standby, ie something is happening in background, yet UI may tick.
+function UIManager:preventStandby()
+    self._prevent_standby_count = self._prevent_standby_count + 1
+end
+
+-- Allow/prevent calls above can interminently allow standbys, but we're not interested until
+-- the state change crosses UI tick boundary, which is what self._prev_prevent_standby_count is tracking.
+function UIManager:_standbyTransition()
+    if self._prevent_standby_count == 0 and self._prev_prevent_standby_count > 0 then
+        -- edge prevent->allow
+        logger.dbg("allow standby")
+        Device:setAutoStandby(true)
+        self:broadcastEvent(Event:new("AllowStandby"))
+    elseif self._prevent_standby_count > 0 and self._prev_prevent_standby_count == 0 then
+        -- edge allow->prevent
+        logger.dbg("prevent standby")
+        Device:setAutoStandby(false)
+        self:broadcastEvent(Event:new("PreventStandby"))
+    end
+    self._prev_prevent_standby_count = self._prevent_standby_count
 end
 
 function UIManager:flushSettings()
--- a/plugins/coverbrowser.koplugin/bookinfomanager.lua
+++ b/plugins/coverbrowser.koplugin/bookinfomanager.lua
@@ -544,6 +544,8 @@ function BookInfoManager:collectSubprocesses()
             local pid = self.subprocesses_pids[i]
             if util.isSubProcessDone(pid) then
                 table.remove(self.subprocesses_pids, i)
+                -- Prevent has been issued for each bg task spawn, we must allow for each death too.
+                UIManager:allowStandby()
             else
                 i = i + 1
             end
@@ -621,6 +623,9 @@ function BookInfoManager:extractInBackground(files)
         logger.warn("Failed lauching background extraction sub-process (fork failed)")
         return false -- let caller know it failed
     end
+    -- No straight control flow exists for background task completion here, so we bump prevent
+    -- counter on each task, and undo that inside collectSubprocesses() zombie reaper.
+    UIManager:preventStandby()
     table.insert(self.subprocesses_pids, task_pid)
     self.subprocesses_last_added_ts = util.gettime()
 
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -300,6 +300,11 @@ function Device:info()
     return self.model
 end
 
+-- Hardware specific method for UI to signal allowed/disallowed standby.
+-- The device is allowed to enter standby only from within waitForEvents,
+-- and only if allowed state is true at the time of waitForEvents() invocation.
+function Device:setAutoStandby(isAllowed) end
+
 -- Hardware specific method to handle usb plug in event
 function Device:usbPlugIn() end
 
--- a/plugins/backgroundrunner.koplugin/commandrunner.lua
+++ b/plugins/backgroundrunner.koplugin/commandrunner.lua
@@ -1,4 +1,5 @@
 local logger = require("logger")
+local UIManager = require("ui/uimanager")
 
 local CommandRunner = {
     pio = nil,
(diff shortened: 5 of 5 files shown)


plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -621,6 +622,7 @@ function BookInfoManager:extractInBackground(files)
         logger.warn("Failed lauching background extraction sub-process (fork failed)")
         return false -- let caller know it failed
     end
+    UIManager:preventStandby()
```
poire-z, line 625: I guess you put it there to not have to put allowStandby above in if not task_pid.
Dunno if the self:terminateBackgroundJobs() above should also be inside preventStandby - it feels like it should, but may be you're fine with small delay, like here just after runInSubProcess is launched.

plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -544,6 +544,7 @@ function BookInfoManager:collectSubprocesses()
             local pid = self.subprocesses_pids[i]
             if util.isSubProcessDone(pid) then
                 table.remove(self.subprocesses_pids, i)
+                UIManager:allowStandby()
```
poire-z, line 547: We may have multiple pids to collect (`while i <= #self.subprocesses_pids` - I dont remember if this really happens, but the code allows for it), so allowing standby after only one collected might be wrong. Best to do it between original lines 575/576.

frontend/ui/trapper.lua
```diff
@@ -126,6 +126,7 @@ function Trapper:info(text, fast_refresh)
         return true -- not dismissed
     end
 
+    UIManager:preventStandby()
```
poire-z, line 129: Not so sure it's to be done here, in Trapper:info().
It's been a while since I wrote this, so I'm not so sure anymore :) but I think Trapper:info() runs for 100ms, and then give control back to the calling code - and it's in that calling code that some possibly long running blocking stuff is happening (like when downloading an EPUB with 100 images, you'll get 100 Trapper:info("Downloading image 4/100...") - but stuff is done in wikipedia.lua.
So, I guess this shouldn't be done in Trapper:info() and Trapper:confirm() - but in the few bits of code that calls them.

But it's fine in the other dismissable... below.

plugins/coverbrowser.koplugin/bookinfomanager.lua
ezdiy: The logic here is: create pid, ban one standby. reap pid, allow one standby. These always come in pairs, nothing else matters. This is necessary because stuck/previous pid killing is ad-hoc with no real top-down flow.

plugins/coverbrowser.koplugin/bookinfomanager.lua
ezdiy: This is the opposite part of the "1 pid = 1 ban" logic. The allow is called for each pid that terminated either on its own, or got explicitly nuked by terminateBackgroundJobs() at some essentially *random* point earlier.

plugins/coverbrowser.koplugin/bookinfomanager.lua
poire-z: OK, it makes sense, I got fooled by the name "allowStandby" :) which doesn't convey the fact that it's some counter that we're incrementing. Might be worth adding a small comment on each of them in this "track pid" cases:
`-- increase/decrease standby preventers count by 1`.

(It feels a bit heavy to track each individual pid, but I guess you have to - even if there's a single place where we're sure all pids are done, the other place where we create pids might get called when there are already old pids created and a standby already prevented. So, ok.)

frontend/ui/trapper.lua
ezdiy: > Trapper:info() runs for 100m

Yup, whenever it yields to ui loop is the moment of UI getting stuck in a standby.

> and it's in that calling code that some possibly long running blocking stuff is happening

All of that long running blocking stuff is wrapped in trapper calls in the first place. Stuff like `ffi.sleep` or io.popen doesn't concern us, they can never invoke ui loop that does the actual standby.

With trapper, we don't care about whatever it is happening in background, we care about telling ui loop to not standby whenever we yield control to it, which is why it's fine to do it for each info(). While this indeed creates needless standby bans, it vastly simplifies the code compared to micromanaging this for each trapper end user individually.

The only time you want to micromanage is when the place runs outside of trapper - polling manually via its own `UIManager:scheduleIn()`. That is certainly the case for covers, bgrunner and as well as whole slew of network stuff (calibre, kosync, evernote..). We don't care about the network ones for the time being, as that would mandate a bit more invasive changes down to luasocket to make it robust enough. For now, it's only safe to autostandby with no network, and in the hands of standby driver to ensure that.

poire-z: > All of that long running blocking stuff is wrapped in trapper calls in the first place. Stuff like ffi.sleep or io.popen doesn't concern us, they can never invoke ui loop that does the actual standby.

But being "wrapped in trapper calls" changes what? It's just being in a subroutine (still in the main and single thread) and being able to yield() to go to UI loop for some small period of time.

If the code does trapper.info("Processing...") - which preventStandby, waits 100ms, allosStandby and give control back to the original code - which then does a computation that would take 20s, the PB system will notice no input for the last 5s and will do standby - while the user is still thinking the computation is happening.

Or may be you should preventStandby/allowStandby in Trapper:wrap() (inside pcalled_func) - so it's prevented for as long as the wrapped function is running ?

Or am I missing something ? (it's been a long time - it's impressive how you got that quickly a hold of our code base :)

ezdiy: > It's just being in a subroutine (still in the main and single thread) and being able to yield() to go to UI loop for some small period of time.

There's no harm in, however, freely lifting the ban in the context of the coroutine interminently as long there are no yields inbetween (this happens at the boundary between two info()) - you must ban only before you yield.

You can of course do this from the outside, but effect would be the same, while putting significant burden on all trapper clients.

>Or may be you should preventStandby/allowStandby in Trapper:wrap() (inside pcalled_func) - so it's prevented for as long as the wrapped function is running ?

Yes, you mean:
```
function Trapper:wrap(func)
    -- Catch and log any error happening in func (an error happening
    -- in a coroutine just aborts silently the coroutine)
    local pcalled_func = function()
        -- we use xpcall as it can give a whole stacktrace, unlike pcall
        local ok, err = xpcall(function()
            UIManager:preventStandby()
            func()
            UIManager:allowStandby()
        end, debug.traceback)
        if not ok then
            logger.warn("error in wrapped function:", err)
            return false
        end
        return true
        -- As a coroutine, we will return at first coroutine.yield(),
        -- and the above true/false won't probably be caught by
        -- any code, but let's do it anyway.
    end
    local co = coroutine.create(pcalled_func)
    return coroutine.resume(co)
end
```

This is what I've started with, the problem is error(). I'd need to wrap all coroutine.resume() in callbacks with error checks to unwind the ban, making an elegant solution rather messy in practice.

The current approach counts on that we have a good control over error in limited section before each yield, and in wherever else the standby ban is un-bumped so no harm done on error(). Whereas holding the ban for entire duration of trapper that is series of yields, a plugin can mess up, crash the coroutine, and subsequently leave the counter stuck.

poire-z: > I guess this is the source of misunderstanding

Yes, I might not understand how PB standby happens.

> The invariant everywhere is that standby can't just kick in willy nilly even if unbanned. Only a routine called by next UI loop trip (input event hook/driver, looper, timer..) can actually freeze the system

Not sure I understand. Why can't it just kick in?
And which is that routine ? Isn't it the PB system that will just decide to go into standby without knowing KOReader working state ? Or do/will you have to, in our UI loop, do the inkview.allowStanby / sleep+waitforinut / disallowStandby ? May be it's this that I don't get.

> Yes, you mean:

Or rather if you don't want to miss allowStandby in case of error:

```lua
function Trapper:wrap(func)
    -- Catch and log any error happening in func (an error happening
    -- in a coroutine just aborts silently the coroutine)
    local pcalled_func = function()
        UIManager:preventStandby()
        -- we use xpcall as it can give a whole stacktrace, unlike pcall
        local ok, err = xpcall(function()
            func()
        end, debug.traceback)
        UIManager:allowStandby()
        if not ok then
            logger.warn("error in wrapped function:", err)
            return false
        end
        return true
        -- As a coroutine, we will return at first coroutine.yield(),
        -- and the above true/false won't probably be caught by
        -- any code, but let's do it anyway.
    end
    local co = coroutine.create(pcalled_func)
    return coroutine.resume(co)
end
```

ezdiy: >waitforinut
In case of pocketbook, it's waitforinput that kicks in standby if it's unbanned at that point. In case of kobo, you'd do the same from your scheduled timer. Effect is the same - both are run from ui loop, *outside* of context of any trapper coroutine - you have to yield into such code path.

```
        if not ok then
            logger.warn("error in wrapped function:", err)
            return false
        end
```

This does nothing on error() past first yield(). I'm not opposed to this approach, *provided* coroutine.resume() in callbacks actually does catch the arrors in some central place. This amounts to more changes than I'm comfortable with.

poire-z: > In case of pocketbook, indeed it's waitforinput that kicks in standby if it's unbanned 

OK, if that's true, I think I'm fine with your current approach :) 
Can you point at which part of the pocketbook code this is allowed ? I don't see anything explicitely mentionning that in https://github.com/koreader/koreader-base/pull/1168.
May be add some comments before/after the explicite inkview call (I guess), Could it be this?

```lua
inkview.ProcessEventLoop() -- standby is allowed when in this
-- standby won't happen after this
```

> This does nothing on error() past first yield(). 

With what did you test it ? The wikipedia Save as EPUB has its own handler for error that will prevent the main Trapper:wrap one to kick in. If I remove it and have the main trapper one handle it, I get after a few Trapper:info():
```
08/25/20-16:04:25 WARN  error in wrapped function: frontend/ui/wikipedia.lua:1352: attempt to index global 'yyy' (a nil value)
stack traceback:
 frontend/ui/wikipedia.lua:1352: in function 'createEpub'
 frontend/ui/wikipedia.lua:1423: in function <frontend/ui/wikipedia.lua:1421>
 [C]: in function 'xpcall'
 frontend/ui/trapper.lua:47: in function <frontend/ui/trapper.lua:45>
```

poire-z: > Ok, I'll force-push this with xpcall global wrapper in a while

Don't feel obliged if it works with that UI loop wait for input granularity.
Might be better if the user falls asleep while on a `Trapper:confirm("Do you want to proceed?") `call :)

frontend/ui/uimanager.lua
```diff
@@ -80,12 +82,11 @@ function UIManager:init()
         end)
     end
     if Device:isPocketBook() then
+        -- Only fg/bg state plugin notifiers, not real power event.
         self.event_handlers["Suspend"] = function()
             self:_beforeSuspend()
-            Device:onPowerEvent("Power")
         end
         self.event_handlers["Resume"] = function()
-            Device:onPowerEvent("Power")
```
ezdiy, line 88: Note that this doesn't really belong this PR thematically, but I'm keeping it in track with changes to UIManager. See https://github.com/koreader/koreader/pull/6559#discussion_r476898817

poire-z (approved):
Looks sane to me.

Issue #6321, opened 2020-06-27
Koreader doesn't reuse existing running instances


xahon:
* KOReader version: 2020.06
* Device: Pocketbook 740 pro

#### Issue

When I open book, then go to home screen and open the same book again, then there are two tasks with the same name in task manager (both with title of my book). I can open lots of instances of koreader
Expected behaviour: When I open the same book koreader reuses existing instance of it. Like in iOs or android, when you open application that is still present in RAM it reopens again without extra instances

#### Steps to reproduce
1. Open book 'a'
2. Go home screen
3. Repeat 1 one or more times
4. Open up task manager

pazos:
Good catch!

> Expected behaviour: When I open the same book koreader reuses existing instance of it. Like in iOs or android, when you open application that is still present in RAM it reopens again without extra instances

That is exactly what I would expect in a mobile device (and what happens on android). On the desktop I think we're not aim at a single instance app, as having few instances is actually cool and useful (and not too resource hungry).

A proper single-instance implementation for pocketbook would mean:

- check that we can, in fact, repurpose the running instance. It is trivial on desktop but I have no idea how to do that on pocketbook (mechanism to raise a window?, some inkview api call?)
- on startup script check if other instances are running
- if an instance is already running forward messages to it using some IPC mechanism.

I would suggest a named pipe as a way to do chepo IPC and check if running instances exist.

Given that nobody has a pocketbook I would suggest a simpler solution: `killall`. Or, if you value current position, annotations, etc in the previous document opened, use a named pipe to tell it to close saving settings first. That should work without having to research how to raise the window and other things that are platform specific and noboby can answer without the device in hand.

pazos:
Can't reproduce. Maybe is something of old FW or specific to a device?

roshavagarga:
@xahon Are you running firmware 5.20 or firmware 6 that released in May 2020?

tastytea:
I can reproduce the issue with a PB632 (PocketBook Touch HD 3) running FW U632.6.0.926, released 2020-04-06 and KOReader 2020.08.

pazos:
@tastytea: thank you. Just to see if I understand correctly:

You can reproduce it following:

```
Open book 'a' (from PB library)
Go home screen
Repeat 1 one or more times
Open up task manager
```

But is not reproducible by starting the KOReader icon under applications n times. Is that correct?. If so please attach the extensions.cfg used to open books from PB system. Thanks!

tastytea:
[...]

PR #6559, opened 2020-08-25
Pocketbook new poll frontend

ezdiy:
Needs https://github.com/koreader/koreader-base/pull/1168 and #6558
RFC #6545
    
This one is slightly stripped down compared to the gist.
Out of the box, it behaves exactly as native pocketbook apps should, ie aggressive standby, but no freezing at operations in progress.
    
Config UI will be done via koplugin that will also do adaptive standby. This is because the API is now device independent (albeit PB is the only implemented user).

This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6559)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -1,55 +1,17 @@
 local Generic = require("device/generic/device") -- <= look at this file!
 local logger = require("logger")
 local ffi = require("ffi")
+local C = ffi.C
 local inkview = ffi.load("inkview")
+local band = require("bit").band
 
--- luacheck: push
--- luacheck: ignore
-local EVT_INIT = 21
-local EVT_EXIT = 22
-local EVT_SHOW = 23
-local EVT_REPAINT = 23
-local EVT_HIDE = 24
-local EVT_KEYDOWN = 25
-local EVT_KEYPRESS = 25
-local EVT_KEYUP = 26
-local EVT_KEYRELEASE = 26
-local EVT_KEYREPEAT = 28
-local EVT_FOREGROUND = 151
-local EVT_BACKGROUND = 152
+require("ffi/posix_h")
+require("ffi/linux_input_h")
+require("ffi/inkview_h")
 
-local KEY_POWER  = 0x01
-local KEY_DELETE = 0x08
-local KEY_OK     = 0x0a
-local KEY_UP     = 0x11
-local KEY_DOWN   = 0x12
-local KEY_LEFT   = 0x13
-local KEY_RIGHT  = 0x14
-local KEY_MINUS  = 0x15
-local KEY_PLUS   = 0x16
-local KEY_MENU   = 0x17
-local KEY_PREV   = 0x18
-local KEY_NEXT   = 0x19
-local KEY_HOME   = 0x1a
-local KEY_BACK   = 0x1b
-local KEY_PREV2  = 0x1c
-local KEY_NEXT2  = 0x1d
-local KEY_COVEROPEN = 0x02
-local KEY_COVERCLOSE = 0x03
-
-local CONNECTING = 1
-local CONNECTED = 2
-local NET_OK = 0
--- luacheck: pop
-
-ffi.cdef[[
-char *GetSoftwareVersion(void);
-char *GetDeviceModel(void);
-int GetNetState(void);
-int GetFrontlightColor(void);
-int NetConnect(const char *name);
-int NetDisconnect();
-]]
+-- FIXME: Signal ffi/input.lua (brought in by device/input later on) that we want to use poll mode backend.
+-- Remove this once backend becomes poll-only.
+_G.POCKETBOOK_FFI = true
 
 local function yes() return true end
 local function no() return false end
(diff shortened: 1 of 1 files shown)


frontend/device/pocketbook/device.lua
```diff
@@ -112,14 +119,11 @@ function PocketBook:init()
...
-                return "Suspend"
+            if ev.code == EVT_HIDE or ev.code == EVT_BACKGROUND then
+                return "SaveState"
             elseif ev.code == EVT_FOREGROUND then
-                if self.isInBackGround then
-                    self.isInBackGround = false
-                    return "Resume"
-                end
+                require("ui/uimanager"):setDirty('all', 'partial')
+                return "Foreground"
```
pazos, line 126: The foreground event handler doesn't exist. It is intended?

pazos:
I think the first commit belongs to #6558. The rest looks fine.

frontend/device/pocketbook/device.lua
ezdiy: Yes, the intent is to translate it to something "human readable", and none of the existing frontent->uimanager events really fit. I'm not fond of spoofing 'Resume' and such when it's not - it's been source of trouble before. Better just name it properly, and let ui care (it bubbles down to `UIManager:sendEvent()` if not handled explicitly in manager/plugin).

frontend/device/pocketbook/device.lua
```diff
@@ -79,9 +87,8 @@ function PocketBook:blacklistCBB()
     local C = ffi.C
 
     -- As well as on those than can't do HW inversion, as otherwise NightMode would be ineffective.
-    --- @fixme Either relax the HWInvert check, or actually enable HWInvert on PB if it's safe and it works,
-    --        as, currently, no PB device is marked as canHWInvert, so, the C BB is essentially *always* blacklisted.
-    if not self:canUseCBB() or not self:canHWInvert() then
+    -- Do this only if settings override isn't present.
+    if (not G_reader_settings:has("dev_no_c_blitter")) and (not self:canUseCBB() or not self:canHWInvert()) then
```
NiLuJe, line 91: The `G_reader_settings` bit is not necessary, `reader.lua` already sets the env var checked below ;).

(This is weirdly convoluted for... reasons I don't even remember :D).

frontend/device/pocketbook/device.lua
pazos: Ok, that makes sense.

Probably the only bit we might want from `onResume` is to update the footer to show the real battery percent

ezdiy: Now looking at the plugins, the situation is rather unfortunate that those all treat Resume/Suspend as equivalent of foreground/background (launch/stop timers to update stuff etc). Guess there's no avoiding spoofing the event, instead one has to be careful to relegate its function only to "notify plugins", and don't *don't* hang actual Resume/Suspend state modifiers as the event may not come from these.

Blah.

frontend/device/pocketbook/device.lua
ezdiy: Ok, brought it back as it was.

frontend/device/pocketbook/device.lua
```diff
@@ -58,26 +20,23 @@ local function no() return false end
...
     model = "PocketBook",
     isPocketBook = yes,
-    isInBackGround = false,
     hasOTAUpdates = yes,
     hasWifiToggle = yes,
     isTouchDevice = yes,
     hasKeys = yes,
     hasFrontlight = yes,
     canSuspend = no,
-    emu_events_dev = "/dev/shm/emu_events",
```
pazos, line 68: @ezdiy: I see a point in keeping that fifo. It turns out that the Pocketbook uses a new instance each time a book is opened from the PB Library.  https://github.com/koreader/koreader/issues/6321. We can repurpose the same instance feeding the new file as ev.value with our own ev.code, for example:

Instead of having the same shell script duplicated in `/mnt/ext1/applications/koreader.app` and `/mnt/ext1/system/bin/koreader.app`, replace the later with

```bash
#!/bin/sh

# IPC for pocketbook. Used to redirect commands to a running instance.
# Events generated with this script need to be handled by handleMiscEv.

# $1: full path of the file to open

KO_EVENT_PIPE=/dev/shm/emu_events
KO_SCRIPT=/mnt/ext1/applications/koreader.app
KO_DIR=/mnt/ext1/applications/koreader

cd "${KO_DIR}" || exit

NewMiscEv() {
    ./luajit <<EOF
    require("setupkoenv")
    local dump = require("dump")
    local ffi = require("ffi")
    local ev_time = ffi.new("timeval")
    ffi.C.gettimeofday(ev_time, nil)
    local ev = {
        type = 4,
        code = 999,
        value = "${1}",
        time = { sec = tonumber(ev_time.tv_sec), usec = tonumber(ev_time.tv_usec) }
    }
    local pipe = io.open(KO_EVENT_PIPE, "wb")
    if pipe then
        pipe:write(dump(ev))
        pipe:close()
    end
EOF
}


if [ -p "${KO_EVENT_PIPE}" ]; then
    NewMiscEv "${@}"
else
    "${KO_SCRIPT}" "${@}"
fi
```

Probably there are better ways of doing the same and I'm fine with all of them :D, including leave this bug unfixed, but this is just FYI.

ezdiy: @pazos Noted. Definitely wontfix in this PR as there's no sensible way to revive the pipe just to inject events. Scope of future ones though - https://github.com/koreader/koreader/issues/6597

pazos (approved):
LGTM!!!

@ezdiy: shall I merge #6558 and wait for a rebase in this pull or merge this directly and omit #6558 ?

ezdiy:
@pazos Try merging this tip directly, I'm quite curious if github is smart enough to recognize the dependency as merged (if not, can just close the second one...).

NiLuJe:
Since we generally squash and don't rebase, I'm not *sure* it'll grok that all on its own ;).

ezdiy:
@NiLuJe If I squash this, that would detach #6558. If you need to ensure 1-PR, 1-commit then definitely merge #6558 first, and I rebase this one on resulting master.

NiLuJe:
Yep, my point exactly ;).

NiLuJe:
Done :).

(Don't forget to bump base in this one, you'll pickup https://github.com/koreader/koreader-base/pull/1168, https://github.com/koreader/koreader-base/pull/1175 & https://github.com/koreader/koreader-base/pull/1176)

ptrm:
The only reason I did not even consider having a PocketBoo,  after owning one for w short while, is now gone. I'm both really grateful and worried about my wallet 😥

PR #6566, opened 2020-08-25
Allow i18n in KoboUSBMS

NiLuJe:
(... at least in theory ;p)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6566)

### DIFF
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -45,10 +45,7 @@ if Device:canToggleMassStorage() then
     }
 
     -- mass storage actions
-    common_settings.mass_storage_actions = {
-        text = _("Start USB storage"),
-        callback = function() MassStorage:start(true) end,
-    }
+    common_settings.mass_storage_actions = MassStorage:getActionsMenuTable()
 end
 
 -- This affects the topmenu, we want to be able to access it even if !Device:setDateTime()
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -47,7 +47,7 @@ if Device:canToggleMassStorage() then
     -- mass storage actions
     common_settings.mass_storage_actions = {
         text = _("Start USB storage"),
-        callback = function() MassStorage:start() end,
+        callback = function() MassStorage:start(true) end,
     }
 end
 
--- a/frontend/ui/elements/mass_storage.lua
+++ b/frontend/ui/elements/mass_storage.lua
@@ -13,7 +13,7 @@ function MassStorage:getSettingsMenuTable()
     return {
         {
             text = _("Disable confirmation popup"),
-            help_text = _([[This will NOT affect what happens when you simply plug in the device!]]),
+            help_text = _([[This will ONLY affect what happens when you simply plug in the device!]]),
             checked_func = function() return not self:requireConfirmation() end,
             callback = function()
                 G_reader_settings:saveSetting("mass_storage_confirmation_disabled", self:requireConfirmation())
@@ -35,8 +35,8 @@ function MassStorage:getActionsMenuTable()
 end
 
 -- exit KOReader and start mass storage mode.
-function MassStorage:start(always_ask)
-    if self:requireConfirmation() or always_ask then
+function MassStorage:start(never_ask)
+    if not never_ask and self:requireConfirmation() then
         local ConfirmBox = require("ui/widget/confirmbox")
         UIManager:show(ConfirmBox:new{
             text = _("Share storage via USB?"),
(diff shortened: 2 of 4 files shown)


NiLuJe:
Pending actual support in the version we use ;p.

@Frenzie: How would this work, actually? Because I'd need the .po at build time, and then compile them properly so they end up in the tarball...

On the upside, the in-repo test with a French translation works, so, there's that.

Frenzie:
You mean including it in the big file might be problematic? In that case I guess we'd have to go with the new module option.

NiLuJe:
Yeah, it's a dedicated domain.

In practice, that means moving the repo to the org, a new weblate component, and either weblate pushing to the repo or a new l10n subrepo or something?

And then a crapload of Makefile magic to hide the mess?

Can Weblate ship compiled .mo files for us, or am I hoping for too much? ^^.

Frenzie:
> In practice, that means moving the repo to the org, a new weblate component, and either weblate pushing to the repo or a new l10n subrepo or something?

Pretty much, yes. subrepo vs just the repo depends on if you're okay with some noise.

There's an addon to generate `.mo` files.

NiLuJe:
@Frenzie: Okay, transfer is done ;).

in-repo is fine by me for such a small tool. And yes please for the .mo add-on ;).

oldman63:
> Yeah, it's [a dedicated domain](https://github.com/NiLuJe/KoboUSBMS/compare/v0.9.1...master).
> 
> In practice, that means moving the repo to the org, a new weblate component, and either weblate pushing to the repo or a new l10n subrepo or something?

The Hungarian translation is ready ...
Where can I upload the po file?

Frenzie:
@oldman63 You should be able to use "upload file" in https://github.com/koreader/KoboUSBMS to create a PR.

Frenzie:
@NiLuJe I dropped them an e-mail. Depending on how busy they are + covid and whatnot I don't know how long it'll take since we're not paying for anything, and of course paying customers come first. ;-)

NiLuJe:
@Frenzie: Cool, thanks ;).

Frenzie:
@NiLuJe I just noticed you pushed exit to page 2. This'll need a bit more thought. ;-)

NiLuJe:
@Frenzie: Ah, didn't notice, it's still on page 1 w/ my set of plugins...

NiLuJe:
Err, actually, there are no plugins in that list, AFAICT?

o_O

Frenzie:
System stats is a plugin.

Frenzie:
But also I was looking in the reader, I think it has more menu items. ;-)

NiLuJe:
Ah, okay, good point, I hadn't checked the Reader ;).

(I do have system stats enabled ;)).

NiLuJe:
I admit that if the UI code wasn't already there, my first reflex would have been to put it inside the Exit > menu (because it's technically an exit state).

Then again, I kind of like it as a top-level option.

But I don't really see anything that could be moved out of the way in the Reader...

https://github.com/koreader/koreader/blob/06bce655c527735d1b331b71889e79124dfb31b4/frontend/ui/elements/reader_menu_order.lua#L160-L177

NiLuJe:
Or, hmm, maybe moving Help after Exit?

poire-z:
> maybe moving Help after Exit?

Not very standard, Exit is the last mostly everywhere I guess.
You could move system_statistics elsewhere, or in a submenu with other stuff.
I wouldn't mind ota_update, help and version in a submenu (as that's not something one would use when reading) but how to name it?

Frenzie:
> I admit that if the UI code wasn't already there, my first reflex would have been to put it inside the Exit > menu (because it's technically an exit state).

The exit menu itself is already questionable for new users. :-P

NiLuJe:
> 
> 
> > maybe moving Help after Exit?
> 
> Not very standard, Exit is the last mostly everywhere I guess.
> You could move system_statistics elsewhere, or in a submenu with other stuff.
> I wouldn't mind ota_update, help and version in a submenu (as that's not something one would use when reading) but how to name it?

I like both these ideas ;p.

Say, punt off device statistics to the fancy new "I don't know where to put this" 'More tools' menu, and then at *least* move Version inside the Update menu?

poire-z:
> The exit menu itself is already questionable for new users. :-P

Yes... (in my own patch, I moved it into Help>, and just have Restart KOReader and Exit)

> Say, punt off device statistics to the fancy new "I don't know where to put this" 'More tools' menu, and then at least move Version inside the Update menu?

Fine with me. (System statistics could go to Hellp> :)
For what it's worth, I also have:
```diff
 common_info.version = {
-    text = _("Version"),
+    text = T(_("Version: %1"), version),
```
(which I think might be truncated with some DPI settings, but if it does not appear in the main first panel, it could be fine)

Frenzie:
> Fine with me. (System statistics could go to Hellp> :)

Yup, makes sense.

Frenzie:
> Yes... (in my own patch, I moved it into Help>, and just have Restart KOReader and Exit)

Something like that might not be a bad idea either. I kinda of like what Windows does, although I don't know how that tested with focus groups. (Well enough, one imagines). You have the big exit/shutdown button and if you click on the right side of it you get the options like sleep/suspend and restart.

PR #6567, opened 2020-08-25
[Pocketbook]: enable usage of system fonts

pazos:
They're not encrypted, so we can use them.

By default "use system fonts" is disabled, like in other platforms.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6567)

### DIFF
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -353,7 +353,7 @@ end
 function ReaderFont:getFontSettingsTable()
     local settings_table = {}
 
-    if Device:isAndroid() or Device:isDesktop() or Device:isEmulator() then
+    if Device:isAndroid() or Device:isDesktop() or Device:isEmulator() or Device:isPocketBook() then
         for _, item in ipairs(require("ui/elements/font_settings"):getSystemFontMenuItems()) do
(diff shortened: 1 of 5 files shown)


frontend/apps/reader/modules/readerfont.lua
```diff
@@ -353,7 +353,7 @@ end
 function ReaderFont:getFontSettingsTable()
     local settings_table = {}
 
-    if Device:isAndroid() or Device:isDesktop() or Device:isEmulator() then
+    if Device:isAndroid() or Device:isDesktop() or Device:isEmulator() or Device:isPocketBook() then
```
Frenzie, line 356: I'd say it's starting to look like it's time for some kind of Device:hasSystemFonts() abstraction but I suppose it's just this one file?

Frenzie (approved):

frontend/apps/reader/modules/readerfont.lua
pazos: Exactly. But the same happens with Device:hasExitOptions, which is only used on reader and FM menu.

Frenzie: Either way's fine by me for now. ;-)

PR #6568, opened 2020-08-25
android: fix frontlight dialog gesture

pazos:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6568)

### DIFF
--- a/frontend/device/devicelistener.lua
+++ b/frontend/device/devicelistener.lua
@@ -227,8 +227,7 @@ if Device:hasFrontlight() then
     end
 
     function DeviceListener:onShowFlDialog()
-        local FrontLightWidget = require("ui/widget/frontlightwidget")
-        UIManager:show(FrontLightWidget:new{})
+        Device:showLightDialog()
     end
 
 end
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -296,6 +296,12 @@ function Device:onPowerEvent(ev)
     end
 end
 
+function Device:showLightDialog()
+    local FrontLightWidget = require("ui/widget/frontlightwidget")
+    local UIManager = require("ui/uimanager")
+    UIManager:show(FrontLightWidget:new{})
+end
+
 function Device:info()
     return self.model
 end


zwim:
LGTM. Works on Epos, Phone and Emulator. :+1:

zwim:
@pazos: Bad news on Tolino. On the phone it there is no issue.

Native Frontlight Dialog on Tolino:
* When called from the menu it works always. Sometimes there is a noticeable short lag (maybe 1/10s) before the sliders are moved.
* When the native dialog is called by gesture it freezes most of the times (not always). Then I get an ANR. 
[logcat.frontlight.docx](https://github.com/koreader/koreader/files/5128912/logcat.frontlight.docx)

zwim:
@pazos. The PR works on Android and on Tolino.
Short press and menu OK.
But when the dialog is called by a long press, I get an ANR on Android and Tolino. (Most of the times)

pazos:
@zwim: it looks like the event queue still holds a non consumed event when the dialog is shown. Events that are not consumed in 5 seconds will trigger an ANR.

Sadly there's no "simple" way of doing the right thing there. Probably the easier workaround is something like we do on the backend example: https://github.com/koreader/android-luajit-launcher/blob/master/examples/helloWorld/main_loop.lua#L28-L31

That would mean finish all events before processing them. I think it is safe for all touch events since they are never relayed. So my suggestion would be:

for key events leave the code as is (consume events after adding them to the queue) and for touch events always consume them before it. 

See https://github.com/koreader/koreader-base/blob/master/ffi/input_android.lua#L153-L162

Something like this could work:

```lua
if android.lib.AInputQueue_preDispatchEvent(android.app.inputQueue, event[0]) == 0 then
    local event_type = android.lib.AInputEvent_getType(event[0])
    if event_type == C.AINPUT_EVENT_TYPE_MOTION then
        -- always consume event first
        android.lib.AInputQueue_finishEvent(android.app.inputQueue, event[0], 1)
        motionEventHandler(event[0])
    elseif event_type == C.AINPUT_EVENT_TYPE_KEY then
        local handled = keyEventHandler(event[0])
        android.lib.AInputQueue_finishEvent(android.app.inputQueue, event[0], handled)
    end
end
```

zwim:
@pazos: I did it like you have proposed. It did not help. 
I have made a logcat for hold corner with nativeDialog (left bottom corner) and TableOfContents (right bottom corner)
[logcat-hold-corner-native-dialog.docx](https://github.com/koreader/koreader/files/5140657/logcat-hold-corner-native-dialog.docx)
[logcat-hold-corner-widget.docx](https://github.com/koreader/koreader/files/5140658/logcat-hold-corner-widget.docx)

The difference is, that when the widget is called, events are processed.
But where are the events processed in the native dialog. We have to consume the events in nativeDialog (before the usleep).
I have an almost working solution (no ANRs, but some gestures are swallowed). I will do a pull request on that, today.
Thanks.

zwim:
@pazos: #6583 fixes the ANRs. Let's discuss there.

PR #6583, opened 2020-08-28
resolve ANR on hold and native light dialog

zwim:
tested on Epos2 and phone.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6583)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -362,13 +362,41 @@ function Device:canExecuteScript(file)
     end
 end
 
+--swallow all events
+local function processEvents()
+    local events = ffi.new("int[1]")
+    local source = ffi.new("struct android_poll_source*[1]")
+    local poll_state = android.lib.ALooper_pollAll(-1, nil, events, ffi.cast("void**", source))
+    if poll_state >= 0 then
+        if source[0] ~= nil then
+            if source[0].id == C.LOOPER_ID_MAIN then
+                local cmd = C.android_app_read_cmd(android.app)
+                C.android_app_pre_exec_cmd(android.app, cmd)
+                C.android_app_post_exec_cmd(android.app, cmd)
+            elseif source[0].id == C.LOOPER_ID_INPUT then
+                local event = ffi.new("AInputEvent*[1]")
+                while android.lib.AInputQueue_getEvent(android.app.inputQueue, event) >= 0 do
+                    if android.lib.AInputQueue_preDispatchEvent(android.app.inputQueue, event[0]) == 0 then
+                        android.lib.AInputQueue_finishEvent(android.app.inputQueue, event[0], 1)
+                    end
+                end
+            end
+        end
+    end
+end
+
 function Device:showLightDialog()
     local usleep = require("ffi/util").usleep
     local title = android.isEink() and _("Frontlight settings") or _("Light settings")
     android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("OK"), _("Cancel"))
     repeat
+        processEvents() -- swallow all events, including the last one
         usleep(25000) -- sleep 25ms before next check if dialog was quit
     until (android.lights.dialogState() ~= C.ALIGHTS_DIALOG_OPENED)
+
+    local GestureDetector = require("device/gesturedetector")
+    GestureDetector:clearStates()
+
     local action = android.lights.dialogState()
     if action == C.ALIGHTS_DIALOG_OK then
         self.powerd.fl_intensity = self.powerd:frontlightIntensityHW()


frontend/device/android/device.lua
```diff
@@ -362,13 +362,41 @@ function Device:canExecuteScript(file)
     end
 end
 
+--swallow all events
+local function processEvents()
```
Frenzie, line 366: Probably not an objection, but this feels like something that might be cleaner as part of luajit-launcher? I'll leave it to @pazos.

pazos: Looks like a duplicate of https://github.com/koreader/koreader-base/blob/master/ffi/input_android.lua#L139-L165 but without generating linux input events.

I'm sure there're different ways of doing this but not sure which is the best one.

frontend/device/android/device.lua
```diff
@@ -362,13 +362,41 @@ function Device:canExecuteScript(file)
...
+        end
+    end
+end
+
 function Device:showLightDialog()
     local usleep = require("ffi/util").usleep
     local title = android.isEink() and _("Frontlight settings") or _("Light settings")
     android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("OK"), _("Cancel"))
     repeat
+        processEvents() -- swallow all events, including the last one
```
pazos, line 393: in a loop? or just once before `android.lights.showDialog` ?

frontend/device/android/device.lua
Frenzie: Oh right, I forgot about that one. That also feels cleaner to me. The effect is the same etc. but it feels more logical to me.

frontend/device/android/device.lua
zwim: >in a loop?

I am afraid yes.
Doing the processEvents() before or after showDialog and doing the processEvents() after a usleep(50000) **not** in the loop does not help. Instead I am getting different events (like single taps).
If you are concerned about the load we could lengthen the usleep to 75000.

pazos: > If you are concerned about the load we could lengthen the usleep to 75000.

Nope. I'm trying to figure out what happens. Could you attach a logcat of a debug build (it needs to be debug so it logs changes in activity state).

IIRC the dialog stays on top an catches all input events while it is running, so they're not forwarded to native code. Maybe that's not true anymore with the native surface implementation. Anyways using https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L271-L273 should get rid of noisy input events and shouldn't affect dialog controls.

zwim: @pazos: The `setIgnoreInput` was the first trick I tired, before all the event things. Nope, did not change anything.

Sorry, the compilation of the debug apk, took long. (It's time to parallelize the make.)

zwim:
These are the logcats. `logcat -c`; `logcat > logcat1.txt`; start koreader -> long press on edge -> open of the dialog -> change brightness, warmth -> close dialog. Did it two times.

[logcat1.txt](https://github.com/koreader/koreader/files/5143672/logcat1.txt)
[logcat2.txt](https://github.com/koreader/koreader/files/5143673/logcat2.txt)

frontend/device/android/device.lua
pazos: setIgnoreInput won't help to empty the event pool, but can prevent to fill it again with new events.

> These are the logcats. logcat -c; logcat > logcat1.txt; start koreader -> long press on edge -> open of the dialog -> change brightness, warmth -> close dialog. Did it two times.

Thanks for the logs!. As they don't show any crash I think you made them with the workaround. I'm sadly unable to reproduce the original issue on the emulator:

test

(FWIW I'm holding the down right corner to show the dialog)

zwim:
My bad. Here are the logs for the unpatched version.

On the phone I never had issues, just on both Tolinos (Epos2, Vision4).
My feelings are, that it is more possible to get the ANR if you touch really long (>2s). But it can happen even on a short hold touch.

[logcat3.txt](https://github.com/koreader/koreader/files/5143877/logcat3.txt)

pazos:
> My bad. Here are the logs for the unpatched version.

That makes sense:

```
V/KOReader( 7990):  in tap state...
V/KOReader( 7990):  hold gesture detected in slot 0
V/KOReader( 7990):  in hold state...
V/KOReader( 7990):  adjusted ges: hold
D/KOReader( 7990): [NativeGlue]: WindowFocusChanged: 0x4005de98 -- 0
I/InputDispatcher( 2501): Application is not responding: Window{4143ef90 u0 org.koreader.launcher/org.koreader.launcher.MainActivity}.  It has been 5006.5ms since event, 5005.7ms since wait started.  Reason: Waiting because the touched window has not finished processing the input events that were previously delivered to it.
I/WindowManager( 2501): Input event dispatching timed out sending to org.koreader.launcher/org.koreader.launcher.MainActivity.  Reason: Waiting because the touched window has not finished processing the input events that were previously delivered to it.
```

My guts say that once `WindowFocusChanged: 0x4005de98 -- 0` happens there no way to get an ANR unless there's still unfinished events on the pool. Also that happens after the native dialog is called, so it should be possible to use `processEvents` once in https://github.com/koreader/android-luajit-launcher/blob/master/assets/android.lua#L1045 if the light dialog is present.



> On the phone I never had issues, just on both Tolinos (Epos2, Vision4).
> My feelings are, that it is more possible to get the ANR if you touch really long (>2s). But it can happen even on a short hold touch.
> 

Sounds like a race condition. Normally the input queue is drained as fast as possible but that just happens when the window is in the foreground. A hold event is in fact a lof of touch events filling the pool, and while most of them can be drained before loosing focus some of them can stay there and trigger the ANR.

The fact that this doesn't happen on phones, even on old borked ones, is unfortunate :/

zwim:
>My guts say that once WindowFocusChanged: 0x4005de98 -- 0 happens there no way to get an ANR unless there's still unfinished events on the pool. Also that happens after the native dialog is called, so it should be possible to use processEvents once in https://github.com/koreader/android-luajit-launcher/blob/master/assets/android.lua#L1045 if the light dialog is present.

I have the same suspect as you. But as I mentioned earlier, I am not so firm on Android/Kotlin to do that. I would need your help.

>Sounds like a race condition. Normally the input queue is drained as fast as possible but that just happens when the window is in the foreground. A hold event is in fact a lof of touch events filling the pool, and while most of them can be drained before loosing focus some of them can stay there and trigger the ANR.

>The fact that this doesn't happen on phones, even on old borked ones, is unfortunate :/

Yeah. But the Tolinos are even more borked than the most borked phone. The have a upToDate(?) `1.000 MHz NXP i.MX6 SoloLite/ 512MB ` single core processor.
This separats the wheat from the chaff on apk-programming. ;-)

pazos:
> I have the same suspect as you. But as I mentioned earlier, I am not so firm on Android/Kotlin to do that. I would need your help.

Sure, at the end of [`handleMiscEv`](https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L129) add:

```lua
    elseif ev.code == C.APP_CMD_LOST_FOCUS then
        if android.lights.dialogState() == C.ALIGHTS_DIALOG_OPENED then
            logger.info("window lost focus caused by light dialog")
        end
    end
```

and see if the log statement happens between `WindowsFocusChanged 0` and the ANR. It should. Then replace the logger call with the processEvent that cleans all the inputs. If it works you don't need to include that function in a loop.

zwim:
Sorry again: With
```   
   elseif ev.code == C.APP_CMD_LOST_FOCUS then
        if android.lights.dialogState() == C.ALIGHTS_DIALOG_OPENED then
            logger.info("window lost focus caused by light dialog")
        end
    end
```
the logger.info doesn`t get called. Even with
```
   elseif ev.code == C.APP_CMD_LOST_FOCUS then
        logger.info("window lost focus caused by light dialog")
    end
```
the logger is also not called.

pazos:
@zwim: it is called when the app goes to the foreground again, so useless for us :/

```
08-29 16:33:44.199  3701  3714 W KOReader: Unable to lock window buffer
08-29 16:33:44.235  3701  3714 I KOReader:  window gained focus
08-29 16:33:48.805  3701  3714 I KOReader:  window lost focus
08-29 16:33:48.805  3701  3714 I KOReader:  window gained focus
08-29 16:33:53.232  3701  3714 I KOReader:  window lost focus
08-29 16:33:53.235  3701  3714 I KOReader:  window gained focus
08-29 16:33:56.154  3701  3714 I KOReader:  window lost focus
08-29 16:33:56.155  3701  3714 I KOReader:  window gained focus
08-29 16:33:59.293  3701  3714 I KOReader:  window lost focus
08-29 16:33:59.318  3701  3714 I KOReader:  window gained focus
08-29 16:34:07.013  3701  3714 I KOReader:  window lost focus
08-29 16:34:07.034  3701  3714 I KOReader:  window gained focus
08-29 16:34:12.900  3701  3714 I KOReader:  window lost focus
08-29 16:34:12.935  3701  3714 I KOReader:  window gained focus
08-29 16:34:16.037  3701  3714 I KOReader:  window lost focus
08-29 16:34:16.038  3701  3714 I KOReader:  window gained focus
08-29 16:34:22.957  3701  3714 I KOReader:  window lost focus
08-29 16:34:22.967  3701  3714 I KOReader:  window gained focus
```

frontend/device/android/device.lua
pazos: @Frenzie: I would say LGTM as is. In the future we can refactor as a base/input function.

I don't want to do it at the moment :(

Frenzie (approved):
lgtm, merge @pazos?

PR #6584, opened 2020-08-28
update statusline after native light dialog

zwim:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6584)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -373,14 +373,13 @@ function Device:showLightDialog()
     if action == C.ALIGHTS_DIALOG_OK then
         self.powerd.fl_intensity = self.powerd:frontlightIntensityHW()
         logger.dbg("Dialog OK, brightness: " .. self.powerd.fl_intensity)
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
Looks fine.

pazos (changes requested):
To match the behaviour of the title, please broadcast the event in the android implementation of showLightDialog and *only* if the user pressed the ok button.

frontend/device/android/device.lua
```diff
@@ -376,6 +376,9 @@ function Device:showLightDialog()
         if android.isWarmthDevice() then
```
pazos, line 376: just for warmth devices? Don't you need to broadcast frontlight state change after a intensity change?.

zwim: Ups. Of course!

pazos: moved at the end of the block

pazos (approved):

PR #6586, opened 2020-08-28
turn on frontlight switch on Tolino.

zwim:
Has to be done, so that the frontlight toggle also works when started with frontlightSwitch==off.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6586)

### DIFF
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -75,6 +75,7 @@ function AndroidPowerD:turnOnFrontlightHW()
     if self:isFrontlightOn() and self:isFrontlightOnHW() then
         return
     end
+    android.enableFrontlightSwitch()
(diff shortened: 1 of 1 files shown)


frontend/device/android/powerd.lua
```diff
@@ -75,6 +75,7 @@ function AndroidPowerD:turnOnFrontlightHW()
     if self:isFrontlightOn() and self:isFrontlightOnHW() then
         return
     end
+    android.enableFrontlightSwitch()
```
Frenzie, line 78: On non-Tolinos this just a no-op, right?

zwim: Yes, a method in `GenericController.kt` returning true.

Frenzie (approved):
Seems fine, a small comment for future clarity might be worthwhile?

PR #6587, opened 2020-08-28
Better zipping

zwim:
This gives about 1Meg smaller apk.
Compression time on a quad-core is about twice as `zip -r9`, when using `mx=7` its about the same time (of course then the compression is not as good, but saves 800k)
Decompression memory usage does not seem to be bigger: Tested with grep `grep VmPeak /proc/$PID/status` and `grep VmHWM /proc/$PID/status` showed almost identical values.

I have used this compression the last weeks on Tolino and phone without strange effects.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6587)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -354,15 +354,16 @@ androidupdate: all
 	# in runtime luajit-launcher's libluajit.so will be loaded
 	-rm $(INSTALL_DIR)/koreader/libs/libluajit.so
 
-	# assets are compressed manually and stored inside the APK.
-	cd $(INSTALL_DIR)/koreader && zip -r9 \
+	cd $(INSTALL_DIR)/koreader && 7z a -l -mx=9 -mfb=256 -mmt=on \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).zip * \
-		--exclude=*resources/fonts* \
-		--exclude=*resources/icons/src* \
-		--exclude=*share/man* \
-		--exclude=*spec* \
-		--exclude=*COPYING* \
-		--exclude=*README.md*
+		-xr!*resources/fonts* \
+		-xr!*resources/icons/src* \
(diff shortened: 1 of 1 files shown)


pazos:
Looks cool. An alternative to save a few MB would be https://commons.apache.org/proper/commons-compress/examples.html. and switch back to xz compression.

pazos (approved):

Frenzie (approved):
Is 7z available on F-Droid?

pazos:
> Is 7z available on F-Droid?

@licaon-kter ^^. Can we assume that the build machine has 7z binary installed?

Also, for future proof, it is ok to have this artifact as a dependency: https://mvnrepository.com/artifact/org.apache.commons/commons-compress ?

licaon-kter:
If Debian Stretch (for now) has it...

I'd go with xz though, as 7zip is soo old. But whatever you want.

zwim:
> I'd go with xz though, as 7zip is soo old. But whatever you want.

zip is even older!

I would have no problems with xz. In fact I would prefer it.
But there was a discussion #5264 (almost exact one year ago). I did not get the point, why KOReader moved to zip.

By the way. Using 7z for zip-file creation might be less invasive than switching to xz, my 5ct.

licaon-kter:
I meant old as in stuck at version 16 and no newer source code.

pazos:
> But there was a discussion #5264 (almost exact one year ago). I did not get the point, why KOReader moved to zip

Because ZipInputStream is part of java/android libraries. That makes possible to catch errors uncompressing, which was not possible using 7z via ffi. It should be a matter of replacing the offending function with an updated one using apache common compress artifact. See https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/utils/AssetsUtils.kt#L101-L151

Issue #6582, opened 2020-08-28
USBMS: no restart & when connecting immediately after start


gerhaher:
* KOReader version: v2020.08.1-21-g8296bea_2020-08-28/   
* Device: KA1

#### Issue
1. It doesn't restart when it says it should. It got stuck at 'Done! KOReader will now restart...' so I have to restart my device.

2. If I connect the device immediately after having started KOReader (I use KSM 09), I have to fast double tap on 'Share' to make it work. With normal single taps the popupwindow just toggle translucent and never disappears, so I have to restart the device.

NiLuJe:
1. You need to have done a full exit & start cycle after the update for this to actually work.

2. That's an unrelated issue with devices with a certain input event handler that mangles the first tap. Although I don't think I can reproduce that on my H2O, so it might be compounded by the terrible touch input of the KA1?

NiLuJe:
And even if I do force a long-hold to make the popup transparent, I can still tap or dismiss it just fine over here?

NiLuJe:
The only thing I can possibly trigger with a fast session start on plug is ending up with a busy mountpoint if I have very unlucky timing with the Wi-Fi script (which implies having restore Wi-Fi on resume enabled).

In my test, I hit the check *just* as dhcpd started, so it prevented the mount, but by the tem the process check ran, it was gone, so the list was empty ;).

NiLuJe:
Yeah, on my H2O, the first tap just goes to the ether, it's not detected as a hold, so at worst I need to tap twice, but that's it.

And that's not anything new, either.

gerhaher:
I tried my Kobo Aura: I could reproduce neither 1 nor 2. So everything OK with that one.

But KA1. Restart just doesn't work. The only thing that happens is that the light goes off after 90 seconds. Stuck!

NiLuJe:
Is that after a full reboot?

If so, check the syslog over Wi-Fi `logread -F`.

NiLuJe:
And also double-check the env from a running luajit process, check that `KOREADER_DIR` is sane.

Is your launcher setup strictly identical between the two devices?

gerhaher:
I have no Wifi at the moment.

> Is your launcher setup strictly identical between the two devices?

I think so.

> And also double-check the env from a running luajit process, check that KOREADER_DIR is sane.

Teach me how.


[...]

PR #6590, opened 2020-08-28
More KoboUSBMS related tweaks & cleanups

NiLuJe:
* Add some debugging in the startup script to aid in figuring out #6582 
* Bump KoboUSBMS for a minor tweak ensuring the final screen will never be optimized out by the EPDC
* Move some items around in the Hamburger menu to keep it single page (c.f., discussions in #6566)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6590)


NiLuJe:
I also checked that we don't get a double-separator on device without the UBSMS cap ;) (Yay, menu code!).

poire-z (approved):

PR #6594, opened 2020-08-29
[Android] update status bar on toggle of frontlight switch on Tolinos and status bar integration with native light dialog on all Android

zwim:
This is an extensive change, focused mainly on the status bar updates.
It can only applied together/after with https://github.com/koreader/android-luajit-launcher/pull/249

Only Android systems which have a frontlight toggle switch should be affected by this PR. On the other Androids is_fl_sw_on (is frontligh switch on) is always true.
Just as a remainder: Tolino Epos 2 has a software toggle (which emulates a keypress) and Tolino Vision4Hd has the same software toggle (in addition there is a hardware button, too).

So we have two different toggles. Our KOReader thing and that software/hardware button on the Tolinos.

On each keypress, change of intensity the statusline has to be updated.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6594)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -25,6 +25,7 @@ local _ = require("gettext")
 local C_ = _.pgettext
 local Screen = Device.screen
 
+
 local MODE = {
     off = 0,
     page_progress = 1,
@@ -106,7 +107,7 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].frontlight
         local powerd = Device:getPowerDevice()
-        if powerd:isFrontlightOn() then
+        if powerd:isFrontlightOn() and powerd:getFrontlightSwitchState() then
             if Device:isCervantes() or Device:isKobo() then
                 return (prefix .. " %d%%"):format(powerd:frontlightIntensity())
             else
@@ -114,6 +115,7 @@ local footerTextGeneratorMap = {
             end
         else
             return T(_("%1 Off"), prefix)
+
         end
     end,
     battery = function(footer)
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -107,7 +107,7 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].frontlight
         local powerd = Device:getPowerDevice()
-        if powerd:isFrontlightOn() and powerd:getFrontlightSwitchState() then
+        if powerd:isFrontlightOn() then
             if Device:isCervantes() or Device:isKobo() then
                 return (prefix .. " %d%%"):format(powerd:frontlightIntensity())
             else
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -387,6 +387,14 @@ end
 
 function Device:showLightDialog()
     local usleep = require("ffi/util").usleep
+    local old_is_fl_sw_on = self.powerd.is_fl_sw_on
+    if not self.powerd.is_fl_sw_on then
+        android:enableFrontlightSwitch()
+        self.powerd.is_fl_sw_on = true
+    end
+
+    self.powerd:setIntensityHW(self.powerd.fl_intensity)
+
     local title = android.isEink() and _("Frontlight settings") or _("Light settings")
     android.lights.showDialog(title, _("Brightness"), _("Warmth"), _("OK"), _("Cancel"))
     repeat
@@ -400,21 +408,38 @@ function Device:showLightDialog()
     local action = android.lights.dialogState()
     if action == C.ALIGHTS_DIALOG_OK then
         self.powerd.fl_intensity = self.powerd:frontlightIntensityHW()
+        if self.powerd.fl_intensity > self.powerd.fl_min then
+            self.powerd.is_fl_on = true
+        else
+            self.powerd.is_fl_on = false
+        end
         logger.dbg("Dialog OK, brightness: " .. self.powerd.fl_intensity)
         if android.isWarmthDevice() then
             self.powerd.fl_warmth = self.powerd:getWarmth()
             logger.dbg("Dialog OK, warmth: " .. self.powerd.fl_warmth)
         end
-        local Event = require("ui/event")
-        local UIManager = require("ui/uimanager")
-        UIManager:broadcastEvent(Event:new("FrontlightStateChanged"))
+        return true
     elseif action == C.ALIGHTS_DIALOG_CANCEL then
         logger.dbg("Dialog Cancel, brightness: " .. self.powerd.fl_intensity)
-        self.powerd:setIntensityHW(self.powerd.fl_intensity)
+        if not old_is_fl_sw_on then
+            local old_intensity = self.powerd.fl_intensity
+            self.powerd:setIntensityHW(self.powerd.fl_min)
+            self.powerd.fl_intensity = old_intensity
+            self.powerd.is_fl_on = false
+        else
+            if self.powerd.is_fl_on then
+                self.powerd:setIntensityHW(self.powerd.fl_intensity)
+            else
+                local old_intensity = self.powerd.fl_intensity
+                self.powerd:setIntensityHW(self.powerd.fl_min)
+                self.powerd.fl_intensity = old_intensity
+            end
+        end
         if android.isWarmthDevice() then
             logger.dbg("Dialog Cancel, warmth: " .. self.powerd.fl_warmth)
             self.powerd:setWarmth(self.powerd.fl_warmth)
         end
+        return false
     end
 end
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -418,6 +418,7 @@ function Device:showLightDialog()
             self.powerd.fl_warmth = self.powerd:getWarmth()
             logger.dbg("Dialog OK, warmth: " .. self.powerd.fl_warmth)
         end
+        self.powerd:broadcastLightChanges()
         return true
     elseif action == C.ALIGHTS_DIALOG_CANCEL then
         logger.dbg("Dialog Cancel, brightness: " .. self.powerd.fl_intensity)
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -6,22 +6,16 @@ local AndroidPowerD = BasePowerD:new{
     fl_max = 100,
 }
 
--- Let the footer know of the change
-local function broadcastLightChanges()
-    if package.loaded["ui/uimanager"] ~= nil then
-        local Event = require("ui/event")
-        local UIManager = require("ui/uimanager")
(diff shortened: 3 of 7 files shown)


pazos:
I'm not very fan of mapping specific buttons on all android devices and I dislike adding new variables, like, `powerd.is_fl_sw_on`.

Nothing against this specific PR, but I had "review" a few of yours related to the frontlight behaviour of the Tolinos. I would suggest to keep this open until you finally came with a solution that fullfits your needs :)

zwim:
@pazos: This version is working on Tolino Epos2 and Vision4HD as well as on a phone.

frontend/device/android/powerd.lua
```diff
@@ -6,22 +6,16 @@ local AndroidPowerD = BasePowerD:new{
...
-end
-
 function AndroidPowerD:frontlightIntensityHW()
     return math.floor(android.getScreenBrightness() / self.bright_diff * self.fl_max)
 end
 
 function AndroidPowerD:setIntensityHW(intensity)
     -- if frontlight switch was toggled of, turn it on
-    android:enableFrontlightSwitch()
+    if not self.is_fl_sw_off then
```
pazos, line 15: typo? off != on

frontend/device/devicelistener.lua
```diff
@@ -227,7 +232,9 @@ if Device:hasFrontlight() then
     end
 
     function DeviceListener:onShowFlDialog()
-        Device:showLightDialog()
+        if Device:showLightDialog() then
+            UIManager:broadcastEvent(Event:new("FrontlightStateChanged"))
```
pazos, line 236: I don't like it. showLightDialog should broadcast events directly (as it does now). Is there a benefit of doing this way?

frontend/device/input.lua
```diff
@@ -341,6 +341,14 @@ function Input:handleKeyBoardEv(ev)
         return
     end
 
+    -- A Button to turn on/off the frontlight
+    if keycode == "FLSwitch" and ev.value == EVENT_VALUE_KEY_RELEASE then
```
pazos, line 345: I'm afraid I don't understand. This detects a keycode and calls `powerd:detectedFrontlightSwitchToggle()` which indeed generates a key event that is detected by this¿?

frontend/device/input.lua
```diff
@@ -341,6 +341,14 @@ function Input:handleKeyBoardEv(ev)
...
+        local Device = require("device")
+        local powerd = Device:getPowerDevice()
+        powerd:detectedFrontlightSwitchToggle()
+        return "FLSwitch"
```
pazos, line 349: why?. You don't handle it in UIManager. Should be `return` instead?

frontend/device/android/powerd.lua
```diff
@@ -68,20 +64,45 @@ function AndroidPowerD:turnOffFrontlightHW()
...
+    self:broadcastLightChanges()
+end
+
+function AndroidPowerD:getFrontlightSwitchState()
+    return self.is_fl_sw_on
+end
+
+function AndroidPowerD:detectedFrontlightSwitchToggle()
+    if self.is_fl_sw_on and not android:getFrontlightSwitchState() then
+        android:enableFrontlightSwitch() --this sends keypress to android (and KOReader)
```
pazos, line 93: it looks like a loop

frontend/device/android/device.lua
```diff
@@ -400,21 +408,38 @@ function Device:showLightDialog()
...
+                local old_intensity = self.powerd.fl_intensity
+                self.powerd:setIntensityHW(self.powerd.fl_min)
+                self.powerd.fl_intensity = old_intensity
+            end
+        end
         if android.isWarmthDevice() then
             logger.dbg("Dialog Cancel, warmth: " .. self.powerd.fl_warmth)
             self.powerd:setWarmth(self.powerd.fl_warmth)
         end
+        return false
```
pazos, line 442: the whole function looks a bit complicate right now and I don't even understand why all that is needed :(

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -106,14 +107,15 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].frontlight
         local powerd = Device:getPowerDevice()
-        if powerd:isFrontlightOn() then
+        if powerd:isFrontlightOn() and powerd:getFrontlightSwitchState() then
```
pazos, line 110: I would like to include getFrontlightSwitchState as part of the return value of isFrontlightOn and leave this as is

frontend/device/android/device.lua
```diff
@@ -387,6 +387,14 @@ end
 
 function Device:showLightDialog()
     local usleep = require("ffi/util").usleep
+    local old_is_fl_sw_on = self.powerd.is_fl_sw_on
+    if not self.powerd.is_fl_sw_on then
+        android:enableFrontlightSwitch()
+        self.powerd.is_fl_sw_on = true
+    end
+
+    self.powerd:setIntensityHW(self.powerd.fl_intensity)
```
pazos, line 396: I don't understand this one. Why we set fl intensity before showing the dialog?

frontend/device/android/device.lua
```diff
@@ -400,21 +408,38 @@ function Device:showLightDialog()
...
+        end
         logger.dbg("Dialog OK, brightness: " .. self.powerd.fl_intensity)
         if android.isWarmthDevice() then
             self.powerd.fl_warmth = self.powerd:getWarmth()
             logger.dbg("Dialog OK, warmth: " .. self.powerd.fl_warmth)
         end
-        local Event = require("ui/event")
-        local UIManager = require("ui/uimanager")
-        UIManager:broadcastEvent(Event:new("FrontlightStateChanged"))
+        return true
```
pazos, line 421: The old way was fine as it was. You can change it to self.powerd:broadcast if you want

pazos (commented):
Changes look a bit overcomplicated.

I don't understand the need to generate a input event and catch that input event in the input handler.
I'm afraid it would be easier to understand with the device in hand.

DId you test your changes in a Tolino device without root?. Does it return the current state as well?

frontend/device/android/powerd.lua
zwim: Yes. This leads to a minimal screenflicker. Corrected

frontend/device/input.lua
zwim: Yes. 
1) Assume KOreader is started with HardwareSwitch set to on. In KOReader we toggle FL off. Then someone toggles the Hardware Switch on the Device -> FL is set to off by hardware and not toggled on again.
Therefore if a Hardware keypress (FLSwitch) is detected by KOReader AND we know the HardwareSwitch is off then we send another keypress and handle is_fl_on to on.
2) Assume KOReader all intensity on. Turn off the HardwareSwitch. We check that in the first `    if self.is_fl_sw_on and not android:getFrontlightSwitchState() then` and send another key to the system (Hardware light is on, but we want no intensity -> therefore we can set is_fl_on to off
1+2) can be combined with `is_fl_on = not is_fl_on`
At the end of 1+2 we are sure SoftwareSwitch is on, and is_fl_on shows KOReaders toggle state

3) Assume KOReader is started with HardwareSwitch set to off. If we want to change the intensity, we send a keypress to the system (this turns Hardware-Lighting on) and unfortunately we get it also in KOReader as a keyevent. In this case we just have to toggle is_fl_sw_on 

As we might have touched `is_fl_on` we have to set the correct android intensity at the end of the method. And broadcast a changedLightLevel.

frontend/device/input.lua
zwim: OK.

frontend/device/android/powerd.lua
zwim: Only at the first glance. 
Imagine:
1.) HardwareSwitch is on first -> is_fl_sw_on==true and getFrontligthSwitchState==true
2.) Someone toggle the Hardware switch -> is_fl_sw_on=true and getFrontlightSwitchState==false
3.) We send another keypress to the system (to toggle HardwareSwitch on again to on)
4.) This time is_fl_sw_on==true and getFrontlightSwitchState==true -> no keypress

frontend/device/android/device.lua
zwim: Yes. Does not look simple. But with that HardwareSwitch things got complicated.
It could be, that the dialog is started with HardwareSwitch off of on.
We have to turn on HardwareSwitch on start of the dialog, so that someone can see the dialog in the dark (perhaps with the intensity which was before a toggle). 

At a cancel we have to restore the state before, but we have turned on the HardwareSwitch. If it was off before the dialog, we turn of is_fl_on and change the screen brightness to zero, but we also don't want to loose the old fl_intensity (`self.powerd:setIntensityHW(self.powerd.fl_min` changes fl_intensity). Here we don't need a broadcastLightChange.

At a accept we have to turn on is_fl_on, when the fl_intensity>fl_min, else if_sw_on==false.

frontend/apps/reader/modules/readerfooter.lua
zwim: Good point.

frontend/device/devicelistener.lua
zwim: As you wish.

frontend/device/android/device.lua
zwim: Line393: If HardwareSwitch is, the dialog could not change the intensity.

Line 396: If the screen was off, but we want to see the native dialog (in the dark) witch the intensity it had before the toggle.

frontend/device/android/device.lua
zwim: OK

frontend/device/android/powerd.lua
pazos: But `android.enableFrontlightSwitch()` (there's no need to pass self) requires root. Or not? what happens if it requires root and the device is not rooted?

frontend/device/android/device.lua
pazos: I still don't understand why we need to handle the state of the hw/sw switch. Benefits?. If I did understand correctly we use the switch to being able to change fl without root, but toggling the switch needs root, I'm totally lost here :(

frontend/device/android/device.lua
pazos: sorry, don't follow. On most android devices this will set intensity to itself, and this is wrong.

pazos (commented):
I still think that I'm going to be unable to understand the changes in the frontend. They're quite targeted at a specific device with an odd behaviour.

I think most of the changes can be done on the backend without reaching the device input handler (look how certain events are skipped in base, based on user config) But need to consider first if we need to handle them at all. If the device requires root to change fl and still needs root after this PR I don't understand the purpose.

At this point I would like to delegate the review process, because it drives me mad (oh, boy I hate the frontlight logic :cry: )

zwim:
I would like to wait for a review. This works on a rooted device. 
For a non rooted device there would be three addions necessary:
1.) Detect if device is rooted -> `is_rooted` 2.) change `detectedFrontlightSwitchToggle` to do nothing if `not is_rooted` 3.) change `isFrontlightOn` to honor `not is_rooted` to display the correct front light level.

As I have a rooted device, I will add this, when I get the principally OK for this work.

Frenzie:
@pazos I don't have anything to add to what you wrote. Thanks for the effort. :-)

PR #6596, opened 2020-08-29
Enabled Serbian typography rules

strn:
Serbian as language exists in CRengine.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6596)


Frenzie (approved):

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -41,6 +41,7 @@ local HYPH_DICT_NAME_TO_LANG_NAME_TAG = {
     ["Russian_EnGB.pattern"] = { _("Russian + English (UK)"), "ru-GB" },
     ["Russian_EnUS.pattern"] = { _("Russian + English (US)"), "ru-US" },
     ["Russian.pattern"]      = { _("Russian"),      "ru" },
+    ["Serbian.pattern"]      = { _("Serbian"),      "sr" },
```
poire-z, line 44: We don't have any Serbian.pattern in https://github.com/koreader/crengine/tree/master/cr3gui/data/hyph/ 
The fact you added it here is may be a non-issue (it's for migration) - but I think then you could just let it out.

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -86,6 +87,7 @@ local LANGUAGES = {
     { "ru-GB",                 {},   "Hb  ",   _("Russian + English (UK)") },
     { "ru-US",                 {},   "Hb  ",   _("Russian + English (US)") },
     { "ru",               {"rus"},   "Hb  ",   _("Russian") },
+    { "sr",               {"srp"},   "H   ",   _("Serbian") },
```
poire-z, line 90: And if you're no bringing a Serbian.pattern, just remove the H from the features string (it's for display in the menu, describing what features are available). Let it as for "ja" and "ko".

frontend/apps/reader/modules/readertypography.lua
Frenzie: Thanks, now I'm confused by this PR.

frontend/apps/reader/modules/readertypography.lua
strn: Sorry about the confusion. Can you please suggest what else I need to add in order to proper add typographic rules for Serbian language?

poire-z: As you added _`duplicate_real_hyphen_on_next_line` for Serbian , you can replace your `H` here with `HB`.

poire-z:
I'm not allowed to push commits to your branch. Can you enable it (somewhere on github)?
I'd like to get this merged, so it does not block koreader-base.
If no news in the afternoon, I'll merge it and do another PR to bump base and update "H" to "HB".

strn:
Sorry for late reply - cannot access email or Github from work. What shall I do? Can you point it to me?

poire-z:
No need to anymore, I did the bump earlier in #6618.
So, it will all be in tomorrow's nightly.

PR #6598, opened 2020-08-29
Add gesture for invoking terminal

smartscripts-nl:
Make the terminal emulator available through a gesture. Handy for technical minded users.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6598)

### DIFF
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -49,6 +49,7 @@ local settingsList = {
     history = { category="none", event="ShowHist", title=_("History"), device=true,},
     open_previous_document = { category="none", event="OpenLastDoc", title=_("Open previous document"), device=true,},
     filemanager = { category="none", event="Home", title=_("File browser"), device=true,},
+    show_terminal = { category = "none", event = "TerminalStart", title = 'Show terminal', device = true, },
     dictionary_lookup = { category="none", event="ShowDictionaryLookup", title=_("Dictionary lookup"), device=true,},
     wikipedia_lookup = { category="none", event="ShowWikipediaLookup", title=_("Wikipedia lookup"), device=true,},
     fulltext_search = { category="none", event="ShowFulltextSearchInput", title=_("Fulltext search"), device=true,},
@@ -156,6 +157,7 @@ local dispatcher_menu_order = {
     "open_previous_document",
     "favorites",
     "filemanager",
+    "show_terminal",
     "stats_calendar_view",
 
     "dictionary_lookup",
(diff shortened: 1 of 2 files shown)


frontend/dispatcher.lua
```diff
@@ -49,6 +49,7 @@ local settingsList = {
     history = { category="none", event="ShowHist", title=_("History"), device=true,},
     open_previous_document = { category="none", event="OpenLastDoc", title=_("Open previous document"), device=true,},
     filemanager = { category="none", event="Home", title=_("File browser"), device=true,},
+    show_terminal = { category = "none", event = "TerminalStart", title = 'Show terminal', device = true, },
```
Galunid, line 52: `title = 'Show terminal'` doesn't allow for translation, it should be `title = _('Show terminal')` instead, and no spaces between assignments inside table, to be consequential ;-). Otherwise lgtm.

poire-z: And may be move them (with `edit_last_edited_file` from #6600), later in the list and the order, after Wallabag/calibre, in a separator section of their own (these are plugins, and aren't much related to reading like the first entries are).

Frenzie: What they said, and thanks for the PR! :-)

poire-z:
Now that #6600 is merged, this has conflict because they touched the same area.
So just rebase master, and decices which of the two is to be first, and remove separator=true from one of them (the first or the last, dunno, please test it :) so they are in the same subsection.

poire-z:
Please fix https://github.com/koreader/koreader/pull/6600#discussion_r480170435 in this one.
(Can you get quickly done with this one ? :) The more PRs you have opened, the more risk of conflicts and the need to fix any of them.)

smartscripts-nl:
> Please fix [#6600 (comment)](https://github.com/koreader/koreader/pull/6600#discussion_r480170435) in this one.
> (Can you get quickly done with this one ? :) The more PRs you have opened, the more risk of conflicts and the need to fix any of them.)

Maybe I can now remove this PR, because I have fixed #6600 elsewhere?

poire-z:
But there is currently no way for a gesture to show terminal...  which was the purpose of this PR (you could have added that to #6609).
So, either you don't really need it :) or you clean up this PR and continue doing it here.

smartscripts-nl:
> But there is currently no way for a gesture to show terminal... which was the purpose of this PR (you could have added that to #6609).
> So, either you don't really need it :) or you clean up this PR and continue doing it here.

I don't really know how to clean this up. So I think I better close this PR and start a new one with the same name, to avoid conflicts with the most recent state of upstream master.

poire-z:
As you wish.
(But in the future, avoid multiple related PRs so you don't get into such situations.)

smartscripts-nl:
Yes, I have learned that lesson... 😊

PR #6599, opened 2020-08-29
Terminal: support multiple lines and make it re-invokeable

smartscripts-nl:
With this changes you can enter multiple lines with commands and after viewing the output you can re-invoke the terminal with a button

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6599)

### DIFF
--- a/plugins/terminal.koplugin/main.lua
+++ b/plugins/terminal.koplugin/main.lua
@@ -1,5 +1,4 @@
 local DataStorage = require("datastorage")
-local Event = require("ui/event")
 local Font = require("ui/font")
 local InfoMessage = require("ui/widget/infomessage")
 local InputDialog = require("ui/widget/inputdialog")
@@ -82,7 +81,7 @@ function Terminal:execute()
                     text = _("Back"),
                     callback = function()
                         UIManager:close(viewer)
-                        self.ui:handleEvent(Event:new("TerminalStart"))
+                        self:onTerminalStart()
                     end,
                 },
                 {
--- a/plugins/terminal.koplugin/main.lua
+++ b/plugins/terminal.koplugin/main.lua
@@ -21,13 +21,16 @@ function Terminal:init()
     self.ui.menu:registerToMainMenu(self)
 end
 
-function Terminal:start()
+function Terminal:onTerminalStart()
     self.input = InputDialog:new{
         title =  _("Enter a command and press \"Execute\""),
         input = self.command,
         para_direction_rtl = false, -- force LTR
         text_height = math.floor(Screen:getHeight() * 0.4),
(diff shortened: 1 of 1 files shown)


plugins/terminal.koplugin/main.lua
```diff
@@ -66,12 +68,32 @@ function Terminal:execute()
...
         title = _("Command output"),
         text = table.concat(entries, "\n"),
         justified = false,
         text_face = Font:getFace("smallinfont"),
-    })
+        -- support re-invoking the terminal with a button:
+        buttons_table = {
+            {
+                {
+                    text = 'New command(s)...',
```
poire-z, line 81: I'd like something better, but no idea what. Re-run ? Edit ?
Also, our prefered quotes for strings are "...", not '...'

poire-z:
Dunno if it's your bug, but while you're at it:
- quit & launch KOReader
- Terminal emulator: enter "ls"
- New command: blank window (I don't get my "ls" back)
- Enter "df"
- New command: I get my "df" back.

plugins/terminal.koplugin/main.lua
Frenzie: I think edit would be welcome but that's not what this does?

In this context, probably just `Run` would suffice.

plugins/terminal.koplugin/main.lua
```diff
@@ -78,10 +79,10 @@ function Terminal:execute()
         buttons_table = {
             {
                 {
-                    text = 'New command(s)...',
+                    text = _("Back"),
                     callback = function()
                         UIManager:close(viewer)
-                        Terminal:onTerminalStart()
+                        self.ui:handleEvent(Event:new("TerminalStart"))
```
poire-z, line 85: May be just `self:onTerminalStart()` (the instance, that remembers the input) rather than your initial `Terminal:onTerminalStart()` (the class object, which was another object), was just enough.

plugins/terminal.koplugin/main.lua
smartscripts-nl: I've now updated this to "Back". I have almost no experience with PR's: do you see my edits automatically, or do I have to do a PR again?

plugins/terminal.koplugin/main.lua
smartscripts-nl: You're right, thanks! I've modified the code accordingly.

plugins/terminal.koplugin/main.lua
Galunid: @smartscripts-nl You just commit and push to your remote branch, PR will update automatically ;-)

poire-z (approved):

PR #6602, opened 2020-08-30
Auto-standby plugin

ezdiy:
Adaptively pins the auto-standby UI lock. When frequent input from user is observed, we'll prevent auto-standby for increasingly longer periods of time, and revert back to more aggressive standby with infrequent input (ie when reading a book). Auto-standby is set aggressive (not adaptive) if battery is below certain threshold. Can also configure standby to be always prevented for lower input lag, shorter battery charge.

Depends on #6558

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6602)

### DIFF
--- a/plugins/autostandby.koplugin/main.lua
+++ b/plugins/autostandby.koplugin/main.lua
@@ -1,0 +1,160 @@
+local Device = require("device")
+
+if not Device:isPocketBook() --[[and not Device:isKobo()]] then
+    return { disabled = true }
+end
+
+local PowerD = Device:getPowerDevice()
+local DataStorage = require("datastorage")
+local LuaSettings = require("luasettings")
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local SpinWidget = require("ui/widget/spinwidget")
+local logger = require("logger")
+local _ = require("gettext")
+
+local AutoStandby = WidgetContainer:new{
+    is_doc_only = false,
+    name = "autostandby",
+    settings = LuaSettings:open(DataStorage:getSettingsDir() .. "/autostandby.lua"),
+    delay = 0,
+    lastInput = 0,
+    preventing = false,
+}
+
+function AutoStandby:init()
+    if not self.settings:has("filter") then
+        logger.dbg("AutoStandby: No settings found, initializing defaults")
+        self.settings.data = {
+            forbidden = false,  -- If forbidden, standby is never allowed to occur
+            filter = 1,         -- Consider input only further than this many seconds apart
+            min = 1,            -- Initial delay period during which we won't standby
+            mul = 1.5,          -- Multiply the delay with each subsequent input that happens, scales up to max
+            max = 30,           -- Input that happens further than 30 seconds since last input one reset delay back to 'min'
+            win = 5,            -- Additional time window to consider input contributing to standby delay
+            bat = 60,           -- If battery is below this percent, make auto-standby aggressive again (disables scaling by mul)
+        }
+        self.settings:flush()
+    end
+
+    UIManager.event_hook:registerWidget("InputEvent", self)
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function AutoStandby:addToMainMenu(menu_items)
+    menu_items.autostandby = {
+        text = _("Auto-standby settings"),
+        sub_item_table = {
+            {
+                keep_menu_open = true,
+                text = _("Allow auto-standby"),
+                checked_func = function() return self:isAllowedByConfig() end,
+                callback = function() self.settings:saveSetting("forbidden", self:isAllowedByConfig()):flush() end,
+            },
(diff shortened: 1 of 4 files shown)


plugins/autostandby.koplugin/main.lua
```diff
@@ -0,0 +1,166 @@
...
+}
+
+function AutoStandby:init()
+    if not self.settings:has('filter') then
+        logger.dbg("AutoStandby: No settings found, initializing defaults")
+        self.settings
+        -- If forbidden, standby is never allowed to occur
+        :saveSetting('forbidden', false)
+        -- Consider input only further than this many seconds apart
+        :saveSetting('filter', 1)
```
poire-z, line 32: Not used to see that kind of constructs in our code base :) but I guess it's fine - the alternative being assigning a plain table with these settings directly to self.settings.data I think.

plugins/autostandby.koplugin/main.lua
```diff
@@ -0,0 +1,166 @@
...
+            {
+                keep_menu_open = true,
+                text = _("Allow auto-standby"),
+                checked_func = function() return self:isAllowedByConfig() end,
+                callback = function() self.settings:saveSetting('forbidden', self:isAllowedByConfig()):flush() end,
+            },
+            self:_config_item(_("Min input idle seconds"), "min", function() return 0 end, function() return self.settings:readSetting("max") end),
+            self:_config_item(_("Max input idle seconds"), "max", function() return 0 end),
+            self:_config_item(_("Input window seconds"), "win", function() return 0 end, function() return self.settings:readSetting("max") end),
+            self:_config_item(_("Always standby if battery below"), "bat", function() return 0 end, function() return 100 end),
```
poire-z, line 63: In the rest of the code, such fonction is usually named getXXXMenuItem() or genXXXMenu():
```
Translator:genSettingsMenu()
ReaderHighlight:genHighlightDrawerMenu()
ReaderTypeset:genStyleSheetMenu()
sub_item_table = self:genStyleSheetMenu(),
ReaderView:genOverlapStyleMenu(overlap_enabled_func)
self:genDictionariesMenu(),
self:_genDownloadDictionariesMenu()
ReaderDictionary:_genDownloadDictionariesMenu()
ReaderDictionary:genDictionariesMenu()
genAutoSaveMenuItem(value)
genAutoSaveMenuItem(false),
genAutoSaveMenuItem(5),
genAutoSaveMenuItem(15),
genAutoSaveMenuItem(60),
```
So may be stick to that kind of name, so we know immediately what this is about.

poire-z (commented):
Not digging into the technical stuff, I trust you. But the code looks fine.

plugins/autostandby.koplugin/main.lua
NiLuJe: Yeah, I'm not sure what this actually does, actually (lua is weird)? implied self?

poire-z: Not Lua, but may be our Luasettings :) It's just that most of our `LuaSettings:saveSetting/delSetting/flipNilOrFalse` return self, so you can chain them like that.
I think that's something quite common in Javascript, to chain stuff like that.

NiLuJe:
Oh, okay, I wasn't aware that linebreaks could be abused like that, but the
rest makes sense, thanks :).

plugins/autostandby.koplugin/main.lua
Frenzie: We normally prefer double quotes btw.

plugins/autostandby.koplugin/main.lua
Frenzie: Agreed, something like `genConfigMenuItem` (?) would be more in line with our idiom.

Frenzie:
@NiLuJe It's not Python; you can do that in just about anything. ;-)

plugins/autostandby.koplugin/main.lua
ezdiy: Switched to setting data directly. This introduces new contract (of data being settable), but I reckon there's no scenario where that would hurt, as all the funky MT stuff should be hung on top of the settings object.

plugins/autostandby.koplugin/main.lua
ezdiy: went with genSpinMenuItem (as the widget template is specifically only for spinner, other widgets would get their own generator)

Frenzie (approved):
lgtm

ezdiy:
Hmm, now testing it now and the stylistic changes broke it somewhere.

poire-z:
Oups, didn't see that:
```
Luacheck results
Checking plugins/autostandby.koplugin/main.lua    1 warning

    plugins/autostandby.koplugin/main.lua:96:12: accessing undefined variable 'delay'
```

ezdiy:
Ye, for some reason luacheck hates me. Should I add on top of this branch, or new one+PR?

poire-z:
New PR I guess (dunno what happens if you push again here, but let's try it just to see :)

ezdiy:
Pushed on top of here, indeed it's deaf post-merge.

PR #6603, opened 2020-08-30
Detect early KoboUSBMS returns, and just restart KOReader

NiLuJe:
We know onboard is safe in these cases.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6603)


NiLuJe:
Depends on https://github.com/koreader/koreader-base/pull/1176

So, pending on the PB stuff being pushed to frontend & base bumped ;).

pazos (approved):

Frenzie (approved):

PR #3608, opened 2018-01-16
Do not execute background runner if device is suspended

Hzj-jie:
On kindle voyage, the background runner may still execute a while even device has been suspended. Because kindle stock software may wake up the device to update itself. The behavior of the stock software impacts KOReader. For example, the AutoFrontlight plugin may turn on the frontlight unexpectedly.

PR #6605, opened 2020-08-30
Tame BackgroundRunner: stop running when no more job

poire-z:
(I really don't like this plugin, and for a few years I've been having bad sleep, knowing that it _is_ running for nothing on thousands of devices and that users don't even know it - and it _is_ surely responsible for climate change and global warming :/ So, may be it's not too late... :)

A BackgroundRunner plugin instance will stop running (rescheduling a check every 2 seconds) when there is no (or no more) job to run.
Clients of this service now have to emit an event after adding a job into PluginShare.backgroundJobs, so an already loaded but stopped BackgroundRunner can notice it and start running again.
Ref https://github.com/koreader/koreader/pull/3608#issuecomment-357929349

Not tested on devices where it is really used: some kindle for autofrontlight, some kobo and cervantes for autowarmth. So, please do if you can (or report once it's merged).
Tested with this and a click on the Version menu item:
```diff
--- a/frontend/ui/elements/common_info_menu_table.lua
+++ b/frontend/ui/elements/common_info_menu_table.lua
@@ -20,6 +20,15 @@ common_info.version = {
         UIManager:show(InfoMessage:new{
             text = Version:getCurrentRevision(),
         })
+        table.insert(require("pluginshare").backgroundJobs, {
+                         when = 10,
+                         repeated = 3,
+                         executable = function()
+                            print("test")
+                         end,
+        })
+        local Event = require("ui/event")
+        UIManager:broadcastEvent(Event:new("BackgroundJobsUpdated"))
     end
 }
```

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6605)

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -137,6 +137,7 @@ files["spec/unit/*"].globals = {
     "package",
     "requireBackgroundRunner",
     "stopBackgroundRunner",
+    "notifyBackgroundJobsUpdated",
 }
 
 -- TODO: clean up and enforce max line width (631)
--- a/frontend/device/cervantes/powerd.lua
+++ b/frontend/device/cervantes/powerd.lua
@@ -178,6 +178,11 @@ function CervantesPowerD:calculateAutoWarmth()
                              end
                          end,
         })
+        if package.loaded["ui/uimanager"] ~= nil then
+            local Event = require("ui/event")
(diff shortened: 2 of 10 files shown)


Frenzie (approved):
Cool, I always disable this one tbh. :-P

PR #6608, opened 2020-08-31
Add a crash.log viewer plugin

smartscripts-nl:
Search for errors in the crash log. Has two viewing modes: only show lines with errors, or show more info (with lots of lines which are irrelevant for the error filtered away). Also has a gesture to invoke the viewer.

I've incorporated some helper methods which in my own installation I've put in external helper files.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6608)


smartscripts-nl:
What does it mean that tests on CircleCI failed?

smartscripts-nl:
I could use some help with this plugin: for some reason my attempts to empty/reset crash.log don't work, even if I try to accomplish this with a shellscript.

smartscripts-nl:
Reader_2020-08-31_125454
Reader_2020-08-31_125553

poire-z:
Mhhh, a bit tedious to make that a plugin.
You could just add some menu items to `Developer options>` like (from my own patch, in which I have a few more tools):
```lua
{
    text = _("Truncate crash.log"),
    keep_menu_open = true,
    callback = function()
            text = _("Really truncate crash.log?"),
            ok_callback = function()
                local file = io.open("crash.log", "w")
                if file then file:close() end
                logger.info("[crash.log truncated]")
            end,
        })
    end,
},
{
    text = _("Show full crash.log"),
    keep_menu_open = true,
    callback = function()
        local TextViewer = require("ui/widget/textviewer")
        local file = io.open("crash.log", "r")
        local content = "no content"
        if file then
            content = file:read("*all")
            file:close()
        end
        local textviewer = TextViewer:new{
            title = "crash.log",
            text = content,
            text_face = Font:getFace("infont", 10),
            justified = false,
        }
        -- scroll to bottom
        textviewer.scroll_text_w:scrollToRatio(1)
        UIManager:show(textviewer)
        return
    end,
},
{
    text = _("Show crash.log tail"),
    keep_menu_open = true,
    callback = function()
        local TextViewer = require("ui/widget/textviewer")
        local file = io.open("crash.log", "r")
        local content = "no content"
        local maxsize = 20000
        if file then
            file:seek("end", -maxsize)
            content = file:read("*all")
            file:close()
            if content:len() == maxsize then
                local first_cr = content:find("\n")
                if first_cr then
                    content = content:sub(first_cr+1)
                end
                content = "[Start of crash.log not shown]\n" .. content
            else
                content = "[Start of crash.log]\n" .. content
            end
        end
        local textviewer = TextViewer:new{
            title = "crash.log",
            text = content,
            text_face = Font:getFace("infont", 10),
            justified = false,
        }
        -- scroll to bottom
        textviewer.scroll_text_w:scrollToRatio(1)
        UIManager:show(textviewer)
        return
    end,
    separator = true,
},
```

And I'm not sure filtering ERROR has any real interest: we don't output much of them, and if any, the context (lines without ERROR) is probably as important.
(And scripts in scripts/ are a bit of some overhead, you could just inline the shell command in the lua.)

If we should have more stuff like that, may be it could be in a `developpertools.plugin` which would export just a bunch of menu items.

NiLuJe:
`os.system()` > one-line shell scripts ;).

Also, I don't think every busybox implementation out there will be able to deal sanely with something mangling the log from under its feet?

smartscripts-nl:
> Mhhh, a bit tedious to make that a plugin.
> You could just add some menu items to `Developer options>` like (from my own patch, in which I have a few more tools):
> 
> ```lua
> {
>     text = _("Truncate crash.log"),
>     keep_menu_open = true,
>     callback = function()
>         local ConfirmBox = require("ui/widget/confirmbox")
>         UIManager:show(ConfirmBox:new{
>             text = _("Really truncate crash.log?"),
>             ok_callback = function()
>                 local file = io.open("crash.log", "w")
>                 if file then file:close() end
>                 local logger = require("logger")
>                 logger.info("[crash.log truncated]")
>             end,
>         })
>     end,
> },
> {
>     text = _("Show full crash.log"),
>     keep_menu_open = true,
>     callback = function()
>         local TextViewer = require("ui/widget/textviewer")
>         local file = io.open("crash.log", "r")
>         local content = "no content"
>         if file then
>             content = file:read("*all")
>             file:close()
>         end
>         local Font = require("ui/font")
>         local textviewer = TextViewer:new{
>             title = "crash.log",
>             text = content,
>             text_face = Font:getFace("infont", 10),
>             justified = false,
>         }
>         -- scroll to bottom
>         textviewer.scroll_text_w:scrollToRatio(1)
>         UIManager:show(textviewer)
>         return
>     end,
> },
> {
>     text = _("Show crash.log tail"),
>     keep_menu_open = true,
>     callback = function()
>         local TextViewer = require("ui/widget/textviewer")
>         local file = io.open("crash.log", "r")
>         local content = "no content"
>         local maxsize = 20000
>         if file then
>             file:seek("end", -maxsize)
>             content = file:read("*all")
>             file:close()
>             if content:len() == maxsize then
>                 local first_cr = content:find("\n")
>                 if first_cr then
>                     content = content:sub(first_cr+1)
>                 end
>                 content = "[Start of crash.log not shown]\n" .. content
>             else
>                 content = "[Start of crash.log]\n" .. content
>             end
>         end
>         local Font = require("ui/font")
>         local textviewer = TextViewer:new{
>             title = "crash.log",
>             text = content,
>             text_face = Font:getFace("infont", 10),
>             justified = false,
>         }
>         -- scroll to bottom
>         textviewer.scroll_text_w:scrollToRatio(1)
>         UIManager:show(textviewer)
>         return
>     end,
>     separator = true,
> },
> ```
> 
> And I'm not sure filtering ERROR has any real interest: we don't output much of them, and if any, the context (lines without ERROR) is probably as important.
> (And scripts in scripts/ are a bit of some overhead, you could just inline the shell command in the lua.)
> 
> If we should have more stuff like that, may be it could be in a `developpertools.plugin` which would export just a bunch of menu items.

But I like to be able to call the viewer with a gesture. That's why I made a plugin for it.

poire-z:
> But I like to be able to call the viewer with a gesture. That's why I made a plugin for it.

OK then, but so, let's have that plugin more generic and call it `developertools.koplugin/` `DeveloperTools`, so we can extend it and add various other stuff.

@NiLuJe : thoughts ? vs the current Developer options (could/should they be migrated in there) ?

We'd also need Dispatcher to provide a way for it to be fed with actions from plugins. @yparitcher : can you give this some thought ? With a new category `Plugins>` maybe ?

For what it's worth, I add these menu items in some core patch of mine:
```lua
text = _("Various stuff"),
    return _("Temperature: ") -- from "/sys/devices//platform/imx_epdc_fb/graphics/fb0/temperature"
    text = _("Logs, settings & system"),
        text = _("Debug mode"),
        text = _("Log scheduled action runs"),
        text = _("Truncate crash.log"),
        text = _("Show full crash.log"),
        text = _("Show crash.log tail"),
        text = _("Backup settings.reader.lua"),
        text = _("Show settings.reader.lua"),
        text = _("Show document metadata.lua"),
        text = _("Show ps output"),
    text = _("Caches"),
        T(_("Set crengine storage size factor (%1)")
        text = _("Cache small cre files too"),
        text = _("Disable crengine cache"),
        text = _("Don't compress crengine cache files"),
        text = _("Store crengine cache files in RAM"),
        text = _("Store other caches (PDF) in RAM"),
        text = _("Show caches size"),
        text = _("Show caches content"),
        text = _("Clear readers' caches"),
        return T(_("Book cre DOM version:  %1 %2  "), current, state)
            return T(_("Engine latest DOM version: %1"), ReaderUI:_getRunningInstance() and
                text = "20171225: oldest",
                text = "20180503: html BR fix",
                text = "20180524: LI block, inline as default",
                text = "20180528: epub.css, img inline, more block elems",
                text = "20200223: complete incomplete tables",
                text = "20200824: new HTML parser",
        return T(_("Book cre cache size:  %1%2"), cachesize, kept)
            return _("No cre cache file used")
            return _("Not a credocument, no cre cache")
        return _("Reload current book with cache cleaned")
```
(Most of the stuff I rarely or never use - but I had some plan to add the last ones about cre cache and dom versions, but I never found a good place for them.)

NiLuJe:
Some of the current developer options are pretty low-level/core, and would probably not be terribly suited for a plugin?

NiLuJe:
On the other hand, I agree that *this* would definitely belong in a plugin :s.

Also, it needs to be disabled on Android, because there's no logfile there.

smartscripts-nl:
There is still a conflict with master, but I don't know how to fix this.

This plugin is mainly directed at finding errors in my own scripts and works very good for that.

I'll push a separate PR for a DeveloperTools plugin. That is no more than a simple button dialog, which can start other utilities. Like the Terminal and this crash.log viewer.
errors-only
errors-with-context

Frenzie:
You can't fix a conflict without explicitly merging. See https://git-scm.com/docs/git-merge but since this was basically a non-issue I used the GH web GUI.

scripts/tail-crashlog.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+## BEREKEN TOTAAL AANTAL BOEKEN ##
+
+tail -n 600 ./crash.log  > ./scripts/tail-crashlog.txt
```
Frenzie, line 5:
```suggestion
## BEREKEN TOTAAL AANTAL BOEKEN ##

tail -n 600 ./crash.log >./scripts/tail-crashlog.txt
```

(Although does this require a shell invocation?)

NiLuJe: I vaguely remember mentioning using `os.exec` instead of shell one-liners in the past, yeah ^^.

ezdiy:
> This plugin is mainly directed at finding errors in my own scripts and works very good for that.

NiLuJe:
@ezdiy: That would be at the very least be tied to the glibc version. Possibly the GCC one, since I *think* we link against the STL dynamically in emulator builds...

I've tried that in the past, and, being on Gentoo, people in Debian-land weren't super enthused about my bleeding edge glibc versions ;p. (Hi, @pgaskin ;p).

ezdiy:
@NiLuJe presumably build it in virtenv with sufficiently ancient debian inside. glibc has fairly good backcompat, for as long you version the symbols low during the build. I'm more worried about SDL.

NiLuJe:
@ezdiy: That's a somewhat different proposition that "just throw the output of kodev in a tarball" ^^. But, yeah, doing it from a stable Debian docker image should generate something that mostly works on any setup that that doesn't deserve to be put down in the first place ;).

We don't ship SDL (at all, I think, even on AppImage?). That might have changed at one point, though. But, yeah, that's still an issue, because we might be expecting symbols from 2.x.y at runtime. I *think* that (i.e., the minimum version) may be documented in the sdl ffi/cdecl commits, though.

pazos:
> On that note, I notice there are no public emulator builds for plain and simple script editing. Closest to this is appimage, but it seems to be too much of "drag and drop" for its own good, as it hides the scripts from the user

There's a hint in https://github.com/koreader/koreader/blob/master/doc/Building.md

> If you only want to work with Lua frontend stuff, you can grab the AppImage and run it with --appimage-extract.

----

> We don't ship SDL (at all, I think, even on AppImage?). That might have changed at one point, though

We copy the so from Ubuntu 16.04 -> https://github.com/koreader/koreader/blob/master/Makefile#L324.

SDL was added as thirdparty dep for macos, to avoid runtime dependencies from brew, so it is a matter of building that for appimage too if that's needed.

Frenzie:
@ezdiy 
> On that note, I notice there are no public emulator builds for plain and simple script editing. Closest to this is appimage, but it seems to be too much of "drag and drop" for its own good, as it hides the scripts from the user. Not sure how viable is to simply ship the output of ./kodev build as is. IMO it should work pretty well on most linux distros, or even windows linux subsystem.

It might hide it a little bit, but that's why I wrote this at the top of the README file (as I just noticed @pazos already pointed out):

> If you only want to work with Lua frontend stuff, you can grab the AppImage and run it with --appimage-extract.

<hr>

@NiLuJe 
> We don't ship SDL (at all, I think, even on AppImage?).

Since the very beginning, but there was some minor issue with loading it at one point (perhaps only on Fedora). It'd been quite a while since I wrote the thing, so I forgot I'd done that and accidentally fell back to regular emulator setup support saying it looked like the system was missing SDL2. I assume this is what you're remembering. ;-)

<hr>

Wrt to shipping it without using the appimagetool on it, *if* really desired you'd probably just want to stick a zip command together with the appimagetool command. That way you'd have something like a pre-extracted AppImage which should pretty much Just Work™.

https://github.com/koreader/koreader/blob/a50774003f8ca670b783347e6761949b4682ef9c/Makefile#L347-L350

pazos:
@smartscripts-nl: please have a try with https://github.com/koreader/koreader/pull/6756 and see if you miss something from this PR. I didn't work on it on the past days, but I think I'm going to add user hooks to it, so it will be possible to add a filter for each log based on user preferences.

PR #6607, opened 2020-08-31
Terminal: add shortcuts for regularly used commands

smartscripts-nl:
Terminal: add code to store sets of commands as shortcuts, which can then later be executed with one click. These shortcuts can also be edited.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6607)

PR #6609, opened 2020-08-31
Terminal: add shortcuts for re-useable commands

smartscripts-nl:
To be able to quickly repeat often used terminal commands

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6609)

### DIFF
--- a/plugins/terminal.koplugin/main.lua
+++ b/plugins/terminal.koplugin/main.lua
@@ -350,7 +350,7 @@ function Terminal:terminal()
                           fullscreen = true,
                           condensed = true,
                           allow_newline = false,
-                          cursor_at_end = false,
+                          cursor_at_end = true,
                           buttons = {{{
                                   text = _("Cancel"),
                                   callback = function()
--- a/plugins/terminal.koplugin/main.lua
+++ b/plugins/terminal.koplugin/main.lua
@@ -1,7 +1,11 @@
+local ButtonDialog = require("ui/widget/buttondialog")
+local CenterContainer = require("ui/widget/container/centercontainer")
 local DataStorage = require("datastorage")
 local Font = require("ui/font")
 local InfoMessage = require("ui/widget/infomessage")
 local InputDialog = require("ui/widget/inputdialog")
+local LuaSettings = require("luasettings")
+local Menu = require("ui/widget/menu")
 local TextViewer = require("ui/widget/textviewer")
 local Trapper = require("ui/trapper")
 local UIManager = require("ui/uimanager")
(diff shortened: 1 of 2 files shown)


smartscripts-nl:
Reader_2020-08-31_131714
Reader_2020-08-31_131722
Reader_2020-08-31_131732
Reader_2020-08-31_131746
Reader_2020-08-31_131817

poire-z:
Seems to work fine enough.

This panel is a bit ugly with its 2 titles:
<kbd>image</kbd>

The shortcuts Menu is a bit small font and hard to read. Don't like having it full screen with ?
```
        width = Screen:getWidth(),
        height = Screen:getHeight(),
        covers_fullscreen = true, -- hint for UIManager:_repaint()
        is_borderless = true,
        is_popout = false,
```

Frenzie:
Some more background in #6607.

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,378 @@
...
+local LuaSettings = require("luasettings")
+local Menu = require("ui/widget/menu")
 local TextViewer = require("ui/widget/textviewer")
 local Trapper = require("ui/trapper")
 local UIManager = require("ui/uimanager")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
 local logger = require("logger")
 local util = require("ffi/util")
 local _ = require("gettext")
+local lfs = require("libs/libkoreader-lfs")
```
Frenzie, line 16: As I said, I think this would be per our style guide? ;-)
```suggestion
local lfs = require("libs/libkoreader-lfs")
local logger = require("logger")
local util = require("ffi/util")
local _ = require("gettext")
```

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,378 @@
...
 local Screen = require("device").screen
 
-local Terminal = WidgetContainer:new{
+local Terminal = WidgetContainer:new {
```
Frenzie, line 19: No need to change that.
```suggestion
local Terminal = WidgetContainer:new{
```

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,378 @@
...
     name = "terminal",
     dump_file = util.realpath(DataStorage:getDataDir()) .. "/terminal_output.txt",
     command = "",
+    items_per_page = G_reader_settings:readSetting("items_per_page") or 16,
+    source = 'terminal',
```
Frenzie, line 24: Please stick with our preferred double quotes unless there's a clear reason not to. ^_^
```suggestion
    source = "terminal",
```

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,378 @@
...
 }
 
 function Terminal:init()
     self.ui.menu:registerToMainMenu(self)
+    self:createSettingsFileIfNotExists()
+    local file_handle = LuaSettings:open(self.shortcuts_file)
+    self.shortcuts = file_handle:readSetting("shortcuts") or {}
+end
+
+function Terminal:createSettingsFileIfNotExists()
```
Frenzie, line 38: This function seems either redundant or a bug in LuaSettings. Just doing a quick glance atm so I don't know which, although redundant seems more likely after years of using LuaSettings. ;-)

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,378 @@
...
+        local file = io.open(self.shortcuts_file, "w")
+        local content = "-- we can read Lua syntax here!\nreturn {}\n"
+        if file then
+            file:write(content)
+            file:close()
+        end
+    end
+end
+
+function Terminal:saveShortcuts()
```
Frenzie, line 49: Same here, this seems to be duplicating things?

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,394 @@
...
+        self:insertPageActions(item_table)
+    end
+    local title = #self.shortcuts == 1 and _("Shortcut") or _("Shortcuts")
+    self.shortcuts_menu:switchItemTable(tostring(#self.shortcuts) .. " " .. title, item_table)
+    UIManager:show(self.shortcuts_dialog)
+end
+
+function Terminal:insertPageActions(item_table)
+    table.insert(item_table, {
+        text = "   " .. string.lower(_("to terminal")) .. "...",
```
poire-z, line 123: Not sure you need to lower() them (I guess I wouldn't mind "To terminal", even without the indentation) - and you don't know how that will work with translations.
<kbd>image</kbd>.
And make the "..." a proper ellipsis (single char), and keep it part of the translated string (we don't know how some other languages might prefer that)

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,394 @@
...
+        timeout = 2
+    })
+    self:saveShortcuts()
+    self:manageShortcuts()
+end
+
+function Terminal:editCommands(item)
+    local edit_dialog
+    edit_dialog = InputDialog:new {
+        title = _("Edit commands for") .. " \"" .. item.text .. "\"",
```
poire-z, line 219: `T(_('Edit commands for "%1"'), item.text)`

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,394 @@
...
 function Terminal:init()
     self.ui.menu:registerToMainMenu(self)
+    self.shortcuts = self.settings:readSetting("shortcuts") or {}
+end
+
+function Terminal:saveShortcuts()
+    self.settings:saveSetting("shortcuts", self.shortcuts)
+    self.settings:flush()
+    UIManager:show(InfoMessage:new {
+        text = _("Shortcuts saved!"),
```
poire-z, line 40: As you must have noticed, we're some rather boring and not very enthusiastic people - and we made a point to not have any enthusiastic message. So, no `!` :) Just a plain `Shorcuts saved.` (Really, we don't have any, except one in the goodreads plugin: "Search not found!")

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,394 @@
...
+end
+
+function Terminal:copyCommands(item)
+    local new_item = {
+        text = item.text .. " (copy)",
+        commands = item.commands
+    }
+    table.insert(self.shortcuts, new_item)
+    UIManager:show(InfoMessage:new {
+        text = _("Shortcut copied!"),
```
poire-z, line 209: Ditto! (oups)

poire-z (commented):
Still works for me.

plugins/terminal.koplugin/main.lua
```diff
@@ -1,58 +1,394 @@
+local ButtonDialog = require("ui/widget/buttondialog")
+local CenterContainer = require("ui/widget/container/centercontainer")
 local DataStorage = require("datastorage")
 local Font = require("ui/font")
 local InfoMessage = require("ui/widget/infomessage")
 local InputDialog = require("ui/widget/inputdialog")
+local LuaSettings = require("luasettings")
+local Menu = require("ui/widget/menu")
+local Screen = require("device").screen
```
Frenzie, line 9: Not this one. ;-) (In fact the reason I commented has disappeared completely?)

plugins/terminal.koplugin/main.lua
```diff
@@ -69,29 +405,61 @@ function Terminal:execute()
...
                         UIManager:close(viewer)
+                        self:terminal()
                     end,
                 },
+                close_button,
             },
-        },
+        }
+    end
+    viewer = TextViewer:new {
```
Frenzie, line 457:
```suggestion
    viewer = TextViewer:new{
```

plugins/terminal.koplugin/main.lua
smartscripts-nl: > Not sure you need to lower() them (I guess I wouldn't mind "To terminal", even without the indentation) - and you don't know how that will work with translations.
> image.
> And make the "..." a proper ellipsis (single char), and keep it part of the translated string (we don't know how some other languages might prefer that)

I like the lower case and indentation better, for clearer separation from regular shortcuts (as you might have guessed).

Frenzie: What's the string.lower about when you're already giving it lowercase characters? It'll just mess up a poor little language like German. ;-)

smartscripts-nl:
The conversation here has apparently stopped and the PR hasn't been merged. Have I overlooked some necessary steps before that can be done?

Frenzie:
Sometimes it'll just take a few days, especially if you do so many somewhat overlapping ones at once. ;-)

poire-z:
You continued adding some commits after the conversation had stopped. So, we couldn't know you were done.
So, I guess you're telling us you are done ? :)

plugins/terminal.koplugin/main.lua
```diff
@@ -69,29 +414,61 @@ function Terminal:execute()
...
+                        self:terminal()
                     end,
                 },
+                close_button,
             },
-        },
+        }
+    end
+    viewer = TextViewer:new{
+        title = _("Command output"),
```
Frenzie, line 467: E.g., I recall we were still discussing changing something like "output" for "results," not in #6598 either, and I don't recall if we even resolved that.

poire-z:
I don't want to look much into the code, but it is working allright.
Minor issue: when back to the command input, the cursor is at the beginning. Would be more convenient if it were at the end.

plugins/terminal.koplugin/main.lua
smartscripts-nl: I thought someone else would suggest another term, so I've done nothing with this.

smartscripts-nl:
> You continued adding some commits after the conversation had stopped. So, we couldn't know you were done.
> So, I guess you're telling us you are done ? :)

Indeed, that was my line of thinking... 😄

smartscripts-nl:
> I don't want to look much into the code, but it is working allright.
> Minor issue: when back to the command input, the cursor is at the beginning. Would be more convenient if it were at the end.

I'll look into this.

plugins/terminal.koplugin/main.lua
Frenzie: Is this the exact same phrase as used elsewhere?

Frenzie: And where was the discussion?

Frenzie: Oh, never mind. This isn't new, just moved.

plugins/terminal.koplugin/main.lua
Frenzie: @smartscripts-nl You never addressed this one.

plugins/terminal.koplugin/main.lua
```diff
@@ -1,57 +1,402 @@
...
+                          local new_shortcut = {
+                              text = name,
+                              commands = input,
+                          }
+                          table.insert(self.shortcuts, new_shortcut)
+                          self:saveShortcuts()
+                      end
+
+                      local prompt
+                      prompt = InputDialog:new{
```
Frenzie, line 346: You don't have to separate all these out.

plugins/terminal.koplugin/main.lua
```diff
@@ -1,57 +1,402 @@
...
+                  end,
+              }}},
+    }
+    UIManager:show(edit_dialog)
+    edit_dialog:onShowKeyboard()
+end
+
+function Terminal:editName(item)
+    local edit_dialog
+    edit_dialog = InputDialog:new{
```
Frenzie, line 255:
```suggestion
    local edit_dialog = InputDialog:new{
```

smartscripts-nl: I've done this on purpose. In the past I used the construction you suggest, but then UIManager:close(edit_dialog) in the callback functions didn't work. Or am I wrong?

smartscripts-nl:
> I don't want to look much into the code, but it is working allright.
> Minor issue: when back to the command input, the cursor is at the beginning. Would be more convenient if it were at the end.

I don't know why the cursor is not at the end: in Terminal:terminal() I have used cursor_at_end = true, but still the cursor is placed at the start of the field.

Does somebody have a clue why this happens?

plugins/terminal.koplugin/main.lua
poire-z: You're not wrong :) you actually have to do it that way when you reference it inside some included callbacks - which is the case here. (But try to not do it when it's not needed, it's kinda a clue about it happening or not.)

poire-z:
> Does somebody have a clue why this happens?

OK, it seems we cancel `cursor_at_end` when fullscreen : 
https://github.com/koreader/koreader/blob/f15aa7103f5e3a0225c8b65146eccc597e1d0dc5/frontend/ui/widget/inputdialog.lua#L342-L352
I'm not sure why I did that but well, I must have had a reason I don't really remember :/
May be it should be done only `if self.text_height > available_height`, dunno.
Well, ok, if that does not bother you, let's keep it that way.

plugins/terminal.koplugin/main.lua
```diff
@@ -1,57 +1,402 @@
...
+                })
+                -- actions:
+                self:insertPageActions(item_table)
+            end
+        end
+        -- no shortcuts defined yet:
+    else
+        self:insertPageActions(item_table)
+    end
+    local title = #self.shortcuts == 1 and _("Terminal shortcut") or _("Terminal shortcuts")
```
yparitcher, line 117: this should probably use http://koreader.rocks/doc/modules/gettext.html#ngettext
```suggestion
    local title = N(_("Terminal shortcut"), _("Terminal shortcuts"), #self.shortcuts)
```
with this at the top
local N_ = _.ngettext

plugins/terminal.koplugin/main.lua
```diff
@@ -1,57 +1,402 @@
...
+    table.insert(item_table, {
+        text = "   " .. _("to terminal…"),
+        deletable = false,
+        editable = false,
+        callback = function()
+            self:terminal()
+        end,
+    })
+    table.insert(item_table, {
+        text = "   " .. _("close…"),
```
yparitcher, line 132:
```suggestion
        text = _("   close…"),
```

plugins/terminal.koplugin/main.lua
```diff
@@ -1,57 +1,402 @@
...
+        self:insertPageActions(item_table)
+    end
+    local title = #self.shortcuts == 1 and _("Terminal shortcut") or _("Terminal shortcuts")
+    self.shortcuts_menu:switchItemTable(tostring(#self.shortcuts) .. " " .. title, item_table)
+    UIManager:show(self.shortcuts_dialog)
+end
+
+function Terminal:insertPageActions(item_table)
+    table.insert(item_table, {
+        text = "   " .. _("to terminal…"),
```
yparitcher, line 124:
```suggestion
        text = _("   to terminal…"),
```

plugins/terminal.koplugin/main.lua
Frenzie: That one should be okay with bidi and all that, plus it's actually nice not to bother the translators with how many spaces to add. :-)

Issue #6562, opened 2020-08-25
FR: A way to follow links in documents on non-touch devices


vovavili:
**Does your feature request involve difficulty completing a task? Please describe.**
My Kindle 4 NT has neither has a physical keyboard nor a touchscreen, which makes it impossible to follow links inside documents in any format in KOReader.

**Describe the solution you'd like**
Some shortcut which enables you to cycle through links on a page you're currently viewing - for example, in [CoolReader 3 for Kindle](https://github.com/CrazyCoder/coolreader-kindle-qt), all you had to do was press Keyboard+Up.

poire-z:
I don't know anything about how keys/pad are used on such non-touch devices, and what's still available or what would conflict or not, and how to trigger or go back.
But if one wants to go on with this, here's some patch to readerlink to allow cycling thru page links, and to jump to the selected one - on CreDocuments only (EPUB, FB2):
```diff
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -24,6 +24,18 @@ local ReaderLink = InputContainer:new{
 }

 function ReaderLink:init()
+    if Device:hasKeys() then
+        self.key_events.SelectNextPageLink = {
+            { {"L" } },
+            doc = "select next page link",
+            event = "SelectNextPageLink",
+        }
+        self.key_events.GotoSelectedPageLink = {
+            { { "G" } },
+            doc = "go to selected page link",
+            event = "GotoSelectedPageLink",
+        }
+    end
     if Device:isTouchDevice() then
         self.ui:registerTouchZones({
             {
@@ -937,6 +949,75 @@ function ReaderLink:onGoToInternalPageLink(ges)
     self:onGoToPageLink(ges, true)
 end

+function ReaderLink:onSelectNextPageLink()
+    if self.ui.document.info.has_pages then
+        return
+    end
+    local links = self.ui.document:getPageLinks()
+    if not links or #links == 0 then
+        return
+    end
+    if self.cur_selected_page_link_num then
+        self.cur_selected_page_link_num = self.cur_selected_page_link_num + 1
+        if self.cur_selected_page_link_num > #links then
+            self.cur_selected_page_link_num = 1
+        end
+    else
+        self.cur_selected_page_link_num = 1
+    end
+    local selected_link = links[self.cur_selected_page_link_num]
+    logger.dbg("selected_link", selected_link)
[...]

PR #6619, opened 2020-09-01
Allow following links and footnotes with keys

poire-z:
Only with CreDocuments (as no way currently to highlight links in PDFs).
<kbd>Tab</kbd> or <kbd>Shift</kbd>+<kbd>Tab</kbd> to select next or previous links (as in web browsers).
Press / <kbd>Enter</kbd> to follow (or show footnote in popup, and in there Press to follow), back to go back.

Probably not enough to allow closing #6562 , but things are there for people with such devices to select which keys to use instead of Tab.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6619)


Frenzie (approved):
I suspect that Tab/Shift+Tab aren't available on the Kindle 4 NT, but as a proof of concept lgtm.

poire-z:
@Goramur : I don't know what you've been doing with your koreader clone, but the way you proceed in keeping in sync has all our PRs added with the message you see just above:
`Goramur added a commit to Goramur/koreader that referenced this pull request 10 hours ago`
which is not really cool and adds noise on your side.
(Might be a github bug, or you testing/making some tools, dunno.)

Goramur:
> @Goramur : I don't know what you've been doing with your koreader clone, but the way you proceed in keeping in sync has all our PRs added with the message you see just above:
> `Goramur added a commit to Goramur/koreader that referenced this pull request 10 hours ago`
> which is not really cool and adds noise on your side.
> (Might be a github bug, or you testing/making some tools, dunno.)

@poire-z I am sorry. I deleted my fork now to hopefully remove the messages. I didn't know i would cause these messages by testing git on my fork.

Frenzie:
@poire-z Btw, if you want to see how convenient link navigation works install old Opera/Presto (7-12) or slightly less well done Vivaldi. You can spatially navigate with Shift+arrows, but in Vivaldi you have to enable it in the settings first.

The way it works in Links 2 is similar too.

poire-z:
> I deleted my fork now to hopefully remove the messages

Oh, thank you - that indeed removed all the messages! :)
(I thought such messages would get somehow engraved on our side - but it seems not :)

poire-z:
OK, I made this PR, but I hate since then that I assigned GoBackLink to the Back key: <kbd>Esc</kbd> on the emulator - and <kbd>Tab</kbd> in my local override. Now, everytime I quit the emulator with these keys (my habbit), I jumps back all the Skim/TOC/Bookmarks jump I have done, and lose the page I was on - and that I would like to find back when relaunching (often, a bug or something on this page that's relevant to my current dev).

I wanted to avoid that this way:
```diff
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
         }
         self.key_events.GoBackLink = {
             { "Back" },
             doc = "go back from link",
             event = "GoBackLink",
         }
+        if G_reader_settings:isFalse("enable_back_history") then
+            -- As in ReaderBack: have Back key quit immediately
+            -- (useful for some developers)
+            self.key_events.GoBackLink = nil
+        end
     end
     if Device:isTouchDevice() then
         self.ui:registerTouchZones({
             {
                 id = "tap_link",
```
Which would be ok for me, but uses a remotely related setting (used only in ReaderBack - offtopic: I think there might be interference between ReaderLink and ReaderBack - we don't specify any ordering to their handling of the Back key with "overrides=" - and depending on ReaderLink:onGoBackLink() returning true or not, it may or not be handled by ReaderBack.. They both manage their own location_stack. I have ReaderBack disable on the emulator and my devices, so I don't know nor care much about it).

Given that we use Back generically on the emulator to close stuff, I wonder if we'd better not register Back>GoBackLink - at least:
- when `Device:isTouchDevice()` and there is another way to go back from link ?
- or just when `Device:isEmulator()` ? 
- or via an hidden setting just for me and other devs in the know?

Any thoughts?

NiLuJe:
No personal experience with this myself, just chiming in that Back is an important key for non  touch devices (namely, the K4), because it's one of the few physically available keys ;).

poire-z:
> that Back is an important key for non touch devices (namely, the K4), 

Is it used for exiting KOReader too?
How would that work on such a device: open a book, use TOC to jump to 4th chapter, read a few pages - and then, would/could one use Back to leave the book/KOReader ?
(I guess the Skim dialog is not available ? As it also add the initial location to the location stack.)
(And if Navigation>Enable back history is checked, any page forward/backwad also add the previous page to another location stack, that would be dequeued by Back...)

NiLuJe:
> 
> 
> > that Back is an important key for non touch devices (namely, the K4),
> 
> Is it used for exiting KOReader too?

Yes. Not quite sure everything works all that well, though. The defaults say "Back to exit (prompt)" and "Back in FM: Go to parent folder", but when hit Back in the FM, it exits with no prompt, and when I hit Back in a book, it goes to the book's folder.

> How would that work on such a device: open a book, use TOC to jump to 4th chapter, read a few pages - and then, would/could one use Back to leave the book/KOReader ?

Back history is enabled, so, the first Back goes back to the initial position (i.e., where you opened the ToC from), and cycles through the history, then finally goes back to the FM when it's exhausted.

> (I guess the Skim dialog is not available ? As it also add the initial location to the location stack.)

It's available, but unusable without a touch screen. You can only close it with... the Back button ;).

> (And if Navigation>Enable back history is checked, any page forward/backwad also add the previous page to another location stack, that would be dequeued by Back...)

Yeah, it does that indeed, as I mentioned earlier.

poire-z:
> Yes. Not quite sure everything works all that well, though.

Yes, everything is a bit of a mess (The back to exit without prompt is because some random pairs/ordering when iterating thru the handlers of Back - so, sometimes it work, sometimes not...)

I'm trying to make all that a bit better, by adding a `Back in reader >` that will propose:
<kbd>image</kbd>
Spent already a few hours on this, everything works fine, except the "Go to previous read page" (= the former "Enable back history") - which is (and was already) a mess with PDF and its various view states (scroll, offset position in page) and many "no change" when "Back" or crashes when changing stuff...

Issue #6597, opened 2020-08-29
Pocketbook OS integration


ezdiy:
Currently, pocketbook OS integration is shoddy at best, large part of it being bitrotten. This issue is to track tasks that need to be done to fix this.

- [x] **Open on power-on** book in `/mnt/ext1/system/state/current`.
- [x] **Poweroff decal** screensaver (book cover, or whatever) burned to screen right before powering off is in `/mnt/ext1/system/logo/bookcover`.
- [x] **Poweron decal** screencap koreader just before poweroff into `/mnt/ext1/system/logo/bootlogo/`
- [x] **Retain screen on de-focus**  Quite likely needed for the two above. Currently our screen blanks when OS switches focus away from us, because OS assumes we use inkview provided back-buffer.
- [x] **Switch book** - when koreader is in background, and new book opened from explorer = discover the background instance, and tell it to open the book. While this could be done through [event pipe in the past](https://github.com/koreader/koreader/pull/6559#discussion_r479684398), that one is gone for good, so we need some new way to ping the task. Just SIGUSR1 + filename in /tmp might do?
- [ ] **History** - it's possible to keep the explorer synchronized with us by telling it books we're opening via `AddLastOpen()`. Bonus: Make it stop indexing random crap.
- [x] **Exit options** - restart koreader is borked. Others like poweroff device also have OS API.
- [x] **File associations** - instead of shipping extensions.cfg, which depends on fw version, parse system extensions (or user extensions if exist), and append koreader.app to the begin/end of supported apps for each supported file.
- [x] **Wifi annoyances** The way OS handles wifi is supremely annoying, so come up with something so as to not have to babysit connect/disconnect states all the time.

ref #6321

pazos:
another one:

- [ ] **File associations** - instead of shipping `extensions.cfg`, which depends on fw version, parse system extensions (or user extensions if exist), and append `koreader.app` to the begin/end of supported apps for each supported file.

Probably worth the time to make part of the code platform agnostic and repurpose it for other "portable" environments, like AppImage

pazos:
> Poweroff decal screensaver burned to screen right before powering off is in /mnt/ext1/system/logo/bookcover.

I remember reading something labeled `???` in inkview documentation :smile:, good to know the file :+1: 

IIRC we need to write a png file when EVT_HIDE is received but we have some time before shutdown. Please correct me if I'm wrong :p

ezdiy:
[...]

PR #6620, opened 2020-09-01
Pocketbook: Launcher OS integration

ezdiy:
* Bring in restart koreader & shutdown device exit options
* Selecting books in stock explorer reuses running instance

ref #6597

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6620)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -29,8 +29,7 @@ local PocketBook = Generic:new{
     hasKeys = yes,
     hasFrontlight = yes,
     canSuspend = no,
-    hasExitOptions = no,
-    canRestart = no,
+    canPowerOff = yes,
     needsScreenRefreshAfterResume = no,
     home_dir = "/mnt/ext1",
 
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Seems fine.

frontend/device/pocketbook/device.lua
```diff
@@ -29,8 +29,9 @@ local PocketBook = Generic:new{
     hasKeys = yes,
     hasFrontlight = yes,
     canSuspend = no,
-    hasExitOptions = no,
-    canRestart = no,
+    hasExitOptions = yes,
+    canRestart = yes,
```
pazos, line 33: hasExitOptions and canRestart return true on generic device, so there's no need to override them here (it shouldn't hurt either)

ezdiy:
Force because of:

```
+++ b/platform/pocketbook/koreader.app
@@ -55,6 +55,7 @@ fi
 export KO_EXIT_CODE="/tmp/.koreader.exit"
 RETURN_VALUE="85"
 while [ "${RETURN_VALUE}" = "85" ]; do
+    rm -f "${KO_EXIT_CODE}"
     ./reader.lua "${args}" >>crash.log 2>&1
     RETURN_VALUE=$(cat ${KO_EXIT_CODE})
 done
```

Exitcode got misinterpreted when koreader crashed (and didn't overwrite the previous file).

pazos:
Closes #6321

Issue #6415, opened 2020-07-22
Integration in pocketbook menu


vimalaguti:
Hi, with the latest pocketbook update there is a new menu. When you hold press a book from the pocketbook library, you can now select an "open with" submenu, from where you can choose pbreader 2 or reader 3. 
Would be nice if koreader can be added in this list too.
However I'm not sure where to look for, guess there is a documentation for the pocketbook SDK but I couldn't find it

roshavagarga:
@vimalaguti Which pocketbook? Which firmware?

Galunid:
This is a polish blog I came across that described something like this. It may be a bit outdated, so you can only **try this at your own risk**: http://cyfranek.booklikes.com/post/1862057/koreader-dla-czytnikow-pocketbook-rozszerza-mozliwosci
I assume you already have the firmware for your device downloaded and extracted:
> - copy contents of "applications" directory to "applications" directory on the ereader

> copy `koreader.app` file from "system/bin" directory to ereader's "system/bin" directory

> edit `extensions.cfg` file on ereader(in `system/config`). Change `epub:@EPUB_file:1:eink-reader.app:ICON_EPUB` to `epub:@EPUB_file:1:eink-reader.app,koreader.app:ICON_EPUB`; - important note: it's worth making a backup of this file, to edit it use Notepad++, or similar software(not MS Word/Libre office/etc.)

> save changed `extensions.cfg` file

> disconnect ereader

**Please remember, that I never owned Pocketbook nor do I plan to. Official installation guide is here:** https://github.com/koreader/koreader/wiki/Installation-on-PocketBook-devices

LucyOne:
Galunid's plan should work. Any external reading software like KoReader and CoolReader needs to be listed in "extensionx.cfg". If the file doesn't exist, you can create it with Notepad++.

vimalaguti:
That works!
After the april Pocketbook update, I've found (in handlers.cfg) the new names of the default reader apps:
- "eink-reader_with_blink.app" which I suppose is the new pbreader3
- "eink-reader_with_epub2.app"

Because I don't know how to find the names of all the default apps, I've just added the epub line to "extensions.cfg":
epub:@EPUB_file:1:koreader.app,eink-reader_with_blink.app,eink-reader_with_epub2.app:ICON_EPUB

Now there is just a new "koreader.app" line in the menu and all 3 choices works fine. I think this info can be added to the wiki

pazos:
Duplicate of #3544

PR #6624, opened 2020-09-02
Pocketbook: Add file associations into "Open with" menu

ezdiy:
We used to put koreader alone in there, but apparently other
handlers can be left in as optional - see #6415 

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6624)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -206,7 +206,7 @@ function PocketBook:associateFileExtensions(assoc)
         local t = info[k]
         if t then
             -- A system entry exists, so just change app, and reuse the rest
-            t[4] = app_name
+            t[4] = app_name .. "," .. t[4]
         else
             -- Doesn't exist, so hallucinate up something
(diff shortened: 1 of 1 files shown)


ezdiy:
Also, glad to see you back, CI.

Frenzie:
Did you do anything (e.g., sign up for CircleCI)?

Frenzie (approved):

ezdiy:
@Frenzie Yep, unsigned and re-signed (the original issue arose when i first signed up and *tried* to set up CI for my fork, which didn't do well as I have no idea about what to put into 'context' to make it work). That said, I'll just stick with draft PRs to check CI for now.

Frenzie:
Nothing's necessary in context. There are a couple of tokens to do some stuff after the build has already completed, concretely to push docs & translations. If desired I'm sure we could add a check to make that show a nice green checkmark even on a fork's master (i.e., only do that on koreader/koreader) but since you're normally working on a branch that shouldn't matter.

Or are you saying that CircleCI doesn't want to run if you don't enter some nonsense?

PR #6625, opened 2020-09-02
Pocketbook: Add reboot option.

ezdiy:
Turns out there's an IPC command for it.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6625)


Frenzie (commented):
Nice, really attacking that bitrot & never implemented functionality. 👍

PR #6627, opened 2020-09-02
[wip]: Turn the bookmark detail dialog into a bookmark navigator, with search button

smartscripts-nl:
With this changes you can easily navigate between bookmarks, without the need to jump to them in the text. When search term entered, easily navigate between bookmarks with hits for that term (then arrow buttons available).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6627)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -452,7 +452,7 @@ function ReaderBookmark:onShowBookmark()
                     end,
                     save_button_text = _("Search")
                 }
-                self:prompt(arguments)
+                bookmark:prompt(arguments)
             end,
         })
         if bookmark.search_value ~= "" then
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -69,7 +69,7 @@ function ReaderBookmark:search(query, bookmarks, current_bookmark, bm_menu, view
     local found_index = 0
     local start
     local use_second_loop = true
-    if direction == "forward" then
+    if direction == 1 then
         start = current_bookmark + 1
         if start > #bookmarks then
             start = 1
@@ -91,7 +91,7 @@ function ReaderBookmark:search(query, bookmarks, current_bookmark, bm_menu, view
                 end
             end
         end
-    -- direction == "previous":
+    -- direction == -1:
     else
         start = current_bookmark - 1
         if start < 1 then
@@ -119,9 +119,8 @@ function ReaderBookmark:search(query, bookmarks, current_bookmark, bm_menu, view
     if found_index > 0 then
         UIManager:close(viewer_instance)
         bm_menu:onMenuHold(bookmarks[found_index])
-        UIManager:show(InfoMessage:new { text = _("Found"), timeout = 1 })
     else
-        UIManager:show(InfoMessage:new { text = T(_('No bookmarks found with query "%1"'), query), timeout = 2 })
+        UIManager:show(InfoMessage:new { text = T(_('No bookmarks found with query "%1"'), query) })
     end
 end
 
@@ -392,20 +391,52 @@ function ReaderBookmark:onShowBookmark()
         end
         local first_button_row = {
             {
-                text = _("Go to"),
+                text = _("Remove"),
                 callback = function()
-                    UIManager:close(self.textviewer)
-                    UIManager:close(bookmark.bookmark_menu)
-                    bookmark.ui.link:addCurrentLocationToStack()
-                    bookmark:gotoBookmark(item.page)
+                    UIManager:show(ConfirmBox:new {
+                        text = _("Do you want remove this bookmark?"),
+                        cancel_text = _("Cancel"),
+                        cancel_callback = function()
+                            return
+                        end,
+                        ok_text = _("Remove"),
+                        ok_callback = function()
+                            local org_item = bookmark.bookmarks[item.index]
+                            bookmark:removeHighlight(org_item)
+                            UIManager:close(self.textviewer)
+                            if #bookmark.bookmarks > 1 then
+                                bm_menu:switchItemTable(nil, bookmark.bookmarks, -1)
+                                -- if no bookmarks left after removal, return to reader:
+                            else
+                                return false
+                            end
+                        end,
+                    })
                 end,
-            }
+            },
+            {
+                text = _("Rename"),
+                callback = function()
+                    local org_item = bookmark.bookmarks[item.index]
+                    bookmark:renameBookmark(org_item)
+                    UIManager:close(self.textviewer)
+                end,
+            },
         }
+        local from_start_text = "▕◁"
+        local backward_text = "◁"
+        local forward_text = "▷"
+        local from_end_text = "▷▏"
+        if BD.mirroredUILayout() then
+            backward_text, forward_text = forward_text, backward_text
+            -- Keep the LTR order of |< and >|:
+            from_start_text, from_end_text = BD.ltr(from_end_text), BD.ltr(from_start_text)
+        end
         if bookmark.search_value ~= "" then
             table.insert(first_button_row, {
-                text = "<",
+                text = backward_text,
                 callback = function()
-                    bookmark:search(bookmark.search_value, bookmarks, current_bookmark, bm_menu, self.textviewer, "previous")
+                    bookmark:search(bookmark.search_value, bookmarks, current_bookmark, bm_menu, self.textviewer, -1)
                 end,
             })
         end
@@ -413,11 +444,11 @@ function ReaderBookmark:onShowBookmark()
             text = _("Search"),
             callback = function()
                 local arguments = {
-                    title = _("Search bookmark"),
+                    title = _("Search in bookmarks"),
                     value = bookmark.search_value,
                     hint = _("Needle"),
                     callback = function(query)
-                        bookmark:search(string.lower(query), bookmarks, current_bookmark, bm_menu, self.textviewer, "forward")
+                        bookmark:search(string.lower(query), bookmarks, current_bookmark, bm_menu, self.textviewer, 1)
                     end,
                     save_button_text = _("Search")
                 }
@@ -426,54 +457,33 @@ function ReaderBookmark:onShowBookmark()
         })
         if bookmark.search_value ~= "" then
             table.insert(first_button_row, {
-                text = ">",
+                text = forward_text,
                 callback = function()
-                    bookmark:search(bookmark.search_value, bookmarks, current_bookmark, bm_menu, self.textviewer, "forward")
+                    bookmark:search(bookmark.search_value, bookmarks, current_bookmark, bm_menu, self.textviewer, 1)
                 end,
             })
         end
         table.insert(first_button_row, {
-            text = _("Rename"),
+            text = _("Go to"),
             callback = function()
-                local org_item = bookmark.bookmarks[item.index]
-                bookmark:renameBookmark(org_item)
                 UIManager:close(self.textviewer)
-            end,
-        })
-        table.insert(first_button_row, {
-            text = _("Remove"),
-            callback = function()
-                UIManager:show(ConfirmBox:new {
-                    text = _("Do you want remove this bookmark?"),
-                    cancel_text = _("Cancel"),
-                    cancel_callback = function()
-                        return
-                    end,
-                    ok_text = _("Remove"),
-                    ok_callback = function()
-                        local org_item = bookmark.bookmarks[item.index]
-                        bookmark:removeHighlight(org_item)
-                        UIManager:close(self.textviewer)
-                        if #bookmark.bookmarks > 1 then
-                            bm_menu:switchItemTable(nil, bookmark.bookmarks, -1)
-                            -- if no bookmarks left after removal, return to reader:
-                        else
-                            return false
-                        end
-                    end,
-                })
+                UIManager:close(bookmark.bookmark_menu)
+                bookmark.ui.link:addCurrentLocationToStack()
+                bookmark:gotoBookmark(item.page)
             end,
         })
         local second_button_row = {
             {
-                text = "<<",
+                text = from_start_text,
+                enabled = #bookmark.bookmarks > 1,
                 callback = function()
                     UIManager:close(self.textviewer)
                     bm_menu:onMenuHold(bookmarks[1])
                 end,
             },
             {
-                text = "<",
+                text = backward_text,
+                enabled = #bookmark.bookmarks > 1,
                 callback = function()
                     UIManager:close(self.textviewer)
                     local prev = current_bookmark - 1
@@ -493,7 +503,8 @@ function ReaderBookmark:onShowBookmark()
                 end,
             },
             {
-                text = ">",
+                text = forward_text,
+                enabled = #bookmark.bookmarks > 1,
                 callback = function()
                     UIManager:close(self.textviewer)
                     local next = current_bookmark + 1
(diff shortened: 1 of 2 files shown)


smartscripts-nl:
Reader_2020-09-03_012710
Reader_2020-09-03_012728
Reader_2020-09-03_012744
Reader_2020-09-03_012754

poire-z:
Too bad you passed the english strings thru some obfuscator or broken random word generator before you made the screenshots :)

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -362,6 +556,51 @@ function ReaderBookmark:onShowBookmark()
     return true
 end
 
+-- argument is a table, containing: value, hint, callback, cancel_callback
+function ReaderBookmark:prompt(args)
+    local value = args.value
+    local description = args.description
+    local callback = args.callback
+    local cancel_callback = args.cancel_callback or nil
+    local title = args.title or 'Bewerk'
```
poire-z, line 565: Untranslatable random word.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -362,6 +556,51 @@ function ReaderBookmark:onShowBookmark()
...
+        condensed = true,
+        allow_newline = false,
+        cursor_at_end = true,
+        buttons = {
+            {
+                {
+                    text = _("Cancel"),
+                    callback = function()
+                        UIManager:close(prompt_dialog)
+                        if cancel_callback ~= nil then
```
poire-z, line 583: `if cancel_callback then` is usually just fine.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -61,6 +63,68 @@ function ReaderBookmark:addToMainMenu(menu_items)
...
+                    break
+                end
+            end
+        end
+    end
+
+    if found_index > 0 then
+        UIManager:close(viewer_instance)
+        bm_menu:onMenuHold(bookmarks[found_index])
+        UIManager:show(InfoMessage:new { text = _("Found"), timeout = 1 })
```
poire-z, line 122: If found_index, you'll be showing a window. So no need for this InfoMessage, as it's obvious something was found.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -61,6 +63,68 @@ function ReaderBookmark:addToMainMenu(menu_items)
...
+    else
+        UIManager:show(InfoMessage:new { text = T(_('No bookmarks found with query "%1"'), query), timeout = 2 })
```
poire-z, line 124: No timeout needed, let the user time to digest its disappointment, look at the query term he entered, etc...

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -300,48 +364,178 @@ function ReaderBookmark:onShowBookmark()
...
+                    UIManager:close(self.textviewer)
+                    UIManager:close(bookmark.bookmark_menu)
+                    bookmark.ui.link:addCurrentLocationToStack()
+                    bookmark:gotoBookmark(item.page)
+                end,
+            }
+        }
+        if bookmark.search_value ~= "" then
+            table.insert(first_button_row, {
+                text = "<",
```
poire-z, line 406: Please use (here and on the 2nd row) the same arrow symbols we use elsewhere, and make it BiDi aware (test the UI with arabic, `KO_RTL=yes ./kodev run`), see:
https://github.com/koreader/koreader/blob/4e281f00edaae361dc4a2325cd8e9a0e5622b9f7/frontend/apps/reader/modules/readersearch.lua#L147-L155

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -300,48 +364,178 @@ function ReaderBookmark:onShowBookmark()
...
+                callback = function()
+                    bookmark:search(bookmark.search_value, bookmarks, current_bookmark, bm_menu, self.textviewer, "previous")
+                end,
+            })
+        end
+        table.insert(first_button_row, {
+            text = _("Search"),
+            callback = function()
+                local arguments = {
+                    title = _("Search bookmark"),
```
poire-z, line 416: `Search bookmarks` plural ? `Search among/in bookmarks` ?

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -300,48 +364,178 @@ function ReaderBookmark:onShowBookmark()
...
+        table.insert(first_button_row, {
+            text = _("Rename"),
+            callback = function()
+                local org_item = bookmark.bookmarks[item.index]
+                bookmark:renameBookmark(org_item)
+                UIManager:close(self.textviewer)
+            end,
+        })
+        table.insert(first_button_row, {
+            text = _("Remove"),
```
poire-z, line 444: For the order in the first row, I'd rather see the most destructive/rarely used on the left:
`Remove | Rename | < | Search | > | Go to`

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -300,48 +364,178 @@ function ReaderBookmark:onShowBookmark()
...
+                    bookmark.ui.link:addCurrentLocationToStack()
+                    bookmark:gotoBookmark(item.page)
+                end,
+            }
+        }
+        if bookmark.search_value ~= "" then
+            table.insert(first_button_row, {
+                text = "<",
+                callback = function()
+                    bookmark:search(bookmark.search_value, bookmarks, current_bookmark, bm_menu, self.textviewer, "previous")
```
poire-z, line 408: We use `-1` or `1` instead of `previous` or `forward` in other places.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -300,48 +364,178 @@ function ReaderBookmark:onShowBookmark()
...
+            {
+                text = ">>",
+                callback = function()
+                    UIManager:close(self.textviewer)
+                    bm_menu:onMenuHold(bookmarks[#bookmarks])
+                end,
+            },
+        }
+        -- if there is only one bookmark, we need only to retain the close button:
+        if #bookmark.bookmarks == 1 then
```
poire-z, line 515: For consistency, better to have always the same buttons. Just may be make some of them disabled with `enabled = #bookmark.bookmarks > 1`

poire-z (commented):
(I'm off soon and won't really be able to review for a week.)

frontend/apps/reader/modules/readerbookmark.lua
smartscripts-nl: I had done this like so because you can't mark the found text with a highlight. But I am fine with removing this.

frontend/apps/reader/modules/readerbookmark.lua
smartscripts-nl: Would you say the same about the navigation buttons on the first row (for navigating found terms)?

Frenzie: Just a quick comment from my phone, so grain of salt & all (may not apply), but generally speaking consistency with grayed out is better for expectations/user training than forcing reanalysis.

smartscripts-nl:
Question: would it be possible to link the hardware buttons of my Kobo Forma to the goto previous and next bookmark buttons, instead of to the scrollbar in the textviewer? So that these buttons kind of have focus, instead of the textviewer?

frontend/apps/reader/modules/readerbookmark.lua
NiLuJe: Or `Bookmark search` ;).

Frenzie: +1 on bookmark search; this is a header, isn't it?

smartscripts-nl: Done

smartscripts-nl:
In my personal version I've added a third row (in the middle), with a button for randomly displaying bookmarks and two buttons for transferring the bookmark notes to two of my other plugins. But I suppose this is a bit too specific and not desirable for other users?

smartscripts-nl:
@Frenzie Is my last commit for improving indentation of bookmark notes OK?

Frenzie:
Could you add before/after screenshots? That way it's not just accessible to people who know how to handle Git. ^_^

smartscripts-nl:
> Could you add before/after screenshots? That way it's not just accessible to people who know how to handle Git. ^_^

OK, later today.

smartscripts-nl:
> > Could you add before/after screenshots? That way it's not just accessible to people who know how to handle Git. ^_^
> 
> OK, later today.



> Could you add before/after screenshots? That way it's not just accessible to people who know how to handle Git. ^_^

I specifically meant my local functions. Those probably should be put elsewhere?

smartscripts-nl:
> Could you add before/after screenshots? That way it's not just accessible to people who know how to handle Git. ^_^

Ignore the second row of buttons. They are in my personal version (translations for who want to know: Random, Quote, Idea).

Without indentation:
Reader_2020-09-04_135602

With indentation:
Reader_2020-09-04_135742

Frenzie:
Not necessarily. That substrCount() is fairly generic and should probably go into util if we feel that's worthwhile, although I think it might be a bit too simple/basic to make into a function. But those other ones seem fairly specific to their context.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -259,12 +332,13 @@ function ReaderBookmark:onShowBookmark()
             end
         end
         if v.text == nil or v.text == "" then
+            self.garbage = k
```
poire-z, line 335: This garbage is not used :)

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -292,56 +366,262 @@ function ReaderBookmark:onShowBookmark()
...
+        local ipage = item.page
+        if not bookmark.ui.document.info.has_pages then
+            page = bookmark.ui.document:getPageFromXPointer(ipage)
+        end
+        if page ~= "" then
+            title = title .. "  -  " .. _("page") .. " " .. page
+        end
+        -- show search term in title:
+        if bookmark.search_value and bookmark.search_value ~= "" then
+            title = title .. '  -  "' .. bookmark.search_value .. '"'
```
poire-z, line 405: These 2 title = might be more readable if using `T('%1 - "%2"')`. Dunno if these constructs need to be translatable.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -292,56 +366,262 @@ function ReaderBookmark:onShowBookmark()
...
+            if last_end <= #str then
+                cap = str:sub(last_end)
+                table.insert(t, cap)
+            end
+            return t
+        end
+
+        -- HELPER FUNCTIONS WHICH PROBABLY SHOULD BE PUT SOMEWHERE ELSE:
+
+        local function substrCount(subject, needle)
```
poire-z, line 564: At least, put them at the module level, outside of (above) `function ReaderBookmark:onShowBookmark()`, so they are not re-defined each time this function is called.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -292,56 +366,262 @@ function ReaderBookmark:onShowBookmark()
...
+
         self.textviewer = TextViewer:new{
-            title = _("Bookmark details"),
-            text = item.notes,
+            title = title,
+            -- add indentation for better readability:
+            text = indent(text),
+            -- make text easier to read:
+            justified = false,
+            text_face = Font:getFace("xx_smallinfofont"),
```
poire-z, line 619: It looks like you reduced the font size from x_smallinfofont to xx_smallinfofont.
May be it's nicer, may be it's not, I don't know.
May be it's just because _you_ usually highlight huge text segments - but people who highlight smaller segments will feel robed of their readability :)
May be you could re-use the font size a user has set for dictionary/wikipedia results (even if it's unrelated, it might be more the size a user expect to read such text with than your hardcoded smaller one):
https://github.com/koreader/koreader/blob/c28f1b3450b9db073a072255a247744b5afb20ed/frontend/ui/widget/dictquicklookup.lua#L75-L77

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -359,9 +639,62 @@ function ReaderBookmark:onShowBookmark()
     end
 
     UIManager:show(self.bookmark_menu)
+    if open_navigator then
+        if #self.bookmarks > 0 then
+            -- open first bookmark in page order:
+            bm_menu:onMenuHold(self.bookmarks[#self.bookmarks])
+        else
+            UIManager:show(InfoMessage:new { text = _('Bookmarks navigator: no bookmarks defined yet') })
```
poire-z, line 647: Have we named that thing `Bookmarks navigator` anywhere else?
If not, may be just `'Bookmarks: no bookmark defined yet'`.

poire-z:
> In my personal version I've added a third row (in the middle), with a button for randomly displaying bookmarks and two buttons for transferring the bookmark notes to two of my other plugins. But I suppose this is a bit too specific and not desirable for other users?

Yeah, that might make that dialog a bit too crowded :)

> Question: would it be possible to link the hardware buttons of my Kobo Forma to the goto previous and next bookmark buttons, instead of to the scrollbar in the textviewer? So that these buttons kind of have focus, instead of the textviewer?

You'd need to add some event handlers. And probably not _instead of_ , but _after_, so they work like Tap in dict lookup results: you scroll down first, and when you reach bottom, you go to next result/bookmark. (But I don't think keys work in dict lookup yet.) That was quite complicated to do for dict lookup though.

> Is my last commit for improving indentation of bookmark notes OK?

I don't really use/read bookmarks, so I don't know how not doing any text formatting was an issue.
But it feels your choices might be some heuristics that work for you, but might not always be right for some other text or other languages (like adding a line break after words in CAPS; indenting text, but not  if it is more than 3 lines smaller than 52 chars...)
I get the idea of making it more readable, but that might be presumptuous :) Showing the raw text is safer, and we won't be accused of messing it if it didn't work.
So, dunno. @Frenzie ?
(We could let it in, and if complains, we could add some menu toggle to have that additional formatting applied or not.)

Frenzie:
> So, dunno. @Frenzie ?

I think if you were going to do that, based on an HTML extract with `<p>` and `<em>`, etc. might be a lot less error prone. But the paragraph things probably looks reasonable enough.

frontend/apps/reader/modules/readerbookmark.lua
smartscripts-nl: That's right. I only use it to keep the code inspection of my IDE silent, complaining otherwise about an unused variable k. But I'll remove it.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -317,6 +317,76 @@ function ReaderBookmark:updateHighlightsIfNeeded()
     self.ui.doc_settings:saveSetting("bookmarks_version", 20200615)
 end
 
+local function split(str, pat)
```
poire-z, line 320: We already have a split function in util: `util.gsplit()`, a more versatile one, so no need to add another one here.
It's not strictly usable as you use this one to make a table, but as then, you iterate that table, you can use it. It's used in frontend/apps/reader/modules/readertypography.lua and frontend/ui/bidi.lua and frontend/device/kobo/device.lua if you want to see how to use it.

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -448,11 +448,11 @@ function ReaderBookmark:onShowBookmark(open_navigator)
             page = bookmark.ui.document:getPageFromXPointer(ipage)
         end
         if page ~= "" then
-            title = title .. "  -  " .. _("page") .. " " .. page
+            title = T("%1  -  %2 %3", title, _("page"), page)
```
poire-z, line 451: Probably not really good, for languages that would display it like "4'paged". Best to include the whole %1 %2 %3 in a _ template, and have T(_(template), value1, value2...).
with a comment above like:
`-- @translators %1 is the bookmark menu title with index. %2 is the page number`

Or `title .. " - " .. T(_("page %1"), page)` if you think like I do that no translator would want to re-order title and page.

Frenzie: @poire-z In America they say today's September 15, 2020.

poire-z:
Closing as many of these features have since been implemented.

PR #6629, opened 2020-09-03
Add bookmark count as footer item

smartscripts-nl:
Optionally display number of bookmarks in footer. If no bookmarks defined, don't show this item

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6629)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -50,7 +50,7 @@ local symbol_prefix = {
         -- @translators This is the footer letter prefix for battery % remaining.
         battery = C_("FooterLetterPrefix", "B:"),
         -- @translators This is the footer letter prefix for bookmarks count.
-        bookmarks_count = C_("FooterLetterPrefix", "Bk:"),
+        bookmarks_count = C_("FooterLetterPrefix", "BM:"),
         -- @translators This is the footer letter prefix for percentage read.
         percentage = C_("FooterLetterPrefix", "R:"),
         -- @translators This is the footer letter prefix for book time to read.
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -40,7 +40,7 @@ local MODE = {
     wifi_status = 10,
     book_title = 11,
     book_chapter = 12,
-    bookmarks_count = 13,
+    bookmark_count = 13,
 }
 
 local symbol_prefix = {
@@ -50,7 +50,7 @@ local symbol_prefix = {
         -- @translators This is the footer letter prefix for battery % remaining.
         battery = C_("FooterLetterPrefix", "B:"),
         -- @translators This is the footer letter prefix for the number of bookmarks (bookmark count).
-        bookmarks_count = C_("FooterLetterPrefix", "BM:"),
+        bookmark_count = C_("FooterLetterPrefix", "BM:"),
         -- @translators This is the footer letter prefix for percentage read.
         percentage = C_("FooterLetterPrefix", "R:"),
         -- @translators This is the footer letter prefix for book time to read.
@@ -68,7 +68,7 @@ local symbol_prefix = {
         time = "⌚",
         pages_left = BD.mirroredUILayout() and "⇐" or "⇒",
         battery = "",
-        bookmarks_count = "☆",
+        bookmark_count = "☆",
         percentage = BD.mirroredUILayout() and "⤟" or "⤠",
         book_time_to_read = "⏳",
         chapter_time_to_read = BD.mirroredUILayout() and "⥖" or "⤻",
@@ -160,18 +160,18 @@ local footerTextGeneratorMap = {
             return BD.wrap(prefix) .. " " .. (powerd:isCharging() and "+" or "") .. batt_lvl .. "%"
         end
     end,
-    bookmarks_count = function(footer)
+    bookmark_count = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
-        local prefix = symbol_prefix[symbol_type].bookmarks_count
-        --retrieve bookmarks count:
+        local prefix = symbol_prefix[symbol_type].bookmark_count
+        --retrieve bookmark count:
         local config = DocSettings:open(G_reader_settings:readSetting("lastfile"))
         local bookmarks = config:readSetting("bookmarks") or {}
-        local bookmarks_count = tostring(#bookmarks)
+        local bookmark_count = tostring(#bookmarks)
         -- if no bookmarks defined, don't show icon:
-        if bookmarks_count == "0" then
+        if bookmark_count == "0" then
             return ""
         end
-        return prefix .. bookmarks_count
+        return prefix .. bookmark_count
     end,
     time = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
@@ -318,7 +318,7 @@ function ReaderFooter:init()
         reclaim_height = false,
         toc_markers = true,
         battery = Device:hasBattery(),
-        bookmarks_count = true,
+        bookmark_count = true,
         time = true,
         page_progress = true,
         pages_left = true,
@@ -711,7 +711,7 @@ function ReaderFooter:textOptionTitles(option)
     local option_titles = {
         all_at_once = _("Show all at once"),
         reclaim_height = _("Reclaim bar height from bottom margin"),
-        bookmarks_count = T("Bookmark count (%1)", symbol_prefix[symbol].bookmarks_count),
+        bookmark_count = T("Bookmark count (%1)", symbol_prefix[symbol].bookmark_count),
         page_progress = T(_("Current page (%1)"), "/"),
         time = symbol_prefix[symbol].time
             and T(_("Current time (%1)"), symbol_prefix[symbol].time) or _("Current time"),
@@ -1623,7 +1623,7 @@ function ReaderFooter:addToMainMenu(menu_items)
     if Device:hasBattery() then
         table.insert(sub_items, getMinibarOption("battery"))
     end
-    table.insert(sub_items, getMinibarOption("bookmarks_count"))
+    table.insert(sub_items, getMinibarOption("bookmark_count"))
     table.insert(sub_items, getMinibarOption("percentage"))
     table.insert(sub_items, getMinibarOption("book_time_to_read"))
     table.insert(sub_items, getMinibarOption("chapter_time_to_read"))
@@ -1876,7 +1876,7 @@ function ReaderFooter:applyFooterMode(mode)
     -- 10 for Wi-Fi status
     -- 11 for book title
     -- 12 for current chapter
-    -- 13 for bookmarks count
+    -- 13 for bookmark count
 
     if mode ~= nil then self.mode = mode end
     local prev_visible_state = self.view.footer_visible
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -49,7 +49,7 @@ local symbol_prefix = {
         pages_left = "=>",
         -- @translators This is the footer letter prefix for battery % remaining.
         battery = C_("FooterLetterPrefix", "B:"),
-        -- @translators This is the footer letter prefix for bookmarks count.
+        -- @translators This is the footer letter prefix for the number of bookmarks (bookmark count).
         bookmarks_count = C_("FooterLetterPrefix", "BM:"),
         -- @translators This is the footer letter prefix for percentage read.
         percentage = C_("FooterLetterPrefix", "R:"),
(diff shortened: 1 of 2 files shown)


hius07:
I'd shorten as BM:, not Bk:
https://acronyms.thefreedictionary.com/bookmark

smartscripts-nl:
> I'd shorten as BM:, not Bk:
> https://acronyms.thefreedictionary.com/bookmark

Thanks! Done.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -47,6 +49,8 @@ local symbol_prefix = {
         pages_left = "=>",
         -- @translators This is the footer letter prefix for battery % remaining.
         battery = C_("FooterLetterPrefix", "B:"),
+        -- @translators This is the footer letter prefix for bookmarks count.
```
Frenzie, line 52: That should be "bookmark count" but I think "number of bookmarks" would be clearer.
```suggestion
        -- @translators This is the footer letter prefix for the number of bookmarks (bookmark count).
```

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -155,6 +160,19 @@ local footerTextGeneratorMap = {
...
+    bookmarks_count = function(footer)
+        local symbol_type = footer.settings.item_prefix or "icons"
+        local prefix = symbol_prefix[symbol_type].bookmarks_count
+        --retrieve bookmarks count:
+        local config = DocSettings:open(G_reader_settings:readSetting("lastfile"))
+        local bookmarks = config:readSetting("bookmarks") or {}
+        local bookmarks_count = tostring(#bookmarks)
+        -- if no bookmarks defined, don't show icon:
+        if bookmarks_count == '0' then
+            return ''
```
Frenzie, line 172:
```suggestion
            return ""
```

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -692,6 +711,7 @@ function ReaderFooter:textOptionTitles(option)
     local option_titles = {
         all_at_once = _("Show all at once"),
         reclaim_height = _("Reclaim bar height from bottom margin"),
+        bookmarks_count = T("Bookmarks count (%1)", symbol_prefix[symbol].bookmarks_count),
```
Frenzie, line 714:
```suggestion
        bookmarks_count = T("Bookmark count (%1)", symbol_prefix[symbol].bookmarks_count),
```

Frenzie (commented):
Seems fine besides a couple of nits.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -39,6 +40,7 @@ local MODE = {
     wifi_status = 10,
     book_title = 11,
     book_chapter = 12,
+    bookmarks_count = 13,
```
Frenzie, line 43: Please also change the variable name while there's no legacy yet. ^_^
```suggestion
    bookmark_count = 13,
```

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1855,6 +1876,7 @@ function ReaderFooter:applyFooterMode(mode)
     -- 10 for Wi-Fi status
     -- 11 for book title
     -- 12 for current chapter
+    -- 13 for bookmarks count
```
Frenzie, line 1879:
```suggestion
    -- 13 for bookmark count
```

smartscripts-nl:
The sidecar file is opened to determine the bookmark count every time the footer is refreshed. That seems costly to me. I will try to modify the code, so that the sidecar file is only opened when its modification time has changed.

Is my line of thinking correct? Or is this peanuts and not worth the trouble?

Frenzie:
I imagine you can just grab it with something like `self.ui.bookmark:getNumberOfHighlightsAndNotes()`.

smartscripts-nl:
> I imagine you can just grab it with something like `self.ui.bookmark:getNumberOfHighlightsAndNotes()`.

OK, I'll try that.

smartscripts-nl:
> I imagine you can just grab it with something like `self.ui.bookmark:getNumberOfHighlightsAndNotes()`.

But I now see that ReaderBookmark:getNumberOfHighlightsAndNotes() only counts highlighted notes. So it seems to me dogeared bookmarks are left out of the count.

Maybe I should introduce a new method in ReaderBookmark: getNumberOfBookmarks() ?

Frenzie:
That would fall under "something like." ;-)

smartscripts-nl:
> That would fall under "something like." ;-)

When I use this:

local bookmark_count = footer.ui.bookmark:getNumberOfBookmarks()

that method which I added in readerbookmark.lua in fact gets not called. What am I doing wrong?

Frenzie:
How did you check if it gets called? What's in the method?

smartscripts-nl:
> How did you check if it gets called? What's in the method?

No method needed, I realise now. This should work (I'll test it in a second):

`bookmark_count = function(footer)
        local prefix = symbol_prefix[symbol_type].bookmark_count
        --retrieve bookmark count:
        local bookmark_count = #footer.ui.bookmark.bookmarks or 0
        -- if no bookmarks defined, don't show icon:
        if not bookmark_count or bookmark_count == 0 then
            return ""
        end
        return prefix .. tostring(bookmark_count)
    end`

Frenzie:
> This should work (I'll test it in a second)

Yes, but that should be the content of the method.

Frenzie:
Think of it in terms of public/private, just not strictly enforced.

Frenzie:
`#footer.ui.bookmark.bookmarks or 0`

Btw, that doesn't seem to make much sense? If `footer.ui.bookmark.bookmarks` is nil that'd be an error.

Frenzie:
You want something like `return self.bookmarks and #self.bookmarks or 0`.

smartscripts-nl:
> `#footer.ui.bookmark.bookmarks or 0`
> 
> Btw, that doesn't seem to make much sense? If `footer.ui.bookmark.bookmarks` is nil that'd be an error.

Yes, but then I can return "" in the next lines of code.

But #footer.ui.bookmark.bookmarks also returns nothing. Strange, because in comparable functions footer.ui is being used. Don't know why footer.ui.bookmark can't be reached (or so it seems to me).

Frenzie:
Works as expected for me (diff with the current status of the PR).
```diff
diff --git a/frontend/apps/reader/modules/readerfooter.lua b/frontend/apps/reader/modules/readerfooter.lua
index 038716da..9dbc7865 100644
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -165,8 +165,9 @@ local footerTextGeneratorMap = {
         local prefix = symbol_prefix[symbol_type].bookmark_count
         --retrieve bookmark count:
         local config = DocSettings:open(G_reader_settings:readSetting("lastfile"))
-        local bookmarks = config:readSetting("bookmarks") or {}
-        local bookmark_count = tostring(#bookmarks)
+        --~ local bookmarks = config:readSetting("bookmarks") or {}
+        local bookmark_count = #footer.ui.bookmark.bookmarks
+        error(bookmark_count)
         -- if no bookmarks defined, don't show icon:
         if bookmark_count == "0" then
             return ""
```

Frenzie:
> Yes, but then I can return "" in the next lines of code.

And error as in a crash.

smartscripts-nl:
> ```diff
> local bookmark_count = #footer.ui.bookmark.bookmarks
> +        error(bookmark_count)
> ```

Then the explanation must be that my version of readerfooter.lua is too heavily modified and because of that something goes awry. I'll try substituting it with the standard version.

Frenzie:
In full:
```diff
diff --git a/frontend/apps/reader/modules/readerbookmark.lua b/frontend/apps/reader/modules/readerbookmark.lua
index 238bf0c1..b583549d 100644
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -681,6 +681,10 @@ function ReaderBookmark:hasBookmarks()
     return self.bookmarks and #self.bookmarks > 0
 end
 
+function ReaderBookmark:getNumberOfBookmarks()
+    return self.bookmarks and #self.bookmarks or 0
+end
+
 function ReaderBookmark:getNumberOfHighlightsAndNotes()
     local highlights = 0
     local notes = 0
diff --git a/frontend/apps/reader/modules/readerfooter.lua b/frontend/apps/reader/modules/readerfooter.lua
index 038716da..d65fb4e1 100644
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -164,9 +164,7 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].bookmark_count
         --retrieve bookmark count:
-        local config = DocSettings:open(G_reader_settings:readSetting("lastfile"))
-        local bookmarks = config:readSetting("bookmarks") or {}
-        local bookmark_count = tostring(#bookmarks)
+        local bookmark_count = footer.ui.bookmark:getNumberOfBookmarks()
         -- if no bookmarks defined, don't show icon:
         if bookmark_count == "0" then
             return ""
```

smartscripts-nl:
> ```diff
> local bookmark_count = footer.ui.bookmark:getNumberOfBookmarks()
> ```

Thanks! I can now confirm that your suggested adaptations work. So it must indeed have something to do with my adaptations. I'll push the updates.

smartscripts-nl:
> In full:

I now am under the impression that it doesn't work because of the code following below this comment:

````
-- NOTE: Getting the dimensions of the widget is impossible without having drawn it first
````

UI not refreshed c.q. available in all situations? I look into this some more.

**Comment by smartscripts-nl** [comment]
Hm, I've developed a bit of code blindness. The version I thought worked, still used the DocSettings module. So adapted code doesn't work for me. I'll look into this again tomorrow.

**Comment by smartscripts-nl** [comment]
Problem solved. There was no error in the code in footer.lua. I realised that in the footer settings I had in the past activated the item bookmarks_count, but not bookmark_count, of course. After activating the setting again in KOReader, the bookmark count was displayed again!

**Comment by smartscripts-nl** [comment]
> Problem solved.

Thanks to the new procedure the display of bookmarks is also more responsive. In the past the bookmark count in the footer was only updated after the bookmarks had been saved to the sidecar file. That lag is now out of the loop.

**Comment by smartscripts-nl** [comment]
> > Problem solved.
> 
> Thanks to the new procedure the display of bookmarks is more responsive and adequate. In the past the bookmark count in the footer was only updated after the bookmarks had been saved to the sidecar file. That lag is now out of the loop.

**Review by Frenzie — APPROVED** [review]
lgtm, any more comments @poire-z @NiLuJe ?

File: frontend/apps/reader/modules/readerfooter.lua [hunk 6]
```diff
@@ -155,6 +159,17 @@ local footerTextGeneratorMap = {
...
+    bookmark_count = function(footer)
+        local symbol_type = footer.settings.item_prefix or "icons"
+        local prefix = symbol_prefix[symbol_type].bookmark_count
+        --retrieve bookmark count:
+        local bookmark_count = footer.ui.bookmark:getNumberOfBookmarks()
+        -- if no bookmarks defined, don't show icon:
+        if bookmark_count == 0 then
+            return ""
+        end
+        return prefix .. tostring(bookmark_count)
```
yparitcher, line 171: You forgot a space between the prefix and number
```suggestion
        return prefix .. " " .. tostring(bookmark_count)
```

smartscripts-nl: No, I hadn't forgotten that, it was a choice of mine to make this item as compact as possible (my footer has many items, so every space extra comes at a cost). I actually like it better without the extra space.

If more people support this extra space, I am allright with adding it.

Frenzie: Of course I don't think this one specific item should randomly be different.

I will point out that there exist such things as thin spaces. It's not a matter of all or nothing. (In the space sense. It's all or nothing in the consistency sense.)

https://commons.wikimedia.org/wiki/File:Punctuation-Spaces.svg

smartscripts-nl: Thin spaces? How to accomplish in KOReader? Also with a svg? Seems unlikely to me.

I understand the argument for consistency, but still think the item looks much nicer without the space.

But as I said, then I'll insert it (and remove it in my personal version 😄 ).

Frenzie: lol no, you'd use it in the string: ` `

But not in this PR please. ;-)

smartscripts-nl: I don't understand: it _was_ the intention that I'd add the space in this PR? Then I've misunderstood. Where do you want me to add the extra space?

Frenzie: The space is good now, all's fine in this PR as far as I'm concerned.

But changing the space to a thin space or a hair space or something — that should be a different PR.

smartscripts-nl: Yes, I think it is best to leave things as they are now.

PR #6631, opened 2020-09-04
Add EyeSaver plugin

smartscripts-nl:
This plugin, if enabled, displays every 20 minutes a message that the user should look away from his screen for 20 seconds, to spare his eyes.

Currently doesn't work yet, because I have problems computing the next timestamp to display the message. These lines result in an error:
````
local now = util.getTimestamp()
self.display_timestamp = now + time
````

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6631)

### CODE REVIEW

**Comment by smartscripts-nl** [comment]
The error: util.getTimestamp not found. I_did_ load this class at the start of main.lua: local util = require("ffi/util").

**Comment by smartscripts-nl** [comment]
When I use this:

````
local now_t = os.time(os.date("*t"))
self.display_timestamp = now_t + time
````

the error I get is: attempt to index local 'time' (a number value).

I don't understand this error... 😊

**Comment by pazos** [comment]
Not against the plugin itself but shipping this enabled in all devices looks like a very bad idea to me.

I would like to improve integration with user plugins (translations, hooks for gestures, installation from zip files...) and leave this kind of plugins outside the main repo (like https://github.com/shezik/SuspendHack, for instance)

**Comment by smartscripts-nl** [comment]
> Not against the plugin itself but shipping this enabled in all devices looks like a very bad idea to me.
> 
> I would like to improve integration with user plugins (translations, hooks for gestures, installation from zip files...) and leave this kind of plugins outside the main repo (like https://github.com/shezik/SuspendHack, for instance)

That seems like a good idea.

PS: in my plugin the messages are disabled by default, you have to enable them from the menu to start showing them.

**Comment by smartscripts-nl** [comment]
> Not against the plugin itself but shipping this enabled in all devices looks like a very bad idea to me.
> 
> I would like to improve integration with user plugins (translations, hooks for gestures, installation from zip files...) and leave this kind of plugins outside the main repo (like https://github.com/shezik/SuspendHack, for instance)

It seems to me that in that setup there should be hooks available for plugin developers with which they can add their plugins to dispatcher and hooks with which they can integrate their plugins in specific locations in the menu (if applicable).

**Comment by Frenzie** [comment]
Tracked in #6575.

>  hooks with which they can integrate their plugins in specific locations in the menu

That already exists, add a `sorting_hint` to the menu entry.

**Comment by smartscripts-nl** [comment]
> Tracked in #6575.
> 
> > hooks with which they can integrate their plugins in specific locations in the menu
> 
> That already exists, add a `sorting_hint` to the menu entry.

Nice!

**Comment by Frenzie** [comment]
With regard to the plugin itself, there's already a timer plugin. From a UI perspective I think adding a recurring timer option to that one would add less noise.

Screenshot_20200904_150833

File: frontend/ui/elements/filemanager_menu_order.lua [hunk 1]
```diff
@@ -104,6 +104,7 @@ local order = {
         "news_downloader",
         "send2ebook",
         "text_editor",
+        "eyesaver",
```
poire-z, line 107: Obsolete (x2)

plugins/readtimer.koplugin/main.lua
```diff
@@ -50,6 +86,89 @@ function ReadTimer:unschedule()
...
+
+function ReadTimer:eyeSaverResetTimer()
+    self.eyesaver.display_timestamp = 0
+    self.eyesaver.display_time_clock = ''
+    self:eyeSaverSaveSettings()
+end
+
+function ReadTimer:eyeSaverSaveSettings()
+    self.eyesaver.settings:saveSetting("eyesaver_timestamp", self.eyesaver.display_timestamp)
+    self.eyesaver.settings:saveSetting("eyesaver_display_time_clock", self.eyesaver.display_time_clock)
```
poire-z, line 136: I don't get why this needs to be saved as a setting: it's just HH:MM representation of self.eyesaver.display_timestamp.

plugins/readtimer.koplugin/main.lua
```diff
@@ -50,6 +86,89 @@ function ReadTimer:unschedule()
...
+end
+
+function ReadTimer:eyeSaverResetTimer()
+    self.eyesaver.display_timestamp = 0
+    self.eyesaver.display_time_clock = ''
+    self:eyeSaverSaveSettings()
+end
+
+function ReadTimer:eyeSaverSaveSettings()
+    self.eyesaver.settings:saveSetting("eyesaver_timestamp", self.eyesaver.display_timestamp)
```
poire-z, line 135: I guess you only need to save it as a setting because the next timer would be lost when you switch books or between reader and filemanager? As when you restart the reader, one should be fine to restart the timer to have the message in 20mn since the restart ?
In which case, you could just save it into another long-living table, like PluginShare (which is already used to keep some state across plugin reload).

plugins/readtimer.koplugin/main.lua
```diff
@@ -19,9 +31,33 @@ function ReadTimer:init()
...
+    self:initEyeSaver()
     self.ui.menu:registerToMainMenu(self)
 end
 
+function ReadTimer:initEyeSaver()
+    self.eyesaver.display_time_clock = self.eyesaver.settings:readSetting("eyesaver_display_time_clock") or ""
+    self.eyesaver.display_timestamp = self.eyesaver.settings:readSetting("eyesaver_timestamp") or 0
+    self.eyesaver.enabled = self.eyesaver.settings:readSetting("eyesaver_enabled") or false
+
+    self.eyesaver.timed_display_callback = function()
```
poire-z, line 43: In other places, we call that kind of schedule functions: `something_action` (`_action` helping to distinguish from UI `_callback`).

poire-z:
It hurt my eyes to read everywhere `EyeSaver`, `self.eyesaver` :) 
This naming might be unecessary too specific, as it could more broadly be named `AssSaver`, `self.asssaver` and used for other things:
- move your ass out of the couch every 20mn and do some exercise
- check the meat in the oven every 10mn
- check the kids in the next room every 30mn
- go give some water to drink to your grandma in the other next room every hour

Seriously, I'd rather see no mention of "eye saving", and just consider this a "periodic timer", and see `self.periodic_timer` when I read the code.
I don't think it needs any message, just some kind of "periodic timing reached" (or something a bit better) as people would probably remember why they did set it.

If you really need a message, then you'll need to save it as a setting - and then I could be fine with just using "Look a few seconds away from your reader to save you eyes" as a hint= for the message.
And the timing out of 20s is also too specific (it make sense in the eye saving usage, and not much in others, but one could as well count from 1 to 20.)

Frenzie:
@poire-z 
> With regard to the plugin itself, there's already a timer plugin. From a UI perspective I think adding a recurring timer option to that one would add less noise.

pazos:
@smartscripts-nl: are you still interested in submitting this one?.

PR #6633, opened 2020-09-05
Pocketbook misc fixes

pazos:
After the awesome work made by @ezdiy:

- Remove empty dir from release zip
- Remove misc files (txt, md, html, gif) to prevent displaying them in PB library
- Do not asociate executable files (sh and py for the moment), just openable files
- Rename the font sample from html to template (to prevent wiping it during clean)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6633)


frontend/document/documentregistry.lua
```diff
@@ -160,6 +160,15 @@ function DocumentRegistry:getExtensions()
     return t
 end
 
+--- Get mapping of openable file extensions to providers
+-- @treturn table mapping openable file extensions to a list of providers
+function DocumentRegistry:getOpenableExtensions()
+    local t = self:getExtensions()
+    t["sh"] = nil
```
Frenzie, line 167: You can just write t.sh

frontend/document/documentregistry.lua
```diff
@@ -160,6 +160,15 @@ function DocumentRegistry:getExtensions()
     return t
 end
 
+--- Get mapping of openable file extensions to providers
+-- @treturn table mapping openable file extensions to a list of providers
+function DocumentRegistry:getOpenableExtensions()
```
Frenzie, line 165: Hm, not the biggest fan of the name. Why not just `getSupportedExtensions()` or some such?

Frenzie (approved):
lgtm besides the nits

frontend/document/documentregistry.lua
```diff
@@ -160,6 +160,15 @@ function DocumentRegistry:getExtensions()
...
+    local t = self:getExtensions()
```
NiLuJe, line 166: Unless I'm missing something, that's not a `DocumentRegistry` method?

frontend/document/documentregistry.lua
ezdiy: I'm slightly in favor of keeping strings in cases like this in lua, because it makes it obvious it's about the string content in the table, and not something that is an object field. While semantics are the same, the syntax distinction indicates intent.

Frenzie: Hm, alright.

frontend/document/documentregistry.lua
pazos: Not sure, .sh and .py are supported extensions from the POV of DocumentRegistry. It just happens that they're not openable.

Probably makes more sense to make it a local function in `file_ext_assoc.lua` ??

ezdiy: @pazos In fact, koreader might very well have a nice reason to route `.lua` and `.sh` to itself - former running in its own context for "quick koreader apps", the latter to run in a nifty terminal or such.

For hard filtering, the only ones we'd really care about is `.app` and `.run` - that could possibly hijack launching of everything else, leading to a world of trouble:

```
app:@APP_file:4:start.app:ICON_APP
run:@APP_file:4:start.app:ICON_APP
```

Since these are pocketbook specific (and very likely it's going to be the same way on other devices too), perhaps it should responsibility of the frontend that is aware of the specifics.

pazos: If there's a way to avoid indexing `/mnt/ext1/applications/koreader` from pocketbook launcher/library I'm fine with leaving the code as is.

My train of thoughts says than most users that discover the file assoc options will click on associate all. It would be a drag to have these users with hundreds of lua and shell files.

PR #6639, opened 2020-09-07
Update KoboUSBMS for CJK support

NiLuJe:
It's not particularly pretty, but still better than shipping a 16MB font twice ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6639)


Frenzie (approved):
Alright.

NiLuJe:
Yup, because of course NotoCJK is ~16MB, while /tmp is... also 16MB :D.

PR #6645, opened 2020-09-11
Only ask to start an USBMS session when plugged into a USB host on Kobo

NiLuJe:
Also, fix a typo that broke (PC) plug out detection on Cervantes ;).

Depends on https://github.com/koreader/koreader-base/pull/1183

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6645)

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -159,13 +159,27 @@ function UIManager:init()
             -- NOTE: Plug/unplug events will wake the device up, which is why we put it back to sleep.
             if Device.screen_saver_mode then
                 self:suspend()
+            end
+        end
+        self.event_handlers["NotCharging"] = function()
+            -- We need to put the device into suspension, other things need to be done before it.
+            self:_afterNotCharging()
+            if Device.screen_saver_mode then
+                self:suspend()
+            end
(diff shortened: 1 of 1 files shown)


NiLuJe:
Previously, you could start a USBMS session when plugged into a wall charger.

Which was technically harmless, but made no sense ;).

NiLuJe:
Since v0.9.8, KoboUSBMS itself will attempt to prevent that, too, but it can only do so on Mk. 7.

Since we can't guarantee that we're the one catching the plug in event either, I refrained from keeping track of it here. Especially since, again, while nonsensical, it's harmless in practice.

Frenzie (approved):

NiLuJe:
Sidebar: figuring out whether I could detect the charger type post plug event was a fun romp through kernel sources (c.f., https://github.com/koreader/KoboUSBMS/compare/v0.9.7...v0.9.8#diff-889ee969b59698c9f8f129167b74967bR908), so, if someone with a Mk. 6 or < Mk. 5 devices could double-check my theories, that would be neat ;).

I doubt there's anything we can do on < Mk. 5, but it *might* be salvageable on Mk. 6.

PR #6648, opened 2020-09-12
Prevent automatic footer refreshes from drawing the footer on top of stuff that isn't ReaderUI

NiLuJe:
Issue exposed by #6540, as long as `auto_refresh_time` is enabled.

Fix #6616

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6648)

Issue #6616, opened 2020-09-01
Refreshing time shows up minibar


hius07:
Kindle PW2, 2020.08.1-44

Refreshing time shows up minibar on pages where it shouldn't be shown (History, TOC, popup footnotes)

![1](https://user-images.githubusercontent.com/62179190/91834367-a18afa00-ec50-11ea-910b-3b9b8f87a5f2.png)

hius07:
The issue has occurred recently, maybe after https://github.com/koreader/koreader/commit/bafc52afd1da7cd19a699fbb7a5f52dcc66eb00e?

hius07:
@poire-z, can you please look at it.

NiLuJe:
It was never supposed *not* to do that ;).

It possibly didn't happen before #6563, but that was arguably a happy coincidence ;).

I'm not *quite* sure how it can trigger in History, though, unless the History view doesn't trigger a `CloseDocument`, for some reason?

But it's perfectly "normal" in anything that's technically still ReaderUI (i.e., ToC, Footnotes).

I don't *think* it hurts for footnotes, though?

Not quite sure how we could easily prevent that in the ToC, I'm not familiar with how this bit works.

NiLuJe:
Can we get essentially bracket `ReaderToc:onShowToc()` between `ReaderFooter:onSuspend()` and ``ReaderFooter:onResume()`?

hius07:
Is it OK to see the minibar over the menu?

![Reader_2020-09-12_193138](https://user-images.githubusercontent.com/62179190/93000136-04bd3c00-f52f-11ea-8b77-d574d8069229.png)

NiLuJe:
Ah, yep, that applies to `ConfigDialog`, too :D.

NiLuJe:
The least dodgy way to fix it is probably to whip up a new event, emit it where needed, and let `ReaderFooter` handle it?

NiLuJe:
@hius07: Any other places where it shouldn't show up but does?

poire-z:
Shouldn't it be UIManager job to manage the stack/covering and prevent paintTo if covered, or issue a paintTo to the covering widgets too ?
If we don't have that, that's probably because we never needed it as nothing else redraw itself in the background?

Or just add a `self.view.footer_covered` that the few things that know they are covering the footer could set and unset ?
At minima, it could check if there's something with `covers_fullscreen` in the widget stack (but it will then still show over footnotes popup, configdialog...)

[...]

PR #6650, opened 2020-09-13
Hilariously simpler fix for #6616

NiLuJe:
Reverts #6648 for a much, much simpler solution, one that will actually
cover every possible situation automagically.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6650)


NiLuJe:
Namely, if ReaderUI is not the top-level widget, don't repaint the footer.

That covers menus, popups, dialogs, everything ;).

Frenzie (approved):
That's more like it. ;-)

poire-z (approved):

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -20,6 +20,7 @@ local VerticalGroup = require("ui/widget/verticalgroup")
 local VerticalSpan = require("ui/widget/verticalspan")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
 local util = require("util")
+local logger = require("logger")
```
Frenzie, line 23: Overlooked this but that doesn't look very alphabetical. ^_^
```suggestion
local logger = require("logger")
local util = require("util")
```

Issue #6649, opened 2020-09-12
OPDS "by newest" view skips over entries


alethiophile:
* KOReader version: v2020.08.1-58-g4e281f0_2020-09-03
* Device: Kobo Forma

#### Issue
When viewing the "By newest" view in a Calibre OPDS library display, the list skips over blocks of entries on later loads, such that many books are never seen in this view.

#### Steps to reproduce
Unclear. Seems to occur only on subsequent loads, maybe a caching issue?

Best guess: load the "By newest" view once; wait a while/add more books to Calibre; load view again

NiLuJe:
Might be another weird parser quirk, àla #5949?

NiLuJe:
(i.e., we'll need a real repro testcase that we can look at in a browser, to confirm that theory ^^).

alethiophile:
In my experience it seems to be more about the location of entries in the list rather than their contents. But I've got a reproducing data test case, I'll see about uploading it.

alethiophile:
These are the first three pages in the OPDS XML that Calibre is generating:

[4f6e6577657374?library_id=ebooks.txt](https://github.com/koreader/koreader/files/5213121/4f6e6577657374.library_id.ebooks.txt)
[4f6e6577657374?library_id=ebooks;offset=30.txt](https://github.com/koreader/koreader/files/5213122/4f6e6577657374.library_id.ebooks.offset.30.txt)
[4f6e6577657374?library_id=ebooks;offset=60.txt](https://github.com/koreader/koreader/files/5213123/4f6e6577657374.library_id.ebooks.offset.60.txt)

All the entries in the first page display correctly. The second page displays correctly up until "Mud's Mission"; then it skips all the remaining entries in the second page and shows the first entry on the third page, "Mine Were of Trouble", next.

NiLuJe:
Hmm, yeah, nothing jumps out, and I guess that if you *can* actually get some of the skipped ones to show up in other view, then it's not a parsing error ;).

NiLuJe:
At a quick glance, I can't reproduce with a "static" Calibre Library (i.e., no change in Last-Modified).

I've added more verbose logging around caching in #6653, might be worth looking at debug logs once it lands, see if something emerges ;).

PR #6653, opened 2020-09-13
More verbose debug logging around OPDS caching

NiLuJe:
Re #6649

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6653)


NiLuJe:
(Also, a sligthly less confusing hash key ^^).

PR #6656, opened 2020-09-15
Only show Wi-Fi icon/symbol in footer when Wi-Fi has been enabled

smartscripts-nl:
The Wi-Fi icon is very visible in the footer. So I think it better to hide it when Wi-Fi has been disabled. In which case the "on" and "off" status indications in the footer are also unnecessary.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6656)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -245,10 +245,11 @@ local footerTextGeneratorMap = {
         local prefix = symbol_prefix[symbol_type].wifi_status
         local NetworkMgr = require("ui/network/manager")
         if NetworkMgr:isWifiOn() then
-            return T(_("%1 On"), prefix)
-        else
-            return T(_("%1 Off"), prefix)
+            -- only show wifi-icon/symbol when wifi enabled:
+            return prefix
         end
+        -- wifi off, so no icon/symbol:
+        return ""
     end,
     book_title = function(footer)
         local doc_info = footer.ui.document:getProps()
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -245,10 +245,10 @@ local footerTextGeneratorMap = {
         local prefix = symbol_prefix[symbol_type].wifi_status
         local NetworkMgr = require("ui/network/manager")
         if NetworkMgr:isWifiOn() then
-            -- only show wifi-icon/symbol when wifi enabled:
+            -- only show Wi-Fi-icon/symbol when Wi-Fi enabled:
             return prefix
         end
-        -- wifi off, so no icon/symbol:
+        -- Wi-Fi off, so no icon/symbol:
         return ""
     end,
     book_title = function(footer)


smartscripts-nl:
In several places in readerfooter.lua there are conditions which ensure the Wi-Fi status icon/symbol is only available for display in the footer on Android devices.

However, in my personal installation I've pruned these conditions away. On my Kobo Forma the Wi-Fi status is then exactly displayed as I expected. So what's the reason for these conditions? Maybe newer Kobo e-readers don't have the problems that older versions had with determining the Wi-Fi status? Or something like that?

Frenzie:
I don't recall otoh if that's still relevant after #6424, but it's not a matter of *getting* the status but the *speed* of getting it.

smartscripts-nl:
> I don't recall otoh if that's still relevant after #6424, but it's not a matter of _getting_ the status but the _speed_ of getting it.

The speed of determining the status is quite satisfactory on my Kobo Forma.

Frenzie:
That depends on your perspective. ~1.5 ms (best case scenario only) is an eternity.

Walk to the very back of your house/outside so that your wifi claims it's connected but not really and see if it doesn't freeze up for a few seconds. That's fine…ish for actions that logically require network activity. Displaying the footer is not one of those kinds of actions. ;-)

smartscripts-nl:
Hm, I'll try that later. But could the status of Wi-Fi not be cached in a variable, so the device doesn't have to query the connection every time the footer is being refreshed?

smartscripts-nl:
I wonder: is there a global registry anywhere in KOReader where statuses can be registered in memory across reloads or opening of another ebook (apart from registering such statuses in settings.reader.lua)?

Frenzie:
> But could the status of Wi-Fi not be cached in a variable, so the device doesn't have to query the connection every time the footer is being refreshed?

That's very much the point, yes. You would need an immediate status return when queried that regularly updates itself in the background.

> I wonder: is there a global registry anywhere in KOReader where statuses can be registered in memory across reloads or opening of another ebook (apart from registering such statuses in settings.reader.lua)?

What would the advantage of that be? Wifi status is definitely *not* something you'd want to store in any way whatsoever. :-P

pazos:
> I wonder: is there a global registry anywhere in KOReader where statuses can be registered in memory across reloads or opening of another ebook (apart from registering such statuses in settings.reader.lua)?

That would defeat the whole purpose of the connectivity check, which is to check connectivity :)

> I wonder: is there a global registry anywhere in KOReader where statuses can be registered in memory across reloads or opening of another ebook (apart from registering such statuses in settings.reader.lua)?

You don't need that. You can query the network device directly for a wireless carrier. It works since linux 2.6.18. It won't show if the device is online because being online is not different from being associated to AP at this level, so it could report you're online even when the wifi client is assoc to an offline AP.

https://lwn.net/Articles/103957/

But things other than reading sysfs would block for the whole eternity each page turn, that includes `ping`

NiLuJe:
Yeah, #6424 hasn't changed anything on that front, and I'll probably still veto that kind of thing ad vitam eternam, even if, admittedly, there are worse `isWifiOn` implementation that on Kobo's (especially since it doesn't actually fork lsmod, and just checks sysfs directly)  ;).

NiLuJe:
As far as this specific tweak is concerned, I don't particularly care myself, but another option is the crossed-off Wi-Fi icon, like I used in [KoboUSBMS](https://github.com/koreader/KoboUSBMS/blob/46629994a7d2dcbe973fe7fb93466b4cc80cd859/usbms.c#L280).

(Except you'd have to check the codepoint, the nerdfont we bundles in KOReader has been patched, unlike the one bundled in KoboUSBMS).

NiLuJe:
I recall extremely weird corner-casey issues with the carrier check on Kobo when we tried that, which is why we *don't* do it on Kobo, unlike on Cervantes. But  it may have been a Lua snafu that's since been fixed, I don't recall the details.

In fact, that hasn't stopped me from doing it elsewhere (most recently in KoboUSBMS).

NiLuJe:
Okay, see https://github.com/koreader/koreader/pull/5211 for a recap of the `isWifiOn` situation, if someone feels like actually relaxing those checks for real for the ReaderFooter icon ;).

(I'm not... super-fond of the idea, but, eh, probably not too terrible).

NiLuJe:
(Because, yeah, reading procfs/sysfs incurs a latency cost, even if it's likely to be the least terrible option for that kind of things).

pazos:
> (Because, yeah, reading procfs/sysfs incurs a latency cost, even if it's likely to be the least terrible option for that kind of things).
> 
> It's likely going to be in the high-range of sub-10ms, FWIW.
> 
> Which is... already way too much for my tastes, but, there you have it.

I think it should be much less than that (something like 0.00x ms) and the cheapest thing available. Also it is much cheaper than the JNI call we do on android. It is also very reliable

Frenzie:
> (Because, yeah, reading procfs/sysfs incurs a latency cost, even if it's likely to be the least terrible option for that kind of things).
> 
> It's likely going to be in the high range of sub-10ms, FWIW.

Surely not? O_o

NiLuJe:
Okay, yeah, a quick round of benchmarking would hint towards 0.300ms ;).

```
 #!./luajit

local ffi = require("ffi")
local C = ffi.C

ffi.cdef[[
struct timeval {
  long int tv_sec;
  long int tv_usec;
};
int gettimeofday(struct timeval *restrict, struct timezone *restrict) __attribute__((nothrow, leaf));
]]

timeval = ffi.new("struct timeval")
function gettime()
	C.gettimeofday(timeval, nil)
	return tonumber(timeval.tv_sec), tonumber(timeval.tv_usec)
end

function getTimestamp()
	local secs, usecs = gettime()
	return secs + usecs/1000000
end

function getDuration(from_timestamp)
	return getTimestamp() - from_timestamp
end



function isWifiOn()
	local fd = io.open("/proc/modules", "r")
	if fd then
		local lsmod = fd:read("*all")
		fd:close()
		-- lsmod is usually empty, unless Wi-Fi or USB is enabled
		-- We could alternatively check if lfs.attributes("/proc/sys/net/ipv4/conf/" .. os.getenv("INTERFACE"), "mode") == "directory"
		-- c.f., also what Cervantes does via /sys/class/net/eth0/carrier to check if the interface is up.
		-- That said, since we only care about whether *modules* are loaded, this does the job nicely.
		if lsmod:len() > 0 then
		local module = os.getenv("WIFI_MODULE") or "sdio_wifi_pwr"
		if lsmod:find(module) then
			return true
		end
		end
	end
	return false
end

--start_clock = os.clock()
start_ts = getTimestamp()
wifi_status = isWifiOn()
end_ts = getTimestamp()
--end_clock = os.clock()
print("isWifiOn:", wifi_status)
--print(string.format("isWifiOn took %9.3f ms", (end_clock - start_clock) * 1000))
print(string.format("isWifiOn took %9.3f ms", (end_ts - start_ts) * 1000))
```

Both os.clock & gettimeofday agree, FWIW.

NiLuJe:
TL;DR: https://github.com/koreader/koreader/pull/5211#issuecomment-521304139 stands ;).

Should be okay on Cervantes & Kobo, as well as possibly on PB.

NiLuJe:
Also, I misremembered, the carrier thingy I mentioned earlier was on Kindle, not Kobo, and it was probably because Kindle reasons ;p.

NiLuJe:
So to recap:

* Feel free to relax the check to allow Android + Kobo + Cervantes.
* I'd personally prefer toggling the prefix icon itself, but that would need some special-casing, as it wouldn't work for the text variant of the prefixes. (Actually, the current approach *also* doesn't make sense for text only prefixes anyway ^^).

smartscripts-nl:
Don't know if this is still relevant, but in my personal installation I'm now caching the wifi status in reader.settings.lua. Works like a charm. This setting is toggled by NetworkListener:onToggleWifi(). For the symbol in the footer the actual wifi status only gets queried when the cached wifi status in reader.settings.lua is -1. So almost never a delay when generating the footer.

It seems to me the worst thing that can happen is that my footer indicates that I have wifi, but in fact it doesn't work, because I am too far away from my router. I can live with that...

NiLuJe:
> Don't know if this is still relevant, but in my personal installation I'm now caching the wifi status in reader.settings.lua. Works like a charm. This setting is toggled by NetworkListener:onToggleWifi(). For the symbol in the footer the actual wifi status only gets queried when the cached wifi status in reader.settings.lua is -1. So almost never a delay when generating the footer.
> 
> It seems to me the worst thing that can happen is that my footer indicates that I have wifi, but in fact it doesn't work, because I am too far away from my router. I can live with that...

`ToggleWifi` is only emitted by Dispatcher. There are approximately seven other billion ways that WiFI could be toggled on or off ;).

`NetworkConnected` and `NetworkDisconnected` are only *mildly* more reliable, and that only on some platforms.

Then again, as we've said earlier, the whole caching thing is overkill and asking for trouble anyway ;).

NiLuJe:
Ping?

NiLuJe:
FWIW, what I had in mind looks like this:

```diff
diff --git a/frontend/apps/reader/modules/readerfooter.lua b/frontend/apps/reader/modules/readerfooter.lua
index 9d4cfeb7..10984b7b 100644
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -75,6 +75,7 @@ local symbol_prefix = {
         frontlight = "☼",
         mem_usage = "",
         wifi_status = "",
+        wifi_status_off = "",
     }
 }
 if BD.mirroredUILayout() then
@@ -239,13 +240,22 @@ local footerTextGeneratorMap = {
         return ""
     end,
     wifi_status = function(footer)
+        -- NOTE: This one deviates a bit from the mold because, in icons mode, we simply use two different icons and no text.
         local symbol_type = footer.settings.item_prefix or "icons"
-        local prefix = symbol_prefix[symbol_type].wifi_status
         local NetworkMgr = require("ui/network/manager")
-        if NetworkMgr:isWifiOn() then
-            return T(_("%1 On"), prefix)
+        if symbol_type == "icons" then
+            if NetworkMgr:isWifiOn() then
+                return symbol_prefix.icons.wifi_status
+            else
+                return symbol_prefix.icons.wifi_status_off
+            end
         else
-            return T(_("%1 Off"), prefix)
+            local prefix = symbol_prefix[symbol_type].wifi_status
+            if NetworkMgr:isWifiOn() then
+                return T(_("%1 On"), prefix)
+            else
+                return T(_("%1 Off"), prefix)
+            end
         end
     end,
     book_title = function(footer)
@@ -334,7 +344,7 @@ function ReaderFooter:init()
     }
 
     -- Remove items not supported by the current device
-    if not Device:isAndroid() then
+    if not Device:hasFastWifiStatusQuery() then
         MODE.wifi_status = nil
     end
     if not Device:hasFrontlight() then
@@ -1633,7 +1643,7 @@ function ReaderFooter:addToMainMenu(menu_items)
         table.insert(sub_items, getMinibarOption("frontlight"))
     end
     table.insert(sub_items, getMinibarOption("mem_usage"))
-    if Device:isAndroid() then
+    if Device:hasFastWifiStatusQuery() then
         table.insert(sub_items, getMinibarOption("wifi_status"))
     end
     table.insert(sub_items, getMinibarOption("book_title"))
@@ -2016,6 +2026,18 @@ function ReaderFooter:onFrontlightStateChanged()
     end
 end
 
+function ReaderFooter:onNetworkConnected()
+    if self.settings.wifi_status then
+        self:onUpdateFooter(true)
+    end
+end
+
+function ReaderFooter:onNetworkDisconnected()
+    if self.settings.wifi_status then
+        self:onUpdateFooter(true)
+    end
+end
+
 function ReaderFooter:onSetRotationMode()
     self:updateFooterContainer()
     self:resetLayout(true)
diff --git a/frontend/device/android/device.lua b/frontend/device/android/device.lua
index 0311f7b3..13c59bb1 100644
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -86,6 +86,7 @@ local Device = Generic:new{
     isHapticFeedbackEnabled = yes,
     hasClipboard = yes,
     hasOTAUpdates = canUpdateApk,
+    hasFastWifiStatusQuery = yes,
     canOpenLink = yes,
     openLink = function(self, link)
         if not link or type(link) ~= "string" then return end
diff --git a/frontend/device/cervantes/device.lua b/frontend/device/cervantes/device.lua
index 78ffca39..1d1bfc5d 100644
--- a/frontend/device/cervantes/device.lua
+++ b/frontend/device/cervantes/device.lua
@@ -58,6 +58,7 @@ local Cervantes = Generic:new{
     touch_mirrored_x = true,
     touch_probe_ev_epoch_time = true,
     hasOTAUpdates = yes,
+    hasFastWifiStatusQuery = yes,
     hasKeys = yes,
     hasWifiManager = yes,
     canReboot = yes,
diff --git a/frontend/device/generic/device.lua b/frontend/device/generic/device.lua
index 9bef87e9..d0c27272 100644
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -95,6 +95,10 @@ local Device = {
     -- set to yes on devices that support over-the-air incremental updates.
     hasOTAUpdates = no,
 
+    -- set to yes on devices that have a non-blocking isWifiOn implementation
+    -- (c.f., https://github.com/koreader/koreader/pull/5211#issuecomment-521304139)
+    hasFastWifiStatusQuery = no,
+
     canOpenLink = no,
     openLink = no,
     canExternalDictLookup = no,
diff --git a/frontend/device/kobo/device.lua b/frontend/device/kobo/device.lua
index c6cb7e69..0476beb8 100644
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -25,6 +25,7 @@ local Kobo = Generic:new{
     isKobo = yes,
     isTouchDevice = yes, -- all of them are
     hasOTAUpdates = yes,
+    hasFastWifiStatusQuery = yes,
     hasWifiManager = yes,
     canReboot = yes,
     canPowerOff = yes,
```

I can PR that if that's fine with you?

smartscripts-nl:
> I can PR that if that's fine with you?

No problem, go ahead!

PR #6657, opened 2020-09-15
Minor util & ffi/util cleanups

NiLuJe:
Procrastination is fun! :D.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6657)

### DIFF
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -2,6 +2,7 @@ local Blitbuffer = require("ffi/blitbuffer")
 local CanvasContext = require("document/canvascontext")
 local DataStorage = require("datastorage")
 local Document = require("document/document")
+local FFIUtil = require("ffi/util")
 local FontList = require("fontlist")
 local Geom = require("ui/geometry")
 local RenderImage = require("ui/renderimage")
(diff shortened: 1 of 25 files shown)


NiLuJe:
And in case we ever decide on unifying the variable name: I kinda like `ffiUtil`, as that means it's not lost in the maze of CamelCase widgets ;).

poire-z:
> Procrastination is fun! :D.

I understand, I too would do anything to not get started with https://github.com/koreader/koreader-base/pull/762#issuecomment-664003274 :)

> And in case we ever decide on unifying the variable name: I kinda like ffiUtil,

I'm fine with `ffiUtil` too.

frontend/document/credocument.lua
```diff
@@ -1137,20 +1140,14 @@ function CreDocument:setupCallCache()
...
         -- cache statistics
         self._call_cache_stats = {}
-        local _gettime = require("ffi/util").gettime
-        getTime = function()
-            local secs, usecs = _gettime()
-            return secs + usecs/1000000
-        end
         addStatMiss = function(name, starttime, not_cached)
-            local duration = getTime() - starttime
+            local duration = FFIUtil.getTimestamp() - starttime
```
poire-z, line 1150: You could also switch them to use  `ffiUtil.getDuration(starttime)` like I started doing in:
https://github.com/koreader/koreader/blob/078b579081a29d24f229d87b08701ca6985a75fd/frontend/apps/reader/readerui.lua#L284-L286

NiLuJe: Yeah, I kept them separate mostly to avoid the noise of the time it'll take for string formatting, which is likely irrelevant in the grand scheme of things, but, oh, well ;p.

NiLuJe:
> > Procrastination is fun! :D.
> 
> I understand, I too would do anything to not get started with [koreader/koreader-base#762 (comment)](https://github.com/koreader/koreader-base/pull/762#issuecomment-664003274) :)
> 

:D.

I'm unlikely to have the bandwidth to deal with it until winter break anyway ;).

poire-z (approved):

Frenzie (approved):
lgtm, @poire-z 's comment makes sense

poire-z:
@Frenzie : when do you plan to release 2020.09?
I tested this morning nightly with this, seems fine.

Frenzie:
I was planning this morning but something got in between.

PR #6658, opened 2020-09-15
remarkable: fix fbdepth use in startup script

tcrs:
The reMarkable version of fbdepth doesn't support the -1 rotation so
this was actually not setting the rotation at all. I didn't notice this
as I was always launching koreader from xochitl so the rotation was
already set correctly.

This means that if koreader is started up as the default UI then the
display actually gets set up correctly!

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6658)

### DIFF
--- a/platform/remarkable/koreader.sh
+++ b/platform/remarkable/koreader.sh
@@ -108,7 +108,7 @@ ko_do_fbdepth() {
         # Swap to 8bpp if things look sane
         if [ -n "${ORIG_FB_BPP}" ]; then
             echo "Switching fb bitdepth to 8bpp & rotation to Portrait" >>crash.log 2>&1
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Thanks! :-)

NiLuJe:
Oops, I may actually have inadvertently broken it in https://github.com/NiLuJe/FBInk/commit/0e8fd3b06b084aecea3b6549da7dce711b24beca (because ntxBootRota is set to CW (i.e., 1) on rM, with a comment specifically mentioning that it was for fbdepth).

In any case, doing it this way certainly makes sense, since there's only a single supported rM variant right now ;). (Meaning this might need to be revistied for the rM 2 ^^).

PR #6662, opened 2020-09-16
kindle: honor --asap

yparitcher:
Recently I started switching between stock kindle & koreader a lot. I got frustrated at how long it takes to startup (my laptop boots faster ;)  to that end i trimmed down KUAL and made a dedicated launcher for koreader. but the koreader.sh still takes ~5s and reader.lua ~5sec

this is the first of a few fixes to speed up startup

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6662)


NiLuJe (approved):
Fine by me ;).

PR #6664, opened 2020-09-17
ReaderFooter: update FL on toggle

yparitcher:
`isFrontlightOnHW()` was returning a value from powerd not from hardware before powerd was updated, therefore `_decideFrontlightState()` would have the old state.

The `FrontlightStateChanged` event used by `ReaderFooter` was called by `_setIntensity` before `powerd.is_fl_on` was toggled by `_decideFrontlightState()`

Also add `isFrontlightOnHW()` for the emulator (it was broken)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6664)

### DIFF
--- a/frontend/device/generic/powerd.lua
+++ b/frontend/device/generic/powerd.lua
@@ -32,8 +32,8 @@ function BasePowerD:setDissmisBatteryStatus(status) self.battery_warning = statu
 function BasePowerD:isChargingHW() return false end
 function BasePowerD:frontlightIntensityHW() return 0 end
 function BasePowerD:isFrontlightOnHW() return self.fl_intensity > self.fl_min end
-function BasePowerD:turnOffFrontlightHW() self:_setIntensity(self.fl_min) end
-function BasePowerD:turnOnFrontlightHW() self:_setIntensity(self.fl_intensity) end --- @fixme: what if fl_intensity == fl_min (c.f., kindle)?
+function BasePowerD:turnOffFrontlightHW() self:setIntensityHW(self.fl_min) end
+function BasePowerD:turnOnFrontlightHW() self:setIntensityHW(self.fl_intensity) end --- @fixme: what if fl_intensity == fl_min (c.f., kindle)?
 -- Anything needs to be done before do a real hardware suspend. Such as turn off
 -- front light.
 function BasePowerD:beforeSuspend() end
@@ -79,6 +79,7 @@ function BasePowerD:turnOffFrontlight()
     if self:isFrontlightOff() then return false end
     self:turnOffFrontlightHW()
     self.is_fl_on = false
+    self:stateChanged()
     return true
 end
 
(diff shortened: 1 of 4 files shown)


NiLuJe:
The extremely roundabout way is unfortunately on purpose, because Kobo can't query the HW.

So, I'm afraid this doesn't quite work on Kobo, at least when some shitty corner-cases are involved (e.g., launching KOReader w/ the light off).

NiLuJe:
`BasePowerD:new` should be setting `fl_intensity` and `is_fl_on` already. Might be a chicken-egg issue on Kindle because lipc is not ready at that point?

NiLuJe:
And, checking whether the toggle updates the footer on Kobo, I just found out that while it does, the repaint is wonky as hell (it repaints on top of the existing footer? o_O).

yparitcher:
Reading the frontlight code in middle of the night does give one a headache :+1: 

my comment wasn't very clear but the issue is when going from on -> off via `turnOffFrontlight()`, `is_fl_on` is only updated after the `FrontlightStateChanged` event. this is **not** an issue on startup.

if we cant make `isFrontlightOnHW()` rely on HW (against its name), then we would have to copy the event code to `turnOffFrontlight()/turnOnFrontlight()` and also to any platform that overrides them leading to a lot of duplication. (this was already duplicated only to kobo which is why it worked there until now)

NiLuJe:
Oh, right, thanks for the clarification ;).

Unfortunately, yeah, I don't think we have a nicer solution than the duplication fest in the toggles?

yparitcher:
> Oh, right, thanks for the clarification, that makes more sense ;).
> 
> Unfortunately, yeah, I don't think we have a nicer solution than the duplication fest in the toggles?

but then we will probably have the issue of #6667 with multiple events being triggered

NiLuJe:
> but then we will probably have the issue of #6667 with multiple events being triggered :(
> unless we get rid of the event in `_setIntensity` and only have it everywhere else (much more work)

No, #6667 was pretty specifically tied to the "fancy" ramp-up/ramp-down from a different process. That's fairly specific to Kobo, unless I did it elsewhere and I forgot? ^^.

yparitcher:
@NiLuJe 
I updated it and it didn't need to much duplication. this obsoletes your fix.
works on kindle & emulator . please test on kobo (I don't think i broke it)

yparitcher:
tests pass on my end not sure why circleci failed

NiLuJe:
Will test tomorrow, but, yeah, that looks good ;).

Didn't remember that the whole `_setIntensity` weirdness came to be solely for the toggles, this definitely feels cleaner ;).

NiLuJe (changes requested):
There's a `_setIntensity` reference left in a comment @ `frontend/device/kindle/powerd.lua`:`24`, but other than that, LGTM, and it appears to behave on Kobo ;).

Thanks!

yparitcher:
> There's a `_setIntensity` reference left in a comment @ `frontend/device/kindle/powerd.lua`:`24`, but other than that, LGTM, and it appears to behave on Kobo ;).
> 
> Thanks!

done

NiLuJe (approved):

yparitcher:
no clue why CircleCi decided to fail. otherwise I am ready to merge

NiLuJe:
Reran it for fun, but you're good to go, feel free to merge ;).

PR #6665, opened 2020-09-17
correct bug introduced with #6641

zwim:
getWarmth must not be called before warm_diff is initialized.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6665)

### DIFF
--- a/frontend/device/android/powerd.lua
+++ b/frontend/device/android/powerd.lua
@@ -39,10 +39,10 @@ function AndroidPowerD:init()
     end
 
     if self.device:hasNaturalLight() then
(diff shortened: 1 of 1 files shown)


Frenzie:
What happens if you do?

zwim:
If you call getWarmth before initialization of warm_diff you get a nil-Pointer access and KOReader crashes.

Please look at https://github.com/koreader/koreader/blob/master/frontend/device/android/powerd.lua#L54.

Frenzie:
The point is I was about to click publish. ;-) I'll hold off due to the crash.

PR #6666, opened 2020-09-17
Autostandby: Fix wrong assumptions about instance liveness.

### LINKED ISSUE #6647: PB626: auto-standby plugin not always enters device in standby state
* KOReader version: v2020.08.1-60-g051c8c0_2020-09-07 
* Device: PB626 sw:U626.5.20.1157

#### Issue:
auto-standby plugin not always enters device into standby hence the power consumption is much higher. When auto-standby plugin is working properly, the LED is ON for 2 seconds and then OFF until next tap ore button press. The LED on power button is ON when alive and OFF when in standby. So this is indirect indicator of actual device state for me :) 

#### Steps to reproduce:

I have preset this options:
- allow auto-standby ON
- min input idle seconds 1
- max input idle seconds 2
- input window seconds 2

1. start koreader
2. open ebook, in my case I opened PDF file
3. tap screen e.g. move to previous or next page
4. LED indicator on power button is lite still after 2 seconds .... 1 minute .... 5 minutes ....

If I start koreader, and **before opening any book** I browse in history or file browser after tapping screen or pressing button the LED always shuts OFF after 2 seconds, therefore I suppose that the plugin is working correctly in this case. After opening an ebook and tapping screen or pressing button standby never occurs even if I invoke menu or file browser.

##### Crash log:
[crash_20200912_185526_autostandby.log](https://github.com/koreader/koreader/files/5213011/crash_20200912_185526_autostandby.log)

ezdiy:
Got instantiated multiple times, when our state is meant to be global.
Fixes #6647

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6666)

### DIFF
--- a/plugins/autostandby.koplugin/main.lua
+++ b/plugins/autostandby.koplugin/main.lua
@@ -16,6 +16,8 @@ local _ = require("gettext")
 local AutoStandby = WidgetContainer:new{
     is_doc_only = false,
     name = "autostandby",
+
+    -- static for all plugin instances
     settings = LuaSettings:open(DataStorage:getSettingsDir() .. "/autostandby.lua"),
     delay = 0,
     lastInput = 0,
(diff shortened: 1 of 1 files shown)


ezdiy:
It's still kinda messy as the assumption here is that app transitions are atomic, ie no ui loop runs while the RD widget replaces FM or vice versa.

PR #6667, opened 2020-09-17
[Kobo] Don't send a flood a FrontlightStateChanged events during the FL ramps

NiLuJe:
This made the ReaderFooter refresh go super-wonky...

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6667)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -1704,6 +1704,7 @@ function ReaderFooter:getDataFromStatistics(title, pages)
 end
 
 function ReaderFooter:onUpdateFooter(force_repaint, force_recompute)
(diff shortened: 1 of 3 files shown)


Frenzie:
Wonky how? The force pushed v2020.09 tag with https://github.com/koreader/koreader/pull/6665 has just finished building. :-P

poire-z (approved):
No need to send an event for the last one of the ramp ?

Frenzie (approved):

NiLuJe:
@Frenzie: It basically either failed to refresh, or refresh a bunch of times on top of each other, which basically made the footer (and only the footer) unreadable until the next refresh.

@poire-z: Nope, it's a toggle, so we basically only want to know about the final state (i.e., on/off).

Frenzie:
@NiLuJe Sounds harmless enough. ;-)

Issue #6677, opened 2020-09-18
Wi-Fi / Network Manager not implemented on reMarkable


gbyl:
* KOReader version: 2020.09
* Device: reMarkable 1

#### Issue
The following toggle does not work on the reMarkable port:
Gear Icon > Network > Wi-Fi connection

#### Workaround
Switching the Wi-Fi on and off is essential to preserve battery life. Thus, I find myself running these complementary pairs of lines often:
`ifconfig wlan0 down`
`ifconfig wlan0 up`
Or:
`iwconfig wlan0 txpower off`
`rfkill unblock wifi`

While the above lines disable/enable Wi-Fi, I am not sure they are all that is needed to preserve battery life. As a beginner, I have not been able to integrate them into a toggle script, let alone add them in any meaningful way to the interface.

Eeems:
The built-in application (xochitl) uses wpa_supplicant to control the wifi connection. I haven't done a ton of digging into how it handles turning it off, but all I've been doing for disabling wifi has been `ifconfig wlan0 down` (https://github.com/Eeems/oxide/blob/master/applications/system-service/wlan.h#L26). I haven't done any experimentation around battery impact versus when xochitl disables wifi though.

pazos:
@gabyhk @Eeems: Yup, wifi was left unimplemented. It should work with little changes.

Have a look at https://github.com/koreader/koreader/blob/master/frontend/device/cervantes/device.lua#L224-L255 and the helper scripts in https://github.com/koreader/koreader/tree/master/platform/cervantes.

I would suggest using the cervantes as a template if you want to split network management between xochitl and KOReader. If you want to repurpose the wpa_supplicant daemon run by xochitl you can use kobo as the template.

On kobos wifi stays alive between Nickel -> KO and killed before running Nickel again
On cervantes wifi is killed between QBookApp -> KO and killed again at exit.

Eeems:
I've also been doing a lot of work in my own project for using the networks that xochitl has stored when dealing with wifi: https://github.com/Eeems/oxide/blob/master/applications/system-service/wifiapi.h#L560-L602

It might make sense to leverage some of that work.

gabyhk:
@pazos @Eeems I might not be the right person to implement those changes due to a lack of expertise, but I certainly appreciate both your comments pointing me in the right direction!

pazos:
[...]

PR #6681, opened 2020-09-19
[wip] Network selection on remarkable

Eeems:
I've started some work on #6677. This expects to be running underneath [Eeems/oxide](https://github.com/Eeems/oxide) which is where I've done a lot more work with actually handling the networks. I've got some outstanding issues with the network API in oxide that I've got to work out. When that's done and I've officially released a version that this will work with I can finish up this PR.

I wouldn't mind a code review though and some pointers at what utility functions to replace portions of my code with as I'm not really that familiar with this codebase, and I'm still not that confident in lua.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6681)

### DIFF
--- a/frontend/device/remarkable/device.lua
+++ b/frontend/device/remarkable/device.lua
@@ -181,17 +181,14 @@ function Remarkable:initNetworkManager(NetworkMgr)
         os.execute("rot wifi call enable")
         local UIManager = require("ui/uimanager")
         local InfoMessage = require("ui/widget/infomessage")
-        local info = InfoMessage:new{text = _("Turning on wifi…")}
+        local info = InfoMessage:new{text = _("Connecting to Wi-Fi…")}
         UIManager:show(info)
         UIManager:nextTick(function()
             -- Wait for wifi to enable
             while getWifiProperty("state") < 1 do
                 os.execute("sleep 1")
             end
-            UIManager:close(info)
             logger.info("Scanning for networks...")
-            info = InfoMessage:new{text = _("Scanning for networks…")}
-            UIManager:show(info)
             -- Initiate active scan
             os.execute("rot wifi call scan bool:true")
             -- Wait for scan to finish
--- a/frontend/device/remarkable/device.lua
+++ b/frontend/device/remarkable/device.lua
@@ -1,5 +1,6 @@
 local Generic = require("device/generic/device") -- <= look at this file!
 local logger = require("logger")
+local rapidjson = require("rapidjson")
 
 local function yes() return true end
 local function no() return false end
@@ -123,6 +124,117 @@ end
 
 function Remarkable:reboot()
     os.execute("systemctl reboot")
+end
+
+function os.capture(cmd, raw)
+    local f = assert(io.popen(cmd, 'r'))
+    local s = assert(f:read('*a'))
+    f:close()
+    if raw then return s end
+    s = string.gsub(s, '^%s+', '')
+    s = string.gsub(s, '%s+$', '')
+    s = string.gsub(s, '[\n\r]+', ' ')
+    return s
+end
+
+function getNetworkProperty(path, name)
+    path = strsub(path, strlen("/codes/eeems/oxide1/"))
+    local json, err = rapidjson.decode(os.capture("rot --object Network:" .. path .. " wifi get " .. name))
+    return json
+end
+function getBSSProperty(path, name)
+    path = strsub(path, strlen("/codes/eeems/oxide1/"))
+    local json, err = rapidjson.decode(os.capture("rot --object BSS:" .. path .. " wifi get " .. name))
+    return json
+end
+function getWifiProperty(name)
+    local json, err = rapidjson.decode(os.capture("rot wifi get " .. name))
+    return json
+end
+local function isempty(s)
+  return s == nil or s == ''
+end
+
+-- wireless
+function Remarkable:initNetworkManager(NetworkMgr)
+    if isempty(os.capture("rot")) then
+        return
+    end
+    function NetworkMgr:turnOffWifi(complete_callback)
+        logger.info("Remarkable: disabling Wi-Fi")
+        os.capture("rot wifi call disable")
+        if complete_callback then
+            complete_callback()
+        end
+    end
+    function NetworkMgr:turnOnWifi(complete_callback)
+        logger.info("Remarkable: enabling Wi-Fi")
+        os.capture("rot wifi call enable")
+        self:reconnectOrShowNetworkMenu(complete_callback)
+    end
+    function NetworkMgr:getNetworkInterfaceName()
+        return "wlan0"
+    end
+    NetworkMgr:setWirelessBackend("rot"})
+    function NetworkMgr:obtainIP()
+        os.capture("dhcpcd")
+    end
+    function NetworkMgr:releaseIP()
+        os.capture("dhcpcd -k")
+    end
+    function NetworkMgr:isWifiOn()
+        return tonumber(os.capture("rot wifi get state")) > 1
+    end
+
+    function NetworkMgr:getNetworkList()
+        local results = {}
+        local currentNetwork = getWifiProperty("network")
+        for path in getWifiProperty("bSSs") do
+            local flags = {}
+            for flag in getBSSProperty(path, "key_mgmt") do
+                table.insert(flags, "[" .. strupper(flag) .. "]")
+            end
+            local network = {
+                bssid = getBSSProperty(path, "bssid"),
+                ssid = getBSSProperty(path, "ssid"),
+                frequency = getBSSProperty(path, "frequency"),
+                signal_level = getBSSProperty(path, "signal"),
+                flags = flags,
+            }
+            network.signal_quality = math.min(math.max((network.signal + 100) * 2, 0), 100)
+            local networkPath = getBSSProperty(path, "network")
+            if networkPath then
+                if networkPath == currentNetwork then
+                    network.connected = true
+                end
+            end
+            table.insert(results, network)
+        end
+        return results
+    end
+    function NetworkMgr:getCurrentNetwork()
+        local path = getWifiProperty("network")
+        local results = {}
+        if path and path ~= "/" then
+            table.insert(results, {
+                ssid = getNetworkProperty(path, "ssid"),
+            })
+        end
+        return results
+    end
+    function NetworkMgr:authenticateNetwork(network)
+        local properties = {
+            ssid = network.ssid,
+        }
+        if network.psk then
+            properties.key_mgmt = "WPA-PSK"
+            properties.psk = network.password
+        end
+        local path, err = rapidjson.decode(os.capture("rot wifi call addNetwork 'QVariantMap:" .. rapidjson.encode(properties) .. "'"))
+    end
+    function NetworkMgr:disconnectNetwork(network)
+        os.execute("rot wifi call disconnect")
+    end
 end
 
 return Remarkable
--- a/frontend/device/remarkable/device.lua
+++ b/frontend/device/remarkable/device.lua
@@ -126,7 +126,7 @@ function Remarkable:reboot()
     os.execute("systemctl reboot")
 end
 
-function osCapture(cmd, raw)
+local function osCapture(cmd, raw)
     local f = io.popen(cmd, 'r')
     if not f then
         return nil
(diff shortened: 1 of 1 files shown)


frontend/device/remarkable/device.lua
```diff
@@ -125,5 +126,159 @@ function Remarkable:reboot()
...
+    f:close()
+    if not s then
+        return nil
+    end
+    if raw then
+        return s
+    end
+    s = string.gsub(s, '^%s+', '')
+    s = string.gsub(s, '%s+$', '')
+    s = string.gsub(s, '[\n\r]+', ' ')
```
Frenzie, line 144: I'm not sure if I'd expect those things to be filtered out by default. Could you clarify the thought process?

Frenzie (commented):
Looks reasonable enough at a glance. I'd prefer it if you didn't include the unrelated submodule bump though. ;-)

Eeems:
> Looks reasonable enough at a glance. I'd prefer it if you didn't include the unrelated submodule bump though. ;-)

I didn't do that on purpose, the submodule bumped when running kodev, so I assumed it should be included. I can revert that.

frontend/device/remarkable/device.lua
```diff
@@ -125,5 +126,159 @@ function Remarkable:reboot()
     os.execute("systemctl reboot")
 end
 
+function os.capture(cmd, raw)
```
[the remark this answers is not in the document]
Eeems: Sure, I just grabbed this from a Stack Overflow and don't really know enough about lua to know what is and isn't good practice.

frontend/device/remarkable/device.lua
Eeems: https://stackoverflow.com/a/326715
Basically it's removing extra whitespace and newlines. I left it in since I don't expect extra newlines in my results anyways.

Frenzie: Involving a particular set of results in the thought process seems contradictory to putting it on `os.*`.

Taking a peek around the code it looks like we normally just use something simple like this, but I can see how that'd be annoyingly repetitive for your use case here.

```lua
local std_out = io.popen("rot wifi")
if std_out then
    local result = std_out:read("*all")
    std_out:close()
end
```

This function might be useful in util, but it seems to me that the unfiltered output is what you'd expect by default. What do you guys think @poire-z @NiLuJe @pazos?

(If you just keep it as a local here the logic should be fine, but it feels like it could have some generic use.)

frontend/device/remarkable/device.lua
```diff
@@ -125,5 +126,159 @@ function Remarkable:reboot()
...
+                ssid = getNetworkProperty(path, "ssid"),
+            })
+        end
+        return results
+    end
+    function NetworkMgr:authenticateNetwork(network)
+        local InfoMessage = require("ui/widget/infomessage")
+        local UIManager = require("ui/uimanager")
+        local BD = require("ui/bidi")
+        local info = InfoMessage:new{text = "Connecting to " .. BD.wrap(network.ssid) }
```
Frenzie, line 265: Is this a temporary debugging aid? The interface has some `Connecting to Wi-Fi…` messages sprinkled around.

If not, it should be something like:

```suggestion
        local info = InfoMessage:new{text = T(_("Connecting to %1.", BD.wrap(network.ssid))) }
```

Eeems: When I was trying to use the template it was never replacing the value.

Frenzie: If you mean specifically what I wrote, that'd be because I typod.

Should be:
```lua
local info = InfoMessage:new{text = T(_("Connecting to %1."), BD.wrap(network.ssid)) }
```

frontend/device/remarkable/device.lua
```diff
@@ -125,5 +126,159 @@ function Remarkable:reboot()
...
+    if raw then
+        return s
+    end
+    s = string.gsub(s, '^%s+', '')
+    s = string.gsub(s, '%s+$', '')
+    s = string.gsub(s, '[\n\r]+', ' ')
+    return s
+end
+
+function getNetworkProperty(path, name)
```
Frenzie, line 148: Yes, these should all be local.

frontend/device/remarkable/device.lua
poire-z: > but it seems to me that the unfiltered output is what you'd expect by default. 

To me too. And it looks like with how it is used here, it should be fine with leading spaces or `\n`: tonumber() is fine with them, I guess it's still valid json so rapidjson should be fine too. May be just have isempty() checks also for `"\n"`.

> it feels like it could have some generic use

Well, we might want to strip some but not all, so, that would need many parameters :) I prefer the strippings done locally, so it can be more easily customized (and we won't go adding one parameter when we need to strip something else :)

Frenzie: So something like this? Or keep it completely local here (but still in roughly this form).

```lua
function util.osCapture(cmd)
    local f = io.popen(cmd, "r")
    if not f then
        return nil
    end
    local s = f:read("*a")
    f:close()
    if not s then
        return nil
    end
    return s
end
```

pazos: util.stdout?

Frenzie: Or `util.somethingPopen`/`util.popenSomething`.

Eeems:
I'll work through the build errors now.

frontend/device/remarkable/device.lua
```diff
@@ -125,5 +126,161 @@ function Remarkable:reboot()
...
+    local json, err = rapidjson.decode(osCapture("rot wifi get " .. name))
+    return json
+end
+local function isempty(s)
+  return s == nil or s == ''
+end
+
+-- wireless
+function Remarkable:initNetworkManager(NetworkMgr)
+    if isempty(osCapture("rot")) then
```
pazos, line 168: if `rot` is an executable and its path is always the same I would like to check if exists using `util.fileExists("/path/to/rot")` and perform the check once at device init. This makes possible to to decide which wireless backend is going to be used at runtime.

frontend/device/remarkable/device.lua
```diff
@@ -125,5 +126,161 @@ function Remarkable:reboot()
...
+    s = string.gsub(s, '^%s+', '')
+    s = string.gsub(s, '%s+$', '')
+    s = string.gsub(s, '[\n\r]+', ' ')
+    return s
+end
+
+local function getNetworkProperty(path, name)
+    path = string.sub(path, string.len("/codes/eeems/oxide1/") + 1)
+    local json, err = rapidjson.decode(osCapture("rot --object Network:" .. path .. " wifi get " .. name))
+    return json
```
Frenzie, line 151: You could simply pass it on. Or you could write the above as `local json = rapidjson.decode(osCapture("rot --object Network:" .. path .. " wifi get " .. name))`
```suggestion
    return json, err
```

frontend/device/remarkable/device.lua
Frenzie: Or just os.execute("which rot")

1 = doesn't exists, 0 = exists

frontend/device/remarkable/device.lua
poire-z: `util.getCommandOutput(cmd, stripped)`
(with may be `stripped=true/false` just to remove leading and trailing spaces)

The first 3 lines are not needed in there, are they ?
```
    if not s then
        return nil
    end
    return s
```

poire-z: Might as well have util.getFileContent(path) as we have a few things like:
https://github.com/koreader/koreader/blob/27fe0d7957fd48c375cdae48bc3bf30ceef1edeb/frontend/apps/reader/modules/readerdictionary.lua#L71-L80

Frenzie: I don't know, depends on if io.popen could return `false` and if we would want to hide that. But indeed it looks unnecessary.

frontend/device/remarkable/device.lua
```diff
@@ -125,5 +129,220 @@ function Remarkable:reboot()
...
+        if complete_callback then
+            complete_callback()
+        end
+    end
+    function NetworkMgr:turnOnWifi(complete_callback)
+        logger.info("Remarkable: enabling Wi-Fi")
+        os.execute("rot wifi call enable")
+        local UIManager = require("ui/uimanager")
+        local InfoMessage = require("ui/widget/infomessage")
+        local info = InfoMessage:new{text = _("Turning on wifi…")}
```
Frenzie, line 184: The message already elsewhere in the code is:
```suggestion
        local info = InfoMessage:new{text = _("Connecting to Wi-Fi…")}
```

Frenzie: My bad, I was just skimming. In any case, unfortunately our current house style is Wi-Fi.

Eeems: Right, but I show another message `"Scanning for networks…"` after wifi properly enables. Should I just say `"Connecting to Wi-Fi..."` and not report the differences?

Frenzie: I'm not sure if a message should really be shown only to be replaced by another in a split second, but turning on wifi is theoretically fine then.

pazos (commented):
Looks good.

I would like to move everything `oxide` to its own file (except `which oxide` check). Something like

### local function to check if the launcher is installed

```lua
local function hasOxide()
    return os.execute("which rot") == 0
end
```

### on network init

```lua
    local ok_oxide, oxide = hasOxide()
    if ok_oxide then
        oxide = require("device/remarkable/oxide")
        -- require here oxide specific functions to extend NetworkMgr
        function NetworkMgr:getCurrentNetwork()
            return oxide:getCurrentNetwork()
        end
        function NetworkMgr:getNetworkList()
            return oxide:getNetworkList()
        end
        function NetworkMgr:authenticateNetwork(network)
            return oxide:authenticateNetwork(network)
        end
        function NetworkMgr:disconnectNetwork(network)
            oxide:disconnectNetwork()
        end
    end

    -- here functions that are common for oxide/wpa_supplicant like obtain/releaseIP and getNetworkInterfaceName.

    -- here functions the are expected to be implemented on device, for example:
    function NetworkMgr:turnOffWifi(complete_callback)
        if ok_oxide then
            oxide:turnOffWifi()
       else
            -- leave unimplemented if you want
       end
       -- finally execute the callback
       if complete_callback then
           complete_callback()
       end
   end
```

pazos (commented):
Looks good.

I would like to move everything `oxide` to its own file (except `which oxide` check). Something like

### local function to check if the launcher is installed

```lua
local function hasOxide()
    return os.execute("which rot") == 0
end
```

### on network init

```lua
    local ok_oxide, oxide = hasOxide()
    if ok_oxide then
        oxide = require("device/remarkable/oxide")
        -- require here oxide specific functions to extend NetworkMgr
        function NetworkMgr:getCurrentNetwork()
            return oxide:getCurrentNetwork()
        end
        function NetworkMgr:getNetworkList()
            return oxide:getNetworkList()
        end
        function NetworkMgr:authenticateNetwork(network)
            return oxide:authenticateNetwork(network)
        end
        function NetworkMgr:disconnectNetwork(network)
            oxide:disconnectNetwork()
        end
    end

    -- here functions that are common for oxide/wpa_supplicant like obtain/releaseIP and getNetworkInterfaceName.

    -- here functions the are expected to be implemented on device, for example:
    function NetworkMgr:turnOffWifi(complete_callback)
        if ok_oxide then
            oxide:turnOffWifi()
       else
            -- leave unimplemented if you want
       end
       -- finally execute the callback
       if complete_callback then
           complete_callback()
       end
   end
```

Eeems:
Doing some work on oxide at the moment, but I'll get back to this at some point later this week.

Witos:
@Eeems , it crashes for me if "rot" is not available:
sh: rot: command not found
./luajit: frontend/device/remarkable/device.lua:233: attempt to compare number with nil
stack traceback:
	frontend/device/remarkable/device.lua:233: in function 'isWifiOn'
	frontend/ui/network/manager.lua:90: in function 'init'
	frontend/ui/network/manager.lua:657: in main chunk
	[C]: in function 'require'
	frontend/apps/reader/modules/readerdictionary.lua:14: in main chunk
	[C]: in function 'require'
	frontend/apps/reader/readerui.lua:31: in main chunk
	[C]: in function 'require'
	./reader.lua:263: in main chunk
	[C]: at 0x0001415d
!!!!

Don't take the lines number into account, the crash is in isWifiOn.

Eeems:
@Witos Are you sure you are using the latest commit? The line it's failing on (frontend/device/remarkable/device.lua:233) is commented out code in the latest commit.

Witos:
@Eeems, yes, do not take line numbers into account (as I wrote in my previous post), the file contained changes irrelevant to wifi logic. I can generate the crash with your version, but it still be inside isWifiOn.

Eeems:
@Eeems which is still od as https://github.com/koreader/koreader/pull/6681/files#diff-1640ae3ecad5211dafcbe28fa71bac211342288d3dafd4280d68491598891cafR168 checks and exits early if rot doesn't exist. Which would keep isWifiOn from being defined.

pazos:
@Eeems: I'm going to close this PR. Now we have direct wifi support on the Remarkable. Feel free to reopen if you still want to make improvements/integrations with Oxide.

PR #6684, opened 2020-09-19
Pocketbook: Implement HW screen rotation, bring in PB specific kludges from linuxfb

ezdiy:
Needs https://github.com/koreader/koreader-base/pull/1194

* Uses bunch of new plumbing in base to configure screen rotations in hardware (koreader normally does this via blit buffer rotations, except for android).

* Some PB specific kludges that used to pollute core/framebuffer_linux are brought into PBs frontend driver.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6684)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -272,10 +272,10 @@ end
 
 -- Pocketbook HW rotation modes start from landsape, CCW
 local landscape_ccw = {
-  1, 0, 3, 2,         -- PORTRAIT, LANDSCAPE, PORTRAIT_180, LANDSCAPE_180
-  every_paint = true, -- inkview will try to steal the rot mode frequently
-  restore = false,    -- no need, because everything using inkview forces 3 on focus
-  default = nil,      -- usually 3
+    1, 0, 3, 2,         -- PORTRAIT, LANDSCAPE, PORTRAIT_180, LANDSCAPE_180
+    every_paint = true, -- inkview will try to steal the rot mode frequently
+    restore = false,    -- no need, because everything using inkview forces 3 on focus
+    default = nil,      -- usually 3
 }
 
 -- PocketBook Mini (515)
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -39,6 +39,10 @@ local PocketBook = Generic:new{
 
     -- Apparently, HW inversion is a pipedream on PB (#6669)
     canHWInvert = no,
+
+    -- Private per-model kludges
+    _fb_init = function() end,
+    _model_init = function() end,
 }
 
 -- Make sure the C BB cannot be used on devices with a 24bpp fb
@@ -75,7 +79,13 @@ function PocketBook:init()
     -- Blacklist the C BB before the first BB require...
     self:blacklistCBB()
 
-    self.screen = require("ffi/framebuffer_mxcfb"):new{device = self, debug = logger.dbg}
+    self.screen = require("ffi/framebuffer_mxcfb"):new {
+        device = self,
+        debug = logger.dbg,
+        is_always_portrait = self.isAlwaysPortrait(),
+        forced_rotation = self.usingForcedRotation,
+        fbinfoOverride = self._fb_init,
+    }
     self.powerd = require("device/pocketbook/powerd"):new{device = self}
 
     -- Whenever we lose focus, but also get suspended for real (we can't reliably tell atm),
@@ -143,12 +153,7 @@ function PocketBook:init()
         end
     end)
 
-    -- fix rotation for Color Lux device
-    if PocketBook:getDeviceModel() == "PocketBook Color Lux" then
-        self.screen.blitbuffer_rotation_mode = self.screen.ORIENTATION_PORTRAIT
-        self.screen.native_rotation_mode = self.screen.ORIENTATION_PORTRAIT
-    end
-
+    self._model_init()
     self.input.open()
     self:setAutoStandby(true)
     Generic.init(self)
@@ -265,6 +270,14 @@ function PocketBook:getDeviceModel()
     return ffi.string(inkview.GetDeviceModel())
 end
 
+-- Pocketbook HW rotation modes start from landsape, CCW
+local landscape_ccw = {
+  1, 0, 3, 2,         -- PORTRAIT, LANDSCAPE, PORTRAIT_180, LANDSCAPE_180
+  every_paint = true, -- inkview will try to steal the rot mode frequently
+  restore = false,    -- no need, because everything using inkview forces 3 on focus
+  default = nil,      -- usually 3
+}
+
 -- PocketBook Mini (515)
 local PocketBook515 = PocketBook:new{
     model = "PB515",
@@ -377,7 +390,7 @@ local PocketBook627 = PocketBook:new{
 local PocketBook628 = PocketBook:new{
     model = "PBTouchLux5",
     display_dpi = 212,
-    isAlwaysPortrait = yes,
+    usingForcedRotation = landscape_ccw,
     hasNaturalLight = yes,
 }
 
@@ -399,7 +412,7 @@ local PocketBook631 = PocketBook:new{
 local PocketBook632 = PocketBook:new{
     model = "PBTouchHDPlus",
     display_dpi = 300,
-    isAlwaysPortrait = yes,
+    usingForcedRotation = landscape_ccw,
     hasNaturalLight = yes,
 }
 
@@ -409,7 +422,7 @@ local PocketBook633 = PocketBook:new{
     display_dpi = 300,
     hasColorScreen = yes,
     canUseCBB = no, -- 24bpp
-    isAlwaysPortrait = yes,
+    usingForcedRotation = landscape_ccw,
 }
 
 -- PocketBook Aqua (640)
@@ -434,7 +447,7 @@ local PocketBook650 = PocketBook:new{
 local PocketBook740 = PocketBook:new{
     model = "PBInkPad3",
     display_dpi = 300,
-    isAlwaysPortrait = yes,
+    usingForcedRotation = landscape_ccw,
     hasNaturalLight = yes,
 }
 
@@ -442,7 +455,7 @@ local PocketBook740 = PocketBook:new{
 local PocketBook740_2 = PocketBook:new{
     model = "PBInkPad3Pro",
     display_dpi = 300,
-    isAlwaysPortrait = yes,
+    usingForcedRotation = landscape_ccw,
     hasNaturalLight = yes,
 }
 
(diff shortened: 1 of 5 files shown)


Frenzie (approved):
I think @NiLuJe touched the rotation stuff last.

NiLuJe:
At a very quick glance, I'd say CRe remembers the fb at document creation in order to deal with its own internal blitbuffer properly, probably around here:

https://github.com/koreader/koreader/blob/27fe0d7957fd48c375cdae48bc3bf30ceef1edeb/frontend/document/credocument.lua#L148

Then again, don't we already have some platforms that re-init the framebuffer? SDL? Android?

NiLuJe:
If not, I'd usually blame a bogus `line_length` for rendering like that... ;).

frontend/ui/uimanager.lua
```diff
@@ -1062,6 +1062,7 @@ function UIManager:_repaint()
             -- pass hint to widget that we got when setting widget dirty
             -- the widget can use this to decide which parts should be refreshed
             logger.dbg("painting widget:", widget.widget.name or widget.widget.id or tostring(widget))
+            Screen:forceRotation()
```
NiLuJe, line 1065: I'd ideally prefer if we could (somehow ^^) ensure that either `forceRotation` or the actual `ioctl` only get called *once* per `_repaint`/stack, instead of before each widget.

Which is easier said than done ^^.

NiLuJe: Oh, you did, that's the `_forced_rotation` check, right?

(It's late, I haven't slept much,, I *really* shouldn't be trying to review this right now, sorry for the dumb comments ^^).

Frenzie:
> Then again, don't we already have some platforms that re-init the framebuffer at runtime? SDL? Android? Isn't this already broken there? And/Or how does it work there?

Yes, otherwise you'd get nicely stretched pixels on SDL. You can see in front how I cobbled together a few events for that.

ezdiy:
> At a very quick glance, I'd say CRe remembers the fb state at document creation in order to deal with its own internal blitbuffer properly, probably around here:
> 
> https://github.com/koreader/koreader/blob/27fe0d7957fd48c375cdae48bc3bf30ceef1edeb/frontend/document/credocument.lua#L148
> 
> Then again, don't we already have some platforms that re-init the framebuffer at runtime? SDL? Android? Isn't this already broken there? And/Or how does it work there?

Good idea with android.

```
-- resize on rotation or split view.
function framebuffer:resize()
    android.screen.width = android.getScreenWidth()
    android.screen.height = android.getScreenHeight()
    self.bb:free()
    self.bb = BB.new(android.screen.width, android.screen.height, BB.TYPE_BBRGB32)
    self.invert_bb = BB.new(android.screen.width, android.screen.height, BB.TYPE_BBRGB32)
    self.bb:fill(BB.COLOR_WHITE)
    self:_updateWindow()
end
```
This one is called *later* when OS even arrives, in addition to emitting:

```
                    UIManager:broadcastEvent(Event:new("SetDimensions", new_size))
                    UIManager:broadcastEvent(Event:new("ScreenResize", new_size))
                    UIManager:broadcastEvent(Event:new("RedrawCurrentPage"))
```
in frontend. So I'll try to emulate what android does.

Frenzie:
@ezdiy Should you happen to mention that in any comments, please refer to SDL for clarity since Android is just copying SDL. ;-)

ezdiy:
@Frenzie Not quite. From what I gather, the SDL driver does *not* use any sort of "hw" rotation, it relies solely on the blitter doing the rotation. Which is in large part source of my confusion - how can this work at all? Of course the underlying bb swaps W/H in accordance of setRotation - but so do we (we just reinstante the bb anew at "0 rotation", but with W/H swapped regardless from HW driver).

Frenzie:
@ezdiy You'd see the exact same type of corruption without those (quick somewhat hackish) workarounds. You're drawing a distinction between resizing and rotating that unless I'm overlooking something PB-specific just isn't there.

ezdiy:
The corrupted screen *is* the result of wrong line pitch though. Meaning something should be told new W, and isn't, or it misinterprets H. Other than that, the document is always oriented correctly.

ezdiy:
```
         self:forceRotation()
         self:reinit()
         self:restoreRotation()
+        local Event = require("ui/event")
+        local new_size = self:getSize()
+        UIManager:broadcastEvent(Event:new("SetDimensions", new_size))
+        UIManager:broadcastEvent(Event:new("ScreenResize", new_size))
+        UIManager:broadcastEvent(Event:new("RedrawCurrentPage"))
```

Did nilch. Besides, cre does get the memo on this level:

```
09/20/20-19:24:30 DEBUG CreDocument: set view dimen {
    ["w"] = 1404,
    ["h"] = 1872
}
```

Even if I don't stuff the UI events. @NiLuJe got it right however. It's not convas context either, that one is given correct dimensions too:

```
     local ok
+    logger.dbg("CreDocument:init(): ", CanvasContext:getSize())
     ok, self._document = pcall(cre.newDocView, CanvasContext:getWidth(), CanvasContext:getHeight(), self._view_mode)
     if not ok then

09/20/20-19:39:59 DEBUG CreDocument:init():  {
    ["h"] = 1872,
    ["w"] = 1404
}
```

NiLuJe:
I'd double-check the [various fast-paths](https://github.com/koreader/koreader/pull/6684#issuecomment-695369741) to see if something's going wrong somewhere.

I don't think I left commented out stuff in the Lua blitter, but I'm fairly sure the C one has commented out logging that prints which type of codepath gets invoked.

poire-z:
Beware with cre, and the w/h you give to it: it might, depending on the functions that may be called (even if not yet repainting anything) notice a view size change and decide to re-render the whole document for the new dimensions.
So, try your stuff with viewing a big EPUB, so you can notice a pause while it is re-rendering (there's no explicite log, it does it internally when needed).
Obviously, it would be preferable if it never gets any view size change if it is not really user view change that legitimately requires a re-rendering for the new view.

NiLuJe:
@poire-z: Yeah, even without HW rotation, that's already visible when toggling between Landscape/Portrait (vs. just Inverting your current orientation).

ezdiy:
I think we're too harsh on cre here. Exact same thing happens with mupdf ('pdfdocument.lua'). Yet CBZ viewed through mupdf (same handler) display (mostly) fine - until I enable continues mode when things get staircased, but only on page boundary. As if things weren't weird enough, mupdf can display the epub just fine on top of all this:

```
-require("document/credocument"):register(DocumentRegistry)
+--require("document/credocument"):register(DocumentRegistry)
 require("document/pdfdocument"):register(DocumentRegistry)
-require("document/djvudocument"):register(DocumentRegistry)
-require("document/picdocument"):register(DocumentRegistry)
+--require("document/djvudocument"):register(DocumentRegistry)
+--require("document/picdocument"):register(DocumentRegistry)
```

ezdiy:
@NiLuJe You were right:

```
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -1020,7 +1020,7 @@ BB_mt.__index.blitToRGB32 = BB_mt.__index.blitDefault
 -- Same to same fast blitting
 function BB8_mt.__index:blitTo8(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     -- We can only do fast copy for simple blitting with no processing (setPixel, no rota, no invert)
-    if setter ~= self.setPixel or self:getRotation() ~= 0 or dest:getRotation() ~= 0 or (self:getInverse() ~= dest:getInverse()) then
+    if true or setter ~= self.setPixel or self:getRotation() ~= 0 or dest:getRotation() ~= 0 or (self:getInverse() ~= dest:getInverse()) then
         return self:blitDefault(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     end
```

Fixes it. This raises more questions though.

poire-z:
> I think we're too harsh on cre here

I'm not talking about blitbuffers getting corrupt, more about:
- painting a page (set fb to canvas w/h, paint, restore fb to OS w/h, tell OS w/h to cre)
- do a dict lookup, full text search, bookmark page: this might happen while cre think the document has OS w/h: it might give wrong results, or causes cre to decide to re-render (and if that doesn't happen, it may be because of race conditions, because a menu refresh caused canvas w/h to have been provided to cre while the cre other methods were happening :)

ezdiy:
> painting a page (set fb to canvas w/h, paint, restore fb to OS w/h, tell OS w/h to cre)

The OS rotation is masked completely from rest of code. The only rotation user code sees is one that is asked for *whenever* it starts painting (assuming uimanager ui:paintTo is pen-ultimate entry for that). The race may occur indeed if something actually flips OS hwrot behind our back (but we'll switch it back soon enough on next draw so ... meh). The race cannt manifest on pocketbook because whenever OS wants to paint over us, they blank the FB anyway - and then tell us via event that we got focus back (during which we force repaint, thus as re-set of hwrot).

NiLuJe (approved):

ezdiy:
This one should go next in precedence to #6696, so as to resolve the minor merge conflict this one will introduce in there.

frontend/device/pocketbook/device.lua
```diff
@@ -265,6 +270,14 @@ function PocketBook:getDeviceModel()
     return ffi.string(inkview.GetDeviceModel())
 end
 
+-- Pocketbook HW rotation modes start from landsape, CCW
+local landscape_ccw = {
+  1, 0, 3, 2,         -- PORTRAIT, LANDSCAPE, PORTRAIT_180, LANDSCAPE_180
+  every_paint = true, -- inkview will try to steal the rot mode frequently
+  restore = false,    -- no need, because everything using inkview forces 3 on focus
+  default = nil,      -- usually 3
```
Frenzie, line 278:
```suggestion
    1, 0, 3, 2,         -- PORTRAIT, LANDSCAPE, PORTRAIT_180, LANDSCAPE_180
    every_paint = true, -- inkview will try to steal the rot mode frequently
    restore = false,    -- no need, because everything using inkview forces 3 on focus
    default = nil,      -- usually 3
```

Frenzie (commented):
nitpick

Frenzie:
You probably need to fake that `getTouchRotation()`.
```
[  ERROR   ] 4 errors, listed below:
[  ERROR   ] spec/front/unit/gesturedetector_spec.lua @ 26: gesturedetector module adjustGesCoordinate should not translate rotation 0
[  ERROR   ] spec/front/unit/gesturedetector_spec.lua @ 33: gesturedetector module adjustGesCoordinate should translate rotation 270
[  ERROR   ] spec/front/unit/gesturedetector_spec.lua @ 40: gesturedetector module adjustGesCoordinate should translate rotation 180
[  ERROR   ] spec/front/unit/gesturedetector_spec.lua @ 47: gesturedetector module adjustGesCoordinate should translate rotation 90
```

PR #6685, opened 2020-09-19
reMarkable: add environment variable to tell koreader.sh to not set bpp

raisjnn:
related to https://github.com/Eeems/oxide/issues/69, this diff adds an env variable to remarkable platform's koreader.sh to allow someone to tell koreader to not set the bpp to 8 from outside koreader.sh. the env variable is called `KO_DONT_SET_DEPTH`

we don't use the dev setting because someone might change the dev setting without realizing what it does

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6685)

### DIFF
--- a/platform/remarkable/koreader.sh
+++ b/platform/remarkable/koreader.sh
@@ -97,7 +97,7 @@ esac
 # The actual swap is done in a function, because we can disable it in the Developer settings, and we want to honor it on restart.
 ko_do_fbdepth() {
     if [ -n "${KO_DONT_SET_DEPTH}" ]; then
-      return
+        return
     fi
 
     # Check if the swap has been disabled...
--- a/platform/remarkable/koreader.sh
+++ b/platform/remarkable/koreader.sh
@@ -96,6 +96,10 @@ esac
 
 # The actual swap is done in a function, because we can disable it in the Developer settings, and we want to honor it on restart.
 ko_do_fbdepth() {
+    if [ -n "${KO_DONT_SET_DEPTH}" ]; then
+      return
+    fi
+
     # Check if the swap has been disabled...
     if grep -q '\["dev_startup_no_fbdepth"\] = true' 'settings.reader.lua' 2>/dev/null; then
         # Swap back to the original bitdepth (in case this was a restart)


platform/remarkable/koreader.sh
```diff
@@ -96,6 +96,10 @@ esac
 
 # The actual swap is done in a function, because we can disable it in the Developer settings, and we want to honor it on restart.
 ko_do_fbdepth() {
+    if [ -n "${KO_DONT_SET_DEPTH}" ]; then
+      return
```
Frenzie, line 100:
```suggestion
        return
```

Frenzie (approved):
Seems fine, also pinging @NiLuJe

NiLuJe:
I was under the impression that this wouldn't be necessary given https://github.com/Eeems/oxide/issues/69#issuecomment-695145631 ;).

(Because, I mean, you *really* don't want a RGB565 fb/bb when you can avoid it, really, it's just the worst ;)).

But, barring that, yeah, that works ^^.

raisjnn:
> (Because, I mean, you _really_ don't want a RGB565 fb/bb when you can avoid it, really, it's just the worst ;)).

you've mentioned this twice and it's also written in the koreader.sh file, i'm curious if you can elaborate on why or point me to some tasks / commits about it. is it because of the read or write path while accessing framebuffer? is it because refreshing eink display is slower when in rgb565? or...?

NiLuJe:
RGB565 as a pixel-format is dumb as rocks when you're on a grayscale device, to begin with, and it's even more stupid when you don't have hardware blitting.

Qt5 itself jumps through a billion of shitty hoops to try to hide the pain, because it was sponsored to handle those crappy pixel-formats in the early days of mobile on ARMv7 + NEON, but even Kobo finally realized than eating the bandwidth cost of moving from 16bpp to 32bpp turned out to be a win regardless.

So, when you *can* actually go down to 8bpp (which isn't necessarily a win as far as Qt5 is concerned, because Y8 is much less of a first rate citizen as a QPainter target), everything becomes rainbows and unicorns: a pixel format that makes sense for the device, much tinier bandwidth requirement, no shitty computations involved, which just makes it plain faster ;).

raisjnn:
I'd like to see some performance analysis of this, tbh  - even something as simple as "drawing on kindle takes X ms in 16bpp and Y in 8bpp". This seems like it is specifically in the write path of the framebuffer which is not exercised frequently on an eink device, so I wonder how many milliseconds it is adding to the write path (i know it's device specific)

You can still memcpy rgb565 from a virtual framebuffer to real framebuffer (and maybe even zero copy?). If you are drawing in full grayscale, I understand why this might be slower (converting between rgb and grayscale) but still don't think its end of world. 

also, while the remarkable has QT support and there's a proprietary eink plugin for fast refreshes, there are three alternate libraries (one in rust, one in cpp and one in .NET) for drawing to framebuffer with low latency. i personally stay away from QT. 

the remarkable is well known for its low latency writing experience and it is in rgb565 (iirc).  it does this by only drawing black/white pixels (no grayscale) when writing with the stylus, so its likely rgb565 isn't coming into consideration since its just an assignment operation to draw a pixel (and no shifts involved). 

the reason for not wanting to drop to 8bpp (for me) is because our launchers take screenshots when we switch away from an app, then when we return we restore the screenshot and poke the app awake. when an app is using 8bpp, the launcher will have to do scaling or be aware of the bpp and adjust the screenshot restoration appropriately.

raisjnn:
also, i'm not bothered by the rgb565 code you linked - i've been to the depths of organizations of thousands of engineers and witnessed the horrors that lurk within.

NiLuJe:
The pixelformat has (nearly) zero impact on the eInk stuff itself on Freescale HW: that's handled by the PxP, and it's generally very good at doing what it needs to do as fast as possible regardless.

That doesn't mean that there isn't still an extra fast-path for untransformed (!rotated, !inverted, !dithered) 8bpp fbs in the EPDC: there is ^^.

So, the main pain-point is the actual blitting. RGB565 starts on the wrong foot because it's not a common "source" format. Your image assets are going to be RGB32, not RGB565.

If you can get parts of your pipeline to use RGB565 blitbuffers, that helps, but that's not always possible.

Qt5 hides most of the pain involved in format-shifting with a lot of hand-crafted NEON assembly. But in some case, it's still faster to not have to do that and just take twice as much memory to store your pixels by jumping to RGB32.

NiLuJe:
To drop hard numbers on it, for CRe+KOReader, not having access to a fast-path is the difference between a 3ms blit and a 50 to 100ms one ;).

NiLuJe:
CRe itself *can* render to a RGB565 blitbuffer, so, someone motivated enough could implement some of those fast paths for RGB565. But that's the only one of our renderer that can do that, and on every platform that's natively RGB565, we can switch to Y8, so it was less effort and more performance to go that route ;).

NiLuJe:
As a very quick and not terribly interesting benchmark, because I do a lot of weird crap in the image codepaths in FBInk, and there is absolutely no NEON involved:

Decoding + blitting an RGB32 (to avoid the Y8 -> Y8 fast-path ;p) full-screen image on an H2O (1080x1440):

* 83ms @ 8bpp
* 142ms @ 16bpp
* 122ms @ 32bpp

raisjnn:
great, thank you! so my understanding (from the above) is that CRe is rendering engine for KO and the paths inside it are slower with rgb565 because source material is rgb32, which requires an extra pass over data and transformations that are slow to go from rgb32 -> rgb565. but when transforming from rgb32 -> Y8, it uses a faster transform path and it has to copy less memory over (and maybe has better cache properties) and has optimizations for faster display in EDPC driver. when going from rgb32 -> rgb32, it skips a transform (full pass over data). 

did i miss anything? (did i get stuff wrong?)

NiLuJe:
CRe (the main reflowable document engine, i.e., ePub) just happens to be the *one* thing in our pipeline that *internally* knows how to deal with it and spit out an RGB565 blitbuffer. Even in these cases, it's still mainly assuming RGB32 internally, though.

Whereas if it's flipped to Grayscale mode, it's Y8 all the way internally (well, except for image decoding, which is still mostly always RGB32 inside CRe).

But that's the only thing that does so. MuPDF, DjVuLibre, image decoders and all that jazz couldn't care less about RGB565, while most of them can do full Y8 pipelines ;).

Our UI toolkit was originally developed on Kindle, so it's pretty much geared towards Grayscale as truth, too.

So, yeah, while we could have technically dealt with CRe @ RGB565 slightly better, the effort wasn't worth it when everything's just plain better all around @ Y8 ;).

NiLuJe:
The main interest is basically having as few pixel-format shifts as possible in the pipeline, and, if you can do that in a format that match the actual framebuffer's, all the better.

Which is why we do that @ Y8 on eInk if possible, and @ RGB32 on the rest if possible.

NiLuJe:
I *may* be a wee bit of a one-man crusade as far as the "RGB565 is terrible and it needs to die" thing ^^.

NiLuJe:
Sidebar: where RGB32 is involved, stuff can also get a bit anoying when stuff doesn't agree on BGRA vs. RGBA, although that's a more common issue, so there are generally not-too-terrible ways to deal with it ;).

PR #6686, opened 2020-09-19
[PB] Crash screen & OTA feedback

NiLuJe:
Much like on other platforms w/ an FBInk port ;).

Depends on https://github.com/koreader/koreader-base/pull/1192

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6686)

### DIFF
--- a/platform/pocketbook/koreader.app
+++ b/platform/pocketbook/koreader.app
@@ -151,11 +151,7 @@ while [ ${RETURN_VALUE} -ne 0 ]; do
 
         # Pause a bit if it's the first crash in a while, so that it actually has a chance of getting noticed ;).
         if [ ${CRASH_COUNT} -eq 1 ]; then
-            # NOTE: We don't actually care about what read read, we're just using it as a fancy sleep ;).
-            #       i.e., we pause either until the 15s timeout, or until the user touches the screen.
-            # FIXME: That probably won't work on PB? Just stick a sleep 15 here?
-            # shellcheck disable=SC2039
-            read -r -t 15 </dev/input/event1
+            sleep 15
         fi
         # Cycle the last crash timestamp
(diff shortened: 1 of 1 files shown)


NiLuJe:
Untested ;p.

If you require FBInk binaries to help test this, they're [here](https://github.com/NiLuJe/FBInk/pull/47#issuecomment-695152346) ;).

Frenzie (approved):
I assume the shell script is the same as on Kobo and it's really all about FBInk? ^_^

NiLuJe:
Essentially, yes ;).

There are a few platform-specific stuff done differently, but hopefully I didn't botch that too much ;p.

platform/pocketbook/koreader.app
```diff
@@ -76,11 +78,108 @@ if [ -e crash.log ]; then
...
+        ko_update_check
+    fi
+
     ./reader.lua "${args}" >>crash.log 2>&1
-    RETURN_VALUE=$(cat ${KO_EXIT_CODE})
+    RETURN_VALUE="$(cat ${KO_EXIT_CODE})"
+    rm -f "${KO_EXIT_CODE}"
+
+    # Did we crash?
+    if [ ${RETURN_VALUE} -ne 0 ] && [ ${RETURN_VALUE} -ne ${KO_RC_RESTART} ]; then
```
ezdiy: On  crash, the file in ${KO_EXIT_CODE} will be always empty, so perhaps one could just do `if [ -e ${KO_EXIT_CODE} ]`. While these conditions will hold true in the empty case, it's more of a side-effect fluke than stating what really goes on explicitly.

NiLuJe: Yeah, good point ;).

NiLuJe:
Let's break all the things! (hopefully not ^^).

PR #6693, opened 2020-09-21
Fix point release and revision detection in Version:getShortVersion

NiLuJe:
Turns out match returns an empty string instead of nil for unmatched groups?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6693)


pazos (approved):

Frenzie:
What was the error string? For https://github.com/koreader/koreader/blob/221b71062a9c0d330b3b8d9ff5f54402d4a4be64/spec/unit/version_spec.lua

NiLuJe:
The latest stable, i.e., v2020.09 ;).

Frenzie:
Oh wait, you added this function and I didn't complain about or missed the lack of tests with the usual suspects! :-P

PR #6694, opened 2020-09-21
Dispatcher: crengine contrast (gamma)

yparitcher:
It would crashing when i tried to enable it originally, but we must have fixed it with some other bugfix along the way :)
___
 Gestures/Profiles: make sure dispatcher is initalized even when the menu is not opened. this was causing a crash when the event name was parsed from creoptions

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6694)


yparitcher:
please do not squash, 2 separate fixes

PR #6696, opened 2020-09-22
Clean up C blitbuffer kludges.

ezdiy:
Needs https://github.com/koreader/koreader-base/pull/1195

CBB now handles nightmode correctly (by deferring to Lua), so we no longer
need to do monkey dances about disabling it when hw invert is missing.

canUseCBB cap is resolved by generic device re-configuring blitbuffer
on the go, so as to avoid repeating the same thing in every device driver.

The dev setting can now flip cbb on the go, so one can gloat at the near
meaningless perf difference - 2Mp draw is 15ms Lua / 10ms C on 1GHz Cortex A7.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6696)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -332,18 +332,18 @@ function FileManagerMenu:setUpdateItemTable()
         })
     end
     if not Device.should_restrict_JIT then
-        local bb = require("ffi/blitbuffer")
+        local Blitbuffer = require("ffi/blitbuffer")
         table.insert(self.menu_items.developer_options.sub_item_table, {
             text = _("Disable C blitter"),
             enabled_func = function()
-                return bb.has_cblitbuffer
+                return Blitbuffer.has_cblitbuffer
             end,
             checked_func = function()
                 return G_reader_settings:isTrue("dev_no_c_blitter")
             end,
             callback = function()
                 G_reader_settings:flipNilOrFalse("dev_no_c_blitter")
-                bb:enableCBB(G_reader_settings:nilOrFalse("dev_no_c_blitter"))
+                Blitbuffer:enableCBB(G_reader_settings:nilOrFalse("dev_no_c_blitter"))
             end,
         })
     end
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -332,20 +332,18 @@ function FileManagerMenu:setUpdateItemTable()
         })
     end
     if not Device.should_restrict_JIT then
+        local bb = require("ffi/blitbuffer")
         table.insert(self.menu_items.developer_options.sub_item_table, {
             text = _("Disable C blitter"),
             enabled_func = function()
-                return lfs.attributes("libs/libblitbuffer.so", "mode") == "file"
+                return bb.has_cblitbuffer
             end,
             checked_func = function()
                 return G_reader_settings:isTrue("dev_no_c_blitter")
             end,
             callback = function()
                 G_reader_settings:flipNilOrFalse("dev_no_c_blitter")
-                local InfoMessage = require("ui/widget/infomessage")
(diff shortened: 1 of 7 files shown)


frontend/device/kindle/device.lua
```diff
@@ -258,25 +258,6 @@ function Kindle:ambientBrightnessLevel()
     return 4
 end
 
---- Makes sure the C BB cannot be used on devices with a 4bpp fb.
-function Kindle:blacklistCBB()
```
Frenzie, line 262: I was momentarily confused by these removals without apparent additions in return but the relevant device definitions already specify `canUseCBB = no`. ^_^

Frenzie (approved):
Nice!

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -332,20 +332,17 @@ function FileManagerMenu:setUpdateItemTable()
...
             checked_func = function()
                 return G_reader_settings:isTrue("dev_no_c_blitter")
             end,
             callback = function()
-                G_reader_settings:flipNilOrFalse("dev_no_c_blitter")
-                local InfoMessage = require("ui/widget/infomessage")
-                UIManager:show(InfoMessage:new{
-                    text = _("This will take effect on next restart."),
-                })
+                bb:enableCBB(G_reader_settings:flipNilOrFalse("dev_no_c_blitter"):nilOrFalse("dev_no_c_blitter"))
```
poire-z, line 345: For readability, can you make this into 2 lines?
```lua
G_reader_settings:flipNilOrFalse("dev_no_c_blitter")
bb:enableCBB(G_reader_settings:nilOrFalse("dev_no_c_blitter"))
```

reader.lua
```diff
@@ -37,23 +37,10 @@ if lang_locale then
...
-else
-    if ffi.os == "Windows" then
-        C._putenv("KO_NO_CBB=false")
-    else
-        C.unsetenv("KO_NO_CBB")
-    end
-end
+
+-- Make the C blitter optional and synchronize the setting with current state
+G_reader_settings:saveSetting("dev_no_c_blitter", not require("ffi/blitbuffer"):enableCBB(G_reader_settings:nilOrFalse("dev_no_c_blitter")))
```
poire-z, line 43: Same and even more needed for this one (I'm not even guessing how many lines it should be ;).

frontend/apps/filemanager/filemanagermenu.lua
ezdiy: On one hand, it's nice of luasettings to speak english predicates. On another, it returning self is utterly useless because it just encourages horrible things like this. Ideally one should be doing just 
```
bb:enableCBB(G_reader_settings:flipNilOrFalse("dev_no_c_blitter"))
```
Let's everyone now make pinkie pie promise to never use self chaining in settings predicates ever again, so this can be made lil more sensible in the future by making settings return new value whenever it flips something.

NiLuJe (approved):
Thanks ;).

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -332,20 +332,18 @@ function FileManagerMenu:setUpdateItemTable()
         })
     end
     if not Device.should_restrict_JIT then
+        local bb = require("ffi/blitbuffer")
```
poire-z, line 335: Naming the imported module `bb` is a bit confusing (although not that much here as it's used immediately) because `bb` is often used just for a buffer instance (`self.bb / local bb = Blitbuffer.new(...)`).
Best to stick with one the 2 variations we use everywhere:
```lua
local BlitBuffer = require("ffi/blitbuffer")
local Blitbuffer = require("ffi/blitbuffer")
```

ezdiy: It's how you can tell apart vim person from emacs person without asking em what editor they use.

pazos (approved):

PR #6707, opened 2020-09-24
Sort folder shortcuts

smartscripts-nl:
In the current KOReader version the folder shortcuts are sorted by date of addition. When a user has many folder shortcuts (like I do 😊 ) they are easier to find again when sorted alphabetically.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6707)

### DIFF
--- a/frontend/apps/filemanager/filemanagershortcuts.lua
+++ b/frontend/apps/filemanager/filemanagershortcuts.lua
@@ -63,7 +63,7 @@ function FileManagerShortcuts:updateItemTable()
     end
 
     self.fm_bookmark:switchItemTable(nil,
-        item_table, select_number)
+                                     item_table, select_number)
 end
 
 function FileManagerShortcuts:addNewFolder()
--- a/frontend/apps/filemanager/filemanagershortcuts.lua
+++ b/frontend/apps/filemanager/filemanagershortcuts.lua
@@ -48,16 +48,12 @@ function FileManagerShortcuts:updateItemTable()
     table.sort(item_table, function(l, r)
         return l.text < r.text
     end)
-    local item_table_sorted = {}
-    table.insert(item_table_sorted, {
+    table.insert(item_table, 1, {
         text = _("Add new folder shortcut"),
         callback = function()
             self:addNewFolder()
         end,
     })
-    for _, item in ipairs(item_table) do
-        table.insert(item_table_sorted, item)
-    end
 
     -- try to stay on current page
     local select_number
@@ -67,7 +63,7 @@ function FileManagerShortcuts:updateItemTable()
     end
 
     self.fm_bookmark:switchItemTable(nil,
-        item_table_sorted, select_number)
+        item_table, select_number)
 end
 
 function FileManagerShortcuts:addNewFolder()
--- a/frontend/apps/filemanager/filemanagershortcuts.lua
+++ b/frontend/apps/filemanager/filemanagershortcuts.lua
@@ -6,21 +6,22 @@ local InputDialog = require("ui/widget/inputdialog")
 local Menu = require("ui/widget/menu")
 local Screen = require("device").screen
 local UIManager = require("ui/uimanager")
+local lfs = require("libs/libkoreader-lfs")
 local util = require("ffi/util")
 local _ = require("gettext")
 local T = util.template
 
 local FileManagerShortcuts = InputContainer:extend{}
 
+local function tableSortByTextProp(subject)
+    table.sort(subject, function (v1, v2)
+        return v1.text < v2.text
+    end)
+end
+
 function FileManagerShortcuts:updateItemTable()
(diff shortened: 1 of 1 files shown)


frontend/apps/filemanager/filemanagershortcuts.lua
```diff
@@ -6,21 +6,22 @@ local InputDialog = require("ui/widget/inputdialog")
 local Menu = require("ui/widget/menu")
 local Screen = require("device").screen
 local UIManager = require("ui/uimanager")
+local lfs = require("libs/libkoreader-lfs")
```
NiLuJe, line 9: Unneeded?

smartscripts-nl: The script *does* work without this inclusion. But then lfs is loaded from a global place. While in most classes lfs is loaded as a local library. So it seems more consistent to do that here also.

frontend/apps/filemanager/filemanagershortcuts.lua
```diff
@@ -6,21 +6,22 @@ local InputDialog = require("ui/widget/inputdialog")
...
 local util = require("ffi/util")
 local _ = require("gettext")
 local T = util.template
 
 local FileManagerShortcuts = InputContainer:extend{}
 
+local function tableSortByTextProp(subject)
+    table.sort(subject, function (v1, v2)
+        return v1.text < v2.text
```
Frenzie, line 18: Are you sure this isn't equivalent to table.sort without supplying a custom sorting function?

smartscripts-nl: The context is that in my private installation I have several helper methods for custom sorting operations which I can quickly use, without typing much code. But now I've pushed a sorting operation that doesn't require a local function.

Frenzie: That doesn't address my question though: isn't this quite literally the default behavior?

Frenzie: Okay, the answer is the table is always deeper. So carry on, but there was no hidden meaning. ;-)

frontend/apps/filemanager/filemanagershortcuts.lua
```diff
@@ -50,15 +45,29 @@ function FileManagerShortcuts:updateItemTable()
...
+    local item_table_sorted = {}
+    table.insert(item_table_sorted, {
+        text = _("Add new folder shortcut"),
+        callback = function()
+            self:addNewFolder()
+        end,
+    })
+    for _, item in ipairs(item_table) do
+        table.insert(item_table_sorted, item)
+    end
```
Frenzie, line 60: I have no idea what's going on here tbh. ;-) Why not just:

```diff
diff --git a/frontend/apps/filemanager/filemanagershortcuts.lua b/frontend/apps/filemanager/filemanagershortcuts.lua
index 3f54acce..8fa5d113 100644
--- a/frontend/apps/filemanager/filemanagershortcuts.lua
+++ b/frontend/apps/filemanager/filemanagershortcuts.lua
@@ -48,16 +48,12 @@ function FileManagerShortcuts:updateItemTable()
     table.sort(item_table, function(l, r)
         return l.text < r.text
     end)
-    local item_table_sorted = {}
-    table.insert(item_table_sorted, {
+    table.insert(item_table, 1, {
         text = _("Add new folder shortcut"),
         callback = function()
             self:addNewFolder()
         end,
     })
-    for _, item in ipairs(item_table) do
-        table.insert(item_table_sorted, item)
-    end
 
     -- try to stay on current page
     local select_number
@@ -67,7 +63,7 @@ function FileManagerShortcuts:updateItemTable()
     end
 
     self.fm_bookmark:switchItemTable(nil,
-        item_table_sorted, select_number)
+        item_table, select_number)
 end
 
 function FileManagerShortcuts:addNewFolder()
```

smartscripts-nl: I simply wasn't to aware of the possibility to insert at the start of a table. I've pushed a commit implementing your suggestion.

frontend/apps/filemanager/filemanagershortcuts.lua
```diff
@@ -50,15 +45,25 @@ function FileManagerShortcuts:updateItemTable()
...
-    local select_number = nil
+    local select_number
 
     if self.fm_bookmark.page and self.fm_bookmark.perpage then
         select_number = (self.fm_bookmark.page - 1) * self.fm_bookmark.perpage + 1
     end
 
     self.fm_bookmark:switchItemTable(nil,
-                                     item_table, select_number)
+        item_table, select_number)
```
Frenzie, line 66: Only ended up in my diff because it was with the version with the second item_table.
```suggestion
                                     item_table, select_number)
```

PR #6709, opened 2020-09-24
Panel zoom - improve output image resolution

Galunid:
Implementing what @Frenzie mentioned in https://github.com/koreader/koreader/pull/6511#issuecomment-698584844
It should improve resolution of output "panel" (https://github.com/koreader/koreader/pull/6511#issuecomment-698346603)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6709)

### DIFF
--- a/frontend/document/document.lua
+++ b/frontend/document/document.lua
@@ -415,7 +415,7 @@ end
 
 function Document:getPagePart(pageno, rect, rotation)
     local canvas_size = CanvasContext:getSize()
-    local zoom = math.min(canvas_size.w / rect.w, canvas_size.h / rect.h)
+    local zoom = math.min(canvas_size.w*2 / rect.w, canvas_size.h*2 / rect.h)
     -- it's really, really important to do math.floor, otherwise we get image projection
     local scaled_rect = {
         x = math.floor(rect.x * zoom),
--- a/frontend/document/document.lua
+++ b/frontend/document/document.lua
@@ -414,9 +414,18 @@ function Document:getDrawnImagesStatistics()
 end
 
 function Document:getPagePart(pageno, rect, rotation)
-    local tile = self:renderPage(pageno, rect, 1, rotation, 1, 0)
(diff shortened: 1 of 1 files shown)


poire-z:
Works better indeed.
One small issue, when clicking on the first full-width panel:
image

it's extracted fine and shown, but its original size is nearly the same as on the page - so, when zooming it in ImageViewer, you don't get the same quality as on half-width panels.

Galunid:
It actually works as intended, but I didn't have a better idea, how to handle this case. Do you think it would be better to rotate the image, so it actually is zoomed?

frontend/document/document.lua
```diff
@@ -414,9 +414,18 @@ function Document:getDrawnImagesStatistics()
 end
 
 function Document:getPagePart(pageno, rect, rotation)
-    local tile = self:renderPage(pageno, rect, 1, rotation, 1, 0)
-    local target = Blitbuffer.new(rect.w, rect.h, self.render_color and self.color_bb_type or nil)
-    target:blitFrom(tile.bb, 0, 0, rect.x, rect.y, rect.w, rect.h)
+    local canvas_size = CanvasContext:getSize()
+    local zoom = math.min(canvas_size.w / rect.w, canvas_size.h / rect.h)
```
Frenzie, line 418: The zoom factor can also make the resulting picture smaller. I imagine we don't want to enforce a minimum zoom factor of 1 (that could be huge! sometimes), so a stupidly simple workaround to both that and @poire-z's comment https://github.com/koreader/koreader/pull/6709#issuecomment-698624533 could also be something like:
```suggestion
    local zoom = math.min(canvas_size.w*2 / rect.w, canvas_size.h*2 / rect.h)
```

Frenzie: Or of course:
```suggestion
    local zoom = math.max(canvas_size.w / rect.w, canvas_size.h / rect.h)
```

Frenzie (approved):
@poire-z Good for you?

poire-z:
Yes, working fine with my full-width panel above (can't find any full-height panel to test :).

PR #6712, opened 2020-09-25
linuxfb: frontends need to pass is_always_portrait

### LINKED ISSUE #6711: Koreader unusable after updating to the latest unstable version (version 2020-09-20, Cervantes 4)
I just updated to the latest unstable version my Cervantes 4
After the update, the screen was automatically switched to landscape mode, but the reader behaves as if it was still in portrait mode: I need to guess where the element I want to touch should be if the display was in portrait mode.
I tried to switch to portrait in the interface, but the operation is not possible: it's like the first button of the bottom bar was ouside of the screen. And it is the same for the 3 first elements of the top bar.

ezdiy:
This is because fb driver should not be inspecting self.device for low level
caps anymore.

Fixes #6711

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6712)

### DIFF
--- a/frontend/device/cervantes/device.lua
+++ b/frontend/device/cervantes/device.lua
@@ -157,7 +157,7 @@ function Cervantes:initEventAdjustHooks()
 end
 
 function Cervantes:init()
-    self.screen = require("ffi/framebuffer_mxcfb"):new{device = self, debug = logger.dbg}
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Yup, quite true. :+1:

PR #6713, opened 2020-09-25
Save document settings before activating USBMS

smartscripts-nl:
As things were, when I activated USBMS, any bookmarks recently added and not yet stored in the sidecar file would be lost after returning to the reader. These changes fix that.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6713)

### DIFF
--- a/frontend/ui/elements/mass_storage.lua
+++ b/frontend/ui/elements/mass_storage.lua
@@ -46,6 +46,14 @@ function MassStorage:getActionsMenuTable()
     }
 end
 
+local function saveDocSettings()
+    local
+    readerui_instance = require("apps/reader/readerui"):_getRunningInstance()
+    if readerui_instance then
+        readerui_instance:saveSettings()
+    end
+end
+
 -- exit KOReader and start mass storage mode.
 function MassStorage:start(never_ask)
     if not Device:canToggleMassStorage() or not self:isEnabled() then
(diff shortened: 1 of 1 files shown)


frontend/ui/elements/mass_storage.lua
```diff
@@ -54,15 +62,19 @@ function MassStorage:start(never_ask)
 
     if not never_ask and self:requireConfirmation() then
         local ConfirmBox = require("ui/widget/confirmbox")
-        UIManager:show(ConfirmBox:new{
+        UIManager:show(ConfirmBox:new {
             text = _("Share storage via USB?"),
             ok_text = _("Share"),
             ok_callback = function()
+                -- save document settings before activating USBMS:
+                saveDocSettings()
```
poire-z, line 70: I think you could just do here (and below) (and remove your function above):
`UIManager:flushSettings()`
which will also have global settings saved when in file manager.

smartscripts-nl: You're right! Thanks. I'll modify my PR accordingly.

poire-z (approved):

frontend/ui/elements/mass_storage.lua
```diff
@@ -54,15 +54,19 @@ function MassStorage:start(never_ask)
 
     if not never_ask and self:requireConfirmation() then
         local ConfirmBox = require("ui/widget/confirmbox")
-        UIManager:show(ConfirmBox:new{
+        UIManager:show(ConfirmBox:new {
```
Frenzie, line 57: No need to change ;-)
```suggestion
        UIManager:show(ConfirmBox:new{
```

smartscripts-nl: I now see it's my IDE that does this! I wasn't aware of this change. I'll fix it.

smartscripts-nl: No need anymore to change it, because you already merged my commit.

Frenzie: There would still be a need of sorts to change it if I'd accidentally put it in that way but for something this minor it hardly matters who commits it or how. ^_^ (I hope you agree, anyway!)

Issue #976, opened 2014-10-08
808: Double click to next chapter crashes


WS64:
./luajit: frontend/apps/reader/modules/readertoc.lua:113: attempt to compare number with nil
stack traceback:
    frontend/apps/reader/modules/readertoc.lua:113: in function 'getTocTicks'
    frontend/apps/reader/modules/readertoc.lua:133: in function 'getNextChapter'
    frontend/apps/reader/modules/readerrolling.lua:250: in function 'handleEvent'
    frontend/ui/widget/container/widgetcontainer.lua:68: in function 'propagateEvent'
    frontend/ui/widget/container/widgetcontainer.lua:80: in function 'handleEvent'
    frontend/ui/uimanager.lua:246: in function 'sendEvent'
    frontend/ui/uimanager.lua:91: in function '**default**'
    frontend/ui/uimanager.lua:434: in function 'run'
    ./reader.lua:176: in main chunk
    [C]: at 0x0000bde5

This still worked in 804

Issue #6339, opened 2020-07-01
Slightly malformed TOC not showing


Frenzie:
* KOReader version: 09c654c
* Device: all

#### Issue
I noticed the TOC in my newspaper doesn't show anything, while it does in calibre.

Here's a scrambled copy: [NRC Handelsblad 2020-07-01 - NRC Handelsblad_scrambled.epub.zip](https://github.com/koreader/koreader/files/4858053/NRC.Handelsblad.2020-07-01.-.NRC.Handelsblad_scrambled.epub.zip) (remove .zip from the name)

It looks like it's not quite proper, but I suppose that's how things can be "in the wild."
```
$ epubcheck nrc_20200701.epub 
Validating using EPUB version 2.0.1 rules.
ERROR(RSC-005): nrc_20200701.epub/toc.ncx(10,12): Error while parsing file: element "doctitle" not allowed anywhere; expected element "docTitle"
ERROR(RSC-005): nrc_20200701.epub/toc.ncx(13,10): Error while parsing file: element "doctitle" not allowed anywhere; expected element "docTitle"
ERROR(RSC-005): nrc_20200701.epub/toc.ncx(15,10): Error while parsing file: element "navmap" not allowed anywhere; expected element "docTitle"
ERROR(RSC-005): nrc_20200701.epub/toc.ncx(20,12): Error while parsing file: element "navmap" not allowed anywhere; expected element "docTitle"
ERROR(RSC-005): nrc_20200701.epub/toc.ncx(23,14): Error while parsing file: element "navmap" not allowed anywhere; expected element "docTitle"
ERROR(RSC-005): nrc_20200701.epub/toc.ncx(28,13): Error while parsing file: element "navmap" not allowed anywhere; expected element "docTitle"
ERROR(RSC-005): nrc_20200701.epub/toc.ncx(31,15): Error while parsing file: element "navmap" not allowed anywhere; expected element "docTitle"
```

[NRC Handelsblad 2020-07-01 - NRC Handelsblad_scrambled.epub.zip](https://github.com/koreader/koreader/files/4858045/NRC.Handelsblad.2020-07-01.-.NRC.Handelsblad_scrambled.epub.zip)

poire-z:
Its toc.ncx file has lowercased tag names:
```xml
<?xml version='1.0' encoding='utf-8'?>
<ncx xmlns="http://www.daisy.org/z3986/2005/ncx/" version="2005-1">
 <head>
[...]

PR #6716, opened 2020-09-26
Revamp ToC ticks handling

### LINKED ISSUE #5500: When chapters have subchapters

NiLuJe:
Instead of the fancyish heuristics implemented in #976 & #984, simply flatten the full list of ToC nodes, dropping duplicates in the process.

It turns out that nothing really makes use of the depth level, so, in practice, mostly everything was relying on the level 0 heuristic.

See #5500 for more details on the subject.

Fix #5500

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6716)


NiLuJe:
Because I'm an idiot, I don't have "before" pictures, but here's the "after" on the testcase from https://www.mobileread.com/forums/showpost.php?p=4038495&postcount=4 (thanks!)

testcase

NiLuJe:
I haven't actually run the testcase (which, fun fact: is the only thing actually using the depth levels directly ;p), so, heh, it might have blown up a bit ;p.

NiLuJe:
I, err, haven't slept much this week, so, glad for reviews and tests, as I have no actual idea how I'm still conscious ^^.

Frenzie (commented):
Even when my documents come with a theoretically good TOC that stuff is [still broken](
https://github.com/koreader/koreader/issues/6339) so I'm not really the person to test/comment.

NiLuJe:
Random thought: I vaguely remember someone bench'ing `ipairs` vs? the `for i = 1, #t` approach: we don't care, right? :D.

(I vaguely tend to find ipairs more readable myself, FWIW).

NiLuJe:
Oh, I quickly checked a random Wikipedia entry as @poire-z mentioned, and it appears to behave.

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -275,147 +280,174 @@ function ReaderToc:getTocTitleOfCurrentPage()
...
+    local ticks = self:getTocTicks()
     local ticks_candidates = {}
-    local max_level = self:getMaxDepth()
-    for i = 0, -max_level, -1 do
-        local ticks = self:getTocTicks(i)
-        table.insert(ticks_candidates, ticks)
+
+    for _, v in ipairs(ticks) do
+        for depth, page in ipairs(v) do
+            table.insert(ticks_candidates, page)
```
poire-z, line 358: You could do the duplicate/hash check here, and not insert it if page already seen. (And then, no need for the intermediate ticks_candidate.)

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -275,147 +280,174 @@ function ReaderToc:getTocTitleOfCurrentPage()
...
-        -- Find the finest toc ticks by sorting out the largest one
-        table.sort(ticks_candidates, function(a, b) return #a > #b end)
-        return ticks_candidates[1]
+
+    -- Then drop duplicates (by using the value as a key to make a quick check) (c.f., https://stackoverflow.com/a/20067270)
+    local hash = {}
+    local ticks_flattened = {}
+    for _, v in ipairs(ticks_candidates) do
+        if (not hash[v]) then
+            ticks_flattened[#ticks_flattened+1] = v
```
poire-z, line 367: Or use `table.insert()` as above :)

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -275,147 +280,174 @@ function ReaderToc:getTocTitleOfCurrentPage()
...
 
-function ReaderToc:getNextChapter(cur_pageno, level)
-    local ticks = self:getTocTicks(level)
+function ReaderToc:getNextChapter(cur_pageno)
+    local ticks = self:getTocTicksFlattened()
     local next_chapter = nil
-    for i = 1, #ticks do
-        if ticks[i] > cur_pageno then
-            next_chapter = ticks[i]
+    for _, v in ipairs(ticks) do
```
poire-z, line 382: (As value is a page, I would have named it p: `for _, p in ipairs(ticks) do if p > cur_pageno then` is a tad more readable/hinting.)

frontend/apps/reader/skimtowidget.lua
```diff
@@ -359,9 +347,9 @@ end
 
 function SkimToWidget:getNextChapter(cur_pageno)
     local next_chapter = nil
-    for i = 1, #self.ticks_candidates do
-        if self.ticks_candidates[i] > cur_pageno then
-            next_chapter = self.ticks_candidates[i]
+    for _, v in ipairs(self.ticks_flattened) do
+        if v > cur_pageno then
+            next_chapter = v
```
poire-z, line 352: Can we re-use `ReaderToc:getNextChapter(cur_pageno)` ? `return self.ui.toc:getNextChapter(cur_pageno)` or even use that at the calling places.

poire-z:
(Not tested yet, will do.)

So, there are really nothing using level other than 0?

Just a thought for the future, and so you can have a quick see (or even a go at it :) if that would fit in your refactoring:
People will now have many ticks on some books. We might want to have something UI to limit ticks to some level (per book, always start with all of them): so, when we have many sublevels (might not be really chapters, could be alinas, lists..., we might want to limit ticks to show only level 1 and 2 (also used by getNextChapter()) so one can browse by top-level chapters if he considers subchapters to be crap/noisy.
A little ProgressBar like for frontlight with `toc_depth` nb of squares could do :)

frontend/apps/reader/skimtowidget.lua
NiLuJe: Oh, duh, didn't catch that among the rest of the code duplication in there, thanks ;).

NiLuJe:
> (Not tested yet, will do.)
> 
> So, there are really nothing using level other than 0?
> 
> Just a thought for the future, and so you can have a quick see (or even a go at it :) if that would fit in your refactoring:
> People will now have many ticks on some books. We might want to have something UI to limit ticks to some level (per book, always start with all of them): so, when we have many sublevels (might not be really chapters, could be alinas, lists..., we might want to limit ticks to show only level 1 and 2 (also used by getNextChapter()) so one can browse by top-level chapters if he considers subchapters to be crap/noisy.
> A little ProgressBar like for frontlight with `toc_depth` nb of squares could do :)

I mean, that's certainly technically doable, I kept the full hierarchical tree intact for just such potential use-cases.

Right now, this appears to work for me, so, unless I actively find something bothering me with it, we'll see, but I probably won't dig into it further (any time soon, that is) ;).

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -275,147 +280,169 @@ function ReaderToc:getTocTitleOfCurrentPage()
...
+
+    -- It hasn't been cached yet, compute it.
+    local ticks = self:getTocTicks()
+    local ticks_flattened = {}
+
+    -- Keep track of what we add to avoid duplicates (c.f., https://stackoverflow.com/a/20067270)
+    local hash = {}
+    for _, v in ipairs(ticks) do
+        for depth, page in ipairs(v) do
+            if (not hash[page]) then
```
poire-z, line 360: (Smells C :)

NiLuJe: It was actually as-is in the SO snippet, and I kept it, because, indeed, I like parenthesis ;p

poire-z:
Btw, I have in my own patch:
```lua
function ReaderToc:onPageUpdate(pageno)
    if UIManager.FULL_REFRESH_COUNT == -1 or G_reader_settings:isTrue("refresh_on_chapter_boundaries") then
        if self.avoid_chapter_flash == nil then -- avoid that on Wikipedia EPUBs
            -- No flash on chapter if too many ticks
(L71)       self.avoid_chapter_flash = (#self:getTocTicks(0) / self.ui.document:getPageCount()) > 0.3
        end
        if self.avoid_chapter_flash then
            self.pageno = pageno
            return
        end
```
which crashed with:
`frontend/apps/reader/modules/readertoc.lua:71: attempt to get length of a nil value`
So, I guess `self:getTocTicks(0)` does not work as before.
You tried to keep that working even if not used, but may be you missed the specific level=0 ?

Using getTocTicksFlattened() instead in my patch, and it works.

Checked on 2 documents, and I get the same ticks as with my:
```
-                depth = v.depth
+                -- depth = v.depth -- so we get them all and don't ignore toc entries that have children
```

So, looks good to me :) :+1:

NiLuJe:
Oh, yeah, asking for level 0 is definitely not handled.

Basically, whatever called getTocTicks(0) should now call getTocTicksFlattened ;).

PR #6721, opened 2020-09-27
Pocketbook: Try to auto-detect HW feature caps more precisely

ezdiy:
The differences between Allwinner/MXC matter surprisingly little in practice (https://github.com/koreader/koreader-base/issues/1202), it's not even worth it to bother the base mxc driver with it. Instead, frontend can make some informed choice by probing for the hardware and tweaking requested device caps accordingly.

On Allwinner B288, we disable both hwrot (done in software in kernel and terribly, ie not worth it) and hwinverse (no support whatsoever). On NXT we'll keep both on, as the kernel driver seems to be [using the hardware properly](https://github.com/pocketbook/Platform_MX6/blob/master/Kernel_631/drivers/video/mxc/mxc_epdc_fb.c#L2218)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6721)

### DIFF
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -84,11 +84,11 @@ local Device = {
     isAlwaysPortrait = no,
     -- On some devices (eg newer pocketbook) we can force HW rotation on the fly (before each update)
     -- The value here is table of 4 elements mapping the sensible linux constants to whatever
-    -- nonsense the device actually has. Canonically it should be { 0, 1, 2, 3 } if the device
+    -- nonsense the device actually has. Canonically it should return { 0, 1, 2, 3 } if the device
     -- matches <linux/fb.h> FB_ROTATE_* constants.
     -- See https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer.lua for full template
     -- of the table expected.
-    usingForcedRotation = nil,
+    usingForcedRotation = function() return nil end,
     -- needs full screen refresh when resumed from screensaver?
     needsScreenRefreshAfterResume = yes,
 
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -37,8 +37,10 @@ local PocketBook = Generic:new{
     -- all devices that have warmth lights use inkview api
     hasNaturalLightApi = yes,
 
-    -- Apparently, HW inversion is a pipedream on PB (#6669)
-    canHWInvert = no,
+    -- NOTE: Apparently, HW inversion is a pipedream on PB (#6669), ... well, on sunxi chipsets anyway.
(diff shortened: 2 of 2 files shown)


frontend/device/pocketbook/device.lua
```diff
@@ -37,8 +37,10 @@ local PocketBook = Generic:new{
     -- all devices that have warmth lights use inkview api
     hasNaturalLightApi = yes,
 
-    -- Apparently, HW inversion is a pipedream on PB (#6669)
-    canHWInvert = no,
+    -- NOTE: Apparently, HW inversion is a pipedream on PB (#6669), ... well, on sunxi chipsets anyway.
+    -- For which we now probe in fbinfoOverride() and tweak the flag to "no".
+    -- NXT/other chipsets *should* work (PB631), but in case it doesn't on your device, set this to "no" in here.
```
NiLuJe, line 42: Nitpick: It's NTX, not NXT ;).

(And the SoCs are from NXP (ex-Freescale), which doesn't help avoid confusion ^^).

NiLuJe (approved):
Besides the NTX nit, LGTM ;).

Random query, though: what does `islwaysPortrait` does now on devices with `usingForcedRotation`?

ezdiy:
> Random query, though: what does `isAlwaysPortrait` does now on devices with `usingForcedRotation`?

The flags are mutually exclusive. So even though the device desc sets *both*, the tweak decides on one, another or none. Will add some comment about too I guess.

frontend/device/generic/device.lua
```diff
@@ -84,11 +84,11 @@ local Device = {
     isAlwaysPortrait = no,
     -- On some devices (eg newer pocketbook) we can force HW rotation on the fly (before each update)
     -- The value here is table of 4 elements mapping the sensible linux constants to whatever
-    -- nonsense the device actually has. Canonically it should be { 0, 1, 2, 3 } if the device
+    -- nonsense the device actually has. Canonically it should return { 0, 1, 2, 3 } if the device
     -- matches <linux/fb.h> FB_ROTATE_* constants.
     -- See https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer.lua for full template
     -- of the table expected.
-    usingForcedRotation = nil,
+    usingForcedRotation = function() return nil end,
```
Frenzie, line 91:
```suggestion
    usingForcedRotation = function() return end,
```

Frenzie (approved):
Sure, lgtm

NiLuJe (approved):

frontend/device/generic/device.lua
ezdiy: (intentional, to signal there's meant to be no table)

PR #6725, opened 2020-09-28
Fancier Wi-Fi status icons in ReaderFooter

NiLuJe:
And also allow enabling said icon on Kobo & Cervantes.

(Supersedes #6656, see discussions over there ;)).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6725)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -75,6 +75,7 @@ local symbol_prefix = {
         frontlight = "☼",
         mem_usage = "",
         wifi_status = "",
+        wifi_status_off = "",
     }
 }
 if BD.mirroredUILayout() then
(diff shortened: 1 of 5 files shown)


Frenzie (approved):
Sure thing.

pazos (approved):
lgtm

poire-z:
(I'll not be seeing these icons, as I hacked myself a big black solic square in the bottom right corner when my wifi is enabled :)

But I find the 2 icons quite a bit too similar (left and right most, hacked there for a quick look):

image

I'd rather see something a bit more flat (top half of the glyph blank) so it's visually easier to notice the difference. No other option?
(and I'm not telling again how I dislike this solid black battery icon - oups :)

NiLuJe:
Nope, the only other decent source is MDI, and they're [actually solid black](https://material.io/resources/icons/?search=wifi&style=baseline) ;p.

(And, also, not in NerdFont anyway ^^)

NiLuJe:
The only other things that vaguely fit would be the [cellular signal](https://material.io/resources/icons/?search=signal&style=baseline) ones (as well as the airplane mode ones).

But, err, not a fan of either of those ^^.

poire-z (approved):
Well, ok then :)

Issue #6726, opened 2020-09-28
Android detects VPN (wireguard) connection as WI-FI enabled


Galunid:
* KOReader version: v2020.09
* Device: Xiaomi MI A2 Lite/Onyx Boox Nova 2 (should be device independent)

#### Issue
When wifi is disabled, and device has VPN (in my case wireguard) enabled KOReader detects wifi as enabled. It's problematic because KOSync plugin/Calibre plugin have to timeout if triggered in this state.
#### Steps to reproduce
Connect to wireguard -> Disconnect wifi -> KOReader still detects wifi enabled (because VPN connection is still active)

pazos:
See https://github.com/koreader/android-luajit-launcher/blob/e3d58c6432ba06d43bdd609c38e6315367825d7e/app/src/org/koreader/launcher/BaseActivity.kt#L338-L402

It is expected. VPN will prevail over wifi (ie: if both are active). Maybe if connection type is VPN do aditional online check?.

I wouldn't want to remove VPN as a valid network type. I use gnirehtet a lot (which is recongized as VPN network).

Galunid:
Perhaps it's enough to just not `return true`  or return `connected` in those snippets
https://github.com/koreader/android-luajit-launcher/blob/e3d58c6432ba06d43bdd609c38e6315367825d7e/app/src/org/koreader/launcher/BaseActivity.kt#L364-L367
https://github.com/koreader/android-luajit-launcher/blob/e3d58c6432ba06d43bdd609c38e6315367825d7e/app/src/org/koreader/launcher/BaseActivity.kt#L392-L395
I'm not actually too clear on how the `when` works when multiple conditions are true, but my idea is to not change "connected" state with VPN connection, because user being connected to VPN doesn't imply internet connection. Hope it makes sense

pazos:
@Galunid: the real problem is android != all the other platforms.

On embedded linux devices network == wifi
On desktop we don't care about network

On android network means a lot of things and VPN is a valid kind of network.
It's usually used over WiFi, over mobile and over Ethernet.

You might want to try

```diff
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -259,9 +259,10 @@ function Device:initNetworkManager(NetworkMgr)
     end
 
     function NetworkMgr:isWifiOn()
-        local ok = android.getNetworkInfo()
+        local ok, transport = android.getNetworkInfo()
         ok = tonumber(ok)
         if not ok then return false end
[...]

PR #6733, opened 2020-09-30
KOSync improvements

Galunid:
If user manually triggered update, allow for automatically enabling WiFi, add new menu entry "Sync every # pages", so users don't have to rely on modifying `defaults.lua.persistent` and set global variable `DAUTO_SAVE_PAGING_COUNT`. It won't break like #6489, because we check if user triggered the update. Some discussion in https://github.com/koreader/koreader/issues/6726#issuecomment-700679676. It should be alright, but I'll go over it once again after I get some sleep ;P

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6733)

### DIFF
--- a/plugins/kosync.koplugin/main.lua
+++ b/plugins/kosync.koplugin/main.lua
@@ -23,9 +23,6 @@ end
 -- that was also used here as the interval for auto sync.
 -- DAUTO_SAVE_PAGING_COUNT has been removed, but let's allow
 -- this plugin to still pick it from defaults.persistent.lua.
---- @todo make this tunable via an added menu item below
-local SYNC_PAGING_COUNT = DAUTO_SAVE_PAGING_COUNT -- luacheck: ignore
-
 local KOSync = InputContainer:new{
     name = "kosync",
     is_doc_only = true,
@@ -111,6 +108,7 @@ function KOSync:onReaderReady()
     self.kosync_username = settings.username
     self.kosync_userkey = settings.userkey
     self.kosync_auto_sync = not (settings.auto_sync == false)
+    self.kosync_pages_before_update = settings.pages_before_update or DAUTO_SAVE_PAGING_COUNT
     self.kosync_whisper_forward = settings.whisper_forward or SYNC_STRATEGY.DEFAULT_FORWARD
     self.kosync_whisper_backward = settings.whisper_backward or SYNC_STRATEGY.DEFAULT_BACKWARD
     self.kosync_device_id = G_reader_settings:readSetting("device_id")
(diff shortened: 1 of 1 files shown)


plugins/kosync.koplugin/main.lua
```diff
@@ -23,9 +23,6 @@ end
 -- that was also used here as the interval for auto sync.
 -- DAUTO_SAVE_PAGING_COUNT has been removed, but let's allow
 -- this plugin to still pick it from defaults.persistent.lua.
---- @todo make this tunable via an added menu item below
-local SYNC_PAGING_COUNT = DAUTO_SAVE_PAGING_COUNT -- luacheck: ignore
```
Frenzie, line 27: This comes along with editing defaults.lua. But there we comment it to signal it's no longer used, see some other values.

Galunid: Not exactly sure what you mean. Do you want me to get rid of that variable completely? I left it for the sake of compatibility (the comments above the highlighted lines). I'm fine removing `DAUTO_SAVE_PAGING_COUNT` completely, since in most cases it's `nil`

Frenzie: I mean this looks like a PR similar to https://github.com/koreader/koreader/pull/5754. :-)

Galunid: I removed `DAUTO_SAVE_PAGING_COUNT`, should be good to go

plugins/kosync.koplugin/main.lua
```diff
@@ -275,10 +268,39 @@ function KOSync:addToMainMenu(menu_items)
                     }
                 end,
             },
+            {
+                text = _("Sync every # pages"),
+                keep_menu_open = true,
+                callback = function()
+                    local SpinWidget = require('ui/widget/spinwidget')
```
Frenzie, line 275:
```suggestion
                    local SpinWidget = require("ui/widget/spinwidget")
```

plugins/kosync.koplugin/main.lua
```diff
@@ -275,10 +268,39 @@ function KOSync:addToMainMenu(menu_items)
...
+                    local items = SpinWidget:new{
+                        text = _([[This value decides how many pages it takes to update book progress.
+If set to 0, disables updating progress based on page turns.]]),
```
Frenzie, line 278:
```suggestion
If set to 0, updating progress based on page turns will be disabled.]]),
```

plugins/kosync.koplugin/main.lua
```diff
@@ -275,10 +268,39 @@ function KOSync:addToMainMenu(menu_items)
                     }
                 end,
             },
+            {
+                text = _("Sync every # pages"),
+                keep_menu_open = true,
+                callback = function()
+                    local SpinWidget = require('ui/widget/spinwidget')
+                    local items = SpinWidget:new{
+                        text = _([[This value decides how many pages it takes to update book progress.
```
Frenzie, line 277:
```suggestion
                        text = _([[This value determines how many page turns it takes to update book progress.
```

Frenzie (approved):

PR #6735, opened 2020-09-30
Pocketbook: Fix silly rotation bug on PB1040

ezdiy:
Ditto as PB740

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6735)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -494,6 +494,7 @@ local PocketBook840 = PocketBook:new{
 local PocketBook1040 = PocketBook:new{
     model = "PB1040",
     display_dpi = 227,
+    isAlwaysPortrait = yes,
     usingForcedRotation = landscape_ccw,
     hasNaturalLight = yes,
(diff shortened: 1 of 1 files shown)


ezdiy:
Blah.

Frenzie:
A mismatch of expectations I suppose. Please request a detailed review when desired in that case. ^_^

Generally I focus more on the general concepts and maintainability, while for the details I tend to stick to a more superficial survey of traps I've experienced or know of. Ideally of course there'd be two or three editing passes, one for structure and another one for details. Except for specifically something like this PB-refactor where no one double checked the definitions were quite right, I do think that does tend to happen organically, simply with two or three different sets of eyeballs focusing on slightly different aspects of details. (E.g., I tend to take a more detailed look at the language/internationalization and UX aspects.)

PS I'm moving this month, so just a heads-up that my reviews probably won't be particularly thorough beyond checking if things look sensible and if they work.

ezdiy:
I jest :>

PR #6737, opened 2020-10-01
Terminal: add placeholders

smartscripts-nl:
Typing on an e-reader (e.g. a Kobo device) is not the most pleasant experience. Placeholders for terminal commands take away some of that pain.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6737)

### DIFF
--- a/plugins/terminal.koplugin/main.lua
+++ b/plugins/terminal.koplugin/main.lua
@@ -102,7 +102,7 @@ function Terminal:showHelp()
         local placeholder = v[2]
         message = message .. "\n%" .. placeholder .. " = " .. label
     end
-    message = message .. _('\n%v = add value from prompt\n\nExample:\ncat "%s"\nto display sidecar file for current ebook')
+    message = message .. _('\n%v = add value from prompt\n\nExample:\ncat "%s/%v"\nto get a prompt for displaying a file in the settings dir')
     UIManager:show(InfoMessage:new{
         text = message
     })
--- a/plugins/terminal.koplugin/main.lua
+++ b/plugins/terminal.koplugin/main.lua
@@ -29,6 +29,13 @@ local Terminal = WidgetContainer:new{
     -- placeholders in commands are marked with "%[placeholder]":
     placeholders = {
         {
+            _("sidecar file current ebook"),
+            "c",
+            function(command)
+                return command:gsub("%%c", DocSettings:getSidecarFile(G_reader_settings:readSetting("lastfile")))
+            end
+        },
+        {
             _("directory of current ebook"),
             "d",
             function(command)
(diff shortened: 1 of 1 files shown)


plugins/terminal.koplugin/main.lua
```diff
@@ -13,15 +15,51 @@ local WidgetContainer = require("ui/widget/container/widgetcontainer")
 local logger = require("logger")
 local util = require("ffi/util")
 local _ = require("gettext")
+local BaseUtil = require("ffi/util")
 local N_ = _.ngettext
 local Screen = require("device").screen
 local T = util.template
 
+local function getPlaceholder(label)
+    return string.lower(string.sub(label, 1, 1))
```
poire-z, line 24: Best to get rid of that function (which get the first char of "KOReader dir" and lowercase it to have `%k` working) and just add the explicite char in your placeholders table:
```lua
    placeholders = {
        {
            -- is it wise to translate these labels? It would have to have consequences for the placeholder strings:
            "directory of current ebook",
+           "d"
            function(command)
                return command:gsub("%%d", BaseUtil.dirname(G_reader_settings:readSetting("lastfile")))
            end
        },
```

plugins/terminal.koplugin/main.lua
```diff
@@ -120,6 +169,63 @@ function Terminal:updateItemTable()
...
+        if self.command:match("%%" .. placeholder) then
+            local substitution = v[2]
+            self.command = substitution(self.command)
+        end
+    end
+end
+
+function Terminal:commandHandler(commands)
+    self.command = self:ensureWhitelineAfterCommands(commands)
+    self:substitutions()
```
poire-z, line 185: Please use a verb :)  like `self:substitutePlaceHolders()`
And better passe it the string, which makes it clearer:
`self.command = self:substitutePlaceHolders(self.command)`
and better to make it a local function then
`self.command = substitutePlaceHolders(self.command)`

poire-z:
I don't mind much you frankenstein'ing this plugin (as you might be the only one using it that much :) but the "show file size" option to add "-lh" to any "ls" feels a bit excessive - as it's a personal option (me, my prefered ones are "-ltra").
Next, you'll be adding "show df in kilo|mega" do add -k or -m to "df" :/

smartscripts-nl:
> I don't mind much you frankenstein'ing this plugin (as you might be the only one using it that much :) but the "show file size" option to add "-lh" to any "ls" feels a bit excessive - as it's a personal option (me, my prefered ones are "-ltra").
> Next, you'll be adding "show df in kilo|mega" do add -k or -m to "df" :/

I was merely testing the waters... 😄  I already had a notion this reaction would come. Allright, then I'll remove it later today and keep it in my personal version.

Hzj-jie:
Are you interested to merge it into https://github.com/koreader/contrib?
You may integrate the plugin with https://github.com/koreader/koreader/blob/master/frontend/ui/plugin/insert_menu.lua to make it visible in "More Tools" menu if users decided to install it themselves. An example can be found at https://github.com/koreader/contrib/tree/main/findhistory.koplugin

smartscripts-nl:
Hello Hzj_jie,

I've stopped contributing to the KOReader repository. Do you need an action from me to execute the merge? Otherwise you are welcome to do the merge yourself, with my blessings!

Issue #6107, opened 2020-05-01
Hide default fonts in test document


CalbertAmu:
* KOReader version: 2020.04.1-12
* Device: KA1

A request to create an option to hide the default fonts from appearing in the Font menu,
such as Droid Sans Mono, Free Sans, Noto Sans, etc.

I'm testing different typefaces to see how they appear on e-ink. Was hoping there was a way to hide some unused fonts. I can't delete them because they are used by the system.

However it's not an issue.

pazos:
@CalbertAmu 

I didn't notice your edition.

> I'm testing different typefaces to see how they appear on e-ink. Was hoping there was a way to hide some unused fonts. I can't delete them because they are used by the system.

disable default fonts on test document sounds good!

> A request to create an option to hide the default fonts from appearing in the Font menu

Here my reaction to https://github.com/koreader/koreader/issues/6107#issue-610583384 :smile:

poire-z:
> disable default fonts on test document sounds good!

Whether in the menu or the test document, it's a bit too user specific to bother making that an option.
You might not want to see all our shipped fonts, but I might still want to see NotoSerif - and an arabic reader would not care, like us, about NotoSansCJK, but might want to see NotoSansArabic...

But for the font test document, you can help yourself with a user styletweak:
```css
#Droid_Sans_Mono, #Droid_Sans_Mono + div,
#Noto_Sans_Arabic_UI, #Noto_Sans_Arabic_UI + div,
#Noto_Sans_Devanagari_UI, #Noto_Sans_Devanagari_UI + div,
#Noto_Sans_CJK_SC, #Noto_Sans_CJK_SC + div,
#Dummy_that_should_not_get_a_comma_at_the_end
{ display: none; }
```
Apply that once to your `font-test.html`, and it will keep working when you regenerate it.

pazos:
> > disable default fonts on test document sounds good!
> 
> Whether in the menu or the test document, it's a bit too user specific to bother making that an option.
> You might not want to see all our shipped fonts, but I might still want to see NotoSerif - and an arabic reader would not care, like us, about NotoSansCJK, but might want to see NotoSansArabic...

Good point. I played a bit last night to exclude all embedded fonts, with:

```diff
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -10,6 +10,7 @@ local MultiConfirmBox = require("ui/widget/multiconfirmbox")
[...]

Issue #4174, opened 2018-08-19
Font weight


robert00s:
Is there any way to read the weight of a given font in cre?
Some fonts have several types of thickness but in KOReader see only main name e.g. Phenomena

![zaznaczenie_169](https://user-images.githubusercontent.com/22982594/44313229-ba773880-a404-11e8-9c38-6f1fa7a8498b.png) 

![obraz](https://user-images.githubusercontent.com/22982594/44313292-4047b380-a406-11e8-831f-7e2d74e7168c.png)
Linux Mint font manager see main name: Phenomena and Style(weight): Bold
![zaznaczenie_170](https://user-images.githubusercontent.com/22982594/44313323-c3690980-a406-11e8-99e0-c9ae44d18429.png)

It would be a good idea to use this to change the font thickness.

KenMaltby:
I have my "Stone Serif" font and also the "Stone Serif Semi Bold" on my KOReader fonts list.  They are the Stone Serif Regular.ttf and 2Stone Serif Semi Bold Regular.ttf font files, that I have in the /freefont folder..

robert00s:
After adding one line:
```
*** before/lvfntman.cpp	2018-08-20 09:46:54.888993947 +0200
--- after/lvfntman.cpp	2018-08-20 08:43:33.300993947 +0200
***************
*** 734,739 ****
--- 734,740 ----
          faceName << " " << face->style_name;
      else if ( /*faceName == "Arial" &&*/ face->style_name && strstr(face->style_name, "Condensed") )
          faceName << " " << "Condensed";
+     faceName << " " << face->style_name;
      return faceName;
```
we have: 
![romeo and juliet - koreader_023](https://user-images.githubusercontent.com/22982594/44327511-25555d80-a45f-11e8-9778-04c8400bcce2.png)

Of course we shouldn't show all fonts in menu `Font` but why not in bottom config or another place as option.

poire-z:
I really don't know much about fonts, but looking at crengine code, I would came to understand the following (which may be wrong :):

Font files advertise their family name, style name, bold and italic status.
I don't see that they advertise their weight (100, 200...900) - no API in freetype doc -  and anyway, crengine sets the weight to 400 if no bold flag, to 700 if bold flag.
(crengine, later, when dealing with css `font-weight:`, is able to do some arithmetic with weight to find the nearest font, and is able to embolden a regular font of 400 to 700 when it doesn't find a bold font of 700 for that family name.)

[...]

PR #6738, opened 2020-10-01
Frontend for mupdf / kopt update

ezdiy:
Relating to:

https://github.com/koreader/koreader-base/pull/1203 and https://github.com/ezdiy/koreader-fonts

This is a starting point of a "branch" that builds koreader with new mupdf.

Regarding font stuff:

~The idea here is to delegate to mupdf for near-universal set of fallback fonts. But why? New mupdf comes with no brutalist font rerouting the old patch used to do. Meaning fallback noto sans fonts for exotic languages are now expected to actually exist. This isn't as heavy as it looks (adds about 5M, almost nothing compared to ginormous CJK).~

Too much rocking of the boat. [The approach taken now](https://github.com/ezdiy/mupdf/blob/koreader/source/fitz/noto.c#L70) is make mupdf *try* use our own fonts, and complain about it in log, so that one could troubleshoot what's wrong (mupdf tends to happpily fall on its face as it expects builtin fonts to be well .. builtin). This means things need to be fixed up on case-by-case basis, but we can avoid the bloat of carrying all fonts.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6738)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "koreader-base"]
 	path = base
-	url = https://github.com/koreader/koreader-base.git
+	url = https://github.com/ezdiy/koreader-base
 [submodule "platform/android/luajit-launcher"]
 	path = platform/android/luajit-launcher
 	url = https://github.com/koreader/android-luajit-launcher.git
@@ -13,7 +13,7 @@
 	shallow = true
 [submodule "resources/fonts"]
 	path = resources/fonts
-	url = https://github.com/koreader/koreader-fonts.git
+	url = https://github.com/ezdiy/koreader-fonts
 	shallow = true
 [submodule "l10n"]
 	path = l10n
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -55,6 +55,13 @@ local CreDocument = Document:new{
         "Noto Sans",
     },
 
+    -- lua patterns
+    blacklist_fonts = {
+        "/urw/",
+        "/sil/",
+        "/nerdfonts/symbols%.",
+    },
+
     default_css = "./data/cr3.css",
     provider = "crengine",
     provider_name = "Cool Reader Engine",
@@ -94,6 +101,13 @@ function CreDocument:cacheInit()
         G_reader_settings:readSetting("cre_storage_size_factor") or default_cre_storage_size_factor)
 end
 
+function CreDocument:isFontBlacklisted(name)
+    for _, v in ipairs(self.blacklist_fonts) do
+        if name:match(v) then return true end
+    end
+    return false
+end
+
 function CreDocument:engineInit()
     if not engine_initialized then
         require "libs/libkoreader-cre"
@@ -106,7 +120,7 @@ function CreDocument:engineInit()
         -- we need to initialize the CRE font list
         local fonts = FontList:getFontList()
         for _k, _v in ipairs(fonts) do
-            if not _v:find("/urw/") and not _v:find("/nerdfonts/symbols.ttf") then
+            if not self:isFontBlacklisted(_v) then
                 local ok, err = pcall(cre.registerFont, _v)
                 if not ok then
                     logger.err("failed to register crengine font:", err)
--- a/frontend/document/pdfdocument.lua
+++ b/frontend/document/pdfdocument.lua
@@ -19,8 +19,21 @@ local PdfDocument = Document:new{
     provider_name = "MuPDF",
 }
 
+local function userFont(ctx, name, bold, italic, exact)
+end
+
+local function cjkFont(ctx, name, ordering, serif)
+end
+
+local function fallbackFont(ctx, script, language, serif, bold, italic)
+end
+
+
 function PdfDocument:init()
-    if not pdf then pdf = require("ffi/mupdf") end
+    if not pdf then
+        pdf = require("ffi/mupdf")
+        pdf:setFontCallbacks(userFont, cjkFont, fallbackFont)
+    end
     -- mupdf.color has to stay false for kopt to work correctly
     -- and be accurate (including its job about showing highlight
     -- boxes). We will turn it on and off in PdfDocument:preRenderPage()
@@ -190,19 +203,26 @@ function PdfDocument:saveHighlight(pageno, item)
     self.is_edited = true
     -- will also need mupdf_h.lua to be evaluated once
     -- but this is guaranteed at this point
-    local n = #item.pboxes
-    local quadpoints = ffi.new("float[?]", 8*n)
-    for i=1, n do
-        -- The order must be left bottom, right bottom, left top, right top.
-        -- https://bugs.ghostscript.com/show_bug.cgi?id=695130
-        quadpoints[8*i-8] = item.pboxes[i].x
-        quadpoints[8*i-7] = item.pboxes[i].y + item.pboxes[i].h
-        quadpoints[8*i-6] = item.pboxes[i].x + item.pboxes[i].w
-        quadpoints[8*i-5] = item.pboxes[i].y + item.pboxes[i].h
-        quadpoints[8*i-4] = item.pboxes[i].x
-        quadpoints[8*i-3] = item.pboxes[i].y
-        quadpoints[8*i-2] = item.pboxes[i].x + item.pboxes[i].w
-        quadpoints[8*i-1] = item.pboxes[i].y
+    local quadpoints = ffi.new("fz_quad[?]", #item.pboxes)
+    for i, pb in ipairs(item.pboxes) do
+        quadpoints[i-1] = ffi.new("fz_quad", {
+            ul = {
+                x = pb.x,
+                y = pb.y,
+            },
+            ur = {
+                x = pb.x + pb.w,
+                y = pb.y,
+            },
+            ll = {
+                x = pb.x,
+                y = pb.y + pb.h,
+            },
+            lr = {
+                x = pb.x + pb.w,
+                y = pb.y + pb.h,
+            },
+        })
     end
     local page = self._document:openPage(pageno)
     local annot_type = C.PDF_ANNOT_HIGHLIGHT
@@ -211,9 +231,9 @@ function PdfDocument:saveHighlight(pageno, item)
     elseif item.drawer == "underscore" then
         annot_type = C.PDF_ANNOT_UNDERLINE
     elseif item.drawer == "strikeout" then
-        annot_type = C.PDF_ANNOT_STRIKEOUT
+        annot_type = C.PDF_ANNOT_STRIKE_OUT
     end
-    page:addMarkupAnnotation(quadpoints, n, annot_type)
+    page:addMarkupAnnotation(quadpoints, #item.pboxes, annot_type)
     page:close()
 end
 


poire-z:
Mhhh, many comments upcoming :)

A) dunno how NimbusSans looks, and how it will change our feeling of the UI. Can you provide a few screenshots (menu with checkmarks and default/fallback symbols, wikipedia lookup result...)

B) we have thought quite a bit about the order of our fallback fonts, and it might matter in various context (symbols used in the UI, chapter leading symbol in Wikipedia lookup results...)

C) might need to check how many things look in the UI, like checkmarks in menu: a small shift in the font's line height or baseline might screw the good look we managed to have

D1) I'm not fond of depending on MuPDF for the choice and update of fonts. I rather depend on @NiLuJe :) We might have in the past tweaked some font files - we would now not be able to do so as easily
D2) Well, depending on MuPDF for fonts leaves some work out of our hands :)

F1) We have a small set of fallback fonts that was good enough. FreeSerif has a lot of coverage and was good enough as the last one.
F2) Adding all the other NotoSans* might feel nice, but it might cause a few issues:
F3) base/xtext.cpp (used to shape text for the UI) currently has a limit of 16 fallback fonts. More fallback fonts might increase the processing time and the amount of instantiated font objects.

G1) crengine would be fed all these fonts, which would be given back and listed in the Fonts> menu - which would be too crowded. We prevent that on Android, which has many such fonts (we have a toggle to use or not use them).
G2) most people would prefer Serif fonts for reading - while all the MuPDF exotic fonts are Sans
G3) with crengine, fortunately, we can tweak which fonts it will use as fallbacks, so we're a bit less impacted on this side:
https://github.com/koreader/koreader/blob/6b4188074ece1d87e9b9b07a14134fb5dcee8908/frontend/document/credocument.lua#L33-L56

So, may be we could ship the fonts provided by MuPDF (@NiLuJe might know more if it's a good idea or not), but for how/what we use in frontend, it might be safer to just keep using what we have :)
(And possibly have another menu in the UI for CRE documents to list all the other fallback fonts, and allow users to select which one they want to use and have fed to crengine so they are available in the main font menu.)

ezdiy:
@poire-z 
Reader_2020-10-01_123222
FileManager_2020-10-01_123135
FileManager_2020-10-01_123102
FileManager_2020-10-01_122826

roshavagarga:
@ezdiy A kind reminder, that FreeSerif supports more locales/regional variations of certain alphabets. Point in case, Noto Sans does not support what is colloquially known as Bulgarian Cyrillic, while FreeSerif does. For those interested, other varieties are Serbian and Macedonian, while standard Cyrillic follows, if memory serves well, the Russian reforms for a long time ago, which became mainstream.

@poire-z On a separate note, do we have any way to check supported glyphs versus languages in an automated manner? I'm only thinking of the UI stuff from translations.

ezdiy:
@roshavagarga Freeserif is still a second choice right after latin based scripts. Noto should be relegated to a scope of far more exotic things.

poire-z:
> do we have any way to check supported glyphs versus languages in an automated manner? I'm only thinking of the UI stuff from translations.

Nope we don't (but well, might as well let that to users that use these languages to report it :)

> Screens https://imgur.com/a/LFbLuoX

Somehow, the bold looks like the synthetized bold we get with xtext/harfbuzz (advances of the regular glyphs kept, just the shape are emboldened, so the space between glyphs is a bit constrained - but personally, I like it with our Noto Sans).
Are you sure Nimbus Bold is used? You can test if there's a difference by manually adding `use_bold_font_for_bold=false` into your settings.reader.lua :
https://github.com/koreader/koreader/blob/6b4188074ece1d87e9b9b07a14134fb5dcee8908/frontend/ui/font.lua#L30-L32

> Overall the little aesthetic difference with nimbus/charis there is I'd describe as "edgy 2000s, windows/android 10 era", whereas noto/freefont is more like "round 90s/android 6 era".

Well, I can't really get any impression with your screenshots :) Could you package a zip file with the new fonts, the updated font.lua (and any other things needed) that we could just unzip over a nighytly on our devices, to quickly get a feel of how it would look - so if it looks really nice, you'll gain more followers :)

> (can be kept in just for ui & cre, but mupdf won't have any of it unless specificly asked for as a family in epub)

Well, as a first step, I'd rather not touch UI and cre. And I don't care much about MuPDF, but it behaving more like upstream MuPDF is a target goal I guess (it currently behaving differently is not wished, just a side effects of shipping less fonts).

>  I'd prefer to have a single set of serif+sans fallback fonts (oldschool noto+freeserif, or newer nimbus+charis), instead of mixing both unless there's a good reason to.

I don't know anything about nimbus/charis. Are they carefully and regularly updated like Google may care about Noto ?

> I was somewhat confused by it. Of course we can add b/i variants for devanagari, hebrew and ... the 30 most commonly used scripts. It just struck me odd why arabic in particular, is there readability problem with line-pitched autobold renderer is ought to do in there (apart from being ugly, but so it is for all scripts)?

It's just that we had more feedback/tests from Arabic users when we added support for RTL one year ago. As it was quite some work and we got nice results, we might as well add the necessary fonts (SansUI for the UI, Nasqh variants for CRE cause they prefer it for text), so it's perfect (and so the Bold, so it doesn't even have to be "a little ugly" on this side).
For other languages, no feedback. I'm not sure we have a single devanagari reader (we added it because it covers billions on virtual users :) For Hebrew, @yparitcher must be using his own fonts - and we might be fine with FreeSerif otherwise. For other "rarer" languages, well, waiting for any request.

ezdiy:
> G1) crengine would be fed all these fonts, which would be given back and listed in the Fonts> menu - which would be too crowded. We prevent that on Android, which has many such fonts (we have a toggle to use or not use them).

The list of *our* fonts is trivial, since technically we have like what, mere 7 faces?  - charis serif, nimbus roman, nimbus sans, noto sans, noto serif, freesans, freeserif. Listing files in that menu is of course barbaric.

poire-z:
> Listing files in that menu is of course barbaric, it must be grouped by typeface prefix.

We don't list files. We give files to crengine, which sort/categorize them to put r/b/i/bi in font family slots, and we get back from it font family names, that we put in the menu (cf https://github.com/koreader/koreader/issues/4174#issuecomment-698520071, https://github.com/koreader/koreader/issues/6107#issuecomment-672712604)
I'd prefer to keep a one-level font menu, and not submenu for `Noto Sans>Devanagari|ArabicUI|Nasqh` :) 
Although if we end up using all the Noto Sans variations, it may make sense, but for this Noto Sans only.

> Guess I'll just try to franken the fonts together and roll back ui to original and see how much bloat ensues. From what I can see it won't be all that bad

OK, safer first step :) But I would still like to see how a Nimbused KOReader would look in real life:

> Could you package a zip file with the new fonts, the updated font.lua (and any other things needed) that we could just unzip over a nighytly on our devices, to quickly get a feel of how it would look

poire-z:
Btw, does that mean that MuPDF, which if I remember right was making .o files from font files to have them statically linked, does not do that anymore and just ship its font files as real untouched font files?

ezdiy:
@poire-z https://github.com/ezdiy/mupdf/commit/23dcc8b4ad120df63c9273461391db02219f9b87

> Could you package a zip file with the new fonts, the updated font.lua (and any other things needed) that we could just unzip over a nighytly on our devices, to quickly get a feel of how it would look

Will do with the franken attempt + put some toggle in there to flip between cut/uncut ordering.

ezdiy:
Soooo ... the whole story with mupdf caring about fonts starts kinda falling apart. I was wondering why its fonts in ui look kinda smudged, turns out they're using unhinted fonts. Which make.s kinda sense, it's a fallback after all. So might turn out out koreader will carry all the noto fonts it already has authoritatively - *overriding* select few mupdf overlaps with higher quality hinted ones.

NiLuJe:
Yeah, that's *hard* veto on touching the UI on my end ;). The symbols font also happens to be very specifically tailored for Noto Sans, and hand-tweaked in a couple places.

And a softer veto on touching CRe, as, as was mentioned, that would bloat the font list, and mess with user's font choices (e.g., Charis is fairly popular, and often hand-tweaked).

I also hate Nimbus' look with a fiery passion, which doesn't help ;).
Charis is fine, but probably has coverage issues vs. FreeSerif (which, yeah, happens to be dog-ugly, but we rarely actually see it display things other than symbols).

ezdiy:
@NiLuJe 

>I also hate Nimbus' look with a fiery passion, which doesn't help ;).

Ye, the affinity or fiery hate for halvetica is a very religious issue in typesetting ^_^

> Yeah, that's a _hard_ veto on touching the UI on my end ;). The symbols font also happens to be very specifically tailored for Noto Sans, and hand-tweaked in a couple places.

Yea figured as much in the end. Mupdf *can* be made mostly happy with franken mongrel of noto from koreader, though I'll try to keep em all as unhinted opentype and see how horrible it looks (faster to load, and 6MB->3MB and on high DPI e-ink i can't tell the difference, LCD it can be seen).

The big guy in town here is CJK though. Koreader uses ancient Noto Sans CJK SC. This is good for menus, but not all that helpful for books. Whereas mupdf uses modern [Source Han Serif](https://github.com/adobe-fonts/source-han-serif) which is great for documents, but awkward for UI. On one hand, it's nice to have both CJK sans and serif now, with mupdf running with fonts it expects. On another this results maybe +12MB bloat on release zip. Whether this is a reasonable default ultimately depends on how many cn/kr/jp users we have (I have no idea).

The changes to font.lua now in current state effectively nop for majority of scripts except the odd balls - in those cases it recycles the noto ranges for scripts that mupdf brings touched whenever the range falls through freefont/cjk.

NiLuJe:
I generally don't have an issue with unhinted fonts (I'm running unhinted CRe settings on anything > 250DPI), but it *might* be more noticeable for UI stuff?

The old Noto CJK is on purpose, newer ones have positioning issues with some punctuation marks, IIRC.
Aren't the CJK bits of Noto sourced from Source Han anyway?

IIRC, the last time we asked, Serif for CJK made zero practical sense, so we dropped the idea entirely.

NiLuJe:
Meaning I'd keep at the very least Noto Sans, and the existing UI variants hinted if they already are.

FreeSerif/FreeSans is already CFF, I kept the old file extension to avoid duplicate files after an updates and because I couldn't be arsed to update the patch ;p.

ezdiy:
@NiLuJe Some are (like the arabic ones), some aren't, some were even outright poorly compressed with unshared glyphs (i just look fc-cache output, not actualy glyphs so not really sure). The Noto CJK font should work with hopefuly only light messing with mupdf to look elsewhere. The point is to not make this daily occurence and end up with that insane patch it was before.

yparitcher:
> For other languages, no feedback. I'm not sure we have a single devanagari reader (we added it because it covers billions on virtual users :) For Hebrew, @yparitcher must be using his own fonts - and we might be fine with FreeSerif otherwise. For other "rarer" languages, well, waiting for any request.

I do indeed use my own font for Hebrew [SBL hebrew](https://www.sbl-site.org/educational/BiblicalFonts_SBLHebrew.aspx).
the default one works fine for plain text, but with diacritics it messes up a little, but I like SBL better anyway.

NiLuJe:
I'd possibly blacklist the bundled Charis from CRe, but this looks much better, yeah :).

ezdiy:
@NiLuJe I'm still not really a fan of gutting full language support to basic 8 scripts or so, and then slapping SC-cjk next to it that's 10x larger than all other languages combined, yet isn't actually the OTC CJK mupdf is expecting. As it stands, "weird" scripts (such as hangul) pdfs  *will* open cause we fool mupdf by giving it bogus fonts, but will display wrong.

But that's for another day and religious discussion - since the route is "screw what mupdf thinks is best" the defaults should be at least reasonable :>

NiLuJe:
IIRC, we somewhat recently considered the OTC, and decided against it, I don't really remember why? Quite possibly as unnecessary?

ezdiy:
@NiLuJe Possibly size, OTCs tend to be lil bit more bloaty (16 vs 18mb). Note that the issue isn't glyphs as such (except the noto that's used being kinda gutted compared to full source han sans), but the differences in cmap handling. Without collection, mupdf runs into trouble of discerning what's missing where and what to default to.

poire-z:
> I'm still not really a fan of gutting full language support to basic 8 scripts or so

We have many other scripts supported by FreeSans and FreeSerif (from Hebrew to Runes :) 

> then slapping SC-cjk next to it that's 10x larger than all other languages combined, yet isn't actually the OTC CJK mupdf is expecting. As it stands, "weird" scripts (such as hangul) pdfs will open cause we fool mupdf by giving it bogus fonts, but will display wrong.

Dunno about MuPDF - but SC-cjk contains Traditional Chinese, Japanese and Hangul glyphs. By default, localized variations of the same glyphs will be the SC ones - but the SC_cjk contains all the variations (and crengine and the UI/xtext requests and gets the good ones by specifying the opentype locl as specified by lang= tags, the Typography> lang selected, or the UI language set).
I don't know if that's not the best for MuPDF - and if it ships a fake collection with the 4 SC/TC/JA/KO fonts so it doesn't need to use opentype features.

> IIRC, we somewhat recently considered the OTC, and decided against it, I don't really remember why? Quite possibly as unnecessary?

It was not an OTC question at first, but bugs with the latest NotoSansCJK from github. The ones downloadable from Google noto site are older ones, but don't have bugs.
(Another issue might have been that being an OTC, we got the 4 font names in CRE font list - and we figure we can just do with having one and using opentype locl features via the Typography> setting and lang=tags).
But the bugs with the latest GitHub noto fonts are huge.
One mentionned at: https://github.com/koreader/koreader-fonts/pull/6#issuecomment-619373571
An other I think I didn't write about is that for some simple diacritic combination (don't remember which, but let's say A + ^), the advance/shifts for ^ were wrong in NotoSansCJK - It's fine when it's not the first font, but it being one of the first fallback made it be used for these when then main font happen to not have them.
So, 2 noticable bugs with github's newer NotoSans - and not there with the Google NotoSans made us stay with the Google one.
I don't know which version of NotoSans MuPDF is shipping.
(But given what you discovered about mix of hinted/unhinted fonts - I'd rather not trust at all MuPDF fonts for our main fonts - so we don't get to discover bugs elsewhre (UI or cre) when we update MuPDF :)

NiLuJe:
@poire-z: Thanks, that's what U vaguely remembered (and what I meant by "unnecessary": we get most of the benefits of the OTC by our use of OpenType features without the downside of having 4 explicit fonts listed. And the "fallback" (SC) matches our most prominent CJK user base anyway).

NiLuJe:
And, yeah, I don't think we currently have a coverage issue.

We handle LGC, CJK, Arabic & Indic, and possibly quite a few bits of more exotic stuff. I don't think we're really missing anything with a significant user base?

(And the size issue is pretty specific to CJK, adding Devanagari support cost us less than a MB).

poire-z:
I've tested this on my Kobo: https://r12a.github.io/scripts/phrases
From https://github.com/koreader/crengine/pull/339#issue-408536015 , only issues are with:
> Balinese, Khmer, Mongolian, Myanmar, Telugu, Tibetan (and hyeroglyphs for Ancient egyptian :)
And possibly Urdu which should be a bit tilted arabic and expects a "nastaliq" font we don't provide - and that it seems we/HB wouldn't render very well if provided.

ezdiy:
@poire-z 

> (Another issue might have been that being an OTC, we got the 4 font names in CRE font list - and we figure we can just do with having one and using opentype locl features via the Typography> setting and lang=tags).

This is bug on our side. Barbaric (as in oldschool and showing what *really* is with no regard for UX) font viewers do this too - listing font files instead of typefaces.

> An other I think I didn't write about is that for some simple diacritic combination (don't remember which, but let's say A + ^), the advance/shifts for ^ were wrong in NotoSansCJK - It's fine when it's not the first font, but it being one of the first fallback made it be used for these when then main font happen to not have them.

Ye, gotchas like this are common with no collection.

> So, 2 noticable bugs with github's newer NotoSans - and not there with the Google NotoSans made us stay with the Google one.

This is why nobody uses note, and stay with han sans (google noto is simply han from 2015 or so, modern noto is a divergent fork since).

> (But given what you discovered about mix of hinted/unhinted fonts - I'd rather not trust at all MuPDF fonts for our main fonts - so we don't get to discover bugs elsewhre (UI or cre) when we update MuPDF :)

The mixed hinting exists only in koreader fonts. All mupdf fonts are unhinted.

poire-z:
> This is bug on our side. Barbaric (as in oldschool and showing what really is with no regard for UX) font viewers do this too - listing font files instead of typefaces.

I don't get that. If a font collection provides multiple fonts, shouldn't we show them ? Or allow setting some precedence/order (as they may all provide at least some sets of same glyphs).
Are Noto Sans and Noto Sans Arabic the same "typeface" ?
You may consider these all the same typeface, but we should allow selecting which to use from the UI , as characteristics (glyphs for ascii chars, numbers, line-heights...) might be different (barbaric if you want :) , but practical).

> Ye, gotchas like this are common with no collection.

I don't get that either :) Where is the link between wrong glyph positionning and collections ?

> This is why nobody uses noto, and stay with han sans (google noto is simply han from 2015 or so, modern noto is a divergent fork since).

No opinion on that. If Han Sans provides SC/TC/JA/KO glyphs via OpenType locl features, and it has no bug :) , and our CJK readers are fine with it - I think I wouldn't mind switching.

> The mixed hinting exists only in koreader fonts. All mupdf fonts are unhinted.

Oh, ok (I thought you said at https://github.com/koreader/koreader/pull/6738#issuecomment-702202973 Noto Sans Arabic was hinted in MuPDF).

NiLuJe:
> > This is why nobody uses noto, and stay with han sans (google noto is simply han from 2015 or so, modern noto is a divergent fork since).
> 
> No opinion on that. If Han Sans provides SC/TC/JA/KO glyphs via OpenType locl features, and it has no bug :) , and our CJK readers are fine with it - I think I wouldn't mind switching.

Same here ;).

ezdiy:
@poire-z Ooops, edited your comment instead of replied. This is some really horrible ux, dear github :>

> Are Noto Sans and Noto Sans Arabic the same "typeface" ?

My apologies for mixing up jargon - font family. This stems from the modern practice of promoting `<font face=Noto Sans>arabic text</font>` to Noto family,  and `Noto Sans Arabic` subsequently. This exists in most renderers to deal with "old school" of always setting only face.

> You may consider these all the same typeface, but we should allow selecting which to use from the UI , as characteristics (glyphs for ascii chars, numbers, line-heights...) might be different (barbaric if you want :) , but practical).

Oh most definitely. Font selection for document rendering, not ui (I don't think user even has a choice in fonts used for ui in koreader, or do they?).

> I don't get that either :) Where is the link between wrong glyph positionning and collections ?

Without collection you have a single multilingual font, with amalgam language OT rules with preference for one. It should work in theory, in practice, most engines fare much better when they have 4 different fonts for 4 different languages, each font snowflaked with whatever rules it needs.

> No opinion on that. If Han Sans provides SC/TC/JA/KO glyphs via OpenType locl features, and it has no bug :) , and our CJK readers are fine with it - I think I wouldn't mind switching.

It's not about that, but whether it's an amalgam font (with SC or whichever preference), or OTC. That choice is same be it noto or source han.

> The mixed hinting exists only in koreader fonts. All mupdf fonts are unhinted.

Oh, ok (I thought you said at https://github.com/koreader/koreader/pull/6738#issuecomment-702202973 Noto Sans Arabic was hinted in MuPDF). Yup, thats when koreader's font are used to override whatever overlaps with mupdf fonts. Mixed hinting is largely a non issue tbh.

poire-z:
> My apologies for mixing up jargon - font family. 

I'm not really familiar with all these terminology.
It looks like from FreeType, crengine gets face->family_name and face->style_name (Narrow, Condensed). and uses mostly only family_name to tag the font (that we get back to fill the UI Font> menu).

>  (I don't think user even has a choice in fonts used for ui in koreader, or do they?).

Nope. We use the main font from the aliases (cfont, tfont...), and the ones from the hardcoded fallback font list.

> This exists in most renderers to deal with "old school" of always setting only face.
> [...]
> in practice, most engines fare much better when they have 4 different fonts for 4 different languages, each font snowflaked with whatever rules it needs.

The thing is that both crengine and our UI (via xtext) don't know much about languages and fonts, and don't decide to use an appropriate font per language.
That's because of me when integrating Harfbuzz and fallbacks, to keep things simpler (because language detection and font associations is another big mess that I don't want to know about :) 
We give a language (if there is one around) to Harfbuzz and a font, and it decices what to do behind the scene, and we don't want to know how it does it :)

Just so you know:
What out code does is that it starts with the main font, and go pick missing glyphs in fallback fonts, trying any next falback font until the glyph is found.
This is a bit ugly (as for laying out `CJK CJK, HEB'HEB.` (CJK there being 3 CJK glyphs and HEB some hebrew ones :), it will pick the spaces, punctuation and quote from the main font, and the CJK from the 2nd fallback fonts, and the HEB from the 3rd. While ideally, the 1st space should be from the CJK font and the quote from the HEB font.
But it works quite OK in practice (and I have no desire to change that :) it had been enough of an adventure)

So, we expect fonts to be fine and not to dissilimar (regarding line height and baseline), and if they have opentype features, HarfBuzz may use them - if not, no worry, it' just a simple font. If buggy opentype features, we'll see the bugs. Having a collection of bugless fonts but with buggy opentype features won't do good to us :)

ezdiy:
> face->family_name

This is odd. Family name *should be* same for "Noto Sans" and "Noto Sans Arabic" - both are just "Noto Sans", but each font announcing different languages. Note that font selection is hot mess everywhere, and usually works something like this:

1. If face is supplied, it exists, and has the glyph, great.
2. No face exists, but we know its family (either from font that is missing the glyph, or just guess it by face name prefix). Then pick another face in same family, bias by language hint (this is what selects Sans Arabic if user specifies "Noto Sans"), and start from 1.
3. If 2 fails, run the same thing with "superfamily" (Noto Sans -> Sans). That way you select, say, Droid Sans or FreeSans gets selected, if it has the glyph.

> What out code does is that it starts with the main font, and go pick missing glyphs in fallback fonts, trying any next falback font until the glyph is found.

Just yolo selecting fonts on single line by glyph is normal, though the heuristics tend to be more involved. First, it's to deal with stuff like some silly font putting in fallback glyphs (rectangle) an such. But more importantly, it plays mayhem with glyph positioning the more you distance yourself in terms of family, which is why the algorithm tries real hard to stay within one. The algorithm also gets supremely confused when you give it "amalgam" fonts, but it depends how it implements face selection internally - some can discern the internal cmap tables of amalgams, but most don't.

poire-z:
(Not sure which you mean by "The algorithm..." here and there - ours, or any other good one :) 

Anyway, FreeType is quite limited in the info it gives us (we can't know if a font is Sans or Serif, Noto Sans and Noto Sans Arabic are different families - and we would have to do some heuristics on the font family name strings to classify them furthermore) - and we don't want to bring fontconfig to the party :)

But I was just stating how our current algorithms (quite similar in crengine and in the UI) work currently, so you know what we expect from/can do with fonts (and don't go assuming they work differently and expect different things from fonts - or that we could easily adapt said algorithms to comply to changes in fonts).
I really have no plan on spending time tweaking these algorithms (and I'm not really willing anyone to want to change them - as it's really involved, and it works well enough :)

ezdiy:
> But I was just stating how our current algorithms (quite similar in crengine and in the UI) work currently, so you know what we expect from/can do with fonts (and don
't go assuming they work differently and expect different things from fonts - or that we could easily adapt said algorithms to comply to changes in fonts).

It's much simpler, neiether UI nor crengine can't deal with arbitrary supplied fonts, they must be given only extremely limited subset to work consistently (I guess this explains the massive font blacklist of system fonts :). On that note, I'm not sure if loading system fonts is a wise idea at all if there's no effort to actually select said fonts.

poire-z:
>  (I guess this explains the massive font blacklist of system fonts :).

That's for Kindle only - may be because the fonts are obfuscated, or have some twisted hinting or something (@NiLuJe ?)

> On that note, I'm not sure if loading system fonts is a wise idea at all if there's no effort to actually select said fonts.

It's an option on Android: a user can select and use Noto Sans Balinese that way :) or other user font he has set globally on Android (dunno if that's possible).

Frenzie:
It says in the comment. ;-) (So a little bit of both I'd say.)

https://github.com/koreader/koreader/blob/c7f77de72aefa4a23b22201363c53a0c050e2aa7/frontend/fontlist.lua#L9-L11

ezdiy:
> a user can select and use Noto Sans Balinese that way

Yes, manually. It will render Balinese fine then, but will fall back to builtin koreader fonts for anything else. Thats why the heuristic of font selector is done in the first place - user (or document) selects "one font", but what that does really is select a wide filter for all available fonts.

poire-z:
> Yes, manually. It will render Balinese fine then, but will fall back to builtin koreader fonts for anything else. 

Yes, but I'm still just describing what we can do and do :) and what a user can do to help himeself with how our stuff work - and why, thus not perfect, it is good enough:
A KOReader user can still set Noto Sans Balinese as the first fallback font, and still set his prefered latin font as the main one - and the other fallback fonts (Noto Sans CJK, FreeSerif..) will be used for other scripts not found in these first 2 fonts.

ezdiy:
@poire-z I'm quite fine with the approach, as the simplicity is probably worth it (for small amount of fonts). Doesn't make sense to impose the limit anywhere else though (ie mupdf can be given all fonts just fine and its own crappy heuristic can do something with it). It makes sharing the fonts between the two kinda awkward though, see the most recent patch here.

ezdiy:
Apart from some minor issues, this thing should be now buildable/useable for common devices (ie not android :)

You may need to clone from scratch due to multiple stomping on git submodules.

poire-z:
Pinging @ezdiy :(

NiLuJe:
Not quite the right place to discuss this ;).

And, as usual with fonts, the issue is package size. Given the size of even their "lite" package: that's probably a nope.

(Plus, Noto has the advantage of matching our stock UI font).

mergen3107:
Is this PR still relevant?

Frenzie:
@benoit-pierre Do you think there's still something of interest in the font stuff? 

@ezdiy Thank you for looking into it once again. 👍

benoit-pierre:
> @benoit-pierre Do you think there's still something of interest in the font stuff?

I don't know, I have not looked at this PR.

Frenzie:
Closing now. :-)

PR #6743, opened 2020-10-02
Add Developer Tools plugin

smartscripts-nl:
A simple buttondialog to start Developer Tools. In this example I've linked to the terminal and to de crash.log viewer I pushed in another commit.

The idea is that other developers can link other useful tools via this plugin. Or should all tools be made part of this plugin?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6743)

### DIFF
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -77,6 +77,8 @@ local settingsList = {
     calibre_search = { category="none", event="CalibreSearch", title=_("Search in calibre metadata"), device=true,},
     calibre_browse_tags = { category="none", event="CalibreBrowseTags", title=_("Browse all calibre tags"), device=true,},
     calibre_browse_series = { category="none", event="CalibreBrowseSeries", title=_("Browse all calibre series"), device=true, separator=true,},
+
+    show_developer_tools = { category = "none", event = "ShowDeveloperTools", title = _("Show developer tools"), device = true, },
     show_terminal = { category = "none", event = "TerminalStart", title = _("Show terminal"), device = true, },
     edit_last_edited_file = { category = "none", event = "OpenLastEditedFile", title = _("Texteditor: open last file"), device = true, separator = true, },
     favorites = { category="none", event="ShowColl", arg="favorites", title=_("Favorites"), device=true,},
@@ -210,6 +212,7 @@ local dispatcher_menu_order = {
     "calibre_browse_tags",
     "calibre_browse_series",
 
+    "show_developer_tools",
     "show_terminal",
     "edit_last_edited_file",
 
--- a/frontend/ui/elements/reader_menu_order.lua
+++ b/frontend/ui/elements/reader_menu_order.lua
@@ -139,6 +139,7 @@ local order = {
         "synchronize_time",
         "keep_alive",
         "doc_setting_tweak",
+        "show_developer_tools",
         "terminal",
         "----------------------------",
         "plugin_management",
--- a/plugins/developertools.koplugin/_meta.lua
+++ b/plugins/developertools.koplugin/_meta.lua
@@ -1,0 +1,7 @@
+local _ = require("gettext")
+return {
+    name = "developertools",
+    fullname = _("Developer Tools"),
+    description = _("Buttondialog linking to tools for developers"),
+    sorting_hint = "more_tools",
+}
--- a/plugins/developertools.koplugin/main.lua
+++ b/plugins/developertools.koplugin/main.lua
@@ -1,0 +1,52 @@
+-- overview of available modules: http://koreader.rocks/doc/index.html
+local ButtonDialog = require("ui/widget/buttondialog")
+local Event = require("ui/event")
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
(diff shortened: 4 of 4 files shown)


poire-z:
I'd say such a plugin should just plug a menu with subitems, where we could categorize dev tools (view settings, view logs, enable various options) - a ButtonTable would limit a bit the size of what we could add (although a single gesture to get a selected set might be useful, dunno).
Other developpers would just add stuff to this plugin - and your crash log viewer would go into it, with may be other kind of crash.log views like see all, see tail only).
(Terminal could stay the plugin it currently is because it's already quite complex - but if it wouldn't already be here, I wouldn't mind it being in this dev tools plugins.)
Stuff in that plugin that might be useful to be invoked as a gesture could then plug actions into dispatcher (but the hooks for that are not yet available :/).

smartscripts-nl:
I find a gesture to invoke a buttondialog with tools and then quickly select a tool very handy. Navigating through all the submenu's is a bit cumbersome imho.

pazos:
> A simple buttondialog to start Developer Tools. In this example I've linked to the terminal and to de crash.log viewer I pushed in another commit.
> 
> The idea is that other developers can link other useful tools via this plugin. Or should all tools be made part of this plugin?

Developing on device is a bit of madness. There's the emulator for that!

> I think a gesture to invoke a buttondialog with tools and then quickly select a tool is very handy. Navigating through all the submenu's is a bit cumbersome imho.

You can put dev options wherever you want. Just copy filemanager and reader menu orders from `frontend/ui/elements` to `settings` and modify the file as you want. Here's an example with dev options just above exit options:

dev

I don't like this "plugin". It doesn't add value for final users. It doesn't make sense at all to me :P

Frenzie:
I agree with @pazos that you don't have to navigate through menus; you can just put it somewhere more in your face instead.

Galunid:
I agree with what @pazos said, I can't really justify having this plugin, provided we already have `Developer options`, and that they are purposely "hidden" from users, and even those aren't used often. If Terminal/Crashlog viewer are something you use often, you can assign them separate gestures. In my opinion it doesn't provide much value to the developers, neither does it to users.
> Developing on device is a bit of madness. There's the emulator for that!

Emulator pretty much defeats any need for build-in developer tools, it provides better interface and more tools, since you can pipe logs and grep for the interesting parts, run tests. 

OT: I'd suggest you don't open so many pull requests. It's better to open one, get it merged, then move onto the next one. This way, you don't introduce any conflicts, that either you or someone else would have to resolve. In my opinion it's also simpler to review one pull request, instead of multiple ones, since it's less code to familiarize yourself with. You also get the added benefit of learning as you write, thus avoiding someone pointing you the same mistake in multiple PRs. It saves everyone's time, yours included.

Frenzie:
> Emulator pretty much defeats any need for build-in developer tools, it provides better interface and more tools, since you can pipe logs and grep for the interesting parts, run tests.

You can tail the log perfectly fine on a device too of course. ;-)

smartscripts-nl:
> I agree with what @pazos said, I can't really justify having this plugin, provided we already have `Developer options`, and that they are purposely "hidden" from users, and even those aren't used often. If Terminal/Crashlog viewer are something you use often, you can assign them separate gestures. In my opinion it doesn't provide much value to the developers, neither does it to users.
> 
> > Developing on device is a bit of madness. There's the emulator for that!
> 
> Emulator pretty much defeats any need for build-in developer tools, it provides better interface and more tools, since you can pipe logs and grep for the interesting parts, run tests.
> 
> OT: I'd suggest you don't open so many pull requests. It's better to open one, get it merged, then move onto the next one. This way, you don't introduce any conflicts, that either you or someone else would have to resolve. In my opinion it's also simpler to review one pull request, instead of multiple ones, since it's less code to familiarize yourself with. You also get the added benefit of learning as you write, thus avoiding someone pointing you the same mistake in multiple PRs. It saves everyone's time, yours included.

OK, point taken. I already noticed there wasn't coming a lot of appreciation for my suggestions. So I'll stop wasting time.

NiLuJe:
While I generally don't develop on the emulator, that doesn't mean that I *actually* develop on device ;).

I just *test* on it ;). (i.e., I have a one-liner in my shell backlog that scp's anything diff' from upstream/master)

And then I generally have two different shells open, one of which will be a tail -f on crash.log.

Frenzie:
> I just test on it ;). (i.e., I have a one-liner in my shell backlog that scp's anything diff' from upstream/master)

What's it say? Is that something with `git ls-files`/`ls-tree` or some filter on the output of `git diff`?

> And then I generally have two different shells open, one of which will be a tail -f on crash.log.

Yup, me too. Unless we're specifically dealing with the device (e.g., wifi, light, suspend, 
mxcfb) I don't though.

NiLuJe:
On mobile right now, but it's based on diff --names-only

NiLuJe:
`for file in $(git diff --name-only upstream/master) ; do ; [[ -f "${file}" ]] && scp "${file}" "root@kobo:/mnt/onboard/.adds/koreader/${file}" ; done`

Frenzie:
Yeah, that's pretty much exactly what I was thinking, might be annoying to get quite right though. :-P

Issue #5417, opened 2019-09-23
Dictionary order


noembryo:
This is just a question:
Is there a way to change the Dictionary order?

Just added some more dictionaries to KOReader.

My problem is that I can not re-arrange them the way I like.
They _do_ change positions somehow, but I think it has to do with the dict's folders creation date(?)

Now, I know that I can make one the default for the current book, but..

1. I would like to set the default for the entire program, and
2. the way it is now configured:
    - Start with the second dict (my default)
    - Go to first
    - Go to second again (can I skip this?)
    - Finally go to third

I must be missing something here, because all those years I was using just one dictionary, so I don't have much experience with them.. :o)

poire-z:
Seriously ?! :)

![image](https://user-images.githubusercontent.com/24273478/65437411-07a5b500-de24-11e9-82e9-72d035f988e1.png)

poire-z:
Some notes also on the Wiki: https://github.com/koreader/koreader/wiki/Dictionary-support
Technical details in https://github.com/koreader/koreader/issues/4504#issuecomment-457352313.
(So, beware that that trick may not always work - the inode ordering can sometime not follow the order you put things in it - might help to delete the original dict dir and recreate it to start from a fresh inodes chain.)

Frenzie:
Maybe the new sorting widget by @robert00s could be used here? ;-)

poire-z:
That new sorting widget isn't smart enough yet to code in C and patch sdcv to make it accept some ordering hints :)

Frenzie:
Oh, sdcv doesn't care about the order of arguments or something like that? Never mind then (for now). :-)

Issue #5714, opened 2019-12-31
Fuzzy search not working for lookup


abdrau26:
* KOReader version:latest
* Device:kobo clara hd

#### Issue
I have installed many dictionaries. Fuzzy search works for all except one. Fuzzy search is already checked.

##### `crash.log` (if applicable)
`crash.log` is a file that is automatically created when KOReader crashes. It can
normally be found in the KOReader directory:

* `/mnt/private/koreader` for Cervantes
* `koreader/` directory for Kindle
* `.adds/koreader/` directory for Kobo
* `applications/koreader/` directory for Pocketbook

Android won't have a crash.log file because Google restricts what apps can log, so you'll need to obtain logs using `adb logcat KOReader:I ActivityManager:* AndroidRuntime:* *:F`.

Please try to include the relevant sections in your issue description.
You can upload the whole `crash.log` file on GitHub by dragging and
dropping it onto this textbox, or you can use one of these paste services.
Please don't forget to set a high expiration date if you do.
https://pastebin.com https://slexy.org https://paste2.org https://fpaste.org
http://ix.io https://paste.kde.org https://paste.debian.net.

Frenzie:
Could you please give slightly more info? Ideally which dictionary, but for starters what you're searching for and what you're expecting to find.

abdrau26:
Here is a link for the dictionary
http://download.huzheng.org/misc/stardict-AbuBakar_EnglUrduDictionary-2.4.2.tar.bz2
For example when I search "explained", all dictionaries come up with definition of "explain". But the dictionary do not show definition although it contains "explain"

Frenzie:
I suspect that upper case may not be considered in fuzzy search.

![Screenshot_2019-12-31_15-34-45](https://user-images.githubusercontent.com/202757/71624477-15b55f80-2be3-11ea-8edb-4732e3bdab3a.png)

poire-z:
It works when this dict is used alone:
<kbd>![image](https://user-images.githubusercontent.com/24273478/71624575-8ceaf380-2be3-11ea-8cc6-07512a8b94a9.png)</kbd>

It's probably because you have other dicts that have a definition for the exact word "explained".
As an exact definition is found, there's no need to go fuzzy and find other fuzzy definitions in any dict (the one where the exact word was found, and others).

[...]

PR #6751, opened 2020-10-04
Make dictionary lookup order configurable via SortWidget

### LINKED ISSUE #4504: ticking on dictionaries installed could define their lookup order (LIFO)
* KOReader version: v2019.01.1-29-gb629881_2019-01-22
* Device: Kobo Glo HD

#### Issue
"If you'd like to change the order, move dict files out and move them back one by on in the order you want"

#### Steps to reproduce
tap on top => MENU => Lens => Info on dictionary order

#### Suggestion
I understand having a table with arrows to define the lookup order is a whole thing to do. I'd suggest the following logic: I select the dictionaries that I'm interested in - in reverse order; so that 
1) the worst dictionary I select first by ticking checkbox in the "Installed Dictionaries";
2) so-so dictionary I select the second;
3) the dictionary I want the first to show up I select the last;
...this way whatever I select next and next and next takes higher and higher priority.

The order can be marked right in the list of dictionaries with checkboxes, in MENU > Lens icon > Installed dictionaries (3/15), like this:
[_] {00} GNU Collaborative Dictionary of English
[X] {03} English-Russian full dictionary
[X] {01} English-Russian short dictionary
[_] {00} English-Russian slang dictionary
[X] {02} New big English-Russian dictionary

where (to explain the above 5 lines):
[X] - ticked checkbox
[_] - unticked checkbox
{02} - order of the dictionary in the lookup, for {02} two taps on ">>" button will be needed to reach that dictionary in the lookup results. 

After I tick the checkbox, the order could recalculate and the menu could repaint; this way without any explanations the user understands that the priority changed; reverse order is a bit counter-intuitive but this is the price for this cheap & dirty solution (without the need to "move out all folders and drop them back one-by-one").

"last selected dictionary will be used first to bring lookup results" represents LIFO, whereas for FIFO the developer would need to provide "reset all checkboxes" and then first ticked is first in the lookup results, second selected dictionary will show up after ">>" button and the third dict goes to the next ">>" page... FIFO is too much to code...

georgeto:
Previously you on had move files in and out of the dictionary directory to change the order in which dictionaries are queried.
With the changes in this pull request, the dictionary order can instead be configured via a SortWidget based menu.

Since I thought it would be impractical to have two menus for managing dictionaries, one for switching between enable/disable and one for the sorting order, I decided to add checkbox support to SortWidget. The check mark is toggled with a long press.

New dictionaries, i.e. those that were not present when the sort menu was last called, are placed at the end of the dictionary list.

### Breaking changes
The previous implementation relied on file order provided the filesystem implementation. Opposing to that the new implementation sorts the dictionaries by their name.
Users that sorted their dictionaries by moving files in and out of the dictionary directory, will experience their existing dictionary order being messed up.

### Screenshots
<details>
  <summary>Old interface</summary>

  DictBefore
</details>

<details>
  <summary>New interface</summary>

  DictAfter
</details>

---
Resolves  #4504 and #5417

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6751)

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -12,6 +12,7 @@ local KeyValuePage = require("ui/widget/keyvaluepage")
 local LuaData = require("luadata")
 local MultiConfirmBox = require("ui/widget/multiconfirmbox")
 local NetworkMgr = require("ui/network/manager")
+local SortWidget = require("ui/widget/sortwidget")
 local Trapper = require("ui/trapper")
 local UIManager = require("ui/uimanager")
 local ffiUtil  = require("ffi/util")
@@ -28,11 +29,6 @@ local lookup_history = nil
 
 local function getIfosInDir(path)
     -- Get all the .ifo under directory path.
-    -- We use the same logic as sdcv to walk directories and ifos files
-    -- (so we get them in the order sdcv queries them) :
-    -- - No sorting, entries are processed in the order the dir_read_name() call
-    --   returns them (inodes linked list)
-    -- - If entry is a directory, Walk in it first and recurse
     -- Don't walk into "res/" subdirectories, as per Stardict specs, they
     -- may contain possibly many resource files (image, audio files...)
     -- that could slow down our walk here.
@@ -138,6 +134,12 @@ function ReaderDictionary:init()
         end
         logger.dbg("found", #available_ifos, "dictionaries")
 
+        if not G_reader_settings:readSetting("dicts_order") then
+            G_reader_settings:saveSetting("dicts_order", {})
+        end
+
+        self:sortAvailableIfos()
+
         if not G_reader_settings:readSetting("dicts_disabled") then
             -- Create an empty dict for this setting, so that we can
             -- access and update it directly through G_reader_settings
@@ -145,30 +147,42 @@ function ReaderDictionary:init()
             G_reader_settings:saveSetting("dicts_disabled", {})
         end
     end
-    -- Prepare the -u options to give to sdcv if some dictionaries are disabled
+    -- Prepare the -u options to give to sdcv the dictionary order and if some are disabled
     self:updateSdcvDictNamesOptions()
     if not lookup_history then
         lookup_history = LuaData:open(DataStorage:getSettingsDir() .. "/lookup_history.lua", { name = "LookupHistory" })
     end
 end
 
+function ReaderDictionary:sortAvailableIfos()
+    local dicts_order = G_reader_settings:readSetting("dicts_order")
+
+    table.sort(available_ifos, function(lifo, rifo)
+        local lord = dicts_order[lifo.file]
+        local rord = dicts_order[rifo.file]
+
+        -- Both ifos without an explicit position -> lexical comparison
+        if lord == rord then
+            return ffiUtil.strcoll(lifo.name, rifo.name)
+        end
+
+        -- Ifos without an explicit position come last.
+        return lord ~= nil and (rord == nil or lord < rord)
+    end)
+end
+
+
 function ReaderDictionary:updateSdcvDictNamesOptions()
-    self.enabled_dict_names = nil
+    self.enabled_dict_names = {}
 
     -- We cannot tell sdcv which dictionaries to ignore, but we
     -- can tell it which dictionaries to use, by using multiple
     -- -u <dictname> options.
-    -- (The order of the -u does not matter, and we can not use
-    -- them for ordering queries and results)
+    -- The order of the -u options controls the dictionary order
+    -- that sdcv uses to order its results.
     local dicts_disabled = G_reader_settings:readSetting("dicts_disabled")
-    if not next(dicts_disabled) then
-        return
-    end
     for _, ifo in pairs(available_ifos) do
         if not dicts_disabled[ifo.file] then
-            if not self.enabled_dict_names then
-                self.enabled_dict_names = {}
-            end
             table.insert(self.enabled_dict_names, ifo.name)
         end
     end
@@ -214,29 +228,22 @@ function ReaderDictionary:addToMainMenu(menu_items)
         text = _("Dictionary settings"),
         sub_item_table = {
             {
+                keep_menu_open = true,
                 text_func = function()
                     local nb_available, nb_enabled, nb_disabled = self:getNumberOfDictionaries()
                     local nb_str = nb_available
                     if nb_disabled > 0 then
                         nb_str = nb_enabled .. "/" .. nb_available
                     end
-                    return T(_("Installed dictionaries (%1)"), nb_str)
+                    return T(_("Manage dictionaries (%1)"), nb_str)
                 end,
                 enabled_func = function()
                     return self:getNumberOfDictionaries() > 0
                 end,
-                sub_item_table = self:genDictionariesMenu(),
-            },
-            {
-                text = _("Info on dictionary order"),
-                keep_menu_open = true,
-                callback = function()
-                    UIManager:show(InfoMessage:new{
-                        text = T(_([[
-If you'd like to change the order in which dictionaries are queried (and their results displayed), you can:
-- move all dictionary directories out of %1.
-- move them back there, one by one, in the order you want them to be used.]]), BD.dirpath(self.data_dir))
-                    })
+                callback = function(touchmenu_instance)
+                    self:showDictionariesMenu(function()
+                        if touchmenu_instance then touchmenu_instance:updateItems() end
+                    end)
                 end,
             },
             {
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -138,37 +134,55 @@ function ReaderDictionary:init()
...
-    self.enabled_dict_names = nil
+    self.enabled_dict_names = {}
 
     -- We cannot tell sdcv which dictionaries to ignore, but we
     -- can tell it which dictionaries to use, by using multiple
     -- -u <dictname> options.
-    -- (The order of the -u does not matter, and we can not use
-    -- them for ordering queries and results)
+    -- The order of the -u options controls the dictionary order
+    -- that sdcv uses to order its results.
```
poire-z, line 182: Really ?! Does that really work now?
I wrote that comment 2017-08-29.
We last updated sdcv around 2017-10 (to 0.5.2) and I added a few patches related to -u. Did I miss the fact that it would now follow the order of the -u ? :/

Btw, sdcv is now at 0.5.3.
(And re-btw, I still use our faster previous version 0.4 :/)

ezdiy:
On a related note, people playing with this might be interested in https://github.com/ezdiy/libsd to get rid of the general sdcv clunkyness.

frontend/apps/reader/modules/readerdictionary.lua
georgeto: According to the [changelog](https://github.com/Dushistov/sdcv/blob/e73388c726430ad724104b4a16d6c1c9821f35e9/NEWS#L17) and the [corresponding commit](https://github.com/Dushistov/sdcv/pull/11/commits/2d95bd0b12e5c977f1f4cb69ad3b5dc881a318ec) the `dictionary order by -u switches` feature was added in version 0.5.1.

georgeto:
> On a related note, people playing with this might be interested in https://github.com/ezdiy/libsd to get rid of the general sdcv clunkyness.

Mhm, I didn't look at the project in detail, but I noticed that it doesn't have a license, which will probably be a problem if someone wants to integrate it into KOReader.

ezdiy:
@georgeto It's gutted from qstardict (and ported to c), which itself is gutted from stardict 2.3, whose original license is GPL2, thus infecting everything that based its code on it (including sdcv).

The public API is https://github.com/ezdiy/libsd/blob/master/DictEng.h#L72

frontend/apps/reader/modules/readerdictionary.lua
poire-z: OK, it indeed works ! :)

poire-z:
OK, tested, works quite fine :) 

A few comments (not all related to your changes and neither for you to fix, but mentionning them anyway as they're related to the whole user experience) :

A) There might be a few bugs in the SortWidget: I have 22 dicts installed, so it makes multiple pages. When selecting a dict and changing pages, it moves to the top of the new page (which is fine). But the one that was at the top of the new page is moved at the top of the previous page (which is not fine, it should move below the one inserted at top of new page) - or even sometimes disappear (not displayed anywhere - but back when you relaunch the sort widget). This happens too when sorting Status bar items, so it's a pure SortWidget bug.

B) Tap to select a dict to be sorted vs Hold to enable/disable it is a bit tedious. Usually, you would sort your dicts once, and then use that "Manage installed dictionaries" to just enable/disable some.
I would prefer tap to enable/disable and hold to select for sorting (even if it's a bit the reverse of the Status bar item sorting).
May be the change of behaviour in SortWidget could be toggled whether the items (any one item) has a checkmark/is toggable ?
Or at least have Tap on the checkmark do a checkmark toggling instead of item selection?

C) We have a "feature" that seems to ~~then be no more ensured~~ (well, it already didnt work before this PR): when you tap on the title (with the dict name the result currently displayed is from) to make it the "default dictionary for this document".
I do think this feature sucks :) I was too often taping on it and gettting this message when I didn't want that, that I commented it out in my own patch. Just in case, this is what I patch in `ui/widget/dictquicklookup.lua`:
```diff
 function DictQuickLookup:init()
-    self:changeToDefaultDict()
+    -- self:changeToDefaultDict() -- we don't care about defaut dict
+    self:changeDictionary(1) -- always use the 1st
[...]
 function DictQuickLookup:onTapCloseDict(arg, ges_ev)
     if ges_ev.pos:notIntersectWith(self.dict_frame.dimen) then
         self:onClose()
         return true
     elseif not ges_ev.pos:notIntersectWith(self.dict_title.dimen) and not self.is_wiki then
-        self.ui:handleEvent(Event:new("UpdateDefaultDict", self.dictionary))
+        -- we don't care about setting a default dict
+        -- self.ui:handleEvent(Event:new("UpdateDefaultDict", self.dictionary))
         return true
     end
     -- Allow for changing dict with tap (tap event will be first
```
So, it didn't work before this PR, and still doesn't with this PR. Should we remove that bit of code that says something we don't ensure - or fix it to make it work?

D) Probably no fix needed, but mentionning it:
See https://github.com/koreader/koreader/issues/5714#issuecomment-569939052
One (probably only me) can have dicts in a `dict_ext/` subdirectory (sibling to `dict/`), that would be queried in a 2nd query and results shown last. But these dicts still appear in the SortWidget, and you can sort them - but the results will still be shown last (as they should :)
Ideally, they could be unsortable and stay at the end - or just not shown. But no real need to implement that.

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -214,6 +228,7 @@ function ReaderDictionary:addToMainMenu(menu_items)
         text = _("Dictionary settings"),
         sub_item_table = {
             {
+                keep_menu_open = true,
```
poire-z, line 231: May be the menu item text below should be:
```diff
- return T(_("Installed dictionaries (%1)"), nb_str)
+ return T(_("Managed dictionaries (%1)"), nb_str)
```

frontend/util.lua
```diff
@@ -2,9 +2,11 @@
 This module contains miscellaneous helper functions for the KOReader frontend.
 ]]
 
+local ffi = require("ffi")
 local BaseUtil = require("ffi/util")
 local dbg = require("dbg")
 local _ = require("gettext")
+local C = ffi.C
```
poire-z, line 9: I think `util.strcoll()` should probably be moved into `base/ffi/util.lua` where all the stuff needing ffi/C is - so that our `frontend/util.lua` keeps being pure Lua and ffi/C free :)
Might be a bit tedious as we need there to add a check for isKobo() and not sure how that would be done. Or check the ffi strcoll is available or not (if that's the reason for it to not be used on Kobo). And possibly get rid of the global DALPHA_SORT_CASE_INSENSITIVE or does it have a purpose? @Frenzie @NiLuJe .

It means making a PR to base and bumping it with this PR (I can do that in a coming up base/crengine bump if that feels too tedious for you).

NiLuJe: `strcoll` is unusable on Kobo because the system lacks compiled locales.

The check could arguably be made Device-less by checking for the existence of `/usr/lib/locale/locale-archive` instead (c.f., -f /usr/lib/locale/locale-archive).

NiLuJe: But, regardless, +1 that it belongs in ffi/util, not front's util ;).

NiLuJe: I'm actually mildly puzzled on *which* of the platforms we target (besides emulator) where `strcoll` actually behaves sanely: it *requires* compiled locales for the current `LC_COLLATE`, something which I highly doubt is happening for !POSIX on eInk readers...

NiLuJe: And of course, those don't bundle `locale`, so we can't check `locale -a` for the actually compiled locales...

georgeto:
> B) ...
> Or at least have Tap on the checkmark do a checkmark toggling instead of item selection?

A good idea, I implemented it!

frontend/util.lua
NiLuJe: On the other hand, we never *actually* set locales to anything other than `en_US.UTF-8`, so just checking for the default locales archive ought to be enough, because if locales are compiled in, there's a good chance that that includes en_US.UTF-8 ;).

ezdiy: If it's just to sort dictionary book names, lexical utf8 suffices because this seems like more about some sort of canonical more than anything and unicode point order will "reasonably" sort this alphabetically even in weird scripts.

Heavyweight strcoll stuff is necessary when case gets involved (stricoll), then you need full on accent mark tables.

georgeto:
> A) There might be a few bugs in the SortWidget: I have 22 dicts installed, so it makes multiple pages. When selecting a dict and changing pages, it moves to the top of the new page (which is fine). But the one that was at the top of the new page is moved at the top of the previous page (which is not fine, it should move below the one inserted at top of new page) - or even sometimes disappear (not displayed anywhere - but back when you relaunch the sort widget). This happens too when sorting Status bar items, so it's a pure SortWidget bug.
> 

I think it is not a bug, but a feature, albeit a very unintuitive one...

If you switch pages while an item is selected, the item is moved to the top of the target page.
I don't know what to think about this feature. On the one hand it can be useful, but on the other hand it is rather confusing.
https://github.com/koreader/koreader/blob/a50774003f8ca670b783347e6761949b4682ef9c/frontend/ui/widget/sortwidget.lua#L366-L386

poire-z:
> I think it is not a bug, but a feature, albeit a very unintuitive one...

I have no issue with this being a feature (that I'd consider useful) - it's the side effects I mentionned (switch items rather than move selected one above existing one, or items disappearance) that I have an issue with :)

frontend/util.lua
NiLuJe: @ezdiy: Yep, this lived in filemanager because it was basically only used by the "Sort By Title" option in the FM until this PR ;).

And, to answer @poire-z's other query: `DALPHA_SORT_CASE_INSENSITIVE` is still enabled by default.

georgeto:
> I have no issue with this being a feature (that I'd consider useful) - it's the side effects I mentionned (switch items rather than move selected one above existing one, or items disappearance) that I have an issue with :)

You are right, my bad, the SortWidget should now work as expected :)

poire-z:
Thanks :) (I didn't read the SortWidget code and didn't see it was actually really doing a swap - but indeed better to do a move with table.insert(...table.remove()).)

NiLuJe:
@georgeto: I took care of the strcoll move to ffi/util, so you'll basically just have to rebase to drop the duplicate changes, and switch to `ffiUtil.strcoll` instead of `util.strcoll` for the new stuff ;).

(Once #6753 is merged ;)).

frontend/apps/reader/modules/readerdictionary.lua
georgeto: What about `Manage dictionaries (%1)`, I think this fits better to the following `Download dictionaries` menu option.

poire-z (approved):
Works for me - and the SortWidget bugs I witnessed earlier seem also fixed.

PR #6752, opened 2020-10-05
ReaderFooter: Don't generate separators for empty generators

NiLuJe:
I'm half of a mind to drop that behavior from the Bookmarks generator, because I feel like the generator itself deciding to go into hiding breaks some implicit contract with the user that explicitly asked to display it...

But, it's been implemented that way from the start, so, for the sake of backward compatibility, eh.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6752)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -164,9 +164,8 @@ local footerTextGeneratorMap = {
     bookmark_count = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].bookmark_count
-        --retrieve bookmark count:
         local bookmark_count = footer.ui.bookmark:getNumberOfBookmarks()
-        -- if no bookmarks defined, don't show icon:
+        -- Don't show anything if there aren't any bookmarks
         if bookmark_count == 0 then
             return ""
         end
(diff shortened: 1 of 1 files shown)


NiLuJe:
(xref: https://www.mobileread.com/forums/showthread.php?t=333670)

poire-z:
> I feel like the generator itself deciding to go into hiding breaks some implicit contract with the user that explicitly asked to display it...

( For the record, I never expected nor agreed with this implicit contract :)
I would be perfectly fine with some items that I asked for to appear only when they make sense, when some state has changed (there is always a time, we are always in some chapter, there is always a battery level or a memory usage ... but there might be no bookmark, wifi might be disabled).
)

NiLuJe:
It feels *extremely* counter-intuitive to me, but, the main actual issue I had with it were the stray separators, which is now taken care of ;).

KenMaltby:
I would think that a user is looking to be able to keep a check on those items/conditions that they select for display on the "status bar".  Being able to establish if an item is off/not available, and that the device is aware of that, can be of some value to the user.

Luck;
  Ken

NiLuJe:
I mean, I, personally, may not like that behavior, but I can understand other PoV on the subject ;).

So I guess the only way to achieve that in a sane manner would be a new setting to toggle that "hide me if zero/off" behavior on a *global* scale.

Because the current inconsistency kind of drives me nuts ^^.

poire-z:
> So I guess the only way to achieve that in a sane manner would be a new setting to toggle that "hide me if zero/off" behavior on a global scale for all the generators where it makes sense.

Agreed that this would be the way to go (not that I personally need it).

NiLuJe:
#6754 ;)

PR #6754, opened 2020-10-05
ReaderFooter: Make the wonky "hide me if I'm off" behavior consistent and user-configurable.

NiLuJe:
Make the wonky "hide me if I'm off" behavior consistent and user-configurable.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6754)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -911,7 +911,7 @@ function ReaderFooter:addToMainMenu(menu_items)
             },
             {
                 text = _("Hide empty items"),
-                help_text = _([[This will hide an item from the status bar if it were to display 0 or an off state.]]),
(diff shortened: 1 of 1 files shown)


NiLuJe:
Applies to FL, BM, Wi-Fi

poire-z (approved):

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -898,6 +909,17 @@ function ReaderFooter:addToMainMenu(menu_items)
                     self:refreshFooter(true, true)
                 end,
             },
+            {
+                text = _("Hide empty items"),
+                help_text = _([[This will hide an item from the status bar if it were to display 0 or an off state.]]),
```
Frenzie, line 914: Maybe something short and to the point like:

This will hide values like 0 or off.

PR #6755, opened 2020-10-05
Font menu: show font names with their own font. Bump crengine

### LINKED ISSUE #6717: German hyphenation: Quotation mark is speparated from related word
* KOReader version: v2020.09-24
* Device: Phone with Anroid 10 and PBInkPad3

#### Issue
In a book with 'unusual' german quotation maks the quotaion mark is separted from its related word by line breaks.

grafik

### LINKED ISSUE #6718: Line break at the footnote number
Shouldn't we avoid line breaks before the footnote numbers?
Reader_2020-09-27_095627

### LINKED ISSUE #6739: FR: Font preview in fonts menu
When choosing fonts it would be great to get a preview of what a font looks like. A solution other reading software (Kindle, Kobo, Pocketbook, Tolino) already has implemented is to display the name of the font written in the font itself.
See Pocketbook font menu as an example:

poire-z:
Includes https://github.com/koreader/crengine/pull/385 :
- TextLang: German: loosen quotation marks handling - Closes #6717.
- Fonts: add LVFontManager::getFontFileNameAndFaceIndex()
- fb2.css: avoid break before footnote links - Closes #6718.
- FlowState: fix possible wrong baseline with nested inline-table
- Tables: fix possible layout issues with rowspan
- Table cells: minor refactoring of baseline handling
- getRenderedWidths(): better estimate table width
- lvstyles: allow for more than 64 styles

Font menu: show font names with their own font
See https://github.com/koreader/koreader/issues/6739#issuecomment-703131167. Closes #6739.

Any rewording suggestions for the following and the setting name `font_menu_use_font_face` ?

<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6755)

### DIFF
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -3,6 +3,7 @@ local CenterContainer = require("ui/widget/container/centercontainer")
 local ConfirmBox = require("ui/widget/confirmbox")
 local Device = require("device")
 local Event = require("ui/event")
+local Font = require("ui/font")
 local Input = Device.input
 local InputContainer = require("ui/widget/container/inputcontainer")
 local Menu = require("ui/widget/menu")
@@ -74,6 +75,14 @@ function ReaderFont:init()
                 end
                 return text
             end,
+            font_func = function(size)
+                if G_reader_settings:nilOrTrue("font_menu_use_font_face") then
+                    local font_filename, font_faceindex = cre.getFontFaceFilenameAndFaceIndex(v)
+                    if font_filename and font_faceindex then
+                        return Font:getFace(font_filename, size, font_faceindex)
+                    end
+                end
+            end,
             callback = function()
                 self:setFont(v)
             end,
@@ -84,7 +93,6 @@ function ReaderFont:init()
                 return v == self.font_face
             end
         })
-        face_list[k] = {text = v}
     end
     self.ui.menu:registerToMainMenu(self)
 end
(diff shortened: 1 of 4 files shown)


Frenzie (commented):
Preview font through name?

poire-z:
> Preview font through name?

I think it's totally uneeded :)
"The quick fox..." would show only letters, which is the thing I care the less about when appreciating fonts: letters are usually coherent. What I care more about are numbers, punctuations, line heights, and we can get a better feel about that with the "Generate font test document".
Oh, and well, we can also apply the font to the current book for a better feel :)
And most of the times, there are fonts we selected and know a bit about.

Also, when long-press, we get "Would you like Ahem to be used...", so we have a chance to see the font name in the normal UI font in case it's unreadable in its own font.

poire-z:
> Preview font through name?

Oh, that was a rewording suggestion for the menu item :) 
I thought you were mentionning your https://github.com/koreader/koreader/issues/6739#issuecomment-703145195 :/

`Preview font through name` could work, a bit not explicite enough may be?

Frenzie:
`Show preview of fonts` perhaps?

poire-z:
> `Show preview of fonts` perhaps?

Still not explicite enough :) One might expect to see these preview in a popup, etc... rather than in the menu item. `Preview font through name` is a bit more explicite then :)

NiLuJe:
I'm mostly fine with @poire-z's wording, perhaps using "display" instead of "show".

Otherwise, it gets tricky real quick, e.g., something based on `Render font names` or something...

NiLuJe:
Or, to combine a bit of everything, `Preview font rendering in font names`

poire-z:
Not fond of "Render", a bit too technical a this level of the menus :) "Display" is fine:
<kbd>image</kbd>

Grammar: Display ... _with/in_ their own font   ?

Galunid:
Works great, thanks! :)

PR #6756, opened 2020-10-05
[RFC]: logviewer

pazos:
I would like to try hacking on top of #6608 in the emulator, but the emulator lacks a proper crash.log

If this patch doesn't sound stupid I can include it on top of the changes and enable the crash.log viewer on the emulator too. If it sounds stupid I will keep it to myself, hack the code on the emulator, and disable the crashlog viewer on all platforms without a crash.log

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6756)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -17,11 +17,18 @@ local util  = require("util")
 local _ = require("gettext")
 local T = FFIUtil.template
 
+local crashlog = require("ui/elements/logviewer"):new{ file = "crash.log" }
+
 local FileManagerMenu = InputContainer:extend{
     tab_item_table = nil,
     menu_items = {},
     registered_widgets = nil,
 }
+
+function FileManagerMenu:onShowCrashlog()
+    crashlog:show()
+    return true
+end
 
 function FileManagerMenu:init()
     self.menu_items = {
@@ -300,6 +307,14 @@ function FileManagerMenu:setUpdateItemTable()
             },
         }
     }
+    if crashlog:exists() then
+        table.insert(self.menu_items.developer_options.sub_item_table, {
+            text = _("Show crash.log"),
+            keep_menu_open = true,
+            callback = function() crashlog:show() end,
+        })
+    end
+
     if Device:isKobo() then
         table.insert(self.menu_items.developer_options.sub_item_table, {
             text = _("Disable forced 8-bit pixel depth"),
--- a/frontend/logger.lua
+++ b/frontend/logger.lua
@@ -10,7 +10,12 @@ Example:
 ]]
 
 local dump = require("dump")
+local isSDL = require("ffi/util").isSDL()
 local isAndroid, android = pcall(require, "android")
+
+-- write to crash.log on SDL platform based on environment.
+-- on other platforms crash.log is generated redirecting stdout in init shell scripts.
+local logfile = isSDL and os.getenv("EMULATE_CRASH_LOG") and io.open("crash.log", "w+")
 
 local DEFAULT_DUMP_LVL = 10
 
@@ -60,8 +65,13 @@ local function log(log_lvl, dump_lvl, ...)
             android.LOGE(line)
         end
     else
-        io.stdout:write(os.date("%x-%X"), " ", LOG_PREFIX[log_lvl], line, "\n")
+        local msg = string.format("%s %s%s\n", os.date("%x-%X"), LOG_PREFIX[log_lvl], line)
+        io.stdout:write(msg)
         io.stdout:flush()
+        if logfile then
+            logfile:write(msg)
+            logfile:flush()
+        end
     end
 end
 
--- a/frontend/ui/elements/logviewer.lua
+++ b/frontend/ui/elements/logviewer.lua
@@ -1,0 +1,287 @@
+--[[ This module implements a log viewer.
+
+ Its intended usage is for crash.log, but you can repurpose it for your own file.
+ All logviewer instances share the same settings but you can override them if you wish.
+
+]]--
+
+local Font = require("ui/font")
+local TextViewer = require("ui/widget/textviewer")
+local UIManager = require("ui/uimanager")
+local util = require("util")
+local _ = require("gettext")
+local T = require("ffi/util").template
+
+local BUTTONS_IN_ROW = 4
+local MIN_FONT_SIZE, MAX_FONT_SIZE = 6, 40
+local MIN_FILE_SIZE, MAX_FILE_SIZE = 4 * 1024, 512 * 1024
+
+local function getSize(file)
+    if not file then return end
+    local handle, err = io.open(file, "r")
+    if handle then
+        local bytes = handle:seek("end")
+        handle:close()
+        return bytes
+    end
+end
+
+local function scaleFont(size, direction)
+    if type(size) ~= "number" then return end
+    if size < MAX_FONT_SIZE and direction == "up" then
+        return size + 1
+    elseif size > MIN_FONT_SIZE and direction == "down" then
+        return size - 1
+    else
+        return size
+    end
+end
+
+local function scaleBytes(file, bytes)
+    local full_size = getSize(file)
+    local max_size = full_size <= MAX_FILE_SIZE and full_size or MAX_FILE_SIZE
+    
+    if type(bytes) ~= "number" or bytes <= 0 then
+        if full_size >= MIN_FILE_SIZE then
+            return MIN_FILE_SIZE
+        else
+            return nil
+        end
+    else
+        local result = bytes * 2
+        if result >= max_size then
+            result = nil
+        end
+        return result
+    end
+end
+
+local LogViewer = {}
+
+function LogViewer:new(o)
+    setmetatable(o, self)
+    self.__index = self
+    return o:init(o.file)
+end
+
+function LogViewer:init(file)
+    if not file then return nil, "file not specified" end
+    local handle = io.open(file, "r")
+    if handle then
+        handle:close()
+        self._exists = true
+    end
+
+    -- default settings if there's no override
+    for k, v in pairs({
+        font_face = "infont",
+        font_size = 10,
+        tail_bytes = 512 * 1024,
+    }) do
+        if not self[k] then
+            self[k] = v
+        end
+    end
+    return self
+end
+
+-- getters and setters for G_reader_settings
+function LogViewer:getFontSize()
+    return G_reader_settings:readSetting("logviewer_font_size") or self.font_size
+end
+
+function LogViewer:setFontSize(size)
+    G_reader_settings:saveSetting("logviewer_font_size", size or self.font_size)
+end
+
+function LogViewer:getTailBytes()
+    return G_reader_settings:readSetting("logviewer_tail_bytes") or self.tail_bytes
+end
+
+function LogViewer:setTailBytes(bytes)
+    G_reader_settings:saveSetting("logviewer_tail_bytes", bytes or self.tail_bytes)
+end
+
+function LogViewer:exists()
+    return self._exists
+end
+
+function LogViewer:truncate()
+    if not self._exists then return end
+    UIManager:show(require("ui/widget/confirmbox"):new{
+        text = T(_("Really truncate %1?"), self.file),
+        ok_callback = function()
+            local handle, err = io.open(self.file, "w")
+            if not handle then return nil, err end
+            handle:write(T(_("%1 [%2 truncated]"), os.date("%x-%X"), self.file))
+            handle:close()
+        end,
+    })
+end
+
+function LogViewer:read(tail_bytes)
+    local handle, err = io.open(self.file, "r")
+    if not handle then return nil, err end
+    local tailOnly = type(tail_bytes) == "number"
+
+    if tailOnly then
+        handle:seek("end", -tail_bytes)
+    end
+
+    local content = handle:read("*all")
+    handle:close()
+
+    if tailOnly then
+        if content:len() == tail_bytes then
+            local first_cr = content:find("\n")
+            if first_cr then
+                content = content:sub(first_cr + 1)
+            end
+            content = T(_("[Start of %1 not shown]\n"), self.file) .. content
+        else
+            content = T(_("[Start of %1]\n"), self.file) .. content
+        end
+    end
+
+    return content
+end
+
+function LogViewer:show()
+    if not self._exists then return end
+    self:update()
+end
+
+function LogViewer:getTitle(bytes)
+    local path, fname = util.splitFilePathName(self.file)
+    if fname == "" then
+        fname = path
+    end
+
+    if bytes and bytes > 0 then
+        size = string.format("Last %dKB", bytes/1024)
+    else
+        size = string.format("All %dKB", getSize(self.file)/1024)
+    end
+    return string.format("%s (%s)", fname, size)
+end
+
+function LogViewer:update(bytes, size, filter)
+    if not size then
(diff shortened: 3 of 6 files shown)


frontend/logger.lua
```diff
@@ -11,6 +11,11 @@ Example:
 
 local dump = require("dump")
 local isAndroid, android = pcall(require, "android")
+local isEmulator, crashlog = require("ffi/util").isSDL()
+    and not os.getenv("APPIMAGE")
+    and not os.getenv("KO_MULTIUSER")
+    and not os.getenv("UBUNTU_APPLICATION_ISOLATION")
+    and true, io.open("crash.log", "a+") or false
```
Frenzie, line 18: This isn't very clear to read at a glance imo. Also, wouldn't it make a lot of sense for Android too? Presumably much more so than for the emulator in fact, since anyone running the emulator will have all kinds of tools at their disposal while most Android users probably don't have adb. ;-)

Frenzie (commented):
The concept seems fine by me.

frontend/logger.lua
pazos: Yup. Having a crash.log probably makes sense on all platforms that don't  have one.

In the case of android having a crash.log could make sense for *some* things, but we are going to miss all logs that use `android.LOG()/__android_log_print` directly and just redirect the ones that use the logger wrapper.

Most of the android errors that are logged to adb are catched in https://github.com/koreader/android-luajit-launcher/blob/master/jni/android-main.c#L117

pazos: So not sure about android, but if it is ok to have a crash.log on all sdl platforms we can make the thing simpler

```lua
local crashlog =  require("ffi/util").isSDL() and io.open("crash.log", "a+")
```

Frenzie: > In the case of android having a crash.log could make sense for some things, but we are going to miss all logs that use android.LOG()/__android_log_print directly and just redirect the ones that use the logger wrapper.

Yeah, that's fair.

> So not sure about android, but if it is ok to have a crash.log on all sdl platforms we can make the thing simpler

I'm talking more about what almost looks like hiding what's going on to me. Write it like this and it's clear at a skim; you don't have to hunt for the comma.
```suggestion
local isEmulator = require("ffi/util").isSDL()
    and not os.getenv("APPIMAGE")
    and not os.getenv("KO_MULTIUSER")
    and not os.getenv("UBUNTU_APPLICATION_ISOLATION")
    and true
local crashlog = io.open("crash.log", "a+") or false
```

But looking at it again, I think what you may have actually meant was:

```suggestion
-- not a function ;-)
local is_emulator = require("ffi/util").isSDL()
    and not os.getenv("APPIMAGE")
    and not os.getenv("KO_MULTIUSER")
    and not os.getenv("UBUNTU_APPLICATION_ISOLATION")
    and true
local crashlog = is_emulator and io.open("crash.log", "a+") or false
```

poire-z:
> if it is ok to have a crash.log on all sdl platforms we can make the thing simpler

For me it's absolutely not ok :)
I don't need at all a crash.log. I launch/relaunch the emulator so often that its (debug) logs would make an enormous crash.log that I don't want (and don't even want to have rotated/truncated at some size like on Kobo).
So, at minima, make it disable'able via another environment variable - and don't kill the regular stdout/stderr output :)

As for hacking on top of this other PR, I'd rather you start from scratch - or just add a few items to `Developer tools`, it doesn't need to be a plugin.
This other PR is too targeted at plugins writers (that don't crash the app, and thus, you can read/grep the errors from the still running app). Its reformatting/displaying feels like no value added for non-plugins hacks/devs. (and go at keeping updated the list of strings to ignore :/).
For info, in my own patch, I add [these 3 menus items](https://github.com/koreader/koreader/pull/6608#issuecomment-683713275) to have a crash.log viewer. After a real crash, I can just see the tail with all the context of what happens. Filtering kills that context.

NiLuJe:
I do agree that it, at the very least, requires the same "trim if > 500KB" logic than our startup scripts.

Then again, I oh so very rarely use the emulator that I'm not really the target audience ;).

Frenzie:
> For me it's absolutely not ok :)
I don't need at all a crash.log.

I'm on board with that. I don't see the need. But I suppose it wouldn't bother me much.

> I don't need at all a crash.log. I launch/relaunch the emulator so often that its (debug) logs would make an enormous crash.log that I don't want (and don't even want to have rotated/truncated at some size like on Kobo).

I noticed that too but forgot to mention it.

https://github.com/koreader/koreader/blob/0fa45c21cfdb6e7f3d5676235a2b35c95f9a674f/platform/kobo/koreader.sh#L257-L261

pazos:
What about:

don't write to crash.log if crash.log doesn't exists:
    - If the user wants to view the crash log on these platforms it needs to create a new, empty one.
    - use "w+": update mode, all previous data is erased; to prevent enormousness on crash.log

> As for hacking on top of this other PR, I'd rather you start from scratch

black box paradigm. I would like a similar "UX" output. I will try to filter by LOG_LVL instead of having including/excluding lists.

NiLuJe:
> 
> 
> What about:
> 
> don't write to crash.log if crash.log doesn't exists:
> - If the user wants to view the crash log on these platforms it needs to create a new, empty one.
> - use "w+": update mode, all previous data is erased; to prevent enormousness on crash.log
> 

Seems like an elegant alternative that fits the intended design, :+1: ;).

poire-z:
> don't write to crash.log if crash.log doesn't exists:
> - If the user wants to view the crash log on these platforms it needs to create a new, empty one.
> - use "w+": update mode, all previous data is erased; to prevent enormousness on crash.log

I could be fine with that.
But it's a bit too implicit. Copy a crash.log from an issue there to investigate, and hop, you get them unwillingly (and the user crash.log is erased).

Why not just a new `kodev run` option?  `./kodev run --crashlog` or `-l` or `-c` so we (including me) can get them when needed (currently, I just do `./kodev run | tee out.log` which is fine enough).

pazos:
@poire-z: 

> Why not just a new kodev run option? ./kodev run --crashlog or -l or -c so we (including me) can get them when needed (currently, I just do ./kodev run | tee out.log which is fine enough).

Agree, `./kodev run | tee out.log` is fine enough to not bothering with a new kodev option.

> I could be fine with that.
But it's a bit too implicit. Copy a crash.log from an issue there to investigate, and hop, you get them unwillingly (and the user crash.log is erased).

I would stay with:

> If it sounds stupid I will keep it to myself, hack the code on the emulator, and disable the crashlog viewer on all platforms without a crash.log

I'm not personally affected by the lack of logs. Hell I even disable most debug statements while doing some things. My intent is to address #6608 (and maybe the idea behind #6743) without making them plugins. Now I'm starting to wonder if it makes sense at all or the time is better spent into documentation, creating a new label "rejected", use the label on these PRs and point to documentation there.

That's the beauty of RFC when there're comments. We can all change our minds :smile:

ezdiy:
> We can all change our minds

[I have no strong feelings one way or the other](https://www.youtube.com/watch?v=ussCHoQttyQ)

But if I "had to": One would expect the emulator to do the same as it does on device - ie no built in log writer until controller script actually redirects/tees the stdout from the outside. So it's a bit odd that the behavior would be built in just for emulator, as it violates "least surprise" principle. The lack of ability to tee seems specific to windows. Can koreader really build on windows? And are there even people out there using that?

For standalone emulator binary distribution without kodev, there would be need for a controller script anyway (to setup LD_LIBRARY and stuff I guess, just like kodev does?). I'd expect such a script ("koreader.sh"?) to redirect the same just as it does on device, and perhaps even "emulate" things like restart exit codes.

pazos:
Nah, current logic is cool. We redirect stdout to a file on embedded targets where is not easy to get logs. On sdl targets just running the program on the console will output everything. There's no need to handle that on a shell script (even more, on some platforms there's no init shell script, like on mac).

> We can all change our minds

It is linked to

> Now I'm starting to wonder if it makes sense at all or the time is better spent into documentation, creating a new label "rejected", use the label on these PRs and point to documentation there.

Maybe a bit passive-aggresive rant about how to handle PR's like the two mentioned above. I don't need them. I think nobody needs them for various reasons, but mostly: "there's always a better way to debug on each platform, here (finger points to docs) they are"

poire-z:
> Now I'm starting to wonder if it makes sense at all or the time is better spent into documentation, creating a new label "rejected", use the label on these PRs and point to documentation there.

Or a label "User-Plugin" for plugins we don't feel like they should be shipped - but are available for users who'd like to try them (and it's the author's job to keep them updated - not ours so we don't have to maintain complicated stuff).

pazos:
> Or a label "User-Plugin" for plugins we don't feel like they should be shipped - but are available for users who'd like to try them (and it's the author's job to keep them updated - not ours so we don't have to maintain complicated stuff).

I'm not sure if these plugins are ok as PR in the main repo. Still not sure how to implement standalone plugins.

@poire-z: I've implemented your suggestions in a way that can be repurposed to view other files too. Not sure about the prototype yet. Maybe we can add the scroll to bottom as part of the function signature?.

poire-z:
> I'm not sure if these plugins are ok as PR in the main repo.

I was just suggesting tagging them as "User-Plugin" and closing and not caring much anymore - as totally the same as tagging them as "Rejected", but in a more polite/passive-passive/we-ll-sleep-well way :) 

> I've implemented your suggestions in a way that can be repurposed to view other files too. Not sure about the prototype yet. Maybe we can add the scroll to bottom as part of the function signature?.

I'm fine with that and scroll to bottom as a parameter.
(A bit hard at first to get into your code, must be the short `M` and `M:new` I'm not used to meet :)

frontend/ui/elements/log_viewer.lua
```diff
@@ -11,56 +14,109 @@ local function checkFile(f)
...
+        buttons_table = {
+            {
+                {
+                    text = "<<",
+                    callback = function()
+                        self.textviewer.scroll_text_w:scrollToTop()
+                    end,
+                },
+                {
+                    text = "<",
```
Frenzie, line 84: Don't forget our nice arrow symbols. ;-)

pazos:
Sin nombre

@poire-z: is it possible to reload the text or to change text size without recreating the widget?

frontend/ui/elements/log_viewer.lua
pazos: a link would be awesome. I tried to grep sources but didn't spend much time :)

poire-z: https://github.com/koreader/koreader/blob/6b90239e7c6bc72c13efd54cbffc14bbbbce2c77/frontend/apps/reader/modules/readersearch.lua#L147-L150

https://github.com/koreader/koreader/blob/6b90239e7c6bc72c13efd54cbffc14bbbbce2c77/frontend/ui/widget/inputdialog.lua#L744-L753

poire-z:
> is it possible to reload the text or to change text size without recreating the widget?

Currently not.
Might need to do the kind of thing we do in other widgets: split half of TextViewer:init() into TextViewer:update() (and re-create `self.scroll_text_w = ScrollTextWidget:new()` in there) - and call this update() from the button that change size or reload the text.

Or if you don't really want/need to tweak TextViewer itself: just tweak it here that way:
Have `self.textviewer = TextViewer:new{...)` done in a new `LogViewer:createAndShowTextViewer(text, size)`, and in your buttons:
```lua
text = "Update",
callback = function()
  UIManager:close(self.textviewer)
  -- reload text, update self.size
  self:createAndShowTextViewer(text, size)
end

```
But does it really matter there ? The crash.log won't change much as long as you're there reading crash.log and no other thing can happen :) But ok, increasing/decreasing font size might be interesting.

pazos:
> But does it really matter there ? The crash.log won't change much as long as you're there reading crash.log and no other thing can happen :) But ok, increasing/decreasing font size might be interesting.

Yeah, mostly font size changes and toggle between INFO and DEBUG log levels. Thanks for all the help :)

pazos:
to-do: arrows, review strings

I added shortcuts to dispatcher. Maybe we can create a new category "developer" instead of plugin them in device.

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -300,6 +317,46 @@ function FileManagerMenu:setUpdateItemTable()
             },
         }
     }
+    if crashlog.exists then
```
poire-z, line 320: Reading this feels like a typo :) We mostly always wrap that and call a function.
So may be store the boolean as `_exists` and have a `function zzz:exists() return _exists end`

frontend/dispatcher.lua
```diff
@@ -210,6 +214,10 @@ local dispatcher_menu_order = {
     "calibre_browse_tags",
     "calibre_browse_series",
 
+    "show_crashlog",
+    "show_taillog",
+    "truncate_crashlog",
```
poire-z, line 219: > Maybe we can create a new category "developer" instead of plugin them in device.

I'm fine with them being in Device (no need for another Developer category that most people don't care about - unless we end up adding more?).
Even push them at the end tail of Device, below everything else (below terminal, lasteditedfile, favorites....)

frontend/logger.lua
```diff
@@ -10,8 +10,12 @@ Example:
 ]]
 
 local dump = require("dump")
+local isSDL = require("ffi/util").isSDL()
 local isAndroid, android = pcall(require, "android")
 
+-- emulate crash.log on SDL platform
+local emulated = isSDL and os.getenv("EMULATE_CRASH_LOG") and io.open("crash.log", "w+")
```
poire-z, line 17: May be name it `emulated_crash_log` (`emulated` sounds like it is a boolean).

frontend/ui/elements/logviewer.lua
```diff
@@ -0,0 +1,150 @@
...
+local LogViewer = {}
+
+function LogViewer:new(o)
+    o.exists = checkFile(o.file)
+    setmetatable(o, self)
+    self.__index = self
+    return o
+end
+
+function LogViewer:savedLines()
```
poire-z, line 35: why "_saved_ lines" ? Do you mean "_tailed_ lines" ?

frontend/dispatcher.lua
poire-z: And "show_crashlog_tail" rather than "show_taillog".

pazos: >  unless we end up adding more?

That's the point. IMHO it is better a category than most people ignore than a few options that clutter the category most people use.

Also, some of them can be added from https://github.com/koreader/koreader/pull/6608#issuecomment-683762231

poire-z: May be they don't need to be all available to dispatcher/gestures. show_crashlog_tail might be the most interesting candidate. Full crash log and truncate might have very little use via gestures.
As for the others I might be using, it's so rare I need them that just having them hidden in a far away menu is fine enough :)

frontend/ui/elements/logviewer.lua
pazos: what about `getTail()` and `setTail(lines)`

poire-z: The name "getTail()" sounds like it will return some content, not a number.
so, get/setNbLinesForTail or getNbTailLines ?

NiLuJe: `(g|s)etTailCount`?

pazos:
I played a bit a few days ago with user configurable buttons. I was waiting for @smartscripts-nl feedback because he proposed the initial crash log viewer with some fancy options. As I didn't received feedback yet I just pushed the local branch here (with a bunch of test code, including a dmesg and a whole plugin that's not intended to be reviewed or merged) and let this stale a bit more :)

PR #6759, opened 2020-10-06
Statistics plugin: fix in-memory statistics.

marek-g:
I have found a new bug in statistics plugin. The remaining time to read (both for a book & for a chapter) is decreasing a way too fast when navigating betweeg the same set of pages (for example going forward and backward a few times), especially on newly opened books, so it becomes useless very quickly.

**The reason:**

The `mem_read_pages` is calculated wrongly in `ReaderStatistics:onPageUpdate(pageno)` method for two reasons:
- it counts duplicated page numbers (every visit of the same page as it was a "new page")
- it counts pages even if you spent less time on them than `self.page_min_read_sec`

This fix solves the bug.

---

_P.S. This is my first contribution to KOReader. It's currently my favorite reader app. I'm really excited that it was so easy to fix something myself that was bothering me for some time :-)._

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6759)

### DIFF
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1867,21 +1867,24 @@ function ReaderStatistics:onPageUpdate(pageno)
     local mem_read_pages = 0
     local mem_read_time = 0
     if util.tableSize(self.pages_stats) > 1 then
-        mem_read_pages = util.tableSize(self.pages_stats) - 1
         local sorted_performance = {}
         for time, page in pairs(self.pages_stats) do
             table.insert(sorted_performance, time)
         end
         table.sort(sorted_performance)
+        local read_pages_set = {}
         local diff_time
         for i=1, #sorted_performance - 1 do
             diff_time = sorted_performance[i + 1] - sorted_performance[i]
             if diff_time <= self.page_max_read_sec and diff_time >= self.page_min_read_sec  then
                 mem_read_time = mem_read_time + diff_time
+                read_pages_set[self.pages_stats[sorted_performance[i]]] = true
             elseif diff_time > self.page_max_read_sec then
                 mem_read_time = mem_read_time + self.page_max_read_sec
+                read_pages_set[self.pages_stats[sorted_performance[i]]] = true
             end
         end
+        mem_read_pages = util.tableSize(read_pages_set)
     end
     -- every 50 pages we write stats to database
     if util.tableSize(self.pages_stats) % PAGE_INSERT == 0 then


poire-z:
I'm not familiar with the stats computations, but it looks allright.
Pinging @robert00s (if you have a few minutes to review this - and to be sure the previous behaviour was not intended).

Btw, this (already before, one more with your PR) looks like a lot (?) of (small?) table manipulation/creation/sort on each page turn. Any idea how costly/battery consuming this can be vs all the other stuff we do? If non negligeable, any way to reduce that?

> I'm really excited that it was so easy to fix something myself that was bothering me for some time :-).

Yep, happened to most of us I guess :)

marek-g:
> Btw, this (already before, one more with your PR) looks like a lot (?) of (small?) table manipulation/creation/sort on each page turn. Any idea how costly/battery consuming this can be vs all the other stuff we do? If non negligeable, any way to reduce that?

Yes, you are right. It would be better to change `self.pages_stat` to be the dictionary (page number -> total time spent on that page) and have another variable to keep total time. With that, only one dictionary lookup would be needed every page update.

But it looks like refactoring that touches more methods. Do you think it is in the scope of this bug or should there be another task created for that?

About the impact on battery life I think it is very very negligible as the maximum size of table is 50 elements (then it is swapped to database) that are processed every two minutes or so.

Galunid:
Just to confirm, do you want to update `mem_read_time` regardless whether the page was visited earlier?

marek-g:
> Just to confirm, do you want to update `mem_read_time` regardless whether the page was visited earlier?

Yes. Every further visit adds to the total time of reading the page. This is the same behavior as it was before and I think that's correct. For example:

I visit page #1 one time - for 20 seconds,
I visit page #2 twice - for 10 seconds and 20 seconds for the second time,

the `mem_read_time` will be sum of all the visits = 20 + 10 + 20 = 50 seconds
the `mem_read_pages` will be 2 as there is only page #1 and #2, it was 3 before the fix hence the bug

average page reading time = read time / read pages = 50 / 2 = 25 seconds per page

marek-g:
My use case is - I sometimes look at the next page or at the previous one for a few seconds only. The original code treats it as I had read whole new page in about 2 seconds. Anyway, it was also inconsistent with what is saved to the database later. Stats saved to database are doing:

SELECT sum(period), count(DISTINCT page)

and it is aligned with my change.

poire-z:
> About the impact on battery life I think it is very very negligible 

OK, good then.

> Do you think it is in the scope of this bug or should there be another task created for that?

No, not in the scope of this PR.
But as you might be the only person to put its eyes on this module for some time, if you're getting comfortable with how it all works, and if you see other bugs or possible optimisations, feel free to go at them.

marek-g:
> But as you might be the only person to put its eyes on this module for some time, if you're getting comfortable with how it all works, and if you see other bugs or possible optimisations, feel free to go at them

Sure, thanks! In this case I think the trouble is not worth the profit, but of course, I would love to improve the parts where I can see benefits if I can!

marek-g:
> if you see other bugs or possible optimisations, feel free to go at them

[OFF TOPIC]

Actually, I can see a nice to have fundamental change to statistics module. Maybe I could work on it, but it is much bigger thing and probably needs a little brain storming :)

Statistics are page size (window/font) dependent. They rely on page numbers. The collected data are becoming totally wrong when user change font or window size in the middle of reading (the number of pages in a book changes and the stats are immediately wrong). The page is probably wrong unit to store reading speed. Maybe words per second is better but its not for comics / pdfs. Probably we should record percentage (area) of the book read in time unit and then recalculate it dynamically to current page size....

Sorry for the off topic.

NiLuJe:
@marek-g: Or, just drop the current stats on (significant?) total page count changes?

marek-g:
[OFF TOPIC]

@NiLuJe: Yes, dropping the stats in that case is better than having it wrong. This is an easy solution.

The more I think about it, I think that tracking read parts of the document in percent ranges, not in pages (in real numbers, for example 0.23 - 0.26 percent of the book instead of page number 5) would work independently of page size but it is hard to implement, but also very interesting (finding intersections of ranges when scaling etc). Maybe on some boring winter night... :D

NiLuJe:
Currently playing with indexing by page, instead of by timestamp, because that's the root of the current craziness...

marek-g:
Ok. That's good. Indexing by page will be faster. I think that you need to 
keep list of pairs of (timestamp and duration) as a value.

So I assume this pull request is on hold, is it?

Dnia wtorek, 6 października 2020 15:51:32 CEST NiLuJe pisze:
> Currently playing with indexing by page, instead of by timestamp, because
> that's the root of the current craziness...

NiLuJe:
@marek-g: Yeah, I basically track two arrays (ts & duration), much like the DB itself.

And, no, I'm starting on top of your fix, so I'll merge this first ;).

poire-z:
> just drop the _current_ stats 

Not all the stats for the book I hope ? :) 
Dunno if you're just talking about the "in memory" temp stats before they are flushed to sqlite, or the full sqlite tables you want to redesign (and migrate old stats to :).

(Also thinking about the calendar view: if I read 3 pages at 16h50 - and read the same 3 pages at 17h05, they should generate bars for the 16h slot, and for the 17h slot - not huger bars for the 16h slot.)

> Statistics are page size (window/font) dependent. They rely on page numbers

I really don't mind that (but each of us might probably care more about some statistics than others).

Just a question: with PDF documents, we might scroll/page/pan the same original PDF page: does that make 3 onPageUpdate(same_page_num) or just one? Do the durations on these 3 intermediate views contribute to the duration on the PDF original page? Or is it messed up?

As for EPUB, we could use percentage - but that also has its caveats: some pages might have a smaller slice of the whole document height than others. So, using that and browsing 5 consecutives pages with a small height, it might account like I have read only 1 page.

NiLuJe:
> > just drop the _current_ stats
> 
> Not all the stats for the book I hope ? :)
> Dunno if you're just talking about the "in memory" temp stats before they are flushed to sqlite, or the full sqlite tables you want to redesign (and migrate old stats to :).
> 

That's a good point. I was going to drop the DB stats for the book, too, because I had the "on open" workflow (i.e., essentially meaningless db stats).

But, you're right, only dropping the volatile stats *probably* makes more sense ;).

NiLuJe:
> Just a question: with PDF documents, we might scroll/page/pan the same original PDF page: does that make 3 onPageUpdate(same_page_num) or just one? Do the durations on these 3 intermediate views contribute to the duration on the PDF original page? Or is it messed up?
> 

That'd be `PosUpdate`s events, and the handler in the Stats plugin only fires the PageUpdate if the page has actually changed. So it should be fine.

NiLuJe:
> > (Also thinking about the calendar view: if I read 3 pages at 16h50 - and read the same 3 pages at 17h05, they should generate bars for the 16h slot, and for the 17h slot - not huger bars for the 16h slot.)
> 
> Will depend on timing. The timestamps will be truly coalesced by session. If there are > 50 page turns between the two (or anything else that closes a session, e.g., suspend), you'll get two sessions, otherwise, they'll end up in the first session.
> 
> Losing that bit of granularity there is the only downside of using pages as keys instead of ts. Can only have one ts per page per session ;).
> 
> The DB itself still tracks multiple sessions, though.

I guess one could forcibly closes the session on the hour?

NiLuJe:
> > Losing that bit of granularity there is the only downside of using pages as keys instead of ts. Can only have one ts per page per session ;).
> 
> Well, that's quite some loss, it feels:/ If I re-read the same 20 pages, I didn't mind being told I have read 40 pages. And I'd expect to see in the calendar view hourly bars when I did read the 20+20 pages (and not have them coalesced with the same page I read at the start of the hour).
> 
> If you are only caring the things in the footer (I'm guessing ?), may be you could just have tweaked accounting for that either in the stats plugin, or directly in ReaderFooter (instead of having the Statistics plugin plugging in ReaderFooter).

The page count being inflated on re-reads was the whole point of *this* PR, so, re-reads won't count. The *time* spent re-reading, will, though.

I don't particularly care about any consumers of the stats, mainly about the insanity happening in the PageUpdate handler ^^.

----

I can *fake* a session closure on the hour (I wouldn't really want to tweak the real session start, as that dictates the "on open" timestamp), which should essentially hide the internal implementation from CalendarView, since I'm guessing it only reads the DB?

(SO making sure the DB is flushed on the hour essentially takes care of that).

poire-z:
> The page count being inflated on re-reads was the whole point of this PR

Reading this PR, I get that it just replaces:
`mem_read_pages = util.tableSize(self.pages_stats) - 1`
with
`mem_read_pages = util.tableSize(read_pages_set)`
I haven't read what it does with mem_read_pages :) but it feels like it doesn't change any of the info that will be saved in the DB. It doesn't coalesce any data, it just compute differently the mem_read_pages number.

> I don't particularly care about any consumers of the stats

It you're going to coalesce a set of data, if it's the data in memory that will end up in the DB, we're losing some data (that we have currently, so it's a loss).
The multiple data (page_num, ts_start, duration) for a same page is information. It is currently saved to the db as is. Coalescing before saving is losing it.
I think any need for coalescing is to be done by the user/caller as usage time, if that's what they want. That's what this PR seems to do for me.

> should essentially hide the internal implementation from CalendarView, since I'm guessing it only reads the DB?

I remember we force a flush to DB when invoking it.

I haven't look at the whole stats.plugin code and I don't want to. But this thing about sessions does not talk to me. I don't care about sessions. I know the plugin care about them when displaying some info. But it does look to me that this does not go in the DB.

May be I don't know what I'm talking about - and not sure about what you're doing either :) Will you change the amount of data that will be saved in the DB ?

NiLuJe:
Yeah, I'm talking about my WIP tree, not necessarily this PR, which doesn't help ^^.

marek-g:
Thanks for merging my PR, I'm really happy that this bug won't be present in the next release :)

Issue #6461, opened 2020-08-02
Reading Statistics recording long time for certain pages.


rigogsilva:
* KOReader version: v2020.07.1
* Device: Boox Nova 2

#### Issue
Looking in the statistics.sqlite3 file (table page_stat), I have found some pages to have a higher number of `period` (seconds) recorded for them. See arrows below:
![image](https://user-images.githubusercontent.com/31558765/89126604-7922ab00-d4ac-11ea-98e4-cd8473158c94.png)

Note `page` 154. I think this may be due because I am still learning how to use koreader and playing around with the settings? That is maybe being recorded as time read? 

I took a deep dive into that page to see what records exist for it and here is what I found:
![image](https://user-images.githubusercontent.com/31558765/89126630-c868db80-d4ac-11ea-8b7f-271c86399932.png)

Thoughts on what is happening?

#### Steps to reproduce
Not sure yet. Probably playing with settings while in a page? But that is just a wild guess.

Frenzie:
There's a maximum time somewhere in the settings. I think it defaults to a minute or two but presumably you've set it to a much higher value.

rigogsilva:
Yes, I have set that to 210. But note that it was logged multiple times for the same page, multiple days also. That is in the second screenshot.

poire-z:
The fact that on the 2nd screenshot, everything is below 210, is somehow reasurring we don't have a bug on that period limitation front. If you stay one hour on a page, you get saved 210s. If you move to some next pages, and get back to it, and stay one hour on it, you get saved a new record with 210s.
And on your 1st screenshot, you're showing the cumulative durations, which can go above 210s.

That sounds to me like the logical (and technically simple) behaviour.

May be you expect that if you spend 30s on a page, then move around and go back to that page and look at it again, it's considered "already read" and your second visit should not be accounted ? This might be wrong in some cases, if you first skim quickly (but slowly enough so it gets acccounted), and later do a serious reading when you stay longer on the pages you previously skimmed - you would get not really relevant stats... And your cumulative time reading this book would also be wrong.

So, the real questions is: _but what was on that page 154 , that had you visiting it 19+ times and contemplating it for a cumulative 27 minutes ?!!_
(I hope you'll answer that question, I'm curious :)

poire-z:
> I think this may be due because I am still learning how to use koreader and playing around with the settings? That is maybe being recorded as time read?

[...]

Issue #6454, opened 2020-07-30
Ability to export Reading Statistics data


rigogsilva:
**Does your feature request involve difficulty completing a task? Please describe.**
I am not sure if this is available yet, or maybe you can point me to where the data is stored. I would like to export the reading data so I can do some of device data analyses on it. But there is no button to export that data.

**Describe the solution you'd like**
In the Reading Statistics, add a button to export. Or just tell me what folder the data is and I can get it. Hopefully is in a usable format.

**Describe alternatives you've considered**
None yet.

**Additional context**
None

FYI: I just started using Koreader, so I may not know what I am talking about. Yet.

poire-z:
There's no export button, because we would have to decide about a format, and nobody has ever cared enough to define one - and it depends on what one wants.

But the stats are in a sqlite DB with simple tables, that one can query and export what he wants (assuming he knows a bit of SQL):
```
$ sqlite3 koreader/settings/statistics.sqlite3
sqlite>
sqlite> .schema
CREATE TABLE book
            (
                id integer PRIMARY KEY autoincrement,
                title text,
                authors text,
                notes      integer,
                last_open  integer,
                highlights integer,
                pages      integer,
                series text,
                language text,
                md5 text,
                total_read_time  integer,
                total_read_pages integer
            );
CREATE TABLE sqlite_sequence(name,seq);
CREATE TABLE page_stat
            (
                id_book    integer,
                page       integer NOT NULL,
                start_time integer NOT NULL,
                period     integer NOT NULL,
                UNIQUE (page, start_time),
                FOREIGN KEY(id_book) REFERENCES book(id)
             );
CREATE TABLE info
             (
                 version integer
             );
CREATE INDEX page_stat_id_book ON page_stat(id_book);
CREATE INDEX book_title_authors_md5 ON book(title, authors, md5);

sqlite> .help
.dump ?TABLE? ...        Render all database content as SQL
.headers on|off          Turn display of headers on or off
.log FILE|off            Turn logging on or off.  FILE can be stderr/stdout
.mode MODE ?TABLE?       Set output mode
.nullvalue STRING        Use STRING in place of NULL values
.output ?FILE?           Send output to FILE or stdout if FILE is omitted
.separator COL ?ROW?     Change the column and row separators
```
Here are some SQL used by the Calendar view to get what it needs to draw stuff:
[...]

Issue #6155, opened 2020-05-16
Finer resolution of statistics history for the last few hours


protist:
* KOReader version: 2020.04.1
* Device: Kobo H2O

#### Issue
Attempt to work out how long I have been reading for in the past few hours.

#### Background
The statistics for the "current session" are great, but often I briefly stop my reading, turn my ebook reader off for a few minutes, and then begin reading again. This resets the "current session", and there is no way to tell how long I have been reading for, i.e. ignoring any brief breaks.

#### Possible solutions
1. Is there a way to see when I have been reading over the past few hours? Similar to the calendar view, but with finer resolution?
2. If not, is this something that might be possible to create?
3. Alternatively, in the meantime is there a text file or something that I can access via koreader to check manually?

(This issue is related to #2978, but is a slightly different problem and possible solution.)

poire-z:
No text file, but your start time and duration on each page is stored in a SQLite db. How it's used for the current statistics stuff is all in https://github.com/koreader/koreader/blob/master/plugins/statistics.koplugin/main.lua (which is huge, and the SQLite API is a bit tedious).

For the calendar view, I only needed to add that to get some data I could use to build the UI stuff:
https://github.com/koreader/koreader/blob/58a6b01d039d3c383ea2f2a0eda76cec85eb30ac/plugins/statistics.koplugin/main.lua#L2001-L2063

I guess with some thinking, a bit of SQL to get contiguous periods, and/or some Lua to aggragate that, you could get a list of `{ book, starttime, duration_without_pause_for_multiple_pages, nb_of_pages_read_without_pause }`, that you could then display in some UI (using as a start KeyValuePage, as most of this plugin uses).

PR #6761, opened 2020-10-06
Statistics: Tone down the insanity going on in the PageUpdate handler

NiLuJe:
Basically invert the tracking logic: index on pages, instead of timestamps.

This simplifies the logic a great deal, at the cost of losing a tiny bit of granularity in the timestamp tracking in some edge cases (mainly as interpreted by CalendarView), so we force flushes to DB every hour to hide the implementation details away without much effect on the visible results ;).

Also, reset in-memory stats if a font size change has significantly changed the amount of total pages (> 10%).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6761)

### DIFF
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -209,8 +209,8 @@ function ReaderFont:onSetFontSize(new_size)
         timeout = 2,
     })
 
-    -- Let it propagate, to allow stuff to react to font size changes!
-    --return true
+    --- @note: Shouldn't we let this propagate, to allow stuff to react to font size changes?
+    return true
 end
 
 function ReaderFont:onSetLineSpace(space)
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -209,7 +209,8 @@ function ReaderFont:onSetFontSize(new_size)
         timeout = 2,
     })
 
-    return true
+    -- Let it propagate, to allow stuff to react to font size changes!
+    --return true
 end
 
 function ReaderFont:onSetLineSpace(space)
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -188,17 +188,20 @@ function ReaderStatistics:initData()
     end
 end
 
--- Reset the stats on significant page count changes after a font size update (> 10%)
+-- Reset the (volatile) stats on significant page count changes after a font size update (> 10%)
 function ReaderStatistics:onFontSizeUpdate()
     local new_pagecount = self.view.document:getPageCount()
 
     local page_diff = math.abs(new_pagecount - self.data.pages)
     if page_diff >= math.floor(PAGECOUNT_DIFF_THRESHOLD * self.data.pages) then
-        -- Clear DB & volatile stats for current book
-        self:deleteBook(self.id_curr_book)
+        -- Clear volatile stats for current book
         self:resetVolatileStats()
+        -- NOTE: If we were to clear the DB stats, too:
+        --[[
+        self:deleteBook(self.id_curr_book)
         -- Re-create empty entry for the book
         self.id_curr_book = self:getIdBookDB()
+        --]]
     end
 
     -- Update our copy of the page count
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -27,6 +27,7 @@ local T = FFIUtil.template
 local statistics_dir = DataStorage:getDataDir() .. "/statistics/"
 local db_location = DataStorage:getSettingsDir() .. "/statistics.sqlite3"
 local PAGE_INSERT = 50
+local PAGECOUNT_DIFF_THRESHOLD = 0.10
 local DEFAULT_MIN_READ_SEC = 5
 local DEFAULT_MAX_READ_SEC = 120
 local DEFAULT_CALENDAR_START_DAY_OF_WEEK = 2 -- Monday
@@ -192,7 +193,7 @@ function ReaderStatistics:onFontSizeUpdate()
     local new_pagecount = self.view.document:getPageCount()
 
     local page_diff = math.abs(new_pagecount - self.data.pages)
-    if page_diff >= math.floor(0.1 * self.data.pages) then
+    if page_diff >= math.floor(PAGECOUNT_DIFF_THRESHOLD * self.data.pages) then
         -- Clear DB & volatile stats for current book
         self:deleteBook(self.id_curr_book)
         self:resetVolatileStats()
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1905,8 +1905,11 @@ function ReaderStatistics:onPageUpdate(pageno)
         return
     end
 
-    -- Compute the difference between now and the previous page's last timestamp
     local curr_duration = data_tuple[2] or 0
+    -- Make sure our tuple is sane
+    data_tuple = { then_ts, curr_duration }
+
+    -- Compute the difference between now and the previous page's last timestamp
     local diff_time = now_ts - then_ts
     if diff_time >= self.page_min_read_sec and diff_time <= self.page_max_read_sec then
         self.mem_read_time = self.mem_read_time + diff_time
@@ -1914,21 +1917,15 @@ function ReaderStatistics:onPageUpdate(pageno)
         if curr_duration == 0 then
             self.mem_read_pages = self.mem_read_pages + 1
         end
-        -- Create a new data tuple
-        local data = { now_ts, curr_duration + diff_time }
-        -- Update the current page's data list
-        table.insert(page_data, data)
-        self.page_stat[self.curr_page] = page_data
+        -- Update the tuple with the computed duration
+        data_tuple[2] = curr_duration + diff_time
     elseif diff_time > self.page_max_read_sec then
         self.mem_read_time = self.mem_read_time + self.page_max_read_sec
         if curr_duration == 0 then
             self.mem_read_pages = self.mem_read_pages + 1
         end
-        -- Create a new data tuple
-        local data = { now_ts, curr_duration + self.page_max_read_sec }
-        -- Update the current page's data list
-        table.insert(page_data, data)
-        self.page_stat[self.curr_page] = page_data
+        -- Update the tuple with the computed duration
+        data_tuple[2] = curr_duration + self.page_max_read_sec
     end
 
     -- We want a flush to db every 50 page turns
@@ -1944,6 +1941,10 @@ function ReaderStatistics:onPageUpdate(pageno)
 
     -- We're done, update the current page tracker
     self.curr_page = pageno
+    -- And, in the new page's list, append a new tuple with the current timestamp (duration will be computed next pageturn)
+    local new_page_data = self.page_stat[pageno] or {}
+    table.insert(new_page_data, { now_ts, 0 })
+    self.page_stat[pageno] = new_page_data
 end
 
 -- For backward compatibility
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -217,7 +217,6 @@ function ReaderStatistics:resetVolatileStats()
 
     -- Volatile storage pending flush to db
     self.page_stat = {}
-    logger.info("ReaderStatistics:resetVolatileStats")
 end
 
 function ReaderStatistics:getStatsBookStatus(id_curr_book, stat_enable)
@@ -587,12 +586,10 @@ function ReaderStatistics:insertDB(id_book)
     conn:exec('BEGIN')
     local stmt = conn:prepare("INSERT OR IGNORE INTO page_stat VALUES(?, ?, ?, ?)")
     for page, data_list in pairs(self.page_stat) do
-        logger.info("ReaderStatistics:insertDB: Iterating on page", page)
         for _, data_tuple in ipairs(data_list) do
             -- See self.page_stat declaration above about the tuple's layout
             local ts = data_tuple[1]
             local duration = data_tuple[2]
-            logger.info("ReaderStatistics:insertDB: opened @", ts, "read in", duration)
             -- Skip placeholder durations
             if duration > 0 then
                 stmt:reset():bind(id_book, page, ts, duration):step()
@@ -634,7 +631,6 @@ function ReaderStatistics:insertDB(id_book)
     self:resetVolatileStats()
     -- Re-seed the volatile stats with minimal data about the current page
     self.page_stat[self.curr_page] = { { now_ts, 0 } }
-    logger.info("ReaderStatistics:insertDB: Inserted a placeholder duration for page", self.curr_page, "@ ts", now_ts)
     conn:close()
 end
 
@@ -692,7 +688,6 @@ function ReaderStatistics:getStatisticEnabledMenuItem()
                 self.start_current_period = TimeVal:now().sec
                 self.curr_page = self.ui:getCurrentPage()
                 self.page_stat[self.curr_page] = { { self.start_current_period, 0 } }
-                logger.info("ReaderStatistics:getStatisticEnabledMenuItem: Inserted a placeholder duration for page", self.curr_page, "@ ts", self.start_current_period)
             end
             self:saveSettings()
             if not self:isDocless() then
@@ -1899,8 +1894,6 @@ function ReaderStatistics:onPageUpdate(pageno)
         return
     end
 
-    logger.info("ReaderStatistics:onPageUpdate: From page", self.curr_page, "to page", pageno)
-
     -- We only care about *actual* page turns ;)
     if self.curr_page == pageno then
         return
@@ -1919,13 +1912,9 @@ function ReaderStatistics:onPageUpdate(pageno)
     if not then_ts then
         logger.dbg("ReaderStatistics: No timestamp for previous page", self.curr_page)
         self.page_stat[pageno] = { { now_ts, 0 } }
-        logger.info("ReaderStatistics:onPageUpdate: Inserted a placeholder duration for page", pageno, "@ ts", now_ts)
         self.curr_page = pageno
         return
     end
-
-    local dump = require("dump")
-    logger.info("self.page_stat[self.curr_page]:", dump(self.page_stat[self.curr_page]))
 
     -- By now, we're sure that we actually have a tuple (and the rest of the code ensures they're sane, i.e., zero-initialized)
     local curr_duration = data_tuple[2]
@@ -1959,13 +1948,6 @@ function ReaderStatistics:onPageUpdate(pageno)
         self.avg_time = (self.total_read_time + self.mem_read_time) / (self.total_read_pages + self.mem_read_pages)
     end
 
-    logger.info("data_tuple:", dump(data_tuple))
-    logger.info("page_data[#page_data]:", dump(page_data[#page_data]))
-    logger.info("page_data:", dump(page_data))
-    logger.info("self.page_stat[self.curr_page]:", dump(self.page_stat[self.curr_page]))
-
-    logger.info("self.page_stat[pageno]:", dump(self.page_stat[pageno]))
-
     -- We're done, update the current page tracker
     self.curr_page = pageno
     -- And, in the new page's list, append a new tuple with the current timestamp and a placeholder duration
@@ -1973,9 +1955,6 @@ function ReaderStatistics:onPageUpdate(pageno)
     local new_page_data = self.page_stat[pageno] or {}
     table.insert(new_page_data, { now_ts, 0 })
     self.page_stat[pageno] = new_page_data
-
-    logger.info("new_page_data:", dump(new_page_data))
-    logger.info("self.page_stat[pageno]:", dump(self.page_stat[pageno]))
 end
 
 -- For backward compatibility
@@ -2035,7 +2014,6 @@ end
 function ReaderStatistics:onResume()
     self.start_current_period = TimeVal:now().sec
     self:resetVolatileStats()
-    logger.info("ReaderStatistics:onResume: Inserted a placeholder duration for page", self.curr_page, "@ ts", self.start_current_period)
     self.page_stat[self.curr_page] = { { self.start_current_period, 0 } }
 end
 
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -43,8 +43,6 @@ local ReaderStatistics = Widget:extend{
     start_current_period = 0,
     curr_page = nil,
     id_curr_book = nil,
-    curr_total_time = 0,
-    curr_total_pages = 0,
     is_enabled = nil,
     convert_to_db = nil, -- true when migration to DB has been done
     pageturn_count = 0
@@ -180,8 +178,6 @@ function ReaderStatistics:initData()
     -- Update these numbers to what's actually stored in the settings
     -- (not that "notes" is invalid and does not represent edited highlights)
     self.data.highlights, self.data.notes = self.ui.bookmark:getNumberOfHighlightsAndNotes()
-    self.curr_total_time = 0
-    self.curr_total_pages = 0
     self.id_curr_book = self:getIdBookDB()
     self.total_read_pages, self.total_read_time = self:getPageTimeTotalStats(self.id_curr_book)
     if self.total_read_pages > 0 then
@@ -539,7 +535,7 @@ function ReaderStatistics:getIdBookDB()
         stmt = conn:prepare("INSERT INTO book VALUES(NULL, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)")
         stmt:reset():bind(self.data.title, self.data.authors, self.data.notes,
             TimeVal:now().sec, self.data.highlights, self.data.pages,
-            self.data.series, self.data.language, self.data.md5, self.curr_total_time, self.curr_total_pages):step()
+            self.data.series, self.data.language, self.data.md5, 0, 0):step()
         sql_stmt = [[
             SELECT last_insert_rowid() AS num;
         ]]
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -582,15 +582,19 @@ function ReaderStatistics:getIdBookDB()
 end
 
 function ReaderStatistics:insertDB(id_book)
-    if id_book == nil or #self.pages_stat_ts < 2 then
+    logger.info("ReaderStatistics:insertDB", id_book)
+    if id_book == nil or util.tableSize(self.pages_stat_ts) < 2 then
         return
     end
     local now_ts = TimeVal:now().sec
+    logger.info("now", now_ts)
     local conn = SQ3.open(db_location)
     conn:exec('BEGIN')
     local stmt = conn:prepare("INSERT OR IGNORE INTO page_stat VALUES(?, ?, ?, ?)")
-    for page, ts in ipairs(self.pages_stat_ts) do
-        local duration = self.pages_stat_duration[k]
+    for page, ts in pairs(self.pages_stat_ts) do
+        logger.info("Iterating on page", page, "last opened @", ts)
+        local duration = self.pages_stat_duration[page]
+        logger.info("Duration is", duration)
         if duration and duration > 0 then
             stmt:reset():bind(id_book, page, ts, duration):step()
         end
@@ -1075,6 +1079,7 @@ function ReaderStatistics:getCurrentBookStats()
 end
 
 function ReaderStatistics:getCurrentStat(id_book)
+    logger.info("ReaderStatistics:getCurrentStat", id_book)
     if id_book == nil then
         return
     end
@@ -1939,6 +1944,7 @@ function ReaderStatistics:onPageUpdate(pageno)
 
     -- Update the total read duration for the *current* page for this session
     self.pages_stat_duration[self.curr_page] = duration
+    logger.info("Set read duration to", duration, "for page", self.curr_page)
 
     -- See if we'll want to flush volatile stats to the DB
     local flush_stats = false
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -27,6 +27,7 @@ local T = FFIUtil.template
 local statistics_dir = DataStorage:getDataDir() .. "/statistics/"
 local db_location = DataStorage:getSettingsDir() .. "/statistics.sqlite3"
 local MAX_PAGETURNS_BEFORE_FLUSH = 50
+local MIN_PAGES_READ_TO_FLUSH = 2
 local DEFAULT_MIN_READ_SEC = 5
 local DEFAULT_MAX_READ_SEC = 120
 local DEFAULT_CALENDAR_START_DAY_OF_WEEK = 2 -- Monday
@@ -193,7 +194,7 @@ function ReaderStatistics:onUpdateToc()
 
     if new_pagecount ~= self.data.pages then
         -- NOTE: insertDB does the exact same check
-        if not self.id_curr_book or self.mem_read_pages < 2 then
+        if not self.id_curr_book or self.mem_read_pages < MIN_PAGES_READ_TO_FLUSH then
             logger.info("ReaderStatistics: Pagecount change, clearing volatile book statistics")
             -- Clear volatile stats
             self:resetVolatileStats()
@@ -579,7 +580,7 @@ function ReaderStatistics:getIdBookDB()
 end
 
 function ReaderStatistics:insertDB(id_book)
-    if id_book == nil or self.mem_read_pages < 2 then
+    if id_book == nil or self.mem_read_pages < MIN_PAGES_READ_TO_FLUSH then
         return
     end
     local now_ts = TimeVal:now().sec
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1915,17 +1915,19 @@ function ReaderStatistics:onPageUpdate(pageno)
     if diff_time >= self.page_min_read_sec and diff_time <= self.page_max_read_sec then
         self.mem_read_time = self.mem_read_time + diff_time
         duration = duration + diff_time
+        -- If the page hadn't been read already, count it as read
+        if not self.pages_stat_duration[self.curr_page] then
+            self.mem_read_pages = self.mem_read_pages + 1
+        end
         -- Update the reading duration for the current page for this session
         self.pages_stat_duration[self.curr_page] = duration
     elseif diff_time > self.page_max_read_sec then
         self.mem_read_time = self.mem_read_time + self.page_max_read_sec
         duration = duration + self.page_max_read_sec
+        if not self.pages_stat_duration[self.curr_page] then
+            self.mem_read_pages = self.mem_read_pages + 1
+        end
         self.pages_stat_duration[self.curr_page] = duration
-    end
-
-    -- If the page hadn't been read already, count it as read
-    if duration > 0 then
-        self.mem_read_pages = self.mem_read_pages + 1
     end
 
     -- See if we'll want to flush volatile stats to the DB...
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1940,7 +1940,7 @@ function ReaderStatistics:onPageUpdate(pageno)
         duration = duration + self.page_max_read_sec
         self.read_pages_set[pageno] = true
     end
-    self.mem_read_pages = #self.read_pages_set
+    self.mem_read_pages = util.tableSize(self.read_pages_set)
 
     -- Update the total read duration for the *current* page for this session
     self.pages_stat_duration[self.curr_page] = duration
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -192,7 +192,12 @@ function ReaderStatistics:onUpdateToc()
     local new_pagecount = self.view.document:getPageCount()
 
     if new_pagecount ~= self.data.pages then
-        if self.id_curr_book then
+        -- NOTE: insertDB does the exact same check
+        if not self.id_curr_book or self.mem_read_pages < 2 then
+            logger.info("ReaderStatistics: Pagecount change, clearing volatile book statistics")
+            -- Clear volatile stats
+            self:resetVolatileStats()
+        else
             logger.info("ReaderStatistics: Pagecount change, flushing volatile book statistics")
             -- Flush volatile stats to DB for current book
             self:insertDB(self.id_curr_book)
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -197,7 +197,7 @@ function ReaderStatistics:onUpdateToc()
         if not self.id_curr_book or self.mem_read_pages < MIN_PAGES_READ_TO_FLUSH then
             logger.info("ReaderStatistics: Pagecount change, clearing volatile book statistics")
             -- Clear volatile stats
-            self:resetVolatileStats()
+            self:resetVolatileStats(TimeVal:now().sec)
         else
             logger.info("ReaderStatistics: Pagecount change, flushing volatile book statistics")
             -- Flush volatile stats to DB for current book
@@ -211,7 +211,7 @@ function ReaderStatistics:onUpdateToc()
     self.data.pages = new_pagecount
 end
 
-function ReaderStatistics:resetVolatileStats()
+function ReaderStatistics:resetVolatileStats(now_ts)
     -- Computed by onPageUpdate
     self.pageturn_count = 0
     self.mem_read_time = 0
@@ -219,6 +219,12 @@ function ReaderStatistics:resetVolatileStats()
 
     -- Volatile storage pending flush to db
     self.page_stat = {}
+
+    -- Re-seed the volatile stats with minimal data about the current page.
+    -- If a timestamp is passed, it's the caller's responsibility to ensure that self.curr_page is accurate.
+    if now_ts then
+        self.page_stat[self.curr_page] = { { now_ts, 0 } }
+    end
 end
 
 function ReaderStatistics:getStatsBookStatus(id_curr_book, stat_enable)
@@ -630,9 +636,7 @@ function ReaderStatistics:insertDB(id_book)
         self.total_read_time = 0
     end
 
-    self:resetVolatileStats()
-    -- Re-seed the volatile stats with minimal data about the current page
-    self.page_stat[self.curr_page] = { { now_ts, 0 } }
+    self:resetVolatileStats(now_ts)
     conn:close()
 end
 
@@ -686,10 +690,9 @@ function ReaderStatistics:getStatisticEnabledMenuItem()
             -- if was disabled have to get data from db
             if self.is_enabled and not self:isDocless() then
                 self:initData()
-                self:resetVolatileStats()
                 self.start_current_period = TimeVal:now().sec
                 self.curr_page = self.ui:getCurrentPage()
-                self.page_stat[self.curr_page] = { { self.start_current_period, 0 } }
+                self:resetVolatileStats(self.start_current_period)
             end
             self:saveSettings()
             if not self:isDocless() then
@@ -2019,8 +2022,7 @@ end
 -- screensaver off
 function ReaderStatistics:onResume()
     self.start_current_period = TimeVal:now().sec
-    self:resetVolatileStats()
-    self.page_stat[self.curr_page] = { { self.start_current_period, 0 } }
+    self:resetVolatileStats(self.start_current_period)
 end
 
 function ReaderStatistics:saveSettings()
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -580,7 +580,7 @@ function ReaderStatistics:getIdBookDB()
 end
 
 function ReaderStatistics:insertDB(id_book)
-    if id_book == nil or util.tableSize(self.pages_stat_ts) < 2 then
+    if id_book == nil or self.mem_read_pages < 2 then
         return
     end
     local now_ts = TimeVal:now().sec
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -582,19 +582,15 @@ function ReaderStatistics:getIdBookDB()
 end
 
 function ReaderStatistics:insertDB(id_book)
-    logger.info("ReaderStatistics:insertDB", id_book)
     if id_book == nil or util.tableSize(self.pages_stat_ts) < 2 then
         return
     end
     local now_ts = TimeVal:now().sec
-    logger.info("now", now_ts)
     local conn = SQ3.open(db_location)
     conn:exec('BEGIN')
     local stmt = conn:prepare("INSERT OR IGNORE INTO page_stat VALUES(?, ?, ?, ?)")
     for page, ts in pairs(self.pages_stat_ts) do
-        logger.info("Iterating on page", page, "last opened @", ts)
         local duration = self.pages_stat_duration[page]
-        logger.info("Duration is", duration)
         if duration and duration > 0 then
             stmt:reset():bind(id_book, page, ts, duration):step()
         end
@@ -1079,7 +1075,6 @@ function ReaderStatistics:getCurrentBookStats()
 end
 
 function ReaderStatistics:getCurrentStat(id_book)
-    logger.info("ReaderStatistics:getCurrentStat", id_book)
     if id_book == nil then
         return
     end
@@ -1907,7 +1902,7 @@ function ReaderStatistics:onPageUpdate(pageno)
 
     -- First page update of the session, just update the current page's timestamp
     if not self.curr_page then
-        logger.dbg("No previous page")
+        logger.dbg("First PageUpdate of the statistics session")
         self.pages_stat_ts[pageno] = now_ts
         self.curr_page = pageno
         self.pageturn_ts = now_ts
@@ -1944,7 +1939,6 @@ function ReaderStatistics:onPageUpdate(pageno)
 
     -- Update the total read duration for the *current* page for this session
     self.pages_stat_duration[self.curr_page] = duration
-    logger.info("Set read duration to", duration, "for page", self.curr_page)
 
     -- See if we'll want to flush volatile stats to the DB
     local flush_stats = false
@@ -1952,7 +1946,8 @@ function ReaderStatistics:onPageUpdate(pageno)
     if self.pageturn_count >= PAGE_INSERT then
         flush_stats = true
     end
-    -- We also want a flush to DB on the hour, to allow CalendarView from accurately tracking per-hour stats in the DB
+    -- We also want a flush to DB on the hour, to allow CalendarView to accurately track per-hour stats in the DB,
+    -- since we only keep track of a single timestamp per page in memory, unlike in the DB.
     if os.date("%H", now_ts) ~= os.date("%H", self.pageturn_ts) then
         flush_stats = true
     end
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1901,8 +1901,7 @@ function ReaderStatistics:onPageUpdate(pageno)
     -- If we don't have a previous timestamp to compare to, abort early
     if not then_ts then
         logger.dbg("ReaderStatistics: No timestamp for previous page", self.curr_page)
-        table.insert(curr_page_ts_list, now_ts)
-        self.pages_stat_ts[pageno] = curr_page_ts_list
+        self.pages_stat_ts[pageno] = { now_ts }
         self.curr_page = pageno
         return
     end
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -201,6 +201,8 @@ function ReaderStatistics:onUpdateToc()
         -- Re-create empty entry for the book
         self.id_curr_book = self:getIdBookDB()
         --]]
+        --- @fixme: This does mean that you may end up with conflicting page numbers in the database...
+        ---         On the other hand, wiping the book off the db on every minor pagination change feels a bit harsh...
     end
 
     -- Update our copy of the page count
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -27,7 +27,7 @@ local T = FFIUtil.template
 local statistics_dir = DataStorage:getDataDir() .. "/statistics/"
 local db_location = DataStorage:getSettingsDir() .. "/statistics.sqlite3"
 local PAGE_INSERT = 50
-local PAGECOUNT_DIFF_THRESHOLD = 0.10
+local PAGECOUNT_DIFF_THRESHOLD = 0.075
 local DEFAULT_MIN_READ_SEC = 5
 local DEFAULT_MAX_READ_SEC = 120
 local DEFAULT_CALENDAR_START_DAY_OF_WEEK = 2 -- Monday
@@ -189,7 +189,7 @@ function ReaderStatistics:initData()
     end
 end
 
--- Reset the (volatile) stats on significant page count changes after a font size update (> 10%)
+-- Reset the (volatile) stats on significant page count changes after a font size update
 function ReaderStatistics:onSetFontSize()
     local new_pagecount = self.view.document:getPageCount()
 
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -212,6 +212,7 @@ end
 function ReaderStatistics:resetVolatileStats()
     -- Computed by onPageUpdate
     self.pageturn_count = 0
+    self.pageturn_ts = 0
     self.mem_read_time = 0
     self.read_pages_set = {}
     self.mem_read_pages = 0
@@ -628,6 +629,7 @@ function ReaderStatistics:insertDB(id_book)
     self:resetVolatileStats()
     -- last page must be added once more
     self.pages_stat_ts[self.curr_page] = now_ts
+    self.pageturn_ts = now_ts
     conn:close()
 end
 
@@ -685,6 +687,7 @@ function ReaderStatistics:getStatisticEnabledMenuItem()
                 self.start_current_period = TimeVal:now().sec
                 self.curr_page = self.ui:getCurrentPage()
                 self.pages_stat_ts[self.curr_page] = self.start_current_period
+                self.pageturn_ts = self.start_current_period
             end
             self:saveSettings()
             if not self:isDocless() then
@@ -2020,6 +2023,7 @@ function ReaderStatistics:onResume()
     self.start_current_period = TimeVal:now().sec
     self:resetVolatileStats()
     self.pages_stat_ts[self.self.curr_page] = self.start_current_period
+    self.pageturn_ts = self.start_current_period
 end
 
 function ReaderStatistics:saveSettings()
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1918,19 +1918,20 @@ function ReaderStatistics:onPageUpdate(pageno)
 
     -- By now, we're sure that we actually have a tuple (and the rest of the code ensures they're sane, i.e., zero-initialized)
     local curr_duration = data_tuple[2]
+    -- NOTE: If all goes well, given the earlier curr_page != pageno check, curr_duration should always be 0 here.
     -- Compute the difference between now and the previous page's last timestamp
     local diff_time = now_ts - then_ts
     if diff_time >= self.page_min_read_sec and diff_time <= self.page_max_read_sec then
         self.mem_read_time = self.mem_read_time + diff_time
-        -- If the page hadn't been read already, count it as read
-        if curr_duration == 0 then
+        -- If it's the first time we're computing a duration for this page, count it as read
+        if #page_data == 1 and curr_duration == 0 then
             self.mem_read_pages = self.mem_read_pages + 1
         end
         -- Update the tuple with the computed duration
         data_tuple[2] = curr_duration + diff_time
     elseif diff_time > self.page_max_read_sec then
         self.mem_read_time = self.mem_read_time + self.page_max_read_sec
-        if curr_duration == 0 then
+        if #page_data == 1 and curr_duration == 0 then
             self.mem_read_pages = self.mem_read_pages + 1
         end
         -- Update the tuple with the computed duration
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -42,7 +42,7 @@ local ReaderStatistics = Widget:extend{
     calendar_show_histogram = true,
     calendar_browse_future_months = false,
     start_current_period = 0,
-    curr_page = nil,
+    curr_page = 0,
     id_curr_book = nil,
     is_enabled = nil,
     convert_to_db = nil, -- true when migration to DB has been done
@@ -1885,7 +1885,7 @@ end
 
 
 function ReaderStatistics:onPosUpdate(pos, pageno)
-    if not self.curr_page or self.curr_page ~= pageno then
+    if self.curr_page ~= pageno then
         self:onPageUpdate(pageno)
     end
 end
@@ -1900,15 +1900,6 @@ function ReaderStatistics:onPageUpdate(pageno)
 
     local now_ts = TimeVal:now().sec
 
-    -- First page update of the session, just update the current page's timestamp
-    if not self.curr_page then
-        logger.dbg("First PageUpdate of the statistics session")
-        self.pages_stat_ts[pageno] = now_ts
-        self.curr_page = pageno
-        self.pageturn_ts = now_ts
-        return
-    end
-
     -- Compute the difference between the previous page's timestamp (if there is one)
     local then_ts
     if self.pages_stat_ts[self.curr_page] then
@@ -1917,7 +1908,7 @@ function ReaderStatistics:onPageUpdate(pageno)
 
     -- If we don't have a previous timestamp to compare to, abort early
     if not then_ts then
-        logger.dbg("No timestamp for previous page", self.curr_page)
+        logger.dbg("ReaderStatistics: No timestamp for previous page", self.curr_page)
         self.pages_stat_ts[pageno] = now_ts
         self.curr_page = pageno
         self.pageturn_ts = now_ts
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -187,6 +187,23 @@ function ReaderStatistics:initData()
     end
 end
 
+-- Reset the stats on significant page count changes after a font size update (> 10%)
+function ReaderStatistics:onFontSizeUpdate()
+    local new_pagecount = self.view.document:getPageCount()
+
+    local page_diff = math.abs(new_pagecount - self.data.pages)
+    if page_diff >= math.floor(0.1 * self.data.pages) then
+        -- Clear DB & volatile stats for current book
+        self:deleteBook(self.id_curr_book)
+        self:resetVolatileStats()
+        -- Re-create empty entry for the book
+        self.id_curr_book = self:getIdBookDB()
+    end
+
+    -- Update our copy of the page count
+    self.data.pages = new_pagecount
+end
+
 function ReaderStatistics:resetVolatileStats()
     -- Computed by onPageUpdate
     self.pageturn_count = 0
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -26,7 +26,7 @@ local T = FFIUtil.template
 
 local statistics_dir = DataStorage:getDataDir() .. "/statistics/"
 local db_location = DataStorage:getSettingsDir() .. "/statistics.sqlite3"
-local PAGE_INSERT = 50
+local MAX_PAGETURNS_BEFORE_FLUSH = 50
 local DEFAULT_MIN_READ_SEC = 5
 local DEFAULT_MAX_READ_SEC = 120
 local DEFAULT_CALENDAR_START_DAY_OF_WEEK = 2 -- Monday
@@ -1939,7 +1939,7 @@ function ReaderStatistics:onPageUpdate(pageno)
     end
 
     -- We want a flush to db every 50 page turns
-    if self.pageturn_count >= PAGE_INSERT then
+    if self.pageturn_count >= MAX_PAGETURNS_BEFORE_FLUSH then
         self:insertDB(self.id_curr_book)
         -- insertDB will call resetVolatileStats for us ;)
     end
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -47,7 +47,6 @@ local ReaderStatistics = Widget:extend{
     is_enabled = nil,
     convert_to_db = nil, -- true when migration to DB has been done
     pageturn_count = 0,
-    pageturn_ts = 0,
     mem_read_time = 0,
     mem_read_pages = 0,
     total_read_pages = 0,
@@ -212,7 +211,6 @@ end
 function ReaderStatistics:resetVolatileStats()
     -- Computed by onPageUpdate
     self.pageturn_count = 0
-    self.pageturn_ts = 0
     self.mem_read_time = 0
     self.mem_read_pages = 0
 
@@ -587,10 +585,13 @@ function ReaderStatistics:insertDB(id_book)
     local conn = SQ3.open(db_location)
     conn:exec('BEGIN')
     local stmt = conn:prepare("INSERT OR IGNORE INTO page_stat VALUES(?, ?, ?, ?)")
-    for page, ts in pairs(self.pages_stat_ts) do
-        local duration = self.pages_stat_duration[page]
-        if duration then
-            stmt:reset():bind(id_book, page, ts, duration):step()
+    for page, ts_list in pairs(self.pages_stat_ts) do
+        for k, ts in ipairs(ts_list) do
+            -- Both pages_stat_ts & pages_stat_duration lists have stable, matching indices
+            local duration = self.pages_stat_duration[page][k]
+            if duration then
+                stmt:reset():bind(id_book, page, ts, duration):step()
+            end
         end
     end
     conn:exec('COMMIT')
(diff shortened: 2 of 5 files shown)


NiLuJe:
Should make the tail end of the discussion in #6759 clearer ;).

NiLuJe:
Any idea why `ReaderFont` was eating the `SetFontSize` event?

poire-z:
I still _really_ don't like it :/

If we forget the so-called "insanity" of current code, what was the aimed result for the changes? I still don't get it.

> Also, reset in-memory stats if a font size change has significantly changed the amount of total pages (> 10%).

Why ? (And if I understood why, I'd asked you to check that also if I change margins, interline space... But don't ! :) at least, flush what you have instead of dropping - it's not less valuable that what was saved 15mn before in another session.

> This simplifies the logic a great deal

May be, it's possibly more understandable with the 2 tables. But it worked before with a single set, and stuff could be fixed easily, like with #6759. 

> at the cost of losing a tiny bit of granularity in the timestamp tracking in some edge cases (mainly as interpreted by CalendarView), so we force flushes to DB every hour to hide the implementation details away without much effect on the visible results ;).

Adding edgy code tweaks (and not justified) for edges cases kills the "simplified logic" for me :)
If we had been going with your initial idea (always coalesce, not just the session), CalendarView hourly progress wouldn't have been possible.
So, coalesce per session, and flush at the hour... and save CalendarView, thanks.
If tomorrow, one wants to add a hourly per-minutes graphs (#6155) or do more stats datamining (#6454, #6461), there will be not as much information - for no real reason.

So, I think you should not change what will be saved in the DB.
You could keep these alternative slots for some other computations like you want them if needed.
But not sacrifice what is saved.

(And I'm not a huged fan of statistics - I just care about CalendarView -  but I just don't get the principle of these changes to the DB.)

NiLuJe:
> Why ? (And if I understood why, I'd asked you to check that also if I change margins, interline space... But don't ! :) at least, flush what you have instead of dropping - it's not less valuable that what was saved 15mn before in another session.
> 

That's a valid point. The current approach is a bit of a shortcut. Another approach would be to copy the same handler for anything relevant.

Or, simply move the logic to the PageUpdate handler.

NiLuJe:
As for the rest, that's arguably a design decision. The DB needs the handler to collect data for it ;).

I mean, I can easily collect the two things together, but then you'll possibly get the two subsystems disagreeing with each other.

NiLuJe:
Dinner time, germinating a design tweak that should accommodate both use-cases... ;).

poire-z:
You haven't mention precisely what's the _use-case_ you're trying to solve (footer info?)
(Until you answer that, this PR feels more insane than before: diffs +111/-56 - PAGECOUNT_DIFF_THRESHOLD to decide, 0.10, oh no, maybe 0.075 ? :)  - flush at wall clock bell, _drop_ stuff on _some_ rendering changes...)

> you'll possibly get the two subsystems disagreeing with each other on where/when exactly the time was spent ;).

Quantum mechanics and classical physics disagree on that too :) But we're still alive!

You might do what you want with your volatile stuff - but the classical "real" data that we have should go raw in the db - to be cooked later at any sauce (have a good dinner!).
(We might even be able to resurrect your volatile states from the classical raw data - the inverse is impossible).

> Or, simply move the logic to the PageUpdate handler.

With ReaderRolling, there's a chain of onUpdatePos>updatePos>UpdateToc triggered when a rendering has changed document full-height or nb of pages. So, UpdateToc might be the one you wan't to catch.

NiLuJe:
(Untested for now, dinner time ;p).

NiLuJe:
Okay, this should be in a fairly good state now.

Going with a single nested list may or may not have made the whole thing more readable ^^.

plugins/statistics.koplugin/main.lua
```diff
@@ -175,17 +176,49 @@ function ReaderStatistics:initData()
...
 end
 
+-- Reset the (volatile) stats on page count changes (e.g., after a font size update)
+function ReaderStatistics:onUpdateToc()
+    local new_pagecount = self.view.document:getPageCount()
+
+    if new_pagecount ~= self.data.pages then
+        logger.info("ReaderStatistics: Pagecount change, clearing volatile book statistics")
+        -- Clear volatile stats for current book
+        self:resetVolatileStats()
```
poire-z, line 197: You reset/drop what you have (might be 30 pages and then I change font size) without saving them to DB ?

plugins/statistics.koplugin/main.lua
```diff
@@ -175,17 +176,49 @@ function ReaderStatistics:initData()
...
+        -- NOTE: If we were to clear the DB stats, too:
+        --[[
+        self:deleteBook(self.id_curr_book)
+        -- Re-create empty entry for the book
+        self.id_curr_book = self:getIdBookDB()
+        --]]
+        --- @fixme: This does mean that you may end up with conflicting page numbers in the database...
+        ---         On the other hand, wiping the book off the db on every minor pagination change feels a bit harsh...
```
poire-z, line 205: _a bit_ harsh :)

frontend/apps/reader/modules/readerfont.lua
```diff
@@ -209,6 +209,7 @@ function ReaderFont:onSetFontSize(new_size)
         timeout = 2,
     })
 
+    --- @note: Shouldn't we let this propagate, to allow stuff to react to font size changes?
```
poire-z, line 212: I guess nothing else _should_ really have to care about this module "private" event.

plugins/statistics.koplugin/main.lua
```diff
@@ -544,26 +578,22 @@ function ReaderStatistics:getIdBookDB()
...
-    for i=1, #sorted_performance - 1 do
-        diff_time = sorted_performance[i+1] - sorted_performance[i]
-        if diff_time >= self.page_min_read_sec then
-            stmt:reset():bind(id_book,
-                self.pages_stats[sorted_performance[i]],
-                sorted_performance[i],
-                math.min(diff_time, self.page_max_read_sec)):step()
+    for page, data_list in pairs(self.page_stat) do
+        for _, data_tuple in ipairs(data_list) do
+            -- See self.page_stat declaration above about the tuple's layout
```
poire-z, line 590: I don't know if previously, the entries were inserted (and so, shown when doing a SELECT) in timestamp order - and it shouldn't matter much with SQL (as we should use `order by timestamp` if needed) - but having them inserted by event order would be a small plus (when debugging, I remember I once did look at that).

plugins/statistics.koplugin/main.lua
```diff
@@ -1862,38 +1893,71 @@ function ReaderStatistics:onPageUpdate(pageno)
...
         end
-        mem_read_pages = util.tableSize(read_pages_set)
+        -- Update the tuple with the computed duration
+        data_tuple[2] = curr_duration + self.page_max_read_sec
     end
-    -- every 50 pages we write stats to database
-    if util.tableSize(self.pages_stats) % PAGE_INSERT == 0 then
+
+    -- We want a flush to db every 50 page turns
+    if self.pageturn_count >= PAGE_INSERT then
```
poire-z, line 1942: (might give a more explicite name to that PAGE_INSERT variable, while you're at it :)

poire-z (commented):
Looks _A LOT_ saner to me ! Thank you :)

plugins/statistics.koplugin/main.lua
NiLuJe: I wondered about that, but if I flush to db instead, I flush potentially bogus data (c.f., the fixme below).

I don't think there's a perfect solution here :/.

NiLuJe: (The flush to db will call resetVolatileStats, which is why it's a potentially viable alternative, with a different set of drawbacks).

poire-z: Then, the 50 pages you flushed the last time before these new 30 ones are as bogus - or as valuable - as these 30 ones. No reason to make a distinction.

poire-z: Either you drop them all (and I don't want you to :) - or you/we live with mismatched (but complete/not dropped) set of data like we did - until we find a better idea (like suggested in the other PR, %ooooo of doc height (millionth of doc height) to rescale to page height when displaying stats - but it has some caveats, small pages, etc...)

plugins/statistics.koplugin/main.lua
NiLuJe: It was indeed inserted in *timestamp* order before, because of the mad temporary sorted array.

It's now supposedly completely random (`pairs` \o/), although it *tends* to be sorted in *page* order (pairs appears to be friendly to purely numerical hashes).

plugins/statistics.koplugin/main.lua
poire-z: (For the record, we already display different data, like the nb of pages read I think, when viewing the stats for the currently opened book vs. viewing them for this book later once it is closed - not the same data is available - but everybody and me seem fine with that :) so, I wouldn't bother if the displayed info might have some mismatch - as long as the data in the DB is sane and display glitches can be fixed later.)

plugins/statistics.koplugin/main.lua
NiLuJe: As a small example:

```
INSERT INTO page_stat VALUES(1,8,1602022935,16);
INSERT INTO page_stat VALUES(1,9,1602022925,10);
INSERT INTO page_stat VALUES(1,9,1602022951,9);
INSERT INTO page_stat VALUES(1,8,1602022964,120);
INSERT INTO page_stat VALUES(1,13,1602024202,5);
INSERT INTO page_stat VALUES(1,14,1602024091,68);
INSERT INTO page_stat VALUES(1,15,1602024190,8);
INSERT INTO page_stat VALUES(1,16,1602024185,5);
INSERT INTO page_stat VALUES(1,17,1602024164,14);
INSERT INTO page_stat VALUES(1,17,1602024180,5);
INSERT INTO page_stat VALUES(1,9,1602024279,9);
INSERT INTO page_stat VALUES(1,10,1602024268,11);
INSERT INTO page_stat VALUES(1,10,1602024288,8);
INSERT INTO page_stat VALUES(1,11,1602024256,12);
INSERT INTO page_stat VALUES(1,11,1602024296,10);
INSERT INTO page_stat VALUES(1,12,1602024248,8);
INSERT INTO page_stat VALUES(1,12,1602024306,14);
INSERT INTO page_stat VALUES(1,13,1602024320,8);
```

plugins/statistics.koplugin/main.lua
NiLuJe: As a compromise, in ec14c34, we flush, and in the rare cases `insertDB` wouldn't actually flush (very few pages read), we clear the volatile data.

This should hopefully cover the use-case where you open a new book, realize the font sizes are all out of whack, and tweak that before actually starting reading ;).

marek-g:
Hi! I would like to add my point of view to the discussion from yesterday. This is only my very humble opinion. You don't have to agree :)

**Coalescence**

I think that we should keep as complete information as possible. I'm generally against data merging - the more information, the better. That's why I wrote to @NiLuJe yesterday about keeping a *list* of pairs of (timestamp and duration) as values to keep status quo, so the amount of data stored in the database stays the same.

**Dropping data from the database**

I don't have an opinion what do about it now, because I think the best solution is to fix the resizing bug and don't have the problem at all.

**The resizing bug**

My proposed solution was also aligned with the idea of keeping as much complete information as possible. Replacing page numbers with percentage ranges of the document is a lossless change - it's just more information. We can convert it back to pages again when presenting to the user (with some rounding at the other scale levels). But the information kept is very accurate. If we change the scale and calculate that user has read 2 and a half pages, that's what exactly happened in the new scale ;) How we present it is another thing.

This of course requires database migration (which is possible - replace one column with two and recalculate data).

About partially filled pages - I was thinking to treat them as every other page (as a good enough estimation), but the level of granulance is of course to be decided (if the plugin has access to such details of course).

I wonder how other apps are doing it.

poire-z (approved):
Not tested, but it looks fine and sane to me.
(May be merge after 2020.10, or soon and delay 2020.10 to not the coming up weekend? just to have this tested quite a bit and to be sure we won't mess users' stats - even if it looks like a no-change on what gets into the DB)

Frenzie:
Not this weekend would probably mean a delay until the 24th if not the 31st. I'm moving the week of the 19th so I'll probably be busy packing, also don't know if I'll still have my desktop operational.

tl;dr let's just wait and I'll shoot for a release on Saturday or Sunday. :-)

NiLuJe:
> Replacing page numbers with percentage ranges of the document is a lossless change - it's just more information.

My first concern would be *how* to compute those. The engine(s) don't really give us anything finer grained than pages, right?
(I mean, CRe could probably spit out XPointers, but that's another kettle of fish, and more of an absolute adress than a relative position info).
Which leaves, what, computing it based on pages? So that page 100 in a 200 pages book is range [0.49-0.51]?

marek-g:
> Which leaves, what, computing it based on pages? So that page 100 in a 200 pages book is range [0.5-0.505]?

Yes, that's how I would do it. Maybe we should also look how it is done in other apps.

We need to be careful when calculating read part of the book (equivalent of today's pages_read). For example, having two areas [0.100 - 0.110] and [0.095 - 0.108] - the read part is [0.095 - 0.110] = 0.015. We cannot read more than 100% of the book :) But we should add durations for both areas normally. This is a generalized solution to the bug fixed here.

I know it looks a little complicated, so I don't expect it to be implemented anytime soon. Maybe we could create an experimental branch to make a research. I don't have too much time, but I can help if you want :)

poire-z:
@marek-g : thanks for getting into the discussion.
(I obviously agree on your first 2 points. :)
About the 3rd point, I also agree on the principle - it just feels like a lot of work and changes in the other stats computations (and possible edgy bad info: if you paged thru 3 small pages, stats would tell you you have read only 1 page).
If we end up with that, it would be nice if the conversion from slices ranges to number of pages is done in the SQL, directly by sqlite, so it's a bit abstracted. (May be we can define/reset a user-defined SQL function to do that? :)

> My first concern would be how to compute those. The engine(s) don't really give us anything finer grained than pages, right?

It could:
https://github.com/koreader/koreader/blob/6b90239e7c6bc72c13efd54cbffc14bbbbce2c77/frontend/apps/reader/modules/readerrolling.lua#L462-L470
I think the 2nd part is doable even in page mode to get the doc_y of top of page - and we can get the height of the current page... Looks like we might even get them more cheaply:
https://github.com/koreader/koreader-base/blob/10965387e5993289e0fc573914ca0e476f02cd76/cre.cpp#L890-L906
(dunno if that works also in scroll mode :)

Also, given page splitting, we would mostly always get a slice for the current page not exactly the available page height but a bit lower (because last line of text not fitting being pushed on next page makes for some blank at bottom _not_ part of the slice. So, most pages ranges will be a tad lower than UI requested page height - and document_full_height / requested_page_height will be smaller than the nb of pages.

> Which leaves, what, computing it based on pages? So that page 100 in a 200 pages book is range [0.5-0.505]?

That's another idea. No opinion yet which is best :)

marek-g:
Thanks @poire-z. This is a very welcoming project. I expected to be ignored when I mentioned the resolution bug (because it is not obvious bug with not obvious solution). But the things started to move so fast! That explains why KOReader is my favourite ebook reader. That's thanks to hard work of the team!

> if you paged thru 3 small pages, stats would tell you you have read only 1 page

Hm... When I said "treat them as every other page" I was thinking about counting page as a whole page no matter how long it is. Using the information we already have, that is a page number and total number of pages. That's an easy solution (we don't have to have access to more grained API) and 3 small pages will be counted as expected - 3 pages. That will also relieve us of thinking about repeated lines etc.

I mean we could do it more accurate if needed (personally I don't think we need). Even later. I think most of the changes will be backward compatible **after** the change to the common (resolution independent) units is made.

**I love** the idea with SQL View. This will make the solution compatible with all other tools on the database level (I'm not aware how many of them exists)! The "percentage range" unit can be translated to current schema by knowning only one parameter - the total number of pages. We've already got this number in the `book.pages` column. The view could use it. Maybe this would also limit the number of changes outside the statistics module. But here comes another question to my mind - how the `book.pages` is calculated now if number of pages is resolution dependent?

poire-z:
>  We've already got this number in the book.pages column. The view could use it. Maybe this would also limit the number of changes outside the statistics module. But here comes another question to my mind - how the book.pages is calculated now if number of pages is resolution dependent?

It seems to be updated evertytime data is flushed into the DB:
https://github.com/koreader/koreader/blob/6b90239e7c6bc72c13efd54cbffc14bbbbce2c77/plugins/statistics.koplugin/main.lua#L577-L589
And there are a few (two):
`self.data.pages = self.view.document:getPageCount()`
that happen at book reopening time and when showing some statistics.

I guess the ReaderStatistics:onUpdateToc() added in this PR might/should set it too once it gets the new info (after flushing the current volative data, so it's flushed with the future volatile data).

NiLuJe:
Yeah, I did add the page count refresh in the UpdateToc, yeah ;).

And, yeah, I was thinking of leaving the range computation to SQL, too. Or, if not to SQL itself, to insertDB.

(i.e., add one more field to the collected tuple: max_page. Then three more columns in the SQL db: max_page, range_start, range_end. That way, most volatile stuff still thinks in terms of pages in the current scale; and the DB queries could do fun stuff like "you've read n pages at current settings").

NiLuJe:
@poire-z: What's the relation to the stuff you mentioned vs. `self.ui.document:getCurrentPos()`?

That seems to accurately reports the start_y of the current page? Total is at `self.ui.document.info.doc_height`.

poire-z:
`getPageStartY(pageno)` returns the doc_y for any pageno.
`getPageHeight(pageno)` returns this pageno page height (so, possibly 1/3 of UI height for a small page that have a page-break followed).

`getCurrentPos()` returns its internal current doc_y (so, for scroll mode, and top of current page in page mode).
So, logically, in page mode, `getCurrentPos() == getPageStartY(cur_page_num) `

(Don't forget "scroll mode" in your thinking about stats :)

NiLuJe:
Okay, so, basically, to account for both page/scroll mode, simply diff'ing getCurrentPos does the trick (we're already computing duration on the next page anyway, so, we can defer computing the page height without relying on getpageHeight ;)).

New problem, then: getCurrentPos() isn't implemented in PDFs ^^.

poire-z:
PDFs have fixed page. So no need for that.
But need to account for both kind of page stats :/

NiLuJe:
On the other hand... since we're going by page number changes anyway, it's probably less complicated to lose a tiny bit of granularity and just compute the range based on page numbers only...

poire-z:
> Yeah, I did add the page count refresh in the UpdateToc handler, yeah ;).

Oh, right, didn't see it at the bottom :) 

> And, yeah, I was thinking of leaving the range computation to SQL, too. Or, if not to SQL itself, to insertDB.

The issue is more about "sql select" by other code that do compute stats for display, to get page backs from the ranges stored - than about insertDB, isn't it?

>  I suck at maths, so I'm not up to figuring out how to

So do I, so I'm out of getting deeply involved in that :) 

>  I was thinking of a shitty 1000 element array, with basically one bit set for for every 0.1% read, then you just count the true bits at the end and boom, % read.

But not as a replacement of the current plain (page|range, start_ts, duration), which allows many things a bitmap would not, right ? :) In an additional table, maybe...

> we're already computing duration on the next page anyway, so, we can defer computing the page height without relying on getPageHeight ;)

Except when one go pages backward, or jump to other pages.
But I guess I'm ok with assuming the page height is the view/UI height.

NiLuJe:
> 
> The issue is more about "sql select" by other code that do compute stats for display, to get page backs from the ranges stored - than about insertDB, isn't it?
> 

I'm currently thinking that the ranges wouldn't actually make it to the DB, as we don't actually really care about the *where*, we care about the *how much* was read. The range intermediate format simply helps us computing that "how much" in a layout-agnostic way.

> 
> But not as a replacement of the current plain (page|range, start_ts, duration), which allows many things a bitmap would not, right ? :) In an additional table, maybe...
> 

 Oh, yeah, definitely something new, it'd mainly be useful as extra data to rescale the "read pages" count to the current layout.

> 
> Except when one go pages backward, or jump to other pages.

Oh, good point, the "jump" case would be another thing that just works when using pages but would be heinously tricky to deal with with positions ;p.

poire-z:
So, we would continue storing (`page`, start_ts, duration), with still bogus/incoherent `page` values after size change ?
And your new stuff would be _just_ used to compute % or nb of (current layout) pages read ?
Only these little fields need to be accurate ? You're really sure nothing else really needs to work with currently valid/scaled pages ? (Fearing you might again have a too narrow horizon :) but also, the simpler the changes, the better.)
@marek-g : thoughts ?

Oh, we have a statistics panel with pages read per day. Which would then not be rescaled to current page size (but might be OK: 10 days ago, you read 23 pages at the size they had 10 days ago :)

marek-g:
> I'm currently thinking that the ranges wouldn't actually make it to the DB, as we don't actually really care about the _where_, we care about the _how much_ was read.

Hm... Statistics care about _how much_ but what about other plugins? Calendar? Future plugins? I don't know, but I think that you cannot create compatible SQL View to what we have now without storing every slice in DB.

I can see it as you are storing slices in a new SQL table and replace existing SQL table with SQL View. SQL View translates every slice row to one ore more page rows. Technically maybe you can make sql that joins with results from function that returns a few rows...


> I suck at maths, so I'm not up to figuring out how to merge ranges sanely and smartly,

I think it is easy with a smart approach.

Let's imagine you keep both - slices table and page-based table (the same as it is now) - in memory.

- adding new slice is very easy - just record start and the end of it + duration in slices table
- at the same time add one page to page-based table (note that new slice = exactly one page at the moment of adding!)
- if total number of pages has changed (note that you don't have to watch for all the resize related events, just look if the number of pages has changed), then you need to recalculate whole page-based table, you also need to do it on init (or load it from sql view)

How to recalculate page-base table? By converting every slice to page (please note, that one slice can become more than one page, up to a few pages). But it can never have holes as we are treating every slice individually and do not have to merge it in this approach (we will merge pages instead but this is already implemented, one problem less).

Conversion can also be done by swapping slices to database and reading results from SQL View. That way the conversion logic is only in SQL.

Every step is pretty simple now. And we can hide the slices not only for plugins but also keep the existing statistics working on page level. Pretty neat. That's why thinking before writing is so beneficial.


~~The only flaw I can see is when zooming out (going in less pages in the book direction) we will get many page duplicates But hey, that's also correct. If there is more text on a new page => we spent more time reading it => so we have multiple slices that match with that page => if we sum their times, we've got duration that matches bigger page. Yes, that scenario is also fine :)~~

marek-g:
> @marek-g : thoughts ?

I think as I described in previous post, that we should store slices in the database as we get them and recalculate them to pages with SQL view (consistency).

> we would continue storing (page, start_ts, duration), with still bogus/incoherent page values after size change ?

I'm not sure I'm exactly following, but bogus page numbers would be a problem sooner or later. Even statistic reads previous data from database to initialize. But we don't have to keep bogus pages. Dynamic SQL View will keep them always properly scaled to what we have in `book.pages`.

> Oh, we have a statistics panel with pages read per day.

SQL View will also make page numbers recalculated here, which I think is a good thing.

poire-z:
(~~Of possible interest as an alternative to views: https://www.sqlite.org/gencol.html~~ no access to data from another table :( )

Why did you strike your last paragraph? (it kinda makes sense to me :) like the rest)

marek-g:
> Why did you strike your last paragraph? (it kinda makes sense to me :)

It was like thinknig loud. Makes sense but doesn't add anything to the conversation ;)

> like the rest)

Thanks. All you write also makes sense to me, you are good at paying attention to every detail.

> https://www.sqlite.org/gencol.html

Generated columns cannot be used also because the variable number of rows. But I hope that either SQL is flexible enough to do so or either it supports joining with function returning table.

poire-z:
(It feels this may have been suggested above:)
```diff
        CREATE TABLE IF NOT EXISTS page_stat
            (
                id_book    integer,
                page       integer NOT NULL,
                start_time integer NOT NULL,
                period     integer NOT NULL,
+               total_page_count_at_insert_time integer
            }
```
By adding this new column, and populating it with the current value for the book from book.page, and inserting new values with the current pagecount.
Then, isn't the "scaled" page value just:
`page_stat.page / page_stat.total_page_count_at_insert_time  * book.page` ? (book.page being the current page count)

NiLuJe:
> So, we would continue storing (`page`, start_ts, duration), with still bogus/incoherent `page` values after size change ?
> And your new stuff would be _just_ used to compute % or nb of (current layout) pages read ?
> Only these little fields need to be accurate ? You're really sure nothing else really needs to work with currently valid/scaled pages ? (Fearing you might again have a too narrow horizon :) but also, the simpler the changes, the better.)
> @marek-g : thoughts ?
> 
> Oh, we have a statistics panel with pages read per day. Which would then not be rescaled to current page size (but might be OK: 10 days ago, you read 23 pages at the size they had 10 days ago :)

Yeah, most everything appears to be doing a `sum(period), count(DISTINCT page)` query for those, completely ignoring the `total_read_*` columns in the `book` table... ^^.

(This may or may not require fixing in a few instances, whether using a scaled value makes sense or not).

NiLuJe:
Do note that I'm terrible at maths, SQL, and Lua in general, so, err, I don't quite follow your slices approach.

But what @poire-z just said is basically what I went with, see https://github.com/NiLuJe/koreader/tree/stat-range-exp

marek-g:
Good Morning :)

Ok, I see now. You want to have columns: page & total_page_count. That will work of course, because it is the same information as percentage range. You can easily convert between these two forms. Like between Celsius and Fahrenheit.  They are equivalent. However :)

Pros:

- it is an integer format, it is more accurate (no rounding needed)
- it is easier to embrace by someone new

Cons:

- it is very easy to misinterpret and forget that page 5 in one row may be different than page 5 in another row if that row has different value in total_page_count column! 
  If you do "sum(period), count(DISTINCT page)" on a whole table - the result will likely have no sense and will be wrong if you don't recalcule page numbers to a common denominator first

So, if you want to just keep the table like showed above, I'm not a fan of it, because I think it is a very error prone approach. But go ahead if that's easier for you. Just be warned that every "sum(period), count(DISTINCT page)" will have to be replaced with more advanced logic. And everyone new digging for data in database may fall into that bug.

If you want to keep integer version behind the scenes and converted them to page_stat table with a view, I think it will be fine. If you don't want a view, I would go with different units than pages.

> Then, isn't the "scaled" page value just:
> page_stat.page / page_stat.total_page_count_at_insert_time * book.page ? (book.page being the current page count)

It isn't. That's why it is so dangerous. It is just the number of the first of possible page range only! Imagine, that you have scaled a book from 100 pages to 300 pages. Every recorded page number have to be translated to 3 page numbers, where the first one can be calculated with the equation above and the last one with something similar like (need to be verified):

page_stat.page / page_stat.total_page_count_at_insert_time * (book.page + 1) - 1?

Now imagine that every plugin needs to implement it in his logic. That's why I think that SQL View was a great idea. Underhood it may be page & total pages - no problem at all.

marek-g:
To be clear, I'm not telling you what to do, I'm only sharing my point of view why I can see this approach very error prone. If you can catch all the places that needs to be fixed after the change, I'm ok with it.

> Do note that I'm terrible at maths, SQL especially, and Lua in general, so, err, 

Haha. Don't worry. I'm sure, I am worst at Lua. I have written 4 lines of Lua in my life. And deleted one (if that matters) :) All as a part of a pull request. But after all they were accepted :D

marek-g:
And thanks for that you are working on it, anyway! I really appreciate that :)

marek-g:
And two days ago I thought is a pretty hard problem to solve with a dirty looking code, but now I think working together we are close to a very elegant solution.

marek-g:
About the SQL view I can help with that if you want....

marek-g:
Ok. Here it is - working proposition, guys. I think it is a clean solution.

1. Create `page_stat_internal` table - it is exactly the same table as `page_stat` is now but with one additional column - `total_pages`:

```SQL
CREATE TABLE "page_stat_internal" (
	"id_book"	integer,
	"page"	integer NOT NULL,
	"start_time"	integer NOT NULL,
	"period"	integer NOT NULL,
	"total_pages"	integer NOT NULL,
	FOREIGN KEY("id_book") REFERENCES "book"("id"),
	UNIQUE("page","start_time")
);
```

2. Create `page_stat` view - 100% compatible with the current `page_stat` table - takes data from `page_stat_internal` and scales them to the current `book.pages` (on the occasion I've learned how to use the WITH statement):

```SQL
CREATE VIEW page_stat AS
	WITH cte AS (
		SELECT id_book, page, total_pages, pages, start_time, period,
			((page - 1) * pages) / total_pages + 1 AS first_page,
			max(((page - 1) * pages) / total_pages + 1, (page * pages) / total_pages) AS last_page
		FROM page_stat_internal
		JOIN book on book.id = id_book
		
		UNION ALL
		
		SELECT id_book, page, total_pages, pages, start_time, period,
			first_page + 1, last_page
		FROM cte
		WHERE first_page + 1 <= last_page
	)
	SELECT id_book, first_page as page, start_time, period FROM cte
```
Sample database in the attachment:
[statistics2.sqlite3.zip](https://github.com/koreader/koreader/files/5346515/statistics2.sqlite3.zip)

That's it. You can record your data in `page_stat_internal` table using that additional column and read size-corrected results from `page_stat` view. Minimal changes in the statics plugin code are needed and no changes outside the plugin at all! Nice and simple.

What do you think?

marek-g:
Except that if one page becomes many pages, we should also divide `period` equally between them. It needs to be addressed in the SQL code (as you can see it is too troublesome code to write to put it on plugin writers). But I need a break now ;)

marek-g:
Here:

```SQL
DROP VIEW page_stat
CREATE VIEW page_stat AS
	WITH cte AS (
		SELECT id_book, page, total_pages, pages, start_time, 
			period / (last_page - first_page + 1) as period,
			first_page, last_page
		FROM
		(SELECT id_book, page, total_pages, pages, start_time, period,
			((page - 1) * pages) / total_pages + 1 AS first_page,
			max(((page - 1) * pages) / total_pages + 1, (page * pages) / total_pages) AS last_page
		FROM page_stat_internal
		JOIN book on book.id = id_book)
		
		UNION ALL
		
		SELECT id_book, page, total_pages, pages, start_time, period,
			first_page + 1, last_page
		FROM cte
		WHERE first_page + 1 <= last_page
	)
	SELECT id_book, first_page as page, start_time, period FROM cte
```

But note, that we should change type `period` to decimal. Otherwise, when dividing 7 minutes by 2 = 3 minutes, we are loosing 1 minute.

poire-z:
> You want to have columns: page & total_page_count. That will work of course, because it is the same information as percentage range. 

That's just what I meant (no need for slices start & end, page + page_count are enough to infere the others, and it's easy to migrate past values by just filling the new column with an info we have).

> But go ahead if that's easier for you

Not my war :) but yes, having that simple thing as an "internal" representation (that we have, and can migrate easily) and a view to make it look like before feels like the perfect solution.

> Here it is - working proposition, guys. I think it is a clean solution.
> Minimal changes in the statics plugin code are needed and no changes outside the plugin at all! Nice and simple.
Very little code to be changed. No more places to introduce errors. No one who reads data from the view doesn't need to think about calculations.

Same feeling.

(You'll have to do a bit of benchmarking just so we get an idea of the overhead involved - even if it wouldn't matter much as it's stats that are going to be displayed and we won't mind much if it takes a bit more time - but I'd like to know.)

No time right now to dig into the SQL of your view, but does this (possibly recursive?) usage of the `cte` table from inside what builds it work ?

`WITH cte AS ( [...] SELECT ... from cte ...)`

And if it works, why does it work, what's happening ? :)

poire-z:
@NiLuJe : not fond of the permilles 1000 bits bitmap. It's arbitrary limiting granularity. WIth a 3000 pages book, we'll get read pages a multiple of 3 - which feels buggy looking.

marek-g:
> Not my war :) but yes, having that simple thing as an "internal" representation (that we have, and can migrate easily) and a view to make it look like before feels like the perfect solution.

I'm glad to hear it. I think the same.

> No time right now to dig into the SQL of your view, but does this (possibly recursive?) usage of the cte table from inside what builds it work ?
> 
> WITH cte AS ( [...] SELECT ... from cte ...)
> 
> And if it works, why does it work, what's happening ? :)

The sub-select in parentheses calculates `first_page` and `last_page`. These are the results of scaling the `page` to new number of book pages. `last_page` can be the same or larger than `first_page`. If it is larger, we need to generate additional rows. The recursive table does exactly that.

Because it is all a single SQL statement (not using cursor here), I _hope_ (I haven't check) that SQL is able to use `book_id` index on the SQL View. If it is, that's all what we need to efficiently get all the data for a given book (I think this is our use case).

poire-z:
Just playing with your suggestion, on a manually updated copy of my emulator statistics.sqlite3:

```
sqlite> .timer on
sqlite> select count(1) from page_stat_internal where id_book=1453 order by start_time asc;
count(1)
----------
116
Run Time: real 0.000 user 0.000000 sys 0.000431

sqlite> select * from page_stat_internal where id_book=1453 order by start_time asc;
[...]
1453        15          1602102642  30          49
1453        15          1602102674  19          49
1453        15          1602102695  17          49
1453        15          1602102738  18          49
1453        15          1602102788  37          49
1453        15          1602102839  21          49
1453        15          1602102864  20          49
1453        15          1602102887  180         49
1453        15          1602103910  180         49
1453        15          1602106148  11          49
1453        15          1602106162  12          49
1453        16          1602155597  12          72 [ change of font size]
1453        24          1602155614  12          72
1453        25          1602155626  16          72
1453        26          1602155642  75          72
1453        27          1602155717  21          72
1453        28          1602155825  9           72
Run Time: real 0.002 user 0.000000 sys 0.001950

sqlite> select count(1) from page_stat where id_book=1453 order by start_time asc;
count(1)
----------
201
Run Time: real 0.385 user 0.384820 sys 0.000000

sqlite> select * from page_stat where id_book=1453 order by start_time asc;
1453        21          1602102642  15
1453        22          1602102642  15
1453        21          1602102674  9
1453        22          1602102674  9
1453        21          1602102695  8
1453        22          1602102695  8
1453        21          1602102738  9
1453        22          1602102738  9
1453        21          1602102788  18
1453        22          1602102788  18
1453        21          1602102839  10
1453        22          1602102839  10
1453        21          1602102864  10
1453        22          1602102864  10
1453        21          1602102887  90
1453        22          1602102887  90
1453        21          1602103910  90
1453        22          1602103910  90
1453        21          1602106148  5
1453        22          1602106148  5
1453        21          1602106162  6
1453        22          1602106162  6
1453        16          1602155597  12
1453        24          1602155614  12
1453        25          1602155626  16
1453        26          1602155642  75
1453        27          1602155717  21
1453        28          1602155825  9
Run Time: real 0.386 user 0.371557 sys 0.014196
```

The results look coherent (page 15 read 180 s becomes page 21 read 90s + page 22 read 90s).
But the additional time it takes (386ms vs 2ms) is quite noticable.

> that's all what we need to efficiently get all the data for a given book (I think this is our use case).

There are statitics panels that show the nb of pages and reading duration per period with agregating the info about all the books read during that period :/

We have the tools to try finding a faster query :):
```
sqlite> explain query plan select * from page_stat_internal where id_book=1453 order by start_time asc;
QUERY PLAN
|--SEARCH TABLE page_stat_internal USING INDEX page_stat_id_book (id_book=?)
`--USE TEMP B-TREE FOR ORDER BY

sqlite> explain query plan select * from page_stat where id_book=1453 order by start_time asc;
QUERY PLAN
|--CO-ROUTINE 3
|  |--SETUP
|  |  |--SCAN TABLE page_stat_internal
|  |  `--SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
|  `--RECURSIVE STEP
|     `--SCAN TABLE cte
|--SCAN SUBQUERY 3
`--USE TEMP B-TREE FOR ORDER BY

sqlite> explain select * from page_stat_internal where id_book=1453 order by start_time asc;
addr  opcode         p1    p2    p3    p4             p5  comment
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     27    0                    00  Start at 27
[...]
28    Goto           0     1     0                    00

sqlite> explain select * from page_stat where id_book=1453 order by start_time asc;
addr  opcode         p1    p2    p3    p4             p5  comment
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     117   0                    00  Start at 117
[...]
120   Goto           0     1     0                    00
```

poire-z:
Without the `UNION ALL and what follows`, it's as fast (but bogus of course):
```
1453        21          1602102642  15
1453        21          1602102674  9
1453        21          1602102695  8
1453        21          1602102738  9
1453        21          1602102788  18
1453        21          1602102839  10
1453        21          1602102864  10
1453        21          1602102887  90
1453        21          1602103910  90
1453        21          1602106148  5
1453        21          1602106162  6
1453        16          1602155597  12
1453        24          1602155614  12
1453        25          1602155626  16
1453        26          1602155642  75
1453        27          1602155717  21
1453        28          1602155825  9
Run Time: real 0.002 user 0.000884 sys 0.001327

sqlite> explain query plan select * from page_stat where id_book=1453 order by start_time asc;
QUERY PLAN
|--SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH TABLE page_stat_internal USING INDEX page_stat_id_book (id_book=?)
`--USE TEMP B-TREE FOR ORDER BY

```

marek-g:
Great findings @poire-z! So it is not using indexes at all. Good catch with the query plan!

So we've got 3 table scans. I think this one is to be optimized:
- SCAN TABLE page_stat_internal

The second one: "SCAN TABLE cte" in RECURSIVE STEP is fine - cte table has 1 up to a few rows and it is build in runtime.

What we can do I think is to make a subquery somehow to force sqlite to use the id_book index first to narrow data belonging to one book first and then use recursive steps on smaller subset of data....

marek-g:
This is pretty exciting problem with that resizing bug. If I was a blogger I would write a post about it :D:D

marek-g:
We can try to use function that takes two parameters: ``generate_rows(page_start, page_end)`` that returns temporary table with rows with values from `page_start` to `page_end` and join with it. I hope SQLite supports it. That way we can remove the recursive step.

marek-g:
What about this one?

```SQL
-- this view rescales data from page_stat_internal table to current number of book pages
CREATE VIEW page_stat AS
    
    SELECT id_book, first_page + N - 1 as page, start_time, period / (last_page - first_page + 1) as period FROM (

		SELECT id_book, page, total_pages, pages, start_time, period,
		
				-- first page_number of the page after rescaling single row
				((page - 1) * pages) / total_pages + 1 AS first_page,
				
				-- last page_number of the page after rescaling single row
				max(((page - 1) * pages) / total_pages + 1, (page * pages) / total_pages) AS last_page,

				N

			FROM page_stat_internal
			
			JOIN book on book.id = id_book
			
			-- duplicate rows for multiple pages if needed (as a result of rescaling)
			JOIN (
				-- maximum scale level is 10x
				SELECT 1 AS N UNION ALL
				SELECT 2 AS N UNION ALL
				SELECT 3 AS N UNION ALL
				SELECT 4 AS N UNION ALL
				SELECT 5 AS N UNION ALL
				SELECT 6 AS N UNION ALL
				SELECT 7 AS N UNION ALL
				SELECT 8 AS N UNION ALL
				SELECT 9 AS N UNION ALL
				SELECT 10 AS N
			) ON N <= (last_page - first_page + 1)
			
		)
```

The execution plan:

```
explain query plan select * from page_stat where id_book=2 order by start_time asc;
```

```

MATERIALIZE 4
--
COMPOUND QUERY
LEFT-MOST SUBQUERY
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
UNION ALL
SCAN CONSTANT ROW
SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
SEARCH TABLE page_stat_internal USING INDEX page_stat_id_book (id_book=?)
SCAN SUBQUERY 4
USE TEMP B-TREE FOR ORDER BY
```

It starts by constructing 10-elements table (can be also physical table to be even faster). And after that it goes as smooth as possible (using `page_stat_id_book` index) :D It's also simpler to understand.

NiLuJe:
Okay, that was definitely way over my head ^^.

But, that iteration seems to behave far batter, yeah ;).

```
SQLite version 3.32.3 2020-06-18 14:00:33
Enter ".help" for usage hints.
sqlite> .timer on
sqlite> select count(1) from page_stat_internal where id_book=78 order by start_time asc;
692
Run Time: real 0.004 user 0.000000 sys 0.000000
sqlite> select count(1) from page_stat_v1 where id_book=78 order by start_time asc;
692
Run Time: real 0.026 user 0.020000 sys 0.000000
sqlite> select count(1) from page_stat_v0 where id_book=78 order by start_time asc;
692
Run Time: real 0.218 user 0.210000 sys 0.000000
sqlite> 
```

(That's on an actual H2O, btw).

(Note that there's no actual scaling involved, I just set `total_pages` to the actual pagecount unilaterally).

i.e., starting from my most complete stat db (the one on my Forma)

```
ALTER TABLE page_stat RENAME TO page_stat_internal;
ALTER TABLE page_stat_internal ADD COLUMN total_pages integer NOT NULL DEFAULT 0;
UPDATE page_stat_internal SET total_pages = 607 WHERE id_book = 78;
```

Then each view created via a dedicated .db text file because it was a bit clunky from the shell ;p.

NiLuJe:
The only downside being that the query cache has, obviously, a harder time making subsequent queries any faster, unlike with plain queries.

Which may not matter all that much if we get it fast enough in the first place ;).

NiLuJe:
Given my lack of SQL chops, my original plan was to simply move the computation to insert time, hence the bitfield trickery, and basically just query a computed value instead of doing it at query time.

On that note, 

> @NiLuJe : not fond of the permilles 1000 bits bitmap. It's arbitrary limiting granularity. WIth a 3000 pages book, we'll get read pages a multiple of 3 - which feels buggy looking.

I'm not sure I grok that? Only the ranges themselves are scaled, the scaling factor isn't, so, while there is a bit of rounding error, it's in the margin of +/- 1 if I do a quick bit of napkin maths.

i.e., say, you've read 1658 pages of that 3000 pages book, assuming you've started at the cover, that's range [0 - (1658 / 3000. * 1000)], i.e., [0 - 553] after rounding. So, 553 permilles. That makes the pagecount round trip computation: 553 / (1 / 300. * 1000) = 1659.

NiLuJe:
My only issue is that I have no clue whatsoever what's happening in the query ^^.

Am I right to assume that the `first_page` - `last_page` business is just trying to find ranges of consecutive `page` with a matching `total_pages`?

Next question: What is `pages`? What is `N`?

----

Sidebar: remind me to rename the `period` column to `duration` because that's been driving me nuts.

marek-g:
Thank you @NiLuJe for the timings! The new version of the query is slower than raw table of course but it should be slower by a constant factor now (which is 10x, but we still fit in 20 ms, and that should be independent of number of books, where the first one was getting slower and slower)...

Let me explain how the SQL works... Please ask about any point if you need more details :)

1. Important thing to understand is that one page after resizing can become 1 or more pages (when zooming in). This is the problematic thing that makes this query complex.

2. The most important part are the equations:

- ((page - 1) * pages) / total_pages + 1 AS first_page
- max(((page - 1) * pages) / total_pages + 1, (page * pages) / total_pages) AS last_page

Calculates the new page number range for a given `page` knowing that total number of pages has changed from `total_pages` to `pages`. I had to translate 1-based index to 0-based and vice-versa, hence some "- 1" and "+ 1". The "max" function is the guard that last_page >= first_page. It's needed because I calculate `last_page` as the index of the next page resized - 1. And sometimes this -1 (it's really "+1 - 1"  = 0 so you don't see "-1" easily) is a reason it smaller than `first_page`. It's hard to explain here :) Try to write the equations yourself and see if you get the same. I'm sure you will get to something very similar.

3. If we know `first_page` and `last_page` we also know that we want to generate `last_page` - `first_page` + 1 rows on the output from a single input row.

4. It is hard to generate more rows on the output in SQL. One trick is to join with another table that has desired number of rows. Join is like double for each here and is done to multiply rows (it creates every one with every one pairs).

5. This is a table that has 10 rows (I assume we don't need to zoom in more than 10x):

```SQL
				SELECT 1 AS N UNION ALL
				SELECT 2 AS N UNION ALL
				SELECT 3 AS N UNION ALL
				SELECT 4 AS N UNION ALL
				SELECT 5 AS N UNION ALL
				SELECT 6 AS N UNION ALL
				SELECT 7 AS N UNION ALL
				SELECT 8 AS N UNION ALL
				SELECT 9 AS N UNION ALL
				SELECT 10 AS N
```

Simple right? :)

6. So we can join with it taking from it only the needed number of rows (look at the condition):

```SQL
            JOIN (
				-- maximum scale level is 10x
				SELECT 1 AS N UNION ALL
				SELECT 2 AS N UNION ALL
				SELECT 3 AS N UNION ALL
				SELECT 4 AS N UNION ALL
				SELECT 5 AS N UNION ALL
				SELECT 6 AS N UNION ALL
				SELECT 7 AS N UNION ALL
				SELECT 8 AS N UNION ALL
				SELECT 9 AS N UNION ALL
				SELECT 10 AS N
			) ON N <= (last_page - first_page + 1)
```

7. Now our result set has desired (probably higher) number of rows. The N is just an successive integer taken from joined table.

8. We can use N to calculate new `page` for every new row number which is `first_page` + N - 1.

9. We also need to divide duration between number of generated rows (pages) from a single one. It's the equation `period / (last_page - first_page + 1) as period`

NiLuJe:
2. Oh, okay, I get it. I just don't quite get where `pages` comes from in the query, but at least now I now what it's supposed to represent ^^.

3. Ohkay, what happens to duration though? Do we divide it by the amounts of row created? Okay, yeah, that's 9. ;). Yay me :D

NiLuJe:
Okay, another random question, this time more directly related to #6759...

Shouldn't the average time per page *actually* count *every* page, even re-reads? (i.e., I see it as a kind of "you generally spend n seconds staring at a screenful of words" metric).

Meaning the fix would have to be the other way around: fix (that incarnation of) total_read_pages to count *every* page in page_stat, not every distinct one.

And then update the stats recap where necessary to make that clear (i.e., unique pages read vs. pages read).

poire-z:
(Sorry, been busy) no time yet to reread and understand the changes.
Just giving the timing to compare to previous ones:
```
1453        21          1602102642  15
1453        22          1602102642  15
1453        21          1602102674  9
1453        22          1602102674  9
1453        21          1602102695  8
1453        22          1602102695  8
1453        21          1602102738  9
1453        22          1602102738  9
1453        21          1602102788  18
1453        22          1602102788  18
1453        21          1602102839  10
1453        22          1602102839  10
1453        21          1602102864  10
1453        22          1602102864  10
1453        21          1602102887  90
1453        22          1602102887  90
1453        21          1602103910  90
1453        22          1602103910  90
1453        21          1602106148  5
1453        22          1602106148  5
1453        21          1602106162  6
1453        22          1602106162  6
1453        16          1602155597  12
1453        24          1602155614  12
1453        25          1602155626  16
1453        26          1602155642  75
1453        27          1602155717  21
1453        28          1602155825  9
Run Time: real 0.005 user 0.005023 sys 0.000000
It's a bit variable, some other runs similar, some like:
Run Time: real 0.012 user 0.002632 sys 0.003209
Run Time: real 0.010 user 0.000996 sys 0.004871
Run Time: real 0.006 user 0.002421 sys 0.002908
Run Time: real 0.007 user 0.005219 sys 0.000711
```
So, rather good ! (but a bit ugly with the N and the limitation to x10)

marek-g:
> Oh, okay, I get it now, thanks. I just don't quite get where pages comes from in the query, but at least now I know what it's supposed to represent ^^.

`pages` come from `book.pages` through `JOIN book on book.id = id_book


> shouldn't the average time per page actually count every page, even re-reads?

I think I don't follow. Can you explain more, please? That was the state before.

My use case was. Sometimes (sometimes means really frequently, so frequently that it ruins totally the old stats for me) I'm going back one page. Either to look for some name. Maybe I missed something and I need to reread one sentence. Sometimes I look one page or more forward to look how far is it to the end of chapter. Maybe I'm nervous :) There are short looking times at the pages. They are not re-reads. If we would count them as re-reads, it would became that I've read 300 pages from the book of 100 pages.

Maybe it is a problem of what is reread and what is "just looking at the page but not reading it as a whole"?

I think this is something that is hard to guess. You can try to guess based on the time parameter. We can add such threshold parameter. If you set it to "0" you will have old behavior. I would set it to "100000" for myself. But something like 1-2 minutes should be fine. If a look at the page takes more than the parameter - it's reread. If it is lower, it's "just browsing through pages",

The new behavior seams more sane for my use case. If I reread a page it's maybe there was something difficult. And maybe there will be another "difficult" page later and it should take that into account that I will read the rest of the book slower. So that's I like the change. However, I can see now that someone else may have mutually exclusive needs. In that case maybe parametrize the behavior?

NiLuJe:
>
> `pages` come from `book.pages` through `JOIN book on book.id = id_book
> 

Oh, so that's what `JOIN` is for :D. Thanks ;).

----

See https://github.com/NiLuJe/koreader/blob/9bd7354f929bd834981ef1331a9330e19b52a995/plugins/statistics.koplugin/main.lua#L2054-L2058 for context.

And, okay, I get what you're experiencing. The `page_min_read_sec` setting is what's supposed to help deal with this. I understand that, in your case, it may be too low to account for most of your "peeks". And I understand that getting it "right" for every use case is impossible ;).

My thinking was that, assuming `page_min_read_sec` does its job mostly properly, I actually expect the `avg_time` to be computed based on time spent reading *a* page, not on reading *unique* pages.

(Re: the comment snippet linked above: the potential overlap in using two different sources of "unique" pages, with each of them having a separate idea about what constitutes "unique" also potentially adds some noise to the data. If they were both using "all pages that pass the min/max check", that goes away).

NiLuJe:
The state before was a bit weirder: `total_*` was always counting *distinct* pages, while `mem_*` was counting *all* pages.

The new behavior at least align them to both count *distinct* pages, although what they consider distinct may actually overlap until insertDB clears it up.

I'm just wondering if doing it the other way around (i.e., count *all* pages, both for `total_*` and `mem_*`) makes more sense, assuming the min/max stuff is properly setup.

NiLuJe:
That raises a valid question: does this attempt to correct this, arguably rare, potential overlap, one that will actually be cleared up on its own anyway, ends up being more destructive to the potentially more common use-cases of the min range not being properly setup, like you experienced ;).

NiLuJe:
Because right now, if I understand correctly, assuming you *do* end up with a bit of "re-reads" in the DB, the current `total_*` query, `count(DISTINCT page), sum(duration)` means that every duration gets accounted for, but not every page. So a re-reads means that a single page end up being counted as read for potentially much longer than `page_max_read_sec`.

Which leads me to another idea: clamp duration to `page_max_read_sec` for those re-reads? So that each *unique* page can actually never go above `page_max_read_sec` in the sum.

Not *quite* sure how to do that in the query, though...

NiLuJe:
Third random idea: only keep a single duration for re-reads, the longest one?

marek-g:
> The state before was a bit weirder: total_* was always counting distinct pages, while mem_* was always counting all pages.

Haha. Yes, exactly. It was a mix :) That's true.

> The page_min_read_sec setting is what's supposed to help deal with this

The current way is the best for me and will always work better than `page_min_read_sec`, because I never read a page twice. If I read it twice - that's because I'm slow or easily distracted or short memory (you know, I'm old :P and Sunday reader) and I want it being reflected in the stats of the remaining reading time.

NiLuJe, please do what you think is the best as default - I'm fine with any, but also please leave me on option to set the current behavior. I just don't want to maintain the fork of the app especially for this reason. That was the reason for me to move my ass and look at the code, so I really care ;)))


> Which leads me to another idea: clamp duration to page_max_read_sec for those re-reads? So that each unique page can actually never go above page_max_read_sec in the sum.

Good catch! I think you are right. In my opinion we should definitely respect the `page_max_read_sec` setting.

> Not quite sure how to do that in the query, though...

It's to be done outside the SQL View. It may be a part of SQL that you query that view:

```SQL
SELECT count(*), sum(duration) FROM (
	SELECT page, MIN(SUM(period), 210) as duration FROM page_stat WHERE id_book = 2
	GROUP BY page
)
```

> Third random idea: only keep a single duration for re-reads, the longest one? (that should more-or-less always end up being the actual "read").

Except when you read half a page one day and another half another day, but it would be acceptable in my case....

However, the choice between algorithms would be the best if you want to experiment (I love the open source because it gives so so many configuration options vs propitiatory software). And isn't it changing the current behavior on the database stored data (as you have noticed how `total_*` works).


where replace 210 with the correct max value.

NiLuJe:
Oh, I'm very much thinking out loud here, so, any and all comments are very much appreciated ;).

poire-z:
I agree that the use case of looking back or browsing forward for small duration is to be taken care by the min time reading setting. You probably didn't know about it . I have that at 30, and I know when browsing I should not stay longer than 25s :) even when opening a new book to have a quick look at it... hurry, browse before the 30s countdown :)

I don't have much input to give on the above few posts, as I rarely look at these kind of stats. But it feels what to use/cap/do may depend on the context of the stats.
Also, I think:
- page_stats_internal should keep getting everything raw
- page_stats is just an scale/split of these raw datas, not an aggregation, and should keep everything as raw and equivalent (so, I don't think any cap/sum/mean should be done in this VIEW)
- it's then all the current queries in the plugin that use "page_stats" that, depending on their need (and what we want the result to represent/assume), would do the adequate cap/sum/mean.

Currently, althouhg I don't care much, I think I can make sense of the values shown because I know all durations add up. Having simple and coherent rules, even if not perfect, makes interpreting the results by someone knowing these rules easier. (I think I wrote about that somewhere.)

Also, for me, `page_max_read_sec` is just a safety for when you fell asleep while reading :) 
(Remaining time to read book might probably use all the re-reads duration, and assume that if the user has re-read (or pause or fall sleep) that much up to current page, it will re-read as much the next pages in the book - so, in that case, capping feels wrong - although capping might give an over optimistic ETA :) (Do we want to lie and tell the user is better, faster, stronger that what he is ? :)

marek-g:
> So, rather good ! (but a bit ugly with the N and the limitation to x10)

Other options are:

1. The RANGE() function could be added in C to SQLite like in a SQL Notebook app - https://sqlnotebook.com/range-func.html. Cons: the view would be incorrect outside the KOReader, so it is not an option.

2. Use recursive select (WITH CTE) - in 4 lines we can have parametrized number of rows. Cons: the table needs to be scanned as a whole with every row. The number of rows is our slowdown factor. It is the same as 3. - only the shortcut. We can do it as it is nicer to parametrize.

3. We can manually add some more rows. Cons: the same as above. If we have 20 rows, we will be 2x slower than now.

4. Create new table with rows and join with it. Pros: we can create 1000 rows in primary index. It will be faster than it is now. The limit will be 1000x zoom. I've got a feeling it is not very beauty, too :)

marek-g:
> Do we want to lie and tell the user is better, faster, stronger that what he is ? :)

I don't want that ;) I hate when it says I will read a chapter in 30 minutes and finally it takes an hour ;)


> I have that at 30, and I know when browsing I should not stay longer than 25s :)

I didn't know about this parameter and this is what bring me here, to be honest. So maybe for people like me it should be set to 30 as a default. I really thought that there is something broken and was wondering why is nobody complaining :) I just assumed that people just read the book, not playing like me ;)

Anyway, still - how do you count the number of read pages in the summary? If I read 10 pages twice (or just asleep 10 times during reading, which is the same). And I am currently on the page 50 of 200. Will it tell that I have read 60 pages or 50 pages in the book summary page?

marek-g:
I need to sleep with that idea of using `page_max_read_sec`. With it estimates of my way of reading will be bearable.

However, because I read a page twice only when I'm loosing my concentration, that will be probably happen in the next chapter.

Can we have that as a parameter? If not, I will probably stay with the `page_max_read_sec`.

Almost a week of discussion. Thousands of words. Many SQL queries. 4 lines of code (not including 1 deleted) by me !!!!! Much more by you NiLuJe. What a overlook !!!!! grrrrrrrrrr ;-)))))))

poire-z:
> it should be set to 30 as a default. I really thought that there is something broken and was wondering why is nobody complaining :) I 

From some screenshots I've seen in past issues, there are surprisingly quite a few people that read with a large font size and have pages with 10 lines of each 5 words. So, I guess they spend less than 30 seconds on a page.

> And I am currently on the page 50 of 200. Will it tell that I have read 60 pages or 50 pages in the book summary page?

I think in that context, it should tell you you have read 50 pages. 50 distinct pages.
In other contexts like time per page, or pages read by day, we're not showing either 50 or 60, and it might be different. We might want to use all occurences, coalesce globally or over a day, etc... (dunno really what is used currently, just saying what is used may depend on context :)

> Can we have that as a parameter? 

I didn't really get what behaviour change you'd like. How would you phrase the checkboxed menu item, and what would be it's help_text giving it more details, like in:
https://github.com/koreader/koreader/blob/0ed46809675f50c2d87a1e95d94b7d466f209f6c/plugins/statistics.koplugin/main.lua#L702-L706

NiLuJe:
> 
> It's to be done outside the SQL View. It may be a part of SQL that you query that view:
> 
> ```sql
> SELECT count(*), sum(duration) FROM (
> 	SELECT page, MIN(SUM(period), 210) as duration FROM page_stat WHERE id_book = 2
> 	GROUP BY page
> )
> ```
> 
> replace 210 with the correct max value.

Ah, progress, I was almost there myself, thanks ;).
(As you may have realized by now, I've *really* almost never done SQL, ever ^^).

> 
> > Third random idea: only keep a single duration for re-reads, the longest one? (that should more-or-less always end up being the actual "read").
> 
> Except when you read half a page one day and another half another day, but it would be acceptable in my case....
> 

Hmm, good point ;).

FWIW, since that's what I was playing at, and it's a similar query layout:

```sql
SELECT count(*), sum(durations) FROM ( SELECT max(duration) AS durations FROM page_stat where id_book = 43 GROUP BY page );
```

(NOTE: I've indeed renamed the awfully-named `period` column to `duration` ;)).

NiLuJe:
* I think we all agree that the data mangling needs to be contextual, and vary depending on context ;).

* I think we agree that when the summary says we've read x pages out of the  y total pages of the book, that x is the amount of *distinct* pages is implicit.

* Where things get murkier is where time is involved... I think we agree that here, some form of data wrangling is necessary.
At least as far as computing the average time spent reading a page, I'm on the fence between the simple "count all pages, count all duration" approach, which assume the min/max ranges are working as intended, and what we've proposed above with counting distinct pages, but clamping duration to `page_max_read_sec`, which is more aggressive with the data, but possibly easier on "bad" setups...

NiLuJe:
>     4. Create new table with rows and join with it. Pros: we can create 1000 rows in primary index. It will be faster than it is now. The limit will be 1000x zoom. I've got a feeling it is not very beauty, too :)
> 
> 
> Edit: Hm.. the 4. is the best from a pragmatic look. It's the fastest and practically with no limit.

I would tend to agree. But the SQL neophyte that I am, how would that look, in practice?

NiLuJe:
As a *really* obvious example, I'm using the random book that sits at the top of my FL on my test device: it spends an inordinate amount of time being stuck on page 1 while I'm doing stuff in a shell.

So I have  *a lot* of entries on said first page stuck at `page_max_read_sec`.

In this, admittedly extremely unlikely example, the current approach is hilariously wrong:

```
sqlite> SELECT count(DISTINCT page), sum(duration) FROM page_stat where id_book = 43;
22|19236

=> 874.36363636363637s per page
```

The "count everything" approach yields a *much* lower average, possibly too low, because it exposes the other common on my test book: skipping pages very quickly, so, a lot of pages close to `page_min_read_sec`

```
sqlite> SELECT count(page), sum(duration) FROM page_stat where id_book = 43;
534|19236

=> 36.022471910112358s
```

Both data wrangling approaches yield something reasonable:

```
sqlite> SELECT count(*), sum(durations) FROM ( SELECT min(sum(duration), 120) AS durations FROM page_stat where id_book = 43 GROUP BY page );
22|1183

53.772727272727273.
```

```
sqlite> SELECT count(*), sum(durations) FROM ( SELECT max(duration) AS durations FROM page_stat where id_book = 43 GROUP BY page );
22|1094

49.727272727272727s
```

NiLuJe:
And because I happened to have a copy of my Forma DB on hand, if I look at actual data from a read novel:

```
sqlite> SELECT count(DISTINCT page), sum(period) FROM page_stat where id_book = 76;
358|32276

90.16s
```

```
sqlite> SELECT count(page), sum(period) FROM page_stat where id_book = 76;
419|32276

79.03s
```

```
sqlite> SELECT count(*), sum(durations) FROM ( SELECT min(sum(period), 120) AS durations FROM page_stat where id_book = 76 GROUP BY page );
358|31809

88.85s
```

```
sqlite> SELECT count(*), sum(durations) FROM ( SELECT max(period) AS durations FROM page_stat where id_book = 76 GROUP BY page );
358|31334

87.53s
```

NiLuJe:
TL;DR: I think I'm fine with the clamp variant ;).

marek-g:
Sorry for repeating and thinking out loud - I'm a little tired today, but I'll try to summarize - maybe I need to sleep with it after I write it down.

---

What is the desired intention? So far it was inconsistent. The number of read pages was:

1. For the data from database - count(DISTINCT page)
2. For the part in the memory (up to 50 pages) - count(page), after some time (50 pages) - count(DISTINCT page).

Or as NiLuJe said: "The state before was a bit weirder: total_* was always counting distinct pages, while mem_* was always counting all pages.".

My PR has changed mem pages to count distinct pages. Do you say that the intention was to do the opposite?

---

If we go with the DISTINCT way - what worse can happen? We can read a page twice and have estimation wrong by a small factor. I assume that we don't read pages twice "for fun".

---

If we go with the every page counts way - what worse can happen? I think this:

You open new book. You look at the cover, look at the chapter list. Once again. The remaining time starts going down like a crazy. And is wrong so much that you think this is a bug until you find the min read limit setting. It is useless for the first chapter so you forget about the stats and turn them off.

You cannot set it to 30 as default, because it will disable the estimation on high zoom levels.

---

I think that I understand now what you were trying to say. The "every page" (second) approach is probably better in general. The fact that I read the page twice doesn't mean that I read slower now. Even if I had to read it again because I was de-concentrated. It doesn't matter - my speed of reading is like measured.

Even because of that, I like the DISTINCT method more for two reasons:
- it doesn't require me to set min read limit to work at all for me (subjective) - this is really disappointing if you don't know about it
- even if it doesn't precisely measure my speed of reading, it gives me more accurate results as it is measuring my speed of reading and level of de-concentration (assuming it is more or less constant during reading)

---

Summary:

Please, choose whichever version you like more. I will use the min read limit setting when needed. We don't need more settings as it's too much code to be written and it isn't really worth it. I should spend more time on reading books than playing with them and looking a way out to write code ;-))))

marek-g:
> Ha! Progress, I was almost there myself, thanks ;).
(As you may have realized by now, I've really almost never done SQL, ever ^^).

Good job, NiLuJe. SQL can be fun :)

> (NOTE: I've indeed renamed the awfully-named period column to duration ;)).

is it because a period is also a dot or an epoch or a menstruation? ;-))))))

NiLuJe:
> > Ha! Progress, I was almost there myself, thanks ;).
> > (As you may have realized by now, I've really almost never done SQL, ever ^^).
> 
> Good job, NiLuJe. SQL can be fun :)

Surprisingly versatile, too, for someone completely unfamiliar with it ;).

> 
> > (NOTE: I've indeed renamed the awfully-named period column to duration ;)).
> 
> is it because a period is also a dot or an epoch or a menstruation? ;-))))))

All of the above? ;P

Mostly it was confusing given that stuff is *actually* dealing with periods of time in there (e.g., hourly, weekly, daily, ...), and the "start of a session" timestamp is also called something_something_period...

So, making clear that this was a plain old duration felt better (in fact, someone had already realized that and stored a `sum(period) AS duration` in some existing queries ;p.

marek-g:
> TL;DR: I think I'm fine with the clamp variant ;).

It is too late to think clearly, but I think that if we got the setting to clamp with, we should use it anyway. The max() version looks similar to it only because you were not asleep :))) The max() could be.... another setting ;)

I think we deserve a PhD now :))))

NiLuJe:
Cool, I'll sleep on it, too, and I'll clean this up tomorrow ;). Many thanks for the lively discussion :).

Small reminder for you to look at https://github.com/koreader/koreader/pull/6761#issuecomment-705812502 once you've got the bandwidth to deal with it, because I really suck at SQL ;p.

poire-z:
> 4. Create new table with rows and join with it. Pros: we can create 1000 rows in primary index. It will be faster than it is now. The limit will be 1000x zoom. I've got a feeling it is not very beauty, too :)

> I would tend to agree. But for the SQL neophyte that I am, how would that look, in practice?

I don't think there's any mechanism in SQL or SQLite for this. So, probably just:
```
If table does not exist or older user_version:
  sqlite(CREATE TABLE zzz)
  for i=1, i<=1000 do -- Lua :)
     sqlite(INSERT INTO zzz (i))
  end
```
(I let @marek-g give you the table definition :)

marek-g:
Ok. If you are a perfectionist, then to the end :))) Thanks @poire-z for giving me that pleasure ;)

The `numbers` table creation script:

```SQL
CREATE TABLE IF NOT EXISTS numbers(
   number INTEGER PRIMARY KEY
);
WITH RECURSIVE counter AS 
(
   SELECT 1 as N UNION ALL
   SELECT N + 1 FROM counter WHERE N < 1000
)
INSERT INTO numbers SELECT N AS number FROM counter;
```

The new view (@NiLuJe please replace _period_ with _duration_ where needed) ;):

```SQL
-- this view rescales data from page_stat_internal table to current number of book pages
CREATE VIEW page_stat AS

    SELECT id_book, first_page + idx - 1 AS page, start_time, period / (last_page - first_page + 1) AS period FROM (

		SELECT id_book, page, total_pages, pages, start_time, period,
		
				-- first page_number of the page after rescaling single row
				((page - 1) * pages) / total_pages + 1 AS first_page,
				
				-- last page_number of the page after rescaling single row
				max(((page - 1) * pages) / total_pages + 1, (page * pages) / total_pages) AS last_page,

				idx

		FROM page_stat_internal
			
		JOIN book ON book.id = id_book
			
		-- duplicate rows for multiple pages if needed (as a result of rescaling)
		JOIN (SELECT number as idx FROM numbers) AS N ON idx <= (last_page - first_page + 1)
			
	)
```

And guys, I don't have data (yeah, I need to actually start reading... someday), but the query plan couldn't be better - pure index searches:

```
SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
SEARCH TABLE page_stat_internal USING INDEX page_stat_id_book (id_book=?)
SEARCH TABLE numbers USING INTEGER PRIMARY KEY (rowid<?)
```

I cannot believe that together we've come to such a clean solution to such a hard looking problem. I love that beauty :) Think of what a huge mess in the code could have we done if we had started from coding instead of planning.

marek-g:
You want believe! I almost started reading a book, but then came a little question to my mind. It's always like that ;)
And as I am directly talking with the most active contributors of KOReader I couldn't resist to ask:

Is adding TTS synthesis to KOReader welcomed?

poire-z:
> I don't think there's any mechanism in SQL or SQLite for this

OK, so there is some :) 
(@marek-g : did you already have some background in SQL ? Or have you just discovered and digested all that the last few days ?!)

> Is adding TTS synthesis to KOReader welcomed?

Not my be :)
Most devices don't have audio support. Possibly only Android, and we're not really an Android app, there are many layers till the Android core and its TTS stuff.
There is some kind of helpers in crengine to get Sentence start/end, because CoolReader (which is a more classic Android app) can to TTS.
But I may have messed with some text functions to better find word boundaries, and I know I did not update all the Sentence functions, because we didn't care about them. Upstream/CoolReader has since done some fix to that, but well, I still don't care about them.
So, possibly needs some work and glue in many places, for little benefit.
(So, not welcomed by me these days where many things are happening - may be in a few months when we are bored and lack ideas about what to fix and add :)

Frenzie:
Somewhat related: https://gitter.im/koreader/koreader?at=5f606ef35fde567ffa4a903e

poire-z:
Some timings with the latest query:

```
Run Time: real 0.005 user 0.004362 sys 0.000000      (more stable, around 0.005/0.007)
sqlite> explain query plan select * from page_stat where id_book=1453 order by start_time asc;
QUERY PLAN
|--SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH TABLE page_stat_internal USING INDEX page_stat_id_book (id_book=?)
|--SEARCH TABLE numbers USING INTEGER PRIMARY KEY (rowid<?)
`--USE TEMP B-TREE FOR ORDER BY
```
Some other with some of the Calendar View queries:

```
ReaderStatistics:getReadBookByDay(month)
Run Time: real 0.145 user 0.131525 sys 0.013609           (was 0.120 previously)
sqlite> explain query plan SELECT strftime('%Y-%m-%d', start_time, 'unixepoch', 'localtime') day, sum(period) duration, id_book book_id, book.title book_title FROM   page_stat JOIN   book on book.id = page_stat.id_book WHERE  strftime('%Y-%m', start_time, 'unixepoch', 'localtime') = '2020-09'  GROUP  BY strftime('%Y-%m-%d', start_time, 'unixepoch', 'localtime'), id_book, title ORDER BY day, duration desc, book_id, book_title;
QUERY PLAN
|--SCAN TABLE page_stat_internal USING INDEX page_stat_id_book
|--SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH TABLE numbers USING INTEGER PRIMARY KEY (rowid<?)
|--USE TEMP B-TREE FOR GROUP BY
`--USE TEMP B-TREE FOR ORDER BY

ReaderStatistics:getReadingRatioPerHourByDay(month)
Run Time: real 0.134 user 0.131879 sys 0.000017    (was 0.126 previously)
sqlite> explain query plan SELECT strftime('%Y-%m-%d', start_time, 'unixepoch', 'localtime') day, strftime('%H', start_time, 'unixepoch', 'localtime') hour, sum(period)/3600.0 ratio FROM   page_stat WHERE  strftime('%Y-%m', start_time, 'unixepoch', 'localtime') = '2020-09' GROUP  BY strftime('%Y-%m-%d', start_time, 'unixepoch', 'localtime'), strftime('%H', start_time, 'unixepoch', 'localtime') ORDER BY day, hour;
QUERY PLAN
|--SCAN TABLE page_stat_internal
|--SEARCH TABLE book USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH TABLE numbers USING INTEGER PRIMARY KEY (rowid<?)
`--USE TEMP B-TREE FOR GROUP BY
```

So, quite fine for the things that matter to me :)

May be give the "number" table a more explicite name: "_internal_number_table_for_pivoting_" or something stating what it's really used for.

marek-g:
Thanks @poire-z and @Frenzie for the info. By saying "welcomed" I meant preparing the PR myself, not asking you for doing it, because it is o much fun - I would be able to do it myself, but it is only wishful thinking as we are all short of time.

Yes, Android is possible out of the box. Windows also via it's SAPI5 API. And all the others platforms could by using open source tts frameworks. But there are other apps that we can use now for it, so this is not really needed now.

> (@marek-g : did you already have some background in SQL ? Or have you just discovered and digested all that the last few days ?!)

Yes, I know SQL. I'm using it once every a few years when project needs it :) But I what I have learned here is the "WITH RECURSIVE" statement. In my daily job I'm using C# and writing desktop apps, sometimes Android in Java. I know C++ and Qt, too. And at home I'm a huge fan of Rust (my number one favorite language).

I'm glad the timings look good now :)

Do you guys also have jobs as professional developers or are you programming for fun?

poire-z:
(I'm not a professional developer, more a system admin/engineer: linux, web stuff, dbs... I like coding to make my life and work easier and saner - but I hate being told what to develop :)

NiLuJe:
Not a developer either, at all, my major is English Lit. ;).

marek-g:
This is how the "migration" SQL script could look like:

```SQL
-- create page_stat_internal table
CREATE TABLE IF NOT EXISTS "page_stat_internal" (
	"id_book"	integer,
	"page"	integer NOT NULL,
	"start_time"	integer NOT NULL,
	"period"	integer NOT NULL,
	"total_pages"	INTEGER,
	FOREIGN KEY("id_book") REFERENCES "book"("id"),
	UNIQUE("page","start_time")
);

-- copy data
INSERT INTO page_stat_internal
SELECT id_book, page, start_time, period, pages as total_pages FROM page_stat
LEFT JOIN book on book.id = id_book;

-- remove page_stat table
DROP TABLE IF EXISTS page_stat;

-- create numbers table
CREATE TABLE IF NOT EXISTS numbers(
   number INTEGER PRIMARY KEY
);
WITH RECURSIVE counter AS 
(
   SELECT 1 as N UNION ALL
   SELECT N + 1 FROM counter WHERE N < 1000
)
INSERT INTO numbers SELECT N AS number FROM counter;

-- create page_stat view
-- this view rescales data from page_stat_internal table to current number of book pages
CREATE VIEW page_stat AS
SELECT id_book, first_page + idx - 1 AS page, start_time, period / (last_page - first_page + 1) AS period
FROM (
	SELECT id_book, page, total_pages, pages, start_time, period,
		
		-- first page_number of the page after rescaling single row
		((page - 1) * pages) / total_pages + 1 AS first_page,
				
		-- last page_number of the page after rescaling single row
		max(((page - 1) * pages) / total_pages + 1, (page * pages) / total_pages) AS last_page,

		idx

	FROM page_stat_internal
			
	JOIN book ON book.id = id_book
			
	-- duplicate rows for multiple pages if needed (as a result of rescaling)
	JOIN (SELECT number as idx FROM numbers) AS N ON idx <= (last_page - first_page + 1)		
)
```

I think the next steps (required in statistics plugin only) could be:
- check if this is the first run of a new version of plugin and then run the migration script, mark somewhere that it was ran
- plugin should record total number of pages additional to every page number in update method
- insertToDb should write to `page_stat_internal`, including `total_pages` column
- when change of total number of pages is detected in the update method, reload the data from DB

NiLuJe:
@marek-g: Thanks!

The db schema migration is already in place in the other branch (just not with the *right* schema yet ^^), and I've started handling the per-book migration (i.e., make sure total_pages is set to something).

have to run right now, will finish it later this afternoon ;).

marek-g:
> have to run right now, will finish it later this afternoon ;).

Thank you for taking care of it!! I really appreciate that! And take your time 
:) No hurry.

NiLuJe:
Okay, the heavy lifting should be done now, I'll just be going over what's not using total_ stuff correctly, and/or what needs clarification on what kind of page/time it's displaying ;).

But the good news is: it works ^^.

marek-g:
@NiLuJe That's fantastic :)) You're so quick!! Thanks!

NiLuJe:
Superseded by #6778 :)

PR #6766, opened 2020-10-07
enable RTL InfoMessage

yparitcher:
as suggested by @poire-z in https://github.com/koreader/koreader/pull/6729#issuecomment-700765094

I need it for a custom plugin https://github.com/yparitcher/chitas.koplugin that displays a InfoMessage in Hebrew (RTL)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6766)


Frenzie (approved):
lgtm but I'll wait for @poire-z to merge in case there's some RTL gotcha I might be missing.

poire-z (approved):
Looks fine, couldn't have written it better :)

PR #6767, opened 2020-10-08
Remove PB hardware requests

pazos:
Because @ezdiy has one (or more?) and is doing a great work with it.

I also have one (Touch HD3) which I don't use and I don't plan to use, so it's available for any contributor that wants to tinker with the platform. Thanks to @Cellaris for providing the device :smile:

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6767)


ptrm:
I continuously upgrade my PB632 to the nightlies, and leave this here if you need a test or two done by more people

Issue #6769, opened 2020-10-08
AutoSuspend plugin should not be disabled on remarkable


soaringgaijin:
I am running KOReader on a Remarkable tablet. I noted that the device never enters standby while running KOReader. Apparently, the original system software must be running to have the device sleep automatically.

However, setting "auto_suspend_time_seconds" in settings.reader.lua never seemed to change anything. I checked the plugin itself and noted that autosuspend.koplugin/main.lua starts with

```
if not Device:isCervantes() and not Device:isKobo() and not Device:isSDL() and not Device:isSonyPRSTUX()
then
    return { disabled = true, }
end
```

Comenting out those lines makes it work as intended, putting the Remarkable to sleep after the defined number of seconds.

Is there some specific reason why the plugin is not enabled by default on Remarkable? It is immensely useful, since I constantly forget to push the hardware button for standby when interrupting my reading, only to come back to a device with empty battery a few hours later.

pazos:
It was untested. Feel free to submit a PR if it works for you

gbyl:
@soaringgaijin please do submit a pull request. I'm going to test this out myself. I asked a similar question and this is the reply that I got https://github.com/koreader/koreader/issues/6028#issuecomment-698566960

gbyl:
I can confirm that this worked! With a minor caveat… it's *timeout* not "time". I commented out those lines from `/koreader/plugins/autosuspend.koplugin/main.lua` and added `["auto_suspend_timeout_seconds"] = 300,` to my `settings.reader.lua` and after 5 min it suspended automatically. Let's add it to the main branch.

gbyl:
Not only that, but autoshutdown works too now! @Eeems This might be of interest to you too.

Eeems:
> Not only that, but autoshutdown works too now! @Eeems This might be of interest to you too.

I already have a good idea of what has to happen for auto shutdown in Oxide. I just haven't had time to implement it.

That said, I will have to test to see if this plays nicely with oxide or not. Likely I'll have to submit a PR to make it play nice when running under Oxide.

Issue #6676, opened 2020-09-18
Screensaver not loading on reMarkable: Requires extra button press


gbyl:
* KOReader version: 2020.09
* Device: reMarkable 1

#### Issue
Power button suspends the device – but does not show the screensaver until one of the three main buttons is pressed

#### Steps to reproduce
1. Open epub/pdf file
2. Press power button
3. Screensaver does not load
4. Press either the left, middle or right button
5. Screensaver loads

### Additional information
For some strange reason it behaves correctly with pdf files but only if you suspend the device from the file browser/main screen:
1. Open pdf file
2. Press middle button
3. Press power button
4. Screensaver loads successfully

Eeems:
I'm not able to replicate this on my own device. That said, I do have a service running that grabs button input before anything starts and then is only passing through press events when they aren't long presses.

gabyhk:
@Eeems Is that service included in the latest release or do you have a custom build? Are there any other benefits of running that service?

Eeems:
@gabyhk It's not a KOReader service, and I'm not convinced it's what's keeping you from seeing the message. As for if there are any benefits, yes but it doesn't play nicely with KOReader yet: https://github.com/Eeems/oxide/issues/69

gabyhk:
@Eeems I am out of ideas at this point then. I thought it would be an easy fix, but apparently it isn't as reproducible as I assumed. Thank you for looking into it nonetheless!

Eeems:
@gabyhk I'll ask on the reMarkable community discord to see if anybody else can replicate.

gabyhk:
@Eeems I just made a little discovery! If I comment out `os.execute("systemctl suspend")` from here:

https://github.com/koreader/koreader/blob/2e731dd4ddc55711734cca3759f3a72d3747787b/frontend/device/remarkable/device.lua#L113-L115

then the screensaver loads as expected every single time! Somehow that `suspend` command fails to execute and keeps it from going to the next action which triggers the screensaver. This has to be related to https://github.com/koreader/koreader/issues/6028

Eeems:
So it might be a timing issue, I guess KOReader might draw the image after calling the suspend command? On my device that works though.

gabyhk:
@Eeems It seems like that is the case. I really don't know why my particular device (RM102) would be the only one afflicted by this. I did a factory reset (FW 2.3.0.16) and reinstalled KOReader from zero twice and the problem never went away… until I commented out that line.

[...]

PR #6772, opened 2020-10-08
A few reMarkable QoL fixes

NiLuJe:
* Enable AutoSuspend (#6769)
* Better power events handling (#6676)
* Enable HW inversion, because I don't see why it wouldn't work ;p.

Many thanks to the actual testers and their feedback in the respective issues :).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6772)

### DIFF
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -217,7 +217,7 @@ function Device:rescheduleSuspend()
     UIManager:scheduleIn(self.suspend_wait_timeout, self.suspend)
 end
 
--- ONLY used for Kobo and PocketBook devices
+-- Only used on platforms where we handle suspend ourselves.
 function Device:onPowerEvent(ev)
     if self.screen_saver_mode then
         if ev == "Power" or ev == "Resume" then
@@ -228,10 +228,12 @@ function Device:onPowerEvent(ev)
                 logger.dbg("Resuming...")
                 local UIManager = require("ui/uimanager")
                 UIManager:unschedule(self.suspend)
-                local network_manager = require("ui/network/manager")
-                if network_manager.wifi_was_on and G_reader_settings:isTrue("auto_restore_wifi") then
-                    network_manager:restoreWifiAsync()
-                    network_manager:scheduleConnectivityCheck()
+                if self:hasWifiManager() and not self:isEmulator() then
+                    local network_manager = require("ui/network/manager")
+                    if network_manager.wifi_was_on and G_reader_settings:isTrue("auto_restore_wifi") then
+                        network_manager:restoreWifiAsync()
+                        network_manager:scheduleConnectivityCheck()
+                    end
                 end
                 self:resume()
                 -- Restore to previous rotation mode, if need be.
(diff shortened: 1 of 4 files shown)


ezdiy:
hwinvert: `Hardware        : Freescale i.MX6 SoloLite (Device Tree)` should be fine

NiLuJe:
Especially since rM are probably the only ones actually using the latest NXP+EPDC kernel tree ;p.

Frenzie (approved):
lgtm

frontend/device/generic/device.lua
```diff
@@ -289,20 +291,30 @@ function Device:onPowerEvent(ev)
...
-          --       This is critical on Kobos (c.f., #3936), where it might still be on from KSM or Nickel,
-          --       without us being aware of it (i.e., wifi_was_on still unset or false),
-          --       because suspend will at best fail, and at worst deadlock the system if Wi-Fi is on,
-          --       regardless of who enabled it!
-          if network_manager:isWifiOn() then
-              network_manager:releaseIP()
-              network_manager:turnOffWifi()
+          -- NOTE: This side of the check needs to be laxer, some platforms can handle Wi-Fi without WifiManager ;).
+          if self:hasWifiToggle() then
+            local network_manager = require("ui/network/manager")
```
poire-z, line 308: Strange indentation around there (already there before your changes - but fixing it won't increase the diff :)

gbyl:
@NiLuJe Just letting you know that this bfe6820b58c2e8456db073f87f2c64b2002b8a24 actually broke Night Mode. It appears that whatever mechanism was triggering HW inversion before is running at the same time. So with that new commit, once you tap on Night Mode, it inverts twice – effectively undoing the inversion once you tap away to hide the menu.

NiLuJe:
@gbyl: I assume you're confusing a black flash along the way with a double-inversion, and this just means that HW inversion is borked for some strange reason.

Oh, well.

NiLuJe:
@gbyl: Just to double check, if you run

`/mnt/root/koreader/fbink -MmH "Hello World"`, is it displayed white on black?

(Feel free to fix KOReader's actual install path on the rM ^^).

gbyl:
@NiLuJe That's correct! "Hello World' is displayed white on black. rM's path is `home/root/koreader` I always wondered why it was never changed here https://github.com/koreader/koreader/blob/39faa8b860325c3c18372a4b24e50e1a1b407a8b/frontend/device/remarkable/device.lua#L20

NiLuJe:
Huh, then HW inversion *does* work. Grmbl.

NiLuJe:
Or it's buggy in *some* weird circumstances, which, err, we've unfortunately seen before ;p.

NiLuJe:
Next question: Since I've kind of lost track, is this with KOReader running @ 8bpp or at xochitl's stock 16bpp?

gbyl:
hmm I don't know how to check that

NiLuJe:
That would have been mentioned in the FBInk console output ;).

gbyl:
It says `8bpp`

NiLuJe:
Then, assuming the KOReader install is sane, that's... strange ;).

----

Re-reading your message, though, it *vaguely* looks like the PxP ignoring the flags passed to it after a (short) while in your case, which is similar to some of the weird issues I've mentioned before.

Just to confirm:

* Start with nightmode disabled.
* Open menu
* Tick Night mode
* Screen refreshes (the flash may be wonky or stroby, that's normal, nightmode is terrible).
* Screen is properly completely inverted
* Close menu (again, possibly stuttery/strobey flash, nightmode sucks)
* Screen is no longer completely inverted ?!

----

If that's what happens, can you try doing the same thing after unticking "Flash buttons and menu items", and ticking "Avoid mandatory black flashes in UI" in Gear > Screen >E-ink settings

Are the results similar? Or do you perhaps end up with only bits of screen inverted?

gbyl:
Actually… I forgot that I already changed `canHWInvert = yes,` to `canHWInvert = no,` before I ran the Hello World message.

On an unmodified v2020.09-57-g39faa8b, `/mnt/root/koreader/fbink -MmH "Hello World"` prints black on white.

gbyl:
> Just to confirm:
> 
> * Start with nightmode disabled.
> * Open menu
> * Tick Night mode
> * Screen refreshes (the flash may be wonky or stroby, that's normal, nightmode is terrible).
> * Screen is properly completely inverted
> * Close menu (again, possibly stuttery/strobey flash, nightmode sucks)
> * Screen is now  completely NOT inverted ?!

This is a perfect step-by-step description of what happens.

> If that's what happens, can you try doing the same thing after unticking "Flash buttons and menu items", and ticking "Avoid mandatory black flashes in UI" in Gear > Screen >E-ink settings
> Are the results similar? Or do you perhaps end up with only bits of screen inverted?

The results are the same.

NiLuJe:
`canHWInvert` should have zero incidence on FBInk (the two are completely orthogonal), so, I'm going to go ahead and assume, that, yes, HWInvert is wonky as hell :D.

gbyl:
Yeah, no big deal. Our time is better spent implementing more essential features like Wi-Fi management.

PR #6778, opened 2020-10-09
ReaderStatistics: Data collection improvements

NiLuJe:
This is essentially the continuation/culmination of both #6759 & #6761, many thanks to all those involved in the discussions and code :).

* Update the data collection format & handler to make it much less tortuous
* Update the pagecount & resync the stats on document layout changes
* Update the database schema to allow doing most queries against a SQL view that rescales the collected data to be accurate regardless of document layout (thanks to @marek-g for the SQL magic ;)).
* Add a "reset stats for current book" entry in the list of reset options, one that won't horribly break stats in said book ;).
* Fixed a couple of resource (SQL connection) leaks (in ReaderStatistics:getCurrentBookStats & ReaderStatistics:getCurrentBookStats).
* Flush stats to the DB on periodical metadata saves.
* Minor cosmetic tweaks to the code

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6778)

### DIFF
--- a/plugins/coverbrowser.koplugin/bookinfomanager.lua
+++ b/plugins/coverbrowser.koplugin/bookinfomanager.lua
@@ -108,10 +108,10 @@ end
 -- Build our most often used SQL queries according to columns
 local BOOKINFO_INSERT_SQL = "INSERT OR REPLACE INTO bookinfo " ..
                             "(" .. table.concat(BOOKINFO_COLS_SET, ",") .. ") " ..
-                            "VALUES (" .. table.concat(bookinfo_values_sql, ",") .. ")"
+                            "VALUES (" .. table.concat(bookinfo_values_sql, ",") .. ");"
 local BOOKINFO_SELECT_SQL = "SELECT " .. table.concat(BOOKINFO_COLS_SET, ",") .. " FROM bookinfo " ..
-                            "WHERE directory=? and filename=? and in_progress=0"
-local BOOKINFO_IN_PROGRESS_SQL = "SELECT in_progress, filename, unsupported FROM bookinfo WHERE directory=? and filename=?"
+                            "WHERE directory=? AND filename=? AND in_progress=0;"
+local BOOKINFO_IN_PROGRESS_SQL = "SELECT in_progress, filename, unsupported FROM bookinfo WHERE directory=? AND filename=?;"
 
 
 local BookInfoManager = {}
@@ -203,10 +203,10 @@ function BookInfoManager:compactDb()
     -- is bigger than available memory...)
     local prev_size = self:getDbSize()
     self:openDbConnection()
-    self.db_conn:exec("PRAGMA temp_store = 2") -- use memory for temp files
+    self.db_conn:exec("PRAGMA temp_store = 2;") -- use memory for temp files
     -- self.db_conn:exec("VACUUM")
     -- Catch possible "memory or disk is full" error
-    local ok, errmsg = pcall(self.db_conn.exec, self.db_conn, "VACUUM") -- this may take some time
+    local ok, errmsg = pcall(self.db_conn.exec, self.db_conn, "VACUUM;") -- this may take some time
     self:closeDbConnection()
     if not ok then
         return T(_("Failed compacting database: %1"), errmsg)
@@ -224,7 +224,7 @@ function BookInfoManager:loadSettings()
     end
     self.settings = {}
     self:openDbConnection()
-    local res = self.db_conn:exec("SELECT key, value FROM config")
+    local res = self.db_conn:exec("SELECT key, value FROM config";)
     local keys = res[1]
     local values = res[2]
     for i, key in ipairs(keys) do
@@ -247,7 +247,7 @@ function BookInfoManager:saveSetting(key, value)
         end
     end
     self:openDbConnection()
-    local query = "INSERT OR REPLACE INTO config (key, value) VALUES (?, ?)"
+    local query = "INSERT OR REPLACE INTO config (key, value) VALUES (?, ?);"
     local stmt = self.db_conn:prepare(query)
     if value == false then -- convert false to NULL
         value = nil
@@ -486,7 +486,7 @@ function BookInfoManager:setBookInfoProperties(filepath, props)
     self:openDbConnection()
     -- Let's do multiple one-column UPDATE (easier than building
     -- a multiple columns UPDATE)
-    local base_query = "UPDATE bookinfo SET %s=? WHERE directory=? AND filename=?"
+    local base_query = "UPDATE bookinfo SET %s=? WHERE directory=? AND filename=?;"
     for k, v in pairs(props) do
         local this_prop_query = string.format(base_query, k) -- add column name to query
         local stmt = self.db_conn:prepare(this_prop_query)
@@ -502,7 +502,7 @@ end
 function BookInfoManager:deleteBookInfo(filepath)
     local directory, filename = util.splitFilePathName(filepath)
     self:openDbConnection()
-    local query = "DELETE FROM bookinfo WHERE directory=? AND filename=?"
+    local query = "DELETE FROM bookinfo WHERE directory=? AND filename=?;"
     local stmt = self.db_conn:prepare(query)
     stmt:bind(directory, filename)
     stmt:step() -- commited
@@ -511,7 +511,7 @@ end
 
 function BookInfoManager:removeNonExistantEntries()
     self:openDbConnection()
-    local res = self.db_conn:exec("SELECT bcid, directory || filename FROM bookinfo")
+    local res = self.db_conn:exec("SELECT bcid, directory || filename FROM bookinfo;")
     if not res then
         return _("Cache is empty. Nothing to prune.")
     end
@@ -523,7 +523,7 @@ function BookInfoManager:removeNonExistantEntries()
             table.insert(bcids_to_remove, tonumber(bcids[i]))
         end
     end
-    local query = "DELETE FROM bookinfo WHERE bcid=?"
+    local query = "DELETE FROM bookinfo WHERE bcid=?;"
     local stmt = self.db_conn:prepare(query)
     for i=1, #bcids_to_remove do
         stmt:bind(bcids_to_remove[i])
--- a/plugins/coverbrowser.koplugin/bookinfomanager.lua
+++ b/plugins/coverbrowser.koplugin/bookinfomanager.lua
@@ -224,7 +224,7 @@ function BookInfoManager:loadSettings()
     end
     self.settings = {}
     self:openDbConnection()
-    local res = self.db_conn:exec("SELECT key, value FROM config";)
+    local res = self.db_conn:exec("SELECT key, value FROM config;")
     local keys = res[1]
     local values = res[2]
     for i, key in ipairs(keys) do
--- a/plugins/coverbrowser.koplugin/bookinfomanager.lua
+++ b/plugins/coverbrowser.koplugin/bookinfomanager.lua
@@ -173,7 +173,7 @@ function BookInfoManager:openDbConnection()
         self:createDB()
     end
     self.db_conn = SQ3.open(self.db_location)
-    xutil.sqlite_set_timeout(self.db_conn, 5000) -- 5 seconds
+    self.db_conn:set_busy_timeout(5000) -- 5 seconds
 
     -- Prepare our most often used SQL statements
     self.set_stmt = self.db_conn:prepare(BOOKINFO_INSERT_SQL)
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -214,6 +214,11 @@ function ReaderStatistics:onUpdateToc()
 
     -- Update our copy of the page count
     self.data.pages = new_pagecount
+
+    -- FIXME: Regardless of what we did, we need to update book's pages with the new pagecount to make the VIEW behave...
+    --        insertDB itself does it, but it runs *before* we update the var.
+    --        We'll probably be able to update the var first when switching to the VIEW approach,
+    --        but we'll still need an extra query in the cases where insertDB returns early (i.e., the resetVolatileStats branch).
 end
 
 function ReaderStatistics:resetVolatileStats(now_ts)
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1932,7 +1932,7 @@ function ReaderStatistics:resetCurrentBook()
             logger.info("ReaderStatistics: Initializing average time per page at 75% of the max value, i.e.,", self.avg_time)
 
             -- And the current volatile stats
-            self:resetVolatileStats()
+            self:resetVolatileStats(TimeVal:now().sec)
 
             -- And re-create the Book's data in the book table and get its new ID...
             self.id_curr_book = self:getIdBookDB()
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -491,10 +491,10 @@ function ReaderStatistics:upgradeDB(conn)
             LEFT JOIN book on book.id = id_book;
 
         -- Update the index, too
-        DROP INDEX IF EXISTS page_stat_id_book;
         CREATE INDEX IF NOT EXISTS page_stat_data_id_book ON page_stat_data(id_book);
 
         -- Drop old page_stat table
+        DROP INDEX IF EXISTS page_stat_id_book;
         DROP TABLE IF EXISTS page_stat;
     ]]
     conn:exec(sql_stmt)
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -27,7 +27,6 @@ local T = FFIUtil.template
 local statistics_dir = DataStorage:getDataDir() .. "/statistics/"
 local db_location = DataStorage:getSettingsDir() .. "/statistics.sqlite3"
 local MAX_PAGETURNS_BEFORE_FLUSH = 50
-local MIN_PAGES_READ_TO_FLUSH = 2
 local DEFAULT_MIN_READ_SEC = 5
 local DEFAULT_MAX_READ_SEC = 120
 local DEFAULT_CALENDAR_START_DAY_OF_WEEK = 2 -- Monday
@@ -218,30 +217,9 @@ function ReaderStatistics:onUpdateToc()
     local new_pagecount = self.view.document:getPageCount()
 
     if new_pagecount ~= self.data.pages then
-        -- NOTE: insertDB does the exact same check
-        if not self.id_curr_book or self.mem_read_pages < MIN_PAGES_READ_TO_FLUSH then
-            logger.info("ReaderStatistics: Pagecount change, clearing volatile book statistics")
-            -- Clear volatile stats
-            self:resetVolatileStats(TimeVal:now().sec)
-
-            -- Update the pagecount in the database to let the view do its job properly
-            if self.id_curr_book then
-                local conn = SQ3.open(db_location)
-                local sql_stmt = [[
-                    UPDATE book
-                    SET    pages = ?
-                    WHERE  id = ?;
-                ]]
-                local stmt = conn:prepare(sql_stmt)
-                stmt:reset():bind(new_pagecount, self.id_curr_book):step()
-                stmt:close()
-                conn:close()
-            end
-        else
-            logger.info("ReaderStatistics: Pagecount change, flushing volatile book statistics")
-            -- Flush volatile stats to DB for current book, and update pagecount and average time per page stats
-            self:insertDB(self.id_curr_book, new_pagecount)
-        end
+        logger.info("ReaderStatistics: Pagecount change, flushing volatile book statistics")
+        -- Flush volatile stats to DB for current book, and update pagecount and average time per page stats
+        self:insertDB(self.id_curr_book, new_pagecount)
     end
 
     -- Update our copy of the page count
@@ -717,7 +695,7 @@ function ReaderStatistics:getIdBookDB()
 end
 
 function ReaderStatistics:insertDB(id_book, updated_pagecount)
-    if id_book == nil or self.mem_read_pages < MIN_PAGES_READ_TO_FLUSH then
+    if not id_book then
         return
     end
     local now_ts = TimeVal:now().sec
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1247,10 +1247,6 @@ function ReaderStatistics:getCurrentStat(id_book)
     --       which is computed slightly differently (c.f., insertDB), we'll be using this tweaked book read time
     --       to compute the other time-based statistics...
     local book_read_pages, book_read_time = self:getPageTimeTotalStats(id_book)
-    -- Warn if the book recap doesn't agree with the page_stat data...
-    if total_read_pages ~= book_read_pages then
-        logger.dbg("Tables disagree on the total amount of distinct pages read! page_stat says", total_read_pages, "while book says", book_read_pages, "id:", id_book)
-    end
 
     if total_time_book == nil then
         total_time_book = 0
@@ -1346,10 +1342,6 @@ function ReaderStatistics:getBookStat(id_book)
     conn:close()
 
     local book_read_pages, book_read_time = self:getPageTimeTotalStats(id_book)
-    -- Warn if the book recap doesn't agree with the page_stat data...
-    if total_read_pages ~= book_read_pages then
-        logger.dbg("Tables disagree on the total amount of distinct pages read! page_stat says", total_read_pages, "while book says", book_read_pages, "id:", id_book)
-    end
 
     if total_time_book == nil then
         total_time_book = 0
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -604,7 +604,9 @@ function ReaderStatistics:getIdBookDB()
         result = stmt:reset():bind(self.data.title, self.data.authors, self.data.md5):step()
         id_book = result[1]
     end
+    stmt:close()
     conn:close()
+
     return tonumber(id_book)
 end
 
@@ -691,6 +693,9 @@ function ReaderStatistics:insertDB(id_book)
     stmt = conn:prepare(sql_stmt)
     stmt:reset():bind(now_ts, self.data.notes, self.data.highlights, total_read_time, rescaled_total_read_pages,
         self.data.pages, progress_str, id_book):step()
+    stmt:close()
+    conn:close()
+
     if total_read_pages then
         self.total_read_pages = tonumber(total_read_pages)
     else
@@ -703,7 +708,6 @@ function ReaderStatistics:insertDB(id_book)
     end
 
     self:resetVolatileStats(now_ts)
-    conn:close()
 end
 
 function ReaderStatistics:getPageTimeTotalStats(id_book)
@@ -718,6 +722,8 @@ function ReaderStatistics:getPageTimeTotalStats(id_book)
         WHERE  id = %d;
     ]]
     local total_pages, total_time = conn:rowexec(string.format(sql_stmt, id_book))
+    conn:close()
+
     if total_pages then
         total_pages = tonumber(total_pages)
     else
@@ -728,7 +734,6 @@ function ReaderStatistics:getPageTimeTotalStats(id_book)
     else
         total_time = 0
     end
-    conn:close()
     return total_pages, total_time
 end
 
@@ -1107,6 +1112,8 @@ function ReaderStatistics:getTodayBookStats()
                 );
     ]]
     local today_pages, today_duration = conn:rowexec(string.format(sql_stmt, start_today_time))
+    conn:close()
+
     if today_pages == nil then
         today_pages = 0
     end
@@ -1115,7 +1122,6 @@ function ReaderStatistics:getTodayBookStats()
     end
     today_duration = tonumber(today_duration)
     today_pages = tonumber(today_pages)
-    conn:close()
     return today_duration, today_pages
 end
 
@@ -1132,6 +1138,8 @@ function ReaderStatistics:getCurrentBookStats()
                );
     ]]
     local current_pages, current_duration = conn:rowexec(string.format(sql_stmt, self.start_current_period))
+    conn:close()
+
     if current_pages == nil then
         current_pages = 0
     end
@@ -1261,7 +1269,6 @@ function ReaderStatistics:getBookStat(id_book)
         WHERE  id_book = %d;
     ]]
     local total_time_book, total_read_pages, first_open = conn:rowexec(string.format(sql_stmt, id_book))
-
     conn:close()
 
     if total_time_book == nil then
@@ -1471,6 +1478,7 @@ function ReaderStatistics:getDatesFromAll(sdays, ptype, book_mode)
     local conn = SQ3.open(db_location)
     local result_book = conn:exec(string.format(sql_stmt_res_book, period_begin))
     conn:close()
+
     if result_book == nil then
         return {}
     end
@@ -1562,6 +1570,7 @@ function ReaderStatistics:getDaysFromPeriod(period_begin, period_end)
     local conn = SQ3.open(db_location)
     local result_book = conn:exec(string.format(sql_stmt_res_book, period_begin, period_end - 1))
     conn:close()
+
     if result_book == nil then
         return {}
     end
@@ -1605,6 +1614,7 @@ function ReaderStatistics:getBooksFromPeriod(period_begin, period_end, callback_
     local conn = SQ3.open(db_location)
     local result_book = conn:exec(string.format(sql_stmt_res_book, period_begin + 1, period_end))
     conn:close()
+
     if result_book == nil then
         return {}
     end
@@ -1668,6 +1678,8 @@ function ReaderStatistics:getReadingProgressStats(sdays)
         ORDER  BY dates DESC;
     ]]
     local result_book = conn:exec(string.format(sql_stmt, period_begin))
+    conn:close()
+
     if not result_book then return end
     for i = 1, sdays do
         local pages = tonumber(result_book[2][i])
@@ -1681,7 +1693,6 @@ function ReaderStatistics:getReadingProgressStats(sdays)
             date_read
         })
     end
-    conn:close()
     return results
 end
 
@@ -1699,6 +1710,7 @@ function ReaderStatistics:getDatesForBook(id_book)
     ]]
     local result_book = conn:exec(string.format(sql_stmt, id_book))
     conn:close()
+
     if result_book == nil then
         return {}
     end
@@ -1774,6 +1786,7 @@ function ReaderStatistics:getTotalStats()
         })
     end
     conn:close()
+
     return T(_("Total time spent reading: %1"), util.secondsToClock(total_books_time, false)), total_stats
 end
 
@@ -1880,6 +1893,7 @@ function ReaderStatistics:resetBook()
         end
     end
     conn:close()
+
     kv_reset_book = KeyValuePage:new{
         title = _("Reset book statistics"),
         value_align = "right",
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -722,6 +722,7 @@ function ReaderStatistics:insertDB(id_book, updated_pagecount)
     -- NOTE: See the tail end of the discussions in #6761 for more context on the choice of this heuristic.
     --       Basically, we're counting distinct pages,
     --       while making sure the sum of durations per distinct page is clamped to self.page_max_read_sec
+    --       This is expressly tailored to a fairer computation of self.avg_time ;).
     sql_stmt = [[
         SELECT count(*),
                sum(durations)
@@ -1227,6 +1228,9 @@ function ReaderStatistics:getCurrentStat(id_book)
     ]]
     local total_days = conn:rowexec(string.format(sql_stmt, id_book))
 
+    -- NOTE: Here, we really want to account for the *full* amount of time spent reading this book.
+    --       The "Average time per page" entry is already re-using self.avg_time, which is computed slightly differently,
+    --       c.f., insertDB
     sql_stmt = [[
         SELECT sum(duration),
                count(DISTINCT page),
@@ -1316,6 +1320,7 @@ function ReaderStatistics:getBookStat(id_book)
     ]]
     local total_days = conn:rowexec(string.format(sql_stmt, id_book))
 
+    -- NOTE: Unlike getCurrentStat, we can't use selv.avg_time here, because this may not be the current book.
     sql_stmt = [[
         SELECT sum(duration),
                count(DISTINCT page),
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1643,7 +1643,6 @@ end
 
 function ReaderStatistics:getReadingProgressStats(sdays)
     local results = {}
-    local pages, duration, date_read
     local now_t = os.date("*t")
     local from_begin_day = now_t.hour *3600 + now_t.min*60 + now_t.sec
     local now_stamp = os.time()
@@ -1669,9 +1668,9 @@ function ReaderStatistics:getReadingProgressStats(sdays)
     local result_book = conn:exec(string.format(sql_stmt, period_begin))
     if not result_book then return end
     for i = 1, sdays do
-        pages = tonumber(result_book[2][i])
-        duration = tonumber(result_book[3][i])
-        date_read = result_book[1][i]
+        local pages = tonumber(result_book[2][i])
+        local duration = tonumber(result_book[3][i])
+        local date_read = result_book[1][i]
         if pages == nil then pages = 0 end
         if duration == nil then duration = 0 end
         table.insert(results, {
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -402,19 +402,13 @@ local STATISTICS_DB_PAGE_STAT_VIEW_SCHEMA = [[
         SELECT id_book, first_page + idx - 1 AS page, start_time, duration / (last_page - first_page + 1) AS duration
         FROM (
             SELECT id_book, page, total_pages, pages, start_time, duration,
-
                 -- First page_number for this page after rescaling single row
                 ((page - 1) * pages) / total_pages + 1 AS first_page,
-
                 -- Last page_number for this page after rescaling single row
                 max(((page - 1) * pages) / total_pages + 1, (page * pages) / total_pages) AS last_page,
-
                 idx
-
             FROM page_stat_data
-
             JOIN book ON book.id = id_book
-
             -- Duplicate rows for multiple pages as needed (as a result of rescaling)
             JOIN (SELECT number as idx FROM numbers) AS N ON idx <= (last_page - first_page + 1)
         );
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -140,15 +140,15 @@ function ReaderStatistics:init()
     Screensaver.getReaderProgress = function()
         local readingprogress
         self:insertDB(self.id_curr_book)
-        local current_period, current_pages = self:getCurrentBookStats()
-        local today_period, today_pages = self:getTodayBookStats()
+        local current_duration, current_pages = self:getCurrentBookStats()
+        local today_duration, today_pages = self:getTodayBookStats()
         local dates_stats = self:getReadingProgressStats(7)
         if dates_stats then
             readingprogress = ReaderProgress:new{
                 dates = dates_stats,
-                current_period = current_period,
+                current_duration = current_duration,
                 current_pages = current_pages,
-                today_period = today_period,
+                today_duration = today_duration,
                 today_pages = today_pages,
                 readonly = true,
             }
@@ -250,7 +250,7 @@ function ReaderStatistics:getStatsBookStatus(id_curr_book, stat_enable)
     ]]
     local total_days = conn:rowexec(string.format(sql_stmt, id_curr_book))
     sql_stmt = [[
-        SELECT sum(period),
+        SELECT sum(duration),
                count(DISTINCT page)
         FROM   page_stat
         WHERE  id_book = '%s'
@@ -387,7 +387,7 @@ function ReaderStatistics:createDB(conn)
                 id_book     integer,
                 page        integer NOT NULL DEFAULT 0,
                 start_time  integer NOT NULL DEFAULT 0,
-                period      integer NOT NULL DEFAULT 0,
+                duration    integer NOT NULL DEFAULT 0,
                 total_pages integer NOT NULL DEFAULT 0,
                 UNIQUE (page, start_time),
                 FOREIGN KEY(id_book) REFERENCES book(id)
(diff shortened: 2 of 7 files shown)


NiLuJe:
There's obviously a database migration involved, so, testers, backup your `settings/statistics.sqlite3` DB first ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -1862,38 +2102,75 @@ function ReaderStatistics:onPageUpdate(pageno)
...
+    if self.curr_page == pageno then
+        return
+    end
+
+    self.pageturn_count = self.pageturn_count + 1
+    local now_ts = TimeVal:now().sec
+
+    -- Get the previous page's last timestamp (if there is one)
+    local page_data = self.page_stat[self.curr_page]
+    -- This is a list of tuples, in insertion order, we want the last one
```
poire-z, line 2116: (Haven't re-read this part. Can I assume it hasn't changed much since the other PR?)

poire-z (commented):
While we were caring mostly about EPUBs and layout changes, has all this been checked with PDF books?

marek-g:
> Or having one that includes (book_id, page) could be more useful

I think you are right. AFAIK, if we always narrow results by book_id than any other index that don't start with the book_id won't be used.

Galunid (commented):
Just two overall comments, I believe you don't need ';' at the end of SQL statement, if there's no following statement, kind of like lua ;)
Another thing is usage of something like `conn:exec(string.format("QUERY WHERE='%s'"), var)`. This is generally bad idea, because of SQL Injection attack. The preferred way to do this is something similar to `conn:prepare("QUERY WHERE=?"):bind(variable)`. It has added benefit of being faster. In this case we don't care too much about that tough. It's just something worth keeping in mind.

plugins/statistics.koplugin/main.lua
```diff
@@ -533,47 +684,59 @@ function ReaderStatistics:getIdBookDB()
...
-               sum(period)
-        FROM   page_stat
-        WHERE  id_book = '%s'
+        SELECT count(*),
+               sum(durations)
+        FROM (
+            SELECT min(sum(duration), %d) AS durations
+            FROM page_stat_data
+            WHERE id_book = %d
+            GROUP BY page
```
marek-g, line 736: As I'm ok with the count(*), we have definitely forget here about the `self.page_min_read_sec`. (the funny thing is that page_min_read_sec is not working correctly even on the last official release build).

I think the we should check this in the HAVING statement (per every entry, not on the grouped by). And also the change with min(sum()) to sum(min()) with page_max_read_sec - see what's going on if we read page twice (total_pages = 2, read_time = once? - this is wrong).

I propose:

SELECT sum(min(duration, %d)) AS durations --- sum(min())
            FROM page_stat_data
            WHERE id_book = %d
            GROUP BY page
            HAVING duration > %d -- page_min_read_sec

Sorry for that at it was mine quick sql example :)

plugins/statistics.koplugin/main.lua
```diff
@@ -1862,38 +2102,75 @@ function ReaderStatistics:onPageUpdate(pageno)
...
+
+    -- By now, we're sure that we actually have a tuple (and the rest of the code ensures they're sane, i.e., zero-initialized)
+    local curr_duration = data_tuple[2]
+    -- NOTE: If all goes well, given the earlier curr_page != pageno check, curr_duration should always be 0 here.
+    -- Compute the difference between now and the previous page's last timestamp
+    local diff_time = now_ts - then_ts
+    if diff_time >= self.page_min_read_sec and diff_time <= self.page_max_read_sec then
+        self.mem_read_time = self.mem_read_time + diff_time
+        -- If it's the first time we're computing a duration for this page, count it as read
+        if #page_data == 1 and curr_duration == 0 then
```
marek-g, line 2136: I would resign from the #page_data == 1. Isn't it equivalent of count(DISTINCT page) in SQL?

plugins/statistics.koplugin/main.lua
```diff
@@ -1862,38 +2102,75 @@ function ReaderStatistics:onPageUpdate(pageno)
...
-                read_pages_set[self.pages_stats[sorted_performance[i]]] = true
-            elseif diff_time > self.page_max_read_sec then
-                mem_read_time = mem_read_time + self.page_max_read_sec
-                read_pages_set[self.pages_stats[sorted_performance[i]]] = true
-            end
+        -- Update the tuple with the computed duration
+        data_tuple[2] = curr_duration + diff_time
+    elseif diff_time > self.page_max_read_sec then
+        self.mem_read_time = self.mem_read_time + self.page_max_read_sec
+        if #page_data == 1 and curr_duration == 0 then
```
marek-g, line 2143: And the same here.

marek-g:
Thanks @NiLuJe for the PR. It's a really great job.

I was able to quickly scan through code and checked mostly the read_pages and read_time variables.

Unfortunatelly, I'm out on the weekend. Will be able to look closer on Monday. Have a good weekend, guys!

marek-g (changes requested):

poire-z:
> Or having one that includes (book_id, page) could be more useful

> I think you are right. 

Although I dunno if anything would use the `page` part of such index? Would it be helpful for the VIEW ?

marek-g:
> > Or having one that includes (book_id, page) could be more useful
> 
> > I think you are right.
> 
> Although I dunno if anything would use the `page` part of such index? Would it be helpful for the VIEW ?

Now I think that UNIQUE index is added to check if there are unique pairs in the database. Not to speed up queries. If that's the case, it's probably fine, however the better would be a triple (book_id, page, start_time):
- if you could merge two databases in the future and you read at the same time on two devices ;)
- it will be used when looking for a given page / pages in a book

poire-z:
> however the better would be a triple (book_id, page, start_time):

Agreed.

> it will be used when looking for a given page / pages in a book

I don't think this would even happen (?).
But would it be useful/used for/by your VIEW, or not at all?

plugins/statistics.koplugin/main.lua
```diff
@@ -175,14 +180,65 @@ function ReaderStatistics:initData()
...
+-- Reset the (volatile) stats on page count changes (e.g., after a font size update)
+function ReaderStatistics:onUpdateToc()
+    local new_pagecount = self.view.document:getPageCount()
+
+    if new_pagecount ~= self.data.pages then
+        -- NOTE: insertDB does the exact same check
+        if not self.id_curr_book or self.mem_read_pages < MIN_PAGES_READ_TO_FLUSH then
+            logger.info("ReaderStatistics: Pagecount change, clearing volatile book statistics")
+            -- Clear volatile stats
+            self:resetVolatileStats(TimeVal:now().sec)
```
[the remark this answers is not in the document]
NiLuJe: Nope, it's the exact same check as insertDB, it aborts early below MIN_PAGES_READ_TO_FLUSH (i.e., 2).

This just ensures the collected data is dropped, too, because it'll have a bogus pagecount.

plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -108,10 +108,10 @@ end
 -- Build our most often used SQL queries according to columns
 local BOOKINFO_INSERT_SQL = "INSERT OR REPLACE INTO bookinfo " ..
                             "(" .. table.concat(BOOKINFO_COLS_SET, ",") .. ") " ..
-                            "VALUES (" .. table.concat(bookinfo_values_sql, ",") .. ")"
+                            "VALUES (" .. table.concat(bookinfo_values_sql, ",") .. ");"
 local BOOKINFO_SELECT_SQL = "SELECT " .. table.concat(BOOKINFO_COLS_SET, ",") .. " FROM bookinfo " ..
-                            "WHERE directory=? and filename=? and in_progress=0"
-local BOOKINFO_IN_PROGRESS_SQL = "SELECT in_progress, filename, unsupported FROM bookinfo WHERE directory=? and filename=?"
+                            "WHERE directory=? AND filename=? AND in_progress=0;"
+local BOOKINFO_IN_PROGRESS_SQL = "SELECT in_progress, filename, unsupported FROM bookinfo WHERE directory=? AND filename=?;"
```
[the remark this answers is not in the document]
NiLuJe: Sure: that was mainly to make the seven billion characters ones and/or those split among several lines clearer. (as in, "the buck stops here" ^^).

And then to make it consistent I also tweaked the short ones.

But, yes, not technically necessary ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -26,12 +26,16 @@ local T = FFIUtil.template
...
-local PAGE_INSERT = 50
+local MAX_PAGETURNS_BEFORE_FLUSH = 50
+local MIN_PAGES_READ_TO_FLUSH = 2
 local DEFAULT_MIN_READ_SEC = 5
 local DEFAULT_MAX_READ_SEC = 120
 local DEFAULT_CALENDAR_START_DAY_OF_WEEK = 2 -- Monday
 local DEFAULT_CALENDAR_NB_BOOK_SPANS = 3
 
+-- Current DB schema version
+local DB_SCHEMA_VERSION = 1
```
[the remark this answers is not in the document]
NiLuJe: Works for me, as long as it's an int ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -305,14 +368,68 @@ function ReaderStatistics:partialMd5(file)
...
+-- Mainly so we don't duplicate the schema twice between the creation/upgrade codepaths
+local STATISTICS_DB_PAGE_STAT_DATA_SCHEMA = [[
+    CREATE TABLE IF NOT EXISTS page_stat_data
+        (
+            id_book     integer,
+            page        integer NOT NULL DEFAULT 0,
+            start_time  integer NOT NULL DEFAULT 0,
+            duration    integer NOT NULL DEFAULT 0,
+            total_pages integer NOT NULL DEFAULT 0,
+            UNIQUE (page, start_time),
```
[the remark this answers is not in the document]
NiLuJe: The UNIQUE constraint was always there, FWIW, I haven't touched that ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -328,34 +445,67 @@ function ReaderStatistics:createDB(conn)
...
+        -- NOTE: While doing a per-book migration could ensure a potentially more accurate page count,
+        --       we need to populate total_pages *now*, or queries against unopened books would return completely bogus values...
+        --       We'll just have to hope the current value of the column pages in the book table is not too horribly out of date,
+        --       and not too horribly out of phase with the actual page count at the time the data was originally collected...
+        INSERT INTO page_stat_data
+            SELECT id_book, page, start_time, duration, pages as total_pages FROM page_stat
+            LEFT JOIN book on book.id = id_book;
+
+        -- Update the index, too
+        DROP INDEX IF EXISTS page_stat_id_book;
```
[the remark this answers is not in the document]
NiLuJe: I wanted to drop the index *after* the table, because I suppose it's still useful during the migration?

(Remember, I suck at SQL ;o)).

plugins/statistics.koplugin/main.lua
```diff
@@ -328,34 +445,67 @@ function ReaderStatistics:createDB(conn)
...
-    stmt:close()
+
+    -- page_stat view
+    conn:exec(STATISTICS_DB_PAGE_STAT_VIEW_SCHEMA)
+
+    -- DB schema version
+    conn:exec(string.format("PRAGMA user_version=%d;", DB_SCHEMA_VERSION))
+end
+
+function ReaderStatistics:upgradeDB(conn)
```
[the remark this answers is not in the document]
NiLuJe: Stuff blows up in fun and interesting ways?

plugins/statistics.koplugin/main.lua
```diff
@@ -533,47 +684,59 @@ function ReaderStatistics:getIdBookDB()
...
+                stmt:reset():bind(id_book, page, ts, duration, self.data.pages):step()
+            end
         end
     end
-    conn:exec('COMMIT')
+    conn:exec('COMMIT;')
+    -- NOTE: See the tail end of the discussions in #6761 for more context on the choice of this heuristic.
+    --       Basically, we're counting distinct pages,
+    --       while making sure the sum of durations per distinct page is clamped to self.page_max_read_sec
+    --       This is expressly tailored to a fairer computation of self.avg_time ;).
```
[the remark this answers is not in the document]
NiLuJe: It's also stored in the book table in the total_* fields, and it is also used for other time-based estimates/averages down the line, so, this is its rightful place ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -1445,22 +1634,23 @@ function ReaderStatistics:getDaysFromPeriod(period_begin, period_end)
...
+               sum(sum_duration)    AS durations,
                start_time
         FROM   (
                     SELECT strftime('%%Y-%%m-%%d', start_time, 'unixepoch', 'localtime') AS dates,
-                           sum(period)                                                   AS sum_period,
+                           sum(duration)                                                 AS sum_duration,
                            start_time
                     FROM   page_stat
-                    WHERE  start_time >= '%s' AND start_time < '%s'
+                    WHERE  start_time BETWEEN %d AND %d
```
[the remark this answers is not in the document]
NiLuJe: Yep ^^.

plugins/statistics.koplugin/main.lua
NiLuJe: It hasn't, at all ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -1086,24 +1264,27 @@ function ReaderStatistics:getCurrentStat(id_book)
...
-        { _("Time spent reading today"), util.secondsToClock(today_period, false) },
+        { _("Time spent reading today"), util.secondsToClock(today_duration, false) },
         { _("Pages read today"), tonumber(today_pages) },
         "----",
         -- Current book statistics
-        { _("Time spent reading this book"), util.secondsToClock(total_time_book, false) },
+        -- Includes re-reads
+        { _("Total time spent on this book"), util.secondsToClock(total_time_book, false) },
+        -- Capped to self.page_max_read_sec per distinct page
+        { _("Time spent reading this book"), util.secondsToClock(book_read_time, false) },
```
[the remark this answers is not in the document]
NiLuJe: c.f., https://github.com/koreader/koreader/pull/6778#discussion_r502790021

`book.total_read_time` is only accessed:

* Directly by `insertDB` to `compute self.avg_time`
* `getPageTimeTotalStats`, as used by `getCurrentStat` & `getBookStat` to compute time-based estimates & averages.

plugins/statistics.koplugin/main.lua
NiLuJe: It does ensure that we only account for a page the first time we collect it (i.e., this is essentially #6759 adapted to the new data layout).

See the NOTE right below ;).

plugins/statistics.koplugin/main.lua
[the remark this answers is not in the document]
poire-z: Oh, ok, got confused, thought it was the other variable with 50.

plugins/statistics.koplugin/main.lua
NiLuJe: The data collection handler ensures that *nothing* will ever be collected below `self.page_min_read_sec`.

... at least as far as `self.page_min_read_sec` is concerned *at the time of collection* ;).

Adding a constraint here would essentially only be useful if `self.page_min_read_sec` was modified in the middle of reading *this* book.

Which is fine by me, but may not actually be all that useful in practice ^^.

plugins/statistics.koplugin/main.lua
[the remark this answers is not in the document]
poire-z: Yep, just thinking while we're at it if it's useful or not.
Having UNIQUE (id_book, page, start_time) logically makes more sense (a page is nothing without its book).
But then, it becomes an index, and the other one explicitely created is an uneeded duplicate.

NiLuJe:
> Just two overall comments, I believe you don't need ';' at the end of SQL statement, if there's no following statement, kind of like lua ;)

Answered https://github.com/koreader/koreader/pull/6778#discussion_r502789430 ;)

> Another thing is usage of something like `conn:exec(string.format("QUERY WHERE='%s'"), var)`. This is generally bad idea, because of SQL Injection attack. The preferred way to do this is something similar to `conn:prepare("QUERY WHERE=?"):bind(variable)`. It has added benefit of being faster. In this case we don't care too much about that tough. It's just something worth keeping in mind.

It's indeed harmless here (we mostly ever format integers), but I do agree about that in principle, and the Jekyll/Hyde mixed approach in the existing code left me puzzled. I was just too lazy to fix it ;p.

plugins/statistics.koplugin/main.lua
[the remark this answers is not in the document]
poire-z: Yes, it would still be after :) "Move this DROP INDEX _before_ the DROP TABLE _below_":

```
        CREATE INDEX IF NOT EXISTS page_stat_data_id_book ON page_stat_data(id_book);

        -- Drop old page_stat table
        DROP INDEX IF EXISTS page_stat_id_book;
        DROP TABLE IF EXISTS page_stat;
```

NiLuJe:
> While we were caring mostly about EPUBs and layout changes, has all this been checked with PDF books?

I did vaguely check that nothing horrible happened, but I'm not used to the behavior there in the first place, so, who knows ;).

plugins/statistics.koplugin/main.lua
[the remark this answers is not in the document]
NiLuJe: Ok ;).

I'm happy with whatever you guys actually knowing SQL settle for ;).

So just ping me if something needs to change :).

plugins/statistics.koplugin/main.lua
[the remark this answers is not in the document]
NiLuJe: Oh. :D.

poire-z:
(I hate discussion in the narrow review comments parts, so moving it here)

> It's also stored in the book table in the total_read_time field, and it is also used for other time-based estimates/averages down the line, so, this is its rightful place ;).
Mostly everything still uses sum(duration), this is purely used for averages and estimates.

Note that my concerns are not about what you use for what - just about the DB being clean, sane and not confusing.

So, if I get that right:
- `page_stat_data` is the only place where we still have all the raw pages and durations, including re-reads (but indeed raw data cap'ed to `page_max_read_sec`, and skipped if `duration < page_min_read_sec`, which is fine by me and what I expected initially solely from these 2 settings).
- `book.total_read_time` used to be `sum(page_stat.durations for page in this book)` - but is now a fomula of `sum(the same but cap'ed to page_max_read_sec)`. (the fact the meaning has changed, if anybody is already parsing our stats.sqlite db from outside might be surprising - and the column name is then ambiguous).

So, from a proper DB degisn POV, `book.total_read_time` is redundant : it's a bit of a cache of a pre-computed value, we can always obtain/refresh from `page_stat_data`. Right ?

So, if it's just a cache, why not have both, or none at all? Does one of them deserves to be pre-computed more than the other?

When was/is `book.total_read_time` updated for a book? Is it always recomputed when book opened? Or previous value completed with latest pages collector data?

An issue now with `book.total_read_time`  being dependant on `page_max_read_sec` is that when I change the value of `page_max_read_sec`, `book.total_read_time` will change. But it might not be updated for previous books not re-opened since - but would be when re-opened? So, we might see some read times oscillating without any much coherency.
So, may be not caching it as it's dependant of a setting might be a better idea for things to be sane and not have stalled values?

NiLuJe:
> So, if I get that right:
> 
>     * `page_stat_data` is the only place where we still have all the raw pages and durations, including re-reads (but indeed raw data cap'ed to `page_max_read_sec`, and skipped if `duration < page_min_read_sec`, which is fine by me and what I expected initially solely from these 2 settings).
> 
>     * `book.total_read_time` used to be `sum(page_stat.durations for page in this book)` - but is now a fomula of `sum(the same but cap'ed to page_max_read_sec)`. (the fact the meaning has changed, if anybody is already parsing our stats.sqlite db from outside might be surprising - and the column name is then ambiguous).
>

Yep! 

> So, from a proper DB degisn POV, `book.total_read_time` is redundant : it's a bit of a cache of a pre-computed value, we can always obtain/refresh from `page_stat_data`. Right ?
> 
> So, if it's just a cache, why not have both, or none at all? Does one of them deserves to be pre-computed more than the other?
> 

Hmm, that's an interesting way to look at it ;).

> When was/is `book.total_read_time` updated for a book? Is it always recomputed when book opened? Or previous value completed with latest pages collector data?

It's only updated alongside `page_stat_data` (i.e., in `insertDB`).
It's *queried* on book-open to seed the `self.` vars of the same name (which are also updated in `insertDB`).

Anything that queries the current book does an `insetDB` first, so it's always up to date in the "current book" context.

> 
> An issue now with `book.total_read_time` being dependant on `page_max_read_sec` is that when I change the value of `page_max_read_sec`, `book.total_read_time` will change. But it might not be updated for previous books not re-opened since - but would be when re-opened? So, we might see some read times oscillating without any much coherency.
> So, may be not caching it as it's dependant of a setting might be a better idea for things to be sane and not have stalled values?

Hmm, that's a fair point.

Then again, usage in the "not the current book" case is fairly limited (`getBookStat`, and that's it).

plugins/statistics.koplugin/main.lua
marek-g: Ok. And what about the second part: "And also the change with min(sum()) to sum(min()) with page_max_read_sec - see what's going on if we read page twice (total_pages = 2, read_time = once? - this is wrong)." ?

plugins/statistics.koplugin/main.lua
marek-g: In #6759 the change was to make counting pages in mem_read_pages as count(DISTINCT page), like we counted total_read_pages in database.

Now total_read_pages are counted as count(page), so here mem_read_pages should be counted the same way.

The example: you read the page twice. You will increase mem_read_time. You should also increase the mem_read_pages. Otherwise we are calculation count(DISTINCT page) in memory and count(page) in dabase.

NiLuJe:
Ah, may have figured out why `total_read_time` is cached: it appears to be for `deleteBooksByTotalDuration`...

plugins/statistics.koplugin/main.lua
```diff
@@ -309,11 +309,11 @@ Do you want to create an empty database?
 
         -- Check if we need to migrate to a newer schema
         local db_version = conn:rowexec("PRAGMA user_version;")
```
poire-z, line 311: tonumber() it once, early ?

NiLuJe:
@poire-z: 5e9dad8 should have clarified the `book.total_read_time` situation ;).

plugins/statistics.koplugin/main.lua
marek-g: In #6759 the change was to make counting pages in mem_read_pages as count(DISTINCT page), like we counted total_read_pages in database.

Now total_read_pages are counted as count(page), so here mem_read_pages should be counted the same way.

The example: you read the page twice. You will increase mem_read_time. You should also increase the mem_read_pages. Otherwise we are calculation count(DISTINCT page) in memory and count(page) in dabase.

plugins/statistics.koplugin/main.lua
marek-g: Ok. And what about the second part: "And also the change with min(sum()) to sum(min()) with page_max_read_sec - see what's going on if we read page twice (total_pages = 2, read_time = once? - this is wrong)." ?

marek-g (changes requested):

plugins/statistics.koplugin/main.lua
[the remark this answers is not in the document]
marek-g: Yes. We are using "GROUP  BY id_book, page" in ReaderStatistics:getTodayBookStats(). Let's change the index to:
UNIQUE (id_book, page, start_time) and benefit from it.

marek-g (changes requested):

NiLuJe:
@marek-g:

Re: https://github.com/koreader/koreader/pull/6778#discussion_r502807547

> Now total_read_pages are counted as count(page), so here mem_read_pages should be counted the same way.

It isn't, it's still counted as (DISTINCT page) ;). Hence why this does the same.

plugins/statistics.koplugin/main.lua
```diff
@@ -36,6 +36,28 @@ local DEFAULT_CALENDAR_NB_BOOK_SPANS = 3
 -- Current DB schema version
 local DB_SCHEMA_VERSION = 20201010
 
+-- This is the query used to compute the total time spent reading distinct pages of the book,
+-- capped at self.page_max_read_sec per distinct page.
+-- c.f., comments in insertDB
+local STATISTICS_SQL_BOOK_CAPPED_TOTALS_QUERY = [[
```
poire-z, line 42: May be include the word TIME or DURATION in the name of the variable, because TOTALS what ? :)

NiLuJe:
@marek-g:

Re: https://github.com/koreader/koreader/pull/6778#discussion_r502807661

Oh, that may very well be my own typo somewhere along the way, good catch ;).

plugins/statistics.koplugin/main.lua
NiLuJe: It includes both time & pages, though ;).

poire-z:
> @poire-z: 5e9dad8 should have clarified the book.total_read_time situation ;).

Yep, looks fine, thanks.
(A bit hard to get into what the variables mean, book_read_time vs total_read_time... Your distinction in the menu title, "reading this book" vs "spent on this book" is a lot clearer - dunno if you feel like adjusting the variable names to be as clear as that :) - although one maps to an existing DB column name...)

NiLuJe:
@marek-g;

Re: https://github.com/koreader/koreader/pull/6778#discussion_r502807661

Err, actually, no, that's the whole point: for each distinct page, capping the *sum* at page_max. So that even re-reads don't go over page_max.

poire-z:
(I checked the new small queries timing and plan: all is fine, they are fast.)

NiLuJe:
> > @poire-z: [5e9dad8](https://github.com/koreader/koreader/commit/5e9dad8ea7ea149eb185eabc8f2096a2d7cb2d50) should have clarified the book.total_read_time situation ;).
> 
> Yep, looks fine, thanks.
> (A bit hard to get into what the variables mean, book_read_time vs total_read_time... Your distinction in the menu title, "reading this book" vs "spent on this book" is a lot clearer - dunno if you feel like adjusting the variable names to be as clear as that :) - although one maps to an existing DB column name...)

Yeah, I'm not super happy that book also happens to be a table name while the data isn't actually pulled from that table at all, but it's good enough for me ;p.

plugins/statistics.koplugin/main.lua
marek-g: > It isn't, it's still counted as (DISTINCT page) ;). Hence why this does the same.

Yes, you are right, sorry. Too many similar names 👍 

But forget me that I'll drill it a little bit more :(

`self.avg_time = (self.book_read_time + self.mem_read_time) / (self.book_read_pages + self.mem_read_pages)`

we are adding book_read_pages with mem_read_pages. And book_read_pages is STATISTICS_SQL_BOOK_CAPPED_TOTALS_QUERY which is count(*) !!

plugins/statistics.koplugin/main.lua
marek-g: > It isn't, it's still counted as (DISTINCT page) ;). Hence why this does the same.

Yes, you are right, sorry. Too many similar names 👍

But forget me that I'll drill it a little bit more :(

self.avg_time = (self.book_read_time + self.mem_read_time) / (self.book_read_pages + self.mem_read_pages)

we are adding book_read_pages to mem_read_pages. And book_read_pages is STATISTICS_SQL_BOOK_CAPPED_TOTALS_QUERY which is count(*) !!

marek-g: > It isn't, it's still counted as (DISTINCT page) ;). Hence why this does the same.

Yes, you are right, sorry. Too many similar names 👍

But forget me that I'll drill it a little bit more :(

self.avg_time = (self.book_read_time + self.mem_read_time) / (self.book_read_pages + self.mem_read_pages)

we are adding book_read_pages to mem_read_pages. And book_read_pages is STATISTICS_SQL_BOOK_CAPPED_TOTALS_QUERY which is count(*) !!

marek-g (changes requested):

NiLuJe:
@marek-g: Re: https://github.com/koreader/koreader/pull/6778#discussion_r502811350

It's `count(*)` ... against a `GROUP BY page`, which ends up being the same as `count(DISTINCT page)` ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -328,34 +464,67 @@ function ReaderStatistics:createDB(conn)
...
-        CREATE INDEX IF NOT EXISTS page_stat_id_book ON page_stat(id_book);
+    ]]
+    conn:exec(sql_stmt)
+
+    -- page_stat_data
+    conn:exec(STATISTICS_DB_PAGE_STAT_DATA_SCHEMA)
+
+    sql_stmt = [[
+        -- Indexes
+        CREATE INDEX IF NOT EXISTS page_stat_data_id_book ON page_stat_data(id_book);
```
poire-z, line 475: No more necessary now that we have `UNIQUE (id_book, page, start_time),` (which creates implicitely an index - more precise but as usable as this one).
@marek-g : do you agree?

marek-g:
> It's count(*) ... against a GROUP BY page SELECT, which ends up being the same as count(DISTINCT page) ;).

Ahhh. Stupid me!! Thanks @NiLuJe for your patience!!! You are right!!!! Sorry.

plugins/statistics.koplugin/main.lua
marek-g: My fault. That's ok!

plugins/statistics.koplugin/main.lua
NiLuJe: That was indeed going to be my next question ^^

NiLuJe:
@marek-g: Eh, don't worry, that's the beauty of code reviews ;).

plugins/statistics.koplugin/main.lua
marek-g: I think you are right, but it would be good to verify it with execution plan to ensure that SQLite will use the unique index.

marek-g: Let me test it.

marek-g: Yes. The `page_stat_data_id_book` is not needed. SQLite will happily use the UNIQUE index when quering by id_book.

marek-g:
I think I'm happy with the changes :) Thanks @NiLuJe.

poire-z:
Looks fine (not re-read, but I applied the latest commits in my head :)

OK, quickly tested on my Kobo.
DB migrated correctly.
When running sqlite3 ".schema", I see these little unexpected thingies:
`CREATE TABLE sqlite_sequence(name,seq);` (but it was there on the old version, so ok - I still don't know why it shows in the user .schema)
and this at the end of the CREATE VIEW:
`/* page_stat(id_book,page,start_time,duration) */;`
which I don't know where it comes from.

Comparing my current book stats:
Before:
<kbd>image</kbd>
After:
<kbd>image</kbd>

I don't care much about the values - just showing this to you so you can tell me it's the expected corrections :)

(CalendarView looks unchanged, so good!)

marek-g:
@poire-z The statistics AFTER make sense. The statistics BEFORE are inconsitent (estimated time to read too low? average time per page too low?). So it is an improvement even for old books.

NiLuJe:
@poire-z:

I noticed that, too. `sqlite_sequence` is an internal table needed to keep track of the `autoincrement` on book id's.

I'm not *quite* sure why it may or may not have bene visible in the .schema output before, but it's definitely supposed to be there ^^.

I assume the comment is also added by SQLite to recap the view's layout.

(SQLite's schema handling is peculiar: it's stored in plain-text, so any modifications to it are also essentallty string manipulations).

poire-z:
> The statistics BEFORE are inconsitent (estimated time to read too low? average time per page too low?). 

OK, but was it expected/observed they were inconsistent that way ? :)

NiLuJe:
I mean, the results looks like stuff is at least working as it should *now* ;o).

marek-g:
> OK, but was it expected/observed they were inconsistent that way ? :)

I don't know. I haven't been using it long enough to tell.

marek-g:
I've run the code. Migration went ok. The behavior of statistics when reading also looks ok.

NiLuJe:
Sidebar: 5e4cb4a now requires https://github.com/koreader/koreader-base/pull/1212

poire-z:
Dunno if it's my own patch messing with this PR (PR up to but without last commit), but I got no new data in page_stat_data/page_stat. My stats are still just as in the screenshot above :) Opening a new book gets it inserted in the book table, but no page with that id_book goes into page_stat_data.
Can you check that all still works fine on your side ? And I'll investigate why it does not on mine tomorrow :)
(Might depend on your other stuff in https://github.com/koreader/koreader-base/pull/1212#partial-timeline that I haven't installed ?)

NiLuJe:
@poire-z: Yup, works for me.

```
SELECT * FROM page_stat_data WHERE id_book = 43 ORDER BY start_time;

...
43|8|1601909189|40|303
43|8|1602374583|23|303
43|9|1602374606|12|303
```

Last two are right now ;).

NiLuJe:
Will get back to you with actual real reading data from the Forma tomorrow, but the quick glance I took at it earlier seemed fine, too (on 2020.10 + both PRs).

NiLuJe:
Fun fact I just realized: the `-readonly` switch on the CLI tool on WAL databases means that it doesn't necessarily COMMIT on exit.

If you have a RO shell holding a lock on the DB, shit may go weird ^^.

poire-z:
Browsed a few page this morning, and I got some data saved, for both books:
```
1153        18          1602331745  23          197
1153        19          1602331768  56          197
1153        20          1602331824  240         197
1153        20          1602353607  22          197
1153        20          1602353629  22          197 yesterday 20h14
    black hole
1154        9           1602401075  71          224 this morning
1154        11          1602401150  76          224
1154        12          1602401226  30          224
1153        30          1602401346  25          197
1153        31          1602401371  51          197
1153        32          1602401422  28          197
1153        33          1602401450  23          197
1153        30          1602401495  76          197
1153        29          1602401571  21          197
```
I might have not spend 2 pages before viewing the stats on that new book (1154), but I did read 10 pages of 1153 withouht caring about stats last evening.

Going to install latest nightly + the 2 PRs, and restore backup'ed db.

poire-z:
No issue this morning:
```
1153        20          1602331824  240         197
1153        20          1602353607  22          197
1153        20          1602353629  22          197

1153        30          1602403375  24          197
1153        30          1602403399  240         197
1153        30          1602404338  20          197
1153        29          1602404358  26          197
1153        30          1602404384  70          197
1153        31          1602404454  95          197
1153        32          1602404549  75          197
1153        33          1602404646  65          197
1153        34          1602404717  240         197
1153        34          1602405514  115         197
1153        35          1602405629  52          197
1153        36          1602405681  80          197
1153        37          1602405761  109         197
1154        6           1602405928  47          224
1154        7           1602405975  35          224
1154        8           1602406010  33          224
```
No idea what did happen last evening... (I did not play with sqlite CLI on the device.)

NiLuJe:
No issues on my end on the Forma, FWIW.

NiLuJe:
The only thing I can think of is an extremely unlucky combination of events that call insertDB -> resetVolatileStats while being under the 2 pages read mark.

Notably, suspend/resume or layout changes can do that.

NiLuJe:
So, @poire-z, I kinda have a feeling I already know your answer, but, what would you say about getting rid of those `MIN_PAGES_READ_TO_FLUSH` checks? ^^

poire-z:
> I kinda have a feeling I already know your answer (^^), but, what would you say about getting rid of those MIN_PAGES_READ_TO_FLUSH checks?

I'd say I don't care much, as I don't really know why it was there :)  but it being `=2`, I'd think it shouldn't hurt either way.
(Does that match your expected/guessed answer ? :)

marek-g:
I just quickly tested (on emulator) that removing MIN_PAGES_READ_TO_FLUSH doesn't slow down emulator when changing the font size, so it's a good move.

I didn't play too much with it, but works great for me so far. It is fantastic, when I change font size and book status is correctly recalculated. Great job!

poire-z:
Not working in this morning Kobo nightly, see : https://github.com/koreader/koreader-base/pull/1212#issuecomment-708950220

NiLuJe:
Gah, I knew I'd forgotten to check something :/.

Completely skipped my mind >_<".

poire-z:
I've been having some traceback on the emulator for some days `common/lua-ljsqlite3/init.lua:58: in function 'err'` and didn't care much, assuming it was some inconsistent state since we played with this.

So, I restored the backup of statistics.sqlite3 and let it migrate properly... but:
```
10/19/20-12:37:46 INFO  ReaderStatistics: Migrating DB from schema 0 to schema 20201010 ...
10/19/20-12:37:46 INFO  ReaderStatistics: Old DB backed up as ./settings/statistics.sqlite3.bkp.0-to-20201010
10/19/20-12:37:47 ERROR Failed to initialize statistics plugin:  common/lua-ljsqlite3/init.lua:58: ljsqlite3[constraint] NOT NULL constraint failed: page_stat_data.total_pages
stack traceback:
        common/lua-ljsqlite3/init.lua:58: in function 'err'
        common/lua-ljsqlite3/init.lua:233: in function 'E_conn'
        common/lua-ljsqlite3/init.lua:603: in function '_step'
        common/lua-ljsqlite3/init.lua:619: in function 'resultset'
        common/lua-ljsqlite3/init.lua:384: in function 'exec'
        plugins/statistics.koplugin/main.lua:500: in function 'upgradeDB'
        plugins/statistics.koplugin/main.lua:314: in function 'checkInitDatabase'
        plugins/statistics.koplugin/main.lua:150: in function 'init'
```
and after this, ignoring it and relaunching KOReader, I get again what I had before:
```
10/19/20-12:44:22 ERROR Failed to initialize statistics plugin:  common/lua-ljsqlite3/init.lua:58: ljsqlite3[error] no such column: "period"
stack traceback:
        common/lua-ljsqlite3/init.lua:58: in function 'err'
        common/lua-ljsqlite3/init.lua:233: in function 'E_conn'
        common/lua-ljsqlite3/init.lua:239: in function 'T_okcode'
        common/lua-ljsqlite3/init.lua:371: in function 'prepare'
        common/lua-ljsqlite3/init.lua:383: in function 'exec'
        plugins/statistics.koplugin/main.lua:481: in function 'upgradeDB'
        plugins/statistics.koplugin/main.lua:314: in function 'checkInitDatabase'
        plugins/statistics.koplugin/main.lua:150: in function 'init'
```
because of a half-migration and the DB being inconsistent.

Anyway, the reason for the original faillure was that I have in the old `page_stat` pages for books that I explicitely DELETEd from `book` before:
```
sqlite> SELECT id_book, page, start_time, period, pages as total_pages FROM page_stat LEFT JOIN book on book.id = id_book where pages is NULL;
id_book     page        start_time  period      total_pages
----------  ----------  ----------  ----------  -----------
1219        1           1586890414  5           NulL
1275        1           1589355537  6           NulL
1275        36          1589355545  8           NulL
```
and I have in my .sqlite_history:
```
delete from book where id>=1219;
delete from book where id=1275
```

So, I guess it's my fault and we might not really need to bother :)

But the little annoying thing is that, with these 2 half-migrations, statistics.sqlite3.bkp.0-to-20201010 is no more the original file before the first migration: it's a half migrated mess :)
(Fortunatly, I had a manual backup of it - so I cleaned things up with `delete from page_stat where id_book in (1219, 1275)` and the migration succeeded.)

So, may be we could arrange things a bit (pragma user_version early ?) so we don't go into this situation when things are a bit user-messy ?

NiLuJe:
@poire-z: The backup is definitely done before any kind of migration is attempted. Which means either it was already broken before the backup, or it wasn't closed properly before that.

poire-z:
Yes, the first time all was fine: the original db was copied.

But the first migration failed with (because of my inconsistent data):
`NOT NULL constraint failed: page_stat_data.total_pages`
Before this failure, it had the time to do and commit:
```
ALTER TABLE page_stat RENAME COLUMN period TO duration;
DROP TABLE info;
CREATE TABLE IF NOT EXISTS page_stat_data
```
and failed before doing:
```
DROP INDEX IF EXISTS page_stat_id_book
DROP TABLE IF EXISTS page_stat
STATISTICS_DB_PAGE_STAT_VIEW_SCHEMA
PRAGMA user_version=%d
```

So, on next KOReader launch, because the user_version is still 0, it redo:
`FFIUtil.copyFile(db_location, bkp_db_location)`
After this, the original copy/backup of the db (with my inconsistent data) is lost: the user is stuck with a DB with an inconsistent schema and stats failing.

NiLuJe:
Oh, right, probably ought not to stomp on an existing backup, true ;).

poire-z:
And may be replace in (which looks like the only thing that could fail):
```
        INSERT INTO page_stat_data
            SELECT id_book, page, start_time, duration, pages as total_pages FROM page_stat
            LEFT JOIN book on book.id = id_book;
```
either use `IFNULL(pages, some_default_value) as total_pages ` (but which default value ? 0 might cause a division by 0)
or remove the `LEFT` from `LEFT JOIN` so we don't migrate pages for books that are no more in the book table.

NiLuJe:
Apparently SQLite simply returns `NULL` for a division by zero (and `total_pages` already has a `DEFAULT 0` constraint), so that shouldn't be an issue down the line, IFNULL should do.

plugins/statistics.koplugin/main.lua
```diff
@@ -175,14 +200,44 @@ function ReaderStatistics:initData()
...
+        logger.info("ReaderStatistics: Initializing average time per page at 50% of the max value, i.e.,", self.avg_time)
+    end
+end
+
+-- Reset the (volatile) stats on page count changes (e.g., after a font size update)
+function ReaderStatistics:onUpdateToc()
+    local new_pagecount = self.view.document:getPageCount()
+
+    if new_pagecount ~= self.data.pages then
+        logger.info("ReaderStatistics: Pagecount change, flushing volatile book statistics")
```
poire-z, line 219: Do these 2 needs to be logger.info() ? I was surprised to see them in my log on my Kobo and wondered what to make out of that info :)

NiLuJe: I think there may be a few other instances of those in there that might be better classified as something other than info, yeah.

Certainly, the second one here. I could argue that the first one is mildly informative, but then again the fact that it'll show on every new book makes it fairly verbose.

poire-z:
There is a small issue since (and possibly before) this PR:
- we update the duration in `self.page_stat[cur_page] = { { now_ts, 0 } }` only when going to the next page when `onPageUpdate(next pageno)` is called.
- when flushing data, when onSuspend or when showing some statistics page, we have :insertDB() called, which will NOT insert a page_stat tuple that has a duration of 0. After this is done, we reset page_stat to empty - possibly in some case adding a new `self.page_stat[cur_page] = { { an_other_now_ts, 0 } }` entry.

So, if you have spent 90s on reading a page - and you suspend or show calendar view when at the bottom of the page - then you resume or close the statistic info page, and in the next few seconds you tap to go to next page, this page is not accounted as read (and so the stats may get a bit wrong if you suspend often).
Been noticing some holes in my #8426 experiments :):
image

Not sure what's the right way to handle that:
- update in insertDB() `self.page_stat[cur_page] = { { an_other_now_ts, 0 } }` with the current time, so at least the first part of the time spent is saved if > min_read_secs
- don't insert anything for the current page, but keep a backup of it (with `{start_now_ts, 0}`) and restore it as-is in resetVolatileStats() - so it will be updated on the next onPageUpdate() with the full duration since its initial now_ts ?
- other ideas ?

poire-z:
On the same point, there's also the case that I sometime, when about to start reading the last paragraph of a page, I tap to see how long it is and if it ends or not on the next page - to see if I can read it for a minute or so, or if I better stop there to keep the longer paragraph for a future session.
If I do that too early (before my min_read_secs of 20s), these 19s I have spend on that page are then ignored - and if I take another 19s to read the 2nd part of the page, I have my 48s spent reading this page not accounted.
So, since I noticed that, I'm a bit hesitating looking ahead :)

NiLuJe:
I'm into my second 58-hour week, so take that with a grian of salt, as my brain is pretty much shot, but:

>     * don't insert anything for the current page, but keep a backup of it (with `{start_now_ts, 0}`) and restore it as-is in resetVolatileStats() - so it will be updated on the next onPageUpdate() with the full duration since its initial now_ts ?

That sounds feasible and mostly coherent with the current design (or what I remember of it, anyway ;p).

NiLuJe:
> On the same point, there's also the case that I sometime, when about to start reading the last paragraph of a page, I tap to see how long it is and if it ends or not on the next page - to see if I can read it for a minute or so, or if I better stop there to keep the longer paragraph for a future session. If I do that too early (before my min_read_secs of 20s), these 19s I have spend on that page are then ignored - and if I take another 19s to read the 2nd part of the page, I have my 48s spent reading this page not accounted. So, since I noticed that, I'm a bit hesitating looking ahead :)

Tap where/when/how?

(Feels intractable, though ^^)

poire-z:
Something strange that I had noticed since I've been using Book map: when you change font size or anything that rerenders and changes the page numbers, it happens that I see in Book map some holes (which might be ok if some small-duration is split/aggragated in some way), but also some page slots for the 4 or 5 next pages FILLED, as if I have read this part of the book, but I have not.
(I haven't investigated, just mentionning it if someone feels like having a look.)

poire-z:
^ I guess I'll consider this a non issue, or at least, explainable.
Extreme case of switching font size from 24 to 44 on a book of 151 pages at 24 becoming 479 pages at 44.
Before:
```
sqlite> select pages from book where id=1779;
pages
-----
151

sqlite> select * from page_stat_data where id_book=1779;
id_book  page  start_time  duration  total_pages
-------  ----  ----------  --------  -----------
1779     126   1643305069  13        151
1779     127   1643305082  21        151
1779     127   1643305104  5         151

sqlite> select * from page_stat where id_book=1779;-- that is the dynamic VIEW
id_book  page  start_time  duration
-------  ----  ----------  --------
1779     126   1643305069  13
1779     127   1643305082  21
1779     127   1643305104  5
```
After:
```
sqlite> select pages from book where id=1779;
pages
-----
479

sqlite> select * from page_stat_data where id_book=1779;
id_book  page  start_time  duration  total_pages
-------  ----  ----------  --------  -----------
1779     126   1643305069  13        151
1779     127   1643305082  21        151
1779     127   1643305104  5         151
1779     127   1643305358  6         151

sqlite> select * from page_stat where id_book=1779; -- that is the dynamic VIEW
id_book  page  start_time  duration
-------  ----  ----------  --------
1779     397   1643305069  4
1779     398   1643305069  4
1779     399   1643305069  4
1779     400   1643305082  7
1779     401   1643305082  7
1779     402   1643305082  7
1779     400   1643305104  1
1779     401   1643305104  1
1779     402   1643305104  1
1779     400   1643305358  2
1779     401   1643305358  2
1779     402   1643305358  2
```
There is also the fact that CRE repositions after re-rendering the book to the page containing the xpointer of the top of page before re-rendering: when switching font size, I ended up on page 395 (with 0 duration, page unread), and I'm indeed seeing in Book map pages 395 and 396 unread, and 397 to 402 read:
image
And it's possible that the content I was seeing on pages 126 and 127 may actually now spans pages 395 to 400 (actually, the last word on page 127 is really on the first line of page 402!).

It's a bit ridiculous :/ but I guess that with storing (pagenum, totalpages) and scaling that to a new totalpages, we can't do better...

Issue #6768, opened 2020-10-08
[RFC] On the whole user-plugin thingy...


NiLuJe:
I was thinking of going with a `contrib` folder in the repo, which could include a `plugins` folder for, well, plugins, or platform folders for platform-specific scripts (e.g., https://github.com/koreader/koreader/issues/6399#issuecomment-681026114).

The code could either be in-repo, or added as submodules?

This would allow PRs to simply be steered toward that after initial review (i.e., either ""merge it", "drop it", or "move it to contrib").

Then you just document that globally and/or individually somewhere/somehow in docs and/or wiki.

Thoughts?

ezdiy:
I'd go even further, and just be a git URL registry the same way luarocks is. Red tape is what @smartscripts-nl perhaps felt a little discouraged by, and we should do our best to encourage people just doing stuff by removing maximum amount of it.

Ideally:

* User writes a plugin
* Adds his repo url into koreader *somehow*. Perhaps just opening PR "heres my repo, add it".  
Perhaps go even a lil bit riskey, and just parse it from "Plugin Registry" wiki page on build, making the stuff completely unattended.
* We don't care what the plugin does, editorial role is reduced to checking "its not malicious / likely not to damage devices".  
  Or in case of the bolder, unattended option there's implicit assumption there's a degree of accountability to edit github wikis.
* Hide those plugins behind - but visible - checkbox, first on plugin menu. On enable, flash infobox that there's absolutely zero QA whatsoever, and all issues should be directed at owner repo of any given plugin.

As for nuts and bolts, just fetching the wiki page and pulling git urls in a makefile from there should be enough. Later on somebody might come by and make it an online fetch/update feature, but given the current interest that's overkill.

pazos:
> User writes a plugin

self contained plugins are somehow limited:

- translations: they don't receive community translations from weblate, they need to provide their own and currently there's no mechanism to translate based on a different po file.

- integration with gestures

- platform limitations (at least **Android**): there's no way to install an user plugin in android on demand. Creating a new folder in internal files dir will crash the Java VM with err=13.  A good workaround is keeping user plugins in dataDir and add that to path and cpath. I guess we can do that for all platforms and keep bundled and user plugins separated.

- optional: put plugin files in a container (a zip file?) with a different mimetype and handle installation from the program.

ezdiy:
[...]

PR #6785, opened 2020-10-12
Dispatcher: allow registering actions at runtime

yparitcher:
as discussed in #6768 and elsewhere.

Dispatcher was updated as not to crash on not existent entries.

I extended the Hello World Plugin to serve as an example.

the important parts.

the plugin must define a `DispatcherRegisterActions` event like so, and call it on init in case the plugin is loaded after Dispatcher:
```lua
function Hello:onDispatcherRegisterActions()
    Dispatcher:RegisterAction("helloworld_action", {category="none", event="HelloWorld", title=_("Hello World"), filemanager=true,})
end
```

If there is a non registered item attached to a gesture/profile (someone enabled a gesture and then disabled the appropriate plugin). then the menu will keep that item, but it will not show in any of the sections as there is no corresponding entry. However the `None` box will **not** be checked so one will realize that something is different. Also one can at any time click on `None` and clear out the gesture/profile.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6785)

### DIFF
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -1,5 +1,28 @@
 --[[--
-    Dispatcher module
+This module is responsible for dispatching events.
+
+To add a new action an entry must be added to `settingsList` & `dispatcher_menu_order`
+This can also be done at runtime via @{registerAction}().
+
+`settingsList` contains the list of dispatchable settings.
+
+Each setting contains:
+
+* category: one of:
+    * none: a direct event call
+    * arg: a event that expects a gesture object or an argument
+    * absolutenumber: event that sets a number
+    * incrementalnumber: event that increments a number & accepts a gesture object
+    * string: event with a list of arguments to chose from
+* event: what to call.
+* title: for use in ui.
+* section: under which menu to display (currently: device, filemanager, rolling, paging)
+    and optionally
+* min/max: for number
+* default
+* args: allowed values for string.
+* toggle: display name for args
+* separator: put a separator after in the menu list
 --]]--
 
 local CreOptions = require("ui/data/creoptions")
@@ -15,25 +38,7 @@ local Dispatcher = {
     initialized = false,
 }
 
---[[--
-contains a list of a dispatchable settings
-each setting contains:
-    category: one of
-       none: a direct event call
-       arg: a event that expects a gesture object or an argument
-       absolutenumber: event that sets a number
-       incrementalnumber: event that increments a number & accepts a gesture object
-       string: event with a list of arguments to chose from
-    event: what to call.
-    title: for use in ui.
-    section: under which menu to display (currently: device, filemanager, rolling, paging)
-and optionally
-    min/max: for number
-    default
-    args: allowed values for string.
-    toggle: display name for args
-    separator: put a separator after in the menu list
---]]--
+-- See above for description.
 local settingsList = {
     -- Device settings
     show_frontlight_dialog = { category="none", event="ShowFlDialog", title=_("Show frontlight dialog"), device=true, condition=Device:hasFrontlight(),},
@@ -322,10 +327,20 @@ function Dispatcher:init()
 end
 
 --[[--
-    add settings at runtime
-    @param name: the key to use in the table
-    @param value: a table per settingsList above.
-    see helloworld plugin for an example.
+Adds settings at runtime.
+
+@usage
+    function Hello:onDispatcherRegisterActions()
+        Dispatcher:registerAction("helloworld_action", {category="none", event="HelloWorld", title=_("Hello World"), filemanager=true,})
+    end
+
+    function Hello:init()
+        self:onDispatcherRegisterActions()
+    end
+
+
+@param name the key to use in the table
+@param value a table per settingsList above
 --]]--
 function Dispatcher:registerAction(name, value)
     if settingsList[name] == nil then
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -1,3 +1,7 @@
+--[[--
+    Dispatcher module
+--]]--
+
 local CreOptions = require("ui/data/creoptions")
 local Device = require("device")
 local Event = require("ui/event")
@@ -73,12 +77,6 @@ local settingsList = {
     toggle_flash_on_chapter_boundaries = { category="none", event="ToggleFlashOnChapterBoundaries", title=_("Toggle flashing on chapter boundaries"), device=true, condition=Device:hasEinkScreen(),},
     set_no_flash_on_second_chapter_page = { category="string", event="SetNoFlashOnSecondChapterPage", title=_("Never flash on chapter's 2nd page"), device=true, condition=Device:hasEinkScreen(), args={true, false}, toggle={_("On"), _("Off")},},
     toggle_no_flash_on_second_chapter_page = { category="none", event="ToggleNoFlashOnSecondChapterPage", title=_("Toggle flashing on chapter's 2nd page"), device=true, condition=Device:hasEinkScreen(), separator=true,},
-    wallabag_download = { category="none", event="SynchronizeWallabag", title=_("Wallabag retrieval"), device=true,},
-    calibre_search = { category="none", event="CalibreSearch", title=_("Search in calibre metadata"), device=true,},
-    calibre_browse_tags = { category="none", event="CalibreBrowseTags", title=_("Browse all calibre tags"), device=true,},
-    calibre_browse_series = { category="none", event="CalibreBrowseSeries", title=_("Browse all calibre series"), device=true, separator=true,},
-    show_terminal = { category = "none", event = "TerminalStart", title = _("Show terminal"), device = true, },
-    edit_last_edited_file = { category = "none", event = "OpenLastEditedFile", title = _("Texteditor: open last file"), device = true, separator = true, },
     favorites = { category="none", event="ShowColl", arg="favorites", title=_("Favorites"), device=true,},
 
     -- filemanager settings
@@ -113,8 +111,6 @@ local settingsList = {
     toggle_inverse_reading_order = { category="none", event="ToggleReadingOrder", title=_("Toggle page turn direction"), rolling=true, paging=true,},
     cycle_highlight_action = { category="none", event="CycleHighlightAction", title=_("Cycle highlight action"), rolling=true, paging=true,},
     cycle_highlight_style = { category="none", event="CycleHighlightStyle", title=_("Cycle highlight style"), rolling=true, paging=true,},
-    kosync_push_progress = { category="none", event="KOSyncPushProgress", title=_("Push progress from this device"), rolling=true, paging=true,},
-    kosync_pull_progress = { category="none", event="KOSyncPullProgress", title=_("Pull progress from other devices"), rolling=true, paging=true, separator=true,},
     page_jmp = { category="absolutenumber", event="GotoViewRel", min=-100, max=100, title=_("Go %1 pages"), rolling=true, paging=true,},
     panel_zoom_toggle = { category="none", event="TogglePanelZoomSetting", title=_("Toggle panel zoom"), paging=true, separator=true,},
 
@@ -205,14 +201,6 @@ local dispatcher_menu_order = {
     "wifi_off",
     "toggle_wifi",
 
-    "wallabag_download",
-    "calibre_search",
-    "calibre_browse_tags",
-    "calibre_browse_series",
-
-    "show_terminal",
-    "edit_last_edited_file",
-
     "rotation_mode",
 
     -- filemanager
@@ -262,9 +250,6 @@ local dispatcher_menu_order = {
     "cycle_highlight_action",
     "cycle_highlight_style",
     "panel_zoom_toggle",
-
-    "kosync_push_progress",
-    "kosync_pull_progress",
 
     "visible_pages",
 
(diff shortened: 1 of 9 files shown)


yparitcher:
@Frenzie I would like to have `Dispatcher` come up in the docs. What do I have to do to get it working to use the function comment etc?

Frenzie:
Very sparse on time for the next couple of weeks, but see something like gettext for pretty much all the tricks in use. The basics are you need enough dashes, typically done as ---- for single line and --[-- for multiline. (Also you can just run ldoc locally to check, piece of cake. ;-) )

frontend/dispatcher.lua
```diff
@@ -1,3 +1,7 @@
+--[[--
+    Dispatcher module
```
Frenzie, line 2: Don't indent this.

frontend/dispatcher.lua
```diff
@@ -332,11 +336,29 @@ function Dispatcher:init()
     for i=1,#CreOptions do
         parseoptions(CreOptions, i)
     end
+    UIManager:broadcastEvent(Event:new("DispatcherRegisterActions"))
     Dispatcher.initialized = true
 end
 
+--[[--
+    add settings at runtime
```
Frenzie, line 344: Adds settings at runtime.

frontend/dispatcher.lua
```diff
@@ -332,11 +336,29 @@ function Dispatcher:init()
...
+    @param name: the key to use in the table
+    @param value: a table per settingsList above.
+    see helloworld plugin for an example.
```
Frenzie, line 347: I believe this should go above the params separated by newlines. Also you can/should include a link.

frontend/dispatcher.lua
```diff
@@ -332,11 +336,29 @@ function Dispatcher:init()
...
+--]]--
+function Dispatcher:RegisterAction(name, value)
+    if settingsList[name] == nil then
+        settingsList[name] = value
+        table.insert(dispatcher_menu_order, name)
+    end
+    return true
```
Galunid, line 354: `return true` is probably not needed, since nothing really uses it

poire-z:
Looks good and simple to me.
Pinging @smartscripts-nl , if you can test this and integrate it in some of your user plugins, and tell us if that works for your use cases.

smartscripts-nl:
> Looks good and simple to me.
> Pinging @smartscripts-nl , if you can test this and integrate it in some of your user plugins, and tell us if that works for your use cases.

Works great! It's nice that I don't have to edit dispatcher.lua to register my custom actions. Also nice that these custom actions can always be found at the end of the actions list, when I want to assign them to a gesture.

smartscripts-nl:
I've use this to add ca. 40 custom actions to the dispatcher and didn't encounter any problem.

yparitcher:
@smartscripts-nl 
40 :)
You might want to try adding a gesture and then disabling the plugin to realy put this through its corner cases.
Thanks for testing

yparitcher:
The names of the function were not very thought out, if anyone has better ideas please let me know

poire-z:
> The names of the function were not very thought out, if anyone has better ideas please let me know

Not really.
Just a convention that method names start with a lowercase, so may be `Dispatcher:registerAction` instead of `Dispatcher:RegisterAction`.
```
$ rgrep ':[A-Z]' frontend/|wc -l
35
$ rgrep ':[a-z]' frontend/|wc -l
13795
```
The event and target functions name is a bit ugly `:onDispatcherRegisterActions()` - but it maps nicely to what is to be done :) : calling `Dispatcher:RegisterAction()`

smartscripts-nl:
> @smartscripts-nl
> 40 :)
> You might want to try adding a gesture and then disabling the plugin to realy put this through its corner cases.
> Thanks for testing

Disabling a plugin that registers custom actions causes no problems.

NiLuJe:
About the names: Yeah, what @poire-z said, basically. No better idea comes to mind ;).

frontend/dispatcher.lua
```diff
@@ -1,3 +1,30 @@
+--[[--
+Dispatcher module
+
+To add a new action an entry must be added to `settingsList` & `dispatcher_menu_order`
+This can also be done at runtime via `registerAction`
```
Frenzie, line 5: You can make that a direct reference btw, something like `{@registerAction}` otoh, or with a pipe if you want to customize the text.

yparitcher: right now it is a direct link.
 {@registerAction} doesn't do anything

Frenzie: Um, it most certainly does. :-P

frontend/dispatcher.lua
```diff
@@ -1,3 +1,30 @@
...
+    Dispatcher:registerAction("helloworld_action", {category="none", event="HelloWorld", title=_("Hello World"), filemanager=true,})
```
Frenzie, line 6: Something more like this (a description, "Dispatcher module" is semi-meaningless):
```suggestion
--[[--
This module is responsible for dispatching events.

To add a new action an entry must be added to `settingsList` & `dispatcher_menu_order`
This can also be done at runtime via @{registerAction}().

    Dispatcher:registerAction("helloworld_action", {category="none", event="HelloWorld", title=_("Hello World"), filemanager=true,})

`settingsList` contains the list of dispatchable settings. Each setting contains:
```

Frenzie: The newlines are for us; they don't affect rendering.

frontend/dispatcher.lua
```diff
@@ -332,11 +341,30 @@ function Dispatcher:init()
...
+    UIManager:broadcastEvent(Event:new("DispatcherRegisterActions"))
     Dispatcher.initialized = true
 end
 
+--[[--
+    Adds settings at runtime.
+    see helloworld plugin for an example.
+
+    @param name: the key to use in the table
+    @param value: a table per settingsList above.
```
Frenzie, line 353:
```suggestion
Adds settings at runtime.

See the helloworld plugin for an example.

@param name: the key to use in the table
@param value: a table per settingsList above.
```

Frenzie: You can btw add it to the config, with `examples = {'examples', exclude = {'examples/slow.lua'}}`, and link to it.

frontend/dispatcher.lua
yparitcher: I am using ldoc 1.4.6 from archlinux with lua 5.4

yparitcher: your comment here has a typo `{@}` further down you wrote the correct way `@{}` It is working now. :)

pazos:
@yparitcher: Is it possible to move plugin actions already included in the dispatcher to their relative plugins?. I'm talking about https://github.com/koreader/koreader/blob/master/frontend/dispatcher.lua#L76-L81 and https://github.com/koreader/koreader/blob/master/frontend/dispatcher.lua#L116-L117 and their entries in `dispatcher_menu_order` array

yparitcher:
> @yparitcher: Is it possible to move plugin actions already included in the dispatcher to their relative plugins?. I'm talking about https://github.com/koreader/koreader/blob/master/frontend/dispatcher.lua#L76-L81 and https://github.com/koreader/koreader/blob/master/frontend/dispatcher.lua#L116-L117 and their entries in `dispatcher_menu_order` array

Yes.
Thanks for pointing this out.

frontend/dispatcher.lua
yparitcher: > You can btw add it to the config, with `examples = {'examples', exclude = {'examples/slow.lua'}}`, and link to it with `@see`.

?
I don't get this.
I am trying to use the docs from https://stevedonovan.github.io/ldoc/manual/doc.md.html Do you have a better reference  for me to use?
Thanks for your help with this.

Frenzie: I copied that from there. ;-) You don't have to leave it as an exercise to the reader to find the hello world plugin. At its simplest by linking to it on GH, but I was saying you could experiment with that too if you want.

yparitcher:
@pazos done.

frontend/dispatcher.lua
Frenzie: Then you can link to it with `@script` or something, I haven't really used that functionality yet.

plugins/wallabag.koplugin/main.lua
```diff
@@ -38,6 +39,9 @@ local Wallabag = WidgetContainer:new{
     name = "wallabag",
 }
 
+function Wallabag:onDispatcherRegisterActions()
+    Dispatcher:registerAction("wallabag_download", { category="none", event="SynchronizeWallabag", title=_("Wallabag retrieval"), device=true,})
+end
```
Frenzie, line 44: nit
```suggestion
end

```

frontend/dispatcher.lua
yparitcher: fixed

yparitcher:
I am ready to merge this, tested by me & @smartscripts-nl.
If anyone has any more reviews or comments, please post.

poire-z (approved):

Frenzie:
Please wait a couple of days to merge.

frontend/dispatcher.lua
```diff
@@ -1,3 +1,29 @@
+--[[--
+This module is responsible for dispatching events.
+
+To add a new action an entry must be added to `settingsList` & `dispatcher_menu_order`
+This can also be done at runtime via @{registerAction}.
```
Frenzie, line 5: I prefer it like this.
```suggestion
This can also be done at runtime via @{registerAction}().
```

frontend/dispatcher.lua
```diff
@@ -332,11 +321,42 @@ function Dispatcher:init()
...
+        Dispatcher:registerAction("helloworld_action", {category="none", event="HelloWorld", title=_("Hello World"), filemanager=true,})
+    end
+
+    function Hello:init()
+        self:onDispatcherRegisterActions()
+    end
+
+
+@param name
+the key to use in the table
```
Frenzie, line 342: Not strictly necessary, but I find this multiline thing a bit hard to read. Same for the rest.
```suggestion
@param name the key to use in the table
```

plugins/hello.koplugin/main.lua
```diff
@@ -1,9 +1,15 @@
+--[[--
+This is a debug plugin to test Plugin functionality
+@module koplugin.HelloWorld
```
Frenzie, line 3: Reverse of the other comment, makes it seem like it's part of it when it's not.
```suggestion
This is a debug plugin to test Plugin functionality

@module koplugin.HelloWorld
```

yparitcher:
The only downside is that if you set a gesture/profile to one of these actions and disable the plugin you menu will not have anything checked which is a little confusing (but this means that when you reenable the plugin the profile will have the correct action). and the user can press `Nothing` and they will be deleted correctly. It will look like this.

Instead of:

frontend/dispatcher.lua
```diff
@@ -332,11 +321,42 @@ function Dispatcher:init()
...
+
+@param value
+a table per settingsList above.
```
Frenzie, line 345: Also stray period.
```suggestion
@param value a table per settingsList above
```

Frenzie (approved):
lgtm besides the doc nits

ezdiy:
Looks far less horrible to me :)

plugins/calibre.koplugin/metadata.lua
```diff
@@ -1,10 +1,11 @@
---[[
+--[[--
     This module implements functions for loading, saving and editing calibre metadata files.
 
     Calibre uses JSON to store metadata on device after each wired transfer.
     In wireless transfers calibre sends the same metadata to the client, which is in charge
     of storing it.
```
Frenzie, line 6: I assume this would show up as a code block? I suppose it explains where you copied it from (I said copy from gettext! :-P)
```suggestion
This module implements functions for loading, saving and editing calibre metadata files.

Calibre uses JSON to store metadata on device after each wired transfer.
In wireless transfers calibre sends the same metadata to the client, which is in charge
of storing it.
```

plugins/calibre.koplugin/metadata.lua
```diff
@@ -1,10 +1,11 @@
...
---]]
+@module koplugin.calibre.metadata
```
Frenzie, line 7:
```suggestion

@module koplugin.calibre.metadata
```

plugins/hello.koplugin/main.lua
```diff
@@ -1,9 +1,16 @@
+--[[--
+This is a debug plugin to test Plugin functionality
```
Frenzie, line 2:
```suggestion
This is a debug plugin to test Plugin functionality.
```

frontend/dispatcher.lua
```diff
@@ -332,11 +321,40 @@ function Dispatcher:init()
...
+    end
+
+    function Hello:init()
+        self:onDispatcherRegisterActions()
+    end
+
+
+@param name the key to use in the table
+
+@param value a table per settingsList above
```
Frenzie, line 343: For me there doesn't need to be a newline in between, but if you prefer that's fine.
```suggestion
@param name the key to use in the table
@param value a table per settingsList above
```

plugins/calibre.koplugin/metadata.lua
yparitcher: It does, I didnt Format it at all I just left it as is. Will fix up.

yparitcher:
> @pazos wrote at https://www.mobileread.com/forums/showthread.php?p=4048983#post4048983:
> 
> > AFAIK gestures can't invoke profile changes. These are two different plugins. Both are using the actions dispatcher, so they can do exactly the same, but you need to configure them separately. (The rationale is: a plugin can be disabled without affecting the behaviour of the rest).
> 
> So, now that plugins car register actions, the "profile" plugin could register its profiles (of dispatcher action) as new dispatcher actions :) that the gesture plugin could use.
> Probably a big bag of messy recursion, but just an idea :)

Bad idea, it would only lead to messy recursion (a profile that calls itself, for an eternal loop), and you can get all the benifits right now, just assign the actions you want to the gesture.

Issue #6763, opened 2020-10-07
FR: [show CJK font full name instead font family name when preview fonts]


windirt:
**Does your feature request involve difficulty completing a task? Please describe.**
not difficult

**Describe the solution you'd like**
when preview font name, if detect CJK font, showing the full font name which has CJK characters, instead showing the font family name which only contain english characters.
![koreader](https://user-images.githubusercontent.com/6456747/95346523-89149c00-08ee-11eb-9f2b-370d7b9009b9.png)

windirt:
dupe to https://github.com/koreader/koreader/issues/6739

poire-z:
No alternative name seems to be accessible from the font library we use:
https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_facerec

windirt:
in the doc, I found this. maybe useful.

family_name  The face's family name. This is an ASCII string, usually in English,  that describes the typeface's family (like ‘Times New Roman’, ‘Bodoni’,  ‘Garamond’, etc). This is a least common denominator used to list fonts.  <ins>Some formats (TrueType & OpenType) provide localized and Unicode  versions of this string. Applications should use the format-specific  interface to access them</ins>. Can be NULL (e.g., in fonts embedded in a PDF file). In case the font doesn't provide a specific family name entry,  FreeType tries to synthesize one, deriving it from other name entries.

windirt:
for example, here is the same ttf in kindle's custom font menu, it's CJK name can be recognized by kindle.
<img width="799" alt="QQ20201007-233423@2x" src="https://user-images.githubusercontent.com/6456747/95353328-bb75c780-08f5-11eb-865b-77613c9f412c.png">

poire-z:
>  Some formats (TrueType & OpenType) provide localized and Unicode versions of this string. Applications should use the format-specific interface to access them

I read that :) and what I meant is that our current library does not have the interface to access them. And adding other libraries just to get them feel a bit heavy :/

NiLuJe:
Does HB have methods to read those?

(Also: that kind of metadata is often quirky and/or bogus and/or wrong and/or outdated, so, here be dragons anyway).

poire-z:
> Does HB have methods to read those?

Possibly: https://harfbuzz.github.io/harfbuzz-hb-ot-name.html 
```c
/**
 * hb_ot_name_id_t:
 * @HB_OT_NAME_ID_INVALID: Value to represent a nonexistent name ID.
[...]

PR #6786, opened 2020-10-12
fontlist: Cache categorized font info (for mupdf)

ezdiy:
Info about each face (l10n, name, family, style etc) is
now cached offline, so fonts can be queried ahead of time.

Fixes https://github.com/koreader/koreader/issues/6763
needs https://github.com/koreader/koreader-base/pull/1215

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6786)

### DIFF
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -4,6 +4,7 @@ local ConfirmBox = require("ui/widget/confirmbox")
 local Device = require("device")
 local Event = require("ui/event")
 local Font = require("ui/font")
+local FontList = require("fontlist")
 local Input = Device.input
 local InputContainer = require("ui/widget/container/inputcontainer")
 local Menu = require("ui/widget/menu")
@@ -61,12 +62,17 @@ function ReaderFont:init()
     -- Font list
     local face_list = cre.getFontFaces()
     for k,v in ipairs(face_list) do
+        local font_filename, font_faceindex = cre.getFontFaceFilenameAndFaceIndex(v)
         table.insert(self.face_table, {
             text_func = function()
                 -- defaults are hardcoded in credocument.lua
                 local default_font = G_reader_settings:readSetting("cre_font") or self.ui.document.default_font
                 local fallback_font = G_reader_settings:readSetting("fallback_font") or self.ui.document.fallback_fonts[1]
                 local text = v
+                if font_filename and font_faceindex then
+                    text = FontList:getLocalizedFontName(font_filename, font_faceindex) or text
+                end
+
                 if v == default_font then
                     text = text .. "   ★"
                 end
@@ -77,7 +83,6 @@ function ReaderFont:init()
             end,
             font_func = function(size)
                 if G_reader_settings:nilOrTrue("font_menu_use_font_face") then
-                    local font_filename, font_faceindex = cre.getFontFaceFilenameAndFaceIndex(v)
                     if font_filename and font_faceindex then
                         return Font:getFace(font_filename, size, font_faceindex)
                     end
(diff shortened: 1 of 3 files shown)


frontend/apps/reader/modules/readerfont.lua
```diff
@@ -66,7 +67,8 @@ function ReaderFont:init()
                 -- defaults are hardcoded in credocument.lua
                 local default_font = G_reader_settings:readSetting("cre_font") or self.ui.document.default_font
                 local fallback_font = G_reader_settings:readSetting("fallback_font") or self.ui.document.fallback_fonts[1]
-                local text = v
+                local text = FontList:getLocalizedFontName(cre.getFontFaceFilenameAndFaceIndex(v)) or v
```
poire-z, line 70: Might do the `cre.getFontFaceFilenameAndFaceIndex(v)` at the start of the for loop, as it's now done for every font, and it might (the default) also be done in the font_func.

frontend/fontlist.lua
```diff
@@ -95,21 +102,65 @@ local function getExternalFontDir()
...
+end
+
+local font_exts = {
+    ["ttf"] = true,
+    ["ttc"] = true,
+    ["cff"] = true,
+    ["otf"] = true,
+}
+
+local function _readList(target, tinfo, mark, dir)
```
poire-z, line 139: Might give a better name to this `tinfo` (which is `fontinfo` in the caller, so why not `fontinfo` here too ?)

frontend/fontlist.lua
```diff
@@ -119,13 +170,64 @@ end
 
 function FontList:getFontList()
     if #self.fontlist > 0 then return self.fontlist end
-    _readList(self.fontlist, self.fontdir)
+
+    local cache_file = DataStorage:getDataDir() .. "/cache/fontinfo.lua"
```
poire-z, line 174: I would have said it's sad (although needed) to need to have and maintain a cache for that - because it's painful to code - but you seem to have done that really well, so ok :)

frontend/apps/reader/modules/readerfont.lua
poire-z: About `FontList:getLocalizedFontName()` - I wonder if we need an option to _not_ get/show the localized one (in case it is crap and people were fine with the regular names). We might add that later if needed.

frontend/fontlist.lua
ezdiy: I'm tempted to move this into a local lambda. There are now 3 variables passed in the recursive call, all of which are fixed state. Would be much better off being captured by the closure at this point.

frontend/apps/reader/modules/readerfont.lua
ezdiy: The field is surprisingly accurate, if I'd wager for user requested option, it would be an option to display *filenames* - as those quite often don't match the "fancy names" of ours now, so its difficult to discern which font is what on storage.

frontend/fontlist.lua
ezdiy: (split into util.findFiles in 670473d3b70d0388355da92e88617d1d40232aca, as this was badly nested already)

Frenzie (commented):
Looks fine at a glance, but I won't have time to really read or test for the next two weeks. (Probably more like three, but in any case I shouldn't have to worry about moving then, only about unpacking. xP

ezdiy:
@Frenzie It's not nearly as invasive as it seems, @poire-z and @NiLuJe having to babysit me will do fine :)

I'll wait with bulk of mupdf until then tho, as that one will need all the eyes due to the horrors around 3rd party building stuff/integration coordination of 4 separate repos etc.

NiLuJe:
Yeah, I won't pretend that I grok everything involved in here, but it looks neat and orderly ;).

PR #6791, opened 2020-10-14
Pocketbook: use raw input I/O

ezdiy:
This allows for better energy efficiency (no more 50Hz tick poll),
as well as lower input lag / higher precision - touch events are
native linux ones.

In addition, auto off/suspend plugin is used in this mode, as we need
to trigger (timed) sleep / poweroff on our own, since the OS ones
will no longer work whenever koreader has focus.

This is for rooted devices only, and possibly somewhat FW
specific, so enabled only on PB740-2 where it's reasonably tested.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6791)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -42,6 +42,19 @@ local PocketBook = Generic:new{
     -- NTX chipsets *should* work (PB631), but in case it doesn't on your device, set this to "no" in here.
     canHWInvert = yes,
 
+    -- If we can access the necessary devices, input events can be handled directly.
+    -- This improves latency (~40ms), as well as power usage - we can spend more time asleep,
+    -- instead of busy looping at 50Hz the way inkview insists on doing.
+    -- In case this method fails (no root), we fallback to classic inkview api.
+    raw_input = nil, --[[{
+        -- value or function to adjust touch matrix orientiation.
+        touch_rotation = -3+4,
+        -- Works same as input.event_map, but for raw input EV_KEY translation
+        keymap = { [scan] = event },
+    }]]
+    -- Runtime state: whether raw input is actually used
+    is_using_raw_input = nil,
+
     -- Private per-model kludges
     _fb_init = function() end,
     _model_init = function() end,
@@ -69,6 +82,9 @@ local function isB288(fb)
 end
 
 function PocketBook:init()
+    local raw_input = self.raw_input
+    local touch_rotation = raw_input and raw_input.touch_rotation or 0
+
     self.screen = require("ffi/framebuffer_mxcfb"):new {
         device = self,
         debug = logger.dbg,
@@ -87,8 +103,14 @@ function PocketBook:init()
             end
             return self._fb_init(fb, finfo, vinfo)
         end,
+        -- raw touch input orientiation is different from the screen
+        getTouchRotation = function(fb)
+            if type(touch_rotation) == "function" then
+                return touch_rotation(self, fb:getRotationMode())
+            end
+            return (4 + fb:getRotationMode() + touch_rotation) % 4
+        end,
     }
-    self.powerd = require("device/pocketbook/powerd"):new{device = self}
 
     -- Whenever we lose focus, but also get suspended for real (we can't reliably tell atm),
     -- plugins need to be notified to stop doing foreground stuff, and vice versa. To this end,
@@ -98,7 +120,8 @@ function PocketBook:init()
 
     self.input = require("device/input"):new{
         device = self,
-        event_map = {
+        raw_input = raw_input,
+        event_map = setmetatable({
             [C.KEY_MENU] = "Menu",
             [C.KEY_PREV] = "LPgBack",
             [C.KEY_NEXT] = "LPgFwd",
@@ -107,7 +130,7 @@ function PocketBook:init()
             [C.KEY_LEFT] = "Left",
             [C.KEY_RIGHT] = "Right",
             [C.KEY_OK] = "Press",
-        },
+        }, {__index=raw_input and raw_input.keymap or {}}),
         handleMiscEv = function(this, ev)
             local ui = require("ui/uimanager")
             if ev.code == C.EVT_HIDE or ev.code == C.EVT_BACKGROUND then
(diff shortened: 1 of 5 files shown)


ezdiy:
Ninja fast track due to https://www.mobileread.com/forums/showpost.php?p=4048095&postcount=327 ... since we're in habit of breaking eggs in master now :)

ptrm:
Sounds like a reason to root my PB. This means that a 20min sleep time, unsettablw in PB settings, could be enable when using KOReader, right?

SmartSelect_20201017-132751_Firefox
Though does the above and the fact that raw events are used also mean, that from KOReader I could no longer use Task Manager mapped to the menu button?

ezdiy:
@ptrm 

Ye, just copy that block over to your device and it should switch to raw input where koreader's autosuspend works. The menu button mappings will still work, but koreader will see em as well (as its no longer filtered by OS).

A possible workaround might be to *not* specify keymap (ie just set `keymap = nil`) - in that case, OS keys should be respected more precisely, though this might be a bit more wonky fw version wise.

There are also some gotchas with the sleep - koreader can indeed set sleep (really just autolock) timer arbitrarily with, but once locked, OS takes over (and then *OS shutdown* timer starts counting, if set). Or you can set shutdown timer in koreader *before* sleep, and that will be respected only if it triggers before autolock. A bit messy, so just "kinda works".

j4w0r:
I have tried it after setting raw_input for PB740 with assert error: "No /hwevent, probably too old firmware".
My Inkpad 3 has latest firmware 5.20.1010 and jailbreak applied.

ezdiy:
@j4w0r I take it falls back to inkview - or does it fail completely (it really shouldn't do that). As noted in the comment, the hwevent may not be necessary.

Here is a change to make it not use /hwinput - (here a copy with [change applied](https://gist.githubusercontent.com/ezdiy/ad2968bd8db5882a00037d4c6f203f67/raw/46e6da7ef1f27ddf1ec0eb229ba635826dc98d54/input_pocketbook.lua))
```diff
diff --git a/ffi/input_pocketbook.lua b/ffi/input_pocketbook.lua
index 93058f7..0f83045 100644
--- a/ffi/input_pocketbook.lua
+++ b/ffi/input_pocketbook.lua
@@ -1,3 +1,4 @@
+local use_hwinput = false
 local ffi = require("ffi")
 local C = ffi.C
 local rt = ffi.load("librt.so")
@@ -107,10 +108,12 @@ function input:open()
         -- Open the monitor queue. We could technically live without it on older firmwares
         -- that don't have it, but I'm not sure how events are meant to be consumed on there
         -- to prevent stall, so for the time being we bail when we don't see it.
-        local hwinput = rt.mq_open("/hwevent", C.O_RDONLY+C.O_NONBLOCK)
-        assert(hwinput >= 0, "No /hwevent, probably too old firmware")
-        poll_fds[0].fd = hwinput
-        poll_fds[0].events = C.POLLIN
+        if use_hwinput then
+            local hwinput = rt.mq_open("/hwevent", C.O_RDONLY+C.O_NONBLOCK)
+            assert(hwinput >= 0, "No /hwevent, probably too old firmware")
+            poll_fds[0].fd = hwinput
+            poll_fds[0].events = C.POLLIN
+        end
         poll_fds_count = 1
 
         -- Open the raw input devices.
@@ -205,7 +208,7 @@ local function waitForEventRaw(t)
         for i=0, poll_fds_count-1 do
             poll_fds[i].revents = 0
         end
-        local res = C.poll(poll_fds, poll_fds_count, ms)
+        local res = C.poll(poll_fds + (use_hwinput and 0 or 1), poll_fds_count - (use_hwinput and 0 or 1), ms)
         if res < 0 then
             error("poll(): " .. ffi.string(C.strerror(ffi.errno())))
         end
@@ -214,7 +217,7 @@ local function waitForEventRaw(t)
         -- thats particularly useful. Keys are nice as they have symbolic names already, but
         -- touch events are utter mess. In any case, this queue *must* be consumed, lest otherwise
         -- monitor would spam queued events to whaetever else gets focus after us.
-        if band(poll_fds[0].revents, C.POLLIN) ~= 0 then
+        if use_hwinput and band(poll_fds[0].revents, C.POLLIN) ~= 0 then
             updateTimestamp() -- single 'ts' copy for genEmuEvent inside the loop
             while rt.mq_receive(poll_fds[0].fd, ffi.cast("char*", hwmsg), hwmsg_len, nil) > 0 do
                 local m = hwmsg[0]
```

j4w0r:
Yes, it was falling back to inkview. Now with hwinput disabled raw_input is working. Thanks!

ezdiy:
@j4w0r Can you see what happens when if exit koreader? On my device (fw6) if I disable /hwinput, all the pent up touches were replayed right after exit, or switch to background. If that doesn't happen on your fw, I can make the hwinput check soft (ie consume it only if it exists).

Uwe-B:
Understanding, that raw_input could somehow improve the touch sensitivity of my PB 740 FW6.1 rooted (pinch to zoom is working sometimes but never spread to zoom), I did try this patch (Adding raw_input ...  to device.lua and replacing input_pocketbook.lua on koreader-pocketbook-v2020.10.1-8-g0406be3_2020-10-18.
Didn't seem to have any effects on my device and after exiting KoReader no effects are registered too.
Tried it with:
use_hwinput = false
use_hwinput = true

ezdiy:
@Uwe-B The effects are marginal for the most part (I dont even bother with complex gestures like pinch). On PB740-2 swipes tend to register better but that's pretty much end of it, changes will need to be made to gesture detection to get better response. Not getting spammed by events when using use_hwinput = false FW6.1 (the one /hwinput is for) is odd though, here it opens a book from the front explorer if I tap the same place in koreader and then switch it to background.

Uwe-B:
I'm not running Koreader as root, could that be the reason.

ezdiy:
@Uwe-B no, but you must chmod the input devices (otherwise it will just fallback to inkview). 

https://github.com/koreader/koreader/pull/6791/files#diff-95367fa064a757a6c7baf98c0a38ee04487e047444cf7a620b3f7321c6d9ed0dR17

Uwe-B:
I did have the lines:
# try to bring in raw device input (on rooted devices)
/mnt/secure/su /bin/chmod 644 /dev/input/*
in /applications/koreader.app
but not in
/system/bin/koreader.app
Adding it to the later did not change the behaviour while setting :
use_hwinput = false

j4w0r:
> On my device (fw6) if I disable /hwinput, all the pent up touches that were done in koreader were replayed right after exit (pressing random stuff on main screen). Happens even after switching koreader to background via back button.

Same behavior on my PB740 (fw5).

PR #6794, opened 2020-10-15
Pocketbook: waveform presets

ezdiy:
cf https://github.com/koreader/koreader-base/issues/1202

On pocketbook, update modes are not as clear cut due to overall
chipset and kernel version mess. Inkview solves this by always
using the slowest (and safe) GC16 waveform. We now do that too
by default.

Slow updates suck though, so there's now a menu entry to configure
it for speed (with mild artifacts at best, kernel panic at worst).

This is a generic interface (any eink Screen can announce support).
The driver may interpret the slow/fast range however they want.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6794)

### DIFF
--- a/frontend/ui/elements/waveform_level.lua
+++ b/frontend/ui/elements/waveform_level.lua
@@ -9,9 +9,9 @@ local items = {}
 for i=0, Screen.wf_level_max do
     local info
     if i == 0 then
-        info = _("Level 0: High quality, slow")
+        info = _("Level 0: high quality, slowest")
     elseif i == Screen.wf_level_max then
-        info = T(_("Level %1: Low quality, fast"), i)
+        info = T(_("Level %1: low quality, fastest"), i)
     else
         info = T(_("Level %1"), i)
     end
@@ -30,7 +30,7 @@ for i=0, Screen.wf_level_max do
 end
 
 return {
-    text = _("Update aggressiveness"),
+    text = _("Refresh speed/fidelity"),
     sub_item_table = items,
 }
 
--- a/frontend/ui/elements/waveform_level.lua
+++ b/frontend/ui/elements/waveform_level.lua
@@ -1,0 +1,36 @@
+local _ = require("gettext")
+local Device = require("device")
+local InfoMessage = require("ui/widget/infomessage")
+local UIManager = require("ui/uimanager")
+local Screen = Device.screen
+local T = require("ffi/util").template
+
+local items = {}
+for i=0, Screen.wf_level_max do
+    local info
+    if i == 0 then
+        info = _("Level 0: High quality, slow")
+    elseif i == Screen.wf_level_max then
+        info = T(_("Level %1: Low quality, fast"), i)
+    else
+        info = T(_("Level %1"), i)
+    end
+
+    table.insert(items, {
+        text = info,
+        checked_func = function() return Screen.wf_level == i end,
+        callback = function()
+            Screen.wf_level = i
+            G_reader_settings:saveSetting("wf_level", i)
+            UIManager:show(InfoMessage:new{
+                text = _("This will take effect on next restart."),
+            })
+        end,
+    })
+end
+
+return {
+    text = _("Update aggressiveness"),
+    sub_item_table = items,
+}
+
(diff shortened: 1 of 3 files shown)


ezdiy:
For review: The backend pocketbook stuff is on me, but wording of the UI entries, or how the UI is done more generally am not all that sure about.

frontend/ui/elements/waveform_level.lua
```diff
@@ -0,0 +1,36 @@
...
+local InfoMessage = require("ui/widget/infomessage")
+local UIManager = require("ui/uimanager")
+local Screen = Device.screen
+local T = require("ffi/util").template
+
+local items = {}
+for i=0, Screen.wf_level_max do
+    local info
+    if i == 0 then
+        info = _("Level 0: High quality, slow")
```
Frenzie, line 12: Maybe something like this?
```suggestion
        info = _("Level 0: high quality, slowest")
```

frontend/ui/elements/waveform_level.lua
```diff
@@ -0,0 +1,36 @@
...
+            G_reader_settings:saveSetting("wf_level", i)
+            UIManager:show(InfoMessage:new{
+                text = _("This will take effect on next restart."),
+            })
+        end,
+    })
+end
+
+return {
+    text = _("Update aggressiveness"),
```
poire-z, line 33: It will be below `Full refresh rate`, so no verb needed, and a bit more context:
`Waveform aggressiveness` (which sounds scary :) 
`Waveform type`
`Waveform level`
(or plural `Waveforms ...` ?
or `Refresh level/type` as "Waveform" might be too technical?

Frenzie: Update is a noun here, not a verb. ;-P I'm more concerned about the "aggressiveness." +1 on something like `Refresh type` to fit in better with the rest of the menu and getting rid of aggressiveness.

ezdiy: 'Update' is meant as a noun in a poor bid to get rid of 'waveform'. Aggressiveness is indeed a jargon that can be scary. Maybe just go with runabout 'Performance level' - because that's ultimately what's it about from user PoV?

Frenzie: Quality instead of performance? Highest performance would imply fastest.

ezdiy: @Frenzie 

>Highest performance would imply fastest.

That's the idea in the first place (low setting => slow, high setting => fast). Though with "performance" it's probably not clear performance of what exactly. Just 'Refresh speed' maybe?

Frenzie: You're reasoning from irrelevant technical bg PB mumbo jumbo. ;-) 

Quality is the variable that matters, speed is just a trade off. Talk about speed and people will enable the highest speed and wonder why it looks shitty. Talk about quality and the highest setting's not slow at all regardless.

ezdiy: > speed and wonder why it looks shitty.

Isn't that obvious from the min/max sentinel though? In addition, purpose of this menu isn't exactly a tradeoff - it works more like a cut-off. Up until point it's just faster with no quality downgrade, and only later on glitches appear (exact point depends on device).

Frenzie: > Isn't that obvious from the min/max sentinel though?

If you'd said "qualification" that might be up in the air, but any time you use jargon like "sentinel" the UX answer is almost certainly a big fat no. :-P

Ftr, you don't need to pick and choose: `Speed and quality`

NiLuJe: My $0.02:

`Refresh speed & quality`, ranging from `Highest quality, slowest` to `Lowest quality, fastest`.

Sidebar: speed/quality are inverted between the two, but I kind of like it that way ;D.

Also, `quality` could be swapped for `fidelity`.

Frenzie: ^ I like that.

ezdiy: Fair enough :)

PR #6796, opened 2020-10-16
Bump base for sqlite.

ezdiy:
+ make CI happy.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6796)


ezdiy:
This is needed for snapshot to pick it up, i think. The pocketbook/harfbuzz stuff should be compatible with older frontend so can sit there for a while.

PR #1527, opened 2015-05-15
Kobo Glo HD support

tarlou:
See discussions in https://github.com/koreader/koreader/issues/1525

Issue #1525, opened 2015-05-11
New device: Kobo Glo HD


tarlou:
Hi, it is my first time to try koreader. It seems that the device Glo HD is not supported yet. Koreader won't start unless I added the following to */.kobo/koreader/frontend/device/kobo/device.lua* (found this with the help of crash.log).
```
local KoboAlyssum = Kobo:new{
    model = "Kobo_alyssum",
    hasFrontlight = yes,
    display_dpi = 300,
}
```
and near the end of the file
```
elseif codename == "alyssum" then
	return KoboAlyssum
```
However, I'm not sure if there are other parameters that need to be set. So far koreader is working except all touches are treated as "long press". It is very tricky to make a short press (turn page or open menu). For most touches, dictionary (or cut/copy/delete/... at the file manager screen) is opened. I haven't figured out a way to fix this.

It would be great if Glo HD can be supported!

Markismus:
[Official specs Kobo GLo HD](http://gl.kobobooks.com/products/kobo-glo-hd):
6” HD Carta E Ink touchscreen, 1448 x 1072 resolution 300 pixels per inch with bezel like the Aura H2O.
[Official specs Aura HD H2O](http://gl.kobobooks.com/products/kobo-aura-h2o): 
Extra-large 6.8” Carta E Ink HD touchscreen with high, 1430 x 1080 resolution, 265 dpi

Looking at the specs for the Dahlia (Aura HD H2O ):
```
-- Kobo Aura H2O:
local KoboAlyssum = Kobo:new{
    model = "Kobo_alyssum",
    hasFrontlight = yes,
    touch_phoenix_protocol = true,
    display_dpi = 300,
    -- the bezel covers the top 11 pixels:
    viewport = Geom:new{x=0, y=11, w=1080, h=1429},
}
```

I would suggest as a first try to use _all_ the info from the H2O. 
The line about `touch_phoenix_protocol` could already solve your long press problem. 
The line about the viewport allows you to specify your width and height. Start with a viewport of `x=0, y=11, w=1072, h=1447`. You can tweak those to get the ideal viewport given the bezel of the Glo HD.
The result would be:
```
-- Kobo Glo HD:
local KoboDahlia = Kobo:new{
    model = "Kobo_dahlia",
    hasFrontlight = yes,
    touch_phoenix_protocol = true,
    display_dpi = 265,
    -- the bezel covers the top 11 pixels:
    viewport = Geom:new{x=0, y=11, w=1072, h=1447},
[...]

PR #6798, opened 2020-10-16
GestureDetector: add Tap interval setting, to avoid bounces

poire-z:
See https://github.com/koreader/koreader/pull/5138#issuecomment-707733771 and above.
And https://www.mobileread.com/forums/showthread.php?t=333958 for the original issue.

Also fixed info_text not displayed, and added more descriptive ones.
Rewording suggestions welcomed (for this one and the 6 other info_text I tweaked).

<kbd>image</kbd>

Works for me instead of my previous handling of that described at https://github.com/koreader/koreader/pull/5138#issuecomment-707384339.
Quickly tested with double tap enabled on the emulator, seems OK.

(There might be already some issues when intervals overlaps, or when we go above 1 second - but not really caring much. I.e. with this one, if single_tap_interval > double_tap_interval, double tap would never trigger.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6798)

### DIFF
--- a/frontend/device/gesturedetector.lua
+++ b/frontend/device/gesturedetector.lua
@@ -430,7 +430,10 @@ function GestureDetector:handleDoubleTap(tev)
     -- a timer if no second tap happened in the double tap delay.
     logger.dbg("set up single/double tap timer")
     -- deadline should be calculated by adding current tap time and the interval
-    local deadline = cur_tap.timev + TimeVal:new{
+    -- (No need to compute self._has_real_clock_time_ev_time here, we should always
+    -- have been thru handleNonTap() where it is computed, before getting here)
+    local ref_time = self._has_real_clock_time_ev_time and cur_tap.timev or TimeVal:now()
+    local deadline = ref_time + TimeVal:new{
         sec = 0,
         usec = not self.input.disable_double_tap and ges_double_tap_interval or 0,
     }
@@ -456,8 +459,22 @@ function GestureDetector:handleNonTap(tev)
         -- switched from other state, probably from initialState
         -- we return nil in this case
         self.states[slot] = self.tapState
+        if self._has_real_clock_time_ev_time == nil then
+            if tev.timev.sec < TimeVal:now().sec - 600 then
+                -- ev.timev is probably the uptime since device boot
+                -- (which might pause on suspend) that we can't use
+                -- with setTimeout(): we'll use TimeVal:now()
+                self._has_real_clock_time_ev_time = false
+                logger.info("event times are not real clock time: some adjustments will be made")
+            else
+                -- assume they are real clock time
+                self._has_real_clock_time_ev_time = true
+                logger.info("event times are real clock time: no adjustment needed")
+            end
+        end
         logger.dbg("set up hold timer")
-        local deadline = tev.timev + TimeVal:new{
+        local ref_time = self._has_real_clock_time_ev_time and tev.timev or TimeVal:now()
+        local deadline = ref_time + TimeVal:new{
             sec = 0, usec = ges_hold_interval
         }
         -- Be sure the following setTimeout only react to this tapState
--- a/frontend/device/gesturedetector.lua
+++ b/frontend/device/gesturedetector.lua
@@ -47,22 +47,31 @@ local TimeVal = require("ui/timeval")
 local logger = require("logger")
 local util = require("util")
 
--- all the time parameters are in us
-local ges_double_tap_interval = G_reader_settings:readSetting("ges_double_tap_interval") or 300 * 1000
-local ges_two_finger_tap_duration = G_reader_settings:readSetting("ges_two_finger_tap_duration") or 300 * 1000
-local ges_hold_interval = G_reader_settings:readSetting("ges_hold_interval") or 500 * 1000
-local ges_pan_delayed_interval = G_reader_settings:readSetting("ges_pan_delayed_interval") or 500 * 1000
-local ges_swipe_interval = G_reader_settings:readSetting("ges_swipe_interval") or 900 * 1000
+-- default values (all the time parameters are in microseconds)
+local TAP_INTERVAL = 0 * 1000
+local DOUBLE_TAP_INTERVAL = 300 * 1000
+local TWO_FINGER_TAP_DURATION = 300 * 1000
+local HOLD_INTERVAL = 500 * 1000
+local PAN_DELAYED_INTERVAL = 500 * 1000
+local SWIPE_INTERVAL = 900 * 1000
+-- current values
+local ges_tap_interval = G_reader_settings:readSetting("ges_tap_interval") or TAP_INTERVAL
+local ges_double_tap_interval = G_reader_settings:readSetting("ges_double_tap_interval") or DOUBLE_TAP_INTERVAL
+local ges_two_finger_tap_duration = G_reader_settings:readSetting("ges_two_finger_tap_duration") or TWO_FINGER_TAP_DURATION
+local ges_hold_interval = G_reader_settings:readSetting("ges_hold_interval") or HOLD_INTERVAL
+local ges_pan_delayed_interval = G_reader_settings:readSetting("ges_pan_delayed_interval") or PAN_DELAYED_INTERVAL
+local ges_swipe_interval = G_reader_settings:readSetting("ges_swipe_interval") or SWIPE_INTERVAL
 
 local GestureDetector = {
     -- must be initialized with the Input singleton class
     input = nil,
-    -- default values (all the time parameters are in us)
-    DOUBLE_TAP_INTERVAL = 300 * 1000,
-    TWO_FINGER_TAP_DURATION = 300 * 1000,
-    HOLD_INTERVAL = 500 * 1000,
-    PAN_DELAYED_INTERVAL = 500 * 1000,
-    SWIPE_INTERVAL = 900 * 1000,
+    -- default values (accessed for display by plugins/gestures.koplugin)
+    TAP_INTERVAL = TAP_INTERVAL,
+    DOUBLE_TAP_INTERVAL = DOUBLE_TAP_INTERVAL,
+    TWO_FINGER_TAP_DURATION = TWO_FINGER_TAP_DURATION,
+    HOLD_INTERVAL = HOLD_INTERVAL,
+    PAN_DELAYED_INTERVAL = PAN_DELAYED_INTERVAL,
+    SWIPE_INTERVAL = SWIPE_INTERVAL,
     -- pinch/spread direction table
     DIRECTION_TABLE = {
         east = "horizontal",
@@ -103,6 +112,7 @@ function GestureDetector:init()
     -- distance parameters
     self.TWO_FINGER_TAP_REGION = 20 * scaler
     self.DOUBLE_TAP_DISTANCE = 50 * scaler
+    self.SINGLE_TAP_BOUNCE_DISTANCE = self.DOUBLE_TAP_DISTANCE
     self.PAN_THRESHOLD = self.DOUBLE_TAP_DISTANCE
     self.MULTISWIPE_THRESHOLD = self.DOUBLE_TAP_DISTANCE
 end
@@ -144,6 +154,15 @@ end
 --[[
 tap2 is the later tap
 --]]
+function GestureDetector:isTapBounce(tap1, tap2)
+    local tv_diff = tap2.timev - tap1.timev
+    return (
+        math.abs(tap1.x - tap2.x) < self.SINGLE_TAP_BOUNCE_DISTANCE and
+        math.abs(tap1.y - tap2.y) < self.SINGLE_TAP_BOUNCE_DISTANCE and
+        (tv_diff.sec == 0 and (tv_diff.usec) < ges_tap_interval)
+    )
+end
+
 function GestureDetector:isDoubleTap(tap1, tap2)
     local tv_diff = tap2.timev - tap1.timev
     return (
@@ -244,7 +263,9 @@ function GestureDetector:clearState(slot)
 end
 
 function GestureDetector:setNewInterval(type, interval)
-    if type == "ges_double_tap_interval" then
+    if type == "ges_tap_interval" then
+        ges_tap_interval = interval
+    elseif type == "ges_double_tap_interval" then
         ges_double_tap_interval = interval
     elseif type == "ges_two_finger_tap_duration" then
         ges_two_finger_tap_duration = interval
@@ -258,7 +279,9 @@ function GestureDetector:setNewInterval(type, interval)
 end
 
 function GestureDetector:getInterval(type)
-    if type == "ges_double_tap_interval" then
+    if type == "ges_tap_interval" then
+        return ges_tap_interval
+    elseif type == "ges_double_tap_interval" then
         return ges_double_tap_interval
     elseif type == "ges_two_finger_tap_duration" then
         return ges_two_finger_tap_duration
@@ -330,6 +353,8 @@ function GestureDetector:tapState(tev)
                 self:clearState(slot)
             end
         elseif self.last_tevs[slot] ~= nil then
+            -- Normal single tap seems to always go thru here
+            -- (the next 'else' might be there for edge cases)
             return self:handleDoubleTap(tev)
         else
             -- last tev in this slot is cleared by last two finger tap
@@ -368,6 +393,17 @@ function GestureDetector:handleDoubleTap(tev)
         timev = tev.timev,
     }
 
+    -- We do tap bounce detection even when double tap is enabled (so, double tap
+    -- is triggered when: ges_tap_interval <= delay < ges_double_tap_interval)
+    if ges_tap_interval > 0 and self.last_taps[slot] ~= nil and self:isTapBounce(self.last_taps[slot], cur_tap) then
+        logger.dbg("tap bounce detected in slot", slot, ": ignored")
+        -- Simply ignore it, and clear state as this is the end of a touch event
+        -- (this doesn't clear self.last_taps[slot], so a 3rd tap can be detected
+        -- as a double tap)
+        self:clearState(slot)
+        return
+    end
+
     if not self.input.disable_double_tap and self.last_taps[slot] ~= nil and
                 self:isDoubleTap(self.last_taps[slot], cur_tap) then
         -- it is a double tap
@@ -381,16 +417,27 @@ function GestureDetector:handleDoubleTap(tev)
     -- set current tap to last tap
     self.last_taps[slot] = cur_tap
 
-    logger.dbg("set up tap timer")
+    if self.input.disable_double_tap then
+        -- We can send the event immediately (no need for the
+        -- timer stuff needed for double tap support)
+        logger.dbg("single tap detected in slot", slot, ges_ev.pos)
+        self:clearState(slot)
+        return ges_ev
+    end
+
+    -- Double tap enabled: we can't send this single tap immediately as it
+    -- may be the start of a double tap. We'll send it as a single tap after
+    -- a timer if no second tap happened in the double tap delay.
+    logger.dbg("set up single/double tap timer")
     -- deadline should be calculated by adding current tap time and the interval
     local deadline = cur_tap.timev + TimeVal:new{
         sec = 0,
         usec = not self.input.disable_double_tap and ges_double_tap_interval or 0,
     }
     self.input:setTimeout(function()
-        logger.dbg("in tap timer", self.last_taps[slot] ~= nil)
+        logger.dbg("in single/double tap timer", self.last_taps[slot] ~= nil)
         -- double tap will set last_tap to nil so if it is not, then
-        -- user must only tapped once
+        -- user has not double-tap'ed: it's a single tap
         if self.last_taps[slot] ~= nil then
             self.last_taps[slot] = nil
             -- we are using closure here
--- a/plugins/gestures.koplugin/main.lua
+++ b/plugins/gestures.koplugin/main.lua
@@ -437,6 +437,7 @@ function Gestures:addIntervals(menu_items)
         sub_item_table = {
             {
                 text = _("Text selection rate"),
+                keep_menu_open = true,
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
                     local current_value = G_reader_settings:readSetting("hold_pan_rate")
@@ -444,18 +445,19 @@ function Gestures:addIntervals(menu_items)
                         current_value = Screen.low_pan_rate and 5.0 or 30.0
                     end
                     local items = SpinWidget:new{
-                        text = T(_([[
+                        title_text = _("Text selection rate"),
+                        info_text = T(_([[
 The rate is how often screen will be refreshed per second while selecting text.
 Higher values mean faster screen updates, but also use more CPU.
+
 Default value: %1]]), Screen.low_pan_rate and 5.0 or 30.0),
-                        width = math.floor(Screen:getWidth() * 0.6),
+                        width = math.floor(Screen:getWidth() * 0.75),
                         value = current_value,
                         value_min = 1.0,
                         value_max = 60.0,
                         value_step = 1,
                         value_hold_step = 15,
                         ok_text = _("Set rate"),
-                        title_text = _("Text selection rate"),
                         default_value = Screen.low_pan_rate and 5.0 or 30.0,
                         callback = function(spin)
                             G_reader_settings:saveSetting("hold_pan_rate", spin.value)
@@ -467,23 +469,54 @@ Default value: %1]]), Screen.low_pan_rate and 5.0 or 30.0),
                 separator = true,
             },
             {
-                text = _("Double tap interval"),
+                text = _("Tap interval"),
+                keep_menu_open = true,
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
-                        text = T(_([[
-Set double tap interval in milliseconds.
-The interval value can range from 100 (0.1 seconds) to 2000 (2 seconds).
+                        title_text = _("Tap interval"),
+                        info_text = T(_([[
+Any other taps made within this interval after a first tap will be considered accidental and ignored.
+
+The interval value is in milliseconds and can range from 0 (0 second) to 2000 (2 seconds).
+Default value: %1]]), GestureDetector.TAP_INTERVAL/1000),
+                        width = math.floor(Screen:getWidth() * 0.75),
+                        value = GestureDetector:getInterval("ges_tap_interval")/1000,
+                        value_min = 0,
+                        value_max = 2000,
+                        value_step = 50,
+                        value_hold_step = 200,
+                        ok_text = _("Set interval"),
+                        default_value = GestureDetector.TAP_INTERVAL/1000,
+                        callback = function(spin)
+                            G_reader_settings:saveSetting("ges_tap_interval", spin.value*1000)
+                            GestureDetector:setNewInterval("ges_tap_interval", spin.value*1000)
+                        end
+                    }
+                    UIManager:show(items)
+                end,
+            },
+            {
+                text = _("Double tap interval"),
+                keep_menu_open = true,
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local GestureDetector = require("device/gesturedetector")
+                    local items = SpinWidget:new{
+                        title_text = _("Double tap interval"),
+                        info_text = T(_([[
+When double tap is enabled, this sets the time to wait for the second tap. A single tap will take at least this long to be detected.
+
+The interval value is in milliseconds and can range from 100 (0.1 seconds) to 2000 (2 seconds).
 Default value: %1]]), GestureDetector.DOUBLE_TAP_INTERVAL/1000),
-                        width = math.floor(Screen:getWidth() * 0.6),
+                        width = math.floor(Screen:getWidth() * 0.75),
                         value = GestureDetector:getInterval("ges_double_tap_interval")/1000,
                         value_min = 100,
                         value_max = 2000,
                         value_step = 100,
                         value_hold_step = 500,
                         ok_text = _("Set interval"),
-                        title_text = _("Double tap interval"),
                         default_value = GestureDetector.DOUBLE_TAP_INTERVAL/1000,
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_double_tap_interval", spin.value*1000)
@@ -495,22 +528,24 @@ Default value: %1]]), GestureDetector.DOUBLE_TAP_INTERVAL/1000),
             },
             {
                 text = _("Two finger tap duration"),
+                keep_menu_open = true,
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
-                        text = T(_([[
-Set two finger tap duration in milliseconds.
-The duration value can range from 100 (0.1 seconds) to 2000 (2 seconds).
+                        title_text = _("Two finger tap duration"),
+                        info_text = T(_([[
+This sets the allowed duration of any of the two fingers touch/release for the combined event to be considered a two finger tap.
+
+The duration value is in milliseconds and can range from 100 (0.1 seconds) to 2000 (2 seconds).
 Default value: %1]]), GestureDetector.TWO_FINGER_TAP_DURATION/1000),
-                        width = math.floor(Screen:getWidth() * 0.6),
+                        width = math.floor(Screen:getWidth() * 0.75),
                         value = GestureDetector:getInterval("ges_two_finger_tap_duration")/1000,
                         value_min = 100,
                         value_max = 2000,
                         value_step = 100,
                         value_hold_step = 500,
                         ok_text = _("Set duration"),
-                        title_text = _("Two finger tap duration"),
                         default_value = GestureDetector.TWO_FINGER_TAP_DURATION/1000,
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_two_finger_tap_duration", spin.value*1000)
@@ -522,22 +557,24 @@ Default value: %1]]), GestureDetector.TWO_FINGER_TAP_DURATION/1000),
             },
             {
                 text = _("Hold interval"),
+                keep_menu_open = true,
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
-                        text = T(_([[
-Set hold interval in milliseconds.
-The interval value can range from 100 (0.1 seconds) to 2000 (2 seconds).
+                        title_text = _("Hold interval"),
+                        info_text = T(_([[
+If a touch is not released in this interval, it is considered a hold (or long-press). On document's text, single word selection is then triggered.
+
+The interval value is in milliseconds and can range from 100 (0.1 seconds) to 2000 (2 seconds).
 Default value: %1]]), GestureDetector.HOLD_INTERVAL/1000),
-                        width = math.floor(Screen:getWidth() * 0.6),
+                        width = math.floor(Screen:getWidth() * 0.75),
                         value = GestureDetector:getInterval("ges_hold_interval")/1000,
                         value_min = 100,
                         value_max = 2000,
                         value_step = 100,
                         value_hold_step = 500,
                         ok_text = _("Set interval"),
-                        title_text = _("Hold interval"),
                         default_value = GestureDetector.HOLD_INTERVAL/1000,
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_hold_interval", spin.value*1000)
@@ -549,22 +586,24 @@ Default value: %1]]), GestureDetector.HOLD_INTERVAL/1000),
             },
             {
                 text = _("Pan delay interval"),
+                keep_menu_open = true,
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
-                        text = T(_([[
-Set pan delay interval in milliseconds.
-The interval value can range from 100 (0.1 seconds) to 2000 (2 seconds).
+                        title_text = _("Pan delay interval"),
+                        info_text = T(_([[
+This is used where necessary to reduce potential activation of panning when swiping is intended (e.g., for the menu or for multiswipe).
+
+The interval value is in milliseconds and can range from 100 (0.1 seconds) to 2000 (2 seconds).
 Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
-                        width = math.floor(Screen:getWidth() * 0.6),
+                        width = math.floor(Screen:getWidth() * 0.75),
                         value = GestureDetector:getInterval("ges_pan_delayed_interval")/1000,
                         value_min = 100,
                         value_max = 2000,
                         value_step = 100,
                         value_hold_step = 500,
                         ok_text = _("Set interval"),
-                        title_text = _("Pan delay interval"),
                         default_value = GestureDetector.PAN_DELAYED_INTERVAL/1000,
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_pan_delayed_interval", spin.value*1000)
@@ -576,22 +615,24 @@ Default value: %1]]), GestureDetector.PAN_DELAYED_INTERVAL/1000),
             },
             {
                 text = _("Swipe interval"),
+                keep_menu_open = true,
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
-                        text = T(_([[
-Set swipe interval in milliseconds.
-The interval value can range from 100 (0.1 seconds) to 2000 (2 seconds).
+                        title_text = _("Swipe interval"),
+                        info_text = T(_([[
+This sets the maximum delay between the start and the end of a swipe for it to be considered a swipe. Above this interval, it's considered panning.
+
+The interval value is in milliseconds and can range from 100 (0.1 seconds) to 2000 (2 seconds).
 Default value: %1]]), GestureDetector.SWIPE_INTERVAL/1000),
-                        width = math.floor(Screen:getWidth() * 0.6),
+                        width = math.floor(Screen:getWidth() * 0.75),
                         value = GestureDetector:getInterval("ges_swipe_interval")/1000,
                         value_min = 100,
                         value_max = 2000,
                         value_step = 100,
                         value_hold_step = 500,
                         ok_text = _("Set interval"),
-                        title_text = _("Swipe interval"),
                         default_value = GestureDetector.SWIPE_INTERVAL/1000,
                         callback = function(spin)
                             G_reader_settings:saveSetting("ges_swipe_interval", spin.value*1000)
--- a/frontend/device/cervantes/device.lua
+++ b/frontend/device/cervantes/device.lua
@@ -1,5 +1,4 @@
 local Generic = require("device/generic/device")
-local TimeVal = require("ui/timeval")
 local logger = require("logger")
 
 local function yes() return true end
@@ -56,7 +55,6 @@ local Cervantes = Generic:new{
     touch_legacy = true, -- SingleTouch input events
     touch_switch_xy = true,
     touch_mirrored_x = true,
-    touch_probe_ev_epoch_time = true,
     hasOTAUpdates = yes,
     hasFastWifiStatusQuery = yes,
     hasKeys = yes,
@@ -119,37 +117,16 @@ local Cervantes4 = Cervantes:new{
 }
 
 -- input events
-local probeEvEpochTime
--- this function will update itself after the first touch event
-probeEvEpochTime = function(self, ev)
-    local now = TimeVal:now()
-    -- This check should work as long as main UI loop is not blocked for more
-    -- than 10 minute before handling the first touch event.
-    if ev.time.sec <= now.sec - 600 then
-        -- time is seconds since boot, force it to epoch
-        probeEvEpochTime = function(_, _ev)
-            _ev.time = TimeVal:now()
-        end
-        ev.time = now
-    else
-        -- time is already epoch time, no need to do anything
-        probeEvEpochTime = function(_, _) end
-    end
-end
 function Cervantes:initEventAdjustHooks()
     if self.touch_switch_xy then
         self.input:registerEventAdjustHook(self.input.adjustTouchSwitchXY)
     end
+
     if self.touch_mirrored_x then
         self.input:registerEventAdjustHook(
             self.input.adjustTouchMirrorX,
             self.screen:getWidth()
         )
-    end
-    if self.touch_probe_ev_epoch_time then
-        self.input:registerEventAdjustHook(function(_, ev)
-            probeEvEpochTime(_, ev)
-        end)
     end
 
     if self.touch_legacy then
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -1,6 +1,5 @@
 local Generic = require("device/generic/device")
 local Geom = require("ui/geometry")
-local TimeVal = require("ui/timeval")
 local WakeupMgr = require("device/wakeupmgr")
 local logger = require("logger")
 local util = require("ffi/util")
@@ -59,10 +58,6 @@ local KoboTrilogy = Kobo:new{
     model = "Kobo_trilogy",
     needsTouchScreenProbe = yes,
     touch_switch_xy = false,
-    -- Some Kobo Touch models' kernel does not generate touch event with epoch
-    -- timestamp. This flag will probe for those models and setup event adjust
-    -- hook accordingly
-    touch_probe_ev_epoch_time = true,
     hasKeys = yes,
     hasMultitouch = no,
 }
@@ -80,7 +75,6 @@ local KoboPixie = Kobo:new{
 local KoboDaylight = Kobo:new{
     model = "Kobo_daylight",
     hasFrontlight = yes,
-    touch_probe_ev_epoch_time = true,
     touch_phoenix_protocol = true,
     display_dpi = 300,
     hasNaturalLight = yes,
@@ -136,7 +130,6 @@ local KoboSnow = Kobo:new{
     hasFrontlight = yes,
     touch_snow_protocol = true,
     touch_mirrored_x = false,
-    touch_probe_ev_epoch_time = true,
     display_dpi = 265,
     hasNaturalLight = yes,
     frontlight_settings = {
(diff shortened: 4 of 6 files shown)


ezdiy:
Looks good to detect actual hw bounces, as well as deal with koreader induced bounces (for the time being).

Do note, there are 2 major sources of tap bounces on my device:

* Incorrect timestamp comparisons in gesture detector (don't consider seconds wrap-around)

* Gesture detector is a busy wait hogging 100% cpu until the ges interval hits (in this case, when detecting tap). The busy loop  fails to jit due to constant error(), and steals cpu time from kernel interrupt handlers, thus making most of touch events rather choppy.

Unfortunately fixing this is not trivial, as that would entail unifying setTimeout and UIManager timers.

poire-z:
Mhhh, some issue with this when a full refresh is triggered.
I have full refresh set to every 6 pages. And double-tap disabled. And I set this new tap interval to 400ms (which seems fine for me).
When I do some double-tap to go forward or backward, my first tap changes pages, and the second tap (usually 200ms later) is considered a bounce.
Except when I happen on the 6th page that causes a full refresh: the second tap is now 600ms later, and is no more considered a bounce.
The 200ms or 600ms delay is due to the timetstamp of the 2nd tap being 200ms or 600ms later, even if I did all the double taps at the same rhythm, so with a real delay of around 200ms.
Any idea on how to deal with that?

plugins/gestures.koplugin/main.lua
```diff
@@ -466,24 +468,55 @@ Default value: %1]]), Screen.low_pan_rate and 5.0 or 30.0),
...
+            {
+                text = _("Tap interval"),
+                keep_menu_open = true,
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local GestureDetector = require("device/gesturedetector")
+                    local items = SpinWidget:new{
+                        title_text = _("Tap interval"),
+                        info_text = T(_([[
+Another tap made in this interval after a first tap will be considered a bounce and ignored.
```
Frenzie, line 480: Maybe something like this?
```suggestion
Any other taps made within this interval after a first tap will be considered accidental and ignored.
```

plugins/gestures.koplugin/main.lua
```diff
@@ -466,24 +468,55 @@ Default value: %1]]), Screen.low_pan_rate and 5.0 or 30.0),
...
                 callback = function()
                     local SpinWidget = require("ui/widget/spinwidget")
                     local GestureDetector = require("device/gesturedetector")
                     local items = SpinWidget:new{
-                        text = T(_([[
-Set double tap interval in milliseconds.
-The interval value can range from 100 (0.1 seconds) to 2000 (2 seconds).
+                        title_text = _("Double tap interval"),
+                        info_text = T(_([[
+When double tap is enabled, this sets the time to wait for the second tap. Only after this interval has ellapsed will a single tap be emitted.
```
Frenzie, line 509: elapsed, but maybe something like
```suggestion
When double tap is enabled, this sets the time to wait for the second tap. A single tap will take at least this long to be detected.
```

frontend/device/gesturedetector.lua
```diff
@@ -144,6 +147,15 @@ end
 --[[
 tap2 is the later tap
 --]]
+function GestureDetector:isTapBounce(tap1, tap2)
+    local tv_diff = tap2.timev - tap1.timev
+    return (
+        math.abs(tap1.x - tap2.x) < self.SINGLE_TAP_BOUNCE_DISTANCE and
```
Frenzie, line 153: Taking an absolute value looks a bit suspicious at a glance? Could you explain the rationale?

frontend/device/gesturedetector.lua
```diff
@@ -57,7 +58,8 @@ local ges_swipe_interval = G_reader_settings:readSetting("ges_swipe_interval") o
 local GestureDetector = {
     -- must be initialized with the Input singleton class
     input = nil,
-    -- default values (all the time parameters are in us)
+    -- default values (all the time parameters are in milliseconds)
+    TAP_INTERVAL = 50 * 1000,
```
Frenzie, line 62: I strongly I suspect I do *not* want this except possibly with a lower value like 10 or 20, but I can't quickly test atm due to whatever this is, please resolve it if you haven't yet guys ;-)

```
ffi.load (warning): /home/frans/src/kobo/koreader/koreader-emulator-x86_64-linux-gnu-debug/koreader/libs/libfreetype.so.6: undefined symbol: ft_smooth_lcd_renderer_class
./luajit: ./setupkoenv.lua:27: Not able to load dynamic library: libs/libmupdf.so
```

Frenzie (commented):
Looks reasonable at a glance, very strong doubts about the proposed default though.

frontend/device/gesturedetector.lua
ezdiy: @Frenzie you need clean rebuild of ft and mupdf. Makefile has never tracked those deps, ie you have to rm manually. This stuff will be fixed with mupdf revamp (mostly because its supremely annoying when constantly rebuilding this stuff).

Frenzie: Ah right, I saw something about that a few days ago. But I'm really just dropping by for 5-10 mins. ;-)

frontend/device/gesturedetector.lua
poire-z: Not my rationale. This is a copy of isDoubleTap() just after - just to use different constants.

But taking the absolute value looks right to me when we want a distance - and make sure the 2 taps (whether bounce or real double tap) are near each others - so the user is not taping with 2 fingers far from each other.
So, 2 taps far from each other is not considered a bounce (or, it's an ultra wide bounce :) and probably not an accident :)
(I guess this might also be to not catch two-fingers tap - but not certain as 2 fingers tap must use 2 different slots.)

frontend/device/gesturedetector.lua
poire-z: I don't care setting it to a lower value, even 0 if you feel like it.
10-20 is a bit awkward as we allow steps of 50 (and we don't really need steps lower than that).
And franckly, I challenge you to do with your fingers on a screen or a mouse on any device 2 taps (touch release) in less than 100ms :)
The real usable values are in the range 200ms - 600ms I guess.
I've set 50 just because I'm quite sure it wouldn't be triggered in real-life - while still allowing the computation/checks to be done, so the new codepaths/checks would be taken by all users, and we'd be more assured there's no issue :)

Frenzie: Okay, 50 ms seems acceptable, at least on my H2O.

Frenzie: > And franckly, I challenge you to do with your fingers on a screen or a mouse on any device 2 taps (touch release) in less than 100ms :)

I probably don't give a hoot about 2 taps, but I do care about ~5-20 quick taps in succession.

poire-z:
Mhhh, one more issue with this:
after a second tap is prevented, the next swipe or multiswipe is also prevented....

frontend/device/gesturedetector.lua
Frenzie: I mean suspicious more in the sense that abs(3-4) = 1, for example. But I don't recall otoh how all this works.

Galunid: @Frenzie This one is alright, what it does is check if the difference in x/y coordinates between taps is in a square of length `self.SINGLE_TAP_BOUNCE_DISTANCE`.

poire-z:
> after a second tap is prevented, the next swipe or multiswipe is also prevented....

... because the state was switched to being in "Pan state", so we got a pan instead of the swipe.

Anyway, 400ms was fine for me to avoid bounces (and allowed faster page browsing than my previous rate=1.5 which must have behaved as setting this to 666ms).
But to prevent a bounce after a full refresh, I need to set it to 700ms (and I'm back to my previous feeling).

@NiLuJe : am I right in understanding that a full refresh would cause (on Kobo) a "wait for update complete", and that this would cause KOReader code to block? And so, the next event, when read from the driver events queue, would be given a timestamp of "now" (so, after the time spent blocking)?

I don't want to mess with the low level event stuff :)
Could somehow UIManager detect such wait for completion and blocking - to somehow set into gesturedetector 2 infos: (last_blocking_start_ts, last_blocking_duration) so we could use it to adjust delays when needed?
(If ugly or too complicated or too edge case'y, I'm fine with just using 700ms and call this a day :)

frontend/device/gesturedetector.lua
NiLuJe: You need a clean rebuild (or a targeted one, c.f., https://github.com/koreader/koreader-base/pull/1214) ;).

NiLuJe:
@poire-z: Yep, that's the whole point of the mechanism: it blocks.

The timestamp *should* be sane, though, except possibly on devices where we mangle timestamps. What [@ezdiy mentioned](https://github.com/koreader/koreader/pull/6798#issuecomment-710761672) might also be relevant here. 

We currently don't, but I *can* compute the exact amount of time it blocked for (with a shitty 10ms precision).

frontend/device/gesturedetector.lua
NiLuJe: I do agree on the defaults, though ;p. And possibly a bypass if set to 0.

NiLuJe: Wheee, magic refresh 3 hours late :D.

NiLuJe: And, yeah, the quick tapping test I did on the Forma generated a tap every ~50ms ;).

poire-z: > quick tapping test I did on the Forma generated a tap every ~50ms ;).

<=49 ms ?! I don't believe you :)
But OK, I see that 50ms is hurting some feelings :) so ok, let's go with a default of 0 and a bypass (so at least, my clearState stuff can be bypassed in case it does more bad than good).

poire-z:
> except possibly on devices where we mangle timestamps (which is unfortunately most anything < Mk. 7 on Kobo).

So, no issue for you on the Forma when a full refresh happens ?
Only older Kobos like my GloHD would see this issue? Any other devices/brands?

> I can compute the exact amount of time it blocked for (with a shitty 10ms precision).

Anyway, with the mangled timestamps, and having that info, we wouldn't know if the event was really triggered just before or just after the blocking full refresh. So, handling this would cause any tap to be avoided for tap_interval after the full refresh is done. Not sure which is preferable...

frontend/device/gesturedetector.lua
NiLuJe: Well, *almost* ;).

```
E: 171.731913 0001 014a 0000    # EV_KEY / BTN_TOUCH            0
E: 171.731913 0001 0145 0000    # EV_KEY / BTN_TOOL_FINGER      0
E: 171.731913 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +30ms
E: 171.787921 0001 0145 0001    # EV_KEY / BTN_TOOL_FINGER      1
E: 171.787921 0001 014a 0001    # EV_KEY / BTN_TOUCH            1
E: 171.787921 0003 0039 0000    # EV_ABS / ABS_MT_TRACKING_ID   0
E: 171.787921 0003 003b 0000    # EV_ABS / ABS_MT_DISTANCE      0
E: 171.787921 0003 0035 1224    # EV_ABS / ABS_MT_POSITION_X    1224
E: 171.787921 0003 0036 0947    # EV_ABS / ABS_MT_POSITION_Y    947
E: 171.787921 0003 003a 0058    # EV_ABS / ABS_MT_PRESSURE      58
E: 171.787921 0003 0030 0011    # EV_ABS / ABS_MT_TOUCH_MAJOR   11
E: 171.787921 0003 0031 0023    # EV_ABS / ABS_MT_TOUCH_MINOR   23
E: 171.787921 0003 0034 0125    # EV_ABS / ABS_MT_ORIENTATION   125
E: 171.787921 0000 0002 0000    # ++++++++++++ SYN_MT_REPORT (0) ++++++++++
E: 171.787921 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +56ms
E: 171.842254 0003 0039 0000    # EV_ABS / ABS_MT_TRACKING_ID   0
E: 171.842254 0003 003b 0000    # EV_ABS / ABS_MT_DISTANCE      0
E: 171.842254 0003 0035 1224    # EV_ABS / ABS_MT_POSITION_X    1224
E: 171.842254 0003 0036 0949    # EV_ABS / ABS_MT_POSITION_Y    949
E: 171.842254 0003 003a 0032    # EV_ABS / ABS_MT_PRESSURE      32
E: 171.842254 0003 0030 0011    # EV_ABS / ABS_MT_TOUCH_MAJOR   11
E: 171.842254 0003 0031 0023    # EV_ABS / ABS_MT_TOUCH_MINOR   23
E: 171.842254 0003 0034 0125    # EV_ABS / ABS_MT_ORIENTATION   125
E: 171.842254 0000 0002 0000    # ++++++++++++ SYN_MT_REPORT (0) ++++++++++
E: 171.842254 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +55ms
E: 171.873092 0001 014a 0000    # EV_KEY / BTN_TOUCH            0
E: 171.873092 0001 0145 0000    # EV_KEY / BTN_TOOL_FINGER      0
E: 171.873092 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +31ms
E: 171.939893 0001 0145 0001    # EV_KEY / BTN_TOOL_FINGER      1
E: 171.939893 0001 014a 0001    # EV_KEY / BTN_TOUCH            1
E: 171.939893 0003 0039 0000    # EV_ABS / ABS_MT_TRACKING_ID   0
E: 171.939893 0003 003b 0000    # EV_ABS / ABS_MT_DISTANCE      0
E: 171.939893 0003 0035 1223    # EV_ABS / ABS_MT_POSITION_X    1223
E: 171.939893 0003 0036 0959    # EV_ABS / ABS_MT_POSITION_Y    959
E: 171.939893 0003 003a 0054    # EV_ABS / ABS_MT_PRESSURE      54
E: 171.939893 0003 0030 0011    # EV_ABS / ABS_MT_TOUCH_MAJOR   11
E: 171.939893 0003 0031 0023    # EV_ABS / ABS_MT_TOUCH_MINOR   23
E: 171.939893 0003 0034 0125    # EV_ABS / ABS_MT_ORIENTATION   125
E: 171.939893 0000 0002 0000    # ++++++++++++ SYN_MT_REPORT (0) ++++++++++
E: 171.939893 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +66ms
E: 171.983499 0003 0039 0000    # EV_ABS / ABS_MT_TRACKING_ID   0
E: 171.983499 0003 003b 0000    # EV_ABS / ABS_MT_DISTANCE      0
E: 171.983499 0003 0035 1228    # EV_ABS / ABS_MT_POSITION_X    1228
E: 171.983499 0003 0036 0959    # EV_ABS / ABS_MT_POSITION_Y    959
E: 171.983499 0003 003a 0039    # EV_ABS / ABS_MT_PRESSURE      39
E: 171.983499 0003 0030 0011    # EV_ABS / ABS_MT_TOUCH_MAJOR   11
E: 171.983499 0003 0031 0023    # EV_ABS / ABS_MT_TOUCH_MINOR   23
E: 171.983499 0003 0034 0125    # EV_ABS / ABS_MT_ORIENTATION   125
E: 171.983499 0000 0002 0000    # ++++++++++++ SYN_MT_REPORT (0) ++++++++++
E: 171.983499 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +44ms
E: 172.014154 0001 014a 0000    # EV_KEY / BTN_TOUCH            0
E: 172.014154 0001 0145 0000    # EV_KEY / BTN_TOOL_FINGER      0
E: 172.014154 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +31ms
E: 172.070519 0001 0145 0001    # EV_KEY / BTN_TOOL_FINGER      1
E: 172.070519 0001 014a 0001    # EV_KEY / BTN_TOUCH            1
E: 172.070519 0003 0039 0000    # EV_ABS / ABS_MT_TRACKING_ID   0
E: 172.070519 0003 003b 0000    # EV_ABS / ABS_MT_DISTANCE      0
E: 172.070519 0003 0035 1252    # EV_ABS / ABS_MT_POSITION_X    1252
E: 172.070519 0003 0036 0960    # EV_ABS / ABS_MT_POSITION_Y    960
E: 172.070519 0003 003a 0060    # EV_ABS / ABS_MT_PRESSURE      60
E: 172.070519 0003 0030 0011    # EV_ABS / ABS_MT_TOUCH_MAJOR   11
E: 172.070519 0003 0031 0023    # EV_ABS / ABS_MT_TOUCH_MINOR   23
E: 172.070519 0003 0034 0125    # EV_ABS / ABS_MT_ORIENTATION   125
E: 172.070519 0000 0002 0000    # ++++++++++++ SYN_MT_REPORT (0) ++++++++++
E: 172.070519 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +56ms
E: 172.133521 0001 014a 0000    # EV_KEY / BTN_TOUCH            0
E: 172.133521 0001 0145 0000    # EV_KEY / BTN_TOOL_FINGER      0
E: 172.133521 0000 0000 0000    # ------------ SYN_REPORT (0) ---------- +63ms
```

NiLuJe:
Haven't actually tried the code, but there's no timestamp mangling on Forma, so I suspect it'll work.

On Kobo, anything using the "alyssum" proto, and anything where the probe_ev_epoch_time hakc actually takes *will* generate timestamps from within KOReader, though. IIRC, that's almost everything < Mk. 7, like I said ;).

NiLuJe:
Yep, just tried it, set to 400ms, if I do a succession of quick double-taps, I always get a single page-turn per pair, regardless of flashes.

NiLuJe:
Same on a H2O, where the kernel timestamps are also sane ;).

frontend/device/gesturedetector.lua
Frenzie: > let's go with a default of 0

:balloon: :+1: :100:

NiLuJe (approved):

poire-z:
(Don't merge yet, I'll try fixing the full refresh issue on old kobo.)

poire-z:
> `default to 0, and bypass check if 0`

50 or 0, it wouldn't be triggered on any device in the hands of a real human, so ok for a default of 0 and bypass :)

> `handle tap interval over blocking screen refreshes`

@NiLuJe : any opinion on https://github.com/koreader/koreader/pull/6798/commits/9bedf0a4fe0bc952f1bcc82719c12910cb58171c ? No issue you can foresee on other devices than Kobo?
It's a bit much code for handling this case, but it works rather fine for me.
On my GloHD, refreshes duration are around 1ms when non blocking, and around 600ms when blocking.
With a tap_interval of 400ms, and quick double taping, I'd now get a single page forward.
When browsing page fast, I could still get a wished tap prevented when a full-refresh happens, but well, can't do magic without the real timestamps.
Anyway, with this commit, it might give a better feeling that it works generally well than without.

NiLuJe:
I'm... not super fond of 9bedf0a to be honest ^^.

Would a quick'n dirty fudging of the tap_interval (say, +500ms) if the refresh to be sent is flashing work instead?

poire-z:
> Kobo, alyssum proto 

I didn't want in gesturedetector.lua or uimanager.lua to put any device specific stuff.

> or the probe timestamp thingy being still set to a function

It looks to me it is always and stays a function. We just replace stuff with a no-op function when it's not needed.
And there are x/y mirroring/switch stuff that are added in that function too.

> Would a quick'n dirty fudging of the tap_interval (say, +500ms [...]) 

I initially went with this: just adding the refresh duration. But then, tap was prevented for 600/700ms instead of my wished 400ms - so more after a full refresh, and when doing quick browsing with tap, I got stopped after a full refresh.

> if the refresh to be sent is flashing work instead?

Are you sure in there (uimanager.lua where I put the before/after stuff) from the mode that the refresh you send is going to block for >500ms vs the other at 1ms - on all devices?
Somehow, deciding what to do or not depending on that refresh duration (known after it happened) seems like it would work in all cases, without having to know about devices and type of refreshes.

> I'm... not super fond of 9bedf0a to be honest ^^.

Well :) It's a bit long, and the fact that for genericity, I also check the 2nd slot [1] although we don't really care much about two fingets tap in that case. If we're sure a single tap happens always in slot [0], I could shorten it :)

NiLuJe:
> > Kobo, alyssum proto
> 
> I didn't want in gesturedetector.lua or uimanager.lua to put any device specific stuff.
> 
> > or the probe timestamp thingy being still set to a function
> 
> It looks to me it is always and stays a function. We just replace stuff with a no-op function when it's not needed.
> And there are x/y mirroring/switch stuff that are added in that function too.
> 

I was thinking in terms of yet-another-Device cap ^^. But, yeah, we can only account for probe_ev_epoch_time being *asked* for, not necessarily being still active. Supposedly, we only probe it on devices where it's necessary, so, might not be a real problem for a device cap.

> > Would a quick'n dirty fudging of the tap_interval (say, +500ms [...])
> 
> I initially went with this: just adding the refresh duration. But then, tap was prevented for 600/700ms instead of my wished 400ms - so more after a full refresh, and when doing quick browsing with tap, I got stopped after a full refresh.
> 
> > if the refresh to be sent is flashing work instead?
> 
> Are you sure in there (uimanager.lua where I put the before/after stuff) from the mode that the refresh you send is going to block for >500ms vs the other at 1ms - on all devices?

If there's a mech_wait_update_complete set, it's going to block for around ~550ms after a flash, yeah.

NiLuJe:
```diff
diff --git a/ffi/framebuffer_mxcfb.lua b/ffi/framebuffer_mxcfb.lua
index 28a16e70..79a0a2a6 100644
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -113,16 +113,28 @@ end
 
 --[[ handlers for the wait API of the eink driver --]]
 
+-- Helper function to make sense of the return value from the mxcfb WAIT_FOR_UPDATE_COMPLETE ioctls,
+-- which is the amount of jiffies left in the timeout in the kernel's wait-for-completion handler.
+local function jiffies_to_ms(jiffies)
+    -- (jiffies * 1000 / USER_HZ)
+    -- USER_HZ (i.e., sysconf(_SC_CLK_TCK)) is pretty much always going to be 100 on these devices...
+    return jiffies * 10
+end
+
 -- Kindle's MXCFB_WAIT_FOR_UPDATE_COMPLETE_PEARL == 0x4004462f
 local function kindle_pearl_mxc_wait_for_update_complete(fb, marker)
     -- Wait for a specific update to be completed
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_PEARL, ffi.new("uint32_t[1]", marker))
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_PEARL, ffi.new("uint32_t[1]", marker))
+    -- Timeout is set to 5000ms
+    return 5000 - jiffies_to_ms(jiffies_left)
 end
 
 -- Kobo's MXCFB_WAIT_FOR_UPDATE_COMPLETE_V1 == 0x4004462f
 local function kobo_mxc_wait_for_update_complete(fb, marker)
     -- Wait for a specific update to be completed
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_V1, ffi.new("uint32_t[1]", marker))
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_V1, ffi.new("uint32_t[1]", marker))
+    -- Timeout is set to 10000ms
+    return 10000 - jiffies_to_ms(jiffies_left)
 end
 
 -- Kobo's Mk7 MXCFB_WAIT_FOR_UPDATE_COMPLETE_V3
@@ -133,7 +145,9 @@ local function kobo_mk7_mxc_wait_for_update_complete(fb, marker)
     -- NOTE: 0 seems to be a fairly safe assumption for "we don't care about collisions".
     --       On a slightly related note, the EPDC_FLAG_TEST_COLLISION flag is for dry-run collision tests, never set it.
     mk7_update_marker[0].collision_test = 0
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_V3, mk7_update_marker)
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_V3, mk7_update_marker)
+    -- Timeout is set to 5000ms
+    return 5000 - jiffies_to_ms(jiffies_left)
 end
 
 -- Pocketbook's MXCFB_WAIT_FOR_UPDATE_COMPLETE_PB... with a twist.
@@ -149,7 +163,10 @@ local function pocketbook_mxc_wait_for_update_complete(fb, marker)
     -- NOTE: 0 seems to be a fairly safe assumption for "we don't care about collisions".
     --       On a slightly related note, the EPDC_FLAG_TEST_COLLISION flag is for dry-run collision tests, never set it.
     update_marker[0].collision_test = 0
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_PB, update_marker)
+    -- NOTE: Return value potentially extremely unreliable on PB...
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_PB, update_marker)
+    -- Timeout is set to 5000ms
+    return 5000 - jiffies_to_ms(jiffies_left)
 end
 
 -- Remarkable MXCFB_WAIT_FOR_UPDATE_COMPLETE
@@ -160,19 +177,25 @@ local function remarkable_mxc_wait_for_update_complete(fb, marker)
     -- NOTE: 0 seems to be a fairly safe assumption for "we don't care about collisions".
     --       On a slightly related note, the EPDC_FLAG_TEST_COLLISION flag is for dry-run collision tests, never set it.
     update_marker[0].collision_test = 0
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, update_marker)
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, update_marker)
+    -- Timeout is set to 5000ms
+    return 5000 - jiffies_to_ms(jiffies_left)
 end
 
 -- Sony PRS MXCFB_WAIT_FOR_UPDATE_COMPLETE
 local function sony_prstux_mxc_wait_for_update_complete(fb, marker)
     -- Wait for a specific update to be completed
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, ffi.new("uint32_t[1]", marker))
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, ffi.new("uint32_t[1]", marker))
+    --- @fixme: Timeout is set to 5000ms?
+    return 5000 - jiffies_to_ms(jiffies_left)
 end
 
 -- BQ Cervantes MXCFB_WAIT_FOR_UPDATE_COMPLETE == 0x4004462f
 local function cervantes_mxc_wait_for_update_complete(fb, marker)
     -- Wait for a specific update to be completed
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, ffi.new("uint32_t[1]", marker))
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, ffi.new("uint32_t[1]", marker))
+    -- Timeout is set to 5000ms
+    return 5000 - jiffies_to_ms(jiffies_left)
 end
 
 -- Kindle's MXCFB_WAIT_FOR_UPDATE_COMPLETE == 0xc008462f
@@ -183,13 +206,17 @@ local function kindle_carta_mxc_wait_for_update_complete(fb, marker)
     -- NOTE: 0 seems to be a fairly safe assumption for "we don't care about collisions".
     --       On a slightly related note, the EPDC_FLAG_TEST_COLLISION flag is for dry-run collision tests, never set it.
     carta_update_marker[0].collision_test = 0
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, carta_update_marker)
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE, carta_update_marker)
+    -- Timeout is set to 5000ms
+    return 5000 - jiffies_to_ms(jiffies_left)
 end
 
 -- Kindle's MXCFB_WAIT_FOR_UPDATE_SUBMISSION == 0x40044637
 local function kindle_mxc_wait_for_update_submission(fb, marker)
     -- Wait for a specific update to be submitted
-    return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_SUBMISSION, ffi.new("uint32_t[1]", marker))
+    local jiffies_left = C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_SUBMISSION, ffi.new("uint32_t[1]", marker))
+    -- Timeout is set to 1000ms
+    return 1000 - jiffies_to_ms(jiffies_left)
 end
 
 
@@ -216,6 +243,9 @@ local function mxc_update(fb, update_ioctl, refarea, refresh_type, waveform_mode
 
     w = w * fb.refresh_pixel_size
 
+    -- Keep track of how long the various WAIT_FOR_UPDATE ioctls blocked for...
+    local blocked_for = 0
+
     -- NOTE: If we're trying to send a:
     --         * true FULL update,
     --         * GC16_FAST update (i.e., popping-up a menu),
@@ -231,7 +261,10 @@ local function mxc_update(fb, update_ioctl, refarea, refresh_type, waveform_mode
       or fb:_isUIWaveFormMode(waveform_mode))
       and (marker ~= 0 and marker ~= fb.dont_wait_for_marker) then
         fb.debug("refresh: wait for submission of (previous) marker", marker)
-        fb.mech_wait_update_submission(fb, marker)
+        local pre_submit_wait = fb.mech_wait_update_submission(fb, marker)
+        if pre_sub_wait > 0 then
+            blocked_for = blocked_for + pre_sub_wait
+        end
         -- NOTE: We don't set dont_wait_for_marker here,
         --       as we *do* want to chain wait_for_submission & wait_for_complete in some rare instances...
     end
@@ -248,7 +281,10 @@ local function mxc_update(fb, update_ioctl, refarea, refresh_type, waveform_mode
       and fb.mech_wait_update_complete
       and (marker ~= 0 and marker ~= fb.dont_wait_for_marker) then
         fb.debug("refresh: wait for completion of (previous) marker", marker)
-        fb.mech_wait_update_complete(fb, marker)
+        local pre_complete_wait = fb.mech_wait_update_complete(fb, marker)
+        if pre_complete_wait > 0 then
+            blocked_for = blocked_for + pre_complete_wait
+        end
     end
 
     refarea[0].update_mode = refresh_type or C.UPDATE_MODE_PARTIAL
@@ -319,10 +355,16 @@ local function mxc_update(fb, update_ioctl, refarea, refresh_type, waveform_mode
     if refarea[0].update_mode == C.UPDATE_MODE_FULL
       and fb.mech_wait_update_complete then
         fb.debug("refresh: wait for completion of marker", marker)
-        fb.mech_wait_update_complete(fb, marker)
+        local post_complete_wait = fb.mech_wait_update_complete(fb, marker)
+        if post_complete_wait > 0 then
+            blocked_for = blocked_for + post_complete_wait
+        end
         -- And make sure we won't wait for it again, in case the next refresh trips one of our wait_for_*  heuristics ;).
         fb.dont_wait_for_marker = marker
     end
+
+    -- Barring wait_for_*, return the ioctl's return value
+    return blocked_for > 0 and blocked_for or rv
 end
 
 local function refresh_k51(fb, refreshtype, waveform_mode, x, y, w, h)
@@ -523,7 +565,7 @@ end
 
 function framebuffer:refreshPartialImp(x, y, w, h, dither)
     self.debug("refresh: partial", x, y, w, h, dither and "w/ HW dithering")
-    self:mech_refresh(C.UPDATE_MODE_PARTIAL, self.waveform_partial, x, y, w, h, dither)
+    return self:mech_refresh(C.UPDATE_MODE_PARTIAL, self.waveform_partial, x, y, w, h, dither)
 end
 
 -- NOTE: UPDATE_MODE_FULL doesn't mean full screen or no region, it means ask for a black flash!
@@ -531,27 +573,27 @@ end
 --       That's regardless of whether the REAGL waveform mode is of the "always enforce FULL" variety or not ;).
 function framebuffer:refreshFlashPartialImp(x, y, w, h, dither)
     self.debug("refresh: partial w/ flash", x, y, w, h, dither and "w/ HW dithering")
-    self:mech_refresh(C.UPDATE_MODE_FULL, self.waveform_partial, x, y, w, h, dither)
+    return self:mech_refresh(C.UPDATE_MODE_FULL, self.waveform_partial, x, y, w, h, dither)
 end
 
 function framebuffer:refreshUIImp(x, y, w, h, dither)
     self.debug("refresh: ui-mode", x, y, w, h, dither and "w/ HW dithering")
-    self:mech_refresh(C.UPDATE_MODE_PARTIAL, self.waveform_ui, x, y, w, h, dither)
+    return self:mech_refresh(C.UPDATE_MODE_PARTIAL, self.waveform_ui, x, y, w, h, dither)
 end
 
 function framebuffer:refreshFlashUIImp(x, y, w, h, dither)
     self.debug("refresh: ui-mode w/ flash", x, y, w, h, dither and "w/ HW dithering")
-    self:mech_refresh(C.UPDATE_MODE_FULL, self.waveform_flashui, x, y, w, h, dither)
+    return self:mech_refresh(C.UPDATE_MODE_FULL, self.waveform_flashui, x, y, w, h, dither)
 end
 
 function framebuffer:refreshFullImp(x, y, w, h, dither)
     self.debug("refresh: full", x, y, w, h, dither and "w/ HW dithering")
-    self:mech_refresh(C.UPDATE_MODE_FULL, self.waveform_full, x, y, w, h, dither)
+    return self:mech_refresh(C.UPDATE_MODE_FULL, self.waveform_full, x, y, w, h, dither)
 end
 
 function framebuffer:refreshFastImp(x, y, w, h, dither)
     self.debug("refresh: fast", x, y, w, h, dither and "w/ HW dithering")
-    self:mech_refresh(C.UPDATE_MODE_PARTIAL, self.waveform_fast, x, y, w, h, dither)
+    return self:mech_refresh(C.UPDATE_MODE_PARTIAL, self.waveform_fast, x, y, w, h, dither)
 end
 
 -- Detect Allwinner boards. Those emulate mxcfb API in a custom driver (poorly).
```

And

```diff
diff --git a/frontend/ui/uimanager.lua b/frontend/ui/uimanager.lua
index 409d702f..62dbd577 100644
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -1094,6 +1094,7 @@ function UIManager:_repaint()
     end
 
     -- execute refreshes:
+    local blocked_for = 0
     for _, refresh in ipairs(self._refresh_stack) do
         -- Honor dithering hints from *anywhere* in the dirty stack
         refresh.dither = update_dither(refresh.dither, dithered)
@@ -1127,12 +1128,19 @@ function UIManager:_repaint()
             refresh.region.w = ALIGN_UP(refresh.region.w + (x_fixup * 2), 8)
             refresh.region.h = ALIGN_UP(refresh.region.h + (y_fixup * 2), 8)
         end
-        Screen[refresh_methods[refresh.mode]](Screen,
+        local rv = Screen[refresh_methods[refresh.mode]](Screen,
             refresh.region.x, refresh.region.y,
             refresh.region.w, refresh.region.h,
             refresh.dither)
+        -- NOTE: Only framebuffer_mxcfb's mech_refresh returns positive values, corresponding to the amount of time (in ms)
+        --       the flashing update blocked for.
+        if rv and rv > 0 then
+            blocked_for = blocked_for + rv
+            logger.dbg("Refresh blocked for", rv, "ms")
+        end
     end
-    Screen:afterPaint()
+    logger.dbg("Refresh queue blocked for", blocked_for, "ms")
+    Screen:afterPaint(blocked_for)
     self._refresh_stack = {}
     self.refresh_counted = false
 end
```

Appears to behave on my end, and passes that value to `Screen:afterPaint`, which you might be able to piggyback on for this?

poire-z:
Minor typo at:
```
+        local pre_submit_wait = fb.mech_wait_update_submission(fb, marker)
+        if pre_sub_wait > 0 then
```

> Screen:afterPaint, which you might be able to piggyback on for this?

I'll see about that tomorrow. (Input passing a afterRepaintCallback() to fb/Screen when device caps feels like adjustment needed?)

Another issue that this specific refresh handling does not solve is all the things that takes cpu time blocking (dict lookup, some ConfirmBox somewhere I can't remember which that I was dismissing unwillingly often...) My UIManager:discardEventFor(1s) was somehow fine with these.
So, still wondering if there's a more generic way to handle all these cases...

NiLuJe:
> Another issue that this specific refresh handling does not solve is all the things that takes cpu time blocking (dict lookup, some ConfirmBox somewhere I can't remember which that I was dismissing unwillingly often...) My UIManager:discardEventFor(1s) was somehow fine with these.

That's admittedly another kettle of fish... ;).

> So, still wondering if there's a more generic way to handle all these cases...

I'm not a huge fan of a generic approach, because it's very much *not* a generic issue: it happens *because* of the crap Kobo workarounds.

poire-z:
> I'm not a huge fan of a generic approach here, because it's very much not a generic issue: it happens because of the crap Kobo workarounds.

Ok, so, a specific approach at solving this specific issue in a more generic specific way :) 

But ... all of this is because we use TimeVal:now() when the ev.time we get is the time since boot.
I just checked that this happens too with my Alyssum GloHD:
```
10/18/20-00:42:42 WARN  alyssum 1602974562 ev: {
    ["type"] = 3,
    ["value"] = 87,
    ["time"] = {
        ["sec"] = 33,
        ["usec"] = 562278
    },
    ["code"] = 53
}
10/18/20-00:42:42 WARN  alyssum 1602974562 ev: {
    ["type"] = 3,
    ["value"] = 0,
    ["time"] = {
        ["sec"] = 33,
        ["usec"] = 562281
    },
    ["code"] = 24
}
```

But, in this business of event handling, do we really care about epoch time and matching os.time() ?
It looks to me we're only computing time differences, that would be the same no matter the origin (epoch or boot time).
(May be some issue with suspend/resume if the boot time doesn't advance - but I don't really see why - first tap at resume prevented by Tap on Suspend meu item by double-tap handling ? :)

Any thoughts ? Something I don't see?

NiLuJe:
@poire-z: That thought has indeed crossed my mind ;).

Not having an affected device, I don't really know, but I'd tend to agree with you that it feels a tad heavy-handed.

I don't *think* Plato does anything of the sort, for example.

poire-z:
Mhh, just commenting this out makes my GloHD all messy :) Taps generating dict long-press or being ignored...
```diff
 local adjustTouchAlyssum = function(self, ev)
-    ev.time = TimeVal:now()
+    -- ev.time = TimeVal:now()
     if ev.type == EV_ABS and ev.code == ABS_MT_TRACKING_ID then
         ev.value = ev.value - 1
     end
 end
```

poire-z:
From https://github.com/koreader/koreader/pull/1527#discussion_r30461109 : 
> ev.time = TimeVal:now() changes the time stamp in events so that they have the same format as the tv_now in Input:waitEvent (which is also obtained by TimeVal:now()). We have discussed this. This will make the time of the events not so accurate, but I don't see a problem, and I don't know another way to solve it..

Also more details in: https://github.com/koreader/koreader/issues/1525#issuecomment-101973618 (I'll read that tomorrow - unless you beat me to it :)

NiLuJe:
At a *very* quick glance, that's because of the whole `setTimeout`/`timer_callbacks` thingy in `waitEvent`.

Not *quite* sure how this all works, but I suppose one could mangle that into being fully relative to the last input event's timestamp instead of aligning to now's epoch?

poire-z:
Or just computing at start up and onResume a kernelTimeToEpochTime constant that we'd just add to ev.time (time ellapsed since boot) to get a time since epoch.
Any other possibiity for kernel time since boot to pause except on Suspend/Resume ? (except for my own private auto-standby that does echo standby > /sys/power/state) ?

Is there a generic way (/proc /sys or some C syscall) to get the "current kernel time ellapsed since boot" ?

NiLuJe:
`clock_gettime` can output a variety of "since boot" timestamps.

NiLuJe:
Okay, I still had the tests on my H2O :D.

The `_COARSE` variants are indeed faster, but only as precise as USER_HZ (i.e., 10ms).

The precision of `CLOCK_MONOTONIC_RAW` is unknown (the call returns garbage on my H2O), but is likely slightly lower than `CLOCK_MONOTONIC`'s 1ns.

Other than than, perf is roughly identical between the `_COARSE` variants and the others, so, it basically just depends on which one you need ;).

NiLuJe:
Random question: why do we even care about suspend/resume? Shouldn't we just drop the slots and essentially reset the state on suspend?

poire-z:
> Random question: why do we even care about suspend/resume? Shouldn't we just drop the slots and essentially reset the state on suspend?

I didn't think about caring much more than that, but _just drop the slots and essentially reset the state on suspend_ is caring :)
But may be we don't have to do all that:

> At a very quick glance, that's because of the whole setTimeout/timer_callbacks thingy in waitEvent.
Not quite sure how this all works, but I suppose one could mangle that into being fully relative to the last input event's timestamp instead of aligning to now's epoch?

Well, looks like there are only 2 things using setTimeout, and instead of mangling all events to use TimeVal:now(), we can just have these 2 setTimeout use TimeVal:now() instead of ev.timev.

So, it's just when handling double tap and hold/long-press that we would have a possibly shift between the even timestamp and the now() used for the timeout.
When everything is not busy, these should be quite similar and this change should have no impact.
And when we long-press or do double tap, things shouldn't be busy (when double tap, the first tap does not trigger anything so no reason to get busy).
Looks ok on my GloHD, i'll be testing that today.
If that looks fine, we could get rid of all the `self.touch_probe_ev_epoch_time` and `probeEvEpochTime`, in the Kobo, Cervantes and sony-prstux device.lua.
Devices that use that are:
```
Kobo Touch (touch_probe_ev_epoch_time)
Kobo Aura One (touch_probe_ev_epoch_time)
Kobo Aura H2O2 (touch_probe_ev_epoch_time)
Kobo Aura second edition (touch_probe_ev_epoch_time)
Kobo Aura second edition, Rev 2 (touch_probe_ev_epoch_time)

Kobo Glo HD (touch_alyssum_protocol)
Kobo Touch 2.0 (touch_alyssum_protocol)

All Cervantes (touch_probe_ev_epoch_time)
Sony PRSTUX (touch_probe_ev_epoch_time)
```

ezdiy:
@NiLuJe
 
>jiffies_to_ms

No such thing on B288. For platforms that need to serialize refreshes, it might be even worth it to do this in a worker thread. Hogging up the main thread for such long periods introduces a host of trouble in other places too (essentially anything that needs precise timers) - not just in touch detection.

poire-z:
> in addition to forcing the poor precision on platforms where the timestamp works ok and touch_probe_ev_epoch_time hack isn't enabled

I understand. Then, I could in the last commit, instead of:
```
-        local deadline = tev.timev + TimeVal:new{
+        local deadline = TimeVal:now() + TimeVal:new{
```
use:
```
        local ref_time = TimeVal:now()
        if tev.timev.sec > ref_time.sec - 600 then
            -- event time is not an uptime, and is usable
            ref_time = tev.timev
        end
        local deadline = ref_time + TimeVal:new{
```

(I understand _most_ of everything you wrote :) but this being ultrasensitive, the less I touch it, the less I'll mess that up on any device :) I'll let any refactoring of that kind to a more adventurous person :)

Anyway, I still have some issues with swipe sometimes being detected as multiswipe... (small diagonal swipe to full refresh, working sometimes, sometimes not...)

ezdiy:
@poire-z 

I don't see why you're erasing the timev value in the first place - that value is synthesized exactly as you do by device frontend driver workaround when it needs be - and if not, fix those :)

Here is how erased timestamps behave on my device - when there is some complex draw operation and two taps happen (try this with high dpi cbz), the bump detection will consider those two spaced taps as one, as synthesized timestamp will both arrive as "now". With real timestamps (that arrive in the past in this case), it behaves better because the timeout/slot it installs immediately expires and the second tap processed next round registers properly. Note that this doesn't all that reliably because the tv_diff comparisons can't handle deltas larger than 1 second, but in the usual case its fine :)

poire-z:
> I don't see why you're erasing the timev value in the first place

But I'm not, and now not anymore by getting rid of the alysumm/proveEv stuff: we let it be the kernel uptime (or whatever it is).
It's just there in the setTimeout() stuff, that ends up into Input:waitEvent(), where we need to compute a time relative to the real clock time, that we can't use the relative kernel uptime.

> that value is synthesized exactly as you do by device frontend driver workaround when it needs be - and if not, fix those :)
>  (i think you're actually removing that workaround for kobo, not sure why that is either).

Yes, because these workarounds make them ALL imprecise - even where comparing relative timestamps would be just fine. So, removing them make them ALL precise, except in the rare places where we need a real clock time comparisons - and the question is: would we be still a bit better with that or not ?

ezdiy:
@poire-z Ah, makes sense. Indeed then yes, https://github.com/koreader/koreader/pull/6798#issuecomment-711149184 works to cater to both cases - both timeouts (that can use precise ts if available), and relative swipes that can make use of broken but precise ts differences.

poire-z:
> works to cater to both cases 

Done in last commit.
Is there a better name that "real clock time" to name the good time type and the variable?

poire-z:
> Anyway, I still have some issues with swipe sometimes being detected as multiswipe... (small diagonal swipe to full refresh, working sometimes, sometimes not...)

I can reproduce these misses with the code before my changes... and logically, I don't see how these could happen more often with the changes.
So, I'll assume I'm dizzy and not drawing straight swipes :)

NiLuJe:
> @NiLuJe
> 
> > jiffies_to_ms
> 
> No such thing on B288 (they serialize updates in-kernel, and the ioctl wait is typically instant return). Might be the case for reagl too.

Yeah, I know, hence the comment ;). This was more of a quick PoC.
(REAGL has no impact on this on Kindle/Kobo, FWIW. Hell, you can wait for a non-flashing A2 update just fine, and the timing will be accurate, too).

> For platforms that need to serialize the waits on markers, it might be even worth it to do this in a worker thread. Hogging up the main thread for such long periods introduces a host of trouble in other places too (essentially anything that needs precise timers) - not just in touch detection (though gesturedetector currently hogs the main thread too as it isn't asynchronous either).
> 

It *needs* to block blitting, though, or blocking serves no purpose whatsoever ;). So, even assuming threading weren't a gigantic PITA in Lua, that would be much easier said than done ;).

NiLuJe:
@poire-z: If you get rid of the Alyssum hack, you can probably also ditch the whole `touch_probe_ev_epoch_time` mechanism, which essentially does the very thing you wiped off Alyssum ;).

NiLuJe:
> > works to cater to both cases
> 
> Done in last commit.
> Is there a better name that "real clock time" to name the good time type and the variable?

wall-clock, which isn't necessary any clearer for non CS persons ^^.

poire-z:
> you can probably also ditch the whole touch_probe_ev_epoch_time mechanism

Done :)
So, we're all fine with the idea of doing that "probe epoch time" only in GestureDetector, for the two timers (long press + double tap) that use setTimeout() ?

NiLuJe:
It needs to consume an input event to make a decision, so, eh, that looks fine by me?

ezdiy:
@poire-z lgtm.

@NiLuJe 

>It *needs* to block blitting, though, or blocking serves no purpose whatsoever ;). So, even assuming threading weren't a gigantic PITA in Lua, that would be much easier said than done ;).

I think best we could do ATM is actually check for collisions - fire up update & and just remember marker - and roll forward with no wait. When we try to fire up next update, and it collides, the kernel *should* signal us the collision, including the marker it is colliding with we should now wait on - only in that case we should start wait for it in a blocking call. This would complicate mxbc code somewhat, and I'm not sure if it would get rid of blocking all that much, as nature of collisions is highly irregular (shared 16x16 pixel blocks, available pipelines...).

pazos (approved):
Nothing to say. I leave comments to clever guys :)

About myself testing in cervantes I'm a bit hessitant (I'm still running a heavy modified version 2020.08.?), but I will test it when it lands master, ofc.

NiLuJe:
@ezdiy: The EPDC already does most of that much better than we could ever hope for, which is why I'm quite fine with the "wait before/after a flash" fire and forget approach ;).

NiLuJe:
I've also never actually seen anything making use the `TEST_COLLISION` feature, and timing issues make me wonder if it'd actually be an improvement ;)...

NiLuJe:
Fun fact: Haven't checked older kernels, but the EPDCv2 driver already internally checks for completion of all previous updates on anything flagged `FULL` :).

NiLuJe:
By which I mean, the main source of *actual* block will be the wait_for_complete *after* a flashing update, and [this comment](https://github.com/koreader/koreader-base/blob/78204b02feb4e7b00d110626634eb8c00e1d503b/ffi/framebuffer_mxcfb.lua#L310-L318) explains this choice ;).

NiLuJe:
> > clock_gettime
> 
> This is always very slow unless it calls gettimeofday (which is just VDSO read, ie very fast). gtod is used for _REALTIME and _COARSE.

Depends on the kernel version ;). On the Forma (4.1), basically everything is provided by the vdso (`gettimeofday` *and* `clock_gettime` w/ `CLOCK_REALTIME_COARSE, CLOCK_MONOTONIC_COARSE, CLOCK_REALTIME, CLOCK_MONOTONIC`. On the H2O (2.6.35), *nothing* is implemented via the vdso, not even `gettimeofday`!

poire-z:
Annoying issue with having a Tap interval 400ms (which is perfect for me for page turning):
now, I can't type fast on the keyboard :/ I'm missing too much letters, so I'm forced to type slow...

If i'd find a way to ignore that interval when the keyboard is shown (haven't looked yet), would it be reasonable to have Tap interval not ensured there - and lying a bit - or would some people really want to have bounce avoidance with they keyboard too?

NiLuJe:
Do you really get real bounces at 400ms intervals?

Otherwise, well, that'd be kinda like replicating mechanical chatter, so, yay?

Joking aside, as far as I'm concerned, it's set to 0, so, if you find an unintrusive way of doing that, why not?

poire-z:
> Do you really get real bounces at 400ms intervals?

Well, may be not. But with my previous use of `rate = 1.5` in the handler for Tap in ReaderRolling, I got used to sometimes quick-multi-tap'ing out of nervosity/frustration/joy and expecting a single page forward :) 
Tried with 200ms earlier, and it's too low. Also, 400ms limits the rate of page turn and avoid the competitions of refreshes when you turn super fast (that you can do with 0ms) - that I find quite ugly, and I don't really need this super fast skimming :)

NiLuJe:
> 400ms limits the rate of page turn and avoid the competitions of refreshes when you turn super fast (that you can do with 0ms) - that I find quite ugly, and I don't really need this super fast skimming :)

Okay, that I can get behind ;). (FWIW, that's no longer an issue on REAGL devices, because there's better separation between pages without requiring a flash).

plugins/gestures.koplugin/main.lua
```diff
@@ -466,24 +468,55 @@ Default value: %1]]), Screen.low_pan_rate and 5.0 or 30.0),
...
+                keep_menu_open = true,
+                callback = function()
+                    local SpinWidget = require("ui/widget/spinwidget")
+                    local GestureDetector = require("device/gesturedetector")
+                    local items = SpinWidget:new{
+                        title_text = _("Tap interval"),
+                        info_text = T(_([[
+Any other taps made within this interval after a first tap will be considered accidental and ignored.
+
+The interval value is in milliseconds and can range from 0 (0 second) to 2000 (2 seconds).
```
Frenzie, line 482: 0 seconds

PR #6799, opened 2020-10-17
Don't shadow gettext

NiLuJe:
Fixes the "current book" stats page.

(Regression in #6796)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6799)


ezdiy:
Oops, force of habit (and rather unfortunate of GT to collide with lua convention). Still, use `-- luacheck: ignore 211` to retain the original intent.

NiLuJe:
Random related comment: where the hell is the ID-to-error mapping hidden for LuaCheck? Every time I wonder where those magic number come from, i just give up and use a global plain ignore ;).

NiLuJe:
https://luacheck.readthedocs.io/en/stable/warnings.html

Well, that wasn't so hard :D

poire-z:
> Add a hint that we do want to ingore the first arg 

Isn't this already taken care of by:
https://github.com/koreader/koreader/blob/893909146d6e406ca36c5f2d95b82526a4ed6167/.luacheckrc#L143-L152

PR #6801, opened 2020-10-18
Random nano optimizations

NiLuJe:
* Avoid spurious Screen:afterPaint calls
* Cache os.time() calls where it makes sense to do so

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6801)

### DIFF
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -927,11 +927,11 @@ function ReaderView:checkAutoSaveSettings()
     if not interval then -- no auto save
         return
     end
-    if os.time() - self.settings_last_save_ts >= interval*60 then
-        self.settings_last_save_ts = os.time()
+    local now_ts = os.time()
+    if now_ts - self.settings_last_save_ts >= interval*60 then
+        self.settings_last_save_ts = now_ts
         UIManager:nextTick(function()
             self.ui:saveSettings()
-            self.settings_last_save_ts = os.time() -- re-set when saving done
         end)
     end
 end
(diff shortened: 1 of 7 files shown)


frontend/apps/reader/modules/readerview.lua
```diff
@@ -927,11 +927,11 @@ function ReaderView:checkAutoSaveSettings()
...
         return
     end
-    if os.time() - self.settings_last_save_ts >= interval*60 then
-        self.settings_last_save_ts = os.time()
+    local now_ts = os.time()
+    if now_ts - self.settings_last_save_ts >= interval*60 then
+        self.settings_last_save_ts = now_ts
         UIManager:nextTick(function()
             self.ui:saveSettings()
-            self.settings_last_save_ts = os.time() -- re-set when saving done
```
NiLuJe, line 934: The delay to the next tick, plus the actual time it took to save should be meaningless in the grand scheme of things, hence the deletion.

plugins/autosuspend.koplugin/main.lua
```diff
@@ -83,11 +84,11 @@ function AutoSuspend:_schedule()
         UIManager:suspend()
     else
         if self:_enabled() then
-            logger.dbg("AutoSuspend: schedule suspend at ", os.time() + delay_suspend)
+            logger.dbg("AutoSuspend: schedule suspend in", delay_suspend)
```
NiLuJe, line 87: Logger entries are already timestamped, and the scheduled action itself will also trigger a log entry, so it feels like simply loading the actual scheduleIn delay is actually more useful?

NiLuJe:
*Proudly splitting hairs since 1988!* :D.

poire-z (approved):

Frenzie (approved):
Sure. ^_^

PR #6804, opened 2020-10-19
Address some DB migration concerns

NiLuJe:
* Don't stomp on an existing backup
* Try to handle inconsistent DBs better

Re: #6778

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6804)

### DIFF
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -520,8 +520,8 @@ function ReaderStatistics:upgradeDB(conn)
         --       We'll just have to hope the current value of the column pages in the book table is not too horribly out of date,
         --       and not too horribly out of phase with the actual page count at the time the data was originally collected...
         INSERT INTO page_stat_data
-            SELECT id_book, page, start_time, duration, IFNULL(pages, 0) as total_pages FROM page_stat
-            LEFT JOIN book on book.id = id_book;
+            SELECT id_book, page, start_time, duration, pages as total_pages FROM page_stat
+            JOIN book on book.id = id_book;
 
         -- Drop old page_stat table
         DROP INDEX IF EXISTS page_stat_id_book;
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -15,8 +15,6 @@ local Screensaver = require("ui/screensaver")
 local SQ3 = require("lua-ljsqlite3/init")
 local UIManager = require("ui/uimanager")
 local Widget = require("ui/widget/widget")
-local ffi = require("ffi")
-local C = ffi.C
 local lfs = require("libs/libkoreader-lfs")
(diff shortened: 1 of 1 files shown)


plugins/statistics.koplugin/main.lua
```diff
@@ -306,7 +308,21 @@ Do you want to create an empty database?
             logger.info("ReaderStatistics: Migrating DB from schema", db_version, "to schema", DB_SCHEMA_VERSION, "...")
             -- Backup the existing DB first
             conn:close()
-            local bkp_db_location = db_location .. ".bkp." .. db_version .. "-to-" .. DB_SCHEMA_VERSION
+            local bkp_base_location = db_location .. ".bkp." .. db_version .. "-to-" .. DB_SCHEMA_VERSION
+            local bkp_db_location = bkp_base_location
+            -- Don't overwrite an existing backup
+            local bkp_rev = 1
+            while C.access(bkp_db_location, C.F_OK) == 0 do
```
poire-z, line 315: What's wrong with our usual `lfs.attributes(file, "mode") == "file"` ? :)

I'd say that if there is one that exists, and if we're re-migrating (so, because something failed and the current db is bad), just don't do anything - no need to keep backups of bad dbs. ?

NiLuJe: > What's wrong with our usual `lfs.attributes(file, "mode") == "file"` ? :)
> 

I wasn't super-convinced it wouldn't do weird things in a while loop, so I stuck with what I knew would work ;p.

> I'd say that if there is one that exists, and if we're re-migrating (so, because something failed and the current db is bad), just don't do anything - no need to keep backups of bad dbs. ?

It's indeed an *extremely* defensive approach, true.
Then again, it's still useful to debug *what* exactly failed to migrate ;).

NiLuJe: > > What's wrong with our usual `lfs.attributes(file, "mode") == "file"` ? :)
> 
> I wasn't super-convinced it wouldn't do weird things in a while loop, so I stuck with what I knew would work ;p.
> 

Okay, it's essentially just a wrapper around `stat`, so, that works ;). A tiny bit overkill, but safe.

NiLuJe: > > I'd say that if there is one that exists, and if we're re-migrating (so, because something failed and the current db is bad), just don't do anything - no need to keep backups of bad dbs. ?
> 
> It's indeed an _extremely_ defensive approach, true.
> Then again, it's still useful to debug _what_ exactly failed to migrate ;).

Well, there's the actual botched migrated db for that, duh'. >_<".

NiLuJe: (Yes, I'm talking to myself, don't mind me, I haven't slept much ^^)

poire-z (approved):

PR #6807, opened 2020-10-21
Statistics: speed up Calendar view

poire-z:
Just by tweaking the SQL queries.
Also, in the hourly histogram, show at least a 1px bar when there was any reading this hour.

I always thought the slowness of showing and browsing the calendar views was caused by the many little widgets to size and draw, but no, it was really the SQL queries (and it was as slow before the recent refactoring).
On my Kobo, each of the 2 SQL queries used to take 500ms (and the UI drawing around 20ms)- and it was quite noticable.
With these tweaks, each SQL query take around 100ms.

(By adding an index `CREATE INDEX zzz ON page_stat_data(start_time)`, we'd get even better results, 50ms + 30ms - but well, it makes the DB file size grow by 30%, and it's not worth again upgrading the schema.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6807)

### DIFF
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -2299,8 +2299,8 @@ function ReaderStatistics:getReadingRatioPerHourByDay(month)
             strftime('%H', start_time, 'unixepoch', 'localtime') hour,
             sum(duration)/3600.0 ratio
         FROM   page_stat
-        WHERE  start_time >= strftime('%s', ?, 'utc')
-          AND  start_time < strftime('%s', ?, 'utc', '33 days', 'start of month')
+        WHERE  start_time BETWEEN strftime('%s', ?, 'utc')
+                              AND strftime('%s', ?, 'utc', '+33 days', 'start of month', '-1 second')
         GROUP  BY
             strftime('%Y-%m-%d', start_time, 'unixepoch', 'localtime'),
             strftime('%H', start_time, 'unixepoch', 'localtime')
@@ -2333,8 +2333,8 @@ function ReaderStatistics:getReadBookByDay(month)
             book.title book_title
         FROM   page_stat
         JOIN   book ON book.id = page_stat.id_book
-        WHERE  start_time >= strftime('%s', ?, 'utc')
-          AND  start_time < strftime('%s', ?, 'utc', '33 days', 'start of month')
+        WHERE  start_time BETWEEN strftime('%s', ?, 'utc')
+                              AND strftime('%s', ?, 'utc', '+33 days', 'start of month', '-1 second')
         GROUP  BY
             strftime('%Y-%m-%d', start_time, 'unixepoch', 'localtime'),
             id_book,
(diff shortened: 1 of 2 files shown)


NiLuJe:
It's never hit a stable yet, and we've got the migration down to pat now, so I don't really have an issue with updating the schema, FWIW ;).

plugins/statistics.koplugin/main.lua
```diff
@@ -2285,21 +2285,30 @@ function ReaderStatistics:getFirstTimestamp()
...
+    -- month; we need the start of the month to be a real date:
+    month = month.."-01"
     local sql_stmt = [[
         SELECT
             strftime('%Y-%m-%d', start_time, 'unixepoch', 'localtime') day,
             strftime('%H', start_time, 'unixepoch', 'localtime') hour,
             sum(duration)/3600.0 ratio
         FROM   page_stat
-        WHERE  strftime('%Y-%m', start_time, 'unixepoch', 'localtime') = ?
+        WHERE  start_time >= strftime('%s', ?, 'utc')
```
NiLuJe, line 2302: a `WHERE start_time BETWEEN x AND y` would possibly be a smidge faster ;).

poire-z: Done. Not tested if it changes speed.

poire-z:
OK, added the creation of that index.
A bit sad that our valuable nightly users won't get it (while our lazy or fearful :) stable users will), unless they add it manually with:
```
$ sqlite3 statistics.sqlite3
sqlite> CREATE INDEX IF NOT EXISTS page_stat_data_start_time ON page_stat_data(start_time);
sqlite> VACUUM;
```

Also added a VACUUM after migration to reclaim the space of the old deleted table page_stat. It's always a bit worrying as it may fail (but the migration too) depending on disk space or memory usage - but at least, the migration happen after a KOReader upgrade, so when the process is just launched, so probably lots of RAM available (more than after running a while).
(My DB size went from 7M to 4M with this VACUUM.)

NiLuJe:
With a bit of massaging, we could arguably make `upgradeDB` safe to re-run, which would allow bumping the schema version and adding that index to the migration, too.

poire-z:
(Not really wanting to spend time on multiple/iterative upgrades, so I'll leave that to you if you feel like it.
May be the safest would be to have a list of user_version tags, and just split the upgrade steps in various
```
function updateDBto20201010 ... end
function updateDBto20201021 ... end
```
and call the functions as needed till we reach the current DB_SCHEMA_VERSION.)

NiLuJe:
> (Not really wanting to spend time on multiple/iterative upgrades, so I'll leave that to you if you feel like it.
> May be the safest would be to have a list of user_version tags, and just split the upgrade steps in various
> 
> ```
> function updateDBto20201010 ... end
> function updateDBto20201021 ... end
> ```
> 
> and call the functions as needed till we reach the current DB_SCHEMA_VERSION.)

Yup! Feel free to merge then, I'll look into it later tonight ;).

poire-z:
It's all yours, good luck :)

Issue #6527, opened 2020-08-19
[Kobo] Charging led does not light up


SilverGreen93:
* KOReader version: 2020.08
* Device: Kobo Forma

#### Issue
When plugging in the usb charger, the charging led does not light up even when the device is sleeping, so there is no indication of the chraging state.

#### Steps to reproduce
Open KOReader.
Put the device to sleep.
Plug in the USB charger.
No indication of charging is shown.
Charging LED does not light up.

pazos:
IMO too much work for litle benefit. Patches welcome

NiLuJe:
We've already established that playing with the LEDs led to... unwanted and weird behavior in weird and interesting ways, so, yeaaah, not happening ;).

NiLuJe:
(I can't *quite* recall the issue where we played with this last, but it was a couple years ago).

SilverGreen93:
Maybe at least display on the lockscreen the charging message and update it accordingly?

NiLuJe:
I would think the giant cable hanging from the device would be obvious enough?

(We don't update the screen on suspend, so having a live tracking % is not happening either).

Frenzie:
That means it either doesn't actually sleep while charging or that it wakes up every x minutes to update the status.

NiLuJe:
And that's potentially device-specific. Mk. 7 might be able to suspend while charging, like they do when USBNet is enabled.

While earlier devices will actually fail to sleep.

(FWIW, Nickel itself prevents suspend entirely when charging, on every generation).

SilverGreen93:
> I would think the giant cable hanging from the device would be obvious enough?
> 
> (We don't/can't update the screen on suspend, so having a live tracking % is not happening either).

No, the cable connected to the device does not mean it's charging...
On my first Kobo device I had a defective microUSB connector and didn't know that it wasn't charging, so I replaced the defective unit. It was charging only if the connector was held at a specific angle. Cables can get broken etc...

Yes, Nickel wakes during charging and updates the screen.

NiLuJe:
> No, the cable connected to the device does not mean it's charging...
> On my first Kobo device I had a defective microUSB connector and didn't know that it wasn't charging, so I replaced the defective unit. It was charging only if the connector was held at a specific angle. Cables can get broken etc...
> 

That's veering into extreme edge-casey behavior ^^.

NiLuJe:
[...]

Issue #5479, opened 2019-10-11
Feature Request: Enable the builtin led light ( if available ) during shutdown


BrainDamage:
* KOReader version:  v2019.09.3-64g5209494_2019-10-10
* Device: Kobo Aura H2O2

#### Issue
When issuing a shutdown, the screen immediately switches to the power off status while the shutdown procedure is still running for a while, therefore, if someone tries to press the on button again, he'd find it non-functional for a while.
At least in my device there's present a led which is lit during boot, which helps knowing that the device is working and not just dead. Setting it on before initializing the shutdown procedure should suffice, since once the device is powered off it'll naturally turn off as well the light.

Frenzie:
I don't think Nickel does that on shutdown, does it? Curiously, I've never investigated the light.

BrainDamage:
That is correct, I consider it a deficiency on Nickel as well

pazos:
Probably a **cannot fix**. You're describing a known hardware limitation of netronix based ereaders. (They have two cores, one is the ARM SoC where linux run and the other one is based on MSP430 which controls the power/battery and most of the hardware and exposes stuff to linux via i2c)

Even when we can turn on the light at poweroff this won't change the thing because the led will be turned off a few seconds before the reset circuit becomes available again (which happens about 500ms after the msp430 PMIC receives a reset command from the linux core).

This affects Kobos, Cervantes and most Freescale/NTX e-books out there. I guess it will also affect Kindles (and maybe that's one of the reasons why lab126 choose to avoid poweroff).

Please test it on your device and tell us: turn the light on yourself (it is exposed as sysfs node) and issue a `poweroff`. I may be wrong.

NiLuJe:
For reference:

On:

```
echo "ch 4" >| /sys/devices/platform/pmic_light.1/lit && echo "cur 1" >| /sys/devices/platform/pmic_light.1/lit && echo "dc 1" >| /sys/devices/platform/pmic_light.1/lit
```

Off:

```
echo "ch 4" >| /sys/devices/platform/pmic_light.1/lit && echo "cur 0" >| /sys/devices/platform/pmic_light.1/lit && echo "dc 0" >| /sys/devices/platform/pmic_light.1/lit
```

BrainDamage:
I just tested it now through ssh and it worked like intended:
[root@(none) ~]# echo 'dc 1' > /sys/devices/platform/pmic_light.1/lit && poweroff
the led lit, and shut down at the same time as the screen backlight

[...]

PR #6810, opened 2020-10-21
[Kobo, Mk. 7] Enable the power LED when charging

NiLuJe:
What it says on the tin ;).

Re #6527 

c.f., #5479 for more details about why it's not enabled anywhere else.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6810)


frontend/device/kobo/device.lua
```diff
@@ -740,6 +750,62 @@ function Kobo:toggleGSensor(toggle)
...
+        return false
+    end
+
+    -- c.f., strace -fittvyy -e trace=ioctl,file,signal,ipc,desc -s 256 -o /tmp/nickel.log -p $(pidof -s nickel) &
+    -- This was observed on a Forma, so I'm mildly hopeful that it's safe on other Mk. 7 devices ;).
+    if toggle == true then
+        -- NOTE: Technically, Nickel forces a toggle off before that, too.
+        --       But since we do that on startup, it shouldn't be necessary here...
+        f:write("ch 4")
+        f:flush()
```
pazos, line 780: Do you need to flush after each write?, if so, wouldn't be more legible to write `f:write("ch4"):flush()`

poire-z (approved):

frontend/device/kobo/device.lua
NiLuJe: They are individual commands, so, yeah, I assume they really need to be each their own separate write syscall.

(That's certainly how it happens in Nickel).

pazos: > That doesn't actually work ;p. (io.write doesn't actually return the I/O object)

Duh, sorry for suggesting without testing :p

NiLuJe: Eh, seemed sensible enough ;).

PR #6811, opened 2020-10-21
android: handle onDestroy

pazos:
We need it to exit the lua state and return from android_main. This avoid a *zombie state** on the following scenarios:

- App is killed by the activity manager  (ie: battery optimizations or OOM)
- We call finish() from Kotlin

**zombie state**: app is running, does nothing, won't easily trigger ANRs and needs to be killed manually by the user

In the real world the only related bug I found was on android10+, when running KO ocassionally and leaving the activity in the background for a week. The app is then destroyed by the system to "save battery".

As android has a way to "upgrade" or "downgrade" app battery optimization based on usage, YMMV.

Requires https://github.com/koreader/koreader-base/pull/1224

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6811)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -134,6 +134,9 @@ function Device:init()
             logger.dbg("Android application event", ev.code)
             if ev.code == C.APP_CMD_SAVE_STATE then
                 return "SaveState"
+            elseif ev.code == C.APP_CMD_DESTROY then
+                self:exit()
+                os.exit()
             elseif ev.code == C.APP_CMD_GAINED_FOCUS
                 or ev.code == C.APP_CMD_INIT_WINDOW
                 or ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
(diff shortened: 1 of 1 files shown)


pazos:
There's no much to review here, but I will wait for an approval :smile:

frontend/device/android/device.lua
```diff
@@ -134,6 +134,9 @@ function Device:init()
             logger.dbg("Android application event", ev.code)
             if ev.code == C.APP_CMD_SAVE_STATE then
                 return "SaveState"
+            elseif ev.code == C.APP_CMD_DESTROY then
+                self:exit()
+                os.exit()
```
NiLuJe, line 139: Would a UIManager:quit or whatever the Exit menu does make sense here?

pazos: Yeah, makes all the sense. It is also slightly more verbose :)

poire-z (approved):

PR #6812, opened 2020-10-21
Statistics: Handle DB migration to 20201022

NiLuJe:
By chunking DB migration steps per schema revision, as discussed in #6807

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6812)


poire-z (approved):
Tested on my Kobo with no issue:
0 > 20201022
20201010 > 20201022

PR #6813, opened 2020-10-22
Plugin for saving the cover image to file

zwim:
If the plugin is enabled the following actions can be selected:

1. On opening of a document, the cover is saved to a file (format png). 
If the filename exists, a backup of the original is created.

2. On closing of the document this file can be deleted and the original file can be restored.

3. Certain documents can be excluded from image generation.

This plugin can be used to change the sleep-screen on Tolinos (filename: /sdcard/suspend_others.jpg) as long KOReader has an opened document. On close of KOReader the original sleep screen is restored.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6813)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -382,11 +382,6 @@ function Device:isValidPath(path)
     return android.isPathInsideSandbox(path)
 end
 
-function Device:isValidFile(filename)
-    local path, name = util.splitFilePathName(filename)
-    return Device:isValidPath(path) and name ~= "" and lfs.attributes(filename, "mode") ~= "directory"
-end
-
 --swallow all events
 local function processEvents()
     local events = ffi.new("int[1]")
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -375,18 +375,11 @@ function Device:canExecuteScript(file)
     end
 end
 
-function Device:isValidPath(name)
-    local path, _ = util.splitFilePathName(name)
-    if android.isPathInsideSandbox(name) then
+function Device:isValidPath(path)
+    if path:find("^/sdcard") ~= nil then
         return true
-    else -- test if name is a symlink to the sandbox
-        path = path:gsub("/$", "") -- lua does not like trailing slash
-        if lfs.symlinkattributes(path, "mode") == "link" then
-            return android.isPathInsideSandbox(lfs.symlinkattributes(path, "target"))
-        else
-            return false
-        end
     end
+    return android.isPathInsideSandbox(path)
 end
 
 --swallow all events
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -376,9 +376,6 @@ function Device:canExecuteScript(file)
 end
 
 function Device:isValidPath(path)
-    if path:find("^/sdcard") ~= nil then
-        return true
-    end
     return android.isPathInsideSandbox(path)
 end
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -94,7 +94,6 @@ local Device = Generic:new{
     end,
     canImportFiles = function() return android.app.activity.sdkVersion >= 19 end,
     importFile = function(path) android.importFile(path) end,
-    isValidPath = function(path) return android.isPathInsideSandbox(path) end,
     canShareText = yes,
     doShareText = function(text) android.sendText(text) end,
 
@@ -373,6 +372,20 @@ function Device:canExecuteScript(file)
     local file_ext = string.lower(util.getFileNameSuffix(file))
     if android.prop.flavor ~= "fdroid" and file_ext == "sh"  then
         return true
+    end
+end
+
+function Device:isValidPath(name)
+    local path, _ = util.splitFilePathName(name)
+    if android.isPathInsideSandbox(name) then
+        return true
+    else -- test if name is a symlink to the sandbox
+        path = path:gsub("/$", "") -- lua does not like trailing slash
+        if lfs.symlinkattributes(path, "mode") == "link" then
+            return android.isPathInsideSandbox(lfs.symlinkattributes(path, "target"))
+        else
+            return false
+        end
     end
 end
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -382,6 +382,11 @@ function Device:isValidPath(path)
     return android.isPathInsideSandbox(path)
 end
 
+function Device:isValidFile(filename)
+    local path, name = util.splitFilePathName(filename)
+    return Device:isValidPath(path) and name ~= "" and lfs.attributes(filename, "mode") ~= "directory"
+end
+
 --swallow all events
 local function processEvents()
     local events = ffi.new("int[1]")
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -491,4 +491,9 @@ function Device:isValidPath(path)
     return util.pathExists(path)
 end
 
+function Device:isValidFile(filename)
+    local path, name = util.splitFilePathName(filename)
+    return Device:isValidPath(path) and name ~= "" and lfs.attributes(filename, "mode") ~= "directory"
+end
+
 return Device
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -487,9 +487,7 @@ function Device:canExecuteScript(file)
     end
 end
 
-function Device:isValidPath(name)
-    local path, _ = util.splitFilePathName(name)
-    -- pathExists wants trailing slash
+function Device:isValidPath(path)
     return util.pathExists(path)
 end
 
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -487,4 +487,10 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:isValidPath(name)
+    local path, _ = util.splitFilePathName(name)
+    -- pathExists wants trailing slash
+    return util.pathExists(path)
+end
+
 return Device
--- a/frontend/device/generic/device.lua
+++ b/frontend/device/generic/device.lua
@@ -491,9 +491,4 @@ function Device:isValidPath(path)
     return util.pathExists(path)
 end
 
-function Device:isValidFile(filename)
-    local path, name = util.splitFilePathName(filename)
-    return Device:isValidPath(path) and name ~= "" and lfs.attributes(filename, "mode") ~= "directory"
-end
-
 return Device
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -87,24 +87,6 @@ If the filename is empty or the file does not exist, the cover file will be dele
 
 If fallback is not activated the screensaver image remains after closing a book.]])
 
-local function isValidPath(name)
-    local path, _ = util.splitFilePathName(name)
-    if Device.isValidPath then -- on Android
-        if Device.isValidPath(name) then
-            return true
-        else -- test if name is a symlink to Sandbox
-            path = path:gsub("/$", "") -- lua does not like trailing slash
-            if lfs.symlinkattributes(path, "mode") == "link" then
-                return Device.isValidPath(lfs.symlinkattributes(path, "target"))
-            else
-                return false
-            end
-        end
-    else -- on emulator
-        return util.pathExists(path) -- pathExists wants trailing slash
-    end
-end
-
 function CoverImage:addToMainMenu(menu_items)
     menu_items.coverimage = {
 --        sorting_hint = "document",
@@ -129,7 +111,7 @@ function CoverImage:addToMainMenu(menu_items)
             {
                 text = _("Set system screensaver image"),
                 checked_func = function()
-                    return self.cover_image_path ~= "" and isValidPath(self.cover_image_path)
+                    return self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path)
                 end,
                 help_text = _("This is the filename, where the cover of the actual book is stored to."),
                 keep_menu_open = true,
@@ -158,7 +140,7 @@ function CoverImage:addToMainMenu(menu_items)
                                             self:cleanUpImage() -- with old filename
                                             self.cover_image_path = new_cover_image_path -- update filename
                                             G_reader_settings:saveSetting("cover_image_path", self.cover_image_path)
-                                            if self.cover_image_path ~= "" and isValidPath(self.cover_image_path) then
+                                            if self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path) then
                                                 self:onReaderReady(self.ui.doc_settings) -- with new filename
                                             else
                                                 self.enabled = false
@@ -180,10 +162,10 @@ function CoverImage:addToMainMenu(menu_items)
             {
                 text = _("Save current book cover as screensaver image"),
                 checked_func = function()
-                    return self:_enabled() and isValidPath(self.cover_image_path)
+                    return self:_enabled() and Device:isValidPath(self.cover_image_path)
                 end,
                 enabled_func = function()
-                    return self.cover_image_path ~= "" and isValidPath(self.cover_image_path)
+                    return self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path)
                 end,
                 callback = function()
                     if self.cover_image_path ~= "" then
@@ -269,8 +251,6 @@ function CoverImage:addToMainMenu(menu_items)
                 callback = function()
                     self.fallback = not self.fallback
                     G_reader_settings:saveSetting("cover_image_fallback", self.fallback)
---                    self:cleanUpImage()
---                    self:onReaderReady(self.ui.doc_settings)
                 end,
                 separator = true,
             },
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -7,6 +7,7 @@ end
 local InfoMessage = require("ui/widget/infomessage")
 local UIManager = require("ui/uimanager")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local ffiutil = require("ffi/util")
 local lfs = require("libs/libkoreader-lfs")
 local logger = require("logger")
 local util = require("util")
@@ -19,7 +20,7 @@ local function pathOk(filename)
 end
 
 local CoverImage = WidgetContainer:new{
-    name = 'coverimage',
+    name = "coverimage",
     is_doc_only = true,
 }
 
@@ -56,8 +57,8 @@ function CoverImage:cleanUpImage()
             timeout = 10,
         })
         os.remove(self.cover_image_path)
-    else
-        os.execute("cp \"" .. self.cover_image_fallback_path .. "\" \"" .. self.cover_image_path .. "\"")
+    elseif Device:isValidFile(self.cover_image_path) then
+        ffiutil.copyFile(self.cover_image_fallback_path, self.cover_image_path)
     end
 end
 
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -13,6 +13,12 @@ local util = require("util")
 local _ = require("gettext")
 local T = require("ffi/util").template
 
+local function pathOk(filename)
+    filename = filename:gsub("/$", "")
+    local path = util.splitFilePathName(filename)
+    return Device:isValidPath(path)
+end
+
 local CoverImage = WidgetContainer:new{
     name = 'coverimage',
     is_doc_only = true,
@@ -56,7 +62,7 @@ function CoverImage:cleanUpImage()
         self:dubiousFallbackImage()
         os.remove(self.cover_image_path)
     else
-        os.execute("cp " ..self.cover_image_fallback_path .. " " .. self.cover_image_path)
+        os.execute("cp " .. self.cover_image_fallback_path .. " " .. self.cover_image_path)
     end
 end
 
@@ -73,6 +79,7 @@ function CoverImage:onReaderReady(doc_settings)
         local image = self.ui.document:getCoverPageImage()
         if image then
             image:writePNG(self.cover_image_path, false)
+            logger.dbg("CoverImage: image written to " .. self.cover_image_path)
         end
     end
 end
@@ -111,7 +118,7 @@ function CoverImage:addToMainMenu(menu_items)
             {
                 text = _("Set system screensaver image"),
                 checked_func = function()
-                    return self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path)
+                    return self.cover_image_path ~= "" and pathOk(self.cover_image_path)
                 end,
                 help_text = _("This is the filename, where the cover of the actual book is stored to."),
                 keep_menu_open = true,
@@ -140,7 +147,7 @@ function CoverImage:addToMainMenu(menu_items)
                                             self:cleanUpImage() -- with old filename
                                             self.cover_image_path = new_cover_image_path -- update filename
                                             G_reader_settings:saveSetting("cover_image_path", self.cover_image_path)
-                                            if self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path) then
+                                            if self.cover_image_path ~= "" and pathOk(self.cover_image_path) then
                                                 self:onReaderReady(self.ui.doc_settings) -- with new filename
                                             else
                                                 self.enabled = false
@@ -162,10 +169,10 @@ function CoverImage:addToMainMenu(menu_items)
             {
                 text = _("Save current book cover as screensaver image"),
                 checked_func = function()
-                    return self:_enabled() and Device:isValidPath(self.cover_image_path)
+                    return self:_enabled() and pathOk(self.cover_image_path)
                 end,
                 enabled_func = function()
-                    return self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path)
+                    return self.cover_image_path ~= "" and pathOk(self.cover_image_path)
                 end,
                 callback = function()
                     if self.cover_image_path ~= "" then
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -25,21 +25,6 @@ function CoverImage:_restore()
     return self.enabled and self.restore
 end
 
--- check if settings of lastfile prohibits creation of cover image
-function CoverImage:excluded()
-    local lastfile = G_reader_settings:readSetting("lastfile")
-    local exclude_ss = false -- consider it not excluded if there's no docsetting
-    if DocSettings:hasSidecarFile(lastfile) then
-        if self and self.ui then
-            exclude_ss = self.ui.doc_settings:readSetting("exclude_cover_image")
-        else
-            local doc_settings = DocSettings:open(lastfile)
-            exclude_ss = doc_settings:readSetting("exclude_cover_image")
-        end
-    end
-    return exclude_ss or false
-end
-
 function CoverImage:init()
     self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "Cover.png"
     self.enabled = G_reader_settings:isTrue("cover_image_enabled")
@@ -47,35 +32,37 @@ function CoverImage:init()
     self.ui.menu:registerToMainMenu(self)
 end
 
+function CoverImage:cleanUpImage()
+    local lfs = require("libs/libkoreader-lfs")
+    local has_bak = lfs.attributes(self.cover_image_path .. ".bak")
+    -- Delete image if backup exists, the backup could contain a user defined sleep image
+    if lfs.attributes(self.cover_image_path) and has_bak then
+        os.remove(self.cover_image_path)
+    end
+    -- Restore backup if it exists, so we can use the last image.
+    -- On Tolino a user defined /sdcard/suspend_others.jgp can be used as screensaver on system sleep.
+    if has_bak then
+        os.rename(self.cover_image_path .. ".bak", self.cover_image_path)
+    end
+end
+
 function CoverImage:onCloseDocument()
     logger.dbg("CoverImage: onCloseDocument")
-    local lfs = require("libs/libkoreader-lfs")
-    -- Try to restore the state before start of KOReader.
-    -- If KOReader gets updated during operation (at least on Android) the backup survives.
-    -- Under normal conditions this should not happen.
-    if self.restore and not self.excluded() then
-        -- Delete image unconditionally, so if no backup exists, there will be no cover file.
-        -- This is useful on Tolino to use the system sleep image after KOReader exit.
-        if lfs.attributes(self.cover_image_path) then
-            os.remove(self.cover_image_path)
-        end
-        -- Restore backup if it exists, so we can use the last image.
-        -- On Tolino a user defined /sdcard/tolino_others.jgp can be used as screensaver on system sleep.
-        if lfs.attributes(self.cover_image_path ..".bak") then
-            os.rename(self.cover_image_path .. ".bak", self.cover_image_path)
-        end
+    -- Try to restore the state before the opening of a document.
+    if self.restore then
+        self:cleanUpImage()
     end
 end
 
 function CoverImage:onReaderReady(doc_settings)
    logger.dbg("CoverImage: onReaderReady")
-   if self.enabled and self.ui and self.ui.document and not self.excluded() then
+   if self.enabled and not doc_settings:readSetting("exclude_cover_image") then
         local lfs = require("libs/libkoreader-lfs")
         local image = self.ui.document:getCoverPageImage()
         if image then
             -- Do not override an existing backup, as this only exists on a crash or abnormal update of KOReader.
             -- This is very usefull, if an image shall be used as a system screensaver (e.g. on Tolino as /sdcard/suspend_other.jpg)
-            if not lfs.attributes(self.cover_image_path ..".bak") then
+            if not lfs.attributes(self.cover_image_path .. ".bak") then
                 os.rename(self.cover_image_path, self.cover_image_path .. ".bak" )
             end
             image:writePNG(self.cover_image_path, false)
@@ -85,7 +72,6 @@ end
 
 function CoverImage:addToMainMenu(menu_items)
     menu_items.coverimage = {
---        sorting_hint = "device", -- maybe this plugin is better situated in device?
         sorting_hint = "document",
         text_func = function()
             return _("Cover Image")
@@ -146,7 +132,8 @@ function CoverImage:addToMainMenu(menu_items)
                 end,
                 callback = function()
                     self.enabled = not self.enabled
-                    G_reader_settings:saveSetting("cover_image_restore", self.enabled)
+                    G_reader_settings:saveSetting("cover_image_enabled", self.enabled)
+                    G_reader_settings:flush()
                 end,
             },
             -- menu entry: restore
@@ -160,28 +147,28 @@ function CoverImage:addToMainMenu(menu_items)
                 callback = function()
                     self.restore = not self.restore and self.enabled
                     G_reader_settings:saveSetting("cover_image_restore", self.restore)
+                    G_reader_settings:flush()
                 end,
+                separator = true,
             },
             -- menu entry: exclude this cover
             {
                 text_func = function()
-                    return _("Exclude this cover image")
+                    return _("Exclude cover image of this document")
                 end,
                 checked_func = function()
                     return self.ui and self.ui.doc_settings and self.ui.doc_settings:readSetting("exclude_cover_image") == true
                 end,
                 callback = function()
-                    if CoverImage:excluded() then
+                    if self.ui.doc_settings:readSetting("exclude_cover_image") == true then
                         self.ui.doc_settings:saveSetting("exclude_cover_image", false)
-                        self.ui:saveSettings() -- save here, because self:onReaderReady needs it
                         self:onReaderReady(self.ui.doc_settings)
                     else
-                        self:onCloseDocument() -- do this before self.ui:saveSettings()
                         self.ui.doc_settings:saveSetting("exclude_cover_image", true)
-                        self.ui:saveSettings()
+                        self:cleanUpImage()
                     end
+                    self.ui:saveSettings()
                 end,
-                added_by_readermenu_flag = true,
             }
         },
     }
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -4,6 +4,7 @@ if not Device.isAndroid() and not Device.isEmulator() then
     return { disabled = true }
 end
 
+local DocSettings = require("docsettings")
 local UIManager = require("ui/uimanager")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
 local logger = require("logger")
@@ -24,6 +25,17 @@ function CoverImage:_restore()
     return self.enabled and self.restore
 end
 
+-- check if settings of lastfile prohibits creation of cover image
+function CoverImage:excluded()
+    local lastfile = G_reader_settings:readSetting("lastfile")
+    local exclude_ss = false -- consider it not excluded if there's no docsetting
+    if DocSettings:hasSidecarFile(lastfile) then
+        local doc_settings = DocSettings:open(lastfile)
+        exclude_ss = doc_settings:readSetting("exclude_cover_image")
+    end
+    return exclude_ss or false
+end
+
 function CoverImage:init()
     self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "Cover.png"
     self.enabled = G_reader_settings:isTrue("cover_image_enabled")
@@ -37,7 +49,7 @@ function CoverImage:onCloseDocument()
     -- Try to restore the state before start of KOReader.
     -- If KOReader gets updated during operation (at least on Android) the backup survives.
     -- Under normal conditions this should not happen.
-    if self.restore then
+    if self.restore and not self.excluded() then
         -- Delete image unconditionally, so if no backup exists, there will be no cover file.
         -- This is useful on Tolino to use the system sleep image after KOReader exit.
         if lfs.attributes(self.cover_image_path) then
@@ -53,7 +65,7 @@ end
 
 function CoverImage:onReaderReady(doc_settings)
    logger.dbg("CoverImage: onReaderReady")
-   if self.enabled and self.ui and self.ui.document then
+   if self.enabled and self.ui and self.ui.document and not self.excluded() then
         local lfs = require("libs/libkoreader-lfs")
         local image = self.ui.document:getCoverPageImage()
         if image then
@@ -146,6 +158,27 @@ function CoverImage:addToMainMenu(menu_items)
                     G_reader_settings:saveSetting("cover_image_restore", self.restore)
                 end,
             },
+            -- menu entry: exclude this cover
+            {
+                text_func = function()
+                    return _("Exclude this cover image")
+                end,
+                checked_func = function()
+                    return self.ui and self.ui.doc_settings and self.ui.doc_settings:readSetting("exclude_cover_image") == true
+                end,
+                callback = function()
+                    if CoverImage:excluded() then
+                        self.ui.doc_settings:saveSetting("exclude_cover_image", false)
+                        self.ui:saveSettings() -- save here, because self:onReaderReady needs it
+                        self:onReaderReady(self.ui.doc_settings)
+                    else
+                        self:onCloseDocument() -- do this before self.ui:saveSettings()
+                        self.ui.doc_settings:saveSetting("exclude_cover_image", true)
+                        self.ui:saveSettings()
+                    end
+                end,
+                added_by_readermenu_flag = true,
+            }
         },
     }
 end
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -87,6 +87,9 @@ function CoverImage:addToMainMenu(menu_items)
                 text_func = function()
                     return self.cover_image_path and _("Set system screensaver image")
                 end,
+                checked_func = function()
+                    return self.cover_image_path ~= ""
+                end,
                 keep_menu_open = true,
                 callback = function(menu)
                     local InputDialog = require("ui/widget/inputdialog")
@@ -176,6 +179,9 @@ function CoverImage:addToMainMenu(menu_items)
             {
                 text_func = function()
                     return self.cover_image_path and _("Set fallback image when no cover or excl. book")
+                end,
+                checked_func = function()
+                    return self.cover_image_fallback_path ~= ""
                 end,
                 keep_menu_open = true,
                 callback = function(menu)
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -1,0 +1,134 @@
+local Device = require("device")
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local logger = require("logger")
+local _ = require("gettext")
+local T = require("ffi/util").template
+
+local CoverImage = WidgetContainer:new{
+    name = 'coverimage',
+    is_doc_only = true,
+    settings_id = 0,
+}
+
+function CoverImage:_enabled()
+    return self.enabled
+end
+
+function CoverImage:_restore()
+    return self.enabled and self.restore
+end
+
+function CoverImage:init()
+    self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "Cover.jpg"
+    self.enabled = G_reader_settings:isTrue("cover_image_enabled")
+    self.restore = G_reader_settings:isTrue("cover_image_restore") and self.enabled
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function CoverImage:onCloseDocument()
+    logger.dbg("CoverImage: onCloseDocument")
+    local lfs = require("libs/libkoreader-lfs")
+    -- Try to restore the state before start of KOReader.
+    -- If KOReader gets updated during operation (at least on Android) the backup survives.
+    -- Under normal conditions this should not happen.
+    if self.restore then
+        -- Delete image unconditionally, so if no backup exists, there will be no cover file.
+        -- This is useful on Tolino to use the system sleep image after KOReader exit.
+        if lfs.attributes(self.cover_image_path) then
+            os.execute("rm " .. self.cover_image_path)
+        end
+        -- Restore backup if it exists, so we can use the last image.
+        -- On Tolino a user defined /sdcard/tolino_others.jgp can be used as screensaver on system sleep.
+        if lfs.attributes(self.cover_image_path ..".bak") then
+            os.execute("mv " .. self.cover_image_path .. ".bak " .. self.cover_image_path)
+        end
+    end
+end
+
+function CoverImage:onDocSettingsLoad(doc_settings, document)
+   logger.dbg("CoverImage: onDocSettingsLoad")
+   if self.enabled then
+        local lfs = require("libs/libkoreader-lfs")
+        local image = document:getCoverPageImage()
+        if image then
+            -- Do not override an existing backup, as this only exists on a crash or abnormal update of KOReader.
+            -- This is very usefull, if an image shall be used as a system screensaver (e.g. on Tolino as /sdcard/suspend_other.jpg)
+            if not lfs.attributes(self.cover_image_path ..".bak") then
+                os.execute("cp " .. self.cover_image_path .. " " .. self.cover_image_path .. ".bak" )
+            end
+            Device.screen.bb.writePNG(image, self.cover_image_path, false)
+        end
+    end
+end
+
+function CoverImage:addToMainMenu(menu_items)
+    menu_items.coverimage = {
+--        sorting_hint = "device", -- maybe this plugin is better situated in device?
+        text_func = function() return _("Cover Image") end,
+        checked_func = function() return self:_enabled() end,
+        sub_item_table = {
+            {
+                text_func = function() return self.cover_image_path and T(_("Cover Image (%1)"), self.cover_image_path)
+                    or _("Cover Image (none)") end,
+                checked_func = function() return self:_enabled() end,
+                callback = function(menu)
+                local InputDialog = require("ui/widget/inputdialog")
+                local sample_input
+                sample_input = InputDialog:new{
+                    title = _("Filename for Cover-Image"),
+                    input = self.cover_image_path,
+                    input_type = "string",
+                    description = _("You can enter the filename of the cover image here."),
+                    buttons = {
+                        {
+                            {
+                                text = _("Cancel"),
+                                callback = function()
+                                    UIManager:close(sample_input)
+                                    menu:updateItems()
+                                end,
+                            },
+                            {
+                                text = _("Disable"),
+                                callback = function()
+                                    self.cover_image_path = sample_input:getInputText()
+                                    G_reader_settings:saveSetting("cover_image_path", self.cover_image_path)
+                                    self.enabled = false
+                                    G_reader_settings:saveSetting("cover_image_enabled", false)
+                                    UIManager:close(sample_input)
+                                    menu:updateItems()
+                                end,
+                            },
+                            {
+                                text = _("Enable"),
+                                is_enter_default = true,
+                                callback = function()
+                                    self.cover_image_path = sample_input:getInputText()
+                                    G_reader_settings:saveSetting("cover_image_path", self.cover_image_path)
+                                    self.enabled = true
+                                    G_reader_settings:saveSetting("cover_image_enabled", true)
+                                    UIManager:close(sample_input)
+                                    menu:updateItems()
+                                end,
+                            },
+                        }
+                    },
+                }
+                UIManager:show(sample_input)
+                sample_input:onShowKeyboard()
+                end
+            },
+            {
+                text_func = function() return _("Restore image on exit") end,
+                checked_func = function() return self:_restore() end,
+                callback = function()
+                self.restore = not self.restore and self.enabled
+                G_reader_settings:saveSetting("cover_image_restore", self.restore)
+                end,
+            },
+        },
+    }
+end
+
+return CoverImage
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -25,6 +25,14 @@ function CoverImage:_fallback()
     return self.fallback
 end
 
+function CoverImage:dubiousFallbackImage()
+    UIManager:show(InfoMessage:new{
+        text = T(_("\"%1\" \nis not a valid image file!\nPlease correct fallback image in Cover-Image"), self.cover_image_fallback_path),
+        show_icon = true,
+        timeout = 10,
+    })
+end
+
 function CoverImage:init()
     self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "Cover.png"
     self.cover_image_fallback_path = G_reader_settings:readSetting("cover_image_fallback_path") or "Cover_fallback.png"
@@ -37,11 +45,7 @@ function CoverImage:cleanUpImage()
     if self.cover_image_fallback_path == "" or not self.fallback then
         os.remove(self.cover_image_path)
     elseif lfs.attributes(self.cover_image_fallback_path, "mode") ~= "file" then
-        UIManager:show(InfoMessage:new{
-            text = T(_("\"%1\" \nis not a valid image file!\nPlease correct fallback image in Cover-Image"), self.cover_image_fallback_path),
-            show_icon = true,
-            timeout = 10,
-        })
+        self:dubiousFallbackImage()
         os.remove(self.cover_image_path)
     else
         os.execute("cp " ..self.cover_image_fallback_path .. " " .. self.cover_image_path)
@@ -85,11 +89,12 @@ function CoverImage:addToMainMenu(menu_items)
             -- menu entry: filename dialog
             {
                 text_func = function()
-                    return self.cover_image_path and _("Set system screensaver image")
+                    return _("Set system screensaver image")
                 end,
                 checked_func = function()
                     return self.cover_image_path ~= ""
                 end,
+                help_text = _("This is the filename, where the cover of the actual book is stored to."),
                 keep_menu_open = true,
                 callback = function(menu)
                     local InputDialog = require("ui/widget/inputdialog")
@@ -143,6 +148,9 @@ function CoverImage:addToMainMenu(menu_items)
                 checked_func = function()
                     return self:_enabled()
                 end,
+                enabled_func = function()
+                    return self.cover_image_path ~= ""
+                end,
                 callback = function()
                     if self.cover_image_path ~= "" then
                         self.enabled = not self.enabled
@@ -178,11 +186,12 @@ function CoverImage:addToMainMenu(menu_items)
             -- menu entry: set fallback image
             {
                 text_func = function()
-                    return self.cover_image_path and _("Set fallback image when no cover or excl. book")
+                    return _("Set fallback image when no cover or excl. book")
                 end,
                 checked_func = function()
-                    return self.cover_image_fallback_path ~= ""
+                    return lfs.attributes(self.cover_image_fallback_path, "mode") == "file"
                 end,
+                help_text =  _("File to use as cover image, when no cover is wanted or book is excluded.\nLeave it blank to use nothing."),
                 keep_menu_open = true,
                 callback = function(menu)
                     local InputDialog = require("ui/widget/inputdialog")
@@ -207,6 +216,9 @@ function CoverImage:addToMainMenu(menu_items)
                                     callback = function()
                                         self.cover_image_fallback_path = sample_input:getInputText()
                                         G_reader_settings:saveSetting("cover_image_fallback_path", self.cover_image_fallback_path)
+                                        if lfs.attributes(self.cover_image_fallback_path, "mode") ~= "file" then
+                                            self:dubiousFallbackImage()
+                                        end
                                         UIManager:close(sample_input)
                                         menu:updateItems()
                                     end,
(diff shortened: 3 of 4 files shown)


plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+                                    menu:updateItems()
+                                end,
+                            },
+                            {
+                                text = _("Disable"),
+                                callback = function()
+                                    self.cover_image_path = sample_input:getInputText()
+                                    G_reader_settings:saveSetting("cover_image_path", self.cover_image_path)
+                                    self.enabled = false
+                                    G_reader_settings:saveSetting("cover_image_enabled", false)
```
poire-z, line 98: I haven't seen a screenshot of this widget - but it might be confusing to have a InputDialog with the path and `Cancel|Disable|Enable` buttons.

May be you should split that to 2 menu items:
`Set cover image path` => InputDialog with just `Cancel | Save` (dunno if we should show the path in the menu item)
`[ ] Save cover image on book opening`
`[ ] Restore image on exit`

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+                    },
+                }
+                UIManager:show(sample_input)
+                sample_input:onShowKeyboard()
+                end
+            },
+            {
+                text_func = function() return _("Restore image on exit") end,
+                checked_func = function() return self:_restore() end,
+                callback = function()
```
poire-z, line 125: Indent the function code.

poire-z (commented):
I'm fine with the principle (and quite a few people were requested that :)
Dunno if we should disable this plugin on some platforms like Kobo or Kindle where it would not make any sense as there are no other tool that could use such generated image?

(I think I'd also wouldn't mind if this was not a plugin but part of the core Screensaver code and menu.)

elvvis:
The plugin will not work on the Tolino. You can select a custom image (in png format) as your screensaver, but if the content of this image is changed later, this will not affect the screensaver (the screensaver image will be converted to jpg format after the png image is selected).
The correct way is to save the book cover as jpg (for Tolino Epos 2 in the directory /data/data/de.telekom.epub/files/lastread.jpg), but for this the device must be rooted.

zwim:
@elvvis: Are you sure? Have you tried it? I am using this plugin for two weeks now and it works on my Epos2.

> The plugin will not work on the Tolino. You can select a custom image (in png format) as your screensaver, but if the content of this image is changed later, this will not affect the screensaver (the screensaver image will be converted to jpg format after the png image is selected).

You have to store the image under /sdcard/suspend_others.jpg (use jpg even if the image is png).

> The correct way is to save the book cover as jpg (for Tolino Epos 2 in the directory /data/data/de.telekom.epub/files/lastread.jpg), but for this the device must be rooted.

You don't have to root your device for the screensaver! 
With this plugin I have my "normal" screensaver image in `/sdcard/suspend_others.jpg`. If I open a document in KOReader the original image is renamed to `/sdcard/suspend_others.jpg.bak` and the actual cover is stored as `/sdcard/suspend_others.jpg`. So I have the book-cover as screensaver. When I end KOReader, the old image is deleted and the backup gets renamed to `/sdcard/suspend_others.jpg`.

elvvis:
> @elvvis: Are you sure? Have you tried it? I am using this plugin for two weeks now and it works on my Epos2.
> 
I didn't know the trick to save the png-image as jpg, but it works great this way.
Fantastic that you wrote this plugin. 👍

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+                os.execute("cp " .. self.cover_image_path .. " " .. self.cover_image_path .. ".bak" )
+            end
+            Device.screen.bb.writePNG(image, self.cover_image_path, false)
+        end
+    end
+end
+
+function CoverImage:addToMainMenu(menu_items)
+    menu_items.coverimage = {
+--        sorting_hint = "device", -- maybe this plugin is better situated in device?
```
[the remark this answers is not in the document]
zwim: I do not find the screensaver menu on android (Tolino).
But I could put it in the Gear/Document menu.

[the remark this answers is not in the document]
elvvis: There is no screensaver menu on android.

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,134 @@
...
+        end
+        -- Restore backup if it exists, so we can use the last image.
+        -- On Tolino a user defined /sdcard/tolino_others.jgp can be used as screensaver on system sleep.
+        if lfs.attributes(self.cover_image_path ..".bak") then
+            os.execute("mv " .. self.cover_image_path .. ".bak " .. self.cover_image_path)
+        end
+    end
+end
+
+function CoverImage:onDocSettingsLoad(doc_settings, document)
```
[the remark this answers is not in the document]
zwim: In `onReaderReady` I have added:
```
function CoverImage:onReaderReady(doc_settings)
   logger.dbg("CoverImage: onDocSettingsLoad")
   if self.enabled and self.ui and self.ui.document then 
```

The plugin is not shown in FileManager, because I have set `    is_doc_only = true,`

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,153 @@
+local Device = require("device")
+
+if Device.isKindle == nil and Device.isKobo == nil then
```
pazos, line 3: you can also disable it on cervantes, remarkable, desktop ...

Or just enable it on android (and the emulator for testing)

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,153 @@
...
+   if self.enabled and self.ui and self.ui.document then
+        local lfs = require("libs/libkoreader-lfs")
+        local image = self.ui.document:getCoverPageImage()
+        if image then
+            -- Do not override an existing backup, as this only exists on a crash or abnormal update of KOReader.
+            -- This is very usefull, if an image shall be used as a system screensaver (e.g. on Tolino as /sdcard/suspend_other.jpg)
+            if not lfs.attributes(self.cover_image_path ..".bak") then
+                os.rename(self.cover_image_path, self.cover_image_path .. ".bak" )
+            end
+            Device.screen.bb.writePNG(image, self.cover_image_path, false)
```
elvvis, line 65: Instead of "Device.screen.bb.writePNG(image, self.cover_image_path, false)" you can use "image:writePNG(self.cover_image_path, false)", so writePNG does not need to be changed in base/ffi/blitbuffer.lua.

plugins/coverimage.koplugin/main.lua
zwim: I have disabled the plugin on all devices except Android and Emulator

plugins/coverimage.koplugin/main.lua
zwim: @elvvis: As you can see this PR just adds two files. For this PR nothing else has to be changed. So I don't understand your comment about base/ffi/blitbuffer.lua

NiLuJe: `getCoverPageImage()` returns a BB, `writePNG` is a BB method.

You can just do `image:writePNG(self.cover_image_path)` ;).

zwim: @NiLuJe: Ahh, now I see :-)

poire-z:
For the normal non-Android screensaver stuff, we have another option, `Exclude this book's cover from screensaver` - so when you don't like a cover for a book, you can ignore it and fallback to the other options.
You might want to duplicate it in your plugin (I actually use it quite often), and don't save the cover image on opening when it is set - or restore/save when toggling it.
https://github.com/koreader/koreader/blob/99ad13b13c896eeca0c1d41b957d05e61d3ab9de/frontend/apps/reader/modules/readermenu.lua#L152-L171

zwim:
@poire-z: Good point, now I can suppress embarrassing images on screen :-)

plugins/coverimage.koplugin/main.lua
```diff
@@ -24,6 +25,17 @@ function CoverImage:_restore()
     return self.enabled and self.restore
 end
 
+-- check if settings of lastfile prohibits creation of cover image
+function CoverImage:excluded()
+    local lastfile = G_reader_settings:readSetting("lastfile")
+    local exclude_ss = false -- consider it not excluded if there's no docsetting
+    if DocSettings:hasSidecarFile(lastfile) then
+        local doc_settings = DocSettings:open(lastfile)
+        exclude_ss = doc_settings:readSetting("exclude_cover_image")
```
poire-z, line 34: This is re-opening/parsing the metatada.zzz.lua, and might be costly. In most (or all, not sure about onCloseDocument) your contexts, you already have self.ui.odc_settings or doc_settings provided, so you can just `readSetting("exclude_cover_image")` on it - without having to re-open/parse it.

zwim: In line 34:
` exclude_ss = self.ui.doc_settings:readSetting("exclude_cover_image") ` brings `./luajit: plugins/coverimage.koplugin/main.lua:35: attempt to index field 'ui' (a nil value)`

and 
`        exclude_ss = DocSettings:readSetting("exclude_cover_image")`
 brings up `./luajit: frontend/docsettings.lua:170: attempt to index field 'data' (a nil value)`

So i have combined the old method with self.ui.doc_settings

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,190 @@
...
+function CoverImage:_restore()
+    return self.enabled and self.restore
+end
+
+-- check if settings of lastfile prohibits creation of cover image
+function CoverImage:excluded()
+    local lastfile = G_reader_settings:readSetting("lastfile")
+    local exclude_ss = false -- consider it not excluded if there's no docsetting
+    if DocSettings:hasSidecarFile(lastfile) then
+        if self and self.ui then
```
poire-z, line 33: No need to check `if self` (it always exists), I guess you meant `if self.ui and self.ui.doc_settings then`.

But I still don't like all that :)
It feels none of this is needed, and you don't need to care about `lastfile`:
- in `CoverImage:onCloseDocument()`, you don't really need to know if `excluded`: if `restore` and there's a .bak: restore it.
- in `CoverImage:onReaderReady(doc_settings)`, you have the doc_settings, so just read it

If there's some edge case because you call these 2 methods when toggling stuff in the menu, may be you should have 2 other methods: `CoverImage:backupPreviousAndCreateNewOne()` and `CoverImage:restorePreviousIfAny()` (with better names :) to do the low level check and action - and onCloseDocument/OnReaderReady/your menu callbacks would do the check they can from what they have - and call the 2 lowlevel methods if stuff should be done.

plugins/coverimage.koplugin/main.lua
poire-z: For the 2nd one:
`exclude_ss = DocSettings:readSetting("exclude_cover_image")`
`DocSettings` is a module/class, not an instance, so it has no data, nothing to readerSetting from :)

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,177 @@
...
+    local has_bak = lfs.attributes(self.cover_image_path .. ".bak")
+    -- Delete image if backup exists, the backup could contain a user defined sleep image
+    if lfs.attributes(self.cover_image_path) and has_bak then
+        os.remove(self.cover_image_path)
+    end
+    -- Restore backup if it exists, so we can use the last image.
+    -- On Tolino a user defined /sdcard/suspend_others.jgp can be used as screensaver on system sleep.
+    if has_bak then
+        os.rename(self.cover_image_path .. ".bak", self.cover_image_path)
+    end
```
poire-z, line 46: Looks like all this can be wrap in a single `if has_bak then`.
(And may be have `self.cover_image_path .. ".bak"` in a local variable to avoid doing the concat twice.

About the last comment:
> On Tolino a user defined /sdcard/suspend_others.jgp can be used as screensaver on system sleep

You mean one could chose to not have a main image (the one that becomes .bak)? In this case, shouldn't you also `os.remove(self.cover_image_path)` when `has_bak` is false ?

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,177 @@
...
+                text_func = function()
+                    return _("Save cover on book opening")
+                end,
+                checked_func = function()
+                    return self:_enabled()
+                end,
+                callback = function()
+                    self.enabled = not self.enabled
+                    G_reader_settings:saveSetting("cover_image_enabled", self.enabled)
+                    G_reader_settings:flush()
```
poire-z, line 136: Why did you feel the need to add this `:flush()` ?
It womm flush to disk settings.reader.lua - which shouldn't be needed to be done explicitely. `:saveSettings()` updates the table in memory, and it will always be flushed with the other stuff when needed (regularly, on switching documents, and always on KOReader exit).

poire-z (commented):
Looks simpler and sane :)

plugins/coverimage.koplugin/main.lua
zwim: Thank you for your hint. I have used `watch -n 0.5 grep cover_image settings.reader.lua` to watch changes. With this I noticed, that no changes in settings.reader.lua showed up. On the next start of KOReader the old values where still  there. Maybe this was caused by kill of KOReader-emulator (ctrl-c). 
Now I have checked this all over again, and could NOT confirm this strange behavior.

You are right, the flush is not necessary.

plugins/coverimage.koplugin/main.lua
zwim: Thank you, the has_bak on the remove is wrong.

plugins/coverimage.koplugin/main.lua
NiLuJe: There's no SIGQUIT handler, so, yeah, a ^C will murder the state ;).

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,176 @@
...
+    self.restore = G_reader_settings:isTrue("cover_image_restore") and self.enabled
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function CoverImage:cleanUpImage()
+    local lfs = require("libs/libkoreader-lfs")
+    -- Delete image the backup (if it exits) will contain a user defined sleep image
+    if lfs.attributes(self.cover_image_path) then
+        os.remove(self.cover_image_path)
+    end
```
poire-z, line 40: As this is always called by onCloseDocument() (if restore=true), if there is no coverimage, this might delete the default image (which would not have been renamed to .bak) ?

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,190 @@
...
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local logger = require("logger")
+local _ = require("gettext")
+local T = require("ffi/util").template
+
+local CoverImage = WidgetContainer:new{
+    name = 'coverimage',
+    is_doc_only = true,
+    settings_id = 0,
```
poire-z, line 16: unused (probably a left over of the copied base plugin)

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,190 @@
...
+function CoverImage:init()
+    self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "Cover.png"
+    self.enabled = G_reader_settings:isTrue("cover_image_enabled")
+    self.restore = G_reader_settings:isTrue("cover_image_restore") and self.enabled
+    self.ui.menu:registerToMainMenu(self)
+end
+
+function CoverImage:cleanUpImage()
+    if not self.ui.doc_settings:readSetting("exclude_cover_image") == true then
+        local lfs = require("libs/libkoreader-lfs")
```
poire-z, line 36: Move this lfs at top with the other requires (it's always loaded, so costs nothing even if not used).
(The other one below can then be removed.)

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,190 @@
...
+        self:cleanUpImage()
+    end
+end
+
+function CoverImage:onReaderReady(doc_settings)
+   logger.dbg("CoverImage: onReaderReady")
+   if self.enabled and not doc_settings:readSetting("exclude_cover_image") == true then
+        local lfs = require("libs/libkoreader-lfs")
+        local image = self.ui.document:getCoverPageImage()
+        if image then
```
poire-z, line 64: Still an issue in onCloseDocument/cleanUpImage() if there were no image here: the one that is `os.remove(self.cover_image_path)` is actually the main fallback image.

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,190 @@
...
+                    return _("Save book cover")
+                end,
+                checked_func = function()
+                    return self:_enabled()
+                end,
+                callback = function()
+                    self.enabled = not self.enabled
+                    G_reader_settings:saveSetting("cover_image_enabled", self.enabled)
+                    if self.enabled then
+                        self.restore = true
```
poire-z, line 142: Can't one toggle "enabled" while still wanting "restore" to stay false?
I assume the combination of `enabled=true` and `restore=false` is when one wants to keep the last book cover as the screensaver when exiting KOReader, or exiting it while in Filebrowser ?

poire-z (commented):
You must have thought more than I did about all the logic and edge cases - but reading it from a distance, the logic feels a bit fuzzy and unclear. It feels there _could_ be a cleaner and simpler logic to managing all that.

zwim:
You are right, the logic is a bit complicated.
I will let it sit for a while, now.

rola25:
Does this also work on Tolino firmware 13.2.1?

zwim:
@rola25: Actually I have 14.0.0, so I can not test it on 13.2.1. But the trick with putting a screensaver image in `/sdcard/suspend_others.jpg` also works on 13.2.1. 
So I think, the plugin should work. 

On old firmware versions (IIRC 10.0.x) the filename in `/sdcard/` was different.

If you try it, could you give me a feedback please?

rola25:
I can try it. Should I just copy the coverimage.koplugin Folder to /data/data/org.kor..... or do I need anything else? Which Screensaver do I have to choose in the tolinio-settings? Just the tolino-face screensaver?

zwim:
You just have to copy folder `coverimage.koplugin` in `/data/data/...`. You find the plugin settings, if you open Menu-Gear-Dokument, when you have a document opened. You won't see it in the filemanager.

pazos:
> On old firmware versions (IIRC 10.0.x) the filename in `/sdcard/suspend.jpg` (IIRC) was different.

Not sure about the file, but the path is highly specific. For example, in @rola25 [logcat](https://github.com/koreader/koreader/issues/6819#issuecomment-715600333) 

```
I/KOReader( 5021): Application info {
I/KOReader( 5021):   VM heap: 128MB (normal)
I/KOReader( 5021):   Flags: user.
I/KOReader( 5021):   Paths {
I/KOReader( 5021):     Assets: /data/data/org.koreader.launcher/files
I/KOReader( 5021):     Library: /data/app-lib/org.koreader.launcher-2
I/KOReader( 5021):     Storage: /storage/sdcard1
I/KOReader( 5021):   }
I/KOReader( 5021): }
```

The proper way of handling that is to use `android.getExternalStoragePath()`, which will always return a valid path to primary storage. (`/storage/sdcard1` in this case). Usually `/sdcard` is a symlink to that folder, but that's another topic.

zwim:
@pazos: I let the user enter a filename for the image. So the user is responsible to enter a valid path. If the plugin is (ab)used to change the screensaver image, this has to be entered only once (and it is device-specific).

Maybe later, if someone requests, we could improve the dialog for the filename with a path-selector and then we would use `android.getExternalStoragePath()', or we use the filemanager to select the image ...

pazos:
Yeah, I know and I'm fine with your proposal :+1: 

> Maybe later, if someone requests, we could improve the dialog for the filename with a path-selector and then we would use `android.getExternalStoragePath()', or we use the filemanager to select the image ...

You can do a runtime check with `Device:isValidPath(path)` when the user enters the file path. On the emulator it will return always true. On android it will return true if the path starts with `android.getExternalStoragePath()`

rola25:
I tried it, but it didn't work. I creates the file on the right place but it didn't use it, so I made a logcat and it said:
I/PowerManagerService( 2477): Going to sleep by user request...
D/FrontLight( 2477): Light====>SYS. brightness orig 0 conv 0
I/Terry-FB( 2477): strSuspendImgFileName = /storage/sdcard1/suspend.jpg
I/Terry-FB( 2477): strEPubFolder + "/" + strDefaultImgFileName = /data/data/de.telekom.epub/files//suspend.jpg
I/Terry-FB( 2477): strLocaleImgFolder + "/" + strDefaultImgFileName = /system/usr/sleep/drawable-de-nodpi/suspend.jpg
I/Terry-FB( 2477): strEPubFolder + "/" + strChargeImgFileName = /data/data/de.telekom.epub/files//suspend_charging.jpg
I/Terry-FB( 2477): strLocaleImgFolder + "/" + strChargeImgFileName = /system/usr/sleep/drawable-de-nodpi/suspend_charging.jpg
I/Terry-FB( 2477): strLocaleImgFolder + "/" + strFullImgFileName = /system/usr/sleep/drawable-de-nodpi/suspend_batteryfull.jpg
I/DANIEL-FB( 2477): strCurrentImgFileName =============== suspend.jpg
I/FrameBufferEx( 2477): Disp File: /data/data/de.telekom.epub/files/suspend.jpg

So I tried to set it to /sdcard/suspend.jpg, but it also doesn't work ... I just see the tolino face

rola25:
Perhaps because it has the wrong size?

pazos:
`I/Terry-FB( 2477): strSuspendImgFileName = /storage/sdcard1/suspend.jpg`

Try that. If that doesn't work it could be that it **needs** a valid jpg and not a png (or the size, I dunno)

zwim:
The right name of the file is `/sdcard/suspend_others.jpg`.
And the Tolino must not be connected via usb.

rola25:
Aaah, it works now, I was connected via ADB+USB, that was the problem. I  use /sdcard/suspend.jpg as Cover Image and that works when disconnected from USB

zwim:
@rola25: Good to hear :-)
On my Tolino `/sdard/suspend.jpg` does not work and on yours it does. Could you please check if `suspend_others.jpg` works on your device, too?

rola25:
I just tried it, and it doesn't work with suspend_others.jpg. I have a Toline Epos **1** with Firmware 13.2.1

zwim:
@pazos: I have a logical much easier implementation. The only "drawback" is, that the user has to enter the file name of his default screensaver image, which is chosen, when the Cover Image is deleted onDocumentClose. 

The UI would look like this
screenshot

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,236 @@
...
+    return self.enabled
+end
+
+function CoverImage:_restore()
+    return self.restore
+end
+
+function CoverImage:init()
+    self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "Cover.png"
+    self.cover_restore_image_path = G_reader_settings:readSetting("cover_restore_image_path") or "Cover_restore.png"
```
poire-z, line 30: Name the setting for this one `cover_image_restore_path` so they all have the same prefix `cover_image_`.

poire-z:
> The only "drawback" is, that the user has to enter the file name of his default screensaver image, which is chosen, when the Cover Image is deleted

I'm lacking knowledge of the contexts of how this could be used on Tolinos and other Androids.
But yes, it feels safer than trying to handle a .bak.

May be the wording (in menu and the code) could be changed to avoid "restore"?
Do I get it right that there is only the main cover_image_path that matters, and is to be configured here and possibly elsewhere in the system or in some other app that will show it as screensaver?
What happens with the system or other app when this file doesn't exist?

May be something like that would be clearer:
```
Set system screensaver image path
[ ] Save current book cover as system screensaver image
[ ] Exclude this book cover
----
Set fallback image path (to be copied) when no cover or excluded
[ ] Use fallback image when leaving book         (otherwise, keep current book cover (or fallback image))
```

zwim:
> Do I get it right that there is only the main cover_image_path that matters, and is to be configured here and possibly elsewhere in the system or in some other app that will show it as screensaver?

Yes `cover_image_path` is the file which is used by system for the screensaver (if not conneted by USB).
There is no way of changing this name I know of, but different Firmware uses slightly different names.

> What happens with the system or other app when this file doesn't exist?

Then the default system image is used. (Tolino sleeping)
 
> May be something like that would be clearer:
> 
> ```
> Set system screensaver image path
> [ ] Save current book cover as system screensaver image
> [ ] Exclude this book cover
> ----
> Set fallback image path (to be copied) when no cover or excluded
> [ ] Use fallback image when leaving book         (otherwise, keep current book cover (or fallback image))
> ```

Yes, is clearer to read.
We should also check the wording (of the dialogs and so on).

pazos:
> We should also check the wording (of the dialogs and so on).

An info message reminding users that the ability to set the image as a screensaver depends on the firmware and the plugin only saves the cover image to the specified file is a cool addon.

It doesn't need to be too technical (ie: this plugin saves current book cover to a file,  so it can be used as a screensaver if your android version supports it. e.g: Tolinos)

poire-z:
Probably needs another option:
```
Set fallback image path (to be copied) when no cover or excluded
[ ] Use fallback image when leaving book   
[x] Keep cover image when leaving book
```
So one can have the original "Tolino sleeping..." message if he prefers it.

zwim:
> Probably needs another option:
> 
> ```
> Set fallback image path (to be copied) when no cover or excluded
> [ ] Use fallback image when leaving book   
> [x] Keep cover image when leaving book
> ```
> 
> So one can have the original "Tolino sleeping..." message if he prefers it.

I can add that additional menu item, but it is not necessary. If the fallback image is an empty string, the "Tolino sleeping" message is used. I stated that in the "Set fallback image"-Dialog (better wording appreciated).

We have in total 16 different cases on closing a book (2 for each): use cover, use fallback, exclude book and empty fallback;
We can collapse them to the following cases:
1.) Use fallback checked and a fallback image given, current cover and exclude book don't care  -> use fallback 
2.) Use fallback checked and empty fallback image, current cover and exclude book don't care -> use system screensaver
3.) Use fallback and save current cover both not checked, current cover don't care -> use system screensaver
4.) Use fallback not checked and current cover checked, current cove not checked -> Cover image remains

An not to forget, if we close a book and the fallback image is not there, the user gets a message to correct the settings in Cover-Image. In this case we are using the system screensaver (that's no harm).

zwim:
> An info message reminding users that the ability to set the image as a screensaver depends on the firmware and the plugin only saves the cover image to the specified file is a cool addon.
> 
> It doesn't need to be too technical (ie: this plugin saves current book cover to a file, so it can be used as a screensaver if your android version supports it. e.g: Tolinos)

This is a very good point. I am not sure where/how to implement the info. Should I put it in the menu as a menu-item "Info/Usage" (which eveyone sees) or a long press on any item (which almost nobody does) or ...?

poire-z:
OK, looks sane.

> if we close a book and the fallback image is not there, the user gets a message to correct the settings in Cover-Image. In this case we are using the system screensaver 

Why bother him with a message - if it's a legitimate use case?

Also note that you can add to the menu item a `checked_func = function() end` to it's grey/disabled or black/enable if that function returns false or true. It's a kind of intuitive self documentation of the menu dependancies.
So, if fallback page image not set, the Use fallback image could stay grey.
Same for "Exclude..." if "Save cover image" is not set.

zwim:
> OK, looks sane.
> 
> > if we close a book and the fallback image is not there, the user gets a message to correct the settings in Cover-Image. In this case we are using the system screensaver
> 
> Why bother him with a message - if it's a legitimate use case?
> 

I don't want anyone to enter a misspelled fallback file name and then wondering why there is the default image. This reduces frustration and issue requests. If one wants the legitimate case, an empty filename would be right. 

> Also note that you can add to the menu item a `checked_func = function() end` to it's grey/disabled or black/enable if that function returns false or true. It's a kind of intuitive self documentation of the menu dependancies.
> So, if fallback page image not set, the Use fallback image could stay grey.
> Same for "Exclude..." if "Save cover image" is not set.

Thanks for the note. I will look at this and push it (tomorrow).

plugins/coverimage.koplugin/main.lua
```diff
@@ -176,7 +175,7 @@ function CoverImage:addToMainMenu(menu_items)
             -- menu entry: set fallback image
             {
                 text_func = function()
-                    return self.cover_image_path and _("Set fallback image when no cover")
+                    return self.cover_image_path and _("Set fallback image when no cover or excl. book")
```
poire-z: This text_func would return `nil` when self.cover_image_path is false.
I guess you can just let it be `text = _("Set fallback image path")` and add a `help_text = _("...")` with mode detailed info and advices (shown on long-press).

poire-z: And yes, for such a specific plugin, you can add a 1st menu item with detailed info (like we have in `Typography rules> About...`)

zwim:
In the settings menu there would be an additional page with only this entry, so I put it to the screen menu.
Any suggestion which `sorting_hint` is better is welcome.

I think the wording in the about message is suboptimal, too. Suggestions are welcome.

poire-z:
> Also note that you can add to the menu item a checked_func = function() end to it's grey/disabled or black/enable if that function returns false or true

Sorry, I meant to write `enabled_func=` (that you discovered)
(`checked_func` decides the state of a checkbox - which is not what I described :)

And, you don't need a text_func that may return nil and no text - or when the text is static.

As for the rewording suggestions, @Frenzie should be back in a few days and is better at that exercise than I am.
I'm wondering if this plugin shouldn't be renamed to systemscreensaverhelper.koplugin or something - the notion of "Cover Image/coverimage" is a bit generic and does not tell much about what it does.
Or could it be used for other things than system screensaver -  and so should be let generic?

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,281 @@
...
+This plugin saves the current book cover to a file, so it can be used as a screensaver, if your android version and firmware supports it (e.g: Tolinos).
+
+If enabled the cover image of the actual file is stored to the selected screensaver file. Certain books can be excluded.
+
+If fallback is activated, the fallback file will be copied to the screensaver file on book closing.
+If the filename is empty or the file does not exist, the cover file will be deleted and system screensaver is used.
+
+If fallback is not activated the screensaver image remains after closing a book.]])
+
+local function isValidPath(name)
```
pazos, line 90: I do understand what you did here. You're checking if the given path contains a symbolic link, like in /sdcard -> /storage/emulated/0. The right place to do that is in `frontend/device/android/device.lua`, in the specific Device:isValidPath implementation.

That way it can be repurposed for all the rest of the code that uses the function, and here you can do Device:isValidPath(path) directly without this very specific function.

The same with emulator. Right now `isValidPath` returns always true. Your proposal (check if pathExists) makes sense. Do it in `frontend/device/generic/device.lua` in the specific Device:isValidPath implementation.

zwim: > That way it can be repurposed for all the rest of the code that uses the function, and here you can do Device:isValidPath(path) directly without this very specific function.

Yes, you are right,  I tried to make this plugin as least invasive to the rest of the code as possible. But if you think this change is desirable, I like to put it in `android/device.lua`. But the proposed method is suboptimal. It only works on `/sdcard/` and not on `/sdcard/koreader/`.
I would like to improve the symlink detection on a different PR, 'cause I want this plugin first, and I only want one thing in one PR.

> The same with emulator. Right now `isValidPath` returns always true. Your proposal (check if pathExists) makes sense. Do it in `frontend/device/generic/device.lua` in the specific Device:isValidPath implementation.

Are you sure it is there? I don't find any `isValidPath` with grep on `device/*`, but it's no problem to add it to `generic/device.lua`.

I have done it now, but I like to test the changes before I push it.

frontend/device/android/device.lua
```diff
@@ -373,6 +372,20 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:isValidPath(name)
```
pazos, line 375: this function receives a path, not a name

frontend/device/android/device.lua
```diff
@@ -373,6 +372,20 @@ function Device:canExecuteScript(file)
...
+    local path, _ = util.splitFilePathName(name)
+    if android.isPathInsideSandbox(name) then
+        return true
+    else -- test if name is a symlink to the sandbox
+        path = path:gsub("/$", "") -- lua does not like trailing slash
+        if lfs.symlinkattributes(path, "mode") == "link" then
```
pazos, line 381: since the symbolic link is always between "/sdcard" and the primary storage partition you can simplify the whole function to

```lua
function Device:isValidPath(path)
    if path:find("^/sdcard") ~= nil then
        path = path:gsub("/sdcard", android.getExternalStoragePath())
    end
    return android.isPathInsideSandbox(path)
end
```

frontend/device/generic/device.lua
```diff
@@ -487,4 +487,10 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:isValidPath(name)
```
pazos, line 490: same as above, the function receives a path, not a filename

```lua
function Device:isValidPath(path)
    return util.pathExists(path)
end
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -87,24 +87,6 @@ If the filename is empty or the file does not exist, the cover file will be dele
 
 If fallback is not activated the screensaver image remains after closing a book.]])
 
-local function isValidPath(name)
```
pazos: split path and filename here, for example (call it like you want)

```lua
local function pathOk(filename)
    filename = filename:gsub("/$", "")
    local path = util.splitFilePathName(filename)
    return Device:isValidPath(path)
end
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -129,7 +111,7 @@ function CoverImage:addToMainMenu(menu_items)
             {
                 text = _("Set system screensaver image"),
                 checked_func = function()
-                    return self.cover_image_path ~= "" and isValidPath(self.cover_image_path)
+                    return self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path)
```
pazos, line 114: replace `Device:isValidPath(filename)` with `pathOk(filename)`

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,7 +140,7 @@ function CoverImage:addToMainMenu(menu_items)
                                             self:cleanUpImage() -- with old filename
                                             self.cover_image_path = new_cover_image_path -- update filename
                                             G_reader_settings:saveSetting("cover_image_path", self.cover_image_path)
-                                            if self.cover_image_path ~= "" and isValidPath(self.cover_image_path) then
+                                            if self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path) then
```
pazos, line 143: replace `Device:isValidPath(filename)` with `pathOk(filename)`

plugins/coverimage.koplugin/main.lua
```diff
@@ -180,10 +162,10 @@ function CoverImage:addToMainMenu(menu_items)
             {
                 text = _("Save current book cover as screensaver image"),
                 checked_func = function()
-                    return self:_enabled() and isValidPath(self.cover_image_path)
+                    return self:_enabled() and Device:isValidPath(self.cover_image_path)
```
pazos, line 165: replace `Device:isValidPath(filename)` with `pathOk(filename)`

plugins/coverimage.koplugin/main.lua
```diff
@@ -180,10 +162,10 @@ function CoverImage:addToMainMenu(menu_items)
...
                 end,
                 enabled_func = function()
-                    return self.cover_image_path ~= "" and isValidPath(self.cover_image_path)
+                    return self.cover_image_path ~= "" and Device:isValidPath(self.cover_image_path)
```
pazos, line 168: replace `Device:isValidPath(filename)` with `pathOk(filename)`

pazos (changes requested):

zwim:
@pazos: I have applied your requested changes. What is the procedure now?

pazos (approved):
> @pazos: I have applied your requested changes. What is the procedure now?

https://github.com/koreader/koreader/pull/6813#issuecomment-716584394

poire-z:
@Frenzie : can you have a quick look at the translatable strings ? (The wording sounds a bit more verbose to me vs what we're used to, i.e. "Please correct it." might be unneeded.)
Also, I'm not sure about the plugin name. CoverImage is simple and generic (and a bit tab-key-unfriendly to CoverBrowser :) ) but "Save cover image" or "System screensaver" is more true to what it does. Dunno, may be CoverImage is fine :)

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+                        self.ui.doc_settings:saveSetting("exclude_cover_image", true)
+                        self:cleanUpImage()
+                    end
+                    self.ui:saveSettings()
+                end,
+                separator = true,
+            },
+            -- menu entry: set fallback image
+            {
+                text = _("Set fallback image when no cover or excl. book"),
```
Frenzie, line 209:
```suggestion
                text = _("Set fallback image"),
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+                checked_func = function()
+                    return lfs.attributes(self.cover_image_fallback_path, "mode") == "file"
+                end,
+                help_text =  _("File to use as cover image, when no cover is wanted or book is excluded.\nLeave it blank to use nothing."),
```
Frenzie, line 213:
```suggestion
                help_text =  _("File to use when no cover is wanted (found ???) or book is excluded.\nLeave this blank to turn off the fallback image."),
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+                keep_menu_open = true,
+                callback = function(menu)
+                    local InputDialog = require("ui/widget/inputdialog")
+                    local sample_input
+                    sample_input = InputDialog:new{
+                        title = _("Filename of fallback image"),
+                        input = self.cover_image_fallback_path,
+                        input_type = "string",
+                        description = _("You can enter the filename of the fallback image here.\n" ..
+                            "Leave it empty to clean up on close."),
```
Frenzie, line 223: For translation with multiline use single line with `\n` or `[[]]`, no concat.
```suggestion
                        description = _("Leave this empty to remove the cover when the document is closed."),
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+                checked_func = function()
+                    return lfs.attributes(self.cover_image_fallback_path, "mode") == "file"
+                end,
+                help_text =  _("File to use as cover image, when no cover is wanted or book is excluded.\nLeave it blank to use nothing."),
+                keep_menu_open = true,
+                callback = function(menu)
+                    local InputDialog = require("ui/widget/inputdialog")
+                    local sample_input
+                    sample_input = InputDialog:new{
+                        title = _("Filename of fallback image"),
```
Frenzie, line 219: Also see remark above about path vs filename.
```suggestion
                        title = _("Fallback image filename"),
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+                            }
+                        },
+                    }
+                    UIManager:show(sample_input)
+                    sample_input:onShowKeyboard()
+                end,
+            },
+            -- menu entry: fallback
+            {
+                text = _("Use fallback image when leaving book"),
```
Frenzie, line 254:
```suggestion
                text = _("Turn on fallback image"),
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+end
+
+function CoverImage:cleanUpImage()
+    if self.cover_image_fallback_path == "" or not self.fallback then
+        os.remove(self.cover_image_path)
+    elseif lfs.attributes(self.cover_image_fallback_path, "mode") ~= "file" then
+        self:dubiousFallbackImage()
+        os.remove(self.cover_image_path)
+    else
+        os.execute("cp " .. self.cover_image_fallback_path .. " " .. self.cover_image_path)
```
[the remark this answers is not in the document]
NiLuJe: Err, speaking of, what's wrong with https://github.com/koreader/koreader-base/blob/master/ffi/util.lua#L197?

We shouldn't *ever* need to shell out to cp?

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+local T = require("ffi/util").template
+
+local function pathOk(filename)
+    filename = filename:gsub("/$", "")
+    local path = util.splitFilePathName(filename)
+    return Device:isValidPath(path)
+end
+
+local CoverImage = WidgetContainer:new{
+    name = 'coverimage',
```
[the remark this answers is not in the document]
zwim: Sorry I don't see it?

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
poire-z: What's wrong with it is that I didn't even know/remember we had it :) (I looked around frontend util.removeFile for a cp equivalent - forgot to look in base ffi/util.lua)

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
NiLuJe: Use double quotes instead of single quotes ;).

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
NiLuJe: I'll admit I kinda expected it in frontend, too. But then it's around path-related shenanigans that do belong in ffi because of win32 shenanigans, so, eeeh.

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+local logger = require("logger")
+local util = require("util")
+local _ = require("gettext")
+local T = require("ffi/util").template
+
+local function pathOk(filename)
+    filename = filename:gsub("/$", "")
+    local path = util.splitFilePathName(filename)
+    return Device:isValidPath(path)
+end
```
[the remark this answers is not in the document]
zwim: In my feeling it makes the code more readable than to put the same
`Device:isValidPath(util.splitFilePathName(self.cover_image_path)` multiple times in the code.

The line would get quite long and confusing to read.

I will add a check to see, if filename really contains a filename and not just a path.

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
Frenzie: > Err, speaking of, what's wrong with

Argh, I was semi-thinking "I guess we really need one of those" but figured I already had so much to say. Since we already have it that's a different story! ;-)

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
Frenzie: > In my feeling it makes the code more readable than to put the same
Device:isValidPath(util.splitFilePathName(self.cover_image_path) multiple times in the code.

And I would still say: why is util.splitetc. _in there_? Deal with it there.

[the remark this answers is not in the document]
Frenzie: Or in luajit-launcher.

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
zwim: Ohh, Ok

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
zwim: Ok, it uses `ffiutil.copyFile` now.

plugins/coverimage.koplugin/main.lua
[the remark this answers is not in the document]
zwim: Ok, I will put it in Device

pazos (changes requested):
Please remove the last commit. We don't need Device:isValidFile(filename). It just don't do what its name implies (check that a file is valid). It just check that its path is valid. You can do that on the plugin.

Also, for future reference:

1. Use `self:isValidPath` instead of `Device:isValidPath`
2. You don't need to declare a function in android device if it does the same than generic. You just declare the generic one, use self to access to other methods and it will automagically use the `isValidPath()` method defined for the platform.

frontend/device/android/device.lua
```diff
@@ -375,6 +374,13 @@ function Device:canExecuteScript(file)
     end
 end
 
+function Device:isValidPath(path)
+    if path:find("^/sdcard") ~= nil then
```
Frenzie, line 378: This doesn't belong here.

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,268 @@
...
+local lfs = require("libs/libkoreader-lfs")
+local logger = require("logger")
+local util = require("util")
+local _ = require("gettext")
+local T = require("ffi/util").template
+
+local function pathOk(filename)
+    local path, name = util.splitFilePathName(filename)
+    return Device:isValidPath(path) and name ~= "" and lfs.attributes(filename, "mode") ~= "directory"
+end
```
Frenzie, line 20: You never answered why this exists.

frontend/device/android/device.lua
zwim: `/sdcard` is a symbolic link which points to ` /storage/sdcard1` on Android. So it the path begins with /sdcard it points to our sandbox.

plugins/coverimage.koplugin/main.lua
zwim: This is for safety, so that we do not overwrite an existing file somewhere outside the sandbox.

1.) We could drop the `isValidPath` check (maybe for flavor "rocks"):
  * Bonus: We could overwrite a lot of image files (sandbox + all in KOReader app directory + a lot of in `/data/data`).
  * Malus: We could write to places, where we should not (e.g. other apps).
  * If we drop `isValidPath` then we should at least check if the extension is `png, jpg, jpeg`.

2.) We have to check if the cover image filename is not a directory, otherwise `copyFile` crashes.

3.) We have to check that the path does not end with a slash, otherwise `copyFile` crashes.

frontend/device/android/device.lua
zwim: Or do you mean it should be in `android.isPathInsideSandbox`

Frenzie: > Or do you mean it should be in android.isPathInsideSandbox

I mean that's where it should be if it weren't, but that's its very raison d'être! ;-)  All it does is check if the path starts with `Environment.getExternalStorageDirectory()`, which will return a value like `/sdcard`, `/storage/sdcard0`, `/storage/sdcard1`, etc. as appropriate.

plugins/coverimage.koplugin/main.lua
Frenzie: So something closer to:

```suggestion
local function pathOk(filename)
    return Device:isValidPath(filename) and util.fileExists(filename)
end
```

frontend/device/android/device.lua
pazos: > which will return a value like /sdcard, /storage/sdcard0, /storage/sdcard1, etc. as appropriate.

`Environment.getExternalStorageDirectory()`  will return the real path (in that case /storage/sdcard1). `/sdcard` is just a bogus path that points there on 95% of devices 90% of the time :p 

So, yeah. If a path starts with `/sdcard` the thing that will work on 100% of devices would be get the real path of the symbolic link and check if that matches the path returned by `Environment.getExternalStorageDirectory()`. I agree that can be moved to `android.isPathInsideSandbox`

Frenzie: In any case, either it's useful and it should be in luajit-launcher or it shouldn't be at all.

Frenzie: Is there really a use case to *hardcode* /sdcard?

pazos: where? :P 

All started in https://github.com/koreader/koreader/pull/6813#issuecomment-716124026 (and following 2 comments). Then @zwim did wrote the code and I answered in https://github.com/koreader/koreader/pull/6813#discussion_r513034218

The main usage *here* is to validate a input dialog entry that a user typed. Users don't know the real path and they're used to know that koreader and other apps have "folders" in /sdcard.

Anyways, the initial implementation (check if the path is a symbolic link using lfs) cannot be done in luajit-launcher, as we're not having access to lua modules there. As @zwim talked about how his proposal would work only sometimes, I suggested current implementation an asume that if a path starts with /sdcard this path will be a symlink to a path that it's under external storage path.

So 4 ways to solve that:

- do not validate user input at all
- validate user input properly with:
    1. check if path starts with `/sdcard`
    2. if that happens resolve the path using lfs
    3. check the resolved path against insideSandbox
- leave the stuff as is (without resolving the real path, just assuming that /sdcard indeed points to external storage). In that case we can move the function to luajit-launcher
- Implement the code to check if /sdcard == storage path in kotlin and update things accordly.

pazos: As always, this is full of android sh*t

Frenzie: Er, I'm talking about *this line*, which is fully separate from any other comments. ;-)

It's a matter of maintainability. Stick this [here](https://github.com/koreader/android-luajit-launcher/blob/f2d946b3b49e728272df4cb56185a2fe57cdb4ff/app/src/org/koreader/launcher/MainActivity.kt#L486-L490) or it will be super confusing to track down how the program actually works. There's no lfs here or some such.

Whether you stick it in plain library-less Lua in this very form or in Java/Kotlin with startsWith/regex/whatever all works the same

pazos: In that case the following should work:

```diff
diff --git a/app/src/org/koreader/launcher/MainActivity.kt b/app/src/org/koreader/launcher/MainActivity.kt
index 7c0039c..c4adc95 100644
--- a/app/src/org/koreader/launcher/MainActivity.kt
+++ b/app/src/org/koreader/launcher/MainActivity.kt
@@ -21,7 +21,7 @@ import androidx.core.app.ActivityCompat
 import androidx.core.content.ContextCompat
 import org.koreader.launcher.interfaces.JNILuaInterface
 import org.koreader.launcher.utils.*
-import java.io.IOException
+import java.io.*
 import java.util.*
 
 class MainActivity : NativeActivity(), JNILuaInterface,
@@ -483,9 +483,15 @@ class MainActivity : NativeActivity(), JNILuaInterface,
         }
     }
 
+    @SuppressLint("SdCardPath")
     override fun isPathInsideSandbox(path: String?): Int {
         return path?.let {
-            if (it.startsWith(device.externalStorage)) 1 else 0
+            val realPath = if (it.startsWith("/sdcard")) {
+                it.replace("/sdcard", File("/sdcard").canonicalPath)
+            } else {
+                it
+            }
+            if (realPath.startsWith(device.externalStorage)) 1 else 0
         } ?: 0
     }
```

@zwim: could you test it?

zwim: @pazos: I am afraid, it does not work. I have deleted `path:find("^/sdcard")` thing in `android/device.lua`.

`File("/sdcard").canonicalPath` points to `/mnt/media_rw/sdcard1` and 
`device.externalStorage` is `/storage/sdcard1` on my device.

When I replace `it.replace("/sdcard", File("/sdcard").canonicalPath)` with `it.replace("sdcard",device.externalStorage)` it works.

plugins/coverimage.koplugin/main.lua
zwim: @Frenzie: Hmm, your suggested solution is not right. If the file exists, it gets overwritten. This should happen on opening a new book. 

But we need to check:
1.) Is the file on a valid path -> isValidPath().
2.) Does the path to the file really exist -> util.pathExists()
3.) If the user enters a path with an empty filename -> name~=""
4.) Has the user entered a directoryname instead of a filename -> lfs.attributes()

So we can catch with 1.) `/data/data/cover.jpg` not valid
2.) `/sdcard/kor/cover.jpg`, if `kor` does not exist
3.) `/sdcard/` empty filename
4.) `/sdcard/koreader` directoryname

In each of these cases we should tell the user, to correct the image filename.

Frenzie: > Hmm, your suggested solution is not right.

It's not meant to be right; it's meant to look right. Slight difference. ;-)

> So we can catch with 1.) /data/data/cover.jpg not valid
2.) /sdcard/kor/cover.jpg, if kor does not exist
3.) /sdcard/ empty filename
4.) /sdcard/koreader directoryname

And there's the rub. That's not good UX, which explains why the function's suspicious to me. It's doing either too much or too little. The above was from the too much perspective, the following general idea from too little.

```suggestion
local function pathOk(filename)
    local path, name = util.splitFilePathName(filename)
    if not Device:isValidPath(path) then
        return false, _("Path isn't in a writable location.")
    elseif name == ""
        return false, _("Please enter a filename at the end of the path.")
    elseif lfs.attributes(filename, "mode") == "directory"
        return false, _("The path must point to a file, but it points to a directory.")
    end
    return true
end
```

The in between doesn't sit well with me.

plugins/coverimage.koplugin/main.lua
```diff
@@ -16,7 +16,17 @@ local T = require("ffi/util").template
 
 local function pathOk(filename)
     local path, name = util.splitFilePathName(filename)
-    return Device:isValidPath(path) and name ~= "" and lfs.attributes(filename, "mode") ~= "directory"
+    if not Device:isValidPath(path) then -- isValidPath expects a trailing slash
+        return false, T(_("Path \"%1\" isn't in a writable location."), path)
+    elseif not util.pathExists(path:gsub("/$", "")) then -- pathExists expects no trailing slash
+        return false, T(_("The path \"%1\" doesn't exist"), path)
```
Frenzie, line 22:
```suggestion
        return false, T(_("The path \"%1\" doesn't exist."), path)
```

Frenzie (approved):
Now I like it, just the one mini-nit. :-)

plugins/coverimage.koplugin/main.lua
zwim: I like it, too. Thank your for your guidance.

zwim:
As I see, one last issue is open: In `launcher/MainActivity.kt` the method `isPathInsideSandbox` has to be fixed for `/sdcard`

frontend/device/android/device.lua
pazos: what a shame, based on File documentation I would expect them to match.

zwim: Maybe this is due to Tolino Firmware 14.0.1 (which is some kind of Android 4.4.2).
Here I get with `ls -al /sd*`
```
lrwxrwxrwx root     root              2020-11-05 11:40 sdcard -> /storage/sdcard1
```
and with `ls -al /storage/*`
```
lrwxrwxrwx root     root              2020-11-05 11:40 extsd -> /mnt/media_rw/extsd
lrwxrwxrwx root     root              2020-11-05 11:40 sdcard1 -> /mnt/media_rw/sdcard1
lrwxrwxrwx root     root              2020-11-05 11:40 udisk -> /mnt/media_rw/udisk
```

sndwichm4n:
Wow, this will be a great function. I am already very excited! You are all great!!!

pazos:
> As I see, one last issue is open: In launcher/MainActivity.kt the method isPathInsideSandbox has to be fixed for /sdcard

@zwim: ready, you can bump android-luajit-launcher on this PR.

I marked as resolved some conversations with @Frenzie that I think they were fixed, not sure about @poire-z comments.

poire-z:
I don't want to re-read the whole things and check my comments :) so I let @zwim do that :) 
My main one is that :onReaderReady should be renamed to :something (so that only this :something() is called in the callbacks), and a :onReaderReady() added that would just call :something().
(Just like we have :onCloseDocument call :cleanUpImage, and everything else calls :cleanUpImage().)

zwim:
> @zwim: ready, you can bump android-luajit-launcher on this PR.

Just for safety, is this right:
``` 
git add platform/android/luajit-launcher
git push origin coverimage
```

pazos:
> > @zwim: ready, you can bump android-luajit-launcher on this PR.
> 
> Just for safety, is this right:
> 
> ```
> git add platform/android/luajit-launcher
> git push origin coverimage
> ```

If you're in koreader dir, you just need to do:

```
# rebase master first (just for safety)
git checkout master
git pull origin master
git checkout coverimage
git rebase master

# bump luajit-launcher
cd platform/android/luajit-launcher
git checkout master
git pull origin master
cd
git add platform/android/luajit-launcher
git commit ...
```

pazos (changes requested):
@zwim: you need to remove the base bump. There are also leftovers from previous PRs (QR code, highlights ...)

zwim:
@pazos: Thank you!

pazos (approved):

pazos:
@zwim: unfortunately I removed you as the author of the commits (I tried not to but I don't seem to have the git foo needed). I think you can re-author yourself by doing

```bash
git checkout coverimage
git pull yourrepo coverimage
./kodev fetch-thirdparty
git reset --soft HEAD~2
git commit -m "name of commit"
git push yourrepo coverimage -f
```

sndwichm4n:
Is this cover to screenshot tool in the next nightly?

NiLuJe:
It hasn't been merged yet, so, no?

sndwichm4n:
Ok, I am no programmer. When is the merging?

sndwichm4n:
I ask because pazos has approved and the plugin tool must be awesome. :)

zwim:
I have added a short description of the usage of this plugin on 
https://github.com/koreader/koreader/wiki/Android-tips-and-tricks

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,279 @@
...
+                                                UIManager:show(InfoMessage:new{
+                                                    text = is_path_ok_message,
+                                                    show_icon = true,
+                                                })
+                                            end
+                                        end
+                                        UIManager:close(sample_input)
+                                        menu:updateItems()
+                                    end,
+                                },
```
Galunid, line 167: Wouldn't it be better to check if the path is correct, and if it is only then do the clean up? Right now if user accidentally put in wrong path they can loose they cover images, right?

Galunid: Guess not, since entering invalid path is meant to disable the plugin.

Galunid (approved):
LGTM
@pazos Is there any way to broadcast android `Intent` from lua? I took a look at [Onyx SDK set cover](https://github.com/onyx-intl/OnyxAndroidDemo/blob/master/app/src/main/java/com/onyx/android/demo/ScreensaverActivity.java) and it doesn't seem too hard to add this, except for broadcasting the intent. [The part of interest](https://github.com/onyx-intl/OnyxAndroidDemo/blob/81a9a9de64784901006ca41b3f5d15572fafed75/app/src/main/java/com/onyx/android/demo/ScreensaverActivity.java#L80-L83) in the demo.

pazos:
> I have added a short description of the usage of this plugin on
> https://github.com/koreader/koreader/wiki/Android-tips-and-tricks

Thanks for the wiki stuff @zwim. I think it is better to document it on a separate plugin page (The text and all the samples are ok) and leave Android-tips-and-tricks for platform code.

pazos:
> @pazos Is there any way to broadcast android Intent from lua? I took a look at Onyx SDK set cover and it doesn't seem too hard to add this, except for broadcasting the intent. The part of interest in the demo.

Yup. There are plenty of samples in the code: a dict lookup, open a link in the browser, start the EPD test activity ... all use startActivity. There's also a wrapper called `startActivityIfSafe` that queries the packageManager before calling startActivity, so it is only called if there's a package able to resolve the intent. It should be useful to prevent ActivityNotFound exceptions.

Still not sure about your intent with the intent :)

pazos:
@Galunid: we don't use `sendBroadcast` anywhere on the code. After a quick read of https://developer.android.com/reference/android/content/Context#sendBroadcast(android.content.Intent) it looks like sendBroadcast(intent) is the same than startActivity(intent). The former is used to communicate with Broadcast receivers and the later is to communicate with activities.

ptrm:
Is this plugin checking for android or disabling itself because of some deps furthers down the code, or is there a chance it might work for Pocketbooks (they also can have an image specified for screensaver)

Frenzie:
> Is this plugin checking for android

Yes.

> is there a chance it might work for Pocketbooks

Well… also yes? ;-P Depending on what PB requires, only minimal adjustments may be needed.

zwim:
> Is this plugin checking for android or disabling itself because of some deps furthers down the code, or is there a chance it might work for Pocketbooks (they also can have an image specified for screensaver)?

All the plugin does, is to save the cover to a certain file (it must be in a place which is writeable for KOReader) and restore a fallback image file on close if wanted.

There is no Android mystery hidden in it.

As I don't have a pocketbook I can not test if it works there. But you can test it. Just disable lines 3-5 on `plugins/coverimage.koplugin/main.lua` and try it out.

I would be interested in the results.

ptrm:
The plugin itself works very well. However, the book cover location requires the file to be in... BMP format 😁. Below are the cover file dumps before and after using the plugin. Png is not recognised amd results in the generic "Touch HD 3" screen with features listed.


 On the other hand, tryung to use a cusrom image as power down logo also ezports the image to tge system/logo dir as BMP, so without BMP support the pkugin seems unusable on PocketBooks 😁 On which I don't insist, but if it's a matter of a single small dependency, that would be great :)

That beibg said, Pocketbook has a cool deature with it's native reader to shiw last rread page during biot, so you get to the most recent book quickly:

This might be a handy feature on Android as a screensaver, too, with just a screenshot instead of a cover.

NiLuJe:
IIRC, we don't currently have anything that supports writing BMP files.

On the other hand, BMP is a pretty "dumb" format, c.f., [stbiw](https://github.com/nothings/stb/blob/master/stb_image_write.h).

Frenzie:
Maybe PB comes with a built-in utility like ImageMagick for such a conversion? Also, I don't know much about BMP files besides having used 'em in Microsoft Paintbrush/Paint when I was little. If they're mostly compatible with PNM for example we already have MuPDF.

zwim:
Couldn't we use the example of lodepng for creating BMP files:
https://raw.githubusercontent.com/lvandeve/lodepng/master/examples/example_png2bmp.cpp

If no one else stands up, I could look into it more deeply tomorrow.

zwim:
I have done a quick test with https://github.com/nothings/stb and compiled a command line tool (Android, Emulator) to convert between bmp, jpg and png. For jpg even a compression (quality) setting is possible.

It seems to be usable on my Tolino.

`usage: image-convert infile outfile format quality` infile and outfile can be the same.

As I am not familiar to integrate a new library or C-functions to KOReader, I will definitively stop work here.
I hope someone wants to continue from here.

I could integrate the new function in the cover-image plugin.

```
/*
 * simple programm to convert files between bmp, jpg, png
 * Autor: @zwim
 * thanks to https://github.com/nothings/stb 
 *
 * Date: 11/12/2020
 * 
*/

#include <stdio.h>
#include <stdlib.h>
#include <strings.h>
 
#define STB_IMAGE_IMPLEMENTATION
#include "stb_image.h"
#define STB_IMAGE_WRITE_IMPLEMENTATION
#include "stb_image_write.h"
 
 int main(int argc, char *argv[]) {
    int width, height, channels;
    unsigned char *img;

    if ( argc < 5 ) {
        printf("usage: image-convert infile outfile format quality\n format can be: bmp, jpg, png\nquality is only valid for jpg: 1-100\n");
        exit(1);
    }

    img = stbi_load(argv[1], &width, &height, &channels, 0);

    if(img == NULL) {
        printf("Error in loading the image\n");
        exit(1);
    }
    printf("Loaded image with a width of %dpx, a height of %dpx and %d channels\n", width, height, channels);

    if ( strcasecmp(argv[3], "PNG" ) == 0 )
        stbi_write_png(argv[2], width, height, channels, img, width * channels);
    else if ( strcasecmp( argv[3], "JPG" ) == 0 )
        stbi_write_jpg(argv[2], width, height, channels, img, atoi(argv[4]) );
    else if ( strcasecmp( argv[3], "BMP" ) == 0 )
        stbi_write_bmp(argv[2], width, height, channels, img);

    stbi_image_free(img);    
    return 0; 
}
```

Frenzie:
> Couldn't we use the example of lodepng for creating BMP files:

I didn't realize the format was that simple, but I suppose it doesn't have compression. In that case I'd suggest simply popping in https://github.com/luapower/bmp/blob/25733a99f077bde8215d7d7939c96ea2371659bc/bmp.lua

zwim:
>  simple, but I suppose it doesn't have compression. 

In my younger days I was forced quite often to use bmp. It was a really dump format and had no compression. Don't know, if that was added later.

But I would prefer the stb solution, as we could also save images in the jpg format, which would save quite a bit of flash.
You can see a short demonstration (Android, Linux) on https://github.com/zwim/image-converter.

Frenzie:
> But I would prefer the stb solution, as we could also save images in the jpg format, which would save quite a bit of flash.

Er, if PB takes jpeg then we don't need any extra anything. ;-)

plugins/coverimage.koplugin/main.lua
```diff
@@ -0,0 +1,279 @@
...
+                end,
+                separator = true,
+            },
+            -- menu entry: set fallback image
+            {
+                text = _("Set fallback image"),
+                checked_func = function()
+                    return lfs.attributes(self.cover_image_fallback_path, "mode") == "file"
+                end,
+                help_text =  _("File to use when no cover is wanted (found ???) or book is excluded.\nLeave this blank to turn off the fallback image."),
```
Frenzie, line 220: @zwim Leftover draft of some sort, will you submit a PR to fix it?

zwim: Oh, thank you.

Yes, we can delete the help_text, as in the dialog shows instructions.
But I have noticed, the wording of just this dialog is wrong:
 https://github.com/koreader/koreader/blob/master/plugins/coverimage.koplugin/main.lua#L325 
It should be: "Leave this empty to keep the cover when the document is closed.", as this is what the plugin actually does.

I do a PR a bit later, today.

Issue #6808, opened 2020-10-21
KOReader doesn't reuse instance sometimes with new PB-firmware


tastytea:
* KOReader version: 2020.10.1-12
* Device: Pocketbook Touch HD 3 (PB632)

#### Issue

After updating the firmware from 6.0.926 to 6.1.900, KOReader doesn't reuse existing instances anymore under some circumstances.

A similar issue existed before (#6321) but was fixed and it worked with the old firmware.

#### Steps to reproduce (1)

* Open book from library.
* Suspend to background.
* Start KOReader from menu or open a different book from the library.
* → A new KOReader instance is opened.

Opening the same book from the library brings the backgrounded instance of KOReader to the foreground, as expected.

#### Steps to reproduce (2)

* Open KOReader from menu.
* Open book.
* Suspend to background.
* Open a different book from the library.
* → A new KOReader instance is opened.

Opening KOReader from menu brings the backgrounded instance of KOReader to the foreground, as expected.

##### `/system/config/extensions.cfg`

```
cbz:@ZIP_file:1:koreader.app,eink-reader.app:ICON_ZIP
djvu:@DJVU_file:1:koreader.app,eink-reader.app:ICON_DJVU
epub:@EPUB_file:1:koreader.app,eink-reader_with_blink.app,eink-reader_with_epub2.app:ICON_EPUB
fb2:@FB2_file:1:koreader.app,eink-reader.app:ICON_FB2
mobi:@MOBI_file:1:koreader.app,eink-reader.app:ICON_MOBI
pdf:@PDF_file:1:koreader.app,eink-reader_with_pdfium.app,eink-reader_with_rmsdk.app:ICON_PDF
zip:@ZIP_file:1:koreader.app:ICON_ZIP
```

ezdiy:
This is really odd, but unfortunately I have no way to investigate this (my device is fw6.0 only). Right after bringing koreadder to background, can you check that /tmp/koreader.pid still exists? Only if it were deleted somehow we'd be firing up new instance in the first place - https://github.com/koreader/koreader/blob/master/platform/pocketbook/koreader.app#L11

tastytea:
* Open book from library: `/tmp/koreader.pid` doesn't exist.
* Open KOReader from menu: PID file exists and contains correct PID.
* Open KOReader from menu, then open book from library: 2 instances of KOReader are running, PID file contains PID from the one started from the menu.
* Open KOReader automatically at boot with last book: PID file doesn't exists.

[...]

PR #6814, opened 2020-10-22
Pocketbook: Use a single launcher

ezdiy:
And make the system one a "symlink"
Fixes https://github.com/koreader/koreader/issues/6808

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6814)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -254,7 +254,7 @@ pbupdate: all
 	mkdir -p $(INSTALL_DIR)/applications
 	mkdir -p $(INSTALL_DIR)/system/bin
 	cp $(POCKETBOOK_DIR)/koreader.app $(INSTALL_DIR)/applications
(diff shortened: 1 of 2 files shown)


pazos:
A question, since I don't have the Pocketbook with me right now: 

Is `/mnt/ext1/applications/koreader.app` updated during an OTA update?. The koreader dir is `/mnt/ext1/applications/koreader`. If that's not the case I guess we can have the start script in `/mnt/ext1/applications/koreader/koreader.sh` and both `/mnt/ext1/applications/koreader.app` and `/mnt/ext1/system/bin/koreader.app` "symlinked" to it.

tastytea:
`/mnt/ext1/applications/koreader.app` is updated via OTA updates.

pazos (approved):

PR #6828, opened 2020-10-25
UIManager: Handle covers_fullscreen properly in close

NiLuJe:
`_repaint` already ensures that widgets below a full-screen one never get painted.
`close` wasn't doing that, which meant that it was sending `setDirty` calls on behalf of stuff that `_repaint` would later skip.

This was causing the refresh region to get needlessly inflated.

In the same vein, prevent being infected by the dither flag if said dithered widget is covered.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6828)

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -425,6 +425,8 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
     -- make it disabled by default and check if any widget wants it disabled or enabled
     Input.disable_double_tap = true
     local requested_disable_double_tap = nil
+    local something_covers_fullscreen = false
+    local start_idx = 1
     -- then remove all references to that widget on stack and refresh
     for i = #self._window_stack, 1, -1 do
         if self._window_stack[i].widget == widget then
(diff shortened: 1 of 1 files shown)


NiLuJe:
My original test-case was the couple of InfoMessage being shown/closed by the "Exec script" feature from the Favorites page.

I noticed that, when it was started from a dithered FM page, the refresh were wonky as hell (i.e., dithered, and full-screen).

This fixes it.

(I'd already noticed some funky business with the refresh queue in some instances the last time I touched UIManager, this should hopefully clear some of those up).

frontend/ui/uimanager.lua
```diff
@@ -425,19 +425,32 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
...
         else
-            -- If anything else on the stack was dithered, honor the hint
-            if self._window_stack[i].widget.dithered then
+            -- If anything else on the stack not already hidden by (i.e., below) a fullscreen widget was dithered, honor the hint
+            if self._window_stack[i].widget.dithered and not is_covered then
                 refreshdither = true
                 logger.dbg("Lower widget", self._window_stack[i].widget.name or self._window_stack[i].widget.id or tostring(self._window_stack[i].widget), "was dithered, honoring the dithering hint")
             end
 
+            -- Remember the uppermost widget that covers the full screen, so we don't bother calling setDirty on hidden (i.e., lower) widgets in the following dirty loop.
```
poire-z, line 443: Yeah, that loop is really dirty :)

poire-z (approved):

PR #6827, opened 2020-10-25
KoptInterface:getWordFromBoxes: guard against nil boxes

pazos:
Fixes #6825

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6827)

Issue #6825, opened 2020-10-25
Crash on higlight (attempt to get length of local 'boxes')


sr1920:
* KOReader version: 2020.10.1
* Device: reMarkable

#### Issue

Sometimes I observe crashes when highlighting on a PDF file.

#### Steps to reproduce

I have not identified a clear pattern for reproduction. It seems to depend on the PDF file and the specific part the user wants to highlight.

##### `crash.log` (if applicable)

Example of crash:

10/25/20-10:10:11 ERROR failed to call event handler onAddHighlight plugins/statistics.koplugin/main.lua:1923: attempt to perform arithmetic on field 'highlights' (a nil value)
Error in pixWordMaskByDilation: imin = 2 is too small
Error in pixConnComp: pixs not defined
Error in boxaSelectBySize: boxas not defined
Error in boxaSelectBySize: boxas not defined
Error in pixaCreateFromBoxa: boxa not defined
Error in boxaSort2d: boxas not defined
Error in pixaSort2dByIndex: pixas not defined
Error in pixaaFlattenToPixa: paa not defined
Error in pixaGetBoxa: pixa not defined
./luajit: frontend/document/koptinterface.lua:802: attempt to get length of local 'boxes' (a nil value)
stack traceback:
	frontend/document/koptinterface.lua:802: in function 'getWordFromBoxes'
	frontend/document/koptinterface.lua:1036: in function 'getTextFromPositions'
	frontend/apps/reader/modules/readerhighlight.lua:824: in function 'handler'
	frontend/ui/widget/container/inputcontainer.lua:255: in function 'handleEvent'
	frontend/ui/uimanager.lua:781: in function 'sendEvent'
	frontend/ui/uimanager.lua:241: in function '__default__'
	frontend/ui/uimanager.lua:1170: in function 'handleInputEvent'
	frontend/ui/uimanager.lua:1238: in function 'handleInput'
	frontend/ui/uimanager.lua:1282: in function 'run'
	./reader.lua:300: in main chunk
	[C]: at 0x0001415d
!!!!
Uh oh, something went awry... (Crash n°1: 10/25/20 @ 10:10:19)
Running on Linux 4.9.84-zero-gravitas (#1 Fri Jun 26 08:19:19 UTC 2020)
Attempting to restart KOReader . . .
!!!!

sr1920:
[...]

PR #6829, opened 2020-10-25
Statistics: Don't blow up when annotating/highlighting stuff with the plugin loaded, but disabled

NiLuJe:
Re #6825
Re #6827

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6829)


NiLuJe:
I also got rid of the inconsistent `return true`, because I guess we don't want to stop these events here?

(Although, currently, this is the sole bit of code handling those).

NiLuJe:
The actual accurate count is computed properly in initData, which is done when enabling the plugin, so we don't really have to think about the whole lifecycle any more ;).

poire-z (approved):

PR #6830, opened 2020-10-26
Another round of ReaderFooter fixes

NiLuJe:
(Some of this is related to the last one, #6540)

* When auto_refresh_time is enabled, don't actually refresh anything when the footer is hidden.
* Fix a bunch of state tracking related to height computations, meaning `getHeight()` is now actually accurate, always, and we don't need shitty workarounds anymore.
  * `footer_container.dimen.h` *includes* the progress bar, so, never reset it to 0 unless the progress bar is disabled (some `settings.progress_bar_position` codepaths were mistakenly doing just that).
  * More aggressively set/reset `footer_text.height` (not sure this one makes much of a difference, and/or if it's actually useful at all, but, the tracking was already there, but hella inconsistent, so, just fix it).
* Honor `settings.reclaim_height` in other bits of code that were only checking `footer_visible` to figure out the visible page area.
* Ask ReaderView to update the `visible_area` *now* when toggling the footer's visibility (for ReaderPaging).

This, among probably other minor quirks, makes the `_updateFooterText` repaint/refresh hacks mostly unnecessary, and fixes a whole lot of inconsistencies with the footer & ReaderPaging.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6830)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -665,8 +665,14 @@ function ReaderFooter:resetLayout(force_reset)
 end
 
 function ReaderFooter:getHeight()
+    print("ReaderFooter:getHeight", self.footer_content:getSize().h, self.vertical_frame:getSize().h, self.bottom_padding)
     if self.footer_content then
-        return self.footer_content:getSize().h
+        if self.view.footer_visible then
+            return self.vertical_frame:getSize().h + self.bottom_padding
+        else
+            -- When going invisible, the text is no longer visible, so the frame's height is off by self.height
+            return self.vertical_frame:getSize().h + self.height + self.bottom_padding
+        end
     else
         return 0
     end
@@ -1830,6 +1836,7 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
                 refresh_dim.h = self.vertical_frame:getSize().h + self.height + self.bottom_padding
             end
             refresh_dim.y = self._saved_screen_height - refresh_dim.h
+            print("_updateFooterText ->", self:getHeight())
         end
         -- If we're making the footer visible (or it already is), we don't need to repaint ReaderUI behind it
         if self.view.footer_visible then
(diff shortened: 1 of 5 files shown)


poire-z:
Can't comment much on all that, I've been too long away from ReaderFooter, and I don't want to go back in :) so, trusting you - but looks ok.

Just be sure changing footer mode or visibility does not trigger a CRE height change and re-rendering.

(I thought I had some readerfooter related tweak in my own patch - but no, the only bit I have is:
```diff
--- a/frontend/apps/reader/modules/readertypeset.lua
+++ b/frontend/apps/reader/modules/readertypeset.lua
@@ -487,7 +487,16 @@
     if self.view.footer.reclaim_height then
         bottom = Screen:scaleBySize(margins[4])
     else
-        bottom = Screen:scaleBySize(margins[4]) + self.view.footer:getHeight()
+        -- bottom = Screen:scaleBySize(margins[4]) + self.view.footer:getHeight()
+        -- somehow, the previous buggy behaviour felt nicer:
+        --   bottom = Screen:scaleBySize(margins[4] + DMINIBAR_HEIGHT)
+        -- but last line gets truncated with small margins.
+        -- So, tweak that a bit
+        local footer_height = self.view.footer:getHeight()
+        bottom = Screen:scaleBySize(margins[4]) + Math.round(footer_height / 2)
+        if bottom < footer_height then
+            bottom = footer_height
+        end
     end
     self.ui.document:setPageMargins(left, top, right, bottom)
     self.ui:handleEvent(Event:new("UpdatePos"))
```
I don't really remember what was that unpleasant :) - I don't use reclaim_height - may be you can read my mind in the past ? :)

NiLuJe:
@poire-z: `getHeight` was returning potentially bogus values in some setups, which may have been a factor we weren't really aware of at the time, otherwise, I can't really think of anything else than our old discussions about fitting the final line of text on screen and really unlucky margins? ;).

zwim:
> (I've certainly never seen CRe actually _clip_ the bottom line, and descenders are often very very close to the bottom edge with my usual settings, so I ought to have noticed ;p).

I had them sometimes, but it did not bother me enough to fill an issue.
Next time I notice them I will report it ;-)

Cheers

NiLuJe:
> Just be sure changing footer mode or visibility does not trigger a CRE height change and re-rendering.
 
Oops, fixed that in d9e9420 ;).

PR #6832, opened 2020-10-26
H2O: Fix initial tap detection

NiLuJe:
On a Kobo H2O, the very first tap after startup could be ignored, because the input code assumes the main finger gets assigned to slot 0. This isn't the case on the H2O, where it gets assigned slot 1. Fudge this on Device init to fix it.

(It's fairly finicky to reproduce, because if you move your finger a tiny bit, you get an extra "down" event which magically fixes everything).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6832)

### DIFF
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -91,6 +91,8 @@ local KoboDahlia = Kobo:new{
     hasFrontlight = yes,
     hasMultitouch = no,
     touch_phoenix_protocol = true,
+    -- There's no slot 0, the first fingers gets assigned slot 1, and the second slot 2
+    main_finger_slot = 1,
     display_dpi = 265,
     -- the bezel covers the top 11 pixels:
     viewport = Geom:new{x=0, y=11, w=1080, h=1429},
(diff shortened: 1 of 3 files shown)


poire-z (approved):
Not familiar with the cur_slot/ev_slots stuff, but looks harmless with other devices.

NiLuJe:
Just to confirm that it's indeed being a special snowflake, I checked the Forma, and the main finger is indeed on slot 0 there ;).

frontend/device/kobo/device.lua
```diff
@@ -310,6 +313,16 @@ function Kobo:init()
     }
     self.wakeup_mgr = WakeupMgr:new()
 
+    -- Tweak initial slot, if necessary
+    if self.main_finger_slot then
+        self.input.cur_slot = self.main_finger_slot
+        self.input.ev_slots = {
+            [self.main_finger_slot] = {
```
pazos, line 320: does this needs the [] ??.

frontend/device/kobo/device.lua
```diff
@@ -310,6 +313,16 @@ function Kobo:init()
...
+                slot = self.main_finger_slot,
```
pazos, line 321: ^^ this way is easier to read

frontend/device/kobo/device.lua
NiLuJe: c.f., https://github.com/koreader/koreader/blob/ac3c97af49e62c792cc7bdcfc51603c6289c63ed/frontend/device/input.lua#L144-L148

It's a *new* nested table, with hashes failing to mimic arrays ^^.

pazos (approved):

PR #6838, opened 2020-10-28
GestureDetector: add Tap interval on keyboard setting

### LINKED ISSUE #6836: same description for footnote-inpage_classic_classname and ..._smaller
Minor documentation bug: the explanatory text for 

"footnote-inpage_classic_classnames" and "footnote-inpage_classic_classnames_smaller",
[frontend/ui/data/css_tweaks.lua:794](https://github.com/koreader/koreader/blob/master/frontend/ui/data/css_tweaks.lua#L794) and [frontend/ui/data/css_tweaks.lua:810](https://github.com/koreader/koreader/blob/master/frontend/ui/data/css_tweaks.lua#L810) is identical

Seen on Weblate
https://hosted.weblate.org/translate/koreader/koreader/en/?checksum=513b951d8ce8d199#comments

poire-z:
`Style tweaks: fix some wording`
See https://github.com/koreader/koreader/issues/6836#issuecomment-717910988. Closes #6836. 

`GestureDetector: add Tap interval on keyboard setting`
Follow up to #6798: allow specifying an other value for tap interval when the keyboard is shown
(a good value for tap interval on reader and UI elements might be too long on the keyboard, and prevent typing fast). (And may be some other people will want a bigger one on the keyboard and 0 in the UI/Reader.)
See https://github.com/koreader/koreader/pull/6798#issuecomment-716852598

<kbd>image</kbd>
(same description text as on the other one)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6838)

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -399,6 +399,8 @@ function UIManager:show(widget, refreshtype, refreshregion, x, y, refreshdither)
     else
         Input.disable_double_tap = true
     end
+    -- a widget may override tap interval (when it doesn't, nil restores the default)
+    Input.tap_interval_override = widget.tap_interval_override
 end
 
 --[[--
@@ -459,6 +461,10 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
     end
     if requested_disable_double_tap ~= nil then
         Input.disable_double_tap = requested_disable_double_tap
+    end
+    if #self._window_stack > 0 then
+        -- set tap interval override to what the topmost widget specifies (when it doesn't, nil restores the default)
+        Input.tap_interval_override = self._window_stack[#self._window_stack].widget.tap_interval_override
     end
     if dirty and not widget.invisible then
         -- schedule the remaining visible (i.e., uncovered) widgets to be painted
--- a/frontend/ui/widget/virtualkeyboard.lua
+++ b/frontend/ui/widget/virtualkeyboard.lua
@@ -587,6 +587,7 @@ function VirtualKeyPopup:init()
             }
         },
     }
+    self.tap_interval_override = G_reader_settings:readSetting("ges_tap_interval_on_keyboard") or 0
 
     if Device:hasDPad() then
         self.key_events.PressKey = { {"Press"}, doc = "select key" }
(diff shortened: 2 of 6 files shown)


frontend/ui/uimanager.lua
```diff
@@ -460,6 +462,10 @@ function UIManager:close(widget, refreshtype, refreshregion, refreshdither)
     if requested_disable_double_tap ~= nil then
         Input.disable_double_tap = requested_disable_double_tap
     end
+    if #self._window_stack > 0 then
```
NiLuJe, line 465: Skip this if tap interval is set to 0?

NiLuJe: (Although, granted, it's not technically the end of the world, and it may make the logic murkier).

frontend/ui/widget/virtualkeyboard.lua
```diff
@@ -587,6 +587,7 @@ function VirtualKeyPopup:init()
             }
         },
     }
+    self.tap_interval_override = G_reader_settings:readSetting("ges_tap_interval_on_keyboard") or 0
```
NiLuJe, line 590: Would it make sense to 'nil' it instead of setting it to 0 in the fallback?

(Again, that may not be a great idea, and might make the logic/intent less clear).

NiLuJe (approved):
If you forget my bad ideas and stupid questions, LGTM ;).

frontend/ui/uimanager.lua
poire-z: > Skip this if tap interval is set to 0?

One might have the generic tap interval (used when turning page by tap) set to 0 - but still wanted to have a tap interval on keyboard non-0.
(But not sure what you mean, stkip the `#self._window_stack>0` test only ? And the generic tap interval is not readily accessible from there, it's in GestureDetector, so, a bit of ugly reaching it via Input.gesture_detector...)

frontend/ui/widget/virtualkeyboard.lua
poire-z: `nil` 'ing it would make the keyboard not override and use the generic `Tap interval`.
I think it would make the 2 menu items a bit unintuitive:
```
Tap interval
Tap interval on keyboard
```
Currently, the first can be seen as "everywhere else but with the keyboard).

With `nil`, it would mean "everywhere else including the keyboard, until you once use and apply `Tap interval on keyboard` " - which would set it to something else than nil.
Or I would have to save nil when the spinwidget value is 0 - making that number picker unintuitive - and this would prevent getting 0 with the keyboard and 400ms for the generic value (my use case :).

PR #6840, opened 2020-10-31
Reader: rationalize "Back" key/action handling

poire-z:
Have ReaderBack be the sole handler of onBack.
Add 4 mutually exclusive options for the Back key, to avoid ReaderLink and ReaderBack location stacks from interfering (ReaderBack's stack being always a superset of ReaderLink's stack).
So, remove "Enable back history", which is replaced by Back option "Go to previous read page".
Fix a few possible crashes and inconsistencies (when zoom, scroll or reflow have changed) with ReaderPaging and ReaderView when restoring previous context (page/view).

Some details about the issues in https://github.com/koreader/koreader/pull/6619#issuecomment-717990825 and follow up.

It can still be a bit of a mess (mostly, one back out of two not changing anything) in ReaderPaging, when switching some zoom/scroll/reflow settings. I tried and failed fixing more stuff in ReaderPaging/ReaderView, so I ended up being the less intrusive in them - at the expense of catching many events, some possibly uneeded/duplicating. But well...

The default for this new option is now the `Back to exit` action. No need to have "Go to previous read page" (old "Enable back history") be the default as most devices don't have a back key, and I think not many users might use the back action (which was quite confusing with its inconsitencies).
I don't know if we need to have an other default for devices with hasKeys and not isTouch.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6840)

### DIFF
--- a/frontend/apps/reader/modules/readerback.lua
+++ b/frontend/apps/reader/modules/readerback.lua
@@ -118,7 +118,7 @@ ReaderBack.onViewRecalculate = ReaderBack._onViewPossiblyUpdated
 ReaderBack.onPagePositionUpdated = ReaderBack._onViewPossiblyUpdated
 
 function ReaderBack:onBack()
-    local back_in_reader = G_reader_settings:readSetting("back_in_reader") or "default"
+    local back_in_reader = G_reader_settings:readSetting("back_in_reader") or "previous_location"
     local back_to_exit = G_reader_settings:readSetting("back_to_exit") or "prompt"
 
     if back_in_reader == "previous_read_page" then
--- a/frontend/apps/reader/modules/readerback.lua
+++ b/frontend/apps/reader/modules/readerback.lua
@@ -1,89 +1,193 @@
+local ConfirmBox = require("ui/widget/confirmbox")
 local Device = require("device")
 local Event = require("ui/event")
 local EventListener = require("ui/widget/eventlistener")
+local Notification = require("ui/widget/notification")
+local UIManager = require("ui/uimanager")
 local logger = require("logger")
 local util = require("util")
 local _ = require("gettext")
+
+-- This module handles the "Back" key (and the "Back" gesture action).
+-- When global setting "back_in_reader" == "previous_read_page", it
+-- additionally handles a location stack for each visited page or
+-- page view change (when scrolling in a same page)
 
 local ReaderBack = EventListener:new{
     location_stack = {},
     -- a limit not intended to be a practical limit but just a failsafe
     max_stack = 5000,
-    -- allow for disabling back history, and having Back key
-    -- quit immediately (useful for some developers)
-    disabled = G_reader_settings:isFalse("enable_back_history"),
 }
 
 function ReaderBack:init()
     if Device:hasKeys() then
         self.ui.key_events.Back = { {"Back"}, doc = "Reader back" }
     end
+    -- Regular function wrapping our method, to avoid re-creating
+    -- an anonymous function at each page turn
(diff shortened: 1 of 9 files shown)


NiLuJe:
> I don't know if we need to have an other default for devices with hasKeys and not isTouch.
> 

There *may* be a case to be made for "previous location" there, but no strong feelings about that (and no real first-hand experience either, I've never actually used KOReader on a K4 ;)).

frontend/apps/reader/modules/readerback.lua
```diff
@@ -1,89 +1,193 @@
...
+    end
+    self.back_resist = nil
+end
 
-    if self.disabled then
-        self.ui:handleEvent(Event:new("Close"))
-    elseif #location_stack > 1 then
-        local saved_location = table.remove(location_stack)
+-- Hook to events that do/may change page/view (more than one of these events
+-- might be send on a single page turn/scroll, _addPreviousLocationToStack()
```
NiLuJe, line 107:
```suggestion
-- may be sent on a single page turn/scroll, _addPreviousLocationToStack()
```

frontend/apps/reader/modules/readerback.lua
```diff
@@ -1,89 +1,193 @@
...
+                timeout = 2,
+            })
             return true
+        else
+            self.back_resist = nil
         end
-    else
-        logger.dbg("[ReaderBack] no location history, closing")
+    elseif back_in_reader == "previous_location" then
+        -- ReaderLink maintain its own location_stack of less frequent jumps
```
NiLuJe, line 152:
```suggestion
        -- ReaderLink maintains its own location_stack of less frequent jumps
```

frontend/apps/reader/modules/readerback.lua
```diff
@@ -1,89 +1,193 @@
...
+        -- (links or TOC entries followed, skim document...)
+        if back_to_exit == "disable" then
+            -- Let ReaderLink always show its notification if empty
+            self.ui.link:onGoBackLink(true) -- show_notification_if_empty=true
```
NiLuJe, line 156: Ooh, good idea of the Python-like named argument comment, makes those kind of tricks much more readable ;).

frontend/apps/reader/modules/readerback.lua
```diff
@@ -1,89 +1,193 @@
...
+function ReaderBack:_onViewPossiblyUpdated()
+    if G_reader_settings:readSetting("back_in_reader") == "previous_read_page" then
+        -- As multiple modules will have their :onPageUpdate()/... called,
+        -- and some of them will set up the new page with it, we need to
+        -- delay our handling after all of them are called (otherwise,
+        -- depending on the order of the calls, we may be have the location
+        -- of either the previous page or the current one).
+        UIManager:nextTick(self._addPreviousLocationToStackCallback)
+    end
+    self.back_resist = nil
```
NiLuJe, line 103: I'm not super-fond of the variable name, but I don't really have any better idea either ;D.

NiLuJe (approved):
At a quick glance, LGTM ;).

poire-z:
> There may be a case to be made for "previous location" there

I agree, so done. And having "previous location" as the default restores the ReaderLink "Back" handling I added for following links with keys (and when one has not followed any link or jumped around, it fallbacks to "Back to exit (prompt|always|disable)" - which should be fine.

poire-z:
Latest commit according to https://github.com/koreader/koreader/pull/6798#pullrequestreview-522658919

PR #6844, opened 2020-11-02
Qrcode widgets

jperon:
This PR defines 2 new widgets :
- QRWidget, that's like an ImageWidget, but with a `text` property that will be converted to a QR code ;
- QRMessage, that's like an InfoMessage, but shows the message as QR code.

Moreover, it adds the ability to export QR codes to the text editor.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6844)


frontend/qrencode.lua
```diff
@@ -0,0 +1,1353 @@
+--- The qrcode library is licensed under the 3-clause BSD license (aka "new BSD")
```
Frenzie, line 1: This file would probably fit better in base, or at least a lib subfolder. Also it looks like CircleCI isn't running on your PR but this file will likely need to be excluded. ^_^

Frenzie (commented):
Sounds cool! No time to review properly. ;-)

frontend/ui/widget/qrwidget.lua
```diff
@@ -0,0 +1,38 @@
...
+        for _, col in ipairs(ret) do
+            for _, lgn in ipairs(col) do
+                table.insert(qr, lgn > 0 and "1 " or "0 ")
+            end
+            table.insert(qr, "\n")
+        end
+        qr = table.concat(qr, '')
+        local dim = math.min(self.width, self.height)
+        self.image = RenderImage:renderImageData(qr, #qr, dim, dim)
+        return ImageWidget:new(self)
```
poire-z, line 34: I'm not sure this is a proper way to use our widgets :new()/:init(). `QRWidget:init()` does not init a proper QRWidget, but returns an ImageWidget. (May be it works, but it's a bit confusing vs all our other widgets I think).
If you dont plan on using a QRWidget anyplace else, you might as well just add it as a local function in qrmessage.lua `local getQRImageWidget = function(...)`
Otherwise, QRWidget might probably _extend_ ImageWidget to be more proper.

poire-z:
What's the use case ? (I'm not a heavy phone user, I never did anything QR and I have no idea what this could be used for :)

Frenzie:
I don't know in KOReader specifically, but you can quickly load links on your phone that way for example without having to type them. (Museums sometimes put QR codes to webpages with more info about a painting or something.)

frontend/qrencode.lua
pazos: I see a lot of bitwise and hex 2 bin ops (without luajit bit module). This is going to be really slow.

Maybe it is better to integrate https://github.com/vincascm/qrencode in base instead?.

jperon: OK to put it into base, but I don't understand what I have to do for CircleCI.

pazos (commented):
It's a very cool feature IMO.

Just some things:

- the button in the text editor works for single lines or very few characters, but hangs the program when used in a file of a few KB.

- the QrMessage widget produces a QrCode not readable on my phone. Also the qrcode looks low quality, zoomed¿?

frontend/ui/widget/qrmessage.lua
```diff
@@ -0,0 +1,117 @@
...
+local InputContainer = require("ui/widget/container/inputcontainer")
+local UIManager = require("ui/uimanager")
+local Input = Device.input
+local Screen = Device.screen
+
+local QRMessage = InputContainer:new{
+    modal = true,
+    timeout = nil, -- in seconds
+    text = nil,  -- The text to encode.
+    width = nil,  -- The width. Keep it nil to use original width.
```
pazos, line 36: if width or height are nil the thingy crash

```
./luajit: frontend/ui/widget/qrmessage.lua:65: attempt to perform arithmetic on field 'width' (a nil value)
stack traceback:
	frontend/ui/widget/qrmessage.lua:65: in function 'init'
	frontend/ui/elements/common_settings_menu_table.lua:19: in function 'callback'
	frontend/ui/widget/touchmenu.lua:802: in function 'action'
	frontend/ui/uimanager.lua:876: in function '_checkTasks'
	frontend/ui/uimanager.lua:1213: in function 'handleInput'
	frontend/ui/uimanager.lua:1306: in function 'run'
	./reader.lua:300: in main chunk
	[C]: at 0x55f4a029979e
```

jperon:
@poire-z I'm thinking about some uses, like an address book in simple text format : you could transfer some details to another person by flashing the QR code.

Another use case that comes from actuality : in France, to get out during confinement, you may have a numeric attestation in form of a QR code.

jperon:
@pazos 

>     * the button in the text editor works for single lines or very few characters, but hangs the program when used in a file of a few KB.

I think I'm going to truncate texts that are too long. The specification allows up to 4 296 characters, but I'm not sure how much resources it would require.

>     * the QrMessage widget produces a QrCode not readable on my phone. Also the qrcode looks low quality, zoomed¿?

Yes, it is zoomed. I don't know how to avoid smoothing while scaling, using `RenderImage`.

poire-z:
(^^ That's fine for some generic use case of QR codes. But I don't see much how these 2 cases could work with the TextEditor plugin: many adresses in a text file, that you would have to delete many lines to get a single one to export as a QR code ? - and I don't think cops will understand if I type my own attestation in our TextEditor and export it as a QR code they can scan :)  )

jperon:
(About the attestation, you reason as if French regulations were consistent…
About the address book, I'd plan to add QR generation to the menu that appears when one highlights words in a text. But for that, I'll have to understand KOReader a bit better !)

frontend/qrencode.lua
Frenzie: You probably need to go on there and allow it access. The automated tests protect against a lot of issues automatically.

You only need to exclude the file from luacheck if it doesn't pass.

jperon:
@Frenzie How would I have to proceed to put `qrencode.lua` into `base` ? Within a new subfolder or otherwise ?

Frenzie:
You can probably put it in ffi/ even if it's not actually FFI (judging by @pazos email; didn't take much of a look).

NiLuJe:
Yeah, that's what I was going to suggest ;).

(Meaning it'll require another PR against https://github.com/koreader/koreader-base; then bumping base here after the base one is merged).

frontend/ui/widget/qrmessage.lua
```diff
@@ -0,0 +1,117 @@
...
+                    w = Screen:getWidth(),
+                    h = Screen:getHeight(),
+                }
+            }
+        }
+    end
+
+    local image_widget = QRWidget:new{
+        text = self.text,
+        width = .9 * self.width,
```
NiLuJe, line 65: Will require rounding, c.f., every other instance of similar computations in the codebase ;).

frontend/ui/widget/qrmessage.lua
```diff
@@ -0,0 +1,117 @@
...
+        height = .9 * self.height,
```
NiLuJe, line 66: Ditto

frontend/ui/widget/qrmessage.lua
```diff
@@ -0,0 +1,117 @@
...
+        alpha = self.alpha,
+        scale_factor = self.scale_factor,
+    }
+
+    local frame = FrameContainer:new{
+        background = Blitbuffer.COLOR_WHITE,
+        padding = .05 * math.min(self.width, self.height),
```
NiLuJe, line 73: Ditto

frontend/ui/widget/qrwidget.lua
NiLuJe: +1 on the extend proposal ;).

frontend/ui/widget/qrwidget.lua
```diff
@@ -0,0 +1,38 @@
...
+    local ok, ret = qrencode.qrcode(self.text)
+    if not ok then
+        logger.info("QRWidget: failed to generate QR code.")
+        return
+    else
+        -- We generate pbm data
+        local qr = {"P1\n", #ret[1], " ", #ret, "\n"}
+        for _, col in ipairs(ret) do
+            for _, lgn in ipairs(col) do
+                table.insert(qr, lgn > 0 and "1 " or "0 ")
```
NiLuJe: You could arguably generate a 4x4 square instead of single pixels to alleviate the "smooth scaling" issue.

jperon: Just did a 2x2…

NiLuJe: (Might require some creative intermediary format better suited to that than a Lua table; I'd go with a C char array that can be properly addresses via pointers, because that's what I'm familiar with ^^).

NiLuJe: I say familiar, because that's roughly what happens here, in FBInk: https://github.com/NiLuJe/FBInk/blob/d8a9e81335fa57eaf31c228456e3b5e4c5f1ffb5/fbink.c#L1433-L1504

poire-z: Or maybe for something that square'ish, you might not need any image: you could just have a widget that itself just paint black rects on its area with something like:
`bb:paintRect(x, y, w, h, Blitbuffer.COLOR_BLACK)`

frontend/ui/widget/qrwidget.lua
```diff
@@ -4,34 +4,43 @@ QRWidget shows a QR code for a given text.
...
-local QRWidget = Widget:new{
+local QRWidget = ImageWidget:extend{
     scale_factor = 0,
     text = ""
     -- see ImageWidget for other options.
 }
 
 function QRWidget:init()
-    local ok, ret = qrencode.qrcode(self.text)
+    local ok, ret = qrencode.qrcode(self.text:sub(1, 2953))
```
poire-z, line 17: I would check if it overflows, and if yes, truncate a bit less to have room to add `[...truncated...]` so the receiver know it did not get the full intended text.

frontend/ui/widget/qrmessage.lua
```diff
@@ -62,15 +62,18 @@ function QRMessage:init()
 
     local image_widget = QRWidget:new{
         text = self.text,
-        width = .9 * self.width,
-        height = .9 * self.height,
+        width = self.width and .9 * self.width,
+        height = self.height and .9 * self.height,
```
poire-z, line 66: I think we all like our `0` in `0.9` :) (I don't remember I've seen any `.9` in our code.)

frontend/ui/widget/qrwidget.lua
```diff
@@ -4,34 +4,43 @@ QRWidget shows a QR code for a given text.
...
-        local qr = {"P1\n", #ret[1], " ", #ret, "\n"}
+        local qr = {"P1\n", 2*#ret[1], " ", 2*#ret, "\n"}
         for _, col in ipairs(ret) do
-            for _, lgn in ipairs(col) do
-                table.insert(qr, lgn > 0 and "1 " or "0 ")
+            for i = 1, 2 do
+                for _, lgn in ipairs(col) do
+                    table.insert(qr, lgn > 0 and "1 1 " or "0 0 ")
+                end
+                table.insert(qr, "\n")
```
NiLuJe, line 29: I'm a bit fuzzy on the PBM format, but doesn't that introduce newlines after each "output" pixel, instead of after each scanline?

frontend/ui/widget/qrwidget.lua
```diff
@@ -4,34 +4,43 @@ QRWidget shows a QR code for a given text.
...
         logger.info("QRWidget: failed to generate QR code.")
         return
     else
         -- We generate pbm data
-        local qr = {"P1\n", #ret[1], " ", #ret, "\n"}
+        local qr = {"P1\n", 2*#ret[1], " ", 2*#ret, "\n"}
         for _, col in ipairs(ret) do
-            for _, lgn in ipairs(col) do
-                table.insert(qr, lgn > 0 and "1 " or "0 ")
+            for i = 1, 2 do
```
NiLuJe, line 25: (I'd also make the scaling factor a variable to make the intent slightly clearer, here and in the MUL above).

frontend/ui/widget/qrwidget.lua
jperon: The examples on [Wikipedia](https://en.wikipedia.org/wiki/Netpbm#PBM_example) put newlines only after each scanline, but I'm not sure at all !

frontend/ui/widget/qrmessage.lua
NiLuJe: (That, and, also, rounding ;p).

jperon: I'll have to be careful, because I'm very used to this way.

frontend/ui/widget/qrwidget.lua
NiLuJe: That was my understanding, too, but I'll admit I'd need to actually run the code myself to check if that's what happens here, and I'm not on my dev box, so I can't ;).

frontend/ui/widget/qrwidget.lua
NiLuJe: Yep, that's what I was hinting at ;).

jperon: I'd like to use BlitBuffer methods, but I don't really understand them (my programming skills aren't *that* good…). Moreover, I thought `QRMessage` could evolve towards something that would allow saving the result as a file, in which case the intermediate image representation would be useful.

roshavagarga:
Quick idea for possible use: Having the option of sharing a book's basic info as a QR code would make it easier if you want to send it to somebody from your phone with a few clicks/taps, or they can just scan it instead of typing?

jperon:
@poire-z Did I understand the extend proposal well ?

@pazos I think the bugs you saw should be gone.

frontend/ui/widget/qrwidget.lua
```diff
@@ -0,0 +1,56 @@
...
+                for _, lgn in ipairs(col) do
+                    for j = 1, scaling do
+                        table.insert(qr, lgn > 0 and "1 " or "0 ")
+                    end
+                end
+                table.insert(qr, "\n")
+            end
+        end
+        qr = table.concat(qr, '')
+        local dim
```
poire-z, line 42: When I read `dim`, I think about the diming option (to make it grey instead of black) we use elsewhere - and `dimen` is usually a table of x/y/w/h.
Might be better to name it `square_size` ?

poire-z:
> Did I understand the :extend proposal well ?

Yes, perfectly understood and implemented :)

frontend/ui/widget/qrwidget.lua
jperon: Here you are !

frontend/ui/widget/qrwidget.lua
poire-z: Would mostly be just some Lua fun :) Like:
```lua
function QRWidget:paintTo(bb, x, y)
  -- x/y are the top left position of the top/left of widget in the screen bb
  local grid_size = 64 --dunno what it is really :)
  local sq_size = square_size / grid_size -- square_size is your original "dim"
         -- so best to ensure before that your dim/square_size is a multiple of 64
  for sy=1, #grid_size do
    local sqy = sy*sq_size
    for sx=1, #grid_size do
      if that square is black then
        local sqx = sx*sq_size
        bb:paintRect(x+sqx, y+sqy, sq_size, sq_size, Blitbuffer.COLOR_BLACK)
      end
    end
  end
end
```
(You could then extract that bb from the Screen bb and writePNG() it if you want to save it as a file - or use an intermediate bb here to draw the square to, and at the end write that bb to the screen bb.)

NiLuJe: You could still feed a BB to RenderImage (I don't recall which method handles that exactly, but this is what the FM does for cover thumbnails, they're stored as raw BBs in the DB).

Basically, start with a w*h*scale BB, fill it white, and then for each input pixel, do a paintRect of scale by scale @ appropriate coordinates. You'll probably need a nested loop, the outer on (input) rows, the inner on (input) columns.
Scaled y == iy*scale (with iy being the input row, starting at row 0)
Scaled x == ix*scale (with ix being the input column, starting at col 0)

jperon:
As soon as https://github.com/koreader/koreader-base/pull/1227 is merged, I'll remove `frontend/qrencode.lua`.

frontend/ui/widget/qrmessage.lua
jperon: Should be fixed.

frontend/ui/widget/qrwidget.lua
```diff
@@ -29,8 +29,8 @@ function QRWidget:init()
         local grid_size = scale * #grid
         local bb = Blitbuffer.new(grid_size, grid_size)
         local white = Blitbuffer.COLOR_WHITE
-        for x, col in ipairs(grid) do
-            for y, lgn in ipairs(col) do
+        for x, col in pairs(grid) do
```
poire-z: If you use #grid, it's an array. I don't think pairs() would be more efficient that ipairs() - but ipairs() is clearer (and won't have us wondering whether this is a bug in a few years :)

jperon: Sorry, I was misinformed and thought that ipairs was slower than pairs.

jperon:
Just added QR generation to plugins and to highlight, if there is no objections.

poire-z:
That's a bit too much QR invasion everywhere to my liking :/ (for something I still see no obvious use case - and if it has a rare use case, there's no need to make it that present everywhere...)

I guess on the highlight dialog, that makes a 5th wide row with a little `QR` in it :) which must looks qruite ugqrly...

Also, can something goes into the clipboard without going thru this highlight dialog first to select "Copy" ? and if we have "QR" in there, there's no need for a plugin to additionally transform that copied text into a QR.

Or keep the plugin, and remove the QR from the highlight dialog, as we can copy and then use the plugin to make some QR from the copied stuff. That would allow an easy opt-out by just disabling this plugin.

frontend/ui/widget/qrwidget.lua
poire-z: (I really don't know if it is slower or not - it just feels wrong to use pairs on an array.)

jperon:
OK, just reverted the last one.

jperon:
I think it should be ready now. Do I have to do some history cleaning / force push, or is it mergeable as is ?

plugins/qrclipboard.koplugin/main.lua
```diff
@@ -0,0 +1,33 @@
...
+--]]--
+
+local Device = require("device")
+local QRMessage = require("ui/widget/qrmessage")
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local _ = require("gettext")
+
+local QRClipboard = WidgetContainer:new{
+    name = 'hello',
```
pazos, line 12: name = 'qrclipboard'

jperon: Sorry. Just fixed.

frontend/ui/widget/qrwidget.lua
```diff
@@ -0,0 +1,43 @@
...
+local QRWidget = ImageWidget:extend{
+    scale_factor = 0,
+    text = ""
+    -- see ImageWidget for other options.
+}
+
+function QRWidget:init()
+    local text = self.text
+    if #text > 2953 then
+        local truncated = _('... (truncated...)')
```
Galunid, line 20:
```suggestion
        local truncated = _("... (truncated...)")
```

plugins/qrclipboard.koplugin/_meta.lua
```diff
@@ -0,0 +1,6 @@
+local _ = require("gettext")
+return {
+    name = 'qrclipboard',
```
Galunid, line 3:
```suggestion
    name = "qrclipboard",
```

plugins/qrclipboard.koplugin/main.lua
```diff
@@ -0,0 +1,33 @@
...
+--]]--
+
+local Device = require("device")
+local QRMessage = require("ui/widget/qrmessage")
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local _ = require("gettext")
+
+local QRClipboard = WidgetContainer:new{
+    name = 'qrclipboard',
```
Galunid, line 12:
```suggestion
    name = "qrclipboard",
```

frontend/ui/widget/qrwidget.lua
```diff
@@ -0,0 +1,43 @@
...
+        text = text:sub(1, 2953 - #truncated) .. truncated
+    end
+    local ok, grid = qrencode.qrcode(text)
+    if not ok then
+        logger.info("QRWidget: failed to generate QR code.")
+        return
+    else
+        local scale = 5
```
NiLuJe, line 28: I'd rather this weren't hard-coded, but instead computed in a DPI-aware fashion via `Screen:scaleBySize`.

c.f., https://github.com/koreader/koreader/blob/master/frontend/ui/size.lua for examples ;).

Galunid (commented):
Just a few nits, other than that, lgtm

frontend/ui/widget/qrmessage.lua
```diff
@@ -0,0 +1,120 @@
...
+    return true
+end
+
+function QRMessage:onAnyKeyPressed()
+    -- triggered by our defined key events
+    self.dismiss_callback()
+    UIManager:close(self)
+    if self.readonly ~= true then
+        return true
+    end
```
poire-z, line 109: You can remove these here and below - you picked that from InfoMessage which have that (undocumented...) but it seems used only when the InfoMessage is used as a screensaver widget (which I guess this QRMessage won't ever need to be :)

frontend/ui/widget/qrwidget.lua
jperon: Just fixed ; thanks to your link, I discovered that there were default sizes, so I changed qrmessage to use default padding.

frontend/ui/widget/qrmessage.lua
```diff
@@ -0,0 +1,119 @@
...
+                    x = 0, y = 0,
+                    w = Screen:getWidth(),
+                    h = Screen:getHeight(),
+                }
+            }
+        }
+    end
+
+    local padding = Size.padding.fullscreen
+    logger.info("PADDING:", padding)
```
poire-z, line 66: Done logging & debugging ? :)

jperon: Sorry…

poire-z (approved):

poire-z:
Just had a try: the resulting QR is a bit blurred (the black edges are not really strong).
It seems the image you're creating is then rescaled:
```
11/03/20-23:58:50 DEBUG ImageWidget: _render'ing data 572 582
11/03/20-23:58:50 DEBUG ImageWidget: scale to fit, setting scale_factor to 5.447619047619
11/03/20-23:58:50 DEBUG ImageWidget: scaling by 5.447619047619
11/03/20-23:58:50 DEBUG RenderImage:scaleBlitBuffer: scaling
11/03/20-23:58:50 DEBUG ImageWidget: initial offsets 0 -5
```
Using `scale_factor = 1,` in qrmessage makes it sharp - but smaller than fullscreen.

Can't you just draw it to the right size so scale_factor = 1 (implicit if the right size?) can be used?

jperon:
@poire-z Couldn't get it to have scale_factor = 1 because of padding, but now it's as close as possible to 1 when using scale_factor = 0.

frontend/ui/widget/qrwidget.lua
```diff
@@ -26,7 +26,18 @@ function QRWidget:init()
...
+                scale = math.min(self.width, self.height)/#grid
+            else
+                scale = self.width/#grid
+            end
+        elseif self.height then
+            scale = self.height/#grid
+        else scale = 2
+        end
+        scale = Screen:scaleBySize(scale)
         local grid_size = scale * #grid
```
poire-z: Not sure I understand all that: self.width and self.height are absolute sizes in px.
So, then, your 'scale' is an absolute size too: no need to scaleBySize() then (possibly, on your emulator, scaleBySize does *1). Your `scale` here might better be named something else (square_size?)

jperon: Perhaps I didn't understand well what scaleBySize() does, but it seems it at least rounds values : when I don't use it (or math.ceil instead), I get "parasite" lines between "pixels". Perhaps I should just use math.ceil instead ?

jperon: Example without rounding:

qr

poire-z:
Feels like you should aim at using scale_factor=1, for quality, and it feels possible:
1) you start somehow from the full screen size
2) you remove padding
3) you divice by #grid, math.floor() it, *#grid you have some leftover you might not even have to care about, because your final stuff is in a CenterContainer.

May be you'd just need to re-order your containers:
Previously: qrmessage = CenterContainer > FrameContainer(white background, padding) > qrwidget
Could be: qrmessage = FrameContainer(white background, padding(a)) > CenterContainer  > qrwidget

(a) this padding might not even need to be used in FrameContainer, it would just be used for the computation of the qrwidget width/height

jperon:
> May be you'd just need to re-order your containers:
> Previously: qrmessage = CenterContainer > FrameContainer(white background, padding) > qrwidget
> Could be: qrmessage = FrameContainer(white background, padding(a)) > CenterContainer > qrwidget

That way, qrmessage should always be fullscreen ; with CenterContainer > FrameContainer, it's possible to have it centered on screen, not necessarily maximized.

frontend/ui/widget/qrwidget.lua
poire-z: scaleBySize just multiplies by a static factor made out of the `Screen>DPI` value and the screen w/h.
It's just so that we can use values like 5 or 10, which, once scaleBySize() should work/look just magically the same on large screen devices.
(I don't really get how that formula is made out, I just know its *1 on my emulator, and *1.79 on my Kobo :)

jperon:
Even without changing containers order, I can get `scale_factor = 1`, but then `scale` isn't an integer, and I get the result you may see on the image I just sent.

frontend/ui/widget/qrwidget.lua
poire-z: > I get "parasite" lines between "pixels". 

Looks like you're painting white squares on a black background. Better reverse that: create a BB, fill it white, paint black square.
And you might miss 1px (due to some rounding error?) in the squares you paint, so the parasite lines.

poire-z:
> That way, qrmessage should always be fullscreen ; with CenterContainer > FrameContainer, it's possible to have it centered on screen, not necessarily maximized.

So, have:
CenterContainer (fullscreen) > FrameContainer (qrmessage specified size, white background) > CenterContainer (dimen=same as parent FrameContainer) > qrwidget.

> I can get scale_factor = 1, but then scale isn't an integer, and I get the result you may see on the image I just sent.

Yep. So, your target is having scale_factor = 1 and scale (square_size) be an integer. That's only that way you'll have sharp edges.
scale(square_size) wanted to be an integer, you'll have to use math.floor(), your qrwidget will then be smaller than the ideal width/height, and a CenterContainer will center it, inside its parent FrameContainer with a white background.

jperon:
I think that time it's good.

frontend/ui/widget/qrwidget.lua
```diff
@@ -27,17 +27,19 @@ function QRWidget:init()
         return
     else
         local scale
-        if self.width then
+        if self.scale_factor then
+            scale = self.scale_factor
```
poire-z, line 31: I don't see the relation between scale (badly named, I take it is the individual square_size, right?) and self.scale_factor (the one used by ImageWidget)

jperon: OK, I was confusing two problems (scale and square size: I thought I had to care about scaling "manually", and I saw square size as the mean to get it).

poire-z:
OK, looks alright.

poire-z:
Some issue with the CircleCI "docs and coverage" steps since this PR was merged:
https://github.com/koreader/koreader/commits/master
```
Checking out koreader/doc for update.
Cloning into 'koreader_doc'...
remote: Enumerating objects: 123, done.        
remote: Counting objects: 100% (123/123), done.        
remote: Compressing objects: 100% (12/12), done.        
remote: Total 161 (delta 111), reused 123 (delta 111), pack-reused 38        
Receiving objects: 100% (161/161), 132.41 KiB | 0 bytes/s, done.
Resolving deltas: 100% (132/132), done.
Checking connectivity... done.
~/project/doc ~/project
reading configuration from config.ld
Failed to generate documents...

Exited with code exit status 1

CircleCI received exit code 1
```
Dunno if it's related to some fancy doc string, or totally unrelated. @Frenzie any idea ?

poire-z:
Or something above that error (that I don't see in previous logs):
```
~/project/l10n ~/project
M	templates/koreader.pot
Previous HEAD position was 498b0ed... Update translation files
Switched to branch 'master'
Your branch is behind 'origin/master' by 9 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)
[master ec24e47] Updated translation source file
 1 file changed, 10 insertions(+), 10 deletions(-)
To https://****************************************@github.com/koreader/koreader-translations.git
 ! [rejected]        master -> master (non-fast-forward)
error: failed to push some refs to 'https://****************************************@github.com/koreader/koreader-translations.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

Translation update pushed.
```

Frenzie:
@NiLuJe Did you do something with the translations repo?

Frenzie:
@poire-z Seems to be working fine now, probably just a random occurrence in the 5-10 mins it takes to run tests (i.e., presumably it's only in that very specific log you looked at):

```
Updating translation source file.
************ CMAKE_MAKE_PROGRAM: "/usr/bin/ninja" **********
git submodule update --remote l10n
Warning: Permanently added the RSA host key for IP address '140.82.114.4' to the list of known hosts.
Submodule path 'l10n': checked out '498b0ed68fbc5ada544f3e2879817fa147cc6bf0'
mkdir -p l10n/templates
xgettext --from-code=utf-8 \
	--keyword=C_:1c,2 --keyword=N_:1,2 --keyword=NC_:1c,2,3 \
	--add-comments=@translators \
	reader.lua `find frontend -iname "*.lua" | sort` \
	`find plugins -iname "*.lua" | sort` \
	`find tools -iname "*.lua" | sort` \
	-o l10n/templates/koreader.pot
~/project/l10n ~/project
M	templates/koreader.pot
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
[master c407031] Updated translation source file
 1 file changed, 11 insertions(+), 11 deletions(-)
```

NiLuJe:
Nope, I checked and I can't actually do much on the main repo actually ^^.

Issue #6831, opened 2020-10-26
Progress bar ticks off by one page (sort of)


Jellby:
* KOReader version: 2020.10.1
* Device: Desktop

#### Issue

This is a matter of taste to some extent, but I think the progress bar ticks (and maybe the bar itself) are off by 1 page. Particularly noticeable in the skim document widget.

What I mean is that I would expect the ticks to appear at the "beginning" of the page. So a tick on page 1 should be at the left edge, and a tick on the last page should *not* be at the right edge, but one "page" shy of it. It seems to be the opposite case right now, because the position of the ticks is computed as `page/total`, which gives 1 for the last page and never gives 0, because there's no page 0. I think I'd prefer `(page-1)/total`.

But then maybe how to draw the bar itself and how to match tapping positions to pages should be changed accordingly? As it is now, when on the first page of chapter the bar looks something like this:
```
#####|#########|____|_____
```
This looks "clean", but is is the best? I'd interpret it as "the chapter ends here". I'm thinking that maybe this would convey more "realistic" information:
```
#####|#########|#___|_____
```
i.e., that this is the first page of a new chapter (except that we don't know if the chapter starts at the top or bottom of the page... but I guess assuming the top is safer). If this is ok, then probably just changing the ticks would be enough.

NiLuJe:
AFAIU, the assumption is that the filled-in portion of the bar corresponds to *read* content. Which makes sense in my mind, but then I'm probably half-unconsciously formatted to understand it this way because it's always been displayed that way ;p.

But I do get what you're saying ^^.

NiLuJe:
(And there's *possibly* a quirk in place to make sure 0/100 are filled-in).

Jellby:
An is that intentional? What I would expect, for a book with 10 pages and chapters *starting* on pages 1, 5 and 10, is something like:

on page 1 (start of first chapter):
```
|#___|_____|_
```
on page 9 (end of second chapter, as far as we can guess):
```
|####|#####|_
```
and of course on page 10:
```
|####|#####|#
```
So the bar would be the same as it is now, I believe, never completely empty, and completely full on the last page. It's just the ticks that seem off.

I somehow consider that the current page is already "read". If that is not liked, I'd suggest maybe marking the current page with a lighter gray, but still darker than white (so it would be the same but replacing the last `#` with something else like `¤`, but this is just a mockup anyway).

NiLuJe:
[...]

PR #3292, opened 2017-10-02
credocument reader optimisation (avoid multiple TOC builds)

poire-z:
Avoid unnecessary work in `ReaderView:onSetViewMode() `and `ReaderRolling:updatePos() `that would result in TOC being reset and rebuilt (which can take time on books with huge TOC) during reader setup.
See https://github.com/koreader/koreader/issues/3261#issuecomment-333560902.
(IT works fine in my tests and it looks simple enough, but given the heavy use of indirections in readerui/readerview setup - postSomethingCallback, events propagated to all widgets... - it's hard to foresee side effects...)
credocument.lua updated just to logically use `DCREREADER_VIEW_MODE `(and to mention the value must be the same as in readerview) , which defaults to `page`, even if we know that `scroll` mode currently crashes.

PR #6847, opened 2020-11-04
Support for hiding (skipping) non-linear EPUB fragments

### LINKED ISSUE #6809: FR: Support for `linear="no"` files in EPUB
The EPUB specification provides for "non-linear" documents, files inside the EPUB that are not included in the normal flow of the text (turning pages from first to last), but that can be accessed though the TOC or hyperlinks. As far as I can see, this property is not supported by KOReader (or most other reading systems), and the non-linear files are treated as any other file. See https://www.w3.org/publishing/epub/epub-packages.html#attrdef-itemref-linear (Note that non-linear documents must still be in the spine, so it is always clear what should be shown when paging forward/backward past their limits.)

I'd like non-linear documents to be excluded from the normal flow, skipped when paging forward/backward and not counted in the total number of pages for the book or the chapter. If one "visits" these non-linear documents, there should be some indication that the "pages" are now outside the normal flow.

For instance, a book could have several chapters and after the final one a non-linear document with end-notes. When reading the book on page `140/322`, I find a link to an end-note, if I follow the link, I may be in page `[3/18]`. If I page forward to page `[18/18]`, and once again, nothing happens, because that's not the "end of the book" it's just the end of this out-of-flow document. If I page backward to `[1/18]` and one again, I go back to the "normal flow", to page `322/322`. And if I now page forward again, the "end of book" popup appears, because this is the end of the book (the last linear document). Of course I could instead "return to previous position" to go back to page `140/322` and continue reading the main text.

**Describe alternatives you've considered**
Well, the alternative is not doing anything, like currently, and having the book with 322+18 = 340 pages, and after page 322 forcing me to page through 18 pages of end-notes (which I may have already read) to reach the end of the book. Another alternative is excluding the non-linear document from normal paging, but not affecting the number of pages, which would be confusing if the book ends up finishing on page `322/340`.

Jellby:
This is the frontend implementation of https://github.com/koreader/koreader/issues/6809, to be merged after https://github.com/koreader/koreader-base/pull/1228 (see https://github.com/koreader/crengine/pull/391/ too)

The core of the implementation is pretty basic: just create `Document:getPrevPage` and `Document:getNextPage`, which return the prev/next page number. By default they are `page-1` and `page+1`, but if a document contains non-linear flows, they can be skipped. If already inside a non-linear flow, paging works normally, until the boundary is reached, where it continues on the main (linear) flow, skipping other non-linear flows that may be in-between. The non-linear pages are still accessible through links, goto, skim, bookmarks... it's only page back/forward that are affected (if enabled).

Some conventions (that can be discussed and changed):

* In the footer, pages on the linear flow as indicated as `x // y` (double slash), and `x` and `y` refer only to the linear pages (non-linear pages are not counted here). The intent is to succinctly indicate that there are uncounted pages.
* When inside a non-linear flow, the footer shows `[x / y]z`, where `x` and `y` refer only to this particular non-linear flow, and `z` is the number of the flow.
* Pages/time to end of chapter/book are counted as if paging through.
* In toc and bookmarks, page numbers refer to the linear flow. Pages in non-linear flows are indicated as `[x]y` where `x` is the "local" page number and "y" is the number of the flow.
* In goto, a plain number leads to the "global" number. The `[x]y` format is supported, in which case `y=0` corresponds to the linear flow.
* In the skim bar, non-linear pages are marked with a darker background.
* ... and maybe something else I'm forgetting.

Setting help
Screenshot_20201104_184445

Footer, linear
Screenshot_20201104_184303

Footer, non-linear
Screenshot_20201104_184419

Go to window
Screenshot_20201104_184503

Skim window
Screenshot_20201107_161725

Bookmarks
Screenshot_20201104_184521

Closes #6809

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6847)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -257,6 +257,13 @@ function ReaderBookmark:onShowBookmark()
                 page = self.ui.pagemap:getXPointerPageLabel(page, true)
             else
                 page = self.ui.document:getPageFromXPointer(page)
+                if self.ui.document:hasHiddenFlows() then
+                    local flow = self.ui.document:getPageFlow(page)
+                    page = self.ui.document:getPageNumberInFlow(page)
+                    if flow > 0 then
+                        page = T("[%1]%2", page, flow)
+                    end
+                end
             end
         end
         if v.text == nil or v.text == "" then
@@ -491,6 +498,13 @@ function ReaderBookmark:updateBookmark(item)
     for i=1, #self.bookmarks do
         if item.datetime == self.bookmarks[i].datetime and item.page == self.bookmarks[i].page then
             local page = self.ui.document:getPageFromXPointer(item.updated_highlight.pos0)
+            if self.ui.document:hasHiddenFlows() then
+                local flow = self.ui.document:getPageFlow(page)
+                page = self.ui.document:getPageNumberInFlow(page)
+                if flow > 0 then
+                    page = T("[%1]%2", page, flow)
+                end
+            end
             local new_text = item.updated_highlight.text
             self.bookmarks[i].page = item.updated_highlight.pos0
             self.bookmarks[i].pos0 = item.updated_highlight.pos0
@@ -517,6 +531,13 @@ function ReaderBookmark:renameBookmark(item, from_highlight)
                     local page = item.page
                     if not self.ui.document.info.has_pages then
                         page = self.ui.document:getPageFromXPointer(page)
+                        if self.ui.document:hasHiddenFlows() then
+                            local flow = self.ui.document:getPageFlow(page)
+                            page = self.ui.document:getPageNumberInFlow(page)
+                            if flow > 0 then
+                                page = T("[%1]%2", page, flow)
+                            end
+                        end
                     end
                     item.text = T(_("Page %1 %2 @ %3"), page, item.notes, item.datetime)
                 end
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -191,7 +191,19 @@ local footerTextGeneratorMap = {
                 return ("%s / %s"):format(footer.ui.pagemap:getCurrentPageLabel(true),
                                           footer.ui.pagemap:getLastPageLabel(true))
             end
-            return ("%d / %d"):format(footer.pageno, footer.pages)
+            if footer.ui.document:hasHiddenFlows() then
+                -- i.e., if we are hiding non-linear fragments and there's anything to hide,
+                local flow = footer.ui.document:getPageFlow(footer.pageno)
+                local page = footer.ui.document:getPageNumberInFlow(footer.pageno)
+                local pages = footer.ui.document:getTotalPagesInFlow(flow)
+                if flow == 0 then
+                    return ("%d // %d"):format(page, pages)
+                else
+                    return ("[%d / %d]%d"):format(page, pages, flow)
+                end
+            else
+                return ("%d / %d"):format(footer.pageno, footer.pages)
+            end
         elseif footer.position then
             return ("%d / %d"):format(footer.position, footer.doc_height)
         end
@@ -200,32 +212,36 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].pages_left
         local left = footer.ui.toc:getChapterPagesLeft(footer.pageno)
-        return prefix .. " " .. (left and left or footer.pages - footer.pageno)
+        return prefix .. " " .. (left or footer.ui.document:getTotalPagesLeft(footer.pageno))
     end,
     percentage = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].percentage
         local digits = footer.settings.progress_pct_format or "0"
-        local string_percentage
-        if not prefix then
-            string_percentage = "%." .. digits .. "f%%"
-        else
-            string_percentage = prefix .. " %." .. digits .. "f%%"
+        local string_percentage = "%." .. digits .. "f%%"
+        if footer.ui.document:hasHiddenFlows() then
+            local flow = footer.ui.document:getPageFlow(footer.pageno)
+            if flow ~= 0 then
+                string_percentage = "[" .. string_percentage .. "]"
+            end
+        end
+        if prefix then
+            string_percentage = prefix .. " " .. string_percentage
         end
         return string_percentage:format(footer.progress_bar.percentage * 100)
     end,
     book_time_to_read = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].book_time_to_read
-        local current_page = footer.ui:getCurrentPage()
-        return footer:getDataFromStatistics(prefix .. " ", footer.pages - current_page)
+        local left = footer.ui.document:getTotalPagesLeft(footer.pageno)
+        return footer:getDataFromStatistics(prefix .. " ", left)
     end,
     chapter_time_to_read = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].chapter_time_to_read
         local left = footer.ui.toc:getChapterPagesLeft(footer.pageno)
         return footer:getDataFromStatistics(
-            prefix .. " ", (left and left or footer.pages - footer.pageno))
+            prefix .. " ", (left or footer.ui.document:getTotalPagesLeft(footer.pageno)))
     end,
     mem_usage = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
@@ -1724,18 +1740,44 @@ function ReaderFooter:setTocMarkers(reset)
     if self.settings.disable_progress_bar or self.settings.progress_style_thin then return end
     if reset then
         self.progress_bar.ticks = nil
-        self.pages = self.view.document:getPageCount()
+        self.pages = self.ui.document:getPageCount()
     end
     if self.settings.toc_markers then
         self.progress_bar.tick_width = Screen:scaleBySize(self.settings.toc_markers_width)
         if self.progress_bar.ticks ~= nil then -- already computed
             return
         end
-        self.progress_bar.ticks = {}
-        if self.ui.toc then
-            self.progress_bar.ticks = self.ui.toc:getTocTicksFlattened()
+        if self.ui.document:hasHiddenFlows() and self.pageno then
+            local flow = self.ui.document:getPageFlow(self.pageno)
+            self.progress_bar.ticks = {}
+            if self.ui.toc then
+                -- filter the ticks to show only those in the current flow
+                for n, pageno in ipairs(self.ui.toc:getTocTicksFlattened()) do
+                    if self.ui.document:getPageFlow(pageno) == flow then
+                        table.insert(self.progress_bar.ticks, self.ui.document:getPageNumberInFlow(pageno))
+                    end
+                end
+            end
+            self.progress_bar.last = self.ui.document:getTotalPagesInFlow(flow)
+        else
+            if self.ui.toc then
+                self.progress_bar.ticks = self.ui.toc:getTocTicksFlattened()
+            end
+            if self.view.view_mode == "page" then
+                self.progress_bar.last = self.pages or self.ui.document:getPageCount()
+            else
+                -- in scroll mode, convert pages to positions
+                if self.ui.toc then
+                    self.progress_bar.ticks = {}
+                    for n, pageno in ipairs(self.ui.toc:getTocTicksFlattened()) do
+                        local idx = self.ui.toc:getTocIndexByPage(pageno)
+                        local pos = self.ui.document:getPosFromXPointer(self.ui.toc.toc[idx].xpointer)
+                        table.insert(self.progress_bar.ticks, pos)
+                    end
+                end
+                self.progress_bar.last = self.doc_height or self.ui.document.info.doc_height
+            end
         end
-        self.progress_bar.last = self.pages or self.view.document:getPageCount()
     else
         self.progress_bar.ticks = nil
     end
@@ -1770,7 +1812,14 @@ end
 
 function ReaderFooter:updateFooterPage(force_repaint, force_recompute)
     if type(self.pageno) ~= "number" then return end
-    self.progress_bar.percentage = self.pageno / self.pages
+    if self.ui.document:hasHiddenFlows() then
+        local flow = self.ui.document:getPageFlow(self.pageno)
+        local page = self.ui.document:getPageNumberInFlow(self.pageno)
+        local pages = self.ui.document:getTotalPagesInFlow(flow)
+        self.progress_bar.percentage = page / pages
+    else
+        self.progress_bar.percentage = self.pageno / self.pages
+    end
     self:updateFooterText(force_repaint, force_recompute)
 end
 
@@ -1882,19 +1931,39 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
     end
 end
 
+function ReaderFooter:onTocReset()
+    self:setTocMarkers(true)
+    if self.view.view_mode == "page" then
+        self:updateFooterPage()
+    else
+        self:updateFooterPos()
+    end
+end
+
 function ReaderFooter:onPageUpdate(pageno)
+    local toc_markers_update = false
+    if self.ui.document:hasHiddenFlows() then
+        local flow = self.pageno and self.ui.document:getPageFlow(self.pageno)
+        local new_flow = pageno and self.ui.document:getPageFlow(pageno)
+        if pageno and new_flow ~= flow then
+            toc_markers_update = true
+        end
+    end
     self.pageno = pageno
-    self.pages = self.view.document:getPageCount()
+    self.pages = self.ui.document:getPageCount()
+    if toc_markers_update then
+        self:setTocMarkers(true)
+    end
     self.ui.doc_settings:saveSetting("doc_pages", self.pages) -- for Book information
     self:updateFooterPage()
 end
 
 function ReaderFooter:onPosUpdate(pos, pageno)
     self.position = pos
-    self.doc_height = self.view.document.info.doc_height
+    self.doc_height = self.ui.document.info.doc_height
     if pageno then
         self.pageno = pageno
-        self.pages = self.view.document:getPageCount()
+        self.pages = self.ui.document:getPageCount()
         self.ui.doc_settings:saveSetting("doc_pages", self.pages) -- for Book information
     end
     self:updateFooterPos()
@@ -2065,7 +2134,7 @@ function ReaderFooter:refreshFooter(refresh, signal)
     -- We *do* need to ensure we at least re-compute the footer layout, though, especially when going from visible to invisible...
     self:onUpdateFooter(refresh and not signal, refresh and signal)
     if signal then
-        self.ui:handleEvent(Event:new("SetPageBottomMargin", self.view.document.configurable.b_page_margin))
+        self.ui:handleEvent(Event:new("SetPageBottomMargin", self.ui.document.configurable.b_page_margin))
     end
 end
 
--- a/frontend/apps/reader/modules/readergoto.lua
+++ b/frontend/apps/reader/modules/readergoto.lua
@@ -62,6 +62,12 @@ function ReaderGoto:onShowGotoDialog()
     self.goto_dialog = InputDialog:new{
         title = dialog_title,
         input_hint = input_hint,
+        description = self.document:hasHiddenFlows() and
+            _([[
+x for an absolute page number
+[x] for a page number in the main (linear) flow
+[x]y for a page number in the non-linear fragment y]])
+            or nil,
         buttons = {
             {
                 {
@@ -134,20 +140,51 @@ function ReaderGoto:gotoPage()
             end
         end
         self:close()
+    elseif self.ui.document:hasHiddenFlows() then
+        -- if there are hidden flows, we accept the syntax [x]y
+        -- for page number x in flow number y (y defaults to 0 if not present)
+        local flow
+        number, flow = string.match(page_number, "^ *%[(%d+)%](%d*) *$")
+        flow = tonumber(flow) or 0
+        number = tonumber(number)
+        if number then
+            if self.ui.document.flows[flow] ~= nil then
+                if number < 1 or number > self.ui.document:getTotalPagesInFlow(flow) then
+                    return
+                end
+                local page = 0
+                -- in flow 0 (linear), we count pages skipping non-linear flows,
+                -- in a non-linear flow the target page is immediate
+                if flow == 0 then
+                    for i=1, number do
+                        page = self.ui.document:getNextPage(page)
+                    end
+                else
+                    page = self.ui.document:getFirstPageInFlow(flow) + number - 1
+                end
+                if page > 0 then
+                    self.ui:handleEvent(Event:new("GotoPage", page))
+                    self:close()
+                end
+            end
+        end
     end
 end
 
 function ReaderGoto:onGoToBeginning()
-    self.ui.link:addCurrentLocationToStack()
-    self.ui:handleEvent(Event:new("GotoPage", 1))
+    local new_page = self.ui.document:getNextPage(0)
+    if new_page then
+        self.ui.link:addCurrentLocationToStack()
+        self.ui:handleEvent(Event:new("GotoPage", new_page))
+    end
     return true
 end
 
 function ReaderGoto:onGoToEnd()
-    local endpage = self.document:getPageCount()
-    if endpage then
+    local new_page = self.ui.document:getPrevPage(0)
+    if new_page then
         self.ui.link:addCurrentLocationToStack()
-        self.ui:handleEvent(Event:new("GotoPage", endpage))
+        self.ui:handleEvent(Event:new("GotoPage", new_page))
     end
     return true
 end
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -764,7 +764,7 @@ function ReaderRolling:onUpdatePos()
     -- so updatePos() has good info and can reposition
     -- the previous xpointer accurately:
     self.ui.document:getCurrentPos()
-    -- Otherwise, _readMetadata() would do that, but the positionning
+    -- Otherwise, _readMetadata() would do that, but the positioning
     -- would not work as expected, for some reason (it worked
     -- previously because of some bad setDirty() in ConfigDialog widgets
     -- that were triggering a full repaint of crengine (so, the needed
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -55,7 +55,8 @@ local ReaderRolling = InputContainer:new{
     -- With visible_pages=2, in 2-pages mode, ensure the first
     -- page is always odd or even (odd is logical to avoid a
     -- same page when turning first 2-pages set of document)
-    odd_or_even_first_page = 1 -- 1 = odd, 2 = even, nil or others = free
+    odd_or_even_first_page = 1, -- 1 = odd, 2 = even, nil or others = free
+    hide_nonlinear_flows = nil,
 }
 
 function ReaderRolling:init()
@@ -206,7 +207,7 @@ function ReaderRolling:onReadSettings(config)
         self.setupXpointer = function()
             self.xpointer = self.ui.document:getXPointer()
             if self.view.view_mode == "page" then
-                self.ui:handleEvent(Event:new("PageUpdate", 1))
+                self.ui:handleEvent(Event:new("PageUpdate", self.ui.document:getNextPage(0)))
             end
         end
     end
@@ -229,6 +230,12 @@ function ReaderRolling:onReadSettings(config)
     self.visible_pages = config:readSetting("visible_pages") or
         G_reader_settings:readSetting("copt_visible_pages") or 1
     self.ui.document:setVisiblePageCount(self.visible_pages)
+
+    self.hide_nonlinear_flows = config:readSetting("hide_nonlinear_flows")
+    if self.hide_nonlinear_flows == nil then
+        self.hide_nonlinear_flows = G_reader_settings:isTrue("hide_nonlinear_flows")
+    end
+    self.ui.document:setHideNonlinearFlows(self.hide_nonlinear_flows)
 
     -- Set a callback to allow showing load and rendering progress
     -- (this callback will be cleaned up by cre.cpp closeDocument(),
@@ -302,10 +309,14 @@ function ReaderRolling:onSaveSettings()
     self.ui.doc_settings:saveSetting("show_overlap_enable", self.show_overlap_enable)
     self.ui.doc_settings:saveSetting("inverse_reading_order", self.inverse_reading_order)
     self.ui.doc_settings:saveSetting("visible_pages", self.visible_pages)
+    self.ui.doc_settings:saveSetting("hide_nonlinear_flows", self.hide_nonlinear_flows)
 end
 
 function ReaderRolling:onReaderReady()
     self:setupTouchZones()
+    if self.hide_nonlinear_flows then
+        self.ui.document:cacheFlows()
+    end
     self.setupXpointer()
 end
 
@@ -457,6 +468,31 @@ You can set how many lines are shown.]])
         help_text = _([[When page overlap is enabled, some lines from the previous pages are shown on the next page.]]),
         sub_item_table = page_overlap_menu,
     }
+    if self.ui.document:hasNonLinearFlows() then
+        local hide_nonlinear_text = _("When this option is enabled, if a document contains non-linear fragments, they will be hidden from the normal page flow, but they are still accessible through links, Toc or Go to. This currently works only in single-page, non-scrolling mode.")
+        menu_items.hide_nonlinear_flows = {
+            text = _("Hide non-linear fragments"),
+            enabled_func = function()
+                return self.view.view_mode == "page" and self.ui.document:getVisiblePageCount() == 1
+            end,
+            checked_func = function() return self.hide_nonlinear_flows end,
+            callback = function()
+                self:onToggleHideNonlinear()
+            end,
+            hold_callback = function()
+                UIManager:show(ConfirmBox:new{
+                    text = T(
+                        hide_nonlinear_text .. _("\n\nSet default hide non-linear fragments to %1?"),
+                        self.hide_nonlinear_flows and _("enabled") or _("disabled")
+                    ),
+                    ok_callback = function()
+                        G_reader_settings:saveSetting("hide_nonlinear_flows", self.hide_nonlinear_flows)
+                    end,
+                })
+            end,
+            help_text = hide_nonlinear_text,
+        }
+    end
 end
 
 function ReaderRolling:getLastPercent()
@@ -526,8 +562,18 @@ end
 function ReaderRolling:onGotoNextChapter()
     local visible_page_count = self.ui.document:getVisiblePageCount()
     local pageno = self.current_page + (visible_page_count > 1 and 1 or 0)
-    local new_page = self.ui.toc:getNextChapter(pageno, 0)
-    if new_page then
+    local new_page
+    if self.ui.document:hasHiddenFlows() then
+        -- Find next chapter start
+        new_page = self.ui.document:getNextPage(pageno)
+        while new_page > 0 do
+            if self.ui.toc:isChapterStart(new_page) then break end
+            new_page = self.ui.document:getNextPage(new_page)
+        end
+    else
+        new_page = self.ui.toc:getNextChapter(pageno) or 0
+    end
+    if new_page > 0 then
         self.ui.link:addCurrentLocationToStack()
         self:onGotoPage(new_page)
     end
@@ -536,8 +582,18 @@ end
 
 function ReaderRolling:onGotoPrevChapter()
     local pageno = self.current_page
-    local new_page = self.ui.toc:getPreviousChapter(pageno, 0)
-    if new_page then
+    local new_page
+    if self.ui.document:hasHiddenFlows() then
+        -- Find previous chapter start
+        new_page = self.ui.document:getPrevPage(pageno)
+        while new_page > 0 do
+            if self.ui.toc:isChapterStart(new_page) then break end
+            new_page = self.ui.document:getPrevPage(new_page)
+        end
+    else
+        new_page = self.ui.toc:getPreviousChapter(pageno) or 0
+    end
+    if new_page > 0 then
         self.ui.link:addCurrentLocationToStack()
         self:onGotoPage(new_page)
     end
@@ -721,7 +777,23 @@ function ReaderRolling:onGotoViewRel(diff)
         else
             diff = math.floor(diff)
         end
-        self:_gotoPage(self.current_page + diff*page_count)
+        local new_page = self.current_page
+        if self.ui.document:hasHiddenFlows() then
+            local test_page
+            for i=1, math.abs(diff*page_count) do
+                if diff > 0 then
+                    test_page = self.ui.document:getNextPage(new_page)
+                else
+                    test_page = self.ui.document:getPrevPage(new_page)
+                end
+                if test_page > 0 then
+                    new_page = test_page
+                end
+            end
+        else
+            new_page = new_page + diff*page_count
+        end
+        self:_gotoPage(new_page)
         if diff > 0 and old_page == self.current_page then
             self.ui:handleEvent(Event:new("EndOfBook"))
         end
@@ -783,12 +855,13 @@ function ReaderRolling:updatePos()
     local new_height = self.ui.document.info.doc_height
     local new_page = self.ui.document.info.number_of_pages
     if self.old_doc_height ~= new_height or self.old_page ~= new_page then
+        if self.hide_nonlinear_flows then
+            self.ui.document:cacheFlows()
+        end
         self:_gotoXPointer(self.xpointer)
         self.old_doc_height = new_height
         self.old_page = new_page
         self.ui:handleEvent(Event:new("UpdateToc"))
-        self.view.footer:setTocMarkers(true)
-        self.view.footer:onUpdateFooter()
     end
     self:updateTopStatusBarMarkers()
     UIManager:setDirty(self.view.dialog, "partial")
@@ -808,7 +881,6 @@ function ReaderRolling:onChangeViewMode()
     self.old_doc_height = self.ui.document.info.doc_height
     self.old_page = self.ui.document.info.number_of_pages
     self.ui:handleEvent(Event:new("UpdateToc"))
-    self.view.footer:setTocMarkers(true)
     if self.xpointer then
         self:_gotoXPointer(self.xpointer)
         -- Ensure a whole screen refresh is always enqueued
@@ -1308,6 +1380,20 @@ Note that %1 (out of %2) xpaths from your bookmarks and highlights have been nor
             end)
         end,
     })
+end
+
+function ReaderRolling:onToggleHideNonlinear()
+    self.hide_nonlinear_flows = not self.hide_nonlinear_flows
+    self.ui.document:setHideNonlinearFlows(self.hide_nonlinear_flows)
+    -- The document may change due to forced pagebreaks between flows being
+    -- added or removed, so we need to find our location
+    self:onUpdatePos()
+    -- Even if the document doesn't change, we must ensure that the
+    -- flow and call caches are cleared, to get the right page numbers,
+    -- which may have changed, and the correct flow structure. Also,
+    -- the footer needs updating, and TOC markers may come or go.
+    self.ui.document:cacheFlows()
+    self.ui:handleEvent(Event:new("UpdateToc"))
 end
 
 -- Duplicated in ReaderPaging
--- a/frontend/apps/reader/modules/readertoc.lua
+++ b/frontend/apps/reader/modules/readertoc.lua
@@ -62,6 +62,7 @@ end
 
 function ReaderToc:onUpdateToc()
     self:resetToc()
+    self.ui:handleEvent(Event:new("TocReset"))
 
     --- @note: Let this propagate, plugins/statistics uses it to react to changes in document pagination
     --return true
@@ -168,25 +169,35 @@ function ReaderToc:validateAndFixToc()
             nb_bogus = nb_bogus + 1
             -- See how many pages we'd need fixing on either side
             local nb_prev = 0
+            local nb_prev_main = 0
             for j = i-1, first, -1 do
                 local ppage = toc[j].fixed_page or toc[j].page
                 if ppage <= page then
                     break
                 else
                     nb_prev = nb_prev + 1
+                    if self.ui.document:getPageFlow(ppage) == 0 then
+                        nb_prev_main = nb_prev_main + 1
+                    end
                 end
             end
-            local nb_next = 1
-            for j = i+1, last do
+            local nb_next = 0
+            local nb_next_main = 0
+            for j = i, last do
                 local npage = toc[j].fixed_page or toc[j].page
                 if npage >= cur_page then
                     break
                 else
                     nb_next = nb_next + 1
+                    if self.ui.document:getPageFlow(npage) == 0 then
+                        nb_next_main = nb_next_main + 1
+                    end
                 end
             end
             logger.dbg("BOGUS TOC:", i, page, "<", i-1, cur_page, "-", nb_prev, nb_next)
-            if nb_prev <= nb_next then -- less changes when fixing previous pages
+            -- Note: by comparing only the entries that belong to the main (linear) flow
+            -- we give priority to moving non-linear bogus entries
+            if nb_prev_main <= nb_next_main then -- less changes when fixing previous pages
                 local fixed_page
                 if i-nb_prev-1 >= 1 then
                     fixed_page = toc[i-nb_prev-1].fixed_page or toc[i-nb_prev-1].page
@@ -436,21 +447,46 @@ function ReaderToc:isChapterEnd(cur_pageno)
 end
 
 function ReaderToc:getChapterPagesLeft(pageno)
-    --if self:isChapterEnd(pageno) then return 0 end
-    local next_chapter = self:getNextChapter(pageno)
-    if next_chapter then
-        next_chapter = next_chapter - pageno - 1
+    if self.ui.document:hasHiddenFlows() then
+        -- Count pages until new chapter
+        local pages_left = 0
+        local test_page = self.ui.document:getNextPage(pageno)
+        while test_page > 0 do
+            pages_left = pages_left + 1
+            if self:isChapterStart(test_page) then
+                return pages_left - 1
+            end
+            test_page = self.ui.document:getNextPage(test_page)
+        end
+    else
+        local next_chapter = self:getNextChapter(pageno)
+        if next_chapter then
+            next_chapter = next_chapter - pageno - 1
+        end
+        return next_chapter
     end
-    return next_chapter
 end
 
 function ReaderToc:getChapterPagesDone(pageno)
     if self:isChapterStart(pageno) then return 0 end
-    local previous_chapter = self:getPreviousChapter(pageno)
-    if previous_chapter then
-        previous_chapter = pageno - previous_chapter
+    if self.ui.document:hasHiddenFlows() then
+        -- Count pages until chapter start
+        local pages_done = 0
+        local test_page = self.ui.document:getPrevPage(pageno)
+        while test_page > 0 do
+            pages_done = pages_done + 1
+            if self:isChapterStart(test_page) then
+                return pages_done
+            end
+            test_page = self.ui.document:getPrevPage(test_page)
+        end
+    else
+        local previous_chapter = self:getPreviousChapter(pageno)
+        if previous_chapter then
+            previous_chapter = pageno - previous_chapter
+        end
+        return previous_chapter
     end
-    return previous_chapter
 end
 
 function ReaderToc:updateCurrentNode()
@@ -489,7 +525,34 @@ function ReaderToc:onShowToc()
         for _,v in ipairs(self.toc) do
             v.text = self.toc_indent:rep(v.depth-1)..self:cleanUpTocTitle(v.title)
             v.mandatory = v.page
-            if v.orig_page then -- bogus page fixed: show original page number
+            if self.ui.document:hasHiddenFlows() then
+                local flow = self.ui.document:getPageFlow(v.page)
+                if v.orig_page then -- bogus page fixed: show original page number
+                    -- This is an ugly piece of code, which can result in an ugly TOC,
+                    -- but it shouldn't be needed very often, only when bogus page numbers
+                    -- are fixed, and then showing everything gets complicated
+                    local orig_flow = self.ui.document:getPageFlow(v.orig_page)
+                    if flow == 0 and orig_flow == flow then
+                        v.mandatory = T("(%1) %2", self.ui.document:getPageNumberInFlow(v.orig_page), self.ui.document:getPageNumberInFlow(v.page))
+                    elseif flow == 0 and orig_flow ~= flow then
+                        v.mandatory = T("[%1]%2", self.ui.document:getPageNumberInFlow(v.orig_page), self.ui.document:getPageFlow(v.orig_page))
+                    elseif flow > 0 and orig_flow == flow then
+                        v.mandatory = T("[(%1) %2]%3", self.ui.document:getPageNumberInFlow(v.orig_page),
+                                                       self.ui.document:getPageNumberInFlow(v.page), self.ui.document:getPageFlow(v.page))
+                    else
+                        v.mandatory = T("([%1]%2) [%3]%4", self.ui.document:getPageNumberInFlow(v.orig_page), self.ui.document:getPageFlow(v.orig_page),
+                                                           self.ui.document:getPageNumberInFlow(v.page), self.ui.document:getPageFlow(v.page))
+                    end
+                else
+                    -- Plain numbers for the linear entries,
+                    -- for non-linear entries we use the same syntax as in the Go to dialog
+                    if flow == 0 then
+                        v.mandatory = self.ui.document:getPageNumberInFlow(v.page)
+                    else
+                        v.mandatory = T("[%1]%2", self.ui.document:getPageNumberInFlow(v.page), self.ui.document:getPageFlow(v.page))
+                    end
+                end
+            elseif v.orig_page then -- bogus page fixed: show original page number
                 v.mandatory = T("(%1) %2", v.orig_page, v.page)
             end
             if self.ui.pagemap and self.ui.pagemap:wantsPageLabels() then
--- a/frontend/apps/reader/modules/readertoc.lua
+++ b/frontend/apps/reader/modules/readertoc.lua
@@ -185,7 +185,7 @@ function ReaderToc:validateAndFixToc()
                     nb_next = nb_next + 1
                 end
             end
-            logger.dbg("BOGUS TOC:", i, page, ">", i-1, cur_page, "-", nb_prev, nb_next)
+            logger.dbg("BOGUS TOC:", i, page, "<", i-1, cur_page, "-", nb_prev, nb_next)
             if nb_prev <= nb_next then -- less changes when fixing previous pages
                 local fixed_page
                 if i-nb_prev-1 >= 1 then
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -58,6 +58,11 @@ local CreDocument = Document:new{
     default_css = "./data/cr3.css",
     provider = "crengine",
     provider_name = "Cool Reader Engine",
+
+    hide_nonlinear_flows = false,
+    flows = {},
+    page_in_flow = {},
+    last_linear_page = nil,
 }
 
 -- NuPogodi, 20.05.12: inspect the zipfile content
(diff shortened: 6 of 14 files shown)


frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -256,6 +256,11 @@ function ReaderBookmark:onShowBookmark()
                 page = self.ui.pagemap:getXPointerPageLabel(page, true)
             else
                 page = self.ui.document:getPageFromXPointer(page)
+                local flow = self.ui.document:getPageFlow(page)
+                page = self.ui.document:getPageNumberInFlow(page)
+                if flow > 0 then
+                    page = T("[%1]%2", page, flow)
```
Frenzie, line 262: Might need a translator's comment. Also screenshots would be nice. ;-)

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -436,21 +446,32 @@ function ReaderToc:isChapterEnd(cur_pageno)
...
+    local test_page = self.ui.document:getPrevPage(pageno)
+    while test_page > 0 do
+        pages_done = pages_done + 1
+        if self:isChapterStart(test_page) then
+            return pages_done
+        end
+        test_page = self.ui.document:getPrevPage(test_page)
     end
-    return previous_chapter
 end
```
NiLuJe, line 475: Is there something wrong with the existing approach if `!hasFlows`? If not, can we keep it in this case? ;).

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -536,8 +573,13 @@ end
...
+    -- Find previous chapter start
+    local new_page = self.ui.document:getPrevPage(pageno)
+    while new_page > 0 do
+        if self.ui.toc:isChapterStart(new_page) then break end
+        new_page = self.ui.document:getPrevPage(new_page)
+    end
+    if new_page > 0 then
         self.ui.link:addCurrentLocationToStack()
         self:onGotoPage(new_page)
     end
```
NiLuJe, line 585: Same question about the existing approach for the `!hasFlows` case ;).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1720,9 +1731,16 @@ function ReaderFooter:setTocMarkers(reset)
...
         if self.ui.toc then
-            self.progress_bar.ticks = self.ui.toc:getTocTicksFlattened()
+            -- filter the ticks to show only those in the current flow
+            for i,j in pairs(self.ui.toc:getTocTicksFlattened()) do
+                if self.view.document:getPageFlow(j) == self.flow then
+                    self.progress_bar.ticks[i] = self.view.document:getPageNumberInFlow(j)
+                else
+                    self.progress_bar.ticks[i] = nil
+                end
+            end
```
NiLuJe, line 1741: In the same vein, only relevant if `hasFlows` (well, has and show, I guess, actually) ;).

frontend/apps/reader/modules/readerrolling.lua
NiLuJe: (That was a botched multi-line comment on the two preceding functions ;p).

frontend/apps/reader/modules/readerbookmark.lua
Jellby: I confess I've used `T(...)` without knowing what it means... I don't think this should be translated though, so I guess I should use `("[%1]%2"):format(page, flow)`, as in the footer code?

frontend/apps/reader/modules/readertoc.lua
Jellby: Nothing wrong, I just like it better to have a single solution than lots of `if`s.

frontend/apps/reader/modules/readerbookmark.lua
Frenzie: Actually you confused me by needlessly using the template function here when you're just doing string.format.

> I don't think this should be translated though

You and I simply don't know (pinging @WaseemAlkurdi since I'm curious about Arabic), but even for a language like Dutch I might opt to localize it as `()` instead of `[]`.

Jellby: I copied that from the toc code, which is "translated", but the footer code is not (e.g. you cannot translate `"%d / %d"` to `"%d : %d"`)

frontend/apps/reader/modules/readertoc.lua
NiLuJe: `getPreviousChapter` & co happen to be faster by using the ticks directly, though, instead of walking page-by-page ;)

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: You *should* be able to iterate through `getTocTicksFlattened()` via `ipairs`. If it turns out the array is actually sparse, then we have a host of other buggy code that assumes it wouldn't ;).

frontend/apps/reader/modules/readerbookmark.lua
Frenzie: That would be a bug then.

poire-z:
Mhhhh, _first_ impression is not the best.
It feels too integrated: what I knew and could read and understand easily - is now something that looks really different, with the "flow" word everywhere (even if there are places you succeeded in hiding in methods of document - but then the feeling is just deported to credocument :) I'm jumping all around to see what the regular calls do and the new calls do - to in the end happily discovering it's a no-op - but there's too many indirections until I notice that.

I would have expected stuff to be a bit more like:
```lua
if self.ui.pagemap and self.ui.pagemap:wantsPageLabels() then
    page = self.ui.pagemap:getXPointerPageLabel(page, true)
else if self.ui.hide_non_linear == true and document.hasNonLinearStuff() then
    -- The new stuff I can skip when discovering the code.
    your new code and new methods called only when necessary
else
    -- the regular usually simpler stuff
    page = self.ui.document:getPageFromXPointer(page)
end
```

I'll get into your code :) and may be all that tight integration is needed (the diff is not that large) - but after 45mn reading the diff, I'm not yet passed this first impression :/

Letting my dev impressions on the side for now, my user impressions (haven't tried it, just from the screenshots and what I quickly got from the code):
- I find the the Skim widget is ugly with it's half grey bars
- not fond of `x // y` and the other stuff - but I haven't yet thought about alternatives
- but I and other users should not see the complicated Goto description and all the new stuff when hide_non_linear is not true : everything should look as before when a user has not opt-in for the non-linear new behaviour
- (issue with bookmarks which could have some the normal page, some the tweaked non-linear page number, depending on whether they were made in one mode or the other - but that's a more general issue with our messy Bookmark stuff)
- I think the new menu should go into the first top menu (with TOC, Bookmarks, Reference pages) : and should not be shown when the document does not have any non-linear stuff. It should appear there only when there are (like Reference pages do)

NiLuJe:
> 
>     * I find the the Skim widget is ugly with it's half grey bars
> 

Kinda agree ;p.

Ideas:

* Make it a *background* color (i.e., ticks go above it), only slightly darker than the "usual" empty white.
* If something non-linear was read, make the fill color slightly lighter (àla additive blend).

frontend/document/credocument.lua
```diff
@@ -283,10 +287,127 @@ function CreDocument:updateColorRendering()
...
+    else
+        return Document.getNextPage(self, page)
+    end
+end
+
+function CreDocument:getPrevPage(page)
+    if self:hasFlows() then
+        if page > self:getPageCount()+1 then return 0 end
+        local flow = page > 0 and self:getPageFlow(page) or 0
+        for test_page=(page==0) and self:getPageCount() or page-1, 1, -1 do
```
Galunid, line 326: Seems a bit too complicated, it's probably better to use a helper variable
```lua
local start_index = (page == 0) and self:getPageCount() or page - 1
for test_page=start_index, 1, -1 do
...
```

frontend/apps/reader/modules/readertoc.lua
Jellby: >     * but I and other users should not see the complicated Goto description and all the new stuff when hide_non_linear is not true : everything should look as before when a user has not opt-in for the non-linear new behaviour

That's the case. Everything (but the new menu item) should look the same as it does now when disabled.

>     * (issue with bookmarks which could have some the normal page, some the tweaked non-linear page number, depending on whether they were made in one mode or the other - but that's a more general issue with our messy Bookmark stuff)

Right, but that also happens now if I make some bookmarks with large font and others with small font.

>     * I think the new menu should go into the first top menu (with TOC, Bookmarks, Reference pages) : and should not be shown when the document does not have any non-linear stuff. It should appear there only when there are (like Reference pages do)

Hmm...  But I want to be able to enable it for all books by default, even if the current book does not have it. Also, I thought that this is something that affects the formatting of the book and the reading experience, and it's not a tool for navigation, so it seems to fit better the second menu, doesn't it?

Jellby:
>     * but I and other users should not see the complicated Goto description and all the new stuff when hide_non_linear is not true : everything should look as before when a user has not opt-in for the non-linear new behaviour

That's the case. Everything (but the new menu item) should look the same as it does now when disabled.

>     * (issue with bookmarks which could have some the normal page, some the tweaked non-linear page number, depending on whether they were made in one mode or the other - but that's a more general issue with our messy Bookmark stuff)

Right, but that also happens now if I make some bookmarks with large font and others with small font.

>     * I think the new menu should go into the first top menu (with TOC, Bookmarks, Reference pages) : and should not be shown when the document does not have any non-linear stuff. It should appear there only when there are (like Reference pages do)

Hmm...  But I want to be able to enable it for all books by default, even if the current book does not have it. Also, I thought that this is something that affects the formatting of the book and the reading experience, and it's not a tool for navigation, so it seems to fit better the second menu, doesn't it?

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -457,6 +465,30 @@ You can set how many lines are shown.]])
         help_text = _([[When page overlap is enabled, some lines from the previous pages are shown on the next page.]]),
         sub_item_table = page_overlap_menu,
     }
+    local hide_nonlinear_text = _("When this option is enabled, if a document contains non-linear fragments, \z
+                                   they will be hidden from the normal page flow, \z
+                                   but they are still accessible through links, Toc or Go to. \z
+                                   This currently works only in single-page, non-scrolling mode.")
```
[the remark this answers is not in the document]
Jellby: I wanted to wrap it in the code, but not in the GUI. If there's some other way, I'd be happy to know (I don't know Lua, so I just googled for something).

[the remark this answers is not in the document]
Galunid: Oh, in that case you can use something like that
```
_([[multiline
string]])
```

poire-z:
> Hmm... But I want to be able to enable it for all books by default, even if the current book does not have it.

I think you need a way to show the user that this book has non-linear flows or not. No need for him to toggle it and then try to see if something has changed like total nb of pages to notice/guess that if not, there's no linear flow. So, the presence of this menu is a good indicator the document has them, and that there's a feature that may makes the reading experience better. Seeing that menu always and toggling it and seeing it has no effect won't advertize how nice that feature can be.
(Having it grey/not enabled when the book does not have non-linear stuff can be a bit confusing. Our other menu items are grey when they depend on other features/settings - so a user, seeing it grey, might want to try tweaking other stuff to make it non-grey/enabled.)

As for the hold_callback to enable it on all books - well, you'll do that on a book that has it :) (in cre, you can't enable pdf stuff on all books).

> Also, I thought that this is something that affects the formatting of the book and the reading experience, and it's not a tool for navigation, so it seems to fit better the second menu, doesn't it?

It doesn't at all to me :) How it affects formatting is a side effect I'd say. Most of the things this PR changes are related to pages, TOC, Bookmarks, the skimto widget after all :) things that all are in that first menu.

(For these 2 points, I consider your feature not much different than the Reference page stuff I added - So, it'd be nice they show up, work and are advertize quite similarly.)

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -485,6 +490,11 @@ function ReaderBookmark:updateBookmark(item)
     for i=1, #self.bookmarks do
         if item.datetime == self.bookmarks[i].datetime and item.page == self.bookmarks[i].page then
             local page = self.ui.document:getPageFromXPointer(item.updated_highlight.pos0)
+            local flow = self.ui.document:getPageFlow(page)
+            page = self.ui.document:getPageNumberInFlow(page)
+            if flow > 0 then
+                page = T("[%1]%2", page, flow)
+            end
```
poire-z, line 497: I'd be happier if that kind of things would be written:
```lua
            local page = self.ui.document:getPageFromXPointer(item.updated_highlight.pos0)
            local flow = self.ui.document:getPageFlow(page) -- 0 when no (or not hidden) non-linear flows
            if flow > 0 then
                page = self.ui.document:getPageNumberInFlow(page)
                page = T("[%1]%2", page, flow)
            end
```
or even:
```lua
            local page = self.ui.document:getPageFromXPointer(item.updated_highlight.pos0)
            if self.ui.document:hasHiddenFlows() then
                local flow = self.ui.document:getPageFlow(page) -- 0 when no (or not hidden) non-linear flows
                if flow > 0 then
                    page = self.ui.document:getPageNumberInFlow(page)
                    page = T("[%1]%2", page, flow)
                end
            end
```
no matter the additional function call.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -191,39 +191,49 @@ local footerTextGeneratorMap = {
...
         elseif footer.position then
             return ("%d / %d"):format(footer.position, footer.doc_height)
         end
     end,
     pages_left = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].pages_left
-        local left = footer.ui.toc:getChapterPagesLeft(footer.pageno)
+        local left = footer.ui.toc:getChapterPagesLeft(footer.ui:getCurrentPage())
```
poire-z, line 211: Why that change for this one - and not for the others (like the one above you modified)?
Can footer.pageno be different from footer.ui:getCurrentPage() ?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -438,6 +448,7 @@ function ReaderFooter:init()
         return
     end
 
+    self.flow = 0
```
poire-z, line 451: `self.current_flow` might be a more explicite name?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1869,20 +1887,38 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
     end
 end
 
+function ReaderFooter:onChangeViewMode()
+    self:onPageUpdate()
```
poire-z, line 1891: May be a comment about why you need to call it.
(I have the feeling it would be called anyway because of other events, but better explicite than implicite if it being called multiple times is not an issue.)

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -55,7 +55,8 @@ local ReaderRolling = InputContainer:new{
     -- With visible_pages=2, in 2-pages mode, ensure the first
     -- page is always odd or even (odd is logical to avoid a
     -- same page when turning first 2-pages set of document)
-    odd_or_even_first_page = 1 -- 1 = odd, 2 = even, nil or others = free
+    odd_or_even_first_page = 1, -- 1 = odd, 2 = even, nil or others = free
+    hide_nonlinear = nil,
```
poire-z, line 59: `hide_nonlinear_flows` may be, to make it clearer it's related to "flows" elsewhere.

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -526,8 +558,13 @@ end
 function ReaderRolling:onGotoNextChapter()
     local visible_page_count = self.ui.document:getVisiblePageCount()
     local pageno = self.current_page + (visible_page_count > 1 and 1 or 0)
-    local new_page = self.ui.toc:getNextChapter(pageno, 0)
-    if new_page then
+    -- Find next chapter start
+    local new_page = self.ui.document:getNextPage(pageno)
+    while new_page > 0 do
+        if self.ui.toc:isChapterStart(new_page) then break end
+        new_page = self.ui.document:getNextPage(new_page)
```
poire-z, line 565: (Just mentionning I have no issue with that :)

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1025,6 +1076,9 @@ function ReaderRolling:handleEngineCallback(ev, ...)
         -- file or re-rendering)
         self:showEngineProgress(1 - args[1]/100) -- unfill progress
     elseif ev == "OnDocumentReady" or ev == "OnSaveCacheFileEnd" then
+        if ev == "OnDocumentReady" then
+            self.ui.document:cacheFlows()
```
poire-z, line 1080: This has nothing to do in this crengine callback handler (mostly used for loading/rendering progress).
It should better be done in some :onReaderReady() handler I think.

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1310,6 +1364,14 @@ Note that %1 (out of %2) xpaths from your bookmarks and highlights have been nor
     })
 end
 
+function ReaderRolling:onToggleHideNonlinear()
+    self.hide_nonlinear = not self.hide_nonlinear
+    self.ui.document:setHideNonlinear(self.hide_nonlinear)
+    self.ui.document:render()
```
poire-z, line 1370: Nowhere we need to call `self.ui.document:render()` directly (the fact that you included the setting in the global doc properties in crengine will have it notice a re-rendering is needed next time... something requests something :)
Usually, we have ReaderRolling:onUpdatePos() called instead.

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -168,25 +168,35 @@ function ReaderToc:validateAndFixToc()
...
+                        nb_next_main = nb_next_main + 1
+                    end
                 end
             end
-            logger.dbg("BOGUS TOC:", i, page, ">", i-1, cur_page, "-", nb_prev, nb_next)
-            if nb_prev <= nb_next then -- less changes when fixing previous pages
+            logger.dbg("BOGUS TOC:", i, page, "<", i-1, cur_page, "-", nb_prev, nb_next)
+            -- Note: by comparing only the entries that belong to the main (linear) flow
+            -- we give priority to moving non-linear bogus entries
+            if nb_prev_main <= nb_next_main then -- less changes when fixing previous pages
```
poire-z, line 199: No issue when only TOC items in non-linear flows are misordered ? Our TOC code expects ALL TOC entries to be ordered, so if this is screwed by multiple entries in non-linear parts, skipped there, it will be an issue.

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -489,8 +510,31 @@ function ReaderToc:onShowToc()
         for _,v in ipairs(self.toc) do
             v.text = self.toc_indent:rep(v.depth-1)..self:cleanUpTocTitle(v.title)
             v.mandatory = v.page
+            local flow = self.ui.document:getPageFlow(v.page)
+            local orig_flow = flow
             if v.orig_page then -- bogus page fixed: show original page number
-                v.mandatory = T("(%1) %2", v.orig_page, v.page)
+                -- This is an ugly piece of code, which can result in an ugly TOC,
+                -- but it shouldn't be needed very often
```
poire-z, line 517: Good to say it's ugly, but please also say why it's rare and what this intends to do/show :)
I'd also like this wrapped in a `if  ...:hasHiddenFlows() then`.

frontend/document/credocument.lua
```diff
@@ -58,6 +58,10 @@ local CreDocument = Document:new{
     default_css = "./data/cr3.css",
     provider = "crengine",
     provider_name = "Cool Reader Engine",
+
+    hide_nonlinear = false,
```
poire-z, line 62: `hide_nonlinear_flows` here too, to make it clearer it's related to "flows" elsewhere.

frontend/document/credocument.lua
```diff
@@ -283,10 +287,127 @@ function CreDocument:updateColorRendering()
...
+
+function CreDocument:getPageNumberInFlow(page)
+    if self:hasFlows() then
+        return self.page_in_flow[page]
+    else
+        return page
+    end
+end
+
+function CreDocument:hasFlows()
```
poire-z, line 364: I think just seeing this one named `hasHiddenFlows()` and checking for `self.hide_nonlinear_flows` - and abused a bit more as wrappers - would make me quite happy with this PR.
(And moved above before all the other` :*Flow*` functions.)

frontend/document/credocument.lua
```diff
@@ -283,10 +287,127 @@ function CreDocument:updateColorRendering()
...
+    return self.flows[1] ~= nil
+end
+
+function CreDocument:cacheFlows()
+    -- Build the cache tables "flows" and "page_in_flow", if there are
```
poire-z, line 369: I'd rather see this one called and do its work only when `hide_nonlinear_flows == true`

frontend/ui/widget/progresswidget.lua
```diff
@@ -90,8 +92,8 @@ function ProgressWidget:paintTo(bb, x, y)
         local bar_width = (my_size.w-2*self.margin_h)
         local y_pos = y + self.margin_v + self.bordersize
         local bar_height = my_size.h-2*(self.margin_v+self.bordersize)
-        for i=1, #self.ticks do
-            local tick_x = bar_width*(self.ticks[i]/self.last)
+        for i,tick in pairs(self.ticks) do
```
poire-z, line 95: Try to not switch from array iterator to `pairs()`. We had bad experiences with `pairs()` in these kind of situations, and switch a lot to `ipairs` or `for i=1, #something`.

frontend/ui/elements/reader_menu_order.lua
```diff
@@ -26,6 +26,7 @@ local order = {
     typeset = {
         "set_render_style",
         "style_tweaks",
+        "hide_nonlinear",
```
poire-z, line 29: `hide_nonlinear_flows` too to make that more explicite.

poire-z (commented):
OK, better second impression after another hour reading :)
With better names and a bit more wrapping, it could be fine.

frontend/apps/reader/modules/readerbookmark.lua
WaseemAlkurdi: @Frenzie We would flip it normally for RTL, like a page number:
```
[2 / 14] 3
```
(page 2 out of 14 in non-linear flow number 3)
would become:
```
3 [14 / 2]
```

frontend/apps/reader/modules/readerfooter.lua
Jellby: > Can footer.pageno be different from footer.ui:getCurrentPage() ?

Yes, `footer.pageno` is the page number shown in the footer, the "local" page in the current flow, while `footer.ui:getCurrentPage()` is the global page number, counting all flows.

frontend/apps/reader/modules/readerrolling.lua
Jellby: I'll have to test again, but I believe this an the footer's `onChangeViewMode()` were needed to ensure proper updating when changing switching between paged and scroll mode (which can change forced pagebreaks between flows and the total page count). Maybe they're not (or can be moved) after the latest changes.

frontend/apps/reader/modules/readerbookmark.lua
Frenzie: That's the self-evident part. But you'd use those symbols? :-)

frontend/apps/reader/modules/readerrolling.lua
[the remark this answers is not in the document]
Jellby: That leads to the linebreaks being retained in the in the string, which is what I'm trying to avoid:
Screenshot_20201105_174502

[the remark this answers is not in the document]
poire-z: Just make it a single long line in the code - we have already quite a few.

frontend/apps/reader/modules/readerbookmark.lua
WaseemAlkurdi: @Frenzie Ah. The circular brackets would be better, as you stated. Square brackets have no place in user-facing stuff (imagine them being used in a literary text, a novel of some sort?)

frontend/apps/reader/modules/readerrolling.lua
Jellby: Actually, the purpose of this call is to regenerate the cache every time the document is re-rendered (due to any change, like fonts, tweaks, etc.). If I put it in `:onReaderReady`, it's only called once at the beginning. I tried setting a callback in `CreDocument` itself, but I didn't succeed (it was never called).

frontend/apps/reader/modules/readertoc.lua
Jellby: Nothing has changed here, except that:
* The displayed page in the TOC is the local page.
* The decision on whether to fix the bogus pages "before" or "after" is done based on the number of linear entries that need to be fixed, but all linear and non-linear are fixed. The idea is that we prefer fixing non-linear entries than linear ones.

At least that is the intent.

frontend/apps/reader/modules/readerrolling.lua
poire-z: I understand, so it's the same as for the TOC, we rebuild when we notice a rendering change, via:
ReaderRolling:updatePos() > self.ui:handleEvent(Event:new("UpdateToc"))
Except we don't do it onReaderReady, but only the first time it is accessed/needed.

So, you might either add an onUpdateToc() too (and share that event) - or send another one where this one is send. Or may be no need for another event as you're already in ReaderRolling.
So, onReaderReady + that one in ReaderRolling:updatePos().

(Ideally, no Reader/UI callback/stuff in CreDocument - which is a document class that could/should theoretically be usable in other contexts than our Reader.)

Jellby: I'll do this. I understand not wanting this callback mixed with something that has a different purpose, but I still feel that this is something that should be done in `CreDocument`. Having crengine send a signal to `ReaderRolling` and then `ReaderRolling` send it "back" to `CreDocument` looks like an unnecessary roundabout. Isn't there an easy way for `CreDocument` to react to crengine events (specifically a "done rendering" event)?

And another thing. Is there any guaranteed order in which `onUpdateToc` methods of different classes/instances are processed? If I want to make sure that the footer is updated *after* `ReaderToc:onUpdateToc`, can I just add a `ReaderFooter:onUpdateToc` method or should I create a different event in `ReaderToc:onUpdateToc` and make `ReaderFooter` react to that?

poire-z:
Following up on https://github.com/koreader/koreader/pull/6847#discussion_r518893208 above here (as I don't like this discussion hidden in the hidden comments above).

> Having crengine send a signal to ReaderRolling and then ReaderRolling send it "back" to CreDocument looks like an unnecessary roundabout

That's mostly how KOReader works everywhere. Roundabouts is its middle name (think about all the events). So, it's not at all an issue for me.

> Isn't there an easy way for CreDocument to react to crengine events

Well, up until you introduced a cache for page flows in it, it was just a thin wrapper, a pure proxy to crengine, with not much state inside of it. And I think until some bigger refactoring is needed, it should stay that way.
ReaderRolling is the "active" higher level wrapper to CreDocument.
But I'm still fine with you having that cache in CreDocument.

There is one little issue with handleEngineCallback: it's done in some kind of out of the main Lua state, and it won't crash the app if you have some coding error in it - so, we don't want to do too much stuff thru it:
https://github.com/koreader/koreader/blob/5ed240077ee505be70adbfaf27e5b1c00ac3ddbc/frontend/apps/reader/modules/readerrolling.lua#L236-L240

I see you do `self:cacheFlows()` inside `CreDocument:setViewMode()` and when toggling hide_nonlinear - which are two things that do not do any re-rendering (so, they don't change the document height) but needs flows to be updated.
I also realize the thing in `updatePos()` to check for a document-height change to notice there was a re-rendering is not perfect: there may be ultra rare cases when changes of styles may move chapters around a bit (and so, pages of some xpointers) while not changing 1px in document height. Also, toggling hide_nonlinear will most often not change the document-height.
The `handleEngineCallback` was added recently - and before it, there was indeed no way to know that a real re-rendering did happen (so, the check for document-height as the best available solution).

Not for you to do in this PR (unless you think that could make your handling clearer) but a more generic solution to handling that could be:
Here in `ReaderRolling:handleEngineCallback()`:
`if ev == "OnDocumentReady" then self.document.re_rendered = true end`
and in `updatePos()` check for `self.document.re_rendered` (and reset it to false once checked) instead of comparing old and current document height.
"updatePos" is a bad name - but I still think everything should be done in this central place for stuff that should react to a re-rendering - including your call to `CreDocument:cacheFlows()`.

(I'd say that even if your cached page flows are internal to CreDocument, updating them should be done from ReaderRolling, in the upper equivalent that calls the lower `CreDocument:setViewMode()` and `CreDocument:setHideNonlinear()` - but I understand you'd like them internal, so I won't be rigid on that :)

> And another thing. Is there any guaranteed order in which onUpdateToc methods of different classes/instances are processed? If I want to make sure that the footer is updated after ReaderToc:onUpdateToc, can I just add a ReaderFooter:onUpdateToc method or should I create a different event in ReaderToc:onUpdateToc and make ReaderFooter react to that?

No real contract on that I think. But the implementation (`WidgetContainer:propagateEvent()`) uses `ipairs()` to iterate thru the sub widgets. So, the `Reader*` modules would be processeded in the order they are `ReaderUI:registerModule()`'d in ReaderUI.
Currently, I think the ReaderFooter (not a child of ReaderUI, but of its ReaderView child) would get it before ReaderToc (as `self:registerModule("view", ReaderView` happens before `self:registerModule("toc", ReaderToc`).
So, possibly safer and clearer to add another event send by `ReaderToc:onUpdateToc` : `TocUpdated` - that's the way we naturally round about :)

frontend/apps/reader/modules/readerrolling.lua
poire-z: Answered and continued at https://github.com/koreader/koreader/pull/6847#issuecomment-723243231

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -191,39 +191,49 @@ local footerTextGeneratorMap = {
...
+        local left = footer.ui.document:getTotalPagesLeft(footer.ui:getCurrentPage())
+        return footer:getDataFromStatistics(prefix .. " ", left)
     end,
     chapter_time_to_read = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].chapter_time_to_read
-        local left = footer.ui.toc:getChapterPagesLeft(footer.pageno)
+        local left = footer.ui.toc:getChapterPagesLeft(footer.ui:getCurrentPage())
         return footer:getDataFromStatistics(
             prefix .. " ", (left and left or footer.pages - footer.pageno))
```
Jellby, line 238: Not affected by this PR, but is this construction needed? Wouldn't `(left or footer.pages - footer.pageno)` be exactly the same?

poire-z: Looks to me that indeed, it's exactly the same (unless I miss something with Lua pecularities :)

Jellby:
> > * (issue with bookmarks which could have some the normal page, some the tweaked non-linear page number, depending on whether they were made in one mode or the other - but that's a more general issue with our messy Bookmark stuff)
> 
> Right, but that also happens now if I make some bookmarks with large font and others with small font.

Maybe a matter for a different PR, but would it make sense to set the default bookmark title not to the page number, but to some placeholder, like `%p` or whatever, and substitute that with the "current" page number when displaying them (as with TOC page numbers)?

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: If there are no ticks at all (i.e., no chapter breaks), `ReaderToc:getNextChapter` will return `nil`, and `ReaderToc:getChapterPagesLeft` will return it as-is.

poire-z:
> Maybe a matter for a different PR, but would it make sense to set the default bookmark title not to the page number

Yep, that would make sense. And for a different PR. #5794 has some discussion about better handling of all that bookmark mess. (Fell free to take over that task if you are courageous :)

frontend/apps/reader/modules/readerfooter.lua
Jellby: That explains the need for `left or ...`, but not for `left and left`? If `left` is `nil`, `left or "foo"` is `"foo"`, and `left and left or "foo"` is `"foo"` as well... at least as far as I can see.

NiLuJe: Oh, yeah, I totally skipped over the fact that there was an `or` after :D.

frontend/ui/widget/progresswidget.lua
```diff
@@ -90,8 +92,8 @@ function ProgressWidget:paintTo(bb, x, y)
...
+            local tick_x = bar_width*((tick-0.5)/self.last)
```
NiLuJe, line 96: I'm... not necessarily a fan of that ;).

And, at the very least, it'll probably require rounding to ensure it doesn't randomly decide to jump a pixel or not ;).

Jellby: Well, either there is a non-existing "page 0" represented in the progress bar, or the last page is missing. If I use the plain page numbers for the non-linear filling, I get an overflow:
Screenshot_20201107_153402

which I can solve by using `page-1`, but then the ticks are off:
Screenshot_20201107_153445

So with `tick-0.5` the ticks are "midpage", they're still off, but at least they're not at the boundary of any page:
Screenshot_20201107_153527

As for rounding, I don't think it's needed. `tick` should be an integer, and even though `tick_x` can jump one pixel, it will always be between the values corresponding to `tick` and `tick-1`. In the case when each page takes less than one pixel, I don't really care which one it is, there's not enough resolution anyway.

NiLuJe: I'd rather keep the current behavior and just fix the overflow, FWIW ;).

Jellby: To fix the overflow and keep the ticks at the boundary, everything should be moved `-1` page. Then the first page would have no filled progress bar, and the last page would not be 100% filled (which would make sense if the bar means "pages already read"), or the progress bar in divided in `pages-1` chunks, instead of `pages`.

But ok, I'll leave it as is. I don't really use the skim widget enough.

poire-z: > As for rounding, I don't think it's needed

Not rounding what might be passed to the drawing code has been the caused of positionning & fitting issues (even in upper containers far from what you're dealing with) in the past. So, please do if you can :)

I have no real opinion on #6831 because I don't use it that much to jump - I just expect it to be "clean", so my preference among the 3 above screenshots is the first (without the overflow of course :) The 2 others, no matter how correct they may be for the person who gave it some thoughts, will look like a bug for most of the other users :/

Jellby: The first is wronger, because the first page is non-linear, and it appears as the second, it's the same -1 offset. If the first page is non-linear, it should go from 0 to 1, and if the last 2 pages (out of 100) are non-linear, the grey should go from 98 to 100 (even though the pages are 99 and 100).

So, my point is the grey blocks in the 2nd and 3rd example are correct. The ticks in 1st and 2nd are correct, if they are interpreted as marking the end of a page where a chapter starts, and the filled bar is correct if it's interpreted as all pages up to (and including) the current one.

Jellby:
I think I've addressed most, if not all the comments so far with the new commit.

frontend/ui/widget/progresswidget.lua
NiLuJe: My stance on the whole #6381 discussion was that I might be agreeable to modify the way the *bar* is filled in, but that the ticks should very much stick to what they're currently doing.

NiLuJe: > To fix the overflow and keep the ticks at the boundary, everything should be moved `-1` page. Then the first page would have no filled progress bar, and the last page would not be 100% filled (which would make sense if the bar means "pages already read"), or the progress bar should be divided in `pages-1` chunks, instead of `pages`.

FWIW, I'd probably be amenable to either of those approaches, I think ;).

(Never seeing the bar fully filled on the last page kinda makes sense to me).

frontend/ui/widget/progresswidget.lua
```diff
@@ -44,7 +44,7 @@ local ProgressWidget = Widget:new{
     bordercolor = Blitbuffer.COLOR_BLACK,
     bgcolor = Blitbuffer.COLOR_WHITE,
     nlcolor = Blitbuffer.COLOR_LIGHT_GRAY,
-    rectcolor = Blitbuffer.COLOR_DIM_GRAY,
+    rectdim = 2/3,
```
NiLuJe, line 47: I'd rather you use something more clearly in the eInk palette ;) (That'd be `0x44`, I think).

NiLuJe: That's not a color, that's the dim factor, wheee. Never mind me :).

Jellby: But this is not a color, but an "alpha" factor.

frontend/ui/widget/progresswidget.lua
NiLuJe: That one is still outstanding ;).

NiLuJe: And since we were talking about rounding, I realize that `tick_x` was already never rounded hee, wheee... -_-".

(At a quick glance, `bar_width` should always be an int, but the MUL against `tick/last` might change that.

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -558,11 +561,15 @@ end
...
+        -- Find next chapter start
+        local new_page = self.ui.document:getNextPage(pageno)
+        while new_page > 0 do
+            if self.ui.toc:isChapterStart(new_page) then break end
+            new_page = self.ui.document:getNextPage(new_page)
+        end
+    else
+        local new_page = self.ui.toc:getNextChapter(pageno) or 0
     end
     if new_page > 0 then
```
poire-z: Looks like a scope issue there (I'm not even sure the local is limited to the `else` and new_pge is nil after the `end` - but it feels it would be the case).

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -721,7 +774,21 @@ function ReaderRolling:onGotoViewRel(diff)
...
+        local new_page = self.current_page
+        if self.ui.document:hasHiddenFlows() then
+            local test_page = 0
+            for i=1, math.abs(diff*page_count) do
+                if diff > 0 then
+                    test_page = self.ui.document:getNextPage(new_page)
+                else
+                    test_page = self.ui.document:getPrevPage(new_page)
+                end
+                if test_page > 0 then new_page = test_page end
```
poire-z, line 786: 3 lines please (might be a question of taste, but there I was looking for new_page getting updated and didn't see it at first).

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -783,12 +850,13 @@ function ReaderRolling:updatePos()
...
     if self.old_doc_height ~= new_height or self.old_page ~= new_page then
+        if self.hide_nonlinear_flows then
+            self.ui.document:cacheFlows()
+        end
         self:_gotoXPointer(self.xpointer)
         self.old_doc_height = new_height
         self.old_page = new_page
         self.ui:handleEvent(Event:new("UpdateToc"))
-        self.view.footer:setTocMarkers(true)
-        self.view.footer:onUpdateFooter()
```
poire-z, line 791: You removed these ones (and one below), because they are not needed, or triggered by something else?

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1310,6 +1377,13 @@ Note that %1 (out of %2) xpaths from your bookmarks and highlights have been nor
     })
 end
 
+function ReaderRolling:onToggleHideNonlinear()
+    self.hide_nonlinear_flows = not self.hide_nonlinear_flows
+    self.ui.document:setHideNonlinearFlows(self.hide_nonlinear_flows)
+    self:onUpdatePos()
+    self:onChangeViewMode()
```
poire-z, line 1384: A small comment about what you call each of these ones?

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -62,6 +62,7 @@ end
 
 function ReaderToc:onUpdateToc()
     self:resetToc()
+    self.ui:handleEvent(Event:new("TocUpdated"))
```
poire-z, line 65: resetToc() just clears the TOC data (so it's rebuild later when needed).
Don't you need it updated for when you catch TocUpdated (or so that that "TocUpdated" event is true to its name) ?
(But beware, rebuilding the TOC might be expensive, so that's why we delay it until needed.)

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -436,21 +447,48 @@ function ReaderToc:isChapterEnd(cur_pageno)
...
+        -- Count pages until new chapter
+        local pages_left = 0
+        local test_page = self.ui.document:getNextPage(pageno)
+        while test_page > 0 do
+            pages_left = pages_left + 1
+            if self:isChapterStart(test_page) then
+                return pages_left
+            end
+            test_page = self.ui.document:getNextPage(test_page)
+        end
```
poire-z, line 461: Is it ok to return nil if there's no chapterStart after current page?

frontend/document/credocument.lua
```diff
@@ -283,10 +287,132 @@ function CreDocument:updateColorRendering()
...
+    if self:hasHiddenFlows() then
+        if page < 0 then return 0 end
+        local flow = page > 0 and self:getPageFlow(page) or 0
+        for test_page=page+1, self:getPageCount() do
+            local test_page_flow = self:getPageFlow(test_page)
+            if test_page_flow == flow or test_page_flow == 0 then
+                return test_page
+            end
+        end
+        return 0
```
poire-z, line 326: No case for breaking that `for` loop before reaching getPageCount() ? (I guess you can't for flow 0, but just asking).

frontend/document/credocument.lua
```diff
@@ -283,10 +287,132 @@ function CreDocument:updateColorRendering()
     end
 end
 
+function CreDocument:setHideNonlinearFlows(hide_nonlinear_flows)
+    if hide_nonlinear_flows ~= self.hide_nonlinear_flows then
+        self.hide_nonlinear_flows = hide_nonlinear_flows
+        self._document:setIntProperty("crengine.doc.nonlinear.pagebreak.force", self.hide_nonlinear_flows and 1 or 0)
+        self:cacheFlows()
```
poire-z, line 294: This one is not wrapped by `if self.hide_nonlinear_flows then` (thanks for wrapping all the others :) 
Is that wrapping safe enough ? No self.flows to clear to be safer ?

poire-z (commented):
Looks fine (readable, and feeling safe with the hasHiddenFlows() wrappers :)
Mostly nitpicks:

frontend/ui/widget/progresswidget.lua
```diff
@@ -42,7 +43,8 @@ local ProgressWidget = Widget:new{
     bordersize = Screen:scaleBySize(1),
     bordercolor = Blitbuffer.COLOR_BLACK,
     bgcolor = Blitbuffer.COLOR_WHITE,
-    rectcolor = Blitbuffer.COLOR_DIM_GRAY,
+    nlcolor = Blitbuffer.COLOR_LIGHT_GRAY,
```
poire-z, line 46: I thought I had commented on that, but I can't find my comment anymore :/
So, re-mentionning it (but your call).
`nlcolor` (`nl` for non linear) and the use of `self.flows` (which should be documented with `flows = nil -- small comment` in the constructor) might have better more generic names in this generic "ProgressWidget".
"secondaries" or something else generic ?

frontend/apps/reader/modules/readergoto.lua
```diff
@@ -134,20 +140,42 @@ function ReaderGoto:gotoPage()
...
+        local flow
+        number, flow = string.match(page_number, "^ *%[(%d+)%](%d*) *$")
+        flow = tonumber(flow) or 0
+        number = tonumber(number)
+        if number then
+            if self.ui.document.flows[flow] ~= nil then
+                for i=self.ui.document:getFirstPageInFlow(flow), self.ui.document:getPageCount() do
+                    if self.ui.document:getPagenNumberInFlow(i) == number and
+                       self.ui.document:getPageFlow(i) == flow then
+                        self.ui:handleEvent(Event:new("GotoPage", i))
```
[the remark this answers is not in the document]
Jellby: Do you mean if I want to go to page 30 in a flow that has only 5 pages? Sure... also if I want to go to page 0 or -7.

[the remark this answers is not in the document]
poire-z: Yep, that's what I mean (but mostly to not just go scan 2900 pages in a 3000 pages books to realize an early flow won't be found in there :)

frontend/apps/reader/modules/readerrolling.lua
Jellby: They are triggered in `ReaderFooter:onTocUpdated()`. I thought it's better if the footer manages its own updates, rather than rely on the reader asking for explicit updates.

frontend/apps/reader/modules/readerrolling.lua
Jellby: A comment here or in the code? Basically, the document may change (due to the force pagebreaks), and the footer and toc markers need to be updated even if it doesn't change.

frontend/apps/reader/modules/readertoc.lua
Jellby: Hmm... Maybe I should just change the name, because as I see it, when the TOC is reset, the footer needs to rebuild it to get the markers. So yes, I need the updated TOC, but I guess this need triggers the rebuild.

frontend/apps/reader/modules/readertoc.lua
Jellby: It should be, it's the current behaviour, and I have tested it, I believe.

frontend/apps/reader/modules/readerrolling.lua
poire-z: OK, agreed.

frontend/document/credocument.lua
Jellby: I don't think so. You have to keep testing until you find flow 0 or the end of the document. Unless you know in advance that a) you have reached the end of the current flow (this we can know as soon as a different flow is found), and b) there are no remaining pages in flow 0 (this is the tricky part).

frontend/apps/reader/modules/readerrolling.lua
poire-z: In the code, so it survives :) 
And a comment about why onChangeViewMode() (what it does that you need, because one would wonder what view mode is changed here).

frontend/document/credocument.lua
Jellby: This `cacheFlows` will effectively clear `self.flows`, I'll put another test down there.

frontend/apps/reader/modules/readertoc.lua
poire-z: Also please have a quick check that on loading a book (with cache or without), the TOC is only built once.
(I remember #3292 and https://github.com/koreader/koreader/issues/3261#issuecomment-333560902 - which are quite old - dunno if these thoughts are still relevant :)

frontend/ui/widget/progresswidget.lua
```diff
@@ -117,7 +118,7 @@ function ProgressWidget:paintTo(bb, x, y)
         local bar_width = (my_size.w-2*self.margin_h)
         local y_pos = y + self.margin_v + self.bordersize
         local bar_height = my_size.h-2*(self.margin_v+self.bordersize)
-        for i,tick in pairs(self.ticks) do
+        for i,tick in ipairs(self.ticks) do
```
poire-z, line 121: We like a space after a comma, like in English :)

frontend/apps/reader/modules/readertoc.lua
Jellby: Right, I think there was one TOC too many which should be fixed now.

frontend/document/credocument.lua
Jellby: But nothing that another cache item cannot fix.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1869,9 +1905,25 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
     end
 end
 
+function ReaderFooter:onTocReset()
+    self:setTocMarkers(true)
+    self:onUpdateFooter()
```
poire-z, line 1910: I'm not sure how soon after a toc reset the TOC was filled again before this PR (might be that it was delayed after a repaint, dunno). But with this, is will be immediately filled.
(I can't answer that question after a quick look at the code, so, just asking in case someone get an idea :)

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -457,6 +468,31 @@ You can set how many lines are shown.]])
         help_text = _([[When page overlap is enabled, some lines from the previous pages are shown on the next page.]]),
         sub_item_table = page_overlap_menu,
     }
+    if self.view.document:hasNonLinearFlows() then
```
poire-z, line 471: I know we use `self.view.document` in a few modules - but not in this one, readerrolling.
The canonical way to access the document object is usually with `self.ui.document` (`ui` instead of `view`).

frontend/document/credocument.lua
```diff
@@ -283,10 +288,150 @@ function CreDocument:updateColorRendering()
...
+        if page < 0 then return 0 end
+        local last_linear_page = self:getLastLinearPage()
+        local flow = page > 0 and self:getPageFlow(page) or 0
+        for test_page=page+1, self:getPageCount() do
+            local test_page_flow = self:getPageFlow(test_page)
+            if test_page_flow == flow or test_page_flow == 0 then
+                return test_page
+            end
+            if test_page_flow ~= flow and test_page > last_linear_page then
+                break
```
poire-z, line 328: Can't you have/use a getLastPageInFlow(flow) when flow > 0 (like you have in the next function) ?
As when a flow is done, you can't meet it again - so it's no use checking up to the last linear page. No?

Jellby: You can't meet it again, but you will meet flow 0 eventually (unless `test_page > last_linear_page`), and that will be the "next page". I could limit the loop to `math.max(self.getLastPageInFlow(flow), self.getLastLinearPage())`.

poire-z: Or I misunderstood this function :) 
`-- Get the next/prev page number, skipping non-linear flows,`
I thought you could use it to navigate pages inside a non-linear flow - and that if you're in flow No7 out of 19, you don't need to test pages until the end of the document.

Jellby: You skip non-linear flows, but linear pages may be interspersed between them, and we want to stop there.

poire-z: But you really don't use CreDocument:getNextPage(p) to page forward inside non-linear flows (and expecting to get the next page inside this same non-linear flow)?
OK, I guess you do (so, this comment is lying :) and you'll stop looping when you meet back a page from flow 0.

frontend/apps/reader/modules/readerfooter.lua
Jellby: Note that `ReaderRolling:updatePos()` had before:
```
        self.ui:handleEvent(Event:new("UpdateToc"))
        self.view.footer:setTocMarkers(true)
        self.view.footer:onUpdateFooter()
```
so nothing should have changed in that particular chain of events.

frontend/document/credocument.lua
Jellby: There could be another performance improvement: when a different non-linear flow is found, instead of trying pages one by one, I could in principle skip all *n* pages in the flow in one go. But I somehow liked the simplicity of just paging through, it makes fewer assumptions about how the flows behave (for instance, non-linear flows should not be interleaved, but the one-by-one method would work even if that was the case).

frontend/document/credocument.lua
```diff
@@ -283,10 +288,173 @@ function CreDocument:updateColorRendering()
...
+-- i.e. the next/prev page that is either in the current
+-- flow or in the linear flow (flow 0)
+-- If "page" is 0, these give the initial and final linear pages
+function CreDocument:getNextPage(page)
+    if self:hasHiddenFlows() then
+        if page < 0 or page >= self:getPageCount() then
+            return 0
+        elseif page == 0 then
+            return self.getLastLinearPage()
+        end
```
poire-z, line 323: In onGoToBegenning, you provide page=0 to go to the first page - which sounds like it's not what you do :)

Jellby: Of course, I got them mixed.

Jellby:
Maybe I should use #6859 instead of forcing `no_wrap` on selected methods?

poire-z:
Isn't a reset already forced when you call `CreDocument:setHideNonlinearFlows()` (because of set*)? (#6854 added a setMethod, #6859 removed it - but you need one :) 
(Or did you have something else in mind?)

But may be you could have some of them not be `no_wrap` (no cache) and neither `cache_global` (long cache), and have some of them `cache_by_tag` (cache resut by page, with a 10 MRU stack).

Jellby:
I had https://github.com/koreader/koreader/issues/6809#issuecomment-720092992 in mind, but I'll try reproducing/investigating it again, and I'll come back if I can figure it out.

poire-z:
Had a try on my Kobo, looks ok (as far as the things I care about as concerned, not tested bookmarks, some footer items...)
(A small initial feeling that turning pages was slower on books without non-linear stuff - but I can't see any reason in the code, so may be it's a suggestive impression because I was looking for slowness :) 

So, I reported previously that 90% of my books don't have any non-linear stuff.
It turns out that among the 10% that do, 90% of them just have the first page with the cover marked as non-linear :/ 
I managed to find only 2 books (+ your 1001 nights one) that have some non-linear stuff all along the book, mostly footnotes sections. It works fine: you can't reach them with tap to turn page, but you can with the Skim widget +1 / -1. I was looking for these sections, but they were hard to find, so the hiding works well :) (but I'll personnally prefer not hiding anything and skipping them manually).

I think the footer marking is good (`43//67` is nice to indicate you're in this mode, `[1/4]3`) is ok. although the fragment number is a bit useless).

Jellby:
> I think the footer marking is good (`43//67` is nice to indicate you're in this mode, `[1/4]3`) is ok. although the fragment number is a bit useless).

It could be useful if you want to identify this page somehow, for instance to access it via goto, particularly if there are many fragments.

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -436,21 +447,48 @@ function ReaderToc:isChapterEnd(cur_pageno)
...
-    return next_chapter
 end
 
 function ReaderToc:getChapterPagesDone(pageno)
-    if self:isChapterStart(pageno) then return 0 end
-    local previous_chapter = self:getPreviousChapter(pageno)
-    if previous_chapter then
-        previous_chapter = pageno - previous_chapter
+    if self.ui.document:hasHiddenFlows() then
+        if self:isChapterStart(pageno) then return 0 end
```
poire-z, line 473: You forgot to make the non-hasHiddenFlows branch benefit from `if self:isChapterStart(pageno) then return 0 end` - so I guess you can probably move it back at the start of the function.

getChapterPagesDone seems not used anywhere -  but I use it in my own patch to tweak the output of the `pages_left` item - so, I got wrong info.
That's my tweak:
```diff
     pages_left = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].pages_left
         local left = footer.ui.toc:getChapterPagesLeft(footer.pageno)
-         return prefix .. " " .. (left and left or footer.pages - footer.pageno)
+        -- so we see better where we are in that chapter
+        local done = footer.ui.toc:getChapterPagesDone(footer.pageno)
+        left = left or (footer.pages - footer.pageno)
+        done = done or footer.pageno-1
+        return (left+1).." > "..(left+done+1)
```

Jellby:
> Isn't a reset already forced when you call `CreDocument:setHideNonlinearFlows()` (because of set*)?

The problem was, I believe, that `CreDocument:setHideNonlinearFlows()` forced a `cacheFlows()` call, which gets the total page count before the document is actually re-rendered, so it's wrong... but the rendering doesn't clear the cache and the wrong number of pages stays. I've moved the call to `cacheFlows()` elsewhere, after the re-rendering happens, I think that fixes the issues I was seeing.

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1388,6 +1388,13 @@ function ReaderRolling:onToggleHideNonlinear()
     -- The document may change due to forced pagebreaks between flows being
     -- added or removed, so we need to find our location
     self:onUpdatePos()
+    -- When disabling hide_nonlinear_flows, we must ensure that the
+    -- flow and call caches are cleared, to get the right page numbers,
+    -- which may have changed (when enabling it, this is already done
+    -- automatically by updatePos)
+    if not self.hide_nonlinear then
```
poire-z, line 1395: Typo ? Shouldn't it be `self.hide_nonlinear_flows`?

poire-z:
Might still want to think whether the new functions needs to be cached or not.
You might want to check how many milliseconds they take whether cached or not - and only cache them if the not cached value is really larger.
https://github.com/koreader/koreader/blob/48a38504f78ed134308b19a4f5ab88f4ec9dbac0/frontend/document/credocument.lua#L1064-L1067

Currently not cached (but they feel cheap and not needing any cache):
```
CreDocument:hasNonLinearFlows()
CreDocument:hasHiddenFlows()
```
Currently cached globally (not trashed until next re-rendering):
```
CreDocument:getNextPage(page)
CreDocument:getPrevPage(page)
CreDocument:getPageFlow(page)
CreDocument:getLastLinearPage()
CreDocument:getFirstPageInFlow(flow)
CreDocument:getTotalPagesInFlow(flow)
CreDocument:getPageNumberInFlow(page)
CreDocument:getTotalPagesLeft(pageno)
```
These are usually cheap when not using flows - dunno how less cheap they get when using flows.
But caching all those that take a `page` argument will mean a 3000 pages books could get 4 or 5 * 3000 entries in the global cache - which could then make other cached entries get a bit slower (as looking for them in a 10000 entries cache is slower that looking for them in a 30 entries cache).
So, may be the ones with a `page` argument could be `no_wrap` (to the expense of possibly be a bit slower when hiding non-linear flows).
Those with `flow` could be cached, as there are usually not much flows (from 1 to a few dozens entries at max).

Jellby:
Hmm... the toc markers in the footer are "wrong" in scroll mode, even before this PR, are they not?

I mean that they are set by pages, but the progress bar is filled by position. So, if a document has 100 pages and there is a chapter that starts at page 50, there will be a mark at 50% of the bar, and that doesn't change in paged or scroll mode. But if the last 50 pages are very short (e.g., they are short footnotes, with forced pagebreaks), in total height they may account for maybe only 20% of the document, so in scroll mode one would only reach this "chapter" when the progress bar is 80% full, but still the marker would be at 50%.

poire-z:
That's possible. I guess nobody really cares about scroll mode :/ (including me, and I wouldn't really care about fixing that).

Jellby:
OK, I'll ignore it. Fixing that may require changing this in credocument:
```
int LVDocView::getPageStartY(int pageIndex)
{
        if (isPageMode() && pageIndex >= 0 && pageIndex < m_pages.length())
                return m_pages[pageIndex]->start;
        return -1;
}
```
since it's always returning `-1` in scroll mode (or keeping a list of both pages and positions in the toc).

> Note that in page mode, one can be switched to scroll mode when doing text selection across page - so we might not want some TOC reset&recompute to happen and slow that process.

That's interesting, this means trying to select something across a non-linear skip will have some fun effects...

frontend/apps/reader/modules/readerrolling.lua
Jellby: At least readerfooter uses both `self.ui.document` and `self.view.document`. Any reason for this or is it OK if I change them all to `self.ui.document` too?

Jellby:
I couldn't ignore it :D I realized the "xpointer" is already saved in the toc, so I thought this could work:

1. Add an optional argument to `ReaderToc:getTocTicksFlattened`, such that:
```lua
    if return_pos then
        local ticks_pos = {}
        for _, tick in pairs(ticks_flattened) do
            local idx = self:getTocIndexByPage(tick)
            local pos = self.ui.document:getPosFromXPointer(self.toc[idx].xpointer)
            table.insert(ticks_pos, pos)
        end 
        return ticks_pos
    end 
```
that is, return positions instead of page numbers

2. In readerfooter, call this with `return_pos=true` if in scroll mode, and as usual otherwise.

No extra calls to rebuild the toc are needed. Would this be OK?

It seems to work fine for me. The skim widget still shows the ticks by pages, but that's also fine because its progress bar is filled by pages too. Compare the two sets of ticks:
Screenshot_20201112_172232

poire-z:
`getPosFromXPointer()` can be a bit expensive, so doing it for 1000 entries might shows some slow down - possibly only the first time as it would be cached by the cre call cache.
So, you'd have to try/time it with the cre call cache stats on a book with a large TOC.

Otherwise, no problem for me to do a bit more work in scroll mode if you think it makes better sense - but having the footer progress bar change that way could feel strange, although on most books we won't notice any difference.

poire-z:
Looks allright, no comment :) 

We still have a few days before we can merge this and its base counterpart - after 2020.11 is released, + a few more days just in case a hot fix is needed (@Frenzie: still planned for this weekend?).

(Tell me if you want https://github.com/koreader/crengine/pull/392 merged earlier than that.)

Frenzie:
@poire-z I'll prepare it this weekend, probably a couple more days wrt https://github.com/koreader/koreader/pull/6870 though.

poire-z:
Mentionning that just in case it shows a bug or edge case (not investigating it, I don't mind the bogus display):
With latest cre and this PR, and my own tweak to the footer page_left item:
```diff
     pages_left = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].pages_left
         local left = footer.ui.toc:getChapterPagesLeft(footer.pageno)
-        return prefix .. " " .. (left and left or footer.pages - footer.pageno)
+        -- so we see better where we are in that chapter
+        local done = footer.ui.toc:getChapterPagesDone(footer.pageno)
+        left = left or (footer.pages - footer.pageno)
+        done = done or footer.pageno-1
+        return (left+1).." > "..(left+done+1)
     end,
```
When browsing (with the Skim widget <kbd>+1</kbd>) from a 7-pages chapter to a 1-page non-linear footnote pages to a 3 pages chapter, I get this displayed:
```
1/7   (last page of chapter)
1/8   (first and last page of non-linear flow) - the 8 is/feels bogus although probably true :)
3/3   (first page of next chapter)
```
I guess it's fine with the current master code, but may be there is a bug in `getChapterPagesDone()` (still not used anywhere :) - or is that a tricky edge case, or the expected display when in non-linear flows ? (I haven't followed how you deal with these kind of boundaries :)

Jellby:
That is OK, I think. Is that "pages left in chapter" / "total pages in chapter"? Then the thing is when you are in the linear flow, it only counts the linear pages, so you get the `/7`, but when you jump into the non-linear flow, it counts that in the current chapter, and you get `/8` (7+1 because there's one page in that particular non-linear flow). The idea is: if you use page back/forward (not skim -1/+1), how many pages until the previous/next chapter?

poire-z:
(May be for later, and for consistency, this probably should also be made available with FB2 books that have their footnotes in 2 optional `body[name="notes"]` and `body[name="comments"]` usually at the end of the single-fragment document. @hius07: thoughts?)

hius07:
> this probably should also be made available with FB2 books

Do you mean to hide Notes and/or Comments at the end of the book?
I don't see the use of this option.

poire-z:
Yes, as an optionnal/toggable feature. The main benefit I guess for FB2 (where you know you already have reached the end of the book content when you see the footnotes pages) would be that you could get the "real" number of pages (content, excluding footnotes) and a possibly more representative progress bar in the footer.

Jellby:
They would only be "hidden" from normal page forward. They'd still be visible as in-page footnotes (if enabled), and fully accessible through links from the text. The idea is that when you're reading the book, you've probably already seen the notes (at least if you're interested in them), so there's no need to see them again when you reach the end of the main text.

I don't know what are specifications or conventions of FB2. With EPUB, the spec says that non-linear fragments *must* be accessible through other means (links from the text or toc). If that is not the case with FB2, it could happen that some pages are only visible by using skim/goto, with no other indication whatsoever that they exist at all.

poire-z:
Well, the FB2 specs state this (really like if they were you own words :) https://github.com/koreader/crengine/blob/master/fb2props/FictionBook2.2.xsd#L168

> `<xs:element name="body" maxOccurs="unbounded">`
> `<xs:annotation>`
> `<xs:documentation>`Main content of the book, multiple bodies are used for additional information, like footnotes, that do not appear in the main book flow. The first body is presented to the reader by default, and content in the other bodies should be accessible by hyperlinks. Name attribute should describe the meaning of this body, this is optional for the main body.`</xs:documentation>
`

Jellby:
Then one shouldn't look for `body[name="notes"]`, but just for any `body` after the first (`body[position()>1]`).

hius07:
OK, that makes sense.
I personally am happy since @poire-z added the notes first page number to the TOC, so I can see the length of the main content.

poire-z (approved):

poire-z:
Still a few luacheck warnings to solve.

poire-z:
Now some issues with unit tests that might need some update to the new functions.
```
[ RUN      ] spec/front/unit/readerrolling_spec.lua @ 51: Readerrolling module test in portrait screen mode should goto previous chapter
frontend/apps/reader/modules/readerrolling.lua:594: attempt to call method 'getPrevChapter' (a nil value)

stack traceback:
	frontend/apps/reader/modules/readerrolling.lua:594: in function 'onGotoPrevChapter'
	spec/front/unit/readerrolling_spec.lua:55: in function <spec/front/unit/readerrolling_spec.lua:51>

[  ERROR   ] spec/front/unit/readerrolling_spec.lua @ 51: Readerrolling module test in portrait screen mode should goto previous chapter (0.35 ms)

[ RUN      ] spec/front/unit/readerrolling_spec.lua @ 139: Readerrolling module test in landscape screen mode should goto previous chapter
frontend/apps/reader/modules/readerrolling.lua:594: attempt to call method 'getPrevChapter' (a nil value)

stack traceback:
	frontend/apps/reader/modules/readerrolling.lua:594: in function 'onGotoPrevChapter'
	spec/front/unit/readerrolling_spec.lua:143: in function <spec/front/unit/readerrolling_spec.lua:139>

[  ERROR   ] spec/front/unit/readerrolling_spec.lua @ 139: Readerrolling module test in landscape screen mode should goto previous chapter (0.38 ms)
```

poire-z:
@Jellby : hope you haven't forgotten everything about this PR :)

I'm reusing much of it to implement "handmade hidden flows" (https://github.com/koreader/koreader/issues/10193#issuecomment-1685412477), that would allow one to set hidden flows on PDF documents (and also on CRE documents whether they have non-linear fragments or not).
I'm just implementing/plugging/overridding the no-op functions you added to Document, and let all the toc/footer/bookmap/... works just as you made them work with non-linear-fragments hidden flows.

I have a few questions regarding your code:
https://github.com/koreader/koreader/blob/38d468f36cf973838381600c393f6eaf208049f4/frontend/document/document.lua#L235-L244

(I have to make ReaderPaging use these instead of `current_page = current_page+1` / `-1` - for the many ways they are used, paging, scrolling, zoom mode, reflow.... and it's not really fun :/)

You allow these methods to return `0`. I assume it means for the calling code that it has reaches the end or start of the document, and that it's for that calling code to handle that case. Right?

Also, when called with `page=0`, they return the page number of the other side, so, `getPrevPage(0)` returns the last page (of the main flow) - looks like it was not implemented for `Document:getNextPage(0)`, but it is for `CreDocument:getNextPage(0)` which returns `1` or the start of the main flow - so I guess I'd have to fix this, if it is really used?)
Fine with this, just curious about why you needed that behaviour.

Also, in the snippet above, it looks like there's a logical bug (`<` vs `<=`) - shouldn't it be changed to this?:
```diff
- return (new_page > 0 and new_page < self.info.number_of_pages) and new_page or 0 
+ return (new_page > 0 and new_page <= self.info.number_of_pages) and new_page or 0 
```

Also mentionning a little bug (that is not part of my scope :) that I noticed, and that I can reproduce on an EPUB with (sick) non-linear fragments:
image

With chapters spanning fragment boundaries like above, the footer "page left in chapter" would show in the above case at the position of the arrow: `13//21` (so, including the pages in yellow of the chapter continuing when the main flow is reached again - but not the same chapter that was active when the main flow stops) instead of `1//21` (if stopping when the main flow stops, or `3//21` (if accounting the pages in that chapter even if part of a hidden flow).

Jellby:
> You allow these methods to return `0`. I assume it means for the calling code that it has reaches the end or start of the document, and that it's for that calling code to handle that case. Right?

I guess so, or just to return a reproducible and recognizable value whenever something wrong happens.

> Also, when called with `page=0`, they return the page number of the other side, so, `getPrevPage(0)` returns the last page (of the main flow) - looks like it was not implemented for `Document:getNextPage(0)`, but it is for `CreDocument:getNextPage(0)` which returns `1` or the start of the main flow - so I guess I'd have to fix this, if it is really used?) Fine with this, just curious about why you needed that behaviour.

I think that was simply to have kind of "overloaded" functions that still do something useful with `0`. That way I could just say `getNextPage(0)` to find the first (linear) page of the document, which makes sense (as it would by default be page 1), and then `getPrevPage(0)` to find the last one is the obvious partner.. These are already used in a couple of places.

`Document:getNextPage(0)` already returns `1` without a special treatment, or am I missing something?

> Also, in the snippet above, it looks like there's a logical bug (`<` vs `<=`) - shouldn't it be changed to this?:
> 
> ```diff
> - return (new_page > 0 and new_page < self.info.number_of_pages) and new_page or 0 
> + return (new_page > 0 and new_page <= self.info.number_of_pages) and new_page or 0 
> ```

You're probably right.

> With chapters spanning fragment boundaries like above, the footer "page left in chapter" would show in the above case at the position of the arrow: `13//21` (so, including the pages in yellow of the chapter continuing when the main flow is reached again - but not the same chapter that was active when the main flow stops) instead of `1//21` (if stopping when the main flow stops, or `3//21` (if accounting the pages in that chapter even if part of a hidden flow).

I believe I consciously assumed that fragments and chapters would be "properly nested", either a fragment contains full chapters or is fully contained in a chapter. I knew that's not necessarily the case, but it certainly matches my use cases, and figured it would also be the most common situation. Given that there is formally no concept of "chapter" in epub 2 (only TOC and hyperlink targets, which may signal the "beginning", but no the end of a chapter), I wouldn't consider the current behaviour wrong, just... "creative" :)

PR #6848, opened 2020-11-04
Analog clock plugin

jperon:
This plugin shows a fullscreen analog clock.

clock

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6848)


poire-z:
What this has to do with KO**reader** ? :)
(I had my doubts about QRcode, but at least, I can undestand one can export book highlights with it - but I think KOreader is a reading app, and it doesn't need to have all the useless widgets/apps a phone has - https://github.com/koreader/koreader/pull/5854#issuecomment-586579954.) (ok, one might find it useful when doing something else - but should it have a countdown clock too ? as I may be cooking eggs some day and have nothing but my Kobo at hands :)

Galunid:
I kind of feel this one is a bit of a stretch, in the end it's KO**Reader**, I'd rather not include this.

pazos:
I do agree with @poire-z and @Galunid (wow, you're fast). The plugin looks cool and I would like to use it but it's out of the scope of the program. 

We're lucky since it is a plugin, you can upload it to your own account and label it accordly (like https://github.com/search?q=koreader+plugin) and people who want to use it can add the code to their own installation.

jperon:
I was nearly sure it would be the reply… :)

Thanks @pazos for pointing me to #6768, as I was thinking about something similar.

Frenzie (commented):
I do like it and I may well put it on my device, but indeed even if it's just a few strings it's not something we should bother the translators with. ^_^

On the plugin itself, perhaps you'd want to generate some fake input events or some other somewhat hackish workaround to keep autostandby from kicking in for a while longer than usual? (Or extend the plugin with some hooks for that…) That might be overkill though.

jperon:
@Frenzie Thanks for your comment. Would 486a0ea6002364940ba226e11525128a717cd32c be enough for autosuspend ? I don't think it would be useful to add a setting to disable it: if one wants a clock in front of him (I'd use it for example when giving a course), he would expect it to be on.

Frenzie:
You'd still want it to autosuspend eventually when not plugged in I'd say, just after an hour or two instead of the regular 15-30 minutes. Although if you really want to get into it you could go standby all the time and update the clock once every e.g. 15 minutes, in which case you should be able to last for a week or maybe two on battery. See https://github.com/koreader/koreader/blob/91708e25d0c7ad340675e58b634ba0a46e2d40fa/frontend/device/wakeupmgr.lua

Just throwing some random ideas out there, mind you. The current concept of turning off the clock manually when you're done with it is perfectly functional and sensible.

jperon:
For now, I think I'll leave it as is, and keep my free time for #6768. To avoid spamming issues here, I'll close this PR: I have put the plugin [here](https://github.com/jperon/clock.koplugin).

Hzj-jie:
Are you interested to merge it into https://github.com/koreader/contrib?
You may integrate the plugin with a pending https://github.com/koreader/koreader/pull/7419 to make it visible in "More Tools" menu if users decided to install it themselves.

Hzj-jie:
Are you interested to merge it into https://github.com/koreader/contrib?
You may integrate the plugin with a pending https://github.com/koreader/koreader/pull/7419 to make it visible in "More Tools" menu if users decided to install it themselves. An example can be found at https://github.com/koreader/contrib/blob/Hzj-jie-findhistory/findhistory.koplugin/insert_menu.lua

PR #6849, opened 2020-11-05
Default extra_plugin_paths if necessary

jperon:
Some changes working on #6768:

- add a menu option to set extra_plugin_paths;
- add a default extra plugin paths in data_dir when default plugin path
  is read-only.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6849)

### DIFF
--- a/datastorage.lua
+++ b/datastorage.lua
@@ -54,7 +54,7 @@ local function initDataDir()
     local sub_data_dirs = {
         "cache", "clipboard",
         "data", "data/dict", "data/tessdata",
-        "history", "ota", "plugins",
+        "history", "plugins",
         "screenshots", "settings", "styletweaks",
     }
     for _, dir in ipairs(sub_data_dirs) do
--- a/datastorage.lua
+++ b/datastorage.lua
@@ -54,7 +54,7 @@ local function initDataDir()
     local sub_data_dirs = {
         "cache", "clipboard",
         "data", "data/dict", "data/tessdata",
-        "history", "ota",
+        "history", "ota", "plugins",
         "screenshots", "settings", "styletweaks",
     }
     for _, dir in ipairs(sub_data_dirs) do
--- a/frontend/pluginloader.lua
+++ b/frontend/pluginloader.lua
@@ -52,7 +52,6 @@ function PluginLoader:loadPlugins()
                 local extra_path_mode = lfs.attributes(extra_path, "mode")
                 if extra_path_mode == "directory" and extra_path ~= DEFAULT_PLUGIN_PATH then
                     table.insert(lookup_path_list, extra_path)
-                    logger.dbg("\n", "PLUGINS:", extra_path, "\n")
                 end
             end
         else
--- a/frontend/pluginloader.lua
+++ b/frontend/pluginloader.lua
@@ -46,6 +46,7 @@ function PluginLoader:loadPlugins()
     if extra_paths then
         if type(extra_paths) == "string" then
             extra_paths = { extra_paths }
+            G_reader_settings:saveSetting("extra_plugin_paths", extra_paths)
         end
         if type(extra_paths) == "table" then
             for _,extra_path in ipairs(extra_paths) do
--- a/frontend/pluginloader.lua
+++ b/frontend/pluginloader.lua
@@ -1,13 +1,5 @@
-local BD = require("ui/bidi")
-local ButtonDialogTitle = require("ui/widget/buttondialogtitle")
-local DataStorage = require("datastorage")
-local DownloadMgr = require("ui/downloadmgr")
-local InfoMessage = require("ui/widget/infomessage")
-local T = require("ffi/util").template
-local UIManager = require("ui/uimanager")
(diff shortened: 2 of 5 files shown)


jperon:
Just wondering for the menu option: is it better in `"device"` (like screenshot folder selection), in `"more_tools"` or in `"developer_options"` ?

Frenzie:
Sounds like developer options to me. Just adding a default `/sdcard/koreader/plugins` type folder there seems like it should pretty much do the trick.

jperon:
`/sdcard/koreader/plugins` would be specific to Android, and wouldn't cover the case of Appimage, for example. With `getDataDir()`, it covers all cases.

Frenzie:
I know. I mean that in contrast to all the lines about user selection.

frontend/pluginloader.lua
```diff
@@ -49,6 +58,11 @@ function PluginLoader:loadPlugins()
         else
             logger.err("extra_plugin_paths config only accepts string or table value")
         end
+    else
+        local datadir = DataStorage:getDataDir()
+        if datadir ~= "." then
+            G_reader_settings:saveSetting("extra_plugin_paths", { datadir .. "/plugins/" })
+        end
```
Frenzie, line 65: I.e., these lines good, the rest looks like feature creep.

Frenzie: Therefore *if you must*, put it in developer options.

jperon: OK, I see.

frontend/pluginloader.lua
```diff
@@ -49,6 +49,11 @@ function PluginLoader:loadPlugins()
         else
             logger.err("extra_plugin_paths config only accepts string or table value")
         end
+    else
+        local datadir = require("datastorage"):getDataDir()
```
Frenzie, line 53: Probably best to keep it data_dir just like there.

Frenzie (approved):
lgtm

frontend/pluginloader.lua
yparitcher: you can also move this after the `if` so it is not called on other platforms, and then change it to just a `elseif`

poire-z (approved):

pazos (commented):
A minor nit:

Create the `plugins` dir if `data_dir ~= "."`

I think it can be done at the end of [`initDataDir()`](https://github.com/koreader/koreader/blob/master/datastorage.lua#L53)

```lua
if DataStorage:getDataDir() ~= "." then
    local sub_data_dir = string.format("%s/%s", DataStorage:getDataDir(), "plugins")
    if lfs.attributes(sub_data_dir, "mode") ~= "directory" then
        lfs.mkdir(sub_data_dir)
    end
end
```

jperon:
@pazos I just added `"plugins"` to `sub_data_dirs`: if `data_dir` is `.`, `plugins/` already exists, and it will be detected by `if lfs.attributes(sub_data_dir, "mode") ~= "directory"`

datastorage.lua
```diff
@@ -54,7 +54,7 @@ local function initDataDir()
     local sub_data_dirs = {
         "cache", "clipboard",
         "data", "data/dict", "data/tessdata",
-        "history", "ota",
+        "history", "ota", "plugins",
```
Frenzie, line 57: If you're changing this line anyway, I don't think ota is relevant anymore on platforms where this function is relevant. Unless I'm mistaken @pazos?
```suggestion
        "history", "plugins",
```

Frenzie:
> initDataDir()

Yup, good catch. I added that because it was a bit confusing to users on platforms like Android since there was no self-documenting directory structure already in place.

datastorage.lua
pazos: appImage?

Frenzie: While I have considered looking into https://github.com/AppImage/AppImageUpdate (which iirc uses zsync?), I don't think that folder would be relevant to it. But right now it simply opens the link to the latest AppImage.

pazos (approved):

pazos:
@jperon @Frenzie 

It seems that "ota" folder was needed, at least on android. Here's what happens on a new install if not present:

 `11-07 21:27:13.492 17748 17774 W KOReader:  cannot find update file /storage/emulated/0/koreader/ota/koreader-android-latest-nightly: No such file or directory`

and a infomessage "update information not available on xxx channel".

Not a very serious issue since legacy ota dirs are preserved, but we should restore that behaviour to not break OTA updates for new users.

frontend/pluginloader.lua
```diff
@@ -49,6 +49,11 @@ function PluginLoader:loadPlugins()
         else
             logger.err("extra_plugin_paths config only accepts string or table value")
         end
+    else
+        local data_dir = require("datastorage"):getDataDir()
+        if data_dir ~= "." then
+            G_reader_settings:saveSetting("extra_plugin_paths", { data_dir .. "/plugins/" })
+        end
```
Hzj-jie, line 56: Did you forget to add the new extra path to lookup_path_list?

PR #6851, opened 2020-11-05
Methods to modify highlight menu.

jperon:
Still thinking about #6768, I was wondering wether such changes could be approved (IMO in the same vein as `registerToMainMenu()`).

For example, this would allow to implement changes like f2194c1201575c940b30e1117fc195c4d50266f7 without modifying internal files, as follows:

```lua
local EventListener = require("ui/widget/eventlistener")
local QRMessage = require("ui/widget/qrmessage")
local _ = require("gettext")

local QRHighlight = EventListener:new{
    name = "qrhighlight",
    is_doc_only = true,
}

function QRHighlight:init()
    self.ui.highlight:addToHighlightDialog("qrhighlight_QR", function(_reader_highlight_instance)
        return {
            text = _("QR"),
            enabled = Device:hasClipboard(),
            callback = function()
                UIManager:show(QRMessage:new{
                    text = _reader_highlight_instance.selected_text.text,
                    width = Device.screen:getWidth(),
                    height = Device.screen:getHeight()
                })
            end,
        }
    end)
end

return QRHighlight
```

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6851)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -16,7 +16,9 @@ local C_ = _.pgettext
 local T = require("ffi/util").template
 local Screen = Device.screen
 
-local ReaderHighlight = InputContainer:new{}
+local ReaderHighlight = InputContainer:new{
+    _extra_highlight_buttons = {}
+}
 
 function ReaderHighlight:init()
     self.ui:registerPostInitCallback(function()
@@ -499,15 +501,14 @@ function ReaderHighlight:onShowHighlightDialog(page, index)
     return true
 end
 
-local _extra_highlight_buttons = {}
 function ReaderHighlight:addToHighlightDialog(idx, fn_button)
     -- fn_button is a function that takes the ReaderHighlight as argument,
     -- and returns a table describing the button to be added.
-    _extra_highlight_buttons[idx] = fn_button
+   self._extra_highlight_buttons[idx] = fn_button
 end
 
 function ReaderHighlight:removeFromHighlightDialog(idx)
-	_extra_highlight_buttons[idx] = nil
+	self._extra_highlight_buttons[idx] = nil
 end
 
 function ReaderHighlight:onShowHighlightMenu()
@@ -613,7 +614,7 @@ function ReaderHighlight:onShowHighlightMenu()
         })
     end
 
-    for _, fn_button in util.opairs(_extra_highlight_buttons) do
+    for _, fn_button in util.opairs(self._extra_highlight_buttons) do
         if #highlight_buttons[#highlight_buttons] > 1 then
             table.insert(highlight_buttons, {})
         end
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -504,11 +504,13 @@ end
 function ReaderHighlight:addToHighlightDialog(idx, fn_button)
     -- fn_button is a function that takes the ReaderHighlight as argument,
     -- and returns a table describing the button to be added.
-   self._extra_highlight_buttons[idx] = fn_button
+    self._extra_highlight_buttons[idx] = fn_button
 end
 
 function ReaderHighlight:removeFromHighlightDialog(idx)
-	self._extra_highlight_buttons[idx] = nil
+    local button = self._extra_highlight_buttons[idx]
+    self._extra_highlight_buttons[idx] = nil
+    return button
 end
 
 function ReaderHighlight:onShowHighlightMenu()
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -11,6 +11,7 @@ local Translator = require("ui/translator")
 local UIManager = require("ui/uimanager")
 local logger = require("logger")
 local util = require("util")
+local ffiUtil = require("ffi/util")
 local _ = require("gettext")
 local C_ = _.pgettext
 local T = require("ffi/util").template
@@ -144,7 +145,7 @@ function ReaderHighlight:init()
     end
 
     if Device:canShareText() then
-        self:addToHighlightDialog("09_follow_link", function(_self)
+        self:addToHighlightDialog("10_share_text", function(_self)
             return {
                 text = _("Share text"),
                 callback = function()
@@ -618,10 +619,6 @@ function ReaderHighlight:addToHighlightDialog(idx, fn_button)
     -- fn_button is a function that takes the ReaderHighlight instance
     -- as argument, and returns a table describing the button to be added.
     self._highlight_buttons[idx] = fn_button
-    local n = 0
-    for _ in pairs(self._highlight_buttons) do
-        n = n + 1
-    end
 end
 
 function ReaderHighlight:removeFromHighlightDialog(idx)
@@ -634,12 +631,12 @@ function ReaderHighlight:onShowHighlightMenu()
     local highlight_buttons = {{}}
 
     local columns = 2
-    for _, fn_button in util.opairs(self._highlight_buttons) do
-        logger.dbg("ReaderHighlight", _, #highlight_buttons, #highlight_buttons[#highlight_buttons])
+    for idx, fn_button in ffiUtil.orderedPairs(self._highlight_buttons) do
         if #highlight_buttons[#highlight_buttons] >= columns then
             table.insert(highlight_buttons, {})
         end
         table.insert(highlight_buttons[#highlight_buttons], fn_button(self))
+        logger.dbg("ReaderHighlight", idx..": line "..#highlight_buttons..", col "..#highlight_buttons[#highlight_buttons])
     end
 
     self.highlight_dialog = ButtonDialog:new{
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -17,10 +17,146 @@ local T = require("ffi/util").template
 local Screen = Device.screen
 
 local ReaderHighlight = InputContainer:new{
-    _extra_highlight_buttons = {}
 }
 
+local function inside_box(pos, box)
+    if pos then
+        local x, y = pos.x, pos.y
+        if box.x <= x and box.y <= y
+            and box.x + box.w >= x
+            and box.y + box.h >= y then
+            return true
+        end
+    end
+end
+
+local function cleanupSelectedText(text)
+    -- Trim spaces and new lines at start and end
+    text = text:gsub("^[\n%s]*", "")
+    text = text:gsub("[\n%s]*$", "")
+    -- Trim spaces around newlines
+    text = text:gsub("%s*\n%s*", "\n")
+    -- Trim consecutive spaces (that would probably have collapsed
+    -- in rendered CreDocuments)
+    text = text:gsub("%s%s+", " ")
+    return text
+end
+
 function ReaderHighlight:init()
+    self._highlight_buttons = {
+        ["01_highlight"] = function(_self)
+            return {
+                text = _("Highlight"),
+                callback = function()
+                    _self:saveHighlight()
+                    _self:onClose()
+                end,
+                enabled = _self.hold_pos ~= nil,
+            }
+        end,
+        ["02_add_note"] = function(_self)
+            return {
+                text = _("Add Note"),
+                callback = function()
+                    _self:addNote()
+                    _self:onClose()
+                end,
+                enabled = _self.hold_pos ~= nil,
+            }
+        end,
+        ["03_copy"] = function(_self)
+            return {
+                text = C_("Text", "Copy"),
+                enabled = Device:hasClipboard(),
+                callback = function()
+                    Device.input.setClipboardText(cleanupSelectedText(_self.selected_text.text))
+                end,
+            }
+        end,
+        ["05_wikipedia"] = function(_self)
+            return {
+                text = _("Wikipedia"),
+                callback = function()
+                    UIManager:scheduleIn(0.1, function()
+                        _self:lookupWikipedia()
+                        -- We don't call _self:onClose(), we need the highlight
+                        -- to still be there, as we may Highlight it from the
+                        -- dict lookup widget
+                    end)
+                end,
+            }
+        end,
+        ["06_dictionary"] = function(_self)
+            return {
+                text = _("Dictionary"),
+                callback = function()
+                    _self:onHighlightDictLookup()
+                    -- We don't call _self:onClose(), same reason as above
+                end,
+            }
+        end,
+        ["07_translate"] = function(_self)
+            return {
+                text = _("Translate"),
+                callback = function()
+                    _self:translate(_self.selected_text)
+                    -- We don't call _self:onClose(), so one can still see
+                    -- the highlighted text when moving the translated
+                    -- text window, and also if NetworkMgr:promptWifiOn()
+                    -- is needed, so the user can just tap again on this
+                    -- button and does not need to select the text again.
+                end,
+            }
+        end,
+        ["08_search"] = function(_self)
+            return {
+                text = _("Search"),
+                callback = function()
+                    _self:onHighlightSearch()
+                    _self:onClose()
+                end,
+            }
+        end,
+    }
+
+    if not self.ui.document.info.has_pages then
+        self:addToHighlightDialog("04_view_html", function(_self)
+            return {
+                text = _("View HTML"),
+                callback = function()
+                    _self:viewSelectionHTML()
+                end,
+            }
+        end)
+    end
+
+    if self.selected_link ~= nil then
+        self:addToHighlightDialog("09_follow_link", function(_self)
+            return {
+                text = _("Follow Link"),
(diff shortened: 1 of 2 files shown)


poire-z:
Looks like a good idea to me.
addToHighlight**Dialog**() then
(If you manage to have a consistent clear/close/keep highlight dialog open behaviour - or setable by the plugin.)
(I think there were other plugins in the past that would have liked that, or stuff that was made into frontend code just because of that - so, good to have this option.)

jperon:
> (If you manage to have a consistent clear/close/keep highlight dialog open behaviour - or setable by the plugin.)

I must confess I don't understand this point.

> Dunno if plugins are loaded in some kind of order, or if that would make the buttons randomely added.

Good point: it seems to depend on the output of `lfs.dir()`, that's not sorted. Perhaps I could do something like that:

```lua
function ReaderHighlight:addToHighlightDialog(idx, fn_button)
    -- fn_button is a function that takes the ReaderHighlight as argument,
    -- and returns a table describing the button to be added.
    _extra_highlight_buttons[idx] = fn_button
end

function ReaderHighlight:removeFromHighlightDialog(idx)
	_extra_highlight_buttons[idx] = nil
end
```

This would:

- allow some control over the buttons' order (for example if one wants to have two nearby ones);
- allow overwriting / removing a button;
- prevent accidental duplicates.

poire-z:
> (If you manage to have a consistent clear/close/keep highlight dialog open behaviour - or setable by the plugin.)

Some buttons close the highlight dialog (and clear the highlighted text), others not (so we can hit another button when done).
Your callback doesn't do anything of that - dunno if it's done by ReaderHighlight when invoking the callback.

Btw, in your sample in a plugin:
```lua
local ReaderHighlight = require("apps/reader/modules/readerhighlight")
ReaderHighlight:addToHighlightMenu(function(self)
```
ReaderHighlight there is not an instance. You might want to do it thru `self.ui.highlight:addToHighlightMenu` - and be sure it's there, as the plugin might be set not document only (is_docless or something like that).

jperon:
I'll have to understand a bit better `self.ui`… I'll look at this !

poire-z:
self.ui is given/plugged to each plugin and each Reader* modules: it's THE instance of ReaderUI.
Thru it, each module instance can access any other module instance.

jperon:
> Btw, in your sample in a plugin:
> 
> ```lua
> local ReaderHighlight = require("apps/reader/modules/readerhighlight")
> ReaderHighlight:addToHighlightDialog(idx, function(self)
> ```
> 
> ReaderHighlight there is not an instance. You might want to do it thru `self.ui.highlight:addToHighlightDialog` - and be sure it's there, as the plugin might be set not document only (is_docless or something like that).

Perhaps I'm missing something, but if I implement `addToHighlightDialog` as an instance method, it means the button will have to be added each time I highlight text (and the dialog repainted); what I intended was more like a class method that would modify `ReadeHighlight` itself, so that all instances get buttons on creation. Perhaps I should change `ReaderHighlight:addToHighlightDialog` to `ReaderHighlight.addToHighlightDialog`, to make clear it doesn't act on an instance ?

poire-z:
> as an instance method, it means the button will have to be added each time I highlight text 

May be I didn't get how/when these new methods of yours are called.
But I thought, around the way our addToMenu stuff works that:
- Reader* modules instances are created by ReaderUI on each book opening
- Plugin instances are then created by ReaderUI in the same book opening step
- Plugins would call self.ui.highlight:addToHighlightDialog() in that same book opening step (so, only once, at this moment only - not on each highlight made.

(This at least allows plugins to be disabled or options changed with just reloading the book - not needing a full restart.)
So, you addToHighlightDialog() should fill a self.extra_buttons table (self being the ReaderHighlight instance) - and if the final table used for the dialog is created each time, buttons should be picked from that self.extra_buttons and inserte in it.

Frenzie:
> Perhaps I'm missing something, but if I implement addToHighlightDialog as an instance method, it means the button will have to be added each time I highlight text (and the dialog repainted)

It would only be called on init. If you implemented it incorrectly in the way you say then you'd get an ever increasing supply of buttons. :-P

jperon:
OK, it get clearer in my mind (not perfect yet, but it's coming…).

@Frenzie thanks to the key-value `_extra_highlight_buttons`, it wouldn't duplicate buttons, but it would be dirty.

jperon:
So I just updated the PR (and sample plugin code in the first message).

Now, getting further: would it be approved (sure it would be feasible, but would such a change be accepted) to get rid of `_extra_highlight_buttons`, and operate directly on `highlight_buttons` (this one becoming `self._highlight_buttons`, and `ReaderHighlight:addToHighlightDialog` operating on it) ? It would:

1. avoid the current behaviour of conditional buttons "Follow Link" and "Share Text" (single button in an added row);
2. let a plugin remove a button from the menu, for example if someone doesn't want "View HTML" or wants to implement another "Search";
3. let a plugin change buttons order in the menu.

(2. and 3. could be seen as quite invasive, but after all, it would be up to the user to install/enable or not such plugins.)

jperon:
> (If you manage to have a consistent clear/close/keep highlight dialog open behaviour - or setable by the plugin.)

In fact, it would be easily setable by the plugin: in the sample above, one would just add `o:onClose()` at the end of the callback function.

IMO, the fact that the behaviour isn't consistent between buttons has a good reason: you wouldn't want the menu to stay after "Copy" or "Follow Link", but you could want it after a dictionary or Wikipedia lookup (as documented in the code).

poire-z:
In:
```lua
function QRHighlight:init()
    self.ui.highlight:addToHighlightDialog("qrhighlight_QR", function(o)
```
I would wrap `self.ui.highlight:addToHighlightDialog` in a `if self.ui.highlight then`, just as a safety if `self.ui` happens to be a FileManager instance (and not only depend of is_doc_only being correct to avoid a crash).
Also rename `o` to something more explicit: `readerhighlight_instance` or `highlight_dialog_instance` (not sure an instance of what it is :) 

> get rid of _extra_highlight_buttons, and operate directly on highlight_buttons (this one becoming self._highlight_buttons, and ReaderHighlight:addToHighlightDialog operating on it) ? 

Haven't re-read the code - but if all these button tables are static and their name and callback code do not change depending on what's highlighted, I guess it's fine to build them all once and re-use them.

> 1) avoid the current behaviour of conditional buttons "Follow Link" and "Share Text" (single button in an added row);

How ? They would be created once, but still added conditionally, right?
The single row is ugly, but it's just there to not show as an always-there grey text and makes that dialog always 5 rows.

> 2) let a plugin remove a button from the menu, for example if someone doesn't want "View HTML" or wants to implement another "Search";

Well, I don't want your `QR` button - but I don't see why anybody would want to hide my `View HTML` :)
Ok, I understand one not interested might want to hide it - and I comment out quite a few stuff I'm not interested in other places in my own patch.
But I don't really like that a plugin is able to shuffle/re-order the current layout of "core" buttons: people might be used to it - and the idea of enabling a plugin (that might provides other features) could shuffle that (the user might not be interested in its added highlight button feature) feels bad.

> 3) let a plugin change buttons order in the menu.

It could be a plugin, but this might as well be a core feature, like we allow re-ordering footer items.
(or it could be just a user made file, like it's allowed for re-ordering menu items - I guess not many people use that). Having that as a plugin feels like too many indirections. Better it be a refactoring of ReaderHighlight, and it providing services to plugins (like you started with this PR).
But it feels a bit overkill to have that feature (but I say that because it's easy for me to patch the code to have them the way I like it), and quite some work. But it it was there, and available deep in some menu so it's not the first thing users will go mess with and wonder how they screw what used to look nice and go opening issues, why not :)

jperon:
> > get rid of _extra_highlight_buttons, and operate directly on highlight_buttons (this one becoming self._highlight_buttons, and ReaderHighlight:addToHighlightDialog operating on it) ?
> 
> Haven't re-read the code - but if all these button tables are static and their name and callback code do not change depending on what's highlighted, I guess it's fine to build them all once and re-use them.
> 
> > 1. avoid the current behaviour of conditional buttons "Follow Link" and "Share Text" (single button in an added row);
> 
> How ? They would be created once, but still added conditionally, right?
> The single row is ugly, but it's just there to not show as an always-there grey text and makes that dialog always 5 rows.

[Actually](https://github.com/koreader/koreader/blob/91708e25d0c7ad340675e58b634ba0a46e2d40fa/frontend/apps/reader/modules/readerhighlight.lua#L502-L610), I think the menu is re-built each time one highlights text. I could replace `local highlight_buttons = {{...}, {...}}` and each `table.insert highlight_buttons, {{...}}` by `self:addToHighlightDialog()` calls, so that when you have an even number of buttons, they are on two columns, and only the last one in an odd number would be alone on its line.

> But I don't really like that a plugin is able to shuffle/re-order the current layout of "core" buttons: people might be used to it - and the idea of enabling a plugin (that might provides other features) could shuffle that (the user might not be interested in its added highlight button feature) feels bad.

I don't mean it should be a documented feature, or that it should be accepted in "official" plugins; the fact is that my proposed implementation would allow that.

> 
> > 1. let a plugin change buttons order in the menu.
> 
> It could be a plugin, but this might as well be a core feature, like we allow re-ordering footer items.
> (or it could be just a user made file, like it's allowed for re-ordering menu items - I guess not many people use that).

I'd find it good to have user made files, but I think such features should be documented (I just learnt it was possible for menu items by reading you reply !)

> Having that as a plugin feels like too many indirections. Better it be a refactoring of ReaderHighlight, and it providing services to plugins (like you started with this PR).

*That* wouldn't be as a plugin: the plugin would *use* this feature by the mean of `:addToHighlightDialog()` and `:removeFromHighlightDialog()`.

poire-z:
> I don't mean it should be a documented feature, or that it should be accepted in "official" plugins; the fact is that my proposed implementation would allow that.

As soon as you make the button list a member of the instance, it's already allowed. Any plugin can tweak `self.ui.highlight._highlight_buttons`. So ok, no need to document that and it wouldn't need any specific API.

> That wouldn't be as a plugin: the plugin would use this feature by the mean of :addToHighlightDialog() and :removeFromHighlightDialog().

OK, you could let in a removeFromHighlightDialog if you want, but just for consistency/symmetry :) and may be if you give ids to buttons, to more easily allow accessing/removing them, change the name, wrap the callback.
(I thought you wanted to create a plugin that would allow re-ordering/removing the buttons in that dialog - in which case _this_ shouldn't be a plugin).

> wants to implement another "Search";

Search is bit too core and plugged into the rendering engines for any plugin to be able to replace it :)
But OK, Translate could be overriden by a plugin, Wikipedia could be replaced.

So, if you don't plan doing a Menu/UI for re-ordering buttons - no problem for me with your proposed small refactoring of buttons handling - including any manipulation methods as long as they look consistent among themselves (and how user plugins use them is none of our business :)

> I could replace local highlight_buttons = {{...}, {...}} and each table.insert highlight_buttons, {{...}} by self:addToHighlightDialog() calls, so that when you have an even number of buttons, they are on two columns, and only the last one in an odd number would be alone on its line.

Let self:addToHighlightDialog()  be the public API for plugins.
But yes, in ReaderHighlight:onShowHighlightMenu(), you could build the local button by iterating the instance `_highlight_buttons`, doing some checks and ignoring those that don't apply, some possible re-ordering and best-fit/layout if you get an even numbers of them - eventually doing that in another method/local function than addToHighlightDialog().

jperon:
I think it should be ready for review.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -16,9 +16,147 @@ local C_ = _.pgettext
...
+                    local link = _self.selected_link.link or _self.selected_link
+                    _self.ui.link:onGotoLink(link)
+                    _self:onClose()
+                end,
+            }
+        end)
+    end
+
+    if Device:canShareText() then
+        self:addToHighlightDialog("09_follow_link", function(_self)
```
poire-z, line 147: You forgot to change `09_follow_link` :)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -16,9 +16,147 @@ local C_ = _.pgettext
...
+                callback = function()
+                    _self:onHighlightSearch()
+                    _self:onClose()
+                end,
+            }
+        end,
+    }
+
+    if not self.ui.document.info.has_pages then
+        self:addToHighlightDialog("04_view_html", function(_self)
```
poire-z, line 123: This will change the look of the highlight dialog whether in PDF or EPUB (starting from the Wikipedia button, all will be shifted, on the left side of that dialog in one format, on the right side in the other format), which is not great.
Not fond of putting it at the end, as it would shift everything next, and we might be used to finding the buttons where they currently are :/ 
We should invent something equivalent for PDF :)

Or have a bit rethinking about the button orders and positions to justify a global "pérenne" new ordering.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -499,107 +614,32 @@ function ReaderHighlight:onShowHighlightDialog(page, index)
...
-                end,
-            },
-        })
+function ReaderHighlight:addToHighlightDialog(idx, fn_button)
+    -- fn_button is a function that takes the ReaderHighlight instance
+    -- as argument, and returns a table describing the button to be added.
+    self._highlight_buttons[idx] = fn_button
+    local n = 0
+    for _ in pairs(self._highlight_buttons) do
+        n = n + 1
```
poire-z, line 623: (It's short so I don't mind - but we have `util.tableSize(t)`).
I might be confused by the diff - but it seems you don't do anything with that local n ?!

jperon: Sorry, it was for debugging and I forgot it.

frontend/util.lua
```diff
@@ -330,6 +330,24 @@ function util.tableSize(t)
...
+    end
+    table.sort(_t, fn)
+    local i = 0
+    return function()
+        i = i + 1
+        local k = _t[i]
+        if k then return k, t[k] end
+    end
+end
+
```
NiLuJe, line 350: We already have [ffiUtil.orderedPairs](https://github.com/koreader/koreader-base/blob/ec3d5021852a1acf3471364a020d07b7de439356/ffi/util.lua#L675-L678) ;).

jperon:
> Or have a bit rethinking about the button orders and positions to justify a global "pérenne" new ordering.

Would you agree with following logic:

- keep the first line as is: *highlighting* / *bookmarks*;
- then most internal functions, that don't depend on anything else: *copy* / *search*;
- then search functions that have dependencies, but don't depend on the internet: *dictionary* / *translate*;
- then the search function that depends on the internet: *Wikipedia*;
- then what's about exporting text: for now, only *HTML View* (after which I would place QR with an extension as shown above).

IMHO, that way we would put first those that would be the most necessary and the least subject to change.

> We should invent something equivalent for PDF

How about a *Text view*, especially when using OCR ?

jperon:
Just a preview of my proposal:

hlmenu

poire-z:
I'm fine with your proposal (mostly because it feels like you didn't change much :) 

> but don't depend on the internet: dictionary / translate;

Translate depends on the internet (it uses Google translate).

May be just have our current Search & View HTML switched.
Keep Wikipedia and Translate on the left (as they trigger a network connection, so less chance to hit them unwillingly on the left for right handed people).
I have the feeling chinese & japanese users may use Dictionary thru this dialog as they may have to select multiple glyphs (a single hold would select only one) - so it might be best to keep Dictionary on the right as it was (even if for us westerners, it feels a bit like uneeded on that dialog as we mostly lookup single words).

```
   Copy   |  Search
Wikipedia | Dictionary
Translate | View HTML
```

> last follow link, as it is only for links, and generally, a single tap is enough to follow them.

OK for it being last (so, single, on on the right).
It can also be used by those that disabled Tap to follow links (even if it's a bit complicated to use that to follow a link :)

> How about a Text view, especially when using OCR ?

Well, dunno - I don't read much PDF.
But may be it is a good idea :) I see it could make sense as a kind of quick zoom solution: select an area of small text to read it at your prefered font size in a TextViewer - instead of having to mess with zoom mode or reflow.

jperon:
Done.

NiLuJe:
I'm less particular about menu ordering myself, but this sounds fine to me, too, FWIW ;).

poire-z (approved):

poire-z:
Some issue since this: when using <kbd>Search</kbd>, and it should stay on the current page and highlight the current matches on the current page (so, at least the text we selected), there is nothing highlighted on the page.
Looks like we did not discuss this in the above discussion, so I assume it's a small cut & paste error.
Fixable by just restoring what was done before:
```diff
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -83,7 +83,10 @@ function ReaderHighlight:init()
                 text = _("Search"),
                 callback = function()
                     _self:onHighlightSearch()
-                    _self:onClose()
+                    UIManager:close(self.highlight_dialog)
+                    -- We don't call _self:onClose(), crengine will highlight
+                    -- search matches on the current page, and self:clear()
+                    -- would redraw and remove crengine native highlights
                 end,
             }
         end,
```
(I'll make a PR with that in the coming days.)

PR #6852, opened 2020-11-06
Add menu item allowing users to disable "hold on footer to show Skim Widget"

Galunid:
Rationale: 
Today I learned that there are system-wide gestures on Onyx devices, and while performing those, skim widget kept popping up and there is no way to disable it.
image

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6852)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -2043,7 +2043,6 @@ function ReaderFooter:onHoldFooter()
         self.ui:handleEvent(Event:new("ShowSkimtoDialog"))
         return true
     end
-    return false
 end
 
 function ReaderFooter:setVisible(visible)
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -945,7 +945,7 @@ function ReaderFooter:addToMainMenu(menu_items)
                 end,
             },
             {
-                text = _("Show SkimWidget by holding the footer"),
+                text = _("Hold (footer) to skim"),
                 checked_func = function()
                     return self.settings.skim_widget_on_hold
                 end,
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -428,6 +428,9 @@ function ReaderFooter:init()
     if not self.settings.book_chapter_max_width_pct then
         self.settings.book_chapter_max_width_pct = 30
     end
+    if self.settings.skim_widget_on_hold == nil then
+        self.settings.skim_widget_on_hold = true
+    end
     self.mode_list = {}
     for i = 0, #self.mode_index do
         self.mode_list[self.mode_index[i]] = i
@@ -939,6 +942,16 @@ function ReaderFooter:addToMainMenu(menu_items)
                 end,
                 callback = function()
                     self.settings.lock_tap = not self.settings.lock_tap
+                end,
+            },
+            {
+                text = _("Show SkimWidget by holding the footer"),
+                checked_func = function()
+                    return self.settings.skim_widget_on_hold
+                end,
+                callback = function()
+                    self.settings.skim_widget_on_hold = not self.settings.skim_widget_on_hold
+                    G_reader_settings:saveSetting("footer", self.settings)
                 end,
             },
             {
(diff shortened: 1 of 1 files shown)


Galunid:
@Frenzie Please take a look at phrasing when you have a moment

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -2026,7 +2039,9 @@ end
 
 function ReaderFooter:onHoldFooter()
     if self.mode == self.mode_list.off then return end
-    self.ui:handleEvent(Event:new("ShowSkimtoDialog"))
+    if self.settings.skim_widget_on_hold then
+        self.ui:handleEvent(Event:new("ShowSkimtoDialog"))
+    end
     return true
 end
```
poire-z, line 2046: Hold on footer is the last handler for hold (#5789), so this is not really an issue - but logically, you'll return true only if you handled it - and let it return without true if you're skipping handling it, as something next (you don't know it there :) could be handling it.

poire-z:
No problem with adding that (it's discreet in a far down menu).
But are you sure there's no other way to solve the bug you noticed, and that's it's not an edge case we don't handle in our gesture detector handling code? (I'm sure it's far more complicated to solve :)

Galunid:
@poire-z I don't think there is actually any bug. What I think happens is:
1. We start "hold" event
2. We swipe up
3. Onyx (android) consumes the swipe
4. KOReader is still left with "hold" event
5. Hold is released, Skim Widget gets triggered
I can be wrong though. Can't confirm this yet, but I can't reproduce it on any other device

poire-z:
I don't know what there "system-wide gestures" are - but can't this trigger any other hold stuff, like holding on a word to do dict lookup or text selection ? Or just always at screen boundaries ?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -2041,8 +2041,9 @@ function ReaderFooter:onHoldFooter()
     if self.mode == self.mode_list.off then return end
     if self.settings.skim_widget_on_hold then
         self.ui:handleEvent(Event:new("ShowSkimtoDialog"))
+        return true
     end
-    return true
+    return false
```
NiLuJe, line 2046: Probably better as "nothing". I don't exactly remember how this is handled on our side, but Lua has weird conventions where retruning nothing is slightly different from returning false, and also slightly different than returning nil ;).

Frenzie: `return nil` is the same as returning nothing. ;-)

NiLuJe: There are apparently sneaky subtleties: http://lua-users.org/lists/lua-l/2011-09/msg00289.html

Galunid: Was pretty surprised, but there is some difference https://stackoverflow.com/a/18526055

Frenzie: Fair, not 100% the same thing, but not different different like false. Eh, never mind. But just get rid of the line. :-P

Galunid:
The system wide gestures are swipes from the very bottom of the screen up. There are 3: left side, middle, right side and you can assign them action (i.e. back, home, screenshot, none). I believe it has something to do with how onyx handles that. Even if I assign the "none" action, it still gets bugged (swipe up is detected as hold on footer). I couldn't reproduce it on Kobo/Android phone, so it's probably not a problem with our gesture handling.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -941,6 +944,16 @@ function ReaderFooter:addToMainMenu(menu_items)
                     self.settings.lock_tap = not self.settings.lock_tap
                 end,
             },
+            {
+                text = _("Show SkimWidget by holding the footer"),
```
poire-z, line 948: Just "Skim on hold" ? "Long-press to skim" ? @Frenzie ?

Galunid: I was thinking something like this was not really user friendly, but it's fine by me. I'll wait for Frenzie to comment when he has some free time.

Frenzie: "SkimWidget" isn't either. ;-)

I'd go with `Hold (footer) to skim`.

poire-z: I assume `(footer)` was meant not to be included literally, but as to allow @Galunid to choose between these 2 alternatives:
`Hold to skim` or `Hold footer to skim`
(I'm fine with any of them.)

Galunid: My bad, opened PR with fix

PR #6856, opened 2020-11-07
Apply CSS tweaks for links to all elements below a elements.

tastytea:
This is for ebooks that have `<span>` inside of `<a>`.

I've come across ebooks with hyperlinks like this: `<a href="fn.html#fn-1"><span class="a">1</span></span></a>`. The span has CSS that changes the color and overrides the “a_black” CSS fix. With this PR, the CSS for the span gets overridden too.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6856)


poire-z (approved):
Looks alright (can't think of any side effect).

PR #6858, opened 2020-11-08
add Screenshot to Dispatcher

### LINKED ISSUE #6592: FR: [add screenshot to dispatcher]
From https://www.mobileread.com/forums/showpost.php?p=4028556&postcount=23

> I take screenshots on my Kindle and Poke2 with press the bottom left and top right corner at the same time. This does not work in my emulator, because I only have one mouse.

From https://www.mobileread.com/forums/showpost.php?p=4028573&postcount=24

> One thing I would like though is for KOReader to allow for taking the screenshot when the selected image is shown at full screen.

Screenshot can be used with any gesture the user wants.

yparitcher:
Fixes #6592

I needed to add `Screenshoter` to `self` as until now it was just listed as a `active_widget` and would only receive `sendEvent` but not regular `handleEvent` like other widget children. this shouldn't cause too much overhead as they should point to the same table. If there is a better way to fix this please let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6858)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -459,8 +459,11 @@ function FileManager:init()
     self.menu = FileManagerMenu:new{
         ui = self
     }
-    self.active_widgets = { Screenshoter:new{ prefix = 'FileManager' } }
-    table.insert(self, self.active_widgets[1])
+
+    local screenshoter = Screenshoter:new{ prefix = 'FileManager' }
+    table.insert(self, screenshoter) -- for regular events
+    self.active_widgets = { screenshoter } -- to get events even when hidden
+
     table.insert(self, self.menu)
     table.insert(self, FileManagerHistory:new{
         ui = self,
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Seems alright.

frontend/apps/filemanager/filemanager.lua
```diff
@@ -460,6 +460,7 @@ function FileManager:init()
         ui = self
     }
     self.active_widgets = { Screenshoter:new{ prefix = 'FileManager' } }
+    table.insert(self, self.active_widgets[1])
```
poire-z, line 463: May be to have that clearer:
```lua
local screenshoter = Screenshoter:new{ prefix = 'FileManager' }
self.active_widgets = { screenshoter } -- so that xxx
table.insert(self, screenshoter) -- so that yyy
```
(and same comments in readerui so we don't wonder why when we rereader that in some years :)

PR #6854, opened 2020-11-06
[CRe] Ensure toggling nightmode invalidates the drawCurrentView cache

NiLuJe:
Instead of relying on `Screen.night_mode` at the actual call site, rely on an internal flag, toggled via a set* function, which ensures the cache framework will reset the cache, allowing the page to be re-rendered (and the image counter-inverted properly, and, more importantly, reliably ^^).

Fix #6845 (at least as far as I'm concerned, as I couldn't find anything else going awry on Kobo either with HW or SW invert ^^).

Now also includes minor ReaderHighlight tweaks to avoid spurious cache invalidations.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6854)

PR #6859, opened 2020-11-08
allow clearing CreDocument's callcache from frontend

yparitcher:
I think this should be a simpler fix for #6854 allowing us not to track nightmode in 2 places.

unless there is a specific reason not to export `_callCacheReset()`?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6859)

### DIFF
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -877,9 +877,9 @@ function CreDocument:setNightmodeImages(toggle)
     self._nightmode_images = toggle
 end
 
+-- no-op that will be wrapped by setupCallCache
 function CreDocument:resetCallCache()
     logger.dbg("CreDocument: reset call cache")
-    self._callCacheReset()
 end
 
 function CreDocument:setFloatingPunctuation(enabled)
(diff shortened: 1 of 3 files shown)


poire-z (commented):
> unless there is a specific reason not to 

I don't see any. No problem with either solution for me - I let @NiLuJe be the judge :)

NiLuJe:
Yup, LGTM, thanks :).

NiLuJe (approved):

NiLuJe (changes requested):
It's missing the removal of the bit from devicelistener than called the new setNightMode, though ;).

EDIT: No it's not, I'm blind.

NiLuJe (approved):

yparitcher:
I didn't realize the call cache could be disabled, this should do the trick

frontend/document/credocument.lua
```diff
@@ -877,9 +877,9 @@ function CreDocument:setNightmodeImages(toggle)
     self._nightmode_images = toggle
 end
 
+-- no-op that will be wrapped by setupCallCache
 function CreDocument:resetCallCache()
     logger.dbg("CreDocument: reset call cache")
```
NiLuJe, line 882: You can also drop the logging, `add_reset` already takes care of it ;).

PR #6868, opened 2020-11-11
Order keys in settings.reader.lua

jperon:
All is in the title.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6868)

### DIFF
--- a/frontend/dump.lua
+++ b/frontend/dump.lua
@@ -29,15 +29,13 @@ local function _serialize(what, outt, indent, max_lv, history, _pairs)
         local didrun = false
         insert(outt, "{")
         for k, v in _pairs(what) do
-            if didrun then
-                insert(outt, ",")
-            end
             insert(outt, "\n")
             insert(outt, string.rep(indent_prefix, indent+1))
             insert(outt, "[")
             _serialize(k, outt, indent+1, max_lv, new_history, _pairs)
             insert(outt, "] = ")
             _serialize(v, outt, indent+1, max_lv, new_history, _pairs)
+            insert(outt, ",")
             didrun = true
         end
         if didrun then
--- a/frontend/dump.lua
+++ b/frontend/dump.lua
@@ -6,7 +6,7 @@ local isUbuntuTouch = os.getenv("UBUNTU_APPLICATION_ISOLATION") ~= nil
 local insert = table.insert
 local indent_prefix = "    "
 
-local function _serialize(what, outt, indent, max_lv, history)
+local function _serialize(what, outt, indent, max_lv, history, _pairs)
     if not max_lv then
         max_lv = math.huge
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
Seems harmless enough.

poire-z (approved):
Good idea.

Frenzie (commented):
Btw, there's no need to put a period at the end of the commit message.

poire-z:
Should this also be done with DocSettings, so we get nice looking and more human readable metadata.epub.lua ?
```diff
--- a/frontend/docsettings.lua
+++ b/frontend/docsettings.lua
@@ -195,3 +195,3 @@ function DocSettings:flush()
     self:ensureSidecar(self.sidecar)
-    local s_out = dump(self.data)
+    local s_out = dump(self.data, nil, true)
     os.setlocale('C', 'numeric')
```

Frenzie:
It might be good to double check how performance works out on a bookmark-heavy file.

poire-z:
On my Kobo, switching between a zero bookmarks book (small durations below) and the 275 bookmarks book from #5766 (larger durations) :

Currently:
```
WARN    docsettngs dump orig took 0.050 seconds
WARN    docsettngs dump orig took 0.003 seconds
WARN    docsettngs dump orig took 0.077 seconds
WARN    docsettngs dump orig took 0.001 seconds
WARN    docsettngs dump orig took 0.050 seconds
```
With sorting:
```
WARN    docsettngs dump sorted took 0.077 seconds
WARN    docsettngs dump sorted took 0.009 seconds
WARN    docsettngs dump sorted took 0.090 seconds
WARN    docsettngs dump sorted took 0.006 seconds
WARN    docsettngs dump sorted took 0.070 seconds
WARN    docsettngs dump sorted took 0.008 seconds
WARN    docsettngs dump sorted took 0.066 seconds
```

So, yes, takes a bit more time - but I'm not sure how to appreciate the differences :)

Frenzie:
I suppose that means the interaction with FAT32 takes the bulk of the effort, unless it's more of a theoretical concern?

poire-z:
I don't think anything FAT32 is at play here: whether sorted or not, the amount written should be the same - so, I guess it's purely the sorting (and the little temporary table added to each table with sorted keys).
Even if my test samples are small, the worst conclusion would be:
Small metadata: 3ms > 9ms
Larger metadata: 50ms > 90ms

Frenzie:
> I don't think anything FAT32 is at play here: whether sorted or not, the amount written should be the same

Yes, and people say it gets slow to close books. I wouldn't know because I'm a very light bookmark/highlight user. :) But given the minute differences in what you posted above, any slowness would therefore have to come from the writing stage.

Issue #2589, opened 2017-02-26
FR: automate advanced settings doc


Frenzie:
* KOReader version: up to 7fb2dc8

#### Issue

@poire-z made [an interesting observation](https://github.com/koreader/koreader/pull/2583#issuecomment-282516140):

> We should have a wiki page documenting the available settings that are not manageable thru menus (i added quite a few these last months).

This should print an alphabetic list of all the available options.
```
grep --include \*.lua -hoPr "(?<=G_reader_settings:readSetting\(\")\w+(?=\"\))" |sort |uniq
```

Ideally we'd automatically add some explanation as well somehow, but for starters just getting the (slightly further processed) output of that command or something equivalent in auto-generated docs somehow would be nice.

AlanSP1:
Yes, this would be very useful and also necessary. At the moment there's many options, but if we don't know how to use them, or that they even exist, we can't use them. Having them at least listed in wiki (or in some pdf or epub file that comes with koreader) would be great.

poire-z:
There's also some with `G_reader_settings:nilOrTrue` and friends.
But looking at grep outputs, i'm not sure there're really many settings not reachable via menus.
So, I don't know if it would not be better to make menus for them.

These are the ones i added or that have been discussed in topics i followed:

`G_reader_settings:nilOrTrue("wikipedia_prettify")` (wikipedia headings unicode symbol prefix)
`G_reader_settings:nilOrTrue("dict_justify")` (dict & wiki text justification)
those were cause i made them the default after i added these features, to give a chance for people who wouldn't like them to disable them (is there really some reason to dislike them ? may be i was just a bit shy :) should they stay ? should there be menus for that ?)

`wikipedia_languages` : a list of wikipedia languages the user wants, hard to implement as a menu, the available items may change and be huge, and there should be the notion of multiple choice and ordering them... fallback to UI Language + book language (if available) + "EN" if not set
`wikipedia_save_dir`: directory to save wikipedia epub to (fallback to other download dir if not set) - could be as easy as cut & paste from other similar "select directory" menu

`show_file_in_bold` (|false|"opened") : conditions for filenames to be displayed in bold (default: new files (files with no sidecar) in bold) in filemanager
`autoremove_deleted_items_from_history` : self explanatory, filemanager

I haven't really checked each of the other ones, but from their names, there's no surprise and they all look like they are related to some menu entry i've seen :)

retrue:
[...]

PR #6869, opened 2020-11-11
Advanced settings

jperon:
To be rebased on `master` after merge of #6868.

This plugin adds a menu entry into `More Tools` (I wanted to put it into `Developper Options` but as it is only available in FileManager, it isn't possible to use it in `sorting_hint`), to directly change values in `G_reader_settings`.

As there is no list of possible settings, it hooks into `LuaSettings` to gather undefined ones on `readSetting()`, `saveSetting()` and `delSetting()`.

It is proposed as a plugin in order to be the least intrusive possible; but should it be seen as useful, the most tricky part (`main.lua`, lines 25-48) would be cleaner if integrated into `luasettings.lua`.

Should it be rejected (what I'd understand without any difficulty), would it be possible to merge at least 3b49986ba124599a3a2bb0fd8e9261ead8e0ccb1 ? It would make manipulation of table dumps easier, either manually (when copy-pasting within `settings.reader.lua`, for example) or programmatically (cf. `main.lua`, lines 72-82).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6869)

### DIFF
--- a/frontend/dump.lua
+++ b/frontend/dump.lua
@@ -28,16 +28,14 @@ local function _serialize(what, outt, indent, max_lv, history)
         local new_history = { what, unpack(history) }
         local didrun = false
         insert(outt, "{")
-        for k, v in pairs(what) do
-            if didrun then
-                insert(outt, ",")
-            end
+        for k, v in _pairs(what) do
             insert(outt, "\n")
             insert(outt, string.rep(indent_prefix, indent+1))
             insert(outt, "[")
             _serialize(k, outt, indent+1, max_lv, new_history)
             insert(outt, "] = ")
             _serialize(v, outt, indent+1, max_lv, new_history)
+            insert(outt, ",")
             didrun = true
         end
         if didrun then
--- a/plugins/settingtweak.koplugin/_meta.lua
+++ b/plugins/settingtweak.koplugin/_meta.lua
@@ -1,0 +1,6 @@
+local _ = require("gettext")
+return {
+    name = 'settingtweak',
+    fullname = _("Tweak global settings"),
+    description = _([[This plugin allows you to tweak global settings from KOReader.]]),
+}
--- a/plugins/settingtweak.koplugin/main.lua
+++ b/plugins/settingtweak.koplugin/main.lua
@@ -1,0 +1,128 @@
+local DataStorage = require("datastorage")
+local FFIUtil = require("ffi/util")
+local InputDialog = require("ui/widget/inputdialog")
+local LuaSettings = require("luasettings")
+local UIManager = require("ui/uimanager")
+local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local _ = require("gettext")
+local T = FFIUtil.template
+local dump = require("dump")
+local util = require("util")
+
+
+local SettingTweak = WidgetContainer:new{
+    name = 'settingtweak',
+}
+
+
+local undefined_settings_cache_dir = table.concat({DataStorage:getDataDir(), "cache", "settingtweak"}, '/')
+local undefined_settings_cache = FFIUtil.joinPath(undefined_settings_cache_dir, "undefined_settings.lua")
+local initialized = false
+
+function SettingTweak:init()
+    if not initialized then
+        lfs.mkdir(undefined_settings_cache_dir)
+        local luasettings = LuaSettings
+        luasettings._undefined_settings = {}
+        local init = luasettings.init
+        local delSetting = luasettings.delSetting
+        local readSetting = luasettings.readSetting
+        local saveSetting = luasettings.saveSetting
+        function luasettings.init(_self, ...)
+            init(_self, ...)
+            _self._undefined_settings = {}
+        end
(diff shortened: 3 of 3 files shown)


pazos:
@jperon: user case? It can pretty much be a **power user** case but needs to have one. Which is the purpose?. Edit settings within the program? Why?.

I mean, nothing against the work (I didn't review it yet but I do agree it should be integrated in dev opts if useful, not under a plugin) but I miss the purpose/rationale.

poire-z:
I also can't see the use case :/ and I have a hard time guessing what will be proposed to be edited. A screenshot could help :)

I'm fine with the first commit, adding a comma after the last element in a table.
We might have a few people/apps that may parse our settings, mostly only the doc settings (metadata.epub.lua), which may or may not stumble on the change (but we have no contract and made no promises :) 
Pinging the only one I know, @noembryo https://github.com/noembryo/KoHighlights : will you be fine?

NiLuJe:
@poire-z: slpp seems fine with it:

```
┌─(ROOT@dahlia:pts/0)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~)─┐
└─(0.27:23%:01:12:38%:#)── python3                                                                                                                                                                                                                                                                                                                     ──(Thu, Nov 12)─┘
Python 3.8.6 (default, Oct  1 2020, 21:45:24) 
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pathlib import Path
>>> from slpp import slpp as lua
>>> blah_data = Path("blah.lua").read_text()
>>> lua.decode(blah_data)
{'blah': [1, 2, 3]}
>>> 
┌─(ROOT@dahlia:pts/0)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~)─┐
└─(0.19:23%:01:13:38%:#)── cat blah.lua                                                                                                                                                                                                                                                                                                                ──(Thu, Nov 12)─┘
{
    blah = { 1, 2, 3, }
}
```

noembryo:
@poire-z  I use a fork of slpp ([slppu](https://github.com/noembryo/slppu)) that also decodes the
```lua
{
    blah = { 1, 2, 3, }
}
```
as `{'blah': [1, 2, 3]}` with no problems
... but it doesn't put the comma back when re-encodes it...
Is it going to be a problem with that?

jperon:
@noembryo No, it won't be a problem: `dump` gets used only to serialize, not to deserialize tables.

jperon:
Here is the screenshot:

settingtweak

jperon:
As for the purpose:

- it makes it possible to (indirectly) edit `settings.reader.lua`, which would get overwritten on exit if directly edited in TextEditor; I already needed that when not having a cable at hand;
- for newcomers (like me :) who don't know all settings that may be defined in `settings.reader.lua`, it creates (and caches) a list of them;
- when developping a new module or plugin, the setting menus are often (at least in my case…) the last thing one wants to code; in that case, while testing on a physical reader, it might be useful to have a quick-and-dirty access to settings.

poire-z:
Well, not convinced :) 
Nobody needs to _know all settings that may be defined_ , because they mostly all can be set via the UI menus. 4 years ago, there were a lot that could only be set manually, but since then, we made mostly all of them tweakable via the UI - so, there's really no use editing them manually. (https://github.com/koreader/koreader/issues/2589#issuecomment-323927682, not updated for some time, lists some really useless ones that might not be).
And if you were to know the settings name, you might not know the possible values and dependancies on the value of other settings. So, better let that to the UI code that do know and deal with that.
As for developping on a device, well, change your habits and do your menu first :)

Frenzie:
> it makes it possible to (indirectly) edit settings.reader.lua, which would get overwritten on exit if directly edited in TextEditor; I already needed that when not having a cable at hand;

Would you mind expanding on why you needed that? That's either a much more important issue or just a user preference. ;-)

jperon:
@Frenzie In my case, it was due to my bad habits :). But I thought then that it could be handy in other cases.

I must say it's not without reasons I made it a plugin at first… So I'll keep it as such !

Do I have to revert the last commit to get the first one merged, or is it possible to merge one commit without the following one ?

Frenzie:
> Do I have to revert the last commit to get the first one merged, or is it possible to merge one commit without the following one ?

Only by causing us extra work (i.e., cherry picking), so it's preferable if you did that yourself in a second PR with just that commit to keep this PR as an archive. Or you could force push here if you prefer, but then other people wouldn't have access anymore.

poire-z:
> Do I have to revert the last commit to get the first one merged, or is it possible to merge one commit without the following one ?

Or you could include the first commit (added comma) into your not yet merged #6868.

jperon:
Just added to #6868.

Thanks !

Issue #6775, opened 2020-10-09
[translation] Arabic is missing from Weblate


WaseemAlkurdi:
Hi folks! 
I wanted to continue my unfinished Arabic translation of KOReader, only to find that for some reason, it's no longer available on Hosted Weblate. Can anyone check what would the issue possibly be?

WaseemAlkurdi:
Alright, so I requested a new translation, it was approved within a few minutes, and I uploaded `koreader.po` straight from the koreader/koreader-translations repository ... it worked, but isn't it odd why would the translation "vanish" all of the sudden?

Frenzie:
It hasn't, you tricked me (in a manner of speaking).

Frenzie:
The ar_AA name is apparently the name from Transifex but Weblate just uses ar.

Frenzie:
It's in https://github.com/koreader/koreader-translations/blob/b077a9ab66f28393d8048c5078c35dd65cc6c035/ar_AA/koreader.po

Frenzie:
So I think I should delete the new one and rename that one for continuity, probably?

Frenzie:
Since now it's double.

Frenzie:
Also the old one seems to have a few more things in it?
![Screenshot_2020-10-09_19-03-36](https://user-images.githubusercontent.com/202757/95611466-26dca800-0a62-11eb-93d1-85a3b2e2e4e5.png)

Frenzie:
@WaseemAlkurdi I resolved the duplicate as I proposed, by replacing the brand new duplicate with the old one that happened to be hidden from Weblate. :-)

In Weblate there are some automated checks to help with translation quality btw. (Of course you should ignore them if any of those are on purpose.)

https://hosted.weblate.org/projects/koreader/koreader/ar/

![Screenshot_2020-10-09_19-12-43](https://user-images.githubusercontent.com/202757/95612265-6c4da500-0a63-11eb-8ff1-125178c311ef.png)

WaseemAlkurdi:
> The ar_AA name is apparently the name from Transifex but Weblate just uses ar.

That's probably why, though I'm certain I've translated Arabic on Weblate before 😅️



> @WaseemAlkurdi I resolved the duplicate as I proposed, by replacing the brand new duplicate with the old one that happened to be hidden from Weblate. :-)
> 
> In Weblate there are some automated checks to help with translation quality btw. (Of course you should ignore them if any of those are on purpose.)
> 
[...]

PR #6870, opened 2020-11-12
Android bundle

pazos:
- adaptative icons
- minify + shrink
- remove unused assets to save a few KBs
- install apk by clicking on notification after download (doesn't apply to F-Droid)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6870)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -379,8 +379,7 @@ androidupdate: all
 		-xr!*NOTICE$ \
 		-xr!*README.md$ \
 		-xr!*koreader.icns$ \
-		-xr!*.git$ \
-		-xr!*.gitignore$
+		-xr'!.*'
 
 	# make the android APK
 	$(MAKE) -C $(ANDROID_LAUNCHER_DIR) $(if $(KODEBUG), debug, release) \
--- a/Makefile
+++ b/Makefile
@@ -358,14 +358,29 @@ androidupdate: all
 	# assets are compressed manually and stored inside the APK.
 	cd $(INSTALL_DIR)/koreader && 7z a -l -mx=9 -mfb=256 -mmt=on \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).zip * \
+		-xr!*cache$ \
+		-xr!*clipboard$ \
+		-xr!*data/dict$ \
+		-xr!*data/tessdata$ \
+		-xr!*history$ \
+		-xr!*l10n/templates$ \
+		-xr!*ota$ \
 		-xr!*resources/fonts* \
 		-xr!*resources/icons/src* \
+		-xr!*rocks/bin$ \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -358,14 +358,31 @@ androidupdate: all
 	# assets are compressed manually and stored inside the APK.
 	cd $(INSTALL_DIR)/koreader && 7z a -l -mx=9 -mfb=256 -mmt=on \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).zip * \
+		-xr!*cache$ \
+		-xr!*clipboard$ \
+		-xr!*data/dict$ \
+		-xr!*data/tessdata$ \
+		-xr!*history$ \
+		-xr!*l10n/.tx$ \
```
Frenzie, line 366: We can just remove that file. Switching between Weblate and Transifex isn't a simple binary thing anyway, see https://github.com/koreader/koreader/issues/6775.

Makefile
```diff
@@ -358,14 +358,31 @@ androidupdate: all
...
 		-xr!*resources/fonts* \
 		-xr!*resources/icons/src* \
+		-xr!*rocks/bin$ \
+		-xr!*rocks/lib/luarocks$ \
+		-xr!*screenshots$ \
 		-xr!*share/man* \
 		-xr!*spec$ \
+		-xr!*tools$ \
 		-xr!*COPYING$ \
+		-xr!*LICENSE$ \
```
Frenzie, line 378: I probably overlooked that before, but is that a good idea to skip?

pazos: There are a bunch of LICENSE files, all buried in /data/data/org.koreader/launcher/files, which are not available to end users.

Probably a generic license, like https://github.com/koreader/koreader/blob/master/platform/debian/copyright, on top of the APK is more useful than something buried into assets and extracted to private app dirs.

pazos: All LICENSE files re-included for now. It is a major PITA to aknowledge all licenses in single document.

Makefile
```diff
@@ -358,14 +358,29 @@ androidupdate: all
...
 		-xr!*COPYING$ \
+		-xr!*Makefile$ \
+		-xr!*NOTES.txt$ \
+		-xr!*NOTICE$ \
 		-xr!*README.md$ \
-		-xr!git$ \
-		-xr!gitiginore$ 
+		-xr!*koreader.icns$ \
+		-xr!*.git$ \
+		-xr!*.gitignore$
```
Frenzie, line 383: Or maybe something along the lines of `-xr'!.*'`, unless there's something hidden we do want?

Frenzie (approved):
lgtm

Makefile
pazos: yeah, makes sense

PR #6872, opened 2020-11-13
update Stausline immediately when toggling a bookmark

zwim:
When toggling a bookmark, the status line should be updated, so that the actual bookmark count (if enabled) is shown.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6872)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -185,6 +185,7 @@ function ReaderBookmark:onToggleBookmark()
     self.ui:handleEvent(Event:new("SetDogearVisibility",
                                   not self.view.dogear_visible))
     UIManager:setDirty(self.view.dialog, "ui")
+    UIManager:broadcastEvent(Event:new("BookmarksChanged"))
     return true
 end
 
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -185,7 +185,7 @@ function ReaderBookmark:onToggleBookmark()
     self.ui:handleEvent(Event:new("SetDogearVisibility",
                                   not self.view.dogear_visible))
     UIManager:setDirty(self.view.dialog, "ui")
-    UIManager:broadcastEvent(Event:new("BookmarksChanged"))
+    self.ui:handleEvent(Event:new("BookmarksChanged"))
     return true
 end
 
@@ -419,6 +419,7 @@ function ReaderBookmark:addBookmark(item)
         end
     end
     table.insert(self.bookmarks, _middle + direction, item)
+    self.ui:handleEvent(Event:new("BookmarksChanged"))
 end
 
 -- binary search of sorted bookmarks
@@ -461,7 +462,9 @@ function ReaderBookmark:removeBookmark(item)
         _middle = math.floor((_start + _end)/2)
         local v = self.bookmarks[_middle]
         if item.datetime == v.datetime and item.page == v.page then
-            return table.remove(self.bookmarks, _middle)
+            local retval = table.remove(self.bookmarks, _middle)
+            self.ui:handleEvent(Event:new("BookmarksChanged"))
+            return retval
         elseif self:isBookmarkInPageOrder(item, v) then
             _end = _middle - 1
         else
@@ -476,7 +479,9 @@ function ReaderBookmark:removeBookmark(item)
     for i=1, #self.bookmarks do
         local v = self.bookmarks[i]
         if item.datetime == v.datetime and item.page == v.page then
-            return table.remove(self.bookmarks, i)
+            local retval = table.remove(self.bookmarks, i)
+            self.ui:handleEvent(Event:new("BookmarksChanged"))
+            return retval
         end
     end
     logger.warn("removeBookmark: full scan search didn't find bookmark")
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -182,10 +182,10 @@ function ReaderBookmark:onToggleBookmark()
         pn_or_xp = self.ui.document:getXPointer()
     end
     self:toggleBookmark(pn_or_xp)
+    self.view.footer:onUpdateFooter(true)
     self.ui:handleEvent(Event:new("SetDogearVisibility",
                                   not self.view.dogear_visible))
     UIManager:setDirty(self.view.dialog, "ui")
-    self.ui:handleEvent(Event:new("BookmarksChanged"))
     return true
 end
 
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -2119,4 +2119,10 @@ function ReaderFooter:onScreenResize()
     self:resetLayout(true)
 end
 
+function ReaderFooter:onBookmarksChanged()
+    if self.settings.frontlight then
+        self:onUpdateFooter(true)
```
poire-z, line 2124: Why the dependancy on `settings.frontlight` ?

frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -185,6 +185,7 @@ function ReaderBookmark:onToggleBookmark()
     self.ui:handleEvent(Event:new("SetDogearVisibility",
                                   not self.view.dogear_visible))
     UIManager:setDirty(self.view.dialog, "ui")
+    UIManager:broadcastEvent(Event:new("BookmarksChanged"))
```
poire-z, line 188: Can't you re-use the already sent event SetDogearVisibility and rename your handler to `ReaderFooter:onSetDogearVisibility()` ?
(Also, any reason the existing one use self.ui:handleEvent and yours UIManager:broadcastEvent ?)

poire-z (commented):
Dunno if they are relevant, but there are already existing events when adding or removing highlights, as used in ReaderStatistics plugin:
```
ReaderStatistics:onAddHighlight()
ReaderStatistics:onDelHighlight()
ReaderStatistics:onAddNote()
```

frontend/apps/reader/modules/readerfooter.lua
zwim: bad copy paste, removed

frontend/apps/reader/modules/readerbookmark.lua
zwim: > Can't you re-use the already sent event SetDogearVisibility and rename your handler to `ReaderFooter:onSetDogearVisibility()` ?

No, that does not work.

> (Also, any reason the existing one use self.ui:handleEvent and yours UIManager:broadcastEvent ?)

No, I have changed it to `self.ui:handleEvent(Event:new("BookmarksChanged"))`

zwim:
> Dunno if they are relevant, but there are already existing events when adding or removing highlights, as used in ReaderStatistics plugin:
> 
> ```
> ReaderStatistics:onAddHighlight()
> ReaderStatistics:onDelHighlight()
> ReaderStatistics:onAddNote()
> ```

Ah, yes. I did not use the highlights so far. You are right, there will events too.

frontend/apps/reader/modules/readerbookmark.lua
poire-z: > No, that does not work.

I'm curious as to why/how it does not work. With your change, these should work the same:
```
self.ui:handleEvent(Event:new("SetDogearVisibility",...)
self.ui:handleEvent(Event:new("BookmarksChanged"))
```
Probably just because of the `return true` in readerdogear.lua's `ReaderDogear:onSetDogearVisibility(visible)` - that I guess you could remove so the event can propagate to other handlers.

zwim:
Sorry I have not seen your comment on dogear. But isn't the dogear event to specific, as I have added the events to the add/remove bookmarks too (that will cover the un/highlight too).

Frenzie:
> But isn't the dogear event to specific

Didn't look at the code, but changing the name is an option.

zwim:
> Didn't look at the code, but changing the name is an option. (And the better option if it's redundant.)

The status line should be updated if the number of bookmarks changes. This could happen on a dogear toggle, a highlight, an unhighlight and on taking a note.
So using isn't the event-name `onSetDogearVisibility` misleading on the other cases, where no dogear is showed?

poire-z:
There are 2 options:
- adding a new event, sent in multiple places, with a new handler in ReaderFooter - what you did
- re-using current events, _already sent_ in multiple places, with multiple new handlers in ReaderFooter

@Frenzie suggested you could rename `onSetDogearVisibility` to `onToggleBookmark` - and it would sound more generic.

My concern is that we already have events sent when highlights are added and removed:
https://github.com/koreader/koreader/blob/1468646aa5d6cbfca398586fa4b9f24338e54043/frontend/apps/reader/modules/readerhighlight.lua#L1284-L1285
https://github.com/koreader/koreader/blob/1468646aa5d6cbfca398586fa4b9f24338e54043/frontend/apps/reader/modules/readerhighlight.lua#L1400-L1401
and you're adding a new one - in the other part of the code (ReaderBookmark) that also deal with highlights (our bookmark/highlight code is messy...).

A 3rd option, given that your onBookmarksChanged just do onUpdateFooter, would be to just send the UpdateFooter event around all these - or use directly `self.view.footer:onUpdateFooter()` which we already do in multiple ReaderSomething modules (dict, paging, rolling, toc, pagemap).

I really don't know which option is best :)

Frenzie:
That's what I'm saying, any name is just an implementation detail. The event was historically (6+ years ago) implemented for the dog ear marker. If it makes more sense as a generic event in 2020, you can give it a generic name.

More tailored events can definitely make more sense than one generic event with a million options. I'm not saying that's what you should do; I'm just talking in the abstract here. But the name can never be an objection. ^_^

zwim:
> A 3rd option, given that your onBookmarksChanged just do onUpdateFooter, would be to just send the UpdateFooter event around all these - or use directly `self.view.footer:onUpdateFooter()` which we already do in multiple ReaderSomething modules (dict, paging, rolling, toc, pagemap).
> 

If I use the `BookmarksChanged` event, the status line gets updated immediately on highlight mark/unmark dialog.
But if I use the `UpdateFooter` event or `self.view.footer:onUpdateFooter()` the status line gets updated after closing the mark/unmark dialog.


Reffering to you concern about the events in ReaderHighlight: Isn't the count in the status line logically connected to the number of bookmarks? The 'backend' of dealing with bookmarks is readerbookmark. readerhighlight does a lot of things but should not deal with the interna of readerbookmark.

poire-z:
> or self.view.footer:onUpdateFooter() the status line gets updated after closing the mark/unmark dialog.

May be because you then don't give the argument `force_repaint=true` to `onUpdateFooter()` (which you do in your `onBookmarksChanged()`).

> Ergo the event should be sent fom readerbookmark.

OK.
May be just to not have this ugly `retval` in `ReaderBookmark:removeBookmark()`, and as no code currently expects to get the removed bookmark, just get rid of it and just `return`. 
And in onToggleBookmark(), may be move your new event nearer to where the toggleBookmark is done - at least above the setDirty.

zwim:
> May be because you then don't give the argument `force_repaint=true` to `onUpdateFooter()` (which you do in your `onBookmarksChanged()`).

Ups, I need new glasses.

poire-z (approved):
Simpler is better :)

PR #6878, opened 2020-11-14
Ensure minimum height for the in-fill in progress bar

Jellby:
With some screen sizes, the footer progress bar (in thick style) may be empty, because margins and borders take the full height. This should ensure that does not happen, by reserving at least 1 pixel for the actual progress bar.

Left: before the PR. Right: after
Screenshot_20201114_115606 Screenshot_20201114_115723
with `./kodev run -h=720 -w=1024`, for sizes 5 to 9 (which get scaled to 6 to 10, and the default border and margin are scaled to 2)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6878)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -660,7 +660,7 @@ function ReaderFooter:resetLayout(force_reset)
     else
         bar_height = self.settings.progress_style_thick_height or PROGRESS_BAR_STYLE_THICK_DEFAULT_HEIGHT
     end
-    self.progress_bar.height = Screen:scaleBySize(bar_height)
+    self.progress_bar:setHeight(bar_height)
 
     self.horizontal_group:resetLayout()
     self.footer_positioner.dimen.w = new_screen_width
@@ -1820,7 +1820,7 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
     else
         bar_height = self.settings.progress_style_thick_height or PROGRESS_BAR_STYLE_THICK_DEFAULT_HEIGHT
     end
-    self.progress_bar.height = Screen:scaleBySize(bar_height)
+    self.progress_bar:setHeight(bar_height)
 
     if self.separator_line then
         self.separator_line.dimen.w = self._saved_screen_width - 2 * self.horizontal_margin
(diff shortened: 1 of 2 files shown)


frontend/ui/widget/progresswidget.lua
```diff
@@ -50,6 +52,8 @@ local ProgressWidget = Widget:new{
     fill_from_right = false,
     allow_mirroring = true,
     _mirroredUI = BD.mirroredUILayout(),
+    margin_v_orig = nil,
+    bordersize_orig = nil,
```
Frenzie, line 56: These seem to be for internal purposes only.

Frenzie:
Sounds good. "Some screen sizes" is very vague though. What kind of screen sizes? :-)

frontend/ui/widget/progresswidget.lua
```diff
@@ -15,6 +15,8 @@ Configurable attributes:
  * ticks (list)  -- default to nil, use this if you want to insert markers
  * tick_width
  * last  -- maximum tick, used with ticks
+ * margin_v_orig   -- margin_v before adjusting
+ * bordersize_orig -- bordersize before adjusting
```
poire-z, line 19: If there are just internal variables, and they shouldn't ever need to be provided by caller, may be don't document them here.
And name them `_orig_margin_v`, with a leading underscore as a kind of convention these are private.

Jellby:
> Sounds good. "Some screen sizes" is very vague though. What kind of screen sizes? :-)

I don't know, whatever causes `Screen:scaleBySize(1) >1` and `Screen:scaleBySize(5) < 9`, I guess. I gave an example in the top message: 720×1024

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -660,7 +660,7 @@ function ReaderFooter:resetLayout(force_reset)
     else
         bar_height = self.settings.progress_style_thick_height or PROGRESS_BAR_STYLE_THICK_DEFAULT_HEIGHT
     end
-    self.progress_bar.height = Screen:scaleBySize(bar_height)
+    self.progress_bar:setHeight(bar_height)
```
NiLuJe, line 663: There's another instance of this @ L1836

(Which I'm not quite sure is actually necessary at all, because it has nothing whatsoever to do with what that function's supposed to do...)

Jellby: I thought I had searched for other instances, obviously I missed at least one.

PR #6880, opened 2020-11-14
[Kobo] Don't attempt to restart Nickel when asking for a reboot/shutdown

NiLuJe:
I'm not quite sure why I never actually noticed this brainfart before, but it's been stupid like this since basically ever.

(It was made obvious by the inittab changes in FW 4.25, which does a few more things on halt, which gave more time for the on-animator call from our nickel restart script to pop up...).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6880)

### DIFF
--- a/platform/kobo/koreader.sh
+++ b/platform/kobo/koreader.sh
@@ -266,6 +266,7 @@ CRASH_PREV_TS=0
 # List of supported special return codes
 KO_RC_RESTART=85
 KO_RC_USBMS=86
+KO_RC_HALT=88
 # Because we *want* an initial fbdepth pass ;).
 RETURN_VALUE=${KO_RC_RESTART}
 while [ ${RETURN_VALUE} -ne 0 ]; do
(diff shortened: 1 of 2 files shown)


platform/kobo/koreader.sh
```diff
@@ -266,6 +266,7 @@ CRASH_PREV_TS=0
 # List of supported special return codes
 KO_RC_RESTART=85
 KO_RC_USBMS=86
+KO_RC_HALT=88
```
poire-z, line 269: Wikipedia states that [87 (eighty-seven)](https://en.wikipedia.org/wiki/87_(number)) is the natural number following 86 and preceding 88, and I'm inclined to trust it!
So, for what coming up new feature have you reserved yourself 87 for ? :)

NiLuJe: Cervantes uses 87 to restart to the stock App ;).

Frenzie (approved):

PR #6881, opened 2020-11-14
android: switch compression to lzma2

pazos:
requires https://github.com/koreader/android-luajit-launcher/pull/264

I will upload slightly modified versions of 2020.11 with LZMA/ZIP that show a toast with the milliseconds spent uncompressing, for users to benchmark.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6881)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -356,8 +356,8 @@ androidupdate: all
 	-rm $(INSTALL_DIR)/koreader/libs/libluajit.so
 
 	# assets are compressed manually and stored inside the APK.
-	cd $(INSTALL_DIR)/koreader && 7z a -l -mx=9 -mfb=256 -mmt=on \
-		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).zip * \
+	cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=9 \
+		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).7z * \
 		-xr!*cache$ \
 		-xr!*clipboard$ \
 		-xr!*data/dict$ \
--- a/Makefile
+++ b/Makefile
@@ -355,6 +355,10 @@ androidupdate: all
 	# in runtime luajit-launcher's libluajit.so will be loaded
 	-rm $(INSTALL_DIR)/koreader/libs/libluajit.so
 
+	# shared libraries are stored as raw assets
+	rm -rf $(ANDROID_LAUNCHER_DIR)assets/libs
+	cp -pR $(INSTALL_DIR)/koreader/libs $(ANDROID_LAUNCHER_DIR)/assets
+
 	# assets are compressed manually and stored inside the APK.
 	cd $(INSTALL_DIR)/koreader && 7z a -l -m0=lzma2 -mx=9 \
 		../../$(ANDROID_LAUNCHER_DIR)/assets/module/koreader-$(VERSION).7z * \
@@ -364,6 +368,7 @@ androidupdate: all
 		-xr!*data/tessdata$ \
 		-xr!*history$ \
 		-xr!*l10n/templates$ \
+		-xr!*libs$ \
 		-xr!*ota$ \
 		-xr!*resources/fonts* \
 		-xr!*resources/icons/src* \


Frenzie (approved):
Should be okay, you sure you want it in 2020.11 though?

pazos:
> Should be okay, you sure you want it in 2020.11 though?

No, no, no. 2020.12. I was considering the 2021.01 milestone, but looks kind of a joke.

pazos:
Test APKs:

Both are built with the frontend in https://github.com/koreader/koreader/commit/510ea5963cfffbc97aa527e6ae5de2624ff8da20, slightly modified to print time spent uncompressing assets (both are compressed at maximum levels):

device-2020-11-15-131109

Original APK, with zipped assets: https://www.dropbox.com/s/fuya68io26a7dvc/koreader-zip-benchmark.apk?dl=0 (37MB)
New APK, with 7z assets: https://www.dropbox.com/s/5tte7v9vvt2sqpm/koreader-7z-benchmark.apk?dl=0 (27MB)

The new APK is slightly slower uncompressing (less than 5% on the devices I tested, including legacy emulators with little RAM)

Pinging people with access to slow devices, like @Cellaris, @Norbi24, @silberzwiebel and @zwim. I want to check if the ratio still stands in slower devices. Both apps should work. If the 7z one crashes and the other does not then we can forget all of this :)

Steps:

1. uninstall current KOReader APK
2. install the zip one ->  see how much time it takes to uncompress
3. install the 7z with or without uninstalling the zip one. -> see how much time it takes to uncompress
4. uninstall and reinstall one of the official/f-droid releases.

Thanks in advance

zwim:
@pazos: Measured on a Tolino Epos 2:
The zip-version takes 9609ms, the 7z-version 11077ms -> This is a 15% overhead in decompression time, but a 27% improvement in size. :+1: 
I would go for it (although I haven't tested them).

norbi24:
Hola Pazos,

  I tried it on Tolino HD4 with Android 4.4.2 - both versions worked, the
zip showed 8987ms, and 7-zip 10030ms

Saludos, Norbert

pazos:
The bottleneck is IO in seeking assets, not cpu. So reducing compression won't make things faster, but slower because it increments the size of the assets to seek.

Changing from `-mx=9` to `-mx=1` makes things slower:

- Nexus 4 emulator API 25 -> 2296 | 3012
- Moto G5S API28 -> 4086 | 4893
- Samsung S3 Neo API28 (Lineage) -> 6230 | 7159

Probably the slower devices we target are tolinos with internal SD card. Even very old Boyue/Onyx clones running 4.0 should have a faster IO than crappy Class 4 sdcards.

I don't know what's your opinion @Frenzie @NiLuJe @poire-z: size over uncompress speed? uncompress speed over size?.

poire-z:
> size over uncompress speed? uncompress speed over size?

(I haven't followed the whole discussion, but) I'd say smaller APK size is better (just because it will make us looks like tidy people :). I wouldn't really care if the installation/update I do once a week takes 10s or 20s.

Frenzie:
Out of the listed values the Nexus 4 (emulator) is the only type of value where I think it subjectively matters. 4 or 5, 5 or 6, 9 or 10… that's all just slow.

Frenzie:
Btw, you could consider experimenting with https://superuser.com/a/337087 on .png and .so and the like?

pazos:
@Frenzie: nope. Resources are already optimized and shared libraries cannot be shrinked by compressing them again.

OTOH aapt makes a pretty impressive work with shared libraries and it will reduce update time as they're extracted at install time, not at runtime.

But:
- we need to rebuilt shared objects from `.so.*` to `.so`
- certain load paths need to be changed because they're hardcoded (see https://github.com/koreader/koreader/search?q=ffi.load+libs%2F&type=)

Too much work for my taste. It will be required at some point if android restricts dlopen from internal partition.

Frenzie:
> Resources are already optimized and shared libraries cannot be shrinked by compressing them again.

I know they can't; my point is whether any CPU time is wasted on compressing/decompressing them at present, i.e., is there time to be gained here?

NiLuJe:
> 
> 
> > size over uncompress speed? uncompress speed over size?
> 
> (I haven't followed the whole discussion, but) I'd say smaller APK size is better (just because it will make us looks like tidy people :). I wouldn't really care if the installation/update I do once a week takes 10s or 20s.

:+1:

pazos:
> > Resources are already optimized and shared libraries cannot be shrinked by compressing them again.
> 
> I know they can't; my point is whether any CPU time is wasted on compressing/decompressing them at present, i.e., is there time to be gained here?

I will try to benchmark what happens storing blobs as raw assets.

pazos:
> I will try to benchmark what happens storing blobs as raw assets.

Putting libs in the assets store without compressing is indeed faster: 186? vs 229?. The APK grows from 30 to 33 MB in size.

Frenzie:
Sounds worthwhile to me. ^_^ Also don't forget about the PNG files and fonts. :-P

pazos:
aapt won't optimize PNGs. The fonts are a pain to move there, as they're nested in folders, so I would need to check for each asset in assets if the asset is a dir, check if the output dir exists and create it if not.

Just to showcase what involves to move the libs outside the zip/7z container:

```diff
diff --git a/.gitignore b/.gitignore
index 7a60b34..c23f787 100644
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ bin
 build
 app/.*
 app/build
+assets/libs
 assets/module
 jni/logger.h
 jni/luajit/build
diff --git a/app/src/org/koreader/launcher/Assets.kt b/app/src/org/koreader/launcher/Assets.kt
index 7ec2427..7c3ece5 100644
--- a/app/src/org/koreader/launcher/Assets.kt
+++ b/app/src/org/koreader/launcher/Assets.kt
@@ -27,6 +27,68 @@ class Assets() {
         }
     }
 
+    fun copyLibs(context: Context): Boolean {
+        val assetManager = context.assets
+        val libsDir = File(context.filesDir.absolutePath + "/libs")
+        if (!libsDir.exists()) {
+            libsDir.mkdir()
+        }
+
+        val libsPath = libsDir.absolutePath
+        try {
+            val assets = assetManager.list("libs")
+            return if (assets != null) {
+                for (asset in assets) {
+                    val file = File(libsPath, asset)
+                    val input = assetManager.open("libs/$asset")
+                    val output = FileOutputStream(file)
+                    copyFile(input, output)
+                    input.close()
+                    output.flush()
+                    output.close()
+                }
+                true
+            } else {
+                Logger.i("No libraries to copy")
+                true
+            }
+        } catch (e: IOException) {
+            Logger.e(TAG, "error copying libraries: $e")
+            return false
+        }
+    }
+
     /* copy raw assets from the assets module */
     fun copyRawAssets(context: Context): Boolean {
         val assetManager = context.assets
diff --git a/app/src/org/koreader/launcher/MainActivity.kt b/app/src/org/koreader/launcher/MainActivity.kt
index d595e21..5a3cc23 100644
--- a/app/src/org/koreader/launcher/MainActivity.kt
+++ b/app/src/org/koreader/launcher/MainActivity.kt
@@ -298,7 +298,11 @@ class MainActivity : NativeActivity(), JNILuaInterface,
                     showToast("update installed in " + elapsedTime / 1000000 + " milliseconds", true)
                     dismissProgress() // dismiss progress dialog
                 }
-                if (ok) 1 else 0
+                if (!ok) {
+                    0
+                } else {
+                    if (assets.copyLibs(this)) 1 else 0
+                }
             } else {
                 // check if the app has other, non-zipped, raw assets
                 Logger.i("Zip file not found, trying raw assets...")
```

pazos:
Of course all of this can be simplified if we get rid of the useless code that we keep in luajit-launcher to make the impression that the repo is useful outside KOReader :)

In that case we can have a generic function to extract assets that iterates the entire asset store extracting things in their relative paths and, if a file ends in 7z) uncompress it in the root of the internal storage.

pazos:
Yup, and with fonts outsize the 7z the resulting APK is almost the same size as we have now :P

Frenzie:
Fonts compress really well, my bad. I was probably thinking of WOFF or something.

Frenzie:
> aapt won't optimize PNGs

We've optimized them ourselves, haven't we? My point is the same as with the libraries, PNGs don't compress since they're already compressed. ;-)

NiLuJe:
Yeah, the PNGs are dithered down to the eInk palette and optimized ;).

silberzwiebel:
Does it still help if I follow [these steps from @pazos](https://github.com/koreader/koreader/pull/6881#issuecomment-727614593) on my inkbook prime for testing? I'd need some days/weeks, probably. Or is your discussion at a different stage already?

pazos:
@silberzwiebel: still welcome. You have plenty of time to do that on the next weeks, there's no hurry :)

pazos:
> We've optimized them ourselves, haven't we? My point is the same as with the libraries, PNGs don't compress since they're already compressed. ;-)

Just tried. No significative differences in speed (still in ~1860) or size (still in 30/33MB for arm/x86), but the code needed makes the whole thing **very, very messy**, I guess most of the time we save uncompressing the few KBs is spent in listing assets for each folder and create the directories recursively.

So my proposal:

- libs outside the 7z
- all the rest inside the 7z

Frenzie:
Probably also because the PNGs don't take up much space regardless.

pazos:
@zwim: could you test again the 7z APK?. I uploaded a new version with the thanges requested by @Frenzie. It is a bit larger (27->30MB) but should update a bit faster

Thanks!

zwim:
@pazos: Interesting result:
Zip need 10100ms and 
7z needs 8095ms.
I am sure, I have not swapped the results.

Frenzie:
Might be an unfair comparison if the same wasting time on .so thing
applies. (But the space saving is worth an extra second or two when it's
8-10 regardless.)

pazos:
> Might be an unfair comparison if the same wasting time on .so thing applies. (But the space saving is worth an extra second or two when it's 8-10 regardless.)

Huh?. I think I didn't understand. The comparation happens between the zip (with shared libraries inside) and 7z (with shared libraries outside). It seems that 7z + libs outside is the best of both worlds (7MB less, 2 seconds less) on Tolinos.

Frenzie:
> The comparation happens between the zip (with shared libraries inside) and 7z (with shared libraries outside).

Exactly, that's an unfair comparison. You're handicapping the zip file so that it will be slower.

There are at least three different scenarios per compression method:

1. Everything compressed.
2. Tailored in archive.
3. Tailored split.

Currently you've only tested 1 for zip, as well as 1 and 3 for 7z.

Ideally 2 and 3 will behave the same, but that's not a given.

Now I don't particularly care what the results are since the ~33 vs ~37 MB will remain which I think it work 10-20 % more time, but you can't claim it's the best of both worlds unless you do a fair comparison.

pazos:
Oh, I understand. I'm not interested in fair comparisons :smile: . I was comparing between what we have now and how things end with this PR.

Anyways I can upload a test version of the zip file. It shouldn't change time too much as a ZipInputStream should behave almost like a inputStream.

There're other things to optimize, like crunching icons/drawables, which seem to save about 300kb more. IIRC the option to disable crunching was added to allow reproducible builds, but wasn't used in practice.

pazos:
Lets see how it behaves in the wild :+1:

silberzwiebel:
> Lets see how it behaves in the wild +1

Fantastic, so I don't need to test anymore :)

Issue #6876, opened 2020-11-14
Coverimage - Some cover images are cut off due to different aspect ratio to the device.


sndwichm4n:
* KOReader version: 2020.10.1-44
* Device: Poke 2

#### Issue
Some cover images are cut off due to different aspect ratio to the device.

The device has a resolution of 1448 x 1072 ratio 1,350. If a book cover has a resolution of 1186 x 762 ratio 1,556, the image is scaled to the screen size of the device by the screensaver. So the upper and lower side is cut off a little bit.

#### Steps to reproduce
Install Koreader with "coverimage" on a Poke 2 and open a book. Put the device to sleep.

#### Solution
Set the book cover to the same aspect ratio as the device: Depending on the margins insert in black or white.

sndwichm4n:
I love the plugin very much! :)

Frenzie:
https://github.com/koreader/koreader/blob/6cc0ab8dc9da1549f6a6d1abdb8fe190b302faba/frontend/ui/widget/imagewidget.lua#L91-L96

Galunid:
@sndwichm4n Does it "just" work on Poke Pro? I mean, is the cover displayed as screensaver (even if it has wrong dimensions)? If that's the case then it's pretty surprising

sndwichm4n:
@Galunid It is a Poke 2 no Poke pro, the poke 2 is new bought. I have written something wrong out of habit.

zwim:
@sandwichm4n: Until now, coverimage does no scaling. My Tolino is smart enough to scale the image itself.

As the Poke2 does not scale correctly, we will have to help him.

Could you please provide some informations:
1.) Which filename did you enter to store the cover image?
2.) Could you upload the created cover-image somewhere (dropbox)?
3.) Could you upload an image (photo) of the Poke2 with the cut off image?

sndwichm4n:
No problem:

Filename: /sdcard/Pictures/cover.png
Fallback: /sdcard/Pictures/Fallback.png

https://imgur.com/a/4DeOaxG

sndwichm4n:
> @sandwichm4n: Until now, coverimage does no scaling. My Tolino is smart enough to scale the image itself.
> 
> As the Poke2 does not scale correctly, we will have to help him.

I do not think it is about scaling. On the left and right sides 58 pixels each must be added. To get to the ratio of the Poke2 screen.

((1186/1.35)-762)/2 = 58.259

zwim:
[...]

PR #6884, opened 2020-11-14
Scale cover image and add a background where necessary to better fit the screen

zwim:
This should fix https://github.com/koreader/koreader/issues/6876

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6884)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -4,9 +4,11 @@ if not Device.isAndroid() and not Device.isEmulator() then
     return { disabled = true }
 end
 
+local Blitbuffer = require("ffi/blitbuffer")
 local InfoMessage = require("ui/widget/infomessage")
 local UIManager = require("ui/uimanager")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
+local RenderImage = require("ui/renderimage")
 local ffiutil = require("ffi/util")
 local lfs = require("libs/libkoreader-lfs")
 local logger = require("logger")
@@ -67,9 +69,33 @@ end
 
 function CoverImage:createCoverImage(doc_settings)
    if self.enabled and not doc_settings:readSetting("exclude_cover_image") == true then
-        local image = self.ui.document:getCoverPageImage()
-        if image then
+        local cover_image = self.ui.document:getCoverPageImage()
+        if cover_image then
+            local s_w, s_h = Device.screen:getWidth(), Device.screen:getHeight()
+            local i_w, i_h = cover_image:getWidth(), cover_image:getHeight()
+            local scale_factor = math.min(s_w / i_w, s_h / i_h)
+
+            if scale_factor == 1 then
+                cover_image:writePNG(self.cover_image_path, false)
+                return
+            end
+
+            local scaled_w, scaled_h = i_w * scale_factor, i_h * scale_factor
+            local cover_image = RenderImage:scaleBlitBuffer(cover_image, scaled_w, scaled_h)
+
+            -- new buffer with screen dimensions,
+            local image = Blitbuffer.new( s_w, s_h, cover_image:getType() )
+            image:fill( Blitbuffer.COLOR_BLACK )
+
+            -- copy scaled image to buffer
+            if s_w > scaled_w then -- move right
+                image:blitFrom(cover_image, math.floor( (s_w - scaled_w) / 2 ), 0, 0, 0, scaled_w, scaled_h)
+            else -- move down
+                image:blitFrom(cover_image, 0, math.floor( (s_h - scaled_h) / 2 ), 0, 0, scaled_w, scaled_h)
+            end
             image:writePNG(self.cover_image_path, false)
+            image:free()
+            cover_image:free()
             logger.dbg("CoverImage: image written to " .. self.cover_image_path)
         end
     end
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -232,7 +232,7 @@ function CoverImage:addToMainMenu(menu_items)
                 text = _("Scale book cover"),
                 help_text = _("If disabled, the cover will have to be scaled by the FW.\nDisabling is the fastest method, but doesn't work on all devices."),
                 checked_func = function()
-                    return self.ui and self.ui.doc_settings and self.cover_image_background ~= "none"
+                    return self.cover_image_background ~= "none"
                 end,
                 enabled_func = function()
                     return self.enabled
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -38,7 +38,7 @@ local CoverImage = WidgetContainer:new{
 
 function CoverImage:init()
     self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "cover.png"
-    self.cover_image_background = G_reader_settings:readSetting("cover_image_background") or "none"
+    self.cover_image_background = G_reader_settings:readSetting("cover_image_background") or "black"
     self.cover_image_fallback_path = G_reader_settings:readSetting("cover_image_fallback_path") or "cover_fallback.png"
     self.enabled = G_reader_settings:isTrue("cover_image_enabled")
     self.fallback = G_reader_settings:isTrue("cover_image_fallback")
@@ -82,14 +82,12 @@ function CoverImage:createCoverImage(doc_settings)
                 return
             end
 
-            local scaled_w, scaled_h = i_w * scale_factor, i_h * scale_factor
+            local scaled_w, scaled_h = math.floor(i_w * scale_factor), math.floor(i_h * scale_factor)
             local cover_image = RenderImage:scaleBlitBuffer(cover_image, scaled_w, scaled_h)
 
             -- new buffer with screen dimensions,
-            local image = Blitbuffer.new( s_w, s_h, cover_image:getType() )
-            if self.cover_image_background == "black" then
---                image:fill(Blitbuffer.COLOR_BLACK) -- not necessary, as BB are zero-initialized
-            elseif self.cover_image_background == "white" then
+            local image = Blitbuffer.new( s_w, s_h, cover_image:getType() ) -- new buffer, filled with black
+            if self.cover_image_background == "white" then
                 image:fill(Blitbuffer.COLOR_WHITE)
             elseif self.cover_image_background == "gray" then
                 image:fill(Blitbuffer.COLOR_GRAY)
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -38,10 +38,12 @@ local CoverImage = WidgetContainer:new{
 
 function CoverImage:init()
     self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "cover.png"
+    self.cover_image_background = G_reader_settings:readSetting("cover_image_background") or "none"
     self.cover_image_fallback_path = G_reader_settings:readSetting("cover_image_fallback_path") or "cover_fallback.png"
     self.enabled = G_reader_settings:isTrue("cover_image_enabled")
     self.fallback = G_reader_settings:isTrue("cover_image_fallback")
     self.ui.menu:registerToMainMenu(self)
+
 end
 
 function CoverImage:_enabled()
@@ -75,7 +77,7 @@ function CoverImage:createCoverImage(doc_settings)
             local i_w, i_h = cover_image:getWidth(), cover_image:getHeight()
             local scale_factor = math.min(s_w / i_w, s_h / i_h)
 
-            if scale_factor == 1 then
+            if scale_factor == 1 or self.cover_image_background == "none" then
                 cover_image:writePNG(self.cover_image_path, false)
                 return
             end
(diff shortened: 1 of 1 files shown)


poire-z:
Given that these options for the regular screensaver are saved in G_reader_settings, and so easily accessible from your plugin, you could re-use their values:
https://github.com/koreader/koreader/blob/6cc0ab8dc9da1549f6a6d1abdb8fe190b302faba/frontend/ui/screensaver.lua#L168-L178
especially the white background (so people with a white plastic device can get white stripes).

zwim:
Yes, we could add a menu-entry "Scale book cover" with the entries:
[ ] Black background
[ ] White background
[ ] No background

poire-z:
Why not if you're up for that (my suggestion was just to avoid that menu work :)

plugins/coverimage.koplugin/main.lua
```diff
@@ -67,9 +71,40 @@ end
...
+            end
+
+            local scaled_w, scaled_h = i_w * scale_factor, i_h * scale_factor
+            local cover_image = RenderImage:scaleBlitBuffer(cover_image, scaled_w, scaled_h)
+
+            -- new buffer with screen dimensions,
+            local image = Blitbuffer.new( s_w, s_h, cover_image:getType() )
+            logger.err("xxxxxxxxxxxxxxxxxxxxxxxx " .. self.cover_image_background)
+            if self.cover_image_background == "black" then
+                image:fill(Blitbuffer.COLOR_BLACK)
```
NiLuJe, line 92: BBs are zero-initialized, so the explicit zero-fill is technically redundant.

(I'd still keep it just to avoid aking the code more puzzling in that case, this is just me throwing in a "fun fact" here ^^).

plugins/coverimage.koplugin/main.lua
```diff
@@ -67,9 +71,39 @@ end
...
+            local i_w, i_h = cover_image:getWidth(), cover_image:getHeight()
+            local scale_factor = math.min(s_w / i_w, s_h / i_h)
+
+            if self.cover_image_background == "none" or scale_factor == 1 then
+                cover_image:writePNG(self.cover_image_path, false)
+                return
+            end
+
+            local scaled_w, scaled_h = i_w * scale_factor, i_h * scale_factor
+            local cover_image = RenderImage:scaleBlitBuffer(cover_image, scaled_w, scaled_h)
```
Galunid, line 86: Do those work properly? In https://github.com/koreader/koreader/pull/6709 using "float" in blitbuffer caused image to be "projected".

Galunid (commented):
Other than those two comments it looks good to me.

plugins/coverimage.koplugin/main.lua
```diff
@@ -36,10 +38,12 @@ local CoverImage = WidgetContainer:new{
 
 function CoverImage:init()
     self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or "cover.png"
+    self.cover_image_background = G_reader_settings:readSetting("cover_image_background") or "none"
```
Galunid, line 41: It should probably be enabled by default. This shouldn't take too much time and it would ensure it works out of the box on all devices.

plugins/coverimage.koplugin/main.lua
```diff
@@ -67,9 +71,39 @@ end
...
+                return
+            end
+
+            local scaled_w, scaled_h = i_w * scale_factor, i_h * scale_factor
+            local cover_image = RenderImage:scaleBlitBuffer(cover_image, scaled_w, scaled_h)
+
+            -- new buffer with screen dimensions,
+            local image = Blitbuffer.new( s_w, s_h, cover_image:getType() )
+            if self.cover_image_background == "black" then
+--                image:fill(Blitbuffer.COLOR_BLACK) -- not necessary, as BB are zero-initialized
```
poire-z, line 91: Please keep the `--` indented at the code level.
Also, no CI check seems to have run, but it will probably complain about the empty `then elseif`. When that's needed, one have to make it non empty with:
```lua
if something then
elseif
    other stuff
end
```
But I would probably leave the image.fill even if it's redundant. Or just chec for white and gray, and mention as a comment that black doesn't need it.

plugins/coverimage.koplugin/main.lua
zwim: For me it seems, that the images are clear, but a `math.floor` can not hurt.

plugins/coverimage.koplugin/main.lua
zwim: OK, set it to black as this is the fastest method and it fits best to my Tolino

plugins/coverimage.koplugin/main.lua
zwim: Thanks for the hint. I will mention the behavior  on `Blitbuffer.new`.

Galunid (approved):

zwim:
Wouldn't it be better to save the background color globally and not for each doc?
The background color was introduced for people with white plastic cover, this does not depend on the book, but on the device.

poire-z:
> Wouldn't it be better to save the background color globally and not for each doc?

Indeed.
But you read it from G_reader_settings on line 41.
And you save it in the doc_settings in the menu, which looks like an error :)

zwim:
> But you read it from G_reader_settings on line 41.
> And you save it in the doc_settings in the menu, which looks like an error :)

Yes, I have mixed them up. But now it should be corrected, only G_reader_settings is used now.

plugins/coverimage.koplugin/main.lua
```diff
@@ -193,6 +227,70 @@ function CoverImage:addToMainMenu(menu_items)
...
+                        text = _("White background"),
+                        checked_func = function()
+                            return self.cover_image_background == "white"
+                        end,
+                        callback = function()
+                            if self.cover_image_background == "white" then
+                                self.cover_image_background = "none"
+                            else
+                                self.cover_image_background = "white"
+                            end
```
Galunid, line 268: It could be rewritten as something like
```lua
self.cover_image_background = self.cover_image_background == "white" and "none" or "white"
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -193,6 +227,70 @@ function CoverImage:addToMainMenu(menu_items)
...
+                        text = _("Gray background"),
+                        checked_func = function()
+                            return self.cover_image_background == "gray"
+                        end,
+                        callback = function()
+                            if self.cover_image_background == "gray" then
+                                self.cover_image_background = "none"
+                            else
+                                self.cover_image_background = "gray"
+                            end
```
Galunid, line 285: Same as above

plugins/coverimage.koplugin/main.lua
```diff
@@ -193,6 +227,70 @@ function CoverImage:addToMainMenu(menu_items)
                     end
                 end,
             },
+            -- menu entry: scale book cover
+            {
+                text = _("Scale book cover"),
+                help_text = _("If disabled, the cover will have to be scaled by the FW.\nDisabling is the fastest method, but doesn't work on all devices."),
+                checked_func = function()
+                    return self.ui and self.ui.doc_settings and self.cover_image_background ~= "none"
```
Galunid, line 235: No need to check for `self.ui` and `self.ui.doc_settings` here now :)

zwim: Right!

Galunid:
Off topic, why isn't there CI on this PR?

Frenzie:
I think CircleCI quit running for people who aren't on it or something. It's quite annoying. >_<

sndwichm4n:
> CircleCI

What is this?

Frenzie:
See https://app.circleci.com/pipelines/github/koreader/koreader/1429/workflows/6e51085b-43cd-4958-ac16-f765ed357056/jobs/5738 for a sample run. It runs a variety of automated tests we set up to ensure certain things don't break and/or meet a certain level of quality.

sndwichm4n:
Very interesting, thanks.
> Good Job
> 498 tests are passing. 
:D

poire-z:
> I think CircleCI quit running for people who aren't on it or something. It's quite annoying

Indeed :( We (members, any logged in user?) can still re-run "unauthorized" builds - which I just did for our opened PRs that were missing it.

plugins/coverimage.koplugin/main.lua
```diff
@@ -193,6 +228,58 @@ function CoverImage:addToMainMenu(menu_items)
                     end
                 end,
             },
+            -- menu entry: scale book cover
+            {
+                text = _("Scale book cover"),
+                help_text = _("If disabled, the cover will have to be scaled by the FW.\nDisabling is the fastest method, but doesn't work on all devices."),
```
poire-z, line 234: Haven't seen how this is presented, but it seems a bit strange:
`text = _("Scale book cover"),` hints it might be a toggle, or some scaling options
`sub_item_table ` presents "black|white|grey background" options, which seem totally unrelated to scaling :) 
`help_text` is then not helpfull, mentionning "disabling" which one can't find anywhere :/

zwim: Should I rename `Scale book cover` to `Scale book cover and add background` and
the help text to:   _("If a background color is selected, the image will be scaled with its original aspect ratio and a solid backround to fit the device screen.\nIf no background color is selected, the cover is saved in the original aspect ratio and it has to be scaled by the FW.\nThe latter is the fastest method, but doesn't work on all devices.")?

poire-z: We use the word "Stretch" in the regular screensaver - it sounds clearer than "Scale" in expressing that the image will be distorted :) 
May be you should have 4 items instead of 3, mutually exclusive: `[ ] Strech to fit screen`, with always one checked.
Your main menu title could then be `Stretching and background`.

zwim:
@pazos: I have change the sub menu the text. 
When pushing the commit 43651c0 I got the message that I should do a `git pull` first, and now I have a strange commit 29dcc6b in here. If this does any harm, how can I remove it?

plugins/coverimage.koplugin/main.lua
```diff
@@ -270,7 +267,20 @@ function CoverImage:addToMainMenu(menu_items)
...
+                            G_reader_settings:saveSetting("cover_image_background", self.cover_image_background)
+                            if self.enabled then
+                                self:createCoverImage(self.ui.doc_settings)
+                            end
+                        end
+                    },
+                    {
+                        text = _("Stretch to fit screen"),
+                        checked_func = function()
+                            return self.cover_image_background == "none"
```
poire-z, line 280: Is their a 5th alternative behaviour when `self.cover_image_background` is `nil` ?
If yes, it would be nice it appears as the first option "Save cover as-is".
If no, one of these 4 should handle `self.cover_image_background == nil` in the checked_func.

poire-z:
> now I have a strange commit 29dcc6b in here. If this does any harm, how can I remove it?

I'd say: don't worry about it :) as long as it looks good in the global `Files changed` tab.
(We'll squash all your commits when merging, so we won't see that bogus commit.)

plugins/coverimage.koplugin/main.lua
zwim: We should not have a 5th alternative, because actually we do not stretch the image to fullscreen, we save it as is. The firmware of the device can take this and scale, stretch or do whatever the device does. 
So actually the 4th menu entry should be named better, something the way: "Save image as is and let the firmware deal with stretching" (in better word of course).

`self.cover_image_background` cannot be `nil` because we initialize it in line 43 with "black" if it is nil.

zwim: Maybe we could name it "Save the original image" and make a help_text with the info "The firmware can stretch the image to fit the screen"

poire-z:
```
white
gray
black
stretch
as-is
```
That makes 5. Which one would you remove?

(Or do you mean you don't really need stretch ? And what I thought was better named stetch was actually just doing scaling to screen size, preservering aspect ratio? So, as-is just scaling to screen size, keep aspect ratio - and we can actually keep it doing that while still naming it as-is, for simplicity).

Let's keep using "scale" when the aspect ratio is preserved (and "stretch" only when it is not).

zwim: We have:
black -> scale and add black background
white -> scale and add white background
gray -> scale and add gray background
as-is  -> just save 

Until now, the stretch option is not implemented in the plugin, because it is not needed. The firmware takes an image and stretches it to fully fit the screen (which does not work very good on Poke2, but works on Tolino).

Of course, we can add a fifth option `stretch`, if you think that's necessary.

I am not really satisfied with the naming of the menu options, wouldn't it  be better to name the thing: `Size and background` and the submenus like
```
[ ] Original size
[ ] Scale, black background
[ ] Scale, white background
[ ] Scale, gray background
```

poire-z: OK, I'm fine with that proposal (no need for stretch if it's the default behaviour on some FW, and otherwise the least interesting option as the others are better).

poire-z:
Ready to merge ?
@Frenzie: fine with the wording ?

zwim:
@poire-z: From my side it is ready.
Let's see, what magic @Frenzie does with the wording ;-)

plugins/coverimage.koplugin/main.lua
Frenzie: That looks fine to me. :+1:

Frenzie (approved):

PR #6887, opened 2020-11-16
android: fix dict overrides

pazos:
Fix for https://www.mobileread.com/forums/showpost.php?p=4058592&postcount=1

- User overrides marked as `false` were not checked
- Actions without app were skipped.

It seems that I forgot to check app picker settings :P

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6887)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -102,7 +102,7 @@ local Device = Generic:new{
     doExternalDictLookup = function (self, text, method, callback)
         external.when_back_callback = callback
         local _, app, action = external:checkMethod("dict", method)
-        if app and action then
+        if action then
             android.dictLookup(text, app, action)
         end
     end,
--- a/frontend/device/thirdparty.lua
+++ b/frontend/device/thirdparty.lua
@@ -25,16 +25,11 @@ function M:new(o)
         if ok then
             o[role.."s"] = user_dicts
             o.is_user_list = true
-        else
-            local t = o[role.."s"]
-            for i, value in ipairs(t or {}) do
-                local app = value[4]
-                if o:check(app) then
-                    value[3] = true
-                end
-            end
-            if t then
-                self[role.."s"] = t
+        end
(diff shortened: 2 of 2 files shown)


Frenzie (approved):

norbi24:
Hi guys, wouldn't it make sense along with this change to add thise generic dict lines to the dicts array in device.lua like this:
dicts = {
    { "Aard2", "Aard2", false, "itkach.aard2", "aard2" },
    ...
    { "QuickDic", "QuickDic", false, "de.reimardoeffinger.quickdic", "quickdic" },

    { "Send", "Choose Dict by App picker (send action)", true, nil, "send" },
    { "Search", "Choose Dict by App picker (search action)", true, nil, "search" },
    { "Text", "Choose Dict by App picker (text action)", true, nil, "text" }, 
}

so that everyone could use his own dictionary app which he has installed on Android ?
Yes I know this behaves in a little different from pre-configured dict apps, but I think everyone who would select one of this
generic options, would then immediately see how it works ...

pazos:
@norbi24: Actually we cannot enforce `Choose Dict by App picker` inside the app. The intent will be handled by the system and the app picker will be shown only if the user didn't set a default for a given action.

It can be implemented within the program, somehow like CC is doing for each filetype, but I think the current implementation works fine. Anyhow the building blocks are there, so PR welcome against https://github.com/koreader/android-luajit-launcher.

And probably the nº of users that use 3rd party dicts outside defaults can be counted with the fingers of one hand :smile:

norbi24:
Thanks for the explanation, understood, but I assume that I will still be able to add those generic picker lines into my override
dictionaries.lua, right ?
And yes,  originally I thought this could be a breaking feature making usage of good dictionaries from known editorials possible,
but it really looks that no one needs this ...

pazos:
> I assume that I will still be able to add those generic picker lines into my override dictionaries.lua, right ?

Of course. That's the point. **Made for power users** :tm:

PR #6895, opened 2020-11-19
[lang] Minor touch-up to non-linear info text

Frenzie:
As an aside, it should be "ToC," not "Toc." ;-)

Cf. <https://github.com/koreader/koreader/pull/6847>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6895)


Jellby:
Note that "single page, non-scrolling" was meant to indicate "no scroll" and "no double page".

Frenzie:
@Jellby Good like this?

Jellby:
I'm fine with that.

PR #6885, opened 2020-11-14
Pan zoom mode

jperon:
Adds an "horizontal panning" zoom mode: like column mode, but panning horizontally, with overlap.

It should be useful when reflow is impossible or doesn't give a good result. Comics especially are much more readable.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6885)

PR #6898, opened 2020-11-20
ReaderFooter: Cleanup some more messy dimensions handling

### LINKED ISSUE #6893: Can't hide status bar until the page is changed.
* KOReader version: 2020.11 (regression, doesn't affect 2020.10)
* Device: all platforms

#### Issue

Can't hide status bar until the page is changed.

#### Steps to reproduce

1. enable show all at once
2. disable progress bar
3. hide the status bar
4. move to next page
5. show the status bar
6. try to hide the statusbar. It won't work until the page is switched.

Reported in https://www.mobileread.com/forums/showthread.php?p=4059187#post4059187

NiLuJe:
Some stuff was still hacked on manually instead of trusting the widget system to do things right, which it does, if you update the right stuff at the right time the right way ;). *This Is The Way*.

Fix #6893 (and address https://github.com/koreader/koreader/pull/6878#discussion_r523411883, because it was indeed redundant ^^).

Includes a bonus fix for a number of settings not being applied immediately in PDFs.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6898)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -692,6 +692,7 @@ function ReaderFooter:resetLayout(force_reset)
 end
 
 function ReaderFooter:getHeight()
+    print("ReaderFooter:getHeight")
     if self.footer_content then
         -- NOTE: self.footer_content is self.vertical_frame + self.bottom_padding,
         --       self.vertical_frame includes self.text_container (which includes self.footer_text)
@@ -1876,13 +1877,6 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
         self.progress_bar.width = math.floor(
             self._saved_screen_width - self.text_width - self.settings.progress_margin_width*2)
     end
-    local bar_height
-    if self.settings.progress_style_thin then
-        bar_height = self.settings.progress_style_thin_height or PROGRESS_BAR_STYLE_THIN_DEFAULT_HEIGHT
-    else
-        bar_height = self.settings.progress_style_thick_height or PROGRESS_BAR_STYLE_THICK_DEFAULT_HEIGHT
-    end
-    self.progress_bar:setHeight(bar_height)
 
     if self.separator_line then
         self.separator_line.dimen.w = self._saved_screen_width - 2 * self.horizontal_margin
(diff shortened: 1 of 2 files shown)


poire-z (approved):
_Looks_ alright.

Frenzie (approved):
Looks fine, let me know if you want me to test though. ;-)

poire-z:
Btw, @NiLuJe Footer Master, regarding https://github.com/koreader/koreader/pull/6885#issuecomment-731015316 :

In ReaderCropping:onPageCrop(), we backup a lot of settings to reset the view dimension to what's available above the bottom buttons.
Except that ReaderView (I guess) will still add (or remove, depending on how you look at it) the space taken by the footer:
image
If I hide the footer before launching that crop dialog, there is not this bottom grey blank area.

Is there a canonical way to check the footer is visible and hide it if it is (and later unhide it so we get back the footer exactly as it was) I could do in the part where we backup the original dimensions?

NiLuJe:
@poire-z: You can check the non-ReaderFooter bits of #6830 for how the check (should) work.

Actually *toggling* it for real might be a bit trickier, though, because I don't think we have a direct way to do that: the tap *cycles* through modes. Might be able to store/set to 0/restore the mode manually and then call a full redraw, though?

NiLuJe:
@Frenzie: I think I checked the usual corner-cases, but, by all means ;).

poire-z:
> You can check the non-ReaderFooter bits of #6830 for how the check (should) work.
Actually toggling it for real might be a bit trickier, though

Thanks. Actually, I don't really need to toggle it... I just need to lie/tell ReaderView it's not visible :) 
```lua
    -- backup original footer visibility
    self.orig_view_footer_visibility = self.view.footer_visible
    self.view.footer_visible = false
```
works for me!

NiLuJe:
And just fixed a slightly older bug so that toggling some stuff (mainly, the progress bar) in the settings actually repaints stuff immediately and properly in ReaderPaging.

(The old logic was relying on a CRe only event for a bit of refresh optimization).

Frenzie:
@NiLuJe 
> I think I checked the usual corner-cases

And there's a lot of 'em too. Nothing obviously wrong.

PR #6900, opened 2020-11-20
ReaderCropping: prettier bottom buttons

poire-z:
Also remove the grey gap between the page and the buttons that would appear only when the footer is visible.

Suggested at https://github.com/koreader/koreader/pull/6885#issuecomment-731015316. It would look now like:

image

I switched OK and Cancel because we usually have OK on the right (although OK is somehow the most destructive - of the current crop settings - and I like destructive things on the left, less acccessible, but well, OK like all OKs.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6900)

### DIFF
--- a/frontend/apps/reader/modules/readercropping.lua
+++ b/frontend/apps/reader/modules/readercropping.lua
@@ -65,7 +65,7 @@ function ReaderCropping:onPageCrop(mode)
                 callback = function() self:onCancelPageCrop() end,
             },
             {
-                text = _("OK"),
+                text = _("Apply crop"),
                 callback = function() self:onConfirmPageCrop() end,
             },
         }},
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readercropping.lua
```diff
@@ -111,20 +53,59 @@ function ReaderCropping:onPageCrop(mode)
...
+        width = Screen:getWidth(),
+        button_font_face = "cfont",
+        button_font_size = 20,
+        buttons = {{
+            {
+                text = _("Cancel"),
+                callback = function() self:onCancelPageCrop() end,
+            },
+            {
+                text = _("OK"),
```
Frenzie, line 68: In line with our more recent UI guidelines we could also change it to something like `Crop`, `Apply (page) crop`, `Crop to … selection?`.

poire-z: Let's go with `Cancel | Apply crop` to keep it short.

PR #6902, opened 2020-11-21
OPDS*: Mangle Calibre feeds some more so that they don't confuse luxl/us

NiLuJe:
By essentially dropping the whole XHTML block, instead of trying to salvage each and every tag one by one as we did before.

Also, as that's usually the result after broken parsing, handle nil URLs slightly better in the frontend, so that they get caught/reported properly instead of doing nothing and/or crashing half the time.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6902)

### DIFF
--- a/frontend/ui/opdsparser.lua
+++ b/frontend/ui/opdsparser.lua
@@ -31,7 +31,6 @@ local function unescape(str)
 end
 
 function OPDSParser:createFlatXTable(xlex, curr_element)
-    print("OPDSParser:createFlatXTable", xlex, curr_element)
     curr_element = curr_element or {}
 
     local curr_attr_name
@@ -39,9 +38,7 @@ function OPDSParser:createFlatXTable(xlex, curr_element)
 
     -- start reading the thing
     for event, offset, size in xlex:Lexemes() do
-        print("event, offset, size:", event, offset, size)
         local txt = ffi.string(xlex.buf + offset, size)
-        print("txt:", txt)
         if event == luxl.EVENT_START then
             if txt ~= "xml" then
                 -- does current element already have something
@@ -75,9 +72,6 @@ function OPDSParser:createFlatXTable(xlex, curr_element)
 end
 
 function OPDSParser:parse(text)
-    print("OPDSParser:parse")
-    print("Orig text:")
-    print(text)
     -- Murder Calibre's whole "content" block, because it's not XML, it's XHTML, and luxl doesn't like it one bit...
     text = text:gsub('<content type="xhtml">(.-)%</content>', '')
     -- luxl cannot properly handle xml comments and we need first remove them
@@ -93,11 +87,7 @@ function OPDSParser:parse(text)
     text = text:gsub("<!%[CDATA%[(.-)%]%]>", function (s)
         return s:gsub( "%p", {["&"] = "&amp;", ["<"] = "&lt;", [">"] = "&gt;" } )
     end )
-    print("text:")
-    print(text)
     local xlex = luxl.new(text, #text)
-    print("xlex:")
-    print(xlex)
     return assert(self:createFlatXTable(xlex))
 end
 
--- a/frontend/ui/opdsparser.lua
+++ b/frontend/ui/opdsparser.lua
@@ -76,9 +76,9 @@ function OPDSParser:parse(text)
     -- as the list of crappy replacements below attests to...
     -- There's also a high probability of finding orphaned tags or badly nested ones in there, which will screw everything up.
     text = text:gsub('<content type="xhtml">.-</content>', '')
-    -- luxl cannot properly handle xml comments and we need first remove them
+    -- luxl doesn't handle XML comments, so strip them
     text = text:gsub("<!%-%-.-%-%->", "")
-    -- luxl prefers <br />, other two forms are valid in HTML, but will kick luxl's ass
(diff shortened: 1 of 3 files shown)


NiLuJe:
The net effect is that the listing should finally be accurate/complete.

frontend/ui/opdsparser.lua
```diff
@@ -73,6 +72,8 @@ function OPDSParser:createFlatXTable(xlex, curr_element)
 end
 
 function OPDSParser:parse(text)
+    -- Murder Calibre's whole "content" block, because it's not XML, it's XHTML, and luxl doesn't like it one bit...
```
Frenzie, line 75: XHTML is definitely XML. If it's not well-formed XML it's not XHTML. ;-)

NiLuJe: True, what I meant by that is that it's not an HTML parser, so it gets thoroughly thrown by the HTML bits ;).

NiLuJe: At a guess, I'd say, among other things, something like `<em/>` vs. `</em>`...

Frenzie: A self-closing em would be useless but well-formed. :-p

NiLuJe:
Yup, both testcases happened to have an orphaned `<em/>`...

Added it to the cleanup list, but I kept the full block removal, since we don't use it for anything, and it's bound to contain quirky syntax somewhere, somehow.

NiLuJe:
I vaguely looked into investing in an actual XML parser w/ XPath capabilities, and after a quick Google search and a look at http://lua-users.org/wiki/LuaXml, I found http://lua-users.org/wiki/LuaXml, which looks neat.. But it doesn't support (AFAICT) XPath queries with a registered namespace, which is necessary to run XPath queries against an Atom feed (apparently).

Since I don't actually speak XML/XPath, that idea went down the drain, but, food for thought for interested parties...

----

Quick and dirty attempt so that `document:search("//atom:entry", { "atom", "http://www.w3.org/2005/Atom" })` actually returns something apparently useful, but then I still can't run simple queries against each entry, or something, so, meh.

Again, I suck at XML/XPath.

```diff
diff --git a/xmlua/libxml2.lua b/xmlua/libxml2.lua
index 3b7b066..0866158 100644
--- a/xmlua/libxml2.lua
+++ b/xmlua/libxml2.lua
@@ -639,6 +639,8 @@ function libxml2.xmlXPathEvalExpression(expression, context)
 end
 jit.off(libxml2.xmlXPathEvalExpression)
 
+libxml2.xmlXPathRegisterNs = xml2.xmlXPathRegisterNs
+
 function libxml2.xmlStrdup(string)
   return xml2.xmlStrdup(string)
 end
diff --git a/xmlua/libxml2/xpath.lua b/xmlua/libxml2/xpath.lua
index 935e79a..c2ecb11 100644
--- a/xmlua/libxml2/xpath.lua
+++ b/xmlua/libxml2/xpath.lua
@@ -176,4 +176,7 @@ int xmlXPathSetContextNode(xmlNodePtr node, xmlXPathContextPtr ctx);
 xmlXPathObjectPtr xmlXPathEvalExpression       (const xmlChar *str, xmlXPathContextPtr ctxt);
 
 void xmlXPathFreeObject                (xmlXPathObjectPtr obj);
+
+int xmlXPathRegisterNs         (xmlXPathContextPtr ctxt, const xmlChar *prefix, const xmlChar *ns_uri);
+
 ]]
diff --git a/xmlua/searchable.lua b/xmlua/searchable.lua
index f7c2056..4aea1da 100644
--- a/xmlua/searchable.lua
+++ b/xmlua/searchable.lua
@@ -50,12 +50,17 @@ local ERROR_MESSAGES = {
   [ffi.C.XPATH_FORBID_VARIABLE_ERROR]    = "Forbidden variable\n",
 }
 
-function Searchable:search(xpath)
+function Searchable:search(xpath, ns)
   local document = self.document
   local context = libxml2.xmlXPathNewContext(document)
   if not context then
     error("failed to create XPath context")
   end
+  if ns then
+    if not libxml2.xmlXPathRegisterNs(context, ns[1], ns[2]) then
+      error("failed to register XPath namespace:", ns[1])
+    end
+  end
   if self.node then
     if not libxml2.xmlXPathSetContextNode(self.node, context) then
       error("failed to set target node: <" .. tostring(self.node) .. ">")
```

Frenzie:
As an aside, the newsdownloader plugin includes some xml lib.

NiLuJe:
Turns out my fugly xmlua patch works, I just suck at XPath ;p.

```
LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/
JIT: ON SSE2 SSE3 SSE4.1 BMI2 fold cse dce fwd dse narrow loop abc sink fuse
> xmlua = require("xmlua")
> xml_file = io.open("/var/tmp/niluje/test.xml")
> xml = xml_file:read("*all")
> xml_file:close()
> document = xmlua.XML.parse(xml)
> entries = document:search('/atom:feed/atom:entry', { 'atom', 'http://www.w3.org/2005/Atom' })                      
> print(#entries)
3
> files = entries[1]:search('atom:link[@rel="http://opds-spec.org/acquisition"]', { 'atom', 'http://www.w3.org/2005/Atom' })
> print(#files)
2
> print(files[1].href)
/get/azw3/4022/Calibre
> 
```

PR #6904, opened 2020-11-23
Add functionality to write BMP-Images in Cover-Image

zwim:
This PR requires https://github.com/koreader/koreader-base/pull/1233.
It allows to store the image depending on its extension, as PNG or as BMP.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6904)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -87,9 +87,8 @@ function CoverImage:createCoverImage(doc_settings)
             local scale_factor = math.min(s_w / i_w, s_h / i_h)
 
             if self.cover_image_background == "none" or scale_factor == 1 then
-                local start_time = os.clock()
-                cover_image:writePNG(self.cover_image_path, false)
-                require("logger").err("xxxxxxxxxxxxxxxxx time=" .. (os.clock()-start_time))
+                local act_format = self.cover_image_format == "auto" and self.cover_image_extension or self.cover_image_format
+                cover_image:writeToFile(self.cover_image_path, act_format)
                 cover_image:free()
                 return
             end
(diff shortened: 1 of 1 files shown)


zwim:
This PR could solve the problem https://github.com/koreader/koreader/pull/6813#issuecomment-725569462 with the BMP-file-format.
@poire-z

zwim:
Sorry, wrong button.

poire-z:
`base` with these changes has already been bumped and merged (but not your lastest base commit).

You started this PR before that, so your start point is before that.
You need to rebase this PR from the current master.
I usually do it this way:
(Be sure your bmp-Support is ok, everything commited - save a backup just in case: `git diff master > ~/backup.diff`)
```
$ git checkout master
$ git fetch upstream
$ git merge upstream/master
$ git push origin master
$ git checkout bmp-Support
$ git rebase origin/master
  ^ this will set your bmp-Support to start at where master is, and reapply your commits starting from that (there could be some conflict to solve, can be painful)
$ git push -f origin bmp-Support
```
May be you need to drop your base bump commit before doing that, not sure.

zwim:
@poire-z: Thank you

poire-z:
Your "add base change" now looks fine.
But the next one ("Revert...") is strange and reverts your previous one and other previous ones. So, may be drop it?

poire-z:
Looks better :)

poire-z (approved):
Looks OK.
@Frenzie: a quick look at the translatable strings?

Frenzie:
Seems fine at a glance.

poire-z:
Should we merge this? Or do you want to keep it open until you can bump your writeJPG stuff?

zwim:
Merging it now would bring us a working cover image plugin on Pocketbook (https://github.com/koreader/koreader/pull/6813#issuecomment-725569462).
On the other hand jpg stuff seems to be very promising, but with jpg-stuff I have the problem with the jit-optimizer on bmp again, which might take some time to solve. 

I would say, merge it now, one thing at one time.

poire-z:
> would bring us a working cover image plugin on Pocketbook 

Don't you need to enable it then on PocketBook ? It still has:
```lua
if not Device.isAndroid() and not Device.isEmulator() then
    return { disabled = true }
end
```

zwim:
Ohh, I forgot. I will put that in #6906.

poire-z:
OK. So, both PRs are ready and I can merge both?

zwim:
Yes, that would be great.

PR #4114, opened 2018-07-27
CPUFreq shenanigans on Kobo

NiLuJe:
Wild-ass experiment switching to conservative instead of userspace (because, well, there's no userspace anymore :D).

Note that my H2O seems to grandly ignore all of that crap and switch freq all on its own anyway, so this may be a giant NOP.

PR #6905, opened 2020-11-24
Kobo, Mk.5: Tweak CPU frequency scaling

NiLuJe:
A couple years after #4114, after countless times of wondering what the hell was happening with cpufreq on the H2O, I, err, finally figured it out ;p.

Turns out there's a weird old i.MX feature for frequency/voltage scaling, called dvfs, and that's what's being used (hence userspace as the governor, to limit interactions with it).
Except, there's a twist: it's only enabled while Wi-Fi is powered on. And disabled otherwise. And what happens when you disable DVFS with userspace, you ask me? Why, yes, of course, the kernel pegs the CPU to max freq :D.

There's also something that sure looks like a Nickel bug, where the behavior is inverted on the first boot, so, you get Wi-Fi, no DVFS, and the CPU pegged at max clock.
Funnily enough, that's what I was relying on for halfway reliable benchmarks...

----

So, on our end, since we have somewhat predictable load patterns, instead of being stuck at max clock, I've switched to the only actually usable cpufreq governor we have available on those kernels: conservative, but with tunables set somewhat aggressively to attempt to alleviate its high latency.

Here be dragons! But hopefully this should help avoid melting batteries on Mk. 5 ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6905)


NiLuJe:
I have no idea what the situation is on < Mk. 5, so the check is extremely defensive to leave those devices alones.

On Mk. 6 & Mk. 7, we attempt to use interactive, which should be the default anyway.
(It's the kernel default on Mk. 6; while on Mk. 7, it's Nickel's choice. The kernels mysteriously default to ondemand).

I suspect ondemand's default tunables are highly inappropriate, which is why it's never actually used ;).

NiLuJe:
As far as the choice of tunables go, it's basically set as low as possible:

Given that a CRe page-turn requires very few CPU time, the thresholds are set to the minimum (note that down_threshold has a sensibly less braindead behavior in current kernels, so this choice only makes sense on older kernels. I'm basing this on the doc from Linux 4.4 kernels on Android).

Bumped the sampling down factor to twice the sampling rate, to avoid doing the yo-yo too much, but given that said sampling rate is already super high (200ms), eh.

Bumped freq_step to 50%, because this isn't AMD/Intel, there's only three steps available: 166MHz, 500MHz, 1GHz. So we want to jump to at least 500MHz ASAP, and then to max right after.

Frenzie (approved):
Dunno about the factors/thresholds but looks okay.

NiLuJe:
xref to the CPUFreq docs:

Interactive: https://android.googlesource.com/kernel/common/+/refs/heads/android-4.4-p-release/Documentation/cpu-freq/governors.txt

(Modern) mainline: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/pm/cpufreq.rst

NiLuJe:
Since I mentioned the interactive tuneables earlier, I just tried playing with it a bit on the Forma, and I can't really prevent it from dipping its toes into max freq for very small amounts of time, even when completely idle.

Usually for around 100ms, which is suspiciously like the defaults timer_slack + timer_rate...

.ci/install.sh
```diff
@@ -60,8 +60,8 @@ else
 fi
 
 # install shfmt
-SHFMT_URL="https://github.com/mvdan/sh/releases/download/v3.0.1/shfmt_v3.0.1_linux_amd64"
-if [ "$(shfmt --version)" != "v3.0.1" ]; then
+SHFMT_URL="https://github.com/mvdan/sh/releases/download/v3.2.0/shfmt_v3.2.0_linux_amd64"
+if [ "$(shfmt --version)" != "v3.2.0" ]; then
```
Frenzie, line 64: Might as well sneak in a shellcheck update too.

NiLuJe: I checked, it's still up to date ;).

NiLuJe:
> Hold on, will the `#!/usr/bin/env bash` in our CI scripts still result in bash parsing mode?

It does for shellcheck, so I vaguely hope shfmt does the same ;).

.ci/install.sh
Frenzie: I was going to say I'd swear there's been a new release since February… and there was, but I already updated it in https://github.com/koreader/koreader/pull/6445 for other reasons.

Frenzie:
> It does for shellcheck, so I vaguely hope shfmt does the same ;).

Never mind regardless, I was going commit by commit.

NiLuJe:
> I can't think of an example of what you're talking about otoh?

https://github.com/koreader/koreader/blob/0ae86fc0ecc44b6824d30eeb51c95dbede003504/platform/kobo/nickel.sh#L16-L22

It expects `(....` instead of `(` to be aligned with the inner indentation level ;).

Issue #6899, opened 2020-11-20
FR: Enable CoverImage Plugin on Remarkable


gbyl:
@zwim's CoverImage plugin works like a charm on the Remarkable. For example, if you want to change the poweroff screen, simply set `usr/share/remarkable/poweroff.png` as the *Screensaver image filename*

Can we have it enabled by default by adding `and not Device.IsRemarkable()` to the line below?
https://github.com/koreader/koreader/blob/2785f0d1133e023d038a76a7cec7a1aee3f3575b/plugins/coverimage.koplugin/main.lua#L5

The only minor issue I see on the Remarkable is that it leaves a ghosted image of the previous screen before turning off. If someone could add a **full screen refresh** before loading the image that would be awesome!

NiLuJe:
If it's deferred to the system, we have no real control over *when* it's displayed, so adding a refresh becomes a potentially dicey proposition ;).

(And I'd rather not emulate the giant mess that is the Kindle codepath for that sort of things ^^).

zwim:
@gybl: You write, that the coverimage plugin works on remarkable, that's good.

As I see on the web, the remarkable has a lot of screen images in `/usr/share/remarkable/`.
Is it right, that you have used `usr/share/remarkable/poweroff.png` to show the cover at device's poweroff and got the ghosting?

Could you please try to set the image to `usr/share/remarkable/suspended.png` and put the device to suspend? I would be interested, if there is ghosting as well.

@NiLuJe: Does KOReader do a complete screen refresh on exit? 
Just a shot in the blue: Couldn't KOReader clear the screen (white image) and do a refresh before quit (exit or SIGTERM)?

pazos:
@gbyl @zwim: ReMarkable is described as a device that supports screensaver. See https://github.com/koreader/koreader/search?q=supportsScreensaver%28%29

That makes all the sense. It is a freescale board. Probably the "magic" used by stock to set suspend/reboot/poweroff images is in systemd units we're calling. You can replace them with their inner guts:

ie: instead of `systemctl reboot` explore what `systemctl reboot` does. It should do a simple `reboot`. Same for `poweroff`. For suspend things are a bit more complex, but indeed possible as explained in https://github.com/koreader/koreader/issues/6028#issuecomment-610411019.

IMO we shouldn't rely on tricks when full screensaver support is possible.

gbyl:
[...]

PR #6906, opened 2020-11-24
Enable Cover-Image on Remarkable and PocketBook

zwim:
Please see https://github.com/koreader/koreader/issues/6899 and https://github.com/koreader/koreader/issues/6899#issuecomment-732664106

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6906)

### DIFF
--- a/frontend/device/remarkable/device.lua
+++ b/frontend/device/remarkable/device.lua
@@ -95,7 +95,13 @@ function Remarkable:setDateTime(year, month, day, hour, min, sec)
     return os.execute(command) == 0
 end
 
+function Remarkable:_clearScreen()
+    self.screen:clear()
+    self.screen:refreshFull()
+end
+
 function Remarkable:suspend()
+    self:_clearScreen()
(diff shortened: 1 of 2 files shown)


NiLuJe:
Oh, cool, thanks for taking care of that ;).

Would you mind folding https://github.com/koreader/koreader/issues/6899#issuecomment-732664106 in there?

NiLuJe:
I meant the actual code ^^.

zwim:
@NiLuJe: I did not want to include "your" code in my commit.

NiLuJe:
It's trivial enough, and I would have said so if it bothered me (it doesn't), but thanks ;).

zwim:
> It's trivial enough, and I would have said so if it bothered me (it doesn't), but thanks ;).

So is the `and not Device.isRemarkable()` ;-)

NiLuJe (approved):

zwim:
Also enable on PocketBook, as bmp-support is ready now.

PR #6909, opened 2020-11-24
add target: debian-arm64

### LINKED ISSUE #6782: FR: [aarch64 linux]
The Debian builds for KOReader could hopefully include the arm64 architecture too, so that it can be used on mobile devices with linux as well, like the PinePhone, or the Librem5.

pazos:
Requires https://github.com/koreader/koreader-base/pull/1239
Closes #6782

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6909)


Frenzie (approved):
Very straightforward on this side; difficulty should be in base. :-P

PR #6910, opened 2020-11-24
Add Latin hyphenation

jperon:
This PR is also an issue: some months ago (I don't record exactly when), it was possible to choose hyphenation from a big list, an latin hyphenation thanks to [this file](https://jperon.github.io/latin/Latin.pattern), generated from [this project](https://github.com/gregorio-project/hyphen-la) was possible.

Since the only option depends on "typographical rules":

- I tried to add the modification proposed in this PR;

- I edited `data/hyph/languages.json`, adding following lines:
```json
    {
        "name": "Latin",
        "filename": "Latin.pattern",
        "language": "la",
        "left_hyphen_min": 2,
        "right_hyphen_min": 2,
        "aliases": ["lat"]
    },
```
- I added [Latin.pattern](https://jperon.github.io/latin/Latin.pattern) into `data/hyph`.

I'd have expected `Latin.pattern` to be used as hyphen dictionary; but the only option I have is `English-US`, although `typographical rules` are correctly set to `la`.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6910)

### DIFF
--- a/frontend/apps/reader/modules/readertypography.lua
+++ b/frontend/apps/reader/modules/readertypography.lua
@@ -66,6 +66,7 @@ local LANGUAGES = {
     { "it",               {"ita"},   "H   ",   _("Italian"),                "Italian.pattern" },
     { "ja",                    {},   "    ",   _("Japanese") },
     { "ko",                    {},   "    ",   _("Korean") },
+    { "la-lit",       {"lat-lit"},   "H   ",   _("Latin (liturgical)"),     "Latin.pattern" },
     { "la",               {"lat"},   "H   ",   _("Latin"),                  "Latin.pattern" },
     { "lv",               {"lav"},   "H   ",   _("Latvian"),                "Latvian.pattern" },
     { "lt",               {"lit"},   "H   ",   _("Lithuanian"),             "Lithuanian.pattern" },
--- a/frontend/apps/reader/modules/readertypography.lua
+++ b/frontend/apps/reader/modules/readertypography.lua
@@ -66,6 +66,7 @@ local LANGUAGES = {
     { "it",               {"ita"},   "H   ",   _("Italian"),                "Italian.pattern" },
     { "ja",                    {},   "    ",   _("Japanese") },
     { "ko",                    {},   "    ",   _("Korean") },
+    { "la",               {"lat"},   "H   ",   _("Latin"),                  "Latin.pattern" },
     { "lv",               {"lav"},   "H   ",   _("Latvian"),                "Latvian.pattern" },
     { "lt",               {"lit"},   "H   ",   _("Lithuanian"),             "Lithuanian.pattern" },
     { "mk",                  {""},   "H   ",   _("Macedonian"),             "Macedonian.pattern" },


NiLuJe:
This is because things now require a bit more metdata on CRe's side (c.f., https://github.com/koreader/crengine/pull/383, https://github.com/koreader/crengine/pull/378, https://github.com/koreader/crengine/pull/376) ;).

poire-z:
Yes, since a few months ago, some stuff is now hardcoded into crengine.
What you did is just the frontend part, which just makes stuff available in the menu.
You'll need to do a little PR in crengine (see https://github.com/koreader/crengine/commits/master/cr3gui/data/hyph for the little it takes to add each language, some can have a little bit more tweaks if needed).

jperon:
Just submitted as https://github.com/koreader/crengine/pull/393

frontend/ui/translator.lua
```diff
@@ -80,6 +80,7 @@ local SUPPORTED_LANGUAGES = {
     ky = _("Kyrgyz"),
     lo = _("Lao"),
     la = _("Latin"),
+    ["la-lit"] = _("Latin (liturgical)"),
```
poire-z, line 83: Why do you need that one here?
It's not an official lang tag, so I guess Google translate doesn't know anything about "la-lit".
Unless this is used for something else in KOReader ?

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -66,6 +66,7 @@ local LANGUAGES = {
     { "it",               {"ita"},   "H   ",   _("Italian"),                "Italian.pattern" },
     { "ja",                    {},   "    ",   _("Japanese") },
     { "ko",                    {},   "    ",   _("Korean") },
+    { "la",               {"lat"},   "H   ",   _("Latin"),                  "Latin.pattern" },
```
poire-z, line 69: But you miss `la-lit` here.

frontend/ui/translator.lua
jperon: Sorry. I shouldn't have wanted to go so quick…

poire-z:
koreader-base updated with the crengine bits, you'll have to update koreader-base in here (you'll be bringing 5 other little commits with it, no worry).

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -66,6 +66,8 @@ local LANGUAGES = {
     { "it",               {"ita"},   "H   ",   _("Italian"),                "Italian.pattern" },
     { "ja",                    {},   "    ",   _("Japanese") },
     { "ko",                    {},   "    ",   _("Korean") },
+    { "la-lit",       {"lat-lit"},   "H   ",   _("Latin (liturgical)"),     "Latin.pattern" },
```
Frenzie, line 69: Looks like I'm too late, but that should of course be `Latin_liturgical.pattern`. ;-)

poire-z: Going to fix that (no real issue, just a display one, we'll show the hyph filename instead of this pretty language name).
While I'm at it, should these be re-ordered, so the shorter/plain name is first, before the next longer/specific ones?
```diff
     { "ko",                    {},   "    ",   _("Korean") },
-    { "la-lit",       {"lat-lit"},   "H   ",   _("Latin (liturgical)"),
     { "la",               {"lat"},   "H   ",   _("Latin"),
+    { "la-lit",       {"lat-lit"},   "H   ",   _("Latin (liturgical)"),
     { "lv",               {"lav"},   "H   ",   _("Latvian"),
@@ -77,9 +77,9 @@ local LANGUAGES = {
     { "pms",                   {},   "H   ",   _("Piedmontese"),
-    { "pt-BR",                 {},   "HB  ",   _("Portuguese (BR)"),
     { "pt",               {"por"},   "HB  ",   _("Portuguese"),
+    { "pt-BR",                 {},   "HB  ",   _("Portuguese (BR)"),
     { "rm",               {"roh"},   "H   ",   _("Romansh"),
     { "ro",               {"ron"},   "H   ",   _("Romanian"),
+    { "ru",               {"rus"},   "Hb  ",   _("Russian"),
     { "ru-GB",                 {},   "Hb  ",   _("Russian + English (UK)"),
     { "ru-US",                 {},   "Hb  ",   _("Russian + English (US)"),
-    { "ru",               {"rus"},   "Hb  ",   _("Russian"),
     { "sr",               {"srp"},   "HB  ",   _("Serbian"),
```

poire-z: (It needs to be in the current order in crengine to match longer ones first - but in frontend, this order does not matter.)

Frenzie: Definitely makes more sense to me.

poire-z: Done with #6913.

PR #6916, opened 2020-11-27
Kobo: Warn on restart if the startup script is outdated

NiLuJe:
Should help prevent potentially puzzling issues whenever significant changes are made that require startup script tweaks (e.g., USBMS).

Also print a warning right after the update.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6916)

### DIFF
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -819,6 +819,15 @@ function Kobo:toggleChargingLED(toggle)
     io.close(f)
 end
 
+function Kobo:isStartupScriptUpToDate()
+    -- Compare the hash of the *active* script (i.e., the one in /tmp) to the *potential* one (i.e., the one in KOREADER_DIR)
+    local current_script = "/tmp/koreader.sh"
+    local new_script = os.getenv("KOREADER_DIR") .. "/" .. "koreader.sh"
+
+    local md5 = require("ffi/MD5")
+    return (md5.sumFile(current_script) == md5.sumFile(new_script))
+end
+
 -------------- device probe ------------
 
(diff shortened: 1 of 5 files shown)


NiLuJe:
Could be applied to other platforms (hence the check being deferred to a Device method), but, in practice, I don't really see one where it'd matter as much.

poire-z (approved):
Looks alright.

frontend/device/kobo/device.lua
```diff
@@ -819,6 +819,15 @@ function Kobo:toggleChargingLED(toggle)
     io.close(f)
 end
 
+function Kobo:isStartupScriptUpToDate()
+    -- Compare the hash of the *active* script (i.e., the one in /tmp) to the *potential* one (i.e., the one in KOREADER_DIR)
+    local current_script = "/tmp/koreader.sh"
+    local new_script = os.getenv("KOREADER_DIR") .. "/" .. "koreader.sh"
+
+    local md5 = require("ffi/MD5")
+    return (md5.sumFile(current_script) == md5.sumFile(new_script))
```
Frenzie, line 828: Bit C-ish with the redundant parentheses?

```suggestion
    return md5.sumFile(current_script) == md5.sumFile(new_script)
```

Frenzie (approved):
lgtm

frontend/device/kobo/device.lua
NiLuJe: Yeah, I vaguely recalled a few other instances where we do this kind of stuff like that, but I have no preference whatsoever ;).

pazos (approved):

PR #6917, opened 2020-11-27
AutoSave: Delay I/O until after the pageturn

NiLuJe:
`nextTick` was too early ;).

Prevents small hitches when turning the page for the page where this
triggers.

Apply the same trickery to the Stats DB insert, even if that one probably
had a much smaller impact.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6917)

### DIFF
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -959,10 +959,12 @@ function ReaderView:checkAutoSaveSettings()
     if not interval then -- no auto save
         return
     end
+
     local now_ts = os.time()
     if now_ts - self.settings_last_save_ts >= interval*60 then
         self.settings_last_save_ts = now_ts
-        UIManager:nextTick(function()
+        -- I/O, delay until after the pageturn
+        UIManager:tickAfterNext(function()
             self.ui:saveSettings()
         end)
     end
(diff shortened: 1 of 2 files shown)


NiLuJe:
This was driving me nuts on the Forma, until I remembered I'd enabled it every 5 min because I was doing heavy highlighting.

(As @poire-z knows, I usually have this disabled, and the suddenly random hitches were driving me crazy ^^).

poire-z:
Well, couldn't you love these hitches that friendly and discreetly made sign to you that your settings and statistics were safely being saved ?! Now, you won't know that, and will live in uncertainty and doubts ! :)

But ok, fine with me :)
(What's being done between nextTick and tickAfterNext?)

NiLuJe:
> Well, couldn't you love these hitches that friendly and discreetly made sign to you that your settings and statistics were safely being saved ?! Now, you won't know that, and will live in uncertainty and doubts ! :)
> 

:D 

> But ok, fine with me :)
> (What's being done between nextTick and tickAfterNext?)

... another tick? :D

Sorry, couldn't resist ;p. Not sure, but the actual refresh *is* the next tick, so, probably the next iteration of the input loop?

NiLuJe:
And if the actual question was whether the delay was unreliable/potentially "far" in the future: nope, it's really straight after the refresh ;).

PR #2809, opened 2017-04-24
readermenu(fix): override pan gesture from scrolling

houqp:
Without this patch, I got into a really frustrating state where swipes at the top only triggers page scrolls and it took me 20 swipes to get the menu out :(

PR #2761, opened 2017-04-13
Swipe for menu and quickstart guide

Frenzie:
Because swiping for the menu is a big change from what we're used to, this commit includes a new quickstart guide.

Fixes #2608.

screenshot_2017-04-13_13-41-02
screenshot_2017-04-13_13-41-09

PR #6918, opened 2020-11-27
[RFC] Make default UI tapzones more generous

NiLuJe:
By which I mean, make the menu tapzone taller, the configdialog tapzone taller, and the footer tapzone *much* taller.

Those were designed back when devices where 600x800 @ 167DPI.

I've *always* found the footer one nigh unusable, even back then, but it's only gotten worse the higher resolution devices got.

I've been using this set of values on my Forma since basically forever, but, obviously this affects a lot of potentially different use-cases, so, comments welcome :).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6918)


NiLuJe:
(I did make sure this wouldn't affect the corner tapzones).

NiLuJe:
As a FYI: I chiefly open menus via taps. And turn pages via buttons/taps, too, FWIW.

defaults.lua
```diff
@@ -72,9 +72,9 @@ FOLLOW_LINK_TIMEOUT = 0.5
 -- y: y coordinate of top left corner in proportion to screen height
 -- w: tap zone width in proportion to screen width
 -- h: tap zone height in proportion to screen height
-DTAP_ZONE_MENU = {x = 1/8, y = 0, w = 3/4, h = 1/8}
-DTAP_ZONE_CONFIG = {x = 1/8, y = 7/8, w = 3/4, h = 1/8}
-DTAP_ZONE_MINIBAR = {x = 0, y = 31/32, w = 1, h = 1/32}
+DTAP_ZONE_MENU = {x = 1/8, y = 0, w = 3/4, h = 1/3}
+DTAP_ZONE_CONFIG = {x = 1/8, y = 3/4, w = 3/4, h = 1/4}
+DTAP_ZONE_MINIBAR = {x = 0, y = 12/13, w = 1, h = 1/13}
```
poire-z, line 77: 13 in 12/13 does not feel multiple of 2 like all the other do :) Dunno if that matters in any way, and if we math.floor/round all thar properly and if that matters :)

Frenzie:
Could you maybe also do a more visual before/after? (Even if "fake". ;-) )

defaults.lua
NiLuJe: it shouldn't, I'm guessing fractions were used to make the computations easier to grok (i.e., y+h == 1).

I'm actually using `h = 0.075` & y = `0.925` here, and this was the closest.

NiLuJe:
On a H2O (1080x1429), old tapzones in red, new in blue:

Menu:

menu_tz

Config:

config_tz

Footer:

footer_tz

Frenzie:
I'll have to give it a try. I think I might be using some of those blue areas for left/right.

defaults.lua
NiLuJe: Getting to those exact values using fractions would mean using `12.333/13.333` and `1/13.333` which is *definitely* worse ^^.

NiLuJe:
FWIW, the full set of defaults look like this:

page_tz

(XCF source attached, with the layers possibly in the right override order ^^).

[zones.tar.gz](https://github.com/koreader/koreader/files/5609726/zones.tar.gz)

NiLuJe:
(The rationale behind the menu going so far down is to be able to tap the menu open when that very region can contain a giant chapter title link ;)).

NiLuJe:
As a comparison, this is the TZ selection menu in Nickel:

nickel_tz

poire-z:
> Those were designed back when devices where 600x800 @ 167DPI.
> I've always found the footer one nigh unusable, even back then, but it's only gotten worse the higher resolution devices got.

But, as they are fractions of screen width and height, how can the resolution have any effect on the areas ?

Anyway, I don't use the default zones, so no much idea on how the changes will affect people who do.
But I use tap for menu, and I do have tweaked my top area quite a bit - adding myself a second zone in the middle, making it a T shape - so I guess I would be happy with it becoming taller.
I also somehow have tap in the middle open both menus (I thought this might have been the default?), so even with a giant chapter title link, I can get both of them :) but I actually never use that/tap there :)

NiLuJe:
> > Those were designed back when devices where 600x800 @ 167DPI.
> > I've always found the footer one nigh unusable, even back then, but it's only gotten worse the higher resolution devices got.
> 
> But, as they are fractions of screen width and height, how can the resolution have any effect on the areas ?
> 

The DPI affects how much *content* is displayed inside that fraction of screen real-estate ;).

> Anyway, I don't use the default zones, so no much idea on how the changes will affect people who do.
> But I use tap for menu, and I do have tweaked my top area quite a bit - adding myself a second zone in the middle, making it a T shape - so I guess I would be happy with it becoming taller.
> I also somehow have tap in the middle open both menus (I thought this might have been the default?), so even with a giant chapter title link, I can get both of them :) but I actually never use that/tap there :)

There's an option, which might be enabled by default, that does indeed open both menus at one. I have that disabled on my end ;).

Frenzie:
> As a comparison, this is the TZ selection menu in Nickel:

Note that it's thinner. I was mostly referring to this area (see bottom right, but also top right and possibly on the left):

bla

Frenzie:
I should probably point out at this point that you won't get any menu on my H2O by tapping in part because I already find that too easily triggered. But on my Android devices I've been using the default. It's a somewhat important difference because your fingers are much bigger on a phone. ;-)

NiLuJe:
Hmm, the phone use-case does raise an interesting conundrum...

FWIW, added the (default) Nickel zone (37.5%) to help visualize the difference:

page_tz

NiLuJe:
It feels great for the blue region, but way too narrow for the red one.

Which means I don't really see a better solution than splitting those in two zones, which might feel a tad overkill?

NiLuJe:
Something like this :).

page_tz

[zones.tar.gz](https://github.com/koreader/koreader/files/5611265/zones.tar.gz)

poire-z:
(Just wondering if the ext tap zone should really handle swipe/pan.
I think the "pos" of a swipe is its starting position, so it might be enough to handle swipe from the regular zone.
But I have swipe to open menus disabled, so what do I know :) )

NiLuJe:
It's definitely helpful for the *closing* gesture, which, granted, is handled (mostly) completely differently:

* For *Menu, it's the actual size of the menu, which makes perfect sense.
* Fr ConfigDialog, it's All The Things. I vaguely considered simplifying it to just the size of the widget, but, having the ext zone in there is mildly helpful for the (rare ;p) small tabs.

Granted, I'm also not a swiper, so, comments welcome, as @poire-z is right, the gesture is based on the starting point.

Frenzie:
Presumably any rationale to extend the tap zone also applies to the swipe start zone? If anything the difference is that since the swipe zone doesn't interfere with tapping it can more liberally extend itself if desired. Since it's swipe up/down it doesn't interfere with swipe left/right/diagonal/multiswipe either. Only with manual panning.

I'd argue that yes, you may require a tiny bit more precision the way things are (give or take top/bottom zone height unification), but for the most part you can just tap on the page wherever you want to follow links or to go back/forward, which is more important to me than opening the menu.

I'd propose a much more modest extension to alleviate potential issues. Something like this:

bla

hius07:
> I'd propose a much more modest extension to alleviate potential issues.

Yes.
Maybe add the checkbox "Taller tap zones" to the "Activate menu" submenu.

Frenzie:
> Maybe add the checkbox "Taller tap zones" to the "Activate menu" submenu.

Ideally we'd have a few presets (cf. Nickel) and perhaps even easy GUI customization (cf. the crop dialog), but the defaults in their *current*, non-extended form have never bothered me.

NiLuJe:
Well, the fact that neither of us actually *use* the defaults would kinda hint at the fact that the defaults are maybe problematic ;o).

Frenzie:
Fair, but my reasoning is the exact opposite. 

On Sunday, November 29, 2020, NiLuJe wrote:
> Well, the fact that neither of us actually *use* the defaults would kinda hint at the fact that the defaults are maybe problematic ;o).
> 
> -- 
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/koreader/koreader/pull/6918#issuecomment-73540160

-- 
Sent from my Sailfish device

Frenzie:
Also I made it the default for a month or two in 2017 (?).

NiLuJe:
I can only see #2809, which made the menu smaller (from its original state, 7 years ago), and #4099, which made the footer smaller, because of a bug that no longer exists ;).

NiLuJe:
That makes for:

page_tz

[zones.tar.gz](https://github.com/koreader/koreader/files/5612360/zones.tar.gz)

NiLuJe:
Menu:

menu_tz

Config:

config_tz

Footer:

footer_tz

Frenzie:
> I can only see #2809, which made the menu zone smaller (from its original state, 7 years ago), and #4099, which made the footer zone smaller, because of a bug that no longer exists ;).

I use the standard *zones*. But that PR was a fix for https://github.com/koreader/koreader/pull/2761, which is what I was referring to.

NiLuJe:
Anything new to add, or shall we try to unleash this on the masses? ;).

Frenzie:
I still think double is very tall, better 130-160% in my view.

NiLuJe:
@ 160%

page_tz

Frenzie:
lol, going straight for the max for the range I gave. :-P

but fine. ;-)

NiLuJe:
Well, the difference would be nearly useless at 130% ;p.

poire-z:
Just a thought. This PR doesnt change:
```
DTAP_ZONE_MENU = {x = 1/8, y = 0, w = 3/4, h = 1/8}
```

Since the original zones, a few things happened in the code:
- we have proper overrides to avoid conflict and be ok with overlaps
- we have optional corners zones, that can be disabled
- to activate the top menu, if the tap is on the far left or far right, we switch to the first or the last tab

So, I think the `x=1/8 w=3/4` might no more be needed (it was for the dogear bookmarking to work I guess), and we could use `x=0 w=1`. (Havent checked though :) 

(Just for info, I personally have since a long time, and not updated for the points I mentionned above:
`DTAP_ZONE_MENU = {x = 2/5, y = 1/8, w = 1/5, h = 5/8}` which plays the role of your new one
and I hardcoded another one equivalent to `{x=1/16, y=0, w=7/8, h=1/8}` - to which I added support for some more specific swipe directions, that I think I could migrate to the swipe on edge gestures, but well...)

NiLuJe:
@poire-z: That's a very good point, and, yep, it works ;).

NiLuJe:
page_tz

[zones.tar.gz](https://github.com/koreader/koreader/files/5625783/zones.tar.gz)

Frenzie (approved):
Didn't check in detail but seems fine.

PR #6919, opened 2020-11-27
android: switch between internal and external sdcard

### LINKED ISSUE #6912: Can't browse internal memory with home set to SD card
I set home in koreader to /sdcard/Android/data/org.koreader.launcher on an external SD card formatted FAT32.  That works, but having done this, I see no way to browse to internal memory.

Side notes:
koreader shows this path as /storage/0b17-1ef9/Android/...
Hardware is a Likebook Mars with Android 8.1
I didn't want to tie the SD card to one device. I insert the card in a Windows PC to copy in ebooks.
Discovering that koreader and most other apps could only read files in a directory named for the app was not easy.

pazos:
Android nonsense nº 1458349

Switch between two storage volumes that cannot be accesed directly using the filemanager.

test

Fixes #6912 

Requires some changes in luajit-launcher. I will push them once I get a confirmation of https://github.com/koreader/koreader/issues/6912#issuecomment-734943107

Please review naming while we're waiting :smile:

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6919)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -641,11 +641,37 @@ function FileManager:tapPlus()
         table.insert(buttons, 3, {
             {
                 text = _("Import files here"),
-                enabled = Device.isValidPath(self.file_chooser.path),
+                enabled = Device:isValidPath(self.file_chooser.path),
                 callback = function()
                     local current_dir = self.file_chooser.path
                     UIManager:close(self.file_dialog)
                     Device.importFile(current_dir)
+                end,
+            },
+        })
+    end
+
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Sure.

poire-z (approved):
Fine with the strings :)

pazos:
@Frenzie: this bumps the APK to API30 and requires the new Docker image 0.4.0 to build. Gradle will complain about missing toolchains with the old image.

Frenzie:
@pazos Ping me after you merge it; I'll update the line on GitLab then.

PR #6921, opened 2020-11-28
Get rid of ui/data/strings.lua

jperon:
As suggested [here](https://github.com/koreader/koreader/pull/6885#issuecomment-734865881)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6921)


frontend/ui/data/koptoptions.lua
```diff
@@ -297,7 +298,7 @@ left to right or reverse, top to bottom or reverse.]]),
                 enabled_func = function(configurable)
                     return optionsutil.enableIfEquals(configurable, "text_wrap", 1)
                 end,
-                labels = {S.AUTO, S.LEFT, S.CENTER, S.RIGHT, S.JUSTIFY},
+                labels = {_("auto"), _("left"), _("center"), _("right"), _("justify")},
```
Frenzie, line 301: Perhaps we should add contexts to some of these while we're in here. E.g., center can mean both "to center" and "middle" depending on context. My unofficial policy of not adding contexts unless a problem is known is about existing code, but if you're editing things anyway that's a slightly different matter.

```suggestion
                labels = {
                    C_("Alignment", "auto"),
                    C_("Alignment", "left"),
                    C_("Alignment", "center"),
                    C_("Alignment", "right"),
                    C_("Alignment", "justify"),
                },
```

frontend/ui/data/koptoptions.lua
```diff
@@ -422,8 +424,8 @@ This can also be used to remove some gray background or to convert a grayscale o
             },
             {
                 name = "quality",
-                name_text = S.RENDER_QUALITY,
-                toggle = {S.LOW, S.DEFAULT, S.HIGH},
+                name_text = _("Render Quality"),
+                toggle = {_("low"), _("default"), _("high")},
```
Frenzie, line 428: Context candidate for "Quality."

Frenzie (commented):
Cheers. :+1:

Frenzie (approved):
lgtm

poire-z (approved):

PR #6927, opened 2020-11-30
[chore, i18n] Remove extra quotation marks

Frenzie:
@jperon Did my review comment about this get lost? :-/

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6927)


jperon:
Sorry, didn't see it :(

NiLuJe:
Yeah, inline comments have a habit of getting lost in the ether if you keep the PR tab open.

It's become so annoying that I usually just check the email notification, because they're sure to be there in there.

Frenzie:
Actually perhaps I should be using Reviewable more, it being better at not randomly losing/hiding junk like GH itself does for some incomprehensible reason is a major reason I wanted to experiment with it.

PR #6928, opened 2020-11-30
[chore] Don't translate numbers

Frenzie:
Except potentially the likes of 0.7 vs. 0,7 but I'm not convinced that should be done by bothering translators.

Cross-ref to <https://github.com/koreader/koreader/pull/6885>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6928)

### DIFF
--- a/frontend/ui/data/koptoptions.lua
+++ b/frontend/ui/data/koptoptions.lua
@@ -165,7 +165,7 @@ In 'semi-auto' and 'manual' modes, you may need to define areas once on an odd p
                 show_true_value_func = function(str)
                     return string.format("%.1f", str)
                 end,
-                toggle =  {_("0.7"), "1", _("1.5"), "2", "3", "5", "10", "20"},
+                toggle =  {"0.7", "1", "1.5", "2", "3", "5", "10", "20"},
                 more_options = true,
                 more_options_param = {
                     value_step = 0.1, value_hold_step = 1,
--- a/frontend/ui/data/koptoptions.lua
+++ b/frontend/ui/data/koptoptions.lua
@@ -140,7 +140,7 @@ In 'semi-auto' and 'manual' modes, you may need to define areas once on an odd p
                 show_true_value_func = function(str)
                     return string.format("%.1f", str)
                 end,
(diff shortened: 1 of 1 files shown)


frontend/ui/data/koptoptions.lua
```diff
@@ -165,7 +165,7 @@ In 'semi-auto' and 'manual' modes, you may need to define areas once on an odd p
                 show_true_value_func = function(str)
                     return string.format("%.1f", str)
                 end,
-                toggle =  {_("0.7"), _("1"), _("1.5"), _("2"), _("3"), _("5"), _("10"), _("20")},
+                toggle =  {_("0.7"), "1", _("1.5"), "2", "3", "5", "10", "20"},
```
Frenzie, line 168: @NiLuJe Thoughts?

poire-z:
> Except potentially the likes of 0.7 vs. 0,7 but I'm not convinced that should be done by bothering translators.

Not convinced either. We have numbers (and formatting with a period) everywhere, so people must be used to see regular numbers, that way.

As for `_("3")`, I would agree to remove them.
Otherwise, we have some work to support [eastern arabic numerals](https://en.wikipedia.org/wiki/Eastern_Arabic_numerals) everywhere :) Pinging @WaseemAlkurdi for opinions: will far far eastern arabs be fine with our shared western arabic numbers ? :)

Frenzie:
> Not convinced either. 

Thanks, I'll go ahead and merge this quickly then so that the rest of the translators aren't bothered by this like the first half dozen.

NiLuJe:
I tend to vastly prefer the `.` as a decimal separator, even though it's supposed to be `,` over here, so I'm admittedly a bit of an odd duck on that front ;).

TL;DR: Yes ^^.

WaseemAlkurdi:
@poire-z Apologies, I'm too late ... 
Yes, if it takes too much effort to implement eastern Arabic numerals, then western Arabic (i.e. Arabic) numerals would be just fine, both systems are used around here.
I'd be curious to hear what speakers of Farsi (Persian) think about this, since they use the eastern numerals much more than speakers of Arabic or Kurdish.

PR #6930, opened 2020-11-30
Fix #6929

### LINKED ISSUE #6929: Crash when tapping on image
* KOReader version: v2020.11-25-g4ed9f19b_2020-11-29 (after #6885)
* Device: Tolino

#### Issue
After opening an image PNG, JPG and tap on the right or left I get a crash.
It seems that this bug was introduced with #6885

#### Steps to reproduce
see above

[log.txt](https://github.com/koreader/koreader/files/5614710/log.txt)

jperon:
A check was missing on a flag not defined for images.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6930)

### DIFF
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -844,7 +844,8 @@ function ReaderPaging:onGotoPageRel(diff)
     logger.dbg("goto relative page:", diff)
     local new_va = self.visible_area:copy()
     local x_pan_off, y_pan_off = 0, 0
-    local right_to_left = (self.ui.document.configurable.writing_direction > 0)
+    local right_to_left = self.ui.document.configurable.writing_direction
+                        and self.ui.document.configurable.writing_direction > 0
     local bottom_to_top = self.ui.zooming.zoom_bottom_to_top
     local h_progress = 1 - self.ui.zooming.zoom_overlap_h / 100
     local v_progress = 1 - self.ui.zooming.zoom_overlap_v / 100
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -844,7 +844,8 @@ function ReaderPaging:onGotoPageRel(diff)
     logger.dbg("goto relative page:", diff)
     local new_va = self.visible_area:copy()
     local x_pan_off, y_pan_off = 0, 0
-    local right_to_left = (self.ui.document.configurable.writing_direction > 0)
+    local right_to_left = self.ui.document.configurable.writing_direction
+                        and self.ui.document.configurable.writing_direction > 0
```
Frenzie, line 848: Alignment would look nicer like this. But no objection to sticking it on one line on my part, probably clearer if anything.
```suggestion
                          and self.ui.document.configurable.writing_direction > 0
```

jperon: Even though it makes the line longer than 120 characters (usual limit in Lua) ?

Frenzie: A line that's more than some 100-120 characters is a rough indicator that there might be too much nested complexity. Other than that line length has no particular significance.

jperon: OK. Till now, I thought the main reason for that was to avoid lines getting out the screen in editors.

Frenzie (commented):
I think you included something with luajit-launcher by accident.

jperon:
I had thought about it the commit just before… :)

PR #6931, opened 2020-11-30
JPG/PNG: MuPDF as default provider

jperon:
As said in #6929

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6931)

### DIFF
--- a/frontend/document/picdocument.lua
+++ b/frontend/document/picdocument.lua
@@ -50,9 +50,9 @@ end
 
 function PicDocument:register(registry)
     registry:addProvider("gif", "image/gif", self, 100)
-    registry:addProvider("jpg", "image/jpeg", self, 100)
-    registry:addProvider("jpeg", "image/jpeg", self, 100)
-    registry:addProvider("png", "image/png", self, 100)
+    registry:addProvider("jpg", "image/jpeg", self, 80)
+    registry:addProvider("jpeg", "image/jpeg", self, 80)
+    registry:addProvider("png", "image/png", self, 80)
 end
 
 return PicDocument


jperon:
I'd just like to mention something I saw while trying, which IMO is a weird bug:

- in file manager, open an image: it opens within MuPDF;
- back to file manager, long press -> "Open with…" -> "Picture Document": it opens within PicDocument;
- back to file manager, either single click, or long press -> "Open with…" -> "MuPDF": it opens within PicDocument, not MuPDF.

Frenzie:
Probably some issue with the PicDocument close method. As @poire-z said, we (or at the very least us two) don't really test it. :-P

NiLuJe:
Does this only happen between Pic/MuPDF, or also between CRe/MuPDF for ePubs?

(Wild guess: pairs vs. ipairs shenanigans again?).

jperon:
> Does this only happen between Pic/MuPDF, or also between CRe/MuPDF for ePubs?

After a quick test, it doesn't seem so.

frontend/document/picdocument.lua
```diff
@@ -50,9 +50,9 @@ end
 
 function PicDocument:register(registry)
     registry:addProvider("gif", "image/gif", self, 100)
-    registry:addProvider("jpg", "image/jpeg", self, 100)
-    registry:addProvider("jpeg", "image/jpeg", self, 100)
-    registry:addProvider("png", "image/png", self, 100)
+    registry:addProvider("jpg", "image/jpeg", self, 80)
```
Frenzie, line 53: In my mind 100 means "fully intended to open" so saying MuPDF is only 90 would be saying (to the code/maintainers) that there's still something wrong with it. I was just a bit shy about it when I added all those new picture formats I suppose. Anyway, I'll merge it like this since that's more of a separate issue.

NiLuJe:
Possibly MuPDF and/or PicDocument polluting the sidecar config and providing confusing settings?

NiLuJe:
I'm seeing for instance weird stuff going on with the "Overlap" menu being available+disabled or unavailable+enabled depending on which engine is used first...

PR #6935, opened 2020-12-01
kodev: Make the directory finding code work as intended (probably?)

NiLuJe:
It was blowing up when find found multiple matches (c.f., gitter).

Warning: requires bash 4.4. It's not in Ubuntu 16.04, it is in 18.04. Is that good enough?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6935)

### DIFF
--- a/kodev
+++ b/kodev
@@ -107,7 +107,7 @@ function setup_env() {
     fi
     local files=()
     while IFS=  read -r -d $'\0'; do
-        files+=("$REPLY")
+        files+=("${REPLY}")
     done < <(find . -maxdepth 1 -name 'koreader-emulator-*' -print0 | ${SETUP_ENV_GREP_COMMAND})
     assert_ret_zero $? "Emulator not found. Please build it first."
     local idx=0
@@ -116,7 +116,7 @@ function setup_env() {
         echo "Multiple emulator builds found:"
         local ts=()
         # Store list of ts at the same index
-        for i in ${!files[@]}; do
+        for i in "${!files[@]}"; do
             local file="${files[${i}]}/koreader"
             if [ -d "${file}/koreader" ]; then
                 echo "${file} (last modified on $(date -c %y "${file}"))"
@@ -127,7 +127,7 @@ function setup_env() {
         local sorted_ts=()
         IFS=$'\n' read -d '' -r -a sorted_ts < <(printf '%s\n' "${ts[@]}" | sort -r)
         # Find the id of the most recent ts
-        for i in ${!ts[@]}; do
+        for i in "${!ts[@]}"; do
             if [ "${ts[${i}]}" == "${sorted_ts[0]}" ]; then
                 idx="${i}"
                 break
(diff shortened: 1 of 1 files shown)


Frenzie (commented):
> Warning: requires bash 4.4. It's not in Ubuntu 16.04, it is in 18.04. Is that good enough?

Not in the current Docker image. What specifically requires bash 4.4?

Edit: link to somewhere in the conversation for posterity: https://gitter.im/koreader/koreader?at=5fc697ab32334215e6ec4a72

kodev
```diff
@@ -99,15 +99,18 @@ EOL
...
         # for android adb install
         KODEBUG_SUFFIX=-debug
     fi
-    files=$(find . -maxdepth 1 -name 'koreader-emulator-*' | ${SETUP_ENV_GREP_COMMAND})/koreader
+    readarray -d '' files < <(find . -maxdepth 1 -name 'koreader-emulator-*' -print0 | ${SETUP_ENV_GREP_COMMAND})
     assert_ret_zero $? "Emulator not found. Please build it first."
-    EMU_DIR=${files[0]}
+    # Warn if multiple matches were found
+    test ${#files[@]} -eq 1
+    assert_ret_zero $? "Multiple emulator builds found! Pare that down to one."
```
Frenzie, line 112: I don't like this. I'd rather have a forced select or something.

NiLuJe:
Now 200% more over-engineered :D.

NiLuJe:
(To answer the compat question: `readarray` is 4.4+).

kodev
```diff
@@ -105,15 +105,38 @@ function setup_env() {
...
+        for i in ${!files[@]}; do
+            local file="${files[${i}]}/koreader"
+            if [ -d "${file}/koreader" ]; then
+                echo "${file} (last modified on $(date -c %y "${file}"))"
+                ts[${i}]="$(date -c %Y "${file}")"
+            fi
+        done
+        # Sort the list of ts
+        local sorted_ts=()
+        IFS=$'\n' read -d '' -r -a sorted_ts < <(printf '%s\n' "${ts[@]}" | sort -r)
```
Frenzie, line 129: Probably smarter than my suggested prompt. :-D

kodev
```diff
@@ -105,7 +105,10 @@ function setup_env() {
         # for android adb install
         KODEBUG_SUFFIX=-debug
     fi
-    readarray -d '' files < <(find . -maxdepth 1 -name 'koreader-emulator-*' -print0 | ${SETUP_ENV_GREP_COMMAND})
+    files=()
+    while IFS=  read -r -d $'\0'; do
+        files+=("$REPLY")
+    done < <(find . -maxdepth 1 -name 'koreader-emulator-*' -print0 | ${SETUP_ENV_GREP_COMMAND})
```
Frenzie, line 111: The only thing I mean by not in the current Docker image is that any PR that requires a change to it can't be merged, and of course implicitly that as a rule of thumb I prefer not to change it but it needs the occasional maintenance anyway. ;-)

NiLuJe: Oh, no worries, this is exactly why I asked: I had the more verbose but more compatible code on hand, I just wanted to check if we could avoid it ;).

Issue #3459, opened 2017-11-01
Large >10MB mobi and epub files making KOreader slow


tejasvi8874:
* KOReader version: Nightly
* Device: Kindle ereader

Opening and reading books which are large increases the response time of reader drastically. Although this not happens in Amazon reader which although take long to open large and highly compressed books but afterwards skimming and navigation is smooth. Although it's not the case with pdf.

#### Steps to reproduce
Open large epub or mobi file.

jperon:
Same on my Kobo mini; large epub files take ages to open (or even crash KOReader).

Frenzie:
The [latest release](https://github.com/koreader/koreader/releases/tag/v2018.02.12-nightly) includes two potential workarounds.

* The option to disable compression of crengine cache, see https://github.com/koreader/koreader/pull/3670
* With the new <kbd>Open with…</kbd> functionality you could try if MuPDF opens it acceptably, see https://github.com/koreader/koreader/pull/3653

jperon:
I've already tried those options; KOReader crashes anyway.

Frenzie:
Sorry, I overlooked the fact that you said crash. If it crashes that might be because the Kobo Mini has a meager 256 MB of RAM. What do you mean by "large" concretely? Although I suppose it's mostly about number of elements or something along those lines, even though you would need sufficient memory for decompression as well. 

I've got a simple 130 MB EPUB here which uses some 58 MB of RAM, and a more complex 1.8 MB EPUB that uses about 87 MB. But almost nothing in my possession takes more than a few seconds that I'm aware of. I do have this book *Classic Shell Scripting*, published by O'Reilly, that clocks in at a mere 1.7 MB but takes quite a while to load and uses 100 MB. In MuPDF it's almost twice as slow to load and it uses even more RAM (~160 for the EPUB vs ~60 for the PDF).

(Of course my H2O is slightly more powerful, but only my desktop could realistically load a file in a split second even though it'd take over 10 seconds on a device.)

All that being said, I'm not really the right person here. For me EPUB is like a nice bonus. PDF and DjVu are what I want. :-) I'm just trying to help pinpoint the cause of the problem better.

jperon:
I mean basically everything over 10MB: for example, I have [sources](https://github.com/DivinumOfficium/divinum-officium/) that can be compiled either to one epub, or to twelve parts (each for one month). The big one is about 13MB, the small ones about 1.3MB each. The big one crashes KOReader, not the small ones.
[...]

Issue #6896, opened 2020-11-20
FR: Need A Paragraph Change Preview Page 


TaChao:
Every change of paragraph would re-rander the whole document, It will take 10-15 mins for a new book setup.
Please add a page for this change, and  preview for multiple modifications。

NiLuJe:
I'll let other people explain why this is not going to be possible because that's just not how CRe works ;p.

(The only remotely feasible thing would be to be able to enable to tweak multiple things *without* triggering a CRe update, *then* hit an "I'm done" button to batch all the changes in one go. That's... also extreely unlikely, because way, way not how things are implemented).

poire-z:
Closing in favor of #9087 (where I wrote a bit more).

PR #6936, opened 2020-12-01
[RFC] [WIP] [POC] Open big epub on low-end device

jperon:
With my good old kobo mini, I still experiment this issue with big epubs: #3459

This draft is an attempt to unzip, then read an epub on a file-by-file basis. As is, it's very dirty:

- the history get polluted;
- no TOC;
- etc.

But well, it lets one open such epubs on the mini…

This is why I open it as a draft, to have room for discussions and ideas.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6936)

### DIFF
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -603,14 +603,19 @@ function ReaderLink:onGotoLink(link, neglect_current_location, allow_footnote_po
             if anchor then
                 display_filename = display_filename .. anchor
             end
-            UIManager:show(ConfirmBox:new{
-                text = T(_("Would you like to read this local document?\n\n%1\n"), BD.filepath(display_filename)),
-                ok_callback = function()
-                    UIManager:scheduleIn(0.1, function()
-                        self.ui:switchDocument(linked_filename)
-                    end)
-                end
-            })
+            require("logger").dbg("ASU prov", self.document.provider)
+            if self.document.provider ~= "bigepub" then
+                UIManager:show(ConfirmBox:new{
+                    text = T(_("Would you like to read this local document?\n\n%1\n"), BD.filepath(display_filename)),
+                    ok_callback = function()
+                        UIManager:scheduleIn(0.1, function()
+                            self.ui:switchDocument(linked_filename)
+                        end)
+                    end
+                })
+            else
+                self.ui:switchDocument(linked_filename)
+            end
         else
             UIManager:show(InfoMessage:new{
                 text = T(_("Link to unsupported local file:\n%1"), BD.url(link_url)),
--- a/frontend/apps/reader/modules/readerstatus.lua
+++ b/frontend/apps/reader/modules/readerstatus.lua
@@ -41,6 +41,10 @@ function ReaderStatus:addToMainMenu(menu_items)
 end
 
 function ReaderStatus:onEndOfBook()
+    if self.document.provider == "bigepub" then
+        self.ui:switchDocument(self.document:nextItem(), "bigepub")
+        return true
+    end
     Device:performHapticFeedback("CONTEXT_CLICK")
     local settings = G_reader_settings:readSetting("end_document_action")
     local choose_action
--- a/frontend/document/bigepub.lua
+++ b/frontend/document/bigepub.lua
@@ -1,0 +1,237 @@
+--[[--
+This is a debug plugin to allow opening of big epubs,
+especially on low-end devices.
+
+@module koplugin.BigEpub
+--]]--
+
+
+local CreDocument = require("document/credocument")
+local PdfDocument = require("document/pdfdocument")
+local DocSettings = require("docsettings")
+local Document = require("document/document")
+local Event = require("ui/event")
+local InfoMessage = require("ui/widget/infomessage")
+local UIManager = require("ui/uimanager")
+local ffi = require("ffi")
+local ffiUtil = require("ffi/util")
+local luxl = require("luxl")
+local util = require("frontend/util")
+local _ = require("gettext")
+
+
+local unescape_map  = {
+    ["lt"] = "<",
+    ["gt"] = ">",
+    ["amp"] = "&",
+    ["quot"] = '"',
+    ["apos"] = "'"
+}
+
+local gsub = string.gsub
+local function unescape(str)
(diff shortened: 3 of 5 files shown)


Frenzie:
I think you forgot to add a file?

poire-z:
(Missing some code I think.)

But I don't really like it: it's too dirty, and you'd have to add kludge in many places to make it less dirty, if that's possible.
Like changeing font size on a document fragment, won't apply to all the others, so you'd have to do it manually on each of them - or set some as default, but then, you'll have to `Purge .sdr` on the ones you previously opened. Good luck solving that properly :)
So, it's a rare use case (really big epub on low memory device) where KOreader won't be really friendly and won't be able to hide the trick used.

I'd rather just have an (hidable) option to filemanager hold menu when long-press on an EPUB: <kbd>Unzip</kbd> :)

And does crengine work ok with such exploded EPUBs ? Is it able to resolved CSS and images links, and to follow links (to footnotes)? I think it might, I had tested a few things like that when testing with HTML test files.
But also, I guess in-page or popup footnotes are out of the horizon. Limited experience - or added kludge to solve it :(

Frenzie:
I agree that an approach like this is very much a kludge, and the only real solution would be to have crengine do a more on-demand style rendering.

Incidentally MuPDF 1.17 introduced support for that:

> api: Chapter based API for faster EPUB loading. 

Of course someone will still have to implement that in the frontend, but as an fyi.

jperon:
Sorry for the forgotten file, fixed.

For sure, this approach is a hack. What I thought would be using lua's metatable and __index mechanisms to avoid re-inventing the wheel, while (if possible) handling the raised points (settings, fonts and so on). But as I'm very young in KOReader development, there are probably many points that I don't see!

jperon:
> I agree that an approach like this is very much a kludge, and the only real solution would be to have crengine do a more on-demand style rendering.
> 
> Incidentally MuPDF 1.17 introduced support for that:
> 
> > api: Chapter based API for faster EPUB loading.
> 
> Of course someone will still have to implement that in the frontend, but as an fyi.

Why not, but I must confess that, each time I tried to read epubs with MuPDF, I wasn't convinced. For example, the epubs I'd like to handle (Summa and Breviarium, mentionned in #3459) have many links, and I didn't see how to follow them with MuPDF…

jperon:
@poire-z The docs I tested only involved css and links (no images), but I didn't see any problem with them. Didn't try in-page or pop-up footnotes either.

jperon:
> So, it's a rare use case (really big epub on low memory device) where KOreader won't be really friendly and won't be able to hide the trick used.

That said, the "showing quick, loading after / in background" approach (however implemented) would have other benefits, like fixing issues such as #6896.

> I'd rather just have an (hidable) option to filemanager hold menu when long-press on an EPUB: Unzip :)

I thought about this too, but for example, I have an epub with 1536 files, and index about in the middle… So it would need at least some renaming!

NiLuJe:
MuPDF also implemented a new ePub API which caches some stuff to disk to make further re-rendering faster (in addition to the finer-grained API @Frenzie mentioned).

poire-z:
> That said, the "showing quick, loading after / in background" approach (however implemented) would have other benefits, like fixing issues such as #6896.

I had thought about that, but the way crengine works, I can't see any hook where that could start being implemented.
Most obvious idea is creating another internal document and just giving it a single fragment and let it work with that - but the interfacing with frontend would be complicated. So, it would just be a gimmick: _we're giving you a splash screen so you have something to look at while we go on processing the whole book in the background - but you still have to wait._
Too much work for the false sense of confort it would give :) 
And once you know how it works, I'm fine on a new book spending a few minutes playing with style tweaks and waiting 40s for each rendering - once done, that time spend is worth the next 20 hours of happy reading :)

May be something for you to test with one of your big books
- load it on the emulator, tweak what you need, exit.
- create and copy the book.sdr/metadata.epub.lua on your device
- copy the crengine cache file from the emulator koreader.../cache/cr3cache/ (blah.epub.c016eb2f.1.cr3 and possibly cr3cache.inx - make a backup of your original cr3cache in case you have big books already cached and you don't want to lose the time spent making them :)
- open that book in koreader: you may have a re-rendering (because WxH changed, but you might bypass the XML loading/HTML parsing phase)

That's not a solution - just an experiment to see how much the XML/HTMLparsing time/mem/power needed vs the re-rerendering of something already parsed do. Might give some ideas, or not :)

jperon:
For MμPDF 1.17, unless I'm mistaken, it is necessary to wait until [this work](https://github.com/koreader/koreader-base/pull/1203) is finished?

Frenzie:
> For MμPDF 1.17, unless I'm mistaken, it is necessary to wait until this work is finished?

Advisable perhaps, necessary no. You can already work on top of https://github.com/koreader/koreader-base/pull/1203 and https://github.com/koreader/koreader/pull/6738 on your PC.

Frenzie:
> For example, the epubs I'd like to handle (Summa and Breviarium, mentionned in #3459) have many links, and I didn't see how to follow them with MuPDF…

I don't quite understand what you mean. You just tap on them, same as in crengine/PDF/DjVu?

jperon:
> > For example, the epubs I'd like to handle (Summa and Breviarium, mentionned in #3459) have many links, and I didn't see how to follow them with MuPDF…
> 
> I don't quite understand what you mean. You just tap on them, same as in crengine/PDF/DjVu?

Just checked, and while it does work on most epubs, it doesn't work with this one (renamed to `.zip` for upload):

[Uploading Saint Thomas d'Aquin - Somme Théologique RDJ.zip…]()

Frenzie:
Curious, also still an issue in 1.17. I suspect a document issue is more likely than MuPDF as such but we'll need a significantly more minimal testcase first. ;-)

jperon:
> … but we'll need a significantly more minimal testcase first. ;-)

:-) Is it better with this one?

[RDJ.zip](https://github.com/koreader/koreader/files/5630424/RDJ.zip)

Frenzie:
Not quite a minimal testcase, but close enough. :-)

Frenzie:
Slightly more minimized attached. Still a lot of distractions. :-)
[RDJ-1.epub.zip](https://github.com/koreader/koreader/files/5630855/RDJ-1.epub.zip)

poire-z:
Closing as inactive and probably a bit too crazy idea :)

PR #6946, opened 2020-12-04
Fix the frontlight gesture range scaling

NiLuJe:
That seems like an obvious mistake?

xref https://www.mobileread.com/forums/showthread.php?t=335341

(ping @zwim)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6946)


NiLuJe:
(Regression since #6468)

NiLuJe:
(I'm testing this on a !warmth device, though).

zwim:
@NiLuJe: Yes, you have fixed it. It works fine on frontlight and warmth scale on the Tolino.

zwim (approved):

PR #6937, opened 2020-12-02
Icon update using MaterialDesignLight

johnbeard:
Much lighter, simpler and less complex icons.

Uses Austin Andrews' (Templarian)'s MaterialDesignLight
icon theme as a base with modififications for size and
content.

Icons licensed under the SIL Open Font License 1.1
and Apache License 2.0

-----
Demo:

2020-12-02_034410_728x932_screenshot

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6937)

PR #6950, opened 2020-12-04
RenderImage: add renderSVGImageFile()

poire-z:
Adds `RenderImage:renderSVGImageFile()` with 2 implementations:
- one using MuPDF (our current MuPDF 1.13 is not really good at it)
- the other using NanoSVG, via https://github.com/koreader/koreader-base/pull/1250, which is quite good (not really with our current SVG icons, but fine with those from #6937)

ImageWidget: allow picking SVG versions of provided PNG icons if available. To use, just drops SVG versions with the same name of the PNG we use in another directory, and set this variable in `frontend/ui/widget/imagewidget.lua`:
```lua
local ICONS_ALT_SVG_DIR
-- Uncomment to use SVG icons from one of these directories
-- ICONS_ALT_SVG_DIR = "resources/icons/src/"
-- ICONS_ALT_SVG_DIR = "resources/icons/svg/"
```

See #6937.

This should be a no-op if we keep `ICONS_ALT_SVG_DIR` blank.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6950)

### DIFF
--- a/frontend/ui/renderimage.lua
+++ b/frontend/ui/renderimage.lua
@@ -2,12 +2,15 @@
 Image rendering module.
 ]]
 
+local Blitbuffer = require("ffi/blitbuffer")
+local Math = require("optmath")
 local ffi = require("ffi")
 local logger = require("logger")
 
 -- Will be loaded when needed
 local Mupdf = nil
 local Pic = nil
+local NnSVG = nil
 
 local RenderImage = {}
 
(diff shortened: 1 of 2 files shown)


Galunid (approved):
Looks good, one question I have is what happens when no width and height is provided in `RenderImage:renderSVGImageFile`? Does it render image full screen with image ratio potentially disturbed? It's more of a question how Mupdf/NanoSVG handle that.

Frenzie:
I would expect the original size as specified by the image file itself, at least for MuPDF.

Frenzie (approved):
lgtm

poire-z:
> what happens when no width and height is provided in RenderImage:renderSVGImageFile? Does it render image full screen with image ratio potentially disturbed? It's more of a question how Mupdf/NanoSVG handle that.

Well, I haven't tested all combinations really much, and mostly only with squared icons :)
And there's 2 layers of scaling in ImageWidget: on load (scale for DPI is applied there), and on rendering/drawing (scale factor, stretching). I haven't really gotten back into all that code :/ We'll see how it goes and fix what's needed.

But there are 2 use cases that I know of for icons:
- we provide width and height: we'll draw the SVG icon so it fits in that W x H. Both MuPDF and NanoSvg work the same: we need to provide a zoom (same for x and y axis), and provide the target W x H (as values for MuPDF, via a pre-made BB for NanoSVG) where they will draw the native SVG with that zoom into. So, we need to compute the zoom and the target W x H, and some offset if needed to have the native SVG scaled with that zoom and centered in the target W x H. (I didn't do the offset stuff for MuPDF, I thought it was centering it, but I'm not sure it does now.)
- the other use case is we don't provide width and height: the current code, for PNG and other regular images, will take the image's native size - and will scale it for dpi/size (`get_dpi_scale()` at the top of `imagewidget.lua`, dunno if it's identical to `scaleBySize()`). We try to do the same thing: use the native SVG dimensions, use that `DPI_SCALE` as the zoom value. target W x H is just native `W*DPI_SCALE x H*DPI_SCALE`.

That is I think all I tested, looking (but not that much :/) at the logs that the 2nd ImageWidget scaling step isn't doing anything - so we'll see.

So, no: no "full screen" scaling, and it looks that neither MuPDF nor NanoSVG can distort the aspect ratio (if that happens, it's ImageWidget's 2nd step :/).

Issue #6875, opened 2020-11-13
FR: Add a downloadable Mac application bundle to the Releases page


JasonInOttawa:
Is it possible to add a downloadable Mac application bundle to the Releases page?

pazos:
I tried with https://www.macstadium.com/opensource. They granted us a free mac mini server on their farms, but the register process involved giving my credit card, so nope.

I don't know if is possible to build the app/dmg using github actions. As far is I'm concerned I'm aware of the situation of opensource enthusiast who own a mac: play stupid games, win stupid prizes :smile:

JasonInOttawa:
Thanks very much for looking into it.

Frenzie:
@pazos See https://github.com/koreader/koreader-base/blob/ffdb9be5e626adbdc5432e1227aee7ddab1c127a/.github/workflows/build.yml which already does it (except I need to double check if I set up the cache quite right).

Also cross-ref to time-limited discussion in <https://github.com/koreader/koreader-base/pull/1198#issuecomment-706888406>. 

Relevant action:
https://github.com/marketplace/actions/upload-a-build-artifact

pazos:
@Frenzie: thanks!

I'm very bad at deploying things, but looks like we need a similar action on the frontend that is triggered on tagged releases and just run
```
run: export MACOSX_DEPLOYMENT_TARGET=10.14 PATH="/usr/local/opt/gettext/bin:$PATH" && ./kodev release macos
```

?.

Also homebrew deps need to be tweaked a bit: sdl2 can be removed (we're building that from source) and `create-dmg` package added.

I *think* we need the dmg step, as macs out there won't like to install non-notarized apps packaged inside a container, like 7z.

Anyways, I built latest version yesterday, on Big Sur, and compressed it to see if it works: https://www.dropbox.com/s/d5doikl6fz821sf/KOReader.7z?dl=0

@JasonInOttawa, could you test if the app works?. It also has an easter egg which I'm going to submit later :)

Frenzie:
> just run […]

Yup, unless packaging requires anything extra that should all be the same. The only thing that'd be properly different would be the artifact thing.

However, it looks like the Mac CI build has broken itself again through no fault of ours…

JasonInOttawa:
@pazos It's been happily running for a couple hours now. Apart from the "unknown developer" warning everything is fine. I haven't seen the Easter egg yet.

pazos:
[...]

PR #6955, opened 2020-12-06
[CI] Mac OS app

pazos:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6955)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,34 @@
+name: test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [macos-latest]
+
+    steps:
+      - name: XCode version
+        run: xcode-select -p
+
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 200
+
+      - name: Homebrew install dependencies
+        # Compared to the README, adds ccache for faster compilation times and removes sh5sum to prevent some conflict with coreutils
+        # Compared to the emulator, adds p7zip and removes sdl2.
+        run: brew install ccache nasm ragel binutils coreutils libtool autoconf automake cmake makedepend lua@5.1 luarocks gettext pkg-config wget p7zip
+
+      - name: Building in progress…
+        run: export MACOSX_DEPLOYMENT_TARGET=10.14 PATH="/usr/local/opt/gettext/bin:$PATH" && ./kodev release macos
+
+      - name: Uploading artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: osx-artifacts
+          path: '*.7z'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v2
         with:
-          fetch-depth: 200
+          fetch-depth: 0
 
       - name: Homebrew install dependencies
         # Compared to the README, adds ccache for faster compilation times and removes sh5sum to prevent some conflict with coreutils
--- a/datastorage.lua
+++ b/datastorage.lua
@@ -18,7 +18,8 @@ function DataStorage:getDataDir()
         -- confined ubuntu app has write access to this dir
         data_dir = string.format("%s/%s", os.getenv("XDG_DATA_HOME"), package_name)
     elseif os.getenv("APPIMAGE") or os.getenv("KO_MULTIUSER") then
-        data_dir = string.format("%s/%s/%s", os.getenv("HOME"), ".config", "koreader")
+        local user_rw = jit.os == "OSX" and "Library/Application Support" or ".config"
+        data_dir = string.format("%s/%s/%s", os.getenv("HOME"), user_rw, "koreader")
     else
         data_dir = "."
     end
--- a/platform/mac/do_mac_bundle.sh
+++ b/platform/mac/do_mac_bundle.sh
@@ -185,15 +185,10 @@ for path in l10n/*; do
     fi
 done
 
-# package as DMG if create-dmg is available
-# reduces size from 80MB to 40MB
 mv "${APP_PATH}" "${APP_BUNDLE}.app"
 
-if command_exists "create-dmg"; then
-    # create KOReader-$VERSION.dmg with KOReader.app inside
-    create-dmg "${APP_BUNDLE}.app" --overwrite
-    rm -rf "${APP_BUNDLE}.app"
-else
-    # rename as KOReader-$VERSION.app
-    mv -v "${APP_BUNDLE}.app" "${APP_BUNDLE}-${VERSION}.app"
+# package as 7z reduces size from 80MB to 30MB
+if command_exists "7z"; then
+    7z a -l -m0=lzma2 -mx=9 "${APP_BUNDLE}-${VERSION}.7z" "${APP_BUNDLE}.app"
+    rm -rfv "${APP_BUNDLE}.app"
 fi


Frenzie (approved):
Sure. :-)

pazos:
wooops. `$VERSION` doesn't work, so `git-rev` is empty. No idea how to fix that :/

NiLuJe:
Which part doesn't work, exactly?

pazos:
@NiLuJe https://github.com/koreader/koreader/blob/master/kodev#L11

NiLuJe:
Is this a shallow clone, perhaps?

Frenzie:
What's the error message? But yeah, best to do at least 100 or 200 deep if
you want that to work.

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
...
+        os: [macos-latest]
+
+    steps:
+      - name: XCode version
+        run: xcode-select -p
+
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
Frenzie, line 20: You don't need 0, something like 100 (and tags if necessary) will likely do, probably better 200 to play it safe. Our total number of commits is close to 10000, but by limiting it to what's likely to happen within a couple of months you still get most of the speed advantages of a shallow clone. (I.e., it'll take ~1 second as opposed to ~4. Maybe not worth the optimization when the build itself takes over 10 minutes, but just fyi anyway. :-P)

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Cache for ccache
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-ccache
+        with:
+          path: ~/.ccache # ccache cache files are stored in `~/.ccache` on Linux/macOS
+          key: ${{ runner.os }}-build-${{ env.cache-name }}
```
Frenzie, line 28: If you want to run this on every PR, you should cache base since that's only updated occasionally. See my CircleCI config.

https://github.com/koreader/koreader/blob/c20ad8f5e9385dd18138a306b874c5b20302f299/.circleci/config.yml#L20-L26

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
...
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
```
Frenzie, line 32: Incidentally, if you copied this from base it's probably not quite right yet. But with the macOS build constantly breaking itself I haven't really been able to look into it.

.github/workflows/build.yml
pazos: I think running this on tagged releases is enough for https://github.com/koreader/koreader/issues/6875. No idea how to do that, how artifacts work and, generally speaking, the work involved on each release. So I would be happy if you could take care of it and remain ignorant :)

.github/workflows/build.yml
pazos: Yup, totally ripped from base. Remember I have no idea of what I'm doing here :p

.github/workflows/build.yml
pazos: Happy to change it :+1:

pazos:
@Frenzie: I don't speak much git, github, automatisms or build systems.  It is all a bit magic to me :smiley: 

I was trying to address your comment in https://github.com/koreader/koreader/issues/6875#issuecomment-727204973.

.github/workflows/build.yml
Frenzie: In that case you can add a condition like https://github.com/hellerbarde/stapler/commit/48d2dc7e7a683ca3c2b659c1762aa67661f7ad79 

For tags/releases only we don't need to cache base.

It can be slightly annoying to test, but it's not too much of a bother on your own fork.

Anyhow, I can also do that in a follow-up PR some other time.

.github/workflows/build.yml
Frenzie: It's the same general idea as in CircleCI. The cache is a never-changing thing (maybe it's deleted if it's not used for a few months?), so to generate an up to date cache it needs to be sufficiently unique.

https://github.com/koreader/koreader/blob/c20ad8f5e9385dd18138a306b874c5b20302f299/.circleci/config.yml#L19

The problem is that I don't think `${{ runner.os }}-build-${{ env.cache-name }}-` has actually changed since I added the macOS build. So the build time is slowly increasing as the cache becomes increasingly unrelated to the repo.

Frenzie:
> I was trying to address your comment in #6875 (comment).

Oh sure, if it works Good Enough™ for the moment I'm fine with merging it as also indicated by the approval I left.

pazos:
@Frenzie


```
      - name: Check out Git repository
        uses: actions/checkout@v2
        with:
          fetch-depth: 200
```

doesn't work. As you can see in the artifacts we lost $VERSION again.

Frenzie:
By "again" you mean it worked at some point? If you don't have tags... ;-)

pazos:
> By "again" you mean it worked at some point?

It worked with `fetch-depth: 0`.
And by worked I mean the osx-artifacts.zip contained a valid version and a working program inside :100:

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    steps:
+      - name: XCode version
+        run: xcode-select -p
+
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 200
+
```
Frenzie, line 21: try adding:
```
run: git fetch --tags
```

Frenzie: Or just go with 0. It's just 3 seconds. ;-)

NiLuJe: (... and not our bandwidth. Mwhahaha.)

pazos:
@JasonInOttawa: could you test https://github.com/koreader/koreader/suites/1645100449/artifacts/30493747 on your pristine mac?. It seems to work fine on mine.

JasonInOttawa:
A quick download and run worked fine. I got an osx-artifacts.zip file, which contained a .7z file, which contained the application.

I did notice that it's not reading my settings - are they still stored in .config/koreader?

Thanks!

JasonInOttawa:
@pazos Found it in Application Support - very nice!

JasonInOttawa:
The test app has been happily running for a few hours now. I do have this computer set up for builds - could that make any difference?

pazos:
> The test app has been happily running for a few hours now. I do have this computer set up for builds - could that make any difference?

I tested the app in a pristine VM (without brew, xcode or commandline tools). It works fine there. The reason to ask for your feedback is because the VM lacks some security enhancements available on apple hardware (namely System Integrity Protection, aka SIP).

The only reason that prevents us from releasing a wide mac app (with a minimum target of 10.09) is that we're relying on some brew libraries and they're built with newer deployment targets.

So, yeah. It should work fine in Big Sur and Catalina and might work on Mojave too.

JasonInOttawa:
I'm running Big Sur. I did have to right-click - Open - Yes I'm sure as the app isn't notarized. No issues other than that.

Thanks!

pazos:
> I did have to right-click - Open - Yes I'm sure as the app isn't notarized. No issues other than that.

That's the gatekeeper, not SIP. Based on your report SIP issues were fixed with the osx-loader in https://github.com/koreader/koreader/pull/6448#issuecomment-665345741

About the gatekeeper issue. Notarization is a joke for a few reasons:

- Notarized apps still require the same workaround as non notarized apps in default osx setup (gatekeeper enforced)
- Notarized apps require a [hardened runtime](https://developer.apple.com/documentation/security/hardened_runtime). Easy to implement in XCode projects but a PITA elsewhere. Also a hardened runtime disables JIT so we need to add an entitlement to disable that specific protection, maybe others too.
- Notarized apps require a **paid developer account**. Lol

tl;dr: notarization is a marketing thing, unrelated to security, which severly restricts free development for a mac. SIP is a security feature, unrelated to marketing, that doesn't harm free development and it is a cool thing to have it enabled.

JasonInOttawa:
Got it.

It certainly would seem to be a good idea to waive the developer account fee for certain developers on projects that have been around for a while. If I had more pull at Apple, I'd suggest it :)

Looking forward to downloading the app from the Releases page one day!

PR #6957, opened 2020-12-07
FileManager: fix Plus menu override

yparitcher:
manage it in FM not in gestures, as  #6918 made the menu tap zone the entire width so the tap is not passed through to the gesture plugin

also do not allow changing the top right corner gesture in FM, as it does not do anything any more.

This is also in line with the home button handling.

regression caused by  #6918, touching the tap zones always manages to break something :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6957)


Frenzie (approved):

poire-z (approved):

NiLuJe:
Nice catch, I barely if ever use that button ;).

PR #6958, opened 2020-12-07
FileManager: don't load already loaded plugins

jperon:
`FileManager:init` is called by `FileManager:reinit`, which caused duplicate
loads. Those can't be useful (for example to re-load a modified plugin
on runtime), as `PluginLoader:loadPlugins` memoizes loaded plugins.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6958)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -479,19 +479,22 @@ function FileManager:init()
     table.insert(self, DeviceListener:new{ ui = self })
 
     -- koreader plugins
-    for _,plugin_module in ipairs(PluginLoader:loadPlugins()) do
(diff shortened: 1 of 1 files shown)


poire-z (approved):

Frenzie (commented):
Is this the same as in reader now? (Can't check atm)

poire-z:
I don't think we ever need to "reinit()" in Reader - so, the situation may never happen in Reader.

jperon:
> I don't think we ever need to "reinit()" in Reader - so, the situation may never happen in Reader.

You where 30 seconds faster than me… Just checked, no duplicate `RD loaded plugin…`.

Issue #6613, opened 2020-08-31
Android: Downloaded Dictionary is not in list


haymant:
* KOReader version: koreader-android-arm-v2020.08.1.apk
* Device: Android 10 on Huawei P30Pro

![android](https://user-images.githubusercontent.com/700512/91778755-1432b000-ec26-11ea-875f-0c7f63f35141.jpg)

#### Issue

I can download dictionary (e.g., the GNU english dictioanary), the log shows:
```
09-01 07:36:25.552 17918 17961 W KOReader: command ./tar xf /storage/emulated/0/koreader/data/dict/gcide.tar.gz -C /storage/emulated/0/koreader/data/dict returned 139
```
If I try again, we can see koreader knows the dictionary is downloaded
![install](https://user-images.githubusercontent.com/700512/91778790-2dd3f780-ec26-11ea-8f06-e1f09ead1474.jpg)

However, the dictionary does not appear in list.
![list](https://user-images.githubusercontent.com/700512/91778746-0c730b80-ec26-11ea-8be1-f1eff6a95bb0.jpg)

#### Steps to reproduce

1. Goto Dictionary Download menu, download an english dictionary
2. Back one level up
3. The downloaded dictionary is not in list

pazos:
> 09-01 07:36:25.552 17918 17961 W KOReader: command ./tar xf /storage/emulated/0/koreader/data/dict/gcide.tar.gz -C  returned 139

Maybe we need to extract dicts into the internal cache and copy the extracted files to /storage/emulated/0/koreader/data/dict? I never played with dict downloads on android (other than avoiding ANRs while downloading)

Frenzie:
@pazos Are you saying the dictionary itself still works, just not the built-in download?

pazos:
stardict dictionaries work fine, yeah. Not sure about the download but it seems to fail uncompressing.

haymant:
Android 10 didn't untar at all (i.e., unlike Android 8.1 device I tried, which got incomplete directory), I untared them and these dictionary got listed but then stardict returned "no definition found"

tryfon:
I have almost exactly the same issue. In my case the dictionary  was extracted successfully but is not listed. Huawei Media Pad T5, Android 8.0, koreader v2020.08.1
Please prioritize this or provide a workaround. Embedded dictionary is a quite essential feature.

NiLuJe:
The workaround is you unpacking the download manually ;).

There's not much we can do about Android actively working against us...

[...]

PR #6959, opened 2020-12-07
move unpackArchive to Device, use native version on android

pazos:
Requires https://github.com/koreader/android-luajit-launcher/pull/270
Requires https://github.com/koreader/koreader-base/pull/1255
Fixes https://github.com/koreader/koreader/issues/6613

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6959)

### DIFF
--- a/platform/android/llapp_main.lua
+++ b/platform/android/llapp_main.lua
@@ -68,7 +68,6 @@ end
 --- @todo Take care of this on extraction instead.
 -- Cf. <https://github.com/koreader/koreader/issues/5347#issuecomment-529476693>.
 android.execute("chmod", "755", "./sdcv")
-android.execute("chmod", "755", "./tar")
 
 -- set TESSDATA_PREFIX env var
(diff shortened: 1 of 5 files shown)


platform/android/llapp_main.lua
```diff
@@ -68,7 +68,6 @@ end
 --- @todo Take care of this on extraction instead.
 -- Cf. <https://github.com/koreader/koreader/issues/5347#issuecomment-529476693>.
 android.execute("chmod", "755", "./sdcv")
-android.execute("chmod", "755", "./tar")
```
Frenzie, line 71: Should we stop building tar on Android then?

Frenzie (approved):

platform/android/llapp_main.lua
pazos: Please

pazos:
Huh?. It seem to fail to uncompress some dicts while succeeded on others. Still better than we have now on modern android versions but I will take a look before merging :)

Frenzie (approved):

PR #6960, opened 2020-12-08
Add fallback text selection to panel zoom

Galunid:
Attempts to do what was discussed in https://github.com/koreader/koreader/pull/6511#issuecomment-739545598 - should allow zooming images in PDFs.

TODO:
Make the defaults per extension and simplify extension handling (avoid duplicated code)

@poire-z any comments?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6960)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -262,20 +262,6 @@ local highlight_style = {
     invert = _("Invert"),
 }
 
-local function getPanelZoomSupportedExt()
-    local default_supported_ext = {
-        cbz = true,
-        cbt = true,
-    }
-    return G_reader_settings:readSetting("panel_zoom_ext") or default_supported_ext
-end
-
-local function isPanelZoomSupported(file)
-    local filetype = util.getFileNameSuffix(file)
-    local supported_filetypes = getPanelZoomSupportedExt()
-    return supported_filetypes[filetype]
-end
-
 function ReaderHighlight:genPanelZoomMenu()
     return {
         {
@@ -288,9 +274,22 @@ function ReaderHighlight:genPanelZoomMenu()
             end,
             hold_callback = function()
                 local ext = util.getFileNameSuffix(self.ui.document.file)
-                local supported_ext = getPanelZoomSupportedExt()
-                supported_ext[ext] = not supported_ext[ext]
-                G_reader_settings:saveSetting("panel_zoom_ext", supported_ext)
+                local curr_val = G_reader_settings:getSettingForExt("panel_zoom_enabled", ext)
+                G_reader_settings:saveSettingForExt("panel_zoom_enabled", not curr_val, ext)
+            end,
+            separator = true,
+        },
+        {
+            text = _("Fall back to text selection"),
+            checked_func = function()
+                return self.panel_zoom_fallback_to_text_selection
+            end,
+            callback = function()
+                self:onToggleFallbackTextSelection()
+            end,
+            hold_callback = function()
+                local ext = util.getFileNameSuffix(self.ui.document.file)
+                G_reader_settings:saveSettingForExt("panel_zoom_fallback_to_text_selection", self.panel_zoom_fallback_to_text_selection, ext)
             end,
             separator = true,
         },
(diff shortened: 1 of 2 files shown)


frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -294,6 +294,19 @@ function ReaderHighlight:genPanelZoomMenu()
             end,
             separator = true,
         },
+        {
+            text = _("Fallback text selection"),
+            checked_func = function()
+                return self.fallback_text_selection
```
poire-z, line 300: Fallback **to** text selection - probably. @Frenzie?
(In the text, the variable, the function name.)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -294,6 +294,19 @@ function ReaderHighlight:genPanelZoomMenu()
...
+            end,
+            callback = function()
+                self:onToggleFallbackTextSelection()
+            end,
+            hold_callback = function()
+                G_reader_settings:saveSetting("fallback_text_selection", self.fallback_text_selection)
```
poire-z, line 306: But the setting should have the `panel_zoom_` prefix, so we know what it's about/its context when viewing settings.reader.lua.

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -698,13 +716,17 @@ function ReaderHighlight:onPanelZoom(arg, ges)
...
     end
-    return true
+    return false
 end
 
 function ReaderHighlight:onHold(arg, ges)
     if self.document.info.has_pages and self.panel_zoom_enabled then
-        return self:onPanelZoom(arg, ges)
+        local res = self:onPanelZoom(arg, ges)
+        if not self.fallback_text_selection or res then
```
poire-z, line 727: (I find `if res or not self.fallback_text_selection then` easier to read :)

poire-z (commented):
Yep, looks fine to me.

frontend/apps/reader/modules/readerhighlight.lua
Frenzie: That'd be "fall back to…"

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -682,6 +695,11 @@ function ReaderHighlight:onTogglePanelZoomSetting(arg, ges)
     self.panel_zoom_enabled = not self.panel_zoom_enabled
 end
 
+function ReaderHighlight:onToggleFallbackTextSelection(arg, ges)
+    if not self.document.info.has_pages then return end
+    self.fallback_text_selection = not self.fallback_text_selection
```
poire-z, line 700: Mentionned that too quickly previously :)
I think all your "fallback text selection" should be renamed "fallback to text selection": that means: the variable name, this function :) because I would read "fallback text selection" as "another kind of text selection".
And given that we're in ReaderHighlight where everything (but panel zoom) is about text selection :), I would even name all your "fallback text selection" to "panel zoom fallback to text selection" so there's zero risk of confusion.

```lua
function ReaderHighlight:onTogglePanelZoomFallbackToTextSelection(arg, ges)
    self.panel_zoom_fallback_to_text_selection= not self.panel_zoom_fallback_to_text_selection
```

Galunid:
@Frenzie Can you take a look at the latest commit and tell me if it looks sane enough?

frontend/extsettings.lua
```diff
@@ -0,0 +1,29 @@
+local util = require('util')
```
Frenzie, line 1: *cough*

Frenzie:
It's not immediately clear to me what it's for. If it's some kind of missing functionality for ReaderSettings maybe it should just be another method or two there?

Galunid:
Well, it was supposed to be module to deal with settings per file extension, so that `.cbz/.pdf/.cbt/.epub` etc. could have their own settings for stuff like Panel Zoom (and for now only that). The current approach works fine if it's just one setting, but it'd require to duplicate the functions like `getPanelZoomSupportedExt`, `isPanelZoomSupported` etc. for every setting. So now if I want to introduce `fallback text selection` it has to have its own functions. That's the reason I attempted to make it more generic and move it out to its own module. Of course I can move it to `LuaSettings` (I assume that's what you meant by `ReaderSettings`), if you feel it's better suited there.

Frenzie:
> Of course I can move it to LuaSettings (I assume that's what you meant by ReaderSettings), if you feel it's better suited there.

That's what I meant, yes. It seems unnecessarily complex to me to split a couple of extra convenience methods out to their own module.

Frenzie (approved):
lgtm

PR #6970, opened 2020-12-09
Add columns to ReaderZooming paged_modes

### LINKED ISSUE #6822: FR: Should "Zoom to fit column" automatically switch to "Page" view mode?
**Does your feature request involve difficulty completing a task? Please describe.**
I use "continuous" view mode for most documents. So when I read a two column article and switch to "Zoom to fit column" I also need to switch to "page" view mode, or else I will scroll from the left column of a page to the left column of the next page. (Is there a major use case where it makes sense to zoom to fit column AND continuous view mode?)

When switching zoom mode to "Zoom to fit column" automatically switch view mode to "page"

I think koreader is fantastic and have been wanting to help with development so If the you think this enhancement makes sense I might try a PR myself because it seems like it should be very simple and mostly a matter of getting a development environment set up.

johnbeard:
This means it will warn when changing to continuous mode while
zoom to columns is active.

----

This doesn't fix #6822 but it's related to a fix since probably all members of `paged_modes` probably should also warn/autofix in reverse (i.e. when changing the zoom mode)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6970)

### DIFF
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -730,12 +730,15 @@ function ReaderView:onSetFullScreen(full_screen)
 end
 
 function ReaderView:onSetScrollMode(page_scroll)
-    if self.ui.document.info.has_pages and page_scroll and self.ui.zooming.paged_modes[self.zoom_mode] then
+    if self.ui.document.info.has_pages and page_scroll
+            and self.ui.zooming.paged_modes[self.zoom_mode] then
         UIManager:show(InfoMessage:new{
             text = _([[
-Continuous view (scroll mode) works best with zoom to page width or zoom to content width.
+Continuous view (scroll mode) works best with zoom to page width, zoom to \
+content width or zoom to rows.
 
-In combination with zoom to fit page, page height, content height or content, continuous view can cause unexpected shifts when turning pages.]]),
+In combination with zoom to fit page, page height, content height, content or \
+columns, continuous view can cause unexpected shifts when turning pages.]]),
             timeout = 5,
         })
     end
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -734,11 +734,9 @@ function ReaderView:onSetScrollMode(page_scroll)
             and self.ui.zooming.paged_modes[self.zoom_mode] then
         UIManager:show(InfoMessage:new{
             text = _([[
-Continuous view (scroll mode) works best with zoom to page width, zoom to \
-content width or zoom to rows.
+Continuous view (scroll mode) works best with zoom to page width, zoom to content width or zoom to rows.
 
-In combination with zoom to fit page, page height, content height, content or \
-columns, continuous view can cause unexpected shifts when turning pages.]]),
+In combination with zoom to fit page, page height, content height, content or columns, continuous view can cause unexpected shifts when turning pages.]]),
             timeout = 5,
         })
     end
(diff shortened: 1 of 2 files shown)


Frenzie:
> warn/autofix

For the majority of them, autofix. I added a couple of warnings originally because there are certain combinations that don't really make sense (e.g., zoom to page + continuous) but some people still want to use them. But zoom to fit column + continuous is completely broken. I never intended warnings for such a scenario.

frontend/apps/reader/modules/readerview.lua
```diff
@@ -730,12 +730,15 @@ function ReaderView:onSetFullScreen(full_screen)
...
 
 function ReaderView:onSetScrollMode(page_scroll)
-    if self.ui.document.info.has_pages and page_scroll and self.ui.zooming.paged_modes[self.zoom_mode] then
+    if self.ui.document.info.has_pages and page_scroll
+            and self.ui.zooming.paged_modes[self.zoom_mode] then
         UIManager:show(InfoMessage:new{
             text = _([[
-Continuous view (scroll mode) works best with zoom to page width or zoom to content width.
+Continuous view (scroll mode) works best with zoom to page width, zoom to \
+content width or zoom to rows.
```
Frenzie, line 738:
```suggestion
Continuous view (scroll mode) works best with zoom to page width, zoom to content width or zoom to rows.
```

frontend/apps/reader/modules/readerview.lua
```diff
@@ -730,12 +730,15 @@ function ReaderView:onSetFullScreen(full_screen)
...
 
-In combination with zoom to fit page, page height, content height or content, continuous view can cause unexpected shifts when turning pages.]]),
+In combination with zoom to fit page, page height, content height, content or \
+columns, continuous view can cause unexpected shifts when turning pages.]]),
```
Frenzie, line 741:
```suggestion
In combination with zoom to fit page, page height, content height, content or columns, continuous view can cause unexpected shifts when turning pages.]]),
```

Frenzie (changes requested):

johnbeard:
@Frenzie so what to do in this case (changing from paged -> continuous)? still warn?

Frenzie:
That depends on how much time you're willing to put in. The PR is
acceptable to me for the moment as a quick interim solution, but if you
have the time to do it right of course I'd much rather see that instead.

jperon:
I'd plan to add the possibility, with columns and continuous mode, to add the beginning of next column to the end of current one (hope I'm clear enough…); but I won't have much time to work on it for a while, and don't really know all that's involved.

Frenzie (approved):
@johnbeard Please let me know what you plan to do. It's ready to merge, but I don't want to bother the translators with this if it won't be necessary.

johnbeard:
@Frenzie my plan was to use this commit to get the paged_modes array right and then hopefully build on it to autofix in future.

The text will stand anyway: even if there's a dialog in future ("erk, this is probably wrong, what to do: ignore and set anyway, autofix, cancel"), the warning text will be this string.

Frenzie:
Specifically this new string with the addition of rows/columns?

johnbeard:
> Specifically this new string with the addition of rows/columns?

This one:

> Continuous view (scroll mode) works best with zoom to page width, zoom to content width or zoom to rows.
> In combination with zoom to fit page, page height, content height, content or columns, continuous view can cause unexpected shifts when turning pages.

Will be the same no matter if there's an "autofix" option or not.

Frenzie:
I don't think I agree (i.e., why confuse users with nonsense about columns when you can't select continuous+columns at the same time in the first place), but since the strain on translators only consists of applying the diff and the string is accurate to the current situation I won't belabor the point.

PR #6973, opened 2020-12-10
ReaderFooter: Don't duplicate a 12h clock time format option

### LINKED ISSUE #6969: 12 hour time format leading zero
* KOReader version: 2020.11-45
* Device: Emulator, Samsung 10 (android)

#### Issue
When the time format is set to 12 Hour it displays a leading zero. e.g. 02:15 pm, I would expect to see 2:15pm. From 10 through 12 it does not have a leading zero. Screenshot attached
Clipboard01

#### Steps to reproduce
Set the time format to 12 Hour. BTW the setting in the status bar is not controlled by the device setting. Is there a reason for having two seperate settings?

N/A

NiLuJe:
* Use the global one instead (from Device > Time and date)
* Where possible (e.g., !Windows, because MS's CRTs are still terrible), trim the leading zero from the hour (fix #6969).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6973)


NiLuJe:
I just realized that we happen to *default* to 12-hour clock (I think?), which I'm not necessarily super-fond of, but, oh, well ;).

frontend/util.lua
```diff
@@ -224,21 +224,42 @@ end
...
+else
+    function util.secondsToHour(seconds, twelve_hour_clock)
+        local time
+        if twelve_hour_clock then
+            if os.date("%p", seconds) == "AM" then
+                -- @translators This is the time in the morning in the 12-hour clock (%-I is the hour, %M the minute).
+                time = os.date(_("%-I:%M AM"), seconds)
+            else
+                -- @translators This is the time in the morning in the 12-hour clock (%-I is the hour, %M the minute).
+                time = os.date(_("%-I:%M PM"), seconds)
```
poire-z, line 255: in the afteroon :)

NiLuJe: Oops, nice catch, c/p snafu ;).

Frenzie:
> I just realized that we happen to default to 12-hour clock, which I'm not necessarily super-fond of, but, oh, well ;).

Defaulting to 24 is fine by me.

frontend/util.lua
```diff
@@ -226,6 +226,8 @@ end
 ---- @treturn string hour string
 --- @note: The MS CRT doesn't support either %l & %k, or the - format modifier (as they're not technically C99 or POSIX).
 ---        They are otherwise supported on Linux, BSD & Bionic, so, just special-case Windows...
+---        We *could* arguably feed the os.date output to gsub("^0(%d)(.*)$", "%1%2"), but, while unlikely,
+---        it's conceavable that a translator wouldput something other that the hour at the front of the string ;).
```
Frenzie, line 230: typo

NiLuJe: *Inconceivable*! :D

NiLuJe: (I solemnly swear not to have made the typo just to land that reference :D).

Frenzie: Or unpossible? :-P

Frenzie:
@NiLuJe This seems to have broken the CI even though it didn't in this PR?

NiLuJe:
> @NiLuJe This seems to have broken the CI even though it didn't in this PR?

The current failure that I see on master is a random one I've seen before on CircleCI, but that i've never been able to reproduce locally :s.

NiLuJe:
At a *very* *very* quick glance, given the very nasty things this particular test does, it's probably a FS timestamp granularity/instability issue on the CI VMs.

(eg., typically the stuff that I'd tag "may fail" if busted allowed such a thing).

NiLuJe:
c.f., a few of the earlier tests ticking a giant usleep in between `touch`es ;p.

Ideally, it should be rewritten to set hard-coded timestamps via lfs to bypass the issue entirely.

poire-z:
Looks like the one failing is related to this PR - some width measurements that might change now that the hour is shorter - might not fail for commits/PRs checked between 12:00 to 23:59 :)  - or longer if the default is 24:
```
spec/front/unit/readerfooter_spec.lua:308: Expected objects to be the same.
Passed in:
(number) 362
Expected:
(number) 370

stack traceback:
	spec/front/unit/readerfooter_spec.lua:308: in function <spec/front/unit/readerfooter_spec.lua:296>

[  FAILED  ] spec/front/unit/readerfooter_spec.lua @ 296: Readerfooter module should pick up screen resize in resetLayout (85.55 ms)
```
https://github.com/koreader/koreader/blob/38d5510a7870f3de07cff93b190bb4d49f6b9805/spec/unit/readerfooter_spec.lua#L305-L308

NiLuJe:
Oh, right, the trim will do that, yeah, no idea why it passed in the PR o_O.

I'll take care of it ;).

Frenzie:
What if you expect something like a range or one of two values (362 or 370)?

NiLuJe:
I don't see anything like that in https://olivinelabs.com/busted/#asserts at a glance, and I don't feel like overthinking this (this specific test is already highly annoying to update) ;).

Frenzie:
I just mean `assert.is_true(number == 362 or number == 370)`, not as pretty but who cares. ;-)

PR #6975, opened 2020-12-10
Fix crash with RTL UI

poire-z:
Fix crash with RTL UI (little RTL UI specific layout bit not updated after #6885).

Also don't check show_func when sizing the names on the bottom menu, to avoid icons moving or resizing when toggling options. See https://github.com/koreader/koreader/pull/6885#issuecomment-735443154 and above:
> E minor aesthetically annyoing thing: when toggling Zoom to because of the different name_text show on the left, and their possibly different text size (Horizontal overlap is long :), the zoom for these names vs the room available for the widget may change: so, you have the widget images moving a bit, possibly changeing sizes to fit. I guess this will vary with language translation :)
May be we should take into account all the not enabled options' name_text into account to compute the left width?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6975)

### DIFF
--- a/frontend/ui/data/koptoptions.lua
+++ b/frontend/ui/data/koptoptions.lua
@@ -551,7 +551,7 @@ if BD.mirroredUILayout() then
     -- be mirrored - but that's not enough: we need to swap LEFT and RIGHT,
     -- so they appear in a more expected and balanced order to RTL users:
     -- {JUSTIFY, LEFT, CENTER, RIGHT, AUTO}
-    local j = KoptOptions[3].options[5]
+    local j = KoptOptions[4].options[5]
     assert(j.name == "justification")
     j.item_icons[2], j.item_icons[4] = j.item_icons[4], j.item_icons[2]
     j.values[2], j.values[4] = j.values[4], j.values[2]
--- a/frontend/ui/data/koptoptions.lua
+++ b/frontend/ui/data/koptoptions.lua
@@ -1,6 +1,7 @@
 local BD = require("ui/bidi")
 local Device = require("device")
 local optionsutil = require("ui/data/optionsutil")
+local util = require("util")
 local _ = require("gettext")
 local C_ = _.pgettext
 local Screen = Device.screen
@@ -556,6 +557,16 @@ if BD.mirroredUILayout() then
     j.item_icons[2], j.item_icons[4] = j.item_icons[4], j.item_icons[2]
     j.values[2], j.values[4] = j.values[4], j.values[2]
     j.labels[2], j.labels[4] = j.labels[4], j.labels[2]
+    -- The zoom direction items will be mirrored, but we want them to
+    -- stay as is, as the RTL diretions are at the end of the arrays.
+    -- By reverting the mirroring, RTL directions will be on the right,
+    -- so, at the start of the options for a RTL reader.
+    j = KoptOptions[3].options[7]
+    assert(j.name == "zoom_direction")
+    util.arrayReverse(j.item_icons)
+    util.arrayReverse(j.values)
+    util.arrayReverse(j.args)
+    j.default_value = 0
 end
 
 return KoptOptions
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -198,9 +198,14 @@ function ConfigOption:init()
         local show_default = not self.options[c].advanced or show_advanced
         local show = self.options[c].show
         -- Prefer show_func over show if there's one
+        -- Or may be not, as show_func is always used to show/hide some widget depending
+        -- on the value of another widget: it's best to keep it accounted for the names
+        -- max width, and avoid stuff moving when toggling options.
+        --[[
         if self.options[c].show_func then
             show = self.options[c].show_func(self.config.configurable, self.config.document)
         end
+        ]]--
         if show ~= false and show_default then
             local name_font_face = self.options[c].name_font_face and self.options[c].name_font_face or "cfont"
             local name_font_size = self.options[c].name_font_size and self.options[c].name_font_size or default_name_font_size


poire-z:
Mhhh, I guess the Directions icons order should also be re-arranged with the RTL UI, so the RTL ones are nearer the widget name.
Currently in the RTL UI:
image

poire-z:
I think this would make more sense to a RTL reader:
<kbd>image</kbd>
or in 2020.01:
image

poire-z:
Pinging @WaseemAlkurdi for confirmation the above is more natural.

Also, mentionning this here to not add it to the huge #6937 : 
@WaseemAlkurdi : with our new WIP icons, it feels to me some of them would need mirroring (`Aa`, the bottom one with lines of text, the hatched column one, may be the top one with a bookmark):

image

Feelings ? Are some of them more confusing that way than others?
Or would it work if they were just *all* mirrored ?

WaseemAlkurdi:
> I think this would make more sense to a RTL reader:
> image
> or in 2020.01:
> image

@poire-z I definitely agree here, the second variant is much more intuitive!



> Pinging @WaseemAlkurdi for confirmation the above is more natural.
> 
> Also, mentionning this here to not add it to the huge #6937 :
> @WaseemAlkurdi : with our new WIP icons, it feels to me some of them would need mirroring (`Aa`, the bottom one with lines of text, the hatched column one, may be the top one with a bookmark):
> 
> image
> 
> Feelings ? Are some of them more confusing that way than others?
> Or would it work if they were just _all_ mirrored ? (argh, it would obviously not work on the Zoom Direction ones... or I'd just have to not reverse the icons in my last commit, just the values)

@poire-z Very well, but the problem with mirroring icons is that most icons _don't_ need to be flipped, with the exception being the ones that show paragraphs ... so instead of mirroring all the icons when they're rendered, then setting exceptions, it might be better to just create RTL versions of the few icons that need to be mirrored, then leaving the rest as is, without mirroring.

Should it be necessary, I'm ready to go through every icon in the interface, then compile a list of which icons need to be mirrored.

poire-z:
I'm surprised :) You're fine with these, that have a LTR feeling ?:
image image image image

> just create RTL versions of the few icons that need to be mirrored, then leaving the rest as is, without mirroring.
Should it be necessary, I'm ready to go through every icon in the interface, then compile a list of which icons need to have RTL versions.

OK, so let's wait for 2020.01 and you have a chance to test it.
(No need to make mirrored versions of the icons: it shouldn't be too hard to just flag those that need it and have the code flip/mirror them.)

WaseemAlkurdi:
> I'm surprised :) You're fine with these, that have a LTR feeling ?:
> image

This one is an Latin letter A, so it (IMO) shouldn't be RTL'd - except if we could replace it with Arabic script and Hebrew script alternatives?
Shouldn't be necessary anyhow, the meaning is conveyed without having to do this, right?

> image 

For the first one, it depends on whether the zoom direction implied is "the beginning of the line" (which means that mirroring it wouldn't be correct) or "the left side of the screen" (which means that we should mirror it). I understood this to mean the former, hence the suggestion to keep it.

For the second and third icons (the ones with the arrows), reversing them won't matter a lot, so I thought to just keep them (because I thought it would be a lot of work to mirror them)

> image 

This one doesn't matter a lot either, since the direction in which one rotates the screen is more dependent on one's handedness (right-handed vs. left-handed) than it is on the direction of what's on the screen.

> image

Isn't this one already RTL? I don't know, but when placing a (physical) bookmark, one would place it closer to the spine (to stop it from falling out) ... which would imply that the spine is on the right (i.e. RTL book)?

> OK, so let's wait for 2020.01 and you have a chance to test it.
> (No need to make mirrored versions of the icons: it shouldn't be too hard to just flag those that need it and have the code flip/mirror them.)

Fingers crossed!
And BTW, I'm really in love with the new icons! :-D

Frenzie:
It seems to me that the font size icon could just as easily be the other way around. In fact wouldn't that make more sense from an LTR perspective? Left to right, smaller to larger.

NB I'm not suggesting we swap it.

> Isn't [the bookmark] already RTL? I don't know, but when placing a (physical) bookmark, one would place it closer to the spine (to stop it from falling out) ... which would imply that the spine is on the right (i.e. RTL book)?

I think it's upside down. :-P

tl;dr Maybe don't overthink the icons too much as long as they clear. ^_^

poire-z:
> image
> Shouldn't be necessary anyhow, the meaning is conveyed without having to do this, right?

Ok, right :) 

> image 
> For the first one, it depends on whether the zoom direction implied is "the beginning of the line" (which means that mirroring it wouldn't be correct) or "the left side of the screen" (which means that we should mirror it). I understood this to mean the former, hence the suggestion to keep it.

For me it means: the left side (= the beginning of the page for us LTR readers) is shown (hatched)
But I realize the hatched side might also mean: not shown - and the white part is what is shown.
So, overall, it's free to interpretation, and there's may be no unique interpretation, so, well, if you're fine with it, let it be non- mirrorable.

> image
> Isn't this one already RTL? I don't know, but when placing a (physical) bookmark, one would place it closer to the spine (to stop it from falling out) ... which would imply that the spine is on the right (i.e. RTL book)?

Well, that's true, in real life, I would put it nearer the spine. But if we were to put many post-it to make pages, it's safer to put them away from the spine, as they stick and the thickness they add don't make the spine crack :) 
But that's somehow how I would expect this icon to show as LTR :) So, ok, as-is.

> either a rewrite of the icon rendering code to give it mirroring support, or making separate icons for RTL vs. LTR (like in Android apps and localized resources ...)

Well, we don't yet have a flip/mirror method for blitbuffers, but it looks like it would not be complicated to add (@NiLuJe , thoughts ?)

NiLuJe:
An upside down flip is essentially free (e.g., just blit bottom to top). A vertical flip might be much less friendly, depending on context.

(It's also fairly boring work, so, I'm way more in favor of having flipped icons if this affects like 5 of them ;p).

Frenzie:
Guys, guys, guys… remember what I wrote before. I'm still having that black block issue but something like this:

```diff
diff --git a/frontend/ui/renderimage.lua b/frontend/ui/renderimage.lua
index 04084b0f..8c33197f 100644
--- a/frontend/ui/renderimage.lua
+++ b/frontend/ui/renderimage.lua
@@ -211,7 +211,17 @@ function RenderImage:renderSVGImageFileWithNanoSVG(filename, width, height, zoom
     if not NnSVG then
         NnSVG = require("libs/libkoreader-nnsvg")
     end
-    local svg_image = NnSVG.new(filename)
+
+
+    local file = io.open(filename, "r")
+    if not file then return nil end
+    local content = file:read("*a")
+    file:close()
+
+    content = content:gsub("(<.* )", '%1 transform="scale(1, -1) translate(0, -100)" ')
+
+    local svg_image = NnSVG.new(content, true)
+
     local native_w, native_h = svg_image:getSize()
     if not zoom then
         if width and height then
```

NiLuJe:
Oh, right, SVG :D.

Frenzie:
Easier than flipping blitbuffers anyway, but the part about selectively doing it manually being smarter probably applies regardless. ;-)

PR #6976, opened 2020-12-11
The final bit of my great bundle of messy stuff

NiLuJe:
Companion to https://github.com/koreader/koreader-base/pull/1257

I intend to somewhat cleanup the history before merge.

----

* BookInfoManager:
  * Use ZSTD instead of zlib to compress thumbnails, because zlib is the absolute worst in terms of performance nowadays.
     Like in CRe, slightly smaller DB, slightly faster compression, hilariously faster decompression.
  * Also revamps the schema a tiny bit following recent discussions:
    * Added `filesize`, `filemtime` columns, and split `series` into `series` and `series_index`.
  * Made the DB migration slightly less harsh (i.e., preserve settings, and a bit of visual feedbad).

* A few `__gc` metamethod tweaks as mentioned in base.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6976)


NiLuJe:
It's -- admittedly, weird -- but valid syntax, mainly to make the captures more obvious.

plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -423,7 +477,27 @@ function BookInfoManager:extractBookInfo(filepath, cover_specs)
...
+                --         * DjVu, for which I couldn't find a real standard for metadata fields
+                --           (we currently use Series for this field, c.f., https://exiftool.org/TagNames/DjVu.html).
+                --         * CRe, which could offer us a split getSeriesName & getSeriesNumber...
+                --           except getSeriesNumber does an atoi, so it'd murder decimal values.
+                --           So, instead, parse how it formats the whole thing as a string ;).
+                if string.find(props.series, "#") then
+                    dbrow.series, dbrow.series_index = props.series:match("(.*) #(.*)")
+                    if dbrow.series_index then
+                        -- We're inserting via a bind method, so make sure we feed it a Lua number, because it's a REAL in the db.
+                        dbrow.series_index = tonumber(dbrow.series_index)
```
poire-z, line 492: Is it worth it being stored as a REAL (storing `4.0` and `4`, we would be fetching the same value for both ?)  when in the end, you're just using the provided value as-is, as a string?

NiLuJe: The reasoning was the potential pie-in-the-sky query UI, where you would most definitely want it kept as a real to do stuff like "get me all books in series n and order them by index" ;).

(Also, I like being anal about data types in general ^^).

NiLuJe: Given than native Lua numbers are already all doubles, it's not really much of an issue?

NiLuJe: Double-checked, and fun (and welcome) fact: a round integer is displayed without a decimal by Lua.

e.g., Calibre always uses a real for series_index, so given

```
    <meta property="belongs-to-collection" id="id-3">Tinfoil Dossier</meta>
    <meta refines="#id-3" property="collection-type">series</meta>
    <meta refines="#id-3" property="group-position">3.0</meta>
```

CRe honors that, and we get `["series"] = "Tinfoil Dossier #3.0"` in `doc_props`.

BookInfo honors that, and we get `Tinfoil Dossier|3.0` in the DB.

And, then, Lua happens... and, wonders of wonders:

```
> print(tonumber(3.0))
3
```

\o/

NiLuJe: Which means, now I understand what the `gsub` I zapped were for, and I need to restore them for the pure doc_props case ;).

poire-z: > it's not really much of an issue?

It's not because SQLite is friendly (data type not enforced on the column, but stored with the data), and won't fail if you store `"zzz"` in a REAL column.
I think we had that discussion in the past, possibly already about the series stuff (or some other calibre metadata): you trust the data type (so, the user or publisher to have done the right thing) - while I think we can meet garbage in the wild, and that it's safer to keep the garbage by just storing everything as text - and do only the conversion at the end when it matters.

(But never really seeing any series with the kind of book I read, I don't mind much how you deal with that :)

poire-z: Or may be, split by # and only if the right part is a number, store it in series_index.
If it's not a number, let it be as part of series, in case you get "The Avengers vs. #twitter" as a serie :)

NiLuJe: Oh, `tonumber("foo")` returns nil, which I'm guessing will be bound as a NULL, so I assume this all shakes out all right in practice :).

NiLuJe: The only real issue would be roman numerals, but I don't *think* I've ever seen those in the wild ^^.

NiLuJe: > Oh, `tonumber("foo")` returns `nil`, which I'm guessing will be bound as a NULL, so I assume this all shakes out all right in practice :).

Oh, except for not restoring that potential bit of string to `series`, right ;).

NiLuJe: b7339144 ought to take care of that, thanks!

poire-z: If you're confident with that, ok :)
(But I would personally store the raw stuff in the db, and do the series parsing at display/search time - so if you get some case when the logic needs a tweak, you don't have to trash the whole DB to reindex/store something right.)

NiLuJe (commented):
> Reviewed 8 of 11 files at r1, 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/6976)* status: all files reviewed, 1 unresolved discussion (waiting on @poire-z)

NiLuJe:
History cleaned up, and archived in https://github.com/NiLuJe/koreader/tree/cbb-c99-front_archive

poire-z:
(Bravo on the no-blank line target ! :)
There's some uneeded 2 leading spaces before your `*` (git log will already indent the messages by 4 spaces, so that will be a lot of left space wasted).

NiLuJe:
That's to ease separation between multiple h1 level headers ;).

That said, there's never two of those here, so, here we go :).

PR #6977, opened 2020-12-11
Update UI layout code for use of SVG icons

poire-z:
Don't trust the icons' native sizes: replace `scale_for_dpi=true` with `width = height= DGENERIC_ICON_SIZE`, so all icons get the same (tunable) size - except in a few specific use cases.

I've put `DGENERIC_ICON_SIZE = 50` (updated to 48) to provoke a reaction :) as we need to decide a value for this.

More context around https://github.com/koreader/koreader/pull/6937#issuecomment-741964278

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6977)


Frenzie:
50 or 48 (144/3)? I'll take a look at what the size looks like later/tonight.

On Fri Dec 11 2020 10:01:16 GMT+0100 (Central European Standard Time), poire-z wrote:



Don't trust the icons' native sizes: replace scale_for_dpi=true with width = height= DGENERIC_ICON_SIZE, so all icons get the same (tunable) size - except in a few specific use cases.

I've put DGENERIC_ICON_SIZE = 50 to provoke a reaction :) as we need to decide a value for this.

More context around #6937 (comment)

This change is 

You can view, comment on, or merge this pull request online at:

  https://github.com/koreader/koreader/pull/6977

Commit Summary
Update UI layout code for use of SVG icons
File Changes
M .luacheckrc (1)
M defaults.lua (3)
M frontend/apps/filemanager/filemanager.lua (4)
M frontend/apps/reader/modules/readertoc.lua (17)
M frontend/apps/reader/modules/readerview.lua (4)
M frontend/ui/rendertext.lua (6)
M frontend/ui/widget/button.lua (6)
M frontend/ui/widget/configdialog.lua (6)
M frontend/ui/widget/confirmbox.lua (4)
M frontend/ui/widget/container/alphacontainer.lua (8)
M frontend/ui/widget/container/movablecontainer.lua (6)
M frontend/ui/widget/iconbutton.lua (7)
M frontend/ui/widget/imagewidget.lua (6)
M frontend/ui/widget/infomessage.lua (9)
M frontend/ui/widget/multiconfirmbox.lua (4)
M frontend/ui/widget/touchmenu.lua (3)
M spec/unit/defaults_spec.lua (16)
Patch Links:
https://github.com/koreader/koreader/pull/6977.patch
https://github.com/koreader/koreader/pull/6977.diff

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub, or unsubscribe.


-- 
Sent with the new, awesome Vivaldi mail client / browser. Download it at vivaldi.com.

johnbeard:
I recommend 48 as that's 1) a standard icon size and 2) easier to scale up and down to other common sizes.

poire-z:
>  2) easier to scale up and down to other common sizes.

Does that matter at this point ? It's the target size, so no scale up/down (unless scaleBySize would magically round things nicely with 48 as input?).
You could keep your source at 144px - or do you plan to resize them at 48 ?

johnbeard:
>  You could keep your source at 144px - or do you plan to resize them at 48 ?

Well, if we're going to actually render at 48px, it makes more sense to edit the SVGs at that size because then they'll be sharper (4px at 144px rendered at 48px will always be anti-aliased and fuzzy, and any lines not on a multiple of 3 pixel grid will be fuzzy too).

Whether that's an issue depends on the device I think. On my Clara HD, it's pretty sharp anyway (might just be too small for my eyes). On the emulator (and Android maybe??) it's pretty fuzzy.

Plus, if we go to 48px and round up from 4/3 to 2px, then we get thicker lines.

NiLuJe:
The issue here is that the 48 @poire-z  is talking about is not the 48px @johnbeard is talking about ^^.

That 48 is fed through scaleBySize, which will *then* spit out an actual pixel value, which will most likely never ever be 48px ;).

johnbeard:
Oh right, I though it was going to change to be _actually_ 48px (or some other value). Ignore me then, it doesn't matter what the SVG size is :-D

poire-z:
Only on the emulator we'll render at 48px (may be only on my small one?). On my GloHD, I know my scale ratio: x 1.78667, so I'll have them at 85px.
(But I don't understand much about DPI and the scaling stuff, so I let you and @Frenzie take the right decisions :)

poire-z:
(But but but... can I still get thicker lines ? :)

Frenzie:
I don't think it's a relevant concern outside of computers with low-res monitors and very old devices, and it's not very bad on those either.

> (But but but... can I still get thicker lines ? :)

Something like 6px at 144 or 2 at 48 is fine by me. However, I wouldn't worry too much about what it looks like at 48 *unless* that happens to be exactly the value we like. Worry more about how it looks at 96 or 144. ;-)

johnbeard:
Are we talking thicker for _all_ icons or just the `appbar` icons which render smaller (currently?)

Because if it's the latter, we should consider trying to figure out what the actual size ratio is so we can thicken those icons relative to the "normal" icons (e.g. zoom to page) so they come out at the same actual thickness on screen.

Frenzie:
Pretty sure we're talking about all icons.

poire-z:
> which render smaller (currently?)

Smaller than what? _Currently_ which?
Because this PR is about making them all the same size: DGENERIC_ICON_SIZE (=48 currently).
(The zoom and direction one also, but these are allowed to have their size decreased to fit, only when needed.)

So, we all should test from now on with this PR applied on our emulator and devices.

NiLuJe:
Assuming you've not set a custom UI DPI, the `scaleBySize` formula is:

`ceil(input * ((min(viewWidth, viewHeight/600) + (dpi/160)) / 2)`

NiLuJe:
Now that I've stopped mangling the maths, the explanation: it was designed to be 1:1 on the original 6" Vizplex screens: 600x800 @ 127dpi.

poire-z:
> Or, if you've set a custom UI dpi and it doesn't match the device's:

Is there a logic about this distinction ?
So, if I use the default of 300, or if I set a custom DPI of 301, I would get really different results ?

GloHD or ClaraHD (1448 x 1072 @ 300dpi):
```
> print( 48 * (1072/600 + 1072/600) / 2 ) -- with device dpi ("auto")
85.76
> print( 48 * (1072/600 + 301/160) / 2 ) - custom dpi of 301
88.03

> print( 48 * (1072/600 + 300/160) / 2 ) -- what we'd get with using device dpi of 300
87.88
```

NiLuJe:
> > Or, if you've set a custom UI dpi and it doesn't match the device's:
> 
> Is there a logic about this distinction ?

I'm as puzzled as you are by that bit of logic every time I look at it ^^.

It possibly coulda maybe have made sense at the time, but it's certainly fishy nowadays.

NiLuJe:
Rabbit hole:

* The switch to 167 from 160 was explained by @Frenzie in https://github.com/koreader/koreader-base/pull/843
* See also https://github.com/koreader/koreader-base/pull/822
* And the original logic, from https://github.com/koreader/koreader-base/pull/344 was already roughly functioning the same way.

poire-z:
So, I guess it's time to make it right, if we are going to resize things around a bit :)

NiLuJe:
Let's see what @Frenzie has to say about it, but I'd tend to agree, yeah ;).

poire-z:
There's a little top vs bottom menu bar height difference, because of this span:
https://github.com/koreader/koreader/blob/86701f6cccd5c0f4e26cb5719559a3bb526802f5/frontend/ui/widget/touchmenu.lua#L236

image

With:
image

Without:
image

So, we have to decide whether to kill that span at top, or add it at bottom.

Also, this PR does not include the line thichness reduction from https://github.com/koreader/koreader/pull/6937#issuecomment-739385544 .
That's another option to decide.

Frenzie:
> So, if I use the default of 300, or if I set a custom DPI of 301, I would get really different results ?

While quite unlikely for a minor step like the example given, it's certainly possible that all kinds of things could suddenly gain an extra pixel. Your 2px line turns into a 3px line at some point.

Setting aside the scaleBy*Size* part of the equation:

* 1px at 160
* turns into 2px at 240 DPI
  1*240/160 = 1.5 → 2px
* 2px at 320
* turns into 3px at… (and so forth)

> So, I guess it's time to make it right, if we are going to resize things around a bit :)

Make what right, precisely?

Frenzie:
Note that we purposefully play with the above. 0.5px at 160 = 1px, all the way up to far above 320 dpi. It then makes 160 DPI a bit fatter than the look we're actually going for.

Frenzie:
> Make what right, precisely?

Oh, I see what you mean. No, I think it's probably an oversight in custom DPI handling unless comments/git blame have anything pertinent to say.

Frenzie:
> Without:

Looks a bit unbalanced to me.

NiLuJe:
> 
> So, we have to decide whether to kill that span at top, or add it at bottom.
> 

I'd be moderately inclined to have it in both menus, but no super strong opinion about it (e.g., keeping the bottom menu as small as possible has its merits).

poire-z:
> There's a little top vs bottom menu bar height difference

OK, fixed. We can have both menu bars have balanced similar padding, here with `Size.padding.default` (that we can reduce if too much):
image

poire-z:
I think, now with the sample blacker icons at 6px strokes, that I prefer our current smaller size of `DGENERIC_ICON_SIZE = 40` to the 48 that I put to get larger and thicker icons and counteract their grayness.
image
(we just get them in the Zoom panel at 40, vs 50 before)

But as it's now configurable, one can have them to his taste. We just have to decide for a nice default.

Frenzie:
Yup, let's just stick to the same old size. ;-)

NiLuJe:
Currently building a Kobo build right now to check behavior on real nightmode, but I'm confident the patch to handle highlights properly will look very much like that post https://github.com/koreader/koreader-base/pull/1257 & https://github.com/koreader/koreader/pull/6976 ;).

```diff --git a/frontend/ui/widget/imagewidget.lua b/frontend/ui/widget/imagewidget.lua
index b8fb4fbf..2c5053b2 100644
--- a/frontend/ui/widget/imagewidget.lua
+++ b/frontend/ui/widget/imagewidget.lua
@@ -178,6 +178,46 @@ function ImageWidget:_loadfile()
                     self._bb = RenderImage:scaleBlitBuffer(self._bb, math.floor(bb_w * DPI_SCALE), math.floor(bb_h * DPI_SCALE))
                 end
             end
+
+            -- Now, if that was *also* one of our icons, and it has an alpha channel,
+            -- compose it against a background-colored BB now, and cache *that*.
+            -- This helps us avoid repeating alpha-blending steps down the line,
+            -- and also ensures icon highlights/unhilights behave sensibly.
+            if self.is_icon then
+                local bbtype = self._bb:getType()
+                if bbtype == Blitbuffer.TYPE_BB8A or bbtype == Blitbuffer.TYPE_BBRGB32 then
+                    print("ImageWidget:_loadfile:", self.file, "is a transparent icon")
+                    local icon_bb = Blitbuffer.new(self._bb.w, self._bb.h, Screen.bb:getType())
+                    --- @note: Should match the background color. Which is currently hard-coded as white ;).
+                    ---        See the note below in paintTo for how to make the dim flag behave in case
+                    ---        this no longer actually is white ;).
+                    icon_bb:fill(Blitbuffer.COLOR_WHITE)
+
+                    -- And now simply compose the icon on top of that, with dithering if necessary
+                    -- Remembering that NanoSVG feeds us straight alpha, unlike MµPDF
+                    if self._is_straight_alpha then
+                        if Screen.sw_dithering then
+                            icon_bb:ditheralphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        else
+                            icon_bb:alphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        end
+                    else
+                        if Screen.sw_dithering then
+                            icon_bb:ditherpmulalphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        else
+                            icon_bb:pmulalphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        end
+                    end
+
+                    -- Free the original icon w/ an alpha-channel, keep the flattened one
+                    self._bb:free()
+                    self._bb = icon_bb
+
+                    -- There's no longer an alpha channel ;)
+                    self._is_straight_alpha = nil
+                end
+            end
+
             if not self.file_do_cache then
                 self._bb_disposable = true -- we made it, we can modify and free it
             else
@@ -385,19 +425,23 @@ function ImageWidget:paintTo(bb, x, y)
     end
     if do_alpha then
         -- NOTE: MuPDF feeds us premultiplied alpha (and we don't care w/ GifLib, as alpha is all or nothing),
-        -- but NanoSVG feeds us straight alpha
+        --       while NanoSVG feeds us straight alpha
         if self._is_straight_alpha then
-            --- @todo if Screen.sw_dithering then use bb:ditheralphablitFrom() when it's available
-            bb:alphablitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
+            -- NOTE: Our icons are already preprocessed properly, eaither at encoding time, or at caching time.
+            if Screen.sw_dithering and not self.is_icon then
+                bb:ditheralphablitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
+            else
+                bb:alphablitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
+            end
         else
-            if Screen.sw_dithering then
+            if Screen.sw_dithering and not self.is_icon then
                 bb:ditherpmulalphablitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
             else
                 bb:pmulalphablitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
             end
         end
     else
-        if Screen.sw_dithering then
+        if Screen.sw_dithering and not self.is_icon then
             bb:ditherblitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
         else
             bb:blitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
@@ -406,13 +450,22 @@ function ImageWidget:paintTo(bb, x, y)
     if self.invert then
         bb:invertRect(x, y, size.w, size.h)
     end
+    --- @note: This is mainly geared at black icons/text on a *white* background,
+    ---        otherwise the background color itself will shift.
+    ---        i.e., this actually *lightens* the rectangle, but since it's aimed at black,
+    ---        it makes it gray, dimmer; hence the name.
+    ---        TL;DR: If we one day want that to work for icons on a non-white background,
+    ---        a better solution would probably be to take the icon pixmap as an alpha-mask,
+    ---        (which simply involves blending it onto a white background, then inverting the result),
+    ---        and colorBlit it a dim gray onto the target bb.
+    ---        This would require the *original* transparent icon, not the flattened one in the cache.
+    ---        c.f., https://github.com/koreader/koreader/pull/6937#issuecomment-748372429 for a PoC
     if self.dim then
         bb:dimRect(x, y, size.w, size.h)
     end
-    -- If in night mode, invert all rendered images, so the original is
+    -- In night mode, invert all rendered images, so the original is
     -- displayed when the whole screen is inverted by night mode.
-    -- Except for our black & white icon files, that we do want inverted
-    -- in night mode.
+    -- Except for our black & white icons, they're already fine.
     if Screen.night_mode and not self.is_icon then
         bb:invertRect(x, y, size.w, size.h)
     end
```

NiLuJe:
Also, finally tried the thinner lines from https://github.com/koreader/koreader/pull/6937#issuecomment-739385544

I'm not necessarily a fan. While they might feel a bit thick on the emu, the current style is just fine on devices, I think?

Also, random issues with the PoC: In configmenu, it makes the icons shifts ever-so slightly by a few pixels horizontally when you switch tabs.

NiLuJe:
Oh, and, yeah, everything's fine on Kobo with the above patch ;p.

poire-z:
> finally tried the thinner lines from #6937 (comment)
I'm not necessarily a fan

Neither am I. This was just to get friendly with the initially thin icons. But now that they grew fat, no more need for thinner lines.

poire-z:
(^ just a rebase to current master, no change)

poire-z:
Ready to merge, and have our emulators cleaned from PNGs.

Note that I replaced our previous koreader.png (600x600) by @johnbeard koreader.png (244x244), figuring it might require less memory from users (sdl window icon).
I don't know if there are other use case where a 600x600 might be preferable if a large size is used, for better quality.

Frenzie:
244x244 is too small and will look blurry and ugly

poire-z:
> 244x244 is too small and will look blurrier and uglier when larger icons are shown.

That's about the koreader.PNG image, that is now only  used as the SDL KOReader window manager icon.
And as the default/fallback image for screensaver, so, ok, let's keep 600x600.

Frenzie:
That's what I'm talking about. In a large icon window switcher, you want 512 × 512 or better. Also koreader.png is the application icon, where the same thing applies.

Frenzie:
https://github.com/koreader/koreader/blob/f0b09fb4b3ecf98bdd654ebccac4dcc304bed255/Makefile#L408

poire-z:
OK. Now other comment, can I merge it?

poire-z:
Oups, some small cleanup in Makefile needed (I only rgrep'ed for stuff in subdirectories...).

Frenzie:
> Now other comment, can I merge it?

Actually I noticed you're also removing the source files for the old icons. Please don't! ^_^

Frenzie:
I'd like to `git mv` them into an appropriate `classic` folder (and replace the seemingly random YAMLs by the relevant SVGs).

poire-z:
> Actually I noticed you're also removing the source files for the old icons. Please don't! ^_^

You really want me to let them?
(I've been just removing the exclusion from zip and rm of icons/src :)

Frenzie:
> You really want me to let them?

Yup! When I said PNG I meant PNG. :-)

poire-z:
OK, done.

Frenzie:
For example, here's what it looks like in GNOME Shell (I think the icon's drawn at ~ 300 × 300):
Screenshot_2020-12-19_13-43-59

Frenzie (approved):
I didn't read/test the alpha stuff in detail but seems fine while skimming.

poire-z:
Not much alpha stuff changed, but @NiLuJe will soon re tweak all that :)

@NiLuJe : one little useless thing to check anyway for your blit/alpha stuff changes: the alphacontainer, that is only used with PDF by readerview for its arrow - when `Page overlap > Arrow` is used instead of `Gray out`:

<kbd>image</kbd>

It's quite crappy, I remember I spent quite some time last week as the SVG stuff made it no more transparent. But well, rarely used, not important.

pazos:
@poire-z: I tried to submit a PR against your branch but I cannot locate your fork as a base.

Anyway, it was to move platform specific resources outside shared icons:

```diff
diff --git a/Makefile b/Makefile
index 453757ef..17722647 100644
--- a/Makefile
+++ b/Makefile
@@ -383,7 +383,6 @@ androidupdate: all
 		-xr!*NOTES.txt$ \
 		-xr!*NOTICE$ \
 		-xr!*README.md$ \
-		-xr!*koreader.icns$ \
 		-xr'!.*'
 
 	# make the android APK
@@ -427,7 +426,7 @@ macosupdate: all
 		$(INSTALL_DIR)/bundle/Contents/MacOS \
 		$(INSTALL_DIR)/bundle/Contents/Resources
 
-	cp resources/koreader.icns $(INSTALL_DIR)/bundle/Contents/Resources/icon.icns
+	cp $(MACOS_DIR)koreader.icns $(INSTALL_DIR)/bundle/Contents/Resources/icon.icns
 	cp -LR $(INSTALL_DIR)/koreader $(INSTALL_DIR)/bundle/Contents
 	cp -pRv $(MACOS_DIR)/menu.xml $(INSTALL_DIR)/bundle/Contents/MainMenu.xib
 	ibtool --compile "$(INSTALL_DIR)/bundle/Contents/Resources/Base.lproj/MainMenu.nib" "$(INSTALL_DIR)/bundle/Contents/MainMenu.xib"
diff --git a/platform/mac/do_mac_bundle.sh b/platform/mac/do_mac_bundle.sh
index 02fcb400..933ab349 100755
--- a/platform/mac/do_mac_bundle.sh
+++ b/platform/mac/do_mac_bundle.sh
@@ -160,8 +160,7 @@ rm -rf cache clipboard history ota \
     l10n/.git l10n/.tx l10n/templates l10n/LICENSE l10n/Makefile l10n/README.md \
     plugins/SSH.koplugin plugins/hello.koplugin plugins/timesync.koplugin \
     plugins/autofrontlight.koplugin resources/fonts resources/icons/src \
-    resources/kobo-touch-probe.png resources/koreader.icns rocks/bin \
-    rocks/lib/luarocks screenshots spec tools
+    rocks/bin rocks/lib/luarocks screenshots spec tools
 
 # adjust reader.lua a bit.
 
diff --git a/resources/kobo-touch-probe.png b/resources/kobo-touch-probe.png
deleted file mode 100644
index d34260f7..00000000
Binary files a/resources/kobo-touch-probe.png and /dev/null differ
diff --git a/resources/koreader.icns b/resources/koreader.icns
deleted file mode 100644
index 97dd63f0..00000000
Binary files a/resources/koreader.icns and /dev/null differ
diff --git a/tools/kobo_touch_probe.lua b/tools/kobo_touch_probe.lua
index 7e2cf675..673ec778 100755
--- a/tools/kobo_touch_probe.lua
+++ b/tools/kobo_touch_probe.lua
@@ -51,7 +51,7 @@ function TouchProbe:init()
         },
     }
     self.image_widget = ImageWidget:new{
-        file = "resources/kobo-touch-probe.png",
+        file = "tools/kobo-touch-probe.png",
     }
     local screen_w, screen_h = Screen:getWidth(), Screen:getHeight()
     local img_w, img_h = self.image_widget:getSize().w, self.image_widget:getSize().h
```

poire-z:
OK, I'm applying it, thanks.

poire-z:
(Just changed your `cp $(MACOS_DIR)koreader.icns` to `cp -pv $(MACOS_DIR)/koreader.icns` - and renamed the files to their new locations.)

pazos (approved):

poire-z:
@NiLuJe : I haven't thought/I forgot about the inverted colors in nightmode:
I hope your yesterday patch takes care of that :)

Frenzie:
@poire-z Did you delete one too many?

```
12/19/20-16:02:21 DEBUG cache image|resources/icons/dogear.png|23|23
./luajit: frontend/ui/widget/imagewidget.lua:191: attempt to index field 'bb' (a nil value)
stack traceback:
	frontend/ui/widget/imagewidget.lua:191: in function '_loadfile'
	frontend/ui/widget/imagewidget.lua:208: in function '_render'
	frontend/ui/widget/imagewidget.lua:313: in function 'getSize'
	plugins/coverbrowsersearch.koplugin/listmenu.lua:807: in function 'paintTo'
	frontend/ui/widget/verticalgroup.lua:48: in function 'paintTo'
	frontend/ui/widget/verticalgroup.lua:52: in function 'paintTo'
	frontend/ui/widget/overlapgroup.lua:82: in function 'paintTo'
	frontend/ui/widget/container/framecontainer.lua:96: in function 'paintTo'
	frontend/ui/widget/container/inputcontainer.lua:85: in function 'paintTo'
	frontend/ui/widget/verticalgroup.lua:48: in function 'paintTo'
	frontend/ui/widget/container/framecontainer.lua:96: in function 'paintTo'
	frontend/ui/widget/container/inputcontainer.lua:85: in function 'paintTo'
	frontend/ui/uimanager.lua:1089: in function '_repaint'
	frontend/ui/uimanager.lua:1235: in function 'handleInput'
	frontend/ui/uimanager.lua:1312: in function 'run'
	./reader.lua:306: in main chunk
	[C]: at 0x55fd5e4de791
```

NiLuJe:
@poire-z: Nope, colors have to be inverted or the B&W stuff doesn't work ;).

poire-z:
> Did you delete one too many?

You probably did `git pull` and forgot to `./kodev something like build` to have `plugins/*` copied into `koreader-emulator/koreader/plugins/` (which happens to me quite often - and is a bit bothering when modifying plugins).

Frenzie:
> You probably did git pull and forgot to ./kodev something like build to have plugins/* copied into koreader-emulator/koreader/plugins/ (which happens to me quite often - and is a bit bothering when modifying plugins).

I always use `./kodev run` (which does build) actually. It takes only a split second. :-) It's only for Android that I might avoid it. But it's a false alarm because it's just local due to the giant icon name/calling changes.

frontend/ui/widget/iconwidget.lua
```diff
@@ -0,0 +1,70 @@
...
+Subclass of ImageWidget to show icons
+]]
+
+local DataStorage = require("datastorage")
+local ImageWidget = require("ui/widget/imagewidget")
+local Screen = require("device").screen
+
+-- Directories to look for icons by name, with any of the accepted suffixes
+local ICONS_DIRS = {}
+local user_icons_dir = DataStorage:getSettingsDir() .. "/icons"
```
poire-z, line 11: I guess this should be `DataStorage:getDataDir()` instead of `DataStorage:getSettingsDir()`.
(`koreader/icons/` instead of `koreader/settings/icons/`)

Frenzie: Yes, it doesn't seem to have much to do with settings.

PR #6980, opened 2020-12-12
kodev: Use getopt instead of a hand-rolled hack

NiLuJe:
* This means that stuff like passing a custom gdb frontend or custom valgrind args is now actually usable ;).
* Fixed the stock GDB invocation (passing args to the debuggee is... not straightforward).
* Only pass LD_LIBRARY_PATH to LuaJIT, not the script's env, because some GDB frontends (hi, nemiver) may have dependencies on libraries we ship, causing ABI breakages (e.g., happened for glib).
* Made the valgrind defaults slightly more aggressive
* Added a --callgrind shortcut for... callgrind
* Documented how to use a valgrind suppression file to tame mesa/drm on distros that don't build it w/ valgrind markings

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6980)

### DIFF
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -41,12 +41,12 @@ Install the prerequisites using [Homebrew](https://brew.sh/):
 
 ```
 brew install nasm ragel binutils coreutils libtool autoconf automake cmake makedepend \
-sdl2 lua@5.1 luarocks gettext pkg-config wget
+sdl2 lua@5.1 luarocks gettext pkg-config wget gnu-getopt
 ```
 
-You will also have to ensure Homebrew's gettext is in your path, e.g., via
+You will also have to ensure Homebrew's gettext & gnu-getopt are in your path, e.g., via
 ```
-export PATH="/usr/local/opt/gettext/bin:${PATH}"
+export PATH="$(brew --prefix)/opt/gettext/bin:$(brew --prefix)/opt/gnu-getopt/bin:${PATH}"
 ```
 See also `brew info gettext` for details on how to make that permanent in your shell.
 
--- a/kodev
+++ b/kodev
@@ -780,7 +780,6 @@ TARGET:
                 ;;
             -r | --sane-return)
                 export SANE_RETURN=1
-                shift
                 ;;
             -H | --help)
                 echo "${RUN_HELP_MSG}"
--- a/kodev
+++ b/kodev
@@ -563,7 +563,8 @@ OPTIONS:
                              supported model: kobo-aura-one, kindle3, hidpi
     --gdb=X                  run with debugger (default: nemiver)
     --graph                  graph memory use (requires gnuplot)
-    --valgrind=X             run with valgrind (default: valgrind --trace-children=yes)
+    --valgrind=X             run with valgrind (default: valgrind --tool=memcheck --trace-children=yes --leak-check=full --track-origins=yes --show-reachable=yes)
+    --callgrind              run with valgrind's callgrind (valgrind --tool=callgrind --trace-children=yes)
 
 TARGET:
 
@@ -622,6 +623,15 @@ TARGET:
                         echo "Couldn't find valgrind."
                         exit 1
                     fi
+                fi
+                ;;
+            --callgrind)
+                # Try to use sensible defaults for valgrind
+                if command -v valgrind >/dev/null; then
+                    valgrind="valgrind --tool=callgrind --trace-children=yes"
+                else
+                    echo "Couldn't find valgrind."
+                    exit 1
                 fi
                 ;;
             -w | --screen-width)
--- a/kodev
+++ b/kodev
@@ -590,15 +590,15 @@ TARGET:
                     gdb=${VALUE}
                 else
                     # Try to use friendly defaults for gdb
-                    if command -v znemiver >/dev/null; then
+                    if command -v nemiver >/dev/null; then
                         # Nemiver is a nice GUI
-                        gdb=nemiver
+                        gdb="nemiver"
                     elif command -v ddd >/dev/null; then
                         # DDD is a slightly less nice GUI
-                        gdb=ddd
+                        gdb="ddd"
                     elif command -v cgdb >/dev/null; then
                         # cgdb is a nice curses-based gdb front
-                        gdb=cgdb
+                        gdb="cgdb"
                     elif command -v gdb >/dev/null; then
                         # gdb -tui is a slightly less nice terminal user interface
                         gdb="gdb -tui"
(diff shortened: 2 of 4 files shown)


NiLuJe:
(The final bit of trickery requires https://github.com/koreader/koreader/pull/6976/commits/a80d71187e7148ea25f6c85db7823274ee10013e).

This was mainly because I was wondering why the hell weren't finalizers properly collected on KOReader's exit.
It turned out to be: because reasons :D.

Ensuring we actually return from main instead of calling exit fixes that.

Since this is obviously only desirable when instrumenting, this is hidden behind a flag flipped by kodev (it does so automatically when running under gdb/valgrind).

Otherwise, we very much rely on our custom exit codes (and, unfortunately, that's not doable with a return. LuaJIT just checks for the pcall guard's results to chooses between EXIT_SUCCESS and EXIT_FAILURE.

kodev
```diff
@@ -173,16 +174,31 @@ usage: build <OPTIONS> <TARGET>
...
-        VALUE=$(echo "${1}" | awk -F= '{print $2}')
+    declare opt
+    declare -r E_OPTERR=85
+    declare -r short_opts="vhnd"
+    declare -r long_opts="verbose,help,no-debug,debug"
+
+    opt=$(getopt -o "${short_opts}" --long "${long_opts}" --name "kodev" -- "${@}")
+
+    # shellcheck disable=SC2181
+    if [[ $? -ne 0 ]]; then
```
Frenzie, line 192: Keep in mind I'm just skimming, but why not listen to shellcheck? Seems easier to read to me. (Also you seem to be unnecessarily using Bash `[[ ]]`? :-P)

```suggestion
    

    # shellcheck disable=SC2181
    if ! opt=$(getopt -o "${short_opts}" --long "${long_opts}" --name "kodev" -- "${@}"); then
```

kodev
```diff
@@ -596,50 +681,77 @@ TARGET:
...
-                        valgrind="valgrind --trace-children=yes"
+                        valgrind="valgrind --tool=memcheck --trace-children=yes --leak-check=full --track-origins=yes --show-reachable=yes"
                     else
                         echo "Couldn't find valgrind."
                         exit 1
                     fi
                 fi
+                shift
+                ;;
+            -c | --callgrind)
```
Frenzie, line 704: Good point

kodev
```diff
@@ -596,50 +681,77 @@ TARGET:
...
+                    kobo-forma)
+                        screen_width=1440
+                        screen_height=1920
+                        screen_dpi=300
+                        ;;
                     kobo-aura-one)
                         screen_width=1404
                         screen_height=1872
                         screen_dpi=300
                         ;;
```
Frenzie, line 749: Not sure that's really necessary? But why not I suppose. :-P

kodev
NiLuJe: I tend to default to `[[` ^^.

(And the one time I didn't was the one time I actually needed it, for the `gdb*` glob ;p

----

Yeah, the original check included a `$#` check which we don't need/handle differently, which is why it was split off. I'll tweak that ;).

pazos:
@NiLuJe:

`./kodev release macos` fails with:

```
Unsupported target for release: divh.

usage: release <OPTIONS> <TARGET>

OPTIONS:

    -d, --debug               debug-enabled release (for remote gdb)
    -i, --ignore-translation  do not fetch translation for release
    -v, --verbose             make the buildsystem more verbose

TARGET:

    kindle          Compatible with all Kindle models >= Kindle4
    kindlepw2       With compiler optimizations for Kindle models >= Paperwhite 2
    kindle-legacy   Needed only for Kindle2/3/DXG
    kobo
    cervantes
    remarkable
    sony-prstux
    android
    pocketbook
    ubuntu-touch
    appimage
    debian          Debian package for current arch
    debian-armel    Debian package for generic armel devices
    debian-armhf    Debian package for generic armhf devices
    debian-arm64    Debian package for generic 64 bits arm devices
    macos           MacOS app bundle. You need a mac to build this package
    ""
Error: Process completed with exit code 1.
```

`./kodev build` works fine but fails to run because it cannot find the path to the emulator, but that has nothing to do with this PR I think

NiLuJe:
Oh, joy, macOS must be shipping a broke-ass getopt version. >_<".

NiLuJe:
There's a keg-only gnu-getopt formula, lemme add that to the README and remember where those get installed to pick it up in the script 
-_-"

doc/Building.md
```diff
@@ -41,12 +41,12 @@ Install the prerequisites using [Homebrew](https://brew.sh/):
 
 ```
 brew install nasm ragel binutils coreutils libtool autoconf automake cmake makedepend \
-sdl2 lua@5.1 luarocks gettext pkg-config wget
+sdl2 lua@5.1 luarocks gettext pkg-config wget gnu-getopt
```
Frenzie, line 44: You'll also want to add it here:
https://github.com/koreader/koreader/blob/a90038e4ebc501c1c14e06e5a63b3b007fbf15a8/.github/workflows/build.yml#L25

(and possibly in base just to keep it in sync)

NiLuJe: Unless I have an old checkout, it's already not in sync in base ;). @pazos?

Frenzie: Is it not in sync or just not including junk it doesn't need like getopt?

NiLuJe: It's not even using the same XCode version AFAICT ^^.

NiLuJe: Slipped a comment about it in https://github.com/koreader/koreader-base/pull/1257/commits/e47b0a9cbb719fde50595e7ab0fd1ddc0d3d261a

(We're not using kodev in base, so we can do without).

Frenzie:
As an aside and probably not related to this PR, but I think programs compiled against my slightly older system's FT may be having some issues with our preps for launching the program or something.

```
nemiver: symbol lookup error: /usr/lib/x86_64-linux-gnu/libpangoft2-1.0.so.0: undefined symbol: hb_glib_script_from_script
~/src/kobo/koreader
```

NiLuJe:
I had the opposite issue, hence a3a7f72 & co... :s

NiLuJe:
Double-check that you don't have KOReader's libs in your actual `LD_LIBRARY_PATH`, and that nemiver starts properly on its own?

Frenzie:
Yes, Nemiver starts very properly on its own or with something like `nemiver ./luajit reader.lua`.

Frenzie:
Okay, I'm confused. it's working now.

PR #6988, opened 2020-12-13
wip: hardened issue template

pazos:
Shameless copy from [`brew`'s bug template](https://github.com/Homebrew/brew/blob/master/.github/ISSUE_TEMPLATE/bug.md)

Adapted to document crash.log creation and test case reports for documents. The main idea is to make our lives easier.

1. avoid incomplete bug reports
2. avoid out-of-date bug reports
3. avoid including our own instructions on the body of the new ticket
4. prevent us from explaining how-to get debug logs on each ticket

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6988)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,15 +7,37 @@ assignees: ''
 
 ---
 
+# Bug report
+
+**Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to KOReader again.**
+
+- [ ] checked that my device is actually supported?
+- [ ] updated to [last release](https://github.com/koreader/koreader/releases) or [last nightly](http://build.koreader.rocks/download/nightly/) and can still reproduce the bug?
+
+<!-- To help us debug your issue, please complete these sections: -->
+
+<!-- the full KOReader version, for example v2020.11 or v2020.11-55-ge0ac00f
+     saying "last stable" or "last nightly" doesn't age well. -->
 * KOReader version:
 * Device:
 
-#### Issue
+#### What you were trying to do
 
-#### Steps to reproduce
+A clear and concise description of the action that triggered the bug
 
-##### `crash.log` (if applicable)
-`crash.log` is a file that is automatically created when KOReader crashes. It can
+#### What happened
+
+A clear and concise description of what happened
+
+#### Step-by-step reproduction instructions
+
+All steps needed to reproduce the bug
+
+#### `crash.log`
+
+<!--
+
+`crash.log` is a file that is automatically created when KOReader runs. It can
 normally be found in the KOReader directory:
 
 * `/mnt/private/koreader` for Cervantes
@@ -25,7 +47,21 @@ normally be found in the KOReader directory:
 
 Android won't have a crash.log file because Google restricts what apps can log, so you'll need to obtain logs using `adb logcat KOReader:I ActivityManager:* AndroidRuntime:* DEBUG:* *:F`.
 
+Unless you're reporting a crash you need to enable debug logs first.
+To do that go to the `KOReader's file manager` -> `Tools` -> `More Tools` -> `Developer options` and check `enable debug logging`.
 
-Please try to include the relevant sections in your issue description.
-You can upload the whole `crash.log` file on GitHub by dragging and
-dropping it onto this textbox.
+You need to reproduce the issue again *once* debug logs are enabled.
+
+You can upload the whole `crash.log` file on GitHub by dragging and dropping it onto this textbox.
+-->
+
+#### test case (if applicable)
+
+<!--
+
+If you're reporting a bug that happens on a given document you need to attach that document to this issue or, better yet, create a test case that showcases the issue you're having.
+
+For copyright materials you'll need to scramble the document before: have a look at [ScrambleEbook calibre plugin](https://www.mobileread.com/forums/showthread.php?t=267998)
+
+-->
+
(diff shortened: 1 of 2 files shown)


poire-z:
Why not.
How will this look? Just like this?:
image
I'm not sure all users will understand they have to remove `<!-- replace me -->` and we won't see ( :) ) `<!-- my issue is this and that -->` :)

As for always enabling debug logs, that might be a bit excessive: we'll have 2Mb crash.logs to look at - and sometimes, if it ends with a Lua stack trace, that's quite enough.

(What I'd really like is for issue titles to have a length limit - to make it harder on arooni :)

Frenzie:
Other inspiration: https://obsproject.com/blog/how-to-contribute-to-open-source

Frenzie (commented):
For context
https://github.com/Homebrew/brew/commit/6676e4051cd354bf116904b09c62b8a6d36dbc1b

```
https://github.com/Homebrew/brew/pull/3085
```

Frenzie:
@poire-z 
> I'm not sure all users will understand they have to remove  and we won't see ( :) )  :)

Also note how I put it in [the feature request template](https://github.com/koreader/koreader/blob/38d5510a7870f3de07cff93b190bb4d49f6b9805/.github/ISSUE_TEMPLATE/feature_request.md), which I think has been very successful. Feature requests have been much clearer and more targeted.

> **Does your feature request involve difficulty completing a task? Please describe.**
A clear and concise description of what the problem is. Ex. I think it takes too many steps to [...]

pazos:
> How will this look? Just like this?:

Yeah.

> I'm not sure all users will understand they have to remove  and we won't see ( :) )  :)

Valid point, @Frenzie suggestion works for me.

> As for always enabling debug logs, that might be a bit excessive: we'll have 2Mb crash.logs to look at - and sometimes, if it ends with a Lua stack trace, that's quite enough.

Maybe:  "Unless you're reporting a crash you're required to enable debug logs first."

pazos:
Follow up: pushed this to my master branch too, so the changes here can be previewed in https://github.com/pazos/koreader/issues/new/choose

Frenzie:
I have to wonder if we can't keep it a little bit more concise? But looks fine in principle.

poire-z:
Not super fond of the tone :) and the heavy wrapping (10% user text for 90% static text probably). I'm usually quite fine with a single sentence and most often, they make sense even if the version/device is not provided.
(I'm quite sure we'll get unreadable stuff with people getting lost in the markdown and HTML comments - and messed up issues we'll probably have to edit to make them readable :/)
But we can try it.

pazos:
> I have to wonder if we can't keep it a little bit more concise? But looks fine in principle.

Segmentation of text based on the nature of the bug/feature?

For example, a few quick templates:

1. Bug - core (unexpected behaviour on the logic)
2. Bug - UX (unexpected behaviour of the interface)
2. Bug - document (unexpected behaviour rendering or dealing with a document)
3. Bug - device (unexpected behaviour dealing with a supported device)
4. Bug - plugin (unexpected behaviour on a specific plugin)
5. Bug - other (other kind of reproducible issue)
6. Feature - core (an enhancement in the logic of the program)
7. Feature - UX (an enhancement in the interface of the program)
7. Feature - plugin (maybe redirected to discussions and saying: ask first if something like that is nice to have)
8. Feature - document rendering (improvements on document rendering)
9. Feature - document type (maybe redirected to discussions and ...)
10. Feature - other (redirect?)

pazos:
> Not super fond of the tone :)

Is a bit harsh for me too :/. Specially the part of `If you repeatedly fail to use the issue template, we will block you from ever submitting issues to KOReader again`

poire-z:
> **Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to KOReader again.**

**Please read and fill out the issue checklist below, and provide all requested information. If you don't, a kitten will be killed and you won't be our friend. If you end up repeatedly killing kittens, we will block you from submitting issues and killing more kittens.**

Frenzie:
@poire-z 
> Not super fond of the tone :) and the heavy wrapping (10% user text for 90% static text probably). I'm usually quite fine with a single sentence and most often, they make sense even if the version/device is not provided.

I'm on the same page. I don't like the brew bug templates. They're for developers, not users.

@pazos 
> For example, a few quick templates:

That's far too many, but we do need a document bug template, mainly to direct people to programs like epubcheck.

pazos:
Leave it wip. I will try to make it soft while keeping some stuff than can make our janitor tasks easier :)

poire-z:
The following first post has some links to some issue template which happens to look like what you wanted, but also quite more html-form-like than what you got above :)
`https://github.com/arkenfox/user.js/pull/1386`

Frenzie:
> but also quite more html-form-like than what you got above :)

Yeah, [GitHub Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) are new since roughly half a year ago. They still say "Note: Issue forms are currently in beta for public repositories on GitHub.com only."

I don't like a bunch of meaningless checkboxes and too many required fields but it could be useful to absolutely require the version. It surprises me a bit but there's been a number of issues where it turned out the OP was using a version from over a year ago.

zwim:
Are there any new thoughts/insights here?

pazos:
> Are there any new thoughts/insights here?

Not really. I've commented elsewhere: https://github.com/koreader/koreader/pull/8902#issuecomment-1204059808

>> We could stick something like this on a wiki page perhaps? (Plus some stuff about those paths.)
>> https://freshrss.github.io/FreshRSS/en/contributing.html

>>> Created https://github.com/koreader/koreader/wiki/Report-a-bug as a proof of concept

pazos:
Got bored on this one. Closing :d

PR #6992, opened 2020-12-14
remarkable2 support

ddvk:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6992)


frontend/device/remarkable/event_map.lua
```diff
@@ -1,7 +1,4 @@
 return {
-    [102] = "Home",
-    [105] = "LPgBack",
-    [106] = "RPgFwd",
```
pazos, line 4: These are safe to keep. It just maps event codes that will never be fired on a RM2

NiLuJe:
This should ideally *not* murder rM1 support, but creating a new device, instead of mangling the existing one ;).

c.f., any of the other device module for inspiration ;).

ddvk:
> This should ideally _not_ murder rM1 support, but create a new device, instead of mangling the existing one ;).
> 
> c.f., any of the other device module for inspiration (e.g., kobo or kindle) ;).

should I  put the rm2 code in a separate directory?

frontend/device/remarkable/device.lua
```diff
@@ -120,5 +153,11 @@ function Remarkable:reboot()
     os.execute("systemctl reboot")
 end
 
-return Remarkable
-
+if isRM2() then
+    if not os.getenv("RM2FB_SHIM") then
+        error("remarkabl 2 requires RM2FB to work")
```
NiLuJe, line 158: nit:  `reMarkable` ;).

frontend/device/remarkable/device.lua
```diff
@@ -64,16 +86,27 @@ end
...
-    self.input.open("/dev/input/event0") -- Buttons
-    self.input:registerEventAdjustHook(adjustTouchEvt)
+    self.input.open(self.input_wacom) -- Wacom
+    self.input.open(self.input_ts) -- Touchscreen
+    self.input.open(self.input_buttons) -- Buttons
+    
+    self.input:registerEventAdjustHook(adjustAbsEvt)
+
+    if self.invertX() then
+        self.input:registerEventAdjustHook(self.input.adjustTouchMirrorX,self.mt_width)
```
NiLuJe, line 102: nit: space after the comma

pazos:
@ddvk I sent you a pr against this branch to avoid splits https://github.com/ddvk/koreader/pull/1

NiLuJe:
> should I put the rm2 code in a separate directory?

Nope, that was the right approach ;).

frontend/device/remarkable/device.lua
```diff
@@ -149,7 +149,8 @@ local f = io.open("/sys/devices/soc0/machine")
 if not f then error("missing sysfs entry for a remarkable") end
 
 local deviceType = f:read("*all")
```
NiLuJe, line 151: You can probably use `*line` instead of  `*all`, that'll strip the LF for you ;).

(There might even be a `:line` method shortcut for that).

frontend/device/remarkable/device.lua
```diff
@@ -5,21 +5,47 @@ local function yes() return true end
...
-    model = "reMarkable",
     isRemarkable = yes,
+    model = "reMarkable",
     hasKeys = yes,
     needsScreenRefreshAfterResume = no,
     hasOTAUpdates = yes,
     canReboot = yes,
     canPowerOff = yes,
     isTouchDevice = yes,
+    invertX = yes,
```
Frenzie, line 16: This line should be added to generic?

ddvk:
no, only for the rm1

frontend/device/remarkable/device.lua
NiLuJe: We've never done so for platform-specific tweaks like that (e.g., `touch_mirrored_x` on Kobo/Cervantes).

Frenzie: Not quite that generic, but it looked like an incorrect leftover at first glance based on your earlier comments.

ddvk: i'll push  it down and rename it

frontend/device/remarkable/device.lua
```diff
@@ -60,18 +78,35 @@ local adjustTouchEvt = function(self, ev)
...
+
+    if self.invertX() then
+        self.input:registerEventAdjustHook(self.input.adjustTouchMirrorX, self.mt_width)
+    end
+
+    local mt_scale_x = self.mt_width / screen_width
+    local mt_scale_y = self.mt_height / screen_height
+
+    self.input:registerEventAdjustHook(self.input.adjustTouchMirrorY, self.mt_height)
+    self.input:registerEventAdjustHook(self.input.adjustTouchScale, {x=mt_scale_x, y=mt_scale_y})
```
ddvk, line 108: this screws the pen scaling

frontend/device/remarkable/device.lua
```diff
@@ -60,18 +78,35 @@ local adjustTouchEvt = function(self, ev)
...
+    self.input:registerEventAdjustHook(adjustAbsEvt)
+
+    if self.invertX() then
+        self.input:registerEventAdjustHook(self.input.adjustTouchMirrorX, self.mt_width)
+    end
+
+    local mt_scale_x = self.mt_width / screen_width
+    local mt_scale_y = self.mt_height / screen_height
+
+    self.input:registerEventAdjustHook(self.input.adjustTouchMirrorY, self.mt_height)
```
ddvk, line 107: this affects the pen as well

frontend/device/remarkable/device.lua
pazos: @ddvk https://github.com/ddvk/koreader/pull/2 is a quck workaround.

frontend/device/remarkable/device.lua
NiLuJe: That's quite likely, and possibly hard to deal with unless there's a sure-fire way to disciminate the two solely based on the events themselves.

(The input subsystem basically chucks everything in one feed, so you can't really know where the event came from, IIRC).

pazos: @NiLuJe: give it a review on my suggestion in https://github.com/ddvk/koreader/pull/2, yay or nay?

NiLuJe: @pazos: I'd keep the `local TimeVal = require('ui/timeval')`, but, sure, why not. (Note that the fact this only handles the MT events, unlike the global hook, might be enough to skip the pen, since I hope the pen doesn't actually generate MT events ^^).

ddvk: i think i'll just revert the old adjust handler and have a version for the rm2

NiLuJe: Yep, that works, too ;).

frontend/device/remarkable/device.lua
```diff
@@ -118,5 +153,17 @@ function Remarkable:reboot()
...
 
-return Remarkable
+local f = io.open("/sys/devices/soc0/machine")
+if not f then error("missing sysfs entry for a remarkable") end
 
+local deviceType = f:read("*all") 
+if deviceType == "reMarkable 2.0\n" then
+    logger.info("rm2 ", deviceType)
+    if not os.getenv("RM2FB_SHIM") then
+        error("reMarkable2 requires RM2FB to work")
```
NiLuJe, line 163: I think you can feel free to link to the repo here, that'll make for a slightly more user-friendly experience ;).

raisjn:
is there a way to have the CI build the remarkable build on this PR so others can test it?

NiLuJe:
I don't actually remember if CircleCI has an rM workflow, but, oh, well, clicked on it anyway :D.

frontend/device/remarkable/device.lua
```diff
@@ -17,39 +29,61 @@ local Remarkable = Generic:new{
...
+    status_path = "/sys/class/power_supply/bq27441-0/status",
+}
+
+function Remarkable1:adjustTouchEvent(ev, by)
+    if ev.type == EV_ABS then
+        ev.time = TimeVal:now()
+        -- Mirror X and Y and scale up both X & Y as touch input is different res from
+        -- display
+        if ev.code == ABS_MT_POSITION_X then
+            ev.value = (Remarkable1.mt_width - ev.value) *  by.mt_scale_x
```
pazos, line 52: `ev.value = (self.mt_width - ev.value) * by.mt_scale_x`

frontend/device/remarkable/device.lua
```diff
@@ -17,39 +29,61 @@ local Remarkable = Generic:new{
...
+        if ev.code == ABS_MT_POSITION_Y then
+            ev.value = (Remarkable1.mt_height - ev.value) * by.mt_scale_y
+        end
+    end
+end
+
+local Remarkable2 = Remarkable:new{
+    model = "reMarkable 2",
+    mt_width = 1403,
+    mt_height = 1871,
```
pazos, line 63: Please append `-- unscaled_size_check: ignore` to both values in order to pass the CI

frontend/device/remarkable/device.lua
ddvk: it crashes with nil exception

ddvk: it should / could be `self.device.mt_width`

Frenzie:
> is there a way to have the CI build the remarkable build on this PR so others can test it?

Yes, but not to be merged in this repo. ;-)

The tl;dr is to at least provide the bits and pieces to create a Docker image (e.g., how to set up the TC) and then we can add it to the nightly builds. See for example:
https://github.com/koreader/virdevenv/tree/aec9d7e55e3f4f2980e2739005970ff3134d36e6/docker/ubuntu/kokobo

For testing purposes you could create a (potentially suboptimal) Docker image and use it in CircleCI to create a full build artifact if you wish.

ddvk:
any other suggestions? should I remove the wip from the title?

frontend/device/remarkable/device.lua
```diff
@@ -3,9 +3,21 @@ local logger = require("logger")
...
+local ABS_MT_POSITION_X = 53
+local ABS_MT_POSITION_Y = 54
+-- Resolutions from libremarkable src/framebuffer/common.rs
+local screen_width = 1404 -- unscaled_size_check: ignore
+local screen_height = 1872 -- unscaled_size_check: ignore
+local wacom_width = 15725 -- unscaled_size_check: ignore
+local wacom_height = 20967 -- unscaled_size_check: ignore
+local wacom_scale_x = screen_width / wacom_width
+local wacom_scale_y = screen_height / wacom_height
+local TimeVal = require('ui/timeval')
```
NiLuJe, line 18: I'd move this near the very top, between `Generic` & `logger`, to follow the style convention.

Also, in the same vein, double quotes please ;).

frontend/device/remarkable/device.lua
```diff
@@ -3,9 +3,21 @@ local logger = require("logger")
 
 local function yes() return true end
 local function no() return false end
+local EV_ABS = 3
```
Frenzie, line 6:
```suggestion

local EV_ABS = 3
```

frontend/device/remarkable/device.lua
```diff
@@ -17,39 +29,61 @@ local Remarkable = Generic:new{
     display_dpi = 226,
     -- Despite the SoC supporting it, it's finicky in practice (#6772)
     canHWInvert = no,
-    home_dir = "/mnt/root",
+    home_dir = "/home/root",
```
Frenzie, line 32: This is for both devices?

Frenzie (commented):
No remarks, just a couple of very minor nits.

frontend/device/remarkable/device.lua
ddvk: yes, the have the same file system layout

Frenzie (approved):
lgtm

ddvk:
Something minor that I noticed, starting it with `koreader.sh` without the shim results in:
```
[CLI] Failed to initialize FBInk, aborting . . .
[CLI] Failed to initialize FBInk, aborting . . .
[CLI] Failed to initialize FBInk, aborting . . .
./koreader.sh: line 209: read: read error: 0: Invalid argument
[CLI] Failed to initialize FBInk, aborting . . .
[CLI] Failed to initialize FBInk, aborting . . .
[CLI] Failed to initialize FBInk, aborting . . .
```
fyi: line 209 is: `read -r -t 15 < /dev/input/event1` 

the crash.log does contain the RM2FB  required message, but I wonder what could be done to be more user friendly

Frenzie:
You can add some handling of the shim thing in the shell script too I imagine if it's not there yet. (NB Just because that line errors doesn't mean right near that line is the best place.) But I'd say that's for a different PR.
https://github.com/koreader/koreader/blob/348d47365b1c68fae9914acb718526868b33f77a/platform/remarkable/koreader.sh#L204-L210

platform/remarkable/koreader.sh
```diff
@@ -1,12 +1,20 @@
...
 export LC_ALL="en_US.UTF-8"
-
 # working directory of koreader
 KOREADER_DIR="${0%/*}"
 
 # we're always starting from our working directory
 cd "${KOREADER_DIR}" || exit
 
+# reMarkable 2 check
+if [ "reMarkable 2.0" == "$(</sys/devices/soc0/machine)" ]; then
```
NiLuJe, line 10: Does the rM ship with Bash?

I'm not quite sure bb ash supports the io slurp operator like that...

(Also, the posix shell equality test is a *single* equal sign, much to my dismay).

NiLuJe: Which leaves the `read` builtin, we use constructs like

`IFS= read -r ORIG_FB_ROTA <"/sys/class/graphics/fb0/rotate"` elsewhere, for instance.

NiLuJe:
> You can add some handling of the shim thing in the shell script too I imagine if it's not there yet. (NB Just because that line errors doesn't mean right near that line is the best place.) But I'd say that's for a different PR.
> 
> https://github.com/koreader/koreader/blob/348d47365b1c68fae9914acb718526868b33f77a/platform/remarkable/koreader.sh#L204-L210

Which reminds me that the input device in that snippet is not quite-right on the rM 2, either (it's the pen instead of the screen).

Not sure if it's worth bothering about, though. (or just ditch the idea and use a plain sleep like on PB).

platform/remarkable/koreader.sh
Frenzie: That's exactly why I said to leave this mostly unrelated change to another PR, for future reference. Now this can't be merged like it could yesterday. ;-)

ddvk: the read doesn't bother me. the rm has bash.

Frenzie: Shellcheck would presumably complain that you're using a Bashism without using the correct shebang and so will I. ;-)

My preferred solution as a rule of thumb is not to use Bashisms unless it's really awkward not to.

Frenzie: PS Please allow CircleCI access so it'll tell you this automatically.

frontend/device/remarkable/device.lua
pazos: yup, because `self` is Input, right? :smile: 

Feel free to do want you want. If you go the `self.device.mt_width` route we can have a single adjustHook. Just use `touch_x_mirrored` or something similar as part of device definition and query its value with `self.device.touch_x_mirrored`. AFAICT that's the only different in the implementation between these two devices

frontend/device/remarkable/device.lua
```diff
@@ -118,5 +164,16 @@ function Remarkable:reboot()
     os.execute("systemctl reboot")
 end
 
-return Remarkable
+local f = io.open("/sys/devices/soc0/machine")
+if not f then error("missing sysfs entry for a remarkable") end
 
+local deviceType = f:read("*line")
```
pazos, line 170: nit: `f:close()`

pazos: Also since you always have `deviceType` you can log it. We actually do that in other platforms. For instance https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L442-L443 or https://github.com/koreader/koreader/blob/master/frontend/device/sdl/device.lua#L369

platform/remarkable/koreader.sh
```diff
@@ -1,12 +1,21 @@
...
 # working directory of koreader
 KOREADER_DIR="${0%/*}"
 
 # we're always starting from our working directory
 cd "${KOREADER_DIR}" || exit
 
+# reMarkable 2 check
+IFS= read -r MACHINE_TYPE < "/sys/devices/soc0/machine"
+if [ "reMarkable 2.0" = "$MACHINE_TYPE" ]; then
+    if [ -z $RM2FB_SHIM ]; then
```
Frenzie, line 12: ShellCheck would complain about that, and well, I've experienced enough later mysterious minor issues due to thinking quoting shouldn't be necessary in a particular spot that I think it's definitely a good habit to always do it (unless you do want to split on spaces, of course).
```suggestion
    if [ -z "$RM2FB_SHIM" ]; then
```

platform/remarkable/koreader.sh
ddvk: how do i give circleci access?

Frenzie: It used to just run on PRs, but for some reason it looks like people now have to go to https://circleci.com/ and sign up with their GitHub account.

It's very annoying too, because the CI is probably at its most useful with new contributors (not talking about you ;-) ).

But I'll have to check if maybe some config changed on their or GH's side too.

ddvk: it had failed on circleci with unauthorized

Frenzie (approved):
Alright, let's merge it. :-)

pazos (approved):
LGTM

And a healthy (and offtopic) reminder for Rm1 & Rm2 users: WiFi manager still not implemented: https://github.com/koreader/koreader/issues/6677

tealdramatic:
I just opened an issue, because I can't connect to SSH and the device is stuck on startup-screen ("remarkable is starting".)
https://github.com/koreader/koreader/issues/7076

notDavid:
Hey all, i'm a little confused so hoping someone can enlighten me; is remarkable2 support still "pre-release" or "aplha", or is it stable and ready for the rest of us?

ddvk:
the koreader rm2 support is complete (on par with rm1), but the additional
dependencies are considered unstable, but so far seems to be working quite
well.

starnix59:
@ddvk just to be sure before I create a new issue, is it expected behavior that koreader always loads ["lastfile"] and appears to be ignoring ["start_with"] = "filemanager"?

ddvk:
if you are referring to the scripted install, the script is the culprit,
the start folder is missing

pazos:
@starnix59: you'll need to check on last nightly, as the behaviour was changed recently and should work just fine as is.

starnix59:
Thank you @ddvk & @pazos bumping to the latest nightly fixed the issue.

notDavid:
Is there a release available for download anywhere which contains this fix? Like a nightly build for the rm2?

Frenzie:
There are nightly builds at http://build.koreader.rocks/download/nightly/

saulrh:
How long is it expected to take for RM2 to support to land in one of toltec's repos?

NiLuJe:
Asking in the right place would be a good start: this is upstream, you'relooking for [downstream](https://github.com/toltec-dev/toltec/tree/stable/package/koreader).

Frenzie:
At a glance it looks to me like you probably shouldn't expect it any earlier than about two months from now, even though our 2021.01 is coming very soon.

raisjn:
rM2 support has been merged into toltec stable repo 3 days ago (13th of january) and was in testing repo a week ago

Frenzie:
Oh right, I didn't glance close enough. I noticed yyhttps://github.com/toltec-dev/toltec/blob/c44fd5f465c6cd2e8d91ad2067558c6d5bcb52dd/package/koreader/package#L17 but not the patch on top of it.

https://github.com/toltec-dev/toltec/blob/c44fd5f465c6cd2e8d91ad2067558c6d5bcb52dd/package/koreader/rm2-support.patch

saulrh:
> Asking in the right place would be a good start: this is upstream, you're looking for [downstream](https://github.com/toltec-dev/toltec/tree/stable/package/koreader).

Ah, sorry. >_> IIiii also didn't notice this was a PR rather than an issue ticket. Must have set my search criteria too wide.

> rM2 support has been merged into toltec stable repo 3 days ago (13th of january) and was in testing repo a week ago

Thanks!

erahhal:
I've read elsewhere that the instructions for installing on RM2 are difference from [here](https://github.com/koreader/koreader/wiki/Installation-on-Remarkable).  When I tried launching using those instructions, nothing changes on the tablet other than it hangs, so I have to stop the process, which causes xochitl to restart.  What's the proper way to launch this?

PR #6994, opened 2020-12-14
Workaround luxl's issue with self-closing tags once and for all

### LINKED ISSUE #6993: OPDS Browser Incorrectly Handling Some Formats
* KOReader version: v2020.11-57-g222e752
* Device: Kindle Paperwhite 3

#### Issue
When opening some submenus on a amusewiki OPDS servers, only the first entry of each page is displayed.

#### Steps to reproduce

1. Add the following OPDS server entry to your catalog: https://staging.amusewiki.org/opds
2. Open the server.
3. Select Topics or Authors
4. Observe the entries and compare to the OPDS files from the server.

NiLuJe:
By basically fixing *all* of them with a Lua pattern, instead of discovering new ones every few weeks ;).

Fix #6993

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6994)

### DIFF
--- a/frontend/ui/opdsparser.lua
+++ b/frontend/ui/opdsparser.lua
@@ -78,6 +78,8 @@ function OPDSParser:parse(text)
     text = text:gsub('<content type="xhtml">.-</content>', '')
     -- luxl doesn't handle XML comments, so strip them
     text = text:gsub("<!%-%-.-%-%->", "")
+    -- luxl is also particular about the syntax for self-closing, empty & orphaned tags...
+    text = text:gsub("<content/>", "<content />")
     -- luxl prefers <br />, the other two forms are valid in HTML, but will kick luxl's ass
     text = text:gsub("<br>", "<br />")
     text = text:gsub("<br/>", "<br />")
--- a/frontend/ui/opdsparser.lua
+++ b/frontend/ui/opdsparser.lua
@@ -79,7 +79,7 @@ function OPDSParser:parse(text)
     -- luxl doesn't handle XML comments, so strip them
     text = text:gsub("<!%-%-.-%-%->", "")
     -- luxl is also particular about the syntax for self-closing, empty & orphaned tags...
-    text = text:gsub("<(%a+)/>", "<%1 />")
+    text = text:gsub("<(%l+)/>", "<%1 />")
     -- We also need to handle the slash-less variants for br & hr...
     text = text:gsub("<br>", "<br />")
     text = text:gsub("<hr>", "<hr />")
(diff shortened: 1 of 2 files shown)


NiLuJe:
(Double-checked that this didn't regress on my various Calibre testcases from #6902, and it deals with #6993's `<content/>` tags.

frontend/ui/opdsparser.lua
```diff
@@ -78,16 +78,11 @@ function OPDSParser:parse(text)
     text = text:gsub('<content type="xhtml">.-</content>', '')
     -- luxl doesn't handle XML comments, so strip them
     text = text:gsub("<!%-%-.-%-%->", "")
-    -- luxl prefers <br />, the other two forms are valid in HTML, but will kick luxl's ass
+    -- luxl is also particular about the syntax for self-closing, empty & orphaned tags...
+    text = text:gsub("<(%a+)/>", "<%1 />")
```
Frenzie, line 82: Characters like `=`, `'` and `"` count as letters? This can be things like ``.

NiLuJe:
Quick question: tag names are supposed to be lowercase, right?

frontend/ui/opdsparser.lua
NiLuJe: I sure hope not, I'm blindly assuming it's basically `[a-Z]` ^^.

Frenzie:
> Quick question: tag names are supposed to be lowercase, right?

If you mean XHTML tag names, theoretically yes. In XML `<STRONG>` is a different element than `<strong>`. In practice, who knows? ;-) In any case `<STRONG>` is well-formed XML, even if it's meaningless XHTML.

frontend/ui/opdsparser.lua
Frenzie: Well, then there are a lot of self-closing tags that'll still close issues, surely? ;-) But it's more efficient than manually adding the next `<hr/>`, yes.

NiLuJe:
Oh, well, the rest of the mangling only deals with lowercase anyway, so, at least this way it's consistent with the rest ;p.

frontend/ui/opdsparser.lua
NiLuJe:
```
> foo="<blah/>" 
> print(foo:gsub("<(%l+)/>", "<%1 />"))
<blah />        1
> foo='<img src="wheee"/>'
> print(foo:gsub("<(%l+)/>", "<%1 />"))
<img src="wheee"/>      0
```

IIRC, as long as there are attributes, luxl is fine with the space-less `/>`, so this ought to do ;).

Frenzie: > as long as there are attributes, luxl is fine with the space-less 

Well, all right then. O_o

NiLuJe:
> (Double-checked that this didn't regress on my various Calibre testcases from #6902, and it deals with #6993's `<content/>` tags.)

Hell, my Calibre stuff even works without scrapping the whole `content` block now, whee!

(In case someone ever wants to make something out of what's in there one day ^^).

PR #6996, opened 2020-12-14
Calibre plugin fixes

### LINKED ISSUE #6943: Can't change the Inbox directory for Calibre plugin when I change the SD Card
* KOReader version: 2020.11
* Device: Motorola G9 Play with Android 10

#### Issue
I've installed the app and set it up with a SD Card. I've set the home directory and the Calibre plugin inbox directory to a directory in the SD Card. Then I have received a new SD Card, switched the cards in the phone, and now I can't change the Inbox directory in the Calibre plugin, it continues going to the old path in the old SD Card and there's no way to navigate to the internal storage or the new SD card. I have even tried going back to the old card, but it keeps saying that the path is not accesible and I can't change it.

#### Steps to reproduce
Set Calibre plugin Inbox directory to a path in the SD Card. Afterwards change the SD Card to another one.

pazos:
Force primary partition on Android file chooser
Scan external volumes on Android

Fixes #6943

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6996)

### DIFF
--- a/frontend/ui/downloadmgr.lua
+++ b/frontend/ui/downloadmgr.lua
@@ -14,7 +14,8 @@ Example:
                     -- reinitialize dialog
                 end)
             end,
-    }:chooseDir()
+        }:chooseDir()
+    end
 ]]
 
 local PathChooser = require("ui/widget/pathchooser")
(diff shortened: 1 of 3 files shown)


poire-z (approved):
Looks allright for the downloadmgr stuff, if that's why you ping'ed me :)

PR #6997, opened 2020-12-14
Radius corners of more boxes

johnbeard:
This gives a more uniform look than having some boxes rectangular and some
with radiused corners

----
I probably missed some...

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6997)


NiLuJe:
Just to say that I've been testing all the new stuff with this applied, and it hasn't bothered me ;).

It probably does fit in rather well with the new icons, actually ;).

Issue #6850, opened 2020-11-05
FR: swipe to dismiss dictionary popup


doozan:
It would be nice to have the option of swiping to dismiss the dictionary popup (and probably other popups, as well).  I use the dictionary feature regularly and hitting the tiny x to close the dialog always feels like more effort than it should be.

NiLuJe:
Tapping outside of the popup works, too,

doozan:
I have the dictionary popup set as "large window" so there is no outside area, but that's a good workaround for other popups. Thanks!

doozan:
I poked at this awhile and I don't see a good way to accomplish it: swiping up/down scrolls the page, swiping left/right changes dictionaries, tapping the article changes dictionaries/page, tapping the title bar sets the default dictionary. Switching to the smaller dictionary view and tapping outside of it seems to be the simplest solution.

Issue #6679, opened 2020-09-19
Wikipedia integration defaults to "en" regardless of the ePub language


gbyl:
* KOReader version: 2020.09 (and earlier)
* Device: Kobo Forma & reMarkable (and potentially all devices)

#### Issue
When doing a long press on text and choosing "Wikipedia" from the menu, it runs the highlighted text through "Wikipedia EN" ignoring the language tag in the ePub's `content.opf` file.

#### Example
If I open a German book – which is properly formatted (the OPF file contains: `<dc:language>de</dc:language>`) and long press over the word `Geschichte` in a paragraph, it searches the English version of Wikipedia and opens `The Neverending Story` entry. What it should do instead is open the `History` entry in German.

One could manually add the language code in the settings, but the "Translate" functionality from the same menu identifies the language tag successfully every single time already. Is there a way to make it so the Wikipedia plugin does the same thing?

Adding this as an option (or making it the default) would make KOReader an even more outstading language learning tool. I hope it is not too hard to implement.

poire-z:
Some opinion of mine on that previously stated in https://github.com/koreader/koreader/issues/2476#issuecomment-270405792 .

I don't think it should be the default behaviour.
Wikipedia is mainly an encyclopedia, not a definitiion dictionary or translation dictionary.
I mainly read French, but can read English. If I read some English book/article mentionning a location or a person name (or even a concept like "Aufklarung"), I expect getting results from my main own language Wikipedia: French. If I lookup a german word like "erwartet", Wikipedia DE is as useless as Wikpedia FR or EN.

But one can set the languages codes one can read, and it's just easy to toggle from one to another depending on the book you're reading (would be even easier with something like in the screenshots of #4395): an english book with location and names: I would keep Wikipedia FR - and english book with lots of concepts in "-ism", I could switch to Wikipedia EN.
Best to let that decision to the user, rather than to pick that from book metadata (which might be wrong: some web french acticle I save as EPUB get lang=en in their EPUB metadata).

gabyhk:
@poire-z Those are good points. I was not aware this had been discussed before. Making it the default behavior might be going too far and I was not counting on some files having the wrong metadata (or lacking the metadata altogether).

The way it is at the moment is probably already best for the general user. That drop-down menu idea does sound interesting though. Another option could be to add the ability to set the language per book (perhaps somewhere in the bottom menu) rather than globally. This setting would then drive the source language for the dictionary, the translator and wikipedia (each with a toggle).

[...]

Issue #6007, opened 2020-03-27
Make enabled dictionaries stick with documents


rtega:
This is a feature request. I'm reading a lot of documents in 5 different languages. Each time I switch between two documents I have to reset the dictionaries which is rather annoying.

The ideal solution would be that a document declares itself to be in a certain language and that dictionaries can be grouped in language groups. Since this is quite involved and not something that would be easily added to koreader and since not all documents declare themselves to be in a language anyway, let me propose a rather simple half solution to this problem which would make life for people switching between documents in different languages regularly a bit more simple:
- once opened a document retains it's enabled dictionaries
- a newly opened document inherits the enabled dictionaries of the previous document

Enabled dictionaries would thus become part of the state of a document along with other settings like position, layout settings, etc. Is this possible?

poire-z:
> The ideal solution would be that a document declares itself to be in a certain language and that dictionaries can be grouped in language groups. 

That might not be ideal :) Even if a document advertized itself as being FR, I might still find in it quotes in latin, greek, german, that I'd like to lookup in the proper dicts.

> Enabled dictionaries would thus become part of the state of a document along with other settings like position, layout settings, etc. Is this possible?

I guess. But that might be bothering in some other cases. If I get a new dict, and find it better than another one, I'd have to go thru all my previously opened books to uncheck the old one and check the new one.
So, may be it should be dependant on another checkbox in Dict settings: `[ ] dedicated dictionary list for this book` (which could be made a global default so you get your wished behaviour).

rtega:
That would indeed be nice!

Issue #5437, opened 2019-09-29
Dictionary with synonym file is slow


robert00s:
Recently I created a big Polish dictionary with many synonyms (over 2 millions) in stardict format. Unfortunately, sdcv performance is very low. When I'm looking for a word, it takes over 3 seconds on my i7 laptop (over 15 seconds on my kindle) :(
I know this is not a problem with KOReader but with sdcv. I am pretty sure that this method is not optimal.

https://github.com/Dushistov/sdcv/blob/a500176661e9c77492219fcbc0d133807dc6b197/src/stardict_lib.cpp#L831-L854

```c
bool SynFile::load(const std::string &url, gulong wc)
{
    struct stat stat_buf;
    if (!stat(url.c_str(), &stat_buf)) {
        MapFile syn;
        if (!syn.open(url.c_str(), stat_buf.st_size))
            return false;
        const gchar *current = syn.begin();
        for (unsigned long i = 0; i < wc; i++) {
            // each entry in a syn-file is:
            // - 0-terminated string
            // 4-byte index into .dict file in network byte order
            glib::CharStr lower_string{ g_utf8_casefold(current, -1) };
            std::string synonym{ get_impl(lower_string) };
            current += synonym.length() + 1;
            const guint32 idx = g_ntohl(get_uint32(current));
            current += sizeof(idx);
            synonyms[synonym] = idx;
        }
        return true;
    } else {
        return false;
    }
}
```

Each time sdcv is started, the `SynFile :: load` method creates large table synonyms, so it takes a long time. I think this is enough to search for only a word in the * .syn file and read index (idx) to take a definition. 
I'm not a c++ guy, so maybe someone have some advice :) 

Link to my dictionary:
https://github.com/robert00s/sjp/releases

ifo file:

```
StarDict's dict ifo file
version=3.0.0
bookname=Słownik Języka Polskiego
wordcount=150356
idxfilesize=2822544
synwordcount=2396723
```

Info about stardict format: https://github.com/huzheng001/stardict-3/blob/master/dict/doc/StarDictFileFormat

poire-z:
Well, I guess that if a synonym file is just a nearly flat file with no index (unless it has to be ordered, in which case it can act as an index, and sdcv does not use it that way - it could use some dichotomy to find the word), it's doomed to be slow.

I learned c++ while learning crengine - but I absolutely don't plan on learning sdcv :)
[...]

PR #6998, opened 2020-12-14
[WIP] Dictionary flow/ui tweak

Mel-kior:
This remove completely dummy results and replace them with proper info messages.
InfoMessage add timeout, easy tap outside dismiss and nice and simple look.

I didn’t manage to do a test condition for the "no_refresh_on_close" like the one in showDict().

Tested on Kobo H2O, seems working properly

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6998)

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -330,6 +330,17 @@ function ReaderDictionary:addToMainMenu(menu_items)
                     UIManager:show(items_font)
                 end,
                 keep_menu_open = true,
+                separator = true,
+            },
+            { -- setting used by dictquicklookup
+                text = _("Show lookup window on cancel"),
+                checked_func = function()
+                    return G_reader_settings:nilOrTrue("dict_lookuponcancel")
+                end,
+                callback = function()
+                    G_reader_settings:flipNilOrTrue("dict_lookuponcancel")
+                end,
+
             }
         }
     }
@@ -706,11 +717,10 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
     -- early exit if no dictionaries
     if dictDirsEmpty(dict_dirs) then
         final_results = {
-            {
-                dict = "",
-                word = word,
-                definition = _([[No dictionaries installed. Please search for "Dictionary support" in the KOReader Wiki to get more information about installing new dictionaries.]]),
-            }
+            info_message = true,
+            message = _([[No dictionaries installed. Please search for "Dictionary support" in the KOReader Wiki to get more information about installing new dictionaries.]]),
+            timeout = nil,
+            icon_file = "resources/info-error.png",
         }
         return final_results
     end
@@ -773,16 +783,41 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
         end
     end
     if #final_results == 0 then
-        -- dummy results
-        final_results = {
-            {
-                dict = "",
-                word = word,
-                definition = lookup_cancelled and _("Dictionary lookup canceled.") or _("No definition found."),
-            }
-        }
+        if lookup_cancelled then
+            if G_reader_settings:nilOrTrue("dict_lookuponcancel") then
+                final_results = {
+                    {
+                        dict = "",
+                        word = word,
+                        definition = _("Dictionary lookup canceled."),
+                    }
+                }
+            else
+                final_results = {
+                  info_message = true,
+                  message = _("Dictionary lookup canceled.") .. "\n" .. _("Search : ") .. word .. "\n" .. _("Tap to edit"),
+                  timeout = 1,
+                  icon_file = "resources/info-i.png",
+                  word = word,
+                  dummy_lookup = {
+                      {
+                          dict = "",
+                          word = word,
+                          definition = _("Dictionary lookup canceled."),
+                      }
+                  }
+                }
+            end
+        else
+          final_results = {
+              {
+                  dict = "",
+                  word = word,
+                  definition = _("No definition found."),
+              }
+          }
+        end
     end
-
     return final_results
 end
 
@@ -808,12 +843,7 @@ function ReaderDictionary:stardictLookup(word, dict_names, fuzzy_search, box, li
 
     if Device:canExternalDictLookup() and G_reader_settings:isTrue("external_dict_lookup") then
         Device:doExternalDictLookup(word, G_reader_settings:readSetting("external_dict_lookup_method"), function()
-            if self.highlight then
-                local clear_id = self.highlight:getClearId()
-                UIManager:scheduleIn(0.5, function()
-                    self.highlight:clear(clear_id)
-                end)
-            end
+            self:clearHighlightShedule()
         end)
         return
     end
@@ -823,7 +853,52 @@ function ReaderDictionary:stardictLookup(word, dict_names, fuzzy_search, box, li
     end
 
     local results = self:startSdcv(word, dict_names, fuzzy_search)
-    self:showDict(word, tidyMarkup(results), box, link)
+    if results["info_message"] then
+        self:showNoResultMessage(results)
+    else
+        self:showDict(word, tidyMarkup(results), box, link)
+    end
+end
+
+function ReaderDictionary:showNoResultMessage(result)
+    self.dict_info_message = InfoMessage:new{
+        text = result["message"],
+        timeout = result["timeout"],
+        icon_file = result["icon_file"],
+        dismiss_callback = function()
+            self:clearHighlightShedule()
+        end,
+    }
+    -- disable no_refresh_on_close so dismissing LookupInfo works if visible
+    -- (maybe we can we test if LookupInfo IM is fully under the dict_info_message IM ?)
+    if self.lookup_progress_msg then
+        self.lookup_progress_msg.no_refresh_on_close = nil
+    end
+
+    if result["timeout"] ~= nil then -- If cancel search IM
+        self.timeoutCleanHighlight = function()
+            self:clearHighlightShedule()
+        end
+        -- Allow user to get the lookup window if tapping the IM
+        self.dict_info_message.dismiss_callback = function()
+            self:showDict(result["word"], tidyMarkup(result["dummy_lookup"]), box, link)
+            UIManager:unschedule(self.timeoutCleanHighlight) -- dont remove highlight if in lookup
+        end
+        -- no dismiss_callback when IM timeout so we scheldule it
+        UIManager:scheduleIn(result["timeout"], self.timeoutCleanHighlight)
+    end
+    UIManager:show(self.dict_info_message)
+    self:dismissLookupInfo()
+end
+
+function ReaderDictionary:clearHighlightShedule()
+    if self.highlight then
+        -- delay unhighlight of selection, so we can see where we stopped
+        local clear_id = self.highlight:getClearId()
+        UIManager:scheduleIn(0.5, function()
+            self.highlight:clear(clear_id)
+        end)
+    end
 end
 
 function ReaderDictionary:showDict(word, results, box, link)


poire-z:
> This remove completely dummy results and replace them with proper info messages.
InfoMessage add timeout, easy tap outside dismiss and nice and simple look.

Mhhh. As much as I don't care about "no dict installed", when we get "no definition found" in a dict lookup window, we can still do things: look it up on Wikipedia, or tap on the word to edit/correct it and use it as a base for a new lookup.
And when the selection is bad (happens on PDF), you can cancel it (instead of waiting) and correct it in the dict window.
With an InfoMessage, you don't have all these possibilities to go forward.

Mel-kior:
I didn’t take that into account indeed.
I never use wikipedia ( main use offline and I got the wikipedia dict ) and I didn’t know you can edit your search! Btw I have to hold, not tap on my H2o for editing.

I started this because of accidental search when holding the reader. You cancel the search but you can’t continue reading before closing the dict window, witch is impossible one handed. In that case the InfoMessage is perfect. 

So maybe make it an option so we can chose between both behavior when canceling search ?
Or gesture dependent, tap cancel vs hold cancel ? ( Not very user friendly )

For the "no definition found" I agree dict lookup is better but a easier way to close it would be great.

NiLuJe:
Do we actually show something when the search has *actually* been *canceled* (as opposed to the "no match" popup)?

(i.e., tap on the "Searching..." IM to cancel).

poire-z:
> I started this because of accidental search when holding the reader. You cancel the search but you can’t continue reading before closing the dict window, witch is impossible one handed. In that case the InfoMessage is perfect.
> So maybe make it an option so we can chose between both behavior when canceling search ?

I guess you have `[x] Large window` dict results and that's why you have a hard time dismissing it. 
So, I guess the option is already there: don't use a large window :)
Similar to #6850 - where the person gave up finding a solution.

You can try to see at making it a little less tall than what it is:
https://github.com/koreader/koreader/blob/a90038e4ebc501c1c14e06e5a63b3b007fbf15a8/frontend/ui/widget/dictquicklookup.lua#L230-L239
Just checked on my Kobo: it let the footer be shown, which is indeed a bit hard to hit to dismiss - but on my little emulator, it exceeds the screen size - so there's a bit of a computation error in here.
But I like this current large window when viewing full page Wikipedia - but in Wikipedia, the bottom right button is actually <kbd>Close</kbd>, so it's easier to close.
On the initial dict result, the bottom right button is <kbd>Search</kbd>.
May be we can just add a hold_callback to that <kbd>Search</kbd> - so those in the know can dismiss it that way ?
Or have the large window size configurable or multiple size options (so-so).

> Do we actually show something when the search has actually been canceled (as opposed to the "no match" popup shown after an unsuccessful lookup)?

Yes, we show "lookup cancelled" the say way we show "no result". Allows doing the other things from there in both cases. In both cases it is useful.

Mel-kior:
> I guess you have [x] Large window dict results and that's why you have a hard time dismissing it.
So, I guess the option is already there: don't use a large window :)

Yep, but not using it is a no-go for me. Small window is not practical with most dict.
And you still have have to close it. On slow e-ink it’s a good 1/2s.
I think not having the window in the first place is better for this case.

Making a setting to choose between IM or the dict lookup when canceling a search would be ok?

I looked how the tap to cancel get detected and it seem impossible to detect a hold press too, right?
https://github.com/koreader/koreader/blob/83e148bc9e426b8e7413648509ec0ad83155859f/frontend/apps/reader/modules/readerdictionary.lua#L754-L755


> May be we can just add a hold_callback to that Search - so those in the know can dismiss it that way ?

Yes, that would be a good alternative way to dismiss when using large window.

poire-z:
> And you still have have to close it. On slow e-ink it’s a good 1/2s.
I think not having the window in the first place is better for this case.
Making a setting to choose between IM or the dict lookup when canceling a search would be ok?

I don't think we should have this option. An InfoMessage might have looked to you like an easy solution - and may be it is, but for your own personal patch - it's just something that feels not many people but you would like.
And it would be just for "no result found" : if you have more than 10% of your lookups resulting in that, well, update your dicts :) and an InfoMessage would help you with these 10%. You would still need to close the 90% others with the same difficulties.
And it would be also for "lookup cancelled", because of your _accidental search when holding the reader_. If you do that a lot, you may consider enabling `Gear > Document > Highlight action > [x] Ask with popup dialog` and `[x] Enable on single word selection` : from there, you can dismiss easily - you'll just need another tap to launch the dict lookup.

> I looked how the tap to cancel get detected and it seem impossible to detect a hold press too, right?

I didn't consider that, but I don't think that's really useful.

I think the best options if you can't do with the the above suggestions are:
- better size the large window: it might be a bit bogus - taller than my emulator screen - dunno how it looks on your device, how much of the footer do you see? Although I like how the full wikipedia looks on my Kobo, just letting the footer with the hour visible.
- in the large window, for dict lookup - either globally or with an option - replace bottom right button <kbd>Search</kbd> with <kbd>Close</kbd>.

(And the other option that I use a lot: I patch my own koreader to solve my little personnal annoyances that don't feel like they would annoy many people, or would not please some others, or would require too much work from me to make my little hack a more proper patch.)

poire-z:
I'm willing to rework the sizing of DictQuickLookup - because I did the large window stuff some years ago when I didn't have much experience with the UI code and... it's really crappy :) and I can probably do it better now.

Pinging @doozan (from #6850) as another large dict window user, so we can try to solve both issues.

Currently, we have these various bottom buttons:
<kbd>image</kbd>
from bottom to top: small dict result, small wikipedia result, full article wikipedia.

I don't really know how you have the large dict window on your devices, because I have it like this on the emulator (its top and bottom borders out of screen):
image

But I have it somehow like this on my Kobo when viewing full article wikipedia - and I like it that way:
image
I initially wanted it just like this, so we can see the footer and the hour.
Now, I'm not sure you're all getting it like that - and if you have enough bottom room to tap and dismiss it.

So, what should we do ?
**A** just ensure we have enough and consistent bottom room to:
**A1** show the footer (do you see it ? do you want it ? do you prefer the large window to cover the full screen and show nothing of the PDF text or footer below ? (I personally want to see the footer)
**A2** allow tapping there to dismiss it (how much is needed)?

**B** rework bottom buttons so we always have <kbd>Close</kbd> as the bottom right button - so it's always available to close it.
I guess it's hard to get rid of <kbd>Search</kbd> as there's no other way to seach a single word/name.
We could move the optionnal <kbd>Follow link</kbd> in an optionnal third row ? Should if then be above or below the 2 existing rows ?.
We could also just have a full width bottom 3rd row with a long <kbd>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Close&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</kbd> button.

Thoughts?

doozan:
On my Kobo A2, the visible bottom margin is halfway between what you see in the emulator and what you see on your Kobo: about half of the footer is visible, so small than I never even considered trying to tap it to dismiss the window.  I just tested it and I'm unable to dismiss the dialog by trying to precisely tap that margin at the bottom (my device seems a little fussy about detecting presses around the edges, which is why it's annoying to try to hit the [X] in the top-right corner to dismiss the dialog). Surprisingly, I was able to dismiss the large dictionary window by pressing the left edge of the screen. Pressing the top triggers the title bar set-dictionary-as-default action, and pressing the right side activates the scrollbar events.

I'd vote for making the bottom margin large enough that it displays not just the footer but the last line or two of text as a visual hint that it's a tapable target to return to the document.

Mel-kior:
For the "no result found" – as I said before – I agree the dictionary lookup seems to be better for a majority of people so should **not** be replace with IM.

For the "cancel search" there seems to be 2 case :
- the accidental where IM is better
- the bad selection where edit is needed. In that case we can have : 
  - dict lookup > hold name to edit > type search 
good for typo in pdf or word modification but slower
  - IM > dismiss > hold the correct search
good for wrong selection and fast but no edit capability ( power user can still bind dict lookup to a gesture )

Adding a way to get both with gesture ( tap and hold ) would be perfect but don’t seems easily feasible.
So adding it as a option so people can choose what fit them best feels good ( getting farther but making it document dependent may be even better ).

Options "large window" and "ask with popup" aren’t very good solution neither because they add drawback/sluggish flow elsewhere.

**About the resize of the large windows :**
On my kobo H2o the window touch ( pixel perfect ) the top of the footer ( container height = 14 ).

I don’t think it’s a good idea to add bottom margin to get some background to tap ( especially if we add the close button ). 
I personally would make it real fullscreen, removing top and side border to gain more usefull space.
Does it really need to be a movable window ?

I don’t stay long enough in lookup to need the footer so I would cover it. But some may want it so a option to choose would be good.
Making the margin dependent of the footer container height seems needed to get consistent view for everyone.

**About buttons :**
I’m for adding a <kbd>Close</kbd> button.
Removing <kbd>Search</kbd> seems ok for me, I don’t see much use case for the search and it’s still accessible via the long press menu popup, binded to a gesture or via top menu. But people heavily using it may need it there.
Removing <kbd>Follow link</kbd> seems good too. But there again some may need it.

Making them user selectable as option is for me the best solution. And make it create a third row if all are selected.

Frenzie:
> Thoughts?

I think it's pretty good the way it is, but a Close button would be more as generally expected.

One potential way to save some space is by shrinking some buttons to icons instead of text.

https://commons.wikimedia.org/wiki/File:Wikipedia%27s_W.svg

256px-Wikipedia's_W svg

marker

Mel-kior:
Reworked the initial PR :
- "no result" -> lookup window
- "no dict installed" -> IM
- "canceled search" -> lookup or IM
I made it a toggle option in dict config menu.
If IM is used, we can still get the lookup window if we dismiss the IM before it times out.

Frenzie:
> "no dict installed" -> IM

No time atm, but note the newer built-in download functionality.

poire-z:
> `Show lookup window on cancel`
> If IM is used, we can still get the lookup window if we dismiss the IM before it times out.

You phrased it genericaly - and you made it ok so there's a way to get the "current" behaviour even for the person who has unchecked that.
But I still don't like it. I mean, I don't like adding new options for very strange use cases - and may be just because we didn't do something good enough so it felt needed.

OK, you, and may be others, might happen to sometimes have to hold the device by long-pressing the screen, and have a dict lookup happening. This might not be an issue for small window users - but it is with the current hideous "large" window, which is really hard to dismiss (tested on my Kobo for the first time: even I got stopped for a few seconds: how do I get it out...). If it had been easier to dismiss, you might never have thought about this idea of bringing InfoMessage in the play.
Except for the InfoMessage timeout, a proper large Window with a Close button and enough room at the bottom so it's easy to dismiss, and your problem is solved. Without the need for more options.

I feel the same about the numerous options you thought about in the discussion :) :

> So adding it as a option so people can choose what fit them best feels good ( getting farther but making it document dependent may be even better ).
> I don’t stay long enough in lookup to need the footer so I would cover it. But some may want it so a option to choose would be good.
> About buttons [...] Making them user selectable as option is for me the best solution. And make it create a third row if all are selected.
> after thinking about it, making all buttons optional ( even prev/next dict, wikipedia and highlight ) so we cant get only one or no row if we don’t need them would be the best for maximum dictionary space and clean ui.

Even if I personally hack a few things that bothers me out in my own patch, I'd still prefer the shipped KOReader to be simple and that its UI is intuitive and clever and usable in a variety of contexts and use cases, without cluttering it with too many options.

> Does it really need to be a movable window ?

It could be a non-movable window, but it's nice it's movable, so when reading the dict, you can move it out to see the word in the book text context. If you don't need it, just don't move it :)
Same with all the buttons: you may not need prev/next/Wikipedia (I don't need Highlight). The room gained is minimal if you use a large window - and having a UI toolkit to select which buttons and in which order to show them would be overkill and excessive configuration crazyness.

poire-z:
I'm currently reworking DictQuickLookup - I don't plan on changing the appearance (except for proper large window sizing and adding a Close button), just making its coding right and simpler. It's really a mess :/ and hard to get right...

While I'm at it:
we have TAP on the window title - which says which dict the result is from - to make that dict "the default dictionary for this document".
I always found this bothering, as a tap there is so easy. I patched that out of my own patch.
I now realize the way it's done it might even rarely work. May be it works better with CJK dicts/words, but with latin text, it's a mess.
https://github.com/koreader/koreader/blob/348d47365b1c68fae9914acb718526868b33f77a/frontend/ui/widget/dictquicklookup.lua#L721-L748
If the original order of results (now, we can sort dicts, but this is global) is `House | HOUSE | House` and the 3rd dict is the one set as default, the 2nd will stop showing the 3rd.
Anyway, I don't plan on fixing it. I'm just asking: should we remove that "feature" ?

Frenzie:
> If the original order of results (now, we can sort dicts, but this is global) is House | HOUSE | House and the 3rd dict is the one set as default, the 2nd will stop showing the 3rd.
Anyway, I don't plan on fixing it. I'm just asking: should we remove that "feature" ?

I'm not sure I quite understand.

poire-z:
`House | HOUSE | House ` was just me giving an example of getting results from 3 dicts, which each can give back the "lookup word" with different cases : that's enough to make the algo in the above snippet fail to put the 3rd result first if one has set its dict as the "default dict for this document" - because the algo stopped because `"HOUSE" != "house"`.
I don't plan on explaining it more or find ways to fix it - as it never bothered anyone before.
My real and unique question is: should we remove that "feature" ? :)

Frenzie:
Being case insensitive sounds better.

poire-z:
> Being case insensitive sounds better.

Given my real and unique question :) I take it you mean that I should not remove that "feature" (that has been buggy for years - at least with my french dicts - and its little annoyance that it is triggered on any tap on the title bar) ?

doozan:
>  My real and unique question is: should we remove that "feature" ? :)

I'd vote to remove it. It's never worked for me and it's bad UI/UX: there's no visual indication of which dictionary has been selected as the default, there's no hint that tapping the title bar will switch default dictionaries, and there's no easy way to undo an accidental tap on the title bar beyond switching to the desired dictionary and tapping it again.

If you want to keep the functionality, maybe move it to something like a hamburger icon to the title bar that exposes the list of dictionaries and lets you select the default.

Frenzie:
> I take it you mean that I should not remove that "feature" (that has been buggy for years - at least with my french dicts - and its little annoyance that it is triggered on any tap on the title bar) ?

It seems sensible to me to display a dictionary that matches properly over one that doesn't?

Frenzie:
Now if you're talking about the tap thing, sure, get rid of that, although the functionality itself is good.

Mel-kior:
> You phrased it genericaly - and you made it ok so there's a way to get the "current" behaviour even for the person who has unchecked that.
> But I still don't like it. I mean, I don't like adding new options for very strange use cases - and may be just because we didn't do something good enough so it felt needed.
> 
> OK, you, and may be others, might happen to sometimes have to hold the device by long-pressing the screen, and have a dict lookup happening. This might not be an issue for small window users - but it is with the current hideous "large" window, which is really hard to dismiss (tested on my Kobo for the first time: even I got stopped for a few seconds: how do I get it out...). If it had been easier to dismiss, you might never have thought about this idea of bringing InfoMessage in the play.
> Except for the InfoMessage timeout, a proper large Window with a Close button and enough room at the bottom so it's easy to dismiss, and your problem is solved. Without the need for more options.

I’m not trying to solve my problem ( I got a patch ) but to fix what I think is a design flaw.
Canceling a search should do what it mean, and not going though to a new lookup.

- Accidental search ( holding, sleepy finger, liquid drop, etc ), you don’t want the lookup window.
- Wrong selection, you don’t want the lookup. Cancel and redo the search.
- Change of mind ( e.g. you remember the search ), you don’t want the lookup.

The lookup window is only needed if bad OCR/selection and you want to edit.

When you don’t want the lookup, still having it have 2( 3 ) drawbacks :
- you don’t have a timeout so one more action is needed
- you get distracted from where you at in the text or where you touched exactly when you did the wrong selection
- you get more delay ( e-ink ) because lookup window give more screen refresh ( current implementation, this may get better ) 

Getting the lookup window only for fast edit seems not the best for overall smooth and fast using experience. Specially if you can still get to edit if you want ( we can even go directly to the edit window so you get there faster ).

> Even if I personally hack a few things that bothers me out in my own patch, I'd still prefer the shipped KOReader to be simple and that its UI is intuitive and clever and usable in a variety of contexts and use cases, without cluttering it with too many options.

I think it’s only cluttered with too many options if the option menu is a mess. If well thought and done ( I don’t say it’s easy ), option can cover more use case.

For example, "size of the large window", or "footer visibility" can be in a sub menu e.g. "Tweak & optimization" and let the main dictionary configuration option menu free of secondary/small/usecase specific options. New and not worrying user get a simple and clear main options but if more specific/personal options are needed they can still have them.

> Same with all the buttons: you may not need prev/next/Wikipedia (I don't need Highlight). The room gained is minimal if you use a large window - and having a UI toolkit to select which buttons and in which order to show them would be overkill and excessive configuration crazyness.

I was thinking of only options to select which button but then we auto place them, with somes getting specific places ( chevrons, close ).

**About the lookup rework  :**

I agree the "default dict for this doc" should get a proper ui indication/icon.
Like an empty star next to dict name. That get filled if defaulted and can be tap again to undo?

And about the "proper match" bug I think I got some hiccup too with the dict "Littré" that got full caps words. But I rarely use the function so I dont mind. However adding string.lower() so everything is lowercase may be a easy fix?

The hold to edit search is not indicated too, again a edit icon next to the name can be a solution.

**Other thing/idea that would be great to have  :**
- Do the search async and give the result as soon they are found. So you don’t have to wait that all dict are looked up.
Actually I reduced my dict to 5 to get a average 10sec lookup ( with fuzzy ). Having async would give possibility to have more dicts without drawback
- Having multiples dict "lists" were you can select dict and their order. The list can then be switch on the go or get auto assigned to document depending of their language or other tags.
- Preprocess dict so they all got consistent formatting, in the device or do a preprocess dict repository.

poire-z:
> to fix what I think is a design flaw.
Canceling a search should do what it mean, and not going though to a new lookup.

I don't consider it a _design_ flaw. Possibly a _naming_ flaw :) as it's more _interruping_ a seach than _cancelling_ a seach. And that's what I expect from that (for a long time, we couldn't interrupt it, we just had to wait till it's done no matter the context of the need for interrupting).

> Accidental search, Wrong selection, Change of mind
> When you don’t want the lookup, still having it have drawbacks :
> you don’t have a timeout [...] you get distracted  [...] you get more delay 

Except for "Wrong selection", I consider the others as edge rare cases that don't need any tweak - and the drawbacks are so little given the context (and should help you getting less accidental and to make up your mind :)
And usually, lookups are fast: you might get your result in less than 1 second. So, the tap to cancel/interrupt has as much chance to happen before or just after the lookup is displayed. I think it is better to always display the dict window, knowing than after taping to interrupt, whether it interrupted the lookup, or whether it ended before/while you taped, you get to something similar: the lookup window - that indeed you might want to tap again to dismiss (but then, it's consistent: 2 taps - and it's more consistent for you, large window users, where the 2nd tap won't close the window).

> The lookup window is only needed if bad OCR/selection and you want to edit.

Usually, my lookups are fast. It happens sometimes that it takes quite a bit (from 5 to 20s). When 5s has elapsed, depending on what I selected (person name, some person names are in my dicts), I usually interrupt because I somehow know I won't get any valuable result even if I wait 20s - and I do go some Wikipedia (one might also wish to highlight it and do the seach elsewhere later).
Or when I highlight a german word, I know I won't go to Wikipedia, so I wait 20s to possibly get some result from my german dict.
Well, that's my use cases :) So, I may be biased in that, but I consider them less "edge rare cases" than your examples :/

Anyway, other people opinions on this are welcome.

> I think it’s only cluttered with too many options if the option menu is a mess. If well thought and done ( I don’t say it’s easy ), option can cover more use case.
For example, "size of the large window", or "footer visibility" can be in a sub menu e.g. "Tweak & optimization" and let the main dictionary configuration option menu free of secondary/small/usecase specific options. New and not worrying user get a simple and clear main options but if more specific/personal options are needed they can still have them.

2 thoughts about that:
- if we go liberal on that, we'll get menus with tons of _personal_ options :) (and believe me, I will add a lot to them :)
- it makes the code larger and more complex, with more combination cases to test - and maintainability more difficult (now, with my rework of dictquicklookup, I only have to test large or not large window, which is ok).

NiLuJe:
I'm still utterly flabbergasted by the "lookups are slow" reports.

I have three (or four?) dicts on the Forma, one of those beign gigantic, and I can't even *see* the lookup IM most of the time. o_O.

NiLuJe:
(As far as the actual discussions go, my use-cases is I don't use large windows, and lookups are fast, so, err, pass? :D).

Frenzie:
> The lookup window is only needed if bad OCR/selection and you want to edit.

That doesn't describe my use at all. Fuzzy search usually selects a lemma like `select` for `selects` but that's far from certain. To me it makes more sense to let it search for a second and then to edit it down into a better lemma than to perform 50 actions. Perhaps this is more of an issue in French, which is where I primarily or exclusively use the dictionary feature.

Other than that, roughly what @NiLuJe said, except I do use it in large window atm.

doozan:
> I'm still utterly flabbergasted by the "lookups are slow" reports.
> 
> I have three (or four?) dicts on the Forma, one of those being _gigantic_, and I can't even _see_ the lookup IM most of the time. o_O.

Slow dictionary lookups may have something to do with #5437. If you have any dictionary with a large synonyms index, all lookups are extremely slow.  The root issue is bad code in sdcv that loads and parses the entire synonyms list on every lookup instead of treating it as a sorted index and doing a fast search the way it looks up the index words. As a workaround, you can use rebuild the dictionary with the synonyms in the main index instead of the secondary index, which results in very speedy lookups.  It might be worth adding a notice if a dictionary is detected with a .syn file larger than some arbitrary size.

poire-z:
About the other comments (thanks!) on my questions:

> The hold to edit search is not indicated too, again a edit icon next to the name can be a solution.

That's an idea. A pen icon on the right may be. Although when you don't care about it, having it on all lookup results might be heavy/distracting. Or with a very light gray?

> I'd vote to remove it. It's never worked for me and it's bad UI/UX

> I agree the "default dict for this doc" should get a proper ui indication/icon.
Like an empty star next to dict name. That get filled if defaulted and can be tap again to undo?
And about the "proper match" bug I think I got some hiccup too with the dict "Littré" that got full caps words. But I rarely use the function so I dont mind. However adding string.lower() so everything is lowercase may be a easy fix?

I think I'll remove it. Even with a proper indicator, it's too fragile to get right.
And it allow only one single override, only one dict can be moved first.
Implementing https://github.com/koreader/koreader/issues/6007#issuecomment-604998768 might be a more generic/general alternative to it. 

> Having multiples dict "lists" were you can select dict and their order. The list can then be switch on the go or get auto assigned to document depending of their language or other tags.

Solving the above issue might half solve this (also see my comment in there about lang tags - also #6679).

> Do the search async and give the result as soon they are found. So you don’t have to wait that all dict are looked up.
Actually I reduced my dict to 5 to get a average 10sec lookup ( with fuzzy ). Having async would give possibility to have more dicts without drawback

All that is limited by the use of sdcv, and sdcv capabilities :/

> Preprocess dict so they all got consistent formatting, in the device or do a preprocess dict repository.

Not KOReader's job :) (fortunately !)

poire-z:
About:

> still having it have 2( 3 ) drawbacks :
> -    you don’t have a timeout so one more action is needed
> -    you get distracted from where you at in the text or where you touched exactly when you did the wrong selection
> -    you get more delay ( e-ink ) because lookup window give more screen refresh ( current implementation, this may get better )

Given the recent #7012 , and if it can help with "accidental long-press" and avoiding the above drawbacks, I might be OK with an interrupting Tap in the 0.5s-1.5s range (lower than the time required by a human to see the "Searching dictionnary...", focus on the word, read the word, and take a decision - or that I could count in my head "one, two seconds" :) would just interrupt the search without showing any window (no dictquicklookup, no infomessage).

It still doesn't solve the "chances were that the lookup was really fast and you get the window before your interrupting tap was processed" case, and the effect of the tap is still randomely inconsistent.

(I do count "one, two, three, four" when I want a long-long-press to not do a dict and show the highlight dialog - such little hidden features are not really advertized, but are discoverable.)

Mel-kior:
> Slow dictionary lookups may have something to do with #5437. If you have any dictionary with a large synonyms index, all lookups are extremely slow. The root issue is bad code in sdcv that loads and parses the entire synonyms list on every lookup instead of treating it as a sorted index and doing a fast search the way it looks up the index words. As a workaround, you can rebuild the dictionary with the synonyms in the main index instead of the secondary index, which results in very speedy lookups. It might be worth adding a notice if a dictionary is detected with a .syn file larger than some arbitrary size.

My slow lookup seem to be that, when I remove one of my dict with a big .syn file I go from ~10 to ~2sec lookup. I will merge synonym to see if it is really the problem.

> That's an idea. A pen icon on the right may be. Although when you don't care about it, having it on all lookup results might be heavy/distracting. Or with a very light gray?

There is already the dict number so I don’t think a small light gray icon can be that distracting. 

> Implementing #6007 (comment) might be a more generic/general alternative to it.

#6007 seems to propose to do the dict selection/order for every new document. I think preset lists are better in a sense that you don’t have to redo them every book. And being auto selected depending of language would be optional ( or not there at all ).

> Given the recent #7012 , and if it can help with "accidental long-press" and avoiding the above drawbacks, I might be OK with an interrupting Tap in the 0.5s-1.5s range (lower than the time required by a human to see the "Searching dictionnary...", focus on the word, read the word, and take a decision - or that I could count in my head "one, two seconds" :) would just interrupt the search without showing any window (no dictquicklookup, no infomessage).

I think that a great despite the to fast lookup problem. And this will not remove the current behavior, so it’s good for everyone.

Frenzie:
@doozan just submitted a patch upstream, so with a little guard against the current segmentation fault (an assumtion about syn files that breaks down in the wild, I assume) hopefully that'll be a problem of the past soon. ;-)

poire-z:
OK, so far this is what I get (ok, not much visual change :) but the UI components buildup was a bit nightmarish to redo correctly while trying to not get much change to what we're used to see):

<kbd>image</kbd>

For the ultra rare case when we lookup a word in a link:
<kbd>image</kbd>

The large window one should now be consistent (fixed padding from screen edges) and adjust to the footer height.
image

Does the padding above the word and below the definition looks enough ? We had more previously, we can tweak them, but it felt strange to not have the scroll bar reach that near the bottom buttons.

@johnbeard : I used this [square-edit-outline.svg](https://github.com/Templarian/MaterialDesign/blob/master/svg/square-edit-outline.svg) icon (there's a similar Unicode codepoint, but it's too thick). If you feel like making it consistent with the others, lighter than this one (but we would use it with a smaller size than the others in the menu, but it's fine it this one looks grayish, less distracting) with possibly minimal padding, especially on the right), thanks :)  Or if you have ideas for a better icon, free free to suggest alternatives.

NiLuJe:
Stupid question, because the icon looks like it means "edit" which doesn't make much sense in this context, so: what does it do? ;o).

poire-z:
> because the icon looks like it means "edit" which doesn't make much sense in this context, so: what does it do? ;o).
Pops up a text input for manual lookup?

Exactly what it means and you half guessed :) 
It replaces the non-obvious long-press on the result word (`Elève 1/7`) to start a new manual text input, with that string as a pre-fill (so, you can edit, replace suffixes, correct it).
Actually with a twist (for the better I think):
```lua
    local lookup_edit_button = IconButton:new{
        icon = "edit",
        callback = function()
            -- allow adjusting the queried word
            self:lookupInputWord(self.word)
        end,
        hold_callback = function()
            -- allow adjusting the current result word    (that's what was done previously, less useful)
            self:lookupInputWord(self.lookupword)
        end,
```
while keeping the current dict lookup if you hit cancel (previously, it was closed as soon as you hold

johnbeard:
You could just try the MaterialDesignList pencil icon by itself? https://github.com/Templarian/MaterialDesignLight/blob/master/icons/svg/pencil.svg

It should be compatible with our icons out of the box.

johnbeard:
And in terms of UX for the top bar, perhaps have a tap (or long-press) bring up the list of dictionaries and from that list, tap to change dictionary (like tapping >> a few times) or long-press to made default (same method as the usual menu areas)? And use a star to indicate the default as usual?

poire-z:
> You could just try the MaterialDesignList pencil icon by itself pencil.svg

Does not work so well I think: 
<kbd>image</kbd>

The other one, showing a pencil in a window, express more a new window will open.
But opinions welcome.

> And in terms of UX for the top bar, perhaps have a tap (or long-press) bring up the list of dictionaries and from that list, tap to change dictionary (like tapping >> a few times) or long-press to made default (same method as the usual menu areas)? And use a star to indicate the default as usual?

Too much work, and having "first result from default-set dict" work is complicated (it currently hardly never works for me).
Above, we were a few (2 :) thinking tap on the title bar is bad UX. Long-press is already used for moving the window (and that's my first target to move it with, like I do on Windows :).
Do you mean it worked for you and you'd like to keep that feature?

johnbeard:
> Do you mean it worked for you and you'd like to keep that feature?
Not really used it at all TBH. I was just trying to think of a way that would be more obvious what the title bar tap did, but there's still no affordance. Outsourcing to a menu button of some sort makes more sense to me than hiding things behind magical tap areas.

With box (radius and width matches the icon page outlines):

[pencil-box.zip](https://github.com/koreader/koreader/files/5727229/pencil-box.zip)
pencil-box

If it's too small, I can adjust it so it's, say, a 36px icon.

poire-z:
Thanks, looks quite fine to me:

<kbd>image</kbd>

(May be I should add a bit of top padding and/or make it a bit smaller - thru our widget toolkit.)

> If it's too small, I can adjust it so it's, say, a 36px icon.

It's SVG, so we don't care about the native size you set :)
(Looks like you make it 48 like all the others, so I guess it's perfect.)

johnbeard:
Yes, but you may still care about the relative line widths being at least roughly the same:

[pencil-box.zip](https://github.com/koreader/koreader/files/5727254/pencil-box.zip)
pencil-box
pencil-box-36

Using the 36px version would effectively thicken the lines by about 25%.

poire-z:
> Using the 36px version would effectively thicken the lines by about 25%.

OK, I finally get it :)

In this context, we won't see the other icons, so nothing to compare with.
Also, currently, I made it be the size of the font size (its line height actually) use for the word - and made that size used for the word not become smaller than the font size set by the use for the definition... so, it might have different sizes... :/

I'll check tomorrow which give the overall better impression.

johnbeard:
Also if I may make a nitpick: having the "1/7" next to the definition makes it look to me like it's the third definition _in that dictionary_, rather than the (only) entry from the first dictionary of 7. Everything else inside the window frame refers to the entry from a single dictionary at a time. Would it make more sense to put it up in the title bar like "1/7: Dictionnaire...", or is that too wasteful of titlebar space?

NiLuJe:
> Also if I may make a nitpick: having the "1/7" next to the definition makes it look to me like it's the third definition _in that dictionary_, rather than the (only) entry from the first dictionary of 7.

Agreed.

Another idea: a pipe separator between 'em. Or one of them aligned on the right edge?

doozan:
> having the "1/7" next to the definition makes it look to me like it's the third definition _in that dictionary_, rather than the (only) entry from the first dictionary of 7.

If it's possible, it would be nice to make this field display 1/X definitions counter and then put the 1/7 dictionary counter in the titel bar.

poire-z:
You mean showing 1/14 in the definition or word, and 1/7 in the title if each of the 7 dictionary would bring 2 definitions each ?
Well, that would be useless information flood to me :)
Currently, the 1/7 is the number of results, whether they come from 1, 2, 5 or 7 of the 12 dicts we have installed.
I don't think we need more than that - we just need to know how many definitions we can still browse if we're not satisfied with the current one.

And I don't want to overcrowd or change much of this little widget we're all so used to see.
(In case we ever need more actions, no need for a menu, we could just add small icons alongside the new edit icon - but only if really needed.)

doozan:
I understand now, I was confused about what exactly the current counter represented. In that case, I think I prefer it where it is rather that in the title bar, since it's indicating the count of the available definitions and not the count of the dictionaries.

poire-z:
Current / nb of results in the title bar:

<kbd>image</kbd

<kbd>image</kbd

On the right, on the left of the icon:

<kbd>image</kbd>

No preference (yet :)

Frenzie:
I'd just keep it where it is (or maybe further on the right if you think that looks nicer). Putting it next to the dictionary name suggests something about number of dictionaries rather than results, imho.

NiLuJe:
@Frenzie does raise a good point, so I'd vote for the third option ;).

johnbeard:
> In case we ever need more actions

What about "set current dictionary as default"? Keep the long press?

poire-z:
> What about "set current dictionary as default"? Keep the long press?

It will be removed. See bottom of https://github.com/koreader/koreader/pull/6998#issuecomment-749281399

johnbeard:
> > What about "set current dictionary as default"? Keep the long press?
> 
> It will be removed. See bottom of [#6998 (comment)](https://github.com/koreader/koreader/pull/6998#issuecomment-749281399)

Oh right, I thought that was just the long-press not the whole action.

poire-z:
OK, some stuff done in #7029, and a zip file for those wanted to test it.
@Mel-kior : I did (3rd commit) the "show nothing if interrupted quickly". It works - but now that you got your dicts lookup fast, not sure it's really interesting. Please tell how this works for you.

@johnbeard : I used your 36px version, thanks.

Mel-kior:
@poire-z Even with fast lookup I find it useful. But the delay feel a little too short for me some times. The time you react to the popup – with a slow e-ink refresh – you got to be fast to get it.
Can we make that inputs get block for 0.5sec when the lookup window appear ? So that if we try to cancel when the screen refresh we don’t mess with the lookup ?
I will test it more to see in more real use cases.

Find some bug too. I will post them in #7029. ( maybe close this PR and continue there? )

poire-z:
Superseded by #7029.

poire-z:
> I'm still utterly flabbergasted by the "lookups are slow" reports.

Been doing some tests, and it looks like my slowness is caused by a single dict (no synonym):
```
54190794 Jan 11  2018 wikt-en-Western-2016-11-07_fixed.dict.dz
77284227 Jan 11  2018 wikt-en-Western-2016-11-07_fixed.idx
  493522 Dec 31 12:17 wikt-en-Western-2016-11-07_fixed.idx.oft
     474 Jan 11  2018 wikt-en-Western-2016-11-07_fixed.ifo
```
From there I think : http://www.dictinfo.com/

If I remove it, I get all my results in less than 2 seconds. (But I won't, cause it's real nice and helpful :).

Frenzie:
That seems somewhat surprising. Does it have exponentially more entries than the other dictionaries?

Frenzie:
@poire-z I don't notice the 2018 version being noticeably slow.

poire-z:
It does: it has a 77MB .idx file (it's a dictionary of all western languages).
All of my other normal dicts have .idx under 3MB, most under 1MB. Except some latin & greek ones (that I usually disable as too slow) which have a .idx around 25MB.

Frenzie:
This one:
Screenshot_2021-01-07_17-09-05

Frenzie:
> It does: it has a 77MB .idx file (it's a dictionary of all western languages).

I see, the -en- was deceptive I suppose.

doozan:
Just for an additional point of reference, I use a Spanish dictionary with a 40MB .idx and it has no noticeable delay. Is your dictionary still slow if you use sdcv on your pc?

poire-z:
With fuzzy search enabled?
If yes, fast also with a strange word (say, some conjugaison form that is not in the index) which won't match an exact word and so fallback to fuzzy search?
I also don't know what's your device :) kobo/kindle, or super fast top android ?

doozan:
I'm on a kobo v2.  You're right, with fuzzy enabled, misses and fuzzy matches are almost unforgivably slow.

```
[root@(none) koreader]# ls -alh data/dict/es-en.*
-rwxr-xr-x    1 root     root        1.6M Jan  7 11:14 data/dict/es-en.dict.dz
-rwxr-xr-x    1 root     root       38.3M Jan  7 11:14 data/dict/es-en.idx
-rwxr-xr-x    1 root     root      223.0K Jan  7 12:25 data/dict/es-en.idx.oft
-rwxr-xr-x    1 root     root         220 Jan  7 11:14 data/dict/es-en.ifo
```

```
[root@(none) koreader]# time ./sdcv -e -n -2 data/dict -u "Wiktionary (es-en)" NotInTheDictionary
Nothing similar to NotInTheDictionary, sorry :(
real	0m 0.01s
user	0m 0.00s
sys	0m 0.01s
```
```
[root@(none) koreader]# time ./sdcv -n -2 data/dict -u "Wiktionary (es-en)" NotInTheDictionary
Nothing similar to NotInTheDictionary, sorry :(
real	0m 1.67s
user	0m 1.50s
sys	0m 0.16s
```
```
[root@(none) koreader]# time ./sdcv -n -2 data/dict -e -u "Wiktionary (es-en)" tenessee
Nothing similar to tenessee, sorry :(
real	0m 0.01s
user	0m 0.01s
sys	0m 0.00s
```
```
[root@(none) koreader]# time ./sdcv -n -2 data/dict -u "Wiktionary (es-en)" tenessee
Found 2 items, similar to tenessee.
-->Wiktionary (es-en)
-->tenerse
[snip]

-->Wiktionary (es-en)
-->Tennessee
[snip]

real	0m 2.11s
user	0m 1.94s
sys	0m 0.16s
```

PR #6999, opened 2020-12-15
Fix docs

Frenzie:
Follow-up to <https://github.com/koreader/koreader/pull/6973>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6999)

### DIFF
--- a/frontend/util.lua
+++ b/frontend/util.lua
@@ -220,15 +220,15 @@ function util.secondsToHClock(seconds, withoutSeconds, hmsFormat)
     end
 end
 
---- Converts timestamp to an hour string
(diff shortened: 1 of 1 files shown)


NiLuJe:
My original ill-advised idea was to leave the final note rambling about MS CRTs outside of the ldoc stuff, but I don't have any particular strong feelings about that, so, that's entirely fine by me :).

NiLuJe:
And, in fact, the later edits I made to that comment probably do warrant making it to the docs, so, yep!

NiLuJe (approved):

PR #7011, opened 2020-12-19
Implement proper alpha-blending of SVG icons

NiLuJe:
Also, instead of doing that every time, cache a pre-composited version
that matches the screen's BB type.
This is faster, and also has the advantage of making icon highlights
behave.

Jot down a few notes about corner-cases or future improvements, e.g.,
dimming icons on non-white backgrounds, and nightmode with color icons.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7011)

### DIFF
--- a/frontend/ui/widget/iconwidget.lua
+++ b/frontend/ui/widget/iconwidget.lua
@@ -35,7 +35,7 @@ local IconWidget = ImageWidget:extend{
     -- be overriden by callers.
     width = Screen:scaleBySize(DGENERIC_ICON_SIZE), -- our icons are square
     height = Screen:scaleBySize(DGENERIC_ICON_SIZE),
-    alpha = true, -- our icons have a transparent background
+    alpha = false, --- @note: our icons have a transparent background, but we flatten them at caching time.
     is_icon = true, -- avoid dithering in ImageWidget:paintTo()
 }
 
--- a/frontend/ui/widget/iconwidget.lua
+++ b/frontend/ui/widget/iconwidget.lua
@@ -35,7 +35,7 @@ local IconWidget = ImageWidget:extend{
     -- be overriden by callers.
     width = Screen:scaleBySize(DGENERIC_ICON_SIZE), -- our icons are square
     height = Screen:scaleBySize(DGENERIC_ICON_SIZE),
-    alpha = false, --- @note: our icons have a transparent background, but we flatten them at caching time.
+    alpha = false, --- @note: our icons have a transparent background, but we flatten them at caching time, and this flag is only checked at blitting time.
     is_icon = true, -- avoid dithering in ImageWidget:paintTo()
 }
 
--- a/frontend/ui/widget/imagewidget.lua
+++ b/frontend/ui/widget/imagewidget.lua
@@ -426,7 +426,7 @@ function ImageWidget:paintTo(bb, x, y)
         -- NOTE: MuPDF feeds us premultiplied alpha (and we don't care w/ GifLib, as alpha is all or nothing),
         --       while NanoSVG feeds us straight alpha.
         --       SVG icons are currently flattened at caching time, so we'll only go through the straight alpha
-        --       codepath for non-icons PNGs.
+        --       codepath for non-icons SVGs.
         if self._is_straight_alpha then
             -- NOTE: Our icons are already dithered properly, either at encoding time, or at caching time.
             if Screen.sw_dithering and not self.is_icon then
--- a/frontend/ui/widget/imagewidget.lua
+++ b/frontend/ui/widget/imagewidget.lua
@@ -419,7 +419,10 @@ function ImageWidget:paintTo(bb, x, y)
         -- Only actually try to alpha-blend if the image really has an alpha channel...
         local bbtype = self._bb:getType()
         if bbtype == Blitbuffer.TYPE_BB8A or bbtype == Blitbuffer.TYPE_BBRGB32 then
-            do_alpha = true
+            -- NOTE: Icons are flattened at caching time!
+            if not self.is_icon then
+                do_alpha = true
+            end
         end
     end
     if do_alpha then
--- a/frontend/ui/widget/imagewidget.lua
+++ b/frontend/ui/widget/imagewidget.lua
@@ -178,6 +178,45 @@ function ImageWidget:_loadfile()
                     self._bb = RenderImage:scaleBlitBuffer(self._bb, math.floor(bb_w * DPI_SCALE), math.floor(bb_h * DPI_SCALE))
                 end
             end
+
+            -- Now, if that was *also* one of our icons, and it has an alpha channel,
+            -- compose it against a background-colored BB now, and cache *that*.
+            -- This helps us avoid repeating alpha-blending steps down the line,
+            -- and also ensures icon highlights/unhilights behave sensibly.
+            if self.is_icon then
+                local bbtype = self._bb:getType()
+                if bbtype == Blitbuffer.TYPE_BB8A or bbtype == Blitbuffer.TYPE_BBRGB32 then
+                    local icon_bb = Blitbuffer.new(self._bb.w, self._bb.h, Screen.bb:getType())
+                    --- @note: Should match the background color. Which is currently hard-coded as white ;).
+                    ---        See the note below in paintTo for how to make the dim flag behave in case
+                    ---        this no longer actually is white ;).
+                    icon_bb:fill(Blitbuffer.COLOR_WHITE)
+
+                    -- And now simply compose the icon on top of that, with dithering if necessary
+                    -- Remembering that NanoSVG feeds us straight alpha, unlike MµPDF
+                    if self._is_straight_alpha then
+                        if Screen.sw_dithering then
+                            icon_bb:ditheralphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        else
+                            icon_bb:alphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        end
+                    else
+                        if Screen.sw_dithering then
+                            icon_bb:ditherpmulalphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        else
+                            icon_bb:pmulalphablitFrom(self._bb, 0, 0, 0, 0, icon_bb.w, icon_bb.h)
+                        end
+                    end
+
+                    -- Free the original icon w/ an alpha-channel, keep the flattened one
+                    self._bb:free()
+                    self._bb = icon_bb
+
+                    -- There's no longer an alpha channel ;)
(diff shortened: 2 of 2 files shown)


NiLuJe:
@poire-z: I checked the [arrow overlap](https://github.com/koreader/koreader/pull/6977#issuecomment-748471340) thingy. I don't recall how it's supposed to look, but it was transparent before this, and is still transparent after, so I guess we're good?

(I imagine it doesn't go through IconWidget, which is why it's unaffected by this PR?).

ko_arrow_overlap

frontend/ui/widget/imagewidget.lua
```diff
@@ -406,13 +449,27 @@ function ImageWidget:paintTo(bb, x, y)
...
+    -- In night mode, invert all rendered images, so the original is
     -- displayed when the whole screen is inverted by night mode.
-    -- Except for our black & white icon files, that we do want inverted
-    -- in night mode.
+    -- Except for our *black & white* icons: we do *NOT* want to invert them again:
+    -- they should match the UI's text/backgound.
+    --- @note: As for *color* icons, we really *ought* to invert them here,
+    ---        but we currently don't, as we don't really trickle down
+    ---        a way to discriminate them from the B&W ones.
+    ---        Currently, this is *only* the KOReader icon in Help, AFAIK.
```
Frenzie, line 472: Doesn't look half bad, anyway.

Screenshot_2020-12-19_17-57-36

Frenzie (approved):

NiLuJe:
The other approach to bitonal color icons would be kinda similar to the dimming trick: actually store 'em as an alpha-mask, and colroBlit'em the requested color at runtime ;).

Frenzie:
@NiLuJe I think you might be able to guess if I only say:

Or we could…

poire-z:
> I don't recall how it's supposed to look, but it was transparent before this, and is still transparent after, so I guess we're good?

It looks similar with v2020.11, so yes, we're not worse :) 
<kbd>image</kbd>

> (I imagine it doesn't go through IconWidget, which is why it's unaffected by this PR?).

It does go thru IconWidget:
https://github.com/koreader/koreader/blob/642978c97c0ce13a8c5830c9edca0f1eae6cedad/frontend/apps/reader/modules/readerview.lua#L107-L115
https://github.com/koreader/koreader/blob/642978c97c0ce13a8c5830c9edca0f1eae6cedad/frontend/apps/reader/modules/readerview.lua#L177-L187

NiLuJe:
Speaking of (fake) nightmode and colors, one fun thing that works out quite well is the "white as 0xDD" trick in fake night mode: that makes for a rather pleasant dark mode with much a contrast that's much less harsh than pure white on black :).

Screenshot_20201219_180232

NiLuJe:
@poire-z: Huh, well, I guess the magic happens in AlphaContainer, then ^^.

poire-z:
Yep, took me quite some time to get it back working (after IconWidget/SVG) before I get to this :) : 
```diff
     -- now have our childs paint to the private blitbuffer
-    --- @todo should we clean before painting?
+    private_bb:fill(Blitbuffer.COLOR_WHITE)
     self[1]:paintTo(private_bb, 0, 0)
```

poire-z:
> cache a pre-composited version

(Still reading this PR) but don't you need to hack the cache hash if you're caching a tweaked version ?
Or do we never get the chance to get this icon/image in some other context ? (I guess we don't, but well.)

Frenzie:
Here we go.

```diff
diff --git a/frontend/ui/renderimage.lua b/frontend/ui/renderimage.lua
index 0821aedc..b87e91e9 100644
--- a/frontend/ui/renderimage.lua
+++ b/frontend/ui/renderimage.lua
@@ -214,7 +214,22 @@ function RenderImage:renderSVGImageFileWithNanoSVG(filename, width, height, zoom
     if not NnSVG then
         NnSVG = require("libs/libkoreader-nnsvg")
     end
-    local svg_image = NnSVG.new(filename)
+
+    local file = io.open(filename, "r")
+    if not file then return nil end
+    local content = file:read("*a")
+    file:close()
+
+    content = content:gsub("#(%x%x)(%x%x)(%x%x)", function(r, g, b)
+        r, g, b = tonumber(r, 16), tonumber(g, 16), tonumber(b, 16)
+        r, g, b = 255 - r, 255 - g, 255 - b
+        r, g, b = string.format("%x", r), string.format("%x", g), string.format("%x", b)
+        return "#"..r..g..b
+    end)
+
+    local svg_image = NnSVG.new(content, true)
+
+
     local native_w, native_h = svg_image:getSize()
     if not zoom then
         if width and height then
```

Screenshot_2020-12-19_18-14-03

poire-z:
> Here we go.

There your top menu icons go :)

Frenzie:
The idea being to pass an `invert this one please = true` or an `invert_function`, of course. Not indiscriminately, but I'm not going to do that for a five-minute proof of concept. :-P

Frenzie:
Just hammering on my "it's SVG, you can do whatever you want in minutes" refrain. xD

frontend/ui/widget/imagewidget.lua
```diff
@@ -385,19 +424,23 @@ function ImageWidget:paintTo(bb, x, y)
     end
     if do_alpha then
         -- NOTE: MuPDF feeds us premultiplied alpha (and we don't care w/ GifLib, as alpha is all or nothing),
-        -- but NanoSVG feeds us straight alpha
+        --       while NanoSVG feeds us straight alpha
         if self._is_straight_alpha then
-            --- @todo if Screen.sw_dithering then use bb:ditheralphablitFrom() when it's available
-            bb:alphablitFrom(self._bb, x, y, self._offset_x, self._offset_y, size.w, size.h)
+            -- NOTE: Our icons are already preprocessed properly, either at encoding time, or at caching time.
```
poire-z, line 429: so we don't go thru here anymore as you reset `self._is_straight_alpha` ? (might be worth adding to your comment?)

poire-z (approved):

NiLuJe:
> > cache a pre-composited version
> 
> (Still reading this PR) but don't you need to hack the cache hash if you're caching a tweaked version ?
> Or do we never get the chance to get this icon/image in some other context ? (I guess we don't, but well.)

Yep, nothing actually requires the transparent version, so, no need to expose that fact to the cache ;).

If we *did* need the transparent version (e.g., dim in !white bg), then, yeah, something would have to be done to be able to at least know that it's not transparent, and ideally request the cached transparent one instead.

frontend/ui/widget/imagewidget.lua
NiLuJe: That's the "at caching time" ^^.

poire-z: It's just the context where that NOTE is: it's inside `if self._is_straight_alpha then` but talks about our icons... not saying precisely we won't ever go thru this `if` branch. But the presence of the NOTE inside the `if` makes me think it applies to this branch, that our icons will go thur it (might be just me :) but it confused me when reading it)

NiLuJe: Right, the comment actually applies to "why are we never dithering on icons" bit of the logic.

straight vs. pmul doesn't really enter the equation, as both are handled, which makes the fact that, yes, we won't actually get straight alpha with icons anymore somewhat irrelevant?

NiLuJe: That's because I tend to comment branches *before* the if when I can.

I can move it inside the branch if that helps?

poire-z: I was expecting it outside :) or I dunno which branch you're talking about :)
I think this comment just needs to be more explicite than just stating a fact: `Our icons are already preprocessed properly`. _Yay, so what? why tell me that here at this point? what should I make out of that? is there a trick?_ (that was my thinking :)
It needs an additional `so, we don't have self._is_straight_alpha on them, and we go / don't go thru here or there`

NiLuJe: But the two are entirely orthogonal ^^.

The straight alpha thingy is not at all the point of the comment ^^.

The point of the comment is to explain why it skips dithering for icons.

Frenzie:
```suggestion
            --- @note: Our icons are already preprocessed properly, either at encoding time, or at caching time.
```

Frenzie: Actually never mind… ish, it's only fixme, todo and warning that are LDoc. >_>

But elsewhere you've used `@note`.

NiLuJe: Tweaked the comments a bit ;).

----

Also, that made me realize that we weren't actually skipping alpha-blending on BB32, so I fixed that by flipping the alpha flag in IconWidget ;).

poire-z: > But the two are entirely orthogonal ^^.

I know, they were.

> The straight alpha thingy is not at all the point of the comment, and completely irrelevant: if it were still set, this would still be true and work just as well ^^.

Yes, but the point that you talk about icons in a branch where we never go with icons is confusing :) (and I agree that if we were going thru it, the comment is right).

(As you usually don't mind getting verbose in comments, may be just adding `(although we don't go thru here currently as we set straight_alpha=nil for all our icons :-P)` would make me confident I won't get stuck in the future - but ok anyways :)

NiLuJe: > Actually never mind… ish, it's only fixme, todo and warning that are LDoc. >_>
> 
> But elsewhere you've used `@note`.

Huh :D. Well, can't hurt anyway ;).

NiLuJe: > (As you usually don't mind getting verbose in comments, may be just adding `(although we don't go thru here currently as we set straight_alpha=nil for all our icons :-P)` would make me confident I won't get stuck in the future - but ok anyways :)

Sure, that should be part of the latest round of commits ;).

frontend/ui/widget/iconwidget.lua
```diff
@@ -35,7 +35,7 @@ local IconWidget = ImageWidget:extend{
     -- be overriden by callers.
     width = Screen:scaleBySize(DGENERIC_ICON_SIZE), -- our icons are square
     height = Screen:scaleBySize(DGENERIC_ICON_SIZE),
-    alpha = true, -- our icons have a transparent background
+    alpha = false, --- @note: our icons have a transparent background, but we flatten them at caching time, and this flag is only checked at blitting time.
```
poire-z, line 38: No issue with transparent PNG icons (themes, users icons)
(I guess you could imply alpha=true or do_alpha from is_icon then, dunno which is cleaner.)

NiLuJe: Yeah, that was my first implementation, but since nothing else really sets is_icon, I went straight to the source ;).

The flattening would happen for PNGs, too, so, not an issue either on that front, behavior is consistent.

poire-z (approved):

Issue #3030, opened 2017-07-27
"Searching for..." dialog for dictionary lookup is redundant


embryo10:
* KOReader version: 2015.11-1054
* Device: Kobo Glo

Whenever I long-press a word to search in the dictionary I first get a blurring fast drawing of a dialog that says that it searching for the "whatever" word I long pressed, and after that the dictionary is opened.
Since I can't even read this properly (and I have an old Kobo Glo) and the extra drawing and re-drawing of the dialog square costs energy, could we just skip it?

Frenzie:
I rather like the notification but iirc we compromised in changing it so
that the dialog would only show after 200-500 ms. That still has to be
implemented though.

poire-z:
See https://github.com/koreader/koreader/pull/2934 : that "Searching for..." popup can be disabled by setting "Disable fuzzy search", now in the Dictionary settings menu.

Frenzie:
I'm inclined to think that disabling fuzzy search is a highly undesirable side effect just to get rid of a little popup. Of course your mileage may vary. Obviously chrox thinks there's something to it. :-)

embryo10:
Is fuzzy search the searching of the word with different endings?
If its that, I wouldn't like to disable it.
Maybe a configurable (even only in lua scripts) time threshold for the dialog to appear could be more handy.

Frenzie:
Fuzzy search is the thing that matches `epuisante`, `épuisante` and `épuisantes` to `épuisant`even if only the latter has an entry in the dictionary. (Or sometimes, fuzzy search is the thing that matches `épuisant` to `lulz this word doesn't resemble the one you looked up in the slightest`.)

embryo10:
Then we sure do need this most of the time (even if it gives some false positives).
About this delay... ;o)

PR #7012, opened 2020-12-19
[UX] Wait 0.5s before showing "searching for" dialog

Frenzie:
Fixes <https://github.com/koreader/koreader/issues/3030>.

Cross-ref to <https://github.com/koreader/koreader/pull/6998>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7012)

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -819,10 +819,11 @@ function ReaderDictionary:stardictLookup(word, dict_names, fuzzy_search, box, li
     end
 
     if fuzzy_search then
-        self:showLookupInfo(word)
+        UIManager:scheduleIn(0.5, self.showLookupInfo, self)
     end
 
     local results = self:startSdcv(word, dict_names, fuzzy_search)
+    UIManager:unschedule(self.showLookupInfo)
     self:showDict(word, tidyMarkup(results), box, link)
 end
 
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -819,10 +819,11 @@ function ReaderDictionary:stardictLookup(word, dict_names, fuzzy_search, box, li
     end
 
     if fuzzy_search then
-        self:showLookupInfo(word)
+        UIManager:scheduleIn(0.5, self.showLookupInfo, self)
```
poire-z, line 822: No need to pass `word` ?

poire-z:
Works for me.
(I feared that delay could cause a small worry that the lookup didn't work, or a change of habbit :) but 0.5s is small enough that I don't notice anything - and long enough so it avoids in 50% cases showing that message. It's actually in sync with the time my GloHD takes to do it's white>black>gray animation on the selected word :)

poire-z:
I thnk there's some issue with this: depending on when you tap to interrupt the lookup, the state is sometimes a bit crappy: the word stay highlighted - or the interruption works, the message disappear, but a few seconds later, the dict lookup result appears :) 

Before this PR, we created the InfoMessage before lauching sdcv, and then, this InfoMessage was given to:
`local completed, results_str = Trapper:dismissablePopen(cmd, self.lookup_progress_msg or false)`
as the thing that will trap the interruption/dismiss.
Now, we give it `self.lookup_progress_msg = nil`, so it uses its own invisible TrapWidget - and things might then be strange.
Not sure how to handle all that correctly, and a bit busy at the moment to focus on it :/

poire-z:
I think I can get this working by reverting this PR and have InfoMessage accept a `show_delay=0.5` property, making it invisible (UIManager already deals with widget.invisible=true, used with TrapWidget) for this duration (not painted, not sertDirty'ed) while still being in UIManager stack so it can catch the dismiss tap.

PR #7013, opened 2020-12-19
macOS: Comment the brew shenanigans

NiLuJe:
(And support non-standard brew prefixes, just in case)

Depends on https://github.com/koreader/koreader-base/pull/1260

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7013)

PR #7014, opened 2020-12-19
macOS: Linting in do_mac_bundle

NiLuJe:
Button was green in #7013. It was a lie :(.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7014)


Frenzie:
Glade this is in shell, 'cause Make is bad with spaces even if you try to escape things. ;-)

PR #7018, opened 2020-12-20
Use a full-width plus for buttons paired with a unicode minus

NiLuJe:
U+FF0B vs. U+002B

Because otherwise it is not, in fact, the same width as the minus, which looks weird in buttons ;).

Re #3840 & #6939

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7018)


NiLuJe:
Example on the NL widget, because you can see classic hyphen-minus/+ in the same window, and they're much narrower.

fwplus

poire-z (approved):

Frenzie:
> Example on the NL widget, because you can see classic hyphen-minus/+ in the same window

Probably more of an oversight on my part than a feature since I don't have NL. xD

PR #7021, opened 2020-12-21
Margins: allow direct input (more_options)

### LINKED ISSUE #5978: UX Suggestion: set margin values by input (like font size)
I have a suggestion for convenience and consistency.

Font size selection has three dots to manually input the value.

How about implementing the same for margin values too?
For example, If I want to select "size 2" for bottom margin, I need to tap the first square of margin, which gives "size 5" and then tap "minus" three times. Each time I tap, the whole book is re-rendered again. This is not very effective.

yparitcher:
fixes #5978

We may want to get rid of fine_tune on them and just use the spinwidget for finer control.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7021)

### DIFF
--- a/frontend/ui/data/creoptions.lua
+++ b/frontend/ui/data/creoptions.lua
@@ -65,7 +65,6 @@ Note that this may not be ensured under some conditions: in scroll mode, when a 
                 name = "h_page_margins",
                 name_text = _("L/R Margins"),
                 buttonprogress = true,
-                fine_tune = true,
                 values = {
                     DCREREADER_CONFIG_H_MARGIN_SIZES_SMALL,
                     DCREREADER_CONFIG_H_MARGIN_SIZES_MEDIUM,
@@ -93,6 +92,17 @@ Note that this may not be ensured under some conditions: in scroll mode, when a 
                 },
                 delay_repaint = true,
                 name_text_hold_callback = optionsutil.showValuesHMargins,
+                more_options = true,
+                more_options_param = {
+                    left_min = 0,
+                    left_max = 140,
+                    left_step = 1,
+                    left_hold_step = 5,
+                    right_min = 0,
+                    right_max = 140,
+                    right_step = 1,
+                    right_hold_step = 5,
+                },
             },
             {
                 name = "sync_t_b_page_margins",
@@ -115,7 +125,6 @@ In the top menu → Settings → Status bar, you can choose whether the bottom m
                 name = "t_page_margin",
                 name_text = _("Top Margin"),
                 buttonprogress = true,
-                fine_tune = true,
                 values = {
                     DCREREADER_CONFIG_T_MARGIN_SIZES_SMALL,
                     DCREREADER_CONFIG_T_MARGIN_SIZES_MEDIUM,
@@ -143,12 +152,18 @@ In the top menu → Settings → Status bar, you can choose whether the bottom m
                 },
                 delay_repaint = true,
                 name_text_hold_callback = optionsutil.showValues,
+                more_options = true,
+                more_options_param = {
+                  value_min = 0,
+                  value_max = 140,
+                  value_step = 1,
+                  value_hold_step = 5,
+                },
             },
             {
                 name = "b_page_margin",
                 name_text = _("Bottom Margin"),
                 buttonprogress = true,
-                fine_tune = true,
                 values = {
                     DCREREADER_CONFIG_B_MARGIN_SIZES_SMALL,
                     DCREREADER_CONFIG_B_MARGIN_SIZES_MEDIUM,
@@ -177,6 +192,13 @@ In the top menu → Settings → Status bar, you can choose whether the bottom m
                 delay_repaint = true,
                 name_text_hold_callback = optionsutil.showValues,
                 help_text = _([[In the top menu → Settings → Status bar, you can choose whether the bottom margin applies from the bottom of the screen, or from above the status bar.]]),
+                more_options = true,
+                more_options_param = {
+                  value_min = 0,
+                  value_max = 140,
+                  value_step = 1,
+                  value_hold_step = 5,
+                },
             },
         }
     },
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -1280,6 +1280,10 @@ function ConfigDialog:onConfigMoreChoose(values, name, event, args, name_text, d
                     end
                 }
             end
+            if delay_repaint then
+                self._dialog_closed = true
+                self:closeDialog()
+            end
             UIManager:show(widget)
         end
         if not delay_repaint then -- immediate refresh


yparitcher:
I have no clue why the sceenshots only capture the item in focus an not the whole page.

Reader_2020-12-20_200924
Reader_2020-12-20_200927
Reader_2020-12-20_200932

NiLuJe:
Screenshots on which platform and done with what tool?

yparitcher:
> Screenshots on which platform and done with what tool?

emulator, native (2 corner tap)

NiLuJe:
> > Screenshots on which platform and done with what tool?
> 
> emulator, native (2 corner tap)

Ah. Well, I guess that answers [that](https://github.com/koreader/crengine/blob/600811fbb91b6e0ec7430fbbba6281965a4a43f1/crengine/src/lvdrawbuf.cpp#L1959-L1960) ;p.

NiLuJe:
Probably only visible since https://github.com/koreader/koreader-base/pull/1206, because it preserves the alpha byte, while we used to stomp it to fully opaque before that.

Frenzie (approved):

poire-z:
Looks fine.
A few minor hiccups, unrelated to any new code (cause this PR doesn't add any :) : 

When you change a value for the first time, the bottom panel disappear to show more of the page, so we can see the bottom margin in effect (and the number picker widget stays, which is fine).
When you do that a second time (on the same number picker), the bottom panel no longer disappear.

I also got this refresh glitch after playing multiple times (can't find out the steps to reproduce it thus :/):

<kbd>image</kbd>

Probably some stuff to look at in the ugly readerconfig.lua or configdialog.lua .
But I guess it's not a reason to not merge this - and we can fix that later.

Frenzie:
> I also got this refresh glitch after playing multiple times (can't find out the steps to reproduce it thus :/):

Does the info message reach that far down? Looks like a localized update when it should probably just be the whole screen or something.

poire-z:
> Does the info message reach that far down? 

No, the infomessage is smaller than that. That's the shadow of the number picker.

NiLuJe:
The existing fine-tuning hide/show shenanigans are already exhibiting papercut issues on fast devices, so it's not terribly surprising ;).

poire-z:
I think that might have been a _tap outside the number picker_, which repainted the book below, but not the part ot configdialog above it.
(That was on the emulator - never had a papercut glitch on it :) so, I guess this refresh issue can probably be fixed, it might just be something with regions or expecting side effects/overlap/coalescing which did not happen.)

NiLuJe:
Oh, right, if it was on the emu, then, yeah, something's definitely wonky somewhere ^^.

yparitcher:
without `fine tune`
Reader_2020-12-21_174517

yparitcher:
Something need to be done about the screenshot with crengine as i cant take one of the document. The area of the crengine page is just transparent in the screenshot. it works with mupdf though.

NiLuJe:
Yep, I'll be working on that later ;).

----

I'd like to keep the fine-tuning buttons, because it makes the workflow a tiny bit faster when you do really only need to fine-tune it by a couple points ;).

Keeping both might feel a bit overkill, but they serve slightly different purposes, and I don't think it makes the widget cramped or harder to use  ;).

yparitcher:
> I'd like to keep the fine-tuning buttons, because it makes the workflow a tiny bit faster when you do really only need to fine-tune it by a couple points ;).
> 
> Keeping both might feel a bit overkill, but they serve slightly different purposes, and I don't think it makes the widget cramped or harder to use ;).

to me they look crowded, but i will wait and see what the crowd has to say :smile:

NiLuJe:
If that's an issue, I'd rather kill a couple of the 'progress' squares ;).

poire-z:
I don't mind having both, but it indeed feels crowded.
I think I could be fine without the <kbd>-</kbd> <kbd>+</kbd> - as the new fine tuning could allow the same, with more granularity than +1/-1, so you can avoid a few re-render when you know +1/-1 won't be enough.

But only if the bottom panel also disappear on the 2++ hits. Because otherwise it is not helpful :)

1st tap on Apply: you see all the margins
image

2++ taps on Apply: you don't see the bottom one, which is usually the more important target
 image

yparitcher:
@poire-z 
the bottom panel staying on 2++ &  the refresh glitch are caused by the same thing.
(refresh glitch can be reproduced by closing the numberpicker without any adjustment.)

the bottom panel is removed on margin update by tagging it `skip_paint` (via `delay_repaint`) therefore when the margin is adjusted it is painted on **top** of the panel. this also causes any other refresh to to paint an top of the bottom panel. (easily shown by moving the numberpicker around and "erasing" the panel.) this `skip_paint` is reset on successful execution of a config option so after the first adjustment the panel is correctly painted on top of the new margin.

this is a inherent limitation of the hack used to allow showing the margins through the panel and i am not sure how to fix it.

poire-z:
> this is a inherent limitation of the hack used to allow showing the margins through the panel and i am not sure how to fix it.

Dunno... (but too late to look seriously at it).
First quick thoughts: by adding another parameter to some function `...,delay_repaint, keep_delay_repaint)`, or  setting a `self.keep_delay_repaint = true` when coming from onConfigChoose which could be checked for inrefresh_dialog_func.
These kind of quick ideas... Hacking the hack :)

NiLuJe:
Can't we just close the ConfigDialog when popping up the picker, and basically never show it again unless closing the picker?

(I don't rightly recall the code flow either right now ^^).

NiLuJe:
Again, don't remember how these works, but, if you can apply a value *and* keep the picker open, I'd be fine with getting rid of fine-tune, as you can then essentially do the same fine-tuning from within the picker.

yparitcher:
> Again, don't remember how these work, but, if you can apply a value _and_ keep the picker open, I'd be fine with getting rid of fine-tune, as you can then essentially do the same fine-tuning from within the picker itself.

The picker stays open, so there goes fine-tune.


> IIRC, the need to show the ConfigDialog again was because that's where the buttons _were_. That's no longer the case when you have a picker ;).

It is convenient to have the panel open back up as usually I tweak a few settings at a time together. (If i change one thing, I probably also change t least one other thing)

yparitcher:
I used `delay_repaint` to dismiss the `ConfigDialog` when adjusting the margin via the numberpicker, which i think is the best compromise between hack & usability.

poire-z:
We already have a 2-values picker for L/R Margin.
As when tweaking top or bottom, we may want to adjust the other, could it be the same 2-values picker allowing tweaking both top and bottom margin, when launched from either <kbd>...</kbd> ?
(Probably less simple as the current tweaks so far :)

yparitcher:
not easy,
L/R are done with one event hence one picker with 2 values, T/B are 2 separate events and `configdialog` does not provide a way to tie them together.
so out of scope for now.

yparitcher:
squash & rebase on master

Issue #7017, opened 2020-12-20
PB632: "Night mode" setting as displayed in menu, becomes inverted with actually applied setting after poweroff


JasperE84:
* KOReader version: 2020.12
* Device: Pocketbook PB632 - U632.6.1.900 / 20201001_123124 / WW

#### Issue
After powering down with night mode enabled, and then starting the device to resume a book, the "Night mode" setting will show up in menu as enabled while the applied setting is actually disabled. 
If Night Mode is subsequently disabled in the menu (checkbox empty), it will actually be enabled. 
The applied night mode setting becomes inverted from the setting state as shown in menu.

#### Screenshot
In the screenshot below, night mode is supposed to be enabled (checkbox marked) but is actually disabled (white background, black text)
![Reader_2020-12-20_170933](https://user-images.githubusercontent.com/12022716/102718297-b1097e00-42e7-11eb-88a7-a6b4e05bd096.png)

#### Steps to reproduce

1. Open ebook
2. Enable night mode
3. Wait for device to go in battery saving mode
4. Power on pocketbook 
5. KOReader will reopen book and resume with inverted night mode 

(I already reinstalled KOReader overwriting all files to see if that fixes it.)

NiLuJe:
That's a known issue in some (old) i.MX kernels: suspend breaks EPDC flags handling.

You can either live with it, or disable HW inversion on that device.

NiLuJe:
c.f., the [Kobo Aura](https://github.com/koreader/koreader/blob/90ad2bcf521860312c042e4a7c5067f161bc7e60/frontend/device/kobo/device.lua#L118-L128) for an example of what such an exception looks like in practice.

JasperE84:
I have set `canHWInvert = no,` in `\applications\koreader\frontend\device\pocketbook\device.lua` in line 43 for pocketbook defaults and added the same config line in block:

```
-- PocketBook Touch HD Plus / Touch HD 3 (632)
local PocketBook632 = PocketBook:new{
  model = "PBTouchHDPlus",
  display_dpi = 300,
  isAlwaysPortrait = yes,
  usingForcedRotation = landscape_ccw,
  hasNaturalLight = yes,
  canHWInvert = no,
}
```
(PBTouchHDPlus is indeed the modelname that shows when opening the help>report a bug dialog) 
Unfortunately that doesn't fix it. 
I might add that simply exiting KOReader and restarting it also presents the issue.

NiLuJe:
[...]

PR #7022, opened 2020-12-21
HW rotation causing PB632 inversion state issues

JasperE84:
More details here:
https://github.com/koreader/koreader/issues/7017

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7022)

### DIFF
--- a/frontend/device/pocketbook/device.lua
+++ b/frontend/device/pocketbook/device.lua
@@ -453,8 +453,8 @@ local PocketBook632 = PocketBook:new{
     model = "PBTouchHDPlus",
     display_dpi = 300,
     isAlwaysPortrait = yes,
-    usingForcedRotation = landscape_ccw,
     hasNaturalLight = yes,
+    -- Buggy kernel, no usingForcedRotation, fb reinits done by the HW rotation shenanigans forget about inversion, c.f. #7017
 }
 
 -- PocketBook Color (633)


Frenzie (approved):

NiLuJe:
Err, arguably, the inversion flag issue is an actual KOReader bug, not a PB quirk, so I'd rather *that* were fixed in base ;).

(e.g., That's kind of a NACK to merge this, because it just hides the issue on *this* specific device).

NiLuJe:
Given that your device shouldn't be using a B288 chip (or this would already have been disabled), it might actually still be necessary...

So, new round of tests (*without* this PR applied):

* Given your earlier tests, you should be able to reproduce the issue after disabling the C BB: (In the FM): Tools > More Tools > Developer options > Disable C blitter

What about with *this* applied?

```
diff --git a/ffi/framebuffer_linux.lua b/ffi/framebuffer_linux.lua
index d93e4392..7491b32c 100644
--- a/ffi/framebuffer_linux.lua
+++ b/ffi/framebuffer_linux.lua
@@ -180,7 +180,10 @@ function framebuffer:setRotationMode(mode)
         -- as well as prod the driver via new mmap() as some do tie down rotation modes to each mapping.
         self.cur_rotation_mode = mode
         self:setHWRotation(mode)
+        -- Remember the screen bb's invert flag, too
+        local inverse = self.bb:getInverse()
         self:reinit()
+        self.bb:setInverse(inverse)
         assert(self.forced_rotation, "reinit/fb hooks shouldn't flip hw rotation flags")
         if self.forced_rotation.restore then
             self:setHWRotation(self.native_rotation_mode)
```

It *should* fix it with the CBB disabled, but it *might* not with it enabled. So, try both ;).

NiLuJe:
> Given that your device shouldn't be using a B288 chip (or this would already have been disabled), it might actually still be necessary...

Actually, no, HW rota is only disabled on B288 iff the waveform presets are set to "fast".

So, new new question: I'd like to see a debug log to confirm that it's indeed not a B288. (In the same "Developer options" menu, "Enable debug logging". Restart KOReader after toggling it).

JasperE84:
Here you go. https://www.dropbox.com/s/umt0dilless7rr7/crash.log?dl=0

BTW, c blitter was already disabled in dev options

NiLuJe:
Ha!

```
DEBUG mxcfb: Disabling hwinvert on B288 chipset
```

So, at least stuff is starting to make a tiny bit of sense.

Does the patch in https://github.com/koreader/koreader/pull/7022#issuecomment-749112978 help?

JasperE84:
Yep! That fixes it with the PR reverted!

NiLuJe:
Cool, thanks :).

----

I can't really help with the rotation stuff, because that way lies madness, but it's been touched upon before, IIRC ;).

NiLuJe:
Closing in favor of https://github.com/koreader/koreader-base/pull/1264 (which will hit front Soon(TM) ;)).

Unless there are actually other HW-rotation induced issues (besides the aforementioned rotation switch between the InfoMessage and the Reader)? (Which wouldn't be surprising, it's a B288 after all...).

PR #7024, opened 2020-12-22
Android: Don't mark the platform as should_restrict_JIT

NiLuJe:
And see how much stuff breaks \o/

Depends on https://github.com/koreader/android-luajit-launcher/pull/278 (to hopefully break less/more stuff :D).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7024)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -355,7 +355,7 @@ function FileManagerMenu:setUpdateItemTable()
             end,
         })
     end
-    if not Device.should_restrict_JIT then
+    if not (Device.should_restrict_JIT or Device:isAndroid()) then
         local Blitbuffer = require("ffi/blitbuffer")
         table.insert(self.menu_items.developer_options.sub_item_table, {
             text = _("Disable C blitter"),
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -108,19 +108,23 @@ local Device = Generic:new{
         end
     end,
 
-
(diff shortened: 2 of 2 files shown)


Frenzie (approved):
I'm mainly worried about the Android 4 ereaders but damn the torpedoes, full speed ahead! ;-)

NiLuJe:
Yeah, if push comes to shove, we can limit this to Android >= 6 or something as mentioned on Gitter ;).

Frenzie:
Careful, I'll put Android 4 back on my old phone just to complain. :-p

ptrm:
Not sure it's connected, but after upgrading from a15106f to gb2b9b49 Koreader fails to get activated after having started once. It show document loading info nd crashes (not sure how to find the logs). Closing the task and starting anew usually helps. That's on Android 10 on Onyx Boox Nova 3.

Reporting here due to the "And see how much stuff breaks \o/" part ;)

NiLuJe:
I did manage to hit something like that when I was chaining test builds,
but I can't actually reproduce it anymore (after a clean install, i.e.,
uninstall first).

NiLuJe:
Okay, writing that jinxed it ;p. It does start up fine on the second try,
though. Can't look at adb right now, but from what I recall, it was
something about failing to load some bundled libraries...

ptrm:
Yeah, 21af94 still does that. Repro path that works most of time is: launch koreader with book open -> switch to another task -> cloe Koreader task -> launch koreader -> crash. Will try to post a log in a separate issue if I manage to find one.

PR #7027, opened 2020-12-22
htmlToPlainText(): More newlines on HTML paragraphs

Frenzie:
Cf. <https://www.mobileread.com/forums/showthread.php?p=4072308#post4072308>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7027)

### DIFF
--- a/frontend/util.lua
+++ b/frontend/util.lua
@@ -1002,8 +1002,8 @@ This may fail on complex HTML (with styles, scripts, comments), but should be fi
 function util.htmlToPlainText(text)
     -- Replace <br> and <p> with \n
     text = text:gsub("%s*<%s*br%s*/?>%s*", "\n") -- <br> and <br/>
-    text = text:gsub("%s*<%s*p%s*>%s*", "\n") -- <p>
-    text = text:gsub("%s*</%s*p%s*>%s*", "\n") -- </p>
+    text = text:gsub("%s*<%s*p%s*>%s*", "\n\n") -- <p>
+    text = text:gsub("%s*</%s*p%s*>%s*", "\n\n") -- </p>
     text = text:gsub("%s*<%s*p%s*/>%s*", "\n") -- standalone <p/>
     -- Remove all HTML tags
     text = text:gsub("<[^>]*>", "")
--- a/frontend/util.lua
+++ b/frontend/util.lua
@@ -1002,8 +1002,8 @@ This may fail on complex HTML (with styles, scripts, comments), but should be fi
 function util.htmlToPlainText(text)
     -- Replace <br> and <p> with \n
     text = text:gsub("%s*<%s*br%s*/?>%s*", "\n") -- <br> and <br/>
-    text = text:gsub("%s*<%s*p%s*>%s*", "\n\n") -- <p>
-    text = text:gsub("%s*</%s*p%s*>%s*", "\n\n") -- </p>
+    text = text:gsub("%s*<%s*p%s*>%s*", "\n&nbsp;&nbsp;&nbsp;&nbsp;") -- <p>
+    text = text:gsub("%s*</%s*p%s*>%s*", "\n") -- </p>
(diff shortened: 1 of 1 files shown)


poire-z:
Not fond of empty lines (as html5.css) :/
Or a few `&nbsp;` at start of para (as epub.css) ?
Or as an optional parameter `condensed=false`, `altstyle=true` ?

It's also used at:
`frontend/ui/wikipedia.lua: local caption = tremain and util.htmlToPlainText(tremain)`
for the text at the top of ImageViewer. Dunno if we can get P in there - but empty lines in there would waste space.

Frenzie:
True, empty lines are a bit American style. Spaces:

Screenshot_2020-12-22_16-53-54

poire-z:
More to my liking (I'm so unamerican :)
But as most people prefer, I'll see if I'm ever bothered :)
(Looks like Wikipedia captions does not go thru this.)

NiLuJe:
Yep, I'd tend to prefer an indent, too ;).

Frenzie:
I very much prefer indents; I'm just afraid of potentially trying to be a little too clever with the plain text. :-)

Frenzie:
Alright, how about this?
Screenshot_2020-12-22_20-07-42

poire-z:
First paragraph not indented and next ones indented ? Not fond of, I like consistency :) (And I think we never see that in French text.)
(I remember you liked that too when we were talking paragraph style tweaks :)

embamba:
As a user with an interest in this issue, indent works just as well for me as vertical spacing between paragraphs does.

Frenzie:
> (And I think we never see that in French text.)

poire-z:
OK ok :) I think _I_ have never seen that in french text :)

Frenzie:
It does seem to be a standard feature of books published by nrf and other French novels I own to indent the first paragraph, regardless if it's from 1923 or from 2016.

It could be related to the fact that it's a Belgian book. Belgian books also have their backs the correct way around. ;-)

NiLuJe:
Yeah, indents everywhere in fiction ;p.

I'm seeing a *few* examples of the first paragraph being unintended, but it's generally for something specific: a blockquote or an inline title or something.

NiLuJe:
e.g., fiction:

fiction

non-fiction:

nonfic

poire-z:
Small issue - but noticable and as this is all about aesthetics :) with the 4 `&nbsp;` trick:
`&nbsp;` gets flagged with "can extend width" by xtext.cpp, just like spaces.
So, when justifying a line, they all get a few pixels more - depending on how much is needed to reach the right margin. And this "how much" can differ between lines. So, you most often won't get perfectly same-indent paragraphs :(
Using a tab (`\t`) instead of these 4 `&nbsp;` looks quite similar - and make them all indented the same.
(But the regexp stuff is a bit tough to not have another gsub with `\s` cancel the added `\t`.)

Frenzie:
> (But the regexp stuff is a bit tough to not have another gsub with \s cancel the added \t.)

What do you mean exactly?

poire-z:
https://github.com/koreader/koreader/blob/0e494169f237375990a704165d3d0909c9120287/frontend/util.lua#L1003-L1016
All the \s in there will catch the \t.
So, it's not as simple as:
`text = text:gsub("%s*<%s*p%s*>%s*", "\n\t") -- <p>`
You need either to use a placeholder so it's not caught by the \s like:
`text = text:gsub("%s*<%s*p%s*>%s*", "\n###IAMAREALTAB###") -- <p>`
or... think a bit more :)

Frenzie:
I see no reason not to use the easy placeholder solution. ;-)

poire-z:
> (But the regexp stuff is a bit tough to not have another gsub with \s cancel the added \t.)
> All the \s in there will catch the \t.

Turned out a simply reordering of the gsubs was enough to avoid this, #7095.

Issue #7025, opened 2020-12-22
Back/up icon in OPDS catalogs doesn't work as expected from top of catalog


johnbeard:
* KOReader version: nightly
* Device: Clara HD + Emulator

#### Issue

The back/up icon in the bottom left of an ODPS catalog doesn't work as expected from the top level of the catalog

#### Steps to reproduce

* Open OPDS list
* Enter, say, Project Gutenberg, then Popular
* Press back/up
* Now at the PG top level
* Press back/up again
* Should be: at the OPDS catalog
* Actually: tries to open a different catalog and fails

NiLuJe:
I, err, can't reproduce that?

I do end up back at the OPDS catalog list.

----

Can you check what happens on your end with verbose debug enabled?

johnbeard:
Hmm, actually it's not only going to the last OPDS catalog, it's trying to go to the last Dropbox folder I had open.
```
12/22/20-17:15:35 DEBUG Browse opds url https://gallica.bnf.fr/opds
12/22/20-17:15:35 INFO  request {
    ["method"] = "HEAD",
    ["sink"] = function: 0x7f09bebddf10,
    ["url"] = "https://gallica.bnf.fr/opds",
    ["headers"] = {
        ["Host"] = "gallica.bnf.fr",
    },
}
12/22/20-17:15:35 DEBUG Cache hit for opds|catalog|https://gallica.bnf.fr/opds
12/22/20-17:15:35 DEBUG setDirty via a func from widget table: 0x7f09c111eba0
12/22/20-17:15:35 DEBUG setDirty via a func from widget table: 0x7f09c111eba0
12/22/20-17:15:35 DEBUG not painting 1 covered widget(s)
12/22/20-17:15:35 DEBUG painting widget: table: 0x7f09d5de3bf8
12/22/20-17:15:35 DEBUG blitFrom 7 876 0 0 36 36
12/22/20-17:15:35 DEBUG _refresh: Enqueued ui update for region 0 0 800 914 
12/22/20-17:15:35 DEBUG _refresh: Enqueued ui update for region 0 0 800 914 
12/22/20-17:15:35 DEBUG refresh on physical rectangle 0 0 800 914
12/22/20-17:15:35 DEBUG not painting 1 covered widget(s)
12/22/20-17:15:36 DEBUG input event => type: 3, code: 57(nil), value: 0, time: 1608657336.403091
12/22/20-17:15:36 DEBUG not painting 1 covered widget(s)
[...]

PR #7028, opened 2020-12-22
Menu: Clear path history on close

NiLuJe:
At a quick glance, it's only used by CloudStorage & OPDSBrowser

Which explains the weird potential for mixups in #7025

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7028)

### DIFF
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -1244,8 +1244,21 @@ function Menu:onRight()
 end
 
 function Menu:onClose()
+    print("Menu:onClose")
     local table_length = #self.item_table_stack
+    print("table_length:", table_length)
+    logger.dbg("self.paths:", self.paths)
     if table_length == 0 then
+        -- Clear the path history (CloudStorage/OPDSBrowser)
+        for i, _ in ipairs(self.paths) do
+            self.paths[i] = nil
+        end
+
+        --[[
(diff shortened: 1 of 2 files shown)


Frenzie (approved):

poire-z:
There's potentially an issue:
`paths = {}` is created via:
```lua
local Menu = FocusManager:new{
    [...]
    paths = {} -- as a class object
}
function Menu:init()
   -- and not here, which is fine
end
```
So, as a class object, shared by all instances.
I guess probably to keep tracks of paths seen via multiple FileBrowsers instances ?

So, clearing it far away in a ODPS/CloudStorage instance, we might lose the file browser history ?
(I don't think I personally care about this long-lasting-across-Filebrowser-sessions history, but may be it has some purpose for some?)

NiLuJe:
I don't *think* anything else actually uses it, and the fact that it's shared is indeed kinda the root of the issue ^^.

poire-z:
So, may be just put `self.paths = {}` in `Menu:init()` so it's per-instance?

NiLuJe:
The Menu class does use it to decide whether the "Return" button is grayed out or not.

But, at a quick glance, there are no other actual users than OPDSBrowser & CloudStorage ;).

poire-z (approved):

PR #6934, opened 2020-12-01
ReaderDictionary: Avoid an irritating refresh when dismissing the InfoMessage

NiLuJe:
Provided said InfoMessage exists, and is not taller than the Dict/Wiki popup itself.

Re #6932

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6934)

PR #7029, opened 2020-12-22
Dictionary: various tweaks

poire-z:
Various tweaks related to dictionary lookup.

#### cre.getWordFromPosition(): fix a few issues

Drop the use of crengine's getWordFromPosition() which is a bit unreliable: it may returns wrong coordinates, or words from far away in the book (ie. when holding in the margins).
Rely only on the robust getTextFromPositions() that we already use for multi words selection.
Having good coordinates allows refreshing a smaller region (the higlighted word, or the 2 lines if hyphenated).

#### InfoMessage: allows being drawn only after a delay

- InfoMessage: add an option show_delay.
- MovableContainer: avoid possible crash when not yet painted.
- ReaderDictionary: use show_delay=0.5 instead of scheduleIn(0.5n...) (a5b133da, #7012) as we need the InfoWidget to be known to UIManager to catch tap and allow interrupting lookups properly. See https://github.com/koreader/koreader/pull/7012#issuecomment-749054293
- ReaderDictionary: rework handling of no_refresh_on_close (6162c287, #6934) with proper regions comparisons

#### ReaderDictionary: show nothing if interrupted quickly

If interrupted quickly just after lookup launch, don't display anything (this might help avoiding refreshes and the need to dismiss after accidental long-press when manipulating the device).
Might solve the original problem #6998 was attempting to solve. See https://github.com/koreader/koreader/pull/6998#issuecomment-748527179.

#### DictQuickLookup: various tweaks
Discussed all along #6998.

- Rework widget layout to avoid implicit (and wrong) sizing, using proper padding/margins and correct measurements of components.
- Adjust lookup word font size so it's not smaller than the definition.
- Tweak small window positionning: keep it centered if possible, only move it to keep the highlighted word shown when we can.
**This is a change in behaviour**: I was not really happy with the previous behaviour, which always moved the window near a word at top or bottom. Now, we don't and keep it centered - so we see more of the word context.. Only when, if centered, the selected word is hidden, and if by moving a bit we can have it shown, we do that move. I don't know if that's preferable, personally, I could be fine with having it always centered.
- Large window: adjust to screen size correctly, leaving room for footer.
- Always have a "Close" button at bottom right.  Former always displayed button "Follow link" will be added as a 3rd row in the rare case we select a word in a link.
- Replace "tap on lookup word to edit and redo the query" with an icon on the right: tap on it to edit the queries word, long-press to edit the current result word. Only close the current dict window when lookup is launched.
- Remove feature "tap on title to set current dict as the default dict for this document", as it didn't really work, and it was bad UX.

image

image

image

image

image

image

Not heavily tested, but seems to work find on my device.
For anybody wanting to test this on their device, here's a zip to unzip on a current nightly: 
[dict_tweaks1.zip](https://github.com/koreader/koreader/files/5731413/dict_tweaks1.zip)

(I'm off tomorrow and won't be able to do code/test for a week. So, I let you decide if you feel confident to merge it before I'm back, or wait till then.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7029)


NiLuJe:
I'm fine with every UX choice made here, FWIW ;).

(And, as a small window user, I'm also fine with your approach to the "move only if absolutely necessary" conundrum).

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -752,10 +752,10 @@ function ReaderHighlight:onHold(arg, ges)
             table.insert(boxes, self.selected_word.sbox)
             self.view.highlight.temp[self.hold_pos.page] = boxes
         end
-        UIManager:setDirty(self.dialog, "ui")
-        --- @todo only mark word?
         -- Unfortunately, CREngine does not return good coordinates
-        -- UIManager:setDirty(self.dialog, "partial", self.selected_word.sbox)
+        -- UIManager:setDirty(self.dialog, "ui")
+        -- But now it does:
+        UIManager:setDirty(self.dialog, "partial", self.selected_word.sbox)
```
NiLuJe, line 758: Using `partial` here means it'll make the "pages until flash" counter tick. Not sure if we want that here?

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -167,6 +156,11 @@ function DictQuickLookup:init()
             ForwardingPanRelease = { GestureRange:new{ ges = "pan_release", range = range, }, },
         }
     end
+
+    -- We no longer support setting a default dict with Tap on title.
+    -- self:changeToDefaultDict()
+    -- Now, dictionaries can be ordered (although yet not per-book), so trust the order set
```
NiLuJe, line 162: nit: `not yet` ;)

doozan:
What about adding a Maximize/Restore button to the title bar next to the close button? 

It could be helpful in the cases like your Wikipedia lookup when you want to see more of the current article in the screen or if you have the large window set as default but want to see the current word in context. It could even save the state and act as a replacement for the "Use large dictionary" preference.

Minor Nit: "Wikipedia FR" is centered in the titlebar, but "Oxford English" left justified

Is the inverted button color ( "<<", "Cancel", "Save as EPUB") just a feature of the emulator / keyboard navigation or is that some new behavior?

Frenzie:
Yes, that indicates the active button on devices that have a d-pad or keyboard.

frontend/apps/reader/modules/readerhighlight.lua
poire-z: Made it "ui".
First time testing this "ui" on my GloHD, I had it doing a faster animation white>black>grey>black>gray instead of the usual white>black>grey.
After a koreader restart, it was fine...
My GloHD is getting more and more inconsistent...

poire-z:
> What about adding a Maximize/Restore button to the title bar next to the close button?
It could be helpful in the cases like your Wikipedia lookup when you want to see more of the current article in the screen or if you have the large window set as default but want to see the current word in context.
It could even save the state and act as a replacement for the "Use large dictionary" preference.

That's an idea :) could be another button on the right of the edit icon (to not complicate the title). May be for later.
(For Wikipedia, you have these introductions in the small window. When you tap on <kbd>Wikipedia full</kbd>, it opens the full article always in a large window.)

> Minor Nit: "Wikipedia FR" is centered in the titlebar, but "Oxford English" left justified

That's like that for some years, and it was... my intention :) : 
https://github.com/koreader/koreader/blob/a15106fcc6ce5181b0faa22d01d080d5f04c5683/frontend/ui/widget/dictquicklookup.lua#L266-L279

johnbeard:
Future idea/feature: have the edit box contain a filterable list of previous entries?

Mel-kior:
The search icon is great but the placement feels a little weird. It’s not evident what you are editing, same with the number. They may be better placed next to the word ( or it’s just a matter of getting used to ).

I got 2 bugs :
- just after starting the device ( directly on the last document ) if I do a search it keep searching for ~1min before giving the result. If I cancel and redo the search it’s the same.
Only after getting the result, or if I turn page, or if I wait 1/2min that the lookup work as excepted.
- holding on the lookup window except on the definition reset the position of the windows. But if already at that default position, the widow do strange flicker.

( tested with your latest push over 2020.12-39 on kobo H2o )


Other idea :
- Holding on the title make the lookup disappear until we release. So we can see the context and we don’t have to move the window ( not very fluid/reliable on e-ink ).
- Adding proper lemmatization, so we don’t rely on the distance based or english only stemming of sdcv. e.g. in french " fîmes " -> " faire " or " chevaux " -> " cheval " don’t get looked up correctly with sdcv. They are some GPL or MIT lookup files that could be use for that. Probably overkill or better if implemented in sdcv.

poire-z:
> just after starting the device ( directly on the last document ) if I do a search it keep searching for ~1min before giving the result. If I cancel and redo the search it’s the same.
Only after getting the result, or if I turn page, or if I wait 1/2min that the lookup work as excepted.

It that something new with this PR ?! or did you get this before too ?
It feels it would be the time needed for the device to read from disk all the dict files - and then, once cached in memory, things become faster.

> holding on the lookup window except on the definition reset the position of the windows. But if already at that default position, the widow do strange flicker.

That's the behaviour of MovableContainer. Search for it here on github, I might have described why I decided about that behaviour (which might feel arbitrary, may be it is :)

Mel-kior:
>It that something new with this PR ?! or did you get this before too ?
It feels it would be the time needed for the device to read from disk all the dict files - and then, once cached in memory, things become faster.

I tested and it is the same without your PR.

>That's the behaviour of MovableContainer. Search for it here on github, I might have described why I decided about that behaviour (which might feel arbitrary, may be it is :)

Ho I didn’t notice the transparency, it’s very faint ( so not very useful ). But the lookup transparency seems lower than the menu, it’s that normal ? 


faint_transparency_lookup

poire-z:
Following up on https://github.com/koreader/koreader/pull/6998#issuecomment-750942440 here:

> Even with fast lookup I find it useful. But the delay feel a little too short for me some times. The time you react to the popup – with a slow e-ink refresh – you got to be fast to get it.

We would now have 2 timeout values:
https://github.com/koreader/koreader/blob/7f9832c0560f37e3e9226e905f75c7390df1ed44/frontend/apps/reader/modules/readerdictionary.lua#L824-L828

0.5s to show the "Searching..." popup (so, no popup is search was fast, < 0.5s).
1.5s (so, 1s after the "Seaching..." popup is shown) where dismissing would result in nothing displayed.
You might try to tweak these to find values that would suit you - that we could then discuss (I've been fine with these values for a week - and the first 0.5s seems acceptable for those who have fast results).

> Can we make that inputs get block for 0.5sec when the lookup window appear ? So that if we try to cancel when the screen refresh we don’t mess with the lookup ?

Well, we could I guess. I actually have that kind of avoidance in my own patch, I think it's just what I pasted at the bottom of https://github.com/koreader/koreader/pull/5138#issuecomment-707384339 . @NiLuJe @Frenzie thoughts ?

poire-z:
> Ho I didn’t notice the transparency, it’s very faint ( so not very useful ).

Yes, the usefullness depends on the context. Not much with the dict result. I mostly only use it with the Search buttons, so you don't go wondering why the highlighted occurence is - when it might just be hidden:

<kbd>image</kbd>

> But the lookup transparency seems lower than the menu, it’s that normal ?

Indeed. That might be new - and not expected :)

v2020.11 | current (with or without this PR):
<kbd>image</kbd>

Not sure why. Haven't looked at it, but just asking @NiLuJe if you think the recent alpha bliting stuff changes might be at play there?

Frenzie:
> Well, we could I guess. I actually have that kind of avoidance in my own patch, I think it's just what I pasted at the bottom of #5138 (comment) . @NiLuJe @Frenzie thoughts ?

I wonder if 0.5 might be a bit long (compared to 0.2) but sure.

NiLuJe:
@poire-z: The fact that only one of them is affected would probably hint at something simpler, like a different opacity level? ;o).

But the RGB32 stuff might definitely have had an effect on this, so, check on a Kobo?

poire-z:
> But the RGB32 stuff might definitely have had an effect on this, so, check on a Kobo?

Same difference in opacity on my Kobo.

> probably hint at something simpler, like a different opacity level? ;o).

Mhh, what do you call "opacity" ? :/ I logged self.alpha in MovableContainer, and in both cases, it's `= 0.7`. (I don't think we ever specify an alpha= with anything that wraps via MovableContainer, so 0.7 the default must be always used...)

NiLuJe:
Huh. Well, that's strange. I don't recall touching anything related to that in particular, but I'll try to take a look in a few days ;).

NiLuJe:
In the meantime, if you can check whether the Lua blitter reproduces the same behavior (and/or if it was already different in 2020.11) shift would be helpful ;).

poire-z:
> In the meantime, if you can check whether the Lua blitter reproduces the same behavior 

Yes, same behaviour with the Lua blitter and the C blitter.

> (and/or if it was already different in 2020.11)

I mentionned that above with a screenshot: `v2020.11 | current (with or without this PR):`

It was fine with v2020.11 (appimage, with C blitter either enabled or disabled).

poire-z:
(^ just a rebase to master, please ignore)

poire-z:
Minor tweaks:
Don't use bold for the `6 / 7` nb of results:

image

and don't show the edit button in Wikipedia full article:
image

Frenzie:
> and don't show the edit button in Wikipedia full article:

Why not? I use it (well, long press) from there sometimes? ;-)

poire-z:
> Why not? I use it (well, long press) from there sometimes? ;-)

Just because you could do it before reaching that last panel. But OK, put back :)

frontend/apps/reader/modules/readerhighlight.lua
poire-z: Turned out this is temperature dependant :/
At 21°, I get this extra animation, white>black>grey>black>gray 
At 24°, I only get the white>black>grey ...

poire-z:
Ok, merging.
I'll add the discardEvent stuff in another PR, and will tweak the timeouts and do other fixes depending on feedback.

frontend/apps/reader/modules/readerhighlight.lua
Frenzie: 24°? I mean, it's not super cold, but… :-P

Screenshot_2020-12-31_13-29-20

poire-z: That's the device temperature, from `/sys/devices/platform/imx_epdc_fb/graphics/fb0/temperature` on my Kobo.
In my hands after 20mn of reading, it can reach 30° :)

NiLuJe: Yep, colder means slower, and hotter means faster refreshes (and possibly slightly different waveform table). Until it gets too hot and it's slow as all hell again ^^.

NiLuJe:
@poire-z: Well, here's a "WTF?!" for you: the alpha issue stems from https://github.com/koreader/koreader/commit/6162c287e8186e5da04eaa1e92f292d3cd1cbb7b

I... have no idea why :D (I had to bisect this when nothing jumped out between 2020.11 and 2020.20...).

NiLuJe:
Oh. There's an extra `UIManager:show(self.dict_window)`, so we're possibly stacking two identical widgets on top of each other, which doubles the opacity.

poire-z:
> Oh. There's an extra UIManager:show(self.dict_window), 

Oh! Quite likely :) Nice find !
I guess we can just remove the 2nd `if results and results[1] then`.
Opacities are similar if we do that.
(You can let that to me, I'll have another batch of dict related stuff over the weekend, with bits touching that area.)

NiLuJe:
I vaguely remember thinking having the dismiss outside of that loop made sense (possibly for when we don't have any results?), but I might be wrong ;p.

poire-z:
Would feel a bit cleaner to just remove the 2nd `if results and results[1] then` and have `self:dismissLookupInfo()` after everything - unless you think the ordering matters ?
(I think we can't not have `results and results[1] ` , that's just a sanity check I guess.)

NiLuJe:
If we really can't hit a codepath without `results and results[1]` being true, then, yeah, that makes sense.

The ordering only matters pedantically, I guess (e.g., honoring a FIFO order). Even more so since the tweaks and the fact that we'll barely ever actually see that IM anyway ;p.

NiLuJe:
Or, well, actually, it makes sense in that we probably don't want to *see* the IM below the dict popup if we make it transparent, so, dismissing it before showing the dict popup does make sense.

poire-z:
OK, so may be just put them in the right order/chronology in the `if results and results[1] ` branch to keep all that clearer - and add a:
```lua
else  -- shouldn't happen (but clean it if it does :)
  self:dismissLookupInfo() 
end
```
Or at start:
```lua
if not results or not results[1] then
  self:dismissLookupInfo() 
  return
end
```
to gain one level of indentation back :)

NiLuJe:
That's another variant of duplication, but, sure, sounds good to me ;).

PR #7037, opened 2020-12-23
Keyboard input in InputText

johnbeard:
Basic text entry in InputText via keyboard to make using the
emulator a bit easier.

Things that don't work:

* Shift-number
* Arrow keys (they move around the OSK)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7037)


frontend/ui/widget/inputtext.lua
```diff
@@ -182,6 +182,31 @@ else
     function InputText:initEventListener() end
 end
 
+function InputText:onKeyPress(key)
+
+    -- logger.dbg(key)
+
+    if key["key"]:len() == 1 and not key["Ctrl"] and not key["Alt"] then
```
Frenzie, line 189: What does `key["key"]:len() == 1` mean exactly? Without modifiers?

frontend/ui/widget/inputtext.lua
```diff
@@ -182,6 +182,31 @@ else
...
+
+        if key["modifiers"]["Sym"] then
```
Frenzie, line 191: Hm no, apparently not. :-P

johnbeard: Means it's just "A" or "Z" or something, as opposed to "Backspace" or something

johnbeard:
Also, comma keys also don't make it to that function, so I assume something has dibs on that event somehow.

frontend/ui/widget/inputtext.lua
Frenzie: You might also want to take a look at the SDL headers/documentation. I don't know otoh if there's anything relevant in there, but you just might be trying to implement a difficult workaround for something that's a standard part of SDL.

johnbeard: The `key` table looks like this:
```
{
    ["key"] = "F",
    ["F"] = true,
    ["modifiers"] = {
        ["Alt"] = false,
        ["Ctrl"] = false,
        ["Shift"] = false,
        ["Sym"] = false,
    },
}
```
Is there somewhere I can hook into a more complete object?

Frenzie: Yes, in base (see the link I left in Gitter).

johnbeard: Right, but how to get at that from InputText?

Frenzie: Note how I moved some of the logic to front in the past because the quick touch event emulation in base was insufficient. You might have to do something similar. The reverse would be putting front logic in base, which while functional would be unsound.

johnbeard:
So I poked around a bit and I think maybe this is a better solution: add a key handler to the _virtual_ keyboard, so if there are any SDL events, they get handled as if the user pressed the softkey.
Also added a few missing SDL scancodes to the event map (like comma).
Shift+number doesn't work, not sure what to do there since exactly what should happen is up to the keyboard map, which isn't really SDL's problem. Could just chuck a "normal" US mapping in and call it good?

Frenzie:
Like I said, keycode should make more sense for that use than scancode. We're not talking about a game where you always want WASD or something. It'd be imperfect (e.g., I don't think dead keys are supported) but more or less okay for starters.

johnbeard:
The SDL keycode delivered in the `SDL_Keysym` structure doesn't distinguish between, say, `5` and `%`: it just delivers a keycode of 53 (`SDLK_3`) along with a keycode 1073742049 (`SDLK_LSHIFT`).

Frenzie:
From SDL_keycode.h
```c
 *  A special exception is the number keys at the top of the keyboard which
 *  always map to SDLK_0...SDLK_9, regardless of layout.
```

johnbeard:
That said, using keycodes instead of scancodes in the event mapper might be sensible anyway, since the scancode will be wrong for, say, AZERTY or DVORAK layouts, but I *think* the keycodes will be right.

johnbeard:
> From SDL_keycode.h
>  *  A special exception is the number keys at the top of the keyboard which
>  *  always map to SDLK_0...SDLK_9, regardless of layout.

The `;` key (next to `L` for me) always returns 59 (`SDLK_SEMICOLON`), and not 58 (`SDLK_COLON`) even if I'm holding Shift. And there is no separate keycode for `A` and `a`, it's just 97 (`SDLK_a`), either.

Frenzie:
You should be able to do it better with https://wiki.libsdl.org/SDL_StartTextInput

johnbeard:
Needs https://github.com/koreader/koreader-base/pull/1277

Frenzie:
> not quite sure if I should check in the base commit too since it's different repo?

Only after it's merged. Once that happens it's partially a matter of who gets around to it first.

Wrt to the frontend code, if I understand the docs correctly you should call the SDL_StartTextInput wrapper as part of onShowKeyboard (or whatever it's called exactly) and SDL_StopTextInput when hiding it again. On other platforms (like Ubuntu Touch and Sailfish) that'll take care of showing the native keyboard, and it'll also be a blueprint for Android.

Basically something like

Device:startTextInput()

Which on SDL goes to SDL and on Android goes to setSoftInputMode.

frontend/ui/widget/inputtext.lua
```diff
@@ -422,14 +422,55 @@ function InputText:focus()
     self._frame_textwidget.color = Blitbuffer.COLOR_BLACK
 end
 
+-- Handle real keypresses from a physical keyboard, even if the virtual keyboard
+-- is shown. Mostly likely to be in the emulator, but could be Android + BT
+-- keyboard, or a "coder's keyboard" Android input method.
+function InputText:onKeyPress(key)
+
```
Frenzie, line 429: I don't think @poire-z would like this. ^_^

frontend/device/sdl/event_map_sdl2.lua
```diff
@@ -1,39 +1,86 @@
...
+    [62] = ">",
+    [63] = "?",
+    [64] = "@",
+
+    [91] = "[",
+    [92] = "\\",
+    [93] = "]",
+    [94] = "^",
+    [95] = "_",
+    [96] = "`",
```
Frenzie, line 41: Should we keep this in? It probably doesn't really matter, but I assume this is a leftover of sorts and we don't really need _all_ the keycode definitions.

frontend/ui/widget/inputtext.lua
```diff
@@ -422,14 +422,55 @@ function InputText:focus()
...
+-- Handle text coming directly as text from the Device layer (eg. soft keyboard
+-- or via SDL's keyboard mapping).
+function InputText:onTextInput(text)
+    self:addChars(text)
+    return true
+end
+
 function InputText:onShowKeyboard(ignore_first_hold_release)
+    Device:startTextInput()
```
Frenzie, line 464: > Avoid a layer of indirection via VirtualKeyboard - there's
actually nothing specific to the VK, it could equally be used
with a PhysicalKeyboard.

Wouldn't it make more sense to duplicate it in both types of keyboard than in every single type of potential field that takes text input? I suppose all the `InputDialog`s etc. etc. goes back to InputText, but putting it in InputText seems potentially a bit "works right now." Any thoughts @poire-z @pazos @NiLuJe?

Frenzie (approved):
lgtm in principle

frontend/device/sdl/event_map_sdl2.lua
johnbeard: You might want to access keys in general for something like hotkeys where you don't have SDL text mode open?

Probably you won't see all of them, e.g. ";" and ":" would both be ";" in practice on a normal keyboard, but one has a "Shift/Sym" set and one doesn't.

Frenzie: Sure, right now we only use the alphabet for that but I suppose you mean in a hypothetical user configurable sense? Let's just keep it in then.

frontend/device/generic/device.lua
```diff
@@ -66,6 +66,10 @@ local Device = {
     canPowerOff = no,
     canAssociateFileExtensions = no,
 
+    -- Start and stop text input mode (e.g. open soft keyboard, etc)
+    startTextInput = function() return nil end,
+    stopTextInput = function() return nil end,
```
Frenzie, line 71: This would be more in line with the rest of the code.
```suggestion
    startTextInput = function() end,
    stopTextInput = function() end,
```

Frenzie: I see, I suppose you copied it from https://github.com/koreader/koreader/commit/25d5e9322c942862a0a6d88e6e7e06a4ddfed24e which snuck in the same principle.

johnbeard: Exactly right!

frontend/ui/widget/inputtext.lua
poire-z: No real/strong thought - but I don't yet have the full picture of how all this work :)
But it feels fine to have it here, in the single module that deals with input - keyboard modules do their stuff, this is just bypassing them - when they are shown.
Also, it feels self.keyboard stays and is not reinstantiated, so having it in the keyboard module might need an additional state to keep updated.

Frenzie: Alright, thanks. I'll merge it soon then.

poire-z:
Works well, even with my azerty keyboard via a windows X11 server :)

I'll look at this issue of mine when all is merged, but just mentionning it in case you see why, or if it's a bug:
On the emulator, I have
`koreader-emulator-x86_64-linux-gnu-debug/koreader/settings/event_map.lua` containing
```lua
return {
   [42] = "Back"
}
```
which allowed me to exit things and close KOReader with the Backspace key.
This does not work anymore since I bumped base, with or without this frontend PR. It's fine if I remove base's [last commit](https://github.com/koreader/koreader-base/commit/8614f997e487c440a9bb57252c20401f2aec54f3).

Frenzie:
I'd have to double check to be sure, but a quick look suggests the backspace has changed to 8 by changing from keysim to keycode. (And 42 would now be *.)

poire-z:
Indeed, and it's even updated in the code :)
It works for me if I change 42 to 8 in my `event_map.lua`.

yparitcher:
this breaks existing function keys in
https://github.com/koreader/koreader-base/blob/8614f997e487c440a9bb57252c20401f2aec54f3/ffi/SDL2_0.lua#L400-L403

will open a PR to fix

PR #7039, opened 2020-12-23
BookInfo: Saner database migration

NiLuJe:
Avoid opening two connections in //, by feeding the current connection
to saveSetting, like we were already doing for loadSettings...

It somehow worked just fine on Kobo/Emu, but blew up in fun and
interesting ways on Kindle, probably because of the Fuse proxy.

It was still pretty dumb nonetheless.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7039)


NiLuJe:
Try this at home:

```
sqlite3 /mnt/us/koreader/settings/bookinfo_cache.sqlite3 'PRAGMA user_version=0;'
```

Then watch it try to migrate *twice*, and implode on the second one ;p.

Frenzie (approved):

PR #7041, opened 2020-12-23
ReaderActivityIndicator: fix isStub()

yparitcher:
it was being overridden by the EventListener:new

caused by #7002

c.f. https://www.mobileread.com/forums/showthread.php?t=335886

@NiLuJe 
I didn't test this (don't have a KPV device) but it makes sense to me, please test.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7041)


NiLuJe:
Oh, yeah, it probably makes sense (I still suck at OOP), thanks ;).

(I don't have a KPV device either ;p).

yparitcher:
Released into the wild so our user can test. Will report back if it is the fix.

Issue #7049, opened 2020-12-25
FR: Start filebrowser if the last book doesn't exist


hius07:
To start KOReader I use very nice launcher by @yparitcher
https://github.com/yparitcher/KUAL_Booklet/releases

If the last book was deleted (not by KOReader) the program cannot be started, I need to launch it by KUAL (filemanager) or edit defaults.lua.
I propose to start filebrowser if the last book doesn't exist.

Frenzie:
> I propose to start filebrowser if the last book doesn't exist.

That's how it works already. If it's different on Kindle, presumably it's on purpose?

hius07:
Asking @NiLuJe of the purpose?

Frenzie:
I think it's more likely to be part of KUAL or however that works than anything related to KOReader at all. ^_^

NiLuJe:
I can't check the code right now, but it's likely to be launching koreader.sh without arguments, hence the last doc. KUAL has an entry for both options, which obviously doesn't work here ;).

I don't recall how I handle that on Kobo, but it's likely that I default to passing the default Home folder, hence always going to the FM.

NiLuJe:
IIRC we should be handling a deleted document fine, though, but admittedly it's not a codepath I ever test.

hius07:
Now it shows infomessage "filename does not exist" and quits.

yparitcher:
It runs `"/bin/sh", "-c", "{ /mnt/us/koreader/koreader.sh --kual --asap ; } 2>>/var/tmp/KOL.log &"` as I usualy want to go to the last document and the point was to be as quick as possible so koreader.sh w/o arguments.

I traced this bug down to the way reader.lua parses it args (it does deal with deleted documents but we chose not to use that path :smile:), PR coming soon.

PR #6280, opened 2020-06-18
Device abstraction

pazos:
Not sure if Device.home_dir should be "/mnt/us" or "/mnt/us/documents" on Kindles. The rest of strings match.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/6280)

PR #7053, opened 2020-12-25
reader.lua: open the correct item / rework args handling

yparitcher:
it was sometimes crashing when opened with no args

Fixes: #7049

The behavior is now (in order of precedence).

1. open **file** passed on command line
2. show the Quickstart if not shown before.
3. if `start_with` is `last document` try to open the last document (confirming that it exists).
4. else if a directory was passed on the command line or `start_with` is defined with a FM based action start the correct action.
5. else show usage.

The reason I didn't prioritize a directory passed on the command line over `start_with` is that many of the launchers start koreader with the directory to start from by default so `start_with` would end up doing nothing.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7053)

### DIFF
--- a/frontend/device/sdl/device.lua
+++ b/frontend/device/sdl/device.lua
@@ -54,7 +54,7 @@ local external = require("device/thirdparty"):new{
 local Device = Generic:new{
     model = "SDL",
     isSDL = yes,
-    home_dir = os.getenv("HOME"),
+    home_dir = os.getenv("XDG_DOCUMENTS_DIR") or os.getenv("HOME"),
     hasBattery = SDL.getPowerInfo(),
     hasKeyboard = yes,
     hasKeys = yes,
@@ -125,7 +125,6 @@ local Emulator = Device:new{
 local UbuntuTouch = Device:new{
     model = "UbuntuTouch",
     hasFrontlight = yes,
-    home_dir = nil,
 }
 
 function Device:init()
--- a/frontend/device/sony-prstux/device.lua
+++ b/frontend/device/sony-prstux/device.lua
@@ -15,6 +15,7 @@ local SonyPRSTUX = Generic:new{
     canReboot = yes,
     canPowerOff = yes,
     usbPluggedIn = false,
+    home_dir = nil,
 }
 
 
--- a/platform/android/llapp_main.lua
+++ b/platform/android/llapp_main.lua
@@ -75,9 +75,9 @@ C.setenv("TESSDATA_PREFIX", path.."/koreader/data", 1)
 -- create fake command-line arguments
 -- luacheck: ignore 121
 if android.isDebuggable() then
-    arg = {"-d", file or path}
+    arg = {"-d", file}
 else
-    arg = {file or path}
+    arg = {file}
 end
 
 dofile(android.dir.."/reader.lua")
--- a/platform/cervantes/koreader.sh
+++ b/platform/cervantes/koreader.sh
@@ -39,19 +39,12 @@ ko_update_check() {
     fi
 }
 
-# if no args were passed to the script, start the FM on public partition.
-if [ "$#" -eq 0 ]; then
-    args="/mnt/public"
-else
-    args="$*"
-fi
-
 # NOTE: Keep doing an initial update check, in addition to one during the restart loop, so we can pickup potential updates of this very script...
 ko_update_check
 # If an update happened, and was successful, reexec
 if [ -n "${fail}" ] && [ "${fail}" -eq 0 ]; then
     # By now, we know we're in the right directory, and our script name is pretty much set in stone, so we can forgo using $0
-    exec ./koreader.sh "${args}"
+    exec ./koreader.sh "$@"
 fi
 
 # load our own shared libraries if possible
@@ -101,7 +94,7 @@ while [ "${RETURN_VALUE}" -ge "${RESTART_KOREADER}" ]; do
     ko_update_check
 
     # run KOReader
-    ./reader.lua "${args}" >>crash.log 2>&1
+    ./reader.lua "$@" >>crash.log 2>&1
     RETURN_VALUE=$?
 
     # check if KOReader requested to enter in mass storage mode.
--- a/platform/debian/koreader.sh
+++ b/platform/debian/koreader.sh
@@ -4,14 +4,10 @@ export LC_ALL="en_US.UTF-8"
 # writable storage: ${HOME}/.config/koreader.
 export KO_MULTIUSER=1
 
-if [ -z "${1}" ]; then
-    ARGS="${HOME}"
+if [ $# -eq 1 ] && [ -e "$(pwd)/${1}" ]; then
+    ARGS="$(pwd)/${1}"
 else
-    if [ $# -eq 1 ] && [ -e "$(pwd)/${1}" ]; then
-        ARGS="$(pwd)/${1}"
-    else
-        ARGS="${*}"
-    fi
+    ARGS="${*}"
 fi
 
 # working directory of koreader
--- a/reader.lua
+++ b/reader.lua
@@ -80,12 +80,31 @@ local function showusage()
     print("If you give the name of a directory instead of a file path, a file")
     print("chooser will show up and let you select a file")
     print("")
-    print("If you don't pass any path, the last viewed document will be opened")
+    print("If you don't pass any path, the File Manager will be opened")
     print("")
     print("This software is licensed under the AGPLv3.")
     print("See http://github.com/koreader/koreader for more info.")
 end
 
+local function getPathFromURI(str)
+    local hexToChar = function(x)
+        return string.char(tonumber(x, 16))
+    end
+
+    local unescape = function(url)
+       return url:gsub("%%(%x%x)", hexToChar)
+    end
+
+    local prefix = "file://"
+    if str:sub(1, #prefix) ~= prefix then
+        return str
+    end
+    return unescape(str):sub(#prefix+1)
+end
+
+local lfs = require("libs/libkoreader-lfs")
+local file
+local directory
 local Profiler = nil
 local ARGV = arg
 local argidx = 1
@@ -109,8 +128,14 @@ while argidx <= #ARGV do
         Profiler = require("jit.p")
         Profiler.start("la")
     else
-        -- not a recognized option, should be a filename
-        argidx = argidx - 1
+        -- not a recognized option, should be a filename or directory
+        local sanitized_path = getPathFromURI(arg)
+        local mode = lfs.attributes(sanitized_path, "mode")
+        if mode == "file" then
+            file = sanitized_path
+        elseif mode == "directory" or mode == "link" then
+            directory = sanitized_path
+        end
         break
     end
 end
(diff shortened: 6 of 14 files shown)


reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
+    elseif start_with == "folder_shortcuts" then
+        local FileManagerShortcuts = require("apps/filemanager/filemanagershortcuts")
+        UIManager:nextTick(function()
+            FileManagerShortcuts:new{
+                ui = FileManager.instance,
+            }:onShowFolderShortcutsDialog()
+        end)
+    end
     exit_code = UIManager:run()
+-- nothing to do, bailout with help info.
```
Frenzie, line 322:
```suggestion
```

reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
+    exit_code = UIManager:run()
+-- if the quickstart was not shown, show it.
+else
+    local QuickStart = require("ui/quickstart")
+    if not QuickStart:isShown() then
+        start_with = "last"
+        last_file = QuickStart:getQuickStart()
+    end
+end
+-- if start_ with is last document makes sure it exists.
```
Frenzie, line 277:
```suggestion
```

reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
-    -- true, the given file or the last file is opened in the reader
-    if file and file ~= "" then
-        local ReaderUI = require("apps/reader/readerui")
-        UIManager:nextTick(function()
-            ReaderUI:showReader(file)
-        end)
-    -- we assume a directory is given in command line argument
-    -- the filemanger will show the files in that path
+end
+-- if file is given in command line argument the given file is opened in the reader.
```
Frenzie, line 262:
```suggestion
```

reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
+if file and file ~= "" then
+    local ReaderUI = require("apps/reader/readerui")
+    UIManager:nextTick(function()
+        ReaderUI:showReader(file)
+    end)
+    exit_code = UIManager:run()
+-- if the quickstart was not shown, show it.
```
Frenzie, line 269:
```suggestion
```

reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
-        open_last = true
-        last_file = QuickStart:getQuickStart()
-    end
-end
+local file
+local directory
 
 -- Start app
 local exit_code
+-- command line file argument overrides the defaults.
```
Frenzie, line 253:
```suggestion
```

reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
+elseif (directory and directory ~= "") or start_with ~= nil then
+    local FileManager = require("apps/filemanager/filemanager")
+    local home_dir =
+        G_reader_settings:readSetting("home_dir") or directory
     UIManager:nextTick(function()
-        ReaderUI:showReader(last_file)
+        FileManager:setRotationMode(true)
+        FileManager:showFiles(home_dir)
     end)
+    -- always open history on top of filemanager so closing history
```
Frenzie, line 299:
```suggestion
    -- Always open history on top of filemanager so closing history
```

reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
+    -- doesn't result in exit
```
Frenzie, line 300:
```suggestion
    -- doesn't result in exit.
```

Frenzie (changes requested):
I don't like most of those added comments because to me they're just redundantly paraphrasing what the code already clearly says. The history comment is different in that it explains why it's done that way.

pazos (approved):
As long as `home_dir` prevails over the directory passed as an argument it should as expected everywhere.

reader.lua
```diff
@@ -245,76 +245,81 @@ end
...
     exit_code = UIManager:run()
-elseif last_file then
-    local ReaderUI = require("apps/reader/readerui")
+-- or else we assume a directory is given in command line argument
+-- the filemanger will show the files in that path
+-- or the home path if start_with is defined.
+elseif (directory and directory ~= "") or start_with ~= nil then
+    local FileManager = require("apps/filemanager/filemanager")
+    local home_dir =
+        G_reader_settings:readSetting("home_dir") or directory
```
Frenzie, line 294: There's also some kind of platform default dir stuff; is that still properly functional with this?

pazos: That's a fallback to **go to home** that matches `Device.home_dir`

As is now directory is always passed as an argument on all platforms. That's why `G_reader_settings:readSetting("home_dir")` takes precedence over the directory.

It should be possible to rely on `Device.home_dir` to start the filemanager on the right path.

yparitcher: > As is now directory is always passed as an argument on all platforms. 
Besides kindle, hence the source of this bug. :smile: 

That's why `G_reader_settings:readSetting("home_dir")` takes precedence over the directory.
> 
> It should be possible to rely on `Device.home_dir` to start the filemanager on the right path.

yparitcher:
As of now

yparitcher:
as of now:

1. open **file** passed on command line
2. show the Quickstart if not shown before.
3. if `start_with` is `last document` try to open the last document (confirming that it exists). **if not fall back to FM**
4. else if a directory was passed on the command line or `start_with` is defined with a FM based action start the correct action.
5. else show usage.

usage is only shown if no argument was passed and `start_with` is not defined. Does anyone object to me changing the default `start_with` to filemanager and usage will never be shown (worst case will be dropped into FM)?

also can we start changing the launchers **not** to pass a directory and let the Home/device default be defined in KOReader, being simpler and allowing at some future point for the command line directory to actually take precedence? Is there a specific reason for the current method?

pazos:
> also can we start changing the launchers not to pass a directory and let the Home/device default be defined in KOReader (Device.home_dir), being simpler and allowing at some future point for the command line directory to actually take precedence? Is there a specific reason for the current method?

I agree. `Device.home_dir` was [added recently](https://github.com/koreader/koreader/pull/6280) to deal a bit more with device abstraction. It is not guaranteed to return something (for instance it will return nil on the emulator) so a good workaround would be

 `local directory =  G_reader_settings:readSetting("home_dir") or Device.home_dir or lfs.currentdir()`

Also agree with command line precedence. I would expect the same behaviour with both files and directories.

yparitcher:
> I agree. `Device.home_dir` was [added recently](https://github.com/koreader/koreader/pull/6280) to deal a bit more with device abstraction. It is not guaranteed to return something (for instance it will return nil on the emulator) so a good workaround would be

for sdl based platforms (including the emulator) `home_dir` is `os.getenv("HOME")`

I started fixing this, the platforms left to figure out are:
1. Ubuntu touch: `home_dir` is nil. ? Does anyone have a idea for this? it's launcher uses `~/Documents`
2. sony-prstux: cant find the launcher, and `home_dir` is not defined.
3. android: not sure where the launcher defines this to be removed. `home_dir` is correctly defined.
4. the SDL based devices.
    1. mac not sure if or how the launcher passes a path?
    2. appimage: launcher tries to find `XDG_DOCUMENTS_DIR` then `pwd`, as opposed to the `home_dir` which is `os.getenv("HOME")`.
5. kindle: kual FM button will always start with `/mnt/us/documents/` once cli directory overrides, should this be changed? maybe get rid of that button as it can now be customized in KOReader with `start_with`? @NiLuJe 


Please advise.


after at least one release with this working, I will then make directory passed on the command line override.

yparitcher:
Can I just pass `"$*"` on these platforms even if it is nothing?

NiLuJe:
>     5. kindle: kual FM button will always start with `/mnt/us/documents/` once cli directory overrides, should this be changed? maybe get rid of that button as it can now be customized in KOReader with `start_with`? @NiLuJe

Fine by me ;).

NiLuJe:
> Can I just pass `"$*"` on these platforms even if it is empty?

I assume you mean in the shell launchers on platforms where the launcher script is in shell?

In which case, I'd say, yeah, forward `"$@"` so that one can also toggle options and leave it at that?

pazos:
> 1. Ubuntu touch: `home_dir` is nil. ? Does anyone have a idea for this? it's launcher uses `~/Documents`

It is a broken platform from our POV. Leaving it to lfs.currentdir() is fine as it will work, and somebody can add a proper home while fixing the rest of issues.

> 2. sony-prstux: cant find the launcher, and `home_dir` is not defined.

The same as above. Single user platform without toolchain, launch scripts or any kind of instructions. lfs.currentdir() is fine.

> 3. android: not sure where the launcher defines this to be removed. `home_dir` is correctly defined.

Remove path from https://github.com/koreader/koreader/blob/master/platform/android/llapp_main.lua#L77-L81

> 4. the SDL based devices.
>    4.1. mac not sure if or how the launcher passes a path?
    
In https://github.com/koreader/koreader-base/blob/master/osx_loader.c#L69, it should end like https://github.com/koreader/koreader-base/blob/master/osx_loader.c#L76

>    4.2. appimage: launcher tries to find `XDG_DOCUMENTS_DIR` then `pwd`, as opposed to the `home_dir` which is `os.getenv("HOME")`.

`os.getenv("XDG_DOCUMENTS_DIR") or os.getenv("HOME")` as home_dir ?

platform/cervantes/koreader.sh
```diff
@@ -39,19 +39,12 @@ ko_update_check() {
...
-    args="$*"
-fi
-
 # NOTE: Keep doing an initial update check, in addition to one during the restart loop, so we can pickup potential updates of this very script...
 ko_update_check
 # If an update happened, and was successful, reexec
 if [ -n "${fail}" ] && [ "${fail}" -eq 0 ]; then
     # By now, we know we're in the right directory, and our script name is pretty much set in stone, so we can forgo using $0
-    exec ./koreader.sh "${args}"
+    exec ./koreader.sh "$*"
```
Frenzie, line 47: Shouldn't that be $@?

frontend/device/sony-prstux/device.lua
```diff
@@ -15,6 +15,7 @@ local SonyPRSTUX = Generic:new{
     canReboot = yes,
     canPowerOff = yes,
     usbPluggedIn = false,
+    home_dir = nil,
```
pazos, line 18: inherited from generic device, no need to override.

yparitcher: I wanted to make it explicit that this is broken.

platform/android/llapp_main.lua
```diff
@@ -75,9 +75,7 @@ C.setenv("TESSDATA_PREFIX", path.."/koreader/data", 1)
 -- create fake command-line arguments
 -- luacheck: ignore 121
 if android.isDebuggable() then
-    arg = {"-d", file or path}
-else
-    arg = {file or path}
+    arg = {"-d"}
```
pazos, line 78: Please keep (the rest of) things as they are. `file` is needed as it is how we can open files delivered by intents when the app is closed. `"-d"` should only be enabled on debug builds. It can be still overriden by "enable debug logs" on developer options.

yparitcher: whoops, forgot about the file. fixed now

NiLuJe:
I'd be inclined to hide the detail from the KUAL button, and just rename them "Start KOReader", since they will now honor `start_with`, which may not be `last` (especially since it defaults to `filemanager` ;)).

platform/android/llapp_main.lua
pazos: Thanks!. I think it should work as is, but I will test before merging. If, for whatever reason, arg doesn't like to join nil we can write it:

`args = file and {"-d", file} or {"-d"}`
`args = file and {file} or {}`

yparitcher:
Rebased on master

pazos:
@yparitcher: could you bump base to include the changes in osx launcher? Thanks

frontend/device/sdl/device.lua
```diff
@@ -123,7 +124,7 @@ local Emulator = Device:new{
 local UbuntuTouch = Device:new{
     model = "UbuntuTouch",
     hasFrontlight = yes,
-    home_dir = nil,
+    home_dir = nil, --  ~/Documents ?
```
Frenzie, line 127: Just the default for SDL should do afaik. I.e., `os.getenv("XDG_DOCUMENTS_DIR")` which will usually lead to `$HOME/Documents`.

```suggestion
```

platform/debian/koreader.sh
```diff
@@ -4,14 +4,10 @@ export LC_ALL="en_US.UTF-8"
...
-    ARGS="${HOME}"
+if [ $# -eq 1 ] && [ -e "$(pwd)/${1}" ]; then
+    ARGS="$(pwd)/${1}"
 else
-    if [ $# -eq 1 ] && [ -e "$(pwd)/${1}" ]; then
-        ARGS="$(pwd)/${1}"
-    else
-        ARGS="${*}"
-    fi
+    ARGS="${*}"
```
Frenzie, line 10:
```suggestion
    ARGS="$@"
```

$* is seldom what you want

reader.lua
```diff
@@ -242,80 +251,77 @@ local function getPathFromURI(str)
...
 local exit_code
 if ARGV[argidx] and ARGV[argidx] ~= "" then
-    local file
     local sanitized_path = getPathFromURI(ARGV[argidx])
     if lfs.attributes(sanitized_path, "mode") == "file" then
         file = sanitized_path
-    elseif open_last and last_file then
-        file = last_file
+    else
+        directory = ARGV[argidx]
```
Frenzie, line 262: Does the file manager automatically fall back to lfs.currendir() if the path is wrong? It looks a bit odd next to the file one not to have a basic check in place.

yparitcher:
> @yparitcher: could you bump base to include the changes in osx launcher? Thanks

@Frenzie did it in #7088

frontend/device/sdl/device.lua
yparitcher: the SDL default for now is `os.getenv("HOME")` only appimage is `os.getenv("XDG_DOCUMENTS_DIR")` should i have UT be that or change the sdl default?

reader.lua
yparitcher: https://github.com/koreader/koreader/blob/d560f9011d3b4f8bcd2f205019082b6b6c689f1b/frontend/apps/filemanager/filemanager.lua#L1104-L1105

https://github.com/koreader/koreader/blob/d560f9011d3b4f8bcd2f205019082b6b6c689f1b/frontend/apps/filemanager/filemanagerutil.lua#L12-L14

Frenzie: Ok, functionally sound, looks slightly odd reading.

frontend/device/sdl/device.lua
Frenzie: I see no reason to make a distinction, so I'd change the default.

yparitcher: done

Frenzie (approved):
lgtm

yparitcher:
Reworked this a bit to make sure thing were not being triggered twice (reopening on exit), and to simplify a bit.
I decided not to wait and now Directories passed on the cli are given preference.
I don't know how this will affect the tests, so please wait to merge until #7907 so i can test this.

reader.lua
```diff
@@ -206,87 +231,68 @@ if G_reader_settings:isTrue("color_rendering") and not Device:hasColorScreen() t
...
+if file then
+    local ReaderUI = require("apps/reader/readerui")
+    UIManager:nextTick(function()
+        ReaderUI:showReader(file)
+    end)
+    exit_code = UIManager:run()
+elseif directory then
+    local FileManager = require("apps/filemanager/filemanager")
+    UIManager:nextTick(function()
+        FileManager:setRotationMode(true)
```
NiLuJe, line 265: This one becomes redundant after #7096, right?

reader.lua
```diff
@@ -206,87 +231,68 @@ if G_reader_settings:isTrue("color_rendering") and not Device:hasColorScreen() t
...
-    -- we assume a directory is given in command line argument
-    -- the filemanger will show the files in that path
+        exit_code = UIManager:run()
     else
         local FileManager = require("apps/filemanager/filemanager")
         local home_dir =
-            G_reader_settings:readSetting("home_dir") or ARGV[argidx]
+            G_reader_settings:readSetting("home_dir") or Device.home_dir or lfs.currentdir()
         UIManager:nextTick(function()
             FileManager:setRotationMode(true)
```
NiLuJe, line 291: Ditto

reader.lua
yparitcher: It will be, I will take care of it there after this is merged, I just want to keep the changes separate

pazos:
@yparitcher: circleci cries with

```
In platform/debian/koreader.sh line 10:
    ARGS="$@"
         ^--^ SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.

For more information:
  https://www.shellcheck.net/wiki/SC2124 -- Assigning an array to a string! A...
```

yparitcher:
> @yparitcher: circleci cries with
> 
> ```
> In platform/debian/koreader.sh line 10:
>     ARGS="$@"
>          ^--^ SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.
> 
> For more information:
>   https://www.shellcheck.net/wiki/SC2124 -- Assigning an array to a string! A...
> ```

Thanks, as I don't want to give CircleCi write permissions to my repos I can't check the failures by myself

yparitcher:
Didn't work. :( not sure why.

pazos:
> Didn't work. :( not sure why.

```
Checking reader.lua                               1 warning

    reader.lua:278:9: value assigned to variable 'exit_code' is unused

Total: 1 warning / 0 errors in 449 files

Exited with code exit status 1
CircleCI received exit code 1
```

> Thanks, as I don't want to give CircleCi write permissions to my repos I can't check the failures by myself

I think you can log as KOReader instead of your personal account.

Frenzie:
> circleci cries with

I said seldom. The point is to double check what kind of case this is. ;-)

> Thanks, as I don't want to give CircleCi write permissions to my repos I can't check the failures by myself

I could add a line to the Makefile and/or kodev to run those particular tests too. That being said, I tend to run only the ones I want locally. (In this case basically luacheck and shellcheck.)

yparitcher:
> I think you can log as KOReader instead of your personal account.

I only have one github account.

sorry about the luacheck, usually i run `./kodev check` locally. which does luacheck but not shellcheck (I usualy stick to editing lua anyways)

reader.lua
```diff
@@ -206,87 +231,69 @@ if G_reader_settings:isTrue("color_rendering") and not Device:hasColorScreen() t
...
-local exit_code
-if ARGV[argidx] and ARGV[argidx] ~= "" then
-    local file
-    local sanitized_path = getPathFromURI(ARGV[argidx])
-    if lfs.attributes(sanitized_path, "mode") == "file" then
-        file = sanitized_path
-    elseif open_last and last_file then
-        file = last_file
+    if start_with == "last" and last_file and lfs.attributes(last_file, "mode") ~= "file" then
+        UIManager:show(retryLastFile())
```
NiLuJe, line 277: Does this one actually show up, or does it require the same `nextTick` trickery as the others (c.f., https://github.com/koreader/koreader/pull/1877)

yparitcher: It shows, `nextTick` would break it.

yparitcher:
rebase on master

NiLuJe (approved):

Issue #6056, opened 2020-04-15
Dictionary: Support images?


mergen3107:
* KOReader version: 2020.03.02
* Device: Kindle PW3 (5.9.7)

Can I hope about showing pictures in dictionary queries?
They show up in the `=h` output as `<blockquote><blockquote><rref>spire.jpg</rref></blockquote></blockquote>`
For example: 
![Reader_2020-Apr-11_183730](https://user-images.githubusercontent.com/6279855/79299441-0c11e800-7eb2-11ea-9352-a30231469071.png)

I am complaining, just wanted to give you my feedback on how to make it better.
If it is not possible or too hard for the purpose, then I can just hide these references as `.wav` files (#6011)

Frenzie:
Something like this should do the trick. You might have to change it to `<img src="/path/to/dictionary/img/files/%1">`
```lua
return function(html)
html = html:gsub('<rref[^>]*>([^<]*%.jpg)</rref>', '<img src="%1">')
return html
end
```

imkaihao:
@Frenzie 
I do something like this but it still doesn't work.
```lua
return function(html)
    html = html:gsub('<rref[^>]*>([^<]*%.jpg)</rref>', '<img src="/path/to/dictionary/img/files/%1" />')
    return html
end
```
KOReader version: 2020.12
I have tested on Kobo Forma and desktop Ubuntu.
It just shows like this:
![FileManager](https://user-images.githubusercontent.com/35647230/103145272-ae3cdc00-4772-11eb-8010-2e4e66c23dd8.png)

Frenzie:
@imkaihao That's what MuPDF puts in there when it can't find the referenced image file. Which is a virtual certainty if you kept `/path/to/dictionary/img/files`! ;-)

imkaihao:
@Frenzie I mean the real path to images. In Kobo Forma it's like this
```lua
return function(html)
    html = html:gsub('<rref[^>]*>([^<]*%.jpg)</rref>', '<img src="/mnt/onboard/.adds/koreader/data/dict/stardict-Longman_Dictionary_of_Contemporary_English_5th_Ed-2.4.2/res/%1" />')
    return html
end
```
It doesn't work. Just showing the result like the above screenshot.

Frenzie:
If you use `data/dict/etc.` it should work.

imkaihao:
@Frenzie 
[...]

PR #7057, opened 2020-12-26
Pass dict_path to dictionary fix_html_func

Frenzie:
That way images can be made to work in a more generic, portable manner, such as:

```lua
return function(html, dict_path)
    html = html:gsub('<rref[^>]*>([^<]*%.jpg)</rref>', '<img src="'..dict_path..'res/%1">')
    return html
end
```

Cf. <https://github.com/koreader/koreader/issues/6056>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7057)

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -577,7 +577,8 @@ local function tidyMarkup(results)
             result.is_html = ifo.is_html
             result.css = ifo.css
             if ifo.fix_html_func then
-                local ok, fixed_definition = pcall(ifo.fix_html_func, result.definition)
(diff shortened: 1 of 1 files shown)


poire-z (approved):
What kind of images are these ? Curious to see some screenshots :)

Frenzie:
The CIA Factbook is an example of a resource that contains images, although that's actually a slightly different syntax. (It just uses a regular `` but without putting `res/` in front of the filename. I'm not sure if that's supposed to Just Work or if it's a peculiarity of the conversion from regular HTML files.)

Pictures in dictionaries look like this. In electronic dictionaries the same ones are applied.


French Wiktionary also contains many pictures like that: https://fr.wiktionary.org/wiki/%C5%93il

PR #7065, opened 2020-12-27
Fix reMarkable 2 "double-press power button required to wake up"

snelg:
While reMarkable 2 is suspended, pressing the power button wakes up the device but does NOT send an input event. So KOReader does not know the device has exited its suspend mode. Currently, the user has to press the power button twice: once to physically wake the device, and once to send the input event to KOReader.

This PR schedules a manual `UIManager:resume` call, which will trigger after the device wakes up from suspension.

Notes:

 - I don't know if the reMarkable 1 has the same behavior quirk. If so, then it's simple enough to apply this same fix to both device types.
 - I don't know if there's a cleaner way to guarantee a post-wakeup call.
 - A zero-second wait in `scheduleIn` does *not* work; it triggers the `resume` call *before* the device actually suspends, i.e. before the `systemctl suspend` process finishes suspending the device. When that happens, things start going wonky if the device is physically suspended for longer than the `autosuspend` plugin's timeout.
 - I tried various non-zero values in `scheduleIn`. Values as low as 0.1 worked, so I left a little buffer and set it to 0.5.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7065)

### DIFF
--- a/frontend/device/remarkable/device.lua
+++ b/frontend/device/remarkable/device.lua
@@ -141,8 +141,20 @@ function Remarkable:setDateTime(year, month, day, hour, min, sec)
     return os.execute(command) == 0
 end
 
-function Remarkable:suspend()
+function Remarkable1:suspend()
     os.execute("systemctl suspend")
+end
+
+function Remarkable2:suspend()
+    os.execute("systemctl suspend")
+    -- While device is suspended, when the user presses the power button and wakes up the device,
+    -- a "Power" event is NOT sent.
+    -- So we schedule a manual `UIManager:resume` call just far enough in the future that it won't
+    -- trigger before the `systemctl suspend` command finishes suspending the device
+    local UIManager = require("ui/uimanager")
+    UIManager:scheduleIn(0.5, function()
+        UIManager:resume()
+    end)
 end
 
 function Remarkable:resume()


NiLuJe:
On the subject of "zero seconds", we have nextTick and tickAfterNext, which basically do what they say on the tin. Since there shouldn't be any UI scheduling, nextTick might be enough to trip at the next input event.

snelg:
I did try `nextTick` (which I saw is a wrapper for `scheduleIn(0, [...]) `), but the scheduled `resume` call was definitely starting *before* the device physically suspends. I just re-tried it, including some logger calls like I did originally:

```lua
os.execute("systemctl suspend")
logger.info("SCHEDULING resume")
UIManager:nextTick(function()
    logger.info("RUNNING resume")
    UIManager:resume()
    logger.info("RAN resume")
end)
```

Results with `nextTick`, pressing physical Power button, waiting about 30 seconds, and pressing again:
```
12/27/20-11:31:49 INFO  SCHEDULING resume
12/27/20-11:31:49 INFO  RUNNING resume
12/27/20-11:31:49 INFO  RAN resume
```

With `tickAfterNext`:
```
12/27/20-11:34:18 INFO  SCHEDULING resume
12/27/20-11:34:18 INFO  RUNNING resume
12/27/20-11:34:18 INFO  RAN resume
```

With `scheduleIn(0.5`:
```
12/27/20-11:35:55 INFO  SCHEDULING resume
12/27/20-11:36:29 INFO  RUNNING resume
12/27/20-11:36:29 INFO  RAN resume
```

The `nextTick` version (or even skipping a scheduling wrapper altogether and immediately calling `UIManager:resume`) tricked me into thinking it was running fine, UNTIL I left the device suspended for more than 20 minutes. At that point, when I woke it up, the screen flickered on for a fraction of a second and then went *back* to suspended mode. Took me a while to figure out that this was what was happening:
 - `systemctl suspend` is called
 - `UIManager:resume()` is called
 - Device physically suspends
 - [20+ minutes pass]
 - I press Power button
 - Device physically wakes up
 - All of the tasks of the earlier `resume` call kick in, making everything wake up
 - *Autosuspend* hasn't seen any activity and sees that the clock is past its 20-minute window, so it initiates a suspend

That's why I need to use `scheduleIn` to wait long enough that the device has actually, physically suspended before initiating the `UIManager:resume` call: to make sure that `autosuspend` knows there's been recent activity and it should reset its timer.

pazos (approved):

wordpr3ss:
>     * I tried various non-zero values in `scheduleIn`. Values as low as 0.1 worked, so I left a little buffer and set it to 0.5.
Possible to make this value customizable?
I had to increase to 1.5 to make it work reliable for my rm2. btw, I had [this fix](https://github.com/toltec-dev/toltec/pull/183) installed, which was delaying the suspend, I believe.

PR #7094, opened 2020-12-31
LuaJIT on Android fixes

NiLuJe:
Requires https://github.com/koreader/android-luajit-launcher/pull/283 & https://github.com/koreader/koreader-base/pull/1279

Make it a real boy, i.e.:
* No longer flag it as `should_restrict_JIT`
* Allow disabling the C blitter

Tweak kodev to add a -d, --debug flag to log to catch the extra logging added in debug builds for LuaJIT.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7094)

### DIFF
--- a/kodev
+++ b/kodev
@@ -1073,7 +1073,11 @@ OPTIONS:
 
 function kodev-log() {
     LOG_HELP_MSG="
-usage: log <TARGET>
+usage: log <OPTIONS> <TARGET>
+
+OPTIONS:
+
+    -d, --debug               more verbose logs (e.g., debug builds)
 
 TARGET:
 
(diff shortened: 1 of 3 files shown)


kodev
```diff
@@ -1084,13 +1088,53 @@ TARGET:
...
-            echo "${LOG_HELP_MSG}"
-            exit 0
-            ;;
         android)
-            adb logcat 'KOReader:I ActivityManager:* AndroidRuntime:* DEBUG:* *:F'
+            if [ -n "${KODEBUG}" ]; then
+                adb logcat 'KOReader:* ActivityManager:* AndroidRuntime:* DEBUG:* *:F dlopen:* LuaJIT:*'
+            else
+                adb logcat 'KOReader:I ActivityManager:* AndroidRuntime:* DEBUG:* *:F'
+            fi
```
Frenzie, line 1137: You sure the distinction's useful?

Frenzie (approved):

kodev
NiLuJe: KOReader's own debug output is already fairly verbose, and so is dlopen, so, we probably don't want to bother unsuspecting users with it ;).

Frenzie: Unsuspecting users are running `kodev run android --no-debug`? :-P

NiLuJe: That's the other way around, the default is the less verbose version ^^.

Frenzie: Oh, right. The impact will be extremely minimal regardless.

Issue #7093, opened 2020-12-31
90° rotation is lost by restarting Koreader


WS64:
* KOReader version: 2020-12
* Device: Kobo Nia

#### Issue
When I restart Koreader the 90° rotation inside books is not there anymore, books (at least the first book) will start in portrait mode instead of landscape mode

#### Steps to reproduce
Change rotation to 90° (and make it default)
Not sure if it is important, but I also have dual pages on (default).
I can't find right now where the setting is, but I also have set up Koreader to restart with opening the last document.
I also have "Screen/Rotation/Keep current rotation across views" checked.
When I now go to th menu "Exit/Restart Koreader" Koreader will open my last document in potrait mode instead of landscape mode.
By unchecking the "keep rotation" mark it sometimes (but not always) opens in the orrect rotation but nearly all popus are turned. Since I was fighting with long boot times I am not 100% sure anymore what exactly happens here, but I think the 90° rotation should stay even with the "keep rotation" checked.

pazos:
Each document has its own setting that can override defaults.

To make sure you're reporting a bug please do the following:

1. Change rotation to 90° (and make it default)
2. Check Screen/Rotation/Keep current rotation across views
3. Open a new book that wasn't opened before.

Books opened before have their own settings. The same applies for font size, margins and any other rule. When you setup a new rule as a default it will be the default on newer documents but won't override the settings of each particular document.

WS64:
But the book that was reopened by restarting was set to 90° and did not open with this setting.

pazos:
You reported a bug and I asked if you can reproduce on book that was never opened. Other observations about the bug could help but don't answer the question: can you reproduce on a new book **once** you setup rotation on the FM as you wish?

NiLuJe:
> I also have "Screen/Rotation/Keep current rotation across views" checked.

As I've mentioned before, there's no bug, it's doing what it's supposed to ;).

WS64:
@pazos, it is only wrong for the first book. all new (and old books) are rotated.

@Nijule, I want "if you want to use everything in Landscape, keep it enabled, but set the default FM rotation to the same Landscape, too." 
That is exactly what I do. 
I have to admit, I did not know that FM has its own defaults.
I went there, set the FM also to 90° (long press, I can see the default star).
But it does not work.
I have set both FM and the books default to 90°. And "keep rotation" checked.
Then "restart koreader", and it opens with my book not turned.

NiLuJe:
[...]

PR #7096, opened 2020-12-31
 save rotation across restarts.

yparitcher:
Fixes: #7093 

~~Based on #7053 as they change the same file.~~

~~only the last commit is related.~~

we now do not have to reset the FM rotation at start when rotation is locked.

This only affects locked rotation (never change rotation), we will ignore defaults and use the way we left off last time.
If rotation is not locked the FM or book will override.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7096)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -70,13 +70,10 @@ function FileManager:onSetRotationMode(rotation)
     return true
 end
 
--- init should be set to True when starting the FM for the first time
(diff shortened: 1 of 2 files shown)


NiLuJe:
Oh, great, thanks ;).

yparitcher:
conflicts with #7053 will fixup after it is merged.

yparitcher:
Conflicts fixed, ready for review.

NiLuJe (approved):

PR #7047, opened 2020-12-24
Add pages remaining option to status bar

JasonInOttawa:
A trivial implementation of adding pages remaining to the status bar, for those times when you want to know how many pages remain and aren't up to doing the subtraction.

This is added to the Status bar settings:

settings

This is how things look in the status bar

status bar

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7047)

PR #7097, opened 2020-12-31
Fix opacity of the dictionary window

NiLuJe:
Brainfart from 6162c287e8186e5da04eaa1e92f292d3cd1cbb7b

Forgot to zap the original UIManager:show(), so we were stacking two
identical widgets on top of each other, which obviously affected alpha.

Re #7029

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7097)


NiLuJe:
Includes a free fix for #7047 that broke the readerfooter test ;).

(I guess now we know why new values should have been appended at the end of the list ;p).

Frenzie:
> Includes a free fix for #7047 that broke the readerfooter test ;).

The proposed solution is to share secrets… yeah, not happening. We might need to switch to GitHub Actions or something if I can't figure out a reasonable way around this, because it doesn't suffer from this.
https://discuss.circleci.com/t/how-do-i-know-why-a-build-is-unauthorized/36538

Screenshot_2021-01-01_09-48-23

PR #7099, opened 2021-01-01
More dictionary tweaks

### LINKED ISSUE #6877: dash prefix in dictionary lookup is ignored
I have [a Greek dictionary Papyros ](https://e.pcloud.link/publink/show?code=XZhhJ7ZKXO2fHo4W04N0DPj1BQtS71gTU17), that has entries starting with a dash "-", e.g. -άρας. If I do a lookup in Koreader, it changes the loopup entry to άρας and doesn't find the entry. _Where in the code do I have to look to fix this?_
Screenshot from 2020-11-14 11-54-43

Goldendict does:
Screenshot from 2020-11-14 11-55-56

As does the command line sdcv of Koreader:
```bash
$ sdcv
Enter word or phrase: -άρας
Found 1 items, similar to -άρας.
-->Πάπυρος (El-El)
-->-άρας

<div style="margin-left:1em">μεγεθυντική κατάλ. αρσ. ονομάτων της Νέας Ελληνικής, που συνδέεται με την κατάλ. -<i>άρα</i>* <span style="color:red">&lt;</span> -<i>άρι</i>- (<i class="p" style="color:green">πρβλ.</i> <i>ποδάρι</i> - <i>ποδάρας</i>). Κατά κύριο λόγο η κατάλ. -<i>άρας</i> χρησιμοποιήθηκε για τον σχηματισμό ανδρικών μεγεθυντικών κυρίων ονομάτων, πολλά από τα οποία πήραν τη θέση επωνύμων χάνοντας συγχρόνως τη μεγεθυντική τους σημασία (<i class="p" style="color:green">πρβλ.</i> <i>Άγγελος</i> - <i>Αγγελάρας</i>, <i>Βασίλης</i> - <i>Βασιλάρας</i>, <i>Γιώργης</i> - <i>Γιωργάρας</i>, <i>Διαμαντής</i> - <i>Διαμαντάρας</i>, <i>Κωσταντής</i> - <i>Κωσταντάρας</i>, <i>Μήτσος</i> - <i>Μητσάρας</i>, <i>Νικόλας</i> - <i>Νικολάρας</i> κ.ά. Τέλος, από μεγεθυντικά προσηγορικά σε -<i>άρα</i> προέκυψαν επίσης επώνυμα ανδρών σε -<i>άρας</i>, όπως <i>Καλαμάρας</i>, <i>Καρδάρας</i>, <i>Πουρνάρας</i>, <i>Στουρνάρας</i>, <i>Χαντζάρας</i> κ.λπ.</div>
```

poire-z:
More dictionary tweaks, following #7029: 

#### Dict/Wiki lookup: less text cleanup on manual input
Don't cleanup input text as much when entered manually (or when it's sane) than when coming from book text selection.
This may allow looking up words like "-suffix", or do more precise Wikipedia queries.
I haven't have much need for this, but it was thought of as a possible solution to https://github.com/koreader/koreader/issues/6877#issuecomment-727192483 . Possible caveats: people who cut & paste crappy input may need to clean that input manually in the InputText widget.
Closes #6877.

#### DictQuickLookup: hold prev/next buttons go to first/last result
Because sometimes I hit <kbd>></kbd> a few times to see what kind of results I got, and then want to quickly get back to the first result to start reading them.

#### Dict, Trapper: prevent dismissal by past events

Add `UIManager:discardEvents()`, to allow dropping events for a period of time.
Default duration of 600ms on eInk, 300ms otherwise.
Used when showing the dict result window, so that a tap happening just when it's being shown won't discard it. See https://github.com/koreader/koreader/pull/7029#issuecomment-752732528
Used with Trapper:confirm()/:info(), to avoid taps made in a previous processing to dismiss (and so, select the cancel action) a ConfirmBox not yet painted/visible (I'm not sure I didn't add too many, for the info() and the confirmbox happening when interrupting info()s - we'll see.)
(Mentionned previously at https://github.com/koreader/koreader/pull/5138#issuecomment-707384339, been using 500ms with all ConfirmBox and 1000ms with dict for years.)

This is mostly useful for these 2 cases - all other cases of bounces/accidental dismissal may be taken care of by the Tap interval setting from #6798.
And it shouldn't cause much discussion (I hope :) as when a dict lookup result is shown, nobody would ever go at wanting to dismiss it - or find another button to tap - in less thans 1s :)
And for Trapper, it's mostly only used in the Wikipedia Save as EPUB and FM plus button to refresh cached book info for a whole directory, where you'd have to answer a few questions at start.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7099)

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -214,7 +214,8 @@ function ReaderDictionary:addToMainMenu(menu_items)
                     os.date("%Y-%m-%d %H:%M:%S", value.time),
                     value.word,
                     callback = function()
-                        self:onLookupWord(value.word)
+                        -- Word had been cleaned before being added to history
+                        self:onLookupWord(value.word, true)
                     end
                 })
             end
@@ -385,10 +386,10 @@ function ReaderDictionary:addToMainMenu(menu_items)
     end
 end
 
-function ReaderDictionary:onLookupWord(word, box, highlight, link)
+function ReaderDictionary:onLookupWord(word, is_sane, box, highlight, link)
     logger.dbg("dict lookup word:", word, box)
     -- escape quotes and other funny characters in word
-    word = self:cleanSelection(word)
+    word = self:cleanSelection(word, is_sane)
     logger.dbg("dict stripped word:", word)
 
     self.highlight = highlight
@@ -609,7 +610,7 @@ local function tidyMarkup(results)
     return results
 end
 
-function ReaderDictionary:cleanSelection(text)
+function ReaderDictionary:cleanSelection(text, is_sane)
     -- Will be used by ReaderWikipedia too
     if not text then
         return ""
@@ -618,31 +619,33 @@ function ReaderDictionary:cleanSelection(text)
     -- some cleanup is still needed for selection we get from other engines
     -- (example: pdf selection "qu’autrefois," will be cleaned to "autrefois")
     --
+    -- Replace no-break space with regular space
+    text = text:gsub("\xC2\xA0", ' ') -- U+00A0 no-break space
     -- Trim any space at start or end
     text = text:gsub("^%s+", "")
     text = text:gsub("%s+$", "")
-    -- Replace extended quote (included in the general puncturation range)
-    -- with plain ascii quote (for french words like "aujourd’hui")
-    text = text:gsub("\xE2\x80\x99", "'") -- U+2019 (right single quotation mark)
-    -- Strip punctuation characters around selection
-    text = util.stripPunctuation(text)
-    -- Strip some common english grammatical construct
-    text = text:gsub("'s$", '') -- english possessive
-    -- Strip some common french grammatical constructs
-    text = text:gsub("^[LSDMNTlsdmnt]'", '') -- french l' s' t'...
-    text = text:gsub("^[Qq][Uu]'", '') -- french qu'
-    -- Replace no-break space with regular space
-    text = text:gsub("\xC2\xA0", ' ') -- U+00A0 no-break space
-    -- There may be a need to remove some (all?) diacritical marks
-    -- https://en.wikipedia.org/wiki/Combining_character#Unicode_ranges
-    -- see discussion at https://github.com/koreader/koreader/issues/1649
-    -- Commented for now, will have to be checked by people who read
-    -- languages and texts that use them.
-    -- text = text:gsub("\204[\128-\191]", '') -- U+0300 to U+033F
-    -- text = text:gsub("\205[\128-\175]", '') -- U+0340 to U+036F
-    -- Trim any space now at start or end after above changes
-    text = text:gsub("^%s+", "")
-    text = text:gsub("%s+$", "")
+    if not is_sane then
+        -- Replace extended quote (included in the general puncturation range)
+        -- with plain ascii quote (for french words like "aujourd’hui")
+        text = text:gsub("\xE2\x80\x99", "'") -- U+2019 (right single quotation mark)
+        -- Strip punctuation characters around selection
+        text = util.stripPunctuation(text)
+        -- Strip some common english grammatical construct
+        text = text:gsub("'s$", '') -- english possessive
+        -- Strip some common french grammatical constructs
+        text = text:gsub("^[LSDMNTlsdmnt]'", '') -- french l' s' t'...
+        text = text:gsub("^[Qq][Uu]'", '') -- french qu'
+        -- There may be a need to remove some (all?) diacritical marks
+        -- https://en.wikipedia.org/wiki/Combining_character#Unicode_ranges
+        -- see discussion at https://github.com/koreader/koreader/issues/1649
+        -- Commented for now, will have to be checked by people who read
+        -- languages and texts that use them.
+        -- text = text:gsub("\204[\128-\191]", '') -- U+0300 to U+033F
+        -- text = text:gsub("\205[\128-\175]", '') -- U+0340 to U+036F
+        -- Trim any space now at start or end after above changes
+        text = text:gsub("^%s+", "")
+        text = text:gsub("%s+$", "")
+    end
     return text
 end
 
@@ -680,7 +683,8 @@ function ReaderDictionary:onShowDictionaryLookup()
                     is_enter_default = true,
                     callback = function()
                         UIManager:close(self.dictionary_lookup_dialog)
-                        self:onLookupWord(self.dictionary_lookup_dialog:getInputText())
+                        -- Trust that input text does not need any cleaning (allows querying for "-suffix")
+                        self:onLookupWord(self.dictionary_lookup_dialog:getInputText(), true)
                     end,
                 },
             }
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -889,6 +889,12 @@ function ReaderDictionary:showDict(word, results, box, link)
     self:dismissLookupInfo()
     if results and results[1] then
         UIManager:show(self.dict_window)
+        if not results[1].lookup_cancelled then
+            -- Discard queued and coming up events to avoid accidental
+            -- dismissal (but not if lookup interrupted, so 2 quick
+            -- taps can discard everything)
+            UIManager:discardEvents(true)
+        end
     end
 end
 
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -937,14 +937,14 @@ function ReaderHighlight:lookup(selected_word, selected_link)
     -- if we extracted text directly
     if selected_word.word then
         local word_box = self.view:pageToScreenTransform(self.hold_pos.page, selected_word.sbox)
-        self.ui:handleEvent(Event:new("LookupWord", selected_word.word, word_box, self, selected_link))
+        self.ui:handleEvent(Event:new("LookupWord", selected_word.word, false, word_box, self, selected_link))
     -- or we will do OCR
     elseif selected_word.sbox and self.hold_pos then
         local word = self.ui.document:getOCRWord(self.hold_pos.page, selected_word)
         logger.dbg("OCRed word:", word)
         if word and word ~= "" then
             local word_box = self.view:pageToScreenTransform(self.hold_pos.page, selected_word.sbox)
-            self.ui:handleEvent(Event:new("LookupWord", word, word_box, self, selected_link))
+            self.ui:handleEvent(Event:new("LookupWord", word, false, word_box, self, selected_link))
         else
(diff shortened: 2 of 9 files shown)


Mel-kior:
So, I did tested the " fast cancel " of dict search.

When wrong search it’s perfect.
But for the accidental case, if I saw it only when the IM pop up 1.5s is too fast, 2s is ok, 2.5s is good ( I should have more real case to be sure ).

But with sdcv being fast now I often got the result while/before canceling
I personally think more logic that even if the lookup is show, if still in the 1.5sec fast cancel timer, a tap should dismiss the lookup and not just getting blocked by discardEvent().

And the discardEvents() is inconsistent on my H2o. On default 600ms it rarely work, I have to set it to 1sec to get it working most of the times but I still get some unblock event ( like after certain wake up from sleep ).

poire-z:
Can't/don't want to think now again much about it :) but it felt that indeed, these 3 timers/durations could be confusing.
What about removing the fast cancel 1.5s one - and go with something like: if you want to interrupt/dismiss it/show nothing: you need to tap _twice_, with at least a discardEvent duration (600ms) between your 2 taps ?
The first tap will show the "lookup cancelled" box, or will be discarded if already shown - the second will close the window. Ok, some screen refreshes that you don't like, but sounds like it might be more consistent.

poire-z:
> And the discardEvents() is inconsistent on my H2o. On default 600ms it rarely work, I have to set it to 1sec to get it working most of the times but I still get some unblock event ( like after certain wake up from sleep ).

I don't really understand that part. What do you mean with working or not working? Can you elaborate?

For your other comments, I guess this might be what you want, and I feel I would be fine with that:
- lookup launched, keep start_timestamp, show infomessage after 0.5s : unchanged
- have a quick_dismiss_everything=3s or 5s (instead of the 1.5s)
- if dismiss before any result: just show nothing, clear the infomessage if any (as before)
- if result before quick_dismiss_everything duration, show it, don't set any discardEvent
- if result after quick_dismiss_everything duration, show it and set discardEvent

So, it's only after these 3s or 5s that the discardEvent will be ensured: assuming that it's only while you're waiting for a result for a long time that you might tap to interrupt it - but you'd be sad if you were interrupting it when it was just about to be shown after 10s, as you then would have to relaunch the lookup and wait 10s, knowing that you will get some results.
Assuming that if you tap before these 3s or 5s, it's not that you were impatient for a result - it's just that you changed your mind or are cancelling an unwilling lookup.
Thoughs ?

poire-z:
@Mel-kior : I added a 4th commit in #7114 - which may solve your issues - and that I find fine too for my usage. See https://github.com/koreader/koreader/pull/7114#issuecomment-756253006.

Mel-kior:
>if result before quick_dismiss_everything duration, show it, don't set any discardEvent

That’s not good because discardEvent is most wanted before quick_dismiss_everything duration when most result pops up and cancel are done. In that sense, the way it was before #7114 was better.

What I was thinking is when showing the result :
- if after the quick_dismiss_everything duration, set discardEvent
- if before quick_dismiss_everything duration, use a trapwidget that :
  - close the dict window if catching a tap
  - get remove at quick_dismiss_everything timer end
 
That way every tap in the quick_dismiss_everything timer close everything.
If doing this, the duration should not be too long so we don’t have unwanted close on legitimate dict window interaction. 2.5/3sec seems a good compromise.
Or 5sec quick_dismiss_everything and set the trapwidget timeout with a maximum 1.5 or 2 seconds. 

-----------

**About discardEvents() :**
I am getting 2 cases :
1) searching > result found > tap > result show ( display refresh ) > tap discarded
2) searching > result found > tap > result show ( display refresh ) > interaction with dict window ( tap not discarded )

with discardEvents(0.6) I get nearly 100% case 2
with discardEvents(1) I get 70% case 1, 30% case 2

This is very approximate but show what I was getting.

And this was inconsistent because with discardEvents(1) I was getting discarded and undiscarded taps when doing – almost – the same manipulations.

I guess that the internal timers are not always accurate.
Just after resuming from a long sleep was often a moment when the undiscarded taps happened.

This is on a kobo H2o v1

poire-z:
> That’s not good because discardEvent is most wanted before quick_dismiss_everything duration when most result pops up and cancel are done. In that sense, the way it was before #7114 was better.

I may have gotten lost in the journey :) Can you remember me what's the use case?
I thought we were dealing with accidental long-press, that you wish to cancel quickly.

> That way **every** tap in the quick_dismiss_everything timer close everything.

So, a first tap works ok, but you also want a second tap in that time-window to be discarded?
Well, not sure that's a good idea, to block normal (I mean, not you :) users for this period of time.

And if it is because you want to tap everywhere to cancel it when it shows quickly, well, you might need to get the habit to tap in the the bottom right corner, where the Close button is on your large window.

I quite like the latest behaviour, as it's simple to comprehend. I don't feel like complicating it with another hidden TrapWidget and blocking.

> About discardEvents()

I guess that now that it is issued after 3s, the different behaviour is less critical? This is now for my use case (waiting for a result, being impatient, interrupting it, getting a chance to not cancel it if it just happens to be shown) and not really for yours (even more now that your results come quicker) ?

Mel-kior:
> I may have gotten lost in the journey :) Can you remember me what's the use case?
I thought we were dealing with accidental long-press, that you wish to cancel quickly.

Yep, accidental long-press, cancel when wrong selection or change of mind.
Most of these cancel are done in the 3sec where most dict result window appear. So, a high probability of tap during the refresh and the need of discardEvent. A 0.6sec discardEvent delay is still quicker than some unwanted dict window interaction.
That why I think removing the discardEvent before quick_dismiss_everything duration is a bad.

> So, a first tap works ok, but you also want a second tap in that time-window to be discarded?
Well, not sure that's a good idea, to block normal (I mean, not you :) users for this period of time.

Not a second tap. It’s for situation where the tap to cancel happen when or just after the dict "result found" window appear. Instead of just discard the tap we could, if the tap still in the quick_dismiss_everything time-window, close the dict window.

If the delay is keep small I don’t think that will block legitimate dict window interaction ( like the dismissEvent delay ).

> > About discardEvents()
> 
> I guess that now that it is issued after 3s, the different behaviour is less critical? This is now for my use case (waiting for a result, being impatient, interrupting it, getting a chance to not cancel it if it just happens to be shown) and not really for yours (even more now that your results come quicker) ?

Even if used before the 3s it’s not that critical – it merely convenience – but seems to show underlying problem ? ( and having unreliable function is not great usability wise ).

There seems to be some quite big ( up to ~1sec ) internal timer inaccuracy with the kobo H2o that may be fix ( or can’t? ). This would benefit not only discardEvents() but every time dependent functions no ?

I didn’t look and don’t know how timers are used but I can do somes test if that help !

poire-z:
> That why I think removing the discardEvent before quick_dismiss_everything duration is a bad.

But I don't want to block anything about normal users, when they get a result quickly. 3s is enough to notice the result comes from a French dict when you expect to read the one from an English dict that will come after. So, his tap to go to next will get ignored, and he'll have to wait 2s...

> Not a second tap. It’s for situation where the tap to cancel happen when or just after the dict "result found" window appear. Instead of just discard the tap we could, if the tap still in the quick_dismiss_everything time-window, close the dict window.

I wrote ealier:
> And if it is because you want to tap everywhere to cancel it when it shows quickly, well, you might need to get the habit to tap in the the bottom right corner, where the Close button is on your large window.

If you get the habit to tap in the bottom right, it should work for you, no?

poire-z:
> I guess that the internal timers are not always accurate.
Just after resuming from a long sleep was often a moment when the undiscarded taps happened.
This is on a kobo H2o v1

> There seems to be some quite big ( up to ~1sec ) internal timer inaccuracy with the kobo H2o that may be fix ( or can’t? ). This would benefit not only discardEvents() but every time dependent functions no ?

We did a lot (and I think what's possible) about timers in #6798 - especially 2nd commit 633d43b32283fd3c24cedbf7c77db3092f6fec59 .
You probably see when starting KOReader in your crash.log:
`INFO event times are not real clock time: some adjustments will be made`

NiLuJe:
Yep, events are real clock time on a H2O ;).

Mel-kior:
> > That why I think removing the discardEvent before quick_dismiss_everything duration is a bad.
> 
> But I don't want to block anything about normal users, when they get a result quickly. 3s is enough to notice the result comes from a French dict when you expect to read the one from an English dict that will come after. So, his tap to go to next will get ignored, and he'll have to wait 2s...

I may not have been clear sorry, but there are 2 differents things :
- The remove of the 600ms discardEvent when result are show and that I think is necessary. It was doing it’s job quite well ( excluding the H2o inaccuracy ) before you remove it in #7114. The 0.6sec saving if remove aren’t worth the unwanted dict interactions that can occur.
- The idea of replacing the discardEvent ( when in the quick_dismiss_everything time-window ) with a "close dict window" trapwidget that would get a small max timeout ( 0.6sec/1sec /1.5sec? ) so no "normal" user get blocked input.

----------

> If you get the habit to tap in the bottom right, it should work for you, no?

No it’s not convenient for me, and I think the trapwidget idea could be implemented. But I can do it in a personal patch too : )

poire-z:
> The remove of the 600ms discardEvent when result are show and that I think is necessary. It was doing it’s job quite well 

Above, 7 days ago, you wrote:
> I personally think more logic that even if the lookup is show, if still in the 1.5sec fast cancel timer, a tap should dismiss the lookup and not just getting blocked by discardEvent().

These 2 statements feel contradictory.
If accidental touch, and an unwanted result shows quickly, you don't want a discardEvent() that will prevent you from getting rid of it - that was your earlier and first requirement ! :) 

But I expect I understand correctly: you don't want this tap to interact with the dict window, but to close it. A discardEvent is not the tool.

> The idea of replacing the discardEvent ( when in the quick_dismiss_everything time-window ) with a "close dict window" trapwidget 

That's a bit overkill. This could be just handled in the DictQuickLookup onTap handler: get a timetamp in onShow(), and in onTap, it now - timestamp < 0.6sec/1sec /1.5sec, just have it close itself.
But I don't know, if really feels like a stack of little timers and tricks (that I helped starting :/) for both our strange edgy use cases :/

Mel-kior:
> But I expect I understand correctly: you don't want this tap to interact with the dict window, but to close it.

Yes ideally and it seems logic because the tap was to "quick dismiss".
But if this is too complex to implement having a discardEvent is still better than nothing ! Yes if nothing you can close the window 0.6sec earlier but if you interact you can add time too.

> That's a bit overkill. This could be just handled in the DictQuickLookup onTap handler: get a timetamp in onShow(), and in onTap, it now - timestamp < 0.6sec/1sec /1.5sec, just have it close itself.
But I don't know, if really feels like a stack of little timers and tricks (that I helped starting :/) for both our strange edgy use cases :/

This seems a better way to do it indeed.

I don’t think that is that edgy, it’s good shortcuts to have more fluid user experience. It’s little thing like that that differentiate clunky vs fluid UX.
And discartEvent() is more a general thing to circonvent e-ink slow refresh.

PR #7101, opened 2021-01-01
Add deprecation messages to plugins

pazos:
And deprecate goodreads

asdf

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7101)

### DIFF
--- a/frontend/pluginloader.lua
+++ b/frontend/pluginloader.lua
@@ -1,11 +1,24 @@
 local lfs = require("libs/libkoreader-lfs")
 local logger = require("logger")
+local _ = require("gettext")
 
 local DEFAULT_PLUGIN_PATH = "plugins"
+
+-- plugin names that were removed and are no longer available.
 local OBSOLETE_PLUGINS = {
     storagestat = true,
     kobolight = true,
 }
+-- deprecated plugins are still available, but show a hint about its deprecation on UI
+local function getMenuTable(plugin)
+    local t = {}
+    t.name = plugin.name
+    t.fullname = string.format("%s%s", plugin.fullname or plugin.name,
+        plugin.deprecated and " (" .. _("deprecated") .. ")" or "")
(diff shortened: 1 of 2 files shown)


frontend/pluginloader.lua
```diff
@@ -1,11 +1,24 @@
...
+local _ = require("gettext")
 
 local DEFAULT_PLUGIN_PATH = "plugins"
+
+-- plugin names that were removed and are no longer available.
 local OBSOLETE_PLUGINS = {
     storagestat = true,
     kobolight = true,
 }
+-- deprecated plugins are still available, but show a hint about its deprecation on UI
```
Frenzie, line 12:
```suggestion
-- Deprecated plugins are still available, but show a hint about deprecation.
```

frontend/pluginloader.lua
```diff
@@ -1,11 +1,24 @@
...
+local function getMenuTable(plugin)
+    local t = {}
+    t.name = plugin.name
+    t.fullname = string.format("%s%s", plugin.fullname or plugin.name,
+        plugin.deprecated and " (" .. _("deprecated") .. ")" or "")
```
Frenzie, line 17: Would something like `outdated` be clearer? Deprecated in this sense is computer jargon.

plugins/goodreads.koplugin/_meta.lua
```diff
@@ -3,4 +3,5 @@ return {
     name = "goodreads",
     fullname = _("Goodreads"),
     description = _([[Allows browsing and searching the Goodreads database of books.]]),
+    deprecated = "Goodreads API was closed. Some keys still work. This plugin will be removed soon",
```
Frenzie, line 6:
```suggestion
    deprecated = "The Goodreads API has been discontinued. Some keys might still work, but this plugin will be removed soon.",
```

PR #7104, opened 2021-01-02
CRE: allow both Top/Bottom margins fine tuning

poire-z:
By having the same DoubleSpinWidget launched when hitting <kbd>...</kbd> on any of Top margin or Bottom margin progress bars.
DoubleSpinWidget: remove blank space when no info_text.

Follow up to @yparitcher #7021, solving https://github.com/koreader/koreader/pull/7021#issuecomment-749389043:
>> We already have a 2-values picker for L/R Margin.
As when tweaking top or bottom, we may want to adjust the other, could it be the same 2-values picker allowing tweaking both top and bottom margin, when launched from either ... ?
(Probably less simple as the current tweaks so far :)

> not easy,
L/R are done with one event hence one picker with 2 values, T/B are 2 separate events and configdialog does not provide a way to tie them together.
so out of scope for now.

image

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7104)

### DIFF
--- a/frontend/ui/data/creoptions.lua
+++ b/frontend/ui/data/creoptions.lua
@@ -94,6 +94,7 @@ Note that this may not be ensured under some conditions: in scroll mode, when a 
                 name_text_hold_callback = optionsutil.showValuesHMargins,
                 more_options = true,
                 more_options_param = {
+                    name_text = _("Left/Right Margins"),
                     left_min = 0,
                     left_max = 140,
                     left_step = 1,
@@ -154,10 +155,23 @@ In the top menu → Settings → Status bar, you can choose whether the bottom m
                 name_text_hold_callback = optionsutil.showValues,
                 more_options = true,
                 more_options_param = {
-                  value_min = 0,
-                  value_max = 140,
-                  value_step = 1,
-                  value_hold_step = 5,
+                    -- Allow this to tune both top and bottom margins,
+                    -- handling 2 setting names and sending 2 events
+                    -- (we'll get the exact same DoubleSpinWidget in
+                    -- the b_page_margin setting just below)
+                    name_text = _("Top/Bottom Margins"),
+                    names = { "t_page_margin", "b_page_margin" },
+                    events = { "SetPageTopMargin", "SetPageBottomMargin" },
+                    left_text = _("Top"),
+                    left_min = 0,
+                    left_max = 140,
+                    left_step = 1,
+                    left_hold_step = 5,
+                    right_text = _("Bottom"),
+                    right_min = 0,
+                    right_max = 140,
+                    right_step = 1,
+                    right_hold_step = 5,
                 },
             },
             {
@@ -194,10 +208,20 @@ In the top menu → Settings → Status bar, you can choose whether the bottom m
                 help_text = _([[In the top menu → Settings → Status bar, you can choose whether the bottom margin applies from the bottom of the screen, or from above the status bar.]]),
                 more_options = true,
                 more_options_param = {
-                  value_min = 0,
-                  value_max = 140,
-                  value_step = 1,
-                  value_hold_step = 5,
+                    -- Similar as for t_page_margin above
+                    name_text = _("Top/Bottom Margins"),
+                    names = { "t_page_margin", "b_page_margin" },
+                    events = { "SetPageTopMargin", "SetPageBottomMargin" },
+                    left_text = _("Top"),
+                    left_min = 0,
+                    left_max = 140,
+                    left_step = 1,
+                    left_hold_step = 5,
+                    right_text = _("Bottom"),
+                    right_min = 0,
+                    right_max = 140,
+                    right_step = 1,
+                    right_hold_step = 5,
                 },
             },
         }
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -1126,7 +1126,13 @@ function ConfigDialog:onConfigMoreChoose(values, name, event, args, name_text, d
             if more_options_param.left_min then -- DoubleSpingWidget
                 local DoubleSpinWidget = require("ui/widget/doublespinwidget")
                 -- (No support for value_table - add it if needed)
-                local curr_values = self.configurable[name]
+                local curr_values
+                if more_options_param.names then -- allows managing 2 different settings
+                    curr_values = { self.configurable[more_options_param.names[1]],
+                                    self.configurable[more_options_param.names[2]] }
+                else
+                    curr_values = self.configurable[name]
+                end
                 widget = DoubleSpinWidget:new{
                     width = math.floor(Screen:getWidth() * 0.6),
                     left_text = more_options_param.left_text,
@@ -1152,8 +1158,17 @@ function ConfigDialog:onConfigMoreChoose(values, name, event, args, name_text, d
                             self:closeDialog()
                         end
                         local value_tables = { left_value, right_value }
-                        self:onConfigChoice(name, value_tables)
-                        if event then
+                        if more_options_param.names then
+                            self:onConfigChoice(more_options_param.names[1], left_value)
+                            self:onConfigChoice(more_options_param.names[2], right_value)
+                        else
+                            self:onConfigChoice(name, value_tables)
+                        end
+                        if more_options_param.events then
+                            self:onConfigEvent(more_options_param.events[1], left_value, nil)
+                            self:onConfigEvent(more_options_param.events[2], right_value, refresh_callback)
+                            self:update()
+                        elseif event then
                             args = args or {}
                             self:onConfigEvent(event, value_tables, refresh_callback)
                             self:update()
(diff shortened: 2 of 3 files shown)


hius07:
When we hit fine tuning in the Margins tab of the bottom menu, the menu itself disappears.
It doesn't disappear when hitting fine tuning in all other tabs.
It seems good to keep the menu visible.

NiLuJe:
That's on purpose in order to be able to visualize the effect on the bottom margins.

(Similarly, it used to disappear when tapping the fine-tune +/- buttons. Well disappear and then reappear. See https://github.com/koreader/koreader/pull/7021 for more details about why this trickier).

hius07:
Then, after fine tuning left or right margin there is an infomessage Margins set to: with horizontal value only, no left and right margin values (which may differ).

hius07:
> That's on purpose in order to be able to visualize the effect on the bottom margins.

OK, I see

poire-z:
> It doesn't disappear when hitting fine tuning in all other tabs.
It seems good to keep the menu visible.

Except that when tweaking margins, it's better to have all four margins visible, not covered by an opened menu, so you can appreciate their balance.
I think the benefit of having it a bit different is superior to the one of being consistent everywhere :)

Or do you mean that once done, once the numberpicker is closed, we should re-open the menu ? (Might be a bit tedious, dunno - and usually, one may go check how it looks on a few other pages.)

hius07:
> Or do you mean that once done, once the numberpicker is closed, we should re-open the menu ?
BTW, the menu is reopened when selecting fixed values after infomessage disappears.

poire-z:
> Then, after fine tuning left or right margin there is an infomessage Margins set to: with horizontal value only, no left and right margin values (which may differ).

Right.
https://github.com/koreader/koreader/blob/a44c178d7423575b535e521a863c63398a1b360b/frontend/apps/reader/modules/readertypeset.lua#L495-L505
@NiLuJe : we can show left and right here if they are different - should we replace "horizontal" with them - or do you want to keep it when left == right ?

NiLuJe:
@poire-z: I'll admit that I've never really seen the appeal of L/R being different, but tweaking the IM is fine by me ;).

hius07:
And (for further consideration) I would propose to remove the Sync switcher and just have four separate settings L/R/T/B.

poire-z:
> I  would propose to remove the Sync switcher and just have four separate settings L/R/T/B

The Sync T/B has its purpose: once you changed T or B to finally be able to get one more line of text on the page - and then T vs B might be unbalanced. Just hit SyncT/B to have them rebalanced. It avoids the user to have to do the computation to move some pixels to the other.

> I'll admit that I've never really seen the appeal of L/R being different

Yes. But lately, I've seen quite a few French books that had different L/R margins - and it's really noticable, I wonder why publishers would do that. "Ignore publishers page margins" might work, but sometimes not when this difference is on some inner DIV or P. So, use others style tweaks, or make one specific.
I guess that when you're in a hurry or not at ease with CSS, this new L/R margins tuning might be another solution to that issue.

> > Or do you mean that once done, once the numberpicker is closed, we should re-open the menu ?

> BTW, the menu is reopened when selecting fixed values after infomessage disappears.

I can fix that, to have the bottom menu re-shown once closing the numberpickers.
We hide it for margins, but I wonder if we should also hide it for all of them (more room to see the effect of word spacing, etc...) and for consistency. Thoughts?

hius07:
> I wonder if we should also hide it for all of them

Yes, it makes sense, to hide and then reopen, as the bottom menu on my PW2 covers more than one third of the screen.

poire-z:
@NiLuJe : in this situation, the windows stack being:
Notification > SpinWidget > ConfigDialog > ReaderUI.

image

Is there any good reason for this bit of code, when handling an event (say Tap), to do that:
- passed to Notification (top widget, which will let it propagate when it has a timeout)
- **not passed to SpinWidget** (because it has no active_widget and is not always_active)
- passed to ConfigDialog (because it is always_active), and so will be closed if tap outside (and we end up with the SpinWidget alone) or some other toggle handled if tap inside, on such a thing, and we may end up with this :) : 
image

https://github.com/koreader/koreader/blob/47568ae9b48271ef010e58d1deb583e6aa5821aa/frontend/ui/uimanager.lua#L879-L913

```
01/10/21-17:30:53 WARN  top widget frontend/ui/widget/notification.lua
01/10/21-17:30:53 DEBUG close widget: table: 0x7fc548d55208
01/10/21-17:30:53 DEBUG setDirty via a func from widget nil
01/10/21-17:30:53 DEBUG Lower widget ReaderUI covers the full screen
01/10/21-17:30:53 DEBUG setDirty nil from widget ReaderUI w/ NO region
01/10/21-17:30:53 DEBUG setDirty nil from widget table: 0x7fc548e6cf58 w/ NO region
01/10/21-17:30:53 DEBUG setDirty nil from widget table: 0x7fc548b06ca0 w/ NO region
01/10/21-17:30:53 WARN  widget frontend/ui/widget/spinwidget.lua
01/10/21-17:30:53 WARN  widget frontend/ui/widget/spinwidget.lua not yet checked
01/10/21-17:30:53 WARN  widget frontend/ui/widget/spinwidget.lua poor thing, ignored
01/10/21-17:30:53 WARN  widget frontend/ui/widget/configdialog.lua
01/10/21-17:30:53 WARN  widget frontend/ui/widget/configdialog.lua not yet checked
01/10/21-17:30:53 WARN  widget frontend/ui/widget/configdialog.lua is_always_active
```

```diff
@@ -892,22 +893,28 @@ function UIManager:sendEvent(event)
     local checked_widgets = {top_widget}
     for i = #self._window_stack, 1, -1 do
         local widget = self._window_stack[i]
+        logger.warn("widget", debug.getinfo(widget.widget.init).short_src)
         if checked_widgets[widget] == nil then
+        logger.warn("widget", debug.getinfo(widget.widget.init).short_src, "not yet checked")
             -- active widgets has precedence to handle this event
             -- Note: ReaderUI currently only has one active_widget: readerscreenshot
             if widget.widget.active_widgets then
+        logger.warn("widget", debug.getinfo(widget.widget.init).short_src, "has active widgets")
                 checked_widgets[widget] = true
                 for _, active_widget in ipairs(widget.widget.active_widgets) do
+        logger.warn("  sub widget", debug.getinfo(active_widget.init).short_src)
                     if active_widget:handleEvent(event) then return end
                 end
             end
             if widget.widget.is_always_active then
+        logger.warn("widget", debug.getinfo(widget.widget.init).short_src, "is_always_active")
                 -- active widgets will handle this event
                 -- Note: is_always_active widgets currently are widgets that want to show a keyboard
                 -- and readerconfig
                 checked_widgets[widget] = true
                 if widget.widget:handleEvent(event) then return end
             end
+        logger.warn("widget", debug.getinfo(widget.widget.init).short_src, "poor thing, ignored")
         end
     end
 end
```

(Found this handy way to log what a widget is :) `debug.getinfo(widget.widget.init).short_src` )

Reformulated: is there any reason to ignore a widget in the stack ? Why not send it to it ? If a widget does not handle the event, it will let it propagate.

If there is one, any other idea (than fixing UIManager to not ignore anything) to correctly handle the above case?

poire-z:
Well, tried it, and there are some side effects :) In this, taping outside the text dialog and the keyboard allows turning pages :) 
image

I guess, until I discover other case where this behaviour is problematic, it's safer to try finding another way for Notification to be more transparent (like don't let the tap propagate, but resent a new event after they are closed)...

NiLuJe:
Yeah, I'm afraid I have no experience with that specific bit of window stack traversal ;).

NiLuJe:
If anything, the only thing that sounds possibly illogical in your example (not sure of the consequences if it were otherwise, though) would be COnfigDialog being flagged as always_active?

Which wouldn't achieve what you're trying to do anyway, so, just my £0.02 ;p.

NiLuJe:
(Jotting down the debug trick for later, though, that's indeed useful :).

poire-z:
Some idea:
Have Notification be the only one with `Notification.transparent_to_event_and_close_on_any_event`.
And UIManager, in that bit of code, if `top_widget.transparent_to_event_and_close_on_any_event`, just close it, remove from stack, and continue processing the event with the new top_widget.
So, Notification, whether they have a timeout or not, would be auto discarded on any event - which I think is what we expect for them: to just show, and not be bothering. At least I do, they bother me a lot: I never know if they have a timeout or not, and if I need to tap, and the different behaviour depending if I tap just before or just after the timeout: so, I usually wait 5s :) before going with the "it's not a timeout" action :)
Only issue: people might, like you, _expect to need three taps to close three layers of widgets_. So, if they are used to do that additional tap when bothered by a notification, they'll trigger something else.


I just have to find a better name for that property :) any idea ?

NiLuJe:
I'd KISS: `toast`, and just explain the logic in the only place it'll be used, namely the UIManager loop?

NiLuJe:
I personally don't mind either way, but I can very much see some people having the "bat that fly away" reflex when a toast shows up ^^.

PR #7106, opened 2021-01-03
add noop callback to cache size to prevent artifacts

yparitcher:
I was getting this:  (the lighter one is invisible on E-ink)

FileManager_2021-01-02_211039

FileManager_2021-01-02_211044

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7106)

### DIFF
--- a/plugins/coverbrowser.koplugin/main.lua
+++ b/plugins/coverbrowser.koplugin/main.lua
@@ -472,7 +472,8 @@ function CoverBrowser:addToMainMenu(menu_items)
                             local sstr = BookInfoManager:getDbSize()
                             return _("Current cache size: ") .. sstr
                         end,
-                        -- no callback, only for information
+                        keep_menu_open = true,
+                        callback = function() end, -- no callback, only for information
                     },
                     {
                         text = _("Prune cache of removed books"),


NiLuJe:
Yeah, that'd be alpha-blending artifacts ;).

poire-z:
I'm fine with this local fix (dummy callback + keep_menu_open) - but isn't there a need for a more global fix for such cases of no callback? (in the flash menu items code, but then, may be others bits of the code?)
Or is it really an error/ill case to not provide any callback to menu items?
(This happens too with v2020.11, so nothing related to the recent blit changes :)

NiLuJe:
Probably, but think the fix ought to be in Menu (i.e., if no callback, insert a noop; or get rid of the tap/hold handlers altogether or something).

Frenzie:
I think that's what @poire-z said? ^_^

poire-z:
Oh, yes, may be :) 
Yep, may be a fix in the touchmenu code to prevent flashing menus with no callback (or properly unflash it with a dummy callback) ? (so for the next time we add a dummy menu item, it doesn't matter if we forget the callback).
But I was rather thinking about the drawing artifacts, as blitting was the main topic of these last weeks :) Dunno if what we saw here (or rather, that we didn't see - only @yparitcher did :), we might see it (or rather, not see :) elsewhere ?

NiLuJe:
Meh, vaguely looked at it, there's seven billion different types of different callbacks possible, so, this'll have to do ;).

NiLuJe (approved):

PR #7112, opened 2021-01-05
Legacy Kindle: Unbreak screensavers

### LINKED ISSUE #7108: Kindle 4NT:  When pressing the power button to lock, nothing changes
* KOReader version:  2020.12
* Device: Kindle 4NT Black

#### Issue

When pressing the power button to lock the screen, nothing changes (visually), however buttons no longer do anything.  Pressing the power button will unlock the device again, and it continues as normal.

I am running KoReader with KUAL, with the non-firmware button.  I have a screensaver plugin installed, so it should show my screensaver.

NiLuJe:
* Flag the K2/K2i & DX/DXg as `supportsScreensaver`, since these cannot be SO.
* On the K3 & the K4, where we can't detect SO properly, which means they are forcibly flagged as NOT `supportsScreensaver`, unbreak native screensaver handling by doing the SIGSTOP/SIGCONT dance with cvm. I'm fairly sure this used to be implemented, and/or I thought it was and never actually checked. In any case, it's probably been broken for ages.

Fix #7108

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7112)

### DIFF
--- a/frontend/device/kindle/device.lua
+++ b/frontend/device/kindle/device.lua
@@ -58,6 +58,7 @@ end
 
 --[[
 Test if a kindle device has Special Offers
+FIXME: Find a way to probe for SO status on FX < 5.x
 --]]
 local function isSpecialOffers()
     -- Look at the current blanket modules to see if the SO screensavers are enabled...
@@ -125,6 +126,11 @@ function Kindle:initNetworkManager(NetworkMgr)
 end
 
 function Kindle:supportsScreensaver()
+    -- NOTE: The first ad-supported device was the K3
+    if self.model == "Kindle2" or self.model == "KindleDXG" then
+        return true
+    end
+
     if self.isSpecialOffers then
         return false
     else
--- a/frontend/device/kindle/device.lua
+++ b/frontend/device/kindle/device.lua
@@ -126,11 +126,6 @@ function Kindle:initNetworkManager(NetworkMgr)
 end
 
 function Kindle:supportsScreensaver()
-    -- NOTE: The first ad-supported device was the K3
-    if self.model == "Kindle2" or self.model == "KindleDXG" then
-        return true
-    end
-
     if self.isSpecialOffers then
         return false
     else
@@ -276,6 +271,7 @@ local Kindle2 = Kindle:new{
     canHWInvert = no,
     canUseCBB = no, -- 4bpp
     canUseWAL = no, -- Kernel too old to support mmap'ed I/O on /mnt/us
+    supportsScreensaver = yes, -- The first ad-supported device was the K3
 }
 
 local KindleDXG = Kindle:new{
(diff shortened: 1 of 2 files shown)


NiLuJe:
Only very vaguely and quickly tested on a K4, so, it may suffer from any of the known race issues with that approach we had to deal with on FW 5.x.

Qwerty-Space:
I have done some testing on the K4.  I think some of it is better off being said in private, but I have noticed some issues :-)

Apart from that, it works nicely.  Thank you

NiLuJe:
Well, you can reach me on Gitter or on MobileRead, I guess ;).

yparitcher:
not new, confirmed broken on 2019.05 (haven't used my NT in a while :) also didn't notice as i had custom SS showing)
btw do non SO have the directory `/opt/amazon/screen_savers/adunits`?

NiLuJe:
@yparitcher: Alas, yes, but, perhaps the content could be a hint?

```
┌─(ROOT@kindle:pts/0)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~)─┐
└─(1.03:46%:03:47:84%:#)── ls -lash /opt/amazon/screen_saver/adunits/600x800                                                                                                                                                                                                                                                                           ──(Thu, Dec 03)─┘
   1 drwxr-xr-x    2 root     root         1.0k Dec 31  2011 .
   1 drwxr-xr-x    3 root     root         1.0k Sep  3  2011 ..
  12 -rw-r--r--    1 root     root        11.8k Sep  3  2011 _banner_gif
  12 -rw-r--r--    1 root     root        11.7k Sep  3  2011 banner-unregistered.gif
   1 -rw-r--r--    1 root     root          409 Sep  3  2011 details-unregistered.xml
   1 -rw-r--r--    1 root     root          408 Sep  3  2011 details.xml
 144 -rw-r--r--    1 root     root       142.4k Sep  3  2011 screensaver-unregistered.gif
 144 -rw-r--r--    1 root     root       142.5k Sep  3  2011 screensvr.gif
   5 -rw-r--r--    1 root     root         5.0k Sep  3  2011 thumb.gif
```

yparitcher:
Same on a SO device

NiLuJe:
Assuming you actually have been sent a new ad, has perhaps the `screensvr.gif` file has been updated?

(On a non-SO device, it's the stock "letters" "Please connect to Wi-Fi to get spam^Wads" ^^).

yparitcher:
I don't let my kindles talk to amazon (so i cant check), but the downloaded ads are placed in `/mnt/us/system/.assets` (I have a old kual script to delete that folder to clear the ads)

NiLuJe:
Ah! That one definitely doesn't exist, so, eh, it'll have to do, I guess.

yparitcher:
I somehow convinced my device that it is not SO (either by reseting it and not letting it find out otherwise, or by tweeking something) but it was a few years so I don't remeber all the details

NiLuJe:
Yeah, my SO PW1 never did quite manage to get sent new ads after a factory reset when I lent it to someone. I still get the default SO screen (with the swipe), though.

Anyyyyyway ^^.

PR #7113, opened 2021-01-06
A batch of random fixes

NiLuJe:
Stuff mentioned on Gitter over the last few days,, or that I encountered along the way ;).

(Rebase me, history is sane).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7113)


NiLuJe:
As far as the ScreenSaver thing is concerned, I *could* restore the final `else return fallback end` branch, bit this feels exceedingly sneaky and non-obvious, so it'll instead use 0/N/A. Not that this should really happen, anyway.

poire-z (approved):

Frenzie (approved):

PR #7114, opened 2021-01-06
DictQuickLookup: re-add "tap on title" to set as preferred dict

### LINKED ISSUE #7109: Regressions in dictionaries management
* KOReader version: last development version as of today, January 5. 
* Device: Kobo Mini

#### Issue
Until now, if you had several dictionaries installed and activated, you could choose the default dictionary for a determined book by searching a word, going to the prefered dictionary in the dictionary search window and then touching the menu bar. This feature doesn't work anymore. It is not possible to choose a default/prefered dictionary for a given book.

#### Steps to reproduce
See above.

poire-z:
#### Statistics plugin: reorder Reset menu items

First Reset current book, then Reset other books by name, then Reset other books by duration - feels more logical than Reset current book sandwitched between the 2 others.

#### Search: fix matches not highlighted on current page

Regression (typo from cut and paste) from #6851. See https://github.com/koreader/koreader/pull/6851#issuecomment-754672318

#### DictQuickLookup: re-add "tap on title" to set as preferred dict

Restore feature removed in #7029: tap on title to set dict of current result as a preferred dictionary for this document.
Changes from former implementation:
- trust sdcv for ordering, instead of doing it ourselves while looking at the results (which didn't always work)
- allow setting multiple preferred dicts
- order of preference is shown with a number as a prefix to the dictionary name in the title
- taping again allows removing it from preferred dicts, and taping again allows setting it back as the first preferred dict

See discussion in #7109. Closes #7109.

image

image

image

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7114)

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -149,6 +149,7 @@ function ReaderDictionary:init()
     end
     -- Prepare the -u options to give to sdcv the dictionary order and if some are disabled
     self:updateSdcvDictNamesOptions()
+
     if not lookup_history then
         lookup_history = LuaData:open(DataStorage:getSettingsDir() .. "/lookup_history.lua", { name = "LookupHistory" })
     end
@@ -173,16 +174,28 @@ end
 
 
 function ReaderDictionary:updateSdcvDictNamesOptions()
-    self.enabled_dict_names = {}
-
     -- We cannot tell sdcv which dictionaries to ignore, but we
     -- can tell it which dictionaries to use, by using multiple
     -- -u <dictname> options.
     -- The order of the -u options controls the dictionary order
     -- that sdcv uses to order its results.
+
+    self.enabled_dict_names = {}
+
+    -- First, insert any preferred dicts, even if globally disabled
+    -- (this might allow enabling a dict only for a specific book,
+    -- while keeping it disabled for all others)
+    local preferred_names_already_in = {}
+    if self.preferred_dictionaries then
+        for _, name in ipairs(self.preferred_dictionaries) do
+            table.insert(self.enabled_dict_names, name)
+            preferred_names_already_in[name] = true
+        end
+    end
+
     local dicts_disabled = G_reader_settings:readSetting("dicts_disabled")
     for _, ifo in pairs(available_ifos) do
-        if not dicts_disabled[ifo.file] then
+        if not dicts_disabled[ifo.file] and not preferred_names_already_in[ifo.name] then
             table.insert(self.enabled_dict_names, ifo.name)
         end
     end
@@ -856,8 +869,8 @@ function ReaderDictionary:showDict(word, results, box, link)
             -- selected link, if any
             selected_link = link,
             results = results,
-            dictionary = self.default_dictionary,
             word_box = box,
+            preferred_dictionaries = self.preferred_dictionaries,
             -- differentiate between dict and wiki
             is_wiki = self.is_wiki,
             wiki_languages = self.wiki_languages,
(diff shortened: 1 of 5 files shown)


frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -1040,12 +1050,36 @@ end
...
+            removed = true
+            table.remove(self.preferred_dictionaries, idx)
+            break
+        end
+    end
+    if not removed then -- insert it as first
+        table.insert(self.preferred_dictionaries, 1, dict)
+    end
+    UIManager:show(InfoMessage:new{
+        text = removed and T(_("%1 is no more a preferred dictionary for this document."), dict)
```
NiLuJe, line 1075: `no longer` ;).

poire-z: I'll never get that one right :/

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -1040,12 +1050,36 @@ end
...
+                        or T(_("%1 is now the first preferred dictionary for this document."), dict),
```
NiLuJe, line 1076: The `first` feels redundant here. Thoughts?

plugins/statistics.koplugin/main.lua
```diff
@@ -1897,20 +1897,20 @@ end
 function ReaderStatistics:genResetBookSubItemTable()
     local sub_item_table = {}
     table.insert(sub_item_table, {
-        text = _("Reset statistics per book"),
+        text = _("Reset statistics for current book"),
```
NiLuJe, line 1900: Probably worth fixing it to `Reset statistics for the current book` while we're here ;).

frontend/apps/reader/modules/readerdictionary.lua
poire-z: For me, it hints/documents that when you tap twice on something that was (2), it will be re-added as being (1) - as you can have multiple preferred dicts. So, it always becomes the first preferred dict.

Frenzie: most preferred?

First preferred definitely sounds a bit weird.

poire-z: "most preferred" sounds strange to my non-native-english ears :) 
But ok with it or just "preferred" (removing "first") if you feel it's fine and enough.

Frenzie: For reference:

https://eur-lex.europa.eu/LexUriServ/LexUriServ.do?uri=COM:2010:0473:FIN:EN:PDF
> Home-made  explosives,  in  turn,  are  the  tool  most  preferred  by  terrorists  and  other  criminals  to  perpetrate  attacks.

https://eur-lex.europa.eu/LexUriServ/LexUriServ.do?uri=COM:2010:0473:FIN:FR:PDF
> Or  ce  sont  précisément  les  explosifs  de  cette  nature  que  les  terroristes  et  autres  criminels  préfèrent  utiliser  pour  perpétrer des attentats.

Frenzie: > But ok with it or just "preferred" (removing "first") if you feel it's fine and enough.

Fine by me. ;-)

poire-z:
Added a 4th commit to make the dictquicklookup dismissal/block a little bit easier to comprehend - and to hopefully solve https://github.com/koreader/koreader/pull/7099#issuecomment-754980900 : 

#### ReaderDictionary: tweak timings if interrupted quickly

Rework b66d0be3 and 6205f260 from #7099 by making that simpler to comprehend:
- If interruption or search result window is shown in less than 3 secondes: no specific tweak, a tap outside will dismiss everything instantaneously.
- If done after 3 seconds: show the interruption or result window, and discard input for a few 100ms.
- UIManager:discardEvents(): increase a bit default delays.

Issue #1580, opened 2015-07-22
Feature request: support Pocket article retrieval


pdo-smith:
As a valuable enhancement I suggest you include support for Pocket (getpocket.com). This could work in the same way it does for Nickel, under articles.

I find this is the only reason I use Nickel. If you supported Pocket I could give up using Nickel for good.

Frenzie:
Pocket's license agreement includes a lot of text about sharing with third-parties etc. Anyhoo, if anyone's interested, here's the API docs: http://getpocket.com/developer/docs/overview

chrox:
The pity is that the pocket api only export urls to third party apps, no readability processed full-text. That means we need to do a lot of work in the client side to render an article from Pocket's API. And I found https://github.com/wallabag/wallabag is more promising in this respect especially it's upcoming v2 api.

Frenzie:
Wallabag sounds great! Thanks for the tip.

On September 12, 2015 7:36:08 AM GMT+02:00, Huang Xin <notifications@github.com> wrote:
>The pity is that the pocket api only export urls to third party apps,
>no readability processed full-text. That means we need to do a lot of
>work in the client side to render an article from Pocket's API. And I
>found https://github.com/wallabag/wallabag is more promising in this
>respect especially it's upcoming v2 api.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/koreader/koreader/issues/1580#issuecomment-139723330

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Frenzie:
Slightly besides the point, but I figured I'd share that there's a free Wallabag instance over at https://framabag.org/ for people who can't or don't want to bother with self-hosting.

But the actual reason I'm posting: Wallabag is thinking about implementing OPDS: https://github.com/wallabag/wallabag/issues/1253

anarcat:
I wrote a plugin called wallabako for kobo readers that does this. Head over to [gitlab](https://gitlab.com/anarcat/wallabako/) to see what's up with that...

anarcat:
I've tested wallabako with KSM/KoReader and it somewhat works. For some reason, the logfile (a simple .txt file!) doesn't render in Koreader, but otherwise koreader sees the downloaded files correctly.

[...]

PR #7115, opened 2021-01-06
WIP: Add pocket plugin

fenuks:
This is WIP plugin for https://getpocket.com. Code is based on similar plugin for wallabag.

Plugin is roughly working now, but there are few issues that I would like to discuss.

I couldn't make OAuth initialization working in code, therefore user for now is expected to enter access_token by hand. This could be surely improved. For testing purposes, I've created koreader app in Pocket Developer program, but it can be changed to something else (also in settings). In order to obtain valid `access_token` you might use this snippet (not tested, the latest testing update broke python3 for me ;P)
```python
import requests

consumer_key = '95098-ae9dbddd5a88261cc5d7d29d'
headers = {'X-Accept': 'application/json'}
request_token = requests.post('https://getpocket.com/v3/oauth/request', data={'consumer_key': consumer_key, 'redirect_uri': 'https://getpocket.com'}, headers=headers).json()['code']

print(f'''Open https://getpocket.com/auth/authorize?request_token={request_token}&redirect_uri=https://getpocket.com in broser, accept authorization, and then run
curl --request POST \
  --url https://getpocket.com/v3/oauth/authorize \
  --header 'Content-Type: application/json' \
  --header 'X-Accept: application/json' \
  --data '{
	"consumer_key": "{consumer_key}",
	"code": "{request_token}"
}'
to generate access_token
''')
```
Much code is duplicated, I suppose it would be wise to refactor it to base plugin that will be used by `wallabag` and `pocket` plugins (probably it would be easy to add Instapaper support on top of it as well).

As correctly stated in #1580, Pocket API has no readability support, thus articles are downloaded with `EpubDownloadBackend`. It works, but is not optimal, as crust filtering algorithm is not very advanced. I've been thinking about porting https://github.com/mozilla/readability, which is used in Firefox Reader Mode (it works quite well for sites I use), but it appears that `htmlparser` used by koreader doesn't support element deletion from quick skimming through [the project](https://github.com/msva/lua-htmlparser)?

I wonder if we could more feature full parser, or maybe it would make sense to implement readability algorithm in other language, and use it from lua?

There is also subject of tests. Frankly speaking, I was unable to install busted (tried with luarocks, and with AUR package). Wallabag plugin has none, but with base plugin approach, I suppose it would be good idea to add some.

### DIFF
--- a/frontend/ui/elements/filemanager_menu_order.lua
+++ b/frontend/ui/elements/filemanager_menu_order.lua
@@ -103,6 +103,7 @@ local order = {
         "cloud_storage",
         "read_timer",
         "wallabag",
+        "pocket",
         "news_downloader",
         "send2ebook",
         "text_editor",
--- a/frontend/ui/elements/reader_menu_order.lua
+++ b/frontend/ui/elements/reader_menu_order.lua
@@ -128,6 +128,7 @@ local order = {
         "progress_sync",
         "move_to_archive",
         "wallabag",
+        "pocket",
         "zsync",
         "news_downloader",
         "send2ebook",
--- a/plugins/pocket.koplugin/_meta.lua
+++ b/plugins/pocket.koplugin/_meta.lua
@@ -1,0 +1,6 @@
+local _ = require("gettext")
+return {
+    name = "pocket",
+    fullname = _("Pocket"),
+    description = _([[Synchronises articles with a Pocket server.]]),
+}
--- a/plugins/pocket.koplugin/api.lua
+++ b/plugins/pocket.koplugin/api.lua
@@ -1,0 +1,127 @@
+local http = require("socket.http")
+local JSON = require("json")
+local InfoMessage = require("ui/widget/infomessage")
+local logger = require("logger")
+local ltn12 = require("ltn12")
+local socket = require("socket")
+local UIManager = require("ui/uimanager")
+local _ = require("gettext")
+
+local Api = {}
+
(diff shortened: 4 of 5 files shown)


fenuks:
On side note, is there option to hot-reload lua code without restarting whole koreader?

Frenzie (commented):
> Wallabag plugin has none, but with base plugin approach, I suppose it would be good idea to add some.

It's probably simpler to maintain to simply have a single "read it later" plugin, which would internally consist of a base and specific implementations like Wallabag and Pocket.

Frenzie:
> On side note, is there option to hot-reload lua code without restarting whole koreader?

Nope, but feel free to experiment with something like https://github.com/anton-kl/lua-hot-reload

fenuks:
> It's probably simpler to maintain to simply have a single "read it later" plugin, which would internally consist of a base and specific implementations like Wallabag and Pocket.

I also think it should be easier. I will refactor both plugins to common base, then.

> Nope, but feel free to experiment with something like https://github.com/anton-kl/lua-hot-reload

Thank you, I will try that. I must admit I found a working on plugin a bit tedious because of modify-reload cycle (and I'm not that familiar with lua).

poire-z:
What's the status of this PR? Should we close as inactive ?

fenuks:
Sorry about that, yes it can be closed, maybe I will revisit it some time in the future. In the meanwhile I started using wallabag instead, and it has built-in support for exporting epubs, and very well working plugin.

PR #7116, opened 2021-01-07
Dispatcher: correctly set document.configurable

yparitcher:
This allows the ConfigDialog to update properly

@poire-z 
Please see if this helps :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7116)

### DIFF
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -23,6 +23,7 @@ Each setting contains:
 * args: allowed values for string.
 * toggle: display name for args
 * separator: put a separator after in the menu list
+* configurable: can be parsed from cre/kopt and used to set document.configurable. Should not be set manualy
 --]]--
 
 local CreOptions = require("ui/data/creoptions")
@@ -289,6 +290,9 @@ function Dispatcher:init()
         for y=1,#base[i].options do
             local option = base[i].options[y]
             if settingsList[option.name] ~= nil then
+                if option.name ~= nil and option.values ~= nil then
+                    settingsList[option.name].configurable = {name = option.name, values = option.values}
+                end
                 if settingsList[option.name].event == nil then
                     settingsList[option.name].event = option.event
                 end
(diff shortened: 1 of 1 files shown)


poire-z:
Looks like what I had in mind.
But I'm not a heavy dispatcher user to judge how it works.
Just did a quick test, associating to a gesture multiple actions:
- toggle nightmode (nothing to update)
- decreased font size: worked and bottom updated
- set line spacing to 80%: seems to apply to book, but bottom not updated (even after a quit/start)
- set kerning mode to off: it is applied on book, but bottom not updated (even after a quit/start)
(seems to still not work if I only set it alone to this gesture)

Unrelated:
- Line spacing number picker: long-press only increase by 2 (so, not much help vs tap which increase by 1 :)
- Seems I can't disable "Decrease font size by" ? :) I can just set the increment to 0, but it stays checked.

poire-z:
Mhhh, it happened I tested this PR ... without actually applying it on my tree :) 
My above test now seems to all work fine :) Line spacing and kerning are properly reflected in the bottom menu :)
(May be the decrease font size worked because of the onSetFontSize existing handled I mentionned in the other PR?)

frontend/dispatcher.lua
```diff
@@ -289,6 +290,9 @@ function Dispatcher:init()
         for y=1,#base[i].options do
             local option = base[i].options[y]
             if settingsList[option.name] ~= nil then
+                if option.name ~= nil and option.values ~= nil then
+                    settingsList[option.name].configurable = {name = option.name, values = option.values}
```
poire-z, line 294: I haven't looked at all the actions that reflects in the bottom menu - but are there some that would not be handled? Or it just magically works for all of them?

poire-z (approved):

yparitcher:
>     * Seems I can't disable "Decrease font size by" ? :) I can just set the increment to 0, but it stays checked.

hold to clear.

frontend/dispatcher.lua
yparitcher: the action from the bottom menu are:
https://github.com/koreader/koreader/blob/e157395660108032ca02c2acb3275d6012af23e7/frontend/dispatcher.lua#L134-L155

and they all are reflected here. ConfigDialog/Creopts use the same name value as configurable, and the values key it wat to set, (as opposed to args)

poire-z:
> Line spacing number picker: long-press only increase by 2 (so, not much help vs tap which increase by 1 :)

@yparitcher : any reason you can think of to not increase it? (like .max being most often < 5?):
```diff
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -418,3 +418,3 @@ function Dispatcher:addItem(caller, menu, location, settings, section)
                             value_step = 1,
-                            value_hold_step = 2,
+                            value_hold_step = 5,
                             value_max = settingsList[k].max,
@@ -460,3 +460,3 @@ function Dispatcher:addItem(caller, menu, location, settings, section)
                             value_step = 1,
-                            value_hold_step = 2,
+                            value_hold_step = 5,
                             value_max = settingsList[k].max,
```

yparitcher:
I probabaly copied it from somewhere, so go ahead.

PR #7118, opened 2021-01-07
Revamp "flash_ui" handling

NiLuJe:
So, err, this started as a simple investigation into why the SkimTo widget was so weirdly slow on Mk. 7 (spoiler: https://github.com/koreader/koreader-base/pull/1283/commits/926c294d87e90080fb68265c30a64a8f7069bfa9), and it turned into a weird revamp of how flash_ui highlights are handled ;p.

In short:

* Wherever possible, do an actual dumb invert on the Screen BB instead of repainting the widget, *then* inverting it (which is what the "invert" flag does).
* Instead of playing with nextTick/tickAfterNext delays, explicitly fence stuff with forceRePaint
* And, in the few cases where said Mk. 7 quirk kicks in, make the fences more marked by using a well-placed WAIT_FOR_UPDATE_COMPLETE

This revealed a fun amount of bugs and weird quirks, from widgets being completely reinstanciated on each page (hi, ConfigDialog! So glad you can munch on almost 30MB of RAM all on your lonesome!), to only Buttons being reinstanciated a billion times, which caused ghost widgets in the event queue, meaning double highlights ;).

KeyValuePage did some weird things with its Return arrow Button, too. (And highlights were generally already broken there, especially visible in the Dictionary/Wikipedia lookup pages).

I'm not super-proud of the "UIManager:getPreviousRefreshRegion" hack, but it turned out to be mildly useful ;p.

@poire-z: There was a minor quirk with the recently implemented show_delay on IM, where the scheduled show wasn't cleared on close, which caused a stray refresh on the full screen.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7118)

### DIFF
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -272,7 +272,7 @@ function Button:onTapSelectButton()
 
             if not self[1] or not self[1].invert or not self[1].dimen then
                 -- If the widget no longer exists (destroyed, re-init'ed by setText(), or not inverted: nothing to invert back
-                return
+                return true
             end
 
             -- If the callback closed our parent (which ought to have been the top level widget), abort early
@@ -280,7 +280,7 @@ function Button:onTapSelectButton()
                 print("After callback, Button is still shown")
             else
                 print("Button was closed by callback")
-                return
+                return true
             end
 
             self[1].invert = false
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -250,7 +250,7 @@ function Button:onTapSelectButton()
                 end
 
                 UIManager:widgetRepaint(self[1], self[1].dimen.x, self[1].dimen.y)
-                -- Keep the invert flag in both cases, mainly for the check below
+                -- Keep the invert flag in both cases, mainly for the early return check below
                 self[1].invert = true
             else
                 self[1].invert = true
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -254,11 +254,12 @@ function Button:onTapSelectButton()
             local t2 = os.clock()
             print(string.format("It took %9.3f ms", (t2 - t1) * 1000))
 
-            -- Force the repaint *now*, so we don't have to delay the callback to see the invert...
-            UIManager:forceRePaint()
+            -- Force the repaint *now*, so we don't have to delay the callback to see the highlight...
+            UIManager:forceRePaint() -- Ensures we have a chance to see the highlight
             self.callback()
-            UIManager:forceRePaint()
-            UIManager:waitForVSync()
+            UIManager:forceRePaint() -- Ensures whatever the callback wanted to paint will be shown *now*...
+            UIManager:waitForVSync() -- ...and that the EPDC will not wait to coalesce it with the *next* update,
+                                     -- because that would have a chance to noticeably delay it until the unhighlight.
 
             if not self[1] or not self[1].invert or not self[1].dimen then
                 -- If the widget no longer exists (destroyed, re-init'ed by setText(), or not inverted: nothing to invert back
@@ -287,7 +288,7 @@ function Button:onTapSelectButton()
             UIManager:setDirty(nil, function()
                 return "fast", self[1].dimen
             end)
-            UIManager:forceRePaint()
+            UIManager:forceRePaint() -- Ensures the unhilight happens now, instead of potentially waiting and having it batched with something else.
         end
     elseif self.tap_input then
         self:onInput(self.tap_input)
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -40,6 +40,7 @@ local Button = InputContainer:new{
     preselect = false,
     callback = nil,
     enabled = true,
+    hidden = false,
     allow_hold_when_disabled = false,
     margin = 0,
     bordersize = Size.border.button,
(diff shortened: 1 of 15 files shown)


NiLuJe:
Depends on https://github.com/koreader/koreader-base/pull/1283

NiLuJe:
* Fix an issue in Button where show/hide & enable/disable where actually all toggles, which meant that duplicate calls or timing issues would do the wrong thing. (This broke dimming some icons, and mistakenly dropped the background from FM chevrons, for example).
* Speaking of, fix Button's hide/show to actually restore the background properly (there was a stupid typo in the variable name)
* Still in Button, fix the insanity of the double repaint on rounded buttons. Turns out it made sense, after all (and was related to said missing background, and bad interaction with invert & text with no background).
* KeyValuePage suffered from a similar issue with broken highlights (all black) because of missing backgrounds.
* In ConfigDialog, only instanciate IconButtons once (because every tab switch causes a full instantiation; and the initial display implies a full instanciation and an initial tab switch).
* ConfigDialog: Restore the "don't repaint ReaderUI" when switching between similarly sized tabs (re #6131). I never could reproduce that on eInk, and I can't now on the emulator, so I'm assuming @poire-z fixed it during the swap to SVG icons.
* KeyValuePage: Only instanciate Buttons once (again, this is a widget that goes through a full init every page).
* Menu: Ditto.
* TouchMenu: Now home of the gnarliest unhilight heuristics, because of the sheer amount of different things that can happen (and/or thanks to stuff not flagged covers_fullscreen properly ;p).

poire-z:
> There was a minor quirk with the recently implemented show_delay on IM, where the scheduled show wasn't cleared on close, which caused a stray refresh on the full screen.

OK, I guess this happened only with the case that I didn't think about: where there's a timeout, and it's the timeout that is closing it, right ? (I assumed only dismiss was closing it...)

> I'm assuming @poire-z fixed it during the swap to SVG icon

May be, dunno :) The only thing I see is limiting the size of the icons - but I thought this issue was cause by the Font size widget with its increasing size - so may be it wasn't it :) We'll see on various geometry/devices.

Thanks for all the work (haven't look at each in details, but it indeed often felt like a mess the re-init of many little things on a simple update.

poire-z:
> ConfigDialog: Restore the "don't repaint ReaderUI" when switching between similarly sized tabs 

I actually see a differences with PDFs:

NiLuJe:
> > There was a minor quirk with the recently implemented show_delay on IM, where the scheduled show wasn't cleared on close, which caused a stray refresh on the full screen.
> 
> OK, I guess this happened only with the case that I didn't think about: where there's a timeout, and it's the timeout that is closing it, right ? (I assumed only dismiss was closing it...)

Yep, IIRC, that's the exact flow of things ;). (In practice, you ended up with a full-screen "ui" refresh a short time after the actual lookup popup was shown, which ended up being mostly invisible unless you happened to be hunting for weird refreshes ;).)

NiLuJe:
> > ConfigDialog: Restore the "don't repaint ReaderUI" when switching between similarly sized tabs
> 
> I actually see a differences with PDFs:
> 
> image image

Gah, PDF >_<".

poire-z:
> Gah, PDF >_<".

Remove this, and PDF will be like EPUBs, so all good I think :)
https://github.com/koreader/koreader/blob/1c9c40fb953e1872f3ca573d2b1caaefe170c64f/frontend/ui/data/koptoptions.lua#L323

NiLuJe:
@poire-z: Great minds ;p.

NiLuJe:
> Thanks for all the work (haven't look at each in details, but it indeed often felt like a mess the re-init of many little things on a simple update.

Yeah, I didn't try too hard, mostly keeping to Button & IconButton. Hopefully haven't broken anything, because those buttons should stay put (in fact, Menu already handles the hide/show dance mostly properly). A good candidate for a good spring cleanup would perhaps be KVP or Menu, they're both generally much less insane than ConfigDialog to get into ^^.

NiLuJe:
The upside is, if anything's broken, results should be consistent between the emu and actual eInk devices ;).

frontend/ui/widget/infomessage.lua
```diff
@@ -196,6 +196,9 @@ function InfoMessage:init()
 end
 
 function InfoMessage:onCloseWidget()
+    if self._delayed_show_action then
+        UIManager:unschedule(self._delayed_show_action)
+    end
```
poire-z, line 201: Might do here and below in:
```diff
function InfoMessage:dismiss()
    if self._delayed_show_action then
        UIManager:unschedule(self._delayed_show_action)
+       self._delayed_show_action = nil
```
to avoid UIManager:unschedule twice, even if the 2nd is a no-op ?

poire-z (approved):
Looks allright (not tested).
I won't experience much of it as I use flash_ui=false.
(So much code for those little flashes.. :) I would have removed all that just to not have to fix them :/)

frontend/ui/widget/infomessage.lua
NiLuJe: Yeah, good point!

frontend/ui/widget/button.lua
```diff
@@ -225,42 +227,74 @@ function Button:onTapSelectButton()
...
-                    return "fast", self[1].dimen
-                end)
+            else
+                UIManager:widgetInvert(self[1], self[1].dimen.x, self[1].dimen.y)
+            end
+            -- If the button was disabled, switch to UI to make sure the gray comes through unharmed ;).
+            UIManager:setDirty(nil, function()
+                return self.enabled and "fast" or "ui", self[1].dimen
             end)
+            --UIManager:forceRePaint() -- Ensures the unhilight happens now, instead of potentially waiting and having it batched with something else.
```
Frenzie, line 297: (just skimming)
typo

NiLuJe: Err, where?

poire-z: (Hi ! unilight, unhilight, unhighlit... unhighlight)

NiLuJe: :facepalm: :D

PR #7121, opened 2021-01-09
Don't need to keep attempting WiFi connection after success

snelg:
If we're in range of multiple known WiFi Access Points (including multiple instances of the same SSID), we don't need to keep trying to connect after the first successful connection.

Minimal change would have been replacing the `return` inside the `foreach` function with `return [a non-nil value]`.
But foreach is deprecated, and since I was touching the code anyhow, I figured I'd do that tiny update as well.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7121)

### DIFF
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -594,7 +594,7 @@ function NetworkMgr:reconnectOrShowNetworkMenu(complete_callback)
         table.sort(network_list,
            function(l, r) return l.signal_quality > r.signal_quality end)
         local success = false
-        for idx, network in pairs(network_list) do
+        for idx, network in ipairs(network_list) do
             if network.password then
                 success = NetworkMgr:authenticateNetwork(network)
                 if success then
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -594,23 +594,22 @@ function NetworkMgr:reconnectOrShowNetworkMenu(complete_callback)
         table.sort(network_list,
            function(l, r) return l.signal_quality > r.signal_quality end)
         local success = false
-        table.foreach(network_list,
-           function(idx, network)
-               if network.password then
-                   success = NetworkMgr:authenticateNetwork(network)
-                   if success then
-                       NetworkMgr:obtainIP()
-                       if complete_callback then
-                           complete_callback()
-                       end
-                       UIManager:show(InfoMessage:new{
-                           text = T(_("Connected to network %1"), BD.wrap(network.ssid)),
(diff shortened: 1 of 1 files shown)


frontend/ui/network/manager.lua
```diff
@@ -594,23 +594,22 @@ function NetworkMgr:reconnectOrShowNetworkMenu(complete_callback)
...
-                       end
-                       UIManager:show(InfoMessage:new{
-                           text = T(_("Connected to network %1"), BD.wrap(network.ssid)),
-                           timeout = 3,
-                       })
-                       return
-                   end
-               end
-           end)
+        for idx, network in pairs(network_list) do
```
NiLuJe, line 597: It *should* be an array (at least according to `frontend/ui/network/wpa_supplicant.lua`), so you should be using `ipairs` to ensure it's iterated in the right order.

frontend/ui/network/manager.lua
```diff
@@ -594,23 +594,22 @@ function NetworkMgr:reconnectOrShowNetworkMenu(complete_callback)
...
-                       end
-                       UIManager:show(InfoMessage:new{
-                           text = T(_("Connected to network %1"), BD.wrap(network.ssid)),
-                           timeout = 3,
-                       })
-                       return
-                   end
-               end
-           end)
+        for idx, network in ipairs(network_list) do
```
Galunid, line 597: A nit, but we don't use `idx` variable, so it probably should be `__` (or whatever is used as "unused" variable)?

Frenzie: I set up `dummy` in .luacheckrc. I don't like `__` because that looks like a metamethod, and `_` looks like gettext. But we don't have anything concrete written down in the styleguide atm I think.

https://github.com/koreader/koreader/blob/a44c178d7423575b535e521a863c63398a1b360b/.luacheckrc#L154

snelg: I actually did use `_` (single-character) at first, but it crashed since it's already used for `gettext` in this file. So I switched it back to `idx`, since that's what the original `foreach` was using.
I'm fine with `dummy`. I've seen some projects use `ignore` in similar cases.

Frenzie: `dummy` has been the de facto standard here for a few years, so please use that. The part about the style guide was a note to self more than anything.

snelg: Yep, that's what I changed it to when I made that last comment.

Galunid (approved):

NiLuJe (approved):

PR #7122, opened 2021-01-09
Wifi enable/disable for reMarkable 2

snelg:
Copied and then tweaked from the `cervantes` device code.
Addresses #6677 but only for rm2, not rm1.

Related note:
I like the way `reconnectOrShowNetworkMenu` automatically reconnects if it finds a known SSID, without stopping to ask you to manually re-select which SSID you want to use every time. But if I use that one instead of `showNetworkMenu`, then I don't currently see a way in the UI to *force* the `networksettings` widget to open.
For example, if for some reason I *don't* I want to connect to an SSID that's currently in-range, then I have to manually adjust the `settings/network.lua` file.
Since auto-connecting is my preferred behavior 98% of the time, I went with `reconnectOrShowNetworkMenu`

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7122)


NiLuJe:
Yeah, that was mostly tailored for Kobo, where there's a system wpa_supplicant config, and half of the Kobo Wi-Fi stuff either directly relies on it, or will use it over the KOReader NetworkManager.

(Patches to plug the KO UI into system's wpa config welcome ;p).

platform/remarkable/enable-rm2-wifi.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+if lsmod | grep -q brcmfmac; then
+    ifconfig wlan0 up
+else
+    modprobe brcmfmac
+fi
+
+# clean out any other running instances of wpa_supplicant
+killall wpa_supplicant 2>/dev/null
```
NiLuJe, line 10: `wpa_cli terminate` ought to be able to do that, hopefully in a cleaner manner ;).

snelg: Hmm, the rm2's default software runs wpa_supplicant as a DBus service, which wpa_cli doesn't connect to.
But you're right, I do prefer "cleaner" over "simpler" if they're at odds.
So replacing with `systemctl stop wpa_supplicant` AND `wpa_cli terminate`.
Both of them run in a tiny fraction of a second, and both of them behave fine regardless of which flavor of wpa_supplicant is or isn't running.

snelg: > (Patches to plug the KO UI into platform system configs welcome ;p. It was never done in order to keep the client platform-agnostic, I think. And then we kinda ruined that on Kobo by having scripts that half bypass it ;p).

I definitely understand living with and/or hacking workarounds for old architectural decisions :)

Still getting a hang of the overall project structure. I'll probably work up a patch when I decide the current situation bugs me enough ;P

snelg:
> (Patches to plug the KO UI into platform system configs welcome ;p. It was never done in order to keep the client platform-agnostic, I think. And then we kinda ruined that on Kobo by having scripts that half bypass it ;p).

I definitely understand living with and/or hacking workarounds for old architectural decisions :)

Still getting a hang of the overall project structure. I'll probably work up a patch when I decide the current situation bugs me enough ;P

pazos:
> I like the way reconnectOrShowNetworkMenu automatically reconnects if it finds a known SSID, without stopping to ask you to manually re-select which SSID you want to use every time. But if I use that one instead of showNetworkMenu, then I don't currently see a way in the UI to force the networksettings widget to open.

I thought about it a couple of times in the past, but it didn't bother me enough to look at it. My suggestion for devices with standalone network settings would be assign `reconnectOrShowNetworkMenu` to short press and `showNetworkMenu` on long press. That's pretty much how android devices work: if I'm disconnected and press the wifi toggle it will try to reconnect on the background. If I long press the same toggle I will be forwarded to a menu where I can choose a network to join.

frontend/device/remarkable/device.lua
```diff
@@ -18,6 +18,10 @@ local wacom_height = 20967 -- unscaled_size_check: ignore
 local wacom_scale_x = screen_width / wacom_width
 local wacom_scale_y = screen_height / wacom_height
 
+local function turnOffWifi()
+    os.execute("./disable-rm2-wifi.sh")
+end
```
pazos, line 23:
```lua
local function turnOnWifi(model)
    if model == "reMarkable 2" then
       os.execute("./enable-rm2-wifi.sh")
    end
end

local function turnOffWifi(model)
    if model == "reMarkable 2" then
       os.execute("./disable-rm2-wifi.sh")
    end
end
```

frontend/device/remarkable/device.lua
```diff
@@ -130,6 +134,32 @@ end
 
 function Remarkable:supportsScreensaver() return true end
 
+function Remarkable2:initNetworkManager(NetworkMgr)
```
pazos, line 137: I think we can use the same network manager for both remarkables: will add a couple of comments

pazos: on top of `initNetworkManager` do

```lua
local model = self.model
```

frontend/device/remarkable/device.lua
```diff
@@ -130,6 +134,32 @@ end
...
+    function NetworkMgr:turnOnWifi(complete_callback)
+        os.execute("./enable-rm2-wifi.sh")
```
pazos, line 139: `turnOnWifi(model)`

frontend/device/remarkable/device.lua
```diff
@@ -130,6 +134,32 @@ end
...
+        self:reconnectOrShowNetworkMenu(function()
+            self:connectivityCheck(1, complete_callback)
+        end)
+    end
+
+    function NetworkMgr:turnOffWifi(complete_callback)
+        turnOffWifi()
```
pazos, line 146: `turnOffWifi(model)`

frontend/device/remarkable/device.lua
```diff
@@ -146,6 +176,9 @@ function Remarkable1:suspend()
 end
 
 function Remarkable2:suspend()
+    -- Need to remove brcmfmac kernel module before suspend
+    turnOffWifi()
```
pazos, line 180: `turnOnWifi(self.model)`

snelg: "attempt to index local 'self' (a nil value)"
I think I get why that's happening: generic/device does
```
UIManager:scheduleIn(self.suspend_wait_timeout, self.suspend)
```

This is the first time I've ever touched Lua, so instead of trying to do some clever solution, I just hard-coded
```
turnOffWifi("reMarkable 2")
```
here, since it's inside the `Remarkable2` implementation anyhow.

Frenzie: Bit difficult to check context on my phone, but to pass self along:
```lua
UIManager:scheduleIn(self.suspend_wait_timeout, self, self.suspend)
```

snelg: Other way around (`self.suspend, self`), but yeah, I figured it was something like that.
I was hesitant to fiddle around in `generic` since I wasn't sure how hard it would be to trace through possible side-effects on other devices. But you've emboldened me :P
Both `scheduleIn` references to `self.suspend` updated.

pazos:
@snelg: AFAICT both scripts can be applied to the Rm1/Rm2. The only difference is the broadcom module, is that right?.

 In that case I would vote for doing the device type check in bash, like in https://github.com/koreader/koreader/blob/master/platform/remarkable/koreader.sh#L9-L11, and use the same scripts for both files. That would remove the need for the local functions in lua, the need to pass self to the scheduler (my bad, sorry :p) and it would be a matter of @gbyl tweaking these shell scripts to get it working on the Rm1

snelg:
> The only difference is the broadcom module, is that right?.

No idea. I only have a reMarkable 2.
I agree it makes sense to push the device type check down into the bash script if the commands are mostly the same between the two devices.

But I think it's fine to leave the added `self` parameter in those scheduler calls. All other calls to the `suspend` function use the `:` syntax, e.g. `Device:suspend()`, so this seems more consistent to me. [insert "still a Lua n00b" caveat]

pazos:
> > The only difference is the broadcom module, is that right?.
> 
> No idea. I only have a reMarkable 2.
> I agree it makes sense to push the device type check down into the bash script if the commands are mostly the same between the two devices.
> 
> But I think it's fine to leave the added `self` parameter in those scheduler calls. All other calls to the `suspend` function use the `:` syntax, e.g. `Device:suspend()`, so this seems more consistent to me. [insert "still a Lua n00b" caveat]

Sorry for the delay.

In this specific case I would prefer to keep that stuff in bash. I don't mind if the bash script does nothing in a Remarkable1. At least it is easier to tweak if some user wants. Having a single call to `os.execute(something)` makes the code leaner and would make easier to implement other backends to handle the network, like https://github.com/koreader/koreader/pull/6681

snelg:
Moved logic to bash.

frontend/device/generic/device.lua
```diff
@@ -221,7 +221,7 @@ end
 function Device:rescheduleSuspend()
     local UIManager = require("ui/uimanager")
     UIManager:unschedule(self.suspend)
-    UIManager:scheduleIn(self.suspend_wait_timeout, self.suspend)
+    UIManager:scheduleIn(self.suspend_wait_timeout, self.suspend, self)
```
pazos, line 224: no longer needed

frontend/device/generic/device.lua
```diff
@@ -323,7 +323,7 @@ function Device:onPowerEvent(ev)
                     network_manager:turnOffWifi()
                 end
             end
-            UIManager:scheduleIn(self.suspend_wait_timeout, self.suspend)
+            UIManager:scheduleIn(self.suspend_wait_timeout, self.suspend, self)
```
pazos, line 326: ditto ^

pazos (approved):

pazos:
Woops. That was intended for 2021.02?. Shall I revert?

Sorry :P

Frenzie:
No problem, mainly I just didn't want to review it yet. ;-)

gbyl:
> […] and it would be a matter of @gbyl tweaking these shell scripts to get it working on the Rm1

@pazos This works perfectly on RM1. All I did was comment out the following lines:

https://github.com/koreader/koreader/blob/9f4e7456cfc105383cead062eb9cb51d7463eab5/platform/remarkable/enable-wifi.sh#L3-L5

https://github.com/koreader/koreader/blob/9f4e7456cfc105383cead062eb9cb51d7463eab5/platform/remarkable/enable-wifi.sh#L20

https://github.com/koreader/koreader/blob/9f4e7456cfc105383cead062eb9cb51d7463eab5/platform/remarkable/disable-wifi.sh#L5-L7

https://github.com/koreader/koreader/blob/9f4e7456cfc105383cead062eb9cb51d7463eab5/platform/remarkable/disable-wifi.sh#L18

This has been a hugely anticipated feature. Thank you so much for your hard work @snelg !

Frenzie:
That simple? Sounds like good news. :+1:

gbyl:
@Frenzie Yes! Network selection and everything. It even helps simplify the code for both devices.

PR #7124, opened 2021-01-09
ScreenSaver: Unbreak "disable" mode

NiLuJe:
Regression since f67296942f9c28e76c9c45b27c3b8ad0c1dc7860

That's a bit of legacy/backward compat mess, because we now have an explicit "no background" setting, but the "disable" mode predates it (*and* the white bg option, for that matter), and as such, just assumes the the default magically morphs from black to nil in this mode.

I mistakenly reverted that bit of the hack in f67296942f9c28e76c9c45b27c3b8ad0c1dc7860 by "fixing" the check to actually honor settings ^^.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7124)


NiLuJe:
(That was the hunk @ L390 in f672969, FWIW)

Frenzie (approved):
Looks ok, can't test atm.

PR #7127, opened 2021-01-09
Add compact items prefix in status bar

JasonInOttawa:
I prefer to have the progress bar next to the status bar with "Show all at once" turned on. I also like the Progress Bar to be as long as possible, to make chapter ticks easier to see.

This works mostly with the Prefix sets to Icons. This removes the icons and extra spaces around the slashes in Current Page and Pages left in book.

It makes the status bar icons that show a simple number (e.g., Pages left in chapter) ambiguous with each other.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7127)


poire-z:
Screenshot**s** (with various other settings, multi lines, etc...) please! so we can see how ugly this gets :)

JasonInOttawa:
This is the option in Settings

Screen Shot 2021-01-09 at 3 08 36 PM

I'm running this on my laptop with the same size window in both cases below.

This is with the option turned off

Screen Shot 2021-01-09 at 3 07 59 PM

This is with the option turned on. Lots of ambiguities :)

Screen Shot 2021-01-09 at 3 08 15 PM

poire-z:
Thanks.
Yes, not super pretty, but could make sense for people with very few items.
Mind moving the setting down in the menu , among the other more related settings, between
`Show footer separators` and `Hide empty items` - and may be rename it to `Compact items` ?

JasonInOttawa:
Actually the wifi commit seems to have broken things :)

JasonInOttawa:
The latest commit removes the "compact items" setting and replaces it with a prefix. 

Here is the menu:

menu

Here is what the bar looks like with "compact letters"

compact letters

Here is what is looks like with letters

letters

Here is what it looks like with icons

icons

poire-z:
With something like this? : 
```diff
@@ -314,8 +314,9 @@ local footerTextGeneratorMap = {
     book_title = function(footer)
         local doc_info = footer.ui.document:getProps()
         if doc_info and doc_info.title then
+            local title = doc_info.title:gsub(" ", "\xC2\xA0") -- replace space with no-break-space
             local title_widget = TextWidget:new{
-                text = doc_info.title,
+                text = title,
```
and if %s strips this no break space, in your stripper, match explicite normal spaces " " instead of "%s".

JasonInOttawa:
@poire-z Thanks very much! Changing spaces to non-break-spaces worked.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -359,6 +360,7 @@ local footerTextGeneratorMap = {
             if add_ellipsis then
                 fitted_chapter_text = fitted_chapter_text .. "…"
             end
+            fitted_chapter_text = fitted_chapter_text:gsub(" ", "\xC2\xA0") -- replace space with no-break-space
```
poire-z, line 363: Well, at least do it similarly as for title :/
(There, you're doing it too late, after the text has been measured and truncated - so, in case in some font a space and a no-break-space have different widths, there might be strange things.)

JasonInOttawa: @poire-z Thank you. Does this commit look better?

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -333,8 +349,9 @@ local footerTextGeneratorMap = {
     book_chapter = function(footer)
         local chapter_title = footer.ui.toc:getTocTitleByPage(footer.pageno)
         if chapter_title and chapter_title ~= "" then
+            local chapter_title_no_break_spaces = chapter_title:gsub(" ", "\xC2\xA0") -- replace space with no-break-space
```
poire-z, line 352: Yes, better, but you can just update this one, no need for a new variable:
```lua
if chapter_title and chapter_title ~= "" then
  chapter_title = chapter_title:gsub(" ", "\xC2\xA0")
```

JasonInOttawa: True enough! I wonder if the interpreter is smart enough to optimise that out ... :)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1729,7 +1762,14 @@ function ReaderFooter:genAllFooterText()
         -- Skip empty generators, so they don't generate bogus separators
         local text = gen(self)
         if text and text ~= "" then
-            table.insert(info, BD.wrap(text))
+            if self.settings.item_prefix == "compact_letters" then
+                -- remove whitespace from footer items if symbol_type is compact_letters
+                -- Issue: also removes white space from book, chapter titles
```
poire-z, line 1767: No more an issue.

poire-z (approved):
Well, another personal tweak added to the footer - but I guess that's the footer's cross to bear :)
Anyway, quite unintruisive (with the gsub, otherwise, we'd have to tweak each of the litlle generators), so I don't mind.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -77,6 +77,21 @@ local symbol_prefix = {
         mem_usage = "",
         wifi_status = "",
         wifi_status_off = "",
+    },
+    compact_letters = {
+        time = nil,
+        pages_left = BD.mirroredUILayout() and "<" or ">",
+        battery = "",
+        -- @translators This is the footer compact letter prefix for the number of bookmarks (bookmark count).
+        bookmark_count = "B",
```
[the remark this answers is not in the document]
JasonInOttawa: Just in case they want to change the letters used for Memory or Bookmarks :)

frontend/apps/reader/modules/readerfooter.lua
JasonInOttawa: Thanks. Took it out

frontend/apps/reader/modules/readerfooter.lua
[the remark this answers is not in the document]
poire-z: I get it - but they won't be able to :)
You need to use one of the `_(` function, like in the other stuff that you cloned:
`bookmark_count = C_("FooterCompactLetterPrefix", "B"),`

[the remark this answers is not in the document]
JasonInOttawa: Thanks again. I don't really know what a C_ function is :) but the new commit duplicates it from the letters. I also used the letters text verbatim (BM and M) but without the colons (they take up so much space :) )

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -83,13 +83,13 @@ local symbol_prefix = {
         pages_left = BD.mirroredUILayout() and "<" or ">",
         battery = "",
         -- @translators This is the footer compact letter prefix for the number of bookmarks (bookmark count).
-        bookmark_count = "B",
+        bookmark_count = C_("FooterLetterPrefix", "BM"),
```
poire-z, line 86: I think it's better to use another "context name" than `FooterLetterPrefix `, to not confuse the translators.
`FooterCompactLetterPrefix`

frontend/apps/reader/modules/readerfooter.lua
Frenzie: @poire-z Pssht. ;-P
> No more an issue.

poire-z:
Given that in the end, you ended up using both letters and icons for prefix, you could rename the option in the menu from Compact Letters  to Compact Items, and just use for item_prefix/symbol_type : "compact" instead of "compact_letters".
We then get smaller values for the thing: letters/icons/compact.

JasonInOttawa:
I went with item_prefix compact_items rather than compact to keep things a bit more descriptive, and perhaps to avoid name clashes later.

Just as an aside, I had to change the following in my settings.reader.lua file:

`["item_prefix"] = "compact_letters",`  

to 

`["item_prefix"] = "compact_items", 
`

or I got this exception when opening a book:

```
[!] doShowReader coroutine crashed:
frontend/apps/reader/modules/readerfooter.lua:195: attempt to index a nil value
stack traceback:
	frontend/apps/reader/modules/readerfooter.lua:1763: in function 'genFooterText'
	frontend/apps/reader/modules/readerfooter.lua:1881: in function 'updateFooterText'
	frontend/apps/reader/modules/readerfooter.lua:1861: in function 'updateFooterPage'
	frontend/apps/reader/modules/readerfooter.lua:1989: in function 'handleEvent'
	frontend/apps/reader/modules/readerrolling.lua:995: in function '_gotoPage'
	frontend/apps/reader/modules/readerrolling.lua:1003: in function '_gotoXPointer'
	frontend/apps/reader/modules/readerrolling.lua:183: in function 'setupXpointer'
	frontend/apps/reader/modules/readerrolling.lua:320: in function 'handleEvent'
	frontend/apps/reader/readerui.lua:430: in function 'init'
	frontend/apps/reader/readerui.lua:578: in function 'doShowReader'
	frontend/apps/reader/readerui.lua:536: in function <frontend/apps/reader/readerui.lua:535>01/11/21-16:09:29 INFO  FM loaded plugin autosuspend at plugins/autosuspend.koplugin

```

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -78,18 +78,18 @@ local symbol_prefix = {
...
     },
-    compact_letters = {
+    compact_items = {
         time = nil,
         pages_left = BD.mirroredUILayout() and "<" or ">",
         battery = "",
-        -- @translators This is the footer compact letter prefix for the number of bookmarks (bookmark count).
-        bookmark_count = C_("FooterCompactLetterPrefix", "BM"),
+        -- @translators This is the footer compact item prefix for the number of bookmarks (bookmark count).
+        bookmark_count = C_("FooterCompactPrefix", "BM"),
```
poire-z, line 86: Ok with compact_items. So, C_("FooterCompactItemsPrefix",...) ?

poire-z:
> I got this exception when opening a book

We might get many of this kind if we go at putting unexpected values into settings :) We don't really pay (nor need to pay) much attention to that. (but it's a bit sad when this kind of exception happens when you downgrade, and the old code can't cope with a setting that changed meaning/format/type.)
So, don't worry.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1729,7 +1762,13 @@ function ReaderFooter:genAllFooterText()
         -- Skip empty generators, so they don't generate bogus separators
         local text = gen(self)
         if text and text ~= "" then
-            table.insert(info, BD.wrap(text))
+            if self.settings.item_prefix == "compact_items" then
+                -- remove whitespace from footer items if symbol_type is compact_items
+                text = text:gsub('%s', '')
+                table.insert(info, BD.wrap(text))
+            else
+                table.insert(info, BD.wrap(text))
```
poire-z, line 1770: Put the `table.insert(info, BD.wrap(text))` outside the if/else as it's the same. And so, no `else `needed.

JasonInOttawa: Should I hold off on any more commits, or keep putting them in with each comment? :)

I notice a lot of builds happening ...

Frenzie: It's better to the environment to bundle them, but they do auto-cancel.

JasonInOttawa: (I was a software designer in a prior life, but it's been 17 years. I'm rusty)

JasonInOttawa: Thanks. I'll hold off committing until there are a few. Please do let me know if I'm holding anything up.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -77,6 +77,21 @@ local symbol_prefix = {
         mem_usage = "",
         wifi_status = "",
         wifi_status_off = "",
+    },
+    compact_items = {
+        time = nil,
+        pages_left = BD.mirroredUILayout() and "<" or ">",
```
Frenzie, line 83: This probably isn't the desired character, but the one we quickly type because we don't know how to type a more suitable UTF-8 symbol.

JasonInOttawa: I was going for something similar to the right/left pointing arrows as in icons. I can replace them with the icon character if preferred.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -77,6 +77,21 @@ local symbol_prefix = {
...
+        battery = "",
+        -- @translators This is the footer compact item prefix for the number of bookmarks (bookmark count).
+        bookmark_count = C_("FooterCompactItemsPrefix", "BM"),
+        percentage = nil,
+        book_time_to_read = nil,
+        chapter_time_to_read = BD.mirroredUILayout() and "«" or "»",
```
poire-z, line 89: It happens you don't need `BD.mirroredUILayout() and ... or ...`, as the chars you chose are mirrorable, and so, mirrored by the FriBidi code.
You can switch the UI language to Arabic to see.
Without BD.mirroredUILayout() , you'll get > and >> mirrored just fine:
image

But there's an issue, I guess with you removing the spaces :( You don't get the expected RTL order for 1/2, which should look like 2/1:
image

Using a zero-width space instead does not seem to solve it :/
`text = text:gsub('%s', '\xE2\x80\x8B') `

frontend/apps/reader/modules/readerfooter.lua
Frenzie: I don't know what you're referring to precisely, but I'm talking about something like the difference between greater than and a chevron (like `⟩`).

frontend/apps/reader/modules/readerfooter.lua
JasonInOttawa: I switched to a RTL language and see this

Screen Shot 2021-01-11 at 5 04 48 PM

The spaces around the slash appear all right.

frontend/apps/reader/modules/readerfooter.lua
poire-z: Yes, no need for icons, there might be more accurate unicode symbols, the classic < > might not be that symetric in some fonts (or dunno, may be they are ? :)

frontend/apps/reader/modules/readerfooter.lua
JasonInOttawa: Sorry, is the issue that the page numbering should also be mirrored (showing as 683/-672 683/11)?

poire-z: The space might be, but that's not the issue:
an arabic/RTL reader expects to read page 11 of 683 as : 683/11.

Other white space that should behave:
https://www.compart.com/en/unicode/bidiclass/WS

May be use a "hair space" :) if that's not too large for you: 
`text = text:gsub("%s", "\xE2\x80\x8A")` (please use double quote " instead of single ')

image
image

poire-z: For differences in space widths: http://jkorpela.fi/chars/spaces.html

Frenzie: Wikipedia isn't horrible either. Just search for a term like "hair space" or "thin space" and you'll end up on https://en.wikipedia.org/wiki/Hair_space#Unicode

frontend/apps/reader/modules/readerfooter.lua
poire-z: https://shapecatcher.com/ draw > or <, and see what looks like it :) 

But I guess what you might want are the buddies of the "«" or "»" we use, but single quote.
U+203A https://www.fileformat.info/info/unicode/char/203a/index.htm

frontend/apps/reader/modules/readerfooter.lua
JasonInOttawa: I'm happy with hair spaces :)

After changing the gsub that strips whitespace only, I'm seeing this now: 683/11

Pages remaining is showing the minus sign like this: 683/671-. Should that be raised as a bug?

poire-z: No, I think `683/671-` might be just what RTL readers expect.
https://util.unicode.org/UnicodeJsps/bidi.jsp?a=-123+%2F+456&p=RTL

JasonInOttawa: I think so too, particularly if you consider the - to be a symbol.

poire-z:
>  I'll hold off committing until there are a few. Please do let me know if I'm holding anything up.

That was all I guess. Go ahead.

JasonInOttawa:
Thanks. The commit is in

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -77,6 +77,23 @@ local symbol_prefix = {
         mem_usage = "",
         wifi_status = "",
         wifi_status_off = "",
+    },
+    compact_items = {
+        time = nil,
+        pages_left = ">",
+        -- pages_left = BD.mirroredUILayout() and "<" or ">",
```
poire-z, line 84: You can remove the BD.mirroredUILayout()  things left ini comments, and just add a comment after:
`pages_left = ">", -- will be auto-mirrored by BiDi`

But have you tried how https://www.fileformat.info/info/unicode/char/203a/index.htm look ?
Just copy/paste this: `›` (vs `>`)

JasonInOttawa: Thanks. The other character looks nice (and saves a bit of space! :) )

JasonInOttawa:
I pushed a commit after this PR was merged and I see the changed code in koreader master. Should I revert that or something? It didn't go through review so far as I know.

I apologise for the error. I didn't realize the PR had been merged at the time of the push.

PR #7128, opened 2021-01-09
DictQuickLookup: allow continuous reading with key - and other fixes

### LINKED ISSUE #7123: KOReader crashes with Wikipedia plugin
* KOReader version: Last one as of today, 2021 January 9.
* Device: Kobo Mini

#### Issue
The Wikipedia plugin crashes just after stablishing connection to the WiFi.

#### Steps to reproduce
Turn off WiFi if it is on. Open a document, select a word in the document to launch dictionary. In the window there, tap on Wikipedia button below. Accept when it asks for permission to turn on WiFi. You will see the message "Obtaining IP address". Some seconds later, KOReader will crash.

#### Aditional information
KOReader doesn't crash in other situations when using Wikipedia. For example, after the crash described above KOReader is connected to the internet, so if you open the same document and tap over the same word and then over the Wikipedia button, the Wikipedia will work.
It won't crash in any case when you have already stablished a WiFi connection.
It won't crash if you are not connected to the Internet but you access to the Wikipedia through the Menu bar in Magnifier icon/Wikipedia lookout. It will ask you to connect to the Internet and then it will proceed without problems.
By the way, I checked too the plugin of translation in the same way: No WiFi, open document, select several words, push on translate, accept turning on WiFi, and it will work too, no crashes.

[crash.log](https://github.com/koreader/koreader/files/5791364/crash.log)

### LINKED ISSUE #7126: PB626: Dictionary word not visibly highlighted after long tap and HW buttons not functional for "switching" between dictionaries
* KOReader version: v2020.12-91-g1c9c40f_2021-01-07
* Device: PB626 sw:U626.5.20.1157

#### Issue
A)
Single word not visibly highlighted after long tap. It is highlighted in background right after dictionary window is showing the translation of selected word. 

B)
HW device buttons are not functional for "switching" between dictionaries. I have installed 3 dictionaries. If the dictionary lookup result is shown, it shows first translation. To the second/third dictionary I can no more use HW buttons, Only SW ones "<<" ">>" are functional. HW buttons are functional only when there is more information from dictionary lookup and the slider is shown. In this case I can move across text up/down also HW buttons.

#### Steps to reproduce
1. Open some e-book. I have pdf document.
2. Make long tap on some word - no word highlight is visible
3. First dictionary with lookup result for selected word is shown - in the background is now visible the highlighted word
4. Press up/down HW device buttons - not possible to move to next/previous dictionary

Before I used koreader nightly v2020.12-30-gb78d9c1_2020-12-23 and this issues were not there present.

Could you please check/fix it?
Thanks

poire-z:
#### DictQuickLookup: allow continuous reading with keys

We could scroll a definition with keys. Now, when reaching top or bottom, switch to prev/next result - like it happens when tap on left/right of definition. Requested in #7126.

Also includes a few fixes:
####  ReaderDictionary: fix possible crash with Wikipedia

Closes #7123 
Also reword some "...canceled" to "...interrupted" for consistency, like done previously for dictionary lookup.

####  Fix selected word non-highlighted on PDF

Optimisation for CRE from #7029 dd74194 made it not work with KOpt-based documents. Closes #7126.

#### Statistics plugin: change some logger.info() to dbg()

See https://github.com/koreader/koreader/pull/6778#discussion_r553957794

####  Margins info msg: show left/right margin values

As we can now tweak them independantly. See https://github.com/koreader/koreader/pull/7104#issuecomment-757338930

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7128)

### DIFF
--- a/frontend/apps/reader/modules/readertypeset.lua
+++ b/frontend/apps/reader/modules/readertypeset.lua
@@ -497,12 +497,13 @@ function ReaderTypeset:onSetPageMargins(margins, refresh_callback)
             text = T(_([[
 Margins set to:
 
-  horizontal: %1 (%2px)
-  top: %3 (%4px)
-  bottom: %5 (%6px)
+  left: %1 (%2px)
+  right: %3 (%4px)
+  top: %5 (%6px)
+  bottom: %7 (%8px)
 
 Tap to dismiss.]]),
-            margins[1], left, margins[2], top, margins[4], bottom),
+            margins[1], left, margins[3], right, margins[2], top, margins[4], bottom),
             dismiss_callback = refresh_callback,
         })
     end
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -914,7 +914,7 @@ function ReaderDictionary:showDict(word, results, box, link)
     self:dismissLookupInfo()
     if results and results[1] then
         UIManager:show(self.dict_window)
-        if not results.lookup_cancelled and ffiUtil.getDuration(self._lookup_start_ts) > self.quick_dismiss_before_delay then
+        if not results.lookup_cancelled and self._lookup_start_ts and ffiUtil.getDuration(self._lookup_start_ts) > self.quick_dismiss_before_delay then
             -- If the search took more than a few seconds to be done, discard
             -- queued and coming up events to avoid a voluntary dismissal
             -- (because the user felt the result would not come) to kill the
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -762,11 +762,13 @@ function ReaderHighlight:onHold(arg, ges)
             local boxes = {}
             table.insert(boxes, self.selected_word.sbox)
             self.view.highlight.temp[self.hold_pos.page] = boxes
+            -- Unfortunately, getWordFromPosition() may not return good coordinates,
+            -- so refresh the whole page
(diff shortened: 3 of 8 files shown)


Frenzie (approved):
lgtm

frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -497,12 +497,13 @@ function ReaderTypeset:onSetPageMargins(margins, refresh_callback)
...
-  top: %3 (%4px)
-  bottom: %5 (%6px)
+  left: %1 (%2px)
+  right: %3 (%4px)
+  top: %5 (%6px)
+  bottom: %7 (%8px)
 
 Tap to dismiss.]]),
-            margins[1], left, margins[2], top, margins[4], bottom),
+            margins[1], left, margins[3], right, margins[2], top, margins[4], bottom),
```
NiLuJe, line 506: Is the table really inverted? (e.g., shouldn't this be `left, margins[2], right, margins[3]` ?)

poire-z: Yes:
https://github.com/koreader/koreader/blob/a44c178d7423575b535e521a863c63398a1b360b/frontend/apps/reader/modules/readertypeset.lua#L92-L93
https://github.com/koreader/koreader/blob/a44c178d7423575b535e521a863c63398a1b360b/frontend/apps/reader/modules/readertypeset.lua#L433-L435

NiLuJe: Or, right, it's L/T/R/B, I read the formatting args in the wrong order ;).

PR #7130, opened 2021-01-10
Align wording of AltStatusBar menu to the bottom one.

zwim:
Use the same wording, where possible (good for translations).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7130)


hius07:
"Top status bar font size (%1)"
"Size of top status bar"

Maybe some unification here as well? Isn't just 'Font size' enough and clear?

poire-z (approved):

PR #7134, opened 2021-01-11
Update readerdictionary.lua

Markismus:
In the function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search):
Add an escape character to a dash to prevent sdcv from looking at it as an option.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7134)

### DIFF
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -718,8 +718,6 @@ end
 function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
     local final_results = {}
     local seen_results = {}
-    -- Add an escape character to a dash to prevent sdcv from looking at it as an option.
-    word = word:gsub('^-', '/-')
     -- Allow for two sdcv calls : one in the classic data/dict, and
     -- another one in data/dict_ext if it exists
     -- We could put in data/dict_ext dictionaries with a great number of words
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -754,7 +754,7 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
             for _, opt in pairs(dict_names) do
                 table.insert(args, "-u")
                 table.insert(args, opt)
-                end
+            end
         end
         table.insert(args, "--")
         table.insert(args, word)
@@ -780,10 +780,8 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
         -- let stdout empty) by appending an "echo":
 
         cmd = cmd .. "; echo"
-        logger.warn(cmd)
         local completed, results_str = Trapper:dismissablePopen(cmd, self.lookup_progress_msg or false)
         lookup_cancelled = not completed
-        logger.warn(results_str)
         if results_str and results_str ~= "\n" then -- \n is when lookup was cancelled
             local ok, results = pcall(JSON.decode, results_str)
             if ok and results then
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -756,7 +756,7 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
                 table.insert(args, opt)
             end
         end
-        table.insert(args, "--")
+        table.insert(args, "--") -- prevent word starting with a "-" to be interpreted as a sdcv option
         table.insert(args, word)
 
         local cmd = util.shell_escape(args)
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -746,7 +746,7 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
             break -- don't do any more lookup on additional dict_dirs
         end
 
-        local args = {"./sdcv", "--utf8-input", "--utf8-output", "--json-output", "--non-interactive", "--data-dir", dict_dir, word}
+        local args = {"./sdcv", "--utf8-input", "--utf8-output", "--json-output", "--non-interactive", "--data-dir", dict_dir}
         if not fuzzy_search then
             table.insert(args, "--exact-search")
         end
@@ -754,8 +754,10 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
             for _, opt in pairs(dict_names) do
                 table.insert(args, "-u")
                 table.insert(args, opt)
-            end
+                end
         end
+        table.insert(args, "--")
+        table.insert(args, word)
 
         local cmd = util.shell_escape(args)
         -- cmd = "sleep 7 ; " .. cmd     -- uncomment to simulate long lookup time
@@ -776,10 +778,12 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
         -- We must ensure we will have some output to be readable (if no
         -- definition found, sdcv will output some message on stderr, and
         -- let stdout empty) by appending an "echo":
+
         cmd = cmd .. "; echo"
+        logger.warn(cmd)
         local completed, results_str = Trapper:dismissablePopen(cmd, self.lookup_progress_msg or false)
         lookup_cancelled = not completed
-
+        logger.warn(results_str)
         if results_str and results_str ~= "\n" then -- \n is when lookup was cancelled
             local ok, results = pcall(JSON.decode, results_str)
             if ok and results then
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -778,10 +778,10 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
         -- We must ensure we will have some output to be readable (if no
         -- definition found, sdcv will output some message on stderr, and
         -- let stdout empty) by appending an "echo":
-
         cmd = cmd .. "; echo"
         local completed, results_str = Trapper:dismissablePopen(cmd, self.lookup_progress_msg or false)
         lookup_cancelled = not completed
+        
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -718,6 +718,8 @@ end
 function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
     local final_results = {}
     local seen_results = {}
+    -- Add an escape character to a dash to prevent sdcv from looking at it as an option.
+    word = word:gsub('^-', '/-')
```
Frenzie, line 722: Forward slash?

Markismus:
I tried four options:  https://github.com/koreader/koreader/issues/6877#issuecomment-757887352

poire-z:
What with the usual way to stop processing options: `--` ?
`./sdcv -blah --blue=blur -O2 -- -άρας`

Frenzie:
`--` is much better.

As an aside, also note that `/-bla/-bla` will less easily fuzzy match `blabla` than `/-bla-bla`. So while it should be `--` and as such what I'm going to say now should be ignored, it should also be `^-` in the pattern. ;-)

poire-z:
Might need to tweak the orderings/insertion points in there:
https://github.com/koreader/koreader/blob/8991ec1372b47672bbb462f3007bfe0a117a3891/frontend/apps/reader/modules/readerdictionary.lua#L749-L758
with:
```
table.insert(args, "--")
table.insert(args, word)
```
at the very end

Markismus:
```lua
local args = {"./sdcv", "--utf8-input", "--utf8-output", "--json-output", "--non-interactive", "--data-dir", dict_dir, "--", word}
```
Wouldn't this work?

poire-z:
> Wouldn't this work?

You tell us :) 
But after this, we might add into args `--exact-search` and the list of dicts to use with `-u`. With --, these wouldn't be processed as options.

Markismus:
Still compiling.... I had to update my koreader fork for the PR. It was somewhere in 2018. I usually just extact the last appimage, had some problem with hard linked compilers way back.

Markismus:
Replacing the "/" with "-- "  doesn't work:
```lua
word = word:gsub('^-', '-- -')
```
It results in:
Screenshot from 2021-01-11 13-36-31

(Nor does replacing it with "---".)

poire-z:
It seems :) It's not about replacing/adding it into `word` - it's about adding it as another arg.
Try this:
```diff
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -746,7 +746,7 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
             break -- don't do any more lookup on additional dict_dirs
         end

-        local args = {"./sdcv", "--utf8-input", "--utf8-output", "--json-output", "--non-interactive", "--data-dir", dict_dir, word}
+        local args = {"./sdcv", "--utf8-input", "--utf8-output", "--json-output", "--non-interactive", "--data-dir", dict_dir}
         if not fuzzy_search then
             table.insert(args, "--exact-search")
         end
@@ -756,6 +756,8 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
                 table.insert(args, opt)
             end
         end
+        table.insert(args, "--") -- prevent word starting with a "-" to be interpreted as a sdcv option
+        table.insert(args, word)

         local cmd = util.shell_escape(args)
         -- cmd = "sleep 7 ; " .. cmd     -- uncomment to simulate long lookup time
```

Markismus:
This works:
```lua
local args = {"./sdcv", "--utf8-input", "--utf8-output", "--json-output", "--non-interactive", "--data-dir", dict_dir}
        if not fuzzy_search then
            table.insert(args, "--exact-search")
        end
        if dict_names then
            for _, opt in pairs(dict_names) do
                table.insert(args, "-u")
                table.insert(args, opt)
                end
        end
        table.insert(args, "--")
        table.insert(args, word)
```

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -776,10 +778,10 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
         -- We must ensure we will have some output to be readable (if no
         -- definition found, sdcv will output some message on stderr, and
         -- let stdout empty) by appending an "echo":
+
         cmd = cmd .. "; echo"
         local completed, results_str = Trapper:dismissablePopen(cmd, self.lookup_progress_msg or false)
         lookup_cancelled = not completed
-
```
poire-z: Please don't move around my valuable blank lines ! :)
(The comment above describe why `cmd = cmd .. "; echo"` - no need for a blank line to separate logical sections.)

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -756,6 +756,8 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
                 table.insert(args, opt)
             end
         end
+        table.insert(args, "--")
```
poire-z, line 759: Please add a comment about that (before, or on the line itself):
`-- prevent word starting with a "-" to be interpreted as a sdcv option`

frontend/apps/reader/modules/readerdictionary.lua
Markismus: Done

frontend/apps/reader/modules/readerdictionary.lua
Markismus: done

poire-z (approved):

NiLuJe:
Does it?

because on the CLI, while it does stop argument parsing, it's *also* passed as a search term...

```
┌─(ROOT@dahlia:pts/0)────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(/mnt/onboard/.adds/koreader)─┐
└─(0.11:12%:17:03:100%:#)── ./sdcv -x -2 ./data/dict -- -foo                                                                                                                                                                                                                                                                                           ──(Mon, Jan 11)─┘
g_mkdir failed: No such file or directory
Found 15 items, similar to --.
```

That's arguably an sdcv bug in my book.

NiLuJe:
Or a glib GOption bug/quirk:

Pre glib parsing:
```
argv[0] = ./sdcv
argv[1] = -x
argv[2] = -2
argv[3] = ./dict
argv[4] = --
argv[5] = -foo
```

Post parsing:
```
argv[0] = ./sdcv
argv[1] = --
argv[2] = -foo
```

That's not how `getopt` behaves (thanks to optind).

NiLuJe:
So, shitty `sdcv` patch:

```diff
diff --git a/src/sdcv.cpp b/src/sdcv.cpp
index ca222b8..b7924b6 100644
--- a/src/sdcv.cpp
+++ b/src/sdcv.cpp
@@ -210,12 +210,17 @@ try {
     lib.load(dicts_dir_list, order_list, disable_list);
 
     std::unique_ptr<IReadLine> io(create_readline_object());
-    if (optind < argc) {
+    if (argc > 1) {
         search_result rval = SEARCH_SUCCESS;
-        for (int i = optind; i < argc; ++i)
+        for (int i = 1; i < argc; ++i) {
+            // Skip the GNU "stop option parsing" token ("--"), because apparently glib's GOption is weird.
+            if (strcmp(argv[i], "--") == 0) {
+                continue;
+            }
             if ((rval = lib.process_phrase(argv[i], *io, non_interactive)) != SEARCH_SUCCESS) {
                 return rval;
             }
+        }
     } else if (!non_interactive) {
 
         std::string phrase;
```

poire-z:
But, lookup from KOReader seem to still work, aren't they?
Normal words work for me, and I assume -words worked for @Markismus too.

NiLuJe:
Possibly because we drop empty results or something?

NiLuJe:
Actual glib way of doing this right:

```diff
diff --git a/src/sdcv.cpp b/src/sdcv.cpp
index ca222b8..3d21156 100644
--- a/src/sdcv.cpp
+++ b/src/sdcv.cpp
@@ -83,6 +83,7 @@ try {
     glib::CharStr opt_data_dir;
     gboolean only_data_dir = FALSE;
     gboolean colorize = FALSE;
+    glib::StrArr word_list;
 
     const GOptionEntry entries[] = {
         { "version", 'v', 0, G_OPTION_ARG_NONE, &show_version,
@@ -109,11 +110,13 @@ try {
           _("only use the dictionaries in data-dir, do not search in user and system directories"), nullptr },
         { "color", 'c', 0, G_OPTION_ARG_NONE, &colorize,
           _("colorize the output"), nullptr },
+        { G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_STRING_ARRAY, get_addr(word_list),
+          _("search terms"), _(" words") },
         {},
     };
 
     glib::Error error;
-    GOptionContext *context = g_option_context_new(_(" words"));
+    GOptionContext *context = g_option_context_new(nullptr);
     g_option_context_set_help_enabled(context, TRUE);
     g_option_context_add_main_entries(context, entries, nullptr);
     const gboolean parse_res = g_option_context_parse(context, &argc, &argv, get_addr(error));
@@ -210,12 +213,14 @@ try {
     lib.load(dicts_dir_list, order_list, disable_list);
 
     std::unique_ptr<IReadLine> io(create_readline_object());
-    if (optind < argc) {
+    if (word_list != nullptr) {
         search_result rval = SEARCH_SUCCESS;
-        for (int i = optind; i < argc; ++i)
-            if ((rval = lib.process_phrase(argv[i], *io, non_interactive)) != SEARCH_SUCCESS) {
+        gchar **p = get_impl(word_list);
+        while (*p) {
+            if ((rval = lib.process_phrase(*p++, *io, non_interactive)) != SEARCH_SUCCESS) {
                 return rval;
             }
+        }
     } else if (!non_interactive) {
 
         std::string phrase;
```

Frenzie:
Upstream PR? ^_^

NiLuJe:
https://github.com/Dushistov/sdcv/pull/68

PR #7135, opened 2021-01-11
ConfigDialog: temporarily hide when showing SpinWidgets

poire-z:
Some follow up to #7104 , reworking how #5000 does its thing, in a hopefully clearer and cleaner way.

#### Dispatcher: increase hold_step for usability

See https://github.com/koreader/koreader/pull/7116#issuecomment-757691669.

#### Notification: closed by any event, but not consuming it

Make Notification have `toast=true`, and UIManager deal specifically with such widget:
a "toast" widget gets closed by any event, and let the event be handled by a lower widget.
This should allow us to not wait or tap to get rid of a notification, and just go on with what we're doing.
Also make them have a default timeout of 2s used with all existing ones.

Because the way it was already supposed to do that (#4013) did not really work, and has side effects, see https://github.com/koreader/koreader/pull/7104#issuecomment-757505651.

This causes a little change in behaviour and some people habits: those who just wait will still wait and see nothing, but those who quickly did a tap to get rid of it will now have that quick tap doing an unwilling action. But I guess they'll get used to it and abandon that now useless tap habit :)

####  CRE: fix to Top/Bottom margins fine tuning

Rework eb66856 so that "Sync T/B Margins" is disabled when set Top and Bottom margins differ (previously, the bottom one was taking over the top one when it was enabled).

####  ConfigDialog: temporarily hide when showing SpinWidgets

Hide bottom panel when showing a SpinWidget, so we gain on the bottom the space covered by the SpinWidget to see more how a change is affecting the book text.
For clarity, rename `delay_repaint` & `refresh_callback` to `hide_on_apply` & `when_applied_callback`.
Normalize SpinWidgets: have them both accept a cancel_callback and a close_callback.
InfoMessage: call dissmiss_callback also on timeout.
Discussed around https://github.com/koreader/koreader/pull/7104#issuecomment-757438538

#### KOPT bottom menu: hide it when applying Zoom options

Use that for Zoom/Pan options (#6885) with PDF, so we see text borders which helps with fine tuning.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7135)


yparitcher (approved):

frontend/ui/uimanager.lua
```diff
@@ -392,7 +392,7 @@ function UIManager:show(widget, refreshtype, refreshregion, x, y, refreshdither)
     for i = #self._window_stack, 0, -1 do
         local top_window = self._window_stack[i]
         -- skip modal window
-        if widget.modal or not top_window or not top_window.widget.modal then
+        if widget.modal or widget.toast or not top_window or not top_window.widget.modal then
```
NiLuJe, line 395: Shouldn't all toasts be modals to begin with?

frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -430,32 +430,43 @@ function ReaderTypeset:makeDefaultStyleSheet(css, text, touchmenu_instance)
...
-    self.ui:handleEvent(Event:new("SetPageMargins", self.unscaled_margins, refresh_callback))
+    self.ui:handleEvent(Event:new("SetPageMargins", self.unscaled_margins, when_applied_callback))
 end
 
-function ReaderTypeset:onSyncPageTopBottomMargins(toggle, refresh_callback)
+function ReaderTypeset:onSetPageTopAndBottomMargin(t_b_margins, when_applied_callback)
+    local t_margin, b_margin = t_b_margins[1], t_b_margins[2]
+    self.unscaled_margins = { self.unscaled_margins[1], t_margin, self.unscaled_margins[3], b_margin }
+    if t_margin ~= b_margin then
+        -- Set Sync T/B Margins toggle to off, as user explicitely made them unsync
```
NiLuJe, line 462: `explicitly made them differ`

frontend/ui/uimanager.lua
```diff
@@ -876,7 +876,15 @@ function UIManager:sendEvent(event)
         end
     end
 
+    -- The top widget get to be the first to get the event
```
NiLuJe, line 879: First get ought to be `gets` ;).

frontend/ui/uimanager.lua
```diff
@@ -876,7 +876,15 @@ function UIManager:sendEvent(event)
...
     local top_widget = self._window_stack[#self._window_stack]
+    if top_widget.widget.toast then
+        -- Notification is our single widget with toast=true
+        -- A toast widget gets closed by any event, and
+        -- let the event be handled by a lower widget
```
NiLuJe, line 884: `lets`

frontend/ui/widget/configdialog.lua
```diff
@@ -1104,20 +1108,31 @@ function ConfigDialog:onConfigMoreChoose(values, name, event, args, name_text, d
             else
                 -- CreDocument event handlers do their own refresh:
                 -- we can just redraw our frame
-                UIManager:setDirty(self, function()
-                    return "ui", self.dialog_frame.dimen
-                end)
+                if self.skip_paint then
+                    -- Redraw anything below the now hidden ConfigDialog
+                    UIManager:setDirty("all", function()
```
NiLuJe, line 1113: Probably ought to be just ReaderUI instead of "all"? (IIRC, the other "all" instances in there are to catch ConfigDialog itself to repaint the panel. Here, it's hidden ;)).

(Unless there's potentially a stack of widgets in the way?).

frontend/ui/uimanager.lua
poire-z: I guess so. But as we don't use/check widget.modal elsewhere it implies that a widget with .toast is handled just as a widget with .modal. So, I guess there's nothing else to do.

frontend/ui/widget/notification.lua
```diff
@@ -19,29 +19,33 @@ local Screen = Device.screen
 local Notification = InputContainer:new{
     face = Font:getFace("x_smallinfofont"),
     text = "Null Message",
-    timeout = nil,
     margin = Size.margin.default,
     padding = Size.padding.default,
+    timeout = 2, -- default to 2 seconds
+    toast = true, -- closed on any event, and let the event to next top widget
```
NiLuJe, line 25: let the event `propagate` to?

frontend/ui/uimanager.lua
```diff
@@ -876,7 +876,15 @@ function UIManager:sendEvent(event)
         end
     end
 
+    -- The top widget get to be the first to get the event
     local top_widget = self._window_stack[#self._window_stack]
+    if top_widget.widget.toast then
```
poire-z, line 881: I guess this should be a while top_widget.widget.toast  - so we dequeue all .toast that could be stacked (but we'll stop at a modal non .toast)

frontend/ui/widget/configdialog.lua
NiLuJe: Especially given what `skip_paint` used to mean: it meant skip painting *ConfigDialog* ;).

frontend/ui/uimanager.lua
NiLuJe: Good point!

(Might happen for real with stuff batched by Dispatcher?)

frontend/ui/widget/configdialog.lua
poire-z: Yes, but there is a SpinWidget above ConfigDialog (which will get its own setDirty onShow).

frontend/ui/uimanager.lua
```diff
@@ -391,8 +391,14 @@ function UIManager:show(widget, refreshtype, refreshregion, x, y, refreshdither)
     -- put this window on top of the toppest non-modal window
     for i = #self._window_stack, 0, -1 do
         local top_window = self._window_stack[i]
-        -- skip modal window
-        if widget.modal or not top_window or not top_window.widget.modal then
+        -- toasts are stacked at top over other toasts,
```
NiLuJe, line 394: `on top` (missed it the first time, my bad :s).

frontend/ui/uimanager.lua
```diff
@@ -391,8 +391,14 @@ function UIManager:show(widget, refreshtype, refreshregion, x, y, refreshdither)
...
+        -- then are modals, and then other widgets
```
NiLuJe, line 395: `then come modals`

poire-z: My consolation is that my coding is better than my english :) 
Please double check my logic on toasts & modals :)

frontend/ui/uimanager.lua
```diff
@@ -876,8 +882,18 @@ function UIManager:sendEvent(event)
...
 
+    -- The top widget gets to be the first to get the event
     local top_widget = self._window_stack[#self._window_stack]
-    -- top level widget has first access to the event
+
+    -- A toast widget gets closed by any event, and
+    -- lets the event be handled by a lower widget
+    -- (Notification is our single widget with toast=true)
+    while top_widget.widget.toast do -- close them all
+        self:close(top_widget.widget)
```
NiLuJe, line 892: That *may* be problematic inside an actual `self._window_stack` walk, as it reorders the table.

UIManager:close() has to take some precautions because it *has* to do just that. Not *quite* sure how much we should care about it, though.

frontend/ui/uimanager.lua
NiLuJe: Looks fine AFAICT ;).

frontend/ui/uimanager.lua
poire-z: Given that here, we don't really walk it: we always take `self._window_stack[#self._window_stack]`, so the one at the top of stack.
When we self:close() it, UIManger:close does walk from #self._window_stack to 1 - and should find it at top (as we found it at top), and remove it from top, and return.
Feels like no issue to me. To you?

poire-z: OK, didn't see your latest edit. So you agree with me :+1:

NiLuJe (approved):

poire-z:
Merging, to have this out and let people see on nightlies how this does.

poire-z:
Some other idea about Notification: knowing with :onShow()/onCloseWidget() how many are shown, the Notification class could store the list of active/shown instances, so a new one could have another y - so they don't override each other, but are stacked above each others.
(Not that I care much about Notification, but seeing 3 of them hiding each other with Dispatcher is a bit ugly :)

PR #7137, opened 2021-01-12
Bump android-luajit-launcher

NiLuJe:
Mainly for HW rotation on notched devices. Depends on https://github.com/koreader/android-luajit-launcher/pull/288

Includes an untested fix/breakage to viewport handling on the small API range where we do that ^^.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7137)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -321,6 +321,7 @@ function Device:toggleFullscreen()
     elseif api < 19 and api >= 17 then
         local width = android.getScreenWidth()
         local height = android.getScreenHeight()
+        -- NOTE: Since we don't do HW rotation here, this should always match width
         local available_width = android.getScreenAvailableWidth()
         local available_height = android.getScreenAvailableHeight()
         local is_fullscreen = android.isFullscreen()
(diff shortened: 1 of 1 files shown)


pazos:
> Includes an untested fix/breakage to viewport handling on the small API range where we do that ^^.

Commented in https://github.com/koreader/android-luajit-launcher/pull/288#pullrequestreview-567114853. available_width will always be width as these devices don't use HW rotation.

The only reason I know for disabling HW rotation on anything lower than 4.4 is that I don't have a device to test that and the only emulator that works in the 4.0-4.3 range is broken when dealing with surface changes.

So, at this point I would consider 4.0-4.3 a bit EOL'd and wouldn't spend a single thought about how to enable HW rota or fix potential bug until it is reported :p

pazos (approved):

pazos:
@Frenzie: depends on https://github.com/koreader/android-luajit-launcher/pull/288

I will bump it later if nobody beats me :)

Frenzie:
https://github.com/koreader/koreader/pull/7142

PR #7147, opened 2021-01-15
Mark calibrecompanion plugin as obsolete

pazos:
To prevent name clashes on menus, when OTA updated. See https://www.mobileread.com/forums/showpost.php?p=4082337&postcount=4

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7147)

### DIFF
--- a/frontend/pluginloader.lua
+++ b/frontend/pluginloader.lua
@@ -6,6 +6,7 @@ local DEFAULT_PLUGIN_PATH = "plugins"
 
 -- plugin names that were removed and are no longer available.
 local OBSOLETE_PLUGINS = {
+    calibrecompanion = true,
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
That reminds me, wasn't there another one?

frontend/pluginloader.lua
```diff
@@ -6,6 +6,7 @@ local DEFAULT_PLUGIN_PATH = "plugins"
 
 -- plugin names that were removed and are no longer available.
 local OBSOLETE_PLUGINS = {
+    calibrecompanion = true,
     storagestat = true,
```
Frenzie, line 10: Never mind, was thinking of this one.

PR #7154, opened 2021-01-16
Address assorted Weblate comments

Frenzie:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7154)

### DIFF
--- a/frontend/ui/elements/screen_android.lua
+++ b/frontend/ui/elements/screen_android.lua
@@ -26,7 +26,7 @@ local timeout_custom7 = 30 * 60 * 1000
 
 local function humanReadableTimeout(timeout)
     local sec = timeout / 1000
-    if sec >= 120 then
+    if sec >= 60 then
         return T(N_("1 minute", "%1 minutes", sec), sec / 60)
     else
         return T(N_("1 second", "%1 seconds", sec), sec)
--- a/frontend/ui/elements/screen_android.lua
+++ b/frontend/ui/elements/screen_android.lua
@@ -5,6 +5,7 @@ local ffi = require("ffi")
 local logger = require("logger")
 local _ = require("gettext")
 local Input = Device.input
+local N_ = _.ngettext
 local Screen = Device.screen
 local T = require("ffi/util").template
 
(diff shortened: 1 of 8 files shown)


poire-z:
> Change KB/MB/GB to KiB/MiB/GiB since that's what it is
> I'd probably prefer kB/MB/GB instead though. I.e., per 1000 instead of 2^10.

So do I. I never ever see MiB, which just feels pedantic.
So please keep them (and their existing translations) and fix the computation :)

NiLuJe (approved):

poire-z (approved):

frontend/ui/elements/screen_android.lua
```diff
@@ -26,9 +27,9 @@ local timeout_custom7 = 30 * 60 * 1000
 local function humanReadableTimeout(timeout)
     local sec = timeout / 1000
     if sec >= 120 then
```
pazos, line 29: `if sec >= 60 then`

plugins/calibre.koplugin/wireless.lua
```diff
@@ -617,7 +618,7 @@ function CalibreWireless:deleteBook(arg)
             if i == 1 then
                 msg = T(_("Deleted file: %1"), BD.filepath(arg.lpaths[1]))
             else
-                msg = T(_("Deleted %1 files in %2:\n %3"),
+                msg = T(N_("Deleted 1 file in %2:\n %3", "Deleted %1 files in %2:\n %3", #arg.lpaths),
```
pazos, line 621: single files are covered by `i == 1`. I guess you can remove that.

plugins/calibre.koplugin/wireless.lua
```diff
@@ -617,7 +618,7 @@ function CalibreWireless:deleteBook(arg)
...
                     #arg.lpaths, BD.filepath(inbox_dir), titles)
```
pazos, line 622: and remove `#arg.lpaths` from here

frontend/ui/elements/screen_android.lua
Frenzie, line 29:
```suggestion
    if sec >= 60 then
```

PR #7159, opened 2021-01-17
calibre plugin: handle huge metadata files

pazos:
Repurposes parts of @WS64 manual parser to read huge json files.

~~Likely broken for json right now, for whatever reason I need to debug. It works fine with the manual parser.~~

rapidjson is 10x faster than manually parsing but manually parsing avoids OOM.

~~Do not review for the moment, as it is broken. Testing  is welcome but there's a 99'9% chance you'll lose faith in humanity.~~

Pinging @NiLuJe, as our chief calibre user: I hardcoded the max JSON file to 30MB. Files bigger than that are going to be parsed manually. I have no idea how many books can be fit in 30MB on calibre metadata. The two files I've tested are mine (200 books, 200kb) and @WS64 (50000 books, 186MB)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7159)

### DIFF
--- a/plugins/calibre.koplugin/metadata.lua
+++ b/plugins/calibre.koplugin/metadata.lua
@@ -8,33 +8,41 @@ of storing it.
 @module koplugin.calibre.metadata
 --]]--
 
+local lfs = require("libs/libkoreader-lfs")
 local rapidjson = require("rapidjson")
 local logger = require("logger")
+local parser = require("parser")
 local util = require("util")
 
-local unused_metadata = {
-    "application_id",
-    "author_link_map",
-    "author_sort",
-    "author_sort_map",
-    "book_producer",
-    "comments",
-    "cover",
-    "db_id",
-    "identifiers",
-    "languages",
-    "pubdate",
-    "publication_type",
-    "publisher",
-    "rating",
-    "rights",
-    "thumbnail",
-    "timestamp",
-    "title_sort",
-    "user_categories",
-    "user_metadata",
-    "_series_sort_",
+local used_metadata = {
+    "uuid",
+    "lpath",
+    "last_modified",
+    "size",
+    "title",
+    "authors",
+    "tags",
+    "series",
+    "series_index"
 }
+
+local function slim(book)
+    local slim_book = {}
+    for k, _ in pairs(used_metadata) do
+        if k == "series" or k == "series_index" then
+            slim_book[k] = book[k] or rapidjson.null
+        elseif k == "tags" then
+            slim_book[k] = book[k] or {}
+        else
+            slim_book[k] = book[k]
+        end
+    end
+    return slim_book
+end
+
+-- this is the max file size we attempt to decode using json. For larger
+-- files we want to attempt to manually parse the file to avoid OOM errors
+local MAX_JSON_FILESIZE = 30 * 1024 * 1024
 
 --- find calibre files for a given dir
 local function findCalibreFiles(dir)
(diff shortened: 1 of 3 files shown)


NiLuJe:
I'll give it a go later, but, yeah, 30MB sounds like a sensible limit ;).

pazos:
Ready for review. I think it lacks a way to check that the cache used for searchs is newer than the metadata for each library used on the search, but that's nothing new.

Other than that everything should work fine. Searchs with books from different libraries parsed with different codecs was tested. I've also added a couple of checks to prevent stripping metadata files when they're used only for searchs.

NiLuJe:
Cool, will give it a try later tonight :).

NiLuJe:
> Ready for review. I think it lacks a way to check that the cache used for searchs is newer than the metadata for each library used on the search, but that's nothing new.

Err, is it? The current code is handling that just fine, as far as I'm concerned ;).

pazos:
> Err, is it? The current code is handling that just fine, as far as I'm concerned ;).

https://github.com/koreader/koreader/blob/master/plugins/calibre.koplugin/search.lua#L580-L593 looks sane, so it is probably it's something fishy on my end, which is probable because I'm running test with @WS64 `calibre.metadata` but without driveinfo (or books :p)

plugins/calibre.koplugin/search.lua
```diff
@@ -32,6 +32,20 @@ local function getDefaultRootDir()
...
+    end
+    if bytes < 1000 then
+        return string.format("%dB", bytes)
+    elseif bytes >= 1000 and bytes < 1000 * 1000 then
+        return string.format("%4.1fKB", bytes/1000)
+    else
+        return string.format("%4.1fMB", bytes/1000/1000)
+    end
+end
+
```
NiLuJe, line 48: We have `util.getFriendlySize` in (front's) `util` ;).

NiLuJe:
No issues with an existing cache, I'll get back to you after I load a few new books later/tomorrow ;).

NiLuJe:
And all's well with new data ;).

```
01/23/21-03:45:10 WARN  cache is older than metadata, ignoring it
01/23/21-03:45:11 INFO  calibre info loaded from disk (search) in 521.914005 milliseconds: 688 books
01/23/21-03:45:11 INFO  metadata: 688 books imported from calibre in 598.746777 milliseconds
```

plugins/calibre.koplugin/metadata.lua
```diff
@@ -8,34 +8,42 @@ of storing it.
...
+    "title",
+    "authors",
+    "tags",
+    "series",
+    "series_index"
 }
 
+local function slim(book)
+    local slim_book = {}
+    for k, _ in pairs(used_metadata) do
```
NiLuJe, line 31: It's an array, you can go with `ipairs` for that infinitesimal speed bump ;p.

pazos: indeed, it's an array and that's a typo from refactor. It should be `ipairs` for sure

NiLuJe: Yeah, noticed the "same" one used later in `parser` is an `ipairs` ;).

pazos: FWIW, it is infinitesimal speed gain for a single book. For 50000 books managed using rapidjson the bump is noticeable (6 vs 8 seconds on my computer)

NiLuJe: (Mildly related: definitely like the logic swap from a blacklist to a whitelist for the field! :+1:)

plugins/calibre.koplugin/metadata.lua
```diff
@@ -42,7 +42,7 @@ end
 
 -- this is the max file size we attempt to decode using json. For larger
 -- files we want to attempt to manually parse the file to avoid OOM errors
-local MAX_JSON_FILESIZE = 30 * 1024 * 1024
+local MAX_JSON_FILESIZE = 30 * 1000 * 1000
```
NiLuJe, line 45: I was quite fine with that one being in mebibytes, but it's not the end of the world, the difference is fairly meaningless in the grand scheme of things ;p.

NiLuJe (approved):
\o/

plugins/calibre.koplugin/metadata.lua
pazos: It is to match OP :)

PR #7161, opened 2021-01-17
Assorted fixes after #7118

NiLuJe:
* I'd failed to notice that ButtonTable *also* instantiates seven billion Buttons on each update. Unfortunately, that one is way trickier to fix properly, so, work around its behavior in Button. (This fixes multiple issues with stuff using ButtonTable, which is basically anything with a persistent set of buttons. A good and easy test-case is the dictionary popup, e.g., the Highlight button changes text, and the next/prev dic buttons change state. All that, and more, was broken ;p).

* Handle corner-cases related to VirtualKeyboard (e.g., Terminal & Text Editor), which screwed with both TouchMenu & Button heuristics because it's weird.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7161)

### DIFF
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -47,6 +47,7 @@ function ReaderSearch:onShowFulltextSearchInput()
                 },
                 {
                     text = backward_text,
+                    vsync = true,
                     callback = function()
                         self:onShowSearchDialog(self.input_dialog:getInputText(), 1)
                         self:closeInputDialog()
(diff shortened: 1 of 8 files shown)


poire-z:
> ButtonTable also instantiates seven billion Buttons on each update. Unfortunately, that one is way trickier to fix properly

I guess it's not really ButtonTable itself, but the way it is used in upper level widgets. But yes, hard to fix without much rework.

All this is a bit over my head to give a proper review, but nothing jumps at me. So, trusting you :) 

(I had been a bit ashamed with my dictionary timers/discardEvent being complex for little reasons, but thanks to your flash_ui stuff, I feel better :)

Frenzie (approved):
Looks sane in principle, but no time to fully comprehend the details in the repainting logic atm.

NiLuJe:
@poire-z: Yeah, I've jotted down a few things to try, but I wanted to have a quick fix ready for the stable first ;).

PR #7162, opened 2021-01-18
Account for result pagination when fetching notes or notebooks from Joplin API

### LINKED ISSUE #7148: Exporting highlights to Joplin fails
* KOReader version: v2020.12
* Device: Kobo Aura H20 Edition 2 v2

#### Issue
Using the Evernote plugin to export notes to Joplin fails with `Multiple errors occurred while trying to export notes from the book: <first word of book filename>`
A KOReader notebook is created in Joplin, but it is empty.

#### Steps to reproduce
Highlight text in a book, use Evernote plugin to try to export book highlights to Joplin
A verbose debug log contains this line:
`DEBUG Error while exporting book <first word of book filename> plugins/evernote.koplugin/JoplinClient.lua:69: attempt to index local 'note' (a boolean value)`

georgelemental:
The Joplin API endpoints for fetching notes or notebooks are paginated now

Resolves #7148

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7162)

### DIFF
--- a/plugins/evernote.koplugin/JoplinClient.lua
+++ b/plugins/evernote.koplugin/JoplinClient.lua
@@ -71,7 +71,7 @@ function JoplinClient:findNoteByTitle(title, notebook_id)
         url = url_base..page
         local notes = self:_makeRequest(url, "GET")
         has_more = notes.has_more
-        for _, note in pairs(notes.items) do
+        for _, note in ipairs(notes.items) do
             if note.title == title then
                 if notebook_id == nil or note.parent_id == notebook_id then
                     return note.id
@@ -97,7 +97,7 @@ function JoplinClient:findNotebookByTitle(title)
         url = url_base..page
         local folders = self:_makeRequest(url, "GET")
         has_more = folders.has_more
-        for _, folder in pairs(folders.items) do
+        for _, folder in ipairs(folders.items) do
             if folder.title== title then
                 return folder.id
             end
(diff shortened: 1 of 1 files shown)


NiLuJe:
Unrelated sidebar: I'm assuming the items field of the json response is laid out as an array, in which case, you might want to use ipairs instead of pairs to preserve order ;).

georgelemental:
@NiLuJe Thanks for the advice! I have no prior experience with Lua

georgelemental:
Is anything preventing this from being merged?

NiLuJe:
Not that I know of, probably forgot to flag it for a milestone and it slipped through the cracks ;).

plugins/evernote.koplugin/JoplinClient.lua
```diff
@@ -61,34 +61,49 @@ end
...
+        url = url_base..page
+        local folders = self:_makeRequest(url, "GET")
+        has_more = folders.has_more
+        for _, folder in ipairs(folders.items) do
+            if folder.title== title then
+                return folder.id
+            end
+        end
+        page = page + 1
+    until(not(has_more))
```
poire-z, line 106: This `until(not(has_more))` (here and above) has too many parens to feel like Lua :)
And where is this `not()` (a function?) coming from?

Frenzie: It's not a function, just another case of excessive (and misleading) parentheses.

Please just write `until not has_more` for clarity. ;-)

Frenzie (approved):
lgtm

PR #7172, opened 2021-01-19
Fix bug raised in issue 7168

JasonInOttawa:
(crash, attempting to set compact items …in status bar [PB631]

Put wrappers around book_time_to_read nil value

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7172)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -280,7 +280,7 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].book_time_to_read
         local left = footer.ui.document:getTotalPagesLeft(footer.pageno)
-        return footer:getDataFromStatistics(prefix .. " ", left)
+        return footer:getDataFromStatistics(prefix and (prefix.." ") or "", left)
     end,
     chapter_time_to_read = function(footer)
         local symbol_type = footer.settings.item_prefix or "icons"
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -280,7 +280,7 @@ local footerTextGeneratorMap = {
         local symbol_type = footer.settings.item_prefix or "icons"
         local prefix = symbol_prefix[symbol_type].book_time_to_read
```
NiLuJe, line 281: Probably simpler to just fix it here?

e.g., `local prefix = blah or ""`

JasonInOttawa: I did try that, and it works for the bug around line 280, but I was trying to keep things consistent with e.g. 

```
    letters = {
        time = nil,

time = function(footer)
...
        if not prefix then
            return clock
        else
            return prefix .. " " .. clock
        end
```

letters.time is set to null and has a wrapper.

Which would be better?

poire-z: Keep it as it is :) (@NiLuJe trick would have had a space added to the prefix, so we would no longer (yay!) get the empty string we want)

poire-z (approved):

frontend/apps/reader/modules/readerfooter.lua
NiLuJe: > @NiLuJe trick would have had a space added to the prefix

Oh, quite right. Oops ;p.

PR #7179, opened 2021-01-21
Refresh AltStatusBar once a minute, if there are changes

zwim:
Make the top bar a bit smarter: Refresh the content if there are changes (time or battery).

Add a menu item to turn the refresh on/off
Reorganize battery items, so that the menu doesn't use two pages.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7179)

### DIFF
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -65,12 +65,13 @@ end
 function ReaderCoptListener:updateHeader()
     if self.view.view_mode == "page" then
         require("logger").dbg("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx updateHeader")
-        self.ui.document._callCacheSet("current_buffer_tag", nil)
-        self.ui:handleEvent(Event:new("RedrawCurrentView", self.current_page))
---        self.ui.rolling:updateBatteryState()
---        UIManager:setDirty(self.view.dialog, function()
---            return "ui", Geom:new{ w = Device.screen:getWidth(), h = self.ui.document:getHeaderHeight()}
---        end)
+--        self.ui.document._callCacheSet("current_buffer_tag", nil)
+--        self.ui:handleEvent(Event:new("RedrawCurrentView", self.current_page))
+        self.ui.rolling:updateBatteryState()
+        self.ui.document:resetBufferCache()
+        UIManager:setDirty(self.view.dialog, function()
+            return "ui", Geom:new{ w = Device.screen:getWidth(), h = self.ui.document:getHeaderHeight()}
+        end)
     end
 end
 
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -64,14 +64,10 @@ end
 
 function ReaderCoptListener:updateHeader()
     if self.view.view_mode == "page" then
-        require("logger").dbg("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx updateHeader")
---        self.ui.document._callCacheSet("current_buffer_tag", nil)
---        self.ui:handleEvent(Event:new("RedrawCurrentView", self.current_page))
         self.ui.rolling:updateBatteryState()
         self.ui.document:resetBufferCache()
-        UIManager:setDirty(self.view.dialog, function()
-            return "ui", Geom:new{ w = Device.screen:getWidth(), h = self.ui.document:getHeaderHeight()}
-        end)
+        UIManager:setDirty(self.view.dialog, "ui",
+            Geom:new{ w = Device.screen:getWidth(), h = self.ui.document:getHeaderHeight()})
     end
 end
 
@@ -84,7 +80,7 @@ function ReaderCoptListener:setupHeaderRefresh()
                 if self.document.configurable.status_line == 0 then -- is top bar enabled
                     local new_battery_level = Device:getPowerDevice():getCapacity()
                     if self.clock == 1 or (self.battery == 1 and new_battery_level ~= self.old_battery_level) then
-                        self:updateHeader(true)
+                        self:updateHeader()
                         self.old_battery_level = new_battery_level
                     end
                 end
@@ -209,7 +205,7 @@ function ReaderCoptListener:getAltStatusBarMenu()
                     local status = _("off")
                     if self.battery == 1 then
                         if self.battery_percent == 1 then
-                            status = _("percent")
+                            status = _("percentage")
                         else
                             status = _("icon")
                         end
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -136,6 +136,7 @@ function ReaderCoptListener:getAltStatusBarMenu()
                 callback = function()
                     self.header_auto_refresh = self.header_auto_refresh == 0 and 1 or 0
                     G_reader_settings:saveSetting("cre_header_auto_refresh", self.header_auto_refresh)
+                    UIManager:broadcastEvent(Event:new("SetStatusLine", self.document.configurable.status_line, true))
                     self:updateHeaderRefreshSchedule()
                 end,
                 separator = true
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -1,6 +1,7 @@
 local Event = require("ui/event")
 local Device = require("device")
 local EventListener = require("ui/widget/eventlistener")
+local Geom = require("ui/geometry")
 local InfoMessage = require("ui/widget/infomessage")
 local UIManager = require("ui/uimanager")
 local T = require("ffi/util").template
@@ -24,7 +25,7 @@ function ReaderCoptListener:onReadSettings(config)
     -- crengine top status bar can only show author and title together
     self.title = G_reader_settings:readSetting("cre_header_title") or 1
     self.clock = G_reader_settings:readSetting("cre_header_clock") or 1
-    self.clock_auto_refresh = G_reader_settings:readSetting("cre_header_clock_auto_refresh") or 1
+    self.header_auto_refresh = G_reader_settings:readSetting("cre_header_auto_refresh") or 1
     self.page_number = G_reader_settings:readSetting("cre_header_page_number") or 1
     self.page_count = G_reader_settings:readSetting("cre_header_page_count") or 1
     self.reading_percent = G_reader_settings:readSetting("cre_header_reading_percent") or 0
@@ -46,8 +47,8 @@ function ReaderCoptListener:onReadSettings(config)
 
     self.old_battery_level = Device:getPowerDevice():getCapacity()
 
-    if self.clock_auto_refresh then
-        self:setupAutoRefreshTime()
+    if self.header_auto_refresh then
+        self:setupHeaderRefresh()
     end
 end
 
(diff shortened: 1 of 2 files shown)


pazos (commented):
It seems a tad complex. Which is the gain?.

zwim:
The gain is, that the time in the status line is correct, especially when you don't touch the reader for more than 1 minute.
Readerfooter does the same (but only with the footer): Status-bar->Settings->Auto refresh time

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -24,6 +24,7 @@ function ReaderCoptListener:onReadSettings(config)
     -- crengine top status bar can only show author and title together
     self.title = G_reader_settings:readSetting("cre_header_title") or 1
     self.clock = G_reader_settings:readSetting("cre_header_clock") or 1
+    self.clock_auto_refresh = G_reader_settings:readSetting("cre_header_clock_auto_refresh") or 1
```
Frenzie, line 27: Does this need a separate setting? There's already an auto-refresh clock setting.

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -54,6 +61,46 @@ function ReaderCoptListener:setAndSave(setting, property, value)
     UIManager:broadcastEvent(Event:new("SetStatusLine", self.document.configurable.status_line, true))
 end
 
+function ReaderCoptListener:UpdateHeader()
+    if self.view.view_mode == "page" then
+        if self.ui.document.provider == "crengine" then -- can that "if" be dropped
+            self.ui.document._callCacheSet("current_buffer_tag", nil)
+        end
+        self.ui:handleEvent(Event:new("RedrawCurrentView", self.current_page))
```
poire-z, line 69: Yes, it can be dropped, ReaderCoptListener being loaded mean crengine is being used.
`self.ui.document._callCacheSet("current_buffer_tag", nil)` is ugly and some internal function.
edec69ac8 added a no-op `resetCallCache()` - but that's not what you want.
You want a similarly set up function that could be called trashBufferCache() and would just have:
`elseif name == "resetCallCache" then add_buffer_trash = true`.

After that, I think you don't even need to have RedrawCurrentView called: simply calling self.ui.rolling:updateBatteryState(), and then a simple setDirty should be enough for UIManager to redraw ReaderUI, asking crengine to draw it again, hopefully with the updated time and battery.
For the setDirty region, we have the height with `CreDocument:getHeaderHeight()`.May be this would work:
```lua
            UIManager:setDirty(self.view.dialog, function()
                return "ui", Geom:new{ w = Screen:getWidth(), h = self.ui.document:getHeaderHeight()},
            end)
```
(I let @NiLuJe tell us what's the right way to do it.)

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -54,6 +61,46 @@ function ReaderCoptListener:setAndSave(setting, property, value)
...
+    end
+end
+
+function ReaderCoptListener:setupAutoRefreshTime()
+    if not self.autoRefreshTime then
+        self.autoRefreshTime = function()
```
poire-z, line 75: Best to rename all that to AutoRefreshHeader - so there's no confusion with setting the refresh time/duration :)

frontend/apps/reader/modules/readercoptlistener.lua
zwim: I think the refresh of the header is different from the footer.
2.) If the clock in the footer is disabled, the refresh cannot be toggled. (Maybe the logic behind that could be improved: First you have to enable "Current time" in bottom bar, then you can toggle the autorefresh in settings). All the other switches in the bottom menu bar are not connected to the top bar, so mixing seems unlogically for me.
2.) The header gets refreshed **every minute**, if the **time** changes **or** (if time is disabled) the **battery** status changes (if battery status is enabled). So you are right -> a better name would be "cre_header_auto_refresh" as it is not only connected to time.

frontend/apps/reader/modules/readercoptlistener.lua
zwim: Yes, of course.

frontend/apps/reader/modules/readercoptlistener.lua
zwim: The header is not refreshed when I change `updateHeader` to:
```
function ReaderCoptListener:updateHeader()
    if self.view.view_mode == "page" then
        require("logger").dbg("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx updateHeader")
--        self.ui.document._callCacheSet("current_buffer_tag", nil)
--        self.ui:handleEvent(Event:new("RedrawCurrentView", self.current_page))
        self.ui.rolling:updateBatteryState()
        UIManager:setDirty(self.view.dialog, function()
            return "ui", Geom:new{ w = Device.screen:getWidth(), h = self.ui.document:getHeaderHeight()}
        end)
    end
end
```

the log shows:
```
01/21/21-21:24:01 DEBUG xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx updateHeader
01/21/21-21:24:01 DEBUG update battery state
01/21/21-21:24:01 DEBUG CreDocument: set battery state 88
01/21/21-21:24:01 DEBUG setDirty via a func from widget ReaderUI
01/21/21-21:24:01 DEBUG painting widget: ReaderUI
01/21/21-21:24:01 DEBUG callCache: drawCurrentViewByPage setting tag
01/21/21-21:24:01 DEBUG callCache: ---------- drawCurrentView: light draw
01/21/21-21:24:01 DEBUG _refresh: Enqueued ui update for region 0 706 540 14 
01/21/21-21:24:01 DEBUG callCache: getHeaderHeight cache miss: getHeaderHeight
01/21/21-21:24:01 DEBUG _refresh: Enqueued ui update for region 0 0 540 31 
01/21/21-21:24:01 DEBUG refresh on physical rectangle 0 706 540 14
01/21/21-21:24:01 DEBUG refresh on physical rectangle 0 0 540 31
```

(I will wait for @NiLuJe for input.)

poire-z: drawCurrentView: light draw = buffer not trashed.
You need to add resetBufferCache into credocument.lua as I outlined above, and call it here
```
self.ui.rolling:updateBatteryState()
self.ui.document:resetBufferCache()
UIManager..setDirty...
```

zwim: I think I got it now. -> I will do that tomorrow.

zwim:
Now with `resetBufferCache`

frontend/document/credocument.lua
```diff
@@ -1512,6 +1517,7 @@ function CreDocument:setupCallCache()
             elseif name:sub(1,6) == "enable" then add_reset = true
             elseif name == "zoomFont" then add_reset = true -- not used by koreader
             elseif name == "resetCallCache" then add_reset = true
+            elseif name == "resetBufferCache" then add_buffer_trash = true
```
poire-z, line 1520: Right, just move it below with the other add_buffer_trash after the findText one.

frontend/apps/reader/modules/readercoptlistener.lua
NiLuJe: Doesn't even need to be a function if you already know the exact dimension of the refresh ;).

(The "pass a function to setDirty" trick is necessary for stuff that uses a widget's internal `dimen` field as the region, as those as (*generally*) only computed during the widget's `paintTo`).

e.g.,

https://github.com/koreader/koreader/blob/9f4e7456cfc105383cead062eb9cb51d7463eab5/frontend/ui/uimanager.lua#L636

zwim:
Now `headerRefresh` is only scheduled if necessary and unscheduled otherwise.

poire-z (commented):
Except for the possible logic error in self.headerRefresh , yes, that's how I wish the launching/rescheduling to behave :)

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -54,6 +60,42 @@ function ReaderCoptListener:setAndSave(setting, property, value)
...
+        UIManager:setDirty(self.view.dialog, "ui",
+            Geom:new{ w = Device.screen:getWidth(), h = self.ui.document:getHeaderHeight()})
+    end
+end
+
+function ReaderCoptListener:setupHeaderRefresh()
+    self.headerRefresh = function()
+        -- Only actually repaint the header only if it's actually visible
+        if self.document.configurable.status_line == 0 then -- is top bar enabled
+            local new_battery_level = Device:getPowerDevice():getCapacity()
```
poire-z, line 76: No more need for `UIManager:getTopWidget() == "ReaderUI" ` ?!
The logic still seems not right.
I guess we want:
- in updateHeaderRefreshSchedule(): what we know at this point and can be toggled from the ReaderCoptListener menu items: if they say stuff needs to be updated: schedule - else, don't.
- in self.headerRefresh, which happens 10-50s later: we should check stuff that may have changed in-between from elsewhere (configurable.status_line , self.view.view_mode == "page" ) and can possible avoid doing it and avoid rescheduling.
- in ReaderCoptListener:updateHeader(), possibly the very local stuff like topwidget!=ReaderUI that doesn't prevent a rescheduling.

Dunno if we should try to catch with a onEvent any change in configurable.status_line , self.view.view_mode so we can unschedule everything if header is off OR view_mode="scroll" - and reschedule when header is on AND view_mode="page".

zwim: @poire-z: I am loosing my joy in contributing with PRs. This is nothing personal. You and the others are polite, helpfull ... professional ...

Please take over this PR and the other one with the bottom bar, if you think they are worthfull enough.

I will stay with KOReader. You will get issues, FRs and so from me, but I won't send no code anymore.

I wish you all the best .
Cheers

poire-z: > I am loosing my joy in contributing with PRs

I'm sorry to hear that.
You actually got quite quickly and well into the schedudling stuff, and you were almost done.
But if I can think of some cases or combinations where it may not be fine, I'm sorry but I just have to mention them - so we can fix them when it's fresh. This is "core" stuff, we don't really want to let it unfinished (unlike "plugins" where we can care a bit less).
And these additional finishing touches is what makes the whole codebase (hopefully) coherent - and make it (hopefully) easier for people like you to get into quickly (unlike crengine).

And yes, it's not always a joy - including for me. Many times have I spent a few days or weeks making a PR, considered it perfect, pushed it - and got reviews about what I would consider nitpicks, that needed a few more hours of not fun work and the need to re-test the whole thing... (and yes, I sometimes did resent them as harassment in my early PRs :/ so, sorry if it feels that way, but that's not my intention - but I know the feeling).
And it's also not always a joy for me to work on or review bits that I don't care much about, like the top status line or the dual page mode :) which delays my fun work (a bit busy with work's work these last days to work on two columns). But I do it so it doesn't get in the way, and I and other people can move on.
I was actually quite happy you made this PR after I mentionned there was no need to do any crengine tweaks, and a setDirty/painto would probably be enough - as I thought that would have been another thing on my not-fun-but-todo list... And happy too that you were appropriating ReaderCoptListener and pushing new ideas in it.

So, well, if you reconsider, good - if not, I'll probably take over this PR as it's really good and there's not much left to do.

hius07:
@zwim, thank you for your efforts

poire-z:
@zwim: any reason that I fail to see for sending the SetStatusLine event on each setting changed ?
https://github.com/koreader/koreader/blob/f4f8820575a96b102c53fb742bb77d25e131efce/frontend/apps/reader/modules/readercoptlistener.lua#L51-L55
(This is not from this PR, but from a previous one.)
Because this sends the event with the current state, so it won't change anything (it will not enable the header if it is currently disabled). So, any other need for it, or do you think we can drop it?

(You had also added it in this PR in the callback for `Auto refresh the alternate status bar`.)

zwim:
Yes, the broadcastEvent is a leftover. Only when changing the font-size of the alt status bar it has to be called to update immediately.

PR #7182, opened 2021-01-22
refesh footer only if there is a change, rename menu item accordingly

zwim:
There are some quirks with the refresh of the bottom bar.
*) The menu entry "Auto refresh time" is misleading. If it is activated the bottom bar is refreshed every full minute (no matter if "Current time" is set or not) -> So I renamed it to "Auto refresh".
*) You cannot change the "Auto refresh time", when "Current time" is disabled, so you cannot change the refresh of the status bar (battery, WiFi change) -> So I have deleted the enabled_func as you can always switch the refresh behavior of the bottom bar.
*) The auto refresh function is updated, so it just refreshes the footer if time, battery, WiFi of memory changes. -> So a lot of  refreshes can be saved when "Current time" is disabled.
*) Renamed the ***refresh*time to ***refresh[footer] to reflect the actual behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7182)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -563,8 +563,8 @@ function ReaderFooter:init()
     else
         self:applyFooterMode()
     end
-    if self.settings.auto_refresh_time then
-        self:setupAutoRefreshTime()
+    if self.settings.auto_refresh then
+        self:setupAutoRefreshFooter()
     end
 
     self.visibility_change = nil
@@ -644,27 +644,30 @@ function ReaderFooter:updateFooterContainer()
     self[1] = self.footer_positioner
 end
 
-function ReaderFooter:setupAutoRefreshTime()
-    if not self.autoRefreshTime then
-        self.autoRefreshTime = function()
+function ReaderFooter:setupAutoRefreshFooter()
+    if not self.autoRefreshFooter then
+        self.autoRefreshFooter = function()
             -- Only actually repaint the footer if nothing's being shown over ReaderUI (#6616)
             if UIManager:getTopWidget() == "ReaderUI" then
                 -- And that only if it's actually visible
                 if self.view.footer_visible then
-                    self:onUpdateFooter(true)
+                    -- Only update, if there could be a change without user activity
+                    if self.settings.time or self.settings.battery or self.settings.wifi_status or self.settings.mem_usage then
+                        self:onUpdateFooter(true)
+                    end
                 end
             else
                 logger.dbg("Skipping ReaderFooter repaint, because ReaderUI is not the top-level widget")
                 -- NOTE: We *do* keep its content up-to-date, though
                 self:onUpdateFooter()
             end
-            UIManager:scheduleIn(61 - tonumber(os.date("%S")), self.autoRefreshTime)
+            UIManager:scheduleIn(61 - tonumber(os.date("%S")), self.autoRefreshFooter)
         end
     end
     self.onCloseDocument = function()
-        UIManager:unschedule(self.autoRefreshTime)
+        UIManager:unschedule(self.autoRefreshFooter)
     end
-    UIManager:scheduleIn(61 - tonumber(os.date("%S")), self.autoRefreshTime)
+    UIManager:scheduleIn(61 - tonumber(os.date("%S")), self.autoRefreshFooter)
 end
 
 function ReaderFooter:setupTouchZones()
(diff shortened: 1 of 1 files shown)


poire-z:
Ideally, for this and your other header PR, the function that is executed every 60 secondes should not even be scheduled if it has nothing to do - or at least not re-schedule itself it it find out there's nothing to do (I see that currently, this test is inside the scheduled function.)

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -962,19 +965,23 @@ function ReaderFooter:addToMainMenu(menu_items)
...
+                    return self.settings.auto_refresh == true
                 end,
-                -- only enable auto refresh when time is shown
-                enabled_func = function() return self.settings.time end,
+                -- only enable auto refresh when time, battery, memory or wifi is shown
+                -- maybe the enabled_func should be dropped, for simplicity
+                --  enabled_func = function()
+                --    return self.settings.time or self.settings.battery or self.settings.wifi_status
+                --        or self.settings.mem_usage
+                -- end,
```
NiLuJe, line 977: I'd keep this ;).

NiLuJe:
Unscheduling it would be a bit tricky, because, say, if one of the update-able items is suddenly disabled, and you were to unschedule it during the scheduled check; if the user re-enables one of those items, nothing currently re-schedules it.

There's already a crapload of stuff refreshing the footer, and the chekc itself make the update silent if necessary, so i'm not sure it's worth making things even gnarlier, unless you had a better idea of how to handle that?

frontend/apps/reader/modules/readerfooter.lua
zwim: OK.

zwim:
> unless you had a better idea of how to handle that?

Sorry no.

poire-z:
We can have a main method to check that and schedule/reschedule (making sure only its only scheduled once) - that would just be called onReadSettings or init - and at end of each callback that may change one of the settings involved.
And the scheculed action would also just call this method after it has done its thing and is about to end.

> There's already a crapload of stuff refreshing the footer

May be, I just don't want uneeded schedule stuff (I tamed BackgroundRunner just for that).

NiLuJe:
Yeah, no better ideas here.

Because I just realized that keeping the `enabled_func` for it kinda raises the same question: if you disable on of the affected options, you can't disable it from the menu again.

So, having something that ensures it's not even scheduled in these cases does make sense.

Not necessarily pretty, but, oh well. I already don't understand the *raison d'être* of this feature to begin with (I mean, the time *may* be off by, what, 4 minutes? So what? It's already off by as much as 59s with this enabled anyway :/).

zwim:
> Because I just realized that keeping the `enabled_func` for it kinda raises the same question: if you disable one of the affected options, you can't disable it from the menu again.

So we drop the enabled_func?

> 
> So, having something that ensures it's not even scheduled in these cases does make sense.
> 

So add a method which checks if all sources of footer update (time, battery, wifi, memory) are off and unshedule the footerUpdate and vice versa if any of the sources is on shedule it (if "Auto refresh") is checked?

> Not necessarily pretty, but, oh well. I already don't understand the _raison d'être_ of this feature to begin with (I mean, the time _may_ be off by, what, 4 minutes? So what? It's already off by as much as 59s with this enabled anyway :/).

No, the time can be off by hours (at least on Android). If you put the device to sleep it shows a sleep-image. If you turn it on after 5 hours it would show the old - not updated - time until you do a manual footer refresh.

zwim:
To avoid doing work/reviews twice, I am waiting with this PR until #7179 is approved. The optimized schedule is implemented in #7179 now.

zwim:
Take it, if it's feasible
It's all saId. I am out.

NiLuJe:
> No, the time can be off by hours (at least on Android). If you put the device to sleep it shows a sleep-image. If you turn it on after 5 hours it would show the old - not updated - time until you do a manual footer refresh.

Which you'll do in the next 4s, give or take? ^^.

I mean, it' snot against you, I've *never* understood the need for this thing.

The footer is just a giant mess to begin with, se we need to be very careful not making it any worse, which means a lot of painful intricate and interwoven crap to deal with.
It's one of the most annoying piece of code to deal with in KO (... with frontlight handling ;p).

zwim:
Jau. I KNOW s not against me. I really appreciate the manners here!!! The work here is a good together The coments are rational...
I admire the patience of you all here. Go on this way!
No but, just a ... it's toooomuich for me. Thats why I retire.
GO ON. I like KOReader. Make it even better

PR #7183, opened 2021-01-22
RFC: Handle arrow keys and enter in emulator text fields

johnbeard:
The emulator thinks it has a DPad, so arrow keys and enter move
around the virtual keyboard rather than moving the cursor and
adding newlines.

This disables the DPad in the emulator to allow these keys to
propagate to the InputText.

Also make sure special keys don't steal modifed keys, and
ensure InputText:OnKeyPress returns false any time
a key is not handled

-----

I get that the DPad thing might be intentional - so may this could have a config or something to toggle DPad in the emulator for people with keyboards?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7183)


poire-z:
> This disables the DPad in the emulator to allow these keys to
propagate to the InputText.

Does this mean the arrow keys won't turn pages or navigate menu?
Can't their handling be different in general vs when a keyboard is opened? Possibly depending if both hasDPAD and hasKeys (or add hasKeyboard if hasKeys means something else) ?

johnbeard:
Hmm, darn, you're right, it does kill off the menu navigation (but page turn still works).
Looks like it steals the event here:
https://github.com/koreader/koreader/blob/8761181cd5c4db2079b372791befc8c156b353a8/frontend/ui/widget/focusmanager.lua#L38-L45

MAybe the check needs to include something like `hasKeyboard() and not virtualKeyboard.isOpen()`?

frontend/device/sdl/device.lua
```diff
@@ -109,6 +109,7 @@ local Desktop = Device:new{
 
 local Emulator = Device:new{
     model = "Emulator",
+    hasDPad = no,
```
Frenzie, line 112: Quoting @poire-z in https://github.com/koreader/koreader/pull/7183#issuecomment-765800280
> Possibly depending if both hasDPAD and hasKeys (or add hasKeyboard if hasKeys means something else) ?

There's also the issue that which behavior's best for the emulator may not be entirely the same as what's best for the AppImage/Desktop. I know I've often said a better emulator is a better desktop app (e.g., being able to drag files straight onto the window, being able to copy/paste, etc. just making it easier to quickly test things), but when we're talking about control basics like typing that's no longer necessarily as true.

Also note that we already have behavior like `local touchless = os.getenv("DISABLE_TOUCH") == "1"` (see line 138), easily triggered with `-t` in the `kodev` script. That doesn't actually disable touch, but perhaps it already covers what you're looking for? But of course I welcome improvements to the desktop app! But these keyboard type changes will require some careful thought to get right. :-)

johnbeard: I get that, which is why it's an RFC. :-D

Basically, I just want the arrow keys to move the cursor and enter to add a new line (basically so I can just type into fields as normal), but not kill off any other arrow/enter usages. I'm unsure of the interplay between FocusManager, the VirtualKeyboard and InputText, the emulator and the app.

poire-z: > Basically, I just want the arrow keys to move the cursor and enter to add a new line 

And backspace to erase !
(Thank you for these keyboard tweaks, I had wished for that for a long time :)

johnbeard: > And backspace to erase !

That should already work (and delete too)?

Frenzie: Btw, I just noticed you can't focus on other input fields where there are multiple. (Select another one, types in the first.)

Screenshot_20210123_194055

poire-z: > That should already work (and delete too)?

You're right.
It is not working for me because of my koreader/settings/event_map.lua containing:
```lua
return {
   [8] = "Back",
   [42] = "Back"
}
```
(If you have an idea on how I can continue having that, but still have it working in InputText, you'll avoid me looking for that :) if not, I'll look for it when I have time :)

Frenzie: I think you'd expect back to… well, go back or close something. Overriding that in the keyboard seems like a breach of contract, but perhaps you could split it out into a separate keyboard back event, but then of course we'd need to be able to make one key to emit both at once.

johnbeard: > Btw, I just noticed you can't focus on other input fields where there are multiple. (Select another one, types in the first.)
This appears to happen on master too?

johnbeard: > I think you'd expect back to… well, go back or close something. Overriding that in the keyboard seems like a breach of contract,

"Back", yes, but overloading "Backspace" to mean "Go back" is also a bit surprising? Even Firefox is scrapping that as a default: https://www.zdnet.com/article/firefox-to-block-backspace-key-from-working-as-back-button/ (but yes, it's a _totally_ different UX in a web browser :-p)

Frenzie: It's not overloaded, @poire-z just has a personal config.

I wasn't talking about this pr with the input focus thing. 😃

poire-z:
Still RFC ? Or ready for merge ?

johnbeard:
I think it's  OK, but the "RFC" kind of stands in that I don't know if it might be breaking stuff on other platforms. Seems OK on my device, but I don't have a DPad or similar.

NiLuJe:
Yeah, it's *probably* going to need an actual check for the emulator platform, instead of the Keyboard cap (because a few actual devices have both the DPad & KeyBoard cap (most prominently, the K3).

NiLuJe:
I'll give it a try on an actual K3 and get back to you on that ;).

NiLuJe:
Not familiar with the code at all, but, yeah, something's wonky on a K3:

Either the InputDialog & VK get closed immediately (Dictionary Lookup), or everything's frozen and a "Back" only closes the VK, not the ID (Find a file).

poire-z:
I'd really like to have arrows keys and enter work on the emulator.
@johnbeard : any more ideas or work on this ? I might pick it if you're no longer interested.

> The emulator thinks it has a DPad, so arrow keys and enter move around the virtual keyboard rather than moving the cursor and adding newlines.

> I get that the DPad thing might be intentional - so may this could have a config or something to toggle DPad in the emulator for people with keyboards?

Yes, I guess we'd like the emulator arrow keys to still do their things, to emulate non-touch on our PCs.
So, it might be a new checkbox in the `Gear> Navigation> [ ] Use arrow keys for text navigation`, may be enabled on AppImage and Desktop but disabled for Emulator ?
And it should only do its things when the virtual keyboard is shown.

johnbeard:
Sure, go for it :). I don't think I can do much more myself anyway since I don't think I have a way to check that I haven't completely trashed, say, dpad devices (also real life keeps, like, totally _happening_).

poire-z:
I've been contemplating for the last hour all the `hasKeys/hasFewKeys/hasDPad/hasKeyboard` stuff, trying to make sense of how they are used in the UI and how they are assigned to devices.
It's quite nightmarish :/
From what I get:
- `hasKeys` = has left/right/enter/back/menu
- `hasFewKeys` = hasKeys, but when back/menu are missing and so left/right should be used instead
- `hasDPad` = hasKeys + up/down/press - focusManager (to navigate menu/dialogs with left/right/up/down) is only enabled with this
- `hasKeyboard` = can type letters

Only SDL and some Kindle have `hasKeyboard` , and when they have it, they all have `hasKeys` and `hasDPad` enabled. These Kindle with `hasKeyboard` are not `isTouchDevice`.
Except Android which has `hasKeys`, never has `hasDPad`, and may have `hasKeyboard` (and may or not have `isTouchDevice`).
Some Kindles with `hasKeyboard`, the one on the right without it (but with hasDPad):
https://upload.wikimedia.org/wikipedia/commons/e/ee/Kindle_2%2C_Kindle_3_%26_Kindle_4.jpg

So, wondering on which conditions to enable the things from this PR (mostly to not change/break/unbreak anything for Kindles NT :):
- add `hasExtendedKeyboard` (or a better name?) to SDL devices, and just enable this for them and we're done. If we need to test focus navigation as hasDPad on the emulator, just comment out hasExtendedKeyboard (like we do comment out isTouchDevice for testing some stuff).
- or use the combinations of `hasKeyboard and isTouchDevice` to enable this PR: assuming that if it isTouchDevice, we can switch fields with touch and don't need arrows to act as DPad to switch fields and select keys on the keyboard.
Kindle non-touch with keyboard need arrows/Enter/Back to act as DPad to navigate fields (and alas, to then navigate into the virtual keyboard and get stuck there ?! the current behaviour of DPad/FocusManager tested on the emulator does not seem super solid, but that's out of my quest :)

Thoughts ? Anything of that makes sense ?

poire-z:
@johnbeard : if I get this right, the arrow keys events were eaten before they reach InputText, by FocusManager.
So, you did override onFocusMove() to prevent their processing by FocusManager, that would happen before they reach InputText.
There's nothing to tweak then (like I initially thought) in your changes to InputText: we can enable all things, they won't trigger on Kindle were FocusManager will have eat them.

If this is right, we can just remove your scattered VirtualKeyboard changes from this PR, and just add this at end of `VirtualKeyboard:init()` (with the conditions we decide on):

```diff
@@ -800,6 +800,20 @@ function VirtualKeyboard:init()
     if keyboard.wrapInputBox then
         self.uwrap_func = keyboard.wrapInputBox(self.inputbox) or self.uwrap_func
     end
+    if Device:hasDPad() and Device:hasKeyboard() and Device:isTouchDevice() then
+        -- hadDPad() would have FocusManager handle arrow keys strokes to navigate
+        -- and activate this VirtualKeyboard's touch keys (needed on non-touch Kindle).
+        -- If we have a keyboard, we'd prefer arrow keys (and Enter, and Del) to be
+        -- handled by InputText to navigate the cursor inside the text box, and to
+        -- add newline and delete chars. And if we are a touch device, we don't
+        -- need focus manager to help us navigate keys and fields.
+        -- So, disable all key_event handled by FocusManager
+        self.key_events.FocusLeft = nil
+        self.key_events.FocusRight = nil
+        self.key_events.FocusUp = nil
+        self.key_events.FocusDown = nil
+        self.key_events.PressKey = nil -- added above
+    end
 end
```
Does this look sane ?
(If yes, just cleanup the conflicting virtualkeyboard.lua and add this in this PR - or I guess I can do it myself in this PR.)

Frenzie:
> add hasExtendedKeyboard (or a better name?) to SDL devices, and just enable this for them and we're done.

In the longer term that may well make the most sense because `hasKeyboard` really just means something like is a Kindle with a keyboard (or the emulator), where the keyboard pretty much means only letters and numbers.

We (mainly me) have added a few full desktop-keyboard centric features to SDL (like Alt+F4) that are purely there unused otherwise.

On the flip side, at least for the short term `hasKeyboard and isTouchDevice` is much simpler and I'm not really sure what else we might even want to add that'd depend on a dedicated `hasFullKeyboard` or whatever you want to call it..

poire-z:
> at least for the short term hasKeyboard and isTouchDevice is much simpler 

Yes, I agree, even `if Device:hasDPad() and Device:hasKeyboard() and Device:isTouchDevice()` as I made it above - now that we can have this condition in a single place, it's easier to keep track and change when we'll need to.

Now, I need to find myself a solution to my own private personal custom event_map.lua:
```lua
return {
   [8] = "Back"
}
```
which gets me to use <kbd>Backspace</kbd> (code 8) to act just like everybody's <kbd>Escape</kbd>: to close current widget, go back, exit KOreader...
Because with this  `[8] = "Back"`, I now can't use Backspace to delete last char in InputText...

I don't want to have to modify this in my dev tree as it would be gathred in git add and would be a mess.
I can go with:
```diff
     if not key["Ctrl"] and not key["Shift"] and not key["Alt"] then
-        if key["Backspace"] then
+        if key["Backspace"] or key["Back"] then
             self:delChar()
         elseif key["Del"] then
             self:rightChar()
```
Before this, your Escape key would close the virtual keyboard, and then do nothing.
With this, your Escape key will close the virtual keyboard, and then act as Backspace.
Would that be an issue - or an unexpected inconsitency/riduculousness you'd be ashamed of ?

Other idea: add "Backspace" with "Back" in every widget that has these kind of stuff:
```
frontend/ui/widget/keyvaluepage.lua:            Close = { {"Back"}, doc = "close page" },
frontend/ui/widget/spinwidget.lua:            Close = { {"Back"}, doc = "close spin widget" }
```
(or make a `Input.group.Back = { "Back", "Backspace" }` and use that instead of these "Back".)
I then would not need to have my hack, we would just all have Backspace work as Back (which could make sense, afterall, that's what I naively expected and why I hack my event_map.lua).
There is no specific handler for Backspace anywhere (except the VirtualKeyboard, which is just fine :)

Thoughts ?

Frenzie:
> (or make a Input.group.Back = { "Back", "Backspace" } and use that instead of these "Back".)

Conflating back and backspace is dangerous because you might accidentally not be in a textbox and then kill your dialog. But I'm on board with the input group since it's such a basic action.

I have wanted to somewhat overhaul all this event map stuff for years, I just never got around to it. I seem to have lost the branch I created some five years ago, but that was just a quick poc. Unfortunately I was significantly more intimate with the relevant code at the time than I am now…

Stupidly I don't seem to have made any notes either (or none that I can find right now), but in this case I'd prefer your `Input.group.Back = { "Back", "Backspace" }` to be an advanced user config. The `event_map.lua` mechanism is very limited.

poire-z:
> I'd prefer your Input.group.Back = { "Back", "Backspace" } to be an advanced user config. 

Guess I can make it not so advanced, and just shove a `[ ] Allow backspace to go back` (wording?) if `hasKeyboard()` into the `Gear > Navigation >` submenu (which already has a bunch of `Back to/in something` items).

Frenzie:
I suppose that would work for this specific case.

```
Backspace works as back button
```
?

poire-z:
Still _This branch cannot be rebased due to conflicts_ ...
I guess letting a PR sit for one year is not a good idea :)
Better start a new one sync on master than risking mess with trying to continue this one.

poire-z:
Refreshed in #8716 (which includes my "Backspace works as back button")

Issue #7129, opened 2021-01-10
Is it possible to enable 'Dual Pages' also in portrait format?


grabi678:
This would be helpful for reading certain publications on e-ink readers with 8" or higher screen size.

poire-z:
Not easily - it's not just a matter of removing this `enabled_func` check:
https://github.com/koreader/koreader/blob/47568ae9b48271ef010e58d1deb583e6aa5821aa/frontend/ui/data/creoptions.lua#L34-L54

crengine (our EPUB engine) will itself prevent going dual page in some conditions:
https://github.com/koreader/koreader/blob/47568ae9b48271ef010e58d1deb583e6aa5821aa/frontend/apps/reader/modules/readerrolling.lua#L1019-L1032

Would need to remove bits in this check:
https://github.com/koreader/crengine/blob/997a9a8fbe4486942997adf05c3516af32512b08/crengine/src/lvdocview.cpp#L3441-L3444
Dunno how this would affect CoolReader thus.

Would look like this:
![image](https://user-images.githubusercontent.com/24273478/104124676-5f866900-5352-11eb-914d-cbdcb5ee6ac5.png)

Not really nice, but might be useful to people with bad progressive glasses with a very narrow reading view and a stuck neck :) 

So, not currently possible, but might be. Waiting for more feedback if we should.

grabi678:
Yes, there are myopic readers who can easily read small fontsizes without glasses. Aside from that devices are getting bigger such as InkPad X . In my opinion this would give the reader a more fexible way to use his device. If it is not too much effort I would appreceate this option could be enabled also for portrait format.

zwim:
Yes, I am severely nearsighted.
I did a quick test with the changes suggested by @poire-z.
On my Tolino4HD (6" 1448x1072) the result is not really usable. But on the Epos2 (8"1440x1920), when using a smaller font-size books are good readable with two columns in portrait. The only thing is, that the font-size has to be chosen so, that there are approximately 40 characters in a line.
I would suspect, that on larger devices

poire-z:
Ok, I can get that working. 2 questions:
- We have one setting for this, so people enabling dual pages in landscape mode will get it enabled too when switching to portrait mode. Is that ok? Real use cases for reading in dual page in landscane, but when switching to portrait, expecting a single page? (I hope not, I don't want to complicate the handling :) 
[...]

PR #7204, opened 2021-01-27
bump crengine, various dual-page and other tweaks

### LINKED ISSUE #5961: Empty TOC lines
* Device: Kindle PW3

#### Issue
Some fb2 books have empty lines after h1 headings, which get into TOC as a separate entry.
Should we suppress them?

Here is an example:
Reader_2020-Mar-15_020442

The empty line here is right after the title, but inside the same section where the title is. Maybe that's why:
Reader_2020-Mar-15_032607

### LINKED ISSUE #7171: Time in alt status bar
* KOReader version: 2020.12-124
* Device: Samsung s10 android 10

#### Issue
The time displayed in the alt status bar is not controlled by the device setting, it is always in 24h format with a leading zero from 1am to 9am.
I'm assuming this is hardcoded in CR as the coolreader app displays it this way as well, and I'm not aware of any setting to change it.
This is basically the same issue as [https://github.com/koreader/koreader/issues/6969](6969), just with CR.
#### Steps to reproduce
Set the time to display in the alt status bar.
N/A

poire-z:
#### CRe TOC, Bookmarks: show marker when jumping

Like we do when following links. Helps a bit on pages crowded with multiple chapters or highlights.

#### TOC: show en-dash on empty titles
Closes #5961.
<kbd>image</kbd>

#### Footer current chapter: use first chapter on page instead of last
Fixes issue B from https://github.com/koreader/koreader/issues/7129#issuecomment-763602579

#### bump crengine: support min/max-width/height, 2-pages tweaks
https://github.com/koreader/koreader-base/pull/1295

Includes:
- EPUB ncx TOC: allow items with an empty title https://github.com/koreader/crengine/pull/409
- resizeImage(): restore original scaling options code
- CSS: parse and store min/max-width/height
- Add getStyledImageSize() ensuring min/max-width/height
- renderBlockElementEnhanced(): ensure min/max-width/height
- 2-pages mode: add option twoVisiblePagesAsOnePageNumber https://github.com/koreader/crengine/pull/410
- TOC, PageMap: save the visible pages number they were build for
- 2-pages mode: tweak header drawing
- Header: add option to display time as 12-hours clock
- 2-pages mode: decrease min middle margin from 1.2 to 0.8em
- Add LVDocView::getDocumentRenderingHash()
- (Upstream) Text: don't remove trailing space on last word if pre https://github.com/koreader/crengine/pull/410
- (Upstream) Font: replace notfound glyphs only on last fallback font
- (Upstream) Add a few lString8 methods

#### CreDocument: disable crengine image scaling options

Since their handling in crengine has been re-enabled in https://github.com/koreader/crengine/pull/409.
We prefer zooming images via `Zoom/Render DPI`.

#### CRe top status bar: ensure 12-hours clock setting

Closes #7171.

#### CRe: use getDocumentRenderingHash() to detect rendering changes

Instead of just relying on document full height and number of pages.

#### Dual pages: shown as 2 columns on a single page

Rework Dual pages code so that the view is considered a single page number, so it looks more like 2-columns on a single page.
This solves a few issues like:
- Page number and count are consistent between top and bottom status bars
- SkimTo -1/+1 doing nothing every other tap
- Statistics being wrong (like "Pages read" never going over half of the book page count)

Discussed in #7129 and in https://github.com/koreader/crengine/pull/408

<kbd>image</kbd>

Haven't yet decided if/how to rename the toggle:
<kbd>image</kbd>

<kbd>Two Columns</kbd> feels accurate in portrait mode, but a bit strange in landscape mode.
Thought a singular <kbd>Dual Page</kbd> could be better (as it's now a single page, but somehow "dual", whatever a person might think it means :)
Thoughts?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7204)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -209,10 +209,10 @@ function ReaderBookmark:onPosUpdate(pos)
     self:setDogearVisibility(self.ui.document:getXPointer())
 end
 
-function ReaderBookmark:gotoBookmark(pn_or_xp)
+function ReaderBookmark:gotoBookmark(pn_or_xp, marker_xp)
     if pn_or_xp then
         local event = self.ui.document.info.has_pages and "GotoPage" or "GotoXPointer"
-        self.ui:handleEvent(Event:new(event, pn_or_xp))
+        self.ui:handleEvent(Event:new(event, pn_or_xp, marker_xp))
     end
 end
 
@@ -304,7 +304,7 @@ function ReaderBookmark:onShowBookmark()
     local bookmark = self
     function bm_menu:onMenuChoice(item)
         bookmark.ui.link:addCurrentLocationToStack()
-        bookmark:gotoBookmark(item.page)
+        bookmark:gotoBookmark(item.page, item.pos0)
     end
 
     function bm_menu:onMenuHold(item)
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -10,16 +10,14 @@ local ReaderCoptListener = EventListener:new{}
 
 function ReaderCoptListener:onReadSettings(config)
     local view_mode = config:readSetting("copt_view_mode") or
-           G_reader_settings:readSetting("copt_view_mode")
-    if view_mode == 0 then
-        self.ui:registerPostReadyCallback(function()
-            self.view:onSetViewMode("page")
-        end)
-    elseif view_mode == 1 then
-        self.ui:registerPostReadyCallback(function()
-            self.view:onSetViewMode("scroll")
-        end)
-    end
+           G_reader_settings:readSetting("copt_view_mode") or 0 -- default to "page" mode
+    local view_mode_name = view_mode == 0 and "page" or "scroll"
+    -- Let crengine know of the view mode before rendering, as it can
+    -- cause a rendering change (2-pages would become 1-page in
+    -- scroll mode).
+    self.ui.document:setViewMode(view_mode_name)
+    -- ReaderView is the holder of the view_mode state
+    self.view.view_mode = view_mode_name
 
     -- crengine top status bar can only show author and title together
     self.title = G_reader_settings:readSetting("cre_header_title") or 1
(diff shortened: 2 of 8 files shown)


hius07:
Very, very nice, thanks a lot.
I'm OK with Two Columns even in landscape mode.

poire-z:
Probably too verbose/technical/users-dont-care, but that's my suggestion :):

<kbd>image</kbd>

I like people to know how this is done, so they can answer their own questions and kill their own crazy ideas before asking :) I don't want to get "can we still get full width images to display spanning the 2 columns?"...
(But I'm sure we'll get "can we have it on 3 columns?" :/ but I won't be bothered :)

(Or should we have it `Columns: one | two` ? althought I like the off|on idea better, has "on" is just using some additional tweaks.)

Rewording suggestions welcome though :)

Frenzie:
Maybe something like:

> Render the document on half the screen width and display two pages at once with a single page number. This makes it look like two columns.

poire-z:
Reworded the 2nd part too:
<kbd>image</kbd>

Frenzie:
Sounds good to me.

zwim:
Sorry, I found a corner case in the top header, which is not so beautiful. 

If the author and/or the title are very long (depending on the font size too), the title is skipped and only the author is shown.
The following solutions come to my mind:
*) Shorten the author and title with ...
*) Show the author and title alternating, but here it could be, that either the author and title is too long and therefor it doesn't get shown.

poire-z:
Does this happens with 1-page mode too ? (= it's an other issue that existed previously, or not?)

zwim:
I have a test.epub. The title is long, the autho is short.
In 1-page mode, the author and title are shown alternating (if both together are too long) and then the title gets additional shortened with an ellipsis if it is too long.

poire-z:
So, you're saying it's ok in 1-page mode?
This should make it behave just like in 1-page mode:
```diff
--- a/crengine/src/lvdocview.cpp
+++ b/crengine/src/lvdocview.cpp
@@ -1920,7 +1920,8 @@ void LVDocView::drawPageHeader(LVDrawBuf * drawbuf, const lvRect & headerRc,
                }
                int w = info.width() - 10;
                if (authorsw + titlew + 10 > w) {
-                       if ((pageIndex & 1))
+                       // Not enough room for both: alternate them
+                       if (getExternalPageNumber(pageIndex) & 1)
                                text = title;
                        else {
                                text = authors;
```

zwim:
Yes, with the above diff, the problem is solved.

PR #7211, opened 2021-01-29
Refresh AltStatusBar once a minute (continued)

poire-z:
#### Top menu: long-press on truncated menu item to show full text

Close #7201 (even if that's a shitty discussion and can't be closed...)

#### Don't toggle the footer when toggling the top status bar

See #7176. Seems people prefer having both bars - and auto-removing the footer when toggling it is a bit bothering when you have footer's "Lock status bar" enabled.
People who really want only the top one can just disable the footer - so they can hide the top without havig the footer auto-shown back.
Feels more prectical to have them independant.

#### Refresh AltStatusBar once a minute, if there are changes + continue work on previous commit

I'll rebase these 2 commits: the first is @zwim latest work on #7179 - the second my continuation.
(Haven't tested resume/suspend.)
@NiLuJe : what do you think about handling auto refresh in the footer the same way: a function called from all involved menu items and events to cancel/reschedule the refresh, with everything ready for auto refresh. The current footer bits that create/nil'ify `self.onCloseDocument` are really ugly.

(I think there is no need for a `UIManager:getTopWidget() == "ReaderUI"` check - if we are reading a dict or wikipedia result, auto-refresh addicts might still want header and footer time auto-updated.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7211)

### DIFF
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -40,8 +40,9 @@ function ReaderCoptListener:onReadSettings(config)
 
     self:onTimeFormatChanged()
 
+    -- Enable or disable crengine header status line (note that for crengine, 0=header enabled, 1=header disabled)
     local status_line = config:readSetting("copt_status_line") or G_reader_settings:readSetting("copt_status_line") or 1
-    self.ui:handleEvent(Event:new("SetStatusLine", status_line, true))
+    self.ui:handleEvent(Event:new("SetStatusLine", status_line))
 end
 
 function ReaderCoptListener:onSetFontSize(font_size)
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -1,8 +1,10 @@
 local Event = require("ui/event")
 local Device = require("device")
 local EventListener = require("ui/widget/eventlistener")
+local Geom = require("ui/geometry")
 local InfoMessage = require("ui/widget/infomessage")
 local UIManager = require("ui/uimanager")
+local logger = require("logger")
 local T = require("ffi/util").template
 local _ = require("gettext")
 
@@ -22,6 +24,7 @@ function ReaderCoptListener:onReadSettings(config)
     -- crengine top status bar can only show author and title together
     self.title = G_reader_settings:readSetting("cre_header_title") or 1
     self.clock = G_reader_settings:readSetting("cre_header_clock") or 1
+    self.header_auto_refresh = G_reader_settings:readSetting("cre_header_auto_refresh") or 1
     self.page_number = G_reader_settings:readSetting("cre_header_page_number") or 1
     self.page_count = G_reader_settings:readSetting("cre_header_page_count") or 1
     self.reading_percent = G_reader_settings:readSetting("cre_header_reading_percent") or 0
@@ -43,6 +46,20 @@ function ReaderCoptListener:onReadSettings(config)
     -- Enable or disable crengine header status line (note that for crengine, 0=header enabled, 1=header disabled)
     local status_line = config:readSetting("copt_status_line") or G_reader_settings:readSetting("copt_status_line") or 1
     self.ui:handleEvent(Event:new("SetStatusLine", status_line))
+
+    self.old_battery_level = Device:getPowerDevice():getCapacity()
+
+    -- Have this ready in case auto-refresh is enabled, now or later
+    self.headerRefresh = function()
+        -- Only draw it if something has changed
+        local new_battery_level = Device:getPowerDevice():getCapacity()
+        if self.clock == 1 or (self.battery == 1 and new_battery_level ~= self.old_battery_level) then
+            self.old_battery_level = new_battery_level
+            self:updateHeader()
+        end
+        self:rescheduleHeaderRefreshIfNeeded() -- schedule (or not) next refresh
+    end
+    self:rescheduleHeaderRefreshIfNeeded() -- schedule (or not) first refresh
 end
 
 function ReaderCoptListener:onSetFontSize(font_size)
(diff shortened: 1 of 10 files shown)


poire-z:
@zwim: please have a quick look/review (can't select you in the reviewers list), if that fits your original plans and I didn't miss cases that you thought of.

Frenzie (approved):
Truncated commit lgtm, the rest too so far.

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -47,7 +48,17 @@ function ReaderCoptListener:onReadSettings(config)
 
     self.old_battery_level = Device:getPowerDevice():getCapacity()
 
-    self:setupHeaderRefresh()
+    -- Have this ready if now or later auto-refresh is enabled
+    self.headerRefresh = function()
+        -- Only draw it if something has changed
+        local new_battery_level = Device:getPowerDevice():getCapacity()
```
zwim, line 54: nit: Comma between "it, if"

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -58,46 +69,65 @@ function ReaderCoptListener:onTimeFormatChanged()
...
 
-function ReaderCoptListener:setAndSave(setting, property, value)
-    self.ui.document._document:setIntProperty(property, value)
-    G_reader_settings:saveSetting(setting, value)
-    UIManager:broadcastEvent(Event:new("SetStatusLine", self.document.configurable.status_line, true))
+function ReaderCoptListener:updateHeader()
+    -- Have crengine display accurate time and battery on its next drawing
+    self.ui.rolling:updateBatteryState()
+    self.ui.document:resetBufferCache() -- be sure next repaint is a redrawing
+    -- Force a repaint (we could avoid it if the top menu is shown as it
```
zwim, line 76: nit: comma "it, if"

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -58,46 +69,65 @@ function ReaderCoptListener:onTimeFormatChanged()
...
-        -- Only actually repaint the header only if it's actually visible
-        if self.document.configurable.status_line == 0 then -- is top bar enabled
-            local new_battery_level = Device:getPowerDevice():getCapacity()
-            if self.clock == 1 or (self.battery == 1 and new_battery_level ~= self.old_battery_level) then
-                self:updateHeader()
-                self.old_battery_level = new_battery_level
-            end
+function ReaderCoptListener:rescheduleHeaderRefreshIfNeeded()
+    if not self.headerRefresh then return end -- not yet set up
+    local unscheduled = UIManager:unschedule(self.headerRefresh) -- unschedule if already scheduled
```
zwim, line 95: Comma

frontend/apps/reader/modules/readercoptlistener.lua
```diff
@@ -58,46 +69,65 @@ function ReaderCoptListener:onTimeFormatChanged()
...
+    -- Only schedule an update if the header is actually visible
```
zwim, line 96: Comma

zwim:
LGTM.
I like the "Don't toggle the footer when toggling the top status bar" and the "Refresh".

frontend/apps/reader/modules/readercoptlistener.lua
NiLuJe: Sounds fine to me without a comma? (We are talking about the comment, right? ^^).

poire-z: Mhhh, that doesn't really sound how my (so-so) english grammar tells me it should :)
A) I will go out if the weather is good
B) I will go out, if the weather is good.

Calling our english grammar scholars @Frenzie and @NiLuJe :)
May be there's some kind of French (usually no comma) or German (comma) habit that leaked in our englishs ?

NiLuJe: I mean, if something needs a rewrite, that's the comment *above* this one, because that one is genuinely not great ^^.

e.g., i'd even simplify it to something along the lines of `We need this for auto-refresh support`.

Frenzie: "Only draw it if something has changed" is good English.

NiLuJe:
> @NiLuJe : what do you think about handling auto refresh in the footer the same way: a function called from all involved menu items and events to cancel/reschedule the refresh, with everything ready for auto refresh. The current footer bits that create/nil'ify `self.onCloseDocument` are really ugly.

Sounds good to me, that was both ugly and problematic in the sense that while there is support for disabling the handlers, nothing would actually ever re-enable 'em (c.f., the discussion linked to the "don't disable footer on alt toggle" thread).

frontend/apps/reader/modules/readercoptlistener.lua
poire-z: Oh, yay, in there, I would put some commas (as I would in French), but not at the places @zwim might put them :)
```
< Have this ready if now or later auto-refresh is enabled
> Have this ready if, now or later, auto-refresh is enabled
```
But ok with the simpler rewrite.

poire-z:
Do we all agree that top status bar and footer toggling should be independant?
Because it looks like I have to nuke some unit tests and do a more thorough hunt for things that may expect to have them tied...

frontend/apps/reader/modules/readercoptlistener.lua
NiLuJe: Yeah, that's a fun comma-related example that's indeed slightly better with a bit of Oxford comma thrown in ;p.

But it's even better if you move that to the end, e.g., `Have this ready in case auto-refresh is enabled, now or later`.

(In case you wan to keep a wordier variant).

NiLuJe:
> Do we all agree that top status bar and footer toggling should be independant?
> Because it looks like I have to nuke some unit tests and do a more thorough hunt for things that may expect to have them tied...

Yeah, I'm perfectly fine with there being no weird interactions between the two ;).

NiLuJe:
@poire-z: Haven't checked the failures in detail, but don't worry about the segfault, that's "normal" as long as readerfooter fails (it's the only one where the final doc:close is inside a test and not a teardown for... reasons).

poire-z:
I don't see any segfault on CircleCI. I also can't reproduce the errors when running each individual tests standalone...

NiLuJe:
Have to run right now, so this might not be relevant, but I think I cheated in one of them to do a sidecar clear to sidestep stuff like that...

poire-z:
Can anybody confirm that on their local emulator this test fail:
```
rm spec/unit/data/*sdr/*
./kodev test front readerlink_spec.lua
```

Frenzie:
I seem to see 2 spec failures which aren't in CircleCI?

```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] ./spec/front/unit/readerlink_spec.lua @ 29: ReaderLink module should jump to links in pdf page mode
[  FAILED  ] ./spec/front/unit/readerlink_spec.lua @ 79: ReaderLink module should be able to go back after link jump in pdf page mode
```

poire-z:
> I don't see any segfault on CircleCI. I also can't reproduce the errors when running each individual tests standalone...

I didn't see any segfault last evening when testing unit tests, until this morning and seeing my system reports :)
```
kernel: luajit[18255]: segfault at 20 ip 00007f3125137f07 sp 00007ffd7da70fd0 error 4 in libmupdf.so[7f31250ff000+150000]
kernel: Code: a0 55 48 89 e5 41 55 41 54 53 48 83 ec 08 49 89 fd 4c 8b 67 70 bb 00 00 00 00 eb 03 83 c3 01 81 fb fc 01 00 00 7f 17 48 63 c3 <49> 8b 74 c4 20 48 85 f6 74 e8 4c 89 ef e8 de fe ff ff eb e9 49 c7
kernel: luajit[21899]: segfault at 0 ip 0000000000000000 sp 00007fffeb9b73f8 error 14 in luajit[563b7fc2a000+8000]
kernel: Code: Bad RIP value.
kernel: luajit[25092]: segfault at 55aa18ea0660 ip 000055aa18ea0660 sp 00007ffdf1129fe8 error 15
kernel: Code: 61 67 65 42 00 b3 76 7f 00 00 a0 c8 c8 18 aa 55 00 00 21 00 00 00 00 00 00 00 01 00 6e 00 49 6d 61 67 65 49 00 b3 76 7f 00 00 <38> 5c 8d 18 aa 55 00 00 11 02 00 00 00 00 00 00 d0 f7 04 19 aa 55
```
Anyway, not minding them.

poire-z:
> I seem to see 2 spec failures which aren't in CircleCI?

I guess you won't see them if you also `rm koreader-emulator-.../koreader/.reader` (the G_reader_settings of the unit tests).
I added in the readerfooter_spec.lua a teardown() that removes some footer settings from it. Seems to solve the numerous issue I had with the next tests (could be we would need more of that, in case tests running in a different order they would run on CI, dunno).

Frenzie:
The order is different based on timings, yes. I haven't seen any failures related to that in quite a while though. I fixed all or the vast majority of that a couple of years ago.

poire-z:
OK, should be good to go.
I rewrote this help text:
image

poire-z:
@NiLuJe : do you think there is a risk, if both header and footer are enabled and both set to auto-refresh, that at the 61th second, they would both setDirty their own areas, which could be merged/coalesced into a single full screen refresh ? :)
If yes, would using this for the header be ok to avoid it, having a few seconds delay between their refreshes?:
```lua
UIManager:scheduleIn(61 - tonumber(os.date("%S")), ... -- footer
UIManager:scheduleIn(63 or 65 - tonumber(os.date("%S")), ... -- header
```

hius07:
> I rewrote this help text:

Isn't it a good idea to move the toggle to the main menu?
https://github.com/koreader/koreader/pull/7107#issuecomment-766265717

poire-z:
> Isn't it a good idea to move the toggle to the main menu?

I don't think so.
- We can toggle the footer by tapping on it - so from outside the menu
- But it would be fine for the footer when the top menu is open, and you see how it looks below
- For the top status bar, showing/hiding it, you would see nothing as it's below the menu where you're doing that.
- So, I think it makes sense to have it toggable from the bottom (and less work)

After all, once you set what item you want in each bar, you're just toggling it (and most would just put their states as defaults and never bother).
(Pfff, I think I already answered that somewhere, probably in Discussions, but Discussions are such a mess...)

NiLuJe:
> @NiLuJe : do you think there is a risk, if both header and footer are enabled and both set to auto-refresh, that at the 61th second, they would both setDirty their own areas, which could be merged/coalesced into a single full screen refresh ? :)
> If yes, would using this for the header be ok to avoid it, having a few seconds delay between their refreshes?:
> 
> ```lua
> UIManager:scheduleIn(61 - tonumber(os.date("%S")), ... -- footer
> UIManager:scheduleIn(63 or 65 - tonumber(os.date("%S")), ... -- header
> ```

The EPDC *probably* wouldn't, but UIManager might. In any case, if it does, it's harmless, it's an UI update, that waveform mode won't touch unmodified pixels.

(i.e., I wouldn't worry about that).

zwim:
Hello, I found a small bug when changing the size of the top bar. My Test-Document is a txt file with 400 lines of text, counting from L001 .. L400


grafik

As you see the screen below the blue line does not get updated, when applying a new size of the top bar.

poire-z:
This should fix it:
```diff
--- a/frontend/apps/reader/modules/readercoptlistener.lua
+++ b/frontend/apps/reader/modules/readercoptlistener.lua
@@ -303,6 +303,8 @@ function ReaderCoptListener:getAltStatusBarMenu()
                             -- we keep open, no need to handle this case.
                             self:setAndSave("cre_header_status_font_size", "crengine.page.header.font.size", spin.value)
                             if touchmenu_instance then touchmenu_instance:updateItems() end
+                            -- Repaint the whole page, as changing this should cause a re-rendering
+                            UIManager:setDirty(self.view.dialog, "ui")
                         end
                     }
                     UIManager:show(size_spinner)
```

zwim:
yes, it's fixed. Thank you.

PR #7212, opened 2021-01-29
Fix a few things after #7166

### LINKED ISSUE #7209: When looking up in the dictionary a crash occurs
* KOReader version: 2021.01.1-9
* Device: PB740, Kobo Aura HD

Loking up in dictionary results in a segmentation fault. Tested on PB740 and Kobo Aura HD.
I am not a developper, but maybe related to #7166?
[crash.log](https://github.com/koreader/koreader/files/5893481/crash.log)

NiLuJe:
* Switching between HTML/text dicts (Fix #7209)
* Updating the scrollbar and scroll state properly when switching dicts
* Highlights in SortWidget
* Highlights in the Dictionary Download page
* Minor simplification of the tail end of the update process in ImageViewer

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7212)

### DIFF
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -799,11 +799,19 @@ function DictQuickLookup:update()
     if self.is_html and self.shw_widget then
         print("Update HTML widget")
         self.text_widget.htmlbox_widget:setContent(self.definition, self:getHtmlDictionaryCss(), Screen:scaleBySize(self.dict_font_size))
+        -- Scroll back to top
+        self.text_widget.htmlbox_widget.page_number = 1
+        self.text_widget.v_scroll_bar:set((self.text_widget.htmlbox_widget.page_number-1) / self.text_widget.htmlbox_widget.page_count, self.text_widget.htmlbox_widget.page_number / self.text_widget.htmlbox_widget.page_count)
+        --self.text_widget:scrollToRatio(0)
     elseif not self.is_html and self.stw_widget then
         print("Update Text widget")
         self.text_widget.text_widget.text = self.definition
         -- NOTE: The recursive free via our WidgetContainer (self[1]) above already free'd us ;)
         self.text_widget.text_widget:init()
+        -- Scroll back to top
+        local low, high = self.text_widget.text_widget:getVisibleHeightRatios()
+        self.text_widget.v_scroll_bar:set(low, high)
+        --self.text_widget:scrollToRatio(0)
     else
         -- We jumped from HTML to Text of vice-versa, we need a new widget instance
         -- Whee, code duplication! (this is copied verbatim from init)
@@ -821,6 +829,7 @@ function DictQuickLookup:update()
                 end,
             }
             self.text_widget:clear()
+            -- Update *all* the references to self.text_widget
             self.text_widget = self.shw_widget
             self.definition_widget[1] = self.text_widget
             self.stw_widget = nil
@@ -840,6 +849,7 @@ function DictQuickLookup:update()
                 images = self.images,
             }
             self.text_widget:clear()
+            -- Update *all* the references to self.text_widget
             self.text_widget = self.stw_widget
             self.definition_widget[1] = self.text_widget
             self.shw_widget = nil
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -799,19 +799,13 @@ function DictQuickLookup:update()
     if self.is_html and self.shw_widget then
         self.text_widget.htmlbox_widget:setContent(self.definition, self:getHtmlDictionaryCss(), Screen:scaleBySize(self.dict_font_size))
         -- Scroll back to top
-        self.text_widget.htmlbox_widget.page_number = 1
-        self.text_widget.v_scroll_bar:set((self.text_widget.htmlbox_widget.page_number-1) / self.text_widget.htmlbox_widget.page_count, self.text_widget.htmlbox_widget.page_number / self.text_widget.htmlbox_widget.page_count)
-        self.text_widget.v_scroll_bar.enable = self.text_widget.htmlbox_widget.page_count > 1
+        self.text_widget:resetScroll()
     elseif not self.is_html and self.stw_widget then
         self.text_widget.text_widget.text = self.definition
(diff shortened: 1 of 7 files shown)


NiLuJe:
(Left as draft because I haven't look at the visual glitches yet, and I may not have time to do so tonight. The rest has only been minimally tested, but seems to behave ;)).

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -794,12 +796,60 @@ function DictQuickLookup:update()
...
+        local low, high = self.text_widget.text_widget:getVisibleHeightRatios()
+        self.text_widget.v_scroll_bar:set(low, high)
+        local visible_line_count = self.text_widget:getVisLineCount()
+        local total_line_count = self.text_widget:getAllLineCount()
+        self.text_widget.v_scroll_bar.enable = visible_line_count < total_line_count
+    else
+        -- We jumped from HTML to Text of vice-versa, we need a new widget instance
+        -- Whee, code duplication! (this is copied verbatim from init)
+        if self.is_html then
+            self.shw_widget = ScrollHtmlWidget:new{
```
poire-z, line 819: I don't like this code duplication, for having to tweaks the params in both places and we might easily forget the other, and having to duplicate comments, etc...
Can't we create both Scroll*Widget empty (or ready enough) in init, and re-use/swap them as needed ?

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -794,12 +796,60 @@ function DictQuickLookup:update()
...
 
     -- Update main text widgets
-    if self.is_html then
+    if self.is_html and self.shw_widget then
         self.text_widget.htmlbox_widget:setContent(self.definition, self:getHtmlDictionaryCss(), Screen:scaleBySize(self.dict_font_size))
-    else
+        -- Scroll back to top
+        self.text_widget.htmlbox_widget.page_number = 1
+        self.text_widget.v_scroll_bar:set((self.text_widget.htmlbox_widget.page_number-1) / self.text_widget.htmlbox_widget.page_count, self.text_widget.htmlbox_widget.page_number / self.text_widget.htmlbox_widget.page_count)
+        self.text_widget.v_scroll_bar.enable = self.text_widget.htmlbox_widget.page_count > 1
```
poire-z, line 804: This hacking of the internal (yes, we might already do it elsewhere...) could probably be made a method of both Scroll*Widget:scrollToTop().

frontend/ui/widget/dictquicklookup.lua
NiLuJe: That would probably work. I tried without doing that because, arguably, those are "heavy" widgets, but it's probably moot in the grand scheme of things ;).

poire-z: Thought we could even push that further and have these Scroll*Widget instances stored not in `self`, but in the single DictQuickLookup class object, so the 2 initial instances would just be re-used by all further lookups - and may be this would fix the memory usage/leak issue I suspect happens with dicts.

But... we can have stacked DictQuickLookups when re-looking up inside a dict :( So, no go.

NiLuJe:
@pazos: I assume the fact that DictQuickLookup is not flagged `modal` is on purpose?

(Harmless query related to highlights under it in the History lookup KVP).

poire-z:
> DictQuickLookup is not flagged modal is on purpose 

No idea. I don't think I ever wondered if a thing should have modal=true or not. So, its modal'ity has probably not changed since I started babysitting it.

NiLuJe:
Hid the code duplication away inside a function, until I see how feasible the "init and reuse both types of ScrollWidgets" is in practice ;).

NiLuJe:
> Hid the code duplication away inside a function, until I see how feasible the "init and reuse both types of ScrollWidgets" approach is in practice ;).

Well, they both have to instantiate their *BoxWidget at init, so there's no really bypassing that, which means two real, full instances.
It would also introduce the same set of (minimal) gymnastics on CloseWidget to free both.

TL;DR: I feel happier about the current compromise... Thoughts, @poire-z?

poire-z:
OK, so we're not reusing Scroll*Widget when we cross back text>html>text - and we reinstantiate it fully as before your init/update optimisations.
But we keep it when text>text>text or html>html>html - but we do :free() anyway in all cases. Is there still a benefit keeping and reusing the freed and naked widget?

I can be fine with what you got. But it still feels a bit hybrid and complicated :)
> they both have to instantiate their *BoxWidget at init, so there's no really bypassing that

Dropping and reinstantiating fuly would feel clearer (killing a bit on your ambitions from #7166).

NiLuJe:
> OK, so we're not reusing Scroll*Widget when we cross back text>html>text - and we reinstantiate it fully as before your init/update optimisations.

Correct.

> But we keep it when text>text>text or html>html>html - but we do :free() anyway in all cases. Is there still a benefit keeping and reusing the freed and naked widget?
> 

Yes: we have to free it to change its content, even when we re-use it ;). So keeping it essentially comes for... free ;p.

poire-z:
> it's two branches and six LoCs, so, yeah, not a great burden 

Yes, but dropping and reinstantiating inconditionally is 0 branch and 1 LoC :)

> So keeping it essentially then comes for... free

Given that we don't keep much, what is the benefit of keeping these skeleton, even for free? :)
Does it matter in the refresh/keeping the reference of #7166 business ? I'm not sure I see how - and dropping it every other html/text transition but not on text/text introduce a change of behaviour that could complicate testing/reproducing bugs.

I'm still fine how it is, even if over-engineered :) just giving you some thoughts :)

poire-z (approved):
No other comment (not tested though)

NiLuJe:
> Does it matter in the refresh/keeping the reference of #7166 business ? I'm not sure I see how - and dropping it every other html/text transition but not on text/text introduce a change of behaviour that could complicate testing/reproducing bugs.

Nope, it has no impact on the ButtonTable stuff, it just happened than in the process of keeping the ButtonTable in those widgets, it made no sense to only try to care about that single thing when rewiring the whole thing to keep as much stuff as possible felt cleaner and easier, since the re-organizing necessary for the one made the rest "easy".

Plus, I found rethinking how this stuff works beneficial, and I would actually *welcome* it unearthing design issues or actual bugs in the underlying widgets ;).

> I'm still fine how it is, even if over-engineered :) just giving you some thoughts :)

over-engineered is my middle name ;p.

PR #7213, opened 2021-01-31
Tweak footer auto refresh

poire-z:
Similar to how it's been done recently for CRe top status bar in #7211.
Continued from #7182. See https://github.com/koreader/koreader/pull/7211#issuecomment-770003153.

@NiLuJe : please have a look if that looks fine (at least, not worse than before)
It's a bit nightmarish with the self.mode vs self.view.footer_visible possibly getting out of sync when toggle from other modules...
So, there are probably some cases badly handled - but hopefully, it's not worse than before :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7213)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -563,9 +563,6 @@ function ReaderFooter:init()
     else
         self:applyFooterMode()
     end
-    if self.settings.auto_refresh_time then
-        self:setupAutoRefreshTime()
-    end
 
     self.visibility_change = nil
 end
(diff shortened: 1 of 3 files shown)


frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -644,27 +641,57 @@ function ReaderFooter:updateFooterContainer()
...
+end
+
+function ReaderFooter:rescheduleFooterAutoRefreshIfNeeded()
+    if not self.autoRefreshFooter then
+        -- Create this function the first time we're called
+        self.autoRefreshFooter = function()
             -- Only actually repaint the footer if nothing's being shown over ReaderUI (#6616)
+            -- (We want to avoid the footer to be painted over a widget covering it - we would
+            -- be fine refreshing it if the widget is not covering it, but this is hard to
+            -- guess from here.)
```
NiLuJe, line 657: FWIW, this wouldn't necessarily be all that great, so all the more reason not to be too bummed about it ;).

As a rather edge-case-y example, the Bookmark code used to triggering footer refreshes on highlight actions: if the footer was invisible, but the highlight in the footer region, it would catch the ReaderUI repaint that updates the highlights, and possibly erase/half-erase/overlay the highlight ;).

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -2134,19 +2159,6 @@ function ReaderFooter:onHoldFooter()
...
-        -- duplicate onTapFooter() code)
-        if self.mode == self.mode_list.off then
-            self:onTapFooter(true) -- ignore tap lock
-        end
-        self.view.footer_visible = (self.mode ~= self.mode_list.off)
-    else
-        self:applyFooterMode(self.mode_list.off)
-    end
-end
-
```
NiLuJe, line 2149: For people following at home, that's related to the removal of the "hide the footer when the alt bar is enabled" code, which went bye bye in a previous PR ;).

NiLuJe:
Looks good, you kinda went the extra mile there ;).

At a glance, I don't think I can see anything going wrong, you're quite right in actually trusting the mode for the scheduling, and the visibility flag for whether the update refreshes or not, that should be solid.

NiLuJe:
I'd possibly fold the `rescheduleFooterAutoRefreshIfNeeded()` calls inside `applyFooterMode` directly, might limit the game of whack-a-mole and make the thing slightly more future-proof?

NiLuJe:
Ping me if you want me to take care of the test, at a quick glance, I don't think anything touches the really annoying parts of the readerfooter tests, but I know it can be extremely irritating to deal with ;).

poire-z:
> I'd possibly fold the rescheduleFooterAutoRefreshIfNeeded calls inside applyFooterMode directly, might limit the game of whack-a-mole and make the thing slightly more future-proof?

I didn't want rescheduleFooterAutoRefreshIfNeeded called in init(), and there's a applyFooterMode also in init, so we'd need a check we are past :init.
There are other places outside of applyFooterMode that needs it, and doing it _also_ in applyFooterMode would have it called sometimes twice.
So yes, either we go the dirty fullproof way by putting it in applyFooterMode ; or we try to see where it's needed like I did, with the risk of missing some spots :/
(I'd go with the latter because I don't use the feature, and I don't want all that many rescheduleFooterAutoRefreshIfNeeded  :)

poire-z:
> Ping me if you want me to take care of the test, at a quick glance, I don't think anything touches the really annoying parts of the readerfooter tests, but I know it can be extremely irritating to deal with ;).

Thanks ! :) but it wasn't that hard to fix them, so done.

(Put back the setting original name `auto_refresh_time` to avoid having some code to migrate it to the would-have-been-a-better-name  `auto_refresh`.)

PR #7214, opened 2021-01-31
Better image fit in Cover-Image plugin

zwim:
If the aspect ratios of the cover image and the screen are similar, the image should get stretched (instead of being scaled) to fill the screen. Stretching looks better than scaling, if the background is only a small stripe. 

See
Untitled

The menu has to be change also to reflect the new functionality:
grafik

As usual, improvements in the wording are appreciated.
How can I get rid of https://github.com/koreader/koreader/commit/c31278cee1c249ccb29aa372defc3004b2ec4f5e? Does a simple `git revert` will do the job?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7214)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -48,6 +48,7 @@ function CoverImage:init()
     self.cover_image_format = G_reader_settings:readSetting("cover_image_format") or "auto"
     self.cover_image_extension = getExtension(self.cover_image_path)
     self.cover_image_quality = G_reader_settings:readSetting("cover_image_quality") or 75
+    self.cover_image_stretch_limit = G_reader_settings:readSetting("cover_image_stretch_limit") or 10
     self.cover_image_background = G_reader_settings:readSetting("cover_image_background") or "black"
     self.cover_image_fallback_path = G_reader_settings:readSetting("cover_image_fallback_path") or "cover_fallback.png"
     self.enabled = G_reader_settings:isTrue("cover_image_enabled")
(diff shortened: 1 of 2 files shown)


poire-z:
> How can I get rid of c31278c? Does a simple git revert will do the job?

`git revert` will add a 3rd commit.
To get rid of one:
`git rebase -i master`
you see a list of the 2 commits in your editor
replace "pick" with "drop" on the first commit, save & quit your editor
and git log shouldn't show it anymore.

I let somebody else review this, because I'm extremely tired of width and height and aspect ratios :)

NiLuJe:
On mobile right now, bit looks sane at a glance, more on that later ;).

I'm not necessarily a fan of the concept myself, but, it's configurable, and a plug-in, so, fine by me ;). (I would lobby for a default at disabled if this were in core, FWIW).

NiLuJe:
I would probably slightly tweak the logic to make it (possibly) more readable:

* Compute the ratio as `iAR / sAR * 100` (always)
* Switch the check to stretch if `ratio >= 100-limit or ratio <= 100+limit`  (e.g., assuming the defaults: if 90 <= ratio <= 110).

plugins/coverimage.koplugin/main.lua
```diff
@@ -259,7 +272,41 @@ function CoverImage:addToMainMenu(menu_items)
                 end,
                 sub_item_table = {
                     {
-                        text = _("Scale, black background"),
+                        text_func = function()
+                            return T(_("Elastic stretch limit (%1%)"), self.cover_image_stretch_limit )
+                        end,
+                        help_text_func = function()
+                            return T(_("If the aspect ratios of the image and the screen differ less than the selected limit (%1%), the image gets stretched (instead of scaled) to fullscreen."), self.cover_image_stretch_limit )
```
NiLuJe, line 279: `If the image and the screen have a similar aspect ratio (more or less %1%), stretch the image instead of honoring its aspect ratio.`

plugins/coverimage.koplugin/main.lua
```diff
@@ -259,7 +272,41 @@ function CoverImage:addToMainMenu(menu_items)
                 end,
                 sub_item_table = {
                     {
-                        text = _("Scale, black background"),
+                        text_func = function()
+                            return T(_("Elastic stretch limit (%1%)"), self.cover_image_stretch_limit )
```
NiLuJe, line 276: `Aspect ratio stretch threshold`, maybe?

plugins/coverimage.koplugin/main.lua
```diff
@@ -259,7 +272,41 @@ function CoverImage:addToMainMenu(menu_items)
...
+                        keep_menu_open = true,
+                        callback = function(touchmenu_instance)
+                            local old_stretch_limit = self.cover_image_stretch_limit
+                            local SpinWidget = require("ui/widget/spinwidget")
+                            local size_spinner = SpinWidget:new{
+                                width = math.floor(Device.screen:getWidth() * 0.6),
+                                value = old_stretch_limit,
+                                value_min = 0,
+                                value_max = 25,
+                                default_value = 95,
```
NiLuJe, line 290: That doesn't make sense, given the default at 10 above ;).

NiLuJe: (Talking about the default, specifically ^^. The min is fine, and I'd fine with the max).

zwim:
> Barring that, at the very least moving the `* 100` bit outside of the actual test, possibly with a better variable name to indicate that it's now a percentage.

Shouldn't we go even further to :
`   local ratio_divergence_percent = 100 - image_ratio / screen_ratio * 100` as this is what we enter in the menu
`if self.cover_image_stretch_limit > ratio_divergence_percent and ratio_divergence_percent > -self.cover_image_stretch_limit then`

plugins/coverimage.koplugin/main.lua
zwim: Should we use +/- or the cooler ± ?

NiLuJe:
> > Barring that, at the very least moving the `* 100` bit outside of the actual test, possibly with a better variable name to indicate that it's now a percentage.
> 
> Shouldn't we go even further to :
> ` local ratio_divergence_percent = 100 - image_ratio / screen_ratio * 100` as this is what we enter in the spin-widget
> `if self.cover_image_stretch_limit > ratio_divergence_percent and ratio_divergence_percent > -self.cover_image_stretch_limit then`

Yeah, that's the spirit ;).

Taking it futher:

```
local ratio_divergence_percent = math.abs(100 - ((image_ratio / screen_ratio) * 100))
if self.cover_image_stretch_limit > ratio_divergence_percent then
```

plugins/coverimage.koplugin/main.lua
NiLuJe: The cooler one, I had intended to lookup the codepoint and... then I forgot :D.

zwim:
Touché

zwim:
While I was cleaning my old PRs and commits I have deleted my zwim/koreader fork. So this PR is orphaned now:
`zwim wants to merge 2 commits into koreader:master from unknown repository`
I have the sources and the repository local. Meanwhile I have forked zwim/koreader again. What should I do? Is there a possibility to append the new zwim/koreader to this PR or should I do a new PR and close this one?

Frenzie:
It's possible that if you recreate the betterImageFit branch it'll reattach itself to this PR, but I wouldn't count on it. I'd suggest trying to see what happens? If it doesn't you can open a new one.

poire-z:
Dunno, if all is fine on your local tree, try just:
git push -f origin your_branch_with_the_name_it_has_originally
or, if it complains it can't find it, create a new one:
git push -u origin your_branch_with_the_name_it_has_originally

zwim:
@poire-z: Did not work as expected. I have opened a new PR and close this one. Sorry.

PR #7220, opened 2021-01-31
cervantes4: get rid of autowarmth 'feature'

pazos:
Now uses a workaround that I've added for Pocketbooks, to avoid touching the fl widget. If @NiLuJe or somebody else does the same in Kobos I would gladly remove all the junk from fl widget in a companion commit.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7220)

### DIFF
--- a/frontend/device/cervantes/device.lua
+++ b/frontend/device/cervantes/device.lua
@@ -56,6 +56,9 @@ local Cervantes = Generic:new{
     -- all devices, except the original Cervantes Touch, have frontlight
     hasFrontlight = yes,
 
+    -- REMOVE ME: workaround until the frontlight widget is sane enough
+    hasNaturalLightApi = yes,
+
     -- currently only Cervantes 4 has coloured frontlight
     hasNaturalLight = no,
     hasNaturalLightMixer = no,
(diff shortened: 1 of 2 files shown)


NiLuJe:
We unfortunately have a fair number of users for this mess on Kobo :/.

pazos:
> We unfortunately have a fair number of users for this mess on Kobo :/.

And saying: use a gesture isn't an option? :smile: 

Quite frankly I'm in the same position with android custom screen timeouts. The code is ugly and a mess. So much mess that I won't touch it. Which leaves me with the question: is that important? I mean: the only way that I have to change the code is removing it, see how much people is affected, and implement something saner in the future.

I wonder if that's the case here. But no worries if it isn't.

Frenzie:
Wasn't the autowarmth a plugin or am I thinking of something else?

NiLuJe:
There's an autofrontlight stuff which leverages the ALS on the Kindle Voyage & Oasis (1), IIRC.

I don't think anyone ever bothered putting an ambient light sensor on any device since ^^.

pazos:
duh

PR #7221, opened 2021-01-31
Better image fit (continued)

zwim:
Eat this, github.

This continues the work from #7214

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7221)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -102,11 +102,12 @@ function CoverImage:createCoverImage(doc_settings)
 
             local screen_ratio = s_w / s_h
             local image_ratio = i_w / i_h
-            local ratio_factor = image_ratio < screen_ratio and image_ratio/screen_ratio or screen_ratio/image_ratio
-            logger.dbg("CoverImage: geometries screen=" .. screen_ratio .. ", image=" .. image_ratio .. "; ratio=" .. ratio_factor)
+            local ratio_divergence_percent = math.abs(100 - image_ratio / screen_ratio * 100)
+
+            logger.dbg("CoverImage: geometries screen=" .. screen_ratio .. ", image=" .. image_ratio .. "; ratio=" .. ratio_divergence_percent)
 
             local image
-            if 100 - ratio_factor * 100 < self.cover_image_stretch_limit then -- stretch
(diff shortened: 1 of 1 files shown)


Frenzie:
"Better" image fit is disputable. I can see how the result might arguably be more pleasant at 1%, but 10%? Pfoo-ee. :-P

zwim:
Ok, then let's rename it to additional method to fit the image. And well --- for me --- below 10% I like the new look, above 10% stretching looks awfully.
Shall we set the default to let's say 7%, 5%?

Frenzie:
Personally I'd say 5% tops, but as not the target audience it's hard to say.

NiLuJe (approved):
LGTM ;).

As mentioned in #7214, if this were in core, I'd fight for a different default (say, 0% ;p), but I'm not overly bothered by 10% in a plugin. It should be just obvious enough to make people that light it happy, and people that don't unhappy enough to look for the setting ;).

plugins/coverimage.koplugin/main.lua
```diff
@@ -259,7 +273,41 @@ function CoverImage:addToMainMenu(menu_items)
                 end,
                 sub_item_table = {
                     {
-                        text = _("Scale, black background"),
+                        text_func = function()
+                            return T(_("Aspect ratio stretch threshold (%1%)"), self.cover_image_stretch_limit )
+                        end,
+                        help_text_func = function()
+                            return T(_("If the image and the screen have a similar aspect ratio (±%1%), stretch the image instead of honoring its aspect ratio."), self.cover_image_stretch_limit )
```
Frenzie, line 280:
```suggestion
                            return T(_("If the image and the screen have a similar aspect ratio (±%1%), stretch the image instead of keeping its aspect ratio."), self.cover_image_stretch_limit )
```

NiLuJe: That's mine, I'll readily accept the fact that I'm very fond of using 'honor' in this context ;p.

Frenzie: You do love to sprinkle the word honor around everywhere. ^_^

zwim: Honor to whom honor is due. ;)))))

PR #7223, opened 2021-01-31
Button: Better handling of translucent MovableContainer

NiLuJe:
* By actually being aware of that fact, and delaying a single full repaint post-callback and post-unhighlight, in order to avoid flickering glitches (especially when flash_ui is enabled).
* Scroll*Widget: handle belonging to a translucent dialog properly on scroll.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7223)

### DIFF
--- a/frontend/apps/reader/skimtowidget.lua
+++ b/frontend/apps/reader/skimtowidget.lua
@@ -316,6 +316,10 @@ function SkimToWidget:init()
             vertical_group_down,
         }
     }
+    self.movable = MovableContainer:new{
+        -- alpha = 0.8,
+        self.skimto_frame,
+    }
     self[1] = WidgetContainer:new{
         align = "center",
         dimen =Geom:new{
@@ -323,10 +327,7 @@ function SkimToWidget:init()
             w = self.screen_width,
             h = self.screen_height,
         },
-        MovableContainer:new{
-            -- alpha = 0.8,
-            self.skimto_frame,
-        }
+        self.movable,
     }
 
     if Device:hasDPad() then
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -674,6 +674,10 @@ to, among other things, flag the right widget as setDirty (c.f., those pesky deb
 This is why you often see stuff doing, when instantiating a new widget, FancyWidget:new{ show_parent = self.show_parent or self };
 meaning, if I'm already a subwidget, cascade my parent, otherwise, it means I'm a window-level widget, so cascade myself as that widget's parent ;).
 
+Another convention (that a few things rely on) is naming a MovableContainer wrapping a full widget `movable`, accessible as an instance field.
+This is useful when it's used for transparency purposes, which, e.g., Button relies on to handle highlighting inside a transparent widget properly,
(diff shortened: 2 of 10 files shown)


NiLuJe:
This allowed entirely getting rid of the NumberPicker workaround ;).

The "flickering" I'm speaking of can be experienced, for example, with a CRe margin SpinWidget, with flash_ui enabled. It's mildly annoying on emu/Android, and extremely irritating on eInk ;).

NiLuJe:
As for DictQuickLookup/Scroll*Widget: while preserving the alpha feels more natural, I'm wondering if it's actually a good idea to begin with: the default value makes it so translucent that actually being able to read the definition gets tricky, unless you happen to have a helpful river underneath ;).

Thoughts? A few other ideas to tackle that:

* Re-setting the container's alpha instead would be trivial to implement, for example
* Using a different alpha value to begin with so the dialog is *less* transparent (or modify it on scroll only?).

poire-z:
> The "flickering" I'm speaking of can be experienced, for example, with a CRe margin SpinWidget, with flash_ui enabled.

I don't notice anything :/

> It's mildly annoying on emu/Android, and extremely irritating on eInk ;).

Yes, you're talking about flash_ui, right ? :) But we can't get rid of it, as there's been so much work put into it :)

> As for DictQuickLookup/Scroll*Widget: while preserving the alpha feels more natural, I'm wondering if it's actually a good idea to begin with: the default value makes it so translucent that actually being able to read the definition gets tricky, unless you happen to have a helpful river underneath ;).

I guess the movable feature is more interesting than the translucent one, but as it's movable, it gets the other bonus. Not really useful in practice, except twice a year when debugging.
(But you can't get rid of it, as there's been so much work put into it ! :)

> Using a different alpha value to begin with so the dialog is less transparent (or modify it on scroll only?).

Not much interesting I guess.

> Re-setting the container's alpha instead would be trivial to implement, for example

Well, that feels better if you prefer that: you make it translucent somehow to move it away and see what's below. If you move it, you should keep it translucent as it is. Once you begin tapping buttons or the window to scroll content, assume you want reading it again and kill the alpha. Wouldn't look odd to me.

No comment about the PR, looks ok.

NiLuJe:
> > The "flickering" I'm speaking of can be experienced, for example, with a CRe margin SpinWidget, with flash_ui enabled.
> 
> I don't notice anything :/

I haven't actually checked on the emu, it might be very sneaky to catch there ^^. (It's already fairly sneaky on Android).

> 
> > Re-setting the container's alpha instead would be trivial to implement, for example
> 
> Well, that feels better if you prefer that: you make it translucent somehow to move it away and see what's below. If you move it, you should keep it translucent as it is. Once you begin tapping buttons or the window to scroll content, assume you want reading it again and kill the alpha. Wouldn't look odd to me.
> 

I don't really use the feature myself, so, no strong opinion either way, but, yeah, I'd tend to agree with that train of thought ;).

NiLuJe:
There, alpha is now reset on scroll & dict swap ;).

Also fixed a few other widgets to handle flash_ui and transparency properly, and documented that `movable` convention.

poire-z (approved):

PR #7226, opened 2021-02-01
Notifications: stack them instead of overdrawing them

poire-z:
(Emptying my list of fun-things-todo list.)
From https://github.com/koreader/koreader/pull/7135#issuecomment-758292956 : 
> Some other idea about Notification: knowing with :onShow()/onCloseWidget() how many are shown, the Notification class could store the list of active/shown instances, so a new one could have another y - so they don't override each other, but are stacked above each others.
(Not that I care much about Notification, but seeing 3 of them hiding each other with Dispatcher is a bit ugly :)

<kbd>stacked_notifs</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7226)

### DIFF
--- a/frontend/ui/widget/notification.lua
+++ b/frontend/ui/widget/notification.lua
@@ -10,9 +10,11 @@ local FrameContainer = require("ui/widget/container/framecontainer")
 local Geom = require("ui/geometry")
 local GestureRange = require("ui/gesturerange")
 local InputContainer = require("ui/widget/container/inputcontainer")
+local RectSpan = require("ui/widget/rectspan")
 local Size = require("ui/size")
 local TextWidget = require("ui/widget/textwidget")
 local UIManager = require("ui/uimanager")
+local VerticalGroup = require("ui/widget/verticalgroup")
 local Input = Device.input
 local Screen = Device.screen
 
@@ -23,6 +25,8 @@ local Notification = InputContainer:new{
     padding = Size.padding.default,
     timeout = 2, -- default to 2 seconds
     toast = true, -- closed on any event, and let the event propagate to next top widget
+
+    _nums_shown = {}, -- array of stacked notifications
 }
 
 function Notification:init()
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
Sure.

frontend/ui/widget/notification.lua
```diff
@@ -48,44 +53,73 @@ function Notification:init()
...
+    table.insert(Notification._nums_shown, os.time())
+    self.num = #Notification._nums_shown
+
+    self[1] = VerticalGroup:new{
+        align = "center",
+        -- We use Spans to properly position this notification:
+        -- have this VerticalGroup full width, to ensure centering
+        HorizontalSpan:new{ width = Screen:getWidth() },
+        -- push this frame at its y=self.num position
+        VerticalSpan:new{ width = notif_height * (self.num - 1) },
```
NiLuJe, line 86: `width`? Not familiar with VerticalSpan, but given the comment, I would have expected `height`?

NiLuJe: Okay, looking at VerticalSpân, that' sjust verticalSPan itself being *super weird*:

```lua
local VerticalSpan = Widget:new{
    width = 0,
}

function VerticalSpan:getSize()
    return {w = 0, h = self.width}
end

return VerticalSpan
```

o_O

poire-z: Yes, they both have this paramenter named `width`.

I just discovered we have a frontend/ui/widget/rectspan.lua that accepts width + height and that I could use instead of these 2 Horiz+Vert. This RectSpan is currently not used by anything !

frontend/ui/widget/notification.lua
```diff
@@ -48,44 +53,73 @@ function Notification:init()
...
+    -- from the end: new notifications can be shown there.
+    -- As a sanity check, we also forget those shown for
+    -- more than 30s in case no close event was received.
+    local expire_ts = os.time() - 30
+    for i=#Notification._nums_shown, 1, -1 do
+        if Notification._nums_shown[i] and Notification._nums_shown[i] > expire_ts then
+            break -- still shown (or not yet expired)
+        end
+        table.remove(Notification._nums_shown, i)
+    end
```
NiLuJe, line 107: Won't this fail to remove a non-tail index, if the tail itself is still shown and not expired?

(e.g., I would expect a `continue` instead of a `break` given the comment. If you forget the fact that Lua doesn't have a `continue` keyword ;p).

poire-z: My idea was to only remove tails. Even if slots have been freed at top, it felt nicer to add new one at bottom. Easier for the eyes to follow what is happening.
I can add that ^ as a comment if you agree with the idea.

NiLuJe: Right, that's because it's used for the vertical UI alignment, I hadn't made the connection ;).

Yep, makes perfect sense then, so, yeah, I wouldn't mind a better comment ;).

NiLuJe (approved):
:+1:

PR #7237, opened 2021-02-04
move opds to plugin

pazos:
Since it is not a required "app" like the reader or the filemanager. Also all the stuff now is placed together and can be disabled :)

1. replace hardcoded menu entry by menu hint.
2. added a dispatcher action to show catalog (from FM)

**to-do:** better name and description in `_meta.lua`

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7237)


frontend/ui/elements/filemanager_menu_order.lua
```diff
@@ -136,7 +136,6 @@ local order = {
         "find_file",
         "----------------------------",
         "goodreads",
-        "opds_catalog",
```
Frenzie, line 139: Please keep this. Sorting hint is just a backup. ;-)

plugins/opds.koplugin/_meta.lua
```diff
@@ -0,0 +1,6 @@
+local _ = require("gettext")
+return {
+    name = "opds",
+    fullname = _("OPDS"),
+    description = _([[OPDS desc]]),
```
Frenzie, line 5: NB

plugins/opds.koplugin/main.lua
```diff
@@ -0,0 +1,47 @@
...
+    local OPDSCatalog = require("opdscatalog")
+    local filemanagerRefresh = function() self.ui:onRefresh() end
+    function OPDSCatalog:onClose()
+        filemanagerRefresh()
+        UIManager:close(self)
+    end
+    OPDSCatalog:showCatalog()
+end
+
+function OPDS:onShowCatalog()
```
Frenzie, line 32: Maybe onShowOPDSCatalog? `ShowCatalog` seems potentially a bit generic.

Frenzie (commented):
I suggested it in so many words, so what can I say? ;-)

frontend/ui/elements/filemanager_menu_order.lua
pazos: It seems to work fine without it. I was under the impression that sorting_hint was meant to replace hardcoding plugin menu entries. Am I wrong?

Frenzie: Yes, definitely wrong. It only seems to work fine because it's literally doing the exact same thing in this one edge case. The menu is *designed*, not some random collection of items.

Frenzie: NB There's a difference between plugins we ship and *user* plugins.

pazos: So the purpose of the sorting hint for stock plugins is to be a backup in case the user removes the entry from default menu?

In that case the `coverimage` is missing its place on stock menus too :p

Frenzie: > So the purpose of the sorting hint for stock plugins is to be a backup in case the user removes the entry from default menu?

I suppose, but as far as the intent goes it's closer to custom menu + new plugin, and of course trying to provide a good example to plugin authors (because most people just modify something like OPDS and not HelloWorld).

Frenzie: > In that case the coverimage is missing its place on stock menus too :p

And that's definitely an issue, thanks.

plugins/opds.koplugin/_meta.lua
Frenzie: Something like this?

```suggestion
    description = _([[OPDS allows you to download books from online catalogs.]]),
```

PR #7238, opened 2021-02-04
[UX] Add ToC/Bookmarks settings

### LINKED ISSUE #7141: FR: Customize progress bar to display only h1 chapter markers
**Current**
Progress bar shows all chapter markers included in the ToC. For books with a complex structure this can appear too busy.

**Request**
It would be great to have an option to customize the depth of the chapter sampling, e.g. "Show chapter markers for h1 elements only"

poire-z:
- Menu widget: allow specifying the number of items per page and the item font size, so we can use other values than the default File browser ones
- Menu: fix setDirty when a border is used
- ToC: add item per page and font size settings, make Alternative ToC more visible (was previously available on long-press on Table of contents)
- Bookmarks: add item per page, font size, size reduction
- Progress bars (Skim widget and footer): allow selecting ToC depths from which ticks are made.

Discussed all along #7141. Screenshots at https://github.com/koreader/koreader/issues/7141#issuecomment-773289149. Closes #7141.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7238)


frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -59,6 +75,58 @@ function ReaderBookmark:addToMainMenu(menu_items)
...
+                title_text =  _("Bookmark font size"),
+                callback = function(spin)
+                    G_reader_settings:saveSetting("bookmarks_items_font_size", spin.value)
+                end
+            }
+            UIManager:show(items_font)
+        end,
+    }
+    menu_items.bookmarks_items_show_more_text = {
+        text = _("Reduce bookmark font size to show more text"),
```
Frenzie, line 121: Is the "show more text" part necessary?
```suggestion
        text = _("Smaller bookmark font size"),
```

poire-z: I think so, it's supposed to only go smaller when the text is long enough and it helps showing more/all of it.

Frenzie: I misunderstood then.

Something like `Auto(matically) shrink bookmark font size to fit more text` perhaps?

poire-z: image
vs:
image

frontend/ui/widget/menu.lua
```diff
@@ -1108,6 +1111,12 @@ function Menu:updateItems(select_number)
         local refresh_dimen =
             old_dimen and old_dimen:combine(self.dimen)
             or self.dimen
+        if not self.is_borderless then
+            refresh_dimen = refresh_dimen:copy()
+            local bordersize = Size.border.window
+            refresh_dimen.w = refresh_dimen.w + bordersize*2
+            refresh_dimen.h = refresh_dimen.h + bordersize*2
```
NiLuJe, line 1118: And `x` and `y` are accurate?

poire-z: Somehow, yes... I initially went with x = x-bordersize and same for y, didn't solve the right and bottom borders not shown issue.
The Menu indeed uses dimen.w for all inner sizing, so it might have a bigger issue with its layout - but I didn't want to refactor it :) (like I did for DictQuickLookup which had the same issue).

Didn't investigate much, that was just to solve the "hit F in the reader" ugly font menu (somehow, I hit that <kbd>F</kbd> quite often...)

poire-z:
Added a first commit to fix a crash when opening a document that was in scroll mode with an invalid last_xpointer:
```
02/04/21-21:16:20 DEBUG CreDocument: goto xpointer /html/body/divzzz[1]/text()[1].1
02/04/21-21:16:20 DEBUG CoverImage: onReaderReady
02/04/21-21:16:20 DEBUG _refresh: Enqueued partial update for region 0 0 600 610
02/04/21-21:16:20 DEBUG setDirty partial from widget ReaderUI w/ NO region
02/04/21-21:16:20 DEBUG CreDocument: set bookmarks highlight and internal history false
[!] doShowReader coroutine crashed:
frontend/document/document.lua:214: attempt to perform arithmetic on local 'page' (a nil value)
stack traceback:
  frontend/apps/reader/modules/readerfooter.lua:261: in function 'gen'
  frontend/apps/reader/modules/readerfooter.lua:1790: in function 'genFooterText'
  frontend/apps/reader/modules/readerfooter.lua:1907: in function 'updateFooterText'
  frontend/apps/reader/modules/readerfooter.lua:1893: in function 'updateFooterPos'
  frontend/apps/reader/modules/readerfooter.lua:2026: in function 'handleEvent'
  frontend/ui/widget/container/widgetcontainer.lua:95: in function 'propagateEvent'
  frontend/ui/widget/container/widgetcontainer.lua:113: in function 'handleEvent'
  frontend/ui/widget/container/widgetcontainer.lua:95: in function 'propagateEvent'
  frontend/ui/widget/container/widgetcontainer.lua:113: in function 'handleEvent'
  frontend/apps/reader/modules/readerrolling.lua:894: in function 'onRedrawCurrentView'
  frontend/apps/reader/modules/readerrolling.lua:128: in function 'v'
  frontend/apps/reader/readerui.lua:433: in function 'init'
  frontend/apps/reader/readerui.lua:578: in function 'doShowReader'
```

frontend/ui/widget/menu.lua
NiLuJe: Just for future reference, which bits of the menu does that affect, exactly?

NiLuJe: (IIRC, there's a hack where bits of the border are pushed off-screen to replicate a behavior/bug of ConfigDialog... or vic e-versa, I'm sure there's a comment in code about that).

poire-z: > there's a hack where bits of the border are pushed off-screen to replicate a behavior/bug of ConfigDialog

Yes, but that was voluntary, we wanted borders on top and bottom only, FrameContainer puts it on all sides, so trick to make it a bit larger so the L/R borders are not visible.
Which is not the case for Menu.

> which bits of the menu does that affect, exactly?

What _that_ ? My changes (it just extends a bit the setDirty area, for the reasons explain just below) or the bugs in menu.lua (in which case, all of it :)
But it affects only the Menu instances with is_borderless=false, ie not many of them: most are fullscreen & borderless.
The main issue is around there (line 634 is fine, line 640 is not):
https://github.com/koreader/koreader/blob/cf3fed252a8f515461552e1067d8c926d2f9e95e/frontend/ui/widget/menu.lua#L631-L640
All the rest of the code uses `self.dimen.w` for sizing the inner content. (It should use a `self.inner_width = self.width - 2*bordersize` and let `self.dimen.w`  alone for setDirty.)
So, at the end, everything is width'ed self.dimen.w, and we put all that in a FrameContainer{bordersize=2}, which makes it 2*2 pixels larger than `self.dimen.w`.
The positionning is off (and so my setDirty and me not fixing x/y), as can be seen with the `Find a file` result.

Menu is extended by so many other widgets (and more hackily by CoverBrowser) that we'd need to check all of them so they no longer use self.dimen.w/h... Quite some work I think.

yparitcher:
This causes a regression:
next and previous chapter gestures also use `getTocTicksFlattened()` so this PR causes that any chapter whose tick is not shown is also skipped when going to next chapter and in how many pages left to the chapter.
I think this was only supposed to affect the ticks displayed in a progress bar.

poire-z:
A regression, but only if this new feature is used :)

I mentionned it would affect prev/next chapter at https://github.com/koreader/koreader/issues/7141#issuecomment-770425411. But it does not affect the current chapter title, which makes all that a bit inconsistent.
Personally, I think I would not mind, I would happily consider the ticks/depths chosen the chapters that matter to me.
But I guess if this needs fixing, it would be for the next/prev chapter code to _not use_ the now unrelated (and unrelatedly named) `getTocTicksFlattened()`, and may be use a simple iteration of prev/next pages like done there in the first branch:
https://github.com/koreader/koreader/blob/ac668ecb64aea0a0b6593a864828db29ae7658b2/frontend/apps/reader/modules/readertoc.lua#L485-L504
(although that feels a bit expensive - and the isChapterStart() also uses getTocTicksFlattened() :/).

I'm not really interested in digging into that, so, feel free to :)

yparitcher:
K,
My use case i a book with a few sections and a lot of short chapters, so the chapters are important to me, but not very useful in the progress bar.

yparitcher:
the only eficient way to fix this is to have 2 versions of ` getTocTicksFlattened()` one with the ignored and one without, which is quite ugly :(

poire-z:
> which is quite ugly :(

Not that much if we only build a 2nd once we have toc_ticks_ignored_levels non empty.

poire-z:
@yparitcher : are you working on this? If not, I may.
I think my *I would happily consider the ticks/depths chosen the chapters that matter to me: page left/time till next tick* could be made an option in that menu `[ ] Ignore chapters for disabled ticks` or some better wording :)

Frenzie:
Maybe instead of "Ignore …" use the reverse? "Include hidden chapters" or something along those lines.

yparitcher:
> @yparitcher : are you working on this? If not, I may.
> I think my _I would happily consider the ticks/depths chosen the chapters that matter to me: page left/time till next tick_ could be made an option in that menu `[ ] Ignore chapters for disabled ticks` or some better wording :)

Please do, I am a little busy (getting married next month)

poire-z:
English grammarologists:
[ ] Skip chapters **for** ignored ticks
[ ] Skip chapters **of** ignored ticks
or else?

Although "skip" might be felt as stronger than what it does, like it would hide whole book sections from your reading - while it's supposed to only skip them when using prev/next chapter, or showing the chapter title in the footer.
Also dunno about a "reverse" formulation.
Also not sure what state of this feature should be the checked one :/

<kbd>image</kbd>

yparitcher:
`Ignore ticks everywhere`

yparitcher:
`Use ignored ticks as chapters` ?

yparitcher:
or `Chapters: use ignored ticks`

NiLuJe:
I'd add something along the lines of "Jump to" (chapters) or something similar in there, because otherwise it was a bit puzzling at first glance and I had to actively try to remember what you were talking about ^^.

hius07:
`Use for chapters navigation`, default: off

poire-z:
Not obvious this wording :)
"Use" is a bit ambiguous: hard to guess if it will "keep using" ignored level - or if it will also "use ignored level", so ignoring ignored level, for chapter navigation.
Rather than "Also ignore ignored entries", I went with "Keep ignored entries for chapters navigation".
Alternative suggestiosn still welcome.

Initial situation on a new book:
<kbd>image</kbd>

Default situation after unchecking one depth level:

<kbd>image</kbd>
<kbd>image</kbd>
We're still using all chapters, next chapter is on next page.

And after unchecking that menu item, + the help_text:
<kbd>image</kbd>
We don't use chapters for ignored depths: next chapter is in 19 pages.

I would have liked to also have the "chapter title" in the footer skip those ignore level, but it's another bit of code that is less obvious to tweak.

With that new menu item unchecked, the SkimTo prev/next chapter would just only to ticks:
image

(Grammar: "chapter**s** navigation", or "chapter navigation" singular ?)

NiLuJe:
Works for me ;).

> (Grammar: "chapter**s** navigation", or "chapter navigation" singular ?)

Singular ;).

hius07:
`Sync chapter navigation with ticks`, default: off

poire-z:
> `Sync chapter navigation with ticks` 

Good idea to have it default off.
What about something like "Constraint (or another verb) chapter navigation **to** ticks" ?

NiLuJe:
> What about something like "Constraint (or another verb) chapter navigation **to** ticks" ?

I potentially like the idea, but I can't quite find something that actually works better than your more verbose "keep..." proposal :s.

poire-z:
"Keep..." may also work better than "Sync..." if I add the buddy setting:
`[ ] Keep ignored entries for chapter navigation`
`[ ] Keep ignored entries for chapter titles`

`[ ] Sync chapter titles with ticks` (would that work?)

Chapter title would apply to what is shown in the footer, but also to what ends up in bookmark/highlights defaut text.

hius07:
`Bound chapter navgation to ticks`
`Chapter navigation follows ticks`

Separating chapter titles and navigation is too much, sorry.

poire-z:
I like `Bind chapter navigation to ticks`.
`Bind chapter titles to ticks` could also work. (As for being too much, it's buried deep in the menu, so it won't confuse many users, and given that the code involved is different, no reason to stick to using the same setting - might be handy some day on some book :)

NiLuJe:
It's a little less straightforward, but it works and it's possibly prettier?

Frenzie:
Bind chapter titles to ticks sounds quite mysterious… but maybe "bind chapter title navigation to ticks"?

poire-z:
Oh, these would be 2 different menu items. (`chapter title navigation ` is a bit mixed up, as we don't really navigate titles).

`Bind chapter navigation to ticks` (is that ok?) for the issue discussed above: next/prev chapter buttons bring you to next/prev ticks.

`Bind chapter titles to ticks` (ok, the notion of binding is less clear) would be another one to ignore chapter titles of ignored ticks, when displaying chapter title in the footer or in bookmarks.

Frenzie:
> Bind chapter navigation to ticks (is that ok?) for the issue discussed above: next/prev chapter buttons bring you to next/prev ticks.

Browse only visible chapter ticks?

poire-z:
> `Browse only visible chapter ticks` ?

Not really clear, a bit vague. Mentionning "chapter navigation" makes it clear that it will apply _only_ to that - if one can guess that's just the 2 skimto buttons <kbd>|<</kbd> and <kbd>>|</kbd> and the associated gestures actions - and if one does not know what it is, that it won't hurt much :)

gbyl:
@poire-z Is there a way to add an option to set a specific ToC depth as default? So if set to `x entries at ToC depth 1` for example, all books that are opened in the future maintain that sampling depth.

poire-z:
Currently no. Would need some additional work.
But I think for this setting, it's not a good idea: the ToC depth and busyness are something you're most likely to notice and discover on each newly opened book, before making a decision.
So, unless you're reading a serie of similar topic/publisher books that you know have the same ticks layour and issue, it feels you're most likely going to go change that default setting of yours for each new book. Feels saner to start new books with full ticks.

gbyl:
I see what you mean. I usually use ToC ticks in the progress bar as a point of reference and H1 is the right depth 99% of the time. But I understand how that can be very subjective. Thank you for being so responsive!

PR #7239, opened 2021-02-04
[chore] Properly set CoverImage menu order

Frenzie:
Reported by @pazos in <https://github.com/koreader/koreader/pull/7237#discussion_r570395126>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7239)

### DIFF
--- a/frontend/ui/elements/reader_menu_order.lua
+++ b/frontend/ui/elements/reader_menu_order.lua
@@ -100,6 +100,7 @@ local order = {
     },
     screen = {
         "screensaver",
+        "coverimage",
         "----------------------------",
         "screen_rotation",
         "----------------------------",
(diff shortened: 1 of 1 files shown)


poire-z:
May be with and just below "screensaver" ? I don't know if they are mutually exclusive, but the idea is it provides images to the system screensaver, so probably on devices where our native screensaver does not work?

Frenzie:
I don't have a full handle on the exact specifics, but indeed it's for devices where we don't control the screensaver but those devices apparently can take an image as input.

gbyl:
>the idea is it provides images to the system screensaver, so probably on devices where our native screensaver does not work?

>it's for devices where we don't control the screensaver but those devices apparently can take an image as input.

It's also for devices with poor battery performance on suspend where you still want a custom image on shutdown https://github.com/koreader/koreader/issues/6899#issuecomment-731616556

Frenzie:
Right, I forgot about that. That's quite… unique, however. ^_^

PR #7242, opened 2021-02-05
Add low battery status bar item which is hidden when above the low battery threshold

Thra11:
Most of the time I'm not interested in the battery level. However, when the battery's running low it's nice to keep half an eye on how much is left. I've implemented a new status item, "Battery status when low", which is hidden when "Hide empty items" is on, and the battery level is above the threshold configured in Device > Low battery alarm > Low battery threshold, but otherwise identical to the normal battery status.

I've implemented it as a new status on the assumption that the existing battery status should be kept as is.

Tested and works as intended, although ideally I'd like to avoid duplicating the regular battery status code, but I figured I'd get some feedback, see if people think this is useful, etc.

I wanted to do something like this:
```
    battery = function(footer)
         ...
    end,
    low_battery = function(footer)
        local batt_lvl = powerd:getCapacity()
        local threshold = G_reader_settings:readSetting("low_battery_threshold") or 20
        if footer.settings.hide_empty_generators and batt_lvl > threshold then
            return ""
        else
            -- delegate to normal battery status function
            return battery(footer)
        end
    end,
```
but I couldn't work out how to call the normal battery function (or if it's even possible in this scope). Guess I could factor it out into a separate function which they both call?

The other flaw with this implementation is that if you uncheck Device > Low battery alarm > Enable, this low battery status still works, and still uses the configured threshold, but it won't let you change the threshold.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7242)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -42,6 +42,7 @@ local MODE = {
     book_title = 12,
     book_chapter = 13,
     bookmark_count = 14,
+    low_battery = 15,
 }
 
 local symbol_prefix = {
@@ -147,6 +148,53 @@ local footerTextGeneratorMap = {
         local prefix = symbol_prefix[symbol_type].battery
         local powerd = Device:getPowerDevice()
         local batt_lvl = powerd:getCapacity()
+        -- If we're using icons, use fancy variable icons
+        if symbol_type == "icons" or symbol_type == "compact_items" then
+            if powerd:isCharging() then
+                prefix = ""
+            else
+                if batt_lvl >= 100 then
+                    prefix = ""
+                elseif batt_lvl >= 90 then
+                    prefix = ""
+                elseif batt_lvl >= 80 then
+                    prefix = ""
+                elseif batt_lvl >= 70 then
+                    prefix = ""
+                elseif batt_lvl >= 60 then
+                    prefix = ""
+                elseif batt_lvl >= 50 then
+                    prefix = ""
+                elseif batt_lvl >= 40 then
+                    prefix = ""
+                elseif batt_lvl >= 30 then
+                    prefix = ""
+                elseif batt_lvl >= 20 then
+                    prefix = ""
+                elseif batt_lvl >= 10 then
+                    prefix = ""
+                else
+                    prefix = ""
+                end
+            end
+            if symbol_type == "compact_items" then
+                return BD.wrap(prefix)
+            else
+                return BD.wrap(prefix) .. batt_lvl .. "%"
+            end
+        else
+            return BD.wrap(prefix) .. " " .. (powerd:isCharging() and "+" or "") .. batt_lvl .. "%"
+        end
+    end,
+    low_battery = function(footer)
+        local symbol_type = footer.settings.item_prefix or "icons"
+        local prefix = symbol_prefix[symbol_type].battery
+        local powerd = Device:getPowerDevice()
+        local batt_lvl = powerd:getCapacity()
+        local threshold = G_reader_settings:readSetting("low_battery_threshold") or 20
+        if footer.settings.hide_empty_generators and batt_lvl > threshold then
+            return ""
+        end
         -- If we're using icons, use fancy variable icons
         if symbol_type == "icons" or symbol_type == "compact_items" then
             if powerd:isCharging() then
@@ -843,6 +891,7 @@ function ReaderFooter:textOptionTitles(option)
             and T(_("Current time (%1)"), symbol_prefix[symbol].time) or _("Current time"),
         pages_left = T(_("Pages left in chapter (%1)"), symbol_prefix[symbol].pages_left),
         battery = T(_("Battery status (%1)"), symbol_prefix[symbol].battery),
+        low_battery = T(_("Battery status when low (%1)"), symbol_prefix[symbol].battery),
         percentage = symbol_prefix[symbol].percentage
             and T(_("Progress percentage (%1)"), symbol_prefix[symbol].percentage) or _("Progress percentage"),
         book_time_to_read = symbol_prefix[symbol].book_time_to_read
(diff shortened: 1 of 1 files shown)


poire-z:
> I've implemented it as a new status on the assumption that the existing battery status should be kept as is.

But people will probably never enable both.

> I'd like to avoid duplicating the regular battery status code

Me too :)

> I wanted to do something like this [...] but I couldn't work out how to call the normal battery function 

So, why not just hack into the normal battery function? Dunno how people use "Hide empty", but you could use it to now show it if low_battery_threshold is set (or add another such setting "Hide items below threshold" ).

> if you uncheck Device > Low battery alarm > Enable, this low battery status still works

Looks like there's another setting for that, that you'd want to check:
https://github.com/koreader/koreader/blob/d4d5dd25ba61b18cdc52669c8efc4e610e686366/frontend/apps/reader/modules/readerdevicestatus.lua#L58-L64

Thra11:
> 
> 
> > I've implemented it as a new status on the assumption that the existing battery status should be kept as is.
> 
> But people will probably never enable both.

True. I was more thinking that if somebody is happy with the current battery status behaviour, then they might not like me changing it with no option to go back. Having a separate status item would allow them to choose the behaviour they prefer. I'm happy for this to be an modification to the existing battery status instead if nobody minds (I figure it would only be a problem if somebody wanted hide-when-empty on for e.g. the wifi status, but wanted the battery status there all the time).

> 
> > I'd like to avoid duplicating the regular battery status code
> 
> Me too :)
> 
> > I wanted to do something like this [...] but I couldn't work out how to call the normal battery function
> 
> So, why not just hack into the normal battery function? Dunno how people use "Hide empty", but you could use it to now show it if low_battery_threshold is set (or add another such setting "Hide items below threshold" ).
> 

I use hide empty with the wifi status so it gently reminds me if the wifi is left on, but otherwise doesn't distract me. I don't know how other people use it. It could be done as a modification to the existing battery function or with new options. As explained above, I just assumed that some people might like the status quo and not want it to change.

> > if you uncheck Device > Low battery alarm > Enable, this low battery status still works
> 
> Looks like there's another setting for that, that you'd want to check:
> 

Sorry, I wasn't clear enough. The flaw isn't that the battery status hiding behaviour _should_ be disabled if the Low battery alarm is disabled. The way I see it, they are two alternative ways to get the same information, so it's not unreasonable to e.g. use the low battery status bar item to let you know when the battery's getting low _instead_ of the Low battery alarm popup. The problem is that at the moment the Low battery alarm 'owns' the threshold setting and won't let you change it if it's not enabled.

poire-z:
> I just assumed that some people might like the status quo and not want it to change

I don't know, may be. But if we're going to make every little thing configurable, our crowded footer settings menu will be double-crowded.
So, I'd say do as you want it with "hide empty", and see what feedback/complain you get :)

> Most of the time I'm not interested in the battery level

Most of the time I'm not interested by the footer as I'm just reading - but it's there and I don't think about it.
So, what bothers you with always having the battery level ?
Could it be that you find it too solid black and it catch your eyes too easily ? (and you'd share my feeling at https://github.com/koreader/koreader/pull/5537#issuecomment-548020267 - in my own patch, I replace them all by the 0% icon). If that's the case, may be make that an option: less catchy battery icon(s) (or `[x] Ditch that ridiculously big solid black battery icon` as I named it there :)

Thra11:
> Most of the time I'm not interested by the footer as I'm just reading - but it's there and I don't think about it.
So, what bothers you with always having the battery level ?

I just find it distracting / annoying. I like to have progress bar, current page, book time to read, and chapter time to read on my status bar. I'm used to them, so they aren't distracting, but the info's there when I want to look at it. Because I'm used to them, I notice any change, such as the wifi icon showing up, or with this PR, the battery status showing up. For me, it's a nice gentle reminder that you need to address the issue soon, without actually getting in the way like the Low battery alarm popup does.

It's a bit like email notifications on a phone or laptop. You don't want a popup in the middle of the screen demanding you deal with the email now, you just want a subtle notification like the unread email count appearing in the notification area. If there are no new emails, no notification. Same with battery status: if it's above the threshold, there's nothing to notify me about, so I don't want to see it.

Frenzie:
> I just assumed that some people might like the status quo and not want it to change.

I do. If I'm going on a 3+ hour train ride I'll charge it at 50 % (even though I probably don't have to) but at 80 % I won't.

> It's a bit like email notifications on a phone or laptop.

Phones and laptops always show the battery status. ;-)

Frenzie:
PS I have no attachment to the *form* of the battery icon. It's a little too fat.

poire-z:
> It's a bit like email notifications on a phone or laptop. You don't want a popup in the middle of the screen demanding you deal with the email now, you just want a subtle notification like the unread email count appearing in the notification area. If there are no new emails, no notification. Same with battery status: if it's above the threshold, there's nothing to notify me about, so I don't want to see it.

That feels like exactly the purpose of "Hide empty items".
So, either hardcode the hiding of the battery icon if that's enabled.
Or make it a submenu with either a level of hiding:
`only empty 0/"" items`
`same + a bit more like battery level > threshold`
or a few subsettings:
`[ ] empty or zero items` (or make these ones explicits, they already cover various things and a user might want only this but not that
`[x] battery level above threshold` 
`[ ] time before 23:00 when it's not yet time to go to bed`
`[ ] memory usage when < 75% available RAM`

Thra11:
> I do. If I'm going on a 3+ hour train ride I'll charge it at 50 % (even though I probably don't have to) but at 80 % I won't.

I'd do the same, but if I know I'm going to take a train, I'm making a conscious decision whether I need to charge it the night before, so I'm quite happy to look in the menu to check the battery charge. That said, different people use things differently. I don't want to tell anyone how they should use their device, which is why I implemented this as a new option instead of a modification to the existing battery indicator.

Out of interest, do you use the "hide when empty" feature at all? Because in its current state, this will only hide if that's enabled. One option for trying to please absolutely everybody would be to split the global "hide when empty" option into an individual "hide when empty" checkbox on each of the status items which can be hidden when empty.

> Phones and laptops always show the battery status. ;-)

That's not strictly true. Lots of laptop OSs/DEs hide the battery status except when it's doing something interesting (getting low, finished charging, etc.).

yparitcher:
my 2¢:
Please don't combine this with `Hide empty items`.

I use `Hide empty items` to only show the FL level when it is on. I usually leave it off an only want it to steal space from the progress bar when it is on so i can see the level.
Otoh I always like to know the battery level.

In a battery context `Hide empty items` would mean to me hide when at 0% (empty).
ie: for things that have an **off** state hide them when they are off. The battery level does not have an observable off state.

In general though, a setting to only show the battery on low battery could be useful.

PR #7245, opened 2021-02-05
Remove leftover in CoverImage after  #7239

zwim:
`sorting_hint` is not necessary after #7239 anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7245)


Frenzie:
No, it's intended to be there as well. :-)

Frenzie:
Oh, that commented line though, not sure why that's there?

PR #7256, opened 2021-02-06
Button: Fix some more weird highlighting corner-cases

NiLuJe:
Namely, the Terminal plug-in and its chain of InputDialog on top of each other (via the "Save" button, which opens a whole new fs InputDialog to set the filename, while keeping the file's fs InputDialog open).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7256)


poire-z:
(Do you still find all that manageable - and think there's a possible final stable state ? I'm for sure never touching any onTap anymore :) 30% of button.lua is for handling flash_ui - if it were just me, I would have killed the flash_ui code paths long ago to not have to wonder if we're the first or the second topmost... :/ Is it really worth all the hassle ? :)

NiLuJe:
Same answer as last time: I like it, because it unearths all the weird crap we do with the UI ^^.

NiLuJe:
To expand a bit:

* I'm fairly confident that any remaining issues are constrained to "that *one* button in that *one* weird plugin" scenarios ;).
* I actually *enjoy* looking for those/digging into reported issues, as it's extremely informative about how the UI/widget system works, and how we use it.

That said,

* If push comes to shove, there is an easy fallback plan for no/simpler heuristics (which, as I said, basically boils down to `setDirty(self.show_parent, ...)` ;).

poire-z:
OK :) Good that you are confident, it just feels like possibly never ending :)
I'm happy as long as this "noise" is kept wrapped in a `if not G_reader_settings:isFalse("flash_ui")` branch.

> there is an easy fallback plan for no/simpler heuristics (which, as I said, basically boils down to setDirty(self.show_parent, ...) ;).

Dumb question, as I, not using flash_ui, haven't seen how it behave before: what happened previously? Was it always doing it that easy fallback way?

NiLuJe:
>
> I'm happy as long as this "noise" is kept wrapped in a `if not G_reader_settings:isFalse("flash_ui")` branch.

Yeah, of course ;). This does sometimes lead to fun discoveries outside, but they're usually limited to specific new methods, or actual fixes for issues that were simply latent but invisible ;).

>
> Dumb question, as I, not using flash_ui, haven't seen how it behave before: what happened previously? Was it always doing it that easy fallback way?

Essentially, yes, although there's a layer of fun interaction that was effectively hidden away by the previous model that relied on scheduled functions, which, possibly by sheer luck, made the order in which things happened much less relevant.

NiLuJe:
And, of course, writing this down led to another shower-thought:

We could replicate this behavior, without the potential timing quirks since we now control the fencing:

1. Highlight
2. Refresh (e.g., fence)
3. Unhighlight
4. Do *not* refresh
5. callback

Doing the unhighlight pre-callback means we know that the widget is here, and that we'll paint *before* (so, *below*) the callback, removing the need for any heuristics. And the fact that we don't fence it means we don't delay the callback.

I need more sleep than I had last night to think through all the interaction, but this sounds like a better direction. The only (teeny, tiny) drawback would be an "useless" subwidget repaint for callbacks that close the parent, but this is potentially a drop of water in the ocean, and possibly even a win given the fact that we'd drop all the heuristics.

NiLuJe:
Sidebar: *this* is a perfect example of why code-reviews matter: everybody wins ;).

NiLuJe:
(PoC for that at https://github.com/NiLuJe/koreader/tree/flash_ui-v3, will continue tomorrow).

PR #7259, opened 2021-02-08
File browser settings: reorganize into Settings submenu

poire-z:
Move classic file browser settings from CoverBrowser plugin into FileManagerMenu, so they are available when this plugin is disabled (as they also apply to everything based on Menu).
Disccused at https://github.com/koreader/koreader/issues/7141#issuecomment-774727373.

Updated screenshots:

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7259)


hius07:
Great!

PR #7262, opened 2021-02-08
Revamp flash_ui handling, once more, with feeling ;)

### LINKED ISSUE #7320: ToC - page selector flash_ui
* KOReader version: 2021.02
* Device: Kindle PW2

---

NiLuJe:
Following a minor lightbulb moment in #7256 (thanks @poire-z ;)), this keeps the "we control the fencing" idea from v2, but keeps the *effective* order of events from v1, but hopefully laid out in a manner that make that order clearer.

v2, stemming from the naïve thought of "well, we should be able to do things in logical order: highlight -> callback -> unhighlight", hoped to be able to get rid of some unhighlight steps by checking the state of things post-callback, because you would then know if your widget was still visible or not, or stuff like that.
Unfortunately, what should have, at its core, been as simple as a few checks along the lines of
```
If parent is gone then abort 
If parent is entirely covered then abort 
If parent is beneath something then repaint parent 
Else (eg parent still top) unhighlight
```
very quickly proved to become a cesspit of weird heuristics, because we happen to do some fairly crazy/nasty things in the UI code, especially as far as updating existing widgets is concerned.
The good news is, this made this fact very obvious, and allowed to refactor a few of those into a semblance of sanity.
The bad news is, this required a number of weird heuristics to deal with these corner-cases, which, while fun for me, because it required thinking like UIManager, which was a very informative experience, was objectively not easily maintainable, despite many, many comments, and ultimately somewhat brittle (e.g., "it *mostly* works for now" ^^).

The other good news is that making the fencing explicit, instead of relying on tick scheduling (which made the actual order in which things happen slightly confusing unless you're familiar with how UIManager handles repainting and refreshing) makes the code easier to follow, is no longer dependent on something not screwing with the actual UIManager scheduling in the callback, and made some NTH features (e.g., preserving alpha in MovableContainers) much easier to follow and implement.

----

Anyway, the idea of this iteration is simple:

1. Highlight
2. Refresh (so we see the highlight)
3. Unhighlight
4. Do *NOT* refresh, but enqueue a refresh *request*
5. Callback
6. Refresh (so we get the callback results, and make the unhighlight effective).

Since the callback hasn't run when we *paint* the unhighlight, we can be assured that the widget exists, and that we paint before/below whatever the callback might do, and as such prevents bad interaction with said callback, rendering crazy heuristics unnecessary.
At worse, we're "wasting" a simple subwidget repaint (or, more often, a simple rectangle invert) in case the callback closed us. In practice, it's a very minor drawback to begin with (and the removal of the heuristics possibly make the performance point moot compared to v2 anyway), and, the only place where such an interactivity boost would be interesting (assuming an invert would actually be costly, which it isn't ;p) is essentially TouchMenu, since that's the one place where we can quickly chain taps, but, given how TouchMenu is designed, we can *keep* that optimization, because we know in advance if the menu will stay open ;).

----

* Simplify flash_ui handling (by handling the unhighlight pre-callback, c.f., #7262 for more details).
* UIManager: Handle translucent window-level widgets (and those wrapped in a translucent MovableContainer) properly in setDirty directly, making sure what's *underneath* them gets repainted to avoid alpha layering glitches. (This was previously handled via localized hacks).
* Update UIManager's documentation, and format it properly for ldoc parsing, making the HTML docs more useful.
* ReaderView: Reinitialize the various page areas when opening a new document, to prevent poisoning from the previous document.
* Event: Handle nils in an event's arguments.
* CheckButton/RadioButton: Switch to simple inversion to handle highlighting
* CheckButton: Make the highlight span the inner frame's width, instead of just the text's width, if possible.
* AlphaContainer: Fix & simplify, given the UIManager alpha handling.
* MovableContainer: When translucent, cache the canvas bb used for composition.
* Avoid spurious refreshes in a few widgets using various dummy *TextWidgets in order to first compute a text height.
* KeyValuePage: Avoid floats in size computations.

----

Fix #7320

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7262)


NiLuJe:
Planned for 2021.03 mainly for documentation purposes, to be able to easily compare:

v1 @ v2020.12 (e.g., [Button](https://github.com/koreader/koreader/blob/348d47365b1c68fae9914acb718526868b33f77a/frontend/ui/widget/button.lua#L219-L262))
v2 @ v2021.02 (e.g., Button)
v2 @ v2021.03 (e.g., Button)

The description of how things work in v1 is [here](https://github.com/koreader/koreader/pull/7256#issuecomment-774698943).

NiLuJe:
Note to self: double-check ldoc rendering on UIManager.

poire-z:
A bit too late to read all that - but can you just explain what you mean by "fence" ? just in here if that's obvious for an english speaker :), otherwise in the comments. Does it just mean that you block during highlight->forceRepaint->unhighlight?

NiLuJe:
I tweaked a few comments to avoid that wording, IIRC, but I'll do another pass ;).

Basically, setDirty on its own doesn't "do" anything, it just queues stuff for the next tick. By fencing, I meant forcing a drain on those queues at specific moments (namely, after highlight and after callback), instead of waiting for the next tick, in order to have greater control over what exactly gets batched together.

poire-z (commented):
Mostly just read the UIManager comments, went smooth, no remark (!).
Still not familiar with flash_ui to get all this business - will be interesting to do a diff between this and before v1 to see how v1+v2+v3 ends up to may be get a better grasp on all that :)

frontend/ui/uimanager.lua
```diff
@@ -594,13 +613,13 @@ dbg:guard(UIManager, 'unschedule',
 --[[--
 Registers a widget to be repainted and enqueues a refresh.
```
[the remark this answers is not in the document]
NiLuJe: Yeah, I was slightly puzzled at that, and made that clearer in `show`, but hadn't realized that the same wording was used for `setDirty`, good call ;).

NiLuJe:
Now includes a free workaround that allows Event to forward `nil`s properly.

NiLuJe:
And the latest set of commits squishes a few bogus repaint/refresh from dummy *Text* widgets used for text height computation.

I think this kind of shenanigan is limited to the few I caught between fb33fab...32ce63b, but do you think this may happen elsewhere, @poire-z?

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -572,6 +572,7 @@ function DictQuickLookup:init()
             face = self.content_face,
             width = self.content_width,
             height = self.definition_height,
+            _dummy = true,
```
poire-z: Note sure where _dummy is really for internal use or not, but it probably isn't here :) So, drop the leading `_`?

NiLuJe: I meant it as a sort or "this is an internal hack and read-only, don't modify me", so I kind of like the leading `_` for that, even if it is technically set from outside sometime ^^.

poire-z:
> but do you think this may happen elsewhere

Only one in DictQuickLookup ? (haven't looked, but it feels I've been doing that quite often :)
I would usually think about explicitely doing a `dummy_widget:free()` when I did that, so may be hunt for such `:free()`.
I think there may be some in the CoverBrowser plugin, and may be in TextEditor - and possibly in some other layout fixes I may have done over the last two years :/

frontend/ui/widget/dictquicklookup.lua
poire-z: `for_measurement_only = true` feels more explicite :) `_dummy` feels a bit... dummy

NiLuJe:
Yeah, the dreaded CoverMenu "best fit" loop ;p.

That one slips through the cracks unscathed because it isn't editable, and is a straight TBW, so there's no scrollbar or caret, and apparently it's using sane geometry and/or I'm not quite sure how to trip it ^^.

NiLuJe:
CalendarView does it via `TextWidget`, which doesn't do anything (e.g., setDirty) either, so it's safe.

frontend/ui/uimanager.lua
```diff
@@ -662,31 +716,31 @@ As far as the actual lifecycle of a widget goes, the rules are:
   this'll ensure that *any* widget including it will be sure that resources are freed when it (or its parent) are closed.
 * Note that there *is* a `Close` event, but it is *only* broadcast (e.g., sent to every widget in the window stack;
   the same rules about propagation apply, but only per *window-level widget*) at poweroff/reboot, so,
-  refrain from implementing custom onClose methods if that's not their intended purpose ;).
+  refrain from implementing custom `onClose` methods if that's not their intended purpose ;).
```
poire-z, line 719: There is an other "use case" of the onClose method:
```
frontend/apps/reader/skimtowidget.lua:        self.key_events.Close = { { "Back" }, doc = "close skimto page" }
frontend/apps/filemanager/filemanager.lua:            self.file_chooser.key_events.Close = nil
frontend/ui/widget/bboxwidget.lua:        self.key_events.Close = { {"Back"}, doc = "close windows" }
frontend/ui/widget/virtualkeyboard.lua:        self.key_events.Close = { {"Back"}, doc = "close keyboard" }
frontend/ui/widget/virtualkeyboard.lua:        self.key_events.Close = { {"Back"}, doc = "close keyboard" }
frontend/ui/widget/menu.lua:        self.key_events.Close = { {"Back"}, doc = "close menu" }
frontend/ui/widget/menu.lua:            self.key_events.Close = { {"Left"}, doc = "close menu" }
frontend/ui/widget/configdialog.lua:        self.key_events.Close = { { close_keys }, doc = "close config menu" }
```

NiLuJe:
Okay, as mentioned in #7269, moving the handling of translucent widgets inside UIManager made Button's flash_ui branch much less gnarly, and mostly removes the need for actual widgets themselves to care about that.

The added check on UIManager is hopefully not too gnarly.

frontend/ui/uimanager.lua
NiLuJe: Ah, that would explain the trend, thanks!

NiLuJe:
And double-checked that AlphaContainer now behaves ;).

poire-z:
> Planned for 2021.03 mainly for documentation purposes, to be able to easily compare

@Frenzie : when do you plan on releasing 2021.02 ? This weekend as its "Due by" says - or the next one as it usually happens? :)

Frenzie:
I haven't been able to test it. Is it ready?

poire-z:
"Ready", aren't we always - or never - ready ? :) 
Haven't tested with the last 2 commits, but I have with before that (although only the bits I'm using, ie not PDFs).
But it looks like a good time, there's not many risky things in or PR waiting, only a few translatable strings may be not yet done.
This PR and my upcoming SkimTo titlebarless and new translatable strings might require a few nightlies before being considered fine and stable.

Frenzie:
In a manner of speaking perhaps, but there are definitely less stable, less finished days/weeks. :-)

NiLuJe:
FWIW, I also feel like we're in a good spot for this week-end, FWIW ;).

Frenzie:
@poire-z 
> "Ready", aren't we always - or never - ready ? :)

And apparently we're not. ;-)

https://gitter.im/koreader/koreader?at=6027947e726a881d4f707908
https://gitlab.com/koreader/nightly-builds/-/jobs/1028035559
Screenshot_2021-02-13_10-51-14

poire-z:
> And apparently we're not. ;-)

Are we now ? :)

Frenzie:
Should be. I'll (re)tag it.

Frenzie:
(Not quite this very moment.)

poire-z:
We were ready, thanks ! :)
(Guess we should wait till tomorrow evening to begin bumping stuff.)

frontend/ui/widget/dictquicklookup.lua
poire-z: 9 days later, I still think this `_dummy` could have a better/more explicite name :)

poire-z:
I think I did have a quick review, and the _dummy was my only comment.
There are still a few `print()` here and there.

NiLuJe:
Took care of the print ;).

The `_dummy` thing vaguely rings a bell, but I must have missed it :?

poire-z:
You might have got it by mail - otherwise: burried deep above https://github.com/koreader/koreader/pull/7262#discussion_r573332111

frontend/ui/widget/dictquicklookup.lua
NiLuJe: Yup, completely slipped through the cracks ;p

yparitcher:
This is causing paper-cut glitches in menu items on my PW4. (tested by reverting just this commit)

when clicking on an item the black flash is not across the entire item the top left corner stays white the entire time.

this happens on the up arrow in the menu and items with a callback, and on some of the top level icons
This does not happen on items with a submenu.

It happens to different degrees sometimes a little black sometimes a little white, but when it happens **always** the bottom is square and the top is shorter on the left.

It happens more often when i press items quickly in a row (the screen is merging refreshes etc.)

NiLuJe:
Yep, not much we can do about it, the EPDC is being too lax w/ A2/DU updates.

NiLuJe:
Actually, possibly only w/ A2, since I can't reproduce it anymore on Kobo since the switch to DU.

The PW2 is already using DU, though, which means the EPDC is playing fast and loose on Kindle.

NiLuJe:
... and the only difference is that we don't request FORCE_MONOCHROME on Kobo, which means there's less reliance on the PxP, so, try that:

```diff
diff --git a/ffi/framebuffer_mxcfb.lua b/ffi/framebuffer_mxcfb.lua
index ae0bc9ea..068fa97d 100644
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -403,8 +403,8 @@ local function refresh_k51(fb, refreshtype, waveform_mode, x, y, w, h)
     end
     -- TEMP_USE_PAPYRUS on Touch/PW1, TEMP_USE_AUTO on PW2 (same value in both cases, 0x1001)
     refarea[0].temp = C.TEMP_USE_AUTO
-    -- Enable the appropriate flag when requesting what amounts to a 2bit update
-    if waveform_mode == C.WAVEFORM_MODE_DU then
+    -- Enable the appropriate flag when requesting a 2bit update
+    if waveform_mode == C.WAVEFORM_MODE_A2 then
         refarea[0].flags = C.EPDC_FLAG_FORCE_MONOCHROME
     else
         refarea[0].flags = 0
@@ -438,8 +438,8 @@ local function refresh_zelda(fb, refreshtype, waveform_mode, x, y, w, h, dither)
         refarea[0].dither_mode = C.EPDC_FLAG_USE_DITHERING_PASSTHROUGH
         refarea[0].quant_bit = 0;
     end
-    -- Enable the appropriate flag when requesting what amounts to a 2bit update, provided we're not dithering.
-    if waveform_mode == C.WAVEFORM_MODE_DU and not dither then
+    -- Enable the appropriate flag when requesting a 2bit update, provided we're not dithering.
+    if waveform_mode == C.WAVEFORM_MODE_A2 and not dither then
         refarea[0].flags = C.EPDC_FLAG_FORCE_MONOCHROME
     else
         refarea[0].flags = 0
@@ -474,8 +474,8 @@ local function refresh_rex(fb, refreshtype, waveform_mode, x, y, w, h, dither)
         refarea[0].dither_mode = C.EPDC_FLAG_USE_DITHERING_PASSTHROUGH
         refarea[0].quant_bit = 0;
     end
-    -- Enable the appropriate flag when requesting what amounts to a 2bit update, provided we're not dithering.
-    if waveform_mode == C.WAVEFORM_MODE_DU and not dither then
+    -- Enable the appropriate flag when requesting a 2bit update, provided we're not dithering.
+    if waveform_mode == C.WAVEFORM_MODE_A2 and not dither then
         refarea[0].flags = C.EPDC_FLAG_FORCE_MONOCHROME
     else
         refarea[0].flags = 0
@@ -571,7 +571,7 @@ local function refresh_cervantes(fb, refreshtype, waveform_mode, x, y, w, h)
     local refarea = ffi.new("struct mxcfb_update_data[1]")
     refarea[0].temp = C.TEMP_USE_AMBIENT
 
-    if waveform_mode == C.WAVEFORM_MODE_DU then
+    if waveform_mode == C.WAVEFORM_MODE_A2 then
         refarea[0].flags = C.EPDC_FLAG_FORCE_MONOCHROME
     else
         refarea[0].flags = 0
@@ -642,7 +642,7 @@ function framebuffer:init()
         self.mech_wait_update_complete = kindle_pearl_mxc_wait_for_update_complete
         self.mech_wait_update_submission = kindle_mxc_wait_for_update_submission
 
-        self.waveform_fast = C.WAVEFORM_MODE_A2
+        self.waveform_fast = C.WAVEFORM_MODE_DU
         self.waveform_ui = C.WAVEFORM_MODE_GC16_FAST
         self.waveform_flashui = self.waveform_ui
         self.waveform_full = C.WAVEFORM_MODE_GC16
```

yparitcher:
does not help :(

yparitcher:
Interesting it happens when holding an item with a submenu but not when pressing

NiLuJe:
Ah, right, the DU swap fixed that for taps, not holds.

Which makes no sense, but, there you have it ^^.

yparitcher:
That behavior is both w/wo your patch (DU swap does nothing)

NiLuJe:
Yeah, I was talking about prior experiments on Kobo where I managed to reproduce that on taps, too ^^.

yparitcher:
It is happening constantly, like 7 times out of 10

PR #7265, opened 2021-02-09
bump crengine: various fixes, support CSS "visibility"

### LINKED ISSUE #7257: "line-height: 0" breaks cover page
* KOReader version:
* Device: emulator

#### Issue

Adding `line-height: 0` to the code of an SVG-wrapped cover image (and maybe other images too), breaks the cover in two pages, with just a thin strip of the image on the second page.

Screenshot_20210207_093454

Another line-height issue (like #7234), maybe it's not a bug, but it's not the behaviour most people would expect, I guess, and not what e.g. the calibre viewer does.

#### Steps to reproduce

Open [this file](https://github.com/koreader/koreader/files/5938519/test.zip) (rename to epub). First page is the cover with `div { text-align: center; }`. Second page is the same cover, but with `div { text-align: center; line-height: 0; }`

Any line-height below 0.75 apparently causes the issue (with default font and window size). Interestingly, with 0.75, font size 30 is affected, but not 28 or 34 (rounding error?)

poire-z:
Includes https://github.com/koreader/crengine/pull/419 : 
- Fix max img height when small line-height - Closes #7257
- getRenderedWidths(): fix consecutive non-collapsable spaces
- CSS: adds support for "visibility: hidden"
- Text: fix some bgcolor issues, handle fgcolor transparent
- remove global gRootFontSize

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7265)


Frenzie (approved):
👌

PR #7266, opened 2021-02-10
Update macos instructions in Building.md to include homebrew grep

rexkerr:
Add instructions for installing homebrew grep on MacOS since the macos version of grep fails with an error about the -z flag.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7266)

### DIFF
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -41,12 +41,12 @@ Install the prerequisites using [Homebrew](https://brew.sh/):
 
 ```
 brew install nasm ragel binutils coreutils libtool autoconf automake cmake makedepend \
-sdl2 lua@5.1 luarocks gettext pkg-config wget gnu-getopt
+sdl2 lua@5.1 luarocks gettext pkg-config wget gnu-getopt grep
 ```
 
-You will also have to ensure Homebrew's gettext & gnu-getopt are in your path, e.g., via
+You will also have to ensure Homebrew's gettext, gnu-getopt, & grep are in your path, e.g., via
 ```
-export PATH="$(brew --prefix)/opt/gettext/bin:$(brew --prefix)/opt/gnu-getopt/bin:${PATH}"
+export PATH="$(brew --prefix)/opt/gettext/bin:$(brew --prefix)/opt/gnu-getopt/bin:$(brew --prefix)/opt/grep/libexec/gnubin:${PATH}"
 ```
 See also `brew info gettext` for details on how to make that permanent in your shell.
 
@@ -61,8 +61,6 @@ On Mojave, `10.09` has been known to behave with XCode 10, And `10.14` with XCod
 export MACOSX_DEPLOYMENT_TARGET=10.09
 ```
 *Note:* On Catalina (10.15), you will currently *NOT* want to deploy for `10.15`, as [XCode is currently broken in that configuration](https://forums.developer.apple.com/thread/121887)! (i.e., deploy for `10.14` instead).
-
-Next, you will need to install the Oracle JDK.  First, check to see whether you have it installed by running `javac` from the terminal.  If you get usage instructions written to the terminal, you have a JDK installed, but if you get a dialog box indicating that you need to install the JDK to use `javac`, click on the `More Information` button and follow the instructions.
 
 ## Getting the source
 
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -62,6 +62,7 @@ export MACOSX_DEPLOYMENT_TARGET=10.09
 ```
 *Note:* On Catalina (10.15), you will currently *NOT* want to deploy for `10.15`, as [XCode is currently broken in that configuration](https://forums.developer.apple.com/thread/121887)! (i.e., deploy for `10.14` instead).
 
+Next, you will need to install the Oracle JDK.  First, check to see whether you have it installed by running `javac` from the terminal.  If you get usage instructions written to the terminal, you have a JDK installed, but if you get a dialog box indicating that you need to install the JDK to use `javac`, click on the `More Information` button and follow the instructions.
 
 ## Getting the source
 


rexkerr:
Here's a page describing the problem:

https://oneminutemacman.com/how-to-make-the-annoying-to-use-the-java-command-line-tool-you-need-to-install-a-jdk-pop-up-go-away-forever/

The dialog box that macos displays when the `javac` placeholder is executed is here:

https://www.oracle.com/java/technologies/javase-downloads.html

pazos:
@rexkerr: if you're suggesting the change to be able to build android apks from macos please place the instructions in https://github.com/koreader/koreader/blob/master/doc/Building_targets.md#for-android-1

If you're building the emulator or the mac bundle you can ignore the (graphical) warning of javac. It is not a prerequisite to install the JDK and the build go fine without it. You can still add a note saying *please ignore graphical warnings or install a JDK if you want to get rid of them*

Or does the emulator/app bundle fail because there's no JDK?

rexkerr:
> @rexkerr: if you're suggesting the change to be able to build android apks from macos please place the instructions in https://github.com/koreader/koreader/blob/master/doc/Building_targets.md#for-android-1
> 
> If you're building the emulator or the mac bundle you can ignore the (graphical) warning of javac. It is not a prerequisite to install the JDK and the build go fine without it. You can still add a note saying _please ignore graphical warnings or install a JDK if you want to get rid of them_
> 
> Or does the emulator/app bundle fail because there's no JDK?

Thanks, I misunderstood what the warnings were, I was not trying to build the Android version.  That said, I did run into another problem, so I've made another edit to reflect that and removed the text about javac for now.

pazos:
It sounds a little weird. grep is part of the base install. And we actually use it without needing gnu grep. See https://github.com/koreader/koreader/blob/master/platform/mac/do_mac_bundle.sh#L133

It might have changed recently (ie: in `kodev`), I don't know. Could you paste the error you're trying to fix?

rexkerr:
When I edit my path to undo the change described above (so that it uses the default /usr/bin/grep), I get this error:

```
[*] Install resources
cp -R -pL resources/fonts/. koreader-emulator-x86_64-apple-darwin19.6.0-debug/koreader/fonts/.
install -d koreader-emulator-x86_64-apple-darwin19.6.0-debug/koreader/{screenshots,data/{dict,tessdata},fonts/host,ota}
grep: invalid option -- z
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
Emulator not found. Please build it first.
```

Using the homebrew version of grep makes the problem go away and it installs and runs.

pazos:
@rexkerr: good catch!

Does it work with `xargs -0 | grep debug` and `xargs -0 | grep -v debug` instead?

rexkerr:
No, I can't seem to make any variant of that work (the $? value is incorrect, and the for loop cannot parse the results), but looking at the script I wonder if the `-print0` is even necessary since the `find` is using `-maxdepth 1` and the emulator executable filename shouldn't ever have a space in it.  If the `-print0` is removed from `find`, then the `-z` can be removed from `grep`.  The `-print0` wasn't there prior to @NiLuJe's change, and I suspect that rest of the change (looping and building the file list one by one) was probably sufficient to address the previous issue.

NiLuJe:
Yeah, good catch, NUL separated input is a GNU grep feature (also implement in recent busybox versions, not that it helps you much ;p).

NiLuJe (approved):

pazos:
@rexkerr: not sure what you tried.

```diff
diff --git a/kodev b/kodev
index 9ab98a79..3be695a7 100755
--- a/kodev
+++ b/kodev
@@ -99,16 +99,16 @@ EOL
 }

 function setup_env() {
-    SETUP_ENV_GREP_COMMAND="grep -z -v debug"
+    SETUP_ENV_GREP_COMMAND="grep -v debug"
     if [ -n "${KODEBUG}" ]; then
-        SETUP_ENV_GREP_COMMAND="grep -z debug"
+        SETUP_ENV_GREP_COMMAND="grep debug"
         # for android adb install
         KODEBUG_SUFFIX=-debug
     fi
     local files=()
     while IFS= read -r -d $'\0'; do
         files+=("${REPLY}")
-    done < <(find . -maxdepth 1 -name 'koreader-emulator-*' -print0 | ${SETUP_ENV_GREP_COMMAND})
+    done < <(find . -maxdepth 1 -name 'koreader-emulator-*' -print0 | xargs -0 | ${SETUP_ENV_GREP_COMMAND})
     test ${#files[@]} -gt 0
     assert_ret_zero $? "Emulator not found. Please build it first."
     local idx=0
```
Generally speaking yep, gnu counterparts make sense. But in this case there's a posix friendly way available everywhere (even busybox) that worth trying ;)

rexkerr:
Yeah, I tried exactly that without luck.  Running the command within the command substitution `$()` from the bash prompt works as expected, but when run in the script it doesn't work for some reason. I'll give it another try, maybe I did something wrong.

rexkerr:
@pazos I tried it again, got the same result:

```
[*] Install resources
cp -R -pL resources/fonts/. koreader-emulator-x86_64-apple-darwin19.6.0-debug/koreader/fonts/.
install -d koreader-emulator-x86_64-apple-darwin19.6.0-debug/koreader/{screenshots,data/{dict,tessdata},fonts/host,ota}
Emulator not found. Please build it first.
```

Then I restore the original script and it works (with the gnu grep)

pazos (approved):

PR #7269, opened 2021-02-10
Page Overlap: Fix rectangle computation and arrow mode

### LINKED ISSUE #5322: Possibly buggy page overlap arrow
* KOReader version: current nightlies
* Device: Kobo Forma (but probably irrelevant)

#### Issue

The "arrow" variant of the page overlap feature appears to be a bit wonky when paging *backwards*: it's positioned where the overlap cutoff would be if we had paged *forward*.

#### Steps to reproduce

(Say, a PDF page is 125% taller than your screen, if you page forward, you get 75% of overlap, with the arrow near the bottom. If you then page backward, it should be near the top. It stays at the bottom).

Graying out behaves properly, on the other hand.

N/A

NiLuJe:
* In ReaderPaging, the panning step pre-PanningUpdate can be wildly overshot near page edges, so, use the corrected value instead by recomputing it after the panning has been effectively computed by ReaderView.
This fixes slight inaccuracies, as well as glaring mistakes when going backwards, or when near page edges.
This is in line with how ReaderRolling computes the value, which I only realized later because I'm an idiot. 
* Minor cleanups around the handling of the dim_area Geom object in general.

* Fix the "Arrow" page overlap mode to be painted in the right coordinates when going backward. Issue might not have been terribly clear because of the previous issue ;).
* Center the arrow's point, while we're here.
* Don't use AlphaContainer to make it translucent, because AlphaContainer is horribly broken, and has weird quirks and behavior that make no sense to me unless some very specific and unlikely constraints are met, and they definitely aren't here.
This fixes the arrow copying an arrow-sized square of the original page the book was opened on on the top-left corner of *every* page with an arrow. (lol).
* Do real proper alpha-blending via Icon/ImageWidget from the original icon, instead of faking it via addBlitFrom, in order to avoid the dimming *around* the triangle's shape.

Fix #5322

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7269)


poire-z:
> @fixme: Why, oh why? What was the intended use-case?
        --          This makes absolutely no sense in, like, most cases

Well, that was from me :) but I can't say much more... AlphaContainer was really tough to get somehow working...
And as indeed, the arrow is the single user of AlphaContainer, and if you made it work without, with the alphaBlitFrom stuff, and if that's enough to solve future use case, we may as well drop the AlphaContainer widget.

NiLuJe:
Thinking about it not-at-3-AM, the best I can think of is that it was an attempt to avoid having to repaint widgets below it to avoid alpha layering glitches?

Which we've eventually fixed by doing setDirty("all") requests instead, which is costlier, but works all the time ;).

I have vague ideas to: make this one at least slightly less brittle here; and attempt to move the alpha handling to UIManager (in #7262), because, admittedly, that would be its job: if a translucent widget is flagged dirty, flag anything below it dorty too, and stuff then magically works properly without hackery :).

NiLuJe:
Okay, made the IconWidget usage much cleaner, by just allowin to honor `alpha` for icons in ImageWidget, and using a tweaked arrow icon at 75% fill opacity.

This gets rid of the dimming *around* the arrow itself (the square) that was caused by addBlitFrom, since we now do proper alpha-blending, and the surrounding area is truly completely transparent ;).

NiLuJe:
(Fixed it with the power of SVG! :rainbow: :unicorn: @Frenzie would be proud ;p).

poire-z (approved):
Not really familiar with the code paths involved, but looks ok.

Frenzie:
lol, I don't see you manipulating the SVG programmatically :angel:

PR #7272, opened 2021-02-10
Improved ramp-down of warm-light after max hour

rexkerr:
This change is to improve the ramp-down of the warm-light.  Currently the warm light stays at 100% for 3 hours, and then drops to 50% and then 0%, meaning that you get a suddend and major change if you're still up reading at that hour.  If you set your max-warmth to 20:00 (8:00pm), then it'll be at max warmth at 8, 9, 10, then 50% at 11, and back to 0% at midnight.  

Ideally there would be a way to adjust these factors or to set the time at which it starts to return to 0% and calculate from there, which I might implement later if it doesn't look too hard, but for now this is still much better as it drops by 15% per hour giving you:

100,100,100,85,70,55,40,25,10,0 -- if you start at 8:00pm (as above) it'll be back to 0 at 5am.  If you start at 11:00pm, then it'll be back to the coolest setting at 8:00 am.

image

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7272)

### DIFF
--- a/frontend/device/cervantes/powerd.lua
+++ b/frontend/device/cervantes/powerd.lua
@@ -152,18 +152,20 @@ function CervantesPowerD:calculateAutoWarmth()
     local current_time = os.date("%H") + os.date("%M")/60
     local max_hour = self.max_warmth_hour
     local diff_time = max_hour - current_time
+    -- number of hours to stay at full warmth before decreasing
+    local warm_window = 4
     if diff_time < 0 then
         diff_time = diff_time + 24
     end
     if diff_time < 12 then
         -- We are before bedtime. Use a slower progression over 5h.
         self.fl_warmth = math.max(20 * (5 - diff_time), 0)
-    elseif diff_time > 22 then
-        -- Keep warmth at maximum for two hours after bedtime.
+    elseif diff_time > (24 - warm_window) then
+        -- Keep warmth at maximum for 'warm_window' hours after bedtime.
         self.fl_warmth = 100
     else
-        -- Between 2-4h after bedtime, return to zero.
-        self.fl_warmth = math.max(100 - 15 * (22 - diff_time), 0)
+        -- Decrease for each subsequent hour by 20% per hour
+        self.fl_warmth = math.max(100 - 20 * (24 - warm_window - diff_time), 0)
     end
     self.fl_warmth = math.floor(self.fl_warmth + 0.5)
 
--- a/frontend/device/cervantes/powerd.lua
+++ b/frontend/device/cervantes/powerd.lua
@@ -163,7 +163,7 @@ function CervantesPowerD:calculateAutoWarmth()
         self.fl_warmth = 100
     else
         -- Between 2-4h after bedtime, return to zero.
-        self.fl_warmth = math.max(100 - 50 * (22 - diff_time), 0)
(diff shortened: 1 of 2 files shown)


NiLuJe:
Might even switch to 10% steps, as most recent devices only have 11 steps of granularity [0..10].

Frenzie (approved):
Seems fine by me.

pazos:
> Might even switch to 10% steps, as most recent devices only have 11 steps of granularity [0..10]. Our internal scale is kept as a percentage for... reasons.

Yup. There're only two kobo devices (or three?) that have more than 11 steps: Kobo Aura One and Kobo Aura H2O2

rexkerr:
Yeah, I noticed that the kobo device properties only showed a range up to 10, and that it was scaled.  That said, I chose a number > 10 because I thought that if it decreased too slowly that it would still be too warm in the morning for early-birds.

So, I've been thinking about the values that I chose and experimenting and 20% jumps don't seem too bad -- way better than the 50% jumps.  I also think that for most people it's unlikely that they'll be reading in the middle of the night if they set their max hour at 8-10pm or so... so I think that it might actually be better to widen the 100% window to 4-5 hours and then start dropping at 20%.  If nobody objects I might make that change.  I'll convert this to a draft until I hear back from others.

rexkerr:
....oh, also, any tips for centralizing the identical code?  I might do that as a separate pull request as that'll take much more testing.

Frenzie:
> ....oh, also, any tips for centralizing the identical code? I might do that as a separate pull request as that'll take much more testing.

As methods in SysfsLight perhaps, or some sister file anyway. As far as I'm concerned the specifics don't matter too much as long as they're mostly sane, and minor stuff like naming is easy to change regardless.

rexkerr:
After more testing, I didn't like the change that I made as much as I'd hoped, so I tweaked it a bit more, this is the new profile:

image

The max warmth window is now four hours long rather than two, and then I drop by 20% per hour to be sure that it's back to white at a reasonable hour in the morning -- a 20% drop wasn't as dramatic as I expected because it's not dropping all at once, but rather by 10% every half-hour since the calculation as a floating point value that takes minutes into account (see the definition of `local_time`)

rexkerr:
> > ....oh, also, any tips for centralizing the identical code? I might do that as a separate pull request as that'll take much more testing.
> 
> As methods in SysfsLight perhaps, or some sister file anyway. As far as I'm concerned the specifics don't matter too much as long as they're mostly sane, and minor stuff like naming is easy to change regardless.

Thanks, I think I'll do that cleanup as a separate PR if I have time.  I'd like to give this code more attention as I've seen a bunch of little bugs, like not setting the correct warmth value when first resuming about 5-10% of the time.

NiLuJe (approved):
LGTM, thanks!

NiLuJe:
As far as testing goes, we're probably going to have to rely on our guinea pigs^W^Wnightly users, because I only have a Forma, which behaves exactly like your Clara. (And, also, I don't use the feature ^^).

As @pazos said earlier, we'd need someone with a KA1, most likely.

That said, this seems harmless enough, and I don't really expect any kind of weird fallout anywhere ;).

pazos (approved):

PR #7281, opened 2021-02-13
SkimToWidget: remove title bar, add bookmark button

### LINKED ISSUE #6032: FR: Skim window: bookmarks, chapters, go back
I have suggestions to improve functionality of skim window:
1) Add "Add/remove bookmark" button as the top middle button. Use case is when skimming text rapidly to quickly bookmark desired pages.

2) Show current chapter in the skim window. I am reading a book right now which has h3 chapters without page breaks, so the titles can appear in the middle of text. It took me a while to figure out that I actually _was_ jumping from chapter to chapter, but the skim window in its current position simply covered the chapter title. So if we have a chapter name in the skim window, then we will have clear indication of where we are and be able to notice that chapter changed.

3) I hope this won't make things too complicated. Here is the idea. Can you add "Go back to previous position" button to the skim window? By "previous position" I imply the page where the skim window was called. Use case: e.g. I am in the middle of chapter 4, on page 123. Then I open skim window, go back and forth between bookmarks/chapters, take mental notes, and then I just want to continue reading where I was opening skim window. For now I just place a bookmark manually, remember its page number, do skimming, and open that bookmark again.

### LINKED ISSUE #7283: 'Go to Location' vs 'Go to Page'
I propose remove 'Go to Location'.
'Go to Page' is OK everywhere.

1. Yes, we have no fixed pages in cre documents. But reading fb2 or epub book I **do have** pages for my font and page settings.
2. Locations are not mentioned anywhere but in the 'Go to' dialog.
3. Looking into `readergoto.lua` - there are no special actions for locations (besides `curr_page`).
4. We are really talking about pages everywhere.

poire-z:
Remove the title bar, as it was just taking uneeded space over the book content. See https://github.com/koreader/koreader/issues/6032#issuecomment-776902493.

Add a middle button in the top row to toggle bookmark for the current page. See https://github.com/koreader/koreader/issues/6032#issuecomment-776794897.

Rework the UI building to get more consistent and aligned buttons and progress bar.
Move SkimToWidget from apps/reader/ to ui/widgets/.
Closes #6032.

<kbd>image</kbd>

<kbd>image</kbd>

Also:
`Goto dialog: use "Go to Page" also with CreDocuments` Closes #7283 .

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7281)


frontend/ui/widget/skimtowidget.lua
```diff
@@ -264,76 +221,86 @@ function SkimToWidget:init()
...
     }
+    self.button_bookmark_toggle = Button:new{
+        text_func = function()
+            return self.ui.view.dogear_visible and bookmark_enabled_text or bookmark_disabled_text
+        end,
+        radius = 0,
+        enabled = true,
+        width = button_inner_width,
+        show_parent = self,
+        vsync = true,
```
NiLuJe, line 232: Note to self: double-check if vsync is necessary here ;).

frontend/ui/widget/skimtowidget.lua
```diff
@@ -264,76 +221,86 @@ function SkimToWidget:init()
             self:goToBookmark(page)
         end,
     }
+    self.button_bookmark_toggle = Button:new{
```
NiLuJe, line 224: Random inconsequential nit: this one's an instance member, the others are locals.

poire-z: Yes, just like your `self.current_page_text` , because this one, like yours, go thru `Button:setText()` in `SkimToWidget:update()` (that I tweaked a bit to just use `:text_func()` to avoid duplicating that text building code).

NiLuJe: Oh, missed that, cool :).

NiLuJe (approved):

poire-z:
Added a commit to close #7283.

frontend/apps/reader/modules/readergoto.lua
```diff
@@ -60,7 +47,7 @@ function ReaderGoto:onShowGotoDialog()
         input_hint = T("@%1 (1 - %2)", curr_page, self.document:getPageCount())
     end
     self.goto_dialog = InputDialog:new{
-        title = dialog_title,
+        title = _("Go to Page"),
```
Frenzie, line 50: Should that be `Go to page`? I think we might have some inconsistencies.

poire-z: You tell me :) We have a few (mostly in the bottom menu) `Camel Titles`.
But ok, looks ugly in dialog titles.

Frenzie: I'd prefer to use `Go to page`.

poire-z:
Should it also be "Go to page" in the menu?
image

hius07:
> Should it also be "Go to page" in the menu?

I'd say yes, it looks good.

poire-z:
Done:
image

PR #7282, opened 2021-02-13
KeyValuePage: configurable items per page

### LINKED ISSUE #3620: Items per page in reading statistics
Is it possible to manually change the default value (14 for me) somewhere?

poire-z:
See https://github.com/koreader/koreader/issues/3620#issuecomment-777775675. Closes #3620.

#### KeyValuePage: configurable items per page

Tweak building to start from items per page instead of a fixed item height.
Guess the best font size that fit.
Update separator specification from using a `"----"` to the now generic `separator=true` (this allows not wasting a slot for each separator in the page and not have only 12 items and 2 small lines in a 14 items page).

<kbd>image</kbd>

Before | After:
<kbd>image</kbd>

And when switched from the default of 12 (on my emulator with Screen DPI: small) to a custom value of 16:
<kbd>image</kbd>

#### Reading statistics: tweak book stats views

Make "Current statistics" and previsouly opened book statistics display missing info that the other view
has, mostly:
- Pages read: `nb (pct%)`
- Current page/Total pages: `num/total (pct%)`

Stats for the same book as above, but invoked from Calendar or Time range when book is closed:
<kbd>image</kbd>

(The difference between current page 22 and last read page 21 is because I didn't stay long enough on that page 22, and the not-opened book use the last page_stat saved, to avoid opening the docsetting - and because past books may no longer be there so no docsettings to look at.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7282)

### DIFF
--- a/plugins/statistics.koplugin/main.lua
+++ b/plugins/statistics.koplugin/main.lua
@@ -1099,8 +1099,8 @@ function ReaderStatistics:statMenu()
                     }
                     UIManager:show(self.kv)
                 end,
+                separator = true,
             },
-            "----",
             { _("Last week"),"",
                 callback = function()
                     local kv = self.kv
@@ -1303,8 +1303,7 @@ function ReaderStatistics:getCurrentStat(id_book)
         { _("Pages read this session"), tonumber(current_pages) },
         -- today
         { _("Time spent reading today"), util.secondsToClock(today_duration, false) },
-        { _("Pages read today"), tonumber(today_pages) },
-        "----",
+        { _("Pages read today"), tonumber(today_pages), separator = true },
         -- Current book statistics
         -- Includes re-reads
         { _("Total time spent on this book"), util.secondsToClock(total_time_book, false) },
@@ -1324,7 +1323,7 @@ function ReaderStatistics:getCurrentStat(id_book)
         { _("Estimated reading finished"),
             T(N_("%1 (1 day)", "%1 (%2 days)", estimate_days_to_read), estimate_end_of_read_date, estimate_days_to_read) },
 
-        { _("Highlights"), tonumber(highlights) },
+        { _("Highlights"), tonumber(highlights), separator = true },
         -- { _("Total notes"), tonumber(notes) }, -- not accurate, don't show it
     }
 end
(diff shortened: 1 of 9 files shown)


hius07:
Three very pleasant PRs, thanks a lot!

plugins/statistics.koplugin/main.lua
```diff
@@ -1295,16 +1295,15 @@ function ReaderStatistics:getCurrentStat(id_book)
     local time_to_read = (self.data.pages - self.view.state.page) * self.avg_time
     local estimate_days_to_read = math.ceil(time_to_read/(book_read_time/tonumber(total_days)))
     local estimate_end_of_read_date = os.date("%Y-%m-%d", tonumber(now_ts + estimate_days_to_read * 86400))
-    local formatstr = "%.0f%%"
+    local estimates_valid = time_to_read > 0 -- above values could be 'nan' and 'nil'
```
NiLuJe, line 1298: Which *above values* would that be?

Because if it's `time_to_read`, that won't work (can't compare nil to a number, at the very least. Not quite sure about NaN).

plugins/statistics.koplugin/main.lua
```diff
@@ -1314,18 +1313,35 @@ function ReaderStatistics:getCurrentStat(id_book)
...
+        { _("Current page/Total pages"), string.format("%d/%d (%d%%)", self.curr_page, self.data.pages, Math.round(100*self.curr_page/self.data.pages)) },
         -- estimation, from current page to end of book
-        { _("Current page/Total pages"),  self.curr_page .. "/" .. self.data.pages },
-        { _("Percentage completed"), formatstr:format(self.curr_page/self.data.pages * 100) },
-        { _("Estimated time to read"), util.secondsToClock(time_to_read, false) },
-        { _("Estimated reading finished"),
-            T(N_("%1 (1 day)", "%1 (%2 days)", estimate_days_to_read), estimate_end_of_read_date, estimate_days_to_read) },
-
-        { _("Highlights"), tonumber(highlights) },
+        { _("Estimated time to read"), estimates_valid and util.secondsToClock(time_to_read, false) or _("n/a") },
```
NiLuJe, line 1322: nit: I tend to prefer `N/A` to `n/a` (but we do have a few n/a instances in the coded base. less that N/As, though ;p).

plugins/statistics.koplugin/main.lua
poire-z: > Which above values would that be?

Implicit, but `estimates_valid` is about the ones starting with `estimate_` .

I had no crash, I was just seeing NaN and nil for these 2 estimates on a new book without any page_stat_data entry yet.

NiLuJe: Oh, okay, the two `estimates` above, hence the variable name ;p.

And, at a quick eyeballing, `time_to_read` itself should be sane.

plugins/statistics.koplugin/main.lua
poire-z: OK, I switched them all to `N/A` (no preference myself :)

NiLuJe (approved):

PR #7296, opened 2021-02-16
Reversing commit 2348cd, changes to warmth values

rexkerr:
Somewhat urgent PR, please merge before the next stable build -- the changes I made somehow made the warm light hours even worse, so I'm reverting them.

Even though I've plotted the output of the algorithm multiple times in both a spreadsheet, and copied the algorithm into Python and made the minimal changes necessary to make it work in Python and plotted the output using numpy and the algorithm is absolutely correct -- there's something VERY weird about the warm-light code.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7296)

### DIFF
--- a/frontend/device/cervantes/powerd.lua
+++ b/frontend/device/cervantes/powerd.lua
@@ -152,20 +152,18 @@ function CervantesPowerD:calculateAutoWarmth()
     local current_time = os.date("%H") + os.date("%M")/60
     local max_hour = self.max_warmth_hour
     local diff_time = max_hour - current_time
-    -- number of hours to stay at full warmth before decreasing
-    local warm_window = 4
     if diff_time < 0 then
         diff_time = diff_time + 24
     end
     if diff_time < 12 then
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
Could be tonight.

NiLuJe:
On the off-chance someone has a bright (ha!) idea, what were the differences between the expected results and what actually happened?

rexkerr:
@NiLuJe  For a bit of context -- the behavior was very strange -- the peak brightness time changed to 6:00 pm when set to 10:00 pm, which is exactly what my window variable was (coincidence?).  And then instead of dropping slowly it dropped even faster than before.

NiLuJe:
Wild guess (only from the visible context in the diff, moght apply elsewhere in the logic): make sure everything is truncated to an int (e.g., math.floor()).

Lua numbers are `double`, and we've seen *very* weird shit happening in the past when stuff was fed a decimal when it didn't really expect to ever see one.

PR #7300, opened 2021-02-16
Wikipedia results: fix images not updated across results

poire-z:
Some results fields were forgotten by the "smarter update" from #7212.
See https://github.com/koreader/koreader/pull/7166#issuecomment-779654246.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7300)

### DIFF
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -401,12 +401,21 @@ function Button:refresh()
 end
 
 function Button:onHoldSelectButton()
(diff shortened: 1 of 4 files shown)


Frenzie (approved):

NiLuJe (approved):
:lgtm:, thanks!

----

And, yeah, making the "starting" state canonical for the *Release handlers seems sensible; and if something actually relied on the reverse behavior, I'd consider it a bug in that something and not an issue w/ Button ;).

Issue #7304, opened 2021-02-16
Double Tap (left and right sides and top) and Screen orientation issues


sim-mig:
* KOReader version:2021-02.01
* Device:Pocketbook inkpad 3 (newest Fw)

1-Double Tap Issue
The more sophisticated gestures (Swipes, multi-swipes) are not really usable because interaction with the pocketbook settings (they mess with intensity of light and warmth). But he simpler gestures (Taps) work almost perfectly apart Double Tap on "left and right side" and "Top left and right" which don't trigger anything. 

Steps to reproduce
If you allocate whatever event to Double Tap on "left and right side"
or on "Top left and right", this Double Tap will unfortunately have no effect. 

The strange thing is that the event assigned to the "left and right side"
will work for the "Top left and right"… But no work around for "left and right side". 

2-Screen orientation issue
When you read a document in the horizontal mode, the 90° orientation settings will  do the opposite of what they supposed to do : the rotation toward the right will go to left and vice versa, easily reproducible.

Those two issues shouldn't be very difficult to fix. 
Apart that KoReader works pretty well with the Inkpad 3.

NiLuJe:
1. PB shenanigans, we don't have an active PB maintainer.

----

As far as 2. is concerned, the arrow represent a *device* rotation, not a *content* rotation.

c.f., https://github.com/koreader/koreader-base/blob/d980ec9c487a0ab1508b8d7c7c6a46f253fa17a8/ffi/framebuffer.lua#L42-L68 and the various discussions we've had on the subject in the past ^^.

poire-z:
> the various discussions we've had on the subject in the past 

See https://github.com/koreader/koreader/issues/6945#issuecomment-738215400 and #6309.

Frenzie:
I'm fine swapping them, fwiw. I'm inclined to think it's confusing either way.

NiLuJe:
I'm very much *not* fine swapping them,, FWIW ;).

Frenzie:
I primarily mean that a lot of people seem to be confused by it. But yeah, we definitely shouldn't change it willy-nilly.

NiLuJe:
^^

https://xkcd.com/386/

sim-mig:
Hi, 

Doesn't it seem logical that the content turns in the direction of the arrow (following what they do in softs like Photoshop, etc)? 
Otherwise, if you think it's confusing, this menu could be in the mood of the Pocketbook one, with explicit icons :

[...]

PR #7306, opened 2021-02-16
[RFC] Switch ConfigDialog's rotation toggle to custom icons

NiLuJe:
As discussed previously in... a few places, most recently #7304

----

I'm not entirely happy with the UR/UD icons, but I don't really know how to make it any better/clearer?
Stick a 'K' inside the screen, maybe?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7306)


NiLuJe:
Before:

arrows

After:

icons2

poire-z:
Feels a lot better.
Curious to see how it would look with a `Ko` written inside.

(But it will feel odd on a PC screen or on an Android TV :)

NiLuJe:
The two options for a letter inside (with a minor preference for the second, which is reminiscent of how the (old, old) Kindle UI did it:

icons_k

icons_k2

NiLuJe:
With a few graphical snafus fixed:

icons_k3

hius07:
The second picture (without k) is absolutely Okay.
The icon shows how should I hold the device to read the text.
Icons with k inside are confusing.

NiLuJe:
@hius07: I'd tend to agree. It's prettier, but potentially confusing.

hius07:
Menu item might be changed to `Orientation` (as in Kindle)

NiLuJe:
I'm not entirely opposed to it for the user-facing boot of UI, but the code
makes a distinction between orientation and rotation: orientation for
portrait/landscape (inverted or not), and rotation for the four possible
rota.

A strong contender for keeping that distinction is the "lock gyro rotation
to current orientation" option, for instance.

poire-z:
> with a minor preference for the second, which is reminiscent of how the (old, old) Kindle UI did it 

I don't get at all the second one (k not rotating inside the device, following the device): it just means to me nothing on screen is rotating :) you're just yourself rotating the device:
image

----

The "k always being horizontal" (first picture) is a good indicator of how the image will be on the screen after - and may help people who have a different way to think about that:
image

----

Without the "k" is fine too, you get an idea of the text/menu orientation with the surrounding menu itself - but may be the "k" makes the mind gymnastic easier.
image

Frenzie:
> a lowercase o is completely useless in this context, because it looks the same no matter which way you look at it ^^

In most typefaces it's slightly taller than it is wide. /nit

Wrt orientation vs rotation, note that the string `Orientation` is currently not translated and they're basically synonyms.

I prefer the cleaner look without the k.

hius07:
> they're basically synonyms

Okay, I'd say, Rotation is an action, Orientation is a pose.
Since we have arrows in the icons, Rotation is better.
Kindle icons show just a screen (device) pose, so it is Orientation.

poire-z:
> Rotation is an action, Orientation is a pose.
Since we have arrows in the icons, Rotation is better.

That's the main issue, for me: using/expressing actions: it works only in initial portrait mode.

Once you have it rotated 90°:
image
To get back to "normal" mode, you'll hit the button in yellow - which does not indicate an action (or the action to move your device up :) - while the landscape icons indeed show an action.

Showing a state/pose (the current, or the future one) would be the least confusing solution, if we can find one :)

Note that these icons work for a device - but not at all with a PC screen - and for the one with the "k" not always being horizontal, it's actually the reverse of what will happen:
<kbd>image</kbd> would have highlighted: image

NiLuJe:
Can't have context-aware icons in ConfigDialog, AFAIK?

(And, admittedly, having the icons magically move around/change depending on the current orientation might be *moire* confusing?)

poire-z:
> Can't have context-aware icons in ConfigDialog, AFAIK?

Depends on the context:
Static device type: use `item_icons = Device:isHandheld() and { first set } or {second set}`
Dynamic like orientation/rotation: wouldn't be too hard to implement `items_icons_func` :)

I'm fine with your current state of icons (with or without the "k"), showing a device: it's clear enough with the thicker device bottom - for PC and TVs, I guess people don't need this toggle as much as we do on devices, so it's ok if it's a bit less clear.

NiLuJe:
Okay, so, settling on the letter-less variant for now, with a few minor grid-fitting tweaks since the original version:

icons_k6

poire-z:
For the portrait icon, the device being not centered on the underline feels a bit odd :) May be the device frame should be centered in the icon (doesn't matter if there's a big blank area on its left in the icon, and the arrow is near the right)?

image

NiLuJe:
icons_k7

(The Landscape ones were already horizontally centered).

PR #7309, opened 2021-02-18
Adjust wording of coverimage plugin

gbyl:
Standardizes case on `_meta.lua` and improves readability of the plugin

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7309)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -348,7 +348,7 @@ function CoverImage:addToMainMenu(menu_items)
                         end,
                     },
                     {
-                        text = _("Original size"),
+                        text = _("Original image"),
                         checked_func = function()
                             return self.cover_image_background == "none"
                         end,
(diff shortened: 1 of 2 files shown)


plugins/coverimage.koplugin/main.lua
```diff
@@ -349,7 +348,7 @@ function CoverImage:addToMainMenu(menu_items)
                         end,
                     },
                     {
-                        text = _("Original image"),
+                        text = _("Original size"),
```
poire-z, line 351: Are you sure about this change ? The callback seems to be about background.

Frenzie: Something like "original page" might be clearer?

NiLuJe: Given the wider context, I think the current wording is correct (in the sense that it's the image, and only the image).

The wording is not limited to saying "no background" or "keep bg as-is" (or however we word that in ScreenSaver) because, AFAICT, it *also* bypasses scaling, and just dumps the file as-is.

NiLuJe: TL;DR: Yeah, ixnay on that hunk ;).

PR #7310, opened 2021-02-18
show external link as QR code

AntonLee:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7310)

### DIFF
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -661,7 +661,7 @@ function ReaderLink:onGoToExternalLink(link_url)
         end,
     })
     table.insert(alt_handlers_buttons, {
-        text = _("As QR code"),
+        text = _("Show QR code"),
         callback = function()
             UIManager:close(dialog)
             UIManager:show(QRMessage:new{
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -661,7 +661,7 @@ function ReaderLink:onGoToExternalLink(link_url)
         end,
     })
     table.insert(alt_handlers_buttons, {
-        text = _("as QR code"),
+        text = _("As QR code"),
         callback = function()
             UIManager:close(dialog)
             UIManager:show(QRMessage:new{
--- a/frontend/apps/reader/modules/readerlink.lua
+++ b/frontend/apps/reader/modules/readerlink.lua
@@ -11,6 +11,7 @@ local InfoMessage = require("ui/widget/infomessage")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local LinkBox = require("ui/widget/linkbox")
 local Notification = require("ui/widget/notification")
+local QRMessage = require("ui/widget/qrmessage")
 local UIManager = require("ui/uimanager")
 local ffiutil = require("ffi/util")
 local logger = require("logger")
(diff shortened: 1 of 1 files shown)


poire-z:
Screenshot:

<kbd>image</kbd>

Fine by me. May be `As QR code` starting with an uppercase A ?

(I thought I would prefer to have the Cancel as the bigger larger button, but as this is dynamically build and other buttons like Wallabag or Open in browser can be added, and as we can just cancel it by tapping outside, it's fine as it is.)

frontend/apps/reader/modules/readerlink.lua
```diff
@@ -652,6 +653,24 @@ function ReaderLink:onGoToExternalLink(link_url)
...
     local alt_handlers_buttons = {}
+    table.insert(alt_handlers_buttons, {
+        text = _("Copy"),
+        callback = function()
+            UIManager:close(dialog)
+            Device.input.setClipboardText(link_url)
+        end,
+    })
+    table.insert(alt_handlers_buttons, {
+        text = _("As QR code"),
```
Frenzie, line 664: The PR title fits better. Here are some alternatives.
```suggestion
        text = _("URL to QR code"),
```

```suggestion
        text = _("Display QR code"),
```

```suggestion
        text = _("Show QR code"),
```

poire-z: Somehow, "**as** QR code" sounds better to my broken english ears :)
These sounds ok:
```
URL as QR code
Display as QR code
As QR code
```
`Show QR code` or `Show as QR code` a bit less.

localjo: As a native American English speaker, "Show QR Code" sounds the most natural.

Frenzie: I'm definitely also on the `Show QR code` train.

poire-z:
Just wondering: does this "Show QR code" has any use on a device with a webbrowser ?
(You won't scan your phone with another phone to follow a url :)
Shouldn't it not show on Android & SDL ?

localjo:
I use koreader on my e-ink tablet, and scanning a QR code with my android phone would be very useful. I can also imagine it would be useful if I want to share a link with someone else who is in the same room as me, they could just scan my device (phone or tablet) with their phone. My opinion is that it should show on Android.

AntonLee:
I use it quite often, thanks to the developers of koreader for readable
code or it would be harder to do this change by myself. On Android, I
imagine, at least copy option will be useful, in conjunction with
third-party program for qr code creation similar result can be achieved.
But there is already a plugin for this in koreader (qr code from clipboard).

вт, 23 февр. 2021 г., 21:56 Jo Sprague <notifications@github.com>:

frontend/apps/reader/modules/readerlink.lua
AntonLee: Maybe shorter "QR code" would be enough? Probably no need for translation in all languages this way. Also is there a length limit on button text (considering small screens)?

Frenzie:
> Just wondering: does this "Show QR code" has any use on a device with a webbrowser ?

Yes, though I suppose by *this* QR code you mean this one specifically in KOReader rather than a more generic system-wide thing. The answer to that isn't as obvious, but I suppose it might as well be there.

frontend/apps/reader/modules/readerlink.lua
poire-z: The next ones `Add to Wallabag` and `Open in browser` are not smaller than `Show QR code`, so I guess it's fine.

PR #7324, opened 2021-02-20
CoverBrowser: list mode: limit font size of components

poire-z:
Should trigger when per-page <= 6/7, avoids authors and titles to become too big.

Discussed in #7278, a so-called "discussion" that this can't close. Trying it: Closes #7278.
Screenshots before/after in #7278 (good luck finding them in that useless piece of shitcussion).
(Changes the max for file info on the right from 16 to 18 as requested in there - we'll see if that feels too big.)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7324)

### DIFF
--- a/plugins/coverbrowser.koplugin/listmenu.lua
+++ b/plugins/coverbrowser.koplugin/listmenu.lua
@@ -48,7 +48,6 @@ local corner_mark_size = -1
 local corner_mark
 
 local scale_by_size = Screen:scaleBySize(1000000) / 1000000
-local max_fontsize_fileinfo
 
 -- ItemShortCutIcon (for keyboard navigation) is private to menu.lua and can't be accessed,
 -- so we need to redefine it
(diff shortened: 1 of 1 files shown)


NiLuJe (approved):

poire-z:
https://github.com/koreader/koreader/discussions/7278#discussioncomment-391050
> Finally, I will agree with you. I tried the info size 14, 18 and I realized what you said, that it will take valuable space from title/ authors. 

Waiting for more feedback before changing 18 to 16.

NiLuJe:
@poire-z: FWIW, I think the only platform where I tweak the defaults are on Android (tall phone), and nothing jumped out at me (but I'm still at something like 12 items per page, though).

poire-z:
Yes, it should trigger only on 4-6 items per page. Try it on your tall phone and say if, pretending you like that view, you'd be fine with the font size of the size/format/progress (and so, smaller width for author titles, and so, possible really smaller font size for them), or if you'd wish it smaller.
Just tried it on my (not tall, small) phone, and I'm somehow OK with the 18 (should give the same size as the "path" in the top title).

NiLuJe:
Yep, looks fine at 6 (in detailed list view), the date font size essentially matches the one for the author/series lines.

I mean, there's, obviously, a lot of vertical space left blank, but it works, I think?

NiLuJe:
Slightly more marked difference at 4, but still works.

(Note that I possibly already have smaller font sizes, because I have the Series on a dedicated line).

PR #7325, opened 2021-02-21
TouchMenu: Workaround an EPDC race in the hold handler w/ flash_ui

NiLuJe:
Re #7262

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7325)

### DIFF
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -216,7 +216,7 @@ function TouchMenuItem:onHoldSelect(arg, ges)
         UIManager:setDirty(nil, "fast", highlight_dimen)
 
         -- Stupid hack because the EPDC has decided to be race-y in this very specific case.
-        -- Random 100ms is random.
+        -- Empirical evidence suggests using 100ms does the job well enough.
         ffiUtil.usleep(100 * 1000)
         UIManager:forceRePaint()
 
--- a/frontend/ui/widget/touchmenu.lua
+++ b/frontend/ui/widget/touchmenu.lua
@@ -219,7 +219,7 @@ function TouchMenuItem:onHoldSelect(arg, ges)
         --       causing a mild variant of our racey friend the papercut refresh glitch ;).
         --       As it appears to stem from the race between *this* refresh for the highlight and the following writes to the fb,
         --       let the kernel take a breather. It'll yield back to us when it's done.
-        --       Expect it to block for ~150 to 350ms.
+        --       Expect it to block for ~150 to 350ms. Given the context (a hold gesture), we can absorb the latency hit mostly unnnoticed.
         UIManager:waitForVSync()
 
         -- Unhighlight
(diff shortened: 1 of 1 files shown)


NiLuJe:
Ping @yparitcher, although I settled on the nuclear option, so, I don't expect any Kindle-specific weirdness ;).

yparitcher:
fixes hold, but touch is still paper-cutting

NiLuJe:
Any specific places? Because I can't reproduce that (anymore) on Kobo.

yparitcher:
try the version button,
putting a vsync in the tap code helps but putting vsyncs in kind of defeats the purpose.
also none of this helps the top level icon buttons and the up arrow.
it also happens on buttons throught koreader, i just dont use so many buttons so it took a while to notice.
ie. it looks like this is broken everywhere on my pw4.

also a simalar regression from 2020.02 i noticed:
when i dismiss the frontligh dialog i get a black flash on its area, i dont think that happened before. even now i didn't notice it by the skimto or others, just by the frontlight.

NiLuJe:
That was a case of "the H2O is dog slow" ^^.

I can very rarely reproduce minor snafus on the Forma, will see what can be done. On the upside, none of those impede on the readability of the elements, unlike the usual tears.

poire-z:
> try the version button,

Can see the paper cut glitch (with flash_ui enabled) when tapping Version too, on my GloHD. Around the same 70% tearing rate :)

NiLuJe:
> also a simalar regression from 2020.02 i noticed: when i dismiss the frontligh dialog i get a black flash on its area, i dont think that happened before. even now i didn't notice it by the skimto or others, just by the frontlight.

That's by design. The bug is arguably SkimTo *not* using flashui on close ;). (The reasoning behind that is that SkimTo can *only* be shown over the Reader, so it can afford to use partial and risk promotion, because that allows is to take care of the potential for ghosting because partial is REAGL on non-shitty devices; the other widgets don't, so they have to flash to handle ghosting in the cases where they are shown over ReaderUI).

yparitcher:
try also the top level icons. (I think they are buttons not menu items)

NiLuJe:
The topmenu's icons, you mean?

I've never seen it happen there. I've only seen it happens on Buttons, not IconButtons (maaaybe the top chevron, but I can't reproduce that).

yparitcher:
yes, it happens there 80 percent of the time.

NiLuJe:
I'm going to need a colder PW2, then ;).

(Mine's charging, so it's in the sweet spot temp-wise for the screen).

NiLuJe:
I also need more RAM apparently :D.

```
[104522.605908] Out of memory: Kill process 4916 (cvm) score 778 or sacrifice child
```

yparitcher:
this also happened last night when my PW4 was plugged in.

NiLuJe:
Okay, colder definitely helped.

The good news is, it appears that it's really down to a hair's breadth, as simply yielding for 1ms does the trick.

More experiments later :}.

PR #7332, opened 2021-02-21
flash_ui: Workaround potential EPDC races

### LINKED ISSUE #7326: Skim dialog - possible minor flash_ui issue
* KOReader version: 2021.02-20-ge582036_2021-02-21
* Device: Kindle PW2

At first, the new Skim dialog without the header and with the star button looks and works great.

#### Issue
After pressing Next (or Previous) chapter button part of it (vertical line) becomes greyed.

I made a screenshot but the issue is not visible: on the screenshot the full button is black.

NiLuJe:
Followup to https://github.com/koreader/koreader/pull/7325

----

* flash_ui: Yield to the kernel between the HL and the UNHL/CB to let the EPDC do its thing in peace.
* UIManager: Handle nils in task scheduling arguments.
* SkimTo: Use the same, thicker chapter nav icons as ReaderSearch (fix #7326).
* SkimTo: The bookmark toggle button doesn't require a vsync flag.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7332)

### DIFF
--- a/frontend/ui/uimanager.lua
+++ b/frontend/ui/uimanager.lua
@@ -1465,6 +1465,21 @@ function UIManager:waitForVSync()
 end
 
 --[[--
+Yield to the EPDC.
+
+This is a dumb workaround for potential races with the EPDC when we request a refresh on a specific region,
+and then proceed to *write* to the framebuffer, in the same region, very, very, very soon after that.
+
+We basically put ourselves to sleep for a very short amount of time, to let the kernel do its thing.
+]]
+function UIManager:yieldToEPDC()
+    if Device:hasEinkScreen() then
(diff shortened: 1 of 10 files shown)


NiLuJe:
@yparitcher: Right now, I can't reproduce any glitches anymore on Kobo (both Mk. 5 & Mk. 7).

The only thing I can reproduce, but at a much lower rate, is on the PW2, with the TouchMenu highlights (on hold, extreeeeemely rarely on tap, if there's a callback).
So, might go back to a full sync on hold there.

yparitcher (approved):
on menu touch (version) i still get hairline papercuts. hard to notice though (the enire cut is to the left and higher than the text).

PR #7334, opened 2021-02-22
ScreenSaver: Delay footer/header repaint if screensaver_delay is enabled

### LINKED ISSUE #7327: Status bar does not wait for screensaver delay
* KOReader version: v2021.02
* Device: Kobo Aura

#### Issue

With a delay to disable the screensaver image, the status bar appears immediately after waking up the device (i.e. opening the sleep cover), overwriting part of the screensaver image. Then after the set delay the screensaver image disappears (this part is OK).

#### Steps to reproduce

* Enable normal status bar (not tested with "alternate")
* Enable screensaver and "delay when exit from screensaver" (this doesn't sound gramatically correct, by the way)
* Activate the screensaver by suspending (putting to sleep) the device
* Wake up the device
* Look how the status bar "corrupts" the lower part of your nice screensaver image :D

NiLuJe:
Fix #7327

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7334)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -2177,6 +2177,25 @@ function ReaderFooter:refreshFooter(refresh, signal)
 end
 
 function ReaderFooter:onResume()
+    -- Don't repaint the footer until OutOfScreenSaver if screensaver_delay is enabled...
+    local screensaver_delay = G_reader_settings:readSetting("screensaver_delay")
+    if screensaver_delay and (type(screensaver_delay) == "number" or screensaver_delay == "tap") then
+        self._delayed_screensaver = true
+        return
+    end
+
+    -- Force a footer repaint on resume if it was visible
+    self:onUpdateFooter(self.view.footer_visible)
+    self:rescheduleFooterAutoRefreshIfNeeded()
+end
+
+function ReaderFooter:onOutOfScreenSaver()
+    if not self._delayed_screensaver then
+        return
+    end
+
+    self._delayed_screensaver = nil
+
     -- Force a footer repaint on resume if it was visible
     self:onUpdateFooter(self.view.footer_visible)
     self:rescheduleFooterAutoRefreshIfNeeded()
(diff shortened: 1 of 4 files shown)


frontend/ui/elements/screensaver_menu.lua
```diff
@@ -208,7 +208,7 @@ return {
                 }
             },
             {
-                text = _("Delay when exit from screensaver"),
+                text = _("Let the screensaver linger on wakeup"),
```
Frenzie, line 211: rofl

Definitely amusing, but maybe just `leave screensaver after waking up` or something?

Frenzie (approved):

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -2177,6 +2177,25 @@ function ReaderFooter:refreshFooter(refresh, signal)
...
+    self:rescheduleFooterAutoRefreshIfNeeded()
+end
+
+function ReaderFooter:onOutOfScreenSaver()
+    if not self._delayed_screensaver then
+        return
+    end
+
+    self._delayed_screensaver = nil
+
```
poire-z, line 2198: Huge file, more blank lines, more huge file :) (this line does not need to stand out like that)

poire-z (approved):
(Dunno what this screensaver option use case is... Resuming when the lights are out, and wanting to see the cover for a few seconds?)

Jellby:
> (Dunno what this screensaver option use case is... Resuming when the lights are out, and wanting to see the cover for a few seconds?)

In my case, I have nice pictures as screensaver, but I never see them properly, because with a sleep cover they're only shown while the device is covered. With the delay option, I get to enjoy the pictures for a few seconds, it's not like I'm in a rush to start reading anyway.

If I didn't have a sleep cover and had to press some button to wake up, I'd disable the delay indeed.

NiLuJe:
Are we all happy with the reworded options?

frontend/ui/elements/screensaver_menu.lua
```diff
@@ -208,7 +208,7 @@ return {
                 }
             },
             {
-                text = _("Delay when exit from screensaver"),
+                text = _("Keep displaying the screensaver after wakeup"),
```
poire-z, line 211: `Keep screensaver displayed after wakeup` ?
(Somehow, this -ing feels strange for our menus, dunno if we have any :)
Fine with the others.

poire-z (approved):

PR #7335, opened 2021-02-22
[RFC] Pagination UI shenanigans

NiLuJe:
c.f., discussion at the tail end of #3620

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7335)


NiLuJe:
`hold_input` is a bit of a special snowflake, so I'm not super enthused about whatever might be necessary to duplicate it as a `tap_input`...

NiLuJe:
Not super fond of "Single page", but then "Page 1 of 1" feels equally stupid, too.

frontend/ui/widget/keyvaluepage.lua
```diff
@@ -597,7 +597,15 @@ function KeyValuePage:_populateItems()
             end
         end
     end
-    self.page_info_text:setText(T(_("Page %1 of %2"), self.show_page, self.pages))
+
+    -- update page information
+    if self.pages > 1 then
+        self.page_info_text:setText(T(_("Page %1 of %2"), self.show_page, self.pages))
+    else
+        self.page_info_text:setText("Single page");
```
Frenzie, line 605: I know page 1 of 1 looks a bit silly but this

a. isn't much less silly
b. is inconsistent

poire-z: I agree:
"Single page" is more prose and will attract your eyes more easily than a consistent "Page 1 of 1"

frontend/ui/widget/menu.lua
```diff
@@ -971,7 +971,11 @@ function Menu:updatePageInfo(select_number)
             self.page_info_text:setText(FFIUtil.template(_("Page %1 of %2"), self.page, self.page_num))
             self.page_info_text:enable()
         else
-            self.page_info_text:setText("");
+            self.page_info_text:setText("Single page");
+            -- Dirty hackery: Call "disable" to to neuter its input handler...
+            self.page_info_text:disable()
+            -- ...and then fudge the label manually to make sure the text is black...
+            self.page_info_text.label_widget.fgcolor = Blitbuffer.COLOR_BLACK
```
poire-z, line 978: Make that hackery a method of Button ? `Button:disableButLookLikeEnabled()` :/ ?

hius07: It might be enabled for one-page as well, no harm. To simplify the code.

NiLuJe:
Okay, I think I went through most of @hius07's helpful comments in #3620 ;).

To recap:

* Menu/KeyValuePage/ReaderGoTo: Unify the dialogs. (Generally, "Enter page number" as title, and "Go to page" as OK button).
* Allow *tapping* on pagination buttons, too. Added spacers around the text to accommodate for that.
* Disable input handlers when <= 1 pages, while still printing the label in black.
* Always display the chevron icons (except whith no content at all, in Menu).

poire-z:
Tested on my Kobo.
Not superfan on seeing the chevrons on KeyValuePage, even more now that we can configure items per page to be 16 and have most of the interesting ones (Statistics current book, Book info) fit on a single page. But ok, I'll get used to it.

One unrelated issue with KeyValuePage, that may be more noticable with this (as I actually just notice it now) is that, because of the "go up chevron" on the left, which is possibly not displayed, the "Page 1 of 1" may not be centered in the bottom of page (noticable when switching between Book info and ToC). (Might be more noticable because of the chevrons, dunno.)

NiLuJe:
@poire-z: Do you have an example of a KVP with that extra chevron?

I don't recall if OPDS uses Menu or KVP, but it can add a return arrow on the bottom-left corner, and that one doesn't affect alignment.

NiLuJe:
But, yeah, I get what you're seeing:

Stats on top (KVP)
OPDS on bottom (Menu, I guess ;p).

kvp-menu

poire-z:
> Do you have an example of a KVP with that extra up chevron?

`Reading statistics > Time range`  and children show the back/up chevron always
`Reading statistics > Current book` doesn't show it - but it has the same effect on positionning the other chevonand page as in the Time range ones

hius07:
Consistency grows up, thank you!

Looking at the up chevron, I recalled FRs asking for a button in the filebrowser to go up to a parent folder. Maybe someday...

NiLuJe:
@hius07: We have that in Details view when it matters, I don't have any particular interest in the Classic view myself ;).

NiLuJe:
Okay; hopefully the last few commits took care of the KVP layout issue without breaking all the things ^^.

frontend/ui/widget/keyvaluepage.lua
```diff
@@ -435,31 +439,50 @@ function KeyValuePage:init()
...
 
     local padding = Size.padding.large
     self.item_width = self.dimen.w - 2 * padding
+
+    -- Account for the outer FrameContainer's padding'
+    footer.dimen.w = self.item_width
+    page_return.dimen.w = self.item_width
+    -- NOTE: Yep, padding * 1. I'm guessing BottomContainer never pads on one side or something?
+    footer.dimen.h = self.dimen.h - padding
+    page_return.dimen.h = self.dimen.h - padding
```
poire-z, line 484: Haven't checked what kind of things all these are.
But indeed, it seems BottomContainer centers by default:
`x = (self.dimen.x or 0) + math.floor((self.dimen.w - contentSize.w)/2),`

But this 1* padding feel strange.
I guess if you want the chevron+text centered, and this HorizontalSpan(padding)+page_return left aligned, you might need:
```
BottomContainer w = screen_width, containing:
  OverlapGroup, containing:
    LeftContainer or directly WidgetContainer(left_pad+page_return)
    CenterContainer w=screen_width containing chevrons + text
```

NiLuJe: On the upside, it matches menu to the pixel (as far as height is concerned, at least), so, eh ^^.

hius07:
A minor point on buttons sequence:
in reader: `Cancel` `Skim mode` `Go to page`
in filebrowser: `Cancel` `Go to page` `Go to letter`
Maybe, 'Go to page' to be the rightmost everywhere?

NiLuJe:
Yeah, thought about that one, I kinda like it on the edge. I'd be more inclined to actually get rid of the Skim button from there.

I mean, the "Go to" menu entry is right above the "Skim" one already ;).

hius07:
> get rid of the Skim button

As we can switch from the Skim widget to Go to page, it's logical that we can switch back.
Let it be (do we need a word 'mode' in it?)

NiLuJe:
Ah, forgot that bit of interaction with the page number in SkimTo, fair point ;).

"Skim" on its own feels a bit weird on a button. I'm not a fan of the current "Skim mode" either, FWIW, but haven't had any brighter ideas ;).

hius07:
https://github.com/koreader/koreader/issues/7141#issuecomment-773380147 :)

Frenzie:
I don't like skim mode, but it is quite clear.

hius07:
`Cancel` `Skim a book` `Go to page`

Frenzie:
The correct form of that would be something closer to `skim (the/this) document`.

hius07:
There are many `book`s in the ui, so maybe `Skim the book`?
And also change 'Skim document' in the main menu.

Frenzie:
I'd rather just have `Skim` or `Skim mode` to be clear. That was more of an aside.

hius07:
Yeah, I like just `Skim` too.

poire-z:
Some of the UI tweaks of this PR mess with my #7282 . There are still 16 items, but the sizing/spacing is less nice.
It was more noticable/ugly on my Kobo this morning than on the emulator, but here's the diff on the emulator:
Before | After this PR:
<kbd>image</kbd>

hius07:
Left picture is After, I guess?

Frenzie:
Surely left is before?

poire-z:
> Left picture is After, I guess?

Yes (it's showing the chevrons on single page). Corrected above.

frontend/ui/widget/keyvaluepage.lua
```diff
@@ -435,31 +439,50 @@ function KeyValuePage:init()
                     self:goToPage(page)
                 end
             end,
+            ok_text = "Go to page",
         },
+        call_hold_input_on_tap = true,
         bordersize = 0,
-        margin = Screen:scaleBySize(20),
```
poire-z, line 440: Hard to track down, but it's the removal of this that causes the sizing change issue.
Of course, it should be removed so there's the same little/no margin on the bottom as in Menu.
But somehow, this 2*Screen:scaleBySize(20) (or may be a bit less) should be accounted in the available_height computation to get the same results as before (which might have been a hack or just luck...).

poire-z:
Pffff... I get we want to make Menu and KeyValuePage similar - but they both build their UI in different ways, and both in ugly/hacky ways.
Menu doesn't use any padding on the final FrameContainer - but prepend a `VerticalSpan{width=header_padding = Size.padding.large}` to its title.
KeyValuePage do use a `padding = Size.padding.large` (so, on all 4 sides) on the final FrameContainer.
So, Titles look the same - but then the bottom stuff should be hacked differently to look the same too...
So, go add a padding_bottom=0 to it, account for some added space it in available_height, and keep adding hacks/crap... or rewrite both of them in a cleaner way - which I don't want to :(

poire-z:
This might be enough to do it somehow properly:
```diff
--- a/frontend/ui/widget/keyvaluepage.lua
+++ b/frontend/ui/widget/keyvaluepage.lua
@@ -459,14 +459,21 @@ function KeyValuePage:init()
     }

+    local padding = Size.padding.large
+    self.inner_dimen = Geom:new{
+        w = self.dimen.w - 2 * padding,
+        h = self.dimen.h - padding, -- no bottom padding
+    }
+    self.item_width = self.inner_dimen.w
+
     local footer = BottomContainer:new{
-        dimen = self.dimen:copy(),
+        dimen = self.inner_dimen:copy(),
         self.page_info,
     }
     local page_return = BottomContainer:new{
-        dimen = self.dimen:copy(),
+        dimen = self.inner_dimen:copy(),
         WidgetContainer:new{
             dimen = Geom:new{
-                w = Screen:getWidth(),
-                h = self.page_return_arrow:getSize().h,
+                w = self.inner_dimen.w,
+                h = self.return_button:getSize().h,
             },
             self.return_button,
@@ -474,14 +481,4 @@ function KeyValuePage:init()
     }

-    local padding = Size.padding.large
-    self.item_width = self.dimen.w - 2 * padding
-
-    -- Account for the outer FrameContainer's padding
-    footer.dimen.w = self.item_width
-    page_return.dimen.w = self.item_width
-    -- NOTE: Yep, padding * 1. I'm guessing BottomContainer never pads on one side or something?
-    footer.dimen.h = self.dimen.h - padding
-    page_return.dimen.h = self.dimen.h - padding
-
     -- setup title bar
     self.title_bar = KeyValueTitle:new{
@@ -492,7 +489,9 @@ function KeyValuePage:init()
         kv_page = self,
     }
+
     -- setup main content
-    local available_height = self.dimen.h
+    local available_height = self.inner_dimen.h
                          - self.title_bar:getSize().h
+                         - Size.span.vertical_large -- for above page_info (as title_bar adds one itself)
                          - self.page_info:getSize().h
                          - 2*Size.line.thick
@@ -521,5 +520,5 @@ function KeyValuePage:init()
     local content = OverlapGroup:new{
         allow_mirroring = false,
-        dimen = self.dimen:copy(),
+        dimen = self.inner_dimen:copy(),
         VerticalGroup:new{
             align = "left",
@@ -535,4 +534,5 @@ function KeyValuePage:init()
         height = self.dimen.h,
         padding = padding,
+        padding_bottom = 0,
         margin = 0,
         bordersize = 0,
```
(No bottom padding, because the Buttons (the chevrons or the text one) already bring some bottom spacing, which is the only one used in Menu.)

There won't be as much vertical space between the K/Vs and the `Page 1 of 1` as in the "Before" screenshot above - but it was probably too much before (`Screen:scaleBySize(20),`). So, this may keep making the font bigger (I'll see later it it's bearable on my Kobo :)

poire-z:
Somehow, Menu would deserve the same kind of `inner_dimen`, which would account for the added border Menu may get when not fullscreen but is_popout - to solve the positionning/setDirty hack issue we discussed elsewhere.

NiLuJe:
On a related note about Menu and those bottom buttons, on my Forma, the very top of the Buttons actually overlaps the final "separator" line, so you get a few holes in those lines over the buttons ^^.

NiLuJe:
> Somehow, Menu would deserve the same kind of `inner_dimen`, which would account for the added border Menu may get when not fullscreen but is_popout - to solve the positionning/setDirty hack issue we discussed elsewhere.

xref, as I still had the tab open: that was in https://github.com/koreader/koreader/pull/7238#discussion_r571122555 ;).

I may try to fiddle with it tomorrow, if I get more sleep than I did tonight ;).

poire-z:
Well, looks not as small as previously on my Kobo - but it looks fine and balances.
I'll just have to set it to 17 items-per-page to have it like it was before (with then the 17th missing item acting as adding bottom padding on 16 items pages :)

But Menu and KeyValuePage are not alone: CalendarView (that I must have cloned from KeyValuePage I guess) now has the bottom page_info+chevron higher than the others :/

> I may try to fiddle with it tomorrow, if I get more sleep than I did tonight ;).

Sleep long then ! Don't forget Menu leaks into CoverBrowser :/

NiLuJe:
IIRC, CalendarView had its buttons higher even *before* I started tweaking KVP.

poire-z:
> IIRC, CalendarView had its buttons higher even before I started tweaking KVP.

Yes, you didn't touch it I think - It's just another one to consider in this quest of consistency :/

NiLuJe:
@poire-z: So far, this doesn't *seem* to have broken anything, but I'm not *quite* sure how to test the setDirty shenanigans, since I still don't quite understand *where* the !borderless issue was visible ;).

```diff
diff --git a/frontend/ui/widget/menu.lua b/frontend/ui/widget/menu.lua
index 1ad41ab8..49c33bd5 100644
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -51,7 +51,7 @@ function ItemShortCutIcon:init()
     local radius = 0
     local background = Blitbuffer.COLOR_WHITE
     if self.style == "rounded_corner" then
-        radius = math.floor(self.width/2)
+        radius = math.floor(self.width / 2)
     elseif self.style == "grey_square" then
         background = Blitbuffer.COLOR_LIGHT_GRAY
     end
@@ -100,7 +100,7 @@ function MenuCloseButton:init()
     -- diagonally aligned with the top right corner (assuming padding_right=0,
     -- or padding_right = padding_top so the diagonal aligment is preserved).
     local text_size = text_widget:getSize()
-    local text_width_pad = (text_size.h - text_size.w) / 2
+    local text_width_pad = math.floor((text_size.h - text_size.w) / 2)
 
     self[1] = FrameContainer:new{
         bordersize = 0,
@@ -157,7 +157,7 @@ function MenuItem:init()
     self.content_width = self.dimen.w - 2 * Size.padding.fullscreen
     local shortcut_icon_dimen = Geom:new()
     if self.shortcut then
-        shortcut_icon_dimen.w = math.floor(self.dimen.h*4/5)
+        shortcut_icon_dimen.w = math.floor(self.dimen.h * 4/5)
         shortcut_icon_dimen.h = shortcut_icon_dimen.w
         self.content_width = self.content_width - shortcut_icon_dimen.w - Size.span.horizontal_default
     end
@@ -216,7 +216,7 @@ function MenuItem:init()
     }
     local state_indent = self.state and self.state.indent or ""
     local state_container = LeftContainer:new{
-        dimen = Geom:new{w = self.content_width/2, h = self.dimen.h},
+        dimen = Geom:new{w = math.floor(self.content_width / 2), h = self.dimen.h},
         HorizontalGroup:new{
             TextWidget:new{
                 text = state_indent,
@@ -457,8 +457,8 @@ end
 function MenuItem:getGesPosition(ges)
     local dimen = self[1].dimen
     return {
-        x = (ges.pos.x - dimen.x)/dimen.w,
-        y = (ges.pos.y - dimen.y)/dimen.h,
+        x = math.floor((ges.pos.x - dimen.x) / dimen.w),
+        y = math.floor((ges.pos.y - dimen.y) / dimen.h),
     }
 end
 
@@ -590,8 +590,12 @@ function Menu:_recalculateDimen()
     if self.dimen.h > Screen:getHeight() or self.dimen.h == nil then
         self.dimen.h = Screen:getHeight()
     end
+    self.inner_dimen = Geom:new{
+        w = self.dimen.w - 2 * self.border_size,
+        h = self.dimen.h
+    }
     self.item_dimen = Geom:new{
-        w = self.dimen.w,
+        w = self.inner_dimen.w,
         h = Screen:scaleBySize(46),
     }
     local height_dim
@@ -604,7 +608,7 @@ function Menu:_recalculateDimen()
     if self.menu_title and not self.no_title then
         top_height = self.menu_title_group:getSize().h + 2 * Size.padding.small
     end
-    height_dim = self.dimen.h - bottom_height - top_height
+    height_dim = self.inner_dimen.h - bottom_height - top_height
     self.item_dimen.h = math.floor(height_dim / self.perpage)
     self.span_width = math.floor((height_dim - (self.perpage * (self.item_dimen.h ))) / 2 - 1)
     self.page_num = math.ceil(#self.item_table / self.perpage)
@@ -620,6 +624,13 @@ function Menu:init()
     if self.dimen.h > Screen:getHeight() or self.dimen.h == nil then
         self.dimen.h = Screen:getHeight()
     end
+
+    self.border_size = self.is_borderless and 0 or Size.border.window
+    self.inner_dimen = Geom:new{
+        w = self.dimen.w - 2 * self.border_size,
+        h = self.dimen.h
+    }
+
     self.page = 1
 
     self.paths = {}  -- per instance table to trace navigation path
@@ -634,7 +645,7 @@ function Menu:init()
     }
     local menu_title_container = CenterContainer:new{
         dimen = Geom:new{
-            w = self.dimen.w,
+            w = self.inner_dimen.w,
             h = self.menu_title:getSize().h,
         },
         self.menu_title,
@@ -645,12 +656,12 @@ function Menu:init()
         self.path_text = TextWidget:new{
             face = Font:getFace("xx_smallinfofont"),
             text = BD.directory(self.path),
-            max_width = self.dimen.w - 2*Size.padding.small,
+            max_width = self.inner_dimen.w - 2*Size.padding.small,
             truncate_left = true,
         }
         path_text_container = CenterContainer:new{
             dimen = Geom:new{
-                w = self.dimen.w,
+                w = self.inner_dimen.w,
                 h = self.path_text:getSize().h,
             },
             self.path_text,
@@ -668,7 +679,7 @@ function Menu:init()
     end
     -- group for title bar
     self.title_bar = OverlapGroup:new{
-        dimen = {w = self.dimen.w, h = self.menu_title_group:getSize().h},
+        dimen = {w = self.inner_dimen.w, h = self.menu_title_group:getSize().h},
         self.menu_title_group,
     }
     -- group for items
@@ -818,11 +829,11 @@ function Menu:init()
     }
     local body = self.item_group
     local footer = BottomContainer:new{
-        dimen = self.dimen:copy(),
+        dimen = self.inner_dimen:copy(),
         self.page_info,
     }
     local page_return = BottomContainer:new{
-        dimen = self.dimen:copy(),
+        dimen = self.inner_dimen:copy(),
         WidgetContainer:new{
             dimen = Geom:new{
                 w = Screen:getWidth(),
@@ -855,7 +866,7 @@ function Menu:init()
         -- to have this complex Menu, and all widgets based on it,
         -- be mirrored correctly with RTL languages
         allow_mirroring = false,
-        dimen = self.dimen:copy(),
+        dimen = self.inner_dimen:copy(),
         self.content_group,
         page_return,
         footer,
@@ -863,10 +874,10 @@ function Menu:init()
 
     self[1] = FrameContainer:new{
         background = Blitbuffer.COLOR_WHITE,
-        bordersize = self.is_borderless and 0 or Size.border.window,
+        bordersize = self.border_size,
         padding = 0,
         margin = 0,
-        radius = self.is_popout and math.floor(self.dimen.w/20) or 0,
+        radius = self.is_popout and math.floor(self.dimen.w / 20) or 0,
         content
     }
     ------------------------------------------
@@ -1072,12 +1083,6 @@ function Menu:updateItems(select_number)
         local refresh_dimen =
             old_dimen and old_dimen:combine(self.dimen)
             or self.dimen
-        if not self.is_borderless then
-            refresh_dimen = refresh_dimen:copy()
-            local bordersize = Size.border.window
-            refresh_dimen.w = refresh_dimen.w + bordersize*2
-            refresh_dimen.h = refresh_dimen.h + bordersize*2
-        end
         return "ui", refresh_dimen
     end)
 end
@@ -1356,7 +1361,7 @@ end
 function Menu.getItemFontSize(perpage)
     -- Get adjusted font size for the given nb of items per page:
     -- item font size between 14 and 24 for better matching
-    return math.floor(24 - ((perpage - 6)/ 18) * 10 )
+    return math.floor(24 - ((perpage - 6) / 18) * 10)
 end
 
 function Menu.getItemMandatoryFontSize(perpage)
```

poire-z:
> since I still don't quite understand where the !borderless issue was visible

I noticed that (remove my setDirty trick from the other issue to see it again):
- in the emulator, hitting <kbd>F</kbd> show a little Menu with the font names (like in our top TouchMenu, but as a popup Menu)
- File brower > Search > Find a file: enter "t" or "j": shows the Search results in a larger pop out Menu with borders, that you will notice is a bit miscentered (shifted towards right and bottom)

poire-z:
About your above patch, 2 thoughts:
- unlike KeyValuePage, the is no padding excluded from inner_dimen - dunno up to which point you want to make them consistent
- I think CoverBrowser might use self.dimen from Menu - and it may need to be switched to use self.inner_dimen

NiLuJe:
>     * unlike KeyValuePage, the is no padding excluded from inner_dimen - dunno up to which point you want to make them consistent
> 

Yeah, noticed that earlier ;). This actually fixes the popout + border cases (e.g., Calibre, or the Find a file you mentioned above ;)).

>     * I think CoverBrowser might use self.dimen from Menu - and it may need to be switched to use self.inner_dimen

At a quick glance I didn't catch anything there (it reimplements _recalculateDimen), I'll double-check.

NiLuJe:
>     * I think CoverBrowser might use self.dimen from Menu - and it may need to be switched to use self.inner_dimen

It does, but since it's borderless that's harmless ;).

NiLuJe:
Okay, pushed the changes, as I haven't noticed any breakage, and haven't found anything funky after a glance at all the Menu users...

Only thing that's untested is 07c0137, which I guess if the `F` menu in the emu you were talking about? (I don't have an emu build right now ;p).

poire-z:
> It does, but since it's borderless that's harmless ;).

Yes, but that would now be an implementation error - even if it doesn't show :)

> unlike KeyValuePage, there is no padding excluded from inner_dimen - dunno up to which point you want to make them consistent

I meant that currently, in Menu, there is no padding between dimen and inner_dimen - so the padding from the screen edges is added to each sub-element inside inner_dimen (above the title bar, inside the items - but we don't want any below the footer). Which I find a bit ugly.
In KeyValuePage, the padding from screen edges is accounted once, between dimen and inner_dimen. Which I find cleaner.
No real need to make these 2 consistent ! Just mentionning that in case you feel tidy obsessive (don't remember which qualification like this one you applied to you some days ago when dealing with refreshes :)

NiLuJe:
> > It does, but since it's borderless that's harmless ;).
> 
> Yes, but that would now be an implementation error - even if it doesn't show :)
> 

True, that's what I figured, hence the final commit ;).

> > unlike KeyValuePage, there is no padding excluded from inner_dimen - dunno up to which point you want to make them consistent
> 
> I meant that currently, in Menu, there is no padding between dimen and inner_dimen - so the padding from the screen edges is added to each sub-element inside inner_dimen (above the title bar, inside the items - but we don't want any below the footer). Which I find a bit ugly.
> In KeyValuePage, the padding from screen edges is accounted once, between dimen and inner_dimen. Which I find cleaner.
> No real need to make these 2 consistent ! Just mentionning that in case you feel tidy obsessive (don't remember which qualification like this one you applied to you some days ago when dealing with refreshes :)

I'd probably be a tiny bit more obsessed if I were actually familiar with the layout code ;).
But it's somewhat out of my wheelhouse ^^.

I *think* I get what you're saying, that in KVP the layout is clearer (in terms of, well, layout/code) as the padding is applied once, from a "logical" place, while in Menu, it kind of happens through inheritance or something?

poire-z:
> I'd probably be a tiny bit more obsessed if I were actually familiar with the layout code ;).
But it's somewhat out of my wheelhouse ^^.

That's ok :) I'll probably clean that up some day. (Mentionned that in case you had fun getting familiar with the layout code - which I find fun once you're familiar with it, building stuff from scratch - but really time consumming when refactoring/cleaning things up.)

> I think I get what you're saying, that in KVP the layout is clearer (in terms of, well, layout/code) as the padding is applied once, from a "logical" place

Right :)

> while in Menu, it kind of happens through inheritance or something?

Not inheritance, but many manual _somethings_ :)
Because of not applying it outside, the coder of these things had to add it inside each things that needs some padding, so possibly with different padding values :/

L/R global padding, on the side of each items:
https://github.com/koreader/koreader/blob/d243097d75400a27b73f37d7635fca109ce74ca3/frontend/ui/widget/menu.lua#L156-L157

Top global padding, as a span above the title:
https://github.com/koreader/koreader/blob/d243097d75400a27b73f37d7635fca109ce74ca3/frontend/ui/widget/menu.lua#L812-L815

NiLuJe:
Yeah, I was a bit puzzled by that content_width, makes sense now, thanks! ;).

And, yeah, I'd rather not break it any further for now, so, for all intents and purposes, this PR should be ready ;).

frontend/apps/filemanager/filemanagerfilesearcher.lua
```diff
@@ -165,8 +166,8 @@ function FileSearcher:showSearchResults()
         dimen = Screen:getSize(),
     }
     self.search_menu = Menu:new{
-        width = Screen:getWidth()-15,
-        height = Screen:getHeight()-15,
+        width = Screen:getWidth() - (Size.border.window * 4),
+        height = Screen:getHeight() - (Size.border.window * 4),
```
poire-z, line 170: This `Size.border.window * 4` should probably be a new `Size.padding.fullscreen_popout * 2`

frontend/ui/widget/menu.lua
```diff
@@ -590,8 +590,12 @@ function Menu:_recalculateDimen()
     if self.dimen.h > Screen:getHeight() or self.dimen.h == nil then
         self.dimen.h = Screen:getHeight()
     end
+    self.inner_dimen = Geom:new{
+        w = self.dimen.w - 2 * self.border_size,
+        h = self.dimen.h - 2 * self.border_size,
+    }
```
poire-z, line 596: I assume _recalculateDimen is called when the screen size changes or on rotation ? so the need to fix up self.dimen and the new self.inner_dimen ?
And so also the need to have that same thing in CoverBrowser as the one in Menu:init() is not rock solid?

poire-z:
Looks ok, haven't checked if _every_ components you switched from self.dimen to self.item_dimen deserves it, so trusting you :)
Still CalendarView to try to have its footer/chevrons nearer the bottom screen border like in KeyValuePage ? May be my above patch would apply quite fine on it with a little manual help :)

frontend/apps/filemanager/filemanagerfilesearcher.lua
poire-z: Or rather Size.margin.fullscreen_popout (with CSS, margin is outside the borders, padding is inside :)

frontend/ui/widget/menu.lua
NiLuJe: That was my assumption too ;).

poire-z: Which might be wrong :) It's called again in `Menu:updateItems` - would need to see if this updating of items can change geometry. My new assumption would be that they shouldn't, and things like dimen shouldn't change. But ok, let it be :)
Menu:onScreenResize() calles Menu:init().

frontend/ui/size.lua
```diff
@@ -53,6 +53,7 @@ local Size = {
         button = Screen:scaleBySize(2),
         buttontable = Screen:scaleBySize(4),
         fullscreen = Screen:scaleBySize(15),
+        fullscreen_popout = Screen:scaleBySize(3), -- Size.border.window * 2
```
poire-z, line 56: Should probably be in Size.margin rather than Size.padding, cf https://github.com/koreader/koreader/pull/7335#discussion_r582081215

NiLuJe:
Okay, took care of KVP. Content is a *little* taller now (I guess because we've gained one padding worth of available height?).

frontend/ui/widget/menu.lua
NiLuJe: On the other hand, if `dimen` does get updated, I don't see how we could go on using a stale `inner_dimen` anyway?

poire-z: Yes of course. So the question becomes: can self.dimen or Screen:getWidth/height change when this is called :)

poire-z:
Looks ok for me on the emulator and on my Kobo.

With one little sizing issue for the CalendarView, in December 2019 where the additional row has a cell just over the chevron:
<kbd>image</kbd>

Also, just mentionning that, but I think you haven't touched anything related:
After playing with filebrowser, changeing items per page, doing searches, re-opening a document, getting back in a book - and then doing a gesture to close KOReader - I got back to File browser, unresponsive (the kind of touch layer unresponsive glitch I have sometimes on my GloHD). Suspend via power button, wait, resume, got me back to file browser. Then, a gesture to re-open last doc, and:
02/24/21-19:21:36 INFO  Kobo resume: unflagged kernel subsystems for resume, ret: true
```
./luajit: frontend/apps/filemanager/filemanagermenu.lua:134: attempt to index field 'instance' (a nil value)
stack traceback:
    frontend/apps/filemanager/filemanagermenu.lua:134: in function 'handleEvent'
    frontend/ui/widget/container/widgetcontainer.lua:95: in function 'propagateEvent'
    frontend/ui/widget/container/widgetcontainer.lua:113: in function 'handleEvent'
    frontend/dispatcher.lua:608: in function 'execute'
    plugins/gestures.koplugin/main.lua:1110: in function 'handler'
    frontend/ui/widget/container/inputcontainer.lua:255: in function 'handleEvent'
    frontend/ui/uimanager.lua:1129: in function 'sendEvent'
    frontend/ui/uimanager.lua:227: in function '__default__'
    frontend/ui/uimanager.lua:1572: in function 'handleInputEvent'
    frontend/ui/uimanager.lua:1642: in function 'handleInput'
    frontend/ui/uimanager.lua:1690: in function 'run'
    ./reader.lua:298: in main chunk
```
Can't reproduce it :/

poire-z:
Regarding the chevron : aren't our icons supposed to be transparent - and so the drawing of the chevront over the border of the calendar should have no effect?

NiLuJe:
> Regarding the chevron : aren't our icons supposed to be transparent - and so the drawing of the chevront over the border of the calendar should have no effect?

They're flattened before being cached ;).

NiLuJe:
> Also, just mentionning that, but I think you haven't touched anything related:
> After playing with filebrowser, changeing items per page, doing searches, re-opening a document, getting back in a book - and then doing a gesture to close KOReader - I got back to File browser, unresponsive (the kind of touch layer unresponsive glitch I have sometimes on my GloHD). Suspend via power button, wait, resume, got me back to file browser. Then, a gesture to re-open last doc, and:
> 02/24/21-19:21:36 INFO Kobo resume: unflagged kernel subsystems for resume, ret: true
> 
> ```
> ./luajit: frontend/apps/filemanager/filemanagermenu.lua:134: attempt to index field 'instance' (a nil value)
> stack traceback:
>     frontend/apps/filemanager/filemanagermenu.lua:134: in function 'handleEvent'
>     frontend/ui/widget/container/widgetcontainer.lua:95: in function 'propagateEvent'
>     frontend/ui/widget/container/widgetcontainer.lua:113: in function 'handleEvent'
>     frontend/dispatcher.lua:608: in function 'execute'
>     plugins/gestures.koplugin/main.lua:1110: in function 'handler'
>     frontend/ui/widget/container/inputcontainer.lua:255: in function 'handleEvent'
>     frontend/ui/uimanager.lua:1129: in function 'sendEvent'
>     frontend/ui/uimanager.lua:227: in function '__default__'
>     frontend/ui/uimanager.lua:1572: in function 'handleInputEvent'
>     frontend/ui/uimanager.lua:1642: in function 'handleInput'
>     frontend/ui/uimanager.lua:1690: in function 'run'
>     ./reader.lua:298: in main chunk
> ```
> 
> Can't reproduce it :/

I'm unfamiliar with the chain of events, but I supposed a nil guard couldn't hurt?

NiLuJe:
As for the chevrons, this fixes it:

```
diff --git a/plugins/statistics.koplugin/calendarview.lua b/plugins/statistics.koplugin/calendarview.lua
index 1e860ece..64383008 100644
--- a/plugins/statistics.koplugin/calendarview.lua
+++ b/plugins/statistics.koplugin/calendarview.lua
@@ -584,7 +584,7 @@ function CalendarView:init()
     -- At most 6 weeks in a month
     local available_height = self.inner_dimen.h - self.title_bar:getSize().h
                             - self.page_info:getSize().h - self.day_names:getSize().h
-    self.week_height = math.floor((available_height - 5*self.inner_padding) / 6)
+    self.week_height = math.floor((available_height - 7*self.inner_padding) / 6)
     self.day_border = Size.border.default
     if self.show_hourly_histogram then
         -- day num + nb_book_spans + histogram: ceil() as histogram rarely
```

Technically, the border is *outside*, so it feels correct?

(Actual days are still slightly taller than before, FWIW).

NiLuJe:
Yep, it's a `self.main_content` overflow:

At 5:
```
Inner Height:   1894
Title height:   121
Header height:  70
Footer Height:  106
Available Height:       1597
Main_content:   1068
```

I still have 3px unaccounted for at 7:

```
Inner Height:   1894
Title height:   121
Header height:  70
Footer Height:  106
Available Height:       1597
Main_content:   1060
```

```diff
diff --git a/plugins/statistics.koplugin/calendarview.lua b/plugins/statistics.koplugin/calendarview.lua
index 1e860ece..5ae803a3 100644
--- a/plugins/statistics.koplugin/calendarview.lua
+++ b/plugins/statistics.koplugin/calendarview.lua
@@ -584,7 +584,7 @@ function CalendarView:init()
     -- At most 6 weeks in a month
     local available_height = self.inner_dimen.h - self.title_bar:getSize().h
                             - self.page_info:getSize().h - self.day_names:getSize().h
-    self.week_height = math.floor((available_height - 5*self.inner_padding) / 6)
+    self.week_height = math.floor((available_height - 7*self.inner_padding) / 6)
     self.day_border = Size.border.default
     if self.show_hourly_histogram then
         -- day num + nb_book_spans + histogram: ceil() as histogram rarely
@@ -616,6 +616,13 @@ function CalendarView:init()
     self.main_content = VerticalGroup:new{}
     self:_populateItems()
 
+    print("Inner Height:", self.inner_dimen.h)
+    print("Title height:", self.title_bar:getSize().h)
+    print("Header height:", self.day_names:getSize().h)
+    print("Footer Height:", self.page_info:getSize().h)
+    print("Available Height:", available_height)
+    print("Main_content:", self.main_content:getSize().h)
+
     local content = OverlapGroup:new{
         dimen = Geom:new{
             w = self.inner_dimen.w,
```

plugins/statistics.koplugin/calendarview.lua
```diff
@@ -443,14 +443,19 @@ function CalendarView:init()
...
+    self.day_width = math.floor((self.dimen.w - 2*self.outer_padding - 6*self.inner_padding) / 7)
     -- Put back the possible 7px lost in rounding into outer_padding
-    outer_padding = math.floor((self.dimen.w - 7*self.day_width - 6*self.inner_padding) / 2)
-    self.content_width = self.dimen.w - 2*outer_padding
+    self.outer_padding = math.floor((self.dimen.w - 7*self.day_width - 6*self.inner_padding) / 2)
+
+    self.inner_dimen = Geom:new{
+        w = self.dimen.w - 2*self.outer_padding,
+        h = self.dimen.h - self.outer_padding, -- no bottom padding
+    }
```
poire-z, line 457: Isn't there some mess around here, using self.outer_padding (computing from width stuff) for self.inner_dimen.h ?

plugins/statistics.koplugin/calendarview.lua
```diff
@@ -523,25 +528,27 @@ function CalendarView:init()
                 })
             end,
         },
+        call_hold_input_on_tap = true,
         bordersize = 0,
-        margin = Screen:scaleBySize(20),
```
poire-z, line 527: Removing this may cause some stuff to go to near the footer (like previously with KeyValuePage - where I removed some stuff from avalable_height to compensate for this loss).

poire-z:
I think in my computation, I considered that as the inter weeks - so 5 - and I probably accounter for the top (above 1st week) and bottom (below last week) paddings some other way.

plugins/statistics.koplugin/calendarview.lua
```diff
@@ -523,25 +528,27 @@ function CalendarView:init()
...
         self.page_info_spacer,
         self.page_info_last_chev,
     }
 
     local footer = BottomContainer:new{
         dimen = Geom:new{
-            w = self.content_width,
-            h = self.dimen.h,
+            w = self.inner_dimen.w,
+            h = self.inner_dimen.h,
```
poire-z, line 551: (And that's a BottomContainer, so its dimen.h should usually be the screen height.)

poire-z:
Disregard most of the above comments :/ (Hard to get back into all this sizing ...)

> Yep, it's a self.main_content overflow:

OK. Looks like that in CalendarView:_populateItems(), I add:
- one VerticalSpan:new{ width = self.inner_padding } at start, before the first CalendarWeek
- one after each inserted CalendarWeek, so at most 6 others

so that makes indeed 7 self.inner_padding .
So, you must be right.

There's still the removed `margin = Screen:scaleBySize(20)`, that added some spacing above footer, that we may want to still keep a bit of it, doing like I did for KeyValuePage:

```diff
 local available_height = self.inner_dimen.h - self.title_bar:getSize().h
+                             - Size.span.vertical_large -- for above page_info (as title_bar adds one itself)
                              - self.page_info:getSize().h - self.day_names:getSize().h
```

poire-z:
Just 5 becoming `7*self.inner_padding` and **no** added Size.span.vertical_large is just fine on my Kobo.
So :+1: on https://github.com/koreader/koreader/pull/7335#issuecomment-785308157

plugins/statistics.koplugin/calendarview.lua
NiLuJe: Yeah, self.outer_padding gets corrected based on something computed from an estimate of itself first.

NiLuJe: I'm not sure how necessary it is (and how much the estimate and the final result changes, but as it's only done once, I figured I'd let it be ;p).

NiLuJe:
> Just 5 becoming `7*self.inner_padding` and **no** added Size.span.vertical_large is just fine on my Kobo.
> So +1 on [#7335 (comment)](https://github.com/koreader/koreader/pull/7335#issuecomment-785308157)

We might still be missing somewhere and/or it's just accumulated rounding errors, but it works out well enough in practice, and it appears to make sense, so, eh'.

FWIW, I get 7 pixels between the button and the final day's border. `inner_padding` is supposed to be 5px here. 

7_pixels

(Sidebar: trying to take a screenshot of a button highlight without tweaking the code to delay the refresh was a stupid idea :D).

poire-z:
> FWIW, I get 7 pixels between the button and the final day's border. inner_padding is supposed to be 5px here.

Not worried: we do:
`self.week_height = math.floor((available_height - 5*self.inner_padding) / 6)`
So, there's 0 to 5px not distributed, that would be not filled after the the last inner_padding.
So I guess you got 2px lost there, + the normal last inner_padding 5px = 7px.

frontend/ui/widget/menu.lua
NiLuJe: As in, does this (the whole dimen + inner_dimen thing) even belong in _recalcWhatverThisFunctionIsCalled in the first place?

Yeah, definitely a good question, especially if it's not related to rotation/resizes to all ;).

poire-z (approved):

hius07:
One more place to change `ok_text` in the `Enter page number` dialog is sortwidget.

frontend/ui/widget/menu.lua
```diff
@@ -457,8 +457,8 @@ end
 function MenuItem:getGesPosition(ges)
     local dimen = self[1].dimen
     return {
-        x = (ges.pos.x - dimen.x)/dimen.w,
-        y = (ges.pos.y - dimen.y)/dimen.h,
+        x = math.floor((ges.pos.x - dimen.x) / dimen.w),
+        y = math.floor((ges.pos.y - dimen.y) / dimen.h),
```
poire-z, line 461: math.floor() abuse ! (This one feels like déja vu :)
We expect a float between 0 and 1, so we can:
https://github.com/koreader/koreader/blob/03885071b97b63d9025871de2f4c16744056f0be/frontend/apps/reader/modules/readertoc.lua#L715-L730
Currently, in the TOC, you can't follow a TOC entry that is expand'collaps'able by tapping on its right half: it will expand or collapse when tapping anywhere on its width.

Quick check looking at math.floor with git log -p over the commits from February, and it seems this is the only one that is a ratio and should not be math.floor()ed.

NiLuJe: I knew this felt funky the moment I did it, thanks ^

PR #7336, opened 2021-02-22
Legacy Kindle: Actually handle (system) power events

### LINKED ISSUE #7333: Kindle Keyboard 3G US - No screensaver on sleep
* KOReader version: 2021.02
* Device: Kindle Keyboard 3G US

#### Issue
I'm not sure if it's supposed to, but when I flip the switch to sleep the device while koreader is open, there is zero indication that the device is asleep. I can tell that it *is* asleep because none of the buttons work until I hit the switch again, but I would've expected some indication of sleep to happen. I've tried checking every option under Screen->Screensaver to no avail.
#### Steps to reproduce
Freshly install KOReader, start it from KUAL, sleep the device with the switch. Expect to see indication that sleep is enabled, but nothing comes.

NiLuJe:
Otherwise, ScreenSaver handling doesn't work, duh'.
I have no idea how I managed to get that working the last time I tested
it :?.
Possibly I tested the final code on a K4 and not a K3?

Fix #7333

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7336)

### DIFF
--- a/frontend/device/kindle/device.lua
+++ b/frontend/device/kindle/device.lua
@@ -428,6 +428,7 @@ function Kindle2:init()
     }
     self.input.open("/dev/input/event0")
     self.input.open("/dev/input/event1")
+    self.input.open("fake_events")
     Kindle.init(self)
 end
 
@@ -444,6 +445,7 @@ function KindleDXG:init()
     self.keyboard_layout = require("device/kindle/keyboard_layout")
     self.input.open("/dev/input/event0")
     self.input.open("/dev/input/event1")
+    self.input.open("fake_events")
     Kindle.init(self)
 end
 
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
There's no doubling with the fakes?

NiLuJe:
@Frenzie: What ado you mean, exactly?

Frenzie:
I mean that on my phone I missed that this only applies to three specific devices, so I was in effect wondering if on a PW2 or something you wouldn't get both "real" and "fake" events.

NiLuJe:
Oh, FWIW, the effect is fairly similar to other platforms, only the implementation differs: it just generates custom input events for Charging/Discharging/PowerPress/PowerRelease events, so we effectively *always* want it ;).

PR #7341, opened 2021-02-23
Kindle: Unbreak Zelda/Rex

NiLuJe:
Via koreader/koreader-base/pull/1314

Regression since koreader/koreader-base/pull/1312

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7341)


NiLuJe:
This probably also unbreaks night mode, for [similar reasons](https://github.com/NiLuJe/FBInk/blob/96a2cd6a93f5184c595c0e53a844fd883adfd75b/fbink.c#L2422-L2440) as on Kobo Mk. 7.

(a.k.a., why we avoid FORCE_MONOCHROME on those devices ^^).

Issue #7139, opened 2021-01-13
reMarkable 2 wakeup timing


snelg:
>     * I tried various non-zero values in `scheduleIn`. Values as low as 0.1 worked, so I left a little buffer and set it to 0.5.
Possible to make this value customizable?
I had to increase to 1.5 to make it work reliable for my rm2. btw, I had [this fix](https://github.com/toltec-dev/toltec/pull/183) installed, which was delaying the suspend, I believe.

_Originally posted by @wordpr3ss in https://github.com/koreader/koreader/issues/7065#issuecomment-759622353_

snelg:
Dang, didn't consider pre-suspend scripts that might delay the actual suspend longer than usual
Should probably figure out a cleaner solution than my hacky "wait an arbitrary amount of time, and just assume that it was long enough that suspend completed."

Quick recap: after rm2 suspends, when you press the power button to wake up, it does NOT send a message to any of the `/dev/input/eventN`

I'm not an expert on system events and such. I do know that `dmesg` output includes a few particular strings after the system wakes up. So here's a rough pseudocode sketch of a potential still-hacky solution:
(sorry, the pseudocode is more C-ish than Lua-ish)

```
function watchForWakeup() {
  if ('dmesg' output contains "PM: suspend exit") {
    triggerResumeProcesses
  } else {
    scheduleIn(0.5, watchForWakeup)
  }
}

execute('systemctl suspend')
watchForWakeup()
```

Will someone with more reMarkable experience give me some feedback for cleaner ways to watch for system messages? @Eeems maybe?

Frenzie:
As far as wait an arbitrary time goes, it'd probably make more sense to just make it higher rather than customizable.

pazos:
Probably the fix would be out of date when WiFi support is merged. And then your original value would work for all rm2 users.

Also, it seems there're quite a few different patches applied by toltec, oxide and the other Rm hacks available. Including backports, fixes and other workarounds for different problems that might not be known here.

I would suggest to keep the barebones implementation working (ie: KOReader + ddvk's framebuffer) and each launcher/repo/thing that wants to package the app and finds something that doesn't work on barebones open a new issue so we can fix it here.

Otherwise it would be a bit nightmare-ish to keep follow of all changes done on each project and these kind of bugs would be very difficult to repro.

snelg:
> As far as wait an arbitrary time goes, it'd probably make more sense to just make it higher rather than customizable.

[...]

PR #7345, opened 2021-02-24
reMarkable 2 clean wakeup from sleep

snelg:
Solution for #7139 
Also requires a small update to the [Installation on ReMarkable](https://github.com/koreader/koreader/wiki/Installation-on-ReMarkable) wiki page. I don't know if there's a fancy way to coordinate wiki changes with PRs or Releases.
The required wiki change is inserting this after step 3 in the "Manual installation" section:

```
4. **reMarkable 2** only, copy the "wake input" workaround script to systemd:
    ```
    cp -v koreader/rm2-wake-input.sh /lib/systemd/system-sleep/
    ```
```
and then incrementing the remaining existing steps.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7345)


NiLuJe:
Not sure if it's worth going down that road, but I just realized that by bundling the script in the platform folder, it actually ends up in the release package... which means you could arguably install it in the KOReader launch script (we do run as root on rM, right?).

snelg:
Just ran a simple test, and that idea works. Systemd didn't need a reboot/reload/whatever, it just went ahead and ran the `rm2-wake-input.sh` script that my tweaked launch script copied into `/lib/systemd/system-sleep/`

But I'm also not sure if copying scripts into systemd folders during launch is a road we want to go down. I know the scope of hacks/workarounds varies greatly between the different types of devices; are any of the others doing something like this?

NiLuJe:
> But I'm also not sure if copying scripts into systemd folders during launch is a road we want to go down. I know the scope of hacks/workarounds varies greatly between the different types of devices; are any of the others doing something like that during launch?

Nope, none., that's usually bundled with the prerequisites/launcher situation.
At most we bundle configuration (or equivalent) files for launchers to register ourselves with, but that's it.

Eeems:
This likely wont play well with other things in the reMarkable ecosystem. For sure it wont play well with Xochitl if you don't grab key events and it's somehow listening. It might be better to just listen to systemd's resume event through dbus.

snelg:
> For sure it wont play well with Xochitl if you don't grab key events and it's somehow listening.

I tested it and xochitl seemed unaffected (to my surprise)
https://github.com/koreader/koreader/issues/7139#issuecomment-785379777
But yeah, this definitely feels like an ugly hack that probably won't play well with *something* else that people use.

> It might be better to just listen to systemd's resume event through dbus.

I'm happy to keep hacking at this if someone will point me in the right direction. I have no idea how to listen to systemd's resume event through dbus :)

Eeems:
https://github.com/Eeems/oxide/blob/master/applications/system-service/systemapi.h#L66
If you are using Qt you'd do something similar to that.

Likely you'll want to use dbus-monitor and dbus-send instead.

NiLuJe:
c.f., https://github.com/koreader/koreader-base/blob/master/input/input-kindle.h for a similar approach (except with Amazon's slightly-less-insane DBus wrapper).

snelg:
Welp [cracks knuckles], guess I'll have to actually set up a toolchain. So far the tweaks I've done have all been in the .lua files, or direct CLI stuff like this. I haven't had to compile C code yet for this project...

snelg:
Alternate solution: https://github.com/koreader/koreader-base/pull/1318
If that gets approved and merged, then I'll update this PR to completely *remove* the LUA code workaround

snelg:
`koreader-base` version bump, and corresponding rm2 workaround removed from `frontend/device/remarkable/device.lua`

NiLuJe (approved):

PR #7347, opened 2021-02-25
Minor followup to #7335

NiLuJe:
* Unify SortWidget's pagination button (it was using a tap_input, which is why I'd missed it ;)). (https://github.com/koreader/koreader/pull/7335#issuecomment-785628495)
* Menu*: Remove redundant screen-layout updates from `_recalculateDimen`, as it should only be concerned with item layout (https://github.com/koreader/koreader/pull/7335#discussion_r582072032)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7347)

### DIFF
--- a/frontend/ui/widget/menu.lua
+++ b/frontend/ui/widget/menu.lua
@@ -585,19 +585,6 @@ local Menu = FocusManager:new{
 function Menu:_recalculateDimen()
     self.perpage = self.items_per_page or G_reader_settings:readSetting("items_per_page") or self.items_per_page_default
     self.span_width = 0
-    self.dimen.w = self.width
-    self.dimen.h = self.height or Screen:getHeight()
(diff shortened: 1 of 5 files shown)


NiLuJe:
@poire-z: Your analysis of the `_recalculateDimen` stuff appears to hold true, nothing appears to have blown up with it gone ;p.

(And, re-reading how it's used, I would agree with your assessment. Possibly a remnant from back when we didn't handle screen geometry updates sanely in the FM/FM-adjacent stuff?).

poire-z (approved):
The cleaner the better :)

hius07:
Minor issue in page numbers footer:
Reading statistics - Calendar view - empty day - `Page 1 of 0`

NiLuJe:
What do you man by "empty day"?

hius07:
No records for that day

NiLuJe:
Oh, the day view, right :).

PR #7349, opened 2021-02-26
Add OPDS catalog for Library Genesis

avsej:
Library Genesis (libgen) is pretty popular library. Unfortunately it does not have its own OPDS server, but I generated static XML files and hosted them on S3 behind CDN, so they can be accessed pretty fast. I have script to regenerate and update tree from the official SQL dumps of libgen. Right now my tree covers fiction and main libraries of libgen.

I hope koreader users will find it useful.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7349)

### DIFF
--- a/plugins/opds.koplugin/opdsbrowser.lua
+++ b/plugins/opds.koplugin/opdsbrowser.lua
@@ -102,6 +102,10 @@ function OPDSBrowser:init()
           {
              title = "Беларуская палічка",
              url = "https://knihi.com/opds.xml",
+          },
+          {
+             title = "Library Genesis",
+             url = "https://libgen.avsej.download/opds.xml",
           }
         }
         G_reader_settings:saveSetting("opds_servers", servers)


NiLuJe:
Dubious legal status of the source probably makes this a Bad Idea(TM) ;).

avsej:
oh, wow. I didn't know that koreader cares about this matter. You have flibusta already, which is the same kind of library (in terms of legal status)

avsej:
Feel free to close the PR if it does not fit koreader's standards.

Frenzie:
I'll close this one then. But thanks for the PR! ;-)

NiLuJe:
I don't know who added it, but we probably failed to check because we're unfamiliar with the Eastern European internet landscape.

I just happened to have heard about libgen somehow, so it very vaguely rung a bell.

Frenzie:
It snuck in through #4515. It looks to me like a Russian Gutenberg/Archive.org type situation but indeed I'm otherwise unfamiliar with it.

avsej:
@Frenzie, I removed flibusta in #7352

PR #7351, opened 2021-02-26
Get rid of OPDS libraries with content of dubious legal status.

NiLuJe:
Remember, it's life +70 in US & EU, life +50 in Canada.
(It's actually slightly more complex than that, and some periods of
time/countries may use a far longer term. IANAL).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7351)

PR #7352, opened 2021-02-26
Remove flibusta from default list of OPDS catalogs

avsej:
While it is large, popular and open library, its legal status is not clear, so better not add this server by default.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7352)


avsej:
Duplicate of #7351

PR #7353, opened 2021-02-26
KeyValuePages: Handle not having anything to show slightly better

NiLuJe:
By disabling the pagination buttons, and showing a specific label, like Menu (https://github.com/koreader/koreader/pull/7347#issuecomment-786768640)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7353)


poire-z:
"No **choices** available" on this (and in Menu) feels a bit awkward (possibly named "no choice" because nothing to give to onMenuChoice :).
Isn't there a more common/generic wording for this, when nothing shown, no item, book, day ?

Frenzie:
In English perhaps, in other languages that might be more questionable. I.e., you'd need to use a context.

Anyway, a simple `Nothing available`? What exactly are we talking about anyway?

hius07:
`No items`

NiLuJe:
`Nothing to display`?

poire-z:
I prefer a shorter "No items" to a verbose "No items to display" ("to display" feels obvious :) or not enough, because it's not only that there is no item to display, but also that there is no item to tap on, no item to long-press on, no item to look at, no item to sing to :)

poire-z (approved):

Issue #7361, opened 2021-02-27
'Find a file' crashes on ' '


hius07:
* KOReader version: 2021.02-30
* Device: Kindle PW2

#### Issue
KOReader crashes on file search of ' '
#### Steps to reproduce
Filemanager - menu Find a file - enter ' ' (one space) - Current folder
##### `crash.log` (if applicable)
`filemanagerfilesearcher.lua:180: attempt to compare two nil values`

hius07:
Not related to this bug, but maybe can be fixed as well?

Find a file is not case sensitive to latin characters.
But it is case sensitive to non-latin (Cyrillic) chars. This is not good.

BTW, fulltext search in a book is not case sensitive to cyrillic chars.

hius07:
And one more FR for filemanagerfilesearcher:
If we have checked `Show unsupported files` in the filebrowser, we shouldn't check `DocumentRegistry:hasProvider` in the file search.
Tapping on the unsupported file in Search results would call Open with... dialog (where we have Text editor).

NiLuJe:
(Re: non-latin & case)

Unicode case-folding is going to be... somewhat unfun from the Lua side.

I don't recall if we build utf8proc shared, in which case, a bit of FFI magic would be a good start.

yparitcher:
#7362 should fix issue 1 and 3

PR #7362, opened 2021-02-28
FileSearcher: Do not crash on space, show unsupported files if enabled.

yparitcher:
Fixes: #7361

Space is overloaded to show all results. Is this the desired behavior?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7362)

### DIFF
--- a/frontend/apps/filemanager/filemanagerfilesearcher.lua
+++ b/frontend/apps/filemanager/filemanagerfilesearcher.lua
@@ -30,6 +30,9 @@ local FileSearcher = InputContainer:new{
 }
 
 function FileSearcher:readDir()
+    local FileManager = require("apps/filemanager/filemanager")
+    local ReaderUI = require("apps/reader/readerui")
+    local show_unsupported = G_reader_settings:isTrue("show_unsupported")
     self.dirs = {self.path}
     self.files = {}
     while #self.dirs ~= 0 do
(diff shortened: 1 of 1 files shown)


hius07:
> Space is overloaded to show all results. Is this the desired behavior?

Frankly, I see no much sense in that.

yparitcher:
> Space is overloaded to show all results. Is this the desired behavior?

@Frenzie 
Do you still stand by your 6 year old decision?

hius07:
This might work as a filter for supported formats back then, but now we have a checkbox in the file browser.

Frenzie:
> Do you still stand by your 6 year old decision?

I think the idea is that you don't have to open various subfolders to get a
larger overview of everything. I'm going to go with that it sounds useful,
but I have no strong opinion on the matter.

poire-z:
Or switch to one `*` ?  `if keywords == "*" then -- one * to show all files `

hius07:
Tried https://github.com/koreader/koreader/pull/7362/commits/3e9cd4a82c5309d59a0b542e4393c715cddc7bed, strange behaviour with * in the end of the search string, eg:
`q` Found no files, `q*` found several files without any q in the names
`11` Found no files, `11*` found `001.txt`, and so on.

yparitcher:
> Tried [3e9cd4a](https://github.com/koreader/koreader/commit/3e9cd4a82c5309d59a0b542e4393c715cddc7bed), strange behaviour with * in the end of the search string, eg:
> `q` Found no files, `q*` found several files without any q in the names
> `11` Found no files, `11*` found `001.txt`, and so on.

Can you reproduce before this PR? I think this has to do with `*` being a regex and captureing more results.

hius07:
> Can you reproduce before this PR?

Thank you, the commit works fine and these mysterious * results are not caused by the PR.
This for sure is not an issue, just an unexpected strange behaviour.

yparitcher:
caused by `string.find()`
https://github.com/koreader/koreader/blob/3d8d40d7972a41aac68af542f8c2137bb3e342fe/frontend/apps/filemanager/filemanagerfilesearcher.lua#L66

NiLuJe:
Right, it's not a glob, it's not a RE, it's a [lua pattern](https://www.lua.org/pil/20.2.html).

NiLuJe:
Should we enforce plaintext matching? (just pass `true` as `find`'s fourth arg).

Frenzie:
> That would seem to be the intent given the forced lowercasing...

Disagree, upper/lower case distinction is seldom useful while other pattern features are.

poire-z:
(No opinion. Not sure how many users would know Lua patterns syntax :)

NiLuJe:
Same boat ;) (also, not a user of this feature, period ^^).

hius07:
https://github.com/koreader/koreader/wiki/Find-a-file

hius07:
On consistency:
`Find a file` in the main menu
`Search for books by filename` in the input dialog header
`File search` in the Gesture manager

hius07:
As we have `Fulltext search` in reader, I'd propose to change the main menu item and the dialog header to `File search`.
`Search Results` is corresponding (maybe just lower `r`).

PR #7365, opened 2021-02-28
Add explanation for shorten home folder

Frenzie:
Cf. <https://www.mobileread.com/forums/showthread.php?p=4098071#post4098071>.

----------

I'm not quite sure, thoughts?

Screenshot_2021-02-28_11-39-00

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7365)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -5,7 +5,6 @@ local ConfirmBox = require("ui/widget/confirmbox")
 local Device = require("device")
 local Event = require("ui/event")
 local FFIUtil = require("ffi/util")
-local InfoMessage = require("ui/widget/infomessage")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local PluginLoader = require("pluginloader")
 local SetDefaults = require("apps/filemanager/filemanagersetdefaults")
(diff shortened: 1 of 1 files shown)


poire-z:
We have `help_text = "..."` to avoid having to do `UIManager:show(InfoMessage:new...)`.

No opinion about the info and wording, looks ok (even if Kobo specific).
But more help_text is always good :)

Frenzie:
I don't think talking about other possible home folders like I did on the
forum would add much other than verbosity.

I'll update it to help text, but I think there are still many outdated menu
items then.

NiLuJe:
I'd be inclined to use an actual Linux example (e.g., /home/$USER ^^), but it may or may not be any more helpful, so, eh' ^^.

In any case, +1 ;).

NiLuJe:
Actually, we can pull the actual platform-specific from Device.home_dir? (may be nil, though).

Frenzie:
> Actually, we can pull the actual platform-specific value from Device.home_dir? (may be nil, though).

The thought crossed my mind, but it might make localization harder.

hius07:
The help text is clear.

It would be nice if the path in the header changes to `Home` immediately after setting the home folder via long-press on the home icon or via the "+" menu.
`Set as HOME folder` in the "+" menu might be greyed if we are in the Home folder.

PR #7368, opened 2021-02-28
WiFi: Handle inconsistent states a tad better.

### LINKED ISSUE #7358: Wifi checkbox is on, but menu action offers to turn it on again
* KOReader version: v2021.02
* Device: Kobo Aura

#### Issue

After disconnecting from a network (but leaving wifi on) the menu checkbox for wifi is activated, but tapping it opens a dialog to turn it on. The only way to turn it off seems to be connecting to a network again and then turning it off.

#### Steps to reproduce

* Activate wifi
* Connect to a network
* Before the network list disappears (if auto-dismiss is enabled), disconnect from the network
* Dismiss the network list
* The wifi checkbox is on, and tapping it doesn't turn it off

NiLuJe:
Namely, if Wi-Fi is enabled, but the connection is down.

Fix #7358

(Note that we have a Kobo-specific workaround to tear-down Wi-Fi in these cases, that was probably subtly inhibited in most cases by #6424 ^^).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7368)

### DIFF
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -600,6 +600,8 @@ function NetworkMgr:showNetworkMenu(complete_callback)
                 return
             end
         end
+        -- NOTE: Also supports a disconnect_callback, should we use it for something?
+        --       Tearing down Wi-Fi completely when tapping "disconnect" would feel a bit harsh, though...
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
Sure thing ^_^

PR #7369, opened 2021-02-28
ReaderDogEar: Enforce a minimum size, too

### LINKED ISSUE #7364: FR: Minimum (or fixed) size for dogear mark
I like reading with minimalistic margins (the device screen is small enough to waste it with margins). Unfortunately, that makes the "dogear" mark for bookmarks really tiny and hard to see.

I'd like a setting for minimum dogear size, or a fixed size, independent of the margins.

I use bookmarks seldom enough that possible overlapping with the text is not a concern for me, but for robustness, maybe the dogear mark could have a partial opacity, or invert the text underneath?

**Additional context**

Default bookmark with 5-size margins
Screenshot_20210228_112052

15-size boommark with 5-size margins (mockup)
Screenshot_20210228_112754

NiLuJe:
Also, update the actual ReaderDogEar variant of the icon to use a semi-transparent fill for the bottom-left corner.

Fix #7364

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7369)

### DIFF
--- a/resources/icons/mdlight/dogear.alpha.svg
+++ b/resources/icons/mdlight/dogear.alpha.svg
@@ -70,13 +70,9 @@
    style="fill:#ffffff;fill-opacity:1;fill-rule:nonzero;stroke:none;stroke-width:0.12;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:0.999901"
    d="M 48,46 V 0 H 2 Z"
    sodipodi:nodetypes="cccc" /><path
-   style="fill:#ffffff;stroke:#000000;stroke-width:4;stroke-linecap:round;stroke-linejoin:round;stroke-opacity:1;opacity:1;fill-opacity:0"
-   d="M 1.9999949,1.9994296 V 45.999919 H 46.00078"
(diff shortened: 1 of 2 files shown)


NiLuJe:
Haven't checked how the min value looks on truly high-dpi devices (only checked on the H2O, not the Forma or a phone), so, open to more testing ;).

Frenzie:
Meh, 260 vs 300 isn't super meaningful. 600 is though ;)

NiLuJe:
Yeah, i think it ended up being hilariously tiny on my phone, but, I don't really use bookmarks (or my phone for reading, for that matter ;p) so it never really bothered me that much, but it is a low hanging fruit ;).

resources/icons/mdlight/dogear.alpha.svg
```diff
@@ -0,0 +1,78 @@
...
+
+
+
+
+
+	
+
+		
+	</defs>
+
```
poire-z, line 67: Signing your art with blank lines ? :)

NiLuJe: Inkscape doing wierd Inkscape things, I guess ^^.

poire-z (approved):
Not tested but feels ok.

Issue #6054, opened 2020-04-15
(UX) Favorites: add star indicator and availability in History menu


mergen3107:
* KOReader version: 2020.03.02
* Device: Kindle PW3  (5.9.7)

#### Issue
Not an issue by itself, just a small inconvenience. 
I noticed this when I couldn't mark a book as favorite from History screen.

#### Steps to reproduce
Long-tap on a book in different file manager windows. I tested only a few variations.

1) File manager, home folder, "Classic mode": surprisingly there are very few options:
![FileManager_2020-Apr-14_222652](https://user-images.githubusercontent.com/6279855/79298548-707f7800-7eaf-11ea-960c-8eee92a13997.png)

2) File manager, home folder, "Detailed mode with metadata and covers": the most complete among other window choices.
![FileManager_2020-Apr-14_222630](https://user-images.githubusercontent.com/6279855/79298568-7bd2a380-7eaf-11ea-8cd3-9ce1395f09d4.png)

3) History, in "Cover mode" or "Detailed mode with metadata and covers": has no "Add to favorites" button. I understand that there is no need in Cut/Copy/Paste from History, but the lack of favorites is a little sad.
![FileManager_2020-Apr-14_222807](https://user-images.githubusercontent.com/6279855/79298590-8a20bf80-7eaf-11ea-96af-f62e89fb7fcb.png)

Just for the uniformity of the UI, is it possible to make the context menus the same for File Managers display modes, and add "Add to Favorites" button to History window?

Frenzie:
> Just for the uniformity of the UI, is it possible to make the context menus the same for File Managers display modes

I think that only means mark as read & add to favorites. The cover stuff makes no sense without covers.

mergen3107:
Yeah, I mean in the reasonable sense. `Add to favorites` is there in classical mode, but `book description` and `Add to favorites` missing. I don't know if `Refresh cached book information` should apply to classical mode. Does only include cover data and metadata? (I get it that metadata is also doesn't apply here, because only filenames are displayed)

mergen3107:
By the way, forgot to ask: Is it possible to add an indication of items added to favorites? Like the indication that the book was opened before (a small "bookmark" icon in the end of the file name bar), but with a star sign, for example?

poire-z:
[...]

PR #2215, opened 2016-08-11
Automatically connect WIFI after Kobo has been resumed

Hzj-jie:
This change is to add a WIFI_WAS_ON flag in suspend.sh to indicate the original WIFI state. If WIFI was on before suspension, try to reestablish after the device has been resumed.
The resume procedure is running in another process, so it won't delay regular resume.
This change also moved most of WIFI / IP related logic into standalone scripts to share between suspend.sh and kobo/device.lua.

This issue was discussed in #2183, while this is a simple solution to resolve part of the issue. Ideally we should automatically connect to any WIFI APs we known, but this change won't automatically connect to another WIFI AP.

Please have a look, I will squash the commits after review has been finished.

PR #7371, opened 2021-03-01
LuaSettings: Add a method to initialize a setting properly

NiLuJe:
To be used instead of the ubiquitous `readSetting(key) or default` construct, which is subtly broken because it means the default doesn't actually belong to the LuaSettings object without further interaction.

This was, among other things, preventing the ReaderFooter settings from ever getting saved in some cases if starting from a default config.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7371)

### DIFF
--- a/frontend/apps/filemanager/filemanagerbookinfo.lua
+++ b/frontend/apps/filemanager/filemanagerbookinfo.lua
@@ -80,7 +80,7 @@ function BookInfo:show(file, book_props)
                     book_props = doc_settings:readSetting("doc_props")
                 end
                 if not book_props then
-                    -- File last opened before 20170701 may have a "stats" setting
+                    -- File last opened before 20170701 may have a "stats" setting.
                     -- with partial metadata, or empty metadata if statistics plugin
                     -- was not enabled when book was read (we can guess that from
                     -- the fact that stats.page = 0)
--- a/frontend/apps/filemanager/filemanagerbookinfo.lua
+++ b/frontend/apps/filemanager/filemanagerbookinfo.lua
@@ -91,7 +91,7 @@ function BookInfo:show(file, book_props)
                         book_props = stats
                     end
                 end
-                -- Files opened after 20170701 have an accurate "doc_pages" setting
+                -- Files opened after 20170701 have an accurate "doc_pages" setting.
                 local doc_pages = doc_settings:readSetting("doc_pages")
                 if doc_pages and book_props then
                     book_props.pages = doc_pages
--- a/frontend/apps/filemanager/filemanagerbookinfo.lua
+++ b/frontend/apps/filemanager/filemanagerbookinfo.lua
@@ -67,7 +67,7 @@ function BookInfo:show(file, book_props)
     table.insert(kv_pairs, { _("Folder:"), BD.dirpath(filemanagerutil.abbreviate(directory)), separator = true })
 
     -- book_props may be provided if caller already has them available
-    -- but it may lack 'pages', that we may get from sidecar file
+    -- but it may lack "pages", that we may get from sidecar file
     if not book_props or not book_props.pages then
         -- check there is actually a sidecar file before calling DocSettings:open()
         -- that would create an empty sidecar directory
@@ -75,24 +75,24 @@ function BookInfo:show(file, book_props)
             local doc_settings = DocSettings:open(file)
             if doc_settings then
                 if not book_props then
-                    -- Files opened after 20170701 have a 'doc_props' setting with
-                    -- complete metadata and 'doc_pages' with accurate nb of pages
-                    book_props = doc_settings:readSetting('doc_props')
+                    -- Files opened after 20170701 have a "doc_props" setting with
+                    -- complete metadata and "doc_pages" with accurate nb of pages
+                    book_props = doc_settings:readSetting("doc_props")
                 end
                 if not book_props then
-                    -- File last opened before 20170701 may have a 'stats' setting
+                    -- File last opened before 20170701 may have a "stats" setting
                     -- with partial metadata, or empty metadata if statistics plugin
                     -- was not enabled when book was read (we can guess that from
                     -- the fact that stats.page = 0)
-                    local stats = doc_settings:readSetting('stats')
+                    local stats = doc_settings:readSetting("stats")
                     if stats and stats.pages ~= 0 then
                         -- Let's use them as is (which was what was done before), even if
                         -- incomplete, to avoid expensive book opening
                         book_props = stats
                     end
                 end
-                -- Files opened after 20170701 have an accurate 'doc_pages' setting
-                local doc_pages = doc_settings:readSetting('doc_pages')
+                -- Files opened after 20170701 have an accurate "doc_pages" setting
+                local doc_pages = doc_settings:readSetting("doc_pages")
                 if doc_pages and book_props then
                     book_props.pages = doc_pages
                 end
@@ -100,7 +100,7 @@ function BookInfo:show(file, book_props)
         end
     end
 
-    -- If still no book_props (book never opened or empty 'stats'), open the
+    -- If still no book_props (book never opened or empty "stats"), open the
     -- document to get them
     if not book_props then
         local document = DocumentRegistry:openDocument(file)
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -141,18 +141,9 @@ function ReaderDictionary:init()
         end
         logger.dbg("found", #available_ifos, "dictionaries")
 
-        if not G_reader_settings:readSetting("dicts_order") then
-            G_reader_settings:saveSetting("dicts_order", {})
-        end
-
+        self.dicts_order = G_reader_settings:readSetting("dicts_order", {})
         self:sortAvailableIfos()
-
-        if not G_reader_settings:readSetting("dicts_disabled") then
-            -- Create an empty dict for this setting, so that we can
-            -- access and update it directly through G_reader_settings
-            -- and it will automatically be saved.
-            G_reader_settings:saveSetting("dicts_disabled", {})
-        end
+        self.dicts_disabled = G_reader_settings:readSetting("dicts_disabled", {})
     end
     -- Prepare the -u options to give to sdcv the dictionary order and if some are disabled
     self:updateSdcvDictNamesOptions()
@@ -163,11 +154,9 @@ function ReaderDictionary:init()
 end
 
 function ReaderDictionary:sortAvailableIfos()
-    local dicts_order = G_reader_settings:readSetting("dicts_order")
-
     table.sort(available_ifos, function(lifo, rifo)
-        local lord = dicts_order[lifo.file]
-        local rord = dicts_order[rifo.file]
+        local lord = self.dicts_order[lifo.file]
+        local rord = self.dicts_order[rifo.file]
 
         -- Both ifos without an explicit position -> lexical comparison
         if lord == rord then
@@ -468,9 +457,8 @@ function ReaderDictionary:getNumberOfDictionaries()
     local nb_available = #available_ifos
     local nb_enabled = 0
     local nb_disabled = 0
-    local dicts_disabled = G_reader_settings:readSetting("dicts_disabled")
     for _, ifo in pairs(available_ifos) do
-        if dicts_disabled[ifo.file] then
+        if self.dicts_disabled[ifo.file] then
             nb_disabled = nb_disabled + 1
         else
             nb_enabled = nb_enabled + 1
@@ -526,7 +514,7 @@ end
 
 function ReaderDictionary:showDictionariesMenu(changed_callback)
     -- Work on local copy, save to settings only when SortWidget is closed with the accept button
-    local dicts_disabled = util.tableDeepCopy(G_reader_settings:readSetting("dicts_disabled"))
+    local dicts_disabled = util.tableDeepCopy(self.dicts_disabled)
 
     local sort_items = {}
     for _, ifo in pairs(available_ifos) do
@@ -552,6 +540,7 @@ function ReaderDictionary:showDictionariesMenu(changed_callback)
         callback = function()
             -- Save local copy of dicts_disabled
             G_reader_settings:saveSetting("dicts_disabled", dicts_disabled)
+            self.dicts_disabled = dicts_disabled
 
             -- Write back the sorted items to dicts_order
             local dicts_order = {}
@@ -559,6 +548,7 @@ function ReaderDictionary:showDictionariesMenu(changed_callback)
                 dicts_order[sort_item.ifo.file] = i
             end
             G_reader_settings:saveSetting("dicts_order", dicts_order)
+            self.dicts_order = dicts_order
 
             self:sortAvailableIfos()
 
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -58,9 +58,6 @@ end
 local ReaderDictionary = InputContainer:new{
     data_dir = nil,
     dict_window_list = {},
-    disable_lookup_history = G_reader_settings:isTrue("disable_lookup_history"),
-    dicts_order = G_reader_settings:readSetting("dicts_order", {}),
-    dicts_disabled = G_reader_settings:readSetting("dicts_disabled", {}),
     lookup_msg = _("Searching dictionary for:\n%1"),
 }
 
@@ -98,6 +95,10 @@ local function getDictionaryFixHtmlFunc(path)
 end
 
 function ReaderDictionary:init()
+    self.disable_lookup_history = G_reader_settings:isTrue("disable_lookup_history")
+    self.dicts_order = G_reader_settings:readSetting("dicts_order", {})
+    self.dicts_disabled = G_reader_settings:readSetting("dicts_disabled", {})
+
     self.ui.menu:registerToMainMenu(self)
     self.data_dir = STARDICT_DATA_DIR or
         os.getenv("STARDICT_DATA_DIR") or
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -1048,8 +1048,9 @@ function ReaderDictionary:onReadSettings(config)
     if #self.preferred_dictionaries > 0 then
         self:updateSdcvDictNamesOptions()
     end
-    self.disable_fuzzy_search = config:readSetting("disable_fuzzy_search")
-    if self.disable_fuzzy_search == nil then
+    if config:has("disable_fuzzy_search") then
+        self.disable_fuzzy_search = config:isTrue("disable_fuzzy_search")
+    else
         self.disable_fuzzy_search = G_reader_settings:isTrue("disable_fuzzy_search")
     end
 end
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -539,6 +539,7 @@ function ReaderDictionary:showDictionariesMenu(changed_callback)
         callback = function()
             -- Save local copy of dicts_disabled
             G_reader_settings:saveSetting("dicts_disabled", dicts_disabled)
+            -- Update our own reference to that setting
             self.dicts_disabled = dicts_disabled
 
             -- Write back the sorted items to dicts_order
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -1109,13 +1109,13 @@ The current default (★) is enabled.]])
             return disable_fuzzy_search and _("Disable (★)") or _("Disable")
         end,
         choice1_callback = function()
-            G_reader_settings:saveSetting("disable_fuzzy_search", true)
+            G_reader_settings:makeTrue("disable_fuzzy_search")
         end,
         choice2_text_func = function()
             return disable_fuzzy_search and _("Enable") or _("Enable (★)")
         end,
         choice2_callback = function()
-            G_reader_settings:saveSetting("disable_fuzzy_search", false)
+            G_reader_settings:makeFalse("disable_fuzzy_search")
         end,
     })
 end
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -537,18 +537,17 @@ function ReaderDictionary:showDictionariesMenu(changed_callback)
         title = _("Manage installed dictionaries"),
         item_table = sort_items,
         callback = function()
-            -- Save local copy of dicts_disabled
-            G_reader_settings:saveSetting("dicts_disabled", dicts_disabled)
-            -- Update our own reference to that setting
+            -- Update both references to point to that new object
             self.dicts_disabled = dicts_disabled
+            G_reader_settings:saveSetting("dicts_disabled", self.dicts_disabled)
 
             -- Write back the sorted items to dicts_order
             local dicts_order = {}
             for i, sort_item in pairs(sort_items) do
                 dicts_order[sort_item.ifo.file] = i
             end
-            G_reader_settings:saveSetting("dicts_order", dicts_order)
             self.dicts_order = dicts_order
+            G_reader_settings:saveSetting("dicts_order", self.dicts_order)
 
             self:sortAvailableIfos()
 
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -59,6 +59,8 @@ local ReaderDictionary = InputContainer:new{
     data_dir = nil,
     dict_window_list = {},
     disable_lookup_history = G_reader_settings:isTrue("disable_lookup_history"),
+    dicts_order = G_reader_settings:readSetting("dicts_order", {}),
+    dicts_disabled = G_reader_settings:readSetting("dicts_disabled", {}),
     lookup_msg = _("Searching dictionary for:\n%1"),
 }
 
@@ -140,10 +142,7 @@ function ReaderDictionary:init()
             end
         end
         logger.dbg("found", #available_ifos, "dictionaries")
-
-        self.dicts_order = G_reader_settings:readSetting("dicts_order", {})
         self:sortAvailableIfos()
-        self.dicts_disabled = G_reader_settings:readSetting("dicts_disabled", {})
     end
     -- Prepare the -u options to give to sdcv the dictionary order and if some are disabled
     self:updateSdcvDictNamesOptions()
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -1853,7 +1853,7 @@ function ReaderFooter:getAvgTimePerPage()
 end
 
 function ReaderFooter:getDataFromStatistics(title, pages)
-    local sec = "N/A"
+    local sec = _("N/A")
     local average_time_per_page = self:getAvgTimePerPage()
     if average_time_per_page then
         if self.settings.duration_format == "classic" then
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -2061,9 +2061,9 @@ end
 
 function ReaderFooter:onReadSettings(config)
     if not self.ui.document.info.has_pages then
-        local h_margins = config:readSetting("copt_h_page_margins") or
-            G_reader_settings:readSetting("copt_h_page_margins") or
-            DCREREADER_CONFIG_H_MARGIN_SIZES_MEDIUM
+        local h_margins = config:readSetting("copt_h_page_margins")
+                       or G_reader_settings:readSetting("copt_h_page_margins")
+                       or DCREREADER_CONFIG_H_MARGIN_SIZES_MEDIUM
         self.book_margins_footer_width = math.floor((h_margins[1] + h_margins[2])/2)
     end
 end
--- a/frontend/apps/reader/modules/readerkoptlistener.lua
+++ b/frontend/apps/reader/modules/readerkoptlistener.lua
@@ -16,15 +16,17 @@ end
 
 function ReaderKoptListener:onReadSettings(config)
     -- normal zoom mode is zoom mode used in non-reflow mode.
-    local normal_zoom_mode = config:readSetting("normal_zoom_mode") or
-                    G_reader_settings:readSetting("zoom_mode") or "page"
+    local normal_zoom_mode = config:readSetting("normal_zoom_mode")
+                          or G_reader_settings:readSetting("zoom_mode")
+                          or "page"
     normal_zoom_mode = util.arrayContains(ReaderZooming.available_zoom_modes, normal_zoom_mode)
-                    and normal_zoom_mode
+                   and normal_zoom_mode
                     or "page"
     self.normal_zoom_mode = normal_zoom_mode
     self:setZoomMode(normal_zoom_mode)
-    self.document.configurable.contrast = config:readSetting("kopt_contrast") or
-                    G_reader_settings:readSetting("kopt_contrast") or 1.0
+    self.document.configurable.contrast = config:readSetting("kopt_contrast")
+                                       or G_reader_settings:readSetting("kopt_contrast")
+                                       or 1.0
     self.ui:handleEvent(Event:new("GammaUpdate", 1/self.document.configurable.contrast))
     -- since K2pdfopt v2.21 negative value of word spacing is also used, for config
     -- compatability we should manually change previous -1 to a more reasonable -0.2
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -299,7 +299,7 @@ end
 
 function ReaderRolling:onSaveSettings()
     -- remove last_percent config since its deprecated
-    self.ui.doc_settings:saveSetting("last_percent", nil)
+    self.ui.doc_settings:delSetting("last_percent")
     self.ui.doc_settings:saveSetting("last_xpointer", self.xpointer)
     -- in scrolling mode, the document may already be closed,
     -- so we have to check the condition to avoid crash function self:getLastPercent()
@@ -398,14 +398,14 @@ function ReaderRolling:addToMainMenu(menu_items)
                     return inverse_reading_order and _("LTR") or _("LTR (★)")
                 end,
                 choice1_callback = function()
-                     G_reader_settings:saveSetting("inverse_reading_order", false)
+                     G_reader_settings:makeFalse("inverse_reading_order")
                      if touchmenu_instance then touchmenu_instance:updateItems() end
                 end,
                 choice2_text_func = function()
                     return inverse_reading_order and _("RTL (★)") or _("RTL")
                 end,
                 choice2_callback = function()
-                    G_reader_settings:saveSetting("inverse_reading_order", true)
+                    G_reader_settings:makeTrue("inverse_reading_order")
                     if touchmenu_instance then touchmenu_instance:updateItems() end
                 end,
             })
@@ -1351,7 +1351,7 @@ Note that %1 (out of %2) xpaths from your bookmarks and highlights have been nor
         }},
         cancel_text = _("Not for this book"),
         cancel_callback = function()
-            self.ui.doc_settings:saveSetting("cre_keep_old_dom_version", true)
+            self.ui.doc_settings:makeTrue("cre_keep_old_dom_version")
         end,
         ok_text = _("Upgrade now"),
         ok_callback = function()
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -139,9 +139,9 @@ function ReaderRolling:onReadSettings(config)
     -- and highlights with old XPATHs.
     -- (EPUB will use the same correct DOM code no matter what DOM version
     -- we request here.)
-    if not config:readSetting("cre_dom_version") then
+    if config:hasNot("cre_dom_version") then
         -- Not previously set, guess which DOM version to use
-        if config:readSetting("last_xpointer") then
+        if config:has("last_xpointer") then
             -- We have a last_xpointer: this book was previously opened
             -- with possibly a very old version: request the oldest
             config:saveSetting("cre_dom_version", self.ui.document:getOldestDomVersion())
@@ -166,7 +166,7 @@ function ReaderRolling:onReadSettings(config)
         self.ui.typeset:ensureSanerBlockRenderingFlags()
         -- And check if we can migrate to a newest DOM version after
         -- the book is loaded (unless the user told us not to).
-        if not config:readSetting("cre_keep_old_dom_version") then
+        if config:nilOrFalse("cre_keep_old_dom_version") then
             self.ui:registerPostReadyCallback(function()
                 self:checkXPointersAndProposeDOMVersionUpgrade()
             end)
@@ -209,13 +209,13 @@ function ReaderRolling:onReadSettings(config)
             end
         end
     end
-    self.show_overlap_enable = config:readSetting("show_overlap_enable")
-    if self.show_overlap_enable == nil then
+    self.show_overlap_enable = config:isTrue("show_overlap_enable")
+    if config:hasNot("show_overlap_enable") then
         self.show_overlap_enable = DSHOWOVERLAP
     end
 
-    self.inverse_reading_order = config:readSetting("inverse_reading_order")
-    if self.inverse_reading_order == nil then
+    self.inverse_reading_order = config:isTrue("inverse_reading_order")
+    if config:hasNot("inverse_reading_order") then
         self.inverse_reading_order = G_reader_settings:isTrue("inverse_reading_order")
     end
 
@@ -229,8 +229,8 @@ function ReaderRolling:onReadSettings(config)
         G_reader_settings:readSetting("copt_visible_pages") or 1
     self.ui.document:setVisiblePageCount(self.visible_pages)
 
-    self.hide_nonlinear_flows = config:readSetting("hide_nonlinear_flows")
-    if self.hide_nonlinear_flows == nil then
+    self.hide_nonlinear_flows = config:isTrue("hide_nonlinear_flows")
+    if config:hasNot("hide_nonlinear_flows") then
         self.hide_nonlinear_flows = G_reader_settings:isTrue("hide_nonlinear_flows")
     end
     self.ui.document:setHideNonlinearFlows(self.hide_nonlinear_flows)
@@ -645,7 +645,7 @@ function ReaderRolling:onGotoXPointer(xp, marker_xp)
     --   followed_link_marker = <number>: removed after <number> seconds
     -- (no real need for a menu item, the default is the finest)
     local marker_setting = G_reader_settings:readSetting("followed_link_marker")
-    if marker_setting == nil then
+    if G_reader_settings:hasNot("followed_link_marker") then
         marker_setting = 1 -- default is: shown and removed after 1 second
     end
 
@@ -1281,7 +1281,7 @@ function ReaderRolling:checkXPointersAndProposeDOMVersionUpgrade()
         -- (see ReaderTypeset:onReadSettings() for the logic of block_rendering_mode)
         local g_block_rendering_mode = G_reader_settings:readSetting("copt_block_rendering_mode")
         if g_block_rendering_mode ~= 0 then -- default is not "legacy"
-            if not g_block_rendering_mode then -- nil means: use default
+            if G_reader_settings:hasNot("copt_block_rendering_mode") then -- nil means: use default
                 g_block_rendering_mode = 3 -- default in ReaderTypeset:onReadSettings()
             end
             -- This setting is actually saved by self.ui.document.configurable
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -209,13 +209,15 @@ function ReaderRolling:onReadSettings(config)
             end
         end
     end
-    self.show_overlap_enable = config:isTrue("show_overlap_enable")
-    if config:hasNot("show_overlap_enable") then
+    if config:has("show_overlap_enable") then
+        self.show_overlap_enable = config:isTrue("show_overlap_enable")
+    else
         self.show_overlap_enable = DSHOWOVERLAP
     end
 
-    self.inverse_reading_order = config:isTrue("inverse_reading_order")
-    if config:hasNot("inverse_reading_order") then
+    if config:has("inverse_reading_order") then
+        self.inverse_reading_order = config:isTrue("inverse_reading_order")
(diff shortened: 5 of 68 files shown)


NiLuJe:
Draft because I need to audit the numerous instances of this construct, and fix/simplify it.

(Some modules evaded the pitfall by doing an explicit `saveSetting(key) later on, but that's effectively redundant, if the ownership of the key is sane from the get go, it gets handled by the usual flush on the object itself...).

NiLuJe:
Also open to comments about the name of the method itself... :}

poire-z:
readSetting or getSetting, hard to guess why use one or the other

readSettingOrSaveDefault
readSettingOrDefault
readTableSetting
?
Is the issue only with a table subsetting ?

NiLuJe:
Nope, technically also applies to simple values ;).

I could just take over readSetting and simply update its API without creating a new method...

poire-z:
> I could just take over readSetting and simply update its API without creating a new method...

That's a lot of work :/ but it would make things clearer/saner/simpler/more obvious for the future.

NiLuJe:
Well, not necessarily, I can simply ignore the "new" codepath if the new default arg is nil?

NiLuJe:
Like that :}

poire-z:
> Like that :}

That's the fun part. The "lots of work" part is updating ALL the `readSetting("blah") or "blue"` to  `readSetting("blah", "blue")` :)
(But for the sake of git history/blame, may be we can do that only when we meet again such parts of the code?)

NiLuJe:
I'll do a pass on stuff where it actually matters/allows to get rid of redundant calls ;).

(e.g., the recent commits on statistics).

poire-z (approved):

Frenzie (approved):
Looks fine, but NB I did *not* read through all the little changes in detail except for the main function.

NiLuJe:
> Nope, technically also applies to simple values ;).

Well, the main use-case is actually limited to stuff that's assigned by reference, not by values.

So, here, mostly exclusively tables.

Made that clearer in the docs, and stopped randomly breaking stuff that's assigned by value ^^.

(This will make the review easier. This also means I need to stop now and pick it up again tomorrow after a good night's sleep if I want to avoid breaking stuff :D).

plugins/terminal.koplugin/main.lua
```diff
@@ -293,13 +293,11 @@ function Terminal:editName(item)
 end
 
 function Terminal:deleteShortcut(item)
-    local shortcuts = {}
-    for _, element in ipairs(self.shortcuts) do
-        if element.text ~= item.text and element.commands ~= item.commands then
-            table.insert(shortcuts, element)
+    for i, element in ipairs(self.shortcuts) do
+        if element.text == item.text and element.commands == item.commands then
+            table.remove(self.shortcuts, i)
```
poire-z, line 298: Sure this is ok, ipairs & table.remove ?
Usually, we're either break'ing after the first remove, or iterating in reverse (`for i=#table, 1, -1 do`).

NiLuJe: Oh, right, of course ;o).

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -770,14 +770,14 @@ function ReaderTypography:onReadSettings(config)
...
-    if self.text_lang_tag then
+    if config:has("text_lang") then
         logger.dbg("Typography lang: using", self.text_lang_tag, "from doc settings")
         self.ui.document:setTextMainLang(self.text_lang_tag)
         return
     end
     -- Use the one manually set as default (with Hold)
     self.text_lang_tag = G_reader_settings:readSetting("text_lang_default")
-    if self.text_lang_tag then
+    if G_reader_settings:has("text_lang_default") then
```
poire-z, line 780: Worth it, looking twice at G_reader_settings ?
I find it a bit more obscure than the previous code.
Same in a next commit with a docsetting config:
```diff
    self.show_page_labels = config:readSetting("pagemap_show_page_labels")
-   if self.show_page_labels == nil then
+   if config:hasNot("pagemap_show_page_labels") then
        self.show_page_labels = G_reader_settings:nilOrTrue("pagemap_show_page_labels")
    end
```
Or remove the line before and do the read in the `then`. But even then... It's not needed if you use :nilOrTrue, or redundant.
And in the text_lang case, I'd rather explicitely assign something, even if nil, than just not assign anything (and not override stuff that would previously be there, even if it might not happen in this context).

NiLuJe: I find that it actually more clearly indicates the intent:

e.g., DocSetting doesn't contain that key => check elsewhere.

NiLuJe: I'm not sure I get the text_lang comment, because it doesn't affect assignments at all :?.

(And nearly all of those should be functionally identical cosmetic tweaks).

NiLuJe: Another way to write (most of) those DocSettings -> G_reader_settings constructs could perhaps be made clearer with:

```
if doc:has(key) then
  self.var = doc:readSetting(key)
else
  self.var = G:readSetting(key)
end
```

poire-z: > I find that it actually more clearly indicates the intent:
e.g., The DocSettings object doesn't contain that key => check elsewhere.

> Speaking of, (most of) those DocSettings -> G_reader_settings constructs could perhaps be made clearer with something like:

Yes, that's clearer, in that "DocSettings -> G_reader_settings"  case.

But the ones like this one are not clearer:
```
1    self.text_lang_tag = G_reader_settings:readSetting("text_lang_fallback")
2    if G_reader_settings:has("text_lang_fallback") then
3        self.ui.document:setTextMainLang(self.text_lang_tag)
4        return
5    end
```

1 Read a value from the far away setting
2 Check again if there is a value in the far away setting
3 Use the value fetched at 1

I find it more explicite as it was:
```
1    self.text_lang_tag = G_reader_settings:readSetting("text_lang_fallback")
2    if self.text_lang_tag then
3        self.ui.document:setTextMainLang(self.text_lang_tag)
4        return
5    end
```

1 Read a value from the far away setting
2 Check that the value is usable
3 Use this value

> I'm not sure I get the text_lang comment, because it doesn't affect assignments at all :?

I meant that _if_ you were to remove/move the step 1 in this sample to not have 2 calls (like you have done elsewhere), you'd get:

```
2    if G_reader_settings:has("text_lang_fallback") then
1        self.text_lang_tag = G_reader_settings:readSetting("text_lang_fallback")
3        self.ui.document:setTextMainLang(self.text_lang_tag)
4        return
5    end
```
You would not be reseting an existing `self.text_lang_tag = "tagada"` (which, granted, does not matter in most of your PRs where you do things early and where we're sure there is no previous value - but as a general principle I prefer assigning nil and be sure there's a reset.)

> that specific case is a fairly gratuitous use of that method as a sort of syntactic sugar

This syntactic sugar assumes you and the reader know `has()` is mostly a check for the thing not being `nil` - and assuming it's usable. I rather have that clear *here* (as it's really simple) than having to look *there* what `has()` really does - because if we are indeed calling it, it must do something more complicated than that :)

NiLuJe: > > I find that it actually more clearly indicates the intent:
> > e.g., The DocSettings object doesn't contain that key => check elsewhere.
> 
> > Speaking of, (most of) those DocSettings -> G_reader_settings constructs could perhaps be made clearer with something like:
> 
> Yes, that's clearer, in that "DocSettings -> G_reader_settings" case.
> 

Cool will do ;).

> I find it more explicit as it was:
> 
> ```
> 1    self.text_lang_tag = G_reader_settings:readSetting("text_lang_fallback")
> 2    if self.text_lang_tag then
> 3        self.ui.document:setTextMainLang(self.text_lang_tag)
> 4        return
> 5    end
> ```

Fair enough, like I said in the edit, that *was* fairly gratuitous ^^. (And inconsistent, since I left quite a few of those as-is anyway).

> 
> > I'm not sure I get the text_lang comment, because it doesn't affect assignments at all :?
> 
> I meant that _if_ you were to remove/move the step 1 in this sample to not have 2 calls (like you have done elsewhere), you'd get:
> 
> ```
> 2    if G_reader_settings:has("text_lang_fallback") then
> 1        self.text_lang_tag = G_reader_settings:readSetting("text_lang_fallback")
> 3        self.ui.document:setTextMainLang(self.text_lang_tag)
> 4        return
> 5    end
> ```
> 
> You would not be reseting an existing `self.text_lang_tag = "tagada"` (which, granted, does not matter in most of this PR where you do things early and where we're sure there is no previous value - but as a general principle I prefer assigning nil and be sure there's a reset.)

Yeah, I finally figured that's what you meant ^^.
And while I definitely agree with you, if I were to play devil's advocate, I would say that this is Lua: accessing an uninitialized or undefined variable already returns `nil` ;o).

> 
> > that specific case is a fairly gratuitous use of that method as a sort of syntactic sugar
> 
> This syntactic sugar assumes you and the reader know `has()` is mostly a check for the thing not being `nil` - and assuming this means it's usable. I rather have that clear _here_ (as it's really simple) than having to look _there_ what `has()` really does - because if we are indeed calling it, it must do something more complicated than that :)

Well, it just does what it says on the tin ;o).

NiLuJe (commented):
> Reviewed 1 of 1 files at r4, 55 of 55 files at r6.
> Dismissed @poire-z from a discussion.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7371)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[plugins/terminal.koplugin/main.lua, line 298 at r5](https://reviewable.io/reviews/koreader/koreader/7371#-MUk38asqrA1rzGI8OlH-r5-298:-MUtN4P4DJHMj42tjmK3:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/2e040e509bc7ddf982b832addd8d36bc107b9bd9/plugins/terminal.koplugin/main.lua#L298)):*

Oh, right, of course ;o).
</blockquote></details>

Done.

frontend/apps/reader/modules/readertoc.lua
```diff
@@ -137,7 +137,7 @@ end
 function ReaderToc:fillToc()
     if self.toc then return end
     if self.ui.document:canHaveAlternativeToc() then
-        if self.ui.doc_settings:readSetting("alternative_toc") then
+        if self.ui.doc_settings:has("alternative_toc") then
```
poire-z, line 140: You had to check (and so I had to) that I indeed did `self.ui.doc_settings:delSetting("alternative_toc")` when toggling it - before doing this change.
Isn't it safer sometimes to just trust readSetting, or use isTrue (which would work whether I have used delSetting or saveSetting(false)) ?

frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -56,63 +59,69 @@ function ReaderTypeset:onReadSettings(config)
...
+        self.block_rendering_mode = config:readSetting("copt_block_rendering_mode")
+    else
+        if config:has("last_xpointer") then
             -- We have a last_xpointer: this book was previously opened
             self.block_rendering_mode = 0
         else
+            -- For new books, use 'web' mode
             self.block_rendering_mode = G_reader_settings:readSetting("copt_block_rendering_mode")
-                                            or block_rendering_default_mode
+                                     or 3
```
poire-z, line 71: `-- For new books, use the global setting, or 'web' mode if not set`
or move that comment after the `or 3 -- default to 'web' mode`

frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -56,63 +59,69 @@ function ReaderTypeset:onReadSettings(config)
...
 
     -- default to disable smooth scaling for now.
-    self.smooth_scaling = config:readSetting("smooth_scaling")
-    if self.smooth_scaling == nil then
+    if config:has("smooth_scaling") then
+        self.smooth_scaling = config:isTrue("smooth_scaling")
+    else
         local global = G_reader_settings:readSetting("copt_smooth_scaling")
-        self.smooth_scaling = (global == nil or global == 0) and 0 or 1
+        self.smooth_scaling = (global == nil or global == 0) and false or true
```
poire-z, line 115: (This and the next one were settings you added - so I guess you've checked it's fine for credocument and cre.cpp to use false/true instead of 0/1 - in some places for some other settings it wasn't :)

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -701,7 +701,7 @@ end
 -- in book settings, no default lang, and book has some language defined.
 function ReaderTypography:onReadSettings(config)
     -- Migrate old readerhyphenation setting, if one was set
-    if not config:readSetting("text_lang") and config:readSetting("hyph_alg") then
+    if config:hasNot("text_lang") and config:hasNot("hyph_alg") then
```
poire-z, line 704: **Error here**: `hasnot and has` translated to `hasnot and hasnot`

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -760,40 +765,36 @@ function ReaderTypography:onReadSettings(config)
...
     self:onToggleFloatingPunctuation(self.floating_punctuation)
 
     -- Decide and set the text main lang tag according to settings
-    self.allow_doc_lang_tag_override = false
-    -- Use the one manually set for this document
-    self.text_lang_tag = config:readSetting("text_lang")
-    if self.text_lang_tag then
+    if config:has("text_lang") then
+        self.allow_doc_lang_tag_override = false
+        -- Use the one manually set for this document
```
poire-z, line 778: (10mn checking this rewrite :/ - looks sane :)

frontend/apps/reader/modules/readerview.lua
```diff
@@ -965,11 +967,12 @@ function ReaderView:checkAutoSaveSettings()
...
         return
     end
-    local interval = G_reader_settings:readSetting("auto_save_settings_interval_minutes")
-    if not interval then -- no auto save
+    if G_reader_settings:nilOrFalse("auto_save_settings_interval_minutes") then
+        -- no auto save
         return
     end
 
+    local interval = G_reader_settings:readSetting("auto_save_settings_interval_minutes")
```
poire-z, line 975: (This blank line between 2 related bit of codes slowed my review by 20s :)

frontend/docsettings.lua
```diff
@@ -165,19 +165,112 @@ function DocSettings:open(docfile)
     return setmetatable(new, {__index = DocSettings})
 end
 
---- Reads a setting.
-function DocSettings:readSetting(key)
+--[[-- Reads a setting, optionally initializing it to a default.
+
+If default is set, and the key doesn't exist yet, it is initialized to default first.
```
poire-z, line 170: "If default is _provided_, ..." ?

frontend/docsettings.lua
```diff
@@ -165,19 +165,112 @@ function DocSettings:open(docfile)
...
+end
+
+--- Checks if setting does not exist.
+function DocSettings:hasNot(key)
+    return self.data[key] == nil
+end
+
+--- Checks if setting is `true`.
+function DocSettings:isTrue(key)
+    return string.lower(tostring(self.data[key])) == "true"
```
poire-z, line 218: (I have always find these test stranges when re-reading them :/ Does it or did it happen we stored "true" as a string?
There could be edge cases where "true" as a string - a relative path :) - might mean something different than `true` as a boolean.)

frontend/ui/screensaver.lua
```diff
@@ -264,6 +264,8 @@ function Screensaver:show(event, fallback_message)
         return
     end
 
+    -- These historical defaults are image-oriented.
+    -- That'll bite us in the ass later on, both for "disable" mode and show_message...
```
poire-z, line 268: (My english sucks, but is "bite us _in_ the ass" correct ? I don't see how that's physically possible, but well, I haven't seen everything :)

frontend/ui/widget/filechooser.lua
```diff
@@ -39,54 +74,65 @@ local unreadable_dir_content = {}
...
+    self.show_dir = function(dirname)
         for _, pattern in ipairs(self.exclude_dirs) do
             if dirname:match(pattern) then return false end
         end
         return true
     end
+    -- Standard file exclusion list
+    self.show_file = function(filename)
+        for _, pattern in ipairs(self.exclude_files) do
+            if filename:match(pattern) then return false end
```
poire-z, line 87: I guess it might be hard/impossible to make a single pattern from the whole list - to not have to call :match `N exclude patterns` times `N files` :/

poire-z (commented):
(That took some time to review... so probably a lot more time to do :/)
Mostly thoughts - except one error I think I spotted.

poire-z:
I hope you won't add too much to this PR - as it would take as much time to review the final code vs the original code :)

If you feel courageous, frontend/readcollection.lua is another ugly module that could use a reference instead of reading/saving a Luasetting everytime: https://github.com/koreader/koreader/issues/6054#issuecomment-769365372

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -59,6 +59,8 @@ local ReaderDictionary = InputContainer:new{
     data_dir = nil,
     dict_window_list = {},
     disable_lookup_history = G_reader_settings:isTrue("disable_lookup_history"),
+    dicts_order = G_reader_settings:readSetting("dicts_order", {}),
+    dicts_disabled = G_reader_settings:readSetting("dicts_disabled", {}),
```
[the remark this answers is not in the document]
NiLuJe: Yeah, those references are unique, as they should, but that one is a value, I'll move it to init, good catch ;).

Not quite used to OOP, but the idea is the the "class" object is only initialized on require, right?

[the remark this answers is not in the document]
NiLuJe: Actually, no, they're not: the ref is broken and updated in `showDictionariesMenu` during the SortWidget shenanigans.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -388,7 +388,7 @@ local ReaderFooter = WidgetContainer:extend{
 }
 
 function ReaderFooter:init()
-    self.settings = G_reader_settings:readSetting("footer") or {
+    self.settings = G_reader_settings:readSetting("footer", {
         -- enable progress bar by default
```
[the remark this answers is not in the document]
NiLuJe: That was very much the intent ;).

(But, yeah, this means that a user that never touched any footer settings will still get a footer section in its config).

frontend/apps/reader/modules/readerrolling.lua
```diff
@@ -1279,11 +1284,14 @@ function ReaderRolling:checkXPointersAndProposeDOMVersionUpgrade()
...
+        if G_reader_settings:has("copt_block_rendering_mode") then
+            g_block_rendering_mode = G_reader_settings:readSetting("copt_block_rendering_mode")
+        else
+            -- nil means: use default
+            g_block_rendering_mode = 3 -- default in ReaderTypeset:onReadSettings()
+        end
         if g_block_rendering_mode ~= 0 then -- default is not "legacy"
-            if not g_block_rendering_mode then -- nil means: use default
-                g_block_rendering_mode = 3 -- default in ReaderTypeset:onReadSettings()
-            end
```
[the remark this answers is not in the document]
NiLuJe: I did spend a bit of time on it (and its brethren in... RV? I think) ;).

frontend/apps/reader/modules/readertoc.lua
NiLuJe: I did switch quite a few things to isTrue to make those kinds of distinctions much clearer ;).

(e.g. limit readSetting to knobs that read a non-boolean value).

That one was aided by the following comment, but I did double-check ;).

frontend/apps/reader/modules/readertypeset.lua
NiLuJe: Yeah, I wasn't sure quite what to do with those comments, thanks ;).

frontend/apps/reader/modules/readertypeset.lua
NiLuJe: Yeah, that one was actually arguably a bug I sneakily fixed ^^.

(It worked because for all intents and purposes when doing boolean checks 0 is false and 1 is true), but it made the whole thing weird given the legacy settings above that *do* expect a 0/1 ;).

frontend/apps/reader/modules/readertypography.lua
NiLuJe: Oh, should have been obvious given the following line, thanks ;).

frontend/docsettings.lua
NiLuJe: I asked myself the same question ;).

I'll dig in the history, but I suspect they were written that way from the start for... some reason :?.

frontend/ui/screensaver.lua
NiLuJe: It's grammatically weird, but the idiom is indeed frozen that way ^^.

(I'm going to refactor ScreenSaver's show in two to try to make the whole thing less kludgey).

NiLuJe:
> I hope you won't add too much to this PR - as it would take as much time to review the final code vs the original code :)
> 
> If you feel courageous, frontend/readcollection.lua is another ugly module that could use a reference instead of reading/saving a Luasetting everytime: [#6054 (comment)](https://github.com/koreader/koreader/issues/6054#issuecomment-769365372)

Yeah, I opened that one, and then I started screaming. Nope, nope, nope, nope. :D.

If anything, I'd scrap it altogether and implement that via a flag in the BookInfoManager db.

frontend/apps/reader/modules/readerdictionary.lua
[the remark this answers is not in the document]
poire-z: > Not quite used to OOP, but the idea is the the "class" object is only initialized on require, right?

Yes, the one we create in each module with:
`local ReaderDictionary = InputContainer:new{???}`
and that we return, and what we return (any object) is cached by Lua module/require subsystem - and just given from the cache for subsequent require() of the same module.

But Lua OOP (or how the kind of way of doing it we use in KOReader) is a bit special:
We use the require-cached-object's `:new()` method - that is found far away in frontend/ui/widget/widget.lua `Widget:new(o)`: this method accepts an empty table, setmetatable it to the original require-cached-object (so stuff not found in its table are looked also in the original require-cached-object - and then call :init() which is defined in the near require-cached-object - and this :init might stuff to the normal table...

The side effect with the setmetatable stuff picking missing stuff (nil) from the near object in the remote require-cached-object  is that, if in some bit of code, we do `self.some_setting = nil`, we'll erase it only from the near object: it will stay in the remote require-cached-object   and will be found there if defined.

Some kind of object properties shadowing/unshadowing upper super class properties...

frontend/ui/widget/filechooser.lua
NiLuJe: Yeah, those aren't REs, we don't have an OR operator ;).

frontend/docsettings.lua
NiLuJe: It appears to -- weirdly -- have been designed for string values from the get-go: https://github.com/koreader/koreader/pull/2215

This doesn't make much sense to me, and I'm inclined to make it purely boolean and what (nothing) break, because storing true as a string is stupid ^^.

poire-z: ^ I agree to try that :)

NiLuJe: Oh, look: the only thing that was doing stupid shit was some tests :}.

NiLuJe:
There, pushed the Screensaver refactor.

Haven't entirely finished reviewing/testing it yet, but it's mostly done ;).

poire-z (approved):
(I haven't really focused on the screensaver, so trusting you.)

frontend/apps/filemanager/filemanagerbookinfo.lua
```diff
@@ -67,40 +67,40 @@ function BookInfo:show(file, book_props)
     table.insert(kv_pairs, { _("Folder:"), BD.dirpath(filemanagerutil.abbreviate(directory)), separator = true })
 
     -- book_props may be provided if caller already has them available
-    -- but it may lack 'pages', that we may get from sidecar file
+    -- but it may lack "pages", that we may get from sidecar file
```
Frenzie, line 70: if you're going to edit comments, I'd prefer if you make them proper sentences (uppercase first letter at the start, period at the end). ;-)

frontend/apps/filemanager/filemanagerbookinfo.lua
```diff
@@ -67,40 +67,40 @@ function BookInfo:show(file, book_props)
...
         -- that would create an empty sidecar directory
         if DocSettings:hasSidecarFile(file) then
             local doc_settings = DocSettings:open(file)
             if doc_settings then
                 if not book_props then
-                    -- Files opened after 20170701 have a 'doc_props' setting with
-                    -- complete metadata and 'doc_pages' with accurate nb of pages
-                    book_props = doc_settings:readSetting('doc_props')
+                    -- Files opened after 20170701 have a "doc_props" setting with
+                    -- complete metadata and "doc_pages" with accurate nb of pages
```
Frenzie, line 79: Although I admit on this one not having a period gives it a nice balance of exactly the same width.

frontend/apps/filemanager/filemanagerbookinfo.lua
```diff
@@ -67,40 +67,40 @@ function BookInfo:show(file, book_props)
...
+                    local stats = doc_settings:readSetting("stats")
                     if stats and stats.pages ~= 0 then
                         -- Let's use them as is (which was what was done before), even if
                         -- incomplete, to avoid expensive book opening
                         book_props = stats
                     end
                 end
-                -- Files opened after 20170701 have an accurate 'doc_pages' setting
-                local doc_pages = doc_settings:readSetting('doc_pages')
+                -- Files opened after 20170701 have an accurate "doc_pages" setting
```
Frenzie, line 94:
```suggestion
                -- Files opened after 20170701 have an accurate "doc_pages" setting.
```

frontend/apps/filemanager/filemanagerbookinfo.lua
```diff
@@ -67,40 +67,40 @@ function BookInfo:show(file, book_props)
...
-                    -- Files opened after 20170701 have a 'doc_props' setting with
-                    -- complete metadata and 'doc_pages' with accurate nb of pages
-                    book_props = doc_settings:readSetting('doc_props')
+                    -- Files opened after 20170701 have a "doc_props" setting with
+                    -- complete metadata and "doc_pages" with accurate nb of pages
+                    book_props = doc_settings:readSetting("doc_props")
                 end
                 if not book_props then
-                    -- File last opened before 20170701 may have a 'stats' setting
+                    -- File last opened before 20170701 may have a "stats" setting
```
Frenzie, line 83:
```suggestion
                    -- File last opened before 20170701 may have a "stats" setting.
```

frontend/apps/reader/modules/readerkoptlistener.lua
```diff
@@ -16,15 +16,17 @@ end
...
     -- normal zoom mode is zoom mode used in non-reflow mode.
-    local normal_zoom_mode = config:readSetting("normal_zoom_mode") or
-                    G_reader_settings:readSetting("zoom_mode") or "page"
+    local normal_zoom_mode = config:readSetting("normal_zoom_mode")
+                          or G_reader_settings:readSetting("zoom_mode")
+                          or "page"
     normal_zoom_mode = util.arrayContains(ReaderZooming.available_zoom_modes, normal_zoom_mode)
-                    and normal_zoom_mode
+                   and normal_zoom_mode
                     or "page"
```
Frenzie, line 24: Seems it should either be this or the one above with its space there included?
```suggestion
                   or "page"
```

frontend/apps/reader/modules/readertypeset.lua
```diff
@@ -56,63 +59,68 @@ function ReaderTypeset:onReadSettings(config)
...
-        G_reader_settings:readSetting("copt_h_page_margins") or
-        DCREREADER_CONFIG_H_MARGIN_SIZES_MEDIUM
-    local t_margin = config:readSetting("copt_t_page_margin") or
-        G_reader_settings:readSetting("copt_t_page_margin") or
-        DCREREADER_CONFIG_T_MARGIN_SIZES_LARGE
-    local b_margin = config:readSetting("copt_b_page_margin") or
-        G_reader_settings:readSetting("copt_b_page_margin") or
-        DCREREADER_CONFIG_B_MARGIN_SIZES_LARGE
+    local h_margins = config:readSetting("copt_h_page_margins")
+                   or G_reader_settings:readSetting("copt_h_page_margins")
```
Frenzie, line 88: What's the exact rule you've used btw? We should probably add it to the style guide. (I slightly prefer or at the beginning of the line, I care less about the specific indentation, and in any case there's a big effort here to standardize it.)

frontend/luasettings.lua
```diff
@@ -69,8 +69,29 @@ function LuaSettings:child(key)
...
+--[[-- Reads a setting, optionally initializing it to a default.
+
+If default is provided, and the key doesn't exist yet, it is initialized to default first.
+This ensures both that the defaults are actually set if necessary,
+and that the returned reference actually belongs to the LuaSettings object straight away,
+without requiring further interaction (e.g., saveSetting) from the caller.
+
+This is mainly useful if the data type you want to retrieve/store is assigned/returned/passed by reference (e.g., a table),
+and you never actually break that reference by assigning another one to the same variable, (by e.g., assigning it a new object).
+c.f., https://www.lua.org/manual/5.1/manual.html#2.2
```
Frenzie, line 81:
```suggestion
c.f., <https://www.lua.org/manual/5.1/manual.html#2.2>.
```

frontend/ui/screensaver.lua
```diff
@@ -23,18 +23,41 @@ local _ = require("gettext")
...
 local T = require("ffi/util").template
 
-local screensaver_provider = {
-    ["jpg"] = true,
-    ["jpeg"] = true,
-    ["png"] = true,
-    ["gif"] = true,
-    ["tif"] = true,
-    ["tiff"] = true,
+-- Migrate old settings
```
Frenzie, line 26: Some kind of indication might be useful?
```suggestion
-- Migrate old settings from 2021.02 or older.
```

frontend/ui/screensaver.lua
```diff
@@ -23,18 +23,41 @@ local _ = require("gettext")
...
+if G_reader_settings:has("screensaver_white_background") then
+    if G_reader_settings:isTrue("screensaver_white_background") then
+        G_reader_settings:saveSetting("screensaver_background", "white")
+    end
+    G_reader_settings:delSetting("screensaver_white_background")
+end
+
+local Screensaver = {
+    screensaver_provider = {
+        ["jpg"] = true,
```
Frenzie, line 46: or just
```suggestion
        jpg = true,
```

Frenzie: Btw, what's this list of filetypes about exactly? Not really related to this PR of course.

frontend/ui/screensaver.lua
```diff
@@ -61,6 +84,114 @@ local function getRandomImage(dir)
...
+        return fallback
+    end
+
+    local ret = message
+    local lastfile = G_reader_settings:readSetting("lastfile")
+
+    local totalpages = 0
+    local percent = 0
+    local currentpage = 0
+    local title = "N/A"
```
Frenzie, line 106: No translation?

frontend/ui/widget/filechooser.lua
```diff
@@ -25,10 +25,45 @@ local FileChooser = Menu:extend{
...
+        "^System Volume Information$",
+        -- Plato
+        "^%.thumbnail%-previews$",
+        "^%.reading%-states$",
+    },
+    exclude_files = {
+        -- macOS
+        "^%.DS_Store$",
+        -- Windows
+        "^Thumbs%.db$",
```
Frenzie, line 56: also try .directory for Linux (Dolphin at least)

frontend/util.lua
```diff
@@ -373,6 +373,44 @@ function util.arrayAppend(t1, t2)
     end
 end
 
+--[[--
+Remove elements from an array, fast.
+
+Swap & pop, like http://lua-users.org/lists/lua-l/2013-11/msg00027.html / https://stackoverflow.com/a/28942022, but preserving order.
+c.f., https://stackoverflow.com/a/53038524
```
Frenzie, line 380:
```suggestion
Swap & pop, like <http://lua-users.org/lists/lua-l/2013-11/msg00027.html> / <https://stackoverflow.com/a/28942022>, but preserving order.
c.f., <https://stackoverflow.com/a/53038524>
```

Frenzie (commented):
Some nits & comments, nothing big.

frontend/apps/reader/modules/readertypeset.lua
poire-z: > (I slightly prefer or at the beginning of the line)

Mhhh, I really prefer if there is some indentation. As for the level of indentation, I'd say let us have some freedom on that, it may depends on what we 'or':
- if it's too long, I might put it before the `= `  
- if it's short, I might indent it further than the `= '
- or I may just indent it so some part aligns with a part in each other, for aesthetics or fun :)

I'm quite fine with the solution above for this specific case (the or'ed things align).

Frenzie: > Mhhh, I really prefer if there is some indentation.

No, that's not at all what I meant! ;-) It should definitely have one level of indentation (4 spaces), or more.

I meant as in end vs beginning
```lua
bla or
  alb
```
```lua
bla
  or alb
```

frontend/apps/filemanager/filemanagerbookinfo.lua
NiLuJe: That one was an S&R pass ;p.

frontend/apps/reader/modules/readerkoptlistener.lua
NiLuJe: I tend to prefer aligning variables, yeah ;).

(Especially for an assignment).

NiLuJe: (Unless I'm getting cross-eyed, it's aligned on my end).

frontend/apps/reader/modules/readertypeset.lua
NiLuJe: The idea is to align the *variables* on the same column (especially for assignments).

I'm more ambivalent about tests ;).

e.g., [clang-format](https://releases.llvm.org/11.0.0/tools/clang/docs/ClangFormatStyleOptions.html) `AlignOperands`, either `Align` or `AlignAfter` w/ `BreakBeforeBinaryOperators` set to `NonAssignment`. (I tend to prefer Align for C/C++, because of parens and stuff).

frontend/ui/screensaver.lua
NiLuJe: It's the filter for choosing image files manually.

frontend/apps/reader/modules/readerkoptlistener.lua
Frenzie: Carry on then.

Frenzie (approved):

NiLuJe (commented):
> Reviewed 3 of 16 files at r7, 1 of 1 files at r9, 4 of 5 files at r12, 9 of 13 files at r13, 22 of 25 files at r14, 9 of 10 files at r15, 1 of 1 files at r16.
> Dismissed @Frenzie and @poire-z from 14 discussions.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7371)* status: all files reviewed, 11 unresolved discussions (waiting on @Frenzie, @NiLuJe, @pazos, and @poire-z)

___
*[frontend/docsettings.lua, line 170 at r7](https://reviewable.io/reviews/koreader/koreader/7371#-MUg4qJ4-MTW98XfKz_T-r7-170:-MV8LnONE5YKKlzJgKl2:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/ff1256b7bbbf71afd6701e7ad42b948acae61b7c/frontend/docsettings.lua#L170)):*

"If default is _provided_, ..." ?
</blockquote></details>

Done.
___
*[frontend/docsettings.lua, line 218 at r7](https://reviewable.io/reviews/koreader/koreader/7371#-MUg4qJ4-MTW98XfKz_T-r7-218:-MV8LrfP82PIuZ88sBNO:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/ff1256b7bbbf71afd6701e7ad42b948acae61b7c/frontend/docsettings.lua#L218)):*

Oh, look: the only things that were doing stupid shit were some tests :}.
</blockquote></details>

Done.
___
*[frontend/luasettings.lua, line 81 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUg4qJ4-MTW98XfKz_U-r14-81:-MV8M3IY6o6VoMy3Rdo2:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/luasettings.lua#L81)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

```suggestion
c.f., <https://www.lua.org/manual/5.1/manual.html#2.2>.
```
</blockquote></details>

Done.
___
*[frontend/util.lua, line 380 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUo2GHX1Z1Bftd__FRB-r14-380:-MV8MHjADS8ANq7iMUru:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/util.lua#L380)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

```suggestion
Swap & pop, like <http://lua-users.org/lists/lua-l/2013-11/msg00027.html> / <https://stackoverflow.com/a/28942022>, but preserving order.
c.f., <https://stackoverflow.com/a/53038524>
```
</blockquote></details>

Done.
___
*[frontend/apps/filemanager/filemanagerbookinfo.lua, line 83 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZsvNk9hWGlUPhiE-r14-83:-MV8MSZKEM3n8jxOzY5K:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/apps/filemanager/filemanagerbookinfo.lua#L83)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

```suggestion
                    -- File last opened before 20170701 may have a "stats" setting.
```
</blockquote></details>

Done.
___
*[frontend/apps/filemanager/filemanagerbookinfo.lua, line 94 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZsvNk9hWGlUPhiE-r14-94:-MV8MUOHEjP8tiOOqoxK:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/apps/filemanager/filemanagerbookinfo.lua#L94)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

```suggestion
                -- Files opened after 20170701 have an accurate "doc_pages" setting.
```
</blockquote></details>

Done.
___
*[frontend/apps/reader/modules/readerkoptlistener.lua, line 24 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZswuY1o0vSrm_R8-r14-24:-MV8NTfp9pNfFGrp_VZ3:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/apps/reader/modules/readerkoptlistener.lua#L24)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

Carry on then.
</blockquote></details>

Done.
___
*[frontend/apps/reader/modules/readertoc.lua, line 140 at r7](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZswuY1o0vSrm_RF-r7-140:-MV8PZqf6GA-WvctSr__:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/ff1256b7bbbf71afd6701e7ad42b948acae61b7c/frontend/apps/reader/modules/readertoc.lua#L140)):*

I did switch quite a few things to isTrue to make those kinds of distinctions much clearer ;).

(e.g. limit readSetting to knobs that read a non-boolean value).

That one was aided by the following comment, but I did double-check ;).
</blockquote></details>

Done.
___
*[frontend/apps/reader/modules/readertypeset.lua, line 71 at r7](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZswuY1o0vSrm_RG-r7-71:-MV8PhvOEsbwllfwbQ0o:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/ff1256b7bbbf71afd6701e7ad42b948acae61b7c/frontend/apps/reader/modules/readertypeset.lua#L71)):*

Yeah, I wasn't sure quite what to do with those comments, thanks ;).
</blockquote></details>

Done.
___
*[frontend/apps/reader/modules/readertypeset.lua, line 115 at r7](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZswuY1o0vSrm_RG-r7-115:-MV8PpiWC1e9pb8sy65W:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/ff1256b7bbbf71afd6701e7ad42b948acae61b7c/frontend/apps/reader/modules/readertypeset.lua#L115)):*

Yeah, that one was actually arguably a bug I sneakily fixed ^^.

(It worked because for all intents and purposes when doing boolean checks 0 is false and 1 is true), but it made the whole thing weird given the legacy settings above that *do* expect a 0/1 ;).
</blockquote></details>

Done.
___
*[frontend/apps/reader/modules/readertypography.lua, line 704 at r7](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZswuY1o0vSrm_RH-r7-704:-MV8QABs4pjNVfbQlLKC:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/ff1256b7bbbf71afd6701e7ad42b948acae61b7c/frontend/apps/reader/modules/readertypography.lua#L704)):*

Oh, should have been obvious given the following line, thanks ;).
</blockquote></details>

Done.
___
*[frontend/ui/screensaver.lua, line 26 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZswuY1o0vSrm_RS-r14-26:-MV8Qoe5Dfoel8epIfiH:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/ui/screensaver.lua#L26)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

Some kind of indication might be useful?
```suggestion
-- Migrate old settings from 2021.02 or older.
```
</blockquote></details>

Done.
___
*[frontend/ui/screensaver.lua, line 106 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUpQZswuY1o0vSrm_RS-r14-106:-MV8Qrv-1fIrFWMF0Gyp:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/ui/screensaver.lua#L106)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

No translation?
</blockquote></details>

Done.
___
*[frontend/ui/widget/filechooser.lua, line 56 at r14](https://reviewable.io/reviews/koreader/koreader/7371#-MUuMzJ1lG5BqeSRW-xS-r14-56:-MV8TLIh5CpVDwelBD2d:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/e6d667cc92ef51f020317cc5dc98d9f4f264cca7/frontend/ui/widget/filechooser.lua#L56)):*
<details><summary><i>Previously, Frenzie (Frans de Jonge) wrote…</i></summary><blockquote>

also try .directory for Linux (Dolphin at least)
</blockquote></details>

Done.

poire-z:
Upgraded this morning, suspending just now, and I got my cover image with an unwanted "Sleeping" mesasge overlay.
I had `[x] Use last book's cover as screensaver` (using the book cover or a fallback image),  and
`[ ] Add message to screensaver` unchecked - and I got the added message.

Enabling `[x] Add message to screensaver` kept it being displayed (as expected), but unchecking it back to `[ ] Add message to screensaver` got rid of the message.
So, all is fine, just letting you know of this minor issue.

NiLuJe:
I assume you'd never checked `Add message to screensaver` (ever?).

The fallback is probably a bit wonky, I'll double-check.

NiLuJe:
Sidebar: this is why i've never been a fan of the weird flip* APIs that deletes setting keys: it means that two different pieces of code have to be in agreement about what do to as a default: the actual code, *and* the settings UI.

(Here, the menu entry was already cheating and checking type instead of show_message ;p).

Frenzie:
I think those flip things are weird too.

poire-z:
Just a note about the `:readSettings(name, default)`, that many PRs recently have updated many bits of code to use it:
With this, the default as of today is now saved in our settings.reader.lua, even if the user has never tweaked it.
This means that if one day, we update the default because we think it's best, all past users who didn't care tweaking it WILL NOT get the new default - but will keep the one as of today.

I somehow prefered the previous behaviour, and seeing only the settings I had modified in my settings.reader.lua. Over a month, my settings.reader.lua has grown with: :) 
```lua
["bookmarks_items_auto_text"] = false,
["bookmarks_items_font_size"] = 16,
["bookmarks_items_reverse_sorting"] = false,
["copt_status_line"] = 1,
["cre_header_auto_refresh"] = 1,
["cre_header_battery"] = 1,
["cre_header_battery_percent"] = 0,
["cre_header_chapter_marks"] = 1,
["cre_header_clock"] = 1,
["cre_header_page_count"] = 1,
["cre_header_page_number"] = 1,
["cre_header_reading_percent"] = 0,
["cre_header_title"] = 1,
["page_turns_tap_zone_forward_size_ratio"] = 0.4,
["page_turns_tap_zones"] = "default",
["reader_footer_custom_text"] = "KOReader",
["reader_footer_custom_text_repetitions"] = "1",
```

Frenzie:
Yeah, I agree.

NiLuJe:
Yeah, there were some funky corner cases in the migration path, given the unhelpful way this was handled ;).

NiLuJe:
That was indeed by design, migrating defaults ought to be handled via a one time migration pass ;).

NiLuJe:
To explain the design choice further, not relying on magic nil settings
allows us to simplify code as well as said migration paths ;).

poire-z:
I get that this behaviour/way to use readSetting() may have had some use, whether because of migration or not.
But in this case, it has been used for new settings. So I wonder how much we should have advertized this was the "right" way to do any settings handling, or if we should add some note as to "when/why" it should be used or not.
It might be a little too late now :) so I would be satisfied with just some "pros" thoughts about the current trend of having all settings "with their default value at the time" in settings.reader.lua (because I mostly see "cons", even if just little ones - and seeing them too late :)

("Any setting might at some point in the future be migrated, so it will be easier if we save it today with its default value" might be one, but it feels insufficient :)

NiLuJe:
Oh, there's definitely a big part of "I found the old API/status quo mind-boggingly confusing" in there, so, yeah, more documentation could be helpful ;).

Issue #7370, opened 2021-03-01
OPDS - no access to the flibusta mirror


hius07:
OK, flibusta etc. are excluded from the package, but we can add them manually )

flibusta.is works fine, but is blocked in some regions.
The issue: KOReader cannot access the mirror: https://flibusta.appspot.com/opds/
CoolReader can.

NiLuJe:
Looks like a broken mirror: headers say it's gzip encoded and set the length properly (like the master), but it actually serves uncompressed content that's much larger, which fails the length check.

I'm not quite sure where this is failing to get reproted in the Lua stack, but it's visible in HTTPie.

NiLuJe:
Or, well, that's the other way around: it's actually serving gz content directly (c.f., curl).

NiLuJe:
And I can't convince it not to, even via ACcept-Encoding: identity

hius07:
The site seems has gone away so closing the issue.

PR #7372, opened 2021-03-01
OPDS: Don't needlessly setup Basic auth

NiLuJe:
InputDialog returns empty strings, not nil. Explicitly nil username ~~& password~~ when they're empty to avoid enabling authentication.

Also, re #7370, explicitly request no compression. Doesn't fix the actual issue because the server doesn't obey, breaking the RFC. Most servers were already assuming no Accept-Encoding header meant that we really didn't want gzip data ;p.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7372)


hius07:
> Doesn't fix the actual issue

Maybe some workarounds especially for this server?
flibusta.is is blocked in Russia, this is the working mirror for them.

NiLuJe:
PRs welcome to handle `Content-Encoding: gzip` ;o).

hius07:
On that server I cannot even get the books list, OPDS browser gives a message 'Catalog not found'.

NiLuJe:
Yep, because it's serving gzip data instead of plaintext ;).

PR #7378, opened 2021-03-03
Another minor fixup post #7335

NiLuJe:
Remove an unwanted truncation.

This *should* be a float ;).

c.f., https://github.com/koreader/koreader/pull/7335#discussion_r586548610

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7378)


poire-z (approved):
Thanks!

PR #7379, opened 2021-03-03
add update of the top and bottom status bars, on +/-Charging

zwim:
It is desirable to get immediate response on the charging status of a device. (Plug not in,  power switch off, cable broken ...)
This PR tries to improve the display of the TopStatusBar by changing `[ + ]` to `[73%]`.

On platforms using "Charging" and "NotCharging" events, the TopStatusBar should get updated on (un)plugging the power immediately (not tested, as I don't have any of such devices).

On Android the charging status gets polled (every 1.3s when running and after 5s after startup). If the polling detects a change of the charging state the TopStatusBar gets updated.

Draft for now, as I am not sure what the consensus on polling/eventing the charging status is.
Personally I don't have a problem to poll data on a Hz base, but if an event can be triggered easy this would be preferable.
@pazos: 
 1st: How difficult is an event on the change of the charge state?
 2nd: Could you, want you and would you add such an event the Android incarnation of KOReader ????

Bottom-Bar is not covered with this PR for now. If this PR gets approved the Bottom story can be done similarly.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7379)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -199,17 +199,14 @@ function Device:init()
                         end)
                     end
                 end
-            elseif ev.code == C.MSC_CHARGE then
+            elseif ev.code == C.EVENT_POWER_CONNECTED then
                 local Event = require("ui/event")
-                if ev.value == 1 then
-                    UIManager:broadcastEvent(Event:new("Charging"))
-                elseif ev.value == 0 then
-                    UIManager:broadcastEvent(Event:new("NotCharging"))
-                else
-                    logger.err("Unknow event C.EV_MSC=" .. C.MSC_CHARGE .. " with ev.code=" .. ev.code)
-                end
+                UIManager:broadcastEvent(Event:new("Charging"))
+            elseif ev.code == C.EVENT_POWER_DISCONNECTED then
+                local Event = require("ui/event")
+                UIManager:broadcastEvent(Event:new("NotCharging"))
             else
-                logger.err("Unknow event ev.code=" .. ev.code .. " with ev.code=" .. ev.code)
+                logger.err("Unknow event ev.code=" .. ev.code)
             end
         end,
         hasClipboardText = function()
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -109,6 +109,7 @@ local Device = Generic:new{
             android.dictLookup(text, app, action)
         end
     end,
+    isCharging = false, -- used for pollingUSB, can be removed if "Charging" ant "NotCharging" events are implemented
 
     -- Android is very finicky, and the LuaJIT allocator has a tremendously hard time getting the system
     -- to allocate memory for it in the very specific memory range it requires to store generated code (mcode).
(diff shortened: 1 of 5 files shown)


NiLuJe:
Yeah, hard nope on the polling approach on my end.

I imagine/hope there's an actual way of getting that event out of Android, though ;).

NiLuJe:
Especially given that we already have an optional timer for footer/header refreshes ;).

Frenzie:
Android has something along the lines of POWER_CONNECTED, should be easy enough. (But I'm not interested in it; PRs welcome etc. xp)

NiLuJe:
Random google result: https://developer.android.com/training/monitoring-device-state/battery-monitoring

Frenzie:
Yes, the ACTION_ was implicit for people with basic Android familiarity, my bad. ;-)

NiLuJe:
The Charging/NotCharging stuff is not a bad idea, though; feel free to add that to readerfooter, too ;).

pazos:
Sorry for the late reply @zwim: I have been very busy these days.

> 1st: How difficult is an event on the change of the charge state?

I think you figured out your way in https://github.com/koreader/android-luajit-launcher/pull/294. That would be my suggestion too.  I've commented there. The difficult part is to communicate broadcast events back to lua when they happen. The communication is one way `lua->JVM` and to get back results from the JVM we're storing values on variables and reading these values. No much event driven :/

> 2nd: Could you, want you and would you add such an event the Android incarnation of KOReader ????

Nope, but I would be happy if you find a way to tell back to execute lua chunks from JVM methods. That would make possible very interesting features.

zwim:
Things get easy.

zwim:
Not tested too, because I am waiting for https://github.com/koreader/koreader-base/pull/1337.

frontend/device/android/device.lua
```diff
@@ -197,6 +197,14 @@ function Device:init()
                         end)
                     end
                 end
+            elseif ev.code == C.ALOOPER_FIFO_POWER_CONNECTED then
+                local Event = require("ui/event")
+                UIManager:broadcastEvent(Event:new("Charging"))
+            elseif ev.code == C.ALOOPER_FIFO_POWER_DISCONNECTED then
+                local Event = require("ui/event")
+                UIManager:broadcastEvent(Event:new("NotCharging"))
+            else
```
pazos, line 206: please remove the else statement. A few different reasons:

1. is not an error, just a command we don't need to handle.
2. it is already logged in https://github.com/koreader/koreader/blob/99b873ece3f9ddd368037ed490bf0cdb86fc5e36/frontend/device/android/device.lua#L137
3. the message is a bit wrong. Its value is already known (and part of `android.lua`), but we decide to ignore it, because we don't need that.

The debug statement is enough for debugging and release builds shouldn't care about not handled commands.

poire-z:
Haven't kept track of all the base/android changes related to this PR - but this PR is quite small and fine and passed my review :)
Is it ready to merge ?

Frenzie:
The Android stuff isn't merged yet, is it?

pazos:
> The Android stuff isn't merged yet, is it?

Just merged the android stuff & base. please @zwim update both submodules. Thanks!

zwim:
Oh I just questioned me: Did you mean to update the submodules here in this PR? Or is it OK I made two different PR for the pumps?

Frenzie:
Sure, that should be fine.

zwim:
From my side this could be merged, what do you think. The base and android-luajit submodules are pumped.

poire-z:
There's a message down there that says we can't :) "due to conflicts".
So, I guess you're good for rebasing to current master:
```
git checkout master
git fetch upstream ; git merge upstream/master
git checkout statusLineUpdateCharging
git rebase master (this may fail at some point and need you to do some manual cleanup, can be messy)
git push -f origin statusLineUpdateCharging (would make this mergable)
```

zwim:
@poire-z I did the rebase, but I think there is something wrong with the pump-luajit commit!

poire-z:
Yep, looks like you deleted it :)
Well, it's the last commit (and you shouldn't need in this PR anything related to luajit-launcher as you already bumped it with another PR), so just remove it (git rebase -i master : pick > drop on the last commit)

zwim:
Looks better now. Thank you, I have learned something :)

PR #7380, opened 2021-03-04
OPDS: Parse Open Search Definition for feeds that support it

### LINKED ISSUE #5119: OPDS Search
* KOReader version: 2019.07
* Device: Kindle PW4

#### Issue

OPDS Catalogs are not searchable

#### Steps to reproduce

Similar to #2442, but across all OPDS catalogs, not necessarily just Calibre.

A simple searchable public OPDS server that can be used for testing is `opds://www.feedbooks.com/catalog.atom`

ajvpot:
Adds a search option to OPDS feeds that support the Open Search Description spec as described in https://specs.opds.io/opds-1.2#3-search
Removes searchable variants of OPDS catalogs from defaults as they are all supported by Search option.
Existing searchable catalogs will continue to work.

Resolves #5119

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7380)


NiLuJe:
Minor stylistic nit: We tend to limit camelCase to functions/methods, and prefer snake_case for variables ;).

ajvpot:
> Minor stylistic nit: We tend to limit camelCase to functions/methods, and prefer snake_case for variables ;).
> 
> EDIT: Mkay, I realize some of this is forced upon you by the OpenSearch spec itself ;o).

Thanks for the review! Gotcha on the snake_case, I don't think I'll be able to change the others because of OpenSearch.

Inserting an artificial menu item is causing some of the tests to fail and makes me think it may not be the best way to present the search option to the user. Do you think it would make sense to add a Search button w/ callback to `Menu` similar to the Return/Up button? I'm looking into ways to extend OPDSBrowser but I'm unfamiliar with lua.

NiLuJe:
I'm unfamiliar with how the existing support was presented, to be honest ;).

The extent of the updates to the tests doesn't worry me overmuch, FWIW, especially given that touching Menu is... touchy ^^.

ajvpot:
Existing support was to add the menu item manually. This required two options for the same catalog, while the primary catalog contains all the information needed to generate the second option.

image

Clicking the `[Searchable]` option gets you

image.

This patch parses the search info out of the main feed and adds a menu item, removing the need for the second catalog.

Untitled

NiLuJe:
Thanks!

NiLuJe:
Realized this when rebasing #7371: Should we kill the old manual "searchable" checkbox and the accompanying code?

NiLuJe:
Oh, err, we need not to break existing configs, at least, so, just the "new" server bits?

ajvpot:
This code calls into the existing search code, so we'll need to keep the dialog etc. The "searchable" field seems to be derived from the url containing `%s`. I think it would be useful to leave the manual option in case the OPDS feed does not implement the spec correctly (seems to happen a lot :) ).

Issue #6418, opened 2020-07-22
Gallica OPDS doesn't always provide item author


Frenzie:
* KOReader version: 2020.07
* Device: all

#### Issue
The code depends on item.author being available, this doesn't seem to be guaranteed.

```
./luajit: frontend/ui/widget/opdsbrowser.lua:547: attempt to concatenate field 'author' (a nil value)
stack traceback:
	frontend/ui/widget/opdsbrowser.lua:547: in function 'downloadFile'
	frontend/ui/widget/opdsbrowser.lua:650: in function 'callback'
	frontend/ui/widget/button.lua:239: in function 'action'
	frontend/ui/uimanager.lua:821: in function '_checkTasks'
	frontend/ui/uimanager.lua:1141: in function 'handleInput'
	frontend/ui/uimanager.lua:1227: in function 'run'
	./reader.lua:299: in main chunk
	[C]: at 0x563ba7d07791
```

```
07/22/20-18:42:06 DEBUG {
    ["image"] = "http://gallica.bnf.fr/ark:/12148/bpt6k6531118k.highres",
    ["thumbnail"] = "http://gallica.bnf.fr/ark:/12148/bpt6k6531118k.thumbnail",
    ["id"] = "http://gallica.bnf.fr/ark:/12148/bpt6k6531118k",
    ["acquisitions"] = {
        [1] = {
            ["type"] = "application/epub+zip",
            ["href"] = "http://gallica.bnf.fr/ark:/12148/bpt6k6531118k.epub"
        }
    },
    ["updated"] = "2013-08-01T00:00:00",
    ["title"] = "L'escoufle : roman d'aventure / publ. pour la première fois d'après le manuscrit unique de l'Arsenal par H. Michelant et P. Meyer",
    ["text"] = "L'escoufle : roman d'aventure / publ. pour la première fois d'après le manuscrit unique de l'Arsenal par H. Michelant et P. Meyer"
}
```

I came across this while looking into #5997. I'll fix it at the same time.

#### Steps to reproduce
Download something from Gallica OPDS.

hius07:
> bpt6k6531118k.epub

Well, this name is not very good to save. Can we have a choice: use metadata or keep original?

hius07:
Loading two versions of a book from Flibusta (e.g. by different translators) we will get the same metadata author+title filename.

Frenzie:
Technically it's slightly different in content-disposition:

```
[...]

PR #7384, opened 2021-03-04
OPDS: Do not include author in download filename if nil or empty

ajvpot:
Some OPDS servers do not provide authors for all items. In this case, fall back to the item name only.
Resolves crash in #6418, option for downloading by Content-Disposition should still be considered.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7384)


NiLuJe (approved):
:lgtm:, thanks!

Frenzie (approved):

PR #7389, opened 2021-03-05
Dispatcher: nil guard against access to a document instance

### LINKED ISSUE #7388: Dispatcher crash when Dispatcher attempts to set Reader settings in the FM
Btw, not sure if this is meant for here, but there is some crash going on when trying to rotate screen with long tap set at left and right corner on Aura One reader. Works ok when you have a book opened, happens just on homescreen.
https://imgur.com/a/c6hs9J7

_Originally posted by @ReaLx3m in https://github.com/koreader/koreader/issues/7304#issuecomment-791396658_

NiLuJe:
Fix #7388

Related fixes:

* Unbreak the Rotation submenu entries since #7306
* Actually update the touchzones properly when the FM is rotated

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7389)

### DIFF
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -599,7 +599,6 @@ arguments are:
     3) optionally a `gestures`object
 --]]--
 function Dispatcher:execute(ui, settings, gesture)
-    print("Dispatcher:execute", ui, settings, gesture)
     for k, v in pairs(settings) do
         if settingsList[k] ~= nil and (settingsList[k].conditions == nil or settingsList[k].conditions == true) then
             if settingsList[k].category == "none" then
@@ -631,8 +630,6 @@ function Dispatcher:execute(ui, settings, gesture)
                         if v == r then value = settingsList[k].configurable.values[i] break end
                     end
                 end
-                print("Have a configurable:", settingsList[k].configurable.name, value)
-                print("Document?", ui.document)
                 -- CreOptions / KoptOptions settings are exposed in the FM's Gesture Manager!
                 if ui.document then
                     ui.document.configurable[settingsList[k].configurable.name] = value
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -599,6 +599,7 @@ arguments are:
     3) optionally a `gestures`object
 --]]--
 function Dispatcher:execute(ui, settings, gesture)
+    print("Dispatcher:execute", ui, settings, gesture)
     for k, v in pairs(settings) do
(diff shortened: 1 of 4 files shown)


NiLuJe:
@yparitcher: Should the ConfigDialog stuff even be *visible* in the FM's gesture manager menus?

yparitcher:
Dispatcher is reader/FM agnostic, gestures happens to use different instances, however profiles is the same across both (each profile can be used everywhere). This usually works well as events are safe to send even when no handler is registered. to separate then would mean a lot of work for little gain, and would cause ui/menu inconsistencies.

yparitcher (commented):
cleaner
```diff
                 local arg = v ~= 0 and v or gesture or 0
                 ui:handleEvent(Event:new(settingsList[k].event, arg))
             end
-            if settingsList[k].configurable ~= nil then
-                 local value = v
-                 if type(v) ~= "number" then
-                     for i, r in ipairs(settingsList[k].args) do
+            if settingsList[k].configurable ~= nil and ui.document then
+                local value = v
+                if type(v) ~= "number" then
+                    for i, r in ipairs(settingsList[k].args) do
                         if v == r then value = settingsList[k].configurable.values[i] break end
-                     end
-                 end
-                 ui.document.configurable[settingsList[k].configurable.name] = value
+                    end
+                end
+                ui.document.configurable[settingsList[k].configurable.name] = value
             end
         end
     end
```

frontend/dispatcher.lua
```diff
@@ -623,14 +623,14 @@ function Dispatcher:execute(ui, settings, gesture)
                 local arg = v ~= 0 and v or gesture or 0
                 ui:handleEvent(Event:new(settingsList[k].event, arg))
             end
-            if settingsList[k].configurable ~= nil then
-                 local value = v
-                 if type(v) ~= "number" then
-                     for i, r in ipairs(settingsList[k].args) do
+            if settingsList[k].configurable ~= nil and ui.document then
```
Frenzie, line 626: Or the other way around?

Frenzie (approved):

frontend/dispatcher.lua
NiLuJe: Hmm, arguably, you're less likely to hit a configurable than not being in the Reader, but it's pretty much a coin toss ;).

Frenzie: I was talking more or less exclusively about ease of reading, although it could potentially make an impossibly small performance impact one way or the other. ;-)

PR #7391, opened 2021-03-05
ReaderFooter: Fix some interactions between margins and text width

### LINKED ISSUE #7390: Wrong calculation of horizontal space for the status bar when left-aligned
* KOReader version: v2021.02
* Device: Android

#### Issue
The status bar displays ellipsis "..." at the right end (instead of the last item(s)), even though there is plenty of space available on the right.

#### Steps to reproduce
Configure the bottom (not alternative) status bar with the following settings:

1. Do NOT display progress bar
2. Display all items at once
3. Display separator
4. Display: current page, current time, pages left in chapter, battery status, bookmark count, progress percentage, book time to read, chapter time to read, frontlight level, koreader memory usage
5. Set alignment to left

Observe that everything fits nicely and there is still quite a bit of space left on the right. So, now increase the status bar's font size from 14 to 15. Observe that the last item is replaced by "..." even though there is more than enough space to fit it and there would still be room left to go to size 16, but perhaps not 17.

NiLuJe:
Should fix #7390

In theory. In practice, CenterContainer is much better at utilizing the available width, whereas LeftContainer & RightContainer attempt to move text around to... weird effects, so we don't let 'em.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7391)

### DIFF
--- a/frontend/apps/reader/modules/readerfooter.lua
+++ b/frontend/apps/reader/modules/readerfooter.lua
@@ -736,11 +736,7 @@ function ReaderFooter:resetLayout(force_reset)
     if self.settings.disable_progress_bar then
         self.progress_bar.width = 0
     elseif self.settings.progress_bar_position then
-        if self.settings.progress_margin_width == 0 then
-            self.progress_bar.width = new_screen_width
-        else
-            self.progress_bar.width = math.floor(new_screen_width - 2 * self.settings.progress_margin_width)
-        end
+        self.progress_bar.width = math.floor(new_screen_width - 2 * self.settings.progress_margin_width)
     else
         -- In alongside mode, we don't allow custom progressbar margins.
         self.progress_bar.width = math.floor(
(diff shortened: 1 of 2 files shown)


NiLuJe:
Oh, joy, my favorite test in the world is unhappy.

frontend/apps/reader/modules/readerfooter.lua
```diff
@@ -1931,7 +1931,8 @@ function ReaderFooter:_updateFooterText(force_repaint, force_recompute)
             if self.settings.align == "center" then
                 self.footer_text:setMaxWidth(math.floor(self._saved_screen_width - 2 * self.settings.progress_margin_width))
             else
-                self.footer_text:setMaxWidth(self._saved_screen_width - 2 * self.horizontal_margin)
+                -- Otherwise, we have to constrain ourselves to the container, or weird shit happens.
```
NiLuJe, line 1934: Technical term. :D.

poire-z (approved):
Looks allright (except for a iffy "iff text is centered" somewhere).

NiLuJe:
iff == if and only if ;)

poire-z:
English, Higher Mathematical Jargon, this is getting challenging :)
But ok, https://en.wikipedia.org/wiki/If_and_only_if

Frenzie:
I learned about it in high school math. My math teacher was jealous they could express it so succinctly in English.

PR #7395, opened 2021-03-07
Hold "Wi-Fi" menu to show network selection instead of auto-connecting

snelg:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7395)

### DIFF
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -112,15 +112,25 @@ function NetworkMgr:releaseIP() end
 function NetworkMgr:restoreWifiAsync() end
 -- End of device specific methods
 
+function NetworkMgr:toggleWifiOn(complete_callback, long_press)
+    self.wifi_was_on = true
+    G_reader_settings:makeTrue("wifi_was_on")
+    self.wifi_toggle_long_press = long_press
+    self:turnOnWifi(complete_callback)
+end
+
+function NetworkMgr:toggleWifiOff(complete_callback)
+    self.wifi_was_on = false
+    G_reader_settings:makeFalse("wifi_was_on")
+    self:turnOffWifi(complete_callback)
+end
+
 function NetworkMgr:promptWifiOn(complete_callback, long_press)
     UIManager:show(ConfirmBox:new{
         text = _("Do you want to turn on Wi-Fi?"),
         ok_text = _("Turn on"),
         ok_callback = function()
-            self.wifi_was_on = true
-            G_reader_settings:makeTrue("wifi_was_on")
-            self.wifi_toggle_long_press = long_press
-            self:turnOnWifi(complete_callback)
+            self:toggleWifiOn(complete_callback, long_press)
         end,
     })
 end
@@ -130,9 +140,7 @@ function NetworkMgr:promptWifiOff(complete_callback)
         text = _("Do you want to turn off Wi-Fi?"),
         ok_text = _("Turn off"),
         ok_callback = function()
-            self.wifi_was_on = false
-            G_reader_settings:makeFalse("wifi_was_on")
-            self:turnOffWifi(complete_callback)
+            self:toggleWifiOff(complete_callback)
         end,
     })
 end
@@ -142,16 +150,11 @@ function NetworkMgr:promptWifi(complete_callback, long_press)
         text = _("Wi-Fi is enabled, but you're currently not connected to a network.\nHow would you like to proceed?"),
         choice1_text = _("Turn Wi-Fi off"),
         choice1_callback = function()
-            self.wifi_was_on = false
-            G_reader_settings:makeFalse("wifi_was_on")
-            self:turnOffWifi(complete_callback)
+            self:toggleWifiOff(complete_callback)
         end,
         choice2_text = _("Connect"),
         choice2_callback = function()
-            self.wifi_was_on = true
-            G_reader_settings:makeTrue("wifi_was_on")
-            self.wifi_toggle_long_press = long_press
-            self:turnOnWifi(complete_callback)
+            self:toggleWifiOn(complete_callback, long_press)
         end,
     })
 end
@@ -395,11 +398,12 @@ function NetworkMgr:getWifiToggleMenuTable()
             end
         end
         if fully_connected then
-            NetworkMgr:promptWifiOff(complete_callback)
+            NetworkMgr:toggleWifiOff(complete_callback)
         elseif is_wifi_on and not is_connected then
+            -- ask whether user wants to connect or turn off wifi
             NetworkMgr:promptWifi(complete_callback, long_press)
         else
-            NetworkMgr:promptWifiOn(complete_callback, long_press)
+            NetworkMgr:toggleWifiOn(complete_callback, long_press)
         end
     end
 
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -112,13 +112,14 @@ function NetworkMgr:releaseIP() end
 function NetworkMgr:restoreWifiAsync() end
 -- End of device specific methods
 
-function NetworkMgr:promptWifiOn(complete_callback)
+function NetworkMgr:promptWifiOn(complete_callback, long_press)
     UIManager:show(ConfirmBox:new{
         text = _("Do you want to turn on Wi-Fi?"),
         ok_text = _("Turn on"),
         ok_callback = function()
             self.wifi_was_on = true
             G_reader_settings:makeTrue("wifi_was_on")
+            self.wifi_toggle_long_press = long_press
             self:turnOnWifi(complete_callback)
         end,
     })
@@ -136,7 +137,7 @@ function NetworkMgr:promptWifiOff(complete_callback)
     })
 end
 
-function NetworkMgr:promptWifi(complete_callback)
+function NetworkMgr:promptWifi(complete_callback, long_press)
     UIManager:show(MultiConfirmBox:new{
         text = _("Wi-Fi is enabled, but you're currently not connected to a network.\nHow would you like to proceed?"),
         choice1_text = _("Turn Wi-Fi off"),
@@ -149,6 +150,7 @@ function NetworkMgr:promptWifi(complete_callback)
         choice2_callback = function()
             self.wifi_was_on = true
             G_reader_settings:makeTrue("wifi_was_on")
+            self.wifi_toggle_long_press = long_press
             self:turnOnWifi(complete_callback)
         end,
     })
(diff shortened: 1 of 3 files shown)


poire-z:
I'm not really familiar with NetworkMgr and with its interaction with the various devices code.

But it looks like you are hacking/adding stuff to NetworkMgr that are not really Remarkable specific - doing it only if running on Remarkable (so, a hack for you happy fews only :).
It seems the kind of "Hold" handling could just be put into NetworkMgr - not needing a hack monkey-patching functions, and it would be available for all devices.

Is my impression wrong ?

snelg:
I went with the reMarkable-specific monkey patch because the NetworkMgr implementation varies wildly between device types.
Half of them don't even use `showNetworkMenu` or `reconnectOrShowNetworkMenu` at all, which are the two functions I'm differentiating with long-vs-short press in this PR.

Since I don't know the details of what I might or might not break for *other* devices, I stuck with the one device I'm familiar with.
I basically just followed @pazos 's advice in https://github.com/koreader/koreader/pull/7122#issuecomment-757290352

As usual, I'm happy to adjust or completely rewrite my approach if someone wants to point me in a better direction.

NiLuJe:
IIRC, anything that uses NetworkMgr, well, uses it?

snelg:
How about this option?
https://github.com/snelg/koreader/commit/ad869b8355bcf3fb1b5c492c1e831f3e4b33ee52#diff-53bad97419097ca42d8959e50d149bf85fd75bbca2c323ed131c3b491d539596R408-R415
Change is smaller than it looks. I had to extract the existing `callback` function from the `return {` object so I could reference it twice.

The key point is highlighted: in NetworkMgr we set a `menu_long_press` variable to differentiate between short and long press of the "Wi-Fi connection" toggle menu button.
What each device does with that extra piece of information is still device-dependent.

If that looks reasonable, then I'll move that change into this PR.

poire-z:
(Still totally unfamilair with all that...) but it still looks hacky :) having this way back into the device specific code and forward into NetworkMgr.
Can't that be made fully into NetworkMgr, with may be just a device capability yes/no: if `Device:canDoThatNetworkLongPressStuff = yes/no` ?

snelg:
Ah, I see.
This PR now re-written; combined NetworkMgr `showNetworkMenu` and `reconnectOrShowNetworkMenu` into a single function (they were already very similar), and long-vs-short hold on the "Wi-Fi connection" menu determines whether to auto-connect, or to force the network list popup.

snelg:
Thus these changes are no longer reMarkable-specific :)

snelg:
"Ignore whitespace" diff of this PR is much simpler to read, BTW.

frontend/ui/network/manager.lua
```diff
@@ -600,44 +609,29 @@ function NetworkMgr:showNetworkMenu(complete_callback)
                 return
             end
         end
-        -- NOTE: Also supports a disconnect_callback, should we use it for something?
-        --       Tearing down Wi-Fi completely when tapping "disconnect" would feel a bit harsh, though...
```
NiLuJe, line 604: I'd preserve that comment in the remaining invocation ;).

NiLuJe:
Yep, looks better, I'll give it a try on Kobo.

As a quick generic comment, I'd *perhaps* invert the behavior: scan on tap, connect on hold? (RFC ;)).

frontend/ui/network/manager.lua
```diff
@@ -600,44 +609,29 @@ function NetworkMgr:showNetworkMenu(complete_callback)
...
+
         local success = false
-        for dummy, network in ipairs(network_list) do
-            if network.password then
-                success = NetworkMgr:authenticateNetwork(network)
-                if success then
-                    NetworkMgr:obtainIP()
-                    if complete_callback then
-                        complete_callback()
+        if not self.wifi_toggle_long_press then
```
poire-z, line 616: When here, are we sure we have always gone thru where you set it to true/false ? No risk to have this set because leaking from a previous (minutes ago) set to true/false ? No other place it would feel safe to reset it ? May be around here after it has been used?

poire-z (commented):
Looks better, thanks :)
There's just this `self.wifi_toggle_long_press` small-duration-state holder that feels risky. Would feel better as a value passed to the callbacks, but if it has to be carried thru many calls, it will look ugly and complex, and possibly more error prone - so may be this is just fine.

frontend/ui/network/manager.lua
NiLuJe: I was going to make the same comment as you did above, and, indeed, various other things can call `turnWifiOn` (which is the main caller for this function).

frontend/ui/network/manager.lua
snelg: Re-added

NiLuJe:
Okay, turns out Kobo is not a great platform to test this, because the Wi-Fi scripts will already auto-connect if the AP is saved in the wpa_supplicant config (which the native system does) ^^.

So, I'm pretty much always going to get an auto-connect w/ the AP list popping on and off ^^.

snelg:
> As a quick generic comment, I'd _perhaps_ invert the behavior: scan on tap, connect on hold? (RFC ;)).

99% of the time, when I say to any device, "Connect to wi-fi," I want it to just go ahead and re-connect to an already-known network. I don't want it to stop and ask me, "Here's the list of networks, which one do you want to connect to?"

NiLuJe:
Unrelated cleanup: there's a dprecated `table.getn(network_list)` call in `reconnectOrShowNetworkMenu`, could you replace that with a `#network_list` while you're in the vicinity? ;).

snelg:
> Okay, turns out Kobo is not a great platform to test this, because the Wi-Fi scripts will already auto-connect if the AP is saved in the wpa_supplicant config (which the native system does) ^^.

Ah, that probably explains why `showNetworkMenu` and `reconnectOrShowNetworkMenu` were separate in the first place.

Also a good example of why my initial version of this PR *only* fiddled with the reMarkable code, even though it was kinda hacky monkey-patching :P

NiLuJe:
And it turned out that a simple tweak takes care of that:

```diff
diff --git a/frontend/ui/network/manager.lua b/frontend/ui/network/manager.lua
index 60ce1541..0768c43d 100644
--- a/frontend/ui/network/manager.lua
+++ b/frontend/ui/network/manager.lua
@@ -628,6 +628,20 @@ function NetworkMgr:reconnectOrShowNetworkMenu(complete_callback)
                         })
                         break
                     end
+                elseif network.connected then
+                    -- On platforms where we use wpa_supplicant (if we're calling this, we are),
+                    -- the invocation will check its global config, and if an AP configured there is reachable,
+                    -- it'll already have connected to it on its own.
+                    success = true
+                    NetworkMgr:obtainIP()
+                    if complete_callback then
+                        complete_callback()
+                    end
+                    UIManager:show(InfoMessage:new{
+                        text = T(_("Connected to network %1"), BD.wrap(network.ssid)),
+                        timeout = 3,
+                    })
+                    break
                 end
             end
         end
```

NiLuJe:
Possibly ought to have checked those the other way around: e.g., if connected elseif password (i.e., honor native settings first, then koreader settings).

snelg:
> And it turned out that a simple tweak takes care of that:

Whoops, I just committed a third back-to-the-drawing-board version; leaves `showNetworkMenu` in its original form, only adding this "long_press" stuff to `reconnectOrShowNetworkMenu`.

Think I should revert my last commit (i.e. go back to the combined `reconnectOrShow`) and do your simple tweak instead?

frontend/ui/network/manager.lua
```diff
@@ -609,11 +607,31 @@ function NetworkMgr:reconnectOrShowNetworkMenu(complete_callback)
...
+            UIManager:show(InfoMessage:new{text = err})
+            return
+        end
         table.sort(network_list,
            function(l, r) return l.signal_quality > r.signal_quality end)
-
         local success = false
-        if not self.wifi_toggle_long_press then
+        if self.wifi_toggle_long_press then
+            self.wifi_toggle_long_press = false
```
NiLuJe, line 633: Possibly `nil`? (mostly for semantic/documentation concerns, makes it clearer that it's more about clearing a state variable that truly making the state false ;)).

NiLuJe: In which case it also wouldn't hurt to clear it unconditionally ;).

NiLuJe:
> > And it turned out that a simple tweak takes care of that:
> 
> Whoops, I just committed a third back-to-the-drawing-board version; leaves `showNetworkMenu` in its original form, only adding this "long_press" stuff to `reconnectOrShowNetworkMenu`.
> 
> Think I should revert my last commit (i.e. go back to the combined `reconnectOrShow`) and do your simple tweak instead?

Oh, no, my diff was completely orthogonal to the variable thingy ;).

That's just to take care of platforms where wpa_supplicant already auto-connects on its own ;).

NiLuJe:
> > Think I should revert my last commit (i.e. go back to the combined `reconnectOrShow`) and do your simple tweak instead?

Oh, I'd missed the platforms bits ;p.

Well, it certainly worked for me on Kobo with that tweak, so, I'd say yes ;).

----

But the lifecycle of the variable is still a concern, so the parameter passing and clearing approach still holds water ;).

snelg:
> Oh, no, my diff was completely orthogonal to the variable thingy ;).
> It's just to take care of platforms where wpa_supplicant already auto-connects on its own ;).

Right, but with this third re-write, kobo isn't even touching this code any more. It's back to using the original `showNetworkMenu`, which is unchanged in this version.

snelg:
> Well, it certainly worked for me on Kobo with that tweak, so, I'd say yes ;).

To clarify, do you mean "yes, *do* re-combine `showNetworkMenu` and `reconnectOrShowNetworkMenu`"?
Including both your tweak and my latest parameter-passing-and-clearing adjustment, of course.

NiLuJe:
> > Well, it certainly worked for me on Kobo with that tweak, so, I'd say yes ;).
> 
> To clarify, do you mean "yes, _do_ re-combine `showNetworkMenu` and `reconnectOrShowNetworkMenu`"?
> Including both your tweak and my latest parameter-passing-and-clearing adjustment, of course.

Yep, that's what I meant ;).

(And you can forget about my musing in https://github.com/koreader/koreader/pull/7395#issuecomment-793045664, handling "prefer <native/ko> over <ko/native> settings" properly would require looping twice. The list is sorted by link quality, and we have no visibility inside the native config besides wpa_supplicant being connected to *one* AP. Aaaany, I'm rambling. Posted diff is fine ;p).

snelg:
I think it's still okay to do the `if network.connected [...] elseif network.password` order. As you said, the list is sorted by signal strength. So even if there are multiple AP's, and even if they share an SSID, the logic should still connect to the strongest signal for which A) it's already connected or B) it has a password and can successfully connect.

That's what I committed before I saw your "ignore my musing" :)

NiLuJe:
Another tangentially related RFC: do we really need a prompt here? This is a menu entry,  99.9% of the time the answer is going to be "why, yes, I *do* want to do that, I've just tapped on a menu to do it!" ;).

NiLuJe (approved):
:lgtm:, thanks!

snelg:
> Another tangentially related RFC: do we really need a prompt here (for the clear on/off case, I mean ^^)? This is a menu entry, 99.9% of the time the answer is going to be "why, yes, I _do_ want to do that, I've just tapped on a menu to do it!" ;).

Excellent point. Just pushed another commit

PR #7396, opened 2021-03-07
A couple of fixes & enhancements

### LINKED ISSUE #7092: Clock on Nia is losing time
* KOReader version: 2020-12
* Device: Kobo Nia

#### Issue
Kobo Nia seems to lose contact with the internal clock.
When I start Koreader the clock is correct, the longer I use it the more it falls behind. Rebooting/going back to nickel and stuff like that always brings back the correct time. But staying inside Koreader the clock is too slow. It seems to depend on what I do, at some times it does not lose much at all, in some times it loses half of the real time, e.g. when I start it at 9am and read for 10 minutes the Koreader clock tells me it's 9:05 instead of 9:10.
This behavior is confirmed by other users in a German forum, and only for Kobo Nia.

### LINKED ISSUE #7394: menu.lua - chevrons remain after removing the last item

NiLuJe:
* Clock shenanigans on Kobo (fix #7092, pending https://github.com/koreader/koreader-base/pull/1322).
* Menu/KeyValuePage: Make sure the buttons are hidden when no items are shown (fix #7394).
* Screensaver: Use the ReaderUI instance when we can. Because re-opening the document you're currently reading just to get its cover felt fairly stupid...

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7396)

### DIFF
--- a/frontend/ui/screensaver.lua
+++ b/frontend/ui/screensaver.lua
@@ -107,8 +107,21 @@ local function expandSpecial(message, fallback)
     local authors = _("N/A")
     local series = _("N/A")
 
-    local instance = require("apps/reader/readerui"):_getRunningInstance()
-    if instance == nil and DocSettings:hasSidecarFile(lastfile) then
+    local ReaderUI = require("apps/reader/readerui")
+    local ui = ReaderUI:_getRunningInstance()
+    if ui then
+        -- If we have a ReaderUI instance, use it.
+        local doc = ui.document
+        currentpage = ui.view.state.page or currentpage
+        totalpages = doc:getPageCount() or totalpages
+        percent = Math.round((currentpage * 100) / totalpages)
+        local props = doc:getProps()
+        if props then
+            title = props.title and props.title ~= "" and props.title or title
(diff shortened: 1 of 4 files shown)


NiLuJe:
(I'll rebase with a base bump once https://github.com/koreader/koreader-base/pull/1322 lands).

frontend/ui/screensaver.lua
```diff
@@ -120,20 +133,6 @@ local function expandSpecial(message, fallback)
             authors = docinfo.data.doc_props.authors and docinfo.data.doc_props.authors ~= "" and docinfo.data.doc_props.authors or authors
             series = docinfo.data.doc_props.series and docinfo.data.doc_props.series ~= "" and docinfo.data.doc_props.series or series
         end
-        docinfo:close()
```
NiLuJe, line 123: That deletion is on purpose: DocSettings doesn't keep the file open, and close is a synonym for flush (i.e., dump the settings to the file).

Here, we're only reading, so that's completely unnecessary.

poire-z (approved):

PR #7397, opened 2021-03-08
Screensaver: Fix a fallback snafu

NiLuJe:
Regression since #7371 

(c.f., https://github.com/koreader/koreader/pull/7371#issuecomment-792397530)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7397)


NiLuJe:
`that *the* type setting is`

I'm always amazed at how many typos I can squeeze inside a commit message :D.

PR #7398, opened 2021-03-08
Standardize search/find to search

hius07:
There is `File search` in the Gesture manager already.
There is `Fulltext search` in the readermenu.

Some help text added.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7398)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -612,7 +612,13 @@ To:
 
     self.menu_items.find_file = {
         -- @translators Search for files by name.
-        text = _("Find a file"),
+        text = _("File search"),
+        help_text = _([[Search a book by filename in the current or home folder and its subfolders.
+
+Search for '*' will show all files.
+Search string supports Lua patterns.
+
+Tap a book in the search results to open it.]]),
         callback = function()
             self.ui:handleEvent(Event:new("ShowFileSearch"))
         end
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -615,7 +615,7 @@ To:
         text = _("File search"),
         help_text = _([[Search a book by filename in the current or home folder and its subfolders.
 
-Search for '*' will show all files.
+A search for '*' will show all files.
 Search string supports Lua patterns.
 
 Tap a book in the search results to open it.]]),
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -331,8 +331,7 @@ function ReaderDictionary:addToMainMenu(menu_items)
                         value_min = 8,
                         value_max = 32,
                         default_value = 20,
-                        ok_text = _("Set size"),
-                        title_text =  _("Dictionary font size"),
+                        title_text = _("Dictionary font size"),
                         callback = function(spin)
                             G_reader_settings:saveSetting("dict_font_size", spin.value)
                             if touchmenu_instance then touchmenu_instance:updateItems() end
@@ -678,7 +677,7 @@ end
 
 function ReaderDictionary:onShowDictionaryLookup()
     self.dictionary_lookup_dialog = InputDialog:new{
-        title = _("Enter a word to look up"),
+        title = _("Enter word(s) to look up"),
         input = "",
         input_type = "text",
         buttons = {
@@ -690,7 +689,7 @@ function ReaderDictionary:onShowDictionaryLookup()
                     end,
                 },
                 {
-                    text = _("Search dictionary"),
+                    text = _("Look up"),
                     is_enter_default = true,
                     callback = function()
                         UIManager:close(self.dictionary_lookup_dialog)
@@ -797,7 +796,7 @@ function ReaderDictionary:startSdcv(word, dict_names, fuzzy_search)
             {
                 dict = "",
                 word = word,
-                definition = lookup_cancelled and _("Dictionary lookup interrupted.") or _("No definition found."),
+                definition = lookup_cancelled and _("Dictionary lookup interrupted.") or _("No results."),
                 no_result = true,
                 lookup_cancelled = lookup_cancelled,
             }
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -677,7 +677,7 @@ end
 
 function ReaderDictionary:onShowDictionaryLookup()
     self.dictionary_lookup_dialog = InputDialog:new{
-        title = _("Enter search query"),
+        title = _("Enter a word or phrase to look up"),
         input = "",
         input_type = "text",
         buttons = {
@@ -689,7 +689,7 @@ function ReaderDictionary:onShowDictionaryLookup()
                     end,
                 },
                 {
-                    text = _("Look up"),
+                    text = _("Search dictionary"),
                     is_enter_default = true,
                     callback = function()
                         UIManager:close(self.dictionary_lookup_dialog)
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -677,7 +677,7 @@ end
 
 function ReaderDictionary:onShowDictionaryLookup()
     self.dictionary_lookup_dialog = InputDialog:new{
-        title = _("Enter word(s) to look up"),
+        title = _("Enter search query"),
         input = "",
         input_type = "text",
         buttons = {
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -37,7 +37,7 @@ end
 
 function ReaderWikipedia:lookupInput()
     self.input_dialog = InputDialog:new{
-        title = _("Enter search query"),
+        title = _("Enter a word or phrase to look up"),
         input = "",
         input_type = "text",
         buttons = {
@@ -49,7 +49,7 @@ function ReaderWikipedia:lookupInput()
                     end,
                 },
                 {
-                    text = _("Look up"),
+                    text = _("Search Wikipedia"),
                     is_enter_default = true,
                     callback = function()
                         UIManager:close(self.input_dialog)
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -37,7 +37,7 @@ end
 
 function ReaderWikipedia:lookupInput()
     self.input_dialog = InputDialog:new{
-        title = _("Enter words to look up on Wikipedia"),
+        title = _("Enter word(s) to look up"),
         input = "",
         input_type = "text",
         buttons = {
@@ -49,7 +49,7 @@ function ReaderWikipedia:lookupInput()
                     end,
                 },
                 {
-                    text = _("Search Wikipedia"),
+                    text = _("Look up"),
                     is_enter_default = true,
                     callback = function()
                         UIManager:close(self.input_dialog)
@@ -249,7 +249,7 @@ function ReaderWikipedia:addToMainMenu(menu_items)
                             local text = _([[
 Wikipedia articles can be saved as an EPUB for more comfortable reading.
 
-You can select an existing directory, or use a default directory named "Wikipedia" in your reader's home folder.
+You can select an existing folder, or use a default folder named "Wikipedia" in your reader's home folder.
 
 Where do you want them saved?]])
                             UIManager:show(ConfirmBox:new{
@@ -442,7 +442,7 @@ function ReaderWikipedia:lookupWikipedia(word, is_sane, box, get_fullpage, force
     else
         self.lookup_msg = T(_("Searching Wikipedia %2 for:\n%1"), "%1", lang:upper())
         req_failure_text = _("Failed searching Wikipedia.")
-        no_result_text = _("No Wikipedia articles matching search term.")
+        no_result_text = _("No results.")
     end
     self:showLookupInfo(display_word)
 
--- a/frontend/apps/reader/modules/readerwikipedia.lua
+++ b/frontend/apps/reader/modules/readerwikipedia.lua
@@ -37,7 +37,7 @@ end
 
 function ReaderWikipedia:lookupInput()
     self.input_dialog = InputDialog:new{
-        title = _("Enter word(s) to look up"),
+        title = _("Enter search query"),
         input = "",
         input_type = "text",
         buttons = {
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -347,7 +347,7 @@ function DictQuickLookup:init()
                                            or require("apps/filemanager/filemanagerutil").getDefaultDir() end
                         if not dir or not util.pathExists(dir) then
                             UIManager:show(InfoMessage:new{
-                                text = _("No directory to save article to could be found."),
+                                text = _("No folder to save article to could be found."),
                             })
                             return
                         end
@@ -1186,7 +1186,7 @@ end
 
 function DictQuickLookup:lookupInputWord(hint)
     self.input_dialog = InputDialog:new{
-        title = _("Input lookup word"),
+        title = _("Enter word(s) to look up"),
         input = hint,
         input_hint = hint or "",
         input_type = "text",
@@ -1199,7 +1199,7 @@ function DictQuickLookup:lookupInputWord(hint)
                     end,
                 },
                 {
-                    text = _("Lookup"),
+                    text = _("Look up"),
                     is_enter_default = true,
                     callback = function()
                         self:closeInputDialog()
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -1186,7 +1186,7 @@ end
 
 function DictQuickLookup:lookupInputWord(hint)
     self.input_dialog = InputDialog:new{
-        title = _("Enter search query"),
+        title = _("Enter a word or phrase to look up"),
         input = hint,
         input_hint = hint or "",
         input_type = "text",
@@ -1199,7 +1199,11 @@ function DictQuickLookup:lookupInputWord(hint)
                     end,
                 },
                 {
-                    text = _("Look up"),
+                    if self.is_wiki then
+                        text = _("Search Wikipedia")
+                    else
+                        text = _("Search dictionary")
+                    end,
                     is_enter_default = true,
                     callback = function()
                         self:closeInputDialog()
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -1186,7 +1186,7 @@ end
 
 function DictQuickLookup:lookupInputWord(hint)
     self.input_dialog = InputDialog:new{
-        title = _("Enter word(s) to look up"),
+        title = _("Enter search query"),
         input = hint,
         input_hint = hint or "",
         input_type = "text",
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -1199,11 +1199,7 @@ function DictQuickLookup:lookupInputWord(hint)
                     end,
                 },
                 {
-                    if self.is_wiki then
-                        text = _("Search Wikipedia")
-                    else
-                        text = _("Search dictionary")
-                    end,
+                    text = self.is_wiki and _("Search Wikipedia") or _("Search dictionary"),
                     is_enter_default = true,
                     callback = function()
                         self:closeInputDialog()
--- a/plugins/calibre.koplugin/main.lua
+++ b/plugins/calibre.koplugin/main.lua
@@ -59,9 +59,9 @@ function Calibre:closeWirelessConnection()
 end
 
 function Calibre:onDispatcherRegisterActions()
-    Dispatcher:registerAction("calibre_search", { category="none", event="CalibreSearch", title=_("Search in calibre metadata"), device=true,})
-    Dispatcher:registerAction("calibre_browse_tags", { category="none", event="CalibreBrowseTags", title=_("Browse all calibre tags"), device=true,})
-    Dispatcher:registerAction("calibre_browse_series", { category="none", event="CalibreBrowseSeries", title=_("Browse all calibre series"), device=true, separator=true,})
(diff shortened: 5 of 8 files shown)


frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -612,7 +612,13 @@ To:
 
     self.menu_items.find_file = {
         -- @translators Search for files by name.
-        text = _("Find a file"),
+        text = _("File search"),
+        help_text = _([[Search a book by filename in the current or home folder and its subfolders.
+
+Search for '*' will show all files.
```
NiLuJe, line 618:
```suggestion
A search for '*' will show all files.
```

Frenzie:
NB

Screenshot_2021-03-08-09-05-47-21_1aff61e6bd981aaa4ba9d0245eb6c04f

hius07:
> NB

`Calibre search` would be clear and unified.
https://github.com/koreader/koreader/blob/0674e6060aca0b88ea76b3b056db3918ccd71786/plugins/calibre.koplugin/main.lua#L112

Frenzie:
I would keep `Find a file` and change it to `Find in text`.

hius07:
Not a simple question)
https://www.dictionary.com/browse/find
https://www.dictionary.com/browse/search
https://www.differencebetween.com/difference-between-search-and-vs-find-and-vs-seek/
Chrome ctrl+f description: 'Open the Find Bar to search the current page"

Based on the definition, `Find` seems to be a little bit more accurate.
In the meantime, there are many `Search results` window headers. Can it be unified somehow?

Frenzie:
The equivalent with search would be something like this:

* Search for a file
* Search in text

> In the meantime, there are many Search results window headers. Can it be unified somehow?

Most things in the menu are very much on purpose. Plenty of headers are legacy.

hius07:
Kindle uses `Search` everywhere: in text, book names on device, Kindle Store.

These are the similar activities (from a user's point of view):
`Text search`
`File search`
`Calibre search`
`Wikipedia search`
`Dictionary search`
And all of them give `Search results`.



Find, search, lookup, look up - the Occam's razor is needed.

poire-z:
(No real opinion about these - I think I'd like this uniformity, but I don't know if we should really go toward radical uniformity - and impoverish vocabulary and have users learn/get used to poor english expressions :) From French, I would naturally use "search a word in a dict" - but it seems to me that "look up a word in a dict" feels like better english - and I might now use that after seeing it that often in KOReader :)

Frenzie:
```
Text search
File search
Calibre search
Wikipedia search
Dictionary search
```

Find/search is fairly immaterial. In my mind `Find` is the canonical KOReader one. But generally it should be an action.

Find/search in text
Find/search for (a) file
Find/search in calibre catalog
Find/search in Wikipedia
Find/search in dictionary

Incidentally, the Apple dictionary app uses all three. 

* Type a word to look up in…
* Search
* Find → Search For a New Word…

hius07:
We can see the following entries:

1 **Text**
1.1 `Fulltext search` main menu
1.2 `Enter text to search for` dialog header
1.3 `Search` button in the context (highlight) menu
1.4 `Fulltext search` dispatcher

2 **File**
2.1 `Find a file` main menu
2.2 `Search for books by filename` dialog header
2.3 `Search Results` window header
2.4 `File search` dispatcher

3 **OPDS catalog**
3.1 `Search` everywhere

4 **Dictionary**
4.1 `Dictionary lookup` main menu
4.2 `Enable fuzzy search` dictionary settings
4.3 `Enter a word to look up` main dailog header
4.4 `Search dictionary` main dialog button
4.5 `Search` dictionary article button
4.6 `Input lookup word` additional dialog header
4.7 `Lookup` additional dialog button
4.8 `Dictionary lookup` dispatcher

5 **Wikipedia**
5.1 `Wikipedia lookup` main menu
5.2 `Show image in search results` Wikipedia settings
5.3 `Enter words to look up on Wikipedia` main dailog header
5.4 `Search Wikipedia` main dialog button
5.5 `Searching Wikipedia` infomessage
5.6 `Input lookup word` additional dialog header
5.7 `Lookup` additional dialog button
5.8 `Wikipedia lookup` dispatcher

6 **Calibre**
6.1 `Find a book via calibre metadata` main menu
6.2 `Search settings` tools Calibre menu, 5 more `search` entries within
6.3 `Search in calibre metadata` dispatcher

To summarize:
1. `Find` is used in the main menu for Files and Calibre search only (2 entries).
2. `Lookup`/`look up` is used in the main menu and one dialog for Dictionary and Wikipedia.
3. `Search` is used for all settings and results in Dictionary and Wikipedia search.
4. `Search` is prevailing mostly everywhere (so far).

Hope this helps you, the code owners, to make the final decision on terminology unification.

hius07:
I would be happy to provide further unification based on two decisions:
- "Search" or "Find"?
- keep "lookup" for Wiki and Dictionary?

poire-z:
I don't use the 2 "Find..." so I won't notice the change - I'm fine with the current "Search" ones.
I'm also fine with the "lookup" where it's used (the switch to "search" feels not necessary to me - may be just because I'm used to them :)

hius07:
`Search` - a noun and a verb, `lookup` - a noun, `Look up` - a verb.
`Enter smth to search for` in all InputDialog titles.
`No results` everywhere when nothing found.

Sorry, don't understand a conflict with opdsbrowser.lua.

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -678,7 +677,7 @@ end
 
 function ReaderDictionary:onShowDictionaryLookup()
     self.dictionary_lookup_dialog = InputDialog:new{
-        title = _("Enter a word to look up"),
+        title = _("Enter word(s) to look up"),
```
Frenzie, line 680: I don't like this change. It's very unnatural.

plugins/calibre.koplugin/main.lua
```diff
@@ -59,9 +59,9 @@ function Calibre:closeWirelessConnection()
 end
 
 function Calibre:onDispatcherRegisterActions()
-    Dispatcher:registerAction("calibre_search", { category="none", event="CalibreSearch", title=_("Search in calibre metadata"), device=true,})
-    Dispatcher:registerAction("calibre_browse_tags", { category="none", event="CalibreBrowseTags", title=_("Browse all calibre tags"), device=true,})
-    Dispatcher:registerAction("calibre_browse_series", { category="none", event="CalibreBrowseSeries", title=_("Browse all calibre series"), device=true, separator=true,})
+    Dispatcher:registerAction("calibre_search", { category="none", event="CalibreSearch", title=_("Calibre search"), device=true,})
```
Frenzie, line 62: Some calibre nitpickers will be upset.

Frenzie: Concretely, I mean it's purposefully calibre. That's not a typo or a mistake.

Frenzie (commented):
Thanks!

plugins/opds.koplugin/opdsbrowser.lua
```diff
@@ -739,7 +737,7 @@ function OPDSBrowser:browseSearchable(browse_url, username, password)
 end
 
 function OPDSBrowser:onMenuSelect(item)
-    self.catalog_title = item.text or _("OPDS Catalog")
+    self.catalog_title = item.text or _("OPDS catalog")
```
Frenzie, line 740: I resolved it but I can't seem to push to your repo. Did you disable push access for repo owners when creating the PR?

Anyway, the easiest way to resolve in lieu of command line git is to undo this change:

```suggestion
    self.catalog_title = item.text or _("OPDS Catalog")
```

Then click something like "merge with master" which will probably be an option in the PR once you do that to bring it up to date, and re-apply on the current code.

Frenzie: See https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

frontend/apps/reader/modules/readerdictionary.lua
hius07: I can look up `upside down` (2 words) in English-Russian dictionary and get results.

plugins/calibre.koplugin/main.lua
hius07: Sorry, you are absolutely right, this should be reverted.

frontend/apps/reader/modules/readerdictionary.lua
NiLuJe: True, but the *wording* is weird ;).

The same meaning might be able to come across another way, possibly something based on "search term", which is understood to mean word/words/phrase ;).

hius07: "Query" is perfect, it is duplicated in the lookup results window.
`Enter search query` for title, ok-button remains `Look up`.

Frenzie: > I can look up upside down (2 words) in English-Russian dictionary and get results.

The proper term is `lemma` but the better solution is `word or phrase`.

plugins/calibre.koplugin/main.lua
```diff
@@ -109,7 +109,7 @@ function Calibre:addToMainMenu(menu_items)
     -- insert the metadata search
     if G_reader_settings:isTrue("calibre_search_from_reader") or not self.ui.view then
         menu_items.find_book_in_calibre_catalog = {
-            text = _("Find a book via calibre metadata"),
+            text = _("Calibre search"),
```
Frenzie, line 112: While I lean more toward "our style guide says to start a phrase or fragment with an uppercase character," I think `calibre search` would also be included in that.

@NiLuJe probably has a better finger on the calibre pulse.

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -1186,7 +1186,7 @@ end
 
 function DictQuickLookup:lookupInputWord(hint)
     self.input_dialog = InputDialog:new{
-        title = _("Input lookup word"),
+        title = _("Enter search query"),
```
Frenzie, line 1189: "Enter search query" is very Computer English. I'm not a fan.
```suggestion
        title = _("Enter a word or phrase to look up"),
```

plugins/calibre.koplugin/main.lua
NiLuJe: There's a bit of a catch-22 in that the official spelling of the app is actually `calibre` (lowercase), which has a tendency to drive me mildly nuts ;p.

NiLuJe: In any case, in this context, I feel like that would be a bit *too* short, anyway ;).

And people are already confused enough that I think the current spelling is perfectly fine?

frontend/apps/reader/modules/readerdictionary.lua
```diff
@@ -690,7 +689,7 @@ function ReaderDictionary:onShowDictionaryLookup()
                     end,
                 },
                 {
-                    text = _("Search dictionary"),
+                    text = _("Look up"),
```
NiLuJe, line 692: `Lookup` ^^.

I'm half of a mind to actually keep it simple and just use `Search`, actually.

frontend/apps/reader/modules/readerwikipedia.lua
```diff
@@ -49,7 +49,7 @@ function ReaderWikipedia:lookupInput()
                     end,
                 },
                 {
-                    text = _("Search Wikipedia"),
+                    text = _("Look up"),
```
NiLuJe, line 52: Ditto

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -1199,7 +1199,7 @@ function DictQuickLookup:lookupInputWord(hint)
                     end,
                 },
                 {
-                    text = _("Lookup"),
+                    text = _("Look up"),
```
NiLuJe, line 1202: Ditto

plugins/opds.koplugin/opdsbrowser.lua
```diff
@@ -706,13 +706,11 @@ end
 
 function OPDSBrowser:browseSearchable(browse_url, username, password)
     self.search_server_dialog = InputDialog:new{
-        title = _("Search OPDS catalog"),
+        title = _("Enter author or title to search for"),
```
NiLuJe, line 709: A bit reductive, you can arguably match on tags in a Calibre catalog.

(i.e. , I prefer the previous title + hint combo, possibly with minor tweaks).

hius07:
What do you think about unified lookup input dialog?


It's just a prototype, not a working code.

plugins/opds.koplugin/opdsbrowser.lua
hius07: > I prefer the previous title + hint combo

Do you mean `input_hint`?
There is just a `hint` in the code
https://github.com/koreader/koreader/blob/6d6760c2b3530b1469f127e6c06a3b49b3e1eca5/plugins/opds.koplugin/opdsbrowser.lua#L708
but I cannot find it in the InputDialog so I've removed it.

plugins/calibre.koplugin/main.lua
Frenzie: > In any case, in this context, I feel like that would be a bit too short, anyway ;).

You mean it should be at least `calibre metadata search` or just no change at all?

frontend/apps/reader/modules/readerdictionary.lua
Frenzie: No, `look up` the phrasal verb (as in `click look up to perform a lookup`).

> I'm half of a mind to actually keep it simple and just use Search, actually.

I'd prefer to just keep it `Search dictionary`.

frontend/apps/reader/modules/readerwikipedia.lua
Frenzie: Ditto indeed, I prefer Search Wikipedia.

Frenzie:
> What do you think about unified lookup input dialog?

I generally prefer simpler dialogs where possible, with a simple and obvious left = cancel, right = the action, without any middle.

That way the menu may be more complex but the dialog is simpler. It's a trade-off.

hius07:
Thank you.
The reason is that this input dialog is the same in all 3 modules: Dict, Wiki and search results DictQuickLookup.
BTW, what ok_button is recommended in DictQuickLookup (where ok_button action depends on the previous search)?

Now we have a very-long-press easter egg for switching between Dict and Wiki
https://github.com/koreader/koreader/blob/6d6760c2b3530b1469f127e6c06a3b49b3e1eca5/frontend/ui/widget/dictquicklookup.lua#L161
so the idea of domain switching is in the air.
Three buttons allow simple switching and show unified accustomed dialogs.

If not agreed, as a compromise it might be
`Search Dictionary`, `Search Wikipedia` and three buttons in DictQuickLookup.

Frenzie:
> The reason is that this input dialog is the same in all 3 modules: Dict, Wiki and search results DictQuickLookup.

I suspect it's not (except visually), but that's an interesting point.

hius07:
> except visually

Yeah, it's hard for a developer to stand at the average user's point of view. He knows what's under the hood.

hius07:
> what ok_button is recommended in DictQuickLookup (where ok_button action depends on the previous search)?

Answer myself: we can have the same `Dict/Wiki search` button, depending on domain.

hius07:
Since this PR is about wordings only, let's postpone redesign of the search input.
If anybody likes three buttons (for the purpose of simple domain switching), please let me know, I'll make the propositions.

plugins/opds.koplugin/opdsbrowser.lua
```diff
@@ -706,13 +706,11 @@ end
...
         input = "",
-        hint = _("Search string"),
         -- @translators: This is an input hint for something to search for in an OPDS catalog, namely a famous author everyone knows. It probably doesn't need to be localized, but this is just here in case another name or book title would be more appropriate outside of a European context.
         input_hint = _("Alexandre Dumas"),
         input_type = "string",
-        description = _("%s in url will be replaced by your input"),
```
Frenzie, line 715: This strikes me as rather important information?
Screenshot_2021-04-02_13-15-50

hius07: Assume I am not an IT man.
What is `%s`? In what url? I just want to search for Dan Brown.
And even if I understand the above, does it change my input?

Frenzie: Hold on, let me check the git blame for this one. I'm not very much at home in the OPDS plugin. Indeed it seems somewhat random here. But the point remains that it's important information *somewhere*. If it's here rather than where it should be, that may be an accident that should be investigated rather than removed wholesale.

If you add a URL like `https://www.bing.com/search?FORM=INCOH2&PC=IFJ1&PTAG=ICO-c9d0fc87&q=%s` then `%s` will be the encoded search terms. Leave out the %s and you have no search.

Frenzie: Context: https://github.com/koreader/koreader/pull/5823

poire-z:
> If anybody likes three buttons (for the purpose of simple domain switching), please let me know, I'll make the propositions.

When would we see it? Only when hitting the little pencil icon at top right of any lookup result (that I can figure from you screenshot), or elsewhere too?

plugins/opds.koplugin/opdsbrowser.lua
hius07: I remember it was redesigned. But...
https://github.com/koreader/koreader/pull/7380#issuecomment-792099693

hius07:
> When would we see it?

As we have one combined search results window (DictQuickLookup), we can have one combined input invoked from the main menu. Calling it from the DictQuickLookup allows switching domans for a new search.

hius07:
I'm done with wordings, all reprimands fixed. Thank you very much.

poire-z:
So, the little pencil icon + the menu. Is that all ?

When you say "main menu", I assume you mean the 2 entries `Dictionary lookup` and `Wikipedia lookup` , right ?

If these 2 would become a single menu item `Lookup`, it would be a little less clear in this menu, as we currently have 3 items related to dictionary, a separator, 3 same items related to Wikipedia.

Or do you see us keeping that existing menu, and both "lookup" items launching that same dialog with both buttons ? (If yes, do you see these buttons inverted, so the one related to the menu item hit is the right most one?)

hius07:
> So, the little pencil icon + the menu. Is that all ?

Yes, I thought about these 3 modules and to have just one main menu entry for lookup.
Let me think more over it, I see that the advantages (besides of input dialog unification) are no so obvious.

plugins/calibre.koplugin/main.lua
NiLuJe: `calibre metadata search` would work, FWIW, but I also have no issue with the current wording ;).

plugins/opds.koplugin/opdsbrowser.lua
NiLuJe: Possibly there was a typo in the field name, yeah, but I know the feature exists ;p.

plugins/opds.koplugin/opdsbrowser.lua
NiLuJe: IIRC, it puzzled me, and I checked, and I *think* it still makes sense is some edge-cases w/ legacy searches.

But it's weird.

NiLuJe: So, basically, what @Frenzie said, it makes sense in like 0.5% of the callers of this dialog ;p.

hius07: Thanks, have reverted already.

plugins/calibre.koplugin/main.lua
Frenzie: I slightly prefer calibre metadata search in the sense that it's more succinct, but no real preference

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -1199,7 +1199,11 @@ function DictQuickLookup:lookupInputWord(hint)
                     end,
                 },
                 {
-                    text = _("Lookup"),
+                    if self.is_wiki then
+                        text = _("Search Wikipedia")
+                    else
+                        text = _("Search dictionary")
+                    end,
```
NiLuJe, line 1206: Can't do that, you're in a table list initializer ;).

```lua
text = self.is_wiki and _("Search Wikipedia") or _("Search dictionary"),
```

might work, though ;).

hius07: Thank you

plugins/calibre.koplugin/search.lua
```diff
@@ -180,7 +180,7 @@ local CalibreSearch = InputContainer:new{
 
 function CalibreSearch:ShowSearch()
     self.search_dialog = InputDialog:new{
-        title = _("Search books"),
+        title = _("Enter string to search for"),
```
Frenzie, line 183: I prefer to keep programlish like `string` to the backend.

Frenzie: Since you like consistency ;-)
```suggestion
        title = _("Enter a word or phrase to look up"),
```

hius07: Seems 'look up' is not much applicable for calibre.
Maybe, return metadata?
`Enter text to search in metadata`
`Enter words to search in metadata`
Or get rid of "Enter...", just
`Calibre metadata search`

Frenzie: > Calibre metadata search

Sounds good to me, but pinging @NiLuJe since I don't use calibre.

NiLuJe: Yep, sounds good!

Frenzie: (for easy commit)
```suggestion
        title = _("Calibre metadata search"),
```

plugins/calibre.koplugin/main.lua
Frenzie:
```suggestion
            text = _("Calibre metadata search"),
```

plugins/calibre.koplugin/main.lua
```diff
@@ -59,7 +59,7 @@ function Calibre:closeWirelessConnection()
 end
 
 function Calibre:onDispatcherRegisterActions()
-    Dispatcher:registerAction("calibre_search", { category="none", event="CalibreSearch", title=_("Search in calibre metadata"), device=true,})
+    Dispatcher:registerAction("calibre_search", { category="none", event="CalibreSearch", title=_("Calibre search"), device=true,})
```
Frenzie, line 62:
```suggestion
    Dispatcher:registerAction("calibre_search", { category="none", event="CalibreSearch", title=_("Calibre metadata search"), device=true,})
```

hius07:
Great, thanks!

PR #7400, opened 2021-03-09
WakeupMgr: Minor usage tweak

NiLuJe:
* Unify logging with AutoSuspend (e.g., keep ourselves to showing the delay in seconds, not the raw timestamp, as that's way harder to interpret, and the RTC module and/or logger will do that for us when the time comes).
* Speaking of, minor revamp of RTC related logging to make it more human-readable.
* On Kobo, if we hit the unexpected wakeup limit, re-engage AutoSuspend's *suspend* check, so that the device has a chance to poweroff instead of being kept awake.

Depends on https://github.com/koreader/koreader-base/pull/1323

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7400)


Frenzie (approved):
Seems ok.

PR #7408, opened 2021-03-13
SkimToWidget: swap '1' and '10' buttons

hius07:
More logical buttons allocation, increasing from left to right.

https://github.com/koreader/koreader/wiki/Skim-document

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7408)

### DIFF
--- a/frontend/ui/widget/skimtowidget.lua
+++ b/frontend/ui/widget/skimtowidget.lua
@@ -251,15 +251,15 @@ function SkimToWidget:init()
     }
     local bottom_buttons_row = HorizontalGroup:new{
         align = "center",
-        button_minus,
-        small_button_span,
         button_minus_ten,
+        small_button_span,
+        button_minus,
         large_button_span,
         self.current_page_text,
         large_button_span,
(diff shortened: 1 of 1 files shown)


Frenzie:
I was very confused by how changing that would be more logical, but you already updated the picture in the wiki! :rofl: 

So it's

### Before
Screenshot_2021-03-13_13-10-23

### After
Screenshot_2021-03-13_13-10-03

Frenzie (approved):
Makes sense to me, waiting for second/third opinion though.

poire-z (approved):
OK for me too.

hius07:
> updated the picture in the wiki!

It's a new article, is it Okay, any corrections?

Frenzie:
That's cool. ^_^

It can be made to look more like a button by using `<kbd>▕◁</kbd>`, by the by. <kbd>▕◁</kbd>

hius07:
`<kbd>▕◁</kbd>`

Thank you, done, looks better.

PR #7411, opened 2021-03-14
Update PDF annotations when changing bookmark text

Toromtomtom:
This changes the PDF annotation contents each time the corresponding bookmark text is edited. When the annotations are written to the PDF file, the annotation contents are saved as well into the PDF.

In my initial tests it works - I can see the bookmark texts in the PDF viewer Okular. However, there are still some rough edges (partly due to my inexperience with the code base and with Lua):
- To touch as little existing code as possible, I duplicated some code. Once verified that everything works, some refactoring should be done.
- More error checking/reporting should be added.
- Tests should be added for the feature.

Depends on some [low-level additions to koreader-base](https://github.com/koreader/koreader-base/pull/1327).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7411)

### DIFF
--- a/frontend/apps/reader/modules/readerbookmark.lua
+++ b/frontend/apps/reader/modules/readerbookmark.lua
@@ -597,9 +597,11 @@ end
 function ReaderBookmark:renameBookmark(item, from_highlight)
     if from_highlight then
         -- Called by ReaderHighlight:editHighlight, we need to find the bookmark
+        local pboxes = item.pboxes
         for i=1, #self.bookmarks do
             if item.datetime == self.bookmarks[i].datetime and item.page == self.bookmarks[i].page then
                 item = self.bookmarks[i]
+                item.pboxes = pboxes
                 if item.text == nil or item.text == "" then
                     -- Make up bookmark text as done in onShowBookmark
                     local page = item.page
@@ -647,6 +649,10 @@ function ReaderBookmark:renameBookmark(item, from_highlight)
                                 if item.text == self.bookmarks[i].text and  item.pos0 == self.bookmarks[i].pos0 and
                                     item.pos1 == self.bookmarks[i].pos1 and item.page == self.bookmarks[i].page then
                                     self.bookmarks[i].text = value
+                                    local setting = G_reader_settings:readSetting("save_document")
+                                    if setting ~= "disable" then
+                                        self.ui.document:updateHighlightContents(item.page, item, value)
+                                    end
                                     UIManager:close(self.input)
                                     if not from_highlight then
                                         self.refresh()
--- a/frontend/document/pdfdocument.lua
+++ b/frontend/document/pdfdocument.lua
@@ -176,34 +176,11 @@ function PdfDocument:getPageLinks(pageno)
 end
 
 function PdfDocument:saveHighlight(pageno, item)
-    local suffix = util.getFileNameSuffix(self.file)
-    if string.lower(suffix) ~= "pdf" then return end
+    local can_write = self:_checkIfWritable()
+    if can_write ~= true then return can_write end
 
-    if self.is_writable == nil then
-        local handle = io.open(self.file, 'r+b')
-        self.is_writable = handle ~= nil
-        if handle then handle:close() end
-    end
-    if self.is_writable == false then
-        return false
-    end
     self.is_edited = true
-    -- will also need mupdf_h.lua to be evaluated once
-    -- but this is guaranteed at this point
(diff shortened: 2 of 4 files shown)


frontend/apps/reader/modules/readerbookmark.lua
```diff
@@ -597,9 +597,11 @@ end
 function ReaderBookmark:renameBookmark(item, from_highlight)
     if from_highlight then
         -- Called by ReaderHighlight:editHighlight, we need to find the bookmark
+        local pboxes = item.pboxes
         for i=1, #self.bookmarks do
             if item.datetime == self.bookmarks[i].datetime and item.page == self.bookmarks[i].page then
                 item = self.bookmarks[i]
+                item.pboxes = pboxes
```
Toromtomtom, line 604: We need to save the pboxes from the (highlight) item, as we need them later to identify the PDF annotation belonging to this bookmark.

poire-z:
I'm totally unfamiliar with PDF, annotations, highlight in PDF (and if these are different things) and how people use the "Save document" feature.
But I believe currently, any stuff that is to be written in the PDF gets handled at document quit time: that's where we get the prompt to write stuff into the PDF - so I guess that's also when it happens when people have disable prompt to use "always".
Was worried you would write stuff to disk on any wrong highlighy I could make, but I see that you just somehow cloned PdfDocument:saveHighlight() to do not much more than what it does :)
So, how does that work ? These 2 functions just add some stuff in memory when called (so, no I/O write to disk ?), and it's only at close time that things get flushed to disk by PdfDocument:writeDocument() ?

Toromtomtom:
> But I believe currently, any stuff that is to be written in the PDF gets handled at document quit time: that's where we get the prompt to write stuff into the PDF - so I guess that's also when it happens when people have disable prompt to use "always".
> Was worried you would write stuff to disk on any wrong highlighy I could make, but I see that you just somehow cloned PdfDocument:saveHighlight() to do not much more than what it does :)
> So, how does that work ? These 2 functions just add some stuff in memory when called (so, no I/O write to disk ?), and it's only at close time that things get flushed to disk by PdfDocument:writeDocument() ?

Yes, that seems to be the case. For example, I observed that, when I added/edited an annotation but killed the emulator before it wrote the file, the PDF file was unchanged.

The mupdf interface works on the in-memory representation of the PDF file. Interestingly, that in-memory representation also contains PDF annotations for which no highlight/bookmark in koreader exists. Those can in principle be edited as well, although that is quite confusing UI-wise. I therefore made sure that editing PDF annotations is only possible if a corresponding koreader highlight/bookmark exists.

Frenzie (approved):
I'm not really familiar with the API but it looks okay to me.

Toromtomtom:
I will try to do some further testing later today and polish this a bit. Then it will be ready for merging, I guess.

Toromtomtom:
The last commit fixed a crash when editing an annotation in a non-PDF document. From my point of view this is now ready to go.

I will try to do some refactoring and submit a new PR soon.

Toromtomtom:
Anything I can do to move this forward? It seems like this PR has already been approved, so I'm hesitant to make substantial changes now. Or shall I continue pushing changes? There is still some refactoring to be done.

Frenzie:
For me it doesn't really matter in this instance. I can merge it like this and then you can open a follow-up PR, or you could continue adding more commits to this PR.

Toromtomtom:
OK, this should now be good to go. 

I also added a function that deletes annotations from the in-memory representation of the PDF file when the corresponding highlight is removed. Before this, annotations that were added and then removed in koreader were being written to the PDF.

frontend/document/pdfdocument.lua
```diff
@@ -217,6 +194,66 @@ function PdfDocument:saveHighlight(pageno, item)
...
+    local can_write = self:_checkIfWritable()
+    if can_write ~= true then return can_write end
+
+    self.is_edited = true
+    local quadpoints, n = self:_quadpointsFromPboxes(item.pboxes)
+    local page = self._document:openPage(pageno)
+    local annot = page:getMarkupAnnotation(quadpoints, n)
+    if annot ~= nil then
+        page:deleteMarkupAnnotation(annot)
+    end
```
NiLuJe, line 207: `page` isn't `close`d here.

NiLuJe (changes requested):
Besides the `page` mishap, looks good, thanks!

NiLuJe (approved):
:lgtm:

Issue #7340, opened 2021-02-23
Slow (1-4s) page turns every 5-15 pages


angerson:
#### Device Details
- Device: Kobo Libra H2O (has page turn buttons)
- Koreader version: 2021.02 (and all previous versions)

#### Issue
Every once in a while, turning a page in KOreader takes 1-4 seconds. Usually incidents grow in frequency and length until I restart KOreader, at worst ~5 seconds each every 10-20 pages.

Weak points that I think I've noticed:

- Seemingly happens more often if RAM usage is very high
- Seems to be worse with complex fonts, *especially* EB Garamond (compared to Bookerly). I changed from Bookerly to EB Garamond and it started happening a lot more often.
- Usually happens just before a full screen refresh (black flash), or right after. Doesn't happen with every flashing refresh, though.
- I can't tell if it's related to page contents or not. Some pages seem to be more prone than others, if I scroll back and forth over them, but most of the time it just seems random.
- Doesn't seem to be related to dpi, page/continuous mode, Render mode, or font size
- Incidents seemed to go away for a while when I set Font Hinting `auto` -> `off` and Kerning `best` -> `off`, but then they came back just as badly.

Stats from last time it happened:
  - Book: 2.8MB, 1224 pages, epub, just text
  - Font: EB Garamond
  - Storage: 5.1G available of 6.7G
  - System Usage%: 1.16
  - Virtual Memory: 342MB
  - RAM usage: 132MB
  - Total Memory: 496MB
  - Free Memory: 134MB
  - Available Memory: 339.26MB

#### Reproduction
I can't guarantee this will reproduce, but try to read a long book with the smaller version of EB Garamond and turn the pages a lot.

NiLuJe:
Are those *single* page turns, are is that *a* page-turn in the middle of a series of page-turns?

(Or, for that matter, a series of *something* causing a refresh, e.g., if you have the footer set to auto-refresh).

NiLuJe:
Because, for the record: you can't expect *any* kind of latency guarantee during a continuous series of refresh. And the leeway between refreshes is *much* stricter on REAGL devices (as in, essentially: you don't have any leeway at all ^^. A REAGL update better stand on its own with nothing happening roughly a second before and after).

As a fun experiment, you'll notice that if you open a book, and right when it's refreshed the first page, you try to tap on the menu tapzone, the menu won't show up until a noticeable delay.
That wouldn't be the case on the same device when not using REAGL updates, the menu would popup instantly.

[...]

PR #7415, opened 2021-03-16
The great Input/GestureDetector/TimeVal spring cleanup (a.k.a., a saner main loop)

NiLuJe:
* Based on `<bsd/sys/time.h>`, which does it short and sweet, and in a way that makes detecting time going backwards much more sensible than whatever the previous iteration was trying to achieve. (And is also 700% more readable, c.f., the comparisons...).
* Took the opportunity to add a variant of `now` that uses `CLOCK_MONOTONIC`, and ported everything that actually expected a monotonic clock to that.
That's pretty much everything, in fact ;).

That means this affects old and core parts of the event loop logic, both in UIManager & Input/GestureDetector, so, here be dragons!

Requires https://github.com/koreader/koreader-base/pull/1332

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7415)


NiLuJe:
The code should be mostly ready, but I'll finish testing tomorrow (I've only thrown it at the emu for now) ;).

NiLuJe:
Whee, broken tests \o/.

NiLuJe:
MockTime & BackgroundRunner being annoying...

NiLuJe:
TODO: Use `CLOCK_MONOTONIC_COARSE` if possible (do a probe at import), it's more than twice as fast as `CLOCK_MONOTONIC` on devices with a vdso codepath, and still ~25% faster on older, crappier devices.

Resolution is 1ms instead of 1ns (which we truncate to 1µs anyway), which should still be more than enough for us.

(FWIW, `CLOCK_MONOTONIC` exhibits similar performance as `CLOCK_REALTIME`, which, itself is roughly on par w/ `gettimeofday`).

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -1122,8 +1122,8 @@ end
 function ReaderHighlight:onHoldRelease()
     local long_final_hold = false
     if self.hold_last_tv then
-        local hold_duration = TimeVal.now() - self.hold_last_tv
-        hold_duration = hold_duration.sec + hold_duration.usec/1000000
+        local hold_duration = TimeVal:monotonic() - self.hold_last_tv
```
poire-z, line 1125: Does this needs to be named `:monotonic()` ? When reading higher level code, `:now()` was a LOT MORE explicite and read like natural language.
"monotonic" might be some technical term/API that you use internally to get a TimeVal, but I think it should be kept internal, and higher level code to keep being readable :)

frontend/device/gesturedetector.lua
```diff
@@ -461,22 +468,9 @@ function GestureDetector:handleNonTap(tev)
...
-                -- assume they are real clock time
-                self._has_real_clock_time_ev_time = true
-                logger.info("event times are real clock time: no adjustment needed")
-            end
-        end
         logger.dbg("set up hold timer")
-        local ref_time = self._has_real_clock_time_ev_time and tev.timev or TimeVal:now()
-        local deadline = ref_time + TimeVal:new{
+        -- Ditto, start counting from *now*, no matter what the input event timestamp says...
+        local deadline = TimeVal:monotonic() + TimeVal:new{
```
poire-z, line 473: You're sure about all that ? (I'm not putting my mind again in all that, but thinking about 633d43b322 #6798, I'm wondering if you're not missing one of the 2 cases this was supposed to handle both correctly?)

poire-z (commented):
Hoping this will somehow solve the non-responsive touch I get once a week on my GloHD (that I suspect might be related to timestamp of events or something around that, possibly not hardware related).

frontend/device/gesturedetector.lua
NiLuJe: We actually care about computing deadlines based on the moment we *process* them, not whatever time the kernel said they came from.

frontend/apps/reader/modules/readerhighlight.lua
NiLuJe: The other option is making a gettime w/ a clockid parameter, which is possibly worse ;p.

frontend/device/gesturedetector.lua
poire-z: But wasn't there a thing with:
`self._has_real_clock_time_ev_time and tev.timev `
trusting and actually using the time from the event (which might be some time in the past, and not a "now") ?

Or is it just that part, and this disctinction is still used for other things around there? (sorry, too busy to look at the code right now)

frontend/apps/reader/modules/readerhighlight.lua
poire-z: Was just thinking about another option: give it a better less technical name :)
`:rightnow()` :) 
or why not just `:now()` ? Do we still need the current `:now()` elsewhere ?

frontend/device/gesturedetector.lua
poire-z: > We actually care about computing deadlines based on the moment we process them

I think this part of the code was actually meant to compute a deadline from the ev.time if we can trust it - and fallback to now() only when we don't trust it.

frontend/apps/reader/modules/readerhighlight.lua
NiLuJe: Yeah, I'm going to check that, and possibly rename now to `realtime` and alias `now` to `monotonic` if that's viable ;).

frontend/device/gesturedetector.lua
NiLuJe: My point is that trusting it doesn't make much sense to me right now, which is what makes me think that the point is moot ^^.

poire-z: > My point is that trusting it doesn't make much sense to me right now

I think it made sense _at the time_ :)

We get events, that may be queued because we're doing some other processing - then, at the time we can process them, some time has elapsed. Using "now" might increase the interval since the time the user made that event.
I think this part of the code handles double tap, tap intervals, etc... and using now() was an issue. Don't want to re-dig into all that. Some more context around https://github.com/koreader/koreader/pull/6798#issuecomment-711154547 maybe.

NiLuJe: This part is only dealing with the actual event loop deadline; gesture detection itself still trusts the event's timestamps, as it should. (e.g., we compare apples to apples (the event loop), and oranges to oranges (input events themsevles) ;).

poire-z: I'm not convinced :/
These 2 parts set up a delay (a max interval between current event A and a futur event B) after which we change some internal state (so we process that futur event B one way, or another, depending on its delay from event A).
Previously, we used the real event time (which you can make monotonic() if you want) which might be a few 100ms in the past (because we've been busy and our UI loop is catching up, unqueuing past events) as the start of the delay - AND ONLY now() when we knew this ev.time was bad.
So, if:
ev.time = 13h00m00s
now (monotonic if you want) = 13h00m03s
and we have a hold_interval of 5s, we want the timeout to be triggered at 13h00m05s, not at 13h00m08s.

NiLuJe: Sure, but is there actually a practical example of when this might matter?

poire-z: So, it's for me to look at the hyper long #6798, where the thing you quickly removed was painfully added, to find out why we ended up adding it ? :/ 

Looks like I thought like you at https://github.com/koreader/koreader/pull/6798#issuecomment-711139118 : 
> So, it's just when handling double tap and hold/long-press that we would have a possibly shift between the event timestamp and the now() used for the timeout.
When everything is not busy, these should be quite similar and this change should have no impact.
And when we long-press or do double tap, things shouldn't be busy (when double tap, the first tap does not trigger anything so no reason to get busy).

I let you read @ezdiy and your comments after it, that you'll probably understand better than I do.
I guess I just followed @ezdiy request to not drop event times, and to not change too much of how all that works as it was working fine.

My only comment after that, possibly telling what this first naive change (always using now()) might have caused, although I'm not certain I wrote it about that, is:

> Anyway, I still have some issues with swipe sometimes being detected as multiswipe... (small diagonal swipe to full refresh, working sometimes, sometimes not...)

--

> Sure, 

Glad you understand the issue.
Well, when things are accurate and can be kept accurate, and they can cause a delay and we're playing with 50-500 ms delays, I don't see why you would want to make them less accurate.
And furthermore, with this, you'll be adding some expensive os.realtime() call, while you could just benefit of the one computed at the time the event was made - that should appeal you :)

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -339,7 +339,7 @@ end
 -- to ensure current highlight has not already been cleared, and that we
 -- are not going to clear a new highlight
 function ReaderHighlight:getClearId()
-    self.clear_id = TimeVal.now() -- can act as a unique id
+    self.clear_id = {} -- The pointer address should be fairly unique
```
poire-z, line 342: Why play and add uncertainty ? It may _should be fairly unique_ - but it _might not be_ and a bit of memory re-used - may be once a year - and we'll once a year wonder wtf is this bug I can't reproduce :)
I'd rather have a counter, or something that makes it _really unique_.

NiLuJe: We'd need some *really* unlucky GC behavior to get a collision ^^.

But the reasoning behind the change was flawed anyway (In the heat of the moment, I, err, forgot about the `__eq` and assumed it was essentially doing this instead of a proper comparison ^^), c.f., 0e931d4 ;).

frontend/device/gesturedetector.lua
NiLuJe: I'll triple-check, but I'm of the mind that trusting the event's timestamp would be a red herring here ;).

And, so far, nothing's broke on a device with events using a REALTIME clock source.

NiLuJe: AFAICT, on the extremely unlikely case we're significantly behind in reading the input events, it doesn't really matter if the setTimeout deadlines are further than they would be if the event timestamps were honored:

* The gesture detection itself will use the event timestamps, as it should be, so the detection itself is always doing the right thing.
* The only difference is that the Gesture Event will be sent "later", which is potentially a *good* thing:
   Suppose we were trusting the events to compute the deadline: that means we can essentially setup a deadline in the *past*, which means it will possibly expire *before* we've consumed the following events that could make it true.

Now, hitting a stall will unwind the event queue in a consistent manner, and Gestures will be detected as if stuff was not late, instead of potentially dropping early gestures because we caught the events "late".
At worst we end up with a bunch of extra deadlines that'll all expire later than they should and do nothing, which is harmless.

It's somewhat hard to test in practice because nothing much really uses this codepath except the tap -> hold switch, but I really can't make it do anything obviously wrong, even with a slow-ass rotated PDF (on a device with realtime events).

NiLuJe: Now, if you remember something that's easy to break to stress-test this, I'm all ears ;).

NiLuJe: Even on a design level, setTimeout is operating on a completely different timescale: it's concerned with the event loop polling, not the events themselves. As such, make it use the same timescale as the UI just makes more sense that trying to shoehorn potentially unreliable input events timestamps in there.

NiLuJe: > 
>     * The only difference is that the Gesture Event will be sent "later", which is potentially a _good_ thing:
>       Suppose we were trusting the events to compute the deadline: that means we could essentially setup a deadline in the _past_, which means it will possibly expire _before_ we've consumed the following events that could make it true and actually detect the gesture it ought to.

Actually, not even that will differ: the `waitEvent` only checks the deadlines when the event queue has been completely drained (i.e., when the waitForEvent select times out).

Which means that we'll just finalize the Gesture detection "late" (relative to how "late" we were to begin with).

NiLuJe: There's also the fact that the first deadline hit that detects a gesture "wins", and clears *all* the deadlines.
Being "late" doesn't change that fact, though, so, basically still business as usual, it'll just clear more deadlines than usual.

NiLuJe: And since #6798 was concerned abound bounces: that works just fine (and doesn't even rely on setTimeout in the first place anyway :D).

NiLuJe: As for why gloves were necessary before this PR, as in #6798: that's because the event loop was using a timescale that it naively thought would match the input events', so precautions had to be taken in case mismatches were detected.

This PR takes a different approach: it enforces a monotonic timescale in the UI/event loop, making scheduling reliable; and treats the input events timestamps as sacred, but potentially unreliable and in a different timescale: as such, anything UI/scheduling related follows the same timescale (monotonic), and input events are handled in their own little world by themselves, never even attempting to mix the two time scale.

poire-z: Thanks for digging deeper into all that. I'm not following everything, but I trust your feeling it is ok/better.

> that means we could essentially setup a deadline in the past

Somehow hoping this might solve my issue of touch getting locked - and me being in a hold state in some cases when I recover.
So ok, let's go with it and see. May be for after 2021.03 ?

NiLuJe: Happy too, writing it down helps me understand what happens and whether the design decision made sense ;).

I would assume the "sticky" state on resume was because the gesture detection code did not deal with large time deltas very well. That *should* be fixed, yeah ;).

https://github.com/koreader/koreader/issues/7340#issuecomment-799710372 is what spurred me on, but since it'll only "crash" w/ debug enabled, I'm fine with postponing this for 2021.04 (assuming it lands this WE ;)) ;).

poire-z: Btw, taking a step further away: are setTimeout() the only solution to this double tap / long-press interval ? It seems really complicated when I had to understand it some months ago...
No simpler way to handle this, like just storing timestamp in variables and considering then when next event is coming, if they have expired or not, ... ? (Not asking you to revamp the whole thing, just thinking if there's room there to do it better later.) (although if it works... don't fix it :)

poire-z:
Btw, does this TimeVal rework fixes the whole insanity with interval/delay computations only doing their things with the `usecs` and ignoring `secs`, somehow limiting them to be < 999ms ?

frontend/device/gesturedetector.lua
NiLuJe: Yeah, preaching to the choir here ;p.

It indeed feels like an extremely roundabout way to detect a hold, but this is an old and somewhat key piece of code, and I'm not really up to rewriting gesturedetector ;p.

NiLuJe:
Yeah, that's what I meant [above](https://github.com/koreader/koreader/pull/7415#discussion_r595621157) by "large time deltas were not handled very well" ^p^.

I mean, the actual comparison metamethods were unreadable to begin with, and GestureDetector was... not using them and instead doing it by hand only for < 1s o_O.

NiLuJe:
Fiddling with the `input.waitForEvent` `select` timeout is a WIP, but it looks promising, and gets rid of the 100µs select loop monstrosity...

NiLuJe:
Whee, haven't broken Android :D.

NiLuJe:
Okay, I'm gonna do a final review pass, clear the debugging, and I think this will be ready :o}.

frontend/device/gesturedetector.lua
```diff
@@ -253,50 +276,56 @@ function GestureDetector:switchState(state_new, tev, param)
...
     elseif type == "ges_swipe_interval" then
-        return ges_swipe_interval
+        return ges_swipe_interval:tousecs()
     end
 end
 
 function GestureDetector:clearStates()
-    self:clearState(0)
-    self:clearState(1)
+    for k, _ in pairs(self.states) do
```
poire-z, line 326: ~~pairs or ipairs?~~ ok, 0 is not Lua. So, I guess pairs works when we set numbers 0 and 1.

frontend/device/gesturedetector.lua
```diff
@@ -461,30 +501,20 @@ function GestureDetector:handleNonTap(tev)
...
+        print("GestureDetector:handleNonTap: Updating pending_hold_timer for slot", slot)
+        -- Invalidate previous hold timers on that slot so that the following setTimeout will only react to *this* tapState.
+        self.input:clearTimeout(slot, "hold")
+        self.pending_hold_timer[slot] = true
+        self.input:setTimeout(slot, "hold", function()
+            print("GestureDetector: Hold finalization for slot", slot)
+            -- If the pending_hold_timer we set on our first switch to tapState on this slot (e.g., first finger down event),
+            -- back when the timer was setup, is still relevant (e.g., the slot wasn't run through clearState by a finger up gesture),
+            -- then check that we're still in a stationary finger down state (e.g., tapState).
+            if self.pending_hold_timer[slot] and self.states[slot] == self.tapState then
```
poire-z, line 516: You're sure this is ok ? Just having self.pending_hold_timer[slot] = true ?
You clear it, then reset-it to true: a previous setTimeout'ed will see it to true and will proceed.

The previous code was using a unique ID (tev.timev) to have a previous setTimeout just know it's no more its job.

poire-z (commented):
Generic question: is there any reason (except that we use it everywhere and it would be a lot of work to kill it everywhere) to use this TimeVal object?
Couldn't we just use a float like `1616174690.561417` everywhere, and forget about secs and usecs, and just use timeout_ms/1000 and work with a single number?

NiLuJe:
> Generic question: is there any reason (except that we use it everywhere and it would be a lot of work to kill it everywhere) to use this TimeVal object?
> Couldn't we just use a float like `1616174690.561417` everywhere, and forget about secs and usecs, and just use timeout_ms/1000 and work with a single number?

Most of the stuff that deals with TimeVal actually deals with... `timeval`s ;p. (As in, the POSIX data type). So it actually makes it *easier* to keep everything in the native format (besides the accuracy and overflow concerns ;)).

The very few things that ever use the :tonumber / :tousecs method are... logging (or `poll`, which takes a ms value for its timeout argument).

frontend/ui/uimanager.lua
```diff
@@ -1191,7 +1176,14 @@ function UIManager:_checkTasks()
...
 
-    return wait_until, now
+    return wait_until, self._now
+end
+
+--[[--
+Returns a TimeVal object corresponding to the last UI tick.
+]]
+function UIManager:getTime()
+    return self._now
```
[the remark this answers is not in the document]
NiLuJe: Yep, that's to cadge a syscall, mainly when either you don't need insane accuracy (it's updated by UIManager at the top of a tick, so it's never *really* late), or by stuff that's running *right* behind that.

It is very much what the description says: the timestamp of the current UI tick.

frontend/ui/uimanager.lua
```diff
@@ -1772,5 +1772,11 @@ function UIManager:restartKOReader()
     self._exit_code = 85
 end
 
+--- Sanely abort KOReader (e.g., exit sanely, but with a non-zero return code).
+function UIManager:abort()
+    self:quit()
+    self._exit_code = 1
```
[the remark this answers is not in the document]
NiLuJe: Yeah, I seem to remember there's a reason for that, but I don't quite recall it ^^. Other callers do the same, FWIW.

frontend/ui/timeval.lua
```diff
@@ -150,9 +143,70 @@ Creates a new TimeVal object based on the current time.
 
 @treturn TimeVal
 ]]
-function TimeVal:now()
+function TimeVal:realtime()
     local sec, usec = util.gettime()
     return TimeVal:new{sec = sec, usec = usec}
```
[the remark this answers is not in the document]
NiLuJe: It's unrelated, on Linux, it's an ever-increasing clock (i.e., it will *never* go backwards, it may *only* report the exact same value multiple times), starting at 0 on the current cold boot, that stops ticking during suspend.

So, yeah, it's there's for feature-parity and documentation and in case some crazy person ever wants to try wall-to-monotic conversions ;).

frontend/device/gesturedetector.lua
NiLuJe: Yep, it's a hash ;).

frontend/device/gesturedetector.lua
NiLuJe: > You clear it, then reset-it to true: a previous setTimeout'ed will see it to true and will proceed.

But it won't, because the clearTimeout means that there's no longer a previous setTimeout ;).

poire-z: Oh, I see, you tweaked setTimeout so there are no longer "anynoymous function", but they get a name/slot internally.
And well, I guess I thought this was managed by UIManager like what we scheduleIn() - so, ok.

frontend/ui/timeval.lua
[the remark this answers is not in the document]
poire-z: So, which should we use if one day, we wan't to store the data a bookmark or highlight is made as a timestamp (to be converted to a human YYYY-MM-DD date at display time)?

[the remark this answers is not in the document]
Frenzie: I'd be inclined to suggest ISO 8601.

frontend/ui/uimanager.lua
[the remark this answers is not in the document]
poire-z: > updated by UIManager at the top of a tick

You have to tell me what's a tick :) I never needed to know until now, I always just cared about future ticks, never a past one :)
When it's sleeping, and i touch the screen/hit a key, is this tick set ? Or only later when it goes again in the UI idle loop?

poire-z:
> So it actually makes it easier to keep everything in the native format 

Except that you have to translate it and allocate a Lua table to hold it, instead of just a number - which I think might cost more than a a syscall + a few multiplication/division to get a number.

frontend/ui/uimanager.lua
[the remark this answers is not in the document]
NiLuJe: If we're in Lua-land, we're in an UI tick  one way or another (so, sometime "soon" after this has been set).

The cycle goes:

* tick
* check scheduled tasks to compute poll timeout
* poll for input <may wait up to whatever was decided above or the next input event, whichever comes first>
* process *one* input event
* back to the top

This is why basically everything uses getTime (or can use it without thinking too much about it) *but* the input code, which uses now(): it's the only bit of code that may be run *significantly* later than the last time this was set ;).

frontend/ui/timeval.lua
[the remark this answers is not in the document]
NiLuJe: The statistics plugin already uses wall clock time for this very reason (and, in fact, is pretty much the main bit of code left that does that explicitly ;)).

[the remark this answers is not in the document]
NiLuJe: (e.g., os.time(), util.gettime() or TimeVal:realtime()).

frontend/device/gesturedetector.lua
NiLuJe: Yeah, that allowed to clean them up early based on what they were for ;).

(Granted, the only time I actually could end up with a stack of two or three pending hold timers that would never ever be true was mashing on the virtual keyboard like a crazy person ^^).

NiLuJe:
If we're worrying about the cost of allocating a Lua table, we're in deep trouble: everything's a table ^^.

frontend/device/gesturedetector.lua
poire-z: Found the PR I made with these timer ids, where I explained the issue I had and how to reproduce it: #6186

frontend/ui/timeval.lua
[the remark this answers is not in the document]
poire-z: > (e.g., os.time(), util.gettime() or TimeVal:realtime()).

I wouldn't mind a comment somewhere stating that all these 3 return the same kind of time, unlike all the others TimeVal methods that are to be used only for internal time/delay/interval management.

frontend/ui/uimanager.lua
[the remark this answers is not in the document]
poire-z: > - tick
> - check scheduled tasks to compute poll timeout
> - poll for input <may wait up to whatever was decided above or the next input event, whichever comes first>
> - process one input event

With that order, at the moment this input event is processed, we may get a time from the tick that is a few seconds/minutes away (if no poll timeout) , right ?
I understand that before actually using that time to do something with it, we may get some dialogs/UI stuff and we'll get a more recent tick time to work with - but still :) It feels strange that a getTime() could give something staled.

You could update _now between the last 2 steps above, but that's useless if it's not used. Might be better to have UIManager:getTime() get a fresh one - unless it's only for internal UIManager use - but it's used in userland. I wouldn't know which of UIManager/util :getTime to use, I'd probably trust more util from now on :) but strange to make them spit out different things.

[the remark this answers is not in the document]
NiLuJe: The TL;DR is: it's fine for userland.

Only input code has to be careful and request an actual up-to-date time.

NiLuJe:
As a practical example, see the UIManager changes, where moving from a mix of weird tables and µs values to TimeVal all the way made everything that less crazy ;).

frontend/device/gesturedetector.lua
NiLuJe: Ha, thanks ;).

Yep, still fixed (I did manage to inadvertently set the screenshot as the screensaver, which should prove that it behaves ;p).

poire-z: Can you reproduce it if you remove your timeout clearing ?

poire-z:
> As a practical example, see the UIManager changes, where moving from a mix of weird tables and µs values to TimeVal all the way made everything that less crazy ;).

I know the previous stuff was ugly. But it would look as nice (and shorter) if we were using just numbers :) and not have to split things into secs and usecs.
I'm not asking we go the way of numbers - just asking if there's something with our superpowered TimeVal (except that it maps nicely to a tm_something C struct) that would not be possible with just using numbers and number substration/comparisons - if it would be more than just minor replacements at each place. Just asking :)

NiLuJe:
We potentially hit range issues and accuracy woes in the conversions with a float, yes.

frontend/device/gesturedetector.lua
NiLuJe: Most likely, yes.

NiLuJe:
I, err, may have come up with a timerfd backend for the setTimeout mess ;D.

(In short, ask the kernel to wake us up instead of computing deadlines like crazy persons).

This allows to keep the computation in the event's timescale (because the kernel is much better at timekeeping than us ;))... provided we've identified the clock source of tap events properly.
Hence the return of a potentially nasty probe function. Seems to behave on my end, but then I don't really have a device that spits out MONOTONIC clock events, so I'll just have to fake that via ioctl on the Forma.

The C bit if still a bit PoC'ish, but the Lua side of the API should be okay (and required only minimal changes).

NiLuJe:
```
[pid  7437] 00:27:19.173427 [2ac7ae90] _newselect(13, [6</dev/input/event0> 7</dev/input/event1> 12<anon_inode:[timerfd]>], NULL, NULL, {tv_sec=1418, tv_usec=928797}) = 1 (in [12], left {tv_sec=1418, tv_usec=511679}) <0.417197>
```

*Chef's kiss*

NiLuJe:
I'm still not entirely convinced trusting the events won't lead to *more* weirdness, because I've seen some weird things on that front (like, severe timestamp jitter on an idle device (as in, > 500ms, meaning stuff coming in out of order), but, oh, well.

poire-z:
> e.g., a 500ms hold detected during a 1418s poll after 417ms ;).

There's some other 501ms lost somewhere ? :)
(I would expect to read "500ms hold detected during a 918s poll after 417ms" - but as I don't understand everything... :)

NiLuJe:
That's what trusting the input timestamps means (e.g., by the time select *starts*, the input timestamp is already stale by ~100ms).

The delta would be less than 10ms if it were using synthetic timestamps.

(That's on a H2O with its extremely wonky CPU scheduling, things are less jittery on saner devices).

NiLuJe:
Fun fact: this reduced the CPU cost of a single tap by a factor of ten. lol.

plugins/coverbrowser.koplugin/bookinfomanager.lua
```diff
@@ -691,7 +691,6 @@ function BookInfoManager:extractInBackground(files)
             local cover_specs = files[idx].cover_specs
             logger.dbg("  BG extracting:", filepath)
             self:extractBookInfo(filepath, cover_specs)
-            FFIUtil.usleep(100000) -- give main process 100ms of free cpu to do its processing
```
poire-z, line 694: You're sure ? This was mostly to let the main process time to sql query, fetch some data and cover, process the cover, force a repaint/refresh, and the whole thing feeling smooth. You think the kernel multitasking is enough, even on single cpu devices ?

poire-z: Seeing the base PR and kernel priority stuff. In my experience on normal linux servers (when I had 1/2 cores), I just don't trust nice/ionice to give enough smoothness (I use them because why not give us as much chances, but I often end up adding a sleep somewhere if I can when i need smoothness).

NiLuJe: It's definitely more than enough. The main process is not going to be doing anything until the next scheduled task, which is the item update scheduled every second.

That's extremely minimal contention to begin with, even if extractions hasn't finished yet.

NiLuJe: A NORMAL process at nice 0 can can run circles around a BATCH process at nice 5 ;).

NiLuJe: Yep, behaves perfectly fine even on a crappy H2O ;).

NiLuJe:
* ReaderDictionary: Port delay computations to TimeVal
* ReaderHighlight: Port delay computations to TimeVal
* ReaderView: Port delay computations to TimeVal
* Android: Reset gesture detection state on APP_CMD_TERM_WINDOW.
  This prevents potentially being stuck in bogus gesture states when switching apps.
* GestureDetector:
  * Port delay computations to TimeVal
  * Fixed delay computations to handle time warps (large and negative deltas).
  * Simplified timed callback handling to invalidate timers much earlier, preventing accumulating useless timers that no longer have any chance of ever detecting a gesture.
  * Fixed state clearing to handle the actual effective slots, instead of hard-coding slot 0 & slot 1.
  * Simplified timed callback handling in general, and added support for a timerfd backend for better performance and accuracy.
  * The improved timed callback handling allows us to detect and honor (as much as possible) the three possible clock sources usable by Linux evdev events.
    The only case where synthetic timestamps are used (and that only to handle timed callbacks) is limited to non-timerfd platforms where input events use
    a clock source that is *NOT* MONOTONIC.
    AFAICT, that's pretty much... PocketBook and that's it?
* Input:
  * Use the <linux/input.h> FFI module instead of re-declaring every constant
  * Fixed (verbose) debug logging of input events to actually translate said constants properly.
  * Completely reset gesture detection state on suspend. This should prevent bogus gesture detection on resume.
  * Refactored the waitEvent loop to make it easier to comprehend (hopefully) and much more efficient.
    Of specific note, it no longer does a crazy select spam every 100µs, instead computing and relying on sane timeouts,
    as afforded by switching the UI event/input loop to the MONOTONIC time scale.
* reMarkable: Stopped enforcing synthetic timestamps on input events, as it should no longer be necessary.
* TimeVal:
  * Refactored and simplified, especially as far as metamethods are concerned (based on <bsd/sys/time.h>).
  * Added a host of new methods to query the various POSIX clock sources, and made :now default to MONOTONIC.
  * Removed the debug guard in __sub, as time going backwards can be a perfectly normal occurence.
  * New methods:
    * Clock sources: :realtime, :monotonic, :monotonic_coarse, :realtime_coarse, :boottime
    * Utility: :tonumber, :tousecs, :tomsecs, :fromnumber, :isPositive, :isZero
* UIManager:
  * Ported event loop & scheduling to TimeVal, and switched to the MONOTONIC time base.
    This ensures reliable and consistent scheduling, as time is ensured never to go backwards.
  * Added a :getTime() method, that returns a cached TimeVal:now(), updated at the top of every UI frame.
    It's used throughout the codebase to cadge a syscall in circumstances where we are guaranteed that a syscall would return a mostly identical value,
    because very few time has passed.
    The only code left that does live syscalls does it because it's actually necessary for accuracy,
    and the only code left that does that in a REALTIME time base is code that *actually* deals with calendar time (e.g., Statistics).
* DictQuickLookup: Port delay computations to TimeVal
* FootNoteWidget: Port delay computations to TimeVal
* HTMLBoxWidget: Port delay computations to TimeVal
* Notification: Port delay computations to TimeVal
* TextBoxWidget: Port delay computations to TimeVal
* AutoSuspend: Port to TimeVal
* AutoTurn:
  * Fix it so that settings are actually honored.
  * Port to TimeVal
* BackgroundRunner: Port to TimeVal
* Calibre: Port benchmarking code to TimeVal
* BookInfoManager: Removed unnecessary yield in the metadata extraction subprocess now that subprocesses get scheduled properly.

* All in all, these changes reduced the CPU cost of a single tap by a factor of ten (!), and gets rid of an insane amount of weird poll/wakeup cycles that must have been hell on CPU schedulers and batteries..

NiLuJe (commented):
> Reviewed 1 of 10 files at r1, 1 of 6 files at r3, 1 of 6 files at r4, 2 of 14 files at r5, 1 of 3 files at r7, 1 of 1 files at r8, 6 of 12 files at r9, 1 of 5 files at r10, 2 of 5 files at r11, 11 of 13 files at r12, 4 of 4 files at r13.
> Dismissed @poire-z from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7415)* status: all files reviewed, 4 unresolved discussions (waiting on @ezdiy and @poire-z)

___
*[frontend/apps/reader/modules/readerhighlight.lua, line 1125 at r3](https://reviewable.io/reviews/koreader/koreader/7415#-MVsSkwlSUiTxSm140IC-r3-1125:-MWjm-QpCQAnX8MuTTgn:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/5b2a6d4969e5769ff878949e7ecdf3cb6b9e81bb/frontend/apps/reader/modules/readerhighlight.lua#L1125)):*

Yeah, had the same thought, I'm going to check that, and possibly rename now to `realtime` and alias `now` to `monotonic` if that's viable ;).
</blockquote></details>

Done.
___
*[frontend/apps/reader/modules/readerhighlight.lua, line 342 at r5](https://reviewable.io/reviews/koreader/koreader/7415#-MVsSkwlSUiTxSm140IC-r5-342:-MWjlxSxCfMiHg5K62lN:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/7a056877c97596d5719f6d21d4cc33ccfbae4c26/frontend/apps/reader/modules/readerhighlight.lua#L342)):*

We'd need some *really* unlucky GC behavior to get a collision ^^.

But the reasoning behind the change was flawed anyway (In the heat of the moment, I, err, forgot about the `__eq` and assumed it was essentially doing this instead of a proper comparison ^^), c.f., 0e931d4 ;).
</blockquote></details>

Done.
___
*[frontend/device/gesturedetector.lua, line 326 at r11](https://reviewable.io/reviews/koreader/koreader/7415#-MVsSkwlSUiTxSm140IE-r11-326:-MWjm9zY9kviGeT2C8cO:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/cd8c5c67d4fdfd9e5d7ac17d0d330c4b916d1a54/frontend/device/gesturedetector.lua#L326)):*

Yep, it's one of those confusing "haha, I'm actually a hash" ;).
</blockquote></details>

Done.
___
*[frontend/ui/timeval.lua, line 148 at r11](https://reviewable.io/reviews/koreader/koreader/7415#-MVsSkwlSUiTxSm140IH-r11-148:-MWjuunBC6HDiPKRIZJE:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/cd8c5c67d4fdfd9e5d7ac17d0d330c4b916d1a54/frontend/ui/timeval.lua#L148)):*

> (e.g., os.time(), util.gettime() or TimeVal:realtime()).

I wouldn't mind a comment somewhere stating that all these 3 return the same kind of time, unlike all the others TimeVal methods that are to be used only for internal time/delay/interval management.
</blockquote></details>

Done.

NiLuJe (commented):
> Reviewed 5 of 5 files at r14.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7415)* status: all files reviewed, 4 unresolved discussions (waiting on @ezdiy and @poire-z)

frontend/ui/uimanager.lua
```diff
@@ -1082,9 +1073,7 @@ function UIManager:sendEvent(event)
 
     -- Ensure discardEvents
     if self._discard_events_till then
-        local now = { ffiUtil.gettime() }
-        local now_us = now[1] * MILLION + now[2]
-        if now_us < self._discard_events_till then
+        if TimeVal:now() < self._discard_events_till then
```
poire-z, line 1076: Not the cached value UIManager:getTime() for this one ?
(I guess I'll probably often wonder and ask you the question :)

poire-z (commented):
Hard to get into the UIManager/input changes, but nothing jumps out. Trusting you.

frontend/device/input.lua
```diff
@@ -774,90 +875,185 @@ end
...
+                ok, ev, timerfd = input.waitForEvent(poll_timeout.sec, poll_timeout.usec)
+                -- We got an actual input event, go and process it
                 if ok then break end
-                local tv_now = TimeVal:now()
-                if (not timeout_us or tv_now < wait_deadline) then
-                    -- check whether timer is up
-                    if tv_now >= self.timer_callbacks[1].deadline then
+
+                -- If we've drained all pending input events, causing waitForEvent to time out, check our timers
+                if ok == false and ev == C.ETIME then
```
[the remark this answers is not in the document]
NiLuJe: It *can* be `nil`, in the C backend: after a read failure, on Android after a ALooper POLL failure; etc..

It's the "something terrible happened" signal ^^.

While false just means "we didn't get an event", and the error code then tells you why.

frontend/ui/uimanager.lua
NiLuJe: I wasn't *quite* sure it wa snecessary, but I don't have an easy way to test it, because it requires a slow dictionary or something? When I tried to double-check, I found that very few codepaths were using that, and none that I could easily trigger.

So it fell in the "when in doubt" basket, which uses getTime for the *beginning* of a time lapse, and :now for the *end* ;).

ptrm:
Wow, this seems to have fixes an issue with unwanted text selection being started when I used Android gestures on Boox. Back then when I dragged my finger upwards from under the bottom edge of the screen, often text got selected behaving as if the finger was still onscreen (tap extended the selection). Or a word would get looked up. Now this behavior is gone and koreader gestures work only when I want them :D

(I'm leaving this comment cause maybe there were some reports of it or anyone else was also bugged by this on androids)

((Also this seems to be fixing a nag I got on Kobos when timesync in the middle of a gesturs would crash the app due to touch end being in the past from touch start. I'm very grateful :D))

Frenzie:
Weird, that sounds more like a device-specific issue (very slow?). Still good that it was somehow solved. ;-)

ptrm:
Speed out of question I think, Boox Nova 3 is much better than predecessors, has octa 1.8G Snapdragon 636 and renders books much faster than my ClaraHD for example

NiLuJe:
Yep, that's https://github.com/koreader/koreader/pull/7415/commits/4ab0870b0bfe6d0c74174f093f951ddb5f134765

Note that on my phone (Android 10), I can cancel a system gesture in the middle of it (in this case, the "swipe up from bottom" to go the task view, which means the zoom out starts and then zooms back in), and that *doesn't* fire the event. (It's pretty rare, though, and requires some ninja skills to actually trigger in the first place, so much so that it might actually be a bug in the OnePlus ROMs).

poire-z:
Great spring obsolete kernels giveup ? :(
Just installed v2021.03-12-gde543b3_2021-03-31, and:
```
 [*] Version: v2021.03-12-gde543b3_2021-03-31

./luajit: frontend/ui/timeval.lua:31: ./luajit: undefined symbol: clock_getres
stack traceback:
    [C]: in function '__index'
    frontend/ui/timeval.lua:31: in main chunk
    [C]: in function 'require'
    frontend/device/gesturedetector.lua:46: in main chunk
    [C]: in function 'require'
    frontend/device/input.lua:8: in main chunk
    [C]: in function 'require'
    frontend/device/kobo/device.lua:342: in function 'init'
    frontend/device.lua:54: in main chunk
    [C]: in function 'require'
    ./reader.lua:148: in main chunk
    [C]: at 0x00014131
SDL2 not loaded:    ./setupkoenv.lua:30: Not able to load dynamic library: libSDL2-2.0.so.0
!!!!
Uh oh, something went awry... (Crash nÂ°1: 03/31/21 @ 18:44:24)
Running FW 3.19.5761 on Linux 3.0.35+ (#4071 PREEMPT Fri Jul 17 16:56:30 CST 2015)
Aborting . . .
```

My ffi/posix_h.lua does have:
`int clock_getres(clockid_t, struct timespec *) __attribute__((nothrow, leaf));`
so I guess I didn't mess up the install...

Frenzie:
No, I actually kept a side eye on that and @NiLuJe explicitly included this. :-/

```lua
-- clock_gettime & friends require librt on old glibc versions...
if ffi.os == "Linux" then
   pcall(ffi.load, "rt")
end
```

poire-z:
Going on with `if false and ffi.os == "Linux" then ... clock_getres stuff...`.
Now:
```
03/31/21-18:53:50 INFO  initializing for device Kobo_alyssum
03/31/21-18:53:50 INFO  framebuffer resolution: {
    ["h"] = 1448,
    ["w"] = 1072,
}
[ko-input] Forked off fake event generator (pid: 1179)
./luajit: frontend/ui/timeval.lua:181: ./luajit: undefined symbol: clock_gettime
stack traceback:
    [C]: in function '__index'
    frontend/ui/timeval.lua:181: in function 'now'
    frontend/ui/uimanager.lua:38: in main chunk
    [C]: in function 'require'
    ./reader.lua:212: in main chunk
    [C]: at 0x00014131
!!!!
```
What can I have messed ?!

poire-z:
Creating `librt.so -> librt-2.11.1.so` does not help.

NiLuJe:
You you don't protect the `ffi.load` and capture the output, does it succeed?

NiLuJe:
And/or try moving it to the top of reader.lua in case there's a weird twisty require dependency chain...

NiLuJe:
That, or it's not loaded in the GLOBAL namespace on older libcs for some reason...

NiLuJe:
```diff
diff --git a/ffi/posix_h.lua b/ffi/posix_h.lua
index 2ac14f50..69c950cd 100644
--- a/ffi/posix_h.lua
+++ b/ffi/posix_h.lua
@@ -1,10 +1,5 @@
 local ffi = require("ffi")
 
--- clock_gettime & friends require librt on old glibc versions...
-if ffi.os == "Linux" then
-   pcall(ffi.load, "rt")
-end
-
 -- Handle arch-dependent typedefs...
 if ffi.arch == "x64" then
     require("ffi/posix_types_x64_h")
@@ -160,3 +155,21 @@ int sched_getaffinity(int, size_t, cpu_set_t *) __attribute__((nothrow, leaf));
 int sched_setaffinity(int, size_t, const cpu_set_t *) __attribute__((nothrow, leaf));
 int sched_yield(void) __attribute__((nothrow, leaf));
 ]]
+
+-- clock_gettime & friends require librt on old glibc (< 2.17) versions...
+if ffi.os == "Linux" then
+    local C = ffi.C
+    -- If it's already in the libc namespace, we're good
+    if C.clock_gettime == nil then
+        -- Otherwise, try to look inside librt...
+        local ok, rt = pcall(ffi.load, "rt")
+        if ok then
+            -- It's in librt, copy references to the C namespace to make the caller's lives easier.
+            C.clock_getres = rt.clock_getres
+            C.clock_gettime = rt.clock_gettime
+            C.clock_settime = rt.clock_settime
+        else
+            error("System doesn't support clock_gettime & co!")
+        end
+    end
+end
```

NiLuJe:
I actually thought the swap happened much later, and as such that I actually fell into the "need librt" basket, but, no, we're using glibc 2.19, so, no need for librt, which is why I never caught this :/.

NiLuJe:
Eh, probably can't do what I tried above, this is going to be fun...

NiLuJe:
Slightly better chance of working:

```diff
diff --git a/ffi/posix_h.lua b/ffi/posix_h.lua
index 2ac14f50..c515254b 100644
--- a/ffi/posix_h.lua
+++ b/ffi/posix_h.lua
@@ -1,10 +1,5 @@
 local ffi = require("ffi")
 
--- clock_gettime & friends require librt on old glibc versions...
-if ffi.os == "Linux" then
-   pcall(ffi.load, "rt")
-end
-
 -- Handle arch-dependent typedefs...
 if ffi.arch == "x64" then
     require("ffi/posix_types_x64_h")
@@ -160,3 +155,13 @@ int sched_getaffinity(int, size_t, cpu_set_t *) __attribute__((nothrow, leaf));
 int sched_setaffinity(int, size_t, const cpu_set_t *) __attribute__((nothrow, leaf));
 int sched_yield(void) __attribute__((nothrow, leaf));
 ]]
+
+-- clock_gettime & friends require librt on old glibc (< 2.17) versions...
+if ffi.os == "Linux" then
+    local C = ffi.C
+    -- Look inside librt, and load it in the global namespace...
+    local ok, rt = pcall(ffi.load, "/lib/librt.so.1", true)
+    if not ok then
+        error("Failed to load librt for clock_gettime & co!")
+    end
+end
```

poire-z:
> If you don't protect the ffi.load and capture the output, does it succeed?

When I don't have `librt.so -> librt-2.11.1.so`:
`./luajit: ./ffi/posix_h.lua:4: librt.so: cannot open shared object file: No such file or directory`
Once I have it, this passes, but then as above: `./luajit: undefined symbol: clock_gettime`

> And/or try moving it to the top of reader.lua in case there's a weird twisty require dependency chain...

```diff
 #!./luajit
+local ffi = require("ffi")
+ffi.load("rt")
```

Same `./luajit: undefined symbol: clock_gettime`.

> https://github.com/koreader/koreader/pull/7415#issuecomment-811307670 (first version of previous post, I guess no need to test the second :)

Works ! seeing your new rotation icons ! :)

NiLuJe:
Cool, thanks :).

That would explain the Cervantes failures reported last night & today, as those are dunning glibc 2.12.2... (*sigh*).

PR #7416, opened 2021-03-16
Use PluginShare to share the keepalive state

Hzj-jie:
Currently the "checked" mark does not work.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7416)

### DIFF
--- a/plugins/keepalive.koplugin/main.lua
+++ b/plugins/keepalive.koplugin/main.lua
@@ -19,7 +19,7 @@ local function showConfirmBox()
         cancel_text = _("Close"),
         cancel_callback = function()
             disable()
-            PluginShare.keepalive =false
+            PluginShare.keepalive = false
         end,
         ok_text = _("Stay alive"),
         ok_callback = function()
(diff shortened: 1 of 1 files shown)


Hzj-jie:
@Frenzie wanna take a quick look? Thank you.

Hzj-jie:
By the way, maybe we can change the name of this plugin into "demo mode".

plugins/keepalive.koplugin/main.lua
```diff
@@ -18,17 +19,16 @@ local function showConfirmBox()
         cancel_text = _("Close"),
         cancel_callback = function()
             disable()
-            menuItem.checked =false
+            PluginShare.keepalive =false
```
Frenzie, line 22: nit

Frenzie (approved):
I think that should be fine.

plugins/keepalive.koplugin/main.lua
Hzj-jie: oops

PR #3247, opened 2017-09-22
Some History fixes and enhancements

poire-z:
Seen https://www.mobileread.com/forums/showthread.php?t=290611 , and I'm not sure we need a "Purge all books in history" feature. But we should at least be able to purge or delete from History (currently, when I want to do that, i have to: open the book, switch to File browser to get in that book directory, browse the directory to find the book, delete it, go back to History, repeat for the next book).

So, I just wanted to add "Purge .sdr" and "Delete" to the onHold buttons of History... and it lead me to having to do much more:

- Made the onHold buttons table similar to the one of File browser (use ButtonDialogTitle with the full filename as title, instead of a truncated one), just cause it's prettier, and for consistency.
- Added "Purge .sdr" and "Delete" to these buttons.
- Moved the purgeSettings and removeFileFromHistoryIfWanted logic into filemanagerutil functions, now that they are used in 2 places.
- Stay on the same page when manipulating history items (previously, we were always put back on first page).
- Really keep deleted files in history (unless setting says otherwise). Previously, these were always removed on startup, because of kept logic that was duplicated in clearMissing(), so it was applied even if `autoremove_deleted_items_from_history` was false (*).
- Show deleted items in grey or dimmed in classic History and all CoverBrowser display modes.

(There were many combinations to check, and I may have missed a few, so this may introduce some new bugs.)

(*): we disccused this some months ago, about whether History should keep deleted files, which was chosen as to be the default, with a setting `autoremove_deleted_items_from_history` to change this behaviour.
But since then, and even with the apparent clearMissing() fix, they were still removed.
I thought about making that `autoremove_deleted_items_from_history` true as default (because that's what people are accustomed to), but then, they wouldn't see the grey/dim stuff I just added :)

PR #2583, opened 2017-02-25
FileManagerHistory: remove harmful deletion code and turn it into a feature

Frenzie:
Fixes the problem introduced in 51327911f2fbea00258717d7b284e061ad31832e and turns it into a feature that should fix #2477

## New button
screenshot_2017-02-25_19-09-42

PR #7419, opened 2021-03-17
Add a plugins/insert_menu.lua to help "non-official" plugins to place themselves into "More Tools" menu

Hzj-jie:
(The original PR has been moved to contrib: https://github.com/koreader/contrib/tree/Hzj-jie-findhistory/findhistory.koplugin.)

Currently if a menu items is not defined, it will be placed into the first menu folder which is not ideal for contrib/plugins. So this is a simple piece of logic to place them into "More Tools" and expected to be maintained by KOReader.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7419)


plugins/findhistory.koplugin/main.lua
```diff
@@ -0,0 +1,114 @@
...
+    local noSuffix, suffix = file:match("(.*)%.sdr/metadata%.(.+)%.lua")
+    return noSuffix .. "." .. suffix
+end
+
+local function doBuildHistory()
+    --- home_dir should be returned by some shared logic.
+    local dir = G_reader_settings:readSetting("home_dir")
+                or require("apps/filemanager/filemanagerutil").getDefaultDir()
+                or "."
+    local file = io.popen("find '" .. dir .. "' -name 'metadata.*.lua' -exec stat -c '%N %Y' {} \\;")
```
NiLuJe, line 31: Ouch.

LFS directory walking instead, maybe?

(c.f. PathChooser).

Hzj-jie: Any reason to prefer lfs? It uses stat under the hood indeed.

Frenzie: At the very least lfs is more portable. E.g., will it work on super ancient BusyBox/Android, it will definitely not work on Windows, etc.

Hzj-jie:
I barely remember the issue I have opened 4 years ago. LOL

I may want to explain some background about the ReadHistory module change first.
Before ReadHistory, the sidecar was implemented to replaced the /history/mnt#us#documents#file.lua format. So during the transition, a module is a must-have to handle both legacy history folder and the new sidecar, as well as maintaining the history list.
Whether the history view itself needs to be derived directly from sidecar files was not part of the discussion at that time. So to avoid heavily searching for all the sidecar files, I implemented a history.lua file to store all the history entries.
It indeed introduces an interesting duplication problem, history.lua can be rebuilt from sidecar files. And usually, duplications cause conflict. Though I agree, here the duplication can be considered as a feature, the history view can be considered as a "favorite" list, cause users can remove items from it - you cannot "erase" the fact one book has been read before.
But anyway, this is more or less a discussion about the definition of the history feature itself. We can delay it.

I use two devices and regularly synchronize files between them, I'd like to have a way to merge the history views as well. I can definitely copy sidecar files between them, but merging history.lua files requires manual operations.
Another use case I can imagine is to copy books from one device to another, if they have different mount points, history.lua cannot be directly copied from one to another.

plugins/findhistory.koplugin/main.lua
Hzj-jie: Ooops, I thought windows is not expected to be supported. Will update this part anyway.

Frenzie: It's not "expected" but there's no reason to make it harder. ;-)

poire-z:
> It indeed introduces an interesting duplication problem, history.lua can be rebuilt from sidecar files. And usually, duplications cause conflict. Though I agree, here the duplication can be considered as a feature, the history view can be considered as a "favorite" list, cause users can remove items from it - you cannot "erase" the fact one book has been read before.
But anyway, this is more or less a discussion about the definition of the history feature itself. We can delay it.

Nothing to delay or even consider, it's not a problem !
You made ReadHistory what it is, something orthogonal to the presence or absence of DocSettings, and even to the absence or presence of the file. It's super versatile, it allows many use cases, there's no reason to restrict it to one's view of "what history should be".
Related discussions: #2477 #2583 #3247 

Around that, we added settings to tweak the behaviour (Clean history of removed files, Auto-remove deleted or purged items from History) so it helps it working better for some of these use cases.

> I use two devices and regularly synchronize files between them, I'd like to have a way to merge the history views as well. I can definitely copy sidecar files between them, but merging history.lua files requires manual operations.
Another use case I can imagine is to copy books from one device to another, if they have different mount points, history.lua cannot be directly copied from one to another.

OK, I understand. It feels like a quite rare use case though.
I think the most logical way, if we don't want that a plugin, would be to have it available from FileBrowser, may be it's <kbd>+</kbd> menu: you're on a directory (your home folder or somewhere else), and there, you'd have <kbd>Update history with opened books in this directory and under</kbd>: it would **not** clean other files from the History (that's orthogonal, leave that to "Clean history from removed books", keeping that independant allows more use cases). Or you could have it dialog driven like "Refresh cached book information" is.

Or, keep this a user-plugin, or a user shell script (that you can invoke from FileBrowser), as you seem to be fond of using shell commands ("mv ..." , Lua has "os.rename").

Hzj-jie:
Adding a menu item in the file browser menu looks like a quite good solution.

Hzj-jie:
Emmm, I also like the idea of "user-plugin". I'd like this piece of code to be merged only to avoid being removed after an update. But if koreader accepts user defined plugins, we do not need to merge rare-used functions into the main branch anymore.

poire-z:
> But if koreader accepts user defined plugins

An update shouldn't remove files that are not part of the package (dunno if that is true on Android), so that's a solution. There are [a few user-plugins](https://github.com/koreader/koreader/issues?q=label%3A%22user+plugin%22) that people interested can pick/hack.

Hzj-jie:
The menu orders will be removed, so the menu entry will always be placed under the first button. It's not a show-stopper, but if loading user-plugins is officially supported, the issue is better to be addressed. Maybe add a new menu item called "User defined tools" below "More tools"?

Also using issue tags to search for user plugins seems too difficult. I like the idea @NiLuJe mentioned in #6768. It can be a new repo with folders like plugins/, scripts/, etc. Each plugin or script can have a maintainer. They will not be packed in the main package, but users can decide to install and use them at their own risk.

And I still believe providing an extra user-define plugin folder to avoid messing up with the installation on the device would be better, as @jperon mentioned in #6849.

Frenzie:
Note that I wrote "Just adding a default /sdcard/koreader/plugins *type* folder there seems like it should pretty much do the trick." ;-) Meaning I didn't say it should be called plugins by default.

Hzj-jie:
Yes, quite true, and something may be wrong in the current implementation unless I misunderstood it.

`
    else
        local data_dir = require("datastorage"):getDataDir()
        if data_dir ~= "." then
            G_reader_settings:saveSetting("extra_plugin_paths", { data_dir .. "/plugins/" })
        end
 
@Hzj-jie Hzj-jie 21 hours ago Member
Did you forget to add the new extra path to lookup_path_list?
`

Frenzie:
Not forgotten since the main goal there was platforms like Android and AppImage. I don't think anybody's ever minded putting custom plugins in /plugins on Kobo before, or at least I haven't.

Frenzie:
PS If you want to change it the line should be replaced, not added to.

poire-z:
So, user plugin it is ? Can we close this PR and tag it [user plugin](https://github.com/koreader/koreader/issues?q=label%3A%22user+plugin%22) ?

Hzj-jie:
Sorry I lost tracking this PR. I will create a contrib project and merge this one there if you do not have concerns. @NiLuJe #6768

NiLuJe:
Fine by me :).

poire-z:
`plugins/insert_menu.lua` probably shouldn't be the single file in `plugins/` - let this directory only hve sub-directories, one per real plugin.
Looks like you had already make yourself a little place for these kind of plugin helpers some years ago :)
https://github.com/koreader/koreader/tree/master/frontend/ui/plugin

Hzj-jie:
Are these tests known to fail:
[  FAILED  ] spec/front/unit/filemanager_spec.lua @ 24: FileManager module should show error on non-existent file
[  FAILED  ] spec/front/unit/filemanager_spec.lua @ 38: FileManager module should not delete settings for non-document file
[  FAILED  ] spec/front/unit/filemanager_spec.lua @ 73: FileManager module should delete document with its settings
?

NiLuJe:
Yeah, I fixed that a few days ago, you'll want to rebase ;).

mergen3107:
@Hzj-jie 
Thank you so much! Your plugin works [charms](https://github.com/koreader/koreader/issues/7435#issuecomment-878686390)!

PR #7425, opened 2021-03-19
Unbreak HTTP authentication

johnbeard:
It appears that passing `username` and `password` to `http.request()` doesn't work.

http://w3.impa.br/~diego/software/luasocket/http.html#request indicates that the way to set this is with a URL like `http://user:pass@url.com/foo`. This does actually work if you set this URL in the WebDAV settings.

This patch uses the (old) alternative method of direct authentication, so as not to have to mess about with splitting the protocol, adding the `user:password@` and recombining.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7425)

### DIFF
--- a/frontend/apps/cloudstorage/webdavapi.lua
+++ b/frontend/apps/cloudstorage/webdavapi.lua
@@ -6,6 +6,7 @@ local socket = require("socket")
 local socketutil = require("socketutil")
 local util = require("util")
 local _ = require("gettext")
+local logger = require("logger")
 
 local WebDavApi = {
 }
@@ -94,6 +95,11 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
     local headers_request = socket.skip(1, http.request(request))
     socketutil:reset_timeout()
     if headers_request == nil then
+        return nil
+    elseif headers_request < 200 or headers_request >= 300 then
+        -- got a response, but it wasn't a success (e.g. auth failure)
+        logger.dbg(headers_request)
+        logger.dbg(table.concat(sink))
         return nil
     end
 
--- a/frontend/apps/cloudstorage/webdavapi.lua
+++ b/frontend/apps/cloudstorage/webdavapi.lua
@@ -87,7 +87,7 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
             ["Depth"]          = "1",
             ["Content-Length"] = #data,
         },
-        username = user,
+        user     = user,
         password = pass,
         source   = ltn12.source.string(data),
         sink     = ltn12.sink.table(sink),
@@ -162,7 +162,7 @@ function WebDavApi:downloadFile(file_url, user, pass, local_path)
         url      = file_url,
         method   = "GET",
         sink     = ltn12.sink.file(io.open(local_path, "w")),
-        username = user,
+        user     = user,
         password = pass,
     })
     socketutil:reset_timeout()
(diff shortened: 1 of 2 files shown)


NiLuJe:
LuaSockets should already be doing that given the user & password argument not being nil :?

NiLuJe:
e.g., https://github.com/diegonehab/luasocket/blob/5b18e475f38fcf28429b1cc4b17baee3b9793a62/src/http.lua#L243-L248

NiLuJe:
Oh. Obvious typo is obvious ;D.

NiLuJe:
The field should be named `user`, not `username`. :facepalm:

NiLuJe:
```lua
#!./luajit

dofile("setupkoenv.lua")

http = require("socket.http")
ltn12 = require("ltn12")
socket = require("socket")
socketutil = require("socketutil")

socketutil:set_timeout(5, -1)
remote_file = "http://pie.dev/headers"
local_file = "/mnt/niluje/foo"

sink = {}
print(os.date("%T"))

code, headers, status = socket.skip(1, http.request{
    url    = remote_file,
    sink  = socketutil.table_sink(sink),
    user   = "foo",
    password = "bar",
    
})
print(os.date("%T"))

print(code)
for k, v in ipairs(headers or {}) do
   print(k, v)
end
print(status)

print(table.concat(sink))
```

```
15:36:42
15:36:42
200
HTTP/1.1 200 OK
{
  "headers": {
    "Accept-Encoding": "gzip", 
    "Authorization": "Basic Zm9vOmJhcg==", 
    "Cdn-Loop": "cloudflare", 
    "Cf-Connecting-Ip": "86.70.214.143", 
    "Cf-Ipcountry": "FR", 
    "Cf-Ray": "63276f40be20cdc3-FRA", 
    "Cf-Request-Id": "08ec83dc7c0000cdc3ab0e7000000001", 
    "Cf-Visitor": "{\"scheme\":\"http\"}", 
    "Connection": "Keep-Alive", 
    "Host": "pie.dev", 
    "User-Agent": "KOReader/2021.02-65 (https://koreader.rocks/) LuaSocket/3.0-rc1"
  }
}
```

Frenzie:
This almost certainly can't be right. Are you sure there's not simply a typo somewhere?

NiLuJe:
@Frenzie: c.f., above ;).

> The field should be named `user`, not `username`. facepalm

Frenzie:
Weird, for some reason the other e-mails weren't linked to the OP and while I'd glanced at the diff I hadn't clicked on view issue.

But yeah, please revert those commits and make it a one (?) line typo fix. :+1:

johnbeard:
I found one more in the OPDS stuff, but I don't have an auth'd OPDS server to test it on.

frontend/apps/cloudstorage/webdavapi.lua
```diff
@@ -95,6 +96,11 @@ function WebDavApi:listFolder(address, user, pass, folder_path)
     socketutil:reset_timeout()
     if headers_request == nil then
         return nil
+    elseif headers_request < 200 or headers_request >= 300 then
+        -- got a response, but it wasn't a success (e.g. auth failure)
+        logger.dbg(headers_request)
+        logger.dbg(table.concat(sink))
+        return nil
```
Frenzie, line 103: Unless we care about the distinction. Btw, logger supports tables.
```suggestion
        return
```

NiLuJe:
Hey, I managed to get *one* out of the two in opdsbrowser right :D. (probably bu sheer luck, but, shhh. :D).

frontend/apps/cloudstorage/webdavapi.lua
NiLuJe: The branch above did the same, FWIW.

NiLuJe (approved):
Can't find any other occurrence of this SNAFU, thanks! (and my bad :x).

PR #7431, opened 2021-03-19
Kobo: Allow toggling the WAIT_FOR_UPDATE_COMPLETE hack

NiLuJe:
And enable it by default on the Libra (which already happened in base previously, since https://github.com/koreader/koreader-base/pull/1329).

This is hidden away inside the dev menu, because duh'.

Requires https://github.com/koreader/koreader-base/pull/1334

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7431)

### DIFF
--- a/frontend/apps/filemanager/filemanagermenu.lua
+++ b/frontend/apps/filemanager/filemanagermenu.lua
@@ -542,11 +542,8 @@ To:
                 else
                     mxcfb_bypass_wait_for = not Device:hasReliableMxcWaitFor()
                 end
-                G_reader_settings:saveSetting("mxcfb_bypass_wait_for", not mxcfb_bypass_wait_for)
-                local InfoMessage = require("ui/widget/infomessage")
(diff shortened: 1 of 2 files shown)


NiLuJe:
Eh, I can probably tweak it live without requiring a reboot, duh'. (but after dinner ;p).

poire-z (approved):

frontend/apps/filemanager/filemanagermenu.lua
```diff
@@ -522,6 +522,34 @@ To:
             end,
         })
     end
+    --- @note: Currently, only Kobo implements this quirk
+    if Device:hasEinkScreen() and Device:isKobo() then
+        table.insert(self.menu_items.developer_options.sub_item_table, {
+            text = _("Bypass the MXCFB_WAIT_FOR_* ioctls"),
```
Frenzie, line 528: A translator's comment might be useful (to the effect of only translate bypass, probably).

PR #7432, opened 2021-03-20
DictQuickLooup: Fix a few bad interactions with ReaderHighlight's dialog

### LINKED ISSUE #7430: Readerhighlight crashes when nothing selected
#### Steps to reproduce
Hold a word - Highlight menu appears
press Dictionary - in Dictionary window press Search
Search navigation menu appears (over Highlight menu! - bad things start here)
press the screen outside of the both menus - Search navgation menu disappears
Highlight menu remains!
Word hghlighting disappears!
Press Highlight (or Add Note) - got a crash

readerhighlight.lua:1307: attempt to index field 'hold_pos' (a nil value)

NiLuJe:
* Disable the Search button when docless and highlight-less.
* Make sure launching a Search closes ReaderHighlight's dialog in all cases. Fix #7430

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7432)

### DIFF
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -372,7 +372,7 @@ function DictQuickLookup:init()
                                                     UIManager:scheduleIn(1.0, function()
                                                         if self.ui then
                                                             -- close Highlight menu if any still shown
-                                                            if self.ui.highlight then
+                                                            if self.ui.highlight and self.ui.highlight.highlight_dialog then
                                                                 self.ui.highlight:onClose()
                                                             end
                                                             self.ui:onClose()
@@ -882,7 +882,7 @@ function DictQuickLookup:onCloseWidget()
 
     -- If ReaderHighlight's dialog is currently, shown, close it, because we may have just cleared highlights,
     -- and it doesn't have the opportunity to refresh its state.
-    if self.ui and self.ui.highlight then
+    if self.ui and self.ui.highlight and self.ui.highlight.highlight_dialog then
         self.ui.highlight:onClose()
     end
 
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -370,16 +370,14 @@ function DictQuickLookup:init()
                                                     self:onHoldClose(true)
                                                     -- close current ReaderUI in 1 sec, and create a new one
                                                     UIManager:scheduleIn(1.0, function()
-                                                        local ReaderUI = require("apps/reader/readerui")
-                                                        local reader = ReaderUI:_getRunningInstance()
-                                                        if reader then
+                                                        if self.ui then
                                                             -- close Highlight menu if any still shown
-                                                            if reader.highlight then
-                                                                reader.highlight:onClose()
+                                                            if self.ui.highlight then
+                                                                self.ui.highlight:onClose()
                                                             end
-                                                            reader:onClose()
+                                                            self.ui:onClose()
                                                         end
-                                                        ReaderUI:showReader(epub_path)
+                                                        self.ui:showReader(epub_path)
                                                     end)
                                                 end,
                                             })
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -54,7 +54,6 @@ local DictQuickLookup = InputContainer:new{
     -- refresh_callback will be called before we trigger full refresh in onSwipe
     refresh_callback = nil,
     html_dictionary_link_tapped_callback = nil,
-    no_clear = nil,
 }
 
 local highlight_strings = {
@@ -881,6 +880,12 @@ function DictQuickLookup:onCloseWidget()
         end
     end
 
+    -- If ReaderHighlight's dialog is currently shown, close it, because we may have just cleared highlights,
+    -- and it doesn't have the opportunity to refresh its state, so its Highlight & Note buttons are still enabled.
+    if self.ui and self.ui.highlight and self.ui.highlight.highlight_dialog then
+        self.ui.highlight:onClose()
+    end
+
     -- NOTE: Drop region to make it a full-screen flash
     UIManager:setDirty(nil, function()
         return "flashui", nil
@@ -1083,7 +1088,7 @@ function DictQuickLookup:onClose()
             table.remove(self.window_list, i)
         end
     end
-    if self.highlight and not self.no_clear then
+    if self.highlight then
         -- delay unhighlight of selection, so we can see where we stopped when
         -- back from our journey into dictionary or wikipedia
         local clear_id = self.highlight:getClearId()
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -67,6 +67,7 @@ function DictQuickLookup:canSearch()
     end
 
     if self.is_wiki then
+        -- In the Wiki variant of this widget, the Search button is coopted to cycle between enabled languages.
         if #self.wiki_languages > 1 then
             return true
         end
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -1111,6 +1111,14 @@ function DictQuickLookup:onHoldClose(no_clear)
     return true
 end
 
+function DictQuickLookup:onCloseWidget()
+    -- If ReaderHighlight's dialog is currently, shown, close it, because we may have just cleared highlights,
+    -- and it doesn't have the opportunity to refresh its state.
+    if self.ui and self.ui.highlight then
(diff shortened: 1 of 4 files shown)


NiLuJe:
Mmh, wait, that broke docless closes, somehow.

poire-z:
I guess it might be complicated solving the issue an other way - but from a UX perspective, I like the highlight dialog to stay and getting back to it to launch other stuff on the still selected thingie (Dict, ok, nothing > Wiki, bof bof > Translate, so-so, ok, highlight it, I'll look at it later).

I'd rather solve the edge case by just not crashing, no matter the mess displayed, but not lose the above functionnality ;)

NiLuJe:
I got that given the behvaior, but doing it right would involve being able to rebuild the ReaderHighlightMenu.

I don't personally care enough about it (read: at all ^^) to go that far ;).

Just not crashing but having the highlights gone and the buttons not doing anything feels slightly weird.

NiLuJe:
Or, possibly easier: telling DictQuickLookup not to clear the highlights when launched from RH?

poire-z:
But currently there is zero issue when used in the simple 99% workflow:
long-press > Highlight menu > Dictionary > DictWindow > Close it (via button or tap outside) : the highlight is not cleared, and still usable for other stuff, even the same Dictionary lookup.

It's just when doing some other stuff from the DictQuickLookup window, like hitting Search (which may close the Dict window a different way than the regular quit ?) that things get strange as in #7430 - Can probably be solved - but what happens at this point, and if it is strange, I don't really mind as long as it does not crash.

NiLuJe:
> But currently there is zero issue when used in the simple 99% workflow:
> long-press > Highlight menu > Dictionary > DictWindow > Close it (via button or tap outside) : the highlight is not cleared, and still usable for other stuff, even the same Dictionary lookup.

Nope, DictQuickLookup was always clearing highlights on close.

b70b187 is hilariously not pretty, but does the trick ;).

poire-z:
> DictQuickLookup was always clearing highlights on close.

It is not, for me, in the workflow described above, with current master on my Kobo or the emulator.

I guess I'd be ok with the no_clear idea (which, I guess, don't change the behaviour I described :)
No solution by just having the dict/wiki thingie check if there's a highlight dialog shown ? like just checking self.ui.highlight.highlight_dialog ?

NiLuJe:
> It is not, for me, in the workflow described above, with current master on my Kobo or the emulator.
> 

Huh, yeah. No idea how that works, though, because DictQuickLookup's onClose is pretty much always calling clear :?

> I guess I'd be ok with the no_clear idea (which, I guess, don't change the behaviour I described :)
> No solution by just having the dict/wiki thingie check if there's a highlight dialog shown ? like just checking self.ui.highlight.highlight_dialog ?

Err, yes, yes, that would probably work and be way less ugly ;p.

NiLuJe:
And that didn't actually fix #7340 anyway, because the Search dialog is also clearing highlights :D.

NiLuJe:
> > It is not, for me, in the workflow described above, with current master on my Kobo or the emulator.
> 
> Huh, yeah. No idea how that works, though, because DictQuickLookup's onClose is pretty much always calling clear :?
> 

Ah, `self.highlight` is nil in the RH case.

NiLuJe:
Which means the culprit was `Search` all along, which makes the fix much simpler ;p.

NiLuJe:
> Ah, `self.highlight` is `nil` in DQL the RH case.

Is this by design, BTW? (RH is not setting the highlight arg in the Lookup Events).

(Another side-effect is that it disables the Highlight button in DictQuickLookup when launched via ReaderHighlight).

poire-z (approved):
Looks better and fine to me :)

poire-z:
> Ah, self.highlight is nil in DQL when launched via RH.
> Is this by design, BTW? (RH is not setting the highlight arg in the Lookup Events).

Dunno, but I'd say it is by design, as a way to not get the current behaviour of not clearing highlight for sub-actions, so we still have it available for other actions.
(And I guess the highlight menu is most often used with multiple words selection, and Dictionary has little use on these - it's there just in case, but not the main protagonist :)

hius07:
Do you have Search from DictQuickLookup working (before this PR)?
I invoke Dctionary lookup from the reader menu, then press Search in the dictionary article window and got nothing, the window just closes.
I am sure of several entries of the search string in the text.

poire-z:
> Do you have Search from DictQuickLookup working (before this PR)?
I invoke Dctionary lookup from the reader menu, then press Search in the dictionary article window and got nothing, the window just closes.

I'm pretty sure this use case has never worked - and that nobody never really thought about it :)

Frenzie:
It should strictly speaking be disabled, but it seems quite unproblematic. Not sure what you think working would even look like

Frenzie (approved):

NiLuJe:
Disabling the Search button in that instance was *mostly* trivial.

(Had to punt off the logic to a function because fake ternary operators didn't cut it).

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -61,6 +61,25 @@ local highlight_strings = {
     unhighlight = _("Unhighlight"),
 }
 
+function DictQuickLookup:canSearch()
+    if self:isDocless() then
+        return false
+    end
+
+    if self.is_wiki then
+        if #self.wiki_languages > 1 then
```
poire-z, line 70: A small comment mentionning that the Search button is used to show Wikipedia languages would be nice (because I wondered what Wikipedia has to do with that :)

frontend/ui/widget/dictquicklookup.lua
```diff
@@ -61,6 +61,26 @@ local highlight_strings = {
...
+    if self.is_wiki then
+        -- In the Wiki variant of this widget, the Search button is coopted to cycle between enabled languages.
+        if #self.wiki_languages > 1 then
+            return true
+        end
+    else
+        -- This is to prevent an ineffective button when we're launched from the Reader's menu.
+        if self.ui.highlight.selected_text then
+            return true
+        end
```
poire-z, line 78: (Not an issue with out current code because of some side effects, but...)
This would probably be cleaner:
```diff
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -395,4 +395,5 @@ function ReaderHighlight:clear(clear_id)
     if self.hold_pos then
         self.hold_pos = nil
+        self.selected_word = nil
         self.selected_text = nil
         UIManager:setDirty(self.dialog, "ui")
@@ -1195,5 +1196,4 @@ function ReaderHighlight:onHoldRelease()
     elseif self.selected_word then
         self:lookup(self.selected_word, self.selected_link)
-        self.selected_word = nil
     end
     return true
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -76,5 +76,5 @@ function DictQuickLookup:canSearch()
     else
         -- This is to prevent an ineffective button when we're launched from the Reader's menu.
-        if self.ui.highlight.selected_text then
+        if self.ui.highlight.selected_word or self.ui.highlight.selected_text then
             return true
         end
```
Normally, `selected_text` is only set on multiple word selection - it is empty on single word long-press, where `selected_word` is set and used instead.

It happens that when DictQuickLookup is shown that the test for the unrelated <kbd>Highlight</kbd> button is updating selected_text:
```
frontend/document/credocument.lua:616: in function 'getTextFromPositions'
frontend/apps/reader/modules/readerhighlight.lua:1243: in function 'highlightFromHoldPos'
frontend/apps/reader/modules/readerhighlight.lua:1317: in function 'getHighlightedItem'
frontend/ui/widget/dictquicklookup.lua:926: in function 'getHighlightText'
frontend/ui/widget/dictquicklookup.lua:449: in function 'init'
```
So, this side effect makes it work.

It happens that I remove/replace this <kbd>Highlight</kbd> button in my own patch, so I got the <kbd>Search</kbd> button always disabled :) (Had a hard time finding this was the issue.)

poire-z: Or instead, just:
```diff
         -- This is to prevent an ineffective button when we're launched from the Reader's menu.
-        if self.ui.highlight.selected_text then
+        if self.highlight then
```
if we don't want to take much risks, the same check used for the <kbd>Highlight</kbd> button:
https://github.com/koreader/koreader/blob/9f835156d16da2f958944a7f2326ee1cc9b47af3/frontend/ui/widget/dictquicklookup.lua#L446-L449

NiLuJe: Yeah, I remember that some of these were only set in certain circumstances, but I don't remember the details, and I think I only went so far as thinking "depends on what did the selection" or something.

I'm not *quite* sure `self.highlight` was set in the specific set of circumstances this tried to fix though, but if it works, go ahead ;).

PR #7434, opened 2021-03-20
TOC settings: fix possible crash

poire-z:
Fix for https://github.com/koreader/koreader/issues/7428#issuecomment-802959731 , although I can't reproduce it, and this part being run and this crash happening should have been prevented by the enabled_func not allowing it to be run...

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7434)


Frenzie (approved):
Very strange.

PR #7436, opened 2021-03-21
Fulltext search: show notification when nothing found

hius07:
When nothing found now we have the `search_dialog` (arrows) displayed and need to observe the screen whether there are any search results highlighted.
I propose to show the infomessage for 2 sec.
After that we can continue search in the opposite direction with the `search_dialog` if needed.

`vsync` removed to avoid the `search_dialog` buttons showing up over the infomessage and it seems to cause no problems anywhere.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7436)

### DIFF
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -4,10 +4,9 @@ local InputContainer = require("ui/widget/container/inputcontainer")
 local InputDialog = require("ui/widget/inputdialog")
 local Notification = require("ui/widget/notification")
 local UIManager = require("ui/uimanager")
-local T = require("ffi/util").template
 local logger = require("logger")
 local _ = require("gettext")
-
+local T = require("ffi/util").template
 local ReaderSearch = InputContainer:new{
     direction = 0, -- 0 for search forward, 1 for search backward
     case_insensitive = true, -- default to case insensitive
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -152,7 +152,7 @@ function ReaderSearch:onShowSearchDialog(text, direction)
             if no_results then
                 local notification_text
                 if self._expect_back_results then
-                    notification_text = _("No results on previous pages")
+                    notification_text = _("No results on preceding pages")
                 else
                     notification_text = _("No results on following pages")
                 end
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -167,18 +167,22 @@ function ReaderSearch:onShowSearchDialog(text, direction)
             {
                 {
                     text = from_start_text,
+                    vsync = true,
                     callback = do_search(self.searchFromStart, text),
                 },
                 {
                     text = backward_text,
+                    vsync = true,
                     callback = do_search(self.searchNext, text, 1),
                 },
                 {
                     text = forward_text,
+                    vsync = true,
                     callback = do_search(self.searchNext, text, 0),
                 },
                 {
                     text = from_end_text,
+                    vsync = true,
                     callback = do_search(self.searchFromEnd, text),
                 },
             }
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -3,6 +3,8 @@ local ButtonDialog = require("ui/widget/buttondialog")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local InputDialog = require("ui/widget/inputdialog")
 local UIManager = require("ui/uimanager")
+local InfoMessage = require("ui/widget/infomessage")
+local T = require("ffi/util").template
 local logger = require("logger")
 local _ = require("gettext")
 
@@ -144,6 +146,13 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                 end
                 -- Don't add result pages to location ("Go back") stack
                 neglect_current_location = true
+            else
+                UIManager:show(
+                    InfoMessage:new{
+                        text = T(_("'%1' not found"), text),
+                        timeout = 2
+                    }
+                )
             end
         end
     end
@@ -162,22 +171,18 @@ function ReaderSearch:onShowSearchDialog(text, direction)
             {
                 {
                     text = from_start_text,
-                    vsync = true,
                     callback = do_search(self.searchFromStart, text),
                 },
                 {
                     text = backward_text,
-                    vsync = true,
                     callback = do_search(self.searchNext, text, 1),
                 },
                 {
                     text = forward_text,
-                    vsync = true,
                     callback = do_search(self.searchNext, text, 0),
                 },
                 {
                     text = from_end_text,
-                    vsync = true,
                     callback = do_search(self.searchFromEnd, text),
                 },
             }
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -6,7 +6,6 @@ local Notification = require("ui/widget/notification")
 local UIManager = require("ui/uimanager")
 local logger = require("logger")
 local _ = require("gettext")
-local T = require("ffi/util").template
 local ReaderSearch = InputContainer:new{
     direction = 0, -- 0 for search forward, 1 for search backward
     case_insensitive = true, -- default to case insensitive
@@ -148,8 +147,7 @@ function ReaderSearch:onShowSearchDialog(text, direction)
             else
                 UIManager:show(
                     Notification:new{
-                        text = T(_("'%1' not found"), text),
-                        timeout = 3,
+                        text = "Not found",
                     }
                 )
             end
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -2,8 +2,8 @@ local BD = require("ui/bidi")
 local ButtonDialog = require("ui/widget/buttondialog")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local InputDialog = require("ui/widget/inputdialog")
+local Notification = require("ui/widget/notification")
 local UIManager = require("ui/uimanager")
-local InfoMessage = require("ui/widget/infomessage")
 local T = require("ffi/util").template
 local logger = require("logger")
 local _ = require("gettext")
@@ -148,9 +148,9 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                 neglect_current_location = true
             else
                 UIManager:show(
-                    InfoMessage:new{
+                    Notification:new{
                         text = T(_("'%1' not found"), text),
-                        timeout = 2
+                        timeout = 3
                     }
                 )
             end
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -147,7 +147,7 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                 neglect_current_location = true
             else
                 UIManager:show(Notification:new{
-                    text = "Not found",
+                    text = _("Not found"),
                 })
             end
         end
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -74,10 +74,11 @@ function ReaderSearch:onShowSearchDialog(text, direction)
     local current_page
     local do_search = function(search_func, _text, param)
         return function()
-            local valid_link
+            local no_results = true -- for notification
             local res = search_func(self, _text, param)
             if res then
                 if self.ui.document.info.has_pages then
+                    no_results = false
                     self.ui.link:onGotoLink({page = res.page - 1}, neglect_current_location)
                     self.view.highlight.temp[res.page] = res
                 else
@@ -89,6 +90,7 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                     -- sometimes even xpointers that resolve to no page.
                     -- We need to loop thru all the results until we find one suitable,
                     -- to follow its link and go to the next/prev page with occurences.
+                    local valid_link
                     -- If backward search, results are already in a reversed order, so we'll
                     -- start from the nearest to current page one.
                     for _, r in ipairs(res) do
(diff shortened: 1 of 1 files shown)


frontend/apps/reader/modules/readersearch.lua
```diff
@@ -3,6 +3,8 @@ local ButtonDialog = require("ui/widget/buttondialog")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local InputDialog = require("ui/widget/inputdialog")
 local UIManager = require("ui/uimanager")
+local InfoMessage = require("ui/widget/infomessage")
```
Frenzie, line 6: order

poire-z:
> avoid the search_dialog buttons showing up over the infomessage and it seems to cause no problems anywhere.

Still looks a bit ugly to have the InfoMessage over the `<< < > >>` search dialog, which is innefective anyway after the InfoMessage disappears.
Probably hard for you to go further if you're editing all this on github and can't really test on the emulator, but some possible options:
- use a Notification instead of an InfoMessage
- close the search dialog before showing the InfoMessage `UIManager:close(self.search_dialog)`

hius07:
> which is innefective anyway after the InfoMessage disappears.

I thought that one might want to repeat the search in the opposite direction then.

poire-z:
> I thought that one might want to repeat the search in the opposite direction then.

I see that then it would show up only if there's no result on the initial search.
If you get some results, and use `>` or `>>` to navigate results, and reach the page with the last one, nothing is shown. For consistency, I guess in both case, you should show "no previous/further results"  - which could work with a less obstrusive Notification.

hius07:
> I see that then it would show up only if there's no result on the initial search.

Exactly, I felt a little bit uncomfortable trying to find highlighted search results over the screen.

frontend/apps/reader/modules/readersearch.lua
hius07: Done

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -2,7 +2,9 @@ local BD = require("ui/bidi")
 local ButtonDialog = require("ui/widget/buttondialog")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local InputDialog = require("ui/widget/inputdialog")
+local Notification = require("ui/widget/notification")
 local UIManager = require("ui/uimanager")
+local T = require("ffi/util").template
 local logger = require("logger")
 local _ = require("gettext")
```
Frenzie, line 10: Close. ^_^ (See the style guide.)
```suggestion
local logger = require("logger")
local _ = require("gettext")
local T = require("ffi/util").template
```

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -144,6 +146,13 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                 end
                 -- Don't add result pages to location ("Go back") stack
                 neglect_current_location = true
+            else
+                UIManager:show(
+                    Notification:new{
+                        text = T(_("'%1' not found"), text),
+                        timeout = 3
```
Frenzie, line 153:
```suggestion
                        timeout = 3,
```

hius07:
Thank you.

I see a problem with Notification that it is designed for "tiny notifications".
When the search string is long, the notification doesn't fit into the screen.

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -144,6 +145,13 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                 end
                 -- Don't add result pages to location ("Go back") stack
                 neglect_current_location = true
+            else
+                UIManager:show(
+                    Notification:new{
+                        text = T(_("'%1' not found"), text),
+                        timeout = 3,
```
poire-z, line 152: Please get rid of the timeout - I made them all be 2 seconds in https://github.com/koreader/koreader/commit/8f79c662cfc5e1d2fb714c7dcef7c3c9318a62ef , for consistency, your cup of tea :)

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -162,22 +170,18 @@ function ReaderSearch:onShowSearchDialog(text, direction)
             {
                 {
                     text = from_start_text,
-                    vsync = true,
```
poire-z, line 165: You still need to remove these @NiLuJe babies , now that you switched to Notifications ?

poire-z:
> When the search string is long, the notification doesn't fit into the screen.

Yep, that's a caveat with them. May be just don't put the string, just "No result found" ? A 2s or 3s timeout might just be not enough to focus on the search string and see where you made a typo.

poire-z:
> > I see that then it would show up only if there's no result on the initial search.

> Exactly, I felt a little bit uncomfortable trying to find highlighted search results over the screen.

(Dunno how feasible it is) but wouldn't you want these message/Notification also when you're navigating results and you reach the first or last one ? OK, you know you've reached it when the page doesn't change - but for consistency, they could be shown.

frontend/apps/reader/modules/readersearch.lua
hius07: I saw that @NiLuJe removed them when possible.
I played with this menu without `vsync` (flash_ui enabled) and faced no problems.

hius07:
> OK, you know you've reached it when the page doesn't change - but for consistency, they could be shown.

This will require different texts (for forward and backward searches), I'll think of it.

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -2,10 +2,10 @@ local BD = require("ui/bidi")
 local ButtonDialog = require("ui/widget/buttondialog")
 local InputContainer = require("ui/widget/container/inputcontainer")
 local InputDialog = require("ui/widget/inputdialog")
+local Notification = require("ui/widget/notification")
 local UIManager = require("ui/uimanager")
 local logger = require("logger")
 local _ = require("gettext")
-
```
poire-z, line 8: We want this space :)

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -144,6 +144,12 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                 end
                 -- Don't add result pages to location ("Go back") stack
                 neglect_current_location = true
+            else
+                UIManager:show(
+                    Notification:new{
+                        text = "Not found",
+                    }
+                )
```
poire-z, line 152: We have all of these kind of `show(Something:new...` (except one) on just 3 lines:
```lua
            text = "Some text",
        })
```

hius07:
Thank you for support and patience

frontend/apps/reader/modules/readersearch.lua
NiLuJe: You need a very specific range of devices to see the difference, and they're *definitely* needed, with no side-effects and unaffected devices, so, yep: please leave them be ;).

hius07: Excuse me, reverted.

hius07:
Now the notification shows search direction.
Now the notification is shown when we are on the last/first page with search results and there are no results forward/backward.

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -145,6 +146,17 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                 -- Don't add result pages to location ("Go back") stack
                 neglect_current_location = true
             end
+            if res==nil or (current_page~=nil and valid_link==nil) then
```
poire-z, line 149: This is a bit hard to understand - and may give doubts :) valid_link and current_page are not defined on PDFs (when `if self.ui.document.info.has_pages`).
Might be clearer if you add another bool variable `local has_results = false` and do simpler checks in the appropriate branches.

frontend/apps/reader/modules/readersearch.lua
```diff
@@ -145,6 +146,17 @@ function ReaderSearch:onShowSearchDialog(text, direction)
...
+                local notification_text
+                if self._expect_back_results then
+                    notification_text = _("No results backward")
+                else
+                    notification_text = _("No results forward")
```
poire-z, line 154: (Or `No previous result` / `No next result` ?)

frontend/apps/reader/modules/readersearch.lua
hius07: > valid_link and current_page are not defined on PDFs

Yes, that's the fact but...
I've tested it with several pdfs and it does work as designed.
Will you be so kind to check it with a real ebook?

frontend/apps/reader/modules/readersearch.lua
hius07: > (Or `No previous result` / `No next result` ?)

Is it okay for initial search with nothng found?

frontend/apps/reader/modules/readersearch.lua
poire-z: I won't be that kind :) as the issue is not about it working or not, just about the code being clearer :)
I'd rather see something like this:
```diff
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -73,11 +73,13 @@ function ReaderSearch:onShowSearchDialog(text, direction)
     local current_page
     local do_search = function(search_func, _text, param)
         return function()
+            local has_result = false
             local res = search_func(self, _text, param)
             if res then
                 if self.ui.document.info.has_pages then
                     self.ui.link:onGotoLink({page = res.page - 1}, neglect_current_location)
                     self.view.highlight.temp[res.page] = res
+                    has_result = true
                 else
                     -- Was previously just:
                     --   self.ui.link:onGotoLink(res[1].start, neglect_current_location)
@@ -140,10 +142,14 @@ function ReaderSearch:onShowSearchDialog(text, direction)
                     end
                     if valid_link then
                         self.ui.link:onGotoLink({xpointer=valid_link}, neglect_current_location)
+                        has_result = true
                     end
                 end
                 -- Don't add result pages to location ("Go back") stack
                 neglect_current_location = true
+                if not has_result then
+                    -- do notification stuff
+                end
             end
         end
     end
```
than having to wonder what cases `if res==nil or (current_page~=nil and valid_link==nil) then` checks for (because I still wonder :) I'm not sure the above `has_result` stuff is enough, or if you thought about other cases depending on current_page meaning that a first search has already been done, etc... if yes, this could be made  clearer with `has_result=true` in the appropriate places).

frontend/apps/reader/modules/readersearch.lua
poire-z: OK, well, I dunno :) English scholars to the rescue!

frontend/apps/reader/modules/readersearch.lua
hius07: Thank you for your questions, now I can explain why my code works for pdf.
For initial search when nothing found - it's clear, `res=nil`.

More complicated is case when we are, say, on the last page with search results and press `▷`. We want to see notification `No results forward`.

The point is that in that case `findText` returns different results for epub and for pdf!
For `self.ui.document.info.has_page` it returns `res=nil` and we need no more checks.

Regading proposition of adding several `has_result=true`. They all should be under `if` conditions for `current_page` and `valid_link`.
I've just traced the logic and compacted it into one expression.
The code is short and it does work.

poire-z: It's still hard to understand. For example:
`current_page~=nil and valid_link==nil` are you sure the `current_page~=nil ` check is necessary ?
Isn't just `if res==nil or valid_link==nil then` enough ?

> The code is short and it does work.

The fact we're having this discussion may show these 2 conditions are not enough :)

frontend/apps/reader/modules/readersearch.lua
NiLuJe: Oh, yes, this doesn't work ^^.

New question: is the distinction really necessary? I'm inclined to think a simple "No results" works just as well, and will in fact be less confusing.

frontend/apps/reader/modules/readersearch.lua
hius07: > are you sure the `current_page~=nil ` check is necessary ?

Yes it is, without it we shall get notification for every search for `self.ui.document.info.has_page` as `valid_link=nil` always for them.

poire-z: So, `current_page` serves as an indirect far fetched equivalent of `not self.ui.document.info.has_pages` (and not even its main meaning to distinguish initial vs subsequence searches) :/

Make it easy for future readers of that bit of code (that may be me or you :):
- there's no reason to move `valid_link` outside the else where it matters
- add a natural language `has_result` so this reads naturally, without having to wonder how/why/when a convoluted `if res==nil or (current_page~=nil and valid_link==nil) then` will trigger, like we're doing for the last 5 hours :)

frontend/apps/reader/modules/readersearch.lua
hius07: Imagine you are on the last page with the results highlighted and press `▷`.
Is just `No results` good for that?

Also for initial search. `No results` in the entire book or in the next pages only?

NiLuJe: Well, yes, no result for the action the user just requested makes sense to me, you're not supposed to have forgotten which button you've tapped on 10ms ago?

poire-z: Well, I might hit `>` without thinking much that I'm searching forward only.
`No result` won't help me realize I could find results backwards.
As we know that fact, it feels a bit sad to not let it known one way or another (although I probably won't read a 2s notification text - but the longer the text, the more I'll be curious to focus on it :)

NiLuJe: That raises another question, though: Assuming "No results ahead" / "No results behind": ahead/behind of *what*?

hius07: No results or Not found
... on previous pages / ... on next pages
... back / ... ahead
... backward / ... forward (onward)
... before / ... after

NiLuJe: IIRC, "page" isn't really what it's actually relative to, but at least it's something, and something that's mostly accurate in practice, so, yeah, that works.

previous / *following* pages, though ;).

hius07:
Thanks a lot

poire-z (approved):
Thanks for making it that clear :)

frontend/apps/reader/modules/readersearch.lua
Frenzie: I'd be inclined to say preceding / following.

Issue #7007, opened 2020-12-17
Insecure HTTPS->HTTP 302 redirects in OPDS catalogs fail but not made clear why


johnbeard:
* KOReader version: 2020.11.60 (nightly)
* Device: Emulator & Clara HD

#### Issue

Some OPDS catalogs point to URLs which do a 302 Redirect. For example Project Gutenburg links to
* https://www.gutenberg.org/ebooks/46.epub.noimages
but this redirects to:
* http://www.gutenberg.org/cache/epub/46/pg46.epub
and the file that gets saved is:
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="http://www.gutenberg.org/cache/epub/46/pg46.epub">here</a>.</p>
<hr>
<address>Apache Server at www.gutenberg.org Port 80</address>
</body></html>
```

#### Steps to reproduce

* Open Project Gutenburg OPDS catalog
* Try to download any book
* It will fail and a 302 redirect HTML file saved in its place (as above)
----
Similar for the ManyBooks catalog which is itself behind a 302: https://manybooks.net/opds/index.php → http://srv.manybooks.net/opds/index.php

And for books from the Flibuter catalog: https://www.flibusta.is/b/226848/epub → http://static.flibusta.is:443/converter/get/convert?url=http%3A%2F%2Fflibusta.is%2Fb%2F7412a0b774fdf8c0998b13e1ebeac686%2Fdownload&out=epub&md5=7412a0b774fdf8c0998b13e1ebeac686

Frenzie:
The error handling toward the user could be improved slightly, but not allowing downgrading security is not a bug.

Frenzie:
I.e., this describes a request that should not be allowed.
> 
    https://www.gutenberg.org/ebooks/46.epub.noimages
    but this redirects to:
    http://www.gutenberg.org/cache/epub/46/pg46.epub
    and the file that gets saved is:

johnbeard:
Oh, so it's the HTTP -> HTTPS downgrade that's triggering this?

johnbeard:
Annoyingly, for PG at least, HTTPS does work for the target: https://www.gutenberg.org/cache/epub/46/pg46.epub does exist and work, so their OPDS generator is busted.

johnbeard:
[...]

PR #7437, opened 2021-03-21
Bump base

NiLuJe:
https://github.com/koreader/koreader-base/pull/1331
https://github.com/koreader/koreader-base/pull/1333
https://github.com/koreader/koreader-base/pull/1335
https://github.com/koreader/koreader-base/pull/1334

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7437)


NiLuJe:
(Gutenberg is down, hence the test failure).

SteadyQuad:
Gutenberg.org is working but really slow, probably timing out in the test.

btw this is on gutenberg.org frontpage:

> Planned service change In April 2021, www.gutenberg.org plans to remove unencrypted access (http), and will only offer https. The collection remains available from mirrors, as well as via unencrypted ftp and rsync.

There's several gutenberg url's in the sourcecode that use http instead of https, worth creating an issue for?

NiLuJe:
IIRC, they're all in the hilariously outdated OPDS tests, but, yeah, at least this'll help tracking it.

Frenzie:
> There's several gutenberg url's in the sourcecode that use http instead of https, worth creating an issue for?

Probably, yes, but cf. https://github.com/koreader/koreader/issues/7007#issuecomment-749726860 as to why that wasn't the case.

PR #7440, opened 2021-03-22
Rotation icons: make 'em dynamic.

### LINKED ISSUE #7429: Screen rotation buttons counterintuitive
* KOReader version: 2021.02-66
* Device: Kobo Aura One

#### Issue
The buttons seem to show how one must rotate _the device_ in order to see the text upright, rather than how _the text_ will rotate.
To me at least that is highly counter-intuitive: I rotate _my device_ myself, but must tell the device how to rotate _the text_.

If people differ in their intuitions regarding this, then maybe it should be a setting "Natural rotating" (like "natural scrolling") - all it had to do was replace device-rotating icons by text-rotating ones.

#### Steps to reproduce
Click e.g. on the icon with the arcing arrow clockwise, and the text will rotate counterclockwise.

*Maybe* related - the remarks like the one on line 7041  in the [crash.log](https://github.com/koreader/koreader/files/6172569/crash.log).

NiLuJe:
Followup to #7306.

As discussed there, the only way to make everybody happy is to make 'em context-aware. At which point, trying to make the fact that it's a *device* rotation and not a *content* one come across in the icon design loses its interest, and would actually only further muddy the issue.

So, these are similar is spirit to the traditional rotation icons in pretty much anything that can rotate text content, with a special inspiration from the Kindle 4 & PocketBook UI.

Fix #7429

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7440)


NiLuJe:
UR:
0ur

UD:
180ud

CW:
90cw

CCW:
90ccw

NiLuJe:
The lack of rotation metadata in the screengrabs actually *helps* make the point across: notice how the "UR" icon (the second one from the left) always show you an upright A in a Portrait-shape ;).

frontend/ui/widget/configdialog.lua
```diff
@@ -486,8 +486,11 @@ function ConfigOption:init()
             end
 
             -- Icons (ex: columns, text align, with PDF)
-            if self.options[c].item_icons then
-                local items_count = #self.options[c].item_icons
+            local item_icons = self.options[c].item_icons_func and
+                               self.options[c].item_icons_func(self.config.configurable, self.config.document) or
```
poire-z, line 490: The 2nd argument is not used. Early idea abandonned, or ready for future ideas?
And even the 1st one, although passed, is not used in cre/koptions.lua.

NiLuJe: Yeah, that's to make the framework as versatile as the other `*_func` varian,t even if it isn't actually being used here.

(An earlier commit did try to get the value from the configurable, but it's fugly as hell, and less reliable).

PR #7447, opened 2021-03-26
PluginLoader should update lookup_path_list if data_dir is not ".".

Hzj-jie:
It looks like a problem introduced by https://github.com/koreader/koreader/commit/dee72a431c1ef4d59569ffc80b876e2e91765bbf#diff-b3b0e0e90ea7b43a9d97a1c7572acef3f1dc7d1d2f75fe634345a1b841c624fa. Or did I misunderstand it?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7447)

### DIFF
--- a/frontend/pluginloader.lua
+++ b/frontend/pluginloader.lua
@@ -66,8 +66,9 @@ function PluginLoader:loadPlugins()
     else
         local data_dir = require("datastorage"):getDataDir()
         if data_dir ~= "." then
-            extra_paths = { data_dir .. "/plugins/" }
-            G_reader_settings:saveSetting("extra_plugin_paths", extra_paths)
+            local extra_path = data_dir .. "/plugins/"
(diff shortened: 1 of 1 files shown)


poire-z:
Do I get the "problem" right ? : the bug is that extra_path would work only on the 2nd koreader launch, but not on the first ?

NiLuJe:
@poire-z: That's how I understood this when it was mentioned a few days ago, yes. But we very well may be getting it wrong, we'd just be wrong together ;p.

NiLuJe:
Sidebar: this changes does nothing ^^.

You probably wanted to update `lookup_path_list` instead?

Hzj-jie:
lol, it's really hard to make it right.

Hzj-jie:
Good to go?

poire-z (approved):

PR #7459, opened 2021-03-28
Kindle: Fix gyro event on Zelda

### LINKED ISSUE #7458: KOA3 bad orientation detection
* KOReader version: 2021.3
* Device: KOA3

#### Issue

On KOA3, KOReader doesn't rotate correctly based on the g-sensor.

#### Steps to reproduce

Open KOreader and rotate the device physically, counter clockwise. The text will turn upside down after ~45 degrees. 

#### Other info:
This issue was reported on KOA3 devices:
https://www.mobileread.com/forums/showpost.php?p=4106995&postcount=1084
https://www.mobileread.com/forums/showpost.php?p=4106405&postcount=5

Seems like the values [here](https://github.com/koreader/koreader/blob/master/frontend/device/input.lua#L54-L62) are different from KOA and KOA2.

NiLuJe:
Like on Kobo, only honor the face-up events.

Fix #7458


NiLuJe:
(Please hold off on merging, I'll rebase after #7415, the other way around would be extra-painful).

TrYoLz:
I've tested it on my KOA3 and the rotation works correctly now.

If the goal was to reproduce the orientation of the stock reader, I don't think this is this is the intended behaviour. The stock reader only switches between portrait and landscape mode with buttons. With this implementation, KOReader switches between them just from tilting the device. I'm not sure if this is intentional, and there is a way to disable switching between landscape and portrait somewhere else.

NiLuJe:
Yep, there's an orientation lock in the Screen settings.

TrYoLz:
> Yep, there's an orientation lock in the Screen settings.

Found it, my bad! Works perfectly, exactly like the stock reader. Good job! :)

PR #7460, opened 2021-03-29
Text editor: option to show/hide keyboard on start

hius07:
Allows to hide keyboard on start to see more text.
Default is to show keyboard.

Some wordings unification.
`Open file` goes down, closer to the list of opened files.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7460)

### DIFF
--- a/plugins/texteditor.koplugin/main.lua
+++ b/plugins/texteditor.koplugin/main.lua
@@ -478,7 +478,7 @@ function TextEditor:editFile(file_path, readonly)
     local buttons_first_row = {}  -- First button on first row, that will be filled with Reset|Save|Close
     if is_lua then
         table.insert(buttons_first_row, {
-            text = _("Check Lua"),
+            text = _("Lua check"),
             callback = function()
                 local parse_error = util.checkLuaSyntax(input:getInputText())
                 if parse_error then
--- a/plugins/texteditor.koplugin/main.lua
+++ b/plugins/texteditor.koplugin/main.lua
@@ -67,7 +67,7 @@ function TextEditor:loadSettings()
     self.auto_para_direction = self.settings:nilOrTrue("auto_para_direction")
     self.force_ltr_para_direction = self.settings:isTrue("force_ltr_para_direction")
     self.qr_code_export = self.settings:nilOrTrue("qr_code_export")
-    self.show_kbd_on_start = self.settings:nilOrTrue("show_kbd_on_start")
+    self.show_keyboard_on_start = self.settings:nilOrTrue("show_keyboard_on_start")
 end
 
 function TextEditor:onFlushSettings()
@@ -80,7 +80,7 @@ function TextEditor:onFlushSettings()
         self.settings:saveSetting("auto_para_direction", self.auto_para_direction)
         self.settings:saveSetting("force_ltr_para_direction", self.force_ltr_para_direction)
         self.settings:saveSetting("qr_code_export", self.qr_code_export)
-        self.settings:saveSetting("show_kbd_on_start", self.show_kbd_on_start)
+        self.settings:saveSetting("show_keyboard_on_start", self.show_keyboard_on_start)
         self.settings:flush()
     end
 end
@@ -173,10 +173,10 @@ Export text to QR code, that can be scanned, for example, by a phone.]]),
                 {
                     text = _("Show keyboard on start"),
                     checked_func = function()
-                        return self.show_kbd_on_start
+                        return self.show_keyboard_on_start
                     end,
                     callback = function()
-                        self.show_kbd_on_start = not self.show_kbd_on_start
+                        self.show_keyboard_on_start = not self.show_keyboard_on_start
                     end,
                 },
             },
@@ -516,7 +516,7 @@ function TextEditor:editFile(file_path, readonly)
         cursor_at_end = false,
         readonly = readonly,
         add_nav_bar = true,
-        keyboard_hidden = not self.show_kbd_on_start,
+        keyboard_hidden = not self.show_keyboard_on_start,
         scroll_by_pan = true,
         buttons = {buttons_first_row},
         -- Set/save view and cursor position callback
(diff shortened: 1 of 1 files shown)


plugins/texteditor.koplugin/main.lua
```diff
@@ -66,6 +67,7 @@ function TextEditor:loadSettings()
     self.auto_para_direction = self.settings:nilOrTrue("auto_para_direction")
     self.force_ltr_para_direction = self.settings:isTrue("force_ltr_para_direction")
     self.qr_code_export = self.settings:nilOrTrue("qr_code_export")
+    self.show_kbd_on_start = self.settings:nilOrTrue("show_kbd_on_start")
```
poire-z, line 70: No real need to abbreviate it, `show_keyboard_on_start` is not much longer, and reads nicer :)

plugins/texteditor.koplugin/main.lua
```diff
@@ -167,23 +170,32 @@ Export text to QR code, that can be scanned, for example, by a phone.]]),
                         self.qr_code_export = not self.qr_code_export
                     end,
                 },
+                {
+                    text = _("Show keyboard on start"),
+                    checked_func = function()
+                        return self.show_keyboard_on_start
```
poire-z, line 176: This setting is more worthy / "core" text editor than "QR code export", so may be move it above "QR code export" ?
Also, a separator between the settings related to paragraph direction and this Keyboard+QR could be nice ?

poire-z (approved):

plugins/texteditor.koplugin/main.lua
```diff
@@ -96,10 +99,10 @@ function TextEditor:getSubMenuItems()
     self.whenDoneFunc = nil -- discard reference to previous TouchMenu instance
     local sub_item_table = {
         {
-            text = _("Text editor settings"),
+            text = _("Settings"),
             sub_item_table = {
                 {
-                    text = _("Set text font size"),
+                    text = _("Text font size"),
```
Frenzie, line 105: I'd make it either just:
```suggestion
                    text = _("Font size"),
```
Or:
```suggestion
                    text = _("Editor font size"),
```

hius07: I made it unified with
https://github.com/koreader/koreader/blob/861214ce2ab3c820ba2475b7ea4207a7cfdde888/frontend/apps/filemanager/filemanagermenu.lua#L192
https://github.com/koreader/koreader/blob/e33476996acfa9aa2dab2c6e35c8eda4545100ce/frontend/apps/reader/modules/readertoc.lua#L1027
https://github.com/koreader/koreader/blob/e33476996acfa9aa2dab2c6e35c8eda4545100ce/frontend/apps/reader/modules/readerbookmark.lua#L99
In the text edtor there are no other items but text. Looks short and clear.

Frenzie: > I'm fine with either "Text font size" or "Font size" - but not with "Editor font size" (which, for me, could evoke the editor title and buttons font size).

That's exactly my rationale, except "text" could be anything and "editor" specifically means the editor component.

hius07: We have much worse cases in the font size dialogs e.g.
https://github.com/koreader/koreader/blob/bf6c0cdd6c5b22bbb38497b5df8abe428eb80307/frontend/apps/reader/modules/readercoptlistener.lua#L316

Frenzie: Are you saying the size of that status bar and the font size aren't interlinked? Otherwise there's nothing that can be confusing about it.

Frenzie: NB It might just as well be that `cre_header_status_font_size` is misleading and that the text is completely accurate. You'd have to check crengine code to see.

hius07: The menu item `Font size` invokes the dialog titled `Size of top status bar`.
It's not good imho even if the font size and bar height are interlinked.

Frenzie: Apologies, I misunderstood. I don't quite understand its relevance to the discussion here though. ^_^

Frenzie: @poire-z said "editor font size" evokes the size of the title and buttons.

I said text could be anything.

And then I thought you said something like "this one here doesn't even say text!"

hius07:
If we have QR enabled `Check Lua` doesn't fit.



Should it be changed to?

poire-z:
I guess I'm fine with both wording.
(It fits for me on my device, so it depends on screen size & DPI - some other things are truncated on my small emulator, but fine on my larger screen device.)

Frenzie:
A rather geeky alternative could be `Lint`. (No objections, though I prefer "Check Lua". Does this happen on actual commonly used devices?)

hius07:
> Does this happen on actual commonly used devices?)

The screenshots are from my Kindle PW2.

NiLuJe:
Which is 758x1024, IIRC.

poire-z:
Ready to merge, or still some things to decide?

plugins/texteditor.koplugin/main.lua
```diff
@@ -28,6 +28,7 @@ local TextEditor = WidgetContainer:new{
     history_keep_size = 60, -- hom many to keep in settings
     normal_font = "x_smallinfofont",
     monospace_font = "infont",
+    default_font_size = 20, -- x_smallinfofont default size
```
Frenzie, line 31: Should we pull that from Font.sizemap.x_smallinfofont or something?

Frenzie (approved):
Seems okay, added a question up to @hius07 or @poire-z

hius07:
Thank you, I'm done.

plugins/texteditor.koplugin/main.lua
poire-z: Yes, we could have. For next time then :)
(Although all these names don't really mean much when you read it... I usually select a font size that pleases me, and pick a name that happens to have that same size :)

Issue #7228, opened 2021-02-02
Default Fit mode not being applied to manga


NO-ob:
* KOReader version: 2021.01.1, 2021.01, 2020.12
* Device: Kobo clara hd

#### Issue
When loading a manga with view set as page and Fit set to full, the manga will not be in full mode. when trying to scroll it goes down the page a bit as its not being fit to the screen so changing page requires 2 presses. It is fixed by opening the menu and clicking on full so it is applied but this has to be done for every manga

#### Steps to reproduce
-open manga
-load bottom menu
-set fit to full and hold down to set as default
-set view mode to page and hold down to set as default
-load another manga
-settings will not be applied

Frenzie:
Incidentally, is it possible that there's a bad interaction with another setting like reflow?

NO-ob:
I have reflow set to off so not sure if its that, ive also tried setting page crop to none and margin to small so I dont think its those either

Galunid:
So at a quick glance `ui/data/koptoptions.lua` and `apps/reader/modules/readercropping.lua`, `apps/reader/modules/readerzooming.lua` seem to use different settings.

Frenzie:
It was just a quick thought before testing. I can confirm the problem. It doesn't exist in 2020.11, so it's a regression in #6885.

Frenzie:
@Galunid Concretely, I mean it was a hypothesis related to user settings, not to the code.

NO-ob:
> It was just a quick thought before testing. I can confirm the problem. It doesn't exist in 2020.11, so it's a regression in #6885.
The 2020.11 build doesn't have the fit setting?

Frenzie:
In 2020.11 it's in the top menu. The new part is the quickly customizable row/column navigation, the rest is a UX improvement.

Frenzie:
Where by 2020.11 I mean something like 2012.04 – 2020.11, but that aside.

NO-ob:
Oh okay just got confused as I installed that build and didn't see it there. Will use that older build for now thankyou

gbyl:
> when trying to scroll it goes down the page a bit as its not being fit to the screen so changing page requires 2 presses. It is fixed by opening the menu and clicking on full so it is applied but this has to be done for every manga

I can confirm that this is issue exists with PDFs as well

Frenzie:
It'd be quite surprising if it didn't. ^_^

Frenzie:
Will be closed by #7463 soon.

[...]

Issue #7192, opened 2021-01-23
pdf: setting default "Zoom to" lacks description, "Set default Zoom to to 4?"


Thra11:
* KOReader version: v2021.01
* Device: reMarkable 1

#### Issue
The popup to set the default "Zoom to" value just has an integer value for each option, instead of a description/name.

#### Steps to reproduce
When reading a pdf, open the bottom settings and choose the third tab. The second option down is "Zoom to", with 5 icons to choose from. If you press and hold on one of these icons, a popup will appear with, for example, "Set default Zoom to to 4?", as the text (Options are 4,3,2,1,0 from left to right).

Having a name / description in the popup instead of the raw number would also be helpful as clarification if anyone is unsure exactly what each icon means.

Frenzie:
PS Avoid clever programming reflexes. Each type should be a full string.

NiLuJe:
Fixed by #7463

PR #7463, opened 2021-03-29
ReaderZooming: Fix defaults handling

### LINKED ISSUE #7461: Default zoom fit option not taken into account for CBZ, always fit the width
* KOReader version:v2021.03
* Device: PB740 Pocketbook inkpad 3 (Vivlio firmware) 

#### Issue
For CBZ files, the zoom is set by default to fit the width of the page even though it is supposed to be height or full
#### Steps to reproduce
I select zoom 4 and fit to full by default. Whenever I open a new CBZ file, it behaves as if it is set to fit the width by default, even though when I go to change the zoom option, I can see that my default is selected (i.e. zoom 4 and fit to full). I have to change the fit option and then put it back to full again for it to behave as expected.

NiLuJe:
It appears the fancy split settings from #6885 were not being honored at all.

Fix #7461, #7228, #7192

(Also, re-ensured "pagewidth" is actually the default zoom mode, "full" had sneaked in as the default state of the zoom type toggle).

Also, display human-readable values in the "Make as default" popup, instead of the raw, meaningless numerical setting.

Also, disable zoom options when reflow is enabled, and restore 'em properly when it's disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7463)

### DIFF
--- a/frontend/ui/data/koptoptions.lua
+++ b/frontend/ui/data/koptoptions.lua
@@ -270,7 +270,7 @@ In 'semi-auto' and 'manual' modes, you may need to define areas once on an odd p
                 name = "zoom_direction",
                 name_text = _("Direction"),
                 enabled_func = function(config)
-                    return optionsutil.enableIfEquals(configurable, "text_wrap", 0) and config.zoom_mode_genus < 3
+                    return optionsutil.enableIfEquals(config, "text_wrap", 0) and config.zoom_mode_genus < 3
                 end,
                 item_icons = {
                     "direction.LRTB",
--- a/frontend/ui/data/koptoptions.lua
+++ b/frontend/ui/data/koptoptions.lua
@@ -119,6 +119,10 @@ In 'semi-auto' and 'manual' modes, you may need to define areas once on an odd p
             {
                 name = "zoom_overlap_h",
                 name_text = _("Horizontal overlap"),
+                enabled_func = function(configurable, document)
+                    -- NOTE: document.is_reflowable is wonky as hell, don't trust it.
+                    return optionsutil.enableIfEquals(configurable, "text_wrap", 0)
+                end,
                 buttonprogress = true,
                 fine_tune = true,
                 values = {0, 12, 24, 36, 48, 60, 72, 84},
@@ -137,6 +141,10 @@ In 'semi-auto' and 'manual' modes, you may need to define areas once on an odd p
             {
                 name = "zoom_overlap_v",
                 name_text = _("Vertical overlap"),
+                enabled_func = function(configurable, document)
+                    -- NOTE: document.is_reflowable is wonky as hell, don't trust it.
+                    return optionsutil.enableIfEquals(configurable, "text_wrap", 0)
+                end,
                 buttonprogress = true,
                 fine_tune = true,
                 values = {0, 12, 24, 36, 48, 60, 72, 84},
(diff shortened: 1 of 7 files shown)


Frenzie (approved):
Looks okay, no time to test.

Galunid:
It should close https://github.com/koreader/koreader/issues/7228 too (as #7461 is a duplicate of #7228)

poire-z:
Mhhh, as you are in Pan Zoom Fix Mode, I'll shamelessly hint 2 other related issues/questions:
https://github.com/koreader/koreader/pull/6885#issuecomment-743409322 footer (time) not upated when panning
https://github.com/koreader/koreader/pull/6885#issuecomment-757911409 refresh glitches when reflow enabled
Just in case you feel like having more :)

NiLuJe:
@poire-z: Haha, nooope. :D.

Unless we're okay with disabling the icons if reflow is enabled, that I can (probably ;p) do?

frontend/ui/data/koptoptions.lua
```diff
@@ -259,6 +260,16 @@ In 'semi-auto' and 'manual' modes, you may need to define areas once on an odd p
                 },
                 alternate = false,
                 values = {7, 6, 5, 4, 3, 2, 1, 0},
+                labels = {
+                    _("Left to Right, Top to Bottom"),
```
NiLuJe, line 264: FWIW, CamelCase was voluntary here.

(One, it helped me make sure I didn't get it wrong, two, I kind of like it? ^^).

Frenzie:
> Unless we're okay with disabling the icons if reflow is enabled, that, I can (probably ;p) do?

Iirc nothing other than page and page width makes sense combined with reflow.

NiLuJe:
If I close my eyes and sing *lalala* really loud, can I assume that *enabling* reflow murders the settings in question and I don't have to even check for that? ^^.

Galunid:
@NiLuJe Seems closing your eyes and singing *lalala* didn't work https://github.com/koreader/koreader/pull/6885#issuecomment-757911409

poire-z:
> can I assume that enabling reflow murders the settings in question and I don't even have to check for that

I don't read PDF often enough to have any kind of workflow - but I guess if I have sen pan/zoom settings to my liking, and I hit Reflow on (to see, for testing, or by error), and hit Reflow off, I'll be sad losing my previous settings :)
So, may be just don't close your eyes, don(t sing (really!), and don't fix it :)

NiLuJe:
@poire-z: That actually works.... as long as the document stays open.

Al bets are off after a close, when toggling reflow off actually resets the zoom to default.

So, I kept that weirdness alive because I have no intention of ever looking at reflow code, and it will reset the zoom mode properly on doc open if reflow is on.

NiLuJe:
Eh, fixed it.

poire-z (approved):
Thanks!
Not tested, and not familiar with that code, but looks allright :)

poire-z:
> Just in case you feel like having more :)

Just in case ... https://github.com/koreader/koreader-base/pull/1203 :)

NiLuJe:
Well played ;p.

Frenzie:
> Just in case ... koreader/koreader-base#1203 :)

The biggest problem is I don't really know what's not done about it yet. I tested it on desktop and it seemed to be hunky dory, but I'm guessing on some target (*cough*Android presumably*cough*) there's still work to be done.

NiLuJe:
I skimmed it a few weeks ago and there *is* some possibly non-trivial stuff missing (or papered over).

PR #7465, opened 2021-03-30
bump crengine: add support for MathML

### LINKED ISSUE #981: MathML Support
Sorry if this isn't the properly place to ask, but koreader supports MathML?

### LINKED ISSUE #1367: MathML and EPUB
Hello,

First thank you for this program ! I'm sorry if I wrote in the wrong place, but I've a little question and I don't know where I should ask it: is it possible to read MathML formula from an epub file (function available from epub3 that let you write mathematical functions) ?

Thank you,
Tobias.

Example: http://docs.mathjax.org/en/latest/misc/epub.html
http://manual.calibre-ebook.com/typesetting_math.html

-- EDIT --
Sorry, I didn't see this before : https://github.com/koreader/koreader/issues/981. I think it's still the case...

### LINKED ISSUE #6678: KOReader does not render MathML correctly in epub3 books, and does not fall back to altimg
* KOReader version: v2020.09
* Device: Kobo Aura H20 2 v2

#### Issue
The EPUB 3 specification allows XHTML documents to contain MathML markup (http://idpf.org/epub/30/spec/epub30-contentdocs.html#sec-xhtml-mathml), and it describes how documents can use an `altimg` attribute on the `math` element to provide an alternate image to show for readers that don't support MathML (http://idpf.org/epub/30/spec/epub30-contentdocs.html#sec-xhtml-mathml-alt). However, KOReader completely ignores the MathML elements in EPUB 3 documents, and thus it neither renders the MathML content correctly, nor does it use the fallback image.

#### Steps to reproduce
Open an EPUB 3 document containing MathML (an example document is attached) in KOReader.

[mathml.zip](https://github.com/koreader/koreader/files/5248611/mathml.zip)

### LINKED ISSUE #7366: Uneven stretching of space at the beginning of paragraph
* KOReader version: v2021-02-20
* Device: emulator (and Kobo Aura)

#### Issue

When a paragraph begins with `. ` (period, space; this can happen when ellipses are typeset as `. . .`), the first space does not stretch like the other spaces to justify the line, creating a very ugly effect

#### Steps to reproduce

* Open [test.zip](https://github.com/koreader/koreader/files/6056630/test.zip) (rename to .epub if necessary).

* Render with justified text, hyphenation disabled.

Screenshot_20210228_114656

Of course the huge spaces in the first line of the second paragraph are not nice, but that's an intended exaggeration here. The point is the circled ellipsis with an extremely uneven spacing.

poire-z:
Includes (among others):
- LVImg: Tweak JPEG decoding some more https://github.com/koreader/crengine/pull/427
- toStringV2(): fix (again) when target node is a boxing node https://github.com/koreader/crengine/pull/428
- LVFontCache::find(): give more weight to first fonts in list
- Page splitting: more accurate rendering progress
- getRenderedWidths(): fix nowrap around image/inlineBoxes
- Tables rendering: tweak column widths algorithm
- CSS: parse/handle "currentcolor", default for border-color
- CSS: add units 'ch' (just like 'ex')
- SVG images: proper alpha blending
- MathML: add parsing and rendering support files https://github.com/koreader/crengine/pull/429
- MathML: plug MathML code into crengine core
- MathML: <epub:switch/case/default>: accept MathML
- (Upstream) Make crengine.font.fallback.faces plural https://github.com/koreader/crengine/pull/430
- (Upstream) Option to not limit font size to a set
- Text: dont adjust space after consecutive initial marks/dashes - Closes #7366
- Update German hyphenation patterns https://github.com/koreader/crengine/pull/431

Closes #981. Closes #1367. Closes #6678. (MathML issues).

About Math fonts: the code will try to use a OpenType Math font (with specific info about math spacings and thickness, etc...). We do ship one math font: FreeSerif which is quite ok (but with integral signs a bit too thick, and the math axis a bit off).
The code will look for the following other known math fonts in that order - and will use the first one found. So, if you find a nice one among these, just drop this one in your koreader/font/ directory. (You can drop more, and select one from the font menu, but it will then apply also to the non-math text. or you can make user style tweaks with `math { font-family: "XITS Math" !important }` to force one so the code does not have to find another one.
```
"Latin Modern Math"
"STIX Two Math"
"XITS Math"
"Libertinus Math"
"TeX Gyre Termes Math"
"TeX Gyre Bonum Math"
"TeX Gyre Schola Math" (naming typo of mine for this one in the code, will be fixed later)
"TeX Gyre Pagella Math"
"DejaVu Math TeX Gyre"
"Asana Math"
"Cambria Math"
"Lucida Bright Math"
"Minion Math"
"FreeSerif"
```

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7465)

### DIFF
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -934,7 +934,7 @@ function CreDocument:setupFallbackFontFaces()
     -- names than ',' or ';', without the need to have to use quotes.
     local s_fallbacks = table.concat(fallbacks, "|")
     logger.dbg("CreDocument: set fallback font faces:", s_fallbacks)
-    self._document:setStringProperty("crengine.font.fallback.face", s_fallbacks)
+    self._document:setStringProperty("crengine.font.fallback.faces", s_fallbacks)
 end
 
 -- To use the new crengine language typography facilities (hyphenation, line breaking,


Frenzie:
Speaking of style tweaks, DBNL seems to use .footnote-link.

NiLuJe:
And speaking of footnotes, do we actually support the ePub 3 `epub:type="footnote"` [thingy](https://www.w3.org/publishing/epub3/epub-contentdocs.html#sec-epub-type-attribute)?

poire-z:
Of course !
With footnotes popup:
https://github.com/koreader/koreader/blob/ffc7ccb576df7194bbb1dda97a3746db67bc9906/frontend/apps/reader/modules/readerlink.lua#L1186-L1187
and with in-page footnotes:
https://github.com/koreader/koreader/blob/ffc7ccb576df7194bbb1dda97a3746db67bc9906/frontend/ui/data/css_tweaks.lua#L713-L730

> Speaking of style tweaks, DBNL seems to use .footnote-link.

As the source link (a href=) or the target (anyelement id=) ? In-page footnotes want it on the target (which must be a block element that contains the ID if I remember right).
Feel free to test and add it among these if it's that consistent:
https://github.com/koreader/koreader/blob/ffc7ccb576df7194bbb1dda97a3746db67bc9906/frontend/ui/data/css_tweaks.lua#L791-L803

poire-z:
btw @NiLuJe, what's the right way to avoid these? Seems like my prefixing with `(void)` wasn't enough:
```
../../crengine/crengine/src/lvfnt.cpp: In function ‘int lvfontOpen(const char*, void**)’:
../../crengine/crengine/src/lvfnt.cpp:74:37: warning: ignoring return value of ‘size_t fread(void*, size_t, size_t, FILE*)’, declared with attribute warn_unused_result [-Wunused-result]
     (void) fread( *pfont, sz, 1, f );
                                     ^
../../crengine/crengine/src/crtxtenc.cpp: In function ‘void MakeStatsForFile(const char*, const char*, const char*, int, FILE*, lString8&)’:
../../crengine/crengine/src/crtxtenc.cpp:2084:38: warning: ignoring return value of ‘size_t fread(void*, size_t, size_t, FILE*)’, declared with attribute warn_unused_result [-Wunused-result]
    (void) fread(buf, 1, buf_size, in);
                                      ^
```

NiLuJe:
Not ignoring the return value? ;o).

I'm halfway not joking, that's a FORTIFY feature of the glibc that you can't really bypass without killing hardening, combined with a GCC feature that makes the `warn_unused_result` unavoidable.

Which leaves push'ing/pop'ing ignore pragmas around the calls if you *really* don't want to check it.

Frenzie:
> As the source link (a href=) or the target (anyelement id=) ? In-page footnotes want it on the target (which must be a block element that contains the ID if I remember right).

Right, good point. I didn't look into it too much because [the particular book I'm reading](https://www.dbnl.org/tekst/lapi001goet01_01/colofon.php) has a grand total of one footnote. xD

The actual footnotes are `class="voetnoten"`, which could potentially be fairly standard in Dutch. I'll look into it if I remember to.
```css
.voetnoten {-cr-hint: footnote-inpage}
```

Frenzie:
DBNL footnotes added to the standard set in <https://github.com/koreader/koreader/pull/7467>.

poire-z:
Just dropping some attempt to have Wikipedia EPUB show the MathML instead of PNG images - as the Wiki source may contain some hidden MathML, mostly provided for "assistive" stuff.
I've been trying it with https://en.wikipedia.org/wiki/Fourier_transform
It would work, but sometimes, the stuff is wrapped in some ugly table code (and somehow, images survives that as they make surrounding stuff adjust to contain them), or needs some other CSS rules for classnames to hide stuff - that it's most often just ugly. And there's some inline formulas that you would think is MathML - but no, it's a bunch of inline-block & vertical-align & margins...
It's currently already ugly - but I don't want a migration to MathML to make it worse, or get more suspicions that our MathML rendering sucks :)

```diff
--- a/frontend/ui/wikipedia.lua
+++ b/frontend/ui/wikipedia.lua
@@ -690,2 +690,7 @@ function Wikipedia:createEpub(epub_path, page, lang, with_images)
     local processImg = function(img_tag)
+        print(img_tag)
+        if img_tag:find([[ class="mwe-math-fallback-image-inline"]], 1, true) then
+            logger.info("ignoring MathML fallback image", img_tag)
+            return ""
+        end
         local src = img_tag:match([[src="([^"]*)"]])
@@ -784,3 +789,3 @@ function Wikipedia:createEpub(epub_path, page, lang, with_images)
     html = html:gsub("(<%s*img [^>]*>)", processImg)
-    logger.dbg("Images found in html:", images)
+--    logger.dbg("Images found in html:", images)

@@ -1117,2 +1122,11 @@ table {
 /* hiding .noprint may discard some interesting links */
+
+/* Show MathML */
+.mwe-math-element > .mwe-math-mathml-a11y {
+    display: inline !important;
+}
+.mwe-math-element > .mwe-math-mathml-a11y + img {
+    display: none;
+}
+.sr-only { display: none; }
 ]])
@@ -1216,3 +1230,3 @@ table {
     -- So, remove the whole <math>...</math> content
-    html = html:gsub([[<math xmlns="http://www.w3.org/1998/Math/MathML".-</math>]], "")
+--    html = html:gsub([[<math xmlns="http://www.w3.org/1998/Math/MathML".-</math>]], "")

```

Not pursuing on this.

PR #7475, opened 2021-03-31
File manager: disable forbidden or dangerous actions for parent folder

hius07:
For parent folder (..):
disabled `Copy`, `Cut`, `Delete`, `Rename`.
enabled `Paste`, `Set as HOME folder`.

Added `/` to the dialog title for folders (same as in the files list) to distinguish folders and files (in addition to different sets of the bottom buttons).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7475)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -895,7 +895,7 @@ function FileManager:pasteHere(file)
             if mode == "file" then
                 text = T(_("The file %1 already exists. Do you want to overwrite it?"), BD.filename(basename))
             else
-                text = T(_("The directory %1 already exists. Do you want to overwrite it?"), BD.directory(basename))
+                text = T(_("The folder %1 already exists. Do you want to overwrite it?"), BD.directory(basename))
             end
 
             UIManager:show(ConfirmBox:new {
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -227,10 +227,14 @@ function FileManager:setupLayout()
     local fileManager = self
 
     function file_chooser:onFileHold(file)  -- luacheck: ignore
+        local is_file = lfs.attributes(file, "mode") == "file"
+        local is_folder = lfs.attributes(file, "mode") == "directory"
+        local is_not_parent_folder = BaseUtil.basename(file) ~= ".."
         local buttons = {
             {
                 {
                     text = C_("File", "Copy"),
+                    enabled = is_not_parent_folder,
                     callback = function()
                         copyFile(file)
                         UIManager:close(self.file_dialog)
@@ -264,6 +268,7 @@ function FileManager:setupLayout()
             {
                 {
                     text = _("Cut"),
+                    enabled = is_not_parent_folder,
                     callback = function()
                         cutFile(file)
                         UIManager:close(self.file_dialog)
@@ -271,6 +276,7 @@ function FileManager:setupLayout()
                 },
                 {
                     text = _("Delete"),
+                    enabled = is_not_parent_folder,
                     callback = function()
                         UIManager:show(ConfirmBox:new{
                             text = _("Are you sure that you want to delete this file?\n") .. BD.filepath(file) .. ("\n") .. _("If you delete a file, it is permanently lost."),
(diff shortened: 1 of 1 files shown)


frontend/apps/filemanager/filemanager.lua
```diff
@@ -436,8 +443,8 @@ function FileManager:setupLayout()
         end
 
         local title
-        if lfs.attributes(file, "mode") == "directory" then
-            title = BD.directory(file:match("([^/]+)$"))
+        if is_folder then
+            title = BD.directory(file:match("([^/]+)$")).."/"
```
poire-z, line 447: Not super sure about it, but it feels like the trailing / should be part of what will be BD.directory()'ed (some BiDi unicode control chars so it gets ok in BiDi text).

hius07: > the trailing / should be part of what will be BD.directory()'ed

I don't have it on Kindle.
If it is on other systems, my addition should be removed for sure.

poire-z: You don't have what ? :)
It's something that you might only see in an Arabic/Hebrew UI, or in a LTR UI with maybe files and directories in arabic or hebrew.
I don't remember much about all that, but it's probably less risky to have it output what would be happening here:
https://github.com/koreader/koreader/blob/de543b348468820aef5015285e35cd740ea2a4e1/frontend/ui/widget/filechooser.lua#L281-L287

So, in your case, maybe:
```lua
        if is_folder then
            local dir_name = file:match("([^/]+)$") -- I'm not sure this might not returns multiple values, so keep the ../ for the next one
            title = BD.directory(dir_name .. "/")
```

poire-z: > `-- will keep any trailing / on the right`

That's the point: with your current code, in an Arabic UI, if the directory/title is `/koreader`, and you add the `/` after BD.directory(), it may get displayed as : `//koreader`.
If you add the / inside BD.directory(), you should get the expected `/koreader/`

hius07: Thank you, this bidi thing requires more learning (for me), leave it for future.

Frenzie (commented):
I thought I already left this comment, but could you expand on the rationale?

Frenzie:
> Rename always fails.

Excellent.

> Delete is catastrophic for non-expirienced user.

Not sure what you mean by catastrophic exactly, but it does seem to result in a crash (if you click the wrong thing afterward). So it should indeed be disabled since I'm not sure if fixing that crash really makes any sense.

> Copy has no sense as Paste always ask for overwriting of existing folder and it isn't clear what files/folders will be overwritten.

Shouldn't they be both disabled or both enabled? This is the one I was mainly wondering about really.

Frenzie:
But why should it be disabled? Does it not copy what you expect it to copy?

poire-z:
> If we copy/cut a file and Paste over .., the file will be copied/moved to the parent folder.
> I remember @poire-z likes this feature.

Good memory :)
What I like is, when in a directory with files and subdirectories, being able to long-press a file and Cut - and the be able to just long-press a sibling directory there and Paste - without needing to _enter_ this target directory.
Moving a file up by pasting onto `../`, I use this quite rarely, but yes, I think it makes the same kind of sense and shortcut.

hius07:
Are you asking, why to disable `Copy` in the long-press menu on `..`?
Because it copies to the clipboard `..`, and when you try to paste it anywhere, the system says "The directory .. already exists" and propose to overwrite. I haven't tried, which folder would be overwrtten, and don't want to.

Frenzie:
> Because it copies to the clipboard .., and when you try to paste it anywhere, the system says "The directory .. already exists" and propose to overwrite. I haven't tried, which folder would be overwrtten, and don't want to.

Okay, so you're saying it's *broken* so there's no choice but to fix or disable it. That's very different rationale from it not making sense as you were implying.

frontend/apps/filemanager/filemanager.lua
```diff
@@ -888,7 +895,7 @@ function FileManager:pasteHere(file)
             if mode == "file" then
                 text = T(_("The file %1 already exists. Do you want to overwrite it?"), BD.filename(basename))
             else
-                text = T(_("The directory %1 already exists. Do you want to overwrite it?"), BD.directory(basename))
+                text = T(_("The folder %1 already exists. Do you want to overwrite it?"), BD.directory(basename))
```
Frenzie, line 898: A bit wordy.
```suggestion
                text = T(_("The folder %1 already exists. Overwrite folder?"), BD.directory(basename))
```

Frenzie (approved):
Fine by me, dunno what @poire-z thinks?

poire-z (approved):
Fine with it since its opening :)

Issue #7473, opened 2021-03-31
Dictionary: tapping issues which affects the Reader


gerhaher:
* KOReader version: v2021.03-12-gde543b3_2021-03-31/ (firmware 4.26.16704)
* Device: KA1

#### Issue
I made a film that shows what happen after the issue is triggered:
https://drive.google.com/file/d/1gPhUXikzMRKls7pmfyRK9gNZKMKrbUDz/view?usp=sharing
What you see is first that I can't change page with a single tap. Instead I need to double tap to do that. And when I tap << & >> I end up in first and last dictionary (and not in next).
Second, when I go back to the Reader, when I tap on the screen the dictionary opens again, every time. And when I swipe, well the dictionary is there again, intruding.
What is not shown in the video is that I can page backward/forward in the Reader by triple tapping.

So in the end I have to make a forced shut down of the device.

I noticed the issue with v2021.03-6-g1f15184_2021-03-30/, I didn't see it with v2021.03-3-ge334769_2021-03-29/ (although it could be because I didn't trigger it)

#### Steps to reproduce
I trigger the issue by using the dictionary. How many words you have to look up varies.
Before writing this bug report I did the following test:
I started every time from a powered off device.
And then I longtapped the same word ('ulcerous') from the same book over and over again until it was triggered.
Like this:
1. Long tap on 'ulcerous'
2. Tap once 'page forward' in the dictionary to check if the issue has been triggered
3. close dictionary window
Repeat 1... and so on

I did the test 4 times. The result was: the issue was triggered after 1 attempt, 13 attempts, 4 attempts, +50 attempts

verbose debug log: [crash1.log](https://github.com/koreader/koreader/files/6235266/crash1.log)

NiLuJe:
Did you enable double-taps at one point, and if so, what's your double-tap interval set to?

gerhaher:
> New question: what's your hold interval set to?

Default value (500)

NiLuJe:
Does things behave better if you disable the BackgroundRunner plugin?

gerhaher:
No, it happened again.

NiLuJe:
Note to self: I'm *guessing* nothing clears the slot in whatever weird protocol the KA1 is using, which means the hold detection always passes.

Have to run for a bit, but I'll get you something to test later tonight (e.g., have the hold detection actually re-check the delta instead of relying on a previous clear).

NiLuJe:
[...]

PR #7478, opened 2021-03-31
Input: Followup tweaks to #7415

### LINKED ISSUE #7472: Bq Cervantes 4 : koreader doesn't start after updating to the latest dev. version
Hi,
Yesterday evening, I updated my bq Cervantes 4 to the latest developper version of Koreader (I cannot remember the version number but I think it ended with a 6).
The update operation went fine but when Koreader restarted it entered into an infinite loop. The little dots scroll endlessly at the bottom of the screen but nothing happens. 
I restarted my device but it didn't solve the problem.
I also tried to connect to the device by Telnet in order to downgrade koreader manually. I followed the instructions found here : http://web.archive.org/web/20200720044821/https://blog.bq.com/es/bq-ereaders-developers-program/. I used Putty, chose Telnet, entered 192.168.4.1 as IP address and 23 as Port, but the connection was refused. I'm not sure that I proceeded correctly (I don't understand for example "Set your computer usb0 interface with another ip address in that range and you would be able to telnet it through USB interface".
So right now I'm stucked and I don't know how to solve the problem. 
Did I miss something ? Or can I force the device to launch the bq software instead of koreader ? Or is there another way than Telnet to access to the device ?
Any help highly appreciated :-)
Bruno

NiLuJe:
* Actually load librt properly on old Linux systems (fix #7472)
* Made sure SDL & Android honored timeouts properly in some edgy (or not so edgy, in SDL's case) cases (c.f., https://github.com/koreader/koreader-base/pull/1343).
* Try harder not to consume a timer callback when the timerfd backend is in use and `select` timed out *not* because of a timerfd expiring. This would take some particularly heinous timing to reproduce (e.g., a timerfd set, but a task deadline expiring *before* it). May or may not help with #7473.
* Decode `BTN_` KEY codes in verbose debug logs.
* Fix MultiTouch on the H2O (GestureDetector was hard-coding the two slots as 0 & 1, but the H2O kernel is a special snowflake that doesn't use slot 0, instead switching to slot 1 & 2).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7478)

### DIFF
--- a/frontend/device/gesturedetector.lua
+++ b/frontend/device/gesturedetector.lua
@@ -205,18 +205,21 @@ function GestureDetector:isHold(t1, t2)
 end
 
 function GestureDetector:isTwoFingerTap()
-    if self.last_tevs[0] == nil or self.last_tevs[1] == nil then
+    local s1 = self.input.main_finger_slot
+    local s2 = self.input.main_finger_slot + 1
+
+    if self.last_tevs[s1] == nil or self.last_tevs[1] == nil then
         return false
     end
-    local x_diff0 = math.abs(self.last_tevs[0].x - self.first_tevs[0].x)
-    local x_diff1 = math.abs(self.last_tevs[1].x - self.first_tevs[1].x)
-    local y_diff0 = math.abs(self.last_tevs[0].y - self.first_tevs[0].y)
-    local y_diff1 = math.abs(self.last_tevs[1].y - self.first_tevs[1].y)
-    local tv_diff0 = self.last_tevs[0].timev - self.first_tevs[0].timev
+    local x_diff0 = math.abs(self.last_tevs[s1].x - self.first_tevs[s1].x)
+    local x_diff1 = math.abs(self.last_tevs[s2].x - self.first_tevs[s2].x)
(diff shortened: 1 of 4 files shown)


NiLuJe:
Requires https://github.com/koreader/koreader-base/pull/1342

poire-z (approved):
(Well, I reviewed what I could: I just followed the 0/1 > s1/s2 changes, they're fine :)

NiLuJe:
Something isn't, as it broke scheduled tasks on the emu ;).

Frenzie (commented):
whoa whoa whoa I'll need to read this in more detail later

NiLuJe:
Quick test-case for the SDL timeout thing: Move to a FM page w/ thumbnails to generate, and keep moving your mouse around inside the KOReader window.

It won't ever refresh until you stop moving, because SDL keeps spitting a single event that we weren't doing anything with, and never exiting the loop.

I'd *vaguely* noticed something weird originally (which is what made me dig inside runInSubProcess), and I knew that the weird forever loop was weird, but I hadn't quite made the link until last night, where it happened much more reliably because of a gamepad stick being jittery ;).

PR #7490, opened 2021-04-02
Font menu: open on page with currently selected font, and other tweaks

poire-z:
#### Font menu: open on page with currently selected font

Generic feature added to TouchMenu: an optional callback 'open_on_menu_item_id_func()' set on an items table can tell which menu item should be shown when opening a menu.
TouchMenu: allows jumping to first or last page with long-press on the chevrons.
Proposals and discussion at at https://github.com/koreader/koreader/issues/3620#issuecomment-801421408 and https://github.com/koreader/koreader/issues/3620#issuecomment-801516104.
I haven't added it for the other menu suggested there (Languages, Dispatcher actions...) because these are less bothering _to me_ :) Tell me if I should.

I also included something that I proposed previously (can't find where) - because I've been happy having it in my own patch: showing the current font name in the `Font>` menu name itself:
<kbd>image</kbd>
It's now less needed, with the above feature that allows seeing what font is selected by just opening the menu.
Also, it might not be the font currently shown (if the book used embedded fonts), so it might not be needed to really show it that top level.
But it fits nicely with its neighour `Typography language>` - although this one shows rightly some real information about the book (typography from metadata lang= tag) so a user better see it if it's wrong - while Font> will show what the user has selected.
Well, I'm fine with removing it, just tell me :)

#### Notifications: add some top margin

#7226 / 68a5fcdb moved them a bit too near the top than before (but the  before computation was crap: centered in a Screen:getHeight()/10), and it can overlap with CRE rendering progress bar.
Also makes it nicer when multiple notications are stacked.
<kbd>image</kbd>

#### TOC: highlight first chapter on page instead of last

So its in sync with the chapter shown in the footer, which was previously fixed with #7204 / a482baac.

#### Style tweaks: remove "Toggle alternative EPUB content"

With MathML support, this is no longer all or nothing.  As it's quite rare and technical, let this be handled via user style tweak when needed. https://github.com/koreader/koreader/pull/7465#issuecomment-810230767

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7490)

### DIFF
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -96,8 +96,12 @@ function ReaderFont:init()
             end,
             checked_func = function()
                 return v == self.font_face
-            end
+            end,
+            menu_item_id = v,
         })
+    end
+    self.face_table.open_on_menu_item_id_func = function()
+        return self.font_face
     end
     self.ui.menu:registerToMainMenu(self)
 end
@@ -332,7 +336,9 @@ function ReaderFont:addToMainMenu(menu_items)
     self.face_table.max_per_page = 5
     -- insert table to main reader menu
     menu_items.change_font = {
-        text = self.font_menu_title,
+        text_func = function()
+            return T(_("Font: %1"), BD.wrap(self.font_face))
+        end,
         sub_item_table = self.face_table,
     }
 end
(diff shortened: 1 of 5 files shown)


Frenzie (approved):
:+1:

hius07:
> which menu item should be shown when opening a menu.

In the Dispatcher menu the most bothering thing is that when we go down to the certain action's submenu and then go up, we are on the first page.

poire-z:
We could just open on the first selection action, but there can be multiple actions, so dunno.
I might let that to @yparitcher, using what I added to TouchMenu, this dispatcher/action stuff feels like quite a bit more dynamic that my font menu :)

Issue #6399, opened 2020-07-19
FR: Calibre plugin : sync-ing back Date Read & Rating


didierm:
I am not entirely sure what to make of https://github.com/koreader/koreader/pull/6177#issuecomment-645918265 , but a feature I use a lot in Calibre Companion (Android app : https://play.google.com/store/apps/details?id=com.multipie.calibreandroid), is sync'ing back the 'Read' and 'Date Read' information to Calibre.
For this purpose, CC allows you to define the Calibre 'Read' and 'Date Read' columns in its settings (in my case : '#read' and '#date_read').

As KR already logs this information (at the end of the document : [Mark as Read] and/or [Book Status]='Finished'), I was wondering whether the Calibre plugin would allow for an analogous setup.
Sync'ing back the book rating (end of document > [Book Status] > rating) would be so nice too.

pazos:
I'm afraid I cannot help here. The main limitation I reached with the current implementation is the lack of usage/understanding of calibre. I just use calibre to strip DRM of the books I buy online and to remove some corporative BS. I don't keep any specific metadata beyond series & tags or use calibre as a book manager. Most of my documents are managed outside calibre.

I think it is doable and certainly can assist (a bit) on the smart device app interface, but have no interest in doing this. PRs are very welcome.

If somebody wants to contribute keep in mind:

- we load the whole JSON database in memory, so we strip the things we don't use, which are almost all of them (except the things used to navigate/search metadata)
- AFAICT the userdata field can grow a lot, it would be a good idea to test the implementation on a big library (1000-10000+ books) with a big userdata for each entry. I think it is not a good idea to store the whole userdata, just the bits we might want.

The most difficult part is figuring out a way to store the relevant data of fields that are, in most cases, mutable. Once some hardcore calibre user does that sending the info back to calibre shouldn't be a problem.

NiLuJe:
The thread right below the comment you linked kind of show that there's approximately seven billion ways to have a "Read" column in Calibre, which, in addition to the technical quirks mentioned by @pazos, make this hard to get right. (You'd basically have to add a new setting, asking for the right custom column name, and quite possibly its type).

NiLuJe:
Also: which way do you sync? When? For which books? What happens when there's a conflict?

[...]

PR #7492, opened 2021-04-03
Calibre - send files back to calibre for sync

snelg:
Pretty much exactly what @pazos suggested in https://github.com/koreader/koreader/issues/6399#issuecomment-725622329
Also submitting a corresponding tiny patch to @harmtemolder 's Calibre plugin.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7492)

### DIFF
--- a/plugins/calibre.koplugin/wireless.lua
+++ b/plugins/calibre.koplugin/wireless.lua
@@ -662,11 +662,10 @@ function CalibreWireless:sendToCalibre(arg)
 
     self:sendJsonData("OK", { fileLength = file_size })
 
-    local chunk_size = 1024
-    local data = file:read(chunk_size)
-    while data ~= nil do
+    while true do
+        local data = file:read(4096)
+        if not data then break end
         self.calibre_socket:send(data)
-        data = file:read(chunk_size)
     end
 
     file:close()
--- a/plugins/calibre.koplugin/wireless.lua
+++ b/plugins/calibre.koplugin/wireless.lua
@@ -644,9 +644,20 @@ end
 
 function CalibreWireless:sendToCalibre(arg)
     logger.dbg("GET_BOOK_FILE_SEGMENT", arg)
-    -- not implemented yet, we just send an invalid opcode to raise a control error in calibre.
-    -- If we don't do this calibre will wait *a lot* for the file(s)
-    self:sendJsonData('NOOP', {})
+    local inbox_dir = G_reader_settings:readSetting("inbox_dir")
+    local path = inbox_dir .. "/" .. arg.lpath
+
+    local file = io.open(path, "rb")
+    if file == nil then
+        self:sendJsonData('NOOP', {})
+        return
+    end
+
+    local data = file:read("*all")
+    file:close()
+
+    self:sendJsonData('OK', { fileLength = #data })
+    self.calibre_socket:send(data)
 end
 
 function CalibreWireless:isCalibreAtLeast(x,y,z)
--- a/plugins/calibre.koplugin/wireless.lua
+++ b/plugins/calibre.koplugin/wireless.lua
@@ -648,7 +648,7 @@ function CalibreWireless:sendToCalibre(arg)
     local path = inbox_dir .. "/" .. arg.lpath
 
     local file = io.open(path, "rb")
-    if file == nil then
+    if not file then
         self:sendJsonData('NOOP', {})
         return
     end
--- a/plugins/calibre.koplugin/wireless.lua
+++ b/plugins/calibre.koplugin/wireless.lua
@@ -13,6 +13,7 @@ local InputDialog = require("ui/widget/inputdialog")
 local InfoMessage = require("ui/widget/infomessage")
 local NetworkMgr = require("ui/network/manager")
 local UIManager = require("ui/uimanager")
+local lfs = require("libs/libkoreader-lfs")
 local logger = require("logger")
 local rapidjson = require("rapidjson")
 local sha = require("ffi/sha2")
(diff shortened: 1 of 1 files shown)


plugins/calibre.koplugin/wireless.lua
```diff
@@ -644,9 +644,20 @@ end
...
     logger.dbg("GET_BOOK_FILE_SEGMENT", arg)
-    -- not implemented yet, we just send an invalid opcode to raise a control error in calibre.
-    -- If we don't do this calibre will wait *a lot* for the file(s)
-    self:sendJsonData('NOOP', {})
+    local inbox_dir = G_reader_settings:readSetting("inbox_dir")
+    local path = inbox_dir .. "/" .. arg.lpath
+
+    local file = io.open(path, "rb")
+    if file == nil then
+        self:sendJsonData('NOOP', {})
```
Frenzie, line 652: nit

plugins/calibre.koplugin/wireless.lua
```diff
@@ -644,9 +644,20 @@ end
...
 function CalibreWireless:sendToCalibre(arg)
     logger.dbg("GET_BOOK_FILE_SEGMENT", arg)
-    -- not implemented yet, we just send an invalid opcode to raise a control error in calibre.
-    -- If we don't do this calibre will wait *a lot* for the file(s)
-    self:sendJsonData('NOOP', {})
+    local inbox_dir = G_reader_settings:readSetting("inbox_dir")
+    local path = inbox_dir .. "/" .. arg.lpath
+
+    local file = io.open(path, "rb")
+    if file == nil then
```
Frenzie, line 651: or if not file

plugins/calibre.koplugin/wireless.lua
```diff
@@ -644,9 +644,20 @@ end
...
+        self:sendJsonData('NOOP', {})
+        return
+    end
+
+    local data = file:read("*all")
+    file:close()
+
+    self:sendJsonData('OK', { fileLength = #data })
```
Frenzie, line 659: nit

plugins/calibre.koplugin/wireless.lua
```diff
@@ -644,9 +644,20 @@ end
...
+    local inbox_dir = G_reader_settings:readSetting("inbox_dir")
+    local path = inbox_dir .. "/" .. arg.lpath
+
+    local file = io.open(path, "rb")
+    if file == nil then
+        self:sendJsonData('NOOP', {})
+        return
+    end
+
+    local data = file:read("*all")
```
pazos, line 656: Is it possible to split the binary in chunks of a few KB/MB?. That way the available memory isn't a hard limit. We might want to prevent OOM errors as 1GB files are not that rare.

When sending files from calibre to koreader we already do that. See https://github.com/koreader/koreader/blob/master/plugins/calibre.koplugin/wireless.lua#L540-L579.

I guess it works the same way here.

ie: use a buffer of 4096. Read the whole file to know its size and store it in a `to_write_bytes` local variable. Then send the `ok, file size` .  Then use file:seek to move between offsets at the file and send chunks until `to_write_bytes` is equal to 0.

I'm not sure how calibre deals with it, but I'm guessing it receives the bytes count first and reads from the socket until that number of bytes is reached.

plugins/calibre.koplugin/wireless.lua
pazos: :+1: for `if not file`. Sounds more idiomatic.

plugins/calibre.koplugin/wireless.lua
snelg: Correct, calibre receives the byte count and then tries to read that number of bytes from the socket.

You're right, we should chunk the data. As usual, I was only thinking of the scope of my own problem.
I was trying to deal with the "sync stats" plugin, which only transfers the `metadata.[filetype].lua` files back to calibre. I hope that 1GB `.lua` files are rare ;)

plugins/calibre.koplugin/wireless.lua
snelg: Sorry, I'm not a lua expert yet. What's the nit?
Is it "use double-quote by default instead of single-quote"? Or am I not using some idiomatic syntax for `self` function calls?

Frenzie: Not Lua related, but yes, our style is double quotes.

snelg: Done

plugins/calibre.koplugin/wireless.lua
snelg: Simple chunking added

plugins/calibre.koplugin/wireless.lua
```diff
@@ -644,9 +645,31 @@ end
...
+    if not file then
+        self:sendJsonData("NOOP", {})
+        return
+    end
+
+    self:sendJsonData("OK", { fileLength = file_size })
+
+    local chunk_size = 1024
+    local data = file:read(chunk_size)
+    while data ~= nil do
```
Frenzie, line 667: Should work this way, shouldn't it? It's not 100% the same in the sense that something could be `false` but not `nil` for example, but it's just less visual noise when the difference doesn't matter.
```suggestion
    while data do
```

plugins/calibre.koplugin/wireless.lua
```diff
@@ -644,9 +645,31 @@ end
...
+
+    local file = io.open(path, "rb")
+    if not file then
+        self:sendJsonData("NOOP", {})
+        return
+    end
+
+    self:sendJsonData("OK", { fileLength = file_size })
+
+    local chunk_size = 1024
```
Frenzie, line 665: Would that be a touch small compared to 4K?

NiLuJe: Yeah, +1 for page-sized buffers (i.e., 4K ^^).

plugins/calibre.koplugin/wireless.lua
snelg: Slightly restructured to follow the literal textbook example at https://www.lua.org/pil/21.2.2.html

Frenzie (approved):
lgtm

harmtemolder:
@snelg Thanks for the patch. I’ve released it as [v0.1.3-alpha](https://git.sr.ht/~harmtemolder/koreader-calibre-plugin/tree/main/item/releases/KOReader%20Sync%20v0.1.3-alpha.zip)

shermp:
Purely out of curiosity, since I haven't actually tested this, is Calibre happy with you not sending the following fields?
```
willStream
willStreamBinary
```

Answers own question: yeah, looking at the code, current Calibre no longer seems to care, don't know if it ever did. May have been a transitional thing back in the early days of the protocol?

pazos (approved):
lgtm

harmtemolder:
@snelg Just updated KOReader on my Kobo and synced its metadata wirelessly with calibre. Thanks!

PR #7496, opened 2021-04-03
ScreenSaver: Get rid of automagic settings

NiLuJe:
By which I mean, explicitly initialize default settings, making everything seven hundred billion percent less weird.

Also, split the background option between image and (solo) message, because the defaults for those are different, and trying to twist the one into the other was leading to pure madness.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7496)

### DIFF
--- a/frontend/ui/elements/screensaver_menu.lua
+++ b/frontend/ui/elements/screensaver_menu.lua
@@ -123,130 +123,140 @@ return {
                 end,
             },
             {
-                text = _("Black background behind images"),
-                enabled_func = function()
-                    return G_reader_settings:readSetting("screensaver_type") ~= "disable"
-                end,
-                checked_func = function()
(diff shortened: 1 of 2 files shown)


NiLuJe:
xref: https://www.mobileread.com/forums/showpost.php?p=4108834&postcount=553

Frenzie (approved):
Sounds good and nothing jumps out as suspicious.

poire-z:
(Using my joker to skip reviewing screensaver.lua!)

NiLuJe (commented):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7496)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

Issue #7476, opened 2021-03-31
FR: Setting to control highlighting brightness


gbyl:
#### Current Behavior

Under the highlighting menu, there are currently three options: Lighten, Underline and Invert. Lighten seems to apply a 50% gray.

#### Request

Some devices have a [darker than ideal](https://www.youtube.com/watch?v=jm-YBQMj8Jw) background color. An option to set highlighting to a brighter gray would improve contrast/readability in such devices.

gbyl:
It seems that the highlighting brightness is controlled by the following value:

https://github.com/koreader/koreader/blob/03e9fac15617e63f99dab71fd1ac0d36acf2e148/frontend/apps/reader/modules/readerview.lua#L40

As evidenced by:

https://github.com/koreader/koreader/blob/03e9fac15617e63f99dab71fd1ac0d36acf2e148/frontend/apps/reader/modules/readerview.lua#L547-L548

I tried many different `lighten_factor` values with no visible results. Do we know what range of values are allowed?

poire-z:
> I tried many different lighten_factor values with no visible results. Do we know what range of values are allowed?

Just tried it (changeing the 0.2 on line 40 in readerview), with PDF or EPUB, and it worked - with 0.1 (very light) and 0.5 (darker).
It's only the final highlight that gets this value - while you're making one, and still when the popup is shown, it's still the engine that is doing the selected text highlight (some value for EPUB/crengine, black for PDF). Once it's done, and when visiting pages, only then it's readerview that does it.

gbyl:
> and still when the popup is shown, it's still the engine that is doing the selected text highlight

Ah you are so right! I was not aware of that

> It's only the final highlight that gets this value

Is there a way to have it change the factor for past highlights?

And more importantly, is it possible/reasonable to have a setting to be able to control the `lighten_factor`?

poire-z:
> Is there a way to have it change the factor for past highlights?

The current value for this lighten_factor applies to the drawing of past highlights, for me, tested on EPUBs.

> is it possible/reasonable to have a setting to be able to control the lighten_factor?

I suppose so, one just have to do it.
- Either a simple `lighten_factor = G_reader_settings:readSetting("highlight_lighten_factor") or 0.2,` and interested people can add manually `"highlight_lighten_factor" = 0.1` to their settings.reader.lua. (But we got rid of most of these hidden settings, to make them available in some menu).
- Or by adding some menu + number picker, but that's more work.
[...]

PR #7497, opened 2021-04-03
hidden setting: highlight lighten factor

gbyl:
Adds a new hidden setting to address https://github.com/koreader/koreader/issues/7476

The setting takes values from 0 (brightest) to 1 (darkest). For example, to lighten the highlight, users can manually add the line below to their `settings.reader.lua`

```
"highlight_lighten_factor" = 0.1
```

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7497)

### DIFF
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -37,7 +37,7 @@ local ReaderView = OverlapGroup:extend{
     outer_page_color = Blitbuffer.gray(DOUTER_PAGE_COLOR / 15),
     -- highlight with "lighten" or "underscore" or "invert"
     highlight = {
-        lighten_factor = 0.2,
+        lighten_factor = G_reader_settings:readSetting("highlight_lighten_factor") or 0.2,
         temp_drawer = "invert",
         temp = {},
         saved_drawer = "lighten",
--- a/frontend/apps/reader/modules/readerview.lua
+++ b/frontend/apps/reader/modules/readerview.lua
@@ -37,7 +37,7 @@ local ReaderView = OverlapGroup:extend{
     outer_page_color = Blitbuffer.gray(DOUTER_PAGE_COLOR / 15),
     -- highlight with "lighten" or "underscore" or "invert"
     highlight = {
-        lighten_factor = G_reader_settings:readSetting("highlight_lighten_factor") or 0.2,
+        lighten_factor = G_reader_settings:readSetting("highlight_lighten_factor", 0.2),
         temp_drawer = "invert",
         temp = {},
         saved_drawer = "lighten",


poire-z:
And now, as we'd like to avoid hidden settings, you (or @NiLuJe, as he incited you to go with a hidden setting) will have to make a menu for it :)
(Not forcing you, except if you have time to spend digging deeper in KOReader, but that would be in `ReaderHighlight:genHighlightDrawerMenu()`, but there's no simple sample to just cut & paste in there - would have to be picked from elsewhere using a SpinWidget.)

frontend/apps/reader/modules/readerview.lua
```diff
@@ -37,7 +37,7 @@ local ReaderView = OverlapGroup:extend{
     outer_page_color = Blitbuffer.gray(DOUTER_PAGE_COLOR / 15),
     -- highlight with "lighten" or "underscore" or "invert"
     highlight = {
-        lighten_factor = 0.2,
+        lighten_factor = G_reader_settings:readSetting("highlight_lighten_factor") or 0.2,
```
Frenzie, line 40: http://koreader.rocks/doc/modules/luasettings.html#LuaSettings:readSetting
```suggestion
        lighten_factor = G_reader_settings:readSetting("highlight_lighten_factor", 0.2),
```

Frenzie:
>  as we'd like to avoid hidden settings

+1

gbyl:
I agree that we should avoid hidden settings, but I simply do not have the skills (yet) to contribute a proper menu. For this specific feature this may be just fine since `0.1` is about the right value for darker devices. At least for RM1 it is.

poire-z:
While just adding `DHIGHLIGHT_LIGHTEN_FACTOR = 0.2`  in defaults.lua would have been just as good, with already some UI (Advanced settings) to document its existence and allow setting it - and no need to update the burried https://github.com/koreader/koreader/issues/2589#issuecomment-323927682 :)

NiLuJe:
I mean, globals are evil, and while hidden settings are suboptimal, I still view them as a lesser evil (despite the Advanced Settings UI) ;).

I'll plug it to the UI later tonight.

NiLuJe:
c.f., #7505.

Sidebar: should this (and the other page_gap_color hidden setting) be read at `:init` instead?

Frenzie:
Implemented in https://github.com/koreader/koreader/commit/e4adcdf4229f9bbeea18bc260db2ab56115f5c39. Thanks for the proposal. ^_^

poire-z:
I think @NiLuJe PR needs this PR: this PR loads the setting, e4adcdf only allows modifying it (and applied), but it won't load it on re-openings.

Frenzie:
Oops, my bad. I thought it was the same thing with some additions.

PR #7501, opened 2021-04-04
Search dialogs: correct response when nothing entered

hius07:
Do not show "No results" notification when nothing entered to search for.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7501)

### DIFF
--- a/frontend/apps/filemanager/filemanagerfilesearcher.lua
+++ b/frontend/apps/filemanager/filemanagerfilesearcher.lua
@@ -23,10 +23,6 @@ local FileSearcher = InputContainer:new{
     results = {},
     items = 0,
     commands = nil,
-
-    --filemanagersearch
-    use_previous_search_results = false,
-    lastsearch = nil,
 }
 
 function FileSearcher:readDir()
@@ -74,28 +70,22 @@ function FileSearcher:setSearchResults()
 end
 
 function FileSearcher:close()
-    if self.search_value then
         UIManager:close(self.search_dialog)
-        if string.len(self.search_value) > 0 then
-            self:readDir() --- @todo this probably doesn't need to be repeated once it's been done
-            self:setSearchResults() --- @todo doesn't have to be repeated if the search term is the same
-            if #self.results > 0 then
-                self:showSearchResults() --- @todo something about no results
-            else
-                UIManager:show(
-                    InfoMessage:new{
-                        text = BaseUtil.template(_("No results for '%1'"),
-                                             self.search_value)
-                    }
-                )
+        self:readDir() --- @todo this probably doesn't need to be repeated once it's been done
+        self:setSearchResults() --- @todo doesn't have to be repeated if the search term is the same
+        if #self.results > 0 then
+            self:showSearchResults() --- @todo something about no results
+        else
+            UIManager:show(
+                InfoMessage:new{
+                    text = BaseUtil.template(_("No results for '%1'."),
+                                         self.search_value)
+                }
+            )
             end
-        end
-    end
 end
 
 function FileSearcher:onShowFileSearch()
-    local dummy = self.search_value
-    local enabled_search_home_dir = G_reader_settings:has("home_dir")
     self.search_dialog = InputDialog:new{
         title = _("Enter filename to search for"),
         input = self.search_value,
@@ -111,30 +101,23 @@ function FileSearcher:onShowFileSearch()
                     end,
                 },
                 {
-                    text = _("Current folder"),
-                    enabled = true,
+                    text = _("Home folder"),
+                    enabled = G_reader_settings:has("home_dir"),
                     callback = function()
-                        self.path = self.ui.file_chooser and self.ui.file_chooser.path or self.ui:getLastDirFile()
                         self.search_value = self.search_dialog:getInputText()
-                        if self.search_value == dummy then -- probably DELETE this if/else block
-                            self.use_previous_search_results = true
-                        else
-                            self.use_previous_search_results = false
-                        end
+                        if self.search_value == "" then return end
+                        self.path = G_reader_settings:readSetting("home_dir")
                         self:close()
                     end,
                 },
                 {
-                    text = _("Home folder"),
-                    enabled = enabled_search_home_dir,
+                    text = _("Current folder"),
+                    enabled = true,
+                    is_enter_default = true,
                     callback = function()
-                        self.path = G_reader_settings:readSetting("home_dir")
                         self.search_value = self.search_dialog:getInputText()
-                        if self.search_value == dummy then -- probably DELETE this if/else block
-                            self.use_previous_search_results = true
-                        else
-                            self.use_previous_search_results = false
-                        end
+                        if self.search_value == "" then return end
+                        self.path = self.ui.file_chooser and self.ui.file_chooser.path or self.ui:getLastDirFile()
                         self:close()
                     end,
                 },
--- a/frontend/apps/filemanager/filemanagerfilesearcher.lua
+++ b/frontend/apps/filemanager/filemanagerfilesearcher.lua
@@ -70,19 +70,19 @@ function FileSearcher:setSearchResults()
 end
 
 function FileSearcher:close()
-        UIManager:close(self.search_dialog)
-        self:readDir() --- @todo this probably doesn't need to be repeated once it's been done
-        self:setSearchResults() --- @todo doesn't have to be repeated if the search term is the same
-        if #self.results > 0 then
-            self:showSearchResults() --- @todo something about no results
-        else
-            UIManager:show(
-                InfoMessage:new{
-                    text = BaseUtil.template(_("No results for '%1'."),
-                                         self.search_value)
-                }
-            )
-            end
+    UIManager:close(self.search_dialog)
+    self:readDir() --- @todo this probably doesn't need to be repeated once it's been done
+    self:setSearchResults() --- @todo doesn't have to be repeated if the search term is the same
+    if #self.results > 0 then
+        self:showSearchResults()
+    else
+        UIManager:show(
+            InfoMessage:new{
+                text = BaseUtil.template(_("No results for '%1'."),
+                                     self.search_value)
+            }
+        )
+        end
 end
 
 function FileSearcher:onShowFileSearch()
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -50,6 +50,7 @@ function ReaderSearch:onShowFulltextSearchInput()
                 {
                     text = backward_text,
                     callback = function()
+                        if self.input_dialog:getInputText() == "" then return end
                         UIManager:close(self.input_dialog)
                         self:onShowSearchDialog(self.input_dialog:getInputText(), 1)
                     end,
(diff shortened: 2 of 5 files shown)


frontend/apps/reader/modules/readersearch.lua
```diff
@@ -70,6 +70,7 @@ function ReaderSearch:onShowFulltextSearchInput()
 end
 
 function ReaderSearch:onShowSearchDialog(text, direction)
+    if text == '' then return end
```
Frenzie, line 73:
```suggestion
    if text == "" then return end
```

However, it should be in the callback, not here.

Frenzie:
```diff
diff --git a/frontend/apps/reader/modules/readersearch.lua b/frontend/apps/reader/modules/readersearch.lua
index 49d11041..ce81899b 100644
--- a/frontend/apps/reader/modules/readersearch.lua
+++ b/frontend/apps/reader/modules/readersearch.lua
@@ -50,6 +50,7 @@ function ReaderSearch:onShowFulltextSearchInput()
                 {
                     text = backward_text,
                     callback = function()
+                        if self.input_dialog:getInputText() == "" then return end
                         UIManager:close(self.input_dialog)
                         self:onShowSearchDialog(self.input_dialog:getInputText(), 1)
                     end,
@@ -58,6 +59,7 @@ function ReaderSearch:onShowFulltextSearchInput()
                     text = forward_text,
                     is_enter_default = true,
                     callback = function()
+                        if self.input_dialog:getInputText() == "" then return end
                         UIManager:close(self.input_dialog)
                         self:onShowSearchDialog(self.input_dialog:getInputText(), 0)
                     end,
```

Something like this. Or maybe as an enabled_func of some sort but that'd be slightly harder to add.

hius07: Can you please point me to the reasons, why two conditions is better than one?

poire-z: The above patch won't close the input widget if it's empty and the user hits < or >. The user will have to hit Cancel to get out.
With your PR, it closes and no search is performed.
I think I'd prefer the above patch, which makes things more obvious - but I haven't checked how we do that in other input dialogs of this kind.

Frenzie: As @poire-z said, two conditions isn't better than one but not closing the dialog is better than closing the dialog.

Frenzie: That is, I was expressing a rough sketch of the desired behavior in the form of a patch, the actual patch could take a completely different form.

hius07: > With your PR, it closes and no search is performed.

That is by design. This behaviour is the same as in the File search, Dictionary lookup, Wiki lookup.

Frenzie: That behavior would be a *lack* of design. Apparently I never considered that scenario when I wrote (some of) those dialogs. :-) Closing it as if you clicked cancel is very unexpected.

hius07: > Closing it as if you clicked cancel is very unexpected.

Can it be hardcored in the InputDialog to be applied everywhere?
(No reaction to the ok_button(s) if the input field is empty)

Frenzie: In the sense of making that the default (rather than "hardcoded" as such), in general probably. But this particular specimen is a custom dialog iirc.

poire-z: And there are some other input dialog where entering/saving an empty string might make sense (ie Wikipedia languages, to reset it to its default of just using the book or UI language).
But given from this sample that it's the job of the caller to close or not close the dialog in its callback, the current behaviour of InputDialog not caring might be more flexible (and less brain work :)

hius07: OK, what do you recommend right now?

Currently when pressing ok_button with empty input box
- File search, Dictionary lookup, Wiki lookup just close the input dialog
- Fulltext search tries to search and says No result, leaving search_dialog (arrows) open

hius07:
Proposition for `File search`: if nothing entered into the input field, pressing ok_buttons has no actions at all.

frontend/apps/filemanager/filemanagerfilesearcher.lua
```diff
@@ -74,28 +70,22 @@ function FileSearcher:setSearchResults()
...
-                UIManager:show(
-                    InfoMessage:new{
-                        text = BaseUtil.template(_("No results for '%1'"),
-                                             self.search_value)
-                    }
-                )
+        self:readDir() --- @todo this probably doesn't need to be repeated once it's been done
+        self:setSearchResults() --- @todo doesn't have to be repeated if the search term is the same
+        if #self.results > 0 then
+            self:showSearchResults() --- @todo something about no results
```
Frenzie, line 77: Seems to be my fault, in part anyway. https://github.com/koreader/koreader/commit/92f0092ad4ee7de35014a41f75e2f5b4f9e550d3
```suggestion
            self:showSearchResults()
```

frontend/apps/filemanager/filemanagerfilesearcher.lua
```diff
@@ -74,28 +70,22 @@ function FileSearcher:setSearchResults()
 end
 
 function FileSearcher:close()
-    if self.search_value then
         UIManager:close(self.search_dialog)
```
Frenzie, line 73: Indent should go down one level

frontend/apps/filemanager/filemanagerfilesearcher.lua
```diff
@@ -111,30 +101,23 @@ function FileSearcher:onShowFileSearch()
...
+                    text = _("Home folder"),
+                    enabled = G_reader_settings:has("home_dir"),
                     callback = function()
-                        self.path = self.ui.file_chooser and self.ui.file_chooser.path or self.ui:getLastDirFile()
                         self.search_value = self.search_dialog:getInputText()
-                        if self.search_value == dummy then -- probably DELETE this if/else block
-                            self.use_previous_search_results = true
-                        else
-                            self.use_previous_search_results = false
-                        end
```
Frenzie, line 123: Are you sure about this?

hius07: `use_previous_search_results` is not used anywhere in the repository.
Maybe somebody had a plan to develop something?

hius07:
Not only search dialogs suffer blank enter response issue.
File manager, page selector: press `Go to page` - dialog just closes; press `Go to letter` - dialog closes and we go to the first page.

frontend/apps/filemanager/filemanagerfilesearcher.lua
Frenzie: I guess that someone would be me. >_> Anyway, looks like you're right.

hius07:
Thank you.
I'll go for keyvaluepage and menu.lua with blank input check in page selector in future PRs.

hius07:
Escuse me for probably incorrect question. Can you please clarify the difference in the input code for:
https://github.com/koreader/koreader/blob/aef1e271b5f7de45c400378dddc5e2a73f49f180/frontend/ui/widget/menu.lua#L733
and so compact here
https://github.com/koreader/koreader/blob/aef1e271b5f7de45c400378dddc5e2a73f49f180/frontend/ui/widget/keyvaluepage.lua#L437

poire-z:
Hard job to clarify the various ways to do widget stuff :(
Button being a InputContainer gets a .input_dialog, that in the first case is accessed directly (which is rather ugly).
The second case also uses this (quite obscure) InputContainer feature, but in a more proper way: with a callback in the hold_input sub-component - and the callback get the value entered.

Feels tedious documenting more these bits of ~~crap~~code :)
May be tell us what you want to do, and we may quickly give you something that achieves this without having to think more about all that.

hius07:
Thank you very much, it is just for my understanding.
I am a little bit confused reading the code: similar dialogs or windows are coded in different ways and it is hard to find out which is correct.

poire-z:
Well, that's doomed to have happened: [give a keyboard](https://en.wikipedia.org/wiki/Infinite_monkey_theorem) to [162 international monkeys](https://github.com/koreader/koreader/graphs/contributors) , wait 5/10 years, and you get KOReader :)
But this gives a bit more justification for some of us playing gorillas and being insistant/harassing when reviewing PRs.

Frenzie:
> Escuse me for probably incorrect question. Can you please clarify the difference in the input code for:

The second is nicer/more standardized/less hackish but whether it's Just Old or something we let through more recently you do have to add support in the specific widget for it.

> I am a little bit confused reading the code: similar dialogs or windows are coded in different ways and it is hard to find out which is correct.

In case of doubt, also don't forget the manual (or the code comments on top of the widget implementations, which is what it's generated from). For example:

http://koreader.rocks/doc/modules/ui.widget.confirmbox.html
http://koreader.rocks/doc/modules/ui.widget.inputdialog.html

> But this gives a bit more justification for some of us playing gorillas and being insistant/harassing when reviewing PRs.

Yes, pretty much everything I wanted to say something about but decided not to back around 2017 has come back to bite me since. ;-)

hius07:
> May be tell us what you want to do

I want
https://github.com/koreader/koreader/blob/aef1e271b5f7de45c400378dddc5e2a73f49f180/frontend/ui/widget/keyvaluepage.lua#L436-L441
not to close the input dialog when pressing ok_button with blank input field.
I cannot do this because of
https://github.com/koreader/koreader/blob/5b996c8ab76ff4d4875b45695627596876017b68/frontend/ui/widget/container/inputcontainer.lua#L291

It is rather simple for
https://github.com/koreader/koreader/blob/aef1e271b5f7de45c400378dddc5e2a73f49f180/frontend/ui/widget/menu.lua#L732-L738
but you say it is ugly.

poire-z:
Then, just add your checks and not calling the callback and not closing if empty in:
https://github.com/koreader/koreader/blob/5b996c8ab76ff4d4875b45695627596876017b68/frontend/ui/widget/container/inputcontainer.lua#L289-L292

This being a generic feature of anything that extends InputContainer, it will then apply to all these things.

hius07:
But you said https://github.com/koreader/koreader/pull/7501#discussion_r606771433

poire-z:
Oh, right.
Then make that behaviour toggable with a property:
In kvpage (and other places as needed):
```diff
        hold_input = {
            title = _("Enter page number"),
            type = "number",
            hint_func = function()
                return "(" .. "1 - " .. self.pages .. ")"
            end,
+           wants_non_empty = true, -- or a better name
            callback = function(input)
```

In inputdialog:
```diff
                    callback = function()
+                       if input.wants_non_empty and self.input_dialog:getInputText() == "" then
+                           return
+                       end
                        input.callback(self.input_dialog:getInputText())
                        self:closeInputDialog()
                    end,
```

hius07:
Thank you, I thought nobody should touch these low level widgets as they've been polished already. I'll think more over it.

poire-z:
It's not certain they have all been polished - it's just that when touching/refactoring/cleaning them, you risk breaking a lot of stuff (or need to update that lot of stuff when reworking them...)
For this little addition, there shouldn't be any risk, as it's just an addition that won't trigger on existing code.

Issue #6572, opened 2020-08-26
Can't switch zoom mode while Reflow is active


MicaelJarniac:
* KOReader version: v2020.08.1
* Device: [PW4] Kindle Paperwhite 4 (10th Gen)

#### Issue
When reading a PDF, after activating **Reflow**, when trying to switch to **View Mode: continuous**, a warning message explains about the zoom mode, but switching zoom mode isn't possible while **Reflow** is active.

#### Steps to reproduce
1. Open a PDF
1. Make sure **Reflow** is **off*
1. Go to **Switch zoom mode** and select **Zoom to fit page width** (or some other)
1. Enable **Reflow**
1. Set **View Mode** to **page**, if not already
1. Set **View Mode** to **continuous**
1. Read the warning message
1. Try to go to **Switch zoom mode** and notice it's unavailable

It's mostly just an annoyance, as one would have to disable reflow, then switch zoom mode, then enable reflow again.

Frenzie:
@NiLuJe To what extent is this still relevant with your recent changes?

NiLuJe:
Well, I haven't touched the paged/continuous toggle, at least not on purpose, and I'm unfamiliar with the warning itself, but any and all zoom modes are meaningless/broken in Reflow mode, so they are indeed disabled now.

I guess I'll just see what the warning actually say and tweak it?

NiLuJe:
Yep, hiding it when reflow is enabled makes sense now :).

PR #7505, opened 2021-04-04
Two unrelated UX tweaks

NiLuJe:
Re #6572 & #7497

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7505)

### DIFF
--- a/frontend/apps/reader/modules/readerhighlight.lua
+++ b/frontend/apps/reader/modules/readerhighlight.lua
@@ -323,7 +323,7 @@ function ReaderHighlight:genHighlightDrawerMenu()
             callback = function()
                 self.view.highlight.disabled = not self.view.highlight.disabled
             end,
-            hold_callback = function(touchmenu_instance)
+            hold_callback = function()
                 self:toggleDefault()
             end,
             separator = true,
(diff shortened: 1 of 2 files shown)


NiLuJe:
Sidebar: the highlight stuff is weird. The highlight type is baked into the HL themselves, and the lighten factor is only read at loading time (e.g., requires a restart).

poire-z:
> and the lighten factor is only read at loading time (e.g., requires a restart).
> Sidebar: should this (and the other page_gap_color hidden setting) be read at :init instead?

The other callbacks (for highlight style/disable) update directly `self.view.highlight.saved_drawer = style`, so in your callback, you could do the same: 
`self.view.highlight.lighten_factor = spin.value`
and it would then be used without a restart (dunno if it would update the page, or if you'd need a refresh/repaint to have it applied live).

frontend/apps/reader/modules/readerhighlight.lua
```diff
@@ -323,14 +323,45 @@ function ReaderHighlight:genHighlightDrawerMenu()
...
                 self:toggleDefault()
             end,
             separator = true,
         },
         get_highlight_style("lighten"),
         get_highlight_style("underscore"),
         get_highlight_style("invert"),
+        {
+            text_func = function()
+                return T(_("Lightening factor: %1"), G_reader_settings:readSetting("highlight_lighten_factor", 0.2))
```
Frenzie, line 336: confusing name

NiLuJe: True, but haven't really found anything better :/.

NiLuJe: `Highlight contrast`, maybe?

NiLuJe: (It's grayed out in other modes, so it doesn't really need to match/contain "lighten" to make it clear that applies to the ligthen mode).

poire-z: Highlight opacity? (as "the higher the darker")

NiLuJe: Works for me, yeah ;).

Frenzie: Yes, I suppose that'll do.

PR #7507, opened 2021-04-05
Inputtext: don't allow the cursor to move within a hint

hius07:
Before: you can tap a hint and cursor jumps to that position. Not nice, as we haven't started to input yet.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7507)

### DIFF
--- a/frontend/ui/widget/inputtext.lua
+++ b/frontend/ui/widget/inputtext.lua
@@ -120,6 +120,7 @@ if Device:isTouchDevice() or Device:hasDPad() then
             if self.parent.onSwitchFocus then
                 self.parent:onSwitchFocus(self)
             end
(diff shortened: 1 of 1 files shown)


poire-z (approved):

frontend/ui/widget/inputtext.lua
```diff
@@ -120,6 +120,7 @@ if Device:isTouchDevice() or Device:hasDPad() then
             if self.parent.onSwitchFocus then
                 self.parent:onSwitchFocus(self)
             end
+            if #self.charlist == 0 then return end
```
Frenzie, line 123:
```suggestion
            if #self.charlist == 0 then return end -- Avoid cursor moving within a hint.
```

Issue #7508, opened 2021-04-05
Remarkable: "sync" command replaced as of FW 2.6


gbyl:
Running the standard `sync` command (used in `koreader.sh`) no longer does what it was intended to do:

https://github.com/koreader/koreader/blob/17e6049e3004bafbf7897240ba123e1408516222/platform/remarkable/koreader.sh#L53-L54

As of firmware 2.6 the output of `sync` is:

```bash
root@reMarkable:~# sync
Usage: sync [options]
A tool to synchronize files to the reMarkable cloud service.

Options:
  --service                Run the service, ignores basepath and usertoken.
  --no-service             Run sync once and exit. Needs --usertoken and
                           --basepath
  --client                 Connect to the service, asks for a sync. Needs
                           basepath and usertoken.
  --usertoken <usertoken>  The user token.
  --basepath <basepath>    The storage directory.
  -h, --help               Displays help on commandline options.
  --help-all               Displays help including Qt specific options.
```

Can that line be safely deleted from `koreader.sh`? Or should it be replaced by something else?

gbyl:
Ah I think I got it. Running `sync` will run the new `/usr/bin/sync`. The standard sync still exists in `/bin/sync`. Should we just call the latter with a full path?

PR #7509, opened 2021-04-05
remarkable FW 2.6: Set absolute path for 'sync'

gbyl:
Fixes https://github.com/koreader/koreader/issues/7508

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7509)


NiLuJe:
I'd probably prefer `busybox sync`, in case there's further trickery planned down the road ;).

Issue #7506, opened 2021-04-04
CoverImage plugin: Remarkable-specific alternative


gbyl:
As the CoverImage plugin was originally designed with Tolinos in mind, there are a few things that don't work very well with Remarkables:

- The plugin runs automatically on book opening, which can easily double the loading time of files.

- When disabled, on the other hand, the plugin automatically deletes the file generated, which breaks RM1 since it needs a `poweroff.png` file.

- Since the plugin lacks a cache system, it has to generate the image anew every single time.

I don't know how to integrate these changes to the current plugin — a plugin that should perhaps be part of the ScreenSaver menu by @poire-z's [own admission](https://github.com/koreader/koreader/pull/6813#pullrequestreview-514595450). So, I blatantly plagiarized @zwim's work and made a `poweroffimage.koplugin` which addresses the issues listed above while at the same time simplifying the UI to just one item.

_Note: I am sure this is rather sloppy and I have re-invented the wheel in more than one place. So, please share suggestions._

```lua
local Device = require("device")
local Blitbuffer = require("ffi/blitbuffer")
local InfoMessage = require("ui/widget/infomessage")
local UIManager = require("ui/uimanager")
local WidgetContainer = require("ui/widget/container/widgetcontainer")
local RenderImage = require("ui/renderimage")
local ffiutil = require("ffi/util")
local lfs = require("libs/libkoreader-lfs")
local util = require("util")
local _ = require("gettext")

local PowerOffImage = WidgetContainer:new{
    name = "poweroffimage",
    is_doc_only = true,
}

local function getFileName(fullpath)
    local path, name_extension = util.splitFilePathName(fullpath)
    local name, extension = util.splitFileNameSuffix(name_extension)
    return name:gsub(" ","_"):lower()
end

function PowerOffImage:init()
    self.poweroff_image_path = "/usr/share/remarkable/poweroff.png"
    self.poweroff_image_cache = "/home/root/.coverimages/"..getFileName(self.ui.document.file)..".png"
    self.ui.menu:registerToMainMenu(self)
end

function PowerOffImage:createPowerOffImage(doc_settings)
    if not util.pathExists(self.poweroff_image_cache) then
        local poweroff_image = self.ui.document:getCoverPageImage()
        local s_w, s_h = Device.screen:getWidth(), Device.screen:getHeight()
        local i_w, i_h = poweroff_image:getWidth(), poweroff_image:getHeight()
        local scale_factor = math.min(s_w / i_w, s_h / i_h)

[...]

PR #7510, opened 2021-04-05
Add a caching mechanism for CoverImage

zwim:
Caching mechanism for CoverImage.

Suggestions are welcome.

This can solve some issues from https://github.com/koreader/koreader/issues/7506

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7510)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -398,7 +398,12 @@ function Device:isValidPath(path)
     -- the thorough check
     local real_ext_storage = FFIUtil.realpath(android.getExternalStoragePath())
     local real_path = FFIUtil.realpath(path)
-    return real_path:sub(1, #real_ext_storage) == real_ext_storage
+
+    if real_path then
+        return real_path:sub(1, #real_ext_storage) == real_ext_storage
+    else
+        return false
+    end
 end
 
 --swallow all events
@@ -463,7 +468,7 @@ end
 -- todo: Wouldn't we like an android.deviceIdentifier() method, so we can use better default paths?
 function Device:getDefaultCoverPath()
     if android.prop.product == "ntx_6sl" then -- Tolino HD4 and other
-    return android.getExternalStoragePath() .. "/suspend_others.jpg"
+        return android.getExternalStoragePath() .. "/suspend_others.jpg"
     else
         return "cover.jpg"
     end
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -9,7 +9,6 @@ local logger = require("logger")
 local util = require("util")
 local _ = require("gettext")
 local T = FFIUtil.template
-local realpath = FFIUtil.realpath
 
 local function yes() return true end
 local function no() return false end
@@ -397,8 +396,8 @@ function Device:isValidPath(path)
     end
 
     -- the thorough check
-    local real_ext_storage = realpath(android.getExternalStoragePath())
-    local real_path = realpath(path)
+    local real_ext_storage = FFIUtil.realpath(android.getExternalStoragePath())
+    local real_path = FFIUtil.realpath(path)
     return real_path:sub(1, #real_ext_storage) == real_ext_storage
 end
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -470,7 +470,7 @@ function Device:getDefaultCoverPath()
     if android.prop.product == "ntx_6sl" then -- Tolino HD4 and other
         return android.getExternalStoragePath() .. "/suspend_others.jpg"
     else
-        return "cover.jpg"
+        return android.getExternalStoragePath() .. "/cover.jpg"
     end
 end
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -460,6 +460,15 @@ function Device:untar(archive, extract_to)
     return android.untar(archive, extract_to)
 end
 
+-- todo: Wouldn't we like an android.deviceIdentifier() method, so we can use better default paths?
+function Device:getDefaultCoverPath()
+    if android.prop.product == "ntx_6sl" then -- Tolino HD4 and other
+    return android.getExternalStoragePath() .. "/suspend_others.jpg"
+    else
+        return "cover.jpg"
+    end
+end
+
 android.LOGI(string.format("Android %s - %s (API %d) - flavor: %s",
     android.prop.version, getCodename(), Device.firmware_rev, android.prop.flavor))
 
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -9,6 +9,7 @@ local logger = require("logger")
 local util = require("util")
 local _ = require("gettext")
 local T = FFIUtil.template
+local realpath = FFIUtil.realpath
 
 local function yes() return true end
 local function no() return false end
@@ -390,7 +391,15 @@ function Device:canExecuteScript(file)
 end
 
 function Device:isValidPath(path)
-    return android.isPathInsideSandbox(path)
+    -- the fast check
+    if android.isPathInsideSandbox(path) then
+        return true
+    end
+
+    -- the thorough check
+    local real_ext_storage = realpath(android.getExternalStoragePath())
+    local real_path = realpath(path)
+    return real_path:sub(1, #real_ext_storage) == real_ext_storage
 end
 
 --swallow all events
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -263,22 +263,26 @@ function CoverImage:cleanCache()
     logger.dbg("CoverImage: clean - cache size: ".. cache_size_KiB .. " KiB, cached files: " .. cache_count)
 end
 
-function CoverImage:isCacheEnabled()
+function CoverImage:isCacheEnabled(path)
+    if not path then
+        path = self.cover_image_cache_path
+    end
     if Device.isAndroid() then
         return self.cover_image_cache_maxfiles >= 0 and self.cover_image_cache_maxsize >= 0
-            and lfs.attributes(self.cover_image_cache_path, "mode") == "directory"
-            and self.cover_image_cache_path ~= android.getExternalStoragePath() .. "/koreader/cache/"  -- interferes with frontent cache-framework
+            and lfs.attributes(path, "mode") == "directory"
+            and path ~= "/sdcard/koreader/cache/"  -- interferes with frontent cache-framework
+            and realpath(path) ~= realpath(android.getExternalStoragePath() .. "/koreader/cache/")  -- interferes with frontent cache-framework
     else
         return self.cover_image_cache_maxfiles >= 0 and self.cover_image_cache_maxsize >= 0
-            and lfs.attributes(self.cover_image_cache_path, "mode") == "directory"
-            and self.cover_image_cache_path ~= "./cache/" -- interferes with frontent cache-framework; quick and dirty check
-            and self.cover_image_cache_path ~= realpath("./cache/") .. "/" -- interferes with frontent cache-framework; quick and dirty check
+            and lfs.attributes(path, "mode") == "directory"
+            and path ~= "./cache/" -- interferes with frontent cache-framework; quick and dirty check
+            and realpath(path) ~= realpath("./cache/") -- interferes with frontent cache-framework; quick and dirty check
     end
 end
 
 -- callback for choosePathFile()
 function CoverImage:migrateCache(old_path, new_path)
-    if old_path == new_path then
+    if old_path == new_path or not self:isCacheEnabled(new_path) then
         return
     end
     for entry in lfs.dir(old_path) do
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -55,12 +55,17 @@ function CoverImage:init()
     self.cover_image_stretch_limit = G_reader_settings:readSetting("cover_image_stretch_limit") or 8
     self.cover_image_background = G_reader_settings:readSetting("cover_image_background") or "black"
     self.cover_image_fallback_path = G_reader_settings:readSetting("cover_image_fallback_path") or ""
-    self.cover_image_cache_path = G_reader_settings:readSetting("cover_image_cache_path") or DataStorage:getDataDir() .. "/cache/"
+    self.cover_image_cache_path = G_reader_settings:readSetting("cover_image_cache_path") or DataStorage:getDataDir() .. "/data/cover_image.cache/"
     self.cover_image_cache_maxfiles = G_reader_settings:readSetting("cover_image_cache_maxfiles") or 36
     self.cover_image_cache_maxsize = G_reader_settings:readSetting("cover_image_cache_maxsize") or 5 -- MiB
-    self.cover_image_cache_prefix = "CI_CACHE_"
+    self.cover_image_cache_prefix = "cover_"
     self.enabled = G_reader_settings:isTrue("cover_image_enabled")
     self.fallback = G_reader_settings:isTrue("cover_image_fallback")
+
+    if not lfs.attributes(self.cover_image_cache_path, "mode") then
+        lfs.mkdir(self.cover_image_cache_path)
+    end
+
     self.ui.menu:registerToMainMenu(self)
 end
 
@@ -246,6 +251,8 @@ end
 function CoverImage:isCacheEnabled()
     return self.cover_image_cache_maxfiles >= 0 and self.cover_image_cache_maxsize >= 0
         and lfs.attributes(self.cover_image_cache_path, "mode") == "directory"
+        and self.cover_image_cache_path ~= "./cache/" -- interferres with frontent cache-framework; quick and dirty check
+        and self.cover_image_cache_path ~= lfs.currentdir() .. "/cache/"  -- interferres with frontent cache-framework
 end
 
 -- callback for choosePathFile()
@@ -454,6 +461,8 @@ function CoverImage:menu_entry_cache()
             text = _("Cover cache folder"),
             checked_func = function()
                 return lfs.attributes(self.cover_image_cache_path, "mode") == "directory"
+                    and self.cover_image_cache_path ~= "./cache/" -- interferres with frontent cache-framework
+                    and self.cover_image_cache_path ~= lfs.currentdir() .. "/cache/"  -- interferres with frontent cache-framework
             end,
             help_text_func = function()
                 return T(_("Current cache path:\n%1"), self.cover_image_cache_path)
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -392,7 +392,7 @@ You can either choose an existing file:
 
 or specify a new file:
 - First select a directory
-- Then enter a name for the new file]])
+- Then add the name of the new file]])
 
 -- menu entry: Cache settings
 function CoverImage:menu_entry_cache()
@@ -402,23 +402,6 @@ function CoverImage:menu_entry_cache()
             return self:isCacheEnabled()
         end,
         sub_item_table = {
-            {
-            text = _("Cover cache folder"),
-            checked_func = function()
-                return lfs.attributes(self.cover_image_cache_path, "mode") == "directory"
-            end,
-            keep_menu_open = true,
-            callback = function()
-                UIManager:show(ConfirmBox:new{
-                    text = _("Select a cache folder. The contents of the old folder will be migrated."),
-                    ok_text = _("Yes"),
-                    cancel_text = _("No"),
-                    ok_callback = function()
-                        self:choosePathFile("cover_image_cache_path", true, false, self.migrateCache)
-                    end,
-                })
-            end,
-            },
             {
                 text_func = function()
                     local number
@@ -458,6 +441,24 @@ function CoverImage:menu_entry_cache()
                 callback = function(touchmenu_instance)
                     self:sizeSpinner(touchmenu_instance, "cover_image_cache_maxsize", _("Cache size"), -1, 100, 5, self.cleanCache)
                 end,
+            },
+            {
+            text = _("Cover cache folder"),
+            checked_func = function()
+                return lfs.attributes(self.cover_image_cache_path, "mode") == "directory"
+            end,
+            help_text = T(_("The actual cache path is:\n%1"), self.cover_image_cache_path),
+            keep_menu_open = true,
+            callback = function()
+                UIManager:show(ConfirmBox:new{
+                    text = _("Select a cache folder. The contents of the old folder will be migrated."),
+                    ok_text = _("Yes"),
+                    cancel_text = _("No"),
+                    ok_callback = function()
+                        self:choosePathFile("cover_image_cache_path", true, false, self.migrateCache)
+                    end,
+                })
+            end,
             },
             {
                 text = _("Clear cached covers"),
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -51,7 +51,6 @@ local CoverImage = WidgetContainer:new{
 function CoverImage:init()
     self.cover_image_path = G_reader_settings:readSetting("cover_image_path") or ""
     self.cover_image_format = G_reader_settings:readSetting("cover_image_format") or "auto"
-    self.cover_image_extension = getExtension(self.cover_image_path)
     self.cover_image_quality = G_reader_settings:readSetting("cover_image_quality") or 75
     self.cover_image_stretch_limit = G_reader_settings:readSetting("cover_image_stretch_limit") or 8
     self.cover_image_background = G_reader_settings:readSetting("cover_image_background") or "black"
@@ -84,7 +83,7 @@ function CoverImage:getCacheFile()
     local cache_file = self.cover_image_cache_path .. self.cover_image_cache_prefix .. self.ui.document:getProps().title
         .. "_" .. self.cover_image_quality .. "_" .. self.cover_image_stretch_limit .. "_" .. self.cover_image_background
 
-    local act_format = self.cover_image_format == "auto" and self.cover_image_extension or self.cover_image_format
+    local act_format = self.cover_image_format
     if act_format == "auto" then
         cache_file = cache_file .. ".jpg"
     else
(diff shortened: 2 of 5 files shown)


NiLuJe:
A bit late to the party now that this is essentially done, but, a quick entirely theoretical question:

The ScreenSaver code has never cached anything, and that's never been a significant issue. Which leads me to the actual question: I understand the concern is that this runs (roughly) on doc open.
So, my stupid question is: why not just delay that a bit, and simply add (if not already there) a check for what's currently in use so as not to do it twice for the same book?

(Remember that I only have a rough idea of what this attempts to solve in the first place ^^).

In any case, a cache certainly works, at the expense of a (minor) extra storage burden. Will read the code later tonight ;).

Frenzie:
I believe the idea is that the suspend and/or shutdown event is not known, so it has to occur sometime earlier. It does seem that adding an extra second to initial loading time may well be the best option.

In the case of the regular screensaver it adds an extra second to suspend or shutdown, but we know when it occurs so it doesn't have to happen any earlier.

plugins/coverimage.koplugin/main.lua
```diff
@@ -446,6 +533,7 @@ function CoverImage:addToMainMenu(menu_items)
                 checked_func = function()
                     return lfs.attributes(self.cover_image_fallback_path, "mode") == "file"
                 end,
+                help_text = _("This file will used as cover image on document close."),
```
Frenzie, line 536:
```suggestion
                help_text = _("This file will be used as a cover image."),
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -501,6 +583,156 @@ function CoverImage:addToMainMenu(menu_items)
...
+                                        text = _("Save"),
+                                        is_enter_default = true,
+                                        callback = function()
+                                            self.cover_image_cache_path = sample_input:getInputText()
+                                            G_reader_settings:saveSetting("cover_image_cache_path", self.cover_image_cache_path)
+                                            if lfs.attributes(self.cover_image_cache_path, "mode") ~= "directory"
+                                                and self.cover_image_cache_path ~= "" then
+                                                -- todo: Ask if the folder should be created
+                                                UIManager:show(InfoMessage:new{
+                                                    text = T(_("\"%1\" \nis not a valid folder!"),
```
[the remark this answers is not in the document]
NiLuJe: Also, is the space necessary right before a linefeed?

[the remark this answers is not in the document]
NiLuJe: As for what @Frenzie said: `PathChooser`, maybe?

While that may not be the best examples (dunno, really, just judging by the surrounding mess ^^), look at ScreenSaver's `chooseFile`/`chooseFolder`.

plugins/coverimage.koplugin/main.lua
```diff
@@ -168,6 +385,262 @@ If the filename is empty or the file doesn't exist, the cover file will be delet
...
+- Select a file
+
+or specify a new file:
+- First select a directory
+- Then enter a name for the new file]])
+
+-- menu entry: Cache settings
+function CoverImage:cover_image_cache_menu()
+    return {
+        text = _("Cover image cache settings"),
```
gbyl, line 400:
```suggestion
        text = _("Cache settings"),
```

gbyl (changes requested):

plugins/coverimage.koplugin/main.lua
```diff
@@ -189,58 +662,20 @@ function CoverImage:addToMainMenu(menu_items)
             },
             -- menu entry: filename dialog
             {
-                text = _("Set image path"),
+                text = _("Set image"),
```
gbyl, line 665: It would be helpful to display the value (i.e. the full path) right next to the menu item — this goes for the fallback image too. So, perhaps something like?

```suggestion
                text_func = function()
                    return T(_("Set image: %1"), G_reader_settings:readSetting("cover_image_path", self.cover_image_path))
                end,
```

plugins/coverimage.koplugin/main.lua
zwim: Done.

plugins/coverimage.koplugin/main.lua
zwim: I have put it in the help text (long press ) as the path can be quite long.

plugins/coverimage.koplugin/main.lua
```diff
@@ -190,57 +622,24 @@ function CoverImage:addToMainMenu(menu_items)
             -- menu entry: filename dialog
             {
                 text = _("Set image path"),
+                help_text_func = function()
+                    local text = self.cover_image_path
+                    text = text ~= "" and text or _("not set")
+                    return T(_("The actual cover image path is:\n%1"), text)
```
gbyl, line 628:
```suggestion
                    return T(_("The current cover image path is:\n%1"), text)
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -168,6 +399,207 @@ If the filename is empty or the file doesn't exist, the cover file will be delet
...
+                callback = function(touchmenu_instance)
+                    self:sizeSpinner(touchmenu_instance, "cover_image_cache_maxsize", _("Cache size"), -1, 100, 5, self.cleanCache)
+                end,
+            },
+            {
+            text = _("Cover cache folder"),
+            checked_func = function()
+                return lfs.attributes(self.cover_image_cache_path, "mode") == "directory"
+            end,
+            help_text = T(_("The actual cache path is:\n%1"), self.cover_image_cache_path),
```
gbyl, line 464:
```suggestion
            help_text = T(_("The current cache path is:\n%1"), self.cover_image_cache_path),
```

gbyl (changes requested):

plugins/coverimage.koplugin/main.lua
Frenzie: You can just say "current cache path"

Frenzie: Oops, wrong line, but the same applies here with cover image.

zwim: OK

plugins/coverimage.koplugin/main.lua
zwim: OK

zwim:
On the emulator I have no problems.
But on the Tolino: When changing cover images settings (background, format ...) everything seems to work in reader. The cache files are created.
But when I go to the filemanager KOReader fails.



logcat shows:
```
I/KOReader( 7056):  GestureDetector:probeClockSource: Touch event timestamps appear to use CLOCK_MONOTONIC
E/KOReader( 7056): [NativeThread]: failed to run script: frontend/cache.lua:161: attempt to compare nil with number
E/KOReader( 7056): [NativeThread]: Stopping due to previous errors
W/ActivityManager( 2488): Exception thrown during pause
W/ActivityManager( 2488): android.os.DeadObjectException
W/ActivityManager( 2488): 	at android.os.BinderProxy.transact(Native Method)
W/ActivityManager( 2488): 	at android.app.ApplicationThreadProxy.schedulePauseActivity(ApplicationThreadNative.java:660)
W/ActivityManager( 2488): 	at com.android.server.am.ActivityStack.startPausingLocked(ActivityStack.java:761)
W/ActivityManager( 2488): 	at com.android.server.am.ActivityStack.finishActivityLocked(ActivityStack.java:2455)
W/ActivityManager( 2488): 	at com.android.server.am.ActivityStack.requestFinishActivityLocked(ActivityStack.java:2299)
W/ActivityManager( 2488): 	at com.android.server.am.ActivityManagerService.finishActivity(ActivityManagerService.java:3441)
W/ActivityManager( 2488): 	at android.app.ActivityManagerNative.onTransact(ActivityManagerNative.java:265)
W/ActivityManager( 2488): 	at com.android.server.am.ActivityManagerService.onTransact(ActivityManagerService.java:2069)
W/ActivityManager( 2488): 	at android.os.Binder.execTransact(Binder.java:404)
W/ActivityManager( 2488): 	at dalvik.system.NativeStart.run(Native Method)
I/ActivityManager( 2488): Process org.koreader.launcher (pid 7056) has died.
```

zwim:
Could this be, because I use the `koreader/cache` folder as default? If yes, which folder can i use safely?

NiLuJe:
At a very quick glance, that would imply that you moved/fiddled with *other* stuff in the cache folder?

In any case, probably safer in a subfolder inside `data`.

zwim:
I just put my `CI_CACH_xxxxx` files in the cache folder. I do not change/delete files there.

gbyl:
On that same topic, is there a specific reason you chose `CI_CACHE_` as the prefix? Maybe just `cover_` would be better especially if it is already in the cache folder?

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +182,413 @@ function CoverImage:onReaderReady(doc_settings)
...
+    for entry in lfs.dir(self.cover_image_cache_path) do
+        if entry ~= "." and entry ~= ".." then
+            local file = self.cover_image_cache_path .. entry
+            if entry:sub(1,self.cover_image_cache_prefix:len()) == self.cover_image_cache_prefix
+                and lfs.attributes(file, "mode") == "file" then
+                    os.remove(file)
+            end
+        end
+    end
+end
```
Frenzie, line 199: @zwim 
> Could this be, because I use the koreader/cache folder as default? If yes, which folder can i use safely?

At a quick glance it looks like you're deleting *everything* instead of just your stuff?

zwim:
> On that same topic, is there a specific reason you chose `CI_CACHE_` as the prefix? Maybe just `cover_` would be better especially if it is already in the cache folder?

No specific reason, I will change it to cover_ .

plugins/coverimage.koplugin/main.lua
zwim: Maybe I am to "shorteyed":
`entry:sub(1,self.cover_image_cache_prefix:len()) == self.cover_image_cache_prefix` should only select the prefixed files?

zwim:
ToDo: not just avoid `lfs.currentdir() .. "/cache/" ` but also the android  `/sdcard/koreader/cache` dir. Comes soon.

Beside the avoid thing, the thing should work ;)

plugins/coverimage.koplugin/main.lua
Frenzie: Provided the prefix is defined properly that'll be fine. But when I read the other message I skimmed quickly and saw this `for entry in lfs.dir(self.cover_image_cache_path) do` … `os.remove`. But I did correctly myself five seconds later. :-P

zwim: Absolutely no problem. Thanks for your help.

But the underlaying problem consisst: If I use the 'cache' folder I get the error. If i use an other folder, I dont get the error.

gbyl:
I was taking a closer look at the `about_text`. How do you feel about this?

```
This plugin saves a book cover to a file. That file can then be used as a screensaver on certain devices.

If enabled, the cover image of the current file is stored in the set path on book opening. Books can be excluded if desired.

If disabled, the cover file will be deleted.

If fallback is enabled, the fallback file will be copied to the screensaver file on book closing. If the filename is empty or the file doesn't exist, the cover file will be deleted.

If fallback is disabled, the screensaver image will stay in place after closing a book.
```

plugins/coverimage.koplugin/main.lua
Frenzie: And the normal cache is working as expected?

zwim: > And the normal cache is working as expected?
As far as I see yes

I can reproduce the crash on koreader-2021.3: 

1. Copy any file in the `cache` directory.
2. Start koreader, open a book. 
3. Delete jus the file.
4. Open filemanager
5.  crash

poire-z: So, I guess koreader/cache is not just "any" cache, and is expected to be managed only by cache.lua - so we're not free to 't add/remove other stuff.
Although it is fine with subdirectories and the file fontinfo.dat.
So, I guess for your stuff: either in a subdirectory (so, still candidate to the menu item "Clear readers' cache") or elsewhere.

zwim: I haven't looked deeper in the frontend/cache thing. But I suspect all files in there are checked by the cache mechanism. I one vanishes we get the `nil` exception -> So I will use a subfolder in there as default.

NiLuJe: It keeps a ... cached (;p) list of entries, so, yeah, it does not expect stuff to mess it.

I guess `fontinfo.dat` kinda works because either it's created early and/or never removed?

NiLuJe: Ah, a list of *file* entries, actually, which is how CRe gets away with it ;).

I suspect that if you tried to delete `fontinfo.dat`, it would *also* be upset.

NiLuJe: Yes, yes, it does. I'll fix fontlist not to do that.

poire-z: Or maybe move fontlist.dat into koreader/settings/ ? It's some KOReader UI font cache, it does not really belong among "reader's caches" :)

NiLuJe: Well, it *is*, semantically, a cache, though...

zwim:
Now another problem with the filesystem arose. 

On Android `device.isValidPath` does not check the realpath (only "/sdcard" and "device.externalStorage"), but filechooser delivers the realpath. 
I have included a tweak in coverimage for this problem. 

1.) Possible solutions: Extend `device.isValidPath` with this tweak, so we move the nastiness into device.
2.) Better solution: Append the realpath-check  in `MainActivity.kt`. e.g. `path.startsWith(realpath(device.externalStorage))`.
3.) Leave the tweak in the plugin and wait for 2.)
4.) Do 1.)  and wait for 2.)

NiLuJe:
It *definitely* doesn't belong her,e and it *definitely* shouldn't be hard-coding *anything*.

At a very quick glance, I'd fix Android:isValidPath to *actually* append an util.pathExists check to the Android shenanigans, because right now it doesn't actually do the same thing as the Generic variant,n AFAICT?

zwim:
From my point of view, the plugin is finished.
Tested on an Tolino 4HD and in the emulator.

IMHO this PR is a great improvement in user-friendliness and speed. And it adds the cache for speedup.

One improvement is open for future: Add more default cover paths for known devices -> This would need some changes luajit-launcher.

frontend/device/android/device.lua
```diff
@@ -9,6 +9,7 @@ local logger = require("logger")
 local util = require("util")
 local _ = require("gettext")
 local T = FFIUtil.template
+local realpath = FFIUtil.realpath
```
NiLuJe, line 12: Caching that in a local is not strictly necessary here?

(Harmless enough, though, just unusual).

frontend/device/android/device.lua
```diff
@@ -390,7 +391,15 @@ function Device:canExecuteScript(file)
...
-    return android.isPathInsideSandbox(path)
+    -- the fast check
+    if android.isPathInsideSandbox(path) then
+        return true
+    end
+
+    -- the thorough check
+    local real_ext_storage = realpath(android.getExternalStoragePath())
+    local real_path = realpath(path)
+    return real_path:sub(1, #real_ext_storage) == real_ext_storage
```
NiLuJe, line 402: Just for information, what was the issue here, exactly? Symlinks?

plugins/coverimage.koplugin/main.lua
```diff
@@ -18,8 +21,34 @@ local util = require("util")
 local _ = require("gettext")
 local T = require("ffi/util").template
 
+local PathChooser = require("ui/widget/pathchooser")
+local Screen = require("device").screen
+local InputDialog = require("ui/widget/inputdialog")
+local realpath = ffiutil.realpath
```
NiLuJe, line 27: Not in alphabetical order (and not exactly with the rest of the requires, too).

plugins/coverimage.koplugin/main.lua
```diff
@@ -18,8 +21,34 @@ local util = require("util")
...
+-- todo: Wouldn't we like an android.deviceIdentifier() method, so we can use better default paths?
+local function getDefaultCoverPath()
+    if Device.isEmulator() then
+        return DataStorage:getDataDir() .. "/"
+    elseif Device.isAndroid() then
+        if android.prop.product == "ntx_6sl" then -- Tolino HD4 and other
+            return android.getExternalStoragePath() .. "/suspend_others.jpg"
+        else
+            return "cover.jpg"
+        end
```
NiLuJe, line 39: Yeah, for now, I'd probably move that to a method in Android's Device implementation, maybe?

NiLuJe: Speaking of: We traditionally use these as methods, not functions (e.g., Device*:*isWhatever()`)

plugins/coverimage.koplugin/main.lua
```diff
@@ -44,33 +72,32 @@ local CoverImage = WidgetContainer:new{
...
-    self.ui.menu:registerToMainMenu(self)
-
-end
 
-function CoverImage:_enabled()
-    return self.enabled
-end
+    if not lfs.attributes(self.cover_image_cache_path, "mode") then
+        lfs.mkdir(self.cover_image_cache_path)
+    end
```
NiLuJe, line 90: You can probably skip the stat and just call mkdir unconditionally, letting it fail harmlessly if the directory already exists.

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
     self:createCoverImage(doc_settings)
 end
 
+---------------------------
+-- cache handling functions
+---------------------------
+
+function CoverImage:getCacheFile()
+    local path, document_name = util.splitFilePathName(self.ui.document.file) --luacheck: no unsused
+    -- use document_name here. Title may contain characters not allowed on every filesystem (esp. vfat on /sdcard)
```
NiLuJe, line 206: Mimic Cache and use the md5 of the filename, maybe?

NiLuJe: Or even the docsetting's `partial_md5_checksum` if you don't want to re-compute it?

frontend/device/android/device.lua
zwim: The problem on Android is that the external storage path is something like `/storage/sdcard1/xxx`. A shortcut to that would be `/sdcard/xxx`. Both of them are checked by luajit-launcher's `isPathInsideSandbox`.
Unfortunately the pathChooser delivers the realpath, which is `/mnt/media_rw/xxx`

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+    return self.cover_image_cache_path .. self.cover_image_cache_prefix .. document_name
+        .. "_" .. self.cover_image_quality .. "_" .. self.cover_image_stretch_limit .. "_" .. self.cover_image_background
+        .. "_" .. self.cover_image_format .. "." .. getExtension(self.cover_image_path)
+end
+
+function CoverImage:emptyCache()
+    for entry in lfs.dir(self.cover_image_cache_path) do
+        if entry ~= "." and entry ~= ".." then
+            local file = self.cover_image_cache_path .. entry
+            if entry:sub(1,self.cover_image_cache_prefix:len()) == self.cover_image_cache_prefix
```
NiLuJe, line 216: nit: space after comma

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+end
+
+function CoverImage:getCacheFiles(cache_path, cache_prefix)
+    local cache_count = 0
+    local cache_size_KiB = 0
+    local files = {}
+    for entry in lfs.dir(self.cover_image_cache_path) do
+        if entry ~= "." and entry ~= ".." then
+            local file = cache_path .. entry
+            if entry:sub(1,self.cover_image_cache_prefix:len()) == cache_prefix
```
NiLuJe, line 231: nit: space after comma

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+    if Device.isAndroid() then
+        return self.cover_image_cache_maxfiles >= 0 and self.cover_image_cache_maxsize >= 0
+            and lfs.attributes(path, "mode") == "directory"
+            and path ~= "/sdcard/koreader/cache/"  -- interferes with frontent cache-framework
+            and realpath(path) ~= realpath(android.getExternalStoragePath() .. "/koreader/cache/")  -- interferes with frontent cache-framework
+    else
+        return self.cover_image_cache_maxfiles >= 0 and self.cover_image_cache_maxsize >= 0
+            and lfs.attributes(path, "mode") == "directory"
+            and path ~= "./cache/" -- interferes with frontent cache-framework; quick and dirty check
+            and realpath(path) ~= realpath("./cache/") -- interferes with frontent cache-framework; quick and dirty check
```
NiLuJe, line 282: getDataDir()?

NiLuJe: (Reasoning: IIRC, PWD is not necessarily KOReader's folder on every platform).

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+
+-- callback for choosePathFile()
+function CoverImage:migrateCache(old_path, new_path)
+    if old_path == new_path or not self:isCacheEnabled(new_path) then
+        return
+    end
+    for entry in lfs.dir(old_path) do
+        if entry ~= "." and entry ~= ".."   then
+            local old_file = old_path .. entry
+            if lfs.attributes(old_file, "mode") == "file" and entry:sub(1,self.cover_image_cache_prefix:len()) == self.cover_image_cache_prefix then
```
NiLuJe, line 294: nit: space after comma

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+    if old_file ~= new_file then
+        os.rename(old_file, new_file)
+    end
+end
+
+--[[--
+chooses a path or (an existing) file
+
+@touchmenu_instance for updating of the menu
+@string setting is the G_reader_setting which is used and changed
```
NiLuJe, line 315: nit: the G_reader_setting `key`

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+@boolean folder_only just selects a path, no file handling
+@boolean new_file allows to enter a new filename, or use just an existing file
+@function migrate(a,b) callback to a function to mangle old folder/file with new folder/file.
+    Can used for migrating the contents of the old path to the new one
```
NiLuJe, line 319: Can be

plugins/coverimage.koplugin/main.lua
zwim: Good idea. We could make a device specific method `device.defaultCover()` which delivers a sane cover path (if known)

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+]]
+function CoverImage:choosePathFile(touchmenu_instance, setting, folder_only, new_file, migrate)
+    local old_path, old_name = util.splitFilePathName(self[setting]) -- luacheck: no unused
```
NiLuJe, line 322: We have `__` and `dummy` placeholders whitelisted for that when `_` is taken by gettext ;).

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+                if touchmenu_instance then
+                    touchmenu_instance:updateItems()
+                end
+            end
+        end,
+    })
+end
+
+--[[--
+changes a G_reader_setting with a size spinner
```
NiLuJe, line 385: nit: `key` ;).

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,16 +197,440 @@ function CoverImage:onReaderReady(doc_settings)
...
+
+@touchmenu_instance used for updating the menu
+@string setting is the G_reader_setting which is used and changed
```
NiLuJe, line 388: Ditto

plugins/coverimage.koplugin/main.lua
zwim: Good point. partial_md5_checksum is not good, because the same document can have different cover image settings like background, format, stretching...
But md5 is fine.

plugins/coverimage.koplugin/main.lua
NiLuJe: Actually, better reword it, because that would be confusing.

e.g.,

Update a specific G_reader_setting's value via a Spinner

plugins/coverimage.koplugin/main.lua
zwim: Yes, this is just for safety. Normally this case should not happen.
But if someone selects a path, and when entering the filename the path could be replaced by `./cache/`
So I would let it in.

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,270 +190,526 @@ function CoverImage:onReaderReady(doc_settings)
...
+            return T(help, text)
+        end,
+        checked_func = function()
+            return isFileOk(self[key]) or (isPathAllowed(self[key]) and folder_only)
+        end,
+        callback = function(touchmenu_instance)
+            UIManager:show(ConfirmBox:new{
+                text = info,
+                ok_text = _("Yes"),
+                cancel_text = _("No"),
```
Frenzie, line 542: This sounds like a rephrasing of the default Cancel/OK that doesn't add anything?

zwim: You are right -> I have removed the texts.

zwim:
Propably I wont have time in the next time (Covid etc ...) to work on that. So feel free to to delete the PR or take it over.
From my side it is ready.  

@gbyl Does it work for you?

gbyl:
@zwim I'm sorry to hear that. Yes, It works perfectly. Thank you for your work and hope you feel better soon!

plugins/coverimage.koplugin/main.lua
```diff
@@ -158,270 +190,524 @@ function CoverImage:onReaderReady(doc_settings)
...
+            {
+                text = _("Clear cached covers"),
+                help_text_func = function()
+                    local cache_count, cache_size_KiB
+                        = self:getCacheFiles(self.cover_image_cache_path, self.cover_image_cache_prefix)
+                    return T(_("The cache contains %1 files and uses %2 MiB."), cache_count, math.floor((cache_size_KiB + 1023) / 1024))
+                end,
+                callback = function()
+                    UIManager:show(ConfirmBox:new{
+                        text =  _("Do you really want to clear the cover image cache?"),
```
Frenzie, line 502:
```suggestion
                        text =  _("Cear the cover image cache?"),
```

PR #7516, opened 2021-04-08
Input dialogs: do not allow blank input cont'd

hius07:
Continue to implement default behaviour in input dialogs: do not accept empty input field.
Folow up to https://github.com/koreader/koreader/pull/7501.

InputContainer will not allow blank input by deafult.
If needed to allow, set `allow_blank_input = true` before calling.

Affects page selector in the KeyValuePage widget.
Does not affect Wikipedia languages settings (https://github.com/koreader/koreader/pull/7501#discussion_r606771433).

Menu.lua: do not allow blank input in page selector.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7516)

### DIFF
--- a/frontend/ui/widget/keyvaluepage.lua
+++ b/frontend/ui/widget/keyvaluepage.lua
@@ -433,6 +433,7 @@ function KeyValuePage:init()
             hint_func = function()
                 return "(" .. "1 - " .. self.pages .. ")"
             end,
+            deny_blank_input = true
             callback = function(input)
                 local page = tonumber(input)
                 if page and page >= 1 and page <= self.pages then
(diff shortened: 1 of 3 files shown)


NiLuJe:
Anything that asks for a password needs to be able to accept an empty string.

hius07:
InputContainer: allow blank input by default.
If needed to block, set `deny_blank_input = true` before calling.

frontend/ui/widget/keyvaluepage.lua
```diff
@@ -433,6 +433,7 @@ function KeyValuePage:init()
             hint_func = function()
                 return "(" .. "1 - " .. self.pages .. ")"
             end,
+            deny_blank_input = true
```
Galunid, line 436: Missing comma, makes everything crash

Issue #7511, opened 2021-04-06
Multi-search input dialog


hius07:
To be discussed.

(1) KOReader has a powerful translation module (Google translate) invoked via Highlights menu only.
![1](https://user-images.githubusercontent.com/62179190/113664094-d31ae680-96b3-11eb-9afb-5c07daac05e7.png)
**It would be nice to have an option to enter any text for translation.**

(2) If (1) is of interest, input dialogs for Dictionary, Wikipedia and Translate are the same and can be combined to one.
![2](https://user-images.githubusercontent.com/62179190/113664193-fe053a80-96b3-11eb-9a5f-dcf6338da4bb.png)
It can substitute 3 items in the overcrowded Search main menu.

(3) It can substitute the continuous search dialog in the lookup results window
![3](https://user-images.githubusercontent.com/62179190/113664378-50465b80-96b4-11eb-8a18-f58a09682a67.png)
to invoke new search in another domain.

poire-z:
(I'm not much a user of manual lookup.)
I think (2) can be nice when invoked from (3).
I'm not much fond of (2) when invoked from the menu:
- the 4 buttons feel a bit unnatural - and would need a single menu item to do lookups (and this would feel strange, unless you can think of a nice title for it?)
- I prefer 3 menu items, with each a cancel|lookup
- Indeed, looks like we miss a menu item "Translate" just above "Translation settings" :)

hius07:
> menu item "Translate"

Which appears in the file manager menu too, why not.
Thus we get 11 menu items: 3 dict, 3 wiki, 2 translate, calibre, file search, opds.

hius07:
> unless you can think of a nice title for it?

Search for knowledge

Frenzie:
> the 4 buttons feel a bit unnatural - and would need a single menu item to do lookups (and this would feel strange, unless you can think of a nice title for it?)

The menu item could be okay(ish) but the dialog is very overcrowded.

I'd prefer to investigate a little how this kind of problem has been implemented previously, assuming we want to fundamentally change anything at all, which I'm not convinced of. (Of course the dictionary/Wikipedia/whatever settings could be stuck together in their own submenu, that's a different matter.) For example, in browsers you have a concept like search keywords and in DDG you have shebangs, but that's probably less ideal.

`w something to search on Wikipedia`
`w! something to search on Wikipedia`

Another fairly obvious one is to use radio buttons. You'd get an extra tap but less busy visuals.

[...]

PR #7523, opened 2021-04-09
Dict/Wiki lookup results window: improve the 'pencil' button

hius07:
Discussed in https://github.com/koreader/koreader/issues/7511.
Allow continue search in Dict, Wiki or Translate.
New window will stack over the window with the previous lookup results (same as with the long press in the article).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7523)

### DIFF
--- a/frontend/ui/widget/dictquicklookup.lua
+++ b/frontend/ui/widget/dictquicklookup.lua
@@ -1223,11 +1223,11 @@ function DictQuickLookup:lookupInputWord(hint)
                     end,
                 },
                 {
-                    text = _("Search dictionary"),
-                    is_enter_default = not self.is_wiki,
+                    text = _("Search Wikipedia"),
+                    is_enter_default = self.is_wiki,
                     callback = function()
                         if self.input_dialog:getInputText() == "" then return end
-                        self.is_wiki = false
+                        self.is_wiki = true
                         self:closeInputDialog()
                         self:inputLookup()
                     end,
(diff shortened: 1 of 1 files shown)


frontend/ui/widget/dictquicklookup.lua
```diff
@@ -1211,6 +1212,27 @@ function DictQuickLookup:lookupInputWord(hint)
...
+                        self:closeInputDialog()
+                        Translator:showTranslation(self.input_dialog:getInputText())
+                    end,
+                },
+                {
+                    text = _("Search dictionary"),
+                    is_enter_default = not self.is_wiki,
+                    callback = function()
+                        if self.input_dialog:getInputText() == "" then return end
+                        self.is_wiki=false
```
poire-z, line 1230: spaces around `=` please :) Makes it more readable.

poire-z (approved):
Looks fine. Untested.

poire-z:
Btw @hius07, still coding via GitHub web UI? Are you able to / do you test your changes on your device?

hius07:
Yes, it is all via web and testing on my Kindle.

rexkerr:
Thank you!  Just a few days ago I had a book where I wanted to translate a single word that was surrounded by English words, making the language detection not work since I had to select more than one word!  I was thinking of implementing this, and before I could even consider it, it's done! :)

poire-z:
> I had to select more than one

Long-press for more than 3 seconds on a single word, and you'll get the highlight dialog.

rexkerr:
Thank you @poire-z -- I see that works, but isn't very discoverable compared to this update, so I'm glad to learn that there are now two ways to get to it.

PR #7528, opened 2021-04-10
Calibre: Minor QoL fixes

NiLuJe:
* CalibreMetadata: Get rid of the now useless NULL-hunt: here, this was basically looking for `rapidjson.null` to replace them with... `rapidjson.null` :?. IIRC, that's a remnant of a quirk of the previous JSON parser (possibly even the previous, *previous* JSON parser ^^).
* CalibreSearch: Update the actually relevant NULL-hunt to make it explicit: replace JSON NULLs with Lua nils, instead of relying on an implementation detail of Lua-RapidJSON, because that detail just changed data type ;).
* UIManager: Make sure tasks scheduled during the final ZMQ callback are honored. e.g., the Calibre "Disconnect" handler. This happened to mostly work purely by chance before the event loop rework.
* Calibre: Restore a proper receiveCallback handler after receiving a book, in order not to break the "Disconnect" handler's state (and, well, get a working Disconnect handler, period ^^).
* Calibre: Unbreak metadata cache when it's initialized by a search (regression since #7159).
* Calibre: Handle UTC <-> local time conversions when checking the cache's timestamp against the Calibre metadata timestamp.
* Bump base (Unbreak CRe on Android, update RapidJSON)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7528)


pazos:
> Get rid of the useless (I think) NULL-hunt. AFAICT, this was basically looking for rapidjson.null to replace them with... rapidjson.null :?. IIRC, that's a remnant of a quirk of the previous JSON parser (possibly even the previous, previous JSON parser ^^).

Please double and triple check. We don't serialize nil values. So any book without series will miss that part when it is converted to json. rapidjson.null is a function, which we cannot serialize either (nor with `dump` & `bitser`)

NiLuJe:
> > Get rid of the useless (I think) NULL-hunt. AFAICT, this was basically looking for rapidjson.null to replace them with... rapidjson.null :?. IIRC, that's a remnant of a quirk of the previous JSON parser (possibly even the previous, previous JSON parser ^^).
> 
> Please double and triple check. We don't serialize nil values. So any book without series will miss that part when it is converted to json. rapidjson.null is a function, which we cannot serialize either (nor with `dump` & `bitser`)

This basically walked over the results from rapidjson.load, for something that'll ultimately go through rapidjson.dump, and it looked for rapidjson.null to replace it w/ rapidjson.null.

Unless I'm missing something, this no longer made any sense. IIRC, this was a leftover for an old version where the parser could leave us with a representation that blew up during serialization, but with rapidjson this is essentially a gigantic NOP, AFAICT.

And I do have perfectly sane `"series_index": null,` fields in the inbox's .metadata.calibre file.

NiLuJe:
Ah, but Search does involve Persist, and does look for JSON nulls, *that* will need a tweak.

NiLuJe:
The broken cache init from search was a regression from #7159 that I'd missed when testing it :s.

plugins/calibre.koplugin/metadata.lua
```diff
@@ -242,6 +242,7 @@ function CalibreMetadata:init(dir, is_search)
 
     local msg
     if is_search then
+        self:cleanUnused()
```
pazos, line 245: mmm, that removes entries from the json file, not from the cache itself. The cache is always slimed and `cleanUnused` prunes unused entries from the table before flushing the table to the json file. See https://github.com/koreader/koreader/blob/master/plugins/calibre.koplugin/metadata.lua#L207

What is the problem with current implementation?.

NiLuJe: `removeNull` only removes NULLs from `series` & `series_index`; if we don't do this first, we have many more fields with potential NULLs.

(This is what happened before #7159. Plus, we don't actually need those extra fields to be dumped to the cache, and `cleanUnused` only cleans the internal representiation of the loaded JSON, not the original file).

NiLuJe: Oh, wait, yes, it does dump back to the file. >_<".

NiLuJe: (Well, no, it didn't actually go through cleanUnused before #7159, but, still ;p).

NiLuJe: Instead, `getAllMetadata` duplicated cleanUnused'd behavior, without the dump, so, this is pretty much the same thing, only DRYer ;).

Frenzie (approved):
mkay

NiLuJe (commented):
> Reviewed 3 of 5 files at r1, 1 of 3 files at r2, 3 of 3 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7528)* status: all files reviewed, 1 unresolved discussion (waiting on @pazos)

PR #7529, opened 2021-04-11
Numberpicker: show hint with boundaries instead of the current value

hius07:
(1) It is convenient to see the boundaries of the input value instead of the current value (which is seen in the SpinWidget).

Before

After

(2) Is line 143 correct? Can it be removed (the input box is closed with 148)?
https://github.com/koreader/koreader/blob/bbe197bd4b8a1e44520a1a659d91ca1fa8e162c6/frontend/ui/widget/numberpickerwidget.lua#L142-L149

(3) Since (1), as we see the hint with the boundaries before input, do we need the detailed error infomessages?
https://github.com/koreader/koreader/blob/bbe197bd4b8a1e44520a1a659d91ca1fa8e162c6/frontend/ui/widget/numberpickerwidget.lua#L151
https://github.com/koreader/koreader/blob/bbe197bd4b8a1e44520a1a659d91ca1fa8e162c6/frontend/ui/widget/numberpickerwidget.lua#L156

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7529)

### DIFF
--- a/frontend/ui/widget/numberpickerwidget.lua
+++ b/frontend/ui/widget/numberpickerwidget.lua
@@ -126,7 +126,7 @@ function NumberPickerWidget:init()
         callback_input = function()
             input_dialog = InputDialog:new{
                 title = _("Enter number"),
-                input = self.formatted_value,
+                input_hint = T("(%1 - %2)", self.value_min, self.value_max),
                 input_type = "number",
                 buttons = {
                     {
--- a/frontend/ui/widget/numberpickerwidget.lua
+++ b/frontend/ui/widget/numberpickerwidget.lua
@@ -126,7 +126,7 @@ function NumberPickerWidget:init()
         callback_input = function()
             input_dialog = InputDialog:new{
                 title = _("Enter number"),
-                input_hint = T("(%1 - %2)", self.value_min, self.value_max),
+                input_hint = T("%1 (%2 - %3)", self.formatted_value, self.value_min, self.value_max),
                 input_type = "number",
                 buttons = {
                     {
@@ -140,7 +140,6 @@ function NumberPickerWidget:init()
                             text = _("OK"),
                             is_enter_default = true,
                             callback = function()
-                                input_dialog:closeInputDialog()
                                 local input_value = tonumber(input_dialog:getInputText())
                                 if input_value and input_value >= self.value_min and input_value <= self.value_max then
                                     self.value = input_value


frontend/ui/widget/numberpickerwidget.lua
```diff
@@ -126,7 +126,7 @@ function NumberPickerWidget:init()
         callback_input = function()
             input_dialog = InputDialog:new{
                 title = _("Enter number"),
-                input = self.formatted_value,
+                input_hint = T("(%1 - %2)", self.value_min, self.value_max),
```
poire-z, line 129: Not showing the current value as the starting point - to show the min/max as a hint - not sure which is best (feels like I have hated in the past some software that didn't show the current value and started from blank in case like this :)
May be include the current value in the hint ?

poire-z:
> (2) Is line 143 correct? Can it be removed (the input box is closed with 148)?

It feels like it can be removed.

> (3) Since (1), as we see the hint with the boundaries before input, do we need the detailed error infomessages?

Costs nothing, better give the info in case the user is min/max/arithmetic-disabled :)

frontend/ui/widget/numberpickerwidget.lua
Frenzie: > feels like I have hated in the past some software that didn't show the current value and started from blank in case like this 

+1

Frenzie: Just display it in the description.

Frenzie: The range, that is. Not the current value.

NiLuJe (approved):
Looks sensible on my end, any other remarks?

Frenzie:
> If we are calling the numberpicker, usually it means that we want to go far from the current value, so no sense to have it in the input field and delete.

Strongly, extremely strongly disagree.

Frenzie:
Quoting @poire-z 

>  I have hated in the past some software that didn't show the current value and started from blank in case like this

Frenzie:
Emptying things is horrible.

hius07:
> I have hated in the past some software that didn't show the current value and started from blank in case like this :)
> **May be include the current value in the hint ?**

Done

hius07:
Is it better?

Frenzie:
Yes, I think so. ;-)

NiLuJe:
I don't think the commit in question made it to the code ^^?

PR #7531, opened 2021-04-11
Centralize one time migration code after updates

NiLuJe:
There have been a couple of these this month, and keeping stuff that should only ever run once piling up in their respective module was getting ugly, especially when it's usually simple stuff (settings, files).

So, move everything to a dedicated module, run by reader.lua on startup, and that will actually only do things once, when necessary.

~~(Draft because on top of #7528; starts after 7305f8f)~~

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7531)

### DIFF
--- a/frontend/apps/reader/modules/readertypography.lua
+++ b/frontend/apps/reader/modules/readertypography.lua
@@ -17,6 +17,8 @@ local ReaderTypography = InputContainer:new{}
 -- This is used to migrate old hyph settings, and to show the currently
 -- used hyph dict language in the hyphenation menu.
 -- It will be completed with info from the LANGUAGES table below.
+-- NOTE: Actual migration is handled in ui/data/onetime_migration,
+--       which is why this hash is public.
 ReaderTypography.HYPH_DICT_NAME_TO_LANG_NAME_TAG = {
     ["@none"]                = { "@none",           "en" },
     ["@softhyphens"]         = { "@softhyphens",    "en" },
--- a/frontend/apps/reader/modules/readertypography.lua
+++ b/frontend/apps/reader/modules/readertypography.lua
@@ -12,10 +12,12 @@ local C_ = _.pgettext
 local T = require("ffi/util").template
 local Screen = Device.screen
 
+local ReaderTypography = InputContainer:new{}
+
 -- This is used to migrate old hyph settings, and to show the currently
 -- used hyph dict language in the hyphenation menu.
 -- It will be completed with info from the LANGUAGES table below.
-local HYPH_DICT_NAME_TO_LANG_NAME_TAG = {
+ReaderTypography.HYPH_DICT_NAME_TO_LANG_NAME_TAG = {
     ["@none"]                = { "@none",           "en" },
     ["@softhyphens"]         = { "@softhyphens",    "en" },
     ["@algorithm"]           = { "@algorithm",      "en" },
@@ -93,7 +95,7 @@ local LANGUAGES = {
     { "zu",               {"zul"},   "H   ",   _("Zulu"),                   "Zulu.pattern" },
 }
 
-local DEFAULT_LANG_TAG = "en-US" -- English_US.pattern is loaded by default in crengine
+ReaderTypography.DEFAULT_LANG_TAG = "en-US" -- English_US.pattern is loaded by default in crengine
 
 local LANG_TAG_TO_LANG_NAME = {}
 local LANG_ALIAS_TO_LANG_TAG = {}
@@ -106,11 +108,9 @@ for __, v in ipairs(LANGUAGES) do
         end
     end
     if hyph_filename then
-        HYPH_DICT_NAME_TO_LANG_NAME_TAG[hyph_filename] = { lang_name, lang_tag }
+        ReaderTypography.HYPH_DICT_NAME_TO_LANG_NAME_TAG[hyph_filename] = { lang_name, lang_tag }
     end
 end
-
-local ReaderTypography = InputContainer:new{}
 
 function ReaderTypography:init()
     self.menu_table = {}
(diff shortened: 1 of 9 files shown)


poire-z:
Do you plan on keeping settings_migration.lua, or moving its stuff in your onetime_migration.lua ?

NiLuJe:
It's used to upgrade individual DocSettings in ReaderUI, too, so it'd be a bit tricky :s.

NiLuJe (commented):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7531)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

Frenzie (approved):
Seems fine in principle, but *didn't test*.

frontend/ui/data/onetime_migration.lua
```diff
@@ -0,0 +1,190 @@
...
+    for k, v in pairs(ReaderStatistics.default_settings) do
+        if settings[k] == nil then
+            settings[k] = v
+        end
+    end
+    G_reader_settings:saveSetting("statistics", settings)
+end
+
+-- ScreenSaver, https://github.com/koreader/koreader/pull/7496
+if from_version < Version:getNormalizedVersion("v2021.03-35") then
```
poire-z, line 147: When making a PR that will need this kind of migration, might be tedious to guess the version it will start on (you'll have to guess what it would be, and make sure to not merge other PRs before yours).
Some other idea on how we will proceed?

frontend/apps/reader/modules/readertypography.lua
```diff
@@ -12,10 +12,12 @@ local C_ = _.pgettext
 local T = require("ffi/util").template
 local Screen = Device.screen
 
+local ReaderTypography = InputContainer:new{}
+
 -- This is used to migrate old hyph settings, and to show the currently
 -- used hyph dict language in the hyphenation menu.
 -- It will be completed with info from the LANGUAGES table below.
-local HYPH_DICT_NAME_TO_LANG_NAME_TAG = {
+ReaderTypography.HYPH_DICT_NAME_TO_LANG_NAME_TAG = {
```
poire-z, line 20: Might need some comment that the migration is done in onetime_migration.lua.
If I want to refactor all this and rename variables, it's not obvious from here that these variables are used elsewhere remote.

frontend/ui/data/onetime_migration.lua
NiLuJe: That's actually easy: take a [rancom commit](https://github.com/koreader/koreader/commit/2f9db25969f4af2f2961d45d981f57174ee638c0), then look at the bottom of the header:

Screenshot_20210412_163222

You've got a link to the branch, a link to the PR this commit is from... and a link to the first tag that implemented this commit ;).

But that's only if there *is* a "future" tag.

For "current" code, you can just take the revision of the last nightly and just append 1 ;).

Frenzie:
Perhaps the concern is with manual builds? But yes, otherwise I agree that
the current nightly +1 should be fine.

poire-z:
> For "current" code, you can just take the revision of the last nightly and just append 1 ;).

But if I make a PR, with current nightly +1, and it sits a few days and nights before merge, i/we'll have to edit it again just before merging for it to be accurate.
A hardcoded date (updated by the PR) in onetime_migration.lua would have worked just as well, without the little complications of using the git-rev date (even if using that makes it easier to follow up when/what).
(But ok with this PR as it is.)

NiLuJe:
We don't have a date for releases, so, that's kind of a no-go anyway ;).

poire-z:
Not talking about release, real dates, or anything git.
We could have had just:
```lua
local CURRENT_MIGRATION_TAG = 20210412  -- or 3 or anything we can increment
if from_version < 20210317 then ... end
G_reader_settings:saveSetting("last_migration_version", CURRENT_MIGRATION_TAG)
```
which would have worked just as well, without being dependant on a git tag/rev stuff that we're not sure about when initially pushing the PR.

NiLuJe:
Oh, right.

Might even be able to automagic it by using the ctime of the git-rev file...

poire-z:
I'm not sure I'm happier with a build date :) which can be even more random than stuff depending on git-rev - i.e. if people build themselves an older tag.

I was just thinking of hardcoding a number, like you did with `DB_SCHEMA_VERSION = 20201022` in `statistics.koplugin/main.lua`. The day some stuff needed to be migrated, you just did in the PR:
```diff
-local DB_SCHEMA_VERSION = 20201010
+local DB_SCHEMA_VERSION = 20201022
```
And no matter if this PR made it into master on 2020/10/22 or 2020/10/29, it would have used `20201022` as the id of the change.

Could just as well be:
```diff
-local DB_SCHEMA_VERSION = 3
+local DB_SCHEMA_VERSION = 4
```
But an integer that looks like the date the developper implemented the change on its own machine is just a bit better to guess when it happened.

NiLuJe:
Meh, that's much less fun :).

But, yeah, I get your point ^^.

NiLuJe (commented):
> Reviewed 5 of 7 files at r3, 2 of 2 files at r4, 1 of 1 files at r5.
> Dismissed @poire-z from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7531)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

___
*[frontend/apps/reader/modules/readertypography.lua, line 20 at r2](https://reviewable.io/reviews/koreader/koreader/7531#-MY1zc6JQFseJLee7SB2-r2-20:-MY75Gk80KVAiMc1vTki:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/d3c5ab73c785fb098aed27a87df5fe8daaeaace6/frontend/apps/reader/modules/readertypography.lua#L20)):*

Might need some comment that the migration is done in onetime_migration.lua.
If I want to refactor all this and rename variables, it's not obvious from here that these variables are used elsewhere remote.
</blockquote></details>

Done.
___
*[frontend/ui/data/onetime_migration.lua, line 147 at r2](https://reviewable.io/reviews/koreader/koreader/7531#-MY1E6Pxjzv_5J1mnyyW-r2-147:-MY75Xv49HmTe9iHkxad:b-896fix) ([raw file](https://github.com/koreader/koreader/blob/d3c5ab73c785fb098aed27a87df5fe8daaeaace6/frontend/ui/data/onetime_migration.lua#L147)):*

That's actually easy: take a [random commit](https://github.com/koreader/koreader/commit/2f9db25969f4af2f2961d45d981f57174ee638c0), then look at the bottom of the header:

Screenshot_20210412_163222

You've got a link to the branch, a link to the PR this commit is from... and a link to the first tag that implemented this commit ;).

But that's only if there *is* a "future" tag.

For "current" code, you can just take the revision of the last nightly and just append 1 ;).
</blockquote></details>

Done.

poire-z (approved):
Thanks, looks more robust :)
Sure there's no issue re-doing each of these migration steps as a first and last one-time?

NiLuJe:
Well, it's pretty much a one and done affair every time ;).

Frenzie (approved):
Not sure how much it's needed, but there doesn't really seem to be error handling?

NiLuJe:
It would abort early and spam the logs on every subsequent boot since the setting won't be updated in case of failures ^^.

But nothing should really cause an abort, it's just playing with settings and files.

The only thing that needed guarding was the OPDS stuff, since the setting may not exist *at all* yet (fresh install).

Moving non-existing files is perfectly fine, but that's also caught and logged for verbosity's sake.

PR #7535, opened 2021-04-11
add hasSystemFonts device property

pazos:
Add system + user paths to the ReMarkable (it has "normal" linux paths)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7535)

### DIFF
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -89,6 +89,7 @@ local Device = Generic:new{
     hasClipboard = yes,
     hasOTAUpdates = canUpdateApk,
     hasFastWifiStatusQuery = yes,
+    hasSystemFonts = yes,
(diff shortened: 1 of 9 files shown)


NiLuJe:
Just in case: are we sure there aren't encrypted/obfuscated fonts in there?

Frenzie (approved):
lgtm

pazos:
> Just in case: are we sure there aren't encrypted/obfuscated fonts in there?

The path is the same that it used to be. We were looking in a few dirs under /usr/share/fonts and now we iterate the entire dir. If there's a borked font there CRengine will error without problems.

The new path added `/home/root/.local/share/fonts` is empty by default. It just happen to be a good dir to place fonts because it is in the "user" partition and stock Rm software will look there too.

Issue #7530, opened 2021-04-11
Sortwidget footer redesign


hius07:
Sorry for this, but the Sortwidget footer is rather ugly.

![1](https://user-images.githubusercontent.com/62179190/114312389-29948480-9afb-11eb-8394-075127670736.png)
---
Is it worth to redesign it using the standard chevrons and Cancel/Save? Kind of:

![4](https://user-images.githubusercontent.com/62179190/114312969-534eab00-9afd-11eb-9eee-3c38b54075ad.png)
---

Move up/down buttons might be kept.

hius07:
![5](https://user-images.githubusercontent.com/62179190/114315862-9878da00-9b09-11eb-8436-569c618dec49.png)
---
![6](https://user-images.githubusercontent.com/62179190/114315866-a0d11500-9b09-11eb-89bf-c121add45786.png)
---
![7](https://user-images.githubusercontent.com/62179190/114315870-a890b980-9b09-11eb-933a-18080f00ba01.png)

Frenzie:
Be nice. ;-)

Anyway, the second (or is that the third?) proposal looks fine to me.

poire-z:
I agree the look is different from everything else, but I'm not meeting that SortWidget very often.
I'm fine with the proposed changes (including the big black solid triangles, it shows they apply to the solid black highlighted item).

hius07:
> I'm not meeting that SortWidget very often.

Three times only: Status bar, Favorites, Dictionaries.

I'll polish it and make a PR.

hius07:
Any suggestions?

---
(1)
![1](https://user-images.githubusercontent.com/62179190/114429758-65485080-9bc6-11eb-9d53-63fc100eef68.png)
---
(2)
![2](https://user-images.githubusercontent.com/62179190/114429815-73966c80-9bc6-11eb-9d93-54919c2e4bff.png)
---
(3)
![3](https://user-images.githubusercontent.com/62179190/114429874-814bf200-9bc6-11eb-9982-6809b32e5698.png)
---

NiLuJe:
2., definitely ;).

poire-z:
I prefer 2 too.

hius07:
OK, thanks

PR #7544, opened 2021-04-13
SortWidget footer redesign

hius07:
Discussion https://github.com/koreader/koreader/issues/7530

---

---

---

---

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7544)


frontend/ui/widget/sortwidget.lua
```diff
@@ -209,99 +177,89 @@ function SortWidget:init()
...
-        text_font_size = 28,
         padding = 0,
         radius = 0,
         show_parent = self,
     }
-
     self.footer_ok = Button:new{
-        text= "✓",
-        width = self.width_widget * 13 / 100,
+        text= "Save",
```
zwim, line 246: Shouldn't we use `_("Save")` here for translations?

frontend/ui/widget/sortwidget.lua
```diff
@@ -209,99 +177,89 @@ function SortWidget:init()
...
-        text_font_size = 28,
         bordersize = 0,
-        padding = 0,
         radius = 0,
         show_parent = self,
     }
     self.footer_cancel = Button:new{
-        text = "✘",
-        width = self.width_widget * 13 / 100,
+        text = "Cancel",
```
zwim, line 237: Same here `_("Cancel")`?

NiLuJe (changes requested):
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7544)* status: 0 of 1 files reviewed, 7 unresolved discussions (waiting on @hius07 and @zwim)

___
*[frontend/ui/widget/sortwidget.lua, line 56 at r1](https://reviewable.io/reviews/koreader/koreader/7544#-MYAiP119WQBJLNfCtl0:-MYAiP119WQBJLNfCtl1:b-6eaiw3) ([raw file](https://github.com/koreader/koreader/blob/b6b332e311a282f0f80dff24412197b2db5f1d6c/frontend/ui/widget/sortwidget.lua#L56)):*
> ```Lua
>         },
>     }
>     if self.use_top_page_count then
> ```

I assume the feature was never used anywhere?
___
*[frontend/ui/widget/sortwidget.lua, line 180 at r1](https://reviewable.io/reviews/koreader/koreader/7544#-MYAhs2m0sIyymVzH_EK:-MYAhYsODFdBOC4vsTaF:b-w2sn8u) ([raw file](https://github.com/koreader/koreader/blob/899f6e4236dde80aab1f241c10ebdbe3e35c6bd2/frontend/ui/widget/sortwidget.lua#L180)):*
> ```
>     self.footer_center_width = self.width_widget * 22 / 100
> ```

Explanation for the magic numbers?
___
*[frontend/ui/widget/sortwidget.lua, line 209 at r1](https://reviewable.io/reviews/koreader/koreader/7544#-MYAiKWV95TwgVS5eNKA:-MYAiKWV95TwgVS5eNKB:bz71san) ([raw file](https://github.com/koreader/koreader/blob/899f6e4236dde80aab1f241c10ebdbe3e35c6bd2/frontend/ui/widget/sortwidget.lua#L209)):*
> ```Lua
>     }
>     self.footer_first_up = Button:new{
>         icon=chevron_first,
> ```

nit: space around `"`
___
*[frontend/ui/widget/sortwidget.lua, line 223 at r1](https://reviewable.io/reviews/koreader/koreader/7544#-MYAiNNT4eY4m1kkgs_7:-MYAiNNT4eY4m1kkgs_8:b-7ropln) ([raw file](https://github.com/koreader/koreader/blob/899f6e4236dde80aab1f241c10ebdbe3e35c6bd2/frontend/ui/widget/sortwidget.lua#L223)):*
> ```Lua
>     }
>     self.footer_last_down = Button:new{
>         icon=chevron_last,
> ```

ditto
___
*[frontend/ui/widget/sortwidget.lua, line 246 at r1](https://reviewable.io/reviews/koreader/koreader/7544#-MYA-ODRgRpCyxuzI2GI-r1-246:-MYAiGCUEp37vsYLtOhD:b-u7aw4l) ([raw file](https://github.com/koreader/koreader/blob/899f6e4236dde80aab1f241c10ebdbe3e35c6bd2/frontend/ui/widget/sortwidget.lua#L246)):*
<details><summary><i>Previously, zwim wrote…</i></summary><blockquote>

Shouldn't we use `_("Save")` here for translations?
</blockquote></details>

Yup, ditto for `Cancel` above.
___
*[frontend/ui/widget/sortwidget.lua, line 421 at r1](https://reviewable.io/reviews/koreader/koreader/7544#-MYAiYSa-uKzmeCLe2Od:-MYAiYSa-uKzmeCLe2Oe:b-1gjxoq) ([raw file](https://github.com/koreader/koreader/blob/899f6e4236dde80aab1f241c10ebdbe3e35c6bd2/frontend/ui/widget/sortwidget.lua#L421)):*
> ```Lua
>     end
>     if self.marked > 0 then
>         self.footer_first_up:setText("▲", self.footer_button_width)
> ```

I'd switch to a simple SVG icon for those, because the alignment is super-wonky right now (baselines don't match).

NiLuJe (commented):
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7544)* status: 0 of 1 files reviewed, 7 unresolved discussions (waiting on @hius07 and @zwim)

___
*[frontend/ui/widget/sortwidget.lua, line 209 at r1](https://reviewable.io/reviews/koreader/koreader/7544#-MYAiKWV95TwgVS5eNKA:-MYAiw7cCAZW3uDQUlUV:b-fcja8f) ([raw file](https://github.com/koreader/koreader/blob/899f6e4236dde80aab1f241c10ebdbe3e35c6bd2/frontend/ui/widget/sortwidget.lua#L209)):*

nit: space around `"`
</blockquote></details>

Err, `=` ^^

NiLuJe:
Try w/ these :).
[move_SVGs.zip](https://github.com/koreader/koreader/files/6304719/move_SVGs.zip)

hius07:
Thank you. I'll check for the icons.

`use_top_page_count` is in the KVP, DKVP and here only, never used.
22% center button width is from the previous version, looks rather well.

frontend/ui/widget/sortwidget.lua
hius07: Done, thanks

frontend/ui/widget/sortwidget.lua
```diff
@@ -396,7 +342,7 @@ function SortWidget:init()
 end
 
 function SortWidget:nextPage()
-    local new_page = math.min(self.show_page+1, self.pages)
+    local new_page = math.min(self.show_page + 1, self.pages)
```
Frenzie, line 345: Is there a particular reason for these random changes?

frontend/ui/widget/sortwidget.lua
```diff
@@ -505,11 +457,9 @@ function SortWidget:onSwipe(arg, ges_ev)
         self:onNextPage()
     elseif direction == "east" then
         self:onPrevPage()
-    elseif direction == "south" then
-        -- Allow easier closing with swipe down
+    elseif direction == "south" then -- close with swipe down
```
Frenzie, line 460: Please revert.

frontend/ui/widget/sortwidget.lua
```diff
@@ -505,11 +457,9 @@ function SortWidget:onSwipe(arg, ges_ev)
...
         self:onClose()
-    elseif direction == "north" then
-        -- no use for now
+    elseif direction == "north" then -- no use for now
```
Frenzie, line 462: Please revert.

frontend/ui/widget/sortwidget.lua
hius07: Shouldn't math signs be separated with spaces?

Frenzie: We have a tacit agreement that it's part context, part personal preference. There's a difference between `page+1` (fairly clear) and `page*6*3*4*9` (hard to read).

I think we also agree that we rather wouldn't see some unrelated commit mentioned in git blame unless it were necessary.

hius07:
> move_SVGs.zip

Looks good. Got an issue with the icons moving to the left after `setIcon`.





It seems that `setIcon` is not used anywhere so there are no working examples. Is it tested?
I resolved the issue with commenting line 167. Is it correct?
https://github.com/koreader/koreader/blob/47c59e0e5a48f825ad78a7405736ea5fe0267a70/frontend/ui/widget/button.lua#L164-L170

frontend/ui/widget/sortwidget.lua
hius07: Done

frontend/ui/widget/sortwidget.lua
hius07: Reverted

NiLuJe:
Possibly designed for something else, but who knows what ;o).

Probably leaner to "fix" it like setText:

```
diff --git a/frontend/ui/widget/button.lua b/frontend/ui/widget/button.lua
index 4a862d915..ad29168bb 100644
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -161,10 +161,10 @@ function Button:setText(text, width)
     end
 end
 
-function Button:setIcon(icon)
+function Button:setIcon(icon, width)
     if icon ~= self.icon then
         self.icon = icon
-        self.width = nil
+        self.width = width
         self:init()
     end
 end
```
And then just pass an extra argument with the current width in your `setIcon` call ;).

hius07:
Sorry, cannot find out if I can add this new svg files to the repository from web interface?

NiLuJe:
There's an "Add file" button in the top-right corner of the file listing.

frontend/ui/widget/sortwidget.lua
```diff
@@ -209,99 +177,89 @@ function SortWidget:init()
     local padding = Size.padding.large
     self.width_widget = self.dimen.w - 2 * padding
     self.item_width = self.dimen.w - 2 * padding
+    self.footer_center_width = self.width_widget * 22 / 100
```
NiLuJe, line 180:
```suggest
self.footer_center_width = math.floor(self.width_widget * 22 / 100)
```

frontend/ui/widget/sortwidget.lua
```diff
@@ -209,99 +177,89 @@ function SortWidget:init()
...
+    self.footer_button_width = self.width_widget * 12 / 100
```
NiLuJe, line 181:
```suggestion
    self.footer_button_width = math.floor(self.width_widget * 12 / 100)
```

frontend/ui/widget/sortwidget.lua
```diff
@@ -407,7 +353,7 @@ function SortWidget:nextPage()
 end
 
 function SortWidget:prevPage()
-    local new_page = math.max(self.show_page-1, 1)
+    local new_page = math.max(self.show_page - 1, 1)
```
NiLuJe, line 356: Ditto above for `SortWidget:nextPage()`.

NiLuJe:
Unrelated to the PR? but while you're here:

L310 is missing a `math.floor` wrapper:

```
    self.item_margin = math.floor(self.item_height / 8)
```

NiLuJe:
Quickly tested, and it looks good.

One minor potential concern: `footer_button_width` might be a tad low for translations. Even in English, the "Cancel" button is feeling cramped ;).

(I don't actually remember what happens if the text overflows the width. Experiment time!).

NiLuJe:
> (I don't actually remember what happens if the text overflows the width. Experiment time!).

Apparently the answer is "all hell breaks loose and the alignment goes bonkers" ^^.

NiLuJe:
> Unfortunately, I don't really see a better fix than not mixing text & icons.

Or restoring the original cross/checkmarks, which won't suffer from the issue, and won't look all that terrible:

sortwidget

frontend/ui/widget/sortwidget.lua
hius07: https://github.com/koreader/koreader/pull/7544#discussion_r612638902

NiLuJe: Well, I tend to prefer spaces no matter the context, but regardless of what you end up with, it should be consistent ;).

Right now one has spaces and not the other ;).

hius07:
Oh, I'm so sad wth these pseudo-italic cross/checkmark, it was the main reason for redesign.


---

frontend/ui/widget/sortwidget.lua
hius07: Done, I like spaces.

NiLuJe:
Well, we can possibly find something else (icons, I mean), but, FWIW, I prefer 'em to whatever you tried in your latest screengrabs ;).

NiLuJe:
More on that tomorrow, but at a quick glance, I can see a cancel and checkmark icon in mdilight ;).

NiLuJe:
By which I mean: https://github.com/Templarian/MaterialDesignLight/blob/master/icons/svg/cancel.svg & https://github.com/Templarian/MaterialDesignLight/blob/master/icons/svg/check.svg

hius07:
> By which I mean:

Thanks, I'll play with them.

hius07:
I like it very much, light and balanced.

frontend/ui/widget/sortwidget.lua
Frenzie: It's not the spaces, it's the diff polluting.

hius07: Restored originals
https://github.com/koreader/koreader/blob/7092df7f4e09f595239afee0706ce20609abefd1/frontend/ui/widget/sortwidget.lua#L399
https://github.com/koreader/koreader/blob/7092df7f4e09f595239afee0706ce20609abefd1/frontend/ui/widget/sortwidget.lua#L410

Frenzie: Thanks. :-)

NiLuJe:
I'll tweak the stroke width to unify it with our actual stroke width, but otherwise looks okay.

I'd feel better if the Cancel button actually *did* Cancel a move, instead of duplicating the close button, but that's admittedly another kettle of fish ;).

PR #7545, opened 2021-04-13
Calibre: More QoL tweaks

NiLuJe:
* Wireless: Optimize memory usage in StreamMessageQueue (use an array of string ropes, that we only concatenate once). Allowed to relax the throttling, making transfers that much faster.
* Persist: Add a "zstd" codec, that uses the "luajit" codec, but compressed via zstd. Since both of those are very fast, it pretty much trounces everything in terms of speed and size ;).
* Persist: Implemented a "writes_to_file" framework, much like the existing "reads_from_file" one. And use it in the zstd codec to avoid useless temporary string interning.
* Metadata: Switch to the zstd codec.

Requires https://github.com/koreader/koreader-base/pull/1355

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7545)

### DIFF
--- a/frontend/persist.lua
+++ b/frontend/persist.lua
@@ -6,7 +6,7 @@ local lfs = require("libs/libkoreader-lfs")
 local logger = require("logger")
 local zstd = require("ffi/zstd")
 
---local C = ffi.C
+local C = ffi.C
 
 local function readFile(file, bytes)
     local f, str, err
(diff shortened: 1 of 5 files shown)


NiLuJe (commented):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader/7545)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

PR #7546, opened 2021-04-13
Button setIcon: keep button width

hius07:
Allow to keep the icon centered after `setIcon`.
`setIcon` isn't used anywhere at the moment (going to be).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7546)

### DIFF
--- a/frontend/ui/widget/button.lua
+++ b/frontend/ui/widget/button.lua
@@ -161,10 +161,10 @@ function Button:setText(text, width)
     end
 end
 
-function Button:setIcon(icon)
+function Button:setIcon(icon, width)
     if icon ~= self.icon then
         self.icon = icon
-        self.width = nil
+        self.width = width
         self:init()
     end
 end


NiLuJe:
Looking at the code flow, I'm not quite sure *why* this works, actually.

IconWidget's already using a fixed width for our icons.

The only thing that might be throwing a kink in the works is that this causes the FrameContainer to be reinstantiated, which makes me think that something else is actually causing the shift.

hius07:
There was no shift in the first version, with text labels "▲" and `setText` switching to the icon labels.
I needed Button:init() before `setIcon` to remove the text label.

NiLuJe:
Oh, right. That's a Text -> Icon Button switch, hence the width being different.

Which begs the question: Why are they still sometimes Text Buttons?

hius07:
> Why are they still sometimes Text Buttons?

How is this determined?
`Cancel`, `Page 1 of 2`, `Save` are Text.
Chevrons and Move up/down have no text labels.

NiLuJe:
Just that, being instantiated with an `icon` field instead of a `text` field. But never mind me, I was looking at an old diff, that's not the case.

So, nope, don't really understand why this helps, as it should be a NOP.

hius07:
Before this PR, when pressing the button after `setIcon`, flash_ui shows square button, the width has been lost.
So I believe this PR is okay.

NiLuJe:
I'm not saying that it doesn't work, just that I don't understand why it does, since icons have a common and fixed size. And that  I don't have time to investigate right now ;).

NiLuJe:
Okay, that's because those Buttons are instantiated with a custom width, as opposed to let Button's init use the ImageWIdget's dimensions instead.

So, yeah, makes perfect sense :).

NiLuJe (approved):

Issue #7519, opened 2021-04-08
FR: 'suspend', 'reboot' and 'poweroff' routines for remarkable


gbyl:
KOReader currently relies on `systemd` for suspend, reboot and poweroff functions. According to @pazos it should be possible to [implement our own routines](https://github.com/koreader/koreader/issues/7506#issuecomment-813097010) — which would unlock a few other features for remarkable users.

I am not sure what this would entail, as for example `poweroff` is always controlled by the system. This is true even if the line below is changed to `os.execute("poweroff")` or `os.execute("/sbin/poweroff")`

https://github.com/koreader/koreader/blob/5b996c8ab76ff4d4875b45695627596876017b68/frontend/device/remarkable/device.lua#L189

pazos:
probably `poweroff` and `reboot` and symbolic links to `systemctl`

That would explain why you still see the system screensaver when you execute one of these commands.

Since `busybox` is installed on the Remarkable I would check if reboot/halt/poweroff applets are already available and see what happens with `os.execute("busybox poweroff")` or similar.

If my supposition is true there's a systemd service in charge of updating the poweroff screen. Disabling that service would be another way of preventing that behaviour. It should be as simple as `systemctl disable nameOfService`.

Please check both things and provide feedback. Thanks!

gbyl:
You are right! They are both symlinked to `systemctl`. However, `busybox poweroff` returns `poweroff: applet not found`.

snelg:
Yes, looks like it's systemd stuff:

Pertinent lines from `/etc/systemd/system/poweroff.target.wants/remarkable-shutdown.service`
```
Description=Show shutdown image
```
```
ExecStart=-/usr/bin/remarkable-shutdown
```

And from `/etc/systemd/system/reboot.target.wants/remarkable-reboot.service`
```
Description=Show reboot image
```
```
ExecStart=-/usr/bin/remarkable-shutdown --reboot
```

pazos:
Mistery solved: my suggestion would be to rely on the screensaver (vs the screencover plugin) and disable that systemd service if the user doesn't need it. I think you can include it as a section in https://github.com/koreader/koreader/wiki/Installation-on-ReMarkable

Something like:

## Optional: disable system splashscreen.

The system displays stock images when is powered off and rebooted. If you want to get rid of that behaviour just do ...

gbyl:
@pazos That definitely did it! Two small caveats though:

[...]

PR #7563, opened 2021-04-17
remarkable: remove extra refresh before poweroff

gbyl:
"we did move rM to the same power event handler [as] Kobo, so it should already be doing a white flash before [ScreenSaver]" https://github.com/koreader/koreader/issues/7519#issuecomment-819649806

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7563)

### DIFF
--- a/frontend/device/remarkable/device.lua
+++ b/frontend/device/remarkable/device.lua
@@ -190,8 +190,6 @@ function Remarkable:suspend()
 end
 
 function Remarkable:powerOff()
-    self.screen:clear()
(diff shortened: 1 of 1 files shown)


NiLuJe:
IIRC, that one was not for our benefit, but for systemd's ;).

gbyl:
@NiLuJe That's right, in fact I requested it myself originally. I will add info on the wiki to encourage users to use the KOReader poweroff instead of `remarkable-shutdown`.

PR #7559, opened 2021-04-17
Filemanager: long-press the "+" button to refresh displaying content

hius07:
c.f. https://github.com/koreader/koreader/issues/7477

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7559)

PR #7564, opened 2021-04-17
filemanager: add home button refresh

gbyl:
fixes https://github.com/koreader/koreader/issues/7477
and complements https://github.com/koreader/koreader/pull/7559

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7564)


NiLuJe:
Shouldn't it be the other way around?

gbyl:
@NiLuJe I just tested it and it works either way. Should I switch them?

NiLuJe:
Feels slightly more logical the other way around, yeah.

PR #7566, opened 2021-04-18
screensaver: add reboot/poweroff message toggle

gbyl:
UI to enable/disable the Screensaver fallback message as discussed here https://github.com/koreader/koreader/issues/7519#issuecomment-821851836

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7566)

### DIFF
--- a/frontend/ui/screensaver.lua
+++ b/frontend/ui/screensaver.lua
@@ -46,6 +46,9 @@ end
 if G_reader_settings:hasNot("screensaver_delay") then
     G_reader_settings:saveSetting("screensaver_delay", "disable")
 end
+if G_reader_settings:hasNot("disable_fallback_msg") then
+    G_reader_settings:makeFalse("disable_fallback_msg")
+end
 
 local Screensaver = {
     screensaver_provider = {
@@ -385,7 +388,7 @@ function Screensaver:setup(event, fallback_message)
     if G_reader_settings:has(self.prefix .. "screensaver_type") then
         self.screensaver_type = G_reader_settings:readSetting(self.prefix .. "screensaver_type")
     else
-        if event and not G_reader_settings:readSetting("disable_fallback_msg") then
+        if event and not G_reader_settings:isTrue("disable_fallback_msg") then
             -- Display the provided fallback_message over the screensaver,
             -- so the user can distinguish between suspend (no overlay),
             -- and reboot/poweroff (overlaid message).
--- a/frontend/ui/screensaver.lua
+++ b/frontend/ui/screensaver.lua
@@ -46,8 +46,8 @@ end
 if G_reader_settings:hasNot("screensaver_delay") then
     G_reader_settings:saveSetting("screensaver_delay", "disable")
 end
-if G_reader_settings:hasNot("screensaver_disable_fallbackmsg") then
-    G_reader_settings:makeFalse("screensaver_disable_fallbackmsg")
+if G_reader_settings:hasNot("screensaver_hide_fallback_msg") then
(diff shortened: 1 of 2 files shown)


frontend/ui/screensaver.lua
```diff
@@ -385,7 +385,7 @@ function Screensaver:setup(event, fallback_message)
     if G_reader_settings:has(self.prefix .. "screensaver_type") then
         self.screensaver_type = G_reader_settings:readSetting(self.prefix .. "screensaver_type")
     else
-        if event then
+        if event and not G_reader_settings:readSetting("disable_fallback_msg") then
```
NiLuJe, line 388: For consistency's sake, inititialize it to `false` like the rest of the settings (c.f., top of this module), and switch to an `isTrue` check here ;).

gbyl: Done! Let me know if it looks good now.

NiLuJe: Yep, I just forgot to mention that I'd prefer is everything kept a `screensaver_` prefix, too ;).

NiLuJe: (And Luacheck's trailing whitespace warning @ L197 ;)).

gbyl: Sounds good. I renamed them accordingly.

frontend/ui/screensaver.lua
```diff
@@ -385,7 +388,7 @@ function Screensaver:setup(event, fallback_message)
     if G_reader_settings:has(self.prefix .. "screensaver_type") then
         self.screensaver_type = G_reader_settings:readSetting(self.prefix .. "screensaver_type")
     else
-        if event then
+        if event and not G_reader_settings:isTrue("screensaver_disable_fallbackmsg") then
```
NiLuJe, line 391: `isFalse` is slightly less tortuous ^^.

(And being able to use it here, and `toggle` in the Settings, is actually the reasoning behind the forced initialization ;)).

NiLuJe: I'd probably keep a `_` between `fallback` and `msg` in the setting key, while you're there ;).

gbyl: Yeah, I was confused about the convention. Some settings have `msg` others have `message` spelled out. And none seem to exceed 3 words.

NiLuJe: Yeah, that was more coincidence and things growing out in a slightly haphazard fashion that anything.

I'd KISS: a sensible prefix, and a meaningful name. We don't really care about length ;).

NiLuJe (approved):

frontend/ui/screensaver.lua
gbyl: Just realized that on devices with poor power management, it makes more sense to have a message displayed when sleeping (because of the battery drain) and not when powered off. In other words, the opposite of a normal ereader. Is there a not convoluted way of adding that to this PR?

NiLuJe: I would probably no rock the boat too much and leave that one to user preferences (by, e.g., setting a custom message)?

If you suffered through the surrounding code, you'll maybe have noticed that those options support custom variants prefixed by those event names, so you can technically get a custom message that's different for sleep, reboot & poweroff.

Exposing that in the UI would be a bit nightmarish, hence the current status quo ^

gbyl: Fair enough. Sounds like that would be a more substantial overhaul. Let's leave this PR like this then.

Issue #6741, opened 2020-10-01
"Full screen refresh" performed in CR Engine even when the gesture plugin is disabled


gbyl:
* KOReader version: All
* Device: All

#### Issue

When running the engine, one finger swipe configuration always performs a **full screen refresh** even when configured to `nothing`. This only occurs when running the reader engine. When running the file browser the swipe configuration works as expected and there is no fullscreen refresh.

#### Steps to reproduce
1. Open any epub file 
2. Settings > Taps and gestures > Gesture manager > Short diagonal swipe
3. Set as "Nothing"
4. Test the following swipes in the middle of the screen:
    - one finger short diagonal swipe (all directions)
    - one finger vertical swipe (top-down and bottom-up)
5. A "full screen refresh" is performed every single time

NiLuJe:
Since this predates the gesture manager by... nearly ten years, it's entirely possible that some legacy code was missed in the migration ;).

gbyl:
Unlikely this will ever be addressed

yparitcher:
setting it to nothing prevent the refresh on my kindle.

setting it to pass-through will use the default handler which for the reader is also a full refresh see: https://github.com/koreader/koreader/blob/b0a10777561aa081af857b46a3e90ee9a22e64d8/frontend/apps/reader/modules/readerrolling.lua#L522-L527

gbyl:
@yparitcher Unfortunately that does not stop the full refresh from occurring. In fact, it keeps happening even after disabling the gestures plugin. Which makes sense considering https://github.com/koreader/koreader/issues/6741#issuecomment-702684567

gbyl:
> Since this predates the gesture manager by... nearly ten years, it's entirely possible that some legacy code was missed in the migration ;).

@NiLuJe That makes sense. Perhaps someone will find a solution in the future. I'll reopen it so that it is at least documented.

gbyl:
@poire-z That makes perfect sense now! And commenting out that line works as you suggested. Thanks.

> Another question is how/why are you doing these vertical swipes, up to the point of being annoyed by the refreshes?
(I personally do these vertical to get a refresh, I'm used to that, never go the diagonal way.)

I use the two-finger gestures often and ocassionally when the engine fails to detect the input it thinks I am using only one finger and then it triggers a refresh.

> Not sure why you say it happens in CRE only, really not in PDF ?

You are right. It happens with PDFs too. I just tested it.

yparitcher:
[...]

Issue #6071, opened 2020-04-19
One finger gestures not configurable


lschneiderbauer:
* KOReader version: 2020.04.1
* Device: Remarkable

#### Issue
I tried to deactivate the gestures for scrolling through pages (since the remarkable has buttons), while in reading mode. I set every single one of them to "Nothing" in the settings (even taps, corner swipes, etc.). However, the swipe gestures and taps still change the page.

gbyl:
I have also searched for a way to disable taps when reading as it is too easy to accidentally change the page. [This comment](https://www.mobileread.com/forums/showpost.php?s=e7c59b15b6617cab6acd09bba02dd963&p=3657873&postcount=2) by @Frenzie is the closest I found to a solution

hius07:
(1) Disable touch input, or
(2) Set size of backward/forward zones to 0% in the Taps & gestures menu.

PR #7572, opened 2021-04-19
UI: Page Turns Menu

gbyl:
https://github.com/koreader/koreader/issues/6741 (second request) and https://github.com/koreader/koreader/issues/6071

Thanks to @NiLuJe for providing guidance

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7572)

### DIFF
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -167,9 +167,7 @@ function ReaderPaging:setupTouchZones()
             screen_zone = {
                 ratio_x = 0, ratio_y = 0, ratio_w = 1, ratio_h = 1,
             },
-            handler = function(ges)
-                return self:onSwipe(nil, ges)
-            end
+            handler = function(ges) return self:onSwipe(nil, ges) end,
         },
         {
             id = "paging_pan",
@@ -178,7 +176,7 @@ function ReaderPaging:setupTouchZones()
             screen_zone = {
                 ratio_x = 0, ratio_y = 0, ratio_w = 1, ratio_h = 1,
             },
-            handler = function(ges) return self:onPan(nil, ges) end
+            handler = function(ges) return self:onPan(nil, ges) end,
         },
         {
             id = "paging_pan_release",
@@ -186,7 +184,7 @@ function ReaderPaging:setupTouchZones()
             screen_zone = {
                 ratio_x = 0, ratio_y = 0, ratio_w = 1, ratio_h = 1,
             },
-            handler = function(ges) return self:onPanRelease(nil, ges) end
+            handler = function(ges) return self:onPanRelease(nil, ges) end,
         },
     })
 end
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -133,7 +133,7 @@ function ReaderPaging:setupTapTouchZones()
             ges = "tap",
             screen_zone = forward_zone,
             handler = function()
-                if G_reader_settings:readSetting("page_turns") ~= "swipe" then
+                if G_reader_settings:nilOrFalse("page_turns_disable_tap") then
                     return self:onGotoViewRel(1)
                 end
             end,
@@ -143,7 +143,7 @@ function ReaderPaging:setupTapTouchZones()
             ges = "tap",
             screen_zone = backward_zone,
             handler = function()
-                if G_reader_settings:readSetting("page_turns") ~= "swipe" then
+                if G_reader_settings:nilOrFalse("page_turns_disable_tap") then
                     return self:onGotoViewRel(-1)
                 end
             end,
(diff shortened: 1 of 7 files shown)


hius07:
I remember a user of Kindle Voyage (with hard buttons) asked to disable both taps and swipes.
It might be 2 checkboxes instead of the 3 menu items, providing 4 combinations
`Turn pages`
  ` [ ] with taps`
  ` [ ] with swipes`

NiLuJe:
@hius07: Yeah, that's great idea!

(Might be easier to implement by also splitting the actual setting into two bools, which would possibly make it clearer, too ^^).

frontend/apps/reader/modules/readerpaging.lua
```diff
@@ -159,7 +167,11 @@ function ReaderPaging:setupTouchZones()
             screen_zone = {
                 ratio_x = 0, ratio_y = 0, ratio_w = 1, ratio_h = 1,
             },
-            handler = function(ges) return self:onSwipe(nil, ges) end
+            handler = function(ges)
+                if G_reader_settings:readSetting("page_turns") ~= "tap" then
+                    return self:onSwipe(nil, ges)
+                end
+            end
```
NiLuJe, line 174: While I like the idea of nipping this directly in the registerTouchZone closure, this is problematic for the swipe handler, because it doesn't handle *only* page turns.

It's also responsible for the refresh swipe, plus the weird flip mode thingy in PDFs.

gbyl: I'm afraid I don't know how to solve that conflict. It sounds like the solution would be something way beyond my current knowledge. Any ideas?

NiLuJe: Move the check to the handler itself, and only trap west & east gestures?

gbyl: I keep getting crashes when I move the check and I can't figure out how to trap only west and east gestures.

NiLuJe: I pushed some commits in your branch that should do this ;).

gbyl: That looks a lot better than what I could have done on my own. Thanks a lot!

hius07:
Despite of the presence of the new `Page turns` menu in the file manager menu
https://github.com/koreader/koreader/blob/3cb95081851da3637288e41f9ce3c17c4d362173/frontend/ui/elements/filemanager_menu_order.lua#L93
it doesn't affect page turning with swipes in the file browser.
Should it?
If yes, should it be a common setting for the reader and for the file browser, or separate settings?
`[ ] Disable taps in reader`
`[ ] Disable swipes in reader`
`[ ] Disable swipes in file browser`

hius07:
Going further, should it affect swiping in
https://github.com/koreader/koreader/blob/3cb95081851da3637288e41f9ce3c17c4d362173/frontend/ui/widget/menu.lua#L1319
https://github.com/koreader/koreader/blob/3cb95081851da3637288e41f9ce3c17c4d362173/frontend/ui/widget/keyvaluepage.lua#L670
Are there any swipe haters who wants to disable page turning with swipes everywhere?

poire-z:
> Are there any swipe haters who wants to disable page turning with swipes everywhere?

I'm not fond of these recent additions of many little tunings (frankly, if you don't like swipes, just don't do any swipe...) so I'd rather keep these a-bit-crazy additions minimal.

If we disable swipes in FileManager and lists (KeyValuePage), what's left for navigation are the chevrons in the footer. I'm not sure many people would want just that.
I can understand it to be for Reader, for nervous people who can't help having their fingers play on the screen while reading/thinking. In FileManager, they shouldn't spend much nervous time on it :)

poire-z:
> I'm not fond of these recent additions of many little tunings (frankly, if you don't like swipes, just don't do any swipe...)

Correcting this. Actually, disaling swipe or tap and keeping the others have some (other than nervous gesturing) use cases:
- with swipes disabled, Links> Follow nearest & Swipe to go back get less confusing (without swipe disabled), they would do different things whether there is link or jump history
- with tap disabled, possible more accurate tap to follow link

Possible issue, if anyone can confirm: disabled Links> Swipe stuff, and Page turns tap disabled, so only Swipes to change page: I can swipe west only once to to to next page, next swipes west do nothing. No issue with swipe east to go back pages.

poire-z:
So, in summary: disabling swipe or tap in Reader have some use cases.
These shouldn't leak into FileBrowser & lists, where disabling any of them gets no real use case.
So, reader only - and if it would really make sonse sense, independant settings for FileBrowser and lists.

hius07:
> it might be named "Book pages turn"

And moved to Navigation?

poire-z:
> And moved to Navigation?

Dunno, it feels right in Taps & Gestures :)

It's the other two, Navigation & Document, that I find a bit clumsy/random.
Document has some reasoning, and stuff that we couldn't put anywhere else.
I'd rather put the Navigation stuff in a `Tap & Gestures & Buttons>` menu, so we get a slot free in the Gear menu for other stuff. And you could put Invert page turn in there.

NiLuJe:
It's (somewhat) on purpose:

* With the menu, you *have* to leave at least one of swipe or tap.
* For page-turns, you can drop *both* of 'em (leaving physical buttons).

PR #7574, opened 2021-04-20
UI: Add one-finger swipes to gesture manager

gbyl:
https://github.com/koreader/koreader/issues/6741 (first request)

idea endorsed [here](https://github.com/koreader/koreader/issues/6741#issuecomment-771936704) and [here](https://github.com/koreader/koreader/issues/6741#issuecomment-771915636)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7574)


plugins/gestures.koplugin/main.lua
```diff
@@ -41,6 +41,8 @@ local gestures_list = {
     hold_top_right_corner = _("Top right"),
     hold_bottom_left_corner = _("Bottom left"),
     hold_bottom_right_corner = _("Bottom right"),
+    one_finger_swipe_down = _("Down"),
```
Frenzie, line 44: Possibly a candidate for a context (something like `"Swipe"`).

Frenzie (approved):
lgtm

plugins/gestures.koplugin/main.lua
gbyl: I also thought that just one word would look a bit odd. But then the enclosing menu item `One-finger swipe` makes any extra words redundant.

Frenzie: Pardon, I meant `C_()`, not within the phrase itself.

http://koreader.rocks/doc/modules/gettext.html#pgettext

gbyl: I see, another possibility to avoid the ambiguity altogether is to use an icon as is done for the two-finger swipes.

Frenzie: The ambiguity is for translation, not for users.

Frenzie: Screenshot_2021-04-20-21-51-40-50_1aff61e6bd981aaa4ba9d0245eb6c04f

plugins/gestures.koplugin/main.lua
```diff
@@ -41,6 +41,8 @@ local gestures_list = {
     hold_top_right_corner = _("Top right"),
     hold_bottom_left_corner = _("Bottom left"),
     hold_bottom_right_corner = _("Bottom right"),
+    one_finger_swipe_down = _("⇓"),
+    one_finger_swipe_up = _("⇑"),
```
Frenzie, line 45:
```suggestion
    one_finger_swipe_down = "⇓",
    one_finger_swipe_up = "⇑",
```

plugins/gestures.koplugin/main.lua
gbyl: that's definitely better since the diagonal one-finger swipes could also be added in the future, right?

plugins/gestures.koplugin/defaults.lua
```diff
@@ -76,6 +84,14 @@ return {
         hold_top_right_corner = nil,
         hold_bottom_left_corner = nil,
         hold_bottom_right_corner = nil,
+        one_finger_swipe_east = nil,
+        one_finger_swipe_west = nil,
```
gbyl, line 88: The default values here should be:

```lua
one_finger_swipe_east = {page_jmp = -1,},
one_finger_swipe_west = {page_jmp = 1,},
```

But that creates an extra page forward/back on top of what reading rolling/paging already does. Any ideas as to how to deal with this?

NiLuJe: Probably much like the full-refresh itself: by getting rid of the RR/RP handler.

plugins/gestures.koplugin/defaults.lua
```diff
@@ -76,6 +84,14 @@ return {
...
+        one_finger_swipe_south = {full_refresh = true,},
+        one_finger_swipe_north = {full_refresh = true,},
+        one_finger_swipe_northeast = nil,
+        one_finger_swipe_northwest = nil,
+        one_finger_swipe_southeast = nil,
+        one_finger_swipe_southwest = nil,
```
NiLuJe, line 94: Speaking of the full refresh, that's the other way around: it was intended to be bound to *any* diagonal swipe.

gbyl: Ok, I'll switch the defaults. I guess I got confused by [this](https://github.com/koreader/koreader/issues/6741#issuecomment-771851683). But users can configure it later anyway. We just want to make it the least disruptive possible.

yparitcher:
Sorry for not helping with this and #6741 & #7572. I have been very busy (got married in march).
this is going to be more complicated than it looks, every gesture handled by koreader has an override system so higher priority gestures take precedence. this can be seen sprinkled around the different modules and here: https://github.com/koreader/koreader/blob/14f9f9d3fcd7a8bd6664bd2dba641efe227f3529/plugins/gestures.koplugin/main.lua#L825

The document's swipe handler takes precedence so it eats the gesture and processes it, adding it to the gesture plugin won't help. the document swipe handler is needed to handle page flipping, so it can't be easily gotten rid of.

gbyl:
@yparitcher That's great news. Congratulations!

> every gesture handled by koreader has an override system so higher priority gestures take precedence. this can be seen sprinkled around the different modules

Just when I thought I had solved the riddle! I guess at this point I have reached the limit of my understanding of koreader. Do we have to start over or can this PR still be salvaged?

plugins/gestures.koplugin/defaults.lua
gbyl: Just realized this PR now definitely conflicts with https://github.com/koreader/koreader/pull/7572 since `Disable swipes` is rendered redundant by the ability to set it to `Nothing` in the Gesture Manager. At the same time though, the page turns behavior should probably be independent from plugins. No idea what would be best.

gbyl:
@NiLuJe I am unsure as to how to go forward with this PR. One option would be to _not_ make the horizontal swipes customizable and hope the rest does not create other conflicts. The original goal was to just have a way to switch off the fallback refresh, so that would be enough for my use case. But there may be a better way to go about it. Would you mind sharing your opinion?

NiLuJe:
As @yparitcher said, this may interact with *a lot* of custom Swipe handlers, so, I'm a bit wary of touching this, quite frankly. (e.h., in DictQuickLookup, perhabps directly from Scroll*Widget/(Text/Html)BoxWidget; has a custom one that will only refresh the *text* on diagonal swipes).

Also, I'm not a swiper, so this isn't something I'm really willing to spend much time testing/investigating ;).

gbyl:
Sounds good. This definitely proved to be more intricate than expected. As always, I highly appreciate your input!

Issue #7577, opened 2021-04-21
Unsupported documents: do not add to History


hius07:
Currently there are two ways to add an unsupported document to the History list:
(1) File manager - long-press - Open with... - Cool Reader engine (this invokes ReaderUI)
(2) By changing the extension of a supported document which is already in the History, via File manager - Rename file

Any unsupported document cannot be opened from the History.
Moreover, trying to open the unsupported document via `main menu - Previous` causes KOReader to quit.

These options should be blocked by checking `DocumentRegistry:getProviders(file) ~= nil`.
I wonder where are the best places to check and block?

**(1)** When opening or renaming
https://github.com/koreader/koreader/blob/ea3fa5c2c7ed7330e8e1e6ef838b780e7998ea51/frontend/apps/reader/readerui.lua#L577
https://github.com/koreader/koreader/blob/14f9f9d3fcd7a8bd6664bd2dba641efe227f3529/frontend/apps/filemanager/filemanager.lua#L984

**(2)** Or directly in the readhistory
https://github.com/koreader/koreader/blob/14f9f9d3fcd7a8bd6664bd2dba641efe227f3529/frontend/readhistory.lua#L269
https://github.com/koreader/koreader/blob/14f9f9d3fcd7a8bd6664bd2dba641efe227f3529/frontend/readhistory.lua#L240

Frenzie:
"Blocking" seems to suggest a very particular solution that may be suboptimal?

hius07:
Do you mean, let History contain unsupported documents?

Frenzie:
I don't know if it currently does so by someone's explicit design or merely as an accident, but not quite, no.

What I mean is that blocking implies denying an attempt to add something to history at some point. The obvious alternative — but perhaps not obvious within the code? — would be to simply avoid trying to add anything until after success.

But now having taken a quick peek, the code already contains this:
https://github.com/koreader/koreader/blob/ea3fa5c2c7ed7330e8e1e6ef838b780e7998ea51/frontend/apps/reader/readerui.lua#L723-L747

Are you saying that's not working?

hius07:
When we open an unsupported document via Open with... dialog, there is no failure.
`one_time_providers` is used to open the document.
https://github.com/koreader/koreader/blob/ea3fa5c2c7ed7330e8e1e6ef838b780e7998ea51/frontend/ui/widget/filechooser.lua#L524-L532

Frenzie:
[...]

PR #7578, opened 2021-04-21
Always show open with dialog

Frenzie:
`DocumentRegistry:getProviders(file) == nil or #(DocumentRegistry:getProviders(file)) > 1` is just a very verbose way of saying "always".

References <https://github.com/koreader/koreader/issues/7577>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7578)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -381,7 +381,7 @@ function FileManager:setupLayout()
             table.insert(buttons, {
                 {
                     text = _("Open with…"),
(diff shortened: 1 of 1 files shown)


yparitcher:
the current test is false when `#(DocumentRegistry:getProviders(file)) == 1`

Frenzie:
This is indeed not quite the right way to go about it, but it seems like discriminating against `== 1` is no longer relevant since the `== nil` was added. It just makes me want to open djvu in crengine. :-P

PR #3935, opened 2018-05-07
Always close filemanger before open document

robert00s:
See comments:
https://github.com/koreader/koreader/pull/3933#issuecomment-386879892
https://github.com/koreader/koreader/pull/3933#issuecomment-386899058

PR #3933, opened 2018-05-06
Close filemanger after open last document

robert00s:
Close: #3310 
Similar to that: https://github.com/koreader/koreader/blob/e83a0c107843e6577fd308cdab5a7ee3ecaba5dc/frontend/apps/filemanager/filemanager.lua#L170-L171 after open last document (from menu) we should also close filemanager instance.

This change fix problem: #3310 (Menu with plugin isn't synchronized)

Issue #3310, opened 2017-10-06
Menu with plugin isn't synchronized


robert00s:
* KOReader version: latest
* Device: all

#### Issue
Plugin menu in filemanger and reader menu isn't synchronized.

#### Steps to reproduce
Open KO in filemanager mode. Go to menu -> Reader Statistics -> uncheck Enabled (or stay unchecked) -> Open any book -> Go to menu -> Reader Statistics -> check Enabled -> Go to Filemanager -> Open menu -> Reader Statistics and you see unchecked Enabled but you should see checked.
Similar with options Evernote -> Export to local HTML files or Evernote -> Export to local clipping text file. 
Reaer Timer is also not synchronized.

PR #7580, opened 2021-04-22
do not exit when opening unsupported file by last for FM

yparitcher:
fixes: #7577

see prior art in #3935 #3933

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7580)


Frenzie (commented):
Wouldn't that reintroduce https://github.com/koreader/koreader/issues/3310?

yparitcher:
no as since #3935 if the document is sucessfully opened the FM will be closed in ReaderUI, rendering closing it here as introduced in #3933 not needed as mention in those PRs

Issue #7581, opened 2021-04-22
Update failed – but it seems to have been successful


tastytea:
* KOReader version: **2021.04**, **2021.03**
* Device: **PB632** (PocketBook Touch HD 3)

#### Issue

After updating from 2021.03 to 2021.04, I hit the restart button in the dialog box. KOReader restarted, but reported that the update failed. It shows the new version in the help menu. I had the same issue when updating from 2021.02 to 2021.03 but didn't report it at the time.

#### Steps to reproduce

1. Update from 2021.03 to 2021.04
2. Restart KOReader

##### `crash.log` (if applicable)

[crash.log](https://github.com/koreader/koreader/files/6357174/crash.log)

These lines in the log are new after the update:

```
04/22/21-11:25:24 INFO  ./cache/fontlist/fontinfo.dat [string "./cache/fontlist/fontinfo.dat"]:1: unexpected symbol near '.' initializing it
04/22/21-11:25:26 INFO  Performing one-time migration for 20210306 (1/2)
04/22/21-11:25:26 INFO  Performing one-time migration for 20210306 (2/2)
04/22/21-11:25:26 INFO  Performing one-time migration for 20210330
ffi.load: sqlite3
04/22/21-11:25:26 INFO  Performing one-time migration for 20210404
04/22/21-11:25:26 INFO  Performing one-time migration for 20210409
04/22/21-11:25:26 INFO  Performing one-time migration for 20210412
04/22/21-11:25:26 WARN  os.rename: ./cache/calibre-libraries.lua: No such file or directory
04/22/21-11:25:26 WARN  os.rename: ./cache/calibre-books.dat: No such file or directory
04/22/21-11:25:26 INFO  Performing one-time migration for 20210414
04/22/21-11:25:26 WARN  os.remove: ./cache/calibre/books.dat: No such file or directory
```

Frenzie:
> It shows the new version in the help menu.

Strictly speaking that does't mean the update succeeded, just that the update to the one specific file with the version string succeeded.

Is this the specific error message you're referring to btw? That is, after clicking "restart".

https://github.com/koreader/koreader/blob/ea3fa5c2c7ed7330e8e1e6ef838b780e7998ea51/platform/pocketbook/koreader.app#L35-L46

LucyOne:
I have been getting the "Update failed" message on all my Pocketbooks since the update code started being displayed. The updates are still successful and everything works as it should.

tastytea:
[...]

PR #7585, opened 2021-04-22
PocketBook: Workaround platform shenanigans when aplying an OTA

NiLuJe:
"Fix" #7581

Thanks to @tastytea for the tests ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7585)


NiLuJe:
Note: There aren't any actual utime errors on PB, probably because vfat or something.

PR #7589, opened 2021-04-24
Remove notification popups when font face and size are changed

### LINKED ISSUE #7502: Menu is drawn improperly after font selection
* KOReader version: 2021.03
* Device: KPW3

#### Issue
After selecting a font in a fb2.zip book the top menu is redrawn, but only partially. Otherwise, it acts as a usual top menu, so if I tap outside of it, it closes; and if I blindly tap a menu item, it opens properly.

mergen3107:
Fixes #7502

Let me know if instead of commenting out I should just remove the lines

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7589)

### DIFF
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -221,9 +221,10 @@ function ReaderFont:onSetFontSize(new_size)
     self.font_size = new_size
     self.ui.document:setFontSize(Screen:scaleBySize(new_size))
     self.ui:handleEvent(Event:new("UpdatePos"))
-    UIManager:show(Notification:new{
-        text = T( _("Font size set to %1."), self.font_size),
-    })
+    -- avoid refresh competition and messup
+    -- UIManager:show(Notification:new{
+    --     text = T( _("Font size set to %1."), self.font_size),
+    -- })
 
     return true
 end
@@ -300,9 +301,10 @@ end
 function ReaderFont:setFont(face)
     if face and self.font_face ~= face then
         self.font_face = face
-        UIManager:show(Notification:new{
-            text = T( _("Redrawing with font %1."), face),
-        })
+        -- avoid refresh competition and messup
+        -- UIManager:show(Notification:new{
+        --     text = T( _("Redrawing with font %1."), face),
+        -- })
 
         self.ui.document:setFontFace(face)
         -- signal readerrolling to update pos in new height
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -221,7 +221,7 @@ function ReaderFont:onSetFontSize(new_size)
     self.font_size = new_size
     self.ui.document:setFontSize(Screen:scaleBySize(new_size))
     self.ui:handleEvent(Event:new("UpdatePos"))
-        UIManager:show(Notification:new{
+    UIManager:show(Notification:new{
         text = T( _("Font size set to %1."), self.font_size),
     })
     return true
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -221,6 +221,9 @@ function ReaderFont:onSetFontSize(new_size)
     self.font_size = new_size
     self.ui.document:setFontSize(Screen:scaleBySize(new_size))
     self.ui:handleEvent(Event:new("UpdatePos"))
+        UIManager:show(Notification:new{
+        text = T( _("Font size set to %1."), self.font_size),
+    })
     return true
 end
 
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -221,11 +221,6 @@ function ReaderFont:onSetFontSize(new_size)
     self.font_size = new_size
     self.ui.document:setFontSize(Screen:scaleBySize(new_size))
     self.ui:handleEvent(Event:new("UpdatePos"))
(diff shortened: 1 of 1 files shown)


hius07:
To reviewers: which style is recommended?
https://github.com/koreader/koreader/blob/e107aad42e06be6f7c861cad4824636cac67a4d7/frontend/apps/reader/modules/readerfont.lua#L224-L226
or
https://github.com/koreader/koreader/blob/e107aad42e06be6f7c861cad4824636cac67a4d7/frontend/apps/reader/modules/readerfont.lua#L358-L359

Frenzie:
The former for the example.

hius07:
Thanks, noted.

NiLuJe:
Definitely remove the code, that's what git blame is for ;).

(I mean, you can leave the comment explaining the lack of notif ^^).

mergen3107:
Thank you! I feel myself so behind this well-organized high tech community of contribution :D

poire-z:
I was a bit hesitant about removing the notification for font size changes, as this one (unlike changing font face) can be assigned to a gesture (as an increment or an absolute value), and the notification might be an interesting feedback about what the value is.
But why not, we'll see if gesture users complain...

Galunid:
@poire-z I'm gesture user who came here to complain :). I think font size is important notification to keep (it depends on the length(?) of the gesture, so sometimes you adjust the font to 15.5 other times to 16). In all fairness though, I don't do that very often.

zwim:
I would leave the notifications as a feedback what happened. 
On a slow device when you tap on a button it takes quite some time until the display is changed (depending on the book and the operation). Together with the change a notification what happend seems adequate (for my feeleings). 
If the user is used that there comes a notification, he realizes the change is done. Otherwise might get nervous and tap once again ....

I would go even further and add some more notifications -> see #7623

poire-z:
I'd rather not see any notification, than seeing more of them :/
And there's the issue that notifications can cause refreshes congestions and glitches, cf #7502.

You'not getting the feedback I described at https://github.com/koreader/koreader/issues/7502#issuecomment-813068069 : 
> And the font size panels stays open while this is happening, so it's showing the font size we end up with - and we have this little effect (I don't think it's voluntary :) that when changing font size, both he old and the new sizes are underlined while the rendering takes place - and when done, only tne new size is underlined.

Frenzie:
I think the reasoning in favor of (some) notifications is twofold:

1. There's often no clear indication that something's happening.
2. Gestures don't include a menu. (Essentially a stronger version of 1.)

But of course a notification needn't be the solution to that problem.

poire-z:
Ideally:
- no notification when tap in bottom menu, a notification when invoked from gesture/dispatcher/event
- some feedback in the bottom widgets similar to what happens with font size, ie. for toggles: on tap, make the one tapped black, while keeping the previous one black - and only change the previous one to white/disabled when done

NiLuJe:
@poire-z: Is there actually a widget left in ConfigDialog where the new state isn't immediately visible?

(IIRC, the toggles toggle, the progress stuff updates, etc.)

zwim:
I am not quite sure what you mean. 
If you tap on the bottom bar, the status is updated the more or less very fast (slower if a re-rendering happens). Increasing/Decreasing font size shows nothing in the menu.
When opening a selector with the three dots menu, you may (or may not) see a change when tapping apply. This is (esp. if you don't see a change in the document) annoying. Sometime I am not sure, if the Tolino has realized my tapping :/

poire-z:
> I am not quite sure what you mean.
If you tap on the bottom bar, the status is updated the more or less very fast (slower if a re-rendering happens). Increasing/Decreasing font size shows nothing in the menu.

I meant taping on the font size numbers (not the increase|decrease fine tuning toggle, which behave like the other toggles).

> When opening a selector with the three dots menu, you may (or may not) see a change when tapping apply. 

Well, the number picker is another thing to think about - I was thinking more about the main bottom widgets.
But anyway, in all cases, you know they have been applied when... the visual rendering of the book text has changed :) as most often, it will change. The notification would just help you be of help when nothing has changed, to tell you it has been applied but it had no effect - which may affect only Render mode, all the other would cause some visual change.
Do you really notice the Notification ? Do your eyes really jump at looking at it ? Mine don't, I look at the page for changes - the notification is just a glitch in my peripheral vision, and the time I need to react and go look at it, it might already be gone...

Also, regarding #7623: Font Hinting and Contrast apply immediately (no re-rendering needed, just a repaint of the page) unlike the others.

> Is there actually a widget left in ConfigDialog where the new state isn't immediately visible?

Not sure, but looks like most change their visual immediately, before the re-rendering.
Your Rotation icons behave like font size numbers (previous and new both underlined while re-rendering).

poire-z:
@mergen3107 : can you revert (= let it be shown as it was) the one about font size, and just have the Notification about the font family change removed - which should be enough to fix your refresh glitch.

Let notifications from the bottom menu be handled/cleaned/made better for another PR, cf the ongoing discussion in #7623.

mergen3107:
@poire-z 
OK, done :)

frontend/apps/reader/modules/readerfont.lua
```diff
@@ -221,10 +221,9 @@ function ReaderFont:onSetFontSize(new_size)
     self.font_size = new_size
     self.ui.document:setFontSize(Screen:scaleBySize(new_size))
     self.ui:handleEvent(Event:new("UpdatePos"))
-    UIManager:show(Notification:new{
+        UIManager:show(Notification:new{
```
poire-z, line 224:
```suggestion
    UIManager:show(Notification:new{
```

PR #7591, opened 2021-04-24
fix template

zwim:
Fix template, so that translations can be used (`color` is an internal string with the values "black", "white" and "gray").

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7591)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -441,7 +441,7 @@ or delete the path:
 - Clear the name of the file]])
 
 -- menu entry: Cache settings
-function CoverImage:menu_entry_cache()
+function CoverImage:menuEntryCache()
     return {
         text = _("Cache settings"),
         checked_func = function()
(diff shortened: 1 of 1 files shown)


plugins/coverimage.koplugin/main.lua
```diff
@@ -581,7 +581,7 @@ end
 
 function CoverImage:menu_entry_background(color)
     return {
-        text = T(_("Fit to screen, %1 background"), color),
+        text = T(_("Fit to screen, %1 background"), _(color)),
```
Frenzie, line 584: That won't work in that form, but you shouldn't do this regardless. Use three strings written out in full.

Frenzie: Also CoverImage:menu_entry_cache() doesn't abide by the style guide, should be menuEntryCache

Frenzie: That is, the call should look like this (except it shouldn't because it's being clever and Anglocentric).

self:menuEntryBackground(_("gray"))

Frenzie:
Thanks!

PR #7592, opened 2021-04-24
File manager: make "Refresh content" action customizable

hius07:
Holding the "Plus" button is reserved for content refresh in https://github.com/koreader/koreader/pull/7559.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7592)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -131,13 +131,6 @@ function FileManager:setupLayout()
         padding_right = Size.padding.large,
         padding_bottom = 0,
         callback = function() self:onShowPlusMenu() end,
-        hold_callback = function()
-            self:onRefresh()
-            UIManager:show(InfoMessage:new{
-                text = _("Content refreshed."),
-                timeout = 2,
-            })
-        end,
     }
 
     self.path_text = TextWidget:new{
@@ -1153,6 +1146,14 @@ end
 
 function FileManager:onHome()
     return self:goHome()
+end
+
+function FileManager:onRefreshContent()
+    self:onRefresh()
+    UIManager:show(InfoMessage:new{
+        text = _("Content refreshed."),
+        timeout = 2,
+    })
 end
 
 return FileManager
--- a/frontend/dispatcher.lua
+++ b/frontend/dispatcher.lua
@@ -90,6 +90,7 @@ local settingsList = {
     -- filemanager settings
     folder_up = { category="none", event="FolderUp", title=_("Folder up"), filemanager=true},
     show_plus_menu = { category="none", event="ShowPlusMenu", title=_("Show plus menu"), filemanager=true},
+    refresh_content = { category="none", event="RefreshContent", title=_("Refresh content"), filemanager=true},
     folder_shortcuts = { category="none", event="ShowFolderShortcutsDialog", title=_("Folder shortcuts"), filemanager=true, separator=true,},
 
     -- reader settings
@@ -216,6 +217,7 @@ local dispatcher_menu_order = {
     -- filemanager
     "folder_up",
     "show_plus_menu",
+    "refresh_content",
     "folder_shortcuts",
 
     -- reader
--- a/plugins/gestures.koplugin/defaults.lua
+++ b/plugins/gestures.koplugin/defaults.lua
@@ -8,7 +8,7 @@ return {
         tap_right_bottom_corner = nil,
         tap_left_bottom_corner = Device:hasFrontlight() and {toggle_frontlight = true,} or nil,
         hold_top_left_corner = nil,
-        hold_top_right_corner = nil,
+        hold_top_right_corner = {refresh_content = true,},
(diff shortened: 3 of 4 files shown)


Frenzie (changes requested):
That'd be a step backwards. Please assign it as the default action and leave it open for customization. ;-)

hius07:
Should https://github.com/koreader/koreader/pull/7559 be reverted?

Frenzie:
As a part of implementing it properly, yes, but if you mean asap then I don't see why. ^_^

hius07:
Right now in the gesture manager we can assign an action to the hold_top_right_corner, but it has no effect, doesn't work.

hius07:
Should we also free the hold_top_left_corner for customization?
We have "Set as HOME folder" in the Plus menu and in the folder context menu.

Frenzie:
> Should we also free the hold_top_left_corner for customization?

Yes, that's all WIP/TBD.

Frenzie:
But beware, the remainders are on purpose because they're often hard, cf. https://github.com/koreader/koreader/pull/6957

I'm primarily talking about the fact that nothing new of the sort should be added.

hius07:
https://github.com/koreader/koreader/pull/7559 reverted, action added to Dispatcher.

Frenzie:
Which migration?

hius07:
> Which migration?

https://github.com/koreader/koreader/blob/ea6576d2b6b25baf5c5f98293fa9e5e7115ed8c7/plugins/gestures.koplugin/migration.lua#L84-L85

Frenzie:
I'm not quite following — isn't this brand new?

hius07:
Yes, this is a brand new action, so migration isn't required, right?

Frenzie:
I shouldn't think so, pinging @NiLuJe just in case.

NiLuJe:
It wouldn't run on existing setups anyway (it's a one time migration), so, I wouldn't think so, no.

yparitcher:
No need for migration, it is only for old gesture code to plugin conversion.

yparitcher:
Was this tested? is the hold passed through to the gesture plugin? as @Frenzie mentioned see  #6957 for where this approach did not work.

hius07:
> Was this tested?

I assgned `refresh_content` to the `hold_top_right_corner` and it works.
I assigned other actions to the `hold_top_right_corner` and it works as well.

NiLuJe:
By "it", do you mean the Gesture, or the Button? IIRC, the issue was that it broke the Button's own handler (well, not break per-se, but it's consumed earlier and stopped).

yparitcher:
> By "it", do you mean the Gesture, or the Button? IIRC, the issue was that it broke the Button's own handler (well, not break per-se, but it's consumed earlier and stopped).

The issue was that it changing the tap zone prevented the gesture as the menu/button are higher precedence than the gesture. The menu only has a tap zone not a hold zone so this PR should be good.

yparitcher:
To make this the default you should probably change:
```diff --git a/plugins/gestures.koplugin/defaults.lua b/plugins/gestures.koplugin/defaults.lua
index 7f27f34c..4fa1273d 100644
--- a/plugins/gestures.koplugin/defaults.lua
+++ b/plugins/gestures.koplugin/defaults.lua
@@ -8,7 +8,7 @@ return {
         tap_right_bottom_corner = nil,
         tap_left_bottom_corner = Device:hasFrontlight() and {toggle_frontlight = true,} or nil,
         hold_top_left_corner = nil,
-        hold_top_right_corner = nil,
+        hold_top_right_corner = {refresh_content = true,},
         hold_bottom_left_corner = nil,
         hold_bottom_right_corner = nil,
         one_finger_swipe_left_edge_down = Device:hasFrontlight() and {decrease_frontlight = 0,} or nil,
```

hius07:
> By "it", do you mean the Gesture, or the Button?

I meant the gesture (hold). Tapping the Plus button works as well.

hius07:
May I ask another question concerning the Dispatcher.
We have an action
https://github.com/koreader/koreader/blob/86126d042f81e566006072ce310a5493cc403fae/frontend/dispatcher.lua#L62

But we have two different onHome events:
https://github.com/koreader/koreader/blob/86126d042f81e566006072ce310a5493cc403fae/frontend/apps/reader/readerui.lua#L749-L753
https://github.com/koreader/koreader/blob/86126d042f81e566006072ce310a5493cc403fae/frontend/apps/filemanager/filemanager.lua#L1154-L1156
Is it correct? That causes that when we invoke the action named "File browser" being in the file browser, we go to the home folder, which is not an issue but unexpectable.
When we invoke the action being in reader, file browser starts in the folder with the book.

yparitcher (approved):

NiLuJe:
I mean, if you set a "Go to FM" action while *inside* the FM, you sorta expect stuff to do something else ;).

hius07:
This PR is finished and tested. Making https://github.com/koreader/koreader/issues/7477 more flexible.

PR #7595, opened 2021-04-24
UI: Add toggle for screen refresh swipes

gbyl:
https://github.com/koreader/koreader/issues/6741 (first request)

much less ambitious (but conflict-free) version of https://github.com/koreader/koreader/pull/7574

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7595)

### DIFF
--- a/frontend/apps/reader/modules/readerpaging.lua
+++ b/frontend/apps/reader/modules/readerpaging.lua
@@ -444,8 +444,10 @@ function ReaderPaging:onSwipe(_, ges)
     else
         -- update footer (time & battery)
         self.view.footer:onUpdateFooter()
-        -- trigger full refresh
-        UIManager:setDirty(nil, "full")
+        if G_reader_settings:nilOrFalse("disable_refresh_swipes") then
+            -- trigger full refresh
+            UIManager:setDirty(nil, "full")
+        end
     end
 end
 
--- a/frontend/apps/reader/modules/readerrolling.lua
+++ b/frontend/apps/reader/modules/readerrolling.lua
@@ -535,8 +535,10 @@ function ReaderRolling:onSwipe(_, ges)
     else
         -- update footer (time & battery)
         self.view.footer:onUpdateFooter()
-        -- trigger full refresh
-        UIManager:setDirty(nil, "full")
+        if G_reader_settings:nilOrFalse("disable_refresh_swipes") then
+            -- trigger full refresh
+            UIManager:setDirty(nil, "full")
+        end
     end
 end
 
--- a/frontend/ui/elements/common_settings_menu_table.lua
+++ b/frontend/ui/elements/common_settings_menu_table.lua
@@ -200,6 +200,7 @@ common_settings.screen_dpi = require("ui/elements/screen_dpi_menu_table")
 common_settings.screen_eink_opt = require("ui/elements/screen_eink_opt_menu_table")
 common_settings.menu_activate = require("ui/elements/menu_activate")
 common_settings.page_turns = require("ui/elements/page_turns")
+common_settings.disable_refresh_swipes = require("ui/elements/disable_refresh_swipes")
 common_settings.screen_disable_double_tab = require("ui/elements/screen_disable_double_tap_table")
 common_settings.ignore_hold_corners = {
     text = _("Ignore hold on corners"),
--- a/frontend/ui/elements/disable_refresh_swipes.lua
+++ b/frontend/ui/elements/disable_refresh_swipes.lua
@@ -2,6 +2,7 @@ local _ = require("gettext")
 
 return {
     text = _("Disable refresh swipes"),
+    help_text = _("Disables fullscreen refreshes triggered by non-horizontal swipes. This setting applies only to in-reader swipes."),
     checked_func = function()
         return G_reader_settings:isTrue("disable_refresh_swipes")
     end,
--- a/frontend/ui/elements/disable_refresh_swipes.lua
+++ b/frontend/ui/elements/disable_refresh_swipes.lua
@@ -1,0 +1,11 @@
+local _ = require("gettext")
+
+return {
+    text = _("Disable refresh swipes"),
+    checked_func = function()
+        return G_reader_settings:isTrue("disable_refresh_swipes")
+    end,
+    callback = function()
+        G_reader_settings:toggle("disable_refresh_swipes")
+    end
+}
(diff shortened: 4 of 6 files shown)


NiLuJe:
There are a few others hidden here and there in a few widgets, FWIW.

So, either hunt them all down, or simply add a help message saying that this only applies to the main in-reader ones.

NiLuJe:
Mildly curious: What's the rationale for wanting to disable those?

gbyl:
@NiLuJe It is a combination of [this](https://github.com/koreader/koreader/issues/6741#issuecomment-771898241) and the fact that a refresh is already triggered every 5 page turns. So, it feels a tad redundant. Also, UX-wise it feels unexpected, hence why I originally mistook it for a bug — when it was a feature all along.

poire-z:
> It is a combination of _this_ and the fact that a refresh is already triggered every 5 page turns
> _this_ = I use the two-finger gestures often and ocassionally when the engine fails to detect the input it thinks I am using only one finger and then it triggers a refresh.

Well, I'm not fond of exposing every user's little annoyance tuning thru a menu item :) as this one feels really a _little_ annoyance: if you already get a refresh everyt 5 page turns and that does not bother you, how really annoying is this additional refresh _ocassionally_ happening ? :)
I understand it's bothering you if you still want that 6 months after your first issue, so...
@NiLuJe : does the behaviour with the input code failing detecting 2 fingers feels like something that could be fixed (as you played recently with input & timeouts :) ? Or is it just @gbyl 's fingers that need more training ?

Frenzie:
I'm not sure why this exists next to short diagonal swipe. Maybe just remove it altogether?

NiLuJe:
@poire-z: I'm not a swiper, so, hard to judge, at worst I seem to trip a two-finger tap instead of a two-finger swipe.

In any case, I don't recall ever having inadvertently tripped the diagonal swipe. But, again, not a swiper ^^/

NiLuJe:
@Frenzie: I'm not entirely clear on what you're referring to (either with "this" or "it ^^).

poire-z:
I think that's the way "short diagonal swipe" is handled: in the onSwipe handler, we process the directions that make sense (west, east), and then fallback to refresh (so, north, south, diagonal).
We indeed could process north and south as no-op, so only diagonal does refresh.
But some users (including me) may have gotten used to short vertical swipes also doing refresh.

I could be fine with no-op'ing north and south swipes to make @gbyl happy - if that's really enough to make you happy (diagonal refresh would still do refresh).
But the day we assign another action to vertical swipes, these missed 2-fingers swipes will then also trigger these new things...

Frenzie:
@NiLuJe The subject of this thread is vertical swipe actions. Apologies for being unclear with pronouns. Vertical swipe and diagonal swipe do the same thing, and I was under the impression that diagonal swipe was the canonical one.

I'm with @poire-z; I'm not in favor of this PR, since it's just adding complexity. But I'm also not attached to this particular behavior.

NiLuJe:
Oh, yeah. We've all canonically referred to it as "short diagonal swipes", but it's indeed implemented as a "fallback" swipe in the handlers.

----

Re: gesture detection in general: it's entirely possible that some things could be made more accurate, but that way lies maths, so that's a pass from me ;p.

(e.g., I attempted to play with pinch and spread when testing the input stuff, and it feels super flakey; I don't remember exactly how I tested it, but something like only moving a single finger made it slightly better or something).

gbyl:
Gesture detection is already excellent, so no need to work on that front @NiLuJe. The rationale of this PR boils down to _'This `full-screen refresh swipe/fallback swipe` feels like something the user should be able to customize'_. That being said, I agree with the sentiment that this PR is too arbitrary and adds unnecessary complexity.

The best solution remains adding one-finger gestures to the gesture manager. A second-best is @Frenzie's suggestion:

> I'm not sure why this exists next to short diagonal swipe. Maybe just remove it altogether?

Which, if I'm interpreting it correctly, involves just removing [this](https://github.com/koreader/koreader/blob/9ef435c97af5b570f2fc4cac27bc7bbd61a5a3ff/frontend/apps/reader/modules/readerrolling.lua#L538-L539) and [this](https://github.com/koreader/koreader/blob/285fc75aa7830eeb6d0a02e35dff763f758e0d87/frontend/apps/reader/modules/readerpaging.lua#L433-L434). Diagonal swipes will still refresh the screen (since that is the default value of 'short diagonal swipe'), but with the added benefit that they are already part of the UI and therefore customizable.

poire-z:
Oh, indeed, short diagonal swipe is handled elsewhere from these bits of code - and the dispatcher action that does the full refresh is DeviceListener:onFullRefresh(). I guess the other part in these bits of code should also be put in DeviceListener:onFullRefresh() (via en event, so it does not crash in Filebrowser):
```lua
-- update footer (time & battery)
self.view.footer:onUpdateFooter()
```
That would make it indeed a bit more coherent.

If we don't want to disappoint people who got used to refresh with small vertical swipes, I guess it could be made a new gesture.
(My success rate with my clumsy diagonal swipes got a bit worse once I commented out the suggested bits.)

poire-z:
@NiLuJe : considering the differences between these:
https://github.com/koreader/koreader/blob/9f835156d16da2f958944a7f2326ee1cc9b47af3/frontend/device/devicelistener.lua#L358-L361
https://github.com/koreader/koreader/blob/9f835156d16da2f958944a7f2326ee1cc9b47af3/frontend/apps/reader/modules/readerrolling.lua#L535-L540
which both do not provide `force_repaint=true` to onUpdateFooter - but one setDirtys `nil` and the other `"all"`.
The devicelistener one (diagonal swipe) does update the time in ther footer, while the readerolling one (vertical swipe) does not (even if the updateFooter line was actually added to get that :/)

What combinations are best (in all/which contexts) ?
`self.ui:handleEvent(Event:new("UpdateFooter", true))` vs `self.ui:handleEvent(Event:new("UpdateFooter"))`
with
`UIManager:setDirty("all", "full")`  vs `UIManager:setDirty(nil, "full")`

NiLuJe:
Definitely not `"all"`, that causes a repaint of `everything`, while we only want a `refresh`.

So, ask a repaint from UpdateFooter, and enqueue a simple refresh should be

```
self.ui:handleEvent(Event:new("UpdateFooter", true))
UIManager:setDirty(nil, "full")
```

No other combination really makes sense, unless they're trying to do something else ;).

NiLuJe:
The cherry on top is only actually repaint the footer when it's visible, so, something like

`self.ui:handleEvent(Event:new("UpdateFooter", self.ui.view.footer_visible))`

poire-z:
Full refreshes on vertical swipes removed in #7646.
And given that vertical swipes can now be used for scrolling, in continuous mode - and have no effect in pge mode, no need to think about adding them to gesture manager.

PR #7598, opened 2021-04-25
Move to archive on end

grasegger:
Followup to #6164 

This Pull Request introduces:

* A way to add new action to the popup that opens after a document end is reached
* The "move to archive" plugin using that way to provide a quick way to move files to the archive
* Fixed a bug where a old state of the movetoarchive settings was held in memory
* A setting to "move to archive" that allows you to skip the popup after a move
* A setting for the default action when movetoarchive shows no popup

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7598)

### DIFF
--- a/frontend/apps/reader/modules/readerstatus.lua
+++ b/frontend/apps/reader/modules/readerstatus.lua
@@ -18,7 +18,13 @@ local ReaderStatus = InputContainer:new {
         modified = "",
     },
     enabled = true,
-    total_pages = 0
+    total_pages = 0,
+
+    -- Using table.insert plugins can add their actions to self.ui.status.additional_actions
+    -- Example: {title = "title", callback = callback}
+    -- All callbacks get the file.
+    additional_actions = {
+    }
 }
 
 function ReaderStatus:init()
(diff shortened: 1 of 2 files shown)


NiLuJe:
I'd be inclined to drop all the ugliness from readerstatus, add an event handler in the plugin, and send an event, and plugin or no plugin loaded be damned.

(That's a rather strong veto on the whole readerstatus shenanigans there ^^).

grasegger:
@NiLuJe I get what you are saying, but that definitely is out of scope for the PR for me. I am happy to take on a bug report that formalises all of this in another PR though. (Just can't make promises about a delivery date ;) )

grasegger:
Ok, so I figured out how to place the action in the "default actions after end", but I can't figure out how to load the plugin there ... 

Also that made me painfully aware of the lack of a proper api ... :D But for now it's possible to use it via the popup, maybe I should work on the automatic end stuff in another PR, what do you think?

grasegger:
I didn't mean to take credit for @mwoz123's work, but my rebase knowledge made me somehow coauthor. Hope that's ok.

NiLuJe:
I can whip up the PluginLoader bit if that's the holdup, because I'm extremely not a fan of this approach ;).

grasegger:
> I can whip up the PluginLoader bit if that's the holdup, because I'm extremely not a fan of this approach ;).

That would be fantastic.

NiLuJe:
c.f., https://github.com/koreader/koreader/pull/7607#issuecomment-826986379 for a cleaner way forward.

grasegger:
@NiLuJe Nice! I will open another PR with the things I added on top of that. Thank you so much for you work!!!!

sorenpeter:
Was this ever implemented? I can't seem to find the option to "Move to Archive plugin" for the "end-menu"...

PR #7607, opened 2021-04-26
PluginLoader: Allow querying whether a specific plugin is loaded

NiLuJe:
(instantiated, actually).

Re #7598

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7607)

### DIFF
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -760,7 +760,7 @@ end
 
 function FileManager:onClose()
     logger.dbg("close filemanager")
-    PluginLoader:dtor()
+    PluginLoader:finalize()
     self:handleEvent(Event:new("SaveSettings"))
     G_reader_settings:flush()
     UIManager:close(self)
--- a/frontend/apps/filemanager/filemanager.lua
+++ b/frontend/apps/filemanager/filemanager.lua
@@ -760,6 +760,7 @@ end
 
 function FileManager:onClose()
     logger.dbg("close filemanager")
+    PluginLoader:dtor()
     self:handleEvent(Event:new("SaveSettings"))
     G_reader_settings:flush()
     UIManager:close(self)
--- a/frontend/apps/reader/readerui.lua
+++ b/frontend/apps/reader/readerui.lua
@@ -699,7 +699,7 @@ end
 
 function ReaderUI:onClose(full_refresh)
     logger.dbg("closing reader")
-    PluginLoader:dtor()
+    PluginLoader:finalize()
     Device:notifyBookState(nil, nil)
     if full_refresh == nil then
         full_refresh = true
--- a/frontend/apps/reader/readerui.lua
+++ b/frontend/apps/reader/readerui.lua
@@ -699,6 +699,7 @@ end
 
 function ReaderUI:onClose(full_refresh)
     logger.dbg("closing reader")
+    PluginLoader:dtor()
     Device:notifyBookState(nil, nil)
     if full_refresh == nil then
         full_refresh = true
--- a/frontend/pluginloader.lua
+++ b/frontend/pluginloader.lua
@@ -40,6 +40,10 @@ end
 
 local PluginLoader = {
     show_info = true,
+    enabled_plugins = nil,
+    disabled_plugins = nil,
+    loaded_plugins = nil,
+    all_plugins = nil,
 }
 
 function PluginLoader:loadPlugins()
@@ -136,26 +140,26 @@ function PluginLoader:loadPlugins()
 end
 
 function PluginLoader:genPluginManagerSubItem()
-    local enabled_plugins, disabled_plugins = {}, {}
-    if self.all_plugins == nil then
+    if not self.all_plugins then
+        local enabled_plugins, disabled_plugins = self:loadPlugins()
         self.all_plugins = {}
-        enabled_plugins, disabled_plugins = self:loadPlugins()
-    end
 
-    for _, plugin in ipairs(enabled_plugins) do
-        local element = getMenuTable(plugin)
-        element.enable = true
-        table.insert(self.all_plugins, element)
-    end
-
-    for _, plugin in ipairs(disabled_plugins) do
-        local element = getMenuTable(plugin)
-        element.enable = false
-        if not OBSOLETE_PLUGINS[element.name] then
+        for _, plugin in ipairs(enabled_plugins) do
+            local element = getMenuTable(plugin)
+            element.enable = true
             table.insert(self.all_plugins, element)
         end
+
+        for _, plugin in ipairs(disabled_plugins) do
+            local element = getMenuTable(plugin)
+            element.enable = false
+            if not OBSOLETE_PLUGINS[element.name] then
+                table.insert(self.all_plugins, element)
+            end
+        end
+
+        table.sort(self.all_plugins, function(v1, v2) return v1.fullname < v2.fullname end)
     end
-    table.sort(self.all_plugins, function(v1, v2) return v1.fullname < v2.fullname end)
 
     local plugin_table = {}
     for __, plugin in ipairs(self.all_plugins) do
@@ -192,7 +196,7 @@ end
 function PluginLoader:createPluginInstance(plugin, attr)
     local ok, re = pcall(plugin.new, plugin, attr)
     if ok then  -- re is a plugin instance
-        self.loaded_plugins[plugin.name] = true
+        self.loaded_plugins[plugin.name] = re
         return ok, re
     else  -- re is the error message
         logger.err("Failed to initialize", plugin.name, "plugin: ", re)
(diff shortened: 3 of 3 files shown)


NiLuJe:
Meaning #7598 can then essentially boil down to something like

```diff
diff --git a/frontend/apps/reader/modules/readerstatus.lua b/frontend/apps/reader/modules/readerstatus.lua
index b83b522c9..28f73f185 100644
--- a/frontend/apps/reader/modules/readerstatus.lua
+++ b/frontend/apps/reader/modules/readerstatus.lua
@@ -5,6 +5,7 @@ local Device = require("device")
 local Event = require("ui/event")
 local InfoMessage = require("ui/widget/infomessage")
 local InputContainer = require("ui/widget/container/inputcontainer")
+local PluginLoader = require("pluginloader")
 local UIManager = require("ui/uimanager")
 local _ = require("gettext")
 local T = require("ffi/util").template
@@ -118,6 +119,24 @@ function ReaderStatus:onEndOfBook()
                     end,
                 },
             },
+        }
+        -- Optional Plugin stuff, each get their own line
+        if PluginLoader:isPluginLoaded("movetoarchive") then
+            table.insert(buttons,
+                {
+                    {
+                        text = _("Archive"),
+                        callback = function()
+                            self.ui:handleEvent(Event:new("MoveToArchive", self.document.file))
+                            UIManager:close(choose_action)
+                        end,
+                    },
+                }
+            )
+        end
+
+        -- ALways end w/ Cancel
+        table.insert(buttons,
             {
                 {
                     text = _("Cancel"),
@@ -125,8 +144,9 @@ function ReaderStatus:onEndOfBook()
                         UIManager:close(choose_action)
                     end,
                 },
-            },
-        }
+            }
+        )
+
         choose_action = ButtonDialogTitle:new{
             name = "end_document",
             title = _("You've reached the end of the document.\nWhat would you like to do?"),
diff --git a/plugins/movetoarchive.koplugin/main.lua b/plugins/movetoarchive.koplugin/main.lua
index 9d59603fb..94f04ecae 100644
--- a/plugins/movetoarchive.koplugin/main.lua
+++ b/plugins/movetoarchive.koplugin/main.lua
@@ -46,7 +46,10 @@ function MoveToArchive:addToMainMenu(menu_items)
                 text = _("Go to archive folder"),
                 callback = function()
                     if not self.archive_dir_path then
-                        self:showNoArchiveConfirmBox()
+                        local ok_cb = function()
+                            self:openFileBrowser(self.archive_dir_path)
+                        end
+                        self:showNoArchiveConfirmBox(ok_cb)
                         return
                     end
                     self:openFileBrowser(self.archive_dir_path)
@@ -80,6 +83,11 @@ function MoveToArchive:moveToArchive()
     self:commonProcess(true, move_done_text)
 end
 
+-- End of Document
+function MoveToArchive:onMoveToArchive()
+    self:commonProcess(true)
+end
+
 function MoveToArchive:copyToArchive()
     local copy_done_text = _("Book copied.\nDo you want to open it from the archive folder?")
     self:commonProcess(false, copy_done_text)
@@ -87,7 +95,10 @@ end
 
 function MoveToArchive:commonProcess(is_move_process, moved_done_text)
     if not self.archive_dir_path then
-        self:showNoArchiveConfirmBox()
+        local ok_cb = function()
+            self:commonProcess(is_move_process, moved_done_text)
+        end
+        self:showNoArchiveConfirmBox(ok_cb)
         return
     end
     local document_full_path = self.ui.document.file
@@ -97,7 +108,9 @@ function MoveToArchive:commonProcess(is_move_process, moved_done_text)
     self.settings:saveSetting("last_copied_from_dir", self.last_copied_from_dir)
     self.settings:flush()
 
-    self.ui:onClose()
+    if moved_done_text then
+        self.ui:onClose()
+    end
     if is_move_process then
         FileManager:moveFile(document_full_path, self.archive_dir_path)
         FileManager:moveFile(DocSettings:getSidecarDir(document_full_path), self.archive_dir_path)
@@ -108,15 +121,17 @@ function MoveToArchive:commonProcess(is_move_process, moved_done_text)
     local dest_file = string.format("%s%s", self.archive_dir_path, filename)
     ReadHistory:updateItemByPath(document_full_path, dest_file) -- (will update "lastfile" if needed)
     ReadCollection:updateItemByPath(document_full_path, dest_file)
-    UIManager:show(ConfirmBox:new{
-        text = moved_done_text,
-        ok_callback = function ()
-            ReaderUI:showReader(dest_file)
-        end,
-        cancel_callback = function ()
-            self:openFileBrowser(self.last_copied_from_dir)
-        end,
-    })
+    if moved_done_text then
+        UIManager:show(ConfirmBox:new{
+            text = moved_done_text,
+            ok_callback = function ()
+                ReaderUI:showReader(dest_file)
+            end,
+            cancel_callback = function ()
+                self:openFileBrowser(self.last_copied_from_dir)
+            end,
+        })
+    end
 end
 
 function MoveToArchive:setArchiveDirectory()
@@ -129,12 +144,13 @@ function MoveToArchive:setArchiveDirectory()
     }:chooseDir()
 end
 
-function MoveToArchive:showNoArchiveConfirmBox()
+function MoveToArchive:showNoArchiveConfirmBox(success_callback)
     UIManager:show(ConfirmBox:new{
         text = _("No archive directory.\nDo you want to set it now?"),
         ok_text = _("Set archive folder"),
         ok_callback = function()
             self:setArchiveDirectory()
+            success_callback()
         end,
     })
 end
```

grasegger:
Wow I wish I knew it was so easy #firstTimeContributor 😅

Thanks for the work and the example.

NiLuJe:
`:loadPlugins` is called each time ReaderUi or FileManager is instantiated (cf., the "plugin loaded" logs: RD is ReaderUI, FM is FileManager).

So, :loadPlugins computes the list of load*able* plugins, returns a pointer to their chunk of code, then `:createPluginInstance` is called on each of them, ensuring they're instantiated properly.

:loadPlugins itself caches the result during the UI/FM instance lifetime, and resets the lists otherwise.

NiLuJe:
Another approach for `isPluginLoaded` would have been to walk `self.enabled_plugins`, but that would have meant just that: that the plugin is *enabled* and *loadable*, not that it's actually instantiated, which is what we care about here: we want to be sure it can receive Events.

NiLuJe:
(Updated patch above to actually close the document first, and go back to the FM properly).

NiLuJe:
Said patch update made me realize than we can also simply check for `self.ui[plugin_name]`, but, eh, a proper method in PL probably looks better ;p.

Frenzie:
> Said patch update made me realize than we can also simply check for self.ui[plugin_name]

I didn't realize you didn't. I thought you wanted a more "proper" approach.

NiLuJe:
I had completely forgotten ^^. (And I think I'm the one who made those accessible that way, IIRC xD).

NiLuJe:
> It came back to me when I was thinking if there was a way to do that from within PL itself.

And that'd be trickier, because that would imply pinning potentially stale pointers to plugin instances in memory from somewhere that doesn't have destructors, unlike FM/UI.

poire-z:
Just mentionning another way to do plugin <> core interactions: #6851.
It would avoid having core to have to know about a plugin named "move2archive": the plugin would just register itself in readerstatus as something that can do something on end of book.
Dunno if/how that could work with this context.

NiLuJe:
@poire-z: The reverse approach is more interesting for stuff where we *know* we'll have a significant number of Plugins that might want to plug into something else.

Here, this is more of a one-off.

(Plus, I like the locality of this approach here, makes the UI-building cleared, as you *know* the stock stuff may inherit plugin stuff, and from which one. If it were hidden away in the plugin, you'd have to do an extra rg to see if any plugin actually register stuff here).

NiLuJe:
Also, re: https://github.com/koreader/koreader/pull/7607#issuecomment-827666501

c.f., 738a437

Requires calling an explicit destructor, which is not pretty, but since only FM & ReaderUI are (and will probably ever be) concerned, eh.

poire-z:
(I was just mentionning this alternative idea in case you didn't remember it - I don't really care much about plugins, so I trust your choice :)

But I don't like `:dtor()` ! Wh nt mak tha nme readbl ? :)

NiLuJe:
I can offer you `fini` instead ;p.

poire-z:
Or just `:clear()`, as in KOReader lingo :) Or `:cleanUp()`.

NiLuJe:
Eh, I kinda like it, despite the nastiness. the way I see it, `registerModule` llows the *machine* to access the plugin, by making it belong to the top-level widget containers, meaning Events & finalizers propagate properly; and this allows *humans* to access 'em, without relying on that primarily machine-oriented implementation detail.

Or, something. :D.

poire-z:
But the single object PluginLoader is not ever finalized/destructed, in a human or Lua PoV. It's the same object that you just clear some sub-elements. So, clear() feels better than :finalize() :)

NiLuJe:
There's a point to be mode for both PoVs ;).

I definitely agree with you, but I like finalizer because it *also* semantically implies that this has something to do with Lua memory management.

PR #7609, opened 2021-04-27
Gestures menu optimization

hius07:
There are 11 `Gesture manager` submenu items, that creates the second page with `Spread and pinch` alone.
Propose to move `Short diagonal swipe` into `One-finger swipe` (it is the one-finger swipe actually) and get a nice one-page menu.

---

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7609)


Frenzie (approved):
> get a nice one-page menu.

That's a bit device-dependent, but in any case agreed with the sorting.

yparitcher (approved):

poire-z (approved):

PR #7610, opened 2021-04-27
Optimize cache for cover-image

zwim:
A better strategy for cleaning the cache.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7610)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -288,12 +288,11 @@ function CoverImage:cleanCache()
         print("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx remove very oldest: "..files[1].name)
         cache_count = cache_count - 1
         cache_size_KiB = cache_size_KiB - files[1].size
-        table.remove(files, 1)  -- delete entry
 
         -- see function description 2.b)
         if cache_size_KiB > self.cover_image_cache_maxsize * 1024 then
             -- find the maximum number of oldest files to delete (to fit space requirement)
-            local index = 1
+            local index = 2 -- start with second file, because the first (oldest) was deleted in 2.a)
             local largest_oldest_files = {}
             local best_cache_size_KiB = cache_size_KiB
             while best_cache_size_KiB > self.cover_image_cache_maxsize * 1024 and self.cover_image_cache_maxsize ~= 0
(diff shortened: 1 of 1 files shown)


NiLuJe:
That kinda feels super over-engineered for no real good reason, but, ho hum.

zwim:
The reason would be: If there are different sized jpg's or if someone mixes bmp with jpg ... on a cache cleanup a lot of cached files get deleted.
This could also happen if someone decreases the cache size.
Here I try too keep the cache filled as max as possible and just delete what is really necessary. (Keeping in mind to delete the at least the oldest file). 

But if you think KISS is better ...

NiLuJe:
I get why you went to the trouble, I'm just not entirely sure it's worth it in terms of maintenance burden *and* actual real-world use cases.

As a somewhat similar reference, the ScreenSavers hack on Kindle essentially does this in Cover mode, and it *never* clears the cache on its own (instead, there's a button that lets the user choose to do so if need be).

zwim:
Is it OK to convert this PR to draft (maybe someone finds it usefull ;) )?

NiLuJe:
I mean, it's a plugin, you're actually the one maintaining it, so, this is just me telling *I* wouldn't do that.

I'd actually be vetoing it if it were in core, but this is a Plugin, so, this is just a friendly comment ;).

zwim:
I will convert it to draft and think about it. Thank you for your comment.

Frenzie:
> I'd actually be vetoing it if it were in core, but this is a Plugin, so, this is just a friendly comment ;).

We do have to be careful with that though — people tend to take some random plugin as their jumping off point.

Frenzie:
I'm on board with this seems awful complicated and therefore more error prone to the simple existing method.

Galunid:
From a Onyx Nova 2 perspective, I don't really think cache is needed in this plugin. Images are generated fast enough that I don't notice any performance issues. I'm not sure as to how fast it works on Pocketbooks/Tolinos. I believe that keeping *n* most recent covers should be more than enough. Is the new algorithm better performance wise? Probably yes, is it worth it at the cost of readability? Probably not. To understand the previous code I needed a glance, it was extremely simple and straightforward, while I spend a few minutes and still don't understand what exactly happens in the revised version. If you were to abandon the plugin, someone would have to spend much more time to understand what happens if they wanted to change it. If you still want to go ahead with it, perhaps making a few helper functions would improve readability

Galunid:
Also it seems to duplicate https://github.com/koreader/koreader/pull/7611, not sure if it's intended

zwim:
Thank you all for your comments. I will close this now.

@galunid: #7611 will be the survivor of this PR. :))
At least on Tolino the cover creation is slow, a cache noticeable speeds up consecutive book openings. If someone has a faster device, the cache does no harm (and can be disabled).

@frenzie @niluje: You are right, this method is more complicated than the original one. In terms of maintainability over the years the older one is prefferable.

PR #7611, opened 2021-04-27
Better strings

zwim:
This is just a minor improvement in the menu display if the cachsize is turned off.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7611)

### DIFF
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -17,7 +17,6 @@ local UIManager = require("ui/uimanager")
 local RenderImage = require("ui/renderimage")
 local Screen = require("device").screen
 local T = require("ffi/util").template
-local Util = require("util")
 local WidgetContainer = require("ui/widget/container/widgetcontainer")
 local ffiutil = require("ffi/util")
 local lfs = require("libs/libkoreader-lfs")
@@ -473,7 +472,7 @@ function CoverImage:menuEntryCache()
                 text_func = function()
                     local number
                     if self.cover_image_cache_maxsize > 0 then
-                        number = Util.getFriendlySize(self.cover_image_cache_maxsize * 1e6)
+                        number = util.getFriendlySize(self.cover_image_cache_maxsize * 1e6)
                     elseif self.cover_image_cache_maxsize == 0 then
                         number = _("unlimited")
                     else
@@ -496,7 +495,7 @@ function CoverImage:menuEntryCache()
                 help_text_func = function()
                     local cache_count, cache_size_KB
                         = self:getCacheFiles(self.cover_image_cache_path, self.cover_image_cache_prefix)
-                    return T(_("The cache contains %1 files and uses %2."), cache_count, Util.getFriendlySize(cache_size_KB * 1000))
+                    return T(_("The cache contains %1 files and uses %2."), cache_count, util.getFriendlySize(cache_size_KB * 1000))
                 end,
                 callback = function()
                     UIManager:show(ConfirmBox:new{
--- a/plugins/coverimage.koplugin/main.lua
+++ b/plugins/coverimage.koplugin/main.lua
@@ -496,11 +496,7 @@ function CoverImage:menuEntryCache()
                 help_text_func = function()
                     local cache_count, cache_size_KB
                         = self:getCacheFiles(self.cover_image_cache_path, self.cover_image_cache_prefix)
-<<<<<<< HEAD
                     return T(_("The cache contains %1 files and uses %2."), cache_count, Util.getFriendlySize(cache_size_KB * 1000))
-=======
-                    return T(_("The cache contains %1 files and uses %2 MB."), cache_count, math.floor((cache_size_KB + 999) / 1000))
->>>>>>> 00e621f4... translatable, change KiB to KB, MiB to MB
                 end,
                 callback = function()
                     UIManager:show(ConfirmBox:new{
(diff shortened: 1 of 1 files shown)


Frenzie (changes requested):
Not sure what you're going for here, but… no? That's significantly worse. ;-)

plugins/coverimage.koplugin/main.lua
```diff
@@ -472,13 +472,13 @@ function CoverImage:menuEntryCache()
                 text_func = function()
                     local number
                     if self.cover_image_cache_maxsize > 0 then
-                        number = self.cover_image_cache_maxsize
+                        number = self.cover_image_cache_maxsize .. "MiB"
```
Frenzie, line 475: Perhaps you meant something like this? 
```suggestion
                        number = T(_("%1 MB"), self.cover_image_cache_maxsize)
```

NB Using MB (as in 1000 rather than 1024) because:
https://github.com/koreader/koreader/blob/86126d042f81e566006072ce310a5493cc403fae/frontend/util.lua#L939-L942

Frenzie: Which really just means use the util func, btw. ;-)

zwim:
> Not sure what you're going for here, but… no? That's significantly worse. ;-)

Concerning your comment with `number = T(_("%1 MB"), self.cover_image_cache_maxsize)` the suggested solution is worse. But the menu would look better as now. It would be
`Maximum size of cached covers (off)` and not `Maximum size of cached covers (offMB)`.

I will change all `MiB` and KiB to `MB` and `KB` and test it again before committing. (From the user point you are absolutely right, we should use corresponding units. Alltough limiting a cache on powers of 1000 is not a good idea ;) )

Frenzie:
If it's wrong then so is your proposal, because it's the same thing except translatable. ;-)

Frenzie:
I was just saying you can use http://koreader.rocks/doc/modules/util.html#getFriendlySize for display ^_^

plugins/coverimage.koplugin/main.lua
```diff
@@ -496,7 +496,11 @@ function CoverImage:menuEntryCache()
                 help_text_func = function()
                     local cache_count, cache_size_KB
                         = self:getCacheFiles(self.cover_image_cache_path, self.cover_image_cache_prefix)
+<<<<<<< HEAD
```
jimman2003, line 499: Additions from patch util... Plz remove

plugins/coverimage.koplugin/main.lua
```diff
@@ -496,7 +496,11 @@ function CoverImage:menuEntryCache()
...
                     return T(_("The cache contains %1 files and uses %2."), cache_count, Util.getFriendlySize(cache_size_KB * 1000))
+=======
+                    return T(_("The cache contains %1 files and uses %2 MB."), cache_count, math.floor((cache_size_KB + 999) / 1000))
+>>>>>>> 00e621f4... translatable, change KiB to KB, MiB to MB
```
jimman2003, line 503: Same here

plugins/coverimage.koplugin/main.lua
zwim: Thank you, that was a failed rebase.

plugins/coverimage.koplugin/main.lua
```diff
@@ -17,6 +17,7 @@ local UIManager = require("ui/uimanager")
 local RenderImage = require("ui/renderimage")
 local Screen = require("device").screen
 local T = require("ffi/util").template
+local Util = require("util")
```
Frenzie, line 20:
```suggestion
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -472,13 +473,13 @@ function CoverImage:menuEntryCache()
                 text_func = function()
                     local number
                     if self.cover_image_cache_maxsize > 0 then
-                        number = self.cover_image_cache_maxsize
+                        number = Util.getFriendlySize(self.cover_image_cache_maxsize * 1e6)
```
Frenzie, line 476:
```suggestion
                        number = util.getFriendlySize(self.cover_image_cache_maxsize * 1e6)
```

plugins/coverimage.koplugin/main.lua
```diff
@@ -493,9 +494,9 @@ function CoverImage:menuEntryCache()
             {
                 text = _("Clear cached covers"),
                 help_text_func = function()
-                    local cache_count, cache_size_KiB
+                    local cache_count, cache_size_KB
                         = self:getCacheFiles(self.cover_image_cache_path, self.cover_image_cache_prefix)
-                    return T(_("The cache contains %1 files and uses %2 MiB."), cache_count, math.floor((cache_size_KiB + 1023) / 1024))
+                    return T(_("The cache contains %1 files and uses %2."), cache_count, Util.getFriendlySize(cache_size_KB * 1000))
```
Frenzie, line 499:
```suggestion
                    return T(_("The cache contains %1 files and uses %2."), cache_count, util.getFriendlySize(cache_size_KB * 1000))
```

Frenzie (changes requested):

plugins/coverimage.koplugin/main.lua
zwim: I see, that `util` is already included. Thanks.

Frenzie (approved):

Galunid (approved):

poire-z:
(Please, @zwim or the person that merges such PRs, prefix your commit with `CoverImage plugin: ...`, the last related commits with their super short and no context make master history quite cryptic.)

Frenzie:
My bad, I usually adjust while squashing.

zwim:
Thank you

PR #7612, opened 2021-04-27
Kindle: Enable AutoSuspend plugin

NiLuJe:
Kindles are not flagged canPowerOff, although that's technically not
entirely warranted, but sorta makes sense.

The Plugin already handles that sanely, and will only expose/honor the
suspend timer.

(That wasn't the case before #7400, which possibly explains the omission).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7612)

### DIFF
--- a/plugins/autosuspend.koplugin/main.lua
+++ b/plugins/autosuspend.koplugin/main.lua
@@ -1,6 +1,7 @@
 local Device = require("device")
 
 if not Device:isCervantes() and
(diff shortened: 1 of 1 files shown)


NiLuJe:
xref: https://gitter.im/koreader/koreader?at=60881f81bdc0bc32e61fd63b

Frenzie (approved):

PR #7616, opened 2021-04-28
bump crengine: more granular font weights

### LINKED ISSUE #7539: Hyphenation not working in Georgian text (Android)
* KOReader version: 2021.03
* Device: toco_eea and Boox Poke 3

#### Issue
Enabling hyphenation for Georgian language does nothing. It doesn't even show the hyphens already added with Calibre plugin Hyphenate this!

#### Steps to reproduce
Open any book written in Georgian, try to enable hyphenation.

[Example book (with soft-hyphens already in it)](https://github.com/koreader/koreader/files/6297593/kale.blumkvisti.zip)
Uploading it as .zip because .epub isn't supported.

https://user-images.githubusercontent.com/457700/114413938-b2b8c380-9bae-11eb-8012-033eab9afd46.mp4

photo5449421455847699683

poire-z:
Includes:
- MathML: a few minor fixes https://github.com/koreader/crengine/pull/435
- (Upstream) lvtext: fix possible index out of range
- Fonts: RegisterExternalFont() should take a documentId
- Fonts: fix: letter-spacing should not be applied on diacritic
- (Upstream) Fonts: more granular synthetic weights
- Fonts: synthesized weights: tweak some comments
- Fonts: keep hinting with synthetic weight
- Fonts: fix synthesized weight inconsitencies
- Fonts: fix getFontFileNameAndFaceIndex()
- Fonts: adds LVFontMan::RegularizeRegisteredFontsWeights()
- Fonts: handle synth_weight tweaks in glyph/glyphinfo slots
- (Upstream) Fonts: fix some compiler warnings
- Fix hyphenation on Armenian and Georgian text https://github.com/koreader/crengine/pull/436 Closes #7539.

I decided to not care about the migration from the old boolean `font_embolden` to the replacement number `font_base_weight` - this will let people who had <kbd>bold</kbd> know that there is this new setting with more granular weights, as they'll probably enjoy 1.5 more than the old "bold"=3.

<kbd>image</kbd>
(<kbd>2</kbd> in the screenshots has been corrected to <kbd>+2</kbd>, too lazy to remake the screenshot)

I got rid of <kbd>+3</kbd>, I wanted to get rid of <kbd>+2</kbd> but the toggle felt a bit naked then :)

Rewording suggestions welcome.

@yparitcher : any idea if we renaming/removing a dispatcher setting/action (font_weight) can cause issues if a user has been using it?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7616)

### DIFF
--- a/frontend/ui/data/creoptions.lua
+++ b/frontend/ui/data/creoptions.lua
@@ -3,6 +3,7 @@ local Screen = Device.screen
 local optionsutil = require("ui/data/optionsutil")
 local _ = require("gettext")
 local C_ = _.pgettext
+local T = require("ffi/util").template
 
 -- Get font size numbers as a table of strings
 local tableOfNumbersToTableOfStrings = function(numbers)
@@ -493,16 +494,6 @@ Note that your selected font size is not affected by this setting.]]),
         icon = "appbar.contrast",
         options = {
             {
-                name = "font_weight",
-                name_text = _("Font Weight"),
-                toggle = {_("regular"), _("bold")},
-                values = {0, 1},
-                default_value = 0,
-                args = {0, 1},
-                event = "ToggleFontBolder",
-                name_text_hold_callback = optionsutil.showValues,
-            },
-            {
                 name = "font_gamma",
                 name_text = _("Contrast"),
                 buttonprogress = true,
@@ -527,6 +518,39 @@ Note that your selected font size is not affected by this setting.]]),
                                    41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56 };
                     value_step = 1,
                 },
+            },
+            {
+                name = "font_base_weight",
+                name_text = _("Font Weight"),
+                toggle = { "-1", "-½", "0", "+½", "+1", "+1½", "+3" },
+                values = { -1, -0.5, 0, 0.5, 1, 1.5, 3 },
+                args = { -1, -0.5, 0, 0.5, 1, 1.5, 3 },
+                default_value = 0,
+                event = "SetFontBaseWeight",
+                more_options = true,
+                more_options_param = {
+                    value_min = -3,
+                    value_max = 5.5,
+                    value_step = 0.25,
+                    precision = "%+.2f",
+                    value_hold_step = 1,
+                },
+                help_text = _([[Set the font weight delta from "regular" to apply to all fonts.
+
+- 0 will use the "regular (400)" variation of a font.
+- +1 will use the "medium (500)" variation of a font if available
+- +3 will use the "bold (700)" variation of a font if available
+If a font variation is not available, as well as for fractional adjustments, a font variation will be synthesized from the nearest available weight of the font.]]),
+                help_text_func = function(configurable, document)
+                    local font_face = document:getFontFace()
+                    local available_weights = cre.getFontFaceAvailableWeights(font_face)
+                    return T(_("The default font '%1' is available in %2."), font_face, table.concat(available_weights, ", "))
+                end,
+                name_text_hold_callback = optionsutil.showValues,
+                name_text_true_values = true,
+                show_true_value_func = function(val)
+                    return string.format("%d", 400+val*100)
+                end,
             },
             {
                 name = "font_hinting",
--- a/frontend/apps/reader/modules/readerfont.lua
+++ b/frontend/apps/reader/modules/readerfont.lua
@@ -127,10 +127,10 @@ function ReaderFont:onReadSettings(config)
                   or 22
     self.ui.document:setFontSize(Screen:scaleBySize(self.font_size))
 
-    self.font_embolden = config:readSetting("font_embolden")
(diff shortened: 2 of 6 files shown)


NiLuJe:
Oh, I quite like what you settled on to show what the current font supports ;).

frontend/ui/data/creoptions.lua
```diff
@@ -528,6 +519,39 @@ Note that your selected font size is not affected by this setting.]]),
...
+                event = "SetFontBaseWeight",
+                more_options = true,
+                more_options_param = {
+                    value_min = -3,
+                    value_max = 5.5,
+                    value_step = 0.25,
+                    precision = "%+.2f",
+                    value_hold_step = 1,
+                },
+                help_text = _([[Set the font weight delta from "regular" to apply to all fonts.
```
NiLuJe, line 538: That's technically a a weight [class](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass), but since we're not actually displaying that to the user, I'm okay with the simplification ;).

frontend/ui/data/creoptions.lua
```diff
@@ -528,6 +519,39 @@ Note that your selected font size is not affected by this setting.]]),
...
+
+- 0 will use the "regular (400)" variation of a font.
+- 1 will use the "medium (500)" variation of a font if available
+- 3 will use the "bold (700)" variation of a font if available
+If a font variation is not available, and for fractional deltas, a font will be synthesized from the nearest available weight of the font.]]),
```
NiLuJe, line 543: `as well as for fractional deltas`

NiLuJe: Could also possibly keep the `variation` thing going, since that's going to be more and more commonly used with the advent of variable fonts, so, `a font variation will be synthesized`

frontend/ui/data/creoptions.lua
```diff
@@ -528,6 +519,39 @@ Note that your selected font size is not affected by this setting.]]),
...
+                help_text_func = function(configurable, document)
+                    local font_face = document:getFontFace()
+                    local available_weights = cre.getFontFaceAvailableWeights(font_face)
+                    return T(_("The default font '%1' is available in these real weights: %2"), font_face, table.concat(available_weights, ", "))
```
NiLuJe, line 547: Oh, there I'd possibly use `weight class`, though ;).

frontend/ui/data/creoptions.lua
Frenzie: Maybe just a `fractional adjustment`?

poire-z:
Updated help text with suggestions:

<kbd>image</kbd>

frontend/ui/data/creoptions.lua
poire-z: Should I keep the "real" in there ? in these _real_ weight classes ?

poire-z:
Does this <kbd>+3</kbd> look odd, in the serie of numbers ?

<kbd>image</kbd>

<kbd>+2</kbd>, synthesized from the bold, is often thiner (so, surprising and soon forgotten) than the <kbd>+1 1/2</kbd> synthesized from the regular - and we mention `3` in the help_text.
(Looks like I should add some `+` in the help text before the numbers.)

Frenzie:
It does seem odd, but I'm not sure if there's an alternative.

frontend/ui/data/creoptions.lua
NiLuJe: I agree that it feels slightly clunky, but I couldn't really think of anything better. Perhaps getting rid of it entirely does the trick, though, the context should make it clear.

NiLuJe:
If only just because the old setting exposed Bold, and not SemiBold, yeah, I'd keep 3 instead of 2 ;).

frontend/ui/data/creoptions.lua
poire-z: "The default font 'Noto Sans' is available in 400, 700." ? Feels a bit short without "real", one could think +1 and +2 won't work.

poire-z: But might be OK, above we state "If a font variation is not available, a font variation will be synthesized...".

poire-z:
image

poire-z:
Or just "in 400 700" to avoid having to handle singular or plural "in weights 400." :/

yparitcher:
> @yparitcher : any idea if we renaming/removing a dispatcher setting/action (font_weight) can cause issues if a user has been using it?

should be fine, #6785 should have fixed any issues.

frontend/ui/data/creoptions.lua
NiLuJe: Oh, no, sorry, I meant just omitting "real" ;).

poire-z: Oh :/
Anybody feel free to correct this in his next PR if "available in 400, 700." feels like broken english :)

Frenzie: `available in: x, y`?

PR #7618, opened 2021-04-29
A silly solution to make whisper touch works on kindle voyage

Hzj-jie:
by restarting koreader after resume

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader/7618)


Hzj-jie:
This change is not ready to review, see discussion: https://github.com/koreader/koreader/discussions/7619.

poire-z:
What is "whisper touch" ? Even Google can't tell me :)

Hzj-jie:
Ah, I do not know where this term comes from, in koreader it's called WhisperTouch (https://github.com/koreader/koreader/blob/master/frontend/device/kindle/device.lua#L638). It's the not-that-physical page turn buttons on kindle voyage (the white dot and line on both sides of the kindle in the following image).

image

